PR #812, opened 2020-07-13
Use 64bit information in GTargetProcess

antonrohr:
The information if a process is used in `OrbitApp::Disassemble` (App.cpp line 335)

@dpallotti this is what I talked about in the standup this morning


antonrohr:
I think your comment on the bug is very valid, I am going to file an extra bug for that

PR #813, opened 2020-07-14
Use variables for flags in conan profiles

beckerhe:
Using variables allows the repository-provided profile to be used as a
template for own modifications. For example I could have a profile
`debug`:

```conan
include(clang9_debug)

[env]
CMAKE_EXPORT_COMPILE_COMMANDS=On
CONAN_CMAKE_GENERATOR=Ninja
CXXFLAGS=$CXX_FLAGS -fcolor-diagnostics
CFLAGS=$C_FLAGS -fcolor-diagnostics
```

I'm introducing this now, because I want to make the default-profiles updatable (by using an include).

More information about profile-variables and profile-includes is here: https://docs.conan.io/en/latest/reference/profiles.html#profile-includes


beckerhe:
@IrinaShkviro I'm sorry that you're always my victim for conan-related reviews.

IrinaShkviro:
> @IrinaShkviro I'm sorry that you're always my victim for conan-related reviews.

It's not a problem at all!

PR #814, opened 2020-07-14
Ensure correct version of conan on every CI run

beckerhe:
This PR adds an update step to the kokoro build scripts ensuring that
the installed version of conan is always new enough. This decouples
the conan installation from the docker container or VM. Updating
conan takes roughly 7 seconds.

After this change we can start using more modern conan features. This first thing coming will probably be "build profiles", which greatly simplifies cross compiling and will solve the cmake_installer/cmake package problem. Ref: https://docs.conan.io/en/latest/reference/profiles.html#build-profiles-and-host-profiles


beckerhe:
@IrinaShkviro And here is one more. :see_no_evil: :grimacing:

IrinaShkviro:
> This PR adds an update step to the kokoro build scripts ensuring that
> the installed version of conan is always new enough. This decouples
> the conan installation from the docker container or VM. Updating
> conan takes roughly 7 seconds.
> 
> After this change we can start using more modern conan features. This first thing coming will probably be "build profiles", which greatly simplifies cross compiling and will solve the cmake_installer/cmake package problem. Ref: https://docs.conan.io/en/latest/reference/profiles.html#build-profiles-and-host-profiles

Looks good, thanks! Just one question: do we want update required conan version in bootstrap-orbit scripts, to be in line with CI builds?

beckerhe:
> > This PR adds an update step to the kokoro build scripts ensuring that
> > the installed version of conan is always new enough. This decouples
> > the conan installation from the docker container or VM. Updating
> > conan takes roughly 7 seconds.
> > After this change we can start using more modern conan features. This first thing coming will probably be "build profiles", which greatly simplifies cross compiling and will solve the cmake_installer/cmake package problem. Ref: https://docs.conan.io/en/latest/reference/profiles.html#build-profiles-and-host-profiles
> 
> Looks good, thanks! Just one question: do we want to update required conan version in bootstrap-orbit scripts, to be in line with CI builds?

Good question. I was debating that with myself. I was a bit reluctant and didn't want to make any changes for us as developers if not needed, but having these versions in sync is also an important thing. I think I will change it! Thanks.

PR #815, opened 2020-07-14
Fix flaky ThreadPoolTest by increasing SleepFor

antonrohr:
Test 
```
cd build_default
ctest --repeat-until-fail 1000
```
or 
```
build_default/bin/OrbitBaseTests --gtest_repeat=1000 | grep -i "Failed"
```

### DIFF
--- a/OrbitBase/ThreadPoolTest.cpp
+++ b/OrbitBase/ThreadPoolTest.cpp
@@ -121,9 +121,9 @@ TEST(ThreadPool, CheckTtl) {
         absl::Milliseconds(50)));
   }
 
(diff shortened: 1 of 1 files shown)


OrbitBase/ThreadPoolTest.cpp
```diff
@@ -121,9 +121,9 @@ TEST(ThreadPool, CheckTtl) {
         absl::Milliseconds(50)));
   }
 
-  // +1 because there might be some nanoseconds between action is complete and
+  // +10 because there might be some nanoseconds between action is complete and
```
dimitry-, line 124: apparently milliseconds...

PR #816, opened 2020-07-14
Improve error handling for ElfFile

dimitry-:
It now returns human-readable error-message when ElfFile::Create fails.

### DIFF
--- a/ElfUtils/ElfFile.cpp
+++ b/ElfUtils/ElfFile.cpp
@@ -122,7 +122,7 @@ ErrorMessageOr<ModuleSymbols> ElfFileImpl<ElfT>::LoadSymbols() const {
   // TODO: if we want to use other sections than .symtab in the future for
   //       example .dynsym, than we have to change this.
   if (!has_symtab_section_) {
-    return ErrorMessage("Elf file does not contain a .symtab section.");
+    return ErrorMessage("ELF file does not have a .symtab section.");
   }
   bool symbols_added = false;
 
@@ -165,7 +165,7 @@ ErrorMessageOr<ModuleSymbols> ElfFileImpl<ElfT>::LoadSymbols() const {
   }
   if (!symbols_added) {
     return ErrorMessage(
-        "Unable to load symbols from elf file, not even a single symbol of "
+        "Unable to load symbols from ELF file, not even a single symbol of "
         "type function found.");
   }
   return module_symbols;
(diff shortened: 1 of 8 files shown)


ElfUtils/ElfFile.cpp
```diff
@@ -247,15 +246,18 @@ std::unique_ptr<ElfFile> ElfFile::CreateFromBuffer(std::string_view file_path,
           llvm::object::ObjectFile::createObjectFile(buffer->getMemBufferRef());
 
   if (!object_file_or_error) {
-    return nullptr;
+    return ErrorMessage(
+        absl::StrFormat("Unable to create elf-file \"%s\": %s", file_path,
```
dpallotti, line 250: Maybe "ELF file" instead of "elf-file" in these messages?

ElfUtils/ElfFileTest.cpp
```diff
@@ -162,7 +165,19 @@ TEST(ElfFile, CreateFromBuffer) {
...
+
+TEST(ElfFile, FileDoesNotExist) {
+  std::string executable_path = Path::GetExecutablePath();
+  std::string file_path = executable_path + "/testdata/does_not_exist";
+
+  auto elf_file = ElfFile::Create(file_path);
+  ASSERT_FALSE(elf_file);
+  EXPECT_EQ(elf_file.error().message(),
+            absl::StrFormat(
+                "Unable to create elf-file \"%s\": No such file or directory",
```
dpallotti, line 181: Do we want a test that precisely compares the error message? It would fail as soon as one character is changed in the error and requires the two strings to be kept in sync.

beckerhe: I kind of agree. If you want to ensure that certain elements are correct, maybe we should just "grep" for them?

dimitry-: done.

ElfUtils/ElfFile.cpp
dimitry-: done.

ElfUtils/ElfFileTest.cpp
```diff
@@ -2,15 +2,16 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
 
-#include <absl/strings/str_format.h>
```
antonrohr, line 5: Just out of curiosity, when do you choose <> vs ""?

dimitry-: <> for stl/standard c/linux/windows headers, "" for the rest.

PR #819, opened 2020-07-14
remove unused from Timer

IrinaShkviro:
* Remove unused types from Timer::Type

* Remove unused `m_CaptureID` and methods from Timer class

### DIFF
--- a/OrbitGl/ThreadTrack.cpp
+++ b/OrbitGl/ThreadTrack.cpp
@@ -55,7 +55,7 @@ void ThreadTrack::Draw(GlCanvas* canvas, bool picking) {
 std::string ThreadTrack::GetExtraInfo(const Timer& timer) {
   std::string info;
   static bool show_return_value = absl::GetFlag(FLAGS_show_return_values);
-  if (show_return_value && (timer.GetType() == Timer::NONE)) {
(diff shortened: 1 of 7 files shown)


OrbitGl/ThreadTrack.cpp
```diff
@@ -114,13 +114,6 @@ void ThreadTrack::SetTimesliceText(const Timer& timer, double elapsed_us,
           absl::StrFormat("%s %s %s", name, extra_info.c_str(), time.c_str());
 
       text_box->SetText(text);
-    } else if (timer.m_Type == Timer::INTROSPECTION) {
```
ronaldfw, line 117: The introspection stuff should actually be left in, we are using it. It needs to be enabled by defining ORBIT_TRACING_ENABLED in OrbitCore/Introspection.cpp, maybe that's why you missed it?

ronaldfw:
@pierricgimmig should have a look at this to make sure we are not deleting something that introspection or manual instrumentation needs.

OrbitGl/ThreadTrack.cpp
IrinaShkviro: Yes, sorry for this, reverted.

PR #820, opened 2020-07-14
Fix incorrect preset loading

antonrohr:
This fixes an error that the preset loading did not work when symbols were already loaded.
It also displays an additional error message when a process did not load a module that was saved in the preset.

### DIFF
--- a/OrbitGl/App.cpp
+++ b/OrbitGl/App.cpp
@@ -6,6 +6,7 @@
 
 #include <absl/flags/flag.h>
 #include <absl/strings/str_format.h>
+#include <absl/strings/str_join.h>
 
 #include <chrono>
 #include <cmath>
@@ -532,7 +533,7 @@ void OrbitApp::LoadPreset(const std::shared_ptr<Preset>& preset) {
     return;
   }
   if (!SelectProcess(Path::GetFileName(preset->m_ProcessFullPath))) {
-    SendErrorToUi("Session loading failed",
+    SendErrorToUi("Preset loading failed",
                   absl::StrFormat("The process \"%s\" is not running.",
                                   preset->m_ProcessFullPath));
     return;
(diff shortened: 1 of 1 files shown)


OrbitGl/App.cpp
```diff
@@ -805,15 +805,27 @@ void OrbitApp::LoadModules(int32_t process_id,
...
-  std::vector<std::shared_ptr<Module>> modules;
+  std::vector<std::shared_ptr<Module>> modules_to_load;
   for (const auto& pair : preset->m_Modules) {
     const std::string& module_path = pair.first;
     const auto& module = process->GetModuleFromPath(module_path);
-    if (module != nullptr && !module->IsLoaded()) {
-      modules.emplace_back(std::move(module));
+    if (module == nullptr) {
+      SendErrorToUi(
+          "Session loading incomplete",
```
ronaldfw, line 814: Session -> Preset?

antonrohr: Yes, of course. I changed it here and where I copied it from

OrbitGl/App.cpp
```diff
@@ -805,15 +805,27 @@ void OrbitApp::LoadModules(int32_t process_id,
...
-      modules.emplace_back(std::move(module));
+    if (module == nullptr) {
+      SendErrorToUi(
+          "Preset loading incomplete",
+          absl::StrFormat("The module \"%s\" is not loaded by process \"%s\".",
+                          module_path, process->GetName()));
+      continue;
     }
+    if (module->IsLoaded()) {
+      CHECK(module->m_Pdb != nullptr);
```
ronaldfw, line 820: Why is this a CHECK and not an error case that we must handle?

OrbitGl/App.cpp
```diff
@@ -805,15 +805,27 @@ void OrbitApp::LoadModules(int32_t process_id,
...
                                      const std::shared_ptr<Preset>& preset) {
-  std::vector<std::shared_ptr<Module>> modules;
+  std::vector<std::shared_ptr<Module>> modules_to_load;
   for (const auto& pair : preset->m_Modules) {
     const std::string& module_path = pair.first;
     const auto& module = process->GetModuleFromPath(module_path);
-    if (module != nullptr && !module->IsLoaded()) {
-      modules.emplace_back(std::move(module));
+    if (module == nullptr) {
+      SendErrorToUi(
```
ronaldfw, line 813: This does potentially spam the user with dialog boxes, does it? Maybe we should have one error dialog at the end?

OrbitGl/App.cpp
antonrohr: Because a loaded module always has a pdb. Thats kind of the definition if a module being loaded. In an easier world maybe `IsLoaded` would return if the module has a pdb, instead of having a separate bool flag. But also, pdb in general is a windows term and currently a very shallow layer and one of the things we are trying to replace with newer data models

ronaldfw: Ok, sounds good.

OrbitGl/App.cpp
antonrohr: I would judge this as a case that should rarely happen, but your idea is good nevertheless. I aggregated the modules that are not found and send the error after the loop

PR #821, opened 2020-07-14
Add build information to Orbit-builds

beckerhe:
This PR adds build information to Orbit-UI and OrbitService.

This information can be accessed by either calling `--version` on the command line (both Orbit-UI and OrbitService) or by opening the about dialog (only Orbit-UI).

The following pieces of information are shown:
- Version and Commit hash
- Compiler
- Build machine's hostname
- Build machine's operating system
- Build timestamp

Furthermore this PR cleans up the cmake-generated version-files. For reasons there are two. One in OrbitCore and one in OrbitQt. Both had previously the name version.h resp. Version.h which clashed on Windows. So I renamed them to OrbitCoreVersion.h resp. OrbitQtVersion.h.

This is how it looks like for Orbit-UI and OrbitService on the command line:
image

This is how it looks like in the About dialog:
image


ronaldfw:
Haven't looked at the code changes, but wanted to say this is very cool!

beckerhe:
> Haven't looked at the code changes, but wanted to say this is very cool!

Thanks. :innocent:

pierricgimmig:
Where is OrbitQtVersion used by the way?

beckerhe:
> Where is OrbitQtVersion used by the way? Is it still needed?

Good point. Yes, it's currently used for the version number shown in the window title bar. And there was a reason for not using OrbitCore::GetVersion in that place. I've been skimming through the commit messages but I couldn't find the reason and I start to believe that it's not needed anymore. I'm going to remove it. Hold on...

PR #823, opened 2020-07-15
Fix ProcessList file path

antonrohr:
The file path of a process used to be determined by reading
/proc/pid/cmdline which contains the command how the process was
started. This is not absolute file path, but can be relative (./app).
Now it is determined by reading the link /proc/pid/exe which points to
the executable.

Also added todo comments to LinuxUtils. 

Besides just making a thing "more correct", this also fixes a bug where preset loading is not working as intended when the process is started with a different command. for example: /path/to/somewhere/game vs ./game

### DIFF
--- a/OrbitCore/LinuxUtils.cpp
+++ b/OrbitCore/LinuxUtils.cpp
@@ -8,6 +8,7 @@
 #include <asm/unistd.h>
 #include <cxxabi.h>
 #include <dirent.h>
+#include <linux/limits.h>
 #include <linux/perf_event.h>
 #include <linux/types.h>
 #include <linux/version.h>
@@ -33,6 +34,7 @@
 
 #include "ElfUtils/ElfFile.h"
 #include "OrbitBase/Logging.h"
+#include "OrbitBase/Result.h"
 #include "OrbitBase/SafeStrerror.h"
 #include "Path.h"
 #include "Utils.h"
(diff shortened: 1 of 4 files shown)


OrbitService/ProcessList.cpp
```diff
@@ -74,11 +74,21 @@ outcome::result<void, std::string> ProcessList::Refresh() {
...
-    process.set_full_path(cmdline.substr(0, cmdline.find('\0')));
-
     std::replace(cmdline.begin(), cmdline.end(), '\0', ' ');
     process.set_command_line(cmdline);
 
+    // /proc/pid/exe is a link to the executable
+    const std::filesystem::path exe_path = directory_entry.path() / "exe";
+    // -n means no new line
+    const auto exe_result = LinuxUtils::ExecuteCommand(
+        absl::StrFormat("readlink -n %s", exe_path.string()));
```
dimitry-, line 84: just readlink (from unistd.h) - no need to execute commands

dimitry-: and probably extract it to linux utils something like GetExecutabePath(int32_t pid)

dimitry-: and have GetExecutablePath call it on linux with getpid()

antonrohr: done. I also filed some bugs for refactoring and added TODOs for things I did not want to touch in this PR

OrbitCore/LinuxUtils.cpp
```diff
@@ -163,9 +167,26 @@ outcome::result<std::unordered_map<pid_t, double>> GetCpuUtilization() {
...
+//-----------------------------------------------------------------------------
+ErrorMessageOr<std::string> GetExecutablePath(int32_t pid) {
+  char buffer[PATH_MAX];
+
+  ssize_t length = readlink(absl::StrFormat("/proc/%d/exe", pid).c_str(),
+                            buffer, sizeof(buffer));
+  if (length == -1) {
+    return ErrorMessage(absl::StrFormat(
+        "Unable to get executable path of process with pid: %d, error: %s", pid,
+        strerror(errno)));
```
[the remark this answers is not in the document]
antonrohr: done

PR #825, opened 2020-07-15
Use the task_rename,task_newtask tracepoints for thread names, unify (a bit) handling of GPU tracepoints and other tracepoints

dpallotti:
Here's the change that tracks the two `task` tracepoints for thread names.
In particular, the last commit refactors quite a bit how GPU tracepoints are handled.

---
#### Rename PerfEventSampleRaw to RawSamplePerfEvent for consistency
For consistency with the other `PerfEvent` classes.
#### Use perf_event_open to obtain task:task_rename tracepoint events
Only use them to print logs for now.
#### Use the task_rename tracepoint to notify listener_ of a new ThreadName
Instead of periodically calling `UpdateThreadNamesIfDelayElapsed`.
#### Also use the task:task_newtask tracepoint for thread names
When a new thread is created, it inherits the name of the parent. We want to
catch this as the thread doesn't necessarily sets its own name.
#### Make GPU events part of PerfEvent's hierarchy like other tracepoints
This aims to unify how we handle GPU tracepoints and other tracepoints.
Moves the format of GPU tracepoints to `Tracepoints.h`.
Makes the events part of the `PerfEvent` class's hierarchy.
Moves parsing GPU tracepoints from `GpuTracepointEventProcessor` to their
`PerfEvent` class.
Unifies reading tracepoints with `ConsumeTracepointPerfEvent`.
Moves GPU tracepoints' ids out of `GpuTracepointEventProcessor`.
Moves the classification of GPU tracepoints to `TracerThread`, which now records
separate stream ids for GPU tracepoints too.

### DIFF
--- a/OrbitLinuxTracing/PerfEvent.h
+++ b/OrbitLinuxTracing/PerfEvent.h
@@ -380,12 +380,28 @@ class TracepointPerfEventBase : public PerfEvent {
   }
 };
 
+class TaskNewtaskPerfEvent
+    : public TracepointPerfEventBase<task_newtask_tracepoint> {
+ public:
+  void Accept(PerfEventVisitor* visitor) override;
+
+  // The tracepoint format calls this "pid" but it's effectively the thread id.
+  pid_t GetTid() const { return tracepoint_data.pid; }
+
+  const char* GetComm() const { return tracepoint_data.comm; }
+
+  uint64_t GetStreamId() const {
+    return ring_buffer_record.sample_id.stream_id;
+  }
+
+  uint32_t GetCpu() const { return ring_buffer_record.sample_id.cpu; }
+};
+
 class TaskRenamePerfEvent
     : public TracepointPerfEventBase<task_rename_tracepoint> {
  public:
   void Accept(PerfEventVisitor* visitor) override;
 
-  // The tracepoint format calls this "pid" but it's effectively the thread id.
   pid_t GetTid() const { return tracepoint_data.pid; }
 
   const char* GetOldComm() const { return tracepoint_data.oldcomm; }
--- a/OrbitLinuxTracing/PerfEvent.h
+++ b/OrbitLinuxTracing/PerfEvent.h
@@ -9,6 +9,7 @@
 #include <memory>
 
 #include "Function.h"
+#include "KernelTracepoints.h"
 #include "MakeUniqueForOverwrite.h"
 #include "PerfEventRecords.h"
 
@@ -366,6 +367,35 @@ class MapsPerfEvent : public PerfEvent {
  private:
   uint64_t timestamp_;
   std::string maps_;
+};
+
+template <typename TracepointData>
+class TracepointPerfEventBase : public PerfEvent {
+ public:
+  perf_event_raw_sample_fixed ring_buffer_record;
+  TracepointData tracepoint_data;
+
+  uint64_t GetTimestamp() const override {
+    return ring_buffer_record.sample_id.time;
+  }
+};
+
+class TaskRenamePerfEvent
+    : public TracepointPerfEventBase<task_rename_tracepoint> {
+ public:
+  void Accept(PerfEventVisitor* visitor) override;
+
+  // The tracepoint format calls this "pid" but it's effectively the thread id.
+  pid_t GetTid() const { return tracepoint_data.pid; }
+
+  const char* GetOldComm() const { return tracepoint_data.oldcomm; }
+  const char* GetNewComm() const { return tracepoint_data.newcomm; }
+
+  uint64_t GetStreamId() const {
+    return ring_buffer_record.sample_id.stream_id;
+  }
+
+  uint32_t GetCpu() const { return ring_buffer_record.sample_id.cpu; }
 };
 
 class RawSamplePerfEvent {
--- a/OrbitLinuxTracing/PerfEvent.h
+++ b/OrbitLinuxTracing/PerfEvent.h
@@ -369,56 +369,166 @@ class MapsPerfEvent : public PerfEvent {
   std::string maps_;
 };
 
-template <typename TracepointData>
-class TracepointPerfEventBase : public PerfEvent {
+class TracepointPerfEvent : public PerfEvent {
  public:
+  explicit TracepointPerfEvent(uint32_t size)
+      : tracepoint_data{make_unique_for_overwrite<uint8_t[]>(size)} {}
+
   perf_event_raw_sample_fixed ring_buffer_record;
-  TracepointData tracepoint_data;
+  std::unique_ptr<uint8_t[]> tracepoint_data;
 
   uint64_t GetTimestamp() const override {
     return ring_buffer_record.sample_id.time;
   }
+
+  uint64_t GetStreamId() const {
+    return ring_buffer_record.sample_id.stream_id;
+  }
+
+  uint32_t GetCpu() const { return ring_buffer_record.sample_id.cpu; }
+
+  uint16_t GetTracepointId() const { return GetTracepointCommon().common_type; }
+
+ protected:
+  const tracepoint_common& GetTracepointCommon() const {
+    return *reinterpret_cast<const tracepoint_common*>(tracepoint_data.get());
+  }
+
+  template <typename TracepointData>
+  const TracepointData& GetTypedTracepointData() const {
+    return *reinterpret_cast<const TracepointData*>(tracepoint_data.get());
+  }
 };
 
-class TaskNewtaskPerfEvent
-    : public TracepointPerfEventBase<task_newtask_tracepoint> {
+class TaskNewtaskPerfEvent : public TracepointPerfEvent {
  public:
+  explicit TaskNewtaskPerfEvent(uint32_t tracepoint_size)
+      : TracepointPerfEvent(tracepoint_size) {}
+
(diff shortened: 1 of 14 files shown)


ronaldfw:
I'll take a look at this either today or tomorrow, sorry for the delay.

dpallotti:
> I'll take a look at this either today or tomorrow, sorry for the delay.

Don't worry, there's no rush for this and it's not blocking anything else.

OrbitLinuxTracing/Tracepoints.h
```diff
@@ -0,0 +1,65 @@
+// Copyright (c) 2020 The Orbit Authors. All rights reserved.
```
ronaldfw, line 1: Should the filename maybe reflect that these are actually kernel tracepoints?

OrbitLinuxTracing/TracerThread.cpp
```diff
@@ -257,23 +257,71 @@ bool TracerThread::OpenSampling(const std::vector<int32_t>& cpus) {
   return true;
 }
 
-bool TracerThread::InitGpuTracepointEventProcessor() {
-  int amdgpu_cs_ioctl_id = GetTracepointId("amdgpu", "amdgpu_cs_ioctl");
-  if (amdgpu_cs_ioctl_id == -1) {
-    return false;
+bool TracerThread::OpenRingBuffersForTracepointAndRedirectOnExistingIfNecessary(
```
ronaldfw, line 260: Can this easily be split into two methods? The name already indicates this method is doing too many things. 

OpenRingBufffersForTracepoint()
RedirectOnExistingBufferIfPossible()

I would also suggest to change Necessary to Possible. It's not necessary to redirect onto an existing ring buffer but it's advantageous to do so.

OrbitLinuxTracing/TracerThread.cpp
```diff
@@ -257,23 +257,71 @@ bool TracerThread::OpenSampling(const std::vector<int32_t>& cpus) {
...
+    tracing_fds->push_back(fd.second);
+    uint64_t stream_id = perf_event_get_id(fd.second);
+    tracepoint_ids->insert(stream_id);
   }
-  int dma_fence_signaled_id =
-      GetTracepointId("dma_fence", "dma_fence_signaled");
-  if (dma_fence_signaled_id == -1) {
-    return false;
+
+  // Redirect all tracepoints on the same cpu to a single ring buffer.
```
ronaldfw, line 287: It seems this part could be done in a separate method, right?

OrbitLinuxTracing/TracerThread.cpp
```diff
@@ -314,32 +364,49 @@ bool TracerThread::OpenRingBufferForGpuTracepoint(
...
+  for (int fd : amdgpu_sched_run_job_fds) {
+    tracing_fds_.push_back(fd);
+    amdgpu_sched_run_job_ids_.insert(perf_event_get_id(fd));
+  }
+  for (int fd : dma_fence_signaled_fds) {
     tracing_fds_.push_back(fd);
-    gpu_tracing_ids_.insert(perf_event_get_id(fd));
+    dma_fence_signaled_ids_.insert(perf_event_get_id(fd));
   }
+  // TODO: Consider redirecting all GPU events on a CPU on the same ring buffer.
```
ronaldfw, line 409: Isn't this done above? Or do you mean to do all three types of events on the same ring buffer (per CPU).

OrbitLinuxTracing/TracerThread.cpp
```diff
@@ -390,6 +457,8 @@ void TracerThread::Run(
     perf_event_open_errors |= uprobes_event_open_errors;
   }
 
+  perf_event_open_errors |= !OpenTracepoints(cpuset_cpus);
```
ronaldfw, line 460: Nit: This was already here, but I'm wondering if you could make this more readable by changing this to be:

bool perf_event_open_all_succeeded = true;

Then you don't need the ! and can just write

perf_event_open_all_succeeded &= OpenTracepoints(cpuset_cpus);

OrbitLinuxTracing/TracerThread.cpp
```diff
@@ -762,29 +868,19 @@ void TracerThread::ProcessDeferredEvents() {
   }
 }
 
-void TracerThread::UpdateThreadNamesIfDelayElapsed() {
+void TracerThread::PollThreadNames() {
```
ronaldfw, line 871: Just "GetThreadNames"? We are not polling the thread names regularly.

OrbitLinuxTracing/PerfEventOpen.cpp
```diff
@@ -154,6 +154,9 @@ void* perf_event_open_mmap_ring_buffer(int fd, uint64_t mmap_length) {
 int tracepoint_event_open(const char* tracepoint_category,
                           const char* tracepoint_name, pid_t pid, int32_t cpu) {
   int tp_id = GetTracepointId(tracepoint_category, tracepoint_name);
+  if (tp_id == -1) {
```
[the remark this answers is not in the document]
ronaldfw: Ok, sounds good.

OrbitLinuxTracing/TracerThread.cpp
ronaldfw: Yes, I think if we change one, all should be changed. But you make a good point that something uprobes_event_open_succeeded doesn't make sense if you don't actually aim to open uprobes. Let's keep it as it is.

OrbitLinuxTracing/TracerThread.cpp
ronaldfw: Good point, sounds good.

OrbitLinuxTracing/PerfEvent.h
```diff
@@ -368,11 +369,166 @@ class MapsPerfEvent : public PerfEvent {
...
+    return ring_buffer_record.sample_id.stream_id;
+  }
+
+  uint32_t GetCpu() const { return ring_buffer_record.sample_id.cpu; }
+
+  uint16_t GetTracepointId() const { return GetTracepointCommon().common_type; }
+
+ protected:
+  const tracepoint_common& GetTracepointCommon() const {
+    return *reinterpret_cast<const tracepoint_common*>(tracepoint_data.get());
```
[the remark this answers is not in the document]
ronaldfw: We had at some point discussed the "memcpy instead of reinterpret_cast" idea. Also, std::bitcast is in C++20, so eventually we may just want to move to using that. 

Both memcpy and bit_cast are discussed a bit here:
https://en.cppreference.com/w/cpp/numeric/bit_cast

Since we have reinterpret_cast in a lot of places like this, I would say it's ok to leave as is, but we should take a close look at this at some point.

OrbitLinuxTracing/PerfEvent.h
```diff
@@ -368,11 +369,166 @@ class MapsPerfEvent : public PerfEvent {
...
+  }
+
+  template <typename TracepointData>
+  const TracepointData& GetTypedTracepointData() const {
```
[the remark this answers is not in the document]
ronaldfw: Initially I was thinking something like CastToTracepointData, but it's not really appropriate as you are not casting the object, you are casting a member and returning the cast version of that. I'm afraid I don't have a better suggestion, so just leave as-is.

OrbitLinuxTracing/TracerThread.cpp
```diff
@@ -257,46 +257,82 @@ bool TracerThread::OpenSampling(const std::vector<int32_t>& cpus) {
...
+      ring_buffer_fds_per_cpu->emplace(cpu, ring_buffer_fd);
+    }
   }
-  int amdgpu_sched_run_job_id =
-      GetTracepointId("amdgpu", "amdgpu_sched_run_job");
-  if (amdgpu_sched_run_job_id == -1) {
-    return false;
+}
+
+bool TracerThread::OpenRingBuffersForTracepoint(
```
ronaldfw, line 285: Maybe one final suggestion: One way of reducing the amount of arguments is to combine some of them into a struct, if that can be done in a meaningful way. It can also become more readable if you group arguments that belong together into a single structure. 

Feel free to ignore this, I don't think you would gain a lot here (if anything).

dpallotti: Maybe `tracepoint_category`, `tracepoint_name` and `cpus` could be grouped, but the others not really as some are output parameters and others input/output parameters. So I'll leave it as it is as I don't think it would improve a lot.

OrbitLinuxTracing/PerfEvent.h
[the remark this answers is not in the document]
dpallotti: So I'll leave it as it is now but let's discuss it offline soon as I'd also like to understand more. My understanding so far has been that since one of the types is an array of bytes then `reinterpret_cast` works correctly, but it's likely I got it wrong.

OrbitLinuxTracing/TracerThread.cpp
ronaldfw: Agree.

OrbitLinuxTracing/PerfEvent.h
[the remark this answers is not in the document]
ronaldfw: Yes, you may be right that in this case it's actually guaranteed to work correctly.

PR #826, opened 2020-07-15
Changing the way to scroll + refactoring jumps

karupayun:
Creating JumpScope and JumpDirection and making only a function JumpToBox which jumps depending on these parameters.

Create a new function, HorizontallyScrollIntoView. When selecting a text box, depending on its position, it shows in the screen following something similar to the rule of thirds.

### DIFF
--- a/OrbitGl/CaptureWindow.cpp
+++ b/OrbitGl/CaptureWindow.cpp
@@ -489,23 +489,35 @@ void CaptureWindow::KeyPressed(unsigned int a_KeyCode, bool a_Ctrl,
         break;
       case 18:  // Left
         if (a_Shift) {
-          time_graph_.OnShiftLeft();
+          time_graph_.JumpToNeighborBox(Capture::GSelectedTextBox,
+                                        TimeGraph::JumpDirection::kPrevious,
+                                        TimeGraph::JumpScope::kSameFunction);
         } else {
-          time_graph_.OnLeft();
+          time_graph_.JumpToNeighborBox(Capture::GSelectedTextBox,
+                                        TimeGraph::JumpDirection::kPrevious,
+                                        TimeGraph::JumpScope::kSameThread);
         }
         break;
       case 20:  // Right
         if (a_Shift) {
-          time_graph_.OnShiftRight();
+          time_graph_.JumpToNeighborBox(Capture::GSelectedTextBox,
+                                        TimeGraph::JumpDirection::kNext,
+                                        TimeGraph::JumpScope::kSameFunction);
         } else {
-          time_graph_.OnRight();
+          time_graph_.JumpToNeighborBox(Capture::GSelectedTextBox,
+                                        TimeGraph::JumpDirection::kNext,
+                                        TimeGraph::JumpScope::kSameThread);
         }
         break;
       case 19:  // Up
-        time_graph_.OnUp();
+        time_graph_.JumpToNeighborBox(Capture::GSelectedTextBox,
+                                      TimeGraph::JumpDirection::kTop,
+                                      TimeGraph::JumpScope::kSameThread);
         break;
       case 21:  // Down
-        time_graph_.OnDown();
+        time_graph_.JumpToNeighborBox(Capture::GSelectedTextBox,
+                                      TimeGraph::JumpDirection::kDown,
+                                      TimeGraph::JumpScope::kSameThread);
         break;
     }
   }
--- a/OrbitGl/TimeGraph.cpp
+++ b/OrbitGl/TimeGraph.cpp
@@ -225,8 +225,8 @@ void TimeGraph::HorizontallyMoveIntoView(const TextBox* text_box,
     return;
   }
 
-  double start = MicroSecondsFromTicks(m_SessionMinCounter, timer.m_Start);
-  double end = MicroSecondsFromTicks(m_SessionMinCounter, timer.m_End);
+  double start = MicroSecondsFromTicks(capture_min_timestamp_, timer.m_Start);
+  double end = MicroSecondsFromTicks(capture_min_timestamp_, timer.m_End);
   double mid = start + ((end - start) / 2.0);
 
   // Mirror the final center position if we have to move left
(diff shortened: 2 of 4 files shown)


OrbitGl/CaptureWindow.cpp
```diff
@@ -489,16 +489,16 @@ void CaptureWindow::KeyPressed(unsigned int a_KeyCode, bool a_Ctrl,
         break;
       case 18:  // Left
         if (a_Shift) {
-          time_graph_.OnShiftLeft();
+          time_graph_.JumpToBox(TimeGraph::Previous, TimeGraph::SameFunction);
```
ronaldfw, line 492: I would consider keeping the names OnShiftLeft, OnLeft etc. and then have them be a simple implementation like 

void OnShiftLeft() {
  JumpToBox(TimerGraph::Next, TimeGraph::SameFunction);
}

etc. 

In particular, if you keep the OnUp() and OnDown() methods (as below).

OrbitGl/TimeGraph.cpp
```diff
@@ -211,6 +211,31 @@ void TimeGraph::PanTime(int a_InitialX, int a_CurrentX, int a_Width,
   NeedsUpdate();
 }
 
+void TimeGraph::HorizontallyScrollIntoView(const TextBox* text_box) {
```
ronaldfw, line 214: "Scroll" in the title suggests an animation to me. Jump or move?

OrbitGl/TimeGraph.cpp
```diff
@@ -211,6 +211,31 @@ void TimeGraph::PanTime(int a_InitialX, int a_CurrentX, int a_Width,
...
+  const Timer& timer = text_box->GetTimer();
+
+  if (IsVisible(timer)) {
```
ronaldfw, line 217: I feel this is a bit inconsistent. If a timer is not visible, this method will put the textbox, on the 1/3 line in the view, if it is visible it will stay wherever it is. Is there a reason you don't want to always move such that the textbox is on the 1/3 line? You'd probably want to rename the method if you change it.

OrbitGl/TimeGraph.cpp
```diff
@@ -211,6 +211,31 @@ void TimeGraph::PanTime(int a_InitialX, int a_CurrentX, int a_Width,
...
+    return;
+  }
+
+  // Where in the screen is going to be the center of the text box
+  double ratio = 0.3;
```
ronaldfw, line 222: Consider making this a parameter. Depending on the direction you are moving in, you might want the textbox be at 1/3 (when moving left) or at 2/3 (when moving right) (or maybe the other way around...).

OrbitGl/TimeGraph.cpp
ronaldfw: Or maybe this could be (with the suggestions below):

```
// Aligns center of textbox with the horizontal position of the view given by 'position' 
// (which defines the position in [0,1] coordinates). 
HorizontallyAlignInView(const TextBox* text_box, double center_position) {...}
```

OrbitGl/TimeGraph.cpp
ronaldfw: Ah, you already to that below if I understand right.

OrbitGl/TimeGraph.h
```diff
@@ -59,12 +59,14 @@ class TimeGraph {
...
                double a_InitialTime);
+  void HorizontallyScrollIntoView(const TextBox* text_box);
   void VerticallyScrollIntoView(const TextBox* text_box);
   double GetTime(double a_Ratio);
   double GetTimeIntervalMicro(double a_Ratio);
-  void Select(const TextBox* a_TextBox) { SelectRight(a_TextBox); }
-  void SelectLeft(const TextBox* a_TextBox);
-  void SelectRight(const TextBox* a_TextBox);
+  void Select(const TextBox* a_TextBox);
+  enum JumpScope { SameThread, SameFunction, SameThreadSameFunction };
```
beckerhe, line 67: Can you make them `enum class`?

OrbitGl/TimeGraph.cpp
```diff
@@ -726,41 +720,34 @@ void TimeGraph::SortTracks() {
   }
 }
 
-//----------------------------------------------------------------------------
-void TimeGraph::OnLeft() {
+const TextBox* TimeGraph::FindPrevious() {
```
ronaldfw, line 723: This is a very specialized method that works on the currently selected text box. Consider passing in the text box as a parameter or renaming to something like FindPreviousFromCurrentSelection() .

OrbitGl/TimeGraph.cpp
```diff
@@ -390,42 +415,11 @@ void TimeGraph::GetWorldMinMax(float& a_Min, float& a_Max) const {
...
-  double currentTimeWindowUs = m_MaxTimeUs - m_MinTimeUs;
-  m_RefTimeUs = MicroSecondsFromTicks(m_SessionMinCounter, timer.m_End);
-
-  static double ratio = 1.0;
-  double minTimeUs = m_RefTimeUs - ratio * currentTimeWindowUs;
-  double maxTimeUs = m_RefTimeUs + (1 - ratio) * currentTimeWindowUs;
-
-  SetMinMax(minTimeUs, maxTimeUs);
+void TimeGraph::Select(const TextBox* a_TextBox) {
+  TextBox* text_box = const_cast<TextBox*>(a_TextBox);
```
beckerhe, line 419: Is there a chance that we don't need the const_cast here? Probably we have to remove the `const` from several TextBox-pointers, but that feels more appropriate to me than having a const_cast here. (Yes, I know, that was here before, so not your fault, but maybe a good chance to clean that up. :-))

OrbitGl/TimeGraph.cpp
```diff
@@ -726,41 +720,34 @@ void TimeGraph::SortTracks() {
   }
 }
 
-//----------------------------------------------------------------------------
-void TimeGraph::OnLeft() {
+const TextBox* TimeGraph::FindPrevious() {
   TextBox* selection = Capture::GSelectedTextBox;
+  const TextBox* previous = nullptr;
```
ronaldfw, line 725: Consider adding direct return statements. That'd usually easier to read. 

That is: 
```
  if (selection) {
    const Timer& timer = selection->GetTimer();
    if (timer.m_Type == Timer::GPU_ACTIVITY) {
      return GetOrCreateGpuTrack(GetGpuTimelineHash(timer))->GetLeft(selection);
    } else {
      return GetOrCreateThreadTrack(timer.m_TID)->GetLeft(selection);
    }
  }
  return nullptr;
```

ronaldfw: This also makes it clear that you are returning a nullptr when you fall through the cases. Not that obvious with the current code.

OrbitGl/TimeGraph.cpp
```diff
@@ -726,41 +720,34 @@ void TimeGraph::SortTracks() {
...
+      previous = GetOrCreateThreadTrack(timer.m_TID)->GetLeft(selection);
     }
   }
-  NeedsUpdate();
+  return previous;
 }
 
-//----------------------------------------------------------------------------
-void TimeGraph::OnRight() {
+const TextBox* TimeGraph::FindNext() {
```
ronaldfw, line 738: Same here.

OrbitGl/TimeGraph.cpp
```diff
@@ -769,32 +756,32 @@ void TimeGraph::SelectAndZoom(const TextBox* text_box) {
   Select(text_box);
 }
 
-//----------------------------------------------------------------------------
-void TimeGraph::OnShiftLeft() {
+void TimeGraph::JumpToBox(JumpDirection jump_direction, JumpScope jump_scope) {
```
ronaldfw, line 759: Make it clear that you are jumping to a neighbor of the selected box. 

Signature could be:

`void JumpToNeighborBox(TextBox* starting_point, JumpDirection..., JumpScope) {}`

ronaldfw: I would also (as suggested) pass the "starting point" box as an argument.

OrbitGl/TimeGraph.cpp
beckerhe, line 738: Maybe FindNextOfSelected/FindPreviousOfSelected? I'm not sure about that, so feel free to ignore it, if you don't like it.

OrbitGl/TimeGraph.h
ronaldfw, line 67: I was wondering if a case is missing here. I guess we could add a "Global" scope here, which just means jumping to the next textbox.

ronaldfw, line 67: Just one more thing here: Enums should use the constant naming style, that is, 

enum ... {
  kSameThread, 
  kSameFunction, 
  ...
};

OrbitGl/TimeGraph.cpp
ronaldfw: +1

Or pass in the textbox as starting point.

OrbitGl/CaptureWindow.cpp
beckerhe: Ahh, I'm sorry. This was my idea. But I see your point with OnUp/OnDown.

ronaldfw: I think either way is fine, but would prefer consistency. 

Either keep the old names or rename OnUp/OnDown in a way consistent with the new names. 

I think that TimeGraph should not necessarily explicitly mention the buttons that had to be pressed to get a certain action  (OnLeft, OnRight), so I would prefer to rename OnUp/OnDown.

beckerhe: > I think that TimeGraph should not necessarily explicitly mention the buttons that had to be pressed to get a certain action (OnLeft, OnRight), so I would prefer to rename OnUp/OnDown.

Yes, that was also my thought.

OrbitGl/TimeGraph.cpp
karupayun: ok, I will change it.

OrbitGl/CaptureWindow.cpp
karupayun: Yes, was the original plan, but I preferred to start with the change of Left/Right to see opinions.

OrbitGl/TimeGraph.h
karupayun: Thanks!

karupayun: Yeah, will add it. Probably we don't want to make this kind of jump, but having it inside enum doesn't hurt.

OrbitGl/TimeGraph.cpp
karupayun: You are clearly right

OrbitGl/TimeGraph.cpp
karupayun: Let me explain it a little bit.
- Actually, we are moving horizontally as minimum as possible to see the text box completely. So, if the text box wasn't visible is always going to be in a side of the screen.
- On the other hand, if you press ctrl-f to find something in a google docs, it isn't going to move the screen if the word is visible (as we do), but if it isn't visible, the screen moves more than the minimum. It's nicer. In fact, it's a bit more complex how it moves, but the plan here if to make a general function which scrolls automatically to the left or the right depending on the text box position and also scrolling more than the minimum.

ronaldfw: Ah, I recall we discussed this. Sounds good to me. Would it make sense to move the logic for checking if the timer is visible up a bit? So that this function will always move a box to the same place in the view, no matter where it is, but the UI logic implements the behavior you suggest?

OrbitGl/TimeGraph.cpp
karupayun: For this, I have to erase several several const. All from ThreadTrack.h, all from GpuTrack.h... And it's only to erase this const_cast. Feels more appropiate, but too many changes.

On the other hand, we could change GSelectedTextBox to be "const TextBox *". What do you think?

OrbitGl/TimeGraph.cpp
karupayun: As we are using the word "jump", I would call the variable "from"

OrbitGl/CaptureWindow.cpp
karupayun: What do you think now?

OrbitGl/TimeGraph.cpp
beckerhe: If that works for GSelectedTextBox then yes. That's an even better solution!

beckerhe: Thanks for checking this out. I didn't want to impose more trouble. :see_no_evil:

OrbitGl/TimeGraph.cpp
```diff
@@ -726,111 +722,88 @@ void TimeGraph::SortTracks() {
...
-  if (selection) {
-    const TextBox* left = FindPreviousFunctionCall(selection->GetTimer().m_FunctionAddress,
-                                      selection->GetTimer().m_End);
-    if (left) {
-      SelectLeft(left);
-      VerticallyScrollIntoView(left);
+void TimeGraph::JumpToNeighborBox(TextBox* from, JumpDirection jump_direction,
+                                  JumpScope jump_scope) {
+  const TextBox* goal = nullptr;
+  if (from) {
```
beckerhe, line 734: Nit: Can you return early here? `if (!from) {...`

OrbitGl/TimeGraph.cpp
```diff
@@ -7,6 +7,7 @@
 #include <OrbitBase/Logging.h>
 
 #include <algorithm>
+#include <iomanip>
```
beckerhe, line 10: Do you need that include? My guess is, it was automatically added by clangd, wasn' it?

PR #831, opened 2020-07-16
Move to gtest 1.10.0

beckerhe:
This PR updates gtest to version 1.10.0 and make some smaller related changes:

- gtest is now a build requirement in conan (pure semantic change)
- We don't use the gtest-provided cmake config files, rather the integration provided by conan. This makes it more robust in cross compilation scenarios.
- We provide our own gtest_main target which provides the same main function implementation as gtest. This change is needed since because the previous point does not allow us to choose whether we want to link against the gtest-provided main function on a per-test basis.

The initial reason for updating gtest was a bug that turned out unrelated to gtest. But there is no need to waste that work. We have to update eventually.

@antonrohr Since we talked about gtest I guess it's your turn to review that. ;-) But there is no rush. I don't wanna merge that for 1.50.


conanfile.py
```diff
@@ -52,6 +50,11 @@ def config_options(self):
         elif not self.options.with_crash_handling:
             del self.options.crashdump_server
 
+    def build_requirements(self):
+        self.build_requires('protoc_installer/3.9.1@bincrafters/stable#0')
+        self.build_requires('grpc_codegen/1.27.3@orbitdeps/stable#ec39b3cf6031361be942257523c1839a')
+        self.build_requires('gtest/1.10.0#ef88ba8e54f5ffad7d706062d0731a40', force_host_context=True)
```
antonrohr, line 56: what is `force_host_context`?

cmake/tests.cmake
```diff
@@ -13,3 +13,30 @@ function(register_test)
...
+#include <cstdio>\n\
+#include <gtest/gtest.h>\n\
+\n\
+int main(int argc, char* argv[]) {\n\
+  printf(\"Running main() from %s\\n\", __FILE__)\;\n\
+  ::testing::InitGoogleTest(&argc, argv)\;\n\
+  return RUN_ALL_TESTS()\;\n\
+}\n\
+")
+
```
antonrohr, line 35: I am just curious why all this is necessary? I bet there is a good reason

conanfile.py
beckerhe: Conan has two contexts which are often the same, but differ for cross compilation. Build requirements belong by default to the build context. In a cross compilation environment, the host context refers to the target, so it's Linux for the ggp_ configurations, while the build context is the platform you're compiling on (Windows for example). Since gtest needs to be linked in the host context, the package also needs to be from the host context (Linux library in that example.)

protoc_installer and grpc_codegen are code generator which have to run on the build machine, so they don't get the flag. We need Windows executables in that example.

cmake/tests.cmake
beckerhe: I tried to describe that in the PR description. It's needed because the conan-generated cmake config does not support linking to some part of a package. (A conan feature which is currently in development and will be merged soon.) So we either have to globally enable linking against gtest with an integrated main function or provide or own main function and link against that in the cases where it's needed. Since we have one test case which provides their own main function in OrbitSshQt, I had to choose the latter.

beckerhe: That sounds a lot like a good reason for not moving to conan-generated cmake-config files. But doing that will allow us to fully support visual studio as soon as we moved all packages to conan-generated cmake-config files. Qt and GRPC are still missing.

conanfile.py
antonrohr: Makes a lot of sense. I was guessing it has to do with the crosscompilation

cmake/tests.cmake
antonrohr: tbh, I should have read the PR description more closely. Thank you for taking the time. The SshQt integration test was not on the top of my mind anymore. It all makes sense now

PR #832, opened 2020-07-16
Initial version of function / timer iterator in the UI

ronaldfw:
This is a first version of a function/timer iterator UI that allows to select instrumented functions and let users iterate over the timers of that function. As an additional feature, this UI allows to set multiple iterators and displays the total time between the different iterators. It is possible to advance all  iterators at once.  As an example use case, his allows to iterate over"frames" assuming one has a CPU function that defines frame boundaries (like vkQueuePresentKHR). 

This sets the baseline and basic structure of the implementation, but there are more things to come and I'll keep iterating on the UI and code. Some ideas for the future:
- Button to "align" iterators.
- Buttons for jumping to min/max etc. of function timers (same as menu entries). 
- Delta times between all iterators (not just first and last) and better coloring.
- Allows users to set iterator at start/center/end of timer.
- Customizable colors. 
- Heatmap of delta times.


OrbitQt/orbiteventiterator.cpp
```diff
@@ -0,0 +1,93 @@
...
+
+//-----------------------------------------------------------------------------
+void OrbitEventIterator::SetMaxCount(int max_count) {
+  max_count_ = max_count;
+  UpdateCountLabel();
+}
+
+//-----------------------------------------------------------------------------
+void OrbitEventIterator::SetIndex(int current_index) {
+  current_index_ = 0;
```
pierricgimmig, line 53: current_index_ = current_index;

OrbitQt/orbitlivefunctions.h
```diff
@@ -0,0 +1,46 @@
...
+// found in the LICENSE file.
+
+#ifndef ORBIT_LIVE_FUNCTIONS_
+#define ORBIT_LIVE_FUNCTIONS_
+
+#include <QLineEdit>
+#include <QWidget>
+
+#include "LiveFunctions.h"
+#include "OrbitEventIterator.h"
```
pierricgimmig, line 12: This should be all lowercase.

ronaldfw:
> * Can't filter instrumented functions in live tab
> * Using the search toolbar makes the application crash

That's odd, I was sure I fixed that in one of the last commits. Let me take a look why that is.

OrbitGl/TimeGraph.cpp
```diff
@@ -566,11 +569,69 @@ void TimeGraph::Draw(GlCanvas* canvas, bool a_Picking) {
...
 
   m_NeedsRedraw = false;
 }
 
+void TimeGraph::DrawOverlay(GlCanvas* canvas, bool /*picking*/) {
+  float min_x = std::numeric_limits<float>::max();
+  float max_x = std::numeric_limits<float>::lowest();
+
+  uint64_t min_tick = std::numeric_limits<uint64_t>::max();
+  uint64_t max_tick = std::numeric_limits<uint64_t>::min();
```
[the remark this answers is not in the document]
ronaldfw: Yes, should get used to just use lowest in these cases, thanks.

OrbitGl/LiveFunctionsDataView.cpp
```diff
@@ -166,13 +169,14 @@ std::vector<std::string> LiveFunctionsDataView::GetContextMenu(
   if (enable_select) menu.emplace_back(MENU_ACTION_SELECT);
   if (enable_unselect) menu.emplace_back(MENU_ACTION_UNSELECT);
   if (enable_disassembly) menu.emplace_back(MENU_ACTION_DISASSEMBLY);
+  menu.emplace_back(MENU_ACTION_ITERATE);
```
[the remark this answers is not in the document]
ronaldfw: Yes, good point. I wanted to show an error dialog in that case (see further below), but handling it here is easier and better for users (just don't show the option if it's not possible to use it anyway).

ronaldfw:
> > * Can't filter instrumented functions in live tab
> > * Using the search toolbar makes the application crash
> 
> That's odd, I was sure I fixed that in one of the last commits. Let me take a look why that is.

Can you check this again, @pierricgimmig? I cannot reproduce this. If you still see this, can you let me know exact steps to reproduce?

ronaldfw:
> Hi Ronald, in general I think this looks good. Please see minor comments. On a higher level, one thing that I might have done differently is to not "clone" the existing orbitdataviewpanel in the orbitlivefunctions.ui. Maybe we could have created an "OrbitLiveFunctionsPanel" that would have contained an original orbitdataviewpanel plus the additional ui elements.
> 
> This would keep things centralized and we could then tweak the appearance of orbitdataviewpanels in Orbit once for all instead of duplicating work.
> 
> That being said, it's not the end of the world, and I'd be OK with submitting as is (after minor comments have been addressed) and then iterating if need be.

I'll take a look at this. I may have just cloned orbitdataviewpanel as it was easier for me to "bootstrap" the initial implementation this way. Now that I have something that is working, I can probably make this work in a cleaner way.

OrbitGl/TimeGraph.cpp
```diff
@@ -566,11 +569,69 @@ void TimeGraph::Draw(GlCanvas* canvas, bool a_Picking) {
...
+
+  float world_start_x = canvas->GetWorldTopLeftX();
+  float world_width = canvas->GetWorldWidth();
+
+  float world_start_y = canvas->GetWorldTopLeftY();
+  float world_height = canvas->GetWorldHeight();
+
+  for (const auto& current_textbox : overlay_current_textboxes_) {
+    double inv_time_window = 1.0 / GetTimeWindowUs();
```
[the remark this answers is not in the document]
ronaldfw: No, it doesn't, thanks.

OrbitQt/orbitlivefunctions.cpp
```diff
@@ -0,0 +1,128 @@
...
+  ui->iteratorLayout->addWidget(all_events_iterator_);
+}
+
+//-----------------------------------------------------------------------------
+OrbitLiveFunctions::~OrbitLiveFunctions() {
+  for (auto it_ui : iterator_uis) {
+    delete it_ui.second;
+  }
+  delete all_events_iterator_;
+  delete ui;
```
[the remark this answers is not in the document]
ronaldfw: Makes sense, thanks for the suggestion!

OrbitGl/LiveFunctions.h
```diff
@@ -0,0 +1,52 @@
...
+#define LIVE_FUNCTIONS_H_
+
+#include <functional>
+
+#include "LiveFunctionsDataView.h"
+#include "OrbitFunction.h"
+#include "TextBox.h"
+#include "absl/container/flat_hash_map.h"
+
+class LiveFunctions {
```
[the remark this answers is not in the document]
ronaldfw: Agree, I wasn't sure what to name it, I'll use your suggestion.

ronaldfw:
> > Hi Ronald, in general I think this looks good. Please see minor comments. On a higher level, one thing that I might have done differently is to not "clone" the existing orbitdataviewpanel in the orbitlivefunctions.ui. Maybe we could have created an "OrbitLiveFunctionsPanel" that would have contained an original orbitdataviewpanel plus the additional ui elements.
> > This would keep things centralized and we could then tweak the appearance of orbitdataviewpanels in Orbit once for all instead of duplicating work.
> > That being said, it's not the end of the world, and I'd be OK with submitting as is (after minor comments have been addressed) and then iterating if need be.
> 
> I'll take a look at this. I may have just cloned orbitdataviewpanel as it was easier for me to "bootstrap" the initial implementation this way. Now that I have something that is working, I can probably make this work in a cleaner way.

Fixing this was actually fairly straightforward. I'm not sure I'm doing it the most elegant way, but I'm now properly reusing OrbitDataViewPanel as a UI element in OrbitLiveFunctions.

OrbitGl/LiveFunctionsDataView.h
```diff
@@ -24,6 +26,8 @@ class LiveFunctionsDataView : public DataView {
   void OnDataChanged() override;
   void OnTimer() override;
 
+  void JumpToBox(const TextBox* box) const;
```
karupayun, line 29: I think that you have to remove this code, maybe is something old one after merging.

ronaldfw:
Thanks for the comments, @karupayun! 

> * Did you think about adding an iterator directly to a specific TextBox?
>   * Could have an option when right clicking on specific text box in a capture

I think that's a great idea! @pierricgimmig suggested something similar. Not sure how much work it is, I don't think we have context menus on the capture window yet, so maybe this is a bit more work?

>   * Or simpler, if there is already some text box selected (clicked or after jumping) from that function, AddIterator will select that box instead of the first occurrence.

Yes, I want to improve this and make it possible to jump to selected boxes, etc. 

> * Is it meant that when using iterators, you can't press any text box within the green range? I don't like that, but maybe there is a reason.

Should be fixed now; see also #834.

OrbitGl/TimeGraph.cpp
```diff
@@ -136,19 +136,22 @@ void TimeGraph::ZoomAll() {
...
 
-//-----------------------------------------------------------------------------
-void TimeGraph::Zoom(const TextBox* a_TextBox) {
-  const Timer& timer = a_TextBox->GetTimer();
-
-  double start = MicroSecondsFromTicks(m_SessionMinCounter, timer.m_Start);
-  double end = MicroSecondsFromTicks(m_SessionMinCounter, timer.m_End);
+void TimeGraph::Zoom(TickType min, TickType max) {
+  double start = MicroSecondsFromTicks(m_SessionMinCounter, min);
+  double end = MicroSecondsFromTicks(m_SessionMinCounter, max);
```
karupayun, line 141: I really liked the separation you made with Zoom.

I would like to have a parameter here (with a default value) that defines the percentage of the screen that the text boxes will occupy. It is actually 0.9 (or more precisely 1 / 1.1).

In addition, but less important, we can give the user the possibility to change this parameter (make it smaller or larger), so that he can see more to the sides after each zoom (if they wish).

karupayun:
> >   * Could have an option when right clicking on specific text box in a capture
> 
> I think that's a great idea! @pierricgimmig suggested something similar. Not sure how much work it is, I don't think we have context menus on the capture window yet, so maybe this is a bit more work?

I think so, I also though in making the same for jumping to the next/previous. Could be some future-feature.

> > * Is it meant that when using iterators, you can't press any text box within the green range? I don't like that, but maybe there is a reason.
> 
> Should be fixed now; see also #834.

Ahh, Hadn't merge your PR with master. Thanks!

OrbitGl/LiveFunctionsController.cpp
```diff
@@ -0,0 +1,150 @@
...
+    id_to_select_ = 0;
+  }
+  Move();
+}
+
+void LiveFunctionsController::AddIterator(Function* function) {
+  uint64_t id = next_iterator_id_;
+  ++next_iterator_id_;
+
+  auto function_address =
```
karupayun, line 138: As suggested in the comments, an easier modification to support adding an iterator in the selected text box could be something like this:

const TextBox* box = Capture::GSelectedTextBox;
if (!box){
  auto function_address = ...
  box = ....
}

ronaldfw: I'll try to add this, will do in a later PR, though. The main issue with just going to a random (selected) text box is that I don't know the index of that textbox (I display that in the UI, like "5 / 21"). I want to change that to just indicate roughly where in the timeline the iterator is (using the timestamps), but want to experiment with that a bit. Does that sound ok for you?

OrbitGl/TimeGraph.cpp
ronaldfw: Yes, definitely! We will need some UI elements for that, so I think it would be good to postpone to a later PR.

OrbitGl/LiveFunctionsDataView.h
ronaldfw: Thanks, this was just a left over. The implementation didn't exist anymore.

OrbitGl/LiveFunctionsController.cpp
karupayun: Perfect

PR #837, opened 2020-07-20
Added ModuleList refresh button

antonrohr:
This is to address b/156732790

After running into several problems trying to reload modules periodically I decided the best fix for now is to implement a refresh button for the module list.

These problems are because with the current way things are done (`DataView`) there is no good way of saving a multi selection across data changes. This is mainly because `OrbitTreeView` discards the content of the multi selection events (`selectionChanged`) and just passes on a single selection event. This could be changed, but I think this work is not really justified if we want implement a more appropriate Qt Item model soon.

### DIFF
--- a/OrbitGl/DataView.h
+++ b/OrbitGl/DataView.h
@@ -73,6 +73,8 @@ class DataView {
     return false;
   }
   virtual std::string GetLabel() { return ""; }
+  virtual bool HasRefreshButton() const { return false; }
+  virtual void OnRefreshButtonClicked() {}
   virtual void SetGlPanel(class GlPanel* /*a_GlPanel*/) {}
   virtual void LinkDataView(DataView* /*a_DataView*/) {}
   virtual bool ScrollToBottom() { return false; }
--- a/OrbitGl/ModulesDataView.cpp
+++ b/OrbitGl/ModulesDataView.cpp
@@ -14,10 +14,8 @@
 ABSL_FLAG(bool, enable_frame_pointer_validator, false,
           "Enable validation of frame pointers");
 
-//-----------------------------------------------------------------------------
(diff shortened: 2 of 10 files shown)


OrbitGl/ModulesDataView.cpp
```diff
@@ -196,6 +196,11 @@ void ModulesDataView::SetModules(int32_t process_id,
   OnDataChanged();
 }
 
+//-----------------------------------------------------------------------------
```
dimitry-, line 199: nit: drop this

OrbitQt/orbitdataviewpanel.cpp
```diff
@@ -70,3 +78,8 @@ void OrbitDataViewPanel::SetFilter(const QString& a_Filter) {
 void OrbitDataViewPanel::on_FilterLineEdit_textEdited(const QString& a_Text) {
   ui->treeView->OnFilter(a_Text);
 }
+
+//-----------------------------------------------------------------------------
```
dimitry-, line 82: and this

OrbitGl/DataView.h
```diff
@@ -73,6 +73,8 @@ class DataView {
     return false;
   }
   virtual std::string GetLabel() { return ""; }
+  virtual bool HasRefreshButton() { return false; }
```
beckerhe, line 76: const?

OrbitGl/ModulesDataView.cpp
antonrohr: removed it in quite a few places that I touched; done

OrbitQt/orbitdataviewpanel.cpp
antonrohr: done

OrbitGl/DataView.h
antonrohr: done

PR #838, opened 2020-07-20
Improvements to iterator UI: Buttons repeat when pressed down, location shown is time based

ronaldfw:
This PR combines two small improvements for the iterator UI:
1) The buttons just repeat firing a press down signal when the user keeps the button pressed down. This makes it easier to quickly move through the capture. Some variables and methods are renamed for more consistency. 
2) The current location of an iterator is now based on the timestamp and the global min/max timestamps. This may seem like a step back compared with the index, but it makes some future improvements easier. For example, jumping to min/max, setting the iterator to the currently selected textbox, and in general more arbitrary movements.

### DIFF
--- a/OrbitGl/LiveFunctionsController.cpp
+++ b/OrbitGl/LiveFunctionsController.cpp
@@ -26,7 +26,8 @@ void LiveFunctionsController::Move() {
     GCurrentTimeGraph->Zoom(min_max.first, min_max.second);
     if (current_textboxes_.find(id_to_select_) != current_textboxes_.end()) {
       GCurrentTimeGraph->Select(current_textboxes_[id_to_select_]);
-      GCurrentTimeGraph->VerticallyScrollIntoView(current_textboxes_[id_to_select_]);
+      GCurrentTimeGraph->VerticallyScrollIntoView(
+          current_textboxes_[id_to_select_]);
     } else {
       CHECK(false);
     }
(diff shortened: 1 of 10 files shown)


OrbitGl/TimeGraph.cpp
```diff
@@ -100,35 +100,35 @@ void TimeGraph::Clear() {
 double GNumHistorySeconds = 2.f;
 
 //-----------------------------------------------------------------------------
-bool TimeGraph::UpdateSessionMinMaxCounter() {
-  m_SessionMinCounter = LLONG_MAX;
+bool TimeGraph::UpdateCaptureMinMaxCounter() {
+  m_CaptureMinCounter = LLONG_MAX;
```
pierricgimmig, line 104: We should probably replace occurences of LLONG_MAX by std::numeric_limits<uint64_t>::max();  Doesn't need to be in this PR.

OrbitGl/LiveFunctionsController.cpp
```diff
@@ -147,4 +145,19 @@ void LiveFunctionsController::AddIterator(Function* function) {
     add_iterator_callback_(id, function);
   }
   Move();
+}
+
+TickType LiveFunctionsController::GetStartTime(uint64_t index) {
+  auto& it = current_textboxes_.find(index);
```
pierricgimmig, line 151: "const auto&" will fix the linux build

OrbitGl/TimeGraph.h
```diff
@@ -137,8 +144,8 @@ class TimeGraph {
   double m_RefTimeUs = 0;
   double m_MinTimeUs = 0;
   double m_MaxTimeUs = 0;
-  TickType m_SessionMinCounter = 0;
-  TickType m_SessionMaxCounter = 0;
+  TickType m_CaptureMinCounter = 0;
```
pierricgimmig, line 147: Should we use the new style capture_min_counter_ ?

OrbitGl/LiveFunctionsController.cpp
karupayun: yeah, that fix it!

OrbitGl/TimeGraph.cpp
ronaldfw: Done.

OrbitGl/LiveFunctionsController.cpp
ronaldfw: Thanks!

OrbitGl/TimeGraph.h
ronaldfw: Yes, of course, thanks!

PR #839, opened 2020-07-21
Avoid align if Y-coordinate is not set

karupayun:
Only temporal solution while solving the real problem.

### DIFF
--- a/OrbitGl/TimeGraph.cpp
+++ b/OrbitGl/TimeGraph.cpp
@@ -242,7 +242,10 @@ void TimeGraph::HorizontallyMoveIntoView(const TextBox* text_box,
 
 void TimeGraph::VerticallyMoveIntoView(const TextBox* text_box) {
   CHECK(text_box != nullptr);
-  // TODO: Sometimes the Y-coordinate is not set.
+  // TODO: Sometimes the Y-coordinate is not set. We only need to update the
+  // Y-coordinate from this text box
+  UpdatePrimitives();
+
   auto world_top_left_y = m_Canvas->GetWorldTopLeftY();
   auto top_margin =
       m_Layout.GetSchedulerTrackOffset() + m_Layout.GetVerticalMargin();


OrbitGl/TimeGraph.cpp
```diff
@@ -243,6 +243,8 @@ void TimeGraph::HorizontallyMoveIntoView(const TextBox* text_box,
 void TimeGraph::VerticallyMoveIntoView(const TextBox* text_box) {
   CHECK(text_box != nullptr);
   // TODO: Sometimes the Y-coordinate is not set.
+  if (text_box->GetPos() == Vec2::Zero()) return;
```
ronaldfw, line 246: I'm wondering if we should just pay the price here and call UpdatePrimitives(), as least temporarily until we have a proper solution.

karupayun: Yeah, me too. Anyway, I like this temporal solution. We hardly ever wants to VerticallyMove, probably only once. And the Y-coordinate sometimes is set (50%??). So, what do you think?

pierricgimmig: I think zero is a very special case.  If the box's position has been set once before, but the vertical layout changed, either by panning vertically or by filtering tracks, then we'll still have a bug where we'll align to the wrong Y position. Calling UpdatePrimitives seems like the way to go.

ronaldfw: Sounds good from my side. While UpdatePrimitives is rather expensive, this is only called once on a user action. It's probably not going to be noticeable unless captures get fairly large. Eventually we should fix this, but it's likely totally fine for now.

karupayun: ok!

karupayun: I started to code something to solve the issue. Tomorrow I will push it!

PR #841, opened 2020-07-21
Snap newly added iterator to selected or closest textbox

ronaldfw:
When adding a new iterator, instead of jumping to the start of the timeline, we now snap the iterator to either the selected text box (if any) or to the text box whose start marker is closest to the center of the screen.

### DIFF
--- a/OrbitGl/LiveFunctionsController.cpp
+++ b/OrbitGl/LiveFunctionsController.cpp
@@ -9,6 +9,8 @@
 #include "FunctionUtils.h"
 #include "TimeGraph.h"
 
+namespace {
+
 std::pair<uint64_t, uint64_t> ComputeMinMaxTime(
     const absl::flat_hash_map<uint64_t, const TextBox*>& text_boxes) {
   uint64_t min_time = std::numeric_limits<uint64_t>::max();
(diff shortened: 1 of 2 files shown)


OrbitGl/LiveFunctionsController.cpp
```diff
@@ -20,6 +22,68 @@ std::pair<uint64_t, uint64_t> ComputeMinMaxTime(
...
+  if (!box) {
+    return GCurrentTimeGraph->FindPreviousFunctionCall(
+        absolute_function_address, center);
+  }
+
+  // We have to consider the case where center falls into the interior of the
+  // interval of 'box'. In this case, the closest box can be any of two boxes:
+  // 'box' or the next one. It cannot be any box before 'box' because we are
+  // using the start marker to measure the distance.
+  if (box->GetTimer().m_Start <= center && center <= box->GetTimer().m_End) {
```
karupayun, line 67: You maybe should erase the comparison: "center <= box->GetTimer().m_End"

First, because you don't need it (99.9999% of times will be true), but second because in the strange case that the middle of the screen is really really near of this m_End point and this center-1 switch its relative position from right to left, you are comparing two text_box whose starting point are at the left of the center instead of the correct ones.

ronaldfw: Isn't it actually true 100% of the time? My understanding was that FindNextFunctionCall(..., timestamp) will find the next function call with m_End > timestamp. So, here this condition will be true. 

But you are right, even if m_End were center - 1 (which I don't think it can be), then I should still use this case here.

ronaldfw:
> Hi Ronald, I really like this change! :)
> 
> I just wrote a comment and have a question. Did you consider that the user might sometimes want the iterator to start at the beginning of the capture? I'm not sure if we have to worry about that, but I just want to mention it.

Yes, I will add a button (later PR) to just jump to the start (or end) of the capture, same as the jump to first/last menu entries, but for each iterator.

OrbitGl/LiveFunctionsController.cpp
karupayun: Yes, 100%, you are right.

For one moment, I supposed that could be something between center-1 and center at Integer's world.

PR #844, opened 2020-07-21
Replace capture data with proto

IrinaShkviro:
Move from Cereal to Protobuf when saving/loading captures.

The following classes (structs) were moved to proto: `FunctionStats`, `Function`, `Callstack`, `LinuxAddressInfo` (used AddressInfo instead of this struct), `SampledFunction`, `ThreadSampleData` and `CallstackEvent`.

For the following classes were added "data" proto messages, which are stored inside the classes (as unique_ptr in most cases): `SamplingProfiler`, `StringManager`, `EventBuffer`, `CaptureSerializer` (proto class `CaptureHeader` stores the general information about the capture), `Capture`.

Also, proto message `TimerData` was added: as @beckerhe rightly noticed we use "different" timers for captures, and we don't call Start/Stop on it. As the Timer class that's used for saving/loading captures doesn't use any methods from Timer class, it could be converted to the separate `TimerData` proto message.


dpallotti:
One thing before I start: you should (maybe for next time) include the descriptions of the commits in the commit messages itself, not just in the pull request. This way they are kept in the commit history, and can also be read at the top when reviewing commit by commit.

IrinaShkviro:
> One thing before I start: you should (maybe for next time) include the descriptions of the commits in the commit messages itself, not just in the pull request. This way they are kept in the commit history, and can also be read at the top when reviewing commit by commit.

Updated the commit messages

protos/capture.proto
```diff
@@ -218,3 +218,25 @@ message CaptureHeader {
   uint32 num_timers = 2;
   uint32 version = 3;
 }
+
+message TimerData {
```
dpallotti, line 236: In my opinion, timers should be completely re-thought instead of simply translate them to protobufs. So far we've always used "timers" as a catch all for anything that has a start and an end, with some fields that apply to some type but not others in a very confusing way. This is not a nice design. If we have different types of time slices (e.g., scheduling slices, dynamic instrumentation, gpu driver events, etc.) we should just have different types of protobufs, similarly to what is done for the gRPC capture messages.

protos/capture.proto
```diff
@@ -121,6 +121,16 @@ message CaptureEvent {
   }
 }
 
+message CaptureData {
+  map<uint64, uint64> function_count_map = 1;
+  map<uint64, Callstack> callstacks = 2;
```
dpallotti, line 126: Weren't these already saved in the event buffer? I have a feeling we're carrying over the serialization of a lot of duplicated data.

OrbitGl/CaptureSerializer.cpp
```diff
@@ -49,35 +54,80 @@ ErrorMessageOr<void> CaptureSerializer::Save(const std::string& filename) {
...
+                  google::protobuf::io::CodedOutputStream* output) {
+  uint32_t message_size = message->ByteSizeLong();
+  output->WriteLittleEndian32(message_size);
+  message->SerializeToCodedStream(output);
 }
 
 //-----------------------------------------------------------------------------
-template <class T>
-void CaptureSerializer::SaveImpl(T& archive) {
+void CaptureSerializer::SaveImpl(
```
dpallotti, line 80: In my opinion the final result of this should be to have a single protobuf message consisting of a sequence of small independent messages that we write in a unified manner to a file. Here we are porting over dumping very big and heterogeneous structures like before (just in protobuf form) which like before make backwards compatibility extremely hard to maintain.

protos/capture.proto
```diff
@@ -196,3 +196,9 @@ message SamplingProfilerData {
   map<uint64, uint64> exact_address_to_function_address = 7;
   map<uint64, Uint64List> function_address_to_exact_addresses_set = 8;
 }
+
+message CallstackEvent {
+  uint64 time = 1;
+  uint64 callstack_id = 2;
```
dimitry-, line 216: maybe rename it to callstack_hash?

OrbitGl/CaptureSerializer.h
```diff
@@ -21,12 +22,14 @@ class CaptureSerializer {
...
-
-  ORBIT_SERIALIZABLE;
-
  private:
   template <class T>
   void SaveImpl(T& archive);
+
+  CaptureHeader header;
+
+  static const uint32_t CAPTURE_VERSION = 1;
```
dimitry-, line 30: kCurrentCaptureVersion

OrbitGl/CaptureSerializer.h
```diff
@@ -21,12 +22,14 @@ class CaptureSerializer {
...
+  static const uint32_t MIN_CAPTURE_VERSION = 1;
```
dimitry-, line 31: kMinSupportedCaptureVersion

protos/capture.proto
```diff
@@ -212,3 +212,9 @@ message EventBufferData {
   uint64 max_time = 2;
   uint64 min_time = 3;
 }
+
+message CaptureHeader {
+  string capture_name = 1;
+  uint32 num_timers = 2;
```
dimitry-, line 232: timers_count

protos/capture.proto
```diff
@@ -218,3 +218,25 @@ message CaptureHeader {
...
+
+message TimerData {
+  uint64 start = 1;
+  uint64 end = 2;
+  int32 process_id = 3;
+  int32 thread_id = 4;
+  uint32 depth = 5;
+
+  enum Type {
+    NONE = 0;
```
dimitry-, line 244: kNone here and below: https://google.github.io/styleguide/cppguide.html#Constant_Names

OrbitCore/Capture.cpp
```diff
@@ -44,20 +44,15 @@ std::string Capture::GFunctionFilter;
...
-std::unordered_map<int32_t, std::string> Capture::GThreadNames;
-std::unordered_map<uint64_t, AddressInfo> Capture::GAddressInfos;
-std::unordered_map<uint64_t, std::string> Capture::GAddressToFunctionName;
 Mutex Capture::GCallstackMutex;
 std::unordered_map<uint64_t, std::string> Capture::GZoneNames;
 TextBox* Capture::GSelectedTextBox;
 ThreadID Capture::GSelectedThreadId;
 std::chrono::system_clock::time_point Capture::GCaptureTimePoint;
 
+CaptureData Capture::GData;
```
dimitry-, line 54: Is it possible to move this to DataManager?

OrbitGl/CaptureSerializer.cpp
```diff
@@ -100,19 +150,76 @@ ErrorMessageOr<void> CaptureSerializer::Load(const std::string& filename) {
     return Load(file);
   } catch (std::exception& e) {
```
dimitry-, line 151: does protobuf implementation throw exceptions?

OrbitGl/CaptureSerializer.cpp
```diff
@@ -100,19 +150,76 @@ ErrorMessageOr<void> CaptureSerializer::Load(const std::string& filename) {
...
 ErrorMessageOr<void> CaptureSerializer::Load(std::istream& stream) {
-  // Header
-  cereal::BinaryInputArchive archive(stream);
+  google::protobuf::io::IstreamInputStream input_stream(&stream);
+  google::protobuf::io::CodedInputStream coded_input(&input_stream);
+
+  ReadMessage(&header, &coded_input);
+  if (header.version() < MIN_CAPTURE_VERSION) {
+    std::string error_message =
+        "Capture was taken in a previous Orbit version and incompatible with "
```
dimitry-, line 178: also add %s for versions, we might need a mapping between capture version -> Orbit version.

"This capture was taken with Orbit X.YY which is no longer supported, this version of orbit can load captures starting with Orbit K.NN"

PR #846, opened 2020-07-22
Get updated Y-position from only one box

karupayun:
We were updating all primitives in the screen to only get the updated y-position from only one text box. Now, we get directly from the corresponding thread track.

### DIFF
--- a/OrbitGl/ThreadTrack.h
+++ b/OrbitGl/ThreadTrack.h
@@ -58,6 +58,7 @@ class ThreadTrack : public Track {
 
   int32_t GetThreadId() const { return thread_id_; }
   bool IsCollapsable() const override { return depth_ > 1; }
+  float GetYFromDepth(uint32_t depth);
 
  protected:
   void UpdateDepth(uint32_t depth) {
--- a/OrbitGl/TimeGraph.cpp
+++ b/OrbitGl/TimeGraph.cpp
@@ -242,18 +242,19 @@ void TimeGraph::HorizontallyMoveIntoView(const TextBox* text_box,
 
 void TimeGraph::VerticallyMoveIntoView(const TextBox* text_box) {
   CHECK(text_box != nullptr);
-  // TODO: Sometimes the Y-coordinate is not set. We only need to update the
-  // Y-coordinate from this text box
-  UpdatePrimitives();
+  const Timer& timer = text_box->GetTimer();
+  auto thread_track = GetOrCreateThreadTrack(timer.m_TID);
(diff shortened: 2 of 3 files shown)


OrbitGl/TimeGraph.cpp
```diff
@@ -242,18 +242,20 @@ void TimeGraph::HorizontallyMoveIntoView(const TextBox* text_box,
 
 void TimeGraph::VerticallyMoveIntoView(const TextBox* text_box) {
   CHECK(text_box != nullptr);
-  // TODO: Sometimes the Y-coordinate is not set. We only need to update the
-  // Y-coordinate from this text box
-  UpdatePrimitives();
+  const Timer& timer = text_box->GetTimer();
+  auto thread_track = GetOrCreateThreadTrack(timer.m_TID);
+  // TODO: Could we need to update the depth of the timer?
```
ronaldfw, line 247: I don't think so, but not 100% sure.

pierricgimmig: We shouldn't change the m_Depth member of Timer once it's set, no.

pierricgimmig: The depth represents an index into a stack of instrumented functions, it should always remain the same.

OrbitGl/ThreadTrack.h
```diff
@@ -58,6 +58,7 @@ class ThreadTrack : public Track {
 
   int32_t GetThreadId() const { return thread_id_; }
   bool IsCollapsable() const override { return depth_ > 1; }
+  virtual float GetYFromDepth(uint32_t depth);
```
pierricgimmig, line 61: Should this be overridden in SchedulerTrack? If not, does it need to be virtual?

OrbitGl/TimeGraph.cpp
karupayun: Yes, I know. And my doubt was that if it set/updated in UpdatePrimitives, maybe has never set at this moment.

OrbitGl/ThreadTrack.h
karupayun: No! :)

OrbitGl/TimeGraph.cpp
ronaldfw: It is actually set already on each timer, so we should be fine. 

I vaguely recalled there was an UpdateDepth() method called in UpdatePrimitives, but that does not update the depth of the timer, but the (current) depth of the track .

PR #847, opened 2020-07-22
Erase double vertical alignment

karupayun:
No need to vertical move after select because Select already make that.


ronaldfw:
Actually, I think this change is unnecessary after #849, where I change how the movement happens when iterating.

karupayun:
> Actually, I think this change is unnecessary after #849, where I change how the movement happens when iterating.

Agree!

PR #849, opened 2020-07-23
Do not zoom when moving the screen due to iterator movements

ronaldfw:
This should be less confusing for users. In particular, this allows users to zoom out and see an overview of the iterator range.

The screen is still moved to make sure that current iterators are in view.

### DIFF
--- a/OrbitGl/TimeGraph.cpp
+++ b/OrbitGl/TimeGraph.cpp
@@ -214,7 +214,8 @@ void TimeGraph::PanTime(int a_InitialX, int a_CurrentX, int a_Width,
   NeedsUpdate();
 }
 
-void TimeGraph::HorizontallyMoveIntoView(VisibilityType vis_type, TickType min, TickType max, double distance) {
+void TimeGraph::HorizontallyMoveIntoView(VisibilityType vis_type, TickType min,
+                                         TickType max, double distance) {
   if (IsVisible(vis_type, min, max)) {
     return;
   }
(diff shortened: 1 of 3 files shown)


OrbitGl/TimeGraph.cpp
```diff
@@ -214,19 +214,14 @@ void TimeGraph::PanTime(int a_InitialX, int a_CurrentX, int a_Width,
...
 
-// Move the view horizontally if the text box isn't visible. The text box
-// center's position will be at 'distance' to the related border (in [0,1]
-// scale).
-void TimeGraph::HorizontallyMoveIntoView(const TextBox* text_box,
-                                         double distance) {
-  const Timer& timer = text_box->GetTimer();
-
-  if (IsVisible(timer)) {
+void TimeGraph::HorizontallyMoveIntoView(VisibilityType vis_type, TickType min,
```
karupayun, line 217: As I commented, I would like to add some check:

if (vis_type == kFullyVisible && CurrentTimeWindowsUs < end - start){
   Zoom (min, max);
}

or in the end (after setting min-max):
if (!IsVisible(vis_type, min, max)){
   Zoom (min, max);
}

OrbitGl/TimeGraph.h
```diff
@@ -62,7 +62,12 @@ class TimeGraph {
   void SetMinMax(double a_MinTimeUs, double a_MaxTimeUs);
   void PanTime(int a_InitialX, int a_CurrentX, int a_Width,
                double a_InitialTime);
-  void HorizontallyMoveIntoView(const TextBox* text_box, double distance = 0.3);
+  enum VisibilityType {
```
karupayun, line 65: I'm not the expert here. It is ok using enum instead enum class here?

ronaldfw: Fixed.

OrbitGl/TimeGraph.cpp
ronaldfw: Yes, I think you may be right. I'll make the change as you suggest and will experiment a little with it.

ronaldfw:
Thanks a lot for the review Pablo!

PR #852, opened 2020-07-23
Add missing step in CLion setup

dimitry-:


DEVELOPMENT.md
```diff
@@ -169,6 +169,8 @@ between CLion build configurations and conan profiles. Check out
 [this blog post](https://blog.jetbrains.com/clion/2019/05/getting-started-with-the-conan-clion-plugin)
 on how to do it.
 
+Add ```-DCMAKE_CXX_FLAGS=-fsized-deallocation``` to Settings -> Build, Execution Deployment -> CMake -> CMake options.
```
dpallotti, line 172:
```suggestion
Add ```-DCMAKE_CXX_FLAGS=-fsized-deallocation``` to Settings -> Build, Execution, Deployment -> CMake -> CMake options.
```

DEVELOPMENT.md
```diff
@@ -169,6 +169,8 @@ between CLion build configurations and conan profiles. Check out
...
+
```
dpallotti, line 173: Should we add a brief comment with the reason for this? @beckerhe You remember better than me. I think this was a clang problem but I don't remember the details.

DEVELOPMENT.md
dimitry-: done.

DEVELOPMENT.md
beckerhe: So the details: C++14 introduced an overload of `operator delete` taking a pointer and a size instead of just a pointer. This allows more efficient implementations of delete in custom allocators. But it's only available in clang behind the mentioned flag (because it would have been an ABI-breaking change.)
So maybe a sentence like this: This flag is needed because Orbit's codebase makes use of C++14's sized-deallocation feature which is not enabled by default on the clang compiler.

dimitry-: done.

PR #854, opened 2020-07-23
More descriptive tooltips for the capture window

reichlfl:
Added tooltips for certain elements in the capture window. I did not touch the behavior of how elements of type PickingID::Pickable are stored - in short, their user_data cannot be accessed in the hover or pick callbacks, and tooltips won't work for them (since the tooltip is in the user_data).

I added a seperate GetTooltip() method in _Pickable_ that provides an easy way to add tooltips to those as well. In general, this only refers to the "larger elements" of the UI: Tracks (but not their content), Scrollbars, Tabs... I added a Tooltip for the event tracks with a hint that the samples can be selected as this was mentioned in a UX interview.

I have some more work open on the tooltips for samples, but didn't want to put too much into a single request.

### DIFF
--- a/OrbitGl/Batcher.cpp
+++ b/OrbitGl/Batcher.cpp
@@ -94,7 +94,7 @@ void Batcher::AddTriangle(Vec3 v0, Vec3 v1, Vec3 v2, Color color,
 }
 
 std::shared_ptr<PickingUserData> Batcher::GetUserData(PickingID a_ID) {
-  std::shared_ptr<PickingUserData>* data;
+  std::shared_ptr<PickingUserData>* data = nullptr;
   switch (a_ID.m_Type) {
     case PickingID::BOX:
       data = box_buffer_.m_UserData.SlowAt(a_ID.m_Id);
--- a/OrbitGl/Batcher.cpp
+++ b/OrbitGl/Batcher.cpp
@@ -9,112 +9,111 @@
 
 void Batcher::AddLine(const Line& line, const Color* colors,
                       PickingID::Type picking_type,
-                      std::shared_ptr<PickingUserData> user_data) {
+                      std::unique_ptr<PickingUserData> user_data) {
   Color picking_color = PickingID::GetColor(
       picking_type, line_buffer_.m_Lines.size(), batcher_id_);
   line_buffer_.m_Lines.push_back(line);
   line_buffer_.m_Colors.push_back(colors, 2);
   line_buffer_.m_PickingColors.push_back_n(picking_color, 2);
-  line_buffer_.m_UserData.push_back(user_data);
+  line_buffer_.m_UserData.push_back(std::move(user_data));
 }
 
 void Batcher::AddLine(const Line& line, Color color,
                       PickingID::Type picking_type,
-                      std::shared_ptr<PickingUserData> user_data) {
+                      std::unique_ptr<PickingUserData> user_data) {
   Color colors[2];
   Fill(colors, color);
-  AddLine(line, colors, picking_type, user_data);
+  AddLine(line, colors, picking_type, std::move(user_data));
 }
 
 void Batcher::AddLine(Vec2 from, Vec2 to, float z, Color color,
                       PickingID::Type picking_type,
-                      std::shared_ptr<PickingUserData> user_data) {
+                      std::unique_ptr<PickingUserData> user_data) {
   Line line;
   Color colors[2];
   Fill(colors, color);
   line.m_Beg = Vec3(from[0], from[1], z);
   line.m_End = Vec3(to[0], to[1], z);
-  AddLine(line, colors, picking_type, user_data);
+  AddLine(line, colors, picking_type, std::move(user_data));
 }
 
 void Batcher::AddVerticalLine(Vec2 pos, float size, float z, Color color,
                               PickingID::Type picking_type,
-                              std::shared_ptr<PickingUserData> user_data) {
+                              std::unique_ptr<PickingUserData> user_data) {
   Line line;
   Color colors[2];
   Fill(colors, color);
   line.m_Beg = Vec3(pos[0], pos[1], z);
   line.m_End = Vec3(pos[0], pos[1] + size, z);
-  AddLine(line, colors, picking_type, user_data);
+  AddLine(line, colors, picking_type, std::move(user_data));
 }
 
 void Batcher::AddBox(const Box& a_Box, const Color* colors,
                      PickingID::Type picking_type,
-                     std::shared_ptr<PickingUserData> user_data) {
+                     std::unique_ptr<PickingUserData> user_data) {
   Color picking_color = PickingID::GetColor(
       picking_type, box_buffer_.m_Boxes.size(), batcher_id_);
   box_buffer_.m_Boxes.push_back(a_Box);
   box_buffer_.m_Colors.push_back(colors, 4);
   box_buffer_.m_PickingColors.push_back_n(picking_color, 4);
-  box_buffer_.m_UserData.push_back(user_data);
+  box_buffer_.m_UserData.push_back(std::move(user_data));
 }
 
 void Batcher::AddBox(const Box& a_Box, Color color,
                      PickingID::Type picking_type,
-                     std::shared_ptr<PickingUserData> user_data) {
+                     std::unique_ptr<PickingUserData> user_data) {
   Color colors[4];
   Fill(colors, color);
-  AddBox(a_Box, colors, picking_type, user_data);
+  AddBox(a_Box, colors, picking_type, std::move(user_data));
 }
 
 void Batcher::AddShadedBox(Vec2 pos, Vec2 size, float z, Color color,
                            PickingID::Type picking_type,
-                           std::shared_ptr<PickingUserData> user_data) {
+                           std::unique_ptr<PickingUserData> user_data) {
   Color colors[4];
   GetBoxGradientColors(color, colors);
   Box box(pos, size, z);
-  AddBox(box, colors, picking_type, user_data);
+  AddBox(box, colors, picking_type, std::move(user_data));
 }
 
 void Batcher::AddTriangle(const Triangle& triangle, Color color,
                           PickingID::Type picking_type,
-                          std::shared_ptr<PickingUserData> user_data) {
+                          std::unique_ptr<PickingUserData> user_data) {
   Color picking_color = PickingID::GetColor(
       picking_type, triangle_buffer_.triangles_.size(), batcher_id_);
   triangle_buffer_.triangles_.push_back(triangle);
   triangle_buffer_.colors_.push_back_n(color, 3);
   triangle_buffer_.picking_colors_.push_back_n(picking_color, 3);
-  triangle_buffer_.user_data_.push_back(user_data);
+  triangle_buffer_.user_data_.push_back(std::move(user_data));
 }
 
 void Batcher::AddTriangle(Vec3 v0, Vec3 v1, Vec3 v2, Color color,
                           PickingID::Type picking_type,
-                          std::shared_ptr<PickingUserData> user_data) {
-  AddTriangle(Triangle(v0, v1, v2), color, picking_type, user_data);
+                          std::unique_ptr<PickingUserData> user_data) {
+  AddTriangle(Triangle(v0, v1, v2), color, picking_type, std::move(user_data));
 }
 
-std::shared_ptr<PickingUserData> Batcher::GetUserData(PickingID a_ID) {
-  std::shared_ptr<PickingUserData>* data = nullptr;
+PickingUserData* Batcher::GetUserData(PickingID a_ID) {
+  CHECK(a_ID.m_Id >= 0);
+
   switch (a_ID.m_Type) {
     case PickingID::BOX:
-      data = box_buffer_.m_UserData.SlowAt(a_ID.m_Id);
-      break;
+      CHECK(a_ID.m_Id < box_buffer_.m_UserData.size());
+      return box_buffer_.m_UserData[a_ID.m_Id].get();
     case PickingID::LINE:
-      data = line_buffer_.m_UserData.SlowAt(a_ID.m_Id);
-      break;
+      CHECK(a_ID.m_Id < line_buffer_.m_UserData.size());
+      return line_buffer_.m_UserData[a_ID.m_Id].get();
     case PickingID::TRIANGLE:
-      data = triangle_buffer_.user_data_.SlowAt(a_ID.m_Id);
-      break;
+      CHECK(a_ID.m_Id < triangle_buffer_.user_data_.size());
+      return triangle_buffer_.user_data_[a_ID.m_Id].get();
   }
 
-  if (data) {
-    return *data;
-  }
   return nullptr;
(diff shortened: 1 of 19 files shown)


OrbitGl/CaptureWindow.cpp
```diff
@@ -181,40 +181,13 @@ void CaptureWindow::Pick(int a_X, int a_Y) {
 void CaptureWindow::Pick(PickingID a_PickingID, int a_X, int a_Y) {
   uint32_t type = a_PickingID.m_Type;
   uint32_t id = a_PickingID.m_Id;
```
karupayun, line 183: Linux's build complains when there are unused variables.

You can add: 
UNUSED(id);
if you are going to use in the future

OrbitGl/Batcher.cpp
```diff
@@ -9,7 +9,7 @@
 
 void Batcher::AddLine(const Line& line, const Color* colors,
                       PickingID::Type picking_type,
-                      PickingUserData user_data) {
+                      std::shared_ptr<PickingUserData> user_data) {
```
dimitry-, line 12: I would say someone needs to assume ownership over PickingUserData here. Maybe have PickingUserData passed as r-value so you can std::move() it to line_buffer.m_UserData, in which case the onwership will belong to m_UserData

OrbitGl/Batcher.h
```diff
@@ -64,32 +78,40 @@ class Batcher {
...
-               PickingID::Type picking_type, void* user_data = nullptr);
+               PickingID::Type picking_type, 
+               std::shared_ptr<PickingUserData> user_data = nullptr);
   void AddLine(const Line& line, Color color, PickingID::Type picking_type,
-               void* user_data = nullptr);
+               std::shared_ptr<PickingUserData> user_data = nullptr);
   void AddLine(Vec2 from, Vec2 to, float z, Color color,
-               PickingID::Type picking_type, void* user_data = nullptr);
+               PickingID::Type picking_type,
+               std::shared_ptr<PickingUserData> user_data = nullptr);
```
dimitry-, line 87: you might want to create 2 function - one for the case with empty userdata.

OrbitGl/GpuTrack.cpp
```diff
@@ -191,20 +192,22 @@ void GpuTrack::UpdatePrimitives(uint64_t min_tick, uint64_t max_tick) {
         if (is_collapsed) {
           std::string gpu_stage =
               string_manager_->Get(timer.m_UserData[0]).value_or("");
-          constexpr const char* kHwExecutionString = "hw execution";
           if (gpu_stage != kHwExecutionString) {
             continue;
           }
         }
 
+        auto userData = std::make_shared<PickingUserData>(
```
dimitry-, line 200: user_data

OrbitGl/SchedulerTrack.cpp
```diff
@@ -115,11 +120,15 @@ void SchedulerTrack::UpdatePrimitives(uint64_t min_tick, uint64_t max_tick) {
                                     is_same_tid_as_selected,
                                     is_same_pid_as_target, is_inactive);
 
+        auto userData = std::make_shared<PickingUserData>(
```
dimitry-, line 123: user_data

OrbitGl/SchedulerTrack.cpp
```diff
@@ -133,3 +142,20 @@ void SchedulerTrack::UpdatePrimitives(uint64_t min_tick, uint64_t max_tick) {
     }
   }
 }
+
+std::string SchedulerTrack::GetBoxTooltip(PickingID id) const {
+  TextBox* textBox = time_graph_->GetBatcher().GetTextBox(id);
```
dimitry-, line 147: text_box

OrbitGl/ThreadTrack.cpp
```diff
@@ -382,3 +389,30 @@ void ThreadTrack::SetEventTrackColor(Color color) {
 bool ThreadTrack::IsEmpty() const {
   return (GetNumTimers() == 0) && event_track_->IsEmpty();
 }
+
+
+//-----------------------------------------------------------------------------
+std::string ThreadTrack::GetBoxTooltip(PickingID id) const {
+  TextBox* textBox = time_graph_->GetBatcher().GetTextBox(id);
+  if (textBox) {
```
dimitry-, line 397: if (!text_box) {
  return "";
}

...

OrbitGl/ThreadTrack.cpp
```diff
@@ -382,3 +389,30 @@ void ThreadTrack::SetEventTrackColor(Color color) {
...
+    if (textBox->GetTimer().m_Type != Timer::CORE_ACTIVITY) {
```
dimitry-, line 398: same here - if (!...) {
return ...;
}

....

OrbitGl/ThreadTrack.cpp
```diff
@@ -382,3 +389,30 @@ void ThreadTrack::SetEventTrackColor(Color color) {
...
+      Function* func =
+          Capture::GSelectedFunctionsMap[textBox->GetTimer().m_FunctionAddress];
+      if (func) {
```
dimitry-, line 401: and here.

OrbitGl/Batcher.cpp
dimitry-: looking at the rest of the code it seems to be the case that once this is passed to AddLine function it is no longer user by the caller. I think you can do PickingUserData&& user_data here and use std::move() when passing it to this function.

OrbitGl/CaptureWindow.cpp
reichlfl: Is there a way to warn about this when compiling for windows?

OrbitGl/Batcher.h
reichlfl: Can you elaborate why you'd prefer that?
The AddLine e.g. has 3 overloads that all would need to be duplicated, I find a lot of overloads to be more confusing than defaults in those cases.

OrbitGl/Batcher.cpp
reichlfl: I actually wanted to use a unique_ptr here, but this would have required changes to the BlockChain (it always copies on "add"), and I did not want to introduce to many changes since it's quite fundamental. So I stuck with the shared_ptr, also because there may be situations where not taking ownership may actually arise... right now it's not used, correct.
Isn't it a bit counter-intuitive to use a shared_ptr, but then forcing to move it?

OrbitGl/Batcher.h
dimitry-: because there is no nulls for r-values.

OrbitGl/Batcher.cpp
dimitry-: well, the idea here is to not use shared_ptr - we do not really need reference counting here do we?

OrbitQt/orbitglwidget.cpp
```diff
@@ -296,10 +296,12 @@ void OrbitGLWidget::mouseMoveEvent(QMouseEvent* event) {
 }
 
 void OrbitGLWidget::enterEvent(QEvent* event) {
```
dimitry-, line 298: nit: you can just remove event from parameters:
enterEvent(QEvent*) { ...

OrbitQt/orbitglwidget.cpp
```diff
@@ -296,10 +296,12 @@ void OrbitGLWidget::mouseMoveEvent(QMouseEvent* event) {
...
+  UNUSED(event);
```
dimitry-, line 299: in which case you won't need this UNUSED here.

ronaldfw:
Just a small comment: Please update the PR title to something descriptive (maybe "Add proper tooltips in capture window"?). Thank you!

OrbitGl/ThreadTrack.cpp
```diff
@@ -394,25 +394,23 @@ bool ThreadTrack::IsEmpty() const {
...
-      }
-    }
+  if (!text_box || text_box->GetTimer().m_Type == Timer::CORE_ACTIVITY) {
+    return "";
   }
 
-  return "";
+  Function* func =
+      Capture::GSelectedFunctionsMap[text_box->GetTimer().m_FunctionAddress];
+  if (func) {
```
dimitry-, line 403: if !func and drop else

OrbitGl/Batcher.h
reichlfl: Unchanged, because now we're using std::unique_ptr<>

OrbitGl/Batcher.cpp
reichlfl: Changed to use std::unique_ptr<>

OrbitQt/orbitglwidget.cpp
dimitry-: have you decided to go with UNUSED?

OrbitGl/EventTrack.cpp
```diff
@@ -16,6 +16,10 @@ EventTrack::EventTrack(TimeGraph* a_TimeGraph) : Track(a_TimeGraph) {
   m_Color = Color(0, 255, 0, 255);
 }
 
+std::string EventTrack::GetTooltip() const {
+  return "Left-click and drag to select samples.";
```
ronaldfw, line 20: Nit: I think that it's a bit cleaner to not have the "." in the tooltip. (Single data point: Chrome does not have "." in their tooltips at the end of the sentence.)

OrbitGl/GlPanel.h
```diff
@@ -64,4 +67,6 @@ class GlPanel {
   int m_MainWindowWidth;
   int m_MainWindowHeight;
   bool m_NeedsRedraw;
+
+  bool m_IsMouseOver;
```
ronaldfw, line 71: Nit: Newly added members should follow the style is_mouse_over_ (yes, it's inconsistent with what's there, but we are trying to move to Google C++ style over time).

OrbitGl/GpuTrack.cpp
```diff
@@ -238,6 +241,10 @@ void GpuTrack::OnTimer(const Timer& timer) {
   if (timer.m_End > max_time_) max_time_ = timer.m_End;
 }
 
+std::string GpuTrack::GetTooltip() const {
+  return "Shows scheduling and execution times for selected GPU calls.";
```
ronaldfw, line 245: Same here, suggest to get rid of "."

I would change "calls" to "job submissions".

OrbitGl/GpuTrack.cpp
```diff
@@ -347,3 +354,53 @@ std::vector<std::shared_ptr<TimerChain>> GpuTrack::GetAllChains() {
...
+    return GetHwExecutionTooltip(text_box->GetTimer());
+  }
+
+  return "";
+}
+
+std::string GpuTrack::GetSwQueueTooltip(const Timer& timer) const { 
+  return absl::StrFormat(
+    "<b>Software Queue</b><br/>"
+    "<i>Time between amdgpu_cs_ioctl and amdgpu_sched_run_job.</i>"
```
ronaldfw, line 381: Can be done later, but we may want to add an explanation for amdgpu_cs_ioctl and amdgpu_sched_run_job. 

amdgpu_cs_ioctl = job submission from user space
amdgpu_sched_run_job = job scheduled by the driver

OrbitGl/GpuTrack.cpp
```diff
@@ -347,3 +354,53 @@ std::vector<std::shared_ptr<TimerChain>> GpuTrack::GetAllChains() {
...
+    "<br/>"
+    "<br/>"
+    "<b>Submitter thread:</b> %s [%d]<br/>"
```
ronaldfw, line 384: Nit: "Submitted from thread:" ..(?)

OrbitGl/SchedulerTrack.cpp
```diff
@@ -44,6 +45,10 @@ float SchedulerTrack::GetYFromDepth(float track_y, uint32_t depth,
          num_gaps * gap_size;
 }
 
+std::string SchedulerTrack::GetTooltip() const { 
+  return "Shows scheduling information for CPU cores.";
```
ronaldfw, line 49: Nit: Get rid of ".".

OrbitGl/ThreadTrack.cpp
```diff
@@ -262,6 +265,10 @@ void ThreadTrack::OnTimer(const Timer& timer) {
   if (timer.m_End > max_time_) max_time_ = timer.m_End;
 }
 
+std::string ThreadTrack::GetTooltip() const {
+  return "Shows collected samples and instrumented function timings (if existing).";
```
ronaldfw, line 269: Nit: Get rid of ".".

ronaldfw, line 269: I would probably get rid of "if existing", it is implied anyway. WDYT?

ronaldfw: Maybe an idea: If you write "...and timings from dynamically instrumented functions." it is a user hint that one can dynamically instrument functions, which can aid when learning and discovering the tool.

OrbitQt/orbitglwidget.cpp
ronaldfw: We have kind of agreed at some point that we don't use UNUSED anymore (and just omit the name).

OrbitGl/GpuTrack.cpp
ronaldfw: Note that I had used "submitter" earlier because it was the shortest I could think of (but already found it sounding weird).

OrbitQt/orbitglwidget.cpp
reichlfl: I found the UNUSED more verbose, but I've changed it now and removed the parameter name for consistency

OrbitGl/EventTrack.cpp
reichlfl: I was undecided about this, so this tipped the scales and I removed the "." everywhere ;-)

OrbitGl/GpuTrack.cpp
reichlfl: I think that makes sense, I added a bit of a shortened version ("job submitted" and "job scheduled") to keep the tooltip a bit shorter, wdyt?

OrbitGl/ThreadTrack.cpp
reichlfl: I initially added the "if existing" because it might have been confusing if the user does not know about dynamic instrumentation and would be looking for the timings announced in the tooltip, but I think yours is more elegant and also implies this. Changed.

ronaldfw:
Just tried this out and it's really nice. 

Two minor things we should fix: We can now remove some of the text on the GPU timeslices to make them look a bit cleaner (for example "submitter: tid"). I also think all GPU timeslice tooltips should show the thread from where the job was submitted (o/w as a user, I always have to move over to the initial software queue timeslice). WDYT?

reichlfl:
> Two minor things we should fix: We can now remove some of the text on the GPU timeslices to make them look a bit cleaner (for example "submitter: tid"). I also think all GPU timeslice tooltips should show the thread from where the job was submitted (o/w as a user, I always have to move over to the initial software queue timeslice). WDYT?

Makes sense to me, I'll take a look.

PR #856, opened 2020-07-24
Display timings between iterators

ronaldfw:
This PR adds displaying timings between all iterators and also adds a visual indication which time is shown. 

Screenshot what this looks like:
ORBIT_iterator_timings

### DIFF
--- a/OrbitGl/TimeGraph.cpp
+++ b/OrbitGl/TimeGraph.cpp
@@ -590,8 +590,8 @@ void TimeGraph::Draw(GlCanvas* canvas, bool a_Picking) {
 
 namespace {
 
-std::string GetLabelBetweenIterators(const TextBox* box_a,
-                                     const TextBox* box_b) {
+std::pair<std::string, std::string> GetLabelBetweenIterators(
+    const TextBox* box_a, const TextBox* box_b) {
   double micros = MicroSecondsFromTicks(box_a->GetTimer().m_Start,
                                         box_b->GetTimer().m_Start);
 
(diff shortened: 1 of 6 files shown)


OrbitGl/TimeGraph.cpp
```diff
@@ -592,15 +592,73 @@ void TimeGraph::Draw(GlCanvas* canvas, bool a_Picking) {
...
+  std::string function_to =
+      Capture::GAddressToFunctionName[box_b->GetTimer().m_FunctionAddress];
+  return std::make_pair(absl::StrFormat("%s to %s", function_from, function_to),
+                        time);
+}
+
+Color GetIteratorBoxColor(uint64_t index) {
+  constexpr uint64_t kNumColors = 4;
+  Color colors[kNumColors] = {Color(73, 208, 242, 80), Color(0, 90, 112, 80),
+                              Color(0, 192, 240, 80), Color(40, 103, 122, 80)};
```
karupayun, line 616: Shall we give names to these colours (kLightBlue1, kDarkBlue1 (or kGrey1), kLightBlue2, ...)?

OrbitGl/TimeGraph.cpp
```diff
@@ -610,44 +668,59 @@ void TimeGraph::DrawOverlay(GlCanvas* canvas, bool picking) {
...
+
+    auto& labels =
+        GetLabelBetweenIterators(boxes[k - 1].second, boxes[k].second);
 
-    float z = GlCanvas::Z_VALUE_OVERLAY;
-    Color color = GetThreadColor(timer.m_TID);
+    float text_y_offset = world_height / 2.f /
+                          static_cast<float>(overlay_current_textboxes_.size());
+    float text_y =
+        pos[1] + (world_height / 2.f) - static_cast<float>(k) * text_y_offset;
```
karupayun, line 705: I'm having issues here. Maybe is my linux workstation. I think that we have to add the margin to this world_height. It isn't divided in equal parts. In my linux view I had 30 of margin. So the real world_height (from WorldTopLeftY to the real bottom is 30 less).

karupayun: I'm pretty sure that this downMargin of 30 is exactly 2 * m_SliderWidth.

OrbitGl/TimeGraph.cpp
ronaldfw: That's a good idea, done.

ronaldfw: Also reduced the number of colors to just two, because there wasn't really a large difference between 0, 2 and 1, 3.

OrbitGl/TimeGraph.cpp
ronaldfw: I had a look and the definitions of the margins, etc. are not 100% clear to me. 

What is clear: The bottom shows a slider (TimeGraphLayout::GetSliderWidth) and there is also the timer bar with timing values. For the latter I added a new TimerGraphLayout value that specifies the height of the timer bar (previously GetSliderWidth() was used for that). So far, I think this is ok. (This also matches what you said about the down margin being 2 * m_SliderWidth.)

But just using these two values is not quite enough, the text at the bottom iterator still overlaps a little with the timer bar. 

When I add TimeGraphLayout::GetVerticalMargin(), it seems to be better, but I don't fully understand the meaning of this value. In one place it is used for the width of the right margin (see CaptureWindow::DrawScreenSpace).

ronaldfw: I'm also wondering if we should remove a margin at the top and possibly the scheduler track offset. WDYT?

ronaldfw: Main issue should be fixed now. 

@karupayun Do you have an opinion regarding how to handle to top margin, if at all?

karupayun: I think that m_Layout.GetSchedulerTrackOffset() is the topMargin.

karupayun: I prefer to have the total time (n > 2) at the 1/3 line (instead of 1/2) and divided the rest (2/3) for the iterators. But, having it at the 1/2 line is ok if you prefer that.
And, counting both top_margin and down_margin. top_margin is not too interesting now, but if you have to zoom could be.

ronaldfw: Yes, agree. Should I take it into account in the computation? That would push the "center" of where I draw "Total" down a bit. It doesn't really matter and the window can't be resized to be small enough that you would only see the scheduler track. And I'm not even sure how I want things to behave then.

ronaldfw: I'll merge for now, but let's chat if you think this still requires a change.

PR #858, opened 2020-07-24
Remove unused fields from Callstack and FunctionStats

IrinaShkviro:
- Removed m_Address from FunctionStats because it's not used

- m_Depth in CallStack is set up as m_Data.size() and never changed (except the test), so can be removed

- m_ThreadId is used only once: it's set up from CallstackEvent and used to set up the field in another CallstackEvent. We can avoid it by passing CallstackEvent instead of Callstack.

- Replaced m_Hash with Hash (which basically returns m_Hash, but also initialize it if it wasn't done before)

### DIFF
--- a/OrbitCore/SamplingProfiler.cpp
+++ b/OrbitCore/SamplingProfiler.cpp
@@ -58,17 +58,15 @@ std::multimap<int, CallstackID> SamplingProfiler::GetCallstacksFromAddress(
 }
 
 //-----------------------------------------------------------------------------
-void SamplingProfiler::AddCallStack(CallStack& a_CallStack) {
-  CallstackID hash = a_CallStack.Hash();
+void SamplingProfiler::AddCallStack(CallstackEvent& callstack_event) {
+  CallstackID hash = callstack_event.m_Id;
   if (!HasCallStack(hash)) {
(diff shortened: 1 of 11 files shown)


OrbitCore/SamplingProfiler.cpp
```diff
@@ -58,17 +58,15 @@ std::multimap<int, CallstackID> SamplingProfiler::GetCallstacksFromAddress(
 }
 
 //-----------------------------------------------------------------------------
-void SamplingProfiler::AddCallStack(CallStack& a_CallStack) {
-  CallstackID hash = a_CallStack.Hash();
+void SamplingProfiler::AddCallStack(CallstackEvent& callstack_event) {
```
dpallotti, line 61: This seems now very similar to `SamplingProfiler::AddHashedCallstack`, with the only difference being the behavior in case of `!HasCallStack(...)`. Are they both necessary or can one of the two be removed?

IrinaShkviro: I probably want to refactor SamplingProfiler in one of the upcoming PRs, will keep it in mind, thanks!

IrinaShkviro: Done in #865

PR #859, opened 2020-07-24
More descriptive names for the GPU tracks

ronaldfw:
There are three names of queues that we can encounter (on AMD): "gfx", "sdmaX", and "comp_X.Y.Z". After this change, these names are used for more descriptive track names ("Graphics queue", "Transfer queue", "Compute queue"). The original name passed from the tracepoint is still used in the name to disambiguate if, for instance, multiple compute queues are used.

### DIFF
--- a/OrbitGl/GpuTrack.cpp
+++ b/OrbitGl/GpuTrack.cpp
@@ -12,9 +12,28 @@
 #include "absl/flags/flag.h"
 #include "absl/strings/str_format.h"
 
+<<<<<<< HEAD
 constexpr const char* kSwQueueString = "sw queue";
 constexpr const char* kHwQueueString = "hw queue";
 constexpr const char* kHwExecutionString = "hw execution";
+=======
+std::string MapGpuTimelineToTrackLabel(std::string_view timeline) {
+  std::string label;
+  if (timeline.rfind("gfx", 0) == 0) {
+    return absl::StrFormat("Graphics queue (%s)", timeline);
+  } else if (timeline.rfind("sdma", 0) == 0) {
+    return absl::StrFormat("Transfer queue (%s)", timeline);
+  } else if (timeline.rfind("comp", 0) == 0) {
+    return absl::StrFormat("Compute queue (%s)", timeline);
+  } else {
+    // On AMD, this should not happen and we don't support tracepoints for
+    // other GPUs (at the moment). We return the timeline to make sure we
+    // at least display something. When we add support for other GPU
+    // tracepoints, this needs to be changed.
+    return std::string(timeline);
+  }
+}
+>>>>>>> Define GPU timeline to track label conversion in free function; add some clarifying comments.
 
 //-----------------------------------------------------------------------------
 GpuTrack::GpuTrack(TimeGraph* time_graph,
--- a/OrbitGl/GpuTrack.cpp
+++ b/OrbitGl/GpuTrack.cpp
@@ -12,11 +12,12 @@
 #include "absl/flags/flag.h"
 #include "absl/strings/str_format.h"
 
-<<<<<<< HEAD
 constexpr const char* kSwQueueString = "sw queue";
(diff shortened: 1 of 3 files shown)


OrbitGl/GpuTrack.cpp
```diff
@@ -351,3 +351,20 @@ std::vector<std::shared_ptr<TimerChain>> GpuTrack::GetAllChains() {
   }
   return chains;
 }
+
+//-----------------------------------------------------------------------------
+void GpuTrack::SetTimeline(std::string_view timeline) {
```
pierricgimmig, line 356: nit. I find "timeline" to be confusing to describe a label or a name.

beckerhe: I kind of agree but it's also my first time looking at that code. And I don't have a good suggestion for a name, so I personally would "fix" that problem by avoiding that function all the way. I would define a free function that contains the absl::StrFormat-logic from below and maps from a gpu-track name to a gpu-track label. (Please feel free to ignore that.)

OrbitGl/TimeGraph.cpp
```diff
@@ -708,6 +706,7 @@ std::shared_ptr<GpuTrack> TimeGraph::GetOrCreateGpuTrack(
   std::shared_ptr<GpuTrack> track = gpu_tracks_[timeline_hash];
   if (track == nullptr) {
     track = std::make_shared<GpuTrack>(this, string_manager_, timeline_hash);
+    track->SetTimeline(string_manager_->Get(timeline_hash).value_or(""));
```
beckerhe, line 709: When looking at other usages of `track->SetName` and `track->setLabel` in the code base, it seems like we assign the raw value to the name and the processed human-readable value to the label. Here both get the processed value. So I'm wondering, do we actually need the distinction?

ronaldfw: Yes, you are right. I will use the "raw" name (which is the timeline name) for SetName and the descriptive name for SetLabel. I wasn't sure where that made a difference, but it's actually used when filtering tracks. So, the distinction is actually important. Thanks.

OrbitGl/GpuTrack.cpp
ronaldfw: Yes, it is. "Timeline" is the Linux kernel name for this, for example, used in the tracepoint format. 

I will follow @beckerhe 's suggestion to map the timeline to the display name.

ronaldfw: Done.

OrbitGl/TimeGraph.cpp
ronaldfw: Except that we don't seem to filter GPU tracks, but that's something we might want to change?

ronaldfw: Done for the purpose of this PR, I think.

ronaldfw:
Addressed comments. What do you think about adding the OrbitGl namespace? @beckerhe @pierricgimmig

beckerhe:
> Addressed comments. What do you think about adding the OrbitGl namespace? @beckerhe @pierricgimmig

Thanks a lot! I'm in favor of adding the `OrbitGl` namespace. I've done that so far also for other new code in OrbitQt and OrbitCore. All the new modules (OrbitLinuxTracing, OrbitSsh, ElfUtils) are namespaced anyway.

ronaldfw:
> > Addressed comments. What do you think about adding the OrbitGl namespace? @beckerhe @pierricgimmig
> 
> Thanks a lot! I'm in favor of adding the `OrbitGl` namespace. I've done that so far also for other new code in OrbitQt and OrbitCore. All the new modules (OrbitLinuxTracing, OrbitSsh, ElfUtils) are namespaced anyway.

Ok, done. I've only made the minimal change that the new free function is now in OrbitGl. I'll follow-up to expand the scope later.

PR #860, opened 2020-07-24
Clean up iterator UI when function live data changes

ronaldfw:
When live data changes, it is likely necessary to remove the iterators that exist. Even if they still make sense for the newly hooked functions, it would be hard to update all data correctly to the new capture. We therefore completely clean up the UI whenever the live data changes.

### DIFF
--- a/OrbitGl/LiveFunctionsController.cpp
+++ b/OrbitGl/LiveFunctionsController.cpp
@@ -223,3 +223,23 @@ TickType LiveFunctionsController::GetCaptureMin() {
 TickType LiveFunctionsController::GetCaptureMax() {
   return GCurrentTimeGraph->GetCaptureMax();
 }
+
+void LiveFunctionsController::Reset() {
+  function_iterators_.clear();
+  current_textboxes_.clear();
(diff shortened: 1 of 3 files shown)


ronaldfw:
> Hi Ronald! Seems nice but it isn't working in my workspace. The iterators are reseting as soon as I created them. Could be something related with Linux?

Ok, thanks, I'll take a look on Linux.

ronaldfw:
> > Hi Ronald! Seems nice but it isn't working in my workspace. The iterators are reseting as soon as I created them. Could be something related with Linux?
> 
> Ok, thanks, I'll take a look on Linux.

I can reproduce this, will need to take a closer look why this is happening later.

ronaldfw:
> I can reproduce this, will need to take a closer look why this is happening later.

This is actually fully reproducible on Windows as well. I will have to implement this differently. Doing the Reset() on OnDataChanged() does not work. This works when _loading_ a capture, but not when one is connected and has captured a capture from an instance. I had only tested this with loading a capture, so I never noticed.

PR #862, opened 2020-07-27
Tooltips for samples in the capture window

reichlfl:
Added tooltips for samples in the capture window.
The concept is the same as for the remaining tooltips, but there is one key difference: Samples should have a tooltip, but they are not pickable (so it is still possible to "click through" them onto the event track). To achieve this, the "picking" flag is now an enum that distinguishes between "hover" and "click", so each element can decide for itself for which of those actions it needs to draw itself into the picking buffer.
The picking enum is also passed to updatePrimitives(), which is consistent to how the method is being used anyways.

### DIFF
--- a/OrbitGl/EventTrack.cpp
+++ b/OrbitGl/EventTrack.cpp
@@ -5,6 +5,7 @@
 #include "EventTrack.h"
 
 #include "Capture.h"
+#include "SamplingProfiler.h"
 #include "EventTracer.h"
 #include "GlCanvas.h"
 #include "PickingManager.h"
@@ -25,13 +26,10 @@ void EventTrack::Draw(GlCanvas* canvas, bool picking) {
   Batcher* batcher = canvas->GetBatcher();
   PickingManager& picking_manager = canvas->GetPickingManager();
 
-  constexpr float kNormalZ = -0.1f;
-  constexpr float kPickingZ = 0.1f;
-  float z = kNormalZ;
+  constexpr float z = -0.1f;
   Color color = m_Color;
 
   if (picking) {
-    z = kPickingZ;
     color = picking_manager.GetPickableColor(this, PickingID::BatcherId::UI);
   }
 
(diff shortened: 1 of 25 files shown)


OrbitGl/EventTrack.cpp
```diff
@@ -21,21 +22,23 @@ std::string EventTrack::GetTooltip() const {
...
 
-  constexpr float kNormalZ = -0.1f;
-  constexpr float kPickingZ = 0.1f;
-  float z = kNormalZ;
+  const bool picking = picking_mode != PickingMode::kNone;
+  // The sample indicators are at z == 0 and do not respond to clicks, but
+  // have a tooltip. For picking, we want to draw the event bar over them if
+  // handling a click, and underneath otherwise.
+  // This simulates "click-through" behavior.
+  const float eventBarZ = picking_mode == PickingMode::kClick ? 0.1f : -0.1f;
```
ronaldfw, line 34: There are all these different z value definitions in GlCanvas, like GlCanvas::Z_VALUE_EVENT. Would it make sense to add them there?

OrbitGl/EventTrack.cpp
```diff
@@ -152,3 +181,37 @@ bool EventTrack::IsEmpty() const {
...
+  if (!callstack) {
+    return unknown_return_text;
+  }
+
+  std::string function_name = SafeGetFormattedFunctionName(callstack->m_Data[0]);
+  std::string result = absl::StrFormat(
+    "<b>%s</b><br/><i>Sampled event</i><br/><br/><b>Callstack:</b>",
+    function_name.c_str());
+  for (auto addr : callstack->m_Data) {
+    result = result + "<br/>" + SafeGetFormattedFunctionName(addr);
```
ronaldfw, line 214: I suppose this can become pretty long. Is this cut off at some point? If not, should it be?

OrbitGl/SchedulerTrack.h
```diff
@@ -11,10 +11,11 @@ class SchedulerTrack : public ThreadTrack {
  public:
   explicit SchedulerTrack(TimeGraph* time_graph);
   ~SchedulerTrack() override = default;
-
+
```
ronaldfw, line 14: Remove space. (Should be done automatically by clang format?)

OrbitGl/EventTrack.cpp
reichlfl: Good point, I moved them there.

OrbitGl/EventTrack.cpp
reichlfl: Currently it is not cut, no. The Tooltip can get quite large and yes, it's probably a good idea to account for that, even tough I found it acceptable in my tests. 

I'd merge this one though and take a look in a new PR.

PR #864, opened 2020-07-27
Raw serialization

IrinaShkviro:
This PR reduce amount of serialized data:

- for SamplingProfiler we need just m_Callstacks and m_UniqueCallstack, other data are filled during the ProcessSamples call

- FunctionCountMap also could be filled on load (probably, we can use Capture::GSelectedFunctionsMap instead of it)

- seems like Capture::GCallstacks and Capture::GProcessId are not used after loading so we can remove it from saving

- EventBuffer could be filled from Capture::GSamplingProfiler, so not necessary to save it too

### DIFF
--- a/OrbitCore/Capture.cpp
+++ b/OrbitCore/Capture.cpp
@@ -51,6 +51,7 @@ std::string Capture::GProcessName;
 std::unordered_map<int32_t, std::string> Capture::GThreadNames;
 std::unordered_map<uint64_t, LinuxAddressInfo> Capture::GAddressInfos;
 std::unordered_map<uint64_t, std::string> Capture::GAddressToFunctionName;
+std::unordered_map<uint64_t, std::string> Capture::GAddressToModuleName;
 Mutex Capture::GCallstackMutex;
 std::unordered_map<uint64_t, std::string> Capture::GZoneNames;
 TextBox* Capture::GSelectedTextBox = nullptr;
@@ -131,6 +132,7 @@ void Capture::ClearCaptureData() {
   GThreadNames.clear();
   GAddressInfos.clear();
   GAddressToFunctionName.clear();
+  GAddressToModuleName.clear();
   GZoneNames.clear();
   GSelectedTextBox = nullptr;
   GSelectedThreadId = 0;
--- a/OrbitCore/SamplingProfiler.cpp
+++ b/OrbitCore/SamplingProfiler.cpp
@@ -314,6 +314,16 @@ void SamplingProfiler::UpdateAddressInfo(uint64_t address) {
 
   Capture::GAddressToFunctionName[address] = function_name;
   Capture::GAddressToFunctionName[function_address] = function_name;
+
+  std::string module_name = "???";
+  std::shared_ptr<Module> module = m_Process->GetModuleFromAddress(address);
+  if (module != nullptr) {
+    module_name = module->m_Name;
+  } else if (address_info != nullptr) {
+    module_name = Path::GetFileName(address_info->module_name);
+  }
+  Capture::GAddressToModuleName[address] = module_name;
+  Capture::GAddressToModuleName[function_address] = module_name;
 }
 
 //-----------------------------------------------------------------------------
@@ -335,7 +345,10 @@ void SamplingProfiler::FillThreadSampleDataSampleReports() {
           100.f * numOccurences / threadSampleData.m_NumSamples;
 
       SampledFunction function;
-      function.m_Name = Capture::GAddressToFunctionName[address];
+      // GAddressToFunctionName and GAddressToModuleName should be filled in
+      // UpdateAddressInfo()
+      CHECK(Capture::GAddressToFunctionName.count(address) > 0);
+      function.m_Name = Capture::GAddressToFunctionName.at(address);
       function.m_Inclusive = inclusive_percent;
       function.m_Exclusive = 0.f;
       auto it = threadSampleData.m_ExclusiveCount.find(address);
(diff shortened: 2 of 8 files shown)


OrbitCore/SamplingProfiler.cpp
```diff
@@ -344,48 +354,15 @@ void SamplingProfiler::FillThreadSampleDataSampleReports() {
             100.f * it->second / threadSampleData.m_NumSamples;
       }
       function.m_Address = address;
-
-      std::shared_ptr<Module> module = m_Process->GetModuleFromAddress(address);
-      function.m_Module = module ? module->m_Name : "???";
+      function.m_Module = Capture::GAddressToModuleName[address];
```
beckerhe, line 357: I'm not sure if that works since function and module address are not the same. `GetModuleFromAddress` has a different behaviour than `GAddressToModuleName[]`. The former calls `upper_bound` while the latter only checks if an entry for the current address exists. Or did you completely revamp that and I'm wrong?

beckerhe: Never mind. It's fine. ;-)

antonrohr, line 357: I am always somewhat hesitant to use the std containers in this way, since [] will return the default constructed value if the key does not exist. I guess it is fine how it is currently done, but you could also add a CHECK if you are sure that this key always exists, or if you are unsure, maybe test and use "???" as before

OrbitGl/CaptureSerializer.cpp
```diff
@@ -83,16 +83,6 @@ void CaptureSerializer::SaveImpl(T& archive) {
     archive(functions);
   }
 
-  {
-    ORBIT_SIZE_SCOPE("Capture::GCallstacks");
-    archive(Capture::GCallstacks);
-  }
-
-  {
-    ORBIT_SIZE_SCOPE("Capture::GProcessId");
```
dpallotti, line 97: This is used in `SchedulerTrack::UpdatePrimitives` to highlight all and only the scheduling slices that belong to the same process, so it should not be removed until it's replaced by something else.

OrbitCore/SamplingProfiler.cpp
IrinaShkviro: Basically, I followed the logic that we have for Capture::GAddressToFunctionName: we fill map in UpdateAddressInfo() (with "???" if we don't have an information and with function/module name if we have) and then we use this map in other functions. In both functions (UpdateAddressInfo and FillThreadSampleDataSampleReports) we get information about addresses from the one place (m_UniqueCallstacks), so we shouldn't be in the situation when [] calls the default constructor.

OrbitGl/CaptureSerializer.cpp
IrinaShkviro: Yes, you're right, thanks for the catch!

OrbitCore/SamplingProfiler.cpp
IrinaShkviro: Should we use "at" instead of "[]" for this case then?

OrbitCore/SamplingProfiler.h
```diff
@@ -111,6 +107,22 @@ class SamplingProfiler {
   [[nodiscard]] const ThreadSampleData* GetSummary() const;
   [[nodiscard]] uint32_t GetCountOfFunction(uint64_t function_address) const;
 
+  // TODO(irinashkviro): remove this when move to protobuf
+  void SaveCallstacks() {
```
dpallotti, line 115: What's the difference between `m_Callstacks` and `callstacks_vector`, other than the former uses `BlockChain`, which is probably unnecessary? Could we just use one and then we don't need these two methods, which are actually a bit confusing?

dimitry-: agree - I do not think blockchain can be replaced with vector.

OrbitCore/Capture.cpp
```diff
@@ -51,6 +51,7 @@ std::string Capture::GProcessName;
 std::unordered_map<int32_t, std::string> Capture::GThreadNames;
 std::unordered_map<uint64_t, LinuxAddressInfo> Capture::GAddressInfos;
 std::unordered_map<uint64_t, std::string> Capture::GAddressToFunctionName;
+std::unordered_map<uint64_t, std::string> Capture::GAddressToModuleName;
```
dimitry-, line 54: Can these be moved to samplingprofilier? maybe not in this PR.

OrbitCore/SamplingProfiler.cpp
antonrohr: at() can throw exceptions, I would prefer having
`CHECK(Capture::GAddressToModuleName.count(address) != 0);`

OrbitCore/SamplingProfiler.h
IrinaShkviro: The `callstacks_vector` was introduced as a temporary solution and removed in the next PR (#865 ) with these 2 methods. Cereal don't know how to serialize blockchain until I write how to do it, but serialization is easy with the vector. If we don't want to replace blockchain with the vector, having a temporary field just for serialization is the easiest solution. Another option is to add cereal to blockchain, but I preferred to localize the serialization logic in SamplingProfiler in order to remove it in the next PR. Or am I missed something?

OrbitCore/SamplingProfiler.cpp
beckerhe: I would also prefer `at` here.

OrbitCore/SamplingProfiler.h
beckerhe: There are ways to do that without an intermediate vector, but not with the ORBIT_SERIALIZE_... macros.
Out-of-class serialization functions for BlockChain are also possible, but also more work than needed for a temporary fix. I would stick to this one.

OrbitCore/SamplingProfiler.cpp
IrinaShkviro: The general question is "When do we prefer to have an exception and when an abort()"?

OrbitCore/Capture.cpp
IrinaShkviro: Added it to my TODO list

IrinaShkviro:
Thanks a lot for the review! I'm going to rebase the branch now and merge to the master.

PR #865, opened 2020-07-27
Add protobuf for capture

IrinaShkviro:
This PR replace cereal with protobuf in capture serialization. FunctionStats, Function, CallstackEvent and LinuxAddressInfo were replaced with proto messages. Added Callstack, CaptureHeader and TimerData proto messages. All new protos are in the OrbitClientProtos.

Also, save the version with the captures to have better error messages for user when loading incompatible captures. 

The capture consists of CaptureHeader, CaptureData and Timers. Capture data are copied to protobuf message for now.


protos/timer.proto
```diff
@@ -0,0 +1,27 @@
...
+    kGpuActivity = 3;
+  }
+  Type type = 6;
+
+  int32 processor = 7;
+  uint64 callstack_hash = 8;
+  uint64 function_address = 9;
+  uint64 user_data_key = 10;
+  uint64 timeline_hash = 11;
+}
```
IrinaShkviro, line 27: @dpallotti I see your comment about completely changing the TimerData in the previous PR, but I would prefer to postpone it for now, because it requires a lot of changes

protos/function.proto
```diff
@@ -11,3 +11,33 @@ message FunctionStats {
...
+	kOrbitTimerStartAsync = 3;
+	kOrbitTimerStopAsync = 4;
+	kOrbitTrackInt = 5;
+	kOrbitTrackInt64 = 6;
+	kOrbitTrackUint = 7;
+	kOrbitTrackUint64 = 8;
+	kOrbitTrackFloat = 9;
+	kOrbitTrackDouble = 10;
+	kOrbitTrackFloatAsInt = 11;
+	kOrbitTrackDoubleAsInt64 = 12;
```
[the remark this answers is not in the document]
beckerhe: I think clang-format does only support languages supported by clang, so no proto. But we should follow https://developers.google.com/protocol-buffers/docs/style?hl=en

protos/function.proto
```diff
@@ -0,0 +1,43 @@
...
+// found in the LICENSE file.
+
+syntax = "proto3";
+
+message FunctionStats {
+  uint64 count = 1;
+  double total_time_ms = 2;
+  double average_time_ms = 3;
+  double min_ms = 4;
+  double max_ms = 5;
```
IrinaShkviro, line 12: @dpallotti suggested to use uint64 and nanosecond for these fields

OrbitCore/SamplingProfiler.cpp
```diff
@@ -217,13 +218,13 @@ void SamplingProfiler::ProcessSamples() {
...
-    const std::shared_ptr<CallStack> callstack = it.second;
+    const std::shared_ptr<HashedCallstack> callstack = it.second;
     // A "resolved callstack" is a callstack where every address is replaced by
     // the start address of the function (if known).
-    CallStack resolved_callstack = *callstack;
+    HashedCallstack resolved_callstack = *callstack;
 
-    for (uint32_t i = 0; i < callstack->m_Data.size(); ++i) {
-      uint64_t addr = callstack->m_Data[i];
+    for (int i = 0; i < callstack->data_size(); ++i) {
```
antonrohr, line 226: Maybe
```suggestion
    for (uint64_t address: callstack->mutable_data()) 
```

OrbitCore/Capture.cpp
```diff
@@ -237,14 +239,14 @@ void Capture::RegisterZoneName(uint64_t a_ID, const char* a_Name) {
 }
 
 //-----------------------------------------------------------------------------
-void Capture::AddCallstack(CallStack& a_CallStack) {
+void Capture::AddCallstack(HashedCallstack& a_CallStack) {
```
antonrohr, line 242:
```suggestion
void Capture::AddCallstack(HashedCallstack& callstack) {
```

OrbitCore/Capture.cpp
```diff
@@ -237,14 +239,14 @@ void Capture::RegisterZoneName(uint64_t a_ID, const char* a_Name) {
...
   ScopeLock lock(GCallstackMutex);
-  Capture::GCallstacks[a_CallStack.Hash()] =
-      std::make_shared<CallStack>(a_CallStack);
+  Capture::GCallstacks[a_CallStack.callstack_hash()] =
+      std::make_shared<HashedCallstack>(a_CallStack);
 }
 
 //-----------------------------------------------------------------------------
-std::shared_ptr<CallStack> Capture::GetCallstack(CallstackID a_ID) {
+std::shared_ptr<HashedCallstack> Capture::GetCallstack(CallstackID a_ID) {
```
antonrohr, line 249:
```suggestion
std::shared_ptr<HashedCallstack> Capture::GetCallstack(CallstackID hash) {
```
or callstack_hash

OrbitCore/Capture.h
```diff
@@ -63,13 +63,14 @@ class Capture {
   static std::shared_ptr<SamplingProfiler> GSamplingProfiler;
   static std::shared_ptr<Process> GTargetProcess;
   static std::shared_ptr<Preset> GSessionPresets;
-  static std::shared_ptr<CallStack> GSelectedCallstack;
+  static std::shared_ptr<HashedCallstack> GSelectedCallstack;
```
antonrohr, line 66: I think this global is only used in CaptureWindow and there just in two lines right next to each other. I am in favor of removing it.

OrbitCore/SamplingProfiler.cpp
```diff
@@ -177,16 +178,16 @@ void SamplingProfiler::ProcessSamples() {
...
       // exclusive stat
-      threadSampleData.m_ExclusiveCount[resolvedCallstack->m_Data[0]] +=
+      threadSampleData.m_ExclusiveCount[resolvedCallstack->data(0)] +=
           callstackCount;
 
       std::set<uint64_t> uniqueAddresses;
-      for (uint32_t i = 0; i < resolvedCallstack->m_Data.size(); ++i) {
-        uniqueAddresses.insert(resolvedCallstack->m_Data[i]);
+      for (int i = 0; i < resolvedCallstack->data_size(); ++i) {
+        uniqueAddresses.insert(resolvedCallstack->data(i));
```
antonrohr, line 190: insert a copy of the whole vector instead of each element

protos/function.proto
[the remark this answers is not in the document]
antonrohr: $ clang-format -help
OVERVIEW: A tool to format C/C++/Java/JavaScript/Objective-C/__Protobuf__/C# code.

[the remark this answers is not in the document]
antonrohr: My vscode "Format Document" (via clangd extension) works out of the box. I dont know about VS or CLion

IrinaShkviro:
> I somehow started going through this like a regular PR. Lets talk about this in the meeting and then I continue

Yes, let's discuss it in the meeting, then I'll update PR and mark it as "ready for review".

protos/capture_data.proto
```diff
@@ -13,3 +13,33 @@ message FunctionStats {
   uint64 min_ns = 4;
   uint64 max_ns = 5;
 }
+
+message FunctionInfo {
```
IrinaShkviro, line 17: FunctionInfo was introduced only to save hooked functions (that are stored in GSelectedFunctionsMap). Fields `file`, `line`, `type` and `pretty_name` aren't used in context of hooked functions, but the `stats` field is only useful for hooked functions (in other cases it's not filled).

protos/capture_data.proto
```diff
@@ -0,0 +1,10 @@
+// Copyright (c) 2020 The Orbit Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+syntax = "proto3";
+
+message CaptureHeader {
+  string capture_name = 1;
```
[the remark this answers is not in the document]
IrinaShkviro: Yes, it's just a filename and used only in Save(). As we don't use it when loading capture we can remove it for now.

protos/function.proto
IrinaShkviro: Updated these fields

protos/callstack.proto
```diff
@@ -9,3 +9,8 @@ message CallstackEvent {
   uint64 callstack_hash = 2;
   int32 thread_id = 3;
 }
+
+message HashedCallstack {
+  uint64 callstack_hash = 1;
```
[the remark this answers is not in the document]
IrinaShkviro: `callstack_hash` is calculated based on data vector, so storing only the vector would be enough. Updated the PR and replaced this proto message with `CallstackInfo`

protos/address_info.proto
```diff
@@ -0,0 +1,12 @@
+// Copyright (c) 2020 The Orbit Authors. All rights reserved.
```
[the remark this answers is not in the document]
IrinaShkviro: Added all of them into one new file, but from the other hand, it's a bit confusing to include "capture_data.pb.h" when you need just CallstackEvent or FunctionInfo, for example.

IrinaShkviro:
The changes are:
- Moved all the protos in the one file (`capture_data.proto`)
- Renamed proto messages
- Introduced CallstackInfo and save data without hash instead of replacing current CallStack struct (with `hash` field)
- Used nanoseconds instead of milliseconds in FunctionStats
- didn't change Function and FunctionsStats; only translated these structures to the proto messages
- added a package specifier in the new proto file

OrbitCore/Capture.cpp
IrinaShkviro: I think these classes require some additional polishing. I reverted most of the changes for now as those changes are not necessary for this task. So would you mind if I do it in the separate PR?

OrbitCore/Capture.h
IrinaShkviro: Yes, and the same for Capture::GCallstacks; probably, I should remove both.

protos/function.proto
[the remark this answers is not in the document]
dpallotti: Ctrl+Alt+L in CLion works fine for me with protos.

protos/capture_data.proto
```diff
@@ -0,0 +1,15 @@
+// Copyright (c) 2020 The Orbit Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+syntax = "proto3";
+
+package protos;
```
dpallotti, line 7: Don't we want a more specific namespace name that refers to the fact that these are the protos for file serialization? Otherwise all protos should end up in this same namespace and I assume that's not what we want?

OrbitCore/Utils.h
```diff
@@ -290,6 +291,10 @@ inline std::string GetPrettySize(uint64_t size) {
   return absl::StrFormat("%.2f TB", size / TB);
 }
 
+inline double GetMillisecondsFromNanoseconds(uint64_t nanoseconds) {
```
dpallotti, line 294: Do we really want this method to wrap an `absl` one-liner into another one-liner?

OrbitCore/FunctionUtils.cpp
```diff
@@ -14,95 +14,114 @@
...
 }
 
-bool IsOrbitFunc(const Function& func) {
-  return func.orbit_type() != Function::OrbitType::NONE;
+bool IsOrbitFunc(const protos::FunctionInfo& func) {
+  return func.type() != protos::FunctionInfo::kNone;
 }
 
-void Select(Function* func) {
+std::shared_ptr<protos::FunctionInfo> CreateFunction(
```
dpallotti, line 33: `CreateFunctionInfo`?

OrbitGl/CaptureSerializer.cpp
```diff
@@ -72,15 +72,15 @@ void CaptureSerializer::SaveImpl(T& archive) {
...
     for (auto& pair : Capture::GSelectedFunctionsMap) {
-      Function* func = pair.second;
+      protos::FunctionInfo* func = pair.second;
       if (func != nullptr) {
         functions.push_back(*func);
       }
     }
 
-    archive(functions);
+    // archive(functions);
```
dpallotti: Why is this commented?

OrbitGl/CaptureSerializer.cpp
```diff
@@ -26,87 +30,107 @@
...
-  cereal::BinaryOutputArchive CountingArchive(Stream);
+  google::protobuf::io::OstreamOutputStream out_counting_stream(&Stream);
+  google::protobuf::io::CodedOutputStream coded_counting_output(
+      &out_counting_stream);
+
   GStreamCounter.Reset();
-  SaveImpl(CountingArchive);
+  SaveImpl(&coded_counting_output);
   PRINT_VAR(GStreamCounter.Size());
   GStreamCounter.Reset();
```
dpallotti, line 67: I know it has always been there, but honestly do we even need this part? This serializes just to print the size? It's probably quite an unnecessary overhead.

OrbitGl/CaptureSerializer.cpp
```diff
@@ -138,12 +159,24 @@ ErrorMessageOr<void> CaptureSerializer::Load(const std::string& filename) {
...
+}
+
+bool ReadMessage(google::protobuf::Message* message,
+                 google::protobuf::io::CodedInputStream* input) {
+  uint32_t message_size;
+  if (!input->ReadLittleEndian32(&message_size)) {
+    return false;
   }
+
+  char* buffer = new char[message_size];
```
dpallotti, line 172: Is this a case for `make_unique_for_overwrite` that is currently in `OrbitLinuxTracing`? Maybe we should move it to `OrbitBase`?

protos/capture_data.proto
IrinaShkviro: Yes, and also I should move `capture_data` to another folder. But I'm still thinking what name/structure would be better: `protos.ui.serialization` / `protos.file_serialization` / `protos.serialization.file`?

OrbitCore/FunctionUtils.cpp
IrinaShkviro: Yes, thanks, fixed it

OrbitGl/CaptureSerializer.cpp
IrinaShkviro: I removed all this stuff in c431424dc4d4d772a8585c33631ceb205c8eb9c7 when added serialization with protobuf, but it's still a part of the history. I'll update commits after the review.

protos/capture_data.proto
```diff
@@ -0,0 +1,15 @@
...
+
+message FunctionStats {
+  uint64 count = 1;
+  uint64 total_time_ns = 2;
+  uint64 average_time_ns = 3;
+  uint64 min_ns = 4;
+  uint64 max_ns = 5;
+}
```
dimitry-, line 15: maybe move these to OrbitGl? they are not used on the service side are they?

OrbitGl/CaptureSerializer.cpp
IrinaShkviro: I don't know if somebody uses it. But it looks like I already removed the useful part and now it shows only the total  size which is not so interesting. Btw, if we need the size we always can just change WriteMessage and log message_size.

OrbitCore/Utils.h
dimitry-: if you decide to leave it probably rename to NanosecondsToMilliseconds()

OrbitCore/Profiling.h
```diff
@@ -28,6 +28,10 @@ inline TickType OrbitTicks(uint32_t a_Clock = 1 /*CLOCK_MONOTONIC*/) {
   return 1000000000ll * ts.tv_sec + ts.tv_nsec;
 }
 
+inline uint64_t NanoSecondsFromTicks(TickType start, TickType end) {
```
dimitry-, line 33: TicksToNanoseconds? or at least s/NanoSeconds/Nanoseconds/

protos/capture_data.proto
```diff
@@ -13,3 +13,33 @@ message FunctionStats {
...
+    kOrbitTrackUint = 7;
+    kOrbitTrackUint64 = 8;
+    kOrbitTrackFloat = 9;
+    kOrbitTrackDouble = 10;
+    kOrbitTrackFloatAsInt = 11;
+    kOrbitTrackDoubleAsInt64 = 12;
+  }
+  OrbitType type = 10;
+  FunctionStats stats = 11;
+}
```
dimitry-, line 45: looking at existing model - seems like Function should be an aggregation of SymbolInfo, ModuleSymbols & FunctionStats, or am I missing something?

OrbitGl/LiveFunctionsDataView.cpp
```diff
@@ -327,14 +326,15 @@ void LiveFunctionsDataView::OnTimer() {
 }
 
 //-----------------------------------------------------------------------------
-Function& LiveFunctionsDataView::GetFunction(unsigned int a_Row) const {
+protos::FunctionInfo& LiveFunctionsDataView::GetFunction(
```
dimitry-, line 330: should probably be a pointer.

OrbitGl/LiveFunctionsDataView.cpp
```diff
@@ -327,14 +326,15 @@ void LiveFunctionsDataView::OnTimer() {
...
+    unsigned int a_Row) const {
   CHECK(a_Row < m_Functions.size());
   CHECK(m_Functions[m_Indices[a_Row]]);
   return *m_Functions[m_Indices[a_Row]];
 }
 
 std::pair<TextBox*, TextBox*> LiveFunctionsDataView::GetMinMax(
-    Function& function) const {
+    protos::FunctionInfo& function) const {
```
dimitry-, line 338: const?

protos/capture_data.proto
```diff
@@ -56,3 +56,7 @@ message LinuxAddressInfo {
   string function_name = 3;
   uint64 offset_in_function = 4;
 }
+
+message CaptureHeader {
+  uint32 version = 1;
```
dimitry-, line 65: add/replace with string, we need something to show to users. (This capture was saved with v1.42 earliest supported version is v1.80...)

OrbitGl/CaptureSerializer.cpp
dimitry-: yes - it also provides scoping of the buffer

OrbitGl/CaptureSerializer.cpp
```diff
@@ -26,87 +30,107 @@
 #include "absl/strings/str_format.h"
 #include "capture_data.pb.h"
 
+static absl::flat_hash_map<const uint32_t, const std::string>
+    capture_version_to_orbit_version_map{
```
dimitry-, line 34: this also works.. but now we need to keep int -> string mapping in all of our versions, where if we save it to proto - we will only need to keep the latest supported one in the code.

OrbitCore/OrbitModule.cpp
```diff
@@ -53,7 +53,7 @@ void Module::LoadSymbols(const ModuleSymbols& module_symbols) {
 
   for (const SymbolInfo& symbol_info : module_symbols.symbol_infos()) {
     std::shared_ptr<protos::FunctionInfo> function =
-        FunctionUtils::CreateFunction(
+        FunctionUtils::CreateFunctionInfo(
         symbol_info.name(), symbol_info.demangled_name(), symbol_info.address(),
         module_symbols.load_bias(), symbol_info.size(),
         symbol_info.source_file(), symbol_info.source_line(), m_FullName,
```
dimitry-, line 59: m_FullName and m_AddressStart should also be used from symbol_info.

dimitry-: which also means this function is not using anything from Module class.

protos/capture_data.proto
IrinaShkviro: These messages aren't used on the service side. Does `protos` only for service? I thought that it's for all the protos, and because of it placed `capture_data` here. But talking about moving it to another module: why to OrbitGl and not to OrbitCore? We use CallstackEvent and FunctionInfo in OrbitCore, so probably we should move to OrbitCore or change the logic.

OrbitCore/Utils.h
IrinaShkviro: Let's just remove it for now. Ideally, I wanted to have time always in one format (nanoseconds for example) and then make a conversion only when show time to the user. In this case, we can localize all the conversion in one place.

protos/capture_data.proto
IrinaShkviro: let's discuss it offline with @antonrohr

OrbitCore/Profiling.h
antonrohr: Since the other methods in this class are also using the from naming, I would be okay with leaving it as is. Or changing all the names (maybe todo message -> other PR)

OrbitGl/CaptureSerializer.cpp
IrinaShkviro: One more idea: do we want to show the exact version for the user, or the range? For example, "this capture could be opened with v1.42-v1.45" (for this I should change the type of the map and the error message). In this case if capture was taken in 1.43, we won't require v1.43 to open the capture, but any version from the range. But we don't know the max version when we save the capture, and for this purpose mapping could be useful.

OrbitGl/CaptureSerializer.cpp
```diff
@@ -24,114 +28,94 @@
...
-    archive(Capture::GCallstacks);
-  }
+  capture_info->mutable_thread_names()->insert(Capture::GThreadNames.begin(),
+                                               Capture::GThreadNames.end());
 
-  {
-    ORBIT_SIZE_SCOPE("Capture::GProcessId");
-    archive(Capture::GProcessId);
+  capture_info->mutable_address_infos()->Reserve(Capture::GAddressInfos.size());
+  for (const auto& address_indo : Capture::GAddressInfos) {
```
antonrohr, line 80:
```suggestion
  for (const auto& address_info : Capture::GAddressInfos) {
```

OrbitGl/LiveFunctionsDataView.cpp
```diff
@@ -49,28 +51,33 @@ std::string LiveFunctionsDataView::GetValue(int a_Row, int a_Column) {
...
+      return GetPrettyTime(absl::ToDoubleMilliseconds(
+          absl::Nanoseconds(stats.average_time_ns())));
     case COLUMN_TIME_MIN:
-      return GetPrettyTime(stats->m_MinMs);
+      return GetPrettyTime(
+          absl::ToDoubleMilliseconds(absl::Nanoseconds(stats.min_ns())));
     case COLUMN_TIME_MAX:
-      return GetPrettyTime(stats->m_MaxMs);
+      return GetPrettyTime(
+          absl::ToDoubleMilliseconds(absl::Nanoseconds(stats.max_ns())));
```
antonrohr, line 75: I am aware of @dpallotti comment from earlier. This here seems like an unnecessary amount of code duplication. I would propose to add a second GetPrettyTime to Utils.h
`std::string GetPrettyTime(uint64_t nanoseconds) { ... }`

IrinaShkviro: Ideally, I wanted to replace `std::string GetPrettyTime(double milliseconds)` with `std::string GetPrettyTime(uint64_t nanoseconds)` in the next PR to reduce amount  of changes. But probably I can do it in two steps (just add a second function for now and then get rid of the first one).

beckerhe: Please no overloads for two numeric types with different units. It's bad enough to convey the unit in the name of the parameter. GetPrettyTime should take a absl::Duration.

OrbitCore/Profiling.h
IrinaShkviro: I changed the name to `TicksToNanoseconds` and also added TODO in the code.

OrbitGl/LiveFunctionsDataView.cpp
IrinaShkviro: I like the idea to use absl::Duration, and it doesn't require too much changes. I added a commit with this change (469cf9900af0df4f43f87b3dbcbd7ad52765f700).

OrbitCore/Utils.h
```diff
@@ -31,6 +31,7 @@
 #include "absl/strings/match.h"
 #include "absl/strings/str_format.h"
 #include "absl/strings/str_split.h"
+#include "absl/time/time.h"
```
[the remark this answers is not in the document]
IrinaShkviro: I should remove it in 7803da53851fecc6423121dc5445bcf10ca5a3c2, thanks for noticing!

OrbitGl/CaptureSerializer.cpp
IrinaShkviro: Fixed, thanks!

OrbitCore/SamplingProfiler.h
```diff
@@ -74,8 +71,8 @@ class SamplingProfiler {
 
   int GetNumSamples() const { return m_NumSamples; }
 
-  void AddCallStack(CallstackEvent& callstack_event);
-  void AddHashedCallStack(CallstackEvent& a_CallStack);
+  void AddCallStack(protos::CallstackEvent& callstack_event);
+  void AddHashedCallStack(protos::CallstackEvent& a_CallStack);
```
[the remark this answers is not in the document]
IrinaShkviro: Thanks, renamed the argument.

OrbitGl/LiveFunctionsDataView.cpp
IrinaShkviro: Yes, thanks!

OrbitGl/LiveFunctionsDataView.cpp
IrinaShkviro: Fixed it

OrbitGl/CaptureSerializer.cpp
IrinaShkviro: I added the solution with the range of the appropriate Orbit versions that gives more information to the user.

OrbitGl/CaptureSerializer.cpp
IrinaShkviro: It's done, thanks!

OrbitUIProtos/capture_data.proto
```diff
@@ -4,7 +4,7 @@
 
 syntax = "proto3";
 
-package protos;
+package orbit_ui_protos;
```
dpallotti, line 7: Soo, maybe this was discussed offline, but I'm not a big fun of this namespace name. "orbit" feels redundant and "ui" not too appropriate. Maybe "client_protos"?
Same for the module/directory name: I guess we want to keep "Orbit" for consistency with the other modules there, but maybe "OrbitClientProtos"?

OrbitUIProtos/CMakeLists.txt
```diff
@@ -0,0 +1,11 @@
+# Copyright (c) 2020 The Orbit Authors. All rights reserved.
+# Use of this source code is governed by a BSD-style license that can be
+# found in the LICENSE file.
+
+cmake_minimum_required(VERSION 3.15)
+
+project(OrbitUIProtos)
```
dpallotti, line 7: As stated in the other comment, maybe "OrbitClientProtos"?

OrbitCore/Profiling.h
```diff
@@ -30,7 +30,8 @@ inline TickType OrbitTicks(uint32_t a_Clock = 1 /*CLOCK_MONOTONIC*/) {
   return 1000000000ll * ts.tv_sec + ts.tv_nsec;
 }
 
-inline uint64_t NanoSecondsFromTicks(TickType start, TickType end) {
+// TODO: use similar naming for other functions
```
dpallotti, line 33: Maybe it can just be done? It's only two functions.

OrbitCore/SamplingProfiler.cpp
```diff
@@ -74,12 +74,12 @@ void SamplingProfiler::AddCallStack(CallstackEvent& callstack_event) {
 }
 
 //-----------------------------------------------------------------------------
-void SamplingProfiler::AddHashedCallStack(CallstackEvent& a_CallStack) {
-  if (!HasCallStack(a_CallStack.callstack_hash())) {
+void SamplingProfiler::AddHashedCallStack(CallstackEvent& callstack_event) {
```
dpallotti, line 77: Are both `AddCallstack` and `AddHashedCallstack` needed now? After recent changes they conceptually do the same thing, except that the former acts when `!HasCallStack` and the latter prints an error. Can this one just be removed or am I missing something?

OrbitGl/CaptureSerializer.cpp
```diff
@@ -24,94 +29,101 @@
...
+
+using orbit_ui_protos::CallstackEvent;
+using orbit_ui_protos::CallstackInfo;
+using orbit_ui_protos::CaptureInfo;
+using orbit_ui_protos::FunctionInfo;
+using orbit_ui_protos::TimerInfo;
+
+static absl::flat_hash_map<const uint32_t,
+                           std::pair<const std::string, const std::string>>
+    capture_version_to_orbit_versions_map{
```
dpallotti, line 42: Maintaining this mapping and the ranges feels like quite a burden to me. Honestly I would lean towards Dimitry's proposal of just saving the version string in the protos.

IrinaShkviro: It looks a bit confusing for now, because the minimum and maximum version are the same. I think about the situation when we have range of 10 for example captures from different versions that are compatible and trying to open it with a newer version of Orbit. For all 10 we will get the error saying the exact version and to open it we should have all the 10 versions (according to error). But if we show the range, the user knows that it's enough to have just one version for all the captures. But it depends also on how this map looks like: if we always have min = max it doesn't make sense at all.

OrbitUIProtos/capture_data.proto
IrinaShkviro: I don't remember if we agree on the certain name, so we can discuss it now. @dimitry- @antonrohr @beckerhe any other suggestions?

dimitry-: we haven't we just discussed separating it from current protos - client sounds good to me, but I do not have strong feelings one way or the other.

OrbitCore/Profiling.h
IrinaShkviro: It's done

OrbitCore/Profiling.h
```diff
@@ -37,7 +39,14 @@ inline double MicroSecondsFromTicks(TickType a_Start, TickType a_End) {
   return double((a_End - a_Start)) * 0.001;
 }
 
+inline double MilliSecondsFromTicks(TickType start, TickType end) {
```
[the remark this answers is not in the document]
IrinaShkviro: Yes, I renamed functions in Profiling.h, thanks!

OrbitCore/SamplingProfiler.cpp
IrinaShkviro: Removed AddHashedCallstack.

OrbitGl/CaptureSerializer.cpp
dpallotti: It's not so much about how it looks, I'm more unsure about whether it's useful to maintain such a map in relation to the fact that it has to be manually maintained and updated for every release. Considering that we're still in beta, we're not committing to maintaining compatibility yet, and that we are probably breaking compatibility at every release, maybe for user it's just better to get a message saying "This capture was created with version ... and is not compatible with the current version" (by saving the version string in the proto) than suggesting them a range of versions to use to open it.
But I might very well be wrong here. Does anyone else have any thoughts about this?

beckerhe: True, having such a message might be very convenient for the user. So probably we want to introduce this map (or something similar at some point). But for me it feels a bit premature to do it now. We are not loosing anything by postponing the introduction of such a map to the point where we introduce the first (backwards-compatible) breaking change. And I hope at that point we will have a concept on how to handle breaking changes and for how long and so on.

OrbitGl/CaptureWindow.cpp
```diff
@@ -551,12 +554,12 @@ void CaptureWindow::Draw() {
     float from = time_graph_.GetWorldFromTick(minTime);
     float to = time_graph_.GetWorldFromTick(maxTime);
 
-    double micros = MicroSecondsFromTicks(minTime, maxTime);
+    uint64_t nanos = TicksToNanoseconds(minTime, maxTime);
```
beckerhe, line 557: Maybe TicksToDuration? :innocent:

OrbitGl/CaptureSerializer.cpp
IrinaShkviro: Ok, let's postpone it for now. I removed the map and changed int to string for `version` field in "capture_data.proto".

OrbitGl/CaptureWindow.cpp
IrinaShkviro: Done, thanks!

OrbitCore/StringManager.h
```diff
@@ -21,7 +20,9 @@ class StringManager {
   bool Contains(uint64_t key);
   void Clear();
 
-  ORBIT_SERIALIZABLE;
+  const absl::flat_hash_map<uint64_t, std::string> GetKeyToStringMap() const {
```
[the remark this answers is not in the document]
IrinaShkviro: Yes, thanks, fixed

OrbitUIProtos/capture_data.proto
IrinaShkviro: Ok, I'm not against OrbitClientProtos, and as no one objects, I renamed it to OrbitClientProtos.

OrbitUIProtos/CMakeLists.txt
IrinaShkviro: Renamed it, thanks!

OrbitGl/CaptureSerializer.cpp
```diff
@@ -190,31 +196,76 @@ ErrorMessageOr<void> CaptureSerializer::Load(std::istream& stream) {
...
+      "Please check release notes for more information.";
+
+  if (!ReadMessage(&header, &coded_input) || header.version().empty()) {
+    ERROR("%s", error_message);
+    return ErrorMessage(error_message);
+  }
+  if (header.version() != kRequiredCaptureVersion) {
+    std::string incompatible_version_error_message = absl::StrFormat(
+        "This capture format is no longer supported but could be opened with "
+        "%s Orbit version.",
```
dpallotti, line 252: I would maybe put the version at the end, e.g. "This capture format is no longer supported but could be opened with Orbit version %s."

OrbitGl/CaptureSerializer.cpp
```diff
@@ -190,31 +196,76 @@ ErrorMessageOr<void> CaptureSerializer::Load(std::istream& stream) {
...
   FillEventBuffer();
+}
+
+ErrorMessageOr<void> CaptureSerializer::Load(std::istream& stream) {
+  google::protobuf::io::IstreamInputStream input_stream(&stream);
+  google::protobuf::io::CodedInputStream coded_input(&input_stream);
+
+  std::string error_message =
+      "Error parsing the capture.\nNote: If the capture "
+      "was taken in a previous Orbit version, it could be incompatible. "
```
dpallotti, line 242:
```suggestion
      "was taken with a previous Orbit version, it could be incompatible. "
```

IrinaShkviro: Fixed, thanks!

OrbitGl/CaptureSerializer.cpp
IrinaShkviro: Done

IrinaShkviro:
Thanks a lot for the review! I rebased the branch and clean up commit history (probably too early, sorry for that 😬).

PR #866, opened 2020-07-27
Add Vertical Zoom by scaling drawn elements

karupayun:
We can make a vertical zoom pressing ctrl + wheel. For this, we are
scaling the heights of each element in the TimeGraph. Also, we are preserving the
distance from the mouse to make a proper zoom similar to the horizontal
one.

This is actually in draft, to receive feedback.

### DIFF
--- a/OrbitGl/CaptureWindow.cpp
+++ b/OrbitGl/CaptureWindow.cpp
@@ -382,19 +382,21 @@ void CaptureWindow::MouseWheelMoved(int a_X, int a_Y, int a_Delta,
   if (delta > m_MaxWheelDelta) m_MaxWheelDelta = delta;
 
   float mousex = a_X;
+  float mousey = a_Y;
 
   float worldx;
   float worldy;
 
   ScreenToWorld(a_X, a_Y, worldx, worldy);
-  m_MouseRatio = static_cast<double>(mousex) / getWidth();
 
   bool zoomWidth = !a_Ctrl;
   if (zoomWidth) {
+    m_MouseRatio = static_cast<double>(mousex) / getWidth();
     time_graph_.ZoomTime(delta, m_MouseRatio);
     m_WheelMomentum = delta * m_WheelMomentum < 0 ? 0 : m_WheelMomentum + delta;
   } else {
-    // TODO: reimplement vertical zoom by scaling track heights.
+    double mouse_relative_y_position = static_cast<double>(mousey) / getHeight();
+    time_graph_.VerticalZoom(delta, mouse_relative_y_position);
   }
 
   // Use the original sign of a_Delta here.
--- a/OrbitGl/TimeGraph.cpp
+++ b/OrbitGl/TimeGraph.cpp
@@ -196,6 +196,34 @@ void TimeGraph::ZoomTime(float a_ZoomValue, double a_MouseRatio) {
   SetMinMax(minTimeUs, maxTimeUs);
 }
 
+void TimeGraph::VerticalZoom(float zoom_value, double mouse_relative_position) {
+  static double increment_ratio = 0.1;
+
+  double ratio = zoom_value > 0 ? 1 + increment_ratio : 1 - increment_ratio;
+
+  float world_height = m_Canvas->GetWorldHeight();
+  double y_mouse_position = m_Canvas->GetWorldTopLeftY() - mouse_relative_position * world_height;
+  double top_distance = m_Canvas->GetWorldTopLeftY() - y_mouse_position;
+
+  double new_y_mouse_position = y_mouse_position / ratio;
+
+  float new_world_top_left_y = new_y_mouse_position + top_distance;
+
+  // If we zoomed-out, we would like to see most part of the screen with events,
+  // so we set a minimum and maximum for the y-top coordinate.
+  new_world_top_left_y =
+      std::max(new_world_top_left_y, world_height - GetThreadTotalHeight());
+  // TODO: TopMargin has to be 1.5f * m_Layout.GetSliderWidth()?
+  new_world_top_left_y =
+      std::min(new_world_top_left_y, 1.5f * m_Layout.GetSliderWidth());
+
+  m_Canvas->SetWorldTopLeftY(new_world_top_left_y);
+
+  // Finally, we have to scale every item in the layout.
+  float old_scale = m_Layout.GetScale();
+  m_Layout.SetScale(old_scale / ratio);
+}
+
 //-----------------------------------------------------------------------------
 void TimeGraph::SetMinMax(double a_MinTimeUs, double a_MaxTimeUs) {
   double desiredTimeWindow = a_MaxTimeUs - a_MinTimeUs;
--- a/OrbitGl/TimeGraphLayout.cpp
+++ b/OrbitGl/TimeGraphLayout.cpp
@@ -35,6 +35,7 @@ TimeGraphLayout::TimeGraphLayout() {
   m_TextZ = -0.02f;
   m_TrackZ = -0.1f;
   m_ToolbarIconHeight = 24.f;
+  scale_ = 1.f;
   time_bar_height_ = 15.f;
 };
 
@@ -80,6 +81,7 @@ bool TimeGraphLayout::DrawProperties() {
   FLOAT_SLIDER_MIN_MAX(m_TextZ, -1.f, 1.f);
   FLOAT_SLIDER_MIN_MAX(m_TrackZ, -1.f, 1.f);
   FLOAT_SLIDER(m_ToolbarIconHeight);
+  FLOAT_SLIDER_MIN_MAX(scale_, 0.05f, 20.f);
   ImGui::Checkbox("DrawTrackBackground", &m_DrawTrackBackground);
   ImGui::End();
 
@@ -89,5 +91,5 @@ bool TimeGraphLayout::DrawProperties() {
 float TimeGraphLayout::GetBottomMargin() const {
   // The bottom consists of the slider (where we have to take the width, as it
   // is rotated), and the time bar.
-  return GetSliderWidth() + GetTimeBarHeight(); 
+  return GetSliderWidth() + GetTimeBarHeight();
 }
--- a/OrbitGl/TimeGraph.h
+++ b/OrbitGl/TimeGraph.h
@@ -64,6 +64,7 @@ class TimeGraph {
   void Zoom(const TextBox* a_TextBox);
   void Zoom(TickType min, TickType max);
   void ZoomTime(float a_ZoomValue, double a_MouseRatio);
+  void VerticalZoom(float a_ZoomValue, double a_MouseRatio);
   void SetMinMax(double a_MinTimeUs, double a_MaxTimeUs);
   void PanTime(int a_InitialX, int a_CurrentX, int a_Width,
                double a_InitialTime);
--- a/OrbitGl/TimeGraphLayout.h
+++ b/OrbitGl/TimeGraphLayout.h
@@ -10,35 +10,39 @@ class TimeGraphLayout {
  public:
   TimeGraphLayout();
 
-  float GetTextBoxHeight() const { return m_TextBoxHeight; }
-  float GetTextCoresHeight() const { return m_CoresHeight; }
-  float GetEventTrackHeight() const { return m_EventTrackHeight; }
-  float GetGraphTrackHeight() const { return m_GraphTrackHeight; }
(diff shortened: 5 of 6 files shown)


ronaldfw:
What does this do with background elements, like the track backgrounds, and in particular the rounded corners? I recall this was one of the issues why @pierricgimmig turned this off at some point.

karupayun:
> What does this do with background elements, like the track backgrounds, and in particular the rounded corners? I recall this was one of the issues why @pierricgimmig turned this off at some point.

These are images when you zoom-in and zoom-out. I have a TODO related with the text-size.

Screen Shot 2020-07-27 at 14 23 53
Screen Shot 2020-07-27 at 14 22 51
Screen Shot 2020-07-27 at 14 23 25

OrbitGl/TimeGraphLayout.cpp
```diff
@@ -85,3 +85,17 @@ bool TimeGraphLayout::DrawProperties() {
...
+  m_EventTrackHeight *= scale;
+  m_GraphTrackHeight *= scale;
+  m_SpaceBetweenCores *= scale;
+  m_SpaceBetweenTracks *= scale;
+  m_SpaceBetweenTracksAndThread *= scale;
+  m_TrackTabHeight *= scale;
+  m_RoundingRadius *= scale;
+  m_TrackBottomMargin *= scale;
+  m_TrackTopMargin *= scale;
+}
```
pierricgimmig, line 101: Scale should be kept as it's own member variable and all the other variables should be kept independent of the scale. The getters should return the result of the scale multiplied by the value. The problem is that the floating point values will drift over time and we won't be able to return to the exact initial values as floating point error will accumulate forever. Also, having scale separate will allow us to have an interface to set scale directly without having to go through many increments. A reset could then be implemented easily by calling SetScale(1.f);

karupayun: You are completely right

karupayun:
Well, I think that it's ready for review. 

I'm not sure what to do with the text size. I think that one option could be scale similarly having a maximum size. But still have to think how to change it without change the size of the text in the bottom part (the pretty times). Maybe for another PR.

OrbitGl/TimeGraph.cpp
```diff
@@ -189,6 +189,24 @@ void TimeGraph::ZoomTime(float a_ZoomValue, double a_MouseRatio) {
   SetMinMax(minTimeUs, maxTimeUs);
 }
 
+void TimeGraph::VerticalZoom(float zoom_value, double mouse_ratio) {
+  static double increment_ratio = 0.1;
+
+  // In this way, VerticalZoom is reversible
```
ronaldfw, line 195: Can you add more of an explanation what "reversible" here means?

karupayun: The comment was only for reviewers, I'm going to erase it.
In the horizontal zoom, we use (1 + increment_ratio) and (1 - increment_ratio). These aren't reversibles (if you zoom-in and after zoom-out you aren't getting the initial ratio). So, after it, I want to change the other one also.

OrbitGl/CaptureWindow.cpp
```diff
@@ -376,19 +376,23 @@ void CaptureWindow::MouseWheelMoved(int a_X, int a_Y, int a_Delta,
...
-  m_MouseRatio = static_cast<double>(mousex) / getWidth();
 
   bool zoomWidth = !a_Ctrl;
   if (zoomWidth) {
+    m_MouseRatio = static_cast<double>(mousex) / getWidth();
     time_graph_.ZoomTime(delta, m_MouseRatio);
     m_WheelMomentum = delta * m_WheelMomentum < 0 ? 0 : m_WheelMomentum + delta;
   } else {
-    // TODO: reimplement vertical zoom by scaling track heights.
+    m_MouseRatio = static_cast<double>(mousey) / getHeight();
```
pierricgimmig, line 392: m_MouseRatio is being reused for different purposes here, we should probably have separate variables, mouse_x_ratio_ and mouse_y_ratio_ maybe. That being said, I don't think we need it for the vertical case.

OrbitGl/CaptureWindow.cpp
```diff
@@ -376,19 +376,23 @@ void CaptureWindow::MouseWheelMoved(int a_X, int a_Y, int a_Delta,
...
+    time_graph_.VerticalZoom(delta, m_MouseRatio);
+    // TODO: What is m_WheelMomentun about?
+    m_WheelMomentum = delta * m_WheelMomentum < 0 ? 0 : m_WheelMomentum + delta;
```
pierricgimmig, line 395: The wheel momentum code is not needed for vertical zoom, please remove.  It was added so that zoom would continue for a little while after the scroll wheel stopped moving. The effect was modulated by how fast the scroll wheel was moved. I don't think it works anymore actually.

OrbitGl/TimeGraph.cpp
```diff
@@ -189,6 +189,24 @@ void TimeGraph::ZoomTime(float a_ZoomValue, double a_MouseRatio) {
...
+  double scale = zoom_value > 0 ? 1 + increment_ratio : 1 / (1 + increment_ratio);
```
pierricgimmig, line 196: The vertical zoom is inverted, usually we zoom in by scrolling the wheel towards the top of the mouse (see ctrl-scroll behavior in Visual Studio for example).

OrbitGl/TimeGraph.cpp
```diff
@@ -189,6 +189,24 @@ void TimeGraph::ZoomTime(float a_ZoomValue, double a_MouseRatio) {
...
+
+  double world_height = m_Canvas->GetWorldHeight();
+  double y_mouse_position = m_Canvas->GetWorldTopLeftY() - mouse_ratio * world_height;
+  double top_distance = m_Canvas->GetWorldTopLeftY() - y_mouse_position;
+
+  // As we are scaling things, every item is going to scale its position.
+  double new_y_mouse_position = y_mouse_position * scale;
+  m_Canvas->SetWorldTopLeftY(new_y_mouse_position + top_distance);
```
pierricgimmig, line 204: Something is missing in the vertical zoom behavior, we should never see a big empty gap on top of the scheduler track.  This can happen easily when zooming out.  If we then use the slider bar, everything snaps back to the top.

OrbitGl/TimeGraphLayout.cpp
```diff
@@ -80,14 +81,24 @@ bool TimeGraphLayout::DrawProperties() {
   FLOAT_SLIDER_MIN_MAX(m_TextZ, -1.f, 1.f);
   FLOAT_SLIDER_MIN_MAX(m_TrackZ, -1.f, 1.f);
   FLOAT_SLIDER(m_ToolbarIconHeight);
+  FLOAT_SLIDER(scale_);
```
pierricgimmig, line 84: FLOAT_SLIDER_MIN_MAX(scale_, 0.f, 10.f);

OrbitGl/TimeGraphLayout.cpp
```diff
@@ -80,14 +81,24 @@ bool TimeGraphLayout::DrawProperties() {
...
   ImGui::End();
 
   return needs_redraw;
 }
 
+float TimeGraphLayout::GetRoundingRadius() const {
+  float max_radius = m_RoundingRadius * scale_;
+  // The radius has to be smaller than certain values to draw a proper circle
+  max_radius = std::min(max_radius, GetTrackTabHeight() / 3);
+  max_radius = std::min(max_radius, GetTrackTabWidth() / 3);
```
pierricgimmig, line 95: Why 3?

pierricgimmig: The rounding radius is already clamped to half of layout.GetTrackTabHeight() (see Track.cpp, around line 126).

pierricgimmig: Actually, I'd remove the clamping here, and add this to Track.cpp:

float half_label_width = 0.5f * label_width; <- add this
...
float radius = std::min(layout.GetRoundingRadius(), half_label_height);  <- this line already exists
radius = std::min(radius, half_label_width); <- add this

OrbitGl/CaptureWindow.cpp
pierricgimmig: Maybe we do need it to focus the zoom on the current mouse position, I'd still rather have two distinct variables.

OrbitGl/TimeGraph.cpp
pierricgimmig: It's actually panning that makes it snap back.

OrbitGl/TimeGraphLayout.cpp
pierricgimmig: This was just to show that we could specify a custom range.  Not at all critical... :-)

OrbitGl/CaptureWindow.cpp
karupayun: Yes, needed. But changing its name!

OrbitGl/CaptureWindow.cpp
karupayun: Perfect, thank you! 👍

OrbitGl/TimeGraph.cpp
karupayun: Anyway, both changes (+ some refactoring in horizontal zoom) will be in another PR.

OrbitGl/TimeGraph.cpp
karupayun: Yes, you are right. Sorry for that.

OrbitGl/TimeGraph.cpp
karupayun: I didn't know that. Thanks.

After playing a little bit with the layout variables and the slider bar, I realized that the maximum m_WorldTopLeftY is related with the SliderWidth: max (m_WorldTopLeftY) is exactly 1.5* m_SliderWidth.
Shall I maintain this maximum when zooming?

OrbitGl/TimeGraphLayout.cpp
karupayun: Thanks! I'm not being able to reproduce again situations where 0.5 (or divided by 2) is not enough. But I used to have. Maybe when moving to track we also solve a bug.

Anyway, about erasing the next line: 
max_radius = std::min(max_radius, GetEventTrackHeight());

Am I being too perfectionist when I'm being worried about this kind of situations? (the blue line is outside the rounding radius)
Screen Shot 2020-07-31 at 12 26 04

pierricgimmig: This is an issue with the z-ordering, the rounded corners should hide the lines.  The rounded corners are actually a mask, the problem here is that the lines are drawn on top of it.

OrbitGl/TimeGraph.cpp
```diff
@@ -196,6 +196,34 @@ void TimeGraph::ZoomTime(float a_ZoomValue, double a_MouseRatio) {
...
+void TimeGraph::VerticalZoom(float zoom_value, double mouse_relative_position) {
+  static double increment_ratio = 0.1;
+
+  double scale = zoom_value > 0 ? 1 + increment_ratio : 1 - increment_ratio;
+
+  float world_height = m_Canvas->GetWorldHeight();
+  double y_mouse_position = m_Canvas->GetWorldTopLeftY() - mouse_relative_position * world_height;
+  double top_distance = m_Canvas->GetWorldTopLeftY() - y_mouse_position;
+
+  // As we are scaling things, every item is going to scale its position.
```
pierricgimmig, line 208: This comment is very general and doesn't really help understanding the line below. Consider removing it altogether.

OrbitGl/TimeGraph.cpp
```diff
@@ -196,6 +196,34 @@ void TimeGraph::ZoomTime(float a_ZoomValue, double a_MouseRatio) {
...
+  new_world_top_left_y =
+      std::max(new_world_top_left_y, world_height - GetThreadTotalHeight());
+  // TODO: TopMargin has to be 1.5f * m_Layout.GetSliderWidth()?
+  new_world_top_left_y =
+      std::min(new_world_top_left_y, 1.5f * m_Layout.GetSliderWidth());
+
+  m_Canvas->SetWorldTopLeftY(new_world_top_left_y);
+
+  // Finally, we have to scale every item in the layout.
+  m_Layout.VerticalZoom(scale);
```
pierricgimmig, line 224: I'd use the GetScale/SetScale methods instead of introducing the concept of VerticalZoom to the layout.

PR #867, opened 2020-07-27
Added unittests for the BlockChain class

reichlfl:
Initially I added those because I though I needed to add functionality to the BlockChain - in the end I didn't modify BlockChain at all, but can't hurt to have the tests in there.

### DIFF
--- a/OrbitCore/BlockChainTest.cpp
+++ b/OrbitCore/BlockChainTest.cpp
@@ -1,0 +1,188 @@
+// Copyright (c) 2020 The Orbit Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#include <BlockChain.h>
+#include <gtest/gtest.h>
+
+#include <string>
+
+TEST(BlockChain, AddCopyableTypes) {
+  const std::string v1 = "hello world";
+  const std::string v2 = "or not";
+
+  BlockChain<std::string, 1024> chain;
+  EXPECT_EQ(chain.size(), 0);
+  chain.push_back(v1);
+  chain.push_back(v2);
+  EXPECT_EQ(chain.size(), 2);
+
+  // Multi-block test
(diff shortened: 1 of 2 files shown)


reichlfl:
Thanks for checking!

> Thanks for this Florian, one thing I've been wanting to test for a while is reading and writing from different threads under heavy contention. I have a feeling that this data structure is not 100% bulletproof but that we're getting away with it based on our particular usage. This is something we could add in another PR, but this is a great start, thanks!

Good point - I'm actually not sure what would be the best way to add a test for this, we want to avoid that tests are unstable and fail/pass randomly, do you have thoughts on how to do this?

> Also, I think that we might want to explore other data structures altogether, the goal being to have contiguous elements with no copy on resize, this virtual memory trick looks particularly promising: https://ourmachinery.com/post/virtual-memory-tricks/.

Sounds interesting. But at this point the BlockChain seems to work fine for me, do we have any performance issues or concerns (except the potential stability issues when multithreading) with the structure in general?

pierricgimmig:
> Good point - I'm actually not sure what would be the best way to add a test for this, we want to avoid that tests are unstable and fail/pass randomly, do you have thoughts on how to do this?

I was thinking adding known elements from one thread and reading them from another thread and validating that the read element is what we expect. We could have a time limit or a predefined number of elements to write/read. We could do this with elements of different sizes. I'd think we'd catch an issue pretty fast if the test were to fail so we wouldn't have random fails for very long. As I mentioned, this doesn't need to be part of this PR, so no worries there. 

> Sounds interesting. But at this point the BlockChain seems to work fine for me, do we have any performance issues or concerns (except the potential stability issues when multithreading) with the structure in general?

We don't have indexed lookup, we have that SlowAt() method that iterates on blocks in the chain. But you're right, if it ain't broken, don't fix it :-). Also having data in blocks could actually come in handy if we decide stream data to/from disk.

Issue #848, opened 2020-07-23
Standard build fails unless ATL is installed


danielo-unity3d:
I was getting the following linker error:
```
LINK : fatal error LNK1104: cannot open file 'atls.lib' [C:\Users\danielo\tools\orbit\build_default_relwithdebinfo\OrbitQ 
t\OrbitQt.vcxproj]
```
I did not have ATL or MFC components installed in VS 2019 (they are not installed by default).
So I installed them, but unfortunately, I (initially) installed the wrong versions. 

You must choose amongst multiple versions:
1. for build tools v141, or v142 (latest)
1. with and without Spectre mitigations
1. CPU architecture: ARM, ARM64, or "x86 and x64"
1. ATL versions: standard, v14.20, v14.21, v14.22, v14.23, v14.24, v14.25

In total, the installer gives you the choice of installing one or several versions from amongst 56 different versions of ATL, so it's easy to screw up.
I finally found the right one, but as I had added verbosity to the linker because I couldn't tell where the reference was coming from (in `OrbitQt/CMakeLists.txt`):
```
target_link_options(OrbitQt PRIVATE /ignore:4099 /VERBOSE:LIB)
```
Finally, I discovered where ATL was being referenced, in `OrbitCore/Utils.cpp`, where we find:
```
#include <AtlBase.h>
#include <atlconv.h>
```
Removing them reveals that they are required in order to satisfy the following external symbols:
```
OrbitCore\Utils.cpp(51,16): error C3861: 'StringFromCLSID': identifier not found 
OrbitCore\Utils.cpp(55,5): error C3861: 'CoTaskMemFree': identifier not found
```
These functions are found in `combaseapi.h` (as per the documentation).
It is therefore possible to avoid the dependency on ATL by `#include`-ing `combaseapi.h` instead of ATL headers.

Notes:
- The offending library, `atls.lib`, is pulled in when `#include`-ing `AtlBase.h` using `#pragma comment(lib, "alts.lib")`
- This can be avoided by `#define`-ing `_ATL_NO_DEFAULT_LIBS` just before `#include`-ing `AtlBase.h`

I guess, at this point, I should have just made a PR?

DanOlivier:
@googlebot I signed it!

DanOlivier:
@googlebot: I signed it!

pierricgimmig:
[...]

PR #870, opened 2020-07-27
Remove dependency on ATL

DanOlivier:
To resolve issue #848


DanOlivier:
@googlebot I fixed it.

PR #869, opened 2020-07-27
Remove dependency on ATL

danielo-unity3d:
To resolve issue #848

PR #871, opened 2020-07-27
Remove dependency on ATL

DanOlivier:
To resolve issue #848


DanOlivier:
Apologies, for two other PRs (#869 and #870), as the wrong identity (email was my corp email) was embedded within the commit (and I couldn't get past the CLA check).

PR #872, opened 2020-07-27
Clear iterator UI when clearing the capture, loading a new capture, or capturing a new capture

ronaldfw:
Using existing iterators in the UI does not make sense when the capture changes, which can happen due to clearing the capture, loading a capture, or when starting a new capture. In all these cases we reset the iterator UI.

### DIFF
--- a/OrbitQt/orbitlivefunctions.cpp
+++ b/OrbitQt/orbitlivefunctions.cpp
@@ -103,3 +103,14 @@ void OrbitLiveFunctions::AddIterator(size_t id, Function* function) {
 QLineEdit* OrbitLiveFunctions::GetFilterLineEdit() {
   return ui->data_view_panel_->GetFilterLineEdit();
 }
+
+void OrbitLiveFunctions::Reset() {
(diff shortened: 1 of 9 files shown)


OrbitQt/orbitlivefunctions.cpp
```diff
@@ -103,3 +103,14 @@ void OrbitLiveFunctions::AddIterator(size_t id, Function* function) {
 QLineEdit* OrbitLiveFunctions::GetFilterLineEdit() {
   return ui->data_view_panel_->GetFilterLineEdit();
 }
+
+void OrbitLiveFunctions::Reset() {
+  live_functions_.Reset();
+
+  for (auto& [_, iterator_ui] : iterator_uis) {
```
pierricgimmig, line 110: mind blown :-)

karupayun: I believe that its name is "Structured Binding", a C++17 feature that makes your daily life more confortable :)

PR #873, opened 2020-07-28
Tooltips for callstacks are shortened (both in depth and line length) if needed

reichlfl:
Following a comment from Ronald in the last PR, callstacks in tooltips are now shortened if needed.

### DIFF
--- a/OrbitCore/Utils.h
+++ b/OrbitCore/Utils.h
@@ -276,6 +276,25 @@ class CWindowsMessageToString {
 #endif
 
 //-----------------------------------------------------------------------------
+inline std::string ShortenStringWithEllipsis(const std::string& text, int max_len) {
+  if (max_len < 0) {
+    return text;
+  }
+  if (max_len <= 3) {
+    return text.length() <= 3 ? text : "...";
+  }
+  if (text.length() <= max_len) {
+    return text;
+  }
+
+  const size_t chars_to_cut =
+      text.length() - static_cast<uint32_t>(max_len) + 3;
+  const size_t l = (text.length() - chars_to_cut) / 2;
+  const size_t r = l + chars_to_cut;
+  return text.substr(0, l) + "..." + text.substr(r);
+}
+
+//-----------------------------------------------------------------------------
 inline std::string GetPrettySize(uint64_t size) {
   constexpr double KB = 1024.0;
   constexpr double MB = 1024.0 * KB;
--- a/OrbitCore/Utils.h
+++ b/OrbitCore/Utils.h
@@ -276,24 +276,32 @@ class CWindowsMessageToString {
 #endif
 
 //-----------------------------------------------------------------------------
-inline std::string ShortenStringWithEllipsis(const std::string& text,
+inline std::string ShortenStringWithEllipsis(std::string_view text,
                                              int max_len) {
   if (max_len < 0) {
-    return text;
+    return std::string(text);
   }
 
   size_t max_len_unsigned = static_cast<size_t>(max_len);
   if (max_len_unsigned <= 3) {
-    return text.length() <= 3 ? text : "...";
+    return text.length() <= 3 ? std::string(text) : "...";
   }
   if (text.length() <= max_len_unsigned) {
-    return text;
+    return std::string(text);
   }
 
   const size_t chars_to_cut = text.length() - max_len_unsigned + 3;
-  const size_t l = (text.length() - chars_to_cut) / 2;
+  size_t l = text.length() - chars_to_cut;
+  // Integer division by two, rounded up
+  if (l & 0x1) {
+    l = (l + 1) >> 1;
+  } else {
+    l = l >> 1;
+  }
+
   const size_t r = l + chars_to_cut;
-  return text.substr(0, l) + "..." + text.substr(r);
+  return std::string(text.substr(0, l)) + "..." +
+         std::string(text.substr(r));
 }
 
 //-----------------------------------------------------------------------------
--- a/OrbitCore/Utils.h
+++ b/OrbitCore/Utils.h
@@ -276,21 +276,21 @@ class CWindowsMessageToString {
 #endif
 
 //-----------------------------------------------------------------------------
-inline std::string ShortenStringWithEllipsis(std::string_view text,
-                                             int max_len) {
-  if (max_len < 0) {
+enum class EllipsisPosition { kMiddle };
+
+inline std::string ShortenStringWithEllipsis(
+    std::string_view text, size_t max_len,
+    EllipsisPosition pos = EllipsisPosition::kMiddle) {
+  constexpr const size_t kNumCharsEllipsis = 3;
+
+  if (max_len <= kNumCharsEllipsis) {
+    return text.length() <= kNumCharsEllipsis ? std::string(text) : "...";
+  }
+  if (text.length() <= max_len) {
     return std::string(text);
   }
 
-  size_t max_len_unsigned = static_cast<size_t>(max_len);
-  if (max_len_unsigned <= 3) {
-    return text.length() <= 3 ? std::string(text) : "...";
-  }
-  if (text.length() <= max_len_unsigned) {
-    return std::string(text);
-  }
-
-  const size_t chars_to_cut = text.length() - max_len_unsigned + 3;
+  const size_t chars_to_cut = text.length() - max_len + kNumCharsEllipsis;
   size_t l = text.length() - chars_to_cut;
   // Integer division by two, rounded up
   if (l & 0x1) {
@@ -300,8 +300,7 @@ inline std::string ShortenStringWithEllipsis(std::string_view text,
   }
 
   const size_t r = l + chars_to_cut;
-  return std::string(text.substr(0, l)) + "..." +
-         std::string(text.substr(r));
+  return std::string(text.substr(0, l)) + "..." + std::string(text.substr(r));
 }
 
 //-----------------------------------------------------------------------------
--- a/OrbitCore/Utils.h
+++ b/OrbitCore/Utils.h
@@ -276,19 +276,21 @@ class CWindowsMessageToString {
 #endif
 
 //-----------------------------------------------------------------------------
-inline std::string ShortenStringWithEllipsis(const std::string& text, int max_len) {
+inline std::string ShortenStringWithEllipsis(const std::string& text,
+                                             int max_len) {
   if (max_len < 0) {
     return text;
   }
-  if (max_len <= 3) {
+
+  size_t max_len_unsigned = static_cast<size_t>(max_len);
+  if (max_len_unsigned <= 3) {
     return text.length() <= 3 ? text : "...";
   }
-  if (text.length() <= max_len) {
+  if (text.length() <= max_len_unsigned) {
     return text;
   }
 
-  const size_t chars_to_cut =
-      text.length() - static_cast<uint32_t>(max_len) + 3;
+  const size_t chars_to_cut = text.length() - max_len_unsigned + 3;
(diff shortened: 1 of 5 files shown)


OrbitCore/Utils.h
```diff
@@ -275,6 +275,25 @@ class CWindowsMessageToString {
...
+  }
+  if (text.length() <= max_len) {
+    return text;
+  }
+
+  const size_t chars_to_cut =
+      text.length() - static_cast<uint32_t>(max_len) + 3;
+  const size_t l = (text.length() - chars_to_cut) / 2;
+  const size_t r = l + chars_to_cut;
+  return text.substr(0, l) + "..." + text.substr(r);
```
dpallotti, line 294: Thought: would the ellipsis "&mldr;" character (U+2026) be supported? You would save two characters.

OrbitCore/UtilsTest.cpp
```diff
@@ -0,0 +1,17 @@
+#include <gtest/gtest.h>
+
+#include "Utils.h"
+
+TEST(Utils, TestEllipsis) {
+  EXPECT_EQ(ShortenStringWithEllipsis("17 char long text", 18), "17 char long text");
+  EXPECT_EQ(ShortenStringWithEllipsis("17 char long text", 17), "17 char long text");
+  EXPECT_EQ(ShortenStringWithEllipsis("17 char long text", 7), "17...xt");
+  EXPECT_EQ(ShortenStringWithEllipsis("17 char long text", 6), "1...xt");
```
dpallotti, line 9: Do we want the ellipsis in the middle? I find it often undesirable as it takes away characters from the prefix and gives them to a suffix that is unreadable anyway after the ellipsis has been added.
Either way, in cases like this line I think that the tie should be broken in favor of the prefix, and the result should for example be "17...t".

OrbitGl/EventTrack.cpp
```diff
@@ -187,15 +187,53 @@ bool EventTrack::IsEmpty() const {
...
+  fn_name = Replace(fn_name, "&", "&amp;");
+  fn_name = Replace(fn_name, "<", "&lt;");
+  fn_name = Replace(fn_name, ">", "&gt;");
+  return fn_name;
+};
+
+//-----------------------------------------------------------------------------
+std::string FormatCallstackForTooltip(std::shared_ptr<CallStack> callstack,
+                                      int max_line_length = 80,
+                                      int max_lines = 13) {
```
dpallotti, line 207: This seems to be quite a limitation. What if the user prefers to see the entire callstack, even if that might take a lot of screen space?

OrbitCore/Utils.h
```diff
@@ -275,6 +275,25 @@ class CWindowsMessageToString {
 };
 #endif
 
+//-----------------------------------------------------------------------------
+inline std::string ShortenStringWithEllipsis(const std::string& text, int max_len) {
```
beckerhe, line 279: I think the argument `text` can be of type `std::string_view`.

OrbitGl/EventTrack.cpp
```diff
@@ -187,15 +187,53 @@ bool EventTrack::IsEmpty() const {
...
 
 //-----------------------------------------------------------------------------
-std::string EventTrack::GetSampleTooltip(PickingID id) const {
-  auto SafeGetFormattedFunctionName = [](uint64_t addr) -> std::string {
-    auto it = Capture::GAddressToFunctionName.find(addr);
-    return it == Capture::GAddressToFunctionName.end()
-               ? std::string("<i>") + "???" + "</i>"
-               : it->second;
-  };
+std::string SafeGetFormattedFunctionName(uint64_t addr, int max_line_length) {
```
beckerhe, line 190: This function is not declared in any header file, is it? In that case it should be static (or in an anonymous namespace).

OrbitGl/EventTrack.cpp
```diff
@@ -187,15 +187,53 @@ bool EventTrack::IsEmpty() const {
...
+  std::string fn_name = ShortenStringWithEllipsis(it->second, max_line_length);
+  // Simple HTML escaping
+  fn_name = Replace(fn_name, "&", "&amp;");
+  fn_name = Replace(fn_name, "<", "&lt;");
+  fn_name = Replace(fn_name, ">", "&gt;");
+  return fn_name;
+};
+
+//-----------------------------------------------------------------------------
+std::string FormatCallstackForTooltip(std::shared_ptr<CallStack> callstack,
```
beckerhe, line 205: The same as `SafeGetFormattedFunctionName`: I think this can be declared static.

OrbitCore/Utils.h
beckerhe, line 294: If it's the intended behaviour to keep the rest of the string after the ellipsis, the name of the function will feel quite misleading to me. If I saw a function called `ShortenStringWithEllipsis` with a max_len parameter I would assume I get a string with `max_len` characters. But maybe I'm wrong and this is common practice in text processing?

OrbitCore/Utils.h
```diff
@@ -275,6 +275,25 @@ class CWindowsMessageToString {
...
+  }
+  if (max_len <= 3) {
+    return text.length() <= 3 ? text : "...";
+  }
+  if (text.length() <= max_len) {
+    return text;
+  }
+
+  const size_t chars_to_cut =
+      text.length() - static_cast<uint32_t>(max_len) + 3;
```
beckerhe, line 291: Nit: If I'm not mistaken this cast is not needed here, but I guess the `if` before wants one.

OrbitCore/Utils.h
reichlfl: Hm maybe. But it's not so much about the actual text length but rather the size of the text on screen, so we wouldn't really benefit from this? In the end it's a bit heuristically anyways since we're not using a fixed-size font and the length in chars is just an indication and not a real measurement here.

OrbitCore/UtilsTest.cpp
reichlfl: I tried to be consistent with the way QT is doing it in e.g. the tables. But you're right, the tie should prefer the prefix (which is also the case in QT), I'll fix that.

OrbitGl/EventTrack.cpp
reichlfl: You always have the option to select the sample and see the full callstack in the "selection" tab. TBH, probably the same limitation may be too much for some devs and too little for others...
Alternatively, we could start introducing UI settings for things like those, but it seems a bit of an overkill to add this for a relatively small feature like this?

OrbitCore/Utils.h
beckerhe: Ahhh. I'm sorry. I completely misread what it's doing. Please ignore this message. :see_no_evil:

OrbitCore/UtilsTest.cpp
reichlfl: Changed the behavior to cut the postfix before the prefix

OrbitCore/Utils.h
reichlfl: Changed

OrbitGl/EventTrack.cpp
reichlfl: Done

OrbitCore/Utils.h
reichlfl: I actually changed the overall casts now, casting everything to unsigned after checking for >= 0 in the beginning. The kokoro build was complaining anyways.

OrbitCore/Utils.h
dpallotti: Yeah, I was thinking about monospace fonts, I guess that otherwise it doesn't matter.

OrbitCore/Utils.h
```diff
@@ -275,6 +275,35 @@ class CWindowsMessageToString {
...
 
+//-----------------------------------------------------------------------------
+inline std::string ShortenStringWithEllipsis(std::string_view text,
+                                             int max_len) {
+  if (max_len < 0) {
+    return std::string(text);
+  }
+
+  size_t max_len_unsigned = static_cast<size_t>(max_len);
+  if (max_len_unsigned <= 3) {
```
ronaldfw, line 286: Maybe define 3 as a constant. The name can indicate how it is used:

constexpr const size_t kNumCharsEllipsis = 3;

OrbitGl/EventTrack.cpp
ronaldfw: Some limitation is obviously needed. As @reichlfl points out, users can select the sample and see the callstack then on the right. Since this is mentioned in the tooltip, it should be somewhat discoverable / understandable for users and I think that's good enough for now. Eventually, we can add a button ("More", "See entire callstack",  "Select this sample") to the tooltip (though I don't know how much work that would be). In any case, we should aim to solve this directly in the UI, rather than introduce settings that may themselves be confusing for users.

dpallotti: I know that samples can be selected and seen in the "selection" tab, but that's *very* tedious to see a single callstack and one reason for the tooltips. But of course a UI setting for this is definitely overkill. 
So let me say it differently: 13 frames (or 12?) seem to be quite on the small side. It basically means the last six as the frames that get hidden are the ones in the middle. From what I've seen such a number of lines still produces quite a small tooltip. I would prefer to see more information (with some limitation obviously) than a small cute tooltip that can't be enlarged.
(Of course this is very easy to change in the future but again it's one of those things that once there don't really get changed anymore.)

ronaldfw: I was just pointing out that some limitation is always needed (above you were talking about users wanting to see the entire callstack) and that there is a workaround to see the full callstack. Agree that we may want to increase 13 to some larger number, maybe just try 25 to double the number of frames visible? 

We might also want to think about doing this asymmetrically where users can see more frames at the top of the stack (the side including the current function). Though not always true, that seems to be in general more relevant. WDYT?

But in general, you will always have a case where someone wants/needs to see the entire stack. We just want to make that case as rare as possible while maintaining good usability of the tooltips.

OrbitCore/Utils.h
```diff
@@ -275,6 +275,35 @@ class CWindowsMessageToString {
 };
 #endif
 
+//-----------------------------------------------------------------------------
+inline std::string ShortenStringWithEllipsis(std::string_view text,
```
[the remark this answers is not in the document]
reichlfl: I was thinking the same this morning, and @beckerhe was also mislead by the function name. I like the idea of an additional enum, so we could extend it later, I added that.

OrbitCore/Utils.h
```diff
@@ -275,6 +275,35 @@ class CWindowsMessageToString {
...
+                                             int max_len) {
+  if (max_len < 0) {
```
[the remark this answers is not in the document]
reichlfl: I did, but this requirement came from a call in EventTrack and shouldn't reflect in the design of this utility method. I agree that this is confusing. Changed the parameter to size_t now, this also removes some cast in the function and improves it overall. Thanks!

OrbitGl/EventTrack.cpp
dpallotti: @ronaldfw I also discussed this offline with @reichlfl and I think that we all agree on what you said now.

reichlfl: I settled for 20 now, this looks ok on my smaller (1080p) screen - 13 may have been a bit short, true.
@ronaldfw Actually it already is asymmetrical, exactly as you suggest: When the stack is shortened, it only shows a fixed number of lines from the bottom, then "...", then as much as possible from the top. I added the number of lines to display from the bottom as a parameter to make it clearer and more verbose.

ronaldfw: Sounds good!

PR #876, opened 2020-07-29
Maximize screen on startup properly.

danielfenner:
Orbit currently starts with a maximized window but the actual size of the window only covers a small part of the screen.

This workaround is described here:
https://stackoverflow.com/questions/27157312/qt-showmaximized-not-working-in-windows


OrbitQt/main.cpp
```diff
@@ -127,6 +127,7 @@ static outcome::result<void> RunUiInstance(
       absl::StrFormat("127.0.0.1:%d", ports.grpc_port);
 
   OrbitMainWindow w(app, std::move(options));
+  w.resize(1280, 720);
```
beckerhe, line 130: Can you add a comment? Otherwise I would assume somebody will remove this sooner than later.

danielfenner: Good point. Done.

PR #877, opened 2020-07-29
Bugfixes for GPU Track collapse behavior

reichlfl:
* Redraw forced after clearing the timeline
* Initial state of TriangleToggle can be set after construction (through an additional parameter in setState)
* GPUTrack does no longer check for depth_ > 1 to determine if the track is collapsed. The depth_ is updated in the loop after this check, which leads to inconsistent drawing after a capture is loaded

### DIFF
--- a/OrbitGl/TimeGraph.cpp
+++ b/OrbitGl/TimeGraph.cpp
@@ -95,9 +95,9 @@ void TimeGraph::Clear() {
   // The process track is a special ThreadTrack of id "0".
   process_track_ = GetOrCreateThreadTrack(0);
 
-  NeedsUpdate();
-
   SetOverlayTextBoxes({});
+
+  NeedsUpdate();
 }
 
 //-----------------------------------------------------------------------------
--- a/OrbitGl/TimeGraph.cpp
+++ b/OrbitGl/TimeGraph.cpp
@@ -95,6 +95,8 @@ void TimeGraph::Clear() {
   // The process track is a special ThreadTrack of id "0".
   process_track_ = GetOrCreateThreadTrack(0);
 
+  NeedsUpdate();
+
   SetOverlayTextBoxes({});
 }
 
@@ -339,6 +341,8 @@ void TimeGraph::ProcessTimer(const Timer& a_Timer) {
       cores_seen_.insert(a_Timer.m_Processor);
     }
   }
+
+  NeedsUpdate();
 }
 
 //-----------------------------------------------------------------------------
--- a/OrbitGl/TriangleToggle.cpp
+++ b/OrbitGl/TriangleToggle.cpp
@@ -68,9 +68,9 @@ void TriangleToggle::OnRelease() {
   time_graph_->NeedsUpdate();
 }
 
-void TriangleToggle::SetState(State state, bool replace_initial_state) {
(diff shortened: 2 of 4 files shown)


OrbitGl/TimeGraph.cpp
```diff
@@ -95,6 +95,8 @@ void TimeGraph::Clear() {
   // The process track is a special ThreadTrack of id "0".
   process_track_ = GetOrCreateThreadTrack(0);
 
+  NeedsUpdate();
```
ronaldfw, line 98: Move this to the end of the function?

OrbitGl/TriangleToggle.cpp
```diff
@@ -68,4 +68,9 @@ void TriangleToggle::OnRelease() {
   time_graph_->NeedsUpdate();
 }
 
-void TriangleToggle::SetState(State state) { state_ = state; }
+void TriangleToggle::SetState(State state, bool replace_initial_state) {
```
ronaldfw, line 71: It can be hard to understand booleans at callsites. When reading

toggle.SetState(TriangleToggle::State::kCollapsed, true); 

I may understand this as setting the collapsed state to true. 

I would suggest to add a method ReplaceInitialState (or just SetInitialState, but the replace makes it more clear that this a special operation), so you can (at the callsite) just say:

togger.SetState(TriangleToggle::State::kCollapsed);
togger.ReplaceInitialState(TriangleToggle::State::kCollapsed);

WDYT?

reichlfl: I'd generally agree. The reason I added it like this here: It took me a bit to figure out why the state, that was set correctly in the GPUTrack constructor, was suddenly overwritten in a different part of the code and I couldn't find any matching "SetState()" calls. As a user of the class, the second parameter makes it clear to me that there is something else going on that may influence the state.

Thoughts?

reichlfl:
> As discussed over chat, I think we should also consider unifying the UI so that even a one depth track can be collapsed and expanded. That's a little weird, but in particular for the GPU tracks makes the visualization more consistent (collapsed --> only see GPU execution; expanded --> see all events).

After testing this, we both found it a bit confusing to the user and will stick with the old behavior of disabling the collapse-icon when only a single track is present.

OrbitGl/TriangleToggle.cpp
ronaldfw: If you want to keep the SetState call, then I would suggest one of two options:
1) Make the second parameter an enum that provides the different options you have: kKeepInitialState, kReplaceInitialState. For true/false options, this feels a bit like overkill, but is much more readable IMO.
2) At callsites, add a comment what the second parameter represents: 
SetState(..., /* replace_initial_state=*/ true);
This is as readable as the first option and a bit shorter (you don't need to define the enum), but can be easy to forget when writing new code that uses SetState.

OrbitGl/TimeGraph.cpp
reichlfl: Done

OrbitGl/TriangleToggle.cpp
reichlfl: Replaced it with an enum now. I think it's a bit overkill here, but I'm sure the second option would get lost over time when new calls are added, so the first at least will provide permanently improved readabiliy.

PR #880, opened 2020-07-30
Add a top-down view

dpallotti:
This adds a first iteration of the top-down view.

Screenshot from profiling Trata:
topdown-screenshot

Things left to do:
- Add context menus&mdash;this is quite pressing.
- Add a checkbox to switch between grouping by thread and not. For now I went for adding a top-level "all threads" branch next to the other threads (see the screenshot).
- Maybe improve the visualization of percentages, for example with "progress bars" filled to the amount corresponding to the percentage (at least for the "inclusive" column).
- Add a top-down view for the selection&mdash;it might be quite easy.
- Eventually, generalize to other tree views like bottom-up and butterfly.

---

#### Remove obsolete ShowIncludesDialog
Bug: http://b/158950531
#### Move live, sampling, selection tabs back to the .ui xml
This simplifies the code and avoids having all tabs defined in the xml before
all tabs added programmatically.
#### Add the top-down view
- Set up the new top-down tab in the ui.
- Set up the `OrbitMainWindow`-`OrbitApp` callback mechanism for the top-down
  view.
- Implement `TopDownView` with the new ad-hoc Qt `TopDownViewItemModel`.
#### Add percentage "Of parent" column to top-down view


ronaldfw:
I won't be able to review this before next week, but wanted to quickly say this looks really awesome, thanks!

OrbitQt/TopDownViewItemModel.h
```diff
@@ -0,0 +1,45 @@
...
+
+#include "TopDownView.h"
+
+class TopDownViewItemModel : public QAbstractItemModel {
+  Q_OBJECT
+
+ public:
+  explicit TopDownViewItemModel(std::unique_ptr<TopDownView> top_down_view,
+                                QObject* parent = nullptr);
+  ~TopDownViewItemModel() override = default;
```
beckerhe, line 21: I think you don't need this line.

OrbitGl/TopDownView.cpp
```diff
@@ -0,0 +1,102 @@
...
+      for (auto frame_it = resolved_callstack->m_Data.crbegin();
+           frame_it != resolved_callstack->m_Data.crend(); ++frame_it) {
+        uint64_t frame = *frame_it;
+        TopDownFunction* function_node =
+            current_thread_or_function->GetFunctionOrNull(frame);
+        if (function_node == nullptr) {
+          std::string function_name;
+          auto function_name_it = function_names.find(frame);
+          if (function_name_it != function_names.end() &&
+              function_name_it->second != "???") {
```
ronaldfw, line 88: Nit: Add a constant for the "???" string.

OrbitGl/TopDownView.cpp
```diff
@@ -0,0 +1,102 @@
...
+    for (const auto& callstack_id_and_count :
+         thread_sample_data->m_CallstackCount) {
+      std::shared_ptr<CallStack> resolved_callstack =
+          sampling_profiler.GetResolvedCallstack(callstack_id_and_count.first);
+      if (resolved_callstack == nullptr) {
+        continue;
+      }
+      const uint64_t sample_count = callstack_id_and_count.second;
+      // Don't count samples from the all-thread case (tid == 0) again.
+      if (tid != 0) {
```
ronaldfw, line 73: Nit: Add a constant for the 0 process.

OrbitGl/TopDownView.cpp
```diff
@@ -0,0 +1,102 @@
...
+  auto top_down_view = std::make_unique<TopDownView>();
+  for (const ThreadSampleData* thread_sample_data :
+       sampling_profiler.GetThreadSampleData()) {
+    const int32_t tid = thread_sample_data->m_TID;
+    TopDownThread* thread_node = top_down_view->GetThreadOrNull(tid);
+    if (thread_node == nullptr) {
+      std::string thread_name;
+      // Use the node with tid == 0 as the container for all threads, like in
+      // SamplingProfiler, SamplingReport(DataView), TimeGraph.
+      if (tid == 0) {
```
ronaldfw, line 53: Nit: Add a constant for the 0 process.

OrbitGl/TopDownView.cpp
```diff
@@ -0,0 +1,102 @@
...
+        thread_name = process_name;
+      } else {
+        auto thread_name_it = thread_names.find(tid);
```
ronaldfw, line 56: You could write this as: 

} else if (auto thread_name_it = thread_names.find(tid); thread_name_it != thread_names.end()) {
  thread_name = thread_name_it->second;
}

I find that a bit more readable because scopes are clearer.

OrbitGl/TopDownView.cpp
```diff
@@ -0,0 +1,102 @@
...
+        continue;
+      }
+      const uint64_t sample_count = callstack_id_and_count.second;
+      // Don't count samples from the all-thread case (tid == 0) again.
+      if (tid != 0) {
+        top_down_view->IncreaseSampleCount(sample_count);
+      }
+      thread_node->IncreaseSampleCount(sample_count);
+
+      TopDownInternalNode* current_thread_or_function = thread_node;
```
ronaldfw, line 78: Can some parts of this function (like maybe below here) be extracted as a function?

OrbitQt/TopDownViewItemModel.cpp
```diff
@@ -0,0 +1,209 @@
...
+QVariant TopDownViewItemModel::data(const QModelIndex& index, int role) const {
+  if (!index.isValid()) {
+    return QVariant();
+  }
+  switch (role) {
+    case Qt::DisplayRole: {
+      auto* item = static_cast<TopDownNode*>(index.internalPointer());
+      auto thread_item = dynamic_cast<TopDownThread*>(item);
+      auto function_item = dynamic_cast<TopDownFunction*>(item);
+      if (thread_item != nullptr) {
```
ronaldfw, line 20: Indentation is getting pretty deep here. Can you refactor this to move some parts out into functions?

ronaldfw:
Some meta-questions that we should discuss but that are not necessarily relevant for this PR: 
- Is the top-down view affected by selections? Should it be?
- Minor: Should the top-down view be next to sampling?

OrbitQt/TopDownViewItemModel.h
dpallotti: True, thanks.

dpallotti:
Before I get to addressing the review of the code, let me reply to the comments.

@ronaldfw 
> Minor: Should the top-down view be next to sampling?

You mean as in the order of the tabs? In that case, maybe, I haven't really given it a thought, I just added the tab at the end. Do you think it should be between the "sampling" and the "selection" tab?

> Is the top-down view affected by selections? Should it be?

Is it? No. Should it be? I think so. But we should discuss how. 
Should it be in the same tab (basically replacing the "overall" top-down view) or in a different tab?\
If in a different tab, I fear the explosion of the number of tabs&mdash;we already have "sampling"/"selection", we would add "top-down"/"top-down (selection)".\
If in the same, I think that "sampling"/"selection" should also be joined. But Pierric noted that the user could have set up one view (e.g., opened the top-down tree) in a certain way, and then a selection would clear that setup. One way to solve this might be to just retain that setup, e.g. by just hiding such "overall" views while a selection is active.

@pierricgimmig 

> Filtering would be important, not just for consistency with all other data views, but it would be particularly useful to automatically expand/collapse the nodes to quickly see the hierarchies containing the searched function(s).

> "Collapse/Expand all" functionality would be great, globally but also for the selected node.

> We should support copy/pasting like all the other data views in Orbit. This would possibly mean allowing multiple selection.

I agree with all this, I tried to add some things left to do in the PR's description but I forgot many. "Collapse/Expand all" is actually very important, I was already annoyed by not having it while implementing this. But this is just a starting point.

> To me, this should probably be a view mode in the Sampling and Selection tabs. Since it is another way to look at
the same sampling data, I think it might make sense to have the option to switch from "flat" to "top-down" to "butterfly" or "bottom-up" within the original sampling and selection tabs instead of creating a new tab for each type.

I remember we discussed this and we need to experiment and change things. My feeling so far, though, has been that the top-down view needs a lot of vertical space. If it "replaced" (based on a ui switch) the flat sampling data, it might still be ok. Still, the "callstacks" view at the bottom takes away space, and maybe while you're in the top-down view you don't really need to see the callstacks when you click on an element, because you're basically already seeing them in tree form.\
To add to this, I'm also not sure if it's best to have all the flat, top-down, bottom-up, butterfly in the same place under just a switch&mdash;it would feel a bit too cramped to me, but this is minor.\
Lots of things to consider...

OrbitGl/TopDownView.h
```diff
@@ -0,0 +1,166 @@
...
+  TopDownNode* parent_;
+};
+
+class TopDownFunction;
+
+class TopDownInternalNode : public TopDownNode {
+ public:
+  explicit TopDownInternalNode(TopDownNode* parent) : TopDownNode{parent} {}
+
+  TopDownFunction* GetFunctionOrNull(uint64_t function_absolute_address);
```
dimitry-, line 42: const?

OrbitGl/TopDownView.h
```diff
@@ -0,0 +1,166 @@
...
+
+  TopDownFunction* AddAndGetFunction(uint64_t function_absolute_address,
+                                     std::string function_name);
+
+ protected:
+  std::map<uint64_t, TopDownFunction> function_nodes_;
```
dimitry-, line 56: flat_hash_map?

OrbitGl/TopDownView.h
```diff
@@ -0,0 +1,166 @@
...
+    return function_absolute_address_;
+  }
+
+  const std::string& function_name() const { return function_name_; }
+
+  void IncreaseSampleCount(uint64_t sample_count_increase) {
+    inclusive_sample_count_ += sample_count_increase;
+  }
+
+  uint64_t inclusive_sample_count() const { return inclusive_sample_count_; }
```
dimitry-: nodiscard? here and below.

OrbitGl/TopDownView.cpp
```diff
@@ -0,0 +1,97 @@
...
+
+    TopDownThread* thread_node = top_down_view->GetThreadOrNull(tid);
+    if (thread_node == nullptr) {
+      std::string thread_name;
+      auto thread_name_it = thread_names.find(tid);
+      if (thread_name_it != thread_names.end()) {
+        thread_name = thread_name_it->second;
+      }
+      thread_node = top_down_view->AddAndGetThread(tid, std::move(thread_name));
+    }
```
dimitry-: extract to GetOrCreateThreadNode?

OrbitGl/TopDownView.cpp
```diff
@@ -0,0 +1,97 @@
...
+          auto function_name_it = function_names.find(frame);
+          if (function_name_it != function_names.end() &&
+              function_name_it->second != "???") {
+            function_name = function_name_it->second;
+          } else {
+            function_name = absl::StrFormat("[unknown@%#llx]", frame);
+          }
+          function_node = current_thread_or_function->AddAndGetFunction(
+              frame, std::move(function_name));
+        }
```
dimitry-, line 95: Also extract GetOrCreateFunctionNode?

dpallotti: Done.

OrbitGl/TopDownView.cpp
dpallotti: Extracted `GetOrCreateThreadNode`, `AddCallstackToTopDownThread`, `GetOrCreateFunctionNode`. Is it better now? To me it looks more convoluted now because of having to forward all the parameters. Maybe it's because I already have the logic in mind having written the code?

OrbitGl/TopDownView.cpp
dpallotti: Done.

OrbitCore/SamplingProfiler.cpp
```diff
@@ -84,6 +84,23 @@ void SamplingProfiler::AddUniqueCallStack(CallStack& a_CallStack) {
...
+  if (resolved_callstack_id_it ==
+      m_OriginalCallstackToResolvedCallstack.end()) {
+    return nullptr;
+  }
+
+  auto resolved_callstack_it =
+      m_UniqueResolvedCallstacks.find(resolved_callstack_id_it->second);
+  if (resolved_callstack_it == m_UniqueResolvedCallstacks.end()) {
+    return nullptr;
+  }
```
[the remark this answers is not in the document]
dpallotti: Done.

pierricgimmig:
> I remember we discussed this and we need to experiment and change things. My feeling so far, though, has been that the top-down view needs a lot of vertical space. If it "replaced" (based on a ui switch) the flat sampling data, it might still be ok. Still, the "callstacks" view at the bottom takes away space, and maybe while you're in the top-down view you don't really need to see the callstacks when you click on an element, because you're basically already seeing them in tree form.
> To add to this, I'm also not sure if it's best to have all the flat, top-down, bottom-up, butterfly in the same place under just a switch—it would feel a bit too cramped to me, but this is minor.
> Lots of things to consider...

Maybe having tabs inside the "sampling" tab would work? We could replicate the same tab hierarchy in the "selection" tab instead of having a lot of top level tabs. I agree that we wouldn't necessarily need to see the callstacks in the top-down view btw. As I said, I'm happy with submitting this and iterating.

OrbitCore/SamplingProfiler.cpp
```diff
@@ -122,13 +117,15 @@ SamplingProfiler::GetSortedCallstacksFromAddress(uint64_t a_Addr,
...
+const int32_t SamplingProfiler::kAllThreadsFakeTid = 0;
+
 //-----------------------------------------------------------------------------
 void SamplingProfiler::SortByThreadUsage() {
   m_SortedThreadSampleData.clear();
   m_SortedThreadSampleData.reserve(m_ThreadSampleData.size());
 
   // "All"
-  m_ThreadSampleData[0].m_AverageThreadUsage = 100.f;
+  m_ThreadSampleData[kAllThreadsFakeTid].m_AverageThreadUsage = 100.f;
```
dimitry-, line 128: well, it is not fake is it.. Just kAllThreadsTid?

OrbitGl/TopDownView.cpp
```diff
@@ -37,60 +37,83 @@ TopDownThread* TopDownView::AddAndGetThread(int32_t thread_id,
...
+       frame_it != resolved_callstack.m_Data.crend(); ++frame_it) {
+    uint64_t frame = *frame_it;
+    TopDownFunction* function_node = GetOrCreateFunctionNode(
+        current_thread_or_function, frame, function_names);
+    function_node->IncreaseSampleCount(callstack_sample_count);
+    current_thread_or_function = function_node;
+  }
+}
+
+static TopDownThread* GetOrCreateThreadNode(
```
dimitry-, line 76: Should this be a member of TopDownThread? same above

OrbitGl/TopDownView.h
```diff
@@ -0,0 +1,166 @@
...
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#ifndef ORBIT_GL_TOP_DOWN_VIEW_H_
+#define ORBIT_GL_TOP_DOWN_VIEW_H_
+
+#include "SamplingProfiler.h"
+#include "absl/container/flat_hash_map.h"
+
+class TopDownNode {
```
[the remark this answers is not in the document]
dimitry-: yep, actually if you do it on class is pointer also nodiscard? probably no?

OrbitGl/TopDownView.h
dimitry-: node_hash_map then? also probably add a comment about pointer stability.

OrbitGl/TopDownView.cpp
ronaldfw: I don't have a better suggestion and I don't think it's that bad. It clearly tells the reader the purpose of the 0 value.

OrbitGl/TopDownView.cpp
ronaldfw: For me, having the names of the functions that explain what is happening helps when first reading this, so I think it is better. Having the parameters also helps me understand what is going on, as it is more cleanly scoped which parameters are needed where.

OrbitQt/TopDownViewItemModel.cpp
ronaldfw: Thanks, looks good.

ronaldfw:
> You mean as in the order of the tabs? In that case, maybe, I haven't really given it a thought, I just added the tab at the end. Do you think it should be between the "sampling" and the "selection" tab?

Yes, I think it should be between "sampling" and "selection", as it visualizes the same data as the sampling tab (compared with the selection tab, which depending on the selection shows just a subset). That would feel a bit more consistent to me. As you and @pierricgimmig noted, there are many other things to consider, so this might all change. But as the order is a tiny change, I would propose to still do it now.

OrbitGl/TopDownView.cpp
dpallotti: Perfect then!

OrbitGl/TopDownView.h
dpallotti: Done.

dpallotti:
> > You mean as in the order of the tabs? In that case, maybe, I haven't really given it a thought, I just added the tab at the end. Do you think it should be between the "sampling" and the "selection" tab?
> 
> Yes, I think it should be between "sampling" and "selection", as it visualizes the same data as the sampling tab (compared with the selection tab, which depending on the selection shows just a subset). That would feel a bit more consistent to me. As you and @pierricgimmig noted, there are many other things to consider, so this might all change. But as the order is a tiny change, I would propose to still do it now.

It's fine by me, done.

dpallotti:
> As I said, I'm happy with submitting this and iterating.

I know I know, but as we're talking about it...

> Maybe having tabs inside the "sampling" tab would work? We could replicate the same tab hierarchy in the "selection" tab instead of having a lot of top level tabs. I agree that we wouldn't necessarily need to see the callstacks in the top-down view btw. 

Maybe, we would need a prototype. My feeling right now is that having even more subtabs (we already have the threads' subtabs) might be a bit cramped and confusing.

PR #881, opened 2020-07-30
Implement SFTP copy from remote to local

dimitry-:
This change renames SftpOperation to SftpCopyToRemoteOperation and implements SfptCopyToLocalOperation, also fixed crash in integration test.


dimitry-:
actually there is a bug - investigating.

dimitry-:
ok, now it works as expected. ptal

OrbitSshQt/IntegrationTests.cpp
```diff
@@ -173,50 +174,121 @@ TEST(OrbitSshQtTests, IntegrationTest) {
...
 
   session.ConnectToServer(creds);
   LOG("connect to server");
-  QTimer::singleShot(std::chrono::seconds{5}, [&]() {
-    loop.quit();
-    FAIL()
-        << "Timeout occurred. The whole integration test should be done in 5 "
-           "seconds. If not, probably it's stuck somewhere in the callback "
-           "logic.";
+  QTimer::singleShot(std::chrono::seconds{5}, [loop]() {
```
beckerhe, line 209: The bug you mentioned could have also been fixed by using another overload of QTimer::singleShot where you specify a receiving object: `QTimer::singleShot(std::chrono::seconds{5}, &loop, [&]() {...`

Then the lambda will only be executed when loop still exists.

OrbitSshQt/include/OrbitSshQt/SftpCopyToRemoteOperation.h
```diff
@@ -2,8 +2,8 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
 
-#ifndef ORBIT_SSH_QT_SFTP_OPERATION_H_
-#define ORBIT_SSH_QT_SFTP_OPERATION_H_
+#ifndef ORBIT_SSH_QT_SFTP_COPY_TO_REMOTE_OPERATION_H_
```
beckerhe, line 5: What about moving these changes into the first commit?

dimitry-: done.

OrbitSshQt/IntegrationTests.cpp
dimitry-: done. ptal

beckerhe: Thanks. Looks good!

PR #882, opened 2020-07-30
Refactor of tracks by introducing a new base class TimerTrack

ronaldfw:
This is a refactor for all track classes that are using timers (Timer class) as their main primitive to display. Distinct behavior between the classes is defined using virtual functions. Performance difference is neligible: Measurements for UpdatePrimitives ranged from 23-39ms before the change to 10-32ms (due to how we filter timers), so potentially the new code is actually faster than the previous code, though precise measurements are difficult to do.

After this refactor, there is now a single implementation of UpdatePrimitives for timer based tracks instead of three (scheduler, thread, GPU).

### DIFF
--- a/OrbitGl/GpuTrack.cpp
+++ b/OrbitGl/GpuTrack.cpp
@@ -4,13 +4,10 @@
 
 #include "GpuTrack.h"
 
-#include <limits>
-
 #include "Capture.h"
 #include "GlCanvas.h"
 #include "Profiling.h"
 #include "TimeGraph.h"
-#include "absl/flags/flag.h"
 #include "absl/strings/str_format.h"
 
 using orbit_client_protos::TimerInfo;
@@ -44,7 +41,7 @@ std::string MapGpuTimelineToTrackLabel(std::string_view timeline) {
 GpuTrack::GpuTrack(TimeGraph* time_graph,
                    std::shared_ptr<StringManager> string_manager,
                    uint64_t timeline_hash)
-    : Track(time_graph) {
+    : TimerTrack(time_graph) {
   text_renderer_ = time_graph->GetTextRenderer();
   timeline_hash_ = timeline_hash;
 
@@ -60,25 +57,23 @@ GpuTrack::GpuTrack(TimeGraph* time_graph,
       TriangleToggle::InitialStateUpdate::kReplaceInitialState);
 }
 
-//-----------------------------------------------------------------------------
-void GpuTrack::Draw(GlCanvas* canvas, PickingMode picking_mode) {
-  float track_height = GetHeight();
-  float track_width = canvas->GetWorldWidth();
+bool GpuTrack::IsTimerActive(const TimerInfo& timer_info) const {
+  bool is_same_tid_as_selected = timer_info.thread_id() == Capture::GSelectedThreadId;
+  // We do not properly track the PID for GPU jobs and we still want to show
+  // all jobs as active when no thread is selected, so this logic is a bit
+  // different than SchedulerTrack::IsTimerActive.
+  bool no_thread_selected = Capture::GSelectedThreadId == 0;
 
-  SetPos(canvas->GetWorldTopLeftX(), m_Pos[1]);
-  SetSize(track_width, track_height);
-
-  Track::Draw(canvas, picking_mode);
+  return is_same_tid_as_selected || no_thread_selected;
 }
 
-//-----------------------------------------------------------------------------
-Color GpuTrack::GetTimerColor(const TimerInfo& timer_info, bool is_selected,
-                              bool inactive) const {
+// //-----------------------------------------------------------------------------
+Color GpuTrack::GetTimerColor(const TimerInfo& timer_info, bool is_selected) const {
   const Color kInactiveColor(100, 100, 100, 255);
   const Color kSelectionColor(0, 128, 255, 255);
   if (is_selected) {
     return kSelectionColor;
-  } else if (inactive) {
+  } else if (!IsTimerActive(timer_info)) {
     return kInactiveColor;
   }
 
@@ -112,9 +107,25 @@ Color GpuTrack::GetTimerColor(const TimerInfo& timer_info, bool is_selected,
 }
 
 //-----------------------------------------------------------------------------
-inline float GetYFromDepth(const TimeGraphLayout& layout, float track_y,
-                           uint32_t depth) {
-  return track_y - layout.GetTextBoxHeight() * (depth + 1);
+float GpuTrack::GetYFromDepth(uint32_t depth) const {
+  float adjusted_depth = static_cast<float>(depth);
+  if (collapse_toggle_.IsCollapsed()) {
+    adjusted_depth = 0.f;
+  }
+  return m_Pos[1] -
+         time_graph_->GetLayout().GetTextBoxHeight() * (adjusted_depth + 1.f);
+}
+
+// When track is collapsed, only draw "hardware execution" timers.
+bool GpuTrack::TimerFilter(const TimerInfo& timer_info) const {
+  if (collapse_toggle_.IsCollapsed()) {
+    std::string gpu_stage =
+        string_manager_->Get(timer_info.user_data_key()).value_or("");
+    if (gpu_stage != kHwExecutionString) {
+      return false;
+    }
+  }
+  return true;
 }
 
 //-----------------------------------------------------------------------------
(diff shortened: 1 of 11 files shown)


ronaldfw:
One reason for doing this is because I want to introduce another track class (HeatmapTrack), which would have introduced the fourth copy of UpdatePrimitives with only minor changes.

ronaldfw:
@reichlfl Could you take a careful look at the GpuTrack changes? I might have accidentally inverted some of your changes when rebasing. Thank you!

ronaldfw:
> > @reichlfl Could you take a careful look at the GpuTrack changes? I might have accidentally inverted some of your changes when rebasing. Thank you!
> 
> Will do, but I will not get to it today, I'll take a look tomorrow morning!

Thanks, take your time, this is not urgent.

OrbitGl/ThreadTrack.cpp
```diff
@@ -126,7 +142,7 @@ void ThreadTrack::SetTimesliceText(const Timer& timer, double elapsed_us,
                                          time.c_str());
       text_box->SetText(text);
     } else {
-      ERROR("Unexpected case in ThreadTrack::SetTimesliceText");
+      ERROR("Unexpected case in TimerTrack::SetTimesliceText");
```
reichlfl, line 145: Looks like this message should not have been changed, the method remains in ThreadTrack

OrbitGl/ThreadTrack.h
```diff
@@ -2,84 +2,44 @@
...
-  mutable Mutex mutex_;
-  std::map<int, std::shared_ptr<TimerChain>> timers_;
+  std::string GetBoxTooltip(PickingID id) const override;
 
-  std::string GetBoxTooltip(PickingID id) const;
+  std::shared_ptr<EventTrack> event_track_;
+  int32_t thread_id_;
 };
+
+#endif  // ORBIT_GL_THREAD_TRACK_
```
reichlfl, line 45: Nit: Newline for consistency

OrbitGl/TimerTrack.cpp
```diff
@@ -0,0 +1,302 @@
...
+  min_time_ = std::numeric_limits<TickType>::max();
+  max_time_ = std::numeric_limits<TickType>::min();
+}
+
+//-----------------------------------------------------------------------------
+void TimerTrack::Draw(GlCanvas* canvas, PickingMode picking_mode) {
+  float track_height = GetHeight();
+  float track_width = canvas->GetWorldWidth();
+
+  SetPos(canvas->GetWorldTopLeftX(), m_Pos[1]);
```
reichlfl, line 35: I know you didn't introduce this, but I find this pattern to update pos and size in the draw() method not ideal. Both should not be influenced by the actual drawing, but rather updated when the underlying data or layouting change that forces this occurs. This may also happen when the view is scrolled to adjust the track height only to the actual visible timings if this is the desired behavior.

With the current pattern, I cannot get positioning information about the track outside of the draw() method, or only if I am certain that drawing has happened at least once after the last layout-influencing changes, and I have no way to force this since I can't just trigger a draw() at random. In general, the draw() method comes with side effects I would not expect as a user of this class.

I am not sure if this is something we can or even want to address, and if so probably not in this PR, but for me, these kind of side-effects made it hard to understand what's happening when I first started reading the Orbit code.

OrbitGl/TimerTrack.cpp
```diff
@@ -0,0 +1,302 @@
...
+            max_ignore = min_ignore + pixel_delta_in_ticks;
+          }
+        }
+      }
+    }
+  }
+}
+
+//-----------------------------------------------------------------------------
+void TimerTrack::OnDrag(int x, int y) { Track::OnDrag(x, y); }
```
reichlfl, line 172: I'd prefer to omit the method completely if it only calls the parent, so it is clear from the header file already that there is no change in functionality here. What's the reason to have it like this?

OrbitGl/SchedulerTrack.cpp
```diff
@@ -25,127 +25,36 @@ float SchedulerTrack::GetHeight() const {
          layout.GetTrackBottomMargin();
 }
 
-inline Color GetTimerColor(const Timer& timer, TimeGraph* time_graph,
-                           bool is_selected, bool same_tid, bool same_pid,
-                           bool inactive) {
+Color SchedulerTrack::GetTimerColor(const Timer& timer, bool is_selected,
+                                    bool inactive) const {
+  bool is_same_tid_as_selected = timer.m_TID == Capture::GSelectedThreadId;
+  bool is_inactive =
```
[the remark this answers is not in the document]
pierricgimmig: "inactive" is not used here, causing a compilation error on Linux.

OrbitGl/TimerTrack.cpp
```diff
@@ -0,0 +1,302 @@
...
+}
+
+//-----------------------------------------------------------------------------
+void TimerTrack::UpdatePrimitives(uint64_t min_tick, uint64_t max_tick,
+                                  PickingMode picking_mode) {
+  UpdateBoxHeight();
+
+  Batcher* batcher = &time_graph_->GetBatcher();
+  GlCanvas* canvas = time_graph_->GetCanvas();
+  const TimeGraphLayout& layout = time_graph_->GetLayout();
```
pierricgimmig, line 73: "layout" is not used.

OrbitGl/TimerTrack.h
```diff
@@ -0,0 +1,90 @@
...
+
+  // Pickable
+  void Draw(GlCanvas* canvas, PickingMode picking_mode) override;
+  void OnDrag(int x, int y) override;
+  void OnTimer(const Timer& timer);
+  std::string GetTooltip() const override;
+
+  // Track
+  void UpdatePrimitives(uint64_t min_tick, uint64_t max_tick,
+                        PickingMode picking_mode) override;
```
pierricgimmig, line 34: picking_mode is unused.

OrbitGl/GpuTrack.cpp
```diff
@@ -365,27 +179,6 @@ const TextBox* GpuTrack::GetRight(TextBox* text_box) const {
...
-std::vector<std::shared_ptr<TimerChain>> GpuTrack::GetAllChains() {
-  std::vector<std::shared_ptr<TimerChain>> chains;
-  for (const auto& pair : timers_) {
-    chains.push_back(pair.second);
-  }
-  return chains;
-}
-
 //-----------------------------------------------------------------------------
 std::string GpuTrack::GetBoxTooltip(PickingID id) const {
```
pierricgimmig, line 183: "id" is not used.

OrbitGl/TimerTrack.cpp
```diff
@@ -0,0 +1,302 @@
...
+}
+
+//-----------------------------------------------------------------------------
+bool TimerTrack::IsEmpty() const { return GetNumTimers() == 0; }
+
+//-----------------------------------------------------------------------------
+std::string TimerTrack::GetBoxTooltip(PickingID id) const { return ""; }
+
+//-----------------------------------------------------------------------------
+Color TimerTrack::GetTimerColor(const Timer& /*timer*/, bool /*is_selected*/,
```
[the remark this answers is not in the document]
ronaldfw: Good point!

OrbitGl/SchedulerTrack.cpp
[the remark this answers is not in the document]
ronaldfw: Good point, @reichlfl. I essentially followed your suggestion and added a method to determine (in)activeness of timers. Because the logic is easier to follow (IMO), I added a virtual method IsTimerActive() that returns true by default but is implemented differently by SchedulerTrack and ThreadTrack.

OrbitGl/ThreadTrack.cpp
ronaldfw: Thanks, I guess this was part of TimerTrack at some point during my refactor :)

OrbitGl/TimerTrack.cpp
ronaldfw: Good point, let's follow up on this. I think it's out of scope for this PR (and I wouldn't even know how to address this without looking into it, maybe it's easy, maybe it's hard).

OrbitGl/TimerTrack.cpp
ronaldfw: Yes, fixed.

OrbitGl/TimerTrack.cpp
ronaldfw: Fixed, thanks.

OrbitGl/TimerTrack.h
ronaldfw: Fixed. I left it in the signature, just marked as unused (commented out), as it is used by EventTrack.

OrbitGl/GpuTrack.cpp
ronaldfw: It's used on the first line (maybe changed?).

OrbitGl/SchedulerTrack.cpp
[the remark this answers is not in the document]
ronaldfw: I actually realized that GpuTrack::IsTimerActive can and should actually follow the same logic that is used in SchedulerTrack::IsTimerActive. When a thread is selected, only show GPU jobs that were submitted from that thread.

OrbitGl/TimerTrack.cpp
```diff
@@ -0,0 +1,302 @@
...
+            static_cast<float>(normalized_length * world_width);
+        float world_timer_x =
+            static_cast<float>(world_start_x + normalized_start * world_width);
+        float world_timer_y = GetYFromDepth(timer.m_Depth);
+
+        bool is_visible_width = normalized_length * canvas->getWidth() > 1;
+        bool is_selected = &text_box == Capture::GSelectedTextBox;
+        bool is_inactive =
+            !Capture::GVisibleFunctionsMap.empty() &&
+            Capture::GVisibleFunctionsMap[timer.m_FunctionAddress] == nullptr;
```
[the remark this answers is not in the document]
ronaldfw: Fixed.

OrbitGl/SchedulerTrack.cpp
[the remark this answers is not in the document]
reichlfl: Thanks! Bit of a nitpick to the changes: Is there a reason you made `IsTimerActive` public? Same for `IsTimerActive` and `SetTimesliceText` (from the base class).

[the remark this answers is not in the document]
ronaldfw: That was not intentional. Should be fixed as soon as I push the new changes.

ronaldfw:
I just realized that I never got an approval for this. Sorry, it was not the intention to merge without that. 

@pierricgimmig, @reichlfl If you have additional requests here, I will take care of them asap in a new PR.

reichlfl:
> I just realized that I never got an approval for this. Sorry, it was not the intention to merge without that.
> 
> @pierricgimmig, @reichlfl If you have additional requests here, I will take care of them asap in a new PR.

Should be fine, I've checked the changes from the last commits and wanted to approve anyways, just wasn't sure if you were done. Not sure if I checked the latest commit, and the changes from the master rebase weren't super easy to follow as I never looked at the proto refactoring, but it looked good to me.

PR #883, opened 2020-07-30
Add jumps in the same thread using alt

karupayun:
We add a new kind of jump. Also name kSameDepth to jumps in the same
depth (when you move left without shift and alt).

### DIFF
--- a/OrbitGl/TimeGraph.cpp
+++ b/OrbitGl/TimeGraph.cpp
@@ -442,7 +442,8 @@ void TimeGraph::Select(const TextBox* a_TextBox) {
 }
 
 const TextBox* TimeGraph::FindPreviousFunctionCall(
-    uint64_t function_address, TickType current_time) const {
+    uint64_t function_address, TickType current_time,
+    std::optional<int32_t> thread_ID) const {
   TextBox* previous_box = nullptr;
   TickType previous_box_time = std::numeric_limits<TickType>::lowest();
   std::vector<std::shared_ptr<TimerChain>> chains =
@@ -456,6 +457,8 @@ const TextBox* TimeGraph::FindPreviousFunctionCall(
         TextBox& box = block[i];
         auto box_time = box.GetTimerInfo().end();
         if ((box.GetTimerInfo().function_address() == function_address) &&
+            (!thread_ID ||
+             thread_ID.value() == box.GetTimerInfo().thread_id()) &&
             (box_time < current_time) && (previous_box_time < box_time)) {
           previous_box = &box;
           previous_box_time = box_time;
@@ -466,8 +469,9 @@ const TextBox* TimeGraph::FindPreviousFunctionCall(
   return previous_box;
 }
 
-const TextBox* TimeGraph::FindNextFunctionCall(uint64_t function_address,
-                                               TickType current_time) const {
+const TextBox* TimeGraph::FindNextFunctionCall(
+    uint64_t function_address, TickType current_time,
+    std::optional<int32_t> thread_ID) const {
   TextBox* next_box = nullptr;
   TickType next_box_time = std::numeric_limits<TickType>::max();
   std::vector<std::shared_ptr<TimerChain>> chains =
@@ -481,6 +485,8 @@ const TextBox* TimeGraph::FindNextFunctionCall(uint64_t function_address,
         TextBox& box = block[i];
         auto box_time = box.GetTimerInfo().end();
         if ((box.GetTimerInfo().function_address() == function_address) &&
+            (!thread_ID ||
+             thread_ID.value() == box.GetTimerInfo().thread_id()) &&
             (box_time > current_time) && (next_box_time > box_time)) {
           next_box = &box;
           next_box_time = box_time;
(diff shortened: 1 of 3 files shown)


OrbitGl/TimeGraph.cpp
```diff
@@ -914,21 +914,28 @@ void TimeGraph::JumpToNeighborBox(TextBox* from, JumpDirection jump_direction,
...
       goal = FindPreviousFunctionCall(function_address, current_time);
-    }
+    }    
+    if (jump_scope == JumpScope::kSameThreadSameFunction) {
+      goal = FindPreviousFunctionCall(function_address, current_time, thread_id);
+    }    
   }
   if (jump_direction == JumpDirection::kNext) {
-    if (jump_scope == JumpScope::kSameThread) {
+    if (jump_scope == JumpScope::kSameDepth) {
```
beckerhe, line 930: Can you turn that into a switch?

beckerhe: I think it's worthwhile because the compiler will warn about about unhandled enum values.

OrbitGl/TimeGraph.cpp
```diff
@@ -914,21 +914,28 @@ void TimeGraph::JumpToNeighborBox(TextBox* from, JumpDirection jump_direction,
   }
   auto function_address = from->GetTimer().m_FunctionAddress;
   auto current_time = from->GetTimer().m_End;
+  auto thread_id = from->GetTimer().m_TID;
   if (jump_direction == JumpDirection::kPrevious) {
-    if (jump_scope == JumpScope::kSameThread) {
+    if (jump_scope == JumpScope::kSameDepth) {
```
beckerhe, line 919: Can you turn that into a switch (see comment below)?

OrbitGl/TimeGraph.h
```diff
@@ -67,23 +67,30 @@ class TimeGraph {
...
+  void HorizontallyMoveIntoView(VisibilityType vis_type,
+                                const TextBox* text_box, double distance = 0.3);
   void VerticallyMoveIntoView(const TextBox* text_box);
   double GetTime(double a_Ratio);
   double GetTimeIntervalMicro(double a_Ratio);
   void Select(const TextBox* a_TextBox);
   enum class JumpScope {
     kGlobal,
+    kSameDepth,
     kSameThread,
```
beckerhe, line 81: From what I see kSameThread is not used anywhere, isn't it?

karupayun: If you move On and Down (arrow keys) you are moving in the same thread (at least by now).
If you move Left and Right you are moving in the same thread but also in the same depth (for example saying that we are jumping to the closer previous event in the same thread is false, because we are preserving the Depth). Do you have some suggestion for better names?
Maybe I have to split the jumpings and the usual-arrow-keys-move.

beckerhe: Ah sorry. No it's fine. My bad. I only scrolled up, but not down. Now I see the usages of kSameThread. Thanks!

OrbitGl/TimeGraph.cpp
karupayun: Is ok now? Or you also suggested to do a switch in the jump_direction?

beckerhe: No, a second nested switch would harm readability. But regarding the switch on `jump_scope`, you shouldn't provide a `default`-case. When having a `default`-case the compiler won't warn you about non-handled cases (which was the whole purpose of that change.)

OrbitGl/TimeGraph.cpp
```diff
@@ -456,7 +456,7 @@ const TextBox* TimeGraph::FindPreviousFunctionCall(
       for (uint64_t i = 0; i < block.size(); i++) {
         TextBox& box = block[i];
         auto box_time = box.GetTimer().m_End;
-        if ((box.GetTimer().m_FunctionAddress == function_address) &&
+        if ((box.GetTimer().m_FunctionAddress == function_address) && (!thread_ID || thread_ID.value() == box.GetTimer().m_TID) &&
```
ronaldfw, line 459: Nit: This should be formatted (line is too long). Did you run clang-format?

OrbitGl/TimeGraph.cpp
```diff
@@ -481,7 +481,7 @@ const TextBox* TimeGraph::FindNextFunctionCall(uint64_t function_address,
...
```
ronaldfw, line 484: Same here.

OrbitGl/TimeGraph.cpp
```diff
@@ -914,20 +914,36 @@ void TimeGraph::JumpToNeighborBox(TextBox* from, JumpDirection jump_direction,
...
+        break;
+      case JumpScope::kSameFunction:
+        goal = FindPreviousFunctionCall(function_address, current_time);
+        break;
+      case JumpScope::kSameThreadSameFunction:
+        goal =
+            FindPreviousFunctionCall(function_address, current_time, thread_id);
+        break;
+      default:
+        break;
```
beckerhe, line 931: Like discussed oob, this should be a CHECK(false). Apart from that it's fine from my side.

OrbitGl/TimeGraph.cpp
karupayun: I don't know what happened, probably I forgot it

karupayun:
Finished the modifications we talked with @beckerhe + fixes. Squashed in only one commit.

Will think how to communicate the options to users.

PR #884, opened 2020-07-31
Manual Instrumentation ground work.

pierricgimmig:
- Sample 6 integer registers on function entry.
- Add support for special TimerStart and TimerStop functions.
- TimerStart functions only need uprobes.
- TimerStop functions only need uretprobes.

### DIFF
--- a/OrbitLinuxTracing/ManualInstrumentationConfig.h
+++ b/OrbitLinuxTracing/ManualInstrumentationConfig.h
@@ -1,0 +1,35 @@
+// Copyright (c) 2020 The Orbit Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#ifndef ORBIT_LINUX_TRACING_MANUAL_INSTRUMENTATION_CONFIG_H_
+#define ORBIT_LINUX_TRACING_MANUAL_INSTRUMENTATION_CONFIG_H_
+
+#include <cstdint>
+
+#include "absl/container/flat_hash_set.h"
+
+class ManualInstrumentationConfig {
+ public:
+  inline void AddTimerStartAddress(uint64_t address) {
+    timer_start_addresses_.insert(address);
+  }
+
+  inline void AddTimerStopAddress(uint64_t address) {
+    timer_stop_addresses_.insert(address);
+  }
+
+  inline bool IsTimerStartAddress(uint64_t address) const {
+    return timer_start_addresses_.contains(address);
+  }
+
+  inline bool IsTimerStopAddress(uint64_t address) const {
+    return timer_stop_addresses_.contains(address);
+  }
+
+ private:
+  absl::flat_hash_set<uint64_t> timer_start_addresses_;
+  absl::flat_hash_set<uint64_t> timer_stop_addresses_;
+};
+
+#endif
--- a/OrbitLinuxTracing/ManualInstrumentationConfig.h
+++ b/OrbitLinuxTracing/ManualInstrumentationConfig.h
@@ -11,19 +11,19 @@
 
 class ManualInstrumentationConfig {
  public:
-  inline void AddTimerStartAddress(uint64_t address) {
+  void AddTimerStartAddress(uint64_t address) {
     timer_start_addresses_.insert(address);
   }
 
-  inline void AddTimerStopAddress(uint64_t address) {
+  void AddTimerStopAddress(uint64_t address) {
     timer_stop_addresses_.insert(address);
   }
 
-  inline bool IsTimerStartAddress(uint64_t address) const {
+  bool IsTimerStartAddress(uint64_t address) const {
     return timer_start_addresses_.contains(address);
   }
 
-  inline bool IsTimerStopAddress(uint64_t address) const {
+  bool IsTimerStopAddress(uint64_t address) const {
     return timer_stop_addresses_.contains(address);
   }
 
--- a/OrbitLinuxTracing/PerfEventOpen.h
+++ b/OrbitLinuxTracing/PerfEventOpen.h
@@ -101,9 +101,9 @@ static constexpr uint64_t SAMPLE_REGS_USER_SP_IP =
 // PerfEventRecords.h.
 static constexpr uint64_t SAMPLE_REGS_USER_AX = (1lu << PERF_REG_X86_AX);
 
-// This must be in sync with struct perf_event_uprobe in
-// PerfEventRecords.h.
-static constexpr uint64_t SAMPLE_REGS_UPROBE =
+// This must be in sync with struct perf_event_sample_regs_user_sp_ip_arguments
+// in PerfEventRecords.h.
+static constexpr uint64_t SAMPLE_REGS_USER_SP_IP_ARGUMENTS =
     (1lu << PERF_REG_X86_CX) | (1lu << PERF_REG_X86_DX) |
     (1lu << PERF_REG_X86_SI) | (1lu << PERF_REG_X86_DI) |
     (1lu << PERF_REG_X86_SP) | (1lu << PERF_REG_X86_IP) |
--- a/OrbitLinuxTracing/PerfEventOpen.h
+++ b/OrbitLinuxTracing/PerfEventOpen.h
@@ -101,6 +101,14 @@ static constexpr uint64_t SAMPLE_REGS_USER_SP_IP =
 // PerfEventRecords.h.
 static constexpr uint64_t SAMPLE_REGS_USER_AX = (1lu << PERF_REG_X86_AX);
 
+// This must be in sync with struct perf_event_uprobe in
+// PerfEventRecords.h.
+static constexpr uint64_t SAMPLE_REGS_UPROBE =
+    (1lu << PERF_REG_X86_CX) | (1lu << PERF_REG_X86_DX) |
+    (1lu << PERF_REG_X86_SI) | (1lu << PERF_REG_X86_DI) |
+    (1lu << PERF_REG_X86_SP) | (1lu << PERF_REG_X86_IP) |
+    (1lu << PERF_REG_X86_R8) | (1lu << PERF_REG_X86_R9);
+
 // Max to pass to perf_event_open without getting an error is (1u << 16u) - 8,
 // because the kernel stores this in a short and because of alignment reasons.
 // But the size the kernel actually returns is smaller, because the maximum size
--- a/OrbitLinuxTracing/PerfEventOpen.h
+++ b/OrbitLinuxTracing/PerfEventOpen.h
@@ -92,11 +92,6 @@ static constexpr uint64_t SAMPLE_REGS_USER_ALL =
     (1lu << PERF_REG_X86_R12) | (1lu << PERF_REG_X86_R13) |
     (1lu << PERF_REG_X86_R14) | (1lu << PERF_REG_X86_R15);
 
-// This must be in sync with struct perf_event_sample_regs_user_sp_ip in
-// PerfEventRecords.h.
-static constexpr uint64_t SAMPLE_REGS_USER_SP_IP =
-    (1lu << PERF_REG_X86_SP) | (1lu << PERF_REG_X86_IP);
-
 // This must be in sync with struct perf_event_ax_sample in
 // PerfEventRecords.h.
 static constexpr uint64_t SAMPLE_REGS_USER_AX = (1lu << PERF_REG_X86_AX);
--- a/OrbitLinuxTracing/PerfEventRecords.h
+++ b/OrbitLinuxTracing/PerfEventRecords.h
@@ -72,12 +72,18 @@ struct __attribute__((__packed__)) perf_event_sample_regs_user_ax {
   uint64_t ax;
 };
 
-// This struct must be in sync with the SAMPLE_REGS_USER_SP_IP in
+// This struct must be in sync with the SAMPLE_REGS_USER_SP_IP_ARGUMENTS in
 // PerfEventOpen.h.
-struct __attribute__((__packed__)) perf_event_sample_regs_user_sp_ip {
+struct __attribute__((__packed__)) perf_event_sample_regs_user_sp_ip_arguments {
   uint64_t abi;
+  uint64_t cx;
+  uint64_t dx;
+  uint64_t si;
+  uint64_t di;
   uint64_t sp;
   uint64_t ip;
+  uint64_t r8;
+  uint64_t r9;
 };
 
 struct __attribute__((__packed__)) perf_event_sample_stack_user {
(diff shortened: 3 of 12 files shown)


OrbitLinuxTracing/TracerThread.cpp
```diff
@@ -92,28 +103,48 @@ void TracerThread::InitUprobesEventProcessor() {
...
-        break;
+      if (open_uprobe) {
+        int uprobes_fd = uprobes_retaddr_event_open(
+            function.BinaryPath().c_str(), function.FileOffset(), -1, cpu);
+        if (uprobes_fd < 0) {
+          function_uprobes_open_error = true;
+          break;
+        }
+        function_uprobes_fds_per_cpu.emplace(cpu, uprobes_fd);
+        fds_per_cpu[cpu].emplace_back(uprobes_fd);
```
dpallotti, line 135: If (for some reason I'm not aware of, but bear with me) this uprobes fails to open not on the first cpu but on a subsequent one, all file descriptors for this function get closed (in the `if (function_uprobes_open_error)`), but don't you end up having recorded the file descriptor in `fds_per_cpu` (which you use later for the redirection) and that stays there? Or am I understanding incorrectly?

OrbitLinuxTracing/TracerThread.cpp
```diff
@@ -152,31 +183,20 @@ bool TracerThread::OpenUprobes(const std::vector<int32_t>& cpus) {
...
-        // will be redirected to this ring buffer.
-        int ring_buffer_fd = uprobes_fd;
-        std::string buffer_name = absl::StrFormat("uprobes_uretprobes_%d", cpu);
-        ring_buffers_.emplace_back(ring_buffer_fd, UPROBES_RING_BUFFER_SIZE_KB,
-                                   buffer_name);
-        uprobes_ring_buffer_fds_per_cpu[cpu] = ring_buffer_fd;
-        // Must be called after the ring buffer has been opened.
-        perf_event_redirect(uretprobes_fd, ring_buffer_fd);
-      }
+  for (auto& [/*int32_t*/ cpu, /*std::vector<int>*/ fds] : fds_per_cpu) {
```
dpallotti, line 188: The reason for changing this code is because you don't always open both the uprobe and the uretprobe, right? But as I said in the previous comment, I think the behavior on errors is different now (I might be wrong though). It can be made the same, but it needs additional/[even more] bookkeeping of file descriptors. That's why it was implemented that way. Wouldn't it be better to add `if`s based on `open_uprobe`\`open_uretprobe` to the old code?

protos/capture.proto
```diff
@@ -42,6 +49,7 @@ message FunctionCall {
   uint64 end_timestamp_ns = 5;
   int32 depth = 6;
   uint64 return_value = 7;
+  repeated uint64 registers = 8;
```
dpallotti, line 52: Are we sure we shouldn't have some more precise mapping here? This way which register is which is determined only by the order and it can quickly get confusing.

OrbitLinuxTracing/UprobesFunctionCallManagerTest.cpp
```diff
@@ -13,8 +13,9 @@ TEST(UprobesFunctionCallManager, OneUprobe) {
   constexpr pid_t tid = 42;
   std::optional<FunctionCall> processed_function_call;
   UprobesFunctionCallManager function_call_manager;
+  perf_event_uprobe_regs registers;
 
-  function_call_manager.ProcessUprobes(tid, 100, 1);
+  function_call_manager.ProcessUprobes(tid, 100, 1, registers);
```
dpallotti, line 18: Maybe add at least some assertions based on the new `registers` field?

OrbitLinuxTracing/PerfEventRecords.h
```diff
@@ -118,6 +118,25 @@ struct __attribute__((__packed__)) perf_event_sp_ip_8bytes_sample {
...
+  uint64_t dx;
+  uint64_t si;
+  uint64_t di;
+  uint64_t sp;
+  uint64_t ip;
+  uint64_t r8;
+  uint64_t r9;
+};
+
+struct __attribute__((__packed__)) perf_event_uprobe {
```
dpallotti, line 133: After adding this, is `perf_event_sp_ip_8bytes_sample` still used or should this just replace it?

OrbitLinuxTracing/ManualInstrumentationConfig.h
```diff
@@ -0,0 +1,35 @@
...
+#ifndef ORBIT_LINUX_TRACING_MANUAL_INSTRUMENTATION_CONFIG_H_
+#define ORBIT_LINUX_TRACING_MANUAL_INSTRUMENTATION_CONFIG_H_
+
+#include <cstdint>
+
+#include "absl/container/flat_hash_set.h"
+
+class ManualInstrumentationConfig {
+ public:
+  inline void AddTimerStartAddress(uint64_t address) {
```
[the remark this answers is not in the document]
pierricgimmig: Yep, fixed.

OrbitLinuxTracing/PerfEventOpen.h
```diff
@@ -101,6 +101,14 @@ static constexpr uint64_t SAMPLE_REGS_USER_SP_IP =
 // PerfEventRecords.h.
 static constexpr uint64_t SAMPLE_REGS_USER_AX = (1lu << PERF_REG_X86_AX);
 
+// This must be in sync with struct perf_event_uprobe in
+// PerfEventRecords.h.
+static constexpr uint64_t SAMPLE_REGS_UPROBE =
```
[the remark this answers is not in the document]
pierricgimmig: Done.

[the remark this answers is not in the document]
pierricgimmig: Done.

OrbitLinuxTracing/PerfEventRecords.h
```diff
@@ -118,6 +118,25 @@ struct __attribute__((__packed__)) perf_event_sp_ip_8bytes_sample {
   perf_event_sample_stack_user_8bytes stack;
 };
 
+struct __attribute__((__packed__)) perf_event_uprobe_regs {
```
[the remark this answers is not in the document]
pierricgimmig: Done.

OrbitLinuxTracing/PerfEventRecords.h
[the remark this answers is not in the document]
pierricgimmig: Done.

OrbitLinuxTracing/TracerThread.cpp
pierricgimmig: You're right. I completely rewrote the loop that creates uprobes and uretprobes. I think it's easier to reason about now. Please let me know what you think.

OrbitLinuxTracing/TracerThread.cpp
pierricgimmig: Please see new implementation.

protos/capture.proto
pierricgimmig: This was done so that we can decide to not send any registers and save on bandwidth. Eventually we could enable register tracking only for specific functions.

OrbitLinuxTracing/UprobesFunctionCallManagerTest.cpp
pierricgimmig: Done.

OrbitLinuxTracing/PerfEventRecords.h
pierricgimmig: I removed it, thanks.

pierricgimmig:
@dpallotti, thanks for the thorough review. I rewrote the main loop that creates the userspace probes in TracerThread, I think it's cleaner now. Please let me know what you think. Cheers!

OrbitLinuxTracing/TracerThread.cpp
```diff
@@ -89,98 +107,122 @@ void TracerThread::InitUprobesEventProcessor() {
...
-    }
-
-    if (function_uprobes_open_error) {
-      ERROR("Opening u(ret)probes for function at %#016lx",
-            function.VirtualAddress());
-      uprobes_event_open_errors = true;
-      for (const auto& uprobes_fd : function_uprobes_fds_per_cpu) {
-        close(uprobes_fd.second);
+    } else if (manual_instrumentation_config_.IsTimerStopAddress(address)) {
+      // Only open uretprobe for a "timer stop" manual instrumentation function.
```
dpallotti, line 182:
```suggestion
      // Only open uretprobes for a "timer stop" manual instrumentation function.
```

OrbitLinuxTracing/TracerThread.cpp
```diff
@@ -89,98 +107,122 @@ void TracerThread::InitUprobesEventProcessor() {
...
-          function.BinaryPath().c_str(), function.FileOffset(), -1, cpu);
-      if (uprobes_fd < 0) {
-        function_uprobes_open_error = true;
-        break;
+    absl::flat_hash_map<int32_t, int> uprobes_fds_per_cpu;
+    absl::flat_hash_map<int32_t, int> uretprobes_fds_per_cpu;
+    uint64_t address = function.VirtualAddress();
+
+    if (manual_instrumentation_config_.IsTimerStartAddress(address)) {
+      // Only open uprobe for a "timer start" manual instrumentation function.
```
dpallotti, line 175:
```suggestion
      // Only open uprobes for a "timer start" manual instrumentation function.
```

OrbitLinuxTracing/TracerThread.cpp
```diff
@@ -89,98 +107,122 @@ void TracerThread::InitUprobesEventProcessor() {
...
+      // Only open uretprobe for a "timer stop" manual instrumentation function.
+      if (!OpenUretprobes(function, cpus, &uretprobes_fds_per_cpu)) {
+        CloseFileDescriptors(uretprobes_fds_per_cpu);
+        uprobes_event_open_errors = true;
+        continue;
       }
-      for (const auto& uretprobes_fd : function_uretprobes_fds_per_cpu) {
-        close(uretprobes_fd.second);
+    } else {
+      // Open both uprobe and uretprobes for regular functions.
```
dpallotti, line 189:
```suggestion
      // Open both uprobes and uretprobes for regular functions.
```

OrbitLinuxTracing/TracerThread.h
```diff
@@ -58,10 +59,26 @@ class TracerThread {
...
+                   const std::vector<int32_t>& cpus,
+                   absl::flat_hash_map<int32_t, int>* fds_per_cpu);
+  bool OpenUretprobes(const LinuxTracing::Function& function,
+                      const std::vector<int32_t>& cpus,
+                      absl::flat_hash_map<int32_t, int>* fds_per_cpu);
   bool OpenMmapTask(const std::vector<int32_t>& cpus);
   bool OpenSampling(const std::vector<int32_t>& cpus);
 
+  void AddUprobesFileDescriptors(
+      const absl::flat_hash_map<int32_t, int>& uretprobes_fds_per_cpu,
```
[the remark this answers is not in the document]
pierricgimmig: thanks

protos/capture.proto
pierricgimmig: Oh I guess we could use optional here and explicitly name the registers. But also, the register names are architecture specific... not sure if naming registers is what we want. It seems having a generic array is more flexible. We can always revisit later if we really have a need, but I'd keep this implementation for now. I have an another PR coming up that makes use of those registers, we can discuss further there if needed.

OrbitLinuxTracing/TracerThread.cpp
```diff
@@ -89,98 +107,122 @@ void TracerThread::InitUprobesEventProcessor() {
...
-bool TracerThread::OpenUprobes(const std::vector<int32_t>& cpus) {
+bool TracerThread::OpenUprobes(const LinuxTracing::Function& function,
+                               const std::vector<int32_t>& cpus,
+                               absl::flat_hash_map<int32_t, int>* fds_per_cpu) {
+  const char* module = function.BinaryPath().c_str();
+  const uint64_t offset = function.FileOffset();
+  for (int32_t cpu : cpus) {
+    int fd = uprobes_retaddr_event_open(module, offset, -1, cpu);
+    if (fd < 0) {
+      ERROR("Uprobe 0x%llx on cpu %u", function.VirtualAddress(), cpu);
```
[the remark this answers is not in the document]
pierricgimmig: done

OrbitLinuxTracing/TracerThread.cpp
```diff
@@ -89,98 +107,122 @@ void TracerThread::InitUprobesEventProcessor() {
...
+
+bool TracerThread::OpenUretprobes(
+    const LinuxTracing::Function& function, const std::vector<int32_t>& cpus,
+    absl::flat_hash_map<int32_t, int>* fds_per_cpu) {
+  const char* module = function.BinaryPath().c_str();
+  const uint64_t offset = function.FileOffset();
+  for (int32_t cpu : cpus) {
+    int fd = uretprobes_event_open(module, offset, -1, cpu);
+    if (fd < 0) {
+      ERROR("Uretprobe 0x%llx on cpu %u", function.VirtualAddress(), cpu);
```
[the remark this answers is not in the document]
pierricgimmig: done

OrbitLinuxTracing/TracerThread.cpp
```diff
@@ -89,98 +107,122 @@ void TracerThread::InitUprobesEventProcessor() {
...
-                                   buffer_name);
-        uprobes_ring_buffer_fds_per_cpu[cpu] = ring_buffer_fd;
-        // Must be called after the ring buffer has been opened.
-        perf_event_redirect(uretprobes_fd, ring_buffer_fd);
-      }
+  return !uprobes_event_open_errors;
+}
+
+void TracerThread::OpenUserSpaceProbesRingBuffers() {
+  for (auto& [/*int32_t*/ cpu, /*std::vector<int>*/ fds] : fds_per_cpu_) {
```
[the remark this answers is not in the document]
pierricgimmig: yep

OrbitLinuxTracing/TracerThread.cpp
```diff
@@ -89,98 +107,122 @@ void TracerThread::InitUprobesEventProcessor() {
...
+    uprobes_ids_.insert(stream_id);
+    tracing_fds_.push_back(fd);
+    fds_per_cpu_[cpu].push_back(fd);
+  }
+}
+
+void TracerThread::AddUretprobesFileDescriptors(
+    const absl::flat_hash_map<int32_t, int>& uretprobes_fds_per_cpu,
+    const LinuxTracing::Function& function) {
+  for (const auto [cpu, fd] : uretprobes_fds_per_cpu) {
```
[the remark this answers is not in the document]
pierricgimmig: cpu and fd are both 32 bit values...

OrbitLinuxTracing/TracerThread.cpp
pierricgimmig: done

OrbitLinuxTracing/TracerThread.cpp
pierricgimmig: done

OrbitLinuxTracing/TracerThread.cpp
pierricgimmig: done

protos/capture.proto
dpallotti: This is proto3, all fields are optional.

As for registers being architecture specific, I don't think we should worry about that (yet, at least).

We can leave it like this for now and discuss it later on, but I think there should be some more explicit mapping.

OrbitLinuxTracing/TracerThread.cpp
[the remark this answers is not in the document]
dpallotti: Right, sorry.

OrbitLinuxTracing/TracerThread.cpp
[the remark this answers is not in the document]
dpallotti: I guess we can go with uretprobes for now but I would still be curious to get an answer on this (also I guess it would be interesting to measure).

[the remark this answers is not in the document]
pierricgimmig: Ah sorry for I missed the question. It has to do with the duplicate uprobe detection code that relies on the presence of a uretprobe. Also, it's easier to have one uprobe and one uretprobe so that we automatically have the proper stacking of timeslice events on the collector side.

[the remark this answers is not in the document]
dpallotti: Ah right, so that all the other processing is done in exactly the same way as dynamic instrumentation.

PR #885, opened 2020-07-31
Fix crashes when pickable elements are deleted between mouse down and release

reichlfl:
Bugfix for https://b/161966451. Crash occured when Pickable elements where deleted in between mouse down and mouse up since the PickingManager stored a raw pointer to the currently clicked element. Fixed by changing all Pickable pointers to std::weak_ptr.

Added a new UnitTest project for OrbitGl that reproduces the bug.


reichlfl:
@ronaldfw This will probably conflict with your refactoring, it is probably easiest if I wait until your changes are merged and then rebase my comparably small branch.

OrbitQt/orbitmainwindow.ui
```diff
@@ -241,22 +241,10 @@
     <addaction name="actionReport_Missing_Feature"/>
     <addaction name="actionReport_Bug"/>
    </widget>
-   <widget class="QMenu" name="menuCrash">
```
beckerhe: When looking at this PR's commits one by one, it's unclear why these changes to the .ui-file are needed. Looks like an accident. Maybe you could add a sentence to the commit message?

OrbitGl/PickingManagerTest.cpp
```diff
@@ -26,11 +27,92 @@ class PickableMock : public Pickable {
...
+  pickable.OnRelease();
+  ASSERT_FALSE(pickable.dragging_);
+  pickable.Reset();
+  ASSERT_FALSE(pickable.picked_);
+}
+
+// Simulate "rendering" the picking color into a uint32_t target
+PickingID MockRenderPickingColor(const Color& col_vec) {
+  uint32_t col;
+  memcpy(&col, &col_vec[0], sizeof(uint32_t));
```
beckerhe, line 51: Nit: std::memcpy and #include <cstring>

OrbitGl/EventTrack.h
```diff
@@ -11,7 +11,7 @@
 class GlCanvas;
 class TimeGraph;
 
-class EventTrack : public Track {
+class EventTrack : public virtual Track {
```
beckerhe, line 14: I'm currently trying to figure out why virtual inheritance is needed. Can you briefly explain that?

OrbitGl/PickingManager.h
```diff
@@ -101,22 +97,24 @@ class PickingManager {
   void Pick(uint32_t a_Id, int a_X, int a_Y);
   void Release();
   void Drag(int a_X, int a_Y);
-  Pickable* GetPicked();
-  Pickable* GetPickableFromId(uint32_t id);
-  bool IsDragging() const;
+  std::weak_ptr<Pickable> GetPicked();
+  std::weak_ptr<Pickable> GetPickableFromId(uint32_t id);
+  bool IsDragging();
```
beckerhe, line 103: I went throught the different implementations of IsDragging and I don't see why the const needs to go away. Can you briefly explain?

OrbitGl/PickingManager.h
```diff
@@ -101,22 +97,24 @@ class PickingManager {
   void Pick(uint32_t a_Id, int a_X, int a_Y);
   void Release();
   void Drag(int a_X, int a_Y);
-  Pickable* GetPicked();
-  Pickable* GetPickableFromId(uint32_t id);
-  bool IsDragging() const;
+  std::weak_ptr<Pickable> GetPicked();
```
beckerhe, line 101: I know you just changed the return type, but maybe it's a good occasion to add const here?

OrbitGl/PickingManager.h
```diff
@@ -101,22 +97,24 @@ class PickingManager {
...
+  std::weak_ptr<Pickable> GetPickableFromId(uint32_t id);
```
beckerhe, line 102: Same here?

OrbitGl/PickingManager.h
```diff
@@ -101,22 +97,24 @@ class PickingManager {
...
+  bool IsDragging();
+
+  Color GetPickableColor(std::weak_ptr<Pickable> pickable,
+                         PickingID::BatcherId batcher_id);
 
-  Color GetPickableColor(Pickable* pickable, PickingID::BatcherId batcher_id);
+  bool IsThisElementPicked(Pickable* pickable);
```
beckerhe, line 108: If GetPicked got `const` this could also be const. :-)

OrbitGl/TriangleToggle.h
```diff
@@ -12,7 +12,8 @@
 class TimeGraph;
 class GlCanvas;
 
-class TriangleToggle : public Pickable {
+class TriangleToggle : public Pickable,
+                       public std::enable_shared_from_this<Pickable> {
```
beckerhe, line 16: Nit: `enabled_shared_from_this<TriangleToggle>`

OrbitGl/PickingManager.h
```diff
@@ -7,6 +7,7 @@
 #include <cstring>
 #include <unordered_map>
 #include <vector>
+#include <memory>
```
beckerhe, line 10: Nit: Since you're already cleaning up the includes :-) `<type_traits>` and `<string>` are also missing. (Please feel free to ignore. Can also be done later since you haven't introduced these issues.)

reichlfl:
> Thanks.
> 
> Meta: Using shared_ptr/weak_ptr feels a bit like we are just covering up the underlying lifetime issues in the overall design. I bet this is still the best solution since other fixes would require significant refactorings. Can you confirm that? Have you thought about other approaches?

Two other approaches came to mind:
1) Adding a method to the PickingManager to clear the current state, which would be called on loading a new capture to reset the currently stored pointer
2) Make the Pickable keep track of it's registration with the PickingManager, and unregister on destruction

Reason for not choosing 1): It's easy to forget that this is happening, and it implies Pickables to be created and deleted in certain places, and not add themselves to the PickingManager unexpectedly
Reason for not choosing 2): This would have been equally fine, and I don't consider the shared_ptr/weak_ptr solution bad design or just covering up, I think it uses weak_ptrs what they are intended to be used for: Tracking pointers of objects that you know may be correctly deleted in other places. I think the main difference is, is the PickingManager something that the Pickables need to be aware of, or do they not need to care. Main reason for the decision was then that the Tracks where already managed with shared_ptr, and implementing solution 2) would either require each Pickable to potentially keep track of multiple PickingManagers they registered with, or force the PickingManager to be a singleton.

OrbitQt/orbitmainwindow.ui
reichlfl: Did you get this from an intermediate commit? I added a hotkey to the "clear capture" button temporarily to reproduce the behavior, but all UI changes should have been reverted.

beckerhe:
> > Thanks.
> > Meta: Using shared_ptr/weak_ptr feels a bit like we are just covering up the underlying lifetime issues in the overall design. I bet this is still the best solution since other fixes would require significant refactorings. Can you confirm that? Have you thought about other approaches?
> 
> Two other approaches came to mind:
> 
> 1. Adding a method to the PickingManager to clear the current state, which would be called on loading a new capture to reset the currently stored pointer
> 2. Make the Pickable keep track of it's registration with the PickingManager, and unregister on destruction
> 
> Reason for not choosing 1): It's easy to forget that this is happening, and it implies Pickables to be created and deleted in certain places, and not add themselves to the PickingManager unexpectedly
> Reason for not choosing 2): This would have been equally fine, and I don't consider the shared_ptr/weak_ptr solution bad design or just covering up, I think it uses weak_ptrs what they are intended to be used for: Tracking pointers of objects that you know may be correctly deleted in other places. I think the main difference is, is the PickingManager something that the Pickables need to be aware of, or do they not need to care. Main reason for the decision was then that the Tracks where already managed with shared_ptr, and implementing solution 2) would either require each Pickable to potentially keep track of multiple PickingManagers they registered with, or force the PickingManager to be a singleton.

Fair enough. Thanks!

OrbitQt/orbitmainwindow.ui
beckerhe: Yes you revert them in the last commit. (before the fixup commit). But when looking at the first commit it's unclear why the whole crash-menu is deleted. I suspect this might also not compile on its own. Ideally commits should compile on its own to make bisecting possible.

beckerhe: Link: https://github.com/google/orbit/pull/885/commits/1ee8cf2f56a1bfda1ec8f0145bb3c0958ead825f

OrbitGl/EventTrack.h
reichlfl: I though I needed it to be able to inherit from std::enable_shared_from_this<Track> in "Track", but apparently I was mislead by another compiler error, because this should only be needed with multiple inheritance... looks like I can remove this again, thanks.

OrbitGl/PickingManager.h
reichlfl: I was under the impression `lock` wasn't `const`, again because of a compiler error that indicated so... I think at some point there was something else wrong in the code which lead me to change this and the "virtual" from your comment above, that's weird. I'll add it back.

OrbitGl/PickingManagerTest.cpp
reichlfl: Done

OrbitGl/PickingManager.h
reichlfl: Done

OrbitGl/PickingManager.h
reichlfl: I can even add two consts here! ;-)

OrbitGl/PickingManager.h
reichlfl: Added, since I'm already cleaning up ;-)

dpallotti:
One note regarding branches: if you don't create branches/pull requests in/from your own fork of Orbit but directly in `google/orbit` (I think that most are using their own fork by now but we haven't really decided on this), please prepend `<ldap_or_username>/` to the name of the branch.

reichlfl:
> One note regarding branches: if you don't create branches/pull requests in/from your own fork of Orbit but directly in `google/orbit` (I think that most are using their own fork by now but we haven't really decided on this), please prepend `<ldap_or_username>/` to the name of the branch.

Will do, thanks

OrbitQt/orbitmainwindow.ui
beckerhe: As discussed out-of-band: Since this will be squashed it's pointless to fix that. Thanks!

reichlfl: As discussed in GVC, not really needed when we squash/merge - but fixed anyways, thanks for showing me how!

PR #888, opened 2020-08-03
Always load symbols from the local file

dimitry-:
This PR consists of 3 parts

part 1:

    Add service call to find debug symbols
    
    This call attempt to find debug symbols on the collector side
    it does not load them but instead return path which could then
    be used to retrieve module via scp.

part 2:

    Keep sftp_channel alive and implement CopyToLocal
    
    This commit makes ServiceDeployManager keep sftp channel
    alive in order to be able to copy debug symbols from remote to
    local while loading symbols. It also adds CopyToLocal method
    to do that.

part 3:

    Always load symbols from local file
    
    This commit modifies Load Symbols shifting symbol loading from
    the service to the client. The modified workflow works in the following
    way:
    
    1. Orbit tries to load symbols from local machine using paths
       from SymbolsPathFile and Cache directory.
    2. If unsuccessful it asks the service to find a debug info file
       on the server side. The call returns path to the file if successful.
    3. If the debug info file was found on the server side Orbit copies
       it to the cache directory and loads symbols.

Test: Start Orbit, load symbols from the server. Restart Orbit, load symbols from the cache.
 Bug: http://b/161352292

### DIFF
--- a/OrbitCore/SymbolHelper.cpp
+++ b/OrbitCore/SymbolHelper.cpp
@@ -6,6 +6,7 @@
 
 #include <absl/strings/match.h>
 #include <absl/strings/str_format.h>
+#include <absl/strings/str_replace.h>
 
 #include <fstream>
 
@@ -175,4 +176,32 @@ ErrorMessageOr<ModuleSymbols> SymbolHelper::LoadUsingSymbolsPathFile(
 ErrorMessageOr<std::string> SymbolHelper::FindDebugSymbolsFile(
     const std::string& module_path, const std::string& build_id) const {
   return FindSymbolsFile(module_path, collector_symbol_directories_, build_id);
+}
+
+ErrorMessageOr<ModuleSymbols> SymbolHelper::LoadSymbolsFromFile(
+    const std::string& file_path, const std::string& build_id) const {
+  ErrorMessageOr<std::unique_ptr<ElfFile>> elf_file_result =
+      ElfFile::Create(file_path);
+
+  if (!elf_file_result) {
+    return ErrorMessage(
+        absl::StrFormat("Failed to load debug symbols from \"%s\": %s",
+                        file_path, elf_file_result.error().message()));
+  }
+
+  const std::string& target_build_id = elf_file_result.value()->GetBuildId();
+  if (target_build_id != build_id) {
+    return ErrorMessage(
+        absl::StrFormat("Failed to load debug symbols from \"%s\": invalid "
+                        "build id \"%s\", expected: \"%s\"",
+                        file_path, target_build_id, build_id));
+  }
+
+  return elf_file_result.value()->LoadSymbols();
+}
+
+std::string SymbolHelper::GenerateCachedFileName(
+    const std::string& file_path) const {
+  auto file_name = absl::StrReplaceAll(file_path, {{"/", "_"}});
+  return Path::JoinPath({Path::GetCachePath(), file_name});
 }
--- a/OrbitCore/SymbolHelper.cpp
+++ b/OrbitCore/SymbolHelper.cpp
@@ -57,14 +57,16 @@ std::vector<std::string> ReadSymbolsFile() {
   return directories;
 }
 
-ErrorMessageOr<ModuleSymbols> FindSymbols(
-    const std::string& module_path, const std::string& build_id,
-    const std::vector<std::string>& search_directories) {
+ErrorMessageOr<std::string> FindSymbolsFile(
+    const std::string& module_path,
+    const std::vector<std::string>& search_paths, const std::string& build_id) {
   std::string filename = Path::GetFileName(module_path);
   std::string filename_without_extension = Path::StripExtension(filename);
 
   std::vector<std::string> search_file_paths;
-  for (const auto& directory : search_directories) {
+  search_file_paths.push_back(module_path);
+
+  for (const auto& directory : search_paths) {
     search_file_paths.emplace_back(
         Path::JoinPath({directory, filename_without_extension + ".debug"}));
     search_file_paths.emplace_back(
@@ -80,15 +82,47 @@ ErrorMessageOr<ModuleSymbols> FindSymbols(
         ElfFile::Create(symbols_file_path);
     if (!symbols_file) continue;
     if (!symbols_file.value()->HasSymtab()) continue;
+
+    // don't check build_id of module_path itself has symbols
+    // and requested build_id is empty.
+    if (build_id.empty() && symbols_file_path == module_path) {
+      return symbols_file_path;
+    }
+
+    // Otherwise check that build_id is not empty and correct.
+    if (symbols_file.value()->GetBuildId().empty()) continue;
     if (symbols_file.value()->GetBuildId() != build_id) continue;
 
-    LOG("Loading symbols for module \"%s\" from \"%s\"", module_path,
+    LOG("Found debug info for module \"%s\" -> \"%s\"", module_path,
         symbols_file_path);
-    return symbols_file.value()->LoadSymbols();
+    return symbols_file_path;
   }
 
-  return ErrorMessage(
-      absl::StrFormat("Could not find symbols for module \"%s\"", module_path));
+  return ErrorMessage(absl::StrFormat(
+      "Could not find a file with debug symbols for module \"%s\"",
+      module_path));
+}
+
+ErrorMessageOr<ModuleSymbols> FindSymbols(
+    const std::string& module_path, const std::string& build_id,
+    const std::vector<std::string>& search_paths) {
+  ErrorMessageOr<std::string> debug_info_file_path =
+      FindSymbolsFile(module_path, search_paths, build_id);
+
+  if (!debug_info_file_path) {
+    return debug_info_file_path.error();
+  }
+
+  ErrorMessageOr<std::unique_ptr<ElfFile>> elf_file_result =
+      ElfFile::Create(debug_info_file_path.value());
+
+  if (!elf_file_result) {
+    return ErrorMessage(absl::StrFormat(
+        "Failed to load debug symbols for \"%s\" from \"%s\": %s", module_path,
+        debug_info_file_path.value(), elf_file_result.error().message()));
+  }
+
+  return elf_file_result.value()->LoadSymbols();
 }
 
 }  // namespace
@@ -136,4 +170,9 @@ ErrorMessageOr<ModuleSymbols> SymbolHelper::LoadSymbolsCollector(
 ErrorMessageOr<ModuleSymbols> SymbolHelper::LoadUsingSymbolsPathFile(
     const std::string& module_path, const std::string& build_id) const {
   return FindSymbols(module_path, build_id, symbols_file_directories_);
+}
+
+ErrorMessageOr<std::string> SymbolHelper::FindDebugSymbolsFile(
+    const std::string& module_path, const std::string& build_id) const {
+  return FindSymbolsFile(module_path, collector_symbol_directories_, build_id);
 }
--- a/OrbitCore/SymbolHelper.h
+++ b/OrbitCore/SymbolHelper.h
@@ -18,13 +18,18 @@ class SymbolHelper {
       : collector_symbol_directories_(std::move(collector_symbol_directories)),
         symbols_file_directories_(std::move(symbols_file_directories)){};
 
-  ErrorMessageOr<ModuleSymbols> LoadSymbolsCollector(
+  [[nodiscard]] ErrorMessageOr<ModuleSymbols> LoadSymbolsCollector(
       const std::string& module_path) const;
-  ErrorMessageOr<ModuleSymbols> LoadUsingSymbolsPathFile(
+  [[nodiscard]] ErrorMessageOr<ModuleSymbols> LoadUsingSymbolsPathFile(
+      const std::string& module_path, const std::string& build_id) const;
+  [[nodiscard]] ErrorMessageOr<ModuleSymbols> LoadSymbolsFromFile(
+      const std::string& file_path, const std::string& build_id) const;
+
+  [[nodiscard]] ErrorMessageOr<std::string> FindDebugSymbolsFile(
       const std::string& module_path, const std::string& build_id) const;
 
-  ErrorMessageOr<std::string> FindDebugSymbolsFile(
-      const std::string& module_path, const std::string& build_id) const;
+  [[nodiscard]] std::string GenerateCachedFileName(
+      const std::string& file_path) const;
 
  private:
   const std::vector<std::string> collector_symbol_directories_;
--- a/OrbitCore/SymbolHelper.h
+++ b/OrbitCore/SymbolHelper.h
@@ -23,6 +23,9 @@ class SymbolHelper {
   ErrorMessageOr<ModuleSymbols> LoadUsingSymbolsPathFile(
       const std::string& module_path, const std::string& build_id) const;
 
+  ErrorMessageOr<std::string> FindDebugSymbolsFile(
+      const std::string& module_path, const std::string& build_id) const;
+
  private:
   const std::vector<std::string> collector_symbol_directories_;
   const std::vector<std::string> symbols_file_directories_;
--- a/OrbitCore/SymbolHelperTest.cpp
+++ b/OrbitCore/SymbolHelperTest.cpp
@@ -9,6 +9,7 @@
 
 #include "Path.h"
 #include "SymbolHelper.h"
+#include "absl/strings/ascii.h"
 #include "symbol.pb.h"
 
 const std::string executable_directory =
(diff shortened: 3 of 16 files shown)


OrbitQt/main.cpp
```diff
@@ -99,34 +99,33 @@ static outcome::result<void> RunUiInstance(
...
+    return std::make_tuple(tunnel_ports, QString{});
   }());
   const auto& [ports, capture_path] = result;
 
   ApplicationOptions options;
   options.grpc_server_address =
       absl::StrFormat("127.0.0.1:%d", ports.grpc_port);
 
-  OrbitMainWindow w(app, std::move(options));
+  OrbitMainWindow w(app, std::move(options), &service_deploy_manager.value());
```
beckerhe, line 128: Is that a problem in the load-capture-case? `service_deploy_manager` is not set in that case, isn't it?

OrbitQt/servicedeploymanager.h
```diff
@@ -35,6 +36,13 @@ class ServiceDeployManager : public QObject {
       const GrpcPort& grpc_port, QObject* parent = nullptr);
 
   outcome::result<GrpcPort> Exec();
+
+  // This method opens SftpChannel and initiates a asynchronous copy of the
+  // remote source file to destination.
+  void CopyFileToLocal(std::string_view source, std::string_view destination,
+                       std::function<void()> copy_complete_callback,
+                       std::function<void(ErrorMessage)> error_callback);
```
beckerhe, line 44: I see why you're hesitant about callbacks. We have callbacks, inside callbacks, inside callbacks here.

One option to improve that would be not to provide a function here taking two callbacks. I would rather use `SftpCopyToLocalOperation` directly where it's actually used. But that's in OrbitGl/App.cpp, isn't it?

Do you see a sensible way to move that to OrbitQt?

If not, I might have some other idea around QFuture/std::future but I first have to look into it.

OrbitCore/SymbolHelper.cpp
```diff
@@ -165,3 +166,29 @@ ErrorMessageOr<std::string> SymbolHelper::FindDebugSymbolsFile(
     const std::string& module_path, const std::string& build_id) const {
   return FindSymbolsFile(module_path, collector_symbol_directories_, build_id);
 }
+ErrorMessageOr<ModuleSymbols> SymbolHelper::LoadSymbolsFromFile(
```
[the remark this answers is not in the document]
dimitry-: done.

OrbitCore/SymbolHelper.cpp
```diff
@@ -165,3 +166,29 @@ ErrorMessageOr<std::string> SymbolHelper::FindDebugSymbolsFile(
...
+  if (target_build_id != build_id) {
+    return ErrorMessage(
+        absl::StrFormat("Failed to load debug symbols from \"%s\": invalid "
+                        "build id \"%s\", expected: \"%s\"",
+                        file_path, target_build_id, build_id));
+  }
+
+  return elf_file_result.value()->LoadSymbols();
+}
+std::string SymbolHelper::GenerateCachedFileName(
```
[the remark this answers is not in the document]
dimitry-: done.

OrbitCore/SymbolHelperTest.cpp
```diff
@@ -100,3 +101,36 @@ TEST(SymbolHelper, LoadUsingSymbolsPathFile) {
...
+      file_path, "b5413574bbacec6eacb3b89b1012d0e2cd92ec6b");
+
+  ASSERT_TRUE(symbols_result) << symbols_result.error().message();
+  ModuleSymbols symbols = std::move(symbols_result.value());
+
+  EXPECT_EQ(symbols.symbols_file_path(), file_path);
+  EXPECT_FALSE(symbols.symbol_infos().empty());
+}
+
+TEST(SymbolHelper, LoadFromFileIvalidFile) {
```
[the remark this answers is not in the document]
dimitry-: done.

OrbitGl/App.h
```diff
@@ -196,6 +196,14 @@ class OrbitApp final : public DataViewFactory, public CaptureListener {
     clipboard_callback_ = std::move(callback);
   }
 
+  typedef std::function<void(std::string_view, std::string_view,
```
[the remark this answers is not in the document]
dimitry-: done.

OrbitGl/ProcessManager.cpp
```diff
@@ -20,7 +20,7 @@ constexpr uint64_t kGrpcDefaultTimeoutMilliseconds = 1000;
 // The actual symbol loading of a module should be relatively quick, but to
 // support slow internet connections, or multiple requests in parallel, this is
 // set to a generous 60 seconds.
-constexpr uint64_t kGrpcSymbolLoadingTimeoutMilliseconds = 60 * 1000;
+// constexpr uint64_t kGrpcSymbolLoadingTimeoutMilliseconds = 60 * 1000;
```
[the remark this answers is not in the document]
dimitry-: yes. removed.

dimitry-:
Regarding cache files:
1. The names for cached files are generated based on module_path (from maps file) - so for the same module (in case of outdated build_id) it will override the same file. That is unless the name was changed.
2. We still might want to add some startup check and offer user to clean the cache directory if it gets too big, I would say at this point this is a low priority task - created a bug: http://b/162721117

OrbitQt/servicedeploymanager.h
beckerhe: Yes, I see. I guess this one-line should be blocking and it should be callable from another thread?

dimitry-: Yes, that would be the idea. Having a thread with synchronous calls reduces number of callbacks to 1 (usually), which makes code more readable. Also it reduces the scope of some local variables in case you need to do more than one thing.

These are 2 main reasons I do not like asynchronous way of writing the code.

beckerhe: I see. And I kind of agree.

What about a wrapper function which can be called from any thread? Something like this?

```
#include "servicedeploymanager.h"
#include <future>
#include <QMetaObject>
#include <QCoreApplication>

ErrorMessageOr<void> copyFileFromRemote(ServiceDeployManager* mgr, std::string source, std::string dest) {

    std::promise<ErrorMessageOr<void>> promise;
    auto future = promise.get_future();

    // That queues the lambda-execution on the main thread.
    QMetaObject::invokeMethod(QCoreApplication::instance(),
      [promise = std::move(promise), mgr, source, dest]() {
        // This assume the previous (blocking) implementation of CopyFileToLocal
        promise.set_result(mgr.CopyFileToLocal(source, dest));
    });

    // This waits for the result
    return future.get();
}
```

dimitry-: done - in a slightly different way.

dimitry-:
ptal

antonrohr:
I did not look at the code yet, but I have one general idea. I am unsure if this makes things easier and have not looked at the performance implications. 
How about finding symbol file paths while listing the modules on the service? This would add some overhead of checking if files exist, but would get rid of the extra grpc call. We could also indicate whether a module is loadable or not in the module list. This could also be a different PR.

OrbitCore/SymbolHelper.cpp
```diff
@@ -137,3 +161,36 @@ ErrorMessageOr<ModuleSymbols> SymbolHelper::LoadUsingSymbolsPathFile(
...
+                        file_path, target_build_id, build_id));
+  }
+
+  return elf_file_result.value()->LoadSymbols();
+}
+
+std::string SymbolHelper::GenerateCachedFileName(
+    const std::string& file_path) const {
+  auto file_name = absl::StrReplaceAll(file_path, {{"/", "_"}});
+  return Path::JoinPath({Path::GetCachePath(), file_name});
```
antonrohr, line 195: Did you consider also involving the build id here?

OrbitCore/SymbolHelper.cpp
```diff
@@ -137,3 +161,36 @@ ErrorMessageOr<ModuleSymbols> SymbolHelper::LoadUsingSymbolsPathFile(
...
+      ElfFile::Create(file_path);
+
+  if (!elf_file_result) {
+    return ErrorMessage(
+        absl::StrFormat("Failed to load debug symbols from \"%s\": %s",
+                        file_path, elf_file_result.error().message()));
+  }
+
+  const std::string& target_build_id = elf_file_result.value()->GetBuildId();
+  if (target_build_id != build_id) {
```
antonrohr, line 182: could both of these strings be empty? 

Consider an elf file that contains symbols but not a build id. This might actually become a problem at other code locations as well

OrbitGl/App.cpp
```diff
@@ -771,29 +771,60 @@ void OrbitApp::LoadModuleOnRemote(int32_t process_id,
...
-    if (!remote_symbols_result) {
-      SendErrorToUi(
-          "Error loading symbols",
-          absl::StrFormat(
-              "Did not find symbols on local machine for module \"%s\".\n"
-              "Trying to load symbols from remote resulted in error "
-              "message: %s",
-              module->m_Name, remote_symbols_result.error().message()));
+    const std::string& module_path = module->m_FullName;
+    const std::string& build_id = module->m_DebugSignature;
```
antonrohr, line 775: I think `module->m_DebugSignature` can be empty here. This is not an issue here, but might be one later

OrbitCore/SymbolHelper.cpp
```diff
@@ -137,3 +161,36 @@ ErrorMessageOr<ModuleSymbols> SymbolHelper::LoadUsingSymbolsPathFile(
     const std::string& module_path, const std::string& build_id) const {
   return FindSymbols(module_path, build_id, symbols_file_directories_);
 }
+
+ErrorMessageOr<std::string> SymbolHelper::FindDebugSymbolsFile(
+    const std::string& module_path, const std::string& build_id) const {
+  return FindSymbolsFile(module_path, collector_symbol_directories_, build_id);
```
antonrohr, line 167: Here only the `collector_symbol_directories_` are looked at. This means if the symbols are included in the modules elf file, they are not found. 
Symbols included in the library is a common scenario.

OrbitCore/SymbolHelper.cpp
dimitry-: This will increase necessary size of cache directly - currently we have a situation when if a module is updated it will override exiting (old) cache file, which saves space and is most common scenario. It is possible that when jumping between 2 version it will require do re-download file every time but I think it is more rare use-case. Having said that I do not feel strongly about this. It might make sense to add build_id to the name to avoid extra downloads.

OrbitCore/SymbolHelper.cpp
dimitry-: that's a good point.

OrbitCore/SymbolHelper.cpp
dimitry-: Good catch. Will fix

OrbitCore/SymbolHelper.cpp
dimitry-: added this use-case and a test for this.

OrbitCore/SymbolHelper.cpp
dimitry-: I am going to leave it as is for now.

OrbitGl/App.cpp
dimitry-: I added handling of empty build id on the service side - it should correctly check/disregard it now. It disregards it only if build_id is empty and we ended up finding symbols in the original file.

OrbitCore/SymbolHelper.cpp
dimitry-: done.

OrbitCore/SymbolHelperTest.cpp
```diff
@@ -19,12 +20,63 @@ TEST(SymbolHelper, LoadSymbolsCollectorSameFile) {
...
+  const std::string kEmptyBuildId = "invalid_build_id";
+
+  SymbolHelper symbol_helper({executable_directory}, {});
+  const auto debug_symbol_file_path =
+      symbol_helper.FindDebugSymbolsFile(kFilePath, kBuildId);
+  ASSERT_TRUE(debug_symbol_file_path)
+      << debug_symbol_file_path.error().message();
+  EXPECT_EQ(debug_symbol_file_path.value(), kFilePath);
+
+  const auto invalud_build_id_result =
```
antonrohr, line 48:
```suggestion
  const auto invalid_build_id_result =
```

OrbitCore/SymbolHelperTest.cpp
```diff
@@ -19,12 +20,63 @@ TEST(SymbolHelper, LoadSymbolsCollectorSameFile) {
...
+  EXPECT_THAT(invalud_build_id_result.error().message(),
+              testing::HasSubstr("Could not find a file with debug symbols"));
+
+  const auto empty_build_id_result =
+      symbol_helper.FindDebugSymbolsFile(kFilePath, kEmptyBuildId);
+  ASSERT_TRUE(empty_build_id_result) << empty_build_id_result.error().message();
+  EXPECT_EQ(empty_build_id_result.value(), kFilePath);
+}
+
+TEST(SymbolHelper, FindDebugSybmolsSeparateFile) {
```
antonrohr, line 60:
```suggestion
TEST(SymbolHelper, FindDebugSymbolsSeparateFile) {
```

OrbitQt/servicedeploymanager.cpp
```diff
@@ -166,70 +167,78 @@ outcome::result<void> ServiceDeployManager::CopyFileToRemote(
...
+
+  auto quit_handler = ConnectQuitHandler(
+      &operation, &OrbitSshQt::SftpCopyToLocalOperation::stopped);
+
+  auto error_handler = ConnectErrorHandler(
+      &operation, &OrbitSshQt::SftpCopyToLocalOperation::errorOccurred);
+
+  operation.CopyFileToLocal(source, destination);
+
+  auto result = loop_.exec();
```
antonrohr, line 219: Here is some kind of problem.

When symbols of two (or) more modules that are not in the cache are requested at the same time, then this error is printed:`QEventLoop::exec: instance 0x7ffcbf53ae60 has already called exec()` and the loading of any but one of the files fails. 
This situation can occur when the user selects several modules and right clicks load symbols and if a preset is loaded.

dimitry-: Oh well, I guess it forces us back into callbacks world. @beckerhe is there another way this use-case can be handled? I do not think we can create another instance of an event loop since we have a lot of ssh related signals set up in this one.

OrbitCore/SymbolHelperTest.cpp
dimitry-: done.

OrbitCore/SymbolHelperTest.cpp
```diff
@@ -36,6 +36,7 @@ TEST(SymbolHelper, FindDebugInfoFileSameFile) {
       Path::JoinPath({executable_directory, "hello_world_elf"});
   const std::string kBuildId = "d12d54bc5b72ccce54a408bdeda65e2530740ac8";
   const std::string kInvalidBuildId = "invalid_build_id";
+  const std::string kEmptyBuildId = "invalid_build_id";
```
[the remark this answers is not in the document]
dimitry-: Oops.. I must have misread my local test results some how. fixed.

OrbitCore/SymbolHelperTest.cpp
dimitry-: done.

OrbitQt/servicedeploymanager.cpp
beckerhe: In this case don't use the member "loop_". When you create a local `QEventLoop`-instance in that function it should work. But you can't use "ConnectQuitHandler" and "ConnectErrorHandler" though since these connect to `loop_`. You probably should overload those with a version that takes a QEventLoop-pointer.

dimitry-: ok, since this change is actually doing a lot I am going to fixup and rebase everything. Please bare with me and take another look after I force-push next revision.

dimitry-: done.

OrbitCore/SymbolHelper.h
```diff
@@ -18,10 +18,18 @@ class SymbolHelper {
...
         symbols_file_directories_(std::move(symbols_file_directories)){};
 
-  ErrorMessageOr<ModuleSymbols> LoadSymbolsCollector(
+  [[nodiscard]] ErrorMessageOr<ModuleSymbols> LoadSymbolsCollector(
       const std::string& module_path) const;
-  ErrorMessageOr<ModuleSymbols> LoadUsingSymbolsPathFile(
+  [[nodiscard]] ErrorMessageOr<ModuleSymbols> LoadUsingSymbolsPathFile(
       const std::string& module_path, const std::string& build_id) const;
+  [[nodiscard]] ErrorMessageOr<ModuleSymbols> LoadSymbolsFromFile(
+      const std::string& file_path, const std::string& build_id) const;
```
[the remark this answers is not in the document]
dimitry-: Yep this is for the future change - It probably makes sense.

dimitry-:
I am going to merge it now (want to do it before fixit week starts) - @antonrohr, @dpallotti please add a review if you notice something - I will address it in follow-up PRs.

PR #889, opened 2020-08-03
Make gRPC CaptureClient independent from OrbitGl project

kisenshi:
### Move CaptureClient classes to new OrbitCaptureClient project.

Refactoring code related to CaptureClient so it is independent from OrbitGl. This is necessary for the creation of a new client able to handle capture requests without depending on the OrbitGl project.

CaptureClient depends on CaptureListener and CaptureEventProcessor, so it was needed to move these classes as well. Because these classes exist and are all related to a capture and not just gRPC calls, the project has been given the name of OrbitCaptureClient.


OrbitCaptureClient/CaptureListener.h
```diff
@@ -2,8 +2,8 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
 
-#ifndef ORBIT_GL_CAPTURE_LISTENER_H_
-#define ORBIT_GL_CAPTURE_LISTENER_H_
```
dimitry-, line 6: For new components please follow the structure as it is in OrbitBase - please include files under indlude/<component_name>/File.h

OrbitCaptureClient/CMakeLists.txt
```diff
@@ -0,0 +1,29 @@
...
+  PUBLIC CaptureClient.h
+         CaptureEventProcessor.h
+         CaptureListener.h)
+
+target_sources(
+  OrbitCaptureClient
+  PRIVATE CaptureClient.cpp
+          CaptureEventProcessor.cpp)
+
+target_include_directories(OrbitCaptureClient PUBLIC ${CMAKE_CURRENT_SOURCE_DIR})
```
dimitry-, line 23: and change this to ${CMAKE_CURRENT_SOURCE_DIR}/include

OrbitCaptureClient/CaptureListener.h
dimitry-: The idea is to use inculde/ directory for public includes and source directory for private ones. This way the file which is not supposed to be used from the outside will not be on the include path.

kisenshi: I have fixed this, thanks for letting me know and the explanation

OrbitCaptureClient/CMakeLists.txt
kisenshi: Done. I have used {CMAKE_CURRENT_LIST_DIR} instead as it is done in OrbitBase and OrbitLinuxTracing

OrbitCaptureClient/CMakeLists.txt
```diff
@@ -0,0 +1,29 @@
+# Copyright (c) 2020 The Orbit Authors. All rights reserved.
+# Use of this source code is governed by a BSD-style license that can be
+# found in the LICENSE file.
+
+cmake_minimum_required(VERSION 3.15)
```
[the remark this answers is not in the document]
kisenshi: Makes sense; I removed it. Thanks!

kisenshi:
Fixed structure as requested.

PR #892, opened 2020-08-03
Add protobuf for preset

IrinaShkviro:
Convert PresetModule to proto message and add PresetInfo proto message. Use these proto messages for preset serialization instead of cereal.


OrbitCore/OrbitSession.h
```diff
@@ -2,28 +2,14 @@
...
-  Preset();
-  ~Preset();
-
-  ORBIT_SERIALIZABLE;
-
-  std::string m_FileName;
-  std::string m_ProcessFullPath;
-  std::map<std::string, PresetModule> m_Modules;
-};
+#endif  // ORBIT_CORE_ORBIT_SESSION_H_
```
IrinaShkviro, line 15: Probably a good candidate for protobuf message. @dimitry- what do you think?

beckerhe:
Thanks Irina. One question upfront: Is this the last usage of cereal? And if yes, I think we should remove it from cmake and conan. Do you prefer to do it in this PR or a follow-up one?

IrinaShkviro:
> Thanks Irina. One question upfront: Is this the last usage of cereal? And if yes, I think we should remove it from cmake and conan. Do you prefer to do it in this PR or a follow-up one?

Yes, but I wanted to remove it in the next PR.

OrbitGl/App.cpp
```diff
@@ -520,7 +524,7 @@ ErrorMessageOr<void> OrbitApp::OnSavePreset(const std::string& filename) {
 
 //-----------------------------------------------------------------------------
 ErrorMessageOr<void> OrbitApp::ReadPresetFromFile(const std::string& filename,
-                                                  Preset* preset) {
+                                                  PresetInfo* preset_info) {
```
beckerhe, line 527: I think that function should return a `ErrorMessageOr<PresetInfo>` instead of having an out-parameter.

OrbitCore/OrbitSession.h
dimitry-: agree - looks like a proto message to me too.

OrbitGl/App.cpp
```diff
@@ -533,38 +537,40 @@ ErrorMessageOr<void> OrbitApp::ReadPresetFromFile(const std::string& filename,
...
-    ERROR("Loading preset from \"%s\": %s", file_path, e.what());
-    return ErrorMessage(
-        absl::StrFormat("Error reading the preset: %s", e.what()));
+  {
+    SCOPE_TIMER_LOG(absl::StrFormat("Read preset from \"%s\"", file_path));
+    if (!preset_info->ParseFromIstream(&file)) {
+      ERROR("Loading preset from \"%s\" failed", file_path);
+      return ErrorMessage(absl::StrFormat("Error reading the preset"));
+    }
   }
```
beckerhe, line 546: Nit: I don't think that extra scope is needed. The measured difference will be lost in the noise.

OrbitCore/OrbitSession.h
beckerhe: I thought `file_name` is determined at runtime and should not be serialized?

IrinaShkviro: It shouldn't be serialized: I thought about having proto message that we don't want to serialize. In this case we will have all preset-related structs in one place.

OrbitGl/App.cpp
IrinaShkviro: Fixed it, thanks!

OrbitGl/App.cpp
IrinaShkviro: Yes, I see... removed it then.

OrbitCore/OrbitSession.h
```diff
@@ -2,20 +2,14 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
 
-#pragma once
+#ifndef ORBIT_CORE_ORBIT_SESSION_H_
```
dpallotti, line 5: While you're here, this file should probably be renamed to `OrbitPreset.h` (or just `Preset.h` I guess).

IrinaShkviro: I probably want to replace this struct with proto and remove the file in the next PR

IrinaShkviro: OrbitPreset.h was removed in #901

PR #893, opened 2020-08-03
Remove [[nodiscard]] from void SetTimesliceText

dpallotti:
This caused one `warning: attribute 'nodiscard' cannot be applied to functions
without return value` for each .cpp that includes the file.


dimitry-:
also add -Werror=ignored-attributes to STRICT_COMPILE_FLAGS in the root CMakeList.txt

dpallotti:
> also add -Werror=ignored-attributes to STRICT_COMPILE_FLAGS in the root CMakeList.txt

Ah, good point, done!

PR #894, opened 2020-08-03
MainThreadExecutor now uses Qt Event loop

dimitry-:
This commit moves MainThreadExecutor to OrbitGl component
and changes implementation to use Qt mechanism of landing
events/actions on the main thread.


OrbitGl/CaptureSerializerLoadFuzzer.cpp
```diff
@@ -42,7 +42,7 @@ int LLVMFuzzerTestOneInput(uint8_t* buf, size_t len) {
 }
 
 int LLVMFuzzerInitialize(int*, char***) {
-  OrbitApp::Init({});
+  OrbitApp::Init({}, nullptr);
```
dimitry-, line 45: not sure if this is right way - seems like main_thread_executor is not used in this case but I might be wrong.

PR #895, opened 2020-08-03
Fix displaying of function names in the iterator UI overlay

ronaldfw:
When a function was not encountered in any samples, then the (global)
map that we maintain to map function addresses to display names did not
contain that function and the displayed string was incorrect. Whenever
we add an iterator, we insert the function address and display name into
the map now.

### DIFF
--- a/OrbitGl/LiveFunctionsController.cpp
+++ b/OrbitGl/LiveFunctionsController.cpp
@@ -190,10 +190,19 @@ void LiveFunctionsController::OnDeleteButton(uint64_t id) {
 }
 
 void LiveFunctionsController::AddIterator(FunctionInfo* function) {
+  CHECK(function != nullptr);
   uint64_t id = next_iterator_id_;
   ++next_iterator_id_;
 
   auto function_address = FunctionUtils::GetAbsoluteAddress(*function);
+
+  // It's possible that this function does not appear in any samples and
+  // in that case, GAddressToFunctionName would not contain the display
+  // name, so we have to add it here, so we can easily look up the display
+  // name when drawing iterators.
+  std::string function_name = FunctionUtils::GetDisplayName(*function);
+  Capture::GAddressToFunctionName[function_address] = function_name;
+
   const TextBox* box = Capture::GSelectedTextBox;
   // If no box is currently selected or the selected box is a different
   // function, we search for the closest box to the current center of the
--- a/OrbitGl/LiveFunctionsDataView.cpp
+++ b/OrbitGl/LiveFunctionsDataView.cpp
@@ -249,7 +249,7 @@ void LiveFunctionsDataView::OnContextMenu(
   } else if (a_Action == MENU_ACTION_ITERATE) {
     for (int i : a_ItemIndices) {
       FunctionInfo* function = GetFunction(i);
-      if (function->stats().count() > 0) {
+      if (function != nullptr && function->stats().count() > 0) {
         live_functions_->AddIterator(function);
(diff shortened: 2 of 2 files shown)


OrbitGl/LiveFunctionsController.cpp
```diff
@@ -190,10 +190,19 @@ void LiveFunctionsController::OnDeleteButton(uint64_t id) {
...
   ++next_iterator_id_;
 
   auto function_address = FunctionUtils::GetAbsoluteAddress(*function);
+
+  // It's possible that this function does not appear in any samples and
+  // in that case, GAddressToFunctionName would not contain the display
+  // name, so we have to add it here, so we can easily look up the display
+  // name when drawing iterators.
+  std::string function_name = FunctionUtils::GetDisplayName(*function);
+  Capture::GAddressToFunctionName[function_address] = function_name;
```
dimitry-, line 204: I wonder if this can be lazily initialized. We now have 2 places where this global is populated.

karupayun:
Very useful PR! Seems that naming works after loading captures. Thanks Ronald!

OrbitGl/LiveFunctionsController.cpp
ronaldfw: Hmm, you are right, writing to the global from two vastly different places is probably not a good idea. I can see if there is a different way I can do this. I could plumb through the function itself and call GetDisplayName when drawing the overlay (in TimeGraph::DrawOverlay). WDYT?

dimitry-: I think the best way would be to replace GAddressToFunction name with FunctionsInspector (the name is not ideal). The class responsible for looking up function name by address, it can hold address_to_function map as a cache but also implement a slow path of finding this information (and updating the cache). Basically hide this detail from the user. But I do not know how much work is it (there are not many references to this global but there might be additional work involved in implementing slow path for the lookup). 

Are you thinking about adding reference to function to TextBox? This would introduce another dependency on the class we want to remove. I would say if introducing the DataManager/Inspector is too much work - add a todo and merge it as is.

ronaldfw: Seems FunctionInspector could be build on top of StringManager (derive from it, or have that as a member) as the cache and then add the needed methods in addition (like the slow look up). Though the actual implementation is probably not the difficult part.

I do not want to add this to TextBox. I only need the function names in DrawOverlay in TimeGraph and I could just create a class/struct that has a TextBox and a string for the function name and pass that to TimeGraph instead of just the TextBox (in TimeGraph::SetOverlayTextBoxes). That's a fairly local solution, and might be good enough for this case. 

Any thoughts?

dimitry-: Sounds good.

Unrelated question: how does a function that we do not encounter in samples end up in the LiveFuncitonsView?

ronaldfw: Ok, I will change this to just pass the function name in addition to the textbox (assume that's what your "sounds good" refers to?). 

Regarding your question, concrete example how to reproduce: Instrument a function that is called once per frame but does not have a lot of CPU time (example: vkQueuePresent). Take a short capture (longer than one frame, less than a second). It's likely you don't have any callstack samples that contain that function, but it will  show up in "live" as it was called at least once.

ronaldfw:
Since the new solution for this is completely different from what I'm doing here, I'll close this PR and will open a new one.

PR #896, opened 2020-08-03
Make gRPC clients independent from OrbitGl

kisenshi:
### Move CrashManager and ProcessManager to new OrbitClientServices project.

Refactoring code related to gRPC clients so it is independent from OrbitGl. FramePointerValidatorClient depends on App.h so it is not possible to move as the other two.


OrbitClientServices/CMakeLists.txt
```diff
@@ -0,0 +1,26 @@
...
+target_sources(OrbitClientServices PUBLIC 
+        include/OrbitClientServices/CrashManager.h
+        include/OrbitClientServices/ProcessManager.h)
+
+target_sources(OrbitClientServices PRIVATE
+        CrashManager.cpp
+        ProcessManager.cpp)
+
+target_link_libraries(OrbitClientServices PUBLIC 
+        OrbitCore
```
dimitry-, line 25: + OrbitBase?

kisenshi: I tried but it does not build if using OrbitBase instead

dimitry-: github replaced + with dot. I meant in addition to OrbitCore

kisenshi: Ahh, ok; done now. Thanks!

PR #900, opened 2020-08-04
Render ImGui over text.

pierricgimmig:
A bug was recently introduced that would cause text to be rendered on top of 
ImGui windows. Make sure we draw text before the ImGui windows.

b/162485357


pierricgimmig:
> Could you just clarify why changing the drawing order without caring about any z-values is enough in this case? Is z-buffer disabled for UI and Text rendering?

Our imgui implemetation changed recently, I'm guessing that this is when the issue started happening. ImGui disables depth testing (see https://github.com/google/orbit/blob/master/OrbitGl/ImGuiOrbit.cpp#L456) which would explain why text would always be on top if drawn last.  From https://www.khronos.org/opengl/wiki/Depth_Test: "When depth testing is disabled, writes to the depth buffer are also disabled."

PR #901, opened 2020-08-04
replace Preset with PresetFile proto message

IrinaShkviro:
This PR replaces `Preset` struct with `PresetFile` proto message and removes OrbitSession.h. It's not necessary for the serialization (`PresetFile` isn't serialized when we save/load preset), but having all the Preset-related data in one `.proto` file looks better from my perspective.


OrbitClientProtos/preset.proto
```diff
@@ -14,3 +14,8 @@ message PresetInfo {
   string process_full_path = 1;
   map<string, PresetModule> path_to_module = 2;
 }
+
+message PresetFile {
```
dpallotti, line 18: One thing is unclear to me: is this used for serialization?
If not, why does this need to be a proto? Convenience?
If yes (but it doesn't look like it, or is it for the future?), isn't `file_name` redundant (and worse, not in sync with the name of the file in case the file is renamed)?

dpallotti: Ok, sorry, missed the comment.

dpallotti, line 18: So maybe add a comment that this is not used for serialization.

IrinaShkviro: Yes, let's make it clear

PR #906, opened 2020-08-04
Add various "Expand ..." / "Collapse ..." context menu actions to the top-down view

dpallotti:
#### Add "Expand all", "Collapse all" context menu actions to TopDownWidget
Bug: http://b/162704044
#### Allow multiple selections in TopDownWidget
...and hence multiple nodes to be "expanded all" or "collapsed all" at once.
#### Use header()->resizeSections instead of a loop on resizeColumnToContents
This doesn't change the functionality, only shortens the code.
#### Rename "Expand / Collapse all" to "Expand / Collapse recursively"
"Expand / Collapse all" seems to usually refer to the entire tree.
#### Add "Collapse children recursively" to TopDownWidget

### DIFF
--- a/OrbitQt/topdownwidget.cpp
+++ b/OrbitQt/topdownwidget.cpp
@@ -1,0 +1,21 @@
+// Copyright (c) 2020 The Orbit Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#include "topdownwidget.h"
+
+#include "TopDownViewItemModel.h"
+
+void TopDownWidget::SetTopDownView(std::unique_ptr<TopDownView> top_down_view) {
+  auto* model =
+      new TopDownViewItemModel{std::move(top_down_view), ui_->topDownTreeView};
+  auto* proxy_model = new QSortFilterProxyModel{ui_->topDownTreeView};
+  proxy_model->setSourceModel(model);
+  proxy_model->setSortRole(Qt::EditRole);
+  ui_->topDownTreeView->setModel(proxy_model);
+  ui_->topDownTreeView->sortByColumn(TopDownViewItemModel::kInclusive,
+                                     Qt::DescendingOrder);
+  for (int column = 0; column < TopDownViewItemModel::kColumnCount; ++column) {
+    ui_->topDownTreeView->resizeColumnToContents(column);
+  }
+}
--- a/OrbitQt/topdownwidget.cpp
+++ b/OrbitQt/topdownwidget.cpp
@@ -4,6 +4,8 @@
 
 #include "topdownwidget.h"
 
+#include <QMenu>
+
 #include "TopDownViewItemModel.h"
 
 void TopDownWidget::SetTopDownView(std::unique_ptr<TopDownView> top_down_view) {
@@ -17,5 +19,64 @@ void TopDownWidget::SetTopDownView(std::unique_ptr<TopDownView> top_down_view) {
                                      Qt::DescendingOrder);
   for (int column = 0; column < TopDownViewItemModel::kColumnCount; ++column) {
     ui_->topDownTreeView->resizeColumnToContents(column);
+  }
+}
+
+const std::string TopDownWidget::kActionExpandAll = "Expand all";
+const std::string TopDownWidget::kActionCollapseAll = "Collapse all";
+
+static void ExpandRecursively(QTreeView* tree_view, const QModelIndex& index) {
+  if (!index.isValid()) {
+    return;
+  }
+  for (int i = 0; i < index.model()->rowCount(index); ++i) {
+    const QModelIndex& child = index.child(i, 0);
+    ExpandRecursively(tree_view, child);
+  }
+  if (!tree_view->isExpanded(index)) {
+    tree_view->expand(index);
+  }
+}
+
+static void CollapseRecursively(QTreeView* tree_view,
+                                const QModelIndex& index) {
+  if (!index.isValid()) {
+    return;
+  }
+  for (int i = 0; i < index.model()->rowCount(index); ++i) {
+    const QModelIndex& child = index.child(i, 0);
+    CollapseRecursively(tree_view, child);
+  }
+  if (tree_view->isExpanded(index)) {
+    tree_view->collapse(index);
+  }
+}
+
+void TopDownWidget::onCustomContextMenuRequested(const QPoint& point) {
+  QModelIndex index = ui_->topDownTreeView->indexAt(point);
+  if (!index.isValid()) {
+    return;
+  }
+  QMenu menu{ui_->topDownTreeView};
+
+  if (index.model()->rowCount(index) > 0) {
+    // Always show "Expand all", as even if the selected node is expanded there
+    // could be subtrees not expanded, but only show "Collapse all" when the
+    // selected node is expanded, as it would otherwise be unintuitive to
+    // collapse subtrees none of which is visible.
+    menu.addAction(kActionExpandAll.c_str());
+    if (ui_->topDownTreeView->isExpanded(index)) {
+      menu.addAction(kActionCollapseAll.c_str());
+    }
+  }
+
+  QAction* action = menu.exec(ui_->topDownTreeView->mapToGlobal(point));
+  if (action == nullptr) {
+    return;
+  }
+  if (action->text().toStdString() == kActionExpandAll) {
+    ExpandRecursively(ui_->topDownTreeView, index);
+  } else if (action->text().toStdString() == kActionCollapseAll) {
+    CollapseRecursively(ui_->topDownTreeView, index);
   }
 }
--- a/OrbitQt/topdownwidget.cpp
+++ b/OrbitQt/topdownwidget.cpp
@@ -17,9 +17,7 @@ void TopDownWidget::SetTopDownView(std::unique_ptr<TopDownView> top_down_view) {
   ui_->topDownTreeView->setModel(proxy_model);
   ui_->topDownTreeView->sortByColumn(TopDownViewItemModel::kInclusive,
                                      Qt::DescendingOrder);
-  for (int column = 0; column < TopDownViewItemModel::kColumnCount; ++column) {
-    ui_->topDownTreeView->resizeColumnToContents(column);
-  }
+  ui_->topDownTreeView->header()->resizeSections(QHeaderView::ResizeToContents);
 }
 
 const std::string TopDownWidget::kActionExpandAll = "&Expand all";
--- a/OrbitQt/topdownwidget.cpp
+++ b/OrbitQt/topdownwidget.cpp
@@ -24,6 +24,8 @@ const std::string TopDownWidget::kActionExpandRecursively =
     "&Expand recursively";
 const std::string TopDownWidget::kActionCollapseRecursively =
     "&Collapse recursively";
+const std::string TopDownWidget::kActionExpandAll = "Expand all";
+const std::string TopDownWidget::kActionCollapseAll = "Collapse all";
 
 static void ExpandRecursively(QTreeView* tree_view, const QModelIndex& index) {
   if (!index.isValid()) {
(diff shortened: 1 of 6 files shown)


OrbitQt/topdownwidget.h
```diff
@@ -0,0 +1,38 @@
...
+#include "ui_topdownwidget.h"
+
+class TopDownWidget : public QWidget {
+  Q_OBJECT
+
+ public:
+  explicit TopDownWidget(QWidget* parent = nullptr)
+      : QWidget{parent}, ui_{std::make_unique<Ui::TopDownWidget>()} {
+    ui_->setupUi(this);
+    connect(ui_->topDownTreeView,
```
pierricgimmig, line 21: Qt's new signal-slot syntax is preferable: https://wiki.qt.io/New_Signal_Slot_Syntax

pierricgimmig:
Hi @dpallotti , thanks for this, code looks good but please see comment about signal-slot syntax. 

Concerning functionality, What would you think of the "collapse all" applying to child nodes only. I'd prefer it that way actually, because you can already close the entire node by clicking on the triangle toggle. It's useful to see the first level of functions under a specific node, with the current implementation, I'd have to "collapse all" then reopen the node.

OrbitQt/topdownwidget.cpp
```diff
@@ -0,0 +1,104 @@
...
+  auto* model =
+      new TopDownViewItemModel{std::move(top_down_view), ui_->topDownTreeView};
+  auto* proxy_model = new QSortFilterProxyModel{ui_->topDownTreeView};
+  proxy_model->setSourceModel(model);
+  proxy_model->setSortRole(Qt::EditRole);
+  ui_->topDownTreeView->setModel(proxy_model);
+  ui_->topDownTreeView->sortByColumn(TopDownViewItemModel::kInclusive,
+                                     Qt::DescendingOrder);
+  for (int column = 0; column < TopDownViewItemModel::kColumnCount; ++column) {
+    ui_->topDownTreeView->resizeColumnToContents(column);
```
pierricgimmig, line 21: This doesn't seem to work for the name column on my side, unless there is a maximum width specified somewhere?

OrbitQt/topdownwidget.h
dpallotti: Done, thank you.

OrbitQt/topdownwidget.cpp
dpallotti: I think that this works, but not like we'd like because the resizing is done based on the visible elements, which when this code is called are only the thread names. I'm not sure what a good way of doing this is.

dpallotti:
> Concerning functionality, What would you think of the "collapse all" applying to child nodes only. I'd prefer it that way actually, because you can already close the entire node by clicking on the triangle toggle. It's useful to see the first level of functions under a specific node, with the current implementation, I'd have to "collapse all" then reopen the node.

I think we're looking at the problem slightly differently: I was thinking about the "Collapse all" as a way to close the entire subtree without having the problem of it completely re-expanding once the node is expanded. With your suggestion, to get that one has to perform two operations&mdash;collapse the children and then collapse the node.
In general I don't dislike the functionality, but I think it's not what one would expect as it's not symmetrical with "Expand all". So if we want it, we should rename it to "Collapse children recursively" or just add an additional entry with that name.

Speaking of naming, maybe these actions should indeed be called "Expand recursively" and "Collapse recursively". It seems that the common behavior of "Expand all" and "Collapse all" in other applications is to expand or collapse the entire tree, regardless of the selected item(s).

dpallotti:
> > Concerning functionality, What would you think of the "collapse all" applying to child nodes only. I'd prefer it that way actually, because you can already close the entire node by clicking on the triangle toggle. It's useful to see the first level of functions under a specific node, with the current implementation, I'd have to "collapse all" then reopen the node.
> 
> I think we're looking at the problem slightly differently: I was thinking about the "Collapse all" as a way to close the entire subtree without having the problem of it completely re-expanding once the node is expanded. With your suggestion, to get that one has to perform two operations—collapse the children and then collapse the node.
> In general I don't dislike the functionality, but I think it's not what one would expect as it's not symmetrical with "Expand all". So if we want it, we should rename it to "Collapse children recursively" or just add an additional entry with that name.
> 
> Speaking of naming, maybe these actions should indeed be called "Expand recursively" and "Collapse recursively". It seems that the common behavior of "Expand all" and "Collapse all" in other applications is to expand or collapse the entire tree, regardless of the selected item(s).

@pierricgimmig Based on this, I changed "Expand all" / "Collapse all" to "Expand recursively", "Collapse recursively", added "Collapse children recursively", added "Expand all" / Collapse all" to actually expand and collapse everything. Let me know what you think and if it's too many actions.

OrbitQt/topdownwidget.cpp
beckerhe: You can use QHeaderView to achieve what you want (probably). Have a look in one of my branches: https://github.com/beckerhe/orbitprofiler/blob/9e362d1d4611c88ada629c57002cb71ca74b0227/OrbitQt/orbitmainwindow.cpp#L270

beckerhe: And a doc link: https://doc.qt.io/qt-5/qheaderview.html#setSectionResizeMode-1

OrbitQt/topdownwidget.cpp
```diff
@@ -0,0 +1,151 @@
...
+const std::string TopDownWidget::kActionExpandRecursively =
+    "&Expand recursively";
+const std::string TopDownWidget::kActionCollapseRecursively =
+    "&Collapse recursively";
+const std::string TopDownWidget::kActionCollapseChildrenRecursively =
+    "Collapse children recursively";
+const std::string TopDownWidget::kActionExpandAll = "Expand all";
+const std::string TopDownWidget::kActionCollapseAll = "Collapse all";
+
+static void ExpandRecursively(QTreeView* tree_view, const QModelIndex& index) {
```
beckerhe, line 32: Maybe I missed it, but is there a reason not to use QTreeView::expandRecursively?

OrbitQt/topdownwidget.cpp
```diff
@@ -0,0 +1,151 @@
...
+  if (!index.isValid()) {
+    return;
+  }
+  for (int i = 0; i < index.model()->rowCount(index); ++i) {
+    const QModelIndex& child = index.child(i, 0);
+    CollapseRecursively(tree_view, child);
+  }
+}
+
+static void ExpandAll(QTreeView* tree_view) {
```
beckerhe, line 70: As before. This seems to do the same as QTreeView::expandAll, isn't it?

OrbitQt/topdownwidget.cpp
```diff
@@ -0,0 +1,151 @@
...
+  for (int i = 0; i < tree_view->model()->rowCount(); ++i) {
+    const QModelIndex& child = tree_view->model()->index(i, 0);
+    ExpandRecursively(tree_view, child);
+  }
+}
+
+static void CollapseAll(QTreeView* tree_view) {
```
beckerhe, line 77: QTreeView::collapseAll?

OrbitQt/topdownwidget.h
```diff
@@ -18,11 +18,20 @@ class TopDownWidget : public QWidget {
...
+            SLOT(onCustomContextMenuRequested(const QPoint&)));
   }
 
   void SetTopDownView(std::unique_ptr<TopDownView> top_down_view);
 
+ private slots:
+  void onCustomContextMenuRequested(const QPoint& point);
+
  private:
+  static const std::string kActionExpandAll;
```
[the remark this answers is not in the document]
beckerhe: Edit: After seeing the rest of the code: Using QStringView or QStringLiteral would avoid some conversions.

OrbitQt/topdownwidget.ui
```diff
@@ -0,0 +1,32 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<ui version="4.0">
```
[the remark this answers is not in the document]
dpallotti: Exactly, the idea was to add a `QLineEdit` for filtering but then I delayed that as things got trickier than expected. So I'll leave it as the plan would be to get to it soon.

OrbitQt/topdownwidget.cpp
dpallotti: I had actually tried that but if I use `Stretch` the behavior is weird. The section is not manually resizable anymore. You can resize it indirectly by resizing the other ones but as it's the first section (or rather, not the last) the overall behavior is annoying.

OrbitQt/topdownwidget.h
```diff
@@ -18,11 +18,20 @@ class TopDownWidget : public QWidget {
   explicit TopDownWidget(QWidget* parent = nullptr)
       : QWidget{parent}, ui_{std::make_unique<Ui::TopDownWidget>()} {
     ui_->setupUi(this);
+    connect(ui_->topDownTreeView,
+            SIGNAL(customContextMenuRequested(const QPoint&)), this,
```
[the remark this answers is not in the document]
dpallotti: Yep, this is done in a subsequent commit as Pierric also pointed that out.

OrbitQt/topdownwidget.cpp
dpallotti: You caught me off guard for a moment, as I had tried and I didn't remember, but the reason is that it should be there from Qt 5.13 but on my Linux machine I can only get Qt up to 5.12.5... So either I leave it like this or I'll need help.

dpallotti: Also there's no `QTreeView::collapseRecursively` so might as well implement the expand version ourselves as well.

OrbitQt/topdownwidget.h
```diff
@@ -0,0 +1,29 @@
+// Copyright (c) 2020 The Orbit Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#ifndef ORBIT_QT_TOP_DOWN_WIDGET_H_
+#define ORBIT_QT_TOP_DOWN_WIDGET_H_
+
+#include <QSortFilterProxyModel>
```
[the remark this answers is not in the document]
dpallotti: Done.

OrbitQt/topdownwidget.h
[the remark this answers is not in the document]
dpallotti: I went with `QString`, using the `QStringLiteral` macro to create the value. I'm not sure why you suggested `string_view`/`QStringView` here though. Aren't those supposed to be used as parameters to receive strings they don't own?

OrbitQt/topdownwidget.cpp
```diff
@@ -65,36 +67,38 @@ void TopDownWidget::onCustomContextMenuRequested(const QPoint& point) {
...
+        enable_collapse_recursively = true;
       }
     }
   }
 
   QMenu menu{ui_->topDownTreeView};
-  if (enable_expand_all) menu.addAction(kActionExpandAll.c_str());
-  if (enable_collapse_all) menu.addAction(kActionCollapseAll.c_str());
+  if (enable_expand_recursively)
+    menu.addAction(kActionExpandRecursively.c_str());
```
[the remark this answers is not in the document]
dpallotti: The constant is now a `QString`.

pierricgimmig:
> I think we're looking at the problem slightly differently: I was thinking about the "Collapse all" as a way to close the entire subtree without having the problem of it completely re-expanding once the node is expanded. With your suggestion, to get that one has to perform two operations—collapse the children and then collapse the node.
> In general I don't dislike the functionality, but I think it's not what one would expect as it's not symmetrical with "Expand all". So if we want it, we should rename it to "Collapse children recursively" or just add an additional entry with that name.
> 
> Speaking of naming, maybe these actions should indeed be called "Expand recursively" and "Collapse recursively". It seems that the common behavior of "Expand all" and "Collapse all" in other applications is to expand or collapse the entire tree, regardless of the selected item(s).

Having another entry, "Collapse/Expand Children (recursively)", would do the trick yeah, doesn't need to be in this PR. Concerning the naming, "Collapse Recursively' vs "Collapse All", i'm ok with either.

PR #907, opened 2020-08-04
Fix displaying of function names in the iterator UI overlay

ronaldfw:
When a function was not encountered in any samples, then the (global)
map that we maintain to map function addresses to display names did not
contain that function and the displayed string was incorrect. Whenever
we add an iterator, we insert the function address and display name into
the map now.

### DIFF
--- a/OrbitGl/TimeGraph.cpp
+++ b/OrbitGl/TimeGraph.cpp
@@ -99,7 +99,7 @@ void TimeGraph::Clear() {
   // The process track is a special ThreadTrack of id "0".
   process_track_ = GetOrCreateThreadTrack(0);
 
-  SetOverlayTextBoxes({});
+  SetIteratorOverlayData({}, {});
 
   NeedsUpdate();
 }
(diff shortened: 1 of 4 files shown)


OrbitGl/TimeGraph.cpp
```diff
@@ -626,12 +626,10 @@ void TimeGraph::Draw(GlCanvas* canvas, PickingMode picking_mode) {
 
 namespace {
 
-std::string GetLabelBetweenIterators(const TextBox* box_a,
-                                     const TextBox* box_b) {
-  std::string function_from =
-      Capture::GAddressToFunctionName[box_a->GetTimerInfo().function_address()];
-  std::string function_to =
-      Capture::GAddressToFunctionName[box_b->GetTimerInfo().function_address()];
+std::string GetLabelBetweenIterators(const FunctionInfo& function_a,
```
dimitry-, line 629: nit: this should probably be [[nodiscard]]

ronaldfw: Yes, thanks for spotting this!

PR #908, opened 2020-08-04
 Update non-thread-safe data on the main thread

dimitry-:
For global maps which are currently used from the main thread
    we should make sure that during capturing they are updated on
    the main thread as well, to avoid undefined behavior.
    
    This change moves updates of capture data to the main thread.
    It also introduces OnCaptureStarted and OnCaptureComplete
    callbacks to the listener.

### DIFF
--- a/OrbitCore/SamplingProfiler.cpp
+++ b/OrbitCore/SamplingProfiler.cpp
@@ -69,7 +69,7 @@ std::multimap<int, CallstackID> SamplingProfiler::GetCallstacksFromAddress(
 }
 
 //-----------------------------------------------------------------------------
-void SamplingProfiler::AddCallStack(CallstackEvent& callstack_event) {
+void SamplingProfiler::AddCallStack(CallstackEvent callstack_event) {
   CallstackID hash = callstack_event.callstack_hash();
   if (!HasCallStack(hash)) {
     std::shared_ptr<CallStack> callstack =
@@ -77,14 +77,14 @@ void SamplingProfiler::AddCallStack(CallstackEvent& callstack_event) {
     AddUniqueCallStack(*callstack);
   }
 
-  m_Callstacks.push_back(callstack_event);
+  m_Callstacks.push_back(std::move(callstack_event));
 }
 
 //-----------------------------------------------------------------------------
-void SamplingProfiler::AddUniqueCallStack(CallStack& a_CallStack) {
+void SamplingProfiler::AddUniqueCallStack(CallStack call_stack) {
   absl::MutexLock lock(&unique_callstacks_mutex_);
-  unique_callstacks_[a_CallStack.Hash()] =
-      std::make_shared<CallStack>(a_CallStack);
+  auto key = call_stack.Hash();
+  unique_callstacks_[key] = std::make_shared<CallStack>(std::move(call_stack));
 }
 
 const CallStack& SamplingProfiler::GetResolvedCallstack(
--- a/OrbitGl/App.cpp
+++ b/OrbitGl/App.cpp
@@ -93,6 +93,51 @@ std::string OrbitApp::FindFile(const std::string& caption,
   return std::string();
 }
 
+void OrbitApp::OnCaptureStarted() {
+  // We need to block until initialization is complete to
+  // avoid races when capture thread start processing data.
+  absl::Mutex mutex;
+  absl::MutexLock mutex_lock(&mutex);
+  bool initialization_complete = false;
+  captured_address_infos_.clear();
+
+  main_thread_executor_->Schedule([this, &initialization_complete, &mutex] {
+    ClearCapture();
+
+    if (capture_started_callback_) {
+      capture_started_callback_();
+    }
+
+    if (!Capture::GSelectedFunctionsMap.empty() && select_live_tab_callback_) {
+      select_live_tab_callback_();
+    }
+
+    absl::MutexLock lock(&mutex);
+    initialization_complete = true;
+  });
+
+  bool (*IsTrue)(bool*) = [](bool* value) { return *value; };
+  mutex.Await(absl::Condition(IsTrue, &initialization_complete));
+}
+
+void OrbitApp::OnCaptureComplete() {
+  main_thread_executor_->Schedule([this] {
+    Capture::GAddressInfos = std::move(captured_address_infos_);
+    Capture::FinalizeCapture();
+
+    RefreshCaptureView();
+
+    AddSamplingReport(Capture::GSamplingProfiler);
+    AddTopDownView(*Capture::GSamplingProfiler);
+
+    if (capture_stopped_callback_) {
+      capture_stopped_callback_();
+    }
+
+    FireRefreshCallbacks();
+  });
+}
+
 void OrbitApp::OnTimer(const TimerInfo& timer_info) {
   GCurrentTimeGraph->ProcessTimer(timer_info);
 }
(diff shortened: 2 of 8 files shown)


pierricgimmig:
This will have big performance implication and slow down the application quite a bit when streaming in timers. It's always been the goal to process timers and events as fast as possible without blocking the main thread. Through the use of the blockchain data structure, which never migrates elements, a worker thread can append timers while the rendering/main thread displays them without locking.  We are talking 100K timers per second here, scheduling each of them individually seems overkill to me.

> Since we are using global variables to store capture data, we should make sure that access to that data is happening on the single thread. 

I don't see how the presence of global variables entails the single threaded approach. Also, I think we should move away from globals.

dimitry-:
I agree that we need to move away from globals, but this is much larger scope - I am currently trying to fix races in capture process. 

For context: CaptureStarted/Complete are introduced to start solving races around Capture::State in order to do that I am moving lifecycle inside CaptureClient and this is the first step for this process.

I moved thread-safe operations back to capture thread, this requires CaptureStarted to block the capture thread until initialization is done on the main thread. ptal.

OrbitGl/App.cpp
```diff
@@ -111,6 +111,52 @@ void OrbitApp::SetCommandLineArguments(const std::vector<std::string>& a_Args) {
   }
 }
 
+static bool IsTrue(bool* flag) { return *flag; }
```
dpallotti, line 114: I would prefer the use of lambdas in `absl::Condition`, to avoid these very generic tiny static functions.

OrbitGl/App.cpp
```diff
@@ -120,35 +166,47 @@ void OrbitApp::OnKeyAndString(uint64_t key, std::string str) {
 }
 
 void OrbitApp::OnCallstack(CallStack callstack) {
-  Capture::GSamplingProfiler->AddUniqueCallStack(callstack);
+  main_thread_executor_->Schedule([callstack = std::move(callstack)] {
```
dpallotti, line 169: I have to agree that this seems like *a lot* of scheduling actions if a new one is scheduled for each event, even if this is not done for `Timer`s which are the most frequent ones. Not sure how a reasonably quick fix would look like though. But as this is not directly related to the issue with `CaptureClient` (right?), was this actually causing problems?

OrbitGl/App.cpp
```diff
@@ -111,6 +111,52 @@ void OrbitApp::SetCommandLineArguments(const std::vector<std::string>& a_Args) {
   }
 }
 
+static bool IsTrue(bool* flag) { return *flag; }
+
+void OrbitApp::OnCaptureStarted() {
```
dpallotti, line 116: So this is *not* scheduled on the main thread and has to wait for the scheduled action to complete on the main thread, right?

OrbitGl/App.cpp
ronaldfw: This should at least be measured to understand the performance impact, o/w it seems to be hard to make a decision here.

OrbitGl/App.cpp
dimitry-: right.

OrbitGl/App.cpp
dimitry-: I have not seen crashes because of this - but this is obviously not thread_safe so I think we need to fix it.

dimitry-: ok, I made sampling profiler thread-safe, lets try again.

dimitry-:
sorry, I had to force-push this - ptal.

OrbitGl/App.cpp
pierricgimmig: This I don't understand, the capture start is initiated from the UI, so on the main thread, shouldn't OnCaptureStarted() already be on the main thread? 

In OrbitApp::StartCapture(), on the main thread, we schedule the call to capture_client_->Capture() on a worker thread. We then in turn call "OnCaptureStarted" from the worker thread which needs to schedule some work on the main thread. Are those hops necessary? Why can't we call capture_client_->Capture() from the main thread?

dimitry-: So the end-goal is to move state from Capture to CaptureClient, in order for this to be consitent it will need to be protected by mutex and we want to avoid making the mutex and implementation details for synchronization global. This is why I added CaptureStarted/Complete as part of Capture lifecycle. Note that this will also improve error handling, - for example in the case of unsuccessful capture you do not end up cleaning up previous results.

We do not call anything that involves blocking I/O on the main thread in order to avoid freezes to UI. CaptureClient::Capture calls streaming grpc Read which could block until next event has arrived.

pierricgimmig: Ah yes, thanks for the precision.

OrbitCore/SamplingProfiler.cpp
```diff
@@ -61,22 +61,22 @@ std::multimap<int, CallstackID> SamplingProfiler::GetCallstacksFromAddress(
 }
 
 //-----------------------------------------------------------------------------
-void SamplingProfiler::AddCallStack(CallstackEvent& callstack_event) {
+void SamplingProfiler::AddCallStack(CallstackEvent callstack_event) {
```
pierricgimmig, line 64: Why pass argument by value here? It seems we copy/move on the caller side, then again in this function.

dimitry-: reference leads to one extra-copy in this case: https://g3doc.corp.google.com/devtools/library_club/g3doc/totw/117.md?cl=head

pierricgimmig: @dimitry- , @dpallotti , this still doesn't make a whole lot of sense to me. If the object being passed by value has some POD data, then we end up copying this static data twice. Shouldn't AddCallstack take in a rvalue reference as argument? This way we save a copy if the caller std::moves the argument.  If not, then the caller has to create a temporary which makes the extra copy explicit.

dpallotti: You are correct about POD. Indeed, when the argument is fully POD you get "Clang-Tidy: std::move of the variable '...' of the trivially-constructible type '...' has no effect; remove std::move()".
But passing by const-ref or using the pass-by-valye-and-move pattern doesn't really make a significant difference with just a few POD fields. But you don't want many POD fields anyway, otherwise when you pass an r-value you are not able to move them, and be it that you copy once (passing an r-value reference you can't move from or a const-ref) or twice (passing by value), it's still too much copying. Instead you want to be able to move that data, by storing it inside the class with smart pointers, vectors, strings, etc, all of which can be moved. 
So the fact that you don't have too much POD data to copy regardless of whether you are copying or moving is pretty much an assumption. At this point, using pass-by-value-and-move is about adhering to a pattern that with a lot of movable data becomes quite relevant. If we always use it, then we also use it when we should use it.

You mention rvalue references. That would be the ideal parameter to pass r-values. But what if you want to pass an l-value (and make a copy of it)? Then you need the two overloads, one for the r-value, one for the l-value. That's an extra method and extra code. But what if you have two parameters (both of which to get into the class, are not just read)? Then you need four overloads, for the different combinations. And so on, that's not sustainable.
So what you do instead is have the parameter(s) by value and move from them. If you pass an l-value, the copy happens when you pass, and then you have the move in the method&mdash;if you were passing by const-ref, you would not have the move but you would still have the copy in the method. But if you pass an r-value, you have two moves and no copies&mdash;if you were passing by const-ref, you would not have the first move but then you would have to copy in the method, as you obviously can't move from a const-ref.
So l-value: one copy + one move instead of one copy. But r-value: **two moves instead of one copy**.

pierricgimmig: > You mention rvalue references. That would be the ideal parameter to pass r-values. But what if you want to pass an l-value (and make a copy of it)? Then you need the two overloads, one for the r-value, one for the l-value.

Well in that case you can create a temporary object at the call site without having to declare an extra method.

> **the fact that the move is cheap is an assumption**.
This is exactly my point, yes. You have to know about the type you're passing, in some cases there might be expensive copies of static data.

> So l-value: one copy + one move instead of one copy. But r-value: **two moves instead of one copy**.

But if you have a r-value as argument, you have only one move.

See https://repl.it/repls/UltimateOrangeredWebpages#main.cpp

OrbitGl/App.h
```diff
@@ -302,6 +303,10 @@ class OrbitApp final : public DataViewFactory, public CaptureListener {
   const SymbolHelper symbol_helper_;
 
   std::unique_ptr<FramePointerValidatorClient> frame_pointer_validator_client_;
+
+  // Temporary objects used by CaptureListener implementation
+  std::unordered_map<uint64_t, orbit_client_protos::LinuxAddressInfo>
```
pierricgimmig, line 308: Should this be a absl::flat_hash_map?

dimitry-:
> Hi Dimitry, LGTM, thanks for getting rid of the scheduling of address_infos. The only thing is that callstack_event argument being passed by value and incurring one extra move. Let's discuss it offline.

Submitted this one, created #945 as a foolow up, we can continue discussion there.

OrbitCore/SamplingProfiler.cpp
dpallotti: > Well in that case you can create a temporary object at the call site without having to declare an extra method.

That gets very annoying very quickly in general.

> You have to know about the type you're passing, in some cases there might be expensive copies of static data.

Is there expensive that to pass in this case though? This is a proto. I would guess moves are cheap.

PR #911, opened 2020-08-05
Synchronize access to unique_callstacks_ in SamplingProfiler (mostly)

dimitry-:
This map is modified from the capture thread while accessed from the
main thread which could cause undefined behavior. This commit
introduces a Mutex that guards access to unique_callstacks_ map
Note that this is not final solution it is rather temporary fix
until we move to more safe Capture Model.

### DIFF
--- a/OrbitCore/SamplingProfiler.h
+++ b/OrbitCore/SamplingProfiler.h
@@ -74,10 +74,12 @@ class SamplingProfiler {
   void AddUniqueCallStack(CallStack& a_CallStack);
 
   std::shared_ptr<CallStack> GetCallStack(CallstackID a_ID) {
-    return m_UniqueCallstacks.at(a_ID);
+    absl::MutexLock lock(&unique_callstacks_mutex_);
+    return unique_callstacks_.at(a_ID);
   }
   bool HasCallStack(CallstackID a_ID) {
-    return m_UniqueCallstacks.count(a_ID) > 0;
+    absl::MutexLock lock(&unique_callstacks_mutex_);
+    return unique_callstacks_.count(a_ID) > 0;
   }
 
   const CallStack& GetResolvedCallstack(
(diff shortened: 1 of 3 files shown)


OrbitGl/CaptureSerializer.cpp
```diff
@@ -79,16 +79,16 @@ void CaptureSerializer::FillCaptureData(CaptureInfo* capture_info) {
...
-    CallstackInfo* callstack = capture_info->add_callstacks();
-    *callstack->mutable_data() = {hashed_callstack->m_Data.begin(),
-                                  hashed_callstack->m_Data.end()};
-  }
+  // TODO: this is not really synchronized, since GetCallstacks processing below
+  // is not under the same mutex lock we could end up having list of callstacks
+  // inconsistent with unique_callstacks. Revisit sampling profiler data
+  // thread-safety.
+  Capture::GSamplingProfiler->ForEachUniqueCallstack(
+      [&capture_info](const CallStack* call_stack) {
```
IrinaShkviro, line 87: Shouldn't be const CallStack& call_stack?

dimitry-: done.

OrbitCore/SamplingProfiler.h
```diff
@@ -92,9 +94,12 @@ class SamplingProfiler {
     return &m_Callstacks;
   }
 
-  const std::unordered_map<CallstackID, std::shared_ptr<CallStack>>&
-  GetUniqueCallstacks() const {
-    return m_UniqueCallstacks;
+  template <typename F>
```
dpallotti, line 97: Why the template instead of passing a `std::function<void(const CallStack&)>`?

dimitry-: habbit from before std::function. fixed.

OrbitCore/SamplingProfiler.h
```diff
@@ -92,9 +94,11 @@ class SamplingProfiler {
     return &m_Callstacks;
   }
 
-  const std::unordered_map<CallstackID, std::shared_ptr<CallStack>>&
-  GetUniqueCallstacks() const {
-    return m_UniqueCallstacks;
+  void ForEachUniqueCallstack(std::function<void(const CallStack&)> action) {
```
dpallotti, line 97: Same line as before, sorry... Can `action` be passed as const-ref?

dimitry-: done.

PR #912, opened 2020-08-06
Fix float/double conversions breaking GCC compilation.

pierricgimmig:
Build was broken with GCC:

/home/pierric/git/orbit/OrbitGl/CaptureWindow.cpp: In member function ‘virtual void CaptureWindow::MouseWheelMoved(int, int, int, bool)’:
/home/pierric/git/orbit/OrbitGl/CaptureWindow.cpp:399:37: error: conversion from ‘double’ to ‘float’ may change value [-Werror=float-conversion]
  399 |     time_graph_.VerticalZoom(delta, mouse_relative_y_position);
      |                                     ^~~~~~~~~~~~~~~~~~~~~~~~~


pierricgimmig:
Thanks @dimitry- . We should probably enable the -Werror=float-conversion equivalent for our clang build...

dimitry-:
It is enabled - I think clang and gcc disagree on which conversion deserves a warning. gcc seems to be more strict.

PR #913, opened 2020-08-06
Remove unused code: BlackBoard.h/cpp and Card.h/cpp.

ronaldfw:
All related locations and instances where these were used are also
changed (including removing the "visualize" panel).


ronaldfw:
As I've discussed this with @pierricgimmig over chat, I'll merge without further approval from him. But let me know if you changed your mind :)

PR #914, opened 2020-08-06
Add "searching" to the top-down view

dpallotti:
This builds on the assumption that I don't want to actually hide (i.e., filter
out) entire subtrees if the filter is not matched—I find that confusing.
Instead, I want to expand enough so that the matching nodes are visible, but
collapse everything else. See this as a search functionality rather than a
filter functionality.

---

Let me know it this can make sense.

### DIFF
--- a/OrbitQt/topdownwidget.cpp
+++ b/OrbitQt/topdownwidget.cpp
@@ -4,15 +4,65 @@
 
 #include "topdownwidget.h"
 
+#include <QColor>
 #include <QMenu>
 #include <QSortFilterProxyModel>
 
 #include "TopDownViewItemModel.h"
+#include "absl/container/flat_hash_set.h"
+
+class HighlightingSortFilterProxyModel : public QSortFilterProxyModel {
+ public:
+  explicit HighlightingSortFilterProxyModel(QObject* parent)
+      : QSortFilterProxyModel{parent} {}
+
+  // Specify a set where this ProxyModel should put internalPointers of indices
+  // that match the current filter.
+  void SetFilterAcceptedNodeCollectorSet(
+      absl::flat_hash_set<void*>* filter_accepted_nodes_collector) {
+    filter_accepted_nodes_collector_ = filter_accepted_nodes_collector;
+  }
+
+  // Specify a set of internalPointers whose indices should be highlighted.
+  void SetNodesToHighlightSet(
+      std::unique_ptr<absl::flat_hash_set<void*>> nodes_to_highlight) {
+    nodes_to_highlight_ = std::move(nodes_to_highlight);
+  }
+
+  QVariant data(const QModelIndex& index, int role) const override {
+    if (role == Qt::ForegroundRole && nodes_to_highlight_ != nullptr) {
+      QModelIndex source_index = mapToSource(index);
+      if (nodes_to_highlight_->contains(source_index.internalPointer())) {
+        return QColor{Qt::green};
+      }
+    }
+    return QSortFilterProxyModel::data(index, role);
+  }
+
+ protected:
+  bool filterAcceptsRow(int source_row,
+                        const QModelIndex& source_parent) const override {
+    bool accepts_row =
+        QSortFilterProxyModel::filterAcceptsRow(source_row, source_parent);
+    if (accepts_row && filter_accepted_nodes_collector_ != nullptr) {
+      QModelIndex source_index =
+          sourceModel()->index(source_row, 0, source_parent);
+      filter_accepted_nodes_collector_->insert(source_index.internalPointer());
+    }
+    return accepts_row;
+  }
+
+ private:
+  mutable absl::flat_hash_set<void*>* filter_accepted_nodes_collector_ =
+      nullptr;
+  std::unique_ptr<absl::flat_hash_set<void*>> nodes_to_highlight_ = nullptr;
+};
 
 void TopDownWidget::SetTopDownView(std::unique_ptr<TopDownView> top_down_view) {
   auto* model =
       new TopDownViewItemModel{std::move(top_down_view), ui_->topDownTreeView};
-  auto* proxy_model = new QSortFilterProxyModel{ui_->topDownTreeView};
+  auto* proxy_model =
+      new HighlightingSortFilterProxyModel{ui_->topDownTreeView};
   proxy_model->setSourceModel(model);
   proxy_model->setSortRole(Qt::EditRole);
 
@@ -171,12 +221,14 @@ void TopDownWidget::onCustomContextMenuRequested(const QPoint& point) {
 }
 
 void TopDownWidget::on_searchLineEdit_textEdited(const QString& text) {
-  if (text.isEmpty()) {
+  auto* proxy_model = dynamic_cast<HighlightingSortFilterProxyModel*>(
+      ui_->topDownTreeView->model());
+  if (proxy_model == nullptr) {
     return;
   }
-  auto* proxy_model =
-      dynamic_cast<QSortFilterProxyModel*>(ui_->topDownTreeView->model());
-  if (proxy_model == nullptr) {
+  if (text.isEmpty()) {
+    proxy_model->SetNodesToHighlightSet(nullptr);
+    ui_->topDownTreeView->viewport()->update();
     return;
   }
   // Don't actually hide any node, but expand up to the parents of the matching
@@ -187,8 +239,18 @@ void TopDownWidget::on_searchLineEdit_textEdited(const QString& text) {
   // has no visual effect while the filter is in place, but those nodes would
   // retain the expanded state when the filter is removed). Finally, remove the
   // filter. The initial collapseAll greatly helps filtering performance.
+
+  // In parallel, collect the nodes that match the filter from the
+  // HighlightingSortFilterProxyModel and set them as nodes to highlight.
   ui_->topDownTreeView->collapseAll();
+  auto filter_accepted_nodes_collector =
+      std::make_unique<absl::flat_hash_set<void*>>();
+  proxy_model->SetFilterAcceptedNodeCollectorSet(
+      filter_accepted_nodes_collector.get());
   proxy_model->setFilterFixedString(text);
   ExpandAllButCollapseLeaves(ui_->topDownTreeView);
+  proxy_model->SetFilterAcceptedNodeCollectorSet(nullptr);
+  proxy_model->SetNodesToHighlightSet(
+      std::move(filter_accepted_nodes_collector));
   proxy_model->setFilterFixedString(QStringLiteral(""));
 }
--- a/OrbitQt/topdownwidget.cpp
+++ b/OrbitQt/topdownwidget.cpp
@@ -6,72 +6,22 @@
 
 #include <QColor>
 #include <QMenu>
-#include <QSortFilterProxyModel>
 
 #include "TopDownViewItemModel.h"
-#include "absl/container/flat_hash_set.h"
-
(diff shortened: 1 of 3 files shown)


ronaldfw:
I tried this and I think it's pretty cool!

There's one concern (with suggestions how to mitigate) and one suggestion I have: 

(1) The concern is that the search invalidates any expanding that a user has already done. I think that either search should only expand further (if needed) or we should add a warning text. The latter can be in the search bar ("Search (note that this will invalidate any changes you made to the top-down tree below)") or a dialog box with checkbox ("Don't ask me again") to confirm that they want to search.

(2) Can we highlight the lines that contain the search terms? Not with the "selection blue", but maybe with a white outline? (Unless this is easy and obvious how to do, probably for a later PR.)

OrbitQt/topdownwidget.cpp
```diff
@@ -136,3 +169,23 @@ void TopDownWidget::onCustomContextMenuRequested(const QPoint& point) {
     ui_->topDownTreeView->collapseAll();
   }
 }
+
+void TopDownWidget::on_searchLineEdit_textEdited(const QString& text) {
+  if (text.isEmpty()) {
+    return;
+  }
+  auto* proxy_model =
+      static_cast<QSortFilterProxyModel*>(ui_->topDownTreeView->model());
```
beckerhe, line 178: This feels reverse and is also potentially dangerous. I would suggest either to move the QSortFilterProxyModel into TopDownWidget or to move this method to where QSortFilterProxyModel lives. TopDownWidget could have a Qt-signal "filter_changed(const QString&)" and you can connect that signal to a function or lambda that has proper access to the proxy model.

dpallotti:
> There's one concern (with suggestions how to mitigate) and one suggestion I have:
> 
> (1) The concern is that the search invalidates any expanding that a user has already done. I think that either search should only expand further (if needed) or we should add a warning text. The latter can be in the search bar ("Search (note that this will invalidate any changes you made to the top-down tree below)") or a dialog box with checkbox ("Don't ask me again") to confirm that they want to search.
> 
> (2) Can we highlight the lines that contain the search terms? Not with the "selection blue", but maybe with a white outline? (Unless this is easy and obvious how to do, probably for a later PR.)

2) I wanted to do this and I think it's necessary. But I had quickly tried with the selection but it made everything really slow. As of now I don't really know how to do it with a different "style", so I guess I'll keep it for later.

1) Yep. The problem with only expanding further (without collapsing) is that if two searches in a row are made, the previous expansion is retained and then the second search is really confusing. Highlighting as in point 2 would help with this, but I still think that too many nodes would end up being opened. The even bigger problem is that as for this implementation the first character to be written is already a search, and that obviously expands a lot of nodes (unless we only trigger the search when Enter is hit or a button is clicked).

So I'm for a warning. I have a preference for the first option, not only because it's a one-line change, but also because changing the expansion of the tree is not *that* disruptive and getting a message dialog might be a bit too much / a bit annoying.

OrbitQt/topdownwidget.cpp
dpallotti: I'm not sure I understood. Could you elaborate? Why do you think it's dangerous and why does it feel reverse? Are you suggesting on keeping the `QSortFilterProxyModel` as a field of `TopDownWidget` and why is it necessary when it's stored in the `QTreeView`, or did I just get it all wrong?

pierricgimmig:
> (2) Can we highlight the lines that contain the search terms?
+1. This is super useful!

> (2) I wanted to do this and I think it's necessary. But I had quickly tried with the selection but it made everything really slow. As of now I don't really know how to do it with a different "style", so I guess I'll keep it for later.

This worked pretty fast in the "Show include util", but maybe that's because there's typically less data.  The color highlighting added a lot though. See it in action here: https://youtu.be/VKkbZyFIowc

dpallotti:
So I added highlighting. And as the screenshot shows, this is quite necessary.

highlighting

The code is probably hacky though. Copy-pasting from the commit message:
This introduces `HighlightingSortFilterProxyModel`, which inherits from
`QSortFilterProxyModel` and allows to retrieve the "result" of filter and to
"set" the nodes to highlight.
I can't disagree if you say that this is a hack. But I'm not sure what a better
way would be.
I had a look at how `ShowIncludesDialog`, `OrbitTreeModel`, `OrbitTreeItem` (now
not in the codebase anymore) were doing it. That maybe looked better, but it
brought the filtering logic all the way down to `OrbitTreeItem`. Not only is
that more complicated, but also I don't think is better conceptually, as that
logic should be further away from the model. As a matter of fact, it didn't use
Qt filtering code but custom `absl::StrContains`-based logic.

Let me know what you think.

OrbitQt/topdownwidget.cpp
beckerhe: Ahh. Never mind. It already "lives" in TopDownWidget. So this is fine.

Nit: I would prefer to access the proxy model via a member variable in TopDownWidget, so bascially having the pointer to the proxy model in TopDownWidget like you suggested.

Thanks and sorry for the delay!

beckerhe:
> So I added highlighting. And as the screenshot shows, this is quite necessary.
> 
> highlighting
> 
> The code is probably hacky though. Copy-pasting from the commit message:
> This introduces `HighlightingSortFilterProxyModel`, which inherits from
> `QSortFilterProxyModel` and allows to retrieve the "result" of filter and to
> "set" the nodes to highlight.
> I can't disagree if you say that this is a hack. But I'm not sure what a better
> way would be.
> I had a look at how `ShowIncludesDialog`, `OrbitTreeModel`, `OrbitTreeItem` (now
> not in the codebase anymore) were doing it. That maybe looked better, but it
> brought the filtering logic all the way down to `OrbitTreeItem`. Not only is
> that more complicated, but also I don't think is better conceptually, as that
> logic should be further away from the model. As a matter of fact, it didn't use
> Qt filtering code but custom `absl::StrContains`-based logic.
> 
> Let me know what you think.

I agree that this is a bit hacky and since you were basically asking I can't get around not offering a possible solution. (Sorry for that. ;-)):

First I would have two proxy models. One for filtering, one for highlighting. The filtering one would be the one from before. For highlighting I would subclass QIdentityProxyModel and override the data method. Furthermore this model should have a SetHighlightString method which sets a string member variable which is used to determine the rows to highlight. Check of that string matches can happen in the overriden data-method.

(Feel free to ignore that. That's a major rewrite and is less efficient than the other solution since the string-matching is done twice.)

OrbitQt/topdownwidget.cpp
dpallotti: Done, I had actually done it originally and then for some reason I changed my mind.

dpallotti:
> I agree that this is a bit hacky and since you were basically asking I can't get around not offering a possible solution. (Sorry for that. ;-)):
> 
> First I would have two proxy models. One for filtering, one for highlighting. The filtering one would be the one from before. For highlighting I would subclass QIdentityProxyModel and override the data method. Furthermore this model should have a SetHighlightString method which sets a string member variable which is used to determine the rows to highlight. Check of that string matches can happen in the overriden data-method.
> 
> (Feel free to ignore that. That's a major rewrite and is less efficient than the other solution since the string-matching is done twice.)

It's not such a big rewrite so that would be fine. I would need to check the performance though, the current implementation is not so slow but also already not that snappy.
Apart from that, in general I like the idea. The same could be done with just one proxy model, although two seem more elegant (again, I would need to check the performance), couldn't it? And that I had considered, but my problem is that I wanted to use the internal filtering of the proxy model, without having my own implementation of the match.
The reason for not wanting to add my own matching logic (to the also-used Qt logic) is that I'm not guaranteed they are exactly identical (e.g., in corner cases) and also they need to be kept in sync in case they are changed. I guess it's not a big deal, but that was the reasoning. What do you think @beckerhe?

beckerhe:
> > I agree that this is a bit hacky and since you were basically asking I can't get around not offering a possible solution. (Sorry for that. ;-)):
> > First I would have two proxy models. One for filtering, one for highlighting. The filtering one would be the one from before. For highlighting I would subclass QIdentityProxyModel and override the data method. Furthermore this model should have a SetHighlightString method which sets a string member variable which is used to determine the rows to highlight. Check of that string matches can happen in the overriden data-method.
> > (Feel free to ignore that. That's a major rewrite and is less efficient than the other solution since the string-matching is done twice.)
> 
> It's not such a big rewrite so that would be fine. I would need to check the performance though, the current implementation is not so slow but also already not that snappy.
> Apart from that, in general I like the idea. The same could be done with just one proxy model, although two seem more elegant (again, I would need to check the performance), couldn't it? And that I had considered, but my problem is that I wanted to use the internal filtering of the proxy model, without having my own implementation of the match.
> The reason for not wanting to add my own matching logic (to the also-used Qt logic) is that I'm not guaranteed they are exactly identical (e.g., in corner cases) and also they need to be kept in sync in case they are changed. I guess it's not a big deal, but that was the reasoning. What do you think @beckerhe?

Yeah, this also came to my mind. First I thought you could subclass "QSortFilterProxyModel" and use the filterAcceptRow-method to determine which row to highlight. But this feels very unclean. The "SetFilterFixedString" method would set pattern for highlight matching.

I think alternatively I would overwrite the "filterAcceptRow"-method in the filtering model and call a common free function which does both the filtering in the filter-model and the highlight-matching in the highlightproxymodel. True this means we have to maintain our own string-matching code. But this is hardly more than 3 lines, so probably a fair deal.

beckerhe:
I just had another idea which might improve performance:

1. You could introduce a new DataRole which determines if a node is marked in some sense. Then you add a proxy model which return true for that data role, when the filter string is matched.

2. Then you need a function which expands all the nodes that are marked by that role. Having that marking avoids filtering the model first and removing the filter again.

3. In addition to that marking this new proxy model could also do the highlighting. Or you could have a second proxy model which reads the markings and applies highlighting accordingly.

Of couse that's again a major rewrite. :see_no_evil: But I at least wanted to mention it.

dpallotti:
> I think alternatively I would overwrite the "filterAcceptRow"-method in the filtering model and call a common free function which does both the filtering in the filter-model and the highlight-matching in the highlightproxymodel. True this means we have to maintain our own string-matching code. But this is hardly more than 3 lines, so probably a fair deal.

This sounds fair, I didn't think about it, I would prefer using only Qt matching code but I'm also fine with using only our own matching logic. I just don't want to use both.

> I just had another idea which might improve performance:
> 
> 1. You could introduce a new DataRole which determines if a node is marked in some sense. Then you add a proxy model which return true for that data role, when the filter string is matched.
> 2. Then you need a function which expands all the nodes that are marked by that role. Having that marking avoids filtering the model first and removing the filter again.
> 3. In addition to that marking this new proxy model could also do the highlighting. Or you could have a second proxy model which reads the markings and applies highlighting accordingly.

I would need to think about this a bit more, as it's maybe as complicated as it sounds, or maybe not that bad. But my questions on this are: How do I retrieve the marking returned by `data`? And how do I store it? I'm missing a piece (again, Qt noob here).

dpallotti:
> I just had another idea which might improve performance:
> 
> 1. You could introduce a new DataRole which determines if a node is marked in some sense. Then you add a proxy model which return true for that data role, when the filter string is matched.
> 2. Then you need a function which expands all the nodes that are marked by that role. Having that marking avoids filtering the model first and removing the filter again.
> 3. In addition to that marking this new proxy model could also do the highlighting. Or you could have a second proxy model which reads the markings and applies highlighting accordingly.
> 
> Of couse that's again a major rewrite. 🙈 But I at least wanted to mention it.

I rewrote this following @beckerhe's last suggestion (thanks!). The code is much much cleaner now! Please have a look, Henning in particular.

PR #915, opened 2020-08-06
Fix "Narrowing conversion from 'double' to 'float'" warning.

florian-kuebler:
Fix "Narrowing conversion from 'double' to 'float'" warning. which prevented clang from compilation.

To fix the warning, the doubles were changed to float. The precision lost should be fine, as the actual values (coming in as params) were already floats.
Also applied clang-format to the changed files and added const keyword in touched functions.

### DIFF
--- a/OrbitGl/CaptureWindow.cpp
+++ b/OrbitGl/CaptureWindow.cpp
@@ -382,9 +382,6 @@ void CaptureWindow::MouseWheelMoved(int a_X, int a_Y, int a_Delta,
   if (delta < m_MinWheelDelta) m_MinWheelDelta = delta;
   if (delta > m_MaxWheelDelta) m_MaxWheelDelta = delta;
 
-  const auto mousex = static_cast<float>(a_X);
-  const auto mousey = static_cast<float>(a_Y);
-
   float worldx;
   float worldy;
 
@@ -392,12 +389,13 @@ void CaptureWindow::MouseWheelMoved(int a_X, int a_Y, int a_Delta,
 
   bool zoomWidth = !a_Ctrl;
   if (zoomWidth) {
-    m_MouseRatio = static_cast<double>(mousex) / getWidth();
+    m_MouseRatio = static_cast<double>(a_X) / getWidth();
     time_graph_.ZoomTime(delta_float, m_MouseRatio);
     m_WheelMomentum =
         delta_float * m_WheelMomentum < 0 ? 0 : m_WheelMomentum + delta_float;
   } else {
-    float mouse_relative_y_position = mousey / static_cast<float>(getHeight());
+    float mouse_relative_y_position =
+        static_cast<float>(a_Y) / static_cast<float>(getHeight());
     time_graph_.VerticalZoom(delta_float, mouse_relative_y_position);
   }
 
--- a/OrbitGl/CaptureWindow.cpp
+++ b/OrbitGl/CaptureWindow.cpp
@@ -376,13 +376,14 @@ void CaptureWindow::MouseWheelMoved(int a_X, int a_Y, int a_Delta,
   if (a_Delta == 0) return;
 
   // Normalize and invert sign, so that delta < 0 is zoom in.
-  int delta = a_Delta < 0 ? 1 : -1;
+  const int delta = a_Delta < 0 ? 1 : -1;
+  const auto delta_float = static_cast<float>(delta);
 
   if (delta < m_MinWheelDelta) m_MinWheelDelta = delta;
   if (delta > m_MaxWheelDelta) m_MaxWheelDelta = delta;
 
-  float mousex = a_X;
-  float mousey = a_Y;
+  const auto mousex = static_cast<float>(a_X);
+  const auto mousey = static_cast<float>(a_Y);
 
   float worldx;
   float worldy;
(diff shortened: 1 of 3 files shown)


OrbitGl/TimeGraph.cpp
```diff
@@ -196,16 +196,18 @@ void TimeGraph::ZoomTime(float a_ZoomValue, double a_MouseRatio) {
   SetMinMax(minTimeUs, maxTimeUs);
 }
 
-void TimeGraph::VerticalZoom(float zoom_value, double mouse_relative_position) {
-  static double increment_ratio = 0.1;
+void TimeGraph::VerticalZoom(float zoom_value, float mouse_relative_position) {
+  static const float increment_ratio = 0.1f;
```
beckerhe, line 200: Nit: That looks surprising. No need to make it static if it's const. So I would suggest to remove the static and maybe replace the const by constexpr.

OrbitGl/CaptureWindow.cpp
```diff
@@ -376,13 +376,14 @@ void CaptureWindow::MouseWheelMoved(int a_X, int a_Y, int a_Delta,
...
+  const int delta = a_Delta < 0 ? 1 : -1;
+  const auto delta_float = static_cast<float>(delta);
 
   if (delta < m_MinWheelDelta) m_MinWheelDelta = delta;
   if (delta > m_MaxWheelDelta) m_MaxWheelDelta = delta;
 
-  float mousex = a_X;
-  float mousey = a_Y;
+  const auto mousex = static_cast<float>(a_X);
+  const auto mousey = static_cast<float>(a_Y);
```
beckerhe, line 386: Nit: This looks surprising to me. Making the copy of `a_X` and `a_Y` here makes me think about why. And that makes me believe that ScreenToWorld changes the value of `a_X` and `a_Y`. What about either removing those variables or at least minimizing their scope by moving down?

florian-kuebler:
> I see a lot of things I would change here, but that's all unrelated to your change. :-) So LGTM!
> I added to nitty-gritty comments though.

Right. As I am not that familiar with the overall code here, I just wanted to fix the warning for now and do some obvious improvements without changing the code's structure (i.e. adding the const's).
Though I suggest for now, I will address your comments and we should be fine.
Please also note, that this PR is targeting the release branch.

beckerhe:
> > I see a lot of things I would change here, but that's all unrelated to your change. :-) So LGTM!
> > I added to nitty-gritty comments though.
> 
> Right. As I am not that familiar with the overall code here, I just wanted to fix the warning for now and do some obvious improvements without changing the code's structure (i.e. adding the const's).
> Though I suggest for now, I will address your comments and we should be fine.
> Please also note, that this PR is targeting the release branch.

Ah I didn't notice that. Why do you want to target the release branch with a compilation fix?

OrbitGl/CaptureWindow.cpp
florian-kuebler: > > > I see a lot of things I would change here, but that's all unrelated to your change. :-) So LGTM!
> > > I added to nitty-gritty comments though.
> > 
> > 
> > Right. As I am not that familiar with the overall code here, I just wanted to fix the warning for now and do some obvious improvements without changing the code's structure (i.e. adding the const's).
> > Though I suggest for now, I will address your comments and we should be fine.
> > Please also note, that this PR is targeting the release branch.
> 
> Ah I didn't notice that. Why do you want to target the release branch with a compilation fix?

As we said, that we wanted to keep the release branches for later, it feels super unnatural to have one, that might not compile. Assuming in future, someone checks out that branch, and compilation fails. This would be unfortunate. Thus, I suggest adding this fix to the release. 

Btw, I addressed your comments.

florian-kuebler:
> > > I see a lot of things I would change here, but that's all unrelated to your change. :-) So LGTM!
> > > I added to nitty-gritty comments though.
> > 
> > 
> > Right. As I am not that familiar with the overall code here, I just wanted to fix the warning for now and do some obvious improvements without changing the code's structure (i.e. adding the const's).
> > Though I suggest for now, I will address your comments and we should be fine.
> > Please also note, that this PR is targeting the release branch.
> 
> Ah I didn't notice that. Why do you want to target the release branch with a compilation fix?

We wanted to keep the releases branch as well as the release branch for the future, such that someone can always check it out for a stable version. Thus it would be unfortunate to checkout a release and get a compile error.

I addressed your comments btw.

beckerhe:
> > > > I see a lot of things I would change here, but that's all unrelated to your change. :-) So LGTM!
> > > > I added to nitty-gritty comments though.
> > > 
> > > 
> > > Right. As I am not that familiar with the overall code here, I just wanted to fix the warning for now and do some obvious improvements without changing the code's structure (i.e. adding the const's).
> > > Though I suggest for now, I will address your comments and we should be fine.
> > > Please also note, that this PR is targeting the release branch.
> > 
> > 
> > Ah I didn't notice that. Why do you want to target the release branch with a compilation fix?
> 
> We wanted to keep the releases branch as well as the release branch for the future, such that someone can always check it out for a stable version. Thus it would be unfortunate to checkout a release and get a compile error.
> 
> I addressed your comments btw.

I feel that kind of undermines our testing effort. Also there seems to be another PR #912 from pierric which has already been merged into master. Having the same problem fixed in different ways will make the merge-back unnecessarily hard.

florian-kuebler:
As discussed offline:
#916 also needs to go in to the release, so we need to redo manual testing anyways. As #912 also fixes the warning (but not the const + formatting), I will also backport that one to master. Due to the release playbook, this will resolve in a merge conflict, when merging back, but we will just override this with master's version. So as long as all changes are in master, we are fine.

PR #916, opened 2020-08-06
use SelectedInCaptureFunctions in Live tab and when saving the capture

IrinaShkviro:
Fix for http://b/163014195

Introduce `Capture::GSelectedInCaptureFunctions` to store the functions that were hooked in the capture, and show these functions on the "live" tab and when saving capture.

If we want to refer to currently hooked functions we should use `GSelectedFunctionsMap`.

### DIFF
--- a/OrbitCaptureClient/CaptureClient.cpp
+++ b/OrbitCaptureClient/CaptureClient.cpp
@@ -15,8 +15,7 @@ ABSL_DECLARE_FLAG(bool, frame_pointer_unwinding);
 using orbit_client_protos::FunctionInfo;
 
 void CaptureClient::Capture(
-    int32_t pid,
-    const std::vector<std::shared_ptr<FunctionInfo>>& selected_functions) {
+    int32_t pid, const std::map<uint64_t, FunctionInfo*>& selected_functions) {
   CHECK(reader_writer_ == nullptr);
 
   event_processor_.emplace(capture_listener_);
@@ -40,13 +39,14 @@ void CaptureClient::Capture(
     }
   }
   capture_options->set_trace_gpu_driver(true);
-  for (const std::shared_ptr<FunctionInfo>& function : selected_functions) {
+  for (const auto& pair : selected_functions) {
+    const FunctionInfo& function = *pair.second;
     CaptureOptions::InstrumentedFunction* instrumented_function =
         capture_options->add_instrumented_functions();
-    instrumented_function->set_file_path(function->loaded_module_path());
-    instrumented_function->set_file_offset(FunctionUtils::Offset(*function));
+    instrumented_function->set_file_path(function.loaded_module_path());
+    instrumented_function->set_file_offset(FunctionUtils::Offset(function));
     instrumented_function->set_absolute_address(
-        FunctionUtils::GetAbsoluteAddress(*function));
+        FunctionUtils::GetAbsoluteAddress(function));
   }
 
   if (!reader_writer_->Write(request)) {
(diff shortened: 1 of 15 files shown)


OrbitGl/LiveFunctionsDataView.h
```diff
@@ -34,7 +34,7 @@ class LiveFunctionsDataView : public DataView {
   std::pair<TextBox*, TextBox*> GetMinMax(
       const orbit_client_protos::FunctionInfo& function) const;
 
-  std::vector<orbit_client_protos::FunctionInfo*> m_Functions;
+  std::vector<std::shared_ptr<orbit_client_protos::FunctionInfo>> m_Functions;
```
ronaldfw, line 37: Nit: Maybe take the opportunity and change m_Functions to functions_?

OrbitCaptureClient/CaptureClient.cpp
```diff
@@ -15,8 +15,7 @@ ABSL_DECLARE_FLAG(bool, frame_pointer_unwinding);
 using orbit_client_protos::FunctionInfo;
 
 void CaptureClient::Capture(
-    int32_t pid,
-    const std::vector<std::shared_ptr<FunctionInfo>>& selected_functions) {
+    int32_t pid, const std::map<uint64_t, FunctionInfo*>& selected_functions) {
```
florian-kuebler, line 18: Did this actually led to instrumenting a function that was unselected?

OrbitGl/App.cpp
```diff
@@ -584,8 +584,8 @@ bool OrbitApp::StartCapture() {
   }
 
   int32_t pid = Capture::GProcessId;
-  std::vector<std::shared_ptr<FunctionInfo>> selected_functions =
-      Capture::GSelectedFunctions;
+  std::map<uint64_t, FunctionInfo*> selected_functions =
```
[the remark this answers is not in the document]
florian-kuebler: I guess we should postpone this. In capture there are a lot of std::maps where we should think of using the absl versions. Changing them would require a lot of places to be modified. Further we want to change them to get rid of the global anyways. So I suggest to do this together.

OrbitGl/LiveFunctionsDataView.h
IrinaShkviro: Done, thanks!

OrbitGl/LiveFunctionsDataView.cpp
```diff
@@ -302,17 +303,12 @@ void LiveFunctionsDataView::DoFilter() {
...
-
   m_Functions.clear();
-  for (auto& pair : Capture::GVisibleFunctionsMap) {
-    const ULONG64& address = pair.first;
-    FunctionInfo* func = Capture::GVisibleFunctionsMap[address];
-    m_Functions.push_back(func);
+  size_t functions_count = Capture::GSelectedInCaptureFunctions.size();
+  m_Indices.resize(functions_count);
+  for (size_t i = 0; i < functions_count; ++i) {
+    m_Indices[i] = i;
```
florian-kuebler, line 310: If you go for renaming [m_Functions](https://github.com/google/orbit/pull/916/files#r466394460), maybe also go for m_Indices to indices_.

OrbitGl/App.cpp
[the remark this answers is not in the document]
IrinaShkviro: I think so... We should also change the type of Capture::GSelectedFunctionsMap. But probably, I would prefer to leave it as is for now as it's not necessary for this fix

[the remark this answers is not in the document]
ronaldfw: Ok, makes sense!

OrbitGl/LiveFunctionsDataView.cpp
IrinaShkviro: Done, thanks!

PR #918, opened 2020-08-06
Fix Path::FileExists bug

antonrohr:
Path::FileExists used stat to determine if a file exists. This did not
on Windows when the filename contained a dash (-).
Path::FileExists now uses std::filesystem::exists()

b/163028647

### DIFF
--- a/OrbitCore/Path.cpp
+++ b/OrbitCore/Path.cpp
@@ -64,12 +64,6 @@ std::string Path::GetExecutablePath() {
   return path;
 }
 
-bool Path::FileExists(const std::string& file) {
-  struct stat statbuf;
-  int ret = stat(file.c_str(), &statbuf);
-  return ret == 0;
-}
-
 uint64_t Path::FileSize(const std::string& file) {
   struct stat stat_buf;
   int ret = stat(file.c_str(), &stat_buf);
--- a/OrbitCore/PathTest.cpp
+++ b/OrbitCore/PathTest.cpp
@@ -4,12 +4,14 @@
 
 #include <gtest/gtest.h>
 
+#include <filesystem>
+
 #include "Path.h"
 #include "absl/strings/match.h"
 
 TEST(Path, FileExistsEmptyFilename) {
   std::string filename{};
-  EXPECT_FALSE(Path::FileExists(filename));
+  EXPECT_FALSE(std::filesystem::exists(filename));
 }
 
 TEST(Path, FileExistsRootDir) {
@@ -19,13 +21,13 @@ TEST(Path, FileExistsRootDir) {
 #else
   filename = "/";
 #endif
-  EXPECT_TRUE(Path::FileExists(filename));
+  EXPECT_TRUE(std::filesystem::exists(filename));
 }
 
 #ifndef _WIN32
 TEST(Path, FileExistsDevNull) {
   std::string filename = "/dev/null";
-  EXPECT_TRUE(Path::FileExists(filename));
+  EXPECT_TRUE(std::filesystem::exists(filename));
 }
 #endif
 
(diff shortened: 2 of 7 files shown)


OrbitCore/Path.cpp
```diff
@@ -65,9 +65,7 @@ std::string Path::GetExecutablePath() {
 }
 
 bool Path::FileExists(const std::string& file) {
```
beckerhe, line 67: I think we should add a `[[deprecated("Please use std::filesystem:.exists")]]` here. (Obviously this needs to go into the header file.

ronaldfw, line 67: While we touch this function, can we change this to std::string_view?

OrbitCore/Path.cpp
```diff
@@ -65,9 +65,7 @@ std::string Path::GetExecutablePath() {
...
-  struct stat statbuf;
-  int ret = stat(file.c_str(), &statbuf);
-  return ret == 0;
```
dimitry-, line 70: I have not idea why this doesn't work on windows.  I agree that we should use filesystem::exists but still I am puzzled by this. Is this some kind of known bug in windows stat implementation?

OrbitCore/Path.cpp
dimitry-: I think it is better to just replace calls to this function with filesystem:exists(), 

./OrbitCore/SymbolHelper.cpp:23:  if (!Path::FileExists(file_name)) {
./OrbitCore/SymbolHelper.cpp:80:    if (!Path::FileExists(symbols_file_path)) continue;
./OrbitCore/OrbitModule.cpp:31:  if (!Path::FileExists(file_name)) {
./OrbitCore/LinuxUtils.cpp:121:    if (!Path::FileExists(module_path)) continue;
./OrbitCore/Path.h:45:  static bool FileExists(const std::string& a_File);
./OrbitCore/Path.cpp:67:bool Path::FileExists(const std::string& file) {
./OrbitCore/PathTest.cpp:12:  EXPECT_FALSE(Path::FileExists(filename));
./OrbitCore/PathTest.cpp:22:  EXPECT_TRUE(Path::FileExists(filename));
./OrbitCore/PathTest.cpp:28:  EXPECT_TRUE(Path::FileExists(filename));
./OrbitGl/App.cpp:252:  if (!Path::FileExists(fileName)) {

beckerhe: I won't argue with that. :slightly_smiling_face:

OrbitCore/Path.cpp
antonrohr: I tried to figure it out but i did not make any progress and decided its not worth the effort. 
It did not work for a module that had 3 dashes in the filename.

OrbitCore/Path.cpp
antonrohr: done

OrbitCore/PathTest.cpp
```diff
@@ -4,12 +4,14 @@
...
+#include <filesystem>
+
 #include "Path.h"
 #include "absl/strings/match.h"
 
 TEST(Path, FileExistsEmptyFilename) {
   std::string filename{};
-  EXPECT_FALSE(Path::FileExists(filename));
+  EXPECT_FALSE(std::filesystem::exists(filename));
 }
```
dimitry-, line 15: I think we do not need to test std::filesystem :)

OrbitCore/PathTest.cpp
```diff
@@ -19,13 +21,13 @@ TEST(Path, FileExistsRootDir) {
 #else
   filename = "/";
 #endif
-  EXPECT_TRUE(Path::FileExists(filename));
```
dimitry-, line 22: this also can be removed.

OrbitCore/PathTest.cpp
```diff
@@ -19,13 +21,13 @@ TEST(Path, FileExistsRootDir) {
...
+  EXPECT_TRUE(std::filesystem::exists(filename));
 }
 
 #ifndef _WIN32
 TEST(Path, FileExistsDevNull) {
   std::string filename = "/dev/null";
-  EXPECT_TRUE(Path::FileExists(filename));
```
dimitry-, line 28: and this.

PR #922, opened 2020-08-06
Add command line flag for log path.

danielfenner:
When running the integration tests I'd like to create the log in a place where it gets uploaded automatically.

### DIFF
--- a/OrbitCore/Path.cpp
+++ b/OrbitCore/Path.cpp
@@ -242,14 +242,14 @@ std::string Path::GetAppDataPath() {
 }
 
 std::string Path::GetLogFilePath() {
-  std::string logsDir;
+  std::string logs_dir;
   if (!absl::GetFlag(FLAGS_log_dir).empty()) {
-    logsDir = absl::GetFlag(FLAGS_log_dir);
+    logs_dir = absl::GetFlag(FLAGS_log_dir);
   } else {
-    logsDir = Path::JoinPath({Path::GetAppDataPath(), "logs"});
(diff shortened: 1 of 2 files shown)


OrbitCore/Path.cpp
```diff
@@ -238,7 +242,12 @@ std::string Path::GetAppDataPath() {
 }
 
 std::string Path::GetLogFilePath() {
-  std::string logsDir = Path::JoinPath({Path::GetAppDataPath(), "logs"});
+  std::string logsDir;
```
IrinaShkviro, line 245:
```suggestion
  std::string logs_dir;
```

OrbitQt/main.cpp
```diff
@@ -249,13 +249,14 @@ int main(int argc, char* argv[]) {
...
-    const std::string log_file_path = Path::GetLogFilePath();
-    InitLogFile(log_file_path);
-
     absl::SetProgramUsageMessage("CPU Profiler");
     absl::SetFlagsUsageConfig(
         absl::FlagsUsageConfig{{}, {}, {}, &OrbitCore::GetBuildReport, {}});
     absl::ParseCommandLine(argc, argv);
+
+	const std::string log_file_path = Path::GetLogFilePath();
+    InitLogFile(log_file_path);
```
IrinaShkviro, line 258:
```suggestion
    const std::string log_file_path = Path::GetLogFilePath();
    InitLogFile(log_file_path);
```

danielfenner:
Thanks for having a look!

PR #923, opened 2020-08-06
use sync call with deadline in CrashManager instead of async

IrinaShkviro:
Fix for http://b/163018728

Replace asynchronous call with synchronous one (with deadline) for CrashManager

### DIFF
--- a/OrbitClientServices/CrashManager.cpp
+++ b/OrbitClientServices/CrashManager.cpp
@@ -40,9 +40,7 @@ void CrashManagerImpl::CrashOrbitService(
   grpc::Status status =
       crash_service_->CrashOrbitService(&context, request, &response);
 
-  if (!status.ok()) {
-    ERROR("Grpc call failed: %s", status.error_message());
-  }
+  CHECK(status.error_code() == grpc::StatusCode::DEADLINE_EXCEEDED);
 }
 
 }  // namespace
--- a/OrbitClientServices/CrashManager.cpp
+++ b/OrbitClientServices/CrashManager.cpp
@@ -9,7 +9,7 @@
 
 namespace {
 
-constexpr uint64_t kTimeoutMilliseconds = 10;
+constexpr uint64_t kTimeoutMilliseconds = 100;
 
 class CrashManagerImpl final : public CrashManager {
  public:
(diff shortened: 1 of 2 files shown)


OrbitClientServices/CrashManager.cpp
```diff
@@ -28,16 +30,19 @@ void CrashManagerImpl::CrashOrbitService(
...
 
-  std::function<void(grpc::Status)> callback = [](::grpc::Status status) {
-    if (!status.ok()) {
-      ERROR("Grpc call failed: %s", status.error_message());
-    }
-  };
-  crash_service_->experimental_async()->CrashOrbitService(&context, &request,
-                                                          &response, callback);
+  if (!status.ok()) {
+    ERROR("Grpc call failed: %s", status.error_message());
```
dimitry-, line 44: we kind of expect it to fail with deadline aren't we? maybe check if it returns expected deadline exceeded error?

OrbitClientServices/CrashManager.cpp
```diff
@@ -9,6 +9,8 @@
 
 namespace {
 
+constexpr uint64_t kTimeoutMilliseconds = 10;
```
dimitry-, line 12: I would increase it a little bit - to 100ms probably. I do not know how grpc behaves if deadline is in the past.

OrbitClientServices/CrashManager.cpp
IrinaShkviro: Yes, you're right: I added a check for this, thanks!

OrbitClientServices/CrashManager.cpp
IrinaShkviro: Done, thanks. Probably if deadline in the past, it just do nothing: I set it to 1ms and OrbitService didn't crash (and no log records too).

OrbitClientServices/CrashManager.cpp
```diff
@@ -40,9 +40,7 @@ void CrashManagerImpl::CrashOrbitService(
   grpc::Status status =
       crash_service_->CrashOrbitService(&context, request, &response);
 
-  if (!status.ok()) {
-    ERROR("Grpc call failed: %s", status.error_message());
-  }
+  CHECK(status.error_code() == grpc::StatusCode::DEADLINE_EXCEEDED);
```
dimitry-, line 43: maybe have an if() with log - or even dialog if error_code is not DEADLINE_EXCEEDED. CHECK seems like an overkill here.

IrinaShkviro: Added an error to the log

PR #924, opened 2020-08-07
Manual instrumentation UI.

pierricgimmig:
- User defined scopes are visible in the timeline.
- GraphTracks are created for tracked variables.
- Implement remote string retrieval mechanism.

image

image

TODO:
- properly show manual instrumentation timers in live view
- fix tooltip for manual instrumentation timers
- polish graph track rendering

### DIFF
--- a/OrbitCaptureClient/CaptureClient.cpp
+++ b/OrbitCaptureClient/CaptureClient.cpp
@@ -6,13 +6,24 @@
 
 #include "FunctionUtils.h"
 #include "OrbitBase/Logging.h"
-
 #include "absl/flags/flag.h"
 
 ABSL_DECLARE_FLAG(uint16_t, sampling_rate);
 ABSL_DECLARE_FLAG(bool, frame_pointer_unwinding);
 
 using orbit_client_protos::FunctionInfo;
+
+static CaptureOptions::InstrumentedFunction::FunctionType
+IntrumentedFunctionTypeFromOrbitType(FunctionInfo::OrbitType orbit_type) {
+  switch (orbit_type) {
+    case FunctionInfo::kOrbitTimerStart:
+      return CaptureOptions::InstrumentedFunction::kTimerStart;
+    case FunctionInfo::kOrbitTimerStop:
+      return CaptureOptions::InstrumentedFunction::kTimerStop;
+    default:
+      return CaptureOptions::InstrumentedFunction::kRegular;
+  }
+}
 
 void CaptureClient::Capture(
     int32_t pid, const std::map<uint64_t, FunctionInfo*>& selected_functions) {
@@ -38,6 +49,7 @@ void CaptureClient::Capture(
       capture_options->set_unwinding_method(CaptureOptions::kDwarf);
     }
   }
+
   capture_options->set_trace_gpu_driver(true);
   for (const auto& pair : selected_functions) {
     const FunctionInfo& function = *pair.second;
@@ -47,6 +59,8 @@ void CaptureClient::Capture(
     instrumented_function->set_file_offset(FunctionUtils::Offset(function));
     instrumented_function->set_absolute_address(
         FunctionUtils::GetAbsoluteAddress(function));
+    instrumented_function->set_function_type(
+        IntrumentedFunctionTypeFromOrbitType(function.type()));
   }
 
   if (!reader_writer_->Write(request)) {
--- a/OrbitClientServices/ProcessManager.cpp
+++ b/OrbitClientServices/ProcessManager.cpp
@@ -33,7 +33,7 @@ class ProcessManagerImpl final : public ProcessManager {
                                                 uint64_t size) override;
 
   ErrorMessageOr<std::string> LoadNullTerminatedString(
-      int32_t pid, uint64_t address, uint64_t max_size) override;
+      int32_t pid, uint64_t address) override;
 
   ErrorMessageOr<std::string> FindDebugInfoFile(
       const std::string& module_path, const std::string& build_id) override;
@@ -203,7 +203,8 @@ ErrorMessageOr<std::string> ProcessManagerImpl::LoadProcessMemory(
 }
 
 ErrorMessageOr<std::string> ProcessManagerImpl::LoadNullTerminatedString(
-    int32_t pid, uint64_t address, uint64_t max_size) {
+    int32_t pid, uint64_t address) {
+  constexpr uint64_t max_size = 256;
   auto error_or_string = LoadProcessMemory(pid, address, max_size);
   if (error_or_string.has_value()) {
     std::string str = error_or_string.value();
--- a/OrbitClientServices/ProcessManager.cpp
+++ b/OrbitClientServices/ProcessManager.cpp
@@ -206,9 +206,10 @@ ErrorMessageOr<std::string> ProcessManagerImpl::LoadNullTerminatedString(
     int32_t pid, uint64_t address, uint64_t max_size) {
   auto error_or_string = LoadProcessMemory(pid, address, max_size);
   if (error_or_string.has_value()) {
-    // Make sure the string is null terminated.
     std::string str = error_or_string.value();
-    str[str.size() - 1] = '\0';
+    // The string has a size of max_size at this point. Shrink it by assigning
+    // it its own str.c_str(). The result of c_str() is guaranteed to be null
+    // terminated so this works when the remote string is bigger than max_size.
     error_or_string = str.c_str();
   }
 
--- a/OrbitClientServices/ProcessManager.cpp
+++ b/OrbitClientServices/ProcessManager.cpp
@@ -203,16 +203,16 @@ ErrorMessageOr<std::string> ProcessManagerImpl::LoadProcessMemory(
 }
 
 ErrorMessageOr<std::string> ProcessManagerImpl::LoadNullTerminatedString(
-      int32_t pid, uint64_t address, uint64_t max_size){
-  auto errorOrString = LoadProcessMemory(pid, address, max_size);
-  if(errorOrString.has_value()) {
+    int32_t pid, uint64_t address, uint64_t max_size) {
+  auto error_or_string = LoadProcessMemory(pid, address, max_size);
+  if (error_or_string.has_value()) {
     // Make sure the string is null terminated.
-    std::string str = errorOrString.value();
-    str[str.size()-1] = '\0';
-    errorOrString = str.c_str();
+    std::string str = error_or_string.value();
+    str[str.size() - 1] = '\0';
+    error_or_string = str.c_str();
   }
 
-  return errorOrString;
+  return error_or_string;
 }
 
 }  // namespace
--- a/OrbitClientServices/ProcessManager.cpp
+++ b/OrbitClientServices/ProcessManager.cpp
@@ -209,12 +209,13 @@ ErrorMessageOr<std::string> ProcessManagerImpl::LoadNullTerminatedString(
   if (error_or_string.has_value()) {
     const std::string& str = error_or_string.value();
     if (str.find('\0') == std::string::npos) {
-      ERROR("Remote string is not null terminated: %s", str.c_str());
+      const char* error_msg = "Remote string is not null terminated";
+      ERROR("%s: %s", error_msg, str.c_str());
+      return ErrorMessage(error_msg);
     }
 
     // The string has a size of max_size at this point. Shrink it by assigning
-    // it its own str.c_str(). The result of c_str() is guaranteed to be null
-    // terminated so this works when the remote string is bigger than max_size.
+    // it its own str.c_str(). c_str() is guaranteed to be null terminated.
     error_or_string = str.c_str();
   }
 
--- a/OrbitClientServices/ProcessManager.cpp
+++ b/OrbitClientServices/ProcessManager.cpp
@@ -207,7 +207,11 @@ ErrorMessageOr<std::string> ProcessManagerImpl::LoadNullTerminatedString(
   constexpr uint64_t max_size = 256;
   auto error_or_string = LoadProcessMemory(pid, address, max_size);
   if (error_or_string.has_value()) {
-    std::string str = error_or_string.value();
+    const std::string& str = error_or_string.value();
+    if (str.find('\0') == std::string::npos) {
+      ERROR("Remote string is not null terminated: %s", str.c_str());
+    }
+
     // The string has a size of max_size at this point. Shrink it by assigning
     // it its own str.c_str(). The result of c_str() is guaranteed to be null
     // terminated so this works when the remote string is bigger than max_size.
--- a/OrbitClientServices/ProcessManager.cpp
+++ b/OrbitClientServices/ProcessManager.cpp
@@ -32,6 +32,9 @@ class ProcessManagerImpl final : public ProcessManager {
   ErrorMessageOr<std::string> LoadProcessMemory(int32_t pid, uint64_t address,
                                                 uint64_t size) override;
 
+  ErrorMessageOr<std::string> LoadNullTerminatedString(
+      int32_t pid, uint64_t address, uint64_t max_size) override;
+
   ErrorMessageOr<std::string> FindDebugInfoFile(
       const std::string& module_path, const std::string& build_id) override;
 
@@ -197,6 +200,19 @@ ErrorMessageOr<std::string> ProcessManagerImpl::LoadProcessMemory(
   }
 
   return std::move(*response.mutable_memory());
+}
+
+ErrorMessageOr<std::string> ProcessManagerImpl::LoadNullTerminatedString(
+      int32_t pid, uint64_t address, uint64_t max_size){
+  auto errorOrString = LoadProcessMemory(pid, address, max_size);
+  if(errorOrString.has_value()) {
+    // Make sure the string is null terminated.
+    std::string str = errorOrString.value();
+    str[str.size()-1] = '\0';
+    errorOrString = str.c_str();
+  }
+
+  return errorOrString;
 }
 
 }  // namespace
--- a/OrbitCore/StringManager.cpp
+++ b/OrbitCore/StringManager.cpp
@@ -13,7 +13,14 @@ bool StringManager::AddIfNotPresent(uint64_t key, std::string_view str) {
   return true;
 }
 
-std::optional<std::string> StringManager::Get(uint64_t key) {
+bool StringManager::AddOrReplace(uint64_t key, std::string_view str) {
+  absl::MutexLock lock{&mutex_};
+  bool added = !key_to_string_.contains(key);
+  key_to_string_[key] = str;
+  return added;
+}
+
+std::optional<std::string> StringManager::Get(uint64_t key) const {
   absl::MutexLock lock{&mutex_};
   auto it = key_to_string_.find(key);
   if (it != key_to_string_.end()) {
@@ -23,7 +30,7 @@ std::optional<std::string> StringManager::Get(uint64_t key) {
   }
 }
 
-bool StringManager::Contains(uint64_t key) {
+bool StringManager::Contains(uint64_t key) const {
   absl::MutexLock lock{&mutex_};
   return key_to_string_.contains(key);
 }
--- a/OrbitCore/StringManager.cpp
+++ b/OrbitCore/StringManager.cpp
@@ -15,9 +15,8 @@ bool StringManager::AddIfNotPresent(uint64_t key, std::string_view str) {
 
 bool StringManager::AddOrReplace(uint64_t key, std::string_view str) {
   absl::MutexLock lock{&mutex_};
-  bool added = !key_to_string_.contains(key);
-  key_to_string_.insert_or_assign(key, str);
-  return added;
+  auto result = key_to_string_.insert_or_assign(key, str);
+  return result.second;
 }
 
 std::optional<std::string> StringManager::Get(uint64_t key) const {
--- a/OrbitCore/StringManager.cpp
+++ b/OrbitCore/StringManager.cpp
@@ -16,7 +16,7 @@ bool StringManager::AddIfNotPresent(uint64_t key, std::string_view str) {
 bool StringManager::AddOrReplace(uint64_t key, std::string_view str) {
   absl::MutexLock lock{&mutex_};
   bool added = !key_to_string_.contains(key);
-  key_to_string_[key] = str;
+  key_to_string_.insert_or_assign(key, str);
   return added;
 }
 
--- a/OrbitCore/StringManager.h
+++ b/OrbitCore/StringManager.h
@@ -19,11 +19,12 @@ class StringManager {
   bool AddIfNotPresent(uint64_t key, std::string_view str);
   // Rerturns true if insertion took place.
   bool AddOrReplace(uint64_t key, std::string_view str);
-  std::optional<std::string> Get(uint64_t key) const;
-  bool Contains(uint64_t key) const;
+  [[nodiscard]] std::optional<std::string> Get(uint64_t key) const;
+  [[nodiscard]] bool Contains(uint64_t key) const;
   void Clear();
 
-  const absl::flat_hash_map<uint64_t, std::string>& GetKeyToStringMap() const {
+  [[nodiscard]] const absl::flat_hash_map<uint64_t, std::string>&
+  GetKeyToStringMap() const {
     return key_to_string_;
   }
 
--- a/OrbitCore/StringManager.h
+++ b/OrbitCore/StringManager.h
@@ -15,7 +15,9 @@ class StringManager {
  public:
   StringManager() = default;
 
+  // Rerturns true if insertion took place.
   bool AddIfNotPresent(uint64_t key, std::string_view str);
+  // Rerturns true if insertion took place.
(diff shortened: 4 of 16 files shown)


OrbitClientServices/ProcessManager.cpp
```diff
@@ -202,6 +205,19 @@ ErrorMessageOr<std::string> ProcessManagerImpl::LoadProcessMemory(
...
 }
 
+ErrorMessageOr<std::string> ProcessManagerImpl::LoadNullTerminatedString(
+      int32_t pid, uint64_t address, uint64_t max_size){
+  auto errorOrString = LoadProcessMemory(pid, address, max_size);
+  if(errorOrString.has_value()) {
+    // Make sure the string is null terminated.
+    std::string str = errorOrString.value();
+    str[str.size()-1] = '\0';
+    errorOrString = str.c_str();
```
dimitry-: why do you need this to be null-terminated?

OrbitCore/StringManager.h
```diff
@@ -16,8 +16,9 @@ class StringManager {
   StringManager() = default;
 
   bool AddIfNotPresent(uint64_t key, std::string_view str);
-  std::optional<std::string> Get(uint64_t key);
-  bool Contains(uint64_t key);
+  bool AddOrReplace(uint64_t key, std::string_view str);
+  std::optional<std::string> Get(uint64_t key) const;
+  bool Contains(uint64_t key) const;
```
dimitry-, line 21: [[nodiscard]] for last 2 functions

OrbitGl/App.h
```diff
@@ -233,6 +233,14 @@ class OrbitApp final : public DataViewFactory, public CaptureListener {
 
   DataView* GetOrCreateDataView(DataViewType type) override;
 
+  const std::unique_ptr<ProcessManager>& GetProcessManager() {
+    return process_manager_;
+  }
+  const std::unique_ptr<ThreadPool>& GetThreadPool() { return thread_pool_; }
+  const std::unique_ptr<MainThreadExecutor>& GetMainThreadExecutor() {
```
dimitry-, line 240: return a pointer - we do not transfer ownership here, also [[nodiscard]]

OrbitGl/GraphTrack.h
```diff
@@ -14,17 +14,20 @@ class TimeGraph;
 
 class GraphTrack : public Track {
  public:
-  explicit GraphTrack(TimeGraph* time_graph);
+  explicit GraphTrack(TimeGraph* time_graph, uint64_t graph_id);
   [[nodiscard]] Type GetType() const override { return kGraphTrack; }
   void Draw(GlCanvas* canvas, PickingMode /*picking_mode*/) override;
   [[nodiscard]] float GetHeight() const override;
+  void AddValue(double value, uint64_t time);
+  double GetValueAtTime(uint64_t time, double default_value = 0) const;
```
dimitry-, line 22: [[nodiscard]]

OrbitGl/ThreadTrack.cpp
```diff
@@ -138,10 +138,21 @@ void ThreadTrack::SetTimesliceText(const TimerInfo& timer_info,
...
+      if (func->type() == FunctionInfo::kOrbitTimerStart) {
+        name = time_graph_->GetManualInstrumentationString(
+            timer_info.registers(0));
+        if (name.empty()) {
+          // The remote string hasn't been retrieved yet,
+          // early out and try again on next update.
+          return;
+        }
+      } else {
+        name = FunctionUtils::GetDisplayName(*func).c_str();
```
dimitry-, line 153: drop .c_str()?

OrbitGl/TimeGraph.h
```diff
@@ -141,6 +145,7 @@ class TimeGraph {
   const TextBox* FindDown(TextBox* from);
 
   Color GetThreadColor(ThreadID tid) const;
+  std::string GetManualInstrumentationString(uint64_t string_address) const;
```
dimitry-, line 148: nodiscard

OrbitGl/TimeGraph.h
```diff
@@ -157,6 +162,12 @@ class TimeGraph {
   std::shared_ptr<SchedulerTrack> GetOrCreateSchedulerTrack();
   std::shared_ptr<ThreadTrack> GetOrCreateThreadTrack(ThreadID a_TID);
   std::shared_ptr<GpuTrack> GetOrCreateGpuTrack(uint64_t timeline_hash);
+  std::shared_ptr<GraphTrack> GetOrCreateGraphTrack(uint64_t graph_id);
```
dimitry-, line 165: nodiscard

OrbitGl/TimeGraph.cpp
```diff
@@ -848,6 +935,44 @@ std::shared_ptr<GpuTrack> TimeGraph::GetOrCreateGpuTrack(
   return track;
 }
 
+//-----------------------------------------------------------------------------
+void SetTrackNameFromRemoteMemory(std::shared_ptr<Track> track,
+                                  uint64_t string_address) {
+  const auto& thread_pool = GOrbitApp->GetThreadPool();
```
dimitry-, line 941: consider passing executor/thread pool here instead of using global.

OrbitCore/StringManager.cpp
```diff
@@ -13,7 +13,14 @@ bool StringManager::AddIfNotPresent(uint64_t key, std::string_view str) {
   return true;
 }
 
-std::optional<std::string> StringManager::Get(uint64_t key) {
+bool StringManager::AddOrReplace(uint64_t key, std::string_view str) {
+  absl::MutexLock lock{&mutex_};
+  bool added = !key_to_string_.contains(key);
+  key_to_string_[key] = str;
```
dimitry-, line 19: I think you can use map::insert_or_assign here.

OrbitClientServices/ProcessManager.cpp
```diff
@@ -202,6 +205,19 @@ ErrorMessageOr<std::string> ProcessManagerImpl::LoadProcessMemory(
   return std::move(*response.mutable_memory());
 }
 
+ErrorMessageOr<std::string> ProcessManagerImpl::LoadNullTerminatedString(
+      int32_t pid, uint64_t address, uint64_t max_size){
+  auto errorOrString = LoadProcessMemory(pid, address, max_size);
```
dimitry-: I am not sure this will load null-terminated string. You might want to find first '\0' manually and resize the string.

it looks like you are currently returning just a fragment of memory with size 0, and letting caller figure out what to do with it. In which case this is not better than LoadProcessMemory.

OrbitClientServices/ProcessManager.cpp
```diff
@@ -206,16 +206,16 @@ ErrorMessageOr<std::string> ProcessManagerImpl::LoadProcessMemory(
...
+    int32_t pid, uint64_t address, uint64_t max_size) {
+  auto error_or_string = LoadProcessMemory(pid, address, max_size);
+  if (error_or_string.has_value()) {
     // Make sure the string is null terminated.
-    std::string str = errorOrString.value();
-    str[str.size()-1] = '\0';
-    errorOrString = str.c_str();
+    std::string str = error_or_string.value();
+    str[str.size() - 1] = '\0';
+    error_or_string = str.c_str();
```
dimitry-, line 215: Nevermind - it looks like I missed what this line does. I wonder if in the case of there is no \0 this function should return a error? Also from the caller perspective max_size does not make sense and is an implementation detail. We should probably make it a constant on the service side. So that there is no way to request too large chunks of memory from the client.

OrbitGl/TimeGraph.cpp
dimitry-: or OrbitApp* app

OrbitClientServices/ProcessManager.cpp
pierricgimmig: Modified the logic, please see comment.

OrbitCore/StringManager.h
pierricgimmig: Let's discuss the use of [[nodiscard]] with the team.

OrbitGl/App.h
pierricgimmig: Done for the pointer part. Let's discuss the use of [[nodiscard]] with the team.

OrbitGl/GraphTrack.h
pierricgimmig: Let's discuss the use of [[nodiscard]] with the team.

OrbitGl/ThreadTrack.cpp
pierricgimmig: yes, thanks.

OrbitGl/TimeGraph.h
pierricgimmig: Let's discuss the use of [[nodiscard]] with the team.

OrbitGl/TimeGraph.h
pierricgimmig: Let's discuss the use of [[nodiscard]] with the team.

OrbitGl/TimeGraph.cpp
pierricgimmig: Done.

OrbitCore/StringManager.cpp
pierricgimmig: Done.

OrbitClientServices/ProcessManager.cpp
pierricgimmig: Well if the caller wants to request an unusually long string, having the parameter is handy. I'd prefer keeping it.  I did remove the punching in of the \0 character as it turns out c_str() is guaranteed to be null terminating. Added a comment as well.

dimitry-: I still think this should be service-side resposibility. Specifying size for GetNullTerminatedString which returns std::string is leaking implementation details (we shouldn't change api when we change implementation, say implementation reads /proc/pid/mem directly and scans for \0 in place, in which case there is no need for size, it is meaningless). The caller does not have to know how the function works. If you want to get memory with size - you can use existing call, and then do the null-termination logic on the client side.

pierricgimmig: Please see modified code that hides the max_size value as implementation detail and warns if the string isn't null terminated.

OrbitGl/TimeGraph.h
pierricgimmig: In this case, since the function is GetOrCreate, nodiscard wouldn't be required anyway.

OrbitCore/StringManager.h
pierricgimmig: Done.

OrbitGl/App.h
pierricgimmig: Done (nodiscard).

OrbitGl/GraphTrack.h
pierricgimmig: Done.

OrbitGl/TimeGraph.h
pierricgimmig: Done.

OrbitClientServices/ProcessManager.cpp
```diff
@@ -206,10 +206,15 @@ ErrorMessageOr<std::string> ProcessManagerImpl::LoadProcessMemory(
...
 ErrorMessageOr<std::string> ProcessManagerImpl::LoadNullTerminatedString(
-    int32_t pid, uint64_t address, uint64_t max_size) {
+    int32_t pid, uint64_t address) {
+  constexpr uint64_t max_size = 256;
   auto error_or_string = LoadProcessMemory(pid, address, max_size);
   if (error_or_string.has_value()) {
-    std::string str = error_or_string.value();
+    const std::string& str = error_or_string.value();
+    if (str.find('\0') == std::string::npos) {
+      ERROR("Remote string is not null terminated: %s", str.c_str());
```
dimitry-, line 215: maybe return a error in this case, is there a valid case for processing non-terminated string?

OrbitGl/TimeGraph.cpp
```diff
@@ -848,6 +936,43 @@ std::shared_ptr<GpuTrack> TimeGraph::GetOrCreateGpuTrack(
...
+                                         uint64_t string_address,
+                                         ThreadPool* thread_pool) {
+  thread_pool->Schedule([track, string_address]() {
+    int32_t pid = Capture::GTargetProcess->GetID();
+    const auto& process_manager = GOrbitApp->GetProcessManager();
+    auto error_or_string =
+        process_manager->LoadNullTerminatedString(pid, string_address);
+
+    if (error_or_string.has_value()) {
+      GOrbitApp->GetMainThreadExecutor()->Schedule([track, error_or_string]() {
```
dimitry-, line 950: How about executor - can it be passed as parameter too?

OrbitGl/TimeGraph.cpp
```diff
@@ -848,6 +936,43 @@ std::shared_ptr<GpuTrack> TimeGraph::GetOrCreateGpuTrack(
...
+}
+
+//-----------------------------------------------------------------------------
+std::shared_ptr<GraphTrack> TimeGraph::GetOrCreateGraphTrack(
+    uint64_t graph_id) {
+  ScopeLock lock(m_Mutex);
+  std::shared_ptr<GraphTrack> track = graph_tracks_[graph_id];
+  if (track == nullptr) {
+    track = std::make_shared<GraphTrack>(this, graph_id);
+    SetTrackNameFromRemoteMemory(track, graph_id, GOrbitApp->GetThreadPool());
```
dimitry-, line 968: Maybe pass pointer to OrbitApp to TimeGraph c-tor and use that pointer instead of the global?

OrbitCaptureClient/CaptureClient.cpp
```diff
@@ -6,14 +6,25 @@
...
 using orbit_client_protos::FunctionInfo;
 
+static CaptureOptions::InstrumentedFunction::FunctionType
+IntrumentedFunctionTypeFromOrbitType(FunctionInfo::OrbitType orbit_type) {
+  switch (orbit_type) {
+    case FunctionInfo::kOrbitTimerStart:
+      return CaptureOptions::InstrumentedFunction::kTimerStart;
+    case FunctionInfo::kOrbitTimerStop:
+      return CaptureOptions::InstrumentedFunction::kTimerStop;
+    default:
```
dpallotti, line 23: There are several other values for `OrbitType`, should they all be mapped to `kRegular`?

OrbitCore/StringManager.cpp
```diff
@@ -13,7 +13,14 @@ bool StringManager::AddIfNotPresent(uint64_t key, std::string_view str) {
   return true;
 }
 
-std::optional<std::string> StringManager::Get(uint64_t key) {
+bool StringManager::AddOrReplace(uint64_t key, std::string_view str) {
```
dpallotti, line 16: The return value is not that clear from the method name. The caller needs to look at the implementation to know if `true` means "added" or "replaced". Not too sure what a better name would be though...

OrbitCore/StringManager.cpp
```diff
@@ -13,7 +13,14 @@ bool StringManager::AddIfNotPresent(uint64_t key, std::string_view str) {
...
+  absl::MutexLock lock{&mutex_};
+  bool added = !key_to_string_.contains(key);
+  key_to_string_.insert_or_assign(key, str);
```
dpallotti, line 19: This returns a pair whose first element is `true` if the insertion took place, this should save you the `contains`.

OrbitGl/App.h
```diff
@@ -233,6 +233,14 @@ class OrbitApp final : public DataViewFactory, public CaptureListener {
 
   DataView* GetOrCreateDataView(DataViewType type) override;
 
+  const std::unique_ptr<ProcessManager>& GetProcessManager() {
```
dpallotti, line 236: Returning (just like passing) a `const std::unique_ptr&` is weird. Return a pointer (if you need to check for the `nullptr` case) or a reference.

dpallotti, line 236: Also should this be `nodiscard`, since the other methods that you added are also `nodiscard`?

OrbitGl/GraphTrack.cpp
```diff
@@ -72,7 +74,27 @@ void GraphTrack::Draw(GlCanvas* canvas, PickingMode /*picking_mode*/) {
...
+  if (value_range_ > 0) inv_value_range_ = 1.0 / value_range_;
+}
+
+double GraphTrack::GetValueAtTime(uint64_t time, double default_value) const {
+  auto iterator_lower = values_.lower_bound(time);
+  if (iterator_lower != values_.end()) {
+    return iterator_lower->second;
+  }
+  auto iterator_upper = values_.upper_bound(time);
+  if (iterator_upper != values_.end()) {
```
dpallotti, line 92: Doesn't `iterator_lower == values_.end()` imply `iterator_upper == values_.end()`? I.e., isn't this always false?

OrbitGl/GraphTrack.h
```diff
@@ -14,17 +14,21 @@ class TimeGraph;
 
 class GraphTrack : public Track {
  public:
-  explicit GraphTrack(TimeGraph* time_graph);
+  explicit GraphTrack(TimeGraph* time_graph, uint64_t graph_id);
```
dpallotti, line 17: Remove `explicit` now?

OrbitGl/TimeGraph.cpp
```diff
@@ -377,6 +383,88 @@ void TimeGraph::ProcessTimer(const TimerInfo& timer_info) {
...
+void TimeGraph::ProcessOrbitFunctionTimer(const FunctionInfo* function,
+                                          const TimerInfo& timer_info) {
+  FunctionInfo::OrbitType type = function->type();
+  uint64_t time = timer_info.start();
+  CHECK(timer_info.registers_size() > 0);
+  uint64_t graph_id = timer_info.registers(0);
+
+  switch (type) {
+    case FunctionInfo::kOrbitTrackInt: {
+      int32_t value = static_cast<int32_t>(timer_info.registers(1));
```
dpallotti, line 396: This is why we need an explicit mapping of registers. What registers is this? Why 1? Either way, at least use a constant in place of `1` so that we give it a name.

OrbitGl/TimeGraph.cpp
```diff
@@ -377,6 +383,88 @@ void TimeGraph::ProcessTimer(const TimerInfo& timer_info) {
...
+      uint64_t value = static_cast<uint64_t>(timer_info.registers(1));
+      auto track = GetOrCreateGraphTrack(graph_id);
+      track->AddValue(value, time);
+    } break;
+    case FunctionInfo::kOrbitTrackFloatAsInt: {
+      int32_t int_value = static_cast<int32_t>(timer_info.registers(1));
+      float value = *(reinterpret_cast<float*>(&int_value));
+      auto track = GetOrCreateGraphTrack(graph_id);
+      track->AddValue(value, time);
+      break;
```
dpallotti, line 420: Maybe put this and the following `break`s out of the `}` like the previous ones, for consistency.

OrbitGl/TimeGraph.cpp
```diff
@@ -377,6 +383,88 @@ void TimeGraph::ProcessTimer(const TimerInfo& timer_info) {
...
+    case FunctionInfo::kOrbitTimerStart:
+      ProcessManualIntrumentationTimer(timer_info);
+      break;
+    default:
+      break;
+  }
+}
+
+void TimeGraph::ProcessManualIntrumentationTimer(const TimerInfo& timer_info) {
+  uint64_t string_address = timer_info.registers(0);
```
dpallotti, line 438: Use a constant for 0.

OrbitGl/TimeGraph.cpp
```diff
@@ -377,6 +383,88 @@ void TimeGraph::ProcessTimer(const TimerInfo& timer_info) {
...
+      int32_t pid = Capture::GTargetProcess->GetID();
+      const auto& process_manager = GOrbitApp->GetProcessManager();
+      auto error_or_string =
+          process_manager->LoadNullTerminatedString(pid, string_address);
+
+      if (error_or_string.has_value()) {
+        manual_instrumentation_string_manager_.AddOrReplace(
+            string_address, error_or_string.value());
+      } else {
+        ERROR("Error loading remote string %s",
```
dpallotti, line 455:
```suggestion
        ERROR("Loading remote string: %s",
```

("Error" is already printed by the macro.)

OrbitGl/TimeGraph.cpp
```diff
@@ -848,6 +936,43 @@ std::shared_ptr<GpuTrack> TimeGraph::GetOrCreateGpuTrack(
...
+      });
+    } else {
+      ERROR("Setting track name from remote memory: %s",
+            error_or_string.error().message());
+    }
+  });
+}
+
+//-----------------------------------------------------------------------------
+std::shared_ptr<GraphTrack> TimeGraph::GetOrCreateGraphTrack(
```
dpallotti, line 962: Does the ownership of the returned `GraphTrack` need to be shared with the caller or can you return a `GraphTrack*`?

OrbitTest/OrbitTest.cpp
```diff
@@ -99,11 +99,24 @@ void NO_INLINE OrbitTest::BusyWork(uint64_t microseconds) {
   }
 }
 
+//-----------------------------------------------------------------------------
```
dpallotti, line 102: Remove this comment.

OrbitTest/OrbitTest.cpp
```diff
@@ -99,11 +99,24 @@ void NO_INLINE OrbitTest::BusyWork(uint64_t microseconds) {
...
+void NO_INLINE SleepFor1Milli() {
+  std::this_thread::sleep_for(std::chrono::milliseconds(1));
+}
+
+//-----------------------------------------------------------------------------
```
dpallotti, line 107: Same.

OrbitTest/OrbitTest.cpp
```diff
@@ -99,11 +99,24 @@ void NO_INLINE OrbitTest::BusyWork(uint64_t microseconds) {
   }
 }
 
+//-----------------------------------------------------------------------------
+void NO_INLINE SleepFor1Milli() {
```
dpallotti, line 103: `SleepFor1Ms`?

OrbitGl/TimeGraph.cpp
pierricgimmig: Well it doesn't need to be shared now no, removing the std::shared_ptr.

OrbitTest/OrbitTest.cpp
pierricgimmig: will get removed in bulk.

OrbitTest/OrbitTest.cpp
pierricgimmig: will get removed in bulk.

OrbitTest/OrbitTest.cpp
pierricgimmig: sure, done.

OrbitGl/GraphTrack.cpp
dimitry-: yep - it looks to me too like upper_bound here is always pointing to the end()

OrbitGl/TimeGraph.cpp
```diff
@@ -802,6 +898,43 @@ std::shared_ptr<GpuTrack> TimeGraph::GetOrCreateGpuTrack(
...
+        track->SetLabel(error_or_string.value());
+      });
+    } else {
+      ERROR("Setting track name from remote memory: %s",
+            error_or_string.error().message());
+    }
+  });
+}
+
+//-----------------------------------------------------------------------------
```
dimitry-, line 923: these commends have been removed from the code recently. Can you remove them from this change as well?

OrbitCaptureClient/CaptureClient.cpp
dpallotti: Mmmh, this is out of the scope of this change and complicates things, but why do you need the uretprobe for types like `kOrbitTrackInt`? Don't you only need the parameters?

OrbitCore/StringManager.cpp
dpallotti: Sounds good.

OrbitGl/GraphTrack.h
dpallotti: I'm all for it if it's on purpose, I was pointing it out as it's something we don't usually do anywhere else.

OrbitTest/OrbitTest.cpp
dpallotti: There are none left in the codebase.

OrbitGl/App.h
dpallotti: Trust me, I feel you...

OrbitCaptureClient/CaptureClient.cpp
pierricgimmig: We have the duplicate uprobes on thread migration issue for which we have a workaround that relies on uprobe-uretprobe pairs.

OrbitGl/TimeGraph.cpp
dpallotti: Has this been addressed?

OrbitGl/GraphTrack.cpp
dpallotti: Has this been addressed?

PR #928, opened 2020-08-07
Integration of Chromium into Orbit

beckerhe:
This PR lays all the ground work for integrating Chromium into Orbit. It does not bundle with any UI changes. It merely integrates Chromium on the build system level.

Since this is a complex step, requiring a lot of new dependencies, changes to the CI and changes to the docker containers I decided to create a PR which only includes the groundwork with no actual user facing change.

The only thing that can be observed on the UI is the greater number of third party dependencies in our auto-generated NOTICE file.

To the technical details:

- Chromium is brought in Qt (QtWebEngine) which saves a lot of integration work on our side.
- On Linux we use the system-provided Qt-packages, so the bootstrap-script now installs additional QtWebEngine packages.
- On Windows we use conan-provided Qt-packages. Since the Chromium-version bundled with Qt does not compile with MSVC2017 (apparently due to a regression in MSVC2017) I needed to fork the publicly available conan Qt package and add a patch for MSVC2017. This patch is not needed for MSVC2019 but it also does no harm. When we move away from MSVC2017 we should remove that patch again.
- Due to the sheer number of third party licenses that are pulled in by Chromium some adjustments to our license collection script had to be made such that it is not picking up false positives (files with license in its name which aren't actually license files.)


conanfile.py
```diff
@@ -93,7 +93,7 @@ def requirements(self):
             if not self.options.system_mesa:
                 self.requires("libxi/1.7.10@bincrafters/stable#0")
             if not self.options.system_qt:
-                self.requires("qt/5.14.1@bincrafters/stable#0")
+                self.requires("qt/5.15.0@{}#3b376b030c38da6378a2c2cea5f99418".format(self._orbit_channel))
```
florian-kuebler, line 96: Any reason for that specific version? Maybe add a comment?

beckerhe: I went this the latest one for having the latest version of Chromium. What should I write in the comment? If it's about "why that version change", I feel this should be part of the commit message. I could elaborate there if you prefer.

beckerhe:
> Thanks! LGTM. But to be honest, did not look too closely to the third_party recipes files. Please re-request review, if you feel this is needed.

Well, not really. :-) Most of the files there are forked (copied) from somewhere else. My changes are minor. But thanks for offering that.

beckerhe:
> Why is it that the QT version is only present in the windows lock-files?
> Otherwise LGTM, same as Florian I did not check all details in the recipes.

Excellent question: We only use the conan-qt-package on Windows. On our Linux-build we rely on the system-provided version of Qt. That has the following reason: In the past we had a variety of problems when providing our own version of Qt:
- System fonts were not detected due to incompatible versions of fontconfig
- Orbit crashed when opening a capture because the native file dialog ABI was different.
- Orbit didn't start at all due to incompatible version of glib.
- and so on...

The simplest solution was to use system libraries. I guess we have to reevaluate that when we want to package linux-binaries. But for now that's the best solution.

beckerhe:
I will reopen that PR when the license situation is clarified.

PR #929, opened 2020-08-07
check for nullptr

IrinaShkviro:
Added check for nullptr

### DIFF
--- a/OrbitCaptureClient/CaptureClient.cpp
+++ b/OrbitCaptureClient/CaptureClient.cpp
@@ -40,13 +40,16 @@ void CaptureClient::Capture(
   }
   capture_options->set_trace_gpu_driver(true);
   for (const auto& pair : selected_functions) {
-    const FunctionInfo& function = *pair.second;
+    const FunctionInfo* function = pair.second;
+    if (function == nullptr) {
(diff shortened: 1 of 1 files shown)


OrbitCaptureClient/CaptureClient.cpp
```diff
@@ -40,13 +40,15 @@ void CaptureClient::Capture(
...
   for (const auto& pair : selected_functions) {
-    const FunctionInfo& function = *pair.second;
-    CaptureOptions::InstrumentedFunction* instrumented_function =
-        capture_options->add_instrumented_functions();
-    instrumented_function->set_file_path(function.loaded_module_path());
-    instrumented_function->set_file_offset(FunctionUtils::Offset(function));
-    instrumented_function->set_absolute_address(
-        FunctionUtils::GetAbsoluteAddress(function));
+    const FunctionInfo* function = pair.second;
+    if (function != nullptr) {
```
dimitry-, line 44: if (function == nullptr) {
  continue;
}

dimitry-: I wonder why was not cleared but proper fix should probably go in master.

PR #934, opened 2020-08-07
Remove some unused global from Capture

dimitry-:
There is more... for the next change.

### DIFF
--- a/OrbitCore/Capture.cpp
+++ b/OrbitCore/Capture.cpp
@@ -31,15 +31,12 @@ using orbit_client_protos::PresetInfo;
 Capture::State Capture::GState = Capture::State::kEmpty;
 bool Capture::GInjected = false;
 std::string Capture::GInjectedProcess;
-bool Capture::GIsSampling = false;
-uint32_t Capture::GFunctionIndex = -1;
 uint32_t Capture::GNumInstalledHooks;
 bool Capture::GHasContextSwitches;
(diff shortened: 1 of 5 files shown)


OrbitCore/Capture.cpp
```diff
@@ -230,18 +225,6 @@ void Capture::NewSamplingProfiler() {
...
 
-//-----------------------------------------------------------------------------
-void Capture::RegisterZoneName(uint64_t a_ID, const char* a_Name) {
-  GZoneNames[a_ID] = a_Name;
-}
-
-//-----------------------------------------------------------------------------
-void Capture::AddCallstack(CallStack& a_CallStack) {
-  ScopeLock lock(GCallstackMutex);
-  Capture::GCallstacks[a_CallStack.Hash()] =
```
IrinaShkviro, line 241: Should GCallstacks also be removed?

dimitry-: Interesting - apparently CaptureWindow uses it to select a callstack. It does not work now (because the map is empty) - but I wonder if it should. I will take a closer look. As it is now - it seems like this is unused as well.

dimitry-: It looks like this is a bug - at some point we lost this map and callstack selection stopped working. I am investigating,

dimitry-: this is not a regression from 1.50

PR #936, opened 2020-08-07
Remove unused GCallstacks from Capture

dimitry-:
callstack_hash was always 0 because of b/159107241, once
we have callstacks for uprobes we can revisit this.

### DIFF
--- a/OrbitClientProtos/capture_data.proto
+++ b/OrbitClientProtos/capture_data.proto
@@ -92,7 +92,7 @@ message TimerInfo {
   Type type = 6;
 
   int32 processor = 7;
-  uint64 callstack_hash = 8;
+  uint64 callstack_id = 8;
   uint64 function_address = 9;
   uint64 user_data_key = 10;
   uint64 timeline_hash = 11;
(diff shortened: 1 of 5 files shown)


OrbitClientProtos/capture_data.proto
```diff
@@ -92,7 +92,7 @@ message TimerInfo {
   Type type = 6;
 
   int32 processor = 7;
-  uint64 callstack_hash = 8;
+  uint64 callstack_id = 8;
```
dimitry-, line 95: this field is now unused - I do not know if we should remove it or keep it for the future?

IrinaShkviro: I would leave this question for @pierricgimmig : he wanted to leave this field and use it in near future

IrinaShkviro: So probably just keep it for now

dpallotti: This would be used for callstacks of dynamically instrumented functions and scheduling slices I assume.

PR #809, opened 2020-07-13
Add SamplingUtils and move ThreadSampleData methods to it

IrinaShkviro:
As ThreadSampleData is going to be a proto in future, this PR move it's methods out of the struct.

PR #942, opened 2020-08-07
remove SamplingUtils

IrinaShkviro:
SamplingUtils introduced as we planned to replace ThreadSampleData with proto for capture serialization. When we started to serialize raw data, it became unnecessary change, so I reverted this part of https://github.com/google/orbit/pull/809.


OrbitCore/SamplingProfiler.cpp
```diff
@@ -53,6 +53,14 @@ void ComputeAverageThreadUsage(ThreadSampleData* data) {
 
 }  // namespace
 
+unsigned int ThreadSampleData::GetCountForAddress(uint64_t address) const {
```
dimitry-, line 56: uint32_t

OrbitCore/SamplingProfiler.h
```diff
@@ -34,6 +34,7 @@ struct SampledFunction {
 //-----------------------------------------------------------------------------
 struct ThreadSampleData {
   ThreadSampleData() { m_ThreadUsage.push_back(0); }
+  [[nodiscard]] unsigned int GetCountForAddress(uint64_t address) const;
```
dimitry-, line 37: and here.

OrbitCore/SamplingProfiler.cpp
IrinaShkviro: Done, thank you!

PR #943, opened 2020-08-07
Synchronize access to Address Infos in Capture

dimitry-:
This will let us (relatively) safely add address infos from different threads.
Again this is not final solution this is a quick fix to be replaces by
proper capture model.


OrbitCore/Capture.cpp
```diff
@@ -189,8 +191,8 @@ void Capture::NewSamplingProfiler() {
 
 //-----------------------------------------------------------------------------
 LinuxAddressInfo* Capture::GetAddressInfo(uint64_t address) {
-  auto address_info_it = GAddressInfos.find(address);
-  if (address_info_it == GAddressInfos.end()) {
+  auto address_info_it = address_infos_.find(address);
```
dimitry-, line 194: hmm... seems like something is missing here.

dimitry-:
This is incorrect - need to redo it. Please ignore.

PR #945, opened 2020-08-08
Use rvalue refererce to pass CallstackEvent

dimitry-:
Change SamplingProfiler::AddCallStack parameter to rvalue reference.

### DIFF
--- a/OrbitGl/TimeGraph.cpp
+++ b/OrbitGl/TimeGraph.cpp
@@ -546,20 +546,20 @@ std::vector<CallstackEvent> TimeGraph::SelectEvents(float a_WorldStart,
   }
 
   // Generate selection report.
-  std::shared_ptr<SamplingProfiler> samplingProfiler =
+  std::shared_ptr<SamplingProfiler> sampling_profiler =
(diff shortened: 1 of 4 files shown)


OrbitGl/TimeGraph.cpp
```diff
@@ -546,20 +546,20 @@ std::vector<CallstackEvent> TimeGraph::SelectEvents(float a_WorldStart,
...
   // Generate selection report.
-  std::shared_ptr<SamplingProfiler> samplingProfiler =
+  std::shared_ptr<SamplingProfiler> sampling_profiler =
       std::make_shared<SamplingProfiler>(Capture::GTargetProcess);
 
-  samplingProfiler->SetGenerateSummary(a_TID == 0);
+  sampling_profiler->SetGenerateSummary(a_TID == 0);
 
-  for (CallstackEvent& event : selected_callstack_events) {
+  for (CallstackEvent event : selected_callstack_events) {
```
dimitry-, line 554: actually - this does not look good - it copies an event even if it is not added. will probably need to move the copy inside the if.

PR #947, opened 2020-08-10
Remove unused functions from PerfEventOpen.h/.cpp and PerfEventRecords.h

dpallotti:


OrbitLinuxTracing/PerfEventOpen.cpp
```diff
@@ -109,17 +109,6 @@ int uprobes_retaddr_event_open(const char* module, uint64_t function_offset,
   return generic_event_open(&pe, pid, cpu);
 }
 
-int uprobes_stack_event_open(const char* module, uint64_t function_offset,
```
dimitry-, line 112: since I cannot leave a review on commit message:

Please fix it it should be "Remove unused functions from ...."

dpallotti: Done.

PR #948, opened 2020-08-10
Added clang_format CI presubmit check

beckerhe:
This PR adds a clang_format presubmit check.

This check evaluates whether all your made changes fulfill our code formatting standard by calling `clang-format-diff` on them.
Only changed lines are taken into account for the result of the presubmit change.

However the script always checks all *.cpp/*.h file in the repository (and not under third_party/) for compliance. This result will only be shown in the logfile.

### DIFF
--- a/third_party/conan/docker/build_containers.sh
+++ b/third_party/conan/docker/build_containers.sh
@@ -32,10 +32,11 @@ declare -A profile_to_dockerfile=( \
   ["msvc2019_debug"]="msvc2019" \
   ["msvc2019_release_x86"]="msvc2019" \
   ["msvc2019_relwithdebinfo_x86"]="msvc2019" \
-  ["msvc2019_debug_x86"]="msvc2019" )
+  ["msvc2019_debug_x86"]="msvc2019" \
+  ["clang_format"]="clang_format" )
 
 if [ "$(uname -s)" == "Linux" ]; then
-  for profile in {clang{7,8,9},gcc{8,9},ggp}_{release,relwithdebinfo,debug}; do
+  for profile in {clang{7,8,9},gcc{8,9},ggp}_{release,relwithdebinfo,debug} clang_format; do
     docker build "${DIR}" -f "${DIR}/Dockerfile.${profile_to_dockerfile[$profile]}" -t gcr.io/orbitprofiler/$profile:latest || exit $?
   done
 else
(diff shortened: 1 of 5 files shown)


third_party/conan/docker/build_containers.sh
```diff
@@ -32,10 +32,11 @@ declare -A profile_to_dockerfile=( \
   ["msvc2019_debug"]="msvc2019" \
   ["msvc2019_release_x86"]="msvc2019" \
   ["msvc2019_relwithdebinfo_x86"]="msvc2019" \
-  ["msvc2019_debug_x86"]="msvc2019" )
+  ["msvc2019_debug_x86"]="msvc2019" \
+  ["clang_format"]="clang_format")
```
florian-kuebler, line 36: Is there a space at the end right before the parenthesis missing?

kokoro/gcp_ubuntu/presubmit.cfg
```diff
@@ -1,24 +1,10 @@
...
-    }
-  }
-}
-
-action {
-  define_artifacts {
-    regex: "github/orbitprofiler/build/testresults/*.xml"
-    strip_prefix: "github/orbitprofiler"
-  }
-}
```
florian-kuebler: To be reverted for non-draft PR.

third_party/conan/docker/build_containers.sh
beckerhe: Maybe. I will add one. Thanks!

kokoro/gcp_ubuntu/presubmit.cfg
beckerhe: Already gone. Thanks!

third_party/conan/docker/build_containers.sh
beckerhe: Done.

beckerhe:
The discussed changes were made. Please have another look! Thanks.

beckerhe:
> I understand this correctly that you're doing the same check as before (using the xml result), but soley for information - the return code is then simply determined by the result of the git clang format diff being non-empty?

Yes I left the previous check in there for purely informational purposes. I think it's nice to be able to verify from time to time that nothing sneaked into the codebase. Or that an update of the clang-format-tool made a mess or whatever.

PR #949, opened 2020-08-10
Remove unused OutputDialog

dpallotti:


OrbitQt/processlauncherwidget.ui
```diff
@@ -27,12 +16,7 @@
     <number>0</number>
    </property>
    <item row="0" column="0">
-    <widget class="QSplitter" name="splitter">
-     <property name="orientation">
-      <enum>Qt::Vertical</enum>
-     </property>
-     <widget class="OrbitDataViewPanel" name="LiveProcessList" native="true"/>
-    </widget>
+    <widget class="OrbitDataViewPanel" name="LiveProcessList"/>
```
beckerhe, line 19: I think you'll need that `native="true"` otherwise QtDesigner will add it when opened next time. I recently made all the UI-files compatible with QtDesigner. Have you tried opening them with designer? It would be great if we could preserve the compatibility.

dpallotti: I made these changes with the designer, so I'm missing something.

dpallotti: But I can just add `native="true"` back if you prefer.

beckerhe: If the designer made these changes, I'm just wrong. I apologize. :innocent:

beckerhe: > But I can just add `native="true"` back if you prefer.

Not at all. I was just the case for that designer always added that by itself.

dpallotti: Maybe we just have a different setup? For example, I'm doing this on Linux.

dpallotti:
> Thanks! This looks good, but it would be great if you could check whether the UI-changes are compatible with QtDesigner.

As I said, I made the changes with the designer, so this should be fine.

PR #953, opened 2020-08-10
clear GSelectedFunctionsMap when start a new capture

IrinaShkviro:
Permanent fix for http://b/163091878.

### DIFF
--- a/OrbitCaptureClient/CaptureClient.cpp
+++ b/OrbitCaptureClient/CaptureClient.cpp
@@ -40,18 +40,13 @@ void CaptureClient::Capture(
   }
   capture_options->set_trace_gpu_driver(true);
   for (const auto& pair : selected_functions) {
-    const FunctionInfo* function = pair.second;
-    // TODO: this is temporary fix. We should understand why in
-    // Capture::GSelectedFunctionsMap could be nullptrs and fix it accordingly
-    if (function == nullptr) {
-      continue;
-    }
+    const FunctionInfo& function = *pair.second;
     CaptureOptions::InstrumentedFunction* instrumented_function =
         capture_options->add_instrumented_functions();
-    instrumented_function->set_file_path(function->loaded_module_path());
-    instrumented_function->set_file_offset(FunctionUtils::Offset(*function));
+    instrumented_function->set_file_path(function.loaded_module_path());
+    instrumented_function->set_file_offset(FunctionUtils::Offset(function));
     instrumented_function->set_absolute_address(
-        FunctionUtils::GetAbsoluteAddress(*function));
+        FunctionUtils::GetAbsoluteAddress(function));
   }
 
   if (!reader_writer_->Write(request)) {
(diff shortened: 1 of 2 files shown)


OrbitCore/Capture.cpp
```diff
@@ -100,6 +100,7 @@ void Capture::ClearCaptureData() {
 void Capture::PreFunctionHooks() {
   GSelectedInCaptureFunctions = GetSelectedFunctions();
 
+  GSelectedFunctionsMap.clear();
```
dimitry-, line 103: TBH - I do not understand what sequence of events leads to the situation where selected functions is nullptr, and why calling clear here fixes it. can you please explain it in the commit message?

IrinaShkviro: Steps to reproduce:

1. Start Orbit -> hook function(s) -> start capture -> save capture
2. Restart Orbit ans select process (but don't hook functions)
3. Load the capture from the first step (we fill `Capture::GSelectedFunctionsMap` on this step)
4. Start a new capture. Capture::GSelectedInCaptureFunctions is empty (we updated it 2 lines above) when GSelectedFunctionsMap is not empty (we didn't clear it in any point and it's still have data from the saved capture)

So when we start a capture GSelectedInCaptureFunctions and GSelectedFunctionsMap should be synchronized, but in case described above we see discrepancy. We won't have this situation if open capture without connection.

I updated commit message too.

florian-kuebler: Maybe a bit more detail to add here:
`GetSelectedFunctions()` basically runs over all loaded functions of a process and check whether it is currently in the `GSelectedFunctionsMap` or a special Orbit function. If one of this is the case, it will be returned.
So if your process does not have any functions (because you loaded the capture and did nothing else before), selected functions will be empty, but the map still contains the functions from loading the capture.

OrbitCaptureClient/CaptureClient.cpp
```diff
@@ -40,18 +40,13 @@ void CaptureClient::Capture(
   }
   capture_options->set_trace_gpu_driver(true);
   for (const auto& pair : selected_functions) {
-    const FunctionInfo* function = pair.second;
-    // TODO: this is temporary fix. We should understand why in
-    // Capture::GSelectedFunctionsMap could be nullptrs and fix it accordingly
-    if (function == nullptr) {
```
florian-kuebler, line 46: If I remember correctly, there are a couple of other places where we check function for nullptr. I guess we should also remove them (maybe in a different PR).

IrinaShkviro:
> I still think we should also clear the maps for now, when we load a capture.

I'm not sure that we should do it. We use this map in Live tab for example and show if function hooked based on this map. Also clearing this map on loading means that we should hook all the functions again if we save capture, then load back and start capture. As we allow to start a new capture after loading the capture, we should leave this map for now.

florian-kuebler:
> > I still think we should also clear the maps for now, when we load a capture.
> 
> I'm not sure that we should do it. We use this map in Live tab for example and show if function hooked based on this map. Also clearing this map on loading means that we should hook all the functions again if we save capture, then load back and start capture. As we allow to start a new capture after loading the capture, we should leave this map for now.

Should be also fine for now. We should reconsider, when we have further specifications of how we want allow new capture after loading. So For this PR, this should be fine. Thanks!

PR #955, opened 2020-08-10
License header presubmit check

beckerhe:
This PR adds a presubmit for license headers in cpp-, header- and CMakeLists.txt files. It will fail if our typical copyright notice is missing.

The check only tests the file which have been touched by the current PR. Everything else will not be examined.

### DIFF
--- a/kokoro/checks/license_headers/check.sh
+++ b/kokoro/checks/license_headers/check.sh
@@ -1,0 +1,42 @@
+#!/bin/bash
+#
+# Copyright (c) 2020 The Orbit Authors. All rights reserved.
+# Use of this source code is governed by a BSD-style license that can be
+# found in the LICENSE file.
+
+# Fail on any error.
+readonly DIR="/mnt/github/orbitprofiler"
+readonly SCRIPT="${DIR}/kokoro/checks/license_headers/check.sh"
+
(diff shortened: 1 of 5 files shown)


kokoro/checks/license_headers/check.sh
```diff
@@ -0,0 +1,42 @@
...
+  MERGE_BASE="$(git merge-base $REFERENCE HEAD)" # Merge base is the commit on master this PR was branched from.
+  LICENSE_HEADER_MISSED=""
+
+  echo -e "The following files are missing a license/copyright header:\n"
+  while read file; do
+    if [ ! -f "$file" ]; then
+      continue
+    fi
+
+    if ! grep 'Copyright (c) 2020 The Orbit Authors. All rights reserved.' "$file" > /dev/null; then
```
reichlfl, line 25: This will break when we add files in a few month - would it make sense to replace "2020" with a "\d{4}" regex?

beckerhe: Yes, definitely. Thanks!

beckerhe: Done. PTAL!

reichlfl: LGTM, thanks!

PR #959, opened 2020-08-10
OrbitClientGgp: project structure and basic capture flow

kisenshi:
### Create OrbitClientGgp to manage a capture request from the ggp instance

Main files and the structure of the project has been created. Basic flow is working: initial set up, starts capture, captures for a period of time and stops the capture.

Capture data is not collected or handled yet; this pull request is intended to include the project in Orbit so there is not a huge list of commits later. Also it avoids being 'silently' broken in updates related to Capture, etc. 

The options related to the capture cycle must be provided as an argument to OrbitClientGgp. Some values are given a default value but others are required to be provided by the user:

* pid. (Mandatory) Id of the process to capture.
* grpc_port. (Optional) Required for gRPC, by default is 44765.
* sampling_rate. (Optional) Required for CaptureClient to work, by default is 1000.
* frame_pointer_unwinding. (Optional) Required for CaptureClient to work.
* capture_length. (Optional) Sets the length of the capture in seconds, by default is 10

The new project contains the following files:

+ main.cpp Contains the linear flow and handles the different classes instantiation and requests
+ ClientGgp.h/cc Contains methods that handle the gRPC client and capture requests. Extends and implements CaptureListener (although it does not handle the calls yet).
+ ClientGgpOptions.h  Struct containing the options needed for the client initialisation.

Also a Run_client_ssh.ps1 script has been created to facilitate the deployment and test of OrbitClientGgp.

### DIFF
--- a/OrbitClientGgp/CMakeLists.txt
+++ b/OrbitClientGgp/CMakeLists.txt
@@ -1,0 +1,25 @@
+# Copyright (c) 2020 The Orbit Authors. All rights reserved.
+# Use of this source code is governed by a BSD-style license that can be
+# found in the LICENSE file.
+
+project(OrbitClientGgp)
+
+add_executable(OrbitClientGgp main.cpp)
+
+target_compile_options(OrbitClientGgp PRIVATE
+    ${STRICT_COMPILE_FLAGS})
+
+target_include_directories(OrbitClientGgp PRIVATE 
+    ${CMAKE_CURRENT_LIST_DIR})
+
+target_sources(OrbitClientGgp PRIVATE
+    ClientGgp.cpp
+    ClientGgp.h
+    main.cpp)
+
+target_link_libraries(OrbitClientGgp PUBLIC 
+    OrbitBase
+    OrbitCaptureClient
+    OrbitProtos)
+
+#strip_symbols(OrbitClientGgp)
--- a/OrbitClientGgp/CMakeLists.txt
+++ b/OrbitClientGgp/CMakeLists.txt
@@ -21,5 +21,4 @@ target_link_libraries(OrbitClientGgp PUBLIC
     OrbitBase
     OrbitCaptureClient
     OrbitProtos)
-
-#strip_symbols(OrbitClientGgp)
+    
--- a/OrbitClientGgp/ClientGgp.cpp
+++ b/OrbitClientGgp/ClientGgp.cpp
@@ -15,6 +15,15 @@
 #include "OrbitBase/Result.h"
 #include "capture_data.pb.h"
 
+namespace {
+std::shared_ptr<Process> GetOrbitProcessByPid(int32_t pid) {
+  std::shared_ptr<Process> process = std::make_shared<Process>();
+  // TODO: study if we need more information. Requires extra steps
+  process->SetID(pid);
+  return process;
+}
+}  // namespace
+
 ClientGgp::ClientGgp(ClientGgpOptions&& options)
     : options_(std::move(options)) {}
 
@@ -75,13 +84,6 @@ void ClientGgp::InitCapture() {
   std::shared_ptr<Process> process = GetOrbitProcessByPid(options_.capture_pid);
   CHECK(process != nullptr);
   Capture::SetTargetProcess(std::move(process));
-}
-
-std::shared_ptr<Process> ClientGgp::GetOrbitProcessByPid(int32_t pid) {
-  std::shared_ptr<Process> process = std::make_shared<Process>();
-  // TODO: study if we need more information. Requires extra steps
-  process->SetID(pid);
-  return process;
 }
 
 // CaptureListener implementation
--- a/OrbitClientGgp/ClientGgp.cpp
+++ b/OrbitClientGgp/ClientGgp.cpp
@@ -45,7 +45,8 @@ bool ClientGgp::InitClient() {
   return true;
 }
 
-bool ClientGgp::StartCapture() {
+// Prepare the Capture object to be able to request a capture
+bool ClientGgp::PrepareStartCapture() {
   CHECK(!Capture::IsCapturing());
 
   ErrorMessageOr<void> result = Capture::StartCapture();
@@ -54,7 +55,11 @@ bool ClientGgp::StartCapture() {
     return false;
   }
 
-  // Client capture request
+  return true;
+}
+
+// Client requests to start the capture
+void ClientGgp::RequestStartCapture() {
   int32_t pid = Capture::GProcessId;
   LOG("Capture pid %d", pid);
 
@@ -62,19 +67,17 @@ bool ClientGgp::StartCapture() {
   std::vector<std::shared_ptr<orbit_client_protos::FunctionInfo>> selected_functions =
       Capture::GSelectedFunctions;
   capture_client_->Capture(pid, selected_functions);
-  LOG("Capture started");
-
-  return true;
+  
+  LOG("Start capture requested");
 }
 
 void ClientGgp::StopCapture() {
-  LOG("Stop captured requested");
   CHECK(Capture::IsCapturing());
   Capture::StopCapture();
 
   capture_client_->StopCapture();
 
-  LOG("Capture stopped");
+  LOG("Stop capture requested");
 }
 
 void ClientGgp::InitCapture() {
--- a/OrbitClientGgp/ClientGgp.cpp
+++ b/OrbitClientGgp/ClientGgp.cpp
@@ -1,0 +1,5 @@
+// Copyright (c) 2020 The Orbit Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#include "ClientGgp.h"
--- a/OrbitClientGgp/ClientGgp.cpp
+++ b/OrbitClientGgp/ClientGgp.cpp
@@ -67,8 +67,6 @@ void ClientGgp::RequestStartCapture() {
   std::map<uint64_t, orbit_client_protos::FunctionInfo*> selected_functions =
       Capture::GSelectedFunctionsMap;
   capture_client_->Capture(pid, selected_functions);
-  
-  LOG("Start capture requested");
 }
 
 void ClientGgp::StopCapture() {
@@ -95,9 +93,13 @@ std::shared_ptr<Process> ClientGgp::GetOrbitProcessByPid(int32_t pid) {
 }
 
 // CaptureListener implementation
-void ClientGgp::OnCaptureStarted() {}
+void ClientGgp::OnCaptureStarted() {
+  LOG("Capture started");
+}
 
-void ClientGgp::OnCaptureComplete() {}
+void ClientGgp::OnCaptureComplete() {
+  LOG("Capture completed");
+}
 
 void ClientGgp::OnTimer(const orbit_client_protos::TimerInfo& timer_info) {
   (void)timer_info;
--- a/OrbitClientGgp/ClientGgp.cpp
+++ b/OrbitClientGgp/ClientGgp.cpp
@@ -103,32 +103,26 @@ void ClientGgp::OnCaptureComplete() {
 
 void ClientGgp::OnTimer(const orbit_client_protos::TimerInfo& timer_info) {
   (void)timer_info;
-  //LOG("OnTimer called");
 }
 
 void ClientGgp::OnKeyAndString(uint64_t key, std::string str) {
   (void)key;
   (void)str;
-  //LOG("OnKeyAndString called");
 }
 
 void ClientGgp::OnCallstack(CallStack callstack) {
   (void)callstack;
-  //LOG("OnCallstack called");
 }
 
 void ClientGgp::OnCallstackEvent(orbit_client_protos::CallstackEvent callstack_event) {
   (void)callstack_event;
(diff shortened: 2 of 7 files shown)


OrbitClientGgp/ClientGgp.cpp
```diff
@@ -0,0 +1,134 @@
...
+  LOG("Capture started");
+}
+
+void ClientGgp::OnCaptureComplete() {
+  LOG("Capture completed");
+}
+
+void ClientGgp::OnTimer(const orbit_client_protos::TimerInfo& timer_info) {
+  (void)timer_info;
+  //LOG("OnTimer called");
```
beckerhe, line 106: Can you remove these LOG-statements if they are not needed?

OrbitClientGgp/CMakeLists.txt
```diff
@@ -0,0 +1,25 @@
...
+    ClientGgp.cpp
+    ClientGgp.h
+    main.cpp)
+
+target_link_libraries(OrbitClientGgp PUBLIC 
+    OrbitBase
+    OrbitCaptureClient
+    OrbitProtos)
+
+#strip_symbols(OrbitClientGgp)
```
beckerhe, line 25: Can you remove that line if not needed? Is there a reason why you commented that?

run_client_ssh.ps1
```diff
@@ -0,0 +1,59 @@
+# Connects to the instance and starts OrbitClientGgp. To also deploy OrbitClientGgp,
```
beckerhe, line 1: I think a license header is missing here.

OrbitClientGgp/main.cpp
```diff
@@ -0,0 +1,73 @@
...
+  std::unique_ptr<ThreadPool> thread_pool = ThreadPool::Create(4 /*min_size*/, 256 /*max_size*/, absl::Seconds(1));
+
+  // The request is done in a separate thread to avoid blocking main()
+  thread_pool->Schedule([&]() {
+    client_ggp.RequestStartCapture();
+  });
+
+  LOG("Go to sleep");
+  // Captures for the period of time requested
+  sleep(capture_length);
```
beckerhe, line 61: You're probably calling the POSIX-function `sleep` here, aren't you? Can you replace that with `absl::SleepFor` which makes it more obvious what unit the argument has and which is more portable?

OrbitClientGgp/ClientGgp.cpp
```diff
@@ -0,0 +1,134 @@
...
+// found in the LICENSE file.
+
+#include "ClientGgp.h"
+
+#include "capture_data.pb.h"
+
+#include "Capture.h"
+#include "OrbitBase/Logging.h"
+#include "OrbitBase/Result.h"
+
```
beckerhe, line 12: We should try to explicitly include the headers needed here.

- `std::numeric_limits` needs `<limits>`.
- `std::make_unique` needs `<memory>`. 
- `int32_t` needs `<cstdint>`.
- `std::map` needs `<map>`.
- `std::string` needs `<string>`.

OrbitClientGgp/ClientGgp.h
```diff
@@ -0,0 +1,52 @@
...
+#define ORBIT_CLIENT_GGP_CLIENT_GGP_H_
+
+//#include <string>
+
+#include "grpcpp/grpcpp.h"
+
+//#include "Callstack.h"
+//#include "EventBuffer.h"
+//#include "KeyAndString.h"
+//#include "ScopeTimer.h"
```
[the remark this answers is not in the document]
beckerhe: And add those that are needed for this header? I see `<cstdint>`, `<string>`, and `<memory>`.

OrbitClientGgp/main.cpp
```diff
@@ -0,0 +1,73 @@
...
+ABSL_FLAG(int32_t, pid, 0, "pid to capture");
+ABSL_FLAG(uint32_t, capture_length, 10, "duration of capture in seconds");
+ABSL_FLAG(uint16_t, sampling_rate, 1000,
+          "Frequency of callstack sampling in samples per second");
+ABSL_FLAG(bool, frame_pointer_unwinding, false,
+          "Use frame pointers for unwinding");
+
+int main(int argc, char** argv) {
+  absl::ParseCommandLine(argc, argv);
+  uint64_t grpc_port_ = absl::GetFlag(FLAGS_grpc_port);
```
danielfenner, line 28: Remove underscore at the end - that's for member variables only.

OrbitClientGgp/main.cpp
```diff
@@ -0,0 +1,73 @@
...
+  uint32_t capture_length = absl::GetFlag(FLAGS_capture_length);
+  
+  if (!absl::GetFlag(FLAGS_pid)) {
+    ERROR("pid to capture not provided; set using -pid");
```
danielfenner, line 32: You could just use FATAL here.

OrbitClientGgp/main.cpp
```diff
@@ -0,0 +1,73 @@
...
+  });
+
+  LOG("Go to sleep");
+  // Captures for the period of time requested
+  sleep(capture_length);
+  LOG("Back from sleep");
+
+  client_ggp.StopCapture();
+
+  LOG("Shuting down the thread pool");
```
danielfenner, line 66: Shutting

OrbitClientGgp/main.cpp
```diff
@@ -0,0 +1,73 @@
...
+  LOG("Lets start the capture");
+
+  if(!client_ggp.PrepareStartCapture()){
+    return -1;
+  }
+
+  std::unique_ptr<ThreadPool> thread_pool = ThreadPool::Create(4 /*min_size*/, 256 /*max_size*/, absl::Seconds(1));
+
+  // The request is done in a separate thread to avoid blocking main()
+  thread_pool->Schedule([&]() {
```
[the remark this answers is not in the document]
kisenshi: This makes a lot of sense, I prefer doing this then. Thanks!

OrbitClientGgp/main.cpp
```diff
@@ -0,0 +1,73 @@
...
+// found in the LICENSE file.
+
+//#include <iostream>
+//#include <string>
+
+#include "absl/flags/flag.h"
+#include "absl/flags/parse.h"
+//#include "absl/flags/usage.h"
+//#include "absl/flags/usage_config.h"
+//#include "absl/strings/str_format.h"
```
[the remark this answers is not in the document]
kisenshi: Done, I forgot I had commented them out

OrbitClientGgp/ClientGgp.h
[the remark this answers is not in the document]
kisenshi: Done and added as well

OrbitClientGgp/ClientGgp.cpp
```diff
@@ -0,0 +1,134 @@
...
+void ClientGgp::OnCaptureStarted() {
+  LOG("Capture started");
+}
+
+void ClientGgp::OnCaptureComplete() {
+  LOG("Capture completed");
+}
+
+void ClientGgp::OnTimer(const orbit_client_protos::TimerInfo& timer_info) {
+  (void)timer_info;
```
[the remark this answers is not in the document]
kisenshi: Didn't know this; I updated it, thanks for the tip!

OrbitClientGgp/ClientGgp.cpp
kisenshi: Done

OrbitClientGgp/CMakeLists.txt
kisenshi: Done. I saw it was in the OrbitService so I added and commented out here in case I needed as well. I don't think I do in the end, so I removed it

run_client_ssh.ps1
kisenshi: You are right, added!

OrbitClientGgp/main.cpp
kisenshi: Makes sense, I have updated it, thanks for the suggestion!

OrbitClientGgp/ClientGgp.cpp
kisenshi: Done. I was trying to avoid including headers that are already included by including others; but maybe it makes sense for these, so I have included them.

OrbitClientGgp/main.cpp
kisenshi: True, I didn't realise, thanks!

OrbitClientGgp/main.cpp
kisenshi: Fixed

OrbitClientGgp/main.cpp
kisenshi: I updated the comment because of the use of thread instead

kisenshi:
> Hi Cris,
> I hope you don't mind that I added some minor comments. It looks really good! :-)
> 
> But someone else who is more familiar with what you're trying to do should also have a look. Supposedly Daniel. Thanks.

Hi, 

All feedback is welcome so thanks a lot for taking the time to take a look at the code :)

I have gone through the comments and suggestions and updated the code accordingly. Thanks!

kisenshi:
> Looks good! There is really only some nit picking.

I have updated the code with the suggestions, thanks!

> Besides the points below: maybe have a look into minimize use of vertical whitespace. The style guide asks for as little empty lines as possible. This is not a fixed rule thought, use your best judgement.

Do you mean I should include less empty lines between code? I can take a look and update it. I like including them because I think it helps readability and the separation between different concepts but I can update it to follow the guidelines

kisenshi:
@danielfenner I have removed some of the vertical whitespace as suggested to follow the guidelines. I think the code is still clear like this. Thanks!

kisenshi:
The code will need to be updated as a result of [PR #980](https://github.com/google/orbit/pull/980) which changes Capture and includes the thread creation inside OrbitClient, so I will wait for that to be merged and update the code accordingly. It will need new review when done.

danielfenner:
> Do you mean I should include less empty lines between code? I can take a look and update it. I like including them because I think it helps readability and the separation between different concepts but I can update it to follow the guidelines

Yes exactly, the java style guide encourages exactly this - use empty lines as paragraphs. The c++ style guide asks for the opposite. As I wrote above, just using your own judgement is totaly fine.

OrbitClientGgp/ClientGgp.cpp
beckerhe: Yeah, that's always a thing for debate what headers to include. The Google C++ style guide is pretty clear here though: We should "include all headers defining symbols we rely upon".

Thanks!

OrbitClientGgp/main.cpp
```diff
@@ -0,0 +1,62 @@
...
+
+  ClientGgp client_ggp(std::move(options));
+  if (!client_ggp.InitClient()) {
+    return -1;
+  }
+
+  // The request is done in a separate thread to avoid blocking main()
+  // It is needed to provide a thread pool
+  std::unique_ptr<ThreadPool> thread_pool =
+      ThreadPool::Create(1, 1, absl::Seconds(1));
```
kisenshi, line 39: I am not sure about this initialisation, I only need one thread , is this the right way to do it (We need to use ThreadPool because is the one used in CaptureClient)?

kisenshi:
I have updated the code to integrate the changes related to Capture and CaptureClient

OrbitClientGgp/main.cpp
danielfenner: I think this is just fine as is; as you wrote you need the thread pool since it is in the signature of StartCapture.

OrbitClientGgp/ClientGgp.h
```diff
@@ -0,0 +1,47 @@
...
+#define ORBIT_CLIENT_GGP_CLIENT_GGP_H_
+
+#include <cstdint>
+#include <memory>
+#include <string>
+
+#include "ClientGgpOptions.h"
+#include "OrbitCaptureClient/CaptureClient.h"
+#include "OrbitCaptureClient/CaptureListener.h"
+#include "OrbitProcess.h"  // remove if including Capture.h in here
```
danielfenner, line 15: Is this comment outdated? At least I don't get it.

OrbitClientGgp/ClientGgp.h
```diff
@@ -0,0 +1,47 @@
...
+  void OnAddressInfo(
+      orbit_client_protos::LinuxAddressInfo address_info) override;
+
+ private:
+  ClientGgpOptions options_;
+  std::shared_ptr<grpc::Channel> grpc_channel_;
+  std::unique_ptr<CaptureClient> capture_client_;
+
+  void InitCapture();
+  std::shared_ptr<Process> GetOrbitProcessByPid(int32_t pid);
```
danielfenner, line 44: I think this can be moved into the implementation entirely (e.g. into a free function in the unnamed namespace). No need to hae it in the header.

OrbitClientGgp/ClientGgp.h
kisenshi: It was a note to myself when I was figuring out what it was needed to include and completely forgot to remove it, will update it, thanks!

OrbitClientGgp/ClientGgp.h

kisenshi:
Including OrbitClientGgp into master

PR #962, opened 2020-08-10
Small cleanup of OrbitCore

dpallotti:


OrbitGl/TextBox.h
```diff
@@ -5,6 +5,8 @@
 #ifndef ORBIT_GL_TEXT_BOX_H_
 #define ORBIT_GL_TEXT_BOX_H_
 
+#include <cfloat>
```
beckerhe, line 8: Maybe also `<cstdint>`?

OrbitCore/Utils.cpp
```diff
@@ -306,7 +287,7 @@ std::string OrbitUtils::FormatTime(const time_t& rawtime) {
   return std::string(buffer);
 }
 
-bool ReadProcessMemory(int32_t pid, uint64_t address, byte* buffer,
+bool ReadProcessMemory(int32_t pid, uint64_t address, uint8_t* buffer,
```
dimitry-, line 290: well.. this will conflict with my change :)

OrbitGl/TextBox.h
dpallotti: Done.

PR #964, opened 2020-08-10
GCC9 presubmit build

beckerhe:
This PR adds the repo-side of a GCC9-based presubmit build.

The second commit is temporary and will go aways before merging. So please only review the first one.

The script basically copies what has already been there for the clang-build. It is my intention to deduplicate these scripts in the near future, just not now in this PR.


kokoro/builds/gcc9_relwithdebinfo/build.sh
```diff
@@ -0,0 +1,42 @@
...
+
+  # Building Orbit
+  mkdir -p "${DIR}/build/"
+  cp -v "${DIR}/third_party/conan/lockfiles/linux/${CONAN_PROFILE}/conan.lock" \
+        "${DIR}/build/conan.lock"
+  sed -i -e "s|crashdump_server=|crashdump_server=$CRASHDUMP_SERVER|" \
+            "${DIR}/build/conan.lock"
+  conan install -u -pr ${CONAN_PROFILE} -if "${DIR}/build/" \
+          --build outdated \
+          -o crashdump_server="$CRASHDUMP_SERVER" \
```
IrinaShkviro, line 32: Do we need to define `CRASHDUMP_SERVER=""` and pass it here?

beckerhe: No, we don't need that, but I didn't wanted to change the code too much compared to the build-script from the `gcp_ubuntu`-job, since my intention is to eventually merge them back together into one build script. Does that make sense?

IrinaShkviro: Yes, sure. Thanks for the clarification

PR #966, opened 2020-08-10
Use this->GetCallStack instead of GSamplingProfiler->GetCallStack

florian-kuebler:
`GSamplingProfiler` is currently a global inside the capture class. Within the `SamplingProfiler` class, this global should not be used, but instead `this`, as it is a usual class.

### DIFF
--- a/OrbitCore/SamplingProfiler.cpp
+++ b/OrbitCore/SamplingProfiler.cpp
@@ -70,8 +70,7 @@ std::multimap<int, CallstackID> SamplingProfiler::GetCallstacksFromAddress(
 void SamplingProfiler::AddCallStack(CallstackEvent callstack_event) {
   CallstackID hash = callstack_event.callstack_hash();
   if (!HasCallStack(hash)) {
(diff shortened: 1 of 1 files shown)


dimitry-:
> `GSamplingProfiler` is currently a global inside the capture class. Within the `SamplingProfiler` class, this global should not be used, but instead `this`, as it is a usual class.

Actually question - is there only one instance of SamplingProfiler? I think we had something with multiple instances for selection tab - was it SamplingReport?

florian-kuebler:
> > `GSamplingProfiler` is currently a global inside the capture class. Within the `SamplingProfiler` class, this global should not be used, but instead `this`, as it is a usual class.
> 
> Actually question - is there only one instance of SamplingProfiler? I think we had something with multiple instances for selection tab - was it SamplingReport?

You are totally right. I wasn't aware of this right now. So we probably need the underlying global data first. Will have a look into this.

florian-kuebler:
As discussed offline. Larger refactoring is needed. So I'll close this one.

PR #970, opened 2020-08-11
Remove obsolete OutputTab and now unused LogInterface

dpallotti:


dpallotti:
@beckerhe The header check fails because of `orbitcodeeditor`, which has a Qt license header. Do you know why and what we should do with it? Only add ours or also remove the Qt license?

beckerhe:
> @beckerhe The header check fails because of `orbitcodeeditor`, which has a Qt license header. Do you know why and what we should do with it? Only add ours or also remove the Qt license?

No, since this is not our copyright we can't add our header to that file. All third-party-code needs to go into the third_party/-directory. That will also silence the presubmit-check.

Is it currently in use? Maybe we can remove it for now?

dpallotti:
> > @beckerhe The header check fails because of `orbitcodeeditor`, which has a Qt license header. Do you know why and what we should do with it? Only add ours or also remove the Qt license?
> 
> No, since this is not our copyright we can't add our header to that file. All third-party-code needs to go into the third_party/-directory. That will also silence the presubmit-check.
> 
> Is it currently in use? Maybe we can remove it for now?

It's used at least by the disassembly dialog. This might have been adapted from Qt code, but it looks like it's still mostly our code (I mean, it's called `OrbitCodeEditor`), that's why I'm confused about the copyright.

beckerhe:
> > > @beckerhe The header check fails because of `orbitcodeeditor`, which has a Qt license header. Do you know why and what we should do with it? Only add ours or also remove the Qt license?
> > 
> > 
> > No, since this is not our copyright we can't add our header to that file. All third-party-code needs to go into the third_party/-directory. That will also silence the presubmit-check.
> > Is it currently in use? Maybe we can remove it for now?
> 
> It's used at least by the disassembly dialog. This might have been adapted from Qt code, but it looks like it's still mostly our code (I mean, it's called `OrbitCodeEditor`), that's why I'm confused about the copyright.

Yeah, the name is `OrbitCodeEditor` but it's basically this example: https://doc.qt.io/qt-5/qtwidgets-widgets-codeeditor-example.html

dpallotti:
> > > > @beckerhe The header check fails because of `orbitcodeeditor`, which has a Qt license header. Do you know why and what we should do with it? Only add ours or also remove the Qt license?
> > > 
> > > 
> > > No, since this is not our copyright we can't add our header to that file. All third-party-code needs to go into the third_party/-directory. That will also silence the presubmit-check.
> > > Is it currently in use? Maybe we can remove it for now?
> > 
> > 
> > It's used at least by the disassembly dialog. This might have been adapted from Qt code, but it looks like it's still mostly our code (I mean, it's called `OrbitCodeEditor`), that's why I'm confused about the copyright.
> 
> Yeah, the name is `OrbitCodeEditor` but it's basically this example: https://doc.qt.io/qt-5/qtwidgets-widgets-codeeditor-example.html
> 
> You could also ignore the check and create a follow-up bug. We should have moved that into third_party already a long time ago. No need to force that on you now, just because of the new check.

Alright, so I created http://b/163462463 and I'll ignore the failed check for this pull request.

PR #973, opened 2020-08-11
Improve push_back in BlockChain

dimitry-:
It now can now move pushed values instead of always copying them


reichlfl:
Just FYI, there's an explicit test in `BlockChainTest.cpp`: `AddCopyableTypes` that tests if addition of copy-able types works correctly. Maybe consider adding a small additional test with a std::move usecase.

PR #976, opened 2020-08-11
clang-format: Increase ColumnLimit from 80 to 100

dimitry-:


beckerhe:
> As I said, by playing around with 100 and 120 for two minutes, if we have to change it I would prefer 100. 120 for C++ is really wide.

Since `[[nodiscard]] ` has 14 characters that would be enough to compensate longer lines introduced by nodiscard.

Also fine from my side.

reichlfl:
> As I said, by playing around with 100 and 120 for two minutes, if we have to change it I would prefer 100. 120 for C++ is really wide.

I'd also lean more towards 100 than 120. Makes it easier to view two files in split-mode next to each other.

kisenshi:
I would prefer 100 than 120 as well

dimitry-:
Since many of you prefer 100 over 120 - I will reduce it.

antonrohr:
I don't really care about the actual line width, but I used clang-format on the whole file every time I touched it.

We should be aware that after this change our codebase is in a non-formatted state. We have to keep in mind to only format the lines we touched from now on, or the PRs will be full of formatting related changes. 

I am planning to use a vscode extension called "Run on save" so I can run `git clang-format -f` instead of formatting the whole document.

PR #977, opened 2020-08-11
CallStack refactoring

IrinaShkviro:


OrbitCore/Callstack.h
```diff
@@ -12,15 +12,15 @@
 
 #include "CallstackTypes.h"
 
-struct CallStack {
-  CallStack() = default;
+struct HashedCallstack {
```
dpallotti, line 15: The "hash" field is just a "cache" for what `Hash()` returns and should probably be a private field as it's an implementation detail. As an implementation detail, it should not affect the name. Why not just `Callstack`?
(Note that in the past we have referred to "hashed callstacks" to mean callstacks that we already had and for which we could use the hash to refer to, so the opposite of this.)

OrbitCaptureClient/CaptureEventProcessor.cpp
```diff
@@ -189,10 +189,8 @@ void CaptureEventProcessor::ProcessAddressInfo(
 
 uint64_t CaptureEventProcessor::GetCallstackHashAndSendToListenerIfNecessary(
     const Callstack& callstack) {
-  CallStack cs;
-  for (uint64_t pc : callstack.pcs()) {
-    cs.m_Data.push_back(pc);
-  }
+  HashedCallstack cs;
+  cs.data = {callstack.pcs().begin(), callstack.pcs().end()};
```
dimitry-, line 193: maybe call the constructor explicitly here?

OrbitCore/Callstack.h
```diff
@@ -12,15 +12,15 @@
...
+  HashedCallstack() = default;
```
dimitry-, line 16: or have an explicit c-tor with vector<>

OrbitCore/Callstack.h
```diff
@@ -12,15 +12,15 @@
...
   inline CallstackID Hash() {
-    if (m_Hash != 0) return m_Hash;
-    m_Hash = XXH64(m_Data.data(), m_Data.size() * sizeof(uint64_t), 0xca1157ac);
-    return m_Hash;
+    if (hash != 0) return hash;
```
dimitry-, line 18: and calculate this hash in c-tor, which will let us make this method const.

(please ignore this if we are planning to get rid of this struct in the nearish future)

OrbitCaptureClient/CaptureEventProcessor.cpp
```diff
@@ -1,3 +1,7 @@
+// Copyright (c) 2020 The Orbit Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
```
beckerhe, line 4: Thanks. :innocent:

OrbitCore/Callstack.h
IrinaShkviro: We can use a different name if HashedCallstack is not good.
The problem here is that in proto files we have Callstacks which is actually a vector of addresses. Here we refer to Callstack as Callstack + Hash. I guess that using different name (not just `Callstack`) will make it a bit more clear that these types are not equal. But suggestions on better name are always welcome 🙂
And I totally agree that we shouln't use the `hash` field directly, so will make it private

dpallotti: Mmh, the feeling I have (this is a point of view I guess) is that the `Hash` method is more of a utility method (it could be a static function taking a callstack as argument) than part of the semantics of the class, and just `Callstack` would be fitting (in other words, the two classes you refer to are indeed basically equal, just that we want to have separate ones, as their scope is quite different and also one is a proto). So this now seems to be a namespace problem, where two different parts of Orbit use different classes with the same name. From this point of view, having the capital S before was a hack but it was convenient. I don't really have a suggestion honestly, other than introducing namespaces or leaving the old name.

Maybe someone else has any thoughts on the matter?

IrinaShkviro:
> otherwise lgtm

I reorganized the struct a bit, could you please take another look?

OrbitCore/Callstack.h
```diff
@@ -12,15 +12,20 @@
...
-struct CallStack {
-  CallStack() = default;
-  inline CallstackID Hash() {
-    if (m_Hash != 0) return m_Hash;
-    m_Hash = XXH64(m_Data.data(), m_Data.size() * sizeof(uint64_t), 0xca1157ac);
-    return m_Hash;
-  }
+struct HashedCallstack {
+  HashedCallstack() = default;
+  HashedCallstack(std::vector<uint64_t>&& addresses) {
```
dimitry-, line 17: explicit?

OrbitCore/Callstack.h
IrinaShkviro: I reorganized this struct, could you please take another look? Probably with new changes it makes more sense to leave the previous name (as all data are private now) and add a namespace, but for me it's still a bit weird that in one case Callstack doesn't have `hash` and in another it has.

OrbitCore/Callstack.h
IrinaShkviro: Yes, thanks!

IrinaShkviro:
Ok, as we didn't agree on name (`CallStack` vs `HashedCallstack`), I leave only refactoring changes in this PR which are beyond doubt .

OrbitCore/Callstack.h
```diff
@@ -7,20 +7,24 @@
...
-  }
+  explicit CallStack(std::vector<uint64_t>&& addresses) {
+    data = std::move(addresses);
+    hash = XXH64(data.data(), data.size() * sizeof(uint64_t), 0xca1157ac);
+  };
 
-  CallstackID m_Hash = 0;
-  std::vector<uint64_t> m_Data;
+  CallstackID GetHash() const { return hash; }
+  uint64_t GetData(size_t id) const { return data.at(id); }
```
dpallotti, line 22: I would be more specific with naming here: `GetData` could be `GetFrame` or `GetIp` or `GetPc` and `id` could be `index`.

OrbitCore/Callstack.h
```diff
@@ -7,20 +7,24 @@
...
+  const std::vector<uint64_t>& GetData() const { return data; };
+  size_t GetDataSize() const { return data.size(); }
+
+ private:
+  CallstackID hash = 0;
+  std::vector<uint64_t> data;
```
dpallotti, line 28: `frames_`? `pcs_`? `ips_`? Use the underscore at the end.

OrbitCore/Callstack.h
```diff
@@ -7,20 +7,24 @@
 
 #include <xxhash.h>
 
-#include <string>
 #include <vector>
 
 #include "CallstackTypes.h"
 
 struct CallStack {
```
dpallotti, line 14: Make it `class` as it has private members and add `public:` for the public members.

OrbitCore/Callstack.h
```diff
@@ -7,20 +7,24 @@
...
 
-  CallstackID m_Hash = 0;
-  std::vector<uint64_t> m_Data;
+  CallstackID GetHash() const { return hash; }
+  uint64_t GetData(size_t id) const { return data.at(id); }
+  const std::vector<uint64_t>& GetData() const { return data; };
+  size_t GetDataSize() const { return data.size(); }
+
+ private:
+  CallstackID hash = 0;
```
dpallotti, line 27: `hash_`?

OrbitCore/Callstack.h
```diff
@@ -7,20 +7,24 @@
...
+    data = std::move(addresses);
+    hash = XXH64(data.data(), data.size() * sizeof(uint64_t), 0xca1157ac);
+  };
 
-  CallstackID m_Hash = 0;
-  std::vector<uint64_t> m_Data;
+  CallstackID GetHash() const { return hash; }
+  uint64_t GetData(size_t id) const { return data.at(id); }
+  const std::vector<uint64_t>& GetData() const { return data; };
+  size_t GetDataSize() const { return data.size(); }
```
dpallotti, line 24: Just `size`? Or `GetFrameCount`/`GetIpCount`/`GetPcCount`?

OrbitCore/Callstack.h
```diff
@@ -7,20 +7,24 @@
...
 struct CallStack {
   CallStack() = default;
-  inline CallstackID Hash() {
-    if (m_Hash != 0) return m_Hash;
-    m_Hash = XXH64(m_Data.data(), m_Data.size() * sizeof(uint64_t), 0xca1157ac);
-    return m_Hash;
-  }
+  explicit CallStack(std::vector<uint64_t>&& addresses) {
+    data = std::move(addresses);
+    hash = XXH64(data.data(), data.size() * sizeof(uint64_t), 0xca1157ac);
```
dpallotti, line 18: Would it make sense to initialize it lazily the first time `GetHash` is called?

OrbitCore/Callstack.h
```diff
@@ -7,20 +7,24 @@
...
+  };
 
-  CallstackID m_Hash = 0;
-  std::vector<uint64_t> m_Data;
+  CallstackID GetHash() const { return hash; }
+  uint64_t GetData(size_t id) const { return data.at(id); }
+  const std::vector<uint64_t>& GetData() const { return data; };
```
dpallotti, line 23: `frames`? `ips`? `pcs`?

OrbitCore/Callstack.h
dimitry-: this makes GetHash non-const and also doesn't save us much or does it?

dpallotti: `hash_` would be `mutable` as it's a cached value. Or it's fine as is.

IrinaShkviro: I don't think that we use lazy initialization very often, and I would prefer to initialize hash in the constructor then

OrbitCore/Callstack.h
IrinaShkviro: If it's struct we shouldn't use underscore at the end

dpallotti: That's because structs typically only have public members. These are private members, hence why this should be a class and why these should have the underscore.

OrbitCore/Callstack.h
```diff
@@ -7,20 +7,26 @@
...
+    frames_ = std::move(addresses);
+    hash_ =
+        XXH64(frames_.data(), frames_.size() * sizeof(uint64_t), 0xca1157ac);
+  };
 
-  CallstackID m_Hash = 0;
-  std::vector<uint64_t> m_Data;
+  CallstackID GetHash() const { return hash_; }
+  uint64_t GetFrame(size_t index) const { return frames_.at(index); }
+  const std::vector<uint64_t>& GetFrames() const { return frames_; };
```
dpallotti, line 25: If you want this could be called `frames()` as it's a trivial getter (same for `GetHash`), but I don't have a strong opinion.

IrinaShkviro: I would prefer to have a `GetFrames` here, because otherwise we will have 2 functions with one style and 2 with another. I've seen that we use camel for trivial getters too, so it should be fine to keep this style here.

PR #980, opened 2020-08-11
Move state to CaptureClient

dimitry-:
This change moves capture state from a global
to CaptureClient and syncronizes access to it
it is now thread-safe and consistent.

This also introduces kStarting state which indicates
that start is initiated but is not yet in the correct
state for stopping.  StopCapture waits until the state
becomes something other that kStarting before either
initiating a stop or in the case Start has failed -
ignoring user request.

Bug: http://b/161968742

### DIFF
--- a/OrbitCaptureClient/CaptureClient.cpp
+++ b/OrbitCaptureClient/CaptureClient.cpp
@@ -25,6 +25,23 @@ IntrumentedFunctionTypeFromOrbitType(FunctionInfo::OrbitType orbit_type) {
   }
 }
 
+ErrorMessageOr<void> CaptureClient::StartCapture(
+    ThreadPool* thread_pool, int32_t pid,
+    const std::map<uint64_t, FunctionInfo*>& selected_functions) {
+  absl::MutexLock lock(&state_mutex_);
+  if (state_ != State::kStopped) {
+    return ErrorMessage(
+        "Capture cannot be started, the previous capture is still "
+        "running/stopping.");
+  }
+
+  state_ = State::kStarting;
+  thread_pool->Schedule(
+      [this, pid, selected_functions]() { Capture(pid, selected_functions); });
+
+  return outcome::success();
+}
+
 void CaptureClient::Capture(
     int32_t pid, const std::map<uint64_t, FunctionInfo*>& selected_functions) {
   CHECK(reader_writer_ == nullptr);
@@ -72,10 +89,14 @@ void CaptureClient::Capture(
   LOG("Sent CaptureRequest on Capture's gRPC stream: asking to start "
       "capturing");
 
+  state_mutex_.Lock();
+  state_ = State::kStarted;
+  state_mutex_.Unlock();
+
   capture_listener_->OnCaptureStarted();
 
   CaptureResponse response;
-  while (reader_writer_->Read(&response)) {
+  while (!force_stop_ && reader_writer_->Read(&response)) {
     event_processor_->ProcessEvents(response.capture_events());
   }
   LOG("Finished reading from Capture's gRPC stream: all capture data has been "
@@ -85,14 +106,38 @@ void CaptureClient::Capture(
   FinishCapture();
 }
 
-void CaptureClient::StopCapture() {
+bool CaptureClient::StopCapture() {
+  absl::MutexLock lock(&state_mutex_);
+  if (state_ == State::kStarting) {
+    // Block and wait until the state is not kStarting
+    bool (*IsNotStarting)(State * state) = [](State* state) {
+      return *state != State::kStarting;
+    };
+    state_mutex_.Await(absl::Condition(IsNotStarting, &state_));
+  }
+
+  if (state_ != State::kStarted) {
+    LOG("StopCapture ignored, because it is already stopping or stopped");
+    return false;
+  }
+
   CHECK(reader_writer_ != nullptr);
 
   if (!reader_writer_->WritesDone()) {
-    ERROR("Finishing writing on Capture's gRPC stream");
-    FinishCapture();
+    // Normally the capture thread waits until service stops sending messages,
+    // but in this case since we failed to notify the service we pull emergency
+    // stop plug. Setting this flag forces capture thread to exit as soon
+    // as it notices that it was set.
+    ERROR(
+        "WritesDone on Capture's gRPC stream failed: unable to finish the "
+        "capture in orderly manner, initiating emergency stop");
+    force_stop_ = true;
+  } else {
+    LOG("Finished writing on Capture's gRPC stream: asking to stop capturing");
   }
-  LOG("Finished writing on Capture's gRPC stream: asking to stop capturing");
+
+  state_ = State::kStopping;
+  return true;
 }
 
 void CaptureClient::FinishCapture() {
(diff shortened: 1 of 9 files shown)


OrbitCaptureClient/CaptureClient.cpp
```diff
@@ -71,14 +91,35 @@ void CaptureClient::Capture(
...
-    ERROR("Finishing writing on Capture's gRPC stream");
-    FinishCapture();
+    // It looks like there is no way to recover from that
+    // we can't safely reset reade_writer_ here because it
+    // will likely lead to null pointer dereference on the
+    // capture thread on reader_writer_->Read(). And no good
+    // way to stop the capture thread as well, since we do not
+    // know if this is going to be blocked on the same call
+    // forever. This is the best we can do here:
+    FATAL("WritesDone failed unable to finish the capture in orderly manner");
```
florian-kuebler, line 118: This will lead to Orbit crashes, right? Shouldn't it be handled by the UI in some way?

dimitry-: well I do not see how we can end-up in consistent state. I tried to explain it in the comment above.

florian-kuebler: Okay. Have you considered using Check then? However, either is fine for me.

dimitry-: I think this error message is more expressive than Check failed, and CHECK is for something that shouldn't happen if it happens this is programming error.

In this case the situation is a bit different. This can happen, we just cannot recover from it.

florian-kuebler: Okay good point (even though I though CHECK can also have a message, like assert would), so yes CHECK does not make sense. 
If this can occur, I would still at least inform the user that something bad happend and that Orbit should be restarted.

OrbitGl/App.cpp
```diff
@@ -571,22 +567,23 @@ bool OrbitApp::StartCapture() {
   int32_t pid = Capture::GTargetProcess->GetID();
   std::map<uint64_t, FunctionInfo*> selected_functions =
       Capture::GSelectedFunctionsMap;
-  thread_pool_->Schedule([this, pid, selected_functions] {
-    capture_client_->Capture(pid, selected_functions);
-  });
+
+  result = capture_client_->StartCapture(thread_pool_.get(), pid,
```
kisenshi, line 571: You are getting the result but not checking if it contains error message afterwards

OrbitGl/App.cpp
```diff
@@ -965,3 +961,6 @@ void OrbitApp::CrashOrbitService(
         [crash_type, this] { crash_manager_->CrashOrbitService(crash_type); });
   }
 }
+bool OrbitApp::IsCapturing() {
+  return capture_client_->state() != CaptureClient::State::kStopped;
```
kisenshi, line 965: I think this method could be moved inside CaptureClient so it can be called directly instead, so it can be used in multiple clients

OrbitGl/App.cpp
```diff
@@ -571,22 +567,23 @@ bool OrbitApp::StartCapture() {
...
   return true;
 }
 
 void OrbitApp::StopCapture() {
-  CHECK(Capture::GState == Capture::State::kStarted);
-  Capture::StopCapture();
-
-  capture_client_->StopCapture();
+  if (!capture_client_->StopCapture()) {
+    LOG("StopCapture failed, likely because it is already stopping/stopped");
```
kisenshi, line 579: Maybe this LOG (or ERR maybe) can be done inside CaptureClient::StopCapture so it logs why the error has occurred

OrbitGl/App.cpp
dimitry-: indeed - good catch, thanks!

OrbitCaptureClient/CaptureClient.cpp
dimitry-: I see your point. I will add todo for now, This will probably require another callback. OnAborted/Fatal to let client display a message before we abort.

kisenshi:
Thanks for making this change. I will need to update my client to remove the creation of the thread on it but I think it makes a lot of sense to have it in OrbitClient instead.

I will wait to merge my PR and update it accordingly beforehand; please let me know when you merge this

OrbitCaptureClient/CaptureClient.cpp
florian-kuebler: Okay cool. Thanks again!

OrbitGl/App.cpp
dimitry-: done.

OrbitGl/App.cpp
dimitry-: done.

OrbitGl/App.cpp
dimitry-: done.

OrbitCaptureClient/CaptureClient.cpp
```diff
@@ -85,14 +106,40 @@ void CaptureClient::Capture(
...
+    return false;
+  }
+
   CHECK(reader_writer_ != nullptr);
 
   if (!reader_writer_->WritesDone()) {
-    ERROR("Finishing writing on Capture's gRPC stream");
-    FinishCapture();
+    // It looks like there is no way to recover from that
+    // we can't safely reset reade_writer_ here because it
```
dpallotti, line 128:
```suggestion
    // It looks like there is no way to recover from this.
    // We can't safely reset reade_writer_ here because it
```

OrbitCaptureClient/CaptureClient.cpp
```diff
@@ -85,14 +106,40 @@ void CaptureClient::Capture(
...
+    // will likely lead to null pointer dereference on the
+    // capture thread on reader_writer_->Read(). And no good
```
dpallotti, line 130: Isn't there a way to protect from this by somehow checking if `reader_writer_` has become a null pointer? This is a regression and we should really avoid crashing on purpose.

dimitry-: why is it a regression?

dpallotti: So you mean that the null pointer dereferencing already happened before? Either way, is there really no way we can protect against that instead of just crashing?

dimitry-: Added forced stop flag...

OrbitCaptureClient/CaptureClient.cpp
```diff
@@ -85,14 +106,38 @@ void CaptureClient::Capture(
...
 
   if (!reader_writer_->WritesDone()) {
-    ERROR("Finishing writing on Capture's gRPC stream");
-    FinishCapture();
+    // Normally the capture thread waits until service stops sending messages,
+    // but in this case since we failed to notify the service we pull emergency
+    // stop plug. Setting this flag forces capture thread to exit as soon
+    // as it notices that it was set.
+    ERROR(
+        "WritesDone failed unable to finish the capture in orderly manner, "
```
dpallotti, line 132:
```suggestion
        "WritesDone on Capture's gRPC stream failed: unable to finish the capture in orderly manner, "
```
(with the proper clang-format)

dimitry-: done.

dimitry-:
@dpallotti @kisenshi ptal, is there anything else to address?

PR #981, opened 2020-08-11
GlPanel is destroyed correctly on destruction of OrbitGLWidget

reichlfl:
Small fix to have GlPanel consistently destroyed when the OrbitGLWidget is destroyed. This *probably* doesn't happen during runtime, but won't hurt nevertheless.

### DIFF
--- a/OrbitGl/GlPanel.cpp
+++ b/OrbitGl/GlPanel.cpp
@@ -7,14 +7,13 @@
 #include "CaptureWindow.h"
 
 std::unique_ptr<GlPanel> GlPanel::Create(Type a_Type) {
-  std::unique_ptr<GlPanel> panel = nullptr;
-
-  switch (a_Type) {
-    case CAPTURE:
-      panel = std::make_unique<CaptureWindow>();
-      break;
+  // Intended to be a factory method creating different types of
+  // GlPanels, but right now there's only CAPTURE
+  if (a_Type != CAPTURE) {
+    return nullptr;
   }
 
+  std::unique_ptr<GlPanel> panel = std::make_unique<CaptureWindow>();
   panel->m_Type = a_Type;
 
   return panel;
--- a/OrbitGl/GlPanel.cpp
+++ b/OrbitGl/GlPanel.cpp
@@ -21,6 +21,8 @@ GlPanel* GlPanel::Create(Type a_Type) {
   return panel;
 }
 
+void GlPanel::Destroy(GlPanel* panel) { delete panel; }
+
 GlPanel::GlPanel() {
   m_WindowOffset[0] = 0;
   m_WindowOffset[1] = 0;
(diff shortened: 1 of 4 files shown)


OrbitQt/orbitglwidget.cpp
```diff
@@ -41,16 +41,15 @@ bool OrbitGLWidget::eventFilter(QObject* /*object*/, QEvent* event) {
 
 void OrbitGLWidget::Initialize(GlPanel::Type a_Type,
                                OrbitMainWindow* a_MainWindow) {
-  UNUSED(a_Type);
-  UNUSED(a_MainWindow);
-
   m_OrbitPanel = GlPanel::Create(a_Type);
```
dimitry-, line 44: Should this just be unique_ptr? and modify GlPanel::Create to return unique_ptr as well?

reichlfl: That was also my first change, but this caused Orbit to crash... I didn't spend a lot of time looking into it though, if you want I can have another look at this.

dimitry-: It is strange.. It would be good to know why Orbit crashes in case of unique_ptr, I do not see anything which could imply a different scope.

reichlfl: Ok, so the short answer is: Because I did it wrong :-) I changed the code to return a unique_ptr as you suggested, this is actually cleaner and requires a lot less code... thanks for pushing for it!

OrbitGl/GlPanel.cpp
```diff
@@ -6,23 +6,20 @@
 
 #include "CaptureWindow.h"
 
-GlPanel* GlPanel::Create(Type a_Type) {
-  GlPanel* panel = nullptr;
+std::unique_ptr<GlPanel> GlPanel::Create(Type a_Type) {
+  std::unique_ptr<GlPanel> panel = nullptr;
```
dimitry-, line 10: if (a_Type != CAPTURE) return nullptr;

and then you can just assume this is capture - construct and return it - switch is no longer needed.

OrbitGl/GlPanel.cpp
```diff
@@ -6,18 +6,17 @@
...
   switch (a_Type) {
     case CAPTURE:
-      panel = new CaptureWindow();
+      panel = std::make_unique<CaptureWindow>();
       break;
   }
 
   panel->m_Type = a_Type;
 
-  // Todo: fix leak...
```
beckerhe, line 20: :smile:

OrbitGl/GlPanel.cpp
reichlfl: I think the "switch" made it more clear on what the intention of this method initially was if more types of panels would be added. But when looking at your comment, I realized that this will crash badly if anything else but CAPTURE would be passed here in the future, so I changed it as recommended and added a comment to clarify the intent of this.

PR #983, opened 2020-08-11
Added .clang-tidy file

antonrohr:
This is my attempt on producing a clang-tidy file that is helpful for our codebase. I tried to enable as many useful checks as possible without getting too many warning messages for our codebase. 

The naming rules do not work for class/struct member naming since clang-tidy currently does not differentiate between class and struct, but the google style guide does (class: `lower_case_` struct: `lower_case`). Wrong naming of functions, local variables, constants, classes, struct and parameters should get caught with this file. 

This is best used with an IDE, but can also be triggered via the command line (`clang-tidy <filename>`)

* CLion defaults to using its own clang-tidy checks, using this file needs to be enabled in settings.
* VS2019 supports clang-tidy files (needs to be enabled in settings) (VS2017 no support)
* clangd extension for vscode supports clang-tidy files out of the box

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -1,0 +1,61 @@
+# Copyright (c) 2020 The Orbit Authors. All rights reserved.
+# Use of this source code is governed by a BSD-style license that can be
+# found in the LICENSE file.
+---
+Checks: >
+  -*,
+  abseil-*,
+  bugprone-*,
+  cert-*,
+  -cert-err58-cpp,
+  cppcoreguidelines-*,
+  -cppcoreguidelines-special-member-functions,
+  -cppcoreguidelines-avoid-magic-numbers,
+  -cppcoreguidelines-pro-type-vararg,
+  -cppcoreguidelines-non-private-member-variables-in-classes,
+  -cppcoreguidelines-pro-bounds-pointer-arithmetic,
+  -cppcoreguidelines-avoid-non-const-global-variables,
+  misc-*,
+  -misc-non-private-member-variables-in-classes,
+  google-*,
+  -google-default-arguments,
+  -google-runtime-references,
+  modernize-*,
+  -modernize-use-noexcept,
+  -modernize-use-trailing-return-type,
+  performance-*,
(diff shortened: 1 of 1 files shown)


beckerhe:
Hey, thanks a lot for that.

What about the `cppcoreguidelines-*`-checks? Except one or two they seems pretty valueable, especially the casting stuff, or member initialization and so on. There is also a rule about owning memory!

antonrohr:
I added the `cppcoreguidelines-*` checks and removed some that I think we shouldn't use at this point since they generate a quite a few warnings in code that I personally would consider fine. I would rather have smaller set of rules right now with less but more actionable warnings, and then decide on a case by case basis in the future to add more checks

.clang-tidy
```diff
@@ -0,0 +1,62 @@
...
+  abseil-*,
+  bugprone-*,
+  cert-*,
+  -cert-err58-cpp,
+  cppcoreguidelines-*,
+  -cppcoreguidelines-special-member-functions,
+  -cppcoreguidelines-avoid-magic-numbers,
+  -cppcoreguidelines-pro-type-vararg,
+  -cppcoreguidelines-non-private-member-variables-in-classes,
+  -cppcoreguidelines-pro-type-reinterpret-cast,
```
beckerhe, line 16: I would remove this. `reinterpret_cast` should really be warned about, even though we use it in some places. We can check theses cases and add a `// NOLINT` comment if it's really needed.

antonrohr: I am unsure if you marked these 4 lines explicitly, or if you are just talking about the reinterpret_cast. (For the later I am okay with it and activated that check. 

For the other ones I am of the opinion that we should not activate them yet, or at all.
* "Magic numbers" are often used in tests as test data.
* The vararg check is triggered by the LOG and ERROR macros, which displays a warning everytime we use them (at least we should change those macros first)
* non-private member check: do we actually want to do that?

beckerhe: Was totally by accident. I was only talking about reinterpret_cast.

PR #984, opened 2020-08-11
Fix 100% crash when capturing instrumented functions.

pierricgimmig:
The code assumed that the hashmap contained the thread name. Use operator[] instead of at() to insert and return an empty string if the thread name is not (yet) present.

### DIFF
--- a/OrbitCore/CaptureData.h
+++ b/OrbitCore/CaptureData.h
@@ -65,9 +65,10 @@ class CaptureData {
   }
 
   [[nodiscard]] const std::string& GetThreadName(int32_t thread_id) const {
-    static std::string empty_string;
-    return thread_names_.contains(thread_id) ? thread_names_.at(thread_id)
-                                             : empty_string;
+    static const std::string kEmptyString;
+    auto it = thread_names_.find(thread_id);
+    return it != thread_names_.end() ? thread_names_.at(thread_id)
+                                     : kEmptyString;
   }
 
   void set_thread_names(
--- a/OrbitCore/CaptureData.h
+++ b/OrbitCore/CaptureData.h
@@ -64,8 +64,10 @@ class CaptureData {
     return thread_names_;
   }
 
-  [[nodiscard]] const std::string& GetThreadName(int32_t thread_id) {
-    return thread_names_[thread_id];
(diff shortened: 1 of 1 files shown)


OrbitCore/CaptureData.h
```diff
@@ -64,8 +64,8 @@ class CaptureData {
     return thread_names_;
   }
 
-  [[nodiscard]] const std::string& GetThreadName(int32_t thread_id) const {
-    return thread_names_.at(thread_id);
+  [[nodiscard]] const std::string& GetThreadName(int32_t thread_id) {
+    return thread_names_[thread_id];
```
dimitry-, line 68: You can keep it const by doing find() -> return ""; in case it was not found. No need to modify the map.

pierricgimmig: Then we'd have to change the return type. I guess using a static empty string would do the trick.

OrbitCore/CaptureData.h
```diff
@@ -64,8 +64,10 @@ class CaptureData {
     return thread_names_;
   }
 
-  [[nodiscard]] const std::string& GetThreadName(int32_t thread_id) {
-    return thread_names_[thread_id];
+  [[nodiscard]] const std::string& GetThreadName(int32_t thread_id) const {
+    static std::string empty_string;
```
dimitry-, line 68: kEmptyString, also const

OrbitCore/CaptureData.h
```diff
@@ -64,8 +64,10 @@ class CaptureData {
...
+    return thread_names_.contains(thread_id) ? thread_names_.at(thread_id)
```
dimitry-, line 69: doesn't this do 2 lookups instead of one if you use find?

OrbitCore/CaptureData.h
```diff
@@ -65,7 +65,10 @@ class CaptureData {
   }
 
   [[nodiscard]] const std::string& GetThreadName(int32_t thread_id) const {
-    return thread_names_.at(thread_id);
+    static const std::string kEmptyString;
+    auto it = thread_names_.find(thread_id);
+    return it != thread_names_.end() ? thread_names_.at(thread_id)
```
dimitry-, line 70: still 2 lookups. use it->second instead of at() call

PR #985, opened 2020-08-11
Remove log on scheduling events.

pierricgimmig:
There was a bit too much logging action :-)


dimitry-:
#982 addresses this as well.

pierricgimmig:
Addressed in https://github.com/google/orbit/pull/982/files.

PR #990, opened 2020-08-12
Moved FunctionStats from FunctionInfo to CaptureData. 

florian-kuebler:
This will break compatibility to old captures, but the decoupling of those helps to reduce code complexity (e.g. on cleaning). It will also allow a history of captures at some point.

### DIFF
--- a/OrbitClientProtos/capture_data.proto
+++ b/OrbitClientProtos/capture_data.proto
@@ -15,6 +15,7 @@ message FunctionStats {
 }
 
 message FunctionInfo {
+  reserved 11;
   string name = 1;
   string pretty_name = 2;
   string loaded_module_path = 3;
@@ -41,7 +42,6 @@ message FunctionInfo {
     kOrbitTrackDoubleAsInt64 = 12;
   }
   OrbitType type = 10;
-  FunctionStats stats = 11;
 }
 
 message CallstackEvent {
@@ -74,6 +74,8 @@ message CaptureInfo {
   repeated CallstackInfo callstacks = 6;
   repeated CallstackEvent callstack_events = 7;
   map<uint64, string> key_to_string = 8;
+  // TODO(kuebler): consider removing this field completely.
+  map<uint64, FunctionStats> function_stats = 9;
 }
 
 message TimerInfo {
--- a/OrbitCore/CaptureData.cpp
+++ b/OrbitCore/CaptureData.cpp
@@ -4,6 +4,10 @@
 
 #include "CaptureData.h"
 
+#include "Profiling.h"
+
+using orbit_client_protos::FunctionStats;
+
 orbit_client_protos::LinuxAddressInfo* CaptureData::GetAddressInfo(
     uint64_t address) {
   auto address_info_it = address_infos_.find(address);
@@ -11,4 +15,34 @@ orbit_client_protos::LinuxAddressInfo* CaptureData::GetAddressInfo(
     return nullptr;
   }
   return &address_info_it->second;
+}
+
+const FunctionStats& CaptureData::GetFunctionStatsOrDefault(
+    uint64_t function_address) {
+  static const FunctionStats kDefaultFunctionStats;
+  auto function_stats_it = functions_stats_.find(function_address);
+  if (function_stats_it == functions_stats_.end()) {
+    return kDefaultFunctionStats;
+  }
+  return function_stats_it->second;
+}
+
+void CaptureData::UpdateFunctionStats(
+    orbit_client_protos::FunctionInfo* func,
+    const orbit_client_protos::TimerInfo& timer_info) {
+  const uint64_t function_address = func->address();
+  FunctionStats& stats = functions_stats_[function_address];
+  stats.set_count(stats.count() + 1);
+  uint64_t elapsed_nanos =
+      TicksToNanoseconds(timer_info.start(), timer_info.end());
+  stats.set_total_time_ns(stats.total_time_ns() + elapsed_nanos);
+  stats.set_average_time_ns(stats.total_time_ns() / stats.count());
+
+  if (elapsed_nanos > stats.max_ns()) {
+    stats.set_max_ns(elapsed_nanos);
+  }
+
+  if (stats.min_ns() == 0 || elapsed_nanos < stats.min_ns()) {
+    stats.set_min_ns(elapsed_nanos);
+  }
 }
--- a/OrbitCore/CaptureData.h
+++ b/OrbitCore/CaptureData.h
@@ -20,6 +20,16 @@ class CaptureData {
       : process_id_{process_id},
         process_name_{std::move(process_name)},
         selected_functions_{std::move(selected_functions)} {}
+  explicit CaptureData(
+      int32_t process_id, std::string process_name,
+      std::vector<std::shared_ptr<orbit_client_protos::FunctionInfo>>
+          selected_functions,
+      absl::flat_hash_map<uint64_t, orbit_client_protos::FunctionStats>
+          functions_stats)
+      : process_id_{process_id},
+        process_name_{std::move(process_name)},
+        selected_functions_{std::move(selected_functions)},
+        functions_stats_{std::move(functions_stats)} {}
 
   explicit CaptureData() = default;
   CaptureData(const CaptureData& other) = default;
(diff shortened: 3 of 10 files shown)


OrbitCore/CaptureData.h
```diff
@@ -11,6 +11,8 @@
 #include "absl/container/flat_hash_map.h"
 #include "capture_data.pb.h"
 
+using orbit_client_protos::FunctionStats;
```
IrinaShkviro, line 14: Do we want to use `using` in header files or should we use the fully qualified name?

IrinaShkviro:
I'm looping in Anton as he has worked on Data Model recently.

OrbitCore/CaptureData.h
florian-kuebler: I'm fine with both.

OrbitClientProtos/capture_data.proto
```diff
@@ -74,6 +73,7 @@ message CaptureInfo {
   repeated CallstackInfo callstacks = 6;
   repeated CallstackEvent callstack_events = 7;
   map<uint64, string> key_to_string = 8;
+  map<uint64, FunctionStats> function_stats = 9;
```
IrinaShkviro, line 76: Is this map enough to reconstruct `repeated FunctionInfo selected_functions = 1;` and don't save this field field anymore?

florian-kuebler: Not 100% sure if address_infos contains the information also for the selected functions. Will have a look into this.

OrbitCore/CaptureData.h
```diff
@@ -79,6 +81,11 @@ class CaptureData {
     thread_names_.insert_or_assign(thread_id, std::move(thread_name));
   }
 
+  const FunctionStats& GetOrInsertFunctionStats(uint64_t function_address);
```
dimitry-, line 84: Maybe this needs to be converted to std::optional and the absence of value handled on the caller side?

OrbitCore/CaptureData.cpp
```diff
@@ -12,3 +16,28 @@ orbit_client_protos::LinuxAddressInfo* CaptureData::GetAddressInfo(
...
+const FunctionStats& CaptureData::GetOrInsertFunctionStats(
+    uint64_t function_address) {
+  return functions_stats_[function_address];
+}
+
+void CaptureData::UpdateFunctionStats(
+    orbit_client_protos::FunctionInfo* func,
+    const orbit_client_protos::TimerInfo& timer_info) {
+  const uint64_t function_address = func->address();
+  FunctionStats& stats = functions_stats_[function_address];
```
dimitry-, line 29: It is always better to create a temporary object - compute it and as final stem insert to the map. The reason is that if something goes wrong we do not end up with half baked instance in the map.

OrbitClientProtos/capture_data.proto
florian-kuebler: Seem to be strange. There is PreSave in Capture, which updates addresses of selected functions in the SamplingProfiler. However, address_infos comes directly from App. I suggest to look further into this in a separate CL and just add comment for now.

IrinaShkviro: It's more for future: even if it's so it should/could be in the separate PR

OrbitCore/CaptureData.cpp
dimitry-: actually - this is not relevant in this case since it looks like this is updating existing record.

florian-kuebler:
Thanks for the Review so far. I noticed that I forgot to add the handling of the new map in (de-)serialization and added this. Please have also a look at those changes.

OrbitClientProtos/capture_data.proto
```diff
@@ -41,7 +41,6 @@ message FunctionInfo {
     kOrbitTrackDoubleAsInt64 = 12;
   }
   OrbitType type = 10;
-  FunctionStats stats = 11;
```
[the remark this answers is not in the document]
IrinaShkviro: But do we need it as we will check capture version in header and won't load old captures anyway?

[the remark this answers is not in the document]
dimitry-: well, we still want to do that - it is low cost and good practice.

florian-kuebler:
> Please also fix the commit message it should be a title limited to 50 characters \n\n and then a description.

Thanks. I updated the commit message.

OrbitCore/CaptureData.cpp
florian-kuebler: Right, its updating the existing value or if this is the first update, create empty stats and then update it.

OrbitCore/CaptureData.h
florian-kuebler: Not 100%. Thought a lot of this case. This will be called by the UI to query function stats e.g. to display in the live view. If the hooked function was not called in the capture there will be no call update the stats, and thus no entry for this function. 
I see three alternatives to what I've done:
1. Use std::optional and handle the empty case in the client.
2. Change the method to GetOrDefaultFunctionStats and return an empty FunctionStats. This would require to actually return a copy rather then a const ref.
3. In constructor, create empty entries for all selected functions and make this function fail if function is not there (e.g. by using `at`).

OrbitClientProtos/capture_data.proto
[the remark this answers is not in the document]
florian-kuebler: I added the reserve 11;

OrbitClientProtos/capture_data.proto
florian-kuebler: As I added a todo above, I'll resolve this for now.

OrbitCore/CaptureData.h
dimitry-: it feels not right using using in header files tbh

OrbitClientProtos/capture_data.proto
```diff
@@ -41,7 +41,7 @@ message FunctionInfo {
...
+  reserved 11;
```
dimitry-, line 44: nit: put it on the top - the first line of the message.

OrbitCore/CaptureData.h
dimitry-: in case of 2 you do not have to return copy if you create a static const FunctionStats

florian-kuebler: Right. So which do you prefer?

OrbitCore/CaptureData.h
florian-kuebler: Okay, I remved the using from the header.

OrbitCore/CaptureData.h
florian-kuebler: As discussed, changed it to 2.

OrbitClientProtos/capture_data.proto
florian-kuebler: Thanks, done!

OrbitClientProtos/capture_data.proto
```diff
@@ -75,6 +74,7 @@ message CaptureInfo {
   repeated CallstackInfo callstacks = 6;
   repeated CallstackEvent callstack_events = 7;
   map<uint64, string> key_to_string = 8;
+  // TODO(kuebler): consider removing this field completely.
```
IrinaShkviro, line 77: Looks like the comment changed place accidentally

florian-kuebler: Actually it was on purpose. As discussed with Dimitry offline, we can probably get rid of function stats for serialization completely (as this is aggregated data from timers, which are already part of the capture info proto). But in this case, we need the selected functions somehow.

OrbitGl/CaptureSerializer.cpp
```diff
@@ -78,6 +79,11 @@ void CaptureSerializer::FillCaptureData(CaptureInfo* capture_info) {
     capture_info->add_address_infos()->CopyFrom(address_info.second);
   }
 
+  const absl::flat_hash_map<uint64_t, FunctionStats> functions_stats =
```
antonrohr, line 82: maybe const reference here?

florian-kuebler: You are totally right. I missed the ref. Changed it accordingly. Thanks.

OrbitClientProtos/capture_data.proto
IrinaShkviro: Ah, ok, I didn't know about that. Thanks for explanation!

PR #992, opened 2020-08-12
Modernize BlockChain container

dimitry-:
This commit modernizes BlockChain:
1. Removed unused methods
2. Converted structs to classes and made fields private
3. Removed reference from Block to BlockChain

Test: run OrbitCoreTests

### DIFF
--- a/OrbitCore/BlockChain.h
+++ b/OrbitCore/BlockChain.h
@@ -2,7 +2,8 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
 
-#pragma once
+#ifndef ORBIT_CORE_BLOCK_CHAIN_H_
+#define ORBIT_CORE_BLOCK_CHAIN_H_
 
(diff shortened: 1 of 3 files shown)


OrbitCore/BlockChain.h
```diff
@@ -2,229 +2,244 @@
...
+
+    return nullptr;
+  }
+
+  Block<T, BlockSize>* root_;
+  Block<T, BlockSize>* current_;
+  uint32_t size_;
 };
+
+#endif  // ORBIT_CORE_BLOCK_CHAIN_H_
```
reichlfl, line 245: Nit: Newline? Don't know if we have a "rule" for this, but Github shows it quite explicitly.

dimitry-: I trust clang-format on this :)

reichlfl: Fair enough

PR #993, opened 2020-08-12
Refactored OrbitSsh

antonrohr:
Moved the struct AddrAndPort out of Socket into its own file. Changed
all uses accordingly. refactored namings and execution order to follow
google style guide.

### DIFF
--- a/OrbitQt/OrbitStartupWindow.h
+++ b/OrbitQt/OrbitStartupWindow.h
@@ -50,11 +50,10 @@ class OrbitStartupWindow : public QDialog {
     if (std::holds_alternative<SshInfo>(result_)) {
       auto& ssh_info = std::get<SshInfo>(result_);
       Credentials credentials{};
-      credentials.host = ssh_info.host.toStdString();
+      credentials.addr_and_port = {ssh_info.host.toStdString(), ssh_info.port};
       credentials.key_path = ssh_info.key_path.toStdString();
       credentials.known_hosts_path = ssh_info.known_hosts_path.toStdString();
       credentials.user = ssh_info.user.toStdString();
-      credentials.port = ssh_info.port;
       return outcome::success(std::move(credentials));
     } else if (std::holds_alternative<QString>(result_)) {
       return outcome::success(std::get<QString>(result_));
--- a/OrbitSsh/ContextTests.cpp
+++ b/OrbitSsh/ContextTests.cpp
@@ -12,7 +12,7 @@ namespace OrbitSsh {
 
 TEST(Context, Create) {
   auto context = Context::Create();
-  ASSERT_TRUE(context.has_value());
+  ASSERT_TRUE(context);
   ASSERT_TRUE(context.value().active());
 
   auto context2 = std::move(context.value());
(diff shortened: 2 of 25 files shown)


OrbitQt/OrbitStartupWindow.h
```diff
@@ -50,11 +50,10 @@ class OrbitStartupWindow : public QDialog {
     if (std::holds_alternative<SshInfo>(result_)) {
       auto& ssh_info = std::get<SshInfo>(result_);
       Credentials credentials{};
-      credentials.host = ssh_info.host.toStdString();
```
dimitry-, line 53: Commit message - second line: Capitalize Refactored
last line: google C++ style guilde

OrbitSsh/ContextTests.cpp
```diff
@@ -12,11 +12,11 @@ namespace OrbitSsh {
...
   auto context = Context::Create();
-  ASSERT_TRUE(context.has_value());
-  ASSERT_TRUE(context.value().active());
+  ASSERT_TRUE(context);
+  ASSERT_TRUE(context.value().Active());
 
   auto context2 = std::move(context.value());
-  ASSERT_FALSE(context.value().active());
-  ASSERT_TRUE(context2.active());
+  ASSERT_FALSE(context.value().Active());
```
dimitry-, line 19: this is trivial getter - the reason it was named with active()

OrbitSsh/include/OrbitSsh/Context.h
```diff
@@ -21,7 +21,7 @@ class Context {
 
   ~Context() noexcept;
 
-  [[nodiscard]] bool active() const { return active_; }
+  [[nodiscard]] bool Active() const { return active_; }
```
dimitry-, line 24: if you want to capitalize it it should be IsActive or GetActive to make it clear this is an accessor. But I prefer if it remained active()

OrbitSsh/ContextTests.cpp
antonrohr: Did we decide on doing that? There are several situations related to protos where this also comes up

antonrohr: I found this is the style guide:
> Regular functions have mixed case; accessors and mutators may be named like variables.
so I guess will change it back to `active()`

dimitry-: I am fine both ways, my personal preference is to use them for trivial accessors. Since style guide says may - you can do it one way or the other.

OrbitQt/OrbitStartupWindow.h
antonrohr: I guess commit message changing means a force push is necessary. Does gerrit also solve that? ;)

OrbitSsh/ContextTests.cpp
antonrohr: done

OrbitSsh/include/OrbitSsh/Context.h
antonrohr: done

PR #996, opened 2020-08-12
OrbitLinuxTracing: Reduce number of reinterpret_casts

dimitry-:
When functions expects a chunk of memory
pass void* instead of uint8_t*, then we can
remove unnecessary reinterpret_casts to uint8_t

Also replace casts from void* to T* with static_casts.

There is one place where reinterpret_cast was
used to memcopy - it was replaced with explicit memcpy.

### DIFF
--- a/OrbitLinuxTracing/LibunwindstackUnwinder.cpp
+++ b/OrbitLinuxTracing/LibunwindstackUnwinder.cpp
@@ -31,7 +31,7 @@ const std::array<size_t, unwindstack::X86_64_REG_LAST>
 std::vector<unwindstack::FrameData> LibunwindstackUnwinder::Unwind(
     unwindstack::Maps* maps,
     const std::array<uint64_t, PERF_REG_X86_64_MAX>& perf_regs,
-    const char* stack_dump, uint64_t stack_dump_size) {
+    const void* stack_dump, uint64_t stack_dump_size) {
   unwindstack::RegsX86_64 regs{};
   for (size_t perf_reg = 0; perf_reg < unwindstack::X86_64_REG_LAST;
        ++perf_reg) {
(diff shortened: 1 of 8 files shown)


OrbitLinuxTracing/LibunwindstackUnwinder.cpp
```diff
@@ -31,7 +31,7 @@ const std::array<size_t, unwindstack::X86_64_REG_LAST>
 std::vector<unwindstack::FrameData> LibunwindstackUnwinder::Unwind(
     unwindstack::Maps* maps,
     const std::array<uint64_t, PERF_REG_X86_64_MAX>& perf_regs,
-    const char* stack_dump, uint64_t stack_dump_size) {
+    const void* stack_dump, size_t stack_dump_size) {
```
dpallotti, line 34: Can we please stick to `uint64_t`, like everywhere else in this module?

OrbitLinuxTracing/PerfEventRingBuffer.h
```diff
@@ -36,15 +36,15 @@ class PerfEventRingBuffer {
...
-    ReadAtOffsetFromTail(reinterpret_cast<uint8_t*>(value), offset, sizeof(T));
+    ReadAtOffsetFromTail(value, offset, sizeof(T));
   }
 
-  void ReadRawAtOffset(uint8_t* dest, uint64_t offset, uint64_t count) {
+  void ReadRawAtOffset(void* dest, uint64_t offset, uint64_t count) {
     ReadAtOffsetFromTail(dest, offset, count);
   }
 
   void ReadRawAtOffset(char* dest, uint64_t offset, uint64_t count) {
```
dimitry-: this can be removed now.

OrbitLinuxTracing/LibunwindstackUnwinder.cpp
dimitry-: done.

OrbitLinuxTracing/PerfEventReaders.cpp
```diff
@@ -86,7 +86,7 @@ std::unique_ptr<CallchainSamplePerfEvent> ConsumeCallchainSamplePerfEvent(
   // TODO(kuebler): we should have templated read methods
```
dpallotti, line 86: Does this TODO still apply by the way, now that we pass `void*`?

dimitry-: I think it does - let me see if I can address this (in a follow up PR)

dimitry-: + @florian-kuebler

PR #997, opened 2020-08-12
Fixes a crash when selecting a different process

florian-kuebler:
This bug occurred when there were symbols loaded and functions hooked.
In this case the process gets updated internally and the UI tries to
render the function window before its `OnDataChanged` event arrived,
as `UpdateModules` fires the refresh callbacks at the very end initially
spawned from another thread.

This fix pospones the setting of the global variable GTargetProcess
until all data is ready to propagete the UI update (in particular the
modules list). So if the UI redraws before the UpdateModules function
propagates the data update, there will be no inconsistent state and
thus no crash. As UpdateModules now also sets the process, the method
is also renamed.

This also slightly refactors the Capture::SetTargetProcess by making
it non-ref parameter (as it is a sink) and move the process there and
further rename the parameter.

Bug: http://b/163786513
Test: On Linux: Select a process, load some modules and hook functions.
        Then select another process. The previous crash is gone now.

### DIFF
--- a/OrbitCore/Capture.cpp
+++ b/OrbitCore/Capture.cpp
@@ -33,8 +33,8 @@ void Capture::Init() { GTargetProcess = std::make_shared<Process>(); }
 
 void Capture::SetTargetProcess(std::shared_ptr<Process> process) {
   if (process != GTargetProcess) {
-    GTargetProcess = std::move(process);
     GSamplingProfiler = std::make_shared<SamplingProfiler>(process);
+    GTargetProcess = std::move(process);
     GSelectedFunctionsMap.clear();
   }
 }
--- a/OrbitCore/Capture.cpp
+++ b/OrbitCore/Capture.cpp
@@ -31,10 +31,10 @@ std::shared_ptr<PresetFile> Capture::GSessionPresets = nullptr;
 
 void Capture::Init() { GTargetProcess = std::make_shared<Process>(); }
 
-void Capture::SetTargetProcess(const std::shared_ptr<Process>& a_Process) {
-  if (a_Process != GTargetProcess) {
-    GTargetProcess = a_Process;
-    GSamplingProfiler = std::make_shared<SamplingProfiler>(a_Process);
+void Capture::SetTargetProcess(std::shared_ptr<Process> process) {
+  if (process != GTargetProcess) {
+    GTargetProcess = std::move(process);
+    GSamplingProfiler = std::make_shared<SamplingProfiler>(process);
     GSelectedFunctionsMap.clear();
   }
 }
--- a/OrbitGl/App.cpp
+++ b/OrbitGl/App.cpp
@@ -813,7 +813,7 @@ void OrbitApp::LoadModulesFromPreset(
   }
 }
 
-void OrbitApp::UpdateModuleList(int32_t pid) {
+void OrbitApp::UpdateProcessAndModuleList(int32_t pid) {
   thread_pool_->Schedule([pid, this] {
     ErrorMessageOr<std::vector<ModuleInfo>> result =
         process_manager_->LoadModuleList(pid);
@@ -838,6 +838,7 @@ void OrbitApp::UpdateModuleList(int32_t pid) {
       // TODO: remove this part when all client code is moved to
       // new data model.
       std::shared_ptr<Process> process = FindProcessByPid(pid);
+      CHECK(process != nullptr);
 
       for (const ModuleInfo& info : module_infos) {
         // if module already exists, don't create it again.
@@ -862,6 +863,13 @@ void OrbitApp::UpdateModuleList(int32_t pid) {
       }
       // To this point ----------------------------------
 
+      // To this point all data is ready. We can set the Process and then
+      // propagate the changes to the UI.
+
+      if (pid != Capture::GTargetProcess->GetID()) {
+        Capture::SetTargetProcess(std::move(process));
+      }
+
       FireRefreshCallbacks();
     });
   });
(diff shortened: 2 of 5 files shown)


OrbitGl/App.cpp
```diff
@@ -873,6 +873,7 @@ void OrbitApp::OnProcessSelected(int32_t pid) {
   std::shared_ptr<Process> process = FindProcessByPid(pid);
   CHECK(process != nullptr);
   Capture::SetTargetProcess(std::move(process));
+  FireRefreshCallbacks();
```
pierricgimmig, line 876: Could this be moved below UpdateModuleList(pid); ?

florian-kuebler: This is actually already in UpdateModuleList(pid) at the very end. However, this method spawns a task on another thread (from the main thread) which then again spawns a task on the main thread. In this last task the FireRefreshCallbacks is called at the very end. And the UI seems to redraw the FunctionsView before that call arrives, and thus has inconsistent data (new process (including functions) and old list of indices of functions to draw), which finally leads to the crash.

I am not sure if there is a better way to solve this without blocking the UI for longer time.
Steps to reproduce and further info are in the bug (b/163786513). I would leave this open for now if you want to look a bit further into this and will merge that change for now.

florian-kuebler:
> Hi Florian, thanks for the fix. Your description mentions that this is a naive fix, is there a better fix that we could implement? If so we should open a bug. I'm ok with submitting as is as it fixes a crash, but we should possibly look further into it. Would you have repro steps or a callstack? Thanks.

Thanks for looking into this. See my [comment above](https://github.com/google/orbit/pull/997#discussion_r469718802) for additional information.

OrbitGl/App.cpp
```diff
@@ -873,6 +873,7 @@ void OrbitApp::OnProcessSelected(int32_t pid) {
   std::shared_ptr<Process> process = FindProcessByPid(pid);
   CHECK(process != nullptr);
   Capture::SetTargetProcess(std::move(process));
```
dimitry-, line 875: Should we set taget process after modules have successfully been updated?

florian-kuebler: I guess UpdateModules uses the process. But we can pass it as a param. Maybe this is better. I'll look into this again.

OrbitGl/App.cpp
antonrohr: I am pretty indifferent between this line as a fix or dimitrys idea in the comment below. I think both should work. If you go with this, maybe `FireRefreshCallbacks(DataViewType::FUNCTIONS);` is enough?

OrbitGl/App.cpp
antonrohr: UpdateModuleList(pid) also uses FindProcessById(pid). You can probably leave it as is. If I understood the problem, the problem is having a delay between SetTargetProcess and FireRefreshCallbacks.

florian-kuebler: Yep you were right. I implemented it this way. Seems to be the proper solution for the issue.

florian-kuebler:
Thanks for the reviews. I changed the fix in order to set the process in UpdateModules at the end.

OrbitGl/App.cpp
florian-kuebler: I also tried this, but I am not sure, if there were similar bugs in the other views (even though there shouldn't). Anyways, went for the other solution.

florian-kuebler:
Forgot to mention, also updated the commit message.

OrbitGl/App.cpp
```diff
@@ -873,6 +873,7 @@ void OrbitApp::OnProcessSelected(int32_t pid) {
   std::shared_ptr<Process> process = FindProcessByPid(pid);
```
dimitry-, line 873: ok, one small thing about commit message

Move link to the bug into Bug: stanza

At the end of commit message add
Bug: b/42
Test: <describe how you tested it>

florian-kuebler: Okay done.

OrbitCore/Capture.cpp
```diff
@@ -31,10 +31,10 @@ std::shared_ptr<PresetFile> Capture::GSessionPresets = nullptr;
 
 void Capture::Init() { GTargetProcess = std::make_shared<Process>(); }
```
dimitry-, line 32: nit: you can remove brackets around (b/42) and if you add http:// before you will be able to open it from console while looking at git log (as it is now you will need to copy and paste it)

florian-kuebler: Sure. Done!

dimitry-: ehm, test does not describe how to test this change. I will not be able to reproduce it.

OrbitGl/App.cpp
```diff
@@ -860,7 +861,12 @@ void OrbitApp::UpdateModuleList(int32_t pid) {
         LoadModulesFromPreset(process, preset);
         Capture::GSessionPresets = nullptr;
       }
-      // To this point ----------------------------------
```
dimitry-, line 863: This was a terminator for TODO comment :)

florian-kuebler: Ups.  Sorry. Readded it again.

florian-kuebler:
Added proper "Test:" description.

OrbitCore/Capture.cpp
```diff
@@ -31,10 +31,10 @@ std::shared_ptr<PresetFile> Capture::GSessionPresets = nullptr;
...
 
-void Capture::SetTargetProcess(const std::shared_ptr<Process>& a_Process) {
-  if (a_Process != GTargetProcess) {
-    GTargetProcess = a_Process;
-    GSamplingProfiler = std::make_shared<SamplingProfiler>(a_Process);
+void Capture::SetTargetProcess(std::shared_ptr<Process> process) {
+  if (process != GTargetProcess) {
+    GTargetProcess = std::move(process);
+    GSamplingProfiler = std::make_shared<SamplingProfiler>(process);
```
antonrohr, line 37: Don't use `process` after move

OrbitGl/App.cpp
```diff
@@ -870,12 +878,8 @@ void OrbitApp::UpdateModuleList(int32_t pid) {
 void OrbitApp::OnProcessSelected(int32_t pid) {
```
antonrohr, line 878: Nit: This function can be removed

OrbitCore/Capture.cpp
florian-kuebler: Got me! Thanks!

OrbitGl/App.cpp
florian-kuebler: I removed the function.

PR #999, opened 2020-08-12
SamplingReport refactoring

IrinaShkviro:


OrbitGl/SamplingReport.cpp
```diff
@@ -4,48 +4,44 @@
 
 #include "SamplingReport.h"
 
-#include <utility>
```
dimitry-, line 7: please add a short description to the commit, you renamed fields/variables, added nodiscard where appropriate. get thread report now returns a pointer, and you changed return type from string to std::string.

OrbitGl/SamplingReport.h
```diff
@@ -5,50 +5,55 @@
...
-  }
-  void SetCallstackDataView(class CallStackDataView* a_DataView) {
-    m_CallstackDataView = a_DataView;
-  }
-  void OnSelectAddress(uint64_t a_Address, int32_t a_ThreadId);
-  void OnCallstackIndexChanged(size_t a_Index);
+  [[nodiscard]] std::shared_ptr<SamplingProfiler> GetProfiler() const {
+    return profiler_;
+  };
+  [[nodiscard]] std::vector<SamplingReportDataView>* GetThreadReports() {
```
dimitry-, line 26: I think it is fine to keep a reference here, this is not-nullable. And pointer were required only for output parameters. Also - I think this has changed because I don't find "use pointer for output parameter" rule in the styleguide. Or is there another reason for making this return a pointer?

IrinaShkviro: I thought that we should use const reference if object won't be modified and pointer otherwise, but can't find it in style guide now, so I should revert this change.

dimitry-: Yep there was a rule about this - but iirc it was only for input/output parameters. For return values it did not apply.

OrbitGl/SamplingReport.cpp
```diff
@@ -55,84 +51,84 @@ void SamplingReport::UpdateReport() {
...
 
-  OnCallstackIndexChanged(m_SelectedCallstackIndex);
+  OnCallstackIndexChanged(selected_callstack_index_);
 }
 
-std::string SamplingReport::GetSelectedCallstackString() {
-  if (m_SelectedSortedCallstackReport) {
-    int num_occurrences = m_SelectedSortedCallstackReport
-                              ->callstacks_count[m_SelectedCallstackIndex]
+const std::string& SamplingReport::GetSelectedCallstackString() const {
```
florian-kuebler, line 105: Not sure, if this is correct. Doesn't it create the returned string only on the callee's stack. Thus might be invalid on the caller? Maybe I am wrong, then forget about my comment.

OrbitGl/SamplingReport.h
florian-kuebler: From [here](https://google.github.io/styleguide/cppguide.html#Inputs_and_Outputs):
```
 Input parameters should usually be values or const references, while required (non-nullable) output and input/output parameters should usually be references. Generally, use absl::optional to represent optional inputs, and non-const pointers to represent optional outputs.
```

So I guess optional output params should be pointer, non optional refs.

florian-kuebler: Even though I am not sure about whether this is a good idea. References are always implicit on the caller side, to it makes it more easy to mistakenly pass a ref that gets then modified.

OrbitGl/SamplingReport.cpp
IrinaShkviro: I reverted the change

florian-kuebler: Thanks

OrbitGl/SamplingReport.cpp
IrinaShkviro: I added the commit message in third commit (will squash 2nd and 3rd) to avoid force-update for now

florian-kuebler:
Regarding the commit message: According to the new guidelines (go/stadia-orbit-commit). You should probably use an imperative style in the header:
"Refactor SamplingReport"

I would also add a period at the end of the message.

IrinaShkviro:
> Regarding the commit message: According to the new guidelines (go/stadia-orbit-commit). You should probably use an imperative style in the header:
> "Refactor SamplingReport"
> 
> I would also add a period at the end of the message.

Done, thanks!

PR #1002, opened 2020-08-13
Remove Introspection and its unused dependencies

dpallotti:
After all the changes to the service and communication, this no longer works.
Being it behind `#ifdefs`, this doesn't get maintained.
As we are not using it and as manual instrumentation is being introduced, I
would suggest removing it.
The alternative would be to make this work again, but I'm not sure it's worth
the effort of fixing it and maintaining it

---

@pierricgimmig Please let me know your opinion on the matter. 

@IrinaShkviro For your information as we talked about this.


pierricgimmig:
This needs to stay. I was actually planning on making it always available and making it a capture option. I can have a look at why it's not working anymore.

pierricgimmig:
Honestly I think we should develop with it always on to get a better sense of our own performance on the service side.

dpallotti:
> This needs to stay.

Why?

This hasn't been used for months as it has not worked for months.

Either way, let me be very clear than I'm not saying I want to remove the feature. But the current implementation is extremely outdated. And it carries with it a way too many classes that are not used anymore: `TidAndThreadName`, `LinuxTracingBuffer`, `KeyAndString`, `LinuxCallstackEvent`. This is all dead code and it needs to go.

Also if this is only for the tracer, the implementation can be extremely simplified, without having dependencies across `OrbitCore` and `OrbitBase`. So in other words, this should to be re-implemented entirely, I don't see the value in keeping this old code around.

I guess that having a capture option and a command line option on the client would increase the chances that this keep working.

pierricgimmig:
> > This needs to stay.
> 
> Why?

I'm talking about the feature also, not the dependencies, those can go. I have a new implementation almost working btw. It was my mistake to not make this enabled by default from day 1. I really want to push this feature further actually, we have the chance to easily visualize our own perf, I don't want to lose it.  Please leave the macros and introspection related files. All the LinuxTracingBuffer stuff can go. I can also take it from here if you want merge this PR with my changes.

dpallotti:
> I'm talking about the feature also, not the dependencies, those can go. I have a new implementation almost working btw. It was my mistake to not make this enabled by default from day 1. I really want to push this feature further actually, we have the chance to easily visualize our own perf, I don't want to lose it. Please leave the macros and introspection related files. All the LinuxTracingBuffer stuff can go. I can also take it from here if you want and merge this PR with my changes.

Alright, if you already have a new implementation then I'll leave it to you, you can take this PR as an indication of the old dependencies that should be removed or they can be removed afterwards.

pierricgimmig:
The dependencies were remove in #1039. The introspection revival is coming soon in a new PR.

PR #1003, opened 2020-08-13
Refactored OrbitService

antonrohr:
Introduced namespace orbit_service, moved header files into include
directory and other minor refactorings to adhere to google C++ style
guide.

### DIFF
--- a/OrbitService/CaptureServiceImpl.cpp
+++ b/OrbitService/CaptureServiceImpl.cpp
@@ -4,9 +4,10 @@
 
 #include "CaptureServiceImpl.h"
 
-#include <OrbitBase/Logging.h>
-
 #include "LinuxTracingGrpcHandler.h"
+#include "OrbitBase/Logging.h"
+
+namespace orbit_service {
 
 grpc::Status CaptureServiceImpl::Capture(
     grpc::ServerContext*,
@@ -30,3 +31,5 @@ grpc::Status CaptureServiceImpl::Capture(
   LOG("Finished handling gRPC call to Capture: all capture data has been sent");
   return grpc::Status::OK;
 }
+
+}  // namespace orbit_service
--- a/OrbitService/OrbitService.cpp
+++ b/OrbitService/OrbitService.cpp
@@ -14,6 +14,8 @@
 #include "OrbitBase/Logging.h"
 #include "OrbitGrpcServer.h"
 
+namespace {
+
 static std::string ReadStdIn() {
   int tmp = fgetc(stdin);
   if (tmp == -1) return "";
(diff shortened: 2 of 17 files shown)


OrbitService/CaptureServiceImpl.cpp
```diff
@@ -2,14 +2,15 @@
...
-#include <OrbitBase/Logging.h>
+#include "OrbitService/LinuxTracingGrpcHandler.h"
+#include "OrbitBase/Logging.h"
 
-#include "LinuxTracingGrpcHandler.h"
+namespace orbit_service {
 
 grpc::Status CaptureServiceImpl::Capture(
-    grpc::ServerContext*,
+    grpc::ServerContext* /*context*/,
```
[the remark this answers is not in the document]
antonrohr: No it doesn't
> Unused parameters that are obvious from context may be omitted
I removed the comments

OrbitService/OrbitService.cpp
```diff
@@ -12,9 +12,11 @@
 #include <thread>
 
 #include "OrbitBase/Logging.h"
-#include "OrbitGrpcServer.h"
+#include "OrbitService/OrbitGrpcServer.h"
 
-static std::string ReadStdIn() {
+namespace {
+
+std::string ReadStdIn() {
```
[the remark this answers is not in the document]
antonrohr: I readded the static, but left the namespace

OrbitService/include/OrbitService/ProcessList.h
```diff
@@ -10,14 +10,20 @@
 
 #include "process.pb.h"
 
+namespace orbit_service {
+
 class ProcessList {
  public:
-  outcome::result<void, std::string> Refresh();
-  const std::vector<ProcessInfo>& GetProcesses() { return processes_; }
+  [[nodiscard]] outcome::result<void, std::string> Refresh();
```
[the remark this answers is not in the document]
antonrohr: done

OrbitService/CMakeLists.txt
```diff
@@ -10,26 +10,31 @@ project(OrbitServiceLib)
...
+target_include_directories(OrbitServiceLib PUBLIC ${CMAKE_CURRENT_LIST_DIR}/include)
+
+target_sources(OrbitServiceLib PUBLIC
+        include/OrbitService/CaptureServiceImpl.h
+        include/OrbitService/CrashServiceImpl.h
+        include/OrbitService/FramePointerValidatorServiceImpl.h
+        include/OrbitService/OrbitGrpcServer.h
+        include/OrbitService/OrbitService.h
+        include/OrbitService/ProcessList.h
+        include/OrbitService/ProcessServiceImpl.h)
```
dimitry-, line 22: The idea behind having include dir is to place headers that are used by other modules to it. That way we have a separation between public headers and private ones.

Service is not used by anyone so in this case we do not need to have public headers.

antonrohr: So are you suggesting to also get rid of the include folder or to just not make them public in the cmake file? 
I like having a separate folder and the prefix in the `#includes`, I think it is more organised and easier to understand where the includes are coming from

dimitry-: Since we should only move headers that are supposed to be used from the outside (other modules). I say do not move them to include for the service.

antonrohr: done

PR #1005, opened 2020-08-13
Remove unused code using Log.h

dimitry-:
Remove unused code using Log.h macros
and move the rest of the code to use
OrbitBase/Logging.h.

Test: builds

### DIFF
--- a/OrbitCore/Log.h
+++ b/OrbitCore/Log.h
@@ -10,10 +10,6 @@
 
 #include "Threading.h"
 
-#define ORBIT_LOG(msg) GLogger.Log(OrbitLog::Global, msg)
-#define ORBIT_LOGV(var) GLogger.Log(OrbitLog::Global, #var, var)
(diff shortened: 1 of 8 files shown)


OrbitCore/Log.h
```diff
@@ -10,10 +10,6 @@
 
 #include "Threading.h"
 
-#define ORBIT_LOG(msg) GLogger.Log(OrbitLog::Global, msg)
-#define ORBIT_LOGV(var) GLogger.Log(OrbitLog::Global, #var, var)
-#define ORBIT_VIZV(var) GLogger.Log(OrbitLog::Viz, #var, var)
-
 class OrbitLog {
```
IrinaShkviro, line 13: Do we use `OrbitLog` and `Logger` classes after this change?

IrinaShkviro: Sorry, just noticed that you removed it in the next PR

dimitry-: #1007

PR #1010, opened 2020-08-13
Kill PrintVar.h

dimitry-:
This commit moves VAR_TO_STR macro
to only please where it is used - CaptureWindow
and removes PrintVar.h

Test: builds


OrbitGl/CaptureWindow.cpp
```diff
@@ -662,6 +661,13 @@ float CaptureWindow::GetTopBarTextY() {
          m_TextRenderer.GetStringHeight("FpjT_H") * 0.5f;
 }
 
+template <class T>
+static std::string VariableToString(const char* name, const T& value) {
```
beckerhe, line 665: Nit: name could be a std::string_view

dimitry-: done.

PR #1014, opened 2020-08-13
Remove outdated files

dimitry-:
These files do not seem to be used anymore


florian-kuebler:
Looping in @antonrohr, as he is using VSCode.

beckerhe:
As I've used VSCode more often recently I've also developed an opinion on that.

My feeling is that these configuration files can't be made to suit everybody's workflow. Some parts work on Linux, some on Windows and so on. Every change in the future will probably break someone else's workflow.

Making your own changes gets a lot more annoying because now you're fighting git and git's auto-completion which always wants to git-add the chunks.

So I prefer removing it and also adding `.vscode/` to `.gitignore`.

We could also move these files somewhere else and consider it a template which can be copied to `.vscode/` if needed.

antonrohr:
I generally agree with Hennings comment. I do use my own tasks and launch files. 

> We could also move these files somewhere else and consider it a template which can be copied to .vscode/ if needed
I think this is an excellent idea. I think I remember @ronaldfw wanting those files to work, for reasons I don't know. It might be related to @kisenshi's onboarding.

pierricgimmig:
We should add those files to the gitignore file.

PR #1015, opened 2020-08-13
Fix some warnings and headers in OrbitQt

dpallotti:


OrbitQt/orbitaboutdialog.h
```diff
@@ -22,12 +22,12 @@ class OrbitAboutDialog : public QDialog {
   void SetVersionString(const QString& version);
   void SetBuildInformation(const QString& build_info);
 
-  ~OrbitAboutDialog() noexcept;
+  ~OrbitAboutDialog() noexcept override;
```
dimitry-, line 25: this is interesting. I thought we had a error for this.

dpallotti: You have an error if you put the `= default` here instead of having the definition outside of the class. I didn't understand why.

dpallotti:
> This is coming from clang-tidy, isn't it?

Some of them yes.

OrbitQt/orbitaboutdialog.h
beckerhe: You have the error because Ui::OrbitAboutDialog is an incomplete type (forward declared) and the destructor needs to full type. So we can either include the header for Ui::OrbitAboutDialog or put the destructor somewhere where Ui::OrbitAboutDialog is fully defined. That is what's happening here.

Regarding the error dimitry is asking about: The error does not apply to destructors because there is basically nothing you can do wrong: No parameters and typos are caught anyway.

dpallotti: Yep, so I moved the inclusion of the header to this file.

beckerhe: > Yep, so I moved the inclusion of the header to this file.

:+1:  Thanks!

PR #1017, opened 2020-08-13
Moved GetServiceLogFilePath into OrbitService main

antonrohr:
This is a small part of my effort to get rid of OrbitCore in OrbitService

### DIFF
--- a/OrbitCore/Path.cpp
+++ b/OrbitCore/Path.cpp
@@ -255,12 +255,6 @@ std::string Path::GetHome() {
   std::string home = GetEnvVar("HOME") + "/";
   return home;
 }
-
-std::string Path::GetServiceLogFilePath() {
-  std::string logsDir = Path::JoinPath({"/", "var", "log"});
-  std::filesystem::create_directory(logsDir);
-  return Path::JoinPath({logsDir, "OrbitService.log"});
-}
 #endif
 
 std::vector<std::string> Path::ListFiles(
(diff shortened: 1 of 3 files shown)


OrbitCore/Path.cpp
```diff
@@ -255,12 +255,6 @@ std::string Path::GetHome() {
   std::string home = GetEnvVar("HOME") + "/";
   return home;
 }
-
```
dimitry-, line 258: nit: maybe explain in the commit message that it is only being used on the service side anyways.

OrbitService/main.cpp
```diff
@@ -38,7 +38,9 @@ void InstallSigintHandler() {
 }  // namespace
 
 int main(int argc, char** argv) {
-  const std::string log_file_path = Path::GetServiceLogFilePath();
+  std::filesystem::path var_log{"/var/log"};
+  std::filesystem::create_directory(var_log);
+  const std::string log_file_path = var_log / "OrbitService.log";
```
dimitry-, line 43: also maybe preserve a function here just make it static?

antonrohr: just to declutter the main method? Okay, will do

OrbitCore/Path.cpp
antonrohr: will do on rebase

OrbitService/main.cpp
antonrohr: done

PR #1021, opened 2020-08-13
Add LOG_VAR macro

pierricgimmig:
LOG_VAR logs a variable's name and value in the form "name = value".

This was previously implemented in OrbitCore as PRINT_VAR. I use it all the time 
for debugging so I'm re-adding it, in OrbitBase this time.

### DIFF
--- a/OrbitBase/include/OrbitBase/Logging.h
+++ b/OrbitBase/include/OrbitBase/Logging.h
@@ -8,6 +8,7 @@
 #include <cstdio>
 #include <filesystem>
 #include <fstream>
+#include <string>
 
 #ifdef _WIN32
 #include <Windows.h>
@@ -33,6 +34,8 @@
         "[%28s] " format "\n", file_and_line__.c_str(), ##__VA_ARGS__);       \
     PLATFORM_LOG(formatted_log__.c_str());                                    \
   } while (0)
+
+#define LOG_VAR(x) LOG("%s = %s", #x, std::to_string(x))
 
 #if defined(_WIN32) && defined(ERROR)
 #undef ERROR


OrbitBase/include/OrbitBase/Logging.h
```diff
@@ -34,6 +35,8 @@
     PLATFORM_LOG(formatted_log__.c_str());                                    \
   } while (0)
 
+#define LOG_VAR(x) LOG("%s = %s", #x, std::to_string(x))
```
dimitry-, line 38: Hide it behind ifdef DEBUG?

pierricgimmig: No, I especially need it in release actually, it's kind of the point :-). It's really just syntactic sugar for the LOG macro...

dimitry-: ok, this is something that us used locally but is not supposed to be actually commited to repo, right? If so I want to have some kind of protection against accidental uses of this instead of regular and more readable LOG messages in cases where LOG is intended and not for debug reasons.

dpallotti: I agree. This is a debug utility more than a logging utility. That's why it was removed. We don't want usages of `LOG_VAR` to actually end up in the code base, or at least to not have an effect on logs in release, as the messages are quite cryptic. Of course one might still use it while developing but still building with optimizations, so I guess `#ifndef NDEBUG` is not the way to go, but this shouldn't be added as is to `Logging.h`.

pierricgimmig: Guys, you are killing me. This has been in the code base for ages, anything it can do LOG can do as well, it's pure syntactic sugar.

PR #1022, opened 2020-08-14
Clean up disassembler

pierricgimmig:
Remove LOG and LOGF macros in Disassembler.cpp. @florian-kuebler, I tried to make the the line_to_address_ code more robust, please let me know if there is any functional difference. I might have introduced an off-by-one here, not sure.

### DIFF
--- a/OrbitGl/Disassembler.cpp
+++ b/OrbitGl/Disassembler.cpp
@@ -4,27 +4,9 @@
 
 #include "Disassembler.h"
 
+#include <absl/strings/str_replace.h>
 #include <capstone/capstone.h>
 #include <capstone/platform.h>
-
-#define LOGF(format, ...)                                   \
-  {                                                         \
-    std::string log = absl::StrFormat(format, __VA_ARGS__); \
-    result_ += log;                                         \
-  }
-
-#define LOG(str) \
-  { result_ += str; }
-
-void Disassembler::LogHex(const uint8_t* str, size_t len) {
-  const unsigned char* c;
-
-  LOGF("%s", "Code: ");
-  for (c = str; c < str + len; c++) {
-    LOGF("0x%02x ", *c & 0xff);
-  }
-  LOGF("%s", "\n");
-}
 
 void Disassembler::Disassemble(const void* machine_code, size_t size,
                                uint64_t address, bool is_64bit) {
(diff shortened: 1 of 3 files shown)


OrbitGl/Disassembler.cpp
```diff
@@ -53,27 +33,33 @@ void Disassembler::Disassemble(const void* machine_code, size_t size,
...
 uint64_t Disassembler::GetAddressAtLine(size_t line) const {
   if (line >= line_to_address_.size()) return 0;
   return line_to_address_[line];
-}
\ No newline at end of file
+}
+
+void Disassembler::LogLine(std::string line, uint64_t address) {
+  // Remove any new line character.
+  absl::StrReplaceAll(line, {{"\n", ""}});
```
dimitry-, line 62: where are they coming from?

OrbitGl/Disassembler.cpp
```diff
@@ -53,27 +33,33 @@ void Disassembler::Disassemble(const void* machine_code, size_t size,
...
 }
 
 uint64_t Disassembler::GetAddressAtLine(size_t line) const {
   if (line >= line_to_address_.size()) return 0;
   return line_to_address_[line];
-}
\ No newline at end of file
+}
+
+void Disassembler::LogLine(std::string line, uint64_t address) {
```
dimitry-, line 60: AddLine?

OrbitGl/Disassembler.cpp
pierricgimmig: There's a mapping between lines and addresses that needs to be maintained, if someone was to call LogLine with a string containing a new line, it would mess up that mapping, that's why it's crucial to guard against it. @florian-kuebler, am I understanding this correctly?

OrbitGl/Disassembler.cpp
pierricgimmig: done

OrbitGl/Disassembler.cpp
florian-kuebler: Right. However, I find your current solution much cleaner.

pierricgimmig:
> Thanks for the change. Was on my agenda already quite some time.
> LGTM for now.
> For the future, I would like to refactor this a bit, such that the meta data at the beginning would be a separate member of the disassembler. So would have such a mess with the line-to-address mapping. Also this would allow to display e.g. the filename in the title of the window.
> Anyways, as said, for now its good to get rid of the warnings.

Awesome, sounds good. Thanks for the review!

PR #1023, opened 2020-08-14
Simplified interaction of Batcher and PickingManager

reichlfl:
This fix make the interaction between the Batcher and PickingManager less error-prone and easier to use. Before, `Pickable` elements used a hack in drawing to work correctly - they calculated their picking color as "regular" color and ignored the picking color set by the batcher. It was also possible to add user_data to pickables which could never be retrieved lateron.

The batcher now handles the picking color generation for pickables, and all batchers are drawn with the correct picking flag instead of hacking the picking in as before.

This required moving some parts of CaptureWindow into the parent GlCanvas. These classes require a seperate cleanup in a subsequent PR.

Note that there's two TODOs for text rendering. The commented lines seem like a previously existing bug (passing picking type kPickable, but there is no Pickable element or correctly calculated color), so these changes should not influence anything. THis should also be tackled in a subsequent cleanup.

### DIFF
--- a/OrbitGl/BatcherTest.cpp
+++ b/OrbitGl/BatcherTest.cpp
@@ -1,0 +1,207 @@
+// Copyright (c) 2020 The Orbit Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#include <gtest/gtest.h>
+
+#include "Batcher.h"
+#include "PickingManagerTest.h"
+
+namespace {
+
+class MockBatcher : public Batcher {
+ public:
+  MockBatcher(BatcherId id, PickingManager* picking_manager = nullptr)
+      : Batcher(id, picking_manager) {}
+
+  void ResetMockDrawCounts() {
+    drawn_line_colors_.clear();
+    drawn_triangle_colors_.clear();
+    drawn_box_colors_.clear();
+  }
+
+  const std::vector<Color>& GetDrawnLineColors() const {
+    return drawn_line_colors_;
+  }
+  const std::vector<Color>& GetDrawnTriangleColors() const {
+    return drawn_triangle_colors_;
+  }
+  const std::vector<Color>& GetDrawnBoxColors() const {
+    return drawn_box_colors_;
+  }
+
+  // Simulate drawing by simple appending all colors to internal
+  // buffers. Only a single color per element will be appended
+  // (start point for line, first vertex for triangle and box)
+  void Draw(bool picking = false) const override {
+    if (picking) {
+      for (auto it = line_buffer_.picking_colors_.begin();
+           it != line_buffer_.picking_colors_.end();) {
+        drawn_line_colors_.push_back(*it);
+        ++it;
+        ++it;
+      }
+      for (auto it = triangle_buffer_.picking_colors_.begin();
+           it != triangle_buffer_.picking_colors_.end();) {
+        drawn_triangle_colors_.push_back(*it);
+        ++it;
+        ++it;
+        ++it;
+      }
+      for (auto it = box_buffer_.picking_colors_.begin();
+           it != box_buffer_.picking_colors_.end();) {
+        drawn_box_colors_.push_back(*it);
+        ++it;
+        ++it;
+        ++it;
+        ++it;
+      }
+    } else {
+      for (auto it = line_buffer_.colors_.begin();
+           it != line_buffer_.colors_.end();) {
+        drawn_line_colors_.push_back(*it);
+        ++it;
+        ++it;
+      }
+      for (auto it = triangle_buffer_.colors_.begin();
+           it != triangle_buffer_.colors_.end();) {
+        drawn_triangle_colors_.push_back(*it);
+        ++it;
+        ++it;
+        ++it;
+      }
+      for (auto it = box_buffer_.colors_.begin();
+           it != box_buffer_.colors_.end();) {
+        drawn_box_colors_.push_back(*it);
+        ++it;
+        ++it;
+        ++it;
+        ++it;
+      }
+    }
+  }
+
+ private:
+  mutable std::vector<Color> drawn_line_colors_;
+  mutable std::vector<Color> drawn_triangle_colors_;
+  mutable std::vector<Color> drawn_box_colors_;
+};
+
+void ExpectDraw(MockBatcher& batcher, uint32_t line_count,
+                uint32_t triangle_count, uint32_t box_count) {
+  batcher.ResetMockDrawCounts();
+  batcher.Draw();
+  EXPECT_EQ(batcher.GetDrawnLineColors().size(), line_count);
+  EXPECT_EQ(batcher.GetDrawnTriangleColors().size(), triangle_count);
+  EXPECT_EQ(batcher.GetDrawnBoxColors().size(), box_count);
+}
+
+TEST(Batcher, SimpleElementsDrawing) {
+  MockBatcher batcher(BatcherId::kUi);
+
+  ExpectDraw(batcher, 0, 0, 0);
+  batcher.AddLine(Vec2(0, 0), Vec2(1, 0), 0, Color(255, 255, 255, 255));
+  ExpectDraw(batcher, 1, 0, 0);
+  EXPECT_EQ(batcher.GetDrawnLineColors()[0], Color(255, 255, 255, 255));
+  batcher.AddTriangle(Triangle(Vec3(0, 0, 0), Vec3(0, 1, 0), Vec3(1, 0, 0)),
+                      Color(0, 255, 0, 255));
+  ExpectDraw(batcher, 1, 1, 0);
+  EXPECT_EQ(batcher.GetDrawnTriangleColors()[0], Color(0, 255, 0, 255));
+  batcher.AddBox(Box(Vec2(0, 0), Vec2(1, 1), 0), Color(255, 0, 0, 255));
+  ExpectDraw(batcher, 1, 1, 1);
+  EXPECT_EQ(batcher.GetDrawnBoxColors()[0], Color(255, 0, 0, 255));
+  batcher.Reset();
+  ExpectDraw(batcher, 0, 0, 0);
+}
+
+TEST(Batcher, PickingElementsDrawing) {
+  MockBatcher batcher(BatcherId::kUi);
+  std::shared_ptr<PickableMock> pickable = std::make_shared<PickableMock>();
+  PickingManager pm;
+
+  ExpectDraw(batcher, 0, 0, 0);
+  EXPECT_DEATH(batcher.AddLine(Vec2(0, 0), Vec2(1, 0), 0,
+                               Color(255, 255, 255, 255), pickable),
+               "nullptr");
+  batcher.SetPickingManager(&pm);
+  batcher.AddLine(Vec2(0, 0), Vec2(1, 0), 0, Color(255, 255, 255, 255),
+                  pickable);
+  ExpectDraw(batcher, 1, 0, 0);
+  batcher.Reset();
+  ExpectDraw(batcher, 0, 0, 0);
+}
+
+template <typename T>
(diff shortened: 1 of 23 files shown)


OrbitGl/PickingManager.h
```diff
@@ -77,9 +81,10 @@ struct PickingId {
     PickingId id = absl::bit_cast<PickingId, uint32_t>(value);
     return id;
   }
-  uint32_t id : 28;
-  PickingType type : 3;
-  BatcherId batcher_id : 1;
+  
+  uint32_t id_ : kIDBitSize;
+  PickingType type_ : kPickingTypeBitSize;
+  BatcherId batcher_id_ : kBatcherIDBitSize;
```
dimitry-: static_assert(kIDBitSize + kPickingTypeBitSize + kBatcherIDBitSize == 32)?

OrbitGl/PickingManager.h
```diff
@@ -53,6 +53,10 @@ enum class PickingType : uint32_t {
 enum class BatcherId : uint32_t { kTimeGraph, kUi };
 
 struct PickingId {
+  static constexpr const uint32_t kIDBitSize = 28;
```
dimitry-: since this is constexpr you can remove static. s/kIDBitSize/kIdBitSize/ - same below

OrbitGl/TimerTrack.cpp
```diff
@@ -63,7 +63,8 @@ void TimerTrack::UpdateBoxHeight() {
   box_height_ = time_graph_->GetLayout().GetTextBoxHeight();
 }
 
-void TimerTrack::UpdatePrimitives(uint64_t min_tick, uint64_t max_tick, PickingMode /*picking_mode*/) {
+void TimerTrack::UpdatePrimitives(uint64_t min_tick, uint64_t max_tick,
+                                  PickingMode /*picking_mode*/) {
```
dimitry-, line 67: nit: you can drop /* */ it is clear from context that parameter is picking mode

OrbitGl/PickingManager.cpp
```diff
@@ -7,28 +7,37 @@
 #include "OpenGl.h"
 #include "OrbitBase/Logging.h"
 
-PickingId PickingManager::CreatePickableId(std::weak_ptr<Pickable> a_Pickable,
-                                           BatcherId batcher_id) {
+PickingId PickingManager::CreateOrGetPickableId(
+    std::weak_ptr<Pickable> pickable, BatcherId batcher_id) {
```
dimitry-, line 11: this weak_ptr is never checked - it may result in nullptr dereference in line 14.

dimitry-: Nit: GetOrCreate is probably better. It creates only if was not able to get.

OrbitGl/GlCanvas.cpp
```diff
@@ -474,3 +480,20 @@ int GlCanvas::AddText2D(const char* a_Text, int a_X, int a_Y, float a_Z,
...
+
+void GlCanvas::ResetHoverTimer() {
+  m_HoverTimer.Reset();
+  m_CanHover = true;
+}
+
+[[nodiscard]] PickingMode GlCanvas::GetPickingMode() {
+  PickingMode picking_mode = PickingMode::kNone;
+  if (m_Picking) {
+    picking_mode = PickingMode::kClick;
```
dimitry-, line 492: if (..) {
  return 1
}

if (..) {
  return 2;
} 

return 0;

OrbitGl/GlCanvas.cpp
```diff
@@ -474,3 +480,20 @@ int GlCanvas::AddText2D(const char* a_Text, int a_X, int a_Y, float a_Z,
...
+  }
+  if (m_IsHovering) {
+    picking_mode = PickingMode::kHover;
```
dimitry-, line 495: I am not sure if this is intended - if case of m_IsHovering and m_Picking this will override PickingMode::kClick.

OrbitGl/GlCanvas.h
```diff
@@ -91,13 +91,17 @@ class GlCanvas : public GlPanel {
...
+  void ResetHoverTimer();
+
   float GetDeltaTimeSeconds() const { return m_DeltaTime; }
 
   virtual void Draw() {}
   virtual void DrawScreenSpace() {}
   virtual void RenderUI() {}
   virtual void RenderText() {}
 
+  virtual void Hover(int a_X, int a_Y) {}
```
dimitry-: /* x */, /* y */

OrbitGl/PickingManager.cpp
```diff
@@ -7,28 +7,37 @@
...
   absl::MutexLock lock(&mutex_);
-  ++id_counter_;
-  PickingId id =
-      PickingId::Create(PickingType::kPickable, id_counter_, batcher_id);
-  id_pickable_map_[id_counter_] = a_Pickable;
+  uint32_t pickable_id = 0;
+  auto it = pickable_pid_map_.find(pickable.lock().get());
```
dimitry-, line 14: lock and check here then use shared_ptr.

OrbitGl/PickingManagerTest.h
```diff
@@ -0,0 +1,40 @@
...
+#ifndef ORBIT_GL_TESTS_PICKING_MANAGER_TEST_H_
+#define ORBIT_GL_TESTS_PICKING_MANAGER_TEST_H_
+
+#include "GlCanvas.h"
+#include "PickingManager.h"
+#include "absl/base/casts.h"
+
+class PickableMock : public Pickable {
+ public:
+  void OnPick(int, int) override { picked_ = true; }
```
dimitry-, line 14: here we need context add /* */ with parameter names.

OrbitGl/PickingManager.h
reichlfl: No, constexpr members can't be static, at least the VS compiler complains. I can remove the const though, that's kind of redundant.

OrbitGl/PickingManager.h
reichlfl: It's technically caught by absl::bit_cast<PickingId, uint32_t> - but it's a good idea to add it here, that makes it more explicit for everyone. Added that, thanks.

OrbitGl/GlCanvas.cpp
reichlfl: Good point... it's like this for a while now, and it will only appear if hover and click trigger in the exact same rendering frame, which is super unlikely. But you're right, I've changed it as suggested in the previous comment, which gets rid of this issue as well.

OrbitGl/PickingManager.cpp
reichlfl: Done

OrbitGl/GlCanvas.cpp
reichlfl: Done

OrbitGl/GlCanvas.h
reichlfl: Done

OrbitGl/PickingManagerTest.h
reichlfl: Done

dimitry-:
nit: you are fixing clang builds not linux builds :)

OrbitGl/PickingManager.cpp
```diff
@@ -7,8 +7,10 @@
 #include "OpenGl.h"
 #include "OrbitBase/Logging.h"
 
-PickingId PickingManager::CreateOrGetPickableId(
+PickingId PickingManager::GetOrCreatePickableId(
     std::weak_ptr<Pickable> pickable, BatcherId batcher_id) {
+  CHECK(!pickable.expired());
```
dimitry-, line 12: auto pickable_ptr = pickable.lock();
CHECK(pickable_ptr);

On the other hand if we do not want this to be null - maybe just pass shared_ptr here instead of weak_ptr. CHECK with weak_ptr seems strange... Usually when weak_ptr used this case is handled.

OrbitGl/BatcherTest.cpp
```diff
@@ -36,41 +36,47 @@ class MockBatcher : public Batcher {
...
         ++it;
+        ++it;
       }
       for (auto it = triangle_buffer_.picking_colors_.begin();
-           it != triangle_buffer_.picking_colors_.end(); ++it) {
+           it != triangle_buffer_.picking_colors_.end();) {
         drawn_triangle_colors_.push_back(*it);
         ++it;
         ++it;
+        ++it;
```
dimitry-, line 49: maybe we should implement += ...

OrbitGl/PickingManager.cpp
reichlfl: In this case we'd need to change more methods up to all the "Batcher::AddXXX" methods to accept a shared_ptr, and the reason for the weak_ptr there was to communicate that there won't be any ownership assumed. Nevertheless it does not make sense to pass an expired weak_ptr at this point... I'd prefer to leave the check, this should never happen but would be good to have just in case.

OrbitGl/PickingManager.cpp
reichlfl: Done

OrbitGl/PickingManager.cpp
dimitry-: ok, please then change the check so that weak pointer use is properly scoped.

OrbitGl/TimerTrack.cpp
reichlfl: I'd prefer to leave it as we use this pattern a lot.

OrbitGl/CMakeLists.txt
```diff
@@ -127,7 +127,11 @@ add_executable(OrbitGlTests)
 target_compile_options(OrbitGlTests PRIVATE ${STRICT_COMPILE_FLAGS})
 
 target_sources(OrbitGlTests PRIVATE
-    PickingManagerTest.cpp)
+    PickingManagerTest.h)
+
+target_sources(OrbitGlTests PRIVATE
+    PickingManagerTest.cpp
+	BatcherTest.cpp)
```
dimitry-, line 134: nit: formatting is a bit off.

OrbitGl/PickingManager.cpp
```diff
@@ -7,26 +7,41 @@
...
+    std::weak_ptr<Pickable> pickable, BatcherId batcher_id) {
+  CHECK(!pickable.expired());
+
   absl::MutexLock lock(&mutex_);
-  ++id_counter_;
-  PickingId id =
-      PickingId::Get(PickingType::kPickable, id_counter_, batcher_id);
-  id_pickable_map_[id_counter_] = a_Pickable;
+  uint32_t pickable_id = 0;
+  auto locked_pickable = pickable.lock();
```
dimitry-, line 16: I am still kind-of uneasy with this. Maybe add a comment why this is not checked for null here? Maybe a TODO: replace this with shared_ptr<> since we cannot handle the case of expired pickable.

OrbitGl/GlCanvas.cpp
reichlfl: I've just realized that this breaks hovering, because it actually was intended this way - m_Picking is set on click and on hover. We should change this to be more precise, like distinguish between click and hover or something. I've changed the check now to retain the old behavior, should refactor this in the future.

OrbitGl/PickingManager.cpp
reichlfl: As discussed off-line, we could add this in a subsequent PR

OrbitGl/GlCanvas.cpp
```diff
@@ -487,7 +487,7 @@ void GlCanvas::ResetHoverTimer() {
 }
 
 [[nodiscard]] PickingMode GlCanvas::GetPickingMode() {
-  if (m_Picking) {
+  if (m_Picking && !m_IsHovering) {
     return PickingMode::kClick;
   }
   if (m_IsHovering) {
```
dimitry-, line 493: well, you can just do this check first I think... and avoid && above.

PR #1026, opened 2020-08-14
Refactor Capture to move GSelectedFunctionsMap to DataManager

florian-kuebler:
The GSelectedFunctionsMap global was previously used from various
places assuming different semantics, which already lead to bugs.
In order to get rid of the globals in Capture and make accesses
to the data stored there more explicit, this change moves the map
to OrbitGl/DataManager and passes it explicitly as input to OrbitCore.

This map represents the current state of the running application,
i.e. wich functions are currently hooked and is not related to the
functions hooked in the previous/loaded capture.
Now this map also stores a copy of the FunctionInfo content, to
avoid nullptrs. Making all maps to store raw pointers is unfeasable,
as there is no dedicated owner of these functions, when loading a
capture (otherwise it would be the process's module).
All accessor methods (in particual to select/unselect/check functions)
are moved to App.

Some usages were replaced with the intended selected functions from the
previous capture (i.e. in CaptureData).

Added [[nodiscard]] and const on various places.

Bug: http://b/163020637
Test: Load Modules, Hook/Unhook functions, do captures, load/store captures,
      load/store presets, repeat captures while/after hooking/unhooking function,
      filter functions.

### DIFF
--- a/OrbitCaptureClient/CaptureClient.cpp
+++ b/OrbitCaptureClient/CaptureClient.cpp
@@ -27,7 +27,7 @@ IntrumentedFunctionTypeFromOrbitType(FunctionInfo::OrbitType orbit_type) {
 
 ErrorMessageOr<void> CaptureClient::StartCapture(
     ThreadPool* thread_pool, int32_t pid,
-    const std::map<uint64_t, FunctionInfo*>& selected_functions) {
+    const absl::flat_hash_map<uint64_t, FunctionInfo>& selected_functions) {
   absl::MutexLock lock(&state_mutex_);
   if (state_ != State::kStopped) {
     return ErrorMessage(
@@ -43,7 +43,8 @@ ErrorMessageOr<void> CaptureClient::StartCapture(
 }
 
 void CaptureClient::Capture(
-    int32_t pid, const std::map<uint64_t, FunctionInfo*>& selected_functions) {
+    int32_t pid,
+    const absl::flat_hash_map<uint64_t, FunctionInfo>& selected_functions) {
   CHECK(reader_writer_ == nullptr);
 
   event_processor_.emplace(capture_listener_);
@@ -69,7 +70,7 @@ void CaptureClient::Capture(
 
   capture_options->set_trace_gpu_driver(true);
   for (const auto& pair : selected_functions) {
-    const FunctionInfo& function = *pair.second;
+    const FunctionInfo& function = pair.second;
     CaptureOptions::InstrumentedFunction* instrumented_function =
         capture_options->add_instrumented_functions();
     instrumented_function->set_file_path(function.loaded_module_path());
--- a/OrbitCaptureClient/CaptureClient.cpp
+++ b/OrbitCaptureClient/CaptureClient.cpp
@@ -27,7 +27,7 @@ IntrumentedFunctionTypeFromOrbitType(FunctionInfo::OrbitType orbit_type) {
 
 ErrorMessageOr<void> CaptureClient::StartCapture(
     ThreadPool* thread_pool, int32_t pid,
-    const absl::flat_hash_map<uint64_t, FunctionInfo>& selected_functions) {
+    absl::flat_hash_map<uint64_t, FunctionInfo> selected_functions) {
   absl::MutexLock lock(&state_mutex_);
   if (state_ != State::kStopped) {
     return ErrorMessage(
@@ -37,14 +37,16 @@ ErrorMessageOr<void> CaptureClient::StartCapture(
 
   state_ = State::kStarting;
   thread_pool->Schedule(
-      [this, pid, selected_functions]() { Capture(pid, selected_functions); });
+      [this, pid, selected_functions{std::move(selected_functions)}]() {
+        Capture(pid, selected_functions);
+      });
 
   return outcome::success();
 }
 
 void CaptureClient::Capture(
     int32_t pid,
-    const absl::flat_hash_map<uint64_t, FunctionInfo>& selected_functions) {
+    absl::flat_hash_map<uint64_t, FunctionInfo> selected_functions) {
   CHECK(reader_writer_ == nullptr);
 
   event_processor_.emplace(capture_listener_);
--- a/OrbitCaptureClient/CaptureClient.cpp
+++ b/OrbitCaptureClient/CaptureClient.cpp
@@ -37,16 +37,14 @@ ErrorMessageOr<void> CaptureClient::StartCapture(
 
   state_ = State::kStarting;
   thread_pool->Schedule(
-      [this, pid, selected_functions{std::move(selected_functions)}]() {
-        Capture(pid, selected_functions);
-      });
+      [this, pid, selected_functions]() { Capture(pid, selected_functions); });
 
   return outcome::success();
 }
 
 void CaptureClient::Capture(
     int32_t pid,
-    absl::flat_hash_map<uint64_t, FunctionInfo> selected_functions) {
+    const absl::flat_hash_map<uint64_t, FunctionInfo>& selected_functions) {
   CHECK(reader_writer_ == nullptr);
 
   event_processor_.emplace(capture_listener_);
--- a/OrbitClientGgp/ClientGgp.cpp
+++ b/OrbitClientGgp/ClientGgp.cpp
@@ -55,17 +55,20 @@ bool ClientGgp::InitClient() {
 
 // Client requests to start the capture
 bool ClientGgp::RequestStartCapture(ThreadPool* thread_pool) {
-  ErrorMessageOr<void> result = Capture::StartCapture();
-  if (result.has_error()) {
-    ERROR("Error starting capture: %s", result.error().message());
-    return false;
-  }
   int32_t pid = target_process_->GetID();
   LOG("Capture pid %d", pid);
 
   // TODO: selected_functions available when UploadSymbols is included
-  std::map<uint64_t, orbit_client_protos::FunctionInfo*> selected_functions =
-      Capture::GSelectedFunctionsMap;
+  // TODO(kuebler): function selection should be handled separately by each
+  //  client.
+  absl::flat_hash_map<uint64_t, orbit_client_protos::FunctionInfo>
+      selected_functions;
+  ErrorMessageOr<void> result = Capture::StartCapture(
+      pid, target_process_->GetName(), selected_functions);
+  if (result.has_error()) {
+    ERROR("Error starting capture: %s", result.error().message());
+    return false;
+  }
   result = capture_client_->StartCapture(thread_pool, pid, selected_functions);
   if (result.has_error()) {
     ERROR("Error starting capture: %s", result.error().message());
--- a/OrbitClientGgp/ClientGgp.cpp
+++ b/OrbitClientGgp/ClientGgp.cpp
@@ -59,8 +59,8 @@ bool ClientGgp::RequestStartCapture(ThreadPool* thread_pool) {
   LOG("Capture pid %d", pid);
 
   // TODO: selected_functions available when UploadSymbols is included
-  // TODO(kuebler): function selection should be handled separately by each
-  //  client.
+  // TODO(kuebler): right now selected_functions is only an empty placeholder,
+  //  it needs to be filled separately in each client and then passed.
   absl::flat_hash_map<uint64_t, orbit_client_protos::FunctionInfo>
       selected_functions;
   ErrorMessageOr<void> result = Capture::StartCapture(
--- a/OrbitCore/Capture.cpp
+++ b/OrbitCore/Capture.cpp
@@ -20,8 +20,7 @@ using orbit_client_protos::PresetFile;
 using orbit_client_protos::PresetInfo;
 
 CaptureData Capture::capture_data_;
-std::map<uint64_t, FunctionInfo*> Capture::GSelectedFunctionsMap;
-std::map<uint64_t, FunctionInfo*> Capture::GVisibleFunctionsMap;
+absl::flat_hash_map<uint64_t, FunctionInfo> Capture::GVisibleFunctionsMap;
 TextBox* Capture::GSelectedTextBox = nullptr;
 ThreadID Capture::GSelectedThreadId;
 
@@ -35,21 +34,19 @@ void Capture::SetTargetProcess(std::shared_ptr<Process> process) {
   if (process != GTargetProcess) {
     GSamplingProfiler = std::make_shared<SamplingProfiler>(process);
     GTargetProcess = std::move(process);
-    GSelectedFunctionsMap.clear();
   }
 }
 
-ErrorMessageOr<void> Capture::StartCapture() {
+ErrorMessageOr<void> Capture::StartCapture(
+    uint64_t process_id, std::string process_name,
+    absl::flat_hash_map<uint64_t, FunctionInfo> selectedFunctions) {
   if (GTargetProcess->GetID() == 0) {
     return ErrorMessage(
         "No process selected. Please choose a target process for the capture.");
   }
 
-  capture_data_ =
-      CaptureData(GTargetProcess->GetID(), GTargetProcess->GetName(),
-                  GetSelectedFunctions());
-
-  PreFunctionHooks();
+  capture_data_ = CaptureData(process_id, std::move(process_name),
+                              std::move(selectedFunctions));
 
   Capture::GSamplingProfiler =
       std::make_shared<SamplingProfiler>(Capture::GTargetProcess);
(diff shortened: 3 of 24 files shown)


OrbitCore/Capture.cpp
```diff
@@ -35,21 +34,19 @@ void Capture::SetTargetProcess(std::shared_ptr<Process> process) {
...
     GSamplingProfiler = std::make_shared<SamplingProfiler>(process);
     GTargetProcess = std::move(process);
-    GSelectedFunctionsMap.clear();
   }
 }
 
-ErrorMessageOr<void> Capture::StartCapture() {
+ErrorMessageOr<void> Capture::StartCapture(
+    uint64_t process_id, std::string process_name,
+    absl::flat_hash_map<uint64_t, FunctionInfo> selectedFunctions) {
```
IrinaShkviro, line 42:
```suggestion
    absl::flat_hash_map<uint64_t, FunctionInfo> selected_functions) {
```

OrbitClientGgp/ClientGgp.cpp
```diff
@@ -55,17 +55,21 @@ bool ClientGgp::InitClient() {
...
+  absl::flat_hash_map<uint64_t, orbit_client_protos::FunctionInfo>
+      selected_functions;
+  ErrorMessageOr<void> result = Capture::StartCapture(
+      options_.capture_pid, Capture::GTargetProcess->GetName(),
+      std::move(selected_functions));
+  if (result.has_error()) {
+    ERROR("Error starting capture: %s", result.error().message());
+    return false;
+  }
   result = capture_client_->StartCapture(thread_pool, pid, selected_functions);
```
dimitry-, line 73: you have just moved them... in L 66

OrbitCaptureClient/CaptureClient.cpp
```diff
@@ -27,7 +27,7 @@ IntrumentedFunctionTypeFromOrbitType(FunctionInfo::OrbitType orbit_type) {
 
 ErrorMessageOr<void> CaptureClient::StartCapture(
     ThreadPool* thread_pool, int32_t pid,
-    const std::map<uint64_t, FunctionInfo*>& selected_functions) {
+    const absl::flat_hash_map<uint64_t, FunctionInfo>& selected_functions) {
```
dimitry-, line 30: maybe pass by value here.. we are going to make a copy of this anyways.

OrbitCore/Capture.cpp
dimitry-, line 42: nit: selected_functions here and everywhere.

OrbitCore/Pdb.h
```diff
@@ -34,7 +34,8 @@ class Pdb {
 
   void PopulateFunctionMap();
   void PopulateStringFunctionMap();
-  void ApplyPreset(const orbit_client_protos::PresetFile& preset);
+  [[nodiscard]] std::vector<orbit_client_protos::FunctionInfo*>
+  FunctionsToSelect(const orbit_client_protos::PresetFile& preset) const;
```
dimitry-, line 38: GetSelectedFunctions or GetSelectedFunctionsFromPreset

OrbitClientGgp/ClientGgp.cpp
```diff
@@ -55,17 +55,21 @@ bool ClientGgp::InitClient() {
...
   int32_t pid = Capture::GTargetProcess->GetID();
   LOG("Capture pid %d", pid);
 
   // TODO: selected_functions available when UploadSymbols is included
-  std::map<uint64_t, orbit_client_protos::FunctionInfo*> selected_functions =
-      Capture::GSelectedFunctionsMap;
+  // TODO(kuebler): function selection should be handled separately by each
+  //  client.
+  absl::flat_hash_map<uint64_t, orbit_client_protos::FunctionInfo>
+      selected_functions;
```
dpallotti, line 65: This is a bit confusing as `selected_functions` is not filled and the TODO is two lines above now.

OrbitCore/Pdb.cpp
```diff
@@ -79,8 +79,12 @@ FunctionInfo* Pdb::GetFunctionFromProgramCounter(uint64_t a_Address) {
   return it->second;
 }
 
-void Pdb::ApplyPreset(const PresetFile& preset) {
-  SCOPE_TIMER_LOG(absl::StrFormat("Pdb::ApplyPreset - %s", m_Name.c_str()));
+std::vector<FunctionInfo*> Pdb::FunctionsToSelect(
+    const PresetFile& preset) const {
+  SCOPE_TIMER_LOG(
```
dpallotti, line 84: I would just remove this, it doesn't seem that relevant to time this.

OrbitGl/App.cpp
```diff
@@ -466,12 +467,49 @@ void OrbitApp::SetClipboard(const std::string& text) {
...
+
+  std::string filename_with_ext = filename;
+  if (!absl::EndsWith(filename, ".opr")) {
+    filename_with_ext += ".opr";
+  }
+
+  std::ofstream file(filename_with_ext, std::ios::binary);
+  if (file.fail()) {
+    ERROR("Saving preset in \"%s\": %s", filename_with_ext, "file.fail()");
+    return ErrorMessage("Error opening the file for writing");
```
dpallotti, line 501: Shouldn't the file name also be included here?

OrbitGl/App.cpp
```diff
@@ -466,12 +467,49 @@ void OrbitApp::SetClipboard(const std::string& text) {
...
+  }
+
+  {
+    SCOPE_TIMER_LOG(
```
dpallotti, line 505: Maybe we want to log this, but probably just normally, I don't think we need to time it.

OrbitGl/App.cpp
```diff
@@ -878,6 +936,31 @@ std::shared_ptr<Process> OrbitApp::FindProcessByPid(int32_t pid) {
...
 
+void OrbitApp::SelectFunction(const orbit_client_protos::FunctionInfo& func) {
+  uint64_t absolute_address = FunctionUtils::GetAbsoluteAddress(func);
+  LOG("Selected %s at 0x%" PRIx64 " (address_=0x%" PRIx64
+      ", load_bias_= 0x%" PRIx64 ", base_address=0x%" PRIx64 ")",
+      func.pretty_name(), absolute_address, func.address(), func.load_bias(),
+      func.module_base_address());
+  data_manager_->SelectFunction(absolute_address);
+}
+void OrbitApp::UnSelectFunction(const orbit_client_protos::FunctionInfo& func) {
```
dpallotti, line 947: Add a white line between these methods?

OrbitCore/Capture.cpp
```diff
@@ -35,21 +34,19 @@ void Capture::SetTargetProcess(std::shared_ptr<Process> process) {
...
   if (GTargetProcess->GetID() == 0) {
     return ErrorMessage(
         "No process selected. Please choose a target process for the capture.");
   }
 
-  capture_data_ =
-      CaptureData(GTargetProcess->GetID(), GTargetProcess->GetName(),
```
kisenshi, line 49: Why are you removing the use of GTargetProcess and providing it as a parameter of StartCapture? If GTargetProcess is being deprecated I think it should be removed entirely then. I think it makes confusing the flow of how a process is targeted. And as result, SetTargetProcess should dissapear as well

OrbitCore/Capture.h
```diff
@@ -25,23 +25,20 @@ class Capture {
  public:
   static void Init();
   static void SetTargetProcess(std::shared_ptr<Process> process);
```
kisenshi, line 27: Related to my comment to removing the use of GTargetProcess in Capture.cpp, I think this method should dissappear as well if that is being deprecated

OrbitGl/App.cpp
dpallotti, line 947: I would use `Unselect` (or honestly `Deselect`) instead of `UnSelect`, it's a single word. Here and for the other "unselect" methods in the pull request.

OrbitGl/LiveFunctionsDataView.cpp
```diff
@@ -295,21 +293,23 @@ void LiveFunctionsDataView::DoFilter() {
...
       selected_functions = Capture::capture_data_.selected_functions();
   size_t functions_count = selected_functions.size();
   indices_.resize(functions_count);
-  for (size_t i = 0; i < functions_count; ++i) {
+  size_t i = 0;
+  for (const auto& pair : selected_functions) {
+    functions_.push_back(pair.second);
     indices_[i] = i;
-    functions_.push_back(selected_functions[i]);
+    i++;
```
dpallotti, line 312: More for fun than anything, it's finally time to link this new addition to the styleguide: https://google.github.io/styleguide/cppguide.html#Preincrement_and_Predecrement 😄

OrbitClientGgp/ClientGgp.cpp
```diff
@@ -55,17 +55,21 @@ bool ClientGgp::InitClient() {
...
+  ErrorMessageOr<void> result = Capture::StartCapture(
+      options_.capture_pid, Capture::GTargetProcess->GetName(),
```
kisenshi, line 67:
```suggestion
      pid, Capture::GTargetProcess->GetName(),
```

kisenshi: Options is used in the InitCapture() method to set the process, so here it should be using the pid that is retrieved from that process created and not the value in options

OrbitCore/Capture.cpp
kisenshi: As a clarification/extension of the previous comment: If GTargetProcess is not being removed, I think this should not be updated then because as I understand it, it is just including a extra step: clients use SetGTargetProcess to set the target process, then the client gets the GTargetProcess data again and provides it to Capture, which has the GTargetProcess information already. I think that if we want to provide the target process when StartCapture is called directlly, GTargetProcess is duplicating unnecessary code so it should be removed and handled in the client entirely. If I am not understanding the flow please let me know.

florian-kuebler: GTargetProcess will be removed as a next step. And as we have all information required here also available, I decided to already pass the information here.

OrbitCore/Capture.h
florian-kuebler: Yes, it will be removed in a separate CL. Currently, I still maintain the global, but removed a couple of reads from it.

OrbitClientGgp/ClientGgp.cpp
florian-kuebler: I would be fine, if you store the pid/process in the client, but don't want to use the GTargetProcess here.

OrbitCore/Capture.cpp
kisenshi: I think it should be done in the next step then, to have everything in a same PR

OrbitClientGgp/ClientGgp.cpp
florian-kuebler: Thanks, removed the move

OrbitClientGgp/ClientGgp.cpp
florian-kuebler: The client does not yet handle selecting functions, thus I created an empty object as a placeholder here.
However, I updated the comment to make it more clear.

OrbitClientGgp/ClientGgp.cpp
kisenshi: I agree, I can update the code to handle the process locally but right not it is using anyway by using Capture::GTargetProcess->GetName(). That's why I think it makes sense to not update the StartCapture in this step to provide the Process and do that in a separate PR

florian-kuebler: Anyways, will go for GTargetProcess for now, as it is used anyways.

OrbitCore/Capture.cpp
florian-kuebler: Thanks, I changed it in the last commit.

florian-kuebler: done. Thanks.

OrbitClientGgp/ClientGgp.cpp
kisenshi: I can work to change the ggp client and create a local target process so the other change is easier for you if you remove GTargetProcess in a next step

OrbitCore/Capture.cpp
florian-kuebler: I kind of reverted some of the changes. However, as I needed to change the constructor anyways and already have that information available on the caller side, I would prefer to keep this change here.

OrbitCore/Pdb.cpp
florian-kuebler: done.

OrbitCore/Pdb.h
florian-kuebler: Ah was thinking about the name a lot. thanks for that suggestion. I went with the latter.

OrbitGl/App.cpp
florian-kuebler: Done.

OrbitGl/App.cpp
florian-kuebler: Done.

OrbitGl/App.cpp
florian-kuebler: Done. Thanks.

florian-kuebler: changed to Deselect.

OrbitGl/LiveFunctionsDataView.cpp
florian-kuebler: Well I kind off disagree with the style guide, but well fine to change this :D

OrbitGl/TextBox.cpp
```diff
@@ -113,8 +114,10 @@ void TextBox::Draw(Batcher* batcher, TextRenderer& a_TextRenderer, float a_MinX,
 
     float maxSize = m_Pos[0] + m_Size[0] - posX;
 
-    FunctionInfo* func =
-        Capture::GSelectedFunctionsMap[timer_info_.function_address()];
+    const FunctionInfo* func = Capture::capture_data_.GetSelectedFunction(
+        timer_info_.function_address());
+    CHECK(func != nullptr);
+    CHECK(func != nullptr);
```
antonrohr, line 120:
```suggestion
```

OrbitGl/ThreadTrack.cpp
```diff
@@ -126,8 +129,9 @@ void ThreadTrack::SetTimesliceText(const TimerInfo& timer_info,
   TimeGraphLayout layout = time_graph_->GetLayout();
   if (text_box->GetText().empty()) {
     std::string time = GetPrettyTime(absl::Microseconds(elapsed_us));
-    FunctionInfo* func =
-        Capture::GSelectedFunctionsMap[timer_info.function_address()];
+    const FunctionInfo* func = Capture::capture_data_.GetSelectedFunction(
+        timer_info.function_address());
+    CHECK(func != nullptr);
```
antonrohr, line 134: Two lines after this CHECK there is `if(func)`.

OrbitCaptureClient/CaptureClient.cpp
florian-kuebler: Done, now also move this into the lambda below.

OrbitGl/App.cpp
```diff
@@ -466,12 +467,49 @@ void OrbitApp::SetClipboard(const std::string& text) {
...
+  }
+
+  std::ofstream file(filename_with_ext, std::ios::binary);
+  if (file.fail()) {
+    ERROR("Saving preset in \"%s\": %s", filename_with_ext, "file.fail()");
+    return ErrorMessage(absl::StrFormat("Error opening the file %s for writing",
+                                        filename_with_ext));
+  }
+
+  {
```
dpallotti, line 505: You can remove this scope.

OrbitGl/App.cpp
```diff
@@ -466,12 +467,49 @@ void OrbitApp::SetClipboard(const std::string& text) {
...
+
+  std::string filename_with_ext = filename;
+  if (!absl::EndsWith(filename, ".opr")) {
+    filename_with_ext += ".opr";
+  }
+
+  std::ofstream file(filename_with_ext, std::ios::binary);
+  if (file.fail()) {
+    ERROR("Saving preset in \"%s\": %s", filename_with_ext, "file.fail()");
+    return ErrorMessage(absl::StrFormat("Error opening the file %s for writing",
```
dpallotti, line 501:
```suggestion
    return ErrorMessage(absl::StrFormat("Error opening the file \"%s\" for writing",
```

OrbitGl/App.cpp
```diff
@@ -572,6 +611,18 @@ bool OrbitApp::StartCapture() {
   return true;
 }
 
+absl::flat_hash_map<uint64_t, FunctionInfo>
+OrbitApp::GetSelectedFunctionsAndOrbitFunctions() const {
+  absl::flat_hash_map<uint64_t, FunctionInfo> selected_functions;
+  for (const auto& func : Capture::GTargetProcess->GetFunctions()) {
+    if (IsFunctionSelected(*func) || FunctionUtils::IsOrbitFunc(*func)) {
+      uint64_t address = FunctionUtils::GetAbsoluteAddress(*func);
+      selected_functions[address] = *(func.get());
```
dpallotti, line 620: Do you need the `.get()`?

OrbitClientGgp/ClientGgp.cpp
florian-kuebler: Cool, thanks!

kisenshi: Ok I have merged the change, so now you can use target_process_ instead :)

OrbitCore/Capture.cpp
kisenshi: Ok, I understand, thanks!

OrbitGl/ThreadTrack.cpp
florian-kuebler: Thanks. I removed the check here, as it seems that the introspected functions can also occur here.

OrbitGl/TextBox.cpp
florian-kuebler: Thanks. Removed it in the last commit.

OrbitGl/App.cpp
florian-kuebler: Nope, you are right. Thanks.

OrbitGl/App.cpp
florian-kuebler: Right.

OrbitGl/App.cpp
florian-kuebler: Changed it in the last commit. Thanks.

florian-kuebler:
Rebased to master.
@antonrohr, @dimitry- , @IrinaShkviro anything else I need to look at?

OrbitCaptureClient/CaptureClient.cpp
```diff
@@ -37,13 +37,16 @@ ErrorMessageOr<void> CaptureClient::StartCapture(
...
+        Capture(pid, selected_functions);
+      });
 
   return outcome::success();
 }
 
 void CaptureClient::Capture(
-    int32_t pid, const std::map<uint64_t, FunctionInfo*>& selected_functions) {
+    int32_t pid,
+    absl::flat_hash_map<uint64_t, FunctionInfo> selected_functions) {
```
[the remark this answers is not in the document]
florian-kuebler: changed it back here

PR #1027, opened 2020-08-14
Enforce consistent formatting across different versions of clang-format

florian-kuebler:
Each version of clang-format has its own definition of Google style.
Recent fails in the presubmit check showed that there are differences
in the latest version and the one from the CI (which is clang-format-9).

This PR dumps the Google style definition from clang-format-9 into
the .clang-format file instead of relying on what the version below
has defined as such.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,8 +1,183 @@
 ---
 Language:        Cpp
-BasedOnStyle:  Google
-DerivePointerAlignment: false
----
-Language:        Proto
-BasedOnStyle:  Google
+# BasedOnStyle:  Google
+AccessModifierOffset: -1
+AlignAfterOpenBracket: Align
+AlignConsecutiveMacros: false
+AlignConsecutiveAssignments: false
+AlignConsecutiveBitFields: false
+AlignConsecutiveDeclarations: false
+AlignEscapedNewlines: Left
+AlignOperands:   Align
+AlignTrailingComments: true
+AllowAllArgumentsOnNextLine: true
+AllowAllConstructorInitializersOnNextLine: true
+AllowAllParametersOfDeclarationOnNextLine: true
+AllowShortEnumsOnASingleLine: true
+AllowShortBlocksOnASingleLine: Never
+AllowShortCaseLabelsOnASingleLine: false
+AllowShortFunctionsOnASingleLine: All
+AllowShortLambdasOnASingleLine: All
+AllowShortIfStatementsOnASingleLine: WithoutElse
+AllowShortLoopsOnASingleLine: true
+AlwaysBreakAfterDefinitionReturnType: None
+AlwaysBreakAfterReturnType: None
+AlwaysBreakBeforeMultilineStrings: true
+AlwaysBreakTemplateDeclarations: Yes
+BinPackArguments: true
+BinPackParameters: true
+BraceWrapping:
+  AfterCaseLabel:  false
+  AfterClass:      false
+  AfterControlStatement: Never
+  AfterEnum:       false
+  AfterFunction:   false
+  AfterNamespace:  false
+  AfterObjCDeclaration: false
+  AfterStruct:     false
+  AfterUnion:      false
+  AfterExternBlock: false
+  BeforeCatch:     false
+  BeforeElse:      false
+  BeforeLambdaBody: false
+  BeforeWhile:     false
+  IndentBraces:    false
+  SplitEmptyFunction: true
+  SplitEmptyRecord: true
+  SplitEmptyNamespace: true
+BreakBeforeBinaryOperators: None
+BreakBeforeBraces: Attach
+BreakBeforeInheritanceComma: false
+BreakInheritanceList: BeforeColon
+BreakBeforeTernaryOperators: true
+BreakConstructorInitializersBeforeComma: false
+BreakConstructorInitializers: BeforeColon
+BreakAfterJavaFieldAnnotations: false
+BreakStringLiterals: true
+ColumnLimit:     80
+CommentPragmas:  '^ IWYU pragma:'
+CompactNamespaces: false
+ConstructorInitializerAllOnOneLineOrOnePerLine: true
+ConstructorInitializerIndentWidth: 4
+ContinuationIndentWidth: 4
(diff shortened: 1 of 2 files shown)


florian-kuebler:
Seem that the presubmit check we have currently is failing for #1026.

florian-kuebler:
As discussed offline, the root cause is not ignoring the file but differences in the Google style in different versions.

beckerhe:
> As discussed offline, the root cause is not ignoring the file but differences in the Google style in different versions.

When I take this file and try to format some code with clang-format-7 I get the following error:
```
YAML:11:18: error: invalid boolean
AlignOperands:   Align
                 ^~~~~
```

It seems the config was generated with a more modern version of clang-format. We shouldn't impose that requirement on our potential external contributors. The Google-style between LLVM-7 and LLVM-9 didn't change. It's just the naming in the configuration file. So there are no disadvantages for us using a config generated by LLVM-7. (Sorry for the confusion, I think we discussed that in chat but I failed to make myself clear.)

We should also loop in @pierricgimmig  and @dimitry-. If I remember correctly they both had objections regarding an explicit configuration.

florian-kuebler:
I'd suggest to go then with a config file generated with clang-format-7.

florian-kuebler:
> I think we should go with the short-version and have final arbitrator on the presubmit check side. (we also want it to be the lasters stable version of clang-format)

This is somewhat inconvenient and arbitrary. Locally, you then need to know which version is used at the check side. E.g. you can't rely on your IDE or even pre-commit hooks.

dimitry-:
> > I think we should go with the short-version and have final arbitrator on the presubmit check side. (we also want it to be the lasters stable version of clang-format)
> 
> This is somewhat inconvenient and arbitrary. Locally, you then need to know which version is used at the check side. E.g. you can't rely on your IDE or even pre-commit hooks.

Don't tools use whichever clang-format you have installed on your machine? If this is the case the idea would be to always use the version used by CI.

florian-kuebler:
> > > I think we should go with the short-version and have final arbitrator on the presubmit check side. (we also want it to be the lasters stable version of clang-format)
> > 
> > 
> > This is somewhat inconvenient and arbitrary. Locally, you then need to know which version is used at the check side. E.g. you can't rely on your IDE or even pre-commit hooks.
> 
> Don't tools use whichever clang-format you have installed on your machine? If this is the case the idea would be to always use the version used by CI.

Exactly, tools like IDEs usually use whats on your system. So if you have a newer version then on the kokoro machine (which is true for our workstations), you result in different formattings. The idea of this change would be, that you can use what ever version you want and will result in what is on the kokoro machine.

dimitry-:
> > > > I think we should go with the short-version and have final arbitrator on the presubmit check side. (we also want it to be the lasters stable version of clang-format)
> > > 
> > > 
> > > This is somewhat inconvenient and arbitrary. Locally, you then need to know which version is used at the check side. E.g. you can't rely on your IDE or even pre-commit hooks.
> > 
> > 
> > Don't tools use whichever clang-format you have installed on your machine? If this is the case the idea would be to always use the version used by CI.
> 
> Exactly, tools like IDEs usually use whats on your system. So if you have a newer version then on the kokoro machine (which is true for our workstations), you result in different formattings. The idea of this change would be, that you can use what ever version you want and will result in what is on the kokoro machine.

But this is not the case right? there are bugs which are fixed in later version which could result in different formatting even if you expand everything. I do not think this is a solution. It seems to me it introduces more problems then it solves. It is hard to read, hard to maintain, if are still capped by whichever version is used to generate this file.

beckerhe:
> > > > > I think we should go with the short-version and have final arbitrator on the presubmit check side. (we also want it to be the lasters stable version of clang-format)
> > > > 
> > > > 
> > > > This is somewhat inconvenient and arbitrary. Locally, you then need to know which version is used at the check side. E.g. you can't rely on your IDE or even pre-commit hooks.
> > > 
> > > 
> > > Don't tools use whichever clang-format you have installed on your machine? If this is the case the idea would be to always use the version used by CI.
> > 
> > 
> > Exactly, tools like IDEs usually use whats on your system. So if you have a newer version then on the kokoro machine (which is true for our workstations), you result in different formattings. The idea of this change would be, that you can use what ever version you want and will result in what is on the kokoro machine.
> 
> But this is not the case right? there are bugs which are fixed in later version which could result in different formatting even if you expand everything. I do not think this is a solution. It seems to me it introduces more problems then it solves. It is hard to read, hard to maintain, if are still capped by whichever version is used to generate this file.

As far as I know bugs in clang-format are always fixed by introducing an additional option which is not enabled by default. So having a verbose config will fix that because bugfixes won't be enabled by default.

florian-kuebler:
> > > > > > I think we should go with the short-version and have final arbitrator on the presubmit check side. (we also want it to be the lasters stable version of clang-format)
> > > > > 
> > > > > 
> > > > > This is somewhat inconvenient and arbitrary. Locally, you then need to know which version is used at the check side. E.g. you can't rely on your IDE or even pre-commit hooks.
> > > > 
> > > > 
> > > > Don't tools use whichever clang-format you have installed on your machine? If this is the case the idea would be to always use the version used by CI.
> > > 
> > > 
> > > Exactly, tools like IDEs usually use whats on your system. So if you have a newer version then on the kokoro machine (which is true for our workstations), you result in different formattings. The idea of this change would be, that you can use what ever version you want and will result in what is on the kokoro machine.
> > 
> > 
> > But this is not the case right? there are bugs which are fixed in later version which could result in different formatting even if you expand everything. I do not think this is a solution. It seems to me it introduces more problems then it solves. It is hard to read, hard to maintain, if are still capped by whichever version is used to generate this file.
> 
> As far as I know bugs in clang-format are always fixed by introducing an additional option which is not enabled by default. So having a verbose config will fix that because bugfixes won't be enabled by default.

And we already had it twice, that system was different to kokoro.

dimitry-:
I am still convinced we shouldn't fix the version of Google style. We probably want to move forward with it.

florian-kuebler:
As discussed offline, source of truth is the latest version on the CI and we will add this version to the documentation.

PR #1028, opened 2020-08-14
Unified build script for Kokoro

beckerhe:
This change adds a single build script which can be used for building all the different build configurations.

All features from the various existing build script have been included in this one. Some examples:

- Windows and Linux support
- Symbol uploading
- Signing
- Package uploading
- Zipping

The script requires docker support - also on Windows. That's why it's not yet enabled for the Windows-build. It will be enabled when we switch out the job-configuration there. The future not-yet-active build configuration can be observed in `kokoro/builds/windows/msvc2017_relwithdebinfo`.


beckerhe:
@IrinaShkviro and @danielfenner This is probably something that makes sense to talk through. Ping me if you want some explanation!

kokoro/builds/build.sh
```diff
@@ -0,0 +1,203 @@
...
+    echo -n "Maybe you missed pushing the release version tag?"
+    echo    " Please consult the release playbook for advice."
+    exit 1
+  fi
+  set -e
+
+  # That's a temporary solution. The docker containers should have the
+  # correct version of conan already preinstalled. This step will be removed
+  # when the docker containers are restructured and versioned.
+  pip3 install conan==1.27.1
```
IrinaShkviro, line 61: Should we also add `conan_package_tools==0.34.0`?

beckerhe: Well, we don't need that at all currently. I want to remove that entirely at some point. And since that's just a hack I didn't bother adding it here. It would just increase the duration of the CI run (not by much though).

kokoro/builds/linux/clang7_relwithdebinfo/continuous.cfg
```diff
@@ -0,0 +1,25 @@
...
+      keyname: "orbitprofiler_artifactory_access_token"
+      key_type: RAW
+      backend_type: FASTCONFIGPUSH
+    }
+    keystore_resource {
+      keystore_config_id: 74938
+      keyname: "orbitprofiler_crashdump_collection_server"
+      key_type: RAW
+      backend_type: FASTCONFIGPUSH
+    }
```
IrinaShkviro, line 16: I think that it should be added for `release.cfg` and removed from here.

kokoro/builds/build.sh
IrinaShkviro: ok, thanks

kokoro/builds/linux/clang7_relwithdebinfo/continuous.cfg
beckerhe: Good point. I just copied that from `kokoro/gcp_ubuntu/continuous.cfg`, but it should probably also not be there.
The `release.cfg` should also not exist for `clang7_relwithdebinfo` (No Linux-UI release). I had this for testing and I'm going to remove it.

beckerhe: done.

IrinaShkviro: Is the same applicable for `gcc9_relwithdebingo`?

beckerhe: Yes. thanks.

beckerhe: done.

PR #1030, opened 2020-08-14
Include local target process in ClientGgp

kisenshi:
GTargetProcess in Capture is going to dissapear so this code prepares the ggp client to handle the target process itself.


OrbitClientGgp/ClientGgp.h
```diff
@@ -38,6 +38,7 @@ class ClientGgp final : public CaptureListener {
  private:
   ClientGgpOptions options_;
   std::shared_ptr<grpc::Channel> grpc_channel_;
+  std::shared_ptr<Process> target_process;
```
florian-kuebler, line 41: please use an `_` at the end of member variables as described in the [style guide](https://google.github.io/styleguide/cppguide.html#Variable_Names).

PR #1031, opened 2020-08-14
Multi-pass capabilities for the batcher

reichlfl:
Multiple calls to Batcher.Reset() in GlCanvas caused problems with 
accessing user_data that has been rendered in a previous pass.
This is now fixed as follows:

Each batcher stores it's user_data now independent of the actual
rendering primitives. Reset() accepts an additional parameter to
indicate that a new frame has started.

This way, a batcher can perform rendering multiple times during a frame
without loosing the user_data of primitives rendered in a previous
renderpass. Only one Reset(true) must be called each frame.

Added tests to reproduce the previous behavior and check for correct
behavior. Also added checks to the PickingId to discover potential overflows.

### DIFF
--- a/OrbitGl/Batcher.cpp
+++ b/OrbitGl/Batcher.cpp
@@ -9,8 +9,7 @@
 
 void Batcher::AddLine(Vec2 from, Vec2 to, float z, const Color& color,
                       std::unique_ptr<PickingUserData> user_data) {
-  Color picking_color =
-      PickingId::ToColor(PickingType::kLine, user_data_.size(), batcher_id_);
+  Color picking_color = PickingId::ToColor(PickingType::kLine, user_data_.size(), batcher_id_);
 
   AddLine(from, to, z, color, picking_color, std::move(user_data));
 }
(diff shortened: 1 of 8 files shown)


OrbitGl/Batcher.cpp
```diff
@@ -163,10 +159,14 @@ void Batcher::GetBoxGradientColors(const Color& color,
   (*colors)[3] = color;
 }
 
-void Batcher::Reset() {
+void Batcher::Reset(bool start_new_frame) {
```
pierricgimmig, line 162: Hi Florian, could this be split into 2 functions, Reset() and ResetAndStartNewFrame(), or Reset() and StartNewFrame()? It would be clearer I think. Passing a boolean as argument is not ideal, another option would be to pass an enum.

reichlfl:
> Hi Florian, please see small comment, otherwise the code LGTM. However, conceptually, I'm wondering, is it necessary to render in multiple passes when picking? Could we render in a single pass instead?

I added this because the current implementation causes a crash on hovering over certain elements. The root cause is in `GlCanvas::Render()`: The `ui_batcher_` gets Reset in between because of the change in coordinate system, and there's pickable elements drawn in both passes.

OrbitGl/Batcher.cpp
reichlfl: I've split it into ResetElements() and StartNewFrame() to make this clearer, thanks for pointing this out.

PR #1033, opened 2020-08-14
Move callstacks from SamplingProfiler to CaptureData

IrinaShkviro:
Callstacks are the "raw" data that we get during the capture, so it
should be part of CaptureData. After this change SamplingProfiler
stores only post-processed data.

### DIFF
--- a/OrbitCore/CallstackData.cpp
+++ b/OrbitCore/CallstackData.cpp
@@ -1,0 +1,38 @@
+// Copyright (c) 2020 The Orbit Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#include "CallstackData.h"
+
+#include "Callstack.h"
+#include "Capture.h"
+#include "SamplingProfiler.h"
+
+using orbit_client_protos::CallstackEvent;
+
+void CallstackData::AddCallstackEvent(CallstackEvent callstack_event) {
+  CallstackID hash = callstack_event.callstack_hash();
+  CHECK(HasCallStack(hash));
+  callstack_events_.push_back(std::move(callstack_event));
+}
+
+void CallstackData::AddUniqueCallStack(CallStack call_stack) {
+  absl::MutexLock lock(&unique_callstacks_mutex_);
+  CallstackID hash = call_stack.GetHash();
+  unique_callstacks_[hash] = std::make_shared<CallStack>(std::move(call_stack));
+}
+
+void CallstackData::AddCallStackFromKnownCallstackData(
+    const orbit_client_protos::CallstackEvent& event, const CallstackData& known_callstack_data) {
+  uint64_t hash = event.callstack_hash();
+  std::shared_ptr<CallStack> unique_callstack = known_callstack_data.GetCallstackPtr(hash);
+  if (unique_callstack == nullptr) {
+    return;
+  }
+
+  if (!HasCallStack(hash)) {
+    absl::MutexLock lock(&unique_callstacks_mutex_);
+    unique_callstacks_[hash] = std::move(unique_callstack);
+  }
+  callstack_events_.push_back(CallstackEvent(event));
+}
--- a/OrbitCore/CallstackData.h
+++ b/OrbitCore/CallstackData.h
@@ -1,0 +1,81 @@
+// Copyright (c) 2020 The Orbit Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#ifndef ORBIT_CORE_CALLSTACK_DATA_H_
+#define ORBIT_CORE_CALLSTACK_DATA_H_
+
+#include <memory>
+
+#include "BlockChain.h"
+#include "Callstack.h"
+#include "CallstackTypes.h"
+#include "absl/container/flat_hash_map.h"
+#include "absl/synchronization/mutex.h"
+#include "capture_data.pb.h"
+
+class CallstackData {
+ public:
+  explicit CallstackData() = default;
+
+  // Assume that callstack_event.callstack_hash is filled correctly and the
+  // CallStack with corresponding hash is already in unique_callstacks_
+  void AddCallstackEvent(orbit_client_protos::CallstackEvent callstack_event);
+  void AddUniqueCallStack(CallStack call_stack);
+  void AddCallStackFromKnownCallstackData(const orbit_client_protos::CallstackEvent& event,
+                                          const CallstackData& known_callstack_data);
+
+  [[nodiscard]] const BlockChain<orbit_client_protos::CallstackEvent, 16 * 1024>& callstack_events()
+      const {
+    return callstack_events_;
+  };
+
+  [[nodiscard]] uint32_t GetCallstackEventsSize() const { return callstack_events_.size(); };
+
+  [[nodiscard]] const CallStack* GetCallStack(CallstackID callstack_id) const {
+    absl::MutexLock lock(&unique_callstacks_mutex_);
+    auto it = unique_callstacks_.find(callstack_id);
+    if (it != unique_callstacks_.end()) {
+      return unique_callstacks_.at(callstack_id).get();
+    }
+    return nullptr;
+  };
+
+  [[nodiscard]] bool HasCallStack(CallstackID callstack_id) const {
+    absl::MutexLock lock(&unique_callstacks_mutex_);
+    return unique_callstacks_.count(callstack_id) > 0;
+  }
+
+  void ForEachUniqueCallstack(const std::function<void(const CallStack&)>& action) const {
+    absl::MutexLock lock(&unique_callstacks_mutex_);
+    for (const auto& it : unique_callstacks_) {
+      action(*it.second);
+    }
+  }
+
+  void ForEachFrameInCallstack(uint64_t callstack_id,
+                               const std::function<void(uint64_t)>& action) const {
+    absl::MutexLock lock(&unique_callstacks_mutex_);
+    for (uint64_t frame : unique_callstacks_.at(callstack_id)->GetFrames()) {
+      action(frame);
+    }
+  }
+
+ private:
+  [[nodiscard]] std::shared_ptr<CallStack> GetCallstackPtr(CallstackID callstack_id) const {
+    absl::MutexLock lock(&unique_callstacks_mutex_);
+    auto it = unique_callstacks_.find(callstack_id);
+    if (it != unique_callstacks_.end()) {
+      return unique_callstacks_.at(callstack_id);
+    }
+    return nullptr;
+  };
+
+ private:
+  BlockChain<orbit_client_protos::CallstackEvent, 16 * 1024> callstack_events_;
+
+  mutable absl::Mutex unique_callstacks_mutex_;
+  absl::flat_hash_map<CallstackID, std::shared_ptr<CallStack>> unique_callstacks_;
+};
+
+#endif  // ORBIT_CORE_CALLSTACK_DATA_H_
--- a/OrbitCore/CaptureData.h
+++ b/OrbitCore/CaptureData.h
@@ -8,6 +8,7 @@
 #include <memory>
 #include <vector>
 
+#include "CallstackData.h"
 #include "absl/container/flat_hash_map.h"
 #include "capture_data.pb.h"
 
@@ -18,7 +19,9 @@ class CaptureData {
       absl::flat_hash_map<uint64_t, orbit_client_protos::FunctionInfo> selected_functions)
       : process_id_{process_id},
         process_name_{std::move(process_name)},
-        selected_functions_{std::move(selected_functions)} {}
+        selected_functions_{std::move(selected_functions)},
+        callstack_data_(std::make_unique<CallstackData>()),
+        selection_callstack_data_(std::make_unique<CallstackData>()) {}
   explicit CaptureData(
       int32_t process_id, std::string process_name,
       absl::flat_hash_map<uint64_t, orbit_client_protos::FunctionInfo> selected_functions,
@@ -26,9 +29,13 @@ class CaptureData {
       : process_id_{process_id},
         process_name_{std::move(process_name)},
         selected_functions_{std::move(selected_functions)},
-        functions_stats_{std::move(functions_stats)} {}
+        functions_stats_{std::move(functions_stats)},
+        callstack_data_(std::make_unique<CallstackData>()),
+        selection_callstack_data_(std::make_unique<CallstackData>()) {}
 
-  explicit CaptureData() = default;
+  explicit CaptureData()
+      : callstack_data_(std::make_unique<CallstackData>()),
+        selection_callstack_data_(std::make_unique<CallstackData>()){};
   CaptureData(const CaptureData& other) = default;
   CaptureData& operator=(const CaptureData& other) = default;
   CaptureData(CaptureData&& other) = default;
@@ -89,6 +96,32 @@ class CaptureData {
   void UpdateFunctionStats(orbit_client_protos::FunctionInfo* func,
                            const orbit_client_protos::TimerInfo& timer_info);
 
+  [[nodiscard]] const CallstackData* GetCallstackData() const { return callstack_data_.get(); };
+
+  void AddUniqueCallStack(CallStack call_stack) {
+    callstack_data_->AddUniqueCallStack(std::move(call_stack));
+  }
+
+  void AddCallstackEvent(orbit_client_protos::CallstackEvent callstack_event) {
+    callstack_data_->AddCallstackEvent(std::move(callstack_event));
+  }
+
+  [[nodiscard]] const CallstackData* GetSelectionCallstackData() const {
+    return selection_callstack_data_.get();
+  };
+
+  void AddSelectionCallStack(CallStack call_stack) {
+    selection_callstack_data_->AddUniqueCallStack(std::move(call_stack));
+  }
+
+  void AddSelectionCallstackEvent(orbit_client_protos::CallstackEvent callstack_event) {
+    selection_callstack_data_->AddCallstackEvent(std::move(callstack_event));
+  }
+
+  void AddSelectionCallstackFromCallstackData(const orbit_client_protos::CallstackEvent& event) {
+    selection_callstack_data_->AddCallStackFromKnownCallstackData(event, *callstack_data_);
+  }
+
  private:
   absl::flat_hash_map<uint64_t, orbit_client_protos::LinuxAddressInfo> address_infos_;
   int32_t process_id_ = -1;
@@ -100,6 +133,12 @@ class CaptureData {
   std::chrono::system_clock::time_point capture_start_time_ = std::chrono::system_clock::now();
 
   absl::flat_hash_map<uint64_t, orbit_client_protos::FunctionStats> functions_stats_;
+
+  // std::unique_ptr<> allows to move and copy CallstackData easier
+  // (as CallstackData stores an absl::Mutex inside)
+  std::unique_ptr<CallstackData> callstack_data_;
+  // selection_callstack_data_ is subset of callstack_data_
+  std::unique_ptr<CallstackData> selection_callstack_data_;
 };
 
 #endif  // ORBIT_CORE_CAPTURE_DATA_H_
--- a/OrbitCore/SamplingProfiler.cpp
+++ b/OrbitCore/SamplingProfiler.cpp
@@ -62,22 +62,6 @@ std::multimap<int, CallstackID> SamplingProfiler::GetCallstacksFromAddress(uint6
   return SortCallstacks(thread_id_to_sample_data_[thread_id], callstacks, callstacks_count);
 }
 
-void SamplingProfiler::AddCallStack(CallstackEvent&& callstack_event) {
-  CallstackID hash = callstack_event.callstack_hash();
-  if (!HasCallStack(hash)) {
-    std::shared_ptr<CallStack> callstack = Capture::GSamplingProfiler->GetCallStack(hash);
-    AddUniqueCallStack(*callstack);
-  }
-
-  callstack_events_.push_back(std::move(callstack_event));
-}
-
-void SamplingProfiler::AddUniqueCallStack(CallStack call_stack) {
-  absl::MutexLock lock(&unique_callstacks_mutex_);
-  auto key = call_stack.GetHash();
-  unique_callstacks_[key] = std::make_shared<CallStack>(std::move(call_stack));
-}
-
 const CallStack& SamplingProfiler::GetResolvedCallstack(CallstackID raw_callstack_id) const {
   auto resolved_callstack_id_it = original_to_resolved_callstack_.find(raw_callstack_id);
   CHECK(resolved_callstack_id_it != original_to_resolved_callstack_.end());
@@ -125,7 +109,7 @@ void SamplingProfiler::SortByThreadUsage() {
        });
 }
 
-void SamplingProfiler::ProcessSamples() {
+void SamplingProfiler::ProcessSamples(const CallstackData& callstack_data) {
   // Clear the result of a previous call to ProcessSamples.
   thread_id_to_sample_data_.clear();
   unique_resolved_callstacks_.clear();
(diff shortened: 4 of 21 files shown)


OrbitGl/TimeGraph.cpp
```diff
@@ -646,14 +646,28 @@ std::vector<CallstackEvent> TimeGraph::SelectEvents(float a_WorldStart,
   sampling_profiler->SetGenerateSummary(a_TID == 0);
 
   for (const CallstackEvent& event : selected_callstack_events) {
-    if (Capture::GSamplingProfiler->GetCallStack(event.callstack_hash())) {
-      sampling_profiler->AddCallStack(CallstackEvent(event));
+    uint64_t callstack_id = event.callstack_hash();
+    if (Capture::capture_data_.GetCallstackData()->HasCallStack(callstack_id)) {
```
dimitry-, line 650: introduce a variable for GetCallstackData()/GetSelectionCallstackData() - it will be easier to read code below

OrbitCore/SamplingProfiler.cpp
```diff
@@ -132,7 +115,7 @@ void SamplingProfiler::SortByThreadUsage() {
        });
 }
 
-void SamplingProfiler::ProcessSamples() {
+void SamplingProfiler::ProcessSamples(CallstackData* callstack_data) {
```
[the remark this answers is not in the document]
IrinaShkviro: Made it const ref here and in ResolveCallstack

OrbitCore/CallstackData.h
```diff
@@ -0,0 +1,72 @@
...
+  void ForEachUniqueCallstack(
+      const std::function<void(const CallStack&)>& action) {
+    absl::MutexLock lock(&unique_callstacks_mutex_);
+    for (const auto& it : unique_callstacks_) {
+      action(*it.second);
+    }
+  }
+
+  void ForEachFrameInCallstack(uint64_t callstack_id,
+                               const std::function<void(uint64_t)>& action) {
```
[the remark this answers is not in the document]
IrinaShkviro: Done

OrbitCore/CallstackData.h
```diff
@@ -0,0 +1,72 @@
...
+    return unique_callstacks_.at(callstack_id);
+  }
+
+  bool HasCallStack(CallstackID callstack_id) {
+    absl::MutexLock lock(&unique_callstacks_mutex_);
+    return unique_callstacks_.count(callstack_id) > 0;
+  }
+
+  void ForEachUniqueCallstack(
+      const std::function<void(const CallStack&)>& action) {
```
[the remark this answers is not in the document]
IrinaShkviro: Done

OrbitCore/CallstackData.h
```diff
@@ -0,0 +1,72 @@
...
+  [[nodiscard]] uint32_t GetCallstackEventsSize() {
+    return callstack_events_.size();
+  };
+
+  std::shared_ptr<CallStack> GetCallStack(CallstackID callstack_id) {
+    absl::MutexLock lock(&unique_callstacks_mutex_);
+    return unique_callstacks_.at(callstack_id);
+  }
+
+  bool HasCallStack(CallstackID callstack_id) {
```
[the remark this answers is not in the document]
IrinaShkviro: Done

OrbitCore/CallstackData.h
```diff
@@ -0,0 +1,72 @@
...
+  void AddUniqueCallStack(std::shared_ptr<CallStack> call_stack);
+
+  [[nodiscard]] BlockChain<orbit_client_protos::CallstackEvent, 16 * 1024>&
+  GetCallstackEvents() { return callstack_events_; };
+
+  [[nodiscard]] uint32_t GetCallstackEventsSize() {
+    return callstack_events_.size();
+  };
+
+  std::shared_ptr<CallStack> GetCallStack(CallstackID callstack_id) {
```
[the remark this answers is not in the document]
IrinaShkviro: I think that it could be a const reference

OrbitCore/CallstackData.h
```diff
@@ -0,0 +1,72 @@
...
+  explicit CallstackData() = default;
+
+  void AddCallstackEvent(orbit_client_protos::CallstackEvent&& callstack_event);
+  void AddUniqueCallStack(CallStack call_stack);
+  void AddUniqueCallStack(std::shared_ptr<CallStack> call_stack);
+
+  [[nodiscard]] BlockChain<orbit_client_protos::CallstackEvent, 16 * 1024>&
+  GetCallstackEvents() { return callstack_events_; };
+
+  [[nodiscard]] uint32_t GetCallstackEventsSize() {
```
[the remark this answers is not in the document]
IrinaShkviro: Done

OrbitCore/CallstackData.h
```diff
@@ -0,0 +1,72 @@
...
+
+class CallstackData {
+ public:
+  explicit CallstackData() = default;
+
+  void AddCallstackEvent(orbit_client_protos::CallstackEvent&& callstack_event);
+  void AddUniqueCallStack(CallStack call_stack);
+  void AddUniqueCallStack(std::shared_ptr<CallStack> call_stack);
+
+  [[nodiscard]] BlockChain<orbit_client_protos::CallstackEvent, 16 * 1024>&
```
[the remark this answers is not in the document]
IrinaShkviro: Yes, fixed it

OrbitCore/CaptureData.h
```diff
@@ -103,6 +110,23 @@ class CaptureData {
...
+  [[nodiscard]] std::unique_ptr<CallstackData>& GetCallstackData() {
+    return callstack_data_;
+  }
+
+  void SetSelectionCallstackData(
+      std::unique_ptr<CallstackData> callstack_data) {
+    selection_callstack_data_ = std::move(callstack_data);
+  }
+
+  [[nodiscard]] std::unique_ptr<CallstackData>& GetSelectionCallstackData() {
```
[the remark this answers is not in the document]
IrinaShkviro: Thanks, made it a bare pointer

OrbitCore/SamplingProfiler.cpp
```diff
@@ -217,24 +200,15 @@ void SamplingProfiler::ProcessSamples() {
...
 
-void SamplingProfiler::ResolveCallstacks() {
-  absl::MutexLock lock(&unique_callstacks_mutex_);
-
-  for (const auto& it : unique_callstacks_) {
-    CallstackID raw_callstack_id = it.first;
-    const std::shared_ptr<CallStack> callstack = it.second;
-    // A "resolved callstack" is a callstack where every address is replaced by
-    // the start address of the function (if known).
+void SamplingProfiler::ResolveCallstacks(CallstackData* callstack_data) {
```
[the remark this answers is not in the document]
IrinaShkviro: done

OrbitGl/TimeGraph.cpp
IrinaShkviro: I created a method `AddCallStackFromCallstackData` to fill SelectionCallstackData and introduced variables as well

OrbitCore/CallstackData.cpp
```diff
@@ -0,0 +1,39 @@
...
+
+#include "CallstackData.h"
+
+#include "Callstack.h"
+#include "Capture.h"
+#include "SamplingProfiler.h"
+
+using orbit_client_protos::CallstackEvent;
+
+void CallstackData::AddCallstackEvent(CallstackEvent&& callstack_event) {
```
florian-kuebler, line 13: Can you add a comment at the declaration of this function, that it assumes that the callstack hash is known?

OrbitCore/CallstackData.cpp
```diff
@@ -0,0 +1,39 @@
...
+  CallstackID hash = callstack_event.callstack_hash();
+  CHECK(HasCallStack(hash));
+  callstack_events_.push_back(std::move(callstack_event));
```
florian-kuebler, line 16: Any reason why we need a lock below, but not here?

OrbitCore/CallstackData.cpp
```diff
@@ -0,0 +1,39 @@
...
+}
+
+void CallstackData::AddUniqueCallStack(CallStack call_stack) {
+  absl::MutexLock lock(&unique_callstacks_mutex_);
+  CallstackID hash = call_stack.GetHash();
+  unique_callstacks_[hash] = std::make_shared<CallStack>(std::move(call_stack));
+}
+
+void CallstackData::AddCallStackFromCallstackData(
```
florian-kuebler, line 25: I guess this is used for selections, where other is the global report? Maybe make this a bit more clear in the naming.
E.g. `other` -> `superset`/`all_callstacks`/`known_callstacks`.

OrbitCore/CallstackData.h
```diff
@@ -0,0 +1,82 @@
...
+
+  void AddCallstackEvent(orbit_client_protos::CallstackEvent&& callstack_event);
+  void AddUniqueCallStack(CallStack call_stack);
+  void AddCallStackFromCallstackData(
+      const orbit_client_protos::CallstackEvent& event,
+      const CallstackData& other);
+
+  [[nodiscard]] const BlockChain<orbit_client_protos::CallstackEvent,
+                                 16 * 1024>&
+  GetCallstackEvents() const { return callstack_events_; };
```
florian-kuebler, line 29: For simple getters, the name should be `callstack_events()`.
For setters it would be `set_callstack_events`.
https://google.github.io/styleguide/cppguide.html#Function_Names

OrbitCore/CallstackData.h
```diff
@@ -0,0 +1,82 @@
...
+
+  [[nodiscard]] uint32_t GetCallstackEventsSize() const {
+    return callstack_events_.size();
+  };
+
+  const CallStack& GetCallStack(CallstackID callstack_id) const {
+    absl::MutexLock lock(&unique_callstacks_mutex_);
+    return *unique_callstacks_.at(callstack_id);
```
florian-kuebler, line 37: This will fail, if the callstack is not available.
You should either add a comment, or maybe even better, consider using `find` and return a bare pointer or an `outcome`.

OrbitCore/CallstackData.h
```diff
@@ -0,0 +1,82 @@
...
+  }
+
+  bool HasCallStack(CallstackID callstack_id) const {
+    absl::MutexLock lock(&unique_callstacks_mutex_);
+    return unique_callstacks_.count(callstack_id) > 0;
+  }
+
+  void ForEachUniqueCallstack(
```
florian-kuebler, line 45: As I come from functional programming, I like these for each methods. However, I had a similar discussion with @dimitry- and I believe returning a const ref to `unique_callstacks_`, should be sufficient. But no need to change this from my side.

OrbitCore/CallstackData.h
```diff
@@ -0,0 +1,82 @@
...
+  void ForEachFrameInCallstack(
+      uint64_t callstack_id,
+      const std::function<void(uint64_t)>& action) const {
+    absl::MutexLock lock(&unique_callstacks_mutex_);
+    for (uint64_t frame : unique_callstacks_.at(callstack_id)->GetFrames()) {
+      action(frame);
+    }
+  }
+
+  void ClearCallstacks() {
```
florian-kuebler, line 62: I think at some point soon, we want to get rid of this method and just ensure, that callstack_data (or better its container), gets reassigned. 
Can you add a TODO to carefully remove this at some point?

OrbitCore/CallstackData.h
```diff
@@ -0,0 +1,82 @@
...
+    absl::MutexLock lock(&unique_callstacks_mutex_);
+    unique_callstacks_.clear();
+    callstack_events_.clear();
+  }
+
+ private:
+  std::shared_ptr<CallStack> GetCallstackPtr(CallstackID callstack_id) const {
+    absl::MutexLock lock(&unique_callstacks_mutex_);
+    return unique_callstacks_.at(callstack_id);
```
florian-kuebler, line 71: As this will fail if not present, please use `find` and return `nullptr` if not present.

OrbitCore/CaptureData.h
```diff
@@ -118,6 +142,10 @@ class CaptureData {
 
   absl::flat_hash_map<uint64_t, orbit_client_protos::FunctionStats>
       functions_stats_;
+
+  std::unique_ptr<CallstackData> callstack_data_;
```
florian-kuebler, line 146: Any reason for those two the be unique_ptrs rather than just members? Just a clarification question.

OrbitCore/CaptureData.h
```diff
@@ -103,6 +110,23 @@ class CaptureData {
...
+  [[nodiscard]] CallstackData* GetCallstackData() {
+    return callstack_data_.get();
+  }
+
+  void SetSelectionCallstackData(
+      std::unique_ptr<CallstackData> callstack_data) {
+    selection_callstack_data_ = std::move(callstack_data);
+  }
+
+  [[nodiscard]] CallstackData* GetSelectionCallstackData() {
```
florian-kuebler, line 126: Can this be const?

OrbitCore/CaptureData.h
```diff
@@ -103,6 +110,23 @@ class CaptureData {
   void UpdateFunctionStats(orbit_client_protos::FunctionInfo* func,
                            const orbit_client_protos::TimerInfo& timer_info);
 
+  void SetCallstackData(std::unique_ptr<CallstackData> callstack_data) {
+    callstack_data_ = std::move(callstack_data);
+  }
+
+  [[nodiscard]] CallstackData* GetCallstackData() {
```
florian-kuebler, line 117: Can this method be const and return a const pointer?

OrbitCore/CallstackData.h
```diff
@@ -0,0 +1,82 @@
...
+  }
+
+  void ClearCallstacks() {
+    absl::MutexLock lock(&unique_callstacks_mutex_);
+    unique_callstacks_.clear();
+    callstack_events_.clear();
+  }
+
+ private:
+  std::shared_ptr<CallStack> GetCallstackPtr(CallstackID callstack_id) const {
```
florian-kuebler, line 69: Can it be also return a const ptr?
Also could you add [[nodiscard]]?

OrbitCore/CallstackData.h
```diff
@@ -0,0 +1,82 @@
...
+  [[nodiscard]] uint32_t GetCallstackEventsSize() const {
+    return callstack_events_.size();
+  };
+
+  const CallStack& GetCallStack(CallstackID callstack_id) const {
+    absl::MutexLock lock(&unique_callstacks_mutex_);
+    return *unique_callstacks_.at(callstack_id);
+  }
+
+  bool HasCallStack(CallstackID callstack_id) const {
```
florian-kuebler, line 40: [[nodiscard]]?

OrbitCore/CallstackData.h
```diff
@@ -0,0 +1,82 @@
...
+
+  [[nodiscard]] const BlockChain<orbit_client_protos::CallstackEvent,
+                                 16 * 1024>&
+  GetCallstackEvents() const { return callstack_events_; };
+
+  [[nodiscard]] uint32_t GetCallstackEventsSize() const {
+    return callstack_events_.size();
+  };
+
+  const CallStack& GetCallStack(CallstackID callstack_id) const {
```
florian-kuebler, line 35: [[nodiscard]]?

OrbitCore/CallstackData.h
florian-kuebler: Also, can you add  [[nodiscard]]

OrbitCore/SamplingProfiler.cpp
```diff
@@ -145,8 +128,8 @@ void SamplingProfiler::ProcessSamples() {
   address_to_module_name_.clear();
 
   // Unique call stacks and per thread data
-  for (const CallstackEvent& callstack : callstack_events_) {
-    if (!HasCallStack(callstack.callstack_hash())) {
+  for (const CallstackEvent& callstack : callstack_data.GetCallstackEvents()) {
+    if (!callstack_data.HasCallStack(callstack.callstack_hash())) {
```
florian-kuebler, line 132: This would be a programming error, right? So could you change this into a CHECK here?

OrbitGl/App.cpp
```diff
@@ -123,7 +124,7 @@ void OrbitApp::OnKeyAndString(uint64_t key, std::string str) {
 }
 
 void OrbitApp::OnCallstack(CallStack callstack) {
```
florian-kuebler, line 126: Can you rename this to something like `OnUniqueCallstack`?

OrbitGl/App.cpp
```diff
@@ -123,7 +124,7 @@ void OrbitApp::OnKeyAndString(uint64_t key, std::string str) {
...
-  Capture::GSamplingProfiler->AddUniqueCallStack(callstack);
+  Capture::capture_data_.GetCallstackData()->AddUniqueCallStack(callstack);
```
florian-kuebler, line 127: I am thinking about if it would be maybe better to hide those methods to CaptureData, and then call CaptureData->AddUniqueCallstack. What do you think? Could be also in a separate PR, and I am not quite sure about this.

OrbitCore/CallstackData.cpp
IrinaShkviro: We use lock only for unique_callstacks_, here we change only callstack_events_

florian-kuebler: But should the callstack_events_ also be guarded by a lock?

OrbitCore/CallstackData.h
IrinaShkviro: I thought that it's not a strict rule (so you may use one of two: GetCallstackEvents() or callstack_events()). For which function do you mean add `[[nodiscard]]`?

florian-kuebler: I though we wanted to make it more strict. I would add them to all getters here.
Basically, I would add them always, when you can consider the behavior of the program to be incorrect, when calling the method and dropping the return value.
In this case: Calling a getter and not using the value, can only be wrong. Thus, you want the compiler to tell you this.

OrbitCore/CallstackData.h
IrinaShkviro: For doing something with `unique_callstacks_` we should lock the `unique_callstacks_mutex_`, and it means that if we want to do something outside, we should get access to this mutex as well. But I would prefer to keep this mutex private.

OrbitCore/CaptureData.h
florian-kuebler: As discussed offline, just add a comment with the explanation here.

OrbitGl/App.cpp
IrinaShkviro: Added these methods to CaptureData.

OrbitGl/App.cpp
IrinaShkviro: Done

OrbitCore/SamplingProfiler.cpp
IrinaShkviro: Yes, replaced with CHECK

OrbitCore/CallstackData.h
IrinaShkviro: Done

OrbitCore/CallstackData.h
IrinaShkviro: Done

OrbitCore/CallstackData.h
IrinaShkviro: Let's keep it a shared_ptr for now: I want to try to get rid of the shared_ptr here, and in this case we should duplicate CallStack data. So, before doing that change I want to double check that I won't break/slow down something.

OrbitCore/CaptureData.h
IrinaShkviro: Done

OrbitCore/CaptureData.h
IrinaShkviro: Done

OrbitCore/CallstackData.h
IrinaShkviro: Added `find` and return const pointer here and in GetCallStack function.

OrbitCore/CallstackData.h
IrinaShkviro: There is just one place where we use it (in CaptureSerializer), and it doesn't make any sense to keep it here as we always create a new capture_data there, so I've removed it.

OrbitCore/CallstackData.h
IrinaShkviro: Done

OrbitCore/CallstackData.h
IrinaShkviro: Renamed the function

OrbitCore/CallstackData.cpp
IrinaShkviro: Renamed the argument and the function.

OrbitCore/CallstackData.cpp
IrinaShkviro: Done

OrbitCore/CallstackData.cpp
IrinaShkviro: Probably yes... But we should understand if it implies on performance...

IrinaShkviro:
I was typing too slow :)

OrbitCore/CallstackData.cpp
florian-kuebler: I think we prefer value arguments for sinks, and move them explicitly on the call side, such that you can see on the caller, that the object becomes invalid after the call.

OrbitCore/CallstackData.cpp
```diff
@@ -0,0 +1,40 @@
...
+
+using orbit_client_protos::CallstackEvent;
+
+void CallstackData::AddCallstackEvent(CallstackEvent&& callstack_event) {
+  CallstackID hash = callstack_event.callstack_hash();
+  CHECK(HasCallStack(hash));
+  callstack_events_.push_back(std::move(callstack_event));
+}
+
+void CallstackData::AddUniqueCallStack(CallStack&& call_stack) {
```
florian-kuebler, line 19: Same as above. Should probably value argument and move on call.

OrbitCore/CaptureData.h
```diff
@@ -103,6 +110,38 @@ class CaptureData {
...
 
+  [[nodiscard]] const CallstackData* GetCallstackData() const {
+    return callstack_data_.get();
+  };
+
+  void AddUniqueCallStack(CallStack&& call_stack) {
+    callstack_data_->AddUniqueCallStack(std::move(call_stack));
+  }
+
+  void AddCallstackEvent(
```
florian-kuebler, line 121: As above, probably we want value args here.

OrbitCore/CaptureData.h
```diff
@@ -103,6 +110,38 @@ class CaptureData {
...
+      orbit_client_protos::CallstackEvent&& callstack_event) {
+    callstack_data_->AddCallstackEvent(std::move(callstack_event));
+  }
+
+  [[nodiscard]] const CallstackData* GetSelectionCallstackData() const {
+    return selection_callstack_data_.get();
+  };
+
+  void AddSelectionCallStack(CallStack&& call_stack) {
```
florian-kuebler, line 130: And here.

OrbitCore/CaptureData.h
```diff
@@ -103,6 +110,38 @@ class CaptureData {
...
+    selection_callstack_data_->AddUniqueCallStack(std::move(call_stack));
+  }
+
+  void AddSelectionCallstackEvent(
```
florian-kuebler, line 134: also here.

OrbitCore/CallstackData.cpp
IrinaShkviro: Done

OrbitCore/CallstackData.cpp
IrinaShkviro: Done

OrbitCore/CaptureData.h
IrinaShkviro: Done

OrbitCore/CaptureData.h
IrinaShkviro: Done

OrbitCore/CaptureData.h
IrinaShkviro: Done

PR #1034, opened 2020-08-14
Get the names of all the tracepoint events in the service using protos and grpc

MariaSandru27:


MariaSandru27:
there were checks that failed

PR #1038, opened 2020-08-14
Add OrbitBase/DebugUtils.h

pierricgimmig:


OrbitBase/include/OrbitBase/DebugUtils.h
```diff
@@ -0,0 +1,26 @@
...
+
+#define DEBUG_UTILS_ENABLED 0  // Enable locally only.
+#if DEBUG_UTILS_ENABLED
+
+// Print variable's name and value in the form "name = value".
+#define PRINT_VAR(x) LOG("%s = %s", #x, std::to_string(x))
+// Print current function's name, file and line as url, and thread id.
+#define PRINT_FUNC                                      \
+  LOG("%s %s(%u) %s", __FUNCTION__, __FILE__, __LINE__, \
+      std::to_string(absl::base_internal::GetTID()))
```
dimitry-, line 22: is it different from thread::self::get_id()? (we probably want to avoid using absl internals)

PR #1041, opened 2020-08-15
Add .git-blame-ignore-revs file

dimitry-:
This file contains list of commits doing
bulk changes to repository. Commits listed
in this file are ignored by git blame.

Run following command to enable it:
git config blame.ignoreRevsFile .git-blame-ignore-revs

### DIFF
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -1,0 +1,7 @@
+# This list contains commits for git blame to ignore
+# These are usually commits with bulk changes like
+# clang-format.
+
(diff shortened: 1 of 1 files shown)


pierricgimmig:
Awesome, thanks!

pierricgimmig:
Suggestion: can we add comments with the reason why commits were added to the list?

florian-kuebler:
Thanks. This will make git blame much better for us!

dimitry-:
> Suggestion: can we add comments with the reason why commits were added to the list?

done,

PR #1043, opened 2020-08-15
Add and implement StatusListener

dimitry-:
StatusListener is an interface to be used by controllers
(OrbitApp) to notify user about progress of background tasks

This commit adds the interface and simple Qt implementation using
QStatusBar showMessage/clearMessage.

Test: run OrbitQtTests

### DIFF
--- a/OrbitQt/StatusListenerImpl.cpp
+++ b/OrbitQt/StatusListenerImpl.cpp
@@ -1,0 +1,52 @@
+// Copyright (c) 2020 The Orbit Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#include "StatusListenerImpl.h"
+
+#include "OrbitBase/Logging.h"
+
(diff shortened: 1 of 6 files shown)


OrbitQt/StatusListenerImpl.cpp
```diff
@@ -0,0 +1,60 @@
...
+  uint64_t candidate = current_id;
+  do {
+    if (status_messages_.count(candidate) == 0) {
+      return candidate;
+    }
+    candidate = next_id_++;
+  } while (candidate != current_id);
+
+  FATAL("Failed to find free (64bit) id and somehow we did not run out of memory before that.");
+}
```
pierricgimmig, line 60: Whoa. If I understand correctly, this is to handle a uint64_t value wrapping around? That's an astronomical number of status updates, it is safe to say that this will never happen :-). At 1GHz updates, it would take over 500 years. I'd be happy with a simple "next_id_++".

dimitry-: ok, I simplified this.

PR #1044, opened 2020-08-17
Fix memory leaks when loading and clearing captures

pierricgimmig:
- Clear CaptureData when clearing capture
- Set empty sampling profiler when clearing capture to trigger deallocation of previous sampling related data
- Make TopDownView own its Model and ProxyModel instead relying on Qt for memory management

When loading a capture and clearing it, we held on to a lot of memory. See before and after comparison below where we load and clear the same capture file twice. Before the change, we start off with 83MB and end with 106MB of memory usage whereas after the change, we get back to 83MB after the loading/clearing operations.

In the graphs below, the green curve represents the number of live allocations and the blue curve represents the number of bytes allocated.

image

### DIFF
--- a/OrbitCore/Capture.cpp
+++ b/OrbitCore/Capture.cpp
@@ -15,10 +15,14 @@
 #include "absl/strings/str_format.h"
 
 using orbit_client_protos::FunctionInfo;
+using orbit_client_protos::LinuxAddressInfo;
 using orbit_client_protos::PresetFile;
+using orbit_client_protos::PresetInfo;
 
 CaptureData Capture::capture_data_;
+absl::flat_hash_map<uint64_t, FunctionInfo> Capture::GVisibleFunctionsMap;
 TextBox* Capture::GSelectedTextBox = nullptr;
+ThreadID Capture::GSelectedThreadId;
 
 std::shared_ptr<SamplingProfiler> Capture::GSamplingProfiler = nullptr;
 std::shared_ptr<Process> Capture::GTargetProcess = nullptr;
@@ -53,7 +57,11 @@ void Capture::FinalizeCapture() {
   }
 }
 
-void Capture::ClearCaptureData() { GSelectedTextBox = nullptr; }
+void Capture::ClearCaptureData() {
+  GSelectedTextBox = nullptr;
+  GSelectedThreadId = 0;
+  capture_data_ = CaptureData();
+}
 
 void Capture::PreSave() {
   // Add selected functions' exact address to sampling profiler
--- a/OrbitCore/Capture.cpp
+++ b/OrbitCore/Capture.cpp
@@ -60,7 +60,6 @@ void Capture::FinalizeCapture() {
 void Capture::ClearCaptureData() {
   GSelectedTextBox = nullptr;
   GSelectedThreadId = 0;
-  capture_data_ = CaptureData();
 }
 
 void Capture::PreSave() {
--- a/OrbitCore/Capture.cpp
+++ b/OrbitCore/Capture.cpp
@@ -45,9 +45,6 @@ ErrorMessageOr<void> Capture::StartCapture(
   }
 
   capture_data_ = CaptureData(process_id, std::move(process_name), std::move(selected_functions));
-
-  Capture::GSamplingProfiler = std::make_shared<SamplingProfiler>(Capture::GTargetProcess);
-
   return outcome::success();
 }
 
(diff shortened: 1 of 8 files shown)


OrbitCore/Capture.cpp
```diff
@@ -60,6 +60,7 @@ void Capture::FinalizeCapture() {
 void Capture::ClearCaptureData() {
   GSelectedTextBox = nullptr;
   GSelectedThreadId = 0;
+  capture_data_ = CaptureData();
```
florian-kuebler, line 63: This will break functionality due to recent changes.
ClearCaptureData gets invoked basically from two places:
1. `OnLoadCapture`, invokes this, and then calls `CaptureSerializer` to load the capture (which will set capture data again). So this is no harmful change here, but still not needed, as it will be reset a couple of instructions later.

2. `OnCaptureStarted`: This gets invoked by the `CaptureClient` **after** the `App::StartCapture` resets the `capture_data_`. Thus all information (like the selected_functions), will be gone. Again, this reset would not be needed, as StartCapture, sets new capture data.

pierricgimmig: > This will break functionality due to recent changes.

Yes it seems like it does. I'll look into it, thanks!

OrbitGl/App.cpp
```diff
@@ -597,6 +597,13 @@ void OrbitApp::StopCapture() {
 void OrbitApp::ClearCapture() {
   Capture::ClearCaptureData();
 
+  // Trigger deallocation of previous sampling related data.
+  sampling_report_ = nullptr;
```
reichlfl, line 601: Nit: This should be handled in the subsequent call to OrbitApp::AddSamplingReport, since sampling_report_ is reassigned there. Is there a reason to do this explicit nullptr assignment here?

OrbitGl/App.cpp
```diff
@@ -597,6 +597,13 @@ void OrbitApp::StopCapture() {
...
+  auto empty_sampling_profiler = std::make_shared<SamplingProfiler>();
+  Capture::GSamplingProfiler = empty_sampling_profiler;
```
florian-kuebler, line 603: There might be the same issue as above due to the execution order, when ClearCapture is invoked. However, I think you spotted the right class for the Memory leak. I just looked up when `SamplingProfiler` and found this in `CaptureSerializer:
```
if (Capture::GSamplingProfiler == nullptr) {
    Capture::GSamplingProfiler = std::make_shared<SamplingProfiler>();
  }
  Capture::GSamplingProfiler->ClearCallstacks();
```
I think the if should be removed and the clear callstacks method should be also removed. And we are fine.
Looping in @IrinaShkviro, as she is currently working on moving the callstacks from sampling profiler. So might be interssting.

florian-kuebler:
As stated above, I think you can just remove the reassignment of `capture_data_ ` and `GSamplingProfiler` and remove the if in `CaptureSerializer`, to always reset `GSamplingProfiler`.
Otherwise looks good to me. Thanks for looking into this!

OrbitGl/App.cpp
pierricgimmig: Done, thanks!

OrbitGl/App.cpp
pierricgimmig: I removed that block altogether since it's done in ClearCapture now.

pierricgimmig:
Hi @florian-kuebler , thanks for catching those issues. I modified the PR:
- Make ClearCapture responsible for creating the new SamplingProfiller
- Don't clear CaptureData
- Don't switch to sampling tab if the sampling report has no samples

OrbitGl/SamplingReport.h
```diff
@@ -29,6 +29,7 @@ class SamplingReport {
   [[nodiscard]] std::string GetSelectedCallstackString() const;
   void SetUiRefreshFunc(std::function<void()> func) { ui_refresh_func_ = std::move(func); };
   [[nodiscard]] bool HasCallstacks() const { return selected_sorted_callstack_report_ != nullptr; };
+  [[nodiscard]] bool HasSamples() const { return profiler_->GetNumSamples() > 0; }
```
florian-kuebler, line 32: This will be removed with #1033. However, keep it here, and we ill move it in the other PR.

pierricgimmig:
> LGTM now!. Thanks!

Cool, thanks Florian!

OrbitGl/SamplingReport.h
pierricgimmig: Oh ok, thanks.

PR #1051, opened 2020-08-17
Show message about loading module in status bar

dimitry-:
Show the status bar and show progress messages while loading it.

Bug: http://b/162705265
Test: clear cache, start Orbit, load symbols for multiple modules
      check that messages in status bar showed and updated according
      on what Orbit is currently doing.

### DIFF
--- a/OrbitQt/orbitmainwindow.cpp
+++ b/OrbitQt/orbitmainwindow.cpp
@@ -15,6 +15,7 @@
 #include <QMouseEvent>
 #include <QProgressDialog>
 #include <QSettings>
+#include <QStatusBar>
 #include <QTimer>
 #include <QToolTip>
 #include <utility>
@@ -24,6 +25,7 @@
 #include "OrbitVersion/OrbitVersion.h"
 #include "Path.h"
 #include "SamplingReport.h"
+#include "StatusListenerImpl.h"
 #include "TopDownViewItemModel.h"
 #include "absl/flags/flag.h"
 #include "absl/strings/match.h"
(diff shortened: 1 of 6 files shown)


OrbitQt/orbitmainwindow.cpp
```diff
@@ -64,6 +66,12 @@ OrbitMainWindow::OrbitMainWindow(QApplication* a_App, ApplicationOptions&& optio
   ui->HomeHorizontalSplitter->setSizes(sizes);
   ui->splitter_2->setSizes(sizes);
 
+  QStatusBar* status_bar = new QStatusBar(this);
```
reichlfl, line 69: I'd prefer to define this in the ui file and not create it in code.

OrbitQt/orbitmainwindow.h
```diff
@@ -98,7 +99,8 @@ class OrbitMainWindow : public QMainWindow {
   QIcon icon_start_capture_;
   QIcon icon_stop_capture_;
 
-  std::string m_CurrentPdbName;
```
reichlfl, line 101: Was this intentionally removed, seems quite unrelated to the PR?
Apparently it was unneeded, so it's fine for me, just want to make sure.

OrbitQt/orbitmainwindow.cpp
dimitry-: I think this is supposed to be created in the code? At least I did not see anything it designer. If it needs to be done in UI which section should this be added to?

OrbitQt/orbitmainwindow.h
dimitry-: Not related no - but since it was unused I removed it.

OrbitQt/orbitmainwindow.cpp
reichlfl: It's a bit hidden. In the designer: Right-click the main window, select "Create status bar". Then you can access the status bar in code through `OrbitMainWindow::statusBar()`. The Designer then adds a `<widget class="QStatusBar" name="statusBar"/>` directly as child of the QMainWindow.

dimitry-: dione.

OrbitQt/orbitmainwindow.ui
```diff
@@ -223,7 +223,7 @@
      <x>0</x>
      <y>0</y>
      <width>976</width>
-     <height>22</height>
```
dimitry-, line 226: hmm. I'm not sure about this one.

reichlfl: @beckerhe I think this should be the same as what happened to you when you cleaned up the UI files, right?

dimitry-:
ops. I forgot to rebase - I guess we are stuck with this patchset commit now :(

PR #1053, opened 2020-08-17
Add MESA's opengl32.dll to Win32 Docker containers

beckerhe:
This is necessary since the new OrbitGl-tests require OpenGL to be available.

The MESA software renderer is the best way to fulfill that requirement. So this PR changes the Dockerfiles such that MESA's opengl32.dll will be installed in the containers.


beckerhe:
> Thanks for this! LGTM

Thanks for the quick review!

PR #1054, opened 2020-08-17
Remove tr-tool from docker call in Kokoro

beckerhe:
Windows docker-run-calls were piped to tr to substitute Windows
line-endings `\r\n` by Unix-style line endings `\n`.

But this means the exit code does not properly propagate to Kokoro. This
commit fixes that by removing `tr` for now.

### DIFF
--- a/kokoro/builds/build.sh
+++ b/kokoro/builds/build.sh
@@ -199,5 +199,5 @@ else
     --isolation=process --storage-opt 'size=50GB' \
     ${CONTAINER} \
     'C:/Program Files/Git/bin/bash.exe' -c \
-    "/c/mnt/github/orbitprofiler/kokoro/builds/build.sh in_docker" | tr -d '\r'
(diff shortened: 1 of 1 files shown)


kokoro/builds/build.sh
```diff
@@ -199,5 +199,5 @@ else
     --isolation=process --storage-opt 'size=50GB' \
     ${CONTAINER} \
     'C:/Program Files/Git/bin/bash.exe' -c \
-    "/c/mnt/github/orbitprofiler/kokoro/builds/build.sh in_docker" | tr -d '\r'
```
danielfenner, line 202: If the 'tr' is needed for anything you can try to read PIPESTATUS[0] afterwards and return that to kokoro.

beckerhe: Yeah, good point. I thought about doing something like that. But the `tr`-solution was never finished anyway (logfiles looked better but there still superfluous line breaks.). So I decided to postpone the whole topic which means logfiles continue to look like they are right now.

PR #1056, opened 2020-08-17
Fix typo in register_tests

dimitry-:
Actually registers OrbitQtTests now.

### DIFF
--- a/OrbitQt/CMakeLists.txt
+++ b/OrbitQt/CMakeLists.txt
@@ -133,4 +133,4 @@ target_link_libraries(
           Qt5::Core
           GTest::Main)
 
-register_test(OrbitGlTests)
+register_test(OrbitQtTests)
--- a/kokoro/builds/build.sh
+++ b/kokoro/builds/build.sh
@@ -38,6 +38,8 @@ if [ -n "$1" ]; then
   readonly MOUNT_POINT="$( cd "$( dirname "${BASH_SOURCE[0]}" )/../../../../" >/dev/null 2>&1 && pwd )"
   readonly KEYSTORE_PATH="${MOUNT_POINT}/keystore"
 
+  export QT_QPA_PLATFORM=offscreen
+
   if [ -z "$BUILD_TYPE" ]; then
     readonly BUILD_TYPE="$(basename "$KOKORO_JOB_NAME")"
   fi


dimitry-:
hmm:

```
[ RUN      ] StatusListenerImpl.ShowAndClearOneMessage
qt.qpa.xcb: could not connect to display 
qt.qpa.plugin: Could not load the Qt platform plugin "xcb" in "" even though it was found.
This application failed to start because no Qt platform plugin could be initialized. Reinstalling the application may fix this problem.

Available platform plugins are: eglfs, linuxfb, minimal, minimalegl, offscreen, vnc, xcb.
```

dimitry-:
I guess this is somewhat expected.. not sure what to do with it though. We can leave the test unregistered, I guess.

beckerhe:
> I guess this is somewhat expected.. not sure what to do with it though. We can leave the test unregistered, I guess.

Can you try to add `export QT_QPA_PLATFORM=offscreen` to the bash-CI-build-script in `kokoro/builds/build.sh`?
Somewhere after the `We are inside the docker container`-comment should be fine.

beckerhe:
> > I guess this is somewhat expected.. not sure what to do with it though. We can leave the test unregistered, I guess.
> 
> Can you try to add `export QT_QPA_PLATFORM=offscreen` to the bash-CI-build-script in `kokoro/builds/build.sh`?
> Somewhere after the `We are inside the docker container`-comment should be fine.

Okay, great. So it works at least on Linux. I'm not sure about Windows though. It might not work anymore when we move to docker. Could you wait a little bit longer before merging? I wanna try this out.

beckerhe:
> > > I guess this is somewhat expected.. not sure what to do with it though. We can leave the test unregistered, I guess.
> > 
> > 
> > Can you try to add `export QT_QPA_PLATFORM=offscreen` to the bash-CI-build-script in `kokoro/builds/build.sh`?
> > Somewhere after the `We are inside the docker container`-comment should be fine.
> 
> Okay, great. So it works at least on Linux. I'm not sure about Windows though. It might not work anymore when we move to docker. Could you wait a little bit longer before merging? I wanna try this out.

Unfortunately I won't work when running inside a Windows docker container. We should probably disable the test registration in the windows on docker case.

Feel free to merge that here. I will take care of it in a subsequent PR.

PR #1057, opened 2020-08-17
Remove abseil dependency from Orbit.h

pierricgimmig:
Orbit.h will be used as the public API for manual instrumentation, we shouldn't have it depend on anything.

### DIFF
--- a/Orbit.h
+++ b/Orbit.h
@@ -7,7 +7,7 @@
 
 #include <stdint.h>
 
-#include "absl/base/casts.h"
+#include <cstring>
 
 // Orbit API (header-only)
 //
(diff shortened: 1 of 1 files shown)


Orbit.h
```diff
@@ -92,10 +90,10 @@ ORBIT_STUB void TrackDoubleAsInt64(const char*, int64_t) { ORBIT_NOOP(); }
 // Convert floating point arguments to integer arguments as we can't access
 // XMM registers with our current dynamic instrumentation on Linux (uprobes).
 inline void TrackFloat(const char* name, float value) {
-  TrackFloatAsInt(name, absl::bit_cast<int32_t>(value));
+  TrackFloatAsInt(name, *reinterpret_cast<int32_t*>(&value));
```
dimitry-, line 93: In which case it is better to use memcpy here.

int int_value;
static_assert(sizeof(value == sizeof(int_value);
memcpy(&int_value, &value, sizeof(value));
TrackFloatAsInt(...);

beckerhe: Yeah, I would also vote for std::memcpy. All major compilers optimize that away, so there are no downsides and it has the advantage that it's not UB.

Apart from that. Thanks for fixing this!

dimitry-:
Also maybe add a compile target for something with Orbit.h (depending on nothing) so that we do not break it during future refactorings.

Orbit.h
pierricgimmig: Cool, done, thanks!

PR #1058, opened 2020-08-18
Disabled local symbols when build id is missing.

antonrohr:


OrbitGl/App.cpp
```diff
@@ -767,6 +767,14 @@ void OrbitApp::LoadModules(int32_t process_id, const std::vector<std::shared_ptr
     // TODO (159889010) Move symbol loading off the main thread.
     const std::string& module_path = module->m_FullName;
     const std::string& build_id = module->m_DebugSignature;
+    if (build_id.empty()) {
+      LOG("Warning: Module \"%s\" does not contain a build id. Orbit cannot use local symbol files "
+          "or cache symbols without a build id.",
```
dimitry-, line 772: Yeah - I think we should just stop handling missing build_id use-cases.

antonrohr: I don't know if you saw the chatroom messages. Maybe we could use a hash of the whole file. But thats a thing we have to discuss and its definitely not pressing.

PR #1061, opened 2020-08-18
Remove GSessionPreset global

florian-kuebler:
Capture::GSessionPreset was used to support the following scenario:
1. Load a preset, whose associated process is running but not selected.
2. Loading the preset was asynch. selecting the process and then
  applying the preset.

If the process was not running at all, an error was shown to the user.

This required keeping track of the preset variable through not
straightforward control-flow.
This change will ask the user to select the process by herself, which
allows to remove the global and makes the control-flow explicit.

Bug: http://b/163020637
Test: Compile, Select preset when
      (1) process is runnig and selected,
      (2) process is running but not selected,
      (3) process is not running.

### DIFF
--- a/OrbitCore/Capture.cpp
+++ b/OrbitCore/Capture.cpp
@@ -15,13 +15,11 @@
 #include "absl/strings/str_format.h"
 
 using orbit_client_protos::FunctionInfo;
-using orbit_client_protos::PresetFile;
 
 CaptureData Capture::capture_data_;
 
 std::shared_ptr<SamplingProfiler> Capture::GSamplingProfiler = nullptr;
 std::shared_ptr<Process> Capture::GTargetProcess = nullptr;
-std::shared_ptr<PresetFile> Capture::GSessionPresets = nullptr;
 
 void Capture::Init() { GTargetProcess = std::make_shared<Process>(); }
 
--- a/OrbitCore/Capture.h
+++ b/OrbitCore/Capture.h
@@ -32,7 +32,6 @@ class Capture {
 
   static std::shared_ptr<SamplingProfiler> GSamplingProfiler;
   static std::shared_ptr<Process> GTargetProcess;
-  static std::shared_ptr<orbit_client_protos::PresetFile> GSessionPresets;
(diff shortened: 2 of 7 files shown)


pierricgimmig:
Hi Florian. I'm all for removing globals. However it seems this introduces a regression in terms of user experience. A user now has to manually select a process and then load the preset, this is unfortunate. Could we get rid of the global without changing the behavior?

florian-kuebler:
Hi, yes, I can also pass the pointer all along as an argument. Bit more code and a bit less straightforward control-flow, but should also work.

antonrohr:
You could call it a regression, or you can call it a clarification. 
How the functionality worked beforehand is that the process selection happened automatically and it was not indicated to the user that a preset is bound to a process at all. One could imagine that a user is quite surprised when the process selection changes after a preset is loaded. I am fine with this change for the sake of getting rid of a global.

In the longterm I would like a preset be process agnostic and only be bound to modules and hooked functions.

antonrohr:
One could imagine a user having one preset for libggp functions and two versions of the game (maybe cool_game.elf and cool_game_start_at_level_3.elf). There is no need for the preset to be bound to the process path at all.

florian-kuebler:
I changed the code accordingly and pass it as parameter all the way down. As I said, control-flow is again not that intuitive anymore, but users will keep the same experience.

pierricgimmig:
> In the longterm I would like a preset be process agnostic and only be bound to modules and hooked functions.

I'm not against it, I think that presets are currently flawed in that they depend on an exact executable name. Partners sometimes have different names depending on the configuration. For now, since the preset is linked to an exe name, I thought it was weird to ask the user to select the process that the preset already depends on. 

Ideally a new solution would still allow us to load a preset and select an exe in a single step, but wouldn't tie a preset to a single exe, maybe by having a list of potential executables it can apply to. If any of those exe are running then Orbit would automatically select it and load the debug info. We could also apply any preset to any another exe, maybe by having a contextual menu ("apply to selected process").

PR #1064, opened 2020-08-18
Remove usages of GTargetProcess in OrbitCore

florian-kuebler:
Capture::GTargetProcess will be moved to a non-global in OrbitGl
soon.
This PR prepares this, by removing accesses to it within OrbitCore
and pass it as parameter.
Also removes the Capture::StartCapture method, as it was only setting
capture_data_ and checking the pid to be selected.

Test: Compile
Bug: http:/b/163020637

### DIFF
--- a/OrbitCore/Pdb.h
+++ b/OrbitCore/Pdb.h
@@ -38,7 +38,7 @@ class Pdb {
   orbit_client_protos::FunctionInfo* GetFunctionFromExactAddress(uint64_t a_Address);
   orbit_client_protos::FunctionInfo* GetFunctionFromProgramCounter(uint64_t a_Address);
 
-  void ProcessData(class Process* process);
+  void ProcessData();
 
  private:
   uint64_t m_MainModule = 0;
--- a/OrbitCore/Pdb.h
+++ b/OrbitCore/Pdb.h
@@ -38,7 +38,7 @@ class Pdb {
   orbit_client_protos::FunctionInfo* GetFunctionFromExactAddress(uint64_t a_Address);
   orbit_client_protos::FunctionInfo* GetFunctionFromProgramCounter(uint64_t a_Address);
 
-  void ProcessData();
+  void ProcessData(class Process* process);
 
  private:
   uint64_t m_MainModule = 0;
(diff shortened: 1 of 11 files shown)


OrbitCore/Pdb.h
```diff
@@ -38,7 +38,7 @@ class Pdb {
   orbit_client_protos::FunctionInfo* GetFunctionFromExactAddress(uint64_t a_Address);
   orbit_client_protos::FunctionInfo* GetFunctionFromProgramCounter(uint64_t a_Address);
 
-  void ProcessData();
+  void ProcessData(class Process* process);
```
IrinaShkviro, line 41: Why not just `void ProcessData(Process* process);`

florian-kuebler: Then the headers would depend cyclicly from each other.

kisenshi: To avoid using "class Process* process" in the .cpp as well (as it already includes OrbitProcess.h), would it be possible to use a forward declaration in the .h so you can declare the method as ProcessData(Process* process) instead?

florian-kuebler: Thanks for the suggestions. As discussed with Irina offline, I have removed the dependency to Process in Pdb completely and added a AddFunctions method to the process.

florian-kuebler:
Thanks for the reviews so far. I modified the usage of Process by adding a AddFunctions method and call this from the App directly.

OrbitGl/App.cpp
```diff
@@ -727,6 +729,7 @@ void OrbitApp::LoadModuleOnRemote(int32_t process_id, const std::shared_ptr<Modu
         return;
       }
       module->LoadSymbols(result.value());
+      Capture::GTargetProcess->AddFunctions(module->m_Pdb->GetFunctions());
```
IrinaShkviro, line 732: Should we check if `Capture::GTargetProcess` is nullptr?

OrbitGl/App.cpp
```diff
@@ -779,6 +782,7 @@ void OrbitApp::LoadModules(int32_t process_id, const std::vector<std::shared_ptr
 
     if (symbols) {
       module->LoadSymbols(symbols.value());
+      Capture::GTargetProcess->AddFunctions(module->m_Pdb->GetFunctions());
```
IrinaShkviro, line 785: And the same here

florian-kuebler: Right, I forgot about this.

PR #1065, opened 2020-08-18
Refactor TextBox

reichlfl:
This PR removes most functionality from TextBox.

We create a TextBox for every Timer, but it is only used as a storage class for the related text and potentially for measurements. The Draw() method was never called when rendering timers, and in nearly all cases, the TextBox is only used to access the linked timer.

The only place where TextBox::Draw() was really used is to draw the overlay when selecting a timeslice in the capture window, and this functionality abused TextBox by not passing any valid timer at all, it only required drawing of the green rectangle and some text.

Refactor:
* All rendering-related functionality of TextBox was removed, it is purely used to store a pointer to a timer and some additional layouting information
* Drawing of the selection rectangle has been implemented in the CaptureWindow without use of TextBox

Maybe we could remove TextBox completely in the future and work with the timer pointers directly.

### DIFF
--- a/OrbitGl/CaptureWindow.cpp
+++ b/OrbitGl/CaptureWindow.cpp
@@ -93,6 +93,7 @@ void CaptureWindow::MouseMoved(int a_X, int a_Y, bool a_Left, bool /*a_Right*/, 
   }
 
   if (m_IsSelecting) {
+    ScreenToWorld(std::max(a_X, 0), std::max(a_Y, 0), worldx, worldy);
     m_SelectStop = Vec2(worldx, worldy);
     m_TimeStop = time_graph_.GetTickFromWorld(worldx);
   }
--- a/OrbitGl/CaptureWindow.cpp
+++ b/OrbitGl/CaptureWindow.cpp
@@ -86,7 +86,6 @@ void CaptureWindow::MouseMoved(int a_X, int a_Y, bool a_Left, bool /*a_Right*/, 
     m_WorldTopLeftX = clamp(m_WorldTopLeftX, worldMin, worldMax - m_WorldWidth);
     m_WorldTopLeftY =
         clamp(m_WorldTopLeftY, m_WorldHeight - time_graph_.GetThreadTotalHeight(), m_WorldMaxY);
-    UpdateSceneBox();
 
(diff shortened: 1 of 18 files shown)


OrbitGl/TextBox.h
```diff
@@ -5,70 +5,25 @@
...
-  float GetMaxY() const { return m_Max[1]; }
-
-  Vec2 GetMin() const { return m_Min; }
-  Vec2 GetMax() const { return m_Max; }
+  const Vec2& GetSize() const { return size_; }
+  const Vec2& GetPos() const { return pos_; }
 
   const std::string& GetText() const { return text_; }
-  void SetText(const std::string& a_Text) { text_ = a_Text; }
+  void SetText(std::string_view text) { text_ = text; }
```
beckerhe, line 26: I think this should be a `std::string` and passed by value and then moved into the member variable. `std::string_view` would force a copy if the caller already has a string which could be moved into the member.

OrbitGl/CaptureWindow.cpp
```diff
@@ -805,4 +789,34 @@ void CaptureWindow::RenderTimeBar() {
...
+    Vec2 pos(from, m_WorldTopLeftY - m_WorldHeight);
+    Vec2 size(sizex, m_WorldHeight);
+
+    std::string text = GetPrettyTime(TicksToDuration(minTime, maxTime));
+    const Color color(0, 128, 0, 128);
+
+    Box box(pos, size, GlCanvas::Z_VALUE_BOX_ACTIVE);
+    ui_batcher_.AddBox(box, color);
+
+    static Color text_color(255, 255, 255, 255);
```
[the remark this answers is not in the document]
reichlfl: Right nevertheless, I changed it

OrbitGl/TextBox.h
reichlfl: Changed

reichlfl:
> Thanks. LGTM. I only added some C++-ish comments. But I really don't know much about this part of the code, so someone else should also approve (Gerrit's +1).

SInce Florian commented the same: @pierricgimmig I added you to the reviewer's list since Ronald is OOO - maybe you know of some side effects I was missing, just to be sure.

PR #1067, opened 2020-08-18
Clear callstack and selection views

IrinaShkviro:
Use pointer to the CallStack in the CallStackDataView instead of
creating a copy, remove extra FireRefreshCallbacks call in StopCapture
and clear CallStack and Selection views when clear capture or start a
new one.

Test: Take a capture, make a selection, clear capture and check the
content of the sampling, selection and callstack views; take a capture,
make a selection, start a new capture and check the content of the
sampling, selection and callstack views while capture in process.

Bug: http://b/164971132

### DIFF
--- a/OrbitGl/CallStackDataView.h
+++ b/OrbitGl/CallStackDataView.h
@@ -28,16 +28,16 @@ class CallStackDataView : public DataView {
                      const std::vector<int>& item_indices) override;
   void OnDataChanged() override;
   void SetCallStack(const CallStack& callstack) {
-    callstack_ = std::make_unique<CallStack>(callstack);
+    callstack_ = CallStack(callstack);
     OnDataChanged();
   }
 
-  void ClearCallstack() { callstack_ = nullptr; }
+  void ClearCallstack() { callstack_ = CallStack(); }
 
  protected:
   void DoFilter() override;
 
-  std::unique_ptr<CallStack> callstack_;
+  CallStack callstack_;
 
   struct CallStackDataViewFrame {
     CallStackDataViewFrame() = default;
--- a/OrbitGl/SamplingReport.cpp
+++ b/OrbitGl/SamplingReport.cpp
@@ -19,6 +19,16 @@ SamplingReport::SamplingReport(std::shared_ptr<SamplingProfiler> sampling_profil
   FillReport();
 }
 
+SamplingReport::~SamplingReport() { ClearReport(); }
+
+void SamplingReport::ClearReport() {
+  selected_sorted_callstack_report_ = nullptr;
+  selected_callstack_index_ = 0;
+  if (callstack_data_view_ != nullptr) {
+    callstack_data_view_->ClearCallstack();
+  }
+}
+
 void SamplingReport::FillReport() {
   const auto& sample_data = profiler_->GetThreadSampleData();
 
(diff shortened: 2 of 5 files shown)


OrbitGl/CallStackDataView.h
```diff
@@ -37,7 +37,7 @@ class CallStackDataView : public DataView {
  protected:
   void DoFilter() override;
 
-  std::unique_ptr<CallStack> callstack_;
+  const CallStack* callstack_;
```
reichlfl, line 40: Would it be a lot of effort to change this into a shared_ptr? I think we should avoid working with raw pointers if possible.

dimitry-: We should probably discuss this. We want to avoid shread_ptr because they mask ownership and they come with the cost of thread-safe reference counting. The long version of modivation is here https://groups.google.com/a/google.com/g/stadia-profiling-dev-muc/c/MFBcOfQQQ80/m/JME_UsJyAQAJ

dimitry- (approved):

OrbitGl/CallStackDataView.h
IrinaShkviro: I didn't get it: why do you think we need a shared_ptr here? I don't think that we need to share ownership here...

OrbitGl/CallStackDataView.h
```diff
@@ -27,8 +27,8 @@ class CallStackDataView : public DataView {
   void OnContextMenu(const std::string& action, int menu_index,
                      const std::vector<int>& item_indices) override;
   void OnDataChanged() override;
-  void SetCallStack(const CallStack& callstack) {
-    callstack_ = std::make_unique<CallStack>(callstack);
+  void SetCallStack(const CallStack* callstack) {
```
florian-kuebler, line 30: Do we ensure, that this won't become nullptr?

OrbitGl/CallStackDataView.h
florian-kuebler: Maybe just use a copy, or an optional? and avoid pointers at all?

florian-kuebler (commented):
Thanks for the change! I would avoid to store the raw pointer of a callstack, but otherwise looks good.

OrbitGl/CallStackDataView.h
reichlfl: Thanks for pointing out the mailthread, Dimitry. Maybe we should discuss this offline - I have a different opinion on the use of shared / weak ptr, but I'm happy to discuss and change my mind. Let's touch it in the daily?

OrbitGl/CallStackDataView.h
IrinaShkviro: It becomes nullptr only if we change callstack_data, but we should clear CallstackDataView when changing the callstack_data

OrbitGl/CallStackDataView.h
IrinaShkviro: Agree, let's discuss it in daily today

florian-kuebler: This is still only one callstack, right? So we would not loose much if we would just copy it here and don't have a pointer at all.

OrbitGl/SamplingReport.cpp
```diff
@@ -134,7 +134,7 @@ void SamplingReport::OnCallstackIndexChanged(size_t index) {
   if (index < selected_sorted_callstack_report_->callstacks_count.size()) {
     const CallstackCount& cs = selected_sorted_callstack_report_->callstacks_count[index];
     selected_callstack_index_ = index;
-    callstack_data_view_->SetCallStack(callstack_data_->GetCallStack(cs.callstack_id));
+    callstack_data_view_->SetCallStack(*callstack_data_->GetCallStack(cs.callstack_id));
```
reichlfl: This was changed and reverted again, is this intended?

reichlfl (approved):
Small comment, otherwise LGTM, thanks!

OrbitGl/SamplingReport.cpp
IrinaShkviro: Yes, I'll squash these 2 commits, so it shouldn't affect the history (with bare pointer I think it make sense to have `const CallStack* callstack` in the SetCallStack, but as we create a copy I would prefer to pass a const ref as an argument).

florian-kuebler (approved):
LGTM. Thanks!

PR #1070, opened 2020-08-19
Unit tests

MariaSandru27:


OrbitService/UtilsTest.cpp
```diff
@@ -152,4 +155,66 @@ TEST(Utils, ReadFileToString) {
   }
 }
 
+TEST(LinuxUtils, CategoriesTracepoints) {
+  using orbit_grpc_protos::TracepointInfo;
+  if (getuid() != 0) {
+    GTEST_SKIP() << "test is supported with root uid";
```
danielfenner, line 161: That's nice, I wasn't aware of GTEST_SKIP.

OrbitService/UtilsTest.cpp
```diff
@@ -152,4 +155,66 @@ TEST(Utils, ReadFileToString) {
...
+
+  const auto tracepoint_infos = utils::ReadTracepoints();
+
+  std::deque<std::string> names;
+  std::transform(tracepoint_infos.value().cbegin(), tracepoint_infos.value().cend(),
+                 std::back_inserter(names),
+                 [](const TracepointInfo& value) { return value.name(); });
+
+  ASSERT_TRUE(names.size() > 0);
+  static const std::array<std::string, 10> names_available = {
```
danielfenner, line 204: kNamesAvailable, similar below

OrbitService/Utils.cpp
```diff
@@ -138,13 +139,31 @@ ErrorMessageOr<std::vector<ModuleInfo>> ParseMaps(std::string_view proc_maps_dat
...
   std::vector<TracepointInfo> result;
 
-  for (const auto& category : fs::directory_iterator(kLinuxTracingEvents)) {
-    if (fs::is_directory(category)) {
-      for (const auto& name : fs::directory_iterator(category)) {
-        TracepointInfo tracepoint_info;
-        tracepoint_info.set_name(fs::path(name).filename());
-        tracepoint_info.set_category(fs::path(category).filename());
-        result.emplace_back(tracepoint_info);
+  struct stat info;
```
danielfenner, line 142: That looks a bit like the 90s are back. What was wrong with the std::filesystem based solution?

OrbitService/UtilsTest.cpp
MariaSandru27: Actually, I found out that because I added GTEST_SKIP I no longer need to deal with the filesystem's exceptions since the code is not executed if it is not run with sudo and therefore there are not going to be any exceptions.

danielfenner: as discussed offline .... guard against exceptions in real use anyway...

OrbitService/Utils.cpp
```diff
@@ -7,6 +7,8 @@
 #include <absl/base/casts.h>
 #include <absl/strings/str_split.h>
 #include <cxxabi.h>
+#include <dirent.h>
+#include <stdio.h>
```
dimitry-, line 11: I think you do not need these, or at least dirent

OrbitService/Utils.cpp
```diff
@@ -26,13 +28,14 @@
...
 using ::ElfUtils::ElfFile;
 using orbit_grpc_protos::ModuleInfo;
 using orbit_grpc_protos::TracepointInfo;
 
 namespace {
 
-namespace fs = std::filesystem;
 const char* kLinuxTracingEvents = "/sys/kernel/debug/tracing/events/";
 
 ErrorMessageOr<uint64_t> FileSize(const std::string& file_path) {
```
dimitry-, line 41: As daniel said - I would still modify this code to use error_code variants for std::filesystem

OrbitService/UtilsTest.cpp
```diff
@@ -152,4 +155,66 @@ TEST(Utils, ReadFileToString) {
...
+    ASSERT_TRUE(find(categories.begin(), categories.end(), category_unavailable) ==
+                categories.end());
+  }
+}
+
+TEST(LinuxUtils, NamesTracepoints) {
+  using orbit_grpc_protos::TracepointInfo;
+
+  if (getuid() != 0) {
+    GTEST_SKIP() << "test is supported with root uid";
```
dimitry-, line 193: "This test requires root uid" maybe...

OrbitService/Utils.cpp
```diff
@@ -139,7 +142,8 @@ ErrorMessageOr<std::vector<ModuleInfo>> ParseMaps(std::string_view proc_maps_dat
 ErrorMessageOr<std::vector<orbit_grpc_protos::TracepointInfo>> ReadTracepoints() {
   std::vector<TracepointInfo> result;
 
-  for (const auto& category : fs::directory_iterator(kLinuxTracingEvents)) {
+  for (const auto& category : fs::directory_iterator(
+           kLinuxTracingEvents, std::filesystem::directory_options::skip_permission_denied)) {
```
dimitry-, line 146: https://en.cppreference.com/w/cpp/filesystem/directory_iterator/directory_iterator
construct the iterator using (4) or (5), if it is end() - check the error code and return an error message.

I am also not sure if we want to skip permission denied here.

OrbitService/UtilsTest.cpp
```diff
@@ -1,13 +1,15 @@
 // Copyright (c) 2020 The Orbit Authors. All rights reserved.
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
-
-#include <absl/strings/str_format.h>
-#include <gmock/gmock-matchers.h>
-#include <gtest/gtest.h>
+//
```
dimitry-, line 4: nit:  remove this line

OrbitService/Utils.cpp
```diff
@@ -139,7 +142,8 @@ ErrorMessageOr<std::vector<ModuleInfo>> ParseMaps(std::string_view proc_maps_dat
...
     if (fs::is_directory(category)) {
       for (const auto& name : fs::directory_iterator(category)) {
```
dimitry-, line 148: same here

PR #1071, opened 2020-08-19
Call Initialize() on DataViewPanels for thread sampling reports

reichlfl:
This fixes the missing intialization of the "Refresh" buttons: https://b.corp.google.com/issues/165000881

Tests: Run Orbit, check that the refresh buttons are correctly hidden for the main sampling report and in the selection.


antonrohr (approved):
lgtm, thanks for fixing that!

PR #1073, opened 2020-08-19
Improvements to our fuzzers

beckerhe:
This PR contains some improvements to our fuzzers:

1. `CaptureEventProcessrProcessEventsFuzzer` was moved into `OrbitCaptureClient` since `CaptureEventProcessor` now lives there.
2. `CaptureSerializerLoadFuzzer` is now based on `libprotobuf-mutator` which leads to better coverage and more efficient fuzz-tests.


OrbitClientProtos/capture_data.proto
```diff
@@ -100,3 +100,9 @@ message TimerInfo {
   uint64 timeline_hash = 11;
   repeated uint64 registers = 12;
 }
+
+// libprotobuf-mutator needs a single proto with all the data
+message CaptureSerializerFuzzerInfo {
+  CaptureInfo capture_info = 2;
+  repeated TimerInfo timers = 3;
```
IrinaShkviro, line 107: Should `CaptureHeader header = 1;` be the first line?
or if not the message probably should be
```
CaptureInfo capture_info = 1;
repeated TimerInfo timers = 2;
```

IrinaShkviro (approved):
LGTM, just one question about the `CaptureSerializerFuzzerInfo` message

OrbitClientProtos/capture_data.proto
beckerhe: Good point. I tried with the header there, but that does not give any benefit. I will change the numbers. This can be easily changed later on, since there are no compatibility requirements.

beckerhe: done.

PR #1075, opened 2020-08-19
Move Visual Studio Code configuration files

antonrohr:
Move content of .vscode to /contrib/.vscode. Also add .vscode to .gitignore.
Add a note about configuration files to DEVELOPMENT.md


.gitignore
```diff
@@ -4,3 +4,4 @@ build*/
 .clangd/
 compile_commands.json
 corpora/
+/.vscode/
```
dimitry-, line 7: /?

antonrohr: this only ignores .vscode in the root folder and not in every folder. I think this is a good idea. 
We could also ignore all .vscode folder, then I would need to rename /contrib/.vscode to /contrib/vscode

beckerhe: I'm fine with having the slash here. But I would argue for renaming `/contrib/.vscode` to `contrib/vscode`. Otherwise the directory will be hidden by default which is not what we want, I guess?

antonrohr: I renamed it to `vscode`, but left the gitignore as is

kisenshi:
Great, thanks!

kisenshi (approved):

beckerhe (approved):
Thanks. Looks good to me. But can you wait for Dimitry to get back regarding the `/`. Maybe he had something else in mind that we didn't see.

dimitry- (approved):

antonrohr:
> Thanks. Looks good to me. But can you wait for Dimitry to get back regarding the /. Maybe he had something else in mind that we didn't see.

@dimitry- Anything?

.gitignore
dimitry-: got it - maybe we want to add some more / in this file :)

antonrohr: different PR :)

PR #1076, opened 2020-08-19
UI Updates

reichlfl:


reichlfl:
UI polishing update.

This PR moves most of the icons that were created in code into QT Resource Files (.qrc) and directly links them in the UI files. I also moved some dynamically created parts of the capture toolbar into the UI file.

I've added a bunch of icons for menu actions and updated the "refresh" icon to make it consistent, and replaced the "<>" arrows on buttons with icons.

Overview:
image
image
image
image
image

The "Presets" icon might be up for discussion since it's officially a "bookmark" icon, WDYT?

I also cleaned up a few texts in the UI:
* Tabs start with a capital letter
* The actions which invoke dialogs end with "..."
* Renamed "Save capture" to "Save capture as..." since this is what it does.

OrbitQt/CMakeLists.txt
```diff
@@ -61,7 +61,8 @@ target_sources(
           processlauncherwidget.cpp
           servicedeploymanager.cpp
           TopDownViewItemModel.cpp
-          topdownwidget.cpp)
+          topdownwidget.cpp
+		  ../icons/orbiticons.qrc)
```
beckerhe, line 65: It seems some tabs slipped into the file.

OrbitQt/orbitmainwindow.cpp
```diff
@@ -201,50 +201,28 @@ static QWidget* CreateSpacer(QWidget* parent) {
...
-static QAction* CreateDummyAction(const QIcon& icon, QObject* parent) {
+[[nodiscard]] static QAction* CreateDummyAction(const QIcon& icon, QObject* parent) {
   auto* action = new QAction(icon, "", parent);
   action->setDisabled(true);
   return action;
 }
 
-static QIcon GetIcon(const std::string& icon_name) {
+[[nodiscard]] static QIcon GetIcon(const std::string& icon_name) {
   return QIcon(Path::JoinPath({Path::GetIconsPath(), icon_name}).c_str());
```
beckerhe, line 211: *wish-mode on* :innocent: : Since you're revamping the icons, it would be great to change this function such that it reads the icons from the embedded resource. As far as I've seen you're embedding all icons anyway.

But this is extending the scope quite a bit, so feel free to ignore that.

Bit if you decide to do that then the icons-deployment-logic can/has to be removed. There is a copy target in `OrbitQt/CMakeLists.txt` and a `self.copy(...)`-command in conanfile.py's `package`-routine. Searching for `icons` in the mentioned files should point you to the right spots.

beckerhe (approved):
Very nice. :-) Thanks.

OrbitQt/orbitmainwindow.cpp
reichlfl: > _wish-mode on_ 😇 : Since you're revamping the icons, it would be great to change this function such that it reads the icons from the embedded resource. As far as I've seen you're embedding all icons anyway.
> 
> But this is extending the scope quite a bit, so feel free to ignore that.
> 
> Bit if you decide to do that then the icons-deployment-logic can/has to be removed. There is a copy target in `OrbitQt/CMakeLists.txt` and a `self.copy(...)`-command in conanfile.py's `package`-routine. Searching for `icons` in the mentioned files should point you to the right spots.

I wanted to do that, then I saw that it's a bite more of an effort and decided to leave it for now, let's have a look at that in the future.

pierricgimmig:
@reichlfl , this is beautiful! Thanks for that!

> The "Presets" icon might be up for discussion since it's officially a "bookmark" icon, WDYT?

I think it's perfect. Even conceptually, it's as if we were bookmarking functions to instrument.

OrbitQt/orbitmainwindow.cpp
beckerhe: sounds good! :+1:

OrbitQt/CMakeLists.txt
reichlfl: Fixed

PR #1077, opened 2020-08-19
Move GTargetProcess to DataManager and CaptureData

florian-kuebler:
The global Capture::GTargetProcess was used for two purposes:
1. Get/Add Functions/Modules for the currently selected process.
2. Get/add Functions/Modules for the process from the previous capture.

This explicitly distinguishes these cases by moving the selected
process to the DataManager and when starting a capture, store the
selected process in capture data.
This will also fix symbol loading.

Follow Up: Remove outside usage of the Mutex inside Process. And
  later completely remove Process (replace by ProcessInfo).

Test: Compile
Bug: http://b/163020637 and http://b/165562622

### DIFF
--- a/OrbitClientGgp/ClientGgp.cpp
+++ b/OrbitClientGgp/ClientGgp.cpp
@@ -84,8 +84,6 @@ bool ClientGgp::StopCapture() {
 void ClientGgp::InitCapture() {
   target_process_ = GetOrbitProcessByPid(options_.capture_pid);
   CHECK(target_process_ != nullptr);
-  // TODO: remove this line when GTargetProcess is deprecated in Capture
-  Capture::SetTargetProcess(target_process_);
 }
 
 // CaptureListener implementation
--- a/OrbitClientGgp/ClientGgp.cpp
+++ b/OrbitClientGgp/ClientGgp.cpp
@@ -52,7 +52,7 @@ bool ClientGgp::InitClient() {
 
 // Client requests to start the capture
 bool ClientGgp::RequestStartCapture(ThreadPool* thread_pool) {
-  int32_t pid = target_process_->GetID();
+  int32_t pid = target_process_->GetId();
   if (pid == -1) {
     ERROR(
         "Error starting capture: "
--- a/OrbitClientGgp/ClientGgp.cpp
+++ b/OrbitClientGgp/ClientGgp.cpp
@@ -65,7 +65,8 @@ bool ClientGgp::RequestStartCapture(ThreadPool* thread_pool) {
   // TODO(kuebler): right now selected_functions is only an empty placeholder,
   //  it needs to be filled separately in each client and then passed.
   absl::flat_hash_map<uint64_t, orbit_client_protos::FunctionInfo> selected_functions;
-  Capture::capture_data_ = CaptureData(pid, target_process_->GetName(), selected_functions);
+  Capture::capture_data_ =
+      CaptureData(pid, target_process_->GetName(), target_process_, selected_functions);
 
   ErrorMessageOr<void> result = capture_client_->StartCapture(thread_pool, pid, selected_functions);
   if (result.has_error()) {
@@ -81,7 +82,6 @@ bool ClientGgp::StopCapture() {
 }
 
 void ClientGgp::InitCapture() {
-  Capture::Init();
   target_process_ = GetOrbitProcessByPid(options_.capture_pid);
   CHECK(target_process_ != nullptr);
   // TODO: remove this line when GTargetProcess is deprecated in Capture
--- a/OrbitCore/Capture.cpp
+++ b/OrbitCore/Capture.cpp
@@ -6,15 +6,8 @@
 
 #include <ostream>
 
-#include "EventBuffer.h"
-#include "FunctionUtils.h"
-#include "OrbitBase/Logging.h"
-#include "Path.h"
-#include "Pdb.h"
 #include "SamplingProfiler.h"
 #include "absl/strings/str_format.h"
-
-using orbit_client_protos::FunctionInfo;
 
 CaptureData Capture::capture_data_;
 
--- a/OrbitCore/Capture.cpp
+++ b/OrbitCore/Capture.cpp
@@ -19,15 +19,9 @@ using orbit_client_protos::FunctionInfo;
 CaptureData Capture::capture_data_;
 
 std::shared_ptr<SamplingProfiler> Capture::GSamplingProfiler = nullptr;
-std::shared_ptr<Process> Capture::GTargetProcess = nullptr;
-
-void Capture::Init() { GTargetProcess = std::make_shared<Process>(); }
 
 void Capture::SetTargetProcess(std::shared_ptr<Process> process) {
-  if (process != GTargetProcess) {
-    GSamplingProfiler = std::make_shared<SamplingProfiler>(process);
-    GTargetProcess = std::move(process);
-  }
+  GSamplingProfiler = std::make_shared<SamplingProfiler>(std::move(process));
 }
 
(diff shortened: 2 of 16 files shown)


dimitry- (approved):

OrbitGl/App.cpp
```diff
@@ -893,7 +893,7 @@ void OrbitApp::UpdateProcessAndModuleList(
       // To this point all data is ready. We can set the Process and then
       // propagate the changes to the UI.
 
-      if (pid != GetSelectedProcessID()) {
+      if (pid != GetSelectedProcess()->GetID()) {
```
dimitry-, line 896: hmm.. I am kind of confused. How is GetSelectedProcessId different?

OrbitGl/App.h
```diff
@@ -208,6 +209,9 @@ class OrbitApp final : public DataViewFactory, public CaptureListener {
   [[nodiscard]] int32_t GetSelectedProcessID() const {
```
dimitry-, line 209: ok, is this used anywhere? should it be renamed?

florian-kuebler: Yes, and maybe, but may not.

OrbitClientGgp/ClientGgp.cpp
```diff
@@ -81,7 +82,6 @@ bool ClientGgp::StopCapture() {
 }
 
 void ClientGgp::InitCapture() {
-  Capture::Init();
   target_process_ = GetOrbitProcessByPid(options_.capture_pid);
   CHECK(target_process_ != nullptr);
   // TODO: remove this line when GTargetProcess is deprecated in Capture
```
kisenshi, line 87:
```suggestion
```

kisenshi: As Capture::SetTargetProcess(target_process_); is not needed to be removed in the end this comment is not necessary anymore.

kisenshi (approved):
Checked the code related to ClientGgp and the changes make sense. I just added a minor suggestion to remove a deprecated comment as a result of this change. Thanks!

OrbitClientGgp/ClientGgp.cpp
florian-kuebler: Well `Capture::SetTargetProcess(target_process_);` will be gone completely (hopefully by the end of tomorrow ;)).
As you do not rely on GSamplingProfiler, I will remove the comment and also the call to SetTargetProcess.

kisenshi: Ok, perfect then, thanks!

OrbitGl/App.cpp
```diff
@@ -605,7 +605,7 @@ void OrbitApp::ClearCapture() {
   SelectTextBox(nullptr);
 
   // Trigger deallocation of previous sampling related data.
-  auto empty_sampling_profiler = std::make_shared<SamplingProfiler>(Capture::GTargetProcess);
+  auto empty_sampling_profiler = std::make_shared<SamplingProfiler>(std::make_shared<Process>());
```
pierricgimmig, line 608: Hi Florian, this breaks sampling, we don't have symbols anymore in the sampling report. We need to pass in the target process to the sampling profiler. I think this is related to the fact that ClearCapture is called after we start a catpure, which is a bit weird.  To repro, load symbols, take a capture, and see that there are no symbols in the sampling report.

pierricgimmig: This seems to fix the issue:
auto empty_sampling_profiler = std::make_shared<SamplingProfiler>(GetSelectedProcess());

OrbitClientGgp/ClientGgp.cpp
```diff
@@ -65,7 +65,8 @@ bool ClientGgp::RequestStartCapture(ThreadPool* thread_pool) {
   // TODO(kuebler): right now selected_functions is only an empty placeholder,
   //  it needs to be filled separately in each client and then passed.
   absl::flat_hash_map<uint64_t, orbit_client_protos::FunctionInfo> selected_functions;
-  Capture::capture_data_ = CaptureData(pid, target_process_->GetName(), selected_functions);
+  Capture::capture_data_ =
+      CaptureData(pid, target_process_->GetName(), target_process_, selected_functions);
```
pierricgimmig, line 69: Could we pass in only the target_process_ and get the name in the constructor? Or does it make sense to have the ability to specify a different name?

florian-kuebler: As we are working towards removing OrbitProcess completely, I wanted to avoid this. But anyways, yes I can just pass this, if you wish to.

OrbitGl/App.cpp
florian-kuebler: Okay, that is then super suspicious. I'll fix it with the GetSelectedProcess for now and will have a closer look into why this is the case.

florian-kuebler: It for now uses GetSelectedProcess (which is equivalent to the prior usage of GTargetProcess).
I guess, we should move the assignment of the new SamplingProfiler to StartCapture though.

OrbitCore/Capture.cpp
```diff
@@ -19,15 +19,9 @@ using orbit_client_protos::FunctionInfo;
...
 
 void Capture::SetTargetProcess(std::shared_ptr<Process> process) {
-  if (process != GTargetProcess) {
-    GSamplingProfiler = std::make_shared<SamplingProfiler>(process);
-    GTargetProcess = std::move(process);
-  }
+  GSamplingProfiler = std::make_shared<SamplingProfiler>(std::move(process));
 }
 
 void Capture::FinalizeCapture() {
```
antonrohr, line 27: nit: when you are touching this, remove unused header and using statements

dimitry- (approved):

OrbitCore/OrbitProcess.cpp
```diff
@@ -14,7 +14,7 @@
 using orbit_client_protos::FunctionInfo;
 
 Process::Process() {
-  m_ID = 0;
+  process_id_ = 0;
```
antonrohr, line 17: nit: maybe just `id` ? 
Its a little verbose now: `process.process_id`.

antonrohr: also: You are setting this here to `0`; In OrbitProcess.h its set to `-1`.

OrbitGl/DataManager.h
```diff
@@ -50,6 +54,7 @@ class DataManager final {
   absl::flat_hash_set<uint64_t> visible_functions_;
   int32_t selected_thread_id_ = -1;
   const TextBox* selected_text_box_ = nullptr;
+  std::shared_ptr<Process> selected_process_ = std::make_shared<Process>();
```
antonrohr, line 57: nit: maybe also a TODO comment that we want to get rid of "OrbitProcess" soon. That is the plan right? It feels somewhat wrong to use OrbitProcess in DataManager, but I understand it is necessary.

antonrohr (approved):

OrbitCore/OrbitProcess.cpp
florian-kuebler: Nice, spot. Renamed and made it also -1 here.

OrbitGl/DataManager.h
florian-kuebler: I added a todo.

OrbitCore/Capture.cpp
florian-kuebler: Done.

OrbitGl/App.cpp
pierricgimmig: Assigning the new sampling profiler in ClearCapture makes sense as it frees memory when a user presses the clear capture button, I would leave it here.

PR #1078, opened 2020-08-19
Integration testing tracepoint events retrieval

MariaSandru27:


OrbitService/Utils.cpp
```diff
@@ -136,22 +139,30 @@ ErrorMessageOr<std::vector<ModuleInfo>> ParseMaps(std::string_view proc_maps_dat
...
         result.emplace_back(tracepoint_info);
       }
     }
   }
+
   return result;
 }
 
+ErrorMessageOr<std::vector<orbit_grpc_protos::TracepointInfo>> ReadTracepoints() {
+  OUTCOME_TRY(result, ListTracepoints());
```
danielfenner, line 162: I might have caused some confusion here: OUTCOME_TRY is not catching any exceptions it is merely forwarding errors from "below" in a meaningful way. So there is no need to use it here. Also wrapping ListTracepoints in ReadTracepoints makes no sense. See above for a proper solution.

OrbitService/Utils.cpp
```diff
@@ -136,22 +139,30 @@ ErrorMessageOr<std::vector<ModuleInfo>> ParseMaps(std::string_view proc_maps_dat
...
-      for (const auto& name : fs::directory_iterator(category)) {
+      for (const auto& name : fs::directory_iterator(category, error_code_name)) {
         TracepointInfo tracepoint_info;
         tracepoint_info.set_name(fs::path(name).filename());
         tracepoint_info.set_category(fs::path(category).filename());
         result.emplace_back(tracepoint_info);
       }
     }
   }
+
```
danielfenner, line 157: You avoid std::filesystem::directory_iterator throwing exceptions, that's fine. Something like the code below should do for reporting the actual errors. 

 ```
  std::error_code no_err;
  if (error_code_category != no_err || error_code_name != no_err) {
    return ErrorMessage("Naaaah ...");
  }
```

danielfenner (approved):

PR #1079, opened 2020-08-19
Fix callstack view when switch between sampling and selection tabs

IrinaShkviro:
CallstackDataView was common for both `sampling` and `selection` tabs,
and after switching the tabs CallstackDataView showed the callstack from
the previously selected tab. This commit introduces new DataViewType for
`selection` callstack, and adds corresponding view for it.

Test: take a capture, make a selection, switch from selection to samping
tab and vice versa.

Bug: http://b/165613158

### DIFF
--- a/OrbitGl/App.h
+++ b/OrbitGl/App.h
@@ -200,7 +200,8 @@ class OrbitApp final : public DataViewFactory, public CaptureListener {
 
   void CrashOrbitService(orbit_grpc_protos::CrashOrbitServiceRequest_CrashType crash_type);
 
-  DataView* GetOrCreateDataView(DataViewType type) override;
+  [[nodiscard]] DataView* GetOrCreateDataView(DataViewType type) override;
+  [[nodiscard]] DataView* GetOrCreateSelectionCallstackDataView();
 
   [[nodiscard]] ProcessManager* GetProcessManager() { return process_manager_.get(); }
   [[nodiscard]] ThreadPool* GetThreadPool() { return thread_pool_.get(); }
@@ -273,6 +274,7 @@ class OrbitApp final : public DataViewFactory, public CaptureListener {
   std::unique_ptr<FunctionsDataView> functions_data_view_;
   std::unique_ptr<LiveFunctionsDataView> live_functions_data_view_;
   std::unique_ptr<CallStackDataView> callstack_data_view_;
+  std::unique_ptr<CallStackDataView> selection_callstack_data_view_;
   std::unique_ptr<PresetsDataView> presets_data_view_;
 
   CaptureWindow* capture_window_ = nullptr;
--- a/OrbitGl/App.h
+++ b/OrbitGl/App.h
@@ -19,6 +19,7 @@
 #include "CallstackData.h"
 #include "CaptureWindow.h"
 #include "DataManager.h"
+#include "DataView.h"
 #include "DataViewFactory.h"
 #include "DataViewTypes.h"
 #include "DisassemblyReport.h"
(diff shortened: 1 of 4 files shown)


florian-kuebler (approved):
LGTM, but I guess you should wait for an additional approval, as I am not that familiar with that part of the code.

OrbitGl/CallStackDataView.cpp
```diff
@@ -13,7 +13,7 @@
 
 using orbit_client_protos::FunctionInfo;
 
-CallStackDataView::CallStackDataView() : DataView(DataViewType::kCallstack), callstack_(nullptr) {}
+CallStackDataView::CallStackDataView(DataViewType type) : DataView(type) {}
```
reichlfl, line 16: I'm not too happy with this change, even though I get that it's consistent. I'm fine with leaving it like that since it's more of a conceptual problem with the DataViewFactory, but let me put my thoughts down here nevertheless.

If I have a factory, I expect it to create objects of different types and behaviors (if only slighty) when I pass in different enum value. So when I look at this, I expect a different type of object being constructed for kCallstack and kSelectionCallstack, or at least an object that is initialized in a different way, resulting in - even if slightly - different behavior. This is not the case the Callstack - here it is basically used to create two instances of the same type. For the CallStackDataView, it is completely irrelevant that it is of type kSelectionCallStack. 

As I said, bit of a higher-level conceptual problem - to me, it does not make sense to have a factory with a "GetOrCreate" method. A factory's functionality shouldn't be to manage singleton-like objects. 

In the end, this factory interface is implemented exactly once, in OrbitApp. It could very well be replaced with `OrbitApp::GetOrCreateProcessesView()`, `OrbitApp::GetOrCreateModulesView()`. Or with a `OrbitDataViewCollection` class that does this as a member of OrbitApp, if we want to keep the modularization.

Edit: I've just seen that not all of the DataViewTypes are even supported by the factory... all the more reason to maybe remove it alltogether?

OrbitGl/SamplingReportDataView.cpp
```diff
@@ -223,7 +223,8 @@ void SamplingReportDataView::OnSelect(int index) {
 }
 
 void SamplingReportDataView::LinkDataView(DataView* data_view) {
-  if (data_view->GetType() == DataViewType::kCallstack) {
+  if (data_view->GetType() == DataViewType::kCallstack ||
+      data_view->GetType() == DataViewType::kSelectionCallstack) {
```
reichlfl, line 227: Another reason to strengthen my point above :-)

reichlfl (approved):
LGTM, even though I have some concerns with the factory in general as I've stated below. Removing the factory would make the data view management easier to read and understand, so we may want to tackle this at some point, but it's not super critical I guess.

IrinaShkviro:
> LGTM, even though I have some concerns with the factory in general as I've stated below. Removing the factory would make the data view management easier to read and understand, so we may want to tackle this at some point, but it's not super critical I guess.

Thanks a lot for the suggestion! As discussed, I created a separate method for selection_callstack_data_view_, and used the same type as for callstack_data_view_. Please take another look

OrbitGl/App.h
```diff
@@ -200,6 +201,7 @@ class OrbitApp final : public DataViewFactory, public CaptureListener {
   void CrashOrbitService(orbit_grpc_protos::CrashOrbitServiceRequest_CrashType crash_type);
 
   DataView* GetOrCreateDataView(DataViewType type) override;
+  DataView* GetOrCreateSelectionCallstackDataView();
```
reichlfl, line 204: [[nodiscard]]

reichlfl: Maybe add it for the method above as well since you're at it

reichlfl (approved):
LGTM, thanks for the change!

OrbitGl/App.h
IrinaShkviro: Yes, thanks! Fixed it

PR #1081, opened 2020-08-20
Clear callstack and selection views

IrinaShkviro:
Fixed the previous PR, please have another look.

Store CallStack value in the CallStackDataView instead of
std::unique_ptr<>, remove extra FireRefreshCallbacks call in StopCapture
and clear CallStack and Selection views when clear capture or start a
new one.

Test: Take a capture, make a selection, clear capture and check the
content of the sampling, selection and callstack views; take a capture,
make a selection, start a new capture and check the content of the
sampling, selection and callstack views while capture in process.

Bug: http://b/164971132

### DIFF
--- a/OrbitGl/App.cpp
+++ b/OrbitGl/App.cpp
@@ -89,6 +89,8 @@ void OrbitApp::OnCaptureStarted() {
       select_live_tab_callback_();
     }
 
+    FireRefreshCallbacks();
+
     absl::MutexLock lock(&mutex);
     initialization_complete = true;
   });
@@ -383,6 +385,10 @@ void OrbitApp::AddSamplingReport(std::shared_ptr<SamplingProfiler> sampling_prof
     sampling_reports_callback_(callstack_data_view, report);
   }
 
+  // clear old sampling report
+  if (sampling_report_ != nullptr) {
+    sampling_report_->ClearReport();
+  }
   sampling_report_ = report;
 }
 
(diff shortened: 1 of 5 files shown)


OrbitGl/App.cpp
```diff
@@ -607,14 +615,19 @@ void OrbitApp::ClearCapture() {
   AddTopDownView(*empty_sampling_profiler);
   Capture::GSamplingProfiler = empty_sampling_profiler;
 
+  if (selection_report_) {
+    auto empty_selection_profiler = std::make_shared<SamplingProfiler>(Capture::GTargetProcess);
```
florian-kuebler, line 619: If #1077 will be in before this change, it will silently lead to compilation error, as GTargetProcess will be gone. Otherwise, I'll take care of updating #1077.

florian-kuebler (approved):
LGTM. Thanks!

OrbitGl/App.cpp
IrinaShkviro: I can wait for your PR and rebase then

OrbitGl/App.cpp
```diff
@@ -383,6 +385,10 @@ void OrbitApp::AddSamplingReport(std::shared_ptr<SamplingProfiler> sampling_prof
     sampling_reports_callback_(callstack_data_view, report);
   }
 
+  // clear old sampling report
```
dimitry-, line 388: why is this needed? maybe add a comment explaining?

IrinaShkviro: The main reason for this is clearing CallStackDataView (the field `CallStackDataView* callstack_data_view_;` in the SamplingReport). Should I write more detailed comment here?

PR #1082, opened 2020-08-20
Fix crash when module is not found

pierricgimmig:


florian-kuebler (commented):
Hi Pierric, thanks for the change. 
#1077 will properly fix this, so there is no need for the null check. 
TBH, I would avoid introducing that check here. If we are in the situation that modules shown are not present in the current process, something really serious is going on and the data is totally inconsistent (as this is the case here).
So I would see that broken state bubbling up as soon as possible.

PR #1085, opened 2020-08-20
Fight flakyness in ThreadPoolTest.

danielfenner:
A sleep caused issues. We now busy wait for up to 10 seconds
(which is in this context an excessive amount of time). The average 
case will still be fast.
This is still not a great solution; I just hope it is good enought.

Bug: 165598808
Test: ./OrbitBaseTests.exe --gtest_repeat=100 | grep -i "Failed"

### DIFF
--- a/OrbitBase/ThreadPoolTest.cpp
+++ b/OrbitBase/ThreadPoolTest.cpp
@@ -196,10 +196,14 @@ TEST(ThreadPool, ExtendThreadPool) {
         << "actions_executed=" << actions_executed << ", expected 12";
   }
 
-  // Wait for ttl+10 and check that ThreadPool has reduced
-  // number of worker threads
-  absl::SleepFor(absl::Milliseconds(kThreadTtlMillis + 10));
-  EXPECT_EQ(thread_pool->GetPoolSize(), kThreadPoolMinSize);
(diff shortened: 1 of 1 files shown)


OrbitBase/ThreadPoolTest.cpp
```diff
@@ -196,10 +196,14 @@ TEST(ThreadPool, ExtendThreadPool) {
...
-  absl::SleepFor(absl::Milliseconds(kThreadTtlMillis + 10));
-  EXPECT_EQ(thread_pool->GetPoolSize(), kThreadPoolMinSize);
+  // Check that ThreadPool will reduce the number of worker threads.
+  int loops = 0;
+  constexpr int kMaxLoops = 1000;
+  while (thread_pool->GetPoolSize() > kThreadPoolMinSize && loops < kMaxLoops) {
+    absl::SleepFor(absl::Milliseconds(10));
+    loops++;
+  }
+  EXPECT_LT(loops, kMaxLoops);
```
dimitry-, line 206: I think it is better to check thread_pool->GetPoolSize here... there is a (very) slight possibility that it will succeed.

dimitry- (approved):
otherwise lgtm

danielfenner:
We agreed to instead disable the test on the CI.

PR #1087, opened 2020-08-20
Move GSamplingProfiler to CaptureData

florian-kuebler:
This moves the global Capture::GSamplingProfiler to CaptureData.
By being tight to the CaptureData (of the previous capture), lifetime
is more explicit.

Follow Up:
1. Make the shared_ptr in CaptureData a value type.
2. Move CaptureData to DataManager.
3. Make SamplingProfiler stateless (requires updates of SamplingReport).

Test: Compile, Take Capture, Load Symbols (before and after). Check all
 views. Load a Capture.
Bug: http://b/163020637

### DIFF
--- a/OrbitCaptureClient/CaptureEventProcessorProcessEventsFuzzer.cpp
+++ b/OrbitCaptureClient/CaptureEventProcessorProcessEventsFuzzer.cpp
@@ -21,7 +21,8 @@ using orbit_client_protos::TimerInfo;
 
 class MyCaptureListener : public CaptureListener {
  private:
-  void OnCaptureStarted() override {}
+  void OnCaptureStarted(
+      int32_t, const absl::flat_hash_map<uint64_t, orbit_client_protos::FunctionInfo>&) override {}
   void OnCaptureComplete() override {}
   void OnTimer(const TimerInfo&) override {}
   void OnKeyAndString(uint64_t, std::string) override {}
--- a/OrbitCaptureClient/CaptureEventProcessorProcessEventsFuzzer.cpp
+++ b/OrbitCaptureClient/CaptureEventProcessorProcessEventsFuzzer.cpp
@@ -22,7 +22,9 @@ using orbit_client_protos::TimerInfo;
 class MyCaptureListener : public CaptureListener {
  private:
   void OnCaptureStarted(
-      int32_t, const absl::flat_hash_map<uint64_t, orbit_client_protos::FunctionInfo>&) override {}
+      int32_t /*process_id*/,
+      const absl::flat_hash_map<
+          uint64_t, orbit_client_protos::FunctionInfo>& /*selected_functions*/) override {}
   void OnCaptureComplete() override {}
   void OnTimer(const TimerInfo&) override {}
   void OnKeyAndString(uint64_t, std::string) override {}
--- a/OrbitClientGgp/ClientGgp.cpp
+++ b/OrbitClientGgp/ClientGgp.cpp
@@ -65,8 +65,6 @@ bool ClientGgp::RequestStartCapture(ThreadPool* thread_pool) {
   // TODO(kuebler): right now selected_functions is only an empty placeholder,
   //  it needs to be filled separately in each client and then passed.
   absl::flat_hash_map<uint64_t, orbit_client_protos::FunctionInfo> selected_functions;
-  Capture::capture_data_ =
-      CaptureData(pid, target_process_->GetName(), target_process_, selected_functions);
 
   ErrorMessageOr<void> result = capture_client_->StartCapture(thread_pool, pid, selected_functions);
   if (result.has_error()) {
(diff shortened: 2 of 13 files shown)


OrbitCore/CaptureData.h
```diff
@@ -23,7 +24,8 @@ class CaptureData {
         process_{std::move(process)},
         selected_functions_{std::move(selected_functions)},
         callstack_data_(std::make_unique<CallstackData>()),
-        selection_callstack_data_(std::make_unique<CallstackData>()) {
+        selection_callstack_data_(std::make_unique<CallstackData>()),
+        sampling_profiler_{std::make_shared<SamplingProfiler>(process_)} {
```
dimitry-, line 28: Is there a reason why this is not in DataManager? It seems like this should live separately, even though it uses the capture data and is tied to capture state (it uses capture process not selected process).

florian-kuebler: The idea was that the lifetime is completely bound by the one of capture data (at least in theory).

dimitry-: On the other hand we do not want to serialize it. And the lifetime is actually a bit different - the sampling profiler data can only be calculated once capture is done.

dimitry-: discussed offline - decided to keep it here because we want it to be freed when capture data is deallocated.

OrbitCore/CaptureData.h
```diff
@@ -125,6 +129,10 @@ class CaptureData {
 
   [[nodiscard]] const std::shared_ptr<Process>& process() const { return process_; }
 
+  [[nodiscard]] const std::shared_ptr<SamplingProfiler>& sampling_profiler() const {
```
dimitry-, line 132: should this be a const object? std::shared_ptr<const SamplingProfiler>?

dimitry-: or have 2 functions - one that returns const pointer/const reference another (non-const) which returns shared_ptr.

dimitry- (approved):
otherwise lgtm

OrbitCore/CaptureData.h
florian-kuebler: I added a second getter returning a const ref SamplingProfiler

OrbitGl/App.cpp
```diff
@@ -612,12 +610,9 @@ void OrbitApp::ClearCapture() {
   set_selected_thread_id(-1);
   SelectTextBox(nullptr);
 
-  // Trigger deallocation of previous sampling related data.
-  // TODO(kuebler): Investigate why selected process needs to be passed here.
-  auto empty_sampling_profiler = std::make_shared<SamplingProfiler>(GetSelectedProcess());
-  AddSamplingReport(empty_sampling_profiler, nullptr);
```
IrinaShkviro, line 618: Why it was removed?

OrbitCore/CaptureData.h
IrinaShkviro: I think that if you're using this function to update profiler it shouldn't be `const`. Also I would prefer to use 
`[[nodiscard]] SamplingProfiler* sampling_profiler()` or 
`[[nodiscard]] std::shared_ptr<SamplingProfiler> sampling_profiler()` if you need shared_ptr.

OrbitGl/App.cpp
florian-kuebler: Sorry, this was unintended

OrbitCore/CaptureData.h
florian-kuebler: I'll make it a non-const ref, as it is also stored into the sampling report (and I want avoid storing raw pointer here).
In the future, this will be a value type and the second getter will go away.

florian-kuebler:
@kisenshi Added you as a reviewer FYI.

OrbitCore/CaptureData.h
florian-kuebler: actually made it returning the shared_ptr as value.

OrbitGl/App.cpp
florian-kuebler: For the record, I now moved the initialization of capture data to OnCaptureStarted and also clear capture data on ClearCapture, so that this is again safe and correct here.

OrbitClientGgp/ClientGgp.cpp
```diff
@@ -65,8 +65,6 @@ bool ClientGgp::RequestStartCapture(ThreadPool* thread_pool) {
   // TODO(kuebler): right now selected_functions is only an empty placeholder,
   //  it needs to be filled separately in each client and then passed.
   absl::flat_hash_map<uint64_t, orbit_client_protos::FunctionInfo> selected_functions;
-  Capture::capture_data_ =
```
kisenshi, line 68: Why is this removed? I still need Capture and capture_data for processing the data of the capture (which is not done yet but it is the next step)

OrbitCore/CaptureData.h
```diff
@@ -125,6 +129,12 @@ class CaptureData {
 
   [[nodiscard]] const std::shared_ptr<Process>& process() const { return process_; }
 
+  [[nodiscard]] std::shared_ptr<SamplingProfiler> sampling_profiler() const {
```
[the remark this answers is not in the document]
florian-kuebler: Done

OrbitCaptureClient/CaptureEventProcessorProcessEventsFuzzer.cpp
```diff
@@ -21,7 +21,8 @@ using orbit_client_protos::TimerInfo;
 
 class MyCaptureListener : public CaptureListener {
  private:
-  void OnCaptureStarted() override {}
+  void OnCaptureStarted(
+      int32_t, const absl::flat_hash_map<uint64_t, orbit_client_protos::FunctionInfo>&) override {}
```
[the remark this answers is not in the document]
florian-kuebler: Thanks. Yes added the comments.

kisenshi (approved):
The changes in the ggp client look good to me, thanks!

IrinaShkviro (approved):
Lgtm, thanks for the change!

PR #1089, opened 2020-08-20
Fix `build_and_upload_dependencies.sh`

beckerhe:
Due to recent changes to the docker containers and due to more tests
some adjustments to the dependency build scripts were neded.

1. Fix the conan version to one exact version since we aren't currently
compatible to the most recent one.
2. Set the `QT_QPA_PLATFORM` variable such that the OrbitQt tests can
run in a headless environment.


IrinaShkviro (approved):

beckerhe:
Thanks, Irina!

PR #1091, opened 2020-08-20
Improve bootstrap-orbit.sh                                                      

beckerhe:
1. The script won't try to install debian packages when not on a debian         
based system.                                                                   
2. The script won't try to add universe as a apt-repository if not on           
Ubuntu.                                                                         
3. The script won't request root-rights for installing packages if the          
required packages are already installed. (This can be checked without           
root-rights.)                                                                   
4. The script will install the required conan version, not an arbitrary         
version. That's required since we currently don't support the latest            
version.                                                                        
5. The script now supports all the command line options which can be            
passed to the `third_party/conan/configs/install.sh`-script. This is            
needed for running bootstrap in a specific way inside of oss-fuzz`s             
docker container.

### DIFF
--- a/bootstrap-orbit.sh
+++ b/bootstrap-orbit.sh
@@ -6,26 +6,78 @@
 
 DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
 
-echo "Installing required system dependencies..."
-sudo add-apt-repository universe
-if [ $? -ne 0 ]; then
-  sudo apt-get install -y software-properties-common
-  sudo add-apt-repository universe
-fi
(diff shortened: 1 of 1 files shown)


bootstrap-orbit.sh
```diff
@@ -6,26 +6,70 @@
...
+        sudo add-apt-repository universe
+        if [ $? -ne 0 ]; then
+          sudo apt-get install -y software-properties-common
+          sudo add-apt-repository universe
+        fi
+      fi
+
+      sudo apt-get update || exit $?
+      sudo apt-get install -y ${REQUIRED_PACKAGES[@]} || exit $?
+    fi
```
dimitry-, line 55: maybe extract this to a separate function?

dimitry- (approved):
otherwise lgtm

bootstrap-orbit.sh
beckerhe: good idea. done.

PR #1092, opened 2020-08-20
Remove Samplingprofiler default constructor

antonrohr:
This is not needed anymore and removes one other usage of the OrbitProcess constructor. Unless I missed something.


florian-kuebler (approved):
Thanks!

PR #1093, opened 2020-08-20
Fix border areas drawn incorrectly in capture

karupayun:
Because of problems with z-alignment and wrong starting point of a
margin, some border parts are drawn incorrectly.

- We define new constants for z-values for margins and slider (related
with the problem about the right-down border area)
- We fix a wrong value for the right margin (related with the problem
with the top-right border area)

Before:
Screen Shot 2020-08-20 at 17 18 40
Screen Shot 2020-08-20 at 17 18 33

After:
Screen Shot 2020-08-20 at 17 34 58
Screen Shot 2020-08-20 at 17 35 04


reichlfl (approved):
LGTM, thanks for fixing!
Thinkng about it, the issue with clicking through the horizontal slider on the bottom (https://b.corp.google.com/issues/162485723) may be quite similar, or might be fixed by the update of the Z-Values as well?

reichlfl:
Oh, I missed the state of the checks during review - seems Z_VALUE_EVENT_BAR_PICKING was still used in another place? Please make sure it compiles on your local machine before sending it to review.

karupayun:
Oh, sorry. Two consecutive mistakes. First didn't see the error message, and after that I thought that I have pushed the new commit but seems that no. I didn't mean to erase the old constant, adding again.

OrbitGl/CaptureWindow.cpp
```diff
@@ -537,8 +536,8 @@ void CaptureWindow::DrawScreenSpace() {
   const Color kBackgroundColor(70, 70, 70, 255);
   float margin_x1 = getWidth();
   float margin_x0 = margin_x1 - vertical_margin;
-
-  Box box(Vec2(margin_x0, 0), Vec2(margin_x1 - margin_x0, canvasHeight - height), z);
+  Box box(Vec2(margin_x0, time_graph_.GetLayout().GetSliderWidth()),
+          Vec2(margin_x1 - margin_x0, canvasHeight - height), GlCanvas::Z_VALUE_MARGIN);
```
pierricgimmig, line 540: Good catch! Thanks. 

nit. With the z-ordering fixed, we could simplify further with: 

Box box(Vec2(margin_x0, 0), Vec2(margin_x1 - margin_x0, canvasHeight), z);

It would eliminate the potential seam between the margin and the slider. But this is not a huge deal, feel free to disregard.

karupayun:
yes, @reichlfl  . Seems that bug is fixed by this proper alignment (given to the slider a greater z position than the events).

PR #1094, opened 2020-08-20
Fix graph track rendering

pierricgimmig:
Properly use std::map's lower_bound() method which 
returns the first element that is not less than the key


OrbitGl/GraphTrack.cpp
```diff
@@ -55,6 +55,7 @@ void GraphTrack::Draw(GlCanvas* canvas, PickingMode picking_mode) {
 
   auto it = values_.lower_bound(min_ns);
   if (it == values_.end()) return;
+  if (it != values_.begin()) --it;
```
dimitry-, line 58: I am not sure why you need --it here and below. the lower bound return position of the element not less than, if you want it to be strictly less maybe check here if *it is not already strictly less than min_ns? Also (not sure if this is the case but) what if *--it == it? do you want to check for that?

pierricgimmig: For a time value between two keys at t0 and t1, lower_bound returns t1, we really want t0 in order to draw the full curve. min_ns is not necessarily close to the first key btw, it's dependent on the current viewport and the distribution of the keys. Also, we don't draw all the values all the time, only the ones that are visible given the current zoom. In general, we want to access one key that is < min_ns and one key that is > max_ns to make sure we cover the entire width of the track.

How can *--it == it btw? I do check that "it" is neither end() nor begin(), so it seems safe to do --it.

PR #1096, opened 2020-08-21
Made SamplingProfiler a value member of CaptureData

florian-kuebler:
Prior to this change, SamplingProfiler was a shared_ptr that was
hold by CaptureData, but also passed and stored by the
SamplingReport view.
With this change, ownership is send to CaptureData and passed as
a copy to SamplingReport.

Test: Compile, Capture, Load Symbols,  Select Samples

### DIFF
--- a/OrbitGl/App.cpp
+++ b/OrbitGl/App.cpp
@@ -980,7 +980,7 @@ void OrbitApp::UpdateSamplingReport() {
   }
 
   if (selection_report_ != nullptr) {
-    SamplingProfiler selection_profiler = selection_report_->profiler();
+    SamplingProfiler selection_profiler(Capture::capture_data_.process());
     selection_profiler.ProcessSamples(*Capture::capture_data_.GetSelectionCallstackData());
     selection_report_->UpdateReport(std::move(selection_profiler),
                                     Capture::capture_data_.GetSelectionCallstackData());
--- a/OrbitGl/App.cpp
+++ b/OrbitGl/App.cpp
@@ -109,12 +109,11 @@ void OrbitApp::OnCaptureStarted(
 void OrbitApp::OnCaptureComplete() {
   main_thread_executor_->Schedule([this] {
     Capture::capture_data_.set_address_infos(std::move(captured_address_infos_));
-    Capture::capture_data_.sampling_profiler()->ProcessSamples(
-        *Capture::capture_data_.GetCallstackData());
+    Capture::capture_data_.UpdateSamplingProfiler();
 
     RefreshCaptureView();
 
-    AddSamplingReport(Capture::capture_data_.sampling_profiler(),
+    AddSamplingReport(Capture::capture_data_.GetSamplingProfiler(),
                       Capture::capture_data_.GetCallstackData());
     AddTopDownView(Capture::capture_data_.GetSamplingProfiler());
 
@@ -330,12 +329,12 @@ void OrbitApp::Disassemble(int32_t pid, const FunctionInfo& function) {
     disasm.AddLine(absl::StrFormat("asm: /* %s */", FunctionUtils::GetDisplayName(function)));
     disasm.Disassemble(memory.data(), memory.size(), FunctionUtils::GetAbsoluteAddress(function),
                        is_64_bit);
-    if (!sampling_report_ || !sampling_report_->GetProfiler()) {
+    if (!sampling_report_) {
       DisassemblyReport empty_report(disasm);
       SendDisassemblyToUi(disasm.GetResult(), std::move(empty_report));
       return;
     }
-    std::shared_ptr<SamplingProfiler> profiler = sampling_report_->GetProfiler();
+    const SamplingProfiler& profiler = Capture::capture_data_.GetSamplingProfiler();
 
     DisassemblyReport report(disasm, FunctionUtils::GetAbsoluteAddress(function), profiler,
                              Capture::capture_data_.GetCallstackData()->GetCallstackEventsSize());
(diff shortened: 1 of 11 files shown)


OrbitGl/App.cpp
```diff
@@ -975,11 +974,16 @@ void OrbitApp::SelectTextBox(const TextBox* text_box) {
...
-    sampling_report_->UpdateReport();
+    Capture::capture_data_.UpdateSamplingProfiler();
+    sampling_report_->UpdateReport(Capture::capture_data_.GetSamplingProfiler(),
+                                   Capture::capture_data_.GetCallstackData());
   }
 
   if (selection_report_ != nullptr) {
-    selection_report_->UpdateReport();
+    SamplingProfiler selection_profiler = selection_report_->profiler();
+    selection_profiler.ProcessSamples(*Capture::capture_data_.GetCallstackData());
```
IrinaShkviro, line 984: Maybe 
```suggestion
    selection_profiler.ProcessSamples(*Capture::capture_data_.GetSelectionCallstackData());
```

OrbitGl/CaptureSerializer.cpp
```diff
@@ -35,9 +35,9 @@ using orbit_client_protos::TimerInfo;
 
 ErrorMessageOr<void> CaptureSerializer::Save(const std::string& filename) {
   // Add selected functions' exact address to sampling profiler
-  const std::shared_ptr<SamplingProfiler>& profiler = Capture::capture_data_.sampling_profiler();
+  SamplingProfiler profiler = Capture::capture_data_.GetSamplingProfiler();
```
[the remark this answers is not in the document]
IrinaShkviro: The only line that could change saved data in `UpdateAddressInfo` is
```
address_info->set_function_name(FunctionUtils::GetDisplayName(*function));
```
And this line change function_name only if for selected address we have address_info, function, but don't have unique_callstack (otherwise we've already called this function in `ResolveCallstacks` and updated function_name).

But I agree with @florian-kuebler : if it makes sense and we want to update saved data and/or sampling_profiler, we probably want to do it when finalizing the capture and not when saving. But it changes data only if we didn't call UpdateAddressInfo function  for this address before (that called for each frame of each unique_callstacks in `ResolveCallstacks`).

OrbitGl/App.cpp
florian-kuebler: Thanks

OrbitGl/CaptureSerializer.cpp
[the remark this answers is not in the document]
florian-kuebler: >The only line that could change saved data in UpdateAddressInfo is
address_info->set_function_name(FunctionUtils::GetDisplayName(*function));

Agree, and this will never change anything, as for the selected functions, they must be loaded, thus 
```
  if (function != nullptr) {
    function_address = FunctionUtils::GetAbsoluteAddress(*function);
    function_name = FunctionUtils::GetDisplayName(*function);
```
will always happen. So address_info will not be modified. All other internal mappings are not used in saving afterwards. Thus, I removed that code completely.

OrbitGl/CaptureSerializer.cpp
```diff
@@ -35,9 +35,9 @@ using orbit_client_protos::TimerInfo;
...
   for (auto& pair : Capture::capture_data_.selected_functions()) {
-    profiler->UpdateAddressInfo(pair.first);
+    profiler.UpdateAddressInfo(pair.first);
```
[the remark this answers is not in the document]
florian-kuebler: That code is gone now.

OrbitGl/App.cpp
```diff
@@ -975,11 +974,16 @@ void OrbitApp::SelectTextBox(const TextBox* text_box) {
...
+    sampling_report_->UpdateReport(Capture::capture_data_.GetSamplingProfiler(),
+                                   Capture::capture_data_.GetCallstackData());
   }
 
   if (selection_report_ != nullptr) {
-    selection_report_->UpdateReport();
+    SamplingProfiler selection_profiler = selection_report_->profiler();
+    selection_profiler.ProcessSamples(*Capture::capture_data_.GetSelectionCallstackData());
+    selection_report_->UpdateReport(std::move(selection_profiler),
+                                    Capture::capture_data_.GetSelectionCallstackData());
```
IrinaShkviro, line 986: I think that it would be enough to create an empty profiler in first line (`ProcessSamples` clears data as the first step anyway).

OrbitGl/SamplingReport.cpp
```diff
@@ -28,41 +28,47 @@ void SamplingReport::ClearReport() {
...
     thread_report.SetThreadID(tid);
     thread_report.SetSamplingReport(this);
     thread_reports_.push_back(std::move(thread_report));
   }
+
+  // Refresh the displayed callstacks as they might not be up to date anymore,
+  // for example the number of occurrences or of total callstacks might have
+  // changed (OrbitSamplingReport::RefreshCallstackView will do the actual
+  // update once OrbitApp::FireRefreshCallbacks is called).
+  UpdateDisplayedCallstack();
```
IrinaShkviro, line 49: I don't think that we need it here: FillReport called only in constructor and selected_address_ and selected_thread_id_ are default values. It means that `UpdateDisplayedCallstack` calls `ClearReport` in this case (and it doesn't change anything as we've just created SamplingReport).

dimitry- (approved):
lgtm!

OrbitGl/App.cpp
florian-kuebler: Done. Thanks for spotting this.

OrbitGl/SamplingReport.cpp
florian-kuebler: Thanks, for the hint. I guess I was about to remove the UpdateReport method, but noticed, that it is not that easy, as I was expecting it.
Anyways, removed that call.

IrinaShkviro (approved):
Thanks!

PR #1097, opened 2020-08-21
Add module and symbol loading to OrbitClientGgp

kisenshi:
Retrieve target process data, load module corresponding to the target process binary and its symbols. Update _ClientGgp::GetOrbitProcessByPid_ and create _ClientGgp::LoadModuleAndSymbols_.

**Add _ProcessClient_ instantiation to _ClientGgp_** _ProcessClient_ is required to be able to handle and retrieve information about the modules and the selected process.

**Retrieve list of processes to set target info**  To later being able to find a module, it is needed to have more information about the process than merely its pid so it is included a call to _ProcessClient::GetProcessList_. _ClientGgp::GetOrbitProcessByPid_ has been updated to search the target process in the list of processes retrieved. Its information is populated into the _target_process__. An available target process is essential for the capture to work as expected so the program ends if not found.

**Load module and symbols**  We need to load the list of modules related to the target process to find the module that corresponds to its binary. Its symbols are also loaded. A new method _ClientGgp::LoadModuleAndSymbols_ contains the logic to load the module and its symbols. This information is needed during capture so it has been included to its initialization. After calling this method, _target_process__ has the information about the module corresponding to its binary and its symbols.

### DIFF
--- a/OrbitClientGgp/ClientGgp.cpp
+++ b/OrbitClientGgp/ClientGgp.cpp
@@ -4,11 +4,13 @@
 
 #include "ClientGgp.h"
 
+#include <condition_variable>
 #include <cstdint>
 #include <limits>
 #include <map>
 #include <memory>
 #include <string>
+#include <thread>
 
 #include "Capture.h"
 #include "OrbitBase/Logging.h"
@@ -18,14 +20,7 @@
 #include "absl/time/time.h"
 #include "capture_data.pb.h"
 
-namespace {
-std::shared_ptr<Process> GetOrbitProcessByPid(int32_t pid) {
-  std::shared_ptr<Process> process = std::make_shared<Process>();
-  // TODO: study if we need more information. Requires extra steps
-  process->SetID(pid);
-  return process;
-}
-}  // namespace
+using orbit_grpc_protos::ProcessInfo;
 
 ClientGgp::ClientGgp(ClientGgpOptions&& options) : options_(std::move(options)) {}
 
@@ -48,9 +43,18 @@ bool ClientGgp::InitClient() {
   LOG("Created GRPC channel to %s", options_.grpc_server_address);
 
   process_manager_ = ProcessManager::Create(grpc_channel_, options_.process_refresh_timeout);
+  auto callback = [this](ProcessManager*) { process_manager_ready_.notify_one(); };
+  process_manager_->SetProcessListUpdateListener(callback);
+  // wait for the process list to be updated once
+  std::unique_lock<std::mutex> lk(mutex_);
+  process_manager_ready_.wait(lk);
+  process_manager_->SetProcessListUpdateListener(nullptr);
 
   // Initialisations needed for capture to work
-  InitCapture();
+  if (!InitCapture()) {
+    ShutdownClient();
+    return false;
+  }
   capture_client_ = std::make_unique<CaptureClient>(grpc_channel_, this);
   return true;
 }
@@ -86,9 +90,39 @@ bool ClientGgp::StopCapture() {
   return capture_client_->StopCapture();
 }
 
-void ClientGgp::InitCapture() {
+std::shared_ptr<Process> ClientGgp::GetOrbitProcessByPid(int32_t pid) {
+  // We retrieve the information of the process to later get the module corresponding to its binary
+  const std::vector<ProcessInfo>& process_infos = process_manager_->GetProcessList();
+  LOG("List of processes:");
+  for (const ProcessInfo& info : process_infos) {
+    LOG("pid:%d, name:%s, path:%s, is64:%d", info.pid(), info.name(), info.full_path(),
+        info.is_64_bit());
+  }
+  auto process_it = find_if(process_infos.begin(), process_infos.end(),
+                            [&pid](const ProcessInfo& info) { return info.pid() == pid; });
+  if (process_it != process_infos.end()) {
+    LOG("Found process by pid, set target process");
+    std::shared_ptr<Process> process = std::make_shared<Process>();
+    process->SetID(process_it->pid());
+    process->SetName(process_it->name());
+    process->SetFullPath(process_it->full_path());
+    process->SetIs64Bit(process_it->is_64_bit());
+    LOG("Process info: pid:%d, name:%s, path:%s, is64:%d", process->GetID(), process->GetName(),
+        process->GetFullPath(), process->GetIs64Bit());
+    return process;
+  }
+  return nullptr;
+}
+
+bool ClientGgp::InitCapture() {
   target_process_ = GetOrbitProcessByPid(options_.capture_pid);
-  CHECK(target_process_ != nullptr);
+  if (target_process_ == nullptr) {
+    ERROR("Error: not able to set target process");
+    return false;
+  }
+  // TODO: remove this line when GTargetProcess is deprecated in Capture
+  Capture::SetTargetProcess(target_process_);
+  return true;
 }
 
 // CaptureListener implementation
--- a/OrbitClientGgp/ClientGgp.cpp
+++ b/OrbitClientGgp/ClientGgp.cpp
@@ -92,18 +92,18 @@ std::shared_ptr<Process> ClientGgp::GetOrbitProcessByPid(int32_t pid) {
   }
   auto process_it = find_if(process_infos.begin(), process_infos.end(),
                             [&pid](const ProcessInfo& info) { return info.pid() == pid; });
-  if (process_it != process_infos.end()) {
-    LOG("Found process by pid, set target process");
-    std::shared_ptr<Process> process = std::make_shared<Process>();
-    process->SetID(process_it->pid());
-    process->SetName(process_it->name());
-    process->SetFullPath(process_it->full_path());
-    process->SetIs64Bit(process_it->is_64_bit());
-    LOG("Process info: pid:%d, name:%s, path:%s, is64:%d", process->GetId(), process->GetName(),
-        process->GetFullPath(), process->GetIs64Bit());
-    return process;
+  if (process_it == process_infos.end()) {
+    return nullptr;
   }
-  return nullptr;
+  LOG("Found process by pid, set target process");
+  std::shared_ptr<Process> process = std::make_shared<Process>();
+  process->SetID(process_it->pid());
+  process->SetName(process_it->name());
+  process->SetFullPath(process_it->full_path());
+  process->SetIs64Bit(process_it->is_64_bit());
+  LOG("Process info: pid:%d, name:%s, path:%s, is64:%d", process->GetId(), process->GetName(),
+      process->GetFullPath(), process->GetIs64Bit());
+  return process;
 }
 
 bool ClientGgp::LoadModuleAndSymbols() {
@@ -122,10 +122,11 @@ bool ClientGgp::LoadModuleAndSymbols() {
         info.file_path(), info.file_size(), info.address_start(), info.address_end(),
         info.build_id());
   }
-  std::string_view target_name = target_process_->GetName();
-  auto module_it =
-      find_if(module_infos.begin(), module_infos.end(),
-              [&target_name](const ModuleInfo& info) { return info.name() == target_name; });
+  std::string_view main_executable_name = target_process_->GetName();
+  auto module_it = find_if(module_infos.begin(), module_infos.end(),
+                           [&main_executable_name](const ModuleInfo& info) {
+                             return info.name() == main_executable_name;
+                           });
   if (module_it == module_infos.end()) {
     ERROR("Error: Module correspondent to process binary not found");
     return false;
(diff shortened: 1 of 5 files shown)


kisenshi:
Updated code to use ProcessClient instead of ProcessManager, everything is ready now for review.

OrbitClientGgp/ClientGgp.cpp
```diff
@@ -48,9 +43,18 @@ bool ClientGgp::InitClient() {
   LOG("Created GRPC channel to %s", options_.grpc_server_address);
 
   process_manager_ = ProcessManager::Create(grpc_channel_, options_.process_refresh_timeout);
+  auto callback = [this](ProcessManager*) { process_manager_ready_.notify_one(); };
+  process_manager_->SetProcessListUpdateListener(callback);
+  // wait for the process list to be updated once
+  std::unique_lock<std::mutex> lk(mutex_);
+  process_manager_ready_.wait(lk);
```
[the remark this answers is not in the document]
kisenshi: This is not needed anymore now that ProcessClient is used instead

OrbitClientGgp/ClientGgp.cpp
```diff
@@ -18,14 +20,7 @@
...
 
-namespace {
-std::shared_ptr<Process> GetOrbitProcessByPid(int32_t pid) {
-  std::shared_ptr<Process> process = std::make_shared<Process>();
-  // TODO: study if we need more information. Requires extra steps
-  process->SetID(pid);
-  return process;
-}
-}  // namespace
+using orbit_grpc_protos::ProcessInfo;
```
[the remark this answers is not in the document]
kisenshi: Done!

OrbitClientGgp/ClientGgp.cpp
```diff
@@ -79,9 +78,105 @@ bool ClientGgp::StopCapture() {
...
+  }
+  const std::vector<ProcessInfo>& process_infos = result_process_infos.value();
+  LOG("List of processes:");
+  for (const ProcessInfo& info : process_infos) {
+    LOG("pid:%d, name:%s, path:%s, is64:%d", info.pid(), info.name(), info.full_path(),
+        info.is_64_bit());
+  }
+  auto process_it = find_if(process_infos.begin(), process_infos.end(),
+                            [&pid](const ProcessInfo& info) { return info.pid() == pid; });
+  if (process_it != process_infos.end()) {
```
dimitry-, line 95: if (process_it == process_infos.end()) {
  return nullptr;
}

...

OrbitClientGgp/ClientGgp.cpp
```diff
@@ -79,9 +78,105 @@ bool ClientGgp::StopCapture() {
...
+  }
+  // Find module that corresponds to the binary of target_process
+  const std::vector<ModuleInfo>& module_infos = result_module_infos.value();
+  LOG("List of modules");
+  for (const ModuleInfo& info : module_infos) {
+    LOG("name:%s, path:%s, size:%d, address_start:%d. address_end:%d, build_id:%s", info.name(),
+        info.file_path(), info.file_size(), info.address_start(), info.address_end(),
+        info.build_id());
+  }
+  std::string_view target_name = target_process_->GetName();
```
dimitry-, line 125: process_name or main_executable_name?

OrbitClientGgp/ClientGgp.cpp
```diff
@@ -79,9 +78,105 @@ bool ClientGgp::StopCapture() {
...
+  module->m_DebugSignature = module_it->build_id();
+  target_process_->AddModule(module);
+  LOG("Module info: name:%s, path:%s, size:%d, address_start:%d. address_end:%d, build_id:%s",
+      module->m_Name, module->m_FullName, module->m_PdbSize, module->m_AddressStart,
+      module->m_AddressEnd, module->m_DebugSignature);
+
+  // Load symbols for the module
+  const std::string& module_path = module->m_FullName;
+  LOG("Looking for debug info file for %s", module_path);
+  ErrorMessageOr<std::string> result_debug_file = process_client_->FindDebugInfoFile(module_path);
```
dimitry-, line 149: main_executable_debug_file?

OrbitClientGgp/ClientGgp.cpp
```diff
@@ -79,9 +78,105 @@ bool ClientGgp::StopCapture() {
...
+    process->SetFullPath(process_it->full_path());
+    process->SetIs64Bit(process_it->is_64_bit());
+    LOG("Process info: pid:%d, name:%s, path:%s, is64:%d", process->GetId(), process->GetName(),
+        process->GetFullPath(), process->GetIs64Bit());
+    return process;
+  }
+  return nullptr;
+}
+
+bool ClientGgp::LoadModuleAndSymbols() {
```
dimitry-, line 109: question - why do you need symbols?

OrbitClientGgp/ClientGgp.cpp
kisenshi: this makes a lot of sense, will update it, thanks!

OrbitClientGgp/ClientGgp.cpp
kisenshi: Better names indeed; will update to use main_executable_name, thanks!

OrbitClientGgp/ClientGgp.cpp
kisenshi: Makes sense, especially after the other name change, will update, thanks!

OrbitClientGgp/ClientGgp.cpp
kisenshi: To be able to hook functions to a capture. I am working on that on a separate PR because this one was big enough already

dimitry-: are these symbols limited to symbols from the main executable?

kisenshi: at the moment, for simplicity, yes; if it is extended later I will update the code accordingly

dimitry- (approved):
lgtm

OrbitClientGgp/ClientGgp.cpp
kisenshi: Done

OrbitClientGgp/ClientGgp.cpp
kisenshi: Done

OrbitClientGgp/ClientGgp.cpp
kisenshi: Done, I also have updated the name of _debug_file_path_ to _main_executable_debug_file_ which is related with this one.

OrbitClientGgp/ClientGgp.cpp
```diff
@@ -79,9 +78,107 @@ bool ClientGgp::StopCapture() {
...
+std::shared_ptr<Process> ClientGgp::GetOrbitProcessByPid(int32_t pid) {
+  // We retrieve the information of the process to later get the module corresponding to its binary
+  ErrorMessageOr<std::vector<ProcessInfo>> result_process_infos = process_client_->GetProcessList();
+  if (result_process_infos.has_error()) {
+    return nullptr;
+  }
+  const std::vector<ProcessInfo>& process_infos = result_process_infos.value();
+  LOG("List of processes:");
+  for (const ProcessInfo& info : process_infos) {
+    LOG("pid:%d, name:%s, path:%s, is64:%d", info.pid(), info.name(), info.full_path(),
```
danielfenner, line 90: Just as a remark: This seems to be a lot of logging. Keep as is if you think it makes sense.

OrbitClientGgp/ClientGgp.cpp
danielfenner: libggp and vulkan might make sense later. I agree that this should be added when needed.

danielfenner (approved):
LGTM!

OrbitClientGgp/ClientGgp.cpp
```diff
@@ -79,9 +78,107 @@ bool ClientGgp::StopCapture() {
...
+  for (const ModuleInfo& info : module_infos) {
+    LOG("name:%s, path:%s, size:%d, address_start:%d. address_end:%d, build_id:%s", info.name(),
+        info.file_path(), info.file_size(), info.address_start(), info.address_end(),
+        info.build_id());
+  }
+  std::string_view main_executable_name = target_process_->GetName();
+  auto module_it = find_if(module_infos.begin(), module_infos.end(),
+                           [&main_executable_name](const ModuleInfo& info) {
+                             return info.name() == main_executable_name;
+                           });
```
antonrohr, line 129: You have to be careful here. A name of a process does not necessarily correspond to the name of the executable (elf file). The name can be set to something arbitrary. (It comes from `/proc/[pid]/comm` see here: https://man7.org/linux/man-pages/man5/proc.5.html)


 I would highly recommend using the path to match with modules instead of name.

OrbitClientGgp/ClientGgp.cpp
```diff
@@ -79,9 +78,107 @@ bool ClientGgp::StopCapture() {
...
+      module->m_AddressEnd, module->m_DebugSignature);
+
+  // Load symbols for the module
+  const std::string& module_path = module->m_FullName;
+  LOG("Looking for debug info file for %s", module_path);
+  ErrorMessageOr<std::string> result_main_executable_debug_file =
+      process_client_->FindDebugInfoFile(module_path);
+  if (result_main_executable_debug_file.has_error()) {
+    ERROR("Error loading symbols: %s", result_main_executable_debug_file.error().message());
+    return false;
```
antonrohr, line 154: Same OUTCOME_TRY as earlier

OrbitClientGgp/ClientGgp.cpp
```diff
@@ -79,9 +78,107 @@ bool ClientGgp::StopCapture() {
...
+  }
+  const std::string& main_executable_debug_file = result_main_executable_debug_file.value();
+  LOG("Found file: %s", main_executable_debug_file);
+  LOG("Loading symbols");
+  ErrorMessageOr<orbit_grpc_protos::ModuleSymbols> result_symbols =
+      symbol_helper_.LoadSymbolsFromFile(main_executable_debug_file, module->m_DebugSignature);
```
antonrohr, line 160: Do I understand this correctly, this code (ClientGgp) is also executed on the instance? This loading will not work when this is run on a workstation.

OrbitClientGgp/ClientGgp.cpp
kisenshi: At the beginning I added this LOG with the idea of removing it later once the code was working but I realized it could be quite helpful and informative in case something is wrong so I decided to keep it in the end. E.g. the selected process is not found; could be that they provided the wrong pid so they get aware of the reason by checking the logs; could be the process is not running at all, etc

OrbitClientGgp/ClientGgp.cpp
```diff
@@ -79,9 +78,107 @@ bool ClientGgp::StopCapture() {
...
+bool ClientGgp::LoadModuleAndSymbols() {
+  // Load modules for target_process_
+  ErrorMessageOr<std::vector<ModuleInfo>> result_module_infos =
+      process_client_->LoadModuleList(target_process_->GetId());
+  if (result_module_infos.has_error()) {
+    ERROR("Error retrieving modules: %s", result_module_infos.error().message());
+    return false;
+  }
+  // Find module that corresponds to the binary of target_process
+  const std::vector<ModuleInfo>& module_infos = result_module_infos.value();
```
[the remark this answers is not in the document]
kisenshi: I like this suggestion; I could even do the same with the GetOrbitProcessByPid method! I'm gonna check if the error info of the method called is informative enough and update the code if it makes sense. Thanks!

OrbitClientGgp/ClientGgp.cpp
kisenshi: Oh, I see! Thanks for raising this, I will update it

OrbitClientGgp/ClientGgp.cpp
kisenshi: Yes exactly, ClientGgp runs on the instance

kisenshi:
I have updated the code as suggested to:

1. Use the path when looking for the module for the binary instead of the name 
2. Use OUTCOME_TRY in LoadModuleAndSymbols and GetOrbitProcessByPid as it makes the code much simpler and the error message that it was included was not really necessary; the error messages returned are informative enough. 
I realised that the same modification could be done to InitCapture now but I decided not to update it to use OUTCOME_TRY because I think for that case it makes sense to notify if the error was caused by not being able to set the target process or when loading the modules and symbols.

Thanks!

OrbitClientGgp/ClientGgp.cpp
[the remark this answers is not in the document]

OrbitClientGgp/ClientGgp.cpp

OrbitClientGgp/ClientGgp.cpp
kisenshi: Done!

OrbitClientGgp/ClientGgp.cpp
```diff
@@ -147,35 +137,28 @@ bool ClientGgp::LoadModuleAndSymbols() {
...
+  return outcome::success();
 }
 
 bool ClientGgp::InitCapture() {
-  target_process_ = GetOrbitProcessByPid(options_.capture_pid);
-  if (target_process_ == nullptr) {
-    ERROR("Error: not able to set target process");
+  ErrorMessageOr<std::shared_ptr<Process>> target_process_result =
+      GetOrbitProcessByPid(options_.capture_pid);
+  if (target_process_result.has_error()) {
```
antonrohr, line 153: Nit: just so you know, the check `if(!target_process_result)` would also work. Probably not worth the change, but just in case you did not know.

antonrohr (approved):
thanks for adding the changes, it looks very good now!

OrbitClientGgp/ClientGgp.cpp
kisenshi: Ah, I didn't know, thanks!

kisenshi:
Thanks! merging into master :)

PR #1100, opened 2020-08-21
Add option to bypass the release check

beckerhe:
We have a check which forbids to finish a Kokoro-triggered release build
when the current branch is not on a tag.

For testing purposes or in exceptional situations it makes sense to
overwrite that behaviour.

This commits adds an environment variable which disabled the check. This
environment variable can be set when the job is triggered from Fusion.

### DIFF
--- a/kokoro/builds/build.sh
+++ b/kokoro/builds/build.sh
@@ -85,6 +85,7 @@ if [ -n "$1" ]; then
 
   set +e
   if [ "$BUILD_TYPE" = "release" ] \
+     && [ -z "${ORBIT_BYPASS_RELEASE_CHECK}" ] \
      && ! git -C "${REPO_ROOT}" describe --tags --exact-match > /dev/null; then
     echo -n "We are currently conducting a release build, but we aren't on a tag."
     echo    " Aborting the build..."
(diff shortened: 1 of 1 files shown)


antonrohr (approved):
lgtm
Just throwing some ideas how this could be done slightly different, maybe you like one of them. (I actually do not have enough insight to judge what makes the most sense) 
* Did you consider instead using the variable BUILD_TYPE ? It could have the value "release_without_checks". 
* Since the only check is whether the branch is on a git tag or not, consider naming the variable RELEASE_WITHOUT_GIT_TAG_CHECK to be more verbose.

beckerhe:
> lgtm
> Just throwing some ideas how this could be done slightly different, maybe you like one of them. (I actually do not have enough insight to judge what makes the most sense)
> 
> * Did you consider instead using the variable BUILD_TYPE ? It could have the value "release_without_checks".
> * Since the only check is whether the branch is on a git tag or not, consider naming the variable RELEASE_WITHOUT_GIT_TAG_CHECK to be more verbose.

Regarding the first: I can't change the BUILD_TYPE variable when manually triggering a job, so this won't work.

Regarding the second: I considered different namings and I came to the conclusion I don't wanna bind the name to a specific check. I want to have a name which resembles the semantics not the implementation. It's mainly because I'm already thinking about additional checks which should also be disabled by this one flag. Per-check-flags might be useful but currently I don't see a use-case for that.

Thank you though for thinking about it! :-)

PR #1108, opened 2020-08-24
Fix drawing problem in rounding corners

karupayun:
Related with [b/162754926](url).

Before this commit, tracks were drawn outside of the rounded corners.

Before: 
Screen Shot 2020-08-24 at 12 44 03
before-1

After:
Screen Shot 2020-08-24 at 12 55 05
Screen Shot 2020-08-24 at 12 53 41


reichlfl (approved):
Nice catch, thanks! LGTM.

General comment (which you don't need to address, but keep in mind for the future): I feel the amount of z-values is getting a bit out of hand. I'm quite sure we could reduce these to a few if we gave it some thought, and those could potentially be relative for clarity, example:

Z_VALUE_UI = 0.1
Z_VALUE_SCROLLBAR_BG = Z_VALUE_UI + 0.1
Z_VALUE_SCROLLBAR_BAR = Z_VALUE_SCROLLBAR_BG + 0.1

karupayun:
Hi Florian, thanks for the review! 

Yes, agree with you. It took a few time understand why these constants are defined in that way, but still not understand 100% of them. I leaved as it is, because of a future probable change, but will talk with Pierric about all of them.

PR #1110, opened 2020-08-24
Align track titles with their respective triangles

karupayun:
Related to[ b/164961039](url).
Track titles y-coordinates were drawn independently of those of the
triangles. After vertical zoom, this difference meant that they weren't
aligned. This PR solved that, causing both to draw similarly.

Florian's image:
Mon Aug 17 2020 10_49_30 GMT+0200 (Central European Summer Time)

Now:
Screen Shot 2020-08-24 at 13 50 31
Screen Shot 2020-08-24 at 13 50 11

### DIFF
--- a/OrbitGl/Track.cpp
+++ b/OrbitGl/Track.cpp
@@ -141,17 +141,19 @@ void Track::Draw(GlCanvas* canvas, PickingMode picking_mode) {
 
   // Draw collapsing triangle.
   float button_offset = layout.GetCollapseButtonOffset();
-  Vec2 toggle_pos = Vec2(tab_x0 + button_offset, m_Pos[1] + half_label_height);
+  float toggle_y_pos = m_Pos[1] + half_label_height;
+  Vec2 toggle_pos = Vec2(tab_x0 + button_offset, toggle_y_pos);
   collapse_toggle_->SetPos(toggle_pos);
   collapse_toggle_->Draw(canvas, picking_mode);
 
   if (!picking) {
     // Draw label.
     float label_offset_x = layout.GetTrackLabelOffsetX();
-    float label_offset_y = layout.GetTrackLabelOffsetY();
+    // Vertical offset for the text to be aligned to the center of the triangle.
+    float label_offset_y = GCurrentTimeGraph->GetFontSize() / 3.f;
     const Color kTextWhite(255, 255, 255, 255);
-    canvas->AddText(label_.c_str(), tab_x0 + label_offset_x, y1 + label_offset_y + m_Size[1],
-                    text_z, kTextWhite, label_width - label_offset_x);
+    canvas->AddText(label_.c_str(), tab_x0 + label_offset_x, toggle_y_pos - label_offset_y, text_z,
+                    kTextWhite, label_width - label_offset_x);
   }
 
   m_Canvas = canvas;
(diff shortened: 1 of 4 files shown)


OrbitGl/Track.cpp
```diff
@@ -150,8 +151,8 @@ void Track::Draw(GlCanvas* canvas, PickingMode picking_mode) {
     float label_offset_x = layout.GetTrackLabelOffsetX();
     float label_offset_y = layout.GetTrackLabelOffsetY();
     const Color kTextWhite(255, 255, 255, 255);
-    canvas->AddText(label_.c_str(), tab_x0 + label_offset_x, y1 + label_offset_y + m_Size[1],
-                    text_z, kTextWhite, label_width - label_offset_x);
+    canvas->AddText(label_.c_str(), tab_x0 + label_offset_x, toggle_y_pos - label_offset_y, text_z,
```
reichlfl, line 154: I think this is abusing the TrackLabelOffset - in the end, you want the label to be vertically centered at the center of the triangle, and you're hardcoding this into the label offset. This will be quite unexpected if someone tries to adjust the layout. Can we easily get the height of the drawn label and then v-center it so this would also adjust to changed font sizes?

karupayun: TrackLabelOffsetY was only for that. I erased it and write something which is related with the font size. Now, if you change the font size is still aligned. BTW, this 3.5 isn't very nice, but is the best that I can find.

Anyway, I have a TODO about scaling the text and triangle sizes.

OrbitGl/Track.cpp
```diff
@@ -141,17 +142,19 @@ void Track::Draw(GlCanvas* canvas, PickingMode picking_mode) {
...
+  Vec2 toggle_pos = Vec2(tab_x0 + button_offset, toggle_y_pos);
   collapse_toggle_->SetPos(toggle_pos);
   collapse_toggle_->Draw(canvas, picking_mode);
 
   if (!picking) {
     // Draw label.
     float label_offset_x = layout.GetTrackLabelOffsetX();
-    float label_offset_y = layout.GetTrackLabelOffsetY();
+    // Vertical offset for the text to be aligned to the center of the triangle.
+    float label_offset_y = GParams.font_size / 3.5;
```
pierricgimmig, line 154: This 3.5 value seems arbitrary. As discussed, this doesn't seem work properly when changing the font size.

karupayun: After our talk, I started to look why this 3.5 doesn't work for you, and realized that I wasn't getting the updated font_size if you modify directly in the menu as you were making.

So, I made a GetFontSize() in Timegraph and now it is always updated. BTW, I modify this 3.5 to 3, no theorical explanation for that. With 3, the triangle is aligned with upper-case letters, and with 3.5 the triangle is aligned with lower-cases ones. As 3 in more elegant, I leaved 3 there.

OrbitGl/Track.cpp
karupayun: I erased the TrackLabelOffsetY, and make something related with the font size. 

BTW, It's a temporal solution and didn't want to spend more time here. In the future we might want to modify font_size and triangle size when vertical-zooming. I didn't think too much how to modify track font-size without modifying the one in the TimeBar. Seems a little mess, but maybe it's easier than I expected.

PR #1111, opened 2020-08-24
Use shared_ptr to pass Pickable

dimitry-:
Using weak_ptr implies additional cost of having to check
if the value expired on every call in every function. This
is not necessary if we pass shared_ptr instead. In that case
the only place we have to check is when getting Pickable
from the container. After that we can lock it and pass
back to the caller.

Also changed memcpy to absl::bit_cast in ColorFromPickingID

Test: run OrbitGlTests

### DIFF
--- a/OrbitGl/PickingManager.cpp
+++ b/OrbitGl/PickingManager.cpp
@@ -7,24 +7,21 @@
 #include "OpenGl.h"
 #include "OrbitBase/Logging.h"
 
-PickingId PickingManager::GetOrCreatePickableId(std::weak_ptr<Pickable> pickable,
+PickingId PickingManager::GetOrCreatePickableId(std::shared_ptr<Pickable> pickable,
                                                 BatcherId batcher_id) {
-  auto locked_pickable = pickable.lock();
-  CHECK(locked_pickable != nullptr);
-
   absl::MutexLock lock(&mutex_);
   uint32_t pickable_id = 0;
 
-  auto it = pickable_pid_map_.find(locked_pickable.get());
-  if (it != pickable_pid_map_.end()) {
-    pickable_id = it->second;
-  } else {
+  auto it = pickable_pid_map_.find(pickable.get());
+  if (it == pickable_pid_map_.end()) {
     pickable_id = ++pickable_id_counter_;
     pid_pickable_map_[pickable_id] = pickable;
-    pickable_pid_map_[locked_pickable.get()] = pickable_id;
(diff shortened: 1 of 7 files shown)


OrbitGl/PickingManager.cpp
```diff
@@ -7,21 +7,16 @@
...
-  } else {
+  auto it = pickable_pid_map_.find(pickable.get());
+  if (it == pickable_pid_map_.end()) {
     pickable_id = ++pickable_id_counter_;
     pid_pickable_map_[pickable_id] = pickable;
-    pickable_pid_map_[locked_pickable.get()] = pickable_id;
+    pickable_pid_map_[pickable.get()] = pickable_id;
   }
 
   PickingId id = PickingId::Create(PickingType::kPickable, pickable_id_counter_, batcher_id);
```
dimitry-, line 22: I am beginning to suspect this should use pickable_id here instead of the counter

dimitry-: yep - this is a bug - added a test for it.

OrbitGl/PickingManager.cpp
```diff
@@ -83,20 +80,17 @@ bool PickingManager::IsDragging() const {
...
+  auto picked = GetPicked();
   return picked && picked.get() == pickable;
 }
 
 Color PickingManager::ColorFromPickingID(PickingId id) const {
-  static_assert(sizeof(PickingId) == 4 * sizeof(uint8_t),
-                "PickingId should be same size as 4 * uint8_t");
-  std::array<uint8_t, 4> color_values;
-  std::memcpy(&color_values[0], &id, sizeof(PickingId));
+  auto color_values = absl::bit_cast<std::array<uint8_t, 4>>(id);
```
reichlfl, line 94: Can you check if this causes any warnings during compilation on Linux? I vaguely remember that there was a reason I left the memcpy() in one place, but I might be wrong.

reichlfl (approved):
LGTM - I think I'm wrong about my one comment, but please have a quick look

OrbitGl/PickingManager.cpp
reichlfl: Right, good catch, thanks - I think it only worked by accident, because we are adding all primitives of a Pickable at once...

OrbitGl/PickingManager.cpp
dimitry-: I do not see any warnings...

PR #1113, opened 2020-08-25
Move OrbitCore/Profiling.h to OrbitBase/Profiling.h

pierricgimmig:
- Piece of the upcoming introspection revival (see #1112 for reference).
- Replace OrbitTicks() by MonotonicTimestampNs()


OrbitBase/include/OrbitBase/Profiling.h
```diff
@@ -18,9 +29,9 @@ inline void clock_gettime(uint32_t, struct timespec* spec) {
 }
 #endif
 
-inline TickType OrbitTicks(uint32_t a_Clock = 1 /*CLOCK_MONOTONIC*/) {
+inline TickType MonotonicTimestampNs() {
   timespec ts;
-  clock_gettime(a_Clock, &ts);
+  clock_gettime(CLOCK_MONOTONIC, &ts);
```
dimitry-, line 34: Why do you want to use monotonic here? If this is because of performance maybe consider using CLOCK_BOOTTIME on linux?

OrbitBase/include/OrbitBase/Profiling.h
```diff
@@ -2,10 +2,21 @@
...
+#ifdef _WIN32
+#include <Windows.h>
+using pid_t = uint32_t;
+#define CLOCK_MONOTONIC 1
+#else
+#include <sys/syscall.h>
+#include <unistd.h>
+#endif
 
 typedef uint64_t TickType;
```
dimitry-, line 21: I think we should drop this - I have a question about underlying time every time I see TickType in the code.

OrbitBase/include/OrbitBase/Profiling.h
```diff
@@ -18,9 +29,9 @@ inline void clock_gettime(uint32_t, struct timespec* spec) {
 }
 #endif
 
-inline TickType OrbitTicks(uint32_t a_Clock = 1 /*CLOCK_MONOTONIC*/) {
+inline TickType MonotonicTimestampNs() {
   timespec ts;
-  clock_gettime(a_Clock, &ts);
+  clock_gettime(CLOCK_MONOTONIC, &ts);
   return 1000000000ll * ts.tv_sec + ts.tv_nsec;
 }
```
dimitry-, line 37: maybe inline TickToNanoseconds is it used anywhere? if so change the code to use TickToDuration

The double variant also seems to be not perfect - I think we should use duration where possible.

Add unit-tests for these? The are probably trivial but better to have them.

OrbitBase/include/OrbitBase/Profiling.h
```diff
@@ -38,4 +49,11 @@ inline TickType MicrosecondsToTicks(double a_Micros) {
   return static_cast<TickType>(a_Micros * 1000);
```
dimitry-, line 49: This does not seem right - is this function used?

(if you keep functions - rename the parameters to comply with the styleguide)

OrbitBase/include/OrbitBase/Profiling.h
pierricgimmig: It's the clock used by the perfeventopen events.  I'm not changing the behavior, just shuffling code around.

OrbitBase/include/OrbitBase/Profiling.h
pierricgimmig: I agree! Will do in another PR.

pierricgimmig: Let's do it in this PR actually, this is long overdue :-)

OrbitBase/include/OrbitBase/Profiling.h
pierricgimmig: I removed TickToNanoseconds. Again this PR is simply moving code around... I agree that we should have tests and possibly use duration, but this is out of the scope of this PR.

OrbitBase/include/OrbitBase/Profiling.h
pierricgimmig: Removed the function altogether and replaced by inlined implementation which makes the code clearer at the call site anyway.

OrbitBase/include/OrbitBase/Profiling.h
```diff
@@ -0,0 +1,46 @@
...
+#include <unistd.h>
+#endif
+
+[[nodiscard]] inline uint64_t MonotonicTimestampNs() {
+  timespec ts;
+#ifdef _WIN32
+  __int64 time;
+  GetSystemTimeAsFileTime((FILETIME*)&time);
+  ts.tv_sec = time / 10000000i64;
+  ts.tv_nsec = time % 10000000i64 * 100;
```
dimitry-, line 24: return static_cast<uint64_t>(time)*100?

dimitry-: maybe use std::steady_clock::now() for windows and then get nanoseconds?

pierricgimmig: Yes, much cleaner, thanks.  For steady_clock::now, what advantage does it offer? Better precision? Are we sure that it's returning the values in the same time domain as the ETW events? I'm not against it, but it requires an investigation which is really outside the scope of this PR, which, again, is shuffling existing code around... :-)

dimitry-: just that it does not not use windows-specific api.

dimitry- (approved):

PR #1114, opened 2020-08-25
Introspection as a custom manual instrumentation implementation

pierricgimmig:
- Add ability to define a custom implementation of manual instrumentation
  functions defined in Orbit.h (define ORBIT_API_INTERNAL_IMPL).
- OrbitService and Orbit itself will use this custom implementation for introspection.
- Remove obsolete files.

NOTE: Introspection currently works end to end in #1112.  I'm breaking down the changes in multiple PRs.  As is, this PR introduces no functional change. The UI side introspection code will come in the next PR.


OrbitBase/include/OrbitBase/Tracing.h
```diff
@@ -5,63 +5,59 @@
...
-}  // namespace orbit::tracing
+using TimerCallback = std::function<void(Scope&& scope)>;
 
-#else
+class Listener {
+ public:
+  explicit Listener(const TimerCallback& callback);
+  ~Listener();
+
+  static absl::Mutex* GetMutex() { return &mutex_; }
```
dimitry-, line 52: This is global mutex... it is not clear what it guards.

OrbitBase/Tracing.cpp
```diff
@@ -0,0 +1,115 @@
...
+void TrackScope(const char* name, uint64_t value, orbit::Color color, ScopeType type) {
+  orbit::tracing::Scope scope;
+  scope.begin = MonotonicTimestampNs();
+  scope.name = name;
+  scope.color = color;
+  scope.tracked_value = value;
+  scope.tid = absl::base_internal::GetTID();
+  scope.type = type;
+
+  absl::MutexLock lock(orbit::tracing::Listener::GetMutex());
```
dimitry-, line 75: this can be function-local static mutex if this is the only place where we guard - if not, the functionality should probably be incapacitated into class owning this mutex.

OrbitBase/include/OrbitBase/Tracing.h
pierricgimmig: Hi @dimitry- , thanks for looking into this PR. I modified the code, it should be cleaner now.

pierricgimmig:
Hi @dimitry- , I changed the way scopes are handled, the callback is now called from a worker thread. It was implemented using the ThreadPool class. The idea is to defer the processing onto a worker thread to minimize contention on the instrumented threads. I'd be curious to have your thoughts on the usage of ThreadPool here. I only need a single worker by the way, the callback locks further down the pipe so having many threads won't help.  It's also nice to guarantee that callbacks are called from a single thread.

dimitry- (commented):
> Hi @dimitry- , I changed the way scopes are handled, the callback is now called from a worker thread. It was implemented using the ThreadPool class. The idea is to defer the processing onto a worker thread to minimize contention on the instrumented threads. I'd be curious to have your thoughts on the usage of ThreadPool here. I only need a single worker by the way, the callback locks further down the pipe so having many threads won't help. It's also nice to guarantee that callbacks are called from a single thread.

Looks good I am not sure how expensive is the callback and if the cost of scheduling an event (note that it also has a mutex lock, and we copy the scope object) is comparable with the cost of just calling the callback, but if you want a separate thread to process scopes this will do it.

Also can you please add tests for these?

OrbitBase/include/OrbitBase/Tracing.h
```diff
@@ -2,66 +2,52 @@
...
-#define ORBIT_CALL(f)               \
-  do {                              \
-    if (orbit::tracing::GHandler) { \
-      orbit::tracing::GHandler->f;  \
-    }                               \
-  } while (0)
+#define ORBIT_API_INTERNAL_IMPL
+// NOTE: Orbit.h will be moved to its own
+//       OrbitApi project in a subsequent PR.
+#include "../../../Orbit.h"
```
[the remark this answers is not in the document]
pierricgimmig: I'm tempted to put Orbit.h in OrbitBase actually, I think tracing is akin to logging, and we want to have it available everywhere.

pierricgimmig:
> Looks good I am not sure how expensive is the callback and if the cost of scheduling an event

Yeah same here, but I have a feeling offloading to a worker will be faster. The callback will consist of sending the data over through grpc, but it uses a lock that seems to be under high contention. The real test will be to measure it (using introspection!!). 

I'll add tests and fix the variable names tomorrow...

dimitry-:
> > Looks good I am not sure how expensive is the callback and if the cost of scheduling an event
> 
> Yeah same here, but I have a feeling offloading to a worker will be faster. The callback will consist of sending the data over through grpc, but it uses a lock that seems to be under high contention. The real test will be to measure it (using introspection!!).

I gave it a little bit more thought and it seems to like the threading should be shifted to callback implementation. The idea is that it knows better if the action is expensive or not. It is also easier to read and more visible as of why this is done on a separate thread.

Orbit.h
```diff
@@ -130,7 +130,19 @@ enum class Color : uint32_t {
 namespace orbit_api {
 
 // NOTE: Do not use these directly, use corresponding macros instead.
-ORBIT_STUB void Start(const char*, orbit::Color) { ORBIT_NOOP(); }
+#ifdef ORBIT_API_INTERNAL_IMPL
+void Start(const char* name, uint64_t dummy, orbit::Color color);
```
dpallotti: What is `dummy` for?

OrbitBase/Tracing.cpp
```diff
@@ -9,54 +9,69 @@
...
 #include "OrbitBase/Profiling.h"
+#include "OrbitBase/ThreadPool.h"
 
 using orbit::tracing::Scope;
 using orbit::tracing::ScopeType;
 using orbit::tracing::TimerCallback;
 
-thread_local std::vector<Scope> scopes;
+ABSL_CONST_INIT static absl::Mutex global_callback_mutex(absl::kConstInit);
+ABSL_CONST_INIT static TimerCallback* global_callback;
```
dpallotti: So how is this set? Other PR?

dpallotti (commented):
Looks good overall, I'll leave you to Dimitry's notes.

So for me to understand: the goal is for this to work both on the service and on the client, and they will set completely different `global_callback`s (other PR?) as the path from the introspection scope to the capture is obviously different in the two cases?

OrbitBase/Tracing.cpp
pierricgimmig: Yes, a variation of #1112. Essentially, we'll listen for introspection scopes and send them over grpc in the same way we send FunctionCalls in the collector.

pierricgimmig:
> So for me to understand: the goal is for this to work both on the service and on the client, and they will set completely different `global_callback`s (other PR?) as the path from the introspection scope to the capture is obviously different in the two cases?

Yes that is exactly right. On top of that, we are re-using the same interface (Orbit.h) as the manual instrumentation API that will be distributed to partners.

pierricgimmig:
> I gave it a little bit more thought and it seems to like the threading should be shifted to callback implementation. The idea is that it knows better if the action is expensive or not. It is also easier to read and more visible as of why this is done on a separate thread.

As discussed, since we are measuring time here, we need to make sure that the callback won't affect the measurement, that's why we are doing the work on our side to minimize the overhead of "scope" processing.

dpallotti (approved):
I guess you have some open comments from Dimitry but other than that it looks good.

Issue #1105, opened 2020-08-24
Bootstrap/build fails on latest Ubuntu with GCC 10 alternative


pwnorbitals:
Error : `ERROR: Error reading 'default_relwithdebinfo' profile: Profile not found: gcc10_relwithdebinfo` 
EDIT : Workaround by calling directly `./build.sh gcc9_relwithdebinfo ggp_relwithdebinfo`
EDIT2 : Test Utils.GetCpuUtilization fails (`Unable to determine cpu utilization, wrong format from top command.`) and test script hangs at the end

pwnorbitals:
Thanks a lot, awesome work

beckerhe:
> Thanks a lot, awesome work

Hi Chris,
I just realized I missed to answer here. But apparently compiling with GCC-10 works now for you?

Cheers,
Henning

pwnorbitals:
Yes, looks great, thanks

PR #1115, opened 2020-08-25
Support for new compilers in conan

beckerhe:
This adds support for new compiler version to conan.
GCC 10 is needed by the community (see #1105) and for Fuzzing we need clang-12 (trunk) support.

Fixes: #1105
Test: Did a clang10_release build. Went though without problems - at least on my machine.

### DIFF
--- a/third_party/conan/configs/linux/settings.yml
+++ b/third_party/conan/configs/linux/settings.yml
@@ -42,15 +42,15 @@ os:
 arch: [x86, x86_64, ppc32be, ppc32, ppc64le, ppc64, armv4, armv4i, armv5el, armv5hf, armv6, armv7, armv7hf, armv7s, armv7k, armv8, armv8_32, armv8.3, sparc, sparcv9, mips, mips64, avr, s390, s390x, asm.js, wasm, sh4le]
 compiler:
     sun-cc:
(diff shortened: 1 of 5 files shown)


third_party/conan/configs/linux/settings.yml
```diff
@@ -69,7 +69,7 @@ compiler:
     clang:
         version: ["3.3", "3.4", "3.5", "3.6", "3.7", "3.8", "3.9", "4.0",
                   "5.0", "6.0", "7.0", "7.1",
-                  "8", "9", "10"]
+                  "8", "9", "10", "11", "12"]
```
dimitry-, line 72: do we need profiles for these?

beckerhe: Eventually yes. But for current use-case, no.

I'm a bit reluctant with adding the profiles without having at least once compiled the whole project with these compilers. I have a branch which is adding clang-10 profiles, but I haven't tried that yet. But I guess this will follow in a future PR. clang-11 is not released yet. Release is scheduled for tomorrow. I might add support for 10 and 11 in one PR.

We only need 12 because oss-fuzz uses clang-master which currently identifies as clang-12.

dimitry- (approved):

PR #1116, opened 2020-08-25
Remove X11 linking

beckerhe:
It seems that we don't need to link to X11 anymore. Probably the code that needed that was removed before.

I tested that on Ubuntu 20.04 and on my corporate Linux workstation.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -73,10 +73,6 @@ list(PREPEND CMAKE_FIND_ROOT_PATH "${CMAKE_BINARY_DIR}")
 # scripts in a cross compilation environment.
 list(APPEND CMAKE_FIND_ROOT_PATH "${CONAN_CMAKE_MODULE_PATH}")
 
-if(NOT WIN32 AND WITH_GUI)
-  find_package(X11 REQUIRED)
-endif()
-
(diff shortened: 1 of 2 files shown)


dimitry-:
Apparently ggc disagrees for some reason?

/usr/bin/ld: /home/conan/.conan/data/glew/2.1.0/orbitdeps/stable/package/0ad44a1ce1c56a88a3bda770f3234e5efe9b394c/lib/libGLEW.a(glew.c.o): undefined reference to symbol 'glXGetProcAddressARB'

beckerhe:
> Apparently ggc disagrees for some reason?
> 
> /usr/bin/ld: /home/conan/.conan/data/glew/2.1.0/orbitdeps/stable/package/0ad44a1ce1c56a88a3bda770f3234e5efe9b394c/lib/libGLEW.a(glew.c.o): undefined reference to symbol 'glXGetProcAddressARB'

Interesting. This should not having anything to do with the compiler. And it seems to work with GGC9 on Ubuntu 20.04. But fails on my machine as well. I have the suspicions that this is a problem with the glew package.

Anyway, let's leave the linking against GLX for now and only remove X11-dependencies.

dimitry- (approved):

PR #1117, opened 2020-08-25
Remove Process::GetModuleFromName

antonrohr:
GetModuleFromPath is now used everywhere.

This is a step towards removing OrbitProcess. In the future modules will be identified always by their path not name, because the name can be ambiguous.

### DIFF
--- a/OrbitClientProtos/capture_data.proto
+++ b/OrbitClientProtos/capture_data.proto
@@ -56,7 +56,7 @@ message CallstackInfo {
 
 message LinuxAddressInfo {
   uint64 absolute_address = 1;
-  string module_name = 2;
+  string module_path = 2;
   string function_name = 3;
   uint64 offset_in_function = 4;
 }
--- a/OrbitGl/SamplingReportDataView.cpp
+++ b/OrbitGl/SamplingReportDataView.cpp
@@ -12,6 +12,7 @@
 #include "Capture.h"
 #include "FunctionUtils.h"
 #include "OrbitModule.h"
+#include "Path.h"
 #include "SamplingReport.h"
 #include "absl/strings/str_format.h"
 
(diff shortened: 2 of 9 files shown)


OrbitClientProtos/capture_data.proto
```diff
@@ -56,7 +56,7 @@ message CallstackInfo {
 
 message LinuxAddressInfo {
   uint64 absolute_address = 1;
-  string module_name = 2;
+  string module_name = 2; // This is actually the full module path
```
dimitry-, line 59: should we rename it?

OrbitGl/SamplingReportDataView.cpp
```diff
@@ -259,13 +265,13 @@ void SamplingReportDataView::DoFilter() {
   for (size_t i = 0; i < functions_.size(); ++i) {
     SampledFunction& func = functions_[i];
     std::string name = ToLower(func.name);
-    std::string module = ToLower(func.module);
+    std::string module_name = ToLower(Path::GetFileName(func.module_path));
```
dimitry-, line 268: is this intended? Maybe users want to search in path?

florian-kuebler: Well at least its the current behaviour.

florian-kuebler (approved):
So lgtm. Thanks.

OrbitGl/SamplingReportDataView.cpp
antonrohr: Yes, it is current behaviour, but more importantly, we only display the name and not the full path. It makes sense to me to only filter things that are displayed

OrbitClientProtos/capture_data.proto
antonrohr: I am in favor to rename it. This breaks compatibility between saved captures, correct? I am going to ask in chat if we already broke it. if yes, I will change it.

dimitry- (approved):

PR #1118, opened 2020-08-26
Fix clang-12's unnecessary-copy-warning

beckerhe:
clang-12's range-loop-construct warning complains about copied const
arguments in range-based for-loop. Apparently we have that enabled as an
error. This commit fixes the complaint. That was necessary for oss-fuzz
which uses clang-trunk.


florian-kuebler (approved):
LGTM. Thanks!

dimitry- (approved):

PR #1119, opened 2020-08-26
Fix Z-ordering issues + Refactoring

karupayun:
There were a few bugs related with Z-ordering.

This PR:
1) Move TextZ and TrackZ from Layout to Canvas.
2) Erase some numeric values and changed them to Canvas' constants.
3) Refactor Canvas a little.
4) Fix wrong Z-values (see images).
5) Fix b/162485723

Before:
Iterators with SelectionOverlay:
Screen Shot 2020-08-25 at 15 47 29

Iterators with Track Titles
Screen Shot 2020-08-25 at 14 44 16

Now:
Screen Shot 2020-08-26 at 13 18 50
Screen Shot 2020-08-26 at 13 19 26

### DIFF
--- a/OrbitGl/CaptureWindow.cpp
+++ b/OrbitGl/CaptureWindow.cpp
@@ -495,7 +495,7 @@ void CaptureWindow::Draw() {
 
     Vec2 pos(m_MouseX, m_WorldTopLeftY);
     // Vertical green line at mouse x position
-    ui_batcher_.AddVerticalLine(pos, -m_WorldHeight, Z_VALUE_TEXT, Color(0, 255, 0, 127));
+    ui_batcher_.AddVerticalLine(pos, -m_WorldHeight, kZValueText, Color(0, 255, 0, 127));
   }
 }
 
@@ -537,14 +537,14 @@ void CaptureWindow::DrawScreenSpace() {
   float margin_x1 = getWidth();
   float margin_x0 = margin_x1 - vertical_margin;
 
-  Box box(Vec2(margin_x0, 0), Vec2(margin_x1 - margin_x0, canvasHeight), GlCanvas::Z_VALUE_MARGIN);
+  Box box(Vec2(margin_x0, 0), Vec2(margin_x1 - margin_x0, canvasHeight), GlCanvas::kZValueMargin);
   ui_batcher_.AddBox(box, kBackgroundColor);
 
   // Time bar background
   if (time_graph_.GetCaptureTimeSpanUs() > 0) {
     Box box(Vec2(0, time_graph_.GetLayout().GetSliderWidth()),
             Vec2(getWidth(), time_graph_.GetLayout().GetTimeBarHeight()),
-            GlCanvas::Z_VALUE_TIME_BAR_BG);
+            GlCanvas::kZValueTimeBarBg);
     ui_batcher_.AddBox(box, Color(70, 70, 70, 200));
   }
 }
@@ -779,11 +779,11 @@ void CaptureWindow::RenderTimeBar() {
 
       std::string text = GetPrettyTime(absl::Microseconds(current_micros));
       float worldX = time_graph_.GetWorldFromUs(current_micros);
-      m_TextRenderer.AddText(text.c_str(), worldX + xMargin, worldY, GlCanvas::Z_VALUE_TEXT_UI,
+      m_TextRenderer.AddText(text.c_str(), worldX + xMargin, worldY, GlCanvas::kZValueTextUi,
                              Color(255, 255, 255, 255));
 
       Vec2 pos(worldX, worldY);
-      ui_batcher_.AddVerticalLine(pos, height, GlCanvas::Z_VALUE_UI, Color(255, 255, 255, 255));
+      ui_batcher_.AddVerticalLine(pos, height, GlCanvas::kZValueUi, Color(255, 255, 255, 255));
     }
   }
 }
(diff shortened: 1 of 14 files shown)


ronaldfw:
Quick question: Are the screenshots showing the bugs or are they showing the rendering after the fixes?

karupayun:
> Quick question: Are the screenshots showing the bugs or are they showing the rendering after the fixes?

Sorry for the confusion. Showing the bug. I've just edited and now you can see before & after.

ronaldfw:
> > Quick question: Are the screenshots showing the bugs or are they showing the rendering after the fixes?
> 
> Sorry for the confusion. Showing the bug. I've just edited and now you can see before & after.

Ok, thanks. Looks great!

OrbitGl/CaptureWindow.cpp
```diff
@@ -494,6 +494,7 @@ void CaptureWindow::Draw() {
     RenderTimeBar();
 
     Vec2 pos(m_MouseX, m_WorldTopLeftY);
+    // Green line in mouse's position
```
ronaldfw, line 497: Nit: in --> at

Personally, I would also remove the apostrophe and maybe clarify the dimention, so this would read:
// Vertical green line at mouse x position.

ronaldfw (approved):
Just a small nit, o/w looks reasonable to me. 

My understanding is that this does not fix the issue that text is rendered at the end and thus on top of everything else, right? So, transparent overlays drawn "over" text (like the text on the TimeGraph time slices) do not change the color tint of the text (as it should). Or am I wrong about this?

ronaldfw:
> Just a small nit, o/w looks reasonable to me.
> 
> My understanding is that this does not fix the issue that text is rendered at the end and thus on top of everything else, right? So, transparent overlays drawn "over" text (like the text on the TimeGraph time slices) do not change the color tint of the text (as it should). Or am I wrong about this?

Clarification: I think the green overlay is actually fine, as it's drawn at the very end. But the iterator overlay color does not affect the text on the timeslices (but it should).

OrbitGl/CaptureWindow.cpp
karupayun: Thank you! :)

karupayun:
> > My understanding is that this does not fix the issue that text is rendered at the end and thus on top of everything else, right? So, transparent overlays drawn "over" text (like the text on the TimeGraph time slices) do not change the color tint of the text (as it should). Or am I wrong about this?
> 
> Clarification: I think the green overlay is actually fine, as it's drawn at the very end. But the iterator overlay color does not affect the text on the timeslices (but it should).

Yeah, it doesn't solve the text & color-lint problem. It was one of my goals when I started looking at, but no solution. 
BTW, I don't see any difference between the green overlay and the iterators (except than before they were in differents z-level). Also didn't follow the point about text drawing on top of everything. In one of the images, if the iterator was on top of the text, I can't see the text. Could be some difference between Windows and Linux?

reichlfl (approved):
Looks great, thanks!

OrbitGl/GlCanvas.cpp
```diff
@@ -17,22 +17,21 @@
...
 float GlCanvas::Z_VALUE_TEXT_UI = 0.0f;
-float GlCanvas::Z_VALUE_TEXT_UI_BG = -0.005f;
-float GlCanvas::Z_VALUE_TEXT = -0.01f;
-float GlCanvas::Z_VALUE_OVERLAY = -0.012f;
-float GlCanvas::Z_VALUE_OVERLAY_BG = -0.013f;
-float GlCanvas::Z_VALUE_CONTEXT_SWITCH = -0.015f;
-float GlCanvas::Z_VALUE_EVENT = -0.015f;
-float GlCanvas::Z_VALUE_BOX_ACTIVE = -0.02f;
-float GlCanvas::Z_VALUE_BOX_INACTIVE = -0.03f;
+float GlCanvas::Z_VALUE_UI = 0.0f;
```
dimitry-, line 23: kZValueUi? since you are renaming most of them anyways.

ronaldfw:
> BTW, I don't see any difference between the green overlay and the iterators (except than before they were in differents z-level). Also didn't follow the point about text drawing on top of everything. In one of the images, if the iterator was on top of the text, I can't see the text. Could be some difference between Windows and Linux?

My understanding is the following (please someone correct if any of this is wrong): For the purpose of discussing the issues here, two things matter: drawing order and z values and they influence each other to some degree.

First, the issue that you were not seeing text under the iterator overlay was caused by two things: The text was drawn after the actual transparent overlay _and_ it was drawn with a smaller z value than the overlay. Because of the smaller z value, drawing the text was simply discarded at the depth testing stage. This issue is fixed here, because now the text is drawn with a higher z value. 

Second, I was wrong about there being a difference between the green overlay and the iterator overlay. They are both not color tinting the text at all. This is not very obvious as the overlays are pretty light, but can be seen when you make all transparent overlays a lot more opaque. This is because the drawing order is roughly (I'm omitting stuff here that is not relevant):

1) Draw iterator overlay
2) Draw green overlay 
3) Draw text (on top of everything else)

For the transparent overlays to affect the color of the text, the rendering order needs to be something like this: 

1) Draw capture window
2) Draw capture window text
3) Draw iterator overlays 
4) Draw iterator overlay text (on top of everything so far)
5) Draw green overlay
6) Draw green overlay text (on top of everything so far)

So, this does add rendering passes that are not there currently, at least for the text rendering. I think though that we need to change this, as the white text on top of everything else makes it very hard to discern which text belongs to which overlay (depending on rendering location). 

Not for this PR, but something we should look into.

OrbitGl/GlCanvas.cpp
karupayun: It's Done!

karupayun:
Yes, I agree 100% with you now. Really nice summary :)

Only thing that I want to add, if we change the kZValueOverlay constant to be greater than the kZValueText one, and modify the order of drawing elements (for example, storing every draw-element in a priority queue and after that drawing from kZValue lower to higher ones), I think that there won't be drawing-bugs anymore.

ronaldfw:
> Yes, I agree 100% with you now. Really nice summary :)
> 
> Only thing that I want to add, if we change the kZValueOverlay constant to be greater than the kZValueText one, and modify the order of drawing elements (for example, storing every draw-element in a priority queue and after that drawing from kZValue lower to higher ones), I think that there won't be drawing-bugs anymore.

I like the idea of using something like a priority queue and then just drawing by z value order (in that case we wouldn't even need depth testing on the GPU anymore, so z values are just used for establishing the order). Given the small number of z values we have effectively, a simple data structure would be sufficient, basically a list of pairs of batchers and text renderers that are indexed by z value.

I still think we need a larger number of different z values for overlays and text layers. There is text that is on the iterator overlays, but that should still be below the green overlay (and affected by its transparent color). So, I think three different text z values and two different overlay z values. The priority queue would then enforce exactly the drawing order I stated in my comment above (right?).

dimitry- (approved):

karupayun:
> I like the idea of using something like a priority queue and then just drawing by z value order (in that case we wouldn't even need depth testing on the GPU anymore, so z values are just used for establishing the order). Given the small number of z values we have effectively, a simple data structure would be sufficient, basically a list of pairs of batchers and text renderers that are indexed by z value.
> 
> I still think we need a larger number of different z values for overlays and text layers. There is text that is on the iterator overlays, but that should still be below the green overlay (and affected by its transparent color). So, I think three different text z values and two different overlay z values. The priority queue would then enforce exactly the drawing order I stated in my comment above (right?).

Yes, exactly as you said. The priority queue could be a hash_map, where keys are "kZValue" and values, the list of elements to draw in this level, and after find the order of these hash_keys and draw the lists in that order.

Also, we can make something like start drawing the text elements in some level and after that the box elements, and avoid to duplicate the amount of kZValues.

karupayun:
FYI, I added a new commit which solves b/162485723 (modified a kZValue).

PR #1121, opened 2020-08-26
Implement basic requirements for oss-fuzz

beckerhe:
This is ready for review now.

Please find more details in the commit message, so here only a general overview:

This PR is adding all what's needed to build inside the oss-fuzz environment.

1. It's adding conan options such that OrbitQt can be disabled. That necessary because linking against Qt inside of oss-fuzz is a pain.
2. The lockfiles are updated as new conan options have been added. There is a positional change of the imgui in the lockfiles which is caused by rearranging the requirements in our conanfile.py
3. The fuzzing conan profiles haven been updated to the new options
4. The cmake-fuzzing macro was changed to obey an environment variable set by the oss-fuzz environment.


antonrohr:
Nit from the commit message
> The normal tests might failed when the code-base is compiled with sanitizers for fuzzing.
might failed -> might fail

IrinaShkviro (approved):
Looks good from my perspective

third_party/conan/configs/linux/profiles/libfuzzer_base
```diff
@@ -0,0 +1,17 @@
+# This file is intended to be imported when building for oss-fuzz
+# Compiler and settings are auto-detected in that case, but we want to have
+# an upstream config file where settings can be changed.
+BASE_CFLAGS= -fno-omit-frame-pointer -mno-omit-leaf-frame-pointer -D_FORTIFY_SOURCE=2 -fstack-protector-all -fPIC
+BASE_CXXFLAGS= -fno-omit-frame-pointer -mno-omit-leaf-frame-pointer -fsized-deallocation -D_FORTIFY_SOURCE=2 -fstack-protector-all -fPIC
+BASE_LDFLAGS=-Wl,-z,relro,-z,now,-z,noexecstack -pthread
```
antonrohr, line 6: Is this line correct? 
maybe:
```suggestion
BASE_LDFLAGS= -Wl,-z,relro,-z,now,-z,noexecstack,-pthread
```

antonrohr (approved):
I did not test anything, but the explanations are good and everything seems pretty straight forward. 

I can judge whether the lock files are right. Maybe there is a way to test that? Something like check out your PR, run build.sh, run gen_lockfiles.sh and then the lock files should not change?

third_party/conan/configs/linux/profiles/libfuzzer_base
beckerhe: Yes the line is correct. Adding a comma to the -pthread option will break the build. As far as I know there is no -pthread-option that can be passed to `-Wl` or do you have different information?

beckerhe:
> I did not test anything, but the explanations are good and everything seems pretty straight forward.
> 
> I can judge whether the lock files are right. Maybe there is a way to test that? Something like check out your PR, run build.sh, run gen_lockfiles.sh and then the lock files should not change?

The best way to check lockfiles is the CI, at least for now. If the CI is happy, the lockfiles are fine with a pretty high probability.

third_party/conan/configs/windows/profiles/msvc2017_debug_x86
```diff
@@ -10,6 +10,7 @@ build_type=Debug
 [options]
 OrbitProfiler:system_qt=False
 OrbitProfiler:with_gui=False
+OrbitProfiler:with_orbitgl=False
```
dimitry-, line 13: I was wondering where are these profiles used? This is 32 bit windows is it?

beckerhe:
> Nit from the commit message
> 
> > The normal tests might failed when the code-base is compiled with sanitizers for fuzzing.
> 
> might failed -> might fail

Fixed. Thank you.

dimitry- (approved):

third_party/conan/configs/windows/profiles/msvc2017_debug_x86
beckerhe: Yeah. We once had this discussion that we wanted to have a 32bit build for preserving OrbitDll's functionality (which needs 32bit support).

But since we removed OrbitDll for now anyway, I also feel that we should remove those. But I didn't want to do it in this PR.

third_party/conan/configs/linux/profiles/libfuzzer_base
antonrohr: No I don't

PR #1122, opened 2020-08-26
Fix crash on preset loading.

danielfenner:
Hooking a already hooked function CHECK-fails. I think it is just fine
to ignore any duplicates.

BUG: b/166424640
TEST: Run locally, does not crash anymore.


florian-kuebler (approved):
Yes, you are right. Looks fine! Thanks for the fix!

PR #1123, opened 2020-08-26
Add update-option to build.ps1

beckerhe:
Newer versions of conan require the `--update`option. Otherwise they
will fail if there is deviation in the recipe revision. Always adding
the update option is not ideal but a work-around for now.

### DIFF
--- a/build.ps1
+++ b/build.ps1
@@ -47,7 +47,7 @@ foreach ($profile in $profiles) {
   
   Write-Host "Building Orbit in build_$profile/ with conan profile $profile"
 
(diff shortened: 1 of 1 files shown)


kisenshi:
Great, thanks! So with this update if I get the same issue than the other day I could just do `& .\build.ps1 --update` to fix it right?

kisenshi (approved):

beckerhe:
> Great, thanks! So with this update if I get the same issue than the other day I could just do `& .\build.ps1 --update` to fix it right?

No quite: You just do `& .\build.ps1 <<profile>>` and it should just work.

Thanks!

PR #1124, opened 2020-08-26
Move CaptureData to App

florian-kuebler:
This finally removes Capture.h/Capture.cpp.
Writes to CaptureData should be thread local, thus only occur in the
CaptureListener part of OrbitApp. As we need to access it also while
capturing in real-time (e.g. for selection reports), it can also not
just be moved to process manager, once done with capturing.
There is currently, no synchronization (as there might be reads while
capturing), as there were not before.

Test: Compile, Do captures, select callstacks (while capturing), load
symbols.
Bug: http://163020637

**Follow Up:** 
properly synchronize capture data


OrbitCaptureClient/CaptureClient.cpp
```diff
@@ -30,7 +31,8 @@ static CaptureOptions::InstrumentedFunction::FunctionType IntrumentedFunctionTyp
 }
 
 ErrorMessageOr<void> CaptureClient::StartCapture(
-    ThreadPool* thread_pool, int32_t pid,
+    ThreadPool* thread_pool, int32_t process_id, std::string process_name,
+    std::shared_ptr<Process> process,
     absl::flat_hash_map<uint64_t, FunctionInfo> selected_functions) {
```
dimitry-, line 36: Maybe this is the point to introduce CaptureParameters class, also considering there is soon going to be list of tracepoint and process will get replaced.

CaptureClient only forwards them to OnCaptureStarted anyways.

OrbitCore/BlockChain.h
```diff
@@ -111,8 +111,30 @@ class BlockChain final {
...
   BlockChain() : size_(0) { root_ = current_ = new Block<T, BlockSize>(nullptr); }
 
+  BlockChain(const BlockChain& other) = delete;
+
+  BlockChain& operator=(const BlockChain& other) = delete;
+
+  BlockChain(BlockChain&& other) {
+    root_ = other.root_;
+    current_ = other.current_;
+    size_ = other.size_;
```
dimitry-, line 121: set other.* to nullptr?

florian-kuebler:
Added @beckerhe and @kisenshi FYI for the changes in fuzzer and ggp client

OrbitCore/BlockChain.h
```diff
@@ -111,8 +111,30 @@ class BlockChain final {
...
+  }
+
+  BlockChain& operator=(BlockChain&& other) {
```
dimitry-, line 124: s/other/that/? not sure if this is better.

OrbitCore/BlockChain.h
dimitry-: also - add a test for blockchain moves.

OrbitCore/CallstackData.h
```diff
@@ -18,6 +18,26 @@ class CallstackData {
...
+
+  CallstackData(CallstackData&& other) {
+    absl::MutexLock lock(&(other.unique_callstacks_mutex_));
+    unique_callstacks_ = std::move(other.unique_callstacks_);
+    callstack_events_ = std::move(other.callstack_events_);
+  }
+
+  CallstackData& operator=(CallstackData&& other) {
+    absl::MutexLock lock(&(other.unique_callstacks_mutex_));
+    absl::MutexLock lock_other(&(other.unique_callstacks_mutex_));
```
dimitry-, line 33: this can end up in a deadlock... In this sistuation we probably want to move all at once by moving unique_ptr of another (non-copiable/non-movable class with the mutex).

OrbitCore/CallstackData.h
```diff
@@ -61,6 +81,12 @@ class CallstackData {
     }
   }
 
+  [[nodiscard]] absl::flat_hash_map<CallstackID, std::shared_ptr<CallStack>> unique_callstacks()
```
dimitry-, line 84: this returns a copy... maybe make the name reflect this? GetUniqueCallstacksCopy()?

OrbitGl/TopDownView.h
```diff
@@ -63,7 +63,7 @@ class [[nodiscard]] TopDownInternalNode : public TopDownNode {
 class [[nodiscard]] TopDownFunction : public TopDownInternalNode {
  public:
   explicit TopDownFunction(uint64_t function_absolute_address, std::string function_name,
-                           TopDownNode* parent)
+                           TopDownNode * parent)
```
dimitry-, line 66: is this necessary? :)

OrbitGl/TopDownView.h
```diff
@@ -7,12 +7,12 @@
 
 #include <unordered_map>
 
-#include "SamplingProfiler.h"
+#include "CaptureData.h"
 #include "absl/container/node_hash_map.h"
 
 class [[nodiscard]] TopDownNode {
```
dimitry-, line 13: I think this nodiscard confuses clang-format

OrbitCore/CallstackData.h
florian-kuebler: Done. Thanks.

OrbitCaptureClient/CaptureClient.cpp
florian-kuebler: Yeah, maybe. Still not 100% sure, if I should add all of those, or keep it as it was, and reconstruct the others on the OnCaptureStarted. I guess the only issue was, that there was would be different threads reading/writing involved.

OrbitCore/BlockChain.h
florian-kuebler: you mean the name of the variable? I'd guess "other" should be fine. Otherwise, I could use "that".

OrbitCore/CallstackData.h
florian-kuebler: Not sure, If I got you right? Should I remove the copy assignment operator and on the usage use unique_ptr and move that one?

OrbitGl/TopDownView.h
florian-kuebler: I removed them in the file.

OrbitGl/TopDownView.h
florian-kuebler: Sorry, that was clang-format. I guess now its even worse, but at least correct.

OrbitCore/CallstackData.h
IrinaShkviro: The possible deadlock is the reason why these operators weren't implemented before, and because of it we have std::unique_ptr<CallstackData> in CaptureData. But do you use these operators in this commit or why you want to add it?

antonrohr (commented):
I haven't looked at everything yet, but here are some first comments

OrbitGl/TopDownView.cpp
```diff
@@ -4,6 +4,8 @@
 
 #include "TopDownView.h"
 
+#include <CaptureData.h>
+
```
IrinaShkviro, line 8: Is it needed?

OrbitCore/CallstackData.h
florian-kuebler: Initially, I wanted to get rid of passing the Pointer to CallstackData in particular to SamplingProfiler. However, as passing a non-copyable reference is not better, I've reverted the changes here and explicitly delete move and copy operators.

OrbitCore/BlockChain.h
florian-kuebler: I reverted this change, as it is not needed here anymore.

florian-kuebler: But for now, I made it explicit and deleted the move and copy operations

OrbitCore/CallstackData.h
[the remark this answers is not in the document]
florian-kuebler: Thanks, but I removed that complete block again.

florian-kuebler (commented):
Thanks for the review so far. I addressed most of the comments and deleted the move operators for BlockChain and CallstackData, while passing the latter one as bare pointer (as done before).

OrbitGl/TopDownView.cpp
```diff
@@ -82,11 +82,13 @@ static void AddCallstackToTopDownThread(TopDownThread* thread_node,
...
-    const SamplingProfiler& sampling_profiler, const std::string& process_name,
-    const absl::flat_hash_map<int32_t, std::string>& thread_names) {
+std::unique_ptr<TopDownView> TopDownView::CreateFromCaptureData(const CaptureData& capture_data) {
   auto top_down_view = std::make_unique<TopDownView>();
-  for (const ThreadSampleData* thread_sample_data : sampling_profiler.GetThreadSampleData()) {
+  const std::string& process_name = capture_data.process_name();
+  const absl::flat_hash_map<int32_t, std::string>& thread_names = capture_data.thread_names();
+  const SamplingProfiler& sampling_profiler = capture_data.sampling_profiler();
+  for (const ThreadSampleData* thread_sample_data :
+       capture_data.sampling_profiler().GetThreadSampleData()) {
```
antonrohr, line 91: tinyiest nit: 
```suggestion
       sampling_profiler.GetThreadSampleData()) {
```

antonrohr: Not that it matters at all

OrbitGl/TopDownView.h
```diff
@@ -7,10 +7,10 @@
 
 #include <unordered_map>
 
-#include "SamplingProfiler.h"
+#include "CaptureData.h"
 #include "absl/container/node_hash_map.h"
 
-class [[nodiscard]] TopDownNode {
```
antonrohr, line 13: Just out of curiosity. Why remove those?

antonrohr (commented):
Thank you a lot for this @florian-kuebler ! This is a big step in the right direction

OrbitCore/CaptureData.h
```diff
@@ -100,16 +111,16 @@ class CaptureData {
...
   }
 
-  const absl::flat_hash_map<uint64_t, orbit_client_protos::FunctionStats>& functions_stats() {
+  const absl::flat_hash_map<uint64_t, orbit_client_protos::FunctionStats>& functions_stats() const {
     return functions_stats_;
   }
 
-  const orbit_client_protos::FunctionStats& GetFunctionStatsOrDefault(uint64_t function_address);
+  const orbit_client_protos::FunctionStats& GetFunctionStatsOrDefault(
+      uint64_t function_address) const;
```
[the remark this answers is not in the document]
antonrohr: already done

OrbitCore/SamplingProfiler.cpp
```diff
@@ -100,8 +99,11 @@ void SamplingProfiler::SortByThreadUsage() {
   sorted_thread_sample_data_.clear();
   sorted_thread_sample_data_.reserve(thread_id_to_sample_data_.size());
 
-  // "All"
-  thread_id_to_sample_data_[kAllThreadsFakeTid].average_thread_usage = 100.f;
+  // If "All" exists, set to 100% usage
+  auto all_threads_it = thread_id_to_sample_data_.find(kAllThreadsFakeTid);
+  if (all_threads_it != thread_id_to_sample_data_.end()) {
+    all_threads_it->second.average_thread_usage = 100.f;
+  }
```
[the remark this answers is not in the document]
florian-kuebler: Yes, that was intended. There was a check later on the results, to remove that value again, if there is no data to draw.

OrbitGl/TopDownView.h
florian-kuebler: They confused clang-format (as Dimitry has noted correctly), so I removed them.

antonrohr: okay

OrbitGl/App.cpp
```diff
@@ -334,10 +343,11 @@ void OrbitApp::Disassemble(int32_t pid, const FunctionInfo& function) {
       SendDisassemblyToUi(disasm.GetResult(), std::move(empty_report));
       return;
     }
-    const SamplingProfiler& profiler = Capture::capture_data_.GetSamplingProfiler();
+    const CaptureData& capture_data = GetCaptureData();
+    const SamplingProfiler& profiler = capture_data.sampling_profiler();
 
     DisassemblyReport report(disasm, FunctionUtils::GetAbsoluteAddress(function), profiler,
-                             Capture::capture_data_.GetCallstackData()->GetCallstackEventsSize());
+                             capture_data.GetCallstackData().GetCallstackEventsSize());
```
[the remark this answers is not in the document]
antonrohr: okay. sounds good to me!

OrbitCore/CaptureData.cpp
```diff
@@ -45,12 +45,34 @@ const FunctionInfo* CaptureData::GetSelectedFunction(uint64_t function_address)
...
+  address_infos_.emplace(absolute_address, std::move(address_info));
+}
+
 const std::string CaptureData::kUnknownFunctionOrModuleName{"???"};
 
 const std::string& CaptureData::GetFunctionNameByAddress(uint64_t absolute_address) const {
   const FunctionInfo* function = process_->GetFunctionFromAddress(absolute_address, false);
   if (function != nullptr) {
-    return function->name();
+    return function->pretty_name();
```
IrinaShkviro, line 75: Maybe it makes sense to use FunctionUtils::GetDisplayName() function here? (it returns `pretty_name` if not empty and `name` otherwise)

OrbitCore/SamplingProfiler.cpp
```diff
@@ -251,90 +249,55 @@ uint32_t SamplingProfiler::GetCountOfFunction(uint64_t function_address) const {
...
   // when this fails (e.g., the module containing the function has not
   // been loaded) use (for now) the LinuxAddressInfo that is collected
   // for every address in a callstack. SamplingProfiler relies heavily
   // on the association between address and function address held by
-  // m_ExactAddressToFunctionAddress, otherwise each address is
+  // exact_address_to_function_address_, otherwise each address is
   // considered a different function.
-  uint64_t function_address;
+  uint64_t absolute_function_address;
   // TODO(kuebler): Remove name management here completely
```
IrinaShkviro, line 267:
```suggestion

```

OrbitGl/CaptureSerializer.h
```diff
@@ -5,6 +5,7 @@
 #ifndef ORBIT_GL_CAPTURE_SERIALIZER_H_
 #define ORBIT_GL_CAPTURE_SERIALIZER_H_
 
+#include <CaptureData.h>
```
IrinaShkviro, line 8:
```suggestion
#include "CaptureData.h"
```

OrbitCore/SamplingProfiler.cpp
```diff
@@ -183,44 +182,43 @@ void SamplingProfiler::ProcessSamples(
...
+          }
+
+          auto address_it = exact_address_to_function_address_.find(address);
+          if (address_it != exact_address_to_function_address_.end()) {
+            const uint64_t& function_address = address_it->second;
+            resolved_callstack_data.push_back(function_address);
+            function_address_to_callstack_[function_address].insert(call_stack.GetHash());
+          } else {
+            resolved_callstack_data.push_back(address);
+          }
```
antonrohr, line 208: This branch should never occur

OrbitGl/CaptureSerializer.cpp
```diff
@@ -240,17 +248,27 @@ ErrorMessageOr<void> CaptureSerializer::Load(std::istream& stream) {
...
+  CaptureData capture_data = GenerateCaptureData(capture_info);
 
   // Timers
   TimerInfo timer_info;
   while (ReadMessage(&timer_info, &coded_input)) {
-    time_graph_->ProcessTimer(timer_info);
+    if (timer_info.function_address() > 0) {
+      FunctionInfo* func =
+          capture_data.process()->GetFunctionFromAddress(timer_info.function_address());
+      CHECK(func != nullptr);
```
antonrohr, line 259: This actually does not work. I just played around a little bit and loading a capture from file crashes here

antonrohr (commented):
Thanks for adding all the review comments. Code style and flow lgtm. 

I tested a little bit and sadly capture loading crashes currently

OrbitCore/SamplingProfiler.cpp
florian-kuebler: Why not. We can't be sure to have address infos.

OrbitGl/CaptureSerializer.cpp
IrinaShkviro: We don't store process in capture, so func should be nullptr here. Probably, we want to check if it's selected function instead and call `UpdateFunctionStats` only for this case and then call `time_graph_->ProcessTimer(timer_info, nullptr);` after "if"?

florian-kuebler: Thanks for spotting this. I guess we want to find the function info here from the selected functions rather than from the process.

IrinaShkviro: And why do we need it at all?  I mean we store stats in proto, so we can remove one of those (stats from proto or UpdateFunctionStats from here).

antonrohr:
I also noticed another thing. Orbit shows an empty selection tab after connecting to the instance, instead of the functions tab as before.

OrbitGl/CaptureSerializer.cpp
florian-kuebler: Wow, looked too long on this today. Right. We just need to call process timer here and do not update the stats. That actually might be a bug that existed before (where stats would be updated twice).

florian-kuebler: But we should try to retrieve the function anyways, as in case of an orbit function, it is processed differently.

florian-kuebler:
Converted to Draft, as there are still a couple of bugs, to be fixed before.

OrbitGl/CaptureSerializer.cpp
IrinaShkviro: I think that as function was nullptr this part (with updating stats) was just skipped

dimitry- (approved):
otherwise lgtm

OrbitCore/BlockChain.h
```diff
@@ -111,8 +111,21 @@ class BlockChain final {
...
+
+  BlockChain& operator=(BlockChain&& other) = delete;
+
   ~BlockChain() {
     // Find last block in chain
+    if (current_ == nullptr) {
+      CHECK(root_ == nullptr);
+      CHECK(size_ == 0);
+      return;
+    }
```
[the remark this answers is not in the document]
florian-kuebler: Thanks, that was debug code.

OrbitCore/CaptureData.cpp
florian-kuebler: Thanks done.

OrbitCore/SamplingProfiler.cpp
florian-kuebler: Thanks.

OrbitGl/CaptureSerializer.h
florian-kuebler: Thanks

OrbitCaptureClient/CaptureClient.cpp
florian-kuebler: Lets do this in a sep. PR.

florian-kuebler:
Crash and strange behaviour on symbol loading is fixed now. The bug, with selecting the selection tab on start up is already on master. I'll look into this tomorrow, but that should not block this PR.
Also there seem to be a crash when trying to capture and no process is selected, that seem related with recent changes, but also present on master.

OrbitGl/CaptureSerializer.cpp
```diff
@@ -240,17 +248,25 @@ ErrorMessageOr<void> CaptureSerializer::Load(std::istream& stream) {
...
+  CaptureData capture_data = GenerateCaptureData(capture_info);
 
   // Timers
   TimerInfo timer_info;
   while (ReadMessage(&timer_info, &coded_input)) {
-    time_graph_->ProcessTimer(timer_info);
+    if (timer_info.function_address() > 0) {
+      const FunctionInfo& func =
+          capture_data.selected_functions().at(timer_info.function_address());
+      capture_data.process()->GetFunctionFromAddress(timer_info.function_address());
```
IrinaShkviro, line 259:
```suggestion

```

IrinaShkviro (approved):
Thanks a lot for this change! Looks good, just a couple of small comments.

OrbitCore/CaptureData.h
```diff
@@ -143,9 +137,11 @@ class CaptureData {
 
   [[nodiscard]] const std::shared_ptr<Process>& process() const { return process_; }
 
-  [[nodiscard]] const SamplingProfiler& sampling_profiler() const { return sampling_profiler_; }
+  [[nodiscard]] const class SamplingProfiler& sampling_profiler() const {
+    return sampling_profiler_;
+  }
 
-  void set_sampling_profiler(SamplingProfiler sampling_profiler) {
+  void set_sampling_profiler(class SamplingProfiler sampling_profiler) {
```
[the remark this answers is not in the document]
antonrohr: I thinks it's part of the forward declaration. SamplingProfiler.h includes "CaptureData.h", therefore CaptureData cant include SamplingProfiler.h. To still be able to use SamplingProfiler this trickery is necessary

OrbitGl/CaptureSerializer.cpp
antonrohr: I guess that would have been caught by a [[nodiscard]]

OrbitCore/SamplingProfiler.cpp
florian-kuebler: Well, you are totally right. I removed that branch and directly retrieved the address.

OrbitGl/CaptureSerializer.cpp
florian-kuebler: Thanks

OrbitCore/CaptureData.h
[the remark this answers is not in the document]
IrinaShkviro: Actually, it's vise versa: I see `#include "SamplingProfiler.h"` in CaptureData.h and forward declaration `class CaptureData;` in SamplingProfiler.h. So looks like it's unnecessary change here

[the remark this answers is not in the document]
antonrohr: Ah okay. I did not which way it was done. Thank you for catching this!

antonrohr:
I just caught another crash. (I am unsure if that also happens on master)
Reproduce:
1. Load some symbols
2. Hook a function
3. Take a capture (start & stop)
4. Unhook the function
5. Start capture -> crash

florian-kuebler (commented):
Addressed review comments. Thanks for the reviews.

OrbitCaptureClient/include/OrbitCaptureClient/CaptureClient.h
```diff
@@ -65,8 +66,6 @@ class CaptureClient {
 
   CaptureListener* capture_listener_ = nullptr;
 
-  std::optional<CaptureEventProcessor> event_processor_;
```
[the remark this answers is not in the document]
antonrohr: Now in CaptureClient.cpp the include is missing

florian-kuebler:
> I just caught another crash. (I am unsure if that also happens on master)
> Reproduce:
> 
> 1. Load some symbols
> 2. Hook a function
> 3. Take a capture (start & stop)
> 4. Unhook the function
> 5. Start capture -> crash

Thanks for doing all the testing. I'll have a look into this.

florian-kuebler:
@antonrohr I fixed that bug when doing a capture after unhooking a funciton

antonrohr:
I found another crash.

Reproduce.
1. Take a capture
2. Select a range in the time graph -> selection tab will get focused.
3. Select an entry (sample) in the selection Tab, the callstack will get displayed in the lower half of the selection tab. 
4. press the down arrow key to select one sample after the other. 
5. crash 

The crash happens because of this CHECK in SamplingReport.cpp
```
void SamplingReport::OnCallstackIndexChanged(size_t index) {
  if (index < selected_sorted_callstack_report_->callstacks_count.size()) {
    const CallstackCount& cs = selected_sorted_callstack_report_->callstacks_count[index];
    selected_callstack_index_ = index;
    auto it = unique_callstacks_.find(cs.callstack_id);
    CHECK(it != unique_callstacks_.end());
    callstack_data_view_->SetCallStack(*it->second);
  } else {
    selected_callstack_index_ = 0;
  }
}
```

florian-kuebler:
@antonrohr Thanks for reporting this. I've fixed the issue.

antonrohr (approved):

PR #1126, opened 2020-08-26
Add selected_functions to capture in OrbitClientGgp

kisenshi:
Populate _selected_functions_ to be provided to _CaptureClient_ when requesting to start a capture. Each of the functions provided as argument acts as a filter to find one or more functions loaded in _target_process__ that matches its string.

**Add optional -functions parameter** It is possible to provide the list of functions to hook when capturing so we create a new "functions" argument, which accepts a list of strings separated by comma: `.\OrbitClientGgp -functions "function1,function2"`

**Populate selected_functions based on the list provided** The list of selected functions acts as a "filter" so all the functions existent in _target_process__ that matches any of the strings provided will be included in selected_functions. Three methods has been created:  

1. _ClientGgp::GetSelectedFunctions_ Returns selected_functions
2. _ClientGgp::SelectedFunctionMatch_ Checks if a function matches any of the "filters" in the list and returns the matching string if it does.
3. _ClientGgp::InformUsedSelectedCaptureFunctions_ Go through the filters provided if not all of them got a match when looking for the selected functions. Informs the user by logs that any functions related to them will be included in the capture.

### DIFF
--- a/OrbitClientGgp/ClientGgp.cpp
+++ b/OrbitClientGgp/ClientGgp.cpp
@@ -9,6 +9,7 @@
 #include <map>
 #include <memory>
 #include <string>
+#include <vector>
 
 #include "OrbitBase/Logging.h"
 #include "OrbitBase/Result.h"
--- a/OrbitClientGgp/ClientGgp.cpp
+++ b/OrbitClientGgp/ClientGgp.cpp
@@ -180,10 +180,11 @@ bool ClientGgp::InitCapture() {
   return true;
 }
 
-void ClientGgp::InformUsedSelectedCaptureFunctions() {
-  if (capture_functions_used_.size() != options_.capture_functions.size()) {
-    for (std::string selected_function : options_.capture_functions) {
-      if (!capture_functions_used_.contains(selected_function)) {
+void ClientGgp::InformUsedSelectedCaptureFunctions(
+    absl::flat_hash_set<std::string> capture_functions_used) {
+  if (capture_functions_used.size() != options_.capture_functions.size()) {
+    for (const std::string selected_function : options_.capture_functions) {
+      if (!capture_functions_used.contains(selected_function)) {
         ERROR("Function matching %s not found; will not be hooked in the capture",
               selected_function);
       }
@@ -193,27 +194,29 @@ void ClientGgp::InformUsedSelectedCaptureFunctions() {
   }
 }
 
-bool ClientGgp::IsSelectedFunction(const FunctionInfo& func) {
-  for (std::string selected_function : options_.capture_functions) {
+std::string ClientGgp::SelectedFunctionMatch(const FunctionInfo& func) {
+  for (const std::string selected_function : options_.capture_functions) {
     if (func.pretty_name().find(selected_function) != std::string::npos) {
-      if (!capture_functions_used_.contains(selected_function)) {
-        capture_functions_used_.insert(selected_function);
-      }
-      return true;
+      return selected_function;
     }
   }
-  return false;
+  return {};
 }
 
 absl::flat_hash_map<uint64_t, FunctionInfo> ClientGgp::GetSelectedFunctions() {
   absl::flat_hash_map<uint64_t, FunctionInfo> selected_functions;
+  absl::flat_hash_set<std::string> capture_functions_used;
   for (const auto& func : target_process_->GetFunctions()) {
-    if (IsSelectedFunction(*func)) {
+    const std::string selected_function_match = SelectedFunctionMatch(*func);
+    if (!selected_function_match.empty()) {
       uint64_t address = FunctionUtils::GetAbsoluteAddress(*func);
       selected_functions[address] = *func;
+      if (!capture_functions_used.contains(selected_function_match)) {
+        capture_functions_used.insert(selected_function_match);
+      }
     }
   }
-  InformUsedSelectedCaptureFunctions();
+  InformUsedSelectedCaptureFunctions(capture_functions_used);
   return selected_functions;
 }
 
--- a/OrbitClientGgp/ClientGgp.cpp
+++ b/OrbitClientGgp/ClientGgp.cpp
@@ -11,13 +11,16 @@
 #include <string>
 #include <vector>
 
+#include "FunctionUtils.h"
 #include "OrbitBase/Logging.h"
 #include "OrbitBase/Result.h"
 #include "OrbitCaptureClient/CaptureClient.h"
 #include "OrbitClientServices/ProcessManager.h"
 #include "SymbolHelper.h"
+#include "absl/container/flat_hash_set.h"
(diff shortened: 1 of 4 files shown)


OrbitClientGgp/ClientGgp.cpp
```diff
@@ -164,6 +180,43 @@ bool ClientGgp::InitCapture() {
...
+      return true;
+    }
+  }
+  return false;
+}
+
+absl::flat_hash_map<uint64_t, FunctionInfo> ClientGgp::GetSelectedFunctions() {
+  absl::flat_hash_map<uint64_t, FunctionInfo> selected_functions;
+  for (const auto& func : target_process_->GetFunctions()) {
+    if (IsSelectedFunction(*func)) {
```
kisenshi, line 211: I have a question here: In `App.cpp` when selected_functions is populated in _OrbitApp::GetSelectedFunctionsAndOrbitFunctions_, it also adds a function if `FunctionUtils::IsOrbitFunc(*func)` returns true. Reading the code I understand that this check is related only to the Orbit app so the ggp client does not need it but I just want to be sure. Can you please confirm? Thanks!

danielfenner: I think this has something to do with the manual instrumentation Pierric is working on. The func.type is KNone for all dynamically instrumented functions I think. The proto would profit from some comments there :)
At some point we'll want the signals from manual instrumentation as well but think it is totally sensible to do this later.

OrbitClientGgp/ClientGgp.cpp
```diff
@@ -164,6 +180,43 @@ bool ClientGgp::InitCapture() {
...
+              selected_function);
+      }
+    }
+  } else {
+    LOG("All functions provided had at least a match");
+  }
+}
+
+bool ClientGgp::IsSelectedFunction(const FunctionInfo& func) {
+  for (std::string selected_function : options_.capture_functions) {
```
danielfenner, line 197: Might profit from making it a const ref.

OrbitClientGgp/ClientGgp.cpp
```diff
@@ -164,6 +180,43 @@ bool ClientGgp::InitCapture() {
...
+        ERROR("Function matching %s not found; will not be hooked in the capture",
+              selected_function);
+      }
+    }
+  } else {
+    LOG("All functions provided had at least a match");
+  }
+}
+
+bool ClientGgp::IsSelectedFunction(const FunctionInfo& func) {
```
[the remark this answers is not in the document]
kisenshi: You are right, will update the code; thanks!

OrbitClientGgp/ClientGgp.cpp
kisenshi: I see, thanks for the clarification. If we add the signals from manual instrumentation I guess it will also require some extra work in the client anyways so I think it will make sense to do this change then. Will leave it as it is for now then, thanks!

OrbitClientGgp/ClientGgp.cpp
kisenshi: True, will update the code, thanks!

OrbitClientGgp/ClientGgp.cpp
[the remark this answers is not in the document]
kisenshi: I have updated the code. As a result, _IsSelectedFunction_ has changed aand now is called _SelectedFunctionMatch_ so it returns the matching string if found so it can be included in _capture_functions_used_by the caller.

OrbitClientGgp/ClientGgp.cpp
kisenshi: Done! I also missed the const in other loops so I updated them as well

kisenshi:
I have updated the code with the suggestions; thanks!

OrbitClientGgp/ClientGgp.cpp
```diff
@@ -164,6 +180,46 @@ bool ClientGgp::InitCapture() {
   return true;
 }
 
+void ClientGgp::InformUsedSelectedCaptureFunctions(
+    absl::flat_hash_set<std::string> capture_functions_used) {
```
danielfenner, line 184: `const absl::flat_hash_set<std::string>& capture_functions_used`

OrbitClientGgp/ClientGgp.cpp
```diff
@@ -164,6 +180,46 @@ bool ClientGgp::InitCapture() {
...
+              selected_function);
+      }
+    }
+  } else {
+    LOG("All functions provided had at least a match");
+  }
+}
+
+std::string ClientGgp::SelectedFunctionMatch(const FunctionInfo& func) {
+  for (const std::string selected_function : options_.capture_functions) {
```
danielfenner, line 198: Sorry for nit picking:
`for (const std::string& selected_function : options_.capture_functions) {`
to avoid the copy

OrbitClientGgp/ClientGgp.h
```diff
@@ -8,6 +8,7 @@
 #include <cstdint>
 #include <memory>
 #include <string>
+#include <vector>
```
danielfenner, line 11: Looks like `#include <vector>` is not needed

OrbitClientGgp/ClientGgp.h
```diff
@@ -50,6 +52,9 @@ class ClientGgp final : public CaptureListener {
   ErrorMessageOr<std::shared_ptr<Process>> GetOrbitProcessByPid(int32_t pid);
   bool InitCapture();
   ErrorMessageOr<void> LoadModuleAndSymbols();
+  std::string SelectedFunctionMatch(const orbit_client_protos::FunctionInfo& func);
+  absl::flat_hash_map<uint64_t, orbit_client_protos::FunctionInfo> GetSelectedFunctions();
+  void InformUsedSelectedCaptureFunctions(absl::flat_hash_set<std::string> capture_functions_used);
```
danielfenner, line 57: `  void InformUsedSelectedCaptureFunctions(const absl::flat_hash_set<std::string>& capture_functions_used);
`

OrbitClientGgp/ClientGgp.cpp

OrbitClientGgp/ClientGgp.cpp
kisenshi: Oh right, thanks! I have updated it here and in other places in the code as well

OrbitClientGgp/ClientGgp.h
kisenshi: I removed it, thanks!

OrbitClientGgp/ClientGgp.h

kisenshi:
I have updated the code with the small fixes suggested, thanks!

danielfenner (approved):
LGTM!

kisenshi:
Thanks; merging these changes now

PR #1127, opened 2020-08-26
ElfFile: Add GetLineInfo and HasDebugInfo methods

dimitry-:
LineInfo returns line number information for particular address.
HasDebugInfo detects if there is a debug info available for this
elf file.

Test: run ElfUtilsTests
Bug: http://b/154580143


ElfUtils/ElfFile.cpp
```diff
@@ -200,10 +217,25 @@ std::string ElfFileImpl<ElfT>::GetFilePath() const {
   return file_path_;
 }
 
+template <typename ElfT>
+ErrorMessageOr<LineInfo> ElfUtils::ElfFileImpl<ElfT>::GetLineInfo(uint64_t address) {
+  auto line_info_or_error = symbolizer_.symbolizeCode(
+      *object_file_, {address, llvm::object::SectionedAddress::UndefSection});
+  if (!line_info_or_error) {
```
dimitry-, line 224: I actually do not know when it returns a error.. maybe we should add detection of `<invalid>` and instead of returning `{"<invalid>", 0}` generate a error message.

antonrohr: I am in favor of generating an ErrorMessage for this case

antonrohr: I briefly looked into the llvm source code and it seems that `symbolizeCode` only returns an error when there is something wrong with the object file.

antonrohr (commented):
This looks very good!

Does it also work with separate `.debug` files? It seems like this is an important use case. There should be a test case for that. 

How expensive is this lookup? Currently `ElfFile::LoadSymbols` sets `source_file` and `source_line` of `SymbolInfo` to default values. Maybe this can be filled with the actual values? If not, what is the plan for the line infos, should they be removed from `SymbolInfo` and only requested on demand?

Maybe you consider changing the `SymbolInfo` proto to use a `LineInfo` proto instead of the two fields `source_line` and `source_file`. Than this `LineInfo` proto could be used instead of the `LineInfo` struct in ElfFile.h

ElfUtils/ElfFile.cpp
```diff
@@ -200,10 +217,25 @@ std::string ElfFileImpl<ElfT>::GetFilePath() const {
   return file_path_;
 }
 
+template <typename ElfT>
+ErrorMessageOr<LineInfo> ElfUtils::ElfFileImpl<ElfT>::GetLineInfo(uint64_t address) {
```
florian-kuebler, line 221: If I got it right, this should only be called, if has_debug_info_section_ is present. So maybe add a CHECK here.

ElfUtils/ElfFile.cpp
```diff
@@ -200,10 +217,25 @@ std::string ElfFileImpl<ElfT>::GetFilePath() const {
...
+  auto line_info_or_error = symbolizer_.symbolizeCode(
+      *object_file_, {address, llvm::object::SectionedAddress::UndefSection});
+  if (!line_info_or_error) {
+    return ErrorMessage(absl::StrFormat(
+        "Unable to get line number info for \"%s\", address=0x%x: %s", object_file_->getFileName(),
+        address, llvm::toString(line_info_or_error.takeError())));
+  }
+
+  auto& line_info = line_info_or_error.get();
+  return LineInfo{line_info.FileName, line_info.Line};
```
florian-kuebler, line 231: Out of curiosity: How does file name look like? Is it relative? Do I need to match with a prefix directory afterwards? Maybe this should be reflected in LineInfo name.

ElfUtils/ElfFileTest.cpp
```diff
@@ -156,3 +156,64 @@ TEST(ElfFile, FileDoesNotExist) {
   EXPECT_THAT(absl::AsciiStrToLower(elf_file.error().message()),
               testing::HasSubstr("no such file or directory"));
 }
+
+TEST(ElfFile, HasDebugInfo) {
+  std::string file_path =
+      Path::JoinPath({Path::GetExecutablePath(), "testdata", "hello_world_elf_with_debug_info"});
+
+  auto hello_world = ElfFile::Create(file_path);
+  ASSERT_TRUE(hello_world) << hello_world.error().message();
```
florian-kuebler, line 165: Didn't we want to avoid implicit conversions to bool?

florian-kuebler: If so, please also change the usages below. Otherwise, feel free to ignore.

florian-kuebler (approved):
Just minor comments. Feel free to ignore.
Thanks for that change! That is valuable. LGTM!

ElfUtils/ElfFile.cpp
antonrohr: You can see an example in the test. It seems like its the full path of the file at time of compilation

ElfUtils/ElfFileTest.cpp
antonrohr: I think the discussion was only about pointer and implicit conversions to bool. `hello_world` here is of type `ErrorMessageOr` and this is based on `outcome::result`.

ElfUtils/ElfFile.cpp
dimitry-: done.

dimitry- (commented):
> Maybe you consider changing the `SymbolInfo` proto to use a `LineInfo` proto instead of the two fields `source_line` and `source_file`. Than this `LineInfo` proto could be used instead of the `LineInfo` struct in ElfFile.h

done.

dimitry-:
Also added additional test and changed loadsymbols to return lineinfos

ElfUtils/ElfFile.cpp
dimitry-: done.

dimitry-:
Loading line numbers for all symbols slows down symbol loading 10-15 times. So we do not do it during symbol loading, this can still be done after loading symbols, I am leaving it for another PR.

PR #1128, opened 2020-08-26
Fix default right tab index

pierricgimmig:
Hi Maria, I'm guessing this was changed by accident, right?

### DIFF
--- a/OrbitQt/orbitmainwindow.ui
+++ b/OrbitQt/orbitmainwindow.ui
@@ -106,7 +106,7 @@
       </widget>
       <widget class="QTabWidget" name="RightTabWidget">
        <property name="currentIndex">
-        <number>7</number>
+        <number>0</number>
        </property>
        <widget class="QWidget" name="FunctionsTab">
         <attribute name="title">


reichlfl:
This happened to me as well, I think it's just bound to happen when working in Qt Designer - we should set the startup tabs programmatically when Orbit launches, WDYT?

pierricgimmig:
> This happened to me as well, I think it's just bound to happen when working in Qt Designer - we should set the startup tabs programmatically when Orbit launches, WDYT?

Hmm, well personally I'm leaning towards leaving it in the .ui file and being more careful on the review/commit. Having it in the .ui file only to overwrite it on init is not ideal I think. Or maybe we can remove it from the .ui and only set it it programmatically?

reichlfl:
> > This happened to me as well, I think it's just bound to happen when working in Qt Designer - we should set the startup tabs programmatically when Orbit launches, WDYT?
> 
> Hmm, well personally I'm leaning towards leaving it in the .ui file and being more careful on the review/commit. Having it in the .ui file only to overwrite it on init is not ideal I think. Or maybe we can remove it from the .ui and only set it it programmatically?

I don't think we can remove it from the UI file, it'll be back every time someone saves it in Designer. But I see your point, let's leave it and keep an eye on it in the PRs.

danielfenner (approved):

MariaSandru27 (approved):

PR #1104, opened 2020-08-23
Display panel of tracepoints in UI

MariaSandru27:

PR #1129, opened 2020-08-27
Fix default tab on right hand side.

danielfenner:
In a previous PR (https://github.com/google/orbit/pull/1104) this was
changed to the right most tab. It should be the left most.

Bug: N/A
Test: Ran the change locally


danielfenner:
duplicate - https://github.com/google/orbit/pull/1128

PR #1130, opened 2020-08-27
Mini rework and tests for Path utils

reichlfl:
This came up for a method that was added to Path.cpp in this PR: https://github.com/google/orbit/pull/1120
I decided to consistently add [[nodiscard]] for all methods, and this lead to some more fixes... which lead to tests to be sure I didn't break the rest of the class.

- Added [[nodiscard]] everywhere
- Changed internal logic for GetBasePath() and Init() due to a warning resulting from [[nodiscard]]
- Renamed some methods to make it more obvious that they are creating directories, and to distinguish between "path" and "dir"
- Added tests for uncovered parts of Path.cpp to make sure I didn't break anything

### DIFF
--- a/OrbitCore/Path.cpp
+++ b/OrbitCore/Path.cpp
@@ -24,13 +24,7 @@
 
 ABSL_FLAG(std::string, log_dir, "", "Set directory for the log.");
 
-std::string Path::base_path_;
-
-void Path::Init() { GetBasePath(); }
-
-std::string Path::GetExecutableName() {
-  // TODO(161419404): This function should probably be called GetExecutablePath
-  //  or GetExecutableFullPath.
+std::string Path::GetExecutablePath() {
 #ifdef _WIN32
   WCHAR cwBuffer[2048] = {0};
   LPWSTR pszBuffer = cwBuffer;
@@ -60,63 +54,10 @@ std::string Path::GetExecutableName() {
 #endif
 }
 
-std::string Path::GetExecutablePath() {
-  // TODO(161419404): This function should probably be named
-  //  GetExecutableDirectory.
-  std::string fullPath = GetExecutableName();
-  std::string path = fullPath.substr(0, fullPath.find_last_of('/')) + "/";
-  return path;
-}
-
-uint64_t Path::FileSize(const std::string& file) {
-  struct stat stat_buf;
-  int ret = stat(file.c_str(), &stat_buf);
-  return ret == 0 ? stat_buf.st_size : 0;
-}
-
-bool Path::DirExists(const std::string& dir) {
-#if _WIN32
-  DWORD ftyp = GetFileAttributesA(dir.c_str());
-  if (ftyp == INVALID_FILE_ATTRIBUTES) return false;
-
-  if (ftyp & FILE_ATTRIBUTE_DIRECTORY) return true;
-
-  return false;
-#else
-  // TODO: also use stat here?
-  DIR* pDir;
-  bool exists = false;
-  pDir = opendir(dir.c_str());
-  if (pDir != nullptr) {
-    exists = true;
-    (void)closedir(pDir);
-  }
-
-  return exists;
-#endif
-}
-
-std::string Path::GetBasePath() {
-  if (!base_path_.empty()) {
-    return base_path_;
-  }
-
-  std::string exePath = GetExecutablePath();
-  base_path_ = exePath.substr(0, exePath.find("bin/"));
-
-  return base_path_;
-}
-
-std::string Path::GetDllPath(bool a_Is64Bit) {
-  std::string basePath = GetBasePath();
-
-  return basePath + GetDllName(a_Is64Bit);
-}
-
-std::string Path::GetDllName(bool a_Is64Bit) { return a_Is64Bit ? "Orbit64.dll" : "Orbit32.dll"; }
+std::string Path::GetExecutableDir() { return GetDirectory(GetExecutablePath()); }
 
 static std::string CreateAndGetConfigPath() {
-  std::string configDir = Path::JoinPath({Path::GetAppDataPath(), "config"});
+  std::string configDir = Path::JoinPath({Path::CreateOrGetOrbitAppDataDir(), "config"});
   std::filesystem::create_directory(configDir);
   return configDir;
 }
@@ -129,67 +70,60 @@ std::string Path::GetSymbolsFileName() {
   return Path::JoinPath({CreateAndGetConfigPath(), "SymbolPaths.txt"});
 }
 
-std::string Path::GetCachePath() {
-  std::string cacheDir = Path::JoinPath({Path::GetAppDataPath(), "cache"});
+std::string Path::CreateOrGetCacheDir() {
+  std::string cacheDir = Path::JoinPath({Path::CreateOrGetOrbitAppDataDir(), "cache"});
   std::filesystem::create_directory(cacheDir);
   return cacheDir;
 }
 
-std::string Path::GetPresetPath() {
-  std::string presetDir = Path::JoinPath({Path::GetAppDataPath(), "presets"});
+std::string Path::CreateOrGetPresetDir() {
+  std::string presetDir = Path::JoinPath({Path::CreateOrGetOrbitAppDataDir(), "presets"});
   std::filesystem::create_directory(presetDir);
   return presetDir;
 }
 
-std::string Path::GetPluginPath() {
-  std::string presetDir = Path::JoinPath({Path::GetAppDataPath(), "plugins"});
-  std::filesystem::create_directory(presetDir);
-  return presetDir;
-}
-
-std::string Path::GetCapturePath() {
-  std::string captureDir = Path::JoinPath({Path::GetAppDataPath(), "output"});
+std::string Path::CreateOrGetCaptureDir() {
+  std::string captureDir = Path::JoinPath({Path::CreateOrGetOrbitAppDataDir(), "output"});
   std::filesystem::create_directory(captureDir);
   return captureDir;
 }
 
-std::string Path::GetDumpPath() {
-  std::string captureDir = Path::JoinPath({Path::GetAppDataPath(), "dumps"});
+std::string Path::CreateOrGetDumpDir() {
+  std::string captureDir = Path::JoinPath({Path::CreateOrGetOrbitAppDataDir(), "dumps"});
   std::filesystem::create_directory(captureDir);
   return captureDir;
 }
 
-std::string Path::GetFileName(const std::string& a_FullName) {
-  std::string FullName = a_FullName;
-  std::replace(FullName.begin(), FullName.end(), '\\', '/');
-  auto index = FullName.find_last_of('/');
+std::string Path::GetFileName(const std::string& file_path) {
+  std::string full_name = file_path;
+  std::replace(full_name.begin(), full_name.end(), '\\', '/');
+  auto index = full_name.find_last_of('/');
   if (index != std::string::npos) {
-    std::string FileName = FullName.substr(FullName.find_last_of('/') + 1);
-    return FileName;
+    std::string file_name = full_name.substr(full_name.find_last_of('/') + 1);
+    return file_name;
   }
 
-  return a_FullName;
+  return full_name;
 }
 
-std::string Path::GetFileNameNoExt(const std::string& a_FullName) {
-  return StripExtension(GetFileName(a_FullName));
+std::string Path::StripExtension(const std::string& file_path) {
+  std::string full_name = file_path;
+  std::replace(full_name.begin(), full_name.end(), '\\', '/');
+  const std::string extension = GetExtension(full_name);
+  return full_name.substr(0, file_path.length() - extension.length());
 }
 
-std::string Path::StripExtension(const std::string& a_FullName) {
-  size_t index = a_FullName.find_last_of('.');
-  if (index != std::string::npos) return a_FullName.substr(0, index);
-  return a_FullName;
-}
-
-std::string Path::GetExtension(const std::string& a_FullName) {
+std::string Path::GetExtension(const std::string& file_path) {
   // returns ".ext" (includes point)
-  size_t index = a_FullName.find_last_of('.');
-  if (index != std::string::npos) return a_FullName.substr(index, a_FullName.length());
+  // Perform on file name to make sure we're not detecting "." within the directory path
+  const std::string file_name = GetFileName(file_path);
+  size_t index = file_name.find_last_of('.');
+  if (index != std::string::npos) return file_name.substr(index, file_name.length());
   return "";
 }
 
-std::string Path::GetDirectory(const std::string& a_FullName) {
-  std::string FullName = a_FullName;
+std::string Path::GetDirectory(const std::string& any_path) {
+  std::string FullName = any_path;
   std::replace(FullName.begin(), FullName.end(), '\\', '/');
   auto index = FullName.find_last_of('/');
   if (index != std::string::npos) {
@@ -200,13 +134,13 @@ std::string Path::GetDirectory(const std::string& a_FullName) {
   return "";
 }
 
-std::string Path::GetParentDirectory(std::string a_Directory) {
-  if (a_Directory.empty()) return "";
-  std::replace(a_Directory.begin(), a_Directory.end(), '\\', '/');
-  wchar_t lastChar = a_Directory.c_str()[a_Directory.size() - 1];
-  if (lastChar == '/') a_Directory.erase(a_Directory.size() - 1);
+std::string Path::GetParentDirectory(std::string any_path) {
+  if (any_path.empty()) return "";
+  std::replace(any_path.begin(), any_path.end(), '\\', '/');
+  wchar_t lastChar = any_path.c_str()[any_path.size() - 1];
+  if (lastChar == '/') any_path.erase(any_path.size() - 1);
 
-  return GetDirectory(a_Directory);
+  return GetDirectory(any_path);
 }
 
 std::string Path::JoinPath(const std::vector<std::string>& parts) {
@@ -220,39 +154,32 @@ std::string Path::JoinPath(const std::vector<std::string>& parts) {
   return joined.string();
 }
 
-std::string Path::GetAppDataPath() {
+std::string Path::CreateOrGetOrbitAppDataDir() {
 #ifdef WIN32
   std::string appData = GetEnvVar("APPDATA");
   std::string path = Path::JoinPath({appData, "OrbitProfiler"});
 #else
-  std::string path = Path::JoinPath({Path::GetHome(), ".orbitprofiler"});
+  std::string path = Path::JoinPath({GetEnvVar("HOME"), ".orbitprofiler"});
 #endif
   std::filesystem::create_directory(path);
   return path;
 }
 
-std::string Path::GetLogFilePath() {
+std::string Path::GetLogFilePathAndCreateDir() {
   std::string logs_dir;
   if (!absl::GetFlag(FLAGS_log_dir).empty()) {
     logs_dir = absl::GetFlag(FLAGS_log_dir);
   } else {
-    logs_dir = Path::JoinPath({Path::GetAppDataPath(), "logs"});
+    logs_dir = Path::JoinPath({Path::CreateOrGetOrbitAppDataDir(), "logs"});
   }
   std::filesystem::create_directory(logs_dir);
   return Path::JoinPath({logs_dir, "Orbit.log"});
 }
 
 std::string Path::GetIconsPath() {
-  static std::string icons_path = JoinPath({GetExecutablePath(), "icons"});
+  static std::string icons_path = JoinPath({GetExecutableDir(), "icons"});
   return icons_path;
 }
-
-#ifdef __linux__
-std::string Path::GetHome() {
-  std::string home = GetEnvVar("HOME") + "/";
-  return home;
-}
-#endif
 
 std::vector<std::string> Path::ListFiles(const std::string& directory,
                                          const std::function<bool(const std::string&)>& filter) {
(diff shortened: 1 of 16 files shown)


OrbitQt/orbitmainwindow.cpp
```diff
@@ -189,7 +189,7 @@ OrbitMainWindow::OrbitMainWindow(QApplication* a_App, ApplicationOptions&& optio
           [this](const QString& text) { OnLiveTabFunctionsFilterTextChanged(text); });
 
   SetTitle({});
-  std::string iconFileName = Path::GetExecutablePath() + "orbit.ico";
+  std::string iconFileName = Path::GetExecutableDir() + "orbit.ico";
```
dimitry-, line 192: Path::Join

OrbitCore/Path.cpp
```diff
@@ -97,13 +92,10 @@ bool Path::DirExists(const std::string& dir) {
 }
 
 std::string Path::GetBasePath() {
```
dimitry-, line 94: this is used only Dllpaht which is not used anywhere else - remove this and GetDllPath and base_path_

OrbitCore/Path.h
```diff
@@ -10,48 +10,47 @@
...
-  static std::string GetCapturePath();
-  static std::string GetDumpPath();
-  static std::string GetAppDataPath();
-  static std::string GetLogFilePath();
-  static std::string GetIconsPath();
-  static void Dump();
+  [[nodiscard]] static std::string GetExecutablePath();
+  [[nodiscard]] static std::string GetExecutableDir();
+  [[nodiscard]] static std::string GetBasePath();
+  [[nodiscard]] static std::string GetDllPath(bool a_Is64Bit);
```
dimitry-, line 16: remove this.

OrbitCore/Path.h
```diff
@@ -10,48 +10,47 @@
...
+  [[nodiscard]] static std::string GetDllName(bool a_Is64Bit);
```
dimitry-, line 17: is_64bit

OrbitCore/Path.h
```diff
@@ -10,48 +10,47 @@
...
+  [[nodiscard]] static std::string GetFileMappingFileName();
+  [[nodiscard]] static std::string GetSymbolsFileName();
+  [[nodiscard]] static std::string CreateAndGetCacheDir();
+  [[nodiscard]] static std::string CreateAndGetPresetDir();
+  [[nodiscard]] static std::string CreateAndGetPluginDir();
```
dimitry-, line 22: this is also not used anywhere - remove it.

OrbitCore/Path.h
```diff
@@ -10,48 +10,47 @@
...
 
-  static std::string GetFileName(const std::string& a_FullName);
-  static std::string GetFileNameNoExt(const std::string& a_FullName);
-  static std::string StripExtension(const std::string& a_FullName);
-  static std::string GetExtension(const std::string& a_FullName);
-  static std::string GetDirectory(const std::string& a_FullName);
-  static std::string GetParentDirectory(std::string a_FullName);
-  static std::string JoinPath(const std::vector<std::string>& parts);
+  [[nodiscard]] static std::string GetFileName(const std::string& a_FullName);
+  [[nodiscard]] static std::string GetFileNameNoExt(const std::string& a_FullName);
```
dimitry-, line 34: unused - remove?

OrbitCore/Path.h
```diff
@@ -10,48 +10,47 @@
...
 #endif
 
-  static std::string GetFileName(const std::string& a_FullName);
-  static std::string GetFileNameNoExt(const std::string& a_FullName);
-  static std::string StripExtension(const std::string& a_FullName);
-  static std::string GetExtension(const std::string& a_FullName);
-  static std::string GetDirectory(const std::string& a_FullName);
-  static std::string GetParentDirectory(std::string a_FullName);
-  static std::string JoinPath(const std::vector<std::string>& parts);
+  [[nodiscard]] static std::string GetFileName(const std::string& a_FullName);
```
dimitry-, line 33: full_path

OrbitCore/Path.h
```diff
@@ -10,48 +10,47 @@
...
+  [[nodiscard]] static std::string GetFileNameNoExt(const std::string& a_FullName);
+  [[nodiscard]] static std::string StripExtension(const std::string& a_FullName);
```
dimitry-, line 35: and here... and below.

OrbitCore/Path.h
```diff
@@ -10,48 +10,47 @@
...
+  [[nodiscard]] static std::string GetExtension(const std::string& a_FullName);
+  [[nodiscard]] static std::string GetDirectory(const std::string& a_FullName);
+  [[nodiscard]] static std::string GetParentDirectory(std::string a_FullName);
```
dimitry-, line 38: not used

OrbitCore/Path.h
```diff
@@ -10,48 +10,47 @@
...
+  [[nodiscard]] static std::string JoinPath(const std::vector<std::string>& parts);
 
   // TODO(159868905): This is only used by OrbitModule. Remove this as soon as
   //  OrbitModule is removed.
-  static uint64_t FileSize(const std::string& a_File);
-  static bool DirExists(const std::string& a_Dir);
+  [[nodiscard]] static uint64_t FileSize(const std::string& a_File);
```
dimitry-, line 43: FileSize is not used anymore - remove it.

OrbitCore/Path.h
```diff
@@ -10,48 +10,47 @@
...
-  static std::string GetFileNameNoExt(const std::string& a_FullName);
-  static std::string StripExtension(const std::string& a_FullName);
-  static std::string GetExtension(const std::string& a_FullName);
-  static std::string GetDirectory(const std::string& a_FullName);
-  static std::string GetParentDirectory(std::string a_FullName);
-  static std::string JoinPath(const std::vector<std::string>& parts);
+  [[nodiscard]] static std::string GetFileName(const std::string& a_FullName);
+  [[nodiscard]] static std::string GetFileNameNoExt(const std::string& a_FullName);
+  [[nodiscard]] static std::string StripExtension(const std::string& a_FullName);
+  [[nodiscard]] static std::string GetExtension(const std::string& a_FullName);
```
dimitry-, line 36: and this is also unused.

OrbitCore/Path.cpp
antonrohr: this result then in Path not having any state anymore and only static functions. You can then consider changing it from a class to a namespace Path. (I have a slight preference towards namespace instead static class, but do not care to much)

OrbitCore/Path.h
```diff
@@ -10,48 +10,47 @@
...
+  [[nodiscard]] static std::string CreateAndGetPluginDir();
+  [[nodiscard]] static std::string CreateAndGetCaptureDir();
+  [[nodiscard]] static std::string CreateAndGetDumpDir();
+  [[nodiscard]] static std::string CreateAndGetOrbitAppDataDir();
+  [[nodiscard]] static std::string GetLogFilePathAndCreateDir();
+  [[nodiscard]] static std::string GetIconsPath();
 
 #ifdef __linux__
-  static std::string GetHome();
+  [[nodiscard]] static std::string GetHome();
```
antonrohr, line 30: `GetHome()` is only used in `CreateAndGetOrbitAppDataDir()`. Maybe put the functionality of `GetHome()` into `CreateAndGetOrbitAppDataDir()` directly. Then we don't need this last ifdef in this header and only have ifdefs in the cpp file.

OrbitCore/PathTest.cpp
```diff
@@ -96,3 +96,171 @@ TEST(Path, JoinPathPartsAbsoluteRootBackslash) {
...
+#else
+  V tests = {{"/dir/With/mult_ext.txt.ext", "/dir/With/", "/dir/With/", "/dir/"},
+             {"/no/ext/file", "/no/ext/", "/no/ext/", "/no/"},
+             {"/dir/subdir/", "/dir/subdir/", "/dir/", "/dir/"},
+             {"/file_in_root", "/", "/", ""},
+             {"broken_path", "", "", ""},
+             {"/", "/", "", ""}};
+#endif
+
+  for (size_t i = 0; i < tests.size(); ++i) {
```
antonrohr, line 183: Nit: 
```suggestion
  for (const auto& test: tests) {
```

OrbitCore/PathTest.cpp
```diff
@@ -96,3 +96,171 @@ TEST(Path, JoinPathPartsAbsoluteRootBackslash) {
...
+
+// TODO: This would be more meaningful with absolute paths that are
+// created explicitely for the test
+TEST(Path, DirExists) {
+  auto dir = Path::GetDirectory(Path::GetExecutableDir());
+  EXPECT_TRUE(Path::DirExists(dir));
+  EXPECT_FALSE(Path::DirExists(Path::JoinPath({dir, "__invalid_subdir__"})));
+}
+
+TEST(Path, FileSize) {
```
antonrohr, line 200: Maybe add a test for FileSize of non-existing file. Currently the result is 0 in that case.

OrbitCore/PathTest.cpp
```diff
@@ -96,3 +96,171 @@ TEST(Path, JoinPathPartsAbsoluteRootBackslash) {
...
+}
+
+TEST(Path, AllAutoCreatedDirsExist) {
+  auto test_fns = {Path::CreateAndGetOrbitAppDataDir, Path::CreateAndGetDumpDir,
+                   Path::CreateAndGetPluginDir,       Path::CreateAndGetPresetDir,
+                   Path::CreateAndGetCacheDir,        Path::CreateAndGetOrbitAppDataDir};
+
+  for (auto fn : test_fns) {
+    std::string path = fn();
+    std::cout << "Testing existence of \"" + path + "\"\n";
```
antonrohr, line 239: Maybe use LOG instead of cout

OrbitCore/PathTest.cpp
```diff
@@ -96,3 +96,171 @@ TEST(Path, JoinPathPartsAbsoluteRootBackslash) {
...
+    const std::string result_dir = Path::GetDirectory(tests[i].input);
+    EXPECT_EQ(result_dir, tests[i].dir);
+    EXPECT_EQ(Path::GetParentDirectory(tests[i].input), tests[i].parent_of_input);
+    EXPECT_EQ(Path::GetParentDirectory(result_dir), tests[i].parent_of_dir);
+  }
+}
+
+// TODO: This would be more meaningful with absolute paths that are
+// created explicitely for the test
+TEST(Path, DirExists) {
```
antonrohr, line 194: Maybe add a test that checks the result of DirExist when called with a file instead of a directory

OrbitFramePointerValidator/FramePointerValidatorTest.cpp
```diff
@@ -16,7 +16,7 @@ using orbit_grpc_protos::CodeBlock;
 using orbit_grpc_protos::SymbolInfo;
 
 TEST(FramePointerValidator, GetFpoFunctions) {
-  std::string executable_path = Path::GetExecutablePath();
+  std::string executable_path = Path::GetExecutableDir();
```
antonrohr, line 19: Nit:
```suggestion
  std::string executable_dit = Path::GetExecutableDir();
```

OrbitGl/ImGuiOrbit.cpp
```diff
@@ -582,7 +582,7 @@ void Orbit_ImGui_RenderDrawLists(ImDrawData* draw_data) {
 }
 
 ImFont* AddOrbitFont(float pixel_size) {
-  const auto exe_path = Path::GetExecutablePath();
+  const auto exe_path = Path::GetExecutableDir();
```
antonrohr, line 585: Nit:
```suggestion
  const auto exe_dir = Path::GetExecutableDir();
```

OrbitGl/TextRenderer.cpp
```diff
@@ -55,7 +55,7 @@ void TextRenderer::Init() {
   int atlasSize = 2 * 1024;
   m_Atlas = texture_atlas_new(atlasSize, atlasSize, 1);
 
-  const auto exePath = Path::GetExecutablePath();
+  const auto exePath = Path::GetExecutableDir();
```
antonrohr, line 58: Nit:
```suggestion
  const auto exe_dir = Path::GetExecutableDir();
```

antonrohr (commented):
Thank you a lot for this work. Especially the tests are really helpful!

I agree with @dimitry-'s comments an I have added a few more. 

Please also remove the `#pragma once` in Path.h. 

If you want to do even more work, you could change the whole thing to std::filesystem::path instead of std::string, this makes some small things nicer like concatenating paths with the / operator

OrbitCore/Path.h
```diff
@@ -10,48 +10,47 @@
...
+  [[nodiscard]] static std::string GetDllPath(bool a_Is64Bit);
+  [[nodiscard]] static std::string GetDllName(bool a_Is64Bit);
+  [[nodiscard]] static std::string GetFileMappingFileName();
+  [[nodiscard]] static std::string GetSymbolsFileName();
+  [[nodiscard]] static std::string CreateAndGetCacheDir();
+  [[nodiscard]] static std::string CreateAndGetPresetDir();
+  [[nodiscard]] static std::string CreateAndGetPluginDir();
+  [[nodiscard]] static std::string CreateAndGetCaptureDir();
+  [[nodiscard]] static std::string CreateAndGetDumpDir();
+  [[nodiscard]] static std::string CreateAndGetOrbitAppDataDir();
```
pierricgimmig, line 25: Since the create operation will happen only once, should we rename to CreateOrGetDir instead of CreateANDGet? We use the GetOrCreateX pattern elsewhere in the code, GetOrCreateThreadTrack for example, maybe we should do the same here for uniformity.

OrbitCore/Path.h
pierricgimmig: if abseil or filesystem has the equivalent, I think we should remove them yes

reichlfl: I think this is obsolete, not only because it's unused, but because `StripExtension(GetFileName())` provides the same. Removed.

OrbitCore/Path.h
reichlfl: And there's also `std::filesystem::file_size` in case we need it, so removed this.

OrbitCore/Path.h
reichlfl: I'd argue we should keep this one. It's a method that fits perfectly in the scope of this utility class and neither abseil not std::filesystem provides it. I know we can always "get it back" from Git, but that requires someone to know it was there in the first place.

dimitry-: I do not know what is the use-case for this one.

OrbitCore/PathTest.cpp
```diff
@@ -96,3 +96,171 @@ TEST(Path, JoinPathPartsAbsoluteRootBackslash) {
...
+  EXPECT_EQ(Path::StripExtension(input), "no/ext");
+  EXPECT_EQ(Path::GetFileName(input), "ext");
+  EXPECT_EQ(Path::GetFileNameNoExt(input), "ext");
+
+  input = "double.ext.txt";
+  EXPECT_EQ(Path::GetExtension(input), ".txt");
+  EXPECT_EQ(Path::StripExtension(input), "double.ext");
+  EXPECT_EQ(Path::StripExtension(Path::StripExtension(input)), "double");
+  EXPECT_EQ(Path::GetFileName(input), "double.ext.txt");
+  EXPECT_EQ(Path::GetFileNameNoExt(input), "double.ext");
```
dimitry-, line 138: add test for StripExtenstion("/some/dir.with.dots/filename_without_dots");

reichlfl: Good point, and that actually uncovered a bug. Added the test and fixed this issue.

OrbitQt/orbitmainwindow.cpp
reichlfl: Done

OrbitCore/Path.cpp
reichlfl: True, I changed it to a namespace now, thanks

OrbitCore/Path.h
reichlfl: Done

OrbitCore/Path.h
reichlfl: GetDllName is removed anyways

OrbitCore/Path.h
reichlfl: Done

OrbitCore/Path.h
reichlfl: I adjusted all those parameters also to hint at if they expect a file path, directory, or any path

OrbitCore/Path.h
reichlfl: It's now actually used in the new implementation of `StripExtension()` (bugfix for the unittest below), it makes that one quite straightforward and easy to follow...

OrbitCore/Path.h
reichlfl: Changed, GetHome() was a one-line anyways

OrbitCore/PathTest.cpp
reichlfl: Done

OrbitCore/PathTest.cpp
reichlfl: Removed FileSize, so the test is obsolete

OrbitCore/PathTest.cpp
reichlfl: Removed DirExists, so the test is obsolete

OrbitFramePointerValidator/FramePointerValidatorTest.cpp
reichlfl: Done

OrbitGl/TextRenderer.cpp
reichlfl: Done

OrbitCore/Path.h
reichlfl: True, I changed this to be more consistent

OrbitCore/Path.h
reichlfl: Given that there's std::filesystem::path::parent_path() to provide this if we need it in the future, I removed it, thanks.

reichlfl:
@antonrohr @pierricgimmig @dimitry-  I think I addressed all comments now, let me know if I missed something. I generally do not think we should strictly remove every utility method that is not used if it makes sense to be in the class / namespace - but I removed most (I think all?) of the mentioned ones now since they were either super specific and won't be missed, or std::filesystem provides an alternative anyways. Thanks for the thorough feedback!

OrbitCore/PathTest.cpp
```diff
@@ -96,3 +97,151 @@ TEST(Path, JoinPathPartsAbsoluteRootBackslash) {
...
+  input = "no/ext";
+  EXPECT_EQ(Path::GetExtension(input), "");
+  EXPECT_EQ(Path::StripExtension(input), "no/ext");
+  EXPECT_EQ(Path::GetFileName(input), "ext");
+
+  input = "double.ext.txt";
+  EXPECT_EQ(Path::GetExtension(input), ".txt");
+  EXPECT_EQ(Path::StripExtension(input), "double.ext");
+  EXPECT_EQ(Path::StripExtension(Path::StripExtension(input)), "double");
+  EXPECT_EQ(Path::GetFileName(input), "double.ext.txt");
```
dimitry-, line 136: In general these should also work on linux-style directory separators on windows.. I do not think they should be platform-specific. It is not uncommon to parse a string we recieved via grpc from linux on windows. Or paths from .debug_info section which could be whatever os was using while compiling them.

OrbitCore/PathTest.cpp
```diff
@@ -96,3 +97,151 @@ TEST(Path, JoinPathPartsAbsoluteRootBackslash) {
...
 }
+
+TEST(Path, FileNamesAndExtensions) {
+#ifdef _WIN32
+  std::string abs_input[] = {"C:\\some_abs_path\\file.txt", "C:\\dir.with.dots\\file_without_ext"};
+  std::string abs_stripped[] = {"C:\\some_abs_path\\file", "C:\\dir.with.dots\\file_without_ext"};
+#else
+  std::string abs_input[] = {"/some_abs_path/file.txt", "/dir.with.dots/file_without_ext"};
+  std::string abs_stripped[] = {"/some_abs_path/file", "/dir.with.dots/file_without_ext"};
+#endif
```
dimitry-, line 108: also maybe remove this ifdefs and test for all 4 cases on both platforms?

OrbitCore/PathTest.cpp
reichlfl: The test actually covers the first part (forward slash on windows) - only the first input is platform specific to check if both platforms handle their respective "native" formats for absolute paths correctly. All other tests are using linux-style directories for both platforms.

I changed one of the tests to cover a mixed-style path for both platforms. This exposed that the `StripExtension` isn't consistent with the other methods - all other methods that return a path convert it to contain only forward-slashes, so I changed `StripExtension` to do that as well, WDYT?

OrbitCore/PathTest.cpp
reichlfl: Right, actually that's the nicer solution. I've removed this, tests are now executed the same on both OS and should yield the same results for window-, unix- and mixed-style paths

dimitry- (approved):

OrbitCore/SymbolHelper.cpp
```diff
@@ -52,7 +52,7 @@ std::vector<fs::path> ReadSymbolsFile() {
       if (absl::StartsWith(line, "//") || line.empty()) continue;
 
       const fs::path& dir = line;
-      if (fs::exists(dir)) {
+      if (std::filesystem::is_directory(dir)) {
```
antonrohr, line 55: Nit: 
```suggestion
      if (fs::is_directory(dir)) {
```

antonrohr (approved):

PR #1132, opened 2020-08-27
Fix potential crash in OrbitGgp

beckerhe:
This change prevents an asynchronously executed callback from accessing
a potentially already deleted object of type QProcess.

There are two checks in this commit since we have to check before and
after custom user code is executed which could potentially delete the
QProcess object.

I hope this will fix a crash reported in http://b/166385260, but due to
the asynchronous nature of that problem we can't say for sure.

Tested manually. The timeout works fine.


antonrohr (approved):
Thanks for taking care of this! Lets hope that solves the issue

PR #1133, opened 2020-08-27
Remove left over tracepoint related code from ProcessClient.

danielfenner:


florian-kuebler:
Super nit-picking comment regarding our new commit message guideline (go/stadia-orbit-commit):
The message should contain a way to test (Test:) And for the title should be 
* Capitalized (starts with a capital letter)
* Doesn’t end with a period

But anyways. Thanks for removing that dead code here. LGTM.

florian-kuebler (approved):

PR #1135, opened 2020-08-28
Fix bug occuring when applying preset.

danielfenner:
This fixes two issues:

When applying a preset and the process was already selected there was a redraw
missing.

When applying a preset and a different process was selected the hooked
functions were cleared after the preset was applied.

Bug: b/166450212
Test: Ran locally, works fine now.


IrinaShkviro (approved):
Thanks for the fix!

PR #1136, opened 2020-08-31
Simplify Deserialization of Address Info

florian-kuebler:
This directly insert the address infos in capture data instead
of creating the internal data structure in serialization already.
Also this fix deserialization issues with very recent old captures.


dimitry- (approved):
It would be nice to have a test for it.

florian-kuebler:
Do you mean for that particular part or for capture serialization in general?

dimitry-:
Both :)

IrinaShkviro (approved):

florian-kuebler:
I'll add tests for both together with the upcoming refactorings and merge this for now..

PR #1138, opened 2020-08-31
Fix Crash Unselecting Multiple Functions

florian-kuebler:
When selecting multiple functions (some of which are hooked, others not),
the UI will allow you to unhook (unselect) all. For simplicity
DeselectFunction will be called for all of them.
Prior to this, there was a CHECK, that would fail then.

Test: Compile
Bug: http://167197303


beckerhe (approved):
Thanks!

PR #1139, opened 2020-08-31
Refactoring and removing the support for the sched_switch

MariaSandru27:
tracepoint in favour of a generalization for all tracepoints

### DIFF
--- a/OrbitGl/CMakeLists.txt
+++ b/OrbitGl/CMakeLists.txt
@@ -5,7 +5,7 @@
 cmake_minimum_required(VERSION 3.15)
 
 project(OrbitGl CXX)
-add_library(OrbitGl STATIC)
+add_library(OrbitGl STATIC TracepointCustom.h)
 
 target_compile_options(OrbitGl PRIVATE ${STRICT_COMPILE_FLAGS})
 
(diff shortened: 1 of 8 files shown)


OrbitGl/CMakeLists.txt
```diff
@@ -5,7 +5,7 @@
 cmake_minimum_required(VERSION 3.15)
 
 project(OrbitGl CXX)
-add_library(OrbitGl STATIC)
+add_library(OrbitGl STATIC TracepointCustom.h)
```
dimitry-, line 8: move this to target_sources below

dimitry- (approved):

OrbitGl/TracepointCustom.h
```diff
@@ -0,0 +1,25 @@
+// Copyright (c) 2020 The Orbit Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#ifndef ORBIT_TRACEPOINTCUSTOM_H
+#define ORBIT_TRACEPOINTCUSTOM_H
+
+#include "tracepoint.pb.h"
+
+namespace internal {
```
danielfenner, line 10: By convention stuff from internal namespaces should only occur in in the header where it is defined (or the related .cc). If you need to make these functors more widely  available remove the "namespace internal {".

MariaSandru27: Sure. Thank you so much! I will modify it right now.

danielfenner (approved):
LGTM!

PR #1140, opened 2020-08-31
Add filter icons and placeholder text to all search boxes

reichlfl:
Small UI changes to add icons to the filter boxes, and placeholder texts to make it obvious that those can be used to filter. Moved some more parts of the UI into the UI files.

Screenshots:
image
image


beckerhe (approved):
This looks a lot cleaner than before. Thanks. Code looks also good.

Just one thing that's maybe worth thinking about. Material design suggests to have the search icon on the left (https://material.io/design/navigation/search.html). I'm not sure what's the long term plan but if it is moving to Material, then it means we have to change that eventually once more.

EDIT: I've just checked Visual Studio. They use the icon on the right, so one could also argue we integrate with that product.

reichlfl:
> This looks a lot cleaner than before. Thanks. Code looks also good.
> 
> Just one thing that's maybe worth thinking about. Material design suggests to have the search icon on the left (https://material.io/design/navigation/search.html). I'm not sure what's the long term plan but if it is moving to Material, then it means we have to change that eventually once more.
> 
> EDIT: I've just checked Visual Studio. They use the icon on the right, so one could also argue we integrate with that product.

@beckerhe Good point - actually, if we have them left-aligned, this would easily allow us to enable the "clear button" of the Textboxes, which doesn't work with the right-aligned icons (because they are then underneath the clear buttons, and the buttons are not easily modifyable).

On the downside, I think alignment looks a bit off:
image

This is already the "fixed" alignment where I manually padded the left of the icon by 2 pixels, because apparently for QT the background starts at the outer left of the textbox.

WDYT?

beckerhe:
> > This looks a lot cleaner than before. Thanks. Code looks also good.
> > Just one thing that's maybe worth thinking about. Material design suggests to have the search icon on the left (https://material.io/design/navigation/search.html). I'm not sure what's the long term plan but if it is moving to Material, then it means we have to change that eventually once more.
> > EDIT: I've just checked Visual Studio. They use the icon on the right, so one could also argue we integrate with that product.
> 
> @beckerhe Good point - actually, if we have them left-aligned, this would easily allow us to enable the "clear button" of the Textboxes, which doesn't work with the right-aligned icons (because they are then underneath the clear buttons, and the buttons are not easily modifyable).
> 
> On the downside, I think alignment looks a bit off:
> image
> 
> This is already the "fixed" alignment where I manually padded the left of the icon by 2 pixels, because apparently for QT the background starts at the outer left of the textbox.
> 
> WDYT?

I like it. I also think it's better than the first version, but that's not a strong opinion. And I have to admit this feeling of familiarity probably comes from websites like stack overflow, gmail, etc. Maybe that's not a good data point.

reichlfl:
> I like it. I also think it's better than the first version, but that's not a strong opinion. And I have to admit this feeling of familiarity probably comes from websites like stack overflow, gmail, etc. Maybe that's not a good data point.

I'm sticking with this, especially because the clear-button is very nice to have. Thanks!

PR #1141, opened 2020-08-31
Load symbols off the main thread

dimitry-:
With line number symbol loading is getting slower and for
large binaries when the main thread is in suspension for long
enough the tunnel generates socket error on the next request
(it could have something to do with tcp timeout).

With recent code refactoring it now possible to do symbol loading off
the main thread.

Test: Start Orbit, load symbols see that loading is done off
      the main thread.
Bug: http://b/156460883

### DIFF
--- a/OrbitGl/App.cpp
+++ b/OrbitGl/App.cpp
@@ -843,38 +843,41 @@ void OrbitApp::LoadSymbols(const std::filesystem::path& symbols_path,
                            const std::shared_ptr<Process>& process,
                            const std::shared_ptr<Module>& module,
                            const std::shared_ptr<PresetFile>& preset) {
-  // TODO (159889010) Move symbol loading off the main thread.
-  const auto scoped_status = CreateScopedStatus(absl::StrFormat(
+  auto scoped_status = CreateScopedStatus(absl::StrFormat(
       R"(Loading symbols for "%s" from file "%s"...)", module->m_FullName, symbols_path.string()));
+  thread_pool_->Schedule([this, scoped_status = std::move(scoped_status), symbols_path, process,
+                          module, preset]() mutable {
+    auto symbols_result = SymbolHelper::LoadSymbolsFromFile(symbols_path);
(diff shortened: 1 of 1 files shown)


OrbitGl/App.cpp
```diff
@@ -843,38 +843,41 @@ void OrbitApp::LoadSymbols(const std::filesystem::path& symbols_path,
...
+      process->AddFunctions(module->m_Pdb->GetFunctions());
+      LOG("Loaded %lu function symbols for module \"%s\"", symbols.symbol_infos().size(),
+          module->m_FullName);
+
+      // Applying preset
+      if (preset != nullptr) {
+        auto it = preset->preset_info().path_to_module().find(module->m_FullName);
+        if (it != preset->preset_info().path_to_module().end()) {
+          for (const FunctionInfo* func : module->m_Pdb->GetSelectedFunctionsFromPreset(*preset)) {
+            SelectFunction(*func);
```
florian-kuebler, line 866: I believe, selecting functions require to be done on the main thread.

florian-kuebler (commented):
Thanks for the change. As you said, with the recent changes, it should be fine to load the symbols off the main thread. However, selecting functions currently assumes to be done on the main thread. Easy solution would be to schedule the apply preset part again on main thread.

OrbitGl/App.cpp
dimitry-: it is done on the main thread - the only part which is done of the main thread is line 850

dimitry-: all legacy structures - like Module and Process are also modified on the main thread. Only loading from the elf-file has moved to thread-pool.

florian-kuebler: Ah thanks. I missed line 852. Perfect. Then it seems to be all fine for me.

florian-kuebler (approved):

florian-kuebler:
LGTM, thanks for the explanation and for that change.

PR #1142, opened 2020-08-31
Remove process from preset

IrinaShkviro:
- Removed information about process from the Preset.
- Presets could be applied partially from now (the user will see the warning if preset was applied partially, and error if it can't be applied).
- Updated PresetDataView: add information about modules, hooked functions count, and if preset could be load for the selected process or not (also font color different depending on this parameter).

Presets new

Test: Start Orbit, save several presets, changed process (check that `loadable` changed), apply preset partially (get a warning, some modules loaded), apply not-loadable preset (get an error, nothing changed).
Bug: http://b/160282419

### DIFF
--- a/OrbitGl/App.cpp
+++ b/OrbitGl/App.cpp
@@ -60,6 +60,34 @@ using orbit_grpc_protos::ModuleSymbols;
 using orbit_grpc_protos::ProcessInfo;
 using orbit_grpc_protos::TracepointInfo;
 
+namespace {
+PresetLoadState GetPresetLoadStateForProcess(
+    const std::shared_ptr<orbit_client_protos::PresetFile>& preset,
+    const std::shared_ptr<Process>& process) {
+  if (process == nullptr) {
+    return PresetLoadState::kNotLoadable;
+  }
+
+  int modules_not_found_count = 0;
+  for (const auto& pair : preset->preset_info().path_to_module()) {
+    const std::string& module_path = pair.first;
+    if (process->GetModuleFromPath(module_path) == nullptr) {
+      modules_not_found_count++;
+    }
+  }
+
+  if (modules_not_found_count == preset->preset_info().path_to_module_size()) {
+    return PresetLoadState::kNotLoadable;
+  }
+
+  if (modules_not_found_count == 0) {
+    return PresetLoadState::kLoadable;
+  }
+
+  return PresetLoadState::kPartiallyLoadable;
+}
+}  // namespace
+
 std::unique_ptr<OrbitApp> GOrbitApp;
 bool DoZoom = false;
 
@@ -556,6 +584,12 @@ void OrbitApp::LoadPreset(const std::shared_ptr<PresetFile>& preset) {
   } else {
     SendErrorToUi("Preset loading failed", "Process is not selected");
   }
+}
+
+PresetLoadState OrbitApp::GetPresetLoadState(
+    const std::shared_ptr<orbit_client_protos::PresetFile>& preset) const {
+  const std::shared_ptr<Process> selected_process = GetSelectedProcess();
+  return GetPresetLoadStateForProcess(preset, selected_process);
 }
 
 ErrorMessageOr<void> OrbitApp::OnSaveCapture(const std::string& file_name) {
--- a/OrbitGl/App.cpp
+++ b/OrbitGl/App.cpp
@@ -552,7 +552,7 @@ void OrbitApp::LoadPreset(const std::shared_ptr<PresetFile>& preset) {
   const int32_t selected_process_id = GetSelectedProcessId();
   const ProcessData* selected_process = data_manager_->GetProcessByPid(selected_process_id);
   if (selected_process != nullptr) {
-    GOrbitApp->LoadModulesFromPreset(GetSelectedProcess(), preset);
+    LoadModulesFromPreset(GetSelectedProcess(), preset);
   } else {
     SendErrorToUi("Preset loading failed", "Process is not selected");
   }
@@ -698,6 +698,14 @@ void OrbitApp::SendInfoToUi(const std::string& title, const std::string& text) {
   main_thread_executor_->Schedule([this, title, text] {
     if (info_message_callback_) {
       info_message_callback_(title, text);
+    }
+  });
+}
+
+void OrbitApp::SendWarningToUi(const std::string& title, const std::string& text) {
+  main_thread_executor_->Schedule([this, title, text] {
+    if (warning_message_callback_) {
+      warning_message_callback_(title, text);
     }
   });
 }
@@ -872,11 +880,18 @@ void OrbitApp::LoadModulesFromPreset(const std::shared_ptr<Process>& process,
     modules_to_load.emplace_back(std::move(module));
   }
   if (!modules_not_found.empty()) {
-    SendErrorToUi(
-        "Preset loading incomplete",
-        absl::StrFormat("Unable to load the preset for the following modules:\n\"%s\"\nThe "
-                        "modules are not loaded by process \"%s\".",
-                        absl::StrJoin(modules_not_found, "\"\n\""), process->GetName()));
+    if (static_cast<int>(modules_not_found.size()) == preset->preset_info().path_to_module_size()) {
+      // no modules were loaded
+      SendErrorToUi(
+          "Preset loading failed",
+          absl::StrFormat("None of the modules in the preset are loaded by the process \"%s\".",
+                          process->GetName()));
+    } else {
+      SendWarningToUi(
+          "Preset only partially loaded",
+          absl::StrFormat("The following modules are not loaded by the process \"%s\":\n\"%s\"",
+                          process->GetName(), absl::StrJoin(modules_not_found, "\"\n\"")));
+    }
   }
   if (!modules_to_load.empty()) {
     LoadModules(process, modules_to_load, preset);
--- a/OrbitGl/App.cpp
+++ b/OrbitGl/App.cpp
@@ -489,7 +489,6 @@ ErrorMessageOr<void> OrbitApp::SavePreset(const std::string& filename) {
   PresetInfo preset;
   const int32_t pid = GetSelectedProcessId();
   const std::shared_ptr<Process>& process = FindProcessByPid(pid);
-  preset.set_process_full_path(data_manager_->GetProcessByPid(pid)->full_path());
 
   for (uint64_t function_address : data_manager_->selected_functions()) {
     FunctionInfo* func = process->GetFunctionFromAddress(function_address);
@@ -552,16 +551,10 @@ ErrorMessageOr<void> OrbitApp::OnLoadPreset(const std::string& filename) {
 void OrbitApp::LoadPreset(const std::shared_ptr<PresetFile>& preset) {
   const int32_t selected_process_id = GetSelectedProcessId();
   const ProcessData* selected_process = data_manager_->GetProcessByPid(selected_process_id);
-  const std::string& process_full_path = preset->preset_info().process_full_path();
-  if (selected_process != nullptr && selected_process->full_path() == process_full_path) {
-    // In case we already have the correct process selected
+  if (selected_process != nullptr) {
     GOrbitApp->LoadModulesFromPreset(GetSelectedProcess(), preset);
-    return;
-  }
-  if (!SelectProcess(Path::GetFileName(process_full_path), preset)) {
-    SendErrorToUi("Preset loading failed",
-                  absl::StrFormat("The process \"%s\" is not running.", process_full_path));
-    return;
+  } else {
+    SendErrorToUi("Preset loading failed", "Process is not selected");
   }
 }
 
--- a/OrbitGl/App.cpp
+++ b/OrbitGl/App.cpp
@@ -669,10 +669,9 @@ void OrbitApp::ToggleCapture() {
   }
 }
 
-bool OrbitApp::SelectProcess(const std::string& process,
-                             const std::shared_ptr<PresetFile>& preset) {
+bool OrbitApp::SelectProcess(const std::string& process) {
   if (processes_data_view_) {
-    return processes_data_view_->SelectProcess(process, preset);
+    return processes_data_view_->SelectProcess(process);
   }
 
   return false;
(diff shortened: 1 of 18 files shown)


OrbitGl/PresetsDataView.cpp
```diff
@@ -55,17 +101,20 @@ std::string PresetsDataView::GetToolTip(int row, int /*column*/) {
     return OrbitUtils::Compare(presets_[a]->Member, presets_[b]->Member, ascending); \
   }
 
+#define ORBIT_PRESET_FUNC_SORT(Func) \
```
ronaldfw, line 104: This, and also ORBIT_PRESET_SORT, above are only used in one place, which maybe defeats the purpose of defining these macros. It might be better to have the defined only locally where they are used. (Or maybe I'm missing the reason for defining them as a macro here.)

OrbitGl/App.cpp
```diff
@@ -552,17 +551,40 @@ ErrorMessageOr<void> OrbitApp::OnLoadPreset(const std::string& filename) {
...
   const ProcessData* selected_process = data_manager_->GetProcessByPid(selected_process_id);
-  const std::string& process_full_path = preset->preset_info().process_full_path();
-  if (selected_process != nullptr && selected_process->full_path() == process_full_path) {
-    // In case we already have the correct process selected
-    GOrbitApp->LoadModulesFromPreset(GetSelectedProcess(), preset);
-    return;
+  if (selected_process != nullptr) {
+    LoadModulesFromPreset(GetSelectedProcess(), preset);
+  } else {
+    SendErrorToUi("Preset loading failed", "Process is not selected");
```
beckerhe, line 557: This feels a bit wrong. Since we knew that upfront we could have told the user more clearly by disabling the preset-list when no process is selected (and adding a tooltip). But since this is not a regression I'd suggest we create a bug for that and leave it like it is.

OrbitGl/App.h
```diff
@@ -199,11 +205,12 @@ class OrbitApp final : public DataViewFactory, public CaptureListener {
...
                              const std::shared_ptr<orbit_client_protos::PresetFile>& preset);
-  void UpdateProcessAndModuleList(int32_t pid,
-                                  const std::shared_ptr<orbit_client_protos::PresetFile>& preset);
+  void UpdateProcessAndModuleList(int32_t pid);
 
   void UpdateSamplingReport();
-  void LoadPreset(const std::shared_ptr<orbit_client_protos::PresetFile>& session);
+  void LoadPreset(const std::shared_ptr<orbit_client_protos::PresetFile>& preset);
+  PresetLoadState GetPresetLoadState(
+      const std::shared_ptr<orbit_client_protos::PresetFile>& preset);
```
beckerhe, line 213: Can this function be `const`?

OrbitGl/App.h
```diff
@@ -61,6 +61,8 @@ ABSL_DECLARE_FLAG(bool, enable_tracepoint_feature);
 
 class Process;
 
+enum PresetLoadState { kLoadable, kPartiallyLoadable, kNotLoadable };
```
beckerhe, line 64: This should be an `enum class`

OrbitGl/PresetsDataView.cpp
```diff
@@ -18,15 +18,57 @@
...
 
 using orbit_client_protos::PresetFile;
+using orbit_client_protos::PresetInfo;
 
 PresetsDataView::PresetsDataView() : DataView(DataViewType::kPresets) {}
 
+std::string PresetsDataView::GetModulesList(const std::vector<ModuleView>& modules) const {
+  std::string modules_list;
+  for (const auto& module : modules) {
+    if (modules_list.empty()) {
```
ronaldfw, line 28: I think you can make this (and below) a bit more elegant by using absl::StrJoin (https://abseil.io/tips/36):
```
std::vector<std::string_view> modules_strings;
for (const auto& module : modules) {
  modules_strings.push_back(module.module_name);
}
return absl::StrJoin(modules_strings.begin(), modules_strings.end(), "\n");

```
}

OrbitGl/App.cpp
```diff
@@ -552,17 +551,40 @@ ErrorMessageOr<void> OrbitApp::OnLoadPreset(const std::string& filename) {
...
   }
-  if (!SelectProcess(Path::GetFileName(process_full_path), preset)) {
-    SendErrorToUi("Preset loading failed",
-                  absl::StrFormat("The process \"%s\" is not running.", process_full_path));
-    return;
+}
+
+PresetLoadState OrbitApp::GetPresetLoadState(
+    const std::shared_ptr<orbit_client_protos::PresetFile>& preset) {
```
beckerhe, line 562: I think this should be a free (or static) function, taking the selected process as a second argument. Ideally there would also be a test for this free function.

beckerhe: But the test could also be a follow-up.

OrbitGl/PresetsDataView.cpp
```diff
@@ -18,15 +18,57 @@
...
+  std::string function_count_list;
+  for (const auto& module : modules) {
+    if (function_count_list.empty()) {
+      absl::StrAppend(&function_count_list, module.function_count);
+    } else {
+      absl::StrAppend(&function_count_list, "\n", module.function_count);
+    }
+  }
+  return function_count_list;
+}
```
beckerhe, line 47: This could be simplified by using `return absl::StrJoin(modules, "\n", [](const auto& module){ return module.function_count; });`.

OrbitGl/PresetsDataView.cpp
```diff
@@ -18,15 +18,57 @@
...
+  std::string modules_list;
+  for (const auto& module : modules) {
+    if (modules_list.empty()) {
+      absl::StrAppend(&modules_list, module.module_name);
+    } else {
+      absl::StrAppend(&modules_list, "\n", module.module_name);
+    }
+  }
+  return modules_list;
+}
```
beckerhe, line 35: This could be simplified by using `return absl::StrJoin(modules, "\n", [](const auto& module){ return module.module_name; });`.

OrbitGl/PresetsDataView.cpp
```diff
@@ -18,15 +18,57 @@
...
+  switch (load_state) {
+    case kLoadable:
+      return "Yes";
+    case kPartiallyLoadable:
+      return "Partially";
+    case kNotLoadable:
+      return "No";
+    default:
+      return "";
+  }
```
beckerhe, line 61: That's basically a `to_string` function for the enum defined in App.h. Ideally this is a free function. And there should be no `default`-case. That prevents the compiler from warning about missing enum values. A `return` after the switch is fine though.

OrbitGl/PresetsDataView.cpp
```diff
@@ -55,17 +101,20 @@ std::string PresetsDataView::GetToolTip(int row, int /*column*/) {
     return OrbitUtils::Compare(presets_[a]->Member, presets_[b]->Member, ascending); \
   }
 
+#define ORBIT_PRESET_FUNC_SORT(Func) \
+  [&](int a, int b) { return OrbitUtils::Compare(Func(presets_[a]), Func(presets_[b]), ascending); }
+
```
beckerhe, line 106: We should avoid adding new macros to the code base. I see no disadvantage in adding this code directly to `DoSort`.

beckerhe: EDIT: If there is a reason for having a macro that I don't see, we should at least `#undef` the macro after usage.

OrbitGl/PresetsDataView.cpp
```diff
@@ -153,13 +199,45 @@ void PresetsDataView::DoFilter() {
...
   indices_.resize(presets_.size());
+  modules_.resize(presets_.size());
   for (size_t i = 0; i < presets_.size(); ++i) {
     indices_[i] = i;
+    std::vector<ModuleView> modules;
+    for (const auto& pair : presets_[i]->preset_info().path_to_module()) {
+      modules.push_back(
+          ModuleView(Path::GetFileName(pair.first), pair.second.function_hashes_size()));
+    }
+    modules_[i] = modules;
```
beckerhe, line 210: std::move

beckerhe: EDIT: You could also remove the temporary variable if you want. But I'm also fine with adding move.

OrbitGl/PresetsDataView.cpp
```diff
@@ -153,13 +199,45 @@ void PresetsDataView::DoFilter() {
...
   }
 
   DataView::OnDataChanged();
 }
 
+bool PresetsDataView::GetDisplayColor(int row, int /*column*/, unsigned char& red,
```
beckerhe, line 216: What about returning a `std::optional<Color>`? If I'm not mistaken we have a Color type.

OrbitGl/PresetsDataView.cpp
```diff
@@ -153,13 +199,45 @@ void PresetsDataView::DoFilter() {
...
+      green = 240;
+      blue = 200;
+      return true;
+    case kNotLoadable:
+      red = 230;
+      green = 190;
+      blue = 190;
+      return true;
+  }
+  return false;
```
beckerhe, line 238: this should be a free function just like the enum to_string function mentioned before.

OrbitGl/PresetsDataView.cpp
beckerhe: Ahhh. I just saw, this is an override. Never mind...

OrbitGl/App.cpp
```diff
@@ -880,22 +909,28 @@ void OrbitApp::LoadModulesFromPreset(const std::shared_ptr<Process>& process,
...
-    SendErrorToUi(
-        "Preset loading incomplete",
-        absl::StrFormat("Unable to load the preset for the following modules:\n\"%s\"\nThe "
-                        "modules are not loaded by process \"%s\".",
-                        absl::StrJoin(modules_not_found, "\"\n\""), process->GetName()));
+    if (static_cast<int>(modules_not_found.size()) == preset->preset_info().path_to_module_size()) {
+      // no modules were loaded
+      SendErrorToUi(
+          "Preset loading failed",
+          absl::StrFormat("The modules are not loaded by process \"%s\".", process->GetName()));
```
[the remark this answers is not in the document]
IrinaShkviro: Much clearer message, thanks, done.

OrbitGl/App.cpp
```diff
@@ -880,22 +909,28 @@ void OrbitApp::LoadModulesFromPreset(const std::shared_ptr<Process>& process,
...
+    } else {
+      SendWarningToUi(
+          "Preset loading incomplete",
```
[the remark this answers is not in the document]
IrinaShkviro: Done

beckerhe (commented):
Hey, I added some comments. I'm not through yet.

OrbitGl/PresetsDataView.cpp
IrinaShkviro: Done, thanks

OrbitGl/App.cpp
```diff
@@ -552,17 +551,40 @@ ErrorMessageOr<void> OrbitApp::OnLoadPreset(const std::string& filename) {
...
-  if (!SelectProcess(Path::GetFileName(process_full_path), preset)) {
-    SendErrorToUi("Preset loading failed",
-                  absl::StrFormat("The process \"%s\" is not running.", process_full_path));
-    return;
+}
+
+PresetLoadState OrbitApp::GetPresetLoadState(
+    const std::shared_ptr<orbit_client_protos::PresetFile>& preset) {
+  const std::shared_ptr<Process> selected_process = GetSelectedProcess();
+  if (selected_process == nullptr) {
```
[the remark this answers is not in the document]
florian-kuebler: We had that it the daily once. Our clang-format actually complains about that implicit conversion. So I guess we should stick with that.

OrbitGl/App.cpp
IrinaShkviro: I moved the logic to anonymous namespace, ptal

OrbitGl/App.cpp
```diff
@@ -552,17 +551,40 @@ ErrorMessageOr<void> OrbitApp::OnLoadPreset(const std::string& filename) {
...
+    return PresetLoadState::kNotLoadable;
+  }
+
+  uint32_t modules_not_found_count = 0;
+  uint32_t modules_found_count = 0;
```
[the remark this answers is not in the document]
IrinaShkviro: Done, thanks!

OrbitGl/App.h
IrinaShkviro: I moved it into the struct, ptal

OrbitGl/App.h
IrinaShkviro: Done, thanks

OrbitGl/PresetsDataView.cpp
IrinaShkviro: Moved this function to new PresetLoadState struct

OrbitGl/PresetsDataView.cpp
IrinaShkviro: Removed both macros, thanks

OrbitGl/PresetsDataView.cpp
```diff
@@ -18,15 +18,57 @@
...
+  }
+}
+
 const std::vector<DataView::Column>& PresetsDataView::GetColumns() {
   static const std::vector<Column> columns = [] {
     std::vector<Column> columns;
     columns.resize(kNumColumns);
-    columns[kColumnSessionName] = {"Preset", .49f, SortingOrder::kAscending};
-    columns[kColumnProcessName] = {"Process", .49f, SortingOrder::kAscending};
+    columns[kColumnLoadState] = {"Loadable", .14f, SortingOrder::kAscending};
```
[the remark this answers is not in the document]
IrinaShkviro: I added consts for columns name and width

OrbitGl/PresetsDataView.cpp
IrinaShkviro: I agree, we should return Color, added a TODO for now

OrbitGl/PresetsDataView.cpp
IrinaShkviro: Move to the PresetLoadState struct, ptal

OrbitGl/PresetsDataView.cpp
IrinaShkviro: Thanks for the tip! Applied it.

OrbitGl/PresetsDataView.cpp
IrinaShkviro: Done, thanks a lot!

OrbitGl/PresetsDataView.cpp
IrinaShkviro: Added std::move

IrinaShkviro:
I addressed review comments, ptal

OrbitGl/App.cpp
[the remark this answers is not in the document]
beckerhe: I think Florian meant "clang-tidy".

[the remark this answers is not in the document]
IrinaShkviro: I'll leave it as it because of clang format

OrbitGl/App.cpp
IrinaShkviro: Thanks for the suggestion! Created a bug for it http://b/167380800

beckerhe (approved):
LGTM

PR #1143, opened 2020-08-31
Fix: Show source file and line in Sampling Report

dimitry-:
Line number were not shown in sampling reports. This change propagates
information to SampledFunction, which fixes the problem.

Bug: http://b/154580143
Test: Start Orbit load module and see that line number and source file
      get updated in Sampling Report tab.

### DIFF
--- a/OrbitCore/SamplingProfiler.cpp
+++ b/OrbitCore/SamplingProfiler.cpp
@@ -293,6 +293,12 @@ void SamplingProfiler::FillThreadSampleDataSampleReports(const CaptureData& capt
       function.address = absolute_address;
       function.module_path = capture_data.GetModulePathByAddress(absolute_address);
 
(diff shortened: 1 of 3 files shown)


OrbitCore/SamplingProfiler.cpp
```diff
@@ -11,6 +11,7 @@
 using orbit_client_protos::CallstackEvent;
 using orbit_client_protos::FunctionInfo;
 using orbit_client_protos::LinuxAddressInfo;
+using orbit_grpc_protos::LineInfo;
```
IrinaShkviro, line 14: This line seems unused
```suggestion

```

IrinaShkviro (approved):
Looks good, thanks! Just one small comment

OrbitCore/SamplingProfiler.cpp
dimitry-: done.

PR #1144, opened 2020-08-31
Send the info for the selected tracepoints to the server for the even…

MariaSandru27:
…t collector

### DIFF
--- a/OrbitCaptureClient/CaptureClient.cpp
+++ b/OrbitCaptureClient/CaptureClient.cpp
@@ -17,6 +17,7 @@ using orbit_client_protos::FunctionInfo;
 using orbit_grpc_protos::CaptureOptions;
 using orbit_grpc_protos::CaptureRequest;
 using orbit_grpc_protos::CaptureResponse;
+using orbit_grpc_protos::TracepointInfo;
 
 static CaptureOptions::InstrumentedFunction::FunctionType IntrumentedFunctionTypeFromOrbitType(
     FunctionInfo::OrbitType orbit_type) {
@@ -33,7 +34,9 @@ static CaptureOptions::InstrumentedFunction::FunctionType IntrumentedFunctionTyp
 ErrorMessageOr<void> CaptureClient::StartCapture(
     ThreadPool* thread_pool, int32_t process_id, std::string process_name,
     std::shared_ptr<Process> process,
-    absl::flat_hash_map<uint64_t, FunctionInfo> selected_functions) {
+    absl::flat_hash_map<uint64_t, FunctionInfo> selected_functions,
+    absl::flat_hash_set<TracepointInfo, HashTracepointInfo, EqualTracepointInfo>
+        selected_tracepoints) {
   absl::MutexLock lock(&state_mutex_);
   if (state_ != State::kStopped) {
     return ErrorMessage(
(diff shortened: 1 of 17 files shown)


OrbitCore/CaptureData.h
```diff
@@ -148,6 +153,9 @@ class CaptureData {
   std::string process_name_;
   std::shared_ptr<Process> process_;
   absl::flat_hash_map<uint64_t, orbit_client_protos::FunctionInfo> selected_functions_;
+
+  absl::flat_hash_set<orbit_grpc_protos::TracepointInfo, HashTracepointInfo, EqualTracepointInfo>
+      selected_tracepoints_;
```
danielfenner, line 158: Can you add something like
`using TracepointInfoSet = absl::flat_hash_set<orbit_grpc_protos::TracepointInfo, HashTracepointInfo, EqualTracepointInfo>;`
to TracepointCustom.h?

Maybe it makes sense to make the HashTracepointInfo, EqualTracepointInfo internal again and just expose the typedef (if that is all you need from the outside).

OrbitCore/CaptureData.h
```diff
@@ -8,6 +8,7 @@
 #include <memory>
 #include <vector>
 
+#include "../../OrbitGl/TracepointCustom.h"
```
danielfenner, line 11: I left a similar comment below: OrbitCore must not depend on OrbitGl. Consider moving TracepointCustom.h to a shared place.

OrbitLinuxTracing/TracerThread.h
```diff
@@ -16,6 +17,7 @@
 #include <regex>
 #include <vector>
 
+#include "../../OrbitGl/TracepointCustom.h"
```
danielfenner, line 20: This looks very wrong. We do not want a dependency from the linux specific tracing code into the ui code in OrbitGl (this is why you have to specify the path in this weird way, the includes are not (really) visible). 
Can you move TracepointCustom.h to a place that is shared between the service and the UI? OrbitCore maybe?

MariaSandru27: Of course. Thanks for the advice!

OrbitCore/CaptureData.h
MariaSandru27: Yes, this is right. I will move it to make the includes more visible.

OrbitCore/CaptureData.h
MariaSandru27: I placed the implementation for the HashTracepointInfo and EqualTracepointInfo in a internal namespace and created a typedef in TracepointCustom.h for "absl::flat_hash_set<orbit_grpc_protos::TracepointInfo, HashTracepointInfo, EqualTracepointInfo>". Thanks so much for the advice!

OrbitCaptureClient/CaptureClient.cpp
```diff
@@ -84,6 +88,13 @@ void CaptureClient::Capture(int32_t process_id, std::string process_name,
     instrumented_function->set_function_type(IntrumentedFunctionTypeFromOrbitType(function.type()));
   }
 
+  for (const auto& tracepoint : selected_tracepoints) {
+    CaptureOptions::InstrumentedTracepoint* instrumented_tracepoint =
+        capture_options->add_instrumented_tracepoint();
```
dimitry-, line 93: Can you reuse TracepontInfo proto here? it looks pretty much the same.

OrbitLinuxTracing/TracerThread.h
```diff
@@ -16,6 +17,7 @@
 #include <regex>
 #include <vector>
 
+#include "../OrbitCore/TracepointCustom.h"
```
dimitry-, line 20: just TracepointCustom.h

OrbitCore/TracepointCustom.h
```diff
@@ -19,5 +22,10 @@ struct EqualTracepointInfo {
     return left.category().compare(right.category()) == 0 && left.name().compare(right.name()) == 0;
   }
 };
+}  // namespace internal
+
+typedef absl::flat_hash_set<orbit_grpc_protos::TracepointInfo, internal::HashTracepointInfo,
+                            internal::EqualTracepointInfo>
+    TracepointInfoSet;
```
dimitry-, line 29: nit: `using TacepointInfoSet=absl::flat_hash_set ...` instead of typedef

OrbitLinuxTracing/TracerThread.h
MariaSandru27: Maybe I am missing something but this file is in OrbitLinuxTracing and TracepointCustom.h is in OrbitCore. By only letting "TracepointCustom.h" the file won't be recognised.

OrbitCore/TracepointCustom.h
MariaSandru27: Sure. Thank you!

OrbitCaptureClient/CaptureClient.cpp
MariaSandru27: I was actually curious how I could use TracepointInfo. Thank you so much for the quick chat in which you have explained me!

OrbitLinuxTracing/TracerThread.h
danielfenner: You need to add OrbitCore to the dependencies of OrbitLinuxTracing in the CMakeLists.txt. The include dir in OrbitCore is then added to the appropriate search path during compilation.  

This is a way to enfore declaring dependencies in the build system.

MariaSandru27: I looked closer into the code and I do not need anything from TracepointCustom after all. At least, I know what to do in the future to get rid of relative paths. Thank you so much for the explanations!

dimitry- (approved):

danielfenner (approved):
LGTM!

PR #1145, opened 2020-09-01
Add TimerInfosIterator for Usage in Serializer

florian-kuebler:
Add an interator for the TimerChains and use it in Serializer.
This is the first step of removing dependencies of OrbitGl from
CaptureSerializer, such that it can be used also by other clients.

Test: Compile, Save/Load capture.

### DIFF
--- a/OrbitGl/CaptureSerializer.h
+++ b/OrbitGl/CaptureSerializer.h
@@ -38,8 +38,7 @@ class CaptureSerializer {
 
  private:
   template <class TimersIterator>
-  static void Save(std::ostream& stream, const orbit_client_protos::CaptureHeader& header,
-                   const CaptureData& capture_data,
+  static void Save(std::ostream& stream, const CaptureData& capture_data,
                    const absl::flat_hash_map<uint64_t, std::string>& key_to_string_map,
                    TimersIterator timers_iterator_begin, TimersIterator timers_iterator_end);
 
@@ -56,13 +55,14 @@ class CaptureSerializer {
 };
 
 template <class TimersIterator>
-void CaptureSerializer::Save(std::ostream& stream, const orbit_client_protos::CaptureHeader& header,
-                             const CaptureData& capture_data,
+void CaptureSerializer::Save(std::ostream& stream, const CaptureData& capture_data,
                              const absl::flat_hash_map<uint64_t, std::string>& key_to_string_map,
                              TimersIterator begin, TimersIterator end) {
   google::protobuf::io::OstreamOutputStream out_stream(&stream);
   google::protobuf::io::CodedOutputStream coded_output(&out_stream);
 
+  orbit_client_protos::CaptureHeader header;
+  header.set_version(kRequiredCaptureVersion);
   WriteMessage(&header, &coded_output);
 
   orbit_client_protos::CaptureInfo capture_info =
@@ -80,9 +80,6 @@ ErrorMessageOr<void> CaptureSerializer::Save(
     const std::string& filename, const CaptureData& capture_data,
     const absl::flat_hash_map<uint64_t, std::string>& key_to_string_map,
     TimersIterator timers_iterator_begin, TimersIterator timers_iterator_end) {
-  orbit_client_protos::CaptureHeader header;
-  header.set_version(kRequiredCaptureVersion);
-
   std::ofstream file(filename, std::ios::binary);
   if (file.fail()) {
     ERROR("Saving capture in \"%s\": %s", filename, "file.fail()");
@@ -91,8 +88,7 @@ ErrorMessageOr<void> CaptureSerializer::Save(
 
   {
     SCOPE_TIMER_LOG(absl::StrFormat("Saving capture in \"%s\"", filename));
-
-    Save(file, header, capture_data, key_to_string_map, std::move(timers_iterator_begin),
+    Save(file, capture_data, key_to_string_map, std::move(timers_iterator_begin),
          std::move(timers_iterator_end));
   }
 
--- a/OrbitGl/CaptureSerializer.h
+++ b/OrbitGl/CaptureSerializer.h
@@ -57,7 +57,8 @@ class CaptureSerializer {
 template <class TimersIterator>
 void CaptureSerializer::Save(std::ostream& stream, const CaptureData& capture_data,
                              const absl::flat_hash_map<uint64_t, std::string>& key_to_string_map,
-                             TimersIterator begin, TimersIterator end) {
+                             TimersIterator timers_iterator_begin,
+                             TimersIterator timers_iterator_end) {
   google::protobuf::io::OstreamOutputStream out_stream(&stream);
   google::protobuf::io::CodedOutputStream coded_output(&out_stream);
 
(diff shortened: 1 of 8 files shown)


OrbitGl/CaptureSerializer.h
```diff
@@ -41,4 +49,50 @@ class CaptureSerializer {
...
+
+  CHECK(time_graph_ != nullptr);
+
+  WriteMessage(&header, &coded_output);
+
+  orbit_client_protos::CaptureInfo capture_info = GenerateCaptureInfo(capture_data);
+  WriteMessage(&capture_info, &coded_output);
+
+  // Timers
+  int timers_count = time_graph_->GetNumTimers();
```
kisenshi, line 66: Maybe _timers_count_ could be provided as an argument to the method as well to remove also the call to `time_graph_->GetNumTimers();`

OrbitGl/TimerInfosIterator.h
```diff
@@ -0,0 +1,52 @@
...
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#ifndef ORBITGL_TIMER_INFOS_ITERATOR_H_
+#define ORBITGL_TIMER_INFOS_ITERATOR_H_
+
+#include <memory>
+#include <vector>
+
+#include "TimerChain.h"
```
kisenshi, line 11: I am thinking that having _TimerInfosIterator_ dependant on _TimerChain_ (which depends on _TextBox_) will make not possible to move this class outside OrbitGl within the serialisation, so the _Write_ method could not be moved to be independent from OrbitGl. Or am I missing something? Are _TimerChain_ and _TextBox_ meant to be moved outside _OrbitGl_ as well? Would that be possible?

kisenshi: Maybe _TimerInfosIterator_ should contain _TimerInfo_ instead of _TimerChain_ as checking the old code of where timers are being written, I see that _TimerInfo_ it is the type of data that ultimately is being written: the _TimerChains_ are retrieved, then from these, a _TimerBlock_ , that is used to get and write the _TimerInfo_

kisenshi: So if we get those _TimerInfo_ beforehand and pass it to the write method, the serialiser would receive all the data it needs and can be completely independent from _TimerChain_ (and therefore _TextBox_ and _OrbitGl_)

kisenshi: For clarity, the piece of code I mean is:

```
// Timers
int writes_count = 0;
  std::vector<std::shared_ptr<TimerChain>> chains = time_graph_->GetAllTimerChains();
  for (auto& chain : chains) {
    if (!chain) continue;
    for (TimerChainIterator it = chain->begin(); it != chain->end(); ++it) {
      TimerBlock& block = *it;
      for (uint32_t k = 0; k < block.size(); ++k) {
        WriteMessage(&block[k].GetTimerInfo(), &coded_output);
        if (++writes_count > timers_count) {
          return;
        }
      }
    }
  }
```
So the important bit I refer to is `WriteMessage(&block[k].GetTimerInfo(), &coded_output);`,

kisenshi (commented):
Thanks for this! I added some comments that hope make sense; please let me know if I have misunderstood or it is not clear what I wrote; will be happy to discuss :)

OrbitGl/TimerInfosIterator.h
florian-kuebler: The idea would be that this TimerInfosIterator is highly dependent from OrbitGl. On the Serializer it is then only a generic template. You can call "Write(..., vector.begin(), vector.end())" in your case and would be fine.

florian-kuebler: So that Iterator basically should just define how to iterate of the current state inside orbitgl's timer storage. But serializer remains independent.

OrbitGl/CaptureSerializer.h
florian-kuebler: Thanks, I'll change that.

OrbitGl/TimerInfosIterator.h
kisenshi: Ahhh, ok! Got it, makes sense, thanks!

florian-kuebler: Welcome.

OrbitGl/CaptureSerializer.h
IrinaShkviro: Do we need `timers_count` at all?

florian-kuebler: I don't think so. But haven't done any experiments yet. If you like, I can also just remove it, as it would be a design flaw anyways (if there would be more timers then numtimers).

OrbitGl/CaptureSerializer.h
```diff
@@ -14,31 +15,88 @@
...
+  }
+}
+
+template <class TimersIterator>
+ErrorMessageOr<void> CaptureSerializer::Save(
+    const std::string& filename, const CaptureData& capture_data,
+    const absl::flat_hash_map<uint64_t, std::string>& key_to_string_map,
+    TimersIterator timers_iterator_begin, TimersIterator timers_iterator_end) {
+  orbit_client_protos::CaptureHeader header;
+  header.set_version(kRequiredCaptureVersion);
```
IrinaShkviro, line 84: Could you move it to another `Save` method and drop `const orbit_client_protos::CaptureHeader& header` from it?

OrbitGl/CaptureSerializer.h
```diff
@@ -14,31 +15,88 @@
...
+
+  std::ofstream file(filename, std::ios::binary);
+  if (file.fail()) {
+    ERROR("Saving capture in \"%s\": %s", filename, "file.fail()");
+    return ErrorMessage("Error opening the file for writing");
+  }
+
+  {
+    SCOPE_TIMER_LOG(absl::StrFormat("Saving capture in \"%s\"", filename));
+
```
IrinaShkviro, line 94: I would prefer to remove the blank line as it doesn't help readability (https://google.github.io/styleguide/cppguide.html#Vertical_Whitespace)
```suggestion
```

OrbitGl/CaptureSerializer.h
florian-kuebler: Done. Thanks!

OrbitGl/CaptureSerializer.h
florian-kuebler: Right. Thanks!

IrinaShkviro (approved):
LGTM, thanks a lot!

OrbitGl/CaptureSerializer.h
```diff
@@ -14,31 +15,84 @@
...
+  orbit_client_protos::CaptureHeader header_;
 
-  const std::string kRequiredCaptureVersion = "1.52";
+  static inline const std::string kRequiredCaptureVersion = "1.52";
 };
 
+template <class TimersIterator>
+void CaptureSerializer::Save(std::ostream& stream, const CaptureData& capture_data,
+                             const absl::flat_hash_map<uint64_t, std::string>& key_to_string_map,
+                             TimersIterator begin, TimersIterator end) {
```
kisenshi, line 60: Maybe rename this to _timer_info_begin_ and _timer_info_end_ or _timers_begin_ or sth like that for more clarity when reading the code below

kisenshi: Or _timers_iterator_begin_ and  _timers_iterator_end_ like in the method below

florian-kuebler: Thanks. I made naming consistent.

kisenshi: Thanks!

kisenshi:
Looks good, thanks a lot for this!

kisenshi (approved):

PR #1147, opened 2020-09-01
Change the text of the top-down view's search field back to "search"

dpallotti:
No filtering is actually done, i.e., no node is hidden: the matching nodes are
highlighted and all their ancestors expanded. I think "search" is more accurate
for this.


dpallotti:
What do you think?

reichlfl (approved):
Right, sorry for the change - I tried to keep it consistent across all fields and misinterpreted the way it works for the Top-Down view.

PR #1151, opened 2020-09-01
Listen on local interfaces only in SSH tunnel

beckerhe:
By default QTcpServer listens on all available interfaces. We should
only listen on the local interface avoiding any security complications
and also the firewall dialog popping up.

Tests: Manual test of Orbit


dimitry- (approved):
do you want to cherry-pick this to 1.52?

PR #1152, opened 2020-09-01
Implement CaptureListener methods to collect capture data

kisenshi:
Implemented CaptureListener methods in ClientGgp so the data related to the capture is collected.

The data that is necessary to collect for the capture is: CaptureData, a vector containing the timers (TimerInfo) and a string manager.

OnCaptureComplete is also updated to set SamplingProfiler data when the capture has finished.

### DIFF
--- a/OrbitClientGgp/ClientGgp.cpp
+++ b/OrbitClientGgp/ClientGgp.cpp
@@ -16,11 +16,16 @@
 #include "OrbitBase/Result.h"
 #include "OrbitCaptureClient/CaptureClient.h"
 #include "OrbitClientServices/ProcessManager.h"
+#include "SamplingProfiler.h"
+#include "StringManager.h"
 #include "SymbolHelper.h"
 #include "absl/container/flat_hash_set.h"
 #include "capture_data.pb.h"
 
+using orbit_client_protos::CallstackEvent;
 using orbit_client_protos::FunctionInfo;
+using orbit_client_protos::LinuxAddressInfo;
+using orbit_client_protos::TimerInfo;
 using orbit_grpc_protos::ModuleInfo;
 using orbit_grpc_protos::ProcessInfo;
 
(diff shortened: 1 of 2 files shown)


OrbitClientGgp/ClientGgp.cpp
```diff
@@ -234,16 +242,39 @@ void ClientGgp::OnCaptureStarted(
...
+  SamplingProfiler sampling_profiler(*capture_data_.GetCallstackData(), capture_data_);
+  capture_data_.set_sampling_profiler(sampling_profiler);
+}
 
-void ClientGgp::OnTimer(const orbit_client_protos::TimerInfo&) {}
+void ClientGgp::OnTimer(const orbit_client_protos::TimerInfo& timer_info) {
+  if (timer_info.function_address() > 0) {
+    FunctionInfo* func =
+        capture_data_.process()->GetFunctionFromAddress(timer_info.function_address());
+    CHECK(func != nullptr);
```
danielfenner, line 255: Is the Check-fail intended here? I briefly looked into GetFunctionFromAddress and it seems returning nullptr is not that far fetched. 
If all these cases are "impossible" (Is this the case? We only can get samples with symbols from our instrumentation?) add a brief comment in front of the CHECK.

florian-kuebler: For function timers, the function *must* be present in the process. If this is not the case, our data model has changed, thus failing is good here.

IrinaShkviro (approved):
lgtm, thanks!

OrbitClientGgp/ClientGgp.cpp
kisenshi: I'm gonna add a brief comment before the check to clarify it, thanks!

kisenshi: Done!

florian-kuebler: Thanks!

danielfenner (approved):
LGTM!

kisenshi:
Merging this, thanks!

PR #1153, opened 2020-09-02
[CHERRYPICK] Listen on local interfaces only in SSH tunnel

beckerhe:
Dimitry suggested to cherry pick that bug fix. Please push back if you think that it's too late for proper testing.

By default QTcpServer listens on all available interfaces. We should
only listen on the local interface avoiding any security complications
and also the firewall dialog popping up.

Tests: Manual test of Orbit

### DIFF
--- a/OrbitSshQt/Tunnel.cpp
+++ b/OrbitSshQt/Tunnel.cpp
@@ -4,6 +4,7 @@
 
 #include "OrbitSshQt/Tunnel.h"
 
+#include <QHostAddress>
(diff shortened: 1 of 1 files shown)


florian-kuebler (approved):
LGTM. Thanks.

dimitry- (approved):

beckerhe:
I  will merge this since #1156 is targeting 1.52 as well, so more testing is needed anyway.

PR #1154, opened 2020-09-02
 Add "Load Symbols" context menu action to the top-down view

dpallotti:
#### Add "Load Symbols" context menu action to top-down view
Bug: http://b/162707037

N.B.: Loading a module's symbols, regardless of whether it's done from the
modules table, the sampling tab, etc., causes the top-down view to be refreshed
(the same happens with the sampling reports, to take the new information into
account, e.g., to group addresses under the same function) and hence to be
collapsed (unless a search string is in place). This is not caused by this
specific change, but it makes it more evident.

### DIFF
--- a/OrbitQt/TopDownViewItemModel.cpp
+++ b/OrbitQt/TopDownViewItemModel.cpp
@@ -65,6 +65,7 @@ QVariant TopDownViewItemModel::GetEditRoleData(const QModelIndex& index) const {
   if (thread_item != nullptr) {
     switch (index.column()) {
       case kThreadOrFunction:
+        // Threads are sorted by tid, not by name.
         return thread_item->thread_id();
       case kInclusive:
         return static_cast<qulonglong>(thread_item->sample_count());
@@ -90,6 +91,18 @@ QVariant TopDownViewItemModel::GetEditRoleData(const QModelIndex& index) const {
   return QVariant();
 }
 
+QVariant TopDownViewItemModel::GetToolTipRoleData(const QModelIndex& index) const {
+  auto* item = static_cast<TopDownNode*>(index.internalPointer());
+  auto function_item = dynamic_cast<TopDownFunction*>(item);
+  if (function_item != nullptr) {
+    switch (index.column()) {
+      case kModule:
+        return QString::fromStdString(function_item->module_path());
+    }
+  }
+  return QVariant();
+}
+
 QVariant TopDownViewItemModel::data(const QModelIndex& index, int role) const {
   if (!index.isValid()) {
     return QVariant();
@@ -100,6 +113,9 @@ QVariant TopDownViewItemModel::data(const QModelIndex& index, int role) const {
     // The value returned when role == Qt::EditRole is used for sorting.
     case Qt::EditRole:
       return GetEditRoleData(index);
+    // When role == Qt::ToolTipRole more detailed information than it's shown is returned.
+    case Qt::ToolTipRole:
+      return GetToolTipRoleData(index);
   }
   return QVariant();
 }
--- a/OrbitQt/TopDownViewItemModel.cpp
+++ b/OrbitQt/TopDownViewItemModel.cpp
@@ -9,6 +9,7 @@ TopDownViewItemModel::TopDownViewItemModel(std::unique_ptr<TopDownView> top_down
     : QAbstractItemModel{parent}, top_down_view_{std::move(top_down_view)} {}
 
 QVariant TopDownViewItemModel::GetDisplayRoleData(const QModelIndex& index) const {
+  CHECK(index.isValid());
   auto* item = static_cast<TopDownNode*>(index.internalPointer());
   auto thread_item = dynamic_cast<TopDownThread*>(item);
   auto function_item = dynamic_cast<TopDownFunction*>(item);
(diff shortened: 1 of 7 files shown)


OrbitQt/TopDownViewItemModel.cpp
```diff
@@ -15,7 +15,7 @@ QVariant TopDownViewItemModel::GetDisplayRoleData(const QModelIndex& index) cons
   if (thread_item != nullptr) {
     switch (index.column()) {
       case kThreadOrFunction:
-        if (thread_item->thread_id() == 0) {
+        if (thread_item->thread_id() == SamplingProfiler::kAllThreadsFakeTid) {
```
florian-kuebler, line 18: Thanks for adding this!

OrbitQt/TopDownViewItemModel.h
```diff
@@ -32,13 +32,15 @@ class TopDownViewItemModel : public QAbstractItemModel {
...
     kOfParent,
+    kModule,
     kFunctionAddress,
     kColumnCount
   };
 
  private:
   QVariant GetDisplayRoleData(const QModelIndex& index) const;
   QVariant GetEditRoleData(const QModelIndex& index) const;
+  QVariant GetToolTipRoleData(const QModelIndex& index) const;
```
beckerhe, line 43: Probably these should be nodiscard?

OrbitQt/TopDownViewItemModel.cpp
```diff
@@ -79,13 +82,27 @@ QVariant TopDownViewItemModel::GetEditRoleData(const QModelIndex& index) const {
...
+        return QString::fromStdString(function_item->GetModuleName());
       case kFunctionAddress:
         return static_cast<qulonglong>(function_item->function_absolute_address());
     }
   }
   return QVariant();
 }
 
+QVariant TopDownViewItemModel::GetToolTipRoleData(const QModelIndex& index) const {
+  auto* item = static_cast<TopDownNode*>(index.internalPointer());
```
beckerhe, line 95: Nit: maybe a `CHECK(index.isValid())`? Makes it easier to understand the function when it's seen on its own.

OrbitQt/topdownwidget.cpp
```diff
@@ -65,6 +69,32 @@ static void CollapseChildrenRecursively(QTreeView* tree_view, const QModelIndex&
...
+  std::set<std::string> unique_module_paths;
+  for (const auto& index : indices) {
+    std::string module_path =
+        index.model()
+            ->index(index.row(), TopDownViewItemModel::kModule, index.parent())
+            .data(Qt::ToolTipRole)
+            .toString()
+            .toStdString();
+    unique_module_paths.insert(module_path);
+  }
```
beckerhe, line 86: Have you considered storing the `std::shared_ptr<Module>` directly inside of `TopDownFunction` and adding a custom DataRole to the view such that you can directly access the modules from QModelIndexes?

I'm basically asking because adding a dependency to OrbitApp in TopDownWidget feels a bit wrong. Ideally the widget should get all its data from the model (TopDownView)

beckerhe: Also using the ToolTipRole here seems to be brittle. Whenever someone wants to adjust the tooltip in the future this logic might be broken.

florian-kuebler (approved):
Thanks for that change. Looks good to me.

OrbitQt/topdownwidget.cpp
dpallotti: > Have you considered storing the std::shared_ptr<Module> directly inside of TopDownFunction and adding a custom DataRole to the view such that you can directly access the modules from QModelIndexes?
I'm basically asking because adding a dependency to OrbitApp in TopDownWidget feels a bit wrong. Ideally the widget should get all its data from the model (TopDownView)

Personally I would prefer the view to have only a key to the data, not to store a pointer to the data itself, but maybe it's just me. But to answer them main point of your concern, I need `TopDownWidget` to depend on `OrbitApp` anyway, because that's the bridge to being able to perform actions that affect the rest of Orbit (in this case, loading modules).

beckerhe: Ah I see.

OrbitQt/TopDownViewItemModel.cpp
dpallotti: Good point, done.

OrbitQt/TopDownViewItemModel.h
dpallotti: Done.

OrbitQt/topdownwidget.cpp
dpallotti: If you have ideas regarding (not) depending on `OrbitApp` let's discuss them, I don't exactly like such dependency either.

dpallotti: In the meantime I added a custom data role to retrieve the module's path, please have a look.

dimitry- (approved):
regarding collapsing of the tree - is there a bug for it?

dpallotti:
> regarding collapsing of the tree - is there a bug for it?

I guess not, I'll create one.

OrbitQt/topdownwidget.cpp
beckerhe: That looks good to me. Thanks. I don't realy know what's the best way to solve this.

beckerhe (approved):

PR #1156, opened 2020-09-02
Fix Selecting "All Threads" and Display That Track on Linux

florian-kuebler:
With recent changes, we started using "-1" instead of "0" for special
purpose thread/process ids (no selection and all threads). By these
changes, we missed a couple of places.

This change uses SamplingProfiler::kAllThreadsFakeTid on all these
places and thus fixes the issue that scheduler track was not highlighting
anything, when clicking on the "all threads" track and also displays that
track on Linux at all.

Bug: http://167391339 http://167393367
Test: In capture, select different tracks (including all threads), and
      click on the background (to reset). Also check the lable in
      TopDownView and the SamplingViews.

### DIFF
--- a/OrbitCore/EventBuffer.cpp
+++ b/OrbitCore/EventBuffer.cpp
@@ -12,42 +12,42 @@ using orbit_client_protos::CallstackEvent;
 
 EventTracer GEventTracer;
 
-std::vector<CallstackEvent> EventBuffer::GetCallstackEvents(uint64_t a_TimeBegin,
-                                                            uint64_t a_TimeEnd,
-                                                            ThreadID a_ThreadId /*= 0*/) {
-  std::vector<CallstackEvent> callstackEvents;
-  for (auto& pair : m_CallstackEvents) {
-    ThreadID threadID = pair.first;
+std::vector<CallstackEvent> EventBuffer::GetCallstackEvents(
+    uint64_t time_begin, uint64_t time_end, ThreadID thread_id /*= kAllThreadsFakeTid*/) {
+  std::vector<CallstackEvent> callstack_events;
+  for (auto& pair : callstack_events_) {
+    const ThreadID callstack_thread_id = pair.first;
     std::map<uint64_t, CallstackEvent>& callstacks = pair.second;
 
-    if (a_ThreadId == 0 || threadID == a_ThreadId) {
-      for (auto it = callstacks.lower_bound(a_TimeBegin); it != callstacks.end(); ++it) {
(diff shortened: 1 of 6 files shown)


IrinaShkviro (approved):

florian-kuebler:
As this is targeting the release branch, I will wait for a second approval here.

OrbitCore/EventBuffer.cpp
```diff
@@ -12,42 +12,42 @@ using orbit_client_protos::CallstackEvent;
...
   CallstackEvent event;
   event.set_time(time);
   event.set_callstack_hash(cs_hash);
   event.set_thread_id(thread_id);
   event_map[time] = event;
 
-  // Add all callstack events to "thread 0".
-  std::map<uint64_t, CallstackEvent>& event_map_0 = m_CallstackEvents[0];
+  // Add all callstack events to "all threads".
+  std::map<uint64_t, CallstackEvent>& event_map_0 =
```
dpallotti, line 45: Maybe rename this variable, as the 0 was indeed referring to the "fake" thread id.

dpallotti (approved):
Thank you!

OrbitCore/EventBuffer.cpp
florian-kuebler: Done. Thanks!

OrbitCore/EventBuffer.cpp
```diff
@@ -12,42 +12,42 @@ using orbit_client_protos::CallstackEvent;
...
+std::vector<CallstackEvent> EventBuffer::GetCallstackEvents(
+    uint64_t time_begin, uint64_t time_end, ThreadID thread_id /*= kAllThreadsFakeTid*/) {
+  std::vector<CallstackEvent> callstack_events;
+  for (auto& pair : callstack_events_) {
+    const ThreadID callstack_thread_id = pair.first;
     std::map<uint64_t, CallstackEvent>& callstacks = pair.second;
 
-    if (a_ThreadId == 0 || threadID == a_ThreadId) {
-      for (auto it = callstacks.lower_bound(a_TimeBegin); it != callstacks.end(); ++it) {
+    if (thread_id == SamplingProfiler::kAllThreadsFakeTid || callstack_thread_id == thread_id) {
```
dimitry-, line 22: I think we should rename it to just kAllThreadsId, it is technically not real but this is implied anyways. (not for this PR)

dimitry- (approved):

OrbitCore/EventBuffer.cpp
florian-kuebler: I'll do so! Thanks!

PR #1158, opened 2020-09-02
Fix symbol uploading

beckerhe:
breakpad's upload tool needs to find `msdia140.dll` from the DIA SDK which comes with Visual Studio.

Apparently this does not work automatically when Visual Studio is installed in the BuildTools version.

This change adds the DIA SDK-path to $PATH to fix that problem.


IrinaShkviro (approved):
Thanks for the fix!

beckerhe:
> Thanks for the fix!

Thanks for spotting this!

PR #1052, opened 2020-08-17
Move GSelectedThreadID to DataManager

florian-kuebler:
Move the GSelectedThreadID global from Capture to DataManager.
Also rename some variables to match guidelines.

Test: Compile, take capture, select different cpu tracks.
Bug: http://b/163020637

PR #1159, opened 2020-09-02
[REBASE] Fix Selecting "All Threads" and Display That Track on Linux (#1156)

florian-kuebler:
With recent changes, we started using "-1" instead of "0" for special
purpose thread/process ids (no selection and all threads). By these
changes, we missed a couple of places.

This change uses SamplingProfiler::kAllThreadsFakeTid on all these
places and thus fixes the issue that scheduler track was not highlighting
anything, when clicking on the "all threads" track and also displays that
track on Linux at all.

Bug: http://167391339 http://167393367
Test: In capture, select different tracks (including all threads), and
      click on the background (to reset). Also check the lable in
      TopDownView and the SamplingViews.

### DIFF
--- a/OrbitCore/EventBuffer.cpp
+++ b/OrbitCore/EventBuffer.cpp
@@ -12,43 +12,43 @@ using orbit_client_protos::CallstackEvent;
 
 EventTracer GEventTracer;
 
-std::vector<CallstackEvent> EventBuffer::GetCallstackEvents(uint64_t a_TimeBegin,
-                                                            uint64_t a_TimeEnd,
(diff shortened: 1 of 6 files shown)


dimitry- (approved):

ronaldfw:
I somehow missed the change going from -1 to 0. Can you point me to the PR that changed this in the first place? What was the root cause that required this change?

florian-kuebler:
I guess that was the first place: https://github.com/google/orbit/pull/1052#discussion_r471465511

PR #1161, opened 2020-09-02
Mark empty preset as 'loadable'

IrinaShkviro:
Test: Start Orbit, save an empty preset, check that the font for this
empty preset is green, and vakue in Loadable column is 'Yes'.


florian-kuebler (approved):
LGTM. Thanks!

reichlfl (commented):
LGTM, thanks

PR #1163, opened 2020-09-02
Add "Go to Disassembly" context menu action to top-down view

dpallotti:
#### Add "Go to Disassembly" context menu action to top-down view
Bug: http://b/162707037
#### Add tooltips for function names to top-down view
Can be useful when the function name is particularly long and doesn't fit in the
column width.

### DIFF
--- a/OrbitQt/topdownwidget.cpp
+++ b/OrbitQt/topdownwidget.cpp
@@ -10,6 +10,8 @@
 #include "App.h"
 #include "TopDownViewItemModel.h"
 
+using orbit_client_protos::FunctionInfo;
+
 void TopDownWidget::SetTopDownView(std::unique_ptr<TopDownView> top_down_view) {
   CHECK(app_ != nullptr);
 
@@ -32,6 +34,7 @@ const QString TopDownWidget::kActionCollapseChildrenRecursively =
 const QString TopDownWidget::kActionExpandAll = QStringLiteral("Expand all");
 const QString TopDownWidget::kActionCollapseAll = QStringLiteral("Collapse all");
 const QString TopDownWidget::kActionLoadSymbols = QStringLiteral("&Load Symbols");
+const QString TopDownWidget::kActionDisassembly = QStringLiteral("Go to &Disassembly");
 
 static void ExpandRecursively(QTreeView* tree_view, const QModelIndex& index) {
   if (!index.isValid()) {
@@ -71,7 +74,7 @@ static void CollapseChildrenRecursively(QTreeView* tree_view, const QModelIndex&
 
 static std::vector<std::shared_ptr<Module>> GetModulesFromIndices(
     OrbitApp* app, const std::vector<QModelIndex>& indices) {
-  std::shared_ptr<Process> process = app->GetCaptureData().process();
+  const std::shared_ptr<Process>& process = app->GetCaptureData().process();
   CHECK(process != nullptr);
 
   std::set<std::string> unique_module_paths;
@@ -93,6 +96,26 @@ static std::vector<std::shared_ptr<Module>> GetModulesFromIndices(
     }
   }
   return modules;
+}
+
+static std::vector<FunctionInfo*> GetFunctionsFromIndices(OrbitApp* app,
+                                                          const std::vector<QModelIndex>& indices) {
+  const std::shared_ptr<Process>& process = app->GetCaptureData().process();
+  CHECK(process != nullptr);
+
+  absl::flat_hash_set<FunctionInfo*> functions_set;
+  for (const auto& index : indices) {
+    uint64_t absolute_address =
+        index.model()
+            ->index(index.row(), TopDownViewItemModel::kFunctionAddress, index.parent())
+            .data(Qt::EditRole)
+            .toLongLong();
+    FunctionInfo* function = process->GetFunctionFromAddress(absolute_address);
+    if (function != nullptr) {
+      functions_set.insert(function);
+    }
+  }
+  return std::vector<FunctionInfo*>(functions_set.begin(), functions_set.end());
 }
 
 void TopDownWidget::onCustomContextMenuRequested(const QPoint& point) {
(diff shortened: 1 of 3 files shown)


OrbitQt/topdownwidget.cpp
```diff
@@ -32,6 +34,7 @@ const QString TopDownWidget::kActionCollapseChildrenRecursively =
 const QString TopDownWidget::kActionExpandAll = QStringLiteral("Expand all");
 const QString TopDownWidget::kActionCollapseAll = QStringLiteral("Collapse all");
 const QString TopDownWidget::kActionLoadSymbols = QStringLiteral("&Load Symbols");
+const QString TopDownWidget::kActionDisassembly = QStringLiteral("Go to &Disassembly");
```
florian-kuebler, line 37: Why is the "&" there? Is this a special purpose char, or is this by accident.

OrbitQt/topdownwidget.cpp
```diff
@@ -95,6 +98,26 @@ static std::vector<std::shared_ptr<Module>> GetModulesFromIndices(
...
 }
 
+static std::vector<FunctionInfo*> GetFunctionsFromIndices(OrbitApp* app,
+                                                          const std::vector<QModelIndex>& indices) {
+  const std::shared_ptr<Process>& process = app->GetCaptureData().process();
+  CHECK(process != nullptr);
+
+  absl::flat_hash_set<FunctionInfo*> functions_set;
+  for (const auto& index : indices) {
+    uint64_t absolute_address =
```
florian-kuebler, line 108: nit: this can be const

OrbitQt/topdownwidget.cpp
```diff
@@ -95,6 +98,26 @@ static std::vector<std::shared_ptr<Module>> GetModulesFromIndices(
...
+        index.model()
+            ->index(index.row(), TopDownViewItemModel::kFunctionAddress, index.parent())
+            .data(Qt::EditRole)
+            .toLongLong();
+    FunctionInfo* function = process->GetFunctionFromAddress(absolute_address);
```
florian-kuebler, line 113: nit: can this be also const?

OrbitQt/topdownwidget.cpp
```diff
@@ -174,6 +203,10 @@ void TopDownWidget::onCustomContextMenuRequested(const QPoint& point) {
     ui_->topDownTreeView->collapseAll();
   } else if (action->text() == kActionLoadSymbols) {
     app_->LoadModules(app_->GetCaptureData().process(), modules_to_load);
+  } else if (action->text() == kActionDisassembly) {
+    for (FunctionInfo* function : functions) {
```
florian-kuebler, line 207: can be const

OrbitQt/topdownwidget.cpp
```diff
@@ -135,6 +158,9 @@ void TopDownWidget::onCustomContextMenuRequested(const QPoint& point) {
   }
   bool enable_load = !modules_to_load.empty();
 
+  std::vector<FunctionInfo*> functions = GetFunctionsFromIndices(app_, selected_tree_indices);
```
florian-kuebler, line 161: also here can be const.

florian-kuebler (approved):
Thanks for the change. Looks good to me. However, I would prefer marking read only variables as const, whenever possible.

OrbitQt/topdownwidget.cpp
dpallotti: It enables "D" as a shortcut, also see the other actions.

dpallotti (commented):
> I would prefer marking read only variables as const, whenever possible.

Is this a new policy that we recently introduced and I missed? If not, while I'm not against it (and I like it in languages that easily support it like Kotlin's `val`), it's just a preference and as far as I'm aware it's not common in C++.

dpallotti:
> > I would prefer marking read only variables as const, whenever possible.
> 
> Is this a new policy that we recently introduced and I missed? If not, while I'm not against it (and I like it in languages that easily support it like Kotlin's `val`), it's just a preference and as far as I'm aware it's not (that) common in C++.

Style guide: "Using const on local variables is neither encouraged nor discouraged."

OrbitQt/topdownwidget.cpp
florian-kuebler: Thanks for clarifying

florian-kuebler:
> > > I would prefer marking read only variables as const, whenever possible.
> > 
> > 
> > Is this a new policy that we recently introduced and I missed? If not, while I'm not against it (and I like it in languages that easily support it like Kotlin's `val`), it's just a preference and as far as I'm aware it's not (that) common in C++.
> 
> Style guide: "Using const on local variables is neither encouraged nor discouraged."

Right. For me it came up in some reviews. I think we never agreed on it (probably we should). So feel free to drop. Personally I like it (+100 for Kotlin's/Scala's `val` vs. `var`, thats just so clean!).

beckerhe:
> > > > I would prefer marking read only variables as const, whenever possible.
> > > 
> > > 
> > > Is this a new policy that we recently introduced and I missed? If not, while I'm not against it (and I like it in languages that easily support it like Kotlin's `val`), it's just a preference and as far as I'm aware it's not (that) common in C++.
> > 
> > 
> > Style guide: "Using const on local variables is neither encouraged nor discouraged."
> 
> Right. For me it came up in some reviews. I think we never agreed on it (probably we should). So feel free to drop. Personally I like it (+100 for Kotlin's/Scala's `val` vs. `var`, thats just so clean!).

Easy:
```
#define var auto
#define val auto const
```

:slightly_smiling_face:

florian-kuebler:
> > > > > I would prefer marking read only variables as const, whenever possible.
> > > > 
> > > > 
> > > > Is this a new policy that we recently introduced and I missed? If not, while I'm not against it (and I like it in languages that easily support it like Kotlin's `val`), it's just a preference and as far as I'm aware it's not (that) common in C++.
> > > 
> > > 
> > > Style guide: "Using const on local variables is neither encouraged nor discouraged."
> > 
> > 
> > Right. For me it came up in some reviews. I think we never agreed on it (probably we should). So feel free to drop. Personally I like it (+100 for Kotlin's/Scala's `val` vs. `var`, thats just so clean!).
> 
> Easy:
> 
> ```
> #define var auto
> #define val auto const
> ```
> 
> 🙂

I guess that's cheating :D. also `val & foo` may look a bit strange, right?

dpallotti:
> Right. For me it came up in some reviews. I think we never agreed on it (probably we should). So feel free to drop. Personally I like it (+100 for Kotlin's/Scala's `val` vs. `var`, thats just so clean!).

I guess we should discuss it. But until then, as valid as it is, I would stay consistent with the rest of the class and with the vast majority of the codebase.

PR #1164, opened 2020-09-02
Add fallback sorting to SamplingReportDataView

beckerhe:
SamplingReportDataView used to perform stable_sort. stable_sort's outcome
depends on the previous state of the container. Since recent changes
trigger a reconstruction of SamplingReportDataView stable_sort
preserving behaviour doesn't work anymore since the container's previous
state is not preserved. That lead's to reorderings observed by the user.

This can be avoided by establishing a total ordering. In this commit I
use the function address as a unique and totally ordered field.

Test: Manual UI test
Bug: http://b/167541131

### DIFF
--- a/OrbitGl/SamplingReportDataView.cpp
+++ b/OrbitGl/SamplingReportDataView.cpp
@@ -113,9 +113,29 @@ void SamplingReportDataView::DoSort() {
       break;
   }
 
-  if (sorter) {
-    std::stable_sort(indices_.begin(), indices_.end(), sorter);
+  if (!sorter) {
+    return;
   }
+
+  const auto fallback_sorter = [&](const auto& ind_left, const auto& ind_right) {
+    // `SampledFunction::address` is the absolute function address. Hence it is unique and qualifies
+    // for total ordering.
+    return functions[ind_left].address < functions[ind_right].address;
+  };
+
+  const auto combined_sorter = [&](const auto& ind_left, const auto& ind_right) {
+    if (sorter(ind_left, ind_right)) {
+      return true;
+    }
+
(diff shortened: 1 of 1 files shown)


florian-kuebler (approved):
LGTM thanks. Just one minor comment regarding the commit message:
I guess we are using "Bug: " instead of "Fixes: ".

OrbitGl/SamplingReportDataView.cpp
```diff
@@ -113,9 +113,29 @@ void SamplingReportDataView::DoSort() {
...
   }
 
-  if (sorter) {
-    std::stable_sort(indices_.begin(), indices_.end(), sorter);
+  if (!sorter) {
+    return;
   }
+
+  const auto fallback_sorter = [&](const auto& ind_left, const auto& ind_right) {
+    // address is the absolute function address. Hence it is unique and qualifies
```
florian-kuebler, line 121: Also start with capital letters in Comments.

beckerhe: Well, I used "address" because I meant the field name. I will add backticks to make that clear.

IrinaShkviro (approved):
Thanks for the fix!

OrbitGl/SamplingReportDataView.cpp
florian-kuebler: Thanks.

beckerhe:
> LGTM thanks. Just one minor comment regarding the commit message:
> I guess we are using "Bug: " instead of "Fixes: ".

Ah thanks. I mixed this up. Should be fixed now.

florian-kuebler:
Yes, looks good. Thanks!

PR #1166, opened 2020-09-02
Fix Broken Capture Duration for Empty Captures

florian-kuebler:
When opening the capture view for the first time showed "days"
instead of "0". This fixes this issue, by making EventBuffer's
callstacks_ read methods const.
Previousely there was an effective access in EventTrack GEventTracer.GetEventBuffer().GetCallstacks[tid];
which was executed always at the beginning for the all threads track
and added empty callstacks for that thread id. Additionally, the
check for the capture time span asks for EventBuffer::HasEvent, which
just checked whether the map is empty (which it was not).

This also made some minor local cleanup refactorings (nodiscard, naming).

Test: On Empty Capture: Go to Sampling View. Also clear capture.
Bug: http://b/162706447

### DIFF
--- a/OrbitCore/EventBuffer.h
+++ b/OrbitCore/EventBuffer.h
@@ -50,7 +50,15 @@ class EventBuffer {
 
   [[nodiscard]] bool HasEvent() {
     ScopeLock lock(mutex_);
-    return !callstack_events_.empty();
+    if (callstack_events_.empty()) {
+      return false;
+    }
+    for (const auto& pair : callstack_events_) {
+      if (!pair.second.empty()) {
+        return true;
+      }
+    }
+    return true;
   }
 
   [[nodiscard]] size_t GetNumEvents() const;
--- a/OrbitCore/EventBuffer.h
+++ b/OrbitCore/EventBuffer.h
@@ -15,13 +15,14 @@
 
 class EventBuffer {
  public:
-  EventBuffer() : min_time_(0), max_time_(LLONG_MAX) {}
+  EventBuffer() : max_time_(0), min_time_(std::numeric_limits<uint64_t>::max()) {}
 
   void Reset() {
     callstack_events_.clear();
-    max_time_ = LLONG_MAX;
-    min_time_ = 0;
+    min_time_ = std::numeric_limits<uint64_t>::max();
+    max_time_ = 0;
   }
+
   [[nodiscard]] const std::map<int32_t, std::map<uint64_t, orbit_client_protos::CallstackEvent>>&
   GetCallstacks() const {
     return callstack_events_;
@@ -36,12 +37,17 @@ class EventBuffer {
     }
     return it->second;
   }
+
   Mutex& GetMutex() { return mutex_; }
+
   [[nodiscard]] std::vector<orbit_client_protos::CallstackEvent> GetCallstackEvents(
       uint64_t time_begin, uint64_t time_end,
       int32_t thread_id = SamplingProfiler::kAllThreadsFakeTid) const;
-  [[nodiscard]] uint64_t GetMaxTime() const { return min_time_; }
-  [[nodiscard]] uint64_t GetMinTime() const { return max_time_; }
+
+  [[nodiscard]] uint64_t GetMaxTime() const { return max_time_; }
+
+  [[nodiscard]] uint64_t GetMinTime() const { return min_time_; }
+
   [[nodiscard]] bool HasEvent() {
     ScopeLock lock(mutex_);
     return !callstack_events_.empty();
@@ -50,8 +56,8 @@ class EventBuffer {
   [[nodiscard]] size_t GetNumEvents() const;
 
   void RegisterTime(uint64_t time) {
-    if (time > min_time_) min_time_ = time;
-    if (time > 0 && time < max_time_) max_time_ = time;
+    if (time > max_time_) max_time_ = time;
+    if (time > 0 && time < min_time_) min_time_ = time;
   }
 
   void AddCallstackEvent(uint64_t time, CallstackID cs_hash, ThreadID thread_id);
@@ -59,8 +65,8 @@ class EventBuffer {
  private:
   Mutex mutex_;
   std::map<int32_t, std::map<uint64_t, orbit_client_protos::CallstackEvent>> callstack_events_;
-  std::atomic<uint64_t> min_time_;
   std::atomic<uint64_t> max_time_;
+  std::atomic<uint64_t> min_time_;
 };
 
 #endif  // ORBIT_CORE_EVENT_BUFFER_H_
--- a/OrbitCore/EventBuffer.h
+++ b/OrbitCore/EventBuffer.h
@@ -15,43 +15,52 @@
 
 class EventBuffer {
  public:
-  EventBuffer() : m_MaxTime(0), m_MinTime(LLONG_MAX) {}
+  EventBuffer() : min_time_(0), max_time_(LLONG_MAX) {}
 
   void Reset() {
     callstack_events_.clear();
-    m_MinTime = LLONG_MAX;
-    m_MaxTime = 0;
+    max_time_ = LLONG_MAX;
+    min_time_ = 0;
   }
-  std::map<ThreadID, std::map<uint64_t, orbit_client_protos::CallstackEvent> >& GetCallstacks() {
+  [[nodiscard]] const std::map<int32_t, std::map<uint64_t, orbit_client_protos::CallstackEvent>>&
+  GetCallstacks() const {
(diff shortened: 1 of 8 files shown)


OrbitCore/EventBuffer.h
```diff
@@ -15,43 +15,52 @@
...
  public:
-  EventBuffer() : m_MaxTime(0), m_MinTime(LLONG_MAX) {}
+  EventBuffer() : min_time_(0), max_time_(LLONG_MAX) {}
 
   void Reset() {
     callstack_events_.clear();
-    m_MinTime = LLONG_MAX;
-    m_MaxTime = 0;
+    max_time_ = LLONG_MAX;
+    min_time_ = 0;
```
pierricgimmig, line 23: The order here was reversed, max_time_ should be intialized to 0 and min_time_ to LLONG_MAX.

pierricgimmig: We could also use std::numeric_limits<uint64_t>::max(); and std::numeric_limits<uint64_t>::min(); for consistency while we are at it.

OrbitCore/EventBuffer.h
```diff
@@ -15,43 +15,52 @@
...
-  size_t GetNumEvents() const;
-#endif
+  [[nodiscard]] size_t GetNumEvents() const;
 
-  void RegisterTime(uint64_t a_Time) {
-    if (a_Time > m_MaxTime) m_MaxTime = a_Time;
-    if (a_Time > 0 && a_Time < m_MinTime) m_MinTime = a_Time;
+  void RegisterTime(uint64_t time) {
+    if (time > min_time_) min_time_ = time;
+    if (time > 0 && time < max_time_) max_time_ = time;
```
pierricgimmig, line 54: It seems min and max were reversed here as well.

OrbitGl/TimeGraph.cpp
```diff
@@ -872,10 +872,10 @@ void TimeGraph::SortTracks() {
     m_EventCount.clear();
 
     for (auto& pair : GEventTracer.GetEventBuffer().GetCallstacks()) {
-      ThreadID threadID = pair.first;
-      std::map<uint64_t, CallstackEvent>& callstacks = pair.second;
-      m_EventCount[threadID] = callstacks.size();
-      GetOrCreateThreadTrack(threadID);
+      ThreadID thread_id = pair.first;
```
pierricgimmig, line 875: I saw that you replaced ThreadID by uint32_t earlier, do you want to do it here as well?

OrbitCore/EventBuffer.h
```diff
@@ -15,43 +15,52 @@
 
 class EventBuffer {
  public:
-  EventBuffer() : m_MaxTime(0), m_MinTime(LLONG_MAX) {}
+  EventBuffer() : min_time_(0), max_time_(LLONG_MAX) {}
```
pierricgimmig, line 18: Why was min and max reversed here?

OrbitGl/Batcher.h
```diff
@@ -18,7 +18,7 @@ using TooltipCallback = std::function<std::string(PickingId)>;
 struct PickingUserData {
   TextBox* text_box_;
   TooltipCallback generate_tooltip_;
-  void* custom_data_ = nullptr;
+  const void* custom_data_ = nullptr;
```
pierricgimmig, line 21: What is the purpose of adding const here?

OrbitCore/EventBuffer.h
pierricgimmig: min_time_ should be initialized to LLONG_MAX so that the first time that we add will become the new minimum. Inversely for max_time_.

florian-kuebler: Ouh, thanks for spotting this. Seems that I mixed them up while refactoring. Will change that back again.

OrbitGl/EventTrack.cpp
```diff
@@ -70,8 +70,8 @@ void EventTrack::UpdatePrimitives(uint64_t min_tick, uint64_t max_tick, PickingM
   const bool picking = picking_mode != PickingMode::kNone;
 
   ScopeLock lock(GEventTracer.GetEventBuffer().GetMutex());
-  std::map<uint64_t, CallstackEvent>& callstacks =
-      GEventTracer.GetEventBuffer().GetCallstacks()[thread_id_];
```
dpallotti, line 74: Here's one thing I don't understand: This is one of the two lines that caused the insertion. Still, this inserts an empty map of callstack events. Since it's empty and there are no actual events, shouldn't the code that computes the duration of the capture (note: I'm not sure where that is) produce zero anyway, as there are still effectively not callstack events? Somehow I have the feeling that while this is one cause, the real cause could be in that other logic.

OrbitCore/EventBuffer.h
```diff
@@ -15,43 +15,52 @@
...
+  GetCallstacksOfThread(int32_t thread_id) const {
+    static std::map<uint64_t, orbit_client_protos::CallstackEvent> empty;
+    const auto& it = callstack_events_.find(thread_id);
+    if (it == callstack_events_.end()) {
+      return empty;
+    }
+    return it->second;
+  }
+  Mutex& GetMutex() { return mutex_; }
+  [[nodiscard]] std::vector<orbit_client_protos::CallstackEvent> GetCallstackEvents(
```
dpallotti, line 40: Add empty lines between function definitions.

OrbitGl/EventTrack.cpp
florian-kuebler: It should now not insert an empty callstack anymore, but just return the static const empty callstack defined in the callee.

dpallotti: But what is the code that adds an empty callstack? This specific line only adds an empty map.

OrbitCore/EventBuffer.h
florian-kuebler: Done. thanks!

OrbitCore/EventBuffer.h
florian-kuebler: Thanks. Used std::numeric_limits<uint64_t>::max().

OrbitCore/EventBuffer.h
florian-kuebler: Done. Thanks.

florian-kuebler (commented):
I updated the PR. Also updated the HasEvent method to be correct.

OrbitCore/EventBuffer.h
```diff
@@ -15,43 +15,66 @@
...
+    ScopeLock lock(mutex_);
+    if (callstack_events_.empty()) {
+      return false;
+    }
+    for (const auto& pair : callstack_events_) {
+      if (!pair.second.empty()) {
+        return true;
+      }
+    }
+    return true;
```
dpallotti, line 61: Shouldn't this be `return false` as all the maps were found empty?

florian-kuebler: And that's why code reviews are so useful (and test would be even better.. :)) Thanks for spotting this!

florian-kuebler: just updated the branch again.

OrbitGl/EventTrack.cpp
dpallotti: Discussed separately, fixed `HasEvent`.

OrbitGl/EventTrack.cpp
```diff
@@ -208,7 +208,7 @@ std::string EventTrack::GetSampleTooltip(PickingId id) const {
   }
 
   const CallstackData* callstack_data = GOrbitApp->GetCaptureData().GetCallstackData();
-  CallstackEvent* callstack_event = static_cast<CallstackEvent*>(user_data->custom_data_);
+  const auto callstack_event = static_cast<const CallstackEvent*>(user_data->custom_data_);
```
dpallotti, line 211: I don't know if we have general guidance here, but I would use `const auto*` here, `auto` because of the `static_cast` but still making clear it's a pointer. I think it's done like this in many other places when casting pointers.

dpallotti (approved):

OrbitGl/EventTrack.cpp
florian-kuebler: yep, think this is better. thanks

OrbitGl/TimeGraph.cpp
florian-kuebler: Done.

florian-kuebler:
Thanks for the reviews so far. @pierricgimmig I updated the PR with all the changes you suggested.

pierricgimmig (approved):
Thanks for the changes Florian!

PR #1167, opened 2020-09-02
Send the collection of events to capture

MariaSandru27:


MariaSandru27:
Create a hash mechanism is order to not send the tracepoint's category and
name every time an event occurs if these do repet themselves.

The only data that will be sent from the service to the client will be
the one that differs for every single tracepoint event, like the pid,
tid, timestamp and cpu. To distinguish the category and name of the tracepoint, a hash key will be associated to these, such that these are sent only once.

Disable the reading of the specific data to every tracepoint in favour of reading only fields such as pid, tid, cpu and timestamp.

OrbitLinuxTracing/PerfEvent.h
```diff
@@ -369,6 +369,17 @@ class TaskRenamePerfEvent : public TracepointPerfEvent {
   }
 };
 
+class TracepointEvent : public TracepointPerfEvent {
```
dpallotti: The name of this class is too generic: what is the difference between `TracepointEvent` and `TracepointPerfEvent`? Actually, why do you even need this class? Couldn't you use `TracepointPerfEvent` directly and just add `GetPid` and `GetTid` to that?

OrbitCaptureClient/CaptureEventProcessorProcessEventsFuzzer.cpp
```diff
@@ -33,6 +33,8 @@ class MyCaptureListener : public CaptureListener {
   void OnCallstackEvent(CallstackEvent) override {}
   void OnThreadName(int32_t, std::string) override {}
   void OnAddressInfo(LinuxAddressInfo) override {}
+  void OnUniqueTracepoint(uint64_t, orbit_grpc_protos::TracepointInfo) override {}
+  virtual void OnTracepointEvent(orbit_client_protos::TracepointEvent) override {}
```
dpallotti, line 37: You don't need to specify `virtual` when using `override`.

OrbitGrpcProtos/capture.proto
```diff
@@ -78,6 +78,22 @@ message CallstackSample {
   uint64 timestamp_ns = 5;
 }
 
+message InternedTracepointInfo {
+  uint64 key = 1;
+  TracepointInfo intern = 2;
+}
+
+message TracepointSample {
```
dpallotti, line 86: What about calling this `TracepointEvent`? We tend to use "sample" when we refer to time-based callstack samples.

OrbitGrpcProtos/capture.proto
```diff
@@ -78,6 +78,22 @@ message CallstackSample {
...
+  int32 pid = 1;
+  int32 tid = 2;
+  int64 time = 3;
+  int32 cpu = 4;
+  oneof tracepoint_info_or_key {
+    TracepointInfo tracepoint_info = 5;
+    uint64 tracepoint_key = 6;
```
dpallotti, line 93: `tracepoint_info_key`

OrbitCore/TracepointInfoManager.h
```diff
@@ -0,0 +1,34 @@
+// Copyright (c) 2020 The Orbit Authors. All rights reserved.
```
dpallotti, line 1: We are trying to get rid of `OrbitCore` and to at least not add anything to it. In what module is this used? If only `OrbitGl` this can be in `OrbitGl`.

OrbitCore/TracepointInfoManager.h
```diff
@@ -0,0 +1,34 @@
...
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#ifndef ORBIT_TRACEPOINTINFOMANAGER_H
```
dpallotti, line 5: Be consistent with the include guards of other files. This would be `ORBIT_CORE_TRACEPOINT_INFO_MANAGER_H_`.

OrbitLinuxTracing/TracerThread.cpp
```diff
@@ -755,6 +763,17 @@ void TracerThread::ProcessSampleEvent(const perf_event_header& header,
             is_callchain_sample <=
         1);
 
+  absl::flat_hash_map<orbit_grpc_protos::TracepointInfo, bool, internal::HashTracepointInfo,
+                      internal::EqualTracepointInfo>
+      are_instrumented_tracepoints;
```
dpallotti, line 768: This name is weird. Why "are" in front?

dpallotti (commented):
I'll go through it again later but I'll add my first pass of comments now.

OrbitLinuxTracing/PerfEvent.h
MariaSandru27: Thanks for the comment! One of the differences is that TracepointEvent does not inherit from anywhere and the ring buffer has the type perf_event_sample_id_tid_time_streamid_cpu instead of perf_event_raw_sample_fixed since only the contents of that structure is read. Also, TracepointEvent does not  need to have the field 'std::unique_ptr<uint8_t[]> tracepoint_data' since the data particular of every tracepoint is not read. I created TracepointEvent which is an independent class of TracepointPerfEvent because it had fields not necessary at this moment and it reserved additional memory that was not really put to use (such as space for the header, size and data[char]). However, I get what you mean with the name as being too generic and, if you think it would be a good idea, I could use TracepointPerfEvent even though I do not read all its contents at this point.

OrbitGrpcProtos/capture.proto
MariaSandru27: Thanks for pointing that out! I will change it immediately.

OrbitCore/TracepointInfoManager.h
MariaSandru27: Yes, it is used only in OrbitGl. I will move it. Thanks!

OrbitLinuxTracing/TracerThread.cpp
MariaSandru27: It was thought to be more like a question. Such as "are these events the instrumented tracepoints that were selected". But yes, I see how this seems weird and would need to be changed.

OrbitCaptureClient/CaptureEventProcessorProcessEventsFuzzer.cpp
MariaSandru27: Thank you! I will change that.

OrbitLinuxTracing/PerfEvent.h
MariaSandru27: Ah...I have just seen that this is an outdated comment. In the meantime, TracepointEvent does not depend on anything.

PR #1169, opened 2020-09-03
Match exact version when deploying Orbit

beckerhe:
Checking the version that's installed on the gamelet is currently
checked with a simple `grep` which leads to `1.51` matches `1.51-333-xxxxxx`.

This should not be the case, so this commit changes the matching such
that always the full version number is matched.

This is only an issue when using developement versions of Orbit, so not
a customer-facing issue.

### DIFF
--- a/OrbitQt/servicedeploymanager.cpp
+++ b/OrbitQt/servicedeploymanager.cpp
@@ -84,8 +84,8 @@ outcome::result<bool> ServiceDeployManager::CheckIfInstalled() {
     // packages. So we have to remove it.
     version = version.substr(1);
   }
-  const auto command =
-      absl::StrFormat("/usr/bin/dpkg-query -W -f '${Version}' orbitprofiler 2>/dev/null | grep -xF '%s'", version);
+  const auto command = absl::StrFormat(
+      "/usr/bin/dpkg-query -W -f '${Version}' orbitprofiler 2>/dev/null | grep -xF '%s'", version);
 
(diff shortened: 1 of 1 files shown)


OrbitQt/servicedeploymanager.cpp
```diff
@@ -85,7 +85,7 @@ outcome::result<bool> ServiceDeployManager::CheckIfInstalled() {
     version = version.substr(1);
   }
   const auto command =
-      absl::StrFormat("/usr/bin/dpkg-query -W orbitprofiler 2>/dev/null | grep %s", version);
+      absl::StrFormat("/usr/bin/dpkg-query -W orbitprofiler 2>/dev/null | grep -E '%s$'", version);
```
antonrohr, line 88: I am unsure if this `-E '%s$'` does exactly what you want. `version` is a simple (not escaped) std::string, right? So for an example of `version` would be "1.52", which results in `grep -E '1.52$'`. This would then also match i.e. "1352". 
`echo "1352" | grep -E '1.52$'`

beckerhe: Yes. But the problem has been there before and is unrelated to that change. But yeah, maybe we should fix it.

beckerhe: I've just pushed a different solution. PTAL. Thanks.

antonrohr (approved):
Thank you! I think this solves it properly.

PR #1172, opened 2020-09-03
Capture in progress is cleaned up before exiting Orbit

reichlfl:
Bugfix: When Orbit was closed while a capture was in progress, the process never exited correctly. This is now fixed, and the user is asked if the running capture should be aborted when closing.

To test:
Start a capture, close Orbit's main window. Choose Yes: Orbit closes after capture processing has finished. Choose No: Orbit does not close, capturing continues.

image

### DIFF
--- a/OrbitQt/orbitmainwindow.cpp
+++ b/OrbitQt/orbitmainwindow.cpp
@@ -560,3 +560,19 @@ void OrbitMainWindow::on_actionServiceStackOverflow_triggered() {
 }
 
 void OrbitMainWindow::OnCaptureCleared() { ui->liveFunctions->Reset(); }
+
+void OrbitMainWindow::closeEvent(QCloseEvent* event) {
+  if (GOrbitApp->IsCapturing()) {
+    event->ignore();
+
+    if (QMessageBox::question(this, "Capture in progress",
+                              "A capture is currently in progress. Do you want to abort the "
+                              "capture and exit Orbit?") == QMessageBox::Yes) {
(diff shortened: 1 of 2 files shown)


dpallotti (approved):
Looks good, but I have one doubt: If I'm not mistaken, there are cases when the time between `StopCapture` and `OrbitApp::OnCaptureComplete` can be quite long, in particular if data is received by the ui slower than it is produced by the collector (e.g., the connection is slow). One way we were achieving this quite consistently was by instrumenting Trata's `execute_task`, which is called hundreds of thousands of times per second. This is a more general problem as stopping the capture takes a long time regardless of closing Orbit, but I wanted to bring it to your attention as in such a case the behavior when trying to close Orbit might be unexpected.

reichlfl:
> Looks good, but I have one doubt: If I'm not mistaken, there are cases when the time between `StopCapture` and `OrbitApp::OnCaptureComplete` can be quite long, in particular if data is received by the ui slower than it is produced by the collector (e.g., the connection is slow). One way we were achieving this quite consistently was by instrumenting Trata's `execute_task`, which is called hundreds of thousands of times per second. This is a more general problem as stopping the capture takes a long time regardless of closing Orbit, but I wanted to bring it to your attention as in such a case the behavior when trying to close Orbit might be unexpected.

True, I was also thinking about that - but since currently the application would just hang completely it's at least an improvement, and I wanted to keep the scope quite narrow for this. We should investigate that in a new bug.

florian-kuebler (commented):
LGTM Thanks.

PR #1173, opened 2020-09-03
Change EventBuffer::GetCallstackEvents to ...InTimeRangeForThreadId

dpallotti:
Fixes a bug caused by the fact that when
`SamplingProfiler::ProcessSamples` is called for a selection, the
`CallstackData` passed to it contains each CallstackEvent twice, once
for its thread and once for the fake -1 thread. But
`SamplingProfiler::ProcessSamples` assumes the opposite.

Bug: http://b/167672728

The original version is not needed as this method is only used in one place.

### DIFF
--- a/OrbitCore/EventBuffer.cpp
+++ b/OrbitCore/EventBuffer.cpp
@@ -12,11 +12,14 @@ using orbit_client_protos::CallstackEvent;
 
 EventTracer GEventTracer;
 
-std::vector<CallstackEvent> EventBuffer::GetCallstackEvents(
+std::vector<CallstackEvent> EventBuffer::GetCallstackEventsInTimeRangeForThreadId(
     uint64_t time_begin, uint64_t time_end, int32_t thread_id /*= kAllThreadsFakeTid*/) const {
   std::vector<CallstackEvent> callstack_events;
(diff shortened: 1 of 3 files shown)


OrbitCore/EventBuffer.cpp
```diff
@@ -12,12 +12,15 @@ using orbit_client_protos::CallstackEvent;
 
 EventTracer GEventTracer;
 
-std::vector<CallstackEvent> EventBuffer::GetCallstackEvents(
+std::vector<CallstackEvent> EventBuffer::GetCallstackEventsExcludingAllThreadsTid(
```
pierricgimmig, line 15: nit. I would rename to GetCallstackEventsInTimeRangeForThreadId. "ExcludingAllThreadsTid" seems more confusing than anything.

pierricgimmig (approved):
Cool, thanks for this Davide. LGTM, but please see minor comment.

florian-kuebler (approved):
Thanks for that change! As Pierric suggested, I would also rename the function. But otherwise LGTM!

OrbitCore/EventBuffer.cpp
dpallotti: Done.

PR #1178, opened 2020-09-07
Add a top-down view for the selection

dpallotti:
Adds a "Top-Down (selection)" tab whose top-down view is constructed from the
selected callstacks only.
Also renames the "selection" tab to "Sampling (selection)" for clarity.

Bug: http://b/162706739

### DIFF
--- a/OrbitGl/App.cpp
+++ b/OrbitGl/App.cpp
@@ -490,8 +490,20 @@ void OrbitApp::SetTopDownView(const CaptureData& capture_data) {
     return;
   }
 
-  std::unique_ptr<TopDownView> top_down_view = TopDownView::CreateFromCaptureData(capture_data);
+  std::unique_ptr<TopDownView> top_down_view =
+      TopDownView::CreateFromSamplingProfiler(capture_data.sampling_profiler(), capture_data);
   top_down_view_callback_(std::move(top_down_view));
+}
+
+void OrbitApp::SetSelectionTopDownView(const SamplingProfiler& selection_sampling_profiler,
+                                       const CaptureData& capture_data) {
+  if (!selection_top_down_view_callback_) {
+    return;
+  }
+
+  std::unique_ptr<TopDownView> selection_top_down_view =
+      TopDownView::CreateFromSamplingProfiler(selection_sampling_profiler, capture_data);
+  selection_top_down_view_callback_(std::move(selection_top_down_view));
 }
 
 std::string OrbitApp::GetCaptureFileName() {
@@ -704,6 +716,7 @@ void OrbitApp::ClearCapture() {
 
   SetSamplingReport(empty_profiler, empty_unique_callstacks);
   SetTopDownView(GetCaptureData());
+  SetSelectionTopDownView(empty_profiler, GetCaptureData());
 
   if (selection_report_) {
     SetSelectionReport(std::move(empty_profiler), empty_unique_callstacks, false);
@@ -919,8 +932,7 @@ void OrbitApp::LoadSymbols(const std::filesystem::path& symbols_path,
 
       modules_currently_loading_.erase(module->m_FullName);
 
-      UpdateSamplingReport();
-      SetTopDownView(GetCaptureData());
+      UpdateAfterSymbolLoading();
       GOrbitApp->FireRefreshCallbacks();
     });
   });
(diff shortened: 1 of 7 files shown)


ronaldfw:
How much work would it be to, instead of having additional selection tabs, have a checkbox each on the "Sampling" and "Top down" tabs that restricts the view to the selection? (Checkbox: "Show selection only").

(I think @reichlfl suggested this when we talked about this last week.)

OrbitGl/App.cpp
```diff
@@ -490,10 +490,22 @@ void OrbitApp::SetTopDownView(const CaptureData& capture_data) {
...
 
-  std::unique_ptr<TopDownView> top_down_view = TopDownView::CreateFromCaptureData(capture_data);
+  std::unique_ptr<TopDownView> top_down_view =
+      TopDownView::CreateFromSamplingProfiler(capture_data.sampling_profiler(), capture_data);
   top_down_view_callback_(std::move(top_down_view));
 }
 
+void OrbitApp::SetSelectionTopDownView(const SamplingProfiler& selection_sampling_profiler,
+                                       const CaptureData& capture_data) {
+  if (!selection_top_down_view_callback_) {
```
florian-kuebler, line 500: Shouldn't that be a CHECK?

dpallotti:
> How much work would it be to, instead of having additional selection tabs, have a checkbox each on the "Sampling" and "Top down" tabs that restricts the view to the selection? (Checkbox: "Show selection only").

This is a bit out of scope here, but anyway, is that something we want? What would be the benefit? I would think it's less intuitive than having dedicated tabs. In some cases it would also cause the need for additional clicks: consider for example going from the top-down for the selection to the overall sampling report, which would then require switching tab and also unchecking the check box.

OrbitGl/App.cpp
```diff
@@ -704,6 +716,7 @@ void OrbitApp::ClearCapture() {
 
   SetSamplingReport(empty_profiler, empty_unique_callstacks);
   SetTopDownView(GetCaptureData());
+  SetSelectionTopDownView(empty_profiler, GetCaptureData());
```
florian-kuebler, line 719: I saw, that you were grouping the update of the sampling related views below, maybe this should be also done here?

florian-kuebler (approved):
LGTM. Thank you very much. Left some minor comments.

ronaldfw:
> > How much work would it be to, instead of having additional selection tabs, have a checkbox each on the "Sampling" and "Top down" tabs that restricts the view to the selection? (Checkbox: "Show selection only").
> 
> This is a bit out of scope here, but anyway, is that something we want? What would be the benefit? I would think it's less intuitive than having dedicated tabs. In some cases it would also cause the need for additional clicks: consider for example going from the top-down for the selection to the overall sampling report, which would then require switching tab and also unchecking the check box.

Agree, some UI details have to be worked out and refined, but I think at some point we run into the problem of having too many tabs. What if we also have a bottom-up view, butterfly view, the "Live" tab is also supposed to have a selection, tracepoints with/without selection, etc.

florian-kuebler:
> > > How much work would it be to, instead of having additional selection tabs, have a checkbox each on the "Sampling" and "Top down" tabs that restricts the view to the selection? (Checkbox: "Show selection only").
> > 
> > 
> > This is a bit out of scope here, but anyway, is that something we want? What would be the benefit? I would think it's less intuitive than having dedicated tabs. In some cases it would also cause the need for additional clicks: consider for example going from the top-down for the selection to the overall sampling report, which would then require switching tab and also unchecking the check box.
> 
> Agree, some UI details have to be worked out and refined, but I think at some point we run into the problem of having too many tabs. What if we also have a bottom-up view, butterfly view, the "Live" tab is also supposed to have a selection, tracepoints with/without selection, etc.

I actually had a similar thought, but in a different direction. Sampling View and Top Down View are basically two different views of the same data. I guess there should be a sampling tab in which the user can toggle the view. This would also e.g. easily avoid having too many tabs when adding a buttom-up view.

OrbitGl/App.cpp
dpallotti: I introduced `UpdateAfterCaptureCleared`, please have a look.

florian-kuebler (approved):
Thanks! LGTM.

dpallotti:
> Agree, some UI details have to be worked out and refined, but I think at some point we run into the problem of having too many tabs. What if we also have a bottom-up view, butterfly view, the "Live" tab is also supposed to have a selection, tracepoints with/without selection, etc.

Yes, we will run into the problem of having too many tabs and I agree that we need to start thinking about that. But with the current number of tabs I have to say that the current arrangement is the one that convinces me the most among the ones we have thought about. (We do need to hide those "selection" tabs when no selection is active, but that's a different story.)

> I actually had a similar thought, but in a different direction. Sampling View and Top Down View are basically two different views of the same data. I guess there should be a sampling tab in which the user can toggle the view. This would also e.g. easily avoid having too many tabs when adding a buttom-up view.

I remember having a similar conversation with Pierric. Having everything under a single tab in my opinion overloads such tab. The "sampling" tab already contains a lot of stuff, with all the sub-tabs for each thread and the callstack view at the bottom. Adding a top-down view to it and controls to switch to it is too much.

ronaldfw:
> Yes, we will run into the problem of having too many tabs and I agree that we need to start thinking about that. But with the current number of tabs I have to say that the current arrangement is the one that convinces me the most among the ones we have thought about. (We do need to hide those "selection" tabs when no selection is active, but that's a different story.)

Yes, I think for now it's fine, but selection is probably the one that we have to do differently at some point (as it doubles the tabs that have data that can be selected). 
 
> > I actually had a similar thought, but in a different direction. Sampling View and Top Down View are basically two different views of the same data. I guess there should be a sampling tab in which the user can toggle the view. This would also e.g. easily avoid having too many tabs when adding a buttom-up view.
> 
> I remember having a similar conversation with Pierric. Having everything under a single tab in my opinion overloads such tab. The "sampling" tab already contains a lot of stuff, with all the sub-tabs for each thread and the callstack view at the bottom. Adding a top-down view to it and controls to switch to it is too much.

I also think that navigation in the regular sampling view and the top-down view is so different that having that switchable in the same tab could be very confusing.

florian-kuebler:
>>> I actually had a similar thought, but in a different direction. Sampling View and Top Down View are basically two different views of the same data. I guess there should be a sampling tab in which the user can toggle the view. This would also e.g. easily avoid having too many tabs when adding a buttom-up view.

>>I remember having a similar conversation with Pierric. Having everything under a single tab in my opinion overloads such tab. The "sampling" tab already contains a lot of stuff, with all the sub-tabs for each thread and the callstack view at the bottom. Adding a top-down view to it and controls to switch to it is too much.

>I also think that navigation in the regular sampling view and the top-down view is so different that having that switchable in the same tab could be very confusing.

Right might be the case (depending how it looks like). I am probably just used to that, as the profilers I used previously (mainly Yourkit and JVisualVM) had it implemented that way.

PR #1179, opened 2020-09-07
Fix "read access violation" error in CaptureSerializer::Save

kisenshi:
When saving a capture, an exception is thrown. Updated the code to fix it.
The error comes from a small typo in the stop condition in the loop that makes the call to _CaptureSerializer::Write_


ronaldfw (approved):

florian-kuebler (approved):
Thanks for spotting this! LGTM

kisenshi:
Thanks! merging into master

PR #1180, opened 2020-09-07
Split CaptureSerialization in CaptureSerialization and CaptureDeserialization

kisenshi:
_OrbitGl/CaptureSerializer_ is split in two: _OrbitClientModel/CaptureSerializer_ and _OrbitGl/CaptureDeserializer_. The objective is making the serialization of capture data independent from _OrbitGl_ so it can be used in other clients.

The modification has been carried out in a series of different commits to be able to highlight the code that is removed in each of the classes. Git is not able to track the split of a file in two so the copy of the file in two different files has been carried out in distinct commits. The "main" commit containing the removal of unnecessary code and dependencies clean, etc is the last one. The changes carried out are:

+ _OrbitClientModel/CaptureSerializer_: removed dependencies to OrbitGl and methods related to Load a capture

+ _OrbitGl/CaptureDeserializer_: removed methods related to Write a capture

+ _CaptureSerializerLoadFuzzer_: renamed to CaptureDeserializerLoadFuzzer, fixed dependencies and proto to match the name change in the class

+ Added new project independent to _OrbitGl_, fixed dependencies, includes, etc

Test: saving and loading a capture seems to work as before.

Bug: http://b/167200870


kisenshi:
Had an issue with clang format so I pushed the files updated, should pass the tests now.

OrbitClientModel/CaptureSerializer.cpp
```diff
@@ -0,0 +1,82 @@
...
+using orbit_client_protos::CallstackEvent;
+using orbit_client_protos::CallstackInfo;
+using orbit_client_protos::CaptureInfo;
+using orbit_client_protos::FunctionInfo;
+using orbit_client_protos::FunctionStats;
+using orbit_client_protos::TimerInfo;
+
+void CaptureSerializer::WriteMessage(const google::protobuf::Message* message,
+                                     google::protobuf::io::CodedOutputStream* output) {
+  uint32_t message_size = message->ByteSizeLong();
```
danielfenner, line 31: `ByteSizeLong()` returns a `size_t`. So this is usally 64 bit wide. 

`uint64_t` and `WriteLittleEndian64 `would probably have been a better choice. I'm not 100% sure about changing it. My wild guess is that we don't handle captures >4GB great currently (so we don't really win anything). 
On the other hand captures >4GB might be a thing in the long run and the earlier we fix this the easier it is.

florian-kuebler:
Thanks for this, this is very valuable. Haven't looked in the code yet, but can you add "Bug: http://b/167200870" to the end of the commit message?

OrbitClientModel/CaptureSerializer.cpp
```diff
@@ -0,0 +1,82 @@
...
+    // Fix names in address infos (some might only be in process):
+    added_address_info->set_function_name(
+        capture_data.GetFunctionNameByAddress(added_address_info->absolute_address()));
+  }
+
+  const absl::flat_hash_map<uint64_t, FunctionStats>& functions_stats =
+      capture_data.functions_stats();
+  capture_info.mutable_function_stats()->insert(functions_stats.begin(), functions_stats.end());
+
+  // TODO: this is not really synchronized, since GetCallstacks processing below
```
florian-kuebler, line 63: I guess the method was renamed to `GetCallstackData`. Can you adjust this?

OrbitClientModel/include/OrbitClientModel/CaptureSerializer.h
```diff
@@ -15,7 +15,6 @@
 
 #include "CaptureData.h"
 #include "OrbitBase/Result.h"
-#include "TimeGraph.h"
 #include "capture_data.pb.h"
 
 class CaptureSerializer {
```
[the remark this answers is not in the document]
florian-kuebler: Agree.

OrbitGl/CaptureDeserializer.h
```diff
@@ -0,0 +1,40 @@
...
+#include <iosfwd>
+#include <outcome.hpp>
+#include <string>
+
+#include "CaptureData.h"
+#include "OrbitBase/Result.h"
+#include "TimeGraph.h"
+#include "capture_data.pb.h"
+
+class CaptureDeserializer {
```
florian-kuebler, line 21: As a follow-up (in a different PR) and a simple first step into also making the Deserializer independent, can you also make this class to be just a namespace with static methods and pass the time_graph_ as parameter?

florian-kuebler (commented):
Thanks a lot for the change. I left some minor comments.

OrbitClientModel/CaptureSerializer.cpp
beckerhe: For clarification:
- AFAIK protobuf has a maximum message size of 2 gigs.
- A serialized capture is a stream of protos, meaning `WriteMessage` will only encode a single timer at once which should not be more than a couple of bytes.

florian-kuebler: Thanks for that clarification!

OrbitClientModel/CaptureSerializer.cpp
```diff
@@ -0,0 +1,82 @@
+// Copyright (c) 2020 The Orbit Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#include "OrbitClientModel/CaptureSerializer.h"
+
+#include <google/protobuf/io/coded_stream.h>
```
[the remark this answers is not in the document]
kisenshi: Do you mean changing the order of the includes? clang format order the headers like this when I save the file so I can't really move them around

OrbitClientModel/CaptureSerializer.cpp
danielfenner: As discussed above and in the standup - this is not a problem. Should stay as is.

OrbitClientModel/include/OrbitClientModel/CaptureSerializer.h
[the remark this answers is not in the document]
kisenshi: you are right, will update the code, thanks!

OrbitGl/CaptureDeserializer.h
```diff
@@ -0,0 +1,40 @@
...
+  TimeGraph* time_graph_;
+
+  static bool ReadMessage(google::protobuf::Message* message,
+                          google::protobuf::io::CodedInputStream* input);
+
+ private:
+  [[nodiscard]] CaptureData GenerateCaptureData(
+      const orbit_client_protos::CaptureInfo& capture_info);
+
+  orbit_client_protos::CaptureHeader header_;
```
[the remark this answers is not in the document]
kisenshi: Ok, will update the code, thanks!

OrbitClientModel/CaptureSerializer.cpp
kisenshi: Discussed offline, I will leave it as it is. Thanks!

OrbitGl/CaptureDeserializer.h
kisenshi: Sure, makes sense; will do that next in a separate PR, thanks!

OrbitClientModel/CaptureSerializer.cpp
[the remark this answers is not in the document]
danielfenner: Ah, okay. 
I'd use use "..." instead of <...> for everything except standard library header files.  If clang_format then still has an opinion don't fight it ...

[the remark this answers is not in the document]
kisenshi: Ok, I will update them to use "..." instead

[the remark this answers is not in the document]
kisenshi: Ah! And with this change _clang_format_ moves it below! great, thanks!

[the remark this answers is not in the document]
kisenshi: Done!

OrbitClientModel/include/OrbitClientModel/CaptureSerializer.h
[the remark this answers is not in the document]
kisenshi: Done, thanks!

OrbitGl/CaptureDeserializer.h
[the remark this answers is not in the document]
kisenshi: Done, thanks!

OrbitClientModel/CaptureSerializer.cpp
kisenshi: Done, thanks!

kisenshi:
I have updated to code to address the suggestions in the review. Thanks!

OrbitClientModel/include/OrbitClientModel/CaptureSerializer.h
```diff
@@ -0,0 +1,87 @@
...
+ErrorMessageOr<void> Save(const std::string& filename, const CaptureData& capture_data,
+                          const absl::flat_hash_map<uint64_t, std::string>& key_to_string_map,
+                          TimersIterator timers_iterator_begin, TimersIterator timers_iterator_end);
+
+void WriteMessage(const google::protobuf::Message* message,
+                  google::protobuf::io::CodedOutputStream* output);
+
+}  // namespace CaptureSerializer
+
+namespace internal {
```
florian-kuebler, line 31: I think you should make this an anonymous namespace (and move the implementation into this header file).
@danielfenner please correct me, if I am wrong.

OrbitClientModel/include/OrbitClientModel/CaptureSerializer.h
```diff
@@ -0,0 +1,87 @@
...
+
+template <class TimersIterator>
+void internal::Save(std::ostream& stream, const CaptureData& capture_data,
+                    const absl::flat_hash_map<uint64_t, std::string>& key_to_string_map,
+                    TimersIterator timers_iterator_begin, TimersIterator timers_iterator_end) {
+  google::protobuf::io::OstreamOutputStream out_stream(&stream);
+  google::protobuf::io::CodedOutputStream coded_output(&out_stream);
+
+  orbit_client_protos::CaptureHeader header;
+  header.set_version(kRequiredCaptureVersion);
```
kisenshi, line 54: I have a question here which I just realise: is it needed to use `internal::kRequiredCaptureVersion` here instead? If not, why? If it does, why does the code compile and works? Thanks!

florian-kuebler (commented):
I would avoid having an accessible namespace "internal" for private only functions and rather use an anonymous namespace instead.

OrbitClientModel/include/OrbitClientModel/CaptureSerializer.h
kisenshi: I tried using anonymous and it was not working, did not compile and had linking errors. I made a quick search and apparently anonymous namespaces should not be used in a header. However, I may be wrong

OrbitClientModel/include/OrbitClientModel/CaptureSerializer.h
beckerhe: You don't need the `internal::` because it's in the same namespace as the function since `Save` is also part of `internal`.

OrbitClientModel/include/OrbitClientModel/CaptureSerializer.h
beckerhe: I think that's not a good idea. An anonymous namespace only makes sense in a cpp-file. I would stick to the internal-namespace but rather move into the `CaptureSerializer`-namespace because `internal::Save` is too generic. This name could occur somewhere else in the codebase as well but `CaptureSerializer::internal::Save` is fine.

And just like Florian said, I would also move the implementation from below up here. I don't see a need to have a separate declaration in this case.

beckerhe: And I'm sorry for chiming in. :see_no_evil:

florian-kuebler: Thanks for clarifying this. So then I would also appreciate a not that generic name. @beckerhe do you have a link or so, why anonymous namespace is not a good idea here?

OrbitClientModel/include/OrbitClientModel/CaptureSerializer.h
kisenshi: Ok, I understand, thanks a lot!

OrbitClientModel/include/OrbitClientModel/CaptureSerializer.h
kisenshi: Ok, I will update the code. Thanks a lot for the clarification :)

beckerhe: I don't have a link right away. But I think it's pretty clear from what an anonymous namespace does. First of all, it does NOT affect name resolution (that's not a 100% true, but almost). That means for the compiler it's like if there were no namespace. So the `Save` would be in the global namespace. The only thing an anonymous namespace does it changes the linkage of the symbols defined inside to internal. "internal linkage" means the symbols are only visible in the current compilation unit. That means every compilation-unit including this header file will instantiate their own version of this templated function. Instead of having only one copy of that function in the binary we will have as many as we have calls.

beckerhe: The best thing I can offer right away is the reference: https://en.cppreference.com/w/cpp/language/namespace#Unnamed_namespaces

kisenshi: I have updated the code, thanks!

florian-kuebler (approved):
LGTM. Thanks!

OrbitClientModel/include/OrbitClientModel/CaptureSerializer.h
florian-kuebler: Thanks (both for clarification and for the update)

OrbitClientModel/include/OrbitClientModel/CaptureSerializer.h
```diff
@@ -0,0 +1,83 @@
...
+  for (auto it = timers_iterator_begin; it != timers_iterator_end; ++it) {
+    CaptureSerializer::WriteMessage(&(*it), &coded_output);
+  }
+}
+
+}  // namespace internal
+
+}  // namespace CaptureSerializer
+
+template <class TimersIterator>
```
danielfenner, line 62: I would move the implementation up to where the definition is or at least before the` }  // namespace CaptureSerializer`.

OrbitClientModel/CaptureSerializer.cpp
```diff
@@ -0,0 +1,80 @@
...
+#include "google/protobuf/message.h"
+
+using orbit_client_protos::CallstackEvent;
+using orbit_client_protos::CallstackInfo;
+using orbit_client_protos::CaptureInfo;
+using orbit_client_protos::FunctionInfo;
+using orbit_client_protos::FunctionStats;
+using orbit_client_protos::TimerInfo;
+
+void CaptureSerializer::WriteMessage(const google::protobuf::Message* message,
```
danielfenner, line 28: The usual way of handling namespaces in cc files is to use them as in the headers:
```
namespace capture_serializer {

void WriteMessage(const google::protobu ...

namespace internal{

CaptureInfo GenerateCaptureInfo( ...

}  // internal

}  // capture_serializer 

```

OrbitClientModel/include/OrbitClientModel/CaptureSerializer.h
kisenshi: I can't move it where the definition is because it uses internal::Save but I moved it inside the namespace. Thanks!

OrbitClientModel/include/OrbitClientModel/CaptureSerializer.h
```diff
@@ -0,0 +1,83 @@
...
+#include <string>
+
+#include "CaptureData.h"
+#include "OrbitBase/Result.h"
+#include "capture_data.pb.h"
+#include "google/protobuf/io/coded_stream.h"
+#include "google/protobuf/io/zero_copy_stream_impl.h"
+#include "google/protobuf/message.h"
+
+namespace CaptureSerializer {
```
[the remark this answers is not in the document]
kisenshi: Didn't realise this; I have updated it. Thanks for letting me know!

OrbitClientModel/include/OrbitClientModel/CaptureSerializer.h
```diff
@@ -0,0 +1,83 @@
...
+template <class TimersIterator>
+void Save(std::ostream& stream, const CaptureData& capture_data,
+          const absl::flat_hash_map<uint64_t, std::string>& key_to_string_map,
+          TimersIterator timers_iterator_begin, TimersIterator timers_iterator_end) {
+  google::protobuf::io::OstreamOutputStream out_stream(&stream);
+  google::protobuf::io::CodedOutputStream coded_output(&out_stream);
+
+  orbit_client_protos::CaptureHeader header;
+  header.set_version(kRequiredCaptureVersion);
+  CaptureSerializer::WriteMessage(&header, &coded_output);
```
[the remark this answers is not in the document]
kisenshi: Updated; thanks for noticing!

OrbitClientModel/include/OrbitClientModel/CaptureSerializer.h
kisenshi: I can't move it where the definition is because it uses internal::Save but I moved it inside the namespace. Thanks!

OrbitClientModel/CaptureSerializer.cpp
kisenshi: Oh ok, I didn't know; I have updated it; thanks :)

kisenshi:
I have updated the code to address the suggestions in the review. Thanks!

OrbitClientModel/include/OrbitClientModel/CaptureSerializer.h
danielfenner: Ahh, I see. Thanks!

danielfenner (approved):
LGTM!

kisenshi:
Merging this now, thanks!

PR #1182, opened 2020-09-08
Allow Ctrl+C and context-action copying from the top-down view

dpallotti:


ronaldfw (approved):
LGTM! Thanks also for cleaning up (renamings)!

OrbitGl/DataView.cpp
```diff
@@ -102,19 +100,19 @@ void DataView::ExportCSV(const std::string& a_FileName) {
 
 void DataView::CopySelection(const std::vector<int>& selection) {
   std::string clipboard;
-  size_t numColumns = GetColumns().size();
-  for (size_t i = 0; i < numColumns; ++i) {
+  size_t num_columns = GetColumns().size();
+  for (size_t i = 0; i < num_columns; ++i) {
     clipboard += GetColumns()[i].header;
-    if (i < numColumns - 1) clipboard += ", ";
+    if (i < num_columns - 1) clipboard += ", ";
```
beckerhe, line 106: Nit: I know this hasn't actually changed, but since you're refactoring it: I think that would benefit from absl::StrJoin: `clipboard = absl::StrJoin(GetColumns(), ", ", [](std::string* out, const auto& column) { out.append(column.header); });`

OrbitGl/DataView.cpp
```diff
@@ -102,19 +100,19 @@ void DataView::ExportCSV(const std::string& a_FileName) {
...
-  size_t numElements = GetNumElements();
+  size_t num_elements = GetNumElements();
   for (size_t i : selection) {
-    if (i < numElements) {
-      for (size_t j = 0; j < numColumns; ++j) {
+    if (i < num_elements) {
+      for (size_t j = 0; j < num_columns; ++j) {
         clipboard += GetValue(i, j);
-        if (j < numColumns - 1) clipboard += ", ";
+        if (j < num_columns - 1) clipboard += ", ";
```
beckerhe, line 115: Nit: as before: absl::StrJoin

OrbitQt/CopyableTreeView.h
```diff
@@ -0,0 +1,29 @@
...
+#define ORBIT_QT_COPYABLE_TREE_VIEW_H_
+
+#include <QKeyEvent>
+#include <QTreeView>
+
+class CopyableTreeView : public QTreeView {
+  Q_OBJECT
+
+ public:
+  CopyableTreeView(QWidget* parent = nullptr) : QTreeView(parent) {}
```
beckerhe, line 15: explicit

OrbitQt/TopDownWidget.cpp
```diff
@@ -28,7 +28,30 @@ void TopDownWidget::SetTopDownView(std::unique_ptr<TopDownView> top_down_view) {
   ui_->topDownTreeView->sortByColumn(TopDownViewItemModel::kInclusive, Qt::DescendingOrder);
   ui_->topDownTreeView->header()->resizeSections(QHeaderView::ResizeToContents);
 
-  on_searchLineEdit_textEdited(ui_->searchLineEdit->text());
+  onSearchLineEditTextEdited(ui_->searchLineEdit->text());
+}
+
+static void CopyFromIndices(OrbitApp* app, const QModelIndexList& indices) {
```
beckerhe, line 34: I think this function should not depend on OrbitApp. Ideally it returns a std::string. So it can be tested independently. I would move the `app->SetClipboard`-call into the key-press handler or have a separate function.

OrbitQt/CopyableTreeView.h
```diff
@@ -0,0 +1,29 @@
...
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#ifndef ORBIT_QT_COPYABLE_TREE_VIEW_H_
+#define ORBIT_QT_COPYABLE_TREE_VIEW_H_
+
+#include <QKeyEvent>
+#include <QTreeView>
+
+class CopyableTreeView : public QTreeView {
```
beckerhe, line 11: My first thought was this is about the C++-object being copyable, so the name was a little bit misleading to me. But I don't have a good suggestion, so feel free to ignore that comment.

beckerhe: Edit: Maybe ClipboardEnabledTreeView or something like that? I'm not sure...

OrbitQt/TopDownWidget.cpp
```diff
@@ -174,6 +198,9 @@ void TopDownWidget::onCustomContextMenuRequested(const QPoint& point) {
 
   bool enable_disassembly = !functions.empty();
 
+  QModelIndexList selected_indices = ui_->topDownTreeView->selectionModel()->selectedIndexes();
+  bool enable_copy = !selected_indices.empty();
```
beckerhe, line 202: Maybe `ui_->topDownTreeView->selectionModel()->hasSelection()`?

beckerhe (approved):
Thanks!

OrbitQt/orbittreeview.cpp
```diff
@@ -79,8 +79,8 @@ void OrbitTreeView::Initialize(DataView* data_view, SelectionType selection_type
   setAlternatingRowColors(true);
 
   if (font_type == FontType::kFixed) {
-    const QFont fixedFont = QFontDatabase::systemFont(QFontDatabase::FixedFont);
```
florian-kuebler, line 82: Should this file (and the header) also be renamed to camel case?

florian-kuebler (approved):
LGTM. Thanks Davide for making this change!

OrbitGl/DataView.cpp
dpallotti: The last commit does this, please have a look if this is ok/better as there are now two nested `absl::StrJoin`s. In particular I don't like that I need to create an array with the column indices.

beckerhe: Ahh. I see your point. I'm sorry, I didn't see that GetValue needs an instance. I also don't particularly like that. And there is no good solution, at least not without a lazy-range-library like cppitertools or range-v3. So feel free to revert that.

OrbitQt/CopyableTreeView.h
beckerhe: Fine with me. Thanks.

OrbitGl/DataView.cpp
dpallotti: Yeah, I think I'll revert it also to keep the parallelism with `ExportCSV` above.

beckerhe (approved):
Ship it! Thanks.

PR #1183, opened 2020-09-08
Fix crash on symbol loading before any capture

dpallotti:
Bug: http://b/168003624


florian-kuebler (approved):
Ah that makes sense. Thanks for fixing this!

antonrohr (approved):
Thanks for taking care of this!

PR #1184, opened 2020-09-08
Receive the events in the client-side and store them

MariaSandru27:
Receive the events in the client-side and store them in a proto for later serialization.
There are two types of events that can be received from the server-side regarding a generic tracepoint. If the first type of event is being processed (consisting of a key and the corresponding tracepoint's name
and category), the hash is used as a key in a unordered map and is associated to the corresponding tracepoint info. This is done such that, when the second type of event is being processed (the one that does not have the tracepoint info and only the hash key), the client to be able to recognise the tracepoint info using the map filled by processing the first type of event. 
The second type of event also has additional information such as the process id, thread id, cpu and time.
A new type of hash is being created that corresponds to the tracepoint event. The difference between this and the hash in the server-side is that, this one recognises the tracepoint event and not the tracepoint info (name and category). It will later be used for capture drawing such that we will store all the tracepoints events mapped by this hash key.

### DIFF
--- a/OrbitCaptureClient/CaptureEventProcessor.cpp
+++ b/OrbitCaptureClient/CaptureEventProcessor.cpp
@@ -33,8 +33,10 @@ void CaptureEventProcessor::ProcessEvent(const CaptureEvent& event) {
       ProcessCallstackSample(event.callstack_sample());
       break;
     case CaptureEvent::kInternedTracepointInfo:
+      ProcessInternedTracepointInfo(event.interned_tracepoint_info());
       break;
     case CaptureEvent::kTracepointEvent:
+      ProcessTracepointEventInfo(event.tracepoint_event());
       break;
     case CaptureEvent::kFunctionCall:
       ProcessFunctionCall(event.function_call());
@@ -211,11 +213,50 @@ uint64_t CaptureEventProcessor::GetCallstackHashAndSendToListenerIfNecessary(
   return hash;
 }
 
+void CaptureEventProcessor::ProcessInternedTracepointInfo(
+    orbit_grpc_protos::InternedTracepointInfo interned_tracepoint_info) {
+  if (tracepoint_intern_pool_.contains(interned_tracepoint_info.key())) {
+    ERROR("Overwriting InternedTracepointInfo  with key %llu", interned_tracepoint_info.key());
+  }
+  tracepoint_intern_pool_.emplace(interned_tracepoint_info.key(),
+                                  std::move(*interned_tracepoint_info.mutable_intern()));
+}
+void CaptureEventProcessor::ProcessTracepointEventInfo(
+    const orbit_grpc_protos::TracepointEvent& tracepoint_event) {
+  orbit_grpc_protos::TracepointInfo tracepoint_info;
+
+  CHECK(tracepoint_event.tracepoint_info_or_key_case() ==
+        orbit_grpc_protos::TracepointEvent::kTracepointInfoKey);
+  tracepoint_info = tracepoint_intern_pool_[tracepoint_event.tracepoint_info_key()];
+
+  uint64_t hash = GetTracepointInfoHashAndSendToListenerIfNecessary(tracepoint_info);
+  orbit_client_protos::TracepointEventInfo tracepoint_event_info;
+  tracepoint_event_info.set_pid(tracepoint_event.pid());
+  tracepoint_event_info.set_tid(tracepoint_event.tid());
+  tracepoint_event_info.set_time(tracepoint_event.time());
+  tracepoint_event_info.set_cpu(tracepoint_event.cpu());
+  tracepoint_event_info.set_tracepoint_info_key(hash);
+
+  // TODO: Set the capture listener to the tracepoint event
+}
+
 uint64_t CaptureEventProcessor::GetStringHashAndSendToListenerIfNecessary(const std::string& str) {
   uint64_t hash = StringHash(str);
   if (!string_hashes_seen_.contains(hash)) {
     string_hashes_seen_.emplace(hash);
     capture_listener_->OnKeyAndString(hash, str);
+  }
+  return hash;
+}
+
+uint64_t CaptureEventProcessor::GetTracepointInfoHashAndSendToListenerIfNecessary(
+    const orbit_grpc_protos::TracepointInfo& tracepoint_info) {
+  uint64_t hash = StringHash(absl::StrCat(tracepoint_info.category(), ":", tracepoint_info.name()));
+
+  if (!tracepoint_hashes_seen_.contains(hash)) {
+    tracepoint_hashes_seen_.emplace(hash);
+
+    // TODO: Set the capture listener to the hash
   }
   return hash;
 }
--- a/OrbitCaptureClient/CaptureEventProcessor.cpp
+++ b/OrbitCaptureClient/CaptureEventProcessor.cpp
@@ -36,7 +36,7 @@ void CaptureEventProcessor::ProcessEvent(const CaptureEvent& event) {
       ProcessInternedTracepointInfo(event.interned_tracepoint_info());
       break;
     case CaptureEvent::kTracepointEvent:
-      ProcessTracepointEventInfo(event.tracepoint_event());
+      ProcessTracepointEvent(event.tracepoint_event());
       break;
     case CaptureEvent::kFunctionCall:
       ProcessFunctionCall(event.function_call());
@@ -216,12 +216,12 @@ uint64_t CaptureEventProcessor::GetCallstackHashAndSendToListenerIfNecessary(
 void CaptureEventProcessor::ProcessInternedTracepointInfo(
     orbit_grpc_protos::InternedTracepointInfo interned_tracepoint_info) {
   if (tracepoint_intern_pool_.contains(interned_tracepoint_info.key())) {
-    ERROR("Overwriting InternedTracepointInfo  with key %llu", interned_tracepoint_info.key());
+    ERROR("Overwriting InternedTracepointInfo with key %llu", interned_tracepoint_info.key());
   }
   tracepoint_intern_pool_.emplace(interned_tracepoint_info.key(),
                                   std::move(*interned_tracepoint_info.mutable_intern()));
 }
-void CaptureEventProcessor::ProcessTracepointEventInfo(
+void CaptureEventProcessor::ProcessTracepointEvent(
     const orbit_grpc_protos::TracepointEvent& tracepoint_event) {
   orbit_grpc_protos::TracepointInfo tracepoint_info;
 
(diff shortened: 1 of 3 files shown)


OrbitCaptureClient/CaptureEventProcessor.cpp
```diff
@@ -211,6 +213,33 @@ uint64_t CaptureEventProcessor::GetCallstackHashAndSendToListenerIfNecessary(
...
 
+void CaptureEventProcessor::ProcessInternedTracepointInfo(
+    orbit_grpc_protos::InternedTracepointInfo interned_tracepoint_info) {
+  if (tracepoint_intern_pool_.contains(interned_tracepoint_info.key())) {
+    ERROR("Overwriting InternedTracepointInfo  with key %llu", interned_tracepoint_info.key());
+  }
+  tracepoint_intern_pool_.emplace(interned_tracepoint_info.key(),
+                                  std::move(*interned_tracepoint_info.mutable_intern()));
+}
+void CaptureEventProcessor::ProcessTracepointEventInfo(
```
dpallotti, line 224: I think that this should be called `ProcessTracepointEvent` (the names of these methods are based on the parameters).

OrbitCaptureClient/CaptureEventProcessor.cpp
```diff
@@ -211,6 +213,33 @@ uint64_t CaptureEventProcessor::GetCallstackHashAndSendToListenerIfNecessary(
   return hash;
 }
 
+void CaptureEventProcessor::ProcessInternedTracepointInfo(
+    orbit_grpc_protos::InternedTracepointInfo interned_tracepoint_info) {
+  if (tracepoint_intern_pool_.contains(interned_tracepoint_info.key())) {
+    ERROR("Overwriting InternedTracepointInfo  with key %llu", interned_tracepoint_info.key());
```
dpallotti, line 219:
```suggestion
    ERROR("Overwriting InternedTracepointInfo with key %llu", interned_tracepoint_info.key());
```

OrbitCaptureClient/CaptureEventProcessor.cpp
```diff
@@ -211,6 +213,33 @@ uint64_t CaptureEventProcessor::GetCallstackHashAndSendToListenerIfNecessary(
...
+
+  uint64_t hash = GetTracepointInfoHashAndSendToListenerIfNecessary(tracepoint_info);
+  orbit_client_protos::TracepointEventInfo tracepoint_event_info;
+  tracepoint_event_info.set_pid(tracepoint_event.pid());
+  tracepoint_event_info.set_tid(tracepoint_event.tid());
+  tracepoint_event_info.set_time(tracepoint_event.time());
+  tracepoint_event_info.set_cpu(tracepoint_event.cpu());
+  tracepoint_event_info.set_tracepoint_info_key(hash);
+
+  // TODO: Set the capture listener to the tracepoint event
```
dpallotti, line 240: I'm a bit confused by the phrasing of this TODO.

MariaSandru27: Thanks for pointing that out! I will modify the phrasing to better express the intent of that TODO.

OrbitCaptureClient/CaptureEventProcessor.cpp
MariaSandru27: Sure! Thank you so much for pointing that out!

OrbitCaptureClient/CaptureEventProcessor.cpp
```diff
@@ -211,6 +213,33 @@ uint64_t CaptureEventProcessor::GetCallstackHashAndSendToListenerIfNecessary(
...
+  tracepoint_intern_pool_.emplace(interned_tracepoint_info.key(),
+                                  std::move(*interned_tracepoint_info.mutable_intern()));
+}
+void CaptureEventProcessor::ProcessTracepointEvent(
+    const orbit_grpc_protos::TracepointEvent& tracepoint_event) {
+  orbit_grpc_protos::TracepointInfo tracepoint_info;
+
+  CHECK(tracepoint_event.tracepoint_info_or_key_case() ==
+        orbit_grpc_protos::TracepointEvent::kTracepointInfoKey);
+  tracepoint_info = tracepoint_intern_pool_[tracepoint_event.tracepoint_info_key()];
```
danielfenner, line 230: Maybe add a 
`CHECK(tracepoint_intern_pool_.contains(tracepoint_event.tracepoint_info_key())`
to assert that the communication happens in order as we suspect.

Also: 
`auto tracepoint_info = ...` and remove the declaration above.

OrbitCaptureClient/CaptureEventProcessor.cpp
```diff
@@ -211,6 +213,33 @@ uint64_t CaptureEventProcessor::GetCallstackHashAndSendToListenerIfNecessary(
...
+
+  uint64_t hash = GetTracepointInfoHashAndSendToListenerIfNecessary(tracepoint_info);
```
danielfenner, line 232: I don't see why we need to "re-hash": Can't we just use `tracepoint_event_info.set_tracepoint_info_key(tracepoint_event.tracepoint_info_key());`
and propagate the `tracepoint_intern_pool_` to whoever is interested?

OrbitCaptureClient/CaptureEventProcessor.cpp
```diff
@@ -219,3 +248,15 @@ uint64_t CaptureEventProcessor::GetStringHashAndSendToListenerIfNecessary(const
...
 }
+
+uint64_t CaptureEventProcessor::GetTracepointInfoHashAndSendToListenerIfNecessary(
+    const orbit_grpc_protos::TracepointInfo& tracepoint_info) {
+  uint64_t hash = StringHash(absl::StrCat(tracepoint_info.category(), ":", tracepoint_info.name()));
+
+  if (!tracepoint_hashes_seen_.contains(hash)) {
+    tracepoint_hashes_seen_.emplace(hash);
+
+    // TODO: Set the capture listener to the hash
```
dpallotti, line 259: Can you rephrase this as well?

MariaSandru27: Of course!

OrbitCaptureClient/CaptureEventProcessor.cpp
MariaSandru27: Of course! Thank you!

OrbitCaptureClient/CaptureEventProcessor.cpp
```diff
@@ -211,6 +213,33 @@ uint64_t CaptureEventProcessor::GetCallstackHashAndSendToListenerIfNecessary(
...
+
+  uint64_t hash = GetTracepointInfoHashAndSendToListenerIfNecessary(tracepoint_info);
+  orbit_client_protos::TracepointEventInfo tracepoint_event_info;
+  tracepoint_event_info.set_pid(tracepoint_event.pid());
+  tracepoint_event_info.set_tid(tracepoint_event.tid());
+  tracepoint_event_info.set_time(tracepoint_event.time());
+  tracepoint_event_info.set_cpu(tracepoint_event.cpu());
+  tracepoint_event_info.set_tracepoint_info_key(hash);
+
+  // TODO: Store the tracepoint event in a unordered map that has as the key the hash computed
```
florian-kuebler, line 240: Can you please add your LDAP, such that later it is clear who left that todo. I.e. `TODO(ldap): `.

OrbitCaptureClient/CaptureEventProcessor.cpp
MariaSandru27: I have noticed that the hash mechanism is different in the client-side (even the names of the thread tracks are hashed differently , which are just strings) and I wanted to keep it consistent. I will have a chat with someone to make sure that the different hash mechanism is justified. However, of course, I can just use 'tracepoint_event.tracepoint_info_key()' as the hash for the tracepoint event.

dpallotti: There are a couple of reasons why also the other methods here re-hash but in general I guess it's mostly simplicity. This allows to have the intern pools local to the `CaptureEventProcessor` that handle interning during the communication, then other pools (not called like that anymore, e.g., `StringManager`) filled by the `CaptureListener` closer to where the strings are used by the UI (e.g., the capture view). Also keep in mind that the `CaptureEventProcessor` completely translates the protos for the communication (capture.proto) to the protos for the client (capture_data.proto). I think we discussed this problem a few weeks ago and didn't really come to a conclusion.

MariaSandru27: @dpallotti thanks for explaining! In this case, do you think it is justified to re-hash for keeping the intern pool local to the CaptureEventProcessor? I am also planning of having another pool for registering the tracepoints events (such as TracepointEventInfoManager) in another CL.

dpallotti: Since for callstacks and strings we already re-hash and duplicate the intern pool mechanism, for me it's fine to do the same for `TracepointInfo`s. If you want, you could at least avoid re-computing the hash, by changing `GetTracepointInfoHashAndSendToListenerIfNecessary` to `SendTracepointInfoToListenerIfNecessary` and passing a second argument `hash` with the hash that you already have (`tracepoint_event.tracepoint_info_key()`).

MariaSandru27: @danielfenner what do you think?

OrbitCaptureClient/CaptureEventProcessor.cpp
MariaSandru27: Thank you! I will.

OrbitCaptureClient/CaptureEventProcessor.cpp
MariaSandru27: For now, I have done what @dpallotti suggested to avoid re-hashing.

florian-kuebler (approved):
LGTM. Thanks!

dpallotti (approved):

OrbitCaptureClient/CaptureEventProcessor.cpp
danielfenner: Let's stick with "re-hashing" for now. Since all of this happens in the UI I am not too concerned about performance. If that hits us we can refactor later.
Plus it might be a bit weird to have the hash map from the service duplicated here since we can't reproduce the hash (we use std::hash in the service and hash functions are only required to produce the same result for the same input within a single execution of a program).

danielfenner (approved):
LGTM!

ronaldfw (approved):

OrbitCaptureClient/CaptureEventProcessor.cpp
MariaSandru27: Of course! Thank you @danielfenner!

MariaSandru27: As discussed offline, we will keep without the re-hash.

PR #1185, opened 2020-09-08
Always show Orbit's version in the title bar

beckerhe:
This change sets the QApplication::displayName property. This property
is automatically appended to all window titles. It simplifies window
title handling a bit since we don't need to do the append-operation in
code anymore.

### DIFF
--- a/OrbitQt/main.cpp
+++ b/OrbitQt/main.cpp
@@ -253,8 +253,15 @@ int main(int argc, char* argv[]) {
 #endif
 
     QApplication app(argc, argv);
-    QCoreApplication::setApplicationName("Orbit Profiler [BETA]");
-    QCoreApplication::setApplicationVersion(QString::fromStdString(OrbitCore::GetVersion()));
+    QApplication::setApplicationName("orbitprofiler");
+
+    // The application display name is automatically appended to all window titles when shown in the
+    // title bar: <specific window title> - <application display name>
+    const auto version_string = QString::fromStdString(OrbitCore::GetVersion());
+    QApplication::setApplicationDisplayName(
+        QString{"Orbit Profiler %1 [BETA]"}.arg(version_string));
+
+    QApplication::setApplicationVersion(version_string);
     path_to_executable = QCoreApplication::applicationFilePath();
 
 #ifdef ORBIT_CRASH_HANDLING
--- a/OrbitQt/orbitmainwindow.cpp
+++ b/OrbitQt/orbitmainwindow.cpp
@@ -78,7 +78,7 @@ OrbitMainWindow::OrbitMainWindow(QApplication* a_App, ApplicationOptions&& optio
     ui->actionOpen_Preset->setDisabled(true);
     ui->actionSave_Preset_As->setDisabled(true);
     ui->HomeTab->setDisabled(true);
-    SetTitle({});
+    setWindowTitle({});
   });
 
   auto finalizing_capture_dialog =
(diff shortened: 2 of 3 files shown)


OrbitQt/main.cpp
```diff
@@ -253,8 +253,14 @@ int main(int argc, char* argv[]) {
 #endif
 
     QApplication app(argc, argv);
-    QCoreApplication::setApplicationName("Orbit Profiler [BETA]");
-    QCoreApplication::setApplicationVersion(QString::fromStdString(OrbitCore::GetVersion()));
+    QApplication::setApplicationName("orbitprofiler");
+
+    // The application display name is automatically appended to all window titles when shown in the
+    // title bar: <specific window title> - <application display name>
+    const auto versionString = QString::fromStdString(OrbitCore::GetVersion());
```
dpallotti, line 260: `version_string`? Or are we sticking with Qt's casing here?

OrbitQt/main.cpp
```diff
@@ -253,8 +253,14 @@ int main(int argc, char* argv[]) {
 #endif
 
     QApplication app(argc, argv);
-    QCoreApplication::setApplicationName("Orbit Profiler [BETA]");
-    QCoreApplication::setApplicationVersion(QString::fromStdString(OrbitCore::GetVersion()));
+    QApplication::setApplicationName("orbitprofiler");
+
+    // The application display name is automatically appended to all window titles when shown in the
+    // title bar: <specific window title> - <application display name>
```
dpallotti, line 259: So if the window title is empty the " - " is not there, right?

OrbitQt/orbitmainwindow.cpp
```diff
@@ -353,7 +353,7 @@ void OrbitMainWindow::on_actionReport_Bug_triggered() {
 
 void OrbitMainWindow::on_actionAbout_triggered() {
   OrbitQt::OrbitAboutDialog dialog{this};
-  dialog.setWindowTitle(windowTitle());
+  dialog.setWindowTitle("About Orbit");
```
dpallotti, line 356: Do you want the "Orbit" part here when this will be followed by "Orbit Profiler ..." (maybe yes, just pointing this out)?

dpallotti (approved):
Nice.

OrbitQt/main.cpp
beckerhe: Ahh. Yesss.. Sorry. my bad. I will fix it.

OrbitQt/main.cpp
beckerhe: Exactly!

OrbitQt/orbitmainwindow.cpp
beckerhe: I'm undecided here. What do you prefer? `About - Orbit Profiler 1.52 [BETA]` or `About Orbit - Orbit Profiler 1.52 [BETA]`?

dpallotti: I think I prefer the first option. Not that critical though obviously.

beckerhe: Great. Let's take the first one. Absolutely fine with me! Thanks!

beckerhe:
I'm looping in @danielfenner since this might affect the E2E tests. Daniel, what do you think?

danielfenner:
> I'm looping in @danielfenner since this might affect the E2E tests. Daniel, what do you think?

`title_re='Orbit Profiler*'` is what I am matching in the title bar - so lgtm!

danielfenner (approved):

beckerhe:
> > I'm looping in @danielfenner since this might affect the E2E tests. Daniel, what do you think?
> 
> `title_re='Orbit Profiler*'` is what I am matching in the title bar - so lgtm!

Thanks!

PR #1186, opened 2020-09-09
Fix OrbitBase/TracingTest flakiness

pierricgimmig:
- Listener now owns the user callback and thread pool objects
- User callback now valid until thread pool is purged (should fix test)

I was never able to reproduce the failed test, but it must have been caused by the fact that we set the user callback to null *before* remaining in-flight scopes were processed

### DIFF
--- a/OrbitBase/Tracing.cpp
+++ b/OrbitBase/Tracing.cpp
@@ -54,7 +54,7 @@ void Listener::DeferScopeProcessing(const Scope& scope) {
   // User callback is called from a worker thread to
   // minimize contention on the instrumented threads.
   absl::MutexLock lock(&global_tracing_mutex);
-  if (IsActive() == false) return;
+  if (!IsActive()) return;
   global_tracing_listener->thread_pool_->Schedule([scope]() {
     absl::MutexLock lock(&global_tracing_mutex);
     if (IsActive() == false) return;
--- a/OrbitBase/Tracing.cpp
+++ b/OrbitBase/Tracing.cpp
@@ -40,7 +40,7 @@ Listener::Listener(std::unique_ptr<TimerCallback> callback) {
 
 Listener::~Listener() {
   absl::MutexLock lock(&global_tracing_mutex);
-  CHECK(IsActive() == true);
+  CHECK(IsActive());
   // Purge deferred scopes.
   thread_pool_->Shutdown();
   thread_pool_->Wait();
--- a/OrbitBase/Tracing.cpp
+++ b/OrbitBase/Tracing.cpp
@@ -11,67 +11,56 @@
 #include "OrbitBase/Profiling.h"
 #include "OrbitBase/ThreadPool.h"
 
+using orbit::tracing::Listener;
 using orbit::tracing::Scope;
 using orbit::tracing::ScopeType;
 using orbit::tracing::TimerCallback;
 
-ABSL_CONST_INIT static absl::Mutex global_callback_mutex(absl::kConstInit);
-ABSL_CONST_INIT static std::unique_ptr<TimerCallback> global_callback = {};
-ABSL_CONST_INIT static absl::Mutex global_thread_pool_mutex(absl::kConstInit);
-ABSL_CONST_INIT static std::unique_ptr<ThreadPool> global_thread_pool = {};
+ABSL_CONST_INIT static absl::Mutex global_tracing_mutex(absl::kConstInit);
+ABSL_CONST_INIT static Listener* global_tracing_listener = nullptr;
 
 static std::vector<Scope>& GetThreadLocalScopes() {
   thread_local std::vector<Scope> thread_local_scopes;
   return thread_local_scopes;
 }
 
-static void CreateCallbackThreadPool() {
-  constexpr size_t kMinNumThreads = 1;
-  constexpr size_t kMaxNumThreads = 1;
-  absl::MutexLock lock(&global_thread_pool_mutex);
-  CHECK(global_thread_pool == nullptr);
-  global_thread_pool = ThreadPool::Create(kMinNumThreads, kMaxNumThreads, absl::Milliseconds(500));
-}
-
-static void ShutdownCallbackThreadPool() {
-  absl::MutexLock lock(&global_thread_pool_mutex);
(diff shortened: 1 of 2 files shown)


OrbitBase/Tracing.cpp
```diff
@@ -11,68 +11,57 @@
...
+  global_tracing_listener = nullptr;
 }
 
 }  // namespace orbit::tracing
 
+void Listener::DeferScopeProcessing(const Scope& scope) {
+  // User callback is called from a worker thread to
+  // minimize contention on the instrumented threads.
+  absl::MutexLock lock(&global_tracing_mutex);
+  if (IsActive() == false) return;
```
dpallotti, line 57:
```suggestion
  if (!IsActive()) return;
```

OrbitBase/Tracing.cpp
```diff
@@ -11,68 +11,57 @@
...
 }
 
 Listener::~Listener() {
-  {
-    absl::MutexLock lock(&global_callback_mutex);
-    global_callback = nullptr;
-  }
-  ShutdownCallbackThreadPool();
+  absl::MutexLock lock(&global_tracing_mutex);
+  CHECK(IsActive() == true);
```
dpallotti, line 43:
```suggestion
  CHECK(IsActive());
```

OrbitBase/Tracing.cpp
```diff
@@ -11,68 +11,57 @@
...
 namespace orbit::tracing {
 
 Listener::Listener(std::unique_ptr<TimerCallback> callback) {
-  CreateCallbackThreadPool();
-  absl::MutexLock lock(&global_callback_mutex);
+  absl::MutexLock lock(&global_tracing_mutex);
   // Only one listener is supported.
-  CHECK(global_callback == nullptr);
-  global_callback = std::move(callback);
+  CHECK(IsActive() == false);
```
dpallotti, line 32:
```suggestion
  CHECK(!IsActive());
```

OrbitBase/Tracing.cpp
```diff
@@ -11,68 +11,57 @@
...
-  }
-  ShutdownCallbackThreadPool();
+  absl::MutexLock lock(&global_tracing_mutex);
+  CHECK(IsActive() == true);
+  // Purge deferred scopes.
+  thread_pool_->Shutdown();
+  thread_pool_->Wait();
+  active_ = false;
+  global_tracing_listener = nullptr;
 }
```
dpallotti, line 49: Should `user_callback_` also be set to `nullptr`?

OrbitBase/Tracing.cpp
```diff
@@ -11,68 +11,57 @@
...
 }  // namespace orbit::tracing
 
+void Listener::DeferScopeProcessing(const Scope& scope) {
+  // User callback is called from a worker thread to
+  // minimize contention on the instrumented threads.
+  absl::MutexLock lock(&global_tracing_mutex);
+  if (IsActive() == false) return;
+  global_tracing_listener->thread_pool_->Schedule([scope]() {
+    absl::MutexLock lock(&global_tracing_mutex);
+    if (IsActive() == false) return;
```
dpallotti, line 60:
```suggestion
    if (!IsActive()) return;
```

dpallotti (approved):
I think it would be useful if you could reproduce the fail and check that it's now gone. On Linux it works quite consistently for me to just run `OrbitBaseTests` a few times in a row.

pierricgimmig:
> I think it would be useful if you could reproduce the fail and check that it's now gone. On Linux it works quite consistently for me to just run `OrbitBaseTests` a few times in a row.

I must have tried 100 times, I can't reproduce the fail with the old code. I modified the test locally to spawn 1000 threads, still doesn't fail. It's a race condition that I could not repro on my machine...

pierricgimmig:
Hi @dpallotti , as mentioned I can't repro the failed test with the old code. If you have some time, could you try to repro with the new code?

dpallotti:
> > I think it would be useful if you could reproduce the fail and check that it's now gone. On Linux it works quite consistently for me to just run `OrbitBaseTests` a few times in a row.
> 
> I must have tried 100 times, I can't reproduce the fail with the old code. I modified the test locally to spawn 1000 threads, still doesn't fail. It's a race condition that I could not repro on my machine...

Oh wow, ok, I'll try the new code then.

dpallotti:
So I ran `OrbitBaseTests` repeatedly (Linux workstation, clang9, relwithdebinfo) and while it never fails, sometimes `Tracing.Scopes` now hangs indefinitely.

OrbitBase/Tracing.cpp
pierricgimmig: user_callback_ is now owned by the Listener instance, so setting it to nullptr in the destructor is not necessary. active_ is static that's why we set it to false, and global_tracing_listener is global so it needs to be reset as well.

dpallotti: Ah right.

pierricgimmig:
> So I ran `OrbitBaseTests` repeatedly (Linux workstation, clang9, relwithdebinfo) and while it never fails, sometimes `Tracing.Scopes` now hangs indefinitely.

Arrgh, ok thanks for your help Davide, I'll look into it a bit more.

dpallotti:
> > So I ran `OrbitBaseTests` repeatedly (Linux workstation, clang9, relwithdebinfo) and while it never fails, sometimes `Tracing.Scopes` now hangs indefinitely.
> 
> Arrgh, ok thanks for your help Davide, I'll look into it a bit more.

I added some prints and the hang is at the end of the big block, after all threads have been joined, so I would guess in `Listener`'s destructor.

dpallotti:
> I added some prints and the hang is at the end of the big block, after all threads have been joined, so I would guess in `Listener`'s destructor.

In `thread_pool_->Wait();`

pierricgimmig:
> I added some prints and the hang is at the end of the big block, after all threads have been joined, so I would guess in `Listener`'s destructor.

Yep, there was a deadlock if the destructor was called before the worker thread had finished processing the scopes. Thanks for letting me know, again I wasn't able to reproduce on my 72 cores Linux box, but could easily repro on my more reasonable Windows machine :-)

PR #1187, opened 2020-09-09
Resize top-down view columns to contents only once

dpallotti:
This prevents the annoying behavior of columns resizing every time the top-down
view is refreshed (new capture/new selection, symbols loaded), invalidating the
width set by the user.


beckerhe (approved):
LGTM!

PR #1188, opened 2020-09-09
Use progress bars in the "Inclusive" column of the top-down view

dpallotti:
This makes the distribution of samples in the tree much more immediate to see.

### DIFF
--- a/OrbitQt/TopDownWidget.cpp
+++ b/OrbitQt/TopDownWidget.cpp
@@ -418,6 +418,7 @@ void TopDownWidget::ProgressBarItemDelegate::paint(QPainter* painter,
 
   QStyleOptionProgressBar option_progress_bar;
   option_progress_bar.rect = option.rect;
+  option_progress_bar.palette = option.palette;
   option_progress_bar.minimum = 0;
   option_progress_bar.maximum = 100;
   option_progress_bar.progress = static_cast<int>(round(inclusive_percent));
@@ -432,16 +433,15 @@ void TopDownWidget::ProgressBarItemDelegate::paint(QPainter* painter,
       static_cast<int>(round(palette_highlight_color.value() * kBarColorValueReductionFactor)));
   option_progress_bar.palette.setColor(QPalette::Highlight, bar_foreground_color);
 
-  if (!highlight) {
-    option_progress_bar.text = index.data(Qt::DisplayRole).toString();
-    option_progress_bar.textVisible = true;
-  }
-  option.widget->style()->drawControl(QStyle::CE_ProgressBar, &option_progress_bar, painter);
+  option_progress_bar.text = index.data(Qt::DisplayRole).toString();
+  option_progress_bar.textVisible = true;
 
   if (highlight) {
-    QTextOption text_option;
-    text_option.setAlignment(Qt::AlignCenter);
-    painter->setPen(HighlightCustomFilterSortFilterProxyModel::kHighlightColor);
-    painter->drawText(option.rect, index.data(Qt::DisplayRole).toString(), text_option);
+    option_progress_bar.palette.setColor(
+        QPalette::Text, HighlightCustomFilterSortFilterProxyModel::kHighlightColor);
+    option_progress_bar.palette.setColor(
+        QPalette::HighlightedText, HighlightCustomFilterSortFilterProxyModel::kHighlightColor);
   }
+
+  option.widget->style()->drawControl(QStyle::CE_ProgressBar, &option_progress_bar, painter);
 }
--- a/OrbitQt/TopDownWidget.cpp
+++ b/OrbitQt/TopDownWidget.cpp
@@ -6,12 +6,29 @@
 
 #include <QColor>
 #include <QMenu>
+#include <QPainter>
+#include <QPalette>
+#include <QStyleOption>
 
 #include "App.h"
 #include "FunctionsDataView.h"
 #include "TopDownViewItemModel.h"
 
 using orbit_client_protos::FunctionInfo;
+
+TopDownWidget::TopDownWidget(QWidget* parent)
+    : QWidget{parent}, ui_{std::make_unique<Ui::TopDownWidget>()} {
+  ui_->setupUi(this);
+  ui_->topDownTreeView->setItemDelegateForColumn(TopDownViewItemModel::kInclusive,
+                                                 new ProgressBarItemDelegate{ui_->topDownTreeView});
+
+  connect(ui_->topDownTreeView, &CopyKeySequenceEnabledTreeView::copyKeySequencePressed, this,
+          &TopDownWidget::onCopyKeySequencePressed);
+  connect(ui_->topDownTreeView, &QTreeView::customContextMenuRequested, this,
+          &TopDownWidget::onCustomContextMenuRequested);
+  connect(ui_->searchLineEdit, &QLineEdit::textEdited, this,
+          &TopDownWidget::onSearchLineEditTextEdited);
+}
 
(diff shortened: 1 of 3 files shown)


dpallotti:
Screenshot:

Screenshot from 2020-09-09 12-22-43

beckerhe (approved):
This looks great. I'm just a little concerned about the contrast of the highlighted green text on the blue progress bar. Is there a chance to make the progress bar darker? Or making the green darker?

dpallotti:
> I'm just a little concerned about the contrast of the highlighted green text on the blue progress bar. Is there a chance to make the progress bar darker? Or making the green darker?

I thought this could be a problem, but on my screen it actually doesn't look that bad. Making the green darker actually makes things worse. As for making the progress bar darker (I would have to find a way first...) I then have yet another color in the overall color scheme, and the contrast with the dark background becomes then a bit worse. So I agree there's room for improvement but I'm not too sure how.

beckerhe:
> > I'm just a little concerned about the contrast of the highlighted green text on the blue progress bar. Is there a chance to make the progress bar darker? Or making the green darker?
> 
> I thought this could be a problem, but on my screen it actually doesn't look that bad. Making the green darker actually makes things worse. As for making the progress bar darker (I would have to find a way first...) I then have yet another color in the overall color scheme, and the contrast with the dark background becomes then a bit worse. So I agree there's room for improvement but I'm not too sure how.

Yeah. I kinda suspected that. Anyway, thanks for trying.

So from my side this is good to go, but maybe @ronaldfw has a good idea?

dpallotti:
Well, in the meantime I found out how to change the color of the progress bar, so I can experiment.

dpallotti:
So, since I've made it so far, a couple of extra options (note that I changed the background of the progress bar to make it actual length stand out):

Make the progress bar green instead of the text (I like this but maybe it's a bit unexpected):

Screenshot from 2020-09-09 16-01-46

Make the blue a bit darker (the color looks a bit "sad" though):

Screenshot from 2020-09-09 16-10-40

OrbitQt/TopDownWidget.cpp
```diff
@@ -374,3 +396,41 @@ QVariant TopDownWidget::HookedIdentityProxyModel::data(const QModelIndex& index,
       QStringLiteral("] ");
   return kDisplayHookedPrefix + data.toString();
 }
+
+void TopDownWidget::ProgressBarItemDelegate::paint(QPainter* painter,
```
ronaldfw, line 400: Nit: Add a comment that this is not really a progress but that we are using this to visualize the percentage in the cell.

ronaldfw (commented):
This looks super useful!

I do have some nits on the visual style: 
- The rounded corner rectangles used for the "progress" bars don't look consistent with the rest of the UI (the capture windows uses flat and sharp rectangles). Not a big deal and probably something we have to live with for now. 
- The blue color for the bars is very obvious and noisy. I would suggest to make the color more muted. The bars really strongly stand out in the screenshot and draw the eyes towards them immediately.  They should be easy to read / discern, but should not draw attention away from everything else.

ronaldfw:
> This looks super useful!
> 
> I do have some nits on the visual style:
> 
> * The rounded corner rectangles used for the "progress" bars don't look consistent with the rest of the UI (the capture windows uses flat and sharp rectangles). Not a big deal and probably something we have to live with for now.
> * The blue color for the bars is very obvious and noisy. I would suggest to make the color more muted. The bars really strongly stand out in the screenshot and draw the eyes towards them immediately.  They should be easy to read / discern, but should not draw attention away from everything else.

Ah, it seems you've already been discussing this above, sorry I missed that!

ronaldfw:
I prefer the last option, maybe make the blue also less saturated, it still stands out a bit too much.

dpallotti:
> * The rounded corner rectangles used for the "progress" bars don't look consistent with the rest of the UI (the capture windows uses flat and sharp rectangles). Not a big deal and probably something we have to live with for now.

Well, this is how it looks with the standard Qt theme, I would argue it's that the capture view is inconsistent.

> * The blue color for the bars is very obvious and noisy. I would suggest to make the color more muted. The bars really strongly stand out in the screenshot and draw the eyes towards them immediately.  They should be easy to read / discern, but should not draw attention away from everything else.

> I prefer the last option, maybe make the blue also less saturated, it still stands out a bit too much.

Ok, then I'll go this route and change the color a bit more. Or I can change the color to something completely different if you have some other idea in mind.

ronaldfw:
> > * The rounded corner rectangles used for the "progress" bars don't look consistent with the rest of the UI (the capture windows uses flat and sharp rectangles). Not a big deal and probably something we have to live with for now.
> 
> Well, this is how it looks with the standard Qt theme, I would argue that the capture view is inconsistent.

Oh, well, sure :) In any case, it's inconsistent either way (independent of which part is "wrong"). 

> > * The blue color for the bars is very obvious and noisy. I would suggest to make the color more muted. The bars really strongly stand out in the screenshot and draw the eyes towards them immediately.  They should be easy to read / discern, but should not draw attention away from everything else.
> 
> > I prefer the last option, maybe make the blue also less saturated, it still stands out a bit too much.
> 
> Ok, then I'll go this route and tweak the color a bit more. Or I can change the color to something completely different if you have some other idea in mind.

I like the blue, would just make it less obvious (darker, less saturated). We can also tweak this later if we find it to still be distracting.

OrbitQt/TopDownWidget.cpp
dpallotti: Done.

dpallotti:
@ronaldfw What do you think?

Screenshot from 2020-09-09 16-56-15

OrbitQt/TopDownWidget.cpp
```diff
@@ -374,3 +396,50 @@ QVariant TopDownWidget::HookedIdentityProxyModel::data(const QModelIndex& index,
...
+  option_progress_bar.minimum = 0;
+  option_progress_bar.maximum = 100;
+  option_progress_bar.progress = static_cast<int>(round(inclusive_percent));
+
+  option_progress_bar.palette.setColor(QPalette::Base,
+                                       option.palette.color(QPalette::Disabled, QPalette::Base));
+  const QColor bar_color =
+      QColor::fromHsv(option.palette.color(QPalette::Highlight).hue(),
+                      option.palette.color(QPalette::Highlight).saturation(),
+                      option.palette.color(QPalette::Highlight).value() * 2 / 3);
```
pierricgimmig, line 430: maybe define the 2/3 as a constant?

pierricgimmig: nit. would it be possible to cache or compute the color once for all progress bars?

pierricgimmig (approved):
Hi Davide, that looks awesome, thanks!

OrbitQt/TopDownWidget.cpp
dpallotti: > maybe define the 2/3 as a constant?

Done (needs some casting to float and back now, `2 / 3` with `int`s is 0...).

dpallotti: > nit. would it be possible to cache or compute the color once for all progress bars?

I guess, but conceptually I would like to avoid that as `const QStyleOptionViewItem& option` is passed every time (the style could theoretically change). The overhead should be negligible.

dpallotti:
I re-increased the brightness just a tiny bit to improve the black-on-blue contrast.

Screenshot from 2020-09-10 09-44-17

ronaldfw:
LGTM, thanks!

ronaldfw (approved):

PR #1189, opened 2020-09-09
Make CaptureDeserialization a namespace

kisenshi:
First step to make _CaptureDeserializer_ the most independent from _OrbitGl_ as possible

- Remove class to use _capture_deserializer_ namespace instead
- Move private methods to internal namespace
- Provide _TimeGraph_ as an argument
- Make _GenerateCaptureData_ completely independent from _OrbitGl_
    - Remove _TimeGraph_ dependency by providing the _StringManager_ directly as an argument
    - Move code referencing _GOrbitApp_ outside of the method: calls that clean _GOrbitApp_ and sets the visible functions; for which the _CaptureData_ generated after _GenerateCaptureData_ is used.

Test: loading a capture seems to work as before.

Bug: http://b/167200870

### DIFF
--- a/OrbitGl/CaptureDeserializer.cpp
+++ b/OrbitGl/CaptureDeserializer.cpp
@@ -41,6 +41,7 @@ static void FillEventBuffer(const CaptureData& capture_data) {
 }
 
 }  // namespace
+
 namespace capture_deserializer {
 
 ErrorMessageOr<void> Load(const std::string& filename, TimeGraph* time_graph) {
--- a/OrbitGl/CaptureDeserializer.cpp
+++ b/OrbitGl/CaptureDeserializer.cpp
@@ -29,7 +29,7 @@ using orbit_client_protos::FunctionInfo;
 using orbit_client_protos::FunctionStats;
 using orbit_client_protos::TimerInfo;
 
-ErrorMessageOr<void> CaptureDeserializer::Load(const std::string& filename) {
+ErrorMessageOr<void> CaptureDeserializer::Load(const std::string& filename, TimeGraph* time_graph) {
   SCOPE_TIMER_LOG(absl::StrFormat("Loading capture from \"%s\"", filename));
 
   // Binary
@@ -39,7 +39,7 @@ ErrorMessageOr<void> CaptureDeserializer::Load(const std::string& filename) {
     return ErrorMessage("Error opening the file for reading");
   }
 
-  return Load(file);
+  return Load(file, time_graph);
 }
 
 bool CaptureDeserializer::ReadMessage(google::protobuf::Message* message,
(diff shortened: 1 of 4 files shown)


OrbitGl/CaptureDeserializer.cpp
```diff
@@ -39,11 +54,11 @@ ErrorMessageOr<void> CaptureDeserializer::Load(const std::string& filename) {
     return ErrorMessage("Error opening the file for reading");
   }
 
-  return Load(file);
+  return Load(file, time_graph);
 }
 
-bool CaptureDeserializer::ReadMessage(google::protobuf::Message* message,
-                                      google::protobuf::io::CodedInputStream* input) {
+bool ReadMessage(google::protobuf::Message* message,
```
florian-kuebler, line 60: Can this method be also moved to an internal namespace?

OrbitGl/CaptureDeserializer.cpp
```diff
@@ -143,24 +101,80 @@ ErrorMessageOr<void> CaptureDeserializer::Load(std::istream& stream) {
     ERROR("%s", error_message);
     return ErrorMessage(error_message);
   }
-  CaptureData capture_data = GenerateCaptureData(capture_info);
+  time_graph->Clear();
```
florian-kuebler, line 104: Thanks for moving this out of GenerateCaptureData

florian-kuebler (approved):
Thanks a lot for this change. Despite the visibility of the FillEventBuffer helper method, the changes LGTM.

danielfenner (approved):
Thanks for cleaning this up! lgtm

OrbitGl/CaptureDeserializer.cpp
kisenshi: It is called in _CaptureDeserializerFuzzerInfo_ so I didn't move it inside the internal because of that. Do you think I should anyway and make a call to capture_deserializer::internal::Load instead in there?

OrbitGl/CaptureDeserializer.cpp
kisenshi: No problem! :)

OrbitGl/CaptureDeserializer.cpp
kisenshi: Discussed offline. I misunderstood and thought the comment was referring to _Load_ instead of _ReadMessage_. _ReadMessage_ method can, indeed, be moved to internal so I have done so. Thanks!

kisenshi:
I have moved _ReadMessage_ to the internal namespace

florian-kuebler:
Thanks. LGTM.

florian-kuebler (approved):

kisenshi:
Merging now, thanks!

PR #1190, opened 2020-09-09
Make scaling invertible

karupayun:
Before this commit, after zooming-in + zooming-out (both horizontally and
vertically), you did not get the same scale. Now, zoom-in and zoom-out
are inverse.

### DIFF
--- a/OrbitGl/TimeGraph.cpp
+++ b/OrbitGl/TimeGraph.cpp
@@ -167,7 +167,7 @@ void TimeGraph::ZoomTime(float a_ZoomValue, double a_MouseRatio) {
   m_MouseRatio = a_MouseRatio;
 
   static double incrementRatio = 0.1;
-  double scale = a_ZoomValue > 0 ? 1 + incrementRatio : 1 - incrementRatio;
+  double scale = (a_ZoomValue > 0) ? (1 + incrementRatio) : (1 / (1 + incrementRatio));
 
   double CurrentTimeWindowUs = m_MaxTimeUs - m_MinTimeUs;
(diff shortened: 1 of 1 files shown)


OrbitGl/TimeGraph.cpp
```diff
@@ -167,7 +167,7 @@ void TimeGraph::ZoomTime(float a_ZoomValue, double a_MouseRatio) {
   m_MouseRatio = a_MouseRatio;
 
   static double incrementRatio = 0.1;
-  double scale = a_ZoomValue > 0 ? 1 + incrementRatio : 1 - incrementRatio;
+  double scale = a_ZoomValue > 0 ? 1 + incrementRatio : 1 / (1 + incrementRatio);
```
beckerhe, line 170: When you're changing this anyway, could you add parenthesis to the arguments of the ternary operator for all the folks who don't know the operator precedence list by heart. :see_no_evil:

beckerhe (commented):
Thanks Pablo. This is a little bit hard to grasp out of context. Could you add a high-level description to the commit message?

OrbitGl/TimeGraph.cpp
ronaldfw: This is a great suggestion and IMO should just be the default behavior: always add parenthesis to indicate precedence. Harder to make a mistake, easier to read.

beckerhe (approved):
LGTM! Thanks for the explanation!

OrbitGl/TimeGraph.cpp
karupayun: Thanks! I got used to them, but you are right :)

karupayun:
I also edit the PR comment.

PR #1191, opened 2020-09-09
Move automation scripts for e2e tests into github.

danielfenner:
Versioning the scripts became a problem: The nightly test runs the same
script as the release test runs. When something in the script needs to be
updated (to match the current head) we would alter the version that is run
on the release build as well.

Moving the pywinauto scripts into the github repo and releasing them with
Orbit introduces a natural versioning for the scripts.

Bug: b/168096732
Test: N/A


beckerhe (approved):
LGTM!

PR #1192, opened 2020-09-09
Rename right margin

karupayun:
As VerticalMargin attribute is only being used as right margin, and it is a little confusing, this commit renames it.


ronaldfw (approved):
LGTM! This confused me in the beginning, so I think this is a great change.

pierricgimmig (approved):

PR #1193, opened 2020-09-09
Store the tracepoint events in the client

MariaSandru27:
In order for the tracepoint events to recognize the names and categories of the tracepoints, they store a hash key that represents
these fields. The keys are recognized by being separately mapped in an unordered map to the corresponding name and category.


OrbitClientGgp/CMakeLists.txt
```diff
@@ -7,19 +7,20 @@ project(OrbitClientGgp)
...
-target_link_libraries(OrbitClientGgp PUBLIC 
-    OrbitBase
-    OrbitCaptureClient
-    OrbitClientServices
-    OrbitProtos)
+target_link_libraries(OrbitClientGgp PUBLIC
+        OrbitBase
+        OrbitCaptureClient
+        OrbitClientServices
+        OrbitGl
```
kisenshi, line 24: OrbitClientGgp must be independent from OrbitGl so please do not include it, thanks!

MariaSandru27: I understand. Thank you!

OrbitGl/TracepointInfoManager.h
```diff
@@ -0,0 +1,34 @@
...
+#include <optional>
+#include <string>
+
+#include "BlockChain.h"
+#include "absl/container/flat_hash_map.h"
+#include "absl/synchronization/mutex.h"
+#include "capture_data.pb.h"
+#include "tracepoint.pb.h"
+
+class TracepointInfoManager {
```
kisenshi, line 17: As briefly discussed, this should be created outside OrbitGl so the client ggp can make use of it, as it is independent from OrbitGl. Thanks!

kisenshi (changes requested):
Thanks for adding me. I left some comments; Please let me know if you have any questions :)

OrbitGl/TracepointInfoManager.h
MariaSandru27: Thank you so much for pointing that out!

OrbitCore/TracepointInfoManager.cpp
```diff
@@ -0,0 +1,37 @@
+// Copyright (c) 2020 The Orbit Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#include "TracepointInfoManager.h"
+
+bool TracepointInfoManager::AddUniqueTracepointEventInfo(
```
kisenshi, line 7: I don't really have context about what you are doing but just wondering why you return a boolean when this is called in the listener without checking what the method returns.

OrbitClientGgp/CMakeLists.txt
```diff
@@ -7,19 +7,18 @@ project(OrbitClientGgp)
 add_executable(OrbitClientGgp main.cpp)
 
 target_compile_options(OrbitClientGgp PRIVATE
-    ${STRICT_COMPILE_FLAGS})
```
kisenshi, line 10: The CMakeLists.txt is not change at all now  so maybe it is not worth it to include these indentation updates in the commit

kisenshi (approved):

OrbitClientGgp/CMakeLists.txt
MariaSandru27: The identation updates were not intentional but were caused by the formatting in CLion. However, I have found a way to remove the identation updates. Thanks you a lot for taking the time to review this and point this out!

OrbitCore/TracepointInfoManager.cpp
MariaSandru27: This is intended to be useful in the long-term for different checks. If you look at the way it is done for the string manager or the callstacks, the mechanism is similar so, I also wanted to keep it consistent with the rest of the data retrieved from the collector.

MariaSandru27:
> I don't have the context of what you are doing so I can't really say much about the tracepoints updates themselves (althought I left some general comments) so this should be approved by someone who has that context :) However, I took a look at the changes in ClientGgp and they make sense as it is updated the same way to the Orbit App, while keep being independent from OrbitGl. Thanks a lot for updating it as well!

Thanks a lot for taking the time to review my code!;)

OrbitClientGgp/ClientGgp.h
```diff
@@ -40,12 +41,16 @@ class ClientGgp final : public CaptureListener {
...
+  void OnUniqueTracepointInfo(uint64_t key,
+                              orbit_grpc_protos::TracepointInfo tracepoint_info) override;
+  void OnTracepointEvent(orbit_client_protos::TracepointEventInfo tracepoint_event_info) override;
 
  private:
   ClientGgpOptions options_;
   std::shared_ptr<grpc::Channel> grpc_channel_;
   std::shared_ptr<Process> target_process_;
   std::shared_ptr<StringManager> string_manager_;
+  std::shared_ptr<TracepointInfoManager> tracepoint_info_manager_;
```
danielfenner, line 53: Is there a good reason to have TracepointInfoManager stored in a shared_ptr and not an unique_ptr?

This confused me at first (I thought OrbitApp and ClientGgp are sharing the same instance - which is not the case).

I see that others are using shared_ptr as well. So please keep as is if the shared ownership is intended and needed further down the road.

danielfenner (approved):
lgtm - please have a look at the comments. If you decide to keep this as is I think this is good to go.

danielfenner:
The failed check is a flaky test that should be fixed by now. So it is save to ignore (or to push again to have it pass).

PR #1194, opened 2020-09-09
Fix symbol loading with --local flag

antonrohr:
This moves some things around to not attempt symbol loading from remote when not connected to a remote. 

While I was at it, I also added some more tests and changed the error message appropriately (they are on the rather verbose side).


pierricgimmig (approved):
LGTM, but there is an error on the Windows build.

PR #1195, opened 2020-09-10
Disable selection tabs on empty selections & tweak tab-switching logic

dpallotti:
#### Disable selection-related tabs when the selection is empty
Also, if the "Sampling (selection)" tab was the current tab, switch to the
global "Sampling" tab. If the "Top-Down (selection)" tab was the current tab,
switch to the global "Top-Down" tab.

Bug: http://b/168098607
#### Tweak tab-switching logic on selection
When making a (new) selection, don't switch to the "Sampling (selection)" tab if
one of the two top-down tabs is currently active (this was annoying). Also, if
the "Top-Down" tab is the current one, switch to the "Top-Down (selection)" tab
instead.

### DIFF
--- a/OrbitQt/orbitmainwindow.cpp
+++ b/OrbitQt/orbitmainwindow.cpp
@@ -309,7 +309,14 @@ void OrbitMainWindow::OnNewSelectionReport(DataView* callstack_data_view,
                                            std::shared_ptr<SamplingReport> sampling_report) {
   if (sampling_report->HasSamples()) {
     ui->RightTabWidget->setTabEnabled(ui->RightTabWidget->indexOf(ui->selectionTab), true);
-    ui->RightTabWidget->setCurrentWidget(ui->selectionTab);
+    // This condition and the corresponding one in OnNewSelectionTopDownView need to be
+    // complementary, such that one doesn't cause switching away from or to a tab that the other
+    // method would switch from when such a tab is selected. Otherwise, which tab ends up being
+    // selected would depend on the order in which these two methods are called.
+    if (ui->RightTabWidget->currentWidget() != ui->topDownTab &&
+        ui->RightTabWidget->currentWidget() != ui->selectionTopDownTab) {
+      ui->RightTabWidget->setCurrentWidget(ui->selectionTab);
+    }
   } else {
     // If the selection is empty, if this tab is currently selected switch to the corresponding tab
     // for the entire capture...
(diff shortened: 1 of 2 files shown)


OrbitQt/orbitmainwindow.cpp
```diff
@@ -116,13 +116,18 @@ OrbitMainWindow::OrbitMainWindow(QApplication* a_App, ApplicationOptions&& optio
       [this](DataView* callstack_data_view, std::shared_ptr<SamplingReport> report) {
         this->OnNewSamplingReport(callstack_data_view, std::move(report));
       });
+
+  ui->RightTabWidget->setTabEnabled(ui->RightTabWidget->indexOf(ui->selectionTab), false);
```
reichlfl, line 120: Not really related to this PR and a nitpick, but should we rename the "selectionTab" to sampling like "seletionSamplingTab" in analogy to "topDownTab" and "selectionTopDownTab"?

reichlfl (approved):
What happens when a capture is opened or taken, does this trigger the `OnNewSelectionReport` callbacks and disable the tabs correctly? Can you make sure it's disabled in this case please?

Otherwise LGTM, love the idea on selecting the appropriate `selection` tab depending on the currently visible tab, thanks!

OrbitQt/orbitmainwindow.cpp
```diff
@@ -318,6 +337,20 @@ void OrbitMainWindow::OnNewTopDownView(std::unique_ptr<TopDownView> top_down_vie
 
 void OrbitMainWindow::OnNewSelectionTopDownView(
     std::unique_ptr<TopDownView> selection_top_down_view) {
+  if (selection_top_down_view->child_count() > 0) {
+    ui->RightTabWidget->setTabEnabled(ui->RightTabWidget->indexOf(ui->selectionTopDownTab), true);
+    if (ui->RightTabWidget->currentWidget() == ui->topDownTab) {
```
ronaldfw, line 342: If I see this correctly, then this logic here has to be exactly complementary to the logic in OnNewSelectionReport, or otherwise which tab is selected will depend on the order in which OnNewSelectionReport and OnNewSelectionTopDownView are called (they are both called on a new selection, right?). 

If my understanding is correct, I would add a comment to explain this and point out the connection. O/w someone might "optimize" the code here or above later without fully understanding the implications.

dpallotti:
> What happens when a capture is opened or taken, does this trigger the `OnNewSelectionReport` callbacks and disable the tabs correctly? Can you make sure it's disabled in this case please?

Good point. The current capture is cleared before taking (this I had tried) or loading one, and this also clears the selection, so this works as expected.

OrbitQt/orbitmainwindow.cpp
dpallotti: Gladly done as it was bothering me as well. By the way, we should do some general renaming in `orbitmainwindow.ui` to at least use only one casing style (`centralWidget` vs `RightTabWidget` vs `capture_toolbar`...).

OrbitQt/orbitmainwindow.cpp
dpallotti: If I got your question correctly, yes, this condition and the one in `OnNewSelectionReport` need to work together, I added a (hopefully not too convoluted) comment to both functions.

ronaldfw: Sounds good, yes, that captures what I had in mind.

PR #1196, opened 2020-09-10
Save capture in Ggp client

kisenshi:
Include _GgpClient::SaveCapture_ implementation so the capture is saved in disk. The file name can be provided by the user as an argument and, if not provided, a default name is given. A series of changes were needed for this:

* **Move _OrbitApp::GetCaptureFileName_ to _CaptureSerializer_** This change allows giving the capture file the same default name as the one given in the Orbit App. The method has been moved to the _CaptureSerializer_, in a new namespace called _file_management_. A new method has also been created to include the orbit extension to a file name if this has not been provided.

* **_GgpClient::SaveCapture_ implementation** Serialise and save the data collected while the capture was running. Added logic to determine the name of the file: either the one provided by the user (making sure the orbit extension is included) or the default one. Makes use of the _CaptureSerializer_ methods.

* **_OrbitGgpClient/main_ update** To be able to provide the _file_name_ as an argument when executing the program, include it in the options of the client and make the call to save the capture once it has finished.

Tests: 
* Save a capture in the Orbit app to make sure the name generated for the capture still works. 
* Run _OrbitGgpClient_ with different arguments making sure the capture is saved with the name provided or the default name
* Load the capture generated by the _OrbitGgpClient_ in the Orbit app and make sure the information is there

### DIFF
--- a/OrbitClientGgp/ClientGgp.cpp
+++ b/OrbitClientGgp/ClientGgp.cpp
@@ -112,10 +112,10 @@ bool ClientGgp::SaveCapture() {
   const auto& key_to_string_map = string_manager_->GetKeyToStringMap();
   std::string file_name = options_.capture_file_name;
   if (file_name.empty()) {
-    file_name = capture_serializer::file_management::GetCaptureFileName(capture_data_);
+    file_name = capture_serializer::GetCaptureFileName(capture_data_);
   } else {
     // Make sure the file is saved with orbit extension
-    capture_serializer::file_management::IncludeOrbitExtensionInFile(file_name);
+    capture_serializer::IncludeOrbitExtensionInFile(file_name);
   }
 
   ErrorMessageOr<void> result = capture_serializer::Save(
--- a/OrbitClientGgp/ClientGgp.cpp
+++ b/OrbitClientGgp/ClientGgp.cpp
@@ -15,6 +15,7 @@
 #include "OrbitBase/Logging.h"
 #include "OrbitBase/Result.h"
 #include "OrbitCaptureClient/CaptureClient.h"
+#include "OrbitClientModel/CaptureSerializer.h"
 #include "OrbitClientServices/ProcessManager.h"
 #include "SamplingProfiler.h"
 #include "StringManager.h"
@@ -104,6 +105,26 @@ bool ClientGgp::RequestStartCapture(ThreadPool* thread_pool) {
 bool ClientGgp::StopCapture() {
   LOG("Request to stop capture");
   return capture_client_->StopCapture();
+}
+
+bool ClientGgp::SaveCapture() {
+  LOG("Saving capture");
+  const auto& key_to_string_map = string_manager_->GetKeyToStringMap();
+  std::string file_name = options_.capture_file_name;
+  if (file_name.empty()) {
+    file_name = capture_serializer::file_management::GetCaptureFileName(capture_data_);
+  } else {
+    // Make sure the file is saved with orbit extension
+    capture_serializer::file_management::IncludeOrbitExtensionInFile(file_name);
+  }
+
+  ErrorMessageOr<void> result = capture_serializer::Save(
+      file_name, capture_data_, key_to_string_map, timer_infos_.begin(), timer_infos_.end());
+  if (result.has_error()) {
+    ERROR("Could not save the capture: %s", result.error().message());
+    return false;
+  }
+  return true;
 }
 
 ErrorMessageOr<std::shared_ptr<Process>> ClientGgp::GetOrbitProcessByPid(int32_t pid) {
--- a/OrbitClientGgp/ClientGgp.h
+++ b/OrbitClientGgp/ClientGgp.h
@@ -26,7 +26,7 @@ class ClientGgp final : public CaptureListener {
   bool InitClient();
   bool RequestStartCapture(ThreadPool* thread_pool);
   bool StopCapture();
-  void SaveCapture();
+  bool SaveCapture();
 
   // CaptureListener implementation
   void OnCaptureStarted(
--- a/OrbitClientGgp/main.cpp
+++ b/OrbitClientGgp/main.cpp
@@ -14,6 +14,7 @@ ABSL_FLAG(int32_t, pid, 0, "pid to capture");
 ABSL_FLAG(uint32_t, capture_length, 10, "duration of capture in seconds");
 ABSL_FLAG(std::vector<std::string>, functions, {},
           "Comma-separated list of functions to hook to the capture");
+ABSL_FLAG(std::string, file_name, "", "Set file name for saving the capture.");
 ABSL_FLAG(uint16_t, sampling_rate, 1000, "Frequency of callstack sampling in samples per second");
 ABSL_FLAG(bool, frame_pointer_unwinding, false, "Use frame pointers for unwinding");
 
@@ -29,6 +30,7 @@ int main(int argc, char** argv) {
   options.grpc_server_address = absl::StrFormat("127.0.0.1:%d", grpc_port);
   options.capture_pid = absl::GetFlag(FLAGS_pid);
   options.capture_functions = absl::GetFlag(FLAGS_functions);
+  options.capture_file_name = absl::GetFlag(FLAGS_file_name);
 
   ClientGgp client_ggp(std::move(options));
   if (!client_ggp.InitClient()) {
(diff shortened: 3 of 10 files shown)


OrbitClientModel/include/OrbitClientModel/CaptureSerializer.h
```diff
@@ -26,9 +26,18 @@ ErrorMessageOr<void> Save(const std::string& filename, const CaptureData& captur
...
+std::string GetCaptureFileName(const CaptureData& capture_data);
+
+void IncludeOrbitExtensionInFile(std::string& file_name);
+
+}  // namespace file_management
+
 namespace internal {
 
 inline const std::string kRequiredCaptureVersion = "1.52";
+inline const std::string kFileOrbitExtension = ".orbit";
```
danielfenner, line 40: If this is only used in the .cpp I would move it to an unnamed namespace there.

OrbitClientModel/include/OrbitClientModel/CaptureSerializer.h
```diff
@@ -26,9 +26,18 @@ ErrorMessageOr<void> Save(const std::string& filename, const CaptureData& captur
 void WriteMessage(const google::protobuf::Message* message,
                   google::protobuf::io::CodedOutputStream* output);
 
+namespace file_management {
```
danielfenner, line 29: I'd probably put the two functions below in the capture_serializer namespace. Unless I miss the reason for putting them in a nested namespace.

OrbitClientGgp/ClientGgp.h
```diff
@@ -26,7 +26,7 @@ class ClientGgp final : public CaptureListener {
   bool InitClient();
   bool RequestStartCapture(ThreadPool* thread_pool);
   bool StopCapture();
-  void SaveCapture();
+  bool SaveCapture();
```
danielfenner, line 29: Consider to return a ErrorMessageOr (since that is already done this way in capture_serializer::Save).

kisenshi: I did but I realised the rest of methods called in ClientGgp (InitClient, RequestStartCapture and StopCapture) handle the errors themselves and then just return a boolean so I thought it would make sense to have the same kind of return for all

OrbitClientModel/include/OrbitClientModel/CaptureSerializer.h
kisenshi: Good point, will move it, thanks!

OrbitClientModel/include/OrbitClientModel/CaptureSerializer.h
kisenshi: I was not sure which of the two options was better so I decided to put it in there and see if there were any comments about it in the review :D I will move it to its own namespace, thanks!

kisenshi: Ok, I re-read this and I am updating the comment. I wanted to make a distinction about being part of file_management because these methods do not really serialize the data and it would be confusing, but they are related to the data serialization. What it might make sense to do instead of a nested namespace is moving the namespace below so it is independent but also part of CaptureSerializer, would that make sense?

OrbitClientModel/include/OrbitClientModel/CaptureSerializer.h
kisenshi: Done!

OrbitClientModel/include/OrbitClientModel/CaptureSerializer.h
kisenshi: I am thinking that it might also make sense to rename it to sth like _orbit_file_

OrbitClientGgp/ClientGgp.h
danielfenner: then let's stick with that

OrbitClientModel/include/OrbitClientModel/CaptureSerializer.h
danielfenner: capture_serializer contains stuff that is related to serialization. The functions have proper names - so I don't think it is confusing at all. 
Having many nested namespaces unnecessarily complicates names (capture_serializer::file_management::IncludeOrbitExtensionInFile). As a general guidance: Often a namespace contains everything that is in a directory. Sometimes everything that is in a project. The reason why the stuff here is in a distinct namespace is that "Save" is a very generic name prone to collisions (and the fact that it used to be a class). It would have been completly fine to rename "Save" to "SaveCaptureToFile" and leave it in the top level namespace.
 
Another way of thinking of this: Namespaces are not so much used for naming things but to avoid collisions. (Ahh, that is not entirely true. But I think the direction is okay.)

kisenshi: Ok, got it, thanks for the explanation :) Will update the code. Thanks!

kisenshi: Done!

kisenshi:
I removed the nested namespace, thanks!

danielfenner (approved):
lgtm!

OrbitClientGgp/ClientGgp.cpp
```diff
@@ -105,6 +106,26 @@ bool ClientGgp::StopCapture() {
   return capture_client_->StopCapture();
 }
 
+bool ClientGgp::SaveCapture() {
+  LOG("Saving capture");
+  const auto& key_to_string_map = string_manager_->GetKeyToStringMap();
+  std::string file_name = options_.capture_file_name;
```
ronaldfw, line 112: This would create a copy of the string, which I don't think you need. You should be able to use std::string_view here instead.

OrbitClientGgp/main.cpp
```diff
@@ -14,6 +14,7 @@ ABSL_FLAG(int32_t, pid, 0, "pid to capture");
 ABSL_FLAG(uint32_t, capture_length, 10, "duration of capture in seconds");
 ABSL_FLAG(std::vector<std::string>, functions, {},
           "Comma-separated list of functions to hook to the capture");
+ABSL_FLAG(std::string, file_name, "", "Set file name for saving the capture.");
```
ronaldfw, line 17: The "set" in the comment could suggest that this is a sticky setting. I would simply say:

"File name used for saving the capture."

OrbitClientModel/CaptureSerializer.cpp
```diff
@@ -25,6 +26,10 @@ using orbit_client_protos::FunctionInfo;
 using orbit_client_protos::FunctionStats;
 using orbit_client_protos::TimerInfo;
 
+namespace {
+inline const std::string kFileOrbitExtension = ".orbit";
```
ronaldfw, line 30: I think the recommended way for this would be:
inline constexpr std::string_view kFileOrbitExtension = ".orbit";

OrbitClientModel/CaptureSerializer.cpp
```diff
@@ -34,6 +39,23 @@ void WriteMessage(const google::protobuf::Message* message,
   message->SerializeToCodedStream(output);
 }
 
+std::string GetCaptureFileName(const CaptureData& capture_data) {
+  time_t timestamp = std::chrono::system_clock::to_time_t(capture_data.capture_start_time());
+  std::string result;
+  result.append(Path::StripExtension(capture_data.process_name()));
```
ronaldfw, line 45: Not a large difference, but a little more concise with absl::StrCat:
`std::string result = absl::StrCat(Path::StripExtension(...), "_", OrbitUtils::FormatTime(timestamp));`

ronaldfw (approved):
LGTM, just some minor comments!

OrbitClientGgp/main.cpp
kisenshi: That makes sense, I updated it. Thanks!

OrbitClientModel/CaptureSerializer.cpp
kisenshi: Done!

OrbitClientModel/CaptureSerializer.cpp
kisenshi: Done, thanks for the suggestion!

OrbitClientGgp/ClientGgp.cpp
kisenshi: I tried doing this but I keep getting compilation erros, should I update _capture_serializer::IncludeOrbitExtensionInFile_ to receive a _string_view_ as well and the _const str_ in _capture_serializer::Save(_ to a _string_view_? I read the documentation and I am having some difficult to understand the difference between str, str_view and const str and in which cases I should use each of them

ronaldfw: As discussed in person, I was wrong here. You actually need the copy of the string because you are modifying it below.

ronaldfw (approved):
LGTM

kisenshi:
Thanks, merging this now :)

PR #1197, opened 2020-09-10
Map the tracepoint events to the thread ids and time

MariaSandru27:
For every thread id we monitor which tracepoint events have been triggered and at what time. For that, a red-black tree has been utilized (std::map) that has as the key the thread id and as the value another red-black three to map the pair time and tracepoint event. A sorted associative container has been chosen such that the tracepoint events are sorted: for every thread id found in an non-descending order, the events are found in a non-descending order of time.

Every occured tracepoint event not only is it map to the corresponding thread on which it has been activated but also, it is stored in an universal entry dedicated to all the thread ids (SamplingProfiler::kAllThreadsFakeTid).

At the end, the tracepoint event is registered, meaning that the upper and lower bound for the timeline on which the events happen can be updated.


OrbitCore/TracepointEventBuffer.h
```diff
@@ -0,0 +1,58 @@
...
+  }
+  std::map<ThreadID, std::map<uint64_t, orbit_client_protos::TracepointEventInfo> >&
+  GetTracepoints() {
+    return tracepoint_events_;
+  }
+  Mutex& GetMutex() { return m_Mutex; }
+  std::vector<orbit_client_protos::TracepointEventInfo> GetTracepointEvents(
+      uint64_t time_begin, uint64_t time_end,
+      ThreadID thread_id = SamplingProfiler::kAllThreadsFakeTid);
+  uint64_t GetMaxTime() const { return m_MaxTime; }
```
ronaldfw, line 32: `max_time()`

OrbitCore/TracepointEventBuffer.h
```diff
@@ -0,0 +1,58 @@
...
+  uint64_t GetMinTime() const { return m_MinTime; }
```
ronaldfw, line 33: `min_time()`

OrbitCore/TracepointEventBuffer.h
```diff
@@ -0,0 +1,58 @@
...
+  void Reset() {
+    tracepoint_events_.clear();
+    m_MinTime = LLONG_MAX;
+    m_MaxTime = 0;
+  }
+  std::map<ThreadID, std::map<uint64_t, orbit_client_protos::TracepointEventInfo> >&
+  GetTracepoints() {
+    return tracepoint_events_;
+  }
+  Mutex& GetMutex() { return m_Mutex; }
```
danielfenner, line 28: Exporting the mutex seems a bit weird. Wild guess: Someone uses GetTracepoints and GetMutex to get access to the private data of the class and modifies it from the outside. Consider providing member functions for this instead (well, if I am right about that...).

OrbitCore/TracepointEventBuffer.h
```diff
@@ -0,0 +1,58 @@
...
+
+  void RegisterTime(uint64_t a_Time) {
+    if (a_Time > m_MaxTime) m_MaxTime = a_Time;
+    if (a_Time > 0 && a_Time < m_MinTime) m_MinTime = a_Time;
+  }
+
+  void AddTracepointEvent(uint64_t time, uint64_t hash, ThreadID thread_id);
+
+ private:
+  Mutex m_Mutex;
```
ronaldfw, line 51: Consider making the mutex `mutable`, then you can mark functions that use the mutex but don't change anything else as `const`. E.g. `HasEvent()` can be made `const`.

OrbitCore/TracepointEventBuffer.h
```diff
@@ -0,0 +1,58 @@
...
+      uint64_t time_begin, uint64_t time_end,
+      ThreadID thread_id = SamplingProfiler::kAllThreadsFakeTid);
+  uint64_t GetMaxTime() const { return m_MaxTime; }
+  uint64_t GetMinTime() const { return m_MinTime; }
+  bool HasEvent() {
+    ScopeLock lock(m_Mutex);
+    return !tracepoint_events_.empty();
+  }
+
+#ifdef __linux__
```
ronaldfw, line 39: Why is this needed?

OrbitCore/TracepointEventBuffer.h
```diff
@@ -0,0 +1,58 @@
...
+  size_t GetNumEvents() const;
+#endif
+
+  void RegisterTime(uint64_t a_Time) {
```
danielfenner, line 43: Can RegisterTime be made private? It should only be used by AddTracepointEvent I guess.

OrbitCore/TracepointEventBuffer.h
```diff
@@ -0,0 +1,58 @@
...
+#include <set>
+
+#include "BlockChain.h"
+#include "SamplingProfiler.h"
+#include "Threading.h"
+#include "capture_data.pb.h"
+
+class TracepointEventBuffer {
+ public:
+  TracepointEventBuffer() : m_MaxTime(0), m_MinTime(LLONG_MAX) {}
```
danielfenner, line 17: `LLONG_MAX` -> `std::numeric_limits<uint64_t>::max()` to be sure the type fits. One can't be sure "long long"  is the same as "uint64_t"

Add `#include <limits>` on the top.

OrbitCore/TracepointEventTracer.h
```diff
@@ -0,0 +1,17 @@
...
+
+class TracepointEventTracer {
+ public:
+  TracepointEventBuffer& GetTracepointEventBuffer() { return m_TracepointEventBuffer; }
+
+ private:
+  TracepointEventBuffer m_TracepointEventBuffer;
+};
+
+extern TracepointEventTracer GTracepointEventTracer;
```
danielfenner, line 17: Can we get around this global variable?

OrbitCore/TracepointEventBuffer.cpp
```diff
@@ -0,0 +1,64 @@
...
+  orbit_client_protos::TracepointEventInfo event_all_threads;
+  event_all_threads.set_time(time);
+  event_all_threads.set_tracepoint_info_key(hash);
+  event_all_threads.set_tid(SamplingProfiler::kAllThreadsFakeTid);
+  event_map_all_threads[time] = event_all_threads;
+
+  RegisterTime(time);
+}
+
+#ifdef __linux
```
danielfenner, line 55: __linux -> __linux__
But I'd prefer to remove the ifdef entirely as commented below.

dpallotti:
I see that this is based on the model of `EventTracer`/`EventBuffer`, but that's something really old and that we really really need to get rid of&mdash;it should not be taken as an example. I would probably look into `CaptureData`, it should still be accessible everywhere with `GOrbitApp->GetCaptureData()`. Maybe ask @florian-kuebler for advice as he recently introduced it.
(If it ends up taking too long or being too complicated and we really need to go this route, let's at least avoid having two classes, `TracepointEventTracer` doesn't do anything.)

MariaSandru27:
> I see that this is based on the model of `EventTracer`/`EventBuffer`, but that's something really old and that we really really need to get rid of—it should not be taken as an example. I would probably look into `CaptureData`, it should still be accessible everywhere with `GOrbitApp->GetCaptureData()`. Maybe ask @florian-kuebler for advice as he recently introduced it.
> (If it ends up taking too long or being too complicated and we really need to go this route, let's at least avoid having two classes, `TracepointEventTracer` doesn't do anything.)

I will also remove TracepointEventTracer.

OrbitCore/TracepointEventBuffer.cpp
```diff
@@ -0,0 +1,64 @@
...
+// found in the LICENSE file.
+
+#include "TracepointEventBuffer.h"
+
+#include "SamplingProfiler.h"
+#include "TracepointEventTracer.h"
+
+using orbit_client_protos::TracepointEventInfo;
+
+TracepointEventTracer GTracepointEventTracer;
```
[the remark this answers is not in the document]
MariaSandru27: It is not necessary to include new globals. I will modify the code. Thank you!

OrbitCore/TracepointEventBuffer.h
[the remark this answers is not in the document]
ronaldfw: Actually drop the "a_" as well.

MariaSandru27: You are right! Thanks you!

OrbitCore/TracepointEventBuffer.h
```diff
@@ -0,0 +1,58 @@
...
+#ifdef __linux__
+  size_t GetNumEvents() const;
+#endif
+
+  void RegisterTime(uint64_t a_Time) {
+    if (a_Time > m_MaxTime) m_MaxTime = a_Time;
+    if (a_Time > 0 && a_Time < m_MinTime) m_MinTime = a_Time;
+  }
+
+  void AddTracepointEvent(uint64_t time, uint64_t hash, ThreadID thread_id);
```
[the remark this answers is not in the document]
MariaSandru27: Of course! Thank you!

OrbitCore/TracepointEventBuffer.h
```diff
@@ -0,0 +1,58 @@
...
+
+ private:
+  Mutex m_Mutex;
+  std::atomic<uint64_t> m_MaxTime;
+  std::atomic<uint64_t> m_MinTime;
+  std::map<ThreadID, std::map<uint64_t, orbit_client_protos::TracepointEventInfo> >
```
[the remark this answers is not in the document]
MariaSandru27: The reason was such that the elements would be places in order. However, the equivalent of std::map would be absl::btree_map. Both have the elements sorted by key but the first one uses a red-black tree and the second uses a B-tree and is optimized.

OrbitCore/TracepointEventBuffer.h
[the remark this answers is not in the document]
MariaSandru27: Of course! Thank you!

OrbitCore/TracepointEventBuffer.h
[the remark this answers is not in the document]
ronaldfw: I can see why you may need to order by timestamp (though I'm not sure about that either), but I'm not sure why you need the order by thread id. 

I can take a closer look at how this is currently used and done in EventTrack, but I *think* that just iterates (per thread) over all events for drawing anyway (it's not yet optimized in the same way that TimerTrack is). If that is the case, you don't need either order.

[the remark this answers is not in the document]
MariaSandru27: I had a quick chat with @danielfenner and the current version of abseil is too old to support absl::btree_map. However, the container still needs to sort the keys  so we thought about having for now std::map.

OrbitCore/TracepointEventBuffer.cpp
```diff
@@ -0,0 +1,84 @@
...
+  event_all_threads.set_time(time);
+  event_all_threads.set_tracepoint_info_key(tracepoint_hash);
+  event_all_threads.set_tid(SamplingProfiler::kAllThreadsFakeTid);
+  event_map_all_threads[time] = event_all_threads;
+
+  RegisterTime(time);
+}
+
+size_t TracepointEventBuffer::GetNumEvents() const {
+  size_t num_events = 0;
```
danielfenner, line 54: I would expect a 
`ScopeLock lock(mutex_);`
here and also in Reset() below.

MariaSandru27: Yes, you are right. Thank you so much for pointing that out!

OrbitCore/TracepointEventBuffer.h
```diff
@@ -0,0 +1,47 @@
...
+
+  void AddTracepointEvent(uint64_t time, uint64_t tracepoint_hash, ThreadID thread_id);
+
+  std::map<ThreadID, std::map<uint64_t, orbit_client_protos::TracepointEventInfo> >&
+  tracepoint_events();
+
+  uint64_t max_time() const;
+  uint64_t min_time() const;
+
+  const bool HasEvent();
```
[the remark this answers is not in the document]
MariaSandru27: I think I could keep the 'before' const and also put one after. After all, the returned data is not going to be later modified and the data in the function 'HasEvent' is not going to be modified either. Thanks a lot!

OrbitCore/TracepointEventBuffer.h
```diff
@@ -0,0 +1,47 @@
...
+#include <set>
+
+#include "BlockChain.h"
+#include "SamplingProfiler.h"
+#include "Threading.h"
+#include "capture_data.pb.h"
+
+class TracepointEventBuffer {
+ public:
+  TracepointEventBuffer() : max_time_(0), min_time_(std::numeric_limits<long long int>::max()) {}
```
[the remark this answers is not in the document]
MariaSandru27: Thanks so much for explaining! I did not know that.

OrbitCore/TracepointEventBuffer.h
MariaSandru27: The data from TracepiintEventBuffer is read in another function and, in order for the it to not be corrupted that mutex is used.

OrbitCore/TracepointEventBuffer.h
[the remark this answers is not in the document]
MariaSandru27: Yes, thank you!

OrbitCore/TracepointEventBuffer.h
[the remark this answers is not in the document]
dpallotti: My understanding (correct me) was that the tracepoint events would be shown in their own track, not in thread tracks (most of these events would probably have kernel-process pids anyway), so you might not need to group by thread at all.

dpallotti:
> It is indeed based on EventTracer/EventBuffer because in those file the mapping from the thread ids/timestamps to the events is actually done. The data given by 'GOrbitApp->GetCaptureData()' only gives the callstack events without any hint towards an ordering of the thread ids or of the timestamps. I already have a way to store the tracepoint events and this time I wanted a way to keep track of the threads and timestamps. However, I can see that the style used was 'old'. I will change to fit the new code style. If you have any suggestions please let me know.

I see that you moved from a global to storing a `TracepointEventBuffer` in `OrbitApp` which is an improvement. I still think that it should be moved to `CaptureData`. For example there is a `CallstackData` in `CaptureData`. You are right that it doesn't map from tids and timestamps, but don't worry about that (it's actually one reason why `EventBuffer` is still around), add what you need to your class for what your goal is. For example you can have a `TracepointData` there, doing what you are doing with `TracepointEventBuffer`, but without worrying about the parallelism with `EventBuffer` (e.g., you probably don't need to map by thread, you don't need `Reset` as you can just create a new object when the capture is cleared, etc.).

OrbitCore/TracepointEventBuffer.cpp
```diff
@@ -0,0 +1,96 @@
...
+
+#include "TracepointEventBuffer.h"
+
+#include "SamplingProfiler.h"
+
+using orbit_client_protos::TracepointEventInfo;
+
+std::vector<TracepointEventInfo> TracepointEventBuffer::GetTracepointEvents(
+    uint64_t time_begin, uint64_t time_end, ThreadID thread_id /*= kAllThreadsFakeTid*/) {
+  std::vector<TracepointEventInfo> tracepoint_events;
```
florian-kuebler, line 13: I guess there should be a `ScopeLock lock(mutex_)` here.

OrbitCore/TracepointEventBuffer.cpp
```diff
@@ -0,0 +1,96 @@
...
+  for (auto& pair : tracepoint_events_) {
+    const ThreadID tracepoint_thread_id = pair.first;
+    std::map<uint64_t, TracepointEventInfo>& tracepoints = pair.second;
+
+    CHECK(thread_id == SamplingProfiler::kAllThreadsFakeTid || tracepoint_thread_id == thread_id);
+    for (auto it = tracepoints.lower_bound(time_begin); it != tracepoints.end(); ++it) {
+      uint64_t time = it->first;
+      if (time < time_end) {
```
florian-kuebler, line 21: The events in tracepoints are sorted, right? So you should break out of that loop as soon as you see an event with a timestamp greater than the time_end.

OrbitCore/TracepointEventBuffer.cpp
```diff
@@ -0,0 +1,96 @@
...
+void TracepointEventBuffer::AddTracepointEvent(uint64_t time, uint64_t tracepoint_hash,
+                                               ThreadID thread_id) {
+  ScopeLock lock(mutex_);
+  std::map<uint64_t, orbit_client_protos::TracepointEventInfo>& event_map =
+      tracepoint_events_[thread_id];
+  orbit_client_protos::TracepointEventInfo event;
+  event.set_time(time);
+  event.set_tracepoint_info_key(tracepoint_hash);
+  event.set_tid(thread_id);
+  event_map[time] = event;
```
florian-kuebler, line 39: I would recommend to `std::move` the event here.

OrbitCore/TracepointEventBuffer.cpp
```diff
@@ -0,0 +1,96 @@
...
+
+  // Add all tracepoint events to "all threads".
+  std::map<uint64_t, orbit_client_protos::TracepointEventInfo>& event_map_all_threads =
+      tracepoint_events_[SamplingProfiler::kAllThreadsFakeTid];
+  orbit_client_protos::TracepointEventInfo event_all_threads;
+  event_all_threads.set_time(time);
+  event_all_threads.set_tracepoint_info_key(tracepoint_hash);
+  event_all_threads.set_tid(SamplingProfiler::kAllThreadsFakeTid);
+  event_map_all_threads[time] = event_all_threads;
```
florian-kuebler, line 48: Also `move` this event.

OrbitCore/TracepointEventBuffer.cpp
```diff
@@ -0,0 +1,96 @@
...
+  ScopeLock lock(mutex_);
+  size_t num_events = 0;
+  for (const auto& pair : tracepoint_events_) {
+    num_events += pair.second.size();
+  }
+
+  return num_events;
+}
+
+bool TracepointEventBuffer::HasEvent() const {
```
florian-kuebler, line 63: `EventBuffer::HasEvent` is used to check whether the trace is empty, I guess at that point you also want to check that there is no tracepoint event.

OrbitCore/TracepointEventBuffer.h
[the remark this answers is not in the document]
florian-kuebler: @ronaldfw I believe that for Sampling Events, the `GetEvents` with a begin/end timestamp assumes that the inner map is sorted. However, about the outer map you should be right. Also I think that ordering should be stated more explicit on the type-level. However, I am not sure about container classes in C++ to be used for sorted data.

OrbitCore/TracepointEventBuffer.h
```diff
@@ -0,0 +1,49 @@
...
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#ifndef ORBIT_CORE_TRACEPOINT_EVENT_BUFFER_H_
+#define ORBIT_CORE_TRACEPOINT_EVENT_BUFFER_H_
+
+#include <limits>
+#include <set>
+
+#include "BlockChain.h"
```
florian-kuebler, line 11: Does it use BlockChain, SamplingProfiler or Threading?

OrbitCore/TracepointEventBuffer.h
```diff
@@ -0,0 +1,49 @@
...
+
+class TracepointEventBuffer {
+ public:
+  TracepointEventBuffer() : max_time_(0), min_time_(std::numeric_limits<uint64_t>::max()) {}
+
+  std::vector<orbit_client_protos::TracepointEventInfo> GetTracepointEvents(
+      uint64_t time_begin, uint64_t time_end,
+      ThreadID thread_id = SamplingProfiler::kAllThreadsFakeTid);
+
+  void AddTracepointEvent(uint64_t time, uint64_t tracepoint_hash, ThreadID thread_id);
```
florian-kuebler, line 24: Please use `int32_t` instead of `ThreadID`.

OrbitCore/TracepointEventBuffer.cpp
```diff
@@ -0,0 +1,96 @@
...
+        tracepoint_events.push_back(it->second);
+      }
+    }
+  }
+
+  return tracepoint_events;
+}
+
+void TracepointEventBuffer::AddTracepointEvent(uint64_t time, uint64_t tracepoint_hash,
+                                               ThreadID thread_id) {
```
florian-kuebler, line 31: Please use `int32_t` instead of `ThreadID`.

OrbitCore/TracepointEventBuffer.cpp
```diff
@@ -0,0 +1,96 @@
...
+// found in the LICENSE file.
+
+#include "TracepointEventBuffer.h"
+
+#include "SamplingProfiler.h"
+
+using orbit_client_protos::TracepointEventInfo;
+
+std::vector<TracepointEventInfo> TracepointEventBuffer::GetTracepointEvents(
+    uint64_t time_begin, uint64_t time_end, ThreadID thread_id /*= kAllThreadsFakeTid*/) {
```
florian-kuebler, line 12: Please use `int32_t` instead of `ThreadID`.

OrbitCore/TracepointEventBuffer.cpp
```diff
@@ -0,0 +1,96 @@
...
+  ScopeLock lock(mutex_);
+  tracepoint_events_.clear();
+  min_time_ = std::numeric_limits<uint64_t>::max();
+  max_time_ = 0;
+}
+
+uint64_t TracepointEventBuffer::max_time() const { return max_time_; }
+uint64_t TracepointEventBuffer::min_time() const { return min_time_; }
+
+std::map<ThreadID, std::map<uint64_t, orbit_client_protos::TracepointEventInfo> >
```
florian-kuebler, line 86: Please use `int32_t` instead of `ThreadID`.

OrbitCore/TracepointEventBuffer.h
```diff
@@ -0,0 +1,49 @@
...
+
+  void Reset();
+
+ private:
+  void RegisterTime(uint64_t time);
+
+  mutable Mutex mutex_;
+  std::atomic<uint64_t> max_time_;
+  std::atomic<uint64_t> min_time_;
+  std::map<ThreadID, std::map<uint64_t, orbit_client_protos::TracepointEventInfo> >
```
florian-kuebler, line 45: Please use `int32_t` instead of `ThreadID`.

florian-kuebler (commented):
Thanks for adding this change. I added some minor comments.

florian-kuebler:
> > It is indeed based on EventTracer/EventBuffer because in those file the mapping from the thread ids/timestamps to the events is actually done. The data given by 'GOrbitApp->GetCaptureData()' only gives the callstack events without any hint towards an ordering of the thread ids or of the timestamps. I already have a way to store the tracepoint events and this time I wanted a way to keep track of the threads and timestamps. However, I can see that the style used was 'old'. I will change to fit the new code style. If you have any suggestions please let me know.
> 
> I see that you moved from a global to storing a `TracepointEventBuffer` in `OrbitApp` which is an improvement. I still think that it should be moved to `CaptureData`. For example there is a `CallstackData` in `CaptureData`. You are right that it doesn't map from tids and timestamps, but don't worry about that (it's actually one reason why `EventBuffer` is still around), add what you need to your class for what your goal is. For example you can have a `TracepointData` there, doing what you are doing with `TracepointEventBuffer`, but without worrying about the parallelism with `EventBuffer` (e.g., you probably don't need to map by thread, you don't need `Reset` as you can just create a new object when the capture is cleared, etc.).

I think the `TracepointEventBuffer` can either live in the `App` or in `CaptureData` for now. There is one issue to be solved anyways, which is synchronization. The original idea of `CaptureData` was that it lives on the capture thread, and being basically invincible to all other threads while capturing, and then make it visible for read only when capturing is done. But as we can select callstacks while capturing, this isn't solved anyways. Currently, this is somewhat save, as the `CallstackData` is synchronized internally. However, we should in general rework synchronization of `CaptureData`. It should provide a mechanism to thread-safe stream data to the app and should be still write only on the capture thread.

`TracepointEventBuffer` would introduce a similar issue. But as this needs to be done anyways, I think it would be fine to move it to `CaptureData` for now.

OrbitCore/TracepointEventBuffer.cpp
```diff
@@ -0,0 +1,96 @@
...
+  }
+  for (const auto& pair : tracepoint_events_) {
+    if (!pair.second.empty()) {
+      return true;
+    }
+  }
+  return false;
+}
+
+void TracepointEventBuffer::Reset() {
```
florian-kuebler, line 76: This can be removed as soon as it is part of CaptureData.

OrbitCore/TracepointEventBuffer.cpp
MariaSandru27: of course! thank you!

OrbitCore/TracepointEventBuffer.h
MariaSandru27: Only SamplingProfiler. Thanks for pointing that out!

OrbitCore/TracepointEventBuffer.h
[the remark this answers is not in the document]
MariaSandru27: @dpallotti as discussed this morning, the thread tracks do not show in their own track but in another event bar.

florian-kuebler (approved):
Thanks for addressing my comments. LGTM. Thanks!

OrbitCore/CaptureData.h
```diff
@@ -27,7 +28,8 @@ class CaptureData {
         selected_functions_{std::move(selected_functions)},
         selected_tracepoints_{std::move(selected_tracepoints)},
         callstack_data_(std::make_unique<CallstackData>()),
-        selection_callstack_data_(std::make_unique<CallstackData>()) {
+        selection_callstack_data_(std::make_unique<CallstackData>()),
+        tracepoint_info_manager_(std::make_unique<TracepointInfoManager>()) {
```
dpallotti, line 32: `TracepointEventBuffer` was the one supposed to end up in `CaptureData`, not `TracepointInfoManager` (or both, but `TracepointEventBuffer` was the object of the discussion).

OrbitCore/TracepointEventBuffer.h
```diff
@@ -0,0 +1,44 @@
...
+
+  std::map<int32_t, std::map<uint64_t, orbit_client_protos::TracepointEventInfo> >
+  tracepoint_events();
+
+  Mutex& mutex();
+
+  uint64_t max_time() const;
+  uint64_t min_time() const;
+
+  bool HasEvent() const;
```
florian-kuebler, line 32: Forgot in my last review:
You should also use that method in the TimeGraph to check for the time span.

OrbitCore/CaptureData.h
```diff
@@ -132,6 +140,21 @@ class CaptureData {
...
+
+  void AddTracepointEvent(orbit_client_protos::TracepointEventInfo tracepoint_event_info) {
+    tracepoint_info_manager_->AddTracepointEvent(std::move(tracepoint_event_info));
+  }
+
+  void AddTracepointEvent(uint64_t time, uint64_t tracepoint_hash, int32_t thread_id){
+    tracepoint_event_buffer_->AddTracepointEvent(time,
+                                                 tracepoint_hash,
+                                                 thread_id);
+  }
```
dpallotti: Do/will you use/need both these methods or is only one of the two used? In the latter case, I would remove the one that is not necessary, let's try to have only necessary methods in this big class.

OrbitCore/TracepointEventBuffer.h
```diff
@@ -0,0 +1,44 @@
...
+#include <set>
+
+#include "SamplingProfiler.h"
+#include "capture_data.pb.h"
+
+class TracepointEventBuffer {
+ public:
+  TracepointEventBuffer() : max_time_(0), min_time_(std::numeric_limits<uint64_t>::max()) {}
+
+  std::vector<orbit_client_protos::TracepointEventInfo> GetTracepointEvents(
```
dpallotti, line 18: If we are following `EventBuffer` here, this should be called `GetTracepointEventsInTimeRangeForThreadId`.

OrbitCore/TracepointEventBuffer.cpp
```diff
@@ -0,0 +1,79 @@
...
+
+std::vector<TracepointEventInfo> TracepointEventBuffer::GetTracepointEvents(
+    uint64_t time_begin, uint64_t time_end, int32_t thread_id /*= kAllThreadsFakeTid*/) {
+  ScopeLock lock(mutex_);
+  std::vector<TracepointEventInfo> tracepoint_events;
+  for (auto& pair : tracepoint_events_) {
+    const int32_t tracepoint_thread_id = pair.first;
+    std::map<uint64_t, TracepointEventInfo>& tracepoints = pair.second;
+
+    CHECK(thread_id == SamplingProfiler::kAllThreadsFakeTid || tracepoint_thread_id == thread_id);
```
dpallotti, line 17: I'm confused about this CHECK. Isn't it normal that `tracepoint_thread_id != thread_id` and you would just skip to the next iteration? Have a look at this loop in `EventBuffer::GetCallstackEventsInTimeRangeForThreadId`. You might want a condition here instead (`if (thread_id == SamplingProfiler::kAllThreadsFakeTid || callstack_thread_id == thread_id)`).

OrbitCore/TracepointEventBuffer.cpp
```diff
@@ -0,0 +1,79 @@
...
+    for (auto it = tracepoints.lower_bound(time_begin); it != tracepoints.end(); ++it) {
+      uint64_t time = it->first;
+      if (time < time_end) {
+        tracepoint_events.push_back(it->second);
+      } else
```
dpallotti, line 22: Add `{}`.

OrbitCore/TracepointEventBuffer.cpp
```diff
@@ -0,0 +1,79 @@
...
+        return tracepoint_events;
```
dpallotti, line 23: Shouldn't this be a `break` and then you continue with the next iteration of the outer loop? If `thread_id == SamplingProfiler::kAllThreadsFakeTid`, you want to always go through all elements in `tracepoint_events_`.

OrbitCore/CaptureData.h
MariaSandru27: I use both of them.

OrbitCore/CaptureData.h
```diff
@@ -132,6 +140,19 @@ class CaptureData {
...
+
+  void AddTracepointEvent(orbit_client_protos::TracepointEventInfo tracepoint_event_info) {
+    tracepoint_info_manager_->AddTracepointEvent(std::move(tracepoint_event_info));
+  }
+
+  void AddTracepointEvent(uint64_t time, uint64_t tracepoint_hash, int32_t thread_id) {
+    tracepoint_event_buffer_->AddTracepointEvent(time, tracepoint_hash, thread_id);
+  }
+
```
dpallotti, line 155: Do/will you use/need both these methods or is only one of the two used? In the latter case, I would remove the one that is not necessary, let's try to have only necessary methods in this big class.

MariaSandru27: Yes, both of them are used are are necessary.

MariaSandru27: One is for mapping the tracepoint event to the hash key and the second is for mapping to the thread is and timestamp.

OrbitCore/TracepointEventBuffer.h
MariaSandru27: That method does something else. 'GetTracepointEvents()' maches the method 'GetCallstacks() const' in EventBuffer.

MariaSandru27: Yes, you are right. Thank you!

OrbitCore/CaptureData.h
dpallotti: If the do different things (I see that one uses `tracepoint_info_manager_` and one uses `tracepoint_event_buffer_`) they should have different names. 

Besides, why do you need to add tracepoint events to both? Do they both store them? Why? Shouldn't they be stored only in `TracepointEventBuffer`?

OrbitCore/TracepointEventBuffer.cpp
MariaSandru27: Yes, you are right. Thank you!

OrbitCore/CaptureData.h
MariaSandru27: The events are added to TracepointInfoManager such that they could be later used for statistics. However, if you this it would be better to delete saving the tracepoint events for now, I can do it.

dpallotti: Duplicating all the events for possible later usage for statistics sounds wrong, I think we should get rid of that and later have a better way to keeps statistics.

danielfenner (approved):
lgtm

OrbitCore/CaptureData.h
MariaSandru27: Done

PR #1198, opened 2020-09-10
Make CaptureDeserializer independed from OrbitGl

florian-kuebler:
CaptureDesirializer now uses CaptureListener methods in order
to process the data from the capture to be loaded instead.
Prior to this, capture data was loaded and directly set at once to the App.
Also, the timers were stored directly on the time graph.

The calls to capture listener must not be done on the main thread.
Therefor, as a side-effect, deserialization does not block the UI
anymore and is somewhat "interactive", i.e. you see events poping
 up on loading.

Follow up: Make the events not appearing in blocks, which requires
 rethinking of the way we store timers.

Known Issue:
When loading a "broken" capture from the startup window
 the following happend before:
 1. An error msg. appears, that and why loading failed. In the background
    Orbit's main UI is already spawned.
 2. After clicking okay on this, the UI closes, and there is another error
    msg. saying "no such file".
 3. After clicking onkay on this, the startup windows shows up again.
Now:
 1. As before.
 2. On click of okay, you are in an empty Orbit and can try to load another
    capture from the capture view, but in order to connect to a gamelet,
    you need to restart Orbit manually.

Bug: http://b/167200870 and http://b/166094073
Test: Compile, Load Capture (also a broken one) from UI and start-up window.

### DIFF
--- a/OrbitClientModel/CaptureDeserializer.cpp
+++ b/OrbitClientModel/CaptureDeserializer.cpp
@@ -25,7 +25,8 @@ using orbit_client_protos::TimerInfo;
 
 namespace capture_deserializer {
 
-ErrorMessageOr<void> Load(const std::string& filename, CaptureListener* capture_listener) {
+ErrorMessageOr<void> Load(const std::string& filename, CaptureListener* capture_listener,
+                          bool* cancelation_requested) {
   SCOPE_TIMER_LOG(absl::StrFormat("Loading capture from \"%s\"", filename));
 
   // Binary
@@ -35,10 +36,11 @@ ErrorMessageOr<void> Load(const std::string& filename, CaptureListener* capture_
     return ErrorMessage("Error opening the file for reading");
   }
 
-  return Load(file, capture_listener);
+  return Load(file, capture_listener, cancelation_requested);
 }
 
-ErrorMessageOr<void> Load(std::istream& stream, CaptureListener* capture_listener) {
+ErrorMessageOr<void> Load(std::istream& stream, CaptureListener* capture_listener,
+                          bool* cancelation_requested) {
   google::protobuf::io::IstreamInputStream input_stream(&stream);
   google::protobuf::io::CodedInputStream coded_input(&input_stream);
 
@@ -67,7 +69,7 @@ ErrorMessageOr<void> Load(std::istream& stream, CaptureListener* capture_listene
     return ErrorMessage(error_message);
   }
 
-  internal::LoadCaptureInfo(capture_info, capture_listener, &coded_input);
+  internal::LoadCaptureInfo(capture_info, capture_listener, &coded_input, cancelation_requested);
 
   return outcome::success();
 }
@@ -91,8 +93,10 @@ bool ReadMessage(google::protobuf::Message* message,
 }
 
 void LoadCaptureInfo(const CaptureInfo& capture_info, CaptureListener* capture_listener,
-                     google::protobuf::io::CodedInputStream* coded_input) {
+                     google::protobuf::io::CodedInputStream* coded_input,
+                     bool* cancelation_requested) {
   CHECK(capture_listener != nullptr);
+
   absl::flat_hash_map<uint64_t, orbit_client_protos::FunctionInfo> selected_functions;
   for (const auto& function : capture_info.selected_functions()) {
     uint64_t address = FunctionUtils::GetAbsoluteAddress(function);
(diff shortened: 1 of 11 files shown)


OrbitClientModel/CaptureDeserializer.cpp
```diff
@@ -134,45 +90,47 @@ bool ReadMessage(google::protobuf::Message* message,
   return true;
 }
 
-CaptureData GenerateCaptureData(const CaptureInfo& capture_info, StringManager* string_manager) {
+void DoLoadCaptureInfo(const CaptureInfo& capture_info, CaptureListener* capture_listener,
```
pierricgimmig, line 93: nit. Can "LoadCaptureInfo" be used instead of "DoLoadCaptureInfo"? Or is the "Do" meaningful here?

OrbitGl/App.cpp
```diff
@@ -170,12 +170,11 @@ void OrbitApp::OnCaptureComplete() {
 
 void OrbitApp::OnTimer(const TimerInfo& timer_info) {
   if (timer_info.function_address() > 0) {
-    FunctionInfo* func =
-        GetCaptureData().process()->GetFunctionFromAddress(timer_info.function_address());
-    CHECK(func != nullptr);
+    const FunctionInfo func =
```
pierricgimmig, line 173: Shouldn't this be a const ref?

pierricgimmig (commented):
Hi Florian, thanks for the PR! The code looks good to me apart from the minor comments. I haven't tested the change, I'm a little bit worried that migrating loading from the main thread might cause a race condition somewhere, can the selected_functions for example be accessed on the main thread while they are being populated on the worker thread?

florian-kuebler:
> Hi Florian, thanks for the PR! The code looks good to me apart from the minor comments. I haven't tested the change, I'm a little bit worried that migrating loading from the main thread might cause a race condition somewhere, can the selected_functions for example be accessed on the main thread while they are being populated on the worker thread?

Hi Pierric, Thanks for having a look. I don't think there is a general problem here, as we are essentially doing the same as on profiling (which is also on a different thread and calls the CaptureListener methods from that thread). So from guarantees there shouldn't be any difference. But http://b/166767590 is also there, to make capture data itself more thread-safe. 
In particular selected functions are no problem at all, as they are only written once at OnCaptureStarted, so they can be accessed safely while capturing/loading. 

The only thing that might be different, is that we are not switching into "IsCapturing" state, and thus we might try to load/start another capture, while loading. I am going to address this now, and update the PR accordingly.

florian-kuebler:
I disabled the UI elements (similar to start capture), such that the user can't interfere with loading a capture.

dpallotti:
(You forgot "b/" in the links to the bugs.)

florian-kuebler:
thanks!

OrbitGl/App.cpp
florian-kuebler: done.

OrbitClientModel/CaptureDeserializer.cpp
florian-kuebler: done.

dpallotti:
Is the known issue that you explained in the commit message fixed by this, caused by this, or independent of this?

OrbitGl/App.cpp
```diff
@@ -637,13 +636,26 @@ ErrorMessageOr<void> OrbitApp::OnSaveCapture(const std::string& file_name) {
                                   timers_it_end);
 }
 
-ErrorMessageOr<void> OrbitApp::OnLoadCapture(const std::string& file_name) {
+void OrbitApp::OnLoadCapture(const std::string& file_name) {
+  if (open_capture_callback_) {
+    open_capture_callback_();
+  }
   ClearCapture();
+  GCurrentTimeGraph->GetStringManager()->Clear();
```
dpallotti, line 644: Should the `StringManager` be accessed from `GOrbitApp`, which is its actual owner?

dpallotti (approved):

OrbitGl/App.cpp
florian-kuebler: Done. Thanks

florian-kuebler:
> Is the known issue that you explained in the commit message fixed by this, caused by this, or independent of this?

As discussed offline, this issue is kind of independent but still exists. It is basically rooting from the fact, that we do not load on the main thread anymore, and thus can not determine whether there was a failure or not on UI creation. 

That state, that gets created in this case (not being connected and not having a capture) can also right now occur: Load a capture, then load another broken capture.

I will have a look into this in a separate PR.

florian-kuebler:
As discussed offline, I added a progress bar while loading a capture, such that the user can't interact with the UI while loading. 
I also made the "cancel" button of that progress bar stopping the loading and clearing the capture.

dpallotti (approved):
Looks good otherwise.

OrbitClientModel/CaptureDeserializer.cpp
```diff
@@ -26,25 +21,12 @@ using orbit_client_protos::CallstackInfo;
...
-  }
-}
-
-}  // namespace
-
 namespace capture_deserializer {
 
-ErrorMessageOr<void> Load(const std::string& filename, TimeGraph* time_graph) {
+ErrorMessageOr<void> Load(const std::string& filename, CaptureListener* capture_listener,
+                          bool* cancelation_requested) {
```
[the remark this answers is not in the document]
florian-kuebler: Oh yep, you are right. Also then Clion does not complain anymore :). Fixed that.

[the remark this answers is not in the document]
florian-kuebler: Right. Thanks! Made it an atomic bool.

pierricgimmig:
> As discussed offline, I added a progress bar while loading a capture, such that the user can't interact with the UI while loading.
> I also made the "cancel" button of that progress bar stopping the loading and clearing the capture.

Ah perfect, thanks.

pierricgimmig (approved):

OrbitGl/App.cpp
```diff
@@ -638,13 +637,40 @@ ErrorMessageOr<void> OrbitApp::OnSaveCapture(const std::string& file_name) {
...
 
   DoZoom = true;  // TODO: remove global, review logic
-  return outcome::success();
+}
+
+void OrbitApp::OnLoadCatpureCanceled() {
+  capture_loading_cancellation_requested_ = true;
+  if (open_capture_failed_callback_) {
+    open_capture_failed_callback_();
+  }
```
dimitry-, line 672: I am wondering if you should wait here until capture is stopped. before calling ClearCapture.

florian-kuebler: Right, that would be safer. So you mean by introducing a second atomic<bool> that the deserialiser writes?

PR #1199, opened 2020-09-10
Remove top from obtaining CPU usage information

### LINKED ISSUE #1157: Problem: Unable to determine cpu utilization, wrong format from top command
OS: Ubuntu 20.04
Build: Development tip (9b6fbf49e99d65206637427d26cb489e9e590fdd)

Output from GetCpuUtilization() command (Utils.cpp)
```
damian@damian-GP72-7RD:~/projects/dependencies/development/orbit$ top -b -n 1 -w512 | sed -n '8, 1000{s/^ *//;s/ *$//;s/  */,/gp;};1000q'
32677,damian,20,0,20788,4116,3196,R,18.8,0.0,0:00.03,top
10104,damian,20,0,9057244,194136,112008,S,12.5,1.2,3:35.18,chrome
2182,root,20,0,485288,109752,66004,S,6.2,0.7,5:03.74,Xorg
2332,damian,20,0,393964,7792,5548,S,6.2,0.0,0:31.72,ibus-daemon
2448,damian,20,0,4944696,360448,82532,S,6.2,2.2,6:08.59,gnome-shell
3060,damian,20,0,915032,58128,32612,S,6.2,0.4,1:12.80,gnome-terminal-
7974,damian,20,0,670268,128780,64516,S,6.2,0.8,3:14.49,chrome
```
The columns are correct, so an attempt at parsing could be made.
```
damian@damian-GP72-7RD:~/projects/dependencies/development/orbit$ top -b -n 1 -w512
top - 08:38:00 up  1:27,  1 user,  load average: 0.97, 1.14, 1.42
Tasks: 452 total,   1 running, 451 sleeping,   0 stopped,   0 zombie
%Cpu(s):  5.0 us,  5.0 sy,  0.0 ni, 88.8 id,  0.0 wa,  0.0 hi,  1.2 si,  0.0 st
MiB Mem :  15806.1 total,    358.7 free,   5903.6 used,   9543.8 buff/cache
MiB Swap:   2048.0 total,   1957.0 free,     91.0 used.   8267.9 avail Mem 

    PID USER      PR  NI    VIRT    RES    SHR S  %CPU  %MEM     TIME+ COMMAND
```

The issue appears to be that occassionally there are more columns than expected:
```
9669,damian,20,0,4060172,19696,13096,S,11.8,0.1,2:28.33,ignore,-,main 14
618,root,20,0,0,0,0,S,0.0,0.0,0:00.00,UVM,global,queu 14
618,root,20,0,0,0,0,S,0.0,0.0,0:00.00,UVM,global,queu 14
618,root,20,0,0,0,0,S,0.0,0.0,0:00.00,UVM,global,queu 14
```

Modifying the following (OrbitService/Utils.cpp)
```
ErrorMessageOr<std::unordered_map<pid_t, double>> GetCpuUtilization(std::string_view top_data) {
  std::unordered_map<pid_t, double> process_map;
  for (const auto& line : absl::StrSplit(top_data, '\n')) {
    if (line.empty()) continue;
    std::vector<std::string> tokens = absl::StrSplit(line, ',');
    if (tokens.size() < 12) {
      return ErrorMessage(
          "Unable to determine cpu utilization, wrong format from top "
          "command.");
    }

    pid_t pid = strtol(tokens[0].c_str(), nullptr, 10);
    double cpu = strtod(tokens[8].c_str(), nullptr);
    process_map[pid] = cpu;
  }

  return process_map;
}
```

allows this method to work.

beckerhe:
Up to now we use top for getting CPU usage information per process.
This requires a fixed output format of top and is slow since top has to
run a certain time interval for averaging CPU usage.

We can do the same internally by reading /proc/stat and
/proc/[pid]/stat. This commit adds this logic to
orbit_service::ProcessList.

The routines reading from `/proc` live in orbit_service::utils and are
unit-tested.

The CPU usage is averaged between two consecutive calls of
ProcessList::Refresh. This means the averaging time period is not always
constant and depends on when Refresh is called. But I believe it's good
enough for our use case.

Test: Manual test, compared process list cpu output to top running in
shell
Bug: http://b/165823980
Fixes #1157

### DIFF
--- a/OrbitService/CMakeLists.txt
+++ b/OrbitService/CMakeLists.txt
@@ -21,6 +21,8 @@ target_sources(OrbitServiceLib PRIVATE
         OrbitGrpcServer.h
         OrbitService.cpp
         OrbitService.h
+        Process.cpp
+        Process.h
         ProcessList.cpp
         ProcessList.h
         ProcessServiceImpl.cpp
@@ -67,7 +69,8 @@ strip_symbols(OrbitService)
 add_executable(OrbitServiceTests)
 target_compile_options(OrbitServiceTests PRIVATE ${STRICT_COMPILE_FLAGS})
 
-target_sources(OrbitServiceTests PRIVATE UtilsTest.cpp Utils.cpp)
+target_sources(OrbitServiceTests PRIVATE UtilsTest.cpp
+                                         ProcessTest.cpp)
 
 target_link_libraries(OrbitServiceTests PRIVATE 
         OrbitServiceLib
@@ -78,6 +81,7 @@ add_custom_command(TARGET OrbitServiceTests POST_BUILD
         $<TARGET_FILE_DIR:OrbitServiceTests>/testdata)
 
 register_test(OrbitServiceTests)
+set_tests_properties(OrbitService PROPERTIES TIMEOUT 10)
 
 add_fuzzer(OrbitServiceUtilsFindSymbolsFilePathFuzzer
            OrbitServiceUtilsFindSymbolsFilePathFuzzer.cpp)
--- a/OrbitService/CMakeLists.txt
+++ b/OrbitService/CMakeLists.txt
@@ -70,6 +70,7 @@ add_executable(OrbitServiceTests)
 target_compile_options(OrbitServiceTests PRIVATE ${STRICT_COMPILE_FLAGS})
 
 target_sources(OrbitServiceTests PRIVATE UtilsTest.cpp
+                                         ProcessListTest.cpp
                                          ProcessTest.cpp)
 
 target_link_libraries(OrbitServiceTests PRIVATE 
--- a/OrbitService/Process.cpp
+++ b/OrbitService/Process.cpp
@@ -19,7 +19,26 @@
 
 namespace orbit_service {
 
-ErrorMessageOr<Process> Process::FromPidAndCpuUsage(pid_t pid, double cpu_usage) {
+void Process::UpdateCpuUsage(utils::Jiffies process_cpu_time, utils::Jiffies total_cpu_time) {
+  const auto diff_process_cpu_time =
+      static_cast<double>(process_cpu_time.value - previous_process_cpu_time_.value);
+  const auto diff_total_cpu_time =
+      static_cast<double>(total_cpu_time.value - previous_total_cpu_time_.value);
+
+  // When the counters wrap, `cpu_usage` might be smaller than 0.0 or larger than 1.0,
+  // depending on the signedness of `Jiffies`. Reference implementations like top and htop usually
+  // clamp in this case. So that's what we're also doing here.
+  const auto cpu_usage = std::clamp(diff_process_cpu_time / diff_total_cpu_time, 0.0, 1.0);
+
+  // TODO(hebecker): Rename cpu_usage to cpu_usage_rate and normalize. Being in percent was
+  // surprising
+  set_cpu_usage(cpu_usage * 100.0);
+
+  previous_process_cpu_time_ = process_cpu_time;
+  previous_total_cpu_time_ = total_cpu_time;
+}
+
+ErrorMessageOr<Process> Process::FromPid(pid_t pid) {
   const auto path = std::filesystem::path{"/proc"} / std::to_string(pid);
 
   if (!std::filesystem::is_directory(path)) {
@@ -43,7 +62,14 @@ ErrorMessageOr<Process> Process::FromPidAndCpuUsage(pid_t pid, double cpu_usage)
   Process process{};
   process.set_pid(pid);
   process.set_name(name);
-  process.set_cpu_usage(cpu_usage);
+
+  const auto total_cpu_time = utils::GetCumulativeTotalCpuTime();
+  const auto cpu_time = utils::GetCumulativeCpuTimeFromProcess(process.pid());
+  if (cpu_time && total_cpu_time) {
+    process.UpdateCpuUsage(cpu_time.value(), total_cpu_time.value());
+  } else {
+    LOG("Could not update the CPU usage of process %d", process.pid());
+  }
 
   // "The command-line arguments appear [...] as a set of strings
   // separated by null bytes ('\0')".
--- a/OrbitService/Process.cpp
+++ b/OrbitService/Process.cpp
@@ -1,0 +1,77 @@
+// Copyright (c) 2020 The Orbit Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#include "Process.h"
+
+#include <absl/strings/ascii.h>
+#include <absl/strings/str_format.h>
+
+#include <algorithm>
+#include <cmath>
+#include <cstdint>
+#include <filesystem>
+
+#include "ElfUtils/ElfFile.h"
+#include "OrbitBase/Logging.h"
+#include "OrbitBase/Result.h"
+#include "Utils.h"
+
+namespace orbit_service {
+
+ErrorMessageOr<Process> Process::FromPidAndCpuUsage(pid_t pid, double cpu_usage) {
+  const auto path = std::filesystem::path{"/proc"} / std::to_string(pid);
+
+  if (!std::filesystem::is_directory(path)) {
+    return ErrorMessage{absl::StrFormat("PID %d does not exist", pid)};
+  }
+
+  const std::filesystem::path name_file_path = path / "comm";
+  auto name_file_result = utils::ReadFileToString(name_file_path);
+  if (!name_file_result) {
+    return ErrorMessage{absl::StrFormat("Failed to read %s: %s", name_file_path.string(),
+                                        name_file_result.error().message())};
+  }
+
+  std::string name = std::move(name_file_result.value());
+  // Remove new line character.
+  absl::StripTrailingAsciiWhitespace(&name);
+  if (name.empty()) {
+    return ErrorMessage{absl::StrFormat("Could not determine the process name of process %d", pid)};
+  }
+
+  Process process{};
+  process.set_pid(pid);
+  process.set_name(name);
+  process.set_cpu_usage(cpu_usage);
+
+  // "The command-line arguments appear [...] as a set of strings
+  // separated by null bytes ('\0')".
+  const std::filesystem::path cmdline_file_path = path / "cmdline";
+  auto cmdline_file_result = utils::ReadFileToString(cmdline_file_path);
+  if (!cmdline_file_result) {
+    return ErrorMessage{absl::StrFormat("Failed to read %s: %s", cmdline_file_path.string(),
+                                        name_file_result.error().message())};
+  }
+
+  std::string cmdline = std::move(cmdline_file_result.value());
+  std::replace(cmdline.begin(), cmdline.end(), '\0', ' ');
+  process.set_command_line(cmdline);
+
+  auto file_path_result = utils::GetExecutablePath(pid);
+  if (file_path_result) {
+    process.set_full_path(std::move(file_path_result.value()));
+
+    const auto& elf_file = ElfUtils::ElfFile::Create(file_path_result.value().string());
+    if (elf_file) {
+      process.set_is_64_bit(elf_file.value()->Is64Bit());
+    } else {
+      LOG("Warning: Unable to parse the executable \"%s\" as elf file. (pid: %d)",
+          file_path_result.value(), pid);
+    }
+  }
+
+  return process;
+}
+
+}  // namespace orbit_service
--- a/OrbitService/Process.h
+++ b/OrbitService/Process.h
@@ -1,0 +1,24 @@
+// Copyright (c) 2020 The Orbit Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#ifndef ORBIT_SERVICE_PROCESS_H_
+#define ORBIT_SERVICE_PROCESS_H_
+
+#include "OrbitBase/Result.h"
+#include "process.pb.h"
+
+namespace orbit_service {
+
+class Process : public orbit_grpc_protos::ProcessInfo {
+ public:
+  using orbit_grpc_protos::ProcessInfo::ProcessInfo;
+
+  // Creates a `Process` by reading details from the `/proc` filesystem.
+  // This might fail due to a non existing pid or due to permission problems.
+  static ErrorMessageOr<Process> FromPidAndCpuUsage(pid_t pid, double cpu_usage);
+};
+
+}  // namespace orbit_service
+
+#endif  // ORBIT_SERVICE_PROCESS_H_
--- a/OrbitService/Process.h
+++ b/OrbitService/Process.h
@@ -6,6 +6,7 @@
 #define ORBIT_SERVICE_PROCESS_H_
 
 #include "OrbitBase/Result.h"
+#include "Utils.h"
 #include "process.pb.h"
 
 namespace orbit_service {
@@ -14,9 +15,15 @@ class Process : public orbit_grpc_protos::ProcessInfo {
  public:
   using orbit_grpc_protos::ProcessInfo::ProcessInfo;
 
+  void UpdateCpuUsage(utils::Jiffies process_cpu_time, utils::Jiffies total_cpu_time);
+
   // Creates a `Process` by reading details from the `/proc` filesystem.
   // This might fail due to a non existing pid or due to permission problems.
-  static ErrorMessageOr<Process> FromPidAndCpuUsage(pid_t pid, double cpu_usage);
+  static ErrorMessageOr<Process> FromPid(pid_t pid);
+
+ private:
+  utils::Jiffies previous_process_cpu_time_ = {};
+  utils::Jiffies previous_total_cpu_time_ = {};
 };
 
 }  // namespace orbit_service
--- a/OrbitService/ProcessList.cpp
+++ b/OrbitService/ProcessList.cpp
@@ -4,7 +4,6 @@
 
 #include "ProcessList.h"
 
-#include <absl/strings/ascii.h>
 #include <absl/strings/numbers.h>
 #include <absl/strings/str_format.h>
 
@@ -18,14 +17,7 @@
 namespace orbit_service {
 
 ErrorMessageOr<void> ProcessList::Refresh() {
-  auto cpu_result = utils::GetCpuUtilization();
-  if (!cpu_result) {
-    return outcome::failure(absl::StrFormat("Unable to retrieve cpu usage of processes: %s",
-                                            cpu_result.error().message()));
-  }
-  std::unordered_map<int32_t, double> cpu_usage_map = std::move(cpu_result.value());
-
-  std::vector<Process> updated_processes;
+  absl::flat_hash_map<pid_t, Process> updated_processes{};
 
   // TODO(b/161423785): This for loop should be refactored. For example, when
   //  parts are in a separate function, OUTCOME_TRY could be used to simplify
@@ -39,17 +31,28 @@ ErrorMessageOr<void> ProcessList::Refresh() {
     uint32_t pid;
     if (!absl::SimpleAtoi(folder_name, &pid)) continue;
 
-    auto iter = processes_map_.find(pid);
-    if (iter != processes_map_.end()) {
-      Process& process = *(iter->second);
-      process.set_cpu_usage(cpu_usage_map[process.pid()]);
-      updated_processes.push_back(process);
+    const auto iter = processes_.find(pid);
+    if (iter != processes_.end()) {
+      auto process = processes_.extract(iter);
+
+      const auto total_cpu_time = utils::GetCumulativeTotalCpuTime();
+      const auto cpu_time = utils::GetCumulativeCpuTimeFromProcess(process.key());
+      if (cpu_time && total_cpu_time) {
+        process.mapped().UpdateCpuUsage(cpu_time.value(), total_cpu_time.value());
+      } else {
+        // We don't fail in this case. This could be a permission problem which might occur when not
+        // running as root.
+        ERROR("Could not update the CPU usage of process %d", process.key());
+      }
+
+      updated_processes.insert(std::move(process));
       continue;
     }
 
-    auto process = Process::FromPidAndCpuUsage(pid, cpu_usage_map[pid]);
+    auto process = Process::FromPid(pid);
+
     if (process) {
-      updated_processes.emplace_back(std::move(process.value()));
+      updated_processes.emplace(pid, std::move(process.value()));
     } else {
       // We don't fail in this case. This could be a permission problem which is restricted to a
       // small amount of processes.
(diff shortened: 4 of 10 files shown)


OrbitService/Process.cpp
```diff
@@ -0,0 +1,99 @@
...
+    process.set_full_path(std::move(file_path_result.value()));
+
+    const auto& elf_file = ElfUtils::ElfFile::Create(file_path_result.value().string());
+    if (elf_file) {
+      process.set_is_64_bit(elf_file.value()->Is64Bit());
+    } else {
+      LOG("Warning: Unable to parse the executable \"%s\" as elf file. (pid: %d)",
+          file_path_result.value(), pid);
+    }
+  }
```
beckerhe, line 94: This part was previously living in ProcessList.cpp and was moved here.

OrbitService/Process.cpp
```diff
@@ -0,0 +1,99 @@
...
+  std::string name = std::move(name_file_result.value());
+  // Remove new line character.
+  absl::StripTrailingAsciiWhitespace(&name);
+  if (name.empty()) {
+    return ErrorMessage{"Could not determine the process name"};
+  }
+
+  Process process{};
+  process.set_pid(pid);
+  process.set_name(name);
```
beckerhe, line 60: This part was previously living in ProcessList.cpp and was moved here.

OrbitService/Utils.cpp
```diff
@@ -175,29 +158,117 @@ ErrorMessageOr<std::vector<orbit_grpc_protos::TracepointInfo>> ReadTracepoints()
...
+    return {};
+  }
+
+  std::string first_line{};
+  std::getline(stream, first_line);
+
+  std::vector<std::string_view> fields = absl::StrSplit(first_line, ' ', absl::SkipWhitespace{});
+
+  constexpr size_t utime_index = 13;
+  constexpr size_t stime_index = 14;
```
dpallotti, line 215: `kUtimeIndex`, `kStimeIndex`?

OrbitService/Utils.cpp
```diff
@@ -175,29 +158,117 @@ ErrorMessageOr<std::vector<orbit_grpc_protos::TracepointInfo>> ReadTracepoints()
...
+  template <typename T>
+  std::vector<T> operator()(std::vector<T> vec, const std::optional<T>& val) const {
+    if (val.has_value()) {
+      vec.emplace_back(std::move(val.value()));
+    }
+
+    return std::move(vec);
+  }
+};
+}  // namespace
```
dpallotti, line 189: Is it not possible to use a lambda instead?

OrbitService/ProcessList.h
```diff
@@ -9,20 +9,61 @@
...
+    [[nodiscard]] bool operator()(const pid_t lhs, const Process& rhs) const {
+      return lhs == rhs.pid();
+    }
+
+    [[nodiscard]] bool operator()(const Process& lhs, const pid_t rhs) const {
+      return lhs.pid() == rhs;
+    }
+  };
+
+  absl::flat_hash_set<Process, ProcessHash, ProcessEq> processes_;
```
dpallotti, line 66: Wouldn't a map from int32_t to orbit_grpc_protos::ProcessInfo be more intuitive? I have to say that I find it unnecessary complicated to have to define hash and equality that work for both Process and pid_t when the access is (or can be) only by pid.

OrbitService/Process.cpp
```diff
@@ -0,0 +1,99 @@
+// Copyright (c) 2020 The Orbit Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#include "Process.h"
```
dimitry-, line 5: would be nice to have commit moving Process separate from this commit - this way it is more visible what was changed.

OrbitService/Process.h
```diff
@@ -0,0 +1,29 @@
...
+
+#ifndef ORBIT_SERVICE_PROCESS_H_
+#define ORBIT_SERVICE_PROCESS_H_
+
+#include "Utils.h"
+#include "process.pb.h"
+
+namespace orbit_service {
+
+struct Process : orbit_grpc_protos::ProcessInfo {
```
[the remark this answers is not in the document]
beckerhe: `struct` and `class` have identical meaning, but `struct` has more sensible defaults. (Public members by default which is nice when you want to put private members at the bottom / public inheritance). But I can change if you prefer a `class`. I don't have a strong opinion here.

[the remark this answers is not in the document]
beckerhe: It's public inheritance since it's a struct.

OrbitService/ProcessList.cpp
```diff
@@ -41,66 +33,44 @@ ErrorMessageOr<void> ProcessList::Refresh() {
...
-    auto iter = processes_map_.find(pid);
-    if (iter != processes_map_.end()) {
-      ProcessInfo& process(*(iter->second));
-      process.set_cpu_usage(cpu_usage_map[process.pid()]);
-      updated_processes.push_back(process);
-      continue;
-    }
+    const auto iter = processes_.find(pid);
+    if (iter != processes_.end()) {
+      auto process = processes_.extract(iter);
```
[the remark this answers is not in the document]
beckerhe: extract returns the internal `node_type` which can the be inserted in another map container. This is/can be more efficient.

OrbitService/ProcessListTest.cpp
```diff
@@ -0,0 +1,37 @@
...
+#include "OrbitBase/Logging.h"
+#include "ProcessList.h"
+#include "gtest/gtest.h"
+
+namespace orbit_service {
+
+TEST(ProcessList, ProcessList) {
+  ProcessList process_list;
+  const auto result1 = process_list.Refresh();
+  ASSERT_TRUE(result1) << result1.error().message();
```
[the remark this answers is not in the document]
beckerhe: I don't mind to write it differently. In this particular case I copied the code from somewhere else because of the stream-operator which I hadn't used yet in gtest.

OrbitService/ProcessList.h
beckerhe: Since `ProcessInfo` already contained the PID, it would mean that the PID is stored in the key and in the value-struct and must be kept in sync. For me personally this feels bug-prone, semantically wrong and less intuitive than a set.

Can you explain where you see the complexity? I admit, defining ProcessHash and ProcessEq is verbose, but is it difficult? Maybe I'm misjudging that...

OrbitService/Process.cpp
beckerhe: I will tackle that when Davide's comments are resolved.

OrbitService/Process.cpp
```diff
@@ -0,0 +1,99 @@
...
+
+  previous_process_cpu_time_ = process_cpu_time;
+  previous_total_cpu_time_ = total_cpu_time;
+}
+
+ErrorMessageOr<Process> Process::FromPid(pid_t pid) {
+  const auto path = std::filesystem::path{"/proc"} / std::to_string(pid);
+
+  if (!std::filesystem::is_directory(path)) {
+    return ErrorMessage{absl::StrFormat("PID %d does not exist!", pid)};
```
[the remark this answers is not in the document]
beckerhe: Ahhhh. Good point. Fixed.

OrbitService/Process.cpp
```diff
@@ -0,0 +1,99 @@
...
+  if (!name_file_result) {
+    return ErrorMessage{absl::StrFormat("Failed to read %s: %s", name_file_path.string(),
+                                        name_file_result.error().message())};
+  }
+
+  std::string name = std::move(name_file_result.value());
+  // Remove new line character.
+  absl::StripTrailingAsciiWhitespace(&name);
+  if (name.empty()) {
+    return ErrorMessage{"Could not determine the process name"};
```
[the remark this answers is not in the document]
beckerhe: Good idea. Done.

OrbitService/Process.h
```diff
@@ -0,0 +1,29 @@
...
+  using orbit_grpc_protos::ProcessInfo::ProcessInfo;
+
+  void UpdateCpuUsage(utils::Jiffies process_cpu_time, utils::Jiffies total_cpu_time);
+
+  // Creates a `Process` by reading details from the `/proc` filesystem.
+  // This might failed due to a non existing pid or due to permission problems.
```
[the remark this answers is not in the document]
beckerhe: Thanks!

OrbitService/ProcessList.cpp
```diff
@@ -41,66 +33,44 @@ ErrorMessageOr<void> ProcessList::Refresh() {
...
-      } else {
-        LOG("Warning: Unable to parse the executable \"%s\" as elf file. (pid: %d)",
-            file_path_result.value(), pid);
-      }
+    if (process) {
+      updated_processes.emplace(std::move(process.value()));
+    } else {
+      // We don't fail in this case. This could be a permission problem which is restricted to a
+      // small amount of processes.
+      ERROR("Could not create process list entry for pid %d. Details: %s", pid,
```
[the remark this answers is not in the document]
beckerhe: Yes. Thanks. Fixed.

OrbitService/Utils.cpp
beckerhe: Done.

OrbitService/Utils.cpp
beckerhe: Yeah, good point. This was used in two different functions before, but I could simplifiy the code and forgot to remove that helper function again. Thanks!

OrbitService/Process.h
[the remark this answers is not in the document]
beckerhe: I just saw the google style guide prefers `class` in this case. I will change it!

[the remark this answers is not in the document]
beckerhe: Fixed. Thanks!

OrbitService/CMakeLists.txt
```diff
@@ -67,7 +69,8 @@ strip_symbols(OrbitService)
 add_executable(OrbitServiceTests)
 target_compile_options(OrbitServiceTests PRIVATE ${STRICT_COMPILE_FLAGS})
 
-target_sources(OrbitServiceTests PRIVATE UtilsTest.cpp Utils.cpp)
+target_sources(OrbitServiceTests PRIVATE UtilsTest.cpp ProcessTest.cpp
+                                         ProcessListTest.cpp)
```
antonrohr, line 73: nit
```suggestion
target_sources(OrbitServiceTests PRIVATE UtilsTest.cpp
                                         ProcessTest.cpp
                                         ProcessListTest.cpp)
```

OrbitService/ProcessTest.cpp
```diff
@@ -0,0 +1,25 @@
+// Copyright (c) 2020 The Orbit Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#include <unistd.h>
+
+#include "OrbitBase/Logging.h"
```
antonrohr, line 7: nit
```suggestion

```

OrbitService/ProcessList.cpp
```diff
@@ -41,66 +33,43 @@ ErrorMessageOr<void> ProcessList::Refresh() {
...
+      const auto cpu_time = utils::GetCumulativeCpuTimeFromProcess(process.value().pid());
+      if (cpu_time && total_cpu_time) {
+        process.value().UpdateCpuUsage(cpu_time.value(), total_cpu_time.value());
+      } else {
+        // We don't fail in this case. This could be a permission problem which might occur when not
+        // running as root.
+        ERROR("Could not update the CPU usage of process %d", process.value().pid());
+      }
+
+      updated_processes.insert(std::move(process));
```
antonrohr, line 50: Should this maybe only be done when we succeed in updating the cpu usage? Just an idea. Feel free to dismiss

OrbitService/ProcessListTest.cpp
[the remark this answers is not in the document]
antonrohr: I like how it is done currently. (Also, the stream operator is independent from the bool conversion. If i understood it right, the stream operator is only for producing better error messages)

OrbitService/ProcessList.h
beckerhe: It would be great to get some other opinions of that. Obviously I like my code :sweat_smile: , but I'm also happy to comply with the majority. :see_no_evil:

OrbitService/ProcessListTest.cpp
```diff
@@ -0,0 +1,37 @@
...
+
+  const auto process1 = process_list.GetProcessByPid(getpid());
+  EXPECT_TRUE(process1);
+
+  const auto total_cpu_cycles = utils::GetCumulativeTotalCpuTime().value();
+
+  // We wait until the stats have been updated
+  while (utils::GetCumulativeTotalCpuTime().value().value == total_cpu_cycles.value) {
+    // If this loop never ends it will be caught by the automatic timeout feature
+    usleep(10'000);
```
antonrohr, line 27: If you use the test timeout as a thing, consider adding the following to the cmake file. (I think the default timeout is 30 minutes and that is excessive)
`set_tests_properties(OrbitService PROPERTIES TIMEOUT 10)`

OrbitService/ProcessList.h
dimitry-: I also think map looks better here.

OrbitService/ProcessListTest.cpp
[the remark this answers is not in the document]
beckerhe: Yes, I only mentioned the streaming operator to explain that I copied that code from somewhere else where the streaming operator was used.

OrbitService/ProcessList.h
beckerhe: Alright. I changed it to a `absl::flat_hash_map<pid_t, Process>`.

OrbitService/UtilsTest.cpp
```diff
@@ -96,44 +97,57 @@ TEST(Utils, ParseMaps) {
 
 }  // namespace orbit_service::utils
```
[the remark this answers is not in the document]
beckerhe: Done. Thanks.

OrbitService/Utils.cpp
```diff
@@ -175,29 +158,112 @@ ErrorMessageOr<std::vector<orbit_grpc_protos::TracepointInfo>> ReadTracepoints()
...
+std::vector<pid_t> GetAllPids() {
+  const auto emplace_back_if_has_value = [](auto vec, const auto& val) {
+    if (val.has_value()) {
+      vec.emplace_back(std::move(val.value()));
+    }
+
+    return std::move(vec);
+  };
+
+  std::filesystem::directory_iterator proc{"/proc"};
```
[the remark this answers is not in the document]
beckerhe: Yes. Done. Thanks.

OrbitService/CMakeLists.txt
beckerhe: Done.

OrbitService/ProcessTest.cpp
beckerhe: Nice catch! Thanks. Done.

OrbitService/ProcessList.cpp
beckerhe: Good question. I thought about these kind of things and decided that I don't want to change behaviour in this PR. Not reading CPU usage is something which usually happens only when the service is not running as root which is currently out of scope. We should probably think about how to handle that in the future, things like differentiating between processes that can be profiled and these that can't.

OrbitService/ProcessListTest.cpp
beckerhe: Good point. Done. Thanks.

beckerhe:
I'm going to split the commits now and the re-request a review from you.

OrbitService/ProcessList.cpp
antonrohr: Sounds good

OrbitService/Process.cpp
beckerhe: This is now in a separate commit. I'm going to make some more tests before I re-request review.

OrbitService/ProcessList.cpp
```diff
@@ -43,55 +41,20 @@ ErrorMessageOr<void> ProcessList::Refresh() {
 
     auto iter = processes_map_.find(pid);
     if (iter != processes_map_.end()) {
-      ProcessInfo& process(*(iter->second));
+      Process& process(*(iter->second));
```
dimitry-: it looks a bit weird - this initialized reference not an object correct? maybe use = here?

OrbitService/Process.cpp
```diff
@@ -19,7 +19,22 @@
...
-ErrorMessageOr<Process> Process::FromPidAndCpuUsage(pid_t pid, double cpu_usage) {
+void Process::UpdateCpuUsage(utils::Jiffies process_cpu_time, utils::Jiffies total_cpu_time) {
+  const auto diff_process_cpu_time =
+      static_cast<double>(process_cpu_time.value - previous_process_cpu_time_.value);
+  const auto diff_total_cpu_time =
+      static_cast<double>(total_cpu_time.value - previous_total_cpu_time_.value);
+
+  const auto cpu_usage = std::clamp(diff_process_cpu_time / diff_total_cpu_time, 0.0, 1.0);
+
+  // TODO(hebecker): Rename cpu_usage to cpu_usage_percent. Not being normalized was surprising
```
dimitry-, line 30: interesting - maybe we should do the opposite and use rate here - normalize on the client-side.

OrbitService/Process.cpp
```diff
@@ -19,7 +19,22 @@
...
 namespace orbit_service {
 
-ErrorMessageOr<Process> Process::FromPidAndCpuUsage(pid_t pid, double cpu_usage) {
+void Process::UpdateCpuUsage(utils::Jiffies process_cpu_time, utils::Jiffies total_cpu_time) {
+  const auto diff_process_cpu_time =
+      static_cast<double>(process_cpu_time.value - previous_process_cpu_time_.value);
+  const auto diff_total_cpu_time =
+      static_cast<double>(total_cpu_time.value - previous_total_cpu_time_.value);
+
+  const auto cpu_usage = std::clamp(diff_process_cpu_time / diff_total_cpu_time, 0.0, 1.0);
```
dimitry-, line 28: can this be less than 0?

OrbitService/ProcessList.cpp
```diff
@@ -18,14 +17,7 @@
...
 ErrorMessageOr<void> ProcessList::Refresh() {
-  auto cpu_result = utils::GetCpuUtilization();
-  if (!cpu_result) {
-    return outcome::failure(absl::StrFormat("Unable to retrieve cpu usage of processes: %s",
-                                            cpu_result.error().message()));
-  }
-  std::unordered_map<int32_t, double> cpu_usage_map = std::move(cpu_result.value());
-
-  std::vector<Process> updated_processes;
+  decltype(processes_) updated_processes{};
```
dimitry-, line 20: use explicit type here decltype is not very readable.

OrbitService/Process.cpp
dimitry-, line 28: Could it be < 0 or more than 1?

OrbitService/Utils.cpp
```diff
@@ -175,29 +158,112 @@ ErrorMessageOr<std::vector<orbit_grpc_protos::TracepointInfo>> ReadTracepoints()
...
+
+std::optional<Jiffies> GetCumulativeCpuTimeFromProcess(pid_t pid) {
+  const auto stat = std::filesystem::path{"/proc"} / std::to_string(pid) / "stat";
+
+  if (!std::filesystem::exists(stat)) {
+    return {};
+  }
+
+  std::ifstream stream{stat.string()};
+  if (!stream.good()) {
```
dimitry-, line 200: log something here (not sure if it will result in spam in logs though)?

OrbitService/Utils.cpp
```diff
@@ -175,29 +158,112 @@ ErrorMessageOr<std::vector<orbit_grpc_protos::TracepointInfo>> ReadTracepoints()
...
+    return std::move(vec);
+  };
+
+  fs::directory_iterator proc{"/proc"};
+  return std::transform_reduce(begin(proc), end(proc), std::vector<pid_t>{},
+                               emplace_back_if_has_value, ProcEntryToPid);
+}
+
+std::optional<Jiffies> GetCumulativeCpuTimeFromProcess(pid_t pid) {
+  const auto stat = std::filesystem::path{"/proc"} / std::to_string(pid) / "stat";
```
dimitry-, line 193: add comment with format/example of /stat so this is clear which part this function is extracting.

OrbitService/Utils.cpp
```diff
@@ -175,29 +158,112 @@ ErrorMessageOr<std::vector<orbit_grpc_protos::TracepointInfo>> ReadTracepoints()
...
-  std::unordered_map<pid_t, double> process_map;
-  for (const auto& line : absl::StrSplit(top_data, '\n')) {
-    if (line.empty()) continue;
-    std::vector<std::string> tokens = absl::StrSplit(line, ',');
-    if (tokens.size() != 12) {
-      return ErrorMessage(
-          "Unable to determine cpu utilization, wrong format from top "
-          "command.");
+std::optional<Jiffies> GetCumulativeTotalCpuTime() {
+  std::ifstream stat_stream{"/proc/stat"};
```
dimitry-, line 230: same here.

dimitry- (approved):
otherwise lgtm

antonrohr (approved):
Thanks for including the changes

OrbitService/ProcessList.cpp
beckerhe: I agree. done.

OrbitService/Process.cpp
beckerhe: Good point. I adjusted the comment.

OrbitService/Process.cpp
beckerhe: No, it can't because Jiffies is unsigned. But it can be bigger than 1.0 when the counter wraps.

But if Jiffies were signed we would get a negative values in the wrapping case. So I had two choices: Either add a static_assert which ensure that Jiffies is unsigned or add the clamp. I felt that the clamp is less surprising. It's also part of all the examples I've on the internet. (top has clamp, htop as well, the stuff on StackOverflow also has it.)

Do you think it's misleading?

OrbitService/ProcessList.cpp
beckerhe: Done.

OrbitService/Process.cpp
beckerhe: See other comment.

OrbitService/Utils.cpp
beckerhe: Done.

OrbitService/Utils.cpp
beckerhe: Done.

OrbitService/Utils.cpp
beckerhe: Done.

beckerhe:
I made some more tests. Output looks good, even after all the changes. Capturing also seems to work fine.

beckerhe:
I will wait for Davide before merging.

OrbitService/Process.cpp
dimitry-: no but I think it deserves a short comment as of when it can be > 1.0

beckerhe: Yes, that makes sense. Done. And thank you!

OrbitService/ProcessList.cpp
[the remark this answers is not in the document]
dpallotti: Ah interesting, thanks!

dpallotti (approved):

OrbitService/Process.h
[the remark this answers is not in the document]
dpallotti: Yeah, this was part of not being used to `struct` instead of `class` in such a case. Thanks for changing to `class`.

PR #1201, opened 2020-09-11
Automation script and test data for e2e preset test.

danielfenner:
Bug: b/161839506
Test: Copied the presets in right place, python.exe orbit_load_presets.py, works

### DIFF
--- a/contrib/automation_tests/orbit_load_presets.py
+++ b/contrib/automation_tests/orbit_load_presets.py
@@ -36,6 +36,7 @@ def ConnectToGamelet(application):
   dialog = application.window(title_re='orbitprofiler')
   dialog.set_focus()
   # DataItem0 contains the name of the first gamelet.
+  dialog.DataItem0.wait('exists', timeout=100)
   dialog.DataItem0.click_input()
   dialog.OK.click_input()
   logging.info('Connected to Gamelet.')
@@ -49,10 +50,6 @@ def main(argv):
   WaitForOrbit()
 
   application = Application(backend='uia').connect(title_re='orbitprofiler')
-  if application.is_process_running():
-    logging.info('Orbit is running.')
-  else:
-    logging.info('Orbit is not running.')
 
   ConnectToGamelet(application)
 
--- a/contrib/automation_tests/orbit_load_presets.py
+++ b/contrib/automation_tests/orbit_load_presets.py
@@ -1,0 +1,160 @@
+"""Apply two presets in Orbit using pywinauto.
+
+Before this script is run there needs to be a gamelet reserved and
+"hello_ggp_standalone" has to be started. Two presets named
+draw_frame_in_hello_ggp_1_52.opr and ggp_issue_frame_token_in_hello_ggp_1_52
+(hooking the functions DrawFrame and GgpIssueFrameToken) need to exits in the
+preset folder.
+
+The script requires absl and pywinauto. Since pywinauto requires the bitness of
+the python installation to match the bitness of the program under test it needs
+to by run from  64 bit python.
+"""
+import logging
+import time
+from absl import app
+import pywinauto
+from pywinauto.application import Application
+
+
+def WaitForOrbit():
+  while True:
+    try:
+      Application(backend='uia').connect(title_re='orbitprofiler')
+      break
+    except pywinauto.findwindows.ElementNotFoundError:
+      pass
+
+
+def ConnectToGamelet(application):
+  """Connect to the first gamelet in the start dialog.
+
+  Args:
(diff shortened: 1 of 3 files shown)


contrib/automation_tests/orbit_load_presets.py
```diff
@@ -0,0 +1,160 @@
+"""Apply two presets in Orbit using pywinauto.
```
florian-kuebler, line 1: In python, I prefer single-line comments, but if you want to use multi-line strings, you should add a line break after the """.

contrib/automation_tests/orbit_load_presets.py
```diff
@@ -0,0 +1,160 @@
...
+  while True:
+    try:
+      Application(backend='uia').connect(title_re='orbitprofiler')
+      break
+    except pywinauto.findwindows.ElementNotFoundError:
+      pass
+
+
+def ConnectToGamelet(application):
+  """Connect to the first gamelet in the start dialog.
```
florian-kuebler, line 30: Again, I would add a linebreak or use single-line comments.

florian-kuebler: Also, that comment should move above the method, right?

contrib/automation_tests/orbit_load_presets.py
```diff
@@ -0,0 +1,160 @@
...
+  dialog = application.window(title_re='orbitprofiler')
+  dialog.set_focus()
+  # DataItem0 contains the name of the first gamelet.
+  dialog.DataItem0.click_input()
+  dialog.OK.click_input()
+  logging.info('Connected to Gamelet.')
+  # Wait until the service is deployed and the main window is opened.
+  # Waiting for the window does not really work since there is a progress dialog
+  # with an identical title popping up in between. So we just sleep.
+  time.sleep(15)
```
florian-kuebler, line 45: couldn't you instead wait, until the dialog disappears?

contrib/automation_tests/orbit_load_presets.py
```diff
@@ -0,0 +1,160 @@
...
+
+
+def main(argv):
+  WaitForOrbit()
+
+  application = Application(backend='uia').connect(title_re='orbitprofiler')
+  if application.is_process_running():
+    logging.info('Orbit is running.')
+  else:
+    logging.info('Orbit is not running.')
```
florian-kuebler, line 55: If orbit is not running, we should fail, right?

contrib/automation_tests/orbit_load_presets.py
```diff
@@ -0,0 +1,160 @@
...
+
+  ConnectToGamelet(application)
```
florian-kuebler, line 57: Out of curiosity: How do you make sure, that ggp already listed the gamelets?

florian-kuebler (approved):
Thanks Daniel for moving the test to this repo. In general LGTM. I left some minor comments, but feel free to ignore them.

contrib/automation_tests/orbit_load_presets.py
florian-kuebler: Just looked up python's doc-strings again. And you are right about the format. So forget what I said.

contrib/automation_tests/orbit_load_presets.py
florian-kuebler: See above.

contrib/automation_tests/orbit_load_presets.py
danielfenner: There is a sequence of windows popping up during the deployment. I think the right way  would be to query an element from the main window in a try/except block, loop that forever or until a timeout happens and break from the loop when the element from the main window shows up. Pretty much as I do in WaitForOrbit.

I'll refactor this when I see this failing for the first time.

contrib/automation_tests/orbit_load_presets.py
danielfenner: Actually this is left over from debugging when I was unable to start Orbit properly. I don't think this makes sense anymore. Will remove this.

florian-kuebler: Okay, also fine.

contrib/automation_tests/orbit_load_presets.py
florian-kuebler: Fine for me. Thanks for clarifying.

contrib/automation_tests/orbit_load_presets.py
danielfenner: Good question. That just never occurred until now since this script is soooo lame. Will add a proper wait inside ConnectToGamelet.

PR #1202, opened 2020-09-11
Update tracepoint events primitives

MariaSandru27:


OrbitCore/CaptureData.h
```diff
@@ -132,6 +132,15 @@ class CaptureData {
 
   [[nodiscard]] const CallstackData* GetCallstackData() const { return callstack_data_.get(); };
 
+  [[nodiscard]] Mutex& GetTracepointEventBufferMutex() const {
+    return tracepoint_event_buffer_->mutex();
+  }
+
+  [[nodiscard]] std::map<int32_t, std::map<uint64_t, orbit_client_protos::TracepointEventInfo> >
```
danielfenner, line 139: This will make a copy of the map. I guess you want to return a (const?) reference.

OrbitGl/TracepointTrack.cpp
```diff
@@ -0,0 +1,39 @@
...
+  float track_height = layout.GetEventTrackHeight();
+
+  ScopeLock lock(GOrbitApp->GetCaptureData().GetTracepointEventBufferMutex());
+
+  std::map<uint64_t, orbit_client_protos::TracepointEventInfo>& tracepoints =
+      GOrbitApp->GetCaptureData().GetTracepointEventBufferTracepoints().at(thread_id_);
+
+  const Color kWhite(255, 255, 255, 255);
+
+  for (auto& pair : tracepoints) {
```
danielfenner, line 26: The tracepoints are sorted by time so there is no need to loop over all the events. Search for the first > min_tick and the last < max_tick and loop between those.
There is something like map::lower_bound/upper_bound that should do.

OrbitCore/CaptureData.h
MariaSandru27: Yes, of course. Thank you!

OrbitGl/TracepointTrack.cpp
MariaSandru27: Yes, I will that to loop from 'tracepoints.lower_bound(min_tick)' and stop when I find an event that has the time larger than max_tick. Thank you for the suggestion!

OrbitGl/TracepointTrack.cpp
```diff
@@ -0,0 +1,41 @@
...
+  float track_height = layout.GetEventTrackHeight();
+
+  ScopeLock lock(GOrbitApp->GetCaptureData().GetTracepointEventBufferMutex());
+
+  const std::map<uint64_t, orbit_client_protos::TracepointEventInfo>& tracepoints =
+      GOrbitApp->GetCaptureData().GetTracepointsOfThread(thread_id_);
+
+  const Color kWhite(255, 255, 255, 255);
+
+  for (auto it = tracepoints.lower_bound(min_tick); it != tracepoints.end(); ++it) {
```
danielfenner, line 26: Why not
`for (auto it = tracepoints.lower_bound(min_tick); it != tracepoints.upper_bound(max_tick); ++it) {`?

That would make the inner loop even denser.

OrbitCore/TracepointEventBuffer.h
```diff
@@ -24,7 +24,10 @@ class TracepointEventBuffer {
   void AddTracepointEventAndMapToThreads(uint64_t time, uint64_t tracepoint_hash,
                                          int32_t thread_id);
 
-  std::map<int32_t, std::map<uint64_t, orbit_client_protos::TracepointEventInfo> >
+  [[nodiscard]] const std::map<uint64_t, orbit_client_protos::TracepointEventInfo>&
```
danielfenner, line 27: I think you can delete "GetTracepointEventsInTimeRangeForThreadId" from class TracepointEventBuffer. This looks like it was supposed to do the stuff you are now doing in UpdatePrimitives on the map directly.

MariaSandru27: Sure. You are right. Thank you!

OrbitGl/TracepointTrack.cpp
MariaSandru27: Thank you! I will update the code.

OrbitCore/TracepointEventBuffer.h
MariaSandru27: Yes. You are right. Thank you!

OrbitGl/TracepointTrack.h
```diff
@@ -0,0 +1,29 @@
+// Copyright (c) 2020 The Orbit Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#ifndef ORBIT_GL_TRACEPOINT_TRACK_H_
+#define ORBIT_GL_TRACEPOINT_TRACK_H_
+
+#include "Track.h"
+
```
danielfenner, line 9: Add includes for the obvious dependencies. As discussed before we are probably not great at this in the code base but let's try to keep this somewhat in shape.

danielfenner (approved):
Minor nitpicking about incluides - otherwise lgtm!

OrbitGl/TracepointTrack.h
MariaSandru27: Thank you!

PR #1203, opened 2020-09-14
Dynamic string support for manual instrumentation

pierricgimmig:
Use the 6 integer registers visible from uprobes to encode
a tightly packed "orbit_api::Event" object. This allows us
to add support for dynamic strings for manual instrumentation
scopes (30 characters max).

Add "orbit_api::Event" class
- Contains the current orbit_api version in the first byte
- Enables dynamic strings support (up to 30 characters for now)
- Fits in exactly 6 integer registers (48 bytes)

Add "orbit_api::EncodedEvent" class
- union of a orbit_api::Event object and 6 uint64_t's
- Used to convert an orbit_api::Event object to and from 6 uint64_t's

### DIFF
--- a/Orbit.h
+++ b/Orbit.h
@@ -155,17 +155,18 @@ struct Event {
 union EncodedEvent {
   EncodedEvent(orbit_api::EventType type, const char* name = nullptr, uint64_t value = 0,
                orbit::Color color = orbit::Color::kAuto, uint32_t id = 0) {
-    static_assert(sizeof(EncodedEvent) == 48, "EncodedEvent should be 48 bytes.");
-    memset(this, 0, sizeof(*this));
-    event.id = id;
-    event.color = color;
-    event.value = value;
+    static_assert(sizeof(EncodedEvent) == 48, "orbit_api::EncodedEvent should be 48 bytes.");
+    static_assert(sizeof(Event) == 48, "orbit_api::Event should be 48 bytes.");
     event.version = kVersion;
     event.type = static_cast<uint8_t>(type);
+    memset(event.name, 0, kMaxEventStringSize);
     if (name != nullptr) {
       std::strncpy(event.name, name, kMaxEventStringSize - 1);
       event.name[kMaxEventStringSize - 1] = 0;
     }
+    event.value = value;
+    event.color = color;
+    event.id = id;
   }
 
   EncodedEvent(uint64_t a0, uint64_t a1, uint64_t a2, uint64_t a3, uint64_t a4, uint64_t a5) {
--- a/Orbit.h
+++ b/Orbit.h
@@ -15,7 +15,8 @@
 // instrumentation is still possible using the macros below. These macros call empty functions that
 // Orbit dynamically instruments.
 //
-// NOTE: We only support string literals for now. This is enforced through the ORBIT_STR() macro.
+// Note: We currently only support strings smaller than 30 characters, longer strings will be
+//       truncated. An upcoming update will allow for arbitrarily long strings.
 
 // To disable manual instrumentation macros, define ORBIT_API_ENABLED as 0.
 #define ORBIT_API_ENABLED 1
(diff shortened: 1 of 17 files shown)


ronaldfw:
Sorry for jumping on this early. Looks great, and glad we can support dynamic strings with this!

My initial question here was whether users of the API need to know about the 30 char length limitation when using dynamic strings, but I was able to clarify looking at the code (they don't have to know, strings get truncated). 

Two remarks:
- I think we should mark truncated strings in the UI clearly as such (I don't think this fits into this PR, but wanted to mention). 
- Do you see a way to increase the 30 char limit?

pierricgimmig:
> * I think we should mark truncated strings in the UI clearly as such (I don't think this fits into this PR, but wanted to mention).

Ah yes good point, I'll see how we can encode that information, one way would be to dedicate a byte to mark whether or not the string was truncated, all we'd need is a bit though...

> * Do you see a way to increase the 30 char limit?
Once we change to the new dynamic instrumentation implementation, there will be no limit for a dynamic string's length. For now we are limited by the 48 bytes available in integer registers. Although one thing we could do is to break down the long strings into chunks and emit multiple events, we'd then pay the overhead for every event but that would allow for arbitrarily long strings with our current dynamic implementation.

Orbit.h
```diff
@@ -129,49 +126,129 @@ enum class Color : uint32_t {
...
+  uint64_t value;                  // 8
+  orbit::Color color;              // 4
+  uint32_t id;                     // 4
+};
+
+union EncodedEvent {
+  EncodedEvent(orbit_api::EventType type, const char* name = nullptr, uint64_t value = 0,
+               orbit::Color color = orbit::Color::kAuto, uint32_t id = 0) {
+    static_assert(sizeof(EncodedEvent) == 48, "EncodedEvent should be 48 bytes.");
+    memset(this, 0, sizeof(*this));
```
dpallotti, line 159: Is this to clear `name` (the other values are set)? Can we explicitly clear only `name` then?

Orbit.h
```diff
@@ -129,49 +126,129 @@ enum class Color : uint32_t {
...
+  char name[kMaxEventStringSize];  // 30
+  uint64_t value;                  // 8
+  orbit::Color color;              // 4
+  uint32_t id;                     // 4
+};
+
+union EncodedEvent {
+  EncodedEvent(orbit_api::EventType type, const char* name = nullptr, uint64_t value = 0,
+               orbit::Color color = orbit::Color::kAuto, uint32_t id = 0) {
+    static_assert(sizeof(EncodedEvent) == 48, "EncodedEvent should be 48 bytes.");
```
dpallotti, line 158: I would think it's clearer to check the size of `Event` rather than the union. Or do we also want to support `Event` being shorter than 48 bytes?

OrbitGl/ManualInstrumentationManager.h
```diff
@@ -0,0 +1,34 @@
...
+// found in the LICENSE file.
+
+#ifndef ORBIT_GL_MANUAL_INSTRUMENTATION_MANAGER_H_
+#define ORBIT_GL_MANUAL_INSTRUMENTATION_MANAGER_H_
+
+#include "../Orbit.h"
+#include "OrbitBase/Logging.h"
+#include "capture_data.pb.h"
+
+class ManualInstrumentationManager {
```
dpallotti, line 12: This class is basically unused, as it doesn't have fields, `OnTimerInfo` is empty and unused and `ApiEventFromTimerInfo` is static. What is the planned usage? Other `On...` methods are part of `CaptureListener`, implemented by `OrbitApp`, and add to `CaptureData`, but this doesn't seem to fit in it.

OrbitGl/ThreadTrack.cpp
```diff
@@ -90,12 +90,12 @@ bool ThreadTrack::IsTimerActive(const TimerInfo& timer_info) const {
...
 
-  // See Orbit.h for more information about the manual instrumentation API.
-  constexpr int kColorArgumentIndex = 2;
-  constexpr uint64_t kColorAuto = 0x00000001;
-  CHECK(timer_info.registers_size() > kColorArgumentIndex);
-  uint64_t color_arg = timer_info.registers(kColorArgumentIndex);
-  return color_arg != kColorAuto ? ToColor(color_arg) : std::optional<Color>{};
+  orbit_api::Event event = ManualInstrumentationManager::ApiEventFromTimerInfo(timer_info);
+  if (event.color == orbit::Color::kAuto) {
+    return std::optional<Color>{};
```
dpallotti, line 95: I guess you can return `std::nullopt` to avoid specifying to full type.

Orbit.h
pierricgimmig: Yep, done!

Orbit.h
pierricgimmig: I'm checking for both now to be sure.

OrbitGl/ManualInstrumentationManager.h
pierricgimmig: Yes it's a bit weird in this PR, I should have put a note. I removed the unused method. The next PR populates the class properly.

OrbitGl/ThreadTrack.cpp
pierricgimmig: Done, thanks!

OrbitBase/OrbitApiTest.cpp
```diff
@@ -0,0 +1,33 @@
+// Copyright (c) 2020 The Orbit Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#include <gtest/gtest.h>
+
+#include "../Orbit.h"
+
+orbit_api::Event Decode(uint64_t a1, uint64_t a2, uint64_t a3, uint64_t a4, uint64_t a5,
```
dpallotti, line 9: Should this be `static`?

dpallotti (approved):

OrbitBase/OrbitApiTest.cpp
pierricgimmig: yes, done, thanks!

ronaldfw:
Sorry, seems I'm either too early or too late on this :)

I have two questions about the versioning: 

1) Events have a version but we don't seem to ever check which version an event has in the code that interprets the events (or did I miss that?). I guess that's not necessary as long as we only have one version. Where is the place were we need to fork based on event version?

2) I assume for support of older versions, we'll have to keep and maintain the decoding code for all versions so that manual instrumentation in binaries that use an old version will still show up in all future versions of Orbit. Correct? If yes, I think it would be good to document the procedure and rules around that: What steps do I have to do if I increase the version? (Not urgent to take care of right now.)

PR #1204, opened 2020-09-14
Renaming variables 

MariaSandru27:
Rename some variables to follow google3 C++ style 

The main reason for doing this, was to move the EventTrack's member variables to Track and help TracepointTrack inherit EventTrack. The Draw call for the TracepointTrack should be identical to the one in EventTrack so, TracepointTrack should inherit from EventTrack. The variables used in EventTrack::Draw were member variables of EventTrack that followed 'old' C++ standard. 

One solution for the TracepointTrack to use EventTrack::Draw via inheritance, would have been to also have the variables used in EventTrack::Draw as members of TracepointTrack. In order for TracepointTrack to be able to use EventTrack::Draw, the names of the member variable should have been identical to the ones in EventTrack, using 'old' C++.

Creating extra variables in C++ using 'old' style is not desirable, so instead refactoring was chosen. The variables used in EventTrack::Draw were moved to Track and renamed to follow google3 C++ style. Then, TracepointTrack was set to inherit EventTrack. Since the variables used in EventTrack were moved to one level higher, when TracepointTrack::Draw is called, the function EventTrack::Draw is called but with the variables specific to TracepointTrack.

A reason for renaming was that the variables in TracepointTrack::UpdatePrimitives use google3 C++ style. All those variables were moved in Track such that the variables used in TracepointTrack::UpdatePrimitives and Tracepoint::Draw are the same.

### DIFF
--- a/OrbitGl/EventTrack.cpp
+++ b/OrbitGl/EventTrack.cpp
@@ -12,9 +12,9 @@
 using orbit_client_protos::CallstackEvent;
 
 EventTrack::EventTrack(TimeGraph* a_TimeGraph) : Track(a_TimeGraph) {
-  m_MousePos[0] = m_MousePos[1] = Vec2(0, 0);
-  m_Picked = false;
-  m_Color = Color(0, 255, 0, 255);
+  mouse_pos_[0] = mouse_pos_[1] = Vec2(0, 0);
+  picked_ = false;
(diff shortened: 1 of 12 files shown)


dimitry- (approved):

danielfenner (approved):
lgtm

PR #1205, opened 2020-09-14
Add Cancellation and Error Handling to CaptureListener

florian-kuebler:
On capture loading there might be the cases of errors or user cancellation.
Prior to this change, those might not be handled thread-safe. Now, the capture
thread properly calls respective methods on the capture listener.
These callbacks are currently only called on capture loading and not on usual
capturing.

Test: Compile, Loading Captures (also broken ones). Pressed "cancel" during loading.

### DIFF
--- a/OrbitCaptureClient/CaptureClient.cpp
+++ b/OrbitCaptureClient/CaptureClient.cpp
@@ -6,8 +6,10 @@
 
 #include "FunctionUtils.h"
 #include "OrbitBase/Logging.h"
+#include "OrbitBase/Result.h"
 #include "OrbitCaptureClient/CaptureEventProcessor.h"
 #include "absl/flags/flag.h"
+#include "absl/strings/str_format.h"
 
 ABSL_DECLARE_FLAG(uint16_t, sampling_rate);
 ABSL_DECLARE_FLAG(bool, frame_pointer_unwinding);
@@ -98,6 +100,7 @@ void CaptureClient::Capture(int32_t process_id, std::string process_name,
     ERROR("Sending CaptureRequest on Capture's gRPC stream");
     reader_writer_->WritesDone();
     FinishCapture();
+    capture_listener_->OnCaptureFailed(ErrorMessage("Error sending capture request."));
     return;
   }
   LOG("Sent CaptureRequest on Capture's gRPC stream: asking to start "
@@ -147,6 +150,7 @@ bool CaptureClient::StopCapture() {
     ERROR(
         "WritesDone on Capture's gRPC stream failed: unable to finish the "
         "capture in orderly manner, initiating emergency stop");
+    capture_listener_->OnCaptureFailed(ErrorMessage("Error while stopping capture."));
     force_stop_ = true;
   } else {
     LOG("Finished writing on Capture's gRPC stream: asking to stop capturing");
@@ -164,6 +168,8 @@ void CaptureClient::FinishCapture() {
   grpc::Status status = reader_writer_->Finish();
   if (!status.ok()) {
     ERROR("Finishing gRPC call to Capture: %s", status.error_message());
+    capture_listener_->OnCaptureFailed(
+        ErrorMessage(absl::StrFormat("Error while finishing capture: %s", status.error_message())));
   }
   reader_writer_.reset();
 
--- a/OrbitCaptureClient/CaptureClient.cpp
+++ b/OrbitCaptureClient/CaptureClient.cpp
@@ -123,7 +123,7 @@ void CaptureClient::Capture(int32_t process_id, std::string process_name,
   if (force_stop_) {
     capture_listener_->OnCaptureFailed(
         ErrorMessage("WritesDone on Capture's gRPC stream failed: unable to finish the "
-                     "capture in orderly manner, initiating emergency stop"));
+                     "capture in orderly manner, performing emergency stop."));
   } else {
     LOG("Finished reading from Capture's gRPC stream: all capture data has been received");
     capture_listener_->OnCaptureComplete();
--- a/OrbitCaptureClient/CaptureClient.cpp
+++ b/OrbitCaptureClient/CaptureClient.cpp
@@ -120,10 +120,14 @@ void CaptureClient::Capture(int32_t process_id, std::string process_name,
   while (!force_stop_ && reader_writer_->Read(&response)) {
     event_processor.ProcessEvents(response.capture_events());
   }
-  LOG("Finished reading from Capture's gRPC stream: all capture data has been "
-      "received");
-
-  capture_listener_->OnCaptureComplete();
+  if (force_stop_) {
+    capture_listener_->OnCaptureFailed(
+        ErrorMessage("WritesDone on Capture's gRPC stream failed: unable to finish the "
+                     "capture in orderly manner, initiating emergency stop"));
+  } else {
+    LOG("Finished reading from Capture's gRPC stream: all capture data has been received");
+    capture_listener_->OnCaptureComplete();
+  }
   FinishCapture();
 }
 
@@ -150,7 +154,6 @@ bool CaptureClient::StopCapture() {
     ERROR(
         "WritesDone on Capture's gRPC stream failed: unable to finish the "
         "capture in orderly manner, initiating emergency stop");
-    capture_listener_->OnCaptureFailed(ErrorMessage("Error while stopping capture."));
     force_stop_ = true;
   } else {
     LOG("Finished writing on Capture's gRPC stream: asking to stop capturing");
@@ -168,8 +171,6 @@ void CaptureClient::FinishCapture() {
   grpc::Status status = reader_writer_->Finish();
   if (!status.ok()) {
     ERROR("Finishing gRPC call to Capture: %s", status.error_message());
-    capture_listener_->OnCaptureFailed(
-        ErrorMessage(absl::StrFormat("Error while finishing capture: %s", status.error_message())));
   }
   reader_writer_.reset();
 
--- a/OrbitCaptureClient/CaptureEventProcessorProcessEventsFuzzer.cpp
+++ b/OrbitCaptureClient/CaptureEventProcessorProcessEventsFuzzer.cpp
@@ -27,6 +27,8 @@ class MyCaptureListener : public CaptureListener {
       absl::flat_hash_map<uint64_t, orbit_client_protos::FunctionInfo> /*selected_functions*/,
       TracepointInfoSet /*selected_tracepoints*/) override {}
   void OnCaptureComplete() override {}
+  void OnCaptureCanceled() override {}
+  void OnCaptureFailed(std::string /*error_message*/) override {}
   void OnTimer(const TimerInfo&) override {}
   void OnKeyAndString(uint64_t, std::string) override {}
   void OnUniqueCallStack(CallStack) override {}
--- a/OrbitCaptureClient/CaptureEventProcessorProcessEventsFuzzer.cpp
+++ b/OrbitCaptureClient/CaptureEventProcessorProcessEventsFuzzer.cpp
@@ -4,6 +4,7 @@
 
 #include <libfuzzer/libfuzzer_macro.h>
 
+#include "OrbitBase/Result.h"
 #include "OrbitCaptureClient/CaptureEventProcessor.h"
 #include "OrbitCaptureClient/CaptureListener.h"
 #include "TracepointCustom.h"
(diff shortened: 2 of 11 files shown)


OrbitCaptureClient/CaptureEventProcessorProcessEventsFuzzer.cpp
```diff
@@ -27,6 +27,8 @@ class MyCaptureListener : public CaptureListener {
       absl::flat_hash_map<uint64_t, orbit_client_protos::FunctionInfo> /*selected_functions*/,
       TracepointInfoSet /*selected_tracepoints*/) override {}
   void OnCaptureComplete() override {}
+  void OnCaptureCanceled() override {}
+  void OnCaptureFailed(std::string /*error_message*/) override {}
```
dimitry-, line 31: ErrorMessage?

OrbitGl/App.h
```diff
@@ -81,7 +81,7 @@ class OrbitApp final : public DataViewFactory, public CaptureListener {
   ErrorMessageOr<void> OnLoadPreset(const std::string& file_name);
   ErrorMessageOr<void> OnSaveCapture(const std::string& file_name);
   void OnLoadCapture(const std::string& file_name);
-  void OnLoadCatpureCanceled();
+  void OnLoadCaptureCancellationRequest();
```
dimitry-, line 84: CancelRequested?

OrbitGl/App.cpp
```diff
@@ -645,17 +662,7 @@ void OrbitApp::OnLoadCapture(const std::string& file_name) {
...
-        open_capture_failed_callback_();
-      }
-      SendErrorToUi("Error loading capture",
-                    absl::StrFormat("Could not load capture from \"%s\":\n%s", file_name,
-                                    result.error().message()));
-    }
+    capture_deserializer::Load(file_name, this, &capture_loading_cancellation_requested_);
 
     if (open_capture_finished_callback_) {
       open_capture_finished_callback_();
```
dimitry-, line 668: Should this be called from OnCaptureComplete?

OrbitGl/App.cpp
```diff
@@ -168,6 +168,23 @@ void OrbitApp::OnCaptureComplete() {
...
+      open_capture_failed_callback_();
+    }
+    ClearCapture();
+  });
+}
+
+void OrbitApp::OnCaptureFailed(std::string error_message) {
+  if (open_capture_failed_callback_) {
+    open_capture_failed_callback_();
+  }
```
dimitry-, line 183: Should this also be called on the main thread?

dimitry- (commented):
What about CaptureClient? Should it also be changed to call OnCaptureFailed where appropriate?

OrbitGl/App.cpp
florian-kuebler: Well that basically calls `loading_capture_dialog->close();`. Even though it does not hurt if this is called also for normal captures, it would feel a bit strange. Also we need to call this here for either case (successful load, error, cancellation).

OrbitGl/App.cpp
florian-kuebler: There is no need to. But if you think it is more clear, then I can move it to the main thread.

florian-kuebler:
> What about CaptureClient? Should it also be changed to call OnCaptureFailed where appropriate?

Not yet, lets move this into a follow up PR.

OrbitGl/App.h
florian-kuebler: Done. Thanks!

OrbitCaptureClient/CaptureEventProcessorProcessEventsFuzzer.cpp
florian-kuebler: You are right. This is better. Done.

OrbitGl/App.cpp
florian-kuebler: I moved this anyway.

OrbitGl/App.cpp
florian-kuebler: I moved this on the main thread, as I am now also clearing the capture on failure.

OrbitCaptureClient/CaptureEventProcessorProcessEventsFuzzer.cpp
```diff
@@ -27,6 +27,8 @@ class MyCaptureListener : public CaptureListener {
       absl::flat_hash_map<uint64_t, orbit_client_protos::FunctionInfo> /*selected_functions*/,
       TracepointInfoSet /*selected_tracepoints*/) override {}
   void OnCaptureComplete() override {}
+  void OnCaptureCanceled() override {}
```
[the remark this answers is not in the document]
florian-kuebler: Done. Thanks.

OrbitClientGgp/ClientGgp.cpp
```diff
@@ -249,6 +249,10 @@ void ClientGgp::OnCaptureComplete() {
   capture_data_.set_sampling_profiler(sampling_profiler);
 }
 
+void ClientGgp::OnCaptureCanceled(){};
+
+void ClientGgp::OnCaptureFailed(std::string /*error_message*/){};
```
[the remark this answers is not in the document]
florian-kuebler: Done in the last commit. Thanks!

OrbitClientModel/CaptureDeserializer.cpp
```diff
@@ -25,53 +25,57 @@ using orbit_client_protos::TimerInfo;
...
 
   // Binary
-  std::ifstream file(filename, std::ios::binary);
+  std::ifstream file(file_name, std::ios::binary);
   if (file.fail()) {
-    ERROR("Loading capture from \"%s\": %s", filename, "file.fail()");
-    return ErrorMessage("Error opening the file for reading");
+    ERROR("Loading capture from \"%s\": %s", file_name, "file.fail()");
+    capture_listener->OnCaptureFailed(
+        absl::StrFormat("Error opening the file \"%s\"for reading", file_name));
```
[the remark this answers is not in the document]
florian-kuebler: Done in the last commit. Thanks!

florian-kuebler:
Thanks for the reviews so far. I updated the PR to address the comments.

dimitry-:
> > What about CaptureClient? Should it also be changed to call OnCaptureFailed where appropriate?
> 
> Not yet, lets move this into a follow up PR.

Why not in this one?

florian-kuebler:
> > > What about CaptureClient? Should it also be changed to call OnCaptureFailed where appropriate?
> > 
> > 
> > Not yet, lets move this into a follow up PR.
> 
> Why not in this one?

Basically, because I need to understand what is going on in CaptureClient, as it is not returning any errors yet (only has some error logs).

dimitry- (approved):

OrbitCaptureClient/CaptureEventProcessorProcessEventsFuzzer.cpp
```diff
@@ -27,8 +28,8 @@ class MyCaptureListener : public CaptureListener {
       absl::flat_hash_map<uint64_t, orbit_client_protos::FunctionInfo> /*selected_functions*/,
       TracepointInfoSet /*selected_tracepoints*/) override {}
   void OnCaptureComplete() override {}
-  void OnCaptureCanceled() override {}
-  void OnCaptureFailed(std::string /*error_message*/) override {}
+  void OnCaptureCancelled() override {}
+  void OnCaptureFailed(ErrorMessage /*error_message*/) override {}
```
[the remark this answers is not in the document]
florian-kuebler: Done.

florian-kuebler:
> > > > What about CaptureClient? Should it also be changed to call OnCaptureFailed where appropriate?
> > > 
> > > 
> > > Not yet, lets move this into a follow up PR.
> > 
> > 
> > Why not in this one?
> 
> Basically, because I need to understand what is going on in CaptureClient, as it is not returning any errors yet (only has some error logs).

I just updated the PR with also using OnCaptureFailed in capture client. @dimitry-  could you have another brief look?

OrbitCaptureClient/CaptureClient.cpp
```diff
@@ -98,6 +100,7 @@ void CaptureClient::Capture(int32_t process_id, std::string process_name,
     ERROR("Sending CaptureRequest on Capture's gRPC stream");
     reader_writer_->WritesDone();
     FinishCapture();
+    capture_listener_->OnCaptureFailed(ErrorMessage("Error sending capture request."));
```
dimitry-, line 103: FinishCapture can aslo generate failure.

OrbitCaptureClient/CaptureClient.cpp
```diff
@@ -164,6 +168,8 @@ void CaptureClient::FinishCapture() {
   grpc::Status status = reader_writer_->Finish();
   if (!status.ok()) {
     ERROR("Finishing gRPC call to Capture: %s", status.error_message());
+    capture_listener_->OnCaptureFailed(
+        ErrorMessage(absl::StrFormat("Error while finishing capture: %s", status.error_message())));
```
dimitry-, line 172: I think we want to not report this. In one case it was already reported in another - it is not really relevant since the capture was already complete.

OrbitCaptureClient/CaptureClient.cpp
```diff
@@ -147,6 +150,7 @@ bool CaptureClient::StopCapture() {
     ERROR(
         "WritesDone on Capture's gRPC stream failed: unable to finish the "
         "capture in orderly manner, initiating emergency stop");
+    capture_listener_->OnCaptureFailed(ErrorMessage("Error while stopping capture."));
```
dimitry-, line 153: Use the same error message as in ERROR? And note that this function is called off the capture thread - we generally do not want to call listener outside of capture thread - maybe handle this in capture thread?

OrbitCaptureClient/CaptureClient.cpp
florian-kuebler: Okay, I'll remove this call

OrbitCaptureClient/CaptureClient.cpp
florian-kuebler: Should be not the case anymore, I i remove the error in FinishCapture as in https://github.com/google/orbit/pull/1205#discussion_r488711100

OrbitCaptureClient/CaptureClient.cpp
florian-kuebler: Thanks. I moved that to the capture thread and used the same string here.

OrbitCaptureClient/CaptureClient.cpp
```diff
@@ -120,10 +120,14 @@ void CaptureClient::Capture(int32_t process_id, std::string process_name,
...
     event_processor.ProcessEvents(response.capture_events());
   }
-  LOG("Finished reading from Capture's gRPC stream: all capture data has been "
-      "received");
-
-  capture_listener_->OnCaptureComplete();
+  if (force_stop_) {
+    capture_listener_->OnCaptureFailed(
+        ErrorMessage("WritesDone on Capture's gRPC stream failed: unable to finish the "
+                     "capture in orderly manner, initiating emergency stop"));
```
dimitry-, line 126: In this case the emergency stop is just happened. s/initiating emergency stop/capture halted/?

dimitry- (approved):
otherwise lgtm

OrbitCaptureClient/CaptureClient.cpp
florian-kuebler: Done.

OrbitGl/App.cpp
```diff
@@ -163,11 +163,39 @@ void OrbitApp::OnCaptureComplete() {
...
         FireRefreshCallbacks();
       });
 }
 
+void OrbitApp::OnCaptureCancelled() {
+  main_thread_executor_->Schedule([this]() mutable {
+    if (capture_failed_callback_) {
+      capture_failed_callback_();
+    }
+    if (open_capture_failed_callback_) {
```
dpallotti, line 179: I guess the result is correct as the `open_capture_failed_callback_` is effectively just `loading_capture_dialog->close();`, but conceptually it's confusing that both `capture_failed_callback_` and `open_capture_failed_callback_` are called here (they are always set), regardless of whether the capture is being taken or it's being loaded.
Same in `OnCaptureFailed` at line 191 and similarly in `OnCaptureComplete` at line 166 with `open_capture_finished_callback_`.

dpallotti (commented):
The rest looks good, up to you to decide if it's easy/feasible to address the comment I left.

OrbitGl/App.cpp
florian-kuebler: That was part of the idea of combining the handling of loading/capturing. I admit, that it is not ideal yet. Long term it should be fine as soon as we don't have the progress bar for loading anymore. I'll leave it as is for now, and revisit if we have any issues with this.

PR #1206, opened 2020-09-14
Filter events

MariaSandru27:
In the tracepoint track responsible for all the threads running on the process being profiled (such as OrbitALL) , filter the tracepoint events based on process id. This is necessary since the events are collected system-wide not process specific. Without this filtering, on that track would be displayed all the tracepoints in the whole system instead of for the process being profiled.

### DIFF
--- a/OrbitClientGgp/ClientGgp.cpp
+++ b/OrbitClientGgp/ClientGgp.cpp
@@ -292,5 +292,5 @@ void ClientGgp::OnTracepointEvent(orbit_client_protos::TracepointEventInfo trace
 
   capture_data_.AddTracepointEventAndMapToThreads(
       tracepoint_event_info.time(), tracepoint_event_info.tracepoint_info_key(),
-      tracepoint_event_info.tid(), tracepoint_event_info.pid(), is_same_pid_as_target);
+      tracepoint_event_info.pid(), tracepoint_event_info.tid(), is_same_pid_as_target);
 }
--- a/OrbitClientGgp/ClientGgp.cpp
+++ b/OrbitClientGgp/ClientGgp.cpp
@@ -287,7 +287,11 @@ void ClientGgp::OnUniqueTracepointInfo(uint64_t key,
 }
 
 void ClientGgp::OnTracepointEvent(orbit_client_protos::TracepointEventInfo tracepoint_event_info) {
-  capture_data_.AddTracepointEventAndMapToThreads(tracepoint_event_info.time(),
-                                                  tracepoint_event_info.tracepoint_info_key(),
-                                                  tracepoint_event_info.tid());
+  int32_t capture_process_id = capture_data_.process_id();
+  bool is_same_pid_as_target =
+      capture_process_id == 0 || capture_process_id == tracepoint_event_info.pid();
+
+  capture_data_.AddTracepointEventAndMapToThreads(
+      tracepoint_event_info.time(), tracepoint_event_info.tracepoint_info_key(),
+      tracepoint_event_info.tid(), is_same_pid_as_target);
 }
--- a/OrbitClientGgp/ClientGgp.cpp
+++ b/OrbitClientGgp/ClientGgp.cpp
@@ -288,10 +288,9 @@ void ClientGgp::OnUniqueTracepointInfo(uint64_t key,
 
 void ClientGgp::OnTracepointEvent(orbit_client_protos::TracepointEventInfo tracepoint_event_info) {
   int32_t capture_process_id = capture_data_.process_id();
-  bool is_same_pid_as_target =
-      capture_process_id == 0 || capture_process_id == tracepoint_event_info.pid();
+  bool is_same_pid_as_target = capture_process_id == tracepoint_event_info.pid();
 
   capture_data_.AddTracepointEventAndMapToThreads(
       tracepoint_event_info.time(), tracepoint_event_info.tracepoint_info_key(),
-      tracepoint_event_info.tid(), is_same_pid_as_target);
+      tracepoint_event_info.tid(), tracepoint_event_info.pid(), is_same_pid_as_target);
 }
--- a/OrbitCore/CaptureData.h
+++ b/OrbitCore/CaptureData.h
@@ -139,10 +139,11 @@ class CaptureData {
     tracepoint_info_manager_->AddUniqueTracepointEventInfo(key, std::move(tracepoint_info));
   }
 
-  void AddTracepointEventAndMapToThreads(uint64_t time, uint64_t tracepoint_hash, int32_t thread_id,
-                                         int32_t process_id, bool is_same_pid_as_target) {
-    tracepoint_event_buffer_->AddTracepointEventAndMapToThreads(time, tracepoint_hash, thread_id,
-                                                                process_id, is_same_pid_as_target);
+  void AddTracepointEventAndMapToThreads(uint64_t time, uint64_t tracepoint_hash,
+                                         int32_t process_id, int32_t thread_id,
+                                         bool is_same_pid_as_target) {
+    tracepoint_event_buffer_->AddTracepointEventAndMapToThreads(time, tracepoint_hash, process_id,
+                                                                thread_id, is_same_pid_as_target);
   }
 
   [[nodiscard]] const CallstackData* GetSelectionCallstackData() const {
--- a/OrbitCore/CaptureData.h
+++ b/OrbitCore/CaptureData.h
@@ -139,9 +139,10 @@ class CaptureData {
     tracepoint_info_manager_->AddUniqueTracepointEventInfo(key, std::move(tracepoint_info));
   }
 
-  void AddTracepointEventAndMapToThreads(uint64_t time, uint64_t tracepoint_hash,
-                                         int32_t thread_id) {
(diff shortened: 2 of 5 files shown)


OrbitCore/TracepointEventBuffer.cpp
```diff
@@ -8,7 +8,8 @@ using orbit_client_protos::TracepointEventInfo;
 
 void TracepointEventBuffer::AddTracepointEventAndMapToThreads(uint64_t time,
                                                               uint64_t tracepoint_hash,
-                                                              int32_t thread_id) {
+                                                              int32_t thread_id,
+                                                              bool is_same_pid_as_target) {
   ScopeLock lock(mutex_);
   std::map<uint64_t, orbit_client_protos::TracepointEventInfo>& event_map =
       tracepoint_events_[thread_id];
```
dimitry-, line 15: so if this tracepoint event is for different process do we want to store it here?

dimitry- (commented):
question: do you want to filter them out on the service side? Or are we interested in tracepoint events for other processes as well? I do not really see how we would display them.

ronaldfw:
> question: do you want to filter them out on the service side? Or are we interested in tracepoint events for other processes as well? I do not really see how we would display them.

We discussed this last week: We want to have another track ("all processes") for all tracepoints from all processes and visually distinguish those that are from processes other than the target process; e.g. grayed out / with more transparency / dashed line. WDYT?

OrbitCore/TracepointEventBuffer.cpp
MariaSandru27: Yes. After a discussion with @dpallotti, I was informed that the events should be collected system-wide in the collector because you might hook a tracepoint and since it is not the selected process, it will not be displayed. @dpallotti maybe could give more details about it.

OrbitCore/TracepointEventBuffer.cpp
```diff
@@ -18,15 +19,16 @@ void TracepointEventBuffer::AddTracepointEventAndMapToThreads(uint64_t time,
...
-  event_map_all_threads[time] = std::move(event);
-
+  if (is_same_pid_as_target) {
+    // Add all tracepoint events to "all threads".
+    std::map<uint64_t, orbit_client_protos::TracepointEventInfo>& event_map_all_threads =
+        tracepoint_events_[SamplingProfiler::kAllThreadsFakeTid];
+    orbit_client_protos::TracepointEventInfo event_all_threads;
+    event_all_threads.set_time(time);
+    event_all_threads.set_tracepoint_info_key(tracepoint_hash);
+    event_all_threads.set_tid(SamplingProfiler::kAllThreadsFakeTid);
```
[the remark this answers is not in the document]
MariaSandru27: Sure. Thanks for the suggestion!

dimitry- (approved):

dpallotti:
> question: do you want to filter them out on the service side? Or are we interested in tracepoint events for other processes as well? I do not really see how we would display them.

We had a lengthy discussion on this on Friday. Since this feature allows to choose *any* tracepoint, there are for sure at least a few that are sometimes or always associated with a kernel process, or that one might be interested in even when not directly associated with the target process. One such example (for the "sometimes" case) is context switches: probably uninteresting as we have a different mechanism for scheduling slices, but if there's one, there are probably many. Filtering them means hiding potentially interesting information when such a tracepoint is chosen. The idea is, in a second step, to have a dedicated track for tracepoints from all processes, as Ronald described.

OrbitClientGgp/ClientGgp.cpp
```diff
@@ -287,7 +287,11 @@ void ClientGgp::OnUniqueTracepointInfo(uint64_t key,
 }
 
 void ClientGgp::OnTracepointEvent(orbit_client_protos::TracepointEventInfo tracepoint_event_info) {
-  capture_data_.AddTracepointEventAndMapToThreads(tracepoint_event_info.time(),
-                                                  tracepoint_event_info.tracepoint_info_key(),
-                                                  tracepoint_event_info.tid());
+  int32_t capture_process_id = capture_data_.process_id();
+  bool is_same_pid_as_target =
+      capture_process_id == 0 || capture_process_id == tracepoint_event_info.pid();
```
dpallotti, line 292: Can `capture_process_id == 0` happen?

OrbitCore/TracepointEventBuffer.cpp
```diff
@@ -18,15 +19,16 @@ void TracepointEventBuffer::AddTracepointEventAndMapToThreads(uint64_t time,
...
-  std::map<uint64_t, orbit_client_protos::TracepointEventInfo>& event_map_all_threads =
-      tracepoint_events_[SamplingProfiler::kAllThreadsFakeTid];
-  orbit_client_protos::TracepointEventInfo event_all_threads;
-  event_all_threads.set_time(time);
-  event_all_threads.set_tracepoint_info_key(tracepoint_hash);
-  event_all_threads.set_tid(SamplingProfiler::kAllThreadsFakeTid);
-  event_map_all_threads[time] = std::move(event);
-
+  if (is_same_pid_as_target) {
+    // Add all tracepoint events to "all threads".
```
dpallotti, line 23: This comment is now outdated as it's not "all tracepoint events" but only the ones related to the target process.

OrbitCore/TracepointEventBuffer.cpp
```diff
@@ -18,15 +19,16 @@ void TracepointEventBuffer::AddTracepointEventAndMapToThreads(uint64_t time,
   event.set_tid(thread_id);
```
dpallotti, line 19: `TracepointEventInfo` also has a `pid` field, but it's not set. It should probably be set as events are collected system wide and we will probably be interested in the `pid` as well (which is not always the target pid).

OrbitCore/TracepointEventBuffer.cpp
dpallotti: When we show events from all processes, we will want a special map for events from all processes, similar to the `event_map_all_threads` that we now have for all threads of the target process. But I don't think that for events from processes different from the target we should group them by thread here (they will all be displayed in one single track). So in other words, I would put this code also in `if (is_same_pid_as_target) {`. Later, when we support events from all threads, we should have the dedicated code outside of such `if`. Does it make sense?

MariaSandru27: Of course! Thank you so much for the suggestion.

OrbitCore/TracepointEventBuffer.cpp
MariaSandru27: Of course. Thank you!

OrbitCore/TracepointEventBuffer.cpp
MariaSandru27: Thanks for the suggestion! Will update the code.

OrbitCore/TracepointEventBuffer.h
```diff
@@ -17,8 +17,8 @@ class TracepointEventBuffer {
  public:
   TracepointEventBuffer() : max_time_(0), min_time_(std::numeric_limits<uint64_t>::max()) {}
 
-  void AddTracepointEventAndMapToThreads(uint64_t time, uint64_t tracepoint_hash,
-                                         int32_t thread_id);
+  void AddTracepointEventAndMapToThreads(uint64_t time, uint64_t tracepoint_hash, int32_t thread_id,
+                                         int32_t process_id, bool is_same_pid_as_target);
```
dpallotti, line 21: Have the `process_id` parameter before the `thread_id` parameter as that's always the order in which these two fields are handled when together.

OrbitCore/TracepointEventBuffer.cpp
```diff
@@ -8,23 +8,29 @@ using orbit_client_protos::TracepointEventInfo;
 
 void TracepointEventBuffer::AddTracepointEventAndMapToThreads(uint64_t time,
                                                               uint64_t tracepoint_hash,
-                                                              int32_t thread_id) {
+                                                              int32_t thread_id, int32_t process_id,
+                                                              bool is_same_pid_as_target) {
+  if (!is_same_pid_as_target) {
+    return;
```
dpallotti, line 14: Maybe add a `TODO` here explaining that tracepoint events with `!is_same_pid_as_target` will also have to be stored when implementing the track showing tracepoint events from all processes in the system.

OrbitCore/CaptureData.h
```diff
@@ -139,9 +139,10 @@ class CaptureData {
     tracepoint_info_manager_->AddUniqueTracepointEventInfo(key, std::move(tracepoint_info));
   }
 
-  void AddTracepointEventAndMapToThreads(uint64_t time, uint64_t tracepoint_hash,
-                                         int32_t thread_id) {
-    tracepoint_event_buffer_->AddTracepointEventAndMapToThreads(time, tracepoint_hash, thread_id);
+  void AddTracepointEventAndMapToThreads(uint64_t time, uint64_t tracepoint_hash, int32_t thread_id,
+                                         int32_t process_id, bool is_same_pid_as_target) {
```
dpallotti, line 143: Here as well, have the `process_id` parameter before the `thread_id` parameter.

dpallotti (approved):
Looks good otherwise.

danielfenner (approved):
lgtm!

PR #1207, opened 2020-09-14
Don't show percentage "Of parent" for a thread in the top-down view

dpallotti:
For a thread, this is the same as the "Inclusive" percentage, i.e. the number of
samples collected from this thread over the total number of samples. It's
confusing to have the value twice.


ronaldfw (approved):
Makes sense, LGTM!

PR #1208, opened 2020-09-14
Refactor to not use OrbitProcess/OrbitModule/Pdb anymore

antonrohr:
This PR is split up in 3 commits. 

1. Move still needed functionality from OrbitProcess/Module to ProcessData/ModuleData (also moved/added tests)
* I introduced a small struct (MemorySpace) that represents a region in memory where the process loaded a module. This is to decouple process and module more (They are still tighter coupled than I would like, since FunctionInfo contains the address at which the module was loaded by the process). In the future this struct should be used when we add support for dlopen/dlclose while taking a capture.

2. Moved ProcessData, ModuleData and FunctionUtils into new component OrbitClientData. Beforehand Process/ModuleData lived in OrbitGl, hence OrbitCore wasn't able to use them. The new component OrbitClientData is a dependency of OrbitGl and OrbitCore

3. Use Process/ModuleData instead of OrbitProcess/Module everywhere. This required a bunch of smaller changes in a lot of places.

@florian-kuebler & @dimitry-: I tried to make thread locality more obvious by actually creating a copy of ProcessData (and a memory map) for CaptureData. The idea is that CaptureData is for the capture thread and the main thread uses DataManager. Modules (ModuleData) and therefore Symbols live in DataManager and access is synchronised via a mutex in ModuleData (instead of copying all symbol data into the thread)

@kisenshi: I added you, because I did not test the changes I made to ClientGgp (I only know it compiles). Of course I am happy to get a full review, but feel free to just look at the ClientGgp changes


OrbitGl/ProcessData.cpp
```diff
@@ -0,0 +1,93 @@
...
+
+  return std::make_pair(module_path, memory_space->start);
+}
+
+void ProcessData::LoadSymbols(ModuleData* module, const ModuleSymbols& module_symbols) const {
+  uint64_t module_start_address = module_memory_map_.at(module->file_path())->start;
+  module->LoadSymbols(module_symbols, module_start_address);
+}
+
+std::unique_ptr<ProcessData> ProcessData::CreateCopy() const {
```
florian-kuebler: What is the use-case here? Shouldn't we aim for just making ProcessData copyable?

OrbitGl/ProcessData.h
```diff
@@ -5,13 +5,27 @@
...
 #include "module.pb.h"
 #include "process.pb.h"
+#include "symbol.pb.h"
+
+// Small struct to model a space in memory occupied by a module.
+struct MemorySpace {
+  explicit MemorySpace(uint64_t start, uint64_t end) : start(start), end(end) {}
+  uint64_t start;
+  uint64_t end;
+  [[nodiscard]] std::string AddressRange() const {
```
florian-kuebler: Can you name it something like `AddressRangeString` or `FormattedAddressRange`?

OrbitGl/ProcessData.h
```diff
@@ -28,48 +42,36 @@ class ProcessData final {
...
 
  private:
   orbit_grpc_protos::ProcessInfo process_info_;
 
-  // This list omits unloaded modules, note that they are still stored
-  // in the map.
-  std::vector<ModuleData*> current_module_list_;
-  absl::flat_hash_map<uint64_t, std::unique_ptr<ModuleData>> modules_;
+  // This is a map from module_path to the space in memory where that module is loaded
+  absl::flat_hash_map<std::string, std::unique_ptr<MemorySpace>> module_memory_map_;
```
florian-kuebler: Is there any need to explicitly heap allocate the MemorySpace objects here? Could't you just store them directly in that map?

florian-kuebler (commented):
Thank you so much for that change. That is a huge step in the correct direction. I only reviewed the first commit, yet, but left a couple of minor comments and some questions.

kisenshi:
I couldn't take a look yet but just wondering if it would make sense to use _OrbitClientModel_ to allocate the data files instead of creating _OrbitClientData_

OrbitGl/App.cpp
```diff
@@ -937,79 +925,98 @@ ErrorMessageOr<std::filesystem::path> OrbitApp::FindSymbolsLocally(
...
+    return ErrorMessage(absl::StrFormat("* %d preset functions missing from module \"%s\"\n",
+                                        count_missing, module->file_path()));
+  }
+  return outcome::success();
+}
+
+void OrbitApp::LoadModulesFromPreset(const ProcessData* process,
+                                     const std::shared_ptr<PresetFile>& preset_file) {
+  std::vector<ModuleData*> modules_to_load;
+  std::vector<std::string> modules_not_found;  // file path of module
```
florian-kuebler, line 983: just name it `module_paths_not_found` ?

OrbitGl/CallStackDataView.cpp
```diff
@@ -192,15 +194,9 @@ CallStackDataView::CallStackDataViewFrame CallStackDataView::GetFrameFromIndex(
...
-
-  const std::shared_ptr<Process> process = GOrbitApp->GetCaptureData().process();
-  CHECK(process != nullptr);
-  // TODO(kuebler): Get rid of locks on caller side
-  ScopeLock lock(process->GetDataMutex());
-  function = process->GetFunctionFromAddress(address, false);
-  module = process->GetModuleFromAddress(address);
+
+  const FunctionInfo* function = GOrbitApp->GetCaptureData().FindFunctionByAddress(address, false);
+  ModuleData* module = GOrbitApp->GetCaptureData().FindModuleByAddress(address);
```
florian-kuebler, line 199: can this also be const?

OrbitGl/ModulesDataView.h
```diff
@@ -28,15 +28,18 @@ class ModulesDataView : public DataView {
   bool HasRefreshButton() const override { return true; }
   void OnRefreshButtonClicked() override;
 
-  void SetProcess(ProcessData* process);
+  // TODO(antonrohr) make const as soon as process_ is gone
```
florian-kuebler, line 31: don't get the comment. the param is already const, and a const setter sounds strange. Can you rephrase, if it is still relevant?

OrbitGl/ModulesDataView.h
```diff
@@ -28,15 +28,18 @@ class ModulesDataView : public DataView {
...
 
  protected:
   void DoSort() override;
   void DoFilter() override;
 
  private:
-  [[nodiscard]] const ModuleData* GetModule(uint32_t row) const { return modules_[indices_[row]]; }
+  [[nodiscard]] ModuleData* GetModule(uint32_t row) const { return modules_[indices_[row]]; }
 
+  // TODO(antonrohr) remove once GOrbitApp->LoadSymbols does not need process anymore
```
florian-kuebler, line 41: That comment also seems to be outdated. If not, please rephrase or move to a different position.

OrbitGl/FunctionsDataView.cpp
```diff
@@ -145,39 +138,30 @@ void FunctionsDataView::OnContextMenu(const std::string& action, int menu_index,
...
-      GOrbitApp->Disassemble(pid, GetFunction(i));
+      GOrbitApp->Disassemble(GOrbitApp->GetSelectedProcess()->pid(), *GetFunction(i));
     }
   } else {
     DataView::OnContextMenu(action, menu_index, item_indices);
   }
 }
 
 void FunctionsDataView::DoFilter() {
-  // TODO(antonrohr) This filter function can take a lot of time when a large
```
florian-kuebler, line 165: Is that comment not relevant anymore?

florian-kuebler (commented):
I am done with my review now. Thanks again for that change. I left a couple of comments and questions, but in general it LGTM.

OrbitGl/ModuleData.cpp
```diff
@@ -0,0 +1,100 @@
...
+    // __cxxabiv1::__class_type_info::~__class_type_info()
+    // __cxxabiv1::__pbase_type_info::~__pbase_type_info()
+    if (success) {
+      hash_to_function_map_.emplace(FunctionUtils::GetHash(*function), function);
+    } else {
+      address_reuse_counter++;
+    }
+  }
+  if (address_reuse_counter != 0) {
+    LOG("Warning: %d absolute addresses are used by more than one symbol", address_reuse_counter);
```
dimitry-: I think this can be smarter... for example, we can exclude symbols with 0 size (unless we already do that). Maybe there are other way we can get a better idea what is function and what is alias. which binary did you see it in?

OrbitGl/ModuleData.cpp
```diff
@@ -0,0 +1,100 @@
...
+  --it;
+  FunctionInfo* function = it->second.get();
+  CHECK(function->address() <= address);
+
+  if (function->address() + function->size() < address) return nullptr;
+
+  return function;
+}
+
+void ModuleData::LoadSymbols(const orbit_grpc_protos::ModuleSymbols& module_symbols,
```
dimitry-: Does not really load symbols, it set or updates them rather. Maybe AddSymbols?

OrbitGl/ProcessData.cpp
```diff
@@ -0,0 +1,93 @@
...
+  --it;
+  const std::string& module_path = it->second;
+  const MemorySpace* memory_space = module_memory_map_.at(module_path).get();
+  CHECK(absolute_address >= memory_space->start);
+  if (absolute_address > memory_space->end) return not_found_error;
+
+  return std::make_pair(module_path, memory_space->start);
+}
+
+void ProcessData::LoadSymbols(ModuleData* module, const ModuleSymbols& module_symbols) const {
```
dimitry-: same here - AddSymbols()

dimitry- (commented):
Still in the process of review - some comments so far.

OrbitClientData/FunctionUtils.cpp
```diff
@@ -2,13 +2,20 @@
...
 
 #include "OrbitBase/Logging.h"
-#include "Path.h"
-#include "Utils.h"
+#include "absl/container/flat_hash_map.h"
+#include "absl/strings/match.h"
+#include "xxhash.h"
+
+namespace {
+uint64_t StringHash(const std::string& string) {
```
dimitry-, line 15: I wonder if the Hash is still needed. It seems to be used in presets and pdb, but from the looks of it this can be dropped in favor of name.

OrbitClientGgp/ClientGgp.cpp
```diff
@@ -235,11 +234,12 @@ void ClientGgp::ProcessTimer(const TimerInfo& timer_info) { timer_infos_.push_ba
 
 // CaptureListener implementation
 void ClientGgp::OnCaptureStarted(
-    int32_t process_id, std::string process_name, std::shared_ptr<Process> process,
+    std::unique_ptr<ProcessData> process,
+    absl::flat_hash_map<std::string, ModuleData*>&& module_map,
```
dimitry-, line 238: This does not look right. The module map should probably get updated from the service at the start capture - seems like it should be a separate event in the listener. Now we kind of assume that whatever is happened to be on the client at the time of start is the correct module map, which is not necessarily the case.

(not for this PR - but maybe add a todo here and a bug about this.)

OrbitClientData/include/OrbitClientData/ModuleData.h
```diff
@@ -0,0 +1,65 @@
...
+  [[nodiscard]] orbit_client_protos::FunctionInfo* FindFunctionFromHash(uint64_t hash) const {
+    return hash_to_function_map_.contains(hash) ? hash_to_function_map_.at(hash) : nullptr;
+  }
+
+  [[nodiscard]] std::vector<const orbit_client_protos::FunctionInfo*> GetFunctionList() const;
+  [[nodiscard]] absl::flat_hash_map<uint64_t, const orbit_client_protos::FunctionInfo*>
+  GetOrbitFunctions() const;
+
+ private:
+  mutable absl::Mutex functions_mutex;
```
dimitry-, line 55: I think this should protect not only functions but also access to other fields, like is_loaded.

OrbitClientData/include/OrbitClientData/ModuleData.h
```diff
@@ -0,0 +1,65 @@
...
+  orbit_grpc_protos::ModuleInfo module_info_;
```
dimitry-, line 56: const? it effectively is a read-only field.

OrbitClientData/include/OrbitClientData/ModuleData.h
```diff
@@ -0,0 +1,65 @@
...
+  bool is_loaded_;
+  std::map<uint64_t, std::unique_ptr<orbit_client_protos::FunctionInfo>> functions_;
+  // This is a map of hash to function used for preset loading. Currently presets are based on a
+  // hash of the functions pretty name. Maybe consider using something better identifying than
+  // pretty name and maybe consider not hashing it.
```
dimitry-, line 61: Agree - convert it to todo?

OrbitClientData/ModuleData.cpp
```diff
@@ -0,0 +1,100 @@
...
+  CHECK(function->address() <= address);
+
+  if (function->address() + function->size() < address) return nullptr;
+
+  return function;
+}
+
+void ModuleData::LoadSymbols(const orbit_grpc_protos::ModuleSymbols& module_symbols,
+                             uint64_t module_start_address) {
+  CHECK(!is_loaded());
```
dimitry-, line 40: move it after the lock? you modify it there after all. Also maybe check the flag since the function will need to lock too.

OrbitClientData/ModuleData.cpp
```diff
@@ -0,0 +1,100 @@
...
+  }
+  if (address_reuse_counter != 0) {
+    LOG("Warning: %d absolute addresses are used by more than one symbol", address_reuse_counter);
+  }
+
+  is_loaded_ = true;
+}
+
+void ModuleData::ClearSymbols() {
+  CHECK(is_loaded());
```
dimitry-, line 70: same here

OrbitClientData/ModuleData.cpp
```diff
@@ -0,0 +1,100 @@
...
+  std::vector<const FunctionInfo*> result;
+  result.reserve(functions_.size());
+  for (const auto& pair : functions_) {
+    result.push_back(pair.second.get());
+  }
+  return result;
+}
+
+absl::flat_hash_map<uint64_t, const FunctionInfo*> ModuleData::GetOrbitFunctions() const {
+  CHECK(is_loaded_);
```
dimitry-, line 89: and here.

OrbitGl/DataManager.h
```diff
@@ -41,9 +43,11 @@ class DataManager final {
   void set_selected_process(std::shared_ptr<Process> process);
 
   [[nodiscard]] ProcessData* GetProcessByPid(int32_t process_id) const;
-  [[nodiscard]] const std::vector<ModuleData*>& GetModules(int32_t process_id) const;
-  [[nodiscard]] ModuleData* FindModuleByAddressStart(int32_t process_id,
-                                                     uint64_t address_start) const;
+  [[nodiscard]] ModuleData* GetModuleByPath(const std::string& path) const;
```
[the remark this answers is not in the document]
antonrohr: This one can not. The module* that is returned here will be filled with symbols later. Loading symbols alters the module, hence it can't be const here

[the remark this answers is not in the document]
florian-kuebler: Would it be feasible to move the symbol loading modifications to data manager (in a different PR)? At least I would add a comment or rename the method to `GetMutableModuleByPath`, or add it as an additional one.

OrbitClientGgp/ClientGgp.cpp
```diff
@@ -235,11 +234,12 @@ void ClientGgp::ProcessTimer(const TimerInfo& timer_info) { timer_infos_.push_ba
...
     absl::flat_hash_map<uint64_t, orbit_client_protos::FunctionInfo> selected_functions,
     TracepointInfoSet selected_tracepoints) {
-  capture_data_ =
-      CaptureData(process_id, process_name, process, selected_functions, selected_tracepoints);
+  capture_data_ = CaptureData(std::move(process), std::move(module_map), selected_functions,
```
kisenshi, line 241: _process_ and _module_map_ are moved, should _selected_functions_ and _selected_tracepoints_ be moved with `std::move` as well?

OrbitClientGgp/ClientGgp.cpp
```diff
@@ -90,9 +95,11 @@ bool ClientGgp::RequestStartCapture(ThreadPool* thread_pool) {
...
   TracepointInfoSet selected_tracepoints;
 
-  ErrorMessageOr<void> result =
-      capture_client_->StartCapture(thread_pool, pid, target_process_->GetName(), target_process_,
-                                    selected_functions, selected_tracepoints);
+  absl::flat_hash_map<std::string, ModuleData*> memory_map;
+  memory_map[main_module_->file_path()] = main_module_.get();
+
+  ErrorMessageOr<void> result = capture_client_->StartCapture(
+      thread_pool, *(target_process_.get()), memory_map, selected_functions, selected_tracepoints);
```
kisenshi, line 102: Why do you pass `*(target_process_.get())` instead of providing `target_process_` directly?

OrbitGl/ModuleData.h
```diff
@@ -9,35 +9,57 @@
...
-  uint64_t file_size() const { return module_info_.file_size(); }
-  uint64_t address_start() const { return module_info_.address_start(); }
-  uint64_t load_bias() const { return module_info_.load_bias(); }
-
-  std::string address_range() const {
-    return absl::StrFormat("[%016" PRIx64 " - %016" PRIx64 "]", module_info_.address_start(),
-                           module_info_.address_end());
+  // relative_address here is the absolute address minus the address this module was loaded at by
+  // the process (module start address)
+  [[nodiscard]] orbit_client_protos::FunctionInfo* FindFunctionByRelativeAddress(
```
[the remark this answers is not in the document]
antonrohr: Yes, it would be a little cleaner right now, but I specifically designed it here like this for a certain reason. 
I want to decouple Module and Process in a way that Module is independent of the address that its loaded at (its base address / start address). I feel like this makes the structure and interfaces between these classes cleaner. A Process loads Modules in memory regions and Modules contain Functions. The Module and Functions are independent from the Process in the sense that a different Process could load the same Module and nothing would change. Heck, if we loaded the Symbols before, we can reuse them. 

This is at least what I want to work towards to. Currently FunctionInfo still contains the field `module_base_address` which makes this plan not possible for now. I added a couple of TODO comments about this. 

One benefit for example would be in that this structure enables us to support processes dynamically loading and unloading modules (even while we capture).

kisenshi (commented):
I focused mostly on understanding the changes related to the ggp client but I also left some minor comments/questions on some things I noticed while taking an overall look to the code. I am going to fetch the code locally and test the client works as expected so I might include further comments after doing so.

OrbitGl/ProcessData.cpp
antonrohr: I do not want to make ProcessData trivially copyable, because we are using ProcessData* in a lot of places and have to be careful to not accidentally create a copy. That is why I would like the copy constructor to stay deleted. I want to force us to jump through the hoop of calling this function in the rare case that we actually need to make a copy (currently only for a capture thread local instance of ProcessData). 

This is a personal design choice and I am unsure if my reasoning makes sense or if I am creating unnecessary complexity. Any opinions are welcome :)

OrbitGl/ModuleData.h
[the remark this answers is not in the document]
florian-kuebler: Convinced me. Thanks for the clarification.

CMakeLists.txt
```diff
@@ -138,6 +138,7 @@ endif()
 add_subdirectory(ElfUtils)
 add_subdirectory(OrbitBase)
 add_subdirectory(OrbitCaptureClient)
+add_subdirectory(OrbitClientData)
```
kisenshi, line 141: I realised it would make more sense to leave the comment here instead of the main conversation: wondering if it would make sense to use the recently created _OrbitClientModel_ to allocate the data files instead of creating a new _OrbitClientData_ project. I feel that the purpose of both projects is similar

kisenshi:
Ok, I am testing now locally and I have a question and an exception that doesn't allow the client to end properly and therefore, the capure is not finished:

1. I get a warning after loading the symbols and I don't think it did before, is this expected?

```
[           ClientGgp.cpp:169] Found file: /srv/game/assets/benchmark
[           ClientGgp.cpp:170] Loading symbols
[           ScopeTimer.cpp:40] Starting LoadSymbolsFromFile: /srv/game/assets/benchmark...
[           ScopeTimer.cpp:60] LoadSymbolsFromFile: /srv/game/assets/benchmark took 50.600453 ms.
[           ModuleData.cpp:63] Warning: 1040 absolute addresses are used by more than one symbol
```

2. I get an exception after completing the capture and so the client does not end properly and the capture is never saved, etc

```
ClientGgp.cpp:112] Request to stop capture
[       CaptureClient.cpp:161] Finished writing on Capture's gRPC stream: asking to stop capturing
[                 main.cpp:57] Shut down the thread and wait for it to finish
[       CaptureClient.cpp:130] Finished reading from Capture's gRPC stream: all capture data has been received
[           ClientGgp.cpp:247] Capture completed
terminating with uncaught exception of type std::out_of_range: absl::container_internal::raw_hash_map<>::at
Connection to 136.114.97.253 closed.
```
I am trying to understand why this is happening but please could you also take a look? 

Thanks!

OrbitCaptureClient/include/OrbitCaptureClient/CaptureClient.h
```diff
@@ -56,7 +54,8 @@ class CaptureClient {
   }
 
  private:
-  void Capture(int32_t process_id, std::string process_name, std::shared_ptr<Process> process,
+  void Capture(std::unique_ptr<ProcessData> process,
+               absl::flat_hash_map<std::string, ModuleData*>&& module_map,
```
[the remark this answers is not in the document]
antonrohr: Why? I like it this way, because it signals that the caller should provide a object that he resigns ownership of. I am happy to change it, but I don't really see a downside.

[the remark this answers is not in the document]
antonrohr: I bet there is a reason to make it call-by-value. Would you mind explaining?

OrbitCaptureClient/include/OrbitCaptureClient/CaptureListener.h
```diff
@@ -20,7 +19,8 @@ class CaptureListener {
 
   // Called after capture started but before the first event arrived.
   virtual void OnCaptureStarted(
-      int32_t process_id, std::string process_name, std::shared_ptr<Process> process,
+      std::unique_ptr<ProcessData> process,
+      absl::flat_hash_map<std::string, ModuleData*>&& module_map,
```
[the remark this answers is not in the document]
antonrohr: same as last comment

kisenshi:
Ok, Update/clarification on my previous comment: I realised the version of the client in your code is not updated so the capture would not save at all and I was expecing it to do so. However, the exception is still there anyway and the program does not end safely ("All done" is never printed) so there is an issue anyway.

OrbitGl/CallStackDataView.cpp
antonrohr: Nope, since this pointer will potentially be used for symbol loading

OrbitGl/FunctionsDataView.cpp
antonrohr: When I wrote this comment a while ago, I used a debug build instead of release and then it applied. I did not realize at this time that it is only horribly slow when on debug.

kisenshi:
Me again (sorry for so many comments). I was thinking that it might make sense for you to go through all the comments/revision related to the refactoring of the code first and we could the take a look to the client together afterwards if there is still an issue with it, or if it doesn't work as expected once the rest of the code is "stable".

It would also be great if I could test the client twice before merging the code: first using your version of the code to check the error mentioned in the previous comments disappears, etc; and a second/final time just before merging into master, once it has been rebased and the version of the client is the most updated one. This way I could also check the capture is saved/loaded properly and that the ggp client would be still stable/in good shape after this change is included.

Thanks!

OrbitGl/ModuleData.h
```diff
@@ -9,35 +9,57 @@
...
-  const std::string& file_path() const { return module_info_.file_path(); }
-  uint64_t file_size() const { return module_info_.file_size(); }
-  uint64_t address_start() const { return module_info_.address_start(); }
-  uint64_t load_bias() const { return module_info_.load_bias(); }
-
-  std::string address_range() const {
-    return absl::StrFormat("[%016" PRIx64 " - %016" PRIx64 "]", module_info_.address_start(),
-                           module_info_.address_end());
+  // relative_address here is the absolute address minus the address this module was loaded at by
+  // the process (module start address)
```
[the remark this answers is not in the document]
antonrohr: Yes, the load bias is something else than the module_base_address. In the way I designed it right now, the load bias is part of module and will therefore be used in this function (see implementation) 
So what I called here `relative_address` needs to be `absolute_address - module_base_address` or the function does not work as intended. Maybe there is a better wording than `relative_address`, maybe there is also a better way of wording this comment?

OrbitGl/ModuleData.cpp
antonrohr: I actually printed the demangled names of these symbols before, but it is way too verbose. I don't know if we ignore 0 size elsewhere, out of the one that are found here, none of them are 0 size. 

In the Infiltrator demo this occurs 5419 times (total 357245 function symbols) (133 have size 1)
In other libraries it also happens: 
* 3155 duplicates out of 69291 symbols
* 225 out of 50661 
* 1658 out of 31393

I don't know why this happens. I just wanted to print something

OrbitClientData/FunctionUtils.cpp
antonrohr: I am somewhat sure is only used in presets. I would also be in favor to use the function name instead (actually probably demangled name)
I tried to keep this PR as small as possible (haha) (but actually). Since changing the presets requires multiple changes in a few places, I will refer this to a different PR for now.

OrbitGl/ProcessData.cpp
florian-kuebler: Fair enough for me. Thanks for explaining.

OrbitCaptureClient/include/OrbitCaptureClient/CaptureClient.h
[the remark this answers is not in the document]
florian-kuebler: First argument would be why to explicitly disallow copies on the call-side? E.g. for tests it might be useful to use that module map as expectation.
Otherwise, my C++ is a bit too limited. I had that discussion once with @beckerhe and he basically introduced that pattern to me (which is on a data sink, have value params and move on the caller side).

OrbitClientGgp/ClientGgp.cpp
antonrohr: `target_process_` is now of the type `unique_ptr<ProcessData>` and the `StartCapture` takes `const ProcessData&`. So `*(target_process_.get())` does this conversion. I first changed the signature of the `StartCapture` method and then your `target_process_ from a shared_ptr to unique_ptr, so this conversion was the last step necessary. 

While I am looking at this, I just realized there is no real need to use `unique_ptr<ProcessData> target_process_` instead of a simple `ProcessData target_process_`. I think I am going to change it.

OrbitClientData/FunctionUtils.cpp
```diff
@@ -2,22 +2,30 @@
...
 namespace FunctionUtils {
 
 using orbit_client_protos::FunctionInfo;
 using orbit_client_protos::FunctionStats;
 using orbit_client_protos::TimerInfo;
+using orbit_grpc_protos::SymbolInfo;
 
 std::string GetLoadedModuleName(const FunctionInfo& func) {
-  return Path::GetFileName(func.loaded_module_path());
+  return std::filesystem::path(func.loaded_module_path()).filename().string();
```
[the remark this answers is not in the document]
antonrohr: This is a good question. It has actually nothing to do with the actual functionality of the respective methods, its about what lives where and what can be used where. FunctionUtils lives in OrbitClientData and Path lives in OrbitCore. OrbitClientData is a dependency of OrbitCore, so I can use parts of OrbitClientData in OrbitCore. If I wanted to use Path in FunctionUtils, I would need to make OrbitCore a dependency of OrbitClientData and then we have cyclic dependencies (which might work, but is definitely a bad design choice)

What could be done? Path could be moved into another component which is then a dependency of OrbitClientData and OrbitCore. But this seemed a lot of work for only the function GetFileName, so I decided to just use std::filesystem here instead.
 
(The answer to the question about OrbitClientModel vs OrbitClientData is actually kind of the same: avoid cyclic dependencies)

OrbitCaptureClient/include/OrbitCaptureClient/CaptureClient.h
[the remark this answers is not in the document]
florian-kuebler: As discussed offline. Both should be fine.

OrbitGl/ModuleData.cpp
```diff
@@ -0,0 +1,100 @@
...
+#include "ModuleData.h"
+
+#include "FunctionUtils.h"
+#include "OrbitBase/Logging.h"
+#include "absl/synchronization/mutex.h"
+#include "capture_data.pb.h"
+
+using orbit_client_protos::FunctionInfo;
+
+FunctionInfo* ModuleData::FindFunctionByRelativeAddress(uint64_t relative_address,
```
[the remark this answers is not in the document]
antonrohr: yes. And good catch, thank you -> done

OrbitGl/DataManager.h
```diff
@@ -41,9 +43,11 @@ class DataManager final {
...
+  [[nodiscard]] ModuleData* FindModuleByAddress(int32_t process_id, uint64_t absolute_address);
+  [[nodiscard]] orbit_client_protos::FunctionInfo* FindFunctionByAddress(int32_t process_id,
```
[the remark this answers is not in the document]
antonrohr: yes, done

OrbitGl/ModuleData.cpp
```diff
@@ -0,0 +1,100 @@
...
+
+  is_loaded_ = true;
+}
+
+void ModuleData::ClearSymbols() {
+  CHECK(is_loaded());
+  absl::MutexLock lock(&functions_mutex);
+
+  functions_.clear();
+  is_loaded_ = false;
```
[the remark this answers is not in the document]
antonrohr: Thank you for finding the mistakes I do! -> done

OrbitGl/ModuleData.cpp
```diff
@@ -0,0 +1,100 @@
...
+  absl::MutexLock lock(&functions_mutex);
+  std::vector<const FunctionInfo*> result;
+  result.reserve(functions_.size());
+  for (const auto& pair : functions_) {
+    result.push_back(pair.second.get());
+  }
+  return result;
+}
+
+absl::flat_hash_map<uint64_t, const FunctionInfo*> ModuleData::GetOrbitFunctions() const {
```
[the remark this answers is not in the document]
antonrohr: I hear you and did something slightly different. I changed the return type to vector. I think the name fits then better. Let me know if you disagree

OrbitGl/ModuleData.cpp
```diff
@@ -0,0 +1,100 @@
...
+void ModuleData::ClearSymbols() {
+  CHECK(is_loaded());
+  absl::MutexLock lock(&functions_mutex);
+
+  functions_.clear();
+  is_loaded_ = false;
+}
+
+// Sorted by address
+std::vector<const FunctionInfo*> ModuleData::GetFunctionList() const {
```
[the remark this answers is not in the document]
antonrohr: I dont know why I did that. I changed it.

OrbitGl/ModuleDataTest.cpp
```diff
@@ -0,0 +1,184 @@
...
+    // address 2
+    FunctionInfo* result = module.FindFunctionByRelativeAddress(address2, true);
+    ASSERT_NE(result, nullptr);
+    EXPECT_EQ(result->name(), name2);
+  }
+  {
+    // wrong address
+    FunctionInfo* result = module.FindFunctionByRelativeAddress(address1 + 1, true);
+    EXPECT_EQ(result, nullptr);
+  }
```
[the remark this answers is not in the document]
antonrohr: done

OrbitGl/ProcessData.h
antonrohr: done (FormattedAddressRange)

OrbitGl/ProcessData.h
antonrohr: no. I changed it to not use unique_ptr

OrbitGl/ModuleData.cpp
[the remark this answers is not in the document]
florian-kuebler: Sounds good.

OrbitGl/ModuleData.cpp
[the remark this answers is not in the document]
florian-kuebler: Thanks

OrbitCaptureClient/include/OrbitCaptureClient/CaptureClient.h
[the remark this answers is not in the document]
antonrohr: As discussed, I decided to leave it like this

OrbitGl/App.cpp
```diff
@@ -172,6 +178,9 @@ void OrbitApp::OnTimer(const TimerInfo& timer_info) {
   if (timer_info.function_address() > 0) {
     const FunctionInfo& func =
         GetCaptureData().selected_functions().at(timer_info.function_address());
+    // const FunctionInfo* func =
```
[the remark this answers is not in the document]
antonrohr: yes, leftover code. -> Removed it. 
I think this is fine during capture loading, since it only uses `selected_functions()` here, but I am going to test it again when I rebase

OrbitGl/App.cpp
antonrohr: done

OrbitGl/ModulesDataView.h
antonrohr: This comment does not make any sense. I removed it

OrbitGl/ModulesDataView.h
antonrohr: It is actually still relevant. I made it more verbose and hope it makes sense now. If not, please let me know

OrbitGl/ModuleData.h
```diff
@@ -9,35 +9,57 @@
...
+  [[nodiscard]] orbit_client_protos::FunctionInfo* FindFunctionByRelativeAddress(
+      uint64_t relative_address, bool is_exact) const;
+  // TODO(antonrohr): The module_start_address parameter should not be needed here, but it is
+  // because FunctionInfo still contains the field module_base_address. As soon as that field is
+  // gone, remove the parameter here
+  void LoadSymbols(const orbit_grpc_protos::ModuleSymbols& module_symbols,
+                   uint64_t module_start_address);
```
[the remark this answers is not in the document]
antonrohr: good idea, done

OrbitGl/ModuleData.cpp
antonrohr: yes, thats better. done

OrbitGl/ProcessData.cpp
antonrohr: done

OrbitClientGgp/ClientGgp.cpp
antonrohr: I see and agree. I filed a bug and made a TODO comment in CaptureClient

OrbitClientData/include/OrbitClientData/ModuleData.h
antonrohr: I agree. I renamed the mutex to a general `mutex_` and also locked it in the functions I forgot.

OrbitClientData/include/OrbitClientData/ModuleData.h
antonrohr: true. done

OrbitClientData/include/OrbitClientData/ModuleData.h
antonrohr: yes and I created a bug

OrbitClientData/ModuleData.cpp
antonrohr: that makes sense, done

OrbitClientData/ModuleData.cpp

OrbitClientData/ModuleData.cpp
antonrohr: done

OrbitClientGgp/ClientGgp.cpp
antonrohr: yes you are right. I changed it.

OrbitGl/ModulesDataView.h
florian-kuebler: Okay. Thanks!

CMakeLists.txt
antonrohr: See my other comment, it is basically to avoid cyclic dependencies between different components

OrbitGl/DataManager.h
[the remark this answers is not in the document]
antonrohr: yes, that is a good idea. I added an additional `GetMutableModuleByPath`

OrbitClientData/ModuleData.cpp
```diff
@@ -67,16 +72,22 @@ void ModuleData::LoadSymbols(const orbit_grpc_protos::ModuleSymbols& module_symb
...
   functions_.clear();
+  hash_to_function_map_.clear();
   is_loaded_ = false;
 }
 
-// Sorted by address
-std::vector<const FunctionInfo*> ModuleData::GetFunctionList() const {
-  absl::MutexLock lock(&functions_mutex);
+[[nodiscard]] orbit_client_protos::FunctionInfo* ModuleData::FindFunctionFromHash(
```
florian-kuebler, line 83: Not sure if I asked previous, but can the be const? If not, please just resolve the comment.

OrbitClientData/include/OrbitClientData/ModuleData.h
```diff
@@ -29,36 +29,32 @@ class ModuleData final {
...
   void ClearSymbols();
-  [[nodiscard]] orbit_client_protos::FunctionInfo* FindFunctionFromHash(uint64_t hash) const {
-    return hash_to_function_map_.contains(hash) ? hash_to_function_map_.at(hash) : nullptr;
-  }
-
-  [[nodiscard]] std::vector<const orbit_client_protos::FunctionInfo*> GetFunctionList() const;
-  [[nodiscard]] absl::flat_hash_map<uint64_t, const orbit_client_protos::FunctionInfo*>
-  GetOrbitFunctions() const;
+  [[nodiscard]] orbit_client_protos::FunctionInfo* FindFunctionFromHash(uint64_t hash) const;
+  [[nodiscard]] std::vector<const orbit_client_protos::FunctionInfo*> GetFunctions() const;
```
florian-kuebler, line 47: if you are not modifying the vectors, maybe just also make them const?

florian-kuebler (approved):
From my side, looks good. Thanks!

OrbitGl/App.cpp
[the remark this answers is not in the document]
florian-kuebler: Agree. As long as you only use selected_functions, this is fine!

antonrohr:
@kisenshi I hope I fixed everything in ClientGgp now so it works as expected. It would be very nice if you could run it to see that its not just my machine. (And then after the rebase on Monday, we will test it again)

OrbitClientData/ModuleData.cpp
antonrohr: I dont know why I this still wasn't const. I made it const and removed the old comment

OrbitClientData/include/OrbitClientData/ModuleData.h
antonrohr: done

antonrohr:
I did just the big rebase to be up to date with master again. I am going to do some extensive testing tomorrow. 

@florian-kuebler If you would look at the CaptureTests, that would be great.

florian-kuebler (approved):
I looked again into the changes of the serialization tests and they lgtm. Thanks!

OrbitClientData/ModuleData.cpp
```diff
@@ -0,0 +1,109 @@
...
+  absl::MutexLock lock(&mutex_);
+  return is_loaded_;
+}
+
+const FunctionInfo* ModuleData::FindFunctionByRelativeAddress(uint64_t relative_address,
+                                                              bool is_exact) const {
+  absl::MutexLock lock(&mutex_);
+  if (functions_.empty()) return nullptr;
+
+  uint64_t address = relative_address + load_bias();
```
dimitry-, line 24: ok, I think this introduces a new kind of address we did not have before. So now we have absolute address, elf-address, and relative address. Does it give us any advantages? say if I have an address from the elf file symbol table I now have to calculate load-bias and substract it from the address before passing it to this function?

antonrohr: As discussed offline, I added another function called FindFunctionByElfAddress and used it in FindFunctionByRelativeAddress. This should clarify the code.

dimitry- (approved):
lgtm, thanks!

PR #1210, opened 2020-09-15
Display graph track value at current x mouse position

pierricgimmig:
With the manual instrumentation API, a user can track the values of interesting variables which Orbit will plot in graph tracks. This PR adds basic support for displaying the graph value at the current X mouse position (See picture below). This should eventually be  polished a little bit. I might need your help Pablo if you have some time.

image

### DIFF
--- a/OrbitGl/GraphTrack.cpp
+++ b/OrbitGl/GraphTrack.cpp
@@ -35,6 +35,13 @@ void GraphTrack::Draw(GlCanvas* canvas, PickingMode picking_mode) {
   float text_z = GlCanvas::kZValueText;
 
   Box box(pos_, Vec2(size_[0], -size_[1]), track_z);
+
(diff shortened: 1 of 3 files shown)


karupayun (approved):
Really nice Pierric! :)

Looking forward to our meeting. As I pointed in our chat, I want to mention 5 things:
- Mouse (we only want to label this new feature when using the mouse?)
- How to label (Textbox at the top Z or different kind of label).
- Color: Do we have to choose different text color?
- Smart text draw (Something that Ronald mentioned. The problem of drawing text near another one)
- Other profilers: I want to see how different profilers work with this kind of thing, so we can get ideas from there

pierricgimmig:
> Looking forward to our meeting

Likewise! Thanks Pablo!

PR #1212, opened 2020-09-16
Integrate Chromium

beckerhe:
This PR integrates Chromium into Orbit by using Qt's WebEngine module.

It consists of 5 commits:
1. Prepare the windows docker containers for additional build requirements
2. Prepare the linux docker containers for additional build requirements
3. Adds Qt as a downstream fork of bincrafter's conan recipe. This was needed because the latest version of MSVC2017 lead to a couple of compilation errors which needed to be patched. This also updated Qt to the latest version.
4. Integrates the new webengine module into our conan recipe and require it in cmake. No further integration is done in this PR. This will come in multiple follow-up PRs.
5. Add an extra NOTICE file which only contains the licenses from Chromium dependencies. This is needed to simplify the licensing review. All these licenses are also added to our standard NOTICE file.


florian-kuebler (approved):
Thanks for this change. LGTM, but I can't really judge about the conan stuff you are doing in commit 3.

reichlfl (approved):
Thanks for splitting this into separate commits! Can't see anything that looks suspicious to me - as the original Florian pointed out, it's not easy to judge, but LGTM as far as I can tell.

beckerhe:
Hello to both of you,
since you both said you don't know how to review the conan-part some words from me on how to check my work:

1. The conan recipe (commit 3) was copied from bincrafters and changed. So it should contain a license file in the same directory and the license should be permissive. It's MIT in this case which is the only license I've seen so far for conan recipes. Apart from that I would say there is nothing more to review here.
2. The conan-integration (commit 4) mostly proves itself by compilation and running tests. But there are some things which can be verified. Every `self.requires`-call in the conanfile should specify a recipe revision (that's the hash after the pound sign). If it's not there then conan will choose the latest available revision and that might break things. Furthermore the lockfile changes can be roughly checked for validity. These are created by a script, so there is not much that can go wrong, but sometimes package changes slip in here which are unrelated to the actual change. In this case we change the Qt-package. So the lockfile-changes should resemble that. The diff shows the change of the qt-package and a change for the jom-package. jom is Qt's build system, so a new version was needed for the new Qt version. This is fine. 

Thanks for the review by the way!

Henning

PR #1213, opened 2020-09-16
Report a error if grpc server is not started

dimitry-:
Bug: http://b/168295378

### DIFF
--- a/OrbitService/OrbitService.cpp
+++ b/OrbitService/OrbitService.cpp
@@ -44,9 +44,14 @@ namespace orbit_service {
 void OrbitService::Run(std::atomic<bool>* exit_requested) {
   LOG("Running Orbit Service version %s", OrbitCore::GetVersion());
   std::string grpc_address = absl::StrFormat("127.0.0.1:%d", grpc_port_);
-  LOG("Starting GRPC server at %s", grpc_address);
+  LOG("Starting gRPC server at %s", grpc_address);
   std::unique_ptr<OrbitGrpcServer> grpc_server;
   grpc_server = OrbitGrpcServer::Create(grpc_address);
+  if (grpc_server == nullptr) {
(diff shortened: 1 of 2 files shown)


OrbitService/OrbitService.cpp
```diff
@@ -47,6 +47,11 @@ void OrbitService::Run(std::atomic<bool>* exit_requested) {
   LOG("Starting GRPC server at %s", grpc_address);
   std::unique_ptr<OrbitGrpcServer> grpc_server;
   grpc_server = OrbitGrpcServer::Create(grpc_address);
+  if (grpc_server == nullptr) {
+    ERROR("Unable to start GRPC server.");
```
dpallotti, line 51:
```suggestion
    ERROR("Unable to start gRPC server");
```

OrbitService/OrbitService.cpp
```diff
@@ -47,6 +47,11 @@ void OrbitService::Run(std::atomic<bool>* exit_requested) {
...
+    return;
+  }
+  LOG("GRPC server is running");
```
dpallotti, line 54:
```suggestion
  LOG("gRPC server is running");
```

dpallotti (approved):

IrinaShkviro (approved):
Thanks a lot Dimitry! Just one question: is it possible to enable grpc log also?

dimitry-:
> Thanks a lot Dimitry! Just one question: is it possible to enable grpc log also?

I think it is but considering that we do not really expect grpc to fail, I decided not to at least for now.

OrbitService/OrbitService.cpp
dimitry-: done.

OrbitService/OrbitService.cpp
dimitry-: done.

PR #1214, opened 2020-09-16
Client ggp release preparation

kisenshi:
The first version of the ggp client has been finished and works properly. This PR include some small additions to the client that are useful when released:

* See flag configuration with -helpfull command

* Add the creation of a log file in `/var/log` (this is only possible when the client is executed with sudo privileges)

* Add _strip_symbols_ compilation rule so symbols are not included in the executable


danielfenner (approved):
lgtm!

kisenshi:
Merging this now, thanks!

PR #1215, opened 2020-09-16
Fix presets for the e2e test.

danielfenner:
Re-created presets with fitting module path.

Bug: 161839506
Test: N/A


IrinaShkviro (approved):
lgtm, thanks!

PR #1216, opened 2020-09-16
Add Test for CaptureDeserializer

florian-kuebler:


OrbitClientModel/CaptureDeserializerTest.cpp
```diff
@@ -0,0 +1,408 @@
...
+  EXPECT_CALL(listener, OnCaptureComplete).Times(0);
+  EXPECT_CALL(listener, OnCaptureCancelled).Times(0);
+  capture_deserializer::Load("not_existing_test_file", &listener, &cancellation_requested);
+}
+
+TEST(CaptureDeserializer, LoadNoVersion) {
+  MockCaptureListener listener;
+  std::atomic<bool> cancellation_requested = false;
+  ErrorMessage actual_error;
+  EXPECT_CALL(listener, OnCaptureFailed).Times(1);
```
dimitry-, line 71: do you want to check parameter here as well? So that you get the correct error message?

OrbitClientModel/CaptureDeserializerTest.cpp
```diff
@@ -0,0 +1,408 @@
...
+  MOCK_METHOD(void, OnAddressInfo, (LinuxAddressInfo), (override));
+  MOCK_METHOD(void, OnUniqueTracepointInfo, (uint64_t /*key*/, TracepointInfo /*tracepoint_info*/),
+              (override));
+  MOCK_METHOD(void, OnTracepointEvent, (TracepointEventInfo), (override));
+};
+
+TEST(CaptureDeserializer, LoadFileNotExists) {
+  MockCaptureListener listener;
+  std::atomic<bool> cancellation_requested = false;
+  EXPECT_CALL(listener, OnCaptureFailed).Times(1);
```
dimitry-, line 60: also maybe check the parameter?

dimitry- (approved):
otherwise lgtm

OrbitClientModel/CaptureDeserializerTest.cpp
```diff
@@ -0,0 +1,408 @@
...
+
+  capture_deserializer::internal::LoadCaptureInfo(empty_capture_info, &listener, &coded_input,
+                                                  &cancellation_requested);
+
+  EXPECT_EQ(timer_1.start(), actual_timer_1.start());
+  EXPECT_EQ(timer_2.start(), actual_timer_2.start());
+  EXPECT_EQ(timer_1.end(), actual_timer_1.end());
+  EXPECT_EQ(timer_2.end(), actual_timer_2.end());
+  EXPECT_EQ(timer_1.process_id(), actual_timer_1.process_id());
+  EXPECT_EQ(timer_2.process_id(), actual_timer_2.process_id());
```
reichlfl, line 405: And another change for `operator=`

reichlfl (approved):
Some comments which all point out the same thing - just a suggestion. I can't really judge if there are more parts the tests should cover since I've not been involved in the capture deserializer at all so far.

Apart from that, existing tests LGTM, thanks for the effort of adding them!

OrbitClientModel/CaptureDeserializerTest.cpp
florian-kuebler: I was a bit unsure of this. I think it is a bit contradictory to what the test should ensure. The intended behaviour is to have the listener be notified that something went wrong. The exact error message may be well subject to change and should be not that much of the focus of the test.
This is a bit different for the test with the wrong version, because there I expect having an error message that tells me something about the version.
It will also introduce unnecessary work when changing the error message.

But I am not having a very strong opinion on this. So I would be also fine, with adding the check there.

OrbitClientModel/CaptureDeserializerTest.cpp
```diff
@@ -0,0 +1,408 @@
...
+  capture_deserializer::internal::LoadCaptureInfo(capture_info, &listener, &empty_stream,
+                                                  &cancellation_requested);
+
+  ASSERT_EQ(actual_selected_functions.size(), 1);
+  ASSERT_TRUE(actual_selected_functions.contains(selected_function->address()));
+  FunctionInfo actual_function_info = actual_selected_functions.at(selected_function->address());
+
+  EXPECT_EQ(actual_function_info.name(), selected_function->name());
+  EXPECT_EQ(actual_function_info.pretty_name(), selected_function->pretty_name());
+  EXPECT_EQ(actual_function_info.size(), selected_function->size());
```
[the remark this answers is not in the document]
florian-kuebler: Would love the have those. However, they are auto generated by protos. GoogleTest will add a EqualsProto matcher at some point. But right now, I can't modify. I was about implementing my own custom matcher but thought it would be an overkill for this test.

OrbitClientModel/CaptureDeserializerTest.cpp
```diff
@@ -0,0 +1,408 @@
...
+                                                  &cancellation_requested);
+
+  EXPECT_EQ(actual_address_info_1.function_name(), address_info_1.function_name());
+  EXPECT_EQ(actual_address_info_2.function_name(), address_info_2.function_name());
+  EXPECT_EQ(actual_address_info_1.module_path(), address_info_1.module_path());
+  EXPECT_EQ(actual_address_info_2.module_path(), address_info_2.module_path());
+  EXPECT_EQ(actual_address_info_1.offset_in_function(), address_info_1.offset_in_function());
+  EXPECT_EQ(actual_address_info_2.offset_in_function(), address_info_2.offset_in_function());
+  EXPECT_EQ(actual_address_info_1.absolute_address(), address_info_1.absolute_address());
+  EXPECT_EQ(actual_address_info_2.absolute_address(), address_info_2.absolute_address());
```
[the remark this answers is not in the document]
florian-kuebler: As above, they are generated by a proto.

OrbitClientModel/CaptureDeserializerTest.cpp
florian-kuebler: As above, they are generated by a proto.

OrbitClientModel/CaptureDeserializerTest.cpp
florian-kuebler: See above.

OrbitClientModel/CaptureDeserializerTest.cpp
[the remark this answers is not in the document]
reichlfl: Ah, I see, didn't know that - makes sense, thanks for pointing that out.

OrbitClientModel/CaptureDeserializerTest.cpp
```diff
@@ -0,0 +1,408 @@
...
+  address_info_1.set_module_path("/path");
+  address_info_1.set_offset_in_function(0);
+  address_info_1.set_absolute_address(123);
+  capture_info.add_address_infos()->CopyFrom(address_info_1);
+
+  LinuxAddressInfo address_info_2;
+  address_info_2.set_function_name("bar");
+  address_info_2.set_module_path("/path");
+  address_info_2.set_offset_in_function(0);
+  address_info_2.set_absolute_address(123);
```
IrinaShkviro, line 187: Shouldn't be data in `address_info_2` different from `address_info_1`?

OrbitClientModel/CaptureDeserializerTest.cpp
```diff
@@ -0,0 +1,408 @@
...
+  callstack_data_2.push_back(4);
+  callstack_data_2.push_back(5);
+  CallStack callstack_2(std::move(callstack_data_2));
+  CallstackInfo* callstack_info_2 = capture_info.add_callstacks();
+  *callstack_info_2->mutable_data() = {callstack_2.GetFrames().begin(),
+                                       callstack_2.GetFrames().end()};
+  CallstackEvent* callstack_event_2 = capture_info.add_callstack_events();
+  callstack_event_2->set_thread_id(2);
+  callstack_event_2->set_time(3);
+  callstack_event_2->set_callstack_hash(callstack_1.GetHash());
```
IrinaShkviro, line 297: Did you mean `callstack_2`?
```suggestion
  callstack_event_2->set_callstack_hash(callstack_2.GetHash());
```

IrinaShkviro (approved):
Thanks for the change! Please see the minor comments.

OrbitClientModel/CaptureDeserializerTest.cpp
dimitry-: I think certain error messages are expected expected in certain situation and it is ok to check them. If they change - they change, as for other things - we fix test as well to match expected behavior.

In this test you are emulating a specific thing that goes wrong. so it is fine to expect a specific error message in this case.

OrbitClientModel/CaptureDeserializerTest.cpp
florian-kuebler: Sure, thats better

OrbitClientModel/CaptureDeserializerTest.cpp
florian-kuebler: Thanks!

OrbitClientModel/CaptureDeserializerTest.cpp
florian-kuebler: Okay, I added the checks.

OrbitClientModel/CaptureDeserializerTest.cpp
florian-kuebler: Done.

PR #1217, opened 2020-09-17
Add Qt WebEngine to GCC docker containers

beckerhe:
The GCC docker containers must also be able to compile Orbit with
webengine. This was an oversight from yesterday and should have been
part of commit 8391285.


reichlfl (approved):
LGTM, thanks!

PR #1218, opened 2020-09-17
Set Callbacks in Fuzzer and Add CHECKS in App

florian-kuebler:
Recent changes from "if (..._callback_)" to "CHECK(..._callback_)"
led to failed checks in the Fuzzers, as the callbacks were not
explicitly set.

This change sets the required callbacks in the Fuzzer.
Also it replaces the remaining "if"s with "CHECK" to ensure
that the callbacks are properly initialized.

Bug: http://b/168770539
Test: Compile

### DIFF
--- a/OrbitGl/App.cpp
+++ b/OrbitGl/App.cpp
@@ -131,11 +131,9 @@ void OrbitApp::OnCaptureStarted(int32_t process_id, std::string process_name,
         capture_data_ = CaptureData(process_id, std::move(process_name), std::move(process),
                                     std::move(selected_functions), std::move(selected_tracepoints));
 
-        CHECK(capture_started_callback_);
         capture_started_callback_();
 
         if (has_selected_functions) {
-          CHECK(select_live_tab_callback_);
           select_live_tab_callback_();
         }
 
@@ -160,10 +158,8 @@ void OrbitApp::OnCaptureComplete() {
                           GetCaptureData().GetCallstackData()->GetUniqueCallstacksCopy());
         SetTopDownView(GetCaptureData());
 
-        CHECK(capture_stopped_callback_);
         capture_stopped_callback_();
 
-        CHECK(open_capture_finished_callback_);
         open_capture_finished_callback_();
 
         FireRefreshCallbacks();
@@ -172,10 +168,8 @@ void OrbitApp::OnCaptureComplete() {
 
 void OrbitApp::OnCaptureCancelled() {
   main_thread_executor_->Schedule([this]() mutable {
-    CHECK(capture_failed_callback_);
     capture_failed_callback_();
 
-    CHECK(open_capture_failed_callback_);
     open_capture_failed_callback_();
 
     ClearCapture();
@@ -184,10 +178,8 @@ void OrbitApp::OnCaptureCancelled() {
 
 void OrbitApp::OnCaptureFailed(ErrorMessage error_message) {
   main_thread_executor_->Schedule([this, error_message = std::move(error_message)]() mutable {
-    CHECK(capture_failed_callback_);
     capture_failed_callback_();
 
-    CHECK(open_capture_failed_callback_);
     open_capture_failed_callback_();
 
     ClearCapture();
@@ -496,7 +488,6 @@ void OrbitApp::SetSamplingReport(
     absl::flat_hash_map<CallstackID, std::shared_ptr<CallStack>> unique_callstacks) {
   auto report =
       std::make_shared<SamplingReport>(std::move(sampling_profiler), std::move(unique_callstacks));
-  CHECK(sampling_reports_callback_);
   DataView* callstack_data_view = GetOrCreateDataView(DataViewType::kCallstack);
   sampling_reports_callback_(callstack_data_view, report);
 
@@ -511,7 +502,6 @@ void OrbitApp::SetSelectionReport(
     SamplingProfiler sampling_profiler,
     absl::flat_hash_map<CallstackID, std::shared_ptr<CallStack>> unique_callstacks,
     bool has_summary) {
-  CHECK(selection_report_callback_);
   auto report = std::make_shared<SamplingReport>(std::move(sampling_profiler),
                                                  std::move(unique_callstacks), has_summary);
   DataView* callstack_data_view = GetOrCreateSelectionCallstackDataView();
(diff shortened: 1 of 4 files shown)


dpallotti:
As Ronald wrote in the bug, this renders the `CHECK`s useless. Either we accept it and remove the `CHECK`s, or the initialization to dummy functions should happen in the fuzzer(s).

florian-kuebler:
As written in the bug. These checks are basically explicitly stating the assumption that the function is save to call. Yes, currently this will be the case, but they should prevent us from changes in the future.

However, if you both strongly argue to remove the checks, that would be also fine for me to remove them. I would however, strongly avoid setting the callbacks in the fuzzer.

dpallotti:
> As written in the bug. These checks are basically explicitly stating the assumption that the function is save to call. Yes, currently this will be the case, but they should prevent us from changes in the future.
> 
> However, if you both strongly argue to remove the checks, that would be also fine for me to remove them. I would however, strongly avoid setting the callbacks in the fuzzer.

On my side, the point is not about removing the `CHECK`s, but rather about the fact that they no longer guarantee that clients of `OrbitApp` set the callbacks and whether we are fine with this or not. So the question is, why

> I would however, strongly avoid setting the callbacks in the fuzzer.

? Is it because they are many and it's messy to set them all for each fuzzer?

florian-kuebler:
> > As written in the bug. These checks are basically explicitly stating the assumption that the function is save to call. Yes, currently this will be the case, but they should prevent us from changes in the future.
> > However, if you both strongly argue to remove the checks, that would be also fine for me to remove them. I would however, strongly avoid setting the callbacks in the fuzzer.
> 
> On my side, the point is not about removing the `CHECK`s, but rather about the fact that they no longer guarantee that clients of `OrbitApp` set the callbacks and whether we are fine with this or not. So the question is, why
> 
> > I would however, strongly avoid setting the callbacks in the fuzzer.
> 
> ? Is it because they are many and it's messy to set them all for each fuzzer?

I see your point. My point was basically, not messing up the fuzzer here, you are right. Still I have the feeling, that a controller (the app), should be in a well behaving state after just constructing it.

OrbitGl/App.cpp
```diff
@@ -131,11 +131,11 @@ void OrbitApp::OnCaptureStarted(int32_t process_id, std::string process_name,
         capture_data_ = CaptureData(process_id, std::move(process_name), std::move(process),
```
dimitry-, line 131: this is comment on the commit message: s/Dummy/Empty/

dimitry- (approved):
Otherwise lgtm

dimitry-:
As for checks... someone could still set these to nullptr if I am not mistaken.. But I agree they look less useful now. Maybe move these CHECKs to Set*Listener functions?

florian-kuebler:
> As for checks... someone could still set these to nullptr if I am not mistaken.. But I agree they look less useful now. Maybe move these CHECKs to Set*Listener functions?

Just as a note, the callbacks are all value type functions. It is just that the default constructor of std::function will create a function that fails when it is called.
In general, I like the idea of moving the checks to the setter.

florian-kuebler:
Thanks for your reviews so far. I moved the CHECKS to the setter methods

ronaldfw:
> As written in the bug. These checks are basically explicitly stating the assumption that the function is save to call. Yes, currently this will be the case, but they should prevent us from changes in the future.

You are right that the CHECKs are still useful, but they lose a lot of what they were guarding us against before, which is: At some point between creation of the object and calling these methods, these callbacks have been properly initialized. That logic is potentially fragile and I'm not in favor of losing the guards against that. (Echoing what @dpallotti dpallotti said, I think.)

Moving the CHECKs to the setter as suggested by @dimitry- helps a bit, but it doesn't guarantee that the Set* methods are actually ever called (which is what we have as a guarantee now). 

If possible, I'd prefer to fix the root cause in the fuzzer. Can you clarify why you don't like that approach?

ronaldfw:
> I see your point. My point was basically, not messing up the fuzzer here, you are right. Still I have the feeling, that a controller (the app), should be in a well behaving state after just constructing it.

But is it even possible that the app is in a well behaving state just after constructing it? If the callbacks are default initialized to do nothing, then you can call them (in the places where we had the CHECKs), but the UI is broken (unless the Set* methods are properly called, which we would have no guarantee for anymore). I'd not consider that well behaving.

florian-kuebler:
> > I see your point. My point was basically, not messing up the fuzzer here, you are right. Still I have the feeling, that a controller (the app), should be in a well behaving state after just constructing it.
> 
> But is it even possible that the app is in a well behaving state just after constructing it? If the callbacks are default initialized to do nothing, then you can call them (in the places where we had the CHECKs), but the UI is broken (unless the Set* methods are properly called, which we would have no guarantee for anymore). I'd not consider that well behaving.

I'd argue that this depends on the definition of well behaving. The whole idea of the callbacks is to decouple our controller (the app) from the UI. So a controller that has not called the Set* methods, can still count on the behaviours of functions, but with having no visual response. Which is especially true for the fuzzer, but might also be relevant for e.g. tests.

As I said, I also see your point. I would loop in @beckerhe, as this would basically affect the fuzzers.

ronaldfw:
> I'd argue that this depends on the definition of well behaving. The whole idea of the callbacks is to decouple our controller (the app) from the UI. So a controller that has not called the Set* methods, can still count on the behaviours of functions, but with having no visual response. Which is especially true for the fuzzer, but might also be relevant for e.g. tests.

The case we care about the most is the behavior when the app is in front of a user. Maybe there is a different way to guarantee that callbacks have been properly initialized or maybe the fuzzer could use a different type of App class, etc. I don't have insight into all parts that are affected here, but I'd like use to carefully consider what we are giving up here.

florian-kuebler:
> > I'd argue that this depends on the definition of well behaving. The whole idea of the callbacks is to decouple our controller (the app) from the UI. So a controller that has not called the Set* methods, can still count on the behaviours of functions, but with having no visual response. Which is especially true for the fuzzer, but might also be relevant for e.g. tests.
> 
> The case we care about the most is the behavior when the app is in front of a user. Maybe there is a different way to guarantee that callbacks have been properly initialized or maybe the fuzzer could use a different type of App class, etc. I don't have insight into all parts that are affected here, but I'd like use to carefully consider what we are giving up here.

Right. As I said, I have no strong opinion on this. So I changed the code accordingly (default initiazlization and checks in the App and Set* calls in the Fuzzer).

ronaldfw (approved):
Please update the PR description, o/w LGTM!

florian-kuebler:
Done. Will merge now. Thanks for the review!.

PR #1219, opened 2020-09-17
Fix FunctionsDataView::ParallelFilter to actually use the module name

dpallotti:
Like in `FunctionsDataView::DoFilter`. The mistake arose at some point because
"file" was previously used to refer to the module (at least I assume).


florian-kuebler:
Looping in @antonrohr as he recently worked on module names/paths.

dimitry- (approved):

PR #1221, opened 2020-09-17
Add small vertical gap between GPU events

karupayun:
Although GPU events are not related, they look like a flamechart and an
user could think that an event below another could have some kind of
relationship. So, we add a small gap similar to the one in Scheduler tracks.
Related to b/168684948.

Currently:
Screen Shot 2020-09-17 at 12 18 18

Now (distance 2 between gpu depths):
Screen Shot 2020-09-17 at 12 19 10

Another possibility (distance 1):
Screen Shot 2020-09-17 at 12 17 48


ronaldfw:
Distance of two looks good to me!

ronaldfw (approved):
LGTM, thanks a lot!

beckerhe:
> Distance of two looks good to me!

To me too!

beckerhe (approved):

PR #1222, opened 2020-09-17
Update two more conan recipe revisions

beckerhe:
Just like cctz, there have been two more packages (pcre2 and double-conversion) whose recipe revision need to be updated such that they work on the public remotes.

Both are Qt-dependencies, that's why only the windows lockfiles need to be updated. On Linux we use system-installed Qt.


florian-kuebler:
LGTM. Thanks!

florian-kuebler (approved):

beckerhe:
> LGTM. Thanks!

Thanks Florian!

PR #1225, opened 2020-09-17
Manual instrumentation "async" operation tracking

pierricgimmig:
An async operation can start in one scope and end in another,
possibly on a different thread altogether. Start and stop pairs are
matched using a user provided "id". See ASYNC_START and ASYNC_STOP
macros in Orbit.h. A new ASYNC_STRING macro allows to associate an
arbitrarily long string to an async event by specifying the event "id".

On the UI side, a new AsyncTrack takes care of displaying the async
spans in the capture view in a compact way.

See OrbitTest.cpp for example usage.

Notes:
- the async track rendering will be modified to insert a vertical gap between rows of timeslices in a subsequent PR to make it clear that the async tracks are not flame graphs.
- The color of the timeslices is chosen using the hash of their labels.

image

### DIFF
--- a/Orbit.h
+++ b/Orbit.h
@@ -36,6 +36,8 @@
 #define ORBIT_START_ASYNC(name, id) ORBIT_START_ASYNC_WITH_COLOR(name, id, orbit::Color::kAuto)
 #define ORBIT_START_ASYNC_WITH_COLOR(name, id, col) orbit_api::StartAsync(name, id, col)
 #define ORBIT_STOP_ASYNC(id) orbit_api::StopAsync(id)
+#define ORBIT_ASYNC_STRING(str, id) orbit_api::AsyncString(str, id, orbit::Color::kAuto)
+#define ORBIT_ASYNC_STRING_WITH_COLOR(str, id, col) orbit_api::AsyncString(str, id, col)
 
 // ORBIT_[type]: graph variables.
 #define ORBIT_INT(name, val) ORBIT_INT_WITH_COLOR(name, val, orbit::Color::kAuto)
@@ -57,7 +59,7 @@ namespace orbit {
 
 // Material Design Colors #500
 enum class Color : uint32_t {
-  kAuto = 0x00000001,
+  kAuto = 0x00000000,
   kRed = 0xf44336ff,
   kPink = 0xe91e63ff,
   kPurple = 0x9c27b0ff,
@@ -88,6 +90,8 @@ enum class Color : uint32_t {
 #define ORBIT_STOP()
 #define ORBIT_START_ASYNC(name, id)
 #define ORBIT_STOP_ASYNC(id)
+#define ORBIT_ASYNC_STRING(str, id)
+#define ORBIT_ASYNC_STRING_WITH_COLOR(str, id, col)
 #define ORBIT_INT(name, value)
 #define ORBIT_INT64(name, value)
 #define ORBIT_UINT(name, value)
@@ -140,6 +144,7 @@ enum EventType : uint8_t {
   kTrackUint64 = 8,
   kTrackFloat = 9,
   kTrackDouble = 10,
+  kString = 11,
 };
 
 constexpr size_t kMaxEventStringSize = 30;
@@ -224,13 +229,27 @@ inline void Stop() {
 }
 
 inline void StartAsync(const char* name, uint64_t id, orbit::Color color) {
-  EncodedEvent e(EventType::kScopeStartAsync, name, id, color);
+  EncodedEvent e(EventType::kScopeStartAsync, name, /*value*/ 0, color, id);
   StartAsync(e.args[0], e.args[1], e.args[2], e.args[3], e.args[4], e.args[5]);
 }
 
 inline void StopAsync(uint64_t id) {
-  EncodedEvent e(EventType::kScopeStopAsync, nullptr, id);
+  EncodedEvent e(EventType::kScopeStopAsync, /*name*/ nullptr, /*value*/ 0, orbit::Color::kAuto,
+                 id);
   StopAsync(e.args[0], e.args[1], e.args[2], e.args[3], e.args[4], e.args[5]);
+}
+
+inline void AsyncString(const char* str, uint32_t id, orbit::Color color) {
+  if (str == nullptr) return;
+  constexpr size_t chunk_size = kMaxEventStringSize - 1;
+  const char* end = str + strlen(str);
+  while (str < end) {
+    EncodedEvent e(EventType::kString, /*name*/ nullptr, /*value*/ 0, color, id);
+    std::strncpy(e.event.name, str, chunk_size);
+    e.event.name[chunk_size] = 0;
+    TrackValue(e.args[0], e.args[1], e.args[2], e.args[3], e.args[4], e.args[5]);
+    str += chunk_size;
+  }
 }
 
 inline void TrackValue(EventType type, const char* name, uint64_t value, orbit::Color color) {
@@ -244,6 +263,7 @@ void Start(const char* name, orbit::Color color);
 void Stop();
 void StartAsync(const char* name, uint32_t id, orbit::Color color);
 void StopAsync(uint32_t id);
+void AsyncString(const char* str, uint32_t id);
 void TrackValue(EventType type, const char* name, uint64_t value, orbit::Color color);
 
 #endif  // ORBIT_API_INTERNAL_IMPL
--- a/Orbit.h
+++ b/Orbit.h
@@ -218,6 +218,12 @@ ORBIT_STUB void TrackValue(uint64_t, uint64_t, uint64_t, uint64_t, uint64_t, uin
 // NOTE: Do not use these directly, use corresponding macros instead.
 #ifndef ORBIT_API_INTERNAL_IMPL
 
+// Default values.
+constexpr const char* kNameNullPtr = nullptr;
+constexpr uint32_t kIdZero = 0;
+constexpr uint64_t kValueZero = 0;
+constexpr orbit::Color kColorAuto = orbit::Color::kAuto;
+
 inline void Start(const char* name, orbit::Color color) {
   EncodedEvent e(EventType::kScopeStart, name, 0, color);
   Start(e.args[0], e.args[1], e.args[2], e.args[3], e.args[4], e.args[5]);
@@ -228,14 +234,13 @@ inline void Stop() {
   Stop(e.args[0], e.args[1], e.args[2], e.args[3], e.args[4], e.args[5]);
 }
 
-inline void StartAsync(const char* name, uint64_t id, orbit::Color color) {
-  EncodedEvent e(EventType::kScopeStartAsync, name, /*value*/ 0, color, id);
+inline void StartAsync(const char* name, uint32_t id, orbit::Color color) {
+  EncodedEvent e(EventType::kScopeStartAsync, name, kValueZero, color, id);
   StartAsync(e.args[0], e.args[1], e.args[2], e.args[3], e.args[4], e.args[5]);
 }
 
-inline void StopAsync(uint64_t id) {
-  EncodedEvent e(EventType::kScopeStopAsync, /*name*/ nullptr, /*value*/ 0, orbit::Color::kAuto,
-                 id);
+inline void StopAsync(uint32_t id) {
+  EncodedEvent e(EventType::kScopeStopAsync, kNameNullPtr, kValueZero, kColorAuto, id);
   StopAsync(e.args[0], e.args[1], e.args[2], e.args[3], e.args[4], e.args[5]);
 }
 
@@ -244,7 +249,7 @@ inline void AsyncString(const char* str, uint32_t id, orbit::Color color) {
   constexpr size_t chunk_size = kMaxEventStringSize - 1;
   const char* end = str + strlen(str);
   while (str < end) {
-    EncodedEvent e(EventType::kString, /*name*/ nullptr, /*value*/ 0, color, id);
+    EncodedEvent e(EventType::kString, kNameNullPtr, kValueZero, color, id);
     std::strncpy(e.event.name, str, chunk_size);
     e.event.name[chunk_size] = 0;
     TrackValue(e.args[0], e.args[1], e.args[2], e.args[3], e.args[4], e.args[5]);
--- a/OrbitCore/Utils.h
+++ b/OrbitCore/Utils.h
@@ -74,8 +74,8 @@ inline std::string GetEnvVar(const char* a_Var) {
   return var;
 }
 
-inline uint64_t StringHash(const std::string& string) {
-  return XXH64(string.data(), string.size(), 0xBADDCAFEDEAD10CC);
+inline uint64_t StringHash(std::string_view str) {
+  return XXH64(str.data(), str.size(), 0xBADDCAFEDEAD10CC);
 }
 
 template <typename T, typename U>
--- a/OrbitCore/Utils.h
+++ b/OrbitCore/Utils.h
@@ -74,8 +74,8 @@ inline std::string GetEnvVar(const char* a_Var) {
   return var;
 }
 
-inline unsigned long long StringHash(const std::string& a_String) {
-  return XXH64(a_String.data(), a_String.size(), 0xBADDCAFEDEAD10CC);
+inline uint64_t StringHash(const std::string& string) {
+  return XXH64(string.data(), string.size(), 0xBADDCAFEDEAD10CC);
 }
 
 template <typename T, typename U>
--- a/OrbitGl/AsyncTrack.cpp
+++ b/OrbitGl/AsyncTrack.cpp
@@ -52,7 +52,7 @@ void AsyncTrack::OnTimer(const orbit_client_protos::TimerInfo& timer_info) {
   // Find the first row that that can receive the new timeslice with no overlap.
   // If none of the existing rows works, add a new row.
   uint32_t depth = 0;
-  while (max_span_time_by_depth_[depth] >= timer_info.start()) ++depth;
+  while (max_span_time_by_depth_[depth] > timer_info.start()) ++depth;
   max_span_time_by_depth_[depth] = timer_info.end();
 
   orbit_client_protos::TimerInfo new_timer_info = timer_info;
--- a/OrbitGl/AsyncTrack.cpp
+++ b/OrbitGl/AsyncTrack.cpp
@@ -1,0 +1,103 @@
+// Copyright (c) 2020 The Orbit Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#include "AsyncTrack.h"
+
+#include "App.h"
+#include "FunctionUtils.h"
+#include "GlCanvas.h"
+#include "ManualInstrumentationManager.h"
+#include "TimeGraph.h"
+#include "capture_data.pb.h"
+
+using orbit_client_protos::FunctionInfo;
+using orbit_client_protos::TimerInfo;
+
+AsyncTrack::AsyncTrack(TimeGraph* time_graph, const std::string& name) : TimerTrack(time_graph) {
+  SetName(name);
+  SetLabel(name);
+}
+
+[[nodiscard]] std::string AsyncTrack::GetBoxTooltip(PickingId id) const {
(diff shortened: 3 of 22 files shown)


Orbit.h
```diff
@@ -36,6 +36,8 @@
 #define ORBIT_START_ASYNC(name, id) ORBIT_START_ASYNC_WITH_COLOR(name, id, orbit::Color::kAuto)
```
ronaldfw, line 36: Sorry for asking some basic questions, I haven't been following this closely. 

Do "name" here and "str" play the same role? If yes, I'm not sure how, as a user, I would combine ORBIT_ASYNC_STRING and ORBIT_START_ASYNC. I assume that ORBIT_ASYNC_STRING just overrides the string that I pass as "name"?

OrbitCore/Utils.h
```diff
@@ -74,8 +74,8 @@ inline std::string GetEnvVar(const char* a_Var) {
   return var;
 }
 
-inline unsigned long long StringHash(const std::string& a_String) {
-  return XXH64(a_String.data(), a_String.size(), 0xBADDCAFEDEAD10CC);
+inline uint64_t StringHash(const std::string& string) {
```
ronaldfw, line 77: Can/should const std::string& be a string_view?

Orbit.h
```diff
@@ -224,15 +229,29 @@ inline void Stop() {
 }
 
 inline void StartAsync(const char* name, uint64_t id, orbit::Color color) {
```
ronaldfw, line 231: "id" is uint32_t in other signatures (also in the one you are calling a few lines below).

Orbit.h
ronaldfw: Looking at the code below, I guess as a  user, when I have registered a string with ORBIT_ASYNC_STRING, I'll just pass nullptr as "name", correct?

Is the idea to also support arbitrary length strings for the non-async case in the same way?

OrbitGl/AsyncTrack.h
```diff
@@ -0,0 +1,30 @@
+// Copyright (c) 2020 The Orbit Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#ifndef ORBIT_GL_ASYNC_TRACK_H_
+#define ORBIT_GL_ASYNC_TRACK_H_
+
+#include <string>
+
+#include "ThreadTrack.h"
```
ronaldfw, line 10: I think this should be TimerTrack.h (class below is derived from TimerTrack).

OrbitGl/AsyncTrack.h
```diff
@@ -0,0 +1,30 @@
...
+  void OnTimer(const orbit_client_protos::TimerInfo& timer_info) override;
+  void UpdateBoxHeight() override;
+
+ protected:
+  void SetTimesliceText(const orbit_client_protos::TimerInfo& timer, double elapsed_us, float min_x,
+                        TextBox* text_box) override;
+  [[nodiscard]] Color GetTimerColor(const orbit_client_protos::TimerInfo& timer_info,
+                                    bool is_selected) const override;
+
+  absl::flat_hash_map<uint32_t, uint64_t> max_span_time_by_depth_;
```
ronaldfw, line 27: Maybe add a comment why this is needed?

OrbitGl/AsyncTrack.cpp
```diff
@@ -0,0 +1,103 @@
...
+  SetLabel(name);
+}
+
+[[nodiscard]] std::string AsyncTrack::GetBoxTooltip(PickingId id) const {
+  const TextBox* text_box = time_graph_->GetBatcher().GetTextBox(id);
+  if (text_box == nullptr) return "";
+  ManualInstrumentationManager* manager = time_graph_->GetManualInstrumentationManager();
+  orbit_api::Event event = manager->ApiEventFromTimerInfo(text_box->GetTimerInfo());
+
+  const FunctionInfo* func =
```
ronaldfw, line 28: Is it documented somewhere that these are "fake" functions? Even if it is, I would suggest to add a reminder here. Reading this without that knowledge, the code here looks wrong ("why is it getting function names in the case of manual instrumentation?"). 

Something like:
"Reminder that for manually instrumented scopes, we record the name as a fake function name with a high address outside of the range of possible regular function addresses"

WDYT?

OrbitGl/AsyncTrack.cpp
```diff
@@ -0,0 +1,103 @@
...
+  if (collapse_toggle_->IsCollapsed() && depth_ > 0) {
+    box_height_ /= static_cast<float>(depth_);
+  }
+}
+
+void AsyncTrack::OnTimer(const orbit_client_protos::TimerInfo& timer_info) {
+  // Find the first row that that can receive the new timeslice with no overlap.
+  // If none of the existing rows works, add a new row.
+  uint32_t depth = 0;
+  while (max_span_time_by_depth_[depth] >= timer_info.start()) ++depth;
```
ronaldfw, line 55: (I guess these are the two lines one needs to understand why max_span_time_by_depth_ is needed.)

OrbitGl/TimeGraph.cpp
```diff
@@ -308,9 +320,8 @@ void TimeGraph::ProcessTimer(const TimerInfo& timer_info, const FunctionInfo* fu
     capture_max_timestamp_ = timer_info.end();
   }
 
-  if (function != nullptr && function->type() == FunctionInfo::kOrbitTrackValue) {
-    ProcessValueTrackingTimer(timer_info);
-    return;
+  if (function != nullptr && function->type() != FunctionInfo::kNone) {
```
ronaldfw, line 323: This confuses me a bit. It looks like ProcessOrbitFunctionTimer only handles manual instrumentation timers, but the naming here suggests that any timer with a proper function->type() (anything that is not kNone) is processed by that function. But regular timers for function calls are processed below.

ronaldfw: I guess that is how it's intended (the types are only meaningful for manual instrumentation). It would confuse me less if kNone were actually called kRegular, or "type" would be called "orbit_type".

ronaldfw (commented):
Are you going to add the code for making manual instr. appear in the live view in a later PR? (Fine with me, just want to double-check, because if I read things right, this is still missing. Or at least, some things are missing that are needed so that everything works correctly.)

OrbitGl/TimerTrack.h
```diff
@@ -27,7 +27,7 @@ class TimerTrack : public Track {
 
   // Pickable
   void Draw(GlCanvas* canvas, PickingMode picking_mode) override;
-  void OnTimer(const orbit_client_protos::TimerInfo& timer_info);
+  virtual void OnTimer(const orbit_client_protos::TimerInfo& timer_info);
```
dpallotti, line 30: Just want to say that this and `AsyncTrack` show how this is hardly is-a inheritance, but rather a way to share code with the parent class while overwriting some methods. Much bigger than this review, probably never-to-be-addressed, and maybe normal for this kind of ui work, but the relationship between these classes is getting more and more confusing.

Orbit.h
pierricgimmig: "name" here is actually the track name, I should rename the variable to make it clear. A user would never pass nullptr, no. ASYNC_STRING is used as a description for individual timeslices. For example, let's say we wanted to track file IO operations, the "name" (name of the track) would be "File IO operations" and the ASYNC_STRING macro would be used to pass in the full name of the requested file. All file IO operations will be under one single track, but we'll have information about the individual operations.

As for arbitrarily long strings for non async, since there is significant overhead, I'd rather not support it for now. This will be supported with the upcoming userspace dynamic instrumentation however.

Orbit.h
```diff
@@ -224,15 +229,29 @@ inline void Stop() {
...
 
+inline void AsyncString(const char* str, uint32_t id, orbit::Color color) {
+  if (str == nullptr) return;
+  constexpr size_t chunk_size = kMaxEventStringSize - 1;
+  const char* end = str + strlen(str);
+  while (str < end) {
+    EncodedEvent e(EventType::kString, /*name*/ nullptr, /*value*/ 0, color, id);
+    std::strncpy(e.event.name, str, chunk_size);
+    e.event.name[chunk_size] = 0;
+    TrackValue(e.args[0], e.args[1], e.args[2], e.args[3], e.args[4], e.args[5]);
```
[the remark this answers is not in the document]
pierricgimmig: I agree that there is quite a bit of overhead. It should probably be outlined somewhere, maybe in the name of the macro itself i.e. ASYNC_STRING_SLOW for example. However, this feature was requested from partner eng. for cases like file IO monitoring for example, where typical asset paths are very long. When we switch to custom dynamic instrumentation, we won't need to chunk up the strings anymore.

OrbitGl/AsyncTrack.cpp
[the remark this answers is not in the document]
pierricgimmig: No it doesn't assume anything. It keeps track of the current state of every row so that we know if adding a timer on that row would overlap or not. If we can't find a row with no overlap, then we create a new one.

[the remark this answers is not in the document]
pierricgimmig: Looking at the code, there is a bug however, ">=" should be ">". Adding a timer at time 0 would cause an infinite loop otherwise.

OrbitGl/TimerTrack.h
pierricgimmig: An AsyncTrack deals with timers (in the sense of time spans), why is it not a TimerTrack conceptually? Is the "TimerTrack" name the problem? The way I see it, AsyncTrack very much is-a TimerTrack, or a SpanTrack maybe!?

OrbitGl/AsyncTrack.cpp
pierricgimmig: Oh, actually, this is not exactly what is done here. I think this does need a comment though. Essentially, the FunctionInfo here will be one of the empty stubs in Orbit.h. It is used here to add information about the module from which the timer comes from.

OrbitGl/AsyncTrack.cpp
pierricgimmig: I still added a comment in the header file.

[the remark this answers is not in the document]
pierricgimmig: I think I see what you mean by assuming ordering. Currently, timespans are added in order of "stop" time. If the events were received in random order, then we would probably create a lot of rows. The algorithm here doesn't try to find available "gaps" in existing rows, that could be an improvement for another PR.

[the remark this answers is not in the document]
pierricgimmig: Oh no ok, now I understand what you mean... If we receive a stop event before it's corresponding start event, right? I think the assumption of ordered events is made elsewhere in the code too, didn't we have buffering in place to ensure ordering? Are you talking about the fact that because we are dealing with per-core buffers events could possibly be sent out of order?

OrbitGl/TimeGraph.h
```diff
@@ -213,6 +229,7 @@ class TimeGraph {
 
   std::shared_ptr<StringManager> string_manager_;
   StringManager manual_instrumentation_string_manager_;
+  std::unique_ptr<ManualInstrumentationManager> manual_instrumentation_manager_;
```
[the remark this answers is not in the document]
pierricgimmig: Yeah that's a good point

Orbit.h
pierricgimmig: Yes, thanks, fixed.

OrbitGl/AsyncTrack.h
pierricgimmig: Fixed, thanks.

OrbitGl/AsyncTrack.h
pierricgimmig: Done.

Orbit.h
```diff
@@ -224,15 +229,29 @@ inline void Stop() {
 }
 
 inline void StartAsync(const char* name, uint64_t id, orbit::Color color) {
-  EncodedEvent e(EventType::kScopeStartAsync, name, id, color);
+  EncodedEvent e(EventType::kScopeStartAsync, name, /*value*/ 0, color, id);
```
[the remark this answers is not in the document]
pierricgimmig: Good idea! Done.

OrbitGl/ManualInstrumentationManager.h
```diff
@@ -7,26 +7,26 @@
...
-    uint64_t arg_5 = timer_info.registers(5);
-    orbit_api::EncodedEvent encoded_event(arg_0, arg_1, arg_2, arg_3, arg_4, arg_5);
-    return encoded_event.event;
-  }
+      const orbit_client_protos::TimerInfo& timer_info);
+
+ private:
+  TimerInfoCallback timer_info_callback_;
+  absl::flat_hash_map<uint32_t, orbit_client_protos::TimerInfo> async_timer_info_start_by_id_;
+  StringManager string_manager_;
```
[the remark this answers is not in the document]
pierricgimmig: This is populated from a worker thread and accessed from the ui thread.

OrbitGl/TimeGraph.cpp
```diff
@@ -336,8 +347,29 @@ void TimeGraph::ProcessTimer(const TimerInfo& timer_info, const FunctionInfo* fu
   NeedsUpdate();
 }
 
+void TimeGraph::ProcessOrbitFunctionTimer(FunctionInfo::OrbitType type,
+                                          const TimerInfo& timer_info) {
+  switch (type) {
+    case FunctionInfo::kOrbitTrackValue:
+      ProcessValueTrackingTimer(timer_info);
+      break;
+    case FunctionInfo::kOrbitTimerStartAsync:
```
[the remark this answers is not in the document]
pierricgimmig: Done.

OrbitCore/Utils.h
pierricgimmig: Yes it can, I changed it, thanks.

OrbitGl/TimeGraph.cpp
pierricgimmig: I changed type to orbit_type.

OrbitGl/TimeGraph.h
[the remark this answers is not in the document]
pierricgimmig: Moved it to App.h

pierricgimmig:
> Are you going to add the code for making manual instr. appear in the live view in a later PR? (Fine with me, just want to double-check, because if I read things right, this is still missing. Or at least, some things are missing that are needed so that everything works correctly.)

Yes, this will come in another PR.

OrbitGl/AsyncTrack.cpp
pierricgimmig: Actually, I added a temporary variable, module_name, to make it clearer.

OrbitGl/ManualInstrumentationManager.cpp
```diff
@@ -3,3 +3,54 @@
...
+  uint64_t arg_5 = timer_info.registers(5);
+  orbit_api::EncodedEvent encoded_event(arg_0, arg_1, arg_2, arg_3, arg_4, arg_5);
+  return encoded_event.event;
+}
+
+void ManualInstrumentationManager::ProcessAsyncTimer(
+    const orbit_client_protos::TimerInfo& timer_info) {
+  orbit_api::Event event = ApiEventFromTimerInfo(timer_info);
+  if (event.type == orbit_api::kScopeStartAsync) {
+    async_timer_info_start_by_id_[event.id] = timer_info;
```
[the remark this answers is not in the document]
pierricgimmig: This allows to use the same dummy id in the case where the async events are known to be non-overlapping. This will be refined in a PR that will make the event processing more robust (allowing for start and end events to be received out of order).

OrbitGl/AsyncTrack.cpp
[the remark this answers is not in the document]
pierricgimmig: The code can be made more robust to handle the case where an "end" event is received before its corresponding "start". Will be addressed in subsequent PR.

Orbit.h
pierricgimmig: I have a PR coming up with proper documentation inside of Orbit.h that should make this clearer.

pierricgimmig:
Hi @ronaldfw and @dpallotti , thanks for your review. I addressed your comments, please see changes, cheers!

Orbit.h
ronaldfw: Ok, thanks for all the details!

OrbitGl/AsyncTrack.cpp
ronaldfw: Ah, I see, makes sense!

OrbitGl/TimerTrack.h
ronaldfw: Agree with @pierricgimmig here, but I also agree one has to be careful with (virtual) inheritance. It's easy to create a class hierarchy that is hard to use and where the primary reason for its existence is sharing of code. 

But I think there are some good reasons for having it the way it is done right now:
- The base class Track allows TimeGraph to just draw a track without case distinction (with some odd, minor exceptions). 
-  For users (person writing a new track), having TimerTrack actually makes things very easy and straightforward to do. In most cases you can start using TimerTrack and then gradually customize what you need (like tooltips). More generally, TimerTrack provides a framework how to write a new timer-based track.
- And, yes, reuse of code (UpdatePrimitives) is a reason here. I'm very much in favor of a functional style where UpdatePrimitives would just be a free function, but it would require to add a good amount of customization options (maybe passing functions into UpdatePrimitives?).

@dpallotti If you have suggestions how this can be improved, I'd be happy to discuss.

OrbitTest/OrbitTest.cpp
```diff
@@ -139,7 +162,15 @@ void OrbitTest::ManualInstrumentationApiTest() {
       ORBIT_DOUBLE(track_name.c_str(), cos(double_var * static_cast<double>(i)));
     }
 
-    std::this_thread::sleep_for(std::chrono::milliseconds(15));
+    // Asyn spans.
```
ronaldfw, line 165: Typo: s/Asyn/Async/g

ronaldfw (approved):
LGTM, thanks for addressing earlier comments!

OrbitGl/AsyncTrack.cpp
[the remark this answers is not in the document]
dpallotti: > Currently, timespans are added in order of "stop" time. If the events were received in random order, then we would probably create a lot of rows. The algorithm here doesn't try to find available "gaps" in existing rows, that could be an improvement for another PR.

This is what I meant. If the events are not received in order, the number of rows explodes. It's fine to assume that events come in order if they do (and they should as the service orders uprobes events), it would be a waste to make the algorithm much more complicated for a case that doesn't happen. But we need to be aware of the reason for the assumption. Maybe add a comment about the assumption.

Orbit.h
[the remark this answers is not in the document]
dpallotti: Then it should somehow be clear that this causes an unexpected overhead. The name of the macro is one possibility. We are talking microseconds for every 31 bytes, it has to be clear it should only be used in cases where it's really needed.

OrbitGl/TimerTrack.h
dpallotti: > An AsyncTrack deals with timers (in the sense of time spans), why is it not a TimerTrack conceptually?

Inheritance is not about the concept, e.g., the famous "a square is not a rectangle". Rather it's about about the behaviour, and this usually fails when using overriding to replace some behavior of the parent in order to share the code of other methods. That was my point. That's why in most cases a methods should either be pure virtual or final.

But this was just a general thought and not strictly related to this pull request. I realize how ui classes are very tricky in this regard and with so many methods adhering to a concept just for the sake of it makes things unnecessarily complicated. But it's something to keep in mind. E.g., sometimes instead of having a subclass replace a method of the superclass, you should have a superclass with two subclasses that express the two behaviors.

OrbitGl/AsyncTrack.cpp
pierricgimmig: I added a comment.

Orbit.h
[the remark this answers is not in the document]
pierricgimmig: The limitations and overhead will be detailed in subsequent PR.

OrbitTest/OrbitTest.cpp
pierricgimmig: fixed, thanks.

PR #1226, opened 2020-09-18
Join EventBuffer into CallstackData and remove EventBuffer

dpallotti:
Removes `EventBuffer`, `EventTracer`, `GEventTracer`.
Moves the functionality to `CallstackData`.
`CallstackEvent`s are no longer stored twice.
Also moves some methods from `CallstackData.h` to the `.cpp`.


OrbitCore/CallstackData.cpp
```diff
@@ -32,5 +117,51 @@ void CallstackData::AddCallStackFromKnownCallstackData(
...
+  }
+}
+
+absl::flat_hash_map<CallstackID, std::shared_ptr<CallStack>>
+CallstackData::GetUniqueCallstacksCopy() const {
+  absl::MutexLock lock(&unique_callstacks_mutex_);
+  return unique_callstacks_;
+}
+
+std::shared_ptr<CallStack> CallstackData::GetCallstackPtr(CallstackID callstack_id) const {
```
florian-kuebler, line 160: Can this be const? I think we should make it clear, that you can not modify while capturing.

florian-kuebler: Basically, I am aiming with CaptureData towards write only on the capture thread.

OrbitCore/CallstackData.cpp
```diff
@@ -32,5 +117,51 @@ void CallstackData::AddCallStackFromKnownCallstackData(
...
+void CallstackData::ForEachFrameInCallstack(uint64_t callstack_id,
+                                            const std::function<void(uint64_t)>& action) const {
+  absl::MutexLock lock(&unique_callstacks_mutex_);
+  for (uint64_t frame : unique_callstacks_.at(callstack_id)->GetFrames()) {
+    action(frame);
+  }
+}
+
+absl::flat_hash_map<CallstackID, std::shared_ptr<CallStack>>
+CallstackData::GetUniqueCallstacksCopy() const {
```
florian-kuebler, line 155: Is that still called/relevant? What is the user? Seems pretty expensive.

OrbitCore/CallstackData.cpp
```diff
@@ -32,5 +117,51 @@ void CallstackData::AddCallStackFromKnownCallstackData(
     absl::MutexLock lock(&unique_callstacks_mutex_);
     unique_callstacks_[hash] = std::move(unique_callstack);
   }
-  callstack_events_.push_back(CallstackEvent(event));
+  absl::MutexLock lock(&callstack_events_by_tid_mutex_);
```
florian-kuebler, line 120: Might this lead to a deadlock?

OrbitCore/CallstackData.cpp
```diff
@@ -20,8 +25,88 @@ void CallstackData::AddUniqueCallStack(CallStack call_stack) {
...
+absl::flat_hash_map<int32_t, uint32_t> CallstackData::GetCallstackEventsCountsPerTid() const {
+  absl::flat_hash_map<int32_t, uint32_t> counts;
+  for (const auto& tid_and_events : callstack_events_by_tid_) {
+    counts.emplace(tid_and_events.first, tid_and_events.second.size());
+  }
+  return counts;
+}
+
+uint32_t CallstackData::GetCallstackEventsOfTidCount(int32_t thread_id) const {
+  if (thread_id == SamplingProfiler::kAllThreadsFakeTid) {
```
florian-kuebler, line 45: As said in a different place. I am not sure, whether we should keep the allthreads data separate or just have it in the same map. looks like it making stuff slightly more complicated then needed.

OrbitCore/CallstackData.cpp
```diff
@@ -20,8 +25,88 @@ void CallstackData::AddUniqueCallStack(CallStack call_stack) {
...
-    const orbit_client_protos::CallstackEvent& event, const CallstackData* known_callstack_data) {
+uint32_t CallstackData::GetCallstackEventsCount() const {
+  uint32_t count = 0;
+  for (const auto& tid_and_events : callstack_events_by_tid_) {
+    count += tid_and_events.second.size();
+  }
+  return count;
+}
+
+absl::flat_hash_map<int32_t, uint32_t> CallstackData::GetCallstackEventsCountsPerTid() const {
```
florian-kuebler, line 36: That is creating a copy, which should reflect in the name. Where is this used?
Also you should add the mutex, right?

OrbitCore/CallstackData.cpp
```diff
@@ -20,8 +25,88 @@ void CallstackData::AddUniqueCallStack(CallStack call_stack) {
   unique_callstacks_[hash] = std::make_shared<CallStack>(std::move(call_stack));
 }
 
-void CallstackData::AddCallStackFromKnownCallstackData(
-    const orbit_client_protos::CallstackEvent& event, const CallstackData* known_callstack_data) {
+uint32_t CallstackData::GetCallstackEventsCount() const {
+  uint32_t count = 0;
```
florian-kuebler, line 29: Need a mutex lock?

florian-kuebler (commented):
Thank you very much for this! In general lgtm! I left some minor comments/questions.
One overall note from my side. We currently already have a hard time ensuring synchronization correctness of CaptureData, as we had reads (in particular when selecting callstacks while capturing). Now this gets more complex. We really need to be careful here. I am fine with this change now, as this is a big step forward, but we need to make sure that we tackle the synchronization of capture data.
Ideally, you should only write to capture data from capture thread, and all reads during capturing should be replaced by active pushes of that data. But this is a longer trail to go.

OrbitCore/CallstackData.cpp
dpallotti: In what case? Can you elaborate?

OrbitCore/CallstackData.cpp
dpallotti: It's used in `OrbitApp` and the result is passed to the constructor of `SamplingProfiler`, as `SamplingProfiler` holds its own `unique_callstacks_`. We might want to change this but it's probably part of a `SamplingProfiler` rework that we should do (even the name is obsolete...), but it's way out of scope here.

OrbitCore/CallstackData.cpp
dpallotti: What part should be const? This is a private method by the way, used only by `CallstackData::AddCallStackFromKnownCallstackData`, which is used when making a selection.

OrbitGl/TimeGraph.cpp
```diff
@@ -800,16 +799,16 @@ void TimeGraph::SetThreadFilter(const std::string& a_Filter) {
...
-    m_EventCount.clear();
-
-    for (auto& pair : GEventTracer.GetEventBuffer().GetCallstacks()) {
-      int32_t thread_id = pair.first;
-      const std::map<uint64_t, CallstackEvent>& callstacks = pair.second;
-      m_EventCount[thread_id] = callstacks.size();
-      GetOrCreateThreadTrack(thread_id);
-    }
+  m_EventCount.clear();
+  m_EventCount[SamplingProfiler::kAllThreadsFakeTid] =
```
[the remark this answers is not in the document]
dpallotti: You mean the map returned by `GetCallstackEventsCountsPerTid` should have an entry for `kAllThreadsFakeTid` as well? In that case I disagree. I'd prefer `CallstackData` not to have too much to do with `kAllThreadsFakeTid`, at least not in the data that it hold and returns (we could also just not make it handle `kAllThreadsFakeTid` at all and let the client call the correct methods but I found it more convenient for now). Or what do you mean?

OrbitCore/CallstackData.cpp
IrinaShkviro: Looks like we still use it to create the sample report on capture complete. Do you think we can get rid of it?

OrbitGl/TimeGraph.cpp
```diff
@@ -800,16 +799,16 @@ void TimeGraph::SetThreadFilter(const std::string& a_Filter) {
...
-    ScopeLock lock(GEventTracer.GetEventBuffer().GetMutex());
-    m_EventCount.clear();
-
-    for (auto& pair : GEventTracer.GetEventBuffer().GetCallstacks()) {
-      int32_t thread_id = pair.first;
-      const std::map<uint64_t, CallstackEvent>& callstacks = pair.second;
-      m_EventCount[thread_id] = callstacks.size();
-      GetOrCreateThreadTrack(thread_id);
-    }
+  m_EventCount.clear();
```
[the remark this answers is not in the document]
dpallotti: Done.

OrbitGl/TimeGraph.cpp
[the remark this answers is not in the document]
florian-kuebler: Yes I meant adding this entry. But also agree what you said. Maybe, you are right, and we should remove it here completely and let the client handle it.

OrbitCore/CallstackData.cpp
florian-kuebler: I mean if it can return a const shared_ptr, such that you can only call const methods on the underlying Callstack

OrbitCore/CallstackData.cpp
florian-kuebler: @dpallotti Okay, thanks for clarification. Yes should be targeted when doing SamplingProfiler refactoring.

OrbitCore/CallstackData.cpp
dpallotti: Making the `shared_ptr` `const` doesn't achieve that, the pointer becomes const, not the pointee. I can try to make the type argument const but that will require adding const to many other places, let me see.

IrinaShkviro: If I'm not mistaken, CallStack is always const: you can modify it after initialization.

dpallotti: True, it's immutable.

OrbitGl/TimeGraph.cpp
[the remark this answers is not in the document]
dpallotti: I've done that, I think it's cleaner to not have the dependency of `CallstackData` on `kAllThreadsFakeTid`. Please have a look.

OrbitCore/CallstackData.cpp
dpallotti: This is outdated now.

OrbitCore/CallstackData.h
```diff
@@ -32,63 +32,64 @@ class CallstackData {
...
-    return nullptr;
-  };
+  [[nodiscard]] std::shared_ptr<CallStack> GetCallstackPtr(CallstackID callstack_id) const;
 
- private:
-  BlockChain<orbit_client_protos::CallstackEvent, 16 * 1024> callstack_events_;
+  void RegisterTime(uint64_t time) {
+    if (time > max_time_) max_time_ = time;
+    if (time > 0 && time < min_time_) min_time_ = time;
+  }
```
IrinaShkviro, line 82: Should the implementation be in `.cpp`?

OrbitCore/CallstackData.cpp
```diff
@@ -20,8 +24,98 @@ void CallstackData::AddUniqueCallStack(CallStack call_stack) {
...
+  absl::MutexLock lock(&callstack_events_by_tid_mutex_);
+  for (const auto& tid_and_events : callstack_events_by_tid_) {
+    for (const auto& time_and_event : tid_and_events.second) {
+      action(time_and_event.second);
+    }
+  }
+}
+
+void CallstackData::ForEachCallstackEventOfTid(
+    int32_t desired_tid,
```
IrinaShkviro, line 105: Maybe just `tid`?

IrinaShkviro (approved):
lgtm, thanks a lot!

OrbitCore/CallstackData.cpp
florian-kuebler: Haven't looked too closely, but you essentially have two locks in a sequence, which is always at least prune to deadlocks. @dimitry- maybe you might have a brief look.

OrbitCore/CallstackData.cpp
florian-kuebler: I don't quite get why it does not make a copy? it is essentially returning a new map with new elements, right? However, feel free to keep the name, I don't have a very good idea though.

OrbitGl/TimeGraph.cpp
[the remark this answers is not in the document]
florian-kuebler: Yep, much better. thanks!

OrbitCore/CallstackData.cpp
dpallotti: Done.

florian-kuebler (approved):
Despite the double lock issue (where I would like to leave it for @dimitry- to decide), looks good to me.
Thanks again!

OrbitCore/CallstackData.h
dpallotti: Yep, done.

OrbitCore/CallstackData.cpp
dpallotti: But the previous one has already been unlocked.

OrbitCore/CallstackData.cpp
dpallotti: I just meant that the new object is not the same as an existing one, so it's not a copy, it's just a new object.

OrbitCore/CallstackData.cpp
dimitry-: I do not think there is a deadlock but do we want 2 separate mutexes in this class?

OrbitCore/CallstackData.h
```diff
@@ -32,63 +32,61 @@ class CallstackData {
...
+      int32_t tid, uint64_t time_begin, uint64_t time_end) const;
+
+  void ForEachCallstackEvent(
+      const std::function<void(const orbit_client_protos::CallstackEvent&)>& action) const;
+
+  void ForEachCallstackEventOfTid(
+      int32_t tid,
+      const std::function<void(const orbit_client_protos::CallstackEvent&)>& action) const;
+
+  [[nodiscard]] uint64_t max_time() const { return max_time_; }
```
dimitry-, line 60: I would protect this by mutex too, since you are modifying them under mutex.

OrbitGl/TimeGraph.h
```diff
@@ -150,59 +149,59 @@ class TimeGraph {
...
   // First member is id.
   absl::flat_hash_map<uint64_t, const TextBox*> iterator_text_boxes_;
   absl::flat_hash_map<uint64_t, const orbit_client_protos::FunctionInfo*> iterator_functions_;
 
-  double m_RefTimeUs = 0;
-  double m_MinTimeUs = 0;
-  double m_MaxTimeUs = 0;
+  double ref_time_us_ = 0;
+  double min_time_us_ = 0;
+  double max_time_us_ = 0;
```
dimitry-, line 163: what does "us" stand for here?

OrbitCore/CallstackData.cpp
florian-kuebler: okay. agree.

OrbitGl/TimeGraph.h
dpallotti: Microseconds.

OrbitCore/CallstackData.cpp
dpallotti: Should I use a single mutex for the entire class?

OrbitCore/CallstackData.h
dpallotti: Done.

OrbitCore/CallstackData.cpp
dimitry-: It most cases yes. The logic is goes like this (and there might be exceptions)

The data is the class is somehow interconnected so it should probably ensure that it does not end up in logically inconsistent state. And if data is not logically connected to each other - maybe it should be 2 classes instead of 1.

dpallotti: There's a single mutex now, please have a look. Had to make it reentrant though (see the comment), so I'm using `std::recursive_mutex`.

dpallotti: What's even the reason for not using reentrant locks by default? It seems to be unnecessarily prone to self deadlock.

dpallotti: Ok, so I've read a lot of stuff against reentrant mutexes. Still, I think that given the `ForEach...` methods here, this is necessary. Do you have thoughts on this?

OrbitCore/CallstackData.h
```diff
@@ -84,10 +84,10 @@ class CallstackData {
...
 
-  mutable absl::Mutex unique_callstacks_mutex_;
+  // Use a reentrant mutex so that calls to the ForEach... methods can be nested.
+  // E.g., one might want to nest ForEachCallstackEvent and ForEachFrameInCallstack.
+  mutable std::recursive_mutex mutex_;
   absl::flat_hash_map<CallstackID, std::shared_ptr<CallStack>> unique_callstacks_;
-
-  mutable absl::Mutex callstack_events_by_tid_mutex_;
   absl::flat_hash_map<int32_t, std::map<uint64_t, orbit_client_protos::CallstackEvent>>
       callstack_events_by_tid_;
```
dimitry-, line 93: remove atomics below?

dimitry- (approved):
otherwise lgtm

OrbitCore/CallstackData.h
dpallotti: Done.

PR #1227, opened 2020-09-18
Add basic tests for CaptureSerializer

florian-kuebler:


dimitry- (approved):

OrbitClientModel/CMakeLists.txt
```diff
@@ -32,7 +32,8 @@ target_compile_options(OrbitClientModelTests PRIVATE ${STRICT_COMPILE_FLAGS})
 
 
 target_sources(OrbitClientModelTests PRIVATE
-               CaptureDeserializerTest.cpp)
+        CaptureDeserializerTest.cpp
```
reichlfl, line 35: We have a bit of a mix of indentation there, but it should usually be like below, shouldn't it?
```
target_sources(
  OrbitClientModelTests
  PRIVATE CaptureDeserializerTest.cpp
          CaptureSerializerTest.cpp)
```

reichlfl (approved):
LGTM, minor comment, thanks!

OrbitClientModel/CMakeLists.txt
florian-kuebler: Well actually I think the one below was wrong. At least when looking at the formatting in other recent CMakeLists files.
We should really add clang-format checks for the cmake files though.

reichlfl: Ok, so apparently we have what I suggested mainly in OrbitQt and OrbitGl, which are the only ones I was looking at for reference... nvm then, but agree, clang-format for this would be nice, we shouldn't need to care about this, neither when writing nor when reviewing code.

florian-kuebler: Agree.

OrbitClientModel/CaptureSerializerTest.cpp
```diff
@@ -0,0 +1,151 @@
...
+  EXPECT_EQ(0, capture_info.callstacks_size());
+  EXPECT_EQ(0, capture_info.callstack_events_size());
+  EXPECT_EQ(0, capture_info.key_to_string_size());
+  EXPECT_EQ(0, capture_info.function_stats_size());
+}
+
+TEST(CaptureSerializer, GenerateCaptureInfo) {
+  int32_t process_id = 42;
+  std::string process_name = "p";
+  auto process = std::make_shared<Process>();
```
IrinaShkviro, line 62: It looks like @antonrohr is removing the Process in the PR. Did you discuss it or am I misunderstand it?

florian-kuebler: I know. Here I need that process. But it can be easily removed as soon Anton is doing a rebase.
I'll get in touch with him.

PR #1230, opened 2020-09-18
Change log file default behaviour in OrbitClientGgp

kisenshi:
_OrbitClientGgp_ is not expected to be run with admin privileges so the default functionality given initially (logs created in /var/log by default) does not work. 

From now on, by default only _stdout_ is used to print log information. However, the user can include an optional _debug_path_ parameter if they want to also create a debug file in the location provided.

### DIFF
--- a/OrbitClientGgp/main.cpp
+++ b/OrbitClientGgp/main.cpp
@@ -19,18 +19,18 @@ ABSL_FLAG(uint32_t, capture_length, 10, "duration of capture in seconds");
 ABSL_FLAG(std::vector<std::string>, functions, {},
           "Comma-separated list of functions to hook to the capture");
 ABSL_FLAG(std::string, file_name, "", "File name used for saving the capture");
-ABSL_FLAG(std::string, debug_path, "",
+ABSL_FLAG(std::string, log_directory, "",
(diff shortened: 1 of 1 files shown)


OrbitClientGgp/main.cpp
```diff
@@ -19,27 +19,33 @@ ABSL_FLAG(uint32_t, capture_length, 10, "duration of capture in seconds");
 ABSL_FLAG(std::vector<std::string>, functions, {},
           "Comma-separated list of functions to hook to the capture");
 ABSL_FLAG(std::string, file_name, "", "File name used for saving the capture");
+ABSL_FLAG(std::string, debug_path, "",
```
danielfenner, line 22: This is merely naming but I'd call this log_directory instead of debug_path and "log file" instead of "debug file".

danielfenner (approved):
lgtm besides nit picking about nameing.

OrbitClientGgp/main.cpp
kisenshi: oh yes this makes a lot of sense, will rename it, thanks!

kisenshi: Done, thanks!

kisenshi:
Merging this, thanks!

PR #1231, opened 2020-09-18
Save and Load the capture with the tracepoint events and markers

MariaSandru27:


OrbitClientModel/CaptureDeserializer.cpp
```diff
@@ -148,6 +148,28 @@ void LoadCaptureInfo(const CaptureInfo& capture_info, CaptureListener* capture_l
     capture_listener->OnCallstackEvent(std::move(callstack_event));
   }
 
+  for (const orbit_client_protos::TracepointInfo& tracepoint_info :
+       capture_info.tracepoint_infos()) {
+    if (*cancellation_requested) {
+      capture_listener->OnCaptureCancelled();
+      return;
+    }
+    orbit_grpc_protos::TracepointInfo tracepoint_info_translated;
```
florian-kuebler, line 157: It feels a bit strange that you are using a grpc proto for `TracepointInfo` on the Listener, and a client proto for `TracepointEventInfo`, but I see that it is the same for callstacks. So I guess you can leave it as is for now.

OrbitCore/CaptureData.h
```diff
@@ -125,11 +125,19 @@ class CaptureData {
...
+  };
+
   [[nodiscard]] const std::map<uint64_t, orbit_client_protos::TracepointEventInfo>&
   GetTracepointsOfThread(int32_t thread_id) const {
     return tracepoint_event_buffer_->GetTracepointsOfThread(thread_id);
   }
 
+  [[nodiscard]] const std::deque<orbit_client_protos::TracepointEventInfo> GetTracepointEvents()
+      const {
+    return tracepoint_event_buffer_->GetTracepointEvents();
```
florian-kuebler, line 139: From a caller perspective it is not clear that this returns a copy, so it can not be modified by a different thread. 
So I would always need to double-check, if it is safe to use that method. I'd rather introduce a `ForEachTracepointEvent` method, of which it is more obvious, that I am safe to use.

OrbitCore/CaptureData.h
```diff
@@ -125,11 +125,19 @@ class CaptureData {
     return tracepoint_info_manager_->Get(key);
   }
 
+  [[nodiscard]] const TracepointInfoManager* GetTracepointInfoManager() const {
+    return tracepoint_info_manager_.get();
```
florian-kuebler, line 129: Just out of curiosity, what is the reason, why the `tracepoint_info_manager-` and the `tracepoint_event_buffer_` are explicitly allocated on the heap using a unique pointer. Can't they just be *value* member variable of the `CaptureData`?

OrbitCore/TracepointEventBuffer.h
```diff
@@ -23,6 +23,16 @@ class TracepointEventBuffer {
   [[nodiscard]] const std::map<uint64_t, orbit_client_protos::TracepointEventInfo>&
   GetTracepointsOfThread(int32_t thread_id) const;
 
+  const std::deque<orbit_client_protos::TracepointEventInfo> GetTracepointEvents() {
+    std::deque<orbit_client_protos::TracepointEventInfo> tracepoint_events;
```
florian-kuebler, line 27: You should use the mutex here. Also, as argued above, I would use a `ForEachTracepointEvent` method instead.

florian-kuebler (commented):
Thanks for those changes. In general looks good.
But could you please extend the Tests for (de-)serialization? Also you should probably also save/load the selected tracepoints, as they are required during deserialization.

danielfenner (commented):
Looks good besides the issue with the mutex Florian already commented on.

OrbitClientModel/CaptureDeserializer.cpp
dimitry-: I kind of agree that we want to reuse grpc protos where possible. But this is a separate task. Currently client protos are completely independent from grpc ones.

dimitry- (approved):

OrbitCore/TracepointEventBuffer.h
MariaSandru27: Thank you so much for the advice! I will update the code!

OrbitCore/CaptureData.h
MariaSandru27: Thank you so much for the advice! I will update the code to have a 'ForEachTracepointEvent' method.

OrbitCore/CaptureData.h
MariaSandru27: Thank you for asking. A similar decision was made for the callstacks. The motivation was that  std::unique_ptr<> allows to move and copy data easier (especially relevant if there is stored an absl::Mutex inside).

OrbitClientModel/CaptureDeserializer.cpp
MariaSandru27: This is more related to the mechanism used to send data from the service to the client. If you want we can discuss this further in more detail.

MariaSandru27:
> Thanks for those changes. In general looks good.
> But could you please extend the Tests for (de-)serialization? Also you should probably also save/load the selected tracepoints, as they are required during deserialization.

Thank you for taking the time to review this. I was planning another PR for the tests about (de-)serialization. Also for now you can not select tracepoints so that is why there is no support for saving/loading on that regard.

OrbitClientModel/CaptureDeserializer.cpp
```diff
@@ -148,6 +155,28 @@ void LoadCaptureInfo(const CaptureInfo& capture_info, CaptureListener* capture_l
...
+    tracepoint_info_translated.set_category(tracepoint_info.category());
+    tracepoint_info_translated.set_name(tracepoint_info.name());
+    capture_listener->OnUniqueTracepointInfo(tracepoint_info.tracepoint_info_key(),
```
florian-kuebler, line 167: As we have the selected tracepoints in capture_data already, do we actually need to the OnUniqueTracepointInfo part at all? could't we use the ones from selected tracepoint to compute this mapping?

This is more of a general question and should not be addressed in this PR.

florian-kuebler (approved):
LGTM, thanks. 
From what I see, the "unique" tracepoints are kind of redundant in the whole system, as the information is already in the selected tracepoints. A follow up might be to remove them/or precompute them from the capture data. Also then the accessors (the for each) are probably not needed anymore, and the ones in capture data should be used.
But as I said, this is clearly a follow up

OrbitCore/TracepointInfoManager.h
```diff
@@ -23,6 +23,18 @@ class TracepointInfoManager {
   [[nodiscard]] orbit_grpc_protos::TracepointInfo Get(uint64_t key) const;
   [[nodiscard]] bool Contains(uint64_t key) const;
 
+  void ForEachUniqueTracepointInfo(
+      const std::function<void(const orbit_client_protos::TracepointInfo&)>& action) const {
+    absl::MutexLock lock(&unique_tracepoints_mutex_);
+    for (const auto& it : unique_tracepoint_) {
+      orbit_client_protos::TracepointInfo tracepoint_info;
```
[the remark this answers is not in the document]
MariaSandru27: Of course! Thank you!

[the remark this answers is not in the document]
MariaSandru27: As discussed offline, not a good idea.

danielfenner (approved):
lgtm!

PR #1233, opened 2020-09-21
Iterators are now in a QScrollArea

reichlfl:
Fixes the issue that iterators are appended to the live functions panel without displaying a scroll bar - the window just kept growing to fit all iterators.

To test the issue:
1) Capture with hooked functions
2) Add a bunch of iterators -> The window keeps growing and growing.

New behavior: Scrollbar appears when the available space is used up.

Fixed by adding a QScrollArea around the iterators, and some CSS to retain the original look. Also added a splitter to make the size of the Iterator area configurable. Background color of the scrollarea is hardcoded, which is a bit unfortunate, but apparently the color is not part of the palette values.

Result:
image

### DIFF
--- a/OrbitQt/orbitlivefunctions.ui
+++ b/OrbitQt/orbitlivefunctions.ui
@@ -19,7 +19,7 @@
   <property name="windowTitle">
    <string>Form</string>
   </property>
-  <layout class="QGridLayout" name="gridLayout">
+  <layout class="QVBoxLayout" name="verticalLayout_2">
    <property name="leftMargin">
     <number>0</number>
    </property>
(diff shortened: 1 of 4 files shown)


OrbitQt/orbitlivefunctions.ui
```diff
@@ -32,19 +32,103 @@
...
+       </item>
+       <item>
+        <widget class="QScrollArea" name="iteratorScrollArea">
+         <property name="autoFillBackground">
+          <bool>true</bool>
+         </property>
+         <property name="styleSheet">
+          <string notr="true">QScrollArea {
+	border: 0;
+}</string>
```
ronaldfw, line 71: Fix indentation here?

OrbitQt/orbitlivefunctions.ui
```diff
@@ -32,19 +32,103 @@
...
+            <height>187</height>
+           </rect>
+          </property>
+          <property name="autoFillBackground">
+           <bool>false</bool>
+          </property>
+          <property name="styleSheet">
+           <string notr="true">#iteratorFrame {
+	background-color: rgb(67, 67, 67);
+}</string>
```
ronaldfw, line 94: Same here?

ronaldfw (approved):
Awesome, thank you so much for fixing this!

OrbitQt/orbitlivefunctions.ui
reichlfl: QT Designer saves it like this and it will just re-appear if changes are made, so I think we should just leave it, even though it looks weird.

OrbitQt/orbitlivefunctions.ui
reichlfl: Same as above

PR #1234, opened 2020-09-21
Color adjustment of Capture Window Background

reichlfl:
The background color of the capture window was slightly different than the surrounding frame and the other tabs:

Surroundings: rgb(67, 67, 67)
Background: rgb(70, 70, 70)

Adjusted to make the colors match.

image
image

### DIFF
--- a/OrbitGl/CaptureWindow.cpp
+++ b/OrbitGl/CaptureWindow.cpp
@@ -532,7 +532,6 @@ void CaptureWindow::DrawScreenSpace() {
 
   // Right vertical margin.
   time_graph_.SetRightMargin(right_margin);
-  const Color kBackgroundColor(70, 70, 70, 255);
   float margin_x1 = getWidth();
   float margin_x0 = margin_x1 - right_margin;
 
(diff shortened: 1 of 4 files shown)


OrbitGl/CaptureWindow.cpp
```diff
@@ -544,7 +544,7 @@ void CaptureWindow::DrawScreenSpace() {
     Box box(Vec2(0, time_graph_.GetLayout().GetSliderWidth()),
             Vec2(getWidth(), time_graph_.GetLayout().GetTimeBarHeight()),
             GlCanvas::kZValueTimeBarBg);
-    ui_batcher_.AddBox(box, Color(70, 70, 70, 200));
+    ui_batcher_.AddBox(box, Color(67, 67, 67, 200));
```
ronaldfw, line 547: Should this be a defined constant of the class to make it clear that this and the color above have to match (and will match in the future)?

ronaldfw (approved):
Thanks for fixing this! Only thing I would consider is adding a constant (in CaptureWindow class) that defines the background color once and for all. WDYT?

OrbitGl/CaptureWindow.cpp
reichlfl: Good point, I introduced GlCanvas::kBackgroundColor (not in CaptureWindow, because all other constants are defined in GlCanvas - Track uses the same color and already includes constants from GlCanvas, so this seemed to be the better place). The timebar is using the same color with a different alpha value.

PR #1237, opened 2020-09-21
Text rendering fix

reichlfl:
Fixes the broken text background for the time bar.

To reproduce:
1) Load or take a capture
2) Scroll the capture window downwards to that the time bar is on top of any event bar
-> The event bar color is visible "through" the text of the time bar

Since depth write is enabled, the transparent parts of the text overwrite the existing z-values even though they are fully transparent. Commonly, depth-write is disabled for alpha-blended elements. This PR does exactly that and makes sure the text is rendered after the remaining UI elements so it is always rendered "on top".

In addition, it is taken care that all draw() functions correctly restore the GL states they touch, and this was moved into separate methods for clarity.

Before:
image

After:
image

The transparent time bar hides any text that is drawn underneath, which is imho a welcome side-effect because it improves readability of the time indicators.

### DIFF
--- a/OrbitGl/TriangleToggle.cpp
+++ b/OrbitGl/TriangleToggle.cpp
@@ -32,13 +32,13 @@ void TriangleToggle::Draw(GlCanvas* canvas, PickingMode picking_mode) {
 
     Triangle triangle;
     if (state_ == State::kCollapsed) {
-      triangle =
-          Triangle(position + Vec3(-half_h, half_w, 0.f), position + Vec3(-half_h, -half_w, 0.f),
-                   position + Vec3(half_w, 0.f, 0.f));
+      triangle = Triangle(position + Vec3(-half_h, half_w, GlCanvas::kZValueUi),
+                          position + Vec3(-half_h, -half_w, GlCanvas::kZValueUi),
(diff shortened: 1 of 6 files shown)


OrbitGl/TriangleToggle.cpp
```diff
@@ -32,13 +32,13 @@ void TriangleToggle::Draw(GlCanvas* canvas, PickingMode picking_mode) {
 
     Triangle triangle;
     if (state_ == State::kCollapsed) {
-      triangle =
-          Triangle(position + Vec3(-half_h, half_w, 0.f), position + Vec3(-half_h, -half_w, 0.f),
-                   position + Vec3(half_w, 0.f, 0.f));
+      triangle = Triangle(position + Vec3(-half_h, half_w, GlCanvas::kZValueUi),
```
karupayun, line 35: Thanks for that!

karupayun (approved):
Nice PR Florian! 👍 

Nothing to comment from my side.

ronaldfw (approved):
Awesome, thanks!

pierricgimmig (approved):
Awesome! Thanks for this Florian!

PR #1238, opened 2020-09-21
Remove broken CSs based on outer frame not matching thread's majority

dpallotti:
As discussed in http://b/162910324#comment5.
Bug: http://b/162910324

### DIFF
--- a/OrbitCore/CallstackData.cpp
+++ b/OrbitCore/CallstackData.cpp
@@ -183,6 +183,7 @@ void CallstackData::FilterCallstackEventsBasedOnMajorityStart() {
 
   for (auto& tid_and_events : callstack_events_by_tid_) {
     std::map<uint64_t, CallstackEvent>& callstack_events = tid_and_events.second;
+    const uint64_t count_for_this_thread = callstack_events.size();
 
     // Count the number of occurrences of each outer frame for this thread.
     absl::flat_hash_map<uint64_t, uint64_t> count_by_outer_frame;
@@ -211,7 +212,16 @@ void CallstackData::FilterCallstackEventsBasedOnMajorityStart() {
       }
     }
 
-    // Discard the CallstackEvents whose outer frame doesn't match the majority outer frame.
+    static constexpr double kFilterSupermajorityThreshold = 0.6;
+    if (majority_outer_frame_count < count_for_this_thread * kFilterSupermajorityThreshold) {
+      ERROR(
+          "Skipping filtering CallstackEvents for tid %d: majority outer frame has only %lu "
+          "occurrences out of %lu",
+          tid_and_events.first, majority_outer_frame_count, count_for_this_thread);
+      continue;
+    }
+
+    // Discard the CallstackEvents whose outer frame doesn't match the (super)majority outer frame.
     for (auto time_and_event_it = callstack_events.begin();
          time_and_event_it != callstack_events.end();) {
       const CallstackEvent& event = time_and_event_it->second;
--- a/OrbitCore/CallstackData.cpp
+++ b/OrbitCore/CallstackData.cpp
@@ -5,6 +5,7 @@
 #include "CallstackData.h"
 
 #include "Callstack.h"
+#include "absl/container/flat_hash_set.h"
 
 using orbit_client_protos::CallstackEvent;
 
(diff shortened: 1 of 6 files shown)


OrbitCore/CallstackData.cpp
```diff
@@ -175,3 +176,58 @@ std::shared_ptr<CallStack> CallstackData::GetCallstackPtr(CallstackID callstack_
...
+    // Find the outer frame with the most occurrences.
+    uint64_t majority_outer_frame = 0;
+    uint64_t majority_outer_frame_count = 0;
+    for (const auto& outer_frame_and_count : count_by_outer_frame) {
+      if (outer_frame_and_count.second > majority_outer_frame_count) {
+        majority_outer_frame = outer_frame_and_count.first;
+        majority_outer_frame_count = outer_frame_and_count.second;
+      }
+    }
+    if (majority_outer_frame_count == 0) {
```
dimitry-, line 209: nitpick: maybe check if it is empty before for loop?

dpallotti:
Please have a look.

OrbitCore/CallstackData.cpp
```diff
@@ -175,3 +176,58 @@ std::shared_ptr<CallStack> CallstackData::GetCallstackPtr(CallstackID callstack_
...
+      continue;
+    }
+
+    // Discard the CallstackEvents whose outer frame doesn't match the majority outer frame.
+    for (auto time_and_event_it = callstack_events.begin();
+         time_and_event_it != callstack_events.end();) {
+      const CallstackEvent& event = time_and_event_it->second;
+      const std::vector<uint64_t> frames =
```
dimitry-, line 217: &

OrbitCore/CallstackData.cpp
```diff
@@ -175,3 +176,58 @@ std::shared_ptr<CallStack> CallstackData::GetCallstackPtr(CallstackID callstack_
...
+    uint64_t majority_outer_frame_count = 0;
+    for (const auto& outer_frame_and_count : count_by_outer_frame) {
+      if (outer_frame_and_count.second > majority_outer_frame_count) {
+        majority_outer_frame = outer_frame_and_count.first;
+        majority_outer_frame_count = outer_frame_and_count.second;
+      }
+    }
+    if (majority_outer_frame_count == 0) {
+      continue;
+    }
```
dimitry-, line 211: maybe also check for supermajority here? majority_outer_frame_count > 0.6*number of callstacks in the thread.

Just a safeguard for distributions like (21, 20, 20, 19, 20) where 21 is winning but something this picture does not look right.

OrbitCore/CallstackData.cpp
```diff
@@ -175,3 +176,58 @@ std::shared_ptr<CallStack> CallstackData::GetCallstackPtr(CallstackID callstack_
...
+  uint32_t count_before_filtering = GetCallstackEventsCount();
+
+  for (auto& tid_and_events : callstack_events_by_tid_) {
+    std::map<uint64_t, CallstackEvent>& callstack_events = tid_and_events.second;
+
+    // Count the number of occurrences of each outer frame for this thread.
+    absl::flat_hash_map<uint64_t, uint64_t> count_by_outer_frame;
+    for (const auto& time_and_event : callstack_events) {
+      const CallstackEvent& event = time_and_event.second;
+      const std::vector<uint64_t> frames =
```
dimitry-, line 191: &

dimitry- (commented):
Also this function needs a test.

OrbitCore/CallstackData.cpp
dpallotti: Done.

OrbitCore/CallstackData.cpp
dpallotti: Done.

OrbitCore/CallstackData.cpp
dpallotti: What do you think the action should be in that case though? Should the filtering just not happen? Should all events be removed (that's a bit extreme)? Should the filtering stay the same but at least an error logged?

OrbitCore/CallstackData.cpp
dpallotti: Done.

dpallotti (commented):
> Also this function needs a test.

You're right, on it.

OrbitCore/CallstackData.cpp
dimitry-: I think filtering should not be happening at this point - maybe print a warning/error message.

dpallotti: Done, let me know what you think.

dimitry- (approved):

PR #1239, opened 2020-09-21
Better Chromium Notice file generation

beckerhe:
For the Chromium NOTICE file we need to add some more semantic information (Name of the dependency, URL to the Website of the project).

Three commits:
- The first commit changes the Qt-conan-package such that this additional information is extracted from the build tree. 
- The second commit integrates the updated conan package into Orbit and updates the lockfiles.
- The third commit adds logic for reading Chromium's metadata files (`README.chromium`) and generates a proper Chromium NOTICE file containing the mentioned extra pieces of information.


beckerhe:
Note to myself: Don't submit before the Qt-MSVC2019-builds are finished on the CI.

beckerhe:
> Note to myself: Don't submit before the Qt-MSVC2019-builds are finished on the CI.

Builds are all done. So this is ready to go after review.

contrib/python/conan_helpers.py
```diff
@@ -0,0 +1,53 @@
+# Copyright (c) 2020 The Orbit Authors. All rights reserved.
```
reichlfl, line 1: Comments on Python files: 
- Encoding hint at the beginning is missing (usually # -*- coding: utf-8 -*-): https://www.python.org/dev/peps/pep-0263/
- I'd suggest to use `"""Comment"""` instead of `# Comment` for non-inline comments such as the module comment on top as this is the official docstring style (https://www.python.org/dev/peps/pep-0257/)
- For our license comment, we could use __license__ and __author__

Which version of Python are we using, btw? For Python3, we could consider type hints... but since this is all basically helper code for our pipeline, it's probably not worth the effort, just a suggestion.

I know this is not the first python file we add to the repo, so this is probably the case for others as well. Are you following a styleguide here?

contrib/python/conan_helpers.py
```diff
@@ -0,0 +1,53 @@
...
+            parts = line.split(": ", 2)
+            if len(parts) != 2:
+                break
+
+            result[parts[0].lower()] = parts[1].strip()
+
+    return result
+
+
+# There are some typos in the chromium supplied license info.
```
reichlfl, line 21: See the comment on docstring above - a function comment should go underneath the `def` and contained in `"""`

contrib/python/conan_helpers.py
```diff
@@ -0,0 +1,53 @@
+# Copyright (c) 2020 The Orbit Authors. All rights reserved.
+# Use of this source code is governed by a BSD-style license that can be
+# found in the LICENSE file.
+
+import os
+
+def parse_chromium_readme(file_path):
```
reichlfl, line 7: Nit: This and `patch_chromium_readme` could be prefixed with _ as they are internal helper methods and do not need to be public

conanfile.py
```diff
@@ -195,6 +198,29 @@ def imports(self):
...
+        if not self.options.system_qt:
+            chromium_licenses = conan_helpers.gather_chromium_licenses(self.deps_cpp_info["qt"].rootpath)
+            chromium_licenses.sort(key=lambda license_info: license_info["name"].lower())
+
+            with open(os.path.join(self.install_folder, "NOTICE.Chromium.csv"), "w") as fd:
+                writer = csv.DictWriter(fd, fieldnames=["name", "url", "license"], extrasaction='ignore')
+                writer.writeheader()
+
+                for license in chromium_licenses:
+                    writer.writerow(license)
```
reichlfl, line 210: Should we check the entries for commas? Quite unlikely, but it could happen in the url, and we could urlencode the string to be sure.

reichlfl (commented):
Thanks for this! I have some comments on the Python files, most of them styling-related and I don't have a super strong opinion about them.

beckerhe:
Florian, thanks a lot for looking at the python stuff. I have to admit I didn't pay much attention, so thank you a lot for reviewing that. I will address your comments and maybe even run pylint before I get back to you.

conanfile.py
beckerhe: We don't need to check for commas because the csv-writer takes care of proper escaping of the CSV fields.
And we probably also don't want to do URL-encoding since this field sometimes contains explanatory text why no URL is available, so it's not always a URL.

contrib/python/conan_helpers.py
beckerhe: Regarding the python version: It's python-3 but might a pretty old version based on the docker container. From the top of my head I don't know if all containers are on 3.5. That's why I haven't looked into type-hints so far. In general I like this suggestion very much. :smiley: 

And we don't have a styleguide, neither for python, nor for shell or powershell. We should probably introduce some rules for all of them.

beckerhe: > For our license comment, we could use license and author

Can you explain that? Is there some standard pythonic way for license headers? Or do I miss something completely here? :see_no_evil:

contrib/python/conan_helpers.py
beckerhe: Good point. Done. Thanks!

contrib/python/conan_helpers.py
beckerhe: Done. Thanks!

beckerhe:
@reichlfl Could you have another look? Thanks a lot in advance!

contrib/python/conan_helpers.py
reichlfl: I thought this was standard, but apparently it's more of a best practice that some projects are following. The idea is to put this kind of information in module-level dunder variables: https://stackoverflow.com/questions/1523427/what-is-the-common-header-format-of-python-files

But since it's not standard, just including our default copyright notice should be fine, I'd just change it to be included in triple quotes.

conanfile.py
reichlfl: Right, good point, I forgot that the csv writer should take care of this. Disregard my comment :-)

reichlfl (approved):
Suggest to change the license hint to a docstring-style comment (triple quotes), otherwise LGTM, thanks!

contrib/python/conan_helpers.py
beckerhe: Ahh. Got it. Thanks. I have to admit I haven't seen this before. :see_no_evil: 
And as far as I remember we are advised to use the current license header (which was provided by the Google OpenSource team if I'm not mistaken.).

I will change the comments now! Thanks a lot!

beckerhe: By the way I'm planning on applying your suggestions on the whole `conanfile.py`. But there are changes coming up from me (in another branch) and from Cris (due to OrbitlClientGgp), so I will do this in a follow-up PR to avoid unnecessary conflict resolution (A lot of lines need to be shortened to proper length - This would probably mess with everyhting).

PR #1241, opened 2020-09-22
Scale track text when making a vertical zoom

karupayun:
The text inside OrbitGl always had the same font size, so it was impossible to scale track related text and not scale overlay text.
Now, we can have different font sizes and scale the text within the track in the same way as other objects.

Related to b/162797051.

Screenshots:
Screen Shot 2020-09-24 at 17 40 15
Screen Shot 2020-09-24 at 17 40 46

### DIFF
--- a/OrbitGl/CaptureWindow.cpp
+++ b/OrbitGl/CaptureWindow.cpp
@@ -768,7 +768,7 @@ void CaptureWindow::RenderTimeBar() {
       std::string text = GetPrettyTime(absl::Microseconds(current_micros));
       float world_x = time_graph_.GetWorldFromUs(current_micros);
       m_TextRenderer.AddText(text.c_str(), world_x + x_margin, world_y, GlCanvas::kZValueTextUi,
-                             Color(255, 255, 255, 255));
+                             Color(255, 255, 255, 255), GParams.font_size);
 
       Vec2 pos(world_x, world_y);
       ui_batcher_.AddVerticalLine(pos, height, GlCanvas::kZValueTextUi, Color(255, 255, 255, 255));
@@ -798,7 +798,7 @@ void CaptureWindow::RenderSelectionOverlay() {
     float pos_x = pos[0] + size[0];
 
     m_TextRenderer.AddText(text.c_str(), pos_x, select_stop_[1], GlCanvas::kZValueText, text_color,
-                           size[0], true);
+                           GParams.font_size, size[0], true);
 
     const unsigned char g = 100;
     Color grey(g, g, g, 255);
--- a/OrbitGl/TextRenderer.cpp
+++ b/OrbitGl/TextRenderer.cpp
@@ -58,11 +58,10 @@ void TextRenderer::Init() {
   const auto exe_dir = Path::GetExecutableDir();
   const auto fontFileName = exe_dir + "fonts/Vera.ttf";
 
-  static float fsize = GParams.font_size;
+  current_font_size_ = GParams.font_size;
   m_Buffer = vertex_buffer_new("vertex:3f,tex_coord:2f,color:4f");
-  m_Font = texture_font_new_from_file(m_Atlas, fsize, fontFileName.c_str());
-  current_font_size_ = static_cast<int>(fsize);
-  for (int i = 10; i <= 100; i += 1) {
+  m_Font = texture_font_new_from_file(m_Atlas, current_font_size_, fontFileName.c_str());
+  for (int i = 1; i <= 100; i += 1) {
     m_FontsBySize[i] = texture_font_new_from_file(m_Atlas, i, fontFileName.c_str());
   }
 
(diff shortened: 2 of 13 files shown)


OrbitGl/TimeGraph.h
```diff
@@ -107,7 +108,8 @@ class TimeGraph {
   void SetCanvas(GlCanvas* canvas);
   [[nodiscard]] GlCanvas* GetCanvas() { return canvas_; }
   void SetFontSize(int font_size);
-  [[nodiscard]] int GetFontSize() { return GetTextRenderer()->GetFontSize(); }
+  [[nodiscard]] int GetFontSize() { return lround((GParams.font_size) * layout_.GetScale()); }
+  [[nodiscard]] int GetOverlayFontSize() { return lround(GParams.font_size); }
```
pierricgimmig, line 112: These methods should be const and the return type should probably be uint32_t. I know that int was used before, but since we're here... :-)

pierricgimmig, line 112: I think it might be worth it to change the type of font_size in Params.h to uint32_t instead of float also.

pierricgimmig (commented):
Hi Pablo, I tested the change, this is super cool, thank you! I have minor comments concerning the type for font sizes. I think we should change them to uint32_t.

karupayun:
New commit: "Fix: Change font_size as uint32_t"

pierricgimmig (approved):
Great, thanks for the change Pablo!

OrbitGl/TextRenderer.cpp
```diff
@@ -82,15 +81,22 @@ void TextRenderer::Init() {
...
-    current_font_size_ = size;
+void TextRenderer::SetFontSize(uint32_t size) {
+  if (!m_FontsBySize.count(size)) {
+    auto iterator_next = m_FontsBySize.upper_bound(size);
+    // If there isn't that font_size in the map, we will search for the next one or previous one
+    if (iterator_next != m_FontsBySize.end()) {
+      size = iterator_next->first;
+    } else if (iterator_next != m_FontsBySize.begin()) {
+      size = (--iterator_next)->first;
+    }
```
reichlfl, line 92: Do we need an error-handling case here for `m_FontsBySize.begin() == m_FontsBySize.end()` (for the empty map)?

OrbitGl/TimeGraph.h
```diff
@@ -106,8 +107,11 @@ class TimeGraph {
   [[nodiscard]] StringManager* GetStringManager() { return string_manager_.get(); }
   void SetCanvas(GlCanvas* canvas);
   [[nodiscard]] GlCanvas* GetCanvas() { return canvas_; }
-  void SetFontSize(int font_size);
-  [[nodiscard]] int GetFontSize() { return GetTextRenderer()->GetFontSize(); }
+  void SetFontSize(uint32_t font_size);
+  [[nodiscard]] uint32_t GetFontSize() const {
+    return lround((GParams.font_size) * layout_.GetScale());
+  }
+  [[nodiscard]] uint32_t GetOverlayFontSize() const { return GParams.font_size; }
```
reichlfl, line 114: I think the change in `GetFontSize` and this additional getter is misleading. If I have a getter and a setter, I expect the getter to retrieve the value that I set (maybe clamped or adjusted to fit some constraints or the like), not a new value that is calculated from the existing one. Currently, if I call `time_graph_->SetFontSize(time_graph_->GetFontSize())` 10 times when zoomed out, the font size would constantly shrink.

I would keep the existing `GetFontSize` method as it is and introduce a new `CalculateZoomedFontSize` or similar that does the job.

OrbitGl/CaptureWindow.cpp
```diff
@@ -776,7 +776,7 @@ void CaptureWindow::RenderTimeBar() {
       std::string text = GetPrettyTime(absl::Microseconds(current_micros));
       float worldX = time_graph_.GetWorldFromUs(current_micros);
       m_TextRenderer.AddText(text.c_str(), worldX + xMargin, worldY, GlCanvas::kZValueTextUi,
-                             Color(255, 255, 255, 255));
+                             Color(255, 255, 255, 255), time_graph_.GetOverlayFontSize());
```
reichlfl, line 779: See comments for `TimeGraph::GetOverlayFontSize`. I would replace this with `GParams.font_size` directly, there's no need for the TimeGraph to pass through the global font size setting for the UI.

OrbitGl/CaptureWindow.cpp
```diff
@@ -806,7 +806,7 @@ void CaptureWindow::RenderSelectionOverlay() {
     float pos_x = pos[0] + size[0];
 
     m_TextRenderer.AddText(text.c_str(), pos_x, m_SelectStop[1], GlCanvas::kZValueText, text_color,
-                           size[0], true);
+                           time_graph_.GetOverlayFontSize(), size[0], true);
```
reichlfl, line 809: Same here

OrbitGl/TimeGraph.cpp
```diff
@@ -616,7 +616,7 @@ void DrawIteratorBox(GlCanvas* canvas, Vec2 pos, Vec2 size, const Color& color,
   float max_size = size[0];
   canvas->GetTextRenderer().AddTextTrailingCharsPrioritized(
       text.c_str(), pos[0] + kLeftOffset, text_y + kAdditionalSpaceForLine, GlCanvas::kZValueText,
-      Color(255, 255, 255, 255), time.length(), max_size);
+      Color(255, 255, 255, 255), time.length(), GCurrentTimeGraph->GetOverlayFontSize(), max_size);
```
reichlfl, line 619: Same here

reichlfl (commented):
This is super useful, thanks for changing this! I have an issue with the changed GetFontSize implementation, can you look into this?

OrbitGl/TextRenderer.cpp
karupayun: I don't think so. We are filling the array at the first and we are not touching it anymore. But a check doesn't hurt.

OrbitGl/TimeGraph.h
karupayun: ok. Sounds good! Thanks for that!

reichlfl (approved):
LGTM, thanks for the changes!

PR #1244, opened 2020-09-22
Tests for CaptureSerializer

MariaSandru27:


OrbitClientModel/CaptureSerializerTest.cpp
```diff
@@ -70,7 +72,12 @@ TEST(CaptureSerializer, GenerateCaptureInfo) {
   selected_function.set_module_base_address(15);
   selected_functions[FunctionUtils::GetAbsoluteAddress(selected_function)] = selected_function;
 
+  TracepointInfo selected_tracepoint_info;
```
florian-kuebler, line 75: Nit: I think it would improve readability if you would move this one line below, such that all operations on selected_tracepoint_info are together.
Feel free to ignore, though.

OrbitClientModel/CaptureSerializerTest.cpp
```diff
@@ -129,6 +148,20 @@ TEST(CaptureSerializer, GenerateCaptureInfo) {
...
   EXPECT_EQ(callstack_event.callstack_hash(), actual_callstack_event.callstack_hash());
 
+  ASSERT_EQ(1, capture_info.tracepoint_infos_size());
+  const orbit_client_protos::TracepointInfo& actual_tracepoint_info =
+      capture_info.tracepoint_infos(0);
+  ASSERT_EQ(actual_tracepoint_info.category(), selected_tracepoint_info.category());
+  ASSERT_EQ(actual_tracepoint_info.name(), selected_tracepoint_info.name());
+  ASSERT_EQ(actual_tracepoint_info.tracepoint_info_key(), 0);
+
+  ASSERT_EQ(2, capture_info.tracepoint_event_infos_size());
```
danielfenner, line 158: Please add a one line comment here with the explanation from the commit message. This is somewhat unexpected here otherwise.

danielfenner (approved):

OrbitClientModel/CaptureSerializerTest.cpp
```diff
@@ -129,6 +148,20 @@ TEST(CaptureSerializer, GenerateCaptureInfo) {
   EXPECT_EQ(callstack_event.time(), actual_callstack_event.time());
   EXPECT_EQ(callstack_event.callstack_hash(), actual_callstack_event.callstack_hash());
 
+  ASSERT_EQ(1, capture_info.tracepoint_infos_size());
+  const orbit_client_protos::TracepointInfo& actual_tracepoint_info =
+      capture_info.tracepoint_infos(0);
+  ASSERT_EQ(actual_tracepoint_info.category(), selected_tracepoint_info.category());
```
florian-kuebler, line 154: Nit: These can be EXPECT_EQ rather then ASSERT_EQ.
The difference is, that the execution will be stopped, if an assert fails. On expect, only the failure will be reported, but the test continues.
This is why, the size checks are asserts, as we afterwards, access the data.

florian-kuebler (approved):
LGTM. Thanks. I added two nit-picking comments. It's up to you, if you address them.

OrbitClientModel/CaptureSerializerTest.cpp
MariaSandru27: Of course! Thank you! I will.

OrbitClientModel/CaptureSerializerTest.cpp
MariaSandru27: Thank you so much! I will update the code!

OrbitClientModel/CaptureSerializerTest.cpp
MariaSandru27: I will update the code. Thank you!

dimitry- (approved):

PR #1245, opened 2020-09-22
Minor Improvements in SamplingProfiler

florian-kuebler:
Prior to this, GetCallstacksFromAddress was modifying the total
callstack count, which would not be expected from a getter.
Now the caller takes care of that (same runtime costs).

Also GetSortedCallstacksFromAddress was returning a shared_ptr,
which was only stored in the SamplingReport. So unique_ptr makes
the owner ship more clearer.

Bug: http://b/166238019


IrinaShkviro (approved):
looks good from my side, thanks!

PR #1246, opened 2020-09-22
Customise the tracepoints markers in the capture

MariaSandru27:
compute the points of a unit circle only once for the lifetime 
of the app
scale those points by "radius" once per track 
translate the points once per tracepoint rendering

Screenshot 2020-09-19 at 1 42 56 PM
Screenshot 2020-09-19 at 1 47 48 PM
Screenshot 2020-09-19 at 1 48 12 PM

### DIFF
--- a/OrbitGl/TracepointTrack.cpp
+++ b/OrbitGl/TracepointTrack.cpp
@@ -69,6 +69,8 @@ void TracepointTrack::UpdatePrimitives(uint64_t min_tick, uint64_t max_tick,
 
   const Color kWhite(255, 255, 255, 255);
 
+  const Color kTracepoint(255, 255, 255, 155);
+
   const Color kGreenSelection(0, 255, 0, 255);
 
   if (!picking) {
@@ -76,7 +78,27 @@ void TracepointTrack::UpdatePrimitives(uint64_t min_tick, uint64_t max_tick,
          ++it) {
       uint64_t time = it->first;
       Vec2 pos(time_graph_->GetWorldFromTick(time), pos_[1]);
-      batcher->AddVerticalLine(pos, -track_height, z, kWhite);
+      batcher->AddVerticalLine(pos, -track_height, z, kTracepoint);
+      pos[1] -= track_height / 2;
+
+      float x_pos = pos[0];
+      float y_pos = pos[1];
+      float radius = track_height / 2.5;
+      const float steps = 1000;
+      const float angle = (kPiFloat * 2.f) / steps;
+      float prev_x = x_pos;
+      float prev_y = y_pos - radius;
+      for (int i = 1; i <= steps; i++) {
+        float new_x = pos[0] + radius * sinf(angle * i);
+        float new_y = pos[1] - radius * cosf(angle * i);
+        Vec3 x_1 = Vec3(pos[0], pos[1], z);
+        Vec3 x_2 = Vec3(prev_x, prev_y, z);
+        Vec3 x_3 = Vec3(new_x, new_y, z);
+        Triangle triangle(x_1, x_2, x_3);
+        batcher->AddTriangle(triangle, kTracepoint);
+        prev_x = new_x;
+        prev_y = new_y;
+      }
     }
   } else {
     constexpr float kPickingBoxWidth = 9.0f;
--- a/OrbitGl/TracepointTrack.cpp
+++ b/OrbitGl/TracepointTrack.cpp
@@ -15,7 +15,7 @@ TracepointTrack::TracepointTrack(TimeGraph* time_graph, int32_t thread_id)
   for (int i = 1; i <= kSteps; i++) {
     float new_x = sinf(angle * i);
     float new_y = cosf(angle * i);
-    circle_points.emplace_back(std::make_pair(new_x, new_y));
+    circle_points.emplace_back(Vec2(new_x, new_y));
   }
 }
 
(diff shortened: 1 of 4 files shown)


OrbitGl/TracepointTrack.cpp
```diff
@@ -69,14 +69,38 @@ void TracepointTrack::UpdatePrimitives(uint64_t min_tick, uint64_t max_tick,
...
+
   const Color kGreenSelection(0, 255, 0, 255);
 
   if (!picking) {
     for (auto it = tracepoints.lower_bound(min_tick); it != tracepoints.upper_bound(max_tick);
          ++it) {
       uint64_t time = it->first;
       Vec2 pos(time_graph_->GetWorldFromTick(time), pos_[1]);
-      batcher->AddVerticalLine(pos, -track_height, z, kWhite);
+      batcher->AddVerticalLine(pos, -track_height / 4, z, kTracepoint);
```
danielfenner, line 81: I guess it is easier and faster to just draw the entire line. The part in the middle will be overdrawn by the circle anyway.

danielfenner (approved):
Lgtm. Comments are merely nitpicking. 

I don't have a good intuition how expensive the drawing operations are. Have you tried this under load e.g. instrumenting sched:sched_switch?

@pierricgimmig might have a gut feeling here.

OrbitGl/TracepointTrack.cpp
MariaSandru27: Not really, I have tried it and because of the alphas the line in the circle gets lighter and visible. @pierricgimmig  may know more about this because he suggested to split the line.

pierricgimmig: Yeah, because the circle has transparency, drawing a single line looks weird...

OrbitGl/TracepointTrack.cpp
```diff
@@ -69,14 +69,38 @@ void TracepointTrack::UpdatePrimitives(uint64_t min_tick, uint64_t max_tick,
...
+      batcher->AddVerticalLine(Vec2(pos[0], pos[1] - track_height), track_height / 4, z,
+                               kTracepoint);
+      pos[1] -= track_height / 2;
+
+      float x_pos = pos[0];
+      float y_pos = pos[1];
+      float radius = track_height / 4;
+      const float steps = 22;
```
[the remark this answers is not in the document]
pierricgimmig: Also, "track_height / 4" should be stored in a temporary variable instead of being computed 3 times.

OrbitGl/TracepointTrack.cpp
```diff
@@ -69,14 +69,38 @@ void TracepointTrack::UpdatePrimitives(uint64_t min_tick, uint64_t max_tick,
...
+      const float angle = (kPiFloat * 2.f) / steps;
+      float prev_x = x_pos;
+      float prev_y = y_pos - radius;
+      for (int i = 1; i <= steps; i++) {
+        float new_x = pos[0] + radius * sinf(angle * i);
+        float new_y = pos[1] - radius * cosf(angle * i);
```
pierricgimmig, line 95: It would be better to compute the points of the circle once for all the tracepoints of the track. Even better, we could compute the points of a unit circle only once for the lifetime of the app, then scale those points by "radius" once per track "UpdatePrimitives", and then finally translate those points once per tracepoint. This would avoid performing costly trigonometric functions for every tracepoint on every track.

pierricgimmig (commented):
Hi Maria, the new tracepoint rendering looks great! Thanks. Please see comments concerning performance.

OrbitGl/TracepointTrack.cpp
ronaldfw: +1

OrbitGl/TracepointTrack.cpp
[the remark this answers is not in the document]
MariaSandru27: Sure! Thank you!

OrbitGl/TracepointTrack.cpp
```diff
@@ -69,14 +69,38 @@ void TracepointTrack::UpdatePrimitives(uint64_t min_tick, uint64_t max_tick,
...
+        Vec3 x_1 = Vec3(pos[0], pos[1], z);
```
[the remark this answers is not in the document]
MariaSandru27: Of course! Thank you!

OrbitGl/TracepointTrack.cpp
MariaSandru27: Thank you so much! I will update the code!

OrbitGl/TracepointTrack.h
```diff
@@ -26,6 +28,7 @@ class TracepointTrack : public EventTrack {
 
  private:
   [[nodiscard]] bool HasTracepoints() const;
+  std::deque<std::pair<float, float>> circle_points;
```
pierricgimmig, line 31: std::vector<Vec2>? A vector would do fine here since we don't need to insert or remove elements. Also we have the Vec2 class that can be used for the point coordinates.

pierricgimmig (commented):
Hi Maria, thanks for the change, please see comment.

OrbitGl/TracepointTrack.h
MariaSandru27: Thank you @pierricgimmig for pointing this out. I have updated the code.

dimitry- (approved):

pierricgimmig (approved):
Thanks for the changes Maria! LGTM

MariaSandru27:
> Thanks for the changes Maria! LGTM

Thank you for taking the time to review!

PR #1247, opened 2020-09-22
Sort CallstackReports by Sample Count

florian-kuebler:
Prior to this, callstack reports were attempted to be sorted by thread usage.
However, this was basically never set (except for the all thread case).

As the thread usage is also not used anywhere else, this removes this
information and uses the sample count for sorting (as already implicitly
doe in top down view).
This also simplifies the code.

Bug: http://b/169135142
Test: Take a Capture and see that the thread taps in the sampling
report are now actually sorted by number of samples.


pierricgimmig (approved):
Thanks for this Florian, good catch :-)

PR #1248, opened 2020-09-22
Add bottom-up view

dpallotti:
#### Prepare OrbitApp, OrbitMainWindow for bottom-up view
Bug: http://b/168018191
#### Implement CallTreeView::CreateBottomUpViewFromSamplingProfiler
Bug: http://b/168018191
#### Hide "Exclusive" in bottom-up view, hide "Of parent" for first level
`CallTreeViewItemModel` handles `kExclusive`, `kOfParent` for threads too.
These values are now hidden by `CallTreeWidget` based on different logics in
case of a top-down or a bottom-up view.
- Top-down view: for the elements at the first level (i.e., the threads),
  "Exclusive" and "Of parent" are hidden.
- Bottom-up view: the "Exclusive" column is hidden entirely; for elements at the
  first level (i.e., the innermost functions), "Of parent" is also hidden.

---

Screenshot from Trata:

Capture

The number of tabs is not getting any lower...

### DIFF
--- a/OrbitGl/TopDownView.cpp
+++ b/OrbitGl/TopDownView.cpp
@@ -4,40 +4,63 @@
 
 #include "TopDownView.h"
 
-TopDownFunction* TopDownInternalNode::GetFunctionOrNull(uint64_t function_absolute_address) {
-  auto top_down_function_it = function_nodes_.find(function_absolute_address);
-  if (top_down_function_it == function_nodes_.end()) {
-    return nullptr;
+std::vector<const TopDownNode*> TopDownNode::children() const {
+  std::vector<const TopDownNode*> ret;
+  for (const auto& tid_and_thread : thread_children_) {
+    ret.push_back(&tid_and_thread.second);
   }
-  return &top_down_function_it->second;
+  for (auto& address_and_functions : function_children_) {
+    ret.push_back(&address_and_functions.second);
+  }
+  return ret;
 }
 
-TopDownFunction* TopDownInternalNode::AddAndGetFunction(uint64_t function_absolute_address,
-                                                        std::string function_name,
-                                                        std::string module_path) {
-  function_nodes_.insert_or_assign(
-      function_absolute_address,
-      TopDownFunction{function_absolute_address, std::move(function_name), std::move(module_path),
-                      this});
-  return &function_nodes_.at(function_absolute_address);
-}
-
-TopDownThread* TopDownView::GetThreadOrNull(int32_t thread_id) {
-  auto top_down_thread_it = thread_nodes_.find(thread_id);
-  if (top_down_thread_it == thread_nodes_.end()) {
+TopDownThread* TopDownNode::GetThreadOrNull(int32_t thread_id) {
+  auto top_down_thread_it = thread_children_.find(thread_id);
+  if (top_down_thread_it == thread_children_.end()) {
     return nullptr;
   }
   return &top_down_thread_it->second;
 }
 
-TopDownThread* TopDownView::AddAndGetThread(int32_t thread_id, std::string thread_name) {
-  thread_nodes_.insert_or_assign(thread_id, TopDownThread{thread_id, std::move(thread_name), this});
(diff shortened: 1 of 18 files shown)


beckerhe:
Very nice! Short question upfront: Should this go into 1.53? I'm currently working on a bugfix which definitely should go into 1.53, so I need to prioritize a bit. That's why I'm asking.

dpallotti:
> Very nice! Short question upfront: Should this go into 1.53? I'm currently working on a bugfix which definitely should go into 1.53, so I need to prioritize a bit. That's why I'm asking.

I was thinking about 1.54 already, I'm not in a rush with this.

OrbitQt/CallTreeWidget.h
```diff
@@ -46,6 +46,65 @@ class CallTreeWidget : public QWidget {
   static const QString kActionDisassembly;
   static const QString kActionCopySelection;
 
+  template <
+      typename HideValuesProxyModelT,
+      typename = std::enable_if_t<std::is_base_of_v<QIdentityProxyModel, HideValuesProxyModelT>>>
```
beckerhe, line 51: If I'm not mistaken you want the compilation to fail if `HideValuesProxyModelT` does not subclass from QIdentityProxyModel, right?

If yes you probably don't need SFINAE here. A `static_assert` inside the function would be much nicer since you can also specify an error message.

You only need SFINAE if you wanted to remove this function from an overload set, meaning there were another `SetCallTreeView(...)` overload which should be picked instead of this one.

OrbitQt/CallTreeWidget.h
```diff
@@ -46,6 +46,65 @@ class CallTreeWidget : public QWidget {
...
+  void SetCallTreeView(std::unique_ptr<CallTreeView> call_tree_view) {
+    CHECK(app_ != nullptr);
+
+    model_ = std::make_unique<CallTreeViewItemModel>(std::move(call_tree_view));
+
+    hide_values_proxy_model_ = std::make_unique<HideValuesProxyModelT>(nullptr);
+    hide_values_proxy_model_->setSourceModel(model_.get());
```
beckerhe, line 58: Have you considered passing in this proxy model as an argument instead of constructing it here? This would mean the `HideValues`-proxy models get implementation details and can be moved to the cpp file. The template here is probably also not needed in this case. But maybe I'm missing something. :see_no_evil:

OrbitGl/TopDownView.cpp
```diff
@@ -56,7 +79,7 @@ static void AddCallstackToTopDownThread(TopDownThread* thread_node,
                                         const CallStack& resolved_callstack,
                                         uint64_t callstack_sample_count,
                                         const CaptureData& capture_data) {
-  TopDownInternalNode* current_thread_or_function = thread_node;
+  TopDownNode* current_thread_or_function = thread_node;
```
beckerhe: Could you add some words to the commit message explaining what's happening in the first commit? I've seen you've merged Node and InternalNode and thread_children_/function_children_ have been introduced. But it would be great to understand the overall rationale.

beckerhe (commented):
Very nice feature. Thank you so much.
The code also looks good to me, but I need to go through it one more time.

I added some minor comments.

OrbitGl/TopDownView.cpp
dpallotti: Done.

OrbitQt/CallTreeWidget.h
dpallotti: You're right, for some reason I was assuming I was setting the underlying model during construction, but that obviously happens with `setSourceModel`. Done.

OrbitQt/CallTreeWidget.h
dpallotti: Good point. Either way I removed the type parameter alrogether as per the other comment.

beckerhe (approved):
Thanks for the changes, especially for the commit message.

I had another look and all looks good to me.

PR #1249, opened 2020-09-22
Removes "Using" from a Header File and Unused "DataManager::set_selected_functions"

florian-kuebler:


antonrohr (approved):

dimitry- (approved):
Please change the commit message to something like:

This removes using from a header file and an unused DataManager::set_selected_functions

PR #1250, opened 2020-09-22
Fix copy-and-paste bug in orbit_service::Process

beckerhe:
This bug/crash only appeared on the error case because the error message
generating logic was accessing the wrong result variable, most likely
introduced due to copy-and-paste.

Test: Compiles. Tests run.
Bug: http://b/169121244

Anton, I will assign this to you for review - and to a random person on the team.

### DIFF
--- a/OrbitService/Process.cpp
+++ b/OrbitService/Process.cpp
@@ -77,7 +77,7 @@ ErrorMessageOr<Process> Process::FromPid(pid_t pid) {
   auto cmdline_file_result = utils::ReadFileToString(cmdline_file_path);
   if (!cmdline_file_result) {
     return ErrorMessage{absl::StrFormat("Failed to read %s: %s", cmdline_file_path.string(),
-                                        name_file_result.error().message())};
+                                        cmdline_file_result.error().message())};
   }
 
   std::string cmdline = std::move(cmdline_file_result.value());


beckerhe:
I'm currently thinking about improving the tests here: Ideally we test all the error branches. But since this code reads from the "real" /proc this will be hard to do without mocking "/proc". Any ideas/suggestions?

florian-kuebler:
> I'm currently thinking about improving the tests here: Ideally we test all the error branches. But since this code reads from the "real" /proc this will be hard to do without mocking "/proc". Any ideas/suggestions?

Well we could mock utils::ReadFileToString and make it return what we would like to return. As this is non-virtual, this is a bit tricky though. But I guess we should go that direction.

florian-kuebler (approved):
Thanks for fixing this!

antonrohr (approved):
I like flos suggestion with mocking ReadFileToString, but this should only be an addition and at least one test reading the /proc should be there. 

Thank you for fixing this bug!

beckerhe:
> I like flos suggestion with mocking ReadFileToString, but this should only be an addition and at least one test reading the /proc should be there.
> 
> Thank you for fixing this bug!

Yeah, mocking ReadFileToString can be done easily, even without virtual functions but it also complicates the code quite a bit. Ideally unit test should make reading code easier, not harder.

That's why I'm currently thinking about making the procfs-mountpoint adjustable. Right now it's hard coded to `/proc`. Making this changeable would allow tests to point to some prepared fake-directory and the cognitive overhead is similar to removing a magic constant. WDYT?

Anyway I'm not planning on doing this in this PR because this fix should go into 1.53

florian-kuebler:
> > I like flos suggestion with mocking ReadFileToString, but this should only be an addition and at least one test reading the /proc should be there.
> > Thank you for fixing this bug!
> 
> Yeah, mocking ReadFileToString can be done easily, even without virtual functions but it also complicates the code quite a bit. Ideally unit test should make reading code easier, not harder.
> 
> That's why I'm currently thinking about making the procfs-mountpoint adjustable. Right now it's hard coded to `/proc`. Making this changeable would allow tests to point to some prepared fake-directory and the cognitive overhead is similar to removing a magic constant. WDYT?
> 
> Anyway I'm not planning on doing this in this PR because this fix should go into 1.53

SGTM. I see that making non-virtual functions mockable, makes the code less readable, so this sounds like a plan.

beckerhe:
> > > I like flos suggestion with mocking ReadFileToString, but this should only be an addition and at least one test reading the /proc should be there.
> > > Thank you for fixing this bug!
> > 
> > 
> > Yeah, mocking ReadFileToString can be done easily, even without virtual functions but it also complicates the code quite a bit. Ideally unit test should make reading code easier, not harder.
> > That's why I'm currently thinking about making the procfs-mountpoint adjustable. Right now it's hard coded to `/proc`. Making this changeable would allow tests to point to some prepared fake-directory and the cognitive overhead is similar to removing a magic constant. WDYT?
> > Anyway I'm not planning on doing this in this PR because this fix should go into 1.53
> 
> SGTM. I see that making non-virtual functions mockable, makes the code less readable, so this sounds like a plan.

Great. Thank you both!

PR #1252, opened 2020-09-22
Fix Crashes on Context Menu on Loaded Captures

florian-kuebler:
Bug: http://b/168891730
Test: Load capture and try context menues on all views.

### DIFF
--- a/OrbitGl/LiveFunctionsDataView.cpp
+++ b/OrbitGl/LiveFunctionsDataView.cpp
@@ -153,14 +153,14 @@ std::vector<std::string> LiveFunctionsDataView::GetContextMenu(
     const FunctionInfo& selected_function = *GetSelectedFunction(index);
     const uint64_t absolute_address = FunctionUtils::GetAbsoluteAddress(selected_function);
 
-    // Is that function actually inside module of the process (i.e. can we disassemble?)
+    // Is that function actually inside a module of the process (i.e. can we disassemble)?
     const FunctionInfo* actual_function =
         capture_data.FindFunctionByAddress(absolute_address, false);
     const bool function_exists = actual_function != nullptr;
 
     const FunctionStats& stats = capture_data.GetFunctionStatsOrDefault(selected_function);
-    enable_select |= !GOrbitApp->IsFunctionSelected(selected_function) && function_exists;
-    enable_unselect |= GOrbitApp->IsFunctionSelected(selected_function) && function_exists;
+    enable_select |= function_exists && !GOrbitApp->IsFunctionSelected(selected_function);
+    enable_unselect |= function_exists && GOrbitApp->IsFunctionSelected(selected_function);
     enable_iterator |= stats.count() > 0;
     enable_disassembly |= function_exists;
   }
@@ -196,16 +196,17 @@ void LiveFunctionsDataView::OnContextMenu(const std::string& action, int menu_in
     for (int i : item_indices) {
       FunctionInfo* selected_function = GetSelectedFunction(i);
       const uint64_t absolute_address = FunctionUtils::GetAbsoluteAddress(*selected_function);
-      // Is that function actually inside module of the process?
-      if (capture_data.FindFunctionByAddress(absolute_address, false) != nullptr) {
-        if (action == kMenuActionSelect) {
-          GOrbitApp->SelectFunction(*selected_function);
-        } else if (action == kMenuActionUnselect) {
-          GOrbitApp->DeselectFunction(*selected_function);
-        } else if (action == kMenuActionDisassembly) {
-          int32_t pid = capture_data.process_id();
-          GOrbitApp->Disassemble(pid, *selected_function);
-        }
+      // Is that function actually inside a module of the process?
+      if (capture_data.FindFunctionByAddress(absolute_address, false) == nullptr) {
+        continue;
+      }
+      if (action == kMenuActionSelect) {
+        GOrbitApp->SelectFunction(*selected_function);
+      } else if (action == kMenuActionUnselect) {
+        GOrbitApp->DeselectFunction(*selected_function);
+      } else if (action == kMenuActionDisassembly) {
+        int32_t pid = capture_data.process_id();
+        GOrbitApp->Disassemble(pid, *selected_function);
       }
     }
   } else if (action == kMenuActionJumpToFirst) {
--- a/OrbitGl/LiveFunctionsDataView.cpp
+++ b/OrbitGl/LiveFunctionsDataView.cpp
@@ -147,14 +147,17 @@ std::vector<std::string> LiveFunctionsDataView::GetContextMenu(
   bool enable_select = false;
   bool enable_unselect = false;
   bool enable_iterator = false;
-  bool enable_disassembly = !selected_indices.empty();
+  bool enable_disassembly = false;
   const CaptureData& capture_data = GOrbitApp->GetCaptureData();
   for (int index : selected_indices) {
     const FunctionInfo& function = *GetFunction(index);
+    const uint64_t absolute_address = FunctionUtils::GetAbsoluteAddress(function);
+    const bool module_exists = capture_data.FindModuleByAddress(absolute_address) != nullptr;
     const FunctionStats& stats = capture_data.GetFunctionStatsOrDefault(function);
-    enable_select |= !GOrbitApp->IsFunctionSelected(function);
-    enable_unselect |= GOrbitApp->IsFunctionSelected(function);
+    enable_select |= !GOrbitApp->IsFunctionSelected(function) && module_exists;
+    enable_unselect |= GOrbitApp->IsFunctionSelected(function) && module_exists;
     enable_iterator |= stats.count() > 0;
+    enable_disassembly |= module_exists;
   }
 
   std::vector<std::string> menu;
@@ -193,10 +196,14 @@ void LiveFunctionsDataView::OnContextMenu(const std::string& action, int menu_in
       GOrbitApp->DeselectFunction(*function);
     }
   } else if (action == kMenuActionDisassembly) {
-    int32_t pid = GOrbitApp->GetCaptureData().process_id();
+    const CaptureData& capture_data = GOrbitApp->GetCaptureData();
+    int32_t pid = capture_data.process_id();
     for (int i : item_indices) {
       const FunctionInfo& function = *GetFunction(i);
-      GOrbitApp->Disassemble(pid, function);
+      const uint64_t absolute_address = FunctionUtils::GetAbsoluteAddress(function);
+      if (capture_data.FindModuleByAddress(absolute_address) != nullptr) {
+        GOrbitApp->Disassemble(pid, function);
+      }
     }
   } else if (action == kMenuActionJumpToFirst) {
     CHECK(item_indices.size() == 1);
--- a/OrbitGl/LiveFunctionsDataView.cpp
+++ b/OrbitGl/LiveFunctionsDataView.cpp
@@ -45,7 +45,7 @@ std::string LiveFunctionsDataView::GetValue(int row, int column) {
     return "";
   }
 
-  const FunctionInfo& function = *GetFunction(row);
+  const FunctionInfo& function = *GetSelectedFunction(row);
   const FunctionStats& stats = GOrbitApp->GetCaptureData().GetFunctionStatsOrDefault(function);
 
   switch (column) {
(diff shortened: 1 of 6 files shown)


OrbitGl/SamplingReportDataView.cpp
```diff
@@ -138,14 +137,15 @@ void SamplingReportDataView::DoSort() {
   std::sort(indices_.begin(), indices_.end(), combined_sorter);
 }
 
-std::vector<const FunctionInfo*> SamplingReportDataView::GetFunctionsFromIndices(
+absl::flat_hash_set<const FunctionInfo*> SamplingReportDataView::GetFunctionsFromIndices(
```
dpallotti, line 140: Is there a reason for changing this, other than simplifying the code? From what I remember, the rationale for having a vector was returning the functions in the same order as the `indices`.

OrbitGl/SamplingReportDataView.cpp
```diff
@@ -155,19 +155,20 @@ std::vector<const FunctionInfo*> SamplingReportDataView::GetFunctionsFromIndices
     }
   }
 
-  return std::vector<const FunctionInfo*>(functions_set.begin(), functions_set.end());
+  return functions_set;
 }
 
-std::vector<ModuleData*> SamplingReportDataView::GetModulesFromIndices(
+absl::flat_hash_set<ModuleData*> SamplingReportDataView::GetModulesFromIndices(
```
dpallotti, line 161: Same here.

OrbitGl/LiveFunctionsDataView.cpp
```diff
@@ -193,10 +196,14 @@ void LiveFunctionsDataView::OnContextMenu(const std::string& action, int menu_in
...
   } else if (action == kMenuActionDisassembly) {
-    int32_t pid = GOrbitApp->GetCaptureData().process_id();
+    const CaptureData& capture_data = GOrbitApp->GetCaptureData();
+    int32_t pid = capture_data.process_id();
     for (int i : item_indices) {
       const FunctionInfo& function = *GetFunction(i);
-      GOrbitApp->Disassemble(pid, function);
+      const uint64_t absolute_address = FunctionUtils::GetAbsoluteAddress(function);
+      if (capture_data.FindModuleByAddress(absolute_address) != nullptr) {
+        GOrbitApp->Disassemble(pid, function);
```
dpallotti, line 205: What I don't understand here is the relationship between the condition and the action&mdash;while conceptually clear, it is actually very loose. If I understand correctly `Disassemble` doesn't rely on anything related to modules as stored in the client, rather it just tries to contact the remote process and queries its memories. This can be seen in the fact that `FindModuleByAddress` is only based on `absolute_address`, while `Disassemble` uses pid&mdash;what happens if for a coincidence `absolute_address` makes sense but `pid` is not available? Would that cause a crash?
So this actually brings me to another question: Can you explain where the crash was? How does this condition relate to the crash?

OrbitGl/SamplingReportDataView.cpp
florian-kuebler: Yes, otherwise, we would try to load multiple modules at once.

florian-kuebler: Also it was a set for the functions internally already.

OrbitGl/SamplingReportDataView.cpp
florian-kuebler: It was already a std::set internally, which was then converted into a vector. That was unnecessarily complicated.

OrbitGl/SamplingReportDataView.cpp
dpallotti: Ah I see that the implementation of this didn't use a `set` first internally, unlike `GetFunctionsFromIndices`.

OrbitGl/SamplingReportDataView.cpp
dpallotti: You're right, forget about this.

OrbitGl/LiveFunctionsDataView.cpp
florian-kuebler: So okay, let me start by what was the crash here (in fact there were also other crashes).
`pid` was used in order to check if the process is 32 or 64 bit. That process, did not exists, so there was a null deference. 
`CaptureData::FindModuleByAddress` uses also the `process` to find the module. This is the exact same process, where we get the `pid` from. So they can't be off sync.

Also note, that since the recent refactoring the `ModuleData*` will be also present if the module is not loaded. So it might be only null, if we are in a loaded capture (or if we would implement unloading of modules in the future).

However, I see that using the module here to check that we can access the memory of a function is a bit confusing.
In the other views (and also already in the topdownview), the `CaptureData::FindFunctionByAddress` is used, which returns a `FunctionInfo *`. However, as we already have a `FunctionInfo *` here (as we actually reconstruct function infos for the instrumented functions that are shown here), I would find it more confusing to use an additional `FunctionInfo* `

antonrohr: Orbit crashed when the `GOrbitApp->Disassemble` is called without being connected to a remote. Which somewhat makes sense, since the memory cannot be retrieved from remote. 

capture_data does apparently not contain module information when the capture was loaded from a file, hence this actually works. 

I would prefer something else than this "if module available". Preferably some kind of CHECK somewhere when disassembly is attempted without being connected (i dont think this code location is actually suited). A CHECK should be good idea, because we already make sure that disassembly is possible when activating the context menu. 

If you want to go with how it currently is, please at least LOG or ERROR something, and maybe a comment whats happening is a good idea

antonrohr:
Maybe we should consider adding a flag to CaptureData that contains the information whether a capture was loaded from file, or is a "live" capture. And then use this information to decide whether certain actions are possible (like hooking functions or displaying disassembly and so on)

Or maybe have a flag somewhere in OrbitApp that indicates whether Orbit is connected to a remote or not.

OrbitGl/LiveFunctionsDataView.cpp
florian-kuebler: Well the actual crash was already before the attempt to load the memory, when trying to get the bittness of the process.
I added a CHECK there. In case the process exists but the remote process fails, there is already an error message appearing to the user.

florian-kuebler:
> Maybe we should consider adding a flag to CaptureData that contains the information whether a capture was loaded from file, or is a "live" capture. And then use this information to decide whether certain actions are possible (like hooking functions or displaying disassembly and so on)
> 
> Or maybe have a flag somewhere in OrbitApp that indicates whether Orbit is connected to a remote or not.

I am not 100% sure. We had these kind of flags in the past. And that made code base really hard to understand (but probably different reasons, so maybe nevermind).
For `CaptureData`, at least I would argue that adding this would be a step backwards from trying to unify loading and live capturing. First question would be, were to set this?

My intention with these conditions was to make it close to, "the data that is required to do this operation  is not there". But, I am open to getting convinced.

florian-kuebler:
> > Maybe we should consider adding a flag to CaptureData that contains the information whether a capture was loaded from file, or is a "live" capture. And then use this information to decide whether certain actions are possible (like hooking functions or displaying disassembly and so on)
> > Or maybe have a flag somewhere in OrbitApp that indicates whether Orbit is connected to a remote or not.
> 
> I am not 100% sure. We had these kind of flags in the past. And that made code base really hard to understand (but probably different reasons, so maybe nevermind).
> For `CaptureData`, at least I would argue that adding this would be a step backwards from trying to unify loading and live capturing. First question would be, were to set this?
> 
> My intention with these conditions was to make it close to, "the data that is required to do this operation is not there". But, I am open to getting convinced.

Also having the decision based on whether Orbit is connected or not, does not solve the problem here. You can load a capture from a connected state.

OrbitGl/LiveFunctionsDataView.cpp
dpallotti: So, for the crash in `OrbitApp::Disassemble` caused by `data_manager_->GetProcessByPid(pid)` being `nullptr`, I think that this case should be handled gracefully directly there. I think that other crash situations like not being connected to a remote should also first of all be handled gracefully.
But obviously in general we should not provide the option when the action is not possible (and not just give an error afterward). Maybe we could have some kind of `OrbitApp::CanDisassemble` that checks the connection first?
What do you think?

dpallotti:
> > Maybe we should consider adding a flag to CaptureData that contains the information whether a capture was loaded from file, or is a "live" capture. And then use this information to decide whether certain actions are possible (like hooking functions or displaying disassembly and so on)
> > Or maybe have a flag somewhere in OrbitApp that indicates whether Orbit is connected to a remote or not.
> 
> I am not 100% sure. We had these kind of flags in the past. And that made code base really hard to understand (but probably different reasons, so maybe nevermind).
> For `CaptureData`, at least I would argue that adding this would be a step backwards from trying to unify loading and live capturing. First question would be, were to set this?
> 
> My intention with these conditions was to make it close to, "the data that is required to do this operation is not there". But, I am open to getting convinced.

I agree with Florian here, we shouldn't mix the `CaptureData` and the state of Orbit, a fresh capture or a loaded capture should be the same. Certain actions that are seemingly on the capture but are actually referring to the live part of Orbit (such as the disassembly, the capture-related part is the pid/process name/module/function address, but the action is only on the live part, based on the capture-related data), should only depend on the state of the live part.

OrbitGl/LiveFunctionsDataView.cpp
florian-kuebler: So instead of that CHECK in Disassemble, you would have an `if(process==nullptr) SendErrorToUI`? And then have a `CanDisassemble` Method in App?
In general I like the idea. But I am quite unsure of how this `CanDisassemble` method should look like.

florian-kuebler:
As discussed offline, I used `FindFunctionByAddress` now in the live view, which at least makes this consistent. 
To avoid ambiguity, I renamed the other "funtions" in that data view, to selected functions, as this is what they actually are.

antonrohr (approved):
lgtm, thanks for fixing this!
I tested a bunch of things and all looks good to me

OrbitGl/LiveFunctionsDataView.cpp
```diff
@@ -182,58 +190,63 @@ std::vector<std::string> LiveFunctionsDataView::GetContextMenu(
...
+  const CaptureData& capture_data = GOrbitApp->GetCaptureData();
+  if (action == kMenuActionSelect || action == kMenuActionUnselect ||
+      action == kMenuActionDisassembly) {
     for (int i : item_indices) {
-      const FunctionInfo& function = *GetFunction(i);
-      GOrbitApp->Disassemble(pid, function);
+      FunctionInfo* selected_function = GetSelectedFunction(i);
+      const uint64_t absolute_address = FunctionUtils::GetAbsoluteAddress(*selected_function);
+      // Is that function actually inside module of the process?
+      if (capture_data.FindFunctionByAddress(absolute_address, false) != nullptr) {
```
dpallotti, line 200: Use an early continue here `if( capture_data.FindFunctionByAddress(absolute_address, false) == nullptr)`.

OrbitGl/LiveFunctionsDataView.cpp
```diff
@@ -147,14 +147,22 @@ std::vector<std::string> LiveFunctionsDataView::GetContextMenu(
...
   const CaptureData& capture_data = GOrbitApp->GetCaptureData();
   for (int index : selected_indices) {
-    const FunctionInfo& function = *GetFunction(index);
-    const FunctionStats& stats = capture_data.GetFunctionStatsOrDefault(function);
-    enable_select |= !GOrbitApp->IsFunctionSelected(function);
-    enable_unselect |= GOrbitApp->IsFunctionSelected(function);
+    const FunctionInfo& selected_function = *GetSelectedFunction(index);
+    const uint64_t absolute_address = FunctionUtils::GetAbsoluteAddress(selected_function);
+
+    // Is that function actually inside module of the process (i.e. can we disassemble?)
```
dpallotti, line 156:
```suggestion
    // Is that function actually inside a module of the process (i.e. can we disassemble)?
```

OrbitGl/LiveFunctionsDataView.cpp
```diff
@@ -182,58 +190,63 @@ std::vector<std::string> LiveFunctionsDataView::GetContextMenu(
...
-    int32_t pid = GOrbitApp->GetCaptureData().process_id();
+  const CaptureData& capture_data = GOrbitApp->GetCaptureData();
+  if (action == kMenuActionSelect || action == kMenuActionUnselect ||
+      action == kMenuActionDisassembly) {
     for (int i : item_indices) {
-      const FunctionInfo& function = *GetFunction(i);
-      GOrbitApp->Disassemble(pid, function);
+      FunctionInfo* selected_function = GetSelectedFunction(i);
+      const uint64_t absolute_address = FunctionUtils::GetAbsoluteAddress(*selected_function);
+      // Is that function actually inside module of the process?
```
dpallotti, line 199:
```suggestion
      // Is that function actually inside a module of the process?
```

OrbitGl/LiveFunctionsDataView.cpp
```diff
@@ -147,14 +147,22 @@ std::vector<std::string> LiveFunctionsDataView::GetContextMenu(
...
+    const uint64_t absolute_address = FunctionUtils::GetAbsoluteAddress(selected_function);
+
+    // Is that function actually inside module of the process (i.e. can we disassemble?)
+    const FunctionInfo* actual_function =
+        capture_data.FindFunctionByAddress(absolute_address, false);
+    const bool function_exists = actual_function != nullptr;
+
+    const FunctionStats& stats = capture_data.GetFunctionStatsOrDefault(selected_function);
+    enable_select |= !GOrbitApp->IsFunctionSelected(selected_function) && function_exists;
+    enable_unselect |= GOrbitApp->IsFunctionSelected(selected_function) && function_exists;
```
dpallotti, line 163: Only for conceptual reasons, I would put `function_esists` on the left side of the `&&`.

dpallotti (approved):

OrbitGl/LiveFunctionsDataView.cpp
florian-kuebler: Done. Thanks

OrbitGl/LiveFunctionsDataView.cpp
florian-kuebler: Sure.

OrbitGl/LiveFunctionsDataView.cpp
florian-kuebler: Done.

OrbitGl/LiveFunctionsDataView.cpp
florian-kuebler: Good point. Thanks!

florian-kuebler (commented):
Thanks for the comments. I addressed them.

dpallotti (approved):
Looks good, thanks a lot!

PR #1253, opened 2020-09-23
Make manual instrumentation event "id" 64 bits instead of 32

pierricgimmig:
A 64 bits member makes more sense for an async event id as a user can
now use a pointer as unique id. "value" (uint64_t) and "id" (uint32_t)
were merged into a single uint64_t, "data", leaving room to increase
kMaxEventStringSize from 30 to 34 characters.


ronaldfw (approved):
LGTM

PR #1255, opened 2020-09-23
Add QWebChannelExtractor

beckerhe:
(This is not targeting 1.53)

The JavaScript client library for Qt's WebChannel module is embedded
into the QtWebChannel shared library. For the development workflow we
need to extract that file. The simplest portable way to do that is to
compile an exectuable link it to QtWebChannel and output the resource.

On Windows where we use Qt from a conan package we could also copy the
library from the source archive, but this won't work on Linux where
we usually rely on system-installed Qt. That's why this extraction
mechanism is the simplest portable way of getting the right version of
that file.

This mechanism has a limitation: It won't work with cross-compilation
setups. But this can be easily added in the future by separating the
extractor into its own conan package and require that package as a build
requirement which means it will be compiled for the build platform and
not for the target platform. Currently we don't need the cross
compilation workflow for the UI.

Bug: http://b/169113529

The next step will be:
- Create a simple webpack config on top of this
- Having webpack run as part of the cmake build
- Automatically generate a qrc-file for all the files that webpack output
- Embed them into OrbitQt.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -165,6 +165,7 @@ if(WITH_GUI)
   add_subdirectory(OrbitSsh)
   add_subdirectory(OrbitSshQt)
   add_subdirectory(OrbitQt)
+  add_subdirectory(WebUI)
 endif()
 
 if(WIN32)
--- a/WebUI/CMakeLists.txt
+++ b/WebUI/CMakeLists.txt
@@ -1,0 +1,31 @@
+# Copyright (c) 2020 The Orbit Authors. All rights reserved.
+# Use of this source code is governed by a BSD-style license that can be
+# found in the LICENSE file.
+
+project(webUI CXX)
+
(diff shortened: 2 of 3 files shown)


CMakeLists.txt
```diff
@@ -165,6 +165,7 @@ if(WITH_GUI)
   add_subdirectory(OrbitSsh)
   add_subdirectory(OrbitSshQt)
   add_subdirectory(OrbitQt)
+  add_subdirectory(webUI)
```
florian-kuebler, line 168: Is there a reason to not start with a capital letter?

florian-kuebler (approved):
LGTM. Thanks!

CMakeLists.txt
beckerhe: Good point. I didn't even think about. I will change it. Thanks!

florian-kuebler: Thanks.

PR #1256, opened 2020-09-23
Add missing lock to TimeGraph::SortTracks

beckerhe:
`TimeGraph::thread_tracks_` and `TimeGraph::gpu_tracks_` are accessed
from different threads and usually guarded by a mutex. In
`TimeGraph::SortTracks` a lock of this mutex is missing and might have
lead to the crash in the mentioned bug.

Test: Manual test of Orbit UI - Basic tests, Capturing works, Manual instrumentation works
Bug: http://b/169124700


IrinaShkviro (approved):
Thanks!

PR #1260, opened 2020-09-24
Disable Help Text on Capture Loading

florian-kuebler:
Bug: http://b/158448668
Test: Load a capture

### DIFF
--- a/OrbitGl/CaptureWindow.h
+++ b/OrbitGl/CaptureWindow.h
@@ -52,18 +52,18 @@ class CaptureWindow : public GlCanvas {
   void OnVerticalDrag(float a_Ratio);
   void NeedsUpdate();
   void OnCaptureStarted();
-  float GetTopBarTextY();
   std::vector<std::string> GetContextMenu() override;
   void OnContextMenu(const std::string& a_Action, int a_MenuIndex) override;
   void UpdateVerticalSlider();
   void ToggleDrawHelp();
+  void set_draw_help(bool draw_help);
 
   Batcher& GetBatcherById(BatcherId batcher_id);
 
  private:
   TimeGraph time_graph_;
   OutputWindow m_StatsWindow;
-  bool m_DrawHelp;
+  bool draw_help_;
   bool m_DrawFilter;
   bool m_FirstHelpDraw;
   bool m_DrawStats;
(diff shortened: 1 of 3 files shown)


OrbitGl/CaptureWindow.h
```diff
@@ -52,18 +52,18 @@ class CaptureWindow : public GlCanvas {
   void OnVerticalDrag(float a_Ratio);
   void NeedsUpdate();
   void OnCaptureStarted();
-  float GetTopBarTextY();
   std::vector<std::string> GetContextMenu() override;
   void OnContextMenu(const std::string& a_Action, int a_MenuIndex) override;
   void UpdateVerticalSlider();
   void ToggleDrawHelp();
+  void set_draw_help(bool draw_help);
```
reichlfl, line 59: For sake of completeness, I'd suggest to either add `get_draw_help` as well, or make this method private. It's kind of weird having a setter available for a value you can't read.

reichlfl (approved):
Thanks for the fix!

OrbitGl/CaptureWindow.h
florian-kuebler: Well it can't be private, as App needs to call it. And I would avoid adding unused methods.
It is just a method that modifies internal state. I am fine with renaming it, but still it is essentially a getter.

reichlfl: Right... well, I'm ok with leaving it like that, that's why I approved nevertheless :-) I just felt it's a bit weird, the public interface for a class shouldn't imho only be defined by "this is what one other class needs at the moment", but should take consistency in itself into account, and when I see a setter without a getter, I suspect there is some reasoning behind it, e.g. it's something like `set_password`.
But I guess this is nitpicking and personal preference, and the stylguide doesn't say anything about it ;-)

florian-kuebler: I see your point. I think this somehow breaks down to your definition of a public API, i.e. wether it is expected to be a moving target or not. 
And I think beside the obvious set_password case, there is also a "there is no need for that" case. And there already is some kind of setter (with the `ToggleHelp`) function. 
Anyways, interesting discussion, but I'll merge the code as it is.
Thanks a lot for you review, though.

PR #1261, opened 2020-09-24
Run OrbitQt tests always with offscreen rendering

beckerhe:
The OrbitQt tests instantiate widgets and need a drawing context for
that. Usually that's the screen of the calling instance. In a CI context
there is no screen and we use offscreen rendering to run the tests.

This change enables the same for the developer workflow. Offscreen
rendering gets the default when available. And it's currently only
available on Linux.

This has the advantage that the OrbitQt tests don't fail anymore when
manually run from a headless context like a SSH session. Before a SSH
session with X11 forwarding was needed for a successful test. But this
also increased test times by roughly a factor of 50 to 100.

@reichlfl @dimitry- You both have written OrbitQt tests that do rendering. Please let me know if you think this might break something.


reichlfl (approved):
LGTM, thanks!

dimitry- (approved):
Thanks!

PR #1263, opened 2020-09-24
Have cmake always create compile_commands.json

beckerhe:
This enables the generation of a compile_commands.json file in CMake
whenever the used generator allows it (No msbuild support though).

Always having a compile_commands.json is convenient for developers using
clangd and is a requirement for the integration of include_what_you_use.


dimitry- (approved):

antonrohr (approved):
I like this a lot! Thank you

PR #1264, opened 2020-09-24
Disable, don't hide, CallTreeWidget's context actions

dpallotti:
These are the top-down and bottom-up views, which already have full control of
the context menus.

Bug: http://b/169222849

--- 

Untitled

What do you think?

### DIFF
--- a/OrbitQt/CallTreeWidget.cpp
+++ b/OrbitQt/CallTreeWidget.cpp
@@ -269,33 +269,19 @@ void CallTreeWidget::onCustomContextMenuRequested(const QPoint& point) {
   bool enable_copy = ui_->callTreeTreeView->selectionModel()->hasSelection();
 
   QMenu menu{ui_->callTreeTreeView};
-  if (enable_expand_recursively) {
-    menu.addAction(kActionExpandRecursively);
-  }
-  if (enable_collapse_recursively) {
-    menu.addAction(kActionCollapseRecursively);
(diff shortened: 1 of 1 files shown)


OrbitQt/CallTreeWidget.cpp
```diff
@@ -269,33 +269,38 @@ void CallTreeWidget::onCustomContextMenuRequested(const QPoint& point) {
...
   QMenu menu{ui_->callTreeTreeView};
-  if (enable_expand_recursively) {
-    menu.addAction(kActionExpandRecursively);
-  }
-  if (enable_collapse_recursively) {
-    menu.addAction(kActionCollapseRecursively);
-    menu.addAction(kActionCollapseChildrenRecursively);
-  }
+
+  menu.addAction(kActionExpandRecursively);
```
reichlfl, line 273: Not sure if this (and everything below) would be more readable as a one-liner:
`menu.addAction(kActionExpandRecursively)->setEnabled(enable_expand_recursively);`

It's a nitpick though, I can live with the current option as well.

reichlfl (approved):
I like it a lot! Thanks for changing this!

OrbitQt/CallTreeWidget.cpp
dpallotti: Aah that's how you do it, I had completely missed that `addAction` returns the `QAction`.

PR #1265, opened 2020-09-24
Show thread names in sampling reports ("Sampling" tab)

dpallotti:
The tabs get wider but I think the gain compensates. Threads names are limited
to 15 characters anyway.

Bug: http://b/159313842

---

Capture

This might be a bit controversial as tabs take more space now, but they also become much more usable (ok, Trata doesn't really rename most of its threads...). Let me know what you think.

### DIFF
--- a/OrbitGl/SamplingReportDataView.cpp
+++ b/OrbitGl/SamplingReportDataView.cpp
@@ -265,9 +265,9 @@ void SamplingReportDataView::SetSampledFunctions(const std::vector<SampledFuncti
 void SamplingReportDataView::SetThreadID(ThreadID tid) {
   tid_ = tid;
   if (tid == SamplingProfiler::kAllThreadsFakeTid) {
-    name_ = "All";
+    name_ = absl::StrFormat("%s\n(all threads)", GOrbitApp->GetCaptureData().process_name());
   } else {
-    name_ = absl::StrFormat("%d", tid_);
+    name_ = absl::StrFormat("%s\n[%d]", GOrbitApp->GetCaptureData().GetThreadName(tid_), tid_);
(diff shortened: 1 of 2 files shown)


dimitry- (approved):
nit: "Test:" stanza is missing in the commit message.

reichlfl (approved):
I think this is a good thing, even with the increased size of the tabs. It would be nice to have the tabs ordered by thread name / id (preferrably both if you look at Trata, not sure which one should be the main dimension though).

To understand correctly, the thread names are already limited to 15 chars?

dpallotti:
> It would be nice to have the tabs ordered by thread name / id (preferrably both if you look at Trata, not sure which one should be the main dimension though).

Currently they are ordered by the number of callstack samples. This is the same ordering as in the capture view, and as the default ordering in the top-down view. So for now I would really stay consistent here and leave it as it is.

> To understand correctly, the thread names are already limited to 15 chars?

Yes, this is a Linux limitation. Process/thread names are limited to 15 characters.

reichlfl:
> > It would be nice to have the tabs ordered by thread name / id (preferrably both if you look at Trata, not sure which one should be the main dimension though).
> 
> Currently they are ordered by the number of callstack samples. This is the same ordering as in the capture view, and as the default ordering in the top-down view. So for now I would really stay consistent here and leave it as it is.
> 

I think the way of interacting with them is a bit different - the capture view is super visual, so you may be less interested in the ordering. In the tabs, you tend to look for a specific thread that you identified as problematic in the capture view, and you don't care so much about the number of samples.

At least that's how I feel about it. But we can discuss that unrelated of this PR.

PR #1266, opened 2020-09-24
[On Release] Add already loaded functions after process switch

antonrohr:
To fix the regression in 1.53. (b/169305654)

Before: When switching the process the symbols in modules were cleared. FunctionsDataView was also cleared. 
Now: symbols in modules are not cleared but updated if necessary. FunctionsDataView is filled with existing functions.

Whenever the list of modules of a process changes, it could be that a module is now loaded at a different memory address than before (module_base_address). If this is actually the case and the module already contained Functions, the field `module_base_address` of these Functions is updated to the new value. 

This also adds/changes the appropriate tests

### DIFF
--- a/OrbitClientData/ModuleData.cpp
+++ b/OrbitClientData/ModuleData.cpp
@@ -75,13 +75,11 @@ void ModuleData::AddSymbols(const orbit_grpc_protos::ModuleSymbols& module_symbo
   is_loaded_ = true;
 }
 
-void ModuleData::ClearSymbols() {
+void ModuleData::UpdateFunctionsModuleBaseAddress(uint64_t module_base_address) {
   absl::MutexLock lock(&mutex_);
-
-  CHECK(is_loaded_);
-  functions_.clear();
-  hash_to_function_map_.clear();
-  is_loaded_ = false;
+  for (auto& [_, function] : functions_) {
+    function->set_module_base_address(module_base_address);
+  }
 }
 
 const orbit_client_protos::FunctionInfo* ModuleData::FindFunctionFromHash(uint64_t hash) const {
--- a/OrbitClientData/include/OrbitClientData/ModuleData.h
+++ b/OrbitClientData/include/OrbitClientData/ModuleData.h
@@ -37,12 +37,14 @@ class ModuleData final {
       uint64_t relative_address, bool is_exact) const;
   [[nodiscard]] const orbit_client_protos::FunctionInfo* FindFunctionByElfAddress(
       uint64_t elf_address, bool is_exact) const;
-  // TODO(antonrohr): The module_base_address parameter should not be needed here, but it is
+  // TODO(169309553): The module_base_address parameter should not be needed here, but it is
   // because FunctionInfo still contains the field module_base_address. As soon as that field is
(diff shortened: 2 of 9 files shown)


OrbitClientData/ModuleData.cpp
```diff
@@ -75,13 +75,11 @@ void ModuleData::AddSymbols(const orbit_grpc_protos::ModuleSymbols& module_symbo
...
 
-void ModuleData::ClearSymbols() {
+void ModuleData::UpdateFunctionsModuleBaseAddress(uint64_t module_base_address) {
   absl::MutexLock lock(&mutex_);
-
-  CHECK(is_loaded_);
-  functions_.clear();
-  hash_to_function_map_.clear();
-  is_loaded_ = false;
+  for (const auto& [_, function] : functions_) {
```
dimitry-, line 80: the line below does not look like it is const....

OrbitGl/DataManager.cpp
```diff
@@ -37,18 +37,30 @@ void DataManager::UpdateModuleInfos(int32_t process_id,
...
       const auto [inserted_it, success] = module_map_.try_emplace(
           module_info.file_path(), std::make_unique<ModuleData>(module_info));
       CHECK(success);
+    } else {
+      ModuleData* module = (*module_it).second.get();
+      if (!module->is_loaded()) continue;
+      // When a module is already loaded (has loaded symbols), it could be the case that the
+      // modules base address changed. Since the module base address is *currently* saved in
+      // FunctionInfo, these need to be updated.
+      // As soon as module base address is not part of FunctionInfo anymore, remove the following.
```
dimitry-, line 56: nit: link to a bug?

dimitry- (approved):
otherwise lgtm

OrbitClientData/include/OrbitClientData/ModuleData.h
```diff
@@ -42,7 +42,9 @@ class ModuleData final {
   // gone, remove the parameter here
   void AddSymbols(const orbit_grpc_protos::ModuleSymbols& module_symbols,
                   uint64_t module_base_address);
-  void ClearSymbols();
+  // Same as last comment: As soon as FunctionInfo does not contain module_base_address anymore,
+  // completely remove the following method
```
IrinaShkviro, line 46: Maybe add `TODO(antonrohr)` in the beginning of the line as well?

IrinaShkviro (approved):
Thanks Anton!

OrbitClientData/include/OrbitClientData/ModuleData.h
IrinaShkviro: Probably, makes no sense as it's obvious. Feel free to ignore.

antonrohr: I added a bug and used the bug number

OrbitClientData/ModuleData.cpp

OrbitGl/DataManager.cpp
antonrohr: done

florian-kuebler (approved):
Thanks for that fix!. LGTM.

PR #1267, opened 2020-09-24
Don't set system_qt option in build script

beckerhe:
The build.ps1 script used to set the `system_qt` setting when calling
`conan install`. That's wrong because this settings should be set by
the conan profiles.

This change removes the settings which solves a problem when building
for Stadia on Windows.

@pierricgimmig : FYI, that's the fix to the problem you had as well.

### DIFF
--- a/build.ps1
+++ b/build.ps1
@@ -47,7 +47,7 @@ foreach ($profile in $profiles) {
   
   Write-Host "Building Orbit in build_$profile/ with conan profile $profile"
 
-  $process = Start-Process $conan.Path -Wait -NoNewWindow -ErrorAction Stop -PassThru -ArgumentList "install","--update","-pr",$profile,"-o","system_qt=False","-if","build_$profile/","--build","outdated",$PSScriptRoot
+  $process = Start-Process $conan.Path -Wait -NoNewWindow -ErrorAction Stop -PassThru -ArgumentList "install","--update","-pr",$profile,"-if","build_$profile/","--build","outdated",$PSScriptRoot
   if ($process.ExitCode -ne 0) {
     Throw "Error while running conan install."
   }


build.ps1
```diff
@@ -47,7 +47,7 @@ foreach ($profile in $profiles) {
   
   Write-Host "Building Orbit in build_$profile/ with conan profile $profile"
 
-  $process = Start-Process $conan.Path -Wait -NoNewWindow -ErrorAction Stop -PassThru -ArgumentList "install","--update","-pr",$profile,"-o","system_qt=False","-if","build_$profile/","--build","outdated",$PSScriptRoot
+  $process = Start-Process $conan.Path -Wait -NoNewWindow -ErrorAction Stop -PassThru -ArgumentList "install","--update","-pr",$profile,"-if","build_$profile/","--build","outdated",$PSScriptRoot
```
IrinaShkviro, line 50: We passed the `system_qt=False` here, when in conanfile.py default value is `True`. Should we change it as well?

beckerhe: Not quite, the values are properly set in the profiles. system_qt being False even when it should be the default (True) lead to the mentioned problem.

Example:

```
$ conan profile show msvc2017_relwithdebinfo
Configuration for profile msvc2017_relwithdebinfo:

[settings]
os=Windows
os_build=Windows
arch=x86_64
arch_build=x86_64
compiler=Visual Studio
compiler.version=15
build_type=RelWithDebInfo
[options]
OrbitProfiler:system_qt=False
[build_requires]
*: cmake_installer/3.16.3@conan/stable
[env]
CONAN_CMAKE_SYSTEM_VERSION=8.1
LDFLAGS=/FIXED:NO /ignore:4099
```

IrinaShkviro: Thanks for the explanation!

IrinaShkviro (approved):
Lgtm, thank you!

beckerhe:
> Lgtm, thank you!

Thanks a lot for the quick review!

PR #1268, opened 2020-09-24
[On Release] Reenable Context Menus for Loaded Captures if Still Connected

florian-kuebler:
Recent changes (#1252 and #1208) disabled some context menus on
loaded captures based on function/module pointers taken from
capture data.
However, if we just save a capture and load it again (without
switching to another process), these actions are possible with
1.52.

Bug: http://b/169305792
Test: 1. Take capture, load it directly -> context menus are there
      2. Take capture, switch process, load capture -> no context menu
      3. Load a capture -> no context menus


florian-kuebler:
As discussed offline. We are aiming for a proper fix on a conceptual level.

PR #1270, opened 2020-09-24
Introduction of a conan base lockfiles

beckerhe:
Base lockfiles is a new feature in conan which allows to only lock recipe version and revisions.
Since no options and settings are locked it's enough to have a single lock-file for all build configurations
and it also allows to use the lock file for developer builds (not only on the CI like it was the case for the full
lock files).

Handling base lockfile is a lot simpler due to less information in the file. Code reviews will be much simpler since
only a single lock file needs to be reviewed.

Since it requires a new version of conan, this PR also changes the conan-version that is installed by the bootstrap-scripts.


IrinaShkviro (approved):
It's very cool, thank you!

beckerhe:
> It's very cool, thank you!

Thanks. I'm going to merge tomorrow after the release. I can't say for sure that updating conan will work with the old solution. And folks should be able to checkout the 1.53-branch without the need to downgrad conan.

bootstrap-orbit.ps1
```diff
@@ -35,19 +35,19 @@ You can call 'pip3 show -f conan' to figure out where conan.exe was placed.
...
+  $conan_version_minor_min = 29
 
   if ($conan_version_major -eq $conan_version_major_required -and $conan_version_minor -lt $conan_version_minor_min) {
     Write-Host "Your conan version $conan_version is too old. Let's try to update it."
 
     Try {
       $pip3 = Get-Command pip3
-      Start-Process -FilePath $pip3.Path -ArgumentList "install","--upgrade","conan" -Wait -NoNewWindow -PassThru
+      & $pip3.Path install --upgrade conan==1.29.2
     } Catch {
```
antonrohr, line 51: I am no powershell expert, so forgive me if I am missing something. This here seems at least inconsistent with the calls to pip3 in line 22. 
Does pip3 install throw an error on fail that will be caught here, or do you have to check $LastExitCode as in line 22. Aka why don't you use TryCatch in line 22?

bootstrap-orbit.ps1
```diff
@@ -65,7 +65,7 @@ You can call 'pip3 show -f conan' to figure out where conan.exe was placed.
 # Start build
 if ($args) {
   & "$PSScriptRoot\build.ps1" $args
-} elseif (conan remote list | Select-String -NotMatch "Disabled:" | Select-String "artifactory:") {
+} elseif (& $conan.Path remote list | Select-String -NotMatch "Disabled:" | Select-String "artifactory:") {
   & "$PSScriptRoot\build.ps1" "default_relwithdebinfo" "ggp_relwithdebinfo"
 } else {
   & "$PSScriptRoot\build.ps1"
```
antonrohr, line 71: nit: It seems like build.sh/ps1 also does some check whether arguments are provided or not. Maybe these lines can go into build.sh/ps1. I think that would simplify this a tiny little bit

antonrohr (approved):
I only have 2 minor comments. Otherwise lgtm.

I haven't tested it yet. But I am going to now. I will report back.

bootstrap-orbit.ps1
beckerhe: There are two things happening here. The Try-Catch is NOT catching an install error. It's catching when the Get-Command does not find pip3 and `$pip3` will be null, so `$pip3.Path` throws.

The return code is ignored here on purpose for compatibility reasons.

In Line 22 it different because `$pip3` can't be null. This is checked before.

bootstrap-orbit.ps1
beckerhe: Unfortunately this won't work. When the build script is called without arguments it will only build the default profile (`default_relwithdebinfo`). And when introducing this logic I didn't want to change that. Everybody is used to that behaviour.

So I have to decide here whether I wanna build the profiles provided, or default and ggp if possible, or fallback to the build-script default which means only the default profile.

Does that make sense?

antonrohr:
ran bootstrap-orbit.ps1/sh on both linux and windows and updating conan and the build worked great. 

I also ran gen_lockfile.sh with the idea that the base.lock file should not change, but it did. I assume this is okay, but I wanted to let you know anyways

beckerhe:
> ran bootstrap-orbit.ps1/sh on both linux and windows and updating conan and the build worked great.

Awesome, Thanks for testing.

> 
> I also ran gen_lockfile.sh with the idea that the base.lock file should not change, but it did. I assume this is okay, but I wanted to let you know anyways

Might be okay or might not. What was the change. Do you have a diff?

bootstrap-orbit.ps1
antonrohr: > The return code is ignored here on purpose for compatibility reasons.
Sorry that I have to ask again, I don't understand this.. what compatibility?

If pip3 is available, but the install --upgrade fails, this will not be detected (I just checked it)

beckerhe: Yes. Exactly. That is what I was saying. There was a bug about that. If I remember correctly this was coming from Pierric. pip3 was always failing due to some permission problem. But the conan installation was fine, somehow. And since the only point here is to provide a better error message to the user we decided to leave it like so.

antonrohr: Ah okay. Now it makes sense. Thank you for taking the time to explain

bootstrap-orbit.ps1
antonrohr: Yes, it does

PR #1271, opened 2020-09-24
Add a white rectangle near the variable values

karupayun:
Previously, in manual instrumentation we wrote the value of each variable on the top of their track. Now, we draw a white rectangle closer to the corresponding value.

Related to b\169310201.

Before:
Screen Shot 2020-09-24 at 14 04 16

After
Screen Shot 2020-09-24 at 14 20 53

### DIFF
--- a/OrbitGl/GraphTrack.cpp
+++ b/OrbitGl/GraphTrack.cpp
@@ -13,6 +13,7 @@ void GraphTrack::Draw(GlCanvas* canvas, PickingMode picking_mode) {
   Batcher* batcher = canvas->GetBatcher();
 
   float trackWidth = canvas->GetWorldWidth();
+  const TimeGraphLayout& layout = time_graph_->GetLayout();
 
   pos_[0] = canvas->GetWorldTopLeftX();
   SetSize(trackWidth, GetHeight());
@@ -32,15 +33,31 @@ void GraphTrack::Draw(GlCanvas* canvas, PickingMode picking_mode) {
   }
 
   float track_z = GlCanvas::kZValueTrack;
-  float text_z = GlCanvas::kZValueText;
 
   Box box(pos_, Vec2(size_[0], -size_[1]), track_z);
 
   // Draw label with graph value at current mouse position.
-  const Color kLabelTextColor(255, 255, 255, 255);
+  const Color kWhiteColor(255, 255, 255, 255);
+  const Color kBlackColor(0, 0, 0, 255);
   double graph_value = GetValueAtTime(time_graph_->GetCurrentMouseTimeNs());
-  canvas->GetTextRenderer().AddText(std::to_string(graph_value).c_str(), canvas->GetMouseX(), y0,
-                                    text_z, kLabelTextColor);
+
+  int white_text_box_width =
+      canvas->GetTextRenderer().GetStringWidth(std::to_string(graph_value).c_str());
+
+  Vec2 white_text_box_size(white_text_box_width, layout.GetTextBoxHeight());
+  auto rightmost_x_position = canvas->GetWorldTopLeftX() + canvas->GetWorldWidth() -
+                              layout.GetRightMargin() - layout.GetSliderWidth() -
+                              white_text_box_size[0];
+  Vec2 white_text_box_position(
+      std::min(canvas->GetMouseX(), rightmost_x_position),
+      y0 - static_cast<float>((max_ - graph_value) * size_[1] / value_range_) -
+          white_text_box_size[1] / 2.f);
+
+  canvas->GetTextRenderer().AddText(std::to_string(graph_value).c_str(), white_text_box_position[0],
+                                    white_text_box_position[1] + layout.GetTextOffset(),
+                                    GlCanvas::kZValueTextUi, kBlackColor, white_text_box_size[0]);
+  Box white_text_box(white_text_box_position, white_text_box_size, GlCanvas::kZValueUi);
+  batcher->AddBox(white_text_box, kWhiteColor, shared_from_this());
 
   batcher->AddBox(box, color, shared_from_this());
 
@@ -73,7 +90,8 @@ void GraphTrack::Draw(GlCanvas* canvas, PickingMode picking_mode) {
     float x1 = time_graph_->GetWorldFromTick(time);
     float y0 = base_y + static_cast<float>(last_normalized_value) * size_[1];
     float y1 = base_y + static_cast<float>(normalized_value) * size_[1];
-    time_graph_->GetBatcher().AddLine(Vec2(x0, y0), Vec2(x1, y1), text_z, kLineColor);
+    time_graph_->GetBatcher().AddLine(Vec2(x0, y0), Vec2(x1, y1), GlCanvas::kZValueText,
+                                      kLineColor);
 
     previous_time = time;
     last_normalized_value = normalized_value;
--- a/OrbitGl/TextRenderer.cpp
+++ b/OrbitGl/TextRenderer.cpp
@@ -220,8 +220,7 @@ void TextRenderer::AddText(const char* a_Text, float a_X, float a_Y, float a_Z,
 
   if (a_RightJustified) {
     a_MaxSize = FLT_MAX;
-    int stringWidth, stringHeight;
-    GetStringSize(a_Text, stringWidth, stringHeight);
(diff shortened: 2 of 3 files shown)


ronaldfw:
Looks awesome! I'm wondering if text in the iterator overlay should be drawn in the same way (currently, it can be really hard to read sometimes).

karupayun:
> Looks awesome! I'm wondering if text in the iterator overlay should be drawn in the same way (currently, it can be really hard to read sometimes).

Yes, could be. I'm attaching a screenshot of the Iterators, so people could make an opinion.

Screen Shot 2020-09-24 at 15 56 15

In particular, note that it is very difficult to read the total time, because it is on a sample track.

OrbitGl/GraphTrack.cpp
```diff
@@ -32,15 +33,26 @@ void GraphTrack::Draw(GlCanvas* canvas, PickingMode picking_mode) {
...
 
   // Draw label with graph value at current mouse position.
-  const Color kLabelTextColor(255, 255, 255, 255);
+  const Color kWhiteColor(255, 255, 255, 255);
+  const Color kBlackColor(0, 0, 0, 255);
   double graph_value = GetValueAtTime(time_graph_->GetCurrentMouseTimeNs());
-  canvas->GetTextRenderer().AddText(std::to_string(graph_value).c_str(), canvas->GetMouseX(), y0,
-                                    text_z, kLabelTextColor);
+
+  Vec2 arrow_box_size(50., layout.GetTextBoxHeight());
```
ronaldfw, line 44: What is the arrow box? I don't understand why this is called "arrow". 

And why is it okay to use a constant for the width here? If you need the constant, make it a named constant where the name reflects the usage.

OrbitGl/GraphTrack.cpp
```diff
@@ -32,15 +33,26 @@ void GraphTrack::Draw(GlCanvas* canvas, PickingMode picking_mode) {
...
+  auto rightmost_x_position = canvas->GetWorldTopLeftX() + canvas->GetWorldWidth() -
+                              layout.GetRightMargin() - layout.GetSliderWidth() - arrow_box_size[0];
+  Vec2 arrow_box_position(std::min(canvas->GetMouseX(), rightmost_x_position),
+                          y0 - static_cast<float>((max_ - graph_value) * size_[1] / value_range_) -
+                              arrow_box_size[1] / 2);
```
ronaldfw, line 49: Vec2 is float, right? Then 2 should be 2.f here.

pierricgimmig (commented):
Hi Pablo, looks really cool, thanks! I agree with Ronald's comments, otherwise, lgtm!

OrbitGl/GraphTrack.cpp
karupayun: Sorry for that. My plan is to make something like an arrow box, but will come in a future PR. I renamed it.

It's clearly not ok :)

OrbitGl/GraphTrack.cpp
karupayun: Done!

OrbitGl/TimeGraphLayout.h
```diff
@@ -13,6 +13,7 @@ class TimeGraphLayout {
   float GetTextCoresHeight() const { return m_CoresHeight * scale_; }
   float GetEventTrackHeight() const { return m_EventTrackHeight * scale_; }
   float GetGraphTrackHeight() const { return m_GraphTrackHeight * scale_; }
+  float GetGraphTrackBoxWidth() const { return graph_track_box_width_; }
```
pierricgimmig, line 16: nit. I would rename to GetGraphTrackValueLabelWidth() or something similar. I think "BoxWidth" might be too generic.

pierricgimmig (approved):
Thanks Pablo! I have a very minor comment, but apart from that, LGTM!

OrbitGl/GraphTrack.cpp
```diff
@@ -32,15 +33,28 @@ void GraphTrack::Draw(GlCanvas* canvas, PickingMode picking_mode) {
...
 
   // Draw label with graph value at current mouse position.
-  const Color kLabelTextColor(255, 255, 255, 255);
+  const Color kWhiteColor(255, 255, 255, 255);
+  const Color kBlackColor(0, 0, 0, 255);
   double graph_value = GetValueAtTime(time_graph_->GetCurrentMouseTimeNs());
-  canvas->GetTextRenderer().AddText(std::to_string(graph_value).c_str(), canvas->GetMouseX(), y0,
-                                    text_z, kLabelTextColor);
+
+  Vec2 white_text_box_size(layout.GetGraphTrackBoxWidth(), layout.GetTextBoxHeight());
```
ronaldfw, line 44: GetGraphTrackBoxWidth() is a constant, right? Isn't it possible that values are too long for the width chosen? I don't know if we have a good way of predicting text width based on a string, but it seems that's what would be needed here.

ronaldfw: Actually we may have a method to that that. Check TextRenderer::GetStringSize().

OrbitGl/TimeGraphLayout.h
karupayun: Thanks!

OrbitGl/GraphTrack.cpp
```diff
@@ -32,15 +33,32 @@ void GraphTrack::Draw(GlCanvas* canvas, PickingMode picking_mode) {
...
 
   // Draw label with graph value at current mouse position.
-  const Color kLabelTextColor(255, 255, 255, 255);
+  const Color kWhiteColor(255, 255, 255, 255);
+  const Color kBlackColor(0, 0, 0, 255);
   double graph_value = GetValueAtTime(time_graph_->GetCurrentMouseTimeNs());
-  canvas->GetTextRenderer().AddText(std::to_string(graph_value).c_str(), canvas->GetMouseX(), y0,
-                                    text_z, kLabelTextColor);
+
+  int white_text_box_width;
```
ronaldfw, line 44: Maybe initialize with default value?

OrbitGl/GraphTrack.cpp
```diff
@@ -32,15 +33,32 @@ void GraphTrack::Draw(GlCanvas* canvas, PickingMode picking_mode) {
...
+  canvas->GetTextRenderer().GetStringWidth(std::to_string(graph_value).c_str(),
+                                           white_text_box_width);
+
+  Vec2 white_text_box_size(white_text_box_width, layout.GetTextBoxHeight());
```
ronaldfw, line 48: Have you checked if the entire text is rendered when using this size? I had to use a slightly larger size as the last character wasn't rendered when I used the size returned by GetStringSize.

(Could be a mistake I made, but maybe worth double-checking.)

ronaldfw (approved):
LGTM, with two minor questions/nitpicks!

OrbitGl/GraphTrack.cpp
karupayun: With 0 is ok?

OrbitGl/GraphTrack.cpp
karupayun: You are right, I've just modified the GetString to return the ceiling of the float instead of casting directly to an int. Let me know if it works for you or you still have to add the "+1".

OrbitGl/GraphTrack.cpp
ronaldfw: See my other suggestion to just return the width from GetStringWidth, then you don't have to initialize with a default value here.

OrbitGl/GraphTrack.cpp
ronaldfw: That sounds good. I'll try if that works for me once your PR is in.

ronaldfw (approved):
LGTM!

OrbitGl/TextRenderer.cpp
```diff
@@ -315,9 +315,8 @@ int TextRenderer::AddText2D(const char* a_Text, int a_X, int a_Y, float a_Z, con
   return a_X;
 }
 
-void TextRenderer::GetStringSize(const char* a_Text, int& a_Width, int& a_Height) {
+void TextRenderer::GetStringWidth(const char* a_Text, int& a_Width) {
```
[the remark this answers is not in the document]
karupayun: Yes, obvious. Sorry for that and thank you

PR #1272, opened 2020-09-24
Fix Various Clang-Tidy Warnings in CaptureWindow

florian-kuebler:
Test: Compile

### DIFF
--- a/OrbitGl/CaptureWindow.cpp
+++ b/OrbitGl/CaptureWindow.cpp
@@ -76,8 +76,10 @@ void CaptureWindow::MouseMoved(int x, int y, bool left, bool /*right*/, bool /*m
 
     time_graph_.GetWorldMinMax(world_min, world_max);
 
-    world_top_left_x_ = world_click_x_ - static_cast<float>(x) / getWidth() * m_WorldWidth;
-    world_top_left_y_ = world_click_y_ + static_cast<float>(y) / getHeight() * m_WorldHeight;
+    auto width = static_cast<float>(getWidth());
+
+    world_top_left_x_ = world_click_x_ - static_cast<float>(x) / width * m_WorldWidth;
+    world_top_left_y_ = world_click_y_ + static_cast<float>(y) / width * m_WorldHeight;
 
     world_top_left_x_ = clamp(world_top_left_x_, world_min, world_max - m_WorldWidth);
     world_top_left_y_ =
@@ -179,8 +181,6 @@ void CaptureWindow::SelectTextBox(const TextBox* text_box) {
   GOrbitApp->set_selected_thread_id(text_box->GetTimerInfo().thread_id());
 
   const TimerInfo& timer_info = text_box->GetTimerInfo();
-  uint64_t address = timer_info.function_address();
-  FindCode(address);
 
   if (m_DoubleClicking) {
     time_graph_.Zoom(timer_info);
@@ -195,7 +195,7 @@ void CaptureWindow::Hover(int x, int y) {
   auto pick_id = absl::bit_cast<PickingId>(pixels);
   Batcher& batcher = GetBatcherById(pick_id.batcher_id);
 
-  std::string tooltip = "";
+  std::string tooltip;
 
   if (pick_id.type == PickingType::kPickable) {
     auto pickable = GetPickingManager().GetPickableFromId(pick_id);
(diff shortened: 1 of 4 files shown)


beckerhe:
There's also stuff from GlCanvas in here. Was this intended? I'm mainly asking because in GlCanvas not all variables have been renamed.

florian-kuebler:
I am actually not sure how this happened in the refactoring flow. However, as the renamings are legit, I would keep them.
The rest will be probably then my next target.

florian-kuebler:
Well, capture window actually extends GlCanvas, this is why it was also renamed.

dimitry-:
This PR is hard to review because it has renames and also clang-tidy fixes. Can you please separate renames to a separate commit?

florian-kuebler:
> This PR is hard to review because it has renames and also clang-tidy fixes. Can you please separate renames to a separate commit?

I can try my best to extract the non-naming related changes in a different commit and understand that it is hard to read. Just want to mention, that the renaming issues are also clang-tidy warnings :)

florian-kuebler:
I've split the renaming and the other changes.

beckerhe (approved):
Awesome. Thanks a lot for splitting this. It looks good to me. The only thing: I'm still in favor of not leaving GlCanvas.{cpp,h} half-renamed like it is right now.

florian-kuebler:
> Awesome. Thanks a lot for splitting this. It looks good to me. The only thing: I'm still in favour of not leaving GlCanvas.{cpp,h} half-renamed like it is right now.

As I said, the issue here is that CaptureWindow extends GlCanvas, so I need to rename both. And I wanted to keep the PR as small as possible. So I decided to go for GlCanvas afterwards.

beckerhe:
> > Awesome. Thanks a lot for splitting this. It looks good to me. The only thing: I'm still in favour of not leaving GlCanvas.{cpp,h} half-renamed like it is right now.
> 
> As I said, the issue here is that CaptureWindow extends GlCanvas, so I need to rename both. And I wanted to keep the PR as small as possible. So I decided to go for GlCanvas afterwards.

Alright. Let's do it like so.

OrbitGl/CaptureWindow.cpp
```diff
@@ -575,9 +575,9 @@ Batcher& CaptureWindow::GetBatcherById(BatcherId batcher_id) {
       return time_graph_.GetBatcher();
     case BatcherId::kUi:
       return ui_batcher_;
+    default:
+      UNREACHABLE();
```
dimitry-, line 579: Interesting.. The downside here is that you won't get a compiler time error if someone adds new enum value, instead you will get a runtime error - maybe we should suppress this one?

dimitry- (approved):

OrbitGl/CaptureWindow.cpp
beckerhe: It's actually even more complicated. I tried to fix that a while ago. BatcherId is declared with a fixed underlying type uint32_t which means according to the standard that all possible uint32_t-values are valid, not only the ones with names. And the compiler won't work. The proper solution is not to use a fixed underlying type for that enum but that's complicated because it's used in a struct which is bitcast (or something similar - I don't remember from the top of my head.) So I gave up a while ago on this...

florian-kuebler: Then maybe just remove the UNREACHABLE completely? But also fine, with keeping as it was before my change.

florian-kuebler: But then at least we would get the runtime error. Henning, do you suggest reverting my change?

beckerhe: No, I think it's the best we can do for now. Obviously a compile time check is better and a runtime-check just a last resort. So if clang-tidy suggests you to do it this way it's fine for me. I would change it though when we refactor BatcherId

florian-kuebler: It was basically telling me, that UNREACHABLE is unreachable code :). So not 100% the obvious correct solution

dimitry-:
--

clang-format determined the following necessary changes to your PR:
--- OrbitCore/TracepointEventBuffer.cpp	(before formatting)
+++ OrbitCore/TracepointEventBuffer.cpp	(after formatting)
@@ -90,5 +90,4 @@
   } else {
     return GetTracepointsOfThread(thread_id).size();
   }
-
 }


Here is the list of files that need formatting:
OrbitCore/TracepointEventBuffer.cpp
--

PR #1274, opened 2020-09-24
Move OrbitApp and ImGui scope to RunUiInstance

reichlfl:
**Targets release 1.53**

When Orbit's main window is closed, the process does not exit.

Reason: 
Destruction of `TimeGraph` accesses `GOrbitApp::manual_instrumentation_manager_`, but GOrbitApp is destroyed before TimeGraph.

Current Initialization:
OrbitMainWindow is initialized on the Stack by main.cpp::RunUiInstance()
OrbitMainWindow calls GOrbitApp::Init()

Current Deinitialization:
main.cpp::RunUiInstance() calls GOrbitApp::Deinit() after the main window is closed
OrbitMainWindow is deleted when it runs out of scope (when main.cpp::RunUiInstance() returns)

In addition, ImGui is initialized by GLCanvas (which is initialized by OrbitMainWindow), deleted by GOrbitApp, which introduced another bug after fixing this one.

Fix:
main.cpp::RunUiInstance() manages scope of GOrbitApp, ImGui and OrbitMainWindow directly, no more hidden initialization of GOrbitApp and ImGui.

Adding a bunch of reviewers since it's a hotfix for the release.
@pierricgimmig Please take a careful look if this could interfere with anything else regarding manual instrumentation.

### DIFF
--- a/OrbitQt/main.cpp
+++ b/OrbitQt/main.cpp
@@ -24,6 +24,7 @@
 #include "App.h"
 #include "ApplicationOptions.h"
 #include "Error.h"
+#include "MainThreadExecutorImpl.h"
 #include "OrbitBase/Logging.h"
 #include "OrbitGgp/Error.h"
 #include "OrbitSsh/Context.h"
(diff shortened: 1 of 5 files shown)


OrbitQt/main.cpp
```diff
@@ -122,32 +123,44 @@ static outcome::result<void> RunUiInstance(
...
-          w.close();
-          QApplication::quit();
-        })};
-  }();
-
-  if (!capture_path.isEmpty()) {
-    w.OpenCapture(capture_path.toStdString());
+    QApplication::exec();
+
+    Orbit_ImGui_Shutdown();
```
dimitry-, line 159: I wonder if you want to have a scoped object (ScopedOrbitImGui) where this is called in the d-tor. There are no early returns atm but someone could add one in the future.

dimitry- (approved):
otherwise lgtm.

OrbitQt/main.cpp
reichlfl: Yes, that would be a nice improvement. But I'd leave that for the future, I wanted to keep the changes small since it's a release hotfix. I'll track that in a separate bug.

pierricgimmig (approved):
Hi Florian, thanks a lot for this fix. I synched your branch and ran OrbitTest which contains manual instrumentation tests, everything looks fine. Cheers!

PR #1275, opened 2020-09-24
Selection in the Capture Window is no longer cleared when scrolling up

reichlfl:
The selected thread was cleared as soon as anything else was clicked
inside the capture window. This also happened when the user tried
to scroll up to see the scheduler view.

To reproduce:
1) Select a thread
2) Scroll in the capture view
-> Thread is no longer selected.

Fix:
1) Threads are not deselected if any pickable is clicked
2) Threads are not deselected if the capture window background is clicked **and** dragged

In addition, thread selection was improved:
- Threads can be selected by clicking on the tab title
- Titles of selected threads are shown in blue

3 commits:
1) Fix the selection issue
2) Cleanup the use of "-1" as fake thread id
3) Highlight currently selected thread

image

### DIFF
--- a/OrbitGl/App.cpp
+++ b/OrbitGl/App.cpp
@@ -757,7 +757,7 @@ void OrbitApp::StopCapture() {
 
 void OrbitApp::ClearCapture() {
   capture_data_ = CaptureData();
-  set_selected_thread_id(-1);
+  set_selected_thread_id(SamplingProfiler::kAllThreadsFakeTid);
   SelectTextBox(nullptr);
 
   UpdateAfterCaptureCleared();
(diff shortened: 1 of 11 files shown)


beckerhe (approved):
This looks pretty straight forward to me. No complaints.

Thanks a lot for that change!

reichlfl:
I accidentally deleted the branch from my remote... restored and re-opened, sorry for that.

beckerhe:
> I accidentally deleted the branch from my remote... restored and re-opened, sorry for that.

And I was already wondering if I said something wrong. :see_no_evil:

PR #1277, opened 2020-09-24
Remove duplicate events in TracepointEventBuffer

MariaSandru27:
Before, the events were stored in the corresponding entry 
for the thread on which they were activated and also in 
another entry corresponding to all the tracepoint events
triggered on all the threads in the target process. 

To store every tracepoint event in an entry for all the threads
was necessary for the thread track corresponding to 'all threads'.

However, this meant duplicating events so, a solution was to only
store the events in an entry corresponding to the thread on which
they were activated. When we are interested in all the tracepoints
triggered in the current process we iterate though the map that stores
the tracepoints by threads.

Also, we store in a separate map all the tracepoints that have
not been triggered in the target process. 
When the visualization for the system-wide tracepoints
will be enabled, we will just use all the entries in this map and in
the map  of the tracepoints linked to the threads.

### DIFF
--- a/OrbitCore/CaptureData.h
+++ b/OrbitCore/CaptureData.h
@@ -111,10 +111,6 @@ class CaptureData {
 
   [[nodiscard]] const CallstackData* GetCallstackData() const { return callstack_data_.get(); };
 
-  [[nodiscard]] Mutex& GetTracepointEventBufferMutex() const {
-    return tracepoint_event_buffer_->mutex();
-  }
-
   [[nodiscard]] orbit_grpc_protos::TracepointInfo GetTracepointInfo(uint64_t key) const {
     return tracepoint_info_manager_->Get(key);
   }
@@ -125,6 +121,13 @@ class CaptureData {
 
   [[nodiscard]] const TracepointEventBuffer* GetTracepointEventBuffer() const {
     return tracepoint_event_buffer_.get();
+  }
+
+  void ForEachTracepointEventPerThread(
+      int32_t thread_id,
+      const std::function<void(
+          const std::map<uint64_t, orbit_client_protos::TracepointEventInfo>&)>& action) const {
+    return tracepoint_event_buffer_->ForEachTracepointEventPerThread(thread_id, action);
   }
 
   [[nodiscard]] const std::map<uint64_t, orbit_client_protos::TracepointEventInfo>&
--- a/OrbitCore/CaptureData.h
+++ b/OrbitCore/CaptureData.h
@@ -123,11 +123,11 @@ class CaptureData {
     return tracepoint_event_buffer_.get();
   }
 
-  void ForEachTracepointEventPerThread(
-      int32_t thread_id,
-      const std::function<void(
-          const std::map<uint64_t, orbit_client_protos::TracepointEventInfo>&)>& action) const {
-    return tracepoint_event_buffer_->ForEachTracepointEventPerThread(thread_id, action);
+  void ForEachTracepointEventOfThread(
+      int32_t thread_id, uint64_t min_tick, uint64_t max_tick,
+      const std::function<void(const orbit_client_protos::TracepointEventInfo&)>& action) const {
+    return tracepoint_event_buffer_->ForEachTracepointEventOfThread(thread_id, min_tick, max_tick,
+                                                                    action);
   }
 
   [[nodiscard]] const std::map<uint64_t, orbit_client_protos::TracepointEventInfo>&
--- a/OrbitCore/CaptureData.h
+++ b/OrbitCore/CaptureData.h
@@ -123,11 +123,11 @@ class CaptureData {
     return tracepoint_event_buffer_.get();
   }
 
-  void ForEachTracepointEventOfThread(
+  void ForEachTracepointEventOfThreadInTimeRange(
       int32_t thread_id, uint64_t min_tick, uint64_t max_tick,
       const std::function<void(const orbit_client_protos::TracepointEventInfo&)>& action) const {
-    return tracepoint_event_buffer_->ForEachTracepointEventOfThread(thread_id, min_tick, max_tick,
-                                                                    action);
+    return tracepoint_event_buffer_->ForEachTracepointEventOfThreadInTimeRange(thread_id, min_tick,
+                                                                               max_tick, action);
   }
 
   [[nodiscard]] const std::map<uint64_t, orbit_client_protos::TracepointEventInfo>&
--- a/OrbitCore/TracepointEventBuffer.cpp
+++ b/OrbitCore/TracepointEventBuffer.cpp
@@ -11,14 +11,21 @@ void TracepointEventBuffer::AddTracepointEventAndMapToThreads(uint64_t time,
                                                               int32_t process_id, int32_t thread_id,
                                                               int32_t cpu,
                                                               bool is_same_pid_as_target) {
+  ScopeLock lock(mutex_);
   if (!is_same_pid_as_target) {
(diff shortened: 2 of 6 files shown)


OrbitCore/TracepointEventBuffer.cpp
```diff
@@ -51,4 +47,18 @@ TracepointEventBuffer::GetTracepointsOfThread(int32_t thread_id) const {
   return it->second;
 }
 
-Mutex& TracepointEventBuffer::mutex() { return mutex_; }
\ No newline at end of file
+void TracepointEventBuffer::ForEachTracepointEventPerThread(
```
dpallotti, line 50: This is called `ForEachTracepoint...`, but then the `action` takes the whole map and the responsibility of iterating over the tracepoints is left to `action` itself. This seems wrong. Am I missing something? I would expect `action` to be simply `std::function<void(const orbit_client_protos::TracepointEventInfo&)>` and this function to iterate on threads *and* and events.

OrbitCore/TracepointEventBuffer.h
```diff
@@ -34,9 +39,8 @@ class TracepointEventBuffer {
     }
   }
 
-  Mutex& mutex();
-
  private:
+  int32_t kAllTracepointsNotInTargetProcessFakeTid = -1;
```
dimitry-, line 43: kNotTargetProcessThreadId?

OrbitGl/TracepointTrack.cpp
```diff
@@ -62,38 +62,38 @@ void TracepointTrack::UpdatePrimitives(uint64_t min_tick, uint64_t max_tick,
...
-
-      Vec2 pos(time_graph_->GetWorldFromTick(time) - kPickingBoxOffset, pos_[1] - track_height + 1);
-      Vec2 size(kPickingBoxWidth, track_height);
-      auto user_data = std::make_unique<PickingUserData>(
-          nullptr, [&](PickingId id) -> std::string { return GetSampleTooltip(id); });
-      user_data->custom_data_ = &it->second;
-      batcher->AddShadedBox(pos, size, z, kGreenSelection, std::move(user_data));
-    }
-  }
+  GOrbitApp->GetCaptureData().ForEachTracepointEventPerThread(
```
dpallotti, line 65: This is related to the other comment about `ForEachTracepointEventPerThread` taking an action that operates on a map instead of the single `TracepointEvent`s. The action passed here has too big of a scope. It should only do what is already done in the for loops, which is what is specific of the `TracepointEvent`s themselves, that is:
```
uint64_t time = it->first;
batcher->AddVerticalLine(pos, -track_height, z, kWhite);
```
and the other code block in the other branch. 
If you need to filter the events by time you can add a `ForEachTracepointEventOfThreadInTimeRange` that calls `lower_bound` and `upper_bound` internally.

dimitry- (approved):
otherwise lgtm

OrbitCore/TracepointEventBuffer.h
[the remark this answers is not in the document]
ronaldfw: +1 to using -2.

MariaSandru27: Of course! Thanks a lot for the suggestion!

OrbitGl/TracepointTrack.cpp
MariaSandru27: As discussed offline: I will update the code. Thanks a lot for the suggestion!

OrbitCore/TracepointEventBuffer.cpp
MariaSandru27: Thank you!

OrbitCore/TracepointEventBuffer.h
```diff
@@ -16,14 +16,22 @@
...
                                          bool is_same_pid_as_target);
 
   [[nodiscard]] const std::map<uint64_t, orbit_client_protos::TracepointEventInfo>&
   GetTracepointsOfThread(int32_t thread_id) const;
 
+  void ForEachTracepointEventOfThread(
+      int32_t thread_id, uint64_t min_tick, uint64_t max_tick,
+      const std::function<void(const orbit_client_protos::TracepointEventInfo&)>& action) const;
+
+  void ForEachTracepointEventOfThreadInTimeRange(
```
dpallotti, line 30: Why do you need both methods? Can't you have a single public 
`ForEachTracepointEventOfThreadInTimeRange(int32_t thread_id, uint64_t min_tick, uint64_t max_tick,
      const std::function<void(const orbit_client_protos::TracepointEventInfo&)>& action) const` 
with the two nested for loops?

OrbitCore/CaptureData.h
```diff
@@ -127,6 +123,13 @@ class CaptureData {
     return tracepoint_event_buffer_.get();
   }
 
+  void ForEachTracepointEventPerThread(
```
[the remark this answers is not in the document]
MariaSandru27: Thanks!

OrbitCore/TracepointEventBuffer.h
```diff
@@ -24,6 +24,11 @@ class TracepointEventBuffer {
   [[nodiscard]] const std::map<uint64_t, orbit_client_protos::TracepointEventInfo>&
   GetTracepointsOfThread(int32_t thread_id) const;
 
+  void ForEachTracepointEventPerThread(
```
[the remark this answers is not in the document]
MariaSandru27: Thanks!

OrbitCore/TracepointEventBuffer.h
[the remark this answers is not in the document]
MariaSandru27: Of course. Thank you!

OrbitCore/TracepointEventBuffer.h
MariaSandru27: Of course I could. I use 'ForEachTracepointEventOfThreadInTimeRange' in two places so I believed that not writing repeating code would be a good idea and would also make everything cleaner. Let me know if this is a good trade-off.

dpallotti: Actually, since `ForEachTracepointEventOfThreadInTimeRange` is used twice and you might want to keep it, I would do like this:
`ForEachTracepointEventOfThread` should be the only public one and called `ForEachTracepointEventOfThreadInTimeRange`, as that's what it's actually doing;
`ForEachTracepointEventOfThreadInTimeRange` is an internal method: it should be private (and maybe renamed to avoid the overloading here).

dpallotti: Actually the old `ForEachTracepointEventOfThreadInTimeRange` can just be a free-standing static function, and maybe rename to `ForEachTracepointEventInEventMapInTimeRange`

MariaSandru27: Maybe I am missing something, but what overloading? I do not think I have another method with the same name but different parameters.

dpallotti: To summarize:
old `ForEachTracepointEventOfThread` should become `ForEachTracepointEventOfThreadInTimeRange`;
old `ForEachTracepointEventOfThreadInTimeRange` should become a static free-standing function in the .cpp only, and be renamed to something like `ForEachTracepointEventInEventMapInTimeRange`.

MariaSandru27: Thank you so much for the suggestions! I will update the code. 👍

dpallotti (approved):
Thanks for bearing with me!

MariaSandru27:
> Thanks for bearing with me!

Thanks a lot for taking the time to review!

danielfenner (approved):
lgtm! Thanks for refactoring this!

PR #1278, opened 2020-09-24
Fix broken build caused by PR#1251

kisenshi:
Fix the broken build caused by adding the _OrbitClientGgp_ to the collector. 

Update `conanfile`  so the ggp client is also available in `bin/`.

It is also needed to update the `build.sh` so the client is copied from `bin/`


kisenshi:
This should fix the broken build caused by the previous PR. I have included a temp commit so the code is checked in the presubmit before merging.

kokoro/builds/build.sh
```diff
@@ -183,7 +183,7 @@ if [ -n "$1" ]; then
 
   # Package the Debian package, the signature and the ggp client into a zip for integration in the 
   # installer.
-  if [ -f ${KEYSTORE_PATH}/74938_SigningPrivateGpg ] && [[ $CONAN_PROFILE == ggp_* ]]; then
+  if true then
```
beckerhe, line 186: Nit: `if true; then`

But I suggest you leave the `ggp_*` check. Otherwise the windows build will fail.

```
if [[ $CONAN_PROFILE == ggp_* ]]; then
```

kisenshi: Ah ok, I realised about the ; but didnt know this. will update it again. Thanks!

kisenshi:
Build seems to work now :)

kokoro/builds/build.sh
```diff
@@ -183,12 +183,11 @@ if [ -n "$1" ]; then
 
   # Package the Debian package, the signature and the ggp client into a zip for integration in the 
   # installer.
-  if [ -f ${KEYSTORE_PATH}/74938_SigningPrivateGpg ] && [[ $CONAN_PROFILE == ggp_* ]]; then
+  if [[ $CONAN_PROFILE == ggp_* ]]; then
     pushd "${REPO_ROOT}/build/package" > /dev/null
     mkdir -p Orbit/collector
     cp -v OrbitProfiler*.deb Orbit/collector/
-    cp -v OrbitProfiler*.deb.asc Orbit/collector/
```
danielfenner, line 190: The .deb.asc file is the signature of the .deb package. We definitely need that one.

kisenshi: It was removed temporally because the test was failing, will be included when the temp commits are removed

danielfenner (approved):
lgtm!

beckerhe (approved):
Thanks a lot!

kisenshi:
Merging the fix. Thanks a lot!

PR #1279, opened 2020-09-24
Manual instrumentation documentation (Orbit.h)

pierricgimmig:


Orbit.h
```diff
@@ -9,45 +9,173 @@
...
+//
+// Parameters:
+// name: [const char*] Label to be displayed on current time slice (kMaxEventStringSize characters).
+// col: [orbit::Color] User defined color for the current time slice.
+//
 #define ORBIT_SCOPE(name) ORBIT_SCOPE_WITH_COLOR(name, orbit::Color::kAuto)
 #define ORBIT_SCOPE_WITH_COLOR(name, col) orbit_api::Scope ORBIT_VAR(name, col)
 
-// ORBIT_START/ORBIT_STOP: profile time span on a single thread.
+// ORBIT_START/ORBIT_STOP: Profile sections inside a scope.
```
ronaldfw, line 69: Similarly here, also mention ORBIT_{START, STOP}_WITH_COLOR here?

Orbit.h
```diff
@@ -9,45 +9,173 @@
...
+//     DoSomeMoreWork();
+//   } else {
+//     ORBIT_SCOPE_WITH_COLOR("DoSomeOtherWork", orbit::Color::kLightGreen);
+//     DoSomeOtherWork();
+//   }
+// }
+//
+// Parameters:
+// name: [const char*] Label to be displayed on current time slice (kMaxEventStringSize characters).
+// col: [orbit::Color] User defined color for the current time slice.
```
ronaldfw, line 64: Maybe point out that predefined colors can be found below? 

"(see below for color definitions)"

Orbit.h
```diff
@@ -9,45 +9,173 @@
...
+// instruments. For manual instrumentation to appear in your Orbit capture, make sure that symbols
+// have been loaded for the manually instrumented modules.
 
 // To disable manual instrumentation macros, define ORBIT_API_ENABLED as 0.
 #define ORBIT_API_ENABLED 1
 
 #if ORBIT_API_ENABLED
 
-// ORBIT_SCOPE: profile current scope.
+// ORBIT_SCOPE: Profile current scope.
```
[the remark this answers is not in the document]
ronaldfw: Alternatively, mention further above that many of the macros have a _WITH_COLOR variant that allows specifying the timeslice color.

Orbit.h
```diff
@@ -9,45 +9,173 @@
...
+//
 #define ORBIT_START(name) ORBIT_START_WITH_COLOR(name, orbit::Color::kAuto)
 #define ORBIT_START_WITH_COLOR(name, col) orbit_api::Start(name, col)
 #define ORBIT_STOP() orbit_api::Stop()
 
-// ORBIT_START_ASYNC/ORBIT_STOP_ASYNC: profile time span across threads.
+// ORBIT_START_ASYNC/ORBIT_STOP_ASYNC: Profile time spans across scopes or threads.
+//
+// Overview:
+// Async time spans can be started in one scope and stopped on another. They will be displayed
```
ronaldfw, line 107: "stopped in another" (scope)

Orbit.h
```diff
@@ -9,45 +9,173 @@
...
+// in Orbit on a track uniquely identified by the "name" parameter. Note that those time slices
+// do not represent hierarchical information.
+//
+// Note:
+// We limit the maximum number of characters of the "name" parameter to "kMaxEventStringSize". This
+// limitation will be lifted as we roll out a new dynamic instrumentation implementation on Linux.
+// It is possible however to add per-time slice additional information using the ASYNC_STRING macro.
```
ronaldfw, line 114: Should we point out performance considerations when using the ASYNC_STRING macro?

Orbit.h
```diff
@@ -9,45 +9,173 @@
...
+//
+// Example usage: Tracking "File IO" operations.
+// Thread 1: ORBIT_START_ASYNC("File IO", unique_64_bit_id);  // File IO request site.
+// Thread 1 or 2: ORBIT_ASYNC_STRING(unique_64_bit_id, "My very long file path");
+// Thread 1 or 2: ORBIT_STOP_ASYNC(unique_64_bit_id);  // File IO result site.
+//
+// Parameters:
+// name: [const char*] Name of the *track* that will display the async events in Orbit.
+// id: [uint64_t] A user-provided unique id for the time slice.
```
ronaldfw, line 123: I think it could help understanding here to add how the unique id is used: "This unique id is used to match start and stop calls."

IMO, that clearly emphasizes that it *has* to be unique and strange stuff will happen if it is not.

Orbit.h
```diff
@@ -9,45 +9,173 @@
...
+// col: [orbit::Color] User defined color for the current time slice.
+//
 #define ORBIT_START_ASYNC(name, id) ORBIT_START_ASYNC_WITH_COLOR(name, id, orbit::Color::kAuto)
 #define ORBIT_START_ASYNC_WITH_COLOR(name, id, col) orbit_api::StartAsync(name, id, col)
 #define ORBIT_STOP_ASYNC(id) orbit_api::StopAsync(id)
+
+// ORBIT_ASYNC_STRING: Provide additional information for an async time span.
```
ronaldfw, line 130: The "additional information" sounds like more than it is (a string) and would initially confuse me as a user. Maybe just say: "Provide an additional string for an async time span."?

Orbit.h
```diff
@@ -9,45 +9,173 @@
...
+//
+// Overview:
+// Provide additional information to be displayed on the time slice corresponding to "id".
+//
+// Example usage: Tracking "File IO" operations.
+// Thread 1: ORBIT_START_ASYNC("File IO", unique_64_bit_id);
+// Thread 1 or 2: ORBIT_ASYNC_STRING(unique_64_bit_id, "My very long file path");
+// Thread 1 or 2: ORBIT_STOP_ASYNC(unique_64_bit_id);
+//
```
ronaldfw, line 139: IMO, it would help to state what will be displayed on the timeslice in this example.

Orbit.h
```diff
@@ -9,45 +9,173 @@
...
+// col: [orbit::Color] User defined color for the current string.
+//
 #define ORBIT_ASYNC_STRING(str, id) orbit_api::AsyncString(str, id, orbit::Color::kAuto)
 #define ORBIT_ASYNC_STRING_WITH_COLOR(str, id, col) orbit_api::AsyncString(str, id, col)
 
-// ORBIT_[type]: graph variables.
+// ORBIT_[type]: Graph variables.
+//
+// Overview:
+// Send values to be plotted in a track uniquely identified by "name".
```
ronaldfw, line 151: Mention that these are plotted over time? It says this below, but it's easier to understand if you state it here already.

ronaldfw (approved):
Thanks for this, @pierricgimmig! I have mostly minor suggestions, feel free to address what you consider important, and feel free to merge as you see fit!

pierricgimmig:
Hi Ronald, thanks for the comments. I addressed all of them and added a bit more information on the experimental nature of the feature and on the general performance overhead which I'd like to pass by you to double check. Thanks again!

Orbit.h
```diff
@@ -9,45 +9,203 @@
...
+//
+// Note: There is a performance overhead incurred by using the ASYNC_STRING macro. The arbitrarily
+//       long input string will be chunked into substrings of "kMaxEventStringSize" length that will
+//       individually be emitted as multiple profiling events.
+//
+// Example usage: Tracking "File IO" operations.
+// Thread 1: ORBIT_START_ASYNC("File IO", unique_64_bit_id);  // File IO request site.
+// Thread 1 or 2: ORBIT_ASYNC_STRING(unique_64_bit_id, "My very long file path");
+// Thread 1 or 2: ORBIT_STOP_ASYNC(unique_64_bit_id);  // File IO result site.
+// Result: Multiple time slices labeled with the results of "io_request->GetFileName()" will appear
```
ronaldfw, line 163: I don't understand the relation of "io_request->GetFileName()" to the above example. Am I missing something?

ronaldfw (approved):
LGTM, just one question left.

Orbit.h
ronaldfw: Seems this was not addressed?

Orbit.h
```diff
@@ -9,45 +9,173 @@
...
 
-// ORBIT_SCOPE: profile current scope.
+// ORBIT_SCOPE: Profile current scope.
+//
+// Overview:
+// ORBIT_SCOPE will profile the time between "now" and the end of the current scope.
+//
+// Note:
+// We limit the maximum number of characters of the "name" parameter to "kMaxEventStringSize". This
+// limitation will be lifted as we roll out a new dynamic instrumentation implementation on Linux.
```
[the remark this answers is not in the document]
ronaldfw: Thoughts on this?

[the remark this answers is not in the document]
ronaldfw: Sorry, was looking at the wrong version, this was addressed, thanks.

Orbit.h
ronaldfw: Looked at the wrong version, sorry.

ronaldfw:
Merged as we need this for 1.53.

PR #1285, opened 2020-09-25
Disable ImGui ini file writing 

beckerhe:
This disables the ImGui configuration file which we don't use anyway.

I haven't found any official documentation about that because the ImGui author says [here](https://github.com/ocornut/imgui/issues/923) and [here](https://www.reddit.com/r/imgui/comments/4xstmx/no_imguiini/) that setting the filename to nullptr will disable the mechanism.


danielfenner (approved):
Cool, thanks for fixing!

PR #1286, opened 2020-09-25
[TARGETING RELEASE] Fix Potential Nullpointer Crashes in CaptureData

florian-kuebler:
Currently, process_ will be nullptr in default constructed
CaptureData. As we are accessing it from many threads,
we should ensure, that we do not access it if it is nullptr.

Bug: http://b/169391098
Test: Compile


florian-kuebler:
As discussed offline, will not target the release. And on master we properly fix this, by making capture data an optional in app, and removing the default constructor, so that will be never null.

PR #1290, opened 2020-09-25
Scrollbar colors

reichlfl:
**Adjust the colors of scrollbars in the Capture Window**

The capture window scrollbars are drawn in OpenGl, and their Style
does not match the other scrollbars in Orbit. For users, it was not evident
which part of the scrollbar is the bar background and which part it the
slider, especially when the slider occupied about 50% of the bar.

Confusingly, our color choice (dark slider on lighter bar) was the opposite
of our QT bars (light, "paneled" slider on dark bar).

This PR fixes the look of the sliders to have them closer to the QT UI.

**Before**
image

**After**
image

(Vertical slider is drawn regularly, horizontal slider is the "on click" version
during drag)

To simplify the code and to prepare for a potential future extension of the
horizontal slider (e.g. showing frame tracks in the background), this PR 
splits the GlSlider into GlVerticalSlider and GlHorizontal slider.

First commit is the actual changes, second commit the cleanup and renaming
of existing variables.

### DIFF
--- a/OrbitGl/Batcher.cpp
+++ b/OrbitGl/Batcher.cpp
@@ -63,12 +63,32 @@ void Batcher::AddBox(const Box& box, const Color& color, std::shared_ptr<Pickabl
   AddBox(box, colors, picking_color, nullptr);
 }
 
+void Batcher::AddShadedBox(Vec2 pos, Vec2 size, float z, const Color& color) {
+  AddShadedBox(pos, size, z, color, std::unique_ptr<PickingUserData>(),
+               ShadingDirection::kLeftToRight);
+}
+
 void Batcher::AddShadedBox(Vec2 pos, Vec2 size, float z, const Color& color,
-                           std::unique_ptr<PickingUserData> user_data) {
+                           ShadingDirection shading_direction) {
+  AddShadedBox(pos, size, z, color, std::unique_ptr<PickingUserData>(), shading_direction);
+}
+
+void Batcher::AddShadedBox(Vec2 pos, Vec2 size, float z, const Color& color,
+                           std::unique_ptr<PickingUserData> user_data,
+                           ShadingDirection shading_direction) {
   std::array<Color, 4> colors;
-  GetBoxGradientColors(color, &colors);
+  GetBoxGradientColors(color, &colors, shading_direction);
   Box box(pos, size, z);
   AddBox(box, colors, std::move(user_data));
+}
+
+void Batcher::AddShadedBox(Vec2 pos, Vec2 size, float z, const Color& color,
+                           std::shared_ptr<Pickable> pickable, ShadingDirection shading_direction) {
+  std::array<Color, 4> colors;
+  GetBoxGradientColors(color, &colors, shading_direction);
+  Color picking_color = picking_manager_->GetPickableColor(pickable, batcher_id_);
+  Box box(pos, size, z);
+  AddBox(box, colors, picking_color, nullptr);
 }
 
 void Batcher::AddBox(const Box& box, const std::array<Color, 4>& colors, const Color& picking_color,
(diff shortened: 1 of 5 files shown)


OrbitGl/Batcher.cpp
```diff
@@ -155,14 +175,41 @@ const TextBox* Batcher::GetTextBox(PickingId id) {
...
+      (*colors)[2] = color;
+      break;
+    case ShadingDirection::kBottomToTop:
+      (*colors)[1] = Color(static_cast<uint8_t>(dark[0]), static_cast<uint8_t>(dark[1]),
+                           static_cast<uint8_t>(dark[2]), color[3]);
+      (*colors)[2] = (*colors)[1];
+      (*colors)[0] = color;
+      (*colors)[3] = color;
+      break;
+  }
```
pierricgimmig, line 212: Could "Color(static_cast<uint8_t>(dark[0]), static_cast<uint8_t>(dark[1]),  static_cast<uint8_t>(dark[2]), color[3]);" be merged into a single "dark_color" temporary variable? I think it would be more obvious that 2 vertices are dark and the two others are the original color. WDYT?

could

OrbitGl/GlSlider.cpp
```diff
@@ -34,41 +32,76 @@ void GlSlider::SetSliderWidthRatio(float a_WidthRatio)  // [0,1]
...
 }
 
-void GlSlider::OnPickHorizontal(int a_X, int /*a_Y*/) {
-  float canvasWidth = m_Canvas->getWidth();
-  float sliderWidth = m_Length * canvasWidth;
-  float nonSliderWidth = canvasWidth - sliderWidth;
+Color GlSlider::GetDarkerColor(const Color& color) {
+  return Color(static_cast<unsigned char>(color[0] * 0.75),
+               static_cast<unsigned char>(color[1] * 0.75),
+               static_cast<unsigned char>(color[2] * 0.75), 255);
```
pierricgimmig, line 44: 1.25 and 0.75 should probably be constants. Code could maybe be extracted into a function to "scale" a color. It could be used in GetBoxGradientColors as well.

OrbitGl/GlSlider.cpp
```diff
@@ -34,41 +32,76 @@ void GlSlider::SetSliderWidthRatio(float a_WidthRatio)  // [0,1]
...
-  float x = static_cast<float>(a_X);
-  float sliderX = m_Ratio * nonSliderWidth;
-  m_PickingRatio = (x - sliderX) / sliderWidth;
+void GlSlider::DrawBackground(GlCanvas* canvas, float x, float y, float width, float height) {
+  Batcher* batcher = canvas->GetBatcher();
+  const Color dark_border_color = GetDarkerColor(m_BarColor);
+
+  // Dark border
+  {
+    Box box(Vec2(x, y), Vec2(width, height), GlCanvas::kZValueSliderBg - 0.0001f);
```
pierricgimmig, line 53: 0.0001f -> constexpr float kEpsilon= 0.0001.f; ?

OrbitGl/GlSlider.cpp
```diff
@@ -34,41 +32,76 @@ void GlSlider::SetSliderWidthRatio(float a_WidthRatio)  // [0,1]
...
+    batcher->AddBox(box, dark_border_color, shared_from_this());
+  }
+  // Background itself
+  {
+    Box box(Vec2(x + 1, y + 1), Vec2(width - 2, height - 2), GlCanvas::kZValueSliderBg);
```
pierricgimmig, line 58: Vec2 contain floats, we should use 1.f and 2.f, and declare those as constants.

OrbitGl/GlSlider.cpp
```diff
@@ -34,41 +32,76 @@ void GlSlider::SetSliderWidthRatio(float a_WidthRatio)  // [0,1]
...
+    Box box(Vec2(x, y), Vec2(width, height), GlCanvas::kZValueSlider - 0.0002f);
+
+    batcher->AddBox(box, dark_border_color, shared_from_this());
+  }
+  {
+    // Light inner border
+    Box box(Vec2(x + 1, y + 1), Vec2(width - 2, height - 2), GlCanvas::kZValueSlider - 0.0001f);
+
+    batcher->AddBox(box, light_border_color, shared_from_this());
+  }
```
pierricgimmig: Could we rename the "box" variables into "outer_box" and "inner_box" maybe and get rid of the extra scopes?

pierricgimmig (commented):
Hi Florian, this looks really good!! Thanks for the change! I like the new "gradient direction" functionality, it will surely come in handy again. Out of curiosity, had you considered adding the real Qt toolbars?

reichlfl:
> Out of curiosity, had you considered adding the real Qt toolbars?

I have, but I wanted to do some more experiments on the scrollbars (e.g. https://b/162487253), and this would be simpler when keeping the custom scrollbars.

OrbitGl/Batcher.cpp
reichlfl: Good idea, that makes the code look a lot cleaner, thanks!

OrbitGl/GlSlider.cpp
reichlfl: Changed to a constant in GlSlider. `GetBoxGradientColors` uses a very different factor though, so I didn't unify this.

OrbitGl/GlSlider.cpp
reichlfl: Done

OrbitGl/GlSlider.cpp
reichlfl: Done

OrbitGl/GlSlider.cpp
reichlfl: Yes, good point. I think the scopes were a remnant from when all of this was done in a single method, and I didn't want to have 5 different box-variables flying around.

pierricgimmig (approved):
Thanks for the changes!

PR #1291, opened 2020-09-25
Fix Function Offset in Serialization

florian-kuebler:
Sampling information just contains raw absolute addresses.
Mapping those to functions (via offsets) and function names
is done via two mechanisms:
1. If available, the loaded symbols on the UI are used.
2. Otherwise, it will be checked if there are address informations
   from unwinding (the service).
On serialization, we do not store the complete information of loaded symbols (yet?),
but only save the address information, received from the service.

As we do not want to modify this information (potentially during capturing,
as loading symbols is possible while capturing), we already patch the serialized
version of the address information. However, that was limited to
writing the function name. In particular the offset was missing.

This adds this information.

Bug: https://169391773
Test: Take a capture, load symbols, save the capture, restart orbit, load capture -> symbols are correct

### DIFF
--- a/OrbitClientModel/CaptureSerializer.cpp
+++ b/OrbitClientModel/CaptureSerializer.cpp
@@ -4,6 +4,8 @@
 
 #include "OrbitClientModel/CaptureSerializer.h"
 
+#include <OrbitClientData/FunctionUtils.h>
+
 #include <memory>
 
 #include "Callstack.h"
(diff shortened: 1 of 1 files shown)


dpallotti (approved):
Have you tried with a stripped binary like the one Anton was using or like Infiltrator, i.e., one for which `LinuxAddressInfo`s actually need patching?

florian-kuebler:
> Have you tried with a stripped binary like the one Anton was using or like Infiltrator, i.e., one for which `LinuxAddressInfo`s actually need patching?

I have tried with CP, but will try again on Monday with Infiltrator.

OrbitClientModel/CaptureSerializer.cpp
```diff
@@ -64,9 +66,15 @@ CaptureInfo GenerateCaptureInfo(
   for (const auto& address_info : capture_data.address_infos()) {
     orbit_client_protos::LinuxAddressInfo* added_address_info = capture_info.add_address_infos();
     added_address_info->CopyFrom(address_info.second);
-    // Fix names in address infos (some might only be in process):
-    added_address_info->set_function_name(
-        capture_data.GetFunctionNameByAddress(added_address_info->absolute_address()));
+    const orbit_client_protos::FunctionInfo* function =
+        capture_data.FindFunctionByAddress(added_address_info->absolute_address(), false);
```
IrinaShkviro, line 70: I'm curious is it possible to fix capture_data.address_infos before the serialization (e.g. when finalizing capture)? Then we don't need to fix names/offset/module during the serialization.

florian-kuebler: I agree that we need a more conceptual fix. But I don't think we can easily patch the address info right now, due to synchronization. 
Address info might be queried also during a capture (due to selection of callstacks), so it can't be just fixed on capture synchronization. Also you might load symbols during a capture or after the capture. This all would make it not cleaner.

We need at some point a way to get rid of the "reads" from capture data during the capture (and rather stream data to the readers), as well as some clever way to handle the two sources for address information. 
For now, I would argue against doing a somewhat better, but more complicated solution, and go with this quick fix.

antonrohr:
Using my own binary with stripped symbols, the loaded capture still does not quite look like the capture I took.

florian-kuebler:
> Using my own binary with stripped symbols, the loaded capture still does not quite look like the capture I took.

Thanks. That was a result from using `FunctionUtils::Offset`, which is apparently not the Offset inside the function, but the function's offset in the binary file. I fixed it, to use the right offset.

antonrohr (approved):
lgtm

IrinaShkviro (approved):

PR #1292, opened 2020-09-25
Reorder some fields/variables in OrbitLinuxTracing

dpallotti:
Don't split specific tracepoints, bring stack and callchain sampling together.
This is just for consistency reasons.


florian-kuebler (approved):
Makes sense. Thanks!

PR #1293, opened 2020-09-25
Render a tracepoint track for the system-wide tracepoints

MariaSandru27:
Note: In a future PR the tooltip will also specify the process id and thread id. Also there will be a distinction in terms of colors of what tracepoints were triggered in the target process and what not.

Screenshot 2020-09-25 at 5 59 02 PM

### DIFF
--- a/OrbitCore/SamplingProfiler.cpp
+++ b/OrbitCore/SamplingProfiler.cpp
@@ -75,6 +75,7 @@ std::unique_ptr<SortedCallstackReport> SamplingProfiler::GetSortedCallstackRepor
 }
 
 const int32_t SamplingProfiler::kAllThreadsFakeTid = -1;
+const int32_t SamplingProfiler::kAllTracepointsFakeTid = -3;
 
 void SamplingProfiler::SortByThreadUsage() {
   sorted_thread_sample_data_.reserve(thread_id_to_sample_data_.size());
--- a/OrbitCore/SamplingProfiler.cpp
+++ b/OrbitCore/SamplingProfiler.cpp
@@ -75,7 +75,6 @@ std::unique_ptr<SortedCallstackReport> SamplingProfiler::GetSortedCallstackRepor
 }
 
 const int32_t SamplingProfiler::kAllThreadsFakeTid = -1;
-const int32_t SamplingProfiler::kAllTracepointsFakeTid = -3;
 
 void SamplingProfiler::SortByThreadUsage() {
   sorted_thread_sample_data_.reserve(thread_id_to_sample_data_.size());
--- a/OrbitCore/TracepointEventBuffer.cpp
+++ b/OrbitCore/TracepointEventBuffer.cpp
@@ -76,10 +76,10 @@ void TracepointEventBuffer::ForEachTracepointEventOfThreadInTimeRange(
         ForEachTracepointEventInEventMapInTimeRange(min_tick, max_tick, entry.second, action);
       }
     }
-    return;
+  } else {
+    ForEachTracepointEventInEventMapInTimeRange(min_tick, max_tick,
+                                                GetTracepointsOfThread(thread_id), action);
   }
-  ForEachTracepointEventInEventMapInTimeRange(min_tick, max_tick, GetTracepointsOfThread(thread_id),
-                                              action);
 }
 
 uint32_t TracepointEventBuffer::GetNumTracepointsForThreadId(int32_t thread_id) {
--- a/OrbitCore/TracepointEventBuffer.cpp
+++ b/OrbitCore/TracepointEventBuffer.cpp
@@ -78,7 +78,7 @@ void TracepointEventBuffer::ForEachTracepointEventOfThreadInTimeRange(
                                               action);
 }
 
-uint32_t TracepointEventBuffer::GetTracepointsEventsCount(int32_t thread_id) const {
+uint32_t TracepointEventBuffer::GetNumTracepointsForThreadId(int32_t thread_id) const {
   ScopeLock lock(mutex_);
   uint32_t count = 0;
 
@@ -92,8 +92,9 @@ uint32_t TracepointEventBuffer::GetTracepointsEventsCount(int32_t thread_id) con
         count += entry.second.size();
       }
     }
-  } else
+  } else {
     count += GetTracepointsOfThread(thread_id).size();
+  }
 
   return count;
 }
--- a/OrbitCore/TracepointEventBuffer.cpp
+++ b/OrbitCore/TracepointEventBuffer.cpp
@@ -12,6 +12,7 @@ void TracepointEventBuffer::AddTracepointEventAndMapToThreads(uint64_t time,
                                                               int32_t cpu,
                                                               bool is_same_pid_as_target) {
   ScopeLock lock(mutex_);
+  no_total_tracepoints_++;
   if (!is_same_pid_as_target) {
     std::map<uint64_t, orbit_client_protos::TracepointEventInfo>&
         event_map_tracepoints_not_in_target_process = tracepoint_events_[kNotTargetProcessThreadId];
@@ -22,6 +23,7 @@ void TracepointEventBuffer::AddTracepointEventAndMapToThreads(uint64_t time,
     event.set_pid(process_id);
     event.set_cpu(cpu);
     event_map_tracepoints_not_in_target_process[time] = std::move(event);
+    no_tracepoints_not_in_target_process_++;
     return;
   }
 
@@ -80,21 +82,13 @@ void TracepointEventBuffer::ForEachTracepointEventOfThreadInTimeRange(
 
 uint32_t TracepointEventBuffer::GetNumTracepointsForThreadId(int32_t thread_id) const {
   ScopeLock lock(mutex_);
-  uint32_t count = 0;
 
   if (thread_id == SamplingProfiler::kAllTracepointsFakeTid) {
-    for (const auto& entry : tracepoint_events_) {
-      count += entry.second.size();
-    }
+    return no_total_tracepoints_;
   } else if (thread_id == SamplingProfiler::kAllThreadsFakeTid) {
-    for (const auto& entry : tracepoint_events_) {
-      if (entry.first != kNotTargetProcessThreadId) {
-        count += entry.second.size();
-      }
-    }
+    return no_total_tracepoints_ - no_tracepoints_not_in_target_process_;
   } else {
-    count += GetTracepointsOfThread(thread_id).size();
+    return GetTracepointsOfThread(thread_id).size();
   }
 
-  return count;
 }
--- a/OrbitCore/TracepointEventBuffer.cpp
+++ b/OrbitCore/TracepointEventBuffer.cpp
@@ -6,6 +6,9 @@
 
 using orbit_client_protos::TracepointEventInfo;
 
+const int32_t TracepointEventBuffer::kNotTargetProcessThreadId = -2;
+const int32_t TracepointEventBuffer::kAllTracepointsFakeTid = -3;
+
 void TracepointEventBuffer::AddTracepointEventAndMapToThreads(uint64_t time,
                                                               uint64_t tracepoint_hash,
                                                               int32_t process_id, int32_t thread_id,
@@ -23,7 +26,6 @@ void TracepointEventBuffer::AddTracepointEventAndMapToThreads(uint64_t time,
     event.set_pid(process_id);
     event.set_cpu(cpu);
     event_map_tracepoints_not_in_target_process[time] = std::move(event);
-    num_tracepoints_not_in_target_process_++;
     return;
   }
 
@@ -64,7 +66,7 @@ void TracepointEventBuffer::ForEachTracepointEventOfThreadInTimeRange(
     int32_t thread_id, uint64_t min_tick, uint64_t max_tick,
     const std::function<void(const orbit_client_protos::TracepointEventInfo&)>& action) const {
   ScopeLock lock(mutex_);
-  if (thread_id == SamplingProfiler::kAllTracepointsFakeTid) {
+  if (thread_id == TracepointEventBuffer::kAllTracepointsFakeTid) {
     for (const auto& entry : tracepoint_events_) {
       ForEachTracepointEventInEventMapInTimeRange(min_tick, max_tick, entry.second, action);
     }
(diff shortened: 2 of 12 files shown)


OrbitGl/ThreadTrack.cpp
```diff
@@ -145,9 +145,11 @@ void ThreadTrack::Draw(GlCanvas* canvas, PickingMode picking_mode) {
   TimerTrack::Draw(canvas, picking_mode);
 
   float event_track_height = time_graph_->GetLayout().GetEventTrackHeight();
-  event_track_->SetPos(pos_[0], pos_[1]);
-  event_track_->SetSize(canvas->GetWorldWidth(), event_track_height);
-  event_track_->Draw(canvas, picking_mode);
+  if (thread_id_ != SamplingProfiler::kAllTracepointsFakeTid) {
```
pierricgimmig, line 148: Instead of introducing a dependency on SamplingProfiler, maybe we could set a flag on ThreadTrack to control whether or not we want to display event tracks when they are empty. Something like thread_track_->ShowEmptyEventTracks(false);

OrbitGl/ThreadTrack.cpp
```diff
@@ -157,8 +159,10 @@ void ThreadTrack::Draw(GlCanvas* canvas, PickingMode picking_mode) {
 void ThreadTrack::OnPick(int /*x*/, int /*y*/) { GOrbitApp->set_selected_thread_id(thread_id_); }
 
 void ThreadTrack::UpdatePrimitives(uint64_t min_tick, uint64_t max_tick, PickingMode picking_mode) {
-  event_track_->SetPos(pos_[0], pos_[1]);
-  event_track_->UpdatePrimitives(min_tick, max_tick, picking_mode);
+  if (thread_id_ != SamplingProfiler::kAllTracepointsFakeTid) {
+    event_track_->SetPos(pos_[0], pos_[1]);
+    event_track_->UpdatePrimitives(min_tick, max_tick, picking_mode);
+  }
```
pierricgimmig, line 165: Same as above, it would be better to avoid introducing a dependency to SamplingProfiler.

OrbitGl/ThreadTrack.cpp
```diff
@@ -168,7 +172,9 @@ void ThreadTrack::UpdatePrimitives(uint64_t min_tick, uint64_t max_tick, Picking
 
 void ThreadTrack::SetTrackColor(Color color) {
   ScopeLock lock(mutex_);
-  event_track_->SetColor(color);
+  if (thread_id_ != SamplingProfiler::kAllTracepointsFakeTid) {
```
pierricgimmig, line 175: Same as above, it would be better to avoid introducing a dependency to SamplingProfiler.

OrbitGl/ThreadTrack.cpp
```diff
@@ -229,8 +235,11 @@ float ThreadTrack::GetHeight() const {
   const uint32_t collapsed_depth = (GetNumTimers() == 0) ? 0 : 1;
   const uint32_t depth = is_collapsed ? collapsed_depth : GetDepth();
   return layout.GetTextBoxHeight() * depth +
-         (depth > 0 ? layout.GetSpaceBetweenTracksAndThread() : 0) + layout.GetEventTrackHeight() +
-         layout.GetTrackBottomMargin() + tracepoint_track_->GetHeight();
+         (depth > 0 ? layout.GetSpaceBetweenTracksAndThread() : 0) +
+         (thread_id_ != SamplingProfiler::kAllTracepointsFakeTid
+              ? layout.GetEventTrackHeight() + layout.GetTrackBottomMargin()
+              : 0) +
+         tracepoint_track_->GetHeight();
```
pierricgimmig, line 242: Same as above, it would be better to avoid introducing a dependency to SamplingProfiler.

pierricgimmig (commented):
Hi Maria, thanks for the change! The new global tracepoint track seems handy! Please see comments.

OrbitCore/TracepointEventBuffer.cpp
```diff
@@ -72,3 +77,23 @@ void TracepointEventBuffer::ForEachTracepointEventOfThreadInTimeRange(
...
+      count += entry.second.size();
+    }
+  } else if (thread_id == SamplingProfiler::kAllThreadsFakeTid) {
+    for (const auto& entry : tracepoint_events_) {
+      if (entry.first != kNotTargetProcessThreadId) {
+        count += entry.second.size();
+      }
+    }
+  } else
+    count += GetTracepointsOfThread(thread_id).size();
```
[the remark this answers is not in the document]
MariaSandru27: Of course. Thank you!

OrbitCore/TracepointEventBuffer.cpp
```diff
@@ -72,3 +77,23 @@ void TracepointEventBuffer::ForEachTracepointEventOfThreadInTimeRange(
   ForEachTracepointEventInEventMapInTimeRange(min_tick, max_tick, GetTracepointsOfThread(thread_id),
                                               action);
 }
+
+uint32_t TracepointEventBuffer::GetTracepointsEventsCount(int32_t thread_id) const {
```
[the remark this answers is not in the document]
MariaSandru27: Sure. Thank you!

OrbitGl/ThreadTrack.cpp
MariaSandru27: Thank you so much! I eliminated any dependency on SamplingProfiler in ThreadTrack and modified the EventTrack to not show the event bar if there are no callstacks.

OrbitCore/TracepointEventBuffer.cpp
```diff
@@ -61,7 +62,11 @@ void TracepointEventBuffer::ForEachTracepointEventOfThreadInTimeRange(
     int32_t thread_id, uint64_t min_tick, uint64_t max_tick,
     const std::function<void(const orbit_client_protos::TracepointEventInfo&)>& action) const {
   ScopeLock lock(mutex_);
-  if (thread_id == SamplingProfiler::kAllThreadsFakeTid) {
+  if (thread_id == SamplingProfiler::kAllTracepointsFakeTid) {
+    for (const auto& entry : tracepoint_events_) {
+      ForEachTracepointEventInEventMapInTimeRange(min_tick, max_tick, entry.second, action);
```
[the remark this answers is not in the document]
MariaSandru27: The logic behind the name is that the function performs an action (any action you want since you pass that as a parameter) for each tracepoint event in an event map in a time range. @dpallotti suggested me this name but if you have any idea feel free to tell. To come with a name expressing what the function is doing to the passed element is a bit tricky since the action can be any action you want.

OrbitCore/SamplingProfiler.cpp
```diff
@@ -75,6 +75,7 @@ std::unique_ptr<SortedCallstackReport> SamplingProfiler::GetSortedCallstackRepor
 }
 
 const int32_t SamplingProfiler::kAllThreadsFakeTid = -1;
+const int32_t SamplingProfiler::kAllTracepointsFakeTid = -3;
```
dpallotti, line 78: This doesn't really belong in `SamplingProfiler`, I would assume it's better to have it next to `kNotTargetProcessThreadId` in `TracepointEventBuffer`.

OrbitCore/TracepointEventBuffer.h
```diff
@@ -37,8 +37,12 @@ class TracepointEventBuffer {
     }
   }
 
+  uint32_t GetNumTracepointsForThreadId(int32_t thread_id) const;
+
  private:
   int32_t kNotTargetProcessThreadId = -2;
+  int32_t no_total_tracepoints_ = 0;
```
dpallotti, line 44: I would use `num_total_tracepoints_` or `total_tracepoint_count_`, `no` sounds more like&mdash;well&mdash;"no" than like "number".

OrbitCore/TracepointEventBuffer.h
```diff
@@ -37,8 +37,12 @@ class TracepointEventBuffer {
     }
   }
 
+  uint32_t GetNumTracepointsForThreadId(int32_t thread_id) const;
+
  private:
   int32_t kNotTargetProcessThreadId = -2;
```
dpallotti, line 43: This should be a `static const` by the way.

OrbitCore/TracepointEventBuffer.h
```diff
@@ -37,8 +37,12 @@ class TracepointEventBuffer {
...
+  int32_t no_total_tracepoints_ = 0;
+  int32_t no_tracepoints_not_in_target_process_ = 0;
```
dpallotti, line 45: Same here: `num_tracepoints_not_in_target_process_` or `tracepoints_not_in_target_process_count_`.

OrbitCore/TracepointEventBuffer.cpp
```diff
@@ -72,3 +79,16 @@ void TracepointEventBuffer::ForEachTracepointEventOfThreadInTimeRange(
...
+  ScopeLock lock(mutex_);
+
+  if (thread_id == SamplingProfiler::kAllTracepointsFakeTid) {
+    return no_total_tracepoints_;
+  } else if (thread_id == SamplingProfiler::kAllThreadsFakeTid) {
+    return no_total_tracepoints_ - no_tracepoints_not_in_target_process_;
+  } else {
+    return GetTracepointsOfThread(thread_id).size();
+  }
+
```
dpallotti, line 93:
```suggestion
```

(remove empty line)

OrbitGl/TimeGraph.cpp
```diff
@@ -730,7 +734,10 @@ void TimeGraph::DrawTracks(GlCanvas* canvas, PickingMode picking_mode) {
     if (track->GetType() == Track::kThreadTrack) {
       auto thread_track = std::static_pointer_cast<ThreadTrack>(track);
       int32_t tid = thread_track->GetThreadId();
-      if (tid == SamplingProfiler::kAllThreadsFakeTid) {
+      if (tid == SamplingProfiler::kAllTracepointsFakeTid) {
+        thread_track->SetName("all tracepoints");
+        thread_track->SetLabel("all tracepoints");
```
dpallotti, line 739: As suggested offline, I would go for "All tracepoint events".

dimitry-:
```
--

clang-format determined the following necessary changes to your PR:
--- OrbitCore/TracepointEventBuffer.cpp	(before formatting)
+++ OrbitCore/TracepointEventBuffer.cpp	(after formatting)
@@ -90,5 +90,4 @@
   } else {
     return GetTracepointsOfThread(thread_id).size();
   }
-
 }


Here is the list of files that need formatting:
OrbitCore/TracepointEventBuffer.cpp
--
```

OrbitCore/TracepointEventBuffer.h
MariaSandru27: I see what you mean. Thanks a lot! I will update the code.

OrbitCore/TracepointEventBuffer.h
MariaSandru27: Of course! Thanks you!

OrbitCore/TracepointEventBuffer.cpp
MariaSandru27: Sure! Thank you!

OrbitGl/TimeGraph.cpp
MariaSandru27: Of course! Thank you!

OrbitCore/TracepointEventBuffer.cpp
```diff
@@ -22,6 +23,7 @@ void TracepointEventBuffer::AddTracepointEventAndMapToThreads(uint64_t time,
     event.set_pid(process_id);
     event.set_cpu(cpu);
     event_map_tracepoints_not_in_target_process[time] = std::move(event);
+    no_tracepoints_not_in_target_process_++;
```
danielfenner, line 26: You don't really need this: 
`tracepoint_events_[kNotTargetProcessThreadId].size()`
will be the same number.

OrbitCore/SamplingProfiler.cpp
danielfenner: Agree, if there isn't a *really* good reason the constants for the  fake id's should be in the same place.

danielfenner (commented):
one minor comment - rest lgtm

OrbitCore/SamplingProfiler.cpp
MariaSandru27: Absolutely! Thank you!

OrbitCore/TracepointEventBuffer.h
MariaSandru27: Thanks so much for pointing that out!

OrbitCore/TracepointEventBuffer.cpp
MariaSandru27: Thank you so much for pointing that out!

OrbitCore/TracepointEventBuffer.cpp
```diff
@@ -72,3 +81,15 @@ void TracepointEventBuffer::ForEachTracepointEventOfThreadInTimeRange(
   ForEachTracepointEventInEventMapInTimeRange(min_tick, max_tick, GetTracepointsOfThread(thread_id),
                                               action);
```
dpallotti, line 82: I think you should put this call in the `else` branch. Previously, an early return was used. But now there's an early return in the `else if` branch, but not in the `if` branch.

OrbitCore/TracepointEventBuffer.cpp
```diff
@@ -6,12 +6,16 @@
 
 using orbit_client_protos::TracepointEventInfo;
 
+const int32_t TracepointEventBuffer::kNotTargetProcessThreadId = -2;
```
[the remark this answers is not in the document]
danielfenner: If this would touch a bunch of additional files consider doing this in a follow up PR instead.

OrbitCore/TracepointEventBuffer.cpp
MariaSandru27: fixed it

OrbitGl/EventTrack.cpp
```diff
@@ -160,6 +164,8 @@ void EventTrack::SelectEvents() {
 }
 
 bool EventTrack::IsEmpty() const {
+  if (thread_id_ == TracepointEventBuffer::kAllTracepointsFakeTid) return true;
```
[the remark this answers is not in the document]
dpallotti: I would guess that this is saying that this "All tracepoint events" track has no sampling event, which is true, but is this check necessary?

dpallotti (approved):

OrbitGl/EventTrack.cpp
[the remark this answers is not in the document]
danielfenner: Discussed offline, Event track is the line with the call stack samples - we obviously have no calsstacks so this is empty.

OrbitCore/TracepointEventBuffer.cpp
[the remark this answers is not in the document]
danielfenner: kAllThreadsFakeTid has the identical meaning in SamplingProfiler, let's keep as is for now.

danielfenner (approved):
lgtm

OrbitGl/EventTrack.cpp
```diff
@@ -19,6 +19,10 @@ EventTrack::EventTrack(TimeGraph* a_TimeGraph) : Track(a_TimeGraph) {
 std::string EventTrack::GetTooltip() const { return "Left-click and drag to select samples"; }
 
 void EventTrack::Draw(GlCanvas* canvas, PickingMode picking_mode) {
+  if (IsEmpty()) {
+    return;
+  }
+
```
pierricgimmig, line 25: This will change the current behavior for the all the event tracks, which I don't think we want. If a thread has no sampling event, then we won't draw the colored event track. That event track is nice to identify the threads by color, I'd prefer keeping the original behavior.

We should early out only for the case of the AllTracePointsFakeTid.

pierricgimmig (commented):
Hi Maria, looks good but I do have one comment.

OrbitGl/EventTrack.cpp
MariaSandru27: Of course! Thank you!

pierricgimmig (approved):
Thanks for the changes Maria!

PR #1294, opened 2020-09-25
Hide CaptureWindow's Capture-Stats behind a flag

beckerhe:
Pressing `I` or `i` in the capture window brings up a bunch of debug windows
which are very useful when making changes to the capture window, but not at all
to the users.

We should avoid situations where the user accidentally presses `i` and these windows show up.

This PR is one quickly-made possible solution. Please push back if you prefer a different way.

In this PR I tied the availability of the `i`-hotkey to Orbit's developer mode. Meaning pressing `i` only
brings up the windows when Orbit was started in developer mode.

Starting Orbit in developer mode works with the `--devmode` command line option. When working
on the capture window always specfying `--devmode` might be annoying. For this situation I added a
environment variable (`ORBIT_DEV_MODE` or alternatively `ORBIT_DEVELOPER_MODE`) which can
be set globally or in the IDE and always enables the developer mode.

### DIFF
--- a/OrbitGl/CaptureWindow.cpp
+++ b/OrbitGl/CaptureWindow.cpp
@@ -10,7 +10,8 @@
 
 using orbit_client_protos::TimerInfo;
 
-CaptureWindow::CaptureWindow() : GlCanvas() {
+CaptureWindow::CaptureWindow(CaptureWindow::StatsMode stats_mode)
+    : GlCanvas(), stats_enabled_(stats_mode == StatsMode::kEnabled) {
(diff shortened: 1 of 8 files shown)


pierricgimmig (approved):
Yesss!! Thanks for this Henning, it's something I've been wanting to do for a while. LGTM, cheers!

reichlfl (approved):
I think this is a valid solution, LGTM!

beckerhe:
Great. Thanks for looking at this!

Issue #1295, opened 2020-09-27
Build fails on Ubuntu 18 - gcc8


pfeatherstone:
I get the following error:
```
ERROR: conanfile.py (OrbitProfiler/None): Error in build() method, line 164
	cmake.build()
	ConanException: Error 2 while executing cmake --build '/home/pf/Downloads/repos/orbit/build_default_relwithdebinfo' '--' '-j12'
```

pfeatherstone:
```
/home/pf/Downloads/repos/orbit/OrbitService/Utils.cpp:188:15: error: ‘transform_reduce’ is not a member of ‘std’
   return std::transform_reduce(begin(proc), end(proc), std::vector<pid_t>{},
```

pfeatherstone:
I think CMakeLists.txt needs to explicitly set C++17 in CFLAGS somewhere

pfeatherstone:
Don't understand, CMakeLists.txt sets c++17 as the minimum standard.

pfeatherstone:
But OrbitService's CMakeLists.txt file doesn't

pfeatherstone:
What was the last release that built on Ubuntu 18  with either gcc or clang?

rodolfocangiotti:
Hey everybody,
I am facing the same issue while trying to build the code on Debian 10. If it could be useful, the issue persists using the default Conan profile both with `compiler.libcxx=libstdc++` and `compiler.libcxx=libstdc++11`.

beckerhe:
Hello everybody,
thanks for reporting this and sorry for the delay.

I hope this is fixed now.

Cheers,
Henning

PR #1296, opened 2020-09-28
Fix GCC-8 build

beckerhe:
Two small code-fixes to make the GCC-8 build work again.

Fixes: #1295

### DIFF
--- a/OrbitQt/TutorialOverlay.h
+++ b/OrbitQt/TutorialOverlay.h
@@ -8,6 +8,7 @@
 #include <QDialog>
 #include <QLabel>
 #include <QPoint>
+#include <functional>
 #include <memory>
 #include <optional>
 #include <unordered_map>
--- a/OrbitService/Utils.cpp
+++ b/OrbitService/Utils.cpp
@@ -176,17 +176,16 @@ static std::optional<pid_t> ProcEntryToPid(const std::filesystem::directory_entr
 }
 
 std::vector<pid_t> GetAllPids() {
-  const auto emplace_back_if_has_value = [](auto vec, const auto& val) {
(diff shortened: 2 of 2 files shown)


OrbitQt/TutorialOverlay.h
```diff
@@ -10,6 +10,7 @@
 #include <QPoint>
 #include <memory>
 #include <optional>
+#include <functional>
```
dpallotti, line 13: clang-format I guess.

OrbitService/Utils.cpp
```diff
@@ -176,17 +176,16 @@ static std::optional<pid_t> ProcEntryToPid(const std::filesystem::directory_entr
...
-    if (val.has_value()) {
-      vec.emplace_back(std::move(val.value()));
-    }
+  fs::directory_iterator proc{"/proc"};
+  std::vector<pid_t> pids;
 
-    return std::move(vec);
-  };
+  for (const auto& entry : proc) {
+    if (auto pid = ProcEntryToPid(entry); pid) {
```
dpallotti, line 183: As you know well by now, I liked the use of `has_value` in the original code! 😄

dpallotti (approved):

OrbitQt/TutorialOverlay.h
beckerhe: Yessss! I should go back to primary school and learn the alphabet. Thanks, fixed!

OrbitService/Utils.cpp
beckerhe: Sure. Fine with me. :+1: I added the `has_value`. Thanks!

PR #1298, opened 2020-09-28
Add base lockfiles to build_and_upload_dependencies.sh

beckerhe:
This makes use of base lockfiles in our dependency build script.

Furthermore this fixes a regression in `build.*` which used to build dependencies when not available as binary prebuilts. This is not relevant for the team, but for external contributors using the public remotes which don't have all the prebuilts.

### DIFF
--- a/third_party/conan/configs/linux/profiles/clang7_debug
+++ b/third_party/conan/configs/linux/profiles/clang7_debug
@@ -15,7 +15,7 @@ compiler.fpo=False
 build_type=Debug
 [options]
 [build_requires]
-cmake_installer/3.16.3@conan/stable
+&: cmake_installer/3.16.3@conan/stable
 [env]
 CC=$C_COMPILER
 CXX=$CXX_COMPILER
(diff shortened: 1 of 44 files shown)


build.ps1
```diff
@@ -49,6 +49,7 @@ foreach ($profile in $profiles) {
   New-Item -ItemType Directory -Force -Path ".\build_$profile\" | Out-Null
 
   & $conan.Path lock create "$PSScriptRoot\conanfile.py" --user=orbitdeps --channel=stable `
+    --build outdated `
```
IrinaShkviro, line 52: Maybe we could you
```suggestion
    --build=outdated `
```
as below for consistency? (and the same in `build.sh`)

IrinaShkviro (approved):
lgtm, thanks!

build.ps1
beckerhe: Good point. Will fix that!

beckerhe:
So this was more complicated as expected.

The "cmake_installer not in lockfile" error might be a bug in conan or we do something wrong.

But it seems that we can fix it by applying a change I wanted to do for a long time.

The commit messages have more information on the details

@IrinaShkviro PTAL
@danielfenner Adding you as a reviewer.

danielfenner (approved):
Thanks, lgtm.

IrinaShkviro (approved):

PR #1300, opened 2020-09-28
Introduce frame tracks to visualize frame markers

ronaldfw:
Users can now choose functions to use as frame markers. A good
example for a frame marker is vkQueuePresent, but any function that
meaningfully represents a "frame" can be used. Choosing multiple frame
markers is possible and meaninful for games that have subsystems
running at different frequencies (e.g. a physics system running at 30Hz,
while rendering runs at 60Hz).

The following improvements are planned for later (not done in this PR):
- Frame tracks should be stored in presets and captures.
- Frame tracks should persist and be populated when starting a new capture.
- Frame tracks could be used to change track backgrounds to indicate
frame boundaries on any track.
- Manual instrumentation markers as frame markers should be supported.
- GPU events as frame markers should be supported.

### DIFF
--- a/OrbitGl/FrameTrack.cpp
+++ b/OrbitGl/FrameTrack.cpp
@@ -1,0 +1,240 @@
+// Copyright (c) 2020 The Orbit Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#include "FrameTrack.h"
+
+#include "App.h"
+#include "OrbitClientData/FunctionUtils.h"
+#include "TimeGraph.h"
+#include "TimeGraphLayout.h"
+
+using orbit_client_protos::FunctionInfo;
+using orbit_client_protos::TimerInfo;
+
+namespace {
+constexpr const double kHeightCapAverageMultipleDouble = 6.0;
+constexpr const uint64_t kHeightCapAverageMultipleUint64 = 6;
+constexpr const float kBoxHeightMultiplier = 3.f;
+}  // namespace
+
+float FrameTrack::GetMaximumBoxHeight() const {
+  const bool is_collapsed = collapse_toggle_->IsCollapsed();
+  const float box_height_normalizer = is_collapsed ? maximum_box_ratio_ : 1.f;
+  if (maximum_box_ratio_ == 0.f) {
+    return 0.f;
+  } else {
+    return maximum_box_ratio_ * box_height_ / box_height_normalizer;
+  }
+}
+
+float FrameTrack::GetAverageBoxHeight() const {
+  const bool is_collapsed = collapse_toggle_->IsCollapsed();
+  const float box_height_normalizer = is_collapsed ? maximum_box_ratio_ : 1.f;
+  if (maximum_box_ratio_ == 0.f) {
+    return 0.f;
+  } else {
+    return box_height_ / box_height_normalizer;
+  }
+}
+
+FrameTrack::FrameTrack(TimeGraph* time_graph, const FunctionInfo& function)
+    : TimerTrack(time_graph), function_(function) {
+  // TODO(b/169554463): Support manual instrumentation.
+  std::string function_name = FunctionUtils::GetDisplayName(function_);
+  std::string name = absl::StrFormat("Frame track based on %s", function_name);
+  SetName(name);
+  SetLabel(name);
+
+  // Frame tracks are collapsed by default.
+  collapse_toggle_->SetState(TriangleToggle::State::kCollapsed,
+                             TriangleToggle::InitialStateUpdate::kReplaceInitialState);
+}
+
+float FrameTrack::GetHeaderHeight() const { return 0.f; }
+
+float FrameTrack::GetHeight() const {
+  TimeGraphLayout& layout = time_graph_->GetLayout();
+  return GetMaximumBoxHeight() + layout.GetTrackBottomMargin();
+}
+
+float FrameTrack::GetYFromDepth(uint32_t /*depth*/) const {
+  return pos_[1] - GetMaximumBoxHeight();
+}
+
+float FrameTrack::GetTextBoxHeight(const TimerInfo& timer_info) const {
+  uint64_t timer_duration_ns = timer_info.end() - timer_info.start();
+  if (stats_.average_time_ns() == 0) {
+    return 0.f;
+  }
+  double ratio =
+      static_cast<double>(timer_duration_ns) / static_cast<double>(stats_.average_time_ns());
+  ratio = std::min(ratio, kHeightCapAverageMultipleDouble);
+  return static_cast<float>(ratio) * GetAverageBoxHeight();
+}
+
+Color FrameTrack::GetTimerColor(const orbit_client_protos::TimerInfo& timer_info,
+                                bool /*is_selected*/) const {
+  Vec4 min_color(76.f, 175.f, 80.f, 255.f);
+  Vec4 max_color(63.f, 81.f, 181.f, 255.f);
+  Vec4 warn_color(244.f, 67.f, 54.f, 255.f);
+
+  Vec4 color;
+  uint64_t timer_duration_ns = timer_info.end() - timer_info.start();
+
+  if (timer_duration_ns >= kHeightCapAverageMultipleUint64 * stats_.average_time_ns()) {
+    color = warn_color;
+  } else {
+    uint64_t lower_bound = stats_.min_ns();
+    uint64_t upper_bound =
+        std::min(kHeightCapAverageMultipleUint64 * stats_.average_time_ns(), stats_.max_ns());
+    if (upper_bound == lower_bound) {
+      // This implies that min_ns == max_ns and thus all times are the same. We can just render
+      // everything using min_color.
+      color = min_color;
+    }
+    timer_duration_ns = std::min(timer_duration_ns, upper_bound);
+    float fraction = static_cast<float>(timer_duration_ns - lower_bound) /
+                     static_cast<float>(upper_bound - lower_bound);
+    color = fraction * max_color + (1.f - fraction) * min_color;
+  }
+
+  if (timer_info.user_data_key() % 2 == 0) {
+    color = 0.8f * color;
+  }
+  return Color(static_cast<uint8_t>(color[0]), static_cast<uint8_t>(color[1]),
+               static_cast<uint8_t>(color[2]), static_cast<uint8_t>(color[3]));
+}
+
+void FrameTrack::OnTimer(const TimerInfo& timer_info) {
+  uint64_t duration_ns = timer_info.end() - timer_info.start();
+  stats_.set_count(stats_.count() + 1);
+  stats_.set_total_time_ns(stats_.total_time_ns() + duration_ns);
+  stats_.set_average_time_ns(stats_.total_time_ns() / stats_.count());
+
+  if (duration_ns > stats_.max_ns()) {
+    stats_.set_max_ns(duration_ns);
+  }
+  if (stats_.min_ns() == 0 || duration_ns < stats_.min_ns()) {
+    stats_.set_min_ns(duration_ns);
+  }
+
+  double ratio = 0.0;
+  if (stats_.average_time_ns() != 0) {
+    ratio = static_cast<double>(duration_ns) / static_cast<double>(stats_.average_time_ns());
+  }
+
+  maximum_box_ratio_ = std::max(maximum_box_ratio_, static_cast<float>(ratio));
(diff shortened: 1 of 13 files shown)


OrbitGl/FrameTrack.cpp
```diff
@@ -0,0 +1,228 @@
...
+void FrameTrack::OnTimer(const TimerInfo& timer_info) {
+  uint64_t duration_ns = timer_info.end() - timer_info.start();
+  stats_.set_count(stats_.count() + 1);
+  stats_.set_total_time_ns(stats_.total_time_ns() + duration_ns);
+  stats_.set_average_time_ns(stats_.total_time_ns() / stats_.count());
+
+  if (duration_ns > stats_.max_ns()) {
+    stats_.set_max_ns(duration_ns);
+  }
+  if (stats_.min_ns() == 0 || duration_ns < stats_.min_ns()) {
```
pierricgimmig, line 107: This assumes that a duration is always >= 1ns, which I think is the case in practice, but maybe we should put in a check to be sure? Or possibly initialize stats_min_ns() to std::numeric_limits<uint64_t>::max()?

OrbitGl/FrameTrack.h
```diff
@@ -0,0 +1,46 @@
...
+
+#ifndef ORBIT_GL_DELTA_TRACK_H_
+#define ORBIT_GL_DELTA_TRACK_H_
+
+#include "TimerTrack.h"
+
+class FrameTrack : public TimerTrack {
+ public:
+  explicit FrameTrack(TimeGraph* time_graph, const orbit_client_protos::FunctionInfo& function);
+  [[nodiscard]] Type GetType() const override { return kDeltaTrack; }
```
pierricgimmig, line 13: Should kDeltaTrack be named kFrameTrack instead?

OrbitGl/LiveFunctionsController.cpp
```diff
@@ -207,6 +208,50 @@ void LiveFunctionsController::AddIterator(FunctionInfo* function) {
...
+      }
+    }
+  }
+  std::sort(all_start_times.begin(), all_start_times.end());
+
+  for (size_t k = 0; k < all_start_times.size() - 1; ++k) {
+    TimerInfo frame_timer;
+
+    // TID is meaningless for this timer (start and end can be on two different threads).
+    constexpr const int32_t kUnusedThreadId = -1;
```
pierricgimmig, line 236: This might interfere with kAllThreadsFakeTid which is also -1.

OrbitGl/LiveFunctionsController.cpp
```diff
@@ -207,6 +208,50 @@ void LiveFunctionsController::AddIterator(FunctionInfo* function) {
...
+    frame_timer.set_thread_id(kUnusedThreadId);
+    frame_timer.set_start(all_start_times[k]);
+    frame_timer.set_end(all_start_times[k + 1]);
+    // Depth is used to indicate i the timer is at an even or odd position in the sequence
+    // of timers. This influences the coloring of the timer. While this depends on other
+    // data (order of timers here), for rendering this information needs to be
+    // available locally for the timer and hence is stored in the timer data structure.
+    frame_timer.set_depth(k % 2);
```
pierricgimmig, line 244: Suggestion: Maybe we could use the user_data field instead of "depth" and store the "index" (k) of the timer? In the tooltip, we could then add "this is the k-th frame". Then we could also use the index to choose the draw color.

pierricgimmig (approved):
Hi Ronald, this is really cool! Thanks! I tested your change and it seems to work as expected. I added several minor comments, but in general LGTM. Cheers!

OrbitGl/FrameTrack.cpp
```diff
@@ -0,0 +1,228 @@
...
+
+float FrameTrack::GetAverageBoxHeight() const {
+  const bool is_collapsed = collapse_toggle_->IsCollapsed();
+  const float box_height_normalizer = is_collapsed ? maximum_box_ratio_ : 1.f;
+  return box_height_ / box_height_normalizer;
+}
+
+FrameTrack::FrameTrack(TimeGraph* time_graph, const FunctionInfo& function)
+    : TimerTrack(time_graph), function_(function) {
+  // TODO(wotzlaw): Support the case for manual instrumentation.
```
reichlfl, line 36: Could you please add a bug-link? :-)

OrbitGl/FrameTrack.cpp
```diff
@@ -0,0 +1,228 @@
...
+      GetPrettyTime(absl::Nanoseconds(stats_.min_ns())),
+      GetPrettyTime(absl::Nanoseconds(stats_.average_time_ns())));
+}
+
+std::string FrameTrack::GetBoxTooltip(PickingId id) const {
+  const TextBox* text_box = time_graph_->GetBatcher().GetTextBox(id);
+  if (!text_box) {
+    return "";
+  }
+  // TODO(wotzlaw): Support case for manual instrumentation.
```
reichlfl, line 171: Buglink

OrbitGl/FrameTrack.h
```diff
@@ -0,0 +1,46 @@
...
+  void UpdateBoxHeight() override;
+
+ protected:
+  [[nodiscard]] Color GetTimerColor(const orbit_client_protos::TimerInfo& timer_info,
+                                    bool is_selected) const override;
+  [[nodiscard]] float GetHeight() const override;
+
+ private:
+  float GetMaximumBoxHeight() const;
+  float GetAverageBoxHeight() const;
```
reichlfl, line 38: [[nodiscard]] for both above

reichlfl (approved):
Very cool! Small comments on top of Pierric's, otherwise LGTM.

OrbitGl/FrameTrack.cpp
```diff
@@ -0,0 +1,228 @@
...
+  std::string label = absl::StrFormat("Avg: %s", avg_time);
+
+  int string_height_unused = 0;
+  int string_width = 0;
+  canvas->GetTextRenderer().GetStringSize(label.c_str(), string_width, string_height_unused);
+
+  // It seems the size of the box that contains the text needs to be a little larger than the
+  // text size itself. Not adding the 1.f on the line results in the last character of the string
+  // not getting rendered.
+  Vec2 white_text_box_size(static_cast<float>(string_width) + 1.f, layout.GetTextBoxHeight());
```
karupayun, line 213: I prefer to add this in GetStringSize. Let me add that, as my PR is almost mergeable.

karupayun (approved):
Very nice PR Ronald! I really liked it and I feel that this Frame Tracks will be very useful! :)

In particular, I liked the color palette. Although doesn't work very well with functions like vkQueueSubmit, with typical frame-kind functions (like vkQueuePresent) provides very nice visual information.

OrbitGl/FrameTrack.cpp
```diff
@@ -0,0 +1,228 @@
+// Copyright (c) 2020 The Orbit Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#include "FrameTrack.h"
+
+#include "App.h"
+#include "ManualInstrumentationManager.h"
```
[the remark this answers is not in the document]
ronaldfw: Done.

OrbitGl/FrameTrack.cpp
```diff
@@ -0,0 +1,228 @@
...
+float FrameTrack::GetMaximumBoxHeight() const {
+  const bool is_collapsed = collapse_toggle_->IsCollapsed();
+  const float box_height_normalizer = is_collapsed ? maximum_box_ratio_ : 1.f;
+  return maximum_box_ratio_ * box_height_ / box_height_normalizer;
+}
+
+float FrameTrack::GetAverageBoxHeight() const {
+  const bool is_collapsed = collapse_toggle_->IsCollapsed();
+  const float box_height_normalizer = is_collapsed ? maximum_box_ratio_ : 1.f;
+  return box_height_ / box_height_normalizer;
```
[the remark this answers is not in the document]
ronaldfw: Good point, under "normal" circumstances this cannot happen. As soon as you have one timer with a positive duration, all stats are non-zero, which guarantees that this value is non-zero. However, to be a bit more defensive I've added guards for all cases where I divide by a value and return or use the appropriate value in that case. If all timers had zero duration, they would now correctly be drawn with zero height.

OrbitGl/FrameTrack.cpp
```diff
@@ -0,0 +1,228 @@
...
+
+FrameTrack::FrameTrack(TimeGraph* time_graph, const FunctionInfo& function)
+    : TimerTrack(time_graph), function_(function) {
+  // TODO(wotzlaw): Support the case for manual instrumentation.
+  std::string function_name = FunctionUtils::GetDisplayName(function_);
+  std::string name = absl::StrFormat("Frame track based on %s", function_name);
+  SetName(name);
+  SetLabel(name);
+
+  // Delta tracks are collapsed by default.
```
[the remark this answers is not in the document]
ronaldfw: Thanks, one part I missed when renaming.

OrbitGl/FrameTrack.cpp
```diff
@@ -0,0 +1,228 @@
...
+  }
+
+  if (timer_info.depth() == 0) {
+    return Color(static_cast<uint8_t>(color[0]), static_cast<uint8_t>(color[1]),
+                 static_cast<uint8_t>(color[2]), static_cast<uint8_t>(color[3]));
+  } else {
+    color = 0.8f * color;
+    return Color(static_cast<uint8_t>(color[0]), static_cast<uint8_t>(color[1]),
+                 static_cast<uint8_t>(color[2]), static_cast<uint8_t>(color[3]));
+  }
```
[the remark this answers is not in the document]
ronaldfw: Good point. Fixed.

OrbitGl/FrameTrack.h
ronaldfw: Yes, forgot the rename this.

OrbitGl/FrameTrack.cpp
ronaldfw: But doesn't it still work correctly, though? If duration_ns is 0 for at least one timer, then we'll just set the min to 0 again in the if body.

OrbitGl/LiveFunctionsController.cpp
ronaldfw: That's a great suggestion, will do that.

OrbitGl/FrameTrack.cpp
ronaldfw: Done.

OrbitGl/FrameTrack.cpp
ronaldfw: Done.

OrbitGl/LiveFunctionsController.cpp
ronaldfw: Done.

OrbitGl/FrameTrack.h
ronaldfw: Done.

OrbitGl/FrameTrack.cpp
ronaldfw: Fixed to use GetStringWidth().

OrbitGl/LiveFunctionsController.cpp
ronaldfw: Yes, I thought about that.

My thinking was that:
1) There's not much I can do about it. I could use -3 here (as -2 is also already used), but there's no guarantee someone will not introduce -3 in some other context. 
2) It doesn't matter too much, as the timers are distinguished first based on type. We have other places where tids are used in a way that if timers would be added to the wrong track, they would mess up stuff. For example, the GPU job timers all have the tid of the submitting track. But since they are used only to populate the GPU track, this is fine. 

Let me know if you still have concerns and we can discuss if there's a better approach here.

OrbitGl/FrameTrack.cpp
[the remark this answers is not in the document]
pierricgimmig: cool!

OrbitGl/FrameTrack.cpp
pierricgimmig: Yes you're right, nevermind :-)

OrbitGl/LiveFunctionsController.cpp
pierricgimmig: Yeah it's true that the dispatching of timers will occur based on the timer type, so this should not be a problem. I think it's fine.

pierricgimmig (approved):
Thanks for the changes Ronald, LGTM!

PR #1301, opened 2020-09-28
Store ProcessData by Value (Not as Unique Pointer)

florian-kuebler:
ProcessData may be stored at two places, first in DataManager and
second in CaptureData. There is no need to store them as unique
pointers here. Moreover, the absense of a concrete process in
CaptureData already led to a crash.

This change ensures, that we always store the process by value,
which among others allow clear default values for empty captures.

Test: Compile

### DIFF
--- a/OrbitCaptureClient/CaptureClient.cpp
+++ b/OrbitCaptureClient/CaptureClient.cpp
@@ -50,7 +50,7 @@ ErrorMessageOr<void> CaptureClient::StartCapture(
 
   state_ = State::kStarting;
 
-  ProcessData process_copy = process.CreateCopy();
+  ProcessData process_copy = process;
 
   // TODO(168797897) Here a copy of the module_map is created. This module_map likely was downloaded
   // when the process was selected, which might be a while back. Between then and now the loaded
--- a/OrbitCaptureClient/CaptureClient.cpp
+++ b/OrbitCaptureClient/CaptureClient.cpp
@@ -50,7 +50,7 @@ ErrorMessageOr<void> CaptureClient::StartCapture(
 
   state_ = State::kStarting;
 
-  std::unique_ptr<ProcessData> process_copy = process.CreateCopy();
+  ProcessData process_copy = process.CreateCopy();
 
   // TODO(168797897) Here a copy of the module_map is created. This module_map likely was downloaded
   // when the process was selected, which might be a while back. Between then and now the loaded
@@ -68,7 +68,7 @@ ErrorMessageOr<void> CaptureClient::StartCapture(
   return outcome::success();
 }
 
-void CaptureClient::Capture(std::unique_ptr<ProcessData> process,
+void CaptureClient::Capture(ProcessData&& process,
                             absl::flat_hash_map<std::string, ModuleData*>&& module_map,
                             absl::flat_hash_map<uint64_t, FunctionInfo> selected_functions,
                             TracepointInfoSet selected_tracepoints) {
@@ -80,7 +80,7 @@ void CaptureClient::Capture(std::unique_ptr<ProcessData> process,
   CaptureRequest request;
   CaptureOptions* capture_options = request.mutable_capture_options();
   capture_options->set_trace_context_switches(true);
-  capture_options->set_pid(process->pid());
+  capture_options->set_pid(process.pid());
   uint16_t sampling_rate = absl::GetFlag(FLAGS_sampling_rate);
   if (sampling_rate == 0) {
     capture_options->set_unwinding_method(CaptureOptions::kUndefined);
--- a/OrbitClientData/ProcessData.cpp
+++ b/OrbitClientData/ProcessData.cpp
@@ -65,30 +65,3 @@ void ProcessData::AddSymbols(ModuleData* module, const ModuleSymbols& module_sym
   uint64_t module_base_address = module_memory_map_.at(module->file_path()).start;
   module->AddSymbols(module_symbols, module_base_address);
 }
-
-ProcessData ProcessData::CreateCopy() const {
-  ProcessInfo info;
-  info.set_pid(pid());
-  info.set_name(name());
-  info.set_full_path(full_path());
-  info.set_command_line(command_line());
-  info.set_cpu_usage(cpu_usage());
-  info.set_is_64_bit(is_64_bit());
-  ProcessData process_copy(info);
-
-  for (const auto& [module_path, memory_space] : module_memory_map_) {
-    uint64_t start = memory_space.start;
-    uint64_t end = memory_space.end;
-    {
-      const auto [it, success] =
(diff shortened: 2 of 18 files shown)


OrbitClientData/include/OrbitClientData/ProcessData.h
```diff
@@ -67,8 +68,8 @@ class ProcessData final {
   bool IsModuleLoaded(const std::string& module_path) const {
     return module_memory_map_.contains(module_path);
   }
-  static std::unique_ptr<ProcessData> Create(orbit_grpc_protos::ProcessInfo process_info);
-  std::unique_ptr<ProcessData> CreateCopy() const;
+
+  ProcessData CreateCopy() const;
```
dpallotti, line 72: Deleting the copy constructor and copy assignment (that could just be default I think, as all the members are copyable) to then provide a dedicated method for the copy seems quite wrong.

florian-kuebler: @antonrohr made this to ensure that there are no unintended copies. I am not sure, how relevant this with the recent changes and happy to remove this, if needed.

antonrohr: Florian summarized it well, but I am happy to be convinced otherwise. 
ProcessData is passed around quite a bit and used as raw pointer in a lot of places. It is easy to accidentally make a copy when forgetting to use std::move.

OrbitClientGgp/ClientGgp.h
```diff
@@ -30,8 +30,7 @@ class ClientGgp final : public CaptureListener {
 
   // CaptureListener implementation
   void OnCaptureStarted(
-      std::unique_ptr<ProcessData> process,
-      absl::flat_hash_map<std::string, ModuleData*>&& module_map,
+      ProcessData process, absl::flat_hash_map<std::string, ModuleData*>&& module_map,
```
antonrohr, line 33:
```suggestion
      ProcessData&& process, absl::flat_hash_map<std::string, ModuleData*>&& module_map,
```

OrbitClientModel/CaptureDeserializerTest.cpp
```diff
@@ -163,13 +162,13 @@ TEST(CaptureDeserializer, LoadCaptureInfoOnCaptureStarted) {
   EXPECT_CALL(listener, OnCaptureStarted(_, _, _, _)).Times(0);
   EXPECT_CALL(listener, OnCaptureStarted(_, _, _, IsEmpty()))
       .Times(1)
-      .WillOnce([selected_function](std::unique_ptr<ProcessData> process, Unused,
+      .WillOnce([selected_function](ProcessData&& process, Unused,
                                     absl::flat_hash_map<uint64_t, orbit_client_protos::FunctionInfo>
                                         actual_selected_functions,
                                     Unused) {
         ASSERT_NE(process, nullptr);
```
antonrohr, line 169:
```suggestion
```

antonrohr (changes requested):
There are still some compile errors. I am going to continue reviewing when they are resolved

OrbitClientData/include/OrbitClientData/ProcessData.h
dpallotti: As the copy doesn't have any side effect, I don't see the problem of using the copy constructor and assignment. That's what they are for. The fact that it's used as raw pointer in many places doesn't have a connection with this. Using a dedicated method to prevent accidental copies seems a very weird pattern in C++. Based on this reasoning all containers (e.g., even `vector`s) should not have a copy constructor to prevent accidental copies.

florian-kuebler:
Thanks for the review so far. I fixed the compile errors.

OrbitClientData/include/OrbitClientData/ProcessData.h
antonrohr: Okay. I guess the he explicit `ProcessData&&` that is now used in method signatures helps with enforcing move instead of copy. Lets remove the Create and use the default copy constructor/assingment

antonrohr (approved):
lgtm, only remove the CreateCopy() method for a copy constructor. See comment

florian-kuebler:
> lgtm, only remove the CreateCopy() method for a copy constructor. See comment

Thanks for the review. I removed that function.

OrbitCaptureClient/CaptureClient.cpp
```diff
@@ -50,7 +50,7 @@ ErrorMessageOr<void> CaptureClient::StartCapture(
 
   state_ = State::kStarting;
 
-  std::unique_ptr<ProcessData> process_copy = process.CreateCopy();
```
dpallotti, line 53: You could just have the lambda capture `process` by value, instead of creating a copy and the moving the copy. Same goes with `module_map` and `selected_functions`.

OrbitClientData/ProcessData.cpp
```diff
@@ -11,10 +11,11 @@
 #include "absl/strings/str_format.h"
 #include "process.pb.h"
 
-using orbit_client_protos::FunctionInfo;
 using orbit_grpc_protos::ModuleSymbols;
 using orbit_grpc_protos::ProcessInfo;
 
+ProcessData::ProcessData() { process_info_.set_pid(-1); }
```
dpallotti, line 17: Do we have a constant somewhere for invalid pid instead of using -1?

dpallotti (approved):

OrbitCaptureClient/CaptureClient.cpp
florian-kuebler: As, discussed offline, this does not compile as the process is passed as ref and capture will be the ref.
I will keep it as is.

OrbitClientData/ProcessData.cpp
florian-kuebler: No, unfortunately not.

OrbitCaptureClient/CaptureClient.cpp
dpallotti: So yeah, just for your information, I had a quick look and&mdash;more or less what you said&mdash; the problem is that `process` is passed by const reference (which is how it's captured, I didn't know that), *and* that `Capture` requires a (non-const, obviously) r-value as first parameter. You can't make a const-ref an r-value. What one could do to be more concise it to make the copy right when passing the parameter, like
```
  thread_pool->Schedule(
      [this, process, module_map = std::move(module_map_copy),
       selected_functions = std::move(selected_functions), selected_tracepoints]() mutable {
        Capture(ProcessData{process}, std::move(module_map), std::move(selected_functions),
                std::move(selected_tracepoints));
      });
```
which saves one move, the one into the lambda.

florian-kuebler: Would be fine for me. If you like, you can just create a small PR for that.
Thanks for looking into this.

PR #1302, opened 2020-09-28
Fix nightly and release builds

beckerhe:
The new version of conan logs all options to the standard output, which
was not the case before. One of the options logged is `crashdump_server`
which is filled from a secret from Keystore even though it's not so
secret.

Kokoro has a script which aborts the build when something logs a
keystore-secret and that breaks the build.

This commit fixes that be adjusting conan's output with sed.

Test: I created a nightly build from my branch and it seems to work. The output now looks like so:

```
...
[options]
crashump_server=<<hidden>>
OrbitProfiler:system_qt=False
...
```


danielfenner (approved):
Thanks for fixing this! lgtm!

PR #1303, opened 2020-09-28
Collect thread names for all threads in the system

dpallotti:
#### Move GetAllPids from OrbitService/Utils to OrbitLinuxTracing/Utils
This is currently unused. Will be used in the upcoming commits.
#### Initial snapshot of thread names now includes all threads in the system
Bug: http://b/169569531

Test: Take a capture locally and a capture on Trata. Verify that several
scheduling slices not from the target process have an associated thread name in
the scheduler track's tooltips.
#### Rename Utils files to {Core,LinuxTracing,Service}Utils
Despite the different modules, the homonymy of these files made it sometimes
annoying to deal with them.
#### Add license header to CoreUtilsTest.cpp


OrbitLinuxTracing/LinuxTracingUtils.cpp
```diff
@@ -35,40 +37,60 @@ std::string ReadMaps(pid_t pid) {
...
   }
+  return "";
 }
 
-std::optional<std::string> ExecuteCommand(const std::string& cmd) {
-  std::unique_ptr<FILE, decltype(&pclose)> pipe{popen(cmd.c_str(), "r"), pclose};
-  if (!pipe) {
-    ERROR("Could not open pipe for \"%s\"", cmd.c_str());
-    return std::optional<std::string>{};
+static std::optional<pid_t> ProcEntryToPid(const std::filesystem::directory_entry& entry) {
```
florian-kuebler, line 44: where is this pid_t coming from? Ad the client, we switched to int32_t for process/thread id. Does this also make sense to do so here?

OrbitLinuxTracing/LinuxTracingUtils.cpp
```diff
@@ -35,40 +37,60 @@ std::string ReadMaps(pid_t pid) {
...
-  std::vector<pid_t> threads;
-  std::optional<std::string> tasks = ExecuteCommand(absl::StrFormat("ls /proc/%d/task", pid));
-  if (!tasks.has_value()) {
-    return {};
+std::vector<pid_t> GetAllPids() {
+  fs::directory_iterator proc{"/proc"};
+  std::vector<pid_t> pids;
+
+  for (const auto& entry : proc) {
+    if (auto pid = ProcEntryToPid(entry); pid.has_value()) {
```
florian-kuebler, line 66: Is there a reason to not move the declaration/assignment of pid out of the condition. This makes it harder to read.

OrbitLinuxTracing/LinuxTracingUtils.cpp
```diff
@@ -35,40 +37,60 @@ std::string ReadMaps(pid_t pid) {
...
-    threads.push_back(std::stol(line));
+  return pids;
+}
+
+std::vector<pid_t> GetTidsOfProcess(pid_t pid) {
+  fs::directory_iterator proc_pid_task{fs::path{"/proc"} / std::to_string(pid) / "task"};
+  std::vector<pid_t> tids;
+
+  for (const auto& entry : proc_pid_task) {
+    if (auto tid = ProcEntryToPid(entry); tid.has_value()) {
```
florian-kuebler, line 79: See above, can you move tid out of the if?

florian-kuebler (approved):
Despite the minor comments, LGTM! Thanks!

OrbitLinuxTracing/LinuxTracingUtils.cpp
dpallotti: This is a nice C++17 feature. The point of it is to limit the scope of a variable used only in the if to only the if. In other reviews I've actually been told to use the feature, as I tend to forget about it even though I like it (this was actually Henning's code, that's why I "remembered" this time).

OrbitLinuxTracing/LinuxTracingUtils.cpp
florian-kuebler: Okay, I see. Question would be if we might use int32_t anyways, to make it consistent across orbit. If yes, for sure in a different PR.

OrbitLinuxTracing/LinuxTracingUtils.cpp
florian-kuebler: Okay, personally I think it is harder to read, and the effective scope in those cases here, will be exactly the same, as the outer scope (of the loop) also ends with directly after the one of the if.
Anyways, feel free to decide whatever you like more.

And out of curiosity, is that really a C++17 feature? I had the feeling that its been there for years in Java, and the adaption rate there, is not very high. So I thought it would be also there since ever in C++. :)

dpallotti: https://en.cppreference.com/w/cpp/language/if
Is this really in Java? I've never seen it.

OrbitLinuxTracing/LinuxTracingUtils.cpp
dpallotti: I personally prefer using the type from the Linux headers in Linux-only code.

OrbitLinuxTracing/LinuxTracingUtils.cpp
ronaldfw: I think this is indeed a great feature that we should all get used to use more often, and it's also consistent with how declarations are done in for loops. 

> Okay, personally I think it is harder to read, and the effective scope in those cases here, will be exactly the same, as the outer scope (of the loop) also ends with directly after the one of the if.

Unless someone changes the code someday and adds something below the if. Better to reduce the scope to what is actually needed now.

florian-kuebler: Thanks for the Link and for the explanation. Personally, I am still not convinced, though, but that doesn't matter for now.
And @dpallotti, you was correct. Java does not have this (at least Java 8), I wonder why I thought so. Pretty sure, I rarely saw this.

OrbitLinuxTracing/LinuxTracingUtils.cpp
florian-kuebler: Okay, fine for me.

PR #1304, opened 2020-09-29
Add flag to activate preset test.

danielfenner:
Bug: 161839506
Test: N/A


beckerhe (approved):
Thanks.

Nit: I think we agreed on listing bugs as URLs, like http://b/161839506 such that you can click on it.

PR #1306, opened 2020-09-29
Fix sorting reset in functions view

antonrohr:
Also Refactor other data view classes to not sort twice.

To fix b//169389587


florian-kuebler (approved):
LGTM. Thanks!

PR #1307, opened 2020-09-29
Store Selected Functions as Values

florian-kuebler:
This is the first part of a series of moving away from representing
functions by absolute addresses in order to make them independent
of the loaded process.

Prior to this, DataManager stored the selected functions as
as set of their absolute addresses.
This changes this, and stores them by value.
As protos have no hash/equals implementation, FunctionInfoSet
was added.

Test: Compile
Bug: http://b/169309553

### DIFF
--- a/OrbitClientData/FunctionInfoSetTest.cpp
+++ b/OrbitClientData/FunctionInfoSetTest.cpp
@@ -1,0 +1,274 @@
+// Copyright (c) 2020 The Orbit Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#include <gtest/gtest.h>
+
+#include "OrbitClientData/FunctionInfoSet.h"
+#include "capture_data.pb.h"
+
+using orbit_client_protos::FunctionInfo;
+
+TEST(FunctionInfoSet, EqualFunctions) {
+  FunctionInfo left;
+  left.set_name("foo");
+  left.set_pretty_name("void foo()");
+  left.set_loaded_module_path("/path/to/module");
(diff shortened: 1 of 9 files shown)


dimitry- (commented):
are there any advantages in storing selected functions as values?

florian-kuebler:
> are there any advantages in storing selected functions as values?

The idea is the following: We currently store them by absolute address. The long-term idea of that series of changes, is to make function/modules independent from process. So we also want to restrict using absolute addresses (there are some places, where we can't and shouldn't get rid of them, but if it makes sense, we should).
I was thinking of storing the functions as pointers (this is how this was done in the past at some point in time),
but we now have different locations, where the function might be owned, so we can't rely on pointers.

OrbitClientData/FunctionInfoSetTest.cpp
```diff
@@ -0,0 +1,274 @@
...
+  left.set_loaded_module_path("/path/to/module");
+  left.set_module_base_address(42);
+  left.set_address(12);
+  left.set_load_bias(4);
+  left.set_size(16);
+  left.set_file("file.cpp");
+  left.set_line(13);
+
+  FunctionInfo right;
+  right.CopyFrom(left);
```
dimitry-, line 25: maybe just copy and paste in this particular case lines 13-22 for right?

dimitry- (approved):
otherwise lgtm

OrbitClientData/FunctionInfoSetTest.cpp
florian-kuebler: sure. Done

florian-kuebler:
Thanks for the review

PR #1308, opened 2020-09-29
Fix Potential Pointer Stability Issue in DataManager

florian-kuebler:
Recent changes stores processes by value, rather then by unique_ptr
in DataManager. As we allow access to the actualy pointer of the
underlying process (in the container), we should be sure to have
pointer stability in that container.

absl::node_hash_map provides this stability.

Test: Compile


beckerhe (approved):
This sounds like a nice catch and good solution to me. But I only little about this part of the code, so I can't really judge if this has implications (performance or other). So Anton or Dimitry should also have a look.

But thanks from my side. :-)

dimitry- (approved):

PR #1309, opened 2020-09-29
Scale text offset

karupayun:
Now that we are scaling text font size, we also have to scale text
offset.


pierricgimmig (approved):
LGTM, thanks!

PR #1310, opened 2020-09-30
Fix crash on exit due to double deletion of font

pierricgimmig:


karupayun (approved):
Thanks @pierricgimmig !

PR #1312, opened 2020-09-30
Add --build=outdated to conan lock commands

beckerhe:
After a series of tests I came to the conclusion that this flag is
needed when creating the full lock-file. Without it conan won't include
build requirements into the lockfile for dependencies that need to be
built locally. And when these are not included the next step (`conan
install`) will fail because it is missing a locked dependency from the
graph.


beckerhe:
@pierricgimmig This should hopefully also resolve the  "cmake_installer not in lockfile"-problem. Sorry that it took so long.

IrinaShkviro (approved):
lgtm!

PR #1313, opened 2020-09-30
Fix manual instrumentation serialization for async tracks

pierricgimmig:
In the case of async time slices, the start and stop events are their
own individual timers and are already serialized on their initial thread
tracks. Make sure we don't serialize the async timers twice.

This change fixes a bug in the serialization of async timers that would
cause duplication of time slices. The other manual instrumentation
data was already properly serialized (scopes and graphs).

Fixes b/165078963

Test: grab a capture of OrbitTest which uses the manual instrumentation
API, save and reload.

### DIFF
--- a/OrbitGl/AsyncTrack.cpp
+++ b/OrbitGl/AsyncTrack.cpp
@@ -52,8 +52,8 @@ void AsyncTrack::UpdateBoxHeight() {
 }
 
 std::vector<std::shared_ptr<TimerChain>> AsyncTrack::GetAllSerializableChains() {
-  // For async time slices, the start and stop events are their own individual timers and are 
-  // already serialized on their initial thread tracks. Return an empty vector so that we don't 
+  // For async time slices, the start and stop events are their own individual timers and are
+  // already serialized on their initial thread tracks. Return an empty vector so that we don't
   // serialize the async timer twice.
   return {};
 }
--- a/OrbitGl/AsyncTrack.cpp
+++ b/OrbitGl/AsyncTrack.cpp
@@ -51,10 +51,10 @@ void AsyncTrack::UpdateBoxHeight() {
   }
 }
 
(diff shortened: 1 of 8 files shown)


florian-kuebler (approved):
Thanks Pierric for the change. It makes sense, even though this feels a bit hacky. However, I do not have a better solution in my mind. So LGTM.

OrbitGl/AsyncTrack.cpp
```diff
@@ -51,6 +51,13 @@ void AsyncTrack::UpdateBoxHeight() {
   }
 }
 
+std::vector<std::shared_ptr<TimerChain>> AsyncTrack::GetAllChains() {
```
ronaldfw, line 54: There's a similar issue with the frame tracks, which also is using data that is already present in other tracks. Do we need a clear concept of "derived tracks" that are never serialized, but are computed on load?

pierricgimmig: I added an explicit GetAllSerializableChains() method that the tracks can override, I think the solution is cleaner now. Essentially, a track now has control on exactly what timers should be serialized.

pierricgimmig:
> Thanks Pierric for the change. It makes sense, even though this feels a bit hacky.

:-) Yeah... I have to agree that it was a little bit hacky. I cleaned up the code by adding an explicit method to get serializable timers.

OrbitGl/AsyncTrack.cpp
ronaldfw: Great, I was going to suggest the same!

ronaldfw (approved):
LGTM!

OrbitGl/AsyncTrack.cpp
florian-kuebler: Yep, that makes it more clear. Thanks!

florian-kuebler (approved):
LGTM, thanks!

PR #1314, opened 2020-09-30
Check std::error_code set by directory_iterator in GetTidsOfProcess

dpallotti:
Bug: http://b/169727599

Test: Automatically run `LinuxTracingUtilsTest` thousands of times, verify that
the times when the problem occurs and the error is printed the test still
succeeds.

### DIFF
--- a/OrbitLinuxTracing/LinuxTracingUtils.cpp
+++ b/OrbitLinuxTracing/LinuxTracingUtils.cpp
@@ -2,9 +2,6 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
 
-#ifndef ORBIT_LINUX_TRACING_UTILS_H_
-#define ORBIT_LINUX_TRACING_UTILS_H_
-
 #include <OrbitBase/Logging.h>
 #include <OrbitBase/SafeStrerror.h>
 #include <sys/resource.h>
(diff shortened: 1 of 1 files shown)


OrbitLinuxTracing/LinuxTracingUtils.cpp
```diff
@@ -72,15 +69,21 @@ std::vector<pid_t> GetAllPids() {
 }
 
 std::vector<pid_t> GetTidsOfProcess(pid_t pid) {
-  fs::directory_iterator proc_pid_task{fs::path{"/proc"} / std::to_string(pid) / "task"};
-  std::vector<pid_t> tids;
+  fs::directory_iterator proc_pid_task;
+  try {
+    proc_pid_task = fs::directory_iterator{fs::path{"/proc"} / std::to_string(pid) / "task"};
```
dimitry-, line 74: maybe use exception-free variant instead?

beckerhe: I was about to write the same. `directory_iterator::directory_iterator` has an overloard which takes a `std::error_code&` which can be checked, instead of throwing an exception. Ref: https://en.cppreference.com/w/cpp/filesystem/directory_iterator/directory_iterator

dpallotti: Ah better, I had missed that, please have a look.

beckerhe (approved):
Thanks! Looks good!

dimitry- (approved):

PR #1315, opened 2020-09-30
Remove Module Base Address and Load Bias from FunctionInfo

florian-kuebler:
Towards making modules (and thus functions) independent from their
process, module base address and module load bias should not be stored
inside FunctionInfo.
This PR removes these fields, in particular by introducing a new GetAbsoluteAddress
function, that uses the process and module data.

It also serializes basic information about the modules and the process.
Loading a capture, will also create the basic information
in data manager.

Symbol loading does not require the process anymore.

Test: Compile, Take Captures, Load Captures, Save Captures
Bug: http://b/169309553

### DIFF
--- a/OrbitClientData/include/OrbitClientData/FunctionInfoSet.h
+++ b/OrbitClientData/include/OrbitClientData/FunctionInfoSet.h
@@ -15,9 +15,7 @@ struct HashFunctionInfo {
     return std::hash<std::string>{}(function.name()) * 37 +
            std::hash<std::string>{}(function.pretty_name()) * 37 +
            std::hash<std::string>{}(function.loaded_module_path()) * 37 +
-           std::hash<uint64_t>{}(function.module_base_address()) * 37 +
            std::hash<uint64_t>{}(function.address()) * 37 +
-           std::hash<uint64_t>{}(function.load_bias()) * 37 +
            std::hash<uint64_t>{}(function.size()) * 37 +
            std::hash<std::string>{}(function.file()) * 37 + std::hash<uint32_t>{}(function.line());
   }
@@ -26,11 +24,9 @@ struct HashFunctionInfo {
 struct EqualFunctionInfo {
   bool operator()(const orbit_client_protos::FunctionInfo& left,
                   const orbit_client_protos::FunctionInfo& right) const {
-    return left.name().compare(right.name()) == 0 &&
+    return left.address() == right.address() && left.name().compare(right.name()) == 0 &&
            left.pretty_name().compare(right.pretty_name()) == 0 &&
            left.loaded_module_path().compare(right.loaded_module_path()) == 0 &&
-           left.module_base_address() == right.module_base_address() &&
-           left.address() == right.address() && left.load_bias() == right.load_bias() &&
            left.size() == right.size() && left.file().compare(right.file()) == 0 &&
            left.line() == right.line();
   }
--- a/OrbitClientData/include/OrbitClientData/FunctionInfoSet.h
+++ b/OrbitClientData/include/OrbitClientData/FunctionInfoSet.h
@@ -5,8 +5,8 @@
 #ifndef ORBIT_CLIENT_DATA_FUNCTION_INFO_SET_H_
 #define ORBIT_CLIENT_DATA_FUNCTION_INFO_SET_H_
 
-#include <absl/container/flat_hash_set.h>
-
+#include "absl/container/flat_hash_map.h"
+#include "absl/container/flat_hash_set.h"
 #include "capture_data.pb.h"
 
 namespace internal {
@@ -40,6 +40,11 @@ struct EqualFunctionInfo {
 
 using FunctionInfoSet =
     absl::flat_hash_set<orbit_client_protos::FunctionInfo, internal::HashFunctionInfo,
+                        internal::EqualFunctionInfo>;
+
+template <class V>
+using FunctionInfoMap =
+    absl::flat_hash_map<orbit_client_protos::FunctionInfo, V, internal::HashFunctionInfo,
                         internal::EqualFunctionInfo>;
 
 #endif  // ORBIT_CLIENT_DATA_FUNCTION_INFO_SET_H_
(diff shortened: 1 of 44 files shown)


OrbitClientData/include/OrbitClientData/FunctionInfoSet.h
```diff
@@ -42,4 +42,9 @@ using FunctionInfoSet =
     absl::flat_hash_set<orbit_client_protos::FunctionInfo, internal::HashFunctionInfo,
                         internal::EqualFunctionInfo>;
 
+template <class V>
+using FunctionInfoMap =
+    absl::flat_hash_map<orbit_client_protos::FunctionInfo, V, internal::HashFunctionInfo,
+                        internal::EqualFunctionInfo>;
+
```
dimitry-, line 49: We need to be careful here - we should probably keep module_path as part of FunctionInfo forever to avoid unwanted conflict.

Also seems like file and line should not be part of the equals/hash<> operators.

dimitry- (approved):
otherwise lgtm

OrbitClientData/include/OrbitClientData/FunctionInfoSet.h
florian-kuebler: Yes, I agree that we should keep module_path as part of function info.
Regarding the equals/hash, I would say to use all the fields that are in the proto. Ideally a proto generated class should have equals/hash to be used in sets/maps, but this will only come in the future.

OrbitClientModel/CaptureSerializer.cpp
```diff
@@ -74,14 +75,26 @@ CaptureInfo GenerateCaptureInfo(
...
+  for (const auto& function_to_stats : functions_stats) {
+    const FunctionInfo& function = function_to_stats.first;
+    const ModuleData* module = capture_data.FindModuleByPath(function.loaded_module_path());
+    CHECK(module != nullptr);
+    const ProcessData* process = capture_data.process();
+    CHECK(process != nullptr);
+    uint64_t absolute_address =
+        FunctionUtils::GetAbsoluteAddress(function_to_stats.first, *process, *module);
+    capture_info.mutable_function_stats()->operator[](absolute_address) = function_to_stats.second;
+  }
```
antonrohr, line 97: Maybe consider having a method: `uint64_t CaptureData::GetAbsoluteAddress(const FunctionInfo&)`. Then `CaptureData::FindModuleByPath` is probably not needed

antonrohr (approved):
This is really cool. A good step into the right direction!

OrbitClientModel/CaptureSerializer.cpp
florian-kuebler: Thanks for the suggestion. Done!

OrbitClientProtos/capture_data.proto
```diff
@@ -73,9 +92,9 @@ message CaptureHeader {
 }
 
 message CaptureInfo {
+  reserved 2;
+  reserved 3;
```
dimitry-, line 96: reserved 2, 3;

florian-kuebler: Done

OrbitClientData/include/OrbitClientData/ProcessData.h
```diff
@@ -69,6 +69,8 @@ class ProcessData final {
     return module_memory_map_.contains(module_path);
   }
 
+  [[nodiscard]] const orbit_grpc_protos::ProcessInfo& process_info() const { return process_info_; }
```
antonrohr, line 72: This is actually not used anywhere

OrbitClientModel/CaptureSerializer.cpp
```diff
@@ -74,14 +93,18 @@ CaptureInfo GenerateCaptureInfo(
...
+    const uint64_t offset = absolute_address - absolute_function_address;
     added_address_info->set_offset_in_function(offset);
     added_address_info->set_module_path(function->loaded_module_path());
   }
 
-  const absl::flat_hash_map<uint64_t, FunctionStats>& functions_stats =
-      capture_data.functions_stats();
-  capture_info.mutable_function_stats()->insert(functions_stats.begin(), functions_stats.end());
+  const FunctionInfoMap<FunctionStats>& functions_stats = capture_data.functions_stats();
+  for (const auto& function_to_stats : functions_stats) {
```
antonrohr, line 103: nit, maybe:
```suggestion
  for (const auto& [function, function_stats] : functions_stats) {
```

antonrohr (commented):
In CaptureSerializer.cpp:36, it still says: `inline const std::string kRequiredCaptureVersion = "1.52";` maybe that should be changed to 1.54
Same for CaptureDeserializerLoadFuzzer:39

OrbitClientModel/CaptureSerializer.cpp
florian-kuebler: thanks, done

OrbitClientData/include/OrbitClientData/ProcessData.h
florian-kuebler: right. thanks!

OrbitClientModel/CaptureDeserializer.cpp
```diff
@@ -123,10 +133,29 @@ void LoadCaptureInfo(const CaptureInfo& capture_info, CaptureListener* capture_l
...
+    module_info.set_address_end(module.address_end());
+    module_info.set_build_id(module.build_id());
+    module_info.set_load_bias(module.load_bias());
+    modules.emplace_back(std::move(module_info));
+  }
+  process.UpdateModuleInfos(modules);
+  modules_callback(process_info, modules);
+
   capture_listener->OnCaptureStarted(
       std::move(process), absl::flat_hash_map<std::string, ModuleData*>(),
```
antonrohr, line 160: I think this module_map here needs to be set

florian-kuebler (commented):
> In CaptureSerializer.cpp:36, it still says: `inline const std::string kRequiredCaptureVersion = "1.52";` maybe that should be changed to 1.54
> Same for CaptureDeserializerLoadFuzzer:39

Nice spot. I changed them.

florian-kuebler:
I just removed the information for function info. Also loading modules, does now not need the process anymore.
There is one issue left, that allows to load symbols on a loaded capture.
Also, module info should not have the load address (currently needed at the creation)

florian-kuebler:
Continued in #1340

PR #1316, opened 2020-09-30
Fix Crash in IsFunctionSelected

florian-kuebler:
When switching from absolute address to function info values to
represent the selected functions, we querry the selected process
and the absolute address to find the function.
Both, the process and the function might be nullptr.
In those cases, the function can't be selected, thus we return false.

Test: Take a capture, load a capture


antonrohr (approved):
lgtm, thanks!

PR #1317, opened 2020-09-30
Customise tracepoint markers 

MariaSandru27:
For the 'All tracepoint events' extra track,
the tooltip also specified the process and thread.

The tracepoints that are in triggered in 
the target process are colored with
light green, a color that is not part
of Orbit's color scheme for the thread tracks.

Screenshot 2020-10-01 at 6 20 18 PM - Edited
Screenshot 2020-10-01 at 6 22 41 PM - Edited

### DIFF
--- a/OrbitGl/TracepointTrack.cpp
+++ b/OrbitGl/TracepointTrack.cpp
@@ -66,6 +66,8 @@ void TracepointTrack::UpdatePrimitives(uint64_t min_tick, uint64_t max_tick,
 
   const Color kWhiteTransparent(255, 255, 255, 190);
 
+  const Color kGrey(128, 128, 128, 255);
+
   const Color kGreenSelection(0, 255, 0, 255);
 
   if (!picking) {
@@ -76,7 +78,9 @@ void TracepointTrack::UpdatePrimitives(uint64_t min_tick, uint64_t max_tick,
           float radius = track_height / 4;
           Vec2 pos(time_graph_->GetWorldFromTick(time), pos_[1]);
           if (thread_id_ == TracepointEventBuffer::kAllTracepointsFakeTid) {
-            batcher->AddVerticalLine(pos, -track_height, z, kWhiteTransparent);
+            const Color color =
+                tracepoint.pid() == GOrbitApp->GetCaptureData().process_id() ? kGrey : kWhite;
+            batcher->AddVerticalLine(pos, -track_height, z, color);
           } else {
             batcher->AddVerticalLine(pos, -radius, z, kWhiteTransparent);
             batcher->AddVerticalLine(Vec2(pos[0], pos[1] - track_height), radius, z,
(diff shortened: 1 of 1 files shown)


OrbitGl/App.cpp
```diff
@@ -317,10 +317,6 @@ void OrbitApp::PostInit() {
 
   GCurrentTimeGraph->SetStringManager(string_manager_);
 
-  if (!absl::GetFlag(FLAGS_enable_tracepoint_feature)) {
```
ronaldfw, line 320: Since this is effectively enabling this feature for everyone, this change should be done in a different PR.

OrbitQt/orbitmainwindow.cpp
```diff
@@ -234,10 +234,6 @@ OrbitMainWindow::OrbitMainWindow(QApplication* a_App,
     ui->RightTabWidget->removeTab(ui->RightTabWidget->indexOf(ui->CodeTab));
   }
 
-  if (!absl::GetFlag(FLAGS_enable_tracepoint_feature)) {
```
ronaldfw, line 237: Similarly here, this should be in a different PR.

OrbitClientProtos/capture_data.proto
```diff
@@ -59,6 +59,7 @@ message TracepointEventInfo {
   int64 time = 3;
   int32 cpu = 4;
   uint64 tracepoint_info_key = 5;
+  bool is_in_target_process = 6;
```
dpallotti, line 62: I don't think you need this. You should be able to do
`if (tracepoint.pid() == GOrbitApp->GetCaptureData().process_id())`.

OrbitGl/TracepointTrack.cpp
```diff
@@ -142,12 +146,27 @@ std::string TracepointTrack::GetSampleTooltip(PickingId id) const {
...
-      "<b>Name:</b> %s [%s]<br/>",
-      tracepoint_event_info->cpu(), tracepoint_info.name(), tracepoint_info.category());
+  if (thread_id_ == TracepointEventBuffer::kAllTracepointsFakeTid) {
+    return absl::StrFormat(
+        "<b>%s : %s</b><br/>"
+        "<i>Tracepoint event</i><br/>"
+        "<br/>"
+        "<b>Core:</b> %d<br/>"
+        "<b>Process:</b> %d [%s]<br/>"
+        "<b>Thread:</b> %d [%s]<br/>",
```
dpallotti, line 156: Ah, I now see what's off. I would do the opposite, that is:

`%s [%d], <process/thread_name>, <process/thread_id>`

This is more consistent with other parts of Orbit.

OrbitGl/TracepointTrack.cpp
```diff
@@ -76,7 +76,11 @@ void TracepointTrack::UpdatePrimitives(uint64_t min_tick, uint64_t max_tick,
           float radius = track_height / 4;
           Vec2 pos(time_graph_->GetWorldFromTick(time), pos_[1]);
           if (thread_id_ == TracepointEventBuffer::kAllTracepointsFakeTid) {
-            batcher->AddVerticalLine(pos, -track_height, z, kWhiteTransparent);
+            if (tracepoint.is_in_target_process() == false) {
+              batcher->AddVerticalLine(pos, -track_height, z, kWhiteTransparent);
+            } else {
+              batcher->AddVerticalLine(pos, -track_height, z, kGreenSelection);
```
dpallotti, line 82: I'm not too sure about this green color, especially since it's already used for the selection. Not too sure what to suggest though. I would stay along the lines of white for one and grey for the other, or white for one and white but even more transparent for the other.

OrbitQt/orbitmainwindow.cpp
dpallotti, line 237: You should also remove the definition of the flag.

OrbitGl/TracepointTrack.cpp
pierricgimmig: Also, for compactness, I would suggest:

Color color = tracepoint.is_in_target_process() ? kGreenSelection : kWhiteTransparent;
batcher->AddVerticalLine(pos, -track_height, z, color);

pierricgimmig (commented):
Hi Maria, thanks for the change. I do agree with Ronald that we should remove the flag in a separate PR.

OrbitGl/TracepointTrack.cpp
MariaSandru27: Thank you!

OrbitGl/TracepointTrack.cpp
MariaSandru27: Thank you!

danielfenner (approved):
lgtm

pierricgimmig (approved):
Hi Maria, thanks for the modifications! LGTM

PR #1318, opened 2020-09-30
Integration of a web pipeline

beckerhe:
This PR integrates a web-building-and-bundling pipeline based on Webpack into our buildsystem. It allows to create web application which will be embedded into the Orbit executable.

4 commits:
1. Adding a NodeJS conan recipe. This recipe is based on the official distributions of NodeJS available for Linux and Windows.
2. Integrating the NodeJS package as a build dependency
3. Extending the license header check for .js files
4. Adding the webpack based build pipeline and integrating it with cmake.

### DIFF
--- a/WebUI/package.json
+++ b/WebUI/package.json
@@ -1,0 +1,19 @@
+{
+  "name": "orbit-web-ui",
+  "version": "1.0.0",
+  "description": "",
+  "private": true,
+  "scripts": {
+    "serve": "webpack-dev-server --open",
+    "build": "webpack --config=webpack.config.js --mode=production"
+  },
+  "keywords": [],
+  "author": "",
+  "license": "BSD",
+  "devDependencies": {
+    "html-webpack-plugin": "^4.5.0",
+    "webpack": "^4.44.2",
+    "webpack-cli": "^3.3.12",
+    "webpack-dev-server": "^3.11.0"
+  }
+}
--- a/WebUI/plugins/qrc-generator.js
+++ b/WebUI/plugins/qrc-generator.js
@@ -1,0 +1,37 @@
+// Copyright (c) 2020 The Orbit Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+/*
+  The QrcGeneratorPlugin for Webpack generates a Qt resource file
+  listing all the assets produced by Webpack.
+
+  This allows embedding of a webpack-generated bundle into a Qt-based
+  C++-application.
+
+  Qt's WebEngine module supports loading websites from QRC resources
+  out-of-the-box. So this is a convenient way to embed a static webpage.
+*/
+class QrcGeneratorPlugin {
+    apply(compiler) {
+        const onEmit = (comp) => {
+            const prefix = '<!DOCTYPE RCC>\n<RCC version="1.0">\n  <qresource prefix="/WebUI">\n';
+            const suffix = '  </qresource>\n</RCC>\n';
+            const files = Object.keys(comp.assets).reduce((str, filename) => str + `    <file>${filename}</file>\n`, '');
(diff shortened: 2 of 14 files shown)


WebUI/package.json
```diff
@@ -0,0 +1,19 @@
+{
+  "name": "WebUI",
```
reichlfl, line 2: Should probably put "OrbitWebUI" or something

WebUI/plugins/qrc-generator.js
```diff
@@ -0,0 +1,37 @@
...
+  This allows embedding of a webpack-generated bundle into a Qt-based
+  C++-application.
+
+  Qt's WebEngine module supports loading websites from QRC resources
+  out-of-the-box. So this is a convenient way to embed a static webpage.
+*/
+class QrcGeneratorPlugin {
+    apply(compiler) {
+        const onEmit = (comp) => {
+            const prefix = '<!DOCTYPE RCC>\n<RCC version="1.0">\n<qresource prefix="/WebUI">\n';
```
reichlfl, line 18: The QRC files generated by QT Designer don't have the doctype and version tag - which is probably not an issue, just wanted to point that out.

For readability, maybe add a '\t' before the <qresource> tag and '\t\t' before each file, or 4 spaces?
Edit: I see you're doing that for each file, so I'd add another indentation level.

third_party/conan/recipes/nodejs/conanfile.py
```diff
@@ -0,0 +1,48 @@
...
+        self.copy(pattern="npx.cmd", src=self._source_subfolder, dst="bin")
+        self.copy(pattern="*", src=os.path.join(self._source_subfolder, "include"), dst="include")
+        self.copy(pattern="*", src=os.path.join(self._source_subfolder, "lib"), dst="lib")
+        self.copy(pattern="*", src=os.path.join(self._source_subfolder, "node_modules"),
+                               dst=os.path.join("bin", "node_modules"))
+
+    def package_info(self):
+        bin_dir = os.path.join(self.package_folder, "bin")
+        self.output.info('Appending PATH environment variable: {}'.format(bin_dir))
+        self.env_info.PATH.append(bin_dir)
```
reichlfl, line 48: Will this append to the global PATH variable?

reichlfl (commented):
Thanks a lot! Overall looks great, I just have one question about the PATH.append.

third_party/conan/recipes/nodejs/conanfile.py
beckerhe: No, this only happens in conan itself. All build tools called by conan will see this enviroment. Its the way conan makes build tools from packages available to the current build.

In our case it's mostly cmake that will see that changed PATH variable.
That's why the `find_program` call in the `CMakeLists.txt` file will be able to find `npm`.

reichlfl: Ok, makes sense. Other comments are minor, LGTM!

reichlfl (approved):

WebUI/plugins/qrc-generator.js
beckerhe: Sure. I added two more leading whitespaces for the qresource-level.

Regarding the format: I was following the official docs (https://doc.qt.io/qt-5/resources.html) which suggest a format which includes the version and the doctype. So I would rather stick to that.

WebUI/package.json
beckerhe: good point. I changed it to `orbit-web-ui` since camel case is not encouraged/wanted by npm. Doesn't really matter since we won't upload that package, but still. Thanks!

third_party/conan/recipes/nodejs/conanfile.py
beckerhe: Thanks!

PR #1319, opened 2020-09-30
Fix presets for e2e test.

danielfenner:
Previously presets contained a module name /srv/game/hello_ggp_standalone.
Changed that to /mnt/developer/hello_ggp_standalone such that local runs
can use this data as well.

Test: Ran locally.


beckerhe (approved):
LGTM!

IrinaShkviro (approved):

PR #1320, opened 2020-09-30
Fix track filtering in capture view

pierricgimmig:
- Track filtering now works for all track types
- Filters are now case insensitive


florian-kuebler (approved):
Makes sense. Thanks!

PR #1323, opened 2020-10-01
Draw white text boxes with black text for iterators overlay

ronaldfw:
The white text showing iterator timings is hard to read and can be
easily confused with other text. This is in particular true as all
text for textboxes is drawn at the end above essentially everything
else. We fix this here partly by using solid white text boxes with black
text for the iterator timings.

Since the white timer text is still drawn last, the black text for the
iterator timings is still overdrawn with timer text when text is drawn
in the same position.

### DIFF
--- a/OrbitGl/TimeGraph.cpp
+++ b/OrbitGl/TimeGraph.cpp
@@ -621,29 +621,36 @@ std::string GetTimeString(const TimerInfo& timer_a, const TimerInfo& timer_b) {
   return colors[index % kNumColors];
 }
 
-void DrawIteratorBox(GlCanvas* canvas, Vec2 pos, Vec2 size, const Color& color,
-                     const std::string& label, const std::string& time, float text_y) {
+}  // namespace
+
+void TimeGraph::DrawIteratorBox(GlCanvas* canvas, Vec2 pos, Vec2 size, const Color& color,
+                                const std::string& label, const std::string& time,
+                                float text_box_y) {
   Box box(pos, size, GlCanvas::kZValueOverlay);
   canvas->GetBatcher()->AddBox(box, color);
 
   std::string text = absl::StrFormat("%s: %s", label, time);
 
-  constexpr const float kAdditionalSpaceForLine = 10.f;
-  constexpr const float kLeftOffset = 5.f;
-
   float max_size = size[0];
-  canvas->GetTextRenderer().AddTextTrailingCharsPrioritized(
-      text.c_str(), pos[0] + kLeftOffset, text_y + kAdditionalSpaceForLine, GlCanvas::kZValueText,
-      Color(255, 255, 255, 255), time.length(), GParams.font_size, max_size);
 
-  constexpr const float kOffsetBelowText = kAdditionalSpaceForLine / 2.f;
-  Vec2 line_from(pos[0], text_y + kOffsetBelowText);
-  Vec2 line_to(pos[0] + size[0], text_y + kOffsetBelowText);
+  const Color kBlack(0, 0, 0, 255);
+  int text_width = canvas->GetTextRenderer().AddTextTrailingCharsPrioritized(
+      text.c_str(), pos[0], text_box_y + layout_.GetTextOffset(), GlCanvas::kZValueText, kBlack,
+      time.length(), GetFontSize(), max_size);
+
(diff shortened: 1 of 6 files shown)


ronaldfw:
Here's a screenshot.

white_text_boxes

OrbitGl/TimeGraph.cpp
```diff
@@ -621,30 +621,37 @@ std::string GetTimeString(const TimerInfo& timer_a, const TimerInfo& timer_b) {
...
-  canvas->GetTextRenderer().AddTextTrailingCharsPrioritized(
-      text.c_str(), pos[0] + kLeftOffset, text_y + kAdditionalSpaceForLine, GlCanvas::kZValueText,
-      Color(255, 255, 255, 255), time.length(), GParams.font_size, max_size);
+  float text_width = canvas->GetTextRenderer().AddTextTrailingCharsPrioritized(
+      text.c_str(), pos[0], text_y + kAdditionalSpaceForLine, GlCanvas::kZValueText,
+      kBlack, time.length(), GParams.font_size, max_size);
 
   constexpr const float kOffsetBelowText = kAdditionalSpaceForLine / 2.f;
+  Vec2 white_box_size(text_width, GetTextBoxHeight());
+  Vec2 white_box_position(pos[0], text_y + kOffsetBelowText);
```
karupayun, line 644: TextBoxHeight will scale with the vertical zooming, but either GParams.font_size nor kOffsetBelowText not.

This is what you see after scaling the text in manual instrumentation variables:
Screen Shot 2020-10-01 at 11 50 22

This is what you see after scaling with the iterator:
Screen Shot 2020-10-01 at 11 49 38


There are 2 possible solutions:
1- Don't scale the text box height (modifying the method and adding a new one GetScalableTextBoxHeight or whatever)
2- Scale everything by modifying:
GParams.font_size -> GetFontSize()
kOffsetBelowText -> GetTextOffset
kAdditionalSpaceForLine -> GetTextOffset x 2

OrbitGl/TimeGraph.cpp
```diff
@@ -621,30 +621,37 @@ std::string GetTimeString(const TimerInfo& timer_a, const TimerInfo& timer_b) {
...
+
+  Box white_box(white_box_position, white_box_size, GlCanvas::kZValueOverlay);
+  canvas->GetBatcher()->AddBox(white_box, kWhite);
```
karupayun, line 647: I would prefer that the white line for the measuring is in the middle of the box (y-coordinate)
Screen Shot 2020-10-01 at 12 09 50

karupayun (changes requested):
Very nice PR Ronald! In my opinion, they seem much more nicer now. I wrote two comments.

BTW, do you know why the width of the box is not always the correct one? I can't figure it.
Screen Shot 2020-10-01 at 11 49 38
Screen Shot 2020-10-01 at 12 11 08

ronaldfw:
> BTW, do you know why the width of the box is not always the correct one? I can't figure it.

Hmm, ok, that doesn't look good. I'll look into it.

OrbitGl/TimeGraph.cpp
ronaldfw: Agree, done.

ronaldfw:
> > BTW, do you know why the width of the box is not always the correct one? I can't figure it.
> 
> Hmm, ok, that doesn't look good. I'll look into it.

I *think* I mostly fixed this. I was returning the wrong string width from the AddTextTrailingCharsPrioritized method. I now explicitly recompute the width of the string used. However, I don't see the effect that you observed with incorrect box sizes quite as strongly as you did. Could you try again with the latest version to see if this looks better now?

OrbitGl/TimeGraph.cpp
ronaldfw: Done.

karupayun (approved):
It looks perfect now! LGTM!

Thanks for the changes :)

BTW, a very very little detail. I would prefer to have the white box on top of the horizontal lines.
Screen Shot 2020-10-01 at 13 25 24

ronaldfw:
> BTW, a very very little detail. I would prefer to have the white box on top of the horizontal lines.

Good suggestion, fixed.

PR #1324, opened 2020-10-01
Add WebEngine module

beckerhe:
The WebEngine module mainly adds a dialog class which shows
a fullscreen web page.

Currently it's nowhere in use. In a follow-up PR this code will
be used to implement the disassembly view.


WebEngine/JsonTransport.cpp
```diff
@@ -0,0 +1,35 @@
...
+namespace WebEngine {
+
+JsonTransport::JsonTransport(QWebSocket* socket) : socket_(socket) {
+  QObject::connect(socket_.get(), &QWebSocket::textMessageReceived, this,
+                   [this](const QString& msg) {
+                     QJsonParseError parse_error;
+                     const auto doc = QJsonDocument::fromJson(msg.toUtf8(), &parse_error);
+
+                     if (parse_error.error) {
+                       ERROR("Failed to parse web channel message: %s. The error was: %s",
```
dpallotti, line 21: How long is a message typically? If it tends to span over multiple lines, maybe it would be better to have
"Failed to parse web channel message: \<error\>. Message:\n
\<message\>"
So that the error is on top and that the message starts on a new line.

WebEngine/JsonTransport.h
```diff
@@ -0,0 +1,41 @@
...
+/* Json-based transport for QtWebChannel
+
+   A transport in the QtWebChannel framework is the link
+   between to C++ WebChannel object and the websocket talking
+   to the JavaScript part.
+
+   This is a very basic implementation and also the default.
+   It encodes all messages as JSON. The default JavaScript
+   channel implementation (qwebchannel.js) is doing the same.
+   So it can't be easily swapped out without rewriting the JS
```
dpallotti, line 26: I don't really understand this sentence. Why is that that if the JavaScript channel implementation does the same as this implementation, then "it" (what is it?) can't be easily swapped out?

WebEngine/include/WebEngine/DeleteLaterDeleter.h
```diff
@@ -0,0 +1,35 @@
...
+   Often a member function of the object to be deleted is part
+   of the callstack (eg. a triggered signal). Deleting the object
+   right away would lead to a crash higher up in the callstack.
+
+   This can be avoided by deleting "later".
+
+   Example type declaration:
+   std::unique_ptr<QObject, DeleteLaterDeleter> obj;
+*/
+struct DeleteLaterDeleter {
```
dpallotti, line 27: One could say "del*a*ter"...


Ok sorry.

WebEngine/CMakeLists.txt
```diff
@@ -0,0 +1,20 @@
+# Copyright (c) 2020 The Orbit Authors. All rights reserved. Use of this source
+# code is governed by a BSD-style license that can be found in the LICENSE file.
```
dpallotti, line 2: The license headers of the other CMake files have newlines in the same places as the C++ files.

WebEngine/Dialog.cpp
```diff
@@ -0,0 +1,47 @@
...
+#include <QWebSocketServer>
+#include <memory>
+
+#include "JsonTransport.h"
+
+namespace WebEngine {
+
+Dialog::Dialog(QWebEngineProfile* profile, QWidget* parent) : QDialog(parent), page_(profile) {
+  web_socket_server_.reset(
+      // NOLINTNEXTLINE
```
dpallotti, line 21: What is this for?

dpallotti (approved):
I miss *a lot* of background knowledge to understand what's really going on. Looks good though.

WebEngine/Dialog.cpp
```diff
@@ -0,0 +1,47 @@
...
+      new QWebSocketServer{"Web channel server", QWebSocketServer::NonSecureMode});
+  web_socket_server_->listen(QHostAddress::LocalHost, 12345);
```
dimitry-, line 23: what if listen fails?

WebEngine/JsonTransport.h
```diff
@@ -0,0 +1,41 @@
...
+   part as well.
```
dimitry-, line 27: Add an example code?

WebEngine/include/WebEngine/Dialog.h
```diff
@@ -0,0 +1,65 @@
...
+   with Q_INVOKABLE or marked as a slot.
+
+   Signals work as well and can trigger events on the JavaScript side.
+   Qt Properties can also be used.
+
+   Be aware that constructing that object can take around 100 to 200
+   milliseconds, especially when the Chromium engine needs to be fired
+   up first. Depending on the use case that delay might be noticable to
+   the user. In that case it helps to construct the dialog upfront and
+   hide it until it's needed.
```
dimitry-, line 39: Example?

WebEngine/Dialog.cpp
beckerhe: clang-tidy will (and should) warn about the "new". But in this case we need it.

WebEngine/Dialog.cpp
beckerhe: Good point! Will think about something

WebEngine/JsonTransport.cpp
beckerhe: Yes, makes sense!

WebEngine/JsonTransport.h
beckerhe: Good point. Rephrasing...

WebEngine/include/WebEngine/DeleteLaterDeleter.h
beckerhe: :grinning: 

As a big fan of (bad) puns I can appreciate that!

WebEngine/include/WebEngine/Dialog.h
beckerhe: Added!

WebEngine/JsonTransport.cpp
beckerhe: Done.

WebEngine/JsonTransport.h
beckerhe: Done.

WebEngine/CMakeLists.txt
beckerhe: Fixed.

WebEngine/JsonTransport.h
beckerhe: Done.

WebEngine/Dialog.cpp
beckerhe: So, I made two changes:
- I changed the code to have the OS choose the listening port which should already make it hard to fail at that point. That was one thing I wanted to do anyway. Furthermore I added a getter for the WebSocketServer such that the port number can be requested by the consumer.
- I added a log message specifying the error.

This is obviously no hard error. A factory function returning a ErrorMessageOr would be good, but this doesn't work well with Qt-code because QObject-derived classes are neither copyable nor movable. I'm also not sure yet how this error should be handled best on the consumer side. It's probably not recoverable, so an error message box is probably the only solution. But I would prefer to do additional changes (if needed) when the consumer code is ready.

WDYT?

WebEngine/JsonTransport.h
```diff
@@ -0,0 +1,49 @@
...
+   channel implementation (qwebchannel.js) is also expecting
+   the message to be JSON-encoded.
+   So the serialization mechanism can't be easily swapped
+   out without rewriting the JS part as well.
+
+   Usage: Call QWebChannel's connectTo method whenever a new
+   websocket connection is pending:
+
+   QWebChannel* web_channel = GetWebChannel();
+   QWebSocketServer* web_socket_server = GetWebSocketServer();
```
dimitry-, line 34: new

WebEngine/Dialog.cpp
dimitry-: sgtm..

question: do we need another solution for error handling, if ErrorMessageOr does not work here? Maybe have one that works with unique_ptr<_, DeleteLaterDeleter>?

dimitry- (approved):
otherwise lgtm

WebEngine/Dialog.cpp
beckerhe: Good question. I was thinking a lot lately about that. In Qt almost all classes support an invalid state while we usually try to write stateless classes, meaning you can't construct them from invalid input data. And these two concepts don't compose well. I' not certain yet, but I think our code also needs to handle an invalid state when it's based on / interacts with Qt classes. But yeah, still thinking about it.

WebEngine/JsonTransport.h
beckerhe: Ah sorry. I think my example was a bit misleading. I changed it.

PR #1326, opened 2020-10-01
Add file_directory location in ClientGgp

kisenshi:
By default the orbit file is now created in /var/game/ but the user can change it if they want to.
This change allows the orbit file to be created without issue by default when the client is called from a game.


danielfenner (approved):
lgtm

kisenshi:
Merging this; thanks!

PR #1329, opened 2020-10-02
Prioritize the ID when printing track labels

karupayun:
Related to b/169663076.

Screenshots:
prioritize1
prioritize2

### DIFF
--- a/OrbitGl/TimeGraph.cpp
+++ b/OrbitGl/TimeGraph.cpp
@@ -763,10 +763,12 @@ void TimeGraph::DrawTracks(GlCanvas* canvas, PickingMode picking_mode) {
         std::string process_name = GOrbitApp->GetCaptureData().process_name();
         thread_track->SetName(process_name);
         thread_track->SetLabel(process_name + " (all threads)");
+        thread_track->SetNumberOfPrioritizedTrailingCharacters(strlen("(all threads)"));
       } else {
         const std::string& thread_name = GOrbitApp->GetCaptureData().GetThreadName(tid);
         track->SetName(thread_name);
         std::string track_label = absl::StrFormat("%s [%u]", thread_name, tid);
+        thread_track->SetNumberOfPrioritizedTrailingCharacters((std::to_string(tid)).size() + 2);
         track->SetLabel(track_label);
       }
     }
@@ -807,6 +809,8 @@ std::shared_ptr<GpuTrack> TimeGraph::GetOrCreateGpuTrack(uint64_t timeline_hash)
     std::string label = OrbitGl::MapGpuTimelineToTrackLabel(timeline);
     track->SetName(timeline);
     track->SetLabel(label);
+    // This min combine two cases, label == timeline and when label includes timeline
+    track->SetNumberOfPrioritizedTrailingCharacters(std::min(label.size(), timeline.size() + 2));
     tracks_.emplace_back(track);
     gpu_tracks_[timeline_hash] = track;
   }
--- a/OrbitGl/TimeGraph.cpp
+++ b/OrbitGl/TimeGraph.cpp
@@ -762,13 +762,15 @@ void TimeGraph::DrawTracks(GlCanvas* canvas, PickingMode picking_mode) {
         // This is the process_track_.
         std::string process_name = GOrbitApp->GetCaptureData().process_name();
         thread_track->SetName(process_name);
-        thread_track->SetLabel(process_name + " (all threads)");
-        thread_track->SetNumberOfPrioritizedTrailingCharacters(strlen("(all threads)"));
+        const std::string_view all_threads = " (all_threads)";
+        thread_track->SetLabel(process_name.append(all_threads));
+        thread_track->SetNumberOfPrioritizedTrailingCharacters(all_threads.size() - 1);
       } else {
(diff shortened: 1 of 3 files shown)


beckerhe:
Could this mess with custom track names like from the manual instrumentation API or others coming in the future?

I imagine that having a name with '(' or '[' will trigger that behaviour even if it might be unintended.

Maybe that behaviour should be controlled by the code that also generates the label? In this case we wouldn't need to detect the part at the end. 

WDYT?

karupayun:
> Could this mess with custom track names like from the manual instrumentation API or others coming in the future?
> 
> I imagine that having a name with '(' or '[' will trigger that behaviour even if it might be unintended.
> 
> Maybe that behaviour should be controlled by the code that also generates the label? In this case we wouldn't need to detect the part at the end.
> 
> WDYT?

I was not sure. I though about that a few. The another option implies adding a ID number for every different kind of track, when probably there is no going to be a collision there. 
Also I can improve that taking the rightmost opening symbol ('[','{' or '('), but if you think that it's still strange, I can move the to the another option.

beckerhe:
> > Could this mess with custom track names like from the manual instrumentation API or others coming in the future?
> > I imagine that having a name with '(' or '[' will trigger that behaviour even if it might be unintended.
> > Maybe that behaviour should be controlled by the code that also generates the label? In this case we wouldn't need to detect the part at the end.
> > WDYT?
> 
> I was not sure. I though about that a few. The another option implies adding a ID number for every different kind of track, when probably there is no going to be a collision there.
> Also I can improve that taking the rightmost opening symbol ('[','{' or '('), but if you think that it's still strange, I can move the to the another option.

I'm not sure since I don't know much about the code and how it interacts. Having an ID separate from the label and combining them just before rendering sounds to me like a proper solution. But I'm not sure if it's worth it and if that has other problem.

@pierricgimmig What is your opinion?

pierricgimmig:
I think this should be handled by thread track type. @beckerhe is right, this could have have undesired behavior for manual instrumentation tracks.  One possible solution would be to have a "virtual size_t GetNumberOfPrioritizedTralingCharacters() " method that would default to 0 but that individual track types could override. That number could then be computed once when we create the label and be cached.

pierricgimmig (commented):
Hi Pablo, thanks for looking at this. Please see comment.

karupayun:
> @beckerhe is right, this could have have undesired behavior for manual instrumentation tracks. One possible solution would be to have a "virtual size_t GetNumberOfPrioritizedTralingCharacters() " method that would default to 0 but that individual track types could override. That number could then be computed once when we create the label and be cached.

Thanks for your suggestions. You are clearly right. I've pushed a new commit. 

Instead of making a virtual function, I made Get and Set functions, so we can set the number exactly when we are setting the label to be printed. WDYT?

OrbitGl/TimeGraph.cpp
```diff
@@ -763,10 +763,12 @@ void TimeGraph::DrawTracks(GlCanvas* canvas, PickingMode picking_mode) {
         std::string process_name = GOrbitApp->GetCaptureData().process_name();
         thread_track->SetName(process_name);
         thread_track->SetLabel(process_name + " (all threads)");
+        thread_track->SetNumberOfPrioritizedTrailingCharacters(strlen("(all threads)"));
```
pierricgimmig, line 766: Not sure why this is done on every draw, but it seems to be the pattern here. We should probably clean that up in another PR.

pierricgimmig (approved):
LGTM! thanks for the modifications.

pierricgimmig:
> Instead of making a virtual function, I made Get and Set functions, so we can set the number exactly when we are setting the label to be printed. WDYT?

Yeah even better, thanks.

OrbitGl/TimeGraph.cpp
beckerhe, line 766: Can you add a string_view constant for the string " (all threads)"?

OrbitGl/TimeGraph.cpp
```diff
@@ -763,10 +763,12 @@ void TimeGraph::DrawTracks(GlCanvas* canvas, PickingMode picking_mode) {
...
       } else {
         const std::string& thread_name = GOrbitApp->GetCaptureData().GetThreadName(tid);
         track->SetName(thread_name);
         std::string track_label = absl::StrFormat("%s [%u]", thread_name, tid);
+        thread_track->SetNumberOfPrioritizedTrailingCharacters((std::to_string(tid)).size() + 2);
```
beckerhe, line 771: You are converting tid twice to string by different algorithms and expecting the result is the same. Probably it will, so not a big deal. But I think it would be nicer to create `auto tid_str = std::to_string(tid)` and use it in absl::StrFormat and in the .size-call. WDYT?

beckerhe (approved):
LGTM! Minor comments!

OrbitGl/TimeGraph.cpp
karupayun: Perfect, done!

OrbitGl/TimeGraph.cpp
karupayun: Yes, what do you think now?

beckerhe: Great. Thanks!

PR #1331, opened 2020-10-02
Fix crash in picking white boxes

karupayun:
Although it was not intended, white boxes were pickable, which make a
crash.

Related to b/169725465.


reichlfl (approved):
Change LGTM, thanks for fixing!

PR #1332, opened 2020-10-02
Multiple PerfEventVisitors in PerfEventProcessor, don't take ownership

dpallotti:
`PerfEventProcessor` now suports multiple `PerfEventVisitor`s and doesn't take
ownership of them, instead is passed bare pointers.
Having multiple visitors will allow for another visitor to handle thread states.
Not taking ownership will allow to store the `PerfEventVisitor`s independently
of `PerfEventProcessor` and hence to access them.

### DIFF
--- a/OrbitLinuxTracing/GpuTracepointEventProcessor.h
+++ b/OrbitLinuxTracing/GpuTracepointEventProcessor.h
@@ -23,7 +23,7 @@ class GpuTracepointEventProcessor {
 
  private:
   // Keys are context, seqno, and timeline
-  typedef std::tuple<uint32_t, uint32_t, std::string> Key;
+  using Key = std::tuple<uint32_t, uint32_t, std::string>;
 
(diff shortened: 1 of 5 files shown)


florian-kuebler (approved):
Thanks. From the code LGTM. Could you add in the commit message "why" we want multiple visitors and also that the owner will be the tracer-thread.

Meta-question:
Why is the tracer-thread the owner of the visitors and not the eventprocessor.

dpallotti:
> Thanks. From the code LGTM. Could you add in the commit message "why" we want multiple visitors and also that the owner will be the tracer-thread.
> 
> Meta-question:
> Why is the tracer-thread the owner of the visitors and not the eventprocessor.

You're right, I specified both in the commit message. This is all for thread-state-related changes that will come at a later time.

florian-kuebler (approved):
LGTM, thanks!

PR #1334, opened 2020-10-02
Fix "node not found" problem

beckerhe:
NodeJS is installed by conan as a build dependency but not added to the
user's global PATH variable. So the conan-supplied NodeJS executable is
only found when the executing environment is initialized by conan, i.e.
when cmake is called from conan like it is the case in build.{sh,ps1}.

When cmake is called either manually or by a third party like CLion the
NodeJS-executable won't be found because it's not in the PATH. This can
be circumvented by caching the contents of the PATH variable during the
configuration phase and restoring it on every subsequent call of cmake.

Limitation: This requires the initial build to be initiated by conan,
usually by running `build.{sh,ps1}` or the bootstrap script.

### DIFF
--- a/WebUI/CMakeLists.txt
+++ b/WebUI/CMakeLists.txt
@@ -36,11 +36,15 @@ else()
   find_program(NPM npm)
 endif()
 
-execute_process(COMMAND ${NPM} install WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR})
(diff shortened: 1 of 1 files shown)


dpallotti (approved):
This is all sorcery to me!

beckerhe:
> This is all sorcery to me!

Thanks anyway! :-)

PR #1335, opened 2020-10-02
Add LinuxTracingUtils/GetThreadState

dpallotti:
This will be used to fetch the state of threads at the beginning of the capture
(although in an approximate way).

### DIFF
--- a/OrbitLinuxTracing/LinuxTracingUtils.h
+++ b/OrbitLinuxTracing/LinuxTracingUtils.h
@@ -31,6 +31,11 @@ std::vector<pid_t> GetAllTids();
 
 std::string GetThreadName(pid_t tid);
 
+// The association between a character and a thread state is documented at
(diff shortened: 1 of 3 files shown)


OrbitLinuxTracing/LinuxTracingUtils.h
```diff
@@ -31,6 +31,8 @@ std::vector<pid_t> GetAllTids();
 
 std::string GetThreadName(pid_t tid);
```
dimitry-, line 33: Since it returns char - add a comment or link explaining how to decode it.

dimitry- (approved):
otherwise lgtm

OrbitLinuxTracing/LinuxTracingUtils.h
dpallotti: Good point, added links.

beckerhe (approved):
LGTM! Thanks, Davide!

PR #1337, opened 2020-10-05
Handle thread names with spaces when parsing /proc/<pid>/stat

dpallotti:
`GetCumulativeCpuTimeFromProcess` and `GetThreadState` parsed the content of
`/proc/<pid>/stat` by splitting on spaces, but didn't consider that the thread
name enclosed in parentheses could contain spaces itself.

Bug: http://b/170073156

Test: Start a process with one/two/three spaces in its name and check that its
displayed CPU utilization is correct. Update test for `GetThreadState`.

### DIFF
--- a/OrbitLinuxTracing/LinuxTracingUtils.cpp
+++ b/OrbitLinuxTracing/LinuxTracingUtils.cpp
@@ -122,13 +122,25 @@ std::optional<char> GetThreadState(pid_t tid) {
 
   std::string first_line{};
   std::getline(stream, first_line);
-  std::vector<std::string_view> fields = absl::StrSplit(first_line, ' ', absl::SkipWhitespace{});
 
-  constexpr size_t kStateIndex = 2;
-  if (fields.size() <= kStateIndex) {
+  // Remove fields up to comm (process name) as this, enclosed in parentheses, could contain spaces.
+  size_t last_closed_paren_index = first_line.find_last_of(')');
+  if (last_closed_paren_index == std::string::npos) {
     return std::nullopt;
   }
-  return fields[kStateIndex][0];
+  std::string_view first_line_excl_pid_comm =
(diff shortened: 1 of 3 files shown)


OrbitLinuxTracing/LinuxTracingUtils.cpp
```diff
@@ -122,6 +123,12 @@ std::optional<char> GetThreadState(pid_t tid) {
 
   std::string first_line{};
   std::getline(stream, first_line);
+
+  // Remove the process name enclosed in parentheses in the second position
+  // as it could contain spaces.
+  first_line = std::regex_replace(first_line, std::regex{"\\(.*\\)"}, "()",
+                                  std::regex_constants::format_first_only);
```
beckerhe, line 130: Since `std::regex` is known to be slow I was wondering if something `std::string_view{first_line}.substr(first_line.find(')')` wouldn't be enough to get rid of the problem. Of course error handling needs to be added and the indexes need adjustment. WDYT?

Nice catch, by the way!

dpallotti: Makes sense, I guess a regex is a bit overkill here. I'll switch away from it.

dpallotti: I've done what you suggested, please have a look as the code is obviously a bit uglier now.

beckerhe: I like it. Your verbose variable names make it pretty obvious what's happening. At least I believe so - obviously I'm now already familiar with what needed to be solved.

beckerhe (approved):
Thanks!

pierricgimmig:
#1337, LEET!

dpallotti:
> #1337, LEET!

Hehe I didn't want to say it!

PR #1340, opened 2020-10-05
continued: Remove Module Base Address and Load Bias from FunctionInfo

antonrohr:
This continues PR #1315 

Everything should now work as expected, aka the context menus in the different views should be correct and show options like "hook" only if Orbit is connected and the correct process is selected. 

I marked this as draft, because I can't help but think this is currently more complicated than it needs to be. The commit that moves "module_map" out of capture data seemed necessary, because capture deserialization is done off the main thread and the "module_map" is stored in DataManager which can only be accessed from the main thread. 
I would propose to change the deserialization instead and still keep a copy of the module_map in CaptureData as it was done before. I think this would simplify some function calls, but I am unsure if its the best way to go forward. wdyt? Maybe one of you has a good idea on a conceptual level.

### DIFF
--- a/OrbitCaptureClient/CaptureClient.cpp
+++ b/OrbitCaptureClient/CaptureClient.cpp
@@ -38,7 +38,7 @@ static CaptureOptions::InstrumentedFunction::FunctionType IntrumentedFunctionTyp
 
 ErrorMessageOr<void> CaptureClient::StartCapture(
     ThreadPool* thread_pool, const ProcessData& process,
-    const absl::flat_hash_map<std::string, ModuleData*>& module_map,
+    const OrbitClientData::ModuleManager& module_manager,
     absl::flat_hash_map<uint64_t, FunctionInfo> selected_functions,
     TracepointInfoSet selected_tracepoints, bool enable_introspection) {
   absl::MutexLock lock(&state_mutex_);
@@ -50,19 +50,16 @@ ErrorMessageOr<void> CaptureClient::StartCapture(
 
   state_ = State::kStarting;
 
+  // TODO(168797897) Here a copy of the process is created. The loaded modules of this process were
+  // likely filled when the process was selected, which might be a while back. Between then and now
+  // the loaded modules might have changed. Up to date information about which modules are loaded
+  // should be used here. (Even better: while taking a capture this should always be up to date)
   ProcessData process_copy = process;
 
-  // TODO(168797897) Here a copy of the module_map is created. This module_map likely was downloaded
-  // when the process was selected, which might be a while back. Between then and now the loaded
-  // modules might have changed. Up to date information about which modules are loaded should be
-  // used here. (Even better: while taking a capture this should always be up to date)
-  absl::flat_hash_map<std::string, ModuleData*> module_map_copy = module_map;
-
-  thread_pool->Schedule([this, process = std::move(process_copy),
-                         module_map = std::move(module_map_copy),
+  thread_pool->Schedule([this, process = std::move(process_copy), &module_manager,
                          selected_functions = std::move(selected_functions), selected_tracepoints,
                          enable_introspection]() mutable {
-    Capture(std::move(process), std::move(module_map), std::move(selected_functions),
+    Capture(std::move(process), module_manager, std::move(selected_functions),
             std::move(selected_tracepoints), enable_introspection);
   });
 
@@ -70,7 +67,7 @@ ErrorMessageOr<void> CaptureClient::StartCapture(
 }
 
 void CaptureClient::Capture(ProcessData&& process,
-                            absl::flat_hash_map<std::string, ModuleData*>&& module_map,
+                            const OrbitClientData::ModuleManager& module_manager,
                             absl::flat_hash_map<uint64_t, FunctionInfo> selected_functions,
                             TracepointInfoSet selected_tracepoints, bool enable_introspection) {
   CHECK(reader_writer_ == nullptr);
@@ -96,14 +93,14 @@ void CaptureClient::Capture(ProcessData&& process,
 
   capture_options->set_trace_thread_state(absl::GetFlag(FLAGS_thread_state));
   capture_options->set_trace_gpu_driver(true);
-  for (const auto& pair : selected_functions) {
-    const FunctionInfo& function = pair.second;
+  for (const auto& [absolute_address, function] : selected_functions) {
     CaptureOptions::InstrumentedFunction* instrumented_function =
         capture_options->add_instrumented_functions();
     instrumented_function->set_file_path(function.loaded_module_path());
-    instrumented_function->set_file_offset(FunctionUtils::Offset(function));
-    instrumented_function->set_absolute_address(FunctionUtils::GetAbsoluteAddress(
-        function, process, *module_map.at(function.loaded_module_path())));
+    const ModuleData* module = module_manager.GetModuleByPath(function.loaded_module_path());
+    CHECK(module != nullptr);
+    instrumented_function->set_file_offset(FunctionUtils::Offset(function, *module));
+    instrumented_function->set_absolute_address(absolute_address);
     instrumented_function->set_function_type(
         IntrumentedFunctionTypeFromOrbitType(function.orbit_type()));
   }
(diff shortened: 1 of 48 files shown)


OrbitGl/CallStackDataView.cpp
```diff
@@ -191,10 +190,6 @@ CallStackDataView::CallStackDataViewFrame CallStackDataView::GetFrameFromRow(int
 
 CallStackDataView::CallStackDataViewFrame CallStackDataView::GetFrameFromIndex(
     int index_in_callstack) {
-  if (index_in_callstack >= static_cast<int>(callstack_.GetFramesCount())) {
```
florian-kuebler, line 192: I guess you are removing this, as this does never happen. Could you then add a CHECK for now?

OrbitGl/SamplingReportDataView.cpp
```diff
@@ -49,6 +49,8 @@ std::string SamplingReportDataView::GetValue(int row, int column) {
     case kColumnInclusive:
       return absl::StrFormat("%.2f", func.inclusive);
     case kColumnModuleName:
+      // LOG("%s", func.module_path);
```
florian-kuebler, line 52: Can you remove those?

OrbitGl/SamplingReportDataView.cpp
```diff
@@ -159,22 +161,25 @@ absl::flat_hash_set<const FunctionInfo*> SamplingReportDataView::GetFunctionsFro
...
     CHECK(sampled_function.absolute_address != 0);
-    ModuleData* module =
-        capture_data.FindModuleByAddress(sampled_function.absolute_address, module_map);
-    if (module != nullptr) {
-      modules.insert(module);
+    auto result = process->FindModuleByAddress(sampled_function.absolute_address);
+    if (!result) {
+      ERROR("result %s", result.error().message());
+    }
+    if (result) {
```
florian-kuebler, line 177: why not just else?

OrbitGl/LiveFunctionsDataView.cpp
```diff
@@ -163,21 +163,23 @@ std::vector<std::string> LiveFunctionsDataView::GetContextMenu(
...
 
-    // Is that function actually inside a module of the process (i.e. can we disassemble)?
-    const FunctionInfo* actual_function =
-        capture_data.FindFunctionByAddress(absolute_address, module_map, false);
-    const bool function_exists = actual_function != nullptr;
+    if (GOrbitApp->IsCaptureConnected(capture_data)) {
+      // Is that function actually inside a module of the process (i.e. can we disassemble)?
+      const FunctionInfo* actual_function =
+          capture_data.FindFunctionByAddress(absolute_address, module_map, false);
+      const bool function_exists = actual_function != nullptr;
```
florian-kuebler, line 170: I guess, this can now be a check? What happens if you have a process running for long time. Start orbit in a connected state, and load a capture of that process (which is still running). Can you hook then? Currently, we would need to reload the symbols, right?

OrbitGl/DataManager.h
```diff
@@ -39,6 +41,7 @@ class DataManager final {
   void set_selected_process(int32_t pid);
 
   [[nodiscard]] const ProcessData* GetProcessByPid(int32_t process_id) const;
+  [[nodiscard]] ProcessData* GetMutableProcessByPid(int32_t process_id);
```
florian-kuebler, line 44: Where do you use that? In general, I think we should keep those accessors minimal. They basically allow us to completely shortcut the thread-safety measurements of DataManager (here, by passing that pointer to another thread).

florian-kuebler (commented):
Thanks for continuing that work. LGTM. I left some comments and questions.

OrbitCaptureClient/CaptureClient.cpp
```diff
@@ -69,7 +63,7 @@ ErrorMessageOr<void> CaptureClient::StartCapture(
 }
 
 void CaptureClient::Capture(ProcessData&& process,
-                            absl::flat_hash_map<std::string, ModuleData*>&& module_map,
+                            absl::flat_hash_map<std::string, ModuleData*> module_map,
```
dimitry-, line 66: should this be a const ref?

OrbitGl/App.cpp
```diff
@@ -155,25 +154,26 @@ void OrbitApp::OnCaptureStarted(ProcessData&& process,
 
 void OrbitApp::OnCaptureComplete() {
   capture_data_.FilterBrokenCallstacks();
-  SamplingProfiler sampling_profiler(*capture_data_.GetCallstackData(), capture_data_);
-  capture_data_.set_sampling_profiler(sampling_profiler);
-  main_thread_executor_->Schedule(
-      [this, sampling_profiler = std::move(sampling_profiler)]() mutable {
-        RefreshCaptureView();
+  main_thread_executor_->Schedule([this]() {
+    SamplingProfiler sampling_profiler(
```
florian-kuebler, line 158: Can you trie to move this one line above again to do this on the capture thread?

OrbitGl/DataManager.h
antonrohr: Yes, I agree, but it is necessary for one specific thing: Updating the loaded module list of a process. The tradeoff here is between having this function and writing an extra function that updates this list inside of DataManager. I decided for this tradeoff, but I am actually not sure anymore. Let me know what you think.

OrbitGl/LiveFunctionsDataView.cpp
antonrohr: Now: If the same process is still running (determined by isCaptureConnected) then you can select/unselect and disassemble

OrbitGl/SamplingReportDataView.cpp
antonrohr: done

OrbitGl/SamplingReportDataView.cpp
antonrohr: done

OrbitGl/CallStackDataView.cpp
antonrohr: done

OrbitGl/App.cpp
antonrohr: done

antonrohr:
@dimitry- & @florian-kuebler 

This PR is finally in a ready for review state. I cleaned up a lot and squashed (fixup-ed) a few commits into one, because before there were changes introduced that I removed again. 

The commit messages should be a good description what is happening. Please look there. 

I tested a bunch (including some weird edge cases) and I think this is in a good shape, any further testing would be great of course.

OrbitGl/DataManager.h
florian-kuebler: I would argue for the function in data manager. But have no strong opinion on it. So I think it is fine for now. Thanks for the explanation.

florian-kuebler (approved):
Thanks Anton for those changes! I really like the new ModuleManager.
LGTM!

OrbitClientProtos/capture_data.proto
```diff
@@ -106,6 +122,8 @@ message CaptureInfo {
   map<uint64, FunctionStats> function_stats = 9;
   repeated TracepointInfo tracepoint_infos = 10;
   repeated TracepointEventInfo tracepoint_event_infos = 11;
+  ProcessInfo process = 13;
+  repeated ModuleInfo modules = 14;
```
dimitry-, line 126: ok, looks like moving these dependencies and CaptureInfo out of proto is much larger task, lets leave it as is.

OrbitGl/App.cpp
```diff
@@ -785,6 +781,17 @@ bool OrbitApp::SelectProcess(const std::string& process) {
...
 
+bool OrbitApp::IsCaptureConnected(const CaptureData& capture) const {
+  const ProcessData* selected_process = GetSelectedProcess();
+  if (selected_process == nullptr) return false;
+
+  const ProcessData* capture_process = capture.process();
+  CHECK(capture_process != nullptr);
+
+  return selected_process->pid() == capture_process->pid() &&
+         selected_process->full_path() == capture_process->full_path();
```
dimitry-, line 792: TODO: check buildid?

dimitry- (approved):
LGTM, thanks!

OrbitGl/App.cpp
antonrohr: I put a TODO and it's a good idea to think about the details of what that would actually mean. Are build-ids required and  so on.

PR #1341, opened 2020-10-06
Remove GLU from Orbit

reichlfl:


OrbitGl/GlUtils.cpp
```diff
@@ -9,11 +9,47 @@
...
-  GLenum errCode = glGetError();
-  const char* errString;
-  if (errCode != GL_NO_ERROR) {
-    errString = reinterpret_cast<const char*>(gluErrorString(errCode));
-    LOG("OpenGL ERROR : %s", errString);
+  GLenum error_code = glGetError();
+  const char* error_string;
+  if (error_code != GL_NO_ERROR) {
+    error_string = GetGlError(error_code);
+    LOG("OpenGL ERROR : %s", error_string);
```
beckerhe, line 53: Maybe `const char* error_string = GetGlError(error_code)`? As far as I can see the bigger scope is not needed, is it?

beckerhe (approved):
Very cool. Thanks a lot for that change!

OrbitGl/GlUtils.cpp
reichlfl: Changed, thanks

PR #1342, opened 2020-10-06
Don't interpolate values in GraphTrack rendering

pierricgimmig:
Use steps instead of drawing straight lines from point to point. 
The value displayed on the graph now matches the plot.


karupayun (approved):
LGTM!

As I told you in the call, I think the final solution should be discussed here: https://buganizer.corp.google.com/issues/169315390

Tomorrow I will send some PR including prettier white boxes.

PR #1345, opened 2020-10-08
Collect thread states, send them to the client, store and serialize them

dpallotti:
This is the first pull request regarding thread states.
It touches many of Orbit's components hence the number of reviewers.
The only main missing part is the visualization in the capture view.
The unification with the logic that handles scheduling slices is still open.

### DIFF
--- a/OrbitCaptureClient/CaptureEventProcessor.cpp
+++ b/OrbitCaptureClient/CaptureEventProcessor.cpp
@@ -21,6 +21,7 @@ using orbit_grpc_protos::InternedCallstack;
 using orbit_grpc_protos::InternedString;
 using orbit_grpc_protos::SchedulingSlice;
 using orbit_grpc_protos::ThreadName;
+using orbit_grpc_protos::ThreadStateSlice;
 
 void CaptureEventProcessor::ProcessEvent(const CaptureEvent& event) {
   switch (event.event_case()) {
@@ -32,12 +33,6 @@ void CaptureEventProcessor::ProcessEvent(const CaptureEvent& event) {
       break;
     case CaptureEvent::kCallstackSample:
       ProcessCallstackSample(event.callstack_sample());
-      break;
-    case CaptureEvent::kInternedTracepointInfo:
-      ProcessInternedTracepointInfo(event.interned_tracepoint_info());
-      break;
-    case CaptureEvent::kTracepointEvent:
-      ProcessTracepointEvent(event.tracepoint_event());
       break;
     case CaptureEvent::kFunctionCall:
       ProcessFunctionCall(event.function_call());
@@ -51,8 +46,17 @@ void CaptureEventProcessor::ProcessEvent(const CaptureEvent& event) {
     case CaptureEvent::kThreadName:
       ProcessThreadName(event.thread_name());
       break;
+    case CaptureEvent::kThreadStateSlice:
+      ProcessThreadStateSlice(event.thread_state_slice());
+      break;
     case CaptureEvent::kAddressInfo:
       ProcessAddressInfo(event.address_info());
+      break;
+    case CaptureEvent::kInternedTracepointInfo:
+      ProcessInternedTracepointInfo(event.interned_tracepoint_info());
+      break;
+    case CaptureEvent::kTracepointEvent:
+      ProcessTracepointEvent(event.tracepoint_event());
       break;
     case CaptureEvent::EVENT_NOT_SET:
       ERROR("CaptureEvent::EVENT_NOT_SET read from Capture's gRPC stream");
@@ -176,6 +180,11 @@ void CaptureEventProcessor::ProcessGpuJob(const GpuJob& gpu_job) {
 
 void CaptureEventProcessor::ProcessThreadName(const ThreadName& thread_name) {
   capture_listener_->OnThreadName(thread_name.tid(), thread_name.name());
+}
+
+void CaptureEventProcessor::ProcessThreadStateSlice(
+    const ThreadStateSlice& /*thread_state_slice*/) {
+  // TODO(dpallotti,b/168790833): Send to OrbitApp and ClientGgp.
 }
 
 void CaptureEventProcessor::ProcessAddressInfo(const AddressInfo& address_info) {
--- a/OrbitCaptureClient/CaptureEventProcessor.cpp
+++ b/OrbitCaptureClient/CaptureEventProcessor.cpp
@@ -9,6 +9,7 @@
 
 using orbit_client_protos::CallstackEvent;
 using orbit_client_protos::LinuxAddressInfo;
+using orbit_client_protos::ThreadStateSliceInfo;
 using orbit_client_protos::TimerInfo;
 
 using orbit_grpc_protos::AddressInfo;
@@ -182,9 +183,46 @@ void CaptureEventProcessor::ProcessThreadName(const ThreadName& thread_name) {
   capture_listener_->OnThreadName(thread_name.tid(), thread_name.name());
 }
 
-void CaptureEventProcessor::ProcessThreadStateSlice(
-    const ThreadStateSlice& /*thread_state_slice*/) {
-  // TODO(dpallotti,b/168790833): Send to OrbitApp and ClientGgp.
+void CaptureEventProcessor::ProcessThreadStateSlice(const ThreadStateSlice& thread_state_slice) {
+  ThreadStateSliceInfo slice_info;
+  slice_info.set_tid(thread_state_slice.tid());
+  switch (thread_state_slice.thread_state()) {
+    case ThreadStateSlice::kRunning:
+      slice_info.set_thread_state(ThreadStateSliceInfo::kRunning);
+      break;
+    case ThreadStateSlice::kRunnable:
+      slice_info.set_thread_state(ThreadStateSliceInfo::kRunnable);
+      break;
+    case ThreadStateSlice::kInterruptibleSleep:
+      slice_info.set_thread_state(ThreadStateSliceInfo::kInterruptibleSleep);
+      break;
+    case ThreadStateSlice::kUninterruptibleSleep:
+      slice_info.set_thread_state(ThreadStateSliceInfo::kUninterruptibleSleep);
+      break;
+    case ThreadStateSlice::kStopped:
+      slice_info.set_thread_state(ThreadStateSliceInfo::kStopped);
+      break;
+    case ThreadStateSlice::kTraced:
+      slice_info.set_thread_state(ThreadStateSliceInfo::kTraced);
+      break;
+    case ThreadStateSlice::kDead:
+      slice_info.set_thread_state(ThreadStateSliceInfo::kDead);
+      break;
+    case ThreadStateSlice::kZombie:
+      slice_info.set_thread_state(ThreadStateSliceInfo::kZombie);
+      break;
+    case ThreadStateSlice::kParked:
+      slice_info.set_thread_state(ThreadStateSliceInfo::kParked);
+      break;
+    case ThreadStateSlice::kIdle:
+      slice_info.set_thread_state(ThreadStateSliceInfo::kIdle);
+      break;
+    default:
+      UNREACHABLE();
+  }
+  slice_info.set_begin_timestamp_ns(thread_state_slice.begin_timestamp_ns());
+  slice_info.set_end_timestamp_ns(thread_state_slice.end_timestamp_ns());
+  capture_listener_->OnThreadStateSlice(std::move(slice_info));
 }
 
 void CaptureEventProcessor::ProcessAddressInfo(const AddressInfo& address_info) {
--- a/OrbitClientModel/CaptureSerializer.cpp
+++ b/OrbitClientModel/CaptureSerializer.cpp
@@ -62,6 +62,17 @@ CaptureInfo GenerateCaptureInfo(
   capture_info.mutable_thread_names()->insert(capture_data.thread_names().begin(),
                                               capture_data.thread_names().end());
 
+  for (const auto& tid_and_thread_state_slices : capture_data.thread_state_slices()) {
+    // Note that thread state slices are saved in their original order only among the same thread,
+    // but all slices related to the same thread are saved sequentially. This might not be desired
+    // if the capture is opened in a streaming fashion.
+    for (const auto& thread_state_slice : tid_and_thread_state_slices.second) {
+      orbit_client_protos::ThreadStateSliceInfo* added_thread_state_slice =
+          capture_info.add_thread_state_slices();
+      added_thread_state_slice->CopyFrom(thread_state_slice);
+    }
+  }
+
   capture_info.mutable_address_infos()->Reserve(capture_data.address_infos().size());
   for (const auto& address_info : capture_data.address_infos()) {
     orbit_client_protos::LinuxAddressInfo* added_address_info = capture_info.add_address_infos();
--- a/OrbitGrpcProtos/capture.proto
+++ b/OrbitGrpcProtos/capture.proto
@@ -35,6 +35,8 @@ message CaptureOptions {
 
   repeated InstrumentedFunction instrumented_functions = 5;
 
+  bool trace_thread_state = 8;
+
   bool trace_gpu_driver = 6;
 
   repeated TracepointInfo instrumented_tracepoint = 7;
--- a/OrbitGrpcProtos/capture.proto
+++ b/OrbitGrpcProtos/capture.proto
@@ -118,10 +118,39 @@ message GpuJob {
 }
 
 message ThreadName {
-  int32 pid = 1;
+  int32 pid = 1;  // Note that pid is currently not set.
   int32 tid = 2;
   string name = 3;
   uint64 timestamp_ns = 4;
+}
+
+message ThreadStateSlice {
+  int32 pid = 1;  // pid is currently not set as we don't have the information.
+  int32 tid = 2;
+
+  // These are the ones listed in
+  // /sys/kernel/debug/tracing/events/sched/sched_switch/format and in
+  // https://github.com/torvalds/linux/blob/master/fs/proc/array.c in
+  // task_state_array[], with the difference that the OS doesn't distinguish
+  // between "running" (scheduled on a CPU) and "runnable" (ready to be
+  // scheduled but not actually scheduled), rather they are the same state
+  // (often referred to as "running", sometimes as "runnable or running").
+  enum ThreadState {
+    kRunning = 0;
+    kRunnable = 1;
+    kInterruptibleSleep = 2;
+    kUninterruptibleSleep = 3;
+    kStopped = 4;
+    kTraced = 5;
+    kDead = 6;
+    kZombie = 7;
+    kParked = 8;
+    kIdle = 9;
+  }
+  ThreadState thread_state = 3;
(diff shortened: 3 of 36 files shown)


OrbitLinuxTracing/ThreadStateVisitor.cpp
```diff
@@ -0,0 +1,272 @@
...
+  ThreadStateSlice slice;
+  slice.set_tid(tid);
+  slice.set_thread_state(open_state.state);
+  slice.set_begin_timestamp_ns(open_state.begin_timestamp_ns);
+  slice.set_end_timestamp_ns(timestamp_ns);
+  tid_open_states_.insert_or_assign(tid, OpenState{kNewState, timestamp_ns});
+  return slice;
+}
+
+std::optional<ThreadStateSlice> ThreadStateManager::OnContextSwitchOut(
```
florian-kuebler, line 112: Again I would prefer `OnSchedSwitchOut`

OrbitLinuxTracing/ThreadStateVisitor.cpp
```diff
@@ -0,0 +1,272 @@
...
+    uint64_t timestamp_ns, pid_t tid, ThreadStateSlice::ThreadState new_state) {
+  auto open_state_it = tid_open_states_.find(tid);
+  if (open_state_it == tid_open_states_.end()) {
+    ERROR("Processed sched:sched_switch(out) but previous state of thread %d is unknown", tid);
```
florian-kuebler, line 116: (Also for the above): I would assume that this happens always on the beginning of the capture and is something usual. So maybe omit the error?

OrbitLinuxTracing/ThreadStateVisitor.cpp
```diff
@@ -0,0 +1,272 @@
...
+// Note: Since we use PerfEventProcessor to process perf_event_open events in order, OnNewTask,
+// OnSchedWakeup, OnContextSwitchIn, OnContextSwitchOut are expected to be called in order of
+// timestamp. But the initial thread states are retrieved (and OnInitialState is called) after the
+// perf_event_open file descriptors have been enabled, so that we don't lose thread states between
+// retrieving the initial states and enabling the file descriptors. It is then common for some of
+// the first tracepoint events to have a timestamp lower than the timestamp of initial retrieval. In
+// all these cases, we discard the previous known state (the one retrieved at the beginning, with a
+// larger timestamp) and replace it with the thread state carried by the tracepoint.
+
+void ThreadStateManager::OnInitialState(uint64_t timestamp_ns, pid_t tid,
```
florian-kuebler, line 20: Feels a bit strange that the ThreadStateManager and the ThreadStateVisitor are in the same file? Could you separate them?

OrbitLinuxTracing/ThreadStateVisitor.cpp
```diff
@@ -0,0 +1,272 @@
...
+    slice.set_begin_timestamp_ns(open_state.begin_timestamp_ns);
+    slice.set_end_timestamp_ns(timestamp_ns);
+    slices.emplace_back(std::move(slice));
+  }
+  return slices;
+}
+
+void ThreadStateVisitor::ProcessInitialState(uint64_t timestamp_ns, pid_t tid, char state_char) {
+  std::optional<ThreadStateSlice::ThreadState> initial_state = GetThreadStateFromChar(state_char);
+  if (!initial_state.has_value()) {
```
florian-kuebler, line 171: That should be a CHECK, shouldn't it?

OrbitLinuxTracing/ThreadStateVisitor.cpp
```diff
@@ -0,0 +1,272 @@
...
+    default:
+      return std::nullopt;
+  }
+}
+
+// Associates a ThreadStateSlice::ThreadState to the bits of the prev_state field of the
+// sched:sched_switch tracepoint. The association is given away by "print ftm" in
+// /sys/kernel/debug/tracing/events/sched/sched_switch/format or by
+// https://github.com/torvalds/linux/blob/master/fs/proc/array.c.
+ThreadStateSlice::ThreadState ThreadStateVisitor::GetThreadStateFromBits(uint64_t bits) {
```
florian-kuebler, line 260: It feels a bit strange, that this method returns runnable in case of an invalid value, and the one above returns nullopt. I think the should behave the same. Also I think there should be a CHECK somewhere (either on the caller or on the callee), that the will be never an "unexpected" value.

OrbitLinuxTracing/ThreadStateVisitor.h
```diff
@@ -0,0 +1,80 @@
...
+//                                       |                                          .  |
+//                                       |                   sched:sched_switch(in) .  |
+//                                       |               ---------------- . . . . . .  |
+//                                       |               | Not runnable |              |
+//                                       --------------- | incl. exited | <-------------
+//                                sched:sched_wakeup     ----------------    sched_switch(out)
+//                             (OR sched:sched_waking)                      with prev_state!='R'
+//                                                                   (ALSO sched:sched_process_exit)
+
+class ThreadStateManager {
```
florian-kuebler, line 34: Again, I would split them into to files.

OrbitLinuxTracing/TracerThread.cpp
```diff
@@ -409,6 +410,21 @@ bool TracerThread::OpenThreadNameTracepoints(const std::vector<int32_t>& cpus) {
...
+void TracerThread::InitThreadStateVisitor() {
+  thread_state_visitor_ = std::make_unique<ThreadStateVisitor>();
+  thread_state_visitor_->SetListener(listener_);
+  event_processor_.AddVisitor(thread_state_visitor_.get());
+}
+
+bool TracerThread::OpenThreadStateTracepoints(const std::vector<int32_t>& cpus) {
+  absl::flat_hash_map<int32_t, int> thread_state_tracepoint_ring_buffer_fds_per_cpu;
+  return OpenFileDescriptorsAndRingBuffersForAllTracepoints(
+      {{"sched", "sched_switch", &sched_switch_ids_},
```
florian-kuebler, line 422: Aren't you missign the wakeup_new/new_task/sched_waking events? Am I missing something?

florian-kuebler (commented):
Thanks for that change. This is really nice. I left some comments, but in general it looks good.

OrbitGrpcProtos/capture.proto
```diff
@@ -122,6 +124,32 @@ message ThreadName {
   uint64 timestamp_ns = 4;
 }
 
+message ThreadStateSlice {
+  int32 pid = 1;
+  int32 tid = 2;
```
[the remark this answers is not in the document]
dpallotti: The pid is not set as we don't have it, but I want to have the field right away as it's only because of technical reason and we'll probably set it in the future. Let's remember that these are protos v3 and all fields are optional.

pierricgimmig (approved):
Hi Davide, thank you for the change, visualizing thread states will be super useful! I have only minor comments, but in general it looks really good.

OrbitClientModel/CaptureSerializer.cpp
```diff
@@ -62,6 +62,17 @@ CaptureInfo GenerateCaptureInfo(
   capture_info.mutable_thread_names()->insert(capture_data.thread_names().begin(),
                                               capture_data.thread_names().end());
 
+  for (const auto& tid_and_thread_state_slices : capture_data.thread_state_slices()) {
+    // Note that thread state slices are saved in their original order only among the same thread,
+    // but all slices related to the same thread are saved sequentially. This might not be desired
+    // if the capture is opened in a streaming fashion.
```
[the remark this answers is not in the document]
florian-kuebler: Well we store them by thread in our data model. So I think it would be the easiest way to simulate something like real-time, if we just seriealize them also by thread, and then have multiple threads reading the serialized file.

Of curse, we could also just serialize all events together sorted by timestamp.

OrbitCaptureClient/CaptureEventProcessor.cpp
```diff
@@ -178,6 +183,48 @@ void CaptureEventProcessor::ProcessThreadName(const ThreadName& thread_name) {
   capture_listener_->OnThreadName(thread_name.tid(), thread_name.name());
 }
 
+void CaptureEventProcessor::ProcessThreadStateSlice(const ThreadStateSlice& thread_state_slice) {
+  ThreadStateSliceInfo slice_info;
+  slice_info.set_tid(thread_state_slice.tid());
```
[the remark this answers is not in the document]
florian-kuebler: Okay, thanks for the explanation. When writing this, I confused the client and the grpc proto. Maybe (if not already there), add a comment to the grpc proto?

OrbitLinuxTracing/ThreadStateVisitor.cpp
florian-kuebler: I see your point. But couldn't we then make it to a private class and let the test be a friend class?
Anyways. I don't have a strong opinion on that. I personally would separate this (mostly for readability). Probably going the friend class path, would be a bit of an overkill. So feel free to ignore.

OrbitLinuxTracing/ThreadStateVisitor.cpp
```diff
@@ -0,0 +1,272 @@
...
+  if (open_state.state == kNewState || open_state.state == ThreadStateSlice::kRunning) {
+    // It seems to be somewhat common for a thread to receive a wakeup
+    // while already in runnable or running state: disregard the state change.
+    return std::nullopt;
+  }
+
+  if (open_state.state == ThreadStateSlice::kZombie ||
+      open_state.state == ThreadStateSlice::kDead) {
+    ERROR("Processed sched:sched_wakeup for thread %d but unexpected previous state %s", tid,
+          ThreadStateSlice::ThreadState_Name(open_state.state));
```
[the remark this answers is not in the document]
florian-kuebler: Hmm, I don't know. If we think that this should not happen, then we should fail. Tbh, we will never notice that there was an error, when just logging it.
If you say, that this might be a strange kernel thing, that could happen, then drop the error (or use a LOG).

OrbitLinuxTracing/ThreadStateVisitor.cpp
florian-kuebler: Okay, then see above.

OrbitLinuxTracing/TracerThread.cpp
florian-kuebler: Okay, then I misunderstood those. I thought there are cases were sched_waking are used, and others were sched_wakeup are use.d

OrbitLinuxTracing/ThreadStateVisitor.cpp
florian-kuebler: In which case, there would be invalid values again.

OrbitLinuxTracing/ThreadStateVisitor.cpp
florian-kuebler: I understand your point. Especially, as the user experience of a crashing service, is not good atm.
But I still have the impression, that we rely on certain behavior of the system and might be doing something wrong, if that system behaves differently. We probably need the metrics uploader or missuse the crash server, for such kind of errors. Were we notify our system, but do not crash the service.

OrbitGrpcProtos/capture.proto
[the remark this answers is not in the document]
dpallotti: Added comment.

OrbitLinuxTracing/ThreadStateVisitor.cpp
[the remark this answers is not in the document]
dpallotti: Logging it is very useful, as I monitor the service's logs when developing and testing, and from logs like these we can learn what to expect from the system. And as it's an unexpected situation, for me it's an error, which is exactly saying "it's unexpected, but we can/must handle it gracefully".

OrbitLinuxTracing/ThreadStateVisitor.cpp
dpallotti: > Can multiple bits of the last byte of "bits" be set to 1 at the same time?

Maybe, but I don't think I observed it when testing.

> What would that mean?

I don't know.

In https://github.com/torvalds/linux/blob/master/fs/proc/array.c the comment says 
"The task state array is a strange "bitmap" of reasons to sleep. Thus "running" is zero, and you can test for combinations of others with simple bit tests."
In such a situation, I prefer to pick the first state rather than none of them. On the other hand I didn't want to add complexity to support multiple states at once when it might only happen in some very specific situation. It's a compromise.
But I'll definitely add a log for it.

OrbitCaptureClient/CaptureEventProcessor.cpp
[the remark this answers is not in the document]
dpallotti: Yep, I added comments to both the grpc and the client protos.

OrbitClientModel/CaptureSerializer.cpp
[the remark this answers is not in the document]
dpallotti: I think it's overkill to plan for reading with multiple threads.

The reasonable thing to do here is to sort by timestamp, but I didn't do it because it adds an overhead that doesn't provide any value until *all* events are sorted by timestamp. Until then, having states come in order when for example they all come before callstacks which all come before (or after?) timers is not useful.

OrbitLinuxTracing/ThreadStateVisitor.cpp
```diff
@@ -0,0 +1,272 @@
...
+      // 'I' (Idle) only applies to kernel threads. See
+      // https://github.com/torvalds/linux/commit/06eb61844d841d0032a9950ce7f8e783ee49c0d0.
+      return ThreadStateSlice::kIdle;
+    default:
+      return std::nullopt;
+  }
+}
+
+// Associates a ThreadStateSlice::ThreadState to the bits of the prev_state field of the
+// sched:sched_switch tracepoint. The association is given away by "print ftm" in
```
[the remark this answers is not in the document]
dpallotti: Whoops.

OrbitClientModel/CaptureSerializer.cpp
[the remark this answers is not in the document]
florian-kuebler: agree. lets postpone this.

OrbitLinuxTracing/ThreadStateVisitor.cpp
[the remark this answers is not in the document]
florian-kuebler: But thats for development. Logging in production is still useful, if something goes wrong, you could check the log. But here, there will be nothing going wrong. As said in the other comment. for the moment, fine for me (but only as the service crash on CHECK, is not really visible for the user). Longer term solution, must be, to be able to monitor those errors, or have a good handling of a crash in the service, s.t. we can do checks here.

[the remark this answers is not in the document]
dpallotti: I don't think that crashing the service just to get a report for a case that we can easily handle gracefully and that doesn't really depend on us is a good idea, even if the service crashing was handled super nicely. Of course it would be nice to get those logs from production, I totally agree on that, but that's a completely different story and I assume will be part of the work on metrics.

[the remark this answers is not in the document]
florian-kuebler: Yeah mostly agree. 
But, I think it can be argued whether "that doesn't really depend on us". Those cases in the service in general are often certain assumptions on the kernel to behave that way. If it fails, it means our assumptions were wrong and we should investigate why.

But nevermind, lets keep it as it is for now. At one point in time, we need to reargue about all the ERRORS in service anyways.

florian-kuebler (approved):
As far as I see, you addressed/commented on most of my concerns. 
So LGTM! Feel free to merge as soon as you like.

OrbitLinuxTracing/ThreadStateVisitor.cpp
pierricgimmig: cool, lgtm!

pierricgimmig (approved):

OrbitLinuxTracing/ThreadStateVisitor.cpp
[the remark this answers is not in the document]
ronaldfw: Adding a comment here because I think this is important: I'm fully with dpallotti@ on this. We should not potentially make Orbit unusable for our users (what if the way the game uses the kernel triggers a bad assumption every time?) in case that an assumption we make is wrong, in particular when we can handle that case gracefully. Logging an error for sure. CHECKs should be reserved for assumptions about our own code and in cases where not asserting may lead to worse.

OrbitLinuxTracing/ThreadStateVisitor.cpp
```diff
@@ -0,0 +1,276 @@
...
+
+  const OpenState& open_state = open_state_it->second;
+  if (timestamp_ns < open_state.begin_timestamp_ns) {
+    tid_open_states_.insert_or_assign(tid, OpenState{new_state, timestamp_ns});
+    return std::nullopt;
+  }
+
+  // As we are switching out of a CPU, if the previous state was kRunnable, assume it was kRunning.
+  // This is because when we retrieve the initial thread states we have no way to distinguish
+  // between kRunnable and kRunning. After all, "running" is not a real thread state.
```
ronaldfw, line 129: Same question here: Is "running" not a real thread state, or is "runnable" not a real thread state? (TASK_RUNNING seems to be a real state the Linux scheduler code.)

OrbitLinuxTracing/ThreadStateVisitor.h
```diff
@@ -0,0 +1,80 @@
...
+#include "PerfEventVisitor.h"
+#include "absl/container/flat_hash_map.h"
+
+namespace LinuxTracing {
+
+// ThreadStateManager stores the state of threads, handles the state switches,
+// builds and returns ThreadStateSlices.
+// The following diagram shows the relationship between the states and the tracepoints.
+//
+//       task:task_newtask                             sched:sched_switch(in)
```
ronaldfw, line 19: Nice diagram!

ronaldfw (approved):
LGTM, only some minor comments or questions!

OrbitGrpcProtos/capture.proto
```diff
@@ -116,12 +118,38 @@ message GpuJob {
...
   int32 tid = 2;
   string name = 3;
   uint64 timestamp_ns = 4;
 }
 
+message ThreadStateSlice {
+  int32 pid = 1;  // pid is currently not set as we don't have the information.
+  int32 tid = 2;
+
+  // Except for "running" (which is not a real state), these are the ones listed
```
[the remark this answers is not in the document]
dpallotti: Yeah, so this is a matter of naming. In my naming with runnable I mean "can be run on a cpu" ("schedulable"), and with running "is currently run on the cpu" ("scheduled"). This is the naming that for example Perfetto, GAPID, Systrace use. This matches the common language.
But for the OS they are the same state. https://github.com/torvalds/linux/blob/master/fs/proc/array.c and https://man7.org/linux/man-pages/man5/proc.5.html call it "running", but https://www.man7.org/linux/man-pages/man1/ps.1.html#PROCESS_STATE_CODES calls it "running or runnable" (probably giving these words the same meaning as above). I guess that instead of saying that running is not a real state, I should say that running and runnable are the same state. I'll rephrase.

PR #1347, opened 2020-10-08
Fix strange jumping when click-dragging

karupayun:
Related to b/169918990. Simply change a width x height.

### DIFF
--- a/OrbitGl/CaptureWindow.cpp
+++ b/OrbitGl/CaptureWindow.cpp
@@ -80,10 +80,10 @@ void CaptureWindow::MouseMoved(int x, int y, bool left, bool /*right*/, bool /*m
 
     time_graph_.GetWorldMinMax(world_min, world_max);
 
-    auto width = static_cast<float>(GetWidth());
-
-    world_top_left_x_ = world_click_x_ - static_cast<float>(x) / width * world_width_;
-    world_top_left_y_ = world_click_y_ + static_cast<float>(y) / width * world_height_;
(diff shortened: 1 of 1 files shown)


florian-kuebler (approved):
LGTM. Thanks!

OrbitGl/CaptureWindow.cpp
```diff
@@ -81,9 +81,10 @@ void CaptureWindow::MouseMoved(int x, int y, bool left, bool /*right*/, bool /*m
     time_graph_.GetWorldMinMax(world_min, world_max);
 
     auto width = static_cast<float>(GetWidth());
+    auto height = static_cast<float>(GetHeight());
 
     world_top_left_x_ = world_click_x_ - static_cast<float>(x) / width * world_width_;
-    world_top_left_y_ = world_click_y_ + static_cast<float>(y) / width * world_height_;
+    world_top_left_y_ = world_click_y_ + static_cast<float>(y) / height * world_height_;
```
florian-kuebler, line 87: Actually, as there is now only one use of width/height, I would get rid of the variable again, and inline the call+cast.
https://google.github.io/styleguide/cppguide.html#Local_Variables also says that the variable should be as close as possible to the usage. So if you decide to keep it as a variable, you would need to move that variable directly to the line.

karupayun: Perfect, done!

florian-kuebler (approved):
Thanks!

PR #1348, opened 2020-10-08
Refactor: Stop renaming/relabeling in each draw.

karupayun:
Before this commit we renamed each track every time when we
draw them. Now, we are making that only once.

### DIFF
--- a/OrbitGl/TimeGraph.cpp
+++ b/OrbitGl/TimeGraph.cpp
@@ -42,7 +42,7 @@ TimeGraph::TimeGraph() : batcher_(BatcherId::kTimeGraph) {
       GetOrCreateThreadTrack(TracepointEventBuffer::kAllTracepointsFakeTid);
 
   // The process track is a special ThreadTrack of id "kAllThreadsFakeTid".
-  process_track_ = GetOrCreateThreadTrack(SamplingProfiler::kAllThreadsFakeTid);
+  process_track_ = nullptr;
 
   async_timer_info_listener_ =
       std::make_unique<ManualInstrumentationManager::AsyncTimerInfoListener>(
@@ -96,7 +96,7 @@ void TimeGraph::Clear() {
       GetOrCreateThreadTrack(TracepointEventBuffer::kAllTracepointsFakeTid);
 
   // The process track is a special ThreadTrack of id "kAllThreadsFakeTid".
-  process_track_ = GetOrCreateThreadTrack(SamplingProfiler::kAllThreadsFakeTid);
+  process_track_ = nullptr;
 
   SetIteratorOverlayData({}, {});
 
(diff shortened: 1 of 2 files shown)


OrbitGl/TimeGraph.cpp
```diff
@@ -42,7 +42,7 @@ TimeGraph::TimeGraph() : batcher_(BatcherId::kTimeGraph) {
       GetOrCreateThreadTrack(TracepointEventBuffer::kAllTracepointsFakeTid);
 
   // The process track is a special ThreadTrack of id "kAllThreadsFakeTid".
-  process_track_ = GetOrCreateThreadTrack(SamplingProfiler::kAllThreadsFakeTid);
+  process_track_ = nullptr;
```
florian-kuebler, line 45: I though about this again, and that member is only used in the SortTracks method. So I would remove it from the class, and have it as a local variable in SortTracks

OrbitGl/TimeGraph.cpp
```diff
@@ -319,6 +319,9 @@ void TimeGraph::ProcessTimer(const TimerInfo& timer_info, const FunctionInfo* fu
     } else {
       scheduler_track_->OnTimer(timer_info);
       cores_seen_.insert(timer_info.processor());
+      uint32_t num_cores = GetNumCores();
```
florian-kuebler, line 322: I don't think that you should set the name/number of cores everytime you process a scheduling timer. This should be probably done in the GetOrCreateSchedulerTrack

florian-kuebler (approved):
Despite the two minor comments, LGTM. Thanks!

OrbitGl/TimeGraph.cpp
karupayun: Yes, you are right!

OrbitGl/TimeGraph.cpp
karupayun: The little issue here is that we don't know the NumCores before processing all timers, do we?

cores_seen_ is simply a set of cores (in my case 8) and we are inserting each processor several times, I changed it, but not sure if you will like it

pierricgimmig (approved):
Thanks for following up on this Pablo! LGTM.

OrbitGl/TimeGraph.cpp
```diff
@@ -318,7 +312,12 @@ void TimeGraph::ProcessTimer(const TimerInfo& timer_info, const FunctionInfo* fu
       ++thread_count_map_[timer_info.thread_id()];
     } else {
       scheduler_track_->OnTimer(timer_info);
-      cores_seen_.insert(timer_info.processor());
+      if (!cores_seen_.count(timer_info.processor())) {
```
florian-kuebler, line 315: I feel that this cores_seen_ is a bit unintuitive. Couldn't we just have a `num_cores_` variable, and here check `if(num_cores_ < timer_info.processor()) num_cores_ = timer_info.processor(); ....`? 
But anyways. Other than that, it looks good. Also this is just a nit. Feel free to ignore.

florian-kuebler (approved):
Despite the nit, LGTM. Thanks!

OrbitGl/TimeGraph.cpp
karupayun: Oh, I didn't suppose that the processors went from 0 to N-1. Is it a guarantee?

florian-kuebler: Yes, it is.

PR #1349, opened 2020-10-08
Only request data update on live view when needed

ronaldfw:
The live functions data view is handled differently than other data
views as creation of it is a bit more complex than for the view we can
handle generically. Therefore, this data view requires an explicit
extra call to OnDataChanged. So far, this call was done independent of
the type of data views to be updated. This leads to a bug with frame
tracks where some data in the live functions data view class is reset
that should not be reset. Note that this mimics the update behavior of
the generic other data view panels.

Bug: b/170392477.
Test: Manual test of taking a capture, adding and removing frame
track(s).

### DIFF
--- a/OrbitGl/App.h
+++ b/OrbitGl/App.h
@@ -354,7 +354,6 @@ class OrbitApp final : public DataViewFactory, public CaptureListener {
   std::unique_ptr<ProcessesDataView> processes_data_view_;
   std::unique_ptr<ModulesDataView> modules_data_view_;
   std::unique_ptr<FunctionsDataView> functions_data_view_;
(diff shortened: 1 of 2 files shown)


florian-kuebler (approved):
Makes sense. Thanks!
Nit (feel free to ignore): Commit message should contain Bug: (if present) and Test:

dpallotti (approved):

ronaldfw:
Thanks for the review!

ronaldfw:
> Nit (feel free to ignore): Commit message should contain Bug: (if present) and Test:

Good point, added.

PR #1351, opened 2020-10-09
Add more precise labels to manual instrumentation

karupayun:
In this PR we add nicer a more precise labels to variables in manual instrumentation.
The label points to the last measured value of every variable. Also we have modified the
blue lines again to interpolate values.
Related to b/169315390.

interpolate 1
interpolate 2

### DIFF
--- a/OrbitGl/GraphTrack.cpp
+++ b/OrbitGl/GraphTrack.cpp
@@ -13,7 +13,6 @@ void GraphTrack::Draw(GlCanvas* canvas, PickingMode picking_mode) {
   Batcher* batcher = canvas->GetBatcher();
 
   float trackWidth = canvas->GetWorldWidth();
-  const TimeGraphLayout& layout = time_graph_->GetLayout();
   const bool picking = picking_mode != PickingMode::kNone;
 
   pos_[0] = canvas->GetWorldTopLeftX();
@@ -33,30 +32,18 @@ void GraphTrack::Draw(GlCanvas* canvas, PickingMode picking_mode) {
 
   Box box(pos_, Vec2(size_[0], -size_[1]), track_z);
 
-  // Draw label with graph value at current mouse position.
-  const Color kWhiteColor(255, 255, 255, 255);
-  const Color kBlackColor(0, 0, 0, 255);
-  double graph_value = GetValueAtTime(time_graph_->GetCurrentMouseTimeNs());
-
-  int white_text_box_width =
-      canvas->GetTextRenderer().GetStringWidth(std::to_string(graph_value).c_str());
-
-  Vec2 white_text_box_size(white_text_box_width, layout.GetTextBoxHeight());
-  auto rightmost_x_position = canvas->GetWorldTopLeftX() + canvas->GetWorldWidth() -
-                              layout.GetRightMargin() - layout.GetSliderWidth() -
-                              white_text_box_size[0];
-  Vec2 white_text_box_position(
-      std::min(canvas->GetMouseX(), rightmost_x_position),
-      y0 - static_cast<float>((max_ - graph_value) * size_[1] / value_range_) -
-          white_text_box_size[1] / 2.f);
-
   if (!picking) {
-    canvas->GetTextRenderer().AddText(
-        std::to_string(graph_value).c_str(), white_text_box_position[0],
-        white_text_box_position[1] + layout.GetTextOffset(), GlCanvas::kZValueTextUi, kBlackColor,
-        time_graph_->CalculateZoomedFontSize(), white_text_box_size[0]);
-    Box white_text_box(white_text_box_position, white_text_box_size, GlCanvas::kZValueUi);
-    batcher->AddBox(white_text_box, kWhiteColor);
+    auto last_measure = GetPreviousValueAndTime(time_graph_->GetCurrentMouseTimeNs());
+    if (last_measure.has_value()) {
+      auto& [time, value] = last_measure.value();
+      if (time_graph_->IsFullyVisible(time, time)) {
+        float point_x = time_graph_->GetWorldFromTick(time);
+        float point_y = y0 - static_cast<float>((max_ - value) * size_[1] / value_range_);
+        const Color kBlack(0, 0, 0, 255);
+        const Color kWhite(255, 255, 255, 255);
+        DrawLabel(canvas, Vec2(point_x, point_y), std::to_string(value), kBlack, kWhite);
+      }
+    }
   }
 
   batcher->AddBox(box, color, shared_from_this());
@@ -69,6 +56,7 @@ void GraphTrack::Draw(GlCanvas* canvas, PickingMode picking_mode) {
   batcher->AddLine(Vec2(x1, y1), Vec2(x0, y1), track_z, color, shared_from_this());
 
   const Color kLineColor(0, 128, 255, 128);
+  const Color kDotColor(0, 128, 255, 255);
 
   if (!picking) {
     // Current time window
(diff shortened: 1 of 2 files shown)


OrbitGl/GraphTrack.h
```diff
@@ -19,9 +19,10 @@ class GraphTrack : public Track {
   void Draw(GlCanvas* canvas, PickingMode /*picking_mode*/) override;
   [[nodiscard]] float GetHeight() const override;
   void AddValue(double value, uint64_t time);
-  [[nodiscard]] double GetValueAtTime(uint64_t time, double default_value = 0) const;
+  bool GetPreviousValue(uint64_t time, uint64_t* prev_time, double* value_at_time) const;
```
florian-kuebler, line 22: should be [[nodiscard]], right?

karupayun: Yes

OrbitGl/GraphTrack.cpp
```diff
@@ -37,30 +36,17 @@ void GraphTrack::Draw(GlCanvas* canvas, PickingMode picking_mode) {
...
-        white_text_box_position[1] + layout.GetTextOffset(), GlCanvas::kZValueTextUi, kBlackColor,
-        time_graph_->CalculateZoomedFontSize(), white_text_box_size[0]);
-    Box white_text_box(white_text_box_position, white_text_box_size, GlCanvas::kZValueUi);
-    batcher->AddBox(white_text_box, kWhiteColor);
+    double graph_value;
+    uint64_t graph_time;
+    bool is_there_a_value =
+        GetPreviousValue(time_graph_->GetCurrentMouseTimeNs(), &graph_time, &graph_value);
+    if (is_there_a_value) {
+      auto graph_y_pos = y0 - static_cast<float>((max_ - graph_value) * size_[1] / value_range_);
```
pierricgimmig, line 45: This should be computed inside the following if statement.

OrbitGl/GraphTrack.cpp
```diff
@@ -37,30 +36,17 @@ void GraphTrack::Draw(GlCanvas* canvas, PickingMode picking_mode) {
...
-    canvas->GetTextRenderer().AddText(
-        std::to_string(graph_value).c_str(), white_text_box_position[0],
-        white_text_box_position[1] + layout.GetTextOffset(), GlCanvas::kZValueTextUi, kBlackColor,
-        time_graph_->CalculateZoomedFontSize(), white_text_box_size[0]);
-    Box white_text_box(white_text_box_position, white_text_box_size, GlCanvas::kZValueUi);
-    batcher->AddBox(white_text_box, kWhiteColor);
+    double graph_value;
+    uint64_t graph_time;
+    bool is_there_a_value =
+        GetPreviousValue(time_graph_->GetCurrentMouseTimeNs(), &graph_time, &graph_value);
```
pierricgimmig, line 43: "is_there_a_value" is not a great name, "has_value_at_time" maybe? Actually, I think the interface would be cleaner here is we used a std::optional. The function could return a std::optional<std::pair<uint64_t, double>> for example. We could then replace "GetPreviousValue" by "GetPointForTime(...)" maybe, wdyt?

OrbitGl/GraphTrack.cpp
```diff
@@ -86,26 +72,73 @@ void GraphTrack::Draw(GlCanvas* canvas, PickingMode picking_mode) {
     if (it != values_.begin()) --it;
     uint64_t previous_time = it->first;
     double last_normalized_value = (it->second - min_) * inv_value_range_;
+    float kDotRadius = 2.5;
```
pierricgimmig, line 75: constexpr

OrbitGl/GraphTrack.cpp
```diff
@@ -86,26 +72,73 @@ void GraphTrack::Draw(GlCanvas* canvas, PickingMode picking_mode) {
...
       double normalized_value = (it->second - min_) * inv_value_range_;
-      float base_y = pos_[1] - size_[1];
       float x0 = time_graph_->GetWorldFromTick(previous_time);
       float x1 = time_graph_->GetWorldFromTick(time);
       float y0 = base_y + static_cast<float>(last_normalized_value) * size_[1];
       float y1 = base_y + static_cast<float>(normalized_value) * size_[1];
-      time_graph_->GetBatcher().AddLine(Vec2(x0, y0), Vec2(x1, y0), GlCanvas::kZValueText,
-                                        kLineColor);
-      time_graph_->GetBatcher().AddLine(Vec2(x1, y0), Vec2(x1, y1), GlCanvas::kZValueText,
+      // TODO: Draw only the dots when it isn't a continuous variable?
```
pierricgimmig, line 89: Could we keep both implementations in prevision of having some UI to toggle between both?

OrbitGl/GraphTrack.cpp
```diff
@@ -86,26 +72,73 @@ void GraphTrack::Draw(GlCanvas* canvas, PickingMode picking_mode) {
...
   }
 }
 
+void GraphTrack::DrawWhiteLabel(GlCanvas* canvas, double graph_value, uint64_t graph_time,
+                                float graph_y_pos) {
+  const TimeGraphLayout& layout = time_graph_->GetLayout();
+
+  // Draw label with graph value at current mouse position.
+  const Color kWhite(255, 255, 255, 255);
+  const Color kBlack(0, 0, 0, 255);
```
pierricgimmig, line 107: I'd pass those as parameters, or as member variables of a "Label" class.

OrbitGl/GraphTrack.cpp
```diff
@@ -86,26 +72,73 @@ void GraphTrack::Draw(GlCanvas* canvas, PickingMode picking_mode) {
...
+
+  int text_width = canvas->GetTextRenderer().GetStringWidth(std::to_string(graph_value).c_str());
+  Vec2 white_text_box_size(text_width, layout.GetTextBoxHeight());
+  Vec2 value_in_screen(time_graph_->GetWorldFromTick(graph_time), graph_y_pos);
```
pierricgimmig, line 111: value_in_screen is misleading, we are calling GetWorldFromTick for the x component, which makes me think that we are not in screen space but in world space. I would replace by "pos", or "target_pos" maybe?

OrbitGl/GraphTrack.cpp
```diff
@@ -86,26 +72,73 @@ void GraphTrack::Draw(GlCanvas* canvas, PickingMode picking_mode) {
...
+
+  auto arrow_width = white_text_box_size[1] / 2;
```
pierricgimmig, line 113: auto -> float

pierricgimmig: 2 -> 2.f

pierricgimmig (commented):
Hi Pablo, I think visually having points looks good, but I'm worried about performance, please see comments.  Thanks!

OrbitGl/GraphTrack.cpp
```diff
@@ -86,26 +72,73 @@ void GraphTrack::Draw(GlCanvas* canvas, PickingMode picking_mode) {
...
       float y0 = base_y + static_cast<float>(last_normalized_value) * size_[1];
       float y1 = base_y + static_cast<float>(normalized_value) * size_[1];
-      time_graph_->GetBatcher().AddLine(Vec2(x0, y0), Vec2(x1, y0), GlCanvas::kZValueText,
-                                        kLineColor);
-      time_graph_->GetBatcher().AddLine(Vec2(x1, y0), Vec2(x1, y1), GlCanvas::kZValueText,
+      // TODO: Draw only the dots when it isn't a continuous variable?
+      time_graph_->GetBatcher().AddLine(Vec2(x0, y0), Vec2(x1, y1), GlCanvas::kZValueText,
                                         kLineColor);
+      time_graph_->GetBatcher().AddCircle(Vec2(x1, y1), kDotRadius, GlCanvas::kZValueText,
+                                          kLineColor);
```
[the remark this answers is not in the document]
karupayun: AddSquareDot is done!

OrbitGl/GraphTrack.cpp
karupayun: GetLastPointBeforeTime

OrbitGl/GraphTrack.cpp
karupayun: Done!

OrbitGl/GraphTrack.cpp
```diff
@@ -86,26 +72,73 @@ void GraphTrack::Draw(GlCanvas* canvas, PickingMode picking_mode) {
...
 
       previous_time = time;
       last_normalized_value = normalized_value;
     }
   }
 }
 
+void GraphTrack::DrawWhiteLabel(GlCanvas* canvas, double graph_value, uint64_t graph_time,
```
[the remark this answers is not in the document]
karupayun: Sure, WDYT now?

OrbitGl/GraphTrack.cpp
karupayun: GetPointForTime maybe is a little misleading, because is returning a time (related with x-coordinate) but a value who doesn't seem to have relation with y-position. I called it GetLastMeasure, what do you think?

OrbitGl/GraphTrack.cpp
karupayun: What do you mean? Comment these lines?

OrbitGl/GraphTrack.cpp
karupayun: Sure!

karupayun:
Well, I think that it's ready to review. 

It's not perfect:
1- Has the methods DrawLabel and DrawSquareDot that could be in some class to be reusable. We could discuss tomorrow in the daily.

2- We have to include something on what to do after zooming-out a lot in a long capture: We have too many points being drawn.
* Performance problem (which I'm really not too worried)
* Visual problem (too many points in similar x-coordinate).

I think that after zooming-out a lot, we have to draw less points or no points at all.

I will investigate that.

OrbitGl/GraphTrack.cpp
pierricgimmig: I would remove the comment.

OrbitGl/GraphTrack.cpp
```diff
@@ -86,26 +73,72 @@ void GraphTrack::Draw(GlCanvas* canvas, PickingMode picking_mode) {
...
   }
 }
 
+void GraphTrack::DrawSquareDot(GlCanvas* canvas, Vec2 center, float radius, float z, Color color) {
+  Vec2 position(center[0] - radius, center[1] - radius);
+  Vec2 size(2 * radius, 2 * radius);
+  canvas->GetBatcher()->AddBox(Box(position, size, z), color);
+}
+
+void GraphTrack::DrawLabel(GlCanvas* canvas, Vec2 target_pos, std::string text, Color text_color,
```
pierricgimmig, line 107: const std::string& text ?

OrbitGl/GraphTrack.cpp
```diff
@@ -86,26 +73,72 @@ void GraphTrack::Draw(GlCanvas* canvas, PickingMode picking_mode) {
...
+
+  canvas->GetTextRenderer().AddText(text.c_str(), text_box_position[0],
+                                    text_box_position[1] + layout.GetTextOffset(),
+                                    GlCanvas::kZValueTextUi, text_color,
+                                    time_graph_->CalculateZoomedFontSize(), text_box_size[0]);
+
+  Box arrow_text_box(text_box_position, text_box_size, GlCanvas::kZValueUi);
+  Vec3 arrow_extra_point(target_pos[0], target_pos[1], GlCanvas::kZValueUi);
+
+  auto batcher = canvas->GetBatcher();
```
pierricgimmig, line 129: Batcher* batcher

OrbitGl/GraphTrack.h
```diff
@@ -19,9 +20,12 @@ class GraphTrack : public Track {
   void Draw(GlCanvas* canvas, PickingMode /*picking_mode*/) override;
   [[nodiscard]] float GetHeight() const override;
   void AddValue(double value, uint64_t time);
-  [[nodiscard]] double GetValueAtTime(uint64_t time, double default_value = 0) const;
+  [[nodiscard]] std::optional<std::pair<uint64_t, double> > GetLastMeasure(uint64_t time) const;
```
pierricgimmig, line 23: nit. I'm not sure about "Measure", I don't think it represents what is being returned. GetLastPoint maybe? Or GetPointForTime?

pierricgimmig (approved):
Thanks for the change Pablo!

OrbitGl/GraphTrack.h
karupayun: I don't like point, because it isn't a point. It's a pair of (value, time), where after we can make some mathematical operations to get the x-y coordinates of a future point. So, for me it's the last measure. 

What do you think about GetPreviousValueAndTime?

karupayun:
A nit, I modified the alpha value of the blue dots.

PR #1355, opened 2020-10-12
Web-based Disassembly View Implementation

beckerhe:
This introduces a web-based disassembly view component leveraging PrismJS - a JavaScript syntax highlighting library - and QWebEngine - Qt's Chromium integration.

After the change the disassembly view looks like this:
image
image


OrbitQt/SourceCodeView.cpp
```diff
@@ -0,0 +1,49 @@
...
+  if (title != title_ || new_code != source_code_ || language != language_) {
+    title_ = std::move(title);
+    source_code_ = std::move(new_code);
+    language_ = std::move(language);
+
+    emit sourceCodeChanged();
+  }
+}
+
+void SourceCodeView::SetHeatmapData(QVector<int> hit_counts, QVector<double> hit_ratios,
```
florian-kuebler, line 21: How is this data organized? Is this lines to hit_count/ratio/total_ratio?

OrbitQt/SourceCodeView.cpp
```diff
@@ -0,0 +1,49 @@
...
+    hit_counts_ = std::move(hit_counts);
+    hit_ratios_ = std::move(hit_ratios);
+    total_hit_ratios_ = std::move(total_hit_ratios);
+
+    emit sourceCodeChanged();
+  }
+}
+
+int SourceCodeView::exec() {
+  dialog_.resize(QSize{820, 600});
```
florian-kuebler, line 34: Should this be a constant?

OrbitQt/SourceCodeView.cpp
```diff
@@ -0,0 +1,49 @@
...
+}
+
+SourceCodeView::SourceCodeView(QWebEngineProfile* profile, QObject* parent)
+    : QObject(parent), dialog_(profile) {
+  dialog_.RegisterObject("view", this);
+  const auto listen_port = dialog_.GetWebSocketServer()->serverPort();
+  LOG("SourceCodeView's websocket server is listening on port %d.", listen_port);
+
+  dialog_.GetWebEnginePage()->load(
+      QUrl{QString{"qrc:///WebUI/SourceCodeView/index.html?websocket_port=%1"}.arg(listen_port)});
```
florian-kuebler, line 46: Again, consider this being a constant?

OrbitQt/orbitmainwindow.cpp
```diff
@@ -631,15 +644,61 @@ void OrbitMainWindow::OpenCapture(const std::string& filepath) {
...
-void OrbitMainWindow::OpenDisassembly(std::string a_String, DisassemblyReport report) {
-  auto* dialog = new OrbitDisassemblyDialog(this);
-  dialog->SetText(std::move(a_String));
-  dialog->SetDisassemblyReport(std::move(report));
-  dialog->setWindowTitle("Orbit Disassembly");
-  dialog->setAttribute(Qt::WA_DeleteOnClose);
-  dialog->setWindowFlags(dialog->windowFlags() | Qt::WindowMinimizeButtonHint |
-                         Qt::WindowMaximizeButtonHint);
-  dialog->show();
+static std::pair<QString, QString> SplitLineAtFirstLinebreak(const std::string& input) {
```
florian-kuebler, line 647: Let us take this moment to actually change the way we store the disassambly/source-code, such that it has a title and a body/code part.

florian-kuebler: Then you also don't need to do all that shifting with the line numbers. 
If you wish, I can also do this as a follow-up pr, but shouldn't be to hard to achieve.

OrbitQt/SourceCodeView.h
```diff
@@ -0,0 +1,86 @@
...
+
+ private:
+  web_engine::Dialog dialog_;
+
+  QString title_;
+  QString source_code_;
+  QString language_;
+  bool line_numbers_enabled_ = false;
+  bool heatmap_enabled_ = false;
+  QVector<int> hit_counts_;
```
florian-kuebler, line 80: Shouldn't we be able to store the DisassemblyReport here, and then just have functions for the javascript part, to retrieve these numbers?

OrbitQt/orbitmainwindow.cpp
```diff
@@ -631,15 +644,61 @@ void OrbitMainWindow::OpenCapture(const std::string& filepath) {
...
+  QVector<double> hit_ratios;
+  hit_ratios.reserve(number_of_code_lines);
+
+  QVector<double> total_hit_ratios;
+  total_hit_ratios.reserve(number_of_code_lines);
+
+  const auto total_number_of_samples = report.GetNumSamples();
+  const auto number_of_samples_in_function = report.GetNumSamplesInFunction();
+
+  for (int i = 0; i < number_of_code_lines; i++) {
```
florian-kuebler, line 687: As said above, I would rather store the disassembly report into the Codeview (I think there is a super-class).

WebUI/src/x86asm.ts
```diff
@@ -0,0 +1,30 @@
+// Copyright (c) 2020 The Orbit Authors. All rights reserved.
```
florian-kuebler, line 1: Did you crafted this by your own, or is this some third_party script?

WebUI/src/x86asm.test.js
```diff
@@ -2,8 +2,10 @@
...
 
-import QWebChannel from 'qtwebchannel/qwebchannel';
+import x86asm from './x86asm';
 
-const el = document.createElement("div");
-el.innerHTML = "<h1>Hello, World!</h1>";
-document.body.appendChild(el);
+test("x86asm language", () => {
+
+    expect(typeof x86asm).toBe("object");
```
florian-kuebler, line 9: I assume there will be tests following?

WebUI/src/SourceCodeView.css
```diff
@@ -0,0 +1,70 @@
...
+    --feature-color: #19181a;
+}
+
+html,
+body {
+    margin: 0;
+    padding: 0;
+    border: 0;
+    background-color: #2d2d2d;
+    font-family: Arial, Helvetica, sans-serif;
```
florian-kuebler, line 17: can we ensure, to use the same font (and size) as in the rest of orbit?

WebUI/src/Heatmap.ts
```diff
@@ -0,0 +1,115 @@
+// Copyright (c) 2020 The Orbit Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import "./Heatmap.css";
+import Prism from "prismjs";
+
+export default function (view: any) {
```
florian-kuebler, line 8: Thanks for converting my mess to TS! Looks good, as far as I can tell.

OrbitQt/SourceCodeView.h
beckerhe: I changed that for two reasons:
1. I replaced function calls by Q_PROPERTY to not have 3 function calls per line which lead to a lot of webchannel messages. Pushing all values in one message to the JavaScript context is a lot faster.
2. When I made that line number mangling (which you already pointed out) I wanted to have all the code which is messing with the line numbers in one place. That's why the DisassemblyReport is not part of this class anymore.

I think we should talk about this on how to move forward here. I'm happy to change things for better. :smile:

WebUI/src/x86asm.ts
beckerhe: Proudly self-made. :grin: But it's also not a big deal. The main task was getting a list of all x86 instructions and all registers.

WebUI/src/x86asm.test.js
beckerhe: Probably. I still don't know what and how, but I hope to figure that out.

OrbitQt/SourceCodeView.cpp
beckerhe: as discussed offline: obsolete

OrbitQt/orbitmainwindow.cpp
beckerhe: completely removed! thanks!

OrbitQt/SourceCodeView.h
beckerhe: I will resolve this for now. I think you need to have a fresh look at this anyway.

OrbitQt/orbitmainwindow.cpp
beckerhe: Resolving this for now since this radically changed.

WebUI/src/SourceCodeView.css
beckerhe: on it. there is more that needs to be done. The button and the scroll bar for example.

WebUI/src/Heatmap.ts
beckerhe: Glad you like it. :-)

WebUI/src/SourceCodeView.css
florian-kuebler: do you want to address this in this PR?

beckerhe: Yes. The are still quite some (minor) changes coming based on the feedback on Monday. I will mark it ready for review when I'm done.

OrbitQt/SourceCodeView.cpp
beckerhe: Done. Thanks.

OrbitQt/SourceCodeView.cpp
beckerhe: Done. Thanks!

WebUI/src/SourceCodeView.css
beckerhe: The style is now adjusted to look more or less like Orbit.
I'm a bit hesitant to make the font dynamicly adjusting since this
might affect the layout and result in a bigger problem than a font
that is a little bit off.

florian-kuebler: okay. thanks anyways.

OrbitQt/SourceCodeView.cpp
florian-kuebler: thanks

OrbitQt/SourceCodeView.h
```diff
@@ -0,0 +1,92 @@
...
+   are passed through to QWebEngine::Dialog. Check out this class's documentation
+   for more details.
+
+   Example:
+   QWebEngineProfile profile;
+   SourceCodeView view{&profile, std::nullopt};
+   view.SetSourceCode("example.cpp", "int main() {}", "cpp");
+   view.exec(); // Will open the dialog and block until closed.
+*/
+class SourceCodeView : public QObject {
```
florian-kuebler, line 35: should we rename it to CodeView (as it is also used for disassembly)

florian-kuebler (approved):
Went trough the recent changes and in general looks good to me.

OrbitQt/SourceCodeView.h
beckerhe: That's an excellent idea! I always was uncomfortable with the name but didn't come up with a better one.

florian-kuebler: Haha, yeah, sometimes the obvious ones might just work :)

beckerhe: So I renamed everything to CodeView. Thanks!

reichlfl (approved):
So cool to see it come to live. This is really great work, thanks a lot.

I've only looked at the typescript and css part since Florian already gave a thorough review - it's mostly nitpicks and personal preference as food for thoughts, feel free to ignore or point out that this is not how "the internet" usually handles it.

WebUI/src/CodeView.css
```diff
@@ -0,0 +1,80 @@
...
+footer {
+    bottom: 0;
+    text-align: right;
+    display: flex;
+    align-items: center;
+    justify-content: flex-end;
+    flex-direction: row;
+}
+
+/*#close {
```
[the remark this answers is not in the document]
beckerhe: No at all. This was an oversight. Thanks for spotting this!

WebUI/src/Heatmap.ts
```diff
@@ -0,0 +1,129 @@
...
+    const highlights = pre.querySelector(".line-heatmap-highlight-wrapper");
+    if (highlights) {
+      highlights.remove();
+    }
+  });
+
+  Prism.hooks.add("complete", (env) => {
+    const pre = <HTMLElement>env.element.parentNode;
+
+    if (
```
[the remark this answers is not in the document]
beckerhe: Good point. We can use TS's safe navigation operator in this case.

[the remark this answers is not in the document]
beckerhe: Done.

WebUI/src/CodeView.css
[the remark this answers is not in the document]
beckerhe: Removed.

WebUI/src/Heatmap.ts
```diff
@@ -0,0 +1,129 @@
...
+    pre.appendChild(heatmap_wrapper);
+
+    const line_highlight_wrapper = document.createElement("div");
+    line_highlight_wrapper.className = "line-heatmap-highlight-wrapper";
+
+    const hit_ratio_wrapper = pre.querySelector("#hit_ratios");
+
+    view.hit_ratios
+      .map((hit_ratio: number, idx: number) => {
+        if (!hit_ratio || hit_ratio == 0) {
```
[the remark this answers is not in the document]
beckerhe: Good point. We don't even need type-safe comparison here. I removed the !hit_ratio completely.

WebUI/src/Heatmap.ts
```diff
@@ -0,0 +1,129 @@
...
+
+        const hit_ratio_sqrt = Math.sqrt(hit_ratio);
+        highlight_line.style.background = `linear-gradient(to left, rgba(var(--accent-color), ${hit_ratio_sqrt}) 0%, rgba(var(--accent-color), ${
+          hit_ratio_sqrt / 2
+        }) 12%, rgba(var(--accent-color), 0) 90%)`;
+
+        return highlight_line;
+      })
+      .forEach(
+        (el: HTMLElement | null) => el && line_highlight_wrapper.appendChild(el)
```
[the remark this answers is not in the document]
beckerhe: Sounds. reasonable. Replaced by filter and forEach!

WebUI/src/Heatmap.ts
```diff
@@ -0,0 +1,129 @@
...
+          hit_ratio_wrapper.children[idx + 1]
+        );
+
+        const highlight_line = document.createElement("div");
+        highlight_line.classList.add("line-heatmap-highlight");
+        highlight_line.style.top = `${hit_ratio_element.offsetTop}px`;
+        highlight_line.style.height = `${hit_ratio_element.offsetHeight}px`;
+
+        const hit_ratio_sqrt = Math.sqrt(hit_ratio);
+        highlight_line.style.background = `linear-gradient(to left, rgba(var(--accent-color), ${hit_ratio_sqrt}) 0%, rgba(var(--accent-color), ${
```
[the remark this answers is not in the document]
beckerhe: Yeah, I also feel not super comfortable with that part. I took that from Florian and he saw that in another prism plugin. I guess right now it's the best option. Probably we can do better though by having a general gradient object and just change the alpha value per line.

Nevertheless I'd prefer to leave it for now.

WebUI/src/Heatmap.ts
```diff
@@ -0,0 +1,129 @@
...
+          const span = document.createElement("span");
+          span.textContent = format(value);
+          return span;
+        })
+        .forEach((child: HTMLSpanElement) => wrapper.appendChild(child));
+
+      heatmap_wrapper.appendChild(wrapper);
+    };
+
+    view.hit_counts &&
```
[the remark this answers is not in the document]
beckerhe: Fixed. Thanks!

PR #1356, opened 2020-10-12
Add generate_coverage_report.sh and conan profile

antonrohr:
I hope this is pretty much self explanatory. First Orbit needs to be built with the included conan profile, which adds the required build arguments. And then the script is used like this: (Assuming its executed from Orbit root directory)
```
./build.sh coverage_clang9_relwithdebinfo
mkdir build_coverage_clang9_relwithdebinfo/coverage
contrib/unit_test_coverage/generate_coverage_report.sh . build_coverage_clang9_relwithdebinfo build_coverage_clang9_relwithdebinfo/coverage
```
This creates the coverage report at `build_coverage_clang9_relwithdebinfo/coverage/index.html`

I am not experienced in bash scripts, so any advice regarding best practices and similar, is appreciated.


contrib/unit_test_coverage/generate_coverage_report.sh
```diff
@@ -0,0 +1,133 @@
...
+# found in the LICENSE file.
+
+# --- generate_coverage_report.sh ---
+#
+# This script generates a coverage report for the Orbit repository. It uses 
+# llvm-cov and requires a build with build arguments -fprofile-instr-generate 
+# and -fcoverage-mapping. 
+#
+# usage: generate_coverage_report.sh SOURCE_DIR BUILD_DIR OUTPUT_DIR
+#
```
florian-kuebler, line 14: Can you remove that #, or add a new line. Afterwards you have the constants of that file rather then a description of the script.

contrib/unit_test_coverage/generate_coverage_report.sh
```diff
@@ -0,0 +1,133 @@
...
+BINS=$(find $BUILD_DIR/bin/ -maxdepth 1 -type f | grep -v "$BIN_FILE_FILTER")
+FORMATTED_BINARY_FILES=$(echo $LIBS $BINS | sed "s/ / -object=/g")
+SUM_FUNCTIONS_COUNT=0
+SUM_FUNCTIONS_COVERED=0
+SUM_LINES_COUNT=0
+SUM_LINES_COVERED=0
+SUM_REGIONS_COUNT=0
+SUM_REGIONS_COVERED=0
+
+for COMPONENT in $(cd "$BUILD_DIR"; ctest --show-only=json-v1 | jq -r '.tests | .[] | .name')
```
florian-kuebler, line 81: Personally, I'd appreciated a brief comment, what you are iterating here exactly. In particular the piped part. Maybe also move the "cd" out of the loop.

florian-kuebler (approved):
So for me looks good. But, as I said offline, I am not an bash expert. I added some nitpicking comments, feel free to ignore.
Thanks!

contrib/unit_test_coverage/generate_coverage_report.sh
```diff
@@ -0,0 +1,133 @@
...
+fi
+
+BUILD_DIR=$(realpath $2)
+if [ ! -d $BUILD_DIR ]; then
+  echo "error: directory $2 does not exist."
+  exit 1
+fi
+
+OUTPUT_DIR=$(realpath $3)
+if [ ! -d $OUTPUT_DIR ]; then
```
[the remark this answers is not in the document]
beckerhe: I would even argue that we might wanna have that check the other way around. If that script is generating more than one file I would prefer that the script fails when I accidentally pass in an existing directory, to avoid unintended overwrites or that an existing directory is cluttered with unrelated files. WDYT?

contrib/unit_test_coverage/generate_coverage_report.sh
```diff
@@ -0,0 +1,133 @@
...
+SUM_LINES_COVERED=0
+SUM_REGIONS_COUNT=0
+SUM_REGIONS_COVERED=0
+
+for COMPONENT in $(cd "$BUILD_DIR"; ctest --show-only=json-v1 | jq -r '.tests | .[] | .name')
+do
+  echo "$COMPONENT: preprocessing"
+  llvm-profdata merge -sparse \
+    -o $BUILD_DIR/$COMPONENT.profdata \
+    $BUILD_DIR/$COMPONENT/default.profraw
```
beckerhe, line 86: I guess when we compile with clang-9 we also need to use llvm tools from LLVM-9, don't we? So maybe we need `llvm-profdata-9`?

contrib/unit_test_coverage/generate_coverage_report.sh
```diff
@@ -0,0 +1,133 @@
...
+
+  SOURCE_FILES=$(find $SOURCE_DIR/$COMPONENT -regex ".*\.\(h\|cpp\)" | grep -v "$SOURCE_FILE_FILTER")
+
+  echo "$COMPONENT: generating html report"
+  llvm-cov show -format=html \
+    -show-line-counts-or-regions -show-instantiations=false \
+    -output-dir=$OUTPUT_DIR/coverage_$COMPONENT \
+    -instr-profile=$BUILD_DIR/$COMPONENT.profdata \
+    $FORMATTED_BINARY_FILES \
+    $SOURCE_FILES >/dev/null 2>&1
```
beckerhe, line 96: See previous comment: `llvm-cov-9`?

contrib/unit_test_coverage/generate_coverage_report.sh
```diff
@@ -0,0 +1,133 @@
...
+
+  echo "$COMPONENT: generating json summary"
+  JSON=$(llvm-cov export -summary-only \
+    -instr-profile=$BUILD_DIR/$COMPONENT.profdata \
```
beckerhe, line 100: llvm-cov-9?

third_party/conan/configs/linux/profiles/coverage_clang9_relwithdebinfo
```diff
@@ -0,0 +1,24 @@
...
+build_type=RelWithDebInfo
+[options]
+[build_requires]
+&: cmake_installer/3.16.3@conan/stable
+[env]
+CC=$C_COMPILER
+CXX=$CXX_COMPILER
+CFLAGS=$C_FLAGS
+CXXFLAGS=$CXX_FLAGS
+LDFLAGS=$LINKER_FLAGS
```
beckerhe, line 24: The profiles are written to be easily extended by other profile. Maybe it makes sense to write an extending profile, something like this:

```
include(clang9_relwithdebinfo)

[env]
CFLAGS=$CFLAGS -fcoverage-mapping
CXXFLAGS=$CXXFLAGS -fcoverage-mapping

**Review by beckerhe — COMMENTED** [review]
This looks really good. I just added some minor comments.

File: contrib/unit_test_coverage/generate_coverage_report.sh (hunk 3 above)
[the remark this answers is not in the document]
Comment (florian-kuebler) [reply contrib/unit_test_coverage/generate_coverage_report.sh:66]: Kind of agree, but this depends a bit on two things:
1. how does this integrate with the usage scenario we have.
2. if we only add additional data, it might be fine, to have the directory existing.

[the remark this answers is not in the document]
Comment (beckerhe) [reply contrib/unit_test_coverage/generate_coverage_report.sh:66]: Good point. The more I'm thinking about that the less I'm sure what to do here. In general I would consider this script somehow work-in-progress. Something we can iterate on.

My vision for the developer would be to have a single command. Something like `./build.sh coverage` which builds Orbit with the correct flags, runs the tests, and also generates the coverage report. But this means cmake needs to call that script. Before having looked into that I don't know what is the "right" thing to do for the script. So I would argue we can also change that later.

So Anton from my side feel free to take any decision on this. :sweat_smile:

File: contrib/unit_test_coverage/generate_coverage_report.sh (hunk 2 above)
Comment (antonrohr) [reply contrib/unit_test_coverage/generate_coverage_report.sh:81]: I added a comment to clarify, but I like the `cd` as part of this command. 

I am doing the cd, because ctest needs to be run in the build directory. I would prefer to tell ctest the build dir as a parameter, but I was not able to figure out how. I feel the way I did it is the next best thing to that, because it does not actually change the working directory of the script, but only for the subshell this command is executed in.

File: third_party/conan/configs/linux/profiles/coverage_clang9_relwithdebinfo (hunk 7 above)
Comment (antonrohr) [reply third_party/conan/configs/linux/profiles/coverage_clang9_relwithdebinfo]: I did not know this was possible, neat!

File: contrib/unit_test_coverage/generate_coverage_report.sh (hunk 6 above)
Comment (antonrohr) [reply contrib/unit_test_coverage/generate_coverage_report.sh]: done

File: contrib/unit_test_coverage/generate_coverage_report.sh (hunk 5 above)
Comment (antonrohr) [reply contrib/unit_test_coverage/generate_coverage_report.sh]: done

File: contrib/unit_test_coverage/generate_coverage_report.sh (hunk 4 above)
Comment (antonrohr) [reply contrib/unit_test_coverage/generate_coverage_report.sh]: good point. Done

File: contrib/unit_test_coverage/generate_coverage_report.sh (hunk 3 above)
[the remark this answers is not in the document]
Comment (antonrohr) [reply contrib/unit_test_coverage/generate_coverage_report.sh]: I am also not 100% on this aka this will be changed when necessary, but I decided to go a completely different way, so please let me know what you think. 

The output directory is now always BUILD_DIR/coverage. I also added a echo at the end of the script which tells the user where to find the report

File: contrib/unit_test_coverage/generate_coverage_report.sh [hunk 8]
```diff
@@ -0,0 +1,133 @@
...
+# grep inverted matching list for files in BUILD_DIR/bin/ that are not used
+BIN_FILE_FILTER="\.debug\|crashpad_handler"
+# grep inverted matching list for source files that are not used
+SOURCE_FILE_FILTER="Test.cpp\|Tests.cpp\|Fuzzer.cpp"
+# ------------------------------------
+
+# fail on any error
+set -e
+
+function get_css_class_string {
```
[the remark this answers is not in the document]
antonrohr: done

antonrohr:
I addressed the review commits and also tested results and timing of relwithdebinfo vs debug vs release. 

The build times of relwithdebinfo and debug is pretty similar on my machine, release takes longer. 
`generate_coverage_report.sh` execution times are very similar in all three modes. Approx 22seconds.

The coverage report for debug varies slightly to release and reliwithdebinfo. There are some `#ifndef NDEBUG` and this explains why the debug build contains more lines than release. Strangely debug also contains more Regions than release and I am not sure if that is because of the ifndef.

I left the conan profile based on relwithdebinfo for now.

beckerhe (approved):
LGTM! Thanks!

contrib/unit_test_coverage/generate_coverage_report.sh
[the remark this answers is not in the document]
florian-kuebler: Seems legit. Thanks

contrib/unit_test_coverage/generate_coverage_report.sh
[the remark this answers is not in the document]
florian-kuebler: thanks

Issue #1327, opened 2020-10-01
Orbit installation hangs on test #12


jhdcs:
I am trying to install the latest version of Orbit on VirtualBox VM running Ubuntu 20.04. I believe I have all the required dependencies, as the build process seems to run just fine. However, during testing, the `bootstrap_orbit.sh` script gets to test 12 (OrbitQt), and then just hangs. Even after 2 hours, nothing has happened.

Are there logs saved anywhere that can show what's going on internally? I'm not sure what information I can provide to help you debug... All it says is `Start 12: OrbitQT`, and then nothing.

florian-kuebler:
Hi, thanks for reaching out.

We are aware of that issue, and are working on a fix of that. However, it should not effect your actual build or your experience with Orbit.
At that point in time, the complete build should be done already and you should find the binaries under `build_default_relwithdebinfo/bin/Orbit` (or OrbitService respectively).
So you should be able to use those.

Feel free to reach out again, if you have any questions.

(For internal reference, we track this issue as http://b/167365568)

illwieckz:
I got similar issue but on test `#11`:

```
      Start  1: OrbitLinuxTracing
      Start  2: OrbitService
      Start  3: ElfUtils
      Start  4: OrbitBase
      Start  5: OrbitClientData
      Start  6: OrbitClientModel
      Start  7: OrbitCore
      Start  8: OrbitFramePointerValidator
 1/12 Test  #1: OrbitLinuxTracing ................   Passed    0.43 sec
      Start  9: OrbitGl
 2/12 Test  #4: OrbitBase ........................   Passed    0.30 sec
      Start 10: OrbitGgp
 3/12 Test  #5: OrbitClientData ..................   Passed    0.31 sec
      Start 11: OrbitSsh
 4/12 Test  #6: OrbitClientModel .................   Passed    0.30 sec
      Start 12: OrbitQt
 5/12 Test  #9: OrbitGl ..........................   Passed    0.12 sec
 6/12 Test #10: OrbitGgp .........................   Passed    0.12 sec
 7/12 Test #11: OrbitSsh .........................   Passed    0.06 sec
 8/12 Test  #7: OrbitCore ........................   Passed    0.45 sec
 9/12 Test  #8: OrbitFramePointerValidator .......   Passed    0.36 sec
10/12 Test  #3: ElfUtils .........................   Passed    0.67 sec
11/12 Test  #2: OrbitService .....................   Passed    1.15 sec
```

This is stuck since hours now.

[...]

PR #1358, opened 2020-10-13
Test isolation for gtest

beckerhe:
This adds test isolation support for gtest
by leveraging custom modifications done by the Android team.

Test isolation won't be enabled by default for now.

This also updates the clang-format presubmit check which has not been excluding
third_party/ from checks before.

Fixes: #1327 
Bug: http://b/167365568

### DIFF
--- a/cmake/tests.cmake
+++ b/cmake/tests.cmake
@@ -2,39 +2,56 @@
 # Use of this source code is governed by a BSD-style license that can be
 # found in the LICENSE file.
 
-function(register_test)
-  set(TEST_TARGET "${ARGV0}")
+# The register_test function automatically registers a test with CTest.
+# So it will be automatically executed in the CI and a xUnit-based XML
+# report will be generated.
+#
+# Usage examples:
+# - default:
+#  register_test(OrbitQtTests)
+#
+# - with test isolation:
+#  register_test(OrbitQtTests TEST_ISOLATION)
+#
+# - with custom timeout:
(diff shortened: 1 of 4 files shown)


cmake/tests.cmake
```diff
@@ -2,39 +2,56 @@
...
+#  register_test(OrbitQtTests TEST_ISOLATION TIMEOUT 120)
+
+function(register_test TEST_TARGET)
+  cmake_parse_arguments(ARGS "TEST_ISOLATION" "TIMEOUT" "" ${ARGN})
   string(REGEX REPLACE "Tests$" "" TEST_NAME "${TEST_TARGET}")
 
   set(TESTRESULTS_DIRECTORY "${CMAKE_BINARY_DIR}/testresults")
+
+  if(ARGS_TEST_ISOLATION)
+    set(ISOLATE "--isolate")
```
dimitry-, line 29: is there such a parameter?

dimitry- (commented):
And I think we should make all tests isolated by default.

cmake/tests.cmake
beckerhe: Yes. Even though it's the default. So we can also omit it.

beckerhe:
I just realized that this is more complex since Windows is not supported. Getting back to a draft...

beckerhe:
Windows has no equivalent to fork. So I guess this won't happen anytime soon. Sorry for the noise...

PR #1359, opened 2020-10-13
Revert "Remove flag guarding the tracepoint feature."

danielfenner:
This reverts commit 5eb96158a69e20a2cb00b3305cd37075b199273a.

There are concerns about performance of the feature. This needs some investigation.

Bug: http://b/169935775
Test: Local run: Tracepoint feature is not visible.


ronaldfw (approved):
LGTM!

PR #1363, opened 2020-10-14
Fix bug when picking in GraphTrack

karupayun:
A temporary solution while we look for the root cause of the picking problem.


reichlfl (approved):
As I already wrote in the chat: Now that I see this, I'm sure that the text rendering is the issue. The TextRenderer was never built to deal with picking. We should fix the root issue by passing the picking-flag also to TextRenderer::Display(), and not draw any text when picking is enabled.

Nevertheless: I think this change makes sense, because I don't see a reason why the overlay and the lines should be clickable, and this allows the user to "click through" on the underlying track.

LGTM, thanks for the change! Should this also be fixed in the release, or is it not affected?

karupayun:
You are right! I can reproduce this bug in 1.53 only with the text.

BTW, I will push it because it has to be in the release, as you said.

PR #1364, opened 2020-10-15
[REBASE] Fix bug when picking in GraphTrack

karupayun:
A temporary solution while we look for the root cause of the picking problem.


reichlfl (approved):
LGTM. Since this is only a rebase of the other PR, you don't need to get a review, btw.

PR #1365, opened 2020-10-15
Track ordering basics

reichlfl:
Basics for thread pinning and sorting.

**Pinned** tracks are always displayed on top of the capture window,
similar to frozen panes in a spreadsheet. Scrolling will not affect the
position of those tracks, and other tracks are rendered underneath
them.

Track tabs can be clicked and dragged up and down to adjust the
**sorting** in the capture window.

**Core changes**

Changes are distributed across multiple commits. Basically:
- Removed unneeded definition of methods in `Pickable` (now defined
by the tracks)
- Added `z_offset` parameter to all drawing / update functions to
allow stacking thacks "on top" of others
- Auto-sorting of tracks is now only happening during a running capture

**Open issues**

- Sorting / Pinning is not saved with the capture, and lost when the
capture is restarted
- Not all elements are ordered correctly w.r.t. depth (e.g. rounded
corners pose an issue)

Due to the open issues, this change is behind a feature flag.
Use `-track_ordering_feature` to enable.

### DIFF
--- a/OrbitGl/App.cpp
+++ b/OrbitGl/App.cpp
@@ -53,7 +53,7 @@
 ABSL_DECLARE_FLAG(bool, devmode);
 ABSL_DECLARE_FLAG(bool, local);
 ABSL_DECLARE_FLAG(bool, enable_tracepoint_feature);
-ABSL_DECLARE_FLAG(bool, thread_ordering_feature);
+ABSL_DECLARE_FLAG(bool, track_ordering_feature);
 
 using orbit_client_protos::CallstackEvent;
 using orbit_client_protos::FunctionInfo;
--- a/OrbitGl/App.cpp
+++ b/OrbitGl/App.cpp
@@ -53,6 +53,7 @@
 ABSL_DECLARE_FLAG(bool, devmode);
 ABSL_DECLARE_FLAG(bool, local);
 ABSL_DECLARE_FLAG(bool, enable_tracepoint_feature);
+ABSL_DECLARE_FLAG(bool, thread_ordering_feature);
 
 using orbit_client_protos::CallstackEvent;
 using orbit_client_protos::FunctionInfo;
--- a/OrbitGl/EventTrack.cpp
+++ b/OrbitGl/EventTrack.cpp
@@ -18,7 +18,7 @@ EventTrack::EventTrack(TimeGraph* a_TimeGraph) : Track(a_TimeGraph) {
 
 std::string EventTrack::GetTooltip() const { return "Left-click and drag to select samples"; }
 
-void EventTrack::Draw(GlCanvas* canvas, PickingMode picking_mode) {
+void EventTrack::Draw(GlCanvas* canvas, PickingMode picking_mode, float z_offset) {
   if (thread_id_ == TracepointEventBuffer::kAllTracepointsFakeTid) {
     return;
   }
@@ -29,10 +29,11 @@ void EventTrack::Draw(GlCanvas* canvas, PickingMode picking_mode) {
   // have a tooltip. For picking, we want to draw the event bar over them if
   // handling a click, and underneath otherwise.
   // This simulates "click-through" behavior.
-  const float eventBarZ = picking_mode == PickingMode::kClick ? GlCanvas::kZValueEventBarPicking
-                                                              : GlCanvas::kZValueEventBar;
+  float event_bar_z = picking_mode == PickingMode::kClick ? GlCanvas::kZValueEventBarPicking
+                                                          : GlCanvas::kZValueEventBar;
+  event_bar_z += z_offset;
   Color color = color_;
-  Box box(pos_, Vec2(size_[0], -size_[1]), eventBarZ);
+  Box box(pos_, Vec2(size_[0], -size_[1]), event_bar_z);
   batcher->AddBox(box, color, shared_from_this());
 
   if (canvas->GetPickingManager().IsThisElementPicked(this)) {
@@ -44,9 +45,8 @@ void EventTrack::Draw(GlCanvas* canvas, PickingMode picking_mode) {
   float x1 = x0 + size_[0];
   float y1 = y0 - size_[1];
 
-  batcher->AddLine(pos_, Vec2(x1, y0), GlCanvas::kZValueEventBar, color, shared_from_this());
-  batcher->AddLine(Vec2(x1, y1), Vec2(x0, y1), GlCanvas::kZValueEventBar, color,
-                   shared_from_this());
+  batcher->AddLine(pos_, Vec2(x1, y0), event_bar_z, color, shared_from_this());
+  batcher->AddLine(Vec2(x1, y1), Vec2(x0, y1), event_bar_z, color, shared_from_this());
 
   if (picked_) {
     Vec2& from = mouse_pos_[0];
@@ -58,17 +58,18 @@ void EventTrack::Draw(GlCanvas* canvas, PickingMode picking_mode) {
     y1 = y0 - size_[1];
 
     Color picked_color(0, 128, 255, 128);
-    Box box(Vec2(x0, y0), Vec2(x1 - x0, -size_[1]), GlCanvas::kZValueUi);
+    Box box(Vec2(x0, y0), Vec2(x1 - x0, -size_[1]), GlCanvas::kZValueUi + z_offset);
     batcher->AddBox(box, picked_color, shared_from_this());
   }
 
   canvas_ = canvas;
 }
 
-void EventTrack::UpdatePrimitives(uint64_t min_tick, uint64_t max_tick, PickingMode picking_mode) {
+void EventTrack::UpdatePrimitives(uint64_t min_tick, uint64_t max_tick, PickingMode picking_mode,
+                                  float z_offset) {
   Batcher* batcher = &time_graph_->GetBatcher();
   const TimeGraphLayout& layout = time_graph_->GetLayout();
-  float z = GlCanvas::kZValueEvent;
+  float z = GlCanvas::kZValueEvent + z_offset;
   float track_height = layout.GetEventTrackHeight();
   const bool picking = picking_mode != PickingMode::kNone;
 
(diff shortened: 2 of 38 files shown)


karupayun (commented):
Hey Florian, nice PR! I really like how ordering the tracks works :)

I leaved a few comments and after testing I see 3 little issues.
1) If you pick a track and try to start moving it, they first jump a little to the top. More precisely, they jump to the position where the mouse points to the top of the EventBar. You can really notice this if you click on the top of the track to move it to the bottom. 

2) After clicking a track several times, I got a crash. I have it twice. This is the log:
[2901359:2901359:20201020,131343.881348:ERROR elf_dynamic_array_reader.h:61] tag not found

3) Overlay is now on top of the track text.
overlay on top

OrbitGl/App.cpp
```diff
@@ -52,6 +52,7 @@
 ABSL_DECLARE_FLAG(bool, devmode);
 ABSL_DECLARE_FLAG(bool, local);
 ABSL_DECLARE_FLAG(bool, enable_tracepoint_feature);
+ABSL_DECLARE_FLAG(bool, thread_ordering_feature);
```
karupayun, line 55: I prefer to call it "track_ordering_feature"

OrbitGl/EventTrack.cpp
```diff
@@ -18,7 +18,7 @@ EventTrack::EventTrack(TimeGraph* a_TimeGraph) : Track(a_TimeGraph) {
 
 std::string EventTrack::GetTooltip() const { return "Left-click and drag to select samples"; }
 
-void EventTrack::Draw(GlCanvas* canvas, PickingMode picking_mode) {
+void EventTrack::Draw(GlCanvas* canvas, PickingMode picking_mode, float z_offset) {
```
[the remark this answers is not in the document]
reichlfl: I'm torn, TBH. I was thinking about the same, because feeding the z_offset through all methods can be annoying. On the other hand, the member variable makes the scope of when this is set less clear, and it's easy to forget when adding new elements, so I opted for the parameter.

OrbitGl/Track.cpp
```diff
@@ -87,7 +87,7 @@ void Track::Draw(GlCanvas* canvas, PickingMode picking_mode, float z_offset) {
   float y0 = pos_[1];
   float y1 = y0 - size_[1];
   float track_z = GlCanvas::kZValueTrack + z_offset;
-  float text_z = GlCanvas::kZValueText + z_offset;
+  float text_z = GlCanvas::kZValueTrack + z_offset;
```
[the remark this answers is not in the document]
reichlfl: Because kZValueText is on the overlay layer, but I need the text to be on the same layer as the track itself, otherwise the text of underlying tracks will be on top of the moving track.

I'm trying to minimize the usages of different z-values. IMO, text should be on the same layer as the element on which it is drawn, so it should be on the same level as the track background, because it is part of the track. Since it's drawn after the track background, it will correctly show on top of the track.

reichlfl:
> 1. If you pick a track and try to start moving it, they first jump a little to the top. More precisely, they jump to the position where the mouse points to the top of the EventBar. You can really notice this if you click on the top of the track to move it to the bottom.

Right, I've seen the same. Interestingly it's not happening for the GPU tracks... I didn't change anything in that regard, but we should add a bug to track this, thanks for bringing it up, I forgot.

> 2. After clicking a track several times, I got a crash. I have it twice. This is the log:
>    [2901359:2901359:20201020,131343.881348:ERROR elf_dynamic_array_reader.h:61] tag not found

I have no idea how that can be related, but I'll take a look before we merge this. I've not run into this so far.

> 3. Overlay is now on top of the track text.
> 
> overlay on top

Yes, but this can only be solved with the priority queue for z-values. The z-values here should be correct, but the text is drawn after the overlay and isn't displayed because the overlay is writing z-values.
To fix this, we need to draw all elements in order and disable the z-buffer completely.

OrbitGl/App.cpp
reichlfl: Good point, I'll change it.

OrbitGl/GlCanvas.cpp
```diff
@@ -34,6 +34,9 @@ float GlCanvas::kZValueMargin = 0.61f;
 float GlCanvas::kZValueSliderBg = 0.63f;
 float GlCanvas::kZValueSlider = 0.65f;
 
+float GlCanvas::kZOffsetMovingTack = 0.1f;
+float GlCanvas::kZOffsetPinnedTrack = 0.2f;
```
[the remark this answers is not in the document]
reichlfl: Right, I've increased the space in between layers, thanks!

OrbitGl/Track.cpp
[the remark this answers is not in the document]
karupayun: You are right!!

karupayun (approved):
LGTM!

reichlfl:
> 2. After clicking a track several times, I got a crash. I have it twice. This is the log:
>    [2901359:2901359:20201020,131343.881348:ERROR elf_dynamic_array_reader.h:61] tag not found

Thanks for helping to track it down, I've fixed it. PTAL!

pierricgimmig:
Hi Florian, thanks for this, this is super cool! I have a few general comments:

- Pinning multiple tracks or zooming in on a pinned tracked leaves the vertical slider usable but scrolling has no visible effect. Having a dedicated viewport for pinned tracks would be awesome, we could then scroll and resize the area we want to dedicate to pinned tracks.
- Track filtering doesn't work anymore
- GraphTrack's curves appear between pinned track background and timers

OrbitGl/TimeGraph.cpp
```diff
@@ -960,6 +957,61 @@ void TimeGraph::SortTracks() {
   }
 }
 
+void TimeGraph::UpdateMovingTrackSorting() {
+  std::shared_ptr<Track> moving_track = nullptr;
+  for (auto track = sorted_tracks_.begin(); track != sorted_tracks_.end(); ++track) {
```
pierricgimmig, line 967: "auto track_it" would be more readable.

OrbitGl/Track.cpp
```diff
@@ -170,6 +175,15 @@ void Track::Draw(GlCanvas* canvas, PickingMode picking_mode, float z_offset) {
...
 
+void Track::SetPinned(bool value) {
+  pinned_ = value && absl::GetFlag(FLAGS_thread_ordering_feature);
+  if (pinned_) {
+    picking_enabled_ = false;
+  } else {
+    picking_enabled_ = absl::GetFlag(FLAGS_thread_ordering_feature);
+  }
+}
+
```
pierricgimmig: The value of GetFlag could be stored in a local variable. Or maybe a different flag was meant to be checked here? I find it a bit difficult to grasp the logic.

OrbitGl/TimeGraph.cpp
```diff
@@ -535,17 +535,13 @@ void TimeGraph::UpdatePrimitives(PickingMode picking_mode) {
...
 
-  SortTracks();
-
-  float current_y = -layout_.GetSchedulerTrackOffset();
-
-  for (auto& track : sorted_tracks_) {
-    track->SetY(current_y);
-    track->UpdatePrimitives(min_tick, max_tick, picking_mode);
-    current_y -= (track->GetHeight() + layout_.GetSpaceBetweenTracks());
+  if (GOrbitApp->IsCapturing() || sorted_tracks_.empty()) {
```
pierricgimmig, line 538: This breaks tracks filtering which is done inside of SortTracks()

pierricgimmig (approved):
Hi Florian, overall LGTM, but please fix track filtering before submitting. As you mentioned in the notes, there are known issues with the track pinning and reordering, I found a feww also (see comment), but I'm totally fine with submitting behind the feature flag. Thanks!!

reichlfl:
> * Pinning multiple tracks or zooming in on a pinned tracked leaves the vertical slider usable but scrolling has no visible effect. Having a dedicated viewport for pinned tracks would be awesome, we could then scroll and resize the area we want to dedicate to pinned tracks.

As discussed offline, having multiple viewports requires quite some changes and I'm not sure if the effort is justified. This would either require to make the batchers aware of different viewports or duplicate all batchers for each viewport, or we should refactor CaptureWindow to allow for having multiple views on the same capture that share a common zoom / horizontal scroll value.

I think the second option would be quite beneficial in multiple ways, but should not be prioritized right now as it's likely not a small rework.

> * Track filtering doesn't work anymore

Thanks for pointing this out, it's fixed now. @pierricgimmig PTAL - the idea behind the current solution was to also make it easier to add hiding of tracks, so in the end we have a single array of all filtered and sorted tracks that is used for drawing, and another array that is just used to maintain sorting. I didn't want the tracks themselves to be aware of any ordering or filtering outside of their control, but only of their own state. If we allow to hide tracks, each track would then be aware of that since it's local track state.

> * GraphTrack's curves appear between pinned track background and timers

I think we should take another look at this after Pablos z-priority rework.

> * We can drag a track behind a pinned track, constraining the movement would be cleaner I think

True. This will need some polishing - also, we probably want to scroll the view if you drag a track to the very top / bottom. But as I said, let's keep the feature flag for now and do this in a subsequent PR.

OrbitGl/TimeGraph.cpp
reichlfl: Changed

OrbitGl/Track.cpp
reichlfl: You're right, changed this to be more clear, thanks

OrbitGl/TimeGraph.cpp
reichlfl: Fixed with the latest commit, thanks for finding this!

karupayun (approved):
LGTM!

3 things that I want to mention:
1- I really like this feature.
2- Small issue: I find counter-intuitive that if we drag some track in filtered-mode, when you erase the filter, the track go as down as possible (respecting the previous filtered order). I suppose that the most-logical is try to respect the previous order (without the filter) as most as possible or at least that if the track which was moved, now move them as top as possible.
3- This issue (I can select a track behind the pinned scheduler one) is because the drawing order also?
Screen Shot 2020-10-23 at 16 51 43
 
As discussed offline:
1) Thanks :-)
2) True, and I like your idea of making this a bit more intelligent and dynamically selecting if we need to append the track "on top" or "below" the visible tracks. I'll add that to the buganizer bug.
3) Yes. Looks like the z-index when drawing in picking mode may be wrong - we should investigate after your changes to z-ordering.

PR #1368, opened 2020-10-15
Fix collapsed timer track rendering

pierricgimmig:
box_height_ is already properly scaled when collapsed, don't scale it a second time.


pierricgimmig:
image

vs
image

ronaldfw (approved):
LGTM! Sorry, I think messed this up when doing the frame track changes.

pierricgimmig:
> I think messed this up when doing the frame track changes.

No worries!!

reichlfl:
This one should also go into Release 1.54, shouldn't it?

PR #1369, opened 2020-10-16
Fix Crash on Loading a Capture with Frame Tracks

florian-kuebler:
Does not serialize the timers of a frame track.

Bug: http://b/171026702
Test: Save and load a capture with frame tracks


ronaldfw (approved):
LGTM, and thank you so much for the fix!

PR #1374, opened 2020-10-16
Fix incorrect computation of the maximum box ratio for frame tracks

ronaldfw:
Maximum box ratio determines the scaling factor for the boxes on the
track and is based on the largest box height and the average box height.
So far, this ratio was computed online whenever a timer was added. This
is incorrect as the average value (obviously) still changes with future
timers to be added.

We fix this be adding a Finalize() method to the FrameTrack class that
has to be called after all timers have been added.


ronaldfw:
Maybe like this?

ronaldfw:
Closed because this can be solved in a simpler way. Sorry, will follow up.

PR #1376, opened 2020-10-16
Fix incorrect computation of the maximum box ratio for frame tracks

ronaldfw:
Maximum box ratio determines the scaling factor for the boxes on the
track and is based on the largest box height and the average box height.
So far, this ratio was computed online whenever a timer was added. This
is incorrect as the average value (obviously) still changes with future
timers to be added.

We fix this be computing the maxium scale factor (box ratio) only when
needed for drawing when all timers have been added.

This approach should also be correct when live updating a frame track
during a capture, which is not working yet, but planned for the future.

### DIFF
--- a/OrbitGl/FrameTrack.cpp
+++ b/OrbitGl/FrameTrack.cpp
@@ -18,20 +18,36 @@ constexpr const uint64_t kHeightCapAverageMultipleUint64 = 6;
 constexpr const float kBoxHeightMultiplier = 3.f;
 }  // namespace
 
+float FrameTrack::GetMaximumScaleFactor() const {
+  if (stats_.average_time_ns() == 0) {
+    return 0.f;
+  }
+  // Compute the scale factor in double first as we convert time values in nanoseconds to
+  // floating point. Single-precision floating point (float type) can only exactly
+  // represent all integer values up to 2^24 - 1, which given the ns time unit is fairly
+  // small (only ~16ms).
+  double scale_factor =
+      static_cast<double>(stats_.max_ns()) / static_cast<double>(stats_.average_time_ns());
+  scale_factor = std::min(scale_factor, kHeightCapAverageMultipleDouble);
+  return static_cast<float>(scale_factor);
+}
+
 float FrameTrack::GetMaximumBoxHeight() const {
   const bool is_collapsed = collapse_toggle_->IsCollapsed();
-  const float box_height_normalizer = is_collapsed ? maximum_box_ratio_ : 1.f;
-  if (maximum_box_ratio_ == 0.f) {
+  float scale_factor = GetMaximumScaleFactor();
+  const float box_height_normalizer = is_collapsed ? scale_factor : 1.f;
+  if (scale_factor == 0.f) {
     return 0.f;
   } else {
-    return maximum_box_ratio_ * box_height_ / box_height_normalizer;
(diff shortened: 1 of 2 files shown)


OrbitGl/FrameTrack.cpp
```diff
@@ -18,20 +18,32 @@ constexpr const uint64_t kHeightCapAverageMultipleUint64 = 6;
 constexpr const float kBoxHeightMultiplier = 3.f;
 }  // namespace
 
+float FrameTrack::GetMaximumScaleFactor() const {
+  float scale_factor = 0.f;
+  if (stats_.average_time_ns() > 0) {
+    scale_factor =
+      static_cast<double>(stats_.max_ns()) / static_cast<double>(stats_.average_time_ns());
```
karupayun, line 25: You should make scale_factor double (as the variable "radio" was before), or make both of these static_cast<float> instead of double. I think that float is enough here.

karupayun (approved):
LGTM Ronald! Nice PR!

Yeah, this approach should also be correct in live captures. We only need to calculate the maximum scale factor when necessary for drawing, and we don't need anything else.

OrbitGl/FrameTrack.cpp
```diff
@@ -18,20 +18,32 @@ constexpr const uint64_t kHeightCapAverageMultipleUint64 = 6;
...
+    scale_factor = std::min(scale_factor, static_cast<float>(kHeightCapAverageMultipleDouble));
+  }
+  return scale_factor;
```
reichlfl, line 28: More of a nitpick for me, but others pointed this out in PRs a lot - structuring like so could be easier to follow:
```
if (stats_.average_time_ns() == 0) {
  return 0.f;
}

return static_cast<float>(stats_.max_ns()) / stats_.average_time_ns();

reichlfl (approved):
LGTM, thanks for fixing!

OrbitGl/FrameTrack.cpp
ronaldfw: Agree, not sure why it was such a mess before. Trying with float for now.

OrbitGl/FrameTrack.cpp
ronaldfw: I think that's a good idea, done.

ronaldfw:
@karupayun I actually thought about the float/double question and I think I want to do the actual computation for the scale factor in double. Single precision floats can only represent integer values correctly up to 2^24 - 1, which given the ns time unit is only about ~16ms. That is, for times that exceed that, we start to get actual computation errors. Not sure how large they would be and whether they could cause visual artifacts (maximum scale too large / or worse too small), but I think we can avoid this question by just doing the computation in double. Does this make sense?

karupayun:
> @karupayun I actually thought about the float/double question and I think I want to do the actual computation for the scale factor in double. Single precision floats can only represent integer values correctly up to 2^24 - 1, which given the ns time unit is only about ~16ms. That is, for times that exceed that, we start to get actual computation errors. Not sure how large they would be and whether they could cause visual artifacts (maximum scale too large / or worse too small), but I think we can avoid this question by just doing the computation in double. Does this make sense?

Yes, makes sense. BTW, I'm pretty sure that as we are working only with scale and a division, we are not going to have any issue because we don't need to be too precise, but doesn't hurt.

PR #1377, opened 2020-10-19
Enable test isolation and timeouts in CTest

beckerhe:
CTest has integrated support for test isolation with a variety of unit
testing frameworks including GoogleTest. This change enables that
feature.

Unfortunately I had to fork the CMake module that adds this support
since Sponge - the logfile handling tool we use - requires all xUnit XML
testing reports to end with "_sponge_log.xml". The GoogleTest cmake
module by default names them "${test_name}.xml". So as along as we don't
want to add "_sponge_log" to our test names the simplest solution was to
modifiy the code. The adapted script lives in third_party/cmake/Modules.

Furthermore this change enables a default test timeout of 60 seconds.

Bug: http://b/167365568
Bug: http://b/170478170
Fixes: #1327


dimitry- (approved):
thanks!

PR #1379, opened 2020-10-19
Remove package upload from CI build script

beckerhe:
We used to automatically upload built packages to the internal
artifactory server when they have been built by the CI during the
run.

But with build_and_upload_dependencies.sh we have a better solution
for a while now, which we have been using exclusively for uploading
new packages.

Furthermore this upload step once corrupted the cache in the past due
to a partially uploaded package.

This all concludes that the best option is to remove the automatic
package uploading and require manual uploading with
build_and_upload_dependencies.sh.

### DIFF
--- a/kokoro/builds/build.sh
+++ b/kokoro/builds/build.sh
@@ -165,17 +165,6 @@ if [ -n "$1" ]; then
     done
   fi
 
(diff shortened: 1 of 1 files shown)


IrinaShkviro (approved):
LGTM!

PR #1380, opened 2020-10-20
Introspection revival

pierricgimmig:
We can now use the manual instrumentation API inside OrbitService and visualize our own code
(and graphed variables) alongside the target process data in the capture view. Introspection 
tracks are displayed right after the scheduler track and have a darker background. TracerThread 
has been instrumented as starting point. The feature is automatically enabled when in "--devmode".

image

The PR is broken down in 7 independent commits that are in logical order.

### DIFF
--- a/OrbitCaptureClient/CaptureEventProcessor.cpp
+++ b/OrbitCaptureClient/CaptureEventProcessor.cpp
@@ -4,7 +4,9 @@
 
 #include "OrbitCaptureClient/CaptureEventProcessor.h"
 
+#include "../Orbit.h"
 #include "CoreUtils.h"
+#include "OrbitBase/Tracing.h"
 #include "capture_data.pb.h"
 
 using orbit_client_protos::CallstackEvent;
@@ -20,6 +22,7 @@ using orbit_grpc_protos::FunctionCall;
 using orbit_grpc_protos::GpuJob;
 using orbit_grpc_protos::InternedCallstack;
 using orbit_grpc_protos::InternedString;
+using orbit_grpc_protos::IntrospectionCall;
 using orbit_grpc_protos::SchedulingSlice;
 using orbit_grpc_protos::ThreadName;
 using orbit_grpc_protos::ThreadStateSlice;
@@ -37,6 +40,9 @@ void CaptureEventProcessor::ProcessEvent(const CaptureEvent& event) {
       break;
     case CaptureEvent::kFunctionCall:
       ProcessFunctionCall(event.function_call());
+      break;
+    case CaptureEvent::kIntrospectionCall:
+      ProcessIntrospectionCall(event.introspection_call());
       break;
     case CaptureEvent::kInternedString:
       ProcessInternedString(event.interned_string());
@@ -104,6 +110,7 @@ void CaptureEventProcessor::ProcessCallstackSample(const CallstackSample& callst
 
 void CaptureEventProcessor::ProcessFunctionCall(const FunctionCall& function_call) {
   TimerInfo timer_info;
+  timer_info.set_process_id(function_call.pid());
   timer_info.set_thread_id(function_call.tid());
   timer_info.set_start(function_call.begin_timestamp_ns());
   timer_info.set_end(function_call.end_timestamp_ns());
@@ -116,6 +123,22 @@ void CaptureEventProcessor::ProcessFunctionCall(const FunctionCall& function_cal
   for (int i = 0; i < function_call.registers_size(); ++i) {
     timer_info.add_registers(function_call.registers(i));
   }
+
+  capture_listener_->OnTimer(timer_info);
+}
+
+void CaptureEventProcessor::ProcessIntrospectionCall(const IntrospectionCall& introspection_call) {
+  const FunctionCall& function_call = introspection_call.function_call();
+  TimerInfo timer_info;
+  timer_info.set_process_id(function_call.pid());
+  timer_info.set_thread_id(function_call.tid());
+  timer_info.set_start(function_call.begin_timestamp_ns());
+  timer_info.set_end(function_call.end_timestamp_ns());
+  timer_info.set_depth(static_cast<uint8_t>(function_call.depth()));
+  timer_info.set_function_address(0);
+  timer_info.set_processor(-1);
+  timer_info.set_type(TimerInfo::kIntrospection);
+  timer_info.mutable_registers()->CopyFrom(function_call.registers());
 
   capture_listener_->OnTimer(timer_info);
 }
--- a/OrbitCaptureClient/CaptureEventProcessor.cpp
+++ b/OrbitCaptureClient/CaptureEventProcessor.cpp
@@ -22,7 +22,7 @@ using orbit_grpc_protos::FunctionCall;
 using orbit_grpc_protos::GpuJob;
 using orbit_grpc_protos::InternedCallstack;
 using orbit_grpc_protos::InternedString;
-using orbit_grpc_protos::IntrospectionCall;
+using orbit_grpc_protos::IntrospectionScope;
 using orbit_grpc_protos::SchedulingSlice;
 using orbit_grpc_protos::ThreadName;
 using orbit_grpc_protos::ThreadStateSlice;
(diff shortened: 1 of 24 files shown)


OrbitCaptureClient/CaptureEventProcessor.cpp
```diff
@@ -120,6 +127,22 @@ void CaptureEventProcessor::ProcessFunctionCall(const FunctionCall& function_cal
...
 
+void CaptureEventProcessor::ProcessIntrospectionCall(const IntrospectionCall& introspection_call) {
+  const FunctionCall& function_call = introspection_call.function_call();
+  TimerInfo timer_info;
+  timer_info.set_process_id(function_call.pid());
+  timer_info.set_thread_id(function_call.tid());
+  timer_info.set_start(function_call.begin_timestamp_ns());
+  timer_info.set_end(function_call.end_timestamp_ns());
+  timer_info.set_depth(static_cast<uint8_t>(function_call.depth()));
+  timer_info.set_function_address(0);
```
ronaldfw, line 138: I would add a comment why it's okay to use these values (0, -1) here.

ronaldfw (approved):
LGTM

Did you check what happens when saving a capture? Are introspection tracks serialized?

OrbitService/LinuxTracingGrpcHandler.cpp
```diff
@@ -34,6 +38,33 @@ void LinuxTracingGrpcHandler::Start(CaptureOptions capture_options) {
   tracer_->Start();
 
   sender_thread_ = std::thread{[this] { SenderThread(); }};
+
+  if (absl::GetFlag(FLAGS_devmode)) {
+    SetupIntrospection();
+  }
```
dpallotti, line 44: Probably just a preference, but I would like it better if this was moved to where `LinuxTracingGrpcHandler` is created, and a value that enables this was set from the outside. I don't really like a global flag to modify the behavior of this class without it being visible in its interface.

OrbitService/LinuxTracingGrpcHandler.cpp
```diff
@@ -34,6 +38,33 @@ void LinuxTracingGrpcHandler::Start(CaptureOptions capture_options) {
...
+}
+
+void LinuxTracingGrpcHandler::SetupIntrospection() {
+  orbit_tracing_listener_ =
+      std::make_unique<orbit::tracing::Listener>([this](const orbit::tracing::Scope& scope) {
+        IntrospectionCall introspection_call;
+        FunctionCall& function_call = *introspection_call.mutable_function_call();
+
+        function_call.set_tid(scope.tid);
+        function_call.set_pid(getpid());
```
dpallotti, line 54: Move `set_pid` before `set_tid`?

OrbitGl/TimeGraph.cpp
```diff
@@ -956,6 +979,32 @@ void TimeGraph::SortTracks() {
...
+
+    // Scheduler track.
+    if (!scheduler_track_->IsEmpty()) {
+      sorted_tracks_.emplace_back(scheduler_track_);
+    }
+
+    // For now, "external_pid_tracks" should only contain
+    // introspection tracks. Display them on top.
+    Append(sorted_tracks_, external_pid_tracks);
+    Append(sorted_tracks_, capture_pid_tracks);
```
dpallotti, line 1006: Not in favor of these utility functions, `std::insert` should be used.

OrbitGrpcProtos/capture.proto
```diff
@@ -61,6 +61,10 @@ message FunctionCall {
   repeated uint64 registers = 8;
 }
 
+message IntrospectionCall {
+  FunctionCall function_call = 1;
```
dpallotti, line 65: In general this is not a function call, it should be called `IntrospectionScope` and not reuse `FunctionCall`.

OrbitGl/ThreadTrack.cpp
```diff
@@ -53,7 +53,6 @@ std::string ThreadTrack::GetBoxTooltip(PickingId id) const {
 
   const FunctionInfo* func =
       GOrbitApp->GetCaptureData().GetSelectedFunction(text_box->GetTimerInfo().function_address());
-  CHECK(func != nullptr);
```
dpallotti, line 56: Is there a way to keep the `CHECK`? Like `CHECK(function != nullptr || introspection)`?

OrbitGl/ThreadTrack.cpp
```diff
@@ -119,11 +119,19 @@ Color ThreadTrack::GetTimerColor(const TimerInfo& timer_info, bool is_selected)
 
   uint64_t address = timer_info.function_address();
   const FunctionInfo* function_info = GOrbitApp->GetCaptureData().GetSelectedFunction(address);
-  CHECK(function_info);
```
dpallotti, line 122: Same here, is there a way to keep the `CHECK`?

OrbitService/LinuxTracingGrpcHandler.cpp
```diff
@@ -80,6 +111,14 @@ void LinuxTracingGrpcHandler::OnFunctionCall(FunctionCall function_call) {
   }
 }
 
+void LinuxTracingGrpcHandler::OnIntrospectionCall(
+    orbit_grpc_protos::IntrospectionCall introspection_call) {
+  CaptureEvent event;
+  *event.mutable_introspection_call() = std::move(introspection_call);
+  absl::MutexLock lock{&event_buffer_mutex_};
+  event_buffer_.emplace_back(std::move(event));
```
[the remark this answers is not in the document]
pierricgimmig: I tend to add them only when needed, but for consistency with the surrounding code I agree

pierricgimmig:
> Did you check what happens when saving a capture? Are introspection tracks serialized?

Yes, the timers are serialized normally, but not the graph tracks, I know why, I will fix it in an upcoming PR.

OrbitCaptureClient/CaptureEventProcessor.cpp
pierricgimmig: done

OrbitService/LinuxTracingGrpcHandler.cpp
pierricgimmig: Yeah I have a PR where introspection is set in the capture options so that we can eventually toggle it from the UI. I will add you as a reviewer, thanks!

OrbitService/LinuxTracingGrpcHandler.cpp
pierricgimmig: Done

OrbitGl/Track.h
```diff
@@ -86,10 +86,13 @@ class Track : public Pickable, public std::enable_shared_from_this<Track> {
   [[nodiscard]] Vec2 GetPos() const { return pos_; }
   void SetSize(float a_SizeX, float a_SizeY);
   void SetColor(Color a_Color) { color_ = a_Color; }
+  [[nodiscard]] Color GetBackGroundColor() const;
```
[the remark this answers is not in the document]
pierricgimmig: Done.

OrbitGl/TimeGraph.cpp
pierricgimmig: I find Append is much more verbose and concise than the insert syntax. It would take two lines to write using insert!!

OrbitGrpcProtos/capture.proto
pierricgimmig: Done.

OrbitGl/TimerTrack.cpp
```diff
@@ -147,6 +147,10 @@ void TimerTrack::OnTimer(const TimerInfo& timer_info) {
     UpdateDepth(timer_info.depth() + 1);
   }
 
+  if(process_id_ == -1) {
+    process_id_ = timer_info.process_id();
```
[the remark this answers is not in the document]
pierricgimmig: Yeah, I'll see if I can move it to the track creation.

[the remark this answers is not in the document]
pierricgimmig: Yeah I'll see if I can move it to the track creation.

OrbitGl/ThreadTrack.cpp
pierricgimmig: Done.

OrbitGl/ThreadTrack.cpp
pierricgimmig: Done.

pierricgimmig: yep, done.

OrbitGl/TimeGraph.cpp
dpallotti:
```
std::insert(sorted_tracks_.begin(), external_pid_tracks.begin(), external_pid_tracks.end());
std::insert(sorted_tracks_.begin(), capture_pid_tracks.begin(), capture_pid_tracks.end());
```

Sure, you have to write the source twice, but it's a well-known function from the standard, used in many places in the codebase. Having a wrapper around a standard function for such minor overhead is in my opinion not a good practice.

PR #1381, opened 2020-10-20
Disable test detection in cross compilation setup

beckerhe:
CMake's GoogleTest integration calls the test executable to obtain a
list of all the included tests. Afterwards every test can be executed in
isolation by CTest.

This fails in a cross compilation scenario because the test executable
can't be executed. Previously we solved that by just not running the
tests, but this won't work with the new GoogleTest integration since the
test executable will be called as a part of the build process - to
enumerate all the tests.

We resolve that problem by not registering the test executable at all
when in a cross compilation scenario.


ronaldfw (approved):
LGTM, thanks for the quick fix.

PR #1382, opened 2020-10-20
Improve logging by adding time and directory

dimitry-:
Logging now reports time (UTC timezone) and directory:
Instead of main.cpp you will get "OrbitQt/main.cpp"

Test: run Orbit client and check that there are timestamps in the log

### DIFF
--- a/OrbitBase/Logging.cpp
+++ b/OrbitBase/Logging.cpp
@@ -4,12 +4,17 @@
 
 #include <OrbitBase/Logging.h>
 
-static absl::Mutex log_file_mutex;
+static absl::Mutex log_file_mutex(absl::kConstInit);
 std::ofstream log_file;
 
 void InitLogFile(const std::string& path) {
   absl::MutexLock lock(&log_file_mutex);
-  CHECK(!log_file.is_open());
+  // Do not call CHECK here - it will end up calling LogToFile,
+  // which tries to lock on the same mutex a second time. This will
+  // lead to an error since the mutex is not recursive.
+  if (log_file.is_open()) {
+    PLATFORM_ABORT();
+  }
   log_file.open(path, std::ofstream::out);
 }
 
--- a/OrbitBase/include/OrbitBase/Logging.h
+++ b/OrbitBase/include/OrbitBase/Logging.h
@@ -15,21 +15,28 @@
 
 #include "absl/strings/str_format.h"
 #include "absl/synchronization/mutex.h"
+#include "absl/time/time.h"
 
 #ifdef __clang__
 #pragma clang diagnostic push
 #pragma clang diagnostic ignored "-Wgnu-zero-variadic-macro-arguments"
 #endif
 
-#define LOG(format, ...)                                                                \
-  do {                                                                                  \
-    std::string file__ = std::filesystem::path(__FILE__).filename().string();           \
-    std::string file_and_line__ = absl::StrFormat("%s:%d", file__.c_str(), __LINE__);   \
-    if (file_and_line__.size() > 28)                                                    \
-      file_and_line__ = "..." + file_and_line__.substr(file_and_line__.size() - 25);    \
-    std::string formatted_log__ =                                                       \
-        absl::StrFormat("[%28s] " format "\n", file_and_line__.c_str(), ##__VA_ARGS__); \
-    PLATFORM_LOG(formatted_log__.c_str());                                              \
+constexpr const char* kLogTimeFormat = "%Y-%m-%dT%H:%M:%E6S";
+
+#define LOG(format, ...)                                                                     \
+  do {                                                                                       \
+    std::filesystem::path path__ = std::filesystem::path(__FILE__);                          \
+    std::string file__;                                                                      \
+    std::string dir__;                                                                       \
+    auto path_it__ = path__.end();                                                           \
+    if (path_it__ != path__.begin()) file__ = (--path_it__)->string();                       \
+    if (path_it__ != path__.begin()) dir__ = (--path_it__)->string();                        \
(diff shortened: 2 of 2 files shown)


OrbitBase/Logging.cpp
```diff
@@ -4,12 +4,14 @@
...
 
-static absl::Mutex log_file_mutex;
+static absl::Mutex log_file_mutex(absl::kConstInit);
 std::ofstream log_file;
 
 void InitLogFile(const std::string& path) {
   absl::MutexLock lock(&log_file_mutex);
-  CHECK(!log_file.is_open());
+  if (log_file.is_open()) {
+    PLATFORM_ABORT();
```
dpallotti, line 13: Why don't we want the message from `CHECK` here?

OrbitBase/include/OrbitBase/Logging.h
```diff
@@ -15,21 +15,28 @@
...
+#define LOG(format, ...)                                                                     \
+  do {                                                                                       \
+    std::filesystem::path path__ = std::filesystem::path(__FILE__);                          \
+    std::string dir__;                                                                       \
+    auto path_it__ = path__.end();                                                           \
+    if (path_it__ != path__.begin() && --path_it__ != path__.begin() &&                      \
+        --path_it__ != path__.begin()) {                                                     \
+      dir__ = path_it__->string();                                                           \
+    }                                                                                        \
+    std::string time__ = absl::FormatTime(absl::Now(), absl::UTCTimeZone());                 \
```
dpallotti, line 34: I would move this up or down so that it doesn't split `dir__` and `file__`.

OrbitBase/include/OrbitBase/Logging.h
```diff
@@ -15,21 +15,28 @@
...
-    std::string formatted_log__ =                                                       \
-        absl::StrFormat("[%28s] " format "\n", file_and_line__.c_str(), ##__VA_ARGS__); \
-    PLATFORM_LOG(formatted_log__.c_str());                                              \
+#define LOG(format, ...)                                                                     \
+  do {                                                                                       \
+    std::filesystem::path path__ = std::filesystem::path(__FILE__);                          \
+    std::string dir__;                                                                       \
+    auto path_it__ = path__.end();                                                           \
+    if (path_it__ != path__.begin() && --path_it__ != path__.begin() &&                      \
+        --path_it__ != path__.begin()) {                                                     \
```
dpallotti, line 31: I don't really understand the reason for this last `!= path__.begin()`. What's wrong if `path_it__` is now `path__.begin()`?

OrbitBase/include/OrbitBase/Logging.h
```diff
@@ -15,21 +15,28 @@
...
+      dir__ = path_it__->string();                                                           \
+    }                                                                                        \
+    std::string time__ = absl::FormatTime(absl::Now(), absl::UTCTimeZone());                 \
+    std::string file__ = path__.filename().string();                                         \
```
dpallotti, line 35: This is just a thought, but I would find it more consistent to do it the same way as for `dir__`:
```
if (path_it__ != path__.begin()) file__ = (--path_it__)->string();
if (path_it__ != path__.begin()) dir__ = (--path_it__)->string();
```
What do you think?

OrbitBase/include/OrbitBase/Logging.h
dpallotti, line 34: How does an example formatted time look like?
I'm afraid it would look like the example given in the docs, `2018-08-06T23:35:32.637472794+00:00`, in which case some stuff could definitely be removed, like the time zone and honestly at least the last three digits of the decimal part.

OrbitBase/Logging.cpp
dimitry-: this happens under the lock - CHECK will invoke LogToFile which will fail because the mutex is not recursive here. Since this method is part of Logging system initialization it shouldn't call logging.

OrbitBase/include/OrbitBase/Logging.h
dimitry-: right - the last one is not needed. fixed.

OrbitBase/include/OrbitBase/Logging.h
dimitry-: done.

OrbitBase/include/OrbitBase/Logging.h
dimitry-: Good idea - done.

OrbitBase/include/OrbitBase/Logging.h
dimitry-: removed timezone and reduced precision to microseconds (6 digits)

OrbitBase/Logging.cpp
dimitry-: added comment explaining this.

OrbitBase/Logging.cpp
```diff
@@ -4,12 +4,17 @@
...
-static absl::Mutex log_file_mutex;
+static absl::Mutex log_file_mutex(absl::kConstInit);
 std::ofstream log_file;
 
 void InitLogFile(const std::string& path) {
   absl::MutexLock lock(&log_file_mutex);
-  CHECK(!log_file.is_open());
+  // Do not call CHECK here - it will end up calling LogToFile
+  // which tries to lock on the same mutex second time -> it will
+  // lead to a error since the mutex in not recursive.
```
dpallotti, line 14:
```suggestion
  // Do not call CHECK here - it will end up calling LogToFile,
  // which tries to lock on the same mutex second time. This will
  // lead to an error since the mutex is not recursive.
```

dpallotti (approved):

OrbitBase/Logging.cpp
dimitry-: done.

PR #1383, opened 2020-10-20
ClientGgp: Make LoadSelectedFunctions private and add call when functions are updated

kisenshi:
Update the scope of _LoadSelectedFunctions_ as it is just used internally: during the initialisation of the client and when the method to update the capture functions is called (_UpdateCaptureFunctions_).

I realised it was a design flaw when using the _ClientGgp_ in a non-linear client. I consider it makes more sense to update the functions and load them so they are used when capturing in a unique call.

### DIFF
--- a/OrbitClientGgp/ClientGgp.cpp
+++ b/OrbitClientGgp/ClientGgp.cpp
@@ -259,6 +259,7 @@ absl::flat_hash_map<uint64_t, FunctionInfo> ClientGgp::GetSelectedFunctions() {
 
 void ClientGgp::UpdateCaptureFunctions(std::vector<std::string> capture_functions) {
   options_.capture_functions = capture_functions;
+  LoadSelectedFunctions();
   return;
 }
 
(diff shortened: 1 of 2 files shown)


OrbitClientGgp/ClientGgp.cpp
```diff
@@ -259,6 +259,8 @@ absl::flat_hash_map<uint64_t, FunctionInfo> ClientGgp::GetSelectedFunctions() {
 
 void ClientGgp::UpdateCaptureFunctions(std::vector<std::string> capture_functions) {
   options_.capture_functions = capture_functions;
+  // Load updated selected functions
```
ronaldfw, line 262: Remove the comment, it's redundant.

ronaldfw (approved):
LGTM

OrbitClientGgp/ClientGgp.cpp

kisenshi:
Merging this now, thanks!

PR #1385, opened 2020-10-20
Create OrbitCaptureGgpGrpc service and client

kisenshi:
Create a grpc service that makes use of the _ClientGgp_ class and waits for grpc calls from a client to start and save a capture. 

Two different projects have been created, one containing the server and service logic (_OrbitCaptureGgpService_) and the other containing the client side (_OrbitCaptureGgpClient_). 

The client is completely transparent about how the orbit capture is handled and just make the requests to the service (_StartCapture_, _StopAndSaveCapture_, _UpdateSelectedFunctions_ and _ShutdownService_).

The service instantiates _ClientGgp_, which connects with OrbitService to handle and receive the capture data. The service makes sure not more than capture runs at the same time as well as the thread pool and all the bits related to capture are shutdown appropriately when requested. 

_OrbitCaptureGgpService_ runs the server, registers the service and waits for requests from the client. A watchdog has been included to check every few seconds if it has been requested to shutdown to do so.

The grpc calls have been tested using _OrbitCaptureGgpClient/main.cpp_ which has been created for testing purposes. It allows making each of the available grpc calls by entering the right choice in the terminal. 

**Note:** The system assumes OrbitService is already running in the background.


kisenshi:
Apologies for the size of the PR. I have committed each of the different steps I have followed independently, which might help with the review. Thanks!

danielfenner (approved):
Looks good! Merely nitpicking...

OrbitGrpcProtos/services_ggp.proto
```diff
@@ -0,0 +1,33 @@
+// Copyright (c) 2020 The Orbit Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+syntax = "proto3";
+
+package orbit_grpc_protos;
+
+// Hello world test code
```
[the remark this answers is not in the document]
kisenshi: Removed :)

OrbitCaptureGgpService/main.cpp
```diff
@@ -0,0 +1,63 @@
...
+}
+
+}  // namespace
+
+int main(int argc, char** argv) {
+  absl::SetProgramUsageMessage("Orbit CPU Profiler Ggp Client");
+  absl::SetFlagsUsageConfig(absl::FlagsUsageConfig{{}, {}, {}, &OrbitCore::GetBuildReport, {}});
+  absl::ParseCommandLine(argc, argv);
+
+  const std::string log_directory = absl::GetFlag(FLAGS_log_directory);
```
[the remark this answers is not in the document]
kisenshi: If I move it I would not be able to check if the _log_directory_ is empty. If the user does not specify a log directory, it is empty by default and in that case we wouldn't want to call _InitLogFile_ at all as the user would not want to create a log file and it would try to do so. That's why I think it is needed to be here and be provided as an argument instead

OrbitClientGgp/include/OrbitClientGgp/ClientGgp.h
```diff
@@ -22,7 +22,10 @@
 
 class ClientGgp final : public CaptureListener {
  public:
-  ClientGgp(ClientGgpOptions&& options);
+  explicit ClientGgp(){};
```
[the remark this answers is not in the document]
kisenshi: You are right, will remove it, thanks!

OrbitCaptureGgpClient/main.cpp
```diff
@@ -0,0 +1,101 @@
...
+#include "OrbitCaptureGgpClient/OrbitCaptureGgpClient.h"
+#include "absl/flags/flag.h"
+#include "absl/flags/parse.h"
+#include "absl/flags/usage.h"
+#include "absl/flags/usage_config.h"
+#include "services_ggp.grpc.pb.h"
+
+ABSL_FLAG(uint16_t, grpc_port, 44767, "gRPC server port for capture ggp service");
+
+int main(int argc, char** argv) {
```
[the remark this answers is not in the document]
kisenshi: Makes sense, will add it, thanks!

OrbitClientGgp/include/OrbitClientGgp/ClientGgp.h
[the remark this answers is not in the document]
kisenshi: Done!

OrbitCaptureGgpClient/main.cpp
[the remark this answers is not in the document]
kisenshi: Done!

CMakeLists.txt
```diff
@@ -143,6 +143,8 @@ else()
   add_subdirectory(OrbitLinuxTracing)
   add_subdirectory(OrbitService)
   add_subdirectory(OrbitClientGgp)
+  add_subdirectory(OrbitCaptureGgpClient)
```
ronaldfw, line 146: I think we had discussed this before, not sure what the conclusion was: Does this actually have anything to do with ggp, or is this just a generic Linux client? (The name seems to suggest there is something special to ggp about this.)

No need to change anything now, but we should clarify.

OrbitCaptureGgpClient/main.cpp
```diff
@@ -0,0 +1,104 @@
...
+    std::cout << "1 Start capture\n";
+    std::cout << "2 Stop and save capture\n";
+    std::cout << "3 Hook functions\n";
+    std::cout << "4 Shutdown service and exit\n";
+    std::cout << "\n";
+    std::cout << "Introduce your choice (1-4): ";
+
+    std::cin >> i;
+    switch (i) {
+      case 1:
```
ronaldfw, line 62: Nit: Should be defined constants, like `constexpr const int kStartCaptureCommand = 1`.

OrbitCaptureGgpService/main.cpp
```diff
@@ -0,0 +1,63 @@
...
+#include "absl/flags/usage_config.h"
+
+ABSL_FLAG(uint16_t, grpc_port, 44767, "gRPC server port for capture ggp service");
+ABSL_FLAG(uint16_t, orbit_service_grpc_port, 44765, "gRPC server port for OrbitService");
+ABSL_FLAG(int32_t, pid, 0, "pid to capture");
+ABSL_FLAG(std::vector<std::string>, functions, {},
+          "Comma-separated list of functions to hook to the capture");
+ABSL_FLAG(std::string, file_name, "", "File name used for saving the capture");
+ABSL_FLAG(std::string, file_directory, "/var/game/",
+          "Path to locate orbit file. By default it is /var/game/");
```
ronaldfw, line 24: Nit: "orbit" -> ".orbit" to make it clear this is the ".orbit" capture file. (Or, if not, rename to "Orbit".)

ronaldfw (approved):
LGTM!

CMakeLists.txt
kisenshi: It should work in Linux as well but it is meant to be used in ggp that is why I called it like that. Happy to change it to a better and more clear name. I think the names need a review anyway as they can be quite confusing (Daniel mentioned it as well offline) so we can have a meeting to discuss about them, agree which should be keep and which should be updated and then I will create a PR for that.

OrbitCaptureGgpClient/main.cpp
kisenshi: Ok, will update this; thanks!

OrbitCaptureGgpService/main.cpp
kisenshi: Makes sense, will update it as well; thanks!

OrbitCaptureGgpClient/main.cpp
kisenshi: Done!

OrbitCaptureGgpService/main.cpp
kisenshi: Done!

kisenshi:
I've updated the code to solve the nit pick suggestions. Thanks!

PR #1386, opened 2020-10-21
Add introspection to capture options

pierricgimmig:
Hi @kisenshi and @dpallotti , this PR adds a new entry in the capture options.  The idea is to be able to control whether or not we want introspection from the UI (a toggle will be added shortly).


dpallotti (approved):
Yep, I like this better.

You mentioned "a toggle" to be added? You mean a command line option, right?

kisenshi (approved):
lgtm, thanks!

pierricgimmig:
> You mentioned "a toggle" to be added? You mean a command line option, right?

I was thinking of adding a way to disable introspection from the UI in the debug menu, it would still be enabled by default when in "devmode".

dpallotti:
> > You mentioned "a toggle" to be added? You mean a command line option, right?
> 
> I was thinking of adding a way to disable introspection from the UI in the debug menu, it would still be enabled by default when in "devmode".

Ah right, makes sense, I was mainly wondering by what command line flag it would be hidden from users.

PR #1387, opened 2020-10-21
Display sftp error to user

antonrohr:
One more sftp error is now displayed to the user, instead of only being
logged. Before this change the error would lead to a crash, because the
symbol loading mechanism would expect a file to be there, even though it
is not. Compare b/170698772


IrinaShkviro (approved):
Looks good to me, thanks!

dimitry- (approved):

beckerhe (approved):
LGTM!

PR #1389, opened 2020-10-21
Enable PRE_TEST test discovery

beckerhe:
Previously the test binaries have been called during build time to
enumerate all the included tests. This call is now deferred to just
before starting the tests.
    
The previous way made problems with test binaries like
OrbitQtTests.exe which need extra libraries like Qt5Core.dll.
Qt libraries are copied to the binary directory as part of the build
which means we can't be sure that these are available when executing
the test binary for lisitng. It has been working so far due to pure
luck, but it breaks when a race condition.


antonrohr (approved):
Makes sense to me. Thanks!

PR #1391, opened 2020-10-23
Margin change in timegraph forces update

reichlfl:
To reproduce:

* Take a capture
* Filter tracks to make the vertical scrollbar disappear (e.g. type "s" for the scheduler track)

The right margin in the capture window is in an invalid state, rounded corners are drawn incorrectly:
image

This PR fixes that issue.

### DIFF
--- a/OrbitGl/TimeGraph.cpp
+++ b/OrbitGl/TimeGraph.cpp
@@ -1068,6 +1068,15 @@ void TimeGraph::JumpToNeighborBox(const TextBox* from, JumpDirection jump_direct
   }
 }
 
+void TimeGraph::SetRightMargin(float margin) {
+  {
+    if (right_margin_ != margin) {
+      NeedsUpdate();
+    }
+    right_margin_ = margin;
+  }
+}
+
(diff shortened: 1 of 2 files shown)


OrbitGl/TimeGraph.cpp
```diff
@@ -1068,6 +1068,15 @@ void TimeGraph::JumpToNeighborBox(const TextBox* from, JumpDirection jump_direct
   }
 }
 
+void TimeGraph::SetRightMargin(float margin) {
+  {
+    if (right_margin_ != margin) {
+      NeedsUpdate();
+    }
+    right_margin_ = margin;
```
karupayun, line 1076: I would like to have this line inside the if-condition

karupayun (approved):
LGTM!

Only 1 comment.

OrbitGl/TimeGraph.cpp
florian-kuebler: I'd be fine with both cases, but I really think the name should be `UpdateRightMargin`.

PR #1393, opened 2020-10-27
Fix GraphTrack overdraw

pierricgimmig:
Redraw curves only when an unpdate is required (zoom, pan), not
on mouse move. The labels do need to be redrawn on mouse move so
split rendering into "UpdatePrimitives" and "Draw".

### DIFF
--- a/OrbitGl/GraphTrack.cpp
+++ b/OrbitGl/GraphTrack.cpp
@@ -9,85 +9,52 @@
 GraphTrack::GraphTrack(TimeGraph* time_graph, std::string name)
     : Track(time_graph), name_(std::move(name)) {}
 
-void GraphTrack::Draw(GlCanvas* canvas, PickingMode picking_mode, float z_offset) {
-  Batcher* batcher = canvas->GetBatcher();
+void GraphTrack::UpdatePrimitives(uint64_t min_tick, uint64_t max_tick, PickingMode picking_mode,
+                                  float z_offset) {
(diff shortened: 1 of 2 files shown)


OrbitGl/GraphTrack.cpp
```diff
@@ -9,96 +9,84 @@
...
       previous_time = time;
       last_normalized_value = normalized_value;
     }
   }
 }
 
-void GraphTrack::DrawSquareDot(GlCanvas* canvas, Vec2 center, float radius, float z, Color color) {
+void GraphTrack::Draw(GlCanvas* canvas, PickingMode picking_mode, float z_offset) {
+  const bool picking = picking_mode != PickingMode::kNone;
+  if (picking) {
```
karupayun, line 67: Please put the condition inside the if (as we now are using it only once)

if (picking_mode != PickingMode::kNone) ...

karupayun (approved):
LGTM! Thanks for the change Pierric!

OrbitGl/GraphTrack.cpp
karupayun, line 67: Maybe I'd prefer to have the condition inside the if statement, what do you think?

Or you like to have the variable in terms of readability?

pierricgimmig: It is for readability yes, I'll leave it because I do think it's easier to understand.

PR #1394, opened 2020-10-27
 Increase min width of file and line in logs to 40 characters

dpallotti:
This is to increase the times when the prefix of a log line has a fixed width,
which makes the logs easier to read.

Test: Run Orbit and look at the logs.

### DIFF
--- a/OrbitBase/include/OrbitBase/Logging.h
+++ b/OrbitBase/include/OrbitBase/Logging.h
@@ -35,7 +35,7 @@ constexpr const char* kLogTimeFormat = "%Y-%m-%dT%H:%M:%E6S";
     std::string file_and_line__ = absl::StrFormat("%s/%s:%d", dir__, file__, __LINE__);      \
     std::string time__ = absl::FormatTime(kLogTimeFormat, absl::Now(), absl::UTCTimeZone()); \
     std::string formatted_log__ =                                                            \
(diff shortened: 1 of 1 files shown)


OrbitBase/include/OrbitBase/Logging.h
```diff
@@ -33,9 +33,11 @@ constexpr const char* kLogTimeFormat = "%Y-%m-%dT%H:%M:%E6S";
     if (path_it__ != path__.begin()) file__ = (--path_it__)->string();                       \
     if (path_it__ != path__.begin()) dir__ = (--path_it__)->string();                        \
     std::string file_and_line__ = absl::StrFormat("%s/%s:%d", dir__, file__, __LINE__);      \
+    if (file_and_line__.size() > 32)                                                         \
+      file_and_line__ = file_and_line__.substr(file_and_line__.size() - 32);                 \
```
dimitry-, line 37: I would rather see full names than perfect formatting.

dpallotti: So you would want to keep `%28s` or should it be changed to `%s`?

In any case, in my experience reading the logs right now is super annoying. We can increase 32 to an even higher number, but without a fixed initial column the format is all over the place.

dimitry-: I think I prefer %s there, seems like otherwise we are wasting space on the screen...

dimitry- (approved):

PR #1395, opened 2020-10-27
Add automated test for frame tracks

ronaldfw:
This adds an automated test for creation of frame tracks. Since this
requires hooking a function and taking a capture first, the whole
process of hooking and capturing is refactored to be reusable between
the test for instrumenting a function and the new frame tracks test.

Tested: Ran all automated tests locally.


danielfenner (approved):
Thanks for refactoring the commonly used functions. Lgtm!

PR #1396, opened 2020-10-27
Resize columns of top-down/bottom-up based on the size of the views

dpallotti:
Bug: http://b/170946091

Test: Profile a target with long function names and make sure that the columns
of the bottom-up and top-down view have always reasonable sizes.

### DIFF
--- a/OrbitQt/CallTreeWidget.cpp
+++ b/OrbitQt/CallTreeWidget.cpp
@@ -49,13 +49,9 @@ void CallTreeWidget::SetCallTreeView(std::unique_ptr<CallTreeView> call_tree_vie
   ui_->callTreeTreeView->setModel(hooked_proxy_model_.get());
   ui_->callTreeTreeView->sortByColumn(CallTreeViewItemModel::kInclusive, Qt::DescendingOrder);
 
-  // Resize columns only the first time a non-empty CallTreeView is set.
-  if (!columns_already_resized_ && hooked_proxy_model_->rowCount(QModelIndex{}) > 0) {
-    ui_->callTreeTreeView->header()->resizeSections(QHeaderView::ResizeToContents);
-    columns_already_resized_ = true;
-  }
-
   onSearchLineEditTextEdited(ui_->searchLineEdit->text());
+
+  ResizeColumnsIfNecessary();
 }
 
 namespace {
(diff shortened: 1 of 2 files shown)


OrbitQt/CallTreeWidget.h
```diff
@@ -92,13 +95,17 @@ class CallTreeWidget : public QWidget {
...
+
   std::unique_ptr<Ui::CallTreeWidget> ui_;
   OrbitApp* app_ = nullptr;
   std::unique_ptr<CallTreeViewItemModel> model_;
   std::unique_ptr<QIdentityProxyModel> hide_values_proxy_model_;
   std::unique_ptr<HighlightCustomFilterSortFilterProxyModel> search_proxy_model_;
   std::unique_ptr<HookedIdentityProxyModel> hooked_proxy_model_;
-  bool columns_already_resized_ = false;
+
+  enum class ColumnResizingState { kInitial = 0, kWidgetSizeSet = 1, kDone = 2 };
```
dimitry-, line 107: Maybe add explanation for these states here?

OrbitQt/CallTreeWidget.cpp
```diff
@@ -103,6 +99,40 @@ void CallTreeWidget::SetBottomUpView(std::unique_ptr<CallTreeView> bottom_up_vie
...
+  }
+}
+
+void CallTreeWidget::ResizeColumnsIfNecessary() {
+  // Resize columns only once, when the following holds:
+  // - a non-empty CallTreeView is set;
+  // - the size of the QTreeView has been set, which means that the size of CallTreeWidget has been
+  //   set, which in turns means that the tab has been shown at least once.
+  if (column_resizing_state_ == ColumnResizingState::kWidgetSizeSet &&
+      hooked_proxy_model_ != nullptr && hooked_proxy_model_->rowCount(QModelIndex{}) > 0) {
```
dimitry-, line 115: nit: inverse the if - use early return here.

dpallotti: Good point, done.

OrbitQt/CallTreeWidget.h
dpallotti: Added a brief explanation for each state.

OrbitQt/CallTreeWidget.h
```diff
@@ -92,13 +95,21 @@ class CallTreeWidget : public QWidget {
...
   OrbitApp* app_ = nullptr;
   std::unique_ptr<CallTreeViewItemModel> model_;
   std::unique_ptr<QIdentityProxyModel> hide_values_proxy_model_;
   std::unique_ptr<HighlightCustomFilterSortFilterProxyModel> search_proxy_model_;
   std::unique_ptr<HookedIdentityProxyModel> hooked_proxy_model_;
-  bool columns_already_resized_ = false;
+
+  enum class ColumnResizingState {
+    kInitial = 0,        // CallTreeWidget hasn't got its first size set yet
+    kWidgetSizeSet = 1,  // CallTreeWidget has got its first size, resize the columns when possible
```
dimitry-, line 109: is resizeEvent something called only once?

dimitry- (approved):
LGTM

OrbitQt/CallTreeWidget.h
dpallotti: No, it's called every time the widget is resized (including when the first size is set when the widget is first shown). Have you noticed anything that could be wrong based on this?

dimitry-: and we do not want to resize columns if this was not the first one, correct?

OrbitQt/CallTreeWidget.cpp
```diff
@@ -103,6 +99,42 @@ void CallTreeWidget::SetBottomUpView(std::unique_ptr<CallTreeView> bottom_up_vie
...
+      hooked_proxy_model_ == nullptr || hooked_proxy_model_->rowCount(QModelIndex{}) == 0) {
+    return;
+  }
+
+  ui_->callTreeTreeView->header()->resizeSections(QHeaderView::ResizeToContents);
+
+  // Shrink the first column (to a minimum width) so that all columns are visible,
+  // but still make it as wide as possible.
+  static constexpr int kMinThreadOrFunctionColumnSize = 200;
+  int thread_or_function_column_size = ui_->callTreeTreeView->header()->width();
```
antonrohr, line 124: I tested this only on Linux and maybe it works differently on Windows, but this `ui_->callTreeTreeView->header()->width()` does not return the maximum size the header can be (without a scrollbar being necessary), but some smaller size. The result is that the last column is taking up a lot of space and not the first. 

Maybe I am missing something, but intuitively I would say the first column in this screenshot should be the big one. 
Screenshot 2020-10-27 at 16 34 12

OrbitQt/CallTreeWidget.h
dpallotti: Exactly, we want to resize only once just to give a reasonable width. At least that's my intention.

OrbitQt/CallTreeWidget.cpp
dpallotti: In theory this is correct because the first column doesn't need shrinking. Maybe we also want to expand to take all the available space? 

So `ui_->callTreeTreeView->header()->width()` is less that `ui_->callTreeTreeView->width()`? I'll have a look.

dpallotti: So this is because when more horizontal space is available than needed, `resizeSections(QHeaderView::ResizeToContents)` gives all the remaining space to the last column.
This is avoided by temporarily setting `setStretchLastSection(false)` before performing the computations.
Please have a look.

antonrohr: I like this more, thank you!

antonrohr (approved):

PR #1398, opened 2020-10-27
Fix bug where module update is necessary

antonrohr:
Before this change, a module (ModuleData) was considered to not change
during a profiling session. Now modules are updated when they change.
Tests and usages are changed accordingly. b/171765631

### DIFF
--- a/OrbitClientData/ModuleData.cpp
+++ b/OrbitClientData/ModuleData.cpp
@@ -8,12 +8,42 @@
 #include "OrbitClientData/FunctionUtils.h"
 #include "absl/synchronization/mutex.h"
 #include "capture_data.pb.h"
+#include "module.pb.h"
 
 using orbit_client_protos::FunctionInfo;
+using orbit_grpc_protos::ModuleInfo;
 
 bool ModuleData::is_loaded() const {
   absl::MutexLock lock(&mutex_);
   return is_loaded_;
+}
+
+void ModuleData::UpdateIfChanged(ModuleInfo info) {
+  absl::MutexLock lock(&mutex_);
+
+  CHECK(file_path() == info.file_path());
+
+  // TODO(171878807): Remove this as soon as a better way of distinguishing modules is implemented.
+  const bool build_id_matching = !build_id().empty() && build_id() == info.build_id();
+  const bool all_module_properties_matching =
+      build_id() == info.build_id() && name() == info.name() && file_size() == info.file_size() &&
+      load_bias() == info.load_bias();
+
+  module_info_ = std::move(info);
+
+  if (build_id_matching) return;
+
+  if (all_module_properties_matching) return;
+
+  LOG("Module %s changed.", file_path());
+
+  if (!is_loaded_) return;
+
+  LOG("Module %s contained symbols. Because the module changed, those are now removed.",
+      file_path());
+  functions_.clear();
+  hash_to_function_map_.clear();
+  is_loaded_ = false;
 }
 
 const orbit_client_protos::FunctionInfo* ModuleData::FindFunctionByRelativeAddress(
@@ -61,7 +91,13 @@ void ModuleData::AddSymbols(const orbit_grpc_protos::ModuleSymbols& module_symbo
     // __cxxabiv1::__class_type_info::~__class_type_info()
     // __cxxabiv1::__pbase_type_info::~__pbase_type_info()
     if (success) {
-      hash_to_function_map_.emplace(FunctionUtils::GetHash(*function), function);
+      const bool success =
+          hash_to_function_map_.try_emplace(FunctionUtils::GetHash(*function), function).second;
+      if (!success) {
+        LOG("Warning: Multiple functions with the same demangled name: %s (this is currently not "
+            "supported by presets)",
+            function->pretty_name());
+      }
     } else {
       address_reuse_counter++;
     }
--- a/OrbitClientData/ModuleManager.cpp
+++ b/OrbitClientData/ModuleManager.cpp
@@ -15,9 +15,11 @@ using orbit_client_protos::FunctionInfo;
 
 namespace OrbitClientData {
 
-void ModuleManager::AddOrUpdateModules(
+std::vector<ModuleData*> ModuleManager::AddOrUpdateModules(
     const std::vector<orbit_grpc_protos::ModuleInfo>& module_infos) {
   absl::MutexLock lock(&mutex_);
+
+  std::vector<ModuleData*> unloaded_modules;
 
   for (const auto& module_info : module_infos) {
     auto module_it = module_map_.find(module_info.file_path());
@@ -25,9 +27,16 @@ void ModuleManager::AddOrUpdateModules(
       const bool success = module_map_.try_emplace(module_info.file_path(), module_info).second;
       CHECK(success);
     } else {
-      module_it->second.UpdateIfChanged(module_info);
+      ModuleData& module = module_it->second;
+      bool was_loaded = module.is_loaded();
+      module.UpdateIfChanged(module_info);
+      if (was_loaded && !module.is_loaded()) {
+        unloaded_modules.push_back(&module);
+      }
     }
   }
+
+  return unloaded_modules;
 }
 
 const ModuleData* ModuleManager::GetModuleByPath(const std::string& path) const {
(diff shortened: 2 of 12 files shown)


ronaldfw:
What happens to hooked functions in a module that is updated? In particular, what happens during a capture when a module is updated?

antonrohr:
> What happens to hooked functions in a module that is updated?

This is a good point. I added a commit (that will be squashed) to deal with this case. Now, when a module is updated and the symbols (functions) are discarded, the hooked functions are also discarded. 

> In particular, what happens during a capture when a module is updated?

Now we are really in undefined behaviour world. I dont even know if that is possible. I would guess that the main module (the binary that is profiled) cannot change while running. For other modules, I don't know what happens if they change while they are loaded. I would assume the process crashes when a module changes while its loaded. I guess the binary could unload (dlclose) a module, then the Orbit user pushes a new version of that module to the instance and then the binary loads (dlopen) that module again. All while taking a capture. A weird use case like that is definitely not supported by Orbit atm. These are things we should think about as soon as we start to support dlopen/dlclose while taking a capture.

OrbitClientData/ModuleData.cpp
```diff
@@ -8,14 +8,30 @@
...
 
 using orbit_client_protos::FunctionInfo;
+using orbit_grpc_protos::ModuleInfo;
 
 bool ModuleData::is_loaded() const {
   absl::MutexLock lock(&mutex_);
   return is_loaded_;
 }
 
+void ModuleData::Update(ModuleInfo info) {
```
dpallotti, line 21: Would it be better to call this `UpdateIfChanged` as the update doesn't necessarily happen?

OrbitClientData/ModuleManager.cpp
```diff
@@ -7,22 +7,25 @@
...
+void ModuleManager::AddOrUpdateModules(
+    const std::vector<orbit_grpc_protos::ModuleInfo>& module_infos) {
   absl::MutexLock lock(&mutex_);
 
   for (const auto& module_info : module_infos) {
     auto module_it = module_map_.find(module_info.file_path());
     if (module_it == module_map_.end()) {
-      const auto [inserted_it, success] =
-          module_map_.try_emplace(module_info.file_path(), module_info);
+      const auto [_, success] = module_map_.try_emplace(module_info.file_path(), module_info);
```
dpallotti, line 25: While in other languages `_` is part of the language or the standard for unused variables, I'm really not sold on this in C++. Why not just
```
const bool success = module_map_.try_emplace(module_info.file_path(), module_info).second;
```
?

dpallotti (approved):

OrbitClientData/ModuleData.cpp
```diff
@@ -8,14 +8,30 @@
...
+  absl::MutexLock lock(&mutex_);
+
+  CHECK(file_path() == info.file_path());
+
+  // if a property changed, remove functions
```
dimitry-, line 26: I would check build_id and only if it is empty - check the rest of properties instead. It could be that the file_size is changed but build_id is not (in the case we get another debug section stripped for example without a rebuild).

dimitry- (commented):
Ideally we do not want to modify/remove already loaded modules in case someone still relying on these. Can we do proper build_id (or say .text section hash if build_id is not available) module identification?

ronaldfw:
> These are things we should think about as soon as we start to support dlopen/dlclose while taking a capture.

Great, thanks for the details. We definitely want to support swapping out / reloading modules and Orbit should behave conveniently in that case, in particular, if you have hooked functions in a module that is reloaded, we should hook those functions again. 

Whether this needs to work _while_ taking a capture, I'm not entirely convinced it is that important. Let's chat about it more.

OrbitClientData/ModuleManager.cpp
ronaldfw: Not sure if I'm missing something here, but shouldn't this trigger an unused variable error?

(On the use of _: I don't feel strongly either way. I think it's fine if as a team we agree on this convention. In this particular case the use of .second works just as well for me, though.)

dpallotti: In C++ there is a series of rules that determine if an identifier starting with an underscore is reserved and shouldn't be used. This specific name for a local variable doesn't fall into one of the reserved category, but still, in light of this, the common approach is to just avoid identifiers that start with underscores (including this degenerate case), to avoid doubts. So I'm not agreeing on this convention, as convenient as it is in other languages.

antonrohr: I started using this underscore syntax a while ago. I wasn't sure about it, because the style guide does not mention it, hence I started a thread in chat about it and the conclusion was that it is fine when used in a local scope. Also, the compiler optimizes it away anyways. 
I use it mainly in for loops with hash maps and I think it improves readability
```
for (const auto& [_, value] : hash_map) { ... } 
```
instead of 
```
for (const auto& pair: hash_map) {
  const auto& value = pair.second;
  ...
}
```

That said, in this case I actually like your suggestion Davide and changed it.

OrbitClientData/ModuleData.cpp

OrbitClientData/ModuleData.cpp

ElfUtils/ElfFile.cpp
```diff
@@ -50,6 +50,7 @@ class ElfFileImpl : public ElfFile {
   std::string build_id_;
   bool has_symtab_section_;
   bool has_debug_info_section_;
+  llvm::ArrayRef<uint8_t> text_section_;
```
dimitry-, line 53: Doesn't this increase memory consumption? Maybe load it only when it is needed?

OrbitClientData/ModuleData.cpp
```diff
@@ -8,14 +8,40 @@
...
+    LOG("Module %s changed.", file_path());
+    if (is_loaded_) {
+      LOG("Module %s contained symbols. Because the module changed, those are now removed.",
+          file_path());
+      functions_.clear();
+      is_loaded_ = false;
+    }
+  }
+
+  module_info_ = std::move(info);
```
dimitry-, line 42: nit: maybe move this before the if and use early returns?

ElfUtils/ElfFile.cpp
```diff
@@ -89,6 +90,16 @@ void ElfFileImpl<ElfT>::InitSections() {
       has_debug_info_section_ = true;
     }
 
+    if (name.str() == ".text") {
+      llvm::Expected<llvm::ArrayRef<uint8_t>> text_section_contents =
+          elf_file->getSectionContents(&section);
+      if (!text_section_contents) {
+        LOG("Unable to get text section content");
+      } else {
+        text_section_ = text_section_contents.get();
```
dimitry-, line 99: it does not seem to be used atm?

OrbitClientData/ModuleManager.cpp
dpallotti: But then why not use an actual name instead of `_`, even if you're not using it? What's the problem with `for (const auto& [key, value] : hash_map) { ... }` instead of `for (const auto& [_, value] : hash_map) { ... } `?
`_` doesn't have a special meaning like in other languages, so you're still having an actual variable, which is now also weirdly named. This also means that if you do the same below or in a nested scope you actually have a redefinition, which goes directly against the purpose of this (which would be not declaring a variable at all).

ronaldfw: We discussed this before and agreed that it's fine. Unless there are strong (newly known) reasons (which I don't think there are), I'm fine if folks continue using it.

antonrohr: I found another resource which has an opinion about it. Taking this to the chat room

ElfUtils/ElfFile.cpp
antonrohr: Sorry, I played around with it and this was not supposed to be committed. I am going to remove it.

ElfUtils/ElfFile.cpp
antonrohr: Sorry for the mess. It is removed again in the latest commit. The rebase before merge will remove all traces of this.

OrbitClientData/ModuleData.cpp
antonrohr: Thank you!!! This was the hint I needed. ptal

dimitry- (approved):

antonrohr:
I added one other commit (Reload and Rehook symbols of changed modules), so after the PR will have 2 commits in the end, everything else will be squashed. 

This reloads changed modules and rehooks the functions after the reload. 

I did quite some testing with my own test binary which uses dlopen/dlclose, so I am confident that it works properly.

PR #1400, opened 2020-10-27
More verbose logging in E2E tests.

danielfenner:
This should simplify debugging in case one of the commands
runs into a timeout during nightly execution.

Also fixed missing inport in orbit_testing.py.

Bug: http://b/171784639
Test: Locally executed the e2e tests.

### DIFF
--- a/contrib/automation_tests/orbit_testing.py
+++ b/contrib/automation_tests/orbit_testing.py
@@ -6,16 +6,19 @@ found in the LICENSE file.
 
 import logging
 import time
+import pywinauto
 from pywinauto.application import Application
 
 def WaitForOrbit():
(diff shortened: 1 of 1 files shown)


contrib/automation_tests/orbit_testing.py
```diff
@@ -6,23 +6,27 @@
 
 import logging
 import time
+import pywinauto
```
danielfenner, line 9: @ronaldfw : Does this work for you with out this import?

ronaldfw: I ran all tests without issues. But, yes, should be imported here for sure.

ronaldfw (approved):
LGTM!

PR #1402, opened 2020-10-27
Render GraphTrack curves as step functions

pierricgimmig:
- labels don't snap to points anymore
- clamp label x position
- fix division by zero bug

### DIFF
--- a/OrbitGl/GraphTrack.cpp
+++ b/OrbitGl/GraphTrack.cpp
@@ -40,6 +40,7 @@ void GraphTrack::UpdatePrimitives(uint64_t min_tick, uint64_t max_tick, PickingM
     double last_normalized_value = (it->second - min_) * inv_value_range_;
     constexpr float kDotRadius = 2.f;
     float base_y = pos_[1] - size_[1];
+    float y1 = 0;
     DrawSquareDot(batcher,
                   Vec2(time_graph_->GetWorldFromTick(previous_time),
                        base_y + static_cast<float>(last_normalized_value) * size_[1]),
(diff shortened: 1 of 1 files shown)


OrbitGl/GraphTrack.cpp
```diff
@@ -53,8 +53,9 @@ void GraphTrack::UpdatePrimitives(uint64_t min_tick, uint64_t max_tick, PickingM
       float x1 = time_graph_->GetWorldFromTick(time);
       float y0 = base_y + static_cast<float>(last_normalized_value) * size_[1];
       float y1 = base_y + static_cast<float>(normalized_value) * size_[1];
-      batcher->AddLine(Vec2(x0, y0), Vec2(x1, y1), graph_z, kLineColor);
-      DrawSquareDot(batcher, Vec2(x1, y1), kDotRadius, dot_z, kDotColor);
+      batcher->AddLine(Vec2(x0, y0), Vec2(x1, y0), graph_z, kLineColor);
+      batcher->AddLine(Vec2(x1, y0), Vec2(x1, y1), graph_z, kLineColor);
+      DrawSquareDot(batcher, Vec2(x1, y0), kDotRadius, dot_z, kDotColor);
```
karupayun, line 58: The square dot has to be in (x1, y1).

karupayun (approved):
Thanks for fixing this Pierric!

I leaved a comment and I think that we want to continue the blue lines at the end of the capture to be consistent.

Screen Shot 2020-10-29 at 14 06 19

OrbitGl/GraphTrack.cpp
pierricgimmig: yeah you're right! thanks!

pierricgimmig:
> I think that we want to continue the blue lines at the end of the capture to be consistent.

I agree, done. Thanks Pablo!

PR #1404, opened 2020-10-28
A couple of fixes to the collector, in particular Skip events coming before all file descriptors have been enabled

dpallotti:
#### Replace fds_per_cpu_ with a parameter
It's a class member used to pass data from a method to another, make it a local
variable and pass it as a parameter.
#### Use perf_event_open with PERF_FLAG_FD_CLOEXEC
See http://b/170711685#comment17
#### PerfEventProcessor discards events that would be out of order
Count the events discarded in `EventStats`, don't log them as if one event is
discarded, usually hundreds of events are discarded, which would flood the logs.
#### Skip events coming before all file descriptors have been enabled
Enabling the perf_event_open file descriptors can take a while when we have a
lot of them (and also once we go back to allowing to start a new capture while
the file descriptors from the previous ones are still being closed).
Some events can start coming to the ring buffers from the first file
descriptors to have been enabled, while we are still enabling other file
descriptors and hence before we start reading from the ring buffers.
`PerfEventProcessor` would have to buffer for all this arbitrarily long interval
in order to guarantee events to be processed in order.
Therefore, we discard events that came before the last file descriptor had been
enabled.
We prefer this to simply enabling all file descriptors in a different thread, as
we want to perform additional work after enabling the file descriptors but
before starting to read from the ring buffers.
This also prevents the beginning of the capture to be missing parts of
information as some events are already being recorded and others aren't.
Also refer to http://b/170711685#comment4.

Bug: http://b/170711685

Test: Capture Trata hooking all the `vk*` functions in `libvulkan.so.1`,
see from the logs that no event is out of order.
#### Tweak TracerThread::Reset()

### DIFF
--- a/OrbitLinuxTracing/PerfEventProcessor.cpp
+++ b/OrbitLinuxTracing/PerfEventProcessor.cpp
@@ -56,12 +56,12 @@ std::unique_ptr<PerfEvent> PerfEventQueue::PopEvent() {
 }
 
 void PerfEventProcessor::AddEvent(int origin_fd, std::unique_ptr<PerfEvent> event) {
-#ifndef NDEBUG
-  if (last_processed_timestamp_ > 0 &&
-      event->GetTimestamp() < last_processed_timestamp_ - kProcessingDelayMs * 1'000'000) {
-    ERROR("Processed an event out of order");
+  if (last_processed_timestamp_ns_ > 0 && event->GetTimestamp() < last_processed_timestamp_ns_) {
+    if (discarded_out_of_order_counter_ != nullptr) {
+      ++(*discarded_out_of_order_counter_);
+    }
+    return;
   }
-#endif
   event_queue_.PushEvent(origin_fd, std::move(event));
 }
 
@@ -69,9 +69,10 @@ void PerfEventProcessor::ProcessAllEvents() {
   CHECK(!visitors_.empty());
   while (event_queue_.HasEvent()) {
     std::unique_ptr<PerfEvent> event = event_queue_.PopEvent();
-#ifndef NDEBUG
-    last_processed_timestamp_ = event->GetTimestamp();
-#endif
+    // Events are guaranteed to be processed in order of timestamp
+    // as out-of-order events are discarded in AddEvent.
+    CHECK(event->GetTimestamp() >= last_processed_timestamp_ns_);
+    last_processed_timestamp_ns_ = event->GetTimestamp();
     for (PerfEventVisitor* visitor : visitors_) {
       event->Accept(visitor);
     }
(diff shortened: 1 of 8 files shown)


OrbitLinuxTracing/PerfEventProcessor.cpp
```diff
@@ -56,22 +56,21 @@ std::unique_ptr<PerfEvent> PerfEventQueue::PopEvent() {
...
 }
 
 void PerfEventProcessor::ProcessAllEvents() {
   CHECK(!visitors_.empty());
   while (event_queue_.HasEvent()) {
     std::unique_ptr<PerfEvent> event = event_queue_.PopEvent();
-#ifndef NDEBUG
-    last_processed_timestamp_ = event->GetTimestamp();
-#endif
+    CHECK(event->GetTimestamp() >= last_processed_timestamp_ns_);
```
ronaldfw, line 72: I would add a comment why this CHECK must be true (events are discarded in AddEvent).

OrbitLinuxTracing/PerfEventProcessor.cpp
```diff
@@ -80,19 +79,18 @@ void PerfEventProcessor::ProcessAllEvents() {
...
   while (event_queue_.HasEvent()) {
     PerfEvent* event = event_queue_.TopEvent();
 
-    // Do not read the most recent events as out-of-order events could arrive.
-    if (event->GetTimestamp() + kProcessingDelayMs * 1'000'000 >= max_timestamp) {
+    // Do not read the most recent events as out-of-order events could (and will) arrive.
+    if (event->GetTimestamp() + kProcessingDelayMs * 1'000'000 >= current_timestamp_ns) {
       break;
     }
+    CHECK(event->GetTimestamp() >= last_processed_timestamp_ns_);
```
ronaldfw, line 91: Same here.

OrbitLinuxTracing/PerfEventProcessor.h
```diff
@@ -78,19 +78,22 @@ class PerfEventProcessor {
 
   void ClearVisitors() { visitors_.clear(); }
 
+  void SetDiscardedOutOfOrderCounter(
```
ronaldfw, line 81: I think this needs a comment why this counter is passed as a shared pointer.

ronaldfw (approved):
LGTM, some smaller comments to discuss.

OrbitLinuxTracing/TracerThread.h
```diff
@@ -178,6 +180,8 @@ class TracerThread {
     uint64_t gpu_events_count = 0;
     uint64_t lost_count = 0;
     absl::flat_hash_map<PerfEventRingBuffer*, uint64_t> lost_count_per_buffer{};
+    std::shared_ptr<std::atomic<uint64_t>> discarded_out_of_order_count =
+        std::make_unique<std::atomic<uint64_t>>(0);
```
florian-kuebler, line 186: `shared_ptr... = make_unique`... is this intended?

OrbitLinuxTracing/PerfEventProcessor.h
florian-kuebler: I would rather just pass a bare pointer here.

florian-kuebler (commented):
Thanks for the change. If have only concerns of using a shared_ptr<atomic<int>> for the counter. Other then that, LGTM.

OrbitLinuxTracing/TracerThread.h
dpallotti: Nope, it was wrong in the lines below and I copy-pasted the mistake, go me. Anyway, switched away from the `shared_ptr`s altogether.

florian-kuebler (approved):
LGTM. Thanks!

OrbitLinuxTracing/PerfEventProcessor.h
florian-kuebler: Thanks for changing that. From my side, looks good :)

OrbitLinuxTracing/TracerThread.h
florian-kuebler: Thanks!

PR #1405, opened 2020-10-28
Close perf_event_open file descriptors in a separate thread

dpallotti:
Closing those file descriptors can take a long time, but all data has already
been correctly processed and sent.
Note that if a new capture is started while those file descriptors are still
being closed, it can take a while for that capture to be effectively started and
for data to come in. But this is the behavior we had when the file descriptors
were closed by the forked `top` process (see explanation in
http://b/170711685#comment17). So let's quickly fix this regression.

Bug: http://b/171350567

Test: Both locally and on Trata, try taking multiple captures in a row when
instrumenting a relatively large number of functions. Verify that the behavior
is like the one prior to the removal of `top`.

---

As discussed, let's switch back to the previous behavior before possibly
fixing the issue on the service or properly handling it in the client.

### DIFF
--- a/OrbitLinuxTracing/TracerThread.cpp
+++ b/OrbitLinuxTracing/TracerThread.cpp
@@ -731,13 +731,20 @@ void TracerThread::Run(const std::shared_ptr<std::atomic<bool>>& exit_requested)
   }
 
   // Close the file descriptors.
-  {
+  // Do it in a separate dedicated thread as this can take a long time
+  // with a lot of uprobes and uretprobes.
+  std::thread([tracing_fds = tracing_fds_] {
+    pthread_setname_np(pthread_self(), "close(fd)s");
     ORBIT_SCOPE_WITH_COLOR("Closing file descriptors", orbit::Color::kRed);
-    for (int fd : tracing_fds_) {
+    uint64_t fd_close_begin_ns = MonotonicTimestampNs();
+    for (int fd : tracing_fds) {
       ORBIT_SCOPE("Closing fd");
       close(fd);
     }
-  }
+    uint64_t fd_close_end_ns = MonotonicTimestampNs();
+    LOG("Finished closing %lu perf_event_open file descriptors after %.3f ms", tracing_fds.size(),
+        (fd_close_end_ns - fd_close_begin_ns) / 1e6);
+  }).detach();
 }
 
 void TracerThread::ProcessContextSwitchCpuWideEvent(const perf_event_header& header,


OrbitLinuxTracing/TracerThread.cpp
```diff
@@ -731,13 +731,20 @@ void TracerThread::Run(const std::shared_ptr<std::atomic<bool>>& exit_requested)
...
 
   // Close the file descriptors.
-  {
+  // Do it in a separate dedicated thread as this can take a long time
+  // with a lot of uprobes and uretprobes.
+  std::thread([tracing_fds = tracing_fds_] {
+    pthread_setname_np(pthread_self(), "close(fd)s");
     ORBIT_SCOPE_WITH_COLOR("Closing file descriptors", orbit::Color::kRed);
-    for (int fd : tracing_fds_) {
+    uint64_t fd_close_begin_ns = MonotonicTimestampNs();
```
ronaldfw, line 739: Why the additional timing if we have the introspection scope already? Do you always want to have this in the logs?

dpallotti: I want it in the logs.

dimitry-: Don't we have a similar marco for the logs?

OrbitLinuxTracing/TracerThread.cpp
```diff
@@ -731,13 +731,20 @@ void TracerThread::Run(const std::shared_ptr<std::atomic<bool>>& exit_requested)
...
+    for (int fd : tracing_fds) {
       ORBIT_SCOPE("Closing fd");
       close(fd);
     }
-  }
+    uint64_t fd_close_end_ns = MonotonicTimestampNs();
+    LOG("Finished closing %lu perf_event_open file descriptors after %.3f ms", tracing_fds.size(),
+        (fd_close_end_ns - fd_close_begin_ns) / 1e6);
+  }).detach();
```
dimitry-, line 747: I am not sure detach here is a good idea... maybe use thread_pool for this?

OrbitLinuxTracing/TracerThread.cpp
dpallotti: Do we? And with a custom message like done here?

pierricgimmig: We have SCOPE_TIMER_LOG but it's in OrbitCore, it would be cool to have it in OrbitBase

OrbitLinuxTracing/TracerThread.cpp
pierricgimmig: We could use a static thread pool so that we would not spawn a new thread on every capture stop, but reuse the same one. If anything, it would mean that we would have a single introspection track in the UI instead of many.

pierricgimmig (approved):
As mentioned in the comment, I'm also not a huge fan of singe use threads, but it's not a huge deal.

OrbitLinuxTracing/TracerThread.cpp
dpallotti: I mean, sure, I can just add a `static` thread pool as local variable. But apart from the number of introspection tracks, which I don't think should drive this decision, but is still something, I don't understand the arguments in favor of a thread pool. Thread pools are used for multiple and frequent short-lived tasks, and this is the exact opposite of it, with the cost of creating a new thread being irrelevant (and instead having an additional thread pool with one thread always alive).

dimitry-: just to clarify why... detached thread is basically we do not know the scope of this thread - main thread can be dead at any moment which means some shared resources could become unreachable with its resources. TBH I do not even know if there is a valid case when detach should be used.

In other words someone need to manage this thread scope. Another approach is to remove the threading from here and delegate it to the caller via FileDescriptorCloser functor/interface if this helps to manage the scope better (not in this component).

dpallotti: > main thread can be dead at any moment which means some shared resources could become unavailable.

What shared resources are we talking about in this specific case?

OrbitLinuxTracing/TracerThread.cpp
```diff
@@ -731,13 +731,20 @@ void TracerThread::Run(const std::shared_ptr<std::atomic<bool>>& exit_requested)
   }
 
   // Close the file descriptors.
-  {
+  // Do it in a separate dedicated thread as this can take a long time
+  // with a lot of uprobes and uretprobes.
+  std::thread([tracing_fds = tracing_fds_] {
+    pthread_setname_np(pthread_self(), "close(fd)s");
     ORBIT_SCOPE_WITH_COLOR("Closing file descriptors", orbit::Color::kRed);
```
dimitry-, line 738: this macro might use some resource which could disappear if main thread no longer exists.

OrbitLinuxTracing/TracerThread.cpp
dimitry-: log file descriptor?

dpallotti: Great, so it seems that the only thing to do is to remove logging from here! This is a joke, but only half a joke.

It's only half a joke because this is supposed to be just a fix to the regression. Having a detached thread and no logging would then do exactly that. If we want to or need to properly handle synchronization and all, then at this point we should maybe just do it properly and communicate with the ui, at least we get something out of it.

dpallotti:
Closing this based on http://b/171350567#comment6.

PR #1406, opened 2020-10-28
Draw in increasing z-value order in Capture View

karupayun:
Following the discussion in (https://github.com/google/orbit/pull/1119), we draw using layers to avoid the problem of transparency items who, when drawn before, we don't see what there are behind them.

For that, we add a layer to every box, triangle, line and text.

Mainly focusing b/167369762, but also b/167515371, b/170701548.

In the image: Now we can see rounding corners and text behind the transparency overlay.
transparency

Ps: I recommend follow commit by commit.

### DIFF
--- a/OrbitGl/Batcher.cpp
+++ b/OrbitGl/Batcher.cpp
@@ -243,7 +243,7 @@ std::vector<float> Batcher::GetLayers() const {
   return layers;
 };
 
-void Batcher::Draw(float layer, bool picking) const {
+void Batcher::DrawLayer(float layer, bool picking) const {
   if (!primitive_buffers_by_layer_.count(layer)) return;
   glPushAttrib(GL_ENABLE_BIT | GL_COLOR_BUFFER_BIT);
   if (picking) {
@@ -264,6 +264,12 @@ void Batcher::Draw(float layer, bool picking) const {
   glDisableClientState(GL_COLOR_ARRAY);
   glDisableClientState(GL_VERTEX_ARRAY);
   glPopAttrib();
+}
+
+void Batcher::Draw(bool picking) const {
+  for (auto& [layer, unused_buffer] : primitive_buffers_by_layer_) {
+    DrawLayer(layer, picking);
+  }
 }
 
 void Batcher::DrawBoxBuffer(float layer, bool picking) const {
--- a/OrbitGl/Batcher.cpp
+++ b/OrbitGl/Batcher.cpp
@@ -235,10 +235,10 @@ void Batcher::StartNewFrame() {
   user_data_.clear();
 }
 
-std::set<float> Batcher::GetLayers() const {
-  std::set<float> layers;
+std::vector<float> Batcher::GetLayers() const {
+  std::vector<float> layers;
   for (auto& [layer, _] : primitive_buffers_by_layer_) {
-    layers.insert(layer);
+    layers.push_back(layer);
   }
   return layers;
 };
--- a/OrbitGl/Batcher.cpp
+++ b/OrbitGl/Batcher.cpp
@@ -42,10 +42,11 @@ void Batcher::AddLine(Vec2 from, Vec2 to, float z, const Color& color, const Col
   Line line;
   line.start_point = Vec3(from[0], from[1], z);
   line.end_point = Vec3(to[0], to[1], z);
+  auto& buffer = primitive_buffers_by_layer_[z];
 
-  line_buffer_.lines_.push_back(line);
-  line_buffer_.colors_.push_back_n(color, 2);
-  line_buffer_.picking_colors_.push_back_n(picking_color, 2);
+  buffer.line_buffer.lines_.push_back(line);
+  buffer.line_buffer.colors_.push_back_n(color, 2);
+  buffer.line_buffer.picking_colors_.push_back_n(picking_color, 2);
   user_data_.push_back(std::move(user_data));
 }
 
@@ -102,9 +103,11 @@ void Batcher::AddShadedBox(Vec2 pos, Vec2 size, float z, const Color& color,
 
 void Batcher::AddBox(const Box& box, const std::array<Color, 4>& colors, const Color& picking_color,
                      std::unique_ptr<PickingUserData> user_data) {
-  box_buffer_.boxes_.push_back(box);
-  box_buffer_.colors_.push_back(colors);
-  box_buffer_.picking_colors_.push_back_n(picking_color, 4);
+  float layer_z_value = box.vertices[0][2];
+  auto& buffer = primitive_buffers_by_layer_[layer_z_value];
+  buffer.box_buffer.boxes_.push_back(box);
+  buffer.box_buffer.colors_.push_back(colors);
+  buffer.box_buffer.picking_colors_.push_back_n(picking_color, 4);
   user_data_.push_back(std::move(user_data));
 }
 
@@ -126,9 +129,11 @@ void Batcher::AddTriangle(const Triangle& triangle, const Color& color,
 
 void Batcher::AddTriangle(const Triangle& triangle, const Color& color, const Color& picking_color,
                           std::unique_ptr<PickingUserData> user_data) {
-  triangle_buffer_.triangles_.push_back(triangle);
-  triangle_buffer_.colors_.push_back_n(color, 3);
-  triangle_buffer_.picking_colors_.push_back_n(picking_color, 3);
+  float layer_z_value = triangle.vertices[0][2];
+  auto& buffer = primitive_buffers_by_layer_[layer_z_value];
+  buffer.triangle_buffer.triangles_.push_back(triangle);
+  buffer.triangle_buffer.colors_.push_back_n(color, 3);
+  buffer.triangle_buffer.picking_colors_.push_back_n(picking_color, 3);
   user_data_.push_back(std::move(user_data));
 }
 
@@ -220,9 +225,9 @@ void Batcher::GetBoxGradientColors(const Color& color, std::array<Color, 4>* col
 }
 
 void Batcher::ResetElements() {
-  line_buffer_.Reset();
-  box_buffer_.Reset();
-  triangle_buffer_.Reset();
+  for (auto& [unused_layer, buffer] : primitive_buffers_by_layer_) {
+    buffer.Reset();
+  }
 }
 
 void Batcher::StartNewFrame() {
@@ -230,7 +235,16 @@ void Batcher::StartNewFrame() {
   user_data_.clear();
 }
 
-void Batcher::Draw(bool picking) const {
+std::set<float> Batcher::GetLayers() const {
+  std::set<float> layers;
+  for (auto& [layer, _] : primitive_buffers_by_layer_) {
+    layers.insert(layer);
+  }
+  return layers;
+};
+
+void Batcher::Draw(float layer, bool picking) const {
+  if (!primitive_buffers_by_layer_.count(layer)) return;
   glPushAttrib(GL_ENABLE_BIT | GL_COLOR_BUFFER_BIT);
   if (picking) {
     glDisable(GL_BLEND);
@@ -243,20 +257,21 @@ void Batcher::Draw(bool picking) const {
   glEnableClientState(GL_COLOR_ARRAY);
   glEnable(GL_TEXTURE_2D);
 
-  DrawBoxBuffer(picking);
-  DrawLineBuffer(picking);
-  DrawTriangleBuffer(picking);
+  DrawBoxBuffer(layer, picking);
+  DrawLineBuffer(layer, picking);
+  DrawTriangleBuffer(layer, picking);
 
   glDisableClientState(GL_COLOR_ARRAY);
   glDisableClientState(GL_VERTEX_ARRAY);
   glPopAttrib();
 }
 
-void Batcher::DrawBoxBuffer(bool picking) const {
-  const Block<Box, BoxBuffer::NUM_BOXES_PER_BLOCK>* box_block = box_buffer_.boxes_.root();
+void Batcher::DrawBoxBuffer(float layer, bool picking) const {
+  auto& box_buffer = primitive_buffers_by_layer_.at(layer).box_buffer;
+  const Block<Box, BoxBuffer::NUM_BOXES_PER_BLOCK>* box_block = box_buffer.boxes_.root();
   const Block<Color, BoxBuffer::NUM_BOXES_PER_BLOCK * 4>* color_block;
 
-  color_block = !picking ? box_buffer_.colors_.root() : box_buffer_.picking_colors_.root();
+  color_block = !picking ? box_buffer.colors_.root() : box_buffer.picking_colors_.root();
 
   while (box_block != nullptr) {
     if (auto num_elems = box_block->size()) {
@@ -270,11 +285,12 @@ void Batcher::DrawBoxBuffer(bool picking) const {
   }
 }
 
-void Batcher::DrawLineBuffer(bool picking) const {
-  const Block<Line, LineBuffer::NUM_LINES_PER_BLOCK>* line_block = line_buffer_.lines_.root();
+void Batcher::DrawLineBuffer(float layer, bool picking) const {
+  auto& line_buffer = primitive_buffers_by_layer_.at(layer).line_buffer;
+  const Block<Line, LineBuffer::NUM_LINES_PER_BLOCK>* line_block = line_buffer.lines_.root();
   const Block<Color, LineBuffer::NUM_LINES_PER_BLOCK * 2>* color_block;
 
-  color_block = !picking ? line_buffer_.colors_.root() : line_buffer_.picking_colors_.root();
+  color_block = !picking ? line_buffer.colors_.root() : line_buffer.picking_colors_.root();
 
   while (line_block != nullptr) {
     if (auto num_elems = line_block->size()) {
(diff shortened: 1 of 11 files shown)


karupayun:
Seems that I have to modify BatcherTest. I'll do that tomorrow.

OrbitGl/Batcher.cpp
```diff
@@ -102,9 +103,11 @@ void Batcher::AddShadedBox(Vec2 pos, Vec2 size, float z, const Color& color,
 
 void Batcher::AddBox(const Box& box, const std::array<Color, 4>& colors, const Color& picking_color,
                      std::unique_ptr<PickingUserData> user_data) {
-  box_buffer_.boxes_.push_back(box);
-  box_buffer_.colors_.push_back(colors);
-  box_buffer_.picking_colors_.push_back_n(picking_color, 4);
+  float layer = box.vertices[0][2];
```
ronaldfw, line 106: Nit: layer_z_value?

OrbitGl/Batcher.cpp
```diff
@@ -126,9 +129,11 @@ void Batcher::AddTriangle(const Triangle& triangle, const Color& color,
 
 void Batcher::AddTriangle(const Triangle& triangle, const Color& color, const Color& picking_color,
                           std::unique_ptr<PickingUserData> user_data) {
-  triangle_buffer_.triangles_.push_back(triangle);
-  triangle_buffer_.colors_.push_back_n(color, 3);
-  triangle_buffer_.picking_colors_.push_back_n(picking_color, 3);
+  float layer = triangle.vertices[0][2];
```
ronaldfw, line 132: Same nit: layer_z_value

OrbitGl/Batcher.cpp
ronaldfw, line 106: There's an assumption here that all vertices have the same z value, which is indeed true in our case. It might be good to add a comment somewhere about this assumption. (In the Batcher.h header file?)

OrbitGl/Batcher.cpp
```diff
@@ -220,17 +225,28 @@ void Batcher::GetBoxGradientColors(const Color& color, std::array<Color, 4>* col
 }
 
 void Batcher::ResetElements() {
-  line_buffer_.Reset();
-  box_buffer_.Reset();
-  triangle_buffer_.Reset();
+  for (auto& [_, buffer] : buffers_by_layer_) {
+    buffer.line_buffer.Reset();
```
ronaldfw, line 229: Consider defining a Reset() method on the Buffers type so you don't have to call all three Reset methods here (which is a minor inconvenience, but potentially a bit more error prone).

OrbitGl/Batcher.cpp
```diff
@@ -220,17 +225,28 @@ void Batcher::GetBoxGradientColors(const Color& color, std::array<Color, 4>* col
...
+  }
 }
 
 void Batcher::StartNewFrame() {
   ResetElements();
   user_data_.clear();
 }
 
-void Batcher::Draw(bool picking) const {
+std::set<float> Batcher::GetLayers() const {
```
ronaldfw, line 240: Why does this return a set? Do you need this to be a set? If not, use a vector (which should be the default choice unless you need something different).

OrbitGl/Batcher.h
```diff
@@ -142,9 +149,7 @@ class Batcher {
 
   BatcherId batcher_id_;
   PickingManager* picking_manager_;
-  LineBuffer line_buffer_;
-  BoxBuffer box_buffer_;
-  TriangleBuffer triangle_buffer_;
+  std::unordered_map<float, Buffers> buffers_by_layer_;
```
ronaldfw, line 152: We use absl::flat_hash_map as the default type for maps (unless you have good reasons not to, say performance metrics that prove you have significantly better performance using std::unordered_map).

OrbitGl/CaptureWindow.cpp
```diff
@@ -515,6 +515,35 @@ void CaptureWindow::Draw() {
...
+    layers.insert(layer);
+  }
+  for (auto layer : text_renderer_.GetLayers()) {
+    layers.insert(layer);
+  }
+  for (auto layer : time_graph_.GetTextRenderer()->GetLayers()) {
+    layers.insert(layer);
+  }
+
+  const float kScreenSpaceMinimumLayer = 0.8f;
```
ronaldfw, line 532: This should be defined along with the other z values.

OrbitGl/CaptureWindow.cpp
```diff
@@ -515,6 +515,35 @@ void CaptureWindow::Draw() {
...
+  for (float layer : layers) {
+    // We use different coordinate system for ScreenSpace items (margin, scrollbar, ...)
```
ronaldfw, line 534: Some typos here. Replace with:

// We use different coordinate systems for ScreenSpace items (margin, scrollbar, ...)
// and for the text than the rest of items inside CaptureWindow. So, we have to switch
// between these 2 systems while the layer is changing (with these "Prepare.." functions).

OrbitGl/GlCanvas.cpp
```diff
@@ -27,11 +27,12 @@ float GlCanvas::kZValueOverlayTextBackground = 0.45f;
 float GlCanvas::kZValueText = 0.47f;
 float GlCanvas::kZValueEventBarPicking = 0.49f;
 float GlCanvas::kZValueUi = 0.61f;
-float GlCanvas::kZValueTimeBarBg = 0.63f;
-float GlCanvas::kZValueTextUi = 0.65f;
-float GlCanvas::kZValueMargin = 0.81f;
-float GlCanvas::kZValueSliderBg = 0.83f;
-float GlCanvas::kZValueSlider = 0.85f;
+float GlCanvas::kZValueTextUi = 0.61f;
```
ronaldfw, line 30: Define the "cut point" between capture window space and screenspace here.

OrbitGl/TextRenderer.cpp
```diff
@@ -51,7 +51,7 @@ void TextRenderer::Init() {
   const auto exe_dir = Path::GetExecutableDir();
   const auto fontFileName = exe_dir + "fonts/Vera.ttf";
 
-  m_Buffer = vertex_buffer_new("vertex:3f,tex_coord:2f,color:4f");
+  //  m_Buffer = vertex_buffer_new("vertex:3f,tex_coord:2f,color:4f");
```
ronaldfw, line 54: Remove code that is commented out.

OrbitGl/TextRenderer.cpp
```diff
@@ -332,6 +336,14 @@ int TextRenderer::GetStringWidth(const char* a_Text) const {
   return static_cast<int>(ceil(stringWidth));
 }
 
+std::set<float> TextRenderer::GetLayers() const {
```
ronaldfw, line 339: Same as elsewhere: Does this need to be a set?

OrbitGl/TextRenderer.h
```diff
@@ -52,7 +52,7 @@ class TextRenderer {
 
  private:
   texture_atlas_t* m_Atlas;
-  vertex_buffer_t* m_Buffer;
+  std::unordered_map<float, vertex_buffer_t*> buffers_by_layer_;
```
ronaldfw, line 55: Use absl::flat_hash_map as mentioned elsewhere.

ronaldfw (commented):
Generally, this looks great, but there's one thing I'd like to discuss (maybe something to do later): Right now it's entirely possible to break the whole setup here (and cause terrible performance) by using an abundance of different z values that are not among the defined list, causing a huge number of batchers / buffers. We have no mechanism that guards against this mistake. One suggestion could be to define an enum for all the possible layers, and then only map to the actual z values in the batcher internally, so the z value purely becomes an implementation details of the batcher.

pierricgimmig:
> One suggestion could be to define an enum for all the possible layers

I agree with Ronald, I find it a bit weird to interpret floating point values as layers. I think it might make the track re-ordering feature look weird when we are dragging tracks around, but we could find a workaround later. 

Actually, another suggestion would be to disable depth-testing altogether, the order in which we add primitives (back-to-front) would determine precedence. We could save a non-negligible amount memory too by discarding the z-component of our primitives. We would draw the tracks' background, the timers, then the text and finally the overlay.  WDYT @karupayun , @ronaldfw , @reichlfl ?

OrbitGl/Batcher.cpp
karupayun: I'm thinking in making another PR changing the way we construct the Triangles, WDYT?

OrbitGl/Batcher.cpp
karupayun: Because we don't want to have repeated layers, but we can use a vector with sort and unique. I will change to that.

OrbitGl/Batcher.h
karupayun: We have an issue with absl::flat_hash_map because we don't have a move constructor in Blockchain. So, either we have to implement a move constructor there or we have to use unordered_map. For now, I will leave the unordered_map.

OrbitGl/TextRenderer.cpp
karupayun: Oh, sorry for that!

karupayun:
So, after a talk with Pierric we decided to pause this PR until Wednesday and trying to get something simpler by drawing transparent things at the end. I will work on that.

ronaldfw:
> So, after a talk with Pierric we decided to pause this PR until Wednesday and trying to get something simpler by drawing transparent things at the end. I will work on that.

Be aware that there are multiple layers of transparency if we want to get this completely right. You can't just draw stuff and then draw transparent stuff at the end. There is non-transparent stuff in between the transparent layers (text mainly) that thus should only be tinted by the transparent layers above them.

ronaldfw:
> > > One suggestion could be to define an enum for all the possible layers
> > 
> > 
> > I agree with Ronald, I find it a bit weird to interpret floating point values as layers. I think it might make the track re-ordering feature look weird when we are dragging tracks around, but we could find a workaround later.
> > Actually, another suggestion would be to disable depth-testing altogether, the order in which we add primitives (back-to-front) would determine precedence. We could save a non-negligible amount memory too by discarding the z-component of our primitives. We would draw the tracks' background, the timers, then the text and finally the overlay. WDYT @karupayun , @ronaldfw , @reichlfl ?
> 
> I don't have too much experience about how this things are generally done. I remembered that @reichlfl preferred that approach and I also found it clearer. I'm not sure why we are using a 3D way to print 2D elements. So, no z-value and draw in the same order as elements appear.
> 
> But currently it's not that simple. We have text who has to be on top of overlay and text who has to be behind. We have timegraph batcher which is not being calculated again after every mouse movement but also ui_batcher which is drawn much more frequently (for example for the green line). We also have some elements who has to be drawn by a coordinate system and others who follow a different one.
> 
> BTW, my plan here was erase some layers but I agree that there is no way to protect against explosion of layers.

My main argument for going with the layers by z value is that it's fairly easy to determine locally for an element where it ends up from a layer perspective. If you rely entirely on order, you have to get that order right globally. Maybe that's easy, I'm not sure.

ronaldfw:
> Actually, another suggestion would be to disable depth-testing altogether, the order in which we add primitives (back-to-front) would determine precedence. We could save a non-negligible amount memory too by discarding the z-component of our primitives. We would draw the tracks' background, the timers, then the text and finally the overlay. WDYT @karupayun , @ronaldfw , @reichlfl ?

To add another comment here, as there seems to be a misunderstanding about the proposed change here: With this change, we can actually disable depth-testing and can also discard the z-component when the element is getting added to the right batcher. However, I like that it's possible to determine locally for a primitive into which layer it is added, and using the z value of the primitive added is one way of determining the layer (and adding to the right batcher) and not rely on the order in which primitives are added. 

And just to finish the description:
> We would draw the tracks' background, the timers, then the text and finally the overlay.

This is exactly what is done here once all elements have been added to the different batchers (modulo that the order is a bit more complicated with multiple overlays and text in-between overlays). 

So, I think the only difference to what you are describing is how the order is determined in the first place: by insertion order (which I think may be hard to control) or by indicating the layer (which is done here via the z value, but could also be an enum that is passed to the batcher along with each, then 2D, primitive).

OrbitGl/Batcher.cpp
```diff
@@ -220,17 +225,26 @@ void Batcher::GetBoxGradientColors(const Color& color, std::array<Color, 4>* col
 }
 
 void Batcher::ResetElements() {
-  line_buffer_.Reset();
-  box_buffer_.Reset();
-  triangle_buffer_.Reset();
+  for (auto& [_, buffer] : buffers_by_layer_) {
```
pierricgimmig, line 228: nit. replace "_" by "unused_layer"

OrbitGl/Batcher.cpp
```diff
@@ -220,17 +225,26 @@ void Batcher::GetBoxGradientColors(const Color& color, std::array<Color, 4>* col
...
 
 void Batcher::StartNewFrame() {
   ResetElements();
   user_data_.clear();
 }
 
-void Batcher::Draw(bool picking) const {
+std::vector<float> Batcher::GetLayers() const {
+  std::vector<float> layers;
+  for (auto& [layer, _] : buffers_by_layer_) {
```
pierricgimmig, line 240: nit. "_" -> "unused_buffer"

OrbitGl/Batcher.h
```diff
@@ -63,6 +63,18 @@ struct TriangleBuffer {
   BlockChain<Color, 3 * NUM_TRIANGLES_PER_BLOCK> picking_colors_;
 };
 
+struct Buffers {
```
pierricgimmig, line 66: nit. "Buffers" is very generic, consider renaming.

OrbitGl/BatcherTest.cpp
```diff
@@ -28,46 +28,50 @@ class MockBatcher : public Batcher {
...
+          drawn_line_colors_.push_back(*it);
+          ++it;
+          ++it;
+        }
+        for (auto it = buffer.triangle_buffer.picking_colors_.begin();
+             it != buffer.triangle_buffer.picking_colors_.end();) {
+          drawn_triangle_colors_.push_back(*it);
+          ++it;
+          ++it;
+          ++it;
```
pierricgimmig, line 44: nit. the code was probably already there, but can't we do "it+=2" and "it+=3" here?

OrbitGl/CaptureWindow.cpp
```diff
@@ -515,6 +515,34 @@ void CaptureWindow::Draw() {
...
+  DrawScreenSpace();
+
+  // We start by getting all layers
+  std::vector<float> all_layers = time_graph_.GetBatcher().GetLayers();
+  auto ui_layers = ui_batcher_.GetLayers();
+  all_layers.insert(all_layers.end(), ui_layers.begin(), ui_layers.end());
+  auto text_ui_layers = text_renderer_.GetLayers();
+  all_layers.insert(all_layers.end(), text_ui_layers.begin(), text_ui_layers.end());
+  auto text_tg_layers = time_graph_.GetTextRenderer()->GetLayers();
+  all_layers.insert(all_layers.end(), text_tg_layers.begin(), text_tg_layers.end());
```
pierricgimmig, line 527: nit. could be replaced by: 
Append(all_layers, ui_batcher_.GetLayers());
Append(all_layers, text_renderer_.GetLayers());
Append(all_layers, time_graph_.GetTextRenderer()->GetLayers());

It's our own util method that wraps the not-so-eye-pleasing std api. I find it much more readable.

OrbitGl/GlCanvas.cpp
```diff
@@ -27,15 +27,20 @@ float GlCanvas::kZValueOverlayTextBackground = 0.45f;
...
+float GlCanvas::kZValueTimeBar = 0.83f;
+float GlCanvas::kZValueMargin = 0.85f;
+float GlCanvas::kZValueSliderBg = 0.87f;
+float GlCanvas::kZValueSlider = 0.89f;
 
 float GlCanvas::kZOffsetMovingTack = 0.1f;
 float GlCanvas::kZOffsetPinnedTrack = 0.2f;
 
+// Max Number of layers: 16 original, 4 for moving track, 4 for pinned Track
+unsigned GlCanvas::kMaxNumberRealZLayers = 16 + 4 + 4;
```
pierricgimmig, line 42: This is going to be a pain to maintain, ideally we could compute the value dynamically instead of hard-coding it. Let's get your PR in, but we should think about how we could improve this.

OrbitGl/TextRenderer.cpp
```diff
@@ -33,9 +32,10 @@ TextRenderer::~TextRenderer() {
   }
   m_FontsBySize.clear();
 
-  if (m_Buffer) {
-    vertex_buffer_delete(m_Buffer);
+  for (auto& [_, buffer] : buffers_by_layer_) {
```
pierricgimmig, line 35: nit. "_" -> "unused_layer"

OrbitGl/TextRenderer.cpp
```diff
@@ -332,6 +341,14 @@ int TextRenderer::GetStringWidth(const char* a_Text) const {
   return static_cast<int>(ceil(stringWidth));
 }
 
+std::vector<float> TextRenderer::GetLayers() const {
+  std::vector<float> layers;
+  for (auto& [layer, _] : buffers_by_layer_) {
```
pierricgimmig, line 346: "_"

OrbitGl/TextRenderer.cpp
```diff
@@ -349,9 +366,7 @@ float TextRenderer::ToScreenSpace(float a_Size) {
 void TextRenderer::Clear() {
   m_Pen.x = 0.f;
   m_Pen.y = 0.f;
-  if (m_Buffer) {
-    vertex_buffer_clear(m_Buffer);
+  for (auto& [_, buffer] : buffers_by_layer_) {
```
pierricgimmig, line 369: "_"

pierricgimmig (approved):
Hi Pablo, please see small nit comments, but otherwise looks good to me. Let's get this is in so that we have time address any unforeseen issue.

OrbitGl/BatcherTest.cpp
karupayun: it is BlockIterator, and we don't have implemented the "+" operator there, so we can't make it+=n

OrbitGl/GlCanvas.cpp
karupayun: Yes, it's not very nice. But we don't know if we are going to still have this constant in the future, or make a different approach like using enum. For now, I couldn't think something better.

OrbitGl/Batcher.h
karupayun: Renamed to ShapeBuffers

OrbitGl/CaptureWindow.cpp
karupayun: Yes, clearly!

OrbitGl/Batcher.h
pierricgimmig: "Primitive" is usually used in this context, I would prefer "Primitive" over "Shape"

karupayun: done!

PR #1409, opened 2020-10-29
Preparation for frame track serialization

ronaldfw:
Data for frame tracks is now kept in App and that is the source of
truth for added frame tracks. This is in preparation of supporting
serialization of frame tracks (and possibly live updates during a
capture, or at least maintaining added frame tracks in-between
captures).

This change also fixes a bug on master where removing a frame track 
was not possible anymore.

### DIFF
--- a/OrbitGl/App.cpp
+++ b/OrbitGl/App.cpp
@@ -1348,3 +1348,70 @@ void OrbitApp::DeselectTracepoint(const TracepointInfo& tracepoint) {
 [[nodiscard]] bool OrbitApp::IsTracepointSelected(const TracepointInfo& info) const {
   return data_manager_->IsTracepointSelected(info);
 }
+
+void OrbitApp::AddFrameTrack(const FunctionInfo& function) {
+  const CaptureData& capture_data = GetCaptureData();
+  const uint64_t function_address = capture_data.GetAbsoluteAddress(function);
+
+  if (frame_track_functions_.contains(function_address)) {
+    ERROR("Trying to add duplicate frame track for function: %s",
+          FunctionUtils::GetDisplayName(function));
+    DCHECK(false);
+    return;
+  }
+
+  std::vector<std::shared_ptr<TimerChain>> chains =
+      GCurrentTimeGraph->GetAllThreadTrackTimerChains();
+
+  std::vector<uint64_t> all_start_times;
+
+  for (const auto& chain : chains) {
+    if (!chain) continue;
+    for (const TimerBlock& block : *chain) {
+      for (uint64_t i = 0; i < block.size(); ++i) {
+        const TextBox& box = block[i];
+        if (box.GetTimerInfo().function_address() == function_address) {
+          all_start_times.push_back(box.GetTimerInfo().start());
+        }
+      }
+    }
+  }
+  std::sort(all_start_times.begin(), all_start_times.end());
+
+  for (size_t k = 0; k < all_start_times.size() - 1; ++k) {
+    TimerInfo frame_timer;
+
+    // TID is meaningless for this timer (start and end can be on two different threads).
+    constexpr const int32_t kUnusedThreadId = -1;
+    frame_timer.set_thread_id(kUnusedThreadId);
+    frame_timer.set_start(all_start_times[k]);
+    frame_timer.set_end(all_start_times[k + 1]);
+    // We use user_data_key to keep track of the frame number.
+    frame_timer.set_user_data_key(k);
+    frame_timer.set_type(TimerInfo::kFrame);
+
+    GCurrentTimeGraph->ProcessTimer(frame_timer, &function);
+  }
+  frame_track_functions_.insert(std::make_pair(function_address, function));
+}
+
+void OrbitApp::RemoveFrameTrack(const FunctionInfo& function) {
+  const CaptureData& capture_data = GetCaptureData();
+  auto function_address = capture_data.GetAbsoluteAddress(function);
+
+  if (!frame_track_functions_.contains(function_address)) {
+    ERROR("Trying to remove non-existent frame track for function: %s",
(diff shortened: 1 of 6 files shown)


OrbitGl/App.cpp
```diff
@@ -1348,3 +1348,55 @@ void OrbitApp::DeselectTracepoint(const TracepointInfo& tracepoint) {
...
+  std::vector<std::shared_ptr<TimerChain>> chains =
+      GCurrentTimeGraph->GetAllThreadTrackTimerChains();
+
+  std::vector<uint64_t> all_start_times;
+
+  for (auto& chain : chains) {
+    if (!chain) continue;
+    for (TimerChainIterator it = chain->begin(); it != chain->end(); ++it) {
+      TimerBlock& block = *it;
+      for (uint64_t i = 0; i < block.size(); i++) {
```
pierricgimmig, line 1364: mega nit: ++i (go/cstyle#Preincrement_and_Predecrement)

OrbitGl/App.cpp
```diff
@@ -1348,3 +1348,55 @@ void OrbitApp::DeselectTracepoint(const TracepointInfo& tracepoint) {
...
+    frame_timer.set_thread_id(kUnusedThreadId);
+    frame_timer.set_start(all_start_times[k]);
+    frame_timer.set_end(all_start_times[k + 1]);
+    // We use user_data_key to keep track of the frame number.
+    frame_timer.set_user_data_key(k);
+    frame_timer.set_type(TimerInfo::kFrame);
+
+    GCurrentTimeGraph->ProcessTimer(frame_timer, &function);
+  }
+  frame_track_functions_.insert(std::make_pair(function_address, function));
```
pierricgimmig, line 1388: I'm wondering if we should early out in the case where frame_track_functions_ already has an entry for function_address, does it make sense to call AddFrameTrack twice with the same function?

pierricgimmig (approved):
Hi Ronald, please see minor comments, but LGTM.

OrbitGl/App.cpp
ronaldfw: Done.

OrbitGl/App.cpp
ronaldfw: The logic that calls this function already checks if a frame track has been added for this function, but it might be a good idea to check that here as well and early out. If this is run twice, it just adds a duplicate timer to the existing frame track for this function. I don't think we dedup, so this would increase runtime for UpdatePrimitives.

I wasn't sure whether to silently return, add an error or even a CHECK. I've now added an ERROR log message and DCHECK based on the following reasoning:

- If in a shipped version this function is called twice due to a bug, it's not really a big deal. Crashing and making this feature unusable would be worse. Hence I don't want to have a CHECK here.
- It's still a logic mistake that we should fix and to draw attention to it, the combination of ERROR and DCHECK seems to be reasonable. 

I added a similar check to RemoveFrameTrack, which should only be called on functions that have a frametrack.

Does this sound reasonable?

OrbitGl/App.cpp
```diff
@@ -1348,3 +1348,71 @@ void OrbitApp::DeselectTracepoint(const TracepointInfo& tracepoint) {
...
+      GCurrentTimeGraph->GetAllThreadTrackTimerChains();
+
+  std::vector<uint64_t> all_start_times;
+
+  for (auto& chain : chains) {
+    if (!chain) continue;
+    for (TimerChainIterator it = chain->begin(); it != chain->end(); ++it) {
+      TimerBlock& block = *it;
+      for (uint64_t i = 0; i < block.size(); ++i) {
+        TextBox& box = block[i];
```
antonrohr, line 1373:
```suggestion
 for (const auto& chain : chains) {
    if (!chain) continue;
    for (const TimerBlock& block : *chain) {
      for (uint64_t i = 0; i < block.size(); ++i) {
        const TextBox& box = block[i];
```

OrbitGl/App.cpp
antonrohr: Sounds reasonable to me!

antonrohr (approved):
lgtm, only a few nitpicks. Feel free to discard

OrbitGl/App.cpp
```diff
@@ -1348,3 +1348,71 @@ void OrbitApp::DeselectTracepoint(const TracepointInfo& tracepoint) {
 [[nodiscard]] bool OrbitApp::IsTracepointSelected(const TracepointInfo& info) const {
   return data_manager_->IsTracepointSelected(info);
 }
+
+void OrbitApp::AddFrameTrack(const orbit_client_protos::FunctionInfo& function) {
```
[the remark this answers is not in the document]
ronaldfw: Good point, definitely a copy&paste error here.

OrbitGl/App.cpp
```diff
@@ -1348,3 +1348,71 @@ void OrbitApp::DeselectTracepoint(const TracepointInfo& tracepoint) {
...
+  const CaptureData& capture_data = GetCaptureData();
+  auto function_address = capture_data.GetAbsoluteAddress(function);
```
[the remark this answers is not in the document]
ronaldfw: Not really, probably copy & pasted from somewhere else at some point, not sure what the original reasoning for that was. I'm usually in favor of using explicit types unless that's really inconvenient.

OrbitGl/App.cpp
ronaldfw: Done.

OrbitGl/App.cpp
ronaldfw: Ok, thanks!

karupayun: We discussed a little in the daily about this decision. I'm not sure yet which is the best, but sounds reasonable!

ronaldfw:
This fixes b/172046495

PR #1410, opened 2020-10-30
Add more details to the message in the "Finalizing capture" dialog

dpallotti:
Bug: http://b/171350567

### DIFF
--- a/OrbitQt/orbitmainwindow.cpp
+++ b/OrbitQt/orbitmainwindow.cpp
@@ -81,8 +81,17 @@ OrbitMainWindow::OrbitMainWindow(QApplication* a_App,
     setWindowTitle({});
   });
 
+  constexpr const char* kFinalizingCaptureMessage =
+      "<div align=\"left\">"
+      "Please wait while the capture is being finalized..."
+      "<ul>"
+      "<li>Waiting for the remaining capture data</li>"
+      "<li>Processing callstacks</li>"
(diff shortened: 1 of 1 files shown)


dpallotti:
I wrote down something but feel free to suggest a different message.

OrbitQt/orbitmainwindow.cpp
```diff
@@ -81,8 +81,17 @@ OrbitMainWindow::OrbitMainWindow(QApplication* a_App,
     setWindowTitle({});
   });
 
+  constexpr const char* kFinalizingCaptureMessage =
+      "<div align=\"left\">"
+      "Please wait while the capture is being finalized..."
+      "<ul>"
+      "<li>Waiting for the remaining capture data</li>"
+      "<li>Processing callstacks</li>"
+      "<li>Cleaning up dynamic instrumentation trampolines</li>"
```
ronaldfw, line 90: Maybe remove the word "trampolines"? Seems too much of a detail that users don't have to know about necessarily.

ronaldfw (approved):
LGTM, minor suggestion.

OrbitQt/orbitmainwindow.cpp
dpallotti: I'm totally fine with it. So just "Cleaning up dynamic instrumentation"?

ronaldfw: SGTM

PR #1411, opened 2020-10-30
Apply pimpl to OrbitCaptureGgpClient and generate as lib

kisenshi:
Update _OrbitCaptureGgpClient_ so it can be used externally as a library. There are two main updates related to this:

1. Update _CMakeLists_ to make a distinction between the lib (_OrbitCaptureGgpClientLib_) and the executable for testing (_OrbitCaptureGgpClient_) and include it in Orbit so it generates a .lib that could be used externally. Include _OrbitBase_ instead of _OrbitCore_ to reduce the number of dependencies.

2. The interface needs to be as generic as possible. The code has been updated to implement the pimpl technique which allows this.


OrbitCaptureGgpClient/include/OrbitCaptureGgpClient/OrbitCaptureGgpClient.h
```diff
@@ -5,31 +5,24 @@
 #ifndef ORBIT_CAPTURE_GGP_CLIENT_ORBIT_CAPTURE_GGP_CLIENT_H_
 #define ORBIT_CAPTURE_GGP_CLIENT_ORBIT_CAPTURE_GGP_CLIENT_H_
 
-#include <grpcpp/ext/proto_server_reflection_plugin.h>
-#include <grpcpp/grpcpp.h>
-#include <grpcpp/health_check_service_interface.h>
-
-#include <iostream>
-#include <memory>
```
danielfenner, line 13: You need `#include <memory>` for the unique_ptr - that causes the broken presubmit ...

kisenshi: Fixed, thanks!

kisenshi:
Hi! As discussed offline, I have updated the CMakeLists so OrbitCaptureGgpClient is only compiled for Linux. I also have organised the position so the projects are ordered alphabetically.

OrbitCaptureGgpClient/CMakeLists.txt
```diff
@@ -2,27 +2,40 @@
...
-        main.cpp
+target_sources(OrbitCaptureGgpClientLib PRIVATE
         OrbitCaptureGgpClient.cpp)
 
-target_link_libraries(OrbitCaptureGgpClient PUBLIC 
+target_link_libraries(OrbitCaptureGgpClientLib PUBLIC 
         OrbitBase
         OrbitProtos)
 
+if(WIN32)
```
danielfenner, line 27: You can remove these defines since we are not compiling for windows anyway.

danielfenner (approved):
lgtm

OrbitCaptureGgpClient/CMakeLists.txt
kisenshi: You are right, I missed this. I have removed it. Thanks!

kisenshi:
Merging this now, thanks!

PR #1412, opened 2020-10-30
Cleanup Timer class

pierricgimmig:
Remove QueryMillis method and reimplement the ElapsedMillis (and other
ElapsedX methods) so that they return either "now - start" if the
timer hasn't been stopped, or "end - start" otherwise.

### DIFF
--- a/OrbitBase/include/OrbitBase/Profiling.h
+++ b/OrbitBase/include/OrbitBase/Profiling.h
@@ -9,7 +9,9 @@
 #include <Windows.h>
 using pid_t = uint32_t;
 #else
+#include <stdint.h>
 #include <sys/syscall.h>
+#include <time.h>
 #include <unistd.h>
 #endif
 
--- a/OrbitGl/TimeGraph.cpp
+++ b/OrbitGl/TimeGraph.cpp
@@ -36,7 +36,6 @@ TimeGraph* GCurrentTimeGraph = nullptr;
 
 TimeGraph::TimeGraph(uint32_t font_size)
     : font_size_(font_size), text_renderer_static_(font_size), batcher_(BatcherId::kTimeGraph) {
-  last_thread_reorder_.Start();
   scheduler_track_ = GetOrCreateSchedulerTrack();
 
   tracepoints_system_wide_track_ =
@@ -1019,7 +1018,7 @@ void TimeGraph::SortTracks() {
     Append(sorted_tracks_, external_pid_tracks);
     Append(sorted_tracks_, capture_pid_tracks);
 
-    last_thread_reorder_.Reset();
+    last_thread_reorder_.Restart();
 
     UpdateFilteredTrackList();
   }
--- a/OrbitGl/TimeGraph.cpp
+++ b/OrbitGl/TimeGraph.cpp
@@ -949,7 +949,7 @@ void TimeGraph::SortTracks() {
   }
 
   // Reorder threads once every second when capturing
-  if (!GOrbitApp->IsCapturing() || last_thread_reorder_.QueryMillis() > 1000.0) {
+  if (!GOrbitApp->IsCapturing() || last_thread_reorder_.ElapsedMillis() > 1000.0) {
     std::vector<int32_t> sorted_thread_ids = GetSortedThreadIds();
 
     ScopeLock lock(mutex_);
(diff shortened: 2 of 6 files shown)


OrbitGl/Timer.h
```diff
@@ -5,32 +5,28 @@
...
 
-  [[nodiscard]] double ElapsedMicros() const { return TicksToMicroseconds(start_, end_); }
-  [[nodiscard]] double ElapsedMillis() const { return ElapsedMicros() * 0.001; }
-  [[nodiscard]] double ElapsedSeconds() const { return ElapsedMicros() * 0.000001; }
-
-  [[nodiscard]] double QueryMillis() {
-    Stop();
-    return ElapsedMillis();
-  }
+  [[nodiscard]] double ElapsedNanos() const { return static_cast<double>(End() - start_); }
```
dpallotti, line 21: Wouldn't it be better if this returned a `uint64_t`, as we always treat nanoseconds as `uint64_t` in the codebase and consider them the smallest time resolution?

OrbitGl/Timer.h
```diff
@@ -5,32 +5,28 @@
...
+  explicit Timer() { Start(); }
 
+  void Start() { start_ = Now(); }
+  void Stop() { end_ = Now(); }
   void Reset() {
-    Stop();
-    Start();
+    start_ = Now();
+    end_ = 0;
   }
```
dpallotti, line 19: I have to say that I find the interface slightly confusing, maybe because I compare it to how one would use a stopwatch in real life. Basically the constructor, `Start`, `Reset` all do kind of the same thing, but not really.  In particular `Start` and `Reset` are a bit weird as `Start` doesn't change `end_`, which could be older than `start_`, and `Reset` does what I would expect from `Start`.

Maybe what I would personally do is only have `Restart` (which would do what `Reset` does now) and `Stop`, and the constructor calls `Restart`.

pierricgimmig: Yeah that's better.

OrbitGl/Timer.h
pierricgimmig: Good point, I changed it.

OrbitGl/Timer.h
```diff
@@ -5,32 +5,28 @@
...
-    Stop();
-    return ElapsedMillis();
-  }
+  [[nodiscard]] double ElapsedNanos() const { return static_cast<double>(End() - start_); }
+  [[nodiscard]] double ElapsedMicros() const { return ElapsedNanos() * 0.001; }
+  [[nodiscard]] double ElapsedMillis() const { return ElapsedNanos() * 0.000001; }
+  [[nodiscard]] double ElapsedSeconds() const { return ElapsedNanos() * 0.000000001; }
 
  private:
+  [[nodiscard]] uint64_t End() const { return end_ == 0 ? Now() : end_; }
```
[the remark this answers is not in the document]
pierricgimmig: Yes. Done.

OrbitGl/TimeGraph.cpp
```diff
@@ -949,7 +948,7 @@ void TimeGraph::SortTracks() {
   }
 
   // Reorder threads once every second when capturing
-  if (!GOrbitApp->IsCapturing() || last_thread_reorder_.QueryMillis() > 1000.0) {
+  if (!GOrbitApp->IsCapturing() || last_thread_reorder_.ElapsedMillis() > 1000.0) {
```
dimitry-, line 951: nit: maybe use nonos here?

dimitry- (approved):
otherwise lgtm

OrbitBase/include/OrbitBase/Profiling.h
```diff
@@ -9,7 +9,9 @@
 #include <Windows.h>
 using pid_t = uint32_t;
 #else
+#include <stdint.h>
 #include <sys/syscall.h>
+#include <time.h>
```
dpallotti, line 14: Is this intended?

dpallotti (approved):

OrbitBase/include/OrbitBase/Profiling.h
pierricgimmig: Yes, removing <string> and GlUtils.h from Timer.h uncovered those two missing includes.

PR #1414, opened 2020-10-31
Add "nodeploy" flag to disable deployment of OrbitService

pierricgimmig:
This could possibly replace the existing logic of not deploying
OrbitService via the "ORBIT_COLLECTOR_NO_DEPLOYMENT" environment
variable, which doesn't work in conjunction with the
"ORBIT_COLLECTOR_EXECUTABLE_PATH" variable. Having "nodeploy" as a
flag also allows to easily define a "nodeploy" IDE launch target.


IrinaShkviro (approved):
LGTM!

PR #1417, opened 2020-11-02
Fix concatenation in SCOPED_TIMED_LOG

dpallotti:
Bug: http://b/172091627

Test: Use the macro multiple times in the same scope:
builds, prints the expected messages


dimitry- (approved):

pierricgimmig (approved):
Thanks

PR #1418, opened 2020-11-03
Add more verbose log messages on known host fail.

antonrohr:
Should be self explanatory. Does not change any logic, just prints more verbose messages.


danielfenner (approved):
Thanks! lgtm

PR #1420, opened 2020-11-03
Refactor: Add const in a few places

antonrohr:
This adds const in a few places and uses a reference instead of a move in one place.

This is actually not meaningful on its own, but makes some things easier for an upcoming PR. 

Test: compile

### DIFF
--- a/OrbitQt/servicedeploymanager.h
+++ b/OrbitQt/servicedeploymanager.h
@@ -30,8 +30,8 @@ class ServiceDeployManager : public QObject {
     uint16_t grpc_port;
   };
 
-  explicit ServiceDeployManager(DeploymentConfiguration deployment_configuration,
-                                OrbitSsh::Context* context, OrbitSsh::Credentials creds,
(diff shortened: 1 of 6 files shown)


OrbitQt/servicedeploymanager.h
```diff
@@ -49,8 +49,8 @@ class ServiceDeployManager : public QObject {
  private:
   EventLoop loop_;
 
-  DeploymentConfiguration deployment_configuration_;
```
dimitry-, line 52: since this is storing a reference we should probably make it clearer that this cannot be temporary by changing c-tor and this field to const *

dimitry- (approved):

OrbitQt/servicedeploymanager.h
florian-kuebler: Agree. So it is not clear, that the lifetime of the configuration must be at least the same as the lifetime of the manager.

florian-kuebler (approved):
LGTM. Thanks!

PR #1421, opened 2020-11-04
Add automated test scripts for bottom-up and top-down view

dpallotti:
#### Add pywinauto automated test for bottom-up view
Verifies that:
- the bottom-up view contains at least ten rows
- the first element is "ioctl"
- the first child of the first element starts with "drm"

Test: Ran locally.
#### Add pywinauto automated test for top-down view
Verifies that:
- the top-down view contains at least 3 rows
- the first item is "hello_* (all threads)"
- the second item is "GgpSwapchain [*]"
- the children of the first item are "*clone" and "_start"

Test: Ran locally.


danielfenner (approved):
Thanks for looking into this - lgtm!

PR #1424, opened 2020-11-04
Do not log Deadline expired errors

dimitry-:
GetProcessList no longer pollutes log with Deadline expired error messages.
Also add a missing nodiscard

Test: build, run & capture

### DIFF
--- a/OrbitClientServices/ProcessClient.cpp
+++ b/OrbitClientServices/ProcessClient.cpp
@@ -48,8 +48,10 @@ ErrorMessageOr<std::vector<orbit_grpc_protos::ProcessInfo>> ProcessClient::GetPr
 
   grpc::Status status = process_service_->GetProcessList(context.get(), request, &response);
   if (!status.ok()) {
-    ERROR("gRPC call to GetProcessList failed: %s (error_code=%d)", status.error_message(),
-          status.error_code());
+    if (status.error_code() != grpc::StatusCode::DEADLINE_EXCEEDED) {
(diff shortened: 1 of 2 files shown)


antonrohr (approved):

OrbitClientServices/ProcessClient.cpp
```diff
@@ -48,8 +48,10 @@ ErrorMessageOr<std::vector<orbit_grpc_protos::ProcessInfo>> ProcessClient::GetPr
 
   grpc::Status status = process_service_->GetProcessList(context.get(), request, &response);
   if (!status.ok()) {
-    ERROR("gRPC call to GetProcessList failed: %s (error_code=%d)", status.error_message(),
-          status.error_code());
+    if (status.error_code() != 4) {
```
dpallotti, line 51: Doesn't that "4" have an associated constant that we could use instead? Otherwise this is a bit unclear.

dpallotti: Should be `grpc::DEADLINE_EXCEEDED`.

dpallotti: Or `grpc::StatusCode::DEADLINE_EXCEEDED` for better clarity, but it's not an `enum class`.

dimitry-: done.

dimitry-:
decided to keep it as is (hopefully this error goes away once we move tunneling to a separate thread)

PR #1426, opened 2020-11-05
Add "collector" and "collector_root_password" flags

pierricgimmig:
The passed in values have precedence over the environment variables.
This makes it possible to easily launch different configurations of the
collector (OrbitService) from scripts or IDE's. Also, clearly log that when
are using a debug version of OrbitService.

### DIFF
--- a/OrbitQt/main.cpp
+++ b/OrbitQt/main.cpp
@@ -51,6 +51,10 @@ ABSL_FLAG(bool, devmode, false, "Enable developer mode in the client's UI");
 
 ABSL_FLAG(bool, nodeploy, false, "Disable automatic deployment of OrbitService");
 
+ABSL_FLAG(std::string, collector, "", "Full path of collector to be deployed");
+
+ABSL_FLAG(std::string, collector_root_password, "", "Collector's machine root password");
+
 ABSL_FLAG(uint16_t, grpc_port, 44765,
           "The service's GRPC server port (use default value if unsure)");
 ABSL_FLAG(bool, local, false, "Connects to local instance of OrbitService");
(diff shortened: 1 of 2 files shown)


OrbitQt/main.cpp
```diff
@@ -239,9 +243,19 @@ static std::optional<OrbitQt::DeploymentConfiguration> FigureOutDeploymentConfig
...
+  std::string collector_override = absl::GetFlag(FLAGS_collector);
+  std::string collector_password = absl::GetFlag(FLAGS_collector_root_password);
+  std::string password = !collector_password.empty() ? collector_password
+                         : env.contains(kEnvRootPassword)
+                             ? env.value(kEnvRootPassword).toStdString()
+                             : "";
+
+  if (!collector_override.empty() && !password.empty()) {
+    return OrbitQt::BareExecutableAndRootPasswordDeployment{collector_override, password};
+  } else if (env.contains(kEnvExecutablePath) && !password.empty()) {
```
antonrohr, line 256: Why not do the same for both location and password?
```suggestion
// Program argument have higher precedence than environment variables.
  std::string collector_location =
      !absl::GetFlag(FLAGS_collector).empty()
          ? absl::GetFlag(FLAGS_collector)
          : env.contains(kEnvExecutablePath) ? env.value(kEnvExecutablePath).toStdString() : "";
  std::string collector_password =
      !absl::GetFlag(FLAGS_collector_root_password).empty()
          ? absl::GetFlag(FLAGS_collector_root_password)
          : env.contains(kEnvRootPassword) ? env.value(kEnvRootPassword).toStdString() : "";

  if (!collector_location.empty() && !collector_password.empty()) {
    return OrbitQt::BareExecutableAndRootPasswordDeployment{collector_location, collector_password};
```

antonrohr: Also.. should we support empty passwords? Probably not

antonrohr (approved):
My last comment was of rather nitpicky nature. So I trust you to make good choices before you submit :)

OrbitQt/main.cpp
pierricgimmig: Good points @antonrohr , I will clean up the code and add support for empty password, thanks.

beckerhe (approved):
LGTM! Thanks Pierric. I also wanted to add something similar for a long time!

PR #1427, opened 2020-11-05
Crash because too many z-layers

karupayun:
This change includes:
- Misuse of unique (didn't resize after)
- Add 4 to the maximum number of layers (epsilon in GlSlider)
- Change the CHECK for ERROR to not have a unnecesary crash

Solve b/172455056.


florian-kuebler (approved):
LGTM. Thanks!

PR #1428, opened 2020-11-05
Make automated test more flexible to the content of the top-down view

dpallotti:
The expected most sampled thread from `hello_ggp_standalone` was GgpSwapchain,
but this is not always the case. The script now accepts any most sampled
thread of the form "hello_\*" or "Ggp\*".

Test: Ran locally a few times.


danielfenner (approved):
lgtm!

PR #1431, opened 2020-11-05
Actually abort a running capture on exit

dpallotti:
#### Use ClientContext::TryCancel to actually abort the capture when exiting
So far, "aborting" the capture on exit only meant stopping it, which could still
take a long time. Now this is immediate.

This even seems to allow to close Orbit in the situation of http://b/171923841.

Test: In addition to several local and normal captures on Trata, also capture
Trata as per http://b/171923841#comment17 until the issue arises, and try to
(now successfully) close Orbit.
#### force_stop_: reset to false, rename to writes_done_failed_

### DIFF
--- a/OrbitCaptureClient/CaptureClient.cpp
+++ b/OrbitCaptureClient/CaptureClient.cpp
@@ -69,11 +69,13 @@ void CaptureClient::Capture(ProcessData&& process,
                             const OrbitClientData::ModuleManager& module_manager,
                             absl::flat_hash_map<uint64_t, FunctionInfo> selected_functions,
                             TracepointInfoSet selected_tracepoints, bool enable_introspection) {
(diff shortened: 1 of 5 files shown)


OrbitCaptureClient/CaptureClient.cpp
```diff
@@ -179,6 +185,18 @@ bool CaptureClient::StopCapture() {
   return true;
 }
 
+bool CaptureClient::TryAbortCapture() {
+  if (client_context_ == nullptr) {
```
dpallotti, line 189: Mmmh, this might need a mutex or at least to be atomic.

dpallotti: Changed this part.

dimitry- (approved):

pierricgimmig (approved):

PR #1434, opened 2020-11-06
Increase ssh tunnel's kChunkSize for faster reads

pierricgimmig:
This greatly reduces the hangs we are sometimes seeing on capture
stop. The real cause is the fact that the tunnel is updated on the ui
thread, which shouldn't be the case as it can being drastically delayed 
by a long render frame.

See b/172686811

### DIFF
--- a/OrbitSshQt/Tunnel.cpp
+++ b/OrbitSshQt/Tunnel.cpp
@@ -157,7 +157,7 @@ outcome::result<void> Tunnel::shutdown() {
 
 outcome::result<void> Tunnel::readFromChannel() {
   while (true) {
-    const size_t kChunkSize = 8192;
+    const size_t kChunkSize = 1024 * 1024;
     const auto result = channel_->ReadStdOut(kChunkSize);
 
     if (!result && !OrbitSsh::ShouldITryAgain(result)) {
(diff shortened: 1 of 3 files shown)


OrbitSshQt/Tunnel.cpp
```diff
@@ -180,6 +180,7 @@ outcome::result<void> Tunnel::readFromChannel() {
     if (bytes_written == -1) {
       SetError(Error::kLocalSocketClosed);
     } else {
+      // TODO: avoid extra copy b/172686811.
```
dpallotti, line 183:
```suggestion
      // TODO(b/172686811): avoid extra copy.
```

dpallotti (approved):

dimitry- (approved):

ronaldfw:
Should this actually go into release/1.55 or only master at this point?

dpallotti:
> Should this actually go into release/1.55 or only master at this point?

Ah right, this is actually targeting the release branch...

dimitry-:
Right - this should go to master.

ronaldfw:
Per discussion with @pierricgimmig, this should now actually go into release/1.55.

PR #1435, opened 2020-11-09
Clear frame track functions in App when clearing the capture

ronaldfw:
We have to clear the stored frame track functions in App when the
capture is cleared to make sure that the UI behaves consistently.

Fixes b/172452901.


florian-kuebler (approved):
LGTM, thanks!

PR #1437, opened 2020-11-09
Make automated test even more flexible to content of top-down view

dpallotti:
Allow `*clone` and `_start` to be the outermost frames across all threads,
but now in either order.

Bug: http://b/172684376


danielfenner (approved):
lgtm!

PR #1439, opened 2020-11-09
Increase ssh tunnel's kChunkSize for faster reads

pierricgimmig:
This greatly reduces the hangs we are sometimes seeing on capture
stop. The real cause is the fact that the tunnel is updated on the ui
thread, which shouldn't be the case as it can being drastically delayed
by a long render frame.

See b/172686811


pierricgimmig:
Reviewed in #1434, as discussed, pushing to release branch.

PR #1440, opened 2020-11-10
Don't make unnecessary swaps when moving tracks

karupayun:
When we move a track after making a filter, it is moved to the last
possible position (just before the next track in the filtered array).

This is what you expect if you are moving tracks upwards, but if you
move tracks downwards or take a track but don't change its position at
all, this results in  a strange behavior where the moved_track is moved
more to the bottom than what it is expected.

Specially noticed when you move a track to the end. After erasing the
filter, the moved_track will be at the end of the complete list.

Related to b/170370861.

### DIFF
--- a/OrbitGl/TimeGraph.cpp
+++ b/OrbitGl/TimeGraph.cpp
@@ -1052,36 +1052,59 @@ void TimeGraph::UpdateMovingTrackSorting() {
   // Note: We do an O(n) search for the correct position in the sorted_tracks_ array which
   // could be optimized, but this is not worth the effort for the limited number of tracks.
 
-  std::shared_ptr<Track> moving_track = nullptr;
-  for (auto track_it = sorted_filtered_tracks_.begin(); track_it != sorted_filtered_tracks_.end();
-       ++track_it) {
-    if ((*track_it)->IsMoving()) {
-      moving_track = *track_it;
-      sorted_filtered_tracks_.erase(track_it);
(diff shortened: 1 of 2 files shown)


OrbitGl/TimeGraph.cpp
```diff
@@ -1053,33 +1053,52 @@ void TimeGraph::UpdateMovingTrackSorting() {
...
     }
 
-    if (!inserted) {
-      sorted_tracks_.push_back(moving_track);
+    if (moving_track_current_position == -1) {
       sorted_filtered_tracks_.push_back(moving_track);
+      moving_track_current_position = static_cast<int>(sorted_filtered_tracks_.size()) - 1;
+    }
+
+    // Now we have to change the position of the moving_track in the non-filtered array
```
reichlfl, line 1083: Nit: The method is getting quite long could be split, maybe into finding the moving track and inserting the moving track at the correct position.

reichlfl (approved):
Thanks, great change! LGTM, up to you if you want to address my nitpick comment.

OrbitGl/TimeGraph.cpp
karupayun: better now?

reichlfl: Yes, thanks. But another nitpick for the nitpick: `FindMovingTrack` should probably be `FindMovingTrackIndex`, and be protected or private - but visibility is a mess in the TimeGraph class already, so this is at least consistent ;-)

karupayun: haha, done!

PR #1441, opened 2020-11-10
Add class to hold user defined capture data and use it for frametracks

ronaldfw:
We add a class UserDefinedCaptureData to hold any user defined capture
data such as frame tracks, iterators, annotations, comments. It is
currently only used for frame tracks.

Note that this is another preparation step for serialization of frametracks.

Bug: b/171026228

Tested: Unit test; manual test of frame tracks.

### DIFF
--- a/OrbitCore/UserDefinedCaptureData.cpp
+++ b/OrbitCore/UserDefinedCaptureData.cpp
@@ -1,0 +1,22 @@
+// Copyright (c) 2020 The Orbit Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#include "UserDefinedCaptureData.h"
+
+#include "OrbitClientData/FunctionUtils.h"
+
+void UserDefinedCaptureData::InsertFrameTrack(const orbit_client_protos::FunctionInfo& function) {
+  // We do not allow insertion of frame tracks twice.
+  CHECK(frame_track_functions_.insert(function).second);
+}
+
+void UserDefinedCaptureData::EraseFrameTrack(const orbit_client_protos::FunctionInfo& function) {
+  // We do not allow erasing a frame track that does not exist.
+  CHECK(frame_track_functions_.erase(function));
+}
+
+[[nodiscard]] bool UserDefinedCaptureData::ContainsFrameTrack(
+    const orbit_client_protos::FunctionInfo& function) const {
+  return frame_track_functions_.contains(function);
+}
--- a/OrbitCore/UserDefinedCaptureData.h
+++ b/OrbitCore/UserDefinedCaptureData.h
@@ -1,0 +1,25 @@
+// Copyright (c) 2020 The Orbit Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#ifndef ORBIT_CORE_USER_DEFINED_CAPTURE_DATA_H_
+#define ORBIT_CORE_USER_DEFINED_CAPTURE_DATA_H_
+
+#include "OrbitClientData/FunctionInfoSet.h"
+
+// UserDefinedCaptureData holds any capture related data that was added by the user. Examples
+// for this include frame tracks, iterators, timeline annotations or comments.
+// Note that this class is not thread-safe.
+class UserDefinedCaptureData {
+ public:
+  [[nodiscard]] FunctionInfoSet frame_track_functions() const { return frame_track_functions_; }
+  void InsertFrameTrack(const orbit_client_protos::FunctionInfo& function);
+  void EraseFrameTrack(const orbit_client_protos::FunctionInfo& function);
+  [[nodiscard]] bool ContainsFrameTrack(const orbit_client_protos::FunctionInfo& function) const;
+  void Clear() { frame_track_functions_.clear(); }
+
+ private:
+  FunctionInfoSet frame_track_functions_;
+};
+
+#endif  // ORBIT_CORE_USER_DEFINED_CAPTURE_DATA_H_
--- a/OrbitCore/UserDefinedCaptureDataTest.cpp
+++ b/OrbitCore/UserDefinedCaptureDataTest.cpp
@@ -1,0 +1,89 @@
+// Copyright (c) 2020 The Orbit Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#include <gtest/gtest.h>
+
+#include <string>
+
+#include "UserDefinedCaptureData.h"
+#include "capture_data.pb.h"
+
+using orbit_client_protos::FunctionInfo;
+
+FunctionInfo CreateFunctionInfo(const std::string& function_name, uint64_t function_address) {
+  FunctionInfo info;
+  info.set_name(function_name);
+  info.set_pretty_name(function_name);
+  info.set_loaded_module_path("/path/to/module");
+  info.set_address(function_address);
+  info.set_size(16);
+  info.set_file("file.cpp");
+  info.set_line(13);
+  return info;
+}
+
+TEST(UserDefinedCaptureData, InsertFrameTrack) {
+  UserDefinedCaptureData data;
+  FunctionInfo info = CreateFunctionInfo("fun0_name", 0);
+  data.InsertFrameTrack(info);
+  EXPECT_TRUE(data.ContainsFrameTrack(info));
+}
+
+TEST(UserDefinedCaptureData, InsertFrameTrackDuplicateFunctions) {
+  UserDefinedCaptureData data;
+  FunctionInfo info = CreateFunctionInfo("fun0_name", 0);
+  data.InsertFrameTrack(info);
+  EXPECT_DEATH(data.InsertFrameTrack(info), "");
+}
+
+TEST(UserDefinedCaptureData, InsertFrameTrackDifferentFunctions) {
+  UserDefinedCaptureData data;
+  FunctionInfo info0 = CreateFunctionInfo("fun0_name", 0);
+  FunctionInfo info1 = CreateFunctionInfo("fun1_name", 1);
+  data.InsertFrameTrack(info0);
+  data.InsertFrameTrack(info1);
+  FunctionInfoSet set = data.frame_track_functions();
+  EXPECT_EQ(set.size(), 2);
+}
+
+TEST(UserDefinedCaptureData, EraseNonExistentFrameTrack) {
+  UserDefinedCaptureData data;
+  FunctionInfo info = CreateFunctionInfo("fun0_name", 0);
+  EXPECT_DEATH(data.EraseFrameTrack(info), "");
+}
+
+TEST(UserDefinedCaptureData, EraseFrameTrack) {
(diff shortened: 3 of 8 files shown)


OrbitCore/UserDefinedCaptureData.h
```diff
@@ -0,0 +1,22 @@
+// Copyright (c) 2020 The Orbit Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#ifndef ORBIT_CORE_USER_DEFINED_CAPTURE_DATA_H_
+#define ORBIT_CORE_USER_DEFINED_CAPTURE_DATA_H_
+
+#include "OrbitClientData/FunctionInfoSet.h"
+
+class UserDefinedCaptureData {
```
dimitry-, line 10: Add a comment

OrbitCore/UserDefinedCaptureData.cpp
```diff
@@ -0,0 +1,30 @@
+// Copyright (c) 2020 The Orbit Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#include "UserDefinedCaptureData.h"
+
+#include "OrbitClientData/FunctionUtils.h"
+
+void UserDefinedCaptureData::InsertFrameTrack(const orbit_client_protos::FunctionInfo& function) {
```
dimitry-, line 9: Maybe make it bool and document when it returns false, or ErrorMessageOr because nobody reads logs.

Same for Erase

OrbitCore/UserDefinedCaptureData.h
ronaldfw: Done.

OrbitCore/UserDefinedCaptureData.cpp
ronaldfw: I'm not sure how to best handle this case. If I return success/error state here, what would the caller do? I can only think of printing a log message, which just moves the problem to a different place. I don't want this to bubble up to become a user visible error.

In both cases, it's not really a problem to call these functions multiple times with the same function, but it hints at a problem in the logic that calls these functions. 

I'm considering just removing the error check and allowing multiple calls to insert/erase with the same function (where second and subsequent calls will have no effect). 

WDYT?

dimitry-: yeah - this is why I was thinking about returning bool here - caller may choose and in most cases will ignore it but there is still signal if needed. 

Returns true if the value was inserted and false otherwise (in the case when Function Set already contains this function)

dimitry-: (in which case you do not need a check - you can return the second result of the std::set::insert)

florian-kuebler: Can this actually happen, or is the context-menu only shown, if the function is not yet added?
If this is the case, I would actually argue that it should be a CHECK here, as we may have wrong assumptions.
If not, we might think of making it that way.

OrbitCore/UserDefinedCaptureData.h
florian-kuebler: So far it seems that you did not pushed the changes yet.
I have a rather conceptual question on this class. Should (if so how) should it be made thread-safe?
There should be probably a statement about thread-safty in the comment here, as this is a topic of special interest in the area of capture data, data manager, and app.

OrbitCore/UserDefinedCaptureData.cpp
```diff
@@ -0,0 +1,30 @@
...
+    ERROR("Trying to remove non-existent frame track for function: %s",
+          FunctionUtils::GetDisplayName(function));
+    return;
+  }
+  frame_track_functions_.erase(function);
+}
+
+[[nodiscard]] bool UserDefinedCaptureData::ContainsFrameTrack(
+    const orbit_client_protos::FunctionInfo& function) const {
+  return frame_track_functions_.count(function) > 0;
```
florian-kuebler, line 29: FunctionSet should be an absl set. This offers a contains method, which I would use here.

OrbitCore/UserDefinedCaptureDataTest.cpp
```diff
@@ -0,0 +1,101 @@
...
+}
+
+TEST(UserDefinedCaptureData, InsertFrameTrack) {
+  UserDefinedCaptureData data;
+  FunctionInfo info = CreateFunctionInfo("fun0_name", 0);
+  data.InsertFrameTrack(info);
+  EXPECT_TRUE(data.ContainsFrameTrack(info));
+}
+
+TEST(UserDefinedCaptureData, InsertFrameTrackDuplicateFunctions) {
```
florian-kuebler, line 33: Depending on the decision on the above comment, this testcase might be irrelevant.

OrbitCore/UserDefinedCaptureDataTest.cpp
```diff
@@ -0,0 +1,101 @@
...
+
+TEST(UserDefinedCaptureData, EraseFrameTrack) {
+  UserDefinedCaptureData data;
+  FunctionInfo info = CreateFunctionInfo("fun0_name", 0);
+  data.InsertFrameTrack(info);
+  data.EraseFrameTrack(info);
+  EXPECT_FALSE(data.ContainsFrameTrack(info));
+}
+
+TEST(UserDefinedCaptureData, EraseFrameTrackDuplicateFunctions) {
```
florian-kuebler, line 60: Again, might be irrelevant if we decide to check for not inserting duplicates.

OrbitCore/UserDefinedCaptureDataTest.cpp
```diff
@@ -0,0 +1,101 @@
...
+  EXPECT_TRUE(data.ContainsFrameTrack(info1));
+}
+
+TEST(UserDefinedCaptureData, ContainsFrameTrackEmpty) {
+  UserDefinedCaptureData data;
+  FunctionInfo info = CreateFunctionInfo("fun1_name", 0);
+  EXPECT_FALSE(data.ContainsFrameTrack(info));
+}
+
+TEST(UserDefinedCaptureData, ContainsFrameTrackSingleFunction) {
```
florian-kuebler, line 86: Isn't that the exact same test case as `InsertFrameTrack`. We might want to omit it (also following the recent talk about testing in our video hour).

florian-kuebler (commented):
Besides the minor comments LGTM. However, I would really argue for having the CHECKs (or maybe just DCHECKs inside the insert/erase functions.

OrbitCore/UserDefinedCaptureData.cpp
ronaldfw: I had a DCHECK here first and would prefer something like that. The logic around prevents this. However, it's strictly not a problem here, it's just a problem in the logic in the calling code. So maybe the right thing is to do a mix of what @dimitry- and you are suggesting. Return a boolean indicating whether the insertion / erasure was successful and CHECK by the caller (where the logic error happens).

OrbitCore/UserDefinedCaptureDataTest.cpp
ronaldfw: > also following the recent talk about testing in our video hour

I wasn't there to watch it :) What would be the reasoning here to omit this?

OrbitCore/UserDefinedCaptureData.cpp
ronaldfw: Ok, I now added CHECK directly here, as it felt artificial to return the result and then CHECK further out. Let me know if I should change this.

OrbitCore/UserDefinedCaptureData.h
ronaldfw: Yes, sorry, I haven't pushed the change yet. Good point on thread safety, will at least add a comment. Right now, the data here can only be updated by the main thread.

OrbitCore/UserDefinedCaptureDataTest.cpp
florian-kuebler: Right, the idea was to prefer behavioural test cases over method/function base test cases.
The given example was a class simulating a glas, that can be filled and can you can drink from it.
So it has 4 methods: Constructor, Fill, Drink, IsFull.
Function based tests would be:
TestConstructor -> Create a Glass, we assume it is empty, but don't want to check its members (as this would not be a blackbox), so we use the "yet-untested" "IsFull" method.
TestFill -> Create a Glass, fill it, and again use the "IsFull" method, that does not really correspond to the test case.
TestDrink -> ...
TestIsFilled-> Now basically does not test anything new.

Instead it suggests behavioural based test cases:
TestAGlassCanBeCreated
TestAnEmptyGlassCanBeFilled
TestYouCanDrinkFromAFilledGlass

florian-kuebler: FYI: https://youtu.be/FjwayiHNI1w?t=356

florian-kuebler: But also note, this is super nitpicking. Feel free to ignore.

OrbitCore/UserDefinedCaptureData.cpp
florian-kuebler: Both would be fine for me. I'll leave it to @dimitry-.

OrbitCore/UserDefinedCaptureData.h
ronaldfw: Done.

OrbitCore/UserDefinedCaptureData.cpp
ronaldfw: PTAL

OrbitCore/UserDefinedCaptureData.cpp
ronaldfw: Done.

OrbitCore/UserDefinedCaptureDataTest.cpp
ronaldfw: Done. Actually added an EXPECT_DEATH check here.

OrbitCore/UserDefinedCaptureDataTest.cpp
ronaldfw: Done. Actually added an EXPECT_DEATH check here.

OrbitCore/UserDefinedCaptureDataTest.cpp
ronaldfw: Ok, thanks for the explanation. I went through the test cases and didn't really find one where that would make a big difference here, except for removing this test case here, which is obsolete and covered by the above behavior test (insertion causes contains to be true).

ronaldfw:
Note: I'll leave the commits here for reviewing, but will "fixup" the commits into the parent once review is done.

florian-kuebler (approved):
LGTM. Thanks.
One last remark about the original Commit Message. Please add "Bug: " and "Test: " to it.

dimitry- (approved):
lgtm!

ronaldfw:
> LGTM. Thanks.
> One last remark about the original Commit Message. Please add "Bug: " and "Test: " to it.

Good point, added.

PR #1442, opened 2020-11-10
Move SSH tunnel processing to a background thread

beckerhe:
I expected that there is not much more to do than `service_deploy_manager->moveToThread(ssh_background_thread);`, but some cleanup needed to be done as well.

Furthermore some synchronization needed to happen to make the public interface of ServiceDeployManager thread-safe.

### DIFF
--- a/OrbitQt/servicedeploymanager.cpp
+++ b/OrbitQt/servicedeploymanager.cpp
@@ -83,7 +83,6 @@ ServiceDeployManager::ServiceDeployManager(const DeploymentConfiguration* deploy
                                            const ServiceDeployManager::GrpcPort& grpc_port,
                                            QObject* parent)
     : QObject(parent),
-      loop_(this),
       deployment_configuration_(deployment_configuration),
       context_(context),
       credentials_(std::move(credentials)),
@@ -94,6 +93,11 @@ ServiceDeployManager::ServiceDeployManager(const DeploymentConfiguration* deploy
 
   background_thread_.start();
   moveToThread(&background_thread_);
+
+  // Event loop needs to be initialized in background thread to work properly on Linux.
+  QMetaObject::invokeMethod(
+      this, [this]() { loop_.emplace(); }, Qt::BlockingQueuedConnection);
+  CHECK(loop_.has_value());
 }
 
 ServiceDeployManager::~ServiceDeployManager() noexcept {
@@ -107,7 +111,7 @@ ServiceDeployManager::~ServiceDeployManager() noexcept {
 
 void ServiceDeployManager::Cancel() {
   DeferToBackgroundThreadAndWait(
-      this, [this]() { loop_.error(make_error_code(Error::kUserCanceledServiceDeployment)); });
+      this, [this]() { loop_->error(make_error_code(Error::kUserCanceledServiceDeployment)); });
 }
 
 outcome::result<bool> ServiceDeployManager::CheckIfInstalled() {
@@ -127,14 +131,15 @@ outcome::result<bool> ServiceDeployManager::CheckIfInstalled() {
 
   OrbitSshQt::Task check_if_installed_task{&session_.value(), command};
 
-  QObject::connect(&check_if_installed_task, &OrbitSshQt::Task::finished, &loop_, &EventLoop::exit);
+  QObject::connect(&check_if_installed_task, &OrbitSshQt::Task::finished, &loop_.value(),
+                   &EventLoop::exit);
 
   auto error_handler =
       ConnectErrorHandler(&check_if_installed_task, &OrbitSshQt::Task::errorOccurred);
 
   check_if_installed_task.Start();
 
-  OUTCOME_TRY(result, loop_.exec());
+  OUTCOME_TRY(result, loop_->exec());
   if (result == 0) {
     // Already installed
     emit statusMessage("The correct version of OrbitService is already installed.");
@@ -158,7 +163,7 @@ outcome::result<uint16_t> ServiceDeployManager::StartTunnel(
 
   tunnel->value().Start();
 
-  OUTCOME_TRY(MapError(loop_.exec(), Error::kCouldNotStartTunnel));
+  OUTCOME_TRY(MapError(loop_->exec(), Error::kCouldNotStartTunnel));
 
   QObject::connect(&tunnel->value(), &OrbitSshQt::Tunnel::errorOccurred, this,
                    &ServiceDeployManager::handleSocketError);
@@ -197,7 +202,7 @@ outcome::result<void> ServiceDeployManager::CopyFileToRemote(
   LOG("About to start copying from %s to %s...", source, dest);
   operation.CopyFileToRemote(source, dest, dest_mode);
 
-  OUTCOME_TRY(loop_.exec());
+  OUTCOME_TRY(loop_->exec());
   return outcome::success();
 }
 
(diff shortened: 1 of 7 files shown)


OrbitQt/servicedeploymanager.cpp
```diff
@@ -199,6 +222,15 @@ outcome::result<void> ServiceDeployManager::CopyOrbitServicePackage() {
 
 ErrorMessageOr<void> ServiceDeployManager::CopyFileToLocal(std::string_view source,
                                                            std::string_view destination) {
+  if (QThread::currentThread() != thread()) {
+    ErrorMessageOr<void> result = outcome::success();
+    DeferToBackgroundThreadAndWait(
+        this, [&, source = std::string{source}, destination = std::string{destination}]() {
+          result = CopyFileToLocal(source, destination);
```
dimitry-, line 229: Maybe move this to another function and have a CHECK(QThread::currentThread() != thread()) there?

In which case you won't need this if here.

dimitry-: Same for Cancel

OrbitQt/servicedeploymanager.cpp
```diff
@@ -384,6 +416,12 @@ void ServiceDeployManager::StartWatchdog() {
 }
 
 outcome::result<ServiceDeployManager::GrpcPort> ServiceDeployManager::Exec() {
+  if (QThread::currentThread() != thread()) {
+    std::optional<outcome::result<GrpcPort>> result;
+    DeferToBackgroundThreadAndWait(this, [&]() { result = Exec(); });
+    return result ? result.value() : outcome::failure(std::errc::timed_out);
+  }
```
dimitry-, line 423: and here

OrbitQt/servicedeploymanager.cpp
```diff
@@ -478,6 +516,11 @@ void ServiceDeployManager::ShutdownSession() {
 }
 
 void ServiceDeployManager::Shutdown() {
+  if (QThread::currentThread() != thread()) {
+    DeferToBackgroundThreadAndWait(this, [this]() { Shutdown(); });
+    return;
+  }
```
dimitry-, line 522: and here

OrbitQt/servicedeploymanager.cpp
beckerhe: I restructured this entirely and moved the thread into ServiceDeployManager which means it now always uses a background thread.

OrbitQt/servicedeploymanager.cpp
```diff
@@ -384,6 +421,12 @@ void ServiceDeployManager::StartWatchdog() {
 }
 
 outcome::result<ServiceDeployManager::GrpcPort> ServiceDeployManager::Exec() {
+  outcome::result<GrpcPort> result = outcome::success(GrpcPort{0});
+  DeferToBackgroundThreadAndWait(this, [&]() { result = ExecImpl(); });
+  return result;
+}
+
+outcome::result<ServiceDeployManager::GrpcPort> ServiceDeployManager::ExecImpl() {
```
dimitry-, line 429: Do you want to add CHECKS here to make sure this is always called on background thread? just in case...

dimitry- (approved):
otherwise lgtm, thanks!

OrbitQt/servicedeploymanager.cpp
beckerhe: Yes. That's already done but not pushed yet. There is also another problem on Linux which needs to be fixed before merging that! Thanks.

beckerhe: Linux is now also working. QEventLoop didn't like to be created in a different thread. By deferring its initialization to the background thread the problem is gone. Might be a bug or an insufficiently communicated precondition of the class.

beckerhe:
It works now under Linux. (The last commit describes in detail what went wrong). Adding Pierric for review since he might wanna test this before merging anyway.

OrbitQt/servicedeploymanager.cpp
```diff
@@ -94,6 +93,9 @@ ServiceDeployManager::ServiceDeployManager(const DeploymentConfiguration* deploy
 
   background_thread_.start();
   moveToThread(&background_thread_);
+  QMetaObject::invokeMethod(
```
dimitry-, line 96: maybe duplicate the explanation here as well? "Event loop needs to be initialized in background thread to work properly on Linux"?

beckerhe: Yep, good idea. On it.

beckerhe: done.

pierricgimmig:
@beckerhe , I tried your change and it feels really robust now on Linux, I couldn't reproduce the hang. It seems like data is streaming more fluidly too. Thanks for the fix! I'll look at the code shortly.

OrbitQt/servicedeploymanager.cpp
```diff
@@ -313,13 +375,14 @@ outcome::result<void> ServiceDeployManager::StartOrbitServicePrivileged() {
...
 
-  OUTCOME_TRY(loop_.exec());
+  OUTCOME_TRY(loop_->exec());
   QObject::connect(&orbit_service_task_.value(), &OrbitSshQt::Task::errorOccurred, this,
                    &ServiceDeployManager::handleSocketError);
   return outcome::success();
 }
 
 outcome::result<void> ServiceDeployManager::InstallOrbitServicePackage() {
+  CHECK(QThread::currentThread() == thread());
```
pierricgimmig, line 385: How does this work? From Qt: currentThread "Returns a pointer to a QThread which manages the currently executing thread." But what is thread() returning?

OrbitQt/servicedeploymanager.cpp
```diff
@@ -67,16 +86,36 @@ ServiceDeployManager::ServiceDeployManager(const DeploymentConfiguration* deploy
...
-      grpc_port_(grpc_port) {
+      grpc_port_(grpc_port),
+      ssh_watchdog_timer_(this) {
   CHECK(deployment_configuration != nullptr);
   CHECK(context != nullptr);
+
+  background_thread_.start();
+  moveToThread(&background_thread_);
+
+  // Event loop needs to be initialized in background thread to work properly on Linux
```
pierricgimmig, line 97: mega nit. Missing "."

pierricgimmig (approved):

OrbitQt/servicedeploymanager.cpp
```diff
@@ -176,9 +220,12 @@ outcome::result<void> ServiceDeployManager::StopSftpChannel(EventLoop* loop,
   return outcome::success();
 }
 
-void ServiceDeployManager::StopSftpChannel() { (void)StopSftpChannel(&loop_, sftp_channel_.get()); }
+void ServiceDeployManager::StopSftpChannel() {
+  (void)StopSftpChannel(&loop_.value(), sftp_channel_.get());
```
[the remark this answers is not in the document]
beckerhe: Yes, it is. The return type itself is markes as nodiscard.

OrbitQt/servicedeploymanager.cpp
beckerhe: `thread()` is `this->thread()`, so it calls `QObject::thread` [details](https://doc.qt.io/qt-5/qobject.html#thread)

Every QObject has an associated thread (which can be read by QThread::thread and changed by QObject::moveToThread). Whenever this object receives an event, the event dispatcher will determine the associated thread and queue an event handler task (a slot execution for example) on this thread's event loop.

OrbitQt/servicedeploymanager.cpp
beckerhe: Thanks. Easy to fix!

beckerhe: Done.

OrbitQt/servicedeploymanager.cpp
pierricgimmig: Ok, thanks for the info!

PR #1443, opened 2020-11-11
Move event buffering out of LinuxTracingHandler to CaptureEventBuffer

dpallotti:
#### Move event buffering out of LinuxTracingHandler to CaptureEventBuffer
This is in preparation of larger changes to `OrbitService` for the communication
with the Vulkan layer.

The concrete implementation `CaptureEventBufferAndResponseSender` also does
the sending and hence also replaces `CaptureEventListener` and
`GrpcCaptureResponseSender`. The two are merged as buffering is only functional
to sending, in order to avoid many tiny messages.
#### Split CaptureEventBufferAndResponseSender
Add `CaptureEventSender` (this is in addition to `CaptureEventBuffer`) and split
`CaptureEventBufferAndResponseSender` into `SenderThreadCaptureEventBuffer` and
`GrpcCaptureEventSender`.


OrbitService/CaptureServiceImpl.cpp
```diff
@@ -4,7 +4,7 @@
 
 #include "CaptureServiceImpl.h"
 
-#include "CaptureResponseListener.h"
```
dimitry-, line 7: The idea behind this interface was that we can have a second implementation for SaveToFileResponseListener, now the Buffer implemenation has gprc-spcific logic as well as bufferring logic, maybe keep them separate? have the buffer and delegate sending to this listener?

dimitry- (approved):

florian-kuebler (approved):
So from my side it looks good, thanks.
I'll leave it to Dimitry and you to decide if you want to keep the CaptureEventListener and have the CaptureEventBuffer delegating to it.

OrbitService/CaptureServiceImpl.cpp
dpallotti: Ok, I see. The way I was thinking it buffering and sending are tied together here as buffering is done in order to send fewer messages, but to save to file something different could be done, e.g., writing event by event as maybe the buffering is done by the os. That's why `CaptureEventBuffer` (but maybe the name should be different) only has `AddEvent`, and an additional `SaveToFile...` could have a different buffering logic. But this is unless you want to have the same way of buffering and be able to only swap sending on gRPC with saving to file.
So in other words if you think that later only the "sending" part should be swappable (with a fixed buffering strategy) I will separate the two interfaces, otherwise if you think that it is ok the way it is then I'll keep them together. Up to you.

dimitry-: Yep, I think buffering strategy with threading should be separate from sending/saving strategy and will likely be reused for savetofile

dpallotti: Ok, done, please re-review.

dimitry- (approved):
lgtm, thanks!

florian-kuebler (approved):
Yes, LGTM. Thank you!

PR #1444, opened 2020-11-11
Connection ui redesign

antonrohr:
Draft so @beckerhe and I can talk about this


OrbitQt/ProfilingTargetDialog.h
```diff
@@ -0,0 +1,80 @@
...
+  ProcessItemModel process_model_;
+  QSortFilterProxyModel process_proxy_model_;
+  bool connect_clicked_ = false;
+
+  ConnectionArtifacts* connection_artifacts_;
+
+  QPointer<OrbitGgp::Client> ggp_client_;
+  absl::flat_hash_map<std::string, OrbitSsh::Credentials> instance_credentials_;
+  MainThreadExecutor* main_thread_executer_;
+  std::variant<std::monostate, const ConnectionArtifacts*, QString> dialog_result_;
```
beckerhe, line 67: Now I realise that the type ConnectionArtifacts appears twice, once as an input values and once as an output (returned by Exec). Could these be two separate types? Or do they have exactly the same members and invariants?

beckerhe (commented):
Hi Anton, just some preliminary comments. Some might be due to confusion and might have never come up if this were finished already. So feel free to ignore those.

PR #1447, opened 2020-11-12
Text is no longer rendered in picking mode

reichlfl:
The text renderers do not know about picking, so adding text during picking-mode
leads to crashes.

Before, it was the developers' responsibility to not add any text to the renderers when
in picking mode, which is easy to forget and lead to crashes in the past. CaptureWindow
now makes sure that even if text was added to the renderers, it is not drawn while in
picking mode.

Bug: b/170308628

### DIFF
--- a/OrbitGl/CaptureWindow.cpp
+++ b/OrbitGl/CaptureWindow.cpp
@@ -520,8 +520,10 @@ void CaptureWindow::Draw() {
     ui_batcher_.DrawLayer(layer, GetPickingMode() != PickingMode::kNone);
 
     PrepareScreenSpaceViewport();
-    text_renderer_.DisplayLayer(&ui_batcher_, layer);
-    RenderText(layer);
+    if (GetPickingMode() == PickingMode::kNone) {
+      text_renderer_.DisplayLayer(&ui_batcher_, layer);
+      RenderText(layer);
+    }
   }
 }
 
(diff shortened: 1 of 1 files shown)


OrbitGl/CaptureWindow.cpp
```diff
@@ -519,9 +519,11 @@ void CaptureWindow::Draw() {
     time_graph_.GetBatcher().DrawLayer(layer, GetPickingMode() != PickingMode::kNone);
     ui_batcher_.DrawLayer(layer, GetPickingMode() != PickingMode::kNone);
 
-    PrepareScreenSpaceViewport();
-    text_renderer_.DisplayLayer(&ui_batcher_, layer);
-    RenderText(layer);
+    if (GetPickingMode() == PickingMode::kNone) {
+      PrepareScreenSpaceViewport();
```
karupayun, line 523: This method couldn't be in that if, or we have to add it again before also. Probably we can implement it in a different way (like having a bool flag related on which viewport we are), but we need to switch to the screen space viewport not only for the text, also because on some z-layer elements (the ones who are in the screen-space z-layer).

karupayun (approved):
Nice! :)

Did you try erasing all "if (!picking)" before each time that we are adding text?

OrbitGl/CaptureWindow.cpp
reichlfl: Right, thanks, I moved it out.

reichlfl:
> Nice! :)
> 
> Did you try erasing all "if (!picking)" before each time that we are adding text?

I was thinking about it, but I decided to not touch these parts to be sure - it doesn't hurt to not add the text in picking mode, this is just an additional safety net.

PR #1449, opened 2020-11-12
Only allow one Capture at a time in CaptureServiceImpl

dpallotti:
Also handle the `grpc::Status` returned by `Capture`.

The assumption of only one capture active at a time is in preparation of further
changes to `OrbitService` for the communication with the Vulkan layer.

As much as having a completely stateless `CaptureService` sounds nice:
- In the released version where the service is deployed, the fact that only
  one client at a time is connected to the service already enforces this,
  albeit by the client. Multiple captures in parallel are only possible when
  manually connecting to an already-started service.
- Even when it's possible to start multiple captures in parallel, they heavily
  interfere with each other: it's just better to diplay an error instead.

### DIFF
--- a/OrbitCaptureClient/CaptureClient.cpp
+++ b/OrbitCaptureClient/CaptureClient.cpp
@@ -123,8 +123,11 @@ void CaptureClient::Capture(ProcessData&& process,
   if (!reader_writer_->Write(request)) {
     ERROR("Sending CaptureRequest on Capture's gRPC stream");
     reader_writer_->WritesDone();
-    FinishCapture();
-    capture_listener_->OnCaptureFailed(ErrorMessage("Error sending capture request."));
+    ErrorMessageOr<void> finish_result = FinishCapture();
+    std::string error_string =
+        absl::StrFormat("Error sending capture request.%s",
+                        finish_result.has_error() ? ("\n" + finish_result.error().message()) : "");
+    capture_listener_->OnCaptureFailed(ErrorMessage{error_string});
     return;
   }
   LOG("Sent CaptureRequest on Capture's gRPC stream: asking to start capturing");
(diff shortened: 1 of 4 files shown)


OrbitCaptureClient/CaptureClient.cpp
```diff
@@ -200,18 +208,21 @@ bool CaptureClient::TryAbortCapture() {
...
-void CaptureClient::FinishCapture() {
+ErrorMessageOr<void> CaptureClient::FinishCapture() {
   if (reader_writer_ == nullptr) {
-    return;
+    return outcome::success();
   }
 
   grpc::Status status = reader_writer_->Finish();
+  ErrorMessageOr<void> result = outcome::success();
   if (!status.ok()) {
```
dimitry-, line 218: maybe just move this check to line 227, in which case you don't need temporary result variable here.

dimitry- (approved):
otherwize lgtm

OrbitCaptureClient/CaptureClient.cpp
dpallotti: Yep, done.

OrbitCaptureClient/CaptureClient.cpp
```diff
@@ -200,18 +208,23 @@ bool CaptureClient::TryAbortCapture() {
...
   client_context_.reset();
 
-  absl::MutexLock lock(&state_mutex_);
-  state_ = State::kStopped;
+  {
+    absl::MutexLock lock(&state_mutex_);
+    state_ = State::kStopped;
+  }
+
+  if (!status.ok()) {
```
dimitry-, line 225: I think it's fine to do it under the lock, if it makes code more readable.

dpallotti: I think it makes the code less readable, because it carries the question "Why does `status` need to be inspected under the lock?".

dimitry-: well, the answer here is - we want state_ to become kStopped once we exit this function, not earlier :) (I know there is no state change here but still). You can keep it as it is, I do not have strong feelings one way or the other.

florian-kuebler (approved):
LGTM. Thanks!

OrbitCaptureClient/CaptureClient.cpp
dpallotti: Indeed I don't really agree with that answer and that's why that would make it less readable. Why do we want `state_` to become `kStopped` when we exit the function? After all, there is other stuff happening after `FinishCapture` is called, so why exactly when `FinishCapture` exits? Instead, we want `state_` to become `kStopped` after `reader_writer_.reset(); client_context_.reset();`, and having the limited scope of the lock makes it clearer.

Of course this is a minor thing but that's my interpretation of why the lock should be as limited as possible, so that it's clearer where the state change actually happens.

PR #1450, opened 2020-11-12
Implement run capture logic in vulkan layer

kisenshi:
Implement the logic of the _OrbitVulkanLayerClientGgp_ to run Orbit captures automatically when the time per frame is higher than a certain threshold, set to 16.66ms by default. The triggered capture runs for a period of time, set to 10s as default, after which it stops and it is saved. The process continues until the game is closed. The steps followed required for this process to work are the following:
 
1. Intercept _QueuePresentKHR_. It is called once per frame so it is the one used to make the calculations and trigger the capture when needed.
2. Implement LayerLogic.cc/h which contains all the logic related to the process so the layer itself is transparent to it.
    - Execute capture service (which waits for calls related to starting/stopping a capture) in a new process
    - Shutdown the capture service safely when it is not needed anymore.
    - Instantiate capture client
    - Implement _ProcessQueuePresentKHR_
        - StartCapture when time frame drops 16.66ms
        - Stop and save capture after 10s
    - Use absl::time for time calculation

Data initialization and clean up are carried out when _CreateInstance_ and _DestroyInstance_ are called, respectively. It is assumed (and expected) that each of these calls happens once during the execution of the game; specifically when it starts and finishes. However, a flag marking the data initialization has also been included.


kisenshi:
As discussed offline, the arguments for the capture service as well as the frame rate threshold and capture length will be read from a config file, which will be included in a different PR as I am working on it at the moment.

OrbitVulkanLayerClientGgp/LayerTimes.h
```diff
@@ -0,0 +1,15 @@
+// Copyright (c) 2020 The Orbit Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#ifndef ORBIT_VULKAN_LAYER_CLIENT_GGP_LAYER_TIMES_H_
+#define ORBIT_VULKAN_LAYER_CLIENT_GGP_LAYER_TIMES_H_
+
+#include "absl/time/time.h"
+
+struct LayerTimes {
```
danielfenner, line 10: If you don't plan to use this struct anywhere else I would make the two Time's member of LayerLogic.  If you plan to pass these around separately later please ignore this.

OrbitVulkanLayerClientGgp/LayerLogic.h
```diff
@@ -0,0 +1,36 @@
...
+#include "OrbitCaptureGgpClient/OrbitCaptureGgpClient.h"
+
+// Contains the logic of the OrbitVulkanLayerClientGgp to run Orbit captures automatically when the
+// time per frame is higher than a certain threshold. It also instantiates the classes and variables
+// needed for this so the layer itself is transparent to it.
+class LayerLogic {
+ public:
+  LayerLogic() : data_initialised_{false}, orbit_capture_running_{false}, skip_logic_call_{true} {}
+
+  void InitLayerData();
```
danielfenner, line 20: I'd just call this "Init".

OrbitVulkanLayerClientGgp/LayerLogic.h
```diff
@@ -0,0 +1,36 @@
...
+  void CleanLayerData();
```
danielfenner, line 21: Maybe "Destroy" if this can only be called once or "Clear" if the LayerLogic is meant to be, well, recycled.

OrbitVulkanLayerClientGgp/LayerLogic.cpp
```diff
@@ -0,0 +1,116 @@
+// Copyright (c) 2020 The Orbit Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#include "LayerLogic.h"
+
+#include <math.h>
+#include <string.h>
+#include <unistd.h>
```
danielfenner, line 9: I would expect clang format to protest here:

````
#include "LayerLogic.h"

#include <unistd.h>

#include <math.h>
#include <string.h>
````
would be my expectation. Please ignore if clangformat is fine with this.

File: OrbitVulkanLayerClientGgp/LayerLogic.cpp [hunk 5]
```diff
@@ -0,0 +1,116 @@
...
+
+void LayerLogic::StartOrbitCaptureService() {
+  LOG("Starting Orbit capture service");
+  pid_t pid = fork();
+  if (pid < 0) {
+    ERROR("Fork failed; not able to start the capture service");
+  } else if (pid == 0) {
+    std::string game_pid = absl::StrFormat("%d", getppid());
+    // TODO(crisguerrero): Read the arguments from a config file.
+    char* argv[] = {strdup("/mnt/developer/OrbitCaptureGgpService"),
```
danielfenner, line 33: Using strdup is a memory leak. If allocates a copy of the string that is never freed.

OrbitVulkanLayerClientGgp/LayerLogic.cpp
```diff
@@ -0,0 +1,116 @@
...
+    orbit_capture_running_ = false;
+    skip_logic_call_ = true;
+  }
+}
+
+// QueuePresentKHR is called once per frame so we can calculate the time per frame. When this value
+// is higher than a certain threshold, an Orbit capture is started and runs during a certain period
+// of time; after which is stopped and saved.
+void LayerLogic::ProcessQueuePresentKHR() {
+  absl::Time current_frame = absl::Now();
```
danielfenner, line 76: really nit picking here: 
"current_frame" suggests (well, to me) an integer number identifining the frame. I'd call it "current_time".

OrbitVulkanLayerClientGgp/LayerLogic.cpp
```diff
@@ -0,0 +1,116 @@
...
+  // Ignore logic on the first call because times are not initialised. Also skipped right after a
+  // capture has been stopped
+  if (skip_logic_call_ == false) {
+    if (orbit_capture_running_ == false) {
+      absl::Duration frame_time = current_frame - layer_times_.last_frame;
+      if (isgreater(absl::ToDoubleMilliseconds(frame_time), kFrameTimeThreshold)) {
+        RunCapture();
+        LOG("Time frame is %fms and exceeds the %fms threshold; starting capture",
```
danielfenner, line 84: Id puit the logging before the call to RunCapture. The timestap will not include the time spent in RunCapture. And in case something goes wrong there the last log entry makes sense.

OrbitVulkanLayerClientGgp/LayerLogic.cpp
danielfenner: clangformat seems to be fine with this. So please ignore ...

OrbitVulkanLayerClientGgp/LayerTimes.h
kisenshi: That makes sense, I am not planning to use it anywhere else so I removed it. Thanks!

OrbitVulkanLayerClientGgp/LayerLogic.h
kisenshi: Done, thanks!

OrbitVulkanLayerClientGgp/LayerLogic.h
kisenshi: It is expected to only be called once so I have renamed it. Thanks!

OrbitVulkanLayerClientGgp/LayerLogic.cpp
kisenshi: Oh I was not aware of that; I have removed it. Thanks!

OrbitVulkanLayerClientGgp/LayerLogic.cpp
kisenshi: Makes sense, I have renamed it, thanks!

OrbitVulkanLayerClientGgp/LayerLogic.cpp
kisenshi: Yes, you are right; done, thanks!

kisenshi:
I have gone through the comments and made some updates to the code. Thanks!

danielfenner (approved):
lgtm

OrbitVulkanLayerClientGgp/LayerLogic.cpp
```diff
@@ -0,0 +1,113 @@
...
+    }
+  } else {
+    skip_logic_call_ = false;
+  }
+  last_frame_time_ = current_time;
+}
+
+void LayerLogic::RunCapture() {
+  int capture_started = ggp_capture_client_->StartCapture();
+  if (capture_started == 1) {
```
ronaldfw, line 100: The "1" should probably be a defined constant. 

constexpr const int kCaptureStartedResult = 1;

OrbitVulkanLayerClientGgp/LayerLogic.cpp
```diff
@@ -0,0 +1,113 @@
...
+    capture_started_time_ = absl::Now();
```
ronaldfw, line 101: Here and in other places: In other code we use std::chrono::steady_clock::now(), and not absl::Now(). It may use the same clock underneath, but we should keep this unified.

ronaldfw (approved):
LGTM, just some minor comments

OrbitVulkanLayerClientGgp/LayerLogic.cpp
```diff
@@ -0,0 +1,113 @@
...
+#include <unistd.h>
+
+#include "OrbitBase/Logging.h"
+#include "OrbitCaptureGgpClient/OrbitCaptureGgpClient.h"
+#include "absl/strings/str_format.h"
+#include "absl/time/time.h"
+
+namespace {
+static constexpr uint16_t kGrpcPort = 44767;
+static constexpr double kFrameTimeThreshold = 1000 / 60.0;  // milliseconds
```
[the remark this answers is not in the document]
kisenshi: Done!

OrbitVulkanLayerClientGgp/LayerLogic.cpp
```diff
@@ -0,0 +1,113 @@
...
+static constexpr int64_t kCaptureLength = 10;               // seconds
```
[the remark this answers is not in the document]
kisenshi: Done. I have also updated the name variable for kFrameTimeThreshold so the units are also included.

OrbitVulkanLayerClientGgp/LayerLogic.cpp
```diff
@@ -0,0 +1,113 @@
...
+                    NULL};
+
+    LOG("Making call to %s %s %s %s %s", argv[0], argv[1], argv[2], argv[3], argv[4]);
+    execv(argv[0], argv);
+  }
+}
+
+void LayerLogic::Init() {
+  // Although this method is expected to be called just once, we include a flag to make sure the
+  // gRPC service and client are not initialised more than once.
```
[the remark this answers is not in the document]
kisenshi: I have updated this comment as well as some others and the variable names to follow the AE convention. Thanks!

OrbitVulkanLayerClientGgp/LayerLogic.cpp
```diff
@@ -0,0 +1,113 @@
...
+}
+
+// QueuePresentKHR is called once per frame so we can calculate the time per frame. When this value
+// is higher than a certain threshold, an Orbit capture is started and runs during a certain period
+// of time; after which is stopped and saved.
+void LayerLogic::ProcessQueuePresentKHR() {
+  absl::Time current_time = absl::Now();
+  // Ignore logic on the first call because times are not initialised. Also skipped right after a
+  // capture has been stopped
+  if (skip_logic_call_ == false) {
```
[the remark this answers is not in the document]
kisenshi: This is much clearer indeed, thanks for the suggestion! I have updated the code

OrbitVulkanLayerClientGgp/LayerLogic.cpp
```diff
@@ -0,0 +1,113 @@
...
+    if (orbit_capture_running_ == false) {
```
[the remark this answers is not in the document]
kisenshi: Done, thanks!

OrbitVulkanLayerClientGgp/LayerLogic.cpp
```diff
@@ -0,0 +1,113 @@
...
+      absl::Duration frame_time = current_time - last_frame_time_;
+      if (isgreater(absl::ToDoubleMilliseconds(frame_time), kFrameTimeThreshold)) {
+        LOG("Time frame is %fms and exceeds the %fms threshold; starting capture",
+            absl::ToDoubleMilliseconds(frame_time), kFrameTimeThreshold);
+        RunCapture();
+      }
+    } else {
+      // Stop capture if it has been running enough time
```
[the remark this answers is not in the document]
kisenshi: Done!

OrbitVulkanLayerClientGgp/LayerLogic.cpp
kisenshi: As the same applies for the StopAndSaveCapture() call, I have named the const _kCaptureClientResultSuccess_ and I use them in both cases.

OrbitVulkanLayerClientGgp/LayerLogic.cpp
kisenshi: I have updated the code to use std::chrono instead of absl::Time in the file (also for duration calculation, etc). All seems to work as before.

kisenshi:
I have gone through the suggestions and updated the code; thanks!

OrbitVulkanLayerClientGgp/LayerLogic.cpp
```diff
@@ -0,0 +1,120 @@
...
+    LOG("Making call to %s %s %s %s %s", argv[0], argv[1], argv[2], argv[3], argv[4]);
+    execv(argv[0], argv);
+  }
+}
+
+void LayerLogic::Init() {
+  // Although this method is expected to be called just once, we include a flag to make sure the
+  // gRPC service and client are not initialized more than once.
+  if (data_initialized_ == false) {
+    LOG("Making initialisations required in the layer");
```
ronaldfw, line 48: initialisations --> initializations

ronaldfw (approved):
LGTM

OrbitVulkanLayerClientGgp/LayerLogic.cpp
kisenshi: Oops I totally missed this one. I have updated it (and also another comment I also missed in _OrbitCaptureClientLayer.cpp_. Thanks!

kisenshi:
Merging this now. Thanks!

PR #1451, opened 2020-11-12
Live updates to frame tracks during a capture

ronaldfw:
We now update frame tracks live during a capture by processing each
timer that belongs to a function call, creating a frame track timer
from its timestamp and the previously encountered timestamp for this
function call.

Note that timers may arrive out of order, in particular when a frame
track function was chosen that is called from multiple threads. During
live capture we do not correct for this, but at the end of the capture,
the entire frame track is refreshed to correct any mistakes made during
capture.

Bug: http://b/173100696
Tested: Unit tests; various manual tests, including serialization of
frame tracks.

### DIFF
--- a/OrbitGl/FrameTrackOnlineProcessor.cpp
+++ b/OrbitGl/FrameTrackOnlineProcessor.cpp
@@ -1,0 +1,54 @@
+// Copyright (c) 2020 The Orbit Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#include "FrameTrackOnlineProcessor.h"
+
+#include <limits>
+
+#include "CaptureData.h"
+#include "TimeGraph.h"
+#include "UserDefinedCaptureData.h"
+
+FrameTrackOnlineProcessor::FrameTrackOnlineProcessor(const CaptureData& capture_data,
+                                                     TimeGraph* time_graph)
+    : time_graph_(time_graph) {
+  const UserDefinedCaptureData& user_defined_capture_data =
+      capture_data.user_defined_capture_data();
+  for (const auto& function : user_defined_capture_data.frame_track_functions()) {
+    const uint64_t function_address = capture_data.GetAbsoluteAddress(function);
+    current_frame_track_functions_.insert(function_address);
+    previous_timestamp_ns_.insert(
+        std::make_pair(function_address, std::numeric_limits<uint64_t>::max()));
+  }
+}
+
+void FrameTrackOnlineProcessor::ProcessTimer(const orbit_client_protos::TimerInfo& timer_info,
+                                             const orbit_client_protos::FunctionInfo& function) {
+  uint64_t function_address = timer_info.function_address();
+  if (!current_frame_track_functions_.contains(function_address)) {
+    return;
+  }
+  uint64_t previous_timestamp_ns = previous_timestamp_ns_.at(function_address);
+  if (previous_timestamp_ns == std::numeric_limits<uint64_t>::max()) {
+    previous_timestamp_ns_[function_address] = timer_info.start();
+    return;
+  }
+
+  if (previous_timestamp_ns < timer_info.start()) {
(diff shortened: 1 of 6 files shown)


OrbitGl/FrameTrackOnlineProcessor.cpp
```diff
@@ -0,0 +1,50 @@
...
+    current_frame_track_functions_.insert(function_address);
+    previous_timestamp_ns_.insert(
+        std::make_pair(function_address, std::numeric_limits<uint64_t>::max()));
+  }
+}
+
+void FrameTrackOnlineProcessor::ProcessTimer(const orbit_client_protos::TimerInfo& timer_info,
+                                             const orbit_client_protos::FunctionInfo& function) {
+  uint64_t function_address = timer_info.function_address();
+  if (current_frame_track_functions_.empty() ||
```
florian-kuebler, line 24: If you think that this shortcut is really helpful in terms of performance, I'd add a comment about that.

OrbitGl/FrameTrackOnlineProcessor.cpp
```diff
@@ -0,0 +1,50 @@
...
+      !current_frame_track_functions_.contains(function_address)) {
+    return;
+  }
+  uint64_t previous_timestamp_ns = previous_timestamp_ns_[function_address];
```
florian-kuebler, line 28: We had the discussion at a couple of places, and always concluded to use `.at()` over `[]` for reading.
Currently, it is clear that the function is in the map and `[]` has no side-effect, but that might change it the future.

Maybe we should add this to our guidelines (do we actually have them yet?).

OrbitGl/FrameTrackOnlineProcessor.cpp
```diff
@@ -0,0 +1,50 @@
...
+  if (previous_timestamp_ns == std::numeric_limits<uint64_t>::max()) {
+    previous_timestamp_ns_[function_address] = timer_info.start();
+    return;
+  }
+
+  if (previous_timestamp_ns < timer_info.start()) {
+    orbit_client_protos::TimerInfo frame_timer;
+
+    // TID is meaningless for this timer (start and end can be on two different threads).
```
florian-kuebler, line 37: We should probably think of refactoring timers at some point to introduce a class hierarchy for it. It has a couple of fields that are only used for certain types. Also we do have a couple of switches over the type, which can be simplified by using virtual methods or a visitor patter. 
Anyway, this is clearly out of scope for this PR.

florian-kuebler (approved):
Beside very minor thinks, LGTM. Thanks!

OrbitGl/FrameTrackOnlineProcessor.cpp
```diff
@@ -0,0 +1,50 @@
+#include "FrameTrackOnlineProcessor.h"
```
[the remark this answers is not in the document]
ronaldfw: Done.

OrbitGl/FrameTrackOnlineProcessor.cpp
ronaldfw: No, this was just an experiment. I don't think it makes a meaningful difference. Thanks for spotting!

OrbitGl/FrameTrackOnlineProcessor.cpp
ronaldfw: Done.

OrbitGl/FrameTrackOnlineProcessor.cpp
ronaldfw: Yes, it's a bit awkward to put in a meaningless value here. 

Will resolve for now, but let's chat about our options how to handle this in the future.

OrbitGl/FrameTrackOnlineProcessor.cpp
ronaldfw: Good idea to add stuff like that into our coding guidelines. I think the only part we have here is this: https://github.com/google/orbit/blob/master/DEVELOPMENT.md#consistent-code-styling

PR #1452, opened 2020-11-12
Fix crash when dragging a thread state track

karupayun:
Only when you have enabled track_ordering_feature, this class inherit
OnDrag from Track and crashed.

Related to b/173093860.


dpallotti (approved):
Thanks Pablo!

pierricgimmig:
nit: Description should be "*Fix* crash when..."

PR #1453, opened 2020-11-12
[Draft] Optimize buffer handling in OrbitSshQt

beckerhe:
This is based on #1442. Only the last two commits belong to this PR.

@pierricgimmig pointed out that tunnel processing logic makes unnecessary copies.
This fixes that by removing temporaries and replacing std::string::substr by memmove.


beckerhe:
I will wait for Pierric's client intospection to land.

PR #1454, opened 2020-11-12
Fix recently introduced text rendering artifact

pierricgimmig:
Use floorf() over static_cast[int] to make sure that characters are 
consistently positioned at pixel boundary in screen space and that
the "shift" is always in the same direction for negative values.


reichlfl (approved):

karupayun (approved):

ronaldfw (approved):
LGTM

PR #1455, opened 2020-11-13
Remove wrong CHECK in ServiceDeployManager

beckerhe:
CopyFileToLocal is intended to be called from a different thread, so the
CHECK shouldn't be there. (Copy-and-paste error).


florian-kuebler (approved):
Okay, thanks!

PR #1456, opened 2020-11-16
Update dependencies in lockfile

beckerhe:
The official conan-center remote does not contain packages without
revisions anymore. So we should also switch our dependencies to
revisions-enabled ones. Otherwise the external build will be broken.

Bug: http://b/172125379


antonrohr (approved):

pierricgimmig (commented):
Thanks for looking into this Henning!

PR #1458, opened 2020-11-16
Move the view when dragging a track to the border

karupayun:
Before this, we couldn't move a track behind or above to some track that
wasn't in the screen. Now if we drag a track to the top or to the
bottom, the view is going to move.

Also, we refactored and renamed UpdateTopLeftY so each time that we update
where is the topY on the screen, we check that fits in the min-max range
of the slider.

Lastly, temporarily we don't pin the scheduler track.

### DIFF
--- a/OrbitGl/SchedulerTrack.cpp
+++ b/OrbitGl/SchedulerTrack.cpp
@@ -18,7 +18,7 @@ using orbit_client_protos::TimerInfo;
 const Color kInactiveColor(100, 100, 100, 255);
 const Color kSelectionColor(0, 128, 255, 255);
 
-SchedulerTrack::SchedulerTrack(TimeGraph* time_graph) : TimerTrack(time_graph) { SetPinned(true); }
+SchedulerTrack::SchedulerTrack(TimeGraph* time_graph) : TimerTrack(time_graph) { SetPinned(false); }
 
 float SchedulerTrack::GetHeight() const {
   TimeGraphLayout& layout = time_graph_->GetLayout();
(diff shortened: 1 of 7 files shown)


OrbitGl/SchedulerTrack.cpp
```diff
@@ -18,7 +18,7 @@ using orbit_client_protos::TimerInfo;
 const Color kInactiveColor(100, 100, 100, 255);
 const Color kSelectionColor(0, 128, 255, 255);
 
-SchedulerTrack::SchedulerTrack(TimeGraph* time_graph) : TimerTrack(time_graph) { SetPinned(true); }
+SchedulerTrack::SchedulerTrack(TimeGraph* time_graph) : TimerTrack(time_graph) { SetPinned(false); }
```
pierricgimmig, line 21: Is this needed for the track dragging?

pierricgimmig (approved):
Hi Pablo, LGTM, although I had on minor question about the scheduler track not being pinned anymore.

OrbitGl/SchedulerTrack.cpp
karupayun: There are a few small issues.

It isn't needed, but I'm not sure how we want to implement pinning, maybe we can make something different but didn't think too much about that. I'm trying to push moving tracks to 1.56, and I don't have priority on Pinning for now.

pierricgimmig: Maybe we should leave it in for now, and disable in a separate PR? @reichlfl ?

reichlfl: I think it's fine to just disable the pinning of this track. It's independent of the track ordering, and the SchedulerTrack is only pinned to test the pinning-feature (when the feature flag is enabled), so there's no harm in changing this. Agree with Pablo, this way we can safely remove the feature flag since the user has no way of enabling track pinning through the UI.

PR #1460, opened 2020-11-16
Remove before-and-after-expanding row count check from BottomUp e2e test

dpallotti:
When not all rows of the bottom-up view (which can be many) fit on the screen
(i.e., there's a vertical scrollbar), listing all items excludes the
non-visible items (probably because Qt correctly doesn't hold them in memory
when not necessary). Therefore, comparing the number of rows before and after
expanding an item doesn't yield the expected result and causes the test to
fail.

Bug: http://b/173180780

Test: Run locally, resizing Orbit's window to a small size. Verify that the
test was previously failing and now succeeds.


danielfenner (approved):
lgtm!

PR #1461, opened 2020-11-17
Remove ServiceDeployManager::loop_

beckerhe:
loop_ used to be a member variable of ServiceDeployManager which has
been used in a couple of function rather independently.

The recent threading changes have made the usage of loop_ even more
brittle because due to a bug in Qt it has to be constructed on the
thread that will eventually use the event loop.

This highlights that we use QEventLoop/EventLoop in a way it was not
designed for. We should instantiate it locally whenever we need it.

So this commit replaces the "global" long-living loop_ by local
short-living instances inside of the member functions.

That makes the code easier to understand and we don't need to rely on
comments to protect the previous rather brittle behaviour.

### DIFF
--- a/OrbitQt/servicedeploymanager.cpp
+++ b/OrbitQt/servicedeploymanager.cpp
@@ -12,7 +12,6 @@
 
 #include <QApplication>
 #include <QDir>
-#include <QEventLoop>
 #include <chrono>
 #include <system_error>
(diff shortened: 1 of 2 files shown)


antonrohr (approved):
lgtm, thank you!

beckerhe:
Thanks Anton!

PR #1462, opened 2020-11-17
Support saving and loading frame tracks as part of presets

ronaldfw:
Frame tracks based on hooked functions are now also stored in and
loaded from presets.

Bug: http://b/173490995

Tested: Manual tests: load/save presets with frame tracks; load/save
presets without frame tracks, both during different states of Orbit
including before and after capture.

### DIFF
--- a/OrbitGl/App.cpp
+++ b/OrbitGl/App.cpp
@@ -641,6 +641,12 @@ ErrorMessageOr<void> OrbitApp::SavePreset(const std::string& filename) {
     (*preset.mutable_path_to_module())[function.loaded_module_path()].add_function_hashes(hash);
   }
 
+  for (const auto& function : data_manager_->user_defined_capture_data().frame_track_functions()) {
+    uint64_t hash = FunctionUtils::GetHash(function);
+    (*preset.mutable_path_to_module())[function.loaded_module_path()]
+        .add_frame_track_function_hashes(hash);
+  }
+
   std::string filename_with_ext = filename;
   if (!absl::EndsWith(filename, ".opr")) {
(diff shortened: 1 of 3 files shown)


OrbitGl/App.cpp
```diff
@@ -1196,8 +1261,23 @@ void OrbitApp::UpdateProcessAndModuleList(int32_t pid) {
...
+        const ModuleData* module = module_manager_->GetModuleByPath(func.loaded_module_path());
+        if (!process->IsModuleLoaded(module->file_path())) {
+          RemoveFrameTrack(func);
+        } else {
+          if (!module->is_loaded()) {
+            RemoveFrameTrack(func);
+            frame_track_function_hashes_map[module->file_path()].push_back(
+                FunctionUtils::GetHash(func));
+          }
+        }
```
IrinaShkviro, line 1277:
```suggestion
        } else if (!module->is_loaded()) {
            RemoveFrameTrack(func);
            frame_track_function_hashes_map[module->file_path()].push_back(
                FunctionUtils::GetHash(func));
        }
```
Also maybe it worth to add an explanatory comment for 'if' branches.

IrinaShkviro (approved):
Looks great from my perspective!

OrbitGl/App.cpp
ronaldfw: Thanks, good suggestion. Fixed.

antonrohr (approved):
code lgtm and I also tested it with one frame track based on vkQueueSubmit.

ronaldfw:
Thanks for the reviews, @IrinaShkviro and @antonrohr!

PR #1463, opened 2020-11-17
Include config file to capture vulkan layer

kisenshi:
Allow users to provide capture service arguments (file and log directory, functions, sampling rate) and layer options (frame time
threshold and capture length) in a config file. Some of these are given a default value if not provided in the config file. The config file is expected to be `/mnt/developer/orbit_trigger_capture_vulkan_layer_config.pb.txt`

The data allowed in the config file is the following:

- _log_directory_: By default a log file is always created for the logs for _OrbitCaptureGgpService_. This directory is set to `/var/game/` if not provided in the config file.
- _frame_time_threshold_ms_: By default is set to 16.66ms
- _capture_length_s_: By default is set to 10s
- _file_directory_: This variable is given a default value in _OrbitCaptureGgpService_ (`/var/game/`) so if it is not provided it won't be included in the call but it will be set to this directory directly.
- _sampling_rate_: This variable is given a default value in _OrbitCaptureGgpService_ (1000) so if it is not provided it won't be included in the call but it will be set to this value directly.
- _functions_ by default are not set so the capture will run without hooked functions.

All data is managed inside a new class created called _LayerOptions_. This class manages the parsing of the config file into the proto, getting the correct values for the threshold and capture length as well as building the array containing all the string arguments that are provided to _OrbitCaptureGgpService_ in the execv call.

The PR includes an example of the expected config file, containing the default values, as reference.

Finally, the "enable_environment" has been set in the json file of the layer so the environment variable _ENABLE_ORBIT_TRIGGER_CAPTURE_VULKAN_LAYER=1_ is needed to be set to load the layer when running the game.


OrbitVulkanLayerClientGgp/LayerData.cpp
```diff
@@ -0,0 +1,125 @@
...
+  // Config is a proto text file
+  int config_file_descriptor = open(kConfigFileName, O_RDONLY);
+  if (config_file_descriptor < 0) {
+    ERROR("Not being able to open file. Default values will be used");
+    return;
+  }
+
+  google::protobuf::io::FileInputStream config_file_stream(config_file_descriptor);
+  if (!google::protobuf::TextFormat::Parse(&config_file_stream, &layer_config_)) {
+    ERROR("Parsing vulkan layer config file. Default values will be used");
```
ronaldfw, line 43: Would it be an option to actually abort at this point? If you can't read the config file, then there is an issue that the user did not intend that way (maybe the config file is malformed and needs to be fixed). Even though there is an error message in the log, continuing here helps hiding the fact that something is wrong.

OrbitVulkanLayerClientGgp/LayerData.cpp
```diff
@@ -0,0 +1,125 @@
...
+    layer_config_.Clear();
+  } else {
+    LOG("Config data loaded successfully");
+  }
+
+  if (close(config_file_descriptor) < 0) {
+    ERROR("Closing config file");
```
ronaldfw, line 50: Also print a string based on errno to get more information about the error that occurred (using SafeStrerror).

OrbitVulkanLayerClientGgp/LayerData.cpp
```diff
@@ -0,0 +1,125 @@
...
+}  // namespace
+
+void LayerData::Init() {
+  // Load data from config file
+  LOG("Loading  vulkan layer config file from %s", std::string(kConfigFileName));
+
+  // Config is a proto text file
+  int config_file_descriptor = open(kConfigFileName, O_RDONLY);
+  if (config_file_descriptor < 0) {
+    ERROR("Not being able to open file. Default values will be used");
```
ronaldfw, line 37: Also print a string based on errno to get more information about the error that occurred (using SafeStrerror).

OrbitVulkanLayerClientGgp/LayerData.cpp
```diff
@@ -0,0 +1,125 @@
...
+  } else {
+    LOG("Config data loaded successfully");
+  }
+
+  if (close(config_file_descriptor) < 0) {
+    ERROR("Closing config file");
+  }
+}
+
+double LayerData::getFrameTimeThresholdMilliseconds() {
```
ronaldfw, line 54: Function name should start with upper case.

OrbitVulkanLayerClientGgp/LayerData.cpp
```diff
@@ -0,0 +1,125 @@
...
+  if (layer_config_.has_layer_options() &&
+      layer_config_.layer_options().frame_time_threshold_ms() > 0) {
+    return layer_config_.layer_options().frame_time_threshold_ms();
+  }
+  return kFrameTimeThresholdMillisecondsDefault;
+}
+
+uint32_t LayerData::getCaptureLengthSeconds() {
```
ronaldfw, line 62: Function name should start with upper case.

OrbitVulkanLayerClientGgp/LayerData.cpp
```diff
@@ -0,0 +1,125 @@
...
+  if (layer_config_.has_layer_options() && layer_config_.layer_options().capture_length_s() > 0) {
+    return layer_config_.layer_options().capture_length_s();
+  }
+  return kCaptureLengthSecondsDefault;
+}
+
+std::vector<char*> LayerData::buildObritCaptureServiceArgv(const std::string& game_pid) {
```
ronaldfw, line 69: Upper case and fix typo in name ("Obrit").

OrbitVulkanLayerClientGgp/LayerData.cpp
```diff
@@ -0,0 +1,125 @@
...
+  LOG("Building argv");
```
ronaldfw, line 70: Not sure this is a useful log message. Printing the actual argv constructed might be.

OrbitVulkanLayerClientGgp/LayerData.cpp
```diff
@@ -0,0 +1,125 @@
...
+  std::vector<char*> argv;
```
ronaldfw, line 71: I suppose the type here and some of the weirdness below (like having some strings as members) is due to the interface required by execv. I would probably try to keep it clean here (in LayerData) by constructing a vector of std::string that is returned by value and then only converting to the char* array where it's needed. That would make it easier to control the lifetime of the strings involved (just keep the vector alive). 

But feel free to keep as-is, I don't think this is a big problem.

OrbitVulkanLayerClientGgp/layer_config.proto
```diff
@@ -0,0 +1,32 @@
...
+message LayerConfig {
+  CaptureServiceArguments capture_service_arguments = 1;
+
+  LayerOptions layer_options = 2;
+}
+
+message CaptureServiceArguments {
+  repeated string functions = 1;  // Functions to hook to the capture
+
+  string file_directory =
```
ronaldfw, line 18: Formatting looks really weird here. I would put the comment above the line to avoid the line break. 

```
// Path to locate .orbit file...
string file_directory = 2;
```

OrbitVulkanLayerClientGgp/LayerData.cpp
kisenshi: Mmm I am not sure if it is possible to force the abort of the game from the vulkan layer; let me try and let you know.

I decided to print the error and continue with default values because I at the time I considered it would be too extreme to abort the execution of the game because of that buy maybe it would actually make sense, as I guess they would run the layer in the case they actually want to use it. Will check if it is possible to do this.

OrbitVulkanLayerClientGgp/orbit_capture_vulkan_layer_config.pb.txt
```diff
@@ -0,0 +1,18 @@
...
+# found in the LICENSE file.
+#
+# proto-file: layer_config.proto
+# proto-message: orbit_vulkan_capture_protos.LayerConfig
+
+capture_service_arguments {
+  functions: []
+  file_directory: ""
+  log_directory: ""
+  sampling_rate: 0
```
danielfenner, line 12: This is thought to be a template for user provided config files, right?

I'd put the defaults from OrbitVulkanLayerClientGgp/LayerData.cpp into this file so when someone wants to use this there is a reasonable value they can alter.

OrbitVulkanLayerClientGgp/LayerData.cpp
```diff
@@ -0,0 +1,125 @@
+// Copyright (c) 2020 The Orbit Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#include "LayerData.h"
```
danielfenner, line 5: LayerOptions.{h,cpp}  would be the better name imo. Feel free to ignore ...

OrbitVulkanLayerClientGgp/LayerData.cpp
kisenshi: yes, execv needs to get the arguments in a certain way and this was the best I could manage to make it work. I can try to clean it up as you suggest as it is true it looks quite convoluted

OrbitVulkanLayerClientGgp/CMakeLists.txt
```diff
@@ -2,21 +2,33 @@
 # Use of this source code is governed by a BSD-style license that can be
 # found in the LICENSE file.
 
-project(OrbitVulkanLayerClientGgp)
-add_library(OrbitVulkanLayerClientGgp SHARED)
+project(OrbitCaptureVulkanLayer)
```
danielfenner, line 5: I really don't want to overemphasis this naming stuff - fell free to ignore... OrbitTriggerCaptureVulkanLayer would be way more destinctive  imo.

OrbitVulkanLayerClientGgp/LayerData.cpp
kisenshi: If I use FATAL here it prints the error and the vulkan execution finished and the game does not load, however it does not finish either, it just hangs. Would this behavior be what we want?

OrbitVulkanLayerClientGgp/orbit_capture_vulkan_layer_config.pb.txt
kisenshi: Yes, this is just the skeleton as example. I decided not to put the default values mainly because some of the default values are not set in the layer (e.g sampling_rate and log_directory) and frame_time_threshold_ms would not be as exact as the one set in the layer 16.66 !=  1000.0 / 60.0. Should I just go with this approximation and leave the non default ones as they are?

OrbitVulkanLayerClientGgp/LayerData.cpp
kisenshi: Done, thanks!

OrbitVulkanLayerClientGgp/LayerData.cpp
kisenshi: Done, thanks!

OrbitVulkanLayerClientGgp/LayerData.cpp
kisenshi: Done!

OrbitVulkanLayerClientGgp/LayerData.cpp
kisenshi: Done!

OrbitVulkanLayerClientGgp/LayerData.cpp
kisenshi: Done, thanks!

OrbitVulkanLayerClientGgp/LayerData.cpp
kisenshi: Yep that was there for testing and forgot to remove it thanks! The argv elements are printed just before execv is called

OrbitVulkanLayerClientGgp/layer_config.proto
kisenshi: Fixed

OrbitVulkanLayerClientGgp/CMakeLists.txt
kisenshi: That was a quick change, so I have renamed it. I agree it makes sense. Thanks for the suggestion!

OrbitVulkanLayerClientGgp/LayerData.cpp
kisenshi: I have updated and the code is much readable now. Thanks for the suggestion!

OrbitVulkanLayerClientGgp/LayerData.cpp
kisenshi: I think this rename makes sense and it was a quick change so I have renamed the class and variable, thanks for the suggestion!

kisenshi:
I have gone through the comments and addressed most of them. I left two comments/questions that need an answer before I update the code for those. Thanks!

kisenshi:
Also, given that I have this PR and I have been updating/renaming some of the code should I also rename the folder (_OrbitVulkanLayerClientGgp_) to _OrbitVulkanLayerTriggerCapture_?

OrbitVulkanLayerClientGgp/orbit_capture_vulkan_layer_config.pb.txt
kisenshi: Alternatively I could add a comment on the right showing the default values as a reference.

danielfenner: I think the accuracy is not an issue at all. Especially since it is related to timing and that is a bit handwavy anyways.

Even if you put in just some arbitray working values e.g. for sampling_rate I think this is helpful to get a user started.

danielfenner:
> Also, given that I have this PR and I have been updating/renaming some of the code should I also rename the folder (_OrbitVulkanLayerClientGgp_) to _OrbitVulkanLayerTriggerCapture_?

Since the project is now called OrbitTriggerCaptureVulkanLayer I think it makes sense to change the dir to the same name.

OrbitVulkanLayerClientGgp/orbit_capture_vulkan_layer_config.pb.txt
kisenshi: I have included example values (corresponding to the default ones) in the config file :)

kisenshi:
I have included example values in the config file and renamed the project directory to match the project name (_OrbitTriggerCaptureVulkanLayer_)

danielfenner (approved):
lgtm!

OrbitVulkanLayerClientGgp/LayerData.cpp
kisenshi: @ronaldfw What should I do about this?

kisenshi: Discussed offline: fine to just print the error.

kisenshi:
Merging this now. Thanks!

PR #1464, opened 2020-11-17
Add logging to GetModuleList.

danielfenner:
There might be an issue in the code collecting the information about
the modules. Occationally the request fails in the e2e tests due to an
exceeded deadline in the grpc communication.
This logging is a bit too much for my taste so I filed
http://b/173498020 to clean it up after the underlying bug is fixed.

Bug: http://b/173358603
Test: Ran locally, logging looks helpful now.


antonrohr (approved):
lgtm. 
Did not test it, because the code is simple enough.

PR #1465, opened 2020-11-18
Manual instrumentation API fixes

pierricgimmig:
Orbit.h contains two versions of "orbit_api::Start" style functions.
One is usually inlined and optimized out, leaving only the empty stub
to hook into. In certain configurations, the inlining does not occur and
we have two functions with the same name. Make sure we force inlining to
avoid the problem. This fixes the duplicate scope issue with garbled
strings for manually instrumented functions.

Replace direct internal includes of "Orbit.h" by "OrbitBase/Tracing.h".

Add missing internal implementation of AsyncString(...).

Fix static assert errors for C++ versions <17.

Fix compilation warning when _WIN32 is undefined.

Fixes b/173449243

### DIFF
--- a/Orbit.h
+++ b/Orbit.h
@@ -282,7 +282,7 @@ enum class Color : uint32_t {
 
 #if defined(_WIN32)
 #define ORBIT_STUB inline __declspec(noinline)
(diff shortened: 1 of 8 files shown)


ronaldfw (approved):
LGTM

PR #1466, opened 2020-11-18
Use std::optional for CaptureData to indicate when it is available

ronaldfw:
We have relied on CaptureData to be empty in a sane way when no actual
capture data is available, for example, we draw empty captures and we
save empty captures, even when no capture was loaded or captured.

We now use std::optional to indicate whether actual capture data is
available and react accordingly when drawing, updating, and saving
captures. That is, in particular when no capture data is available,
these actions are skipped or not allowed (users now cannot save anymore
when no capture is available).

This is mainly a refactor, but some user visible changes were needed to
disallow saving captures when no capture is available.

Tested: Ran unit tests; lots of manual testing with loading/saving/
capturing profiles, etc.

### DIFF
--- a/OrbitClientGgp/ClientGgp.cpp
+++ b/OrbitClientGgp/ClientGgp.cpp
@@ -101,7 +101,7 @@ bool ClientGgp::SaveCapture() {
   const auto& key_to_string_map = string_manager_->GetKeyToStringMap();
   std::string file_name = options_.capture_file_name;
   if (file_name.empty()) {
-    file_name = capture_serializer::GetCaptureFileName(capture_data_);
+    file_name = capture_serializer::GetCaptureFileName(GetCaptureData());
   } else {
     // Make sure the file is saved with orbit extension
     capture_serializer::IncludeOrbitExtensionInFile(file_name);
@@ -110,7 +110,7 @@ bool ClientGgp::SaveCapture() {
   file_name.insert(0, options_.capture_file_directory);
 
   ErrorMessageOr<void> result = capture_serializer::Save(
-      file_name, capture_data_, key_to_string_map, timer_infos_.begin(), timer_infos_.end());
+      file_name, GetCaptureData(), key_to_string_map, timer_infos_.begin(), timer_infos_.end());
   if (result.has_error()) {
     ERROR("Could not save the capture: %s", result.error().message());
     return false;
@@ -256,7 +256,7 @@ void ClientGgp::UpdateCaptureFunctions(std::vector<std::string> capture_function
 }
 
 void ClientGgp::ClearCapture() {
-  capture_data_ = CaptureData();
+  capture_data_.reset();
   string_manager_->Clear();
   timer_infos_.clear();
 }
@@ -275,8 +275,8 @@ void ClientGgp::OnCaptureStarted(
 
 void ClientGgp::OnCaptureComplete() {
   LOG("Capture completed");
-  SamplingProfiler sampling_profiler(*capture_data_.GetCallstackData(), capture_data_);
-  capture_data_.set_sampling_profiler(sampling_profiler);
+  SamplingProfiler sampling_profiler(*GetCaptureData().GetCallstackData(), GetCaptureData());
+  GetMutableCaptureData().set_sampling_profiler(sampling_profiler);
 }
 
 void ClientGgp::OnCaptureCancelled() {}
@@ -286,11 +286,11 @@ void ClientGgp::OnCaptureFailed(ErrorMessage /*error_message*/) {}
 void ClientGgp::OnTimer(const orbit_client_protos::TimerInfo& timer_info) {
   if (timer_info.function_address() > 0) {
     const FunctionInfo* func =
-        capture_data_.FindFunctionByAddress(timer_info.function_address(), false);
+        GetCaptureData().FindFunctionByAddress(timer_info.function_address(), false);
     // For timers, the function must be present in the process
     CHECK(func != nullptr);
     uint64_t elapsed_nanos = timer_info.end() - timer_info.start();
-    capture_data_.UpdateFunctionStats(*func, elapsed_nanos);
+    GetMutableCaptureData().UpdateFunctionStats(*func, elapsed_nanos);
   }
   ProcessTimer(timer_info);
 }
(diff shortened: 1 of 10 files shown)


OrbitGl/DataManager.h
```diff
@@ -60,6 +60,9 @@ class DataManager final {
   [[nodiscard]] const UserDefinedCaptureData& user_defined_capture_data() const {
     return user_defined_capture_data_;
   }
+  [[nodiscard]] UserDefinedCaptureData& user_defined_capture_data() {
```
florian-kuebler, line 63: Again, I'd prefer to give this method a more explicit name, e.g. containing "Mutable".

OrbitGl/LiveFunctionsDataView.cpp
```diff
@@ -304,26 +304,30 @@ void LiveFunctionsDataView::DoFilter() {
   indices_ = indices;
 
   // Filter drawn textboxes
-  absl::flat_hash_set<uint64_t> visible_functions;
-  const CaptureData& capture_data = GOrbitApp->GetCaptureData();
-  for (size_t i = 0; i < indices_.size(); ++i) {
-    FunctionInfo* func = GetSelectedFunction(i);
-    visible_functions.insert(capture_data.GetAbsoluteAddress(*func));
+  if (GOrbitApp->HasCaptureData()) {
```
florian-kuebler, line 307: Maybe instead `if (!GOrbitApp->HasCaptureData()) { return; }`

OrbitGl/LiveFunctionsDataView.cpp
```diff
@@ -304,26 +304,30 @@ void LiveFunctionsDataView::DoFilter() {
...
-  const absl::flat_hash_map<uint64_t, orbit_client_protos::FunctionInfo>& selected_functions =
-      GOrbitApp->GetCaptureData().selected_functions();
-  size_t functions_count = selected_functions.size();
-  indices_.resize(functions_count);
-  size_t i = 0;
-  for (const auto& pair : selected_functions) {
-    functions_.push_back(pair.second);
-    indices_[i] = i;
-    ++i;
+  if (GOrbitApp->HasCaptureData()) {
```
florian-kuebler, line 320: Again, `if (!GOrbitApp->HasCaptureData()) { return; }` ?

OrbitQt/orbitmainwindow.cpp
```diff
@@ -545,6 +548,7 @@ void OrbitMainWindow::on_actionOpen_Preset_triggered() {
     }
     break;
   }
+  ui->actionSave_Preset_As->setDisabled(false);
```
florian-kuebler, line 551: I am a bit confused, could you explain why this is the correct place to enable preset saving?
Same for the Capture below?
It feels a bit unintuitive.

florian-kuebler (commented):
Thanks that change. I would discuss the way to disable saving.

OrbitQt/orbitmainwindow.cpp
ronaldfw: When you open a preset or a capture, you can then save it again. The other place where we enable preset saving is the capture_finished_callback above, which is called on stopping the capture. 

Regarding presets the situation is also that you can open a preset without actually having capture data.

I'll add comments to clarify if you agree this all makes sense.

OrbitQt/orbitmainwindow.cpp
```diff
@@ -642,6 +646,8 @@ void OrbitMainWindow::OpenCapture(const std::string& filepath) {
   GOrbitApp->OnLoadCapture(filepath);
   setWindowTitle(QString::fromStdString(filepath));
   ui->MainTabWidget->setCurrentWidget(ui->CaptureTab);
+  ui->actionSave_Capture->setDisabled(false);
```
antonrohr, line 649: maybe nit, maybe its fine as is. Do we use `setDisabled` in other places instead of `setEnabled` ? 
```suggestion
  ui->actionSave_Capture->setEnabled(true);
```
I think this is a tiny little bit easier to read.

antonrohr: I just saw `setDisabled` is used often in this context, so feel free to dismiss

OrbitQt/orbitmainwindow.cpp
antonrohr: I am unsure about all possible situations that can happen. For example. 

* I start Orbit, load a module, hook 5 functions. Now I want to save this as preset (before taking a capture) Is this possible?
* I start Orbit, load a capture from file which has hooked functions and a frame track. I want to save this as a preset. Is this possible?

antonrohr (commented):
I looked at the code and it lgtm, but I would like to test this, which I will do after the build issues are resolved

OrbitQt/orbitmainwindow.cpp
florian-kuebler: I actually thought about the same. Shouldn't the disable/enable capture be more related to the "HasCaptureData" Property and the disable/enable preset more to the "Are Functions Hooked"?

ronaldfw: The callbacks for capture data are there (OnCaptureCleared(), etc.), but for presets not. I think it would actually be good to separate out the preset changes to the menu from this PR. For saving presets, CaptureData is not relevant and it work whether that exists or not. WDYT?

OrbitGl/App.h
```diff
@@ -91,8 +91,16 @@ class OrbitApp final : public DataViewFactory, public CaptureListener {
   void StopCapture();
   void AbortCapture();
   void ClearCapture();
+  bool HasCaptureData() const { return capture_data_.has_value(); }
   void SetCaptureData(CaptureData capture_data) { capture_data_ = std::move(capture_data); }
-  [[nodiscard]] const CaptureData& GetCaptureData() const { return capture_data_; }
+  [[nodiscard]] CaptureData& GetCaptureData() {
```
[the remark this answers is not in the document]
florian-kuebler: In that case it is true. But think about just calling a non-const method on the return value directly.

OrbitQt/orbitmainwindow.cpp
florian-kuebler: Sounds good, can you create a bug for the preset change?

florian-kuebler (approved):
Thanks for the changes. Feel free to either address or dismiss the function renaming. Otherwise LGTM! Thanks!

OrbitGl/App.h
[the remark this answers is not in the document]
ronaldfw: Agree there are cases where it's more readable when the name is explicit. Will change it.

OrbitQt/orbitmainwindow.cpp
ronaldfw: b/173679796

OrbitGl/DataManager.h
ronaldfw: Done.

OrbitGl/App.h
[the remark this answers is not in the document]
ronaldfw: Done.

OrbitGl/DataManager.h
```diff
@@ -57,7 +57,7 @@ class DataManager final {
   void set_user_defined_capture_data(const UserDefinedCaptureData& user_defined_capture_data) {
     user_defined_capture_data_ = user_defined_capture_data;
   }
-  [[nodiscard]] const UserDefinedCaptureData& user_defined_capture_data() const {
+  [[nodiscard]] const UserDefinedCaptureData& mutable_user_defined_capture_data() const {
```
florian-kuebler, line 60: Super nit: Following the style guide, the only exception for the PascalCase in function names would be the getter/setter with exactly the name of the field (+`set_`). So I guess `GetMutableUserDefinedCaptureData` would be better here.

ronaldfw: It's not exactly worded like that, if I read the style guide right. I think one can make the case that +mutable_ is fine as well. 

(But I should actually change that in App for GetCaptureData()).

ronaldfw: Actually, no, GetCaptureData() in App is getting capture_data.value(), so it should not follow this rule.

florian-kuebler: Agree.

ronaldfw:
@kisenshi Could you take a look at the changes in ClientGgp?

OrbitClientGgp/ClientGgp.cpp
```diff
@@ -275,8 +275,8 @@ void ClientGgp::OnCaptureStarted(
 
 void ClientGgp::OnCaptureComplete() {
   LOG("Capture completed");
```
kisenshi, line 277: Not related with this PR but just wondering comparing with the code in App if it is missing `GetMutableCaptureData().FilterBrokenCallstacks();` here? I don't think it is a big deal but just wondering if it should be added to the client as well.

kisenshi (approved):
I have checked the updates in ClientGgp, checkout the branch and tested the code and it seems everything is still working as expected. Thanks! 

I left a comment that it is not really related with the PR but I realized when checking/comparing the code with the one in App.

OrbitClientGgp/ClientGgp.cpp
ronaldfw: I would say, yes, it should be added here. Otherwise, we have an inconsistency in captures taken with this client vs the UI client.

ronaldfw: (But would say it's out-of-scope for the PR. Will you take care of adding it?)

antonrohr (approved):
This looks very good! Thank you a lot Ronald, capture data not being nullable was something that bothered me for a while, I am glad its finally resolved. 

I tested a bunch and did not manage to crash Orbit. :)

OrbitClientGgp/ClientGgp.cpp
kisenshi: Yes, I will add it to my todo list. Thanks!

ronaldfw: Thank you!

OrbitQt/orbitmainwindow.cpp
```diff
@@ -642,6 +644,7 @@ void OrbitMainWindow::OpenCapture(const std::string& filepath) {
...
```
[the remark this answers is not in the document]
ronaldfw: Removed, thanks.

ronaldfw:
> This looks very good! Thank you a lot Ronald, capture data not being nullable was something that bothered me for a while, I am glad its finally resolved.
> 
> I tested a bunch and did not manage to crash Orbit. :)

Great, thanks @antonrohr@!

ronaldfw:
Thanks for the reviews, @florian-kuebler, @antonrohr, @kisenshi !

PR #1468, opened 2020-11-18
Add Vulkan Queue Manager for the Vulkan layer.

florian-kuebler:
This will create a new project for Orbit's new vulkan layer.
Its first class is a queue manager. That maintains mappings from
vulkan queues to its corresponding logical device (VkDevice).

Test: Compile & Test
Bug: http://b/168200502


OrbitVulkanLayer/QueueManager.h
```diff
@@ -0,0 +1,37 @@
+// Copyright (c) 2020 The Orbit Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#ifndef ORBIT_VULKAN_LAYER_QUEUE_MANAGER_H_
+#define ORBIT_VULKAN_LAYER_QUEUE_MANAGER_H_
+
+#include "absl/container/flat_hash_map.h"
+#include "absl/container/flat_hash_set.h"
```
beckerhe, line 9: Is that needed?

OrbitVulkanLayer/QueueManager.cpp
```diff
@@ -0,0 +1,22 @@
...
+
+void QueueManager::TrackQueue(const VkQueue& queue, const VkDevice& device) {
+  absl::WriterMutexLock lock(&mutex_);
+  CHECK(!queue_to_device_.contains(queue) || queue_to_device_.at(queue) == device);
+  queue_to_device_[queue] = device;
+}
+
+const VkDevice& QueueManager::GetDeviceOfQueue(const VkQueue& queue) {
+  absl::ReaderMutexLock lock(&mutex_);
+  CHECK(queue_to_device_.contains(queue));
```
beckerhe, line 19: Have you considered returning a std::optional<VkDevice> instead of having a CHECK?

OrbitVulkanLayer/QueueManager.cpp
```diff
@@ -0,0 +1,22 @@
...
+
+#include "QueueManager.h"
+
+#include "OrbitBase/Logging.h"
+
+namespace orbit_vulkan_layer {
+
+void QueueManager::TrackQueue(const VkQueue& queue, const VkDevice& device) {
+  absl::WriterMutexLock lock(&mutex_);
+  CHECK(!queue_to_device_.contains(queue) || queue_to_device_.at(queue) == device);
```
beckerhe, line 13: Have you considered to fail gracefully here? Maybe return a ErrorMessageOr<void>? I'm not sure if that's a good use-case for a CHECK here.

beckerhe (commented):
Thanks Florian. Minor comments!

OrbitVulkanLayer/QueueManager.cpp
```diff
@@ -0,0 +1,22 @@
...
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#include "QueueManager.h"
+
+#include "OrbitBase/Logging.h"
+
+namespace orbit_vulkan_layer {
+
+void QueueManager::TrackQueue(const VkQueue& queue, const VkDevice& device) {
```
beckerhe, line 11: VkQueue and VkDevice seem to be pointers. Does it make sense to pass them by value?

CMakeLists.txt
```diff
@@ -146,6 +146,7 @@ else()
   add_subdirectory(OrbitLinuxTracing)
   add_subdirectory(OrbitService)
   add_subdirectory(OrbitVulkanLayerClientGgp)
+  add_subdirectory(OrbitVulkanLayer)
```
dpallotti, line 149: Keep the alphabetical order and move it up one line?

OrbitVulkanLayer/CMakeLists.txt
```diff
@@ -0,0 +1,38 @@
...
+        ${CMAKE_CURRENT_LIST_DIR}/include)
+
+target_include_directories(OrbitVulkanLayer PRIVATE
+        ${CMAKE_CURRENT_LIST_DIR})
+
+target_sources(OrbitVulkanLayer PRIVATE
+        QueueManager.cpp
+        QueueManager.h)
+
+
```
dpallotti, line 20:
```suggestion
```

OrbitVulkanLayer/CMakeLists.txt
```diff
@@ -0,0 +1,38 @@
...
+target_link_libraries(OrbitVulkanLayer PUBLIC
+        OrbitBase
+        OrbitProtos)
+
+add_executable(OrbitVulkanLayerTests)
+
+target_compile_options(OrbitVulkanLayerTests PRIVATE ${STRICT_COMPILE_FLAGS})
+
+
```
dpallotti, line 29:
```suggestion
```

OrbitVulkanLayer/CMakeLists.txt
```diff
@@ -0,0 +1,38 @@
...
+target_sources(OrbitVulkanLayerTests PRIVATE
+        QueueManagerTest.cpp)
+
+target_link_libraries(
+        OrbitVulkanLayerTests
+        PRIVATE OrbitVulkanLayer
```
dpallotti, line 35:
```suggestion
        OrbitVulkanLayerTests PRIVATE
        OrbitVulkanLayer
```

OrbitVulkanLayer/QueueManager.cpp
```diff
@@ -0,0 +1,22 @@
...
+  absl::WriterMutexLock lock(&mutex_);
+  CHECK(!queue_to_device_.contains(queue) || queue_to_device_.at(queue) == device);
+  queue_to_device_[queue] = device;
```
dpallotti, line 14: I personally would use `queue_to_device_.emplace(queue, device)` to make it clear that you are not actually overwriting (in case `queue_to_device_.at(queue) == device`) (but this is super minor).

OrbitVulkanLayer/QueueManager.h
```diff
@@ -0,0 +1,37 @@
...
+#include "absl/synchronization/mutex.h"
+#include "vulkan/vulkan.h"
+
+namespace orbit_vulkan_layer {
+
+/*
+ * This class tracks the mapping from queues to the device (via vkGetDeviceQueue).
+ *
+ * Note: There is no "untrack" function as there is no such mechanism in vulkan. We could
```
dpallotti, line 18:
```suggestion
 * Note: There is no "untrack" function as there is no such mechanism in Vulkan. We could
```

OrbitVulkanLayer/QueueManager.cpp
florian-kuebler: In the layer we won't be able to handle errors anyways, thus I'd rather keep my expectations on the calls explicit.

OrbitVulkanLayer/QueueManager.cpp
dpallotti: This will occur in so many places in the Vulkan layer, let's think about it well now.

florian-kuebler: Right they are typedefs of `VkQueue_T*` and `VkDevice_T*`. I thought of passing them as values, but then always find this confusing to see objects passed by value, where you not directly see that they are actually just pointers. I am not too strong about this though.

OrbitVulkanLayer/QueueManager.cpp
florian-kuebler: As for the other CHECK, I'd really like to use it to encode certain expectations on that call. The logic should never call this method, when the queue is not there, so I wanted to encode this here.

OrbitVulkanLayer/QueueManagerTest.cpp
```diff
@@ -0,0 +1,33 @@
+// Copyright (c) 2020 The Orbit Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#include "QueueManager.h"
+#include "gtest/gtest.h"
+
+#define UNUSED(x) (void)(x)
```
dpallotti, line 8: Not necessarily a fan of having the macro, I think that using `(void)` directly is a known pattern. You can probably (I think?) also skip the declaration of the variable in the case (if you want), like `(void)manager.GetDeviceOfQueue(queue)`.

OrbitVulkanLayer/QueueManager.h
florian-kuebler: No, thanks!

dpallotti (approved):
Looks good! (All the comments are really really minor.)

OrbitVulkanLayer/CMakeLists.txt
florian-kuebler: Done.

OrbitVulkanLayer/CMakeLists.txt
florian-kuebler: Done.

OrbitVulkanLayer/QueueManager.cpp
florian-kuebler: Done, thanks!

OrbitVulkanLayer/QueueManager.h
florian-kuebler: Done.

OrbitVulkanLayer/QueueManagerTest.cpp
florian-kuebler: Ah nice, thanks!

beckerhe (approved):
Issues were discussed OOB.

OrbitVulkanLayer/QueueManager.cpp
florian-kuebler: As discussed offline, I went with passing them by value.

PR #1469, opened 2020-11-18
Test program that spawns and ends threads.

danielfenner:
Helpful to test the behaviour when attaching to a process that races with us by creating tasks.

### DIFF
--- a/OrbitTest/main_short_lived_threads.cpp
+++ b/OrbitTest/main_short_lived_threads.cpp
@@ -34,7 +34,7 @@ void Worker(int ttl_ms) {
 // make things slightly more interesting.
 int main(int argc, char* argv[]) {
   if (argc != 3) {
-    std::cerr << "Usage:\n./OrbitTestShortLivedThreads number_of_threads time_to_live_per_thread\n";
+    std::cerr << "Usage:\n./OrbitTestShortLivedThreads number_of_threads time_to_live_per_thread_ms\n";
     return -1;
   }
 
--- a/OrbitTest/main_short_lived_threads.cpp
+++ b/OrbitTest/main_short_lived_threads.cpp
@@ -2,16 +2,14 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
 
-#include <string>
 #include <chrono>
 #include <iostream>
+#include <mutex>
+#include <random>
+#include <set>
+#include <string>
 #include <thread>
 #include <vector>
-#include <mutex>
-#include <set>
-#include <random>
-
-uint32_t g_thread_count{0};
 
 std::mutex g_joinable_mutex;
 std::set<std::thread::id> g_joinable_threads;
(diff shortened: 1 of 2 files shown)


OrbitTest/main_short_lived_threads.cpp
```diff
@@ -0,0 +1,72 @@
...
+
+// Main thread goes into a busy loop that spawns and joins threads.
+// The spawned threads go into a busy wait.
+// The first command line parameter gives the number of threads to spawn.
+// The second command line parameter gives the live time of each spawned thread.
+// The actual live time is varied between 100% and 200% of that value to
+// make things slightly more interesting.
+int main(int argc, char* argv[]) {
+  if (argc != 3) {
+    std::cerr << "Usage:\n./OrbitTestShortLivedThreads number_of_threads time_to_live_per_thread\n";
```
pierricgimmig, line 37: Maybe specify the time unit: time_to_live_per_thread_ms?

OrbitTest/main_short_lived_threads.cpp
```diff
@@ -0,0 +1,72 @@
...
+// Busy wait for ttl_ms.
+void Worker(int ttl_ms) {
+  auto deadline = std::chrono::system_clock::now() + std::chrono::milliseconds(ttl_ms);
+  while (true) {
+    if (std::chrono::system_clock::now() > deadline) {
+      break;
+    }
+  }
+  std::lock_guard lock(g_joinable_mutex);
+  g_joinable_threads.emplace(std::this_thread::get_id());
```
pierricgimmig, line 26: Is the g_joinable_threads set really needed? We could simply create all threads and then join them all, no?

danielfenner: I'd like to keep _number_of_threads_ threads running all the time to keep the system under stress. So in every round in the main loop I just want to join the threads that already finished their "work".

OrbitTest/main_short_lived_threads.cpp
danielfenner: yep, done.

OrbitTest/main_short_lived_threads.cpp
```diff
@@ -0,0 +1,74 @@
...
+}
+
+// Main thread goes into a busy loop that spawns and joins threads.
+// The spawned threads perform a busy wait.
+// The first command line parameter gives the number of threads to spawn. When
+// a thread finishs it will be replaced by a new one.
```
pierricgimmig, line 32: nit: finishes

OrbitTest/main_short_lived_threads.cpp
pierricgimmig: Ah ok ok. And the lifecycle of threads is what you are interested in? A comment on how this program will be used to add more context could be useful.

pierricgimmig (approved):

OrbitTest/main_short_lived_threads.cpp
danielfenner: done

OrbitTest/main_short_lived_threads.cpp
danielfenner: Yes, I'd like to provoke errors in places where we need to keep track of the set of threads in the target process.

Added a few lines of comments.

PR #1471, opened 2020-11-19
Merge release/1.55 into master

karupayun:


beckerhe:
Hi Pablo,
there are two additional commits. That looks a bit fishy. Is that intended?

karupayun:
Updated with a force push, because we've erased some PR from 1.55 release.

karupayun:
> Hi Pablo,
> there are two additional commits. That looks a bit fishy. Is that intended?

In a GVC with Henning we agreed that is intended

beckerhe:
> Hi Pablo,
> there are two additional commits. That looks a bit fishy. Is that intended?

For reference: There was a problem with pierric's commit. But having additional commits in here is normal and there was nothing wrong in that regard with the PR.

Looks good!

beckerhe (approved):
LGTM! Thanks Pablo!

IrinaShkviro (approved):

PR #1473, opened 2020-11-19
Put Mock Class in an Anonymous Namespace

florian-kuebler:
This allows other tests also define a `MockDispatchTable`.

Test: Compile & Run Tests

### DIFF
--- a/OrbitVulkanLayer/DeviceManagerTest.cpp
+++ b/OrbitVulkanLayer/DeviceManagerTest.cpp
@@ -18,7 +18,7 @@ class MockDispatchTable {
               (VkPhysicalDevice dispatchable_object));
 };
 
-}  //  namespace
+}  // namespace
 
 TEST(DeviceManager, AnUntrackedDeviceCannotBeQueried) {
   MockDispatchTable dispatch_table;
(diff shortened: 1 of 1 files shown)


OrbitVulkanLayer/DeviceManagerTest.cpp
```diff
@@ -10,12 +10,16 @@ using ::testing::Return;
...
 
+namespace {
+
 class MockDispatchTable {
  public:
   MOCK_METHOD(PFN_vkGetPhysicalDeviceProperties, GetPhysicalDeviceProperties,
               (VkPhysicalDevice dispatchable_object));
 };
 
+}  //  namespace
```
dpallotti, line 21:
```suggestion
}  // namespace
```

(sorry :) )

dpallotti (approved):

OrbitVulkanLayer/DeviceManagerTest.cpp
dpallotti: My OCD is strong.

florian-kuebler: Done :)

PR #1474, opened 2020-11-19
Move UserDefinedCaptureData from CaptureData to DataManager only

ronaldfw:
The recent refactoring of CaptureData to using std::optional and the
need of adding frame tracks before a capture makes it clear that
UserDefinedCaptureData (which owns frame track data) should not be a
member of CaptureData (and indeed does not have to be a member).

This is a mostly mechanical change that removes the
UserDefinedCaptureData member from CaptureData.

Tested: Ran unit tests; manual load/save/capture tests with and without
presets.

Bug: b/173191435

### DIFF
--- a/OrbitClientGgp/ClientGgp.cpp
+++ b/OrbitClientGgp/ClientGgp.cpp
@@ -82,7 +82,7 @@ bool ClientGgp::RequestStartCapture(ThreadPool* thread_pool) {
   bool enable_introspection = false;
   ErrorMessageOr<void> result = capture_client_->StartCapture(
       thread_pool, target_process_, module_manager_, selected_functions_, selected_tracepoints,
-      UserDefinedCaptureData(), enable_introspection);
+      UserDefinedCaptureData{}, enable_introspection);
 
   if (result.has_error()) {
     ERROR("Error starting capture: %s", result.error().message());
@@ -109,8 +109,9 @@ bool ClientGgp::SaveCapture() {
   // Add the location where the capture is saved
   file_name.insert(0, options_.capture_file_directory);
 
-  ErrorMessageOr<void> result = capture_serializer::Save(
-      file_name, GetCaptureData(), key_to_string_map, timer_infos_.begin(), timer_infos_.end());
+  ErrorMessageOr<void> result =
+      capture_serializer::Save(file_name, GetCaptureData(), UserDefinedCaptureData{},
+                               key_to_string_map, timer_infos_.begin(), timer_infos_.end());
   if (result.has_error()) {
     ERROR("Could not save the capture: %s", result.error().message());
     return false;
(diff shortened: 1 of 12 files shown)


kisenshi (approved):
I checked the modifications in _ClientGgp_ and tested it. Everything seems to be working as expected still. Thanks!

ronaldfw:
Thanks for checking @kisenshi. I need to check some things here first before continuing, I think I actually may need to also store the user defined data in CaptureData.

OrbitClientGgp/ClientGgp.cpp
```diff
@@ -82,7 +82,7 @@ bool ClientGgp::RequestStartCapture(ThreadPool* thread_pool) {
   bool enable_introspection = false;
   ErrorMessageOr<void> result = capture_client_->StartCapture(
       thread_pool, target_process_, module_manager_, selected_functions_, selected_tracepoints,
-      UserDefinedCaptureData(), enable_introspection);
+      UserDefinedCaptureData{}, enable_introspection);
```
florian-kuebler, line 85: Couldn't it also just be `...selected_tracepoints, {}, enable_introspection);`?

OrbitGl/App.cpp
```diff
@@ -138,12 +138,12 @@ void OrbitApp::OnCaptureStarted(ProcessData&& process,
...
         // this task is completely executed.
-        capture_data_ =
-            CaptureData(std::move(process), module_manager_.get(), std::move(selected_functions),
-                        std::move(selected_tracepoints), std::move(user_defined_capture_data));
+        capture_data_ = CaptureData(std::move(process), module_manager_.get(),
+                                    std::move(selected_functions), std::move(selected_tracepoints));
 
-        frame_track_online_processor_ =
-            FrameTrackOnlineProcessor(GetCaptureData(), GCurrentTimeGraph);
+        data_manager_->set_user_defined_capture_data(std::move(user_defined_capture_data));
```
florian-kuebler, line 144: It feels a bit strange that we set the user defined capture data here, which will be the same value as it is right now in data_manager.
But probably it is fine for now.

florian-kuebler (approved):
LGTM. Thanks!

ronaldfw:
Thanks @florian-kuebler for the review. Sorry for the back and forth here, but I think I may need to abandon this change. As you pointed out on chat, there is a difference between the selected functions in DataManager and CaptureData (in particular, they don't have to be the same set), and the same is true for frame tracks (as they are tied to selected functions). 

So, I think the right model is to have DataManager to hold the frame track functions, but CaptureData must get a copy of that when taking a capture. This is the same as for selected functions. However, for frame tracks it is (and should be) possible for users to add frame tracks on hooked functions even after taking a capture. This must be reflected both the DataManager and CaptureData.

florian-kuebler:
No worries. We can also set up something "offline", maybe together with @antonrohr?

PR #1476, opened 2020-11-20
Refactor TextRenderer

pierricgimmig:
- Font size is now only passed as argument to the AddText methods
- GetStringWidth now takes font size as input
- Add GetStringHeight method
- Revive DrawOutline for character debug rendering
- Rename variables

There were strange bugs caused by the fact that the text renderer
kept the last font size passed by a call to AddText as the current
font size to be used by other methods like GetStringWidth.

### DIFF
--- a/OrbitGl/CaptureWindow.cpp
+++ b/OrbitGl/CaptureWindow.cpp
@@ -498,7 +498,7 @@ void CaptureWindow::Draw() {
   Append(all_layers, text_renderer_.GetLayers());
   Append(all_layers, time_graph_.GetTextRenderer()->GetLayers());
 
-  // Remove duplicates.
+  // Sort and remove duplicates.
   std::sort(all_layers.begin(), all_layers.end());
   auto it = std::unique(all_layers.begin(), all_layers.end());
   all_layers.resize(std::distance(all_layers.begin(), it));
--- a/OrbitGl/CaptureWindow.cpp
+++ b/OrbitGl/CaptureWindow.cpp
@@ -402,11 +402,6 @@ void CaptureWindow::KeyPressed(unsigned int key_code, bool ctrl, bool shift, boo
       case 'X':
         ToggleRecording();
         break;
-      case 'O':
-        if (ctrl) {
-          text_renderer_.ToggleDrawOutline();
-        }
-        break;
       case 18:  // Left
         if (shift) {
           time_graph_.JumpToNeighborBox(GOrbitApp->selected_text_box(),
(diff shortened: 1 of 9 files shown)


OrbitGl/CaptureWindow.cpp
```diff
@@ -734,11 +743,11 @@ void CaptureWindow::RenderHelpUi() {
       "Measure: \"Right Click + Drag\"\n"
       "Toggle Help: 'H'";
 
-  const uint32_t kHelpMessageFontSize = 2 * font_size_;
   Vec2 text_bounding_box_pos;
   Vec2 text_bounding_box_size;
-  text_renderer_.AddText(help_message, world_x, world_y, GlCanvas::kZValueTextUi,
-                         Color(255, 255, 255, 255), kHelpMessageFontSize, -1.f /*max_size*/,
+  float string_height = text_renderer_.GetStringHeight(help_message, font_size_);
+  text_renderer_.AddText(help_message, world_x, world_y - string_height, GlCanvas::kZValueTextUi,
```
karupayun, line 749: To avoid this world_y - string_height calculation and getting the height before, a function like AddTextGivenTopLimit or something like that, don't you like?

karupayun (approved):
LGTM! :)

Thanks Pierric for solving the Track label height issue :)

I have one question:
1- When I click "Draw Text Outline" in Linux, i see this behavior. Is this expected?
Draw Text Outline

OrbitGl/CaptureWindow.cpp
pierricgimmig: Yes good idea, I'd add a parameter to specify the origin of the text (kBaseLine, kTopLine, kBottomLine)

pierricgimmig: Will do in subsequent pr.

OrbitGl/CaptureWindow.cpp
```diff
@@ -497,11 +492,17 @@ void CaptureWindow::Draw() {
...
+    text_renderer_.RenderDebug(&ui_batcher_);
+  }
+
+  // Get all layers.
   std::vector<float> all_layers = time_graph_.GetBatcher().GetLayers();
   Append(all_layers, ui_batcher_.GetLayers());
   Append(all_layers, text_renderer_.GetLayers());
   Append(all_layers, time_graph_.GetTextRenderer()->GetLayers());
+
+  // Remove duplicates.
```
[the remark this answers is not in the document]
pierricgimmig: Done.

PR #1477, opened 2020-11-20
Add option for enabling a frame track for a function to Functions tab

ronaldfw:
So far it was only possible to add frame tracks after taking a capture.
With this change, it is now possible to enable a frame track (or
multiple) even before the first capture. This is possible from the
Functions tab by choosing the context menu option
"Enable frame track(s)".

Note that a function that is chosen as a frame track must be hooked, so
if a function is not hooked when enabling it as a frame track, it is
automatically hooked.

Bug: http://b/173191435

Tested: Lots of manual tests with taking captures, loading/saving
presets, loading/saving captures.

### DIFF
--- a/OrbitCore/UserDefinedCaptureData.cpp
+++ b/OrbitCore/UserDefinedCaptureData.cpp
@@ -7,13 +7,11 @@
 #include "OrbitClientData/FunctionUtils.h"
 
 void UserDefinedCaptureData::InsertFrameTrack(const orbit_client_protos::FunctionInfo& function) {
-  // We do not allow insertion of frame tracks twice.
-  CHECK(frame_track_functions_.insert(function).second);
+  frame_track_functions_.insert(function);
 }
 
 void UserDefinedCaptureData::EraseFrameTrack(const orbit_client_protos::FunctionInfo& function) {
-  // We do not allow erasing a frame track that does not exist.
-  CHECK(frame_track_functions_.erase(function));
+  frame_track_functions_.erase(function);
 }
 
 [[nodiscard]] bool UserDefinedCaptureData::ContainsFrameTrack(
--- a/OrbitGl/App.cpp
+++ b/OrbitGl/App.cpp
@@ -58,6 +58,7 @@ ABSL_DECLARE_FLAG(bool, track_ordering_feature);
 
 using orbit_client_protos::CallstackEvent;
 using orbit_client_protos::FunctionInfo;
+using orbit_client_protos::FunctionStats;
 using orbit_client_protos::LinuxAddressInfo;
 using orbit_client_protos::PresetFile;
 using orbit_client_protos::PresetInfo;
@@ -1108,7 +1109,7 @@ void OrbitApp::LoadSymbols(const std::filesystem::path& symbols_path, ModuleData
 
       if (!frame_track_function_hashes.empty()) {
         const auto frame_track_result =
-            InsertFrameTracksFromHashes(module_data, frame_track_function_hashes);
+            EnableFrameTracksFromHashes(module_data, frame_track_function_hashes);
         if (!frame_track_result) {
           LOG("Warning, could not insert frame tracks: %s", frame_track_result.error().message());
         }
@@ -1162,12 +1163,12 @@ ErrorMessageOr<void> OrbitApp::SelectFunctionsFromHashes(
   return error;
 }
 
-ErrorMessageOr<void> OrbitApp::InsertFrameTracksFromHashes(
+ErrorMessageOr<void> OrbitApp::EnableFrameTracksFromHashes(
     const ModuleData* module, const std::vector<uint64_t> function_hashes) {
   std::vector<const FunctionInfo*> function_infos;
   const auto& error = GetFunctionInfosFromHashes(module, function_hashes, &function_infos);
   for (const auto* function : function_infos) {
-    data_manager_->user_defined_capture_data().InsertFrameTrack(*function);
+    data_manager_->EnableFrameTrack(*function);
   }
   return error;
 }
@@ -1191,7 +1192,7 @@ void OrbitApp::LoadPreset(const std::shared_ptr<PresetFile>& preset_file) {
       }
       std::vector<uint64_t> frame_track_hashes{preset_module.frame_track_function_hashes().begin(),
                                                preset_module.frame_track_function_hashes().end()};
-      const auto frame_track_result = InsertFrameTracksFromHashes(module_data, frame_track_hashes);
+      const auto frame_track_result = EnableFrameTracksFromHashes(module_data, frame_track_hashes);
       if (!frame_track_result) {
         LOG("Warning: %s", frame_track_result.error().message());
       }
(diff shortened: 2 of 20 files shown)


OrbitGl/App.cpp
```diff
@@ -1523,22 +1524,36 @@ void OrbitApp::DeselectTracepoint(const TracepointInfo& tracepoint) {
...
+  data_manager_->InsertFrameTrack(function);
+
+  if (!HasCaptureData()) {
+    return;
+  }
+  // We only add a frame track to the actual capture data if the function for the frame
+  // track actually has hits in the capture data. Otherwise we can end up in inconsistent
+  // states where "empty" frame tracks exist in the capture data (which would also be
+  // serialized).
+  const FunctionStats& stats = GetCaptureData().GetFunctionStatsOrDefault(function);
```
florian-kuebler, line 1536: That makes very much sense. Maybe we should add a pop-up here, explaining this to the user (by here, I really mean after the !HasCaptureData.

OrbitGl/App.cpp
```diff
@@ -1523,22 +1524,36 @@ void OrbitApp::DeselectTracepoint(const TracepointInfo& tracepoint) {
...
-  data_manager_->set_user_defined_capture_data(GetCaptureData().user_defined_capture_data());
-  GCurrentTimeGraph->RemoveFrameTrack(function);
+  data_manager_->EraseFrameTrack(function);
+  if (HasCaptureData() && GetCaptureData().ContainsFrameTrack(function)) {
+    GetMutableCaptureData().EraseFrameTrack(function);
+    GCurrentTimeGraph->RemoveFrameTrack(function);
+  }
 }
 
 bool OrbitApp::HasFrameTrack(const FunctionInfo& function) const {
```
florian-kuebler, line 1551: Where is that function called from? Just want to double check, that it will return false e.g. if there was not already a capture.

florian-kuebler (approved):
LGTM, thanks! Let me know what you think about the pop-up.

OrbitGl/FunctionsDataView.h
```diff
@@ -49,6 +51,8 @@ class FunctionsDataView : public DataView {
 
   static const std::string kMenuActionSelect;
   static const std::string kMenuActionUnselect;
+  static const std::string kMenuActionFrameTrack;
```
antonrohr, line 54: nit maybe:
```suggestion
  static const std::string kMenuActionAddFrameTrack;
```

antonrohr: Or Enable / Disable

antonrohr (changes requested):
The code looks nice, only one nitpick, but I encountered a crash while testing. (Might actually be unrelated to this PR, but I think this could be fixed here)

1. Connect to instance.
2. load module libvulkan.so
3. Filter functionslist for : vkQueueSubmit
4. Right click enable frametrack (on vkQueueSubmit)
5. Right click unhook (on vkQueueSubmit) -> Now this is in a state where its not hooked, but has a frametrack
6. Start Capture
7. Stop Capture
8. Crash

OrbitGl/App.cpp
ronaldfw: Added pop-up that has checkbox option to not show again.

ronaldfw:
> The code looks nice, only one nitpick, but I encountered a crash while testing. (Might actually be unrelated to this PR, but I think this could be fixed here)
> 
> 1. Connect to instance.
> 2. load module libvulkan.so
> 3. Filter functionslist for : vkQueueSubmit
> 4. Right click enable frametrack (on vkQueueSubmit)
> 5. Right click unhook (on vkQueueSubmit) -> Now this is in a state where its not hooked, but has a frametrack
> 6. Start Capture
> 7. Stop Capture
> 8. Crash

This should be fixed now. Unhooking a function now also disables its frame track.

OrbitGl/App.cpp
ronaldfw: Code changed quite a bit, could you take a look again?

OrbitGl/FunctionsDataView.h
ronaldfw: Done.

OrbitGl/App.cpp
```diff
@@ -1522,23 +1523,55 @@ void OrbitApp::DeselectTracepoint(const TracepointInfo& tracepoint) {
...
+  // track actually has hits in the capture data. Otherwise we can end up in inconsistent
+  // states where "empty" frame tracks exist in the capture data (which would also be
+  // serialized).
+  const FunctionStats& stats = GetCaptureData().GetFunctionStatsOrDefault(function);
+  if (stats.count() > 1) {
+    GetMutableCaptureData().EnableFrameTrack(function);
+    AddFrameTrackTimers(function);
+  } else {
+    CHECK(empty_frame_frame_warning_callback_);
+    empty_frame_frame_warning_callback_();
```
florian-kuebler, line 1552: -> `empty_frame_track_warning_callback_`

OrbitGl/LiveFunctionsDataView.cpp
```diff
@@ -265,20 +264,17 @@ void LiveFunctionsDataView::OnContextMenu(const std::string& action, int menu_in
...
     }
-  } else if (action == kMenuActionFrameTrack) {
+  } else if (action == kMenuActionEnableFrameTrack) {
     for (int i : item_indices) {
       FunctionInfo* function = GetSelectedFunction(i);
-      const FunctionStats& stats = capture_data.GetFunctionStatsOrDefault(*function);
-      if (stats.count() > 1 && !GOrbitApp->HasFrameTrack(*function)) {
-        live_functions_->AddFrameTrack(*function);
-      }
+      GOrbitApp->EnableFrameTrack(*function);
```
florian-kuebler, line 270: What happens to the next capture, if we unselect a function for the next capture, but add the frame track afterwards. The frame track will not appear in the next capture, right?

OrbitQt/orbitmainwindow.cpp
```diff
@@ -607,6 +610,28 @@ void OrbitMainWindow::ShowCaptureOnSaveWarningIfNeeded() {
   }
 }
 
+void OrbitMainWindow::ShowEmptyFrameTrackWarningIfNeeded() {
+  QSettings settings("The Orbit Authors", "Orbit Profiler");
+  const QString empty_frame_track_warning("EmptyFrameTrackWarning");
+  if (!settings.value(empty_frame_track_warning, false).toBool()) {
+    QMessageBox message_box;
+    message_box.setText(
+        "Frame track enabled for the selected function, but since "
```
florian-kuebler, line 619: Do we allow adding multiple frame tracks at once? In that case the warning could be more concrete.

florian-kuebler (commented):
Thanks for the update. Feels better now. I'd like to hear your opinion on the auto-selection comment. Otherwise looks good.

OrbitGl/App.cpp
ronaldfw: Thanks for spotting. Fixed.

OrbitGl/App.cpp
```diff
@@ -1522,23 +1523,55 @@ void OrbitApp::DeselectTracepoint(const TracepointInfo& tracepoint) {
...
+  }
 }
 
 void OrbitApp::RemoveFrameTrack(const FunctionInfo& function) {
-  GetMutableCaptureData().EraseFrameTrack(function);
-  data_manager_->set_user_defined_capture_data(GetCaptureData().user_defined_capture_data());
-  GCurrentTimeGraph->RemoveFrameTrack(function);
+  CHECK(!IsFrameTrackEnabled(function));
+
+  if (HasCaptureData() && GetCaptureData().IsFrameTrackEnabled(function)) {
```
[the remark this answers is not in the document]
ronaldfw: Done.

OrbitGl/FunctionsDataView.cpp
```diff
@@ -156,6 +178,27 @@ void FunctionsDataView::OnContextMenu(const std::string& action, int menu_index,
...
+      GOrbitApp->SelectFunction(function);
+      GOrbitApp->EnableFrameTrack(function);
+      GOrbitApp->AddFrameTrack(function);
+    }
+  } else if (action == kMenuActionDisableFrameTrack) {
+    for (int i : item_indices) {
+      // When we remove a frame track, we do not unhook (deselect) the function as
+      // it may have been selected manually (not as part of adding a frame track).
+      // However, disable the frame track, so it is not recreated on the next capture.
+      GOrbitApp->DisableFrameTrack(*GetFunction(i));
```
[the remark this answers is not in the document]
ronaldfw: What you describe (keep frame track hooked functions separate) sounds like quite a large change. 

I think it's fairly visible to users due to the icons (checkmark and "F") that a function gets hooked when selecting a frame track, so I'm not too concerned about this.

OrbitGl/LiveFunctionsDataView.cpp
ronaldfw: This is actually a mistake. When the function is not hooked, it should get hooked here.

ronaldfw: Fixed.

OrbitQt/orbitmainwindow.cpp
ronaldfw: You mean adding the function name(s) to the message?

OrbitGl/TimeGraph.cpp
```diff
@@ -1275,6 +1275,10 @@ bool TimeGraph::IsVisible(VisibilityType vis_type, uint64_t min, uint64_t max) c
   }
 }
 
+bool TimeGraph::HasFrameTrack(const orbit_client_protos::FunctionInfo& function) const {
+  return frame_tracks_.count(function.address()) > 0;
```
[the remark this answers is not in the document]
ronaldfw: Not sure anymore. I can fix that in a follow-up PR.

OrbitGl/FunctionsDataView.cpp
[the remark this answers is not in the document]
antonrohr: I like the idea that Frametracks are completely independent from hooked functions, because I think it would make it more intuitive. The user does not need to know how frametracks work under the hood. 

But I also agree with Ronald, that it is completely fine how it is right now. Changing that would require a lot of work with rather little benefit.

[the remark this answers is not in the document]
florian-kuebler: Agree. Just wanted to hear your ideas about that. That might be something for a future fixit week?
For now it should be fine.

OrbitGl/TimeGraph.cpp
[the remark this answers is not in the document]
florian-kuebler: Sounds good. Thanks!

OrbitQt/orbitmainwindow.cpp
florian-kuebler: For example. I am not too strong on this. If you think it would be fine as it is, also feel free to just resolve the comment.

OrbitCore/UserDefinedCaptureData.cpp
```diff
@@ -7,13 +7,11 @@
 #include "OrbitClientData/FunctionUtils.h"
 
 void UserDefinedCaptureData::InsertFrameTrack(const orbit_client_protos::FunctionInfo& function) {
-  // We do not allow insertion of frame tracks twice.
-  CHECK(frame_track_functions_.insert(function).second);
+  frame_track_functions_.insert(function);
```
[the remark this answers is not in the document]
ronaldfw: It simplifies the logic a bit and actually doesn't hurt. 

There are now cases where you can end up with frame tracks still in the capture data but the function is unhooked. Then re-renabling the function as a frame track can cause the frame track function to be inserted again in the capture data. I could work around that by checking whether it's already in there, but that defeats the purpose. If I have a check whether it's in there when I insert it, then I just circumvent the CHECK. It's cleaner to just make sure the right calls to insert / erase get made independent of the current state.

OrbitQt/orbitmainwindow.cpp
antonrohr: Currently when I select 3 functions in the live tab that have never been called, and click on Enable Frame Track, then I get 3 times the same error message.

antonrohr (approved):
lgtm, thanks for this change :)

OrbitQt/orbitmainwindow.cpp
ronaldfw: What do you think about the following change: Add the function name where adding a frame track didn't work and show that in the message. Users can still use the checkbox to not get the second, etc. message.

Showing a single aggregated dialogue requires a bigger logic change.

ronaldfw: Note that this should be a super-rare error message anyway, as it's not useful to use functions that are not called at a more or less regular frequency.

OrbitGl/FunctionsDataView.cpp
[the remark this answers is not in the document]
ronaldfw: Sounds good, will resolve this for now.

OrbitQt/orbitmainwindow.cpp
ronaldfw: I've added the function name to the message. The message will still pop up multiple times for each function that has 0 hits.

ronaldfw:
@antonrohr @florian-kuebler Let me know if you have additional comments or concerns, o/w I'll merge this soon.

OrbitCore/UserDefinedCaptureData.cpp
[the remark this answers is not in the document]
florian-kuebler: Okay, thanks for the explanation.

OrbitQt/orbitmainwindow.cpp
florian-kuebler: Thanks. Sounds good to me.

florian-kuebler:
From my side, this is good to go. Thanks!

PR #1478, opened 2020-11-20
Add IntrospectionWindow class

pierricgimmig:
The IntrospectionWindow is a CaptureWindow used to show a second
time graph dedicated to introspection of the client side Orbit app.
This change only adds the new class, a subsequent change will
enable introspection via some new UI in the main CaptureWindow.

Also add a few profiling markers in our code.

### DIFF
--- a/OrbitGl/App.cpp
+++ b/OrbitGl/App.cpp
@@ -509,13 +509,13 @@ void OrbitApp::MainTick() {
   }
 }
 
-void OrbitApp::RegisterCaptureWindow(CaptureWindow* capture) {
+void OrbitApp::SetCaptureWindow(CaptureWindow* capture) {
   CHECK(capture_window_ == nullptr);
   GCurrentTimeGraph = capture->GetTimeGraph();
   capture_window_ = capture;
 }
 
-void OrbitApp::RegisterDebugCanvas(GlCanvas* debug_canvas) {
+void OrbitApp::SetDebugCanvas(GlCanvas* debug_canvas) {
   CHECK(debug_canvas_ == nullptr);
   debug_canvas_ = debug_canvas;
   debug_canvas_->EnableImGui();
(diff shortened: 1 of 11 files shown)


OrbitGl/App.cpp
```diff
@@ -503,6 +511,7 @@ void OrbitApp::MainTick() {
 
 void OrbitApp::RegisterCaptureWindow(CaptureWindow* capture) {
```
reichlfl, line 512: Nit: You obviously didn't introduce this, but I find the function names `RegisteCaptureWindow` (and the other `RegisterXXX` below) confusing - to me, this signals you can register more than one capture window here, when in fact the CHECK at the beginning demands the exact opposite.

Would just calling it `SetCaptureWindow` be clearer?

OrbitGl/IntrospectionWindow.cpp
```diff
@@ -0,0 +1,59 @@
...
+        timer_info.add_registers(scope.encoded_event.args[2]);
+        timer_info.add_registers(scope.encoded_event.args[3]);
+        timer_info.add_registers(scope.encoded_event.args[4]);
+        timer_info.add_registers(scope.encoded_event.args[5]);
+        time_graph_.ProcessTimer(timer_info, /*FunctionInfo*/ nullptr);
+      });
+}
+void IntrospectionWindow::StopIntrospection() { introspection_listener_ = nullptr; }
+
+void IntrospectionWindow::ToggleCapture() {
```
reichlfl, line 51: Another naming nitpick: I can see us running into issues with the word "Capture". Before, it was clear what this is referring to since we were only profiling one process. Now, "Capture" is used to refer to the recording taken of the target process (in contrast to the introspection), so `ToggleCapture` doesn't make a lot of sense here.

I'd suggest to rename this method to `ToggleRecording` or the like.

reichlfl (approved):
Small nitpicks mostly about naming, but overall LGTM. Really nice, I'm pleasantly suprised that we can actually do this with few and clean code changes :-)

OrbitGl/App.cpp
pierricgimmig: Totally agree, thanks, done.

OrbitGl/IntrospectionWindow.cpp
pierricgimmig: Totally agree, thanks, done.

PR #1480, opened 2020-11-21
Client side introspection UI

pierricgimmig:
Activate client side introspection. In "--devmode", a new button in the 
capture view toolbar appears. Pressing it will instantiate an always-on-top 
IntrospectionWindow that can be used to profile Orbit inside of Orbit.
To start and stop a capture, press 'X' just like in the normal cpature window.
When the IntrospectionWindow is closed, introspection automatically stops.
You can add profiling markers in the Orbit code base by using the manual 
instrumentation API in Orbit.h.

### DIFF
--- a/OrbitGl/CaptureWindow.h
+++ b/OrbitGl/CaptureWindow.h
@@ -63,7 +63,7 @@ class CaptureWindow : public GlCanvas {
   void ToggleDrawHelp();
   void set_draw_help(bool draw_help);
   [[nodiscard]] TimeGraph* GetTimeGraph() { return &time_graph_; }
-  [[nodiscard]] CaptureData* GetCaptureData() const { return time_graph_.GetCaptureData(); }
+  [[nodiscard]] const CaptureData* GetCaptureData() const { return time_graph_.GetCaptureData(); }
   void SetCaptureData(CaptureData* capture_data) { time_graph_.SetCaptureData(capture_data); }
 
   Batcher& GetBatcherById(BatcherId batcher_id);
--- a/OrbitGl/CaptureWindow.h
+++ b/OrbitGl/CaptureWindow.h
@@ -63,8 +63,6 @@ class CaptureWindow : public GlCanvas {
   void ToggleDrawHelp();
   void set_draw_help(bool draw_help);
   [[nodiscard]] TimeGraph* GetTimeGraph() { return &time_graph_; }
-  [[nodiscard]] const CaptureData* GetCaptureData() const { return time_graph_.GetCaptureData(); }
-  void SetCaptureData(CaptureData* capture_data) { time_graph_.SetCaptureData(capture_data); }
 
   Batcher& GetBatcherById(BatcherId batcher_id);
 
--- a/OrbitGl/CaptureWindow.h
+++ b/OrbitGl/CaptureWindow.h
@@ -63,6 +63,8 @@ class CaptureWindow : public GlCanvas {
   void ToggleDrawHelp();
   void set_draw_help(bool draw_help);
   [[nodiscard]] TimeGraph* GetTimeGraph() { return &time_graph_; }
+  [[nodiscard]] CaptureData* GetCaptureData() const { return time_graph_.GetCaptureData(); }
+  void SetCaptureData(CaptureData* capture_data) { time_graph_.SetCaptureData(capture_data); }
 
   Batcher& GetBatcherById(BatcherId batcher_id);
 
--- a/OrbitGl/EventTrack.cpp
+++ b/OrbitGl/EventTrack.cpp
@@ -161,6 +161,7 @@ void EventTrack::SelectEvents() {
 }
 
 bool EventTrack::IsEmpty() const {
+  if (!GOrbitApp->HasCaptureData()) return true;
   const uint32_t callstack_count =
       (thread_id_ == SamplingProfiler::kAllThreadsFakeTid)
           ? GOrbitApp->GetCaptureData().GetCallstackData()->GetCallstackEventsCount()
--- a/OrbitGl/EventTrack.cpp
+++ b/OrbitGl/EventTrack.cpp
@@ -75,6 +75,8 @@ void EventTrack::UpdatePrimitives(uint64_t min_tick, uint64_t max_tick, PickingM
 
   const Color kWhite(255, 255, 255, 255);
   const Color kGreenSelection(0, 255, 0, 255);
+  const CaptureData* capture_data = time_graph_->GetCaptureData();
+  CHECK(capture_data);
 
   if (!picking) {
     // Sampling Events
@@ -85,12 +87,12 @@ void EventTrack::UpdatePrimitives(uint64_t min_tick, uint64_t max_tick, PickingM
         batcher->AddVerticalLine(pos, -track_height, z, kWhite);
       }
     };
+
     if (thread_id_ == SamplingProfiler::kAllThreadsFakeTid) {
-      GOrbitApp->GetCaptureData().GetCallstackData()->ForEachCallstackEvent(
-          action_on_callstack_events);
+      capture_data->GetCallstackData()->ForEachCallstackEvent(action_on_callstack_events);
     } else {
-      GOrbitApp->GetCaptureData().GetCallstackData()->ForEachCallstackEventOfTid(
-          thread_id_, action_on_callstack_events);
+      capture_data->GetCallstackData()->ForEachCallstackEventOfTid(thread_id_,
+                                                                   action_on_callstack_events);
     }
 
     // Draw selected events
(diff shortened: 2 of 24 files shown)


OrbitQt/orbitmainwindow.cpp
```diff
@@ -585,6 +586,26 @@ void OrbitMainWindow::on_actionClear_Capture_triggered() { GOrbitApp->ClearCaptu
...
+    introspection_widget_ = new OrbitGLWidget();
+    introspection_widget_->setWindowFlags(Qt::WindowStaysOnTopHint);
+    introspection_widget_->Initialize(GlCanvas::CanvasType::kIntrospectionWindow, this, 14);
+    introspection_widget_->installEventFilter(this);
+  }
+
+  introspection_widget_->show();
+}
+
+bool OrbitMainWindow::eventFilter(QObject* object, QEvent* event) {
```
reichlfl, line 600: Nit: Just FYI, AFAIK the proposed way of doing this would to override the `closeEvent` of OrbitGLWidget and not use the eventFilter. But it's quite annoying that there is no signal emitted on close that you can simply connect to, so I think doing it like this is a workable workaround, but it's somewhat abusing eventFilter IMHO.

OrbitGl/TimeGraph.cpp
```diff
@@ -928,19 +936,23 @@ void TimeGraph::SetThreadFilter(const std::string& filter) {
...
-  auto process_track_ = GetOrCreateThreadTrack(SamplingProfiler::kAllThreadsFakeTid);
-  for (const auto& tid_and_count :
-       GOrbitApp->GetCaptureData().GetCallstackData()->GetCallstackEventsCountsPerTid()) {
-    const int32_t thread_id = tid_and_count.first;
-    const uint32_t count = tid_and_count.second;
-    event_count_[thread_id] = count;
-    GetOrCreateThreadTrack(thread_id);
+  std::shared_ptr<ThreadTrack> process_track = nullptr;
+
+  if (IsMainCaptureTimegraph() && GOrbitApp->HasCaptureData()) {
```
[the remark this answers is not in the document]
pierricgimmig: Yeah in my mind it was an intermediate step. I totally agree that it's not the way to go. I submitted a commit that gets rid of it and cleans up the zooming logic, please let me know what you think.

OrbitGl/TimeGraph.cpp
```diff
@@ -971,13 +983,14 @@ void TimeGraph::SortTracks() {
...
 
+    // Tracepoint tracks.
     if (!tracepoints_system_wide_track_->IsEmpty()) {
       all_processes_sorted_tracks.emplace_back(tracepoints_system_wide_track_);
     }
 
     // Process track.
-    if (!process_track_->IsEmpty()) {
-      all_processes_sorted_tracks.emplace_back(process_track_);
+    if (process_track && !process_track->IsEmpty()) {
```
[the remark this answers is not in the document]
pierricgimmig: yes, process_track_ was already a local variable, so I renamed it.

OrbitQt/orbitmainwindow.cpp
pierricgimmig: Ah ok, thanks for the info, I hadn't looked into it in detail. This will change eventually, I was thinking of making a common widget for both the capturewindow and the introspection window so that we could reuse the filters and toolbar.

OrbitGl/TimeGraph.cpp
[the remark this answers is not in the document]
pierricgimmig: Actually, I can replace some more "GOrbitApp->GetCaptureData()" int the code, I'll have another commit in this PR just for that.

pierricgimmig:
@reichlfl , @ronaldfw , please let me know if the PR is too big, I could split it. @ronaldfw , I'd be particularly interested in having your thoughts on https://github.com/google/orbit/pull/1480/commits/0671da8b23057c193ed9b2609496a297322363db regarding the capture data pointer. Capture data is set per-timegraph so that we can have multiple timegraphs with different (or no) capture data.

OrbitGl/EventTrack.cpp
```diff
@@ -75,6 +75,8 @@ void EventTrack::UpdatePrimitives(uint64_t min_tick, uint64_t max_tick, PickingM
 
   const Color kWhite(255, 255, 255, 255);
   const Color kGreenSelection(0, 255, 0, 255);
+  const CaptureData* capture_data = time_graph_->GetCaptureData();
+  CHECK(capture_data);
```
reichlfl, line 79: Should probably be `capture_data != nullptr` for clarity, same as we agreed on in if-statements. Same in a lot of other places where you access `GetCaptureData()`.

OrbitGl/Track.cpp
```diff
@@ -200,7 +200,8 @@ void Track::SetY(float y) {
 }
 
 Color Track::GetBackgroundColor() const {
-  int32_t capture_process_id = GOrbitApp->GetCaptureData().process_id();
+  const CaptureData* capture_data = time_graph_->GetCaptureData();
+  int32_t capture_process_id = capture_data ? capture_data->process_id() : 0;
```
reichlfl, line 204: Don't we have -1 as "invalid process"? Also see line 207.

OrbitGl/CaptureWindow.h
```diff
@@ -63,6 +63,8 @@ class CaptureWindow : public GlCanvas {
   void ToggleDrawHelp();
   void set_draw_help(bool draw_help);
   [[nodiscard]] TimeGraph* GetTimeGraph() { return &time_graph_; }
+  [[nodiscard]] const CaptureData* GetCaptureData() const { return time_graph_.GetCaptureData(); }
```
reichlfl, line 66: Nit: Is this (and the setter below) needed? As far as I can see, the capture data is accessed through time_graph in all cases. It could be clearer to remove both of them and explicitely use `CaptureWindow::GetTimeGraph()->Get/SetCaptureData` where it's needed.

Reasoning: From the outside I find it is not always clear if these methods are proxies for convenience, or if `CaptureWindow::time_graph_` and `CaptureWindow` could hold different instances of `CaptureData`. Since we're exposing the TimeGraph anyways, I don't see the reason to do this.

OrbitGl/TimeGraph.cpp
```diff
@@ -34,6 +34,8 @@ using orbit_client_protos::TimerInfo;
 
 TimeGraph* GCurrentTimeGraph = nullptr;
 
+bool TimeGraph::IsMainCaptureTimegraph() { return this == GCurrentTimeGraph; }
```
reichlfl, line 37: This is no longer needed, is it?

OrbitGl/Track.h
```diff
@@ -57,8 +57,8 @@ class Track : public Pickable, public std::enable_shared_from_this<Track> {
   void SetVisible(bool value) { visible_ = value; }
 
   [[nodiscard]] uint32_t GetNumTimers() const { return num_timers_; }
-  [[nodiscard]] uint64_t GetMinTime() const { return min_time_; }
-  [[nodiscard]] uint64_t GetMaxTime() const { return max_time_; }
+  [[nodiscard]] virtual uint64_t GetMinTime() const { return min_time_; }
```
reichlfl, line 60: This is only done do have it overridden in `EventTrack`, right? I think this is really confusing, as far as I see the `min_time_` / `max_time_` are only set to meaningful values for the `ThreadTrack`, in which case it is actually computed from it's `EventTrack` child, which in turn passes it through from the `capture_data_`...

But IMHO this is a symptom of `EventTrack` inheriting `Track`, which is something we should get rid of. Not in this PR, since it's quite big anyways, but I'd love to see this refactored later and then have this change actually undone.

reichlfl (approved):
Thanks for changing, overall I think this is a cleaner structure. There's a few nitpicks and a larger issue that I think shouldn't be addressed in this PR though.

OrbitGl/TimeGraph.cpp
pierricgimmig: Ah yes, indeed. Thanks.

OrbitGl/EventTrack.cpp
pierricgimmig: Done, thanks.

OrbitGl/CaptureWindow.h
pierricgimmig: I agree, I removed it, thanks!

OrbitGl/Track.h
pierricgimmig: Yes I agree, this should be cleaned up. I still think it's useful for some types of Tracks, but not all. The GetNumTimers has no business in the main Track class for example, it should be in the TimerTrack.

OrbitGl/Track.cpp
pierricgimmig: Yes, thanks!

PR #1481, opened 2020-11-21
Add ThreadUtils.h to OrbitBase

pierricgimmig:
- Add ThreadUtils to OrbitBase and split Linux and Windows implementations
   for a cleaner separation of platform specific code
- Implement GetThreadName and SetThreadName on both platforms
- Add tests for Get/SetThreadName
- Move MonotonicTimestamp and GetThreadName Linux implementation
   from LinuxTracingUtils down to OrbitBase


OrbitBase/Platform/Windows/Profiling.cpp
```diff
@@ -0,0 +1,63 @@
...
+#include <OrbitBase/Platform/Windows/Profiling.h>
+// clang-format on
+
+#include <processthreadsapi.h>
+
+#include <OrbitBase/Logging.h>
+#include <string>
+
+template <typename T>
+inline T GetProcAddress(const std::string& library, const std::string& procedure) {
```
dimitry-, line 15: I do not think inline does anything here. in any case I would leave it to compiler... also maybe make it static.

OrbitBase/ProfilingTest.cpp
```diff
@@ -27,3 +28,18 @@ TEST(Profiling, ThreadId) {
...
 }
 #endif
+
+TEST(Profiling, GetSetThreadNames) {
+  const std::string kThreadName = "ProfilingTest";
+  SetThreadName(kThreadName);
+  std::string thread_name = GetThreadName(GetCurrentThreadId());
+  EXPECT_EQ(kThreadName, thread_name);
+
+  // On Linux, the maximum length for a thread name is 16 characters including '\0'.
```
dimitry-, line 38: Do we want to test this? Is anything bad going to happen to Orbit if linux start limiting thread name with 32 characters?

OrbitBase/include/OrbitBase/Platform/Linux/Profiling.h
```diff
@@ -0,0 +1,29 @@
...
+#include <sys/syscall.h>
+#include <time.h>
+#include <unistd.h>
+
+#include <string>
+
+[[nodiscard]] inline uint64_t MonotonicTimestampNs() {
+  timespec ts;
+  clock_gettime(CLOCK_MONOTONIC, &ts);
+  return 1'000'000'000llu * ts.tv_sec + ts.tv_nsec;
```
dimitry-, line 18: nit: LL (harder to confuse them with 1)

OrbitBase/include/OrbitBase/Platform/Windows/Profiling.h
```diff
@@ -0,0 +1,24 @@
+// Copyright (c) 2020 The Orbit Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#ifndef ORBIT_BASE_PLATFORM_WINDOWS_PROFILING_H_
+#define ORBIT_BASE_PLATFORM_WINDOWS_PROFILING_H_
+
+#include <Windows.h>
```
dimitry-, line 8: I would rather we had one header and 2 implementations.. for inlined functions I suppose it is ok to use ifdefs in this file.

OrbitBase/include/OrbitBase/Platform/Linux/Profiling.h
```diff
@@ -0,0 +1,29 @@
+// Copyright (c) 2020 The Orbit Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#ifndef ORBIT_BASE_PLATFORM_LINUX_PROFILING_H_
+#define ORBIT_BASE_PLATFORM_LINUX_PROFILING_H_
```
dimitry-, line 6: looking at the functions - this should be called something like Utils or split into MonotonicTime.h (Native?)ThreadUtils.h maybe?

OrbitBase/Platform/Linux/Profiling.cpp
```diff
@@ -0,0 +1,51 @@
...
+#include <OrbitBase/Platform/Linux/Profiling.h>
+// clang-format on
+
+#include <OrbitBase/Logging.h>
+
+#include <fstream>
+#include <optional>
+
+namespace {
+static std::optional<std::string> ReadFile(std::string_view filename) {
```
dpallotti, line 15: Having both the anonymous `namespace` and `static` is redundant.

OrbitBase/CMakeLists.txt
```diff
@@ -35,6 +35,16 @@ target_sources(OrbitBase PRIVATE
         Tracing.cpp
         SafeStrerror.cpp)
 
+if(WIN32)
+target_sources(OrbitBase PRIVATE
+        include/OrbitBase/Platform/Windows/Profiling.h
```
dpallotti: Is the "intermediate" `Platform` directory necessary? It seems that it just makes the path longer but without real need.

OrbitBase/Platform/Linux/Profiling.cpp
```diff
@@ -0,0 +1,51 @@
...
+  std::ifstream file{std::string{filename}, std::ios::in | std::ios::binary};
+  if (!file) {
+    ERROR("Could not open \"%s\"", std::string{filename}.c_str());
+    return std::optional<std::string>{};
```
dpallotti, line 19: You can use `std::nullopt`.

OrbitBase/Platform/Linux/Profiling.cpp
```diff
@@ -0,0 +1,51 @@
...
+// clang-format on
+
+#include <OrbitBase/Logging.h>
+
+#include <fstream>
+#include <optional>
+
+namespace {
+static std::optional<std::string> ReadFile(std::string_view filename) {
+  std::ifstream file{std::string{filename}, std::ios::in | std::ios::binary};
```
dpallotti, line 16: Why `std::ios::binary` when you return a string?

OrbitBase/Platform/Linux/Profiling.cpp
```diff
@@ -0,0 +1,51 @@
...
+  // On Linux, "the thread name is a meaningful C language
+  // string, whose length is restricted to 16 characters,
+  // including the terminating null byte ('\0')".
+  constexpr size_t kMaxStringSize = 16;
+  std::string name = thread_name;
+  if (name.length() >= kMaxStringSize) {
+    name[kMaxStringSize - 1] = 0;
+  }
+  pthread_setname_np(pthread_self(), name.data());
+}
```
dpallotti, line 51: These two methods are somewhat confusing as they seem complementary but in reality one returns the name of any thread while the other sets the name of the current thread.

OrbitBase/Platform/Linux/Profiling.cpp
```diff
@@ -0,0 +1,51 @@
...
+
+void SetThreadName(const std::string& thread_name) {
+  // On Linux, "the thread name is a meaningful C language
+  // string, whose length is restricted to 16 characters,
+  // including the terminating null byte ('\0')".
+  constexpr size_t kMaxStringSize = 16;
+  std::string name = thread_name;
+  if (name.length() >= kMaxStringSize) {
+    name[kMaxStringSize - 1] = 0;
+  }
```
dpallotti, line 49: Is this processing necessary? I'm pretty sure that the system (after calling `pthread_setname_np`) already takes care of it.

OrbitBase/Platform/Linux/Profiling.cpp
```diff
@@ -0,0 +1,51 @@
...
+  content << file.rdbuf();
+  return content.str();
+}
+}  // namespace
+
+std::string GetThreadName(pid_t tid) {
+  std::string comm_filename = absl::StrFormat("/proc/%d/comm", tid);
+  std::optional<std::string> comm_content = ReadFile(comm_filename);
+  if (!comm_content.has_value()) {
+    const std::string kEmptyString;
```
dpallotti, line 32: `static`?

OrbitBase/include/OrbitBase/Platform/Linux/Profiling.h
dpallotti: I would avoid a very generic "Utils" but I agree on renaming to something more descriptive and possibly splitting.

OrbitCore/Threading.h
```diff
@@ -9,6 +9,8 @@
 #include <mutex>
```
dpallotti, line 9: It looks like this file should be emptied and then removed.

OrbitBase/include/OrbitBase/Platform/Windows/Profiling.h
```diff
@@ -0,0 +1,24 @@
...
+
+#include <string>
+
+using pid_t = uint32_t;
```
dpallotti, line 12: I dislike this a lot. `pid_t` is a Linux-specific name and it's very confusing that it's suddenly available in Windows code. These `typedef`s to bring one name to the other platform are bad practice. Even worse, on Linux `pid_t` is a signed type.

OrbitBase/Platform/Windows/Profiling.cpp
```diff
@@ -11,31 +11,53 @@
 #include <OrbitBase/Logging.h>
 #include <string>
 
+template <typename T>
+inline T GetProcAddress(const std::string& library, const std::string& procedure) {
+  HMODULE module_handle = LoadLibraryA(library.c_str());
+  if (module_handle == NULL) {
```
dpallotti, line 17: Is there a reason why `nullptr` is not used?

OrbitBase/Platform/Windows/Profiling.cpp
```diff
@@ -0,0 +1,63 @@
...
+// clang-format off
+#include <OrbitBase/Platform/Windows/Profiling.h>
+// clang-format on
+
+#include <processthreadsapi.h>
+
+#include <OrbitBase/Logging.h>
+#include <string>
+
+template <typename T>
```
dpallotti, line 14: Since it puzzled me for a second, can we name `T` something more descriptive, like `FunctionT` or similar?

OrbitBase/include/OrbitBase/Platform/Linux/Profiling.h
pierricgimmig: Ah sure. I favored the implementation from LinuxTracingUtils, will make the change.

OrbitBase/Platform/Windows/Profiling.cpp
pierricgimmig: Actually this should be in a header. It could be reused elsewhere in the code base.

OrbitBase/Platform/Linux/Profiling.cpp
pierricgimmig: This was copied from LinuxTracingUtils and left as is. What would you suggest here?

OrbitBase/Platform/Linux/Profiling.cpp
pierricgimmig: Yeah I agree, but I don't see a need for a SetThreadName that would take in thread id. We could rename it SetCurrentThreadName(). Bah, I'll make SetThreadName take in a tid for completeness add a default parameter for the current thread id.

OrbitBase/Platform/Linux/Profiling.cpp
dimitry-: Even though this is reading a string we still want to read it as is - hence std::ios::binary here.
https://en.cppreference.com/w/cpp/io/c/FILE#Binary_and_text_modes

OrbitBase/Platform/Linux/Profiling.cpp
dpallotti: Honestly I think that renaming to `SetCurrentThreadName` is already good enough. Unless you really want to support setting the name for another tid.

OrbitBase/Platform/Linux/Profiling.cpp
dpallotti: Interesting, I didn't know that.

pierricgimmig:
Hi @dimitry- and @dpallotti , I ended up adding ThreadUtils.h, ThreadUtilsWindows.cpp and ThreadUtilsLinux.cpp, and leaving Profiling.h mostly as is. The goal is to move thread utility functions down to OrbitBase so that they are available across the code base. Thanks for your patience!

OrbitBase/ThreadUtilsTest.cpp
```diff
@@ -0,0 +1,38 @@
...
+}
+#endif
+
+TEST(ThreadUtils, GetSetThreadNames) {
+  const std::string kThreadName = "ThreadUtilsTest";
+  orbit_base::SetCurrentThreadName(kThreadName);
+  std::string thread_name = orbit_base::GetThreadName(orbit_base::GetCurrentThreadId());
+  EXPECT_EQ(kThreadName, thread_name);
+
+  // On Linux, the maximum length for a thread name is 16 characters including '\0'.
```
dimitry-, line 31: Maybe I missed you reply on this - why do we want to test that Linux obeys 16 character limit? Is something bad is going to happen to Orbit if they extend the limit? Is so - probably add a comment here as of why.

OrbitTest/OrbitTest.cpp
```diff
@@ -27,7 +27,7 @@ uint64_t GetThreadID() {
   return std::stoull(ss.str());
 }
 
-void SetThreadName(const std::string& a_Name) {
+void SetCurrentThreadName(const std::string& a_Name) {
```
dimitry-, line 30: SetCurrentThreadNameIfLinux?

OrbitBase/ThreadUtilsLinux.cpp
```diff
@@ -0,0 +1,45 @@
...
+#include "OrbitBase/ThreadUtils.h"
+
+namespace orbit_base {
+
+[[nodiscard]] uint32_t GetCurrentThreadId() {
+  thread_local pid_t current_tid = syscall(__NR_gettid);
+  return static_cast<uint32_t>(current_tid);
+}
+
+std::string GetThreadName(uint32_t tid) {
```
dpallotti, line 19: Same here, I don't like this is unsigned when on Linux pids are signed and we use `int32_t` in the codebase for pids.

dimitry- (approved):
otherwise lgtm

OrbitBase/ThreadUtilsTest.cpp
```diff
@@ -0,0 +1,38 @@
...
+  const size_t kMaxNonZeroCharactersLinux = 15;
+  const std::string kLongThreadName = "ThreadUtilsTestVeryLongName";
+  orbit_base::SetCurrentThreadName(kLongThreadName);
+  std::string long_thread_name = orbit_base::GetThreadName(orbit_base::GetCurrentThreadId());
+  EXPECT_GE(long_thread_name.size(), kMaxNonZeroCharactersLinux);
+  EXPECT_THAT(kLongThreadName, testing::HasSubstr(long_thread_name.c_str()));
```
dpallotti, line 37: Why not check for equality with `ThreadUtilsTestVeryLongName.substr(0, kMaxNonZeroCharactersLinux)`?

dpallotti (changes requested):
I do not like the sign change on Linux, especially on code that we currently only use on Linux.

(Sorry for the rude "Request changes" but Dimitry approved and I disagree.)

pierricgimmig (commented):
@dpallotti, @dimitry- , thanks again for the review, I have addressed the comments. Mainly, I reverted to keeping the original platform specific types for thread ids.

OrbitBase/include/OrbitBase/ThreadUtils.h
```diff
@@ -0,0 +1,25 @@
...
+
+#ifndef ORBIT_BASE_THREAD_UTILS_H_
+#define ORBIT_BASE_THREAD_UTILS_H_
+
+#include <string>
+
+namespace orbit_base {
+
+#ifdef _WIN32
+using thread_id_t = uint32_t;
```
dimitry-, line 13: This bothers me... we currently have a lot of windows code dealing with linux pid_t, so this is not really platform specific it is specific to the platform we are receiving information from. This might become an inconvenience where we are forced to do a lot of static casts or this type will become unused.

OrbitBase/include/OrbitBase/DebugUtils.h
```diff
@@ -9,15 +9,17 @@
...
 #if DEBUG_UTILS_ENABLED
 
 // Print variable's name and value in the form "name = value".
 #define PRINT_VAR(x) LOG("%s = %s", #x, std::to_string(x))
 // Print current function's name, file and line as url, and thread id.
-#define PRINT_FUNC \
-  LOG("%s %s(%u) %s", __FUNCTION__, __FILE__, __LINE__, std::to_string(GetCurrentThreadId()))
+#define PRINT_FUNC                                      \
+  LOG("%s %s(%u) %s", __FUNCTION__, __FILE__, __LINE__, \
+      std::to_string(orbit_base::GetCurrentThreadId()))
```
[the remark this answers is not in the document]
dpallotti: The point of not having them is `Logging.h` is basically to force writing a bit more than just print strings with names of functions and variables. These are ok for printf debugging, but not for actual logging (where we shouldn't be lazy and just write a bit more). Hence why these fit in `DebugUtils.h`.

OrbitBase/ThreadUtilsTest.cpp
dpallotti: Ah right, forgot about Windows.

OrbitBase/include/OrbitBase/ThreadUtils.h
pierricgimmig: @dpallotti , @dimitry- , again, thanks, please see the modifications.

I understand the concern about using a type alias, and I have made the change... I do find it annoying not to have the same type on Windows and Linux. I am tempted to change the Windows thread id type to int32_t and assert that we can do the conversion from uint32_t (in practice, the thread ids on Windows won't reach 2^31). A thread id in our code base would simply be int32_t and It would simplify the interface. Inversely, asserting that the thread id on Linux is always positive could work too... Anyhow, I've spent enough time on this PR for now.

OrbitBase/include/OrbitBase/DebugUtils.h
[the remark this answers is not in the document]
dimitry-: I think this is low cost medium benefit change and sometimes it is useful to see tid in logs, especially if there are logs from different thread coming in. I think we should add tid to base logging.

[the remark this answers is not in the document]
dpallotti: I thought we were talking about `PRINT_VAR` and `PRINT_FUNC` as a whole. If we are only talking about the thread id then yes, I guess we can just add the thread id to all the logs. But that's independent from which macros do that.

dpallotti (approved):

OrbitBase/include/OrbitBase/ThreadUtils.h
dpallotti: I mean, it surely is very annoying. But that's unfortunately one of those differences among the two platforms that are just there. Personally I'm more strongly against checking that the tid is positive on Linux, not because it isn't, but because the type is signed and we rely on that (for example, in some cases `perf_event_open` gives a `-1` pid), as most of the code that deals with thread ids is Linux as Linux is the target of profiling. I guess for the time being there wouldn't be a downside of assuming and checking that a tid on Windows fits in a signed int (unless of course it can happen to have tids above 2^31). It is still quite an assumption though, so I don't know.

PR #1482, opened 2020-11-23
Add a Thread-Safe Dispatch Table for the Vulkan Layer

florian-kuebler:
This wrapper provides methods to create a Vulkan dispatch table,
perform function look-up (queries on that table) and provide the
information whether certain extensions are available in the next
layer (or in the driver).

Test: Compile & run tests
Bug: http://b/168200502

### DIFF
--- a/OrbitVulkanLayer/DispatchTable.cpp
+++ b/OrbitVulkanLayer/DispatchTable.cpp
@@ -7,7 +7,7 @@
 namespace orbit_vulkan_layer {
 
 void DispatchTable::CreateInstanceDispatchTable(
-    VkInstance instance, const PFN_vkGetInstanceProcAddr& next_get_instance_proc_addr_function) {
+    VkInstance instance, PFN_vkGetInstanceProcAddr next_get_instance_proc_addr_function) {
   VkLayerInstanceDispatchTable dispatch_table;
   dispatch_table.DestroyInstance = absl::bit_cast<PFN_vkDestroyInstance>(
       next_get_instance_proc_addr_function(instance, "vkDestroyInstance"));
@@ -35,7 +35,7 @@ void DispatchTable::RemoveInstanceDispatchTable(VkInstance instance) {
 }
 
 void DispatchTable::CreateDeviceDispatchTable(
-    VkDevice device, const PFN_vkGetDeviceProcAddr& next_get_device_proc_addr_function) {
+    VkDevice device, PFN_vkGetDeviceProcAddr next_get_device_proc_addr_function) {
   VkLayerDispatchTable dispatch_table;
 
   dispatch_table.DestroyDevice = absl::bit_cast<PFN_vkDestroyDevice>(
--- a/OrbitVulkanLayer/DispatchTable.cpp
+++ b/OrbitVulkanLayer/DispatchTable.cpp
@@ -91,7 +91,6 @@ void DispatchTable::CreateDeviceDispatchTable(
       next_get_device_proc_addr_function(device, "vkCmdDebugMarkerEndEXT"));
 
   void* key = GetDispatchTableKey(device);
-
   {
     absl::WriterMutexLock lock(&mutex_);
     CHECK(!device_dispatch_table_.contains(key));
--- a/OrbitVulkanLayer/DispatchTable.cpp
+++ b/OrbitVulkanLayer/DispatchTable.cpp
@@ -29,9 +29,11 @@ void DispatchTable::CreateInstanceDispatchTable(
 
 void DispatchTable::RemoveInstanceDispatchTable(VkInstance instance) {
   void* key = GetDispatchTableKey(instance);
-  absl::WriterMutexLock lock(&mutex_);
-  CHECK(instance_dispatch_table_.contains(key));
-  instance_dispatch_table_.erase(key);
+  {
+    absl::WriterMutexLock lock(&mutex_);
+    CHECK(instance_dispatch_table_.contains(key));
+    instance_dispatch_table_.erase(key);
+  }
 }
 
 void DispatchTable::CreateDeviceDispatchTable(
@@ -109,16 +111,18 @@ void DispatchTable::CreateDeviceDispatchTable(
 
 void DispatchTable::RemoveDeviceDispatchTable(VkDevice device) {
   void* key = GetDispatchTableKey(device);
-  absl::WriterMutexLock lock(&mutex_);
+  {
+    absl::WriterMutexLock lock(&mutex_);
 
-  CHECK(device_dispatch_table_.contains(key));
-  device_dispatch_table_.erase(key);
+    CHECK(device_dispatch_table_.contains(key));
+    device_dispatch_table_.erase(key);
 
-  CHECK(device_supports_debug_utils_extension_.contains(key));
-  device_supports_debug_utils_extension_.erase(key);
+    CHECK(device_supports_debug_utils_extension_.contains(key));
+    device_supports_debug_utils_extension_.erase(key);
 
-  CHECK(device_supports_debug_marker_extension_.contains(key));
-  device_supports_debug_marker_extension_.erase(key);
+    CHECK(device_supports_debug_marker_extension_.contains(key));
+    device_supports_debug_marker_extension_.erase(key);
+  }
 }
 
 }  // namespace orbit_vulkan_layer
--- a/OrbitVulkanLayer/DispatchTable.cpp
+++ b/OrbitVulkanLayer/DispatchTable.cpp
@@ -1,0 +1,124 @@
+// Copyright (c) 2020 The Orbit Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#include "DispatchTable.h"
+
+namespace orbit_vulkan_layer {
+
+void DispatchTable::CreateInstanceDispatchTable(
+    VkInstance instance, const PFN_vkGetInstanceProcAddr& next_get_instance_proc_addr_function) {
+  VkLayerInstanceDispatchTable dispatch_table;
+  dispatch_table.DestroyInstance = absl::bit_cast<PFN_vkDestroyInstance>(
+      next_get_instance_proc_addr_function(instance, "vkDestroyInstance"));
+  dispatch_table.GetInstanceProcAddr = absl::bit_cast<PFN_vkGetInstanceProcAddr>(
+      next_get_instance_proc_addr_function(instance, "vkGetInstanceProcAddr"));
+  dispatch_table.EnumerateDeviceExtensionProperties =
+      absl::bit_cast<PFN_vkEnumerateDeviceExtensionProperties>(
+          next_get_instance_proc_addr_function(instance, "vkEnumerateDeviceExtensionProperties"));
+  dispatch_table.GetPhysicalDeviceProperties = absl::bit_cast<PFN_vkGetPhysicalDeviceProperties>(
+      next_get_instance_proc_addr_function(instance, "vkGetPhysicalDeviceProperties"));
+
+  void* key = GetDispatchTableKey(instance);
+  {
+    absl::WriterMutexLock lock(&mutex_);
+    CHECK(!instance_dispatch_table_.contains(key));
+    instance_dispatch_table_[key] = dispatch_table;
+  }
+}
+
+void DispatchTable::RemoveInstanceDispatchTable(VkInstance instance) {
+  void* key = GetDispatchTableKey(instance);
+  absl::WriterMutexLock lock(&mutex_);
+  CHECK(instance_dispatch_table_.contains(key));
+  instance_dispatch_table_.erase(key);
+}
+
+void DispatchTable::CreateDeviceDispatchTable(
+    VkDevice device, const PFN_vkGetDeviceProcAddr& next_get_device_proc_addr_function) {
+  VkLayerDispatchTable dispatch_table;
+
+  dispatch_table.DestroyDevice = absl::bit_cast<PFN_vkDestroyDevice>(
+      next_get_device_proc_addr_function(device, "vkDestroyDevice"));
+  dispatch_table.GetDeviceProcAddr = absl::bit_cast<PFN_vkGetDeviceProcAddr>(
+      next_get_device_proc_addr_function(device, "vkGetDeviceProcAddr"));
+
+  dispatch_table.ResetCommandPool = absl::bit_cast<PFN_vkResetCommandPool>(
+      next_get_device_proc_addr_function(device, "vkResetCommandPool"));
+
+  dispatch_table.AllocateCommandBuffers = absl::bit_cast<PFN_vkAllocateCommandBuffers>(
+      next_get_device_proc_addr_function(device, "vkAllocateCommandBuffers"));
+  dispatch_table.FreeCommandBuffers = absl::bit_cast<PFN_vkFreeCommandBuffers>(
+      next_get_device_proc_addr_function(device, "vkFreeCommandBuffers"));
+  dispatch_table.BeginCommandBuffer = absl::bit_cast<PFN_vkBeginCommandBuffer>(
+      next_get_device_proc_addr_function(device, "vkBeginCommandBuffer"));
+  dispatch_table.EndCommandBuffer = absl::bit_cast<PFN_vkEndCommandBuffer>(
+      next_get_device_proc_addr_function(device, "vkEndCommandBuffer"));
+  dispatch_table.ResetCommandBuffer = absl::bit_cast<PFN_vkResetCommandBuffer>(
+      next_get_device_proc_addr_function(device, "vkResetCommandBuffer"));
+
+  dispatch_table.QueueSubmit = absl::bit_cast<PFN_vkQueueSubmit>(
+      next_get_device_proc_addr_function(device, "vkQueueSubmit"));
+  dispatch_table.QueuePresentKHR = absl::bit_cast<PFN_vkQueuePresentKHR>(
+      next_get_device_proc_addr_function(device, "vkQueuePresentKHR"));
+
+  dispatch_table.GetDeviceQueue = absl::bit_cast<PFN_vkGetDeviceQueue>(
+      next_get_device_proc_addr_function(device, "vkGetDeviceQueue"));
+  dispatch_table.GetDeviceQueue2 = absl::bit_cast<PFN_vkGetDeviceQueue2>(
+      next_get_device_proc_addr_function(device, "vkGetDeviceQueue2"));
+
+  dispatch_table.CreateQueryPool = absl::bit_cast<PFN_vkCreateQueryPool>(
+      next_get_device_proc_addr_function(device, "vkCreateQueryPool"));
+  dispatch_table.ResetQueryPoolEXT = absl::bit_cast<PFN_vkResetQueryPoolEXT>(
+      next_get_device_proc_addr_function(device, "vkResetQueryPoolEXT"));
+
+  dispatch_table.CmdWriteTimestamp = absl::bit_cast<PFN_vkCmdWriteTimestamp>(
+      next_get_device_proc_addr_function(device, "vkCmdWriteTimestamp"));
+
+  dispatch_table.GetQueryPoolResults = absl::bit_cast<PFN_vkGetQueryPoolResults>(
+      next_get_device_proc_addr_function(device, "vkGetQueryPoolResults"));
+
+  dispatch_table.CmdBeginDebugUtilsLabelEXT = absl::bit_cast<PFN_vkCmdBeginDebugUtilsLabelEXT>(
+      next_get_device_proc_addr_function(device, "vkCmdBeginDebugUtilsLabelEXT"));
+  dispatch_table.CmdEndDebugUtilsLabelEXT = absl::bit_cast<PFN_vkCmdEndDebugUtilsLabelEXT>(
+      next_get_device_proc_addr_function(device, "vkCmdEndDebugUtilsLabelEXT"));
+
+  dispatch_table.CmdDebugMarkerBeginEXT = absl::bit_cast<PFN_vkCmdDebugMarkerBeginEXT>(
+      next_get_device_proc_addr_function(device, "vkCmdDebugMarkerBeginEXT"));
+  dispatch_table.CmdDebugMarkerEndEXT = absl::bit_cast<PFN_vkCmdDebugMarkerEndEXT>(
+      next_get_device_proc_addr_function(device, "vkCmdDebugMarkerEndEXT"));
+
+  void* key = GetDispatchTableKey(device);
+
+  {
+    absl::WriterMutexLock lock(&mutex_);
+    CHECK(!device_dispatch_table_.contains(key));
+    device_dispatch_table_[key] = dispatch_table;
+
+    CHECK(!device_supports_debug_utils_extension_.contains(key));
+    device_supports_debug_utils_extension_[key] =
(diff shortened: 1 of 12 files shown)


OrbitVulkanLayer/DispatchTable.h
```diff
@@ -0,0 +1,413 @@
...
+ * extension is available.
+ *
+ * Thread-Safety: This class is internally synchronized (using read/write locks) and can be safely
+ * accessed from different threads.
+ */
+class DispatchTable {
+ public:
+  DispatchTable() = default;
+
+  void CreateInstanceDispatchTable(
```
danielfenner, line 35: Is there a reason to make this header only? Otherwise I'd prefer to put the non-template stuff into the .cpp.

OrbitVulkanLayer/DispatchTableTest.cpp
```diff
@@ -0,0 +1,803 @@
...
+  uint32_t property_count;
+
+  VkResult result = dispatch_table.EnumerateDeviceExtensionProperties(instance)(
+      device, nullptr, &property_count, nullptr);
+  EXPECT_EQ(result, VK_SUCCESS);
+  EXPECT_EQ(property_count, 42);
+}
+
+TEST(DispatchTable, CanCallGetPhysicalDeviceProperties) {
+  // We can not create an actual VkInstance, but the first bytes of an any dispatchable type in
```
danielfenner, line 228: `the first bytes of an any dispatchable 
`->
`the first bytes of any dispatchable `

danielfenner: Also: I'd probably put this comment somewhere at the top once instead of repeating it so many times.

OrbitVulkanLayer/DispatchTableTest.cpp
```diff
@@ -0,0 +1,803 @@
...
+}
+
+TEST(DispatchTable, CanCallResetQueryPoolEXT) {
+  // We can not create an actual VkInstance, but the first bytes of an any dispatchable type in
+  // Vulkan will be a pointer to a dispatch table. This characteristic will be used by our dispatch
+  // table wrapper, so we need to mimic it.
+  VkLayerDispatchTable some_dispatch_table = {};
+  auto device = absl::bit_cast<VkDevice>(&some_dispatch_table);
+
+  static bool was_called = false;
```
danielfenner, line 611: Why is this static? You don't want to preserve the state across different runs, right?

OrbitVulkanLayer/DispatchTable.h
florian-kuebler: I moved it all to the header, when I started adding the templated methods. If you prefer to have the two create- and the two remove methods in a .cpp file, I am fine with this,

OrbitVulkanLayer/DispatchTableTest.cpp
florian-kuebler: It needs to be static, such that I can use the lambda as function pointer. However, you are right, I should at least set it to false after the check again.

OrbitVulkanLayer/DispatchTable.h
```diff
@@ -0,0 +1,413 @@
...
+
+namespace orbit_vulkan_layer {
+
+/*
+ * A thread-safe dispatch table for Vulkan function look-up.
+ *
+ * It computes/stores the Vulkan dispatch tables for concrete devices/instances and provides
+ * accessors to the functions.
+ *
+ * For functions provided by extensions it also provide predicate functions to check if the
```
dpallotti, line 25:
```suggestion
 * For functions provided by extensions it also provides predicate functions to check if the
```

OrbitVulkanLayer/DispatchTable.h
```diff
@@ -0,0 +1,413 @@
...
+ *
+ * Thread-Safety: This class is internally synchronized (using read/write locks) and can be safely
+ * accessed from different threads.
+ */
+class DispatchTable {
+ public:
+  DispatchTable() = default;
+
+  void CreateInstanceDispatchTable(
+      VkInstance instance, const PFN_vkGetInstanceProcAddr& next_get_instance_proc_addr_function) {
```
dpallotti: Should `next_get_instance_proc_addr_function` also be passed by value similarly to other Vulkan handles as it's always a function pointer (and not, for example, a lambda)?

OrbitVulkanLayer/DispatchTable.h
florian-kuebler: I moved the non-template methods to a cpp file.

OrbitVulkanLayer/DispatchTableTest.cpp
florian-kuebler: I added a comment at the very top and removed them here.

OrbitVulkanLayer/DispatchTable.h
```diff
@@ -0,0 +1,413 @@
...
+  }
+
+ private:
+  // In vulkan, every "dispatchable type" has as a very first field in memory a pointer to the
+  // internal dispatch table. This pointer is unique per device/instance. So for example
+  // for a command buffer allocated on a certain device, this pointer is the same for the buffer
+  // and for the device. So we can use that pointer to uniquely map dispatchable types to their
+  // dispatch table.
+  template <typename DispatchableType>
+  void* GetDispatchTableKey(DispatchableType dispatchable_object) {
```
dpallotti, line 286: `dispatchable_object` is passed by value here but by reference in many other places. Is there a reason for that?

OrbitVulkanLayer/DispatchTable.h
```diff
@@ -0,0 +1,413 @@
...
+  template <typename DispatchableType>
+  bool IsDebugUtilsExtensionSupported(const DispatchableType& dispatchable_object) {
+    absl::ReaderMutexLock lock(&mutex_);
+    void* key = GetDispatchTableKey(dispatchable_object);
+    CHECK(device_supports_debug_utils_extension_.contains(key));
+    return device_supports_debug_utils_extension_.at(key);
+  }
+
+ private:
+  // In vulkan, every "dispatchable type" has as a very first field in memory a pointer to the
```
dpallotti, line 280: Could you elaborate more on what a dispatchable type is? Just a device or instance? Or more? Is a command buffer that you mention in this comment also a dispatchable type?

OrbitVulkanLayer/DispatchTable.h
```diff
@@ -0,0 +1,413 @@
...
+      CHECK(!instance_dispatch_table_.contains(key));
+      instance_dispatch_table_[key] = dispatch_table;
+    }
+  }
+
+  void RemoveInstanceDispatchTable(VkInstance instance) {
+    void* key = GetDispatchTableKey(instance);
+    absl::WriterMutexLock lock(&mutex_);
+    CHECK(instance_dispatch_table_.contains(key));
+    instance_dispatch_table_.erase(key);
```
dpallotti: Add braces around this like in other methods to highlight that `GetDispatchTableKey` is not under the lock?

OrbitVulkanLayer/DispatchTableTest.cpp
florian-kuebler: I've the static bool to false at the end of each test.

florian-kuebler (commented):
Thanks for the review so far, I addressed a couple of comments.

OrbitVulkanLayer/DispatchTable.h
```diff
@@ -0,0 +1,413 @@
...
+    device_supports_debug_utils_extension_.erase(key);
+
+    CHECK(device_supports_debug_marker_extension_.contains(key));
+    device_supports_debug_marker_extension_.erase(key);
+  }
+
+  template <typename DispatchableType>
+  PFN_vkDestroyDevice DestroyDevice(const DispatchableType& dispatchable_object) {
+    absl::ReaderMutexLock lock(&mutex_);
+    void* key = GetDispatchTableKey(dispatchable_object);
```
dpallotti, line 46: This is now under the lock but in previous methods it wasn't. What is the reason?

OrbitVulkanLayer/DispatchTable.h
```diff
@@ -0,0 +1,413 @@
...
+
+ private:
+  // In vulkan, every "dispatchable type" has as a very first field in memory a pointer to the
+  // internal dispatch table. This pointer is unique per device/instance. So for example
+  // for a command buffer allocated on a certain device, this pointer is the same for the buffer
+  // and for the device. So we can use that pointer to uniquely map dispatchable types to their
+  // dispatch table.
+  template <typename DispatchableType>
+  void* GetDispatchTableKey(DispatchableType dispatchable_object) {
+    return *absl::bit_cast<void**>(dispatchable_object);
```
dpallotti, line 287: This is a bit of a hack to access the first 8 bytes. Is there a chance that `hwvulkan_dispatch_t dispatch` of `struct VkInstance_T` and `struct VkDevice_T` (done a minor research here) is consistent enough that you can access it by name just with `dispatchable_object->dispatch`?

OrbitVulkanLayer/DispatchTable.h
```diff
@@ -0,0 +1,413 @@
...
+  // layer in the dispatch chain among our handling of functions we intercept.
+  absl::flat_hash_map<void*, VkLayerInstanceDispatchTable> instance_dispatch_table_;
+  absl::flat_hash_map<void*, VkLayerDispatchTable> device_dispatch_table_;
+  absl::flat_hash_map<void*, bool> device_supports_debug_marker_extension_;
+  absl::flat_hash_map<void*, bool> device_supports_debug_utils_extension_;
+
+  // Must protect access to dispatch tables above by mutex since the Vulkan
+  // application may be calling these functions from different threads.
+  // However, they are usually filled once (per device/instance) at the beginning
+  // and afterwards we only read that data. So we use a read/write lock.
```
dpallotti, line 300:
```suggestion
  // and afterwards we only read that data. So we use read/write locks.
```

OrbitVulkanLayer/DispatchTableTest.cpp
```diff
@@ -0,0 +1,803 @@
...
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#include "DispatchTable.h"
+#include "gtest/gtest.h"
+
+namespace orbit_vulkan_layer {
+
+TEST(DispatchTable, CanInitializeInstance) {
+  // We can not create an actual VkInstance, but the first bytes of an any dispatchable type in
```
dpallotti:
```suggestion
  // We cannot create an actual VkInstance, but the first bytes of any dispatchable type in
```

OrbitVulkanLayer/DispatchTableTest.cpp
```diff
@@ -0,0 +1,803 @@
...
+  // Vulkan will be a pointer to a dispatch table. This characteristic will be used by our dispatch
+  // table wrapper, so we need to mimic it.
+  VkLayerInstanceDispatchTable some_dispatch_table = {};
+  auto instance = absl::bit_cast<VkInstance>(&some_dispatch_table);
+  PFN_vkGetInstanceProcAddr next_get_instance_proc_addr_function =
+      +[](VkInstance /*instance*/, const char * /*name*/) -> PFN_vkVoidFunction { return nullptr; };
```
dpallotti, line 20:
```suggestion
      +[](VkInstance /*instance*/, const char* /*name*/) -> PFN_vkVoidFunction { return nullptr; };
```

dpallotti: Here and everywhere else.

OrbitVulkanLayer/DispatchTableTest.cpp
dpallotti: Here and everywhere else.

OrbitVulkanLayer/DispatchTableTest.cpp
```diff
@@ -0,0 +1,803 @@
...
+
+  DispatchTable dispatch_table = {};
+  dispatch_table.CreateInstanceDispatchTable(instance, next_get_instance_proc_addr_function);
+  dispatch_table.RemoveInstanceDispatchTable(instance);
+}
+
+TEST(DispatchTable, CanInitializeInstanceTwiceAfterRemove) {
```
dpallotti, line 52: Would suggest something like `CanReinitializeInstanceAfterRemove` otherwise "twice after remove" conveys the wrong idea.

OrbitVulkanLayer/DispatchTableTest.cpp
```diff
@@ -0,0 +1,803 @@
...
+  auto device = absl::bit_cast<VkDevice>(&some_dispatch_table);
+  PFN_vkGetDeviceProcAddr next_get_device_proc_addr_function =
+      +[](VkDevice /*instance*/, const char * /*name*/) -> PFN_vkVoidFunction { return nullptr; };
+
+  DispatchTable dispatch_table = {};
+  dispatch_table.CreateDeviceDispatchTable(device, next_get_device_proc_addr_function);
+  dispatch_table.RemoveDeviceDispatchTable(device);
+}
+
+TEST(DispatchTable, CanInitializeDeviceTwiceAfterRemove) {
```
dpallotti, line 98: Same suggestion here as above.

OrbitVulkanLayer/DispatchTableTest.cpp
```diff
@@ -0,0 +1,803 @@
...
+      +[](VkInstance /*instance*/, const char * /*name*/) -> PFN_vkVoidFunction { return nullptr; };
+
+  DispatchTable dispatch_table = {};
+  dispatch_table.CreateInstanceDispatchTable(instance, next_get_instance_proc_addr_function);
+}
+
+TEST(DispatchTable, CannotInitializeInstanceTwice) {
+  // We can not create an actual VkInstance, but the first bytes of an any dispatchable type in
+  // Vulkan will be a pointer to a dispatch table. This characteristic will be used by our dispatch
+  // table wrapper, so we need to mimic it.
```
dpallotti: Ok, I don't know if you really need this comment everywhere, maybe you can have just one that explains what you're doing only once.

OrbitVulkanLayer/DispatchTableTest.cpp
dpallotti: Can we assume that each test is run only once per execution of the test binary?

dpallotti:
The vast majority of comments are really minor, I like the code. I'm curious about `GetDispatchTableKey`.

OrbitVulkanLayer/DispatchTable.h
florian-kuebler: There is no need to put this under a lock. I moved it upwards.

OrbitVulkanLayer/DispatchTable.h
florian-kuebler: I expanded the comment here a bit. Hope it gets clearer now.

OrbitVulkanLayer/DispatchTable.h
florian-kuebler: Well as they are all essentially Vulkan types, and we decided not to pass them as reference, we should not pass them as reference here.
Thanks for spotting the const-refs in all the accessor functions. I removed them and made them pass-by-value.

OrbitVulkanLayer/DispatchTable.h
florian-kuebler: Done, thanks!

OrbitVulkanLayer/DispatchTable.h
florian-kuebler: Unfortunately that does not compile.

OrbitVulkanLayer/DispatchTableTest.cpp
florian-kuebler: Done, thanks!

OrbitVulkanLayer/DispatchTableTest.cpp
florian-kuebler: Sounds good.

OrbitVulkanLayer/DispatchTableTest.cpp
florian-kuebler: Done.

OrbitVulkanLayer/DispatchTableTest.cpp
florian-kuebler: That's a good question. So far, yes. Especially, as we would only have an issue as the same test gets executed from the same process in parallel.

OrbitVulkanLayer/DispatchTable.h
florian-kuebler: Yes, makes sense

OrbitVulkanLayer/DispatchTable.h
florian-kuebler: Done!

OrbitVulkanLayer/DispatchTableTest.cpp
dpallotti: Ok, that was to say that maybe it's a bit weird to reset `was_called` if you don't expect this function to be called again. But on the other hand I guess it's theoretically more precise.

OrbitVulkanLayer/DispatchTableTest.cpp
```diff
@@ -0,0 +1,722 @@
...
+TEST(DispatchTable, CanCallAllocateCommandBuffers) {
+  VkLayerDispatchTable some_dispatch_table = {};
+  auto device = absl::bit_cast<VkDevice>(&some_dispatch_table);
+
+  PFN_vkGetDeviceProcAddr next_get_device_proc_addr_function =
+      +[](VkDevice /*device*/, const char* name) -> PFN_vkVoidFunction {
+    if (strcmp(name, "vkAllocateCommandBuffers") == 0) {
+      PFN_vkAllocateCommandBuffers function =
+          +[](VkDevice /*device*/, const VkCommandBufferAllocateInfo* /*allocate_info*/,
+              VkCommandBuffer *
```
dpallotti, line 309:
```suggestion
              VkCommandBuffer*
```

dpallotti (approved):
To me it looks really good. Maybe `GetDispatchTableKey` can still be made more readable but it's not critical.

OrbitVulkanLayer/DispatchTableTest.cpp
ronaldfw: Sorry if this is not on point, but one can repeat tests multiple times with --gtest_repeat=10 (or whatever number you want), which is useful when debugging a flaky test. This does not run anything in parallel, though.

OrbitVulkanLayer/DispatchTable.h
```diff
@@ -0,0 +1,359 @@
...
+    {
+      absl::ReaderMutexLock lock(&mutex_);
+      CHECK(device_supports_debug_utils_extension_.contains(key));
+      return device_supports_debug_utils_extension_.at(key);
+    }
+  }
+
+ private:
+  // Vulkan has the concept of "dispatchable types". Basically every Vulkan type whose objects can
+  // be associated with a VkInstance or VkDevice are "dispatchable". As an example both, a device
```
ronaldfw, line 332: Comma in the wrong place: "As an example both," --> "As an example, both"

OrbitVulkanLayer/DispatchTable.cpp
```diff
@@ -0,0 +1,128 @@
...
+  dispatch_table.CmdEndDebugUtilsLabelEXT = absl::bit_cast<PFN_vkCmdEndDebugUtilsLabelEXT>(
+      next_get_device_proc_addr_function(device, "vkCmdEndDebugUtilsLabelEXT"));
+
+  dispatch_table.CmdDebugMarkerBeginEXT = absl::bit_cast<PFN_vkCmdDebugMarkerBeginEXT>(
+      next_get_device_proc_addr_function(device, "vkCmdDebugMarkerBeginEXT"));
+  dispatch_table.CmdDebugMarkerEndEXT = absl::bit_cast<PFN_vkCmdDebugMarkerEndEXT>(
+      next_get_device_proc_addr_function(device, "vkCmdDebugMarkerEndEXT"));
+
+  void* key = GetDispatchTableKey(device);
+
```
ronaldfw, line 94: Nit: Remove empty line here.

OrbitVulkanLayer/DispatchTable.h
```diff
@@ -0,0 +1,359 @@
...
+      return device_dispatch_table_.at(key).DestroyDevice;
+    }
+  }
+
+  template <typename DispatchableType>
+  PFN_vkDestroyInstance DestroyInstance(DispatchableType dispatchable_object) {
+    void* key = GetDispatchTableKey(dispatchable_object);
+    {
+      absl::ReaderMutexLock lock(&mutex_);
+      CHECK(instance_dispatch_table_.contains(key));
```
ronaldfw, line 59: Are these methods called on every call to whatever function is being dispatched? How much overhead do all these CHECKS generate? If these methods are only called once for some form of initialization, it doesn't matter.

florian-kuebler: They are called on each dispatch. So yes, overhead might be a problem. I haven't done any proper performance analysis of the layer. So far it does not seem to be an issue.
I'd suggest for now to keep it as it is, and if we have a proper analysis of the overhead, we can check if e.g. DCHECK is reasonable.

OrbitVulkanLayer/DispatchTableTest.cpp
florian-kuebler: Right, I was also thinking of this. In this case, the reset at the end should be the right thing to do.

dpallotti: Maybe you can add a comment in the first test that does this explaining why this needs to be static (basically saying what you replied to Daniel).

ronaldfw: This might be a good case for using a test fixture (TEST_F) as it seems you need to reset state after each test. You can also put most of the common setup code into that.

OrbitVulkanLayer/DispatchTable.h
florian-kuebler: I updated it using memcpy

OrbitVulkanLayer/DispatchTableTest.cpp
florian-kuebler: Regarding the TEST_F, if I get it correctly, then we would have one static bool for all the tests, rather then one per test, right? Are we sure, that no tests are executed in parallel?

florian-kuebler: Anyways added a comment to the first usage.

danielfenner (approved):

OrbitVulkanLayer/DispatchTableTest.cpp
ronaldfw: Depends on whether you create a test fixture for all tests or one for each test, though the latter seems a bit too much overhead and boilerplate. I guess best to ignore this idea :)

florian-kuebler: Agree, I would have been nice, but I guess it would also be too much overhead. Let's keep it as it for now.

florian-kuebler:
I needed to let the layer only compile in ggp. @beckerhe Could you have quick look at those changes?

OrbitVulkanLayer/DispatchTable.h
ronaldfw: Ok, we'll have to take a close look at overhead anyway. Let's keep them for now and change if needed later.

ronaldfw (approved):
LGTM

beckerhe (approved):
I just looked at the cmake/conan part. That looks good!

PR #1485, opened 2020-11-24
Read buffer optimizations

beckerhe:
Two smallish optimizations to the SSH buffer handling code. This avoids superfluous allocations and makes the code faster.

I tried that with @pierricgimmig new introspection feature and saw improvements up to a factor of 2 or 3. No changes for small data chunks (as expected) but for larger ones the time spent in the copy-function went down quite a bit. For my example it was 25ms before, afterwards 10ms - 15ms.

So I'm rather confident this change doesn't make the situation worse. Obviously more can be done.

### DIFF
--- a/OrbitSshQt/Tunnel.cpp
+++ b/OrbitSshQt/Tunnel.cpp
@@ -183,8 +183,8 @@ outcome::result<void> Tunnel::readFromChannel() {
     if (bytes_written == -1) {
       SetError(Error::kLocalSocketClosed);
     } else {
-      // TODO(b/172686811): avoid extra copy.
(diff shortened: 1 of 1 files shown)


ronaldfw:
Typo in "So I'm rather confident this change does make the situation worse."? :P

Improvement numbers look great, thanks!

beckerhe:
> Typo in "So I'm rather confident this change does make the situation worse."? :P
> 
> Improvement numbers look great, thanks!

:see_no_evil: Yes. Thanks. Fixed!

pierricgimmig (approved):
Awesome, thanks Henning!

PR #1486, opened 2020-11-24
Add ProcessItemModel class

antonrohr:
ProcessItemModel is based on ProcessInfo. Tests are included.

This is used in the upcoming UI

### DIFF
--- a/OrbitQt/ProcessItemModel.h
+++ b/OrbitQt/ProcessItemModel.h
@@ -1,0 +1,40 @@
+// Copyright (c) 2020 The Orbit Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#ifndef ORBIT_QT_PROCESS_ITEM_MODEL_H_
+#define ORBIT_QT_PROCESS_ITEM_MODEL_H_
+
+#include <QAbstractItemModel>
+#include <QVariant>
+
+#include "process.pb.h"
+
+namespace OrbitQt {
+
+class ProcessItemModel : public QAbstractItemModel {
+  Q_OBJECT
+ public:
+  enum class Column { kName, kPid, kCpu, kEnd };
+
+  int columnCount(const QModelIndex& parent = {}) const override;
+  QVariant data(const QModelIndex& idx, int role = Qt::DisplayRole) const override;
+  QVariant headerData(int section, Qt::Orientation orientation,
+                      int role = Qt::DisplayRole) const override;
+  QModelIndex index(int row, int column, const QModelIndex& parent = {}) const override;
+  QModelIndex parent(const QModelIndex& parent) const override;
+  int rowCount(const QModelIndex& parent = {}) const override;
+
+  void SetProcesses(std::vector<orbit_grpc_protos::ProcessInfo> processes);
+  bool HasProcesses() const { return !processes_.empty(); }
+  void Clear() { SetProcesses({}); }
+
+ private:
+  std::vector<orbit_grpc_protos::ProcessInfo> processes_;
+};
+
+}  // namespace OrbitQt
+
+Q_DECLARE_METATYPE(const orbit_grpc_protos::ProcessInfo*);
+
+#endif  // ORBIT_QT_PROCESS_ITEM_MODEL_H_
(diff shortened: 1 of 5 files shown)


OrbitQt/ProcessItemModel.h
```diff
@@ -0,0 +1,41 @@
...
+  QVariant data(const QModelIndex& idx, int role = Qt::DisplayRole) const override;
+  // maybe flags?
+  QVariant headerData(int section, Qt::Orientation orientation,
+                      int role = Qt::DisplayRole) const override;
+  QModelIndex index(int row, int column, const QModelIndex& parent = {}) const override;
+  QModelIndex parent(const QModelIndex& parent) const override;
+  int rowCount(const QModelIndex& parent = {}) const override;
+
+  void SetProcesses(std::vector<orbit_grpc_protos::ProcessInfo> processes);
+  bool HasProcesses() { return !processes_.empty(); }
```
beckerhe, line 30: const?

OrbitQt/ProcessItemModel.h
```diff
@@ -0,0 +1,41 @@
...
+namespace OrbitQt {
+
+class ProcessItemModel : public QAbstractItemModel {
+  // Q_OBJECT
+ public:
+  enum class Column { kName, kPid, kCpu, kEnd };
+
+  int columnCount(const QModelIndex& parent = {}) const override;
+  QVariant data(const QModelIndex& idx, int role = Qt::DisplayRole) const override;
+  // maybe flags?
```
beckerhe, line 22: Is this comment intended?

OrbitQt/ProcessItemModel.h
```diff
@@ -0,0 +1,41 @@
...
+
+#include <QAbstractItemModel>
+#include <QVariant>
+
+#include "process.pb.h"
+
+namespace OrbitQt {
+
+class ProcessItemModel : public QAbstractItemModel {
+  // Q_OBJECT
```
beckerhe, line 16: This macro should be there. Is there a reason for commenting it?

OrbitQt/ProcessItemModelTest.cpp
```diff
@@ -0,0 +1,87 @@
...
+class AssertNoQtLogWarnings {
+  static void MessageHandlerTest(QtMsgType type, const QMessageLogContext& context,
+                                 const QString& msg) {
+    static bool NO_WARNING_MSG = true;
+    QByteArray localMsg = msg.toLocal8Bit();
+    const char* file = context.file ? context.file : "";
+    const char* function = context.function ? context.function : "";
+    switch (type) {
+      case QtDebugMsg:
+        fprintf(stderr, "Debug: %s (%s:%u, %s)\n", localMsg.constData(), file, context.line,
```
beckerhe, line 23: fprintf requires "#include <stdio.h>"

OrbitQt/ProcessItemModelTest.cpp
```diff
@@ -0,0 +1,87 @@
...
+    qInstallMessageHandler(nullptr);
+  }
+};
+
+}  // namespace
+
+namespace OrbitQt {
+
+TEST(ProcessItemModel, ProcessItemModel) {
+  AssertNoQtLogWarnings log_qt_test;
```
beckerhe, line 56: This is very cool. Can you add a short comment why you're doing this? My guess is that the AbstractItemModelTester spits out warnings which have to be converted into a failed gtest asserts, right?

beckerhe (approved):
LGTM!

OrbitQt/ProcessItemModel.h
antonrohr: done

OrbitQt/ProcessItemModel.h
antonrohr: no, just a mistakes on my side. I should read over changes one final time before I create the PR. :/

OrbitQt/ProcessItemModel.h
antonrohr: No, mistake on my side

OrbitQt/ProcessItemModelTest.cpp
antonrohr: I used Log instead :)

OrbitQt/ProcessItemModelTest.cpp
antonrohr: I added a short comment, please let me know if insufficient, otherwise I will merge right away :)

beckerhe: I'm sorry, I'm still missing the connection to QAbstractItemModelTester in here. Maybe instead of "produced by Qt"
you could write "produced by QAbstractItemModelTester". Technically that's not fully correct but I think it makes it easier
to understand why we want to fail on log messages.

PR #1488, opened 2020-11-24
Add CaptureStartStopListener and start integrating into OrbitService

dpallotti:
This will be needed for later changes that add actual communication with the
Vulkan layer.

Bug: http://b/174028631

### DIFF
--- a/OrbitService/CaptureServiceImpl.cpp
+++ b/OrbitService/CaptureServiceImpl.cpp
@@ -119,6 +119,33 @@ class GrpcCaptureEventSender final : public CaptureEventSender {
 
 }  // namespace
 
+// LinuxTracingHandler::Stop is blocking, until all perf_event_open events have been processed
+// and all perf_event_open file descriptors have been closed.
+// CaptureStartStopListener::OnCaptureStopRequested is also to be assumed blocking,
+// for example until all CaptureEvents from external producers have been received.
+// Hence why these methods need to be called in parallel on different threads.
+static void StopTracingHandlerAndCaptureStartStopListenersInParallel(
+    LinuxTracingHandler* tracing_handler,
+    absl::flat_hash_set<CaptureStartStopListener*>* capture_start_stop_listeners) {
+  std::vector<std::thread> stop_threads;
+
+  stop_threads.emplace_back([&tracing_handler] {
+    tracing_handler->Stop();
+    LOG("LinuxTracingHandler stopped: perf_event_open tracing is done");
+  });
+
+  for (CaptureStartStopListener* listener : *capture_start_stop_listeners) {
+    stop_threads.emplace_back([&listener] {
+      listener->OnCaptureStopRequested();
+      LOG("CaptureStartStopListener stopped: one or more producers finished capturing");
+    });
+  }
+
+  for (std::thread& stop_thread : stop_threads) {
+    stop_thread.join();
+  }
+}
+
 grpc::Status CaptureServiceImpl::Capture(
     grpc::ServerContext*,
     grpc::ServerReaderWriter<CaptureResponse, CaptureRequest>* reader_writer) {
@@ -137,7 +164,11 @@ grpc::Status CaptureServiceImpl::Capture(
   CaptureRequest request;
   reader_writer->Read(&request);
   LOG("Read CaptureRequest from Capture's gRPC stream: starting capture");
+
   tracing_handler.Start(std::move(*request.mutable_capture_options()));
+  for (CaptureStartStopListener* listener : capture_start_stop_listeners_) {
+    listener->OnCaptureStartRequested(&capture_event_buffer);
+  }
 
   // The client asks for the capture to be stopped by calling WritesDone.
   // At that point, this call to Read will return false.
@@ -146,13 +177,23 @@ grpc::Status CaptureServiceImpl::Capture(
   }
   LOG("Client finished writing on Capture's gRPC stream: stopping capture");
 
-  tracing_handler.Stop();
-  LOG("LinuxTracingHandler stopped: perf_event_open tracing is done");
+  StopTracingHandlerAndCaptureStartStopListenersInParallel(&tracing_handler,
+                                                           &capture_start_stop_listeners_);
 
   capture_event_buffer.StopAndWait();
   LOG("Finished handling gRPC call to Capture: all capture data has been sent");
   is_capturing = false;
   return grpc::Status::OK;
+}
+
+void CaptureServiceImpl::AddCaptureStartStopListener(CaptureStartStopListener* listener) {
+  bool new_insertion = capture_start_stop_listeners_.insert(listener).second;
+  CHECK(new_insertion);
+}
+
+void CaptureServiceImpl::RemoveCaptureStartStopListener(CaptureStartStopListener* listener) {
+  bool was_removed = capture_start_stop_listeners_.erase(listener) > 0;
+  CHECK(was_removed);
 }
 
 }  // namespace orbit_service
--- a/OrbitService/CaptureStartStopListener.h
+++ b/OrbitService/CaptureStartStopListener.h
@@ -1,0 +1,26 @@
+// Copyright (c) 2020 The Orbit Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#ifndef ORBIT_SERVICE_CAPTURE_START_STOP_LISTENER_H_
+#define ORBIT_SERVICE_CAPTURE_START_STOP_LISTENER_H_
+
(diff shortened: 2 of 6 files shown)


OrbitService/CaptureStartStopListener.h
```diff
@@ -0,0 +1,28 @@
...
+#define ORBIT_SERVICE_CAPTURE_START_STOP_LISTENER_H_
+
+#include "CaptureEventBuffer.h"
+
+namespace orbit_service {
+
+// This interface is used to propagate the requests
+// to start and stop the capture received by CaptureServiceImpl,
+// together with the CaptureEventBuffer where to add the generated CaptureEvents.
+
```
florian-kuebler, line 15: I'd remove that empty line.

OrbitService/CaptureStartStopListener.h
```diff
@@ -0,0 +1,28 @@
...
+#ifndef ORBIT_SERVICE_CAPTURE_START_STOP_LISTENER_H_
+#define ORBIT_SERVICE_CAPTURE_START_STOP_LISTENER_H_
+
+#include "CaptureEventBuffer.h"
+
+namespace orbit_service {
+
+// This interface is used to propagate the requests
+// to start and stop the capture received by CaptureServiceImpl,
```
florian-kuebler, line 13: Super nit: But I think your text should be aligned to 100column line.

OrbitService/CaptureServiceImpl.cpp
```diff
@@ -146,13 +150,37 @@ grpc::Status CaptureServiceImpl::Capture(
...
+    }
+  }
 
   capture_event_buffer.StopAndWait();
   LOG("Finished handling gRPC call to Capture: all capture data has been sent");
   is_capturing = false;
   return grpc::Status::OK;
 }
 
+void CaptureServiceImpl::AddCaptureStartStopListener(CaptureStartStopListener* listener) {
```
florian-kuebler, line 176: Would that be the right point in time to start writing tests for this class?

OrbitService/CaptureServiceImpl.cpp
```diff
@@ -146,13 +150,37 @@ grpc::Status CaptureServiceImpl::Capture(
   }
   LOG("Client finished writing on Capture's gRPC stream: stopping capture");
 
-  tracing_handler.Stop();
-  LOG("LinuxTracingHandler stopped: perf_event_open tracing is done");
+  {
```
florian-kuebler, line 153: Is there a reason to put it in its own scope?

florian-kuebler: I'd actually either add comment explaining what goes on here, or move that to a separate method.

OrbitService/CaptureServiceImpl.cpp
```diff
@@ -137,7 +137,11 @@ grpc::Status CaptureServiceImpl::Capture(
   CaptureRequest request;
   reader_writer->Read(&request);
   LOG("Read CaptureRequest from Capture's gRPC stream: starting capture");
+
   tracing_handler.Start(std::move(*request.mutable_capture_options()));
```
florian-kuebler, line 141: You stop using different threads. Do you also want to do this here for the start?

florian-kuebler (commented):
In general looks very good to me. I added some minor comments.

OrbitService/CaptureServiceImpl.cpp
dpallotti: The reason why I'm stopping using different threads is that the "stop" methods are blocking, waiting for remaining events (I'll add a comment there). These "start" methods are not blocking hence why they are in the same thread.

OrbitService/CaptureServiceImpl.cpp
dpallotti: The scope is just to limit the scope of `stop_threads`. I can see how it feels like this scope should be a separate function, although in my opinion it makes things a bit harder to follow here. I'm adding a comment in the meantime.

OrbitService/CaptureServiceImpl.cpp
dpallotti: Which class do you mean? `CaptureServiceImpl`? I'm really not sure how to do that, as it's a gRPC service. Unless we involve gRPC in the tests, but should we go there?

OrbitService/CaptureStartStopListener.h
dpallotti: The rationale here is that when a comment doesn't fit into a smaller number of lines then I put new lines in places that split the sentence the least. Not sure if it's great practice but that's what's going on here.

OrbitService/CaptureStartStopListener.h
dpallotti: Done.

OrbitService/CaptureStartStopListener.h
florian-kuebler: Still feel it looks a bit strange, but nothing I would argue strongly for or against.

OrbitService/CaptureServiceImpl.cpp
florian-kuebler: Yes, I was referring to `CaptureServiceImpl`. Also not sure if we should go this route. Could mock away the gRPC stuff here?
Also feel free, to just ignore that comment.

OrbitService/CaptureServiceImpl.cpp
florian-kuebler: A comment should probably be fine, but do you really think a method makes it harder to read? Would be that method called something like `StopHandlerAndListeners`? In that case, I'd argue that it is better to read.

OrbitService/CaptureServiceImpl.cpp
florian-kuebler: Okay, make sense. Then there should be probably a comment for that on the stop part.

OrbitService/CaptureStartStopListener.h
dpallotti: Rearranged a bit.

OrbitService/CaptureServiceImpl.cpp
dpallotti: We should consider that but it's out of scope here. I guess `grpc::ServerReaderWriter<CaptureResponse, CaptureRequest>` could be mocked and then `Capture` can be called locally. `LinuxTracingHandler` would also need to be mocked. Maybe it can also inherit from `CaptureStartStopListener`.

OrbitService/CaptureServiceImpl.cpp
dpallotti: Alright, done. But the comment is still needed as it explains something else.

OrbitService/CaptureServiceImpl.cpp
dpallotti: Yes, that's there.

OrbitService/CaptureServiceImpl.cpp
florian-kuebler: Agree. Should we create a bug for this?

OrbitService/CaptureServiceImpl.cpp
```diff
@@ -119,6 +119,34 @@ class GrpcCaptureEventSender final : public CaptureEventSender {
...
 
+static void StopTracingHandlerAndCaptureStartStopListenersInParallel(
+    LinuxTracingHandler* tracing_handler,
+    absl::flat_hash_set<CaptureStartStopListener*>* capture_start_stop_listeners) {
+  // LinuxTracingHandler::Stop is blocking, until all perf_event_open events have been processed
+  // and all perf_event_open file descriptors have been closed.
+  // CaptureStartStopListener::OnCaptureStopRequested is also to be assumed blocking,
+  // for example until all CaptureEvents from external producers have been received.
+  // Hence why these methods need to be called in parallel on different threads.
+
```
florian-kuebler, line 130: Either remove the empty line here, or move the comment in front of the function.

florian-kuebler (approved):
LGTM, thanks!

OrbitService/CaptureServiceImpl.cpp
dpallotti: Ok, I can see you don't like vertical space.

florian-kuebler: Hmm, wouldn't say so. But it feels strange to me to have comments not attached to something concretely (either statements, methods, classes,...).

OrbitService/CaptureServiceImpl.cpp
dpallotti: http://b/174145842

florian-kuebler: Thanks!

PR #1489, opened 2020-11-24
Add ProcessItemModel class

antonrohr:
ProcessItemModel is based on ProcessInfo. Tests are included.

This is the same PR as before, but the branch of the one before was accidentally pushed to origin. This is now a PR from my fork. 

@beckerhe I amended the comment, ptal


beckerhe (approved):
Thanks Anton. Great! LGTM!

PR #1490, opened 2020-11-24
Add OverlayWidget class

antonrohr:
This class is an overlay above any other large enough widget
(e.g. a table). The overlay provides an animation to signal to the user
that something is happening an the program is not frozen. It also provides
a status message field and a button to cancel. Whoever uses the
OverlayWidget has to make sure to react to the "Cancelled" signal, and
shutdown the overlay afterwards (either setVisible(false) or remove all
together)

I do not know if having a unit test for this class makes sense. I did not add one for now.

### DIFF
--- a/OrbitQt/OverlayWidget.cpp
+++ b/OrbitQt/OverlayWidget.cpp
@@ -1,0 +1,38 @@
+// Copyright (c) 2020 The Orbit Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#include "OverlayWidget.h"
+
+namespace {
+
+// This color is used as a "background" for the overlay. The alpha value of 128 makes it
+// transparent
+const QColor kOverlayShadeColor{100, 100, 100, 128};
+
+}  // namespace
+
+namespace orbit_qt {
+
+OverlayWidget::OverlayWidget(QWidget* parent)
+    : QWidget(parent), ui_(std::make_unique<Ui::OverlayWidget>()) {
+  CHECK(parent != nullptr);
+  parent->installEventFilter(this);
+  ui_->setupUi(this);
+  ui_->cancelButton->setEnabled(true);
+
+  QObject::connect(ui_->cancelButton, &QPushButton::clicked, this, [this] { emit Cancelled(); });
+}
+
+void OverlayWidget::paintEvent(QPaintEvent* /*event*/) {
+  QPainter(this).fillRect(rect(), kOverlayShadeColor);
+}
+
+bool OverlayWidget::eventFilter(QObject* obj, QEvent* event) {
+  if (obj == parent() && event->type() == QEvent::Resize) {
+    resize(parentWidget()->size());
+  }
+  return false;
+}
+
+}  // namespace orbit_qt
--- a/OrbitQt/OverlayWidget.h
+++ b/OrbitQt/OverlayWidget.h
@@ -1,0 +1,53 @@
+// Copyright (c) 2020 The Orbit Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#ifndef ORBIT_QT_OVERLAY_WIDGET_H_
+#define ORBIT_QT_OVERLAY_WIDGET_H_
+
+#include <QEvent>
+#include <QObject>
+#include <QPainter>
+#include <QWidget>
+#include <memory>
+
+#include "OrbitBase/Logging.h"
+#include "ui_OverlayWidget.h"
+
+namespace orbit_qt {
+
+class OverlayWidget : public QWidget {
+  Q_OBJECT
+  Q_PROPERTY(bool spinning READ IsSpinning WRITE SetSpinning)
+  Q_PROPERTY(bool cancelable READ IsCancelable WRITE SetCancelable);
+  Q_PROPERTY(QString statusMessage READ GetStatusMessage WRITE SetStatusMessage)
+  Q_PROPERTY(QString buttonMessage READ GetButtonMessage WRITE SetButtonMessage);
+
+ public:
+  explicit OverlayWidget(QWidget* parent);
+
+ protected:
+  void paintEvent(QPaintEvent* /*event*/) override;
+  bool eventFilter(QObject* obj, QEvent* event) override;
+
+ private slots:
+  void SetSpinning(bool value) { ui_->progressBar->setVisible(value); }
+  void SetCancelable(bool value) { ui_->cancelButton->setVisible(value); }
+  void SetStatusMessage(const QString& message) { ui_->messageLabel->setText(message); }
+  void SetButtonMessage(const QString& message) { ui_->cancelButton->setText(message); }
+
+ signals:
+  void Cancelled();
+
+ private:
+  std::unique_ptr<Ui::OverlayWidget> ui_;
+
+  [[nodiscard]] bool IsSpinning() const { return ui_->progressBar->isVisible(); }
+  [[nodiscard]] bool IsCancelable() const { return ui_->cancelButton->isVisible(); }
+  [[nodiscard]] QString GetStatusMessage() const { return ui_->messageLabel->text(); }
+  [[nodiscard]] QString GetButtonMessage() const { return ui_->cancelButton->text(); }
+};
+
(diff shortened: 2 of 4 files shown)


OrbitQt/OverlayWidget.cpp
```diff
@@ -0,0 +1,33 @@
...
+OverlayWidget::OverlayWidget(QWidget* parent)
+    : QWidget(parent), parent_(parent), ui_(std::make_unique<Ui::OverlayWidget>()) {
+  CHECK(parent != nullptr);
+  parent->installEventFilter(this);
+  ui_->setupUi(this);
+  ui_->cancelButton->setEnabled(true);
+}
+
+void OverlayWidget::paintEvent(QPaintEvent* /*event*/) {
+  QPainter(this).fillRect(rect(), {100, 100, 100, 128});
```
beckerhe, line 18: Where does this value come from? It's a color isn't it? Could you put that into a const variable?

OrbitQt/OverlayWidget.cpp
```diff
@@ -0,0 +1,33 @@
...
+bool OverlayWidget::eventFilter(QObject* obj, QEvent* event) {
+  if (!obj->isWidgetType()) return false;
+
+  auto* widget = dynamic_cast<QWidget*>(obj);
+  CHECK(widget == parent_);
+
+  if (event->type() == QEvent::Resize) {
+    resize(parent_->size());
+  }
+  return false;
```
beckerhe, line 30: This feels a bit confusing to me. This never filters anything, does it? So this function just acts as a hooking point, doesn't it?

What about:

```
if (obj == parent() && event->type() == QEvent::Resize) {
  resize(parentWidget()->size());
}

return QWidget::eventFilter(obj, event);
```

OrbitQt/OverlayWidget.h
```diff
@@ -0,0 +1,55 @@
...
+
+ public:
+  explicit OverlayWidget(QWidget* parent);
+
+ protected:
+  void paintEvent(QPaintEvent* /*event*/) override;
+  bool eventFilter(QObject* obj, QEvent* event) override;
+
+ private slots:
+  void on_cancelButton_clicked() { emit Cancelled(); }
```
beckerhe, line 34: Personal preference: I don't like slots connected by name because the compiler can't check them. You will only notice at runtime that something changed. Maybe the name of "cancelButton". I would prefer to use QObject::connect and a lambda in the constructor of OverlayWidget, especially since the lambda will contain "emit Cancelled();".

OrbitQt/OverlayWidget.h
```diff
@@ -0,0 +1,55 @@
...
+  void SetSpinning(bool value) { ui_->progressBar->setVisible(value); }
+  void SetCancelable(bool value) { ui_->cancelButton->setVisible(value); }
+  void SetStatusMessage(QString message) { ui_->messageLabel->setText(message); }
+  void SetButtonMessage(QString message) { ui_->cancelButton->setText(message); }
+
+ signals:
+  void Cancelled();
+
+ private:
+  QWidget* parent_;
```
beckerhe, line 44: parent_ is like parent() but a QWidget*? What about parentWidget()?

OrbitQt/OverlayWidget.h
```diff
@@ -0,0 +1,55 @@
...
+  std::unique_ptr<Ui::OverlayWidget> ui_;
+
+  bool IsSpinning() { return ui_->progressBar->isVisible(); }
+  bool IsCancelable() { return ui_->cancelButton->isVisible(); }
+  QString GetStatusMessage() { return ui_->messageLabel->text(); }
+  QString GetButtonMessage() { return ui_->cancelButton->text(); }
```
beckerhe, line 50: const

OrbitQt/OverlayWidget.h
```diff
@@ -0,0 +1,55 @@
...
+
+class OverlayWidget : public QWidget {
+  Q_OBJECT
+  Q_PROPERTY(bool spinning READ IsSpinning WRITE SetSpinning)
+  Q_PROPERTY(bool cancelable READ IsCancelable WRITE SetCancelable);
+  Q_PROPERTY(QString statusMessage READ GetStatusMessage WRITE SetStatusMessage)
+  Q_PROPERTY(QString buttonMessage READ GetButtonMessage WRITE SetButtonMessage);
+
+ public:
+  explicit OverlayWidget(QWidget* parent);
```
beckerhe, line 27: It's best practive in Qt-code to have a default argument for the parent: `QWidget* parent = nullptr`

OrbitQt/OverlayWidget.h
```diff
@@ -0,0 +1,55 @@
...
+#include <QEvent>
+#include <QObject>
+#include <QPainter>
+#include <QWidget>
+#include <memory>
+
+#include "OrbitBase/Logging.h"
+#include "ui_OverlayWidget.h"
+
+namespace OrbitQt {
```
beckerhe, line 17: I think we decided to go with the `orbit_qt` namespace.

OrbitQt/OverlayWidget.h
beckerhe: But we have these in several places in the codebase, so feel free to leave it here as well.

beckerhe (commented):
Great. I added some comments. Nothing serious though. :+1:

OrbitQt/OverlayWidget.h
antonrohr: Yes, but I did this here on purpose to indicate that this OverlayWidget needs a parent. The definition of this constructor also contains a CHECK(parent != nullptr);

OrbitQt/OverlayWidget.h
antonrohr: No, your argumentation is very sound. I agree. 

I was wondering if there is a better way to accomplish what I've done here. Its basically just re-emiting a different signal. I guess `QObject::connect()` is the best way.

OrbitQt/OverlayWidget.h
antonrohr: done

OrbitQt/OverlayWidget.h
antonrohr: done

OrbitQt/OverlayWidget.h
antonrohr: Good idea. I used parentWidget()

OrbitQt/OverlayWidget.cpp
antonrohr: Yes, it does not filter anything, but just resizes when the parent resizes. I changed the code to almost what you suggested, ptal

OrbitQt/OverlayWidget.cpp
antonrohr: It comes from thin air ;) I added a constant for it

OrbitQt/OverlayWidget.h
antonrohr: done

OrbitQt/OverlayWidget.h
antonrohr: done

OrbitQt/OverlayWidget.h
beckerhe: Alright. Makes sense!

OrbitQt/OverlayWidget.cpp
beckerhe: Looks good!

OrbitQt/OverlayWidget.h
beckerhe: Yes, re-emitting is the best we can do here! Thanks!

beckerhe (approved):
LGTM!

OrbitQt/OverlayWidget.h
```diff
@@ -0,0 +1,53 @@
...
+#include <QPainter>
+#include <QWidget>
+#include <memory>
+
+#include "OrbitBase/Logging.h"
+#include "ui_OverlayWidget.h"
+
+namespace orbit_qt {
+
+class OverlayWidget : public QWidget {
```
dimitry-, line 19: Add comment for the class, and maybe how is it supposed to be used

dimitry-: (you can probably copy most of it from the commit decription)

PR #1492, opened 2020-11-24
Draw only in negative y-coordinates on Capture Window

karupayun:
Solving 2 issues.
First b/173575460, a regression in 1.55 about wrong y-initialization when loading
a capture (probably from https://github.com/google/orbit/commit/1a58be3766c411bb2d143872eb9b95299db8f209).

Starting orbit by loading a capture:
Screen Shot 2020-11-24 at 18 16 33

Second, an issue with vertical zooming where you stop seeing the top
part of the scheduler track because it is on positive y-coordinates.

Orbit after zooming-in a lot:
Screen Shot 2020-11-24 at 18 16 51

### DIFF
--- a/OrbitGl/TimeGraph.cpp
+++ b/OrbitGl/TimeGraph.cpp
@@ -1099,7 +1099,8 @@ int TimeGraph::FindMovingTrackIndex() {
 }
 
 void TimeGraph::UpdateTracks(uint64_t min_tick, uint64_t max_tick, PickingMode picking_mode) {
-  float current_y = -layout_.GetSchedulerTrackOffset();
+  // Make sure track tab fits in the viewport.
+  float current_y = -layout_.GetSchedulerTrackOffset() - layout_.GetTrackTabHeight();
   float pinned_tracks_height = 0.f;
 
   // Draw pinned tracks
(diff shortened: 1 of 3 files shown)


OrbitGl/TimeGraph.cpp
```diff
@@ -1099,7 +1099,9 @@ int TimeGraph::FindMovingTrackIndex() {
 }
 
 void TimeGraph::UpdateTracks(uint64_t min_tick, uint64_t max_tick, PickingMode picking_mode) {
-  float current_y = -layout_.GetSchedulerTrackOffset();
+  // As we draw the track tab on top of current_y coordinate, so we need this extra space to not
+  // draw anything on positive y-coordinates.
```
pierricgimmig, line 1103: nit. The wording is weird here. Either remove "so" or use a simpler expression: 

// Make sure track tab fits in the viewport. 

I'd go with the latter.

pierricgimmig (approved):
Thanks Pablo!

OrbitGl/TimeGraph.cpp
karupayun: Done! Thanks :)

PR #1493, opened 2020-11-24
Manual instrumentation bug fixes

pierricgimmig:
- Include parameters when searching for orbit_api functions to hook
  into to  make sure we are hooking the stubs and not the compiled
  implementation of those functions (b/172988899)
- Remove force inlining of non-stub functions as it is not required
  anymore. (Also fixes warning with GCC)
- Include Orbit.h in OrbitTest.cpp, not Tracing.h. (OrbitTest needs to use the api directly)

Tests: 
- Profiling OrbitTest produces expected manual instrumentation scopes
- Profiling OrbitService doesn't generate garbled manual instrumentation scopes

### DIFF
--- a/OrbitClientData/FunctionUtils.cpp
+++ b/OrbitClientData/FunctionUtils.cpp
@@ -55,11 +55,11 @@ std::unique_ptr<FunctionInfo> CreateFunctionInfo(const SymbolInfo& symbol_info,
 
 const absl::flat_hash_map<const char*, FunctionInfo::OrbitType>& GetFunctionNameToOrbitTypeMap() {
   static absl::flat_hash_map<const char*, FunctionInfo::OrbitType> function_name_to_type_map{
-      {"Start" STUB_PARAMS, FunctionInfo::kOrbitTimerStart},
-      {"Stop" STUB_PARAMS, FunctionInfo::kOrbitTimerStop},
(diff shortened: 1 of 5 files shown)


pierricgimmig:
@ronaldfw , @antonrohr, I'm wondering if you got a notification for the review.  The PR used to be a draft but was converted a few days ago.

OrbitClientData/FunctionUtils.cpp
```diff
@@ -50,13 +50,16 @@ std::unique_ptr<FunctionInfo> CreateFunctionInfo(const SymbolInfo& symbol_info,
   return function_info;
 }
 
+#define STUB_PARAMS \
+  "(unsigned long, unsigned long, unsigned long, unsigned long, unsigned long, unsigned long)"
```
antonrohr, line 54: Cant this be solved without a macro? 
I feel less macros is better than more macros.. but I am unsure

antonrohr (approved):
Hey @pierricgimmig, sorry that this slipped through the net. Looks good, only one nitpick

OrbitClientData/FunctionUtils.cpp
pierricgimmig: Yeah I agree, although concatenating string literals happen at compile time... I did change it to make the code cleaner.

pierricgimmig:
> Hey @pierricgimmig, sorry that this slipped through the net. Looks good, only one nitpick

No worries, thanks for the review!

PR #1495, opened 2020-11-25
Add Grpc Protos used for the Vulkan Layer.

florian-kuebler:
This adds a CaptureEvent proto for a Vulkan submission, which gathers
information from a single submission (VkQueueSubmit). In particular
it contains the timings for all the command buffers in that submission
and the debug markers opend/closed in that submission.

These protos can be then used by the layer to send information to the
service.

Test: Compile

### DIFF
--- a/OrbitGrpcProtos/capture.proto
+++ b/OrbitGrpcProtos/capture.proto
@@ -128,6 +128,48 @@ message GpuJob {
   uint64 dma_fence_signaled_time_ns = 11;
 }
 
+message GpuQueueSubmission {
+  GpuQueueSubmissionMetaInfo meta_info = 1;
+  repeated GpuSubmitInfo submit_infos = 2;
+  repeated GpuDebugMarker completed_markers = 3;
+  int32 num_begin_markers = 4;
+}
+
+message GpuQueueSubmissionMetaInfo {
+  int32 tid = 1;
+  uint64 pre_submission_cpu_timestamp = 2;
+  uint64 post_submission_cpu_timestamp = 3;
+}
+
+message GpuSubmitInfo {
+  repeated GpuCommandBuffer command_buffers = 1;
+}
+
+message GpuCommandBuffer {
+  uint64 begin_gpu_timestamp_ns = 1;
+  uint64 end_gpu_timestamp_ns = 2;
+}
+
+message GpuDebugMarkerBeginInfo {
+  GpuQueueSubmissionMetaInfo meta_info = 1;
+  uint64 gpu_timestamp_ns = 2;
+}
+
+message GpuDebugMarker {
+  GpuDebugMarkerBeginInfo begin_marker = 1;
+  uint64 end_gpu_timestamp_ns = 2;
+  uint64 text_key = 3;
+  int32 depth = 4;
+  Color color = 5;
+}
+
+message Color {
+  float red = 1;
+  float green = 2;
+  float blue = 3;
+  float alpha = 4;
+}
+
 message ThreadName {
   int32 pid = 1;  // Note that pid is currently not set.
   int32 tid = 2;
@@ -175,50 +217,6 @@ message AddressInfo {
     string map_name = 5;
     uint64 map_name_key = 6;
   }
-}
-
-message GpuQueueSubmission {
-  GpuQueueSubmissionMetaInfo meta_info = 1;
-  repeated GpuSubmitInfo submit_infos = 2;
-  repeated GpuDebugMarker completed_markers = 3;
-  uint32 num_begin_markers = 4;
-}
-
-message GpuQueueSubmissionMetaInfo {
-  int32 thread_id = 1;
-  uint64 pre_submission_cpu_timestamp = 2;
-  uint64 post_submission_cpu_timestamp = 3;
-}
-
-message GpuSubmitInfo {
-  repeated GpuCommandBuffer command_buffers = 1;
-}
-
-message GpuCommandBuffer {
-  uint64 begin_gpu_timestamp_ns = 1;
-  uint64 end_gpu_timestamp_ns = 2;
-}
-
-message GpuDebugMarkerBeginInfo {
-  GpuQueueSubmissionMetaInfo meta_info = 1;
-  uint64 gpu_timestamp_ns = 2;
-}
-
-message GpuDebugMarker {
-  oneof begin_marker_option {
-    GpuDebugMarkerBeginInfo begin_marker = 1;
-  };
-  uint64 end_gpu_timestamp_ns = 2;
-  uint64 text_key = 3;
-  uint64 depth = 4;
-  Color color = 5;
-}
-
-message Color {
-  float red = 1;
-  float green = 2;
-  float blue = 3;
-  float alpha = 4;
 }
 
 message CaptureEvent {
--- a/OrbitGrpcProtos/capture.proto
+++ b/OrbitGrpcProtos/capture.proto
@@ -227,12 +227,12 @@ message CaptureEvent {
     FunctionCall function_call = 4;
     InternedString interned_string = 5;
     GpuJob gpu_job = 6;
+    GpuQueueSubmission gpu_queue_submission = 13;
     ThreadName thread_name = 7;
     ThreadStateSlice thread_state_slice = 11;
     AddressInfo address_info = 8;
     InternedTracepointInfo interned_tracepoint_info = 9;
     TracepointEvent tracepoint_event = 10;
     IntrospectionScope introspection_scope = 12;
-    GpuQueueSubmission gpu_queue_submission = 13;
   }
 }
--- a/OrbitGrpcProtos/capture.proto
+++ b/OrbitGrpcProtos/capture.proto
@@ -104,7 +104,7 @@ message InternedTracepointInfo {
 message TracepointEvent {
   int32 pid = 1;
   int32 tid = 2;
-  int64 time = 3;
+  uint64 time = 3;
   int32 cpu = 4;
   oneof tracepoint_info_or_key {
     TracepointInfo tracepoint_info = 5;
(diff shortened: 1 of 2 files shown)


OrbitGrpcProtos/capture.proto
```diff
@@ -191,5 +235,6 @@ message CaptureEvent {
     InternedTracepointInfo interned_tracepoint_info = 9;
     TracepointEvent tracepoint_event = 10;
     IntrospectionScope introspection_scope = 12;
+    GpuQueueSubmission gpu_queue_submission = 13;
```
dpallotti, line 238: I would suggest moving it right after `GpuJob`, it makes sense to me if they are next to each other.

OrbitGrpcProtos/capture.proto
```diff
@@ -177,6 +177,50 @@ message AddressInfo {
...
+
+message GpuDebugMarkerBeginInfo {
+  GpuQueueSubmissionMetaInfo meta_info = 1;
+  uint64 gpu_timestamp_ns = 2;
+}
+
+message GpuDebugMarker {
+  oneof begin_marker_option {
+    GpuDebugMarkerBeginInfo begin_marker = 1;
+  };
```
dpallotti, line 210: Why `oneof` when you have only one option?

OrbitGrpcProtos/capture.proto
```diff
@@ -177,6 +177,50 @@ message AddressInfo {
...
 
+message GpuQueueSubmission {
+  GpuQueueSubmissionMetaInfo meta_info = 1;
+  repeated GpuSubmitInfo submit_infos = 2;
+  repeated GpuDebugMarker completed_markers = 3;
+  uint32 num_begin_markers = 4;
+}
+
+message GpuQueueSubmissionMetaInfo {
+  int32 thread_id = 1;
```
dpallotti, line 188: All the other protos here call it `tid`, I would suggest consistency.

OrbitGrpcProtos/capture.proto
```diff
@@ -177,6 +177,50 @@ message AddressInfo {
...
+  uint64 gpu_timestamp_ns = 2;
+}
+
+message GpuDebugMarker {
+  oneof begin_marker_option {
+    GpuDebugMarkerBeginInfo begin_marker = 1;
+  };
+  uint64 end_gpu_timestamp_ns = 2;
+  uint64 text_key = 3;
+  uint64 depth = 4;
```
dpallotti, line 213: This should probably be `int32`, similarly to `FunctionCall`, unless you have a specific reason why you need it unsigned (and 64 bits).

OrbitGrpcProtos/capture.proto
```diff
@@ -177,6 +177,50 @@ message AddressInfo {
   }
 }
 
+message GpuQueueSubmission {
+  GpuQueueSubmissionMetaInfo meta_info = 1;
+  repeated GpuSubmitInfo submit_infos = 2;
+  repeated GpuDebugMarker completed_markers = 3;
+  uint32 num_begin_markers = 4;
```
dpallotti, line 184: This should probably be `int32` unless there's a specific reason for it to be unsigned.

OrbitGrpcProtos/capture.proto
florian-kuebler: That is an purely optional field. Unfortunately, the version of ProtoBuf that we are using, does not support `optional`.

dpallotti: That's not the point of `oneof`. In `proto3` all fields are already optional.

florian-kuebler: But then I can't check if that field is there.
https://stackoverflow.com/questions/42622015/how-to-define-an-optional-field-in-protobuf-3/42634681

florian-kuebler: Wait, I was wrong. That was/is the case for primitives. At the beginning, that field should be a primitive. For the others, there are also the ` has` methods.

dpallotti: I have to give it to you that it's a bit of a mess for primitives, I forgot about that. But yes, in this case there should be a `has_begin_marker`.

dpallotti: Maybe if you want to make it clearer just from the name you could rename `begin_marker` to `optional_begin_marker`, but to me that's not necessary.

florian-kuebler: Changed it accordninly.

OrbitGrpcProtos/capture.proto
florian-kuebler: Done.

OrbitGrpcProtos/capture.proto
florian-kuebler: Done.

OrbitGrpcProtos/capture.proto
florian-kuebler: Done.

OrbitGrpcProtos/capture.proto
florian-kuebler: Done.

OrbitGrpcProtos/capture.proto
florian-kuebler: > Maybe if you want to make it clearer just from the name you could rename begin_marker to optional_begin_marker, but to me that's not necessary.

I think it is fine as it is.

OrbitGrpcProtos/capture.proto
dpallotti: I meant also here in `message CaptureEvent` (even more now that the order of definition of the messages is like that).

dpallotti (approved):

OrbitGrpcProtos/capture.proto
```diff
@@ -104,7 +104,7 @@ message InternedTracepointInfo {
 message TracepointEvent {
   int32 pid = 1;
   int32 tid = 2;
-  int64 time = 3;
+  uint64 time = 3;
```
ronaldfw, line 107: Why this change? Seems to be unrelated to the rest?

OrbitGrpcProtos/capture.proto
```diff
@@ -128,6 +128,48 @@ message GpuJob {
   uint64 dma_fence_signaled_time_ns = 11;
 }
 
+message GpuQueueSubmission {
+  GpuQueueSubmissionMetaInfo meta_info = 1;
+  repeated GpuSubmitInfo submit_infos = 2;
+  repeated GpuDebugMarker completed_markers = 3;
+  int32 num_begin_markers = 4;
```
ronaldfw, line 135: Is this the total number of markers begun or the remaining open markers?

OrbitGrpcProtos/capture.proto
```diff
@@ -128,6 +128,48 @@ message GpuJob {
...
+  repeated GpuCommandBuffer command_buffers = 1;
+}
+
+message GpuCommandBuffer {
+  uint64 begin_gpu_timestamp_ns = 1;
+  uint64 end_gpu_timestamp_ns = 2;
+}
+
+message GpuDebugMarkerBeginInfo {
+  GpuQueueSubmissionMetaInfo meta_info = 1;
```
ronaldfw, line 154: It doesn't seem to be obvious why this message has a GpuQueueSubmissionMetaInfo field. Is this only for convenience and has the same data as the GpuQueueSubmission parent or is this different?

OrbitGrpcProtos/capture.proto
ronaldfw: Independent of the answer, consider adding a comment in the code to clarify.

OrbitGrpcProtos/capture.proto
ronaldfw: Same here, add a comment in the code to clarify why this is needed.

OrbitGrpcProtos/capture.proto
florian-kuebler: Hah, good point. I think that is a left over from extracting the change from my branch. I'd actually like to keep it in, as this might clearly be an issue, though.

OrbitGrpcProtos/capture.proto
florian-kuebler: that is the total number of begin markers in that submission.

OrbitGrpcProtos/capture.proto
florian-kuebler: That differs, if the BeginMarkers originate from a different submission.

OrbitGrpcProtos/capture.proto
ronaldfw: I would not leave changes in a PR with a different purpose, in particular if they have the potential to break things.

OrbitGrpcProtos/capture.proto
ronaldfw: Ok, that should be explained in a code comment.

OrbitGrpcProtos/capture.proto
```diff
@@ -128,6 +128,48 @@ message GpuJob {
...
+message GpuDebugMarker {
+  GpuDebugMarkerBeginInfo begin_marker = 1;
+  uint64 end_gpu_timestamp_ns = 2;
+  uint64 text_key = 3;
+  int32 depth = 4;
+  Color color = 5;
+}
+
+message Color {
+  float red = 1;
```
ronaldfw, line 167: Clarify the range of the numbers in a comment. Probably [0.f, 1.f], but I've also seen [0.f, 255.f] being used (whether that is good practice is a different matter).

OrbitGrpcProtos/capture.proto
florian-kuebler: Done.

OrbitGrpcProtos/capture.proto
florian-kuebler: Okay, agree.

florian-kuebler: Removed it.

OrbitGrpcProtos/capture.proto
florian-kuebler: Done.

florian-kuebler:
@ronaldfw is there anything else from your side to add?

PR #1496, opened 2020-11-25
Add Vulkan Layer json file.

florian-kuebler:
This adds a json describtion file for the vulkan layer.

### DIFF
--- a/OrbitVulkanLayer/resources/VkLayer_Orbit_implicit.json
+++ b/OrbitVulkanLayer/resources/VkLayer_Orbit_implicit.json
@@ -1,0 +1,18 @@
+{ 
+    "file_format_version" : "1.1.0", 
+    "layer" : { 
+        "name": "ORBIT_VK_LAYER", 
+        "type": "GLOBAL", 
+        "library_path": "./libOrbitVulkanLayer.so",
+        "api_version": "1.1.77", 
+        "implementation_version": "1", 
+        "description": "Provides GPU insights for the Orbit Profiler",
+        "disable_environment": {
+            "DISABLE_ORBIT_VULKAN_LAYER": "1"
+        },
+        "functions": {
+            "vkGetInstanceProcAddr": "OrbitGetInstanceProcAddr",
+            "vkGetDeviceProcAddr": "OrbitGetDeviceProcAddr"
+        }  
+    } 
+} 


OrbitVulkanLayer/resources/VkLayer_Orbit_implicit.json
```diff
@@ -0,0 +1,18 @@
+{ 
+    "file_format_version" : "1.1.0", 
+    "layer" : { 
+        "name": "ORBIT_VK_LAYER", 
+        "type": "GLOBAL", 
+        "library_path": "./libOrbitVulkanLayer.so",
+        "api_version": "1.1.77",
```
ronaldfw, line 7: What determines that API version we have to use or we can use? (Why exactly 1.1.77?)

OrbitVulkanLayer/resources/VkLayer_Orbit_implicit.json
```diff
@@ -0,0 +1,18 @@
+{ 
+    "file_format_version" : "1.1.0", 
+    "layer" : { 
+        "name": "ORBIT_VK_LAYER", 
+        "type": "GLOBAL", 
+        "library_path": "./libOrbitVulkanLayer.so",
+        "api_version": "1.1.77", 
+        "implementation_version": "1",
```
ronaldfw, line 8: Does this have to be changed (increased) when we make changes?

OrbitVulkanLayer/resources/VkLayer_Orbit_implicit.json
```diff
@@ -0,0 +1,18 @@
+{
```
ronaldfw, line 1: Is it clear that we want the Orbit layer to be an implicit layer? Obviously that can be changed, but we should have an idea (ideally written down in our design document) what the deployment mechanism etc. will (likely) look like.

OrbitVulkanLayer/resources/VkLayer_Orbit_implicit.json
florian-kuebler: TBH, that was the version that GpuVoyeur was using.
The version basically states which API that layer is using. As all the methods we are using that are not in GpuVoyeur, were present in 1.1, I just sticked with that.

OrbitVulkanLayer/resources/VkLayer_Orbit_implicit.json
florian-kuebler: For implicit layers, that should be rather uninteresting. There is an enumeration function, that states the version of a layer. For explicit layers, games could decide not to use a certain layer in a specific version.

OrbitVulkanLayer/resources/VkLayer_Orbit_implicit.json
florian-kuebler: Yes, explicit layers need to have code on the game side, to enable them. That is nothing, that is on our radar for now. Also note, that it is totally fine for a layer to be implicit and explicit, by providing two json files.
Further, for now, that implicit layer is enabled by default,  but this is something that we can change in the future, when it comes to actual deploying. Short-term it is much more convenient to have it like this.

dpallotti: Should there be a TODO to add `enable_environment` one the time comes?

florian-kuebler: I am not sure. This is nothing that we want for sure. It depends on the way we will deploy the layer. So it will be a reasonable step as soon as we come to that point.

OrbitVulkanLayer/resources/VkLayer_Orbit_implicit.json
ronaldfw: Ok, thanks for the clarification.

OrbitVulkanLayer/resources/VkLayer_Orbit_implicit.json
ronaldfw: Ok, makes sense.

OrbitVulkanLayer/resources/VkLayer_Orbit_implicit.json
ronaldfw: Seems it is then fine to keep as is, and we can adjust if needed.

ronaldfw (approved):
LGTM, sorry for the long delay!

PR #1497, opened 2020-11-25
Fix e2e test for frame tracks due to renaming

ronaldfw:
The option to add a frame track was renamed for "Add frame track(s)" to
"Enable frame track(s)" as it is now possible to enable frame tracks
before taking a capture from the "Functions" tap and hence to have
a consistent naming in the options.

This caused the e2e test to fail, which is fixed with this change.

Bug: http://b/174214930
Tested: Ran e2e test locally with the new version.


danielfenner (approved):
lgtm!

PR #1498, opened 2020-11-25
Reduce ThreadPool.CheckGetNumberOfBusyThreads flakiness

dimitry-:
Sleep before checking that the number of busy threads is reduced to 0

Test: run the test


danielfenner (approved):

kisenshi (approved):
Oh, I didn't realize this could cause flakiness, thanks for fixing it!

PR #1499, opened 2020-11-25
Fix parsing /proc/$pid/maps

danielfenner:
It is possible to map some character or block devices. Previously these
were interpreted as candidates for modules and that lead to problems. We
now filter out every mapping with a pathname under "/dev/".

Bug: http://b/173358603
Test: Unit test & ran locally.

### DIFF
--- a/OrbitService/ServiceUtils.cpp
+++ b/OrbitService/ServiceUtils.cpp
@@ -5,6 +5,7 @@
 #include "ServiceUtils.h"
 
 #include <absl/base/casts.h>
+#include <absl/strings/match.h>
 #include <absl/strings/str_join.h>
 #include <absl/strings/str_split.h>
 #include <cxxabi.h>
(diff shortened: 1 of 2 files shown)


OrbitService/ServiceUtils.cpp
```diff
@@ -75,6 +75,9 @@ ErrorMessageOr<std::vector<ModuleInfo>> ParseMaps(std::string_view proc_maps_dat
 
     const std::string& module_path = tokens[5];
 
+    // This excludes mapped character or block devices.
+    if (module_path.rfind("/dev/", 0) == 0) continue;
```
antonrohr, line 79: I wasn't aware of this function rfind. I assume this usage is basically just checking if the prefix matches? 

I dont really care, but there is also `bool absl::StartsWith(string, prefix)`. Maybe its easier to read

antonrohr (approved):

dimitry- (approved):
otherwise lgtm

OrbitService/ServiceUtils.cpp
danielfenner: +1

PR #1500, opened 2020-11-25
Fix crash in iterators

karupayun:
When you create an iterator where the center of the screen is between the
start and the end of the last function, it produces a crash because
next_box is a null pointer. Add an if there.

I only could reproduce the issue on Windows.


ronaldfw (approved):
LGTM and thank you so much for identifying and fixing this!

PR #1503, opened 2020-11-26
Dynamic tab layout

reichlfl:
Allows users to move tabs in between the two panels.

This way, they can customize the layout at least slightly to best fit the current workflow.
Few things to note:

- Changes are **not** saved across sessions. This would be a nice addition and is in theory not that hard to do, but could introduce issues as soon as we update QT to a new version, or by changes introduced in Orbit itself, and this should be given some thought before implementing a quick-fix
- The functionality is a bit hidden behind a right-click menu on the tab bar. It would be nice if tabs could just be dragged across widgets, but this is more work and could be done in a subsequent PR.
- I've tried to retain the logic of when tabs are enabled / disabled and when Orbit automatically switches tabs (e.g. from the sampling to the sampling selection tab) as much as possible, but it's not as straightforward anymore when we don't know how the tabs are arranged
- Tutorials still work by restoring the default layout first. I've opened a bug to either improve this or to at least notify the user that this is happening before we ship the Tutorials

3 commits for easier review:
- Introduce the changes, disable the automatic tab selection
- Restore the automatic tab selection
- Add the layout restore functionality for Tutorials

image

### DIFF
--- a/OrbitQt/orbitmainwindow.cpp
+++ b/OrbitQt/orbitmainwindow.cpp
@@ -97,13 +97,11 @@ OrbitMainWindow::OrbitMainWindow(QApplication* a_App,
   finalizing_capture_dialog->close();
 
   GOrbitApp->SetCaptureStopRequestedCallback([this, finalizing_capture_dialog] {
-    capture_stop_requested_ = true;
     finalizing_capture_dialog->show();
     UpdateCaptureStateDependentWidgets();
   });
   auto capture_finished_callback = [this, finalizing_capture_dialog] {
     finalizing_capture_dialog->close();
-    capture_stop_requested_ = false;
     UpdateCaptureStateDependentWidgets();
   };
   GOrbitApp->SetCaptureStoppedCallback(capture_finished_callback);
@@ -373,7 +371,8 @@ void OrbitMainWindow::UpdateCaptureStateDependentWidgets() {
   const bool has_data = GOrbitApp->HasCaptureData();
   const bool has_selection = has_data && GOrbitApp->HasSampleSelection();
   const bool is_connected = GOrbitApp->IsConnectedToInstance();
-  const bool is_capturing = GOrbitApp->IsCapturing();
+  CaptureClient::State capture_state = GOrbitApp->GetCaptureState();
+  const bool is_capturing = capture_state != CaptureClient::State::kStopped;
 
   set_tab_enabled(ui->HomeTab, true);
   ui->HomeTab->setEnabled(!is_capturing);
@@ -387,7 +386,8 @@ void OrbitMainWindow::UpdateCaptureStateDependentWidgets() {
   set_tab_enabled(ui->selectionTopDownTab, has_selection);
   set_tab_enabled(ui->selectionBottomUpTab, has_selection);
 
-  ui->actionToggle_Capture->setEnabled(!capture_stop_requested_);
+  ui->actionToggle_Capture->setEnabled(capture_state == CaptureClient::State::kStarted ||
+                                       capture_state == CaptureClient::State::kStopped);
   ui->actionToggle_Capture->setIcon(is_capturing ? icon_stop_capture_ : icon_start_capture_);
   ui->actionClear_Capture->setEnabled(!is_capturing && has_data);
   ui->actionOpen_Capture->setEnabled(!is_capturing);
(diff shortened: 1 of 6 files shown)


OrbitQt/orbitmainwindow.cpp
```diff
@@ -105,19 +97,12 @@ OrbitMainWindow::OrbitMainWindow(QApplication* a_App,
   finalizing_capture_dialog->close();
 
   GOrbitApp->SetCaptureStopRequestedCallback([this, finalizing_capture_dialog] {
-    ui->actionToggle_Capture->setDisabled(true);
+    ui->actionToggle_Capture->setEnabled(false);
```
dpallotti, line 100: I feel like this should also be captured by `UpdateCaptureStateDependentWidgets()`.

OrbitQt/orbitmainwindow.cpp
```diff
@@ -319,6 +310,145 @@ void OrbitMainWindow::SetupCodeView() {
...
+    CHECK(tab_widget != nullptr);
+    tab_widget->setTabEnabled(tab_widget->indexOf(widget), enabled);
+  };
+
+  const bool has_data = GOrbitApp->HasCaptureData();
+  const bool has_selection = has_data && GOrbitApp->HasSampleSelection();
+  const bool is_connected = GOrbitApp->IsConnectedToInstance();
+  const bool is_capturing = GOrbitApp->IsCapturing();
+
+  SetTabEnabled(ui->HomeTab, !is_capturing);
```
dpallotti, line 373: So it's no longer possible to switch to the "home" tab altogether, right? The point of only disabling the content was to still show CPU usage while capturing. I remember a somewhat long discussion on this when we decided to disable the content of the "Home" tab only rather than the tab itself. But I guess we are moving away from that anyway.

OrbitQt/orbitmainwindow.cpp
```diff
@@ -319,6 +310,145 @@ void OrbitMainWindow::SetupCodeView() {
...
+    if (tab->isEnabled()) {
+      other_widget->setCurrentWidget(tab);
+    }
+  });
+  context_menu.addAction(&move_action);
+  context_menu.exec(pos);
+}
+
+void OrbitMainWindow::UpdateCaptureStateDependentWidgets() {
+  auto SetTabEnabled = [this](QWidget* widget, bool enabled) -> void {
```
dpallotti, line 362: This is a clang-tidy "Invalid case style for local variable". I see how the point here is to define a "local" function, but it's a local variable after all. Same for `ShowCorrespondingSelectionTab`.

dpallotti (approved):
A couple of things and maybe you also want Anton to take a look, but looks good.

OrbitQt/orbitmainwindow.cpp
reichlfl: Ah, I see, makes sense. This function needs to be changed / merged when Anton's changes come in anyways, but I've changed it to reflect the previous behaviour.

OrbitQt/orbitmainwindow.cpp
reichlfl: I felt the same, but couldn't find a way to figure out if Orbit is in that state... I've added this as a local state in `OrbitMainWindow` now, let me know if there's a better way.

OrbitQt/orbitmainwindow.cpp
reichlfl: Right, thanks. Changed both of them.

OrbitQt/orbitmainwindow.cpp
dpallotti: I think that the way to do it would be to let `OrbitApp` expose `CaptureClient::state` instead of just `CaptureClient::IsCapturing`, which you use here through `OrbitApp::IsCapturing` and which only represents `state_ != State::kStopped`. "Stop requested" should then correspond to `State::kStopping`.

If the UI can be in all the states that correspond to the `CaptureClient::State`s, to me it makes sense to expose `CaptureClient::state` directly, otherwise with `CaptureClient::IsCapturing` (and `OrbitApp::IsCapturing`) we effectively lose information, that we then need to recover in different ways, such as what you did here with then adding `capture_stop_requested_`.

reichlfl: Thanks for pointing this out, makes perfect sense. I added `GetCaptureState` to `OrbitApp`. I didn't remove `IsCapturing` at this point because it's used in more places and we have this pattern a lot through Orbit, but it could make sense to refactor these occurences in a later PR as well.

dpallotti: Great, thanks!

antonrohr (approved):
I only did a rather shallow review in the sense that I did not test it and I did not mentally explore all states the UI can be in. 
But I like the code a lot and I think its easy to fix something in the case we will find an edge case where something goes wrong.

PR #1504, opened 2020-11-26
Add definition of ProducerSideService

dpallotti:
`ProducerSideService` defines the service for the communication between
producers and OrbitService.

Bug: http://b/174028631

### DIFF
--- a/OrbitGrpcProtos/producer_side_services.proto
+++ b/OrbitGrpcProtos/producer_side_services.proto
@@ -1,0 +1,41 @@
+// Copyright (c) 2020 The Orbit Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+syntax = "proto3";
+
+package orbit_grpc_protos;
+
+import "capture.proto";
+
+message ReceiveCommandsAndSendEventsRequest {
+  message BufferedCaptureEvents {
+    repeated CaptureEvent capture_events = 1;
+  }
+  message AllEventsSent {}
+
+  oneof event {
+    BufferedCaptureEvents buffered_capture_events = 1;
+    AllEventsSent all_events_sent = 2;
+  }
+}
+
+message ReceiveCommandsAndSendEventsResponse {
+  message StartCaptureCommand {}
+  message StopCaptureCommand {}
+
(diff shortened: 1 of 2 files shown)


OrbitGrpcProtos/producer_side_services.proto
```diff
@@ -0,0 +1,41 @@
...
+
+message ReceiveCommandsAndSendEventsRequest {
+  message BufferedCaptureEvents {
+    repeated CaptureEvent capture_events = 1;
+  }
+  message AllEventsSent {}
+
+  oneof event {
+    BufferedCaptureEvents buffered_capture_events = 1;
+    AllEventsSent all_events_sent = 2;
```
dimitry-, line 19: I think it would be better if we had LastEvent as a Capture event.

dpallotti: What's the specific reason why you think that's better?

To me, that doesn't improve things here, but on the other hand makes things more confusing in all other places where we deal with `CaptureEvent`s, as they now have to deal with a new type of (fake) event they are not interested in. In particular, as that event would be the last event sent by the producer, not the last of the capture.

dimitry-: I was going to do that anyways for reading from file, where you want an indication even if events are not enclosed in another message.

dimitry-: EndOfCaptureEvent is a better name.

dpallotti: Here the point is for the single producer to notify that all the events that still needed to be sent have been sent (because of buffering) without closing the connection. It's slightly different, because what you suggest is to indicate the global end of the capture.

I can see how the two uses might and probably will converge, and that this might (or not) benefit from changing once they do, but as of now we shouldn't do something that complicates this specific change (adding special handling of a specific `CaptureEvent::event_case`) because you need it later for something else.

florian-kuebler (approved):
So to me it looks good. 
Regarding the EndOfCaptureEvent vs. the AllEventsSet message, I'd leave it to Dimitry and you, as I don't have a strong opinion on this.

dimitry- (approved):

PR #1505, opened 2020-11-26
Connect to stadia widget

antonrohr:
This is the ConnectToStadiaWidget, which will be a separate Widget in the upcoming ProfilingTargetDialog. This Widget handles the connection up to establishing a grpc channel (tunnel). 

I marked this as a draft PR, because it currently cannot be tested, because the ProfilingTargetDialog is still missing (next PR). 
If you want to you could already do a code review of this, because it will not change significantly anymore (if at all). You could also wait until the ProfilingTargetDialog PR is ready.

### DIFF
--- a/OrbitGgp/InstanceItemModel.cpp
+++ b/OrbitGgp/InstanceItemModel.cpp
@@ -4,9 +4,15 @@
 
 #include "OrbitGgp/InstanceItemModel.h"
 
-namespace {
-enum class Columns { DisplayName, ID, IPAddress, LastUpdated, Owner, Pool, NumberOfColumns };
-}  // namespace
+#include <QDateTime>
+#include <QLatin1String>
+#include <QString>
+#include <algorithm>
+#include <iterator>
+#include <utility>
+
+#include "OrbitBase/Logging.h"
+#include "OrbitGgp/Instance.h"
 
 namespace OrbitGgp {
 
@@ -16,7 +22,7 @@ InstanceItemModel::InstanceItemModel(QVector<Instance> instances, QObject* paren
 }
 
 int InstanceItemModel::columnCount(const QModelIndex& parent) const {
-  return parent.isValid() ? 0 : static_cast<int>(Columns::NumberOfColumns);
+  return parent.isValid() ? 0 : static_cast<int>(Columns::kEnd);
 }
 
 QVariant InstanceItemModel::data(const QModelIndex& index, int role) const {
@@ -31,19 +37,19 @@ QVariant InstanceItemModel::data(const QModelIndex& index, int role) const {
   if (role != Qt::DisplayRole) return {};
 
   switch (static_cast<Columns>(index.column())) {
-    case Columns::DisplayName:
+    case Columns::kDisplayName:
       return current_instance.display_name;
-    case Columns::ID:
+    case Columns::kId:
       return current_instance.id;
-    case Columns::IPAddress:
+    case Columns::kIpAddress:
       return current_instance.ip_address;
-    case Columns::LastUpdated:
+    case Columns::kLastUpdated:
       return current_instance.last_updated.toString(Qt::TextDate);
-    case Columns::Owner:
+    case Columns::kOwner:
       return current_instance.owner;
-    case Columns::Pool:
+    case Columns::kPool:
       return current_instance.pool;
-    case Columns::NumberOfColumns:
+    case Columns::kEnd:
       CHECK(false);
       return {};
   }
@@ -55,7 +61,7 @@ QVariant InstanceItemModel::data(const QModelIndex& index, int role) const {
 QModelIndex InstanceItemModel::index(int row, int col, const QModelIndex& parent) const {
   if (parent.isValid()) return {};
   if (row < 0 || row >= instances_.size()) return {};
-  if (col < 0 || col >= static_cast<int>(Columns::NumberOfColumns)) return {};
+  if (col < 0 || col >= static_cast<int>(Columns::kEnd)) return {};
 
   return createIndex(row, col, nullptr);
 }
@@ -63,25 +69,25 @@ QModelIndex InstanceItemModel::index(int row, int col, const QModelIndex& parent
 QVariant InstanceItemModel::headerData(int section, Qt::Orientation orientation, int role) const {
   if (role != Qt::DisplayRole) return {};
   if (orientation != Qt::Horizontal) return {};
-  if (section < 0 || section >= static_cast<int>(Columns::NumberOfColumns)) {
+  if (section < 0 || section >= static_cast<int>(Columns::kEnd)) {
     return {};
   }
 
   switch (static_cast<Columns>(section)) {
-    case Columns::DisplayName:
+    case Columns::kDisplayName:
       return QLatin1String("Display Name");
-    case Columns::ID:
+    case Columns::kId:
       return QLatin1String("ID");
-    case Columns::IPAddress:
+    case Columns::kIpAddress:
       return QLatin1String("IP Address");
-    case Columns::LastUpdated:
+    case Columns::kLastUpdated:
       return QLatin1String("Last Updated");
-    case Columns::Owner:
+    case Columns::kOwner:
       return QLatin1String("Owner");
-    case Columns::Pool:
+    case Columns::kPool:
       return QLatin1String("Pool");
-    case Columns::NumberOfColumns:
-      CHECK(false);
+    case Columns::kEnd:
+      UNREACHABLE();
       return {};
   }
 
@@ -110,7 +116,7 @@ void InstanceItemModel::SetInstances(QVector<Instance> new_instances) {
       if (*old_iter != *new_iter) {
         *old_iter = *new_iter;
         emit dataChanged(index(current_row, 0, {}),
-                         index(current_row, static_cast<int>(Columns::NumberOfColumns) - 1, {}));
+                         index(current_row, static_cast<int>(Columns::kEnd) - 1, {}));
       }
       ++old_iter;
       ++new_iter;
--- a/OrbitGgp/include/OrbitGgp/InstanceItemModel.h
+++ b/OrbitGgp/include/OrbitGgp/InstanceItemModel.h
@@ -10,25 +10,27 @@
 #include <QObject>
 #include <QVariant>
 #include <QVector>
+#include <Qt>
 
 #include "Instance.h"
-#include "OrbitBase/Logging.h"
 
 namespace OrbitGgp {
 
 class InstanceItemModel : public QAbstractItemModel {
  public:
+  enum class Columns { kDisplayName, kId, kIpAddress, kLastUpdated, kOwner, kPool, kEnd };
+
   explicit InstanceItemModel(QVector<Instance> instances = {}, QObject* parent = nullptr);
 
   void SetInstances(QVector<Instance> instances);
 
-  int columnCount(const QModelIndex& parent = {}) const override;
-  QVariant data(const QModelIndex& index, int role = Qt::DisplayRole) const override;
-  QVariant headerData(int section, Qt::Orientation orientation,
-                      int role = Qt::DisplayRole) const override;
-  QModelIndex index(int row, int col, const QModelIndex& parent = {}) const override;
-  QModelIndex parent(const QModelIndex& parent) const override;
-  int rowCount(const QModelIndex& parent = {}) const override;
+  [[nodiscard]] int columnCount(const QModelIndex& parent = {}) const override;
+  [[nodiscard]] QVariant data(const QModelIndex& index, int role = Qt::DisplayRole) const override;
+  [[nodiscard]] QVariant headerData(int section, Qt::Orientation orientation,
+                                    int role = Qt::DisplayRole) const override;
+  [[nodiscard]] QModelIndex index(int row, int col, const QModelIndex& parent = {}) const override;
+  [[nodiscard]] QModelIndex parent(const QModelIndex& parent) const override;
+  [[nodiscard]] int rowCount(const QModelIndex& parent = {}) const override;
 
  private:
   QVector<Instance> instances_;
--- a/OrbitQt/ConnectToStadiaWidget.cpp
+++ b/OrbitQt/ConnectToStadiaWidget.cpp
@@ -1,0 +1,422 @@
+// Copyright (c) 2020 The Orbit Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#include "ConnectToStadiaWidget.h"
+
+#include <absl/strings/str_format.h>
+#include <grpcpp/create_channel.h>
+#include <grpcpp/security/credentials.h>
+#include <grpcpp/support/channel_arguments.h>
+
+#include <QAbstractButton>
+#include <QAbstractItemView>
+#include <QAbstractState>
+#include <QApplication>
+#include <QCheckBox>
+#include <QHBoxLayout>
+#include <QItemSelectionModel>
+#include <QLayout>
+#include <QMessageBox>
+#include <QModelIndexList>
+#include <QPushButton>
+#include <QRadioButton>
+#include <QSettings>
+#include <QTableView>
+#include <QVariant>
+#include <Qt>
+#include <filesystem>
+#include <memory>
+#include <optional>
+#include <system_error>
+#include <utility>
+
+#include "Error.h"
+#include "OrbitBase/Logging.h"
+#include "OrbitBase/Result.h"
+#include "OrbitGgp/Client.h"
+#include "OrbitGgp/InstanceItemModel.h"
+#include "OrbitSsh/AddrAndPort.h"
+#include "OrbitSshQt/ScopedConnection.h"
+#include "OverlayWidget.h"
+#include "servicedeploymanager.h"
+
+namespace {
+const QString kRememberChosenInstance{"RememberChosenInstance"};
+}  // namespace
+
+namespace orbit_qt {
+
+using OrbitGgp::Instance;
+using OrbitSshQt::ScopedConnection;
+
+ConnectToStadiaWidget::ConnectToStadiaWidget(QWidget* parent)
+    : QWidget(parent),
+      ui_(std::make_unique<Ui::ConnectToStadiaWidget>()),
+      s_idle_(&state_machine_),
+      s_instances_loading_(&state_machine_),
+      s_instance_selected_(&state_machine_),
+      s_waiting_for_creds_(&state_machine_),
+      s_deploying_(&state_machine_),
+      s_connected_(&state_machine_) {
+  ui_->setupUi(this);
+
+  DetachRadioButton();
+
+  QSettings settings;
+  if (settings.contains(kRememberChosenInstance)) {
+    remembered_instance_id_ = settings.value(kRememberChosenInstance).toString();
+    ui_->rememberCheckBox->setChecked(true);
+  }
+
+  ui_->instancesTableView->setModel(&instance_model_);
+
+  QObject::connect(ui_->radioButton, &QRadioButton::clicked, this,
+                   &ConnectToStadiaWidget::OnConnectToStadiaRadioButtonClicked);
+  QObject::connect(this, &ConnectToStadiaWidget::ErrorOccurred, this,
+                   &ConnectToStadiaWidget::OnErrorOccurred);
+  QObject::connect(ui_->instancesTableView->selectionModel(), &QItemSelectionModel::currentChanged,
+                   this, &ConnectToStadiaWidget::OnSelectionChanged);
+  QObject::connect(ui_->rememberCheckBox, &QCheckBox::toggled, this,
+                   &ConnectToStadiaWidget::OnRememberCheckBoxToggled);
+
+  SetupStateMachine();
+}
+
+void ConnectToStadiaWidget::SetActive(bool value) {
+  ui_->contentFrame->setEnabled(value);
+  ui_->radioButton->setChecked(value);
+}
+
+void ConnectToStadiaWidget::Start(SshConnectionArtifacts* ssh_connection_artifacts,
+                                  std::optional<StadiaConnection> connection) {
+  CHECK(ssh_connection_artifacts != nullptr);
+  ssh_connection_artifacts_ = ssh_connection_artifacts;
+
+  auto client_result = OrbitGgp::Client::Create(this);
+  if (!client_result) {
+    ui_->radioButton->setToolTip(QString::fromStdString(client_result.error().message()));
+    setEnabled(false);
+    return;
+  }
+  ggp_client_ = client_result.value();
+
+  if (connection.has_value()) {
+    selected_instance_ = std::move(connection->instance_);
+    service_deploy_manager_ = std::move(connection->service_deploy_manager_);
+    grpc_channel_ = std::move(connection->grpc_channel_);
+    state_machine_.setInitialState(&s_connected_);
+  } else {
+    state_machine_.setInitialState(&s_instances_loading_);
+  }
+
+  state_machine_.start();
+}
+
+std::optional<StadiaConnection> ConnectToStadiaWidget::StopAndClearConnection() {
+  if (!selected_instance_.has_value() || service_deploy_manager_ == nullptr ||
+      grpc_channel_ == nullptr) {
+    return std::nullopt;
+  }
+
+  state_machine_.stop();
+
+  Instance local_copy = std::move(selected_instance_.value());
+  selected_instance_ = std::nullopt;
+  return StadiaConnection(std::move(local_copy), std::move(service_deploy_manager_),
+                          std::move(grpc_channel_));
+}
+
+void ConnectToStadiaWidget::DetachRadioButton() {
+  ui_->titleBarLayout->removeWidget(ui_->radioButton);
+  ui_->radioButton->setParent(ui_->mainFrame);
+  int left = 0;
+  int top = 0;
+  ui_->mainFrame->layout()->getContentsMargins(&left, &top, nullptr, nullptr);
+  int frame_border_width = ui_->mainFrame->lineWidth();
+  ui_->radioButton->move(left + frame_border_width, top + frame_border_width);
+}
+
+void ConnectToStadiaWidget::SetupStateMachine() {
+  state_machine_.setGlobalRestorePolicy(QStateMachine::RestoreProperties);
+
+  // PROPERTIES of states
+  // STATE s_idle
+  s_idle_.assignProperty(ui_->refreshButton, "enabled", true);
+  // STATE s_instances_loading
+  s_instances_loading_.assignProperty(ui_->instancesTableOverlay, "visible", true);
+  s_instances_loading_.assignProperty(ui_->instancesTableOverlay, "statusMessage",
+                                      "Loading instances...");
+  s_instances_loading_.assignProperty(ui_->instancesTableOverlay, "cancelable", false);
+  // STATE s_instance_selected
+  s_instance_selected_.assignProperty(ui_->refreshButton, "enabled", true);
+  s_instance_selected_.assignProperty(ui_->connectButton, "enabled", true);
+  // STATE s_waiting_for_creds
+  s_waiting_for_creds_.assignProperty(ui_->instancesTableOverlay, "visible", true);
+  s_waiting_for_creds_.assignProperty(ui_->instancesTableOverlay, "statusMessage",
+                                      "Loading encryption credentials for instance...");
+  s_waiting_for_creds_.assignProperty(ui_->instancesTableOverlay, "cancelable", true);
+  // STATE s_deploying
+  s_deploying_.assignProperty(ui_->instancesTableOverlay, "visible", true);
+  s_deploying_.assignProperty(ui_->instancesTableOverlay, "cancelable", true);
+  // STATE s_connected
+  s_connected_.assignProperty(ui_->instancesTableOverlay, "visible", true);
+  s_connected_.assignProperty(ui_->instancesTableOverlay, "spinning", false);
+  s_connected_.assignProperty(ui_->instancesTableOverlay, "cancelable", true);
+  s_connected_.assignProperty(ui_->instancesTableOverlay, "buttonMessage", "Disconnect");
+  s_connected_.assignProperty(ui_->rememberCheckBox, "enabled", true);
+
+  // TRANSITIONS (and entered/exit events)
+  // STATE s_idle_
+  s_idle_.addTransition(ui_->refreshButton, &QPushButton::clicked, &s_instances_loading_);
+  s_idle_.addTransition(this, &ConnectToStadiaWidget::InstanceSelected, &s_instance_selected_);
+
+  // STATE s_instances_loading_
+  QObject::connect(&s_instances_loading_, &QState::entered, this,
+                   &ConnectToStadiaWidget::ReloadInstances);
+
+  s_instances_loading_.addTransition(this, &ConnectToStadiaWidget::ErrorOccurred, &s_idle_);
+  s_instances_loading_.addTransition(this, &ConnectToStadiaWidget::ReceivedInstances, &s_idle_);
+
+  // STATE s_instance_selected_
+  s_instance_selected_.addTransition(ui_->refreshButton, &QPushButton::clicked,
+                                     &s_instances_loading_);
+  s_instance_selected_.addTransition(ui_->connectButton, &QPushButton::clicked,
+                                     &s_waiting_for_creds_);
+  s_instance_selected_.addTransition(ui_->instancesTableView, &QTableView::doubleClicked,
+                                     &s_waiting_for_creds_);
+  s_instance_selected_.addTransition(this, &ConnectToStadiaWidget::Connect, &s_waiting_for_creds_);
+
+  // STATE s_waiting_for_creds_
+  QObject::connect(&s_waiting_for_creds_, &QState::entered, this,
+                   &ConnectToStadiaWidget::CheckCredentialsAvailable);
+
+  s_waiting_for_creds_.addTransition(this, &ConnectToStadiaWidget::ReceivedSshInfo,
+                                     &s_waiting_for_creds_);
+  s_waiting_for_creds_.addTransition(this, &ConnectToStadiaWidget::ReadyToDeploy, &s_deploying_);
+  s_waiting_for_creds_.addTransition(ui_->instancesTableOverlay, &OverlayWidget::Cancelled,
+                                     &s_instance_selected_);
+  s_waiting_for_creds_.addTransition(this, &ConnectToStadiaWidget::ErrorOccurred,
+                                     &s_instances_loading_);
+
+  // STATE s_deploying_
+  QObject::connect(&s_deploying_, &QState::entered, this,
+                   &ConnectToStadiaWidget::DeployOrbitService);
+
+  s_deploying_.addTransition(this, &ConnectToStadiaWidget::ErrorOccurred, &s_instance_selected_);
+  s_deploying_.addTransition(ui_->instancesTableOverlay, &OverlayWidget::Cancelled,
+                             &s_instance_selected_);
+  s_deploying_.addTransition(this, &ConnectToStadiaWidget::Connected, &s_connected_);
+
+  // STATE s_connected_
+  QObject::connect(&s_connected_, &QState::entered, this, [this] {
+    ui_->instancesTableOverlay->SetStatusMessage(
+        QString{"Connected to %1"}.arg(selected_instance_->display_name));
+  });
+  QObject::connect(&s_connected_, &QState::exited, this, &ConnectToStadiaWidget::Disconnect);
+
+  s_connected_.addTransition(ui_->instancesTableOverlay, &OverlayWidget::Cancelled,
+                             &s_instance_selected_);
+  s_connected_.addTransition(this, &ConnectToStadiaWidget::ErrorOccurred, &s_instance_selected_);
+}
+
+void ConnectToStadiaWidget::ReloadInstances() {
+  CHECK(ggp_client_ != nullptr);
+  instance_model_.SetInstances({});
+
+  ggp_client_->GetInstancesAsync([this](outcome::result<QVector<Instance>> instances) {
+    OnInstancesLoaded(std::move(instances));
+  });
+}
+
+void ConnectToStadiaWidget::CheckCredentialsAvailable() {
+  CHECK(selected_instance_.has_value());
+
+  const std::string instance_id = selected_instance_->id.toStdString();
+
+  if (!instance_credentials_.contains(instance_id)) return;
+
+  if (instance_credentials_.at(instance_id).has_error()) {
+    emit ErrorOccurred(
+        QString::fromStdString(instance_credentials_.at(instance_id).error().message()));
+    return;
+  }
+
+  emit ReadyToDeploy();
+}
+
+void ConnectToStadiaWidget::DeployOrbitService() {
+  CHECK(service_deploy_manager_ == nullptr);
+  CHECK(selected_instance_.has_value());
+  const std::string instance_id = selected_instance_->id.toStdString();
+  CHECK(instance_credentials_.contains(instance_id));
+  CHECK(instance_credentials_.at(instance_id).has_value());
+
+  const OrbitSsh::Credentials& credentials{instance_credentials_.at(instance_id).value()};
+
+  CHECK(ssh_connection_artifacts_ != nullptr);
+  service_deploy_manager_ = std::make_unique<ServiceDeployManager>(
+      ssh_connection_artifacts_->GetDeploymentConfiguration(),
+      ssh_connection_artifacts_->GetSshContext(), credentials,
+      ssh_connection_artifacts_->GetGrpcPort());
+
+  ScopedConnection label_connection{QObject::connect(
+      service_deploy_manager_.get(), &orbit_qt::ServiceDeployManager::statusMessage,
+      ui_->instancesTableOverlay, &OverlayWidget::SetStatusMessage)};
+  ScopedConnection cancel_connection{
+      QObject::connect(ui_->instancesTableOverlay, &OverlayWidget::Cancelled,
+                       service_deploy_manager_.get(), &orbit_qt::ServiceDeployManager::Cancel)};
+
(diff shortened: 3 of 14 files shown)


OrbitQt/ConnectToStadiaWidget.h
```diff
@@ -0,0 +1,64 @@
...
+
+namespace orbit_qt {
+
+class ConnectToStadiaWidget : public QWidget {
+  Q_OBJECT
+
+ public:
+  explicit ConnectToStadiaWidget(QWidget* parent = nullptr);
+  // This needs to be called before this class can be used.
+  void SetConnectionArtifacts(ConnectionArtifacts* connection_artifacts);
```
reichlfl, line 27: Could this not be added to the constructor?

OrbitQt/ConnectToStadiaWidget.cpp
```diff
@@ -0,0 +1,344 @@
...
+      grpc_server_address, grpc::InsecureChannelCredentials(), grpc::ChannelArguments());
+  CHECK(connection_artifacts_->grpc_channel_ != nullptr);
+
+  emit Connected();
+}
+
+void ConnectToStadiaWidget::Disconnect() {
+  connection_artifacts_->grpc_channel_ = nullptr;
+
+  // TODO(antonrohr) currently does not work
```
reichlfl, line 333: Can you link a bug to the 2do please? I didn't get around to look into it yet.

OrbitQt/ConnectionArtifacts.h
```diff
@@ -0,0 +1,56 @@
...
+#include "OrbitGgp/Instance.h"
+#include "OrbitSsh/Context.h"
+#include "OrbitSsh/Credentials.h"
+#include "grpcpp/channel.h"
+#include "servicedeploymanager.h"
+
+namespace orbit_qt {
+
+// A class to store remote connection related artifacts. This is used by the ProfilingTargetDialog
+// and OrbitApp to be able to use the same connection. The
```
reichlfl, line 23: The...?

OrbitQt/ConnectionArtifacts.h
```diff
@@ -0,0 +1,56 @@
...
+    CHECK(deployment_configuration_ != nullptr);
+  }
+  void CreateServiceDeployManager(OrbitSsh::Credentials credentials) {
+    CHECK(service_deploy_manager_ == nullptr);
+    service_deploy_manager_ = std::make_unique<ServiceDeployManager>(
+        deployment_configuration_, ssh_context_, std::move(credentials), grpc_port_);
+  }
+
+  // Required Member
+  const OrbitSsh::Context* ssh_context_;
```
reichlfl, line 42: Why don't you use references instead of pointers if those are required to be non-null?

reichlfl (commented):
I review everything except `SetupAndStartStateMachine` for now, I'm submitting this nevertheless since I may not get to finish it today.

Overall, I like how it's done, especially that the Connection Artifacts are not owned by the Dialog, there's a few things I'd change though. As for the state machine, I have to admit I'm not sure I like it... especially `SetupAndStartStateMachine` is quite hard to check for correctness, and it feels a bit like overcomplicating the dependencies in the UI. But I get that it may come in handy when this is extended, so it's fine to do it like this.

OrbitQt/ConnectionArtifacts.h
beckerhe: I politely disagree. :smiley:

I think we shouldn't use references as non-nullable pointers. They have never been meant for that. I'm pretty sure there is a section about that in the C++ style guide but I haven't found it so far.

I think the root cause of this problem is that in C-APIs pointers are often considered nullable, but in C++ they shouldn't be by default. If you need something nullable, `std::optional<T*>` can express that a lot more explicit. Of course you never know what contract the API designer followed, that's why wrappers like `gsl::not_null<T*>` had been invented which can explicitly state the non-nullable case.

But that's just my opinion.

reichlfl: References vs. pointers are mentioned in the context of function parameters here:
https://google.github.io/styleguide/cppguide.html#Inputs_and_Outputs

"Input parameters should usually be values or const references, while non-optional output and input/output parameters should usually be references (which cannot be null). Generally, use absl::optional to represent optional by-value inputs, and use a const pointer when the non-optional form would have used a reference. Use non-const pointers to represent optional outputs."

IMHO this is the same case here. This member is passed in through the constructor and should be non-null when passed there. You could argue it should still be stored as a pointer, and only the constructor should be changed to a reference, which would be the case mentioned above in the style guide - fair enough, but what is the drawback of a reference here?

beckerhe: I believe the mentioned rules does not apply for constructors. The main problem I see with a reference in a constructor is that the user does not know that this reference is stored internally, so the user does not know that it's their job to keep the passed-in object alive at least as long as the constructed class exists (That's exactly what needs to happen here). A raw pointer means that the lifetime needs to be managed by the caller and they have to explicitly add a `&` when a local variable is used as an argument. That makes that contract more explicit - at least to me.

beckerhe: Quick Example:
```
class CSVParser {
  explicit CSVParser(const std::string&);
  ...
};

void foo() {
  CSVParser parser{readFromFile()};

  while(const auto& line : parser) {
    // do something
  }
}
```

If the `const std::string&` is not copied, but stored as a ref this won't work. But it feels correct to me. I would assume that the string is copied in that situation.

The other case:

```
class CSVParser {
  explicit CSVParser(const std::string*);
  ...
};

void foo() {
  const auto buf = readFromFile()
  CSVParser parser{&buf};

  while(const auto& line : parser) {
    // do something
  }
}
```

reichlfl: Ok, I can get behind that line of thought. There's a non-null CHECK in the constructor for these parameters which will catch misuse anyways, so you can ignore my comment on this.

beckerhe: We can still discuss the member variable. :smiley:

That's what I thought we would be discussing in the first place. Even if the constructor took a pointer the member variable could be a reference.

OrbitQt/ConnectToStadiaWidget.cpp
```diff
@@ -0,0 +1,344 @@
...
+  connection_artifacts_ = connection_artifacts;
+}
+
+void ConnectToStadiaWidget::SetupAndStartStateMachine() {
+  state_machine_.setGlobalRestorePolicy(QStateMachine::RestoreProperties);
+
+  // STATES list
+  auto* s_ggp_not_available = new QState();
+  state_machine_.addState(s_ggp_not_available);
+  auto* s_instances_empty = new QState();
```
reichlfl, line 88: Is `s_instances_empty` a state that is actually needed? Isn't this just a special case of `s_instances_loaded` with 0 instances? As far as I see, the state of `s_instances_loaded` without an instance selected is equal to `s_instances_empty`?

OrbitQt/ConnectToStadiaWidget.cpp
```diff
@@ -0,0 +1,344 @@
...
+  // STATE s_connected
+  s_connected->assignProperty(ui_->instancesTableOverlay, "visible", true);
+  s_connected->assignProperty(ui_->instancesTableOverlay, "spinning", false);
+  s_connected->assignProperty(ui_->instancesTableOverlay, "cancelable", true);
+  s_connected->assignProperty(ui_->instancesTableOverlay, "buttonMessage", "Disconnect");
+  s_connected->assignProperty(ui_->rememberCheckBox, "enabled", true);
+
+  // STATE s_ggp_not_available
+
+  // TRANSITIONS (and entered/xit events)
```
reichlfl, line 136: Typo: "exit"

reichlfl (commented):
Done reviewing the StateMachine, sorry for the delay! Some minor comments, but overall the state handling looks good.

OrbitQt/ConnectToStadiaWidget.cpp
```diff
@@ -0,0 +1,344 @@
...
+  QObject::connect(ui_->connectToStadiaInstanceRadioButton, &QRadioButton::toggled, this,
+                   [this](bool checked) {
+                     if (checked) {
+                       emit Selected();
+                     } else {
+                       ui_->connectToStadiaInstanceRadioButton->setChecked(true);
+                     }
+                   });
+
+  QObject::connect(this, &ConnectToStadiaWidget::ErrorOccurred, this,
```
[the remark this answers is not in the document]
antonrohr: Yes and no. The main reason is because I react in different ways when an error occurs. One of them is displaying that error (here), others are transitioning into a different state of the state machine

OrbitQt/ConnectToStadiaWidget.cpp
```diff
@@ -0,0 +1,344 @@
...
+                     if (!current.isValid()) return;
+
+                     CHECK(current.model() == &instance_model_);
+                     connection_artifacts_->selected_instance_ =
+                         current.data(Qt::UserRole).value<Instance>();
+                     emit InstanceSelected();
+                   });
+
+  QObject::connect(ui_->rememberCheckBox, &QCheckBox::toggled, this, [this](bool checked) {
+    if (checked) {
```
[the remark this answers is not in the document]
antonrohr: The `rememberCheckBox` is disabled until a connection is established. So only after a successful connection, the user can choose to remember the instance. 
When the instance disconnects (because the user did so) then this is not remembered anymore.

Let me know if that makes sense, or if another way makes more sense

OrbitQt/ConnectToStadiaWidget.h
antonrohr: No it can't, because otherwise it cant be used in a ui file. If its supposed to be instantiated via the ui file, the constructor needs to have only one parameter (QWidget*)

OrbitQt/ConnectionArtifacts.h
antonrohr: I'd like to have this discussion. The class `ConnectionArtifacts` changed in the meantime to 3 structs, but we can talk about `StadiaConnection` and the member `const OrbitSsh::Context* ssh_context` and `const DeploymentConfiguration* deployment_configuration`. 
They are both pointers, because I only use them for the constructor of ServiceDeployManager, which takes pointers (instead of references). 
Since this is the only use case of these variables, I would leave it as is unless there is a downside I don't see. But in general I like the idea of requiring a pointer to make sure the caller knows the lifetime hierarchy, and then saving a reference internally.

OrbitGgp/include/OrbitGgp/InstanceItemModel.h
```diff
@@ -30,6 +30,8 @@ class InstanceItemModel : public QAbstractItemModel {
   QModelIndex parent(const QModelIndex& parent) const override;
   int rowCount(const QModelIndex& parent = {}) const override;
 
+  std::optional<int> GetRowOfInstanceById(const QString& instance_id);
```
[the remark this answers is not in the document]
antonrohr: done

OrbitQt/ConnectToStadiaWidget.cpp
```diff
@@ -0,0 +1,344 @@
...
+
+namespace orbit_qt {
+
+using OrbitGgp::Instance;
+using OrbitSshQt::ScopedConnection;
+
+ConnectToStadiaWidget::ConnectToStadiaWidget(QWidget* parent)
+    : QWidget(parent),
+      ui_(std::make_unique<Ui::ConnectToStadiaWidget>()),
+      settings_("The Orbit Authors", "Orbit Profiler") {
```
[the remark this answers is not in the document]
antonrohr: I solved this by using `QSettings` without arguments. This uses then internally `QCoreApplication::organizationName()` and `QCoreApplication::applicationName()`, which is set in OrbitQt/main.cpp

(In the next (integration) PR, I also clear this up everywhere else)

OrbitQt/ConnectToStadiaWidget.cpp
```diff
@@ -0,0 +1,344 @@
...
+  ui_->setupUi(this);
+
+  if (!settings_.value(kRememberChosenInstance).toString().isEmpty()) {
```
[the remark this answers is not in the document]
antonrohr: thanks 👍

OrbitQt/ConnectToStadiaWidget.cpp
```diff
@@ -0,0 +1,344 @@
...
+                       ui_->connectToStadiaInstanceRadioButton->setChecked(true);
+                     }
+                   });
+
+  QObject::connect(this, &ConnectToStadiaWidget::ErrorOccurred, this,
+                   [this](const QString& message) {
+                     QMessageBox::critical(this, QApplication::applicationName(), message);
+                   });
+
+  QObject::connect(ui_->instancesTableView->selectionModel(), &QItemSelectionModel::currentChanged,
```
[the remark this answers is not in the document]
antonrohr: Done

OrbitQt/ConnectToStadiaWidget.cpp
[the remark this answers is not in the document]
antonrohr: I read your other comment later. And I agree, lets leave it as is for now

OrbitQt/ConnectToStadiaWidget.cpp
```diff
@@ -0,0 +1,344 @@
...
+  }
+
+  state_machine_.start();
+}
+
+void ConnectToStadiaWidget::ReloadInstances() {
+  CHECK(ggp_client_ != nullptr);
+  instance_model_.SetInstances({});
+
+  ggp_client_->GetInstancesAsync([this](outcome::result<QVector<Instance>> instances) {
```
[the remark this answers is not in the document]
antonrohr: I introduced other functions

OrbitQt/ConnectToStadiaWidget.cpp
antonrohr: done

OrbitQt/ConnectionArtifacts.h
antonrohr: I rewrote ConnectionArtifacts.h which is now ConnectionConfiguration.h. While I did that, I removed the comment, because I felt the code is self-explaining enough. Let me know if you think otherwise.

(Ps. the The was a mistake)

OrbitQt/ConnectToStadiaWidget.cpp
```diff
@@ -0,0 +1,344 @@
...
+  s_waiting_for_creds->assignProperty(ui_->instancesTableOverlay, "cancelable", true);
+  // STATE s_deploying
+  s_deploying->assignProperty(ui_->instancesTableOverlay, "visible", true);
+  s_deploying->assignProperty(ui_->instancesTableOverlay, "cancelable", true);
+  // STATE s_connected
+  s_connected->assignProperty(ui_->instancesTableOverlay, "visible", true);
+  s_connected->assignProperty(ui_->instancesTableOverlay, "spinning", false);
+  s_connected->assignProperty(ui_->instancesTableOverlay, "cancelable", true);
+  s_connected->assignProperty(ui_->instancesTableOverlay, "buttonMessage", "Disconnect");
+  s_connected->assignProperty(ui_->rememberCheckBox, "enabled", true);
```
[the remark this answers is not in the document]
antonrohr: I commented on the other comment. -> I agree, lets leave as is for now

OrbitQt/ConnectToStadiaWidget.cpp
antonrohr: You are right. I merged both states into `s_idle`. Thanks for finding that!

karupayun (commented):
Nice PR! Happy to see everything in a PR after the demos :) :) 

Sorry for being useless, but I have no comments. For me looks fine. BTW, there are a few things that I want to test when possible. :)

OrbitQt/ConnectToStadiaWidget.cpp
antonrohr: done

antonrohr:
Hey, so this PR is now officially ready for review. I actually had to change some things (for example instead of ConnectionArtifacts. there is now a ConnectionConfiguration) and I addressed @reichlfl comments. So please take a look again. 

To try this PR, you can locally checkout draft PR #1526 and run Orbit with `--enable_ui_beta`. 

My plan is to finish and merge this PR and then I will rebase the other PR (#1526) onto master.

karupayun (approved):
It seems really nice Anton! :)

Only 3 minor things for future PR's:
1- In the initial window there is no way to close Orbit. Before we had the "Cancel button", but now it's impossible to close.
2- I think that we have to update this message to tell the user to "End Session" and after that, select a process.
no process selected
3- A really minor thing, I would like to have "End Session" bigger and maybe at the bottom right. For me it seems more intuitive if it's there. But feel free to disregard this comment.

OrbitQt/ConnectToStadiaWidget.cpp
```diff
@@ -0,0 +1,378 @@
...
+                   &ConnectToStadiaWidget::OnRememberCheckBoxToggled);
+
+  SetupAndStartStateMachine();
+}
+
+bool ConnectToStadiaWidget::eventFilter(QObject* obj, QEvent* event) {
+  if (obj == parent() && event->type() == QEvent::Resize) {
+    ui_->contentFrame->resize(size());
+  }
+  return false;
```
beckerhe, line 59: According to the docs  you're supposed to call QObject::eventFilter when your custom logic doesn't match.
Not sure if that's just good practice or actually needed.
Details: https://doc.qt.io/qt-5/qobject.html#installEventFilter

OrbitQt/ConnectToStadiaWidget.h
```diff
@@ -0,0 +1,81 @@
...
+
+class ConnectToStadiaWidget : public QWidget {
+  Q_OBJECT
+  Q_PROPERTY(bool active READ IsActive WRITE SetActive)
+
+ public:
+  explicit ConnectToStadiaWidget(QWidget* parent = nullptr);
+  // This needs to be called before this class can be used. (It is not part of the constructor,
+  // because otherwise this class could not be instantiated from a .ui file)
+  void SetStadiaConnection(StadiaConnection* stadia_connection);
```
beckerhe, line 32: In that case we should go the Qt-way which means the widget should be able to work without crashing when no StadiaConnection is set. Just like a QTableView works with no model set.

Maybe that's already the case. I'm not sure.

OrbitQt/ConnectToStadiaWidget.h
```diff
@@ -0,0 +1,81 @@
...
+  bool IsActive() { return ui_->contentFrame->isEnabled(); }
```
beckerhe, line 33: const

OrbitQt/ConnectToStadiaWidget.cpp
```diff
@@ -0,0 +1,378 @@
...
+
+  instance_model_.SetInstances(instances.value());
+  emit ReceivedInstances();
+
+  for (const auto& instance : instances.value()) {
+    std::string instance_id = instance.id.toStdString();
+
+    if (remembered_instance_id_ != std::nullopt) {
+      std::optional<int> row =
+          instance_model_.GetRowOfInstanceById(remembered_instance_id_.value());
```
beckerhe, line 332: There is a generic way to make searches in a model: `QAbstractItemModel::match` (https://doc.qt.io/qt-5/qabstractitemmodel.html#match)

In that case you didn't need a new lookup function `GetRowOfInstanceById`. But I don't feel particurlarly strong about this.

OrbitQt/ConnectionConfiguration.h
```diff
@@ -0,0 +1,77 @@
...
+                            const DeploymentConfiguration* deployment_configuration)
+      : ssh_context(ssh_context),
+        grpc_port(grpc_port),
+        deployment_configuration(deployment_configuration) {
+    CHECK(ssh_context != nullptr);
+    CHECK(deployment_configuration != nullptr);
+  }
+  ~StadiaConnection() {
+    if (process_manager != nullptr) process_manager->Shutdown();
+  }
```
beckerhe, line 36: When defining a destructor you should also define copy/move-constructors/assignment operators. (Rule of 5!)

OrbitQt/ConnectionConfiguration.h
```diff
@@ -0,0 +1,77 @@
+// Copyright (c) 2020 The Orbit Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#ifndef ORBIT_QT_CONNECTION_CONFIGURATION_H_
+#define ORBIT_QT_CONNECTION_CONFIGURATION_H_
+
+#include <memory>
+#include <utility>
```
beckerhe, line 9: also optional

OrbitQt/ConnectionConfiguration.h
```diff
@@ -0,0 +1,77 @@
...
+    if (process_manager != nullptr) process_manager->Shutdown();
+  }
+  void CreateServiceDeployManager(OrbitSsh::Credentials credentials) {
+    CHECK(service_deploy_manager == nullptr);
+    service_deploy_manager = std::make_unique<ServiceDeployManager>(
+        deployment_configuration, ssh_context, std::move(credentials), grpc_port);
+  }
+
+  const OrbitSsh::Context* ssh_context;
+  const ServiceDeployManager::GrpcPort& grpc_port;
```
beckerhe, line 44: When someone passes in a temporary GrpcPort as a constructor argument this will crash when CreateServiceDeployManager is called.

OrbitQt/ConnectionConfiguration.h
beckerhe: `Shutdown` should probably called by ProcessManager's destructor. Forcing that requirement on the user doesn't feel right.

OrbitQt/ConnectionConfiguration.h
beckerhe: (because grpc_port is a reference)

OrbitQt/ConnectionConfiguration.h
```diff
@@ -0,0 +1,77 @@
...
+  }
+  ~StadiaConnection() {
+    if (process_manager != nullptr) process_manager->Shutdown();
+  }
+  void CreateServiceDeployManager(OrbitSsh::Credentials credentials) {
+    CHECK(service_deploy_manager == nullptr);
+    service_deploy_manager = std::make_unique<ServiceDeployManager>(
+        deployment_configuration, ssh_context, std::move(credentials), grpc_port);
+  }
+
```
beckerhe, line 42: That doesn't feel right. This gives that struct invariants which it shouldn't have by definition.

Also this class has state that is not really "runtime-state". It's just mimicking two "programming-stages" (without and with ServiceDeployManager). Ideally we represent that in static types instead of dynamically handling
two branches.

I think you should split that up into two classes/struct. One class that holds all resources that are needed
for constructing the ServiceDeployManager and one class that holds all the elements which are needed
to be given to the main window. And then this function can be a free function taking an r-value-reference of
the first class and those credentials and returning a value of the second class.

OrbitQt/ConnectionConfiguration.h
```diff
@@ -0,0 +1,77 @@
...
+struct NoConnection {
+ public:
+  explicit NoConnection() = default;
+
+  std::filesystem::path capture_file_path;
+};
+
+using ConnectionConfiguration =
+    std::variant<const StadiaConnection*, const LocalConnection*, const NoConnection*>;
```
beckerhe, line 73: It's difficult to say how ConnectionConfiguration will be used, so I have to ask: Why pointers? Why not values?

OrbitQt/ConnectionConfiguration.h
beckerhe: (BTW, maybe you didn't notice but by defining this destructor this struct gets non-movable and it's not copyable because of its members which means you have to head-allocate it all the time what you probably don't want.)

OrbitQt/main.cpp
```diff
@@ -314,6 +314,7 @@ int main(int argc, char* argv[]) {
 #endif
 
     QApplication app(argc, argv);
+    QApplication::setOrganizationName("The Orbit Authors");
```
beckerhe, line 317: Is that changing where QSettings stores its values?

OrbitQt/ConnectToStadiaWidget.h
reichlfl: And [[nodiscard]], unless that causes issues with the Q_PROPERTY declaration

OrbitQt/ConnectionConfiguration.h
```diff
@@ -0,0 +1,77 @@
...
+#include "OrbitClientServices/ProcessManager.h"
+#include "OrbitGgp/Instance.h"
+#include "OrbitSsh/Context.h"
+#include "OrbitSsh/Credentials.h"
+#include "grpcpp/channel.h"
+#include "servicedeploymanager.h"
+
+namespace orbit_qt {
+
+struct StadiaConnection {
```
reichlfl, line 23: This shouldn't be a struct, but a class with private fields and proper getters. Structs should only be used for objects that carry data, and should not enforce any invariants on this data (https://google.github.io/styleguide/cppguide.html#Structs_vs._Classes).

reichlfl (commented):
Some more comments on top of Henning's review.

OrbitQt/main.cpp
antonrohr: Apparently on Windows it is necessary to set the Organization name, or QSettings will not work (when used without arguments). 
Besides that, yes, it will change, when the arguments are removed from the other 2 occurrences of QSettings. Before we used this for QSettings: "The Orbit Authors" ; "Orbit Profiler". 
Since the application name is set to "orbitprofiler" in the next line. It will change to "The Orbit Authors" ; "orbitprofiler". 

Does that answer the question? 

I will write something in chat, to see if anyone has a problem with losing the settings about remembering not to show the warnings

beckerhe: Thanks for figuring that out. Of course that makes a lot of sense. you need that organization name for the registry.
Thanks for fixing that!

Yes, that answers my question!

OrbitQt/ConnectToStadiaWidget.cpp
antonrohr: Ive also read these docs and I feel they are on the rather ambiguous side. The text says to either return true in the case it should not be propagated further (filtered) and return false if the event should not be filtered. 
The example then does not return true or false, but propagates the event manually by calling QObject::eventFilter with the obj as parameter. 
I am unsure what to make of this. It seems to me that the example is rather wrong and might might not work when multiple event filter are installed.

beckerhe: I don't see how this is related to multiple event filters?

I think this example is clearer: https://doc.qt.io/qt-5/qobject.html#eventFilter
In that example you have all three cases. Explicit true, explicit false and forwarding to base class implementation.

I'd argue since we don't want to change any behaviour we should return the value that would be returned without overriding the member function at all. And that's the return value of the base class implementation.

antonrohr: Yes, I of course dont want to change any behaviour. I did not fully understand before how this actually works. But I think I now got it. The return true/false is about whether another event filter, maybe on the parent, or any other installed eventfilter should be called. The direction up the inheritance chain (base class) is independent of that.

reichlfl: To my understanding, the true / false return value is not primarily for other filters. The main purpose of the eventFilter is not to handle events, but to prevent them from being handled by the object. Event filters are called in the order they are installed, and I think if any of them returns `true`, the event is not passed to the target object.

Installing event filters to react to events of other widgets is, IMHO, abusing them a bit. But it's also a quick and easy way to achieve what you are trying to do, otherwise you'd either have to fiddle around with custom layouts, or provide a `Resized()` signal in the parent and connect this to a slot of the child.

OrbitQt/ConnectToStadiaWidget.cpp
```diff
@@ -0,0 +1,378 @@
...
+  auto* s_instances_loading = new QState();
+  state_machine_.addState(s_instances_loading);
+  auto* s_instance_selected = new QState();
+  state_machine_.addState(s_instance_selected);
+  auto* s_waiting_for_creds = new QState();
+  state_machine_.addState(s_waiting_for_creds);
+  auto* s_deploying = new QState();
+  state_machine_.addState(s_deploying);
+  auto* s_connected = new QState();
+  state_machine_.addState(s_connected);
```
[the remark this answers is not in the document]
antonrohr: done

OrbitGgp/InstanceItemModel.cpp
```diff
@@ -140,4 +142,14 @@ void InstanceItemModel::SetInstances(QVector<Instance> new_instances) {
   }
 }
 
+std::optional<int> InstanceItemModel::GetRowOfInstanceById(const QString& instance_id) {
```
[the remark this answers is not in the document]
antonrohr: Its not existing anymore, went with the QAbstractItemModel::match, as suggested

OrbitQt/ConnectToStadiaWidget.cpp
```diff
@@ -0,0 +1,378 @@
...
+namespace orbit_qt {
+
+using OrbitGgp::Instance;
+using OrbitSshQt::ScopedConnection;
+
+ConnectToStadiaWidget::ConnectToStadiaWidget(QWidget* parent)
+    : QWidget(parent), ui_(std::make_unique<Ui::ConnectToStadiaWidget>()) {
+  ui_->setupUi(this);
+
+  parent->installEventFilter(this);
```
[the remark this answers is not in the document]
antonrohr: I did revise this part and its now done differently. The main challenge I had was figuring out the best way of grouping everything in this widget besides the radioButton, so it can be enabled separately. This was problem because the radioButton and the rest of the content are not distinct rectangles, but overlapping. 
I tried QStackedWidget with QStackedLayout, but that sadly did not lead anywhere, because even when all stacks in a stackedlayout are visible at the same time, they are not transparent for mouseevents, so I would needed to have filtered and reapplied the these.

How I did it now is that everything is grouped in a QFrame called contentFrame and after the ui_->setupUi(this) call, I remove the radioButton from contentFrame and place it on top of it at the same position. ptal

OrbitQt/ConnectToStadiaWidget.cpp
antonrohr: I found a way to use a regular layout and not use eventFilter at all. 
Thank you both anyways for taking the time to explain

OrbitQt/ConnectToStadiaWidget.h
antonrohr: This now all changed a little bit, but the general Idea of the widget not crashing when these setters are not called is done.  Its a pretty useless widget then, but it does not crash :)

OrbitQt/ConnectToStadiaWidget.h
antonrohr: done

OrbitQt/ConnectToStadiaWidget.cpp
antonrohr: I like it and used it, thanks for the suggestion

OrbitQt/ConnectionConfiguration.h
antonrohr: See #1547

OrbitQt/ConnectionConfiguration.h
antonrohr: I added it, but I need to use the iwyu presubmit check for all files to correct the includes. I did not do that yet

OrbitQt/ConnectionConfiguration.h
antonrohr: I changed quite some things, and did a copy there. ptal if its good now, or something could still go wrong.

OrbitQt/ConnectionConfiguration.h
antonrohr: I used classes now, instead of structs. The classes in this PR (now in Connections.h) are actually holding less data than before, because some of it went into TargetConfiguration.h which will be part of #1526.

OrbitQt/ConnectionConfiguration.h
antonrohr: This is different now and this specific part uses values instead of pointers. This is part of #1526 now.

OrbitQt/ConnectionConfiguration.h
antonrohr: After a conversation with Hennining, Its split up now into different classes. Some in this PR in Connections.h and some in #1526 (TargetConfiguration.h)

antonrohr:
Hey everyone, 

I again changed quite some things in this PR, so further reviews would be appreciated :)
This can still be tested (and put into context) with #1526.

reichlfl (commented):
Mostly minor stuff, but the part about QState memleak should be addressed.

OrbitQt/ConnectToStadiaWidget.cpp
```diff
@@ -0,0 +1,410 @@
+// Copyright (c) 2020 The Orbit Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#include "ConnectToStadiaWidget.h"
+
+#include <QMessageBox>
+#include <QRadioButton>
+#include <QSettings>
+#include <QStackedLayout>
```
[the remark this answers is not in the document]
antonrohr: no, its gone, because I just fixed the includes

OrbitQt/ConnectToStadiaWidget.h
```diff
@@ -58,22 +62,27 @@ class ConnectToStadiaWidget : public QWidget {
...
+  std::optional<OrbitGgp::Instance> selected_instance_;
+  std::unique_ptr<ServiceDeployManager> service_deploy_manager_;
+  std::shared_ptr<grpc::Channel> grpc_channel_;
   QPointer<OrbitGgp::Client> ggp_client_ = nullptr;
   std::optional<QString> remembered_instance_id_;
 
+  // State Machine
+  QStateMachine state_machine_;
+  QState* s_instances_loading_ = nullptr;
+  QState* s_connected_ = nullptr;
```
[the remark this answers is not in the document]
antonrohr: This is trying to basically follow the Qt way of doing this. With `state_machine_.addState(QState*)`, the state machine takes ownership of the QState* and if I am informed correctly, takes car of deleting the states. 

I am sure that using unique_ptr or an object instead of pointers is possible, but I am unsure what the correct and best way of doing this is, so nothing is deleted twice. I can take some time to figure this out, or if you or @beckerhe know already and have a minute to explain, that would be great.

OrbitQt/Connections.h
```diff
@@ -0,0 +1,84 @@
...
+  }
+
+ private:
+  const OrbitSsh::Context* ssh_context_;
+  const ServiceDeployManager::GrpcPort grpc_port_;
+  const DeploymentConfiguration* deployment_configuration_;
+};
+
+class StadiaConnection {
+  friend class ConnectToStadiaWidget;
```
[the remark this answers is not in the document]
antonrohr: A couple of things come together here. 

On struct vs class, the style guide says the following: "The struct must not have invariants that imply relationships between different fields [...]". When designed as a class, as done here, this is supposed to express to other people , that when this class is instantiated, it is always a valid connection. That's why there are no setter methods to change a field. The fields have relationships between each other in the sense that for example, `service_deploy_manager_` holds the tunnel to a Stadia instance and this is exactly the instance saved in `instance_`. 
StadiaConnection is later moved into MainWindow and used there. MainWindow is not allowed to modify the contents. 

Now when the MainWindow is closed, the StadiaConnection is moved out of MainWindow and the ConnectToStadiaWidget is instantiated again and StadiaConnection is moved into the widget. This is so the user picks up in the ConnectToStadiaWidget, where they left off. A disconnect and reconnect to an instance is not necessary, since StadiaConnection is reused. 

Since ConnectToStadiaWidget needs to be able to only have a partial connection, while establishing the connection, it has its own variables `selected_instance_`, `service_deploy_manager_` and `grpc_channel_`. Now when StadiaConnection is moved into ConnectToStadiaWidget via SetConnection(), the friend access is necessary to move  the members of StadiaConnection into these variables. 

Another consideration was to not use the friend access, but have a `Clear()` method which return the members and then leaves StadiaConnection in a empty state. I decided for the approach via friend, because then StadiaConnection is always valid (when used somewhere else than in ConnectToStadiaWidget).

Does this make sense? 

Ps. I actually just realized that there are no != nullptr checks in the constructor of StadiaConnection. I will add those now

OrbitQt/ConnectToStadiaWidget.cpp
```diff
@@ -30,7 +33,16 @@ ConnectToStadiaWidget::ConnectToStadiaWidget(QWidget* parent)
     : QWidget(parent), ui_(std::make_unique<Ui::ConnectToStadiaWidget>()) {
   ui_->setupUi(this);
 
-  parent->installEventFilter(this);
+  {
+    // Move radioButton from titleBarLayout to mainFrame and position
```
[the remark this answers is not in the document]
antonrohr: done

OrbitQt/Connections.h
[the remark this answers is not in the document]
reichlfl: Hm, ok, then the class makes sense. I still don't like the usage of `friend` here, but I am also not sure what a resonably better design would be TBH.

reichlfl (approved):
LGTM, the QState handling as objects looks cleaner, thanks!
I'd prefer a different solution for the `friend` declaration, but since I also am not sure how this could be done in a more elegant way without it, I'm fine with keeping it like this.

OrbitQt/ConnectToStadiaWidget.cpp
```diff
@@ -0,0 +1,410 @@
...
+
+  ggp_client_->GetInstancesAsync([this](outcome::result<QVector<Instance>> instances) {
+    OnInstancesLoaded(std::move(instances));
+  });
+}
+
+void ConnectToStadiaWidget::CheckCredentialsAvailable() {
+  CHECK(selected_instance_.has_value());
+
+  const std::string& instance_id = selected_instance_->id.toStdString();
```
beckerhe, line 238: That works due to the reference lifetime extension rule, but shouldn't be invoked here.

`toStdString` returns an r-value - not a reference. You should store that into a temporary variable.

OrbitQt/ConnectToStadiaWidget.h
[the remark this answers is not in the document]
beckerhe: I think that's what we discussed OOB.

beckerhe (approved):
Thanks Anton! Apart from the comments LGTM!

OrbitQt/ConnectToStadiaWidget.h
```diff
@@ -0,0 +1,90 @@
...
+  [[nodiscard]] const std::shared_ptr<grpc::Channel>& GetGrpcChannel() { return grpc_channel_; }
+  void Start();
+
+ public slots:
+  void SetActive(bool value);
+
+ signals:
+  void Activated();
+  void Connected();
+  void Disconnected();
```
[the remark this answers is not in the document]
antonrohr: These are public, the other ones are private. I thought there is a difference, but the docs say there isnt. I will change it

OrbitQt/ConnectToStadiaWidget.cpp
```diff
@@ -0,0 +1,410 @@
...
+void ConnectToStadiaWidget::TrySelectRememberedInstance() {
+  if (remembered_instance_id_ == std::nullopt) return;
+
+  QModelIndexList matches = instance_model_.match(
+      instance_model_.index(0, static_cast<int>(OrbitGgp::InstanceItemModel::Columns::kId)),
+      Qt::DisplayRole, QVariant::fromValue(remembered_instance_id_.value()));
+
+  if (!matches.isEmpty()) {
+    ui_->instancesTableView->selectionModel()->setCurrentIndex(
+        matches[0], {QItemSelectionModel::SelectCurrent, QItemSelectionModel::Rows});
```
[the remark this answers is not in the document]
antonrohr: https://doc.qt.io/qt-5/qflags.html
I like it, I dont see any downside and it has the benefit of type checking compared to using the | and the int values of the flags. Will leave as is

OrbitQt/ConnectToStadiaWidget.cpp
antonrohr: done

OrbitQt/ConnectToStadiaWidget.cpp
```diff
@@ -0,0 +1,410 @@
...
+
+std::optional<StadiaConnection> ConnectToStadiaWidget::StopAndClearConnection() {
+  if (!selected_instance_.has_value() || service_deploy_manager_ == nullptr ||
+      grpc_channel_ == nullptr) {
+    return std::nullopt;
+  }
+
+  state_machine_.stop();
+
+  return StadiaConnection(std::move(selected_instance_.value()), std::move(service_deploy_manager_),
```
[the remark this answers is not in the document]
antonrohr: done

OrbitQt/Connections.h
```diff
@@ -0,0 +1,84 @@
...
+#include "OrbitClientServices/ProcessManager.h"
+#include "OrbitGgp/Instance.h"
+#include "OrbitSsh/Context.h"
+#include "OrbitSsh/Credentials.h"
+#include "grpcpp/channel.h"
+#include "servicedeploymanager.h"
+
+namespace orbit_qt {
+
+class SshConnectionArtifacts {
```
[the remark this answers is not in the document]
antonrohr: done

antonrohr:
Thanks for the review @beckerhe and @reichlfl!
I made one additional last minute change, which consolidates the SetSshConnectionArtifacts, SetConnection and Start method and makes the interface of ConnectToStadiaWidget a little be cleaner imho. Its just a few lines change, without any change in logic. 

I soon as the checks are done, I will potentially fix includes, rebase and reorder some commits and merge

OrbitGgp/InstanceItemModel.cpp
```diff
@@ -4,9 +4,9 @@
 
 #include "OrbitGgp/InstanceItemModel.h"
 
-namespace {
-enum class Columns { DisplayName, ID, IPAddress, LastUpdated, Owner, Pool, NumberOfColumns };
-}  // namespace
+#include <iterator>
+
+#include "OrbitBase/Logging.h"
```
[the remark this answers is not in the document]
antonrohr: Will do as part of the final rebase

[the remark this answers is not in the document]
antonrohr: done

PR #1507, opened 2020-11-26
Add ProducerSideServiceImpl and ProducerSideServer

dpallotti:
For the communication between producers and `OrbitService`. Bug: http://b/174028631.

Synchronization mess part 1: I'm aware that the synchronization in `ProducerSideServiceImpl` is a bit tricky and especially not so easy on the reader. I'm not really sure how to improve it, but I tried to test it quite a bit.

### DIFF
--- a/OrbitService/ProducerSideServer.cpp
+++ b/OrbitService/ProducerSideServer.cpp
@@ -1,0 +1,54 @@
+// Copyright (c) 2020 The Orbit Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#include "ProducerSideServer.h"
+
+#include <sys/stat.h>
+
+#include "OrbitBase/SafeStrerror.h"
+
+namespace orbit_service {
+
+bool ProducerSideServer::BuildAndStart(std::string_view unix_domain_socket_path) {
+  CHECK(server_ == nullptr);
+
+  grpc::ServerBuilder builder;
+  builder.AddListeningPort(absl::StrFormat("unix:%s", unix_domain_socket_path),
+                           grpc::InsecureServerCredentials());
+
+  builder.RegisterService(&producer_side_service_);
+
+  server_ = builder.BuildAndStart();
+  if (server_ == nullptr) {
+    return false;
+  }
+
+  // When OrbitService runs as root, also allow non-root producers
+  // (e.g., the game) to communicate over the Unix domain socket.
+  if (chmod(std::string{unix_domain_socket_path}.c_str(), 0777) != 0) {
+    ERROR("Changing mode bits to 777 of \"%s\": %s", unix_domain_socket_path, SafeStrerror(errno));
+    server_->Shutdown();
+    server_->Wait();
+    return false;
+  }
+
+  return true;
+}
+
+void ProducerSideServer::ShutdownAndWait() {
+  CHECK(server_ != nullptr);
+  producer_side_service_.OnExitRequest();
+  server_->Shutdown();
+  server_->Wait();
+}
+
+void ProducerSideServer::OnCaptureStartRequested(CaptureEventBuffer* capture_event_buffer) {
+  producer_side_service_.OnCaptureStartRequested(capture_event_buffer);
+}
+
+void ProducerSideServer::OnCaptureStopRequested() {
+  producer_side_service_.OnCaptureStopRequested();
+}
+
+}  // namespace orbit_service
--- a/OrbitService/ProducerSideServiceImpl.cpp
+++ b/OrbitService/ProducerSideServiceImpl.cpp
@@ -92,7 +92,7 @@ grpc::Status ProducerSideServiceImpl::ReceiveCommandsAndSendEvents(
   // Note that this will also be protected by service_state_mutex_.
   bool all_events_sent_received = true;
 
-  std::atomic<bool> exit_send_commands_thread = false;
+  std::atomic<bool> receive_events_thread_exited = false;
 
   // This thread is responsible for writing on stream, and specifically for
   // sending StartCaptureCommands and StopCaptureCommands to the connected producer.
@@ -101,7 +101,7 @@ grpc::Status ProducerSideServiceImpl::ReceiveCommandsAndSendEvents(
                                    context,
                                    stream,
                                    &all_events_sent_received,
-                                   &exit_send_commands_thread};
+                                   &receive_events_thread_exited};
 
   // This thread is responsible for reading from stream, and specifically for
   // receiving CaptureEvents and AllEventsSent messages.
@@ -111,7 +111,7 @@ grpc::Status ProducerSideServiceImpl::ReceiveCommandsAndSendEvents(
 
   // When receive_events_thread exits because stream->Read(&request) fails,
   // it means that the producer has disconnected: ask send_commands_thread to exit, too.
-  exit_send_commands_thread = true;
+  receive_events_thread_exited = true;
   send_commands_thread.join();
 
   {
@@ -127,7 +127,7 @@ void ProducerSideServiceImpl::SendCommandsThread(
     grpc::ServerContext* context,
     grpc::ServerReaderWriter<orbit_grpc_protos::ReceiveCommandsAndSendEventsResponse,
                              orbit_grpc_protos::ReceiveCommandsAndSendEventsRequest>* stream,
-    bool* all_events_sent_received, std::atomic<bool>* exit_send_commands_thread) {
+    bool* all_events_sent_received, std::atomic<bool>* receive_events_thread_exited) {
   // As a result of initializing prev_capture_status to kCaptureFinished,
   // an initial StartCaptureCommand is sent
   // if service_state_.capture_status is actually CaptureStatus::kCaptureStarted,
@@ -138,11 +138,11 @@ void ProducerSideServiceImpl::SendCommandsThread(
   // This loop keeps track of changes to service_state_.capture_status using
   // conditional critical sections on service_state_mutex_ and updating prev_capture_status,
   // and sends StartCaptureCommands and StopCaptureCommands accordingly.
-  // It exits when *exit_send_commands_thread is true, when service_state_.exit_requested is true,
+  // It exits when one of *receive_events_thread_exited and service_state_.exit_requested is true,
   // or when Write fails (because the producer disconnected or because the context was cancelled).
   while (true) {
     // This is set when ReceiveEventsThread has exited. At that point this thread should also exit.
-    if (*exit_send_commands_thread) {
+    if (*receive_events_thread_exited) {
       return;
     }
 
--- a/OrbitService/ProducerSideServiceImpl.cpp
+++ b/OrbitService/ProducerSideServiceImpl.cpp
@@ -146,13 +146,63 @@ void ProducerSideServiceImpl::SendCommandsThread(
       return;
     }
 
-    service_state_mutex_.Lock();
-    if (service_state_.exit_requested) {
-      service_state_mutex_.Unlock();
-      return;
-    }
+    {
+      absl::MutexLock lock{&service_state_mutex_};
+      if (service_state_.exit_requested) {
+        return;
+      }
 
-    if (service_state_.capture_status != prev_capture_status) {
+      if (service_state_.capture_status == prev_capture_status) {
+        // Wait for service_state_.capture_status to change or for service_state->exit_requested
+        // (the next iteration will handle the change).
+        // Use a timeout to periodically check (in the next iteration)
+        // for *terminate_send_commands_thread, set by ReceiveCommandsAndSendEvents.
+        static constexpr absl::Duration kCheckExitSendCommandsThreadInterval = absl::Seconds(1);
+
+        // The three cases in this switch are almost identical, except
+        // for the value service_state->capture_status is compared with.
+        // The reason for the duplication is that AwaitWithTimeout takes a function pointer, which
+        // means that the lambda cannot capture the initial value of service_state_.capture_status.
+        switch (service_state_.capture_status) {
+          case CaptureStatus::kCaptureStarted: {
+            service_state_mutex_.AwaitWithTimeout(absl::Condition(
+                                                      +[](ServiceState* service_state) {
+                                                        return service_state->exit_requested ||
+                                                               service_state->capture_status !=
+                                                                   CaptureStatus::kCaptureStarted;
+                                                      },
+                                                      &service_state_),
+                                                  kCheckExitSendCommandsThreadInterval);
+          } break;
+
+          case CaptureStatus::kCaptureStopping: {
+            service_state_mutex_.AwaitWithTimeout(absl::Condition(
+                                                      +[](ServiceState* service_state) {
+                                                        return service_state->exit_requested ||
+                                                               service_state->capture_status !=
+                                                                   CaptureStatus::kCaptureStopping;
+                                                      },
+                                                      &service_state_),
+                                                  kCheckExitSendCommandsThreadInterval);
+          } break;
+
+          case CaptureStatus::kCaptureFinished: {
+            service_state_mutex_.AwaitWithTimeout(absl::Condition(
+                                                      +[](ServiceState* service_state) {
+                                                        return service_state->exit_requested ||
+                                                               service_state->capture_status !=
+                                                                   CaptureStatus::kCaptureFinished;
+                                                      },
+                                                      &service_state_),
+                                                  kCheckExitSendCommandsThreadInterval);
+          } break;
+        }
+
+        continue;
+      }
+
+      // service_state_.capture_status has changed compared to prev_capture_status:
+      // handle the change.
       switch (service_state_.capture_status) {
         case CaptureStatus::kCaptureStarted: {
           ++service_state_.producers_remaining;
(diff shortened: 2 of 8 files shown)


OrbitService/ProducerSideServiceImpl.cpp
```diff
@@ -0,0 +1,302 @@
...
+                             orbit_grpc_protos::ReceiveCommandsAndSendEventsRequest>* stream,
+    bool* all_events_sent_received, std::atomic<bool>* exit_send_commands_thread) {
+  // As a result of initializing prev_capture_status to kCaptureFinished,
+  // an initial StartCaptureCommand is sent
+  // if service_state_.capture_status is actually CaptureStatus::kCaptureStarted,
+  // and an initial StopCaptureCommand is sent (with little effect)
+  // if service_state_.capture_status is actually CaptureStatus::kCaptureStopping.
+  CaptureStatus prev_capture_status = CaptureStatus::kCaptureFinished;
+
+  while (true) {
```
florian-kuebler, line 131: A general description of that loop (or just the whole method), would be beneficial.

OrbitService/ProducerSideServiceImpl.cpp
```diff
@@ -0,0 +1,302 @@
...
+    }
+
+    // Wait for service_state_.capture_status to change or for service_state->exit_requested
+    // (the next iteration will handle the change).
+    // Use a timeout to periodically check (in the next iteration)
+    // for *terminate_send_commands_thread, set by ReceiveCommandsAndSendEvents.
+    static constexpr absl::Duration kCheckExitSendCommandsThreadInterval = absl::Seconds(1);
+    switch (service_state_.capture_status) {
+      case CaptureStatus::kCaptureStarted: {
+        service_state_mutex_.AwaitWithTimeout(absl::Condition(
```
florian-kuebler, line 195: It seems that you need this super duplicated code (for all the three cases), as you need a function pointer in the AwaitWithTimout, which can't capture the state at the beginning. If so, could you add comment for that?

OrbitService/ProducerSideServiceImpl.cpp
```diff
@@ -0,0 +1,302 @@
...
+      absl::MutexLock lock{&service_state_mutex_};
+      if (service_state_.exit_requested) {
+        break;
+      }
+    }
+
+    switch (request.event_case()) {
+      case orbit_grpc_protos::ReceiveCommandsAndSendEventsRequest::kBufferedCaptureEvents: {
+        absl::MutexLock lock{&capture_event_buffer_mutex_};
+        if (capture_event_buffer_ != nullptr) {
```
florian-kuebler, line 250: In what cases would the buffer be nullptr? Is that reasonable? Should this be rather a check, or do we want to log some error in this case?

OrbitService/ProducerSideServiceImpl.cpp
```diff
@@ -0,0 +1,302 @@
...
+        }
+      } break;
+
+      case orbit_grpc_protos::ReceiveCommandsAndSendEventsRequest::kAllEventsSent: {
+        LOG("Received AllEventsSent from CaptureEventProducer");
+        absl::MutexLock lock{&service_state_mutex_};
+        switch (service_state_.capture_status) {
+          case CaptureStatus::kCaptureStarted: {
+            ERROR("CaptureEventProducer sent AllEventsSent while still capturing");
+            if (!*all_events_sent_received) {
```
florian-kuebler, line 264: That would benefit from a brief comment.

OrbitService/ProducerSideServiceImpl.cpp
```diff
@@ -0,0 +1,302 @@
...
+          }
+        }
+      } break;
+
+      case orbit_grpc_protos::ReceiveCommandsAndSendEventsRequest::kAllEventsSent: {
+        LOG("Received AllEventsSent from CaptureEventProducer");
+        absl::MutexLock lock{&service_state_mutex_};
+        switch (service_state_.capture_status) {
+          case CaptureStatus::kCaptureStarted: {
+            ERROR("CaptureEventProducer sent AllEventsSent while still capturing");
```
florian-kuebler, line 263: Can this actually happen?

OrbitService/ProducerSideServiceImpl.cpp
```diff
@@ -0,0 +1,302 @@
...
+  }
+}
+
+void ProducerSideServiceImpl::ReceiveEventsThread(
+    grpc::ServerContext* /*context*/,
+    grpc::ServerReaderWriter<orbit_grpc_protos::ReceiveCommandsAndSendEventsResponse,
+                             orbit_grpc_protos::ReceiveCommandsAndSendEventsRequest>* stream,
+    bool* all_events_sent_received) {
+  orbit_grpc_protos::ReceiveCommandsAndSendEventsRequest request;
+  while (stream->Read(&request)) {
```
florian-kuebler, line 239: Shouldn't there also an ` || *all_events_sent_received`?

OrbitService/ProducerSideServiceImplTest.cpp
```diff
@@ -0,0 +1,244 @@
...
+      stream_;
+  std::thread read_thread_;
+};
+
+class MockCaptureEventBuffer : public CaptureEventBuffer {
+ public:
+  MOCK_METHOD(void, AddEvent, (orbit_grpc_protos::CaptureEvent && event), (override));
+};
+
+class ProducerSideServiceImplTest : public ::testing::Test {
```
florian-kuebler, line 101: Could you also add some tests for the corner cases (errors, no capture data, no AllEventsSend, etc.?).

florian-kuebler (changes requested):
Thanks for that, overall looking very good. I was afraid of the synchronization, but it looks good from my side. But beside the minor commits, I really would like to see a couple of tests for the corner cases.

OrbitService/ProducerSideServiceImplTest.cpp
dpallotti: Added some more tests, please have a look.

OrbitService/ProducerSideServiceImpl.h
```diff
@@ -0,0 +1,61 @@
...
+#define ORBIT_SERVICE_PRODUCER_SIDE_SERVICE_IMPL_H_
+
+#include "CaptureStartStopListener.h"
+#include "absl/container/flat_hash_set.h"
+#include "absl/synchronization/mutex.h"
+#include "producer_side_services.grpc.pb.h"
+
+namespace orbit_service {
+
+class ProducerSideServiceImpl final : public orbit_grpc_protos::ProducerSideService::Service,
```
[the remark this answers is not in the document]
florian-kuebler: But there are additional methods. So I was thinking of a very very brief insight of the implementation and the additional methods. 
Regarding thread-safety, basically the same. Some rough ideas whats going on with the mutexes here would be nice.

OrbitService/ProducerSideServiceImpl.cpp
```diff
@@ -0,0 +1,302 @@
...
+                                ::orbit_grpc_protos::ReceiveCommandsAndSendEventsRequest>* stream) {
+  LOG("A CaptureEventProducer has connected calling ReceiveCommandsAndSendEvents");
+
+  {
+    absl::MutexLock lock{&server_contexts_mutex_};
+    server_contexts_.emplace(context);
+  }
+
+  // This will also be protected by service_state_mutex_.
+  bool all_events_sent_received = true;
```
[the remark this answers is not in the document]
florian-kuebler: Looks good now, thanks.

OrbitService/ProducerSideServiceImpl.cpp
florian-kuebler: Okay, thanks for explanation. I leave it up to you on whether to log errors here or not.

OrbitService/ProducerSideServiceImpl.cpp
florian-kuebler: Sounds good. Thanks!

OrbitService/ProducerSideServiceImpl.cpp
florian-kuebler: I see. Thanks!

florian-kuebler (approved):
Perfect, thanks for the additional tests. LGTM now!

OrbitService/ProducerSideServiceImpl.cpp
```diff
@@ -0,0 +1,324 @@
...
+
+    // Wait (for a limited amount of time) for all producers to send AllEventsSent or to disconnect.
+    service_state_mutex_.AwaitWithTimeout(absl::Condition(
+                                              +[](ServiceState* service_state) {
+                                                return service_state->producers_remaining == 0 ||
+                                                       service_state->exit_requested;
+                                              },
+                                              &service_state_),
+                                          max_wait_for_all_events_sent_);
+    CHECK(service_state_.producers_remaining >= 0);
```
dimitry-, line 40: == 0 || exit requested?

OrbitService/ProducerSideServiceImpl.cpp
```diff
@@ -0,0 +1,324 @@
...
+    for (grpc::ServerContext* context : server_contexts_) {
+      // This should cause blocking Reads on ServerReaderWriter to fail immediately.
+      context->TryCancel();
+    }
+  }
+
+  {
+    absl::WriterMutexLock lock{&capture_event_buffer_mutex_};
+    capture_event_buffer_ = nullptr;
+  }
```
dimitry-, line 76: maybe extract this to ClearEventBuffer?

OrbitService/ProducerSideServiceImpl.cpp
```diff
@@ -0,0 +1,324 @@
...
+grpc::Status ProducerSideServiceImpl::ReceiveCommandsAndSendEvents(
+    ::grpc::ServerContext* context,
+    ::grpc::ServerReaderWriter< ::orbit_grpc_protos::ReceiveCommandsAndSendEventsResponse,
+                                ::orbit_grpc_protos::ReceiveCommandsAndSendEventsRequest>* stream) {
+  LOG("A CaptureEventProducer has connected calling ReceiveCommandsAndSendEvents");
+
+  {
+    absl::MutexLock lock{&server_contexts_mutex_};
+    server_contexts_.emplace(context);
+  }
```
dimitry-, line 88: SetContext?

OrbitService/ProducerSideServiceImpl.cpp
```diff
@@ -0,0 +1,324 @@
...
+          command.mutable_start_capture_command();
+          if (!stream->Write(command)) {
+            ERROR("Sending StartCaptureCommand to CaptureEventProducer");
+            LOG("Terminating call to ReceiveCommandsAndSendEvents as Write failed");
+            // Cause Read in ReceiveEventsThread to also fail if for some reason it hasn't already.
+            context->TryCancel();
+            return;
+          }
+          LOG("Sent StartCaptureCommand to CaptureEventProducer");
+        } break;
```
dimitry-, line 173: nit: move break inside {}?

OrbitService/ProducerSideServiceImpl.cpp
```diff
@@ -0,0 +1,324 @@
...
+  CaptureStatus prev_capture_status = CaptureStatus::kCaptureFinished;
+
+  // This loop keeps track of changes to service_state_.capture_status using
+  // conditional critical sections on service_state_mutex_ and updating prev_capture_status,
+  // and sends StartCaptureCommands and StopCaptureCommands accordingly.
+  // It exits when *exit_send_commands_thread is true, when service_state_.exit_requested is true,
+  // or when Write fails (because the producer disconnected or because the context was cancelled).
+  while (true) {
+    // This is set when ReceiveEventsThread has exited. At that point this thread should also exit.
+    if (*exit_send_commands_thread) {
```
dimitry-, line 145: revieve_event_thread_exited?

OrbitService/ProducerSideServiceImpl.cpp
```diff
@@ -0,0 +1,324 @@
...
+      return;
+    }
+
+    if (service_state_.capture_status != prev_capture_status) {
+      switch (service_state_.capture_status) {
+        case CaptureStatus::kCaptureStarted: {
+          prev_capture_status = service_state_.capture_status;
+          ++service_state_.producers_remaining;
+          *all_events_sent_received = false;
+          service_state_mutex_.Unlock();
```
dimitry-, line 161: Yes this is really hard to track. One thing you could do is to set prev_capture_status in L153, You can then use it in switch without locking. As for other fields... What would happen if you reistablished the lock here? Is there a possibility of race condition on them in other words do they have to be scoped with status read?

dimitry- (commented):
Hi, sorry for delay in review - some preliminary comments. Still looking at the change.

OrbitService/ProducerSideServiceImpl.cpp
dpallotti: That would be false in case of timeout. I could do something like `!condition_was_satisfied || service_state_.producers_remaining == 0 || service_state->exit_requested` (I think? Or something like that), but that actually goes beyond the point of this `CHECK`, as it's basically restating the condition, while I just wanted to have a `CHECK` on `service_state_.producers_remaining`.

OrbitService/ProducerSideServiceImpl.cpp
dpallotti: I know this is a minor code duplication, but in this code I really want to avoid extracting code that locks mutexes outside of the already numerous (but necessary) methods that take those mutexes, as doing that distracts from the actual code flow as it requires jumping somewhere else. In other words in this complex synchronized code I want to prioritize following the flow (which is already hard enough) to normal code practices.

OrbitService/ProducerSideServiceImpl.cpp
dpallotti: As above, I don't want to add more methods that take mutexes, even more here are there's no code duplication.

OrbitService/ProducerSideServiceImpl.cpp
dpallotti: The break outside is something we do elsewhere and that I like. The idea is to separate the actual code from the `break` required by the old `switch` construct, which has nothing to do with the content of the braces.

OrbitService/ProducerSideServiceImpl.cpp
dpallotti: I guess it's better, done.

OrbitService/ProducerSideServiceImpl.cpp
dimitry-: maybe just remove the check in that case, I do not think it adds anything.

OrbitService/ProducerSideServiceImpl.cpp
dpallotti: Hmmm, I don't think the mutex should be acquired twice, as `service_state_.capture_status` and `service_state_.producers_remaining` and also `all_events_sent_received` go together. The value assigned to the last two depends quite strongly on the current value of `service_state_.capture_status`. That shouldn't be relaxed only to make it easier to track `Unlock`s.

OrbitService/ProducerSideServer.cpp
```diff
@@ -0,0 +1,55 @@
...
+
+  builder.RegisterService(&producer_side_service_);
+
+  server_ = builder.BuildAndStart();
+  if (server_ == nullptr) {
+    return false;
+  }
+
+  // When OrbitService runs as root, also allow non-root producers
+  // (e.g., the game) to communicate over the Unix domain socket.
```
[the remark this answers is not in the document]
dimitry-: yes - it is about todo.

OrbitService/ProducerSideServiceImpl.cpp
dimitry-: well, I was under impression that producers_remaining cannot go negative. but ok. you can keep the check if you want to make sure it is not negative.

OrbitService/ProducerSideServiceImpl.cpp
dimitry-: ok, then split the case into 2. One scoped on the mutex - and the second one doing communication based on prev_capture_status, it is going to be more readable and will have a clear mutex scope.

OrbitService/ProducerSideServer.cpp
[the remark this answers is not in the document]
dpallotti: Ok, I'm certainly happy to remove the TODO.

OrbitService/ProducerSideServiceImpl.cpp
dpallotti: You mean I should have two separate `switch`es?

dimitry-: yes, switches, it makes code more readable, and mutex scope easier to follow.

dpallotti: Makes sense, done.

OrbitService/ProducerSideServiceImpl.cpp
```diff
@@ -241,9 +246,9 @@ void ProducerSideServiceImpl::SendCommandsThread(
                                                   },
                                                   &service_state_),
                                               kCheckExitSendCommandsThreadInterval);
-        service_state_mutex_.Unlock();
       } break;
     }
+    service_state_mutex_.Unlock();
```
dimitry-, line 251: I wonder if this can be scoped somehow as well. Maybe check == case first with early continue?

dpallotti: I don't understand the suggestion.

OrbitService/ProducerSideServiceImpl.cpp
```diff
@@ -0,0 +1,329 @@
...
+      return;
+    }
+
+    service_state_mutex_.Lock();
+    if (service_state_.exit_requested) {
+      service_state_mutex_.Unlock();
+      return;
+    }
+
+    if (service_state_.capture_status != prev_capture_status) {
```
dimitry-, line 155: if you reverse this if - you can scope service_state_mutex_ better. I think.

dpallotti: Done, it doesn't make the biggest difference but I guess it's a bit better.

Let me know if there's something else that needs to be addressed.

OrbitService/ProducerSideServiceImpl.cpp
```diff
@@ -152,103 +152,104 @@ void ProducerSideServiceImpl::SendCommandsThread(
...
+      } break;
     }
+    prev_capture_status = service_state_.capture_status;
 
-    // Wait for service_state_.capture_status to change or for service_state->exit_requested
-    // (the next iteration will handle the change).
-    // Use a timeout to periodically check (in the next iteration)
-    // for *terminate_send_commands_thread, set by ReceiveCommandsAndSendEvents.
-    static constexpr absl::Duration kCheckExitSendCommandsThreadInterval = absl::Seconds(1);
+    service_state_mutex_.Unlock();
```
dimitry-, line 221: And at this point you can use scoped lock I think: MutexLock lock (....)

dimitry- (approved):
otherwise lgtm

OrbitService/ProducerSideServiceImpl.cpp
dpallotti: Done.

PR #1508, opened 2020-11-26
Fix issues with frame tracks when loading a capture

ronaldfw:
When loading a capture, Orbit can be in a state where hooking a function
that has hits in the "Live" tab must not be allowed, either because
Orbit is not in a connected state (no process selected) or because the
module to which the function belongs is not loaded. In both these cases,
the "Hook" option is disabled in the "Live" context menu, however it was
incorrectly possible to hook a function by enabling a frame track. In
the connected state, this leads to a crash when taking a capture with
the incorrectly hooked function.

We now apply the same check that is used for enabling the "Hook" context
menu entry when enabling the frame track for a function to make sure
it's only hooked when possible.

Bug: http://b/174296341

Tested: Manual tests for the cases described on the bug.

### DIFF
--- a/OrbitGl/FunctionsDataView.cpp
+++ b/OrbitGl/FunctionsDataView.cpp
@@ -33,14 +33,34 @@ const std::vector<DataView::Column>& FunctionsDataView::GetColumns() {
   return columns;
 }
 
+bool FunctionsDataView::ShouldShowSelectedFunctionIcon(const FunctionInfo& function) {
+  return GOrbitApp->IsFunctionSelected(function);
+}
+
+bool FunctionsDataView::ShouldShowFrameTrackIcon(const FunctionInfo& function) {
+  if (GOrbitApp->IsFrameTrackEnabled(function)) {
+    return true;
+  }
+  if (GOrbitApp->HasCaptureData()) {
+    const CaptureData& capture_data = GOrbitApp->GetCaptureData();
+    if (!GOrbitApp->IsCaptureConnected(capture_data) &&
+        GOrbitApp->HasFrameTrackInCaptureData(function)) {
+      // This case occurs when loading a capture. We still want to show the indicator that a frame
+      // track is enabled for the function.
+      return true;
+    }
+  }
+  return false;
+}
+
 std::string FunctionsDataView::BuildSelectedColumnsString(const FunctionInfo& function) {
   std::string result = kUnselectedFunctionString;
-  if (GOrbitApp->IsFunctionSelected(function)) {
+  if (ShouldShowSelectedFunctionIcon(function)) {
     absl::StrAppend(&result, kSelectedFunctionString);
-    if (GOrbitApp->IsFrameTrackEnabled(function)) {
+    if (ShouldShowFrameTrackIcon(function)) {
       absl::StrAppend(&result, " ", kFrameTrackString);
     }
-  } else if (GOrbitApp->IsFrameTrackEnabled(function)) {
+  } else if (ShouldShowFrameTrackIcon(function)) {
     absl::StrAppend(&result, kFrameTrackString);
   }
   return result;
--- a/OrbitGl/FunctionsDataView.h
+++ b/OrbitGl/FunctionsDataView.h
@@ -56,6 +56,8 @@ class FunctionsDataView : public DataView {
   static const std::string kMenuActionDisassembly;
 
  private:
+  static bool ShouldShowSelectedFunctionIcon(const orbit_client_protos::FunctionInfo& function);
+  static bool ShouldShowFrameTrackIcon(const orbit_client_protos::FunctionInfo& function);
   std::vector<const orbit_client_protos::FunctionInfo*> functions_;
 };
 
--- a/OrbitGl/LiveFunctionsDataView.cpp
+++ b/OrbitGl/LiveFunctionsDataView.cpp
@@ -174,8 +174,14 @@ std::vector<std::string> LiveFunctionsDataView::GetContextMenu(
     const FunctionStats& stats = capture_data.GetFunctionStatsOrDefault(selected_function);
     // We need at least one function call to a function so that adding iterators makes sense.
     enable_iterator |= stats.count() > 0;
-    enable_enable_frame_track |= !GOrbitApp->IsFrameTrackEnabled(selected_function);
-    enable_disable_frame_track |= GOrbitApp->IsFrameTrackEnabled(selected_function);
+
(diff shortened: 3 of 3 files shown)


OrbitGl/FunctionsDataView.h
```diff
@@ -56,6 +56,8 @@ class FunctionsDataView : public DataView {
   static const std::string kMenuActionDisassembly;
 
  private:
+  static bool ShowSelectedFunctionIcon(const orbit_client_protos::FunctionInfo& function);
+  static bool ShowFrameTrackIcon(const orbit_client_protos::FunctionInfo& function);
```
dpallotti, line 60: "show" here to me sounds more like performing the actual action. Personally I would use "should show".

OrbitGl/FunctionsDataView.cpp
```diff
@@ -33,14 +33,33 @@ const std::vector<DataView::Column>& FunctionsDataView::GetColumns() {
...
 }
 
+bool FunctionsDataView::ShowSelectedFunctionIcon(const FunctionInfo& function) {
+  return GOrbitApp->IsFunctionSelected(function);
+}
+
+bool FunctionsDataView::ShowFrameTrackIcon(const FunctionInfo& function) {
+  if (GOrbitApp->IsFrameTrackEnabled(function)) {
+    return true;
+  } else if (GOrbitApp->HasCaptureData()) {
```
dpallotti, line 43: This is a clang-tidy warning "do not use else after return".

dpallotti (approved):
Minor comments, looks good. I trust your testing on whether this really covers all cases now. Thanks!

OrbitGl/FunctionsDataView.h
ronaldfw: Yes, agree. Fixed.

OrbitGl/FunctionsDataView.cpp
ronaldfw: Yes, makes sense. Fixed.

ronaldfw:
> Minor comments, looks good. I trust your testing on whether this really covers all cases now. Thanks!

Hmm, I don't know if I trust myself, but I've gone through the cases you mentioned on the bug.

florian-kuebler:
Thanks for fixing this, but it feels a bit hacked and more complicated then it should be. Maybe we can rework once more on the design of frame tracks (who stores what), to make this all clearer.

I think this change is fine for the release (as it absolutely fixes the issue described), but for master we should try to make it cleaner.
WDYT?

ronaldfw:
> Thanks for fixing this, but it feels a bit hacked and more complicated then it should be. Maybe we can rework once more on the design of frame tracks (who stores what), to make this all clearer.
> 
> I think this change is fine for the release (as it absolutely fixes the issue described), but for master we should try to make it cleaner.
> WDYT?

Apart from encapsulating the logic a bit more, I don't see right now how to really make this a lot simpler. The state is unfortunately complex with Orbit being in connected state vs. disconnected state, hooking of functions has to be correct with respect to frame tracks, frame tracks settings has to be reflected in DataManager and CaptureData, etc. 

One option that radically simplifies this would be to disallow enabling frame tracks after capture. That is, you have to enable it in the "Functions" tab. That pretty much removes all weird states, as one has to be connected and has to have loaded the module. 

Maybe that's ok from a user's perspective?

OrbitGl/FunctionsDataView.cpp
```diff
@@ -33,14 +33,34 @@ const std::vector<DataView::Column>& FunctionsDataView::GetColumns() {
   return columns;
 }
 
+bool FunctionsDataView::ShouldShowSelectedFunctionIcon(const FunctionInfo& function) {
+  return GOrbitApp->IsFunctionSelected(function);
+}
+
+bool FunctionsDataView::ShouldShowFrameTrackIcon(const FunctionInfo& function) {
+  if (GOrbitApp->IsFrameTrackEnabled(function)) {
```
florian-kuebler, line 41: Can't we ensure, that this will return "true" whenever a frame tack for that function is drawn (disregarding the capture data - is connected case)?

The idea would be: DataManager stores all the FrameTrack Functions that are drawn or will be drawn in the next capture.
Use CaptureData's FrameTracks is just for serialization, but to do so, store all the functions of FrameTracks currently drawn.

On Deserialization, ensure, that we also enable the FrameTracks on the DataManager, but do not "enable"/"select" the functions, if that capture is not connected.

dpallotti:
> I think this change is fine for the release (as it absolutely fixes the issue described), but for master we should try to make it cleaner.
> WDYT?

Maybe there's room for improvement (although Orbit can be in many states and this is a reflection of that), but in my opinion the fixes on master and on the release should initially be as similar as possible.

florian-kuebler:
>One option that radically simplifies this would be to disallow enabling frame tracks after capture. That is, you have to enable it in the "Functions" tab. That pretty much removes all weird states, as one has to be connected and has to have loaded the module.
>
>Maybe that's ok from a user's perspective?

Hmm, I pretty much like the possibility to actually add frame tracks after capturing, but that might be actually no real-user perspective. There is probably not much of a difference for a user before a capture to hook a function vs. to enable a frametrack. And in the most cases it will be the same function anyways. So maybe thats the way to go.

florian-kuebler:
>Maybe there's room for improvement (although Orbit can be in many states and this is a reflection of that), but in my opinion the fixes on master and on the release should initially be as similar as possible.

Would be also fine to me, if we start working on "a proper" fix (if that even exists, maybe I am also just wrong here).

OrbitGl/FunctionsDataView.cpp
ronaldfw: It seems DataManager is not touched at all when deserializing data. Is it a good idea to now store some of the capture related data in DataManager?

The issue is that you can have frame tracks in the capture data but they will not be drawn when you capture again (as you can just unhook the function).

ronaldfw:
I'll merge this for now, but definitely want to discuss this more, @florian-kuebler and @dpallotti, to see if we can come up with an overall cleaner solution.

OrbitGl/FunctionsDataView.cpp
florian-kuebler: We at least need to think carefully about this.
Right now, I do not see an issue of storing this information in data manager.

I see the issue you describe, that feels a bit strange. But maybe there is a solution also for that.

Unfortunately, need to go now. We can discuss this offline tomorrow in more detail.

PR #1512, opened 2020-11-27
Add a couple of ORBIT_SCOPEs to TracerThread

dpallotti:


OrbitLinuxTracing/TracerThread.cpp
```diff
@@ -1016,20 +1016,27 @@ void TracerThread::ProcessDeferredEvents() {
   pthread_setname_np(pthread_self(), "Proc.Def.Events");
   bool should_exit = false;
   while (!should_exit) {
+    ORBIT_SCOPE("ProcessDeferredEvents iteration");
```
ronaldfw, line 1019: For consistency with above, name this TracerThread::ProcessDeferredEvents?

OrbitLinuxTracing/TracerThread.cpp
```diff
@@ -1016,20 +1016,27 @@ void TracerThread::ProcessDeferredEvents() {
...
     std::vector<std::unique_ptr<PerfEvent>> events = ConsumeDeferredEvents();
     if (events.empty()) {
       // TODO: use a wait/notify mechanism instead of check/sleep.
+      ORBIT_SCOPE("Sleep");
       usleep(IDLE_TIME_ON_EMPTY_DEFERRED_EVENTS_US);
     } else {
-      for (auto& event : events) {
-        int fd = event->GetOriginFileDescriptor();
-        event_processor_.AddEvent(fd, std::move(event));
+      {
```
ronaldfw, line 1029: The extra braces are not needed here, if I'm not mistaken. I would remove them (but also see other comment about consistency).

OrbitLinuxTracing/TracerThread.cpp
```diff
@@ -1016,20 +1016,27 @@ void TracerThread::ProcessDeferredEvents() {
...
   bool should_exit = false;
   while (!should_exit) {
+    ORBIT_SCOPE("ProcessDeferredEvents iteration");
     // When "should_exit" becomes true, we know that we have stopped generating
     // deferred events. The last iteration will consume all remaining events.
     should_exit = stop_deferred_thread_;
     std::vector<std::unique_ptr<PerfEvent>> events = ConsumeDeferredEvents();
     if (events.empty()) {
       // TODO: use a wait/notify mechanism instead of check/sleep.
+      ORBIT_SCOPE("Sleep");
```
ronaldfw, line 1026: In other places you add extra braces to clarify the exact scope, here you don't. I think either would be fine, but should be consistent.

ronaldfw (approved):
LGTM, just minor comments on consistently adding braces to clearly define scope.

OrbitLinuxTracing/TracerThread.cpp
dpallotti: `TracerThread::ProcessDeferredEvents` is too long in the ui and `ProcessDeferredEvents` is quite a unique name, while `Run` isn't (but I wanted to change `Tracer` as it was too generic).

OrbitLinuxTracing/TracerThread.cpp
dpallotti: Braces are not needed because this covers the entire scope, unlike the other case.

OrbitLinuxTracing/TracerThread.cpp
dpallotti: Without these braces this would overlap with `ORBIT_SCOPE("ProcessOldEvents");`.

OrbitLinuxTracing/TracerThread.cpp
dpallotti: I had actually called this "TracerThread::ProcessDeferredEvents" and then decided against it for the reason I mentioned.

OrbitLinuxTracing/TracerThread.cpp
ronaldfw: Ah, I mis-matched the braces, thought this was the only part in the else. Disregard then.

OrbitLinuxTracing/TracerThread.cpp
ronaldfw: Misunderstood the other case, makes sense.

OrbitLinuxTracing/TracerThread.cpp
ronaldfw: I generally think it's better to have a clear consistent rule for naming stuff like this, but I get the reasoning.

dpallotti: I'm generally all for consistency but as I said I tried the longer version and it's really a bit too long for the use we do of this, so I really prefer the current one.

PR #1513, opened 2020-11-27
Solve strange little jump when dragging

karupayun:
Really noticeable when you zoom-in to see details in a particular timer. As soon as
you start dragging the screen jumps to the left, and it makes really difficult to navigate.

Also solve b/171190211.


vickyliu-go4it (approved):
LGTM.  It helps and thanks for that : )

reichlfl (approved):
LGTM, thanks for digging into this and fixing this annoying issue!

PR #1390, opened 2020-10-22
Zooming / Scrolling fixes

reichlfl:
This makes the update relation of world and the scrollbar UI elements clearer and fixes various smaller issues related to scrollbar updates:

b/169552431
b/167190887

Also added the help text for vertical zoom: b/159313319

PR #1515, opened 2020-11-27
Solve mini-jump after each position change (Revert)

karupayun:
Regression from 1.54. After every zooming, or other position changes,
the screen glitched and make a small movement. Noticeable after
zooming-in a lot in the horizontal coordinates.

Small revert from https://github.com/google/orbit/pull/1390.


reichlfl (approved):
As commented in GVC: I thought there was a reason for adding those two lines there, but since we now both couldn't see any issues, this is probably no longer the case.

Thanks for finding this, LGTM!

PR #1517, opened 2020-11-30
Fix the samples jumping around when scrolling

reichlfl:
To reproduce the issue:
* Scroll down by any amount vertically (vertical scroll value must be != 0)
* Zoom in / out horizontally, or move the horizontal slider
-> Samples on event tracks jump around randomly be 1 pixel up/down

b/170334307

### DIFF
--- a/OrbitGl/Batcher.cpp
+++ b/OrbitGl/Batcher.cpp
@@ -4,6 +4,8 @@
 
 #include "Batcher.h"
 
+#include <math.h>
+
 #include "CoreUtils.h"
 #include "OpenGl.h"
 
@@ -40,8 +42,8 @@ void Batcher::AddVerticalLine(Vec2 pos, float size, float z, const Color& color,
 void Batcher::AddLine(Vec2 from, Vec2 to, float z, const Color& color, const Color& picking_color,
                       std::unique_ptr<PickingUserData> user_data) {
   Line line;
-  line.start_point = Vec3(from[0], from[1], z);
-  line.end_point = Vec3(to[0], to[1], z);
+  line.start_point = Vec3(floorf(from[0]), floorf(from[1]), z);
+  line.end_point = Vec3(floorf(to[0]), floorf(to[1]), z);
   auto& buffer = primitive_buffers_by_layer_[z];
 
   buffer.line_buffer.lines_.push_back(line);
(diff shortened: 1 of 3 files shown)


OrbitGl/Batcher.cpp
```diff
@@ -40,8 +42,8 @@ void Batcher::AddVerticalLine(Vec2 pos, float size, float z, const Color& color,
 void Batcher::AddLine(Vec2 from, Vec2 to, float z, const Color& color, const Color& picking_color,
                       std::unique_ptr<PickingUserData> user_data) {
   Line line;
-  line.start_point = Vec3(from[0], from[1], z);
-  line.end_point = Vec3(to[0], to[1], z);
+  line.start_point = Vec3(floorf(from[0]), floorf(from[1]), z);
```
karupayun, line 45: I would like to have some comment explaining why we are making this floorf.

karupayun (approved):
Very nice catch Florian! As I told you before, I'm really happy with this solution in Batcher, I think that it's the right way to go, but prefer to wait for more comments.

1- The solution now seems to work very well 💯 
2- Samples are still behaving strangely. They are sometimes drawn on the top of the track, and sometimes on the bottom. Happily, all of them are drawn on the same side.

pierricgimmig:
@reichlfl , if I understand correctly, this will constrain the world coordinates of primitives to always be at pixel boundary, essentially using floats to express ints. I think this will work as long as our view frustum size in world coordinates matches the canvas size in pixels. Going forward with the change we should keep that in mind. For example, if we ever wanted to draw using standard world coordinates (player position in the world for example), then we'd need to have a different Batcher that wouldn't truncate the floating point values.

I thought that setting the vertex specification to use ints instead of floats for our capture window could be a solution (`glVertexPointer(3, GL_INT, ...)`), but it might complicate things too much on our side. OpenGl would simply convert ints to floats on its side anyway.

pierricgimmig (approved):

OrbitGl/Batcher.cpp
reichlfl: Good point. I've added a documentation for the Batcher-class to the header explaining that this is done and what the reasoning behind it is.

reichlfl:
> 2- Samples are still behaving strangely. They are sometimes drawn on the top of the track, and sometimes on the bottom. Happily, all of them are drawn on the same side.

IMO they should be the same height as the track... I wanted to address that in a separate PR and clean up the EventTrack.

reichlfl:
> @reichlfl , if I understand correctly, this will constrain the world coordinates of primitives to always be at pixel boundary, essentially using floats to express ints. I think this will work as long as our view frustum size in world coordinates matches the canvas size in pixels. Going forward with the change we should keep that in mind. For example, if we ever wanted to draw using standard world coordinates (player position in the world for example), then we'd need to have a different Batcher that wouldn't truncate the floating point values.
> 
> I thought that setting the vertex specification to use ints instead of floats for our capture window could be a solution (`glVertexPointer(3, GL_INT, ...)`), but it might complicate things too much on our side. OpenGl would simply convert ints to floats on its side anyway.

Exactly. If we e.g. choose to implement zooming in a different way, this needs to be changed. But this would probably also mean the problem is back...

I was also thinking about expanding the Batcher to inherit from a base class that provides only the basic AddPrimitive methods (a single AddLine, AddBox, AddTriangle) which can then take care of rounding, or have it configurable through a member. But since this is potentially performance-critical, I wanted to avoid adding more and more indirections, I've already added additional loops and copy-operations for the rounding now, this may get out of hand.

And basically, GL_INT would come with the same drawbacks without a real upside.

karupayun:
> if I understand correctly, this will constrain the world coordinates of primitives to always be at pixel boundary, essentially using floats to express ints. I think this will work as long as our view frustum size in world coordinates matches the canvas size in pixels. Going forward with the change we should keep that in mind. For example, if we ever wanted to draw using standard world coordinates (player position in the world for example), then we'd need to have a different Batcher that wouldn't truncate the floating point values.

Being the fact that we couldn't fix the root issue (why OpenGl draws differently based on x-coordinate), I'm really happy with the solution because it looks cleaner. But I agree that we should be opened to revert the PR if we find a better solution.

PR #1521, opened 2020-11-30
    Allow Windows builds with system-installed Qt

beckerhe:
We can't provide binary packages of Qt for the public and compiling
Qt from source takes a long time and is cumbersome.

To solve that problem this introduces the possibility
to use a system-installed Qt as we support it on Linux
for a long time.

This new support is optional and we will continue building
our own Qt for the official Stadia version of Orbit. But
for external contributors that should simplify getting started
on Windows.

### DIFF
--- a/build.ps1
+++ b/build.ps1
@@ -31,9 +31,28 @@ function conan_create_profile($profile) {
   $build_type = $profile -replace "default_"
   $profile_path = "$conan_dir/profiles/$profile"
 
+  if (!$Env:Qt5_DIR) {
+    Write-Host ""
+    Write-Host "Compile Qt from source or use official distribution?"
+    Write-Host "====================================================`r`n"
+    Write-Host "Orbit depends on the Qt framework which can be either automatically compiled from source (can take several hours!)"
+    Write-Host "or can be provided by an official Qt distribution. We recommend the latter which is less cumbersome."
+    Write-Host "Check out DEVELOPMENT.md for more details on how to use an official Qt distribution.`r`n"
+    Write-Host "Press Ctrl+C to stop here and install Qt first. Press Enter to continue with compiling Qt from source."
+    Write-Host "Google-internal devs: Please press Enter!`r`n"
+    Read-Host  "Answer"
+  }
+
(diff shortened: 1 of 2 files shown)


build.ps1
```diff
@@ -31,9 +31,31 @@ function conan_create_profile($profile) {
   $build_type = $profile -replace "default_"
   $profile_path = "$conan_dir/profiles/$profile"
 
+  if (!$Env:Qt5_DIR) {
+    Write-Host ""
+    Write-Host "Compile Qt from source or use official distribution?"
+    Write-Host "====================================================`r`n"
+    Write-Host "Orbit depends on the Qt framework which can be either automatically compiled from source (takes a long time),"
```
pierricgimmig, line 38: nit. I think we should emphasize that building Qt from source could take *hours*.

"...compiled from source (can take hours)," or something along those lines?

build.ps1
```diff
@@ -31,9 +31,31 @@ function conan_create_profile($profile) {
...
+  if ($Env:Qt5_DIR) {
+    Write-Host "Found Qt5_DIR environment variable. Using system provided Qt distribution."
+    "OrbitProfiler:system_qt=True" | Add-Content -Path $profile_path
+  }
+
   "[build_requires]`r`n[env]" | Add-Content -Path $profile_path
+
+  if ($Env:Qt5_DIR) {
+    "OrbitProfiler:Qt5_DIR=`"$Env:Qt5_DIR`"" | Add-Content -Path $profile_path
+  }
```
pierricgimmig, line 58: Can the two "if ($Env:Qt5_DIR) {" blocks be combined?

pierricgimmig (approved):
Hi Henning, thanks for looking into this, for the detailed doc and the user prompt. Please see minor comments, otherwise LGTM.

build.ps1
beckerhe: Yes, makes sense! Done!

build.ps1
beckerhe: Yes, sure. Done!

PR #1524, opened 2020-12-01
Add CaptureEventProducer and LockFreeBufferCaptureEventProducer

dpallotti:
For the communication between producers and OrbitService. Bug: http://b/174028631.


OrbitProducer/include/OrbitProducer/LockFreeBufferCaptureEventProducer.h
```diff
@@ -0,0 +1,141 @@
...
+  // Subclasses need to implement this method to convert an IntermediateEventT enqueued
+  // in the internal lock-free buffer to a CaptureEvent to be sent to ProducerSideService.
+  [[nodiscard]] virtual orbit_grpc_protos::CaptureEvent TranslateIntermediateEvent(
+      IntermediateEventT&& intermediate_event) = 0;
+
+ private:
+  void ForwarderThread() {
+    constexpr uint64_t kMaxEventsPerRequest = 10'000;
+    std::vector<IntermediateEventT> dequeued_events;
+    dequeued_events.resize(kMaxEventsPerRequest);
```
dimitry-, line 87: nit: you can do it in c-tor.

OrbitProducer/CaptureEventProducer.cpp
```diff
@@ -0,0 +1,203 @@
...
+  // the internal grpc_shutdown will be executed. This can use a detached thread.
+  // From the docs of grpc_shutdown: "The last call to grpc_shutdown will initiate
+  // cleaning up of grpc library internals, which can happen in another thread".
+  // Give that a moment to complete, as otherwise that could lead to a SIGSEGV on exit.
+  std::this_thread::sleep_for(std::chrono::milliseconds{1});
+}
+
+void CaptureEventProducer::OnCaptureStart() { LOG("CaptureEventProducer called OnCaptureStart"); }
+
+void CaptureEventProducer::OnCaptureStop() { LOG("CaptureEventProducer called OnCaptureStop"); }
```
dimitry-, line 55: these 2 look like a separate interface

OrbitProducer/CaptureEventProducer.cpp
```diff
@@ -0,0 +1,203 @@
...
+
+  CHECK(producer_side_service_stub_ != nullptr);
+  {
+    // Acquiring the mutex just for the CHECK might seem expensive,
+    // but the gRPC call that follows is orders of magnitude slower.
+    absl::ReaderMutexLock lock{&shutdown_requested_mutex_};
+    CHECK(!shutdown_requested_);
+  }
+
+  bool write_succeeded;
```
ronaldfw, line 70: I would initialize with a sane default value here, `bool write_succeeded = false`. (And elsewhere.)

OrbitProducer/CaptureEventProducer.cpp
```diff
@@ -0,0 +1,203 @@
...
+      stream_ = producer_side_service_stub_->ReceiveCommandsAndSendEvents(context_.get());
+    }
+
+    if (stream_ == nullptr) {
+      ERROR(
+          "Calling ReceiveCommandsAndSendEvents to establish "
+          "gRPC connection with ProducerSideService");
+      // This is the reason why we protect shutdown_requested_ with a Mutex instead of
+      // using an std::atomic<bool>: so that we can use (Reader)LockWhenWithTimeout.
+      if (shutdown_requested_mutex_.ReaderLockWhenWithTimeout(
```
florian-kuebler, line 136: Is that a fancy way of waiting `reconnection_delay_ms_`, or stopping early when shutdown was requested?
If so, I'd also put this in the comment.

OrbitProducer/CaptureEventProducer.cpp
```diff
@@ -0,0 +1,203 @@
...
+      absl::ReaderMutexLock lock{&context_and_stream_mutex_};
+      context_ = std::make_unique<grpc::ClientContext>();
+      stream_ = producer_side_service_stub_->ReceiveCommandsAndSendEvents(context_.get());
+    }
+
+    if (stream_ == nullptr) {
+      ERROR(
+          "Calling ReceiveCommandsAndSendEvents to establish "
+          "gRPC connection with ProducerSideService");
+      // This is the reason why we protect shutdown_requested_ with a Mutex instead of
```
ronaldfw, line 134: Nit: Mutex -> mutex

OrbitProducer/CaptureEventProducer.cpp
```diff
@@ -0,0 +1,203 @@
...
+              absl::Milliseconds(static_cast<int64_t>(reconnection_delay_ms_)))) {
+        shutdown_requested_mutex_.ReaderUnlock();
+      }
+      continue;
+    }
+    LOG("Called ReceiveCommandsAndSendEvents on ProducerSideService");
+
+    while (true) {
+      ReceiveCommandsAndSendEventsResponse response;
+      bool read_succeeded;
```
ronaldfw, line 149: Also initialize with `false`.

OrbitProducer/FakeProducerSideService.h
```diff
@@ -0,0 +1,91 @@
...
+
+  MOCK_METHOD(void, OnCaptureEventsReceived, (int32_t count), ());
+  MOCK_METHOD(void, OnAllEventsSentReceived, (), ());
+
+ private:
+  grpc::ServerContext* context_ = nullptr;
+  grpc::ServerReaderWriter<orbit_grpc_protos::ReceiveCommandsAndSendEventsResponse,
+                           orbit_grpc_protos::ReceiveCommandsAndSendEventsRequest>* stream_ =
+      nullptr;
+  std::atomic<bool> rpc_allowed = true;
```
ronaldfw, line 86: Member variable should be named `rpc_allowed_`

OrbitProducer/LockFreeBufferCaptureEventProducerTest.cpp
```diff
@@ -0,0 +1,345 @@
...
+  void StopService() {
+    fake_service->FinishAndDisallowRpc();
+    fake_server->Shutdown();
+    fake_server->Wait();
+
+    fake_service.reset();
+    fake_server.reset();
+  }
+
+  std::optional<FakeProducerSideService> fake_service;
```
ronaldfw, line 64: Member vars should end in `_`.

ronaldfw (commented):
The code looks very good and it's great to see the extensive unit tests. I have some minor comments only. 

I know you disagree on initialization of variables, but I think it's overall safer (also for future modifications) to just initialize bools with a safe default value (see a few of my comments).

OrbitProducer/include/OrbitProducer/LockFreeBufferCaptureEventProducer.h
```diff
@@ -0,0 +1,141 @@
...
+// This still abstract implementation of CaptureEventProducer provides a lock-free queue
+// where to write events with low overhead from the fast path where they are produced.
+// The methods EnqueueIntermediateEvent(IfCapturing) allow to enqueue those events.
+//
+// Internally, a thread reads from the lock-free queue and sends CaptureEvents
+// to ProducerSideService using the methods provided by the superclass.
+//
+// Note that the events stored in the lock-free queue, whose type is specified by the
+// type parameter IntermediateEventT, don't need to be CaptureEvents, nor protobufs at all.
+// This is to allow enqueuing objects that are faster to produce than protobufs.
```
florian-kuebler, line 23: Whats really the difference here? How can it be faster to produce a non CaptureEvent and let it then translate here to an CaptureEvent then just storing CaptureEvents directly (and creating that on the user side)?
Is it about potential copies of the proto?

florian-kuebler (approved):
Thanks for this change. Looks very good to me.

OrbitProducer/include/OrbitProducer/CaptureEventProducer.h
```diff
@@ -0,0 +1,75 @@
...
+  // Subclasses that extend this method by overriding it must also call the overridden method.
+  virtual void BuildAndStart(const std::shared_ptr<grpc::Channel>& channel);
+  // This method causes to disconnect from ProducerSideService or to stop trying to reconnect to it.
+  // Subclasses that extend this method by overriding it must also call the overridden method.
+  virtual void ShutdownAndWait();
+
+  // Subclasses can override this method to be notified of a request to start a capture.
+  virtual void OnCaptureStart();
+  // Subclasses can override this method to be notified of a request to stop the capture.
+  virtual void OnCaptureStop();
```
dimitry-, line 41: Make them pure virtual for clarity, and you won't need virtual but implemented d-tor trick anymore.

dimitry- (approved):
Otherwise lgtm

OrbitProducer/CaptureEventProducer.cpp
dpallotti: Discussed offline with Dimitry. `OnCaptureStart`,`OnCaptureStop` shouldn't be separated from `SendCaptureEvents`,`NotifyAllEventsSent` as they basically represent the implementation of a client of `ProducerSideService` as defined in `producer_side_services.proto`.

OrbitProducer/CaptureEventProducer.cpp
dpallotti: I had done that. But the IDE's lint warns me that "The value is never used" (because it's always overwritten). On the other hand using the variable before it has been assigned results in a compilation error (`error: variable '...' is uninitialized when used here [-Werror,-Wuninitialized]`). Not sure it will convince you but that's the reasoning...

OrbitProducer/CaptureEventProducer.cpp
dpallotti: This is intended, as I'm not referring to any mutex, but to `absl::Mutex`, which supports `LockWhenWithTimeout`. Added `absl::` in the comment.

OrbitProducer/CaptureEventProducer.cpp
dpallotti: Yes, we are using this pattern more and more. Expanded the comment.

OrbitProducer/FakeProducerSideService.h
dpallotti: Done.

OrbitProducer/LockFreeBufferCaptureEventProducerTest.cpp
dpallotti: As this is a test fixture, this was on purpose to avoid those underscores everywhere in the tests. But in all the examples I found of fixtures they use the trailing underscore so I adapted.

OrbitProducer/include/OrbitProducer/CaptureEventProducer.h
```diff
@@ -0,0 +1,75 @@
...
+
+#include "absl/synchronization/mutex.h"
+#include "grpcpp/grpcpp.h"
+#include "producer_side_services.grpc.pb.h"
+
+namespace orbit_producer {
+
+// This abstract class offers the subclasses methods
+// to connect and communicate with a ProducerSideService.
+class CaptureEventProducer {
```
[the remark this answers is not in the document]
dpallotti: Discussed offline. Other discussed solutions have their own disadvantages.

OrbitProducer/include/OrbitProducer/CaptureEventProducer.h
dpallotti: Done.

OrbitProducer/include/OrbitProducer/LockFreeBufferCaptureEventProducer.h
dpallotti: The difference is that the translation is done by the reader, not by the writer, and the latter is the one that needs to be fast.

OrbitProducer/include/OrbitProducer/LockFreeBufferCaptureEventProducer.h
dpallotti: This is a local variable (and there's no need for it to not be a local variable).

OrbitProducer/include/OrbitProducer/LockFreeBufferCaptureEventProducer.h
dpallotti: By the way, I'm not saying that protobufs are so slow, but they do have an overhead (if I understand correctly, mostly because of the memory management, as all fields are optional) which in some cases can be relevant if compared to a simple struct of a few bytes.

OrbitProducer/include/OrbitProducer/LockFreeBufferCaptureEventProducer.h
dimitry-: replace 86 and 87 with
std::vector<IntermediateEventT> dequeued_events(kMaxEventsPerRequest);

dpallotti: Ah ok, sorry, I had obviously read your comment completely wrong...

OrbitProducer/include/OrbitProducer/LockFreeBufferCaptureEventProducer.h
florian-kuebler: Okay. Lets see how the users are going to look like. If every implementation will use CaptureEvent anyways, we can remove that later.

OrbitProducer/include/OrbitProducer/LockFreeBufferCaptureEventProducer.h
dpallotti: Done.

OrbitProducer/CaptureEventProducer.cpp
ronaldfw: Ok, that's interesting. Together with the "uninitialized" warning, this is fine and actually arguably better than initializing with a default value.

PR #1525, opened 2020-12-01
Remove unused files from repo root

beckerhe:
This removes the Windows batch script which have been deprecated and
converted to powershell a long time ago.

Furthermore this remove the windows cmake readme which is outdated.

I believe `pack.bat` is also not in use anymore.


dpallotti (approved):
Thanks!

By the way, what about `release_notes.txt`?

pierricgimmig (approved):
Yes, thanks Henning, I've been wanting to do the same!

beckerhe:
> Thanks!
> 
> By the way, what about `release_notes.txt`?

Yes, good Point. @pierricgimmig Do you have any strong feelings about the release_notes.txt file?
Since it's not up-to-date it might be confusing to new potential contributors.

pierricgimmig:
> Yes, good Point. @pierricgimmig Do you have any strong feelings about the release_notes.txt file?

@beckerhe , this flew under my radar, but let's remove it yes!

PR #1528, opened 2020-12-01
Improve OrbitGgp::Client

antonrohr:
There used to be situations where some calls to ggp would fail, but this was not recognized until a timeout was triggered (or not recognized at all). This is now fixed.

I am aware that the whole class (and probably all of OrbitGgp) could use some refactoring. I would like to use more signals and slots and the `ErrorMessageOr` return type. Maybe I get to more, as part of the fixit week.

### DIFF
--- a/OrbitGgp/Client.cpp
+++ b/OrbitGgp/Client.cpp
@@ -8,6 +8,7 @@
 #include <QPointer>
 #include <QProcess>
 #include <QTimer>
+#include <optional>
 
 #include "OrbitBase/Logging.h"
 #include "OrbitGgp/Error.h"
@@ -47,7 +48,7 @@ void RunProcessWithTimeout(const QString& program, const QStringList& arguments,
                    parent,
                    [process, timeout_timer, callback](const int exit_code,
                                                       const QProcess::ExitStatus exit_status) {
-                     if (timeout_timer) {
+                     if (timeout_timer != nullptr) {
                        timeout_timer->stop();
                        timeout_timer->deleteLater();
                      }
@@ -56,7 +57,7 @@ void RunProcessWithTimeout(const QString& program, const QStringList& arguments,
                        ERROR(
                            "Ggp list instances request failed with error: %s (exit code: "
                            "%d)",
-                           process->errorString().toStdString().c_str(), exit_code);
+                           process->errorString().toStdString(), exit_code);
                        callback(Error::kGgpListInstancesFailed);
                        return;
                      }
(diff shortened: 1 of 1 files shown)


OrbitGgp/Client.cpp
```diff
@@ -66,6 +67,17 @@ void RunProcessWithTimeout(const QString& program, const QStringList& arguments,
                      process->deleteLater();
                    });
 
+  QObject::connect(process, &QProcess::errorOccurred, parent, [timeout_timer, process, callback]() {
+    if (timeout_timer != nullptr) {
+      timeout_timer->stop();
+      timeout_timer->deleteLater();
+    }
+    ERROR("Ggp list instances request failed with error: %s",
+          process->errorString().toStdString().c_str());
```
dimitry-, line 76: nit: you can drop c_str() at the end

dimitry- (approved):
otherwise lgtm

OrbitGgp/Client.cpp
```diff
@@ -76,13 +88,25 @@ outcome::result<QPointer<Client>> Client::Create(QObject* parent) {
   QProcess ggp_process{};
   ggp_process.setProgram("ggp");
   ggp_process.setArguments({"version"});
+
+  std::optional<Error> error;
+  QObject::connect(&ggp_process, &QProcess::errorOccurred, parent, [&ggp_process, &error] {
+    ERROR("Ggp command line process failed with error: %s",
+          ggp_process.errorString().toStdString());
+    error = Error::kCouldNotUseGgpCli;
+  });
```
beckerhe, line 97: What situation does this slot cover that is not covered by the `if (ggp_process.exitStatus()...` below?
Since this is all synchronous I don't really understand when this could potentially be called while the if from below is NOT true.

antonrohr: It took me some testing to realize that on if the process can't be started at all (e.g because ggp is not installed/ found in path), then the `if` branch is not taken. This happens only on windows as far as I can tell. 

I tested this by just renaming the executable name `ggp` to something that does not exit.

beckerhe: Ah great. Thanks for the explanation. That makes sense.

I would say this needs at least a comment for explanation. How do you feel about adding an additional condition to the if instead?

This might work:

```
if (ggp_process.exitStatus() != QProcess::NormalExit || ggp_process.exitCode() != 0 || ggp_process.error() != QProcess::UnknownError) {
```

antonrohr: This is even better, thank you!

OrbitGgp/Client.cpp
antonrohr: done

beckerhe (approved):
Great Anton! Thanks a lot! LGTM!

PR #1529, opened 2020-12-01
Configure CMake Graphviz feature.

danielfenner:
CMake can generate Graphviz files showing the dependencies between the targets
in a project, as well as external libraries which are linked against.

This configuration removes the clutter from the dependency graphs of OrbitQt
and OrbitService.


beckerhe:
Hi,
how does this interact with CMake? I would assume this needs to be somehow included in the main CMakeLists.txt, doesn't it?

florian-kuebler (approved):
That looks good to me. Thanks for looking into this!

danielfenner:
> Hi,
> how does this interact with CMake? I would assume this needs to be somehow included in the main CMakeLists.txt, doesn't it?

Good point. I intended to share the doc with the context but forgot. Done that now. 

tl;dr 
```
cd build_default_relwithdebinfo
cmake --graphviz=dependencies_orbit_qt.dot ..
dot -Tpng -o dependencies_orbit_qt.png dependencies_orbit_qt.dot
```

The filename CMakeGraphVizOptions.cmake is magic to CMake and the config gets picked up automatically.

beckerhe:
> > Hi,
> > how does this interact with CMake? I would assume this needs to be somehow included in the main CMakeLists.txt, doesn't it?
> 
> Good point. I intended to share the doc with the context but forgot. Done that now.
> 
> tl;dr
> 
> ```
> cd build_default_relwithdebinfo
> cmake --graphviz=dependencies_orbit_qt.dot ..
> dot -Tpng -o dependencies_orbit_qt.png dependencies_orbit_qt.dot
> ```
> 
> The filename CMakeGraphVizOptions.cmake is magic to CMake and the config gets picked up automatically.

Ah I see. I had a look at the code. It seems `CMakeGraphVizOptions` is a CMake module  which means it is resolved via the `CMAKE_MODULE_PATH`. Our `cmake/` subdirectory is in the module path, so my guess is that it would work fine when the file lives there. It would be great if you could give that a try since I'm currently trying to remove all unnecessary files from the repo root.

danielfenner:
> Ah I see. I had a look at the code. It seems `CMakeGraphVizOptions` is a CMake module which means it is resolved via the `CMAKE_MODULE_PATH`. Our `cmake/` subdirectory is in the module path, so my guess is that it would work fine when the file lives there. It would be great if you could give that a try since I'm currently trying to remove all unnecessary files from the repo root.

Okay, that did not work. 

https://cmake.org/cmake/help/latest/module/CMakeGraphVizOptions.html
> This file is first searched in CMAKE_BINARY_DIR, and then in CMAKE_SOURCE_DIR.

CMAKE_SOURCE_DIR is the repo root. CMAKE_BINARY_DIR are the ephemeral build... directories I guess.

beckerhe:
> > Ah I see. I had a look at the code. It seems `CMakeGraphVizOptions` is a CMake module which means it is resolved via the `CMAKE_MODULE_PATH`. Our `cmake/` subdirectory is in the module path, so my guess is that it would work fine when the file lives there. It would be great if you could give that a try since I'm currently trying to remove all unnecessary files from the repo root.
> 
> Okay, that did not work.
> 
> https://cmake.org/cmake/help/latest/module/CMakeGraphVizOptions.html
> 
> > This file is first searched in CMAKE_BINARY_DIR, and then in CMAKE_SOURCE_DIR.
> 
> CMAKE_SOURCE_DIR is the repo root. CMAKE_BINARY_DIR are the ephemeral build... directories I guess.

I see. Thanks for pointing that out. That's definitely a weird CMake module.

CMakeGraphVizOptions.cmake
```diff
@@ -0,0 +1,23 @@
+# Clean up stuff in OrbitQt
```
beckerhe, line 1: I guess that file needs a license header

beckerhe (approved):
Thanks!

CMakeGraphVizOptions.cmake
danielfenner: done

PR #1530, opened 2020-12-02
Fix track filter background image

reichlfl:
The image was showing up in the right-click context because it was not
restricted to the QLineEdit class

To reproduce:
- Right-click "Filter tracks" textbox
- The "list filter" icon is set as background for the menu

After the fix:
- Context menu looks correct

b/173577095


karupayun (approved):
LGTM! :)

PR #1531, opened 2020-12-02
Determine target branch in presubmit checks

beckerhe:
This prepares our two non-build presubmit checks (clang-format,
license_headers) for the upcoming master to main change by replacing the
hard-coded "origin/master" branch-name in the scripts by an enviroment
variable provided by kokoro.

Additionaly this applies some best practives for shell programming.

### DIFF
--- a/kokoro/checks/clang_format/check.sh
+++ b/kokoro/checks/clang_format/check.sh
@@ -5,8 +5,10 @@
 # found in the LICENSE file.
 
 # Fail on any error.
-DIR="/mnt/github/orbitprofiler"
-SCRIPT="${DIR}/kokoro/checks/clang_format/check.sh"
+set -euo pipefail
+
+readonly REPO_ROOT="$( cd "$( dirname "${BASH_SOURCE[0]}" )/../../../" >/dev/null 2>&1 && pwd )"
+readonly SCRIPT="/mnt/kokoro/checks/clang_format/check.sh"
 
 if [ "$0" == "$SCRIPT" ]; then
   # We are inside the docker container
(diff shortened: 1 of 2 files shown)


kokoro/checks/clang_format/check.sh
```diff
@@ -16,17 +18,16 @@ if [ "$0" == "$SCRIPT" ]; then
...
-  done <<< $(find "$DIR" -name '*.cpp' -o -name '*.h' \
+  done <<< $(find /mnt -name '*.cpp' -o -name '*.h' \
         | grep -v "third_party/" \
         | grep -v "/build" )
   echo -e "--\n"
 
-  cd "$DIR"
-  REFERENCE="origin/master"
+  cd /mnt
+  REFERENCE="${KOKORO_GITHUB_PULL_REQUEST_TARGET_BRANCH:-origin/master}"
```
florian-kuebler, line 30: Why does the REFERENCE variable here still contain the origin/master name? shouldn't that be set from outside?

florian-kuebler (commented):
Thanks for that change! From reading the description I though you would remove the term "master" already completely from the scripts and already pass the branch into them.

kokoro/checks/clang_format/check.sh
beckerhe: It's a fallback when "KOKORO_GITHUB_PULL_REQUEST_TARGET_BRANCH" is not set. (Bash syntax ${name:-fallback_string}")

beckerhe: This allows the script to be called on a standard linux machine for testing purposes. We can change that later to "main" without any pressure.

florian-kuebler: Okay, thanks!

florian-kuebler: Can you maybe add a comment there?

florian-kuebler (approved):

kokoro/checks/clang_format/check.sh
beckerhe: Sure. About what exactly? THe master/main thing? Or about the default string? Regarding the latter I'm a bit hesitant because I feel it's a standard language construct widely used for that purpose and pretty clear on its own.

florian-kuebler: If it is that widely used, then feel free to not add a comment.
Otherwise, I was referring to the default string. Something like "Use orgin/master as reference branch, if not specified otherwise."

beckerhe: Yeah. I can live with that! :sweat_smile: Actually not too bad. Thanks!

beckerhe: Done.

PR #1532, opened 2020-12-02
Parse and send module updates to the client

dimitry-:
In event of a module load/unload during capture send
updated modules list to the client.

Bug: http://b/168797897
Test: capture

### DIFF
--- a/OrbitCaptureClient/CaptureEventProcessor.cpp
+++ b/OrbitCaptureClient/CaptureEventProcessor.cpp
@@ -66,6 +66,8 @@ void CaptureEventProcessor::ProcessEvent(const CaptureEvent& event) {
       break;
     case CaptureEvent::kGpuQueueSubmission:
       UNREACHABLE();
+    case CaptureEvent::kModulesUpdateEvent:
+      // TODO (http://b/168797897): Process module update events
       break;
     case CaptureEvent::EVENT_NOT_SET:
       ERROR("CaptureEvent::EVENT_NOT_SET read from Capture's gRPC stream");
--- a/OrbitGrpcProtos/capture.proto
+++ b/OrbitGrpcProtos/capture.proto
@@ -6,6 +6,7 @@ syntax = "proto3";
 
 package orbit_grpc_protos;
 
+import "module.proto";
 import "tracepoint.proto";
 
 message CaptureOptions {
@@ -224,6 +225,12 @@ message AddressInfo {
   }
 }
 
+message ModulesUpdateEvent {
+  int32 pid = 1;
+  uint64 timestamp_ns = 2;
+  repeated ModuleInfo modules = 3;
+}
+
 message CaptureEvent {
   oneof event {
     SchedulingSlice scheduling_slice = 1;
@@ -239,5 +246,6 @@ message CaptureEvent {
     InternedTracepointInfo interned_tracepoint_info = 9;
     TracepointEvent tracepoint_event = 10;
     IntrospectionScope introspection_scope = 12;
+    ModulesUpdateEvent modules_update_event = 14;
   }
 }
--- a/OrbitLinuxTracing/TracerThread.cpp
+++ b/OrbitLinuxTracing/TracerThread.cpp
@@ -835,7 +835,7 @@ void TracerThread::ProcessMmapEvent(const perf_event_header& header,
 
   // There was a call to mmap with PROT_EXEC, hence refresh the maps.
   // This should happen rarely.
-  auto event = std::make_unique<MapsPerfEvent>(MonotonicTimestampNs(), ReadMaps(pid_));
+  auto event = std::make_unique<MapsPerfEvent>(pid_, MonotonicTimestampNs(), ReadMaps(pid_));
   event->SetOriginFileDescriptor(ring_buffer->GetFileDescriptor());
   DeferEvent(std::move(event));
 }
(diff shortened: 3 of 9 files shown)


OrbitGrpcProtos/capture.proto
```diff
@@ -177,6 +178,11 @@ message AddressInfo {
   }
 }
 
+message ModulesUpdateEvent {
```
dpallotti, line 181: This should also have `pid`.

OrbitLinuxTracing/include/OrbitLinuxTracing/TracerListener.h
```diff
@@ -21,6 +21,7 @@ class TracerListener {
   virtual void OnThreadStateSlice(orbit_grpc_protos::ThreadStateSlice thread_state_slice) = 0;
   virtual void OnAddressInfo(orbit_grpc_protos::AddressInfo address_info) = 0;
   virtual void OnTracepointEvent(orbit_grpc_protos::TracepointEvent tracepoint_event) = 0;
+  virtual void OnMapsUpdate(uint64_t timestamp, const std::string& maps) = 0;
```
dpallotti, line 24: Notice how this method is so inconsistent compared to all the others. They all accept protobufs from `orbit_grpc_protos` and this doesn't. I think that this should take `ModulesUpdateEvent` (maybe renamed to `MapsUpdateEvent`) and that should carry unparsed `maps`, but then those maps should be parsed anyway on the service and the protobuf translated. So more easily (and see the other comment), it could carry maps already parsed in `OrbitLinuxTracing`.

OrbitService/LinuxTracingHandler.cpp
```diff
@@ -143,6 +144,22 @@ void LinuxTracingHandler::OnTracepointEvent(orbit_grpc_protos::TracepointEvent t
   capture_event_buffer_->AddEvent(std::move(event));
 }
 
+void LinuxTracingHandler::OnMapsUpdate(uint64_t timestamp, const std::string& maps) {
+  auto result_or_error = utils::ParseMaps(maps);
```
dpallotti, line 148: This is related to the other comment. This class does buffering and interning. This processing doesn't really belong here, it seems like it's done here just to use `ServiceUtils.h`. Is there a reason why this processing shouldn't be done already in `OrbitLinuxTracing`?

OrbitGrpcProtos/capture.proto
dimitry-: right - done.

OrbitService/LinuxTracingHandler.cpp
dimitry-: we need to extract additional information (load_bias and build_id), this is not currently available by parsing done in LinuxOrbitTracing, also note that OrbitLinuxTracing produces libunwindstack related structures which we need to convert and complete. It is possible to move ParseMaps to OrbitLinuxTracing but it would also introduce some additional dependencies.

dpallotti: > OrbitLinuxTracing produces libunwindstack related structures which we need to convert and complete

But those are only used internally by `OrbitLinuxTracing` and don't leave it, or am I missing what you meant?

> It is possible to move ParseMaps to OrbitLinuxTracing but it would also introduce some additional dependencies.

Are there particular dependencies other than `ElfFile`? Considering that `OrbitLinuxTracing` depends on `libunwindstack`, I wouldn't be against depending also on `ElfFile` or the like.

In general `LinuxTracingHandler` seems a bit of a limbo to put this processing.

OrbitCaptureClient/CaptureEventProcessor.cpp
```diff
@@ -64,6 +64,9 @@ void CaptureEventProcessor::ProcessEvent(const CaptureEvent& event) {
     case CaptureEvent::kTracepointEvent:
       ProcessTracepointEvent(event.tracepoint_event());
       break;
+    case CaptureEvent::kModulesUpdateEvent:
+      // TODO (http://b/168797897): Process module update events
```
antonrohr, line 68: Should we maybe log something in this case?

antonrohr (approved):
I do not really know OrbitLinuxTracing and its architecture, but I feel this part of the review is in good hands with Davide.

OrbitCaptureClient/CaptureEventProcessor.cpp
dimitry-: I am addressing it in the next change. But still - will do.

florian-kuebler: Have you considered adding a test for that already? :)

OrbitLinuxTracing/include/OrbitLinuxTracing/TracerListener.h
dimitry-: as discussed offline - moved some code around to be able to parse maps in OrbitLinuxTracing. Note that first commit is PR#1558

OrbitService/LinuxTracingHandler.cpp
dimitry-: see above.

OrbitLinuxTracing/TracerThread.cpp
```diff
@@ -835,7 +835,7 @@ void TracerThread::ProcessMmapEvent(const perf_event_header& header,
 
   // There was a call to mmap with PROT_EXEC, hence refresh the maps.
   // This should happen rarely.
-  auto event = std::make_unique<MapsPerfEvent>(MonotonicTimestampNs(), ReadMaps(pid_));
+  auto event = std::make_unique<MapsPerfEvent>(pid_, MonotonicTimestampNs(), ReadMaps(pid_));
```
dpallotti, line 838: This is not about this pull request, but I just realized that maybe we should read the exact timestamp from the record in the ring buffer rather than using `MonotonicTimestampNs` here.

dpallotti (approved):

OrbitLinuxTracing/TracerThread.cpp
dimitry-: ops - I kind of missed that. Yes, will do it in follow-up cl.

PR #1533, opened 2020-12-02
Remove unused methods from CaptureData

dimitry-:
Test: builds


florian-kuebler:
Thanks for that change. Regarding the methods about user defined data, I know that Ronald added those just recently. So I'd let him decide.

ronaldfw (approved):
LGTM

PR #1535, opened 2020-12-02
Add metrics uploader client

IrinaShkviro:
Add MetricsUploader subproject to send log events from Orbit. For
communication between Orbit and metrics_uploader_client.dll
OrbitLogEvent proto message is used.

Currently only ORBIT_INITIALIZED and ORBIT_CAPTURE_DURATION
log message are available.

The metrics will be sent only if metrics_uploader_client.dll is available
and --enable_metrics flag is set.

Bug: http://b/170389366

Test: start Orbit, check that the ORBIT_INITIALIZED and
ORBIT_CAPTURE_DURATION log messages appears on the server.

### DIFF
--- a/MetricsUploader/MetricsUploader.cpp
+++ b/MetricsUploader/MetricsUploader.cpp
@@ -7,6 +7,7 @@
 #include <vector>
 
 #include "OrbitBase/Logging.h"
+#include "OrbitVersion/OrbitVersion.h"
 
 namespace orbit_metrics_uploader {
 
@@ -14,6 +15,7 @@ bool MetricsUploader::SendLogEvent(OrbitLogEvent_LogEventType log_event_type) {
   if (send_log_event_addr_ != nullptr) {
     OrbitLogEvent log_event;
     log_event.set_log_event_type(log_event_type);
+    log_event.set_orbit_version(orbit_core::GetVersion());
 
     int message_size = log_event.ByteSize();
     std::vector<uint8_t> buffer(message_size);
--- a/MetricsUploader/MetricsUploader.cpp
+++ b/MetricsUploader/MetricsUploader.cpp
@@ -11,11 +11,13 @@
 
 namespace orbit_metrics_uploader {
 
-bool MetricsUploader::SendLogEvent(OrbitLogEvent_LogEventType log_event_type) {
+bool MetricsUploader::SendLogEvent(OrbitLogEvent_LogEventType log_event_type,
+                                   std::chrono::milliseconds event_duration) {
   if (send_log_event_addr_ != nullptr) {
     OrbitLogEvent log_event;
     log_event.set_log_event_type(log_event_type);
     log_event.set_orbit_version(orbit_core::GetVersion());
+    log_event.set_event_duration_milliseconds(event_duration.count());
 
     int message_size = log_event.ByteSize();
     std::vector<uint8_t> buffer(message_size);
--- a/MetricsUploader/MetricsUploader.cpp
+++ b/MetricsUploader/MetricsUploader.cpp
@@ -1,0 +1,31 @@
+// Copyright (c) 2020 The Orbit Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#include "MetricsUploader/MetricsUploader.h"
+
+#include <vector>
+
+#include "OrbitBase/Logging.h"
+
+namespace orbit_metrics_uploader {
+
+bool MetricsUploader::SendLogEvent(OrbitLogEvent_LogEventType log_event_type) {
+  if (send_log_event_addr_ != nullptr) {
+    OrbitLogEvent log_event;
+    log_event.set_log_event_type(log_event_type);
+
+    int message_size = log_event.ByteSize();
+    std::vector<uint8_t> buffer(message_size);
+    log_event.SerializeToArray(buffer.data(), message_size);
+
+    Result result = send_log_event_addr_(buffer.data(), message_size);
+    if (result == kNoError) {
+      return true;
+    }
+    ERROR("Can't start the metrics uploader client: %s", GetErrorMessage(result));
+  }
+  return false;
+}
+
+}  // namespace orbit_metrics_uploader
--- a/MetricsUploader/MetricsUploaderLinux.cpp
+++ b/MetricsUploader/MetricsUploaderLinux.cpp
@@ -1,0 +1,18 @@
+// Copyright (c) 2020 The Orbit Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#include "MetricsUploader/MetricsUploader.h"
+
+namespace orbit_metrics_uploader {
+
+MetricsUploader::MetricsUploader() = default;
+MetricsUploader::~MetricsUploader() = default;
+MetricsUploader::MetricsUploader(MetricsUploader&& other) = default;
+MetricsUploader& MetricsUploader::operator=(MetricsUploader&& other) = default;
+
+ErrorMessageOr<MetricsUploader> MetricsUploader::CreateMetricsUploader(std::string) {
+  return ErrorMessage("MetricsUploader is not implemented on Linux");
+}
+
+}  // namespace orbit_metrics_uploader
--- a/MetricsUploader/MetricsUploaderWindows.cpp
+++ b/MetricsUploader/MetricsUploaderWindows.cpp
@@ -1,0 +1,100 @@
+// Copyright (c) 2020 The Orbit Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#include <absl/memory/memory.h>
+
+#include "MetricsUploader/MetricsUploader.h"
+#include "OrbitBase/Logging.h"
+
+namespace orbit_metrics_uploader {
+
+constexpr const char* kSendLogEventFunctionName = "SendOrbitLogEvent";
+constexpr const char* kStartUploaderClientFunctionName = "StartUploaderClient";
+
+MetricsUploader::MetricsUploader(std::string client_name,
+                                 Result (*send_log_event_addr)(const uint8_t*, int),
+                                 HMODULE metrics_uploader_client_dll)
+    : client_name_(std::move(client_name)),
+      send_log_event_addr_(send_log_event_addr),
+      metrics_uploader_client_dll_(metrics_uploader_client_dll) {}
+
+MetricsUploader::MetricsUploader(MetricsUploader&& other)
+    : client_name_(std::move(other.client_name_)),
+      metrics_uploader_client_dll_(other.metrics_uploader_client_dll_),
+      send_log_event_addr_(other.send_log_event_addr_) {
+  other.metrics_uploader_client_dll_ = nullptr;
+  other.send_log_event_addr_ = nullptr;
+}
+
+MetricsUploader& MetricsUploader::operator=(MetricsUploader&& other) {
+  if (&other == this) {
+    return *this;
+  }
+
+  client_name_ = std::move(other.client_name_);
+
+  metrics_uploader_client_dll_ = other.metrics_uploader_client_dll_;
+  other.metrics_uploader_client_dll_ = nullptr;
+
+  send_log_event_addr_ = other.send_log_event_addr_;
+  other.send_log_event_addr_ = nullptr;
+
+  return *this;
+}
+
+MetricsUploader::~MetricsUploader() {
+  if (nullptr != metrics_uploader_client_dll_) {
+    FreeLibrary(metrics_uploader_client_dll_);
+  }
+}
+
+ErrorMessageOr<MetricsUploader> MetricsUploader::CreateMetricsUploader(std::string client_name) {
+  HMODULE metrics_uploader_client_dll;
+  if (GetModuleHandleExA(GET_MODULE_HANDLE_EX_FLAG_UNCHANGED_REFCOUNT, client_name.c_str(),
+                         &metrics_uploader_client_dll) != 0) {
+    return ErrorMessage("MetricsUploader is already created");
+  }
+
+  metrics_uploader_client_dll = LoadLibraryA(client_name.c_str());
+  if (nullptr != metrics_uploader_client_dll) {
+    LOG("Metrics uploader client library is loaded");
+
+    auto start_uploader_client_addr = reinterpret_cast<Result (*)()>(
+        GetProcAddress(metrics_uploader_client_dll, kStartUploaderClientFunctionName));
+    if (nullptr == start_uploader_client_addr) {
+      std::string error_message =
+          absl::StrFormat("%s function not found", kStartUploaderClientFunctionName);
+      ERROR("%s", error_message);
+      FreeLibrary(metrics_uploader_client_dll);
+      return ErrorMessage(error_message);
+    }
+
+    auto send_log_event_addr = reinterpret_cast<Result (*)(const uint8_t*, int)>(
+        GetProcAddress(metrics_uploader_client_dll, kSendLogEventFunctionName));
+    if (nullptr == send_log_event_addr) {
+      std::string error_message =
+          absl::StrFormat("%s function not found", kSendLogEventFunctionName);
+      ERROR("%s", error_message);
+      FreeLibrary(metrics_uploader_client_dll);
+      return ErrorMessage(error_message);
+    }
+
+    // set up connection and create a client
+    Result result = start_uploader_client_addr();
+    if (result != kNoError) {
+      std::string error_message = absl::StrFormat(
+          "Error while starting the metrics uploader client: %s", GetErrorMessage(result));
+      ERROR("%s", error_message);
+      FreeLibrary(metrics_uploader_client_dll);
+      return ErrorMessage(error_message);
+    }
+
+    return MetricsUploader(client_name, send_log_event_addr, metrics_uploader_client_dll);
+  }
+  std::string error_message = "Metrics uploader client library is not found";
+  ERROR("%s", error_message);
+  return ErrorMessage(error_message);
+}
+
+}  // namespace orbit_metrics_uploader
--- a/MetricsUploader/MetricsUploaderWindowsTest.cpp
+++ b/MetricsUploader/MetricsUploaderWindowsTest.cpp
@@ -1,0 +1,49 @@
+// Copyright (c) 2020 The Orbit Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#include <gtest/gtest.h>
+
+#include "MetricsUploader/MetricsUploader.h"
+
+using orbit_metrics_uploader::MetricsUploader;
+
+TEST(MetricsUploader, CreateMetricsUploaderFromClientWithoutSendEvent) {
+  auto metrics_uploader =
+      MetricsUploader::CreateMetricsUploader("MetricsUploaderClientWithoutSendEvent");
+  EXPECT_EQ(metrics_uploader.has_value(), false);
+}
+
+TEST(MetricsUploader, CreateMetricsUploaderFromClientWithoutStart) {
+  auto metrics_uploader =
+      MetricsUploader::CreateMetricsUploader("MetricsUploaderClientWithoutStart");
+  EXPECT_EQ(metrics_uploader.has_value(), false);
+}
+
+TEST(MetricsUploader, StartMetricsUploaderWithError) {
+  auto metrics_uploader =
+      MetricsUploader::CreateMetricsUploader("MetricsUploaderStartWithErrorClient");
+  EXPECT_EQ(metrics_uploader.has_value(), false);
+}
+
+TEST(MetricsUploader, SendLogEvent) {
+  auto metrics_uploader = MetricsUploader::CreateMetricsUploader("MetricsUploaderCompleteClient");
+  EXPECT_EQ(metrics_uploader.has_value(), true);
+  bool result = metrics_uploader.value().SendLogEvent(
+      orbit_metrics_uploader::OrbitLogEvent_LogEventType_UNKNOWN_EVENT_TYPE);
+  EXPECT_EQ(result, false);
+  result = metrics_uploader.value().SendLogEvent(
+      orbit_metrics_uploader::OrbitLogEvent_LogEventType_ORBIT_INITIALIZED);
+  EXPECT_EQ(result, true);
+  result = metrics_uploader.value().SendLogEvent(
+      orbit_metrics_uploader::OrbitLogEvent_LogEventType_ORBIT_CAPTURE_DURATION,
+      std::chrono::milliseconds(100));
+  EXPECT_EQ(result, true);
+}
+
+TEST(MetricsUploader, CreateTwoMetricsUploaders) {
+  auto metrics_uploader1 = MetricsUploader::CreateMetricsUploader("MetricsUploaderCompleteClient");
+  EXPECT_EQ(metrics_uploader1.has_value(), true);
+  auto metrics_uploader2 = MetricsUploader::CreateMetricsUploader("MetricsUploaderCompleteClient");
+  EXPECT_EQ(metrics_uploader2.has_value(), false);
+}
--- a/MetricsUploader/include/MetricsUploader/MetricsUploader.h
+++ b/MetricsUploader/include/MetricsUploader/MetricsUploader.h
@@ -9,6 +9,7 @@
 #include <windows.h>
 #endif  // WIN32
 
+#include <chrono>
 #include <cstdint>
 #include <string>
 
@@ -51,7 +52,8 @@ class MetricsUploader {
 
   // Send log events to the server using metrics_uploader.
   // Returns true on success and false otherwise.
-  bool SendLogEvent(OrbitLogEvent_LogEventType log_event_type);
+  bool SendLogEvent(OrbitLogEvent_LogEventType log_event_type,
+                    std::chrono::milliseconds event_duration = std::chrono::milliseconds::zero());
 
  private:
 #ifdef _WIN32
--- a/MetricsUploader/include/MetricsUploader/MetricsUploader.h
+++ b/MetricsUploader/include/MetricsUploader/MetricsUploader.h
@@ -1,0 +1,72 @@
+// Copyright (c) 2020 The Orbit Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#ifndef ORBIT_METRICS_UPLOADER_METRICS_UPLOADER_H_
+#define ORBIT_METRICS_UPLOADER_METRICS_UPLOADER_H_
+
+#ifdef _WIN32
+#include <windows.h>
+#endif  // WIN32
+
+#include <cstdint>
+#include <string>
+
+#include "MetricsUploader/Result.h"
+#include "OrbitBase/Result.h"
+#include "orbit_log_event.pb.h"
+
+namespace orbit_metrics_uploader {
+
+const std::string kMetricsUploaderClientDllName = "metrics_uploader_client";
+
+// This class is used for sending log events from Orbit. It works only on Windows and only if
+// metrics_uploader_client.dll is available. The types of logs that could be send are defined in
+// orbit_client_protos::OrbitLogEvent::LogEventType.
+//
+// Only one instance of the class is allowed.
+//
+// Usage example:
+//
+// auto metrics_uploader = MetricsUploader::CreateMetricsUploader();
+// if (metrics_uploader.has_value()) {
+//   metrics_uploader.value().SendLogEvent(...);
+// }
+//
+class MetricsUploader {
+ public:
+  // Create a MetricsUploader instance, load metrics uploader client library if available and starts
+  // metrics uploader client when called first time. Return the instance if there are no errors and
+  // ErrorMessage otherwise.
+  [[nodiscard]] static ErrorMessageOr<MetricsUploader> CreateMetricsUploader(
+      std::string client_name = kMetricsUploaderClientDllName);
+
+  // Unload metrics_uploader_client.dll
+  ~MetricsUploader();
+
+  MetricsUploader(const MetricsUploader& other) = delete;
+  MetricsUploader(MetricsUploader&& other);
+  MetricsUploader& operator=(const MetricsUploader& other) = delete;
+  MetricsUploader& operator=(MetricsUploader&& other);
+
+  // Send log events to the server using metrics_uploader.
+  // Returns true on success and false otherwise.
+  bool SendLogEvent(OrbitLogEvent_LogEventType log_event_type);
+
+ private:
+#ifdef _WIN32
+  HMODULE metrics_uploader_client_dll_;
+  explicit MetricsUploader(std::string client_name,
+                           Result (*send_log_event_addr)(const uint8_t*, int),
+                           HMODULE metrics_uploader_client_dll);
+#else
+  MetricsUploader();
+#endif
+
+  Result (*send_log_event_addr_)(const uint8_t*, int) = nullptr;
+  std::string client_name_;
+};
+
+}  // namespace orbit_metrics_uploader
+
+#endif  // ORBIT_METRICS_UPLOADER_METRICS_UPLOADER_H_
--- a/MetricsUploader/proto/orbit_log_event.proto
+++ b/MetricsUploader/proto/orbit_log_event.proto
@@ -13,8 +13,10 @@ message OrbitLogEvent {
   enum LogEventType {
     UNKNOWN_EVENT_TYPE = 0;
     ORBIT_INITIALIZED = 1;
+    ORBIT_CAPTURE_DURATION = 2;
   }
   LogEventType log_event_type = 1;
 
   string orbit_version = 2;
+  int64 event_duration_milliseconds = 3;
 }
--- a/MetricsUploader/proto/orbit_log_event.proto
+++ b/MetricsUploader/proto/orbit_log_event.proto
@@ -15,4 +15,6 @@ message OrbitLogEvent {
     ORBIT_INITIALIZED = 1;
   }
   LogEventType log_event_type = 1;
+
+  string orbit_version = 2;
 }
--- a/MetricsUploader/proto/orbit_log_event.proto
+++ b/MetricsUploader/proto/orbit_log_event.proto
@@ -1,0 +1,18 @@
+// Copyright (c) 2020 The Orbit Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+syntax = "proto3";
+
+package orbit_metrics_uploader;
+
+// This proto is used for communication between Orbit and metrics uploader
+// client. Changing this proto requires changing the metrics uploader client
+// library accordingly.
+message OrbitLogEvent {
+  enum LogEventType {
+    UNKNOWN_EVENT_TYPE = 0;
+    ORBIT_INITIALIZED = 1;
+  }
+  LogEventType log_event_type = 1;
+}
--- a/OrbitGl/App.cpp
+++ b/OrbitGl/App.cpp
@@ -69,6 +69,7 @@
 #include "TrackManager.h"
 #include "capture_data.pb.h"
 #include "module.pb.h"
+#include "orbit_log_event.pb.h"
 #include "preset.pb.h"
 #include "process.pb.h"
 #include "symbol.pb.h"
@@ -129,8 +130,9 @@ PresetLoadState GetPresetLoadStateForProcess(
 
 bool DoZoom = false;
 
-OrbitApp::OrbitApp(MainThreadExecutor* main_thread_executor)
-    : main_thread_executor_(main_thread_executor) {
+OrbitApp::OrbitApp(MainThreadExecutor* main_thread_executor,
+                   orbit_metrics_uploader::MetricsUploader* metrics_uploader)
+    : main_thread_executor_(main_thread_executor), metrics_uploader_(metrics_uploader) {
   thread_pool_ = ThreadPool::Create(4 /*min_size*/, 256 /*max_size*/, absl::Seconds(1));
   main_thread_id_ = std::this_thread::get_id();
   data_manager_ = std::make_unique<DataManager>(main_thread_id_);
@@ -312,8 +314,10 @@ void OrbitApp::OnValidateFramePointers(std::vector<const ModuleData*> modules_to
   });
 }
 
-std::unique_ptr<OrbitApp> OrbitApp::Create(MainThreadExecutor* main_thread_executor) {
-  auto app = std::make_unique<OrbitApp>(main_thread_executor);
+std::unique_ptr<OrbitApp> OrbitApp::Create(
+    MainThreadExecutor* main_thread_executor,
+    orbit_metrics_uploader::MetricsUploader* metrics_uploader) {
+  auto app = std::make_unique<OrbitApp>(main_thread_executor, metrics_uploader);
 
 #ifdef _WIN32
   oqpi::default_helpers::start_default_scheduler();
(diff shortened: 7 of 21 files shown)


MetricsUploader/include/MetricsUploader/MetricsUploader.h
```diff
@@ -0,0 +1,42 @@
...
+  MetricsUploader();
+
+  // Unload metrics_uploader_client.dll
+  ~MetricsUploader();
+
+  // Send log events to the server using metrics_uploader.
+  void SendLogEvent(orbit_client_protos::OrbitLogEvent_LogEventType log_event_type);
+
+ private:
+  inline static bool is_init_{false};
```
beckerhe, line 35: We should make that class thread-safe by default which means, `is_init_` should have a mutex.

MetricsUploader/MetricsUploaderWindows.cpp
```diff
@@ -0,0 +1,68 @@
...
+
+using orbit_client_protos::OrbitLogEvent;
+using orbit_client_protos::OrbitLogEvent_LogEventType;
+
+constexpr const wchar_t* kMetricsUploaderClientDllName = L"metrics_uploader_client.dll";
+constexpr const char* kSendLogEventFunctionName = "SendOrbitLogEvent";
+constexpr const char* kStartUploaderClientFunctionName = "StartUploaderClient";
+
+MetricsUploader::MetricsUploader() {
+  CHECK(!is_init_);
```
beckerhe, line 19: I think this can be made more explicit for the user of this class.

You can make the constructor private and add a factory function which returns a `ErrorMessageOr<MetricsUploader>`. So everybody has to go through the factory function and check if they actually got a MetricsUploader object.

A singleton might be another solution.

OrbitClientProtos/orbit_log_event.proto
```diff
@@ -0,0 +1,15 @@
...
+
+package orbit_client_protos;
+
+message OrbitLogEvent {
+  enum LogEventType {
+    UNKNOWN_EVENT_TYPE = 0;
+    ORBIT_INITIALIZED = 1;
+  }
+  LogEventType log_event_type = 1;
+}
```
beckerhe: That's the proto which have to be kept in sync with the library, isn't it? Could you add a comment explaining that to avoid folks accidentally changing that without making the some change in the library?

MetricsUploader/include/MetricsUploader/MetricsUploader.h
```diff
@@ -0,0 +1,42 @@
+// Copyright (c) 2020 The Orbit Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#ifndef ORBIT_METRICS_UPLOADER_METRICS_UPLOADER_H_
+#define ORBIT_METRICS_UPLOADER_METRICS_UPLOADER_H_
+
+#ifdef _WIN32
+#include <windows.h>
```
beckerhe, line 9: It's only about HINSTANCE, isn't it? So `#include <WinDef.h>` should be enough.

MetricsUploader/include/MetricsUploader/MetricsUploader.h
```diff
@@ -0,0 +1,42 @@
...
+  // Unload metrics_uploader_client.dll
+  ~MetricsUploader();
+
+  // Send log events to the server using metrics_uploader.
+  void SendLogEvent(orbit_client_protos::OrbitLogEvent_LogEventType log_event_type);
+
+ private:
+  inline static bool is_init_{false};
+#ifdef _WIN32
+  HINSTANCE metrics_uploader_client_dll_;
```
beckerhe, line 37: Nit: AFAIK the return type of LoadLibrary is HMODULE (which happens to be the same as HINSTANCE - but ideally we don't assume that.)

OrbitClientProtos/CMakeLists.txt
```diff
@@ -8,6 +8,7 @@ project(OrbitClientProtos)
 
 add_library(OrbitClientProtos STATIC)
 protobuf_generate(TARGET OrbitClientProtos PROTOS capture_data.proto
+                                                  orbit_log_event.proto
```
beckerhe, line 11: I think that proto should live in `MetricsUploader` instead. It's tied to the class in there, so there is no use case without the class.

MetricsUploader/include/MetricsUploader/MetricsUploader.h
```diff
@@ -0,0 +1,42 @@
...
+  void (*send_log_event_addr_)(const uint8_t*, int);
```
beckerhe, line 38: Can you initialize this to `nullptr`?

MetricsUploader/MetricsUploaderWindows.cpp
```diff
@@ -0,0 +1,68 @@
...
+}
+
+MetricsUploader::~MetricsUploader() {
+  if (is_init_) {
+    FreeLibrary(metrics_uploader_client_dll_);
+  }
+}
+
+void MetricsUploader::SendLogEvent(OrbitLogEvent_LogEventType log_event_type) {
+  if (is_init_) {
```
beckerhe, line 61: Nit: `if (send_log_event_addr_ != nullptr)` instead

That's basically the same condition but easier to understand for someone just looking at this function.

MetricsUploader/MetricsUploaderWindows.cpp
```diff
@@ -0,0 +1,68 @@
...
+    start_uploader_client_addr_();
+    is_init_ = true;
+  } else {
+    LOG("Metrcs uploader client library is not found");
+  }
+}
+
+MetricsUploader::~MetricsUploader() {
+  if (is_init_) {
+    FreeLibrary(metrics_uploader_client_dll_);
```
beckerhe, line 51: I guess you should reset is_init_ here.

beckerhe (commented):
Hi Irina,
sorry for being so late. I totally missed when that moved from Draft to Ready.

I added some comments, but overall looks good.

But let's quickly talk about tests. :innocent: 

I think that code needs unit tests. And I gave some thought on how to do that.

1. I wouldn't make the library file name fixed. In that case you can load a different DLL in a test.
2. I would add a DLL which implements the same functions as the uploader library. Then you can load that DLL in a test case and see if it behaves correctly.

MetricsUploader/include/MetricsUploader/MetricsUploader.h
```diff
@@ -0,0 +1,44 @@
...
+
+#include "orbit_log_event.pb.h"
+
+namespace orbit_metrics_uploader {
+
+/**
+ * This class is used for sending log events from Orbit. It works only on Windows and only if
+ * metrics_uploader_client.dll is available. The types of logs that could be send are defined in
+ * orbit_client_protos::OrbitLogEvent::LogEventType.
+ */
```
beckerhe, line 20: The text also needs to describe the pre-conditions. In this case: The class can only be instantiated once. But like I said in other comments, ideally we make that clear on the API level and don't rely on comments or CHECKS.

MetricsUploader/MetricsUploaderLinux.cpp
```diff
@@ -0,0 +1,20 @@
...
+
+#include "MetricsUploader/MetricsUploader.h"
+
+namespace orbit_metrics_uploader {
+
+using orbit_client_protos::OrbitLogEvent_LogEventType;
+
+MetricsUploader::MetricsUploader() {}
+
+MetricsUploader::~MetricsUploader() {}
```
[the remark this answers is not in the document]
IrinaShkviro: Done, thanks

MetricsUploader/MetricsUploaderWindows.cpp
```diff
@@ -0,0 +1,76 @@
...
+void MetricsUploader::SendLogEvent(OrbitLogEvent_LogEventType log_event_type,
+                                   int64_t event_duration_ms) {
+  if (is_init_) {
+    OrbitLogEvent log_event;
+    log_event.set_log_event_type(log_event_type);
+    log_event.set_orbit_version(OrbitCore::GetVersion());
+    log_event.set_event_duration_milliseconds(event_duration_ms);
+
+    int message_size = log_event.ByteSize();
+    uint8_t* buffer = new uint8_t[message_size];
```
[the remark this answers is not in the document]
IrinaShkviro: I've added std::vector for the buffer, thanks.

MetricsUploader/MetricsUploaderWindows.cpp
```diff
@@ -0,0 +1,76 @@
...
+    log_event.SerializeToArray(buffer, message_size);
+
+    send_log_event_addr_(buffer, message_size);
+    delete[] buffer;
+  }
+}
```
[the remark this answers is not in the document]
IrinaShkviro: Right, I've moved it to the MetricsUploader.cpp.

MetricsUploader/include/MetricsUploader/MetricsUploader.h
IrinaShkviro: Thanks, I've added Mutex for `is_init_`.

MetricsUploader/MetricsUploaderWindows.cpp
IrinaShkviro: Added `static CreateMetricsUploader()` function and made constructor private.

OrbitClientProtos/orbit_log_event.proto
IrinaShkviro: Yes, right, added a comment about it, thank you!

MetricsUploader/include/MetricsUploader/MetricsUploader.h
IrinaShkviro: Including only `<WinDef.h>` cause `fatal error C1189: #error:  "No Target Architecture"` for me.

MetricsUploader/include/MetricsUploader/MetricsUploader.h
IrinaShkviro: Right, fixed it, than you!

OrbitClientProtos/CMakeLists.txt
IrinaShkviro: I've moved it to the `MetricsUploader/proto` folder.

MetricsUploader/include/MetricsUploader/MetricsUploader.h
IrinaShkviro: Done, thanks.

MetricsUploader/MetricsUploaderWindows.cpp
IrinaShkviro: Changed the condition to `if (send_log_event_addr_ != nullptr)`.

MetricsUploader/MetricsUploaderWindows.cpp
IrinaShkviro: Added `Reset()` function for this, thank you!

MetricsUploader/include/MetricsUploader/MetricsUploader.h
IrinaShkviro: I've changed the description, ptal

IrinaShkviro:
> Hi Irina,
> sorry for being so late. I totally missed when that moved from Draft to Ready.
> 
> I added some comments, but overall looks good.
> 
> But let's quickly talk about tests. 😇
> 
> I think that code needs unit tests. And I gave some thought on how to do that.
> 
> 1. I wouldn't make the library file name fixed. In that case you can load a different DLL in a test.
> 2. I would add a DLL which implements the same functions as the uploader library. Then you can load that DLL in a test case and see if it behaves correctly.

Thanks a lot for the review Henning! I've added several tests and also addressed most of the comments. Could you, please take another look?

MetricsUploader/proto/orbit_log_event.proto
```diff
@@ -0,0 +1,22 @@
...
+message OrbitLogEvent {
+  enum LogEventType {
+    UNKNOWN_EVENT_TYPE = 0;
+    ORBIT_INITIALIZED = 1;
+    ORBIT_CAPTURE_DURATION = 2;
+  }
+  LogEventType log_event_type = 1;
+
+  string orbit_version = 2;
+  int64 event_duration_milliseconds = 3;
```
ronaldfw, line 21: Do all events necessarily have a duration? And maybe we want a timestamp of when an event occurred? That would allow computations of durations after the fact. E.g. types ORBIT_CAPTURE_STARTED and ORBIT_CAPTURE_FINISHED with a timestamp each. Or does that cause other problems in the analysis of the data?

MetricsUploader/MetricsUploader.cpp
```diff
@@ -0,0 +1,69 @@
...
+  }
+}
+
+ErrorMessageOr<std::unique_ptr<MetricsUploader>> MetricsUploader::CreateMetricsUploader(
+    const std::string_view client_name) {
+  absl::MutexLock lock(&init_mutex_);
+  if (is_init_) {
+    return ErrorMessage("MetricsUploader is already created");
+  }
+  MetricsUploader* metrics_uploader = new MetricsUploader(client_name);
```
beckerhe, line 44: You could use std::make_unique here instead of `new`.

ronaldfw (commented):
I don't have much context about this at this point, but would like to discuss the log event definitions. Not sure if my comments make sense, but would love to understand our options better.

MetricsUploader/include/MetricsUploader/MetricsUploader.h
```diff
@@ -0,0 +1,76 @@
...
+  MetricsUploader& operator=(const MetricsUploader& other) = delete;
+  MetricsUploader& operator=(MetricsUploader&& other) = default;
+
+  // Send log events to the server using metrics_uploader.
+  // Returns true on success and false otherwise.
+  bool SendLogEvent(OrbitLogEvent_LogEventType log_event_type);
+
+ private:
+  // Load metrics uploader client library if available and starts metrics uploader client.
+  MetricsUploader(const std::string_view client_name = kMetricsUploaderClientDllName);
```
beckerhe, line 60: `explicit`

MetricsUploader/include/MetricsUploader/MetricsUploader.h
```diff
@@ -0,0 +1,76 @@
...
+
+  void Reset();
```
beckerhe, line 62: If I understand everything correctly: The user should call `CreateMetricsUploader` and will only get a result if no instance has been created before. So that's good.

If no instance has been created yet, the CreateMetricsUploader function will create a new `MetricsUploader` instance. But the construction of that instance can fail which is checked, by `is_init_`.

What about moving all the code from the constructor to `CreateMetricsUploader`? This would simplify the design since `Reset won't be needed anymore and `is_init` will only be changed in one place - in `CreateMetricsUploader`

MetricsUploader/include/MetricsUploader/MetricsUploader.h
```diff
@@ -0,0 +1,76 @@
...
+  std::string GetErrorMessage(Result error);
+
+  inline static bool is_init_ = false;
+  static absl::Mutex init_mutex_;
```
beckerhe, line 66: Personal preference: If `is_init_` was inline, I would also make `init_mutex_` inline. Then you don't need the definition in the cpp file.

OrbitGl/App.cpp
```diff
@@ -126,8 +127,9 @@ PresetLoadState GetPresetLoadStateForProcess(
 
 bool DoZoom = false;
 
-OrbitApp::OrbitApp(MainThreadExecutor* main_thread_executor)
-    : main_thread_executor_(main_thread_executor) {
+OrbitApp::OrbitApp(MainThreadExecutor* main_thread_executor,
+                   orbit_metrics_uploader::MetricsUploader* metrics_uploader)
+    : main_thread_executor_(main_thread_executor), metrics_uploader_(metrics_uploader) {
```
beckerhe, line 132: Could you put all the integration changes in OrbitGl and OrbitQt into a separate commit?

MetricsUploader/proto/orbit_log_event.proto
```diff
@@ -0,0 +1,22 @@
...
+package orbit_metrics_uploader;
+
+// This proto is used for communication between Orbit and metrics uploader
+// client. Changing this proto requires changing the metrics uploader client
+// library accordingly.
+message OrbitLogEvent {
+  enum LogEventType {
+    UNKNOWN_EVENT_TYPE = 0;
+    ORBIT_INITIALIZED = 1;
+    ORBIT_CAPTURE_DURATION = 2;
```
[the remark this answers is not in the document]
IrinaShkviro: Here I've used the similar to others teams approach (to have a separate log type for every event). Having a comment describing an event could be another option, but what's advantage of it?

MetricsUploader/include/MetricsUploader/MetricsUploader.h
```diff
@@ -53,7 +53,7 @@ class MetricsUploader {
 
   // Send log events to the server using metrics_uploader.
   // Returns true on success and false otherwise.
-  bool SendLogEvent(OrbitLogEvent_LogEventType log_event_type);
+  bool SendLogEvent(OrbitLogEvent_LogEventType log_event_type, int64_t event_duration_ms = 0);
```
beckerhe, line 56: What about using `std::chrono::milliseconds` instead of  `int64_t`?

OrbitGl/App.cpp
```diff
@@ -871,6 +871,17 @@ void OrbitApp::StopCapture() {
     return;
   }
 
+  if (metrics_uploader_ != nullptr) {
+    thread_pool_->Schedule([this] {
```
beckerhe, line 875: What is the reason for doing this asynchronously? Does it take a long time to send a log event?

I would expect a couple of milliseconds max. That shouldn't be a problem for the user since this is just called once.

MetricsUploader/proto/orbit_log_event.proto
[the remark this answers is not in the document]
ronaldfw: It would potentially give more flexibility as you don't have to change the LogEventType definition if you introduce a new event somewhere. But if this is common practice, it seems fine to me.

MetricsUploader/TestClients/SendEventClient.cpp
```diff
@@ -0,0 +1,17 @@
+// Copyright (c) 2020 The Orbit Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#include <cstdint>
+
+#include "MetricsUploader/Result.h"
+
+#define DllExport __declspec(dllexport)
```
beckerhe, line 9: Currently there is no reason for having this macro. It's usually there because you surround it by `#ifdef _WIN32` and `#endif`.

So I'd either add the `ifdef` or remove it.

MetricsUploader/TestClients/StartAndSendEventClient.cpp
```diff
@@ -0,0 +1,30 @@
...
+
+#define DllExport __declspec(dllexport)
+
+using orbit_metrics_uploader::OrbitLogEvent;
+using orbit_metrics_uploader::OrbitLogEvent_LogEventType_UNKNOWN_EVENT_TYPE;
+
+using orbit_metrics_uploader::Result;
+
+extern "C" {
+DllExport enum Result StartUploaderClient() { return Result::kNoError; }
```
beckerhe, line 18: Is `enum Result` correct here? Shouldn't this be `int`? What does the header that is generated from the go-code show? Is there an enum?

MetricsUploader/MetricsUploaderWindowsTest.cpp
```diff
@@ -0,0 +1,54 @@
...
+  EXPECT_EQ(result, false);
+  result = metrics_uploader.value()->SendLogEvent(
+      orbit_metrics_uploader::OrbitLogEvent_LogEventType_ORBIT_INITIALIZED);
+  EXPECT_EQ(result, true);
+  result = metrics_uploader.value()->SendLogEvent(
+      orbit_metrics_uploader::OrbitLogEvent_LogEventType_ORBIT_CAPTURE_DURATION, 100);
+  EXPECT_EQ(result, true);
+}
+
+TEST(MetricsUploader, SecondMetricsUploader) {
```
beckerhe, line 47: What do you mean by `SecondMetricsUploader`?

beckerhe (commented):
Very cool Irina. This looks really good now. I added some questions and comments.

MetricsUploader/proto/orbit_log_event.proto
IrinaShkviro: Not all the events have duration, this field is optional. There were 2 options that I've considered: to send event with duration or to send 2 events (start and end) and use sql query to calculate the duration. From the metrics uploader client perspective they look pretty similar, but I guess that having duration field would make queries easier (for example, when we want to get mean/max capture length for the specific time period). If we want to use ORBIT_CAPTURE_STARTED and ORBIT_CAPTURE_FINISHED only to calculate the duration I think it would be better just to send duration event, or am I missed something?

MetricsUploader/MetricsUploader.cpp
IrinaShkviro: I'm not sure I can do it due to private constructor: https://abseil.io/tips/134

beckerhe: Ah, yes you're right. In this case I'd would recommend to move absl::WrapUnique to line 43. This way you still have a unique_ptr the whole.

BTW, you can also write `const auto metrics_uploader = std::unique_ptr{new MetricsUploader(client_name)};`. AFAIK `WrapUnique` also just calls the constructor of unique_ptr.

IrinaShkviro: Done

MetricsUploader/include/MetricsUploader/MetricsUploader.h
IrinaShkviro: fixed, thanks

MetricsUploader/include/MetricsUploader/MetricsUploader.h
IrinaShkviro: I reorganized this part of the code, could you please take another look?

MetricsUploader/include/MetricsUploader/MetricsUploader.h
IrinaShkviro: I've changed this code: now I have a non-static mutex which id used for send_log_event, and check if the library was loaded earlier with GetModuleHandleExA function.

OrbitGl/App.cpp
IrinaShkviro: Done

MetricsUploader/include/MetricsUploader/MetricsUploader.h
IrinaShkviro: Changed to `std::chrono::milliseconds`, thanks!

OrbitGl/App.cpp
IrinaShkviro: Remove async call for now, will modify the MetricsUploader itself to make async calls if needed in future

MetricsUploader/TestClients/SendEventClient.cpp
IrinaShkviro: Removed `#define ` here

MetricsUploader/TestClients/StartAndSendEventClient.cpp
IrinaShkviro: I put the same enum in the preamble, so generated header have the `enum Result` declaration. The signature of the functions are:
```
extern enum Result StartUploaderClient();
extern enum Result SendOrbitLogEvent(uint8_t* serializedProto, int length);
```

MetricsUploader/MetricsUploaderWindowsTest.cpp
IrinaShkviro: In this test I want to check that it's not possible to create MetricsUploader for the second time. Renamed this test

MetricsUploader/MetricsUploaderWindows.cpp
```diff
@@ -0,0 +1,75 @@
...
+constexpr const char* kStartUploaderClientFunctionName = "StartUploaderClient";
+
+MetricsUploader::~MetricsUploader() {
+  absl::MutexLock lock(&send_log_event_mutex_);
+  send_log_event_addr_ = nullptr;
+  HMODULE metrics_uploader_client_dll;
+  if (GetModuleHandleExA(GET_MODULE_HANDLE_EX_FLAG_UNCHANGED_REFCOUNT, client_name_.c_str(),
+                         &metrics_uploader_client_dll) != 0) {
+    FreeLibrary(metrics_uploader_client_dll);
+  }
```
beckerhe, line 23: Could you make `metrics_uploader_client_dll` a member variable of `MetricsUploader` and then just call `FreeLibrary`?

MetricsUploader/MetricsUploaderWindows.cpp
```diff
@@ -0,0 +1,75 @@
...
+
+namespace orbit_metrics_uploader {
+
+constexpr const char* kSendLogEventFunctionName = "SendOrbitLogEvent";
+constexpr const char* kShutdownUploaderClientConnName = "ShutdownUploaderClientConn";
+constexpr const char* kStartUploaderClientFunctionName = "StartUploaderClient";
+
+MetricsUploader::~MetricsUploader() {
+  absl::MutexLock lock(&send_log_event_mutex_);
+  send_log_event_addr_ = nullptr;
```
beckerhe, line 18: Setting this to `nullptr` is not necessary, since the object gets destructed anyway.

MetricsUploader/MetricsUploaderWindows.cpp
```diff
@@ -0,0 +1,75 @@
...
+    if (NULL == start_uploader_client_addr_) {
+      std::string error_message =
+          absl::StrFormat("%s function not found", kStartUploaderClientFunctionName);
+      ERROR("%s", error_message);
+      FreeLibrary(metrics_uploader_client_dll);
+      return ErrorMessage(error_message);
+    }
+
+    auto metrics_uploader = std::unique_ptr<MetricsUploader>{new MetricsUploader(client_name)};
+    metrics_uploader->send_log_event_addr_ = (Result(*)(const uint8_t*, int))GetProcAddress(
```
beckerhe, line 50: Instead of assigning `send_log_event_addr_`, could you make this a constructor argument?

MetricsUploader/include/MetricsUploader/MetricsUploader.h
```diff
@@ -0,0 +1,71 @@
...
+  MetricsUploader(MetricsUploader&& other) = delete;
+  MetricsUploader& operator=(const MetricsUploader& other) = delete;
+  MetricsUploader& operator=(MetricsUploader&& other) = delete;
+
+  // Send log events to the server using metrics_uploader.
+  // Returns true on success and false otherwise.
+  bool SendLogEvent(OrbitLogEvent_LogEventType log_event_type);
+
+ private:
+  explicit MetricsUploader(std::string_view client_name);
```
beckerhe, line 63: Since `client_name` is a sink variable, you should use `std::string` here and move it into place.

MetricsUploader/include/MetricsUploader/MetricsUploader.h
```diff
@@ -0,0 +1,71 @@
...
+  // metrics uploader client when called first time. Return the instance if there are no errors and
+  // ErrorMessage otherwise.
+  [[nodiscard]] static ErrorMessageOr<std::unique_ptr<MetricsUploader>> CreateMetricsUploader(
+      std::string_view client_name = kMetricsUploaderClientDllName);
+
+  // Unload metrics_uploader_client.dll
+  ~MetricsUploader();
+
+  MetricsUploader(const MetricsUploader& other) = delete;
+  MetricsUploader(MetricsUploader&& other) = delete;
```
beckerhe, line 53: With the new way of handling `FreeLibrary` a move constructor/assignment operator can be easily defined:

```
MetricsUploader(MetricsUploader&& other) : 
  send_log_event_addr_{other.send_log_event_addr_},
  client_name{std::move(client_name)}
{
  other.send_log_event_addr_ = nullptr;
}
```

The move assignment-operator is roughly the same, but needs to check for self assignment:
```
MetricsUploader& MetricsUploader::operator= (MetricsUploader&& other) {
  if(&other == this) return *this;

  send_log_event_addr_ = other.send_log_event_addr_;
  other.send_log_event_addr_ = nullptr;

  client_name = std::move(other.client_name);
  return *this;
}
```

And the destructor may only call `FreeLibrary` when `send_log_event_addr_` is `nullptr`:

```
MetricsUploader::~MetricsUploader() {
  if (send_log_event_addr_ == nullptr)
    return;

  // destructor work....
}
```

MetricsUploader/include/MetricsUploader/MetricsUploader.h
```diff
@@ -0,0 +1,71 @@
...
+//   metrics_uploader.value()->SendLogEvent(...);
+// }
+//
+class MetricsUploader {
+ public:
+  // Create a MetricsUploader instance, load metrics uploader client library if available and starts
+  // metrics uploader client when called first time. Return the instance if there are no errors and
+  // ErrorMessage otherwise.
+  [[nodiscard]] static ErrorMessageOr<std::unique_ptr<MetricsUploader>> CreateMetricsUploader(
+      std::string_view client_name = kMetricsUploaderClientDllName);
```
beckerhe, line 47: When a move constructor is defined (see instructions below) you could return a `ErrorMessageOr<MetricsUploader>` and avoid the heap allocation. To me this is a better design, but I don't feel particularly strong about this.

MetricsUploader/MetricsUploaderWindows.cpp
```diff
@@ -0,0 +1,75 @@
...
+#include "OrbitBase/Logging.h"
+
+namespace orbit_metrics_uploader {
+
+constexpr const char* kSendLogEventFunctionName = "SendOrbitLogEvent";
+constexpr const char* kShutdownUploaderClientConnName = "ShutdownUploaderClientConn";
+constexpr const char* kStartUploaderClientFunctionName = "StartUploaderClient";
+
+MetricsUploader::~MetricsUploader() {
+  absl::MutexLock lock(&send_log_event_mutex_);
```
beckerhe, line 17: You can remove this lock. A destructor can't be thread-safe anyway, since it can only be called once.

MetricsUploader/MetricsUploaderWindows.cpp
beckerhe: `HMODULE metrics_uploader_client_dll` also needs to be a constructor argument (see comment above).

MetricsUploader/include/MetricsUploader/MetricsUploader.h
beckerhe: You would also need to move `metrics_uploader_client_dll_` (which I suggested to be a member variable in a comment above)

MetricsUploader/include/MetricsUploader/MetricsUploader.h
```diff
@@ -0,0 +1,71 @@
...
+  // Returns true on success and false otherwise.
+  bool SendLogEvent(OrbitLogEvent_LogEventType log_event_type);
+
+ private:
+  explicit MetricsUploader(std::string_view client_name);
+
+  absl::Mutex send_log_event_mutex_;
+  Result (*send_log_event_addr_)(const uint8_t*, int) = nullptr;
+
+  const std::string client_name_;
```
beckerhe, line 68: When you define move constructors (like suggested above) this should be non-const. Otherwise you can't move the string to another instance.

MetricsUploader/MetricsUploaderWindows.cpp
```diff
@@ -0,0 +1,75 @@
...
+    if (result != kNoError) {
+      std::string error_message = absl::StrFormat(
+          "Error while starting the metrics uploader client: %s", GetErrorMessage(result));
+      ERROR("%s", error_message);
+      return ErrorMessage(error_message);
+    }
+
+    return metrics_uploader;
+  }
+  std::string error_message = "Metrcs uploader client library is not found";
```
beckerhe, line 70: Nit: Metrics

MetricsUploader/MetricsUploaderWindows.cpp
```diff
@@ -0,0 +1,75 @@
...
+    LOG("Metrcs uploader client library is loaded");
+
+    Result (*start_uploader_client_addr_)();
+    start_uploader_client_addr_ =
+        (Result(*)())GetProcAddress(metrics_uploader_client_dll, kStartUploaderClientFunctionName);
+    if (NULL == start_uploader_client_addr_) {
+      std::string error_message =
+          absl::StrFormat("%s function not found", kStartUploaderClientFunctionName);
+      ERROR("%s", error_message);
+      FreeLibrary(metrics_uploader_client_dll);
```
beckerhe, line 45: You probably should do this in the other error branches below as well, don't you?

MetricsUploader/MetricsUploaderWindows.cpp
```diff
@@ -0,0 +1,75 @@
...
+    return ErrorMessage("MetricsUploader is already created");
+  }
+
+  metrics_uploader_client_dll = LoadLibraryA(client_name.data());
+  if (NULL != metrics_uploader_client_dll) {
+    LOG("Metrcs uploader client library is loaded");
+
+    Result (*start_uploader_client_addr_)();
+    start_uploader_client_addr_ =
+        (Result(*)())GetProcAddress(metrics_uploader_client_dll, kStartUploaderClientFunctionName);
```
beckerhe, line 40: Can you replace the c-style cast by static_cast?

MetricsUploader/MetricsUploaderWindows.cpp
```diff
@@ -0,0 +1,75 @@
...
+                         &metrics_uploader_client_dll) != 0) {
+    return ErrorMessage("MetricsUploader is already created");
+  }
+
+  metrics_uploader_client_dll = LoadLibraryA(client_name.data());
+  if (NULL != metrics_uploader_client_dll) {
+    LOG("Metrcs uploader client library is loaded");
+
+    Result (*start_uploader_client_addr_)();
+    start_uploader_client_addr_ =
```
beckerhe, line 39: When using `auto` you could avoid declaring the variable first

MetricsUploader/MetricsUploaderWindows.cpp
```diff
@@ -0,0 +1,75 @@
...
+
+ErrorMessageOr<std::unique_ptr<MetricsUploader>> MetricsUploader::CreateMetricsUploader(
+    std::string_view client_name) {
+  HMODULE metrics_uploader_client_dll;
+  if (GetModuleHandleExA(GET_MODULE_HANDLE_EX_FLAG_UNCHANGED_REFCOUNT, client_name.data(),
+                         &metrics_uploader_client_dll) != 0) {
+    return ErrorMessage("MetricsUploader is already created");
+  }
+
+  metrics_uploader_client_dll = LoadLibraryA(client_name.data());
```
beckerhe, line 34: You can't use `.data()`. `data()` is not null-terminated! You need to use `.c_str()` which is only available in `std::string`, so you have to replace the parameter type.

MetricsUploader/MetricsUploaderWindows.cpp
```diff
@@ -0,0 +1,75 @@
...
+  if (GetModuleHandleExA(GET_MODULE_HANDLE_EX_FLAG_UNCHANGED_REFCOUNT, client_name_.c_str(),
+                         &metrics_uploader_client_dll) != 0) {
+    FreeLibrary(metrics_uploader_client_dll);
+  }
+}
+
+ErrorMessageOr<std::unique_ptr<MetricsUploader>> MetricsUploader::CreateMetricsUploader(
+    std::string_view client_name) {
+  HMODULE metrics_uploader_client_dll;
+  if (GetModuleHandleExA(GET_MODULE_HANDLE_EX_FLAG_UNCHANGED_REFCOUNT, client_name.data(),
```
beckerhe, line 29: You can't use `.data()`. `data()` is not null-terminated! You need to use `.c_str()` which is only available in `std::string`, so you have to replace the parameter type.

MetricsUploader/MetricsUploaderLinux.cpp
```diff
@@ -0,0 +1,16 @@
...
+// found in the LICENSE file.
+
+#include "MetricsUploader/MetricsUploader.h"
+
+namespace orbit_metrics_uploader {
+
+MetricsUploader::~MetricsUploader() = default;
+
+ErrorMessageOr<std::unique_ptr<MetricsUploader>> MetricsUploader::CreateMetricsUploader([
+    [maybe_unused]] std::string_view client_name) {
```
beckerhe, line 12: The line break looks weird. Seems to be a clang-format problem. I'd suggest to not specify the argument name (client_name). Then the attribute is not needed in the first place.

beckerhe (commented):
Hi Irina,
I added some more comments. I'm sorry I hope you're not already fed up with my comments. :see_no_evil: 

Thank you for the changes!

Henning

MetricsUploader/MetricsUploaderWindows.cpp
IrinaShkviro: Done

MetricsUploader/MetricsUploaderWindows.cpp
IrinaShkviro: Right, thank you

MetricsUploader/MetricsUploaderWindows.cpp
IrinaShkviro: Made both variables constructor arguments (for Windows)

MetricsUploader/include/MetricsUploader/MetricsUploader.h
IrinaShkviro: Done, thanks

MetricsUploader/include/MetricsUploader/MetricsUploader.h
IrinaShkviro: Added move constructor/assignment for the MetricsUploader for Windows

MetricsUploader/include/MetricsUploader/MetricsUploader.h
IrinaShkviro: Agree, removed `unique_ptr` here.

MetricsUploader/MetricsUploaderWindows.cpp
IrinaShkviro: Done, thank you!

MetricsUploader/include/MetricsUploader/MetricsUploader.h
IrinaShkviro: Done

MetricsUploader/MetricsUploaderWindows.cpp
IrinaShkviro: fixed, thanks

MetricsUploader/MetricsUploaderWindows.cpp
IrinaShkviro: I've added `reinterpret_cast` here

MetricsUploader/MetricsUploaderWindows.cpp
IrinaShkviro: Right, thanks

MetricsUploader/MetricsUploaderWindows.cpp
IrinaShkviro: thank you, fixed!

MetricsUploader/MetricsUploaderWindows.cpp
IrinaShkviro: Done

MetricsUploader/MetricsUploaderLinux.cpp
IrinaShkviro: Yes, strange for me as well. Removed argument name and it looks better now

IrinaShkviro:
> Hi Irina,
> I added some more comments. I'm sorry I hope you're not already fed up with my comments. 🙈
> 
> Thank you for the changes!
> 
> Henning

Thanks a lot for the review Henning! I've addressed comments and I think it looks much better now =)

MetricsUploader/include/MetricsUploader/MetricsUploader.h
```diff
@@ -0,0 +1,78 @@
...
+// metrics_uploader_client.dll is available. The types of logs that could be send are defined in
+// orbit_client_protos::OrbitLogEvent::LogEventType.
+//
+// Only one instance of the class is allowed.
+//
+// Usage example:
+//
+// auto metrics_uploader = MetricsUploader::CreateMetricsUploader();
+// if (metrics_uploader.has_value()) {
+//   metrics_uploader.value()->SendLogEvent(...);
```
beckerhe, line 38: Since `CreateMetricsUploader` now returns a value, it's either `metrics_uploader.value().SendLogEvent` or `metrics_uploader->SendLogEvent`.

beckerhe (approved):
Great! LGTM now! Thank you, Irina.

I added a minor comment regarding the example in the comments.

MetricsUploader/include/MetricsUploader/MetricsUploader.h
IrinaShkviro: Thanks a lot! updated the comment

PR #1536, opened 2020-12-02
Include .py and .proto files in license_headers check

IrinaShkviro:
Test: delete license headers in .py and .proto files and run check.sh
script locally; verify that the result of the script is that these files
are missing a license/copyright header.


beckerhe (approved):
LGTM!

PR #1537, opened 2020-12-02
Move Classes from OrbitCore to OrbitClientData and OrbitClientModel

florian-kuebler:
This is another step of getting rid of OrbitCore. It moves all the
data classes to OrbitClientData and basically SamplingProfiler and
CaptureData to OrbitClientModul.
After http://b/166238019 is fixed, we can also move CaptureData to
OrbitClientData, as it will be then not dependent to SamplingProfiler
anymore (but only the resuling data class).

TracepointEventProcessor had a dependency to Threading.h, that should
be not moved. So in order to move that class, this also uses
absl::Mutex here, instead of using the typdefs from Threading.h.

Test: Compile & run tests


dimitry- (approved):
LGTM!

antonrohr (approved):
Nice!

beckerhe:
@florian-kuebler Since you already have two approvals: Is there something I should have a look at in particular? Otherwise I would skip that review. :innocent:

florian-kuebler:
> @florian-kuebler Since you already have two approvals: Is there something I should have a look at in particular? Otherwise I would skip that review. 😇

Feel free to skip that. I'll basically wait for the CI to run and then submit this ASAP.

PR #1538, opened 2020-12-02
Move part of codebase to std::filesystem::path

dimitry-:
This change also simplifies most of the functions, and adds error
handling (in form of FATAL in the case executable_path is not available)

Test: build, run OrbitBaseTests

### DIFF
--- a/ElfUtils/ElfFile.cpp
+++ b/ElfUtils/ElfFile.cpp
@@ -28,7 +28,7 @@ using orbit_grpc_protos::SymbolInfo;
 template <typename ElfT>
 class ElfFileImpl : public ElfFile {
  public:
-  ElfFileImpl(std::string_view file_path,
+  ElfFileImpl(std::filesystem::path file_path,
               llvm::object::OwningBinary<llvm::object::ObjectFile>&& owning_binary);
 
   [[nodiscard]] ErrorMessageOr<ModuleSymbols> LoadSymbols() override;
@@ -37,13 +37,13 @@ class ElfFileImpl : public ElfFile {
   [[nodiscard]] bool HasDebugInfo() const override;
   [[nodiscard]] bool Is64Bit() const override;
   [[nodiscard]] std::string GetBuildId() const override;
-  [[nodiscard]] std::string GetFilePath() const override;
+  [[nodiscard]] std::filesystem::path GetFilePath() const override;
   [[nodiscard]] ErrorMessageOr<LineInfo> GetLineInfo(uint64_t address) override;
 
  private:
   void InitSections();
 
-  const std::string file_path_;
+  const std::filesystem::path file_path_;
   llvm::object::OwningBinary<llvm::object::ObjectFile> owning_binary_;
   llvm::object::ELFObjectFile<ElfT>* object_file_;
   llvm::symbolize::LLVMSymbolizer symbolizer_;
@@ -53,9 +53,9 @@ class ElfFileImpl : public ElfFile {
 };
 
 template <typename ElfT>
-ElfFileImpl<ElfT>::ElfFileImpl(std::string_view file_path,
+ElfFileImpl<ElfT>::ElfFileImpl(std::filesystem::path file_path,
                                llvm::object::OwningBinary<llvm::object::ObjectFile>&& owning_binary)
-    : file_path_(file_path),
+    : file_path_(std::move(file_path)),
       owning_binary_(std::move(owning_binary)),
       has_symtab_section_(false),
       has_debug_info_section_(false) {
@@ -119,7 +119,7 @@ ErrorMessageOr<ModuleSymbols> ElfFileImpl<ElfT>::LoadSymbols() {
 
   ModuleSymbols module_symbols;
   module_symbols.set_load_bias(load_bias);
-  module_symbols.set_symbols_file_path(file_path_);
+  module_symbols.set_symbols_file_path(file_path_.string());
 
   for (const llvm::object::ELFSymbolRef& symbol_ref : object_file_->symbols()) {
     if ((symbol_ref.getFlags() & llvm::object::BasicSymbolRef::SF_Undefined) != 0) {
@@ -130,8 +130,8 @@ ErrorMessageOr<ModuleSymbols> ElfFileImpl<ElfT>::LoadSymbols() {
 
     // Unknown type - skip and generate a warning
     if (!symbol_ref.getType()) {
-      LOG("WARNING: Type is not set for symbol \"%s\" in \"%s\", skipping.", name.c_str(),
-          file_path_.c_str());
+      LOG("WARNING: Type is not set for symbol \"%s\" in \"%s\", skipping.", name,
+          file_path_.string());
       continue;
     }
 
@@ -169,7 +169,7 @@ ErrorMessageOr<uint64_t> ElfFileImpl<ElfT>::GetLoadBias() const {
     std::string error = absl::StrFormat(
         "Unable to get load bias of ELF file: \"%s\". No program headers "
         "found.",
-        file_path_);
+        file_path_.string());
     ERROR("%s", error.c_str());
     return ErrorMessage(std::move(error));
   }
@@ -189,7 +189,7 @@ ErrorMessageOr<uint64_t> ElfFileImpl<ElfT>::GetLoadBias() const {
     std::string error = absl::StrFormat(
         "Unable to get load bias of ELF file: \"%s\". No PT_LOAD program "
         "headers found.",
-        file_path_);
+        file_path_.string());
     ERROR("%s", error.c_str());
     return ErrorMessage(std::move(error));
   }
@@ -212,7 +212,7 @@ std::string ElfFileImpl<ElfT>::GetBuildId() const {
 }
 
 template <typename ElfT>
-std::string ElfFileImpl<ElfT>::GetFilePath() const {
+std::filesystem::path ElfFileImpl<ElfT>::GetFilePath() const {
   return file_path_;
 }
 
@@ -253,15 +253,15 @@ bool ElfFileImpl<llvm::object::ELF32LE>::Is64Bit() const {
 
 }  // namespace
 
-ErrorMessageOr<std::unique_ptr<ElfFile>> ElfFile::CreateFromBuffer(std::string_view file_path,
-                                                                   const void* buf, size_t len) {
+ErrorMessageOr<std::unique_ptr<ElfFile>> ElfFile::CreateFromBuffer(
+    const std::filesystem::path& file_path, const void* buf, size_t len) {
   std::unique_ptr<llvm::MemoryBuffer> buffer = llvm::MemoryBuffer::getMemBuffer(
       llvm::StringRef(static_cast<const char*>(buf), len), llvm::StringRef("buffer name"), false);
   llvm::Expected<std::unique_ptr<llvm::object::ObjectFile>> object_file_or_error =
       llvm::object::ObjectFile::createObjectFile(buffer->getMemBufferRef());
 
   if (!object_file_or_error) {
-    return ErrorMessage(absl::StrFormat("Unable to load ELF file \"%s\": %s", file_path,
+    return ErrorMessage(absl::StrFormat("Unable to load ELF file \"%s\": %s", file_path.string(),
                                         llvm::toString(object_file_or_error.takeError())));
   }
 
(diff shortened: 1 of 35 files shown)


OrbitBase/PathTest.cpp
```diff
@@ -0,0 +1,196 @@
...
+  EXPECT_EQ(OrbitBase::GetFileName(input), "ext");
+
+  input = "double.ext.txt";
+  EXPECT_EQ(OrbitBase::GetExtension(input), ".txt");
+  EXPECT_EQ(OrbitBase::StripExtension(input), "double.ext");
+  EXPECT_EQ(OrbitBase::StripExtension(OrbitBase::StripExtension(input)), "double");
+  EXPECT_EQ(OrbitBase::GetFileName(input), "double.ext.txt");
+}
+
+// TODO: A test for GetExecutableDir is missing - using this here makes this test
```
florian-kuebler, line 120: Can you create a bug for this and reference it here?

OrbitCore/Path.h
```diff
@@ -9,8 +9,6 @@
 #include <vector>
 
 namespace Path {
```
florian-kuebler, line 11: Do you plan to remove/move this in the future, or do we want to keep OrbitCore::Path for now?
In the latter case, consider renaming the OrbitCore version.

florian-kuebler (approved):
Thanks for that! Besides the two very minor comments, LGTM!

OrbitBase/Path.cpp
```diff
@@ -0,0 +1,72 @@
...
+#include "OrbitBase/Path.h"
+
+#include <filesystem>
+#include <string>
+
+#include "absl/strings/str_replace.h"
+
+namespace OrbitBase {
+
+std::string GetExecutableDir() { return GetDirectory(GetExecutablePath()); }
```
dpallotti, line 14: This feels slightly out of place here as the other functions are "generic" path-related functions, while this returns a very specific file/directory. Does this even belong in `OrbitBase`?

dpallotti (commented):
One fundamental question that needs to be asked here is whether we really want to provide our own somewhat exhaustive wrapper around `std::filesystem::path` that operates only with strings. We are taking it for granted here just because it was already there, but I'm generally not a big fan of just wrapping basic standard library functionality without an overall big plan.

For example: given `Path` in `OrbitBase`, should we just never use `std::filesystem::path` directly? If yes, that's a strict assumption, as that would mean that `Path` needs to cover everything (look at `ListFiles`, what if one wants something similar but that differs slighly), and that it should be used even if one is comfortable with `std::filesystem::path`. If not, that leads to a general inconsistency, where someone uses `Path` and someone else uses `std::filesystem::path`.

OrbitBase/CMakeLists.txt
```diff
@@ -31,24 +32,35 @@ target_sources(OrbitBase PRIVATE
...
         std::filesystem)
 
+
 add_executable(OrbitBaseTests)
 
 target_compile_options(OrbitBaseTests PRIVATE ${STRICT_COMPILE_FLAGS})
 
 target_sources(OrbitBaseTests PRIVATE
-    UniqueResourceTest.cpp
+    PathTest.cpp
```
dpallotti, line 59: Move down to keep the alphabetical order.

OrbitBase/PathTest.cpp
```diff
@@ -0,0 +1,196 @@
...
+  std::string actual = OrbitBase::JoinPath({"C:", "dir1\\dir2", "dir3\\dir4"});
+#ifdef _WIN32
+  std::string expected = "C:dir1\\dir2\\dir3\\dir4";
+#else
+  std::string expected = "C:/dir1\\dir2/dir3\\dir4";
+#endif
+  EXPECT_EQ(actual, expected);
+}
+
+TEST(Path, FileNamesAndExtensions) {
```
dpallotti, line 77: This misses a case with "dir/.file".

OrbitBase/PathTest.cpp
```diff
@@ -0,0 +1,196 @@
...
+
+  input = "double.ext.txt";
+  EXPECT_EQ(OrbitBase::GetExtension(input), ".txt");
+  EXPECT_EQ(OrbitBase::StripExtension(input), "double.ext");
+  EXPECT_EQ(OrbitBase::StripExtension(OrbitBase::StripExtension(input)), "double");
+  EXPECT_EQ(OrbitBase::GetFileName(input), "double.ext.txt");
+}
+
+// TODO: A test for GetExecutableDir is missing - using this here makes this test
+// kind of invalid, since GetExecutableDir is using GetExecutablePath internally...
```
dpallotti, line 121: This is a TODO but what is there to actually do? Can this be either done or the TODO removed?

OrbitBase/Path.cpp
```diff
@@ -0,0 +1,72 @@
...
+  return std::filesystem::path(ToLinuxPath(file_path)).filename().string();
+}
+
+std::string StripExtension(const std::string& file_path) {
+  const std::string extension = GetExtension(file_path);
+  return file_path.substr(0, file_path.length() - extension.length());
+}
+
+std::string GetExtension(const std::string& file_path) {
+  // returns ".ext" (includes point)
```
dpallotti, line 33: Is this comment referring to the function? If yes, it should be in the header.

OrbitBase/include/OrbitBase/Path.h
```diff
@@ -0,0 +1,29 @@
...
+[[nodiscard]] std::string GetExecutableDir();
+[[nodiscard]] std::string GetFileName(const std::string& file_path);
+[[nodiscard]] std::string StripExtension(const std::string& file_path);
+[[nodiscard]] std::string GetExtension(const std::string& file_path);
+[[nodiscard]] std::string GetDirectory(const std::string& any_path);
+[[nodiscard]] std::string JoinPath(const std::vector<std::string>& parts);
+
+[[nodiscard]] std::vector<std::string> ListFiles(
+    const std::string& directory, const std::function<bool(const std::string&)>& filter =
+                                      [](const std::string&) { return true; });
```
dpallotti, line 25: As this is `OrbitBase`, some small documentation would be beneficial. For example: does `GetExtension` include the `.`? Does `StripExtension` return the full path or only the file name without extension?

beckerhe:
I'm with Davide on this. I would prefer to use std::filesystem::path whenever possible and only have functionality in our library that is not available in a portable way, like `GetExecutablePath` (which should return a std::filesystem::path).

Also: I was under the impression we settled for Google style namespaces, didn't we? That would mean "OrbitBase -> orbit_base".

OrbitBase/CMakeLists.txt
dimitry-: done

OrbitBase/Path.cpp
dimitry-: This is widely used by tests for pretty much all components so I think it belongs here.

OrbitBase/Path.cpp
```diff
@@ -0,0 +1,72 @@
...
+namespace OrbitBase {
+
+std::string GetExecutableDir() { return GetDirectory(GetExecutablePath()); }
+
+static std::string ToLinuxPath(std::string_view path) {
+  return absl::StrReplaceAll(path, {{"\\", "/"}});
+}
+
+std::string GetFileName(const std::string& file_path) {
+  // Ok, we cannot use filesystem::path for windows paths
```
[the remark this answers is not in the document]
dimitry-: this is removed.

OrbitBase/Path.cpp
```diff
@@ -0,0 +1,72 @@
...
+  // because linux implementation does not recognize windows path
+  // while windows implementation understands linux-style paths.
```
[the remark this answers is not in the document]
dimitry-: looks like this is not needed after all - removed and replaced calls with std::filesystem::path

OrbitBase/Path.cpp
```diff
@@ -0,0 +1,72 @@
...
+  const std::string file_name = GetFileName(file_path);
+  size_t index = file_name.find_last_of('.');
+  if (index == std::string::npos) {
+    return {};
+  }
+  return file_name.substr(index, file_name.length());
+}
+
+std::string GetDirectory(const std::string& file_path) {
+  return std::filesystem::path(ToLinuxPath(file_path)).parent_path().string();
```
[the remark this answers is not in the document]
dimitry-: not really - removed this one as well.

OrbitBase/Path.cpp
```diff
@@ -0,0 +1,72 @@
...
+    return "";
+  }
+  std::filesystem::path joined{parts[0]};
+  for (size_t i = 1; i < parts.size(); ++i) {
+    joined.append(parts[i]);
+  }
+  return joined.string();
+}
+
+std::vector<std::string> ListFiles(const std::string& directory,
```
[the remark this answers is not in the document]
dimitry-: this method was moved to one place where it is used (App.cpp)

OrbitBase/PathTest.cpp
dimitry-: these functions have been removed.

dimitry-:
> I'm with Davide on this. I would prefer to use std::filesystem::path whenever possible and only have functionality in our library that is not available in a portable way, like `GetExecutablePath` (which should return a std::filesystem::path).
> 
> Also: I was under the impression we settled for Google style namespaces, didn't we? That would mean "OrbitBase -> orbit_base".

done.

dpallotti:
> done.

Should we review or is it still a draft?

dimitry-:
> > done.
> 
> Should we review or is it still a draft?

I am chasing windows build failure - other than that it is ready for review.

OrbitBase/include/OrbitBase/Path.h
```diff
@@ -0,0 +1,17 @@
+// Copyright (c) 2020 The Orbit Authors. All rights reserved.
```
dpallotti, line 1: I would suggest renaming the file to something more specific, like `ExecutablePath`. `Path` is way too generic now.

dpallotti (commented):
Thanks! As I wrote in the comment, I think we should rename to something more specific. Other than that it looks good.

ElfUtils/ElfFile.cpp
```diff
@@ -253,26 +253,26 @@ bool ElfFileImpl<llvm::object::ELF32LE>::Is64Bit() const {
...
   return ElfFile::Create(file_path, llvm::object::OwningBinary<llvm::object::ObjectFile>(
                                         std::move(object_file_or_error.get()), std::move(buffer)));
 }
 
-ErrorMessageOr<std::unique_ptr<ElfFile>> ElfFile::Create(std::string_view file_path) {
+ErrorMessageOr<std::unique_ptr<ElfFile>> ElfFile::Create(const std::filesystem::path& file_path) {
   // TODO(hebecker): Remove this explicit construction of StringRef when we
   // switch to LLVM10.
-  const llvm::StringRef file_path_llvm{file_path.data(), file_path.size()};
+  const llvm::StringRef file_path_llvm{file_path.c_str()};
```
beckerhe: Personal preference: I'd say `file_path.native()` would be a tiny bit nicer than `file_path.c_str()` because that superfluous call to strlen under the hood hurts. But I don't feel particularly strong about that. :-)

OrbitBase/include/OrbitBase/Path.h
beckerhe: Good point. Especially since OrbitCore/Path is still there. A different name might be less confusing.

beckerhe (approved):
I like that change! Thanks! LGTM!

OrbitBase/include/OrbitBase/Path.h
dimitry-: done.

OrbitBase/include/OrbitBase/Path.h
```diff
@@ -0,0 +1,17 @@
...
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#ifndef ORBIT_BASE_PATH_H_
+#define ORBIT_BASE_PATH_H_
+
+#include <filesystem>
+#include <vector>
+
+namespace orbit_base {
```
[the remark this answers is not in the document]
dimitry-: done.

ElfUtils/ElfFile.cpp
dimitry-: this apparently requires a string intermediate variable on windows so I introduced one. which means this c_str is now called on a string.

OrbitBase/PathTest.cpp
dimitry-: I wrote a smoke test for it - removed todo.

OrbitCore/Path.h
dimitry-: I am thinking about moving it too.. this is for follow up cl

PR #1539, opened 2020-12-02
Include-what-you-use

beckerhe:
This adds include-what-you-use to our build system.

include-what-you-use is a clang-based tool and needs to properly parse all the source code,
which means it needs all the compiler flags a real compiler takes. It needs all source code file,
including headers from dependencies and headers generated by code generators like Qt's uic
or the protobuf compiler.

That's why we handle include-what-you-use as a separate build target inside of CMake, because
we will have all compiler flags available and we can declare code generation targets as dependencies
to this build target.

So there is a new build target `iwyu` which is not built by default, but can be requested by `make iwyu`.
It will invoke the protobuf compiler and uic if this hasn't been done yet.

I also added a conan profile which will only build the `iwyu` target. The profile is also called `iwyu`. It can
be built by calling `./build.sh iwyu`.

The iwyu results will be in a file called `iwyu_results.txt` inside of the build-directory. The CI build script
reads that file and filters it for files touched by the current PR and print the errors to the build log.
It also prints the full, unfiltered list which provides an easy way to get an overview.

The conan-profile and build-target will only be available on Linux. Windows is unfortunately not supported.
That also means that Windows-only-cpp files can't be checked by this presubmit check!

After meging this the presubmit check will run with every PR, but it won't fail. It might take several months
before all includes have been fixed and I don't want to generate the impression that a failed presubmit check
is okay. There is also a chance that adjustments need to be made - especially for external dependencies. Symbols
are often declared in private headers which are not supposed to be included. A mapping file can tell iwyu that instead
it should suggest the public header for inclusion. Some adjustments have already been made, see `contrib/iwyu/` for reference.

The last commit "temp" in here will be removed before merging. It redirects the gcc9 presubmit check to the iwyu-one.
This way I can test the check before enabling it globally. You can check the gcc9-check's build log for the iwyu-output.


cmake/iwyu.cmake
```diff
@@ -0,0 +1,74 @@
...
+
+if(NOT WIN32)
+  include(ProcessorCount)
+  ProcessorCount(CPU_COUNT)
+
+  find_package(Python3 COMPONENTS Interpreter)
+  find_program(iwyu_tool_py NAMES iwyu_tool)
+
+  if (iwyu_tool_py AND TARGET Python3::Interpreter)
+    add_custom_command(OUTPUT iwyu_commands.json
```
florian-kuebler, line 13: Can you add a very brief comment about what you are in each of the two custom commands?

kokoro/builds/build.sh
```diff
@@ -208,6 +214,32 @@ if [ -n "$1" ]; then
...
+    echo -e "\n\n\nThe following include problems occur in files this PR touched:"
+    readonly REFERENCE="${KOKORO_GITHUB_PULL_REQUEST_TARGET_BRANCH:-origin/master}"
+    readonly MERGE_BASE="$(git merge-base $REFERENCE HEAD)" # Merge base is the commit on master this PR was branched from.
+
+    PATTERN_FILE="$(mktemp)"
+    git diff -U0 --no-color --relative --name-only $MERGE_BASE | sed -e 's/$/:/' > ${PATTERN_FILE}
+    if grep -A1 -F "error:" build/iwyu_results.txt | grep -A1 -F -f ${PATTERN_FILE}; then
+      echo -e "\n\n"
+
+      # Replace `true` by `exit 1` here when the check is ready to fail in production!
```
florian-kuebler, line 233: I'd prefer to start having the CI failing for touched files from now. WDYT?

florian-kuebler (approved):
Thanks Henning, this is a really nice improvement!
But I would suggest that the CI should start failing for the touched files starting from now. It should still print all the files to get an overview, but if you are touching a file right now, it would be totally fine to have the CI failing.

cmake/iwyu.cmake
beckerhe: Yes. good idea.

kokoro/builds/build.sh
beckerhe: I long debated that with myself. In the end I decided against it because I expect iwyu not to be correct in many cases yet.
The mapping file for external libraries is probably not complete yet. I tried to work through all the edge cases but it's hard
because there are so many includes missing.

So I'd prefer to not fail for the next couple of days, because otherwise folks have to submit with the test failing. And if that
phase takes too long folks will stop checking the logfile because they assume iwyu failed for the usual reasons.

florian-kuebler: Okay, I am fine with this for a couple of days :). But we really should aim to enable this. Once again, thanks a lot.

cmake/iwyu.cmake
beckerhe: Done

danielfenner (approved):
Really cool, thanks for looking into this!

PR #1540, opened 2020-12-02
Also send CaptureFinishedCommand to clients of ProducerSideService

dpallotti:
This signals the producers that the service has finished collecting data from
all producers.

Bug: http://b/174028631


florian-kuebler (approved):
LGTM. Thanks for making this change!

PR #1541, opened 2020-12-03
Fix Crash in TracepointEventBuffer::GetNumTracepointsForThreadId

florian-kuebler:
With the recent change #1537 a map access in
"GetNumTracePointsForThreadId" was changed from "[]" to ".at",
assuming the key was in that map. This turned out to be a wrong
assumption.

This changes the code to handle the case where the data is not there
without modifying the map.

Bug: http://174730511
Test: Compile and take a capture

### DIFF
--- a/OrbitClientData/TracepointEventBuffer.cpp
+++ b/OrbitClientData/TracepointEventBuffer.cpp
@@ -70,7 +70,7 @@ void TracepointEventBuffer::ForEachTracepointEventOfThreadInTimeRange(
     for (const auto& entry : tracepoint_events_) {
       ForEachTracepointEventInEventMapInTimeRange(min_tick, max_tick, entry.second, action);
     }
-  } else if (thread_id == -1 /*TODO(b/166238019) Use a proper constant again*/) {
+  } else if (thread_id == -1 /*TODO(b/166238019) Use a proper constant for "all fake thread id"*/) {
     for (const auto& entry : tracepoint_events_) {
       if (entry.first != kNotTargetProcessThreadId) {
         ForEachTracepointEventInEventMapInTimeRange(min_tick, max_tick, entry.second, action);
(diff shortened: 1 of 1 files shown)


OrbitClientData/TracepointEventBuffer.cpp
```diff
@@ -89,7 +89,12 @@ uint32_t TracepointEventBuffer::GetNumTracepointsForThreadId(int32_t thread_id)
...
     if (thread_id == -1 /*TODO(b/166238019) Use a proper constant again*/) {
-      return num_total_tracepoints_ - tracepoint_events_.at(kNotTargetProcessThreadId).size();
+      auto const not_target_process_tracepoints_it =
+          tracepoint_events_.find(kNotTargetProcessThreadId);
+      if (not_target_process_tracepoints_it != tracepoint_events_.end()) {
+        return num_total_tracepoints_ - not_target_process_tracepoints_it->second.size();
+      }
+      return num_total_tracepoints_;
     }
   }
```
dpallotti, line 99: This is not related to this change, but here the mutex is released to then be reacquired in `GetTracepointsOfThread`. While this shouldn't cause issue, it seems wrong to me. This changes the previous behaviour. Why is that, when a `std::recursive_mutex` is used anyway?

OrbitClientData/TracepointEventBuffer.cpp
```diff
@@ -89,7 +89,12 @@ uint32_t TracepointEventBuffer::GetNumTracepointsForThreadId(int32_t thread_id)
       return num_total_tracepoints_;
     }
     if (thread_id == -1 /*TODO(b/166238019) Use a proper constant again*/) {
-      return num_total_tracepoints_ - tracepoint_events_.at(kNotTargetProcessThreadId).size();
+      auto const not_target_process_tracepoints_it =
```
dpallotti, line 92: We usually put `const` in front, I think we should do the same even when using `auto`.

OrbitClientData/TracepointEventBuffer.cpp
```diff
@@ -89,7 +89,12 @@ uint32_t TracepointEventBuffer::GetNumTracepointsForThreadId(int32_t thread_id)
...
+          tracepoint_events_.find(kNotTargetProcessThreadId);
+      if (not_target_process_tracepoints_it != tracepoint_events_.end()) {
```
dpallotti, line 94: Personally I would invert what causes the early return: `if (not_target_process_tracepoints_it == tracepoint_events_.end())`. So that the "exception" causes the early return. Does that make sense?

OrbitClientData/TracepointEventBuffer.cpp
```diff
@@ -89,7 +89,12 @@ uint32_t TracepointEventBuffer::GetNumTracepointsForThreadId(int32_t thread_id)
       return num_total_tracepoints_;
     }
     if (thread_id == -1 /*TODO(b/166238019) Use a proper constant again*/) {
```
dpallotti, line 91: Since the constant is no longer used, can there at least be a comment on what this -1 represents, at least in the TODO?

dpallotti (approved):

OrbitClientData/TracepointEventBuffer.cpp
florian-kuebler: Changed it.

OrbitClientData/TracepointEventBuffer.cpp
florian-kuebler: I changed the comment.

OrbitClientData/TracepointEventBuffer.cpp
florian-kuebler: Done.

OrbitClientData/TracepointEventBuffer.cpp
florian-kuebler: Done.

florian-kuebler (commented):
Thanks for the review. I addressed your comments.

danielfenner (approved):
lgtm - let's get this in to fix the crash. 

This looks a bit more convoluted than necessary but let's follow up on that later.

PR #1545, opened 2020-12-03
Optimize PerfEventQueue

dpallotti:


florian-kuebler (commented):
Can you extend the commit message with a sentence of why you moved it to its own file?

OrbitLinuxTracing/PerfEventQueue.h
```diff
@@ -37,20 +33,12 @@ class PerfEventQueue {
...
-      return lhs->front()->GetTimestamp() > rhs->front()->GetTimestamp();
-    }
-  };
-
-  std::priority_queue<std::queue<std::unique_ptr<PerfEvent>>*,
-                      std::vector<std::queue<std::unique_ptr<PerfEvent>>*>,
-                      QueueFrontTimestampReverseCompare>
-      queues_heap_{};
-  absl::flat_hash_map<int, std::unique_ptr<std::queue<std::unique_ptr<PerfEvent>>>> queues_{};
+  void MoveDownHeapFront();
```
florian-kuebler, line 36: Can you add brief comments on what these methods do?

dpallotti:
> Can you extend the commit message with a sentence of why you moved it to its own file?

Will do. It was just for better clarity. Or do you think that they should be in the same file?

OrbitLinuxTracing/PerfEventQueue.cpp
```diff
@@ -26,30 +23,76 @@ void PerfEventQueue::PushEvent(std::unique_ptr<PerfEvent> event) {
...
 }
 
 bool PerfEventQueue::HasEvent() { return !queues_heap_.empty(); }
 
-PerfEvent* PerfEventQueue::TopEvent() { return queues_heap_.top()->front().get(); }
+PerfEvent* PerfEventQueue::TopEvent() { return queues_heap_.front()->front().get(); }
 
 std::unique_ptr<PerfEvent> PerfEventQueue::PopEvent() {
-  std::queue<std::unique_ptr<PerfEvent>>* top_queue = queues_heap_.top();
-  queues_heap_.pop();
```
florian-kuebler: I would really add a comment here why you don not pop the event queue here. I see that it was needed in the empty case anyways, but I feel a comment would help here.

OrbitLinuxTracing/PerfEventQueue.cpp
```diff
@@ -26,30 +23,76 @@ void PerfEventQueue::PushEvent(std::unique_ptr<PerfEvent> event) {
...
-    // The queue with the older PerfEvent is always removed. If it's not empty, re-insert it so that
-    // it is in the right position in the heap after the front of the queue has been removed.
-    queues_heap_.push(top_queue);
+    MoveDownHeapFront();
   }
 
   return top_event;
 }
 
+void PerfEventQueue::MoveDownHeapFront() {
```
florian-kuebler, line 53: To me those methods looks good. But with such algorithmic methods are always error-prune. I saw that you added an additional test case (thanks for that). Have you checked that the tests covers the corner cases here?

florian-kuebler (approved):
Thanks for diving into this. That seems like a huge performance improvement! Only left some minor comments. LGTM!

florian-kuebler:
> > Can you extend the commit message with a sentence of why you moved it to its own file?
> 
> Will do. It was just for better clarity. Or do you think that they should be in the same file?

No, after looking at those files, it made totally sense to have them separate, I'd just prefer a brief statement for the commit log.

florian-kuebler:
I just noticed that we now don't have any test for the PerfEventProcessor. That class seems to be pretty simple now, but could you also add a small test for that?

OrbitLinuxTracing/PerfEventQueue.cpp
florian-kuebler: Okay thanks. Yeah we never know if we have all the corner cases. I just wanted to double check. I guess thats totally fine now. Thanks!

dpallotti:
> > > Can you extend the commit message with a sentence of why you moved it to its own file?
> > 
> > 
> > Will do. It was just for better clarity. Or do you think that they should be in the same file?
> 
> No, after looking at those files, it made totally sense to have them separate, I'd just prefer a brief statement for the commit log.

This is done.

> I just noticed that we now don't have any test for the PerfEventProcessor. That class seems to be pretty simple now, but could you also add a small test for that?

On it.

dpallotti:
> I just noticed that we now don't have any test for the PerfEventProcessor. That class seems to be pretty simple now, but could you also add a small test for that?

Done, please have a look.

OrbitLinuxTracing/PerfEventProcessorTest.cpp
```diff
@@ -0,0 +1,125 @@
...
+}  // namespace
+
+TEST_F(PerfEventProcessorTest, ProcessOldEvents) {
+  EXPECT_CALL(mock_visitor_, visit).Times(0);
+  processor_.AddEvent(MakeFakePerfEvent(11, MonotonicTimestampNs()));
+  processor_.AddEvent(MakeFakePerfEvent(11, MonotonicTimestampNs()));
+  processor_.AddEvent(MakeFakePerfEvent(22, MonotonicTimestampNs()));
+  processor_.ProcessOldEvents();
+
+  ::testing::Mock::VerifyAndClearExpectations(&mock_visitor_);
```
florian-kuebler, line 57: Uh that is a nice feature, I wasn't aware of :)

florian-kuebler (approved):
LGTM. Thanks!

OrbitLinuxTracing/PerfEventProcessorTest.cpp
dpallotti: I used it extensively in the IPC tests. In time-sensitive code, it's the way I found to say "`EXPECT_CALL` by now".

PR #1546, opened 2020-12-03
Apply include-what-you-use findings to OrbitLinuxTracing

beckerhe:
As a test run I manually applied all findings from include-what-you-use
to OrbitLinuxTracing. I was in particular interested how it works with
system headers (Linux, Posix, etc.) which are used a lot by OrbitLinuxTracing.

So far no complications.

### DIFF
--- a/OrbitLinuxTracing/ContextSwitchManager.cpp
+++ b/OrbitLinuxTracing/ContextSwitchManager.cpp
@@ -4,6 +4,10 @@
 
 #include "ContextSwitchManager.h"
 
+#include <utility>
+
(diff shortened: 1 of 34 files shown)


beckerhe:
@dpallotti I can easily rebase that on top of #1545. So I suggest we merge yours first, do you agree?

dpallotti:
> @dpallotti I can easily rebase that on top of #1545. So I suggest we merge yours first, do you agree?

That would be better, thanks! I'll let you know when it's in.

dpallotti (approved):
#1545 is in, please go ahead. Minor details discussed offline.

PR #1547, opened 2020-12-04
Refactor ProcessManagerImpl

antonrohr:
This adds a dtor to ProcessManagerImpl that calls Shutdown.

### DIFF
--- a/OrbitClientServices/ProcessManager.cpp
+++ b/OrbitClientServices/ProcessManager.cpp
@@ -4,15 +4,16 @@
 
 #include "OrbitClientServices/ProcessManager.h"
 
-#include <chrono>
+#include <absl/synchronization/mutex.h>
+
+#include <algorithm>
 #include <memory>
 #include <string>
+#include <thread>
 
 #include "OrbitBase/Logging.h"
+#include "OrbitBase/Result.h"
 #include "OrbitClientServices/ProcessClient.h"
-#include "grpcpp/grpcpp.h"
-#include "outcome.hpp"
-#include "symbol.pb.h"
 
 namespace {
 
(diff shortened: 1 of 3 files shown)


OrbitClientServices/ProcessManager.cpp
```diff
@@ -23,6 +24,9 @@ class ProcessManagerImpl final : public ProcessManager {
  public:
   explicit ProcessManagerImpl(const std::shared_ptr<grpc::Channel>& channel,
                               absl::Duration refresh_timeout);
+  ProcessManagerImpl(ProcessManagerImpl&&) = default;
+  ProcessManagerImpl& operator=(ProcessManagerImpl&&) = default;
+  ~ProcessManagerImpl() override { Shutdown(); }
```
dimitry-, line 29: nit: wanted to do that but while you are here maybe rename Shutdown() to ShutdownAndWait()?

dimitry-: Also are we sure it does not throw bad_alloc? Maybe make it nothrow as well.

antonrohr: I am not completely sure if I understood the nothrow part. I made the function noexcept now. Please take a look

OrbitClientServices/ProcessManager.cpp
```diff
@@ -85,7 +89,7 @@ void ProcessManagerImpl::Start() {
   worker_thread_ = std::thread([this] { WorkerFunction(); });
 }
 
-void ProcessManagerImpl::Shutdown() {
+void ProcessManagerImpl::ShutdownAndWait() noexcept {
   shutdown_mutex_.Lock();
   shutdown_initiated_ = true;
   shutdown_mutex_.Unlock();
```
dimitry-, line 95: ok, this is not safe - worker_thread_.join() below can throw an exception. Maybe we need JoinThreadOrDie function here which does not throw.

beckerhe: I think it's fine to just swallow the exception in that case. Something like:
```
try {
worker_thread_.join();
catch (const std::exception&) {}
```

@dimitry- WDYT?

dimitry-: well, I think we kind of want to add FATAL in exception handling part at least, preferably checking for conditions before too.

antonrohr: I did not add another JoinThreadOrDie function, but catch the potential exception there. ptal

dimitry- (approved):

PR #1548, opened 2020-12-04
Fix save capture when there is no capture data crash

dimitry-:
Test: start orbit -> save capture (without loading/taking capture)
      check that it displays message instead of crashing.


IrinaShkviro (approved):
Thanks a lot! LGTM!

PR #1549, opened 2020-12-04
Build config for the jobs collecting the artefacts.

danielfenner:
The build file does a trivial repackaging to include the collector into
Orbit.zip.

Bug: http://b/167356615
Test: Tried it out in a slightly different setup. What can possibly go wrong?


beckerhe (approved):
Looks fine to me!

PR #1550, opened 2020-12-04
Join TracepointInfoManager and TracepointEventBuffer to TracepointData

florian-kuebler:
This change is a major refactoring of TracepointEventBuffer and
TracepointInfoManager. The classes are joined, as they are both
just data containers for tracepoint related info (similar to
CallstackData).

Most of the methods were also touched and simplified. The use of
absl classes was introduced whereever possible.

Test: Compile & run tests

**Please review this PR as if the code was completely new.**


OrbitBase/include/OrbitBase/ThreadConstants.h
```diff
@@ -9,5 +9,12 @@ namespace orbit_base {
 // Represents a fake thread id to specify the set of all thread ids of the current process.
 static constexpr int32_t kAllProcessThreadsFakeTid = -1;
 
+// Represents a fake thread id to specify the set of all thread ids of all processes on the system.
+static constexpr int32_t kAllThreadsOfAllProcessesFakeTid = -2;
```
dimitry-, line 13: nit: I would drop Fake from the name, it does not add anything.

OrbitClientData/include/OrbitClientData/TracepointData.h
```diff
@@ -0,0 +1,72 @@
...
+  uint32_t GetNumTracepointsForThreadId(int32_t thread_id) const;
+
+  bool AddUniqueTracepointInfo(uint64_t key, orbit_grpc_protos::TracepointInfo tracepoint);
+
+  [[nodiscard]] orbit_grpc_protos::TracepointInfo GetTracepointInfo(uint64_t key) const;
+  [[nodiscard]] bool HasTracepointKey(uint64_t key) const;
+
+  void ForEachUniqueTracepointInfo(
+      const std::function<void(const orbit_client_protos::TracepointInfo&)>& action) const {
+    absl::MutexLock lock(&unique_tracepoints_mutex_);
```
dimitry-, line 51: is there a reason this is in the header file? can this be moved to cpp?

OrbitClientData/TracepointData.cpp
```diff
@@ -0,0 +1,139 @@
...
+// found in the LICENSE file.
+
+#include "OrbitClientData/TracepointData.h"
+
+#include "OrbitBase/Logging.h"
+#include "OrbitBase/ThreadConstants.h"
+
+using orbit_client_protos::TracepointEventInfo;
+
+void TracepointData::EmplaceTracepointEvent(uint64_t time, uint64_t tracepoint_hash,
```
dimitry-, line 12: nit: tracepoint_hash -> tracepoint_key

OrbitClientData/TracepointData.cpp
```diff
@@ -0,0 +1,139 @@
...
+  CHECK(HasTracepointKey(tracepoint_hash));
+  event.set_tracepoint_info_key(tracepoint_hash);
+  event.set_tid(thread_id);
+  event.set_pid(process_id);
+  event.set_cpu(cpu);
+
+  int32_t insertion_thread_id =
+      (is_same_pid_as_target) ? thread_id : orbit_base::kNotTargetProcessThreadId;
+
+  auto [event_map_iterator, unused_inserted] =
```
dimitry-, line 29: we decided not to use _ here?

OrbitClientData/TracepointData.cpp
```diff
@@ -0,0 +1,139 @@
...
+      thread_id_to_time_to_tracepoint_.try_emplace(insertion_thread_id);
+  auto [unused_iterator, event_inserted] =
+      event_map_iterator->second.try_emplace(time, std::move(event));
+  if (!event_inserted) {
+    ERROR(
```
dimitry-, line 34: in which situations this can happen?

OrbitClientData/TracepointData.cpp
```diff
@@ -0,0 +1,139 @@
...
+        "Tracepoint event was not inserted as there was already an event on this time and "
+        "thread.");
+  }
+}
+
+[[nodiscard]] const std::map<uint64_t, orbit_client_protos::TracepointEventInfo>&
+TracepointData::GetTracepointsOfThread(int32_t thread_id) const {
+  absl::MutexLock lock(&mutex_);
+  static std::map<uint64_t, orbit_client_protos::TracepointEventInfo> empty;
+  const auto& it = thread_id_to_time_to_tracepoint_.find(thread_id);
```
dimitry-, line 44: seems like there are multiple problems with this method. the returned map can be modified while the caller is reading it because read is not guarder after this method is called.

I would consider dropping this method and replacing the call with one of ForEach methods.

florian-kuebler: Turns out, it is unused anyways. I'll drop it completely.

OrbitClientData/TracepointData.cpp
florian-kuebler: Daniel and I discussed this. As we have a nanosecond resolution, this is nearly impossible that this happens, as the kernel's tracepoint macro would need to be hit within the same nanosecond on the same thread twice. Still it feels strange to either ignore that fact, or to add a CHECK here.

OrbitClientData/TracepointData.cpp
florian-kuebler: We had a lengthy discussion in our daily some time ago (might be during your vacation). The result was "use unused_ prefix". The main reason for not using "_" was that it will still bind it to a name. So for nested loops etc. it can't be reused.

OrbitClientData/include/OrbitClientData/TracepointData.h
florian-kuebler: Moved it. Thanks!

OrbitBase/include/OrbitBase/ThreadConstants.h
florian-kuebler: Agree, thanks!

OrbitClientData/TracepointData.cpp
dimitry-: ok

dimitry- (approved):

OrbitClientData/include/OrbitClientData/TracepointData.h
```diff
@@ -0,0 +1,60 @@
...
+                              int32_t thread_id, int32_t cpu, bool is_same_pid_as_target);
+
+  void ForEachTracepointEventOfThreadInTimeRange(
+      int32_t thread_id, uint64_t min_tick, uint64_t max_tick_exclusive,
+      const std::function<void(const orbit_client_protos::TracepointEventInfo&)>& action) const;
+
+  void ForEachTracepointEvent(
+      const std::function<void(const orbit_client_protos::TracepointEventInfo&)>& action) const;
+
+  uint32_t GetNumTracepointsForThreadId(int32_t thread_id) const;
```
danielfenner, line 39: nit: GetNumTracepointEventsForThreadId would be a more logical name. I don't know how many files this would touch - feel free to disregard.

OrbitClientData/include/OrbitClientData/TracepointData.h
```diff
@@ -0,0 +1,60 @@
...
+
+  bool AddUniqueTracepointInfo(uint64_t key, orbit_grpc_protos::TracepointInfo tracepoint);
```
danielfenner, line 41: Can orbit_grpc_protos::TracepointInfo tracepoint be handed over as a const ref?

OrbitClientData/include/OrbitClientData/TracepointData.h
```diff
@@ -0,0 +1,60 @@
...
+
+  [[nodiscard]] orbit_grpc_protos::TracepointInfo GetTracepointInfo(uint64_t hash) const;
+  [[nodiscard]] bool HasTracepointKey(uint64_t key) const;
+
+  void ForEachUniqueTracepointInfo(
+      const std::function<void(const orbit_client_protos::TracepointInfo&)>& action) const;
+
+ private:
+  int32_t num_total_tracepoints_ = 0;
```
danielfenner, line 50: for readability: num_total_tracepoint_events_

OrbitClientData/include/OrbitClientData/TracepointData.h
```diff
@@ -0,0 +1,60 @@
...
+#include <functional>
+#include <limits>
+#include <map>
+#include <vector>
+
+#include "capture_data.pb.h"
+#include "tracepoint.pb.h"
+
+/*
+ * Stores `TracepointEventInfos` and a mapping from tracepoint key to unique `TracepointInfo`s.
```
danielfenner, line 21: I like class comments that are understandable with very little context/knowledge. Suggestion:

TracepointData stores all tracepoint related information on the Client/Ui side. 
Single events in which a tracepoint got hit are represented by a 'TracepointEventInfo'. These events are stored ordered by the time stamp of the event such that one can iterate over the events in a time interval in an efficient manner. 
The class offers methods to add events and iterate over over them.

Other than the events themselves the class stores a list containing all the available tracepoints on the target system ('TracepointInfo's). This is stored in a map that from an integer identifier to a 'TracepointInfo'. This map is also used to compress the wire format of the events (the events contain an identifier rather than the full name of the travepoint they correspond to).

OrbitClientData/include/OrbitClientData/TracepointData.h
```diff
@@ -0,0 +1,72 @@
+// Copyright (c) 2020 The Orbit Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#ifndef ORBIT_CORE_TRACEPOINT_EVENT_BUFFER_H_
+#define ORBIT_CORE_TRACEPOINT_EVENT_BUFFER_H_
+
+#include <absl/container/flat_hash_map.h>
```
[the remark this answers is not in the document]
florian-kuebler: Didn't we decided on friday to use <> for all external dependencies?

OrbitClientData/include/OrbitClientData/TracepointData.h
florian-kuebler: I guess that method is considered being a sink. So we encourage to move data into it.

PR #1553, opened 2020-12-05
Move GetExecutablePath(pid) to OrbitBase

dimitry-:
This implementation is only available for Linux
(because atm we need it on Linux only)

Test: run OrbitBaseTests


dpallotti (approved):

dimitry-:
oh... i merged fixup commit :/

PR #1558, opened 2020-12-07
Refactoring: Move some method from ServiceUtils

dimitry-:
ReadFileToString moved to orbit_base, ReadModules and
ParseMaps moved to ElfUtils.

Test: run OrbitBaseTests and ElfUtilsTests

### DIFF
--- a/ElfUtils/LinuxMap.cpp
+++ b/ElfUtils/LinuxMap.cpp
@@ -1,0 +1,117 @@
+// Copyright (c) 2020 The Orbit Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#include "ElfUtils/LinuxMap.h"
+
+#include <absl/strings/str_format.h>
+#include <absl/strings/str_split.h>
+#include <sys/stat.h>
+#include <unistd.h>
+
+#include <filesystem>
+
+#include "ElfUtils/ElfFile.h"
+#include "OrbitBase/Logging.h"
+#include "OrbitBase/ReadFileToString.h"
+#include "OrbitBase/SafeStrerror.h"
+
+namespace orbit_elf_utils {
+
+using ElfUtils::ElfFile;
+using orbit_grpc_protos::ModuleInfo;
+
+static ErrorMessageOr<uint64_t> FileSize(const std::string& file_path) {
+  struct stat stat_buf {};
+  int ret = stat(file_path.c_str(), &stat_buf);
+  if (ret != 0) {
+    return ErrorMessage(absl::StrFormat("Unable to call stat with file \"%s\": %s", file_path,
+                                        SafeStrerror(errno)));
+  }
+  return stat_buf.st_size;
+}
+
+ErrorMessageOr<std::vector<ModuleInfo>> ReadModules(int32_t pid) {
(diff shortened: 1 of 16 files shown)


ElfUtils/LinuxMap.cpp
```diff
@@ -0,0 +1,117 @@
...
+#include <unistd.h>
+
+#include <filesystem>
+
+#include "ElfUtils/ElfFile.h"
+#include "OrbitBase/File.h"
+#include "OrbitBase/Logging.h"
+#include "OrbitBase/SafeStrerror.h"
+
+namespace elf_utils {
```
dpallotti, line 19: Based on the discussions, the namespace should be `orbit_elf_utils` (unlike the dir name).

OrbitBase/File.cpp
```diff
@@ -0,0 +1,24 @@
+// Copyright (c) 2020 The Orbit Authors. All rights reserved.
```
dpallotti, line 1: The name of this file is way too generic (`File.h`/`File.cpp`). If for now it only contains this one function, for me it's also fine to just call it `ReadFileToString.cpp`/`.h`, but feel free to come up with something better, as long as it's more specific.

OrbitBase/File.cpp
```diff
@@ -0,0 +1,24 @@
...
+
+namespace orbit_base {
+
+ErrorMessageOr<std::string> ReadFileToString(const std::filesystem::path& file_name) {
+  std::ifstream file_stream(file_name);
+  if (file_stream.fail()) {
+    return ErrorMessage(
+        absl::StrFormat("Unable to read file %s: %s", file_name.string(), SafeStrerror(errno)));
+  }
+  return std::string{std::istreambuf_iterator<char>{file_stream}, std::istreambuf_iterator<char>{}};
```
dpallotti, line 21: What's the reason to not pass `file_stream.begin()` and `file_stream.end()` instead of explicitly constructing the iterators? Especially the second one, `std::istreambuf_iterator<char>{}`, is quite cryptic.

ElfUtils/LinuxMap.cpp
dimitry-: done.

OrbitBase/File.cpp
dimitry-: done.

OrbitBase/File.cpp
dimitry-: there is no file_stream.begin()

dpallotti: Right, looks like this is the way to do it.

ElfUtils/include/ElfUtils/LinuxMap.h
```diff
@@ -0,0 +1,23 @@
+// Copyright (c) 2020 The Orbit Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#if defined(__linux)
```
dpallotti, line 5: Wouldn't it be more common to have the `#ifndef ELF_UTILS_LINUX_MAP_H_` outside of `#if defined(__linux)`? Of course it makes no practical difference, but one always expects the include guards first.

dpallotti (approved):

ElfUtils/include/ElfUtils/LinuxMap.h
dimitry-: moved

PR #1561, opened 2020-12-08
Speedup builds by faster cleanup.

danielfenner:
Prefer to delete directories using 'rm -rf' instead of iterating over all the files. 

Cuts 6-8 minutes off the windows build times.


kokoro/builds/build.sh
```diff
@@ -52,36 +52,33 @@ if [ -n "$1" ]; then
...
+        grep -v 'build/testresults' | \
         while read file; do
-          if [[ -d $file ]]; then
-            # That might give an error message when the directory is not empty.
-            # That's okay.
-            rmdir --ignore-fail-on-non-empty "$file"
-          elif [[ -e $file ]]; then
-            rm "$file"
-          fi
+          rm -rf "$file"
```
beckerhe, line 67: Have you considered a single find call as well? Might also be performant and a bit cleaner. Something like this:

```
find "${MOUNT_POINT}" ! -path "${REPO_ROOT}/kokoro*" ! -path "${REPO_ROOT}/build" ! -path "${REPO_ROOT}/build/testresults*" -delete
```

The first `! -path` option should exclude the kokoro directory and everything below. The second -path should exclude the build directory but not what's inside. The last -path excludes the build/testresults directory and everything below.

kokoro/builds/build.sh
```diff
@@ -52,36 +52,33 @@ if [ -n "$1" ]; then
...
         done
+      # ... and everything from keystore.
+      rm -rf "$KEYSTORE_PATH"
```
beckerhe, line 81: Should work here as well:

```
find "${MOUNT_POINT}" ! -path "${REPO_ROOT}/kokoro*" ! -path "${REPO_ROOT}/build*" -delete
```

beckerhe (commented):
Thanks Daniel!

I added a suggestion about using a single find-call instead. Looking forward to hear what you think about that.

kokoro/builds/build.sh
danielfenner: I agree that the two find's are not  great for readability but the point here is that I'd like to delete the directories on the upper level and avoid traversing the rubbish below. 
The "-delete" option implies "-depth". So this is a full traversal again.

beckerhe: As far as I know you can't avoid the traversal. It's either done by Bash (previous solution), or `rm -r` (your solution) or `find` (my solution). I would assume find should be equally fast than `rm -r` since both solutions don't spawn new processes like the bash solution does.

danielfenner: Hmm, I hoped that there is optimization going on "further down" when unlinking an entire dir. Not so sure about that anymore. Seems to be quite a [rabbit hole](https://unix.stackexchange.com/questions/37329/efficiently-delete-large-directory-containing-thousands-of-files). One optimization is to remove the files in an order that does not cause intermediate resorting of things in the tree underlying the file system. Both '-delete' and 'rm -rf' seem to do that. I'll just try if this works ...

beckerhe: ...and this might be completely different when these unix tools run on Windows. I guess there is only one way to find out. So thanks a lot for testing this.

danielfenner: The solution using 'find ... -delete' take 5 seconds, the 'rm -rf' takes 8 seconds and the old solution took 7:40

Under Linux there is not much of a difference (only checked the build times, no benchmarking).

fwiw: I think the order of the deletion is the issue here - not the spawning of sub processes. Most of the 7 minutes is spend in "sys", there are no new threads starting up. Something in the stack is not as optimized on Windows as it is on Linux.

beckerhe (approved):
Thanks a lot, Daniel. Looks a lot cleaner now!

PR #1562, opened 2020-12-08
Namespace alignment

radu-danciu:
As proposed by dpallotti@.

Proposed as one PR despite the separate commits as there are a number of dependencies in between modules such that it's unlikely any real benefit could be had from separating (in between too few of those commits is the codebase in a buildable state).

### DIFF
--- a/OrbitBase/Tracing.cpp
+++ b/OrbitBase/Tracing.cpp
@@ -11,19 +11,20 @@
 #include "OrbitBase/Profiling.h"
 #include "OrbitBase/ThreadPool.h"
 
-using orbit_tracing::Listener;
-using orbit_tracing::Scope;
-using orbit_tracing::TimerCallback;
+using orbit_base::TracingListener;
+using orbit_base::TracingScope;
+using orbit_base::TracingTimerCallback;
 
 ABSL_CONST_INIT static absl::Mutex global_tracing_mutex(absl::kConstInit);
-ABSL_CONST_INIT static Listener* global_tracing_listener = nullptr;
+ABSL_CONST_INIT static TracingListener* global_tracing_listener = nullptr;
 
-namespace orbit_tracing {
+namespace orbit_base {
 
-Scope::Scope(orbit_api::EventType type, const char* name, uint64_t data, orbit::Color color)
+TracingScope::TracingScope(orbit_api::EventType type, const char* name, uint64_t data,
+                           orbit::Color color)
     : encoded_event(type, name, data, color) {}
 
-Listener::Listener(TimerCallback callback) {
+TracingListener::TracingListener(TracingTimerCallback callback) {
   constexpr size_t kMinNumThreads = 1;
   constexpr size_t kMaxNumThreads = 1;
   thread_pool_ = ThreadPool::Create(kMinNumThreads, kMaxNumThreads, absl::Milliseconds(500));
@@ -37,7 +38,7 @@ Listener::Listener(TimerCallback callback) {
   active_ = true;
 }
 
-Listener::~Listener() {
+TracingListener::~TracingListener() {
   // Purge deferred scopes.
   thread_pool_->Shutdown();
   thread_pool_->Wait();
@@ -49,9 +50,9 @@ Listener::~Listener() {
   global_tracing_listener = nullptr;
 }
 
-}  // namespace orbit_tracing
+}  // namespace orbit_base
 
-void Listener::DeferScopeProcessing(const Scope& scope) {
+void TracingListener::DeferScopeProcessing(const TracingScope& scope) {
   // User callback is called from a worker thread to
   // minimize contention on the instrumented threads.
   absl::MutexLock lock(&global_tracing_mutex);
@@ -65,8 +66,8 @@ void Listener::DeferScopeProcessing(const Scope& scope) {
 
 #ifdef ORBIT_API_INTERNAL_IMPL
 
-static std::vector<Scope>& GetThreadLocalScopes() {
-  thread_local std::vector<Scope> thread_local_scopes;
+static std::vector<TracingScope>& GetThreadLocalScopes() {
+  thread_local std::vector<TracingScope> thread_local_scopes;
   return thread_local_scopes;
 }
 
@@ -74,7 +75,7 @@ namespace orbit_api {
 
 void Start(const char* name, orbit::Color color) {
   GetThreadLocalScopes().emplace_back(
-      orbit_tracing::Scope(orbit_api::kScopeStart, name, /*data*/ 0, color));
+      TracingScope(orbit_api::kScopeStart, name, /*data*/ 0, color));
   auto& scope = GetThreadLocalScopes().back();
   scope.begin = MonotonicTimestampNs();
 }
@@ -84,45 +85,45 @@ void Stop() {
   scope.end = MonotonicTimestampNs();
   scope.depth = GetThreadLocalScopes().size() - 1;
   scope.tid = GetCurrentThreadId();
-  Listener::DeferScopeProcessing(scope);
+  TracingListener::DeferScopeProcessing(scope);
   GetThreadLocalScopes().pop_back();
 }
 
 void StartAsync(const char* name, uint64_t id, orbit::Color color) {
-  orbit_tracing::Scope scope(orbit_api::kScopeStartAsync, name, id, color);
+  TracingScope scope(orbit_api::kScopeStartAsync, name, id, color);
   scope.begin = MonotonicTimestampNs();
   scope.end = scope.begin;
   scope.tid = GetCurrentThreadId();
-  Listener::DeferScopeProcessing(scope);
+  TracingListener::DeferScopeProcessing(scope);
 }
 
 void StopAsync(uint64_t id) {
-  orbit_tracing::Scope scope(orbit_api::kScopeStopAsync, /*name*/ nullptr, id);
+  TracingScope scope(orbit_api::kScopeStopAsync, /*name*/ nullptr, id);
   scope.begin = MonotonicTimestampNs();
   scope.end = scope.begin;
   scope.tid = GetCurrentThreadId();
-  Listener::DeferScopeProcessing(scope);
+  TracingListener::DeferScopeProcessing(scope);
 }
 
 void AsyncString(const char* str, uint64_t id, orbit::Color color) {
   if (str == nullptr) return;
-  orbit_tracing::Scope scope(orbit_api::kString, /*name*/ nullptr, id, color);
+  TracingScope scope(orbit_api::kString, /*name*/ nullptr, id, color);
   auto& e = scope.encoded_event;
   constexpr size_t chunk_size = kMaxEventStringSize - 1;
   const char* end = str + strlen(str);
   while (str < end) {
     std::strncpy(e.event.name, str, chunk_size);
     e.event.name[chunk_size] = 0;
-    Listener::DeferScopeProcessing(scope);
+    TracingListener::DeferScopeProcessing(scope);
     str += chunk_size;
   }
 }
 
 void TrackValue(orbit_api::EventType type, const char* name, uint64_t value, orbit::Color color) {
-  orbit_tracing::Scope scope(type, name, value, color);
+  TracingScope scope(type, name, value, color);
   scope.begin = MonotonicTimestampNs();
   scope.tid = GetCurrentThreadId();
-  Listener::DeferScopeProcessing(scope);
+  TracingListener::DeferScopeProcessing(scope);
 }
 
 }  // namespace orbit_api
--- a/OrbitBase/Tracing.cpp
+++ b/OrbitBase/Tracing.cpp
@@ -11,14 +11,14 @@
 #include "OrbitBase/Profiling.h"
 #include "OrbitBase/ThreadPool.h"
 
-using orbit::tracing::Listener;
-using orbit::tracing::Scope;
-using orbit::tracing::TimerCallback;
+using orbit_tracing::Listener;
+using orbit_tracing::Scope;
+using orbit_tracing::TimerCallback;
 
 ABSL_CONST_INIT static absl::Mutex global_tracing_mutex(absl::kConstInit);
 ABSL_CONST_INIT static Listener* global_tracing_listener = nullptr;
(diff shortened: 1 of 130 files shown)


OrbitBase/include/OrbitBase/Tracing.h
```diff
@@ -16,7 +16,7 @@
 
 #define ORBIT_SCOPE_FUNCTION ORBIT_SCOPE(__FUNCTION__)
 
-namespace orbit::tracing {
+namespace orbit_tracing {
```
dpallotti, line 19: Hmm here's a tricky case, this is in `OrbitBase` but was following a namespace logic of its own. Should this be `orbit_base` or `orbit_tracing` now? Does anybody have thoughts?

OrbitClientData/FunctionUtils.cpp
```diff
@@ -17,7 +17,7 @@ uint64_t StringHash(const std::string& string) {
 }
 }  // namespace
 
-namespace FunctionUtils {
+namespace function_utils {
```
dpallotti, line 20: So as discussed this namespace should be removed in favor of `orbit_client_data` and the functions contained renamed to include a hint to `FunctionInfo`. Should we do this at a later time?

OrbitCore/CoreUtils.h
```diff
@@ -231,7 +231,7 @@ inline void fopen_s(FILE** fp, const char* fileName, const char* mode) {
 }
 #endif
 
-namespace OrbitUtils {
+namespace orbit_utils {
```
dpallotti, line 234: I would just make this `orbit_core`, too.

OrbitGl/ImGuiOrbit.h
```diff
@@ -20,13 +20,13 @@
...
-#define IMGUI_VAR_TO_TEXT(var) ImGuiOrbit::VariableToText(#var, var)
-#define IMGUI_VARN_TO_TEXT(var, name) ImGuiOrbit::VariableToText(name, var)
+#define IMGUI_VAR_TO_TEXT(var) orbit_gl::imgui::VariableToText(#var, var)
+#define IMGUI_VARN_TO_TEXT(var, name) orbit_gl::imgui::VariableToText(name, var)
 
 #define IMGUI_FLOAT_SLIDER(x) IMGUI_FLOAT_SLIDER_MIN_MAX(x, 0, 100.f)
 #define IMGUI_FLOAT_SLIDER_MIN_MAX(x, min, max) ImGui::SliderFloat(#x, &x, min, max)
 
-namespace ImGuiOrbit {
+namespace orbit_gl::imgui {
```
dpallotti, line 29: Seeing now that this namespace seems to contain only this one function, I would just keep `orbit_gl` only and if we really want to just rename the function to `VariableToImguiText`?

OrbitQt/orbitmainwindow.cpp
```diff
@@ -212,6 +212,14 @@ OrbitMainWindow::OrbitMainWindow(QApplication* a_App,
...
   });
 
+  GOrbitApp->SetTimerSelectedCallback([this](const orbit_client_protos::TimerInfo* timer_info) {
+    OnTimerSelectionChanged(timer_info);
+  });
+
+  GOrbitApp->SetTimerSelectedCallback([this](const orbit_client_protos::TimerInfo* timer_info) {
+    OnTimerSelectionChanged(timer_info);
+  });
+
```
dpallotti, line 222: Is this intended to be part of the change?

antonrohr (commented):
Thank you Radu! This was a pretty monotone code review, but I am glad this change is on the way. 

There are still some build errors on Linux (I assume you did this change on Windows). These should get caught be the CI, which for some reason did not run yet. This is the second time that I encounter that and the last time the simplest solution was to amend the commit and (force) push again. This should hopefully trigger the CI

dpallotti (commented):
Thanks a lot Radu!

Some comments on the special cases, maybe someone else has different or clearer ideas.

OrbitBase/include/OrbitBase/Tracing.h
dimitry-: If this is renamed to orbit_base we would probably want to rename Scope Listener and TimerCallback to TracingScope ...

radu-danciu: I guess it depends on a number of things: nesting policy, whether this merits a namespace or rather better to just remove and rename the structs, etc.

OrbitClientData/FunctionUtils.cpp
radu-danciu: I suggest so. There were two FunctionUtils, one of which was removed by somebody while I was doing this which required a rebase for me. I was thinking alignment would be a first step, followed by removal later - especially since I don't yet have enough context to carry out the renaming in a satisfactory way.

OrbitCore/CoreUtils.h
radu-danciu: I'm missing a lot of context :-/. I'm happy to do a second pass on these suggestions, but sadly I couldn't really arrive at them myself at this time.

OrbitQt/orbitmainwindow.cpp
radu-danciu: Ugh... no, this looks like a rebasing mistake.

radu-danciu:
> Thank you Radu! This was a pretty monotone code review, but I am glad this change is on the way.
> 
> There are still some build errors on Linux (I assume you did this change on Windows). These should get caught be the CI, which for some reason did not run yet. This is the second time that I encounter that and the last time the simplest solution was to amend the commit and (force) push again. This should hopefully trigger the CI

That's interesting. I did run it on Windows only, and on the first build after a rebase I needed to rebuild a module due to outdated protos, but the error that is now in the sponge is new to me. I'll wait to wrap a first pass on all comments and push an update.

OrbitBase/include/OrbitBase/Tracing.h
florian-kuebler: I'd say, that following the discussion we had this week, our policy is `orbit_base` and than rename `Scope` to `TracingScope`,..., as Dimitry suggested.

radu-danciu: Not having been included in that discussion, I'm not sure I'm currently the best person to do that, but I can try.

radu-danciu:
> > Thank you Radu! This was a pretty monotone code review, but I am glad this change is on the way.
> > There are still some build errors on Linux (I assume you did this change on Windows). These should get caught be the CI, which for some reason did not run yet. This is the second time that I encounter that and the last time the simplest solution was to amend the commit and (force) push again. This should hopefully trigger the CI
> 
> That's interesting. I did run it on Windows only, and on the first build after a rebase I needed to rebuild a module due to outdated protos, but the error that is now in the sponge is new to me. I'll wait to wrap a first pass on all comments and push an update.

Ugh, I see it now. The file with this error is not even included in the generated windows solution, so I never saw it... Maybe it should be in the solution but excluded from compilation?

OrbitBase/include/OrbitBase/Tracing.h
florian-kuebler: Basically, the idea would be to always use the module name (in snake case) as namespace and avoid nested namespaces. For free functions in the namespaces, ensure that the name is not ambiguous, and rather be more verbal on the name. I think it would be also fine to skip those cases if you are unsure.

Thanks for the change anyways.

IrinaShkviro:
Should the [OrbitVersion.cpp.in](https://github.com/google/orbit/blob/master/OrbitVersion/OrbitVersion.cpp.in) file be changed as well?

OrbitBase/include/OrbitBase/Tracing.h
dpallotti: So I think we agreed that we should use `orbit_base` as namespace and then make the changes `Scope` -> `TracingScope`, `Listener` -> `TracingListener`. I guess `TimerCallback` can be moved inside `TracingListener`.

radu-danciu: I have a new commit ready with this change, will update the PR in a bit.

OrbitCore/CoreUtils.h
radu-danciu: I have a new commit ready with this change, will update the PR in a bit.

OrbitGl/ImGuiOrbit.h
radu-danciu: Done.

OrbitQt/orbitmainwindow.cpp
radu-danciu: Removed.

radu-danciu:
> > > Thank you Radu! This was a pretty monotone code review, but I am glad this change is on the way.
> > > There are still some build errors on Linux (I assume you did this change on Windows). These should get caught be the CI, which for some reason did not run yet. This is the second time that I encounter that and the last time the simplest solution was to amend the commit and (force) push again. This should hopefully trigger the CI
> > 
> > 
> > That's interesting. I did run it on Windows only, and on the first build after a rebase I needed to rebuild a module due to outdated protos, but the error that is now in the sponge is new to me. I'll wait to wrap a first pass on all comments and push an update.
> 
> Ugh, I see it now. The file with this error is not even included in the generated windows solution, so I never saw it... Maybe it should be in the solution but excluded from compilation?

Fixed, hopefully all occurrences.

radu-danciu:
> Should the [OrbitVersion.cpp.in](https://github.com/google/orbit/blob/master/OrbitVersion/OrbitVersion.cpp.in) file be changed as well?

Done. Thanks for pointing this out.

OrbitBase/include/OrbitBase/Tracing.h
```diff
@@ -16,10 +16,8 @@
 
 #define ORBIT_SCOPE_FUNCTION ORBIT_SCOPE(__FUNCTION__)
 
-namespace orbit::tracing {
```
dpallotti, line 19: (I know you're already fixing this removal.)

OrbitCore/CoreUtils.h
```diff
@@ -315,6 +315,6 @@ std::vector<std::pair<Key, Val> > ReverseValueSort(std::map<Key, Val>& a_Map) {
 }
 
 std::string FormatTime(const time_t& rawtime);
-}  // namespace OrbitUtils
+}  // namespace orbit_utils
```
dpallotti, line 318:
```suggestion
}  // namespace orbit_core
```

dimitry- (approved):

OrbitCore/CoreUtils.h
radu-danciu: Damn.

OrbitBase/Tracing.cpp
```diff
@@ -84,45 +84,45 @@ void Stop() {
...
   scope.depth = GetThreadLocalScopes().size() - 1;
   scope.tid = GetCurrentThreadId();
-  Listener::DeferScopeProcessing(scope);
+  TracingListener::DeferScopeProcessing(scope);
   GetThreadLocalScopes().pop_back();
 }
 
 void StartAsync(const char* name, uint64_t id, orbit::Color color) {
-  orbit::tracing::Scope scope(orbit_api::kScopeStartAsync, name, id, color);
+  orbit_base::TracingScope scope(orbit_api::kScopeStartAsync, name, id, color);
```
dpallotti, line 92: Same here, I guess you don't need `orbit_base::`.

OrbitBase/Tracing.cpp
```diff
@@ -84,45 +84,45 @@ void Stop() {
...
   scope.begin = MonotonicTimestampNs();
   scope.end = scope.begin;
   scope.tid = GetCurrentThreadId();
-  Listener::DeferScopeProcessing(scope);
+  TracingListener::DeferScopeProcessing(scope);
 }
 
 void StopAsync(uint64_t id) {
-  orbit::tracing::Scope scope(orbit_api::kScopeStopAsync, /*name*/ nullptr, id);
+  orbit_base::TracingScope scope(orbit_api::kScopeStopAsync, /*name*/ nullptr, id);
```
dpallotti, line 100: Also here.

OrbitBase/Tracing.cpp
```diff
@@ -84,45 +84,45 @@ void Stop() {
...
   scope.end = scope.begin;
   scope.tid = GetCurrentThreadId();
-  Listener::DeferScopeProcessing(scope);
+  TracingListener::DeferScopeProcessing(scope);
 }
 
 void AsyncString(const char* str, uint64_t id, orbit::Color color) {
   if (str == nullptr) return;
-  orbit::tracing::Scope scope(orbit_api::kString, /*name*/ nullptr, id, color);
+  orbit_base::TracingScope scope(orbit_api::kString, /*name*/ nullptr, id, color);
```
dpallotti, line 109: Also here.

OrbitBase/Tracing.cpp
```diff
@@ -84,45 +84,45 @@ void Stop() {
...
     e.event.name[chunk_size] = 0;
-    Listener::DeferScopeProcessing(scope);
+    TracingListener::DeferScopeProcessing(scope);
     str += chunk_size;
   }
 }
 
 void TrackValue(orbit_api::EventType type, const char* name, uint64_t value, orbit::Color color) {
-  orbit::tracing::Scope scope(type, name, value, color);
+  orbit_base::TracingScope scope(type, name, value, color);
```
dpallotti, line 122: Also here.

OrbitBase/include/OrbitBase/Tracing.h
```diff
@@ -28,22 +28,22 @@ struct Scope {
...
 
  private:
-  TimerCallback user_callback_ = nullptr;
+  TracingTimerCallback user_callback_ = nullptr;
   std::unique_ptr<ThreadPool> thread_pool_ = {};
   inline static bool active_ = false;
 };
 
-}  // namespace orbit::tracing
+}  // namespace orbit_tracing
```
dpallotti, line 47:
```suggestion
}  // namespace orbit_base
```

dpallotti (approved):
Thanks again, Radu! I left a couple of comments on minor things that could be cleaned up.

PR #1563, opened 2020-12-08
Fix compilation when ORBIT_API_ENABLED is set to 0

pierricgimmig:
The ORBIT_API_ENABLED macro can be toggled on/off by external users to
disable manual instrumentation completely. In Orbit, introspection uses
the same manual instrumentation API for self profiling. Make sure we can
also use the same ORBIT_API_ENABLED macro to disable introspection.

The change consists of leaving some of the class declarations available 
even though ORBIT_API_ENABLED is set to 0 instead of #ifdefing the 
introspection code itself which is spread out in various locations. 
When ORBIT_API_ENABLED is set to 0, all the orbit api macros are defined
to "nothing" which means that we are not emitting any introspection data.

b/174565006


beckerhe (approved):
Looks good to me, but Davide should have a look as well!

Thanks Pierric!

dpallotti (approved):
Thanks for taking care of this!

PR #1564, opened 2020-12-09
Fix generate_coverage_report.sh

antonrohr:
The script to generate unit test coverage reports was broken since
ctest was used differently. This commit fixes this and coverage reports
can now be generated again.

Test: 
third_party/conan/configs/install.sh &&
./build.sh coverage_clang9 &&
contrib/unit_test_coverage/generate_coverage_report.sh . build_coverage_clang9/


beckerhe (approved):
Thanks for fixing this! (And sorry for breaking it!)

PR #1566, opened 2020-12-09
Update callstack view when switching between thread tabs

vickyliu-go4it:
Make the callstack view show consistent selections when the user selects in multiple thread tabs and switches tabs.

Current callstack widget is not refreshed when the user switch thread tabs. Therefore, it is not clear which selection is shown if the user selects in multiple thread tabs. We fix this inconsistency by keeping the selection for each thread tab and update the callstack view accordingly when we switch thread tabs.

Fixes the bug b/170468638.

### DIFF
--- a/OrbitGl/LiveFunctionsDataView.cpp
+++ b/OrbitGl/LiveFunctionsDataView.cpp
@@ -77,10 +77,14 @@ std::string LiveFunctionsDataView::GetValue(int row, int column) {
   }
 }
 
-void LiveFunctionsDataView::OnSelect(int row) {
+void LiveFunctionsDataView::OnSelect(std::optional<int> row) {
+  if (!row.has_value()) {
+    return;
+  }
   GOrbitApp->DeselectTextBox();
   const CaptureData& capture_data = GOrbitApp->GetCaptureData();
-  GOrbitApp->set_highlighted_function(capture_data.GetAbsoluteAddress(*GetSelectedFunction(row)));
+  GOrbitApp->set_highlighted_function(
+      capture_data.GetAbsoluteAddress(*GetSelectedFunction(row.value())));
 }
 
 #define ORBIT_FUNC_SORT(Member)                                                      \
--- a/OrbitGl/ProcessesDataView.cpp
+++ b/OrbitGl/ProcessesDataView.cpp
@@ -81,8 +81,11 @@ void ProcessesDataView::DoSort() {
 
 #undef ORBIT_PROC_SORT
 
-void ProcessesDataView::OnSelect(int index) {
-  const ProcessInfo& selected_process = GetProcess(index);
+void ProcessesDataView::OnSelect(std::optional<int> index) {
+  if (!index.has_value()) {
+    return;
+  }
+  const ProcessInfo& selected_process = GetProcess(index.value());
   selected_process_id_ = selected_process.pid();
 
   SetSelectedItem();
--- a/OrbitGl/SamplingReport.cpp
+++ b/OrbitGl/SamplingReport.cpp
@@ -14,7 +14,7 @@ SamplingReport::SamplingReport(
     : post_processed_sampling_data_{std::move(post_processed_sampling_data)},
       unique_callstacks_{std::move(unique_callstacks)},
       has_summary_{has_summary} {
-  selected_address_ = 0;
+  selected_address_ = kInvalidFunctionAddress;
   selected_thread_id_ = 0;
   callstack_data_view_ = nullptr;
   selected_sorted_callstack_report_ = nullptr;
@@ -43,6 +43,11 @@ void SamplingReport::FillReport() {
 }
 
 void SamplingReport::UpdateDisplayedCallstack() {
+  if (selected_address_ == SamplingReport::kInvalidFunctionAddress) {
+    ClearReport();
+    return;
+  }
+
   selected_sorted_callstack_report_ =
       post_processed_sampling_data_.GetSortedCallstackReportFromAddress(selected_address_,
                                                                         selected_thread_id_);
--- a/OrbitGl/SamplingReport.h
+++ b/OrbitGl/SamplingReport.h
@@ -35,6 +35,7 @@ class SamplingReport {
   [[nodiscard]] bool HasSamples() const { return !unique_callstacks_.empty(); }
   [[nodiscard]] bool has_summary() const { return has_summary_; }
   void ClearReport();
+  static const uint64_t kInvalidFunctionAddress = std::numeric_limits<uint64_t>::max();
 
  protected:
   void FillReport();
--- a/OrbitQt/orbitsamplingreport.cpp
+++ b/OrbitQt/orbitsamplingreport.cpp
@@ -92,7 +92,10 @@ void OrbitSamplingReport::on_PreviousCallstackButton_clicked() {
 void OrbitSamplingReport::OnCurrentThreadTabChanged(int current_tab_index) {
   auto treeView = m_OrbitDataViews[current_tab_index];
   QModelIndexList index_list = treeView->GetTreeView()->selectionModel()->selectedIndexes();
-  int row = (!index_list.isEmpty() && index_list.front().isValid()) ? index_list.front().row() : -1;
+  std::optional<int> row = std::nullopt;
+  if (!index_list.isEmpty() && index_list.front().isValid()) {
+    row = index_list.front().row();
+  }
   treeView->GetTreeView()->GetModel()->OnRowSelected(row);
   RefreshCallstackView();
 }
--- a/OrbitQt/orbitsamplingreport.cpp
+++ b/OrbitQt/orbitsamplingreport.cpp
@@ -72,6 +72,9 @@ void OrbitSamplingReport::Initialize(DataView* callstack_data_view,
     QString threadName = QString::fromStdString(report_data_view.GetName());
     ui->tabWidget->addTab(tab, threadName);
   }
+
+  connect(ui->tabWidget, &QTabWidget::currentChanged, this,
+          &OrbitSamplingReport::OnCurrentThreadTabChanged);
 }
 
 void OrbitSamplingReport::on_NextCallstackButton_clicked() {
@@ -83,6 +86,14 @@ void OrbitSamplingReport::on_NextCallstackButton_clicked() {
 void OrbitSamplingReport::on_PreviousCallstackButton_clicked() {
   CHECK(m_SamplingReport != nullptr);
   m_SamplingReport->DecrementCallstackIndex();
+  RefreshCallstackView();
+}
+
+void OrbitSamplingReport::OnCurrentThreadTabChanged(int current_tab_index) {
+  auto treeView = m_OrbitDataViews[current_tab_index];
+  QModelIndexList index_list = treeView->GetTreeView()->selectionModel()->selectedIndexes();
+  int row = (!index_list.isEmpty() && index_list.front().isValid()) ? index_list.front().row() : -1;
+  treeView->GetTreeView()->GetModel()->OnRowSelected(row);
   RefreshCallstackView();
 }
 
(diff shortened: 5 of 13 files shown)


OrbitQt/orbitsamplingreport.cpp
```diff
@@ -86,6 +89,21 @@ void OrbitSamplingReport::on_PreviousCallstackButton_clicked() {
   RefreshCallstackView();
 }
 
+void OrbitSamplingReport::OnCurrentThreadTabChanged(int current_tab_index) {
+  if (current_tab_index == -1) {
```
reichlfl, line 93: When can this case happen? Looks like this should also clear the selection?

OrbitQt/orbitsamplingreport.cpp
```diff
@@ -86,6 +89,21 @@ void OrbitSamplingReport::on_PreviousCallstackButton_clicked() {
...
+    return;
+  }
+
+  auto treeView = m_OrbitDataViews.at(current_tab_index);
```
reichlfl, line 97: Nit: Is there a particular reason for using `.at(current_tab_index)` instead of `[current_tab_index]`?

OrbitQt/orbitsamplingreport.cpp
```diff
@@ -86,6 +89,21 @@ void OrbitSamplingReport::on_PreviousCallstackButton_clicked() {
...
+  QModelIndexList index_list = treeView->GetTreeView()->selectionModel()->selectedIndexes();
+  if (!index_list.isEmpty() && index_list.front().isValid()) {
+    treeView->GetTreeView()->GetModel()->OnRowSelected(index_list.front().row());
+  } else {
+    m_SamplingReport->ClearReport();
```
reichlfl, line 102: Is this correct? Shouldn't it just clear the callstack selection, not the complete sampling report?

OrbitGl/SamplingReport.cpp
```diff
@@ -80,8 +80,8 @@ void SamplingReport::OnSelectAddress(uint64_t address, ThreadID thread_id) {
     if (selected_address_ != address || selected_thread_id_ != thread_id) {
       selected_address_ = address;
       selected_thread_id_ = thread_id;
-      UpdateDisplayedCallstack();
```
karupayun, line 83: I don't understand this, suppose that we are doing for the case that the user switch between tabs. But doesn't have more sense instead of updating this every time, to really update the thread_id when we move to another tab?

karupayun (commented):
Nice to see that you are tackling this inconsistency!

I've left a comment and I have a few questions/future work.

1- Now that we are saving rows for every tab, and update the callstack accordingly, doesn't have sense to also save the number of callstack that the user was currently seeing? So, for example, I imagine a use case where an user is comparing two call stacks between different threads, it would be good that when we switch tabs, the callstack's number is saved. Currently if you move to another tab, it is also possible that you lose the number that you were before.
2- Similar to the sliders position, but probably too less important.
3- For a future PR. Could be really interesting when navigating for "all threads", having a way to reference the thread id where this callstack is happening, so the user could analize that thread directly.

OrbitQt/orbitsamplingreport.cpp
vickyliu-go4it: This happens when the user click on the same thread tab. 

To be more specific, we receive the current_tab_index from the signal currentChanged every time when we select a thread tab. If the selected tab is the same with previous one, current_tab_index is -1; otherwise, current_tab_index illustrates the page index of selected new thread tab.

vickyliu-go4it: But it might be a good idea that we consider the user can click on the same thread tab to clear the previous selection. This might be helpful when for the multiple selection case.

OrbitGl/SamplingReport.cpp
vickyliu-go4it: I'm happy that you point out this issue. It also bothers me for a while.

So for example, if we have three thread tabs: tab 1, tab 2 and tab 3. Both tab 1 and 2 have selections but tab 3 don't.
If we click the thread tab following the order: tab 2 --> tab 3 --> tab 2, the UpdateDisplayedCallstack() will not be called if we don't move it out of the if loop.

The reason is that, when we switch to a new thread tab without making any selection, the selected_address_ and selected_thread_id_ are not updated, so if we switch back to the last tab which has selections, the selected_thread_id and  selected_address_ are still the same so the callstack widget will not be updated.

OrbitQt/orbitsamplingreport.cpp
vickyliu-go4it: Yes, I agree with you. It is impossible to be out of range here, so `[current_tab_index]` would be more efficient. Will change it accordingly.

OrbitGl/SamplingReport.cpp
karupayun: Yes, I supposed that when I erased that line and playing a little with tabs.

But in my opinion doesn't have any sense being in a tab with an outdated reference to selected_thread_id_. This patch could work now, but could generate another issue later. So,

1- We can add at least one comment explaining the necessity of having this line out of that block.
2- Or better again, trying to update the thread_id when we move to another tab and not only when we select a row.

OrbitQt/orbitsamplingreport.cpp
vickyliu-go4it: Thanks for raising this problem. The name might be misleading. `m_SamplingReport->ClearReport()` only clears the callstack report rather than the whole sampling report.

OrbitGl/SamplingReport.cpp
```diff
@@ -43,6 +43,11 @@ void SamplingReport::FillReport() {
 }
 
 void SamplingReport::UpdateDisplayedCallstack() {
+  if (selected_address_ == SamplingReport::kInvalidFunctionAddress) {
+    ClearReport();
```
reichlfl, line 47: Ah... `ClearReport()` only resets the selected callstack, it doesn't clear the complete report? The name is very misleading and I completely misread that, but then this makes sense.

OrbitQt/orbitsamplingreport.cpp
reichlfl: Are you sure about the -1 being sent when the same tab is clicked? That would be super ugly. The way I read the documentation, -1 is passed in exceptional cases, e.g. when the TabWidget does not have any tabs that could be selected. If the same tab is clicked, this signal shouldn't be emitted at all.

Which means clearing the selection would be the right thing to do for -1, even though we do not expect this to happen.

vickyliu-go4it: I checked the qt5 document for [QTabWidget::currentChanged](https://doc.qt.io/qt-5/qtabwidget.html#currentChanged). And I tested on the ui and found that if I click on the same thread tab, current_tab_index is -1.

OrbitGl/SamplingReport.cpp
vickyliu-go4it: Thanks for the suggestion. Now I keep both the selected function address and thread id to be updated: the selected_address is set as kInvalidFunctionAddress to indicate there is no selections; and also the thread id is always the current selected one.

OrbitGl/SamplingReport.cpp
vickyliu-go4it: I change to select function address as kInvalidFunctionAddress to indicate no function is selected when moving to a thread tab without selections. So the action of clearing the callstack report is move from the orbitsamplingreport.cpp to the SamplingReport.cpp, which is closer to the definition of ClearReport. Hope it would be less misleading.

OrbitQt/orbitsamplingreport.cpp
```diff
@@ -94,13 +94,10 @@ void OrbitSamplingReport::OnCurrentThreadTabChanged(int current_tab_index) {
...
 
-  auto treeView = m_OrbitDataViews.at(current_tab_index);
+  auto treeView = m_OrbitDataViews[current_tab_index];
   QModelIndexList index_list = treeView->GetTreeView()->selectionModel()->selectedIndexes();
-  if (!index_list.isEmpty() && index_list.front().isValid()) {
-    treeView->GetTreeView()->GetModel()->OnRowSelected(index_list.front().row());
-  } else {
-    m_SamplingReport->ClearReport();
-  }
+  int row = (!index_list.isEmpty() && index_list.front().isValid()) ? index_list.front().row() : -1;
```
karupayun, line 95: I don't like OnRowSelected being executed with this -1.
I was thinking that we can use std::optional here. So, you give an optional row to OnRowSelected and you don't need this -1 constant here, neither in "OnSelect" function.

Also, if you want, function_address could be a std::optional.

karupayun (approved):
Thanks for addressing my comments!

I'm approving it, but leaving 2 comments :)

OrbitGl/LiveFunctionsDataView.cpp
```diff
@@ -77,10 +77,12 @@ std::string LiveFunctionsDataView::GetValue(int row, int column) {
   }
 }
 
-void LiveFunctionsDataView::OnSelect(int row) {
+void LiveFunctionsDataView::OnSelect(std::optional<int> row) {
+  CHECK(row.has_value());
```
reichlfl, line 81: This looks dangerous, I think the case `!row.has_value()` must be handled. It should be possible to not have any function selected, wasn't this handled before? 

Right now, wouldn't this happen if you hook a function, start a capture, and then unhook that function again and restart?

OrbitGl/ProcessesDataView.cpp
```diff
@@ -81,8 +81,9 @@ void ProcessesDataView::DoSort() {
 
 #undef ORBIT_PROC_SORT
 
-void ProcessesDataView::OnSelect(int index) {
-  const ProcessInfo& selected_process = GetProcess(index);
+void ProcessesDataView::OnSelect(std::optional<int> index) {
+  CHECK(index.has_value());
```
reichlfl, line 85: Same comment as for the LiveFunctionsView, I think this must be handled more gracefully. This is a case that could happen in some circumstances, and the CHECK will crash Orbit.

reichlfl (commented):
I like the change with std::optional<>, but please take a look at the two checks I commented.

OrbitGl/LiveFunctionsDataView.cpp
vickyliu-go4it: Thanks for raising this issue. That's really a good question I should consider.

In the current version, we shouldn't reach the case that `!row.has_value()` for the LiveFunctionsDataView and the ProcessDataView. The reasons are the following:
- If it is an external selection (i.e., the selection made by the users when clicking rows in the data view panel), we always pass `currentIndex().rows()` to `OnSelect()` when `currentIndex().rows()` is valid. Therefore, for the external selections, we always pass a valid row to `OnRowSelected()` and `OnSelect()`.  
- But we do pass `std::nullopt` to `OnRowSelected()` and `OnSelect()` in **some internal selection cases**, which makes us able to clear the selection in the data view and process the no selection case. But we only do this for the `SamplingReportDataView` but not for the other two data views.

But you are right, this CHECK might cause crashes in the future. I will try to find a better way to handle this.

OrbitGl/SamplingReport.h
```diff
@@ -47,7 +48,7 @@ class SamplingReport {
   std::vector<SamplingReportDataView> thread_reports_;
   CallStackDataView* callstack_data_view_;
 
-  uint64_t selected_address_;
+  uint64_t selected_address_ = kInvalidFunctionAddress;
```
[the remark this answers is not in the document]
vickyliu-go4it: Yes, totally agree.

OrbitQt/orbitsamplingreport.cpp
vickyliu-go4it: Thanks so much for bringing that to my eyes! I like std::optional!

reichlfl (approved):
As discussed offline, I think we're not really handling the cases of a user clicking in the empty areas of tableviews correctly, this should clear the selection. But the behavior is consistent, so in the scope of this PR it LGTM, thanks for the change!

reichlfl:
> I like the change with std::optional<>, but please take a look at the two checks I commented.
> Edit: BTW, this may conflict with #1526, it's probably better to wait for Antons PR to be finished before merging this.

@vickyliu-go4it It's fine to merge this before #1526, so just go ahead.

PR #1567, opened 2020-12-09
[Refactor] Move Tracks out of Timegraph

karupayun:
Splitting TimeGraph by the creation of TrackManager, who will be on charge of everything related to the tracks (creation, searching, erasing and sorting). For now, this class have a reference to TimeGraph but in the future we want to erase it. We want to make unit tests on the tracks, which will be better if these test doesn't need to know about TimeGraph.

Future work appeared during the refactor:

**Next Steps in my roadmap:**
- Eliminate the TimeGraph dependence in TrackManager and in Tracks (b/176086740).
- Try to erase StringManager in TrackManager and TimeGraph 
- TrackManager should manage sorting by their own (b/176077097)

**Possible bugs:**
- Use FunctionInfo instead of only their address as key to FrameTracks (b/175865913)
- TimeGraph should not hold and expose CaptureData (b/176056427)

**Better design:**
- Improve the way to store, create and get the Tracks (b/175869409)
- Having tracks unique pointers instead of shared_ptr in TrackManager whenever possible.

### DIFF
--- a/OrbitGl/App.cpp
+++ b/OrbitGl/App.cpp
@@ -541,7 +541,8 @@ void OrbitApp::MainTick() {
   GMainTimer.Restart();
 
   if (DoZoom && HasCaptureData()) {
-    GCurrentTimeGraph->SortTracks();
+    // TODO (b/176077097): TrackManager has to manage sorting by their own.
+    GCurrentTimeGraph->GetTrackManager()->SortTracks();
     capture_window_->ZoomAll();
     NeedsRedraw();
     DoZoom = false;
@@ -768,7 +769,8 @@ PresetLoadState OrbitApp::GetPresetLoadState(
 }
 
 ErrorMessageOr<void> OrbitApp::OnSaveCapture(const std::string& file_name) {
-  const auto& key_to_string_map = GCurrentTimeGraph->GetStringManager()->GetKeyToStringMap();
+  const auto& key_to_string_map =
+      GCurrentTimeGraph->GetTrackManager()->GetStringManager()->GetKeyToStringMap();
 
   std::vector<std::shared_ptr<TimerChain>> chains =
       GCurrentTimeGraph->GetAllSerializableTimerChains();
--- a/OrbitGl/FrameTrack.cpp
+++ b/OrbitGl/FrameTrack.cpp
@@ -246,7 +246,7 @@ void FrameTrack::UpdateBoxHeight() {
   box_height_ = kBoxHeightMultiplier * time_graph_->GetLayout().GetTextBoxHeight();
 }
 
-std::vector<std::shared_ptr<TimerChain>> FrameTrack::GetAllSerializableChains() {
+std::vector<std::shared_ptr<TimerChain>> FrameTrack::GetAllSerializableChains() const {
   // Frametracks are just displaying existing data in a different way.
   // We don't want to write out all the timers of that track.
   // TODO(b/171026228): However, we should serialize them in some form.
--- a/OrbitGl/FrameTrack.cpp
+++ b/OrbitGl/FrameTrack.cpp
@@ -4,7 +4,16 @@
 
 #include "FrameTrack.h"
 
-#include "App.h"
+#include <GteVector.h>
+#include <absl/strings/str_format.h>
+#include <absl/time/time.h>
+
+#include <algorithm>
+
+#include "Batcher.h"
+#include "CoreUtils.h"
+#include "GlCanvas.h"
+#include "GlUtils.h"
 #include "OrbitClientData/FunctionUtils.h"
 #include "TimeGraph.h"
 #include "TimeGraphLayout.h"
--- a/OrbitGl/FrameTrack.h
+++ b/OrbitGl/FrameTrack.h
@@ -14,6 +14,7 @@ class FrameTrack : public TimerTrack {
   explicit FrameTrack(TimeGraph* time_graph, const orbit_client_protos::FunctionInfo& function,
                       OrbitApp* app);
   [[nodiscard]] Type GetType() const override { return kFrameTrack; }
+  [[nodiscard]] uint64_t GetFunctionAddress() const { return function_.address(); }
   [[nodiscard]] bool IsCollapsable() const override { return GetMaximumScaleFactor() > 0.f; }
 
   [[nodiscard]] virtual float GetYFromDepth(uint32_t depth) const override;
@@ -32,7 +33,7 @@ class FrameTrack : public TimerTrack {
 
   void UpdateBoxHeight() override;
 
-  [[nodiscard]] std::vector<std::shared_ptr<TimerChain>> GetAllSerializableChains() override;
+  [[nodiscard]] std::vector<std::shared_ptr<TimerChain>> GetAllSerializableChains() const override;
 
  protected:
   [[nodiscard]] Color GetTimerColor(const orbit_client_protos::TimerInfo& timer_info,
--- a/OrbitGl/FrameTrack.h
+++ b/OrbitGl/FrameTrack.h
@@ -5,7 +5,19 @@
 #ifndef ORBIT_GL_FRAME_TRACK_H_
 #define ORBIT_GL_FRAME_TRACK_H_
 
+#include <stdint.h>
+
+#include <memory>
+#include <string>
+#include <vector>
+
+#include "CoreMath.h"
+#include "PickingManager.h"
+#include "TextBox.h"
+#include "TimerChain.h"
 #include "TimerTrack.h"
+#include "Track.h"
+#include "capture_data.pb.h"
 
 class OrbitApp;
 
--- a/OrbitGl/GpuTrack.cpp
+++ b/OrbitGl/GpuTrack.cpp
@@ -38,8 +38,8 @@ std::string MapGpuTimelineToTrackLabel(std::string_view timeline) {
 
 }  // namespace orbit_gl
 
-GpuTrack::GpuTrack(TimeGraph* time_graph, std::shared_ptr<StringManager> string_manager,
-                   uint64_t timeline_hash, OrbitApp* app)
+GpuTrack::GpuTrack(TimeGraph* time_graph, StringManager* string_manager, uint64_t timeline_hash,
+                   OrbitApp* app)
     : TimerTrack(time_graph, app) {
   text_renderer_ = time_graph->GetTextRenderer();
   timeline_hash_ = timeline_hash;
@@ -129,8 +129,7 @@ void GpuTrack::SetTimesliceText(const TimerInfo& timer_info, double elapsed_us, 
     CHECK(timer_info.type() == TimerInfo::kGpuActivity);
 
     std::string text = absl::StrFormat(
-        "%s  %s", time_graph_->GetStringManager()->Get(timer_info.user_data_key()).value_or(""),
-        time.c_str());
+        "%s  %s", string_manager_->Get(timer_info.user_data_key()).value_or(""), time.c_str());
     text_box->SetText(text);
   }
 
--- a/OrbitGl/TimeGraph.cpp
+++ b/OrbitGl/TimeGraph.cpp
@@ -36,9 +36,7 @@ TimeGraph* GCurrentTimeGraph = nullptr;
 
 TimeGraph::TimeGraph(uint32_t font_size, OrbitApp* app)
     : font_size_(font_size), batcher_(BatcherId::kTimeGraph), app_{app} {
-  scheduler_track_ = GetOrCreateSchedulerTrack();
-
-  tracepoints_system_wide_track_ = GetOrCreateThreadTrack(orbit_base::kAllThreadsOfAllProcessesTid);
+  track_manager_ = std::make_unique<TrackManager>(this, app);
 
   async_timer_info_listener_ =
       std::make_unique<ManualInstrumentationManager::AsyncTimerInfoListener>(
@@ -55,7 +53,7 @@ TimeGraph::~TimeGraph() {
 }
 
 void TimeGraph::SetStringManager(std::shared_ptr<StringManager> str_manager) {
-  string_manager_ = std::move(str_manager);
+  track_manager_->SetStringManager(str_manager.get());
 }
 
 void TimeGraph::SetCanvas(GlCanvas* canvas) {
@@ -73,20 +71,7 @@ void TimeGraph::Clear() {
   capture_max_timestamp_ = 0;
   thread_count_map_.clear();
 
-  tracks_.clear();
-  scheduler_track_ = nullptr;
-  thread_tracks_.clear();
-  gpu_tracks_.clear();
-  graph_tracks_.clear();
-  async_tracks_.clear();
-  frame_tracks_.clear();
-
-  sorted_tracks_.clear();
-  sorted_filtered_tracks_.clear();
-
-  scheduler_track_ = GetOrCreateSchedulerTrack();
-
-  tracepoints_system_wide_track_ = GetOrCreateThreadTrack(orbit_base::kAllThreadsOfAllProcessesTid);
+  track_manager_->Clear();
 
   SetIteratorOverlayData({}, {});
 
@@ -97,7 +82,7 @@ double GNumHistorySeconds = 2.f;
 
 void TimeGraph::UpdateCaptureMinMaxTimestamps() {
   std::lock_guard<std::recursive_mutex> lock(mutex_);
-  for (auto& track : tracks_) {
+  for (auto& track : track_manager_->GetTracks()) {
     if (!track->IsEmpty()) {
       capture_min_timestamp_ = std::min(capture_min_timestamp_, track->GetMinTime());
       capture_max_timestamp_ = std::max(capture_max_timestamp_, track->GetMaxTime());
@@ -227,7 +212,7 @@ void TimeGraph::HorizontallyMoveIntoView(VisibilityType vis_type, const TimerInf
 }
 
 void TimeGraph::VerticallyMoveIntoView(const TimerInfo& timer_info) {
-  VerticallyMoveIntoView(*GetOrCreateThreadTrack(timer_info.thread_id()));
+  VerticallyMoveIntoView(*track_manager_->GetOrCreateThreadTrack(timer_info.thread_id()));
 }
 
 void TimeGraph::VerticallyMoveIntoView(Track& track) {
@@ -262,27 +247,31 @@ void TimeGraph::ProcessTimer(const TimerInfo& timer_info, const FunctionInfo* fu
     ProcessOrbitFunctionTimer(function->orbit_type(), timer_info);
   }
 
+  // TODO (b/175869409): Change the way to create and get the tracks. Move this part to
+  // TrackManager.
   if (timer_info.type() == TimerInfo::kGpuActivity) {
     uint64_t timeline_hash = timer_info.timeline_hash();
-    std::shared_ptr<GpuTrack> track = GetOrCreateGpuTrack(timeline_hash);
+    std::shared_ptr<GpuTrack> track = track_manager_->GetOrCreateGpuTrack(timeline_hash);
     track->OnTimer(timer_info);
   } else if (timer_info.type() == TimerInfo::kFrame) {
-    std::shared_ptr<FrameTrack> track = GetOrCreateFrameTrack(*function);
+    std::shared_ptr<FrameTrack> track = track_manager_->GetOrCreateFrameTrack(*function);
     track->OnTimer(timer_info);
   } else if (timer_info.type() == TimerInfo::kIntrospection) {
     ProcessIntrospectionTimer(timer_info);
   } else {
-    std::shared_ptr<ThreadTrack> track = GetOrCreateThreadTrack(timer_info.thread_id());
+    std::shared_ptr<ThreadTrack> track =
+        track_manager_->GetOrCreateThreadTrack(timer_info.thread_id());
     if (timer_info.type() != TimerInfo::kCoreActivity) {
       track->OnTimer(timer_info);
       ++thread_count_map_[timer_info.thread_id()];
     } else {
-      scheduler_track_->OnTimer(timer_info);
+      auto scheduler_track = track_manager_->GetOrCreateSchedulerTrack();
+      scheduler_track->OnTimer(timer_info);
       if (GetNumCores() <= static_cast<uint32_t>(timer_info.processor())) {
         auto num_cores = timer_info.processor() + 1;
         SetNumCores(num_cores);
         layout_.SetNumCores(num_cores);
-        scheduler_track_->SetLabel(absl::StrFormat("Scheduler (%u cores)", num_cores));
+        scheduler_track->SetLabel(absl::StrFormat("Scheduler (%u cores)", num_cores));
       }
     }
   }
@@ -310,7 +299,8 @@ void TimeGraph::ProcessIntrospectionTimer(const TimerInfo& timer_info) {
 
   switch (event.type) {
     case orbit_api::kScopeStart: {
-      std::shared_ptr<ThreadTrack> track = GetOrCreateThreadTrack(timer_info.thread_id());
+      std::shared_ptr<ThreadTrack> track =
+          track_manager_->GetOrCreateThreadTrack(timer_info.thread_id());
       track->OnTimer(timer_info);
       ++thread_count_map_[timer_info.thread_id()];
     } break;
@@ -340,7 +330,7 @@ void TimeGraph::ProcessValueTrackingTimer(const TimerInfo& timer_info) {
     return;
   }
 
-  auto track = GetOrCreateGraphTrack(event.name);
+  GraphTrack* track = track_manager_->GetOrCreateGraphTrack(event.name);
   uint64_t time = timer_info.start();
 
   switch (event.type) {
@@ -373,19 +363,19 @@ void TimeGraph::ProcessValueTrackingTimer(const TimerInfo& timer_info) {
 }
 
 void TimeGraph::ProcessAsyncTimer(const std::string& track_name, const TimerInfo& timer_info) {
-  auto track = GetOrCreateAsyncTrack(track_name);
+  AsyncTrack* track = track_manager_->GetOrCreateAsyncTrack(track_name);
   track->OnTimer(timer_info);
 }
 
 uint32_t TimeGraph::GetNumTimers() const {
   uint32_t num_timers = 0;
   std::lock_guard<std::recursive_mutex> lock(mutex_);
-  for (const auto& track : tracks_) {
+  for (const auto& track : track_manager_->GetTracks()) {
     num_timers += track->GetNumTimers();
   }
   // Frame tracks are removable by users and cannot simply be thrown into the
   // tracks_ vector.
-  for (const auto& [unused_key, track] : frame_tracks_) {
+  for (const auto& track : track_manager_->GetFrameTracks()) {
     num_timers += track->GetNumTimers();
   }
   return num_timers;
@@ -393,12 +383,12 @@ uint32_t TimeGraph::GetNumTimers() const {
 
 std::vector<std::shared_ptr<TimerChain>> TimeGraph::GetAllTimerChains() const {
   std::vector<std::shared_ptr<TimerChain>> chains;
-  for (const auto& track : tracks_) {
+  for (const auto& track : track_manager_->GetTracks()) {
     Append(chains, track->GetAllChains());
   }
   // Frame tracks are removable by users and cannot simply be thrown into the
   // tracks_ vector.
-  for (const auto& [unused_key, track] : frame_tracks_) {
+  for (const auto& track : track_manager_->GetFrameTracks()) {
     Append(chains, track->GetAllChains());
   }
   return chains;
@@ -406,7 +396,7 @@ std::vector<std::shared_ptr<TimerChain>> TimeGraph::GetAllTimerChains() const {
 
 std::vector<std::shared_ptr<TimerChain>> TimeGraph::GetAllThreadTrackTimerChains() const {
   std::vector<std::shared_ptr<TimerChain>> chains;
-  for (const auto& [_, track] : thread_tracks_) {
+  for (const auto& track : track_manager_->GetThreadTracks()) {
     Append(chains, track->GetAllChains());
   }
   return chains;
@@ -414,7 +404,7 @@ std::vector<std::shared_ptr<TimerChain>> TimeGraph::GetAllThreadTrackTimerChains
 
 std::vector<std::shared_ptr<TimerChain>> TimeGraph::GetAllSerializableTimerChains() const {
   std::vector<std::shared_ptr<TimerChain>> chains;
-  for (const auto& track : tracks_) {
+  for (const auto& track : track_manager_->GetTracks()) {
     Append(chains, track->GetAllSerializableChains());
   }
   return chains;
@@ -425,8 +415,6 @@ void TimeGraph::UpdateMaxTimeStamp(uint64_t time) {
     capture_max_timestamp_ = time;
   }
 };
-
-float TimeGraph::GetThreadTotalHeight() const { return std::abs(min_y_); }
 
 float TimeGraph::GetWorldFromTick(uint64_t time) const {
   if (time_window_us_ > 0) {
@@ -528,7 +516,7 @@ void TimeGraph::NeedsUpdate() {
 
 void TimeGraph::UpdatePrimitives(PickingMode picking_mode) {
   ORBIT_SCOPE_FUNCTION;
-  CHECK(string_manager_);
+  CHECK(track_manager_->GetStringManager() != nullptr);
 
   batcher_.StartNewFrame();
   text_renderer_static_.Clear();
@@ -543,13 +531,9 @@ void TimeGraph::UpdatePrimitives(PickingMode picking_mode) {
   uint64_t min_tick = GetTickFromUs(min_time_us_);
   uint64_t max_tick = GetTickFromUs(max_time_us_);
 
-  if (app_->IsCapturing() || sorted_tracks_.empty() || sorting_invalidated_) {
-    SortTracks();
-    sorting_invalidated_ = false;
-  }
-
-  UpdateMovingTrackSorting();
-  UpdateTracks(min_tick, max_tick, picking_mode);
+  track_manager_->SortTracks();
+  track_manager_->UpdateMovingTrackSorting();
+  track_manager_->UpdateTracks(min_tick, max_tick, picking_mode);
 
   needs_update_primitives_ = false;
 }
@@ -754,7 +738,7 @@ std::string TimeGraph::GetThreadNameFromTid(uint32_t tid) {
 }
 
 void TimeGraph::DrawTracks(GlCanvas* canvas, PickingMode picking_mode) {
-  for (auto& track : sorted_filtered_tracks_) {
+  for (auto& track : track_manager_->GetFilteredTracks()) {
     float z_offset = 0;
     if (track->IsPinned()) {
       z_offset = GlCanvas::kZOffsetPinnedTrack;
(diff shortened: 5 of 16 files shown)


OrbitGl/TrackManager.h
```diff
@@ -0,0 +1,97 @@
...
+  [[nodiscard]] StringManager* GetStringManager() const { return string_manager_.get(); }
+
+  void SortTracks();
+  void SetFilter(const std::string& filter);
+  void UpdateFilteredTrackList();
+
+  void UpdateTracks(uint64_t min_tick, uint64_t max_tick, PickingMode picking_mode);
+  float GetTracksHeight() const { return std::abs(tracks_height_); }
+
+  SchedulerTrack* GetOrCreateSchedulerTrack();
```
florian-kuebler, line 56: Can those be `const ...Track*(...) const;`? I assume that not, but maybe we should work toward this.
Also as soon as we have create and get seperated, we should have the getters be nodisard.

OrbitGl/TrackManager.h
```diff
@@ -0,0 +1,97 @@
...
+  GraphTrack* GetOrCreateGraphTrack(const std::string& name);
+  AsyncTrack* GetOrCreateAsyncTrack(const std::string& name);
+  FrameTrack* GetOrCreateFrameTrack(const orbit_client_protos::FunctionInfo& function);
+
+  void AddTrack(std::shared_ptr<Track> track);
+  void RemoveFrameTrack(uint64_t function_address);
+
+  void UpdateMovingTrackSorting();
+  int FindMovingTrackIndex();
+  [[nodiscard]] std::vector<int32_t> GetSortedThreadIds();
```
florian-kuebler, line 68: can this method be const?

florian-kuebler: If not, consider renaming it.

OrbitGl/TrackManager.h
```diff
@@ -0,0 +1,97 @@
...
+
+ protected:
```
florian-kuebler, line 70: As you don't use protected fields, you can just remove that section.

OrbitGl/TrackManager.cpp
```diff
@@ -0,0 +1,449 @@
...
+  frame_tracks_.clear();
+
+  sorted_tracks_.clear();
+  sorted_filtered_tracks_.clear();
+
+  GetOrCreateSchedulerTrack();
+  tracepoints_system_wide_track_ = GetOrCreateThreadTrack(orbit_base::kAllThreadsOfAllProcessesTid);
+}
+
+void TrackManager::SetStringManager(std::shared_ptr<StringManager> str_manager) {
```
florian-kuebler, line 55: Longterm, we probably want to get rid of the shared pointer here. Have you thoughts on this? Can we just use a raw pointer already?

OrbitGl/TrackManager.cpp
```diff
@@ -0,0 +1,449 @@
...
+  string_manager_ = std::move(str_manager);
+}
+
+std::vector<Track*> TrackManager::GetTracks() const {
+  std::vector<Track*> tracks;
+  for (auto track : tracks_) {
+    tracks.emplace_back(track.get());
```
florian-kuebler, line 62: As far as I know, there is no need to use emplace here, as the data_type is a primitive one anyways. So consider just use `push_back` here.

OrbitGl/TrackManager.cpp
```diff
@@ -0,0 +1,449 @@
...
+  }
+  return tracks;
+}
+
+void TrackManager::SortTracks() {
+  if (!GOrbitApp->IsCapturing() && !sorted_tracks_.empty() && !sorting_invalidated_) return;
+
+  ThreadTrack* process_track = nullptr;
+
+  const CaptureData* capture_data = time_graph_->GetCaptureData();
```
florian-kuebler, line 88: I am really unsure about the general "where does capture data comes from discussion". But for now, I think you really should use `GOrbitApp->GetCaptureData()`. Also the Timegraph should neither hold, and especially not expose this. But this is for a different PR.

OrbitGl/TrackManager.cpp
```diff
@@ -0,0 +1,449 @@
...
+  }
+}
+
+std::vector<int32_t>
+TrackManager::GetSortedThreadIds() {  // Show threads with instrumented functions first
+  std::vector<int32_t> sorted_thread_ids;
+  std::vector<std::pair<int32_t, uint32_t>> sorted_threads =
+      OrbitUtils::ReverseValueSort(thread_count_map_);
+
+  for (auto& pair : sorted_threads) {
```
florian-kuebler, line 215: Consider using for auto [...,...] syntax here and below.

OrbitGl/TrackManager.cpp
```diff
@@ -0,0 +1,449 @@
...
+  // Only drawn (i.e. not filtered out) tracks are taken into account to determine the
+  // insertion position, but both arrays are updated accordingly.
+  //
+  // Note: We do an O(n) search for the correct position in the sorted_tracks_ array which
+  // could be optimized, but this is not worth the effort for the limited number of tracks.
+
+  int moving_track_previous_position = FindMovingTrackIndex();
+
+  if (moving_track_previous_position != -1) {
+    auto moving_track = sorted_filtered_tracks_[moving_track_previous_position];
```
florian-kuebler, line 251: Please avoid using `[]` to access a map. Use `at` for simple access. And e.g. `insert` if you want to add something, only if it is not yet there.

OrbitGl/TrackManager.cpp
```diff
@@ -0,0 +1,449 @@
...
+    sorted_filtered_tracks_.erase(sorted_filtered_tracks_.begin() + moving_track_previous_position);
+
+    int moving_track_current_position = -1;
+    for (auto track_it = sorted_filtered_tracks_.begin(); track_it != sorted_filtered_tracks_.end();
+         ++track_it) {
+      if (moving_track->GetPos()[1] >= (*track_it)->GetPos()[1]) {
```
florian-kuebler, line 257: Again, please avoid `[]`.

OrbitGl/TrackManager.cpp
```diff
@@ -0,0 +1,449 @@
...
+
+    // Now we have to change the position of the moving_track in the non-filtered array
+    if (moving_track_current_position == moving_track_previous_position) {
+      return;
+    }
+    sorted_tracks_.erase(std::find(sorted_tracks_.begin(), sorted_tracks_.end(), moving_track));
+    if (moving_track_current_position > moving_track_previous_position) {
+      // In this case we will insert the moving_track right after the one who is before in the
+      // filtered array
+      auto& previous_filtered_track = sorted_filtered_tracks_[moving_track_current_position - 1];
```
florian-kuebler, line 277: Avoid auto here, and also the `[]`.

OrbitGl/TrackManager.cpp
```diff
@@ -0,0 +1,449 @@
...
+      sorted_tracks_.insert(
+          ++std::find(sorted_tracks_.begin(), sorted_tracks_.end(), previous_filtered_track),
+          moving_track);
+    } else {
+      // In this case we will insert the moving_track right before the one who is after in the
+      // filtered array
+      auto& next_filtered_track = sorted_filtered_tracks_[moving_track_current_position + 1];
```
florian-kuebler, line 284: As above, please avoid `auto` and `[]`.

OrbitGl/TrackManager.cpp
```diff
@@ -0,0 +1,449 @@
...
+    if ((*track_it)->IsMoving()) {
+      return track_it - sorted_filtered_tracks_.begin();
+    }
+  }
+  return -1;
+}
+
+void TrackManager::UpdateTracks(uint64_t min_tick, uint64_t max_tick, PickingMode picking_mode) {
+  // Make sure track tab fits in the viewport.
+  TimeGraphLayout layout = time_graph_->GetLayout();
```
florian-kuebler, line 305: Consider passing the Layout and Canvas (either as method param or just as a member of TrackManager) instead of having time_graph_ here in this class.

OrbitGl/TrackManager.cpp
```diff
@@ -0,0 +1,449 @@
...
+void TrackManager::RemoveFrameTrack(uint64_t function_address) {
+  frame_tracks_.erase(function_address);
+  sorting_invalidated_ = true;
+}
+
+SchedulerTrack* TrackManager::GetOrCreateSchedulerTrack() {
+  std::lock_guard<std::recursive_mutex> lock(mutex_);
+  std::shared_ptr<SchedulerTrack> track = scheduler_track_;
+  if (track == nullptr) {
+    track = std::make_shared<SchedulerTrack>(time_graph_);
```
florian-kuebler, line 354: Again, mid-term, SchedulerTrack should not know time_graph_.

OrbitGl/TrackManager.cpp
```diff
@@ -0,0 +1,449 @@
...
+    AddTrack(track);
+    scheduler_track_ = track;
+    uint32_t num_cores = time_graph_->GetNumCores();
```
florian-kuebler, line 357: Consider using the number of cores from the layout.

OrbitGl/TrackManager.cpp
```diff
@@ -0,0 +1,449 @@
...
+}
+
+ThreadTrack* TrackManager::GetOrCreateThreadTrack(int32_t tid) {
+  std::lock_guard<std::recursive_mutex> lock(mutex_);
+  std::shared_ptr<ThreadTrack> track = thread_tracks_[tid];
+  if (track == nullptr) {
+    track = std::make_shared<ThreadTrack>(time_graph_, tid);
+    AddTrack(track);
+    thread_tracks_[tid] = track;
+    track->SetTrackColor(time_graph_->GetThreadColor(tid));
```
florian-kuebler, line 371: Static methods should be called using the classname and not an object.
`TimeGraph::GetThreadColor()`. Also, in order to get rid of this forward declaration and the usage of timegraph here, consider moving these methods. (Probably in a different PR).

OrbitGl/TrackManager.cpp
```diff
@@ -0,0 +1,449 @@
...
+    if (tid == orbit_base::kAllThreadsOfAllProcessesTid) {
+      track->SetName("All tracepoint events");
+      track->SetLabel("All tracepoint events");
+    } else if (tid == orbit_base::kAllProcessThreadsTid) {
+      // This is the process track.
+      auto capture_data = time_graph_->GetCaptureData();
```
florian-kuebler, line 377: Please avoid auto and also use GOrbitApp->GetCaptureData for now.

florian-kuebler (commented):
Done with my pass over the code. Conceptually looks good. I left a couple of minor comments. 
As a general note: We should really try to not have TimeGraph known in the TrackManager and the Tracks. In most cases it seems to be the Canvas or Layout that is of interest. We should work around this as soon as possible.

reichlfl (commented):
Thanks a lot for this, Pablo, it's a great step towards reducing bloat in classes like the TimeGraph. Agree with FlorianK'd comments, and added a few on top.

OrbitGl/TimeGraph.h
```diff
@@ -5,33 +5,34 @@
...
 #include "TextRenderer.h"
-#include "ThreadTrack.h"
 #include "TimeGraphLayout.h"
-#include "Timer.h"
 #include "TimerChain.h"
+#include "Track.h"  // for Track
 #include "absl/container/flat_hash_map.h"
 #include "capture_data.pb.h"
 
+class TrackManager;
```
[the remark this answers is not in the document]
florian-kuebler: Right, and sorry for the confusion (there is a forward declaration in both, which is definitely an issue).

OrbitGl/TrackManager.h
```diff
@@ -0,0 +1,97 @@
...
+#include "GraphTrack.h"
+#include "PickingManager.h"  // for PickingMode
+#include "SchedulerTrack.h"
+#include "StringManager.h"  // for StringManager
+#include "ThreadTrack.h"
+#include "Timer.h"            // for Timer
+#include "Track.h"            // for Track
+#include "capture_data.pb.h"  // for FunctionInfo
+
+class TimeGraph;
```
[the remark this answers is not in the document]
florian-kuebler: Right. So basically keep this forward declaration and remove the other one. But also add a bug for that. Mid-term, expose all the information that we use from TimeGraph directly (maybe even as parameter instead of having it as a member). Also isn't too much.

OrbitGl/TrackManager.h
```diff
@@ -0,0 +1,97 @@
...
+  void AddTrack(std::shared_ptr<Track> track);
+  void RemoveFrameTrack(uint64_t function_address);
+
+  void UpdateMovingTrackSorting();
+  int FindMovingTrackIndex();
+  [[nodiscard]] std::vector<int32_t> GetSortedThreadIds();
+
+ protected:
+ private:
+  mutable std::recursive_mutex mutex_;
```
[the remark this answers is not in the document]
florian-kuebler: Nope there is not. We also encourage writing code that does not require re-entrancy :)

OrbitGl/FrameTrack.cpp
```diff
@@ -4,10 +4,21 @@
 
 #include "FrameTrack.h"
 
-#include "App.h"
+#include <GteVector.h>                // for Vector, operator*, operator*=, operator+=, operator+
```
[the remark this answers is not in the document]
karupayun: Yes, that lines are entirely suggested by IWYU. I supposed that we wanted that from now on, but no problem in erasing then. @beckerhe ??

And also, I only modified FrameTrack.h, sorry for including the cpp ones.

OrbitGl/TimeGraph.cpp
```diff
@@ -263,25 +259,26 @@ void TimeGraph::ProcessTimer(const TimerInfo& timer_info, const FunctionInfo* fu
 
   if (timer_info.type() == TimerInfo::kGpuActivity) {
     uint64_t timeline_hash = timer_info.timeline_hash();
-    std::shared_ptr<GpuTrack> track = GetOrCreateGpuTrack(timeline_hash);
+    auto track = track_manager_->GetOrCreateGpuTrack(timeline_hash);
```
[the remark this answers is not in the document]
karupayun: I thought about that and finally used auto, my bad. I only used auto because it's easier to switch from different kinds of pointers, but you are right.

OrbitGl/TimeGraph.cpp
```diff
@@ -425,7 +422,7 @@ void TimeGraph::UpdateMaxTimeStamp(uint64_t time) {
   }
 };
 
-float TimeGraph::GetThreadTotalHeight() const { return std::abs(min_y_); }
+float TimeGraph::GetThreadTotalHeight() const { return track_manager_->GetTracksHeight(); }
```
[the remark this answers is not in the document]
karupayun: Thanks. I have to change both of them.

OrbitGl/TimeGraph.cpp
```diff
@@ -542,13 +539,9 @@ void TimeGraph::UpdatePrimitives(PickingMode picking_mode) {
...
 
-  if (GOrbitApp->IsCapturing() || sorted_tracks_.empty() || sorting_invalidated_) {
-    SortTracks();
-    sorting_invalidated_ = false;
-  }
-
-  UpdateMovingTrackSorting();
-  UpdateTracks(min_tick, max_tick, picking_mode);
+  track_manager_->SortTracks();
+  track_manager_->UpdateMovingTrackSorting();
```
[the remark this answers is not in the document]
karupayun: Yes, is part of my next refactoring steps. UpdateMovingTrackSorting could be part of SortTracks.

OrbitGl/TimeGraph.h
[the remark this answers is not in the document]
karupayun: My bad here, sorry. Should include TrackManager.h, instead of forward-declarating.

OrbitGl/TrackManager.h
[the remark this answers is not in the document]
karupayun: Again here, sorry for the other forward declaration. And about this one. I wrote my next steps here, which I'm not planning to do all of them, but erasing TimeGraph dependence in TrackManager and Tracks is my P0, because I wanted to make tests after that. Given the amount of comments which I received, I will do another PR and therefore, write a TODO here.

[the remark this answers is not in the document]
karupayun: And thanks for the link florianK :)

OrbitGl/TrackManager.h
```diff
@@ -0,0 +1,97 @@
...
+  GpuTrack* GetOrCreateGpuTrack(uint64_t timeline_hash);
+  GraphTrack* GetOrCreateGraphTrack(const std::string& name);
+  AsyncTrack* GetOrCreateAsyncTrack(const std::string& name);
+  FrameTrack* GetOrCreateFrameTrack(const orbit_client_protos::FunctionInfo& function);
+
+  void AddTrack(std::shared_ptr<Track> track);
+  void RemoveFrameTrack(uint64_t function_address);
+
+  void UpdateMovingTrackSorting();
+  int FindMovingTrackIndex();
```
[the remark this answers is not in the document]
karupayun: Not, of course, thanks.

OrbitGl/TrackManager.h
```diff
@@ -0,0 +1,97 @@
...
+
+  void SetStringManager(std::shared_ptr<StringManager> str_manager);
+  [[nodiscard]] StringManager* GetStringManager() const { return string_manager_.get(); }
+
+  void SortTracks();
+  void SetFilter(const std::string& filter);
+  void UpdateFilteredTrackList();
+
+  void UpdateTracks(uint64_t min_tick, uint64_t max_tick, PickingMode picking_mode);
+  float GetTracksHeight() const { return std::abs(tracks_height_); }
```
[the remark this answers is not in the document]
karupayun: I've copied it from TimeGraph. This function isn't being very precise (track height seems to be a little bit smaller than the value of the variable), so I want to investigate a little bit.

Basically we set 0 to the top coordinate and draw everything to the bottom (so, in negatives coordinates). And this function actually could return -tracks_height_.

So, will add a TODO here, where next steps will be analizing why this unprecision, and also change tracks_height_ to have a positive value and return directly tracks_height_.

OrbitGl/TrackManager.h
karupayun: Something that with Henning we discussed a little about. For now, I'm leaving as it is, but definitely I want to split it. In fact, I want to have an only list of Tracks and only 1 function to get Tracks (with a template?), but not sure if I will succeed with that.

Will add a TODO here.

OrbitGl/TrackManager.h
karupayun: Sure!

OrbitGl/TrackManager.cpp
karupayun: Not sure but I had a crash when I tried xD. I wrote it down in my infinite list of task to do :)

OrbitGl/TrackManager.cpp
karupayun: In this case GetPos() returns a Vec2. You still didn't like the []?

OrbitGl/TrackManager.cpp
karupayun: Yes, completely agree and part of the next steps.

OrbitGl/TimeGraph.cpp
```diff
@@ -226,7 +222,7 @@ void TimeGraph::HorizontallyMoveIntoView(VisibilityType vis_type, const TimerInf
 }
 
 void TimeGraph::VerticallyMoveIntoView(const TimerInfo& timer_info) {
-  VerticallyMoveIntoView(*GetOrCreateThreadTrack(timer_info.thread_id()));
+  VerticallyMoveIntoView(*track_manager_->GetOrCreateThreadTrack(timer_info.thread_id()));
```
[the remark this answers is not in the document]
karupayun: For a future PR, for sure. Here we are using GetOrCreate for simply a Get (and maybe a CHECK?). The track should exist (but in the future maybe could be not visible).

OrbitGl/TimeGraph.cpp
```diff
@@ -263,25 +259,26 @@ void TimeGraph::ProcessTimer(const TimerInfo& timer_info, const FunctionInfo* fu
 
   if (timer_info.type() == TimerInfo::kGpuActivity) {
```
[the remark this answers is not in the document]
karupayun: I'm undecided here. At the first I was thinking TrackManager as only a manager of the Tracks and their order. So, in this point of view, TimeGraph could know about Track types. But maybe you are right and we could also move the logic to the Tracks, and TimeGraph only needs to have a way to get a Track given a timer. I will add a TODO here and think about that when refactor this GetOrCreate method.

OrbitGl/TimeGraph.cpp
```diff
@@ -764,364 +757,12 @@ void TimeGraph::DrawTracks(GlCanvas* canvas, PickingMode picking_mode) {
...
-
-    const float z_offset = track->IsMoving() ? GlCanvas::kZOffsetMovingTack : 0.f;
-    track->SetY(current_y);
-    track->UpdatePrimitives(min_tick, max_tick, picking_mode, z_offset);
-    current_y -= (track->GetHeight() + layout_.GetSpaceBetweenTracks());
-  }
-
-  min_y_ = current_y;
-}
+void TimeGraph::SortTracks() { track_manager_->SortTracks(); }
```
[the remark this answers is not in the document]
karupayun: I liked this proxy methods to have a better defined parent-child relationship. 

App -> TimeGraph -> TrackManager -> Tracks
Why App has to know about TrackManager?

But, I'm completely open to discuss about that.

OrbitGl/TrackManager.h
```diff
@@ -0,0 +1,97 @@
...
+class TimeGraph;
+
+class TrackManager {
+ public:
+  explicit TrackManager(TimeGraph* time_graph);
+
+  void Clear();
+
+  [[nodiscard]] std::vector<Track*> GetTracks() const;
+  [[nodiscard]] std::vector<Track*> GetFilteredTracks() const { return sorted_filtered_tracks_; }
```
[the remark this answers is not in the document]
karupayun: Of course, a really better name!

OrbitGl/TrackManager.h
```diff
@@ -0,0 +1,97 @@
...
+  [[nodiscard]] ThreadTrack* GetTracepointsSystemWideTrack() const {
+    return tracepoints_system_wide_track_;
+  }
+
+  void SetStringManager(std::shared_ptr<StringManager> str_manager);
+  [[nodiscard]] StringManager* GetStringManager() const { return string_manager_.get(); }
+
+  void SortTracks();
+  void SetFilter(const std::string& filter);
+  void UpdateFilteredTrackList();
```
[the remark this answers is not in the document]
karupayun: It was there before.

We update the Filtered (or Visible) track list when we change the filter and when we Sort Tracks (because we get the filtered from the sorted).

But maybe we can Sort directly both the visible and non-visible Tracks. I'm not seeing a priority here, let me know if I am wrong.

OrbitGl/TrackManager.h
```diff
@@ -0,0 +1,97 @@
...
+
+class TimeGraph;
+
+class TrackManager {
+ public:
+  explicit TrackManager(TimeGraph* time_graph);
+
+  void Clear();
+
+  [[nodiscard]] std::vector<Track*> GetTracks() const;
```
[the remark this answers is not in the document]
karupayun: ok!

OrbitGl/TrackManager.cpp
```diff
@@ -0,0 +1,449 @@
...
+std::vector<FrameTrack*> TrackManager::GetFrameTracks() const {
+  std::vector<FrameTrack*> tracks;
+  for (auto& [unused_key, track] : frame_tracks_) {
+    tracks.emplace_back(track.get());
+  }
+  return tracks;
+}
+
+void TrackManager::SortTracks() {
+  if (!GOrbitApp->IsCapturing() && !sorted_tracks_.empty() && !sorting_invalidated_) return;
```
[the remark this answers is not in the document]
karupayun: Good point, will think about that.

OrbitGl/TimeGraph.cpp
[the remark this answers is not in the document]
reichlfl: Especially since the track manager is already exposed, I see three issues with this:
1) It increases the amount of methods in TimeGraph's API, making it less clear what's the scope of responsibility for this class.
2) Every change done in TrackManager needs to "bubble up". Just renaming methods or changing parameters requires changing this in TimeGraph as well
3) And for me, this is the most important one: It is not clear to the user which method to use, because both "TimeGraph::SortTracks()" and "TimeGraph::GetTrackManager()::SortTracks()" are available to them. You need to look at the implementation to figure out that those actually do the exact same thing.

OrbitGl/TrackManager.h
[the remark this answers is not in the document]
reichlfl: I'm aware I'm blaming you for the bad choices of others (potentially even my own choices) :D
It's fine to have that method if it's used in two different places, but from what you describe it should be private then.

OrbitGl/TimeGraph.cpp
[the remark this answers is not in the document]
karupayun: So, you recommend to do a GetTimeGraph->GetTrackManager -> SortTracks on App?

OrbitGl/TrackManager.h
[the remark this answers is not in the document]
karupayun: UpdateFilteredTrackList could be inside SetFilter and also SortTracks. I was thinking when to make it, but better to make in this PR.

OrbitGl/TimeGraph.cpp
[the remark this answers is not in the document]
reichlfl: If we keep the call to SortTracks in App, yes, I would suggest that.

But this also relates to what I called out regarding the name of `SortTracks`. We're only using this method in `OrbitApp` to update the sorting of the tracks. Why is this needed? Why does `OrbitApp` need to know that the `TimeGraph` wants to do a re-sorting of tracks after timer updates? Sorting is the domain of the `TrackManager`, and it should be responsible to decide when it's time to re-sort the tracks, not the App. **Especially** since the `TrackManager` already decides on its own if it actually wants to perform the sorting as instructed, or just does an early-out in `SortTracks` based on conditions not visible to the user of the class. 

The track manager should know when timers are added to the Tracks, and it can then trigger re-sorting of the tracks when it thinks it's required, but the App should not know nor care about this. 

Since currently timers are added through the `TimeGraph`, it's a reasonable compromise for now to have the `TimeGraph` decide when it's time to sort the tracks, but I'd argue that the Timers should be passed to `TrackManager`, then this class decides which tracks to add them to and if it's time to perform a sorting, and neither the App nor the TimeGraph need to care.

OrbitGl/TrackManager.cpp
florian-kuebler: Oh, then it is probably fine.

OrbitGl/TrackManager.cpp
karupayun: BTW, now seems that is working, probably was another shared_pointer. I will get rid of this one.

OrbitGl/TrackManager.cpp
karupayun: again here, filtered track is a vector

OrbitGl/TrackManager.h
```diff
@@ -0,0 +1,97 @@
...
+#include "SchedulerTrack.h"
+#include "StringManager.h"  // for StringManager
+#include "ThreadTrack.h"
+#include "Timer.h"            // for Timer
+#include "Track.h"            // for Track
+#include "capture_data.pb.h"  // for FunctionInfo
+
+class TimeGraph;
+
+class TrackManager {
```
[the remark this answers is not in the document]
karupayun: something short is OK? What do you think:
// TrackManager is in charge of the active Tracks in Timegraph (their creation, searching, erasing
// and sorting).

OrbitGl/TrackManager.h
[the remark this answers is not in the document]
karupayun: Changing plans:  Not TODO. Follow @reichl idea and make a comment there.

OrbitGl/FrameTrack.h
```diff
@@ -5,12 +5,25 @@
...
+#include "TimerChain.h"      // for TimerChain
 #include "TimerTrack.h"
+#include "Track.h"            // for Track::Type, Track::kFrameTrack
+#include "capture_data.pb.h"  // for TimerInfo, FunctionInfo, FunctionStats
 
 class FrameTrack : public TimerTrack {
  public:
   explicit FrameTrack(TimeGraph* time_graph, const orbit_client_protos::FunctionInfo& function);
   [[nodiscard]] Type GetType() const override { return kFrameTrack; }
+  [[nodiscard]] uint64_t GetFunctionAddress() const { return function_.address(); }
```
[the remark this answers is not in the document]
karupayun: I've decided to make a bug for this issue: http://b/175865913

OrbitGl/TimeGraph.cpp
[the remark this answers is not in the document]
karupayun: Added a comment and a bug here.

OrbitGl/TimeGraph.cpp
```diff
@@ -309,7 +306,7 @@ void TimeGraph::ProcessIntrospectionTimer(const TimerInfo& timer_info) {
 
   switch (event.type) {
```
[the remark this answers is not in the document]
karupayun: b/175869409

OrbitGl/TimeGraph.cpp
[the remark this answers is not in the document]
karupayun: Erased this method, renamed to GetTotalTracksHeight and updated references.

OrbitGl/TimeGraph.cpp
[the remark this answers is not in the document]
karupayun: 90% agree. I'm ok (or at least not 100% sure of moving them) with Timers added from TimeGraph, but TrackManager should manage their own sorting of Tracks. 

Erased the method in TimeGraph and added a TODO for that.

OrbitGl/TimeGraph.cpp
```diff
@@ -1259,11 +900,13 @@ bool TimeGraph::IsVisible(VisibilityType vis_type, uint64_t min, uint64_t max) c
 }
 
 bool TimeGraph::HasFrameTrack(const orbit_client_protos::FunctionInfo& function) const {
-  return frame_tracks_.count(function.address()) > 0;
+  auto frame_tracks = track_manager_->GetFrameTracks();
+  return (std::find_if(frame_tracks.begin(), frame_tracks.end(), [&](auto frame_track) {
```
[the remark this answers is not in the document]
karupayun: b/175865913

OrbitGl/TrackManager.cpp
karupayun: Sorted_filtered_track is a vector :)

OrbitGl/TrackManager.cpp
karupayun: http://b/176056427

OrbitGl/TrackManager.h
```diff
@@ -0,0 +1,97 @@
...
+
+  void UpdateTracks(uint64_t min_tick, uint64_t max_tick, PickingMode picking_mode);
+  float GetTracksHeight() const { return std::abs(tracks_height_); }
+
+  SchedulerTrack* GetOrCreateSchedulerTrack();
+  ThreadTrack* GetOrCreateThreadTrack(int32_t tid);
+  GpuTrack* GetOrCreateGpuTrack(uint64_t timeline_hash);
+  GraphTrack* GetOrCreateGraphTrack(const std::string& name);
+  AsyncTrack* GetOrCreateAsyncTrack(const std::string& name);
+  FrameTrack* GetOrCreateFrameTrack(const orbit_client_protos::FunctionInfo& function);
```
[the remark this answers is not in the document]
karupayun: After splitting in GetFunctions and Create functions, I agree. Now it isn't too clear. For example, I'm using GetOrCreateSchedulerTrack to create the Scheduler.

OrbitGl/TimeGraph.cpp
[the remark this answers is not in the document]
karupayun: b/176077097

OrbitGl/App.cpp
```diff
@@ -756,7 +757,8 @@ PresetLoadState OrbitApp::GetPresetLoadState(
 }
 
 ErrorMessageOr<void> OrbitApp::OnSaveCapture(const std::string& file_name) {
-  const auto& key_to_string_map = GCurrentTimeGraph->GetStringManager()->GetKeyToStringMap();
+  const auto& key_to_string_map =
+      GCurrentTimeGraph->GetTrackManager()->GetStringManager()->GetKeyToStringMap();
```
reichlfl, line 761: The hirarchy "TimeGraph -> TrackManager -> StringManager" feels wrong - can't we use the string manager of the time graph directly?

I'm overall not sure why we're passing the string manager around - there is only a single instance of `StringManager` anyways that is already living in `OrbitApp`, and the only point in our code where this is different is in `CaptureDeserializerLoadFuzzer`, where a new `StringManager` is created and passed to the `TimeGraph`, even though an App exists, not sure why it is done like this?

OrbitGl/GpuTrack.cpp
```diff
@@ -128,9 +127,12 @@ void GpuTrack::SetTimesliceText(const TimerInfo& timer_info, double elapsed_us,
 
     CHECK(timer_info.type() == TimerInfo::kGpuActivity);
 
-    std::string text = absl::StrFormat(
-        "%s  %s", time_graph_->GetStringManager()->Get(timer_info.user_data_key()).value_or(""),
-        time.c_str());
+    std::string text = absl::StrFormat("%s  %s",
+                                       time_graph_->GetTrackManager()
+                                           ->GetStringManager()
```
reichlfl, line 132: There is also `string_manager_`, which should be the same?

OrbitGl/TimeGraph.cpp
```diff
@@ -61,11 +55,9 @@ TimeGraph::~TimeGraph() {
 }
 
 void TimeGraph::SetStringManager(std::shared_ptr<StringManager> str_manager) {
```
reichlfl, line 57: Since you changed this to a raw pointer in `TrackManager`, wouldn't it be consistent to do the same here?

OrbitGl/TimeGraph.h
```diff
@@ -115,7 +113,6 @@ class TimeGraph {
   void SetTextRenderer(TextRenderer* text_renderer) { text_renderer_ = text_renderer; }
   [[nodiscard]] TextRenderer* GetTextRenderer() { return &text_renderer_static_; }
   void SetStringManager(std::shared_ptr<StringManager> str_manager);
-  [[nodiscard]] StringManager* GetStringManager() const;
```
reichlfl: Ah, now I get why you changed it above. In this case, I'd argue that it's fine to have a `GetStringManager()` method here since this is actually a getter for a member. But I'd question if those getters are needed at all, as mentioned above.

OrbitGl/TrackManager.cpp
```diff
@@ -368,15 +364,14 @@ ThreadTrack* TrackManager::GetOrCreateThreadTrack(int32_t tid) {
...
+    track->SetTrackColor(TimeGraph::GetThreadColor(tid));
     if (tid == orbit_base::kAllThreadsOfAllProcessesTid) {
       track->SetName("All tracepoint events");
       track->SetLabel("All tracepoint events");
     } else if (tid == orbit_base::kAllProcessThreadsTid) {
       // This is the process track.
-      auto capture_data = time_graph_->GetCaptureData();
-      CHECK(capture_data);
-      std::string process_name = capture_data->process_name();
+      const CaptureData& capture_data = GOrbitApp->GetCaptureData();
```
reichlfl, line 373: Did you remove the CHECK() on purpose?

OrbitGl/TrackManager.h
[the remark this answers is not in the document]
reichlfl: I think you're missing the main point: I'm mostly arguing you should just move this to the `private` section since it's never called from outside as far as I can see. I'm fine with the method existing, but it should not be public.

OrbitGl/TrackManager.h
reichlfl: Sounds good, it's perfectly fine to address that later.

OrbitGl/TrackManager.h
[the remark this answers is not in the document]
reichlfl: Nit: I'd put a blank line between the forward declaration of Timegraph and the comment.

reichlfl (approved):
Minor comments that should be very quick fixes, would be nice to take care of those - LGTM otherwise. If the comments turn out to be more effort it's fine to address them later.

OrbitGl/TrackManager.cpp
reichlfl: I'd disagree here - I think the layout obviously lives in the TimeGraph (given it's even named `TimeGraphLayout`), I'd avoid passing elements just to reduce indirections.

OrbitGl/TrackManager.cpp
karupayun: There is a CHECK on that method.

OrbitGl/TrackManager.h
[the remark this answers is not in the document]
karupayun: My bad!!! I was 100% sure that we had a reference to this method in TimeGraph and couldn't be private.

OrbitGl/TimeGraph.cpp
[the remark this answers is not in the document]
karupayun: Will be part of b/176077097

OrbitGl/TrackManager.h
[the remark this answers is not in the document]
karupayun: b/176086740

OrbitGl/TrackManager.h
karupayun: Will be part of b/176077097.

OrbitGl/TrackManager.cpp
karupayun: b/176086740

OrbitGl/TrackManager.cpp
karupayun: b/176086740

OrbitGl/TrackManager.cpp
karupayun: b/176086740

OrbitGl/TrackManager.cpp
karupayun: b/176086740

OrbitGl/TrackManager.h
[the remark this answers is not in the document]
karupayun: b/175869409

OrbitGl/App.cpp
karupayun: I will try to change it as my next PR and if not, make a bug there.

OrbitGl/TimeGraph.cpp
karupayun: They are different things. I will try to erase StringManager in TimeGraph and TrackManager and make a new one on GpuTrack.

OrbitGl/TimeGraph.h
karupayun: Agree, part of my next PR.

karupayun:
Updated commit and PR messages to summarize next steps discussed here.

PR #1568, opened 2020-12-09
ReaderMutexLock -> WriterMutexLock in CaptureEventProducer

dpallotti:
Just noticed it by chance.


florian-kuebler (approved):
Yep, that looks better. Thanks!

PR #1569, opened 2020-12-10
Split the mutex for process and callback update

reichlfl:
Update of the process_list_ and process_list_update_listener_ was using the same mutex.

This causes a deadlock when the process_list_update_listener_ calls GetProcessList() when invoked from the WorkerFunction().
Problem could be observed in the new UI beta branch: Process List waits to be loaded indefinitely after connecting to an instance.

### DIFF
--- a/OrbitClientServices/ProcessManager.cpp
+++ b/OrbitClientServices/ProcessManager.cpp
@@ -32,7 +32,7 @@ class ProcessManagerImpl final : public ProcessManager {
 
   ~ProcessManagerImpl() override { ShutdownAndWait(); }
 
-  void SetProcessListUpdateListener(const std::function<void(ProcessManager*)>& listener) override;
+  void SetProcessListUpdateListener(const std::function<void(const ProcessList)> listener) override;
 
   std::vector<ProcessInfo> GetProcessList() const override;
   ErrorMessageOr<std::vector<ModuleInfo>> LoadModuleList(int32_t pid) override;
@@ -60,7 +60,7 @@ class ProcessManagerImpl final : public ProcessManager {
   std::vector<ProcessInfo> process_list_;
 
   mutable absl::Mutex process_list_update_listener_mutex_;
-  std::function<void(ProcessManager*)> process_list_update_listener_;
+  std::function<void(const ProcessList)> process_list_update_listener_;
 
   std::thread worker_thread_;
 };
@@ -72,7 +72,7 @@ ProcessManagerImpl::ProcessManagerImpl(const std::shared_ptr<grpc::Channel>& cha
 }
 
 void ProcessManagerImpl::SetProcessListUpdateListener(
-    const std::function<void(ProcessManager*)>& listener) {
+    const std::function<void(const ProcessList)> listener) {
   absl::MutexLock lock(&process_list_update_listener_mutex_);
   process_list_update_listener_ = listener;
 }
@@ -134,7 +134,7 @@ void ProcessManagerImpl::WorkerFunction() {
     {
       absl::MutexLock callback_lock(&process_list_update_listener_mutex_);
       if (process_list_update_listener_) {
-        process_list_update_listener_(this);
+        process_list_update_listener_(std::move(result.value()));
       }
     }
   }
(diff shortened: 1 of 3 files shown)


antonrohr (approved):
You solve my nitpicks faster than I can comment on them :) 
LGTM, thank you a lot!

OrbitClientServices/ProcessManager.cpp
```diff
@@ -124,11 +126,16 @@ void ProcessManagerImpl::WorkerFunction() {
...
+    {
+      absl::MutexLock process_lock(&process_list_mutex_);
+      const auto& processes = result.value();
+      process_list_.assign(processes.begin(), processes.end());
+    }
+    {
+      absl::MutexLock callback_lock(&process_list_update_listener_mutex_);
+      if (process_list_update_listener_) {
+        process_list_update_listener_(this);
+      }
```
dimitry-, line 138: I am guessing the problem here is that listener may call GetProcessList which will lead to recursive lock. Have you considered passing reference to process_list_ here maybe?

dimitry-:
> Update of the process_list_ and process_list_update_listener_ was using the same mutex.
> 
> This causes a deadlock when the process_list_update_listener_ calls GetProcessList() when invoked from the WorkerFunction().
> Problem could be observed in the new UI beta branch: Process List waits to be loaded indefinitely after connecting to an instance.

also - this should go to commit message probably. I did not read this because I was looking at the commit message :)

OrbitClientServices/ProcessManager.cpp
antonrohr: I like this Idea as well. Then I would get rid of the function GetProcessList() altogether, which requires a small change in App.cpp where GetProcessList() is used. 
I also dont mind how the change is right now.

reichlfl: Good point. I changed that, PTAL. Nevertheless, I think it's a good idea to keep this as two mutexes, since they guard different objects, and it seems excessive to block the whole process list update during the callback when the callback operates on a copy anyways.

reichlfl:
> > Update of the process_list_ and process_list_update_listener_ was using the same mutex.
> > This causes a deadlock when the process_list_update_listener_ calls GetProcessList() when invoked from the WorkerFunction().
> > Problem could be observed in the new UI beta branch: Process List waits to be loaded indefinitely after connecting to an instance.
> 
> also - this should go to commit message probably. I did not read this because I was looking at the commit message :)

True, will do that with the final rebase.

OrbitClientServices/include/OrbitClientServices/ProcessManager.h
```diff
@@ -35,17 +35,19 @@
 //
 class ProcessManager {
  public:
+  using ProcessList = std::vector<orbit_grpc_protos::ProcessInfo>;
```
dimitry-, line 38: maybe keep vector in definitions? this using is hiding information about the type... I almost confused it with class we used to have in OrbitCore.

OrbitClientServices/ProcessManager.cpp
```diff
@@ -124,11 +126,16 @@ void ProcessManagerImpl::WorkerFunction() {
...
 
-    absl::MutexLock callback_lock(&mutex_);
-    const auto& processes = result.value();
-    process_list_.assign(processes.begin(), processes.end());
-    if (process_list_update_listener_) {
-      process_list_update_listener_(this);
+    {
+      absl::MutexLock process_lock(&process_list_mutex_);
+      const auto& processes = result.value();
+      process_list_.assign(processes.begin(), processes.end());
```
dimitry-, line 132: Does anyone use process_list_ ?

OrbitClientServices/include/OrbitClientServices/ProcessManager.h
reichlfl: I think it's more readable, and there's IDEs to display type info ;-) But in all seriousness, I don't have a strong opinion, I wasn't touching this part of the code so much before, so if it's confusing I'll keep the std::vector<> in the definitions.

OrbitClientServices/ProcessManager.cpp
reichlfl: Hm, good point. Actually, no. It's returned by GetProcessList(), but after this change, this isn't used at all. But I'm not sure if it will be needed with @antonrohr 's changes?

The current fix is in his change as a cherrypick because it was causing a crash, so I don't want to change this all too much before rebasing, but we could also not merge this fix and remove the whole GetProcessList() (and probably the mutex altogether...) afterwards.

reichlfl:
Handled with https://github.com/google/orbit/pull/1574, closing

PR #1570, opened 2020-12-10
Fix Vulkan validation layer include path

beckerhe:
Depending on the version of the vulkan validationlayers header package,
the include path is different. Old versions put the headers into the
vulkan/ subdirectory, laters ones don't use that sub-directory.

Since we have to work with both versions, this commit adds a cmake
find-package detection script which adds a cmake target
Vulkan::ValidationLayers which conveys for the proper include path.

The vk_layer_dispatch_table.h header is not self-contained and needs a
special inclusion order. That's why this commit als adds some
clang-format descriptors which avoid reordering of the vulkan layers.
Ideally this will be fixed upstream.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -117,6 +117,11 @@ endif()
 
 find_package(Filesystem REQUIRED)
 
+if(NOT WIN32)
+  find_package(Vulkan REQUIRED)
+  find_package(VulkanValidationLayers REQUIRED)
+endif()
+
 # Set preprocessor defines These are only necessary for windows, but we will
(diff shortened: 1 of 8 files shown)


florian-kuebler (approved):
Thanks a lot. I can't really judge about the cmake script, but it seem to be fine. LGTM!

dpallotti (approved):
Thanks a lot for quickly addressing this!

beckerhe:
Thanks. I will quickly test this on a Windows workstation - just to be sure - and then merge it.

beckerhe:
Seems to work.

PR #1571, opened 2020-12-10
Fix generate_coverage_report.sh

antonrohr:
This change makes the coverage report independent from the location it
is opened. Beforehand one path was an absolute path, now it is relative.


florian-kuebler (approved):
LGTM

PR #1573, opened 2020-12-10
Fix bug in TracepointData.cpp

beckerhe:
An iterator was dereferenced before checking its validity.
Checking before dereferencing fixes it.

Bug: b/175300672


florian-kuebler (approved):
Thanks for that. LGTM, but super nit comment about the commit message.
Could you add "http://" as prefix to the bug?
Also adding how to test that change (e.g. "Test: Compile & run tests.") would be nice.

beckerhe:
> Thanks for that. LGTM, but super nit comment about the commit message.
> Could you add "http://" as prefix to the bug?
> Also adding how to test that change (e.g. "Test: Compile & run tests.") would be nice.

Sure. Done!

florian-kuebler:
Thanks!

PR #1574, opened 2020-12-10
Pass process list directly to ProcessListUpdateListener

dimitry-:
Also remove unused GetProcessList method

Test: builds

### DIFF
--- a/OrbitClientServices/ProcessManager.cpp
+++ b/OrbitClientServices/ProcessManager.cpp
@@ -32,9 +32,9 @@ class ProcessManagerImpl final : public ProcessManager {
 
   ~ProcessManagerImpl() override { ShutdownAndWait(); }
 
-  void SetProcessListUpdateListener(const std::function<void(ProcessManager*)>& listener) override;
+  void SetProcessListUpdateListener(
+      const std::function<void(std::vector<orbit_grpc_protos::ProcessInfo>)>& listener) override;
 
-  std::vector<ProcessInfo> GetProcessList() const override;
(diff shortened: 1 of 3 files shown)


dimitry-:
I think we still need mutex here but now to guard access to listener.

OrbitClientServices/ProcessManager.cpp
```diff
@@ -57,8 +57,7 @@ class ProcessManagerImpl final : public ProcessManager {
   bool shutdown_initiated_;
 
   mutable absl::Mutex mutex_;
```
reichlfl, line 59: Nit: I'd prefer to have this named in a less generic way to indicate what it's supposed to be guarding.

reichlfl (approved):
LGTM, thanks. Maybe we should wait for https://github.com/google/orbit/pull/1526 to be merged before merging though.

antonrohr (approved):
LGTM. You do not need to wait for #1526.

OrbitClientServices/ProcessManager.cpp
dimitry-: done.

OrbitClientServices/ProcessManager.cpp
```diff
@@ -56,9 +56,8 @@ class ProcessManagerImpl final : public ProcessManager {
   absl::Mutex shutdown_mutex_;
   bool shutdown_initiated_;
 
-  mutable absl::Mutex mutex_;
-  std::vector<ProcessInfo> process_list_;
-  std::function<void(ProcessManager*)> process_list_update_listener_;
+  mutable absl::Mutex process_list_update_listener_mutex_;
```
dimitry-, line 59: and removed mutable.

PR #1575, opened 2020-12-10
Add typed ConsumeRecord and CHECK for size

dimitry-:
Provide template variant of ConsumeRecord and add size to void* variant to always explicitly specify size. 

Test: Build, Start Orbit, Capture

### DIFF
--- a/OrbitLinuxTracing/PerfEventRingBuffer.cpp
+++ b/OrbitLinuxTracing/PerfEventRingBuffer.cpp
@@ -121,7 +121,7 @@ void PerfEventRingBuffer::SkipRecord(const perf_event_header& header) {
   WriteRingBufferTail(metadata_page_, new_tail);
 }
 
-void PerfEventRingBuffer::ConsumeRecord(const perf_event_header& header, void* record) {
+void PerfEventRingBuffer::ConsumeRawRecord(const perf_event_header& header, void* record) {
   ReadAtTail(static_cast<uint8_t*>(record), header.size);
   SkipRecord(header);
 }
--- a/OrbitLinuxTracing/PerfEventRingBuffer.h
+++ b/OrbitLinuxTracing/PerfEventRingBuffer.h
@@ -6,9 +6,10 @@
 #define ORBIT_LINUX_TRACING_PERF_RING_BUFFER_H_
 
 #include <linux/perf_event.h>
-#include <stdint.h>
 
 #include <string>
+
+#include "OrbitBase/Logging.h"
 
 namespace LinuxTracing {
 
@@ -30,7 +31,12 @@ class PerfEventRingBuffer {
   bool HasNewData();
   void ReadHeader(perf_event_header* header);
   void SkipRecord(const perf_event_header& header);
-  void ConsumeRecord(const perf_event_header& header, void* record);
+
+  template <typename T>
+  void ConsumeRecord(const perf_event_header& header, T* record) {
+    CHECK(header.size == sizeof(T));
+    ConsumeRawRecord(header, record);
+  }
 
   template <typename T>
   void ReadValueAtOffset(T* value, uint64_t offset) {
@@ -52,8 +58,9 @@ class PerfEventRingBuffer {
   int file_descriptor_ = -1;
   std::string name_;
 
+  // ConsumeRawRecord reads header.size bytes into record buffer and then skips the record.
+  void ConsumeRawRecord(const perf_event_header& header, void* record);
   void ReadAtTail(void* dest, uint64_t count) { return ReadAtOffsetFromTail(dest, 0, count); }
-
   void ReadAtOffsetFromTail(void* dest, uint64_t offset_from_tail, uint64_t count);
 };
 


OrbitLinuxTracing/PerfEventRingBuffer.cpp
```diff
@@ -121,8 +121,11 @@ void PerfEventRingBuffer::SkipRecord(const perf_event_header& header) {
...
 }
 
-void PerfEventRingBuffer::ConsumeRecord(const perf_event_header& header, void* record) {
-  ReadAtTail(static_cast<uint8_t*>(record), header.size);
+void PerfEventRingBuffer::ConsumeRecord(const perf_event_header& header, void* record,
+                                        size_t size) {
+  // Sanity check so that we do not over-read from the ring-buffer and end up with random data.
+  CHECK(header.size >= size);
+  ReadAtTail(static_cast<uint8_t*>(record), size);
   SkipRecord(header);
```
pierricgimmig, line 129: Hi Dimitry, it seems we are reading "sizeof(record)" but then skipping "sizeof(header)", is this normal?

dimitry-: the size of the record includes the header.

dimitry-: oh, sorry misread you comment - yes this is normal. Since we do not always want to read the whole record because of duplicate sample_id at the end.

pierricgimmig: Ah ok I see, thanks.

pierricgimmig (approved):
LGTM, good catch, thanks!

OrbitLinuxTracing/PerfEventRingBuffer.cpp
pierricgimmig: "we do not always want to read the whole **header**" right?

pierricgimmig (approved):

OrbitLinuxTracing/PerfEventRingBuffer.cpp
dimitry-: record, the header contains size information about the whole record, SkipRecord also skips the whole record based on the information from the header.

pierricgimmig: Hmm, this is confusing. header.size >= sizeof(record). We are reading "sizeof(record)" bytes, so we are not always reading the full *header*, no? Then we are calling "SkipRecord" which uses header.size internally, maybe it should be called SkipHeader?

dimitry-: header.size is size of record, the header itself is always fixed size of... some even number of bytes..

pierricgimmig: Ah ok ok, I see.

dpallotti:
I don't understand this change. When do we read more than we have memory for? The point of `ConsumeRecord` is to be only used when we know the full structure of the data and we have memory for it. If not, the corresponding `struct`s should be fixed. Which `struct`s are affected by this?

OrbitLinuxTracing/PerfEventRingBuffer.cpp
```diff
@@ -121,8 +121,11 @@ void PerfEventRingBuffer::SkipRecord(const perf_event_header& header) {
   WriteRingBufferTail(metadata_page_, new_tail);
 }
 
-void PerfEventRingBuffer::ConsumeRecord(const perf_event_header& header, void* record) {
-  ReadAtTail(static_cast<uint8_t*>(record), header.size);
+void PerfEventRingBuffer::ConsumeRecord(const perf_event_header& header, void* record,
+                                        size_t size) {
+  // Sanity check so that we do not over-read from the ring-buffer and end up with random data.
+  CHECK(header.size >= size);
```
dpallotti, line 127: The way `ConsumeRecord` is used, this should check for equality.

dimitry-: yeah, this is what I started with :)

dimitry-: Interesting - I think I did the check wrong yesterday. now it works fine. I will upload the version with more strict check.

OrbitLinuxTracing/PerfEventRingBuffer.h
```diff
@@ -30,7 +31,13 @@ class PerfEventRingBuffer {
...
   void ReadHeader(perf_event_header* header);
   void SkipRecord(const perf_event_header& header);
-  void ConsumeRecord(const perf_event_header& header, void* record);
+
+  template <typename T>
+  void ConsumeRecord(const perf_event_header& header, T* record) {
+    CHECK(header.size == sizeof(T));
+    ConsumeRecord(header, record, sizeof(T));
+  }
+  void ConsumeRecord(const perf_event_header& header, void* record, size_t size);
```
dpallotti, line 40: This should be private. If you need to read from the ring buffer without reading the whole record you are not really "consuming the record", you should use `ReadValueAtOffset` and `SkipRecord` as it's already done in many other places.

OrbitLinuxTracing/PerfEventRingBuffer.cpp
dpallotti: This check should be updated, or removed as it's already done by the public `ConsumeRecord`.

OrbitLinuxTracing/PerfEventRingBuffer.h
dimitry-: done.

OrbitLinuxTracing/PerfEventRingBuffer.cpp
dimitry-: updated

OrbitLinuxTracing/PerfEventRingBuffer.cpp
```diff
@@ -121,8 +121,10 @@ void PerfEventRingBuffer::SkipRecord(const perf_event_header& header) {
   WriteRingBufferTail(metadata_page_, new_tail);
 }
 
-void PerfEventRingBuffer::ConsumeRecord(const perf_event_header& header, void* record) {
-  ReadAtTail(static_cast<uint8_t*>(record), header.size);
+void PerfEventRingBuffer::ConsumeRecord(const perf_event_header& header, void* record,
+                                        size_t size) {
+  DCHECK(header.size == size);
+  ReadAtTail(static_cast<uint8_t*>(record), size);
```
dpallotti, line 127: Now `size` is passed just for the (`D`)`CHECK`. Can we just remove it and use the implementation we had before, with the difference that now it is not called from outside (the public method is templated and does the check, and this check is now redundant)?

OrbitLinuxTracing/PerfEventRingBuffer.h
```diff
@@ -55,6 +61,8 @@ class PerfEventRingBuffer {
   void ReadAtTail(void* dest, uint64_t count) { return ReadAtOffsetFromTail(dest, 0, count); }
 
   void ReadAtOffsetFromTail(void* dest, uint64_t offset_from_tail, uint64_t count);
+
+  void ConsumeRecord(const perf_event_header& header, void* record, size_t size);
```
dpallotti, line 65: Could you move it up before `ReadAtTail` (it keeps the current order and reflects the order in the .cpp file)?

OrbitLinuxTracing/PerfEventRingBuffer.cpp
dpallotti: And also maybe just rename the private method to `ConsumeRecordInternal` or something like that to avoid confusion with overloading.

dpallotti: So again, can we just remove the `size` argument now as it's only used for the `DCHECK`, which is redundant as this is a private method and the caller already does the `CHECK`?

dimitry-: I would like to keep it here.

dpallotti: Ok, then use a normal `CHECK` otherwise it's useless, the service is run in debug mode too rarely (performance reason + it's hard to debug on an instance).

dpallotti (approved):

OrbitLinuxTracing/PerfEventRingBuffer.cpp
dimitry-: well, on second look it does look wired. I guess I will add a comment here and remove the check and size argument.

PR #1576, opened 2020-12-11
Clang-format DispatchTableTest.cpp

pierricgimmig:


pierricgimmig:
@florian-kuebler , running "./run_clang_format.sh" on the repo modifies this file locally. It seems the presubmit check doesn't like the new formatting however. Do you understand what's going on?

florian-kuebler:
Hey Pierric, try to use `clang-format-9` locally (you might need to install it first). There are a couple of changes between the versions, and this is the version we use upstream in the CI.

pierricgimmig:
> Hey Pierric, try to use `clang-format-9` locally (you might need to install it first). There are a couple of changes between the versions, and this is the version we use upstream in the CI.

Cool, thanks

PR #1578, opened 2020-12-11
Fix crash on capture abort on exit

dpallotti:
Change `TryAbort` to `AbortCaptureAndWait` and wait (for a limited time) for the
`CaptureClient::Capture` thread to change `state_` to `State::kStopped`, so that
`client_context_` has been cleaned up.
This is mostly done by improving synchronization in the class, adding a mutex
for `client_context_` and `reader_writer_`.

Bug: http://b/175211279

Test: locally and with Trata both taking "normal" captures and aborting captures
on exit.

### DIFF
--- a/OrbitCaptureClient/CaptureClient.cpp
+++ b/OrbitCaptureClient/CaptureClient.cpp
@@ -24,7 +24,7 @@ using orbit_grpc_protos::CaptureRequest;
 using orbit_grpc_protos::CaptureResponse;
 using orbit_grpc_protos::TracepointInfo;
 
-static CaptureOptions::InstrumentedFunction::FunctionType IntrumentedFunctionTypeFromOrbitType(
+static CaptureOptions::InstrumentedFunction::FunctionType InstrumentedFunctionTypeFromOrbitType(
     FunctionInfo::OrbitType orbit_type) {
   switch (orbit_type) {
     case FunctionInfo::kOrbitTimerStart:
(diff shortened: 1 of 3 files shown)


OrbitCaptureClient/CaptureClient.cpp
```diff
@@ -197,34 +223,50 @@ bool CaptureClient::StopCapture() {
...
+      client_context_->TryCancel();  // reader_writer_->Read in Capture should then fail
+      try_cancel_called = true;
+    }
   }
 
-  CHECK(client_context_ != nullptr);
+  if (!try_cancel_called) {
+    LOG("AbortCaptureAndWait ignored: no ClientContext to TryCancel");
+    return false;
+  }
```
dimitry-, line 244: maybe move it up and do early return on L233? is there a reason for avoiding read lock while LOGging?

dimitry-: (and you can drop try_cancel_called flag)

dimitry- (approved):
otherwise lgtm

OrbitCaptureClient/CaptureClient.cpp
dpallotti: Yeah, in general I try to print logs when not holding a mutex when possible (to do as little as possible when holding the mutex), but I guess it doesn't really matter, and the code improves quite a bit from changing this, so done.

PR #1579, opened 2020-12-11
Instrumentation to monitor number of capture data bytes sent

ronaldfw:
Add a small amount of instrumentation in OrbitService to monitor the
number of bytes of capture data sent during a capture. The average
number of bytes per CaptureEvent message sent is already recorded for
every batch of events sent and shown as a live graph when --devmode
is enabled.

Tested: Manual test.
Bug: http://b/175383580

### DIFF
--- a/OrbitService/CaptureServiceImpl.cpp
+++ b/OrbitService/CaptureServiceImpl.cpp
@@ -91,30 +91,49 @@ class GrpcCaptureEventSender final : public CaptureEventSender {
     CHECK(reader_writer_ != nullptr);
   }
 
+  ~GrpcCaptureEventSender() override {
+    LOG("Total number of events sent: %lu", total_number_of_events_sent_);
+    LOG("Total number of bytes sent: %lu", total_number_of_bytes_sent_);
+    float average_bytes =
+        static_cast<float>(total_number_of_bytes_sent_) / total_number_of_events_sent_;
+    LOG("Average number of bytes per event: %.2f", average_bytes);
+  }
+
   void SendEvents(std::vector<orbit_grpc_protos::CaptureEvent>&& events) override {
     ORBIT_SCOPE_FUNCTION;
-    ORBIT_UINT64("Number of sent buffered events", events.size());
+    ORBIT_UINT64("Number of buffered events sent", events.size());
     if (events.empty()) {
       return;
     }
 
     constexpr uint64_t kMaxEventsPerResponse = 10'000;
+    uint64_t number_of_bytes_sent = 0;
     CaptureResponse response;
     for (CaptureEvent& event : events) {
       // We buffer to avoid sending countless tiny messages, but we also want to
       // avoid huge messages, which would cause the capture on the client to jump
       // forward in time in few big steps and not look live anymore.
       if (response.capture_events_size() == kMaxEventsPerResponse) {
+        number_of_bytes_sent += response.ByteSizeLong();
         reader_writer_->Write(response);
         response.clear_capture_events();
       }
       response.mutable_capture_events()->Add(std::move(event));
     }
+    number_of_bytes_sent += response.ByteSizeLong();
     reader_writer_->Write(response);
+
+    float average_bytes = static_cast<float>(number_of_bytes_sent) / events.size();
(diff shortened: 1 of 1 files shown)


OrbitService/CaptureServiceImpl.cpp
```diff
@@ -91,30 +91,48 @@ class GrpcCaptureEventSender final : public CaptureEventSender {
     CHECK(reader_writer_ != nullptr);
   }
 
+  ~GrpcCaptureEventSender() {
+    LOG("Total number of events sent: %lu", total_number_of_events_sent_);
+    LOG("Total number of bytes sent: %lu", total_number_of_bytes_sent_);
+    LOG("Average number of bytes per event: %.2f",
+        static_cast<float>(total_number_of_bytes_sent_) / total_number_of_events_sent_);
```
dimitry-, line 98: I do not really know: is it ok to divide float to 0 in c++?

dimitry- (approved):
one question, otherwise lgtm

OrbitService/CaptureServiceImpl.cpp
ronaldfw: Good catch, that would cause a divide by zero exception. Will fix.

ronaldfw: Done.

OrbitService/CaptureServiceImpl.cpp
```diff
@@ -91,30 +91,52 @@ class GrpcCaptureEventSender final : public CaptureEventSender {
     CHECK(reader_writer_ != nullptr);
   }
 
+  ~GrpcCaptureEventSender() {
```
dpallotti, line 94: Could you add `override` to this (clang-tidy)?
```suggestion
  ~GrpcCaptureEventSender() override {
```

OrbitService/CaptureServiceImpl.cpp
```diff
@@ -91,30 +91,52 @@ class GrpcCaptureEventSender final : public CaptureEventSender {
...
+    LOG("Total number of events sent: %lu", total_number_of_events_sent_);
+    LOG("Total number of bytes sent: %lu", total_number_of_bytes_sent_);
+    float average_bytes =
+        total_number_of_events_sent_ == 0
+            ? 0.0f
+            : static_cast<float>(total_number_of_bytes_sent_) / total_number_of_events_sent_;
```
dpallotti, line 100: Dividing a float by zero doesn't produce any exception, it just gives NaN (technically more correct than 0...).

OrbitService/CaptureServiceImpl.cpp
ronaldfw: Done.

OrbitService/CaptureServiceImpl.cpp
```diff
@@ -91,30 +91,52 @@ class GrpcCaptureEventSender final : public CaptureEventSender {
...
+    ORBIT_FLOAT("Average bytes per CaptureEvent", average_bytes);
+    total_number_of_events_sent_ += events.size();
+    total_number_of_bytes_sent_ += number_of_bytes_sent;
   }
 
  private:
   grpc::ServerReaderWriter<CaptureResponse, CaptureRequest>* reader_writer_;
+
+  uint64_t total_number_of_events_sent_ = 0;
+  uint64_t total_number_of_bytes_sent_ = 0;
```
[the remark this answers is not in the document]
ronaldfw: No, I don't need it, but would prefer to keep it.

OrbitService/CaptureServiceImpl.cpp
dpallotti: (Well, NaN if the dividend is also 0, plus or minus infinity otherwise.)

ronaldfw: Ah, of course. Was thinking of integer division for a moment. Reverted the change (which is also more concise). Thanks a lot!

OrbitService/CaptureServiceImpl.cpp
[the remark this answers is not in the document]
dpallotti: Then I don't really understand the reasoning, but ok, it doesn't matter.

dpallotti (approved):

PR #1580, opened 2020-12-11
Remove unused DebugUtils.h

pierricgimmig:
Having code compiled out by default is not a good idea. In it's current
form, those utils are too painful to use in day-to-day development as
they require annoying local modifications. Let's remove it and possibly
revive in some always-on form.


dimitry- (approved):

OrbitBase/include/OrbitBase/DebugUtils.h
```diff
@@ -1,24 +0,0 @@
...
-
-#ifndef ORBIT_BASE_DEBUG_UTILS_H_
-#define ORBIT_BASE_DEBUG_UTILS_H_
-
-#include <string>
-
-#include "Logging.h"
-#include "OrbitBase/Profiling.h"
-
-#define DEBUG_UTILS_ENABLED 0  // Enable locally only.
```
dpallotti, line 13: I'm fine with removing them altogether (still kind of against having them back as always-on as written elsewhere, but that's another story). But if the problem is local modifications, isn't there a way to remove this `#define` and allow to set `-DDEBUG_UTILS_ENABLED=1` in the IDE? Or do conan profiles override everything? The point of code always compiled out by default obviously remains though.

dpallotti (approved):

OrbitBase/include/OrbitBase/DebugUtils.h
pierricgimmig: Interesting idea, but yeah, same issue with code compiled out by default. Maybe a presubmit check that makes sure the code doesn't contain debug macros could work.

PR #1583, opened 2020-12-14
Remove flaky assertion from test.

danielfenner:
Bug: http://b/175527472
Test: Ran locally.

### DIFF
--- a/contrib/automation_tests/orbit_top_down.py
+++ b/contrib/automation_tests/orbit_top_down.py
@@ -75,12 +75,7 @@ def main(argv):
 
     logging.info('Re-listing items of the top-down view...')
     tree_items = main_wnd.TreeView.children(control_type='TreeItem')
-    row_count_after_expansion = len(tree_items) / TOP_DOWN_ROW_CELL_COUNT
 
-    if row_count_after_expansion != row_count_before_expansion + 2:
-        raise RuntimeError(
-            'First item of the top-down view doesn\'t have exactly two children'
-        )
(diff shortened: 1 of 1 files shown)


contrib/automation_tests/orbit_top_down.py
```diff
@@ -75,12 +75,7 @@ def main(argv):
 
     logging.info('Re-listing items of the top-down view...')
     tree_items = main_wnd.TreeView.children(control_type='TreeItem')
-    row_count_after_expansion = len(tree_items) / TOP_DOWN_ROW_CELL_COUNT
 
-    if row_count_after_expansion != row_count_before_expansion + 2:
-        raise RuntimeError(
-            'First item of the top-down view doesn\'t have exactly two children'
-        )
```
dpallotti, line 83: What do you think about adjusting this instead of removing it to keep the "at least two children" assertion explicit? E.g.:

```
     row_count_after_expansion = len(tree_items) / TOP_DOWN_ROW_CELL_COUNT	


    if row_count_after_expansion < row_count_before_expansion + 2:	
        raise RuntimeError(	
            'First item of the top-down view doesn\'t have at least two children'	
        )
```

dpallotti (approved):
What do you think about keeping the assertion explicit?

contrib/automation_tests/orbit_top_down.py
danielfenner: Changed as suggested

PR #1585, opened 2020-12-14
Add VulkanLayerProducerImpl to OrbitVulkanLayer

dpallotti:
#### Move FakeProducerSideService.h to OrbitProducer target
This is so that in can be used in tests by other modules (specifically,
`OrbitVulkanLayer`).
#### Add VulkanLayerProducerImpl to OrbitVulkanLayer
This class enables the communication with `OrbitService`'s `ProducerSideService`
by making use of `LockFreeBufferCaptureEventProducer`.

Bug: http://b/174028631


OrbitVulkanLayer/VulkanLayerProducerImplTest.cpp
```diff
@@ -0,0 +1,512 @@
...
+
+    std::shared_ptr<grpc::Channel> channel =
+        fake_server_->InProcessChannel(grpc::ChannelArguments{});
+
+    producer_.emplace();
+    producer_->SetCaptureStatusListener(&mock_listener_);
+    producer_->BringUp(channel);
+
+    // Leave some time for the ReceiveCommandsAndSendEvents RPC to actually happen.
+    std::this_thread::sleep_for(std::chrono::milliseconds(50));
```
florian-kuebler, line 40: Do you see a way to work around those sleeps?

OrbitVulkanLayer/VulkanLayerProducerImplTest.cpp
```diff
@@ -0,0 +1,512 @@
...
+constexpr std::chrono::duration kWaitMessagesSentDuration = std::chrono::milliseconds(25);
+
+}  // namespace
+
+TEST_F(VulkanLayerProducerImplTest, IsCapturingAndListener) {
+  EXPECT_FALSE(producer_->IsCapturing());
+
+  EXPECT_CALL(mock_listener_, OnCaptureStart).Times(1);
+  fake_service_->SendStartCaptureCommand();
+  std::this_thread::sleep_for(kWaitMessagesSentDuration);
```
florian-kuebler, line 92: Could you work around those sleeps, e.g. by setting a flag in the OnCaptureStart/Stop methods?

florian-kuebler (commented):
Thanks. LGTM. I am only a bit worried for the tests, to become flaky with all the timeouts going on. Maybe you can workaround those (at least some), by setting state using the listener.

OrbitVulkanLayer/VulkanLayerProducerImplTest.cpp
dpallotti: I know it's far from ideal, but using `OnCaptureStart` as an indication to continue kind of interferes with the tests (as that's what we are testing), and if the expectation is not satisfied it means that the test will hang instead of failing. By the way, I'm doing the same in the tests for `ProducerSideService` and `CaptureEventProducer`.

florian-kuebler: Yeah you are right, this is also not ideal. I'd still slightly feel it would be better. Two thoughts on this:
1. By keeping the EXPECT_CALL regarding the OnCaptureStart, you would keep test testing this behaviour. You may need to insert a "validate and reset". 
2. You could still add a timeout, to prevent from a hold. That timeout may be much larger, such that it is at least less likely that it failed falsely.

florian-kuebler: But also note, I am not too strong about this.

dpallotti: I see your point. If we need to tackle this, we need to do it together with the tests for `ProducerSideService`, `CaptureEventProducer` and `LockFreeBufferCaptureEventProducer` and give it some good thought. That's because there are quite big drawbacks:

- Every one-line sleep is replaced with much more complicated code (action on listener method, wait-notify, timeout), which complicates the tests even more and makes them even harder to read, as they are already hundreds of lines of code.
- We would need to be sure that every single sleep could be remove. Otherwise the tests would become inconsistent, as there would be two ways to implement a wait. That would mean that the tests would be harder to read, without actually solving the problem fully.

florian-kuebler: Agree. I am fine with tackling this as a follow up. I still think that we can set something up by using the Test class and the MockListener, so that the code complexity will not increase (that much) in the actual tests.
I've done something similar (but in a different context), with the IsCapturing in the SubmissionTrackerTest, before we had the Listener.

dpallotti: Let me at least create a bug at least.

florian-kuebler: Thanks

dpallotti: http://b/175669610

florian-kuebler (approved):
Thanks. Let's follow up on this.

PR #1586, opened 2020-12-14
Unit tests and fix for GpuTracepointEventProcessor

dpallotti:


OrbitLinuxTracing/GpuTracepointEventProcessorTest.cpp
```diff
@@ -0,0 +1,589 @@
...
+
+AmdgpuCsIoctlPerfEvent MakeFakeAmdgpuCsIoctlPerfEvent(pid_t tid, uint64_t timestamp_ns,
+                                                      uint32_t context, uint32_t seqno,
+                                                      const std::string& timeline) {
+  AmdgpuCsIoctlPerfEvent event{
+      static_cast<uint32_t>(sizeof(amdgpu_cs_ioctl_tracepoint) + timeline.length() + 1)};
+  event.ring_buffer_record.sample_id.tid = tid;
+  event.ring_buffer_record.sample_id.time = timestamp_ns;
+  reinterpret_cast<amdgpu_cs_ioctl_tracepoint*>(event.tracepoint_data.get())->context = context;
+  EXPECT_EQ(event.GetContext(), context);
```
florian-kuebler, line 48: This seems to test the `AmdgpuCsIoctrlPerfEvent`, which is not really related to the `GpuTracepointEventProcessor`, right? I am unsure if it is right to have those `EXPECT`s here. 
If this is more about to validate the results from the assignments after reinterpret_cast, CHECK might be the better choice. WDYT?

OrbitLinuxTracing/GpuTracepointEventProcessorTest.cpp
```diff
@@ -0,0 +1,589 @@
...
+  expected_gpu_job.set_timeline(std::move(timeline));
+  expected_gpu_job.set_depth(depth);
+  expected_gpu_job.set_amdgpu_cs_ioctl_time_ns(amdgpu_cs_ioctl_time_ns);
+  expected_gpu_job.set_amdgpu_sched_run_job_time_ns(amdgpu_sched_run_job_time_ns);
+  expected_gpu_job.set_gpu_hardware_start_time_ns(gpu_hardware_start_time_ns);
+  expected_gpu_job.set_dma_fence_signaled_time_ns(dma_fence_signaled_time_ns);
+  return expected_gpu_job;
+}
+
+void ExpectGpuJobEq(const orbit_grpc_protos::GpuJob& actual,
```
florian-kuebler, line 116: I am fine with this. But have you considered making this method a matcher? Might make it even clearer on the caller.

OrbitLinuxTracing/GpuTracepointEventProcessorTest.cpp
```diff
@@ -0,0 +1,589 @@
...
+  GpuTracepointEventProcessor processor;
+  processor.PushEvent(
+      MakeFakeAmdgpuCsIoctlPerfEvent(kTid, kTimestampA, kContext, kSeqno, kTimeline));
+  processor.PushEvent(MakeFakeAmdgpuSchedRunJobPerfEvent(kTimestampB, kContext, kSeqno, kTimeline));
+  EXPECT_DEATH(processor.PushEvent(
+                   MakeFakeDmaFenceSignaledPerfEvent(kTimestampD, kContext, kSeqno, kTimeline)),
+               "listener_ != nullptr");
+}
+
+TEST_F(GpuTracepointEventProcessorTest, OneJob) {
```
florian-kuebler, line 149: How about "WillNotifyListenerWithOneGpuJob" or something more descriptive?

OrbitLinuxTracing/GpuTracepointEventProcessorTest.cpp
```diff
@@ -0,0 +1,589 @@
...
+  EXPECT_CALL(mock_listener_, OnGpuJob).Times(1).WillOnce(::testing::SaveArg<0>(&actual_gpu_job));
+  processor_.PushEvent(
+      MakeFakeAmdgpuSchedRunJobPerfEvent(kTimestampB, kContext, kSeqno, kTimeline));
+  processor_.PushEvent(
+      MakeFakeAmdgpuCsIoctlPerfEvent(kTid, kTimestampA, kContext, kSeqno, kTimeline));
+  processor_.PushEvent(MakeFakeDmaFenceSignaledPerfEvent(kTimestampD, kContext, kSeqno, kTimeline));
+  ExpectGpuJobEq(actual_gpu_job, expected_gpu_job);
+}
+
+TEST_F(GpuTracepointEventProcessorTest, MismatchingContext) {
```
florian-kuebler, line 215: How about "WillNotCreateAGpuJobOnMismatchingContext"?

OrbitLinuxTracing/GpuTracepointEventProcessorTest.cpp
```diff
@@ -0,0 +1,589 @@
...
+
+  EXPECT_CALL(mock_listener_, OnGpuJob).Times(0);
+  processor_.PushEvent(
+      MakeFakeAmdgpuCsIoctlPerfEvent(kTid, kTimestampA, kContext + 1, kSeqno, kTimeline));
+  processor_.PushEvent(
+      MakeFakeAmdgpuSchedRunJobPerfEvent(kTimestampB, kContext, kSeqno, kTimeline));
+  processor_.PushEvent(MakeFakeDmaFenceSignaledPerfEvent(kTimestampD, kContext, kSeqno, kTimeline));
+}
+
+TEST_F(GpuTracepointEventProcessorTest, MismatchingSeqno) {
```
florian-kuebler, line 232: See above.

OrbitLinuxTracing/GpuTracepointEventProcessorTest.cpp
```diff
@@ -0,0 +1,589 @@
...
+
+  EXPECT_CALL(mock_listener_, OnGpuJob).Times(0);
+  processor_.PushEvent(
+      MakeFakeAmdgpuCsIoctlPerfEvent(kTid, kTimestampA, kContext, kSeqno, kTimeline));
+  processor_.PushEvent(
+      MakeFakeAmdgpuSchedRunJobPerfEvent(kTimestampB, kContext, kSeqno + 1, kTimeline));
+  processor_.PushEvent(MakeFakeDmaFenceSignaledPerfEvent(kTimestampD, kContext, kSeqno, kTimeline));
+}
+
+TEST_F(GpuTracepointEventProcessorTest, MismatchingTimeline) {
```
florian-kuebler, line 249: And here.

OrbitLinuxTracing/GpuTracepointEventProcessorTest.cpp
```diff
@@ -0,0 +1,589 @@
...
+
+  orbit_grpc_protos::GpuJob expected_gpu_job1 =
+      MakeGpuJob(kTid, kContext1, kSeqno, kTimeline, 0, kTimestampA1, kTimestampB1, kTimestampC1,
+                 kTimestampD1);
+  orbit_grpc_protos::GpuJob expected_gpu_job2 =
+      MakeGpuJob(kTid, kContext2, kSeqno, kTimeline, 0, kTimestampA2, kTimestampB2, kTimestampC2,
+                 kTimestampD2);
+  orbit_grpc_protos::GpuJob actual_gpu_job1;
+  orbit_grpc_protos::GpuJob actual_gpu_job2;
+  EXPECT_CALL(mock_listener_, OnGpuJob)
```
florian-kuebler, line 291: I am not 100% in the GpuTracepointEventProcessor, but will it guarantee in order outputs? If not consider saving the jobs in a vector, make the "expect" method returning a matcher, and use UnorderedElementsAre.

OrbitLinuxTracing/GpuTracepointEventProcessorTest.cpp
```diff
@@ -0,0 +1,589 @@
...
+  EXPECT_CALL(mock_listener_, OnGpuJob).Times(0);
+  processor_.PushEvent(
+      MakeFakeAmdgpuCsIoctlPerfEvent(kTid, kTimestampA, kContext, kSeqno, kTimeline));
+  processor_.PushEvent(
+      MakeFakeAmdgpuSchedRunJobPerfEvent(kTimestampB, kContext, kSeqno, kTimeline));
+  processor_.PushEvent(
+      MakeFakeDmaFenceSignaledPerfEvent(kTimestampD, kContext, kSeqno, kTimeline + "1"));
+}
+
+TEST_F(GpuTracepointEventProcessorTest, TwoNonOverlappingJobsDifferingByContext) {
```
florian-kuebler, line 267: Again, it would be nice to get from the name an intuition about what to expect.

florian-kuebler (commented):
Thanks very much for this. It is very nice to see the coverage increasing.
I am a bit concerned about the test names. I would prefer to reflect them not only the situation but also what behaviour to expect.

Also, I think the tests might be slightly better when using a Matcher for the GpuJob equality.

OrbitLinuxTracing/GpuTracepointEventProcessorTest.cpp
dpallotti: Yes, I had originally written it as a matcher. Since you don't have the individual `EXPECT_EQ`s the output in case of failure is completely useless, because you don't know where the difference is.

OrbitLinuxTracing/GpuTracepointEventProcessorTest.cpp
dpallotti: You are right, all these `EXPECT` made me forget of `CHECK`. Done.

OrbitLinuxTracing/GpuTracepointEventProcessorTest.cpp
dpallotti: The `GpuJob`s are created and sent to the `Listener` once all the three tracepoint events have been seen, so we can totally the order.

OrbitLinuxTracing/GpuTracepointEventProcessorTest.cpp
dpallotti: I want to avoid mentioning "notifying the listener" in every name as it doesn't really add any information, that's just the basic of how we test the class, using whether the listener is notified or not. Similarly for "gpu job created", what does it add? I'll try to expand the names a bit where applicable.

OrbitLinuxTracing/GpuTracepointEventProcessorTest.cpp
dpallotti: Changed to `NoJobBecauseOfMismatchingContext`.

OrbitLinuxTracing/GpuTracepointEventProcessorTest.cpp
dpallotti: Changed to `NoJobBecauseOfMismatchingSeqno`.

OrbitLinuxTracing/GpuTracepointEventProcessorTest.cpp
dpallotti: Changed to `NoJobBecauseOfMismatchingTimeline`.

OrbitLinuxTracing/GpuTracepointEventProcessorTest.cpp
dpallotti: I added "WithSameDepth" as that's the main factor here. Left out anything like "listener notified" or "jobs created" as they add very little.

OrbitLinuxTracing/GpuTracepointEventProcessorTest.cpp
florian-kuebler: You could use the `AllOf` and `Fields`/`Property` Matcher. That should preserve the output.
See [here](https://chromium.googlesource.com/external/github.com/google/googletest/+/HEAD/googlemock/docs/cook_book.md#validating-a-member-of-an-object).

```
Matcher<GpuJob> GpuJobEq(
    const GpuJob& expected) {
  return AllOf(
      Property(&GpuJob:: tid, expected. tid()),
      Field(&GpuJob:: context, expected. context()));
}
```

OrbitLinuxTracing/GpuTracepointEventProcessorTest.cpp
florian-kuebler: Sounds good. In general, I was just missing the "what to expect" in the names.

OrbitLinuxTracing/GpuTracepointEventProcessorTest.cpp
florian-kuebler: Okay, thanks.

OrbitLinuxTracing/GpuTracepointEventProcessorTest.cpp
florian-kuebler: Sounds good.

florian-kuebler (approved):
Thanks for the changes. Looks good to me. Feel free to implement the `Matcher` or keep it as it is.

OrbitLinuxTracing/GpuTracepointEventProcessor.cpp
```diff
@@ -69,7 +69,7 @@ void GpuTracepointEventProcessor::CreateGpuExecutionEventIfComplete(const Key& k
   // at the timestamp of scheduling the current job, we push the start
   // time for starting on the hardware back.
   if (!timeline_to_latest_dma_signal_.contains(timeline)) {
-    timeline_to_latest_dma_signal_.emplace(timeline, dma_it->second.timestamp_ns);
```
ronaldfw, line 73: I would add a short comment why setting to zero here is fine. 

Suggestion:
"When there is not yet an entry in timeline_to_latest_dma_signal_ for the current timeline, this means that no previous GPU job has been executed on this timeline during our capture. We just have to set a timestamp here that lies before any event on the timeline to make sure that the hardware start time below is set correctly, so "0" is a perfectly valid choice."

OrbitLinuxTracing/GpuTracepointEventProcessorTest.cpp
dpallotti: Interesting, I'll have a look. Then I use the matcher in the `EXPECT_CALL` directly?

ronaldfw (commented):
Thanks so much for creating a unit test for this! @florian-kuebler is already reviewing the test code, I only looked at the logic change to make sure I understand what the problem was. Only a small suggestion to include a comment there why setting the timestamp to 0 there is fine (as the choice is somewhat arbitrary).

OrbitLinuxTracing/GpuTracepointEventProcessor.cpp
dpallotti: Done and thanks for giving a suggestion (just changed a couple of words).

OrbitLinuxTracing/GpuTracepointEventProcessorTest.cpp
florian-kuebler: Either in the `EXPECT_CALL` directly, or via `EXPECT_THAT`.

dpallotti: So I'm trying with the matcher and I kind of have the same problem, that is the output is horrible and it doesn't tell me what field is wrong, just that no call was registered as the parameter didn't match.

For example, by making a test fail on purpose:
```
unknown file: Failure

Unexpected mock function call - returning directly.
    Function call: OnGpuJob(88-byte object <A8-75 22-7D 1D-56 00-00 00-00 00-00 00-00 00-00 00-00 00-00 2A-00 00-00 01-00 00-00 0A-00 00-00 64-00 00-00 00-00 00-00 C8-00 00-00 00-00 00-00 C8-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00 2C-01 00-00 00-00 00-00 B0-54 D1-7D 1D-56 00-00 00-00 00-00 05-00 00-00>)
Google Mock tried the following 1 expectation, but it didn't match:

/usr/local/google/home/dpallotti/CLionProjects/orbitprofiler/OrbitLinuxTracing/GpuTracepointEventProcessorTest.cpp:166: EXPECT_CALL(mock_listener_, OnGpuJob(GpuJobEq(expected_gpu_job)))...
  Expected arg #0: (is an object whose given property is equal to 42) and (is an object whose given property is equal to 1) and (is an object whose given property is equal to 10) and (is an object whose given property is equal to "timeline") and (is an object whose given property is equal to 1) and (is an object whose given property is equal to 100) and (is an object whose given property is equal to 200) and (is an object whose given property is equal to 200) and (is an object whose given property is equal to 300)
           Actual: 88-byte object <A8-75 22-7D 1D-56 00-00 00-00 00-00 00-00 00-00 00-00 00-00 2A-00 00-00 01-00 00-00 0A-00 00-00 64-00 00-00 00-00 00-00 C8-00 00-00 00-00 00-00 C8-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00 2C-01 00-00 00-00 00-00 B0-54 D1-7D 1D-56 00-00 00-00 00-00 05-00 00-00>, whose given property is 0 (of type int)
         Expected: to be called once
           Actual: never called - unsatisfied and active
/usr/local/google/home/dpallotti/CLionProjects/orbitprofiler/OrbitLinuxTracing/GpuTracepointEventProcessorTest.cpp:166: Failure
Actual function call count doesn't match EXPECT_CALL(mock_listener_, OnGpuJob(GpuJobEq(expected_gpu_job)))...
         Expected: to be called once
           Actual: never called - unsatisfied and active
```

vs a much clearer

```
/usr/local/google/home/dpallotti/CLionProjects/orbitprofiler/OrbitLinuxTracing/GpuTracepointEventProcessorTest.cpp:123: Failure
Expected equality of these values:
  actual.depth()
    Which is: 0
  expected.depth()
    Which is: 1
```

dpallotti: Ok, I guess I need to use `EXPECT_THAT` then. I feel like that doesn't really add anything to what was there before, but as I'm already here...

dpallotti: Example output:
```
/usr/local/google/home/dpallotti/CLionProjects/orbitprofiler/OrbitLinuxTracing/GpuTracepointEventProcessorTest.cpp:173: Failure
Value of: actual_gpu_job
Expected: (is an object whose given property is equal to 42) and (is an object whose given property is equal to 1) and (is an object whose given property is equal to 10) and (is an object whose given property is equal to "timeline") and (is an object whose given property is equal to 1) and (is an object whose given property is equal to 100) and (is an object whose given property is equal to 200) and (is an object whose given property is equal to 200) and (is an object whose given property is equal to 300)
  Actual: 88-byte object <A8-55 09-54 6A-55 00-00 00-00 00-00 00-00 00-00 00-00 00-00 2A-00 00-00 01-00 00-00 0A-00 00-00 64-00 00-00 00-00 00-00 C8-00 00-00 00-00 00-00 C8-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00 2C-01 00-00 00-00 00-00 A0-39 4D-55 6A-55 00-00 00-00 00-00 05-00 00-00> (of type orbit_grpc_protos::GpuJob), whose given property is 0 (of type int)
```

So again, it doesn't tell me what field is different. The advantage though is that it tells me the correct line in the source code. So I don't know, pros and cons for both solutions.

florian-kuebler: Yeah, probably the EXPECT_CALL will not tell you this, right. You can specify Strings for the `Field` and `Property` matchers. When I wrote the suggestion, I was mostly thinking of not testing the order of events, using `UnorderedElementsAre`. But as that won't be used anyways, I agree that there is not much benefit here. 

I leave you with the decision.

OrbitLinuxTracing/GpuTracepointEventProcessor.cpp
ronaldfw: Looks great, thanks!

OrbitLinuxTracing/GpuTracepointEventProcessorTest.cpp
dpallotti: I'm keeping the `EXPECT_THAT` with the matcher, it might be useful to have the matcher around. I also added the strings for the names of the properties, they add a bit of value instead of getting "given property". It would be better to be told what property causes the fail but oh well, at least the line number is correct with this.

florian-kuebler: Sounds good.

PR #1589, opened 2020-12-15
Fix the deselection behavior in the data view panel

vickyliu-go4it:
This PR fixes the following misleading deselection behaviours in the data view panels:
* When clicking the empty area in the live tab, sampling tab or callstack tab, the previous selection stays but the highlight disappears. 
* When deselecting a function call in the capture window, or if we change to select another function that cannot be found in the live tab, the previous selection in the live tab stays the same.
* When deselecting a function in the live tab (i.e., click the empty area), the highlighted function calls in the capture window stay the same. 

We now allow the user to deselect functions in data view panel by clicking the empty area. And we fix the above problems by:
* Deselecting previous selections when clicking the empty area in the data view panels.
* Clearing the highlighting of function calls in the capture window if we click the empty area in the live tab.
* Clearing the selection in the live tab if we deselect a function call in the capture window or if we change to select a function which cannot be found in the live tab.

### DIFF
--- a/OrbitQt/orbitlivefunctions.cpp
+++ b/OrbitQt/orbitlivefunctions.cpp
@@ -118,12 +118,13 @@ void OrbitLiveFunctions::Reset() {
   all_events_iterator_->DisableButtons();
 }
 
-void OrbitLiveFunctions::onRowSelected(int row) {
-  CHECK(row >= 0);
+void OrbitLiveFunctions::OnRowSelected(std::optional<int> row) {
   ui->data_view_panel_->GetTreeView()->SetIsInternalRefresh(true);
   QItemSelectionModel* selection = ui->data_view_panel_->GetTreeView()->selectionModel();
-  QModelIndex idx = ui->data_view_panel_->GetTreeView()->GetModel()->CreateIndex(row, 0);
-  selection->select(idx, QItemSelectionModel::ClearAndSelect | QItemSelectionModel::Rows);
-  selection->setCurrentIndex(idx, QItemSelectionModel::Rows);
+  QModelIndex index;
+  if (row.has_value()) {
+    index = ui->data_view_panel_->GetTreeView()->GetModel()->CreateIndex(row.value(), 0);
+  }
+  selection->select(index, QItemSelectionModel::ClearAndSelect | QItemSelectionModel::Rows);
   ui->data_view_panel_->GetTreeView()->SetIsInternalRefresh(false);
 }
--- a/OrbitQt/orbitlivefunctions.h
+++ b/OrbitQt/orbitlivefunctions.h
@@ -33,7 +33,7 @@ class OrbitLiveFunctions : public QWidget {
                   bool is_main_instance = true);
   void Refresh();
   void OnDataChanged();
-  void onRowSelected(int row);
+  void OnRowSelected(std::optional<int> row);
   void Reset();
   void SetFilter(const QString& a_Filter);
   void AddIterator(uint64_t id, orbit_client_protos::FunctionInfo* function);
--- a/OrbitQt/orbitmainwindow.cpp
+++ b/OrbitQt/orbitmainwindow.cpp
@@ -778,16 +778,16 @@ void OrbitMainWindow::RestoreDefaultTabLayout() {
 }
 
 void OrbitMainWindow::OnTimerSelectionChanged(const orbit_client_protos::TimerInfo* timer_info) {
-  if (!timer_info) return;
-  uint64_t function_address = timer_info->function_address();
-  const auto live_functions_controller = ui->liveFunctions->GetLiveFunctionsController();
-  CHECK(live_functions_controller.has_value());
-  LiveFunctionsDataView& live_functions_data_view =
-      live_functions_controller.value()->GetDataView();
-  int selected_row = live_functions_data_view.GetRowFromFunctionAddress(function_address);
-  if (selected_row != -1) {
-    ui->liveFunctions->onRowSelected(selected_row);
+  std::optional<int> selected_row(std::nullopt);
+  if (timer_info) {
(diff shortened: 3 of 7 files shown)


OrbitQt/orbitlivefunctions.cpp
```diff
@@ -108,12 +108,14 @@ void OrbitLiveFunctions::Reset() {
...
   QItemSelectionModel* selection = ui->data_view_panel_->GetTreeView()->selectionModel();
-  QModelIndex idx = ui->data_view_panel_->GetTreeView()->GetModel()->CreateIndex(row, 0);
-  selection->select(idx, QItemSelectionModel::ClearAndSelect | QItemSelectionModel::Rows);
-  selection->setCurrentIndex(idx, QItemSelectionModel::Rows);
+  QModelIndex index;
+  if (row.has_value()) {
+    index = ui->data_view_panel_->GetTreeView()->GetModel()->CreateIndex(row.value(), 0);
+  }
+  selection->select(index, QItemSelectionModel::ClearAndSelect | QItemSelectionModel::Rows);
+  selection->setCurrentIndex(index, QItemSelectionModel::Rows);
```
reichlfl, line 119: Are both calls above really needed? Looks a bit redundant, but I'm just guessing here.

reichlfl: Also I'm assuming the default constructor of `QModelIndex` results in an index that will then clear the selection?

OrbitQt/orbitlivefunctions.h
```diff
@@ -29,7 +29,7 @@ class OrbitLiveFunctions : public QWidget {
   void Initialize(SelectionType selection_type, FontType font_type, bool is_main_instance = true);
   void Refresh();
   void OnDataChanged();
-  void onRowSelected(int row);
+  void onRowSelected(std::optional<int> row);
```
reichlfl, line 32: If you're touching this method, can you change it to follow our coding style (`OnRowSelected`)? Thanks!
(I think this isn't overwritten from `QWidget`, right? If this is the case, add `override` please)

reichlfl (approved):
Just some small nitpicks, overall LGTM, thanks for taking care of this!

OrbitQt/orbitlivefunctions.cpp
vickyliu-go4it: > Also I'm assuming the default constructor of `QModelIndex` results in an index that will then clear the selection?

Yes. More specifically, when we deselect functions from the live tab or the capture window, we pass `std::nullopt` to `OnRowSelected(...)` to clear selections in the live tab.

vickyliu-go4it: > Are both calls above really needed? Looks a bit redundant, but I'm just guessing here.

Thanks for pointing out this! Yes, it's redundant and I will fix it.

OrbitQt/orbitlivefunctions.h
vickyliu-go4it: Thanks for the suggestion! I will update the name accordingly. Also, you are right. I just checked that it is not overwritten from `QWidget`.

OrbitQt/orbitmainwindow.cpp
```diff
@@ -769,14 +769,17 @@ void OrbitMainWindow::RestoreDefaultTabLayout() {
...
-      ui->liveFunctions->GetLiveFunctionsController().GetDataView();
-  int selected_row = live_functions_data_view.GetRowFromFunctionAddress(function_address);
-  if (selected_row != -1) {
-    ui->liveFunctions->onRowSelected(selected_row);
+  std::optional<int> selected_row(std::nullopt);
+  if (timer_info) {
+    uint64_t function_address = timer_info->function_address();
+    LiveFunctionsDataView& live_functions_data_view =
+        ui->liveFunctions->GetLiveFunctionsController().GetDataView();
+    selected_row = live_functions_data_view.GetRowFromFunctionAddress(function_address);
```
karupayun, line 777: Why not also modify this function on DataView to return an optional?

karupayun (approved):
Thanks for this PR Vicky! Very nice one! :)
I feel that works as expected, so nothing to comment about that.

I would like that when clicking on a FrameTrack (attached image), it also highlights the correspondent function on the Live tab. And maybe when clicking on a function who has a FrameTrack, make something on that FrameTrack, but not sure about this last.
frametrack no highlight

OrbitQt/orbitmainwindow.cpp
vickyliu-go4it: Good suggestion! Thanks!

vickyliu-go4it:
> Thanks for this PR Vicky! Very nice one! :)
> I feel that works as expected, so nothing to comment about that.
> 
> I would like that when clicking on a FrameTrack (attached image), it also highlights the correspondent function on the Live tab. And maybe when clicking on a function who has a FrameTrack, make something on that FrameTrack, but not sure about this last.
> frametrack no highlight

Thanks so much for the review an also for the suggestions on the FrameTrack. It's a good idea! I will work on it in the future PR.

reichlfl:
> Thanks for this PR Vicky! Very nice one! :)
> I feel that works as expected, so nothing to comment about that.
> 
> I would like that when clicking on a FrameTrack (attached image), it also highlights the correspondent function on the Live tab. And maybe when clicking on a function who has a FrameTrack, make something on that FrameTrack, but not sure about this last.
> frametrack no highlight

Hm, could be interesting. But we currently have two different "selection" behaviors when clicking on a track (selects the thread, if possible) and on a timer (selects the function, if possible). Not sure if this would be confusing because it mixes the two concepts. But could make sense to think about what is actually expected to happen when clicking on the various tracks, especially the non-thread tracks.

PR #1590, opened 2020-12-15
Fix GetThreadName: it was always returning the current thread's name

dpallotti:
Use an implementation similar to the one previously used in `OrbitLinuxTracing`.
Also add the unit test that would catch this.

Bug: http://b/175663948

### DIFF
--- a/OrbitBase/ThreadUtilsLinux.cpp
+++ b/OrbitBase/ThreadUtilsLinux.cpp
@@ -2,11 +2,13 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
 
-#include <OrbitBase/Logging.h>
+#include <absl/strings/strip.h>
 #include <sys/syscall.h>
 #include <time.h>
 #include <unistd.h>
 
+#include "OrbitBase/Logging.h"
+#include "OrbitBase/ReadFileToString.h"
 #include "OrbitBase/ThreadUtils.h"
 
 namespace orbit_base {
@@ -22,13 +24,16 @@ pid_t GetCurrentThreadId() {
 }
 
 std::string GetThreadName(pid_t tid) {
-  char thread_name[kMaxThreadNameLength];
-  int result = pthread_getname_np(pthread_self(), thread_name, kMaxThreadNameLength);
-  if (result != 0) {
-    ERROR("Getting thread name for tid %d. Error %d", tid, result);
-    return {};
+  std::string comm_filename = absl::StrFormat("/proc/%d/comm", tid);
+  ErrorMessageOr<std::string> comm_content = ReadFileToString(comm_filename);
+  if (!comm_content.has_value()) {
+    ERROR("Getting thread name for tid %d: %s", tid, comm_content.error().message());
+    return "";
   }
-  return thread_name;
+  if (!comm_content.value().empty() && comm_content.value().back() == '\n') {
+    comm_content.value().pop_back();
+  }
+  return comm_content.value();
 }
 
 void SetCurrentThreadName(const std::string& thread_name) {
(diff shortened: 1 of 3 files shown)


dimitry- (approved):

OrbitBase/ThreadUtilsLinux.cpp
```diff
@@ -22,13 +23,16 @@ pid_t GetCurrentThreadId() {
...
-  if (result != 0) {
-    ERROR("Getting thread name for tid %d. Error %d", tid, result);
-    return {};
+  std::string comm_filename = absl::StrFormat("/proc/%d/comm", tid);
+  ErrorMessageOr<std::string> comm_content = ReadFileToString(comm_filename);
+  if (!comm_content.has_value()) {
+    ERROR("Getting thread name for tid %d: %s", tid, comm_content.error().message());
+    return "";
+  }
+  if (comm_content.value().back() == '\n') {
```
dimitry-, line 32: absl::StripSuffix? It is slower though....

dimitry-: also - what if it is empty.. also not sure if this could happen, can you set "" name for a thread?

dimitry-: also this relies on the fact that value() returns a reference. maybe add auto value = std::move(comm_content.value()) and use that here and below?

OrbitBase/ThreadUtilsTest.cpp
```diff
@@ -51,3 +52,43 @@ TEST(ThreadUtils, GetSetThreadNames) {
...
+TEST(ThreadUtils, GetThreadName) {
+  absl::Mutex mutex;
+#if _WIN32
+  using PidType = uint32_t;
+#else
+  using PidType = pid_t;
+#endif
+  PidType other_tid = 0;
+  bool other_name_read = false;
+  static const std::string kThreadName = "OtherThread";
```
dimitry-, line 65: add a test for empty name?

OrbitBase/ThreadUtilsLinux.cpp
dpallotti: I don't mind it, but indeed it copies the string. Up to you.

dimitry-: > I don't mind it, but indeed it copies the string. Up to you.

I would keep it as it is. the '\n' check...

OrbitBase/ThreadUtilsTest.cpp
dpallotti: Added for the current thread. Also rearranged the tests a bit.

OrbitBase/ThreadUtilsLinux.cpp
dpallotti: > also this relies on the fact that value() returns a reference. maybe add auto value = std::move(comm_content.value()) and use that here and below?

No, it's normal that `value` returns a reference, it would be quite a weird interface otherwise.

dpallotti: > also - what if it is empty.. also not sure if this could happen, can you set "" name for a thread?

Done.

dimitry-: I would still add a check for value.empty(), I am not sure if the test works because comm always has \n at the end or we are just lucky.

dimitry- (approved):

OrbitBase/ThreadUtilsLinux.cpp
dpallotti: Aaah, you are right, I didn't understand what you meant above, this worked because the `\n` was always there but if we have the `if` we need to do it right! Added `!comm_content.value().empty() &&` to the condition.

dpallotti: Tell me if it looks good now before I get it wrong again.

dimitry- (approved):
lgtm!

PR #1591, opened 2020-12-15
Clear user defined capture data when switching processes

ronaldfw:
We have to clear user defined capture data including frame track data
when switching processes similar to how this is done with selected
(hooked) functions. This fixes a violated CHECK when switching
processes when frame tracks are enabled.

Test: Manual test.
Bug: http://b/175674080


florian-kuebler (approved):
Yep, that makes sense!

PR #1592, opened 2020-12-15
[rebase on master from release/1.57] Clear user defined capture data when switching processes

ronaldfw:
We have to clear user defined capture data including frame track data
when switching processes similar to how this is done with selected
(hooked functions). This fixes a violated CHECK when switching
processes when frame tracks are enabled.


florian-kuebler (approved):
Thanks!

PR #1593, opened 2020-12-15
Add Vulkan Layer Controller

florian-kuebler:
The `VulkanLayerController` glues together all the logic of the
Vulkan layer by having an representative for every instrumented
Vulkan function ("OnFunctionName").

It is also responsible for the bootstrapping (in particular
vkCreateInstance/Device) and the enumeration functions to expose
the layer's properties.

Bug: http://b/168200502
Test: Compile & Run tests

TODO: The commit introducing the grpc channel is missing, thus
the tests will fail, at the moment.

### DIFF
--- a/OrbitVulkanLayer/VulkanLayerController.h
+++ b/OrbitVulkanLayer/VulkanLayerController.h
@@ -392,14 +392,14 @@ class VulkanLayerController {
     // Append all of our extensions, that are not yet listed.
     // Note as this list of our extensions is very small, we are fine with O(N*M) runtime.
     for (const auto& extension : kDeviceExtensions) {
-      bool is_unique = true;
+      bool already_present = false;
       for (const auto& other_extension : extensions) {
         if (strcmp(extension.extensionName, other_extension.extensionName) == 0) {
-          is_unique = false;
+          already_present = true;
           break;
         }
       }
-      if (is_unique) {
+      if (!already_present) {
         extensions.push_back(extension);
       }
     }
--- a/OrbitVulkanLayer/VulkanLayerController.h
+++ b/OrbitVulkanLayer/VulkanLayerController.h
@@ -227,7 +227,7 @@ class VulkanLayerController {
 
   [[nodiscard]] VkResult OnQueueSubmit(VkQueue queue, uint32_t submit_count,
                                        const VkSubmitInfo* submits, VkFence fence) {
-    auto queue_submission_optional =
+    std::optional<typename SubmissionTracker::QueueSubmission> queue_submission_optional =
         submission_tracker_.PersistCommandBuffersOnSubmit(submit_count, submits);
     VkResult result = dispatch_table_.QueueSubmit(queue)(queue, submit_count, submits, fence);
     submission_tracker_.PersistDebugMarkersOnSubmit(queue, submit_count, submits,
--- a/OrbitVulkanLayer/VulkanLayerController.h
+++ b/OrbitVulkanLayer/VulkanLayerController.h
@@ -1,0 +1,454 @@
+// Copyright (c) 2020 The Orbit Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#ifndef ORBIT_VULKAN_LAYER_LAYER_LOGIC_H_
+#define ORBIT_VULKAN_LAYER_LAYER_LOGIC_H_
+
+#include "OrbitBase/Logging.h"
+#include "VulkanLayerProducerImpl.h"
+#include "absl/base/casts.h"
+#include "vulkan/vk_layer.h"
+#include "vulkan/vulkan.h"
+
+namespace orbit_vulkan_layer {
+
+/**
+ * This class controls the logic of this layer. For the instrumented vulkan functions,
+ * it provides an `On*` function (e.g. for `vkQueueSubmit` there is `OnQueueSubmit`) that delegates
+ * to the driver/next layer (see `DispatchTable`) and calls the required functions for this layer to
+ * function properly. So it ties together the classes like the `SubmissionTracker` or the
+ * `TimerQueryPool`.
+ * In particular it performs the bootstrapping code (OnCreateInstance/Device) and the enumerations
+ * required by every vulkan layer.
+ *
+ * Usage: For an instrumented vulkan function "X" a common pattern from the layers entry (Main.cpp)
+ * `OnX` needs to be called to the controller.
+ *
+ * Note, the main reason, to not expose the vulkan functions directly in this class, is that this
+ * allows us to write tests that check if we glue the code correctly together and does the proper
+ * bootstrapping.
+ */
+template <class DispatchTable, class QueueManager, class DeviceManager, class TimerQueryPool,
+          class SubmissionTracker>
+class VulkanLayerController {
+ public:
+  // layer metadata
+  static constexpr const char* const kLayerName = "ORBIT_VK_LAYER";
+  static constexpr const char* const kLayerDescription =
+      "Provides GPU insights for the Orbit Profiler";
+
+  static constexpr const uint32_t kLayerImplVersion = 1;
+  static constexpr const uint32_t kLayerSpecVersion = VK_API_VERSION_1_1;
+
+  static constexpr std::array<VkExtensionProperties, 3> kDeviceExtensions = {
+      VkExtensionProperties{.extensionName = VK_EXT_DEBUG_MARKER_EXTENSION_NAME,
+                            .specVersion = VK_EXT_DEBUG_MARKER_SPEC_VERSION},
+      VkExtensionProperties{.extensionName = VK_EXT_DEBUG_UTILS_EXTENSION_NAME,
+                            .specVersion = VK_EXT_DEBUG_UTILS_SPEC_VERSION},
+      VkExtensionProperties{.extensionName = VK_EXT_HOST_QUERY_RESET_EXTENSION_NAME,
+                            .specVersion = VK_EXT_HOST_QUERY_RESET_SPEC_VERSION}};
+
+  VulkanLayerController()
+      : device_manager_(&dispatch_table_),
+        timer_query_pool_(&dispatch_table_, kNumTimerQuerySlots),
+        submission_tracker_(&dispatch_table_, &timer_query_pool_, &device_manager_,
+                            std::numeric_limits<uint32_t>::max()) {}
+
+  ~VulkanLayerController() { CloseVulkanLayerProducerIfNecessary(); }
+
+  // ----------------------------------------------------------------------------
+  // Layer bootstrapping code
+  // ----------------------------------------------------------------------------
+
+  [[nodiscard]] VkResult OnCreateInstance(const VkInstanceCreateInfo* create_info,
+                                          const VkAllocationCallbacks* allocator,
+                                          VkInstance* instance) {
+    // The specification ensures that the create_info pointer must not be nullptr.
+    CHECK(create_info != nullptr);
+
+    auto* layer_create_info = absl::bit_cast<VkLayerInstanceCreateInfo*>(create_info->pNext);
+
+    while (layer_create_info != nullptr &&
+           (layer_create_info->sType != VK_STRUCTURE_TYPE_LOADER_INSTANCE_CREATE_INFO ||
+            layer_create_info->function != VK_LAYER_LINK_INFO)) {
+      layer_create_info = absl::bit_cast<VkLayerInstanceCreateInfo*>(layer_create_info->pNext);
+    }
+
+    if (layer_create_info == nullptr) {
+      return VK_ERROR_INITIALIZATION_FAILED;
+    }
+
+    InitVulkanLayerProducerIfNecessary();
+
+    PFN_vkGetInstanceProcAddr next_get_instance_proc_addr_function =
+        layer_create_info->u.pLayerInfo->pfnNextGetInstanceProcAddr;
+
+    // Advance linkage for next layer.
+    layer_create_info->u.pLayerInfo = layer_create_info->u.pLayerInfo->pNext;
+
+    // Need to call vkCreateInstance down the chain to actually create the
+    // instance, as we need it to be alive in the create instance dispatch table.
+    auto create_instance = absl::bit_cast<PFN_vkCreateInstance>(
+        next_get_instance_proc_addr_function(VK_NULL_HANDLE, "vkCreateInstance"));
+    VkResult result = create_instance(create_info, allocator, instance);
+
+    dispatch_table_.CreateInstanceDispatchTable(*instance, next_get_instance_proc_addr_function);
+
+    return result;
+  }
+
+  [[nodiscard]] VkResult OnCreateDevice(VkPhysicalDevice /*physical_device*/ physical_device,
+                                        const VkDeviceCreateInfo* create_info,
+                                        const VkAllocationCallbacks* allocator /*allocator*/,
+                                        VkDevice* device) {
+    auto* layer_create_info = absl::bit_cast<VkLayerDeviceCreateInfo*>(create_info->pNext);
+
+    while (layer_create_info != nullptr &&
+           (layer_create_info->sType != VK_STRUCTURE_TYPE_LOADER_DEVICE_CREATE_INFO ||
+            layer_create_info->function != VK_LAYER_LINK_INFO)) {
+      layer_create_info = absl::bit_cast<VkLayerDeviceCreateInfo*>(layer_create_info->pNext);
+    }
+
+    if (layer_create_info == nullptr) {
+      return VK_ERROR_INITIALIZATION_FAILED;
+    }
+
+    PFN_vkGetInstanceProcAddr next_get_instance_proc_addr_function =
+        layer_create_info->u.pLayerInfo->pfnNextGetInstanceProcAddr;
+    PFN_vkGetDeviceProcAddr next_get_device_proc_addr_function =
+        layer_create_info->u.pLayerInfo->pfnNextGetDeviceProcAddr;
+
+    // Advance linkage for next layer
+    layer_create_info->u.pLayerInfo = layer_create_info->u.pLayerInfo->pNext;
+
+    // Need to call vkCreateInstance down the chain to actually create the
+    // instance, as we need it to be alive in the create instance dispatch table.
+    auto create_device_function = absl::bit_cast<PFN_vkCreateDevice>(
+        next_get_instance_proc_addr_function(VK_NULL_HANDLE, "vkCreateDevice"));
+    VkResult result = create_device_function(physical_device, create_info, allocator, device);
+
+    dispatch_table_.CreateDeviceDispatchTable(*device, next_get_device_proc_addr_function);
+
+    device_manager_.TrackLogicalDevice(physical_device, *device);
+    timer_query_pool_.InitializeTimerQueryPool(*device);
+
+    return result;
+  }
+
+  [[nodiscard]] PFN_vkVoidFunction OnGetDeviceProcAddr(VkDevice device, const char* name) {
+    return dispatch_table_.GetDeviceProcAddr(device)(device, name);
+  }
+
+  [[nodiscard]] PFN_vkVoidFunction OnGetInstanceProcAddr(VkInstance instance, const char* name) {
+    return dispatch_table_.GetInstanceProcAddr(instance)(instance, name);
+  }
+
+  void OnDestroyInstance(VkInstance instance, const VkAllocationCallbacks* allocator) {
+    PFN_vkDestroyInstance destroy_instance_function = dispatch_table_.DestroyInstance(instance);
+    CHECK(destroy_instance_function != nullptr);
+    dispatch_table_.RemoveInstanceDispatchTable(instance);
+
+    destroy_instance_function(instance, allocator);
+
+    CloseVulkanLayerProducerIfNecessary();
+  }
+
+  void OnDestroyDevice(VkDevice device, const VkAllocationCallbacks* allocator) {
+    PFN_vkDestroyDevice destroy_device_function = dispatch_table_.DestroyDevice(device);
+    CHECK(destroy_device_function != nullptr);
+    device_manager_.UntrackLogicalDevice(device);
+    dispatch_table_.RemoveDeviceDispatchTable(device);
+
+    destroy_device_function(device, allocator);
+  }
+
+  // ----------------------------------------------------------------------------
+  // Core layer logic
+  // ----------------------------------------------------------------------------
+
+  [[nodiscard]] VkResult OnResetCommandPool(VkDevice device, VkCommandPool command_pool,
+                                            VkCommandPoolResetFlags flags) {
+    VkResult result = dispatch_table_.ResetCommandPool(device)(device, command_pool, flags);
+    submission_tracker_.ResetCommandPool(command_pool);
+    return result;
+  }
+
+  [[nodiscard]] VkResult OnAllocateCommandBuffers(VkDevice device,
+                                                  const VkCommandBufferAllocateInfo* allocate_info,
+                                                  VkCommandBuffer* command_buffers) {
+    VkResult result =
+        dispatch_table_.AllocateCommandBuffers(device)(device, allocate_info, command_buffers);
+
+    VkCommandPool pool = allocate_info->commandPool;
+    const uint32_t command_buffer_count = allocate_info->commandBufferCount;
+    submission_tracker_.TrackCommandBuffers(device, pool, command_buffers, command_buffer_count);
+    return result;
+  }
+
+  void OnFreeCommandBuffers(VkDevice device, VkCommandPool command_pool,
+                            uint32_t command_buffer_count, const VkCommandBuffer* command_buffers) {
+    submission_tracker_.UntrackCommandBuffers(device, command_pool, command_buffers,
+                                              command_buffer_count);
+    return dispatch_table_.FreeCommandBuffers(device)(device, command_pool, command_buffer_count,
+                                                      command_buffers);
+  }
+
+  [[nodiscard]] VkResult OnBeginCommandBuffer(VkCommandBuffer command_buffer,
+                                              const VkCommandBufferBeginInfo* begin_info) {
+    VkResult result =
+        dispatch_table_.BeginCommandBuffer(command_buffer)(command_buffer, begin_info);
+    submission_tracker_.MarkCommandBufferBegin(command_buffer);
+    return result;
+  }
+
+  [[nodiscard]] VkResult OnEndCommandBuffer(VkCommandBuffer command_buffer) {
+    submission_tracker_.MarkCommandBufferEnd(command_buffer);
+    return dispatch_table_.EndCommandBuffer(command_buffer)(command_buffer);
+  }
+
+  [[nodiscard]] VkResult OnResetCommandBuffer(VkCommandBuffer command_buffer,
+                                              VkCommandBufferResetFlags flags) {
+    submission_tracker_.ResetCommandBuffer(command_buffer);
+    return dispatch_table_.ResetCommandBuffer(command_buffer)(command_buffer, flags);
+  }
+
+  void OnGetDeviceQueue(VkDevice device, uint32_t queue_family_index, uint32_t queue_index,
+                        VkQueue* queue) {
+    dispatch_table_.GetDeviceQueue(device)(device, queue_family_index, queue_index, queue);
+    queue_manager_.TrackQueue(*queue, device);
+  }
+
+  void OnGetDeviceQueue2(VkDevice device, const VkDeviceQueueInfo2* queue_info, VkQueue* queue) {
+    dispatch_table_.GetDeviceQueue2(device)(device, queue_info, queue);
+    queue_manager_.TrackQueue(*queue, device);
+  }
+
+  [[nodiscard]] VkResult OnQueueSubmit(VkQueue queue, uint32_t submit_count,
+                                       const VkSubmitInfo* submits, VkFence fence) {
+    auto queue_submission_optional =
+        submission_tracker_.PersistCommandBuffersOnSubmit(submit_count, submits);
+    VkResult result = dispatch_table_.QueueSubmit(queue)(queue, submit_count, submits, fence);
+    submission_tracker_.PersistDebugMarkersOnSubmit(queue, submit_count, submits,
+                                                    queue_submission_optional);
+    return result;
+  }
+
+  [[nodiscard]] VkResult OnQueuePresentKHR(VkQueue queue, const VkPresentInfoKHR* present_info) {
+    submission_tracker_.CompleteSubmits(queue_manager_.GetDeviceOfQueue(queue));
+    return dispatch_table_.QueuePresentKHR(queue)(queue, present_info);
+  }
+
+  void OnCmdBeginDebugUtilsLabelEXT(VkCommandBuffer command_buffer,
+                                    const VkDebugUtilsLabelEXT* label_info) {
+    if (dispatch_table_.IsDebugUtilsExtensionSupported(command_buffer)) {
+      dispatch_table_.CmdBeginDebugUtilsLabelEXT(command_buffer)(command_buffer, label_info);
+    }
+
+    CHECK(label_info != nullptr);
+    submission_tracker_.MarkDebugMarkerBegin(command_buffer, label_info->pLabelName,
+                                             {
+                                                 .red = label_info->color[0],
+                                                 .green = label_info->color[1],
+                                                 .blue = label_info->color[2],
+                                                 .alpha = label_info->color[3],
+                                             });
+  }
+
+  void OnCmdEndDebugUtilsLabelEXT(VkCommandBuffer command_buffer) {
+    submission_tracker_.MarkDebugMarkerEnd(command_buffer);
+    if (dispatch_table_.IsDebugUtilsExtensionSupported(command_buffer)) {
+      dispatch_table_.CmdEndDebugUtilsLabelEXT(command_buffer)(command_buffer);
+    }
+  }
+
+  void OnCmdDebugMarkerBeginEXT(VkCommandBuffer command_buffer,
+                                const VkDebugMarkerMarkerInfoEXT* marker_info) {
+    if (dispatch_table_.IsDebugMarkerExtensionSupported(command_buffer)) {
+      dispatch_table_.CmdDebugMarkerBeginEXT(command_buffer)(command_buffer, marker_info);
+    }
+    CHECK(marker_info != nullptr);
+    submission_tracker_.MarkDebugMarkerBegin(command_buffer, marker_info->pMarkerName,
+                                             {
+                                                 .red = marker_info->color[0],
+                                                 .green = marker_info->color[1],
+                                                 .blue = marker_info->color[2],
+                                                 .alpha = marker_info->color[3],
+                                             });
+  }
+
+  void OnCmdDebugMarkerEndEXT(VkCommandBuffer command_buffer) {
+    submission_tracker_.MarkDebugMarkerEnd(command_buffer);
+    if (dispatch_table_.IsDebugMarkerExtensionSupported(command_buffer)) {
+      dispatch_table_.CmdDebugMarkerEndEXT(command_buffer)(command_buffer);
+    }
+  }
+
+  // ----------------------------------------------------------------------------
+  // Layer enumeration functions
+  // ----------------------------------------------------------------------------
+
+  [[nodiscard]] VkResult OnEnumerateInstanceLayerProperties(uint32_t* property_count,
+                                                            VkLayerProperties* properties) {
+    // Vulkan spec dictates that we are only supposed to enumerate ourselve.
+    if (property_count != nullptr) {
+      *property_count = 1;
+    }
+    if (properties != nullptr) {
+      snprintf(properties->layerName, sizeof(properties->layerName), kLayerName);
+      snprintf(properties->description, sizeof(properties->description), kLayerDescription);
+      properties->implementationVersion = kLayerImplVersion;
+      properties->specVersion = kLayerSpecVersion;
+    }
+
+    return VK_SUCCESS;
+  }
+
+  [[nodiscard]] VkResult OnEnumerateInstanceExtensionProperties(
+      const char* layer_name, uint32_t* property_count, VkExtensionProperties* /*properties*/) {
+    // Inform the client that we have no extension properties if this layer
+    // specifically is being queried.
+    if (layer_name != nullptr && strcmp(layer_name, kLayerName) == 0) {
+      if (property_count != nullptr) {
+        *property_count = 0;
+      }
+      return VK_SUCCESS;
+    }
+
+    // Vulkan spec mandates returning this when this layer isn't being queried.
+    return VK_ERROR_LAYER_NOT_PRESENT;
+  }
+
+  [[nodiscard]] VkResult OnEnumerateDeviceExtensionProperties(VkPhysicalDevice physical_device,
+                                                              const char* layer_name,
+                                                              uint32_t* property_count,
+                                                              VkExtensionProperties* properties) {
+    // If our layer is queried exclusively, we just return our extensions
+    if (layer_name != nullptr && strcmp(layer_name, kLayerName) == 0) {
+      // If properties == nullptr, only the number of extensions are queried.
+      if (properties == nullptr) {
+        *property_count = kDeviceExtensions.size();
+        return VK_SUCCESS;
+      }
+      uint32_t num_extensions_to_copy = kDeviceExtensions.size();
+      // In the case that less extensions are queried then the layer uses, we copy on this number
+      // and return VK_INCOMPLETE, according to the specification.
+      if (*property_count < num_extensions_to_copy) {
+        num_extensions_to_copy = *property_count;
+      }
+      memcpy(properties, kDeviceExtensions.data(),
+             num_extensions_to_copy * sizeof(VkExtensionProperties));
+      *property_count = num_extensions_to_copy;
+
+      if (num_extensions_to_copy < kDeviceExtensions.size()) {
+        return VK_INCOMPLETE;
+      }
+      return VK_SUCCESS;
+    }
+
+    // If a different layer is queried exclusively, we forward the call.
+    if (layer_name != nullptr) {
+      return dispatch_table_.EnumerateDeviceExtensionProperties(physical_device)(
+          physical_device, layer_name, property_count, properties);
+    }
+
+    // This is a general query, so we need to append our extensions to the once down in the
+    // callchain.
+    uint32_t num_other_extensions;
+    VkResult result = dispatch_table_.EnumerateDeviceExtensionProperties(physical_device)(
+        physical_device, nullptr, &num_other_extensions, nullptr);
+    if (result != VK_SUCCESS) {
+      return result;
+    }
+
+    std::vector<VkExtensionProperties> extensions(num_other_extensions);
+    result = dispatch_table_.EnumerateDeviceExtensionProperties(physical_device)(
+        physical_device, nullptr, &num_other_extensions, extensions.data());
+    if (result != VK_SUCCESS) {
+      return result;
+    }
+
+    // Lets append all of our extensions, that are not yet listed.
+    // Note, as this list of our extensions is very small, we are fine with O(N*M) runtime.
+    for (const auto& extension : kDeviceExtensions) {
+      bool is_unique = true;
+      for (const auto& other_extension : extensions) {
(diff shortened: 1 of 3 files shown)


OrbitVulkanLayer/VulkanLayerController.h
```diff
@@ -0,0 +1,454 @@
...
+    if (layer_create_info == nullptr) {
+      return VK_ERROR_INITIALIZATION_FAILED;
+    }
+
+    InitVulkanLayerProducerIfNecessary();
+
+    PFN_vkGetInstanceProcAddr next_get_instance_proc_addr_function =
+        layer_create_info->u.pLayerInfo->pfnNextGetInstanceProcAddr;
+
+    // Advance linkage for next layer.
```
dpallotti, line 87: Is "linkage" part of Vulkan terminology here? Or just "pointer"?

OrbitVulkanLayer/VulkanLayerController.h
```diff
@@ -0,0 +1,454 @@
...
+    layer_create_info->u.pLayerInfo = layer_create_info->u.pLayerInfo->pNext;
+
+    // Need to call vkCreateInstance down the chain to actually create the
+    // instance, as we need it to be alive in the create instance dispatch table.
```
dpallotti, line 91: What does "it" refer to here?

OrbitVulkanLayer/VulkanLayerController.h
```diff
@@ -0,0 +1,454 @@
...
+    if (layer_create_info == nullptr) {
+      return VK_ERROR_INITIALIZATION_FAILED;
+    }
+
+    PFN_vkGetInstanceProcAddr next_get_instance_proc_addr_function =
+        layer_create_info->u.pLayerInfo->pfnNextGetInstanceProcAddr;
+    PFN_vkGetDeviceProcAddr next_get_device_proc_addr_function =
+        layer_create_info->u.pLayerInfo->pfnNextGetDeviceProcAddr;
+
+    // Advance linkage for next layer
```
dpallotti, line 122: Again, "linkage"?

OrbitVulkanLayer/VulkanLayerController.h
```diff
@@ -0,0 +1,454 @@
...
+  }
+
+  void OnGetDeviceQueue2(VkDevice device, const VkDeviceQueueInfo2* queue_info, VkQueue* queue) {
+    dispatch_table_.GetDeviceQueue2(device)(device, queue_info, queue);
+    queue_manager_.TrackQueue(*queue, device);
+  }
+
+  [[nodiscard]] VkResult OnQueueSubmit(VkQueue queue, uint32_t submit_count,
+                                       const VkSubmitInfo* submits, VkFence fence) {
+    auto queue_submission_optional =
```
dpallotti, line 229: Could you state the type?

OrbitVulkanLayer/VulkanLayerController.h
```diff
@@ -0,0 +1,454 @@
...
+    return dispatch_table_.QueuePresentKHR(queue)(queue, present_info);
+  }
+
+  void OnCmdBeginDebugUtilsLabelEXT(VkCommandBuffer command_buffer,
+                                    const VkDebugUtilsLabelEXT* label_info) {
+    if (dispatch_table_.IsDebugUtilsExtensionSupported(command_buffer)) {
+      dispatch_table_.CmdBeginDebugUtilsLabelEXT(command_buffer)(command_buffer, label_info);
+    }
+
+    CHECK(label_info != nullptr);
```
dpallotti, line 248: Is this guaranteed by the standard?

OrbitVulkanLayer/VulkanLayerController.h
```diff
@@ -0,0 +1,454 @@
...
+    }
+  }
+
+  // ----------------------------------------------------------------------------
+  // Layer enumeration functions
+  // ----------------------------------------------------------------------------
+
+  [[nodiscard]] VkResult OnEnumerateInstanceLayerProperties(uint32_t* property_count,
+                                                            VkLayerProperties* properties) {
+    // Vulkan spec dictates that we are only supposed to enumerate ourselve.
```
dpallotti, line 293:
```suggestion
    // Vulkan spec dictates that we are only supposed to enumerate ourselves.
```

OrbitVulkanLayer/VulkanLayerController.h
```diff
@@ -0,0 +1,454 @@
...
+    if (property_count != nullptr) {
+      *property_count = 1;
+    }
```
dpallotti, line 296: So we only have one property? Could we have more? If that was the case, how could more than one property be set in `properties`.

OrbitVulkanLayer/VulkanLayerController.h
```diff
@@ -0,0 +1,454 @@
...
+          class SubmissionTracker>
+class VulkanLayerController {
+ public:
+  // layer metadata
+  static constexpr const char* const kLayerName = "ORBIT_VK_LAYER";
+  static constexpr const char* const kLayerDescription =
+      "Provides GPU insights for the Orbit Profiler";
+
+  static constexpr const uint32_t kLayerImplVersion = 1;
+  static constexpr const uint32_t kLayerSpecVersion = VK_API_VERSION_1_1;
```
dpallotti, line 42: So these need to be in sync with the json? I would at least add a comment about that.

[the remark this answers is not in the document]
dpallotti: They are redundant.

OrbitVulkanLayer/VulkanLayerController.h
```diff
@@ -0,0 +1,454 @@
...
+
+    // Vulkan spec mandates returning this when this layer isn't being queried.
+    return VK_ERROR_LAYER_NOT_PRESENT;
+  }
+
+  [[nodiscard]] VkResult OnEnumerateDeviceExtensionProperties(VkPhysicalDevice physical_device,
+                                                              const char* layer_name,
+                                                              uint32_t* property_count,
+                                                              VkExtensionProperties* properties) {
+    // If our layer is queried exclusively, we just return our extensions
```
dpallotti, line 326: "exclusively"? Does it mean that if `layer_name == nullptr` then all extensions are requested? In that case add a comment about it.

OrbitVulkanLayer/VulkanLayerController.h
```diff
@@ -0,0 +1,454 @@
...
+    if (layer_name != nullptr && strcmp(layer_name, kLayerName) == 0) {
+      // If properties == nullptr, only the number of extensions are queried.
+      if (properties == nullptr) {
+        *property_count = kDeviceExtensions.size();
+        return VK_SUCCESS;
+      }
+      uint32_t num_extensions_to_copy = kDeviceExtensions.size();
+      // In the case that less extensions are queried then the layer uses, we copy on this number
+      // and return VK_INCOMPLETE, according to the specification.
+      if (*property_count < num_extensions_to_copy) {
```
dpallotti, line 336: I'm confused here. So `property_count` is both an in and an out parameter?

OrbitVulkanLayer/VulkanLayerController.h
```diff
@@ -0,0 +1,454 @@
...
+      return VK_SUCCESS;
+    }
+
+    // If a different layer is queried exclusively, we forward the call.
+    if (layer_name != nullptr) {
+      return dispatch_table_.EnumerateDeviceExtensionProperties(physical_device)(
+          physical_device, layer_name, property_count, properties);
+    }
+
+    // This is a general query, so we need to append our extensions to the once down in the
```
dpallotti, line 355:
```suggestion
    // This is a general query, so we need to append our extensions to the ones down in the
```

OrbitVulkanLayer/VulkanLayerController.h
```diff
@@ -0,0 +1,454 @@
...
+
+    std::vector<VkExtensionProperties> extensions(num_other_extensions);
+    result = dispatch_table_.EnumerateDeviceExtensionProperties(physical_device)(
+        physical_device, nullptr, &num_other_extensions, extensions.data());
+    if (result != VK_SUCCESS) {
+      return result;
+    }
+
+    // Lets append all of our extensions, that are not yet listed.
+    // Note, as this list of our extensions is very small, we are fine with O(N*M) runtime.
```
dpallotti, line 372:
```suggestion
    // Note: as this list of our extensions is very small, we are fine with O(N*M) runtime.
```

OrbitVulkanLayer/VulkanLayerController.h
```diff
@@ -0,0 +1,454 @@
...
+    return VK_ERROR_LAYER_NOT_PRESENT;
+  }
+
+  [[nodiscard]] VkResult OnEnumerateDeviceExtensionProperties(VkPhysicalDevice physical_device,
+                                                              const char* layer_name,
+                                                              uint32_t* property_count,
+                                                              VkExtensionProperties* properties) {
+    // If our layer is queried exclusively, we just return our extensions
+    if (layer_name != nullptr && strcmp(layer_name, kLayerName) == 0) {
+      // If properties == nullptr, only the number of extensions are queried.
```
dpallotti, line 328:
```suggestion
      // If properties == nullptr, only the number of extensions is queried.
```

OrbitVulkanLayer/VulkanLayerController.h
```diff
@@ -0,0 +1,454 @@
...
+
+  [[nodiscard]] const DispatchTable* dispatch_table() const { return &dispatch_table_; }
+
+  [[nodiscard]] const SubmissionTracker* submission_tracker() const { return &submission_tracker_; }
+
+  [[nodiscard]] const DeviceManager* device_manager() const { return &device_manager_; }
+
+  [[nodiscard]] const TimerQueryPool* timer_query_pool() const { return &timer_query_pool_; }
+
+  [[nodiscard]] const QueueManager* queue_manager() const { return &queue_manager_; }
```
dpallotti, line 415: Why are these needed?

OrbitVulkanLayer/VulkanLayerController.h
```diff
@@ -0,0 +1,454 @@
...
+      vulkan_layer_producer_->BringUp(nullptr);
+      submission_tracker_.SetVulkanLayerProducer(vulkan_layer_producer_.get());
+    }
+  }
+
+  void CloseVulkanLayerProducerIfNecessary() {
+    absl::MutexLock lock{&vulkan_layer_producer_mutex_};
+    if (vulkan_layer_producer_ != nullptr) {
+      // TODO: Only do this when DestroyInstance has been called a number of times
+      //  equal to the number of times CreateInstance was called.
```
dpallotti, line 432: Forgot about this... Should we address this TODO?

OrbitVulkanLayer/VulkanLayerController.h
```diff
@@ -0,0 +1,454 @@
...
+      LOG("Taking down VulkanLayerProducer");
```
dpallotti, line 433: Should we have the corresponding `LOG` in `InitVulkanLayerProducerIfNecessary`?

dpallotti (commented):
I'm sending the comments on the class before proceeding with the tests. Mostly only comments about comments or questions, all seems good overall.

OrbitVulkanLayer/VulkanLayerControllerTest.cpp
```diff
@@ -0,0 +1,770 @@
...
+  VkInstanceCreateInfo create_info{.sType = VK_STRUCTURE_TYPE_INSTANCE_CREATE_INFO,
+                                   .pNext = &layer_create_info};
+  VkInstance created_instance;
+  VkResult result = controller_.OnCreateInstance(&create_info, nullptr, &created_instance);
+  EXPECT_EQ(result, VK_SUCCESS);
+  EXPECT_EQ(layer_create_info.u.pLayerInfo, &layer_link_1);
+
+  ::testing::Mock::VerifyAndClearExpectations(absl::bit_cast<void*>(submission_tracker));
+  // There will be a call at the destructor.
+  EXPECT_CALL(*submission_tracker, SetVulkanLayerProducer).Times(1);
```
dpallotti, line 361: You should be able to verify that the argument is `nullptr`, right?

OrbitVulkanLayer/VulkanLayerControllerTest.cpp
```diff
@@ -0,0 +1,770 @@
...
+  const MockDispatchTable* dispatch_table = controller_.dispatch_table();
+  EXPECT_CALL(*dispatch_table, DestroyInstance).Times(1).WillOnce(Return(mock_destroy_instance));
+  EXPECT_CALL(*dispatch_table, RemoveInstanceDispatchTable).Times(1);
+
+  VkInstance instance = {};
+
+  controller_.OnDestroyInstance(instance, nullptr);
+}
+
+TEST_F(VulkanLayerControllerTest, WillClearUpOnDestroyDevice) {
```
dpallotti, line 454: "clean up"?

OrbitVulkanLayer/VulkanLayerControllerTest.cpp
```diff
@@ -0,0 +1,770 @@
...
+
+  VkResult result = controller_.OnAllocateCommandBuffers(device, &allocate_info, &command_buffer);
+  EXPECT_EQ(result, VK_SUCCESS);
+}
+
+TEST_F(VulkanLayerControllerTest, CanDelegateOnFreeCommandBuffers) {
+  PFN_vkFreeCommandBuffers mock_free_command_buffers =
+      +[](VkDevice /*device*/, VkCommandPool /*command_pool*/, uint32_t /*command_buffer_count*/,
+          const VkCommandBuffer*
+          /*command_buffers*/) {};
```
dpallotti, line 517: clang-format being confused?

OrbitVulkanLayer/VulkanLayerControllerTest.cpp
```diff
@@ -0,0 +1,770 @@
...
+  VkDevice device = {};
+
+  controller_.OnDestroyDevice(device, nullptr);
+}
+
+// ----------------------------------------------------------------------------
+// Core layer logic
+// ----------------------------------------------------------------------------
+
+TEST_F(VulkanLayerControllerTest, CanDelegateOnResetCommandPool) {
```
dpallotti, line 472: Is delegate the right word here? Maybe "forward", as `SubmissionTracker::ResetCommandPool` is not the only work done in the system on `OnResetCommandPool` (or is it)? Also why "can"? And maybe specify the receiving end? In other words, if I get the test correctly this would be something like `ForwardsOnResetCommandPoolToSubmissionTracker`? Similarly for other tests below.

OrbitVulkanLayer/VulkanLayerControllerTest.cpp
```diff
@@ -0,0 +1,770 @@
...
+  const MockSubmissionTracker* submission_tracker = controller_.submission_tracker();
+  EXPECT_CALL(*submission_tracker, MarkDebugMarkerBegin).Times(1);
+
+  VkCommandBuffer command_buffer = {};
+  VkDebugUtilsLabelEXT debug_marker = {.sType = VK_STRUCTURE_TYPE_DEBUG_UTILS_LABEL_EXT,
+                                       .pLabelName = "Marker"};
+  controller_.OnCmdBeginDebugUtilsLabelEXT(command_buffer, &debug_marker);
+}
+
+TEST_F(VulkanLayerControllerTest, WillDelegateOnBeginDebugLabelIfDriverDoesSupportDebugUtils) {
```
dpallotti, line 663: Or also just "supports".

dpallotti (commented):
Pretty nice. I'll let you have a look at everything and then I'll have a final look.

OrbitVulkanLayer/VulkanLayerController.h
florian-kuebler: Done.

OrbitVulkanLayer/VulkanLayerController.h
florian-kuebler: Done.

OrbitVulkanLayer/VulkanLayerController.h
florian-kuebler: Yes, it does. For all those functions the pattern is usually the following:
```
vkFoo(&count, nullptr);
// Allocate an array "data" of size "count"
vkFoo(&count,  &data);
```
In that particular function, the count also acts as input. So if the requested count is smaller than the actual count, it only writes the smaller number of elements.

OrbitVulkanLayer/VulkanLayerController.h
florian-kuebler: Done.

OrbitVulkanLayer/VulkanLayerController.h
florian-kuebler: Done.

OrbitVulkanLayer/VulkanLayerController.h
florian-kuebler: We can also do this next year, I guess.

OrbitVulkanLayer/VulkanLayerController.h
florian-kuebler: Done.

florian-kuebler (commented):
Thanks for the review so far, done with the comments on the class itself. Will continue with the tests in an hour.

OrbitVulkanLayer/VulkanLayerControllerTest.cpp
```diff
@@ -0,0 +1,770 @@
...
+namespace {
+
+class MockDispatchTable {
+ public:
+  MOCK_METHOD(PFN_vkEnumerateDeviceExtensionProperties, EnumerateDeviceExtensionProperties,
+              (VkPhysicalDevice));
+  MOCK_METHOD((void), CreateInstanceDispatchTable, (VkInstance, PFN_vkGetInstanceProcAddr));
+  MOCK_METHOD((void), CreateDeviceDispatchTable, (VkDevice, PFN_vkGetDeviceProcAddr));
+  MOCK_METHOD((void), RemoveInstanceDispatchTable, (VkInstance));
+  MOCK_METHOD((void), RemoveDeviceDispatchTable, (VkDevice));
```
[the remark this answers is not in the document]
florian-kuebler: Nope, change it

OrbitVulkanLayer/VulkanLayerControllerTest.cpp
```diff
@@ -0,0 +1,770 @@
...
+  MOCK_METHOD((void), CompleteSubmits, (VkDevice));
+  MOCK_METHOD((void), MarkDebugMarkerBegin, (VkCommandBuffer, const char*, Color));
+  MOCK_METHOD((void), MarkDebugMarkerEnd, (VkCommandBuffer));
+};
+
+using VulkanLayerControllerImpl =
+    VulkanLayerController<MockDispatchTable, MockQueueManager, MockDeviceManager,
+                          MockTimerQueryPool, MockSubmissionTracker>;
+
+Matcher<VkExtensionProperties> VkExtensionPropertiesAreEqual(
```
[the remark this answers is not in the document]
florian-kuebler: That struct contains multiple properties. thats the vulkan terminology.

OrbitVulkanLayer/VulkanLayerControllerTest.cpp
```diff
@@ -0,0 +1,770 @@
...
+  return AllOf(
+      Field("specVersion", &VkExtensionProperties::specVersion, expected.specVersion),
+      Field("extensionName", &VkExtensionProperties::extensionName, StrEq(expected.extensionName)));
+}
+}  // namespace
+
+class VulkanLayerControllerTest : public ::testing::Test {
+ protected:
+  VulkanLayerController<MockDispatchTable, MockQueueManager, MockDeviceManager, MockTimerQueryPool,
+                        MockSubmissionTracker>
```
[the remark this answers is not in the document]
florian-kuebler: thanks!

OrbitVulkanLayer/VulkanLayerControllerTest.cpp
```diff
@@ -0,0 +1,770 @@
...
+      .extensionName = VK_EXT_HOST_QUERY_RESET_EXTENSION_NAME,
+      .specVersion = VK_EXT_HOST_QUERY_RESET_SPEC_VERSION};
+
+  static constexpr VkExtensionProperties kFakeExtension1{.extensionName = "Other Extension 1",
+                                                         .specVersion = 3};
+  static constexpr VkExtensionProperties kFakeExtension2{.extensionName = "Other Extension 2",
+                                                         .specVersion = 2};
+
+  static constexpr PFN_vkEnumerateDeviceExtensionProperties
+      kMockEnumerateDeviceExtensionPropertiesFunction =
```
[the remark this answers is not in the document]
florian-kuebler: right, I always confuse the wording here. Thanks.

OrbitVulkanLayer/VulkanLayerControllerTest.cpp
```diff
@@ -0,0 +1,770 @@
...
+  EXPECT_EQ(result, VK_SUCCESS);
+  EXPECT_THAT(actual_properties,
+              UnorderedElementsAreArray({VkExtensionPropertiesAreEqual(kFakeExtension1),
+                                         VkExtensionPropertiesAreEqual(kFakeExtension2),
+                                         VkExtensionPropertiesAreEqual(kDebugMarkerExtension)}));
+}
+
+TEST_F(VulkanLayerControllerTest,
+       WillReturnErrorOnEnumerateAllLayersDeviceExtensionPropertiesError) {
+  PFN_vkEnumerateDeviceExtensionProperties mock_enumerate_device_extension_properties_function =
```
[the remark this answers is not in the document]
florian-kuebler: done.

OrbitVulkanLayer/VulkanLayerControllerTest.cpp
```diff
@@ -0,0 +1,770 @@
...
+      .u.pLayerInfo = &layer_link_2};
+  VkInstanceCreateInfo create_info{.sType = VK_STRUCTURE_TYPE_INSTANCE_CREATE_INFO,
+                                   .pNext = &layer_create_info};
+  VkInstance created_instance;
+  VkResult result = controller_.OnCreateInstance(&create_info, nullptr, &created_instance);
+  EXPECT_EQ(result, VK_SUCCESS);
+  EXPECT_EQ(layer_create_info.u.pLayerInfo, &layer_link_1);
+
+  ::testing::Mock::VerifyAndClearExpectations(absl::bit_cast<void*>(submission_tracker));
+  // There will be a call at the destructor.
```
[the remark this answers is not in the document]
florian-kuebler: thanks!

OrbitVulkanLayer/VulkanLayerControllerTest.cpp
florian-kuebler: Done.

OrbitVulkanLayer/VulkanLayerControllerTest.cpp
florian-kuebler: Better, yes

OrbitVulkanLayer/VulkanLayerControllerTest.cpp
florian-kuebler: Yeah that is better. Thanks for the suggestion.

OrbitVulkanLayer/VulkanLayerControllerTest.cpp
florian-kuebler: Thanks. Strange.

florian-kuebler (commented):
Thanks for the reviews. They should be now addressed.

OrbitVulkanLayer/VulkanLayerController.h
dpallotti: Perfect then.

OrbitVulkanLayer/VulkanLayerController.h
dpallotti: Don't you have all the types available here? Of course they use the type parameters.

OrbitVulkanLayer/VulkanLayerController.h
dpallotti: Ah I guess that `VkLayerProperties*` is an array?

OrbitVulkanLayer/VulkanLayerController.h
dpallotti: Fine by me. I guess we should address it together with the other TODO above.

OrbitVulkanLayer/VulkanLayerControllerTest.cpp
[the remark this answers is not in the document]
dpallotti: > That struct contains multiple properties.

Does it? Isn't it just the weirdly named

```
typedef struct VkExtensionProperties {
    char        extensionName[VK_MAX_EXTENSION_NAME_SIZE];
    uint32_t    specVersion;
} VkExtensionProperties;
```
?

OrbitVulkanLayer/VulkanLayerControllerTest.cpp
dpallotti: Was this addressed?

dpallotti (approved):
The only thing missing is deciding what to do about those (well) `TODO`s. Should they be addressed in this change or later?

OrbitVulkanLayer/VulkanLayerController.h
florian-kuebler: No, this is an optional<SubmissionTracker::Submission>. Those types are not there.

OrbitVulkanLayer/VulkanLayerController.h
florian-kuebler: Exactly.

OrbitVulkanLayer/VulkanLayerController.h
florian-kuebler: Yep, lets do this then next year.

OrbitVulkanLayer/VulkanLayerControllerTest.cpp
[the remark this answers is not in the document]
florian-kuebler: Well, yes technically those are two properties, so "multiple". I admit, that the name is very strange though.

OrbitVulkanLayer/VulkanLayerControllerTest.cpp
florian-kuebler: Hmm somehow not. Should be done now.

OrbitVulkanLayer/VulkanLayerController.h
dpallotti: Ah, the return type is different in the tests, I noticed `// Note we simplify the return type, as it is not really part of this test.`. Why not add an empty `struct QueueSubmission{}` to `MockSubmissionTracker`? It seems better than using `bool`.

Then you should (maybe...) be able to write 
`std::optional<typename SubmissionTracker::QueueSubmission> queue_submission_optional =`.

OrbitVulkanLayer/VulkanLayerControllerTest.cpp
```diff
@@ -0,0 +1,775 @@
...
+  MOCK_METHOD(void, TrackCommandBuffers,
+              (VkDevice, VkCommandPool, const VkCommandBuffer*, uint32_t));
+  MOCK_METHOD(void, UntrackCommandBuffers,
+              (VkDevice, VkCommandPool, const VkCommandBuffer*, uint32_t));
+  MOCK_METHOD(void, MarkCommandBufferBegin, (VkCommandBuffer));
+  MOCK_METHOD(void, MarkCommandBufferEnd, (VkCommandBuffer));
+  MOCK_METHOD(void, ResetCommandBuffer, (VkCommandBuffer));
+  MOCK_METHOD(std::optional<QueueSubmission>, PersistCommandBuffersOnSubmit,
+              (uint32_t, const VkSubmitInfo* submits));
+  MOCK_METHOD(bool, PersistDebugMarkersOnSubmit,
```
dpallotti, line 99: Super minor thing I noticed: this is originally returning a `void`. You don't need the mock to return a `bool`, right?

florian-kuebler: Very good catch! Thanks!

OrbitVulkanLayer/VulkanLayerController.h
```diff
@@ -0,0 +1,460 @@
...
+#include "vulkan/vk_layer.h"
+#include "vulkan/vulkan.h"
+
+namespace orbit_vulkan_layer {
+
+/**
+ * This class controls the logic of this layer. For the intercepted Vulkan functions,
+ * it provides an `On*` function (e.g. for `vkQueueSubmit` there is `OnQueueSubmit`) that delegates
+ * to the driver/next layer (see `DispatchTable`) and calls the required functions for this layer to
+ * function properly. So it ties together the classes like the `SubmissionTracker` or the
```
ronaldfw, line 20:
```suggestion
 * function properly. So it ties together classes like `SubmissionTracker` or
```

OrbitVulkanLayer/VulkanLayerController.h
```diff
@@ -0,0 +1,460 @@
...
+ * `TimerQueryPool`.
+ * In particular it executes the bootstrapping code (OnCreateInstance/Device) and the enumerations
+ * required by every Vulkan layer, to describe the layer as well as the extensions it uses.
+ *
+ * Usage: For an intercepted Vulkan function "X" in the layers entry (Main.cpp), `OnX` needs to be
```
ronaldfw, line 25: Main.cpp is now called EntryPoints.cpp, I think.

OrbitVulkanLayer/VulkanLayerController.h
```diff
@@ -0,0 +1,460 @@
...
+ *
+ * Note the main reason not to expose the Vulkan functions directly in this class is that this
+ * allows us to write tests. Those tests can check if we glue the code correctly together
+ * and if we do the proper bootstrapping.
+ */
+template <class DispatchTable, class QueueManager, class DeviceManager, class TimerQueryPool,
+          class SubmissionTracker>
+class VulkanLayerController {
+ public:
+  // Layer metadata. This must be in sync with the json file in the resources.
```
ronaldfw, line 36: Does the description have to match exactly?

OrbitVulkanLayer/VulkanLayerController.h
```diff
@@ -0,0 +1,460 @@
...
+  static constexpr const char* kLayerName = "ORBIT_VK_LAYER";
+  static constexpr const char* kLayerDescription = "Provides GPU insights for the Orbit Profiler";
+
+  static constexpr uint32_t kLayerImplVersion = 1;
```
ronaldfw, line 40: Not clear if this is part of the layer metadata? Remove the empty line above?

OrbitVulkanLayer/VulkanLayerController.h
```diff
@@ -0,0 +1,460 @@
...
+  static constexpr uint32_t kLayerSpecVersion = VK_API_VERSION_1_1;
+
+  static constexpr std::array<VkExtensionProperties, 3> kDeviceExtensions = {
```
ronaldfw, line 43: Are these the extensions required by the layer? If yes, the name should reflect that (kRequiredDeviceExtensions).

OrbitVulkanLayer/VulkanLayerController.h
```diff
@@ -0,0 +1,460 @@
...
+        layer_create_info->u.pLayerInfo->pfnNextGetInstanceProcAddr;
+
+    // Advance linkage for next layer.
+    layer_create_info->u.pLayerInfo = layer_create_info->u.pLayerInfo->pNext;
+
+    // Need to call vkCreateInstance down the chain to actually create the
+    // instance, as we need the instance to be alive in the create instance dispatch table.
+    auto create_instance = absl::bit_cast<PFN_vkCreateInstance>(
+        next_get_instance_proc_addr_function(VK_NULL_HANDLE, "vkCreateInstance"));
+    VkResult result = create_instance(create_info, allocator, instance);
```
ronaldfw, line 95: The comment above seems to suggest that this call must not fail before calling "CreateInstanceDispatchTable". Should we check the result here to make sure sure that it succeeded?

OrbitVulkanLayer/VulkanLayerController.h
```diff
@@ -0,0 +1,460 @@
...
+        layer_create_info->u.pLayerInfo->pfnNextGetDeviceProcAddr;
+
+    // Advance linkage for next layer
+    layer_create_info->u.pLayerInfo = layer_create_info->u.pLayerInfo->pNext;
+
+    // Need to call vkCreateInstance down the chain to actually create the
+    // instance, as we need it to be alive in the create instance dispatch table.
+    auto create_device_function = absl::bit_cast<PFN_vkCreateDevice>(
+        next_get_instance_proc_addr_function(VK_NULL_HANDLE, "vkCreateDevice"));
+    VkResult result = create_device_function(physical_device, create_info, allocator, device);
```
ronaldfw, line 130: Similar question here: Should we check "result" before proceeding with the CreateDeviceDispatchTable call?

OrbitVulkanLayer/VulkanLayerController.h
```diff
@@ -0,0 +1,460 @@
...
+    destroy_device_function(device, allocator);
+  }
+
+  // ----------------------------------------------------------------------------
+  // Core layer logic
+  // ----------------------------------------------------------------------------
+
+  [[nodiscard]] VkResult OnResetCommandPool(VkDevice device, VkCommandPool command_pool,
+                                            VkCommandPoolResetFlags flags) {
+    VkResult result = dispatch_table_.ResetCommandPool(device)(device, command_pool, flags);
```
ronaldfw, line 173: Check result before proceeding?

OrbitVulkanLayer/VulkanLayerController.h
```diff
@@ -0,0 +1,460 @@
...
+    submission_tracker_.ResetCommandPool(command_pool);
+    return result;
+  }
+
+  [[nodiscard]] VkResult OnAllocateCommandBuffers(VkDevice device,
+                                                  const VkCommandBufferAllocateInfo* allocate_info,
+                                                  VkCommandBuffer* command_buffers) {
+    VkResult result =
```
ronaldfw, line 181: Check result before proceeding? It seems if allocation here fails, we should not do the next steps..

OrbitVulkanLayer/VulkanLayerController.h
```diff
@@ -0,0 +1,460 @@
...
+                            uint32_t command_buffer_count, const VkCommandBuffer* command_buffers) {
+    submission_tracker_.UntrackCommandBuffers(device, command_pool, command_buffers,
+                                              command_buffer_count);
+    return dispatch_table_.FreeCommandBuffers(device)(device, command_pool, command_buffer_count,
+                                                      command_buffers);
+  }
+
+  [[nodiscard]] VkResult OnBeginCommandBuffer(VkCommandBuffer command_buffer,
+                                              const VkCommandBufferBeginInfo* begin_info) {
+    VkResult result =
```
ronaldfw, line 200: Check result?

OrbitVulkanLayer/VulkanLayerController.h
```diff
@@ -0,0 +1,460 @@
...
+  void OnGetDeviceQueue2(VkDevice device, const VkDeviceQueueInfo2* queue_info, VkQueue* queue) {
+    dispatch_table_.GetDeviceQueue2(device)(device, queue_info, queue);
+    queue_manager_.TrackQueue(*queue, device);
+  }
+
+  [[nodiscard]] VkResult OnQueueSubmit(VkQueue queue, uint32_t submit_count,
+                                       const VkSubmitInfo* submits, VkFence fence) {
+    std::optional<typename SubmissionTracker::QueueSubmission> queue_submission_optional =
+        submission_tracker_.PersistCommandBuffersOnSubmit(submit_count, submits);
+    VkResult result = dispatch_table_.QueueSubmit(queue)(queue, submit_count, submits, fence);
```
ronaldfw, line 232: Check result?

OrbitVulkanLayer/VulkanLayerController.h
```diff
@@ -0,0 +1,460 @@
...
+  }
+
+  [[nodiscard]] VkResult OnQueuePresentKHR(VkQueue queue, const VkPresentInfoKHR* present_info) {
+    submission_tracker_.CompleteSubmits(queue_manager_.GetDeviceOfQueue(queue));
+    return dispatch_table_.QueuePresentKHR(queue)(queue, present_info);
+  }
+
+  void OnCmdBeginDebugUtilsLabelEXT(VkCommandBuffer command_buffer,
+                                    const VkDebugUtilsLabelEXT* label_info) {
+    if (dispatch_table_.IsDebugUtilsExtensionSupported(command_buffer)) {
```
ronaldfw, line 245: Is this "if" needed? Shouldn't we already just fail earlier if this extension is not supported?

ronaldfw: Same question in a few places below.

OrbitVulkanLayer/VulkanLayerController.h
```diff
@@ -0,0 +1,460 @@
...
+    }
+
+    std::vector<VkExtensionProperties> extensions(num_other_extensions);
+    result = dispatch_table_.EnumerateDeviceExtensionProperties(physical_device)(
+        physical_device, nullptr, &num_other_extensions, extensions.data());
+    if (result != VK_SUCCESS) {
+      return result;
+    }
+
+    // Lets append all of our extensions, that are not yet listed.
```
ronaldfw, line 376:
```suggestion
    // Append all of our extensions that are not yet listed.
```

OrbitVulkanLayer/VulkanLayerController.h
```diff
@@ -0,0 +1,460 @@
...
+    // Note as this list of our extensions is very small, we are fine with O(N*M) runtime.
```
ronaldfw, line 377: And this function is also only called during initialization (correct?).

OrbitVulkanLayer/VulkanLayerController.h
```diff
@@ -0,0 +1,460 @@
...
+    for (const auto& extension : kDeviceExtensions) {
+      bool is_unique = true;
```
ronaldfw, line 379: Nit: I find this a bit unclear, but also maybe I'm misunderstanding. If this were:

bool already_present = false;
for () {
   if (strcmp(...)) {
     already_present = true;
   }
} 

if (!already_present) {
...
}

I think it'd a bit easier to understand.

OrbitVulkanLayer/VulkanLayerController.h
```diff
@@ -0,0 +1,460 @@
...
+    }
+
+    // As above, if properties is nullptr, only the number if extensions is queried.
+    if (properties == nullptr) {
+      *property_count = extensions.size();
+      return VK_SUCCESS;
+    }
+
+    uint32_t num_extensions_to_copy = extensions.size();
+    // In the case that less extensions are queried then the layer uses, we copy on this number and
```
ronaldfw, line 398:
```suggestion
    // In the case that less extensions are queried than the layer uses, we copy this number and
```

OrbitVulkanLayer/VulkanLayerController.h
```diff
@@ -0,0 +1,460 @@
...
+  [[nodiscard]] const TimerQueryPool* timer_query_pool() const { return &timer_query_pool_; }
+
+  [[nodiscard]] const QueueManager* queue_manager() const { return &queue_manager_; }
+
+ private:
+  void InitVulkanLayerProducerIfNecessary() {
+    absl::MutexLock lock{&vulkan_layer_producer_mutex_};
+    if (vulkan_layer_producer_ == nullptr) {
+      vulkan_layer_producer_ = std::make_unique<VulkanLayerProducerImpl>();
+      // TODO: As soon as we do have producer side channel for the port, use it.
```
ronaldfw, line 427:
```suggestion
      // TODO: As soon as we have a producer side channel for the port, use it.
```

OrbitVulkanLayer/VulkanLayerController.h
```diff
@@ -0,0 +1,460 @@
...
+      LOG("Bringing up VulkanLayerProducer");
+      vulkan_layer_producer_->BringUp(nullptr);
+      submission_tracker_.SetVulkanLayerProducer(vulkan_layer_producer_.get());
+    }
+  }
+
+  void CloseVulkanLayerProducerIfNecessary() {
+    absl::MutexLock lock{&vulkan_layer_producer_mutex_};
+    if (vulkan_layer_producer_ != nullptr) {
+      // TODO: Only do this when DestroyInstance has been called a number of times
```
ronaldfw, line 437:
```suggestion
      // TODO: Only do this when DestroyInstance has been called the same number of times as CreateInstance.
```

OrbitVulkanLayer/VulkanLayerController.h
```diff
@@ -0,0 +1,460 @@
...
+  std::unique_ptr<VulkanLayerProducer> vulkan_layer_producer_ = nullptr;
+  absl::Mutex vulkan_layer_producer_mutex_;
+
+  DispatchTable dispatch_table_;
+  DeviceManager device_manager_;
+  TimerQueryPool timer_query_pool_;
+  SubmissionTracker submission_tracker_;
+  QueueManager queue_manager_;
+
+  static constexpr uint32_t kNumTimerQuerySlots = 65536;
```
ronaldfw, line 455: Is this a constant from the specification?

ronaldfw (commented):
Looks great, mostly minor comments!

I would like to understand better the places where some calls are made that return a VkResult but our code doesn't act on the result, usually something else is done, and then the result is just returned. Shouldn't we check the result and possibly act differently depending on its value?

OrbitVulkanLayer/VulkanLayerController.h
florian-kuebler: Done.

OrbitVulkanLayer/VulkanLayerController.h
florian-kuebler: Done.

OrbitVulkanLayer/VulkanLayerController.h
florian-kuebler: As above, I added an `if` rather then an `CHECK`.

OrbitVulkanLayer/VulkanLayerController.h
florian-kuebler: As above. Thanks.

OrbitVulkanLayer/VulkanLayerController.h
florian-kuebler: As above, done with an if.

florian-kuebler (commented):
I addressed the comments. Guarded the **post* Vulkan code, such that it only gets executed if the call was successful.

OrbitVulkanLayer/VulkanLayerController.h
ronaldfw: Thanks for the explanation!

OrbitVulkanLayer/VulkanLayerController.h
ronaldfw: I don't see the name changed here or maybe I'm looking at the wrong thing?

OrbitVulkanLayer/VulkanLayerController.h
ronaldfw: Agree we should not fail on error (with CHECK) and actually return the error, but that we also should not do further calls in case of an error. Change with "if" LGTM.

OrbitVulkanLayer/VulkanLayerController.h
ronaldfw: Looks good.

OrbitVulkanLayer/VulkanLayerController.h
ronaldfw: Looks good.

OrbitVulkanLayer/VulkanLayerController.h
ronaldfw: Ok, sounds good.

OrbitVulkanLayer/VulkanLayerController.h
ronaldfw: Can you add a comment how this is chosen and why this value was chosen? It's fine to say that the choice is "arbitrary but large enough" or something like that, but someone not familiar with the code should have an idea where this value came frame (and knowing that it just needs to be "large enough" is valuable information).

ronaldfw (approved):
LGTM, just some smaller things left.

OrbitVulkanLayer/VulkanLayerController.h
florian-kuebler: Oh, sorry missed that one.

OrbitVulkanLayer/VulkanLayerController.h
florian-kuebler: Done.

florian-kuebler (commented):
Addressed the comments. Thanks for the input.

PR #1594, opened 2020-12-15
Add OrbitProducerSideChannel module defining Unix socket for producers

dpallotti:
This defines the Unix socket over which the communication between producers of
`CaptureEvent`s and `OrbitService` will happen.
In the near future, it will be used by `OrbitVulkanLayer` and `OrbitService`.
The name is what it is but it is for consistency with "producer side" used
elsewhere to define the side of `OrbitService` that will communicate with
producers.

Bug: http://b/174028631


dpallotti:
The `CMakeLists.txt` file might very well contain mistakes, I'm a bit lost there with this header-only module.

OrbitProducerSideChannel/CMakeLists.txt
```diff
@@ -0,0 +1,13 @@
+# Copyright (c) 2020 The Orbit Authors. All rights reserved.
+# Use of this source code is governed by a BSD-style license that can be
+# found in the LICENSE file.
+
+project(OrbitProducerSideChannel)
+
+add_library(OrbitProducerSideChannel INTERFACE)
+target_compile_options(OrbitProducerSideChannel INTERFACE ${STRICT_COMPILE_FLAGS})
```
beckerhe, line 8: You probably want to remove this line. This would enforce the strict compiler flags on all users of this module. And that's what we want to avoid be enabling it on a module-by-module-basis.

OrbitProducerSideChannel/include/OrbitProducerSideChannel/ProducerSideChannel.h
```diff
@@ -0,0 +1,26 @@
...
+#ifndef ORBIT_PRODUCER_SIDE_CHANNEL_PRODUCER_SIDE_CHANNEL_H_
+#define ORBIT_PRODUCER_SIDE_CHANNEL_PRODUCER_SIDE_CHANNEL_H_
+
+#include "grpcpp/grpcpp.h"
+
+namespace orbit_producer_side_channel {
+
+// This is the default path of the Unix domain socket used for the communication
+// between producers of CaptureEvents and OrbitService.
+constexpr const char* kProducerSideUnixDomainSocketPath = "/tmp/orbit-producer-side-socket";
```
beckerhe, line 14: Personal preference and very nit: Maybe we could construct a std::string_view here directory?

beckerhe (approved):
LGTM!

CMakeLists.txt
```diff
@@ -135,6 +135,7 @@ else()
   add_subdirectory(OrbitCaptureGgpService)
   add_subdirectory(OrbitLinuxTracing)
   add_subdirectory(OrbitProducer)
+  add_subdirectory(OrbitProducerSideChannel)
```
florian-kuebler, line 138: This thing is Linux-only, but would compile on Windows, right? Should it still be only included on Linux?

florian-kuebler (approved):
In general looks good. Thanks for that.
Im just unsure if this is/should be Linux-only.

OrbitProducerSideChannel/CMakeLists.txt
dpallotti: Done.

OrbitProducerSideChannel/CMakeLists.txt
```diff
@@ -0,0 +1,13 @@
...
+
+target_sources(OrbitProducerSideChannel INTERFACE
+        include/OrbitProducerSideChannel/ProducerSideChannel.h)
+
+target_include_directories(OrbitProducerSideChannel INTERFACE ${CMAKE_CURRENT_SOURCE_DIR}/include)
```
dpallotti, line 12: @beckerhe, should I add 

```
target_link_libraries(OrbitProducerSideChannel INTERFACE
        CONAN_PKG::abseil)
```

here? And if yes, what about gRPC for `#include <grpcpp/grpcpp.h>`? What do I need to link for that?

OrbitProducerSideChannel/include/OrbitProducerSideChannel/ProducerSideChannel.h
dpallotti: Done, I don't have a preference, mostly I'm still a bit lost about `string_view`. I though it was mostly for receiving a string rather than to pass a string around?

CMakeLists.txt
dpallotti: I don't have an opinion. I made it Linux only because it's conceptually Linux only, as it's about Unix domain sockets. Should it also be build on Windows?

florian-kuebler: oh sorry, didn't noticed it is already in the linux only block. Perfect!

OrbitProducerSideChannel/CMakeLists.txt
```diff
@@ -0,0 +1,16 @@
...
+add_library(OrbitProducerSideChannel INTERFACE)
+
+target_sources(OrbitProducerSideChannel INTERFACE
+        include/OrbitProducerSideChannel/ProducerSideChannel.h)
+
+target_include_directories(OrbitProducerSideChannel INTERFACE ${CMAKE_CURRENT_SOURCE_DIR}/include)
+
+target_link_libraries(OrbitProducerSideChannel INTERFACE
+        CONAN_PKG::abseil
+        CONAN_PKG::protobuf)
```
dpallotti, line 16: Hi @beckerhe , sorry to tag you again. I added this here now. Could you tell me if it makes more sense or if it's a mistake with this kind of module?

beckerhe: No, that's totally fine!

OrbitProducerSideChannel/CMakeLists.txt
beckerhe: You should also add `CONAN_PKG::grpc`.

OrbitProducerSideChannel/CMakeLists.txt
beckerhe: And sorry for being late!

OrbitProducerSideChannel/include/OrbitProducerSideChannel/ProducerSideChannel.h
beckerhe: Yes, that's the most common use-case. But it also works great as a string-literal replacement. And there is also a performance consideration: Most string literals (const char*) needs to be converted into a std::string at some point.
That requires std::strlen to be called to determine the length. But the length could already be known since it's static
and a literal. string_view solves that problem. It has a constexpr constructor and determines the length at compile time,
avoiding the strlen-call at runtime.

beckerhe (approved):
Sorry for being late. I added answers to your questions.

OrbitProducerSideChannel/CMakeLists.txt
dpallotti: Ah well, `grpc` is obviously what I need here, not `protobuf`. Whoops. Done.

OrbitProducerSideChannel/CMakeLists.txt
dpallotti: No worries, thank you!

OrbitProducerSideChannel/include/OrbitProducerSideChannel/ProducerSideChannel.h
dpallotti: Ah excellent, thanks for the explanation!

PR #1596, opened 2020-12-15
Removal of global variable GOrbitApp

beckerhe:
This PR removes GOrbitApp from the code base.

It has a lot of commits - mainly separated by class. Almost all of them a straight forward.
They introduce a member variable `OrbitApp* app_` which is initialized via the constructor and used
in the class's implementation instead of GOrbitApp.

All commits compile on its own and I recommend to review them one-by-one. It's a large PR,
but the commits itself are rather small. Let me know if you think this needs to be split up into multiple PRs.

The reason for doing this is mainly #1526. This PR needs to move some resource ownership (i.e. ProcessManager) from OrbitApp to OrbitMainWindow, which means OrbitMainWindow needs to control the lifetime of OrbitApp, otherwise OrbitApp might end up with a dangling pointer when OrbitMainWindow gets out of scope. So after removing all of GOrbitApp, there is a commit that hands ownership of OrbitApp to OrbitMainWindow.

It did some extensive manual testing with the branch and I couldn't crash it or found any bug.

Note: This does not solve the cyclic dependency problem that we have with OrbitApp ("OrbitApp owns X and X needs to access OrbitApp"). It can and has to be refactored independently. This is also the reason why this change introduces a lot of forward declarations of OrbitApp. They are needed due to cyclic dependencies.

EDIT: Note that the license_headers presubmit check is expected to fail due to the known orbitcodeeditor-problem. b/163462463 for more details.

EDIT: Bug added for tracking the progress of follow-up refactorings of the cyclic dependencies: http://b/175759190

### DIFF
--- a/OrbitGl/App.cpp
+++ b/OrbitGl/App.cpp
@@ -1567,7 +1567,7 @@ DataView* OrbitApp::GetOrCreateDataView(DataViewType type) {
 
     case DataViewType::kPresets:
       if (!presets_data_view_) {
-        presets_data_view_ = std::make_unique<PresetsDataView>();
+        presets_data_view_ = std::make_unique<PresetsDataView>(this);
         panels_.push_back(presets_data_view_.get());
       }
       return presets_data_view_.get();
--- a/OrbitGl/App.cpp
+++ b/OrbitGl/App.cpp
@@ -104,7 +104,6 @@ PresetLoadState GetPresetLoadStateForProcess(
 }
 }  // namespace
 
-std::unique_ptr<OrbitApp> GOrbitApp;
 bool DoZoom = false;
 
 OrbitApp::OrbitApp(ApplicationOptions&& options,
--- a/OrbitGl/App.cpp
+++ b/OrbitGl/App.cpp
@@ -462,7 +462,7 @@ void OrbitApp::ListPresets() {
 void OrbitApp::RefreshCaptureView() {
   ORBIT_SCOPE_FUNCTION;
   NeedsRedraw();
-  GOrbitApp->FireRefreshCallbacks();
+  FireRefreshCallbacks();
   DoZoom = true;  // TODO: remove global, review logic
 }
 
@@ -528,15 +528,14 @@ void OrbitApp::Disassemble(int32_t pid, const FunctionInfo& function) {
 
 Timer GMainTimer;
 
-// TODO: make it non-static
 void OrbitApp::MainTick() {
   ORBIT_SCOPE("OrbitApp::MainTick");
   GMainTimer.Restart();
 
-  if (DoZoom && GOrbitApp->HasCaptureData()) {
+  if (DoZoom && HasCaptureData()) {
     GCurrentTimeGraph->SortTracks();
-    GOrbitApp->capture_window_->ZoomAll();
-    GOrbitApp->NeedsRedraw();
+    capture_window_->ZoomAll();
+    NeedsRedraw();
     DoZoom = false;
   }
 }
@@ -1162,7 +1161,7 @@ void OrbitApp::LoadSymbols(const std::filesystem::path& symbols_path, ModuleData
       }
 
       UpdateAfterSymbolLoading();
-      GOrbitApp->FireRefreshCallbacks();
+      FireRefreshCallbacks();
     });
   });
 }
--- a/OrbitGl/App.cpp
+++ b/OrbitGl/App.cpp
@@ -1551,7 +1551,7 @@ DataView* OrbitApp::GetOrCreateDataView(DataViewType type) {
 
     case DataViewType::kModules:
       if (!modules_data_view_) {
-        modules_data_view_ = std::make_unique<ModulesDataView>();
+        modules_data_view_ = std::make_unique<ModulesDataView>(this);
         panels_.push_back(modules_data_view_.get());
       }
       return modules_data_view_.get();
--- a/OrbitGl/App.cpp
+++ b/OrbitGl/App.cpp
@@ -1537,7 +1537,7 @@ DataView* OrbitApp::GetOrCreateDataView(DataViewType type) {
   switch (type) {
     case DataViewType::kFunctions:
       if (!functions_data_view_) {
-        functions_data_view_ = std::make_unique<FunctionsDataView>();
+        functions_data_view_ = std::make_unique<FunctionsDataView>(this);
         panels_.push_back(functions_data_view_.get());
       }
       return functions_data_view_.get();
--- a/OrbitGl/App.cpp
+++ b/OrbitGl/App.cpp
@@ -1544,7 +1544,7 @@ DataView* OrbitApp::GetOrCreateDataView(DataViewType type) {
 
     case DataViewType::kCallstack:
       if (!callstack_data_view_) {
-        callstack_data_view_ = std::make_unique<CallStackDataView>();
+        callstack_data_view_ = std::make_unique<CallStackDataView>(this);
         panels_.push_back(callstack_data_view_.get());
       }
       return callstack_data_view_.get();
@@ -1597,7 +1597,7 @@ DataView* OrbitApp::GetOrCreateDataView(DataViewType type) {
 
 DataView* OrbitApp::GetOrCreateSelectionCallstackDataView() {
   if (selection_callstack_data_view_ == nullptr) {
-    selection_callstack_data_view_ = std::make_unique<CallStackDataView>();
+    selection_callstack_data_view_ = std::make_unique<CallStackDataView>(this);
     panels_.push_back(selection_callstack_data_view_.get());
   }
   return selection_callstack_data_view_.get();
--- a/OrbitGl/App.cpp
+++ b/OrbitGl/App.cpp
@@ -1558,7 +1558,7 @@ DataView* OrbitApp::GetOrCreateDataView(DataViewType type) {
 
     case DataViewType::kProcesses:
       if (!processes_data_view_) {
-        processes_data_view_ = std::make_unique<ProcessesDataView>();
+        processes_data_view_ = std::make_unique<ProcessesDataView>(this);
         processes_data_view_->SetSelectionListener(
             [&](int32_t pid) { UpdateProcessAndModuleList(pid); });
         panels_.push_back(processes_data_view_.get());
--- a/OrbitGl/App.cpp
+++ b/OrbitGl/App.cpp
@@ -118,6 +118,20 @@ OrbitApp::OrbitApp(ApplicationOptions&& options,
 }
 
 OrbitApp::~OrbitApp() {
+  AbortCapture();
+
+  try {
+    process_manager_->ShutdownAndWait();
+  } catch (const std::exception& e) {
+    FATAL("Exception occured in ProcessManager::ShutdownAndWait(): %s", e.what());
+  }
+
+  try {
+    thread_pool_->ShutdownAndWait();
+  } catch (const std::exception& e) {
+    FATAL("Exception occured in ThreadPool::ShutdownAndWait(): %s", e.what());
+  }
+
 #ifdef _WIN32
   oqpi::default_helpers::stop_scheduler();
 #endif
@@ -510,15 +524,6 @@ void OrbitApp::Disassemble(int32_t pid, const FunctionInfo& function) {
                              capture_data.GetCallstackData()->GetCallstackEventsCount());
     SendDisassemblyToUi(disasm.GetResult(), std::move(report));
   });
-}
-
-void OrbitApp::OnExit() {
-  AbortCapture();
-
-  process_manager_->ShutdownAndWait();
-  thread_pool_->ShutdownAndWait();
-
-  GOrbitApp = nullptr;
 }
 
 Timer GMainTimer;
--- a/OrbitGl/App.cpp
+++ b/OrbitGl/App.cpp
@@ -1584,7 +1584,7 @@ DataView* OrbitApp::GetOrCreateDataView(DataViewType type) {
 
     case DataViewType::kTracepoints:
       if (!tracepoints_data_view_) {
-        tracepoints_data_view_ = std::make_unique<TracepointsDataView>();
+        tracepoints_data_view_ = std::make_unique<TracepointsDataView>(this);
         panels_.push_back(tracepoints_data_view_.get());
       }
       return tracepoints_data_view_.get();
--- a/OrbitGl/App.cpp
+++ b/OrbitGl/App.cpp
@@ -580,7 +580,7 @@ void OrbitApp::SetSamplingReport(
     sampling_report_->ClearReport();
   }
 
-  auto report = std::make_shared<SamplingReport>(std::move(post_processed_sampling_data),
+  auto report = std::make_shared<SamplingReport>(this, std::move(post_processed_sampling_data),
                                                  std::move(unique_callstacks));
   CHECK(sampling_reports_callback_);
   DataView* callstack_data_view = GetOrCreateDataView(DataViewType::kCallstack);
@@ -599,7 +599,7 @@ void OrbitApp::SetSelectionReport(
     selection_report_->ClearReport();
   }
 
-  auto report = std::make_shared<SamplingReport>(std::move(post_processed_sampling_data),
+  auto report = std::make_shared<SamplingReport>(this, std::move(post_processed_sampling_data),
                                                  std::move(unique_callstacks), has_summary);
   DataView* callstack_data_view = GetOrCreateSelectionCallstackDataView();
 
--- a/OrbitGl/FunctionsDataView.h
+++ b/OrbitGl/FunctionsDataView.h
@@ -64,8 +64,6 @@ class FunctionsDataView : public DataView {
   static bool ShouldShowFrameTrackIcon(OrbitApp* app,
                                        const orbit_client_protos::FunctionInfo& function);
   std::vector<const orbit_client_protos::FunctionInfo*> functions_;
-
-  OrbitApp* app_ = nullptr;
 };
 
 #endif  // ORBIT_GL_FUNCTIONS_DATA_VIEW_H_
--- a/OrbitGl/FunctionsDataView.h
+++ b/OrbitGl/FunctionsDataView.h
@@ -8,14 +8,17 @@
 #include "DataView.h"
 #include "capture_data.pb.h"
 
+class OrbitApp;
+
 class FunctionsDataView : public DataView {
  public:
-  FunctionsDataView();
+  explicit FunctionsDataView(OrbitApp* app);
 
   static const std::string kUnselectedFunctionString;
   static const std::string kSelectedFunctionString;
   static const std::string kFrameTrackString;
-  static std::string BuildSelectedColumnsString(const orbit_client_protos::FunctionInfo& function);
+  static std::string BuildSelectedColumnsString(OrbitApp* app,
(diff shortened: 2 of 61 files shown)


OrbitGl/App.cpp
```diff
@@ -118,6 +118,11 @@ OrbitApp::OrbitApp(ApplicationOptions&& options,
 }
 
 OrbitApp::~OrbitApp() {
+  AbortCapture();
+
+  process_manager_->ShutdownAndWait();
+  thread_pool_->ShutdownAndWait();
```
dimitry-, line 123: are these noexcept?

beckerhe: No. Should I add a catch-all block? Maybe that's better but this would change behaviour. There was nothing that caught a potential exception before.

If these threw, this would end up in an abort since the destructor is noexcept by default.

dimitry-: I did not know d-tors are noexcept by default. I am thinking it might be better to have an explicit catch all -> FATAL block,

beckerhe: Added that. PTAL.

OrbitQt/orbitcodeeditor.cpp
```diff
@@ -61,7 +61,6 @@
 #include <QtWidgets>
 #include <fstream>
 
-#include "App.h"
```
florian-kuebler, line 64: Maybe we should remove the complete class (maybe in a different PR). WDYT?

OrbitGl/TracepointTrack.h
```diff
@@ -7,9 +7,9 @@
 
 #include "EventTrack.h"
 
-class TracepointTrack : public EventTrack {
+class TracepointTrack final : public EventTrack {
```
florian-kuebler, line 10: Any reason why this is made final? Fine with me, but seems unrelated.

florian-kuebler (approved):
Wow, thanks so much for this. This was on my todo list since our last fixit-week. 
On minor thing:
As you noted, we should really push forward removing the forward references here.
Can you create a bug for that and link this in this PR?

beckerhe:
> Wow, thanks so much for this. This was on my todo list since our last fixit-week.
> On minor thing:
> As you noted, we should really push forward removing the forward references here.
> Can you create a bug for that and link this in this PR?

Hmm. I don't think I have touched a single forwarding reference. Can you explain what you're talking about? Adn why should we remove forwarding references? I'm confused. :see_no_evil:

OrbitGl/TracepointTrack.h
beckerhe: First I added that to help with the refactoring. Then I've seen that all the other Track classes had this final and I left it there for completeness reasons. I can remove it if you prefer.

OrbitQt/orbitcodeeditor.cpp
beckerhe: Yes, this will be revamped with the source code view. So there are already a variety of bugs covering that.

OrbitGl/TracepointTrack.h
florian-kuebler: I thought I saw other tracks that are not final. But hey, fine with keeping this :)

florian-kuebler:
> > Wow, thanks so much for this. This was on my todo list since our last fixit-week.
> > On minor thing:
> > As you noted, we should really push forward removing the forward references here.
> > Can you create a bug for that and link this in this PR?
> 
> Hmm. I don't think I have touched a single forwarding reference. Can you explain what you're talking about? Adn why should we remove forwarding references? I'm confused. 🙈

Sorry, I wanted to say "forward declaration". You added those for OrbitApp, which is necessary for now. But should have a bug tracking to remember us to remove them.

OrbitQt/orbitcodeeditor.cpp
florian-kuebler: Fine. Thanks.

OrbitGl/TracepointTrack.h
beckerhe: Ah sorry. I should have explained that more clearly. Some tracks are inherited from, notabley TimerTrack and EventTrack. But all the tracks without subclasses had the final tag.

beckerhe:
> > > Wow, thanks so much for this. This was on my todo list since our last fixit-week.
> > > On minor thing:
> > > As you noted, we should really push forward removing the forward references here.
> > > Can you create a bug for that and link this in this PR?
> > 
> > 
> > Hmm. I don't think I have touched a single forwarding reference. Can you explain what you're talking about? Adn why should we remove forwarding references? I'm confused. see_no_evil
> 
> Sorry, I wanted to say "forward declaration". You added those for OrbitApp, which is necessary for now. But should have a bug tracking to remember us to remove them.

Ah sorry. Should have gotten that right away. So regarding adding a bug: I thought about this and think this is not feasible or at least not easily done, because like I said in the PR description: It's not about removing the forward-decl, it's about removing the cyclic dependency. And this needs to be tackled differently - for every class. So I think a single bug doesn't cut it. We would need a bug for every class, so some 30-ish additional bugs. I'm not sure if it is worth it creating those.

If we had a mechanical solution for all classes a single bug would work.

OrbitGl/TracepointTrack.h
florian-kuebler: Hmm I don't see the final on any Subclass of TimerTrack.

florian-kuebler:
> > > > Wow, thanks so much for this. This was on my todo list since our last fixit-week.
> > > > On minor thing:
> > > > As you noted, we should really push forward removing the forward references here.
> > > > Can you create a bug for that and link this in this PR?
> > > 
> > > 
> > > Hmm. I don't think I have touched a single forwarding reference. Can you explain what you're talking about? Adn why should we remove forwarding references? I'm confused. see_no_evil
> > 
> > 
> > Sorry, I wanted to say "forward declaration". You added those for OrbitApp, which is necessary for now. But should have a bug tracking to remember us to remove them.
> 
> Ah sorry. Should have gotten that right away. So regarding adding a bug: I thought about this and think this is not feasible or at least not easily done, because like I said in the PR description: It's not about removing the forward-decl, it's about removing the cyclic dependency. And this needs to be tackled differently - for every class. So I think a single bug doesn't cut it. We would need a bug for every class, so some 30-ish additional bugs. I'm not sure if it is worth it creating those.
> 
> If we had a mechanical solution for all classes a single bug would work.
> 
> But I'm open for discussion how we could represent that work in a bug.

What issue do you see in having one bug to bundle all? Even each class needs to be tackled differently. The bug is basically, that we don't forget about this. I agree that 30-ish additional bugs would be a bit much.

OrbitGl/TracepointTrack.h
beckerhe: Hmmm. Maybe I'm completely mistaken. Let me just remove it.

dimitry- (approved):
lgtm!

beckerhe:
> > > > > Wow, thanks so much for this. This was on my todo list since our last fixit-week.
> > > > > On minor thing:
> > > > > As you noted, we should really push forward removing the forward references here.
> > > > > Can you create a bug for that and link this in this PR?
> > > > 
> > > > 
> > > > Hmm. I don't think I have touched a single forwarding reference. Can you explain what you're talking about? Adn why should we remove forwarding references? I'm confused. see_no_evil
> > > 
> > > 
> > > Sorry, I wanted to say "forward declaration". You added those for OrbitApp, which is necessary for now. But should have a bug tracking to remember us to remove them.
> > 
> > 
> > Ah sorry. Should have gotten that right away. So regarding adding a bug: I thought about this and think this is not feasible or at least not easily done, because like I said in the PR description: It's not about removing the forward-decl, it's about removing the cyclic dependency. And this needs to be tackled differently - for every class. So I think a single bug doesn't cut it. We would need a bug for every class, so some 30-ish additional bugs. I'm not sure if it is worth it creating those.
> > If we had a mechanical solution for all classes a single bug would work.
> > But I'm open for discussion how we could represent that work in a bug.
> 
> What issue do you see in having one bug to bundle all? Even each class needs to be tackled differently. The bug is basically, that we don't forget about this. I agree that 30-ish additional bugs would be a bit much.

I'm a bit worried that this will be just one bug which will be around forever. Nobody will ever assign this bug to theirselves because the scope is so huge. Do we have similar bugs? If we are already doing this I will be more comfortable to add another bug in the same manner.

florian-kuebler:
> > > > > > Wow, thanks so much for this. This was on my todo list since our last fixit-week.
> > > > > > On minor thing:
> > > > > > As you noted, we should really push forward removing the forward references here.
> > > > > > Can you create a bug for that and link this in this PR?
> > > > > 
> > > > > 
> > > > > Hmm. I don't think I have touched a single forwarding reference. Can you explain what you're talking about? Adn why should we remove forwarding references? I'm confused. see_no_evil
> > > > 
> > > > 
> > > > Sorry, I wanted to say "forward declaration". You added those for OrbitApp, which is necessary for now. But should have a bug tracking to remember us to remove them.
> > > 
> > > 
> > > Ah sorry. Should have gotten that right away. So regarding adding a bug: I thought about this and think this is not feasible or at least not easily done, because like I said in the PR description: It's not about removing the forward-decl, it's about removing the cyclic dependency. And this needs to be tackled differently - for every class. So I think a single bug doesn't cut it. We would need a bug for every class, so some 30-ish additional bugs. I'm not sure if it is worth it creating those.
> > > If we had a mechanical solution for all classes a single bug would work.
> > > But I'm open for discussion how we could represent that work in a bug.
> > 
> > 
> > What issue do you see in having one bug to bundle all? Even each class needs to be tackled differently. The bug is basically, that we don't forget about this. I agree that 30-ish additional bugs would be a bit much.
> 
> I'm a bit worried that this will be just one bug which will be around forever. Nobody will ever assign this bug to theirselves because the scope is so huge. Do we have similar bugs? If we are already doing this I will be more comfortable to add another bug in the same manner.

IIRC, we had such a bug to get rid of the capture globals. I see you fear, that this bug is around forever, but I guess we can add in the description of the bug, that this needs to be tackled in different ways. Then whenever, someone wants to tackle one of the classes, that person creates a child bug for that specific class. WDYT?

beckerhe:
> > > > > > > Wow, thanks so much for this. This was on my todo list since our last fixit-week.
> > > > > > > On minor thing:
> > > > > > > As you noted, we should really push forward removing the forward references here.
> > > > > > > Can you create a bug for that and link this in this PR?
> > > > > > 
> > > > > > 
> > > > > > Hmm. I don't think I have touched a single forwarding reference. Can you explain what you're talking about? Adn why should we remove forwarding references? I'm confused. see_no_evil
> > > > > 
> > > > > 
> > > > > Sorry, I wanted to say "forward declaration". You added those for OrbitApp, which is necessary for now. But should have a bug tracking to remember us to remove them.
> > > > 
> > > > 
> > > > Ah sorry. Should have gotten that right away. So regarding adding a bug: I thought about this and think this is not feasible or at least not easily done, because like I said in the PR description: It's not about removing the forward-decl, it's about removing the cyclic dependency. And this needs to be tackled differently - for every class. So I think a single bug doesn't cut it. We would need a bug for every class, so some 30-ish additional bugs. I'm not sure if it is worth it creating those.
> > > > If we had a mechanical solution for all classes a single bug would work.
> > > > But I'm open for discussion how we could represent that work in a bug.
> > > 
> > > 
> > > What issue do you see in having one bug to bundle all? Even each class needs to be tackled differently. The bug is basically, that we don't forget about this. I agree that 30-ish additional bugs would be a bit much.
> > 
> > 
> > I'm a bit worried that this will be just one bug which will be around forever. Nobody will ever assign this bug to theirselves because the scope is so huge. Do we have similar bugs? If we are already doing this I will be more comfortable to add another bug in the same manner.
> 
> IIRC, we had such a bug to get rid of the capture globals. I see you fear, that this bug is around forever, but I guess we can add in the description of the bug, that this needs to be tackled in different ways. Then whenever, someone wants to tackle one of the classes, that person creates a child bug for that specific class. WDYT?

Alright. I guees we reached a conclusion. This sounds good to me. :slightly_smiling_face: I will create the bug right away.

florian-kuebler:
> > > > > > > > Wow, thanks so much for this. This was on my todo list since our last fixit-week.
> > > > > > > > On minor thing:
> > > > > > > > As you noted, we should really push forward removing the forward references here.
> > > > > > > > Can you create a bug for that and link this in this PR?
> > > > > > > 
> > > > > > > 
> > > > > > > Hmm. I don't think I have touched a single forwarding reference. Can you explain what you're talking about? Adn why should we remove forwarding references? I'm confused. see_no_evil
> > > > > > 
> > > > > > 
> > > > > > Sorry, I wanted to say "forward declaration". You added those for OrbitApp, which is necessary for now. But should have a bug tracking to remember us to remove them.
> > > > > 
> > > > > 
> > > > > Ah sorry. Should have gotten that right away. So regarding adding a bug: I thought about this and think this is not feasible or at least not easily done, because like I said in the PR description: It's not about removing the forward-decl, it's about removing the cyclic dependency. And this needs to be tackled differently - for every class. So I think a single bug doesn't cut it. We would need a bug for every class, so some 30-ish additional bugs. I'm not sure if it is worth it creating those.
> > > > > If we had a mechanical solution for all classes a single bug would work.
> > > > > But I'm open for discussion how we could represent that work in a bug.
> > > > 
> > > > 
> > > > What issue do you see in having one bug to bundle all? Even each class needs to be tackled differently. The bug is basically, that we don't forget about this. I agree that 30-ish additional bugs would be a bit much.
> > > 
> > > 
> > > I'm a bit worried that this will be just one bug which will be around forever. Nobody will ever assign this bug to theirselves because the scope is so huge. Do we have similar bugs? If we are already doing this I will be more comfortable to add another bug in the same manner.
> > 
> > 
> > IIRC, we had such a bug to get rid of the capture globals. I see you fear, that this bug is around forever, but I guess we can add in the description of the bug, that this needs to be tackled in different ways. Then whenever, someone wants to tackle one of the classes, that person creates a child bug for that specific class. WDYT?
> 
> Alright. I guees we reached a conclusion. This sounds good to me. 🙂 I will create the bug right away.

Thanks!

beckerhe:
> > > > > > > > > Wow, thanks so much for this. This was on my todo list since our last fixit-week.
> > > > > > > > > On minor thing:
> > > > > > > > > As you noted, we should really push forward removing the forward references here.
> > > > > > > > > Can you create a bug for that and link this in this PR?
> > > > > > > > 
> > > > > > > > 
> > > > > > > > Hmm. I don't think I have touched a single forwarding reference. Can you explain what you're talking about? Adn why should we remove forwarding references? I'm confused. see_no_evil
> > > > > > > 
> > > > > > > 
> > > > > > > Sorry, I wanted to say "forward declaration". You added those for OrbitApp, which is necessary for now. But should have a bug tracking to remember us to remove them.
> > > > > > 
> > > > > > 
> > > > > > Ah sorry. Should have gotten that right away. So regarding adding a bug: I thought about this and think this is not feasible or at least not easily done, because like I said in the PR description: It's not about removing the forward-decl, it's about removing the cyclic dependency. And this needs to be tackled differently - for every class. So I think a single bug doesn't cut it. We would need a bug for every class, so some 30-ish additional bugs. I'm not sure if it is worth it creating those.
> > > > > > If we had a mechanical solution for all classes a single bug would work.
> > > > > > But I'm open for discussion how we could represent that work in a bug.
> > > > > 
> > > > > 
> > > > > What issue do you see in having one bug to bundle all? Even each class needs to be tackled differently. The bug is basically, that we don't forget about this. I agree that 30-ish additional bugs would be a bit much.
> > > > 
> > > > 
> > > > I'm a bit worried that this will be just one bug which will be around forever. Nobody will ever assign this bug to theirselves because the scope is so huge. Do we have similar bugs? If we are already doing this I will be more comfortable to add another bug in the same manner.
> > > 
> > > 
> > > IIRC, we had such a bug to get rid of the capture globals. I see you fear, that this bug is around forever, but I guess we can add in the description of the bug, that this needs to be tackled in different ways. Then whenever, someone wants to tackle one of the classes, that person creates a child bug for that specific class. WDYT?
> > 
> > 
> > Alright. I guees we reached a conclusion. This sounds good to me. slightly_smiling_face I will create the bug right away.
> 
> Thanks!

Done. http://b/175759190 Also added to the PR description.

beckerhe:
Heads up, I will remove the fixup commits now, force-push and merge after CI has finished.

If you haven't reviewed this yet and still want to do it, please speak up now.

Florian and Dimitry, thanks a lot for the quick review.

OrbitGl/FunctionsDataView.h
```diff
@@ -8,14 +8,17 @@
 #include "DataView.h"
 #include "capture_data.pb.h"
 
+class OrbitApp;
+
 class FunctionsDataView : public DataView {
  public:
-  FunctionsDataView();
+  FunctionsDataView(OrbitApp* app);
```
dpallotti, line 15: Maybe this should be `explicit` now?

OrbitGl/LiveFunctionsController.h
```diff
@@ -13,9 +13,11 @@
 #include "absl/container/flat_hash_map.h"
 #include "capture_data.pb.h"
 
+class OrbitApp;
+
 class LiveFunctionsController {
  public:
-  LiveFunctionsController() : live_functions_data_view_(this) {}
+  LiveFunctionsController(OrbitApp* app) : live_functions_data_view_(this, app), app_{app} {}
```
dpallotti, line 20: `explicit`?

OrbitGl/PresetsDataView.h
```diff
@@ -10,9 +10,11 @@
 #include "DataView.h"
 #include "preset.pb.h"
 
+class OrbitApp;
+
 class PresetsDataView : public DataView {
  public:
-  PresetsDataView();
+  PresetsDataView(OrbitApp* app);
```
dpallotti, line 17: `explicit`?

OrbitGl/TracepointsDataView.h
```diff
@@ -14,9 +14,11 @@
 
 using orbit_grpc_protos::TracepointInfo;
 
+class OrbitApp;
+
 class TracepointsDataView : public DataView {
  public:
-  TracepointsDataView();
+  TracepointsDataView(OrbitApp* app);
```
dpallotti, line 21: `explicit`?

dpallotti (approved):
Only skimmed through it and added a couple of comments about `explicit`.

Mostly I wanted to say thank you for the change!!

OrbitGl/FunctionsDataView.h
beckerhe: Yesss. Thanks a lot. I missed that in quite a few locations - it seems! Fixing that now.

beckerhe: Done. Thanks again!

OrbitGl/LiveFunctionsController.h
beckerhe: Done. Thanks!

OrbitGl/PresetsDataView.h
beckerhe: Done. Thanks!

OrbitGl/TracepointsDataView.h
beckerhe: Done. Thanks!

beckerhe:
> Only skimmed through it and added a couple of comments about `explicit`.
> 
> Mostly I wanted to say thank you for the change!!

I even found another missing explicit! Thanks a lot for pointing this out.

PR #1598, opened 2020-12-16
Add Entry Points for Orbit Vulkan Layer

florian-kuebler:
This adds the entry points to the Vulkan layer. It is the entry point
that will be hooked up by the Vulkan loaded.

Bug: http://b/168200502
Test: Compile, push to the instance, run a game.

**Note: #1593 needs to be in first**


OrbitVulkanLayer/Main.cpp
```diff
@@ -0,0 +1,270 @@
...
+}
+
+// ----------------------------------------------------------------------------
+// GetProcAddr functions
+// ----------------------------------------------------------------------------
+
+#define ORBIT_GETPROCADDR(func)                              \
+  if (strcmp(name, "vk" #func) == 0) {                       \
+    return absl::bit_cast<PFN_vkVoidFunction>(&Orbit##func); \
+  }
```
dpallotti, line 195: If we want to keep this macro to do the `if` instead of something else like filling a data structure, I still think that it could be renamed to tell what it's doing. For the name, maybe something like "RETURN_ORBIT_FUNCTION_IF_MATCHES_VK_FUNCTION`? Ugly but at least more descriptive. Better ideas are for sure possible.

Also, I would personally wrap it in a `do {} while(false)` so that it requires to be followed by a semicolon.

OrbitVulkanLayer/Main.cpp
```diff
@@ -0,0 +1,270 @@
+// Copyright (c) 2020 The Orbit Authors. All rights reserved.
```
dpallotti, line 1: This file obviously doesn't contain a `main` function as this is a shared library, but that's what one would expect from a file named `Main.cpp`. I don't really have an idea but could we find a better name for the file?

dpallotti:
The commit message refers to a "main class" but there's no class involved here.

OrbitVulkanLayer/Main.cpp
florian-kuebler: In the compiler/static analysis community, those are typically called "entry points". So Maybe `VulkanLayerEntryPoints`? What do you think?

dpallotti: It sounds good, it matches what they actually are. It could also be just `EntryPoints`.

OrbitVulkanLayer/Main.cpp
```diff
@@ -0,0 +1,270 @@
...
+ * OrbitGetDeviceProcAddr and OrbitGetInstanceProcAddr are the actual entry points, called by
+ * the loader and potential other layers, and forward to all the functions that this layer
+ * intercepts. All other functions, will be referenced by those two lookup functions.
+ */
+namespace orbit_vulkan_layer {
+
+#if defined(WIN32)
+#define ORBIT_EXPORT extern "C" __declspec(dllexport) VK_LAYER_EXPORT
+#else
+#define ORBIT_EXPORT extern "C" VK_LAYER_EXPORT
```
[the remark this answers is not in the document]
florian-kuebler: Well, you are right. I just added the extern "C" VK_LAYER_EXPORT on the two functions.

OrbitVulkanLayer/Main.cpp
```diff
@@ -0,0 +1,270 @@
...
+#endif
+
+using DeviceMangerImpl = DeviceManager<DispatchTable>;
+using TimerQueryPoolImpl = TimerQueryPool<DispatchTable>;
+using SubmissionTrackerImpl =
+    SubmissionTracker<DispatchTable, DeviceMangerImpl, TimerQueryPoolImpl>;
+static VulkanLayerController<DispatchTable, QueueManager, DeviceMangerImpl, TimerQueryPoolImpl,
+                             SubmissionTrackerImpl>
+    controller_;
```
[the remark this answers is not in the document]
florian-kuebler: I went for controller.

OrbitVulkanLayer/Main.cpp
```diff
@@ -0,0 +1,270 @@
...
+
+// ----------------------------------------------------------------------------
+// Layer bootstrapping code
+// ----------------------------------------------------------------------------
+
+VKAPI_ATTR VkResult VKAPI_CALL OrbitCreateInstance(const VkInstanceCreateInfo* create_info,
+                                                   const VkAllocationCallbacks* allocator,
+                                                   VkInstance* instance) {
+  VkResult result = controller_.OnCreateInstance(create_info, allocator, instance);
+  return result;
```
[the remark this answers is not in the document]
florian-kuebler: right, missed that. thanks.

OrbitVulkanLayer/Main.cpp
```diff
@@ -0,0 +1,270 @@
...
+// Layer enumeration functions
+// ----------------------------------------------------------------------------
+
+VKAPI_ATTR VkResult VKAPI_CALL
+OrbitEnumerateInstanceLayerProperties(uint32_t* property_count, VkLayerProperties* properties) {
+  return controller_.OnEnumerateInstanceLayerProperties(property_count, properties);
+}
+
+// Deprecated by Khronos, but we'll support it in case older applications still
+// use it.
```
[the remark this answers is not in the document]
florian-kuebler: Thanks

OrbitVulkanLayer/Main.cpp
```diff
@@ -0,0 +1,270 @@
...
+
+#define ORBIT_GETPROCADDR(func)                              \
+  if (strcmp(name, "vk" #func) == 0) {                       \
+    return absl::bit_cast<PFN_vkVoidFunction>(&Orbit##func); \
+  }
+
+ORBIT_EXPORT VKAPI_ATTR PFN_vkVoidFunction VKAPI_CALL OrbitGetDeviceProcAddr(VkDevice device,
+                                                                             const char* name) {
+  // Functions available through GetInstanceProcAddr and GetDeviceProcAddr
+  ORBIT_GETPROCADDR(GetDeviceProcAddr)
```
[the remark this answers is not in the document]
florian-kuebler: Well there will be a return in each "if", right? Not sure, if a map would be much better. This also seem to be the way its done in Vulkan.

OrbitVulkanLayer/Main.cpp
florian-kuebler: Agree.

OrbitVulkanLayer/Main.cpp
```diff
@@ -0,0 +1,270 @@
...
+  ORBIT_GETPROCADDR(QueuePresentKHR)
+  ORBIT_GETPROCADDR(GetDeviceQueue)
+  ORBIT_GETPROCADDR(GetDeviceQueue2)
+
+  ORBIT_GETPROCADDR(CmdBeginDebugUtilsLabelEXT)
+  ORBIT_GETPROCADDR(CmdEndDebugUtilsLabelEXT)
+  ORBIT_GETPROCADDR(CmdDebugMarkerBeginEXT)
+  ORBIT_GETPROCADDR(CmdDebugMarkerEndEXT)
+
+  return controller_.OnGetDeviceProcAddr(device, name);
```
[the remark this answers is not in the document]
florian-kuebler: How about `ForwardGetInstanceProcAddr`..?

florian-kuebler (commented):
Thanks for the review. Addressed most of the comments.

OrbitVulkanLayer/Main.cpp
florian-kuebler: Went for EntryPoints

OrbitVulkanLayer/EntryPoints.cpp
```diff
@@ -0,0 +1,263 @@
...
+#include "TimerQueryPool.h"
+#include "VulkanLayerController.h"
+#include "absl/base/casts.h"
+#include "vulkan/vk_layer.h"
+#include "vulkan/vulkan.h"
+
+/*
+ * The big picture:
+ * This is the main entry point for Orbit's Vulkan layer. The layer is structured as follows:
+ * ° All instrumented Vulkan functions will hook into implementations found here
```
ronaldfw, line 18: Use "-" instead. (This is the "degree" character, right? That shouldn't be used as a bullet point.)

OrbitVulkanLayer/EntryPoints.cpp
```diff
@@ -0,0 +1,263 @@
...
+ *   (e.g. OrbitQueueSubmit) which will delegate to the `VulkanLayerController`.
+ * ° For every `vkX` function, this controller has an `OnX` functions that will perform the actual
+ *   Vulkan call (using DispatchTable), but also glue together the logic of the layer.
+ * ° These are the following helper classes to structure the actual layer logic:
+ *   ° SubmissionTracker: Which is the heart of the layer logic. It keeps track of
```
ronaldfw, line 23: Possibly use "--" as sub-bullet points.

OrbitVulkanLayer/EntryPoints.cpp
```diff
@@ -0,0 +1,263 @@
...
+ *     command buffer usages and timings, debug markers and submissions.
+ *   ° DispatchTable: It provides virtual dispatch for the Vulkan functions to be called.
+ *   ° TimerQueryPool: Which keeps track of query pool slots used for timestamp queries and allows
+ *     to assign those.
+ *   ° VulkanLayerProducer: This is the producer used for the IPC with Orbit. Results will be sent
+ *     by this as CaptureEvent protos.
+ *   ° DeviceManager: It tracks the association of a VkDevice to the VkPhysicalDevice.
+ *   ° QueueManager: It keeps track of association of VkQueue(s) to devices.
+ *
+ *
```
ronaldfw, line 33: Remove empty line.

OrbitVulkanLayer/EntryPoints.cpp
```diff
@@ -0,0 +1,263 @@
...
+ * The free functions in this namespace act as entry points to the layer.
+ * OrbitGetDeviceProcAddr and OrbitGetInstanceProcAddr are the actual entry points, called by
+ * the loader and potential other layers, and return pointers to the functions that this layer
```
ronaldfw, line 36:
```suggestion
 * the loader and potential other layers. They return pointers to the functions that this layer
```

OrbitVulkanLayer/EntryPoints.cpp
```diff
@@ -0,0 +1,263 @@
...
+ * intercepts. All other functions will be referenced by those two lookup functions.
```
ronaldfw, line 37: I don't understand this sentence: "All other functions will be referenced by those two lookup functions."

ronaldfw (approved):
LGTM, just some minor comments!

OrbitVulkanLayer/EntryPoints.cpp
```diff
@@ -0,0 +1,263 @@
...
+  RETURN_ORBIT_FUNCTION_IF_MATCHES_VK_FUNCTION(CmdEndDebugUtilsLabelEXT);
+  RETURN_ORBIT_FUNCTION_IF_MATCHES_VK_FUNCTION(CmdDebugMarkerBeginEXT);
+  RETURN_ORBIT_FUNCTION_IF_MATCHES_VK_FUNCTION(CmdDebugMarkerEndEXT);
+
+  return controller.ForwardGetInstanceProcAddr(instance, name);
+}
+
+#undef RETURN_ORBIT_FUNCTION_IF_MATCHES_VK_FUNCTION
+
+#undef ORBIT_EXPORT
```
dpallotti, line 261: This is no longer defined I think.

OrbitVulkanLayer/EntryPoints.cpp
```diff
@@ -0,0 +1,263 @@
...
+#include "vulkan/vulkan.h"
+
+/*
+ * The big picture:
+ * This is the main entry point for Orbit's Vulkan layer. The layer is structured as follows:
+ * ° All instrumented Vulkan functions will hook into implementations found here
+ *   (e.g. OrbitQueueSubmit) which will delegate to the `VulkanLayerController`.
+ * ° For every `vkX` function, this controller has an `OnX` functions that will perform the actual
+ *   Vulkan call (using DispatchTable), but also glue together the logic of the layer.
+ * ° These are the following helper classes to structure the actual layer logic:
```
dpallotti, line 22: Remove "following"?

dpallotti (approved):

OrbitVulkanLayer/Main.cpp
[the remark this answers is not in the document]
dpallotti: Looks good now.

OrbitVulkanLayer/Main.cpp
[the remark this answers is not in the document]
dpallotti: Sounds good. Another possibility is to add "ToNextLayer" (`ForwardGetInstanceProcAddrToNextLayer`). Not sure if necessary though.

OrbitVulkanLayer/EntryPoints.cpp
florian-kuebler: Done. Thanks

OrbitVulkanLayer/EntryPoints.cpp
florian-kuebler: Done.

OrbitVulkanLayer/EntryPoints.cpp
florian-kuebler: Good suggestion. Done.

OrbitVulkanLayer/EntryPoints.cpp
florian-kuebler: Done.

OrbitVulkanLayer/EntryPoints.cpp
florian-kuebler: Done.

OrbitVulkanLayer/EntryPoints.cpp
florian-kuebler: Rephrased it.

OrbitVulkanLayer/EntryPoints.cpp
florian-kuebler: Good catch. Thanks!

florian-kuebler (commented):
Thanks for the review. I addressed the comments.

ronaldfw (approved):
LGTM

PR #1602, opened 2020-12-16
Introduce CreateModule function to LinuxMap.h

dimitry-:
Extract CreateModule which creates a module based on mapping start/end
address and the elf file path.

Also use std::filesystem::file_size instead of FileSize.

Test: run ElfUtilsTests

### DIFF
--- a/ElfUtils/LinuxMap.cpp
+++ b/ElfUtils/LinuxMap.cpp
@@ -6,29 +6,57 @@
 
 #include <absl/strings/str_format.h>
 #include <absl/strings/str_split.h>
-#include <sys/stat.h>
-#include <unistd.h>
 
 #include <filesystem>
 
 #include "ElfUtils/ElfFile.h"
 #include "OrbitBase/Logging.h"
 #include "OrbitBase/ReadFileToString.h"
-#include "OrbitBase/SafeStrerror.h"
 
 namespace orbit_elf_utils {
 
 using orbit_elf_utils::ElfFile;
 using orbit_grpc_protos::ModuleInfo;
 
-static ErrorMessageOr<uint64_t> FileSize(const std::string& file_path) {
-  struct stat stat_buf {};
-  int ret = stat(file_path.c_str(), &stat_buf);
-  if (ret != 0) {
-    return ErrorMessage(absl::StrFormat("Unable to call stat with file \"%s\": %s", file_path,
-                                        SafeStrerror(errno)));
+ErrorMessageOr<ModuleInfo> CreateModule(const std::filesystem::path& module_path,
+                                        uint64_t start_address, uint64_t end_address) {
+  // This excludes mapped character or block devices.
+  if (absl::StartsWith(module_path.string(), "/dev/")) {
+    return ErrorMessage(absl::StrFormat(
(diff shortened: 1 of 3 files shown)


ElfUtils/LinuxMap.cpp
```diff
@@ -21,14 +21,45 @@ namespace orbit_elf_utils {
...
+
+  ErrorMessageOr<std::unique_ptr<ElfFile>> elf_file = ElfFile::Create(module_path);
+  if (!elf_file) {
+    return ErrorMessage(absl::StrFormat("Unable to load module: %s", elf_file.error().message()));
+  }
+
+  ErrorMessageOr<uint64_t> load_bias = elf_file.value()->GetLoadBias();
+  // Every loadable module contains a load bias.
+  if (!load_bias) {
+    return ErrorMessage(absl::StrFormat("No load bias found for module %s", module_path));
```
dpallotti, line 50: I suggest `: %s` or `\"%s\"` like above.

ElfUtils/LinuxMap.cpp
```diff
@@ -80,35 +108,15 @@ ErrorMessageOr<std::vector<ModuleInfo>> ParseMaps(std::string_view proc_maps_dat
...
-      ERROR("Unable to load module \"%s\": %s - will ignore.", module_path,
-            elf_file.error().message());
-      continue;
-    }
 
-    ErrorMessageOr<uint64_t> load_bias = elf_file.value()->GetLoadBias();
-    // Every loadable module contains a load bias.
-    if (!load_bias) {
-      ERROR("No load bias found for module %s", module_path.c_str());
+    auto module_info =
```
dpallotti, line 112: I generally prefer seeing the type rather than `auto`: for example just from the name it's unclear that `CreateModule` returns `ErrorMessageOr`.

ElfUtils/LinuxMapTest.cpp
```diff
@@ -3,13 +3,77 @@
...
+  EXPECT_EQ(result.value().address_end(), kEndAddress);
+  EXPECT_EQ(result.value().build_id(), "d12d54bc5b72ccce54a408bdeda65e2530740ac8");
+  EXPECT_EQ(result.value().load_bias(), 0x0);
+}
+
+TEST(LinuxMap, CreateModuleOnDev) {
+  using orbit_elf_utils::CreateModule;
+  using orbit_grpc_protos::ModuleInfo;
+
+  const std::filesystem::path hello_world_path = "/dev/zero";
```
dpallotti, line 39: I guess that this variable should be renamed. For example `dev_zero_path`.

dpallotti (approved):

ElfUtils/LinuxMap.cpp
dimitry-: changed to use load_bias.error() here.

ElfUtils/LinuxMap.cpp
dimitry-: done.

ElfUtils/LinuxMapTest.cpp
dimitry-: done.

PR #1603, opened 2020-12-17
Make Qt widgets deinitializable

beckerhe:
You all know the problem, that Qt-classes that need to be declared in a UI-file may not have additional constructor arguments.

The solution to this is having a `Initialize` method which most of these classes have. The `Initialize` method passes pointers
to other resources into the widget (OrbitApp, DataView, etc.) such that the widget can call methods of these resources.

This becomes problematic during destruction. Widgets are destructed by Qt's QObject-based lifetime handling mechanism, which
means OrbitMainWindow's widgets are destructed in QObject's destructor - hence OrbitMainWindow does not exist anymore when
its widgets are destructed.

That's a problem for resources living in OrbitMainWindow and required by these widgets.

To my knowledge the best solution to this problem without refactoring everything is making the `Initialize` method symmetric and
also provide a `Deinitialize` method. So `Initialize` is called in OrbitMainWindows's constructor while `Deinitialize` is called in its destructor. This is what is implemented here.

### DIFF
--- a/OrbitQt/orbitmainwindow.cpp
+++ b/OrbitQt/orbitmainwindow.cpp
@@ -630,8 +630,8 @@ void OrbitMainWindow::OnTimer() {
   ORBIT_SCOPE("OrbitMainWindow::OnTimer");
   app_->MainTick();
 
-  for (OrbitGLWidget* glWidget : m_GlWidgets) {
-    glWidget->update();
+  for (OrbitGLWidget* gl_widget : gl_widgets_) {
+    gl_widget->update();
   }
 
   ui->timerLabel->setText(QString::fromStdString(app_->GetCaptureTime()));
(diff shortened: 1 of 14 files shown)


OrbitQt/orbitmainwindow.cpp
```diff
@@ -630,8 +630,8 @@ void OrbitMainWindow::OnTimer() {
   ORBIT_SCOPE("OrbitMainWindow::OnTimer");
   app_->MainTick();
 
-  for (OrbitGLWidget* glWidget : m_GlWidgets) {
-    glWidget->update();
+  for (OrbitGLWidget* gl_widget : gl_widgets_) {
+    gl_widget->update();
```
dimitry-, line 647: is the order important here?

dimitry- (approved):

OrbitQt/orbitmainwindow.cpp
beckerhe: Good question. Probably not. In our current use case only one widget will be registered anyway - the capture window.

This should probably be refactored. Instead of giving OrbitGLWidget a pointer to OrbitMainWindow for calling RegisterGLWidget, OrbitMainWindow should maintain this list by itself. But I felt that this is out of scope of this PR.

PR #1604, opened 2020-12-17
Also LOG count of thread state slices in TracerThread

dpallotti:


OrbitLinuxTracing/TracerThread.cpp
```diff
@@ -1102,6 +1103,9 @@ void TracerThread::PrintStatsIfTimerElapsed() {
         discarded_samples_in_uretprobes_count / actual_window_s,
         discarded_samples_in_uretprobes_count,
         100.0 * discarded_samples_in_uretprobes_count / stats_.sample_count);
+    uint64_t thread_state_count = stats_.thread_state_count;
+    LOG("  target's thread states: %.0f/s (%lu)", thread_state_count / actual_window_s,
```
dimitry-, line 1107: maybe check before L1072 that actual_window_s is not 0, we are dividing by it a lot here.

dimitry- (approved):
otherwise lgtm

OrbitLinuxTracing/TracerThread.cpp
dpallotti: With "check" you mean `CHECK` or `if`? Because that's already guaranteed by the `if`, so there's no need for another one.  (By the way, worst case you get NaN.)

dimitry-: yes - CHECK() my understanding is that is UB, has anything changed in the standard recently? also stats_.sample_count above should also be checked.

ronaldfw: It seems the C++ standard does not guarantee IEEE 754 floating point. That's a bit disappointing.

ronaldfw: There's a way to check it though: 
https://en.cppreference.com/w/cpp/types/numeric_limits/is_iec559

dpallotti: Well, that's unexpected, although it would be more unexpected if this was actually a problem with any of our compilers.
Anyway, I added the `CHECK` for `actual_window_s`. As for `stats_.sample_count`, than can be zero (and it can be forced to be zero by just using `-sampling_rate=0`), so I'm adding an `if` there.

dimitry- (approved):

OrbitLinuxTracing/TracerThread.cpp
ronaldfw: I guess it's because the C++ standard must also be applicable for implementing C++ on hardware where support for IEEE 754 is not available.

....Well, look at that for some examples:
https://stackoverflow.com/questions/6971886/exotic-architectures-the-standards-committees-care-about

dpallotti: Yeah, that surely makes sense, I guess there's a lot of stuff in the standard to support exotic architectures. I'm wondering if we should just assume IEEE 754 for our case. Or we just adhere to the standard and check for != 0, it shouldn't happen too often after all that we rely on a division by zero to behave well and actually give +/-Inf/NaN.

PR #1607, opened 2020-12-21
Accessible tracks

reichlfl:
Adds accessibility information for `CaptureWindow`, `TimeGraph` and `Track`.

I've decided to not introduce a new base class for our UI elements for now, and instead create custom accessibility interfaces for this. We may want to introduce this later nevertheless to handle at least parent / child relationships and maybe relative positioning, but in this case it will be straightforward to generalize these existing interfaces.

This PR is split into multiple commits that can be reviewed one by one.

### DIFF
--- a/OrbitGl/AccessibleTimeGraph.cpp
+++ b/OrbitGl/AccessibleTimeGraph.cpp
@@ -1,0 +1,29 @@
+// Copyright (c) 2020 The Orbit Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#include "AccessibleTimeGraph.h"
+
+#include "GlCanvas.h"
+#include "TimeGraph.h"
+
+AccessibilityRect TimeGraphAccessibility::AccessibleLocalRect() const {
+  GlCanvas* canvas = time_graph_->GetCanvas();
+  return AccessibilityRect(0, 0, canvas->GetWidth(), canvas->GetHeight());
+}
+
+AccessibilityState TimeGraphAccessibility::AccessibleState() const {
+  return orbit_gl::AccessibilityState::Focusable;
+}
+
+int TimeGraphAccessibility::AccessibleChildCount() const {
+  return time_graph_->GetTrackManager()->GetVisibleTracks().size();
+}
+
+const AccessibleInterface* TimeGraphAccessibility::AccessibleChild(int index) const {
+  return time_graph_->GetTrackManager()->GetVisibleTracks()[index]->AccessibilityInterface();
+}
+
+const AccessibleInterface* TimeGraphAccessibility::AccessibleParent() const {
+  return time_graph_->GetCanvas()->GetOrCreateAccessibleInterface();
+}
--- a/OrbitGl/CMakeLists.txt
+++ b/OrbitGl/CMakeLists.txt
@@ -11,7 +11,8 @@ target_compile_options(OrbitGl PRIVATE ${STRICT_COMPILE_FLAGS})
 
 target_sources(
   OrbitGl
-  PUBLIC App.h
+  PUBLIC AccessibleTimeGraph.h
+         App.h
          AsyncTrack.h
          Batcher.h
          BlockChain.h
@@ -70,7 +71,8 @@ target_sources(
 
 target_sources(
   OrbitGl
-  PRIVATE App.cpp
+  PRIVATE AccessibleTimeGraph.cpp
+          App.cpp
           AsyncTrack.cpp
           Batcher.cpp
           CallStackDataView.cpp
--- a/OrbitGl/CMakeLists.txt
+++ b/OrbitGl/CMakeLists.txt
@@ -154,7 +154,7 @@ target_sources(OrbitGlTests PRIVATE
                ScopedStatusTest.cpp
                SliderTest.cpp
                TimerInfosIteratorTest.cpp
-               UnitTestMockEnvironment.cpp)
+               ClientFlags.cpp)
 
 target_link_libraries(
   OrbitGlTests
--- a/OrbitGl/CMakeLists.txt
+++ b/OrbitGl/CMakeLists.txt
@@ -62,6 +62,7 @@ target_sources(
          TimerInfosIterator.h
          TimerTrack.h
          Track.h
+         TrackAccessibility.h
          TrackManager.h
          TriangleToggle.h
          TracepointsDataView.h
@@ -110,6 +111,7 @@ target_sources(
           ThreadStateTrack.cpp
           ThreadTrack.cpp
           Track.cpp
+          TrackAccessibility.cpp
           TrackManager.cpp
           TriangleToggle.cpp
           TracepointsDataView.cpp
--- a/OrbitGl/GlCanvas.cpp
+++ b/OrbitGl/GlCanvas.cpp
@@ -138,6 +138,13 @@ void GlCanvas::EnableImGui() {
   }
 }
 
+orbit_accessibility::AccessibleInterface* GlCanvas::GetOrCreateAccessibleInterface() {
+  if (accessibility_ == nullptr) {
+    accessibility_ = CreateAccessibilityInterface();
+  }
+  return accessibility_.get();
+}
+
 void GlCanvas::MouseMoved(int x, int y, bool left, bool /*right*/, bool /*middle*/) {
   int mouse_x = x;
   int mouse_y = y;
@@ -412,6 +419,7 @@ PickingMode GlCanvas::GetPickingMode() {
   return PickingMode::kNone;
 }
 
-void GlCanvas::CreateAccessibilityInterface() {
-  accessibility_ = std::make_unique<orbit_accessibility::AccessibleWidgetBridge>();
+std::unique_ptr<orbit_accessibility::AccessibleWidgetBridge>
+GlCanvas::CreateAccessibilityInterface() {
+  return std::make_unique<orbit_accessibility::AccessibleWidgetBridge>();
 }
--- a/OrbitGl/GlCanvas.cpp
+++ b/OrbitGl/GlCanvas.cpp
@@ -333,12 +333,11 @@ void GlCanvas::ScreenToWorld(int x, int y, float& wx, float& wy) const {
       world_top_left_y_ - (static_cast<float>(y) / static_cast<float>(GetHeight())) * world_height_;
 }
 
-Vec2 GlCanvas::ScreenToWorld(Vec2 screen_pos) {
+Vec2 GlCanvas::ScreenToWorld(Vec2 screen_pos) const {
   Vec2 world_pos;
-  world_pos[0] =
-      world_top_left_x_ + (screen_pos[0] / static_cast<float>(GetWidth())) * world_width_;
+  world_pos[0] = world_top_left_x_ + screen_pos[0] / static_cast<float>(GetWidth()) * world_width_;
   world_pos[1] =
-      world_top_left_y_ - (screen_pos[1] / static_cast<float>(GetHeight())) * world_height_;
+      world_top_left_y_ - screen_pos[1] / static_cast<float>(GetHeight()) * world_height_;
   return world_pos;
 }
 
@@ -348,6 +347,21 @@ float GlCanvas::ScreenToWorldHeight(int height) const {
 
 float GlCanvas::ScreenToWorldWidth(int width) const {
   return (static_cast<float>(width) / static_cast<float>(GetWidth())) * world_width_;
+}
+
+Vec2 GlCanvas::WorldToScreen(Vec2 world_pos) const {
+  Vec2 screen_pos;
+  screen_pos[0] = floorf((world_pos[0] - world_top_left_x_) / world_width_ * GetWidth());
+  screen_pos[1] = floorf((world_top_left_y_ - world_pos[1]) / world_height_ * GetHeight());
+  return screen_pos;
+}
+
+int GlCanvas::WorldToScreenHeight(float height) const {
+  return static_cast<int>(height / world_height_ * GetHeight());
+}
+
+int GlCanvas::WorldToScreenWidth(float width) const {
+  return static_cast<int>(width / world_width_ * GetWidth());
 }
 
 int GlCanvas::GetWidth() const { return screen_width_; }
--- a/OrbitGl/Track.h
+++ b/OrbitGl/Track.h
@@ -2,7 +2,8 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
 
-#pragma once
+#ifndef ORBIT_GL_TRACK_H_
+#define ORBIT_GL_TRACK_H_
 
 #include <atomic>
 #include <memory>
@@ -17,6 +18,7 @@
 #include "TextRenderer.h"
 #include "TimeGraphLayout.h"
 #include "TimerChain.h"
+#include "TrackAccessibility.h"
 #include "TriangleToggle.h"
 
 class GlCanvas;
@@ -85,12 +87,17 @@ class Track : public Pickable, public std::enable_shared_from_this<Track> {
   [[nodiscard]] const std::string& GetLabel() const { return label_; }
 
   void SetTimeGraph(TimeGraph* timegraph) { time_graph_ = timegraph; }
+  [[nodiscard]] TimeGraph* GetTimeGraph() { return time_graph_; }
+
   void SetPos(float a_X, float a_Y);
   void SetY(float y);
   [[nodiscard]] Vec2 GetPos() const { return pos_; }
   void SetSize(float a_SizeX, float a_SizeY);
+  [[nodiscard]] Vec2 GetSize() const { return size_; }
   void SetColor(Color a_Color) { color_ = a_Color; }
   [[nodiscard]] Color GetBackgroundColor() const;
+
+  [[nodiscard]] GlCanvas* GetCanvas() const { return canvas_; }
 
   void AddChild(std::shared_ptr<Track> track) { children_.emplace_back(track); }
   virtual void OnCollapseToggle(TriangleToggle::State state);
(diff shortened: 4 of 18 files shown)


OrbitGl/GlCanvas.cpp
```diff
@@ -138,6 +138,13 @@ void GlCanvas::EnableImGui() {
   }
 }
 
+orbit_accessibility::AccessibleInterface* GlCanvas::Accessibility() {
```
pierricgimmig, line 141: We tend to use the GetOrCreateXYZ pattern for lazily initialized objects, consider using the same pattern here, GetOrCreateAccessibilityInterface?

pierricgimmig: Should we use "*Accessible*Interface" rather than "*Accessibility*Interface"? It seems to me that the latter would be more appropriate.

OrbitGl/GlCanvas.cpp
```diff
@@ -138,6 +138,13 @@ void GlCanvas::EnableImGui() {
...
+  if (accessibility_ == nullptr) {
+    accessibility_ = CreateAccessibilityInterface();
```
pierricgimmig, line 143: Now that CreateAccessibilityInterface() is private, if it's not called elsewhere, should we remove it and inline the code here instead?

OrbitGl/CMakeLists.txt
```diff
@@ -152,7 +152,10 @@ target_sources(OrbitGlTests PRIVATE
                ScopedStatusTest.cpp
                SliderTest.cpp
                TimerInfosIteratorTest.cpp
-               UnitTestMockEnvironment.cpp)
+               # Todo (176066030): OrbitGl relies on flags usually defined in the OrbitQt executable, so the
+               # tests needs this to link against them. To avoid multiple definitions, we compile the flags
+               # cpp with the tests
+               ../OrbitQt/ClientFlags.cpp)
```
pierricgimmig, line 162: Ah good call to factor them out, thanks. Could they be moved to OrbitGl however so that OrbitGl wouldn't depend on OrbitQt?

OrbitGl/Track.h
```diff
@@ -99,12 +104,19 @@ class Track : public Pickable, public std::enable_shared_from_this<Track> {
   void SetProcessId(uint32_t pid) { process_id_ = pid; }
   [[nodiscard]] virtual bool IsEmpty() const = 0;
 
+  [[nodiscard]] virtual bool IsTrackSelected() const { return false; }
+
+  [[nodiscard]] bool IsCollapsed() { return collapse_toggle_->IsCollapsed(); }
```
pierricgimmig, line 109: const

OrbitGl/TrackAccessibility.cpp
```diff
@@ -0,0 +1,120 @@
...
+
+AccessibilityRect AccessibleTrackContent::AccessibleLocalRect() const {
+  CHECK(track_ != nullptr);
+  CHECK(track_->GetTimeGraph() != nullptr);
+
+  auto& layout = track_->GetTimeGraph()->GetLayout();
+  return AccessibilityRect(0, static_cast<int>(layout.GetTrackTabHeight()),
+                           static_cast<int>(track_->GetSize()[0]),
+                           static_cast<int>(track_->GetSize()[1]));
+}
```
pierricgimmig, line 32: I would prefer querying the Track for its position rather than going through the layout. It would be easier to support tracks that wouldn't have a tab for example. It seems to me that this logic should centralized in the tracks themselves. Also, shouldn't the rectangle include the tab?

OrbitGl/TrackAccessibility.cpp
```diff
@@ -0,0 +1,120 @@
...
+
+AccessibilityState AccessibleTrackContent::AccessibleState() const {
+  AccessibilityState result;
+  return result;
+}
```
pierricgimmig, line 37: Should this always return a dummy? Maybe use "*static* AccessibilityState result;" ?

OrbitGl/TrackAccessibility.cpp
```diff
@@ -0,0 +1,120 @@
...
+
+AccessibilityRect AccessibleTrackTab::AccessibleLocalRect() const {
+  CHECK(track_ != nullptr);
+  CHECK(track_->GetTimeGraph() != nullptr);
+
+  auto& layout = track_->GetTimeGraph()->GetLayout();
+  return AccessibilityRect(static_cast<int>(layout.GetTrackTabOffset()), 0,
+                           static_cast<int>(layout.GetTrackTabWidth()),
+                           static_cast<int>(layout.GetTrackTabHeight()));
```
pierricgimmig, line 55: Similarly to previous comment, I feel that we are duplicating the track layout logic in different classes. If a track diverges from the common pattern in the future, then we'll break this method and the accessible rectangle will be out of sync with the new track.

OrbitGl/TrackAccessibility.cpp
```diff
@@ -0,0 +1,120 @@
...
+}
+
+AccessibilityState AccessibleTrackTab::AccessibleState() const {
+  AccessibilityState result;
+  return result;
+}
```
pierricgimmig, line 61: "return AccessibilityState::Normal"?

OrbitGl/TrackAccessibility.cpp
```diff
@@ -0,0 +1,120 @@
...
+
+AccessibilityRect AccessibleTrack::AccessibleLocalRect() const {
+  CHECK(track_ != nullptr);
+  CHECK(track_->GetTimeGraph() != nullptr);
+  CHECK(track_->GetTimeGraph()->GetCanvas() != nullptr);
+
+  auto canvas = track_->GetTimeGraph()->GetCanvas();
+  auto pos = track_->GetPos();
+  auto size = track_->GetSize();
+  auto layout = track_->GetTimeGraph()->GetLayout();
```
pierricgimmig, line 81: please use full types instead of "auto"

OrbitGl/TrackAccessibility.cpp
```diff
@@ -0,0 +1,120 @@
...
+
+  int top = static_cast<int>(
+      std::max(-pos[1] + canvas->GetWorldTopLeftY() - layout.GetTrackTabHeight(), 0.0f));
+  int left = static_cast<int>(pos[0]);
+  int width = static_cast<int>(size[0]);
+  int height =
+      std::max(std::min(static_cast<int>(size[1] + layout.GetTrackTabHeight() - std::min(top, 0)),
+                        canvas->GetHeight() - top),
```
pierricgimmig, line 89: Here we are interpreting "world space" coordinates as "screen space" coordinates. It works because our world unit happen to align with the viewport screen space units. We don't really have the proper WorldToScreen conversion util in GlCanvas yet, but I created b/176129272. Maybe add a TODO here?

OrbitGl/TrackAccessibility.cpp
```diff
@@ -0,0 +1,120 @@
...
+  return AccessibilityRect(left, top, width, height);
+}
+
+AccessibilityState AccessibleTrack::AccessibleState() const {
+  CHECK(track_ != nullptr);
+
+  using State = orbit_accessibility::AccessibilityState;
+
+  State result = State::Normal;
+  result |= State::Focusable | State::Movable;
```
pierricgimmig, line 101: Can this be on a single line?

OrbitGl/AccessibleTimeGraph.cpp
```diff
@@ -0,0 +1,30 @@
...
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#include "AccessibleTimeGraph.h"
+
+#include "GlCanvas.h"
+#include "TimeGraph.h"
+
+AccessibilityRect TimeGraphAccessibility::AccessibleLocalRect() const {
+  auto canvas = time_graph_->GetCanvas();
```
pierricgimmig, line 11: Please use full type instead of auto

OrbitGl/AccessibleTimeGraph.cpp
```diff
@@ -0,0 +1,30 @@
...
+  return AccessibilityRect(0, 0, canvas->GetWidth(), canvas->GetHeight());
+}
+
+AccessibilityState TimeGraphAccessibility::AccessibleState() const {
+  using State = orbit_gl::AccessibilityState;
+  return State::Focusable;
```
pierricgimmig, line 17: nit. is the indirection needed? "return orbit_gl::AccessibilityState::Focusable"?

OrbitGl/TrackAccessibility.cpp
pierricgimmig: It will also fix compilation I think.

OrbitGl/GlCanvas.cpp
reichlfl: Good point about GetOrCreate, I'll change that. Regarding "Accessibility" vs "Accessible": I've named the interfaces "Accessible" to have them named similar to the Qt convention, and following the idea of "this is the interface for an accessible object". Also because "Accessible" is slightly less annoying to type than "Accessibility". I've named the module "OrbitAccessibility" though because it contains all the general concepts for Accessibility.

That being said - the method should probably be called `GetOrCreateAccessibleInterface`.

OrbitGl/GlCanvas.cpp
reichlfl: It's overridden in `CaptureWindow`. That's the whole idea of this method - it's a template method that deriving classes can override to construct specialized instances of the accessibility interface.

OrbitGl/CMakeLists.txt
reichlfl: I was thinking the same, but then again, some of the flags are defined in OrbitQt, so it again would be rather inconsistent to define them in OrbitGl, wouldn't it?

But probably from a dependency perspective it's still more consistent, I'll change it.

reichlfl: Changed

OrbitGl/TrackAccessibility.cpp
reichlfl: True, I changed the way `AccessibilityState` is defined during this PR, this made sense before :-) Changed, thanks.

OrbitGl/TrackAccessibility.cpp
reichlfl: I've actually added the methods to make the calculations in `TrackAccessibility` more straightforward and remove some of the required knowledge of GlCanvas internals from the classes here.

OrbitGl/AccessibleTimeGraph.cpp
reichlfl: No, it's not, thanks. The result was composed of more than one state in the past, now this doesn't make sense any more.

OrbitGl/TrackAccessibility.cpp
reichlfl: Overall I agree, but I'd rather do that in a subsequent PR that tackles the problem a bit more generic. The problem here is that I'm splitting the Track into 3 UI components (container, tab, and content), but the Track itself doesn't know about this. Making this work without layouting information would require adding a bunch of methods to the track (e.g. `GetTabPos`, `GetTabSize`, `GetContentPos`, `GetContentSize`).

On top of that, the Tracks are IMO not really consistent in what they consider "position" and "size", because both of them do not take the Tabs into account, so I'd either need to change how `Track::GetSize` and `Track::GetPos` work, or add additional methods.

A better solution would IMO be introducing a general interface for UI-Controls that exposes pos / size and parent / child relationsships (maybe tied to the `Pickable` interface?), then the Tracks could actually create sub-controls for the Tabs and the Content and own the layout.

OrbitGl/GlCanvas.cpp
pierricgimmig: Sounds good!

OrbitGl/GlCanvas.cpp
pierricgimmig: Ah yes, cool.

OrbitGl/CMakeLists.txt
pierricgimmig: Thanks!

karupayun (commented):
Hi Florian! I've left a few comments, feel free to ping me about them tomorrow :)

OrbitGl/TrackAccessibility.cpp
pierricgimmig: Ok sounds good, thanks for the precision.

OrbitGl/GlCanvas.cpp
```diff
@@ -343,6 +349,21 @@ float GlCanvas::ScreenToWorldWidth(int width) const {
   return (static_cast<float>(width) / static_cast<float>(GetWidth())) * world_width_;
 }
 
+Vec2 GlCanvas::WorldToScreen(Vec2 world_pos) const {
+  Vec2 screen_pos;
+  screen_pos[0] = (world_pos[0] - world_top_left_x_) / world_width_ * GetWidth();
+  screen_pos[1] = (world_top_left_y_ - world_pos[1]) / world_height_ * GetHeight();
+  return screen_pos;
+}
```
pierricgimmig, line 357: I think we should use floorf(...), to be consistent with how TextRenderer positions glyphs. Also this will make it easy to spot screen space coordinates as their decimal part will always be 0.f. Wdyt? I was tempted to suggest returning integers for screen space, as GetWidth() and GetHeight() do, but it's true that we probably always convert to float. If there is a need for an "int" version, we should probably add it as a separate function.

OrbitGl/GlCanvas.cpp
```diff
@@ -343,6 +349,21 @@ float GlCanvas::ScreenToWorldWidth(int width) const {
...
+
+int GlCanvas::WorldToScreenHeight(float height) const {
+  return static_cast<int>(height / world_height_ * GetHeight());
+}
+
+int GlCanvas::WorldToScreenWidth(float width) const {
+  return static_cast<int>(width / world_width_ * GetWidth());
+}
```
pierricgimmig, line 365: There is an inconsistency with the WorldToScreen return type, we have both float and int for screen space coordinates. I think we should make that more uniform eventually, perhaps by using float across the board, but it's outside the scope of this PR.

pierricgimmig (approved):
Hi @reichlfl , thanks for the changes, please see minor comment about using floorf(..) but otherwise LGTM.

OrbitGl/GlCanvas.cpp
reichlfl: I think it depends - in case of the tracks, the argument was rather that Tracks shouldn't just be created "on the fly", there usually should be one entity in charge of creating a track when it's needed. 

In this case here, the `accessibility_` could already be created in the constructor, it usually should just always exist. But since I'm using a virtual factory method that may be overridden in the inheriting class, I can't call it in the base constructor.  I could have added an `Init()` method that is called after construction, but I'm not a fan of additional initialization-calls because they violate RAII. In addition, there already is an `Initialize()` method in GlCanvas, but this may be called later because it's called on OpenGl initialization... so I decided that constructing the instance on demand is the cleanest way here.

OrbitGl/TrackAccessibility.cpp
reichlfl: > I'm ok if you change that and I can easily modify TrackManager (I think that is 1-line modification) to draw everything properly.

I'm shying away from this change a bit since I'm not sure which parts directly or indirectly rely on this convention atm. If it's ok with you, I would suggest to merge it as-is, I need to take care of rebasing on your changes anyways, and we could then tackle this afterwards.

OrbitGl/GlCanvas.cpp
```diff
@@ -1,12 +1,12 @@
+#include "GlCanvas.h"
```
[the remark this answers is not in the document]
reichlfl: Huh, interesting. No clue where that comes from, thanks for spotting, I'll remove it.

OrbitGl/GlCanvas.cpp
reichlfl: Good point, I've added that. Regarding returning `int`, and also related to your comment below: I agree, and initially, I wanted to make this return an int-vector. But I've decided against it for consistency since we're using Vec2 for screen-space in a lot of other places as well.

But I agree that it would be nice to change this all to int at some point.

OrbitGl/TrackAccessibility.cpp
```diff
@@ -0,0 +1,120 @@
...
+std::string AccessibleTrackContent::AccessibleName() const {
+  CHECK(track_ != nullptr);
+  return track_->GetName() + "_content";
+}
+
+AccessibilityRect AccessibleTrackContent::AccessibleLocalRect() const {
+  CHECK(track_ != nullptr);
+  CHECK(track_->GetTimeGraph() != nullptr);
+
+  auto& layout = track_->GetTimeGraph()->GetLayout();
```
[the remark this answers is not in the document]
reichlfl: As discussed offline, I'm now passing the layout in the constructors and got rid of all but one dependencies to `TimeGraph`. This last dependency will be solved with a `GetParent()` later.

OrbitGl/TrackAccessibility.cpp
```diff
@@ -0,0 +1,120 @@
...
+  return AccessibilityRect(static_cast<int>(layout.GetTrackTabOffset()), 0,
+                           static_cast<int>(layout.GetTrackTabWidth()),
+                           static_cast<int>(layout.GetTrackTabHeight()));
+}
+
+AccessibilityState AccessibleTrackTab::AccessibleState() const {
+  AccessibilityState result;
+  return result;
```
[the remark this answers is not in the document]
reichlfl: Yes, that was fixed in a later commit.

OrbitGl/TrackAccessibility.cpp
```diff
@@ -0,0 +1,120 @@
...
+
+std::string AccessibleTrack::AccessibleName() const {
+  CHECK(track_ != nullptr);
+  return track_->GetName();
+}
+
+AccessibilityRect AccessibleTrack::AccessibleLocalRect() const {
+  CHECK(track_ != nullptr);
+  CHECK(track_->GetTimeGraph() != nullptr);
+  CHECK(track_->GetTimeGraph()->GetCanvas() != nullptr);
```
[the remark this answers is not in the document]
reichlfl: I am now instead getting the canvas from the track - this makes more sense in the current state as the tracks depend on properties of the canvas (width, scrolling...), and it can then be removed straightforward later if we either decide to remove this dependency to the canvas, or make `GetCanvas()` part of the generic ui element interface.

OrbitGl/TrackAccessibility.cpp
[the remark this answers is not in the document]
reichlfl: Also added a bug for this: b/176216022

OrbitGl/TrackAccessibility.cpp
```diff
@@ -75,30 +79,38 @@ AccessibilityRect AccessibleTrack::AccessibleLocalRect() const {
...
+  int screen_width = canvas->WorldToScreenWidth(size[0]);
+  int screen_height = canvas->WorldToScreenHeight(size[1]);
+
+  // Adjust the coordinates to clamp the result to an on-screen rect
+  // This will "cut" any part that is offscreen due to scrolling, and may result
+  // in a final result with width / height of 0.
+  if (screen_pos[1] + screen_height > canvas->GetHeight()) {
+    screen_height = std::max(0, canvas->GetHeight() - static_cast<int>(screen_pos[1]));
+  }
+  if (screen_pos[1] < 0) {
```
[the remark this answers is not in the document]
reichlfl: Added an additional comment to make this clearer

karupayun (approved):
LGTM!

Thanks for addressing my comments :)

PR #1612, opened 2020-12-23
Add model_ == nullptr conditions in orbittreeview

reichlfl:
Fixes b/176151873.
Reason: model_ can now be nullptr under certain conditions,
this needs to be dealt with.

Added in all methods that could potentially fire during
de-initialization due to signals or user interaction.

Removed unneeded drawRow() method to clean up the class.

### DIFF
--- a/OrbitQt/orbittreeview.cpp
+++ b/OrbitQt/orbittreeview.cpp
@@ -103,23 +103,35 @@ void OrbitTreeView::ClearDataModel() {
 }
 
 void OrbitTreeView::OnSort(int section, Qt::SortOrder order) {
+  if (model_ == nullptr) {
+    return;
+  }
+
   model_->sort(section, order);
   Refresh();
 }
 
 void OrbitTreeView::OnFilter(const QString& filter) {
+  if (model_ == nullptr) {
+    return;
+  }
+
   model_->OnFilter(filter);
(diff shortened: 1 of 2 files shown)


karupayun (approved):
Hi Florian! Thanks for taking care of this bug. Now seems to be solved. BTW, I have a question.

I don't know too much of the code here, but in which cases model_ could be a nullptr?? Doesn't have more sense assure model_ always have a valid pointer instead of doing nothing if there is a nullptr?

reichlfl:
> Hi Florian! Thanks for taking care of this bug. Now seems to be solved. BTW, I have a question.
> 
> I don't know too much of the code here, but in which cases model_ could be a nullptr?? Doesn't have more sense assure model_ always have a valid pointer instead of doing nothing if there is a nullptr?

This can happen in `OrbitTreeView::Deinitialize()` and `OrbitTreeView::ClearDataModel()`. Usually should only happen on destruction, but this causes the selection to be updated before destruction is complete, and some of the events to fire.

Since `ClearDataModel` is part of the public API for `OrbitTreeView`, we should consider the case `model_ == nullptr` to be valid.

beckerhe:
Thanks a lot Florian for fixing this. I must have missed that!

PR #1614, opened 2020-12-23
E2E Framework fixes

reichlfl:
Several fixes / updates to the E2E tests before merging the capture window tests.

- `find_control` did not always work as expected. For some control types, the AccessibleName was not returned as expected, and it looks like pywinauto is not exposing this in all cases. I've changed it to use the underlying COM functionality now, this seems to work a lot more reliable
- `find_control` can be instructed to not recurse
- Renamed `E2ETest` and `Fragment` to `E2ETestSuite` and `E2ETestCase` as this is clearer and more common

### DIFF
--- a/contrib/automation_tests/core/orbit_e2e.py
+++ b/contrib/automation_tests/core/orbit_e2e.py
@@ -6,6 +6,7 @@ found in the LICENSE file.
 
 import logging
 from typing import Iterable, Type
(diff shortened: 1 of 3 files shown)


pierricgimmig (approved):
Hi Florian, I have very limited experience with e2e tests, but from what I can tell, this looks good to me.

PR #1615, opened 2020-12-23
Clean TimeGraph dependences from TrackManager/Track

karupayun:
In the process of making UnitTests for the Tracks, we want to erase
most dependences to TimeGraph to easily test TrackManager. Also we want
to get rid of the circular dependence from TimeGraph and the Tracks.

In this PR we:

 - Use correctly TimeGraph static methods.

 - Get CaptureData from App instead of TimeGraph (CaptureData now isn't
 on TimeGraph anymore).

 - Get StringManager from App instead of passing throught the classes.
 It is only used in GpuTrack.

 - Move initialization (Name, Label, ...) inside the Tracks and no inside
 TrackManager. Now GetOrCreate methods are really similar and could be
 templatized in the future or we can have an only list of tracks and
 easily get from there.

Bugs related:
https://b/176086740
https://b/176056427

Tested with the new_ui: Start a new capture/Load a capture/End session
and load it again.

### DIFF
--- a/OrbitGl/AsyncTrack.cpp
+++ b/OrbitGl/AsyncTrack.cpp
@@ -16,8 +16,8 @@ using orbit_client_protos::TimerInfo;
 
 AsyncTrack::AsyncTrack(TimeGraph* time_graph, const std::string& name, OrbitApp* app)
     : TimerTrack(time_graph, app) {
-  SetName(name);
-  SetLabel(name);
+  name_ = name;
+  label_ = name;
 }
 
 [[nodiscard]] std::string AsyncTrack::GetBoxTooltip(PickingId id) const {
@@ -111,7 +111,7 @@ Color AsyncTrack::GetTimerColor(const TimerInfo& timer_info, bool is_selected) c
   orbit_api::Event event = ManualInstrumentationManager::ApiEventFromTimerInfo(timer_info);
   const uint64_t event_id = event.data;
   std::string name = app_->GetManualInstrumentationManager()->GetString(event_id);
-  Color color = time_graph_->GetColor(name);
+  Color color = TimeGraph::GetColor(name);
 
   constexpr uint8_t kOddAlpha = 210;
   if (!(timer_info.depth() & 0x1)) {
--- a/OrbitGl/EventTrack.cpp
+++ b/OrbitGl/EventTrack.cpp
@@ -76,8 +76,7 @@ void EventTrack::UpdatePrimitives(uint64_t min_tick, uint64_t max_tick, PickingM
 
   const Color kWhite(255, 255, 255, 255);
   const Color kGreenSelection(0, 255, 0, 255);
-  const CaptureData* capture_data = time_graph_->GetCaptureData();
-  CHECK(capture_data != nullptr);
+  const CaptureData& capture_data = app_->GetCaptureData();
 
   if (!picking) {
     // Sampling Events
@@ -90,10 +89,10 @@ void EventTrack::UpdatePrimitives(uint64_t min_tick, uint64_t max_tick, PickingM
     };
 
     if (thread_id_ == orbit_base::kAllProcessThreadsTid) {
-      capture_data->GetCallstackData()->ForEachCallstackEvent(action_on_callstack_events);
+      capture_data.GetCallstackData()->ForEachCallstackEvent(action_on_callstack_events);
     } else {
-      capture_data->GetCallstackData()->ForEachCallstackEventOfTid(thread_id_,
-                                                                   action_on_callstack_events);
+      capture_data.GetCallstackData()->ForEachCallstackEventOfTid(thread_id_,
+                                                                  action_on_callstack_events);
     }
 
     // Draw selected events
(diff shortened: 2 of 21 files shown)


pierricgimmig:
Hi Pablo, I haven't done a full review, but 2 things stand out:

Moving the capture data to the App level means that we can't have different timegraphs loaded at once. Concretely, this will duplicate the capture data in the introspection windows since introspection has its own timegraph that will access the "global" capture data from the App.

We should remove the "SetStringManager" method and its last remaining call site (b/OrbitGl/CaptureDeserializerLoadFuzzer.cpp)

OrbitGl/AsyncTrack.cpp
```diff
@@ -16,8 +16,8 @@ using orbit_client_protos::TimerInfo;
 
 AsyncTrack::AsyncTrack(TimeGraph* time_graph, const std::string& name, OrbitApp* app)
     : TimerTrack(time_graph, app) {
-  SetName(name);
-  SetLabel(name);
+  name_ = name;
+  label_ = name;
```
beckerhe, line 20: Personal preference: I would prefer to initialize these in the member initializer list, something like:
`: TimeTrack(time_graph, app), name_{name}, label_{name} { }`

beckerhe: But that's just personal preference and since it has been like so in the original code, I'm also fine if you just leave it.

OrbitGl/ThreadTrack.cpp
```diff
@@ -18,15 +18,41 @@ using orbit_client_protos::FunctionInfo;
 using orbit_client_protos::TimerInfo;
 
 ThreadTrack::ThreadTrack(TimeGraph* time_graph, int32_t thread_id, OrbitApp* app)
-    : TimerTrack(time_graph, app) {
+    : TimerTrack(time_graph, app), app_{app} {
```
beckerhe, line 21: I'm not sure about this change. There is a protected member variable `TimerTrack::app_` which is now shadowed by `ThreadTrack::app_`. Do you really need this additional variable?

BTW, of course accessing a protected variable in a base class is very bad design, but shadowing a base class member variable is also not ideal.

OrbitGl/TrackManager.cpp
```diff
@@ -83,7 +81,7 @@ void TrackManager::SortTracks() {
 
   ThreadTrack* process_track = nullptr;
 
-  const CaptureData* capture_data = time_graph_->GetCaptureData();
+  const CaptureData* capture_data = app_->HasCaptureData() ? &app_->GetCaptureData() : nullptr;
   if (capture_data != nullptr) {
```
beckerhe, line 85: I think it would be nicer to do it like so:
```
if (app_->HasCaptureData()) {
  const auto& capture_data = app_->GetCaptureData();
  ...
```

beckerhe (commented):
Hi Pablo,
all your changes look good. Thank you for that.

But I believe there are too many unrelated things going on in a single commit.

I think you could have single commits for each of these changes:
- Removing StringManager from TimeGraph
- Properly calling TimeGraph's static methods
- Replacing TimeGraph::GetCaptureData by OrbitApp::GetCaptureData
- Refactoring name_ and label_ in all the tracks
- Moving num_cores_ from TimeGraph to SchedulerTrack

OrbitGl/EventTrack.cpp
```diff
@@ -76,8 +76,7 @@ void EventTrack::UpdatePrimitives(uint64_t min_tick, uint64_t max_tick, PickingM
 
   const Color kWhite(255, 255, 255, 255);
   const Color kGreenSelection(0, 255, 0, 255);
-  const CaptureData* capture_data = time_graph_->GetCaptureData();
-  CHECK(capture_data != nullptr);
+  const CaptureData& capture_data = app_->GetCaptureData();
```
florian-kuebler, line 79: >Moving the capture data to the App level means that we can't have different timegraphs loaded at once. Concretely, this will duplicate the capture data in the introspection windows since introspection has its own timegraph that will access the "global" capture data from the App.

Okay, so if this is the case and we can't use app's capture data here, we should really pass the capture data all the way down to the tracks that use it. So EventTrack could have a pointer to the capture data to be used. 

WDYT?

florian-kuebler (commented):
Thanks for the changes. From my side it looks good, but as Pierric pointed out, there might be an issue when having two capture data objects in the case of introspection. I guess in those cases the Tracks need to know their capture data.

OrbitGl/ThreadTrack.cpp
beckerhe: Yes. It will work just fine - as long as ThreadTrack::app_ and TimerTrack::app_ hold the same value - which is the case here. But it's suprising to us developers. There is a high risk of introducing subtle bugs in the future.

florian-kuebler: Wait, I wouldn't shadow/overload the `app_` in ThreadTrack. From a semantical point of view, I would expect that the ` app_` in TimerTrack is the one, that this track belongs to, and that should be the same for all subclasses.

In general I see the potential issues coming with protected members, but that feels like the use-case for which protected was introduced.

beckerhe: I can't completely follow. I think we might mean the same.

florian-kuebler: I was referring to this part:
>BTW, of course accessing a protected variable in a base class is very bad design, but shadowing a base class member variable is also not ideal.

pierricgimmig (changes requested):
As discussed, the introspection window uses a TimeGraph object to display its information. We shouldn't access the App's CaptureData directly otherwise we'll end up duplicating the main capture in the introspection window.

OrbitGl/EventTrack.cpp
karupayun: As discussed online, I agree

PR #1616, opened 2020-12-23
Some E2E Tests for tracks in the capture window

reichlfl:
**Based on https://github.com/google/orbit/pull/1614**

Ignore the first 2 commits if you want to review this, they are also contained in the linked PR.

This add a couple of E2E tests for tracks inside the capture window:

- Verify selection and deselection of tracks by clicking in the capture
window
- Verify drag / drop behavior for track sorting
- Verify filtering of tracks correctly affects track in the capture
window

I have to admit that it is not super structured in terms of what is tested, and it's not an exhaustive test for the track functionality. This mainly served the purpose of verifying the accessibility of the capture window, and could be extended later.


reichlfl:
This is now contained in https://github.com/google/orbit/pull/1623, sorry for the mess - I ran into several peculiarities with pywinauto while updating the rest of the E2E tests, and this lead to changes in all tests that were hard to pull apart without having reviewers available in the same timezone.

PR #1618, opened 2020-12-29
Better ImGui layout in "Debug" tab

pierricgimmig:
Add a specific "Introspection" section in the Debug tab (visible when 
launching with --devmode). Also use ImGui "CollapsingHeader" to
organize information.

### DIFF
--- a/OrbitGl/IntrospectionWindow.cpp
+++ b/OrbitGl/IntrospectionWindow.cpp
@@ -62,7 +62,6 @@ void IntrospectionWindow::RenderImGui() {
 
   if (ImGui::CollapsingHeader("IntrospectionWindow")) {
     IMGUI_VAR_TO_TEXT(IsIntrospecting());
-    IMGUI_VAR_TO_TEXT(time_graph_.GetTrackManager()->GetAllTracks().size());
   }
 }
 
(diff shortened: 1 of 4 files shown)


OrbitGl/IntrospectionWindow.cpp
```diff
@@ -57,4 +57,13 @@ void IntrospectionWindow::ToggleRecording() {
   }
 }
 
+void IntrospectionWindow::RenderImGui() {
+  CaptureWindow::RenderImGui();
+
+  if (ImGui::CollapsingHeader("IntrospectionWindow")) {
+    IMGUI_VAR_TO_TEXT(IsIntrospecting());
+    IMGUI_VAR_TO_TEXT(time_graph_.GetTrackManager()->GetAllTracks().size());
```
karupayun, line 65: I would like to also have this Variable for CaptureInfo, WDYT?

karupayun (approved):
LGTM!! :)

OrbitGl/IntrospectionWindow.cpp
pierricgimmig: Yes good point, thanks, moved it over to CaptureWindow.

PR #1619, opened 2020-12-30
Fix assert when moving tracks

pierricgimmig:
Use visible_tracks_ iterator before it is invalidated by insertion.

### DIFF
--- a/OrbitGl/TrackManager.cpp
+++ b/OrbitGl/TrackManager.cpp
@@ -252,8 +252,8 @@ void TrackManager::UpdateMovingTrackSorting() {
     int moving_track_current_position = -1;
     for (auto track_it = visible_tracks_.begin(); track_it != visible_tracks_.end(); ++track_it) {
       if (moving_track->GetPos()[1] >= (*track_it)->GetPos()[1]) {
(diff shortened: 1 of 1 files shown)


reichlfl (approved):
LGTM, thanks!

OrbitGl/TrackManager.cpp
```diff
@@ -252,8 +252,8 @@ void TrackManager::UpdateMovingTrackSorting() {
     int moving_track_current_position = -1;
     for (auto track_it = visible_tracks_.begin(); track_it != visible_tracks_.end(); ++track_it) {
       if (moving_track->GetPos()[1] >= (*track_it)->GetPos()[1]) {
-        visible_tracks_.insert(track_it, moving_track);
```
reichlfl, line 255: I already LGTM'd this, but afterwards, I was just wondering: Wouldn't it be clearer to make this

```
auto inserted_it = visible_tracks_.insert(track_it, moving_track);
moving_track_current_position = inserted_it - visible_tracks_.begin();
```

pierricgimmig: Yep lgtm, I made the change, thanks!

karupayun:
Thanks again!

PR #1620, opened 2020-12-30
Fix track sorting

pierricgimmig:
The TrackManager change introduced a bug that causes tracks that don't
have callstack events to never show up. It also broke the previous
sorting policy which was to display tracks with instrumented functions
first. Reimplement track sorting and remove unused and duplicated
"thread_count_map_" and "event_count_" vectors.

This change also fixes introspection tracks not showing up anymore.

### DIFF
--- a/OrbitGl/TimeGraph.cpp
+++ b/OrbitGl/TimeGraph.cpp
@@ -72,7 +72,6 @@ void TimeGraph::Clear() {
   batcher_.StartNewFrame();
   capture_min_timestamp_ = std::numeric_limits<uint64_t>::max();
   capture_max_timestamp_ = 0;
-  thread_count_map_.clear();
 
   track_manager_->Clear();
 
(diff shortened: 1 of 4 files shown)


reichlfl (approved):
Thanks for fixing this, Pierric, LGTM, this also makes the code cleaner!

Could you just verify that I've read this correctly:
1) As far as I understand, the issue was that `GetSortedThreadIds` runs through the `thread_count_map_` instead of all tracks, and Threads without events do not show up there, is that the cause?
2) With this change, `GetCallstackEventsOfTidCount` is called during sorting, but this method should run in O(1) because it returns the size of a vector retrieved from a hashmap, so it should not result in decreased performance compared to keeping the running count in `thread_count_map_`.

pierricgimmig:
Hi @reichlfl , thanks for the review. To answer your questions:

> 1. As far as I understand, the issue was that `GetSortedThreadIds` runs through the `thread_count_map_` instead of all tracks, and Threads without events do not show up there, is that the cause?

The issue was that thread_count_map_ was copied over from TimeGraph to TrackManager but it is not updated, causing tracks with timers but no events to not be displayed. The map is not really needed as we can access that count from the track themselves.

> 2\. With this change, `GetCallstackEventsOfTidCount` is called during sorting, but this method should run in O(1) because it returns the size of a vector retrieved from a hashmap, so it should not result in decreased performance compared to keeping the running count in `thread_count_map_`.

Yes that's exactly right. Also, sorting happens once every second when capturing. For the number of tracks we are dealing with, that's really not a bottleneck.

karupayun:
Thanks for addressing this Pierric! :)

PR #1621, opened 2020-12-30
Fixed a newly introduced bug that slipped through

reichlfl:
Realized that this fix was missing after the last rebase, sorry for that!


pierricgimmig (approved):
LGTM!

PR #1623, opened 2020-12-30
E2E Updates for the new UI

reichlfl:
This is a rather large PR that first fixes several issues with the E2E framework,
introduces a new E2E test for sortable tracks, and updates the existing E2E
tests to work with both the current and the new beta UI.

I've split up everything into logical commits that can be reviewed one by one.
The first three commits contain preperations to support the new UI from C++
and Python side, all other commits introduce new or update existing E2E tests
one by one. Some of them introduce functionality that the later ones rely on,
so they need to be applied in this order and cannot be split into orthogonal PRs.

I've tagged a bunch of reviewers - @dpallotti and @ronaldfw, it's fine if you only
take a look at the changes related to your E2E tests. I made some small updates
to verify additional conditions, but kept all the functionality otherwise. Some
of the magic lookups provided by PyWinAuto have been replaced with our own
helper functions because this provides a significant speedup and makes the tests
more robust to layout changes.


beckerhe (approved):
That looks really great. Thank you Florian! No objections from my side. LGTM!

OrbitQt/AccessibilityAdapter.cpp
```diff
@@ -110,7 +110,20 @@ absl::flat_hash_map<const AccessibleInterface*, QAccessibleInterface*>
...
     AccessibilityAdapter::managed_adapters_;
 
+/*
+ * Utility class to verify all created accessibility adapters have been cleaned up
+ * when the application exits as a sanity check for correct registration behavior.
+ *
+ * A static instance of this is created on AccessibilityAdapter::Init() and will thus
+ * be destructed at exit. At this point, all adapters should have been cleaned up.
+ */
+class TeardownCheck {
```
dpallotti, line 120: Should this be in the anonymous namespace?

contrib/automation_tests/fragments/top_down_tab.py
```diff
@@ -0,0 +1,68 @@
...
+import logging
+
+from core.orbit_e2e import E2ETestCase, find_control
+
+
+class VerifyHelloGgpTopDownContents(E2ETestCase):
+    """
+        TODO: This is a more or less plain copy of the old E2E test. If may need to
+        be updated
+        """
```
dpallotti, line 15: Specify how it should be updated, otherwise if it's unclear the TODO should be removed.

contrib/automation_tests/fragments/bottom_up_tab.py
```diff
@@ -0,0 +1,48 @@
...
+import logging
+
+from core.orbit_e2e import E2ETestCase, find_control
+
+
+class VerifyHelloGgpBottomUpContents(E2ETestCase):
+    """
+    TODO: This is a more or less plain copy of the old E2E test. If may need to
+    be updated
+    """
```
dpallotti, line 15: Same here: how does it need to be updated?

dpallotti (approved):
I only had a (relatively) brief look.

contrib/automation_tests/core/orbit_e2e.py
```diff
@@ -110,9 +133,42 @@ def execute(self):
...
-                 auto_id_leaf=None, qt_class=None, recurse=True) -> BaseWrapper:
+
+def wait_for_condition(callable: Callable, max_seconds: int = 5, interval: int = 1):
+    """
+    Wait until a condition is satisfied.
+
+    :param callable: Any callable that evaluates to True or False. If this returns True, the condition is considered
+        satisfied.
+    :param max_seconds: Maximum time to wait for the condition to be satisfied
+    :param interval: Sleep time in between calls to callable
```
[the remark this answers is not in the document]
reichlfl: Yes, I'll add this to the param comments, thanks for pointing it out.

OrbitQt/AccessibilityAdapter.cpp
reichlfl: Yes, makes sense, thanks! Changed.

contrib/automation_tests/fragments/top_down_tab.py
reichlfl: Now that you mention it, after I realized that the test wasn't working for the new UI, I already updated it slightly. I've changed the comment, it could still be nice to change the test to not contain e.g. the hard-coded number of cells, there's some tooling for this now in common_controls that could be used.

contrib/automation_tests/fragments/bottom_up_tab.py
reichlfl: Same as above

contrib/automation_tests/fragments/symbols_tab.py
```diff
@@ -72,3 +75,58 @@ def _execute(self, function_search_string):
...
+    def _execute(self):
+        _show_symbols_and_functions_tabs(self.suite.top_window())
+
+        self._load_presets()
+        wait_for_condition(self._try_verify_functions_are_hooked)
+
+        Capture().execute(self.suite)
+
+        logging.info('Verifying function call counts')
+        VerifyFunctionCallCount(function_name='DrawFrame', min_calls=30, max_calls=3000).execute(self.suite)
```
danielfenner, line 94: I would have expected "Capture" and "VerifyFunctionCallCount" in the "tests = [..]" list.

No need to change this now if this complicated for some reason I don't see...

contrib/automation_tests/core/orbit_e2e.py
```diff
@@ -51,52 +52,74 @@ def expect_true(self, cond, description):
...
+
 class E2ETestSuite:
-    def __init__(self, test_name: str, tests: Iterable[E2ETestCase], dev_mode: bool = False,
-                 auto_connect: bool = True):
-        try:
-            self._application = Application(backend='uia').connect(title_re='orbitprofiler')
-        except Exception:
-            logging.error("Could not find Orbit application. Make sure to start Orbit before running E2E tests.")
-            raise
+    def __init__(self, test_name: str, tests: Iterable[E2ETestCase], dev_mode: bool = False):
```
danielfenner, line 96: I had troubles getting the concepts of TestSuite, TestCase right at first. I'd suggest calling the param 'test_cases' (here and in files were the suites are implemented)

`def __init__(self, test_name: str, test_cases: Iterable[E2ETestCase], dev_mode: bool = False):`

Also I would rename the folder "fragments" to "test_cases" (since fragments are not a thing anymore, right?)

Even if this is something you agree on it is probably best to do this in a follow up PR.

danielfenner (commented):
Thanks for looking into this! Great to have a way to operate on the capture window!

contrib/automation_tests/fragments/connection_window.py
```diff
@@ -0,0 +1,81 @@
...
+        window.InstanceList.DataItem0.wait('exists', timeout=100)
+        instance_list = self.find_control('Table', 'InstanceList')
+        logging.info('Found %s rows in the instance list', instance_list.item_count())
+        self.expect_true(instance_list.item_count() >= 1, 'Found at least one instance')
+
+        window.InstanceList.DataItem0.double_click_input()
+        logging.info('Connecting to Instance, waiting for the process list...')
+
+        if flags.FLAGS.enable_ui_beta:
+            # In the new UI, use small waits until the process list is active, and then some more for the Overlay
```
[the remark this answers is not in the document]
reichlfl: It's part of the tables in the new connection window, the semi-transparent black overlay indicting progress while connecting. This is not ideal, but right now there is no way for the automation to know when loading of the table is done because the overlay is not correctly exposed through the automation API (for reasons unknown).

But I think this should be looked into on C++ side and we could then get rid of this sleep. I'll create a bug for this and mention it here.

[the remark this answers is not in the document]
reichlfl: Created b/177037834 and added a TODO referencing the bug in the Python code

contrib/automation_tests/fragments/connection_window.py
```diff
@@ -0,0 +1,81 @@
...
+            process_list = window.DataViewProcesses.DataView
+
+        if process_filter:
+            filter_edit.set_edit_text(process_filter)
+        self.expect_true(process_list.item_count() > 0, 'Process list has at least one entry')
+
+        if flags.FLAGS.enable_ui_beta:
+            logging.info('Process selected, continuing to main window...')
+            process_list.DataItem0.double_click_input()
+            time.sleep(2)
```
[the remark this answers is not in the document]
reichlfl: Actually, now that you mention it, yes... I've changed this and the unconditional 15-second-sleep to wait for a window with class name "OrbitMainWindow" to be on top, this is a lot nicer. Thanks for pointing this out, PTAL.

contrib/automation_tests/fragments/connection_window.py
```diff
@@ -0,0 +1,81 @@
...
+            window = self.suite.top_window(True)
+            self.expect_eq(window.class_name(), "OrbitMainWindow", 'Main window is visible')
+        logging.info('Process list ready')
+
+
+class FilterAndSelectFirstProcess(E2ETestCase):
+    """
+    Select the first process in the process list and verify there is at least one entry in the list
+    """
+    def _execute(self, process_filter: str = "hello_ggp"):
```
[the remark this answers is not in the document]
reichlfl: Agreed, done

contrib/automation_tests/core/orbit_e2e.py
reichlfl: True, I was thinking about the same and then worried to make this too hard to review with those changes. I'll do that in a follow-up PR.

contrib/automation_tests/fragments/symbols_tab.py
reichlfl: You're pointing at something that I'm not really happy with here, and I'm not sure how I want to handle this TBH. The reasoning for this is the following: I want the test case `LoadAndVerifyHelloGgpPreset` to verify the loading of the preset self-contained, and the test suite should not need to care that it needs to take a capture and count function calls to verify the loading of the preset.

I can either do this by:
- Duplicating code from `Capture` and `VerifyFunctionCallCount`, which is obviously not ideal
- Moving the functionality to capture and verify the call count into a free function in the respective modules and calling those in all test cases. This means I can't use some of the convenience that the `E2ETestCase` class provides inside these functions, and I somewhat expose internals of the `Capture` and `VerifyFunctionCallCount`
- Using the test cases directly, which results in this slightly weird calling style

I'm a bit torn between 2 and 3. I'm afraid that going the route 2 will lead to a lot of functionality being stripped out into free functions over time again, which is ok, but gives up some of the convenience of the TestCase class.

I think it's also a bit of a core question on how independent a single TestCase actually is, and I tried to find a good balance between independence and reusability. There's preconditions to some test cases, like being connected to an instance, or having a capture taken, but apart from that, I tried to have all test cases verify their success by all means necessary without the need of the test suite to know about this.

WDYT?

ronaldfw (approved):
LGTM! Only looked briefly at E2E tests for iterators and frame tracks, but looks very nice!

contrib/automation_tests/core/orbit_e2e.py
danielfenner: sgtm!

contrib/automation_tests/fragments/connection_window.py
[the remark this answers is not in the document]
danielfenner: nice

contrib/automation_tests/fragments/symbols_tab.py
danielfenner: I really haven't thought of a TestCase to be entirely self-contained. Is this really necessary? As you said the tests currently depend on `ConnectToStadiaInstance` being it the suite before them anyway; they don't call that themselves.

That said I'm totally fine with it the way it is - I don't think this is importent - I just wondered (and 3 looks better than 2 to me fwiw).

contrib/automation_tests/fragments/connection_window.py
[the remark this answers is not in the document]
danielfenner: lg!

danielfenner (approved):
lgtm!

PR #1625, opened 2021-01-03
Add diff generation support to the include-what-you-use presubmit check

beckerhe:


beckerhe:
Note: The license header check is (once again) failing due to a change in orbitcodeeditor.cpp. Time to get rid of it / move it, but that's out of scope of this PR.

cmake/iwyu.cmake
```diff
@@ -7,7 +7,11 @@ if(NOT WIN32)
   ProcessorCount(CPU_COUNT)
 
   find_package(Python3 COMPONENTS Interpreter)
-  find_program(iwyu_tool_py NAMES iwyu_tool)
+  find_program(iwyu_tool_py NAMES iwyu_tool iwyu_tool.py)
+  set(fix_includes_py ${CMAKE_SOURCE_DIR}/third_party/include-what-you-use/fix_includes.py)
```
florian-kuebler, line 11: Don't we usually use CAPS for CMake variables?

cmake/iwyu.cmake
```diff
@@ -20,15 +24,56 @@ if(NOT WIN32)
     # That's the public facing include-what-you-use target which can be invoked by `make iwyu`.
     # It generates the include-what-you-use report and writes it into iwyu_results.txt in the
```
florian-kuebler, line 25: I guess the comment needs to be updated too. We are not creating `iwyu_results.txt` anymore, aren't we?

cmake/iwyu.cmake
```diff
@@ -20,15 +24,56 @@ if(NOT WIN32)
...
       VERBATIM)
+
+    add_custom_command(OUTPUT iwyu_unformatted.diff
+      COMMAND "${Python3_EXECUTABLE}" "${fix_includes_py}" -n -p ${CMAKE_SOURCE_DIR}
+        < include-what-you-use.log > iwyu_unformatted.diff
+      DEPENDS include-what-you-use.log
+      WORKING_DIRECTORY ${CMAKE_BINARY_DIR}
+      VERBATIM)
+
+    set(wrapper_content "\
```
florian-kuebler, line 46: As above, shouldn't it be CAPS?

kokoro/builds/build.sh
```diff
@@ -209,16 +209,17 @@ if [ -n "$1" ]; then
     pushd ${REPO_ROOT} > /dev/null
 
     echo -e "\n\n\nHere is the full list of all things iwyu found:"
-    cat build/iwyu_results.txt
+    cat build/include-what-you-use.log
 
-    echo -e "\n\n\nThe following include problems occur in files this PR touched:"
+    echo -e "\n\n\nThe following is a diff that fixes all problems in files this PR touched:"
```
florian-kuebler, line 214: It would be nice if the script would not just output the plain diff, but also the command to apply this diff, such that you can just copy&paste the line.

third_party/conan/docker/tags.sh
```diff
@@ -0,0 +1,12 @@
+# Copyright (c) 2020 The Orbit Authors. All rights reserved.
+# Use of this source code is governed by a BSD-style license that can be
+# found in the LICENSE file.
+
+# Maps conan profiles to docker image tags. The images are expected to be named after the conan profile,
+# but the tag can be adjusted here to support versioning of docker containers.
+declare -rA docker_image_tag_mapping=( \
+  [iwyu]="2" \
```
florian-kuebler, line 8: From the description, it is a bit unclear to me, why there is only the mapping for iwyu. Could you clarify the comment?

cmake/iwyu.cmake
beckerhe: Yes and no. :sweat_smile: Old-style cmake uses (all-caps) variables to transport information between scopes.
So you're always at risk of overwriting one of these variables. So I try to use lower case variables for all things local - if I  don't forget about it. So I'd rather stick to lower case here.

And yes it's not consistent. `CPU_COUNT` should be lower-case. I'm going to fix this.

cmake/iwyu.cmake
beckerhe: Yessss... Thanks. Will fix that!

kokoro/builds/build.sh
beckerhe: Good idea!

third_party/conan/docker/tags.sh
beckerhe: Yes, makes sense. I will explain that in the comment.

beckerhe: Done.

kokoro/builds/build.sh
beckerhe: Done.

cmake/iwyu.cmake
beckerhe: Done.

cmake/iwyu.cmake
florian-kuebler: Okay, thanks for clarification. At least it is consistent now.

florian-kuebler (approved):
Thanks for addressing the review comments. LGTM!

PR #1626, opened 2021-01-03
Replace isgreater by std::isgreater

beckerhe:
LayerLogic.cpp uses the C99 function isgreater. We should avoid using
C-functions and use the C++ equivalent std::isgreater instead.

This also triggered a bug in include-what-you-use. That's why I'm fixing
it.

@florian-kuebler: I'm not sure why `isgreater` has been used. Do you happen to know
whether it's special behaviour is actually needed here? Using `operator >` would be a lot more readable to me.

### DIFF
--- a/OrbitTriggerCaptureVulkanLayer/LayerLogic.cpp
+++ b/OrbitTriggerCaptureVulkanLayer/LayerLogic.cpp
@@ -8,6 +8,7 @@
 #include <unistd.h>
 
 #include <chrono>
+#include <cmath>
 #include <cstdint>
 #include <string>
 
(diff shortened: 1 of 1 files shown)


dimitry- (approved):

florian-kuebler (approved):
LGTM. But I also don't know why `isgreater` was used in the first place. Maybe we should go for the more readable `>`.

beckerhe:
> LGTM. But I also don't know why `isgreater` was used in the first place. Maybe we should go for the more readable `>`.

Thank you both. Since I don't know the code here very well I was a bit reluctant to change the behaviour. So I'd rather go with std::isgreater for now and we can change that later if needed.

PR #1627, opened 2021-01-04
Make configs/install.{sh,ps1} detect catchall DNS

beckerhe:
Some ISPs have their DNS servers configured in such a way
that they return an A or AAAA record even for invalid DNS
names. That allows them to show the user a "Domain Name Not Found"
website - at least when the wrong request originates from a browser.

This behaviour breaks our CI machine detection which relied on
making an HTTP request to a hostname that only exists internally.

This commit fixes that by making another HTTP request which should does
not exist inside of the CI environment. If it succeeds we will know that
we hit the ISP's catchall website.

Bug: http://b/176184061
Test: Manually tested


pierricgimmig (approved):
I tested the change, it works, thanks!

PR #1628, opened 2021-01-05
Fix clang_format/license_header checks on release branches

beckerhe:
These checks were broken due to the recently introduced automatic
detection of git target branch names. The kokoro variable contains the
branch name but for a checkout we also need to specify the remote name
`origin`. It happened to work for `master` because it's the default
branch and a local copy exists. That's not the case for a release
branch.

Bug: http://b/175678721
Test: Tested by creating a PR against an old release branch

### DIFF
--- a/kokoro/checks/clang_format/check.sh
+++ b/kokoro/checks/clang_format/check.sh
@@ -28,7 +28,7 @@ if [ "$0" == "$SCRIPT" ]; then
 
   cd /mnt
   # Use origin/master as reference branch, if not specified by kokoro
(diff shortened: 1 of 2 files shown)


beckerhe:
Just a friendly ping for visibility.

This should go into master before branching 1.58 on Wed, because otherwise this problem will also occur on the 1.58 release branch.

dimitry- (approved):

danielfenner (approved):
lgtm

beckerhe:
Thank you both!

PR #1629, opened 2021-01-05
Double-click should load presets and symbols.

vickyliu-go4it:
With this change: double-click in the preset view will load preset, and; double click in the modules view will load symbols.

Fixes bug: b/167661761.

### DIFF
--- a/OrbitQt/orbittreeview.cpp
+++ b/OrbitQt/orbittreeview.cpp
@@ -47,6 +47,8 @@ OrbitTreeView::OrbitTreeView(QWidget* parent) : QTreeView(parent), auto_resize_(
 
   connect(verticalScrollBar(), SIGNAL(rangeChanged(int, int)), this,
           SLOT(OnRangeChanged(int, int)));
+
+  connect(this, &OrbitTreeView::doubleClicked, this, &OrbitTreeView::OnDoubleClicked);
 }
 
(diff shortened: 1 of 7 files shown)


OrbitQt/orbittreeview.cpp
```diff
@@ -260,6 +260,19 @@ void OrbitTreeView::mousePressEvent(QMouseEvent* event) {
...
+    return;
+  }
+
+  QModelIndex index = indexAt(event->pos());
+  if (index.isValid()) {
+    model_->GetDataView()->OnDoubleClick(index.row());
+  } else {
+    QTreeView::mouseDoubleClickEvent(event);
+  }
+}
```
beckerhe, line 274: That seems to work, but is probably not the intended way. There is a signal called `doubleClicked` which you can use. Then you don't have to deal with events. More details here: https://doc.qt.io/qt-5/qabstractitemview.html#doubleClicked

beckerhe (approved):
Hi Tingwei,
thanks a lot for this change. I just added one comment in OrbitTreeView. Otherwise LGTM!

OrbitQt/orbittreeview.cpp
vickyliu-go4it: Changed to use `doubleClicked`. Thanks for the comments!

karupayun (approved):
LGTM! :)

PR #1630, opened 2021-01-05
Use information from MapPerfEvent to update process maps

dimitry-:
Instead of reading /proc/pid/maps on every MapPerfEvent
update existing map using information from MapsPerfEvent

Also add a test process that dlopens a library and calls it to
test this scenario.

Test: Run capture OrbitLinuxTracingTestPuppet, check stack-traces.

### DIFF
--- a/OrbitGrpcProtos/capture.proto
+++ b/OrbitGrpcProtos/capture.proto
@@ -225,13 +225,14 @@ message AddressInfo {
   }
 }
 
-message ModulesUpdateEvent {
+message ModuleUpdateEvent {
   int32 pid = 1;
   uint64 timestamp_ns = 2;
-  repeated ModuleInfo modules = 3;
+  ModuleInfo module = 3;
 }
 
 message CaptureEvent {
+  reserved 14;
   oneof event {
     SchedulingSlice scheduling_slice = 1;
     InternedCallstack interned_callstack = 2;
@@ -246,6 +247,6 @@ message CaptureEvent {
     InternedTracepointInfo interned_tracepoint_info = 9;
     TracepointEvent tracepoint_event = 10;
     IntrospectionScope introspection_scope = 12;
-    ModulesUpdateEvent modules_update_event = 14;
+    ModuleUpdateEvent modules_update_event = 15;
   }
 }
--- a/OrbitLinuxTracing/CMakeLists.txt
+++ b/OrbitLinuxTracing/CMakeLists.txt
@@ -67,21 +67,37 @@ target_link_libraries(OrbitLinuxTracing PUBLIC
         CONAN_PKG::abseil
         CONAN_PKG::libunwindstack)
 
+add_executable(OrbitLinuxTracingTestDlopenPuppet)
+
+target_compile_options(OrbitLinuxTracingTestDlopenPuppet PRIVATE ${STRICT_COMPILE_FLAGS})
+
+target_sources(OrbitLinuxTracingTestDlopenPuppet PRIVATE
+        TestPuppet.cpp)
+
+target_link_libraries(OrbitLinuxTracingTestDlopenPuppet PUBLIC
+        dl
+        OrbitBase)
+
+add_library(OrbitLinuxTracingTestPuppetLibrary SHARED)
+
+target_compile_options(OrbitLinuxTracingTestPuppetLibrary PRIVATE ${STRICT_COMPILE_FLAGS})
+
+target_sources(OrbitLinuxTracingTestPuppetLibrary PRIVATE
+        TestPuppetLibrary.cpp)
+
 add_executable(OrbitLinuxTracingTests)
 
 target_compile_options(OrbitLinuxTracingTests PRIVATE ${STRICT_COMPILE_FLAGS})
 
-if (NOT WIN32)
-    target_sources(OrbitLinuxTracingTests PRIVATE
-            ContextSwitchManagerTest.cpp
-            GpuTracepointEventProcessorTest.cpp
-            LinuxTracingUtilsTest.cpp
-            PerfEventProcessorTest.cpp
-            PerfEventQueueTest.cpp
-            ThreadStateManagerTest.cpp
-            UprobesFunctionCallManagerTest.cpp
-            UprobesReturnAddressManagerTest.cpp)
-endif()
+target_sources(OrbitLinuxTracingTests PRIVATE
+        ContextSwitchManagerTest.cpp
+        GpuTracepointEventProcessorTest.cpp
+        LinuxTracingUtilsTest.cpp
+        PerfEventProcessorTest.cpp
+        PerfEventQueueTest.cpp
+        ThreadStateManagerTest.cpp
+        UprobesFunctionCallManagerTest.cpp
+        UprobesReturnAddressManagerTest.cpp)
 
 target_link_libraries(OrbitLinuxTracingTests PRIVATE
         OrbitLinuxTracing
--- a/OrbitLinuxTracing/PerfEvent.h
+++ b/OrbitLinuxTracing/PerfEvent.h
@@ -305,24 +305,27 @@ class UretprobesPerfEvent : public PerfEvent, public AbstractUprobesPerfEvent {
   uint32_t GetCpu() const { return ring_buffer_record.sample_id.cpu; }
 };
 
-// This carries a snapshot of /proc/<pid>/maps and does not reflect a
-// perf_event_open event, but we want it to be part of the same hierarchy.
-class MapsPerfEvent : public PerfEvent {
+class MmapPerfEvent : public PerfEvent {
  public:
-  MapsPerfEvent(int32_t pid, uint64_t timestamp, std::string maps)
-      : pid_{pid}, timestamp_{timestamp}, maps_{std::move(maps)} {}
+  MmapPerfEvent(int32_t pid, uint64_t timestamp, const perf_event_mmap_up_to_pgoff& mmap_event,
+                std::string filename)
+      : pid_{pid}, timestamp_{timestamp}, mmap_event_{mmap_event}, filename_{std::move(filename)} {}
 
   uint64_t GetTimestamp() const override { return timestamp_; }
 
   void Accept(PerfEventVisitor* visitor) override;
 
-  [[nodiscard]] const std::string& GetMaps() const { return maps_; }
-  [[nodiscard]] int32_t GetPid() const { return pid_; }
+  [[nodiscard]] const std::string& filename() const { return filename_; }
+  [[nodiscard]] int32_t pid() const { return pid_; }
+  [[nodiscard]] uint64_t address() const { return mmap_event_.address; }
+  [[nodiscard]] uint64_t length() const { return mmap_event_.length; }
+  [[nodiscard]] uint64_t page_offset() const { return mmap_event_.page_offset; }
 
  private:
   int32_t pid_;
   uint64_t timestamp_;
-  std::string maps_;
+  perf_event_mmap_up_to_pgoff mmap_event_;
+  std::string filename_;
 };
 
 class GenericTracepointPerfEvent : public PerfEvent {
--- a/OrbitLinuxTracing/PerfEventReaders.h
+++ b/OrbitLinuxTracing/PerfEventReaders.h
@@ -21,7 +21,11 @@ namespace orbit_linux_tracing {
 
 // Helper functions for reads from a perf_event_open ring buffer that require
 // more complex operations than simply copying an entire perf_event_open record.
-pid_t ReadMmapRecordPid(PerfEventRingBuffer* ring_buffer);
+
+// This function reads sample_id, which is always the last field
+// in the perf event record unless it is PERF_RECORD_SAMPLE.
+void ReadPerfSampleIdAll(PerfEventRingBuffer* ring_buffer, const perf_event_header& header,
+                         perf_event_sample_id_tid_time_streamid_cpu* sample_id);
 
 uint64_t ReadSampleRecordTime(PerfEventRingBuffer* ring_buffer);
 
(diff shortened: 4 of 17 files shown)


OrbitGrpcProtos/capture.proto
```diff
@@ -226,9 +226,10 @@ message AddressInfo {
 }
 
 message ModulesUpdateEvent {
+  reserved 3;
   int32 pid = 1;
   uint64 timestamp_ns = 2;
-  repeated ModuleInfo modules = 3;
+  ModuleInfo module = 4;
```
dpallotti, line 232: Given that we don't need compatibility with different versions here, can we keep the field numbers clean instead of already having `reserved` numbers?

OrbitGrpcProtos/capture.proto
```diff
@@ -226,9 +226,10 @@ message AddressInfo {
 }
 
 message ModulesUpdateEvent {
```
dpallotti, line 228: Maybe rename to `ModuleUpdateEvent` now that it's only one?

OrbitLinuxTracing/PerfEventRecords.h
```diff
@@ -139,6 +139,17 @@ struct __attribute__((__packed__)) perf_event_lost {
   perf_event_sample_id_tid_time_streamid_cpu sample_id;
 };
 
+// This structure represents mmap event up to filename
+struct __attribute__((__packed__)) perf_event_mmap_up_to_filename {
```
dpallotti, line 143: This is not "up to filename" as `filename` is excluded, rather "up to pgoff" (or at least it's ambiguous). Also adjust the comment.

OrbitLinuxTracing/TestPuppetLibrary.cpp
```diff
@@ -0,0 +1,18 @@
+// Copyright (c) 2020 The Orbit Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#include <cmath>
+
+// extern "C" to have unmangled name in the symbol table
+extern "C" double function_that_that_works_for_considerable_amount_of_time();
```
dpallotti, line 8: Remove redundant "that".

OrbitLinuxTracing/TestPuppet.cpp
```diff
@@ -0,0 +1,50 @@
...
+// found in the LICENSE file.
+
+// This is test executable used by LinuxTracing tests to test profiling functionality
+// The control is done by executing commands
+
+#include <iostream>
+#include <string>
+#include <thread>
+
+
```
dpallotti: Remove empty line.

OrbitLinuxTracing/CMakeLists.txt
```diff
@@ -67,21 +67,45 @@ target_link_libraries(OrbitLinuxTracing PUBLIC
         CONAN_PKG::abseil
         CONAN_PKG::libunwindstack)
 
+set(CMAKE_SKIP_BUILD_RPATH FALSE)
+set(CMAKE_BUILD_WITH_INSTALL_RPATH FALSE)
+set(CMAKE_INSTALL_RPATH $ORIGIN)
+
+
+add_executable(OrbitLinuxTracingTestPuppet)
```
dpallotti, line 75: Two things about this. 

First, I don't like the name "puppet", it should more clearly state what it's doing, i.e., dlopen.

Second, the only place where we have this kind of targets to profile for manual tests is `OrbitTest`. I would keep it like that and move this to `OrbitTest`.

OrbitLinuxTracing/TracerThread.cpp
```diff
@@ -766,16 +766,52 @@ void TracerThread::ProcessExitEvent(const perf_event_header& header,
...
+  size_t filename_offset = sizeof(perf_event_mmap_up_to_filename);
+  // strictly > because filename is null-terminated string
+  CHECK(header.size > (filename_offset + sizeof(perf_event_sample_id_tid_time_streamid_cpu)));
+  size_t filename_size =
+      header.size - filename_offset - sizeof(perf_event_sample_id_tid_time_streamid_cpu);
+  std::vector<char> filename_vector(filename_size);
+  ring_buffer->ReadRawAtOffset(&filename_vector[0], filename_offset, filename_size);
+  std::string filename(&filename_vector[0], filename_size);
+
   ring_buffer->SkipRecord(header);
```
dpallotti, line 798: Move this code to a `ConsumeMmapPerfEvent` in `PerfEventReaders.h`, which is precisely where we have functions to read `PerfEvent`s that are not trivial to read.

It's fine to create the whole `MapsPerfEvent` before the `if`s that follow, it's a rare event anyway.

OrbitLinuxTracing/PerfEventRecords.h
```diff
@@ -139,6 +139,17 @@ struct __attribute__((__packed__)) perf_event_lost {
...
+  perf_event_header header;
+  uint32_t pid;
+  uint32_t tid;
+  uint64_t address;
+  uint64_t length;
+  uint64_t page_offset;
+  // OMITTED: char filename[]
```
dpallotti, line 150: `perf_event_sample_id_tid_time_streamid_cpu sample_id` is also omitted.

OrbitLinuxTracing/TracerThread.cpp
```diff
@@ -766,16 +766,52 @@ void TracerThread::ProcessExitEvent(const perf_event_header& header,
...
+  // };
+  // Because of filename, the layout is not fixed.
+
+  perf_event_sample_id_tid_time_streamid_cpu sample_id;
+  ReadPerfSampleIdAll(ring_buffer, header, &sample_id);
+
+  perf_event_mmap_up_to_filename mmap_event;
+  ring_buffer->ReadValueAtOffset(&mmap_event, 0);
+
+  //  ring_buffer->ConsumeRecord()
```
dpallotti, line 787: Remove this.

OrbitLinuxTracing/TestPuppetLibrary.cpp
dpallotti, line 8: Can't you just add `extern "C"` to the function definition instead of having a separate declaration?

OrbitLinuxTracing/TracerThread.cpp
```diff
@@ -766,16 +766,52 @@ void TracerThread::ProcessExitEvent(const perf_event_header& header,
...
+  if (sample_id.time < effective_capture_start_timestamp_ns_) {
     return;
   }
 
-  // There was a call to mmap with PROT_EXEC, hence refresh the maps. This should happen rarely.
+  // There was a call to mmap with PROT_EXEC, hence refresh the maps.
+  // This should happen rarely.
+  // Note that we do not use filename from the record itself instead we read all maps
+  // from /proc/pid/maps. Consider moving this to MMAP2 event which has more information (like
+  // flags)
```
dpallotti: Isn't this comment wrong? Isn't this moving away from reading `/proc/pid/maps`? Isn't `filename` actually used?

OrbitLinuxTracing/UprobesUnwindingVisitor.cpp
```diff
@@ -186,19 +187,28 @@ void UprobesUnwindingVisitor::visit(UretprobesPerfEvent* event) {
...
-  if (!result_or_error) {
-    ERROR("Couldn't parse maps: %s", result_or_error.error().message());
+  ErrorMessageOr<orbit_grpc_protos::ModuleInfo> module_info_result =
+      orbit_elf_utils::CreateModule(filename, event->address(), event->address() + event->length());
+  if (!module_info_result) {
+    ERROR("Unable create module : %s", module_info_result.error().message());
     return;
   }
 
+  auto& module_info = module_info_result.value();
```
dpallotti, line 199: `const`?

OrbitGrpcProtos/capture.proto
dimitry-: well, numbers are free - this is good practice and we should probably do this by default even if this is not really needed at this point. There are 0 downsides to this.

OrbitLinuxTracing/PerfEventRecords.h
dimitry-: done.

OrbitLinuxTracing/TestPuppetLibrary.cpp
dimitry-: done.

OrbitLinuxTracing/TracerThread.cpp
dimitry-: done.

OrbitLinuxTracing/UprobesUnwindingVisitor.cpp
dimitry-: std::move in line 211

OrbitGrpcProtos/capture.proto
dimitry-: right. done.

OrbitLinuxTracing/TestPuppet.cpp
dimitry-: done.

OrbitLinuxTracing/TracerThread.cpp
dimitry-: yep - updated this as well.

OrbitGrpcProtos/capture.proto
dpallotti: Well, there is one downside to this: it litters the proto. I don't like leaving litter behind when it's not necessary, even if it's a good (or necessary) practice in a different context.

Also this is now basically a new proto as it has different semantics (it was carrying the entire parsed maps before, it carries a single new module now, and indeed it made sense to rename it because of that), so it makes even less sense to have a field number already obsolete.

OrbitLinuxTracing/CMakeLists.txt
dpallotti: Alright.

OrbitLinuxTracing/PerfEventReaders.h
```diff
@@ -19,9 +19,10 @@
 
 namespace orbit_linux_tracing {
 
-// Helper functions for reads from a perf_event_open ring buffer that require
-// more complex operations than simply copying an entire perf_event_open record.
```
dpallotti, line 23: This was not a comment on the function but on the whole file, do you want it removed?

OrbitLinuxTracing/PerfEventReaders.h
```diff
@@ -19,9 +19,10 @@
...
-pid_t ReadMmapRecordPid(PerfEventRingBuffer* ring_buffer);
+// This function reads sample_id , which always is the last field
```
dpallotti, line 22:
```suggestion
// This function reads sample_id, which is always the last field
```

OrbitLinuxTracing/PerfEvent.h
```diff
@@ -309,20 +309,25 @@ class UretprobesPerfEvent : public PerfEvent, public AbstractUprobesPerfEvent {
 // perf_event_open event, but we want it to be part of the same hierarchy.
 class MapsPerfEvent : public PerfEvent {
  public:
-  MapsPerfEvent(int32_t pid, uint64_t timestamp, std::string maps)
-      : pid_{pid}, timestamp_{timestamp}, maps_{std::move(maps)} {}
+  MapsPerfEvent(int32_t pid, uint64_t timestamp, const perf_event_mmap_up_to_pgoff& mmap_event,
```
dpallotti, line 312: As this now now matches a `PERF_RECORD_MMAP` record (unlike before), it should be called `MmapPerfEvent`.

OrbitLinuxTracing/TestPuppet.cpp
```diff
@@ -0,0 +1,49 @@
...
+#include "OrbitBase/ExecutablePath.h"
+#include "OrbitBase/Logging.h"
+
+// dlopen and call function_that_sleeps_for_one_second 10 times.
+static void run_dlopen() {
+  static const char* kLibraryFile = "libOrbitLinuxTracingTestPuppetLibrary.so";
+  static const char* kFunctionName = "function_that_works_for_considerable_amount_of_time";
+  // Setting rpath in CMAKE is a nightmare.. I was not able to do that, so we are going
+  // to emulate "$ORIGIN/../lib" rpath here.
```
dpallotti, line 22: So what are
```
set(CMAKE_SKIP_BUILD_RPATH FALSE)
set(CMAKE_BUILD_WITH_INSTALL_RPATH FALSE)
set(CMAKE_INSTALL_RPATH $ORIGIN)
```
in the `CMakeLists.txt` file for? Maybe add a comment to those as well?

OrbitLinuxTracing/UprobesUnwindingVisitor.cpp
dpallotti: Right.

OrbitLinuxTracing/UprobesUnwindingVisitor.cpp
```diff
@@ -186,19 +187,28 @@ void UprobesUnwindingVisitor::visit(UretprobesPerfEvent* event) {
...
-  current_maps_ = LibunwindstackUnwinder::ParseMaps(event->GetMaps());
+  const std::string& filename = event->filename();
 
-  auto result_or_error = orbit_elf_utils::ParseMaps(event->GetMaps());
-  if (!result_or_error) {
-    ERROR("Couldn't parse maps: %s", result_or_error.error().message());
+  ErrorMessageOr<orbit_grpc_protos::ModuleInfo> module_info_result =
+      orbit_elf_utils::CreateModule(filename, event->address(), event->address() + event->length());
+  if (!module_info_result) {
+    ERROR("Unable create module : %s", module_info_result.error().message());
```
dpallotti, line 195:
```suggestion
    ERROR("Unable create module: %s", module_info_result.error().message());
```

OrbitLinuxTracing/PerfEvent.h
dimitry-: done.

dpallotti (approved):

PR #1632, opened 2021-01-05
Add Vulkan related Timer types.

florian-kuebler:
This adds timers representing Vulkan command buffer timings and
Vulkan debug markers to the capture protos.

Test: Compile
Bug: http://b/176809754

### DIFF
--- a/OrbitClientProtos/capture_data.proto
+++ b/OrbitClientProtos/capture_data.proto
@@ -164,10 +164,10 @@ message TimerInfo {
 
 message Color {
   // Each color must be between 0.0f and 1.0f
-  float red = 1;
-  float green = 2;
-  float blue = 3;
-  float alpha = 4;
+  uint32 red = 1;
+  uint32 green = 2;
+  uint32 blue = 3;
+  uint32 alpha = 4;
 }
 
 // libprotobuf-mutator needs a single proto with all the data
--- a/OrbitClientProtos/capture_data.proto
+++ b/OrbitClientProtos/capture_data.proto
@@ -163,7 +163,7 @@ message TimerInfo {
 }
 
 message Color {
-  // Each color must be between 0.0f and 1.0f
+  // Each color must be between 0 and 255 (including).
   uint32 red = 1;
   uint32 green = 2;
   uint32 blue = 3;
--- a/OrbitClientProtos/capture_data.proto
+++ b/OrbitClientProtos/capture_data.proto
@@ -148,6 +148,8 @@ message TimerInfo {
     kIntrospection = 2;
     kGpuActivity = 3;
     kFrame = 4;
+    kGpuCommandBuffer = 5;
+    kGpuDebugMarker = 6;
   }
   Type type = 6;
 
@@ -157,6 +159,15 @@ message TimerInfo {
   uint64 user_data_key = 10;
   uint64 timeline_hash = 11;
   repeated uint64 registers = 12;
+  Color color = 13;
+}
+
+message Color {
+  // Each color must be between 0.0f and 1.0f
+  float red = 1;
+  float green = 2;
+  float blue = 3;
+  float alpha = 4;
 }
 
 // libprotobuf-mutator needs a single proto with all the data


dimitry- (commented):
I think I need more context here. How is this used?

florian-kuebler:
Sure, so similar to `kGpuActivity` timers (which display Gpu information based on tracepoints), the command buffer and debug marker timers, are timers that should display Gpu related information (this time originating from our new Vulkan layer).

There will be another PR, that extents the CaptureEvenProcessor in order to produce those timers, base on the corresponding `CaptureEvents`. I am working on that one right now.
Further #1642 adds handling for those new timer types in the GpuTrack. Command buffer timers will basically displayed below the corresponding "hw execution" timers that are based on the tracepoints. The debug markers will be displayed on their own track for now, but combined later. 
Further, the debug marker interface allows the game devs. to specify a color. This color should be then also used in our timeline.

If you need any additional input, feel free to set up a meeting to discuss offline.

Also note, we are aware that the way TimerInfo is encoded with its different types is not ideal, and want to work on this (e.g. by introducing "subtypes") in the future. But this is clearly out of scope for this change.

OrbitClientProtos/capture_data.proto
```diff
@@ -157,6 +159,15 @@ message TimerInfo {
   uint64 user_data_key = 10;
   uint64 timeline_hash = 11;
   repeated uint64 registers = 12;
+  Color color = 13;
```
ronaldfw, line 162: We need to be careful about adding additional data to TimerInfo, simply for memory reasons (not during transfer, but in-memory). I think this "only" adds a pointer to a TimerInfo, but we already have a different way to handle colors. The manual instrumentation timers also have a color that can be customized.

ronaldfw: @pierricgimmig How are colors handled for manual instrumentation markers?

florian-kuebler: For manual instrumentation Colors are encoded as part of the `orbit_api::EncodedEvent`, which is encoded using `TimerInfo`'s register fields. You may want to have a look at `ThreadTrack::GetUserColor`. 
To be honest, I have high doubts in also using this for the markers here. The new `Color` field should cost no memory if unused and if we using it, it should be still better than using the 6 64bit registers for encoding 4 float values.

florian-kuebler: Also, I guess mid-term we really want to split the TimerInfo into different protos for each test.

ronaldfw: > The new Color field should cost no memory if unused

It's a pointer in the message, so 8 bytes per TimerInfo, even if the TimerInfo does not use color.

florian-kuebler: >>The new Color field should cost no memory if unused

>It's a pointer in the message, so 8 bytes per TimerInfo, even if the TimerInfo does not use color.

Oh, I thought there was some magic going on, such that unused fields do not use memory at all (even for primitive types). But nevertheless, my argument would be to make it clean and then start separating the timers rather sooner then later.

ronaldfw: Fields don't use space in encoded messages if they are not present, but the generated code has to have a field for them. You can look at the generated code easily (in the build folder). 

I'm not convinced we should have separate messages for all types, but let's have that discussion when we get there. 

Adding the color message is not going to bloat memory too much, relatively speaking, but I want to make sure we consider changes carefully. 

Another thing to consider here is the serialized size (TimerInfo are used in the serialized capture file). floats are always stored as 4 bytes in encoded format, so a capture with a lot of GPU timers could potentially be a lot larger than necessary. How would you feel about storing the colors as 4 int values and just use values [0, 255]?

florian-kuebler: Storing the values as ints [0,255] would be totally fine for me (can't we even use uint8 here?).

Also agree that we should discuss the rest in detail, when it is the time for this.

florian-kuebler: Done!

ronaldfw: There is no uint8/int8 type in protos. But for the encoded format that doesn't matter.

florian-kuebler: Right, also noticed this. I used uint32. Should be fine though.

florian-kuebler: @ronaldfw is there anything else I should take care of?

ronaldfw (approved):
LGTM

PR #1633, opened 2021-01-05
Changed TimerTrack::GetYFromDepth to TimerTrack::GetYFromTimer.

florian-kuebler:
In order to share the implementation of GpuTracks with and without
command buffer information, GpuTrack's `GetY...` implementations
needs to be aware of the timer type (in addition to the plain depth).
Therefore, this changes the method signature, to pass a timer
instead of only the timer's depth.

Test: Compile
Bug: http://b/176809754


karupayun (approved):
LGTM!! :)

PR #1634, opened 2021-01-05
Use `switch` instead of `if(...) else if(...)` in TimerGraph.

florian-kuebler:
This changes the switch over the different timer types from
an `if(...) else if(...)` construct to a switch statement,
which makes the code more readable and error prone (we will fail
if we miss a case).
Also it avoids having a "default" case (as present before) to
make the code more robust by explicitly stating the expected
case.

Test: Compile

### DIFF
--- a/OrbitGl/TimeGraph.cpp
+++ b/OrbitGl/TimeGraph.cpp
@@ -271,6 +271,9 @@ void TimeGraph::ProcessTimer(const TimerInfo& timer_info, const FunctionInfo* fu
       break;
     }
     case TimerInfo::kCoreActivity: {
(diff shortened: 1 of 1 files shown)


karupayun (commented):
Hi Florian, I really liked the change.

BTW, the code runs strangely. Testing a little bit, we are having half the tracks that we used to have, because I'm only seeing ThreadTracks with instrumented functions. Do you know what could had happened?

florian-kuebler:
Thanks for the review and the test. I found the issue, fixed it (the way it was before) and created a bug for a proper fix in the future.

karupayun (approved):
Thanks Florian for addressing my comment and creating the bug :)

LGTM!

PR #1635, opened 2021-01-05
Import std::string and std::to_string

sidhys1:
Added "using std::string;" and "using std::to_string;"


sidhys1:
@googlebot I signed it!

pierricgimmig:
We always qualify std:: types.

PR #1636, opened 2021-01-07
Fix clean-up routine in Kokoro's Build script

beckerhe:
There was a backslash missing which made the `-delete` option be
interpreted as an independent command. Fixed by the commit.


danielfenner (approved):
lgtm, thanks for fixing this!

PR #1637, opened 2021-01-07
Add "Capture Options" button and dialog to toggle the collection of thread states

dpallotti:
Bug: http://b/176791430

Screenshot:
CaptureOptionsDialog2


OrbitGl/App.cpp
```diff
@@ -839,15 +839,16 @@ bool OrbitApp::StartCapture() {
   }
```
dimitry-, line 839: third commit has 2 bug: stanzas - one before commit message and one after.

dimitry- (approved):

OrbitGl/App.cpp
dpallotti: Rearranged the commit message.

OrbitQt/CaptureOptionsDialog.h
```diff
@@ -0,0 +1,31 @@
...
+namespace orbit_qt {
+
+class CaptureOptionsDialog : public QDialog {
+  Q_OBJECT
+
+ public:
+  explicit CaptureOptionsDialog(QWidget* parent = nullptr);
+
+  void SetCollectThreadStates(bool collect_thread_state);
+  [[nodiscard]] bool GetCollectThreadStates();
```
beckerhe, line 23: const?

OrbitQt/orbitmainwindow.cpp
```diff
@@ -937,6 +939,22 @@ void OrbitMainWindow::on_actionToggle_Capture_triggered() { app_->ToggleCapture(
 
 void OrbitMainWindow::on_actionClear_Capture_triggered() { app_->ClearCapture(); }
 
+void OrbitMainWindow::on_actionCaptureOptions_triggered() {
+  static const QString kCollectThreadStates{"CollectThreadStates"};
+  QSettings settings;
+
+  auto dialog = std::make_unique<orbit_qt::CaptureOptionsDialog>(this);
```
beckerhe, line 956: Since `dialog` does not leave this local scope, this should be stack-allocated.

beckerhe (approved):
Very cool. I like the icon.

Two minor comments. Otherwise LGTM.

reichlfl (approved):
Thanks Davide, the code LGTM. Few thoughts from the UI side:

I'm not sure about the solution regarding the settings window - we've been talking about the need for a settings window for some time, so I guess it makes sense to introduce this, but obviously it also looks quite bland and empty with a single setting.

However, do we want to make this window capture-specific, especially by making it available through the settings-button in the Capture window rather than through the main menu? But I get the idea that this draws much more attention. As long as we only have capture-related settings here, it's fine, but what if we start adding additional options that may not directly influence taking the capture?

If we leave the button, maybe we want to clean up the toolbar a bit and remove e.g. Open / Save Capture as they are available in the File menu and more global than the capture window? Maybe also remove the "?" button and add it to the main menu? I'm worried we're wasting a lot of space with this.

Completely different suggestion: You can attach menus to QToolButtons. You could give access to the "Settings" by a menu-entry underneath the Toggle Capture - Button, this would make it super explicit that these are settings that affect the newly taken capture rather than the existing one.

You could also expose the setting directly in the menu like this:

image

by:

```
auto button =
      static_cast<QToolButton*>(ui->capture_toolbar->widgetForAction(ui->actionToggle_Capture));
  button->setMenu(ui->menuCaptureSettings); // New menu
  button->setPopupMode(QToolButton::ToolButtonPopupMode::MenuButtonPopup);
```

Just some food for thoughts. I'm fine with leaving it as-is. Mainly the popup looks a bit unfinished to me as an end-user. Maybe also some minimal design elements, like preparing a grouping for "Capture-related settings", or a hint "These settings will only affect your next capture", or both would improve on this... not sure, TBH, it's just that the dialog feels more like a notification popup than a settings dialog right now.

OrbitQt/CaptureOptionsDialog.h
dpallotti: Yep, thanks.

OrbitQt/orbitmainwindow.cpp
dpallotti: Ah yes, done.

dpallotti:
@reichlfl I share most of your thoughts, this came from Tuesday's discussion and as a start it's a compromised. Let's discuss offline if/what we can improve in this first version.

dpallotti:
@reichlfl I improved the dialog a bit based on our discussion. Here is a new screenshot. Let me know what you think.

CaptureOptionsDialog3

florian-kuebler:
> Thanks Davide, the code LGTM. Few thoughts from the UI side:
> 
> I'm not sure about the solution regarding the settings window - we've been talking about the need for a settings window for some time, so I guess it makes sense to introduce this, but obviously it also looks quite bland and empty with a single setting.
> 
> However, do we want to make this window capture-specific, especially by making it available through the settings-button in the Capture window rather than through the main menu? But I get the idea that this draws much more attention. As long as we only have capture-related settings here, it's fine, but what if we start adding additional options that may not directly influence taking the capture?
> 
> If we leave the button, maybe we want to clean up the toolbar a bit and remove e.g. Open / Save Capture as they are available in the File menu and more global than the capture window? Maybe also remove the "?" button and add it to the main menu? I'm worried we're wasting a lot of space with this.
> 
> Completely different suggestion: You can attach menus to QToolButtons. You could give access to the "Settings" by a menu-entry underneath the Toggle Capture - Button, this would make it super explicit that these are settings that affect the newly taken capture rather than the existing one.
> 
> You could also expose the setting directly in the menu like this:
> 
> image
> 
> by:
> 
> ```
> auto button =
>       static_cast<QToolButton*>(ui->capture_toolbar->widgetForAction(ui->actionToggle_Capture));
>   button->setMenu(ui->menuCaptureSettings); // New menu
>   button->setPopupMode(QToolButton::ToolButtonPopupMode::MenuButtonPopup);
> ```
> 
> Just some food for thoughts. I'm fine with leaving it as-is. Mainly the popup looks a bit unfinished to me as an end-user. Maybe also some minimal design elements, like preparing a grouping for "Capture-related settings", or a hint "These settings will only affect your next capture", or both would improve on this... not sure, TBH, it's just that the dialog feels more like a notification popup than a settings dialog right now.

Sorry for chiming in, but I guess the main "pain point" of this settings window, is that it only contains one setting so far. 
But we could even take the opportunity (not in this PR though), to e.g. also include the "sampling rate" setting now. Maybe add a marker that this is experimental, or that this may effect the performance.
Maybe we can even add the frame-pointer based unwinding here already, of course with a big mark that this is experimental/alpha-level and requires all binaries to be compiled with frame-pointers.
I agree with you Florian that we might also want to make this available via the main menu. We also discussed this on Tuesday and decided that we can add this at any time. For now that should probably be most user visible.

reichlfl:
> Sorry for chiming in, but I guess the main "pain point" of this settings window, is that it only contains one setting so far.
> But we could even take the opportunity (not in this PR though), to e.g. also include the "sampling rate" setting now. Maybe add a marker that this is experimental, or that this may effect the performance.
> Maybe we can even add the frame-pointer based unwinding here already, of course with a big mark that this is experimental/alpha-level and requires all binaries to be compiled with frame-pointers.
> I agree with you Florian that we might also want to make this available via the main menu. We also discussed this on Tuesday and decided that we can add this at any time. For now that should probably be most user visible.

We've discussed this offline before and agreed to leave the dialog, I wasn't aware you already talked about this on Tuesday. Basically, I think that we all agree it looks very empty and a bit out of place right now, but it will be the right place for those options in the future, so we have to start somewhere.

PR #1639, opened 2021-01-07
Draw timebar lines again

karupayun:
Regression from 1.55 (new drawing order).
We are using a wrong coordinate system to draw time bar lines.
Before 1.55 we had two different batchers to draw items for each
coordinate system and we had merged both in one. 

In this PR:
 - We draw time bar lines using screen space coordinate system.
 
 - We make a function to convert Qt-Screen To Gl-Screen 
 
 - We added a comment after a discussion with Pierric about Text should be drawn in screen space.

Related to http://b/175532525.

Test: Run Orbit + Load a Capture.

### DIFF
--- a/OrbitGl/CaptureWindow.cpp
+++ b/OrbitGl/CaptureWindow.cpp
@@ -563,6 +563,7 @@ void CaptureWindow::Draw() {
     ui_batcher_.DrawLayer(layer, GetPickingMode() != PickingMode::kNone);
 
     PrepareScreenSpaceViewport();
+    // Text needs to be drawn in screen space.
     if (GetPickingMode() == PickingMode::kNone) {
       text_renderer_.RenderLayer(&ui_batcher_, layer);
       RenderText(layer);
@@ -859,7 +860,8 @@ void CaptureWindow::RenderTimeBar() {
       text_renderer_.AddText(text.c_str(), world_x + x_margin, world_y, GlCanvas::kZValueTimeBar,
                              Color(255, 255, 255, 255), font_size_);
 
-      Vec2 pos(world_x, world_y);
+      // Lines from time bar are drawn in screen space.
+      Vec2 pos = QtScreenToGlScreen(WorldToScreen(Vec2(world_x, world_y)));
       ui_batcher_.AddVerticalLine(pos, height, GlCanvas::kZValueTimeBar, Color(255, 255, 255, 255));
     }
   }
--- a/OrbitGl/GlCanvas.cpp
+++ b/OrbitGl/GlCanvas.cpp
@@ -359,6 +359,14 @@ Vec2 GlCanvas::WorldToScreen(Vec2 world_pos) const {
   return screen_pos;
 }
 
+// TODO (b/177350599): Unify QtScreen and GlScreen
+// QtScreen(x,y) --> GlScreen(x,height-y)
+Vec2 GlCanvas::QtScreenToGlScreen(Vec2 qt_pos) const {
+  Vec2 gl_pos = qt_pos;
+  gl_pos[1] = GetHeight() - qt_pos[1];
+  return gl_pos;
+}
+
 int GlCanvas::WorldToScreenHeight(float height) const {
   return static_cast<int>(height / world_height_ * GetHeight());
 }
--- a/OrbitGl/GlCanvas.h
+++ b/OrbitGl/GlCanvas.h
@@ -65,6 +65,7 @@ class GlCanvas {
   Vec2 WorldToScreen(Vec2 world_pos) const;
   int WorldToScreenHeight(float height) const;
   int WorldToScreenWidth(float width) const;
+  Vec2 QtScreenToGlScreen(Vec2 qt_pos) const;
 
   // events
   virtual void MouseMoved(int x, int y, bool left, bool right, bool middle);


OrbitGl/CaptureWindow.cpp
```diff
@@ -526,13 +526,16 @@ void CaptureWindow::Draw() {
     // We use different coordinate systems for ScreenSpace items (margin, scrollbar, ...)
     // and for the text than the rest of items inside CaptureWindow. So, we have to switch
     // between these 2 systems while the layer is changing (with these "Prepare.." functions).
-    if (layer < GlCanvas::kScreenSpaceCutPoint) {
+
+    // Lines from TimeBar are also drawn in WorldSpace coordinate system
+    if (layer < GlCanvas::kScreenSpaceCutPoint || layer == GlCanvas::kZValueTimeBar) {
```
pierricgimmig, line 531: Hmmm, this seems a bit weird. I think the problem is that the time bar's background is rendered in screen space while the time lines are rendered in world space. They should be rendered in the same coordinate system, then their z-values will be on the same side of the "kScreenSpaceCutPoint" delimiter.

reichlfl: I agree, we should try to not introduce special cases like this after we've just spent time on simplifying the z-layering. This will just make it harder to catch bugs in the future.

If a good fix takes more time, I'd rather not fix it for the next release and fix it properly, because once this hack is in, other priorities may come along and this stays in forever until it causes issues.

karupayun: Yes, the problem is exactly that.

Ok. I didn't have time today to make it, but I will try to draw both of them in ScreenSpace coordinate system, which for me is what has more sense. 

Thanks!

karupayun:
I've moved the change directly where we are drawing the lines. For that, I've needed to solve an issue with WorldToScreen function. Let me know @reichlfl if it is ok for you.

OrbitGl/GlCanvas.cpp
```diff
@@ -352,7 +352,8 @@ float GlCanvas::ScreenToWorldWidth(int width) const {
 Vec2 GlCanvas::WorldToScreen(Vec2 world_pos) const {
   Vec2 screen_pos;
   screen_pos[0] = floorf((world_pos[0] - world_top_left_x_) / world_width_ * GetWidth());
-  screen_pos[1] = floorf((world_top_left_y_ - world_pos[1]) / world_height_ * GetHeight());
+  screen_pos[1] =
+      floorf((GetHeight() - (world_top_left_y_ - world_pos[1])) / world_height_ * GetHeight());
```
beckerhe, line 356: So this is moving the origin from top-left to bottom-left, right?

This is probably a stupid question, but how has this not affected other code so far? Was WorldToScreen unused?
(I'm sorry I'm lacking a bit of context hear. :sweat_smile: )

One remark regarding the code: It's rather unrelated to your change but affects this line. We should replace the C-function `floorf` by its C++ equivalent `std::floor`. (Can also be fixed somewhen else)

karupayun: Yes. We have 2 different coordinate systems, "screen space" and "world space". I'm not 100% sure about I'm talking below, correct me if I am wrong.

Screen space refers to point in the screen. Makes sense when we want to draw items that don't have to move when we navigate between events. They are centered in the bottom left point of the screen (0,0).

World space is more complicated. They use the world-top-left as a 0,0. This world-top-left seems to be mostly recalculated after each zoom/jump/etc. It seems to be only useful when users navigate in the TimeGraph using the mouse. When this happens, only world-top-left is an invariant after mouse-movements.

There are 2 things that sound strange for me. 
1- Why we aren't using the same point as a reference (bottom left)
2- Why world-coordinates hasn't always the (0,0) at the top/bottom-leftmost point in the capture. 

As I only trying to solve a bug that I've introduced, I didn't investigate why we are drawing on that way.

Second, AFAIR WorldToScreen was created recently for tutorials (is only used there), so I tagged @reichlfl for that.

And third, thanks for the floorf! :)

beckerhe: Ah great. Since WorldToScreen is new, then it makes a lot of sense.

beckerhe (approved):
So this makes sense to me and looks good to me. But Florian should also approve this.

OrbitGl/GlCanvas.cpp
reichlfl: Actually `WorldToScreen` was introduced for the accessibility changes, and this change will break the accessibility.

It's even more complicated than explained above, because what is referred to as "ScreenSpace" in methods like `DrawScreenSpace` is *not* was this method refers to as "ScreenSpace"... we have this weird space called screen space that is in pixel coordinates, but with the bottom left being (0, 0), which is used to draw the elements "around" the capture (like scrollbars, time bar etc).

But this method is the inverse of `ScreenToWorld`, which converts from the QT screen coordinate system to our world... this is used to e.g. convert mouse coordinates from QT to OpenGl. `ScreenToWorld` thus converts from OpenGl world to QT screen, and needs to stay as it is. You'll need to introduce a new method to convert to the space you need here. Ideally we would change the "Screen" space we use in OpenGl to actually be the same as QT Screen space, but this is more effort.

reichlfl (changes requested):
Welcome to the horrors of our coordinate systems :-)

OrbitGl/GlCanvas.h
```diff
@@ -50,6 +50,7 @@ class GlCanvas {
   Vec2 WorldToScreen(Vec2 world_pos) const;
   int WorldToScreenHeight(float height) const;
   int WorldToScreenWidth(float width) const;
+  Vec2 ScreenQtToScreenGl(Vec2 qt_pos) const;
```
reichlfl, line 53: [[nodiscard]] is missing - but it's also missing for all methods above, we could solve this in a seperate cleanup PR.

reichlfl (approved):
Thanks for the changes, Pablo, LGTM! I think introducing this as an additional function to convert between the two "screen spaces" is a good solution for now.

OrbitGl/GlCanvas.h
karupayun: Perfect, I will make a new tiny-PR to merge it in master!

PR #1640, opened 2021-01-07
Fix use-after-move for process executable path

dimitry-:
Test: builds

### DIFF
--- a/OrbitService/Process.cpp
+++ b/OrbitService/Process.cpp
@@ -90,9 +90,9 @@ ErrorMessageOr<Process> Process::FromPid(pid_t pid) {
 
   auto file_path_result = orbit_base::GetExecutablePath(pid);
   if (file_path_result) {
-    process.set_full_path(std::move(file_path_result.value()));
+    process.set_full_path(file_path_result.value());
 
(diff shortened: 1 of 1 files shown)


OrbitService/Process.cpp
```diff
@@ -90,9 +90,9 @@ ErrorMessageOr<Process> Process::FromPid(pid_t pid) {
 
   auto file_path_result = orbit_base::GetExecutablePath(pid);
   if (file_path_result) {
-    process.set_full_path(std::move(file_path_result.value()));
+    process.set_full_path(file_path_result.value());
 
-    const auto& elf_file = orbit_elf_utils::ElfFile::Create(file_path_result.value().string());
+    const auto& elf_file = orbit_elf_utils::ElfFile::Create(file_path_result.value());
```
dimitry-, line 95: hmm.. this shouldn't be a reference.

beckerhe (approved):
Thanks!

OrbitService/Process.cpp
beckerhe: That's not a problem due to the automatic reference-to-temporaty lifetime extensions. https://abseil.io/tips/107 for details

PR #1641, opened 2021-01-07
Add FindFunctionAbsoluteAddressByAddress to CaptureData.

florian-kuebler:
This adds a `GetFunctionBaseAddressByAddress` to `CaptureData` that
computes the functions's base address from an absolute address
inside a function based on either `ModuleData` or `LinuxAddressInfo`.
If both information are not available, it assumes the give address
is the base address.

This information was already computed internally in the
`SamplingDataPostProcessor`, but not made available. Instead of
just making this available in the `PostProcessedSamplingData`,
the new function goes to `CaptureData`, where sematically close
functions like `GetFunctionNameByAddress` are located.
Inside the `SamplingDataPostProcessor`, we now use the new function,
but still store the results in the map to avoid recomputations.

Bug: http://b/176970009
Test: Take a capture validate callstacks, load module and redo
      the validation.

### DIFF
--- a/OrbitClientModel/CaptureData.cpp
+++ b/OrbitClientModel/CaptureData.cpp
@@ -112,11 +112,12 @@ const std::string& CaptureData::GetFunctionNameByAddress(uint64_t absolute_addre
   return function_name;
 }
 
-// Find the start address of the function this address falls inside. Use the Function returned by
+// Find the start address of the function this address falls inside. Use the function returned by
 // FindFunctionByAddress, and when this fails (e.g., the module containing the function has not
 // been loaded) use (for now) the LinuxAddressInfo that is collected for every address in a
 // callstack.
-uint64_t CaptureData::GetFunctionBaseAddressByAddress(uint64_t absolute_address) const {
+std::optional<uint64_t> CaptureData::FindFunctionAbsoluteAddressByAddress(
+    uint64_t absolute_address) const {
   const FunctionInfo* function = FindFunctionByAddress(absolute_address, false);
   if (function != nullptr) {
     return GetAbsoluteAddress(*function);
@@ -125,7 +126,7 @@ uint64_t CaptureData::GetFunctionBaseAddressByAddress(uint64_t absolute_address)
   if (address_info != nullptr) {
     return absolute_address - address_info->offset_in_function();
   }
-  return absolute_address;
+  return std::nullopt;
 }
 
 const std::string& CaptureData::GetModulePathByAddress(uint64_t absolute_address) const {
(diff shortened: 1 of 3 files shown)


vickyliu-go4it (approved):
LGTM! Thanks so much for that!

OrbitClientModel/CaptureData.cpp
```diff
@@ -112,6 +112,22 @@ const std::string& CaptureData::GetFunctionNameByAddress(uint64_t absolute_addre
   return function_name;
 }
 
+// Find the start address of the function this address falls inside. Use the Function returned by
+// FindFunctionByAddress, and when this fails (e.g., the module containing the function has not
+// been loaded) use (for now) the LinuxAddressInfo that is collected for every address in a
+// callstack.
+uint64_t CaptureData::GetFunctionBaseAddressByAddress(uint64_t absolute_address) const {
```
dimitry-, line 119: This returns absolute_address if function was not found. I am not sure this is intuitive. Maybe use std::optional or ErrorMessageOr<> here? and let the caller decide what the value should be in this case using value_or() for example.

OrbitClientModel/CaptureData.cpp
```diff
@@ -112,6 +112,22 @@ const std::string& CaptureData::GetFunctionNameByAddress(uint64_t absolute_addre
...
+  const FunctionInfo* function = FindFunctionByAddress(absolute_address, false);
+  if (function != nullptr) {
+    return GetAbsoluteAddress(*function);
+  }
+  const LinuxAddressInfo* address_info = GetAddressInfo(absolute_address);
+  if (address_info != nullptr) {
+    return absolute_address - address_info->offset_in_function();
+  }
```
dimitry-, line 127: I wonder if we should translate address infos to functions after capture is done so that this function can be simplified.

OrbitClientModel/CaptureData.cpp
dimitry-: Maybe FindFunctionAbsoluteAddressByAddress?

OrbitClientModel/CaptureData.cpp
florian-kuebler: IIRC @antonrohr and I had the discussion at one point in time, but I don't remember what our arguments where.
I agree that the distinction of LinuxAddressData and FunctionInfo is not intuitive. This was the main motivation to keep this local to CaptureData where ever possible, and use the functions provided here. 
So, I'd argue, that this change is a step in this direction.

Also note, that as soon as we turn on frame-pointer-based unwinding, LinuxAddressInfo will be completely gone anyways.

OrbitClientModel/CaptureData.cpp
florian-kuebler: Done

dimitry- (approved):

dimitry-:
also probably update commit message with new function name

florian-kuebler:
Thanks, I'll update the message when doing the squash & merge (after the CI has completed).

PR #1642, opened 2021-01-07
Handle Vulkan-based Timers in GpuTrack.

florian-kuebler:
This adds support for Vulkan debug marker and command buffer timers
to the GpuTrack.
Command buffer timers will be drawn underneath their corresponding
hw execution timer, which requires to double the depth in the case
that we do have those timers.

For now, debug markers will be drawn on a separate track (named by
the timeline + "_marker"). For the future, we want to combine those
into a composed track.
Colors of the debug markers will either be picked up by the user's
choice, or be determined by the marker's text string.

Bug: http://b/176809754
Test: Compile -- Not really testable without producing the timers,
      which will be the next PR.
      
**Note: Requires #1632 to be in first**

### DIFF
--- a/OrbitGl/GpuTrack.cpp
+++ b/OrbitGl/GpuTrack.cpp
@@ -4,6 +4,8 @@
 
 #include "GpuTrack.h"
 
+#include <utility>
+
 #include "App.h"
 #include "GlCanvas.h"
 #include "OrbitBase/Profiling.h"
@@ -16,24 +18,35 @@ using orbit_client_protos::TimerInfo;
 constexpr const char* kSwQueueString = "sw queue";
 constexpr const char* kHwQueueString = "hw queue";
 constexpr const char* kHwExecutionString = "hw execution";
+constexpr const char* kCmdBufferString = "command buffer";
 
 namespace orbit_gl {
 
 std::string MapGpuTimelineToTrackLabel(std::string_view timeline) {
   std::string label;
   if (timeline.rfind("gfx", 0) == 0) {
+    if (timeline.find("_marker") != std::string::npos) {
+      return absl::StrFormat("Graphics queue debug markers (%s)", timeline);
+    }
     return absl::StrFormat("Graphics queue (%s)", timeline);
-  } else if (timeline.rfind("sdma", 0) == 0) {
-    return absl::StrFormat("Transfer queue (%s)", timeline);
-  } else if (timeline.rfind("comp", 0) == 0) {
-    return absl::StrFormat("Compute queue (%s)", timeline);
-  } else {
-    // On AMD, this should not happen and we don't support tracepoints for
-    // other GPUs (at the moment). We return the timeline to make sure we
-    // at least display something. When we add support for other GPU
-    // tracepoints, this needs to be changed.
-    return std::string(timeline);
   }
+  if (timeline.rfind("sdma", 0) == 0) {
+    if (timeline.find("_marker") != std::string::npos) {
+      return absl::StrFormat("Transfer queue debug markers (%s)", timeline);
+    }
+    return absl::StrFormat("Transfer queue (%s)", timeline);
+  }
+  if (timeline.rfind("comp", 0) == 0) {
+    if (timeline.find("_marker") != std::string::npos) {
+      return absl::StrFormat("Compute queue debug markers (%s)", timeline);
+    }
+    return absl::StrFormat("Compute queue (%s)", timeline);
+  }
+  // On AMD, this should not happen and we don't support tracepoints for
+  // other GPUs (at the moment). We return the timeline to make sure we
+  // at least display something. When we add support for other GPU
+  // tracepoints, this needs to be changed.
+  return std::string(timeline);
 }
 
 }  // namespace orbit_gl
@@ -48,6 +61,16 @@ GpuTrack::GpuTrack(TimeGraph* time_graph, StringManager* string_manager, uint64_
   // Gpu tracks are collapsed by default.
   collapse_toggle_->SetState(TriangleToggle::State::kCollapsed,
                              TriangleToggle::InitialStateUpdate::kReplaceInitialState);
+}
+
+void GpuTrack::OnTimer(const orbit_client_protos::TimerInfo& timer_info) {
+  // In case of having command buffer timers, we need to double the depth of the Gpu timers (as we
+  // are drawing the corresponding command buffer timers below them). Therefore, we watch out for
+  // those timers.
+  if (timer_info.type() == TimerInfo::kGpuCommandBuffer) {
+    has_vulkan_layer_command_buffer_timers_ = true;
+  }
+  TimerTrack::OnTimer(timer_info);
 }
 
 bool GpuTrack::IsTimerActive(const TimerInfo& timer_info) const {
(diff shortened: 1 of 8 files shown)


OrbitGl/GpuTrack.cpp
```diff
@@ -16,24 +18,35 @@ using orbit_client_protos::TimerInfo;
...
 constexpr const char* kHwQueueString = "hw queue";
 constexpr const char* kHwExecutionString = "hw execution";
+constexpr const char* kCmdBufferString = "command buffer";
 
 namespace orbit_gl {
 
 std::string MapGpuTimelineToTrackLabel(std::string_view timeline) {
   std::string label;
   if (timeline.rfind("gfx", 0) == 0) {
+    if (timeline.find("_marker") != std::string::npos) {
```
ronaldfw, line 28: Do we ourselves define the timeline strings in this case?

OrbitGl/GpuTrack.cpp
```diff
@@ -103,15 +137,28 @@ float GpuTrack::GetYFromTimer(const TimerInfo& timer_info) const {
...
+  CHECK(timer_info.type() == TimerInfo::kGpuActivity ||
+        timer_info.type() == TimerInfo::kGpuCommandBuffer);
+
+  if (has_vulkan_layer_command_buffer_timers_) {
+    adjusted_depth *= 2.f;
+  }
+
+  float gap_space = adjusted_depth * layout.GetSpaceBetweenGpuDepths();
+  if (timer_info.type() == TimerInfo::kGpuCommandBuffer) {
+    ++adjusted_depth;
```
ronaldfw, line 152: Nit: To my eye, this looks a bit funky, since this is a floating point value. I would write "adjusted_depth += 1.f". (There's no difference, so this is just a preference.)

ronaldfw: Maybe more actionable: It's not clear to me why this +1 is needed here. Can you add a comment to explain this a bit? Is this also correct in the collapsed state?

OrbitGl/GpuTrack.cpp
```diff
@@ -103,15 +137,28 @@ float GpuTrack::GetYFromTimer(const TimerInfo& timer_info) const {
...
+  return pos_[1] - layout.GetTextBoxHeight() * (adjusted_depth + 1.f) - gap_space;
 }
 
-// When track is collapsed, only draw "hardware execution" timers.
+// When track is collapsed, only draw "hardware execution" timers and "debug markers".
 bool GpuTrack::TimerFilter(const TimerInfo& timer_info) const {
   if (collapse_toggle_->IsCollapsed()) {
     std::string gpu_stage = string_manager_->Get(timer_info.user_data_key()).value_or("");
-    if (gpu_stage != kHwExecutionString) {
+    if (gpu_stage != kHwExecutionString && timer_info.type() != TimerInfo::kGpuDebugMarker) {
```
ronaldfw, line 161: I think this is more readable:

`return gpu_stage == kHwExecutionString || timer_info.type() == TimerInfo::kGpuDebugMarker;`

OrbitGl/GpuTrack.cpp
```diff
@@ -240,3 +300,36 @@ std::string GpuTrack::GetHwExecutionTooltip(const TimerInfo& timer_info) const {
       capture_data->GetThreadName(timer_info.thread_id()), timer_info.thread_id(),
       GetPrettyTime(TicksToDuration(timer_info.start(), timer_info.end())).c_str());
 }
+
+std::string GpuTrack::GetCommandBufferTooltip(
+    const orbit_client_protos::TimerInfo& timer_info) const {
+  return absl::StrFormat(
+      "<b>Command Buffer Execution</b><br/>"
+      "<i>At `vkBeginCommandBuffer` and `vkEndCommandBuffer` `vkCmdWriteTimestamp`s have been "
+      "inserted. The gpu timestamps get aligned with the corresponding submit's hardware "
```
ronaldfw, line 309: Although it seems many Vulkan users use "submit" that way, I don't think it's actually a noun. I think it should be "submission": "The GPU timestamps get aligned with the corresponding hardware execution of the submission." 

Also note that we write "GPU" with capitals in other places.

OrbitGl/GpuTrack.cpp
```diff
@@ -240,3 +300,36 @@ std::string GpuTrack::GetHwExecutionTooltip(const TimerInfo& timer_info) const {
...
+      GetPrettyTime(TicksToDuration(timer_info.start(), timer_info.end())).c_str());
+}
+
+std::string GpuTrack::GetDebugMarkerTooltip(
+    const orbit_client_protos::TimerInfo& timer_info) const {
+  std::string marker_text = string_manager_->Get(timer_info.user_data_key()).value_or("");
+  return absl::StrFormat(
+      "<b>Vulkan Debug Marker</b><br/>"
+      "<i>At the marker's begin and end `vkCmdWriteTimestamp`s have been "
+      "inserted. The gpu timestamps get aligned with the corresponding submit's hardware "
```
ronaldfw, line 325: Same here.

OrbitGl/GpuTrack.h
```diff
@@ -38,6 +38,12 @@ class GpuTrack : public TimerTrack {
   [[nodiscard]] float GetYFromTimer(
       const orbit_client_protos::TimerInfo& timer_info) const override;
 
+  void OnTimer(const orbit_client_protos::TimerInfo& timer_info) override;
+
+  [[nodiscard]] bool IsCollapsable() const override {
```
ronaldfw, line 43: `IsCollapsible()`

OrbitGl/GpuTrack.cpp
florian-kuebler: Yes exactly. We are going to add append the "_marker" to the original timeline string.

OrbitGl/GpuTrack.cpp
florian-kuebler: Done. Regarding the collapsed state: Yes that is handled correctly. We are not drawing command buffer timers, so we won't ask for the depth here. Explicitly tested this case, thanks for raising that question.

OrbitGl/GpuTrack.cpp
florian-kuebler: Yes, thanks!

OrbitGl/GpuTrack.cpp
florian-kuebler: Done.

OrbitGl/GpuTrack.cpp
florian-kuebler: Done.

OrbitGl/GpuTrack.h
florian-kuebler: Done.

florian-kuebler (commented):
Thanks for the review, I addressed the comments.

OrbitGl/GpuTrack.cpp
```diff
@@ -50,6 +63,16 @@ GpuTrack::GpuTrack(TimeGraph* time_graph, StringManager* string_manager, uint64_
                              TriangleToggle::InitialStateUpdate::kReplaceInitialState);
 }
 
+void GpuTrack::OnTimer(const orbit_client_protos::TimerInfo& timer_info) {
+  // In case of having command buffer timers, we need to double the depth of the Gpu timers (as we
```
ronaldfw, line 67: Gpu -> GPU for consistency with other comments

OrbitGl/GpuTrack.cpp
```diff
@@ -103,17 +141,35 @@ float GpuTrack::GetYFromTimer(const TimerInfo& timer_info) const {
...
   }
-  return pos_[1] - layout.GetTextBoxHeight() * (adjusted_depth + 1.f) -
-         adjusted_depth * layout.GetSpaceBetweenGpuDepths();
+  if (timer_info.type() == TimerInfo::kGpuDebugMarker) {
+    return pos_[1] - layout.GetTextBoxHeight() * (adjusted_depth + 1.f);
+  }
+  CHECK(timer_info.type() == TimerInfo::kGpuActivity ||
+        timer_info.type() == TimerInfo::kGpuCommandBuffer);
+
+  // Command buffer timers to be drawn underneath the matching "hw execution" timer, which as the
```
ronaldfw, line 150: ``
```suggestion
  // Command buffer timers are drawn underneath the matching "hw execution" timer, which as the
```

ronaldfw, line 150:
```suggestion
  // Command buffer timers to be drawn underneath the matching "hw execution" timer, which has the
```

ronaldfw (approved):
LGTM!

OrbitGl/GpuTrack.cpp
florian-kuebler: Done.

OrbitGl/GpuTrack.cpp
florian-kuebler: Done

florian-kuebler: Done.

ronaldfw (approved):
LGTM!

PR #1643, opened 2021-01-07
Indicate an ended target process in new beta UI

beckerhe:
Since the process list was removed in the new (beta) UI it is
particurlary hard to realize that the target process has ended on the
instance.

This commit is adding a somewhat hacky fix. It's checking the regularly
incoming process list for existence of the target process. If the target
process pid cannot be found the corner widget will color the process
name in orange - indicating that the process has died. Additionally it
will disable the Start-Capture button.

Bug: http://b/176960277

### DIFF
--- a/OrbitQt/orbitmainwindow.cpp
+++ b/OrbitQt/orbitmainwindow.cpp
@@ -118,6 +118,11 @@ constexpr int kHintFramePosX = 21;
 constexpr int kHintFramePosY = 47;
 constexpr int kHintFrameWidth = 140;
 constexpr int kHintFrameHeight = 45;
+
+const QString kTargetLabelDefaultStyleSheet = "#TargetLabel { color: %1; }";
+const QString kTargetLabelColorConnected = "#66BB6A";
+const QString kTargetLabelColorFileTarget = "#BDBDBD";
+const QString kTargetLabelColorTargetProcessDied = "orange";
 }  // namespace
 
 OrbitMainWindow::OrbitMainWindow(orbit_qt::TargetConfiguration target_configuration,
@@ -582,6 +587,14 @@ void OrbitMainWindow::UpdateCaptureStateDependentWidgets() {
   CaptureClient::State capture_state = app_->GetCaptureState();
   const bool is_capturing = capture_state != CaptureClient::State::kStopped;
 
+  // The detection mechanism is only implemented for the new UI, so we maintain
+  // the old behaviour for the old UI and assume the target process is always running.
+  // TODO (170468590): [ui beta] This can be renamed to is_target_process_running when
+  // the feature flag gets removed.
+  const bool assume_target_process_is_running =
+      (!absl::GetFlag(FLAGS_enable_ui_beta)) ||
+      target_process_state_ == TargetProcessState::kRunning;
+
   if (!absl::GetFlag(FLAGS_enable_ui_beta)) {
     set_tab_enabled(ui->HomeTab, true);
     ui->HomeTab->setEnabled(!is_capturing);
@@ -596,8 +609,9 @@ void OrbitMainWindow::UpdateCaptureStateDependentWidgets() {
   set_tab_enabled(ui->selectionTopDownTab, has_selection);
   set_tab_enabled(ui->selectionBottomUpTab, has_selection);
 
-  ui->actionToggle_Capture->setEnabled(capture_state == CaptureClient::State::kStarted ||
-                                       capture_state == CaptureClient::State::kStopped);
+  ui->actionToggle_Capture->setEnabled(
+      capture_state == CaptureClient::State::kStarted ||
+      (capture_state == CaptureClient::State::kStopped && assume_target_process_is_running));
   ui->actionToggle_Capture->setIcon(is_capturing ? icon_stop_capture_ : icon_start_capture_);
   ui->actionClear_Capture->setEnabled(!is_capturing && has_data);
   ui->actionOpen_Capture->setEnabled(!is_capturing);
@@ -1165,11 +1179,19 @@ void OrbitMainWindow::SetTarget(const orbit_qt::StadiaTarget& target) {
   app_->SetGrpcChannel(connection->GetGrpcChannel());
   app_->SetProcessManager(target.GetProcessManager());
   app_->SetTargetProcess(target.GetProcess());
-  // TODO (b/173011168): Color should not be defined inline. Replace with defined "good color"
-  // (green).
-  target_label_->setStyleSheet("#TargetLabel { color: #66BB6A; }");
+
+  target_label_->setStyleSheet(kTargetLabelDefaultStyleSheet.arg(kTargetLabelColorConnected));
   target_label_->setText(QString::fromStdString(target.GetProcess()->name()) + " @ " +
                          target.GetConnection()->GetInstance().display_name);
+
+  using ProcessInfo = orbit_grpc_protos::ProcessInfo;
+  target.GetProcessManager()->SetProcessListUpdateListener([&](std::vector<ProcessInfo> processes) {
+    // This lambda is called from a background-thread, so we use QMetaObject::invokeMethod
+    // to execute our logic on the main thread.
+    QMetaObject::invokeMethod(this, [&, processes = std::move(processes)]() {
+      OnProcessListUpdated(std::move(processes));
+    });
+  });
 }
 
 void OrbitMainWindow::SetTarget(const orbit_qt::LocalTarget& target) {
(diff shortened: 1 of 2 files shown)


OrbitQt/orbitmainwindow.cpp
```diff
@@ -582,6 +582,13 @@ void OrbitMainWindow::UpdateCaptureStateDependentWidgets() {
   CaptureClient::State capture_state = app_->GetCaptureState();
   const bool is_capturing = capture_state != CaptureClient::State::kStopped;
 
+  // The detection mechanism is only implemented for the new UI, so we maintain
+  // the old behaviour for the old UI and assume the target process is always running.
+  // This can be renamed to is_target_process_running when the feature flag gets removed.
```
reichlfl, line 587: IIRC we have a bug id mentioned in all other places concerning removal of the feature flag, can you add this here as well with a TODO?

OrbitQt/orbitmainwindow.cpp
```diff
@@ -1165,11 +1173,38 @@ void OrbitMainWindow::SetTarget(const orbit_qt::StadiaTarget& target) {
...
   // TODO (b/173011168): Color should not be defined inline. Replace with defined "good color"
   // (green).
-  target_label_->setStyleSheet("#TargetLabel { color: #66BB6A; }");
+  static const QString kTargetLabelDefaultStyleSheet = "#TargetLabel { color: #66BB6A; }";
+  target_label_->setStyleSheet(kTargetLabelDefaultStyleSheet);
   target_label_->setText(QString::fromStdString(target.GetProcess()->name()) + " @ " +
                          target.GetConnection()->GetInstance().display_name);
+
+  using ProcessInfo = orbit_grpc_protos::ProcessInfo;
+  target.GetProcessManager()->SetProcessListUpdateListener([&](std::vector<ProcessInfo> processes) {
```
reichlfl, line 1185: For readability and to allow easier modification later, I'd prefer to provide a `OrbitMainWindow::OnProcessListUpdated` and call this inside `invokeMethod`.

OrbitQt/orbitmainwindow.cpp
```diff
@@ -1165,11 +1173,38 @@ void OrbitMainWindow::SetTarget(const orbit_qt::StadiaTarget& target) {
...
+    // This lambda is called from a background-thread, so we use QMetaObject::invokeMethod
+    // to execute our logic on the main thread.
+    QMetaObject::invokeMethod(this, [&, processes = std::move(processes)]() {
+      const auto is_current_process = [this](const auto& process) {
+        const auto target_process = app_->GetTargetProcess();
```
reichlfl, line 1190: I have been scolded for using `auto` too much, so I'll do the same here :-) Same for the `auto`s following below
https://google.github.io/styleguide/cppguide.html#Type_deduction

OrbitQt/orbitmainwindow.cpp
```diff
@@ -1165,11 +1173,38 @@ void OrbitMainWindow::SetTarget(const orbit_qt::StadiaTarget& target) {
...
+        return target_process != nullptr && process.pid() == app_->GetTargetProcess()->pid();
+      };
+      const auto current_process =
+          std::find_if(processes.begin(), processes.end(), is_current_process);
+
+      if (const auto process_died = current_process == processes.end(); process_died) {
```
reichlfl, line 1196: I don't see the benefit of the `process_died` assignment here since it's not used inside the if-block. Is it purely to convey intent of `current_process == processes.end()`? For me, it's more confusing then helping because I've never used this if-syntax in the past, but maybe that's just me.

In any case, `auto` instead of `bool` is certainly not helping ;-)

reichlfl (commented):
Thanks for fixing this so quickly, as discussed I think it's perfectly fine to be somewhat of a workaround for now. Few comments mostly regarding coding style, functionality looks fine to me!

OrbitQt/orbitmainwindow.cpp
beckerhe: Sure. I can add that.

My reasoning for not doing that: Whenever the code involves the feature flag `FLAGS_enable_ui_beta`, it's an even stronger indication that this has to be touched when the feature flag is removed. The compiler would complain if it was forgotten.

OrbitQt/orbitmainwindow.cpp
beckerhe: Makes sense. I debated that with myself as well!

OrbitQt/orbitmainwindow.cpp
beckerhe: Yes, it's merely to explain that `current_process == processes.end()` means that the process died. But I can write that differently.

OrbitQt/orbitmainwindow.cpp
beckerhe: Will change that! :+1:

OrbitQt/orbitmainwindow.cpp
beckerhe: done.

OrbitQt/orbitmainwindow.cpp
beckerhe: done. I just like auto too much. :see_no_evil: Time to move on to C++20!

OrbitQt/orbitmainwindow.cpp
beckerhe: Changed!

OrbitQt/orbitmainwindow.cpp
beckerhe: Done.

beckerhe:
Thanks Florian. I addressed your comments.

I had to make some changes to the stylesheet handling when I factored out this code into its own function. I'm not sure if I like it that way. Can you have another look?

I also realized, that we can enable the detection also for the local target. So I added that as well.

OrbitQt/orbitmainwindow.cpp
```diff
@@ -1242,3 +1236,24 @@ void OrbitMainWindow::SetupGrpcAndProcessManager(std::string grpc_server_address
   app_->SetGrpcChannel(std::move(grpc_channel));
   app_->SetProcessManager(process_manager_.get());
 }
+
+void OrbitMainWindow::OnProcessListUpdated(std::vector<orbit_grpc_protos::ProcessInfo> processes) {
+  const auto is_current_process = [this](const auto& process) {
+    const ProcessData* const target_process = app_->GetTargetProcess();
+    return target_process != nullptr && process.pid() == app_->GetTargetProcess()->pid();
+  };
+  const auto current_process = std::find_if(processes.begin(), processes.end(), is_current_process);
```
reichlfl, line 1245: Another `auto` that's probably not conforming to the style guide

OrbitQt/orbitmainwindow.h
```diff
@@ -95,6 +95,13 @@ class OrbitMainWindow : public QMainWindow {
   // TODO(170468590): [ui beta] When out of ui beta, this can return TargetConfiguration (without
   // std::optional)
   std::optional<orbit_qt::TargetConfiguration> ClearTargetConfiguration() {
```
reichlfl, line 97: I've just realized this is in the header file - suggest to move the implementation to the cpp file.

reichlfl (approved):
> Thanks Florian. I addressed your comments.
> 
> I had to make some changes to the stylesheet handling when I factored out this code into its own function. I'm not sure if I like it that way. Can you have another look?
> 
> I also realized, that we can enable the detection also for the local target. So I added that as well.

Thanks, I've only got small comments, but overall LGTM!

I'm not sure what exactly you're referring to here, you mean replacing the text with constants and pulling them out of the methods? It's still not ideal that we always define our colors locally (by locally, I don't mean only as a local variable in a function, but also across several cpp files and not in one central place), but it's better than before. We're doing this all across Orbit, so it's nothing we should improve in this PR now.

OrbitQt/orbitmainwindow.cpp
beckerhe: I'm not sure if `const std::vector<ProcessInfo>::const_iterator` helps readability here. Actually I believe it rather harms readability. WDYT?

reichlfl: Oh sorry, I misread and thought this was supposed to be `ProcessData*`. My bad, ignore my comment.

OrbitQt/orbitmainwindow.h
beckerhe: Yep. Good point. Done.

OrbitQt/orbitmainwindow.cpp
beckerhe: No worries at all. Glad that we agree on this. ;-)

beckerhe: and FWIW. `std::vector<ProcessInfo>::const_iterator` might be an alias for `const ProcessInfo*` depending on the implemenation of the standard library. But we can't assume that.

PR #1644, opened 2021-01-08
Fix crash in OrbitTableModel

beckerhe:
OrbitMainWindow's widgets are destructed after destructing
OrbitMainWindow. Most widgets like OrbitSamplingReport require resources
that are managed by OrbitMainWindow.

During the deconstruction phase of OrbitMainWindow's widgets,
deconstructions of other widgets trigger resize events in not-yet
deconstructed widgets. These resize events might trigger data access to
resources which have been deconstructed long before.

This is solved by removing access to OrbitMainWindow-managed resources
before starting to deconstruct widgets.

So this commit adds a Deinitalize method to OrbitSamplingReport which
deinitializes all the internal widgets and data views.

Furthermore there is a second commit which removes dead code.


dpallotti (approved):
Thanks!

PR #1646, opened 2021-01-08
Add/Remove FrameTracks properly while capturing

karupayun:
Frame tracks are a special track because could be added or erased while
capturing. In this PR we are solving a few issues about inconsistent
information when adding or removing a frame track while capturing.
Basically, we updated:

 - FrameTrackOnlineProcessor in App (info while capturing)

 - Visible Tracks in TrackManager after removing Frame Tracks.

There are still an issue about multithreading problems (more information
in http://b/176959931).

Test: Start a capture, enable/disable Frame Tracks/ Stop a capture several times.

### DIFF
--- a/OrbitGl/TrackManager.cpp
+++ b/OrbitGl/TrackManager.cpp
@@ -82,8 +82,8 @@ std::vector<FrameTrack*> TrackManager::GetFrameTracks() const {
 void TrackManager::SortTracks() {
   if (!app_->IsCapturing() && !sorted_tracks_.empty() && !sorting_invalidated_) return;
 
-  // Reorder threads once every second when capturing
-  if (!app_->IsCapturing() || last_thread_reorder_.ElapsedMillis() > 1000.0) {
+  // Reorder threads if sorting isn't valid or once per second when capturing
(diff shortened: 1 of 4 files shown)


ronaldfw (approved):
LGTM! 

Should we add some asserts to make sure that methods like RefreshFrameTracks() are only called from the main thread?

OrbitGl/TrackManager.cpp
```diff
@@ -321,6 +321,8 @@ void TrackManager::AddTrack(std::shared_ptr<Track> track) {
 void TrackManager::RemoveFrameTrack(uint64_t function_address) {
   frame_tracks_.erase(function_address);
   sorting_invalidated_ = true;
+  // We need to do SortTracks again to have visible_tracks updated
```
florian-kuebler, line 324: Nit: There is an underscore missing in `visible_tracks_`.

florian-kuebler (approved):
LGTM. Thanks for fixing this. Regarding the checks that the calls are from the main thread. That might be a good idea. You can have a look at `DataManager`, that already has such kind of checks.
But for me it would be also fine to not have them.

karupayun:
Added two CHECK.

PR #1647, opened 2021-01-08
Apply iwyu suggestions to the code base

beckerhe:
This applies all suggestions by include-what-you-use to the code base.

Unfortunately it resulted in a single commit and not in
per-module-commits. Removed includes from one module often results in
compilation errors in other modules which makes it very hard to split
the changes on a module-level.

Let me know if you think it's worthwhile to split this up. Probably it's a lot of work
and so far I don't see a good enough reason to do that.

### DIFF
--- a/ElfUtils/include/ElfUtils/LinuxMap.h
+++ b/ElfUtils/include/ElfUtils/LinuxMap.h
@@ -4,6 +4,11 @@
 
 #ifndef ELF_UTILS_LINUX_MAP_H_
 #define ELF_UTILS_LINUX_MAP_H_
+
+#include <stdint.h>
+
+#include <string_view>
+
 #if defined(__linux)
 
 #include <filesystem>
--- a/OrbitBase/Logging.cpp
+++ b/OrbitBase/Logging.cpp
@@ -3,6 +3,8 @@
 // found in the LICENSE file.
 
 #include <OrbitBase/Logging.h>
+#include <absl/base/const_init.h>
+#include <absl/synchronization/mutex.h>
 
 static absl::Mutex log_file_mutex(absl::kConstInit);
 std::ofstream log_file;
--- a/OrbitBase/OrbitApiTest.cpp
+++ b/OrbitBase/OrbitApiTest.cpp
@@ -3,6 +3,10 @@
 // found in the LICENSE file.
 
 #include <gtest/gtest.h>
+#include <stdint.h>
+#include <string.h>
+
+#include <string>
 
 #include "OrbitBase/Tracing.h"
 
--- a/OrbitBase/ThreadUtilsTest.cpp
+++ b/OrbitBase/ThreadUtilsTest.cpp
@@ -3,16 +3,22 @@
 // found in the LICENSE file.
 
 #include <absl/synchronization/mutex.h>
-#include <gmock/gmock-matchers.h>
 #include <gtest/gtest.h>
 
-#include <chrono>
+#ifdef _WIN32
+// No special windows header needed
+#else
+#include <pthread.h>
+#endif
+
+#include <ctime>
+#include <string>
 #include <thread>
 
 #include "OrbitBase/ThreadUtils.h"
 
 TEST(ThreadUtils, GetCurrentThreadId) {
-#if _WIN32
+#ifdef _WIN32
   uint32_t current_tid = orbit_base::GetCurrentThreadId();
   uint32_t worker_tid = 0;
 #else
--- a/OrbitCaptureClient/CaptureEventProcessorTest.cpp
+++ b/OrbitCaptureClient/CaptureEventProcessorTest.cpp
@@ -1,12 +1,28 @@
 // Copyright (c) 2020 The Orbit Authors. All rights reserved.
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
+
+#include <absl/container/flat_hash_map.h>
+#include <gmock/gmock.h>
+#include <gtest/gtest.h>
+#include <stddef.h>
+#include <stdint.h>
+
+#include <algorithm>
+#include <string>
+#include <vector>
+
+#include "OrbitBase/Result.h"
 #include "OrbitCaptureClient/CaptureEventProcessor.h"
 #include "OrbitCaptureClient/CaptureListener.h"
+#include "OrbitClientData/Callstack.h"
+#include "OrbitClientData/ProcessData.h"
+#include "OrbitClientData/TracepointCustom.h"
+#include "OrbitClientData/UserDefinedCaptureData.h"
+#include "capture.pb.h"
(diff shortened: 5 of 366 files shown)


beckerhe:
The license_header check fails as usual because of `OrbitQt/orbitcodeeditor.{cpp,h}`.

OrbitBase/OrbitApiTest.cpp
```diff
@@ -3,6 +3,10 @@
 // found in the LICENSE file.
 
 #include <gtest/gtest.h>
+#include <stdint.h>
+#include <string.h>
```
florian-kuebler, line 7: Do we really need <string.h> here?

beckerhe: strlen is used below

OrbitBase/Logging.cpp
```diff
@@ -3,6 +3,8 @@
 // found in the LICENSE file.
 
 #include <OrbitBase/Logging.h>
```
dpallotti, line 5: Just mentioning that ideally these would be changed to using quotes.

OrbitCaptureClient/CaptureEventProcessorTest.cpp
```diff
@@ -1,12 +1,27 @@
 // Copyright (c) 2020 The Orbit Authors. All rights reserved.
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
+#include <absl/container/flat_hash_map.h>
```
dpallotti, line 4:
```suggestion

#include <absl/container/flat_hash_map.h>
```

OrbitTest/OrbitTest.h
```diff
@@ -3,6 +3,8 @@
 // found in the LICENSE file.
 
 #pragma once
+#include <stdint.h>
```
dpallotti, line 6:
```suggestion

#include <stdint.h>
```

dpallotti (approved):
Great to see this happen!

OrbitSsh/Socket.cpp
```diff
@@ -4,6 +4,17 @@
 
 #include "OrbitSsh/Socket.h"
 
+#ifndef _WIN32
```
florian-kuebler, line 7: Again, I'd switch the cases and have a `ifdef` instead.

florian-kuebler (approved):
Thanks so much. This is really great!

OrbitBase/Logging.cpp
beckerhe: Yes. I can easily influence that for new includes. But as far as I can see iwyu does not support that for existing and working includes. I guess we have to find another way to fix that.

OrbitBase/ThreadUtilsTest.cpp
```diff
@@ -3,10 +3,14 @@
 // found in the LICENSE file.
 
 #include <absl/synchronization/mutex.h>
-#include <gmock/gmock-matchers.h>
 #include <gtest/gtest.h>
 
-#include <chrono>
+#ifndef _WIN32
```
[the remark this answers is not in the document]
beckerhe: Hmmm. This file has two flavors of that:

1. Line 19 and line 72:
```
#if _WIN32
#else
#endif
```

2. Line 54:
```
#ifdef __linux
#endif
```

Of course, this should be cleaned up and made consistent. But I guess not in this PR.

My thoughts on this:
1. `#if _WIN32` is wrong and should definitely be changed to `#ifdef _WIN32` (or something else).
2. I don't like having two separate branches: `#ifdef _WIN32 ... #endif #ifdef __linux ... #endif`. In that case the code might compile when neither macro is set which is a confusing silent failure. This is especially important since every included header could potentially change an existing macro. Silent bugs can be minimized by using fewer macros and fail if something is unexpected.
3. The header that I include here is technically not a linux-only header, rather a Unix/Posix header. So they would work on any other Unix-based OS like BSD or OSX. That's why I like `#ifndef _WIN32` here. In addition we mostly use `_WIN32` in the codebase (A quick search resulted in 56 usage of _WIN32/WIN32 and 8 usages of __linux/__linux__).
4. It might be confusing that `#ifndef _WIN32` means Linux. I guess that's also the reason why you added that comment. And that might be a strong argument for not doing it.

What are your thoughts on this?

OrbitSsh/Socket.cpp
beckerhe: Good point. Done.

OrbitTest/OrbitTest.h
beckerhe: Yes. thanks. done.

ElfUtils/include/ElfUtils/LinuxMap.h
```diff
@@ -4,6 +4,9 @@
 
 #ifndef ELF_UTILS_LINUX_MAP_H_
 #define ELF_UTILS_LINUX_MAP_H_
+#include <stdint.h>
+
+#include <string_view>
```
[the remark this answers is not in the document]
beckerhe: Good point. Done. Thanks.

OrbitCaptureClient/CaptureEventProcessorTest.cpp
beckerhe: Done.

OrbitBase/OrbitApiTest.cpp
florian-kuebler: Ah thanks.

OrbitBase/ThreadUtilsTest.cpp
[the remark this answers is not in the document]
florian-kuebler: I think I share all of your arguments/thoughts.
I was not aware of point 3. In this case I'd guess I'd go for `#ifdef _WIN32 ... #else ... #endifdef`. I would hope that in this case it is clearer that "Not Windows" means Linux in our case. Further, I'd try to group those together.
This file was the most extreme example, so it would be nice, if you can address this for this file (and maybe the second one, I've commented on).

It doesn't need to be perfect, as it is already a very big step. But it would be cool to see this here a bit more consistent.

OrbitBase/Logging.cpp
dpallotti: Yep, I know (I remember we talked about this), I guess we can just do it manually eventually, there shouldn't be many left.

OrbitBase/ThreadUtilsTest.cpp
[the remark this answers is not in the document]
beckerhe: Sure, makes sense. I also like the idea of having a consistent if-else-construct.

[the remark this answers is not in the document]
beckerhe: Done.

beckerhe:
I will rebase now and if the CI agrees I'm going to merge.

PR #1648, opened 2021-01-08
Prevent Vulkan queue submissions from being sent out of order.

florian-kuebler:
On every vkQueuePresentKHR, we try to retrieve timestamp results
for all unprocessed submissions. It may happen that on one `present`
some submissions are not yet ready, and on the next `present` we
have multiple ones to process.

Before this change, this could lead to us sending submissions of
a single queue out of order, as we might have picked the latest
submission first to check.

It would complicate the client's processing code, if it is not
guaranteed that submissions of a single queue are retrieved in
order.

Thus, this sorts the submissions by their CPU timestamps after
they are retrieved.

Test: Run the new test

### DIFF
--- a/OrbitVulkanLayer/SubmissionTracker.h
+++ b/OrbitVulkanLayer/SubmissionTracker.h
@@ -471,6 +471,14 @@ class SubmissionTracker : public VulkanLayerProducer::CaptureStatusListener {
     std::vector<QueueSubmission> completed_submissions =
         PullCompletedSubmissions(device, query_pool);
 
+    // We sort the completed submissions by timestamp to avoid sending the capture events out of
+    // order.
+    std::sort(completed_submissions.begin(), completed_submissions.end(),
+              [](const QueueSubmission& left, const QueueSubmission& right) {
+                return left.meta_information.pre_submission_cpu_timestamp <
+                       right.meta_information.pre_submission_cpu_timestamp;
+              });
+
     if (completed_submissions.empty()) {
       return;
     }
--- a/OrbitVulkanLayer/SubmissionTracker.h
+++ b/OrbitVulkanLayer/SubmissionTracker.h
@@ -471,14 +471,6 @@ class SubmissionTracker : public VulkanLayerProducer::CaptureStatusListener {
     std::vector<QueueSubmission> completed_submissions =
         PullCompletedSubmissions(device, query_pool);
 
-    // We sort the completed submissions by timestamp to avoid sending the capture events out of
-    // order.
-    std::sort(completed_submissions.begin(), completed_submissions.end(),
-              [](const QueueSubmission& lhs, const QueueSubmission& rhs) {
-                return lhs.meta_information.pre_submission_cpu_timestamp <
-                       rhs.meta_information.pre_submission_cpu_timestamp;
-              });
-
     if (completed_submissions.empty()) {
       return;
     }
@@ -701,6 +693,14 @@ class SubmissionTracker : public VulkanLayerProducer::CaptureStatusListener {
         }
       }
     }
+
+    // We sort the completed submissions by timestamp to avoid sending the capture events out of
+    // order.
+    std::sort(completed_submissions.begin(), completed_submissions.end(),
+              [](const QueueSubmission& lhs, const QueueSubmission& rhs) {
+                return lhs.meta_information.pre_submission_cpu_timestamp <
+                       rhs.meta_information.pre_submission_cpu_timestamp;
+              });
 
     return completed_submissions;
   }
--- a/OrbitVulkanLayer/SubmissionTracker.h
+++ b/OrbitVulkanLayer/SubmissionTracker.h
@@ -474,9 +474,9 @@ class SubmissionTracker : public VulkanLayerProducer::CaptureStatusListener {
     // We sort the completed submissions by timestamp to avoid sending the capture events out of
     // order.
     std::sort(completed_submissions.begin(), completed_submissions.end(),
-              [](const QueueSubmission& left, const QueueSubmission& right) {
-                return left.meta_information.pre_submission_cpu_timestamp <
-                       right.meta_information.pre_submission_cpu_timestamp;
+              [](const QueueSubmission& lhs, const QueueSubmission& rhs) {
+                return lhs.meta_information.pre_submission_cpu_timestamp <
+                       rhs.meta_information.pre_submission_cpu_timestamp;
               });
 
     if (completed_submissions.empty()) {
--- a/OrbitVulkanLayer/SubmissionTrackerTest.cpp
+++ b/OrbitVulkanLayer/SubmissionTrackerTest.cpp
@@ -667,22 +667,22 @@ TEST_F(SubmissionTrackerTest, MultipleSubmissionsWontBeOutOfOrder) {
       .WillRepeatedly(Return(mock_get_query_pool_results_function_all_ready_));
 
   std::vector<uint32_t> actual_reset_slots;
-  auto mock_reset_query_slots = [&actual_reset_slots](VkDevice /*device*/,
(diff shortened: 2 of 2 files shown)


OrbitVulkanLayer/SubmissionTracker.h
```diff
@@ -471,6 +471,14 @@ class SubmissionTracker : public VulkanLayerProducer::CaptureStatusListener {
     std::vector<QueueSubmission> completed_submissions =
         PullCompletedSubmissions(device, query_pool);
 
+    // We sort the completed submissions by timestamp to avoid sending the capture events out of
+    // order.
+    std::sort(completed_submissions.begin(), completed_submissions.end(),
+              [](const QueueSubmission& left, const QueueSubmission& right) {
```
dpallotti, line 477: Little note that common names for these are "lhs" and "rhs" (left-hand side, right-hand side), also with the advantage of being the same length.

OrbitVulkanLayer/SubmissionTrackerTest.cpp
```diff
@@ -660,6 +660,67 @@ TEST_F(SubmissionTrackerTest, StartCaptureDuringSubmissionWontWriteData) {
...
 }
 
+TEST_F(SubmissionTrackerTest, MultipleSubmissionsWontBeOutOfOrder) {
+  ExpectFourNextReadyQuerySlotCalls();
+  EXPECT_CALL(dispatch_table_, GetQueryPoolResults)
+      .WillOnce(Return(mock_get_query_pool_results_function_not_ready_))
+      .WillRepeatedly(Return(mock_get_query_pool_results_function_all_ready_));
+
+  std::vector<uint32_t> actual_reset_slots;
+  auto mock_reset_query_slots = [&actual_reset_slots](VkDevice /*device*/,
```
dpallotti, line 670: Maybe these lambdas are fakes rather than mocks?

OrbitVulkanLayer/SubmissionTrackerTest.cpp
```diff
@@ -660,6 +660,67 @@ TEST_F(SubmissionTrackerTest, StartCaptureDuringSubmissionWontWriteData) {
...
+  tracker_.ResetCommandBuffer(command_buffer_);
+  tracker_.MarkCommandBufferBegin(command_buffer_);
+  tracker_.MarkCommandBufferEnd(command_buffer_);
+  std::optional<QueueSubmission> queue_submission_optional_2 =
+      tracker_.PersistCommandBuffersOnSubmit(1, &submit_info_);
+  tracker_.PersistDebugMarkersOnSubmit(queue_, 1, &submit_info_, queue_submission_optional_2);
+  // Now the results are ready for all submissions.
+  tracker_.CompleteSubmits(device_);
+
+  EXPECT_EQ(actual_capture_events.size(), 2);
```
dpallotti, line 709: I think that this should be `ASSERT` otherwise accessing `actual_capture_events` in the next two lines could fail.

OrbitVulkanLayer/SubmissionTrackerTest.cpp
```diff
@@ -660,6 +660,67 @@ TEST_F(SubmissionTrackerTest, StartCaptureDuringSubmissionWontWriteData) {
...
+  tracker_.CompleteSubmits(device_);
+
+  // Submission #2:
+  tracker_.ResetCommandBuffer(command_buffer_);
+  tracker_.MarkCommandBufferBegin(command_buffer_);
+  tracker_.MarkCommandBufferEnd(command_buffer_);
+  std::optional<QueueSubmission> queue_submission_optional_2 =
+      tracker_.PersistCommandBuffersOnSubmit(1, &submit_info_);
+  tracker_.PersistDebugMarkersOnSubmit(queue_, 1, &submit_info_, queue_submission_optional_2);
+  // Now the results are ready for all submissions.
```
dpallotti, line 706: What causes all submissions here to be ready compared to the `Submission #1` block?

OrbitVulkanLayer/SubmissionTracker.h
```diff
@@ -471,6 +471,14 @@ class SubmissionTracker : public VulkanLayerProducer::CaptureStatusListener {
...
+                return left.meta_information.pre_submission_cpu_timestamp <
+                       right.meta_information.pre_submission_cpu_timestamp;
+              });
+
     if (completed_submissions.empty()) {
```
dpallotti, line 482: One doubt, most likely because I'm missing the full picture: Are all submissions completed in order of their cpu timestamp? Or are all completed submissions obtained only if all the submissions for a `present` have completed? Otherwise, if not all submissions from one `present` are completed, couldn't it happen that we process some that are already completed now and in the following call to this method we process other ones that were previously not ready, but that came before the already processed ones?

florian-kuebler: Submissions on the *same queue*, are guaranteed to be completed in the order as the `vkQueueSubmit` happens on the cpu, i.e. based on this submission_cpu_timestamp. The `present` here is just a random point in time, to check if the submissions are already completed. So if I retrieve a submission here, all the submissions on that queue that happend before, were either already ready in the previous present (and this already sent out), or are part of the result of `PullCompletedSubmission`, in which case the `sort` ensures that we send them out in the correct order.

OrbitVulkanLayer/SubmissionTrackerTest.cpp
florian-kuebler: I guess here both would be fine (as they are also used to validate the result data). But I renamed them.

OrbitVulkanLayer/SubmissionTrackerTest.cpp
florian-kuebler: Line 665:   
```
EXPECT_CALL(dispatch_table_, GetQueryPoolResults)
      .WillOnce(Return(mock_get_query_pool_results_function_not_ready_))
      .WillRepeatedly(Return(mock_get_query_pool_results_function_all_ready_));
```

In the first submission, we try to get the results for the one command buffer, that is part of the submission, and will note that it is not yet ready. Afterwards, every call will succeed.

OrbitVulkanLayer/SubmissionTrackerTest.cpp
florian-kuebler: Right! Thanks.

florian-kuebler: The ones below, can be an `EXPECT` though...

OrbitVulkanLayer/SubmissionTracker.h
dpallotti: Ah alright, thanks for the explanation. So why does `PullCompletedSubmission` return them not in order? And wouldn't it maybe make more sense for `PullCompletedSubmissions` to perform the `sort` instead of returning an undefined order?

florian-kuebler: Yeah you are right. Maybe this is better.

florian-kuebler: Done.

OrbitVulkanLayer/SubmissionTracker.h
florian-kuebler: Done. Thanks.

OrbitVulkanLayer/SubmissionTrackerTest.cpp
dpallotti: Ah, I would suggest adding a quick comment for that (either here or where the `EXPECT_CALL` is).

dpallotti (approved):

OrbitVulkanLayer/SubmissionTrackerTest.cpp
florian-kuebler: Added a comment.

PR #1649, opened 2021-01-08
Handle Vulkan related timers in TimeGraph.

florian-kuebler:
This adds support to the new Vulkan related timer types (command
buffers and debug markers) to `TimeGraph`, which will forward them
to the `GpuTrack` that already has support for those (#1642).

Test: Compile -- Not really testable without producing the timers.
Bug: http://b/176809754

### DIFF
--- a/OrbitGl/TimeGraph.cpp
+++ b/OrbitGl/TimeGraph.cpp
@@ -258,6 +258,18 @@ void TimeGraph::ProcessTimer(const TimerInfo& timer_info, const FunctionInfo* fu
       track->OnTimer(timer_info);
       break;
     }
+    case TimerInfo::kGpuCommandBuffer: {
+      uint64_t timeline_hash = timer_info.timeline_hash();
+      GpuTrack* track = track_manager_->GetOrCreateGpuTrack(timeline_hash);
+      track->OnTimer(timer_info);
+      break;
+    }
+    case TimerInfo::kGpuDebugMarker: {
+      uint64_t timeline_hash = timer_info.timeline_hash();
+      GpuTrack* track = track_manager_->GetOrCreateGpuTrack(timeline_hash);
+      track->OnTimer(timer_info);
+      break;
+    }
     case TimerInfo::kFrame: {
       if (function == nullptr) {
         break;
--- a/OrbitGl/TimeGraph.cpp
+++ b/OrbitGl/TimeGraph.cpp
@@ -252,18 +252,9 @@ void TimeGraph::ProcessTimer(const TimerInfo& timer_info, const FunctionInfo* fu
   // TODO (b/175869409): Change the way to create and get the tracks. Move this part to
   // TrackManager.
   switch (timer_info.type()) {
-    case TimerInfo::kGpuActivity: {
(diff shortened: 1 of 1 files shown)


pierricgimmig (approved):
LGTM. However, is the code duplication intentional? The code is the same for kGpuXYZ types, should we consolidate into a single code block?

florian-kuebler:
I was thinking about the code duplication. We very recently switched to the "switch" statement here, to ensure that we not miss some cases/handle them as intended (before there was a lot going on in an "else" case). Switches seems to be much cleaner, and omitting the "break" in those switches, to handle those GPU cases equal, feels kind of hacky and again a bit error prone. 
I'd suggest to keep it as it is, especially as the code duplication is not large. Further, AFAIK @karupayun is going to rewrite this part of the code anyways.

karupayun:
And writing the 3 cases together?
case TimerInfo::kGpuActivity:
case TimerInfo::kGpuCommandBuffer:
case TimerInfo::kGpuDebugMarker:
{
    body
}

karupayun:
I'd prefer to have the 3 cases together. Seems more probably that we have to modify the 3 of them together (which now seems more error prone) than one of them separately.

case TimerInfo::kGpuActivity:
case TimerInfo::kGpuCommandBuffer:
case TimerInfo::kGpuDebugMarker:
{
      uint64_t timeline_hash = timer_info.timeline_hash();
      GpuTrack* track = track_manager_->GetOrCreateGpuTrack(timeline_hash);
      track->OnTimer(timer_info);
      break;
}

florian-kuebler:
Yeah, that would do the trick. But again feels a bit error-prone, as it is easy to accidentally insert a case in between. But if you both prefer that, I'd be fine with going for that. Miss Scala, where it would be `case kGpuActivity | kGpuCommandBuffer | kGpuDebugMarker: `....

karupayun:
> Yeah, that would do the trick. But again feels a bit error-prone, as it is easy to accidentally insert a case in between. But if you both prefer that, I'd be fine with going for that. Miss Scala, where it would be `case kGpuActivity | kGpuCommandBuffer | kGpuDebugMarker: `....

Yes, I think that is less error prone. And also, there is an use example in Google C++ Style (https://engdoc.corp.google.com/eng/doc/devguide/cpp/styleguide.md?cl=head#Loops_and_Switch_Statements), so I think that is safer to go in that direction.

florian-kuebler:
> > Yeah, that would do the trick. But again feels a bit error-prone, as it is easy to accidentally insert a case in between. But if you both prefer that, I'd be fine with going for that. Miss Scala, where it would be `case kGpuActivity | kGpuCommandBuffer | kGpuDebugMarker: `....
> 
> Yes, I think that is less error prone. And also, there is an use example in Google C++ Style (https://engdoc.corp.google.com/eng/doc/devguide/cpp/styleguide.md?cl=head#Loops_and_Switch_Statements), so I think that is safer to go in that direction.

Okay, that convinced me. Made the cases fall-through. Still not happy, but as I said at the beginning, I guess that code will be changed anyways.

karupayun (approved):
LGTM! :)

PR #1650, opened 2021-01-11
Fixes after E2E test updates

reichlfl:
Fixes two issues on the current E2E tests:

- Accessibility names of some data views were not consistent between old and new UI. This was because the names were based on the labels of the dataview, I've changed this to explicitely set the accessibility names for the panels instead
- Filtering the process list did not always work. I assume it's because there is a default wait of "0" in pywinauto in between setting the filter text and checking the list


danielfenner (approved):
lgtm

PR #1652, opened 2021-01-11
Add smoke test for thread state collection

dpallotti:
Bug: http://b/176960292

### DIFF
--- a/contrib/automation_tests/test_cases/capture_window.py
+++ b/contrib/automation_tests/test_cases/capture_window.py
@@ -188,9 +188,22 @@ class Capture(E2ETestCase):
         logging.info('Showing capture window')
         self.find_control("TabItem", "Capture").click_input()
 
-    def _execute(self, length_in_seconds: int = 5):
-        self._show_capture_window()
+    def _set_capture_options(self, collect_thread_states: bool):
+        capture_tab = self.find_control('Group', "CaptureTab")
 
+        logging.info('Opening "Capture Options" dialog')
+        capture_options_button = self.find_control('Button', 'Capture Options', parent=capture_tab)
+        capture_options_button.click_input()
+
+        collect_thread_states_checkbox = self.find_control('CheckBox', 'Collect thread states')
+        if collect_thread_states_checkbox.get_toggle_state() != collect_thread_states:
+            logging.info('Toggling "Collect thread states" checkbox')
+            collect_thread_states_checkbox.click_input()
+
+        logging.info('Saving "Capture Options"')
+        self.find_control('Button', 'OK').click_input()
+
+    def _take_capture(self, length_in_seconds: int):
         capture_tab = self.find_control('Group', "CaptureTab")
         toggle_capture_button = self.find_control('Button', 'Toggle Capture', parent=capture_tab)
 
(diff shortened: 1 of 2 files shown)


contrib/automation_tests/test_cases/capture_window.py
```diff
@@ -188,9 +188,22 @@ def _show_capture_window(self):
...
-    def _execute(self, length_in_seconds: int = 5):
-        self._show_capture_window()
+    def _set_capture_options(self, collect_thread_states: bool):
+        capture_tab = self.find_control('Group', "CaptureTab")
+
+        logging.info('Opening "Capture Options" dialog')
+        capture_options_button = self.find_control('Button', 'Capture Options', parent=capture_tab)
+        capture_options_button.click_input()
+
+        collect_thread_states_checkbox = self.find_control('CheckBox', 'Collect thread states')
```
reichlfl, line 198: I'm a bit suprised this is working TBH - is the dialog that is being opened parented under the main window? Because by default, the current top window is cached (see `orbit_e2e.E2ETestSuite.top_window`) for performance reasons, so when a new top window appears, this cache needs to be refreshed explicitly.

Could you double-check (in accessibility insights e.g.) if the dialog is located underneath the current main window? Just want to make sure there is no error in the caching and / or this doesn't just work "by accident".

contrib/automation_tests/test_cases/capture_window.py
```diff
@@ -188,9 +188,22 @@ def _show_capture_window(self):
...
+        if collect_thread_states_checkbox.get_toggle_state() != collect_thread_states:
+            logging.info('Toggling "Collect thread states" checkbox')
+            collect_thread_states_checkbox.click_input()
+
+        logging.info('Saving "Capture Options"')
+        self.find_control('Button', 'OK').click_input()
```
reichlfl, line 204: If, as commented below, the dialog is really parented underneath the top window, 'OK' is a likely candidate to be ambiguous. Would be more robust to pass the dialog as explicit parent to all `find_control` calls in this method.

reichlfl (approved):
Thanks for adding this, LGTM! I have some comments on robustness you may want to tackle, but I'm also ok with this state.

contrib/automation_tests/test_cases/capture_window.py
dpallotti: From Accessibility Insights the dialog is indeed underneath the main window. What do you think the best approach is? Probably passing the dialog window to `find_control` as per the other comment?

contrib/automation_tests/test_cases/capture_window.py
dpallotti: I think you are right, I'll try to get the "Capture Options" window and pass it to `find_control`.

contrib/automation_tests/test_cases/capture_window.py
reichlfl: Interesting, because context-menus e.g. show up as new top level window. Probably because the Dialog is modal...
I think the most stable way is to find the dialog using `find_control` without explicit parent once, and then passing it explicitly as parent to the remaining `find_control` calls.

dpallotti: Yep, done like that, please have another look.

contrib/automation_tests/test_cases/capture_window.py
dpallotti: Done.

danielfenner (commented):
lgtm

reichlfl (approved):
Thanks for the changes, LGTM!
One comment: It _might_ happen that the dialog does not pop up fast enough before `capture_options_dialog = self.find_control('Window', 'Capture Options')` - in this case, you may want to wait for its existence with `orbit_e2e.wait_for_condition` (I've seen there's also similar methods in pywinauto.timings that we could use as well, btw).

But I would only care about this if you run into issues with the test execution. There is a default wait time after `click_input()` and we should only introduce additional waits if required.

dpallotti:
> Thanks for the changes, LGTM!
> One comment: It _might_ happen that the dialog does not pop up fast enough before `capture_options_dialog = self.find_control('Window', 'Capture Options')` - in this case, you may want to wait for its existence with `orbit_e2e.wait_for_condition` (I've seen there's also similar methods in pywinauto.timings that we could use as well, btw).
> 
> But I would only care about this if you run into issues with the test execution. There is a default wait time after `click_input()` and we should only introduce additional waits if required.

Thanks for the heads-up, so I'll merge as is and we keep that in mind.

PR #1653, opened 2021-01-11
Disable save capture action when no capture is taken

vickyliu-go4it:
Disable save capture action on the capture tool bar when no capture is 
taken. Also gray out disabled buttons on the capture tool bar.

Bug: http://b/161533525.

### DIFF
--- a/OrbitQt/orbitmainwindow.cpp
+++ b/OrbitQt/orbitmainwindow.cpp
@@ -25,6 +25,7 @@
 #include <QFileDialog>
 #include <QFlags>
 #include <QFontMetrics>
+#include <QGraphicsOpacityEffect>
 #include <QGridLayout>
 #include <QHBoxLayout>
 #include <QIODevice>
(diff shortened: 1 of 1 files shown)


vickyliu-go4it:
When capturing:
when_capturing

When no capture is taken:
when_no_capture_data

karupayun (commented):
Nice!! Should we also disable "save capture" after clearing?

vickyliu-go4it:
> Nice!! Should we also disable "save capture" after clearing?

Yep! With this change, the save capture action is also disabled after clearing capture.

karupayun (approved):
Yes, sorry. You are right, I confused the saved capture button with the open one! :P

LGTM!! :)

OrbitQt/orbitmainwindow.cpp
```diff
@@ -640,6 +641,13 @@ void OrbitMainWindow::UpdateCaptureStateDependentWidgets() {
   if (hint_frame_ != nullptr) {
     hint_frame_->setVisible(!has_data);
   }
+
+  // Gray out disabled actions on the capture toolbar.
+  for (QAction* action : ui->capture_toolbar->actions()) {
+    QGraphicsOpacityEffect* effect = new QGraphicsOpacityEffect;
+    effect->setOpacity(action->isEnabled() ? 1 : 0.3);
+    ui->capture_toolbar->widgetForAction(action)->setGraphicsEffect(effect);
```
antonrohr, line 649: This heap allocation without the corresponding `delete` feels a little bit wrong to me, but I guess this is the qt way and there is no way around it. 

Maybe you can put a small comment explaining that `setGraphicsEffect()` transfers the ownership of effect to the QWidget and it means that effect actually gets deleted at some point.

antonrohr (approved):
Cool change! Thanks for doing that. Only one small comment

PR #1654, opened 2021-01-11
E2E Symbols tab and timeout fixes

reichlfl:
First commit: Fix a broken wait in the symbol tab
Second commit: Replace set_edit_text with send_keys. In the current version of PyWinAuto, the sleep inside set_edit_text is commented out, thus any settings we change in timings.Timings.after_setedittext_wait is ignored.


danielfenner (approved):
lgtm!

PR #1657, opened 2021-01-11
Replace qt5-default dependency by qtbase5-dev

pierricgimmig:
qt5-default is being removed from Debian.

b/177100727


beckerhe (approved):
Thanks Pierric! LGTM!

PR #1660, opened 2021-01-12
Small fix for frame track UI

ronaldfw:
After taking a capture and when unhooking a function that was enabled as
a frame track, the frame track is now removed and the frame track icon
is not shown (was shown before). This is more consistent behaviour than
before where it was possible to end up in a state where the frame track
was enabled for the current capture but disabled for the next capture
(with confusing UI options in the context menu).

Bug: http://b/177302941
Tested: Manual tests (take capture, unhook, load capture with frame
track).

### DIFF
--- a/OrbitGl/FunctionsDataView.cpp
+++ b/OrbitGl/FunctionsDataView.cpp
@@ -58,18 +58,8 @@ bool FunctionsDataView::ShouldShowSelectedFunctionIcon(OrbitApp* app,
 }
 
 bool FunctionsDataView::ShouldShowFrameTrackIcon(OrbitApp* app, const FunctionInfo& function) {
-  if (app->IsFrameTrackEnabled(function)) {
-    return true;
-  }
-  if (app->HasCaptureData()) {
-    const CaptureData& capture_data = app->GetCaptureData();
-    if (!app->IsCaptureConnected(capture_data) && app->HasFrameTrackInCaptureData(function)) {
-      // This case occurs when loading a capture. We still want to show the indicator that a frame
-      // track is enabled for the function.
-      return true;
-    }
-  }
-  return false;
+  return app->IsFrameTrackEnabled(function) ||
+         (app->HasCaptureData() && app->HasFrameTrackInCaptureData(function));
 }
 
 std::string FunctionsDataView::BuildSelectedColumnsString(OrbitApp* app,
@@ -218,10 +208,13 @@ void FunctionsDataView::OnContextMenu(const std::string& action, int menu_index,
   } else if (action == kMenuActionUnselect) {
     for (int i : item_indices) {
       app_->DeselectFunction(*GetFunction(i));
-      // If a function is deselected, we have to make sure that the frame track is
-      // not created for this function on the next capture. However, we do not
-      // want to remove the frame track from the capture data.
+      // Unhooking a function implies disabling (and removing) the frame
+      // track for this function. While it would be possible to keep the
+      // current frame track in the capture data, this would lead to a
+      // somewhat inconsistent state where the frame track for this function
+      // is enabled for the current capture but disabled for the next one.
       app_->DisableFrameTrack(*GetFunction(i));
+      app_->RemoveFrameTrack(*GetFunction(i));
     }
   } else if (action == kMenuActionEnableFrameTrack) {
     for (int i : item_indices) {
--- a/OrbitGl/LiveFunctionsDataView.cpp
+++ b/OrbitGl/LiveFunctionsDataView.cpp
@@ -252,8 +252,13 @@ void LiveFunctionsDataView::OnContextMenu(const std::string& action, int menu_in
         app_->SelectFunction(*selected_function);
       } else if (action == kMenuActionUnselect) {
         app_->DeselectFunction(*selected_function);
-        // We disable the frame track, but do not remove it from current capture data.
+        // Unhooking a function implies disabling (and removing) the frame
+        // track for this function. While it would be possible to keep the
+        // current frame track in the capture data, this would lead to a
+        // somewhat inconsistent state where the frame track for this function
+        // is enabled for the current capture but disabled for the next one.
         app_->DisableFrameTrack(*selected_function);
+        app_->RemoveFrameTrack(*selected_function);
       } else if (action == kMenuActionDisassembly) {
         int32_t pid = capture_data.process_id();
         app_->Disassemble(pid, *selected_function);


karupayun (commented):
Thanks Ronald for the PR! It was something that confused me although it was working properly.

Could I ask you to also correct the "Enable/Disable frame Track(s)"? Because after Unhooking a function with frame tracks, we are still having an inconsistency there.

enable-disable frame tracks

ronaldfw:
> Thanks Ronald for the PR! It was something that confused me although it was working properly.
> 
> Could I ask you to also correct the "Enable/Disable frame Track(s)"? Because after Unhooking a function with frame tracks, we are still having an inconsistency there.

That's a tricky case. Technically, when unhooking a function, the frame track is disabled (for the next capture), but it's also still "enabled" (it's in the current capture data). 

I think the options we have are:

1) Keep it as it is now. 
2) Unhook also removes the frame track (and the "hook" and "enable frame track" options are then available).
3) Unhook does not remove the frame track, after "unhook" we still show option "disable frame track" (which removes the frame track). This means that enabling the frame track for the next capture would now require two steps (disable, then enable).

With 2) there is a small inconvenience that unhooking removes the frame track from the current capture, but it feels the most consistent to me. 

WDYT?

florian-kuebler:
> > Thanks Ronald for the PR! It was something that confused me although it was working properly.
> > Could I ask you to also correct the "Enable/Disable frame Track(s)"? Because after Unhooking a function with frame tracks, we are still having an inconsistency there.
> 
> That's a tricky case. Technically, when unhooking a function, the frame track is disabled (for the next capture), but it's also still "enabled" (it's in the current capture data).
> 
> I think the options we have are:
> 
> 1. Keep it as it is now.
> 2. Unhook also removes the frame track (and the "hook" and "enable frame track" options are then available).
> 3. Unhook does not remove the frame track, after "unhook" we still show option "disable frame track" (which removes the frame track). This means that enabling the frame track for the next capture would now require two steps (disable, then enable).
> 
> With 2) there is a small inconvenience that unhooking removes the frame track from the current capture, but it feels the most consistent to me.
> 
> WDYT?

Sorry for being late (haven't looked at the code yet). 
In fact this is an interesting problem. I'd be afraid of keeping it as it is now, as this feels inconsistent.
Option 2 and 3 both seem to be consistent. Option 3 has the advantage that it separates "frame tracks" and "hooked functions" a bit. This might be more obvious to the user, but on the other hand adds overhead to their workflow -- unsure if we really want this.
Option 2, on the other hand seems very neat from a workflow perspective, but might be surprising. Maybe, we can inform the user the first time that this will also delete the frame track (we already have those pop-ups with a "Dont ask again" option, right).

ronaldfw:
> Option 2 and 3 both seem to be consistent. Option 3 has the advantage that it separates "frame tracks" and "hooked functions" a bit. This might be more obvious to the user, but on the other hand adds overhead to their workflow -- unsure if we really want this.

The inconsistency with option 3 to me is that we show "disable frame track" even though it's _not_ enabled at the moment (after unhooking a function) for the next capture. It's a weird in-between state of enabled and disabled (enabled for the current capture, disabled for the next capture). 

For simplicity right now I think I would prefer option 2), but I'll play around with it a bit to see if the code is as straightforward as I hope it is.

karupayun:
I like option 2 (with Florian's suggestion). Option 1 seems inconsistent and option 3 would always have some confusion with current/next captures that I prefer to avoid.

ronaldfw:
> I like option 2 (with Florian's suggestion). Option 1 seems inconsistent and option 3 would always have some confusion with current/next captures that I prefer to avoid.

Do you think the warning dialogue is necessary? The frame track can be recreated right away with right-click and "Enable frame track" if it was unexpected that unhooking also removed the frame track (it's the same number of clicks needed to click "don't show again" and OK in the dialogue :) ). It's not that the frame track is gone and can't be gotten back. And it would be a lot more code needed as I need to pass the callback for the warning dialogue into the LiveFunctionsDataView and FunctionsDataView.

florian-kuebler:
> > Option 2 and 3 both seem to be consistent. Option 3 has the advantage that it separates "frame tracks" and "hooked functions" a bit. This might be more obvious to the user, but on the other hand adds overhead to their workflow -- unsure if we really want this.
> 
> The inconsistency with option 3 to me is that we show "disable frame track" even though it's _not_ enabled at the moment (after unhooking a function) for the next capture. It's a weird in-between state of enabled and disabled (enabled for the current capture, disabled for the next capture).
> 
> For simplicity right now I think I would prefer option 2), but I'll play around with it a bit to see if the code is as straightforward as I hope it is.

Agree, so lets go for option 2.

Its hard to predict, whether the dialogue is needed. Not sure how much work it is to add it. I'd say, if it is less then 15min, I'd go for it for the moment and eventually turn that dialogue off in the future. If it is more work, I don't think it is worth it. The user will learn the how it works after one or two usages.

florian-kuebler (approved):
Code-wise LGTM. Thanks!

ronaldfw:
> > > Option 2 and 3 both seem to be consistent. Option 3 has the advantage that it separates "frame tracks" and "hooked functions" a bit. This might be more obvious to the user, but on the other hand adds overhead to their workflow -- unsure if we really want this.
> > 
> > 
> > The inconsistency with option 3 to me is that we show "disable frame track" even though it's _not_ enabled at the moment (after unhooking a function) for the next capture. It's a weird in-between state of enabled and disabled (enabled for the current capture, disabled for the next capture).
> > For simplicity right now I think I would prefer option 2), but I'll play around with it a bit to see if the code is as straightforward as I hope it is.
> 
> Agree, so lets go for option 2.
> 
> Its hard to predict, whether the dialogue is needed. Not sure how much work it is to add it. I'd say, if it is less then 15min, I'd go for it for the moment and eventually turn that dialogue off in the future. If it is more work, I don't think it is worth it. The user will learn the how it works after one or two usages.

I might be able to do it in 15 mins, but it's also added code and I'm not sure that's worth it. It's fairly obvious what is happening and can be reverted right away by the user.

karupayun:
> Do you think the warning dialogue is necessary? The frame track can be recreated right away with right-click and "Enable frame track" if it was unexpected that unhooking also removed the frame track (it's the same number of clicks needed to click "don't show again" and OK in the dialogue :) ). It's not that the frame track is gone and can't be gotten back. And it would be a lot more code needed as I need to pass the callback for the warning dialogue into the LiveFunctionsDataView and FunctionsDataView.

You are completely right, you can add FrameTrack again! I'd forgot that!

PR #1661, opened 2021-01-12
Handle Vulkan events in CaptureEventProcessor.

florian-kuebler:
This adds support of the Vulkan related capture events (GpuQueueSubmission)
to CaptureEventProcessor.
For each command buffer and completed debug marker of a queue submission,
we create a TimerInfo object.

In order to convert the GPU timestamps to CPU timestamps, we try match each
queue submission to the corresponding tracepoint-based GpuJob. This requires
us to store the GpuJob/GpuSubmission until the counter-part has been processed
(remember, that the events origin from different data collectors).

If debug markers span across different submissions and we haven't captured
the "begin" marker, we do not discard this timer, but instead use the
begin of capture as start date.

Note, that the change to the TracepointEvent proto was needed, to fix the compiler warning when comparing int64 with uinz64 when computing the capture begin.

Test: Run tests.
Bug: http://b/176809754.

### DIFF
--- a/src/OrbitCaptureClient/CMakeLists.txt
+++ b/src/OrbitCaptureClient/CMakeLists.txt
@@ -16,11 +16,13 @@ target_include_directories(OrbitCaptureClient PRIVATE
 target_sources(OrbitCaptureClient PUBLIC 
         include/OrbitCaptureClient/CaptureClient.h
         include/OrbitCaptureClient/CaptureListener.h
-        include/OrbitCaptureClient/CaptureEventProcessor.h)
+        include/OrbitCaptureClient/CaptureEventProcessor.h
+        include/OrbitCaptureClient/GpuQueueSubmissionProcessor.h)
 
 target_sources(OrbitCaptureClient PRIVATE 
         CaptureClient.cpp
-        CaptureEventProcessor.cpp)
+        CaptureEventProcessor.cpp
+        GpuQueueSubmissionProcessor.cpp)
 
 target_link_libraries(OrbitCaptureClient PUBLIC 
         OrbitCore
--- a/src/OrbitCaptureClient/CaptureEventProcessor.cpp
+++ b/src/OrbitCaptureClient/CaptureEventProcessor.cpp
@@ -16,6 +16,7 @@
 #include "capture_data.pb.h"
 
 using orbit_client_protos::CallstackEvent;
+using orbit_client_protos::Color;
 using orbit_client_protos::LinuxAddressInfo;
 using orbit_client_protos::ThreadStateSliceInfo;
 using orbit_client_protos::TimerInfo;
@@ -25,7 +26,9 @@ using orbit_grpc_protos::Callstack;
 using orbit_grpc_protos::CallstackSample;
 using orbit_grpc_protos::CaptureEvent;
 using orbit_grpc_protos::FunctionCall;
+using orbit_grpc_protos::GpuCommandBuffer;
 using orbit_grpc_protos::GpuJob;
+using orbit_grpc_protos::GpuQueueSubmission;
 using orbit_grpc_protos::InternedCallstack;
 using orbit_grpc_protos::InternedString;
 using orbit_grpc_protos::IntrospectionScope;
@@ -72,7 +75,8 @@ void CaptureEventProcessor::ProcessEvent(const CaptureEvent& event) {
       ProcessTracepointEvent(event.tracepoint_event());
       break;
     case CaptureEvent::kGpuQueueSubmission:
-      UNREACHABLE();
+      ProcessGpuQueueSubmission(event.gpu_queue_submission());
+      break;
     case CaptureEvent::kModuleUpdateEvent:
       // TODO (http://b/168797897): Process module update events
       break;
@@ -91,6 +95,10 @@ void CaptureEventProcessor::ProcessSchedulingSlice(const SchedulingSlice& schedu
   timer_info.set_processor(static_cast<int8_t>(scheduling_slice.core()));
   timer_info.set_depth(timer_info.processor());
   timer_info.set_type(TimerInfo::kCoreActivity);
+
+  if (begin_capture_time_ns_ > scheduling_slice.in_timestamp_ns()) {
+    begin_capture_time_ns_ = scheduling_slice.in_timestamp_ns();
+  }
 
   capture_listener_->OnTimer(timer_info);
 }
@@ -116,6 +124,11 @@ void CaptureEventProcessor::ProcessCallstackSample(const CallstackSample& callst
   callstack_event.set_time(callstack_sample.timestamp_ns());
   callstack_event.set_callstack_hash(hash);
   callstack_event.set_thread_id(callstack_sample.tid());
+
+  if (begin_capture_time_ns_ > callstack_sample.timestamp_ns()) {
+    begin_capture_time_ns_ = callstack_sample.timestamp_ns();
+  }
+
   capture_listener_->OnCallstackEvent(std::move(callstack_event));
 }
 
@@ -135,6 +148,10 @@ void CaptureEventProcessor::ProcessFunctionCall(const FunctionCall& function_cal
     timer_info.add_registers(function_call.registers(i));
   }
 
+  if (begin_capture_time_ns_ > function_call.begin_timestamp_ns()) {
+    begin_capture_time_ns_ = function_call.begin_timestamp_ns();
+  }
+
   capture_listener_->OnTimer(timer_info);
 }
 
@@ -150,44 +167,57 @@ void CaptureEventProcessor::ProcessIntrospectionScope(
   timer_info.set_processor(-1);        // cpu info not available, set to invalid value
   timer_info.set_type(TimerInfo::kIntrospection);
   timer_info.mutable_registers()->CopyFrom(introspection_scope.registers());
+
+  if (begin_capture_time_ns_ > introspection_scope.begin_timestamp_ns()) {
+    begin_capture_time_ns_ = introspection_scope.begin_timestamp_ns();
+  }
+
   capture_listener_->OnTimer(timer_info);
 }
 
 void CaptureEventProcessor::ProcessInternedString(InternedString interned_string) {
-  if (string_intern_pool.contains(interned_string.key())) {
+  if (string_intern_pool_.contains(interned_string.key())) {
     ERROR("Overwriting InternedString with key %llu", interned_string.key());
   }
-  string_intern_pool.emplace(interned_string.key(), std::move(*interned_string.mutable_intern()));
+  string_intern_pool_.emplace(interned_string.key(), std::move(*interned_string.mutable_intern()));
 }
 
 void CaptureEventProcessor::ProcessGpuJob(const GpuJob& gpu_job) {
   std::string timeline;
   if (gpu_job.timeline_or_key_case() == GpuJob::kTimelineKey) {
-    timeline = string_intern_pool[gpu_job.timeline_key()];
+    timeline = string_intern_pool_[gpu_job.timeline_key()];
   } else {
     timeline = gpu_job.timeline();
   }
   uint64_t timeline_hash = GetStringHashAndSendToListenerIfNecessary(timeline);
 
+  int32_t thread_id = gpu_job.tid();
+  uint64_t amdgpu_cs_ioctl_time_ns = gpu_job.amdgpu_cs_ioctl_time_ns();
+
   constexpr const char* sw_queue = "sw queue";
   uint64_t sw_queue_key = GetStringHashAndSendToListenerIfNecessary(sw_queue);
 
   TimerInfo timer_user_to_sched;
-  timer_user_to_sched.set_thread_id(gpu_job.tid());
-  timer_user_to_sched.set_start(gpu_job.amdgpu_cs_ioctl_time_ns());
+  timer_user_to_sched.set_thread_id(thread_id);
+  timer_user_to_sched.set_start(amdgpu_cs_ioctl_time_ns);
   timer_user_to_sched.set_end(gpu_job.amdgpu_sched_run_job_time_ns());
   timer_user_to_sched.set_depth(gpu_job.depth());
   timer_user_to_sched.set_user_data_key(sw_queue_key);
   timer_user_to_sched.set_timeline_hash(timeline_hash);
   timer_user_to_sched.set_processor(-1);
   timer_user_to_sched.set_type(TimerInfo::kGpuActivity);
+
+  if (begin_capture_time_ns_ > gpu_job.amdgpu_cs_ioctl_time_ns()) {
+    begin_capture_time_ns_ = gpu_job.amdgpu_cs_ioctl_time_ns();
+  }
+
   capture_listener_->OnTimer(std::move(timer_user_to_sched));
 
   constexpr const char* hw_queue = "hw queue";
   uint64_t hw_queue_key = GetStringHashAndSendToListenerIfNecessary(hw_queue);
 
   TimerInfo timer_sched_to_start;
-  timer_sched_to_start.set_thread_id(gpu_job.tid());
+  timer_sched_to_start.set_thread_id(thread_id);
   timer_sched_to_start.set_start(gpu_job.amdgpu_sched_run_job_time_ns());
   timer_sched_to_start.set_end(gpu_job.gpu_hardware_start_time_ns());
   timer_sched_to_start.set_depth(gpu_job.depth());
@@ -201,7 +231,7 @@ void CaptureEventProcessor::ProcessGpuJob(const GpuJob& gpu_job) {
   uint64_t hw_execution_key = GetStringHashAndSendToListenerIfNecessary(hw_execution);
 
   TimerInfo timer_start_to_finish;
-  timer_start_to_finish.set_thread_id(gpu_job.tid());
+  timer_start_to_finish.set_thread_id(thread_id);
   timer_start_to_finish.set_start(gpu_job.gpu_hardware_start_time_ns());
   timer_start_to_finish.set_end(gpu_job.dma_fence_signaled_time_ns());
   timer_start_to_finish.set_depth(gpu_job.depth());
@@ -210,6 +240,55 @@ void CaptureEventProcessor::ProcessGpuJob(const GpuJob& gpu_job) {
   timer_start_to_finish.set_processor(-1);
   timer_start_to_finish.set_type(TimerInfo::kGpuActivity);
   capture_listener_->OnTimer(std::move(timer_start_to_finish));
+
+  const GpuQueueSubmission* matching_gpu_submission =
+      FindMatchingGpuQueueSubmission(thread_id, amdgpu_cs_ioctl_time_ns);
+  if (matching_gpu_submission == nullptr || matching_gpu_submission->num_begin_markers() > 0) {
+    tid_to_submission_time_to_gpu_job_[thread_id][amdgpu_cs_ioctl_time_ns] = gpu_job;
+  }
+  if (matching_gpu_submission == nullptr) {
+    return;
+  }
+
+  DoProcessGpuQueueSubmission(*matching_gpu_submission, gpu_job);
+
+  uint64_t post_submission_cpu_timestamp =
+      matching_gpu_submission->meta_info().post_submission_cpu_timestamp();
+
+  if (!HasUnprocessedBeginMarkers(thread_id, post_submission_cpu_timestamp)) {
+    tid_to_post_submission_time_to_gpu_submission_.at(thread_id).erase(
+        post_submission_cpu_timestamp);
+  }
+}
+
+void CaptureEventProcessor::ProcessGpuQueueSubmission(
+    const GpuQueueSubmission& gpu_queue_submission) {
+  int32_t thread_id = gpu_queue_submission.meta_info().tid();
+  uint64_t pre_submission_cpu_timestamp =
+      gpu_queue_submission.meta_info().pre_submission_cpu_timestamp();
+  uint64_t post_submission_cpu_timestamp =
+      gpu_queue_submission.meta_info().post_submission_cpu_timestamp();
+  const GpuJob* matching_gpu_job =
+      FindMatchingGpuJob(thread_id, pre_submission_cpu_timestamp, post_submission_cpu_timestamp);
+
+  if (matching_gpu_job == nullptr || gpu_queue_submission.num_begin_markers() > 0) {
+    tid_to_post_submission_time_to_gpu_submission_[thread_id][post_submission_cpu_timestamp] =
+        gpu_queue_submission;
+  }
+  if (gpu_queue_submission.num_begin_markers() > 0) {
+    tid_to_post_submission_time_to_num_begin_markers_[thread_id][post_submission_cpu_timestamp] =
+        gpu_queue_submission.num_begin_markers();
+  }
+  if (matching_gpu_job == nullptr) {
+    return;
+  }
+
+  DoProcessGpuQueueSubmission(gpu_queue_submission, *matching_gpu_job);
+
+  if (!HasUnprocessedBeginMarkers(thread_id, post_submission_cpu_timestamp)) {
+    tid_to_submission_time_to_gpu_job_.at(thread_id).erase(
+        matching_gpu_job->amdgpu_cs_ioctl_time_ns());
+  }
 }
 
 void CaptureEventProcessor::ProcessThreadName(const ThreadName& thread_name) {
@@ -255,20 +334,25 @@ void CaptureEventProcessor::ProcessThreadStateSlice(const ThreadStateSlice& thre
   }
   slice_info.set_begin_timestamp_ns(thread_state_slice.begin_timestamp_ns());
   slice_info.set_end_timestamp_ns(thread_state_slice.end_timestamp_ns());
+
+  if (begin_capture_time_ns_ > thread_state_slice.begin_timestamp_ns()) {
+    begin_capture_time_ns_ = thread_state_slice.begin_timestamp_ns();
+  }
+
   capture_listener_->OnThreadStateSlice(std::move(slice_info));
 }
 
 void CaptureEventProcessor::ProcessAddressInfo(const AddressInfo& address_info) {
   std::string function_name;
   if (address_info.function_name_or_key_case() == AddressInfo::kFunctionNameKey) {
-    function_name = string_intern_pool[address_info.function_name_key()];
+    function_name = string_intern_pool_[address_info.function_name_key()];
   } else {
     function_name = address_info.function_name();
   }
 
   std::string map_name;
   if (address_info.map_name_or_key_case() == AddressInfo::kMapNameKey) {
-    map_name = string_intern_pool[address_info.map_name_key()];
+    map_name = string_intern_pool_[address_info.map_name_key()];
   } else {
     map_name = address_info.map_name();
   }
@@ -321,6 +405,10 @@ void CaptureEventProcessor::ProcessTracepointEvent(
   tracepoint_event_info.set_cpu(tracepoint_event.cpu());
   tracepoint_event_info.set_tracepoint_info_key(hash);
 
+  if (begin_capture_time_ns_ > tracepoint_event.time()) {
+    begin_capture_time_ns_ = tracepoint_event.time();
+  }
+
   capture_listener_->OnTracepointEvent(std::move(tracepoint_event_info));
 }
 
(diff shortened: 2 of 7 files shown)


OrbitCaptureClient/include/OrbitCaptureClient/CaptureEventProcessor.h
```diff
@@ -57,6 +83,14 @@ class CaptureEventProcessor {
   absl::flat_hash_set<uint64_t> tracepoint_hashes_seen_;
   void SendTracepointInfoToListenerIfNecessary(
       const orbit_grpc_protos::TracepointInfo& tracepoint_info, const uint64_t& hash);
+
+  absl::flat_hash_map<int32_t, std::map<uint64_t, orbit_grpc_protos::GpuJob>>
+      tid_to_submission_time_to_gpu_job_;
```
dpallotti, line 88: Would it be more accurate to call this `tid_to_pre_submission_time_to_gpu_job_` (note the "pre") or did I get it wrong? Or do you think it's better to omit it?

OrbitCaptureClient/CaptureEventProcessor.cpp
```diff
@@ -340,3 +428,254 @@ void CaptureEventProcessor::SendTracepointInfoToListenerIfNecessary(
     capture_listener_->OnUniqueTracepointInfo(hash, tracepoint_info);
   }
 }
+
+const GpuQueueSubmission* CaptureEventProcessor::FindMatchingGpuQueueSubmission(
```
dpallotti, line 432: The relationship between job and submission is 1 to 1, yes?

OrbitCaptureClient/CaptureEventProcessor.cpp
```diff
@@ -340,3 +428,254 @@ void CaptureEventProcessor::SendTracepointInfoToListenerIfNecessary(
...
+  --gpu_job_matching_post_submission_it;
+
+  if (&gpu_job_matching_pre_submission_it->second != &gpu_job_matching_post_submission_it->second) {
+    return nullptr;
+  }
+
+  return &gpu_job_matching_pre_submission_it->second;
+}
+
+void CaptureEventProcessor::DoProcessGpuQueueSubmission(
```
dpallotti, line 494: Maybe call it `ProcessGpuQueueSubmissionWithMatchingGpuJob`?

OrbitCaptureClient/CaptureEventProcessor.cpp
```diff
@@ -340,3 +428,254 @@ void CaptureEventProcessor::SendTracepointInfoToListenerIfNecessary(
...
+      color->set_green(static_cast<uint32_t>(completed_marker.color().green() * 255));
+      color->set_blue(static_cast<uint32_t>(completed_marker.color().blue() * 255));
+      color->set_alpha(static_cast<uint32_t>(completed_marker.color().alpha() * 255));
+    }
+    marker_timer.set_user_data_key(text_key);
+    capture_listener_->OnTimer(marker_timer);
+  }
+}
+
+std::optional<orbit_grpc_protos::GpuCommandBuffer> CaptureEventProcessor::FindFirstCommandBuffer(
```
dpallotti, line 673: I would say this is "get" more than "find". Maybe "extract" if you really don't want to use "get".

OrbitCaptureClient/CaptureEventProcessor.cpp
```diff
@@ -340,3 +428,254 @@ void CaptureEventProcessor::SendTracepointInfoToListenerIfNecessary(
...
+  return true;
+}
+
+void CaptureEventProcessor::DecrementUnprocessedBeginMarkers(int32_t thread_id,
+                                                             uint64_t post_submission_timestamp) {
+  CHECK(tid_to_post_submission_time_to_num_begin_markers_.contains(thread_id));
+  auto& post_submission_time_to_num_begin_markers =
+      tid_to_post_submission_time_to_num_begin_markers_.at(thread_id);
+  CHECK(post_submission_time_to_num_begin_markers.contains(post_submission_timestamp));
+  uint32_t new_num = --post_submission_time_to_num_begin_markers[post_submission_timestamp];
```
dpallotti, line 534: Use `.at`?

OrbitCaptureClient/CaptureEventProcessor.cpp
```diff
@@ -340,3 +428,254 @@ void CaptureEventProcessor::SendTracepointInfoToListenerIfNecessary(
...
+                             begin_submission_first_command_buffer->begin_gpu_timestamp_ns());
+      if (begin_marker_thread_id == gpu_queue_submission.meta_info().tid()) {
+        marker_timer.set_thread_id(begin_marker_thread_id);
+      }
+
+      DecrementUnprocessedBeginMarkers(begin_marker_thread_id,
+                                       begin_marker_post_submission_cpu_timestamp);
+    } else {
+      marker_timer.set_start(begin_capture_time_ns_);
+      marker_timer.set_thread_id(-1);
```
dpallotti, line 646: So the thread id is lost, it would only come with the begin marker?

And should -1 be a constant somehow? It would probably be complicated to propagate it to the entire codebase though. I guess it should be a constant defined in the `TimerInfo` proto but it seems that that can't be done. Hmm.

OrbitCaptureClient/CaptureEventProcessor.cpp
```diff
@@ -340,3 +428,254 @@ void CaptureEventProcessor::SendTracepointInfoToListenerIfNecessary(
...
+      if (submission_pre_submission_cpu_timestamp == begin_marker_pre_submission_cpu_timestamp &&
+          submission_post_submission_cpu_timestamp == begin_marker_post_submission_cpu_timestamp &&
+          submission_thread_id == begin_marker_thread_id) {
+        matching_begin_submission = &gpu_queue_submission;
+      } else {
+        matching_begin_submission = FindMatchingGpuQueueSubmission(
+            begin_marker_thread_id, begin_marker_post_submission_cpu_timestamp);
+      }
+
+      // Note, we receive submissions of a single queue in order (by CPU submission time). So if
```
dpallotti, line 621: "Note:" or "Note that".

OrbitCaptureClient/CaptureEventProcessor.cpp
```diff
@@ -340,3 +428,254 @@ void CaptureEventProcessor::SendTracepointInfoToListenerIfNecessary(
...
+      // there is no matching "begin submission", the "begin" was submitted before the "end" and
+      // we lost the record of the "begin's" submission (which should not happen).
```
dpallotti, line 623: `"begin's" submission` -> `"begin" submission` or (like previous line) `"begin submission"`?

OrbitCaptureClient/CaptureEventProcessor.cpp
```diff
@@ -340,3 +428,254 @@ void CaptureEventProcessor::SendTracepointInfoToListenerIfNecessary(
...
+      CHECK(matching_begin_submission != nullptr);
```
dpallotti, line 624: Are we sure this should be a `CHECK` and not handled gracefully?

OrbitCaptureClient/CaptureEventProcessor.cpp
```diff
@@ -340,3 +428,254 @@ void CaptureEventProcessor::SendTracepointInfoToListenerIfNecessary(
...
+      const auto& begin_marker_info = completed_marker.begin_marker();
+      const auto& begin_marker_meta_info = begin_marker_info.meta_info();
+      const int32_t begin_marker_thread_id = begin_marker_meta_info.tid();
+      uint64_t begin_marker_post_submission_cpu_timestamp =
+          begin_marker_meta_info.post_submission_cpu_timestamp();
+      uint64_t begin_marker_pre_submission_cpu_timestamp =
+          begin_marker_meta_info.pre_submission_cpu_timestamp();
+
+      const GpuQueueSubmission* matching_begin_submission = nullptr;
+      if (submission_pre_submission_cpu_timestamp == begin_marker_pre_submission_cpu_timestamp &&
```
dpallotti, line 612: Could you explain this `if`? I'm confused.

OrbitCaptureClient/CaptureEventProcessor.cpp
```diff
@@ -340,3 +428,254 @@ void CaptureEventProcessor::SendTracepointInfoToListenerIfNecessary(
...
+      CHECK(begin_submission_first_command_buffer.has_value());
+
+      const GpuJob* matching_begin_job = FindMatchingGpuJob(
+          begin_marker_thread_id, begin_marker_meta_info.pre_submission_cpu_timestamp(),
+          begin_marker_post_submission_cpu_timestamp);
+      CHECK(matching_begin_job != nullptr);
+
+      marker_timer.set_start(completed_marker.begin_marker().gpu_timestamp_ns() +
+                             matching_begin_job->gpu_hardware_start_time_ns() -
+                             begin_submission_first_command_buffer->begin_gpu_timestamp_ns());
```
dpallotti, line 637: `matching_begin_job->gpu_hardware_start_time_ns() - begin_submission_first_command_buffer->begin_gpu_timestamp_ns()` is the difference between the cpu and the gpu time, right?

dpallotti (commented):
I will look at the tests now but I'll send out the first part of the review first. I would say that my main thought would be to extract the bookkeeping related the Vulkan layer (basically from line 243 to the .cpp, so including part of `ProcessGpuJob`) to a separate class.

OrbitCaptureClient/CaptureEventProcessorTest.cpp
```diff
@@ -531,6 +545,382 @@ TEST(DISABLED_CaptureEventProcessor, CanHandleGpuJobs) {
   EXPECT_EQ(hw_excecution_timer.user_data_key(), actual_hw_execution_key);
 }
 
+GpuQueueSubmissionMetaInfo* CreateGpuQueueSubmissionMetaInfo(GpuQueueSubmission* submission,
+                                                             uint64_t pre_timestamp,
+                                                             uint64_t post_timestamp) {
+  GpuQueueSubmissionMetaInfo* meta_info = submission->mutable_meta_info();
+  meta_info->set_tid(2);
```
dpallotti, line 552: Shouldn't the tid also be a parameter?

OrbitCaptureClient/CaptureEventProcessorTest.cpp
```diff
@@ -531,6 +545,382 @@ TEST(DISABLED_CaptureEventProcessor, CanHandleGpuJobs) {
...
+void AddGpuDebugMarker(GpuQueueSubmission* submission, GpuQueueSubmissionMetaInfo* begin_meta_info,
+                       uint64_t marker_text_key, uint64_t begin_gpu_timestamp,
+                       uint64_t end_gpu_timestamp) {
+  GpuDebugMarker* debug_marker = submission->add_completed_markers();
+  Color* color = debug_marker->mutable_color();
+  color->set_alpha(1.f);
+  color->set_red(0.75f);
+  color->set_green(0.5f);
+  color->set_blue(0.25f);
+  debug_marker->set_depth(1);
```
dpallotti, line 574: Shouldn't the color and the depth be parameters too?

OrbitCaptureClient/CaptureEventProcessorTest.cpp
```diff
@@ -531,6 +545,382 @@ TEST(DISABLED_CaptureEventProcessor, CanHandleGpuJobs) {
...
+}
+
+void AddGpuCommandBuffer(GpuSubmitInfo* submit_info, uint64_t gpu_begin_timestamp,
+                         uint64_t gpu_end_timestamp) {
+  GpuCommandBuffer* command_buffer = submit_info->add_command_buffers();
+  command_buffer->set_begin_gpu_timestamp_ns(gpu_begin_timestamp);
+  command_buffer->set_end_gpu_timestamp_ns(gpu_end_timestamp);
+}
+
+void AddGpuDebugMarker(GpuQueueSubmission* submission, GpuQueueSubmissionMetaInfo* begin_meta_info,
```
dpallotti, line 565: Maybe `AddGpuDebugMarkerToSubmission` or even `AddGpuDebugMarkerToGpuQueueSubmission`?

OrbitCaptureClient/CaptureEventProcessorTest.cpp
```diff
@@ -531,6 +545,382 @@ TEST(DISABLED_CaptureEventProcessor, CanHandleGpuJobs) {
...
+                                                             uint64_t pre_timestamp,
+                                                             uint64_t post_timestamp) {
+  GpuQueueSubmissionMetaInfo* meta_info = submission->mutable_meta_info();
+  meta_info->set_tid(2);
+  meta_info->set_pre_submission_cpu_timestamp(pre_timestamp);
+  meta_info->set_post_submission_cpu_timestamp(post_timestamp);
+  return meta_info;
+}
+
+void AddGpuCommandBuffer(GpuSubmitInfo* submit_info, uint64_t gpu_begin_timestamp,
```
dpallotti, line 558: `AddGpuCommandBufferToSubmitInfo` or `AddGpuCommandBufferToGpuSubmitInfo`?

OrbitCaptureClient/CaptureEventProcessorTest.cpp
```diff
@@ -531,6 +545,382 @@ TEST(DISABLED_CaptureEventProcessor, CanHandleGpuJobs) {
...
+  EXPECT_CALL(listener, OnKeyAndString(_, "marker"))
+      .Times(1)
+      .WillOnce(SaveArg<0>(&actual_marker_key));
+
+  TimerInfo command_buffer_timer_1;
+  TimerInfo command_buffer_timer_2;
+  TimerInfo debug_marker_timer;
+  EXPECT_CALL(listener, OnTimer)
+      .Times(6)
+      // The first three timers are from the GpuJob, which we not test in here.
```
dpallotti, line 743:
```suggestion
      // The first three timers are from the GpuJob, which we don't test here.
```

florian-kuebler (commented):
Thanks for the review. I addressed most of the comments. I would like to settle this before doing the change of moving the stuff to a separate class.

OrbitCaptureClient/CaptureEventProcessor.cpp
```diff
@@ -340,3 +423,294 @@ void CaptureEventProcessor::SendTracepointInfoToListenerIfNecessary(
...
+
+void CaptureEventProcessor::DecrementUnprocessedBeginMarkers(int32_t thread_id,
+                                                             uint64_t submission_timestamp,
+                                                             uint64_t post_submission_timestamp) {
+  CHECK(tid_to_post_submission_time_to_num_begin_markers_.contains(thread_id));
+  auto& post_submission_time_to_num_begin_markers =
+      tid_to_post_submission_time_to_num_begin_markers_.at(thread_id);
+  CHECK(post_submission_time_to_num_begin_markers.contains(post_submission_timestamp));
+  uint64_t new_num = post_submission_time_to_num_begin_markers.at(post_submission_timestamp) - 1;
+  post_submission_time_to_num_begin_markers[post_submission_timestamp] = new_num;
```
dpallotti, line 533: Related to the outdated comment about using `.at`, why can't you just write `--post_submission_time_to_num_begin_markers.at(post_submission_timestamp)`?

`.at` also returns a reference (or what am I missing?).

Maybe it's clearer now without the pre-increment, but you should still be able to write
`post_submission_time_to_num_begin_markers.at(post_submission_timestamp) = new_num;`.
It doesn't make a big difference but to me using `at` makes it clearer that we know and expect that the key-value pair was already in the map.

OrbitCaptureClient/CaptureEventProcessor.cpp
dpallotti: Thanks!

OrbitCaptureClient/CaptureEventProcessor.cpp
dpallotti: I would at least add a local constant to give a name to the `-1` here.

> It does not need to be lost. Basically the point is that (if the debug marker is begins and ends on a different submission), the thread id of the "begin submission" may be different than the one of the "end submission".

I'm not sure how this applies here. Isn't this code executed when `if (completed_marker.has_begin_marker())` is false? Doesn't that just mean that the begin was missing? Or does it mean it was in a different submission?

OrbitCaptureClient/CaptureEventProcessorTest.cpp
dpallotti: Then I would at least define it as a constant.

OrbitCaptureClient/CaptureEventProcessorTest.cpp
dpallotti: As above, I would at least add constants, especially since these values are reused in `ExpectDebugMarkerTimerEq`.

OrbitCaptureClient/include/OrbitCaptureClient/CaptureEventProcessor.h
```diff
@@ -43,9 +43,35 @@ class CaptureEventProcessor {
   void ProcessInternedTracepointInfo(
       orbit_grpc_protos::InternedTracepointInfo interned_tracepoint_info);
   void ProcessTracepointEvent(const orbit_grpc_protos::TracepointEvent& tracepoint_event);
+  void ProcessGpuQueueSubmission(const orbit_grpc_protos::GpuQueueSubmission& gpu_command_buffer);
+
+  // Vulkan Layer related helpers:
```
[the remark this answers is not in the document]
dpallotti: As discussed offline, let's not pass the listener, but we can return `Timer`s.

OrbitCaptureClient/include/OrbitCaptureClient/CaptureEventProcessor.h
dpallotti: Makes sense!

OrbitCaptureClient/CaptureEventProcessor.cpp
dpallotti: But doesn't this condition depend on a lot of assumptions down the line, in particular both on the Vulkan layer and on the communication?

OrbitCaptureClient/CaptureEventProcessor.cpp
```diff
@@ -210,6 +231,63 @@ void CaptureEventProcessor::ProcessGpuJob(const GpuJob& gpu_job) {
...
+    const GpuQueueSubmission& gpu_queue_submission) {
+  int32_t thread_id = gpu_queue_submission.meta_info().tid();
+  uint64_t pre_submission_cpu_timestamp =
+      gpu_queue_submission.meta_info().pre_submission_cpu_timestamp();
+  uint64_t post_submission_cpu_timestamp =
+      gpu_queue_submission.meta_info().post_submission_cpu_timestamp();
+  const GpuJob* matching_gpu_job =
+      FindMatchingGpuJob(thread_id, pre_submission_cpu_timestamp, post_submission_cpu_timestamp);
+
+  // If we haven't found the matching "GpuSubmission" or the submission contains "begin" markers
```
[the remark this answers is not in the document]
florian-kuebler: Thanks

OrbitCaptureClient/CaptureEventProcessor.cpp
```diff
@@ -340,3 +423,294 @@ void CaptureEventProcessor::SendTracepointInfoToListenerIfNecessary(
     capture_listener_->OnUniqueTracepointInfo(hash, tracepoint_info);
   }
 }
+
+// Finds the GpuQueueSubmission that fully contains the given given timestamp and happened on the
```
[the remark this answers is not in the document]
florian-kuebler: Done.

OrbitCaptureClient/CaptureEventProcessor.cpp
florian-kuebler: Right. I somehow forgot about `.at` also returning a reference. Updated the code.

OrbitCaptureClient/CaptureEventProcessor.cpp
florian-kuebler: If I don't miss anything, it has the assumption that the layer sends the events in-order (for a given queue) and that the communication does neither get the order wrong nor drop messages. 
I agree that these assumptions to the communication are kind of strong, but I feels like there is actually something really going badly wrong if they don't hold anymore.
In general, I think it is a good scheme to make assumptions explicit, and fail if don't meet them.

OrbitCaptureClient/CaptureEventProcessor.cpp
florian-kuebler: You are totally right. That branch was only if we have not recorded the begin marker. The other case was missing (I added it a couple of lines above). Also added a local constant.

OrbitCaptureClient/CaptureEventProcessorTest.cpp
florian-kuebler: Done.

florian-kuebler (commented):
Addressed the review comments (despite the refactoring -- which I will do next).

florian-kuebler:
Just refactored the code, such that the logic for the submissions is extracted into its own class.

OrbitCaptureClient/include/OrbitCaptureClient/GpuQueueSubmissionProcessor.h
```diff
@@ -0,0 +1,110 @@
...
+#include <vector>
+
+#include "capture.pb.h"
+#include "capture_data.pb.h"
+
+// `GpuQueueSubmission` proto objects come with GPU timestamps (rather than CPU timestamps) for the
+// command buffer and debug marker timings. In order convert those timestamps into CPU time, each
+// `GpuQueueSubmission` contains a timestamp before (pre) and after (post) the vkQueueSubmit driver
+// call. For the driver we already have timestamps in the `GpuJob` events. Together with the thread
+// id, this allows us to establish a 1:1 mapping of `GpuJob`s and `GpuQueueSubmission.
```
dpallotti, line 22:
```suggestion
// id, this allows us to establish a 1:1 mapping of `GpuJob`s and `GpuQueueSubmission`s.
```

OrbitCaptureClient/include/OrbitCaptureClient/GpuQueueSubmissionProcessor.h
```diff
@@ -0,0 +1,110 @@
...
+//
+// This class allows to convert `GpuQueueSubmission` events (with GPU timestmaps) to command buffer
+// and debug marker `TimerInfo`s (with CPU timestamps). Thereby, it manages the mapping from
```
dpallotti, line 25: I think "thereby" is not fully appropriate here. "For that"?

OrbitCaptureClient/include/OrbitCaptureClient/GpuQueueSubmissionProcessor.h
```diff
@@ -0,0 +1,110 @@
...
+  // processing.
+  [[nodiscard]] std::vector<orbit_client_protos::TimerInfo> ProcessGpuJob(
+      const orbit_grpc_protos::GpuJob& gpu_job,
+      const absl::flat_hash_map<uint64_t, std::string>& string_intern_pool,
+      const std::function<uint64_t(const std::string& str)>&
+          get_string_hash_and_send_to_listener_if_necessary);
+
+  // In case we have recored the submission containing the "begin" of a certain debug marker, we
+  // use the `begin_capture_time_ns_` as an approximation for the begin CPU timestamp.
+  // This method updates this timestamp with the minimum of the current value and the given value.0
```
dpallotti, line 54:
```suggestion
  // This method updates this timestamp with the minimum of the current value and the given value.
```

OrbitCaptureClient/include/OrbitCaptureClient/GpuQueueSubmissionProcessor.h
```diff
@@ -0,0 +1,110 @@
...
+  // Finds the GpuQueueSubmission that fully contains the given timestamp and happened on the given
+  // thread id. Returns `nullptr` if there is no such submission.
+  [[nodiscard]] const orbit_grpc_protos::GpuQueueSubmission* FindMatchingGpuQueueSubmission(
+      int32_t thread_id, uint64_t submit_time);
+  [[nodiscard]] bool HasUnprocessedBeginMarkers(int32_t thread_id,
+                                                uint64_t post_submission_timestamp) const;
+  void DecrementUnprocessedBeginMarkers(int32_t thread_id, uint64_t submission_timestamp,
+                                        uint64_t post_submission_timestamp);
+  void DeleteSavedGpuJob(int32_t thread_id, uint64_t submission_timestamp);
+  void DeleteSavedGpuSubmission(int32_t thread_id, uint64_t post_submission_timestamp);
```
dpallotti, line 98: Add spaces between these methods. It's quite hard to read otherwise.

OrbitCaptureClient/GpuQueueSubmissionProcessor.cpp
```diff
@@ -0,0 +1,398 @@
...
+  std::optional<GpuCommandBuffer> first_command_buffer =
+      ExtractFirstCommandBuffer(gpu_queue_submission);
+
+  ProcessGpuCommandBuffers(gpu_queue_submission, matching_gpu_job, first_command_buffer,
+                           timeline_hash, get_string_hash_and_send_to_listener_if_necessary,
+                           &result);
+
+  ProcessGpuDebugMarkers(gpu_queue_submission, matching_gpu_job, first_command_buffer, timeline,
+                         string_intern_pool, get_string_hash_and_send_to_listener_if_necessary,
+                         &result);
```
dpallotti, line 174: Wouldn't it be a clearer if these methods returned a `std::vector<TimerInfo>` as well and these were then appended to `result`? Or is it because of performance?

OrbitCaptureClient/include/OrbitCaptureClient/GpuQueueSubmissionProcessor.h
```diff
@@ -0,0 +1,110 @@
...
+
+#include <algorithm>
+#include <functional>
+#include <string>
+#include <vector>
+
+#include "capture.pb.h"
+#include "capture_data.pb.h"
+
+// `GpuQueueSubmission` proto objects come with GPU timestamps (rather than CPU timestamps) for the
```
dpallotti, line 18: This comment is already a big reason why I prefer the separate class much better now!

dpallotti (approved):
I like it much better now. I left a few more small comments but it looks good!

OrbitCaptureClient/CaptureEventProcessor.cpp
dpallotti: I have to say that still I'm a bit unsure, but I'm fine with leaving the CHECK, at least for now. Maybe Ronald has an opinion?

florian-kuebler: @ronaldfw WDYT?

OrbitCaptureClient/GpuQueueSubmissionProcessor.cpp
florian-kuebler: There was no special reason. Changed it.

OrbitCaptureClient/include/OrbitCaptureClient/GpuQueueSubmissionProcessor.h
florian-kuebler: Done.

OrbitCaptureClient/include/OrbitCaptureClient/GpuQueueSubmissionProcessor.h
florian-kuebler: Done.

OrbitCaptureClient/include/OrbitCaptureClient/GpuQueueSubmissionProcessor.h
florian-kuebler: Thanks!

OrbitCaptureClient/include/OrbitCaptureClient/GpuQueueSubmissionProcessor.h
florian-kuebler: Done.

florian-kuebler (commented):
Thanks for the review. I rebased to master and addressed the comments.

OrbitCaptureClient/CaptureEventProcessor.cpp
ronaldfw: I don't think we have as strong assertions elsewhere about the order, do we? I would somewhat be in favor of logging this as an error if we are not really sure that the condition must be true. Assertions are for cases where you know something must hold and if it doesn't you know something bad has happened. Are we sure this condition must hold? I don't have enough context to understand this fully.

If you are sure that this cannot happen (that is, it could be proven to never happen, assuming all other code is correct) or that we cannot recover from this error, then the CHECK is appropriate. Otherwise, error logging would be better.

ronaldfw: Alternatively, leave the CHECK while we are developing and see if we ever see this problem occur. Re-evaluate when we ship (though, easy to forget to do so).

florian-kuebler: As far as I know, this is the only situation on the Client where we depend on the order being correct, which is why have no other places where we have assertions on the order being correct. From the reset of the code base, I am not 100% sure. 
On the Layer side, we have code to ensure the order of sending out is correct and also have test for that. 

So I'd like to ensure that this really the case, and that I not missed something.

florian-kuebler: >Alternatively, leave the CHECK while we are developing and see if we ever see this problem occur. Re-evaluate when we ship (though, easy to forget to do so).

I created http://b/177652631 for this and made it a blocker so that we don't forget about it.
For now, I like to see how it performs.

PR #1664, opened 2021-01-12
Fix move_tabs test

reichlfl:
This fixes several issues that showed in the last failure of move_tabs and should stabilize the tests more w.r.t. connection times and screen sizes.


danielfenner (approved):
lgtm

PR #1665, opened 2021-01-13
Fix build in debug mode

dimitry-:
Test: build clang9_debug

### DIFF
--- a/OrbitLinuxTracing/LibunwindstackUnwinder.cpp
+++ b/OrbitLinuxTracing/LibunwindstackUnwinder.cpp
@@ -11,6 +11,8 @@
 #include <array>
 #include <cstdint>
 
+#include "OrbitBase/Logging.h"  // IWYU pragma: keep
+
 namespace orbit_linux_tracing {
 
 std::unique_ptr<unwindstack::BufferMaps> LibunwindstackUnwinder::ParseMaps(


OrbitLinuxTracing/LibunwindstackUnwinder.cpp
```diff
@@ -11,6 +11,10 @@
 #include <array>
 #include <cstdint>
 
+#ifndef NDEBUG
+#include "OrbitBase/Logging.h"
```
dimitry-, line 15: without it compilation fails on L56

dpallotti: Do we upset IWYU if we include it regardless of `NDEBUG`?

dpallotti (approved):

OrbitLinuxTracing/LibunwindstackUnwinder.cpp
beckerhe: Ah makes sense. Can you also add `// IWYU pragma: keep` behind the include?

beckerhe (approved):
Otherwise LGTM. Thanks!

OrbitLinuxTracing/LibunwindstackUnwinder.cpp
dimitry-: done.

beckerhe: I think it has to go on the same line as the include itself. Ref: https://github.com/include-what-you-use/include-what-you-use/blob/master/docs/IWYUPragmas.md

dimitry-: done.

PR #1667, opened 2021-01-13
Move all code into a src/-subdirectory

beckerhe:
To have a cleaner root directory we move all module directories into a common subdirectory called `src`.

It seems to be the common name for that.

There is a second commit which prepares the presubmit checks for that change. Due to that change file
renames won't be considered modifications to the file.

Bug: http://b/177407068
Test: compiles


dimitry- (approved):

dpallotti (approved):
This is already a candidate for one of my favorite changes of the year! Thanks a lot for taking care of this!

PR #1668, opened 2021-01-13
Use (duration,end) instead of (begin,end) for CaptureEvent gRPC protos

dpallotti:
This saves some bandwidth thanks to variable-length encoding of protobufs.
More details in the bug.
GPU-related events are (at least for now) not included in the change. As the
frequency of those events is relatively low compared to scheduler, thread state,
dynamic instrumentation, that would in any case have a smalelr impact.
This does not affect "client protos" (i.e., serialization to file), so that we
don't break compatibility with previous versions.

Bug: http://b/177222925

Test: Update unit tests. Take a few captures on Trata and Infiltrator and verify
that things look good.

### DIFF
--- a/src/OrbitCaptureClient/CaptureEventProcessor.cpp
+++ b/src/OrbitCaptureClient/CaptureEventProcessor.cpp
@@ -84,7 +84,7 @@ void CaptureEventProcessor::ProcessEvent(const CaptureEvent& event) {
 
 void CaptureEventProcessor::ProcessSchedulingSlice(const SchedulingSlice& scheduling_slice) {
   TimerInfo timer_info;
-  timer_info.set_start(scheduling_slice.in_timestamp_ns());
+  timer_info.set_start(scheduling_slice.out_timestamp_ns() - scheduling_slice.duration_ns());
   timer_info.set_end(scheduling_slice.out_timestamp_ns());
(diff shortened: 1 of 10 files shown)


dpallotti:
By the way, any objections, or downsides I missed?

OrbitGrpcProtos/capture.proto
```diff
@@ -45,29 +45,43 @@ message CaptureOptions {
   bool enable_introspection = 9;
 }
 
+// For CaptureEvents with a duration, excluding for now GPU-related ones, we
```
dimitry-, line 48: why excluding GPU-related ones?

dpallotti: A couple of reasons:
- I don't want to touch the events coming from the Vulkan layer and interfere with Florian's work.
- As for GPU jobs, Florian is also working with them. Also, they have three durations and I didn't want to deal with that for now.
As they are low frequency compared to other events, I just let them alone.

dimitry- (approved):

florian-kuebler (approved):
LGTM, thanks for addressing this.
I agree that we should wait with the GPU events. I can have a look as soon as I am done with my work there.

ronaldfw (approved):
LGTM!

PR #1670, opened 2021-01-14
Optimize EventTrack: only iterate on the CallstackEvent in the visible time range

dpallotti:
#### Optimize `EventTrack`: don't iterate over all events in `CallstackData`
Only iterate on the `CallstackEvent`s in the visible range.
We already had `std::map`s in place to do this, I guess something was lost
during some old refactoring.
Of course this doesn't make a difference when completely zoomed out, but
otherwise it makes the capture windows so much smoother with long captures.
#### Address some lint warnings in `EventTrack`
In particular, I removed some methods that shadowed an identical implementation
in `Track`. But I left `SetPos` as its implementation is different.

### DIFF
--- a/src/OrbitClientData/CallstackData.cpp
+++ b/src/OrbitClientData/CallstackData.cpp
@@ -105,23 +105,39 @@ std::vector<CallstackEvent> CallstackData::GetCallstackEventsOfTidInTimeRange(
 void CallstackData::ForEachCallstackEvent(
     const std::function<void(const orbit_client_protos::CallstackEvent&)>& action) const {
   std::lock_guard lock(mutex_);
-  for (const auto& tid_and_events : callstack_events_by_tid_) {
-    for (const auto& time_and_event : tid_and_events.second) {
-      action(time_and_event.second);
+  for (const auto& [unused_tid, events] : callstack_events_by_tid_) {
+    for (const auto& [unused_timestamp, event] : events) {
+      action(event);
     }
   }
 }
 
-void CallstackData::ForEachCallstackEventOfTid(
-    int32_t tid,
+void CallstackData::ForEachCallstackEventInTimeRange(
+    uint64_t min_timestamp, uint64_t max_timestamp,
     const std::function<void(const orbit_client_protos::CallstackEvent&)>& action) const {
   std::lock_guard lock(mutex_);
+  CHECK(min_timestamp <= max_timestamp);
+  for (const auto& [unused_tid, events] : callstack_events_by_tid_) {
+    for (auto event_it = events.lower_bound(min_timestamp);
+         event_it != events.upper_bound(max_timestamp); ++event_it) {
+      action(event_it->second);
+    }
+  }
+}
+
+void CallstackData::ForEachCallstackEventOfTidInTimeRange(
+    int32_t tid, uint64_t min_timestamp, uint64_t max_timestamp,
+    const std::function<void(const orbit_client_protos::CallstackEvent&)>& action) const {
+  std::lock_guard lock(mutex_);
+  CHECK(min_timestamp <= max_timestamp);
   const auto& tid_and_events_it = callstack_events_by_tid_.find(tid);
   if (tid_and_events_it == callstack_events_by_tid_.end()) {
     return;
   }
-  for (const auto& time_and_event : tid_and_events_it->second) {
-    action(time_and_event.second);
+  const auto& events = tid_and_events_it->second;
+  for (auto event_it = events.lower_bound(min_timestamp);
+       event_it != events.upper_bound(max_timestamp); ++event_it) {
+    action(event_it->second);
   }
 }
 
--- a/src/OrbitGl/EventTrack.cpp
+++ b/src/OrbitGl/EventTrack.cpp
@@ -100,18 +100,18 @@ void EventTrack::UpdatePrimitives(uint64_t min_tick, uint64_t max_tick, PickingM
   if (!picking) {
     // Sampling Events
     auto action_on_callstack_events = [=](const orbit_client_protos::CallstackEvent& event) {
-      uint64_t time = event.time();
-      if (time > min_tick && time < max_tick) {
-        Vec2 pos(time_graph_->GetWorldFromTick(time), pos_[1]);
-        batcher->AddVerticalLine(pos, -track_height, z, kWhite);
-      }
+      const uint64_t time = event.time();
+      CHECK(time >= min_tick && time <= max_tick);
+      Vec2 pos(time_graph_->GetWorldFromTick(time), pos_[1]);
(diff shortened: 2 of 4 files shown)


src/OrbitGl/EventTrack.cpp
```diff
@@ -92,31 +92,30 @@ void EventTrack::UpdatePrimitives(uint64_t min_tick, uint64_t max_tick, PickingM
...
+  static const Color kWhite(255, 255, 255, 255);
+  static const Color kGreenSelection(0, 255, 0, 255);
   const CaptureData* capture_data = time_graph_->GetCaptureData();
   CHECK(capture_data != nullptr);
 
   if (!picking) {
     // Sampling Events
     auto action_on_callstack_events = [=](const orbit_client_protos::CallstackEvent& event) {
-      uint64_t time = event.time();
-      if (time > min_tick && time < max_tick) {
```
florian-kuebler, line 104: Do you want to add a CHECK for being in the range?

src/OrbitGl/EventTrack.cpp
```diff
@@ -128,30 +127,24 @@ void EventTrack::UpdatePrimitives(uint64_t min_tick, uint64_t max_tick, PickingM
     constexpr const float kPickingBoxOffset = (kPickingBoxWidth - 1.0f) / 2.0f;
 
     auto action_on_callstack_events = [=](const orbit_client_protos::CallstackEvent& event) {
-      uint64_t time = event.time();
-      if (time > min_tick && time < max_tick) {
```
florian-kuebler, line 132: Again, maybe add a CHECK for now?

src/OrbitGl/EventTrack.cpp
```diff
@@ -251,7 +242,7 @@ std::string EventTrack::GetSampleTooltip(PickingId id) const {
   static const std::string unknown_return_text = "Function call information missing";
 
   auto user_data = time_graph_->GetBatcher().GetUserData(id);
-  if (!user_data || !user_data->custom_data_) {
+  if (user_data == nullptr || user_data->custom_data_ == nullptr) {
```
florian-kuebler, line 245: Thanks, but I guess we should "briefly" rediscuss. AFAIK we have not decided on anything here, or well we said both is fine. But I know that clang-tidy complains.

florian-kuebler (approved):
Thanks a lot for this change. It looks good. Feel free to ignore my comments.

src/OrbitClientData/CallstackData.cpp
```diff
@@ -105,23 +105,37 @@ std::vector<CallstackEvent> CallstackData::GetCallstackEventsOfTidInTimeRange(
...
+  }
+}
+
+void CallstackData::ForEachCallstackEventInTimeRange(
+    uint64_t min_timestamp, uint64_t max_timestamp,
+    const std::function<void(const orbit_client_protos::CallstackEvent&)>& action) const {
+  std::lock_guard lock(mutex_);
+  for (const auto& [unused_tid, events] : callstack_events_by_tid_) {
+    for (auto event_it = events.lower_bound(min_timestamp);
+         event_it != events.upper_bound(max_timestamp); ++event_it) {
```
beckerhe, line 121: If max_timestamp is smaller than min_timestamp this could potentially be an endless loop. Maybe we should add a check here? Or use `std::min(min_timestamp, max_timestamp)` and `std::max(min_timestamp, max_timestamp)`?

src/OrbitClientData/CallstackData.cpp
```diff
@@ -105,23 +105,37 @@ std::vector<CallstackEvent> CallstackData::GetCallstackEventsOfTidInTimeRange(
...
   std::lock_guard lock(mutex_);
   const auto& tid_and_events_it = callstack_events_by_tid_.find(tid);
   if (tid_and_events_it == callstack_events_by_tid_.end()) {
     return;
   }
-  for (const auto& time_and_event : tid_and_events_it->second) {
-    action(time_and_event.second);
+  const auto& events = tid_and_events_it->second;
+  for (auto event_it = events.lower_bound(min_timestamp);
+       event_it != events.upper_bound(max_timestamp); ++event_it) {
```
beckerhe, line 137: same as before.

src/OrbitGl/EventTrack.cpp
```diff
@@ -92,31 +92,30 @@ void EventTrack::UpdatePrimitives(uint64_t min_tick, uint64_t max_tick, PickingM
   float track_height = layout.GetEventTrackHeight();
   const bool picking = picking_mode != PickingMode::kNone;
 
-  const Color kWhite(255, 255, 255, 255);
-  const Color kGreenSelection(0, 255, 0, 255);
+  static const Color kWhite(255, 255, 255, 255);
+  static const Color kGreenSelection(0, 255, 0, 255);
```
beckerhe, line 96: Is that a clang-tidy suggested fix? Performance wise I would assume that this might be detremental since the static initialization needs to be synchronized between threads.

src/OrbitGl/EventTrack.cpp
```diff
@@ -128,30 +127,24 @@ void EventTrack::UpdatePrimitives(uint64_t min_tick, uint64_t max_tick, PickingM
...
-        auto user_data = std::make_unique<PickingUserData>(
-            nullptr, [&](PickingId id) -> std::string { return GetSampleTooltip(id); });
-        user_data->custom_data_ = &event;
-        batcher->AddShadedBox(pos, size, z, kGreenSelection, std::move(user_data));
-      }
+      const uint64_t time = event.time();
+      Vec2 pos(time_graph_->GetWorldFromTick(time) - kPickingBoxOffset, pos_[1] - track_height + 1);
+      Vec2 size(kPickingBoxWidth, track_height);
+      auto user_data = std::make_unique<PickingUserData>(
+          nullptr, [&](PickingId id) -> std::string { return GetSampleTooltip(id); });
```
beckerhe, line 134: Nit: `[&]` -> `[this]` At least to me that helps to understand that `GetSampleTooltip` is a member function

src/OrbitGl/EventTrack.cpp
beckerhe: In practice this is probably not noticeable. I just wanted to mention it, feel free to ignore.

beckerhe (approved):
Minor comments, otherwise LGTM! Very cool change! Thank you for that!

karupayun (approved):
LGTM!! :) :)

src/OrbitGl/EventTrack.cpp
florian-kuebler: Agree, but then let us decide on this! We should bring this up in our next meeting.

ronaldfw:
Nice, glad to see this!

PR #1671, opened 2021-01-14
Fix Crash on "go to disassembly".

florian-kuebler:
The recent UI change resulted in App having one process (the selected process).
In DataManager the process list is not added anymore, as this is obsolete.

App::Disassemly however still used DataManager to retrieve the process.

This change the usage to `process_` instead of using `DataManager::GetProcessByPid`
and removes this function.

Bug: http://b/177499048.
Test: Go to disassembly.


antonrohr (approved):
Looks good!

dimitry- (approved):

PR #1673, opened 2021-01-14
Hide manual instrumentation API functions in live view.

florian-kuebler:
The internal API functions for manual instrumentation show up in
the live functions view. This is missleading, as they are not
the results of manual instrumentation.

Also related: Adding a frame track for those is currently leading
to a crash. Hiding those, would also prevent the user for adding
a frame track (at least from this view).

Test: Capture OrbitTest.
Bug: http://b/177432700.


antonrohr:
related: b/177522433

pierricgimmig (approved):
Thanks for this Florian, LGTM.

antonrohr (approved):
Since Pierric agrees, LGTM from my side!

PR #1674, opened 2021-01-14
[release/1.58] Fix crash when enabling frame tracks for manual instrumentation functions

ronaldfw:
Functions dynamically instrumented for manual instrumentation (that is,
functions with OrbitType not kNone) show up in the "Live" functions
view (which needs to be changed at some point) and it is possible to
enable a frame track for them. Due to the case distinction in
TimeGraph::ProcessTimer that checks for those special functions with
OrbitType != kNone, when adding a frame track, frame track timers were
processed incorrectly. This is fixed here.

Tested: Manual test with OrbitTest
Bug: http://b/177432700

### DIFF
--- a/OrbitGl/TimeGraph.cpp
+++ b/OrbitGl/TimeGraph.cpp
@@ -250,7 +250,14 @@ void TimeGraph::ProcessTimer(const TimerInfo& timer_info, const FunctionInfo* fu
     capture_max_timestamp_ = timer_info.end();
   }
 
-  if (function != nullptr && function->orbit_type() != FunctionInfo::kNone) {
+  // Functions for manual instrumentation scopes and tracked values are those with orbit_type() !=
+  // FunctionInfo::kNone. All proper timers for these have timer_info.type() == TimerInfo::kNone. It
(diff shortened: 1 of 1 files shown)


florian-kuebler (approved):
Thanks for the change. LGTM.

OrbitGl/TimeGraph.cpp
```diff
@@ -250,7 +250,7 @@ void TimeGraph::ProcessTimer(const TimerInfo& timer_info, const FunctionInfo* fu
     capture_max_timestamp_ = timer_info.end();
   }
 
-  if (function != nullptr && function->orbit_type() != FunctionInfo::kNone) {
+  if (function != nullptr && function->orbit_type() != FunctionInfo::kNone && timer_info.type() == TimerInfo::kNone) {
```
karupayun, line 253: As discussed online, better if we have a comment for this "cryptic" condition

karupayun (approved):
LGTM!

OrbitGl/TimeGraph.cpp
ronaldfw: Done, good suggestion!

PR #1675, opened 2021-01-14
[rebase to master] Fix crash when enabling frame tracks for manual instrumentation functions

ronaldfw:
Functions dynamically instrumented for manual instrumentation (that is,
functions with OrbitType not kNone) show up in the "Live" functions
view (which needs to be changed at some point) and it is possible to
enable a frame track for them. Due to the case distinction in
TimeGraph::ProcessTimer that checks for those special functions with
OrbitType != kNone, when adding a frame track, frame track timers were
processed incorrectly. This is fixed here.

Tested: Manual test with OrbitTest
Bug: http://b/177432700

### DIFF
--- a/src/OrbitGl/TimeGraph.cpp
+++ b/src/OrbitGl/TimeGraph.cpp
@@ -250,7 +250,14 @@ void TimeGraph::ProcessTimer(const TimerInfo& timer_info, const FunctionInfo* fu
     capture_max_timestamp_ = timer_info.end();
   }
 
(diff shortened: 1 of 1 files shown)


beckerhe (approved):
LGTM, just one thing: Could you remove the `[release/1.58]` prefix from the commit message? Thanks.

ronaldfw:
> LGTM, just one thing: Could you remove the `[release/1.58]` prefix from the commit message? Thanks.

Done, thanks for spotting this!

PR #1676, opened 2021-01-14
Draw frametrack text in the correct layer

karupayun:
We were drawing Frame-track-text as text on top in the capture view. So,
this text was drawn on top of other pinned/moving tracks and also over
the green overlay and iterators.

This PR erased completely the text z-layer because it makes confusion
and draw some text in the new track-text layer and other in UI-text
layer.

It solved http://b/177420768.

Test: Load a capture, enable a frame track, add iterators/overlays and
move tracks one on top of another several times.

### DIFF
--- a/src/OrbitGl/GlCanvas.cpp
+++ b/src/OrbitGl/GlCanvas.cpp
@@ -27,9 +27,9 @@ float GlCanvas::kZValueTrack = 0.01f;
 float GlCanvas::kZValueEventBar = 0.03f;
 float GlCanvas::kZValueBox = 0.05f;
 float GlCanvas::kZValueEvent = 0.07f;
+float GlCanvas::kZValueTrackText = 0.09f;
 float GlCanvas::kZValueOverlay = 0.43f;
 float GlCanvas::kZValueOverlayTextBackground = 0.45f;
-float GlCanvas::kZValueText = 0.47f;
 float GlCanvas::kZValueEventBarPicking = 0.49f;
 float GlCanvas::kZValueUi = 0.61f;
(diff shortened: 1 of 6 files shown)


src/OrbitGl/GlCanvas.cpp
```diff
@@ -43,8 +43,8 @@ float GlCanvas::kZValueSlider = 0.89f;
 float GlCanvas::kZOffsetMovingTack = 0.1f;
 float GlCanvas::kZOffsetPinnedTrack = 0.2f;
 
-// Max Number of layers: 16 original, 4 for moving track, 4 for pinned Track, 4 epsilon in Slider
-unsigned GlCanvas::kMaxNumberRealZLayers = 16 + 4 + 4 + 4;
+// Max Number of layers: 16 original, 5 for moving track, 5 for pinned Track, 4 epsilon in Slider
+unsigned GlCanvas::kMaxNumberRealZLayers = 16 + 5 + 5 + 4;
```
ronaldfw, line 47: Nit: Make these named constants, then you can get rid of some of the comment. 

constexpr unsigned kNumberOriginalLayers = 16;
constexpr unsigned kNumberLayersForMovingTracks = 5;
constexpr unsigned kNumberLayersForPinnedTracks = 5;
constexpr unsigned kSliderEpsilon = 4;

ronaldfw (approved):
LGTM

src/OrbitGl/GlCanvas.cpp
karupayun: Changed!

reichlfl (approved):
LGTM, thanks for fixing this so quickly!

PR #1678, opened 2021-01-14
Introduce function_id for instrumented functions

dimitry-:
Previously the client code used absolute function address to identify
the instrumented function in timers and in some other places in UI.

Since using the address assumes that the function address in memory is
fixed (which might not be true) this has been replaced with function_id
for instrumented functions.

Changed code relying on function absolute address where it is not needed
and replaced it with function_id.

Test: run unit tests
Test: hook functions run capture, create iterators and frame tracks
Bug: http://b/177525225
Bug: http://b/175865913

### DIFF
--- a/src/OrbitCaptureClient/CaptureClient.cpp
+++ b/src/OrbitCaptureClient/CaptureClient.cpp
@@ -53,7 +53,7 @@ ErrorMessageOr<void> CaptureClient::StartCapture(
     ThreadPool* thread_pool, const ProcessData& process,
     const orbit_client_data::ModuleManager& module_manager,
     absl::flat_hash_map<uint64_t, FunctionInfo> selected_functions,
-    TracepointInfoSet selected_tracepoints, UserDefinedCaptureData user_defined_capture_data,
+    TracepointInfoSet selected_tracepoints, absl::flat_hash_set<uint64_t> frame_track_function_ids,
     bool collect_thread_state, bool enable_introspection) {
   absl::MutexLock lock(&state_mutex_);
   if (state_ != State::kStopped) {
@@ -72,10 +72,10 @@ ErrorMessageOr<void> CaptureClient::StartCapture(
 
   thread_pool->Schedule([this, process = std::move(process_copy), &module_manager,
                          selected_functions = std::move(selected_functions), selected_tracepoints,
-                         user_defined_capture_data = std::move(user_defined_capture_data),
+                         frame_track_function_ids = std::move(frame_track_function_ids),
                          collect_thread_state, enable_introspection]() mutable {
     Capture(std::move(process), module_manager, std::move(selected_functions),
-            std::move(selected_tracepoints), std::move(user_defined_capture_data),
+            std::move(selected_tracepoints), std::move(frame_track_function_ids),
             collect_thread_state, enable_introspection);
   });
 
@@ -86,7 +86,7 @@ void CaptureClient::Capture(ProcessData&& process,
                             const orbit_client_data::ModuleManager& module_manager,
                             absl::flat_hash_map<uint64_t, FunctionInfo> selected_functions,
                             TracepointInfoSet selected_tracepoints,
-                            UserDefinedCaptureData user_defined_capture_data,
+                            absl::flat_hash_set<uint64_t> frame_track_function_ids,
                             bool collect_thread_state, bool enable_introspection) {
   ORBIT_SCOPE_FUNCTION;
   writes_done_failed_ = false;
@@ -117,14 +117,14 @@ void CaptureClient::Capture(ProcessData&& process,
 
   capture_options->set_trace_thread_state(collect_thread_state);
   capture_options->set_trace_gpu_driver(true);
-  for (const auto& [absolute_address, function] : selected_functions) {
+  for (const auto& [function_id, function] : selected_functions) {
     CaptureOptions::InstrumentedFunction* instrumented_function =
         capture_options->add_instrumented_functions();
     instrumented_function->set_file_path(function.loaded_module_path());
     const ModuleData* module = module_manager.GetModuleByPath(function.loaded_module_path());
     CHECK(module != nullptr);
     instrumented_function->set_file_offset(function_utils::Offset(function, *module));
-    instrumented_function->set_absolute_address(absolute_address);
+    instrumented_function->set_function_id(function_id);
     instrumented_function->set_function_type(
         InstrumentedFunctionTypeFromOrbitType(function.orbit_type()));
   }
@@ -165,7 +165,7 @@ void CaptureClient::Capture(ProcessData&& process,
 
   capture_listener_->OnCaptureStarted(std::move(process), std::move(selected_functions),
                                       std::move(selected_tracepoints),
-                                      std::move(user_defined_capture_data));
+                                      std::move(frame_track_function_ids));
 
   while (!writes_done_failed_ && !try_abort_) {
     CaptureResponse response;
--- a/src/OrbitCaptureClient/CaptureEventProcessor.cpp
+++ b/src/OrbitCaptureClient/CaptureEventProcessor.cpp
@@ -33,6 +33,8 @@ using orbit_grpc_protos::SchedulingSlice;
 using orbit_grpc_protos::ThreadName;
 using orbit_grpc_protos::ThreadStateSlice;
 
+static constexpr uint64_t kInvalidFunctionId = 0;
+
 void CaptureEventProcessor::ProcessEvent(const CaptureEvent& event) {
   switch (event.event_case()) {
     case CaptureEvent::kSchedulingSlice:
@@ -126,7 +128,7 @@ void CaptureEventProcessor::ProcessFunctionCall(const FunctionCall& function_cal
   timer_info.set_start(function_call.end_timestamp_ns() - function_call.duration_ns());
   timer_info.set_end(function_call.end_timestamp_ns());
   timer_info.set_depth(static_cast<uint8_t>(function_call.depth()));
-  timer_info.set_function_address(function_call.absolute_address());
+  timer_info.set_function_id(function_call.function_id());
   timer_info.set_user_data_key(function_call.return_value());
   timer_info.set_processor(-1);
   timer_info.set_type(TimerInfo::kNone);
@@ -146,7 +148,7 @@ void CaptureEventProcessor::ProcessIntrospectionScope(
   timer_info.set_start(introspection_scope.end_timestamp_ns() - introspection_scope.duration_ns());
   timer_info.set_end(introspection_scope.end_timestamp_ns());
   timer_info.set_depth(static_cast<uint8_t>(introspection_scope.depth()));
-  timer_info.set_function_address(0);  // function address n/a, set to invalid value
+  timer_info.set_function_id(kInvalidFunctionId);  // function id n/a, set to 0
   timer_info.set_processor(-1);        // cpu info not available, set to invalid value
   timer_info.set_type(TimerInfo::kIntrospection);
   timer_info.mutable_registers()->CopyFrom(introspection_scope.registers());
--- a/src/OrbitClientGgp/ClientGgp.cpp
+++ b/src/OrbitClientGgp/ClientGgp.cpp
@@ -93,7 +93,7 @@ bool ClientGgp::RequestStartCapture(ThreadPool* thread_pool) {
   bool enable_introspection = false;
   ErrorMessageOr<void> result = capture_client_->StartCapture(
       thread_pool, target_process_, module_manager_, selected_functions_, selected_tracepoints,
-      UserDefinedCaptureData(), collect_thread_state, enable_introspection);
+      absl::flat_hash_set<uint64_t>{}, collect_thread_state, enable_introspection);
 
   if (result.has_error()) {
     ERROR("Error starting capture: %s", result.error().message());
@@ -243,13 +243,13 @@ std::string ClientGgp::SelectedFunctionMatch(const FunctionInfo& func) {
 }
 
 absl::flat_hash_map<uint64_t, FunctionInfo> ClientGgp::GetSelectedFunctions() {
+  uint64_t function_id = 1;
   absl::flat_hash_map<uint64_t, FunctionInfo> selected_functions;
   absl::flat_hash_set<std::string> capture_functions_used;
   for (const FunctionInfo* func : main_module_->GetFunctions()) {
     const std::string& selected_function_match = SelectedFunctionMatch(*func);
     if (!selected_function_match.empty()) {
-      uint64_t address = function_utils::GetAbsoluteAddress(*func, target_process_, *main_module_);
-      selected_functions[address] = *func;
+      selected_functions[function_id++] = *func;
       if (!capture_functions_used.contains(selected_function_match)) {
         capture_functions_used.insert(selected_function_match);
       }
@@ -276,9 +276,10 @@ void ClientGgp::ProcessTimer(const TimerInfo& timer_info) { timer_infos_.push_ba
 void ClientGgp::OnCaptureStarted(
     ProcessData&& process,
     absl::flat_hash_map<uint64_t, orbit_client_protos::FunctionInfo> selected_functions,
-    TracepointInfoSet selected_tracepoints, UserDefinedCaptureData user_defined_capture_data) {
+    TracepointInfoSet selected_tracepoints,
+    absl::flat_hash_set<uint64_t> frame_track_function_ids) {
   capture_data_ = CaptureData(std::move(process), &module_manager_, std::move(selected_functions),
-                              std::move(selected_tracepoints), user_defined_capture_data);
+                              std::move(selected_tracepoints), std::move(frame_track_function_ids));
   LOG("Capture started");
 }
 
@@ -298,9 +299,9 @@ void ClientGgp::OnCaptureFailed(ErrorMessage error_message) {
 }
 
 void ClientGgp::OnTimer(const orbit_client_protos::TimerInfo& timer_info) {
-  if (timer_info.function_address() > 0) {
+  if (timer_info.function_id() != 0) {
     const FunctionInfo* func =
-        GetCaptureData().FindFunctionByAddress(timer_info.function_address(), false);
+        GetCaptureData().GetInstrumentedFunctionById(timer_info.function_id());
     // For timers, the function must be present in the process
     CHECK(func != nullptr);
     uint64_t elapsed_nanos = timer_info.end() - timer_info.start();
--- a/src/OrbitClientGgp/include/OrbitClientGgp/ClientGgp.h
+++ b/src/OrbitClientGgp/include/OrbitClientGgp/ClientGgp.h
@@ -52,7 +52,7 @@ class ClientGgp final : public CaptureListener {
       ProcessData&& process,
       absl::flat_hash_map<uint64_t, orbit_client_protos::FunctionInfo> selected_functions,
       TracepointInfoSet selected_tracepoints,
-      UserDefinedCaptureData user_defined_capture_data) override;
+      absl::flat_hash_set<uint64_t> frame_track_function_ids) override;
   void OnCaptureComplete() override;
   void OnCaptureCancelled() override;
   void OnCaptureFailed(ErrorMessage error_message) override;
--- a/src/OrbitClientModel/CaptureData.cpp
+++ b/src/OrbitClientModel/CaptureData.cpp
@@ -65,12 +65,25 @@ void CaptureData::UpdateFunctionStats(const FunctionInfo& function, uint64_t ela
   }
 }
 
-const FunctionInfo* CaptureData::GetSelectedFunction(uint64_t function_address) const {
-  auto selected_functions_it = selected_functions_.find(function_address);
-  if (selected_functions_it == selected_functions_.end()) {
+const FunctionInfo* CaptureData::GetInstrumentedFunctionById(uint64_t function_id) const {
+  auto selected_functions_it = instrumented_functions_.find(function_id);
+  if (selected_functions_it == instrumented_functions_.end()) {
     return nullptr;
   }
   return &selected_functions_it->second;
+}
+
+std::optional<uint64_t> CaptureData::FindInstrumentedFunctionIdSlow(
+    const orbit_client_protos::FunctionInfo& function) const {
+  for (const auto& it : instrumented_functions_) {
+    const auto& target_function = it.second;
+    if (target_function.file() == function.file() &&
+        target_function.address() == function.address()) {
+      return it.first;
+    }
+  }
+
+  return std::nullopt;
 }
 
 const LinuxAddressInfo* CaptureData::GetAddressInfo(uint64_t absolute_address) const {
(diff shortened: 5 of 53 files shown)


src/OrbitClientGgp/include/OrbitClientGgp/ClientGgp.h
```diff
@@ -80,6 +80,7 @@ class ClientGgp final : public CaptureListener {
   std::shared_ptr<grpc::Channel> grpc_channel_;
   ProcessData target_process_;
   orbit_client_data::ModuleManager module_manager_;
+  std::atomic<uint64_t> next_function_id_;
```
florian-kuebler, line 83: Can you add a comment here? What is the next function representing?

src/OrbitGl/LiveFunctionsController.cpp
```diff
@@ -194,31 +194,25 @@ void LiveFunctionsController::OnDeleteButton(uint64_t id) {
   if (id == id_to_select_ && !current_textboxes_.empty()) {
     id_to_select_ = current_textboxes_.begin()->first;
   } else if (current_textboxes_.empty()) {
-    // TODO: Not sure this is a good idea...
     id_to_select_ = 0;
```
ronaldfw, line 197: kInvalidFunctionId?

And, are these ids now function_ids?

src/OrbitClientModel/CaptureData.cpp
```diff
@@ -65,14 +65,27 @@ void CaptureData::UpdateFunctionStats(const FunctionInfo& function, uint64_t ela
...
+  auto selected_functions_it = instrumented_functions_.find(function_id);
+  if (selected_functions_it == instrumented_functions_.end()) {
     return nullptr;
   }
   return &selected_functions_it->second;
 }
 
+std::optional<uint64_t> CaptureData::GetInstrumentedFunctionId(
+    const orbit_client_protos::FunctionInfo& function) const {
+  for (const auto& it : instrumented_functions_) {
```
florian-kuebler, line 78: Can this become a performance bottleneck? If so, we could arrange the target functions sorted by address and do a binary search.

src/OrbitClientModel/include/OrbitClientModel/CaptureData.h
```diff
@@ -239,7 +243,7 @@ class CaptureData {
 
   std::chrono::system_clock::time_point capture_start_time_ = std::chrono::system_clock::now();
 
-  UserDefinedCaptureData user_defined_capture_data_;
+  absl::flat_hash_set<uint64_t> frame_track_function_ids_;
```
florian-kuebler, line 246: As said above, I am really unsure wether we want this change or not. UserDefinedCaptureData can change even after a capture was taken. This was encapsulated by having this separate class. Also IIRC we planned to add additional stuff there.

src/OrbitClientProtos/capture_data.proto
```diff
@@ -155,7 +156,7 @@ message TimerInfo {
 
   int32 processor = 7;
   uint64 callstack_id = 8;
-  uint64 function_address = 9;
```
florian-kuebler, line 158: should we rather use a new id for this?

src/OrbitGl/App.cpp
```diff
@@ -863,24 +864,28 @@ bool OrbitApp::StartCapture() {
   selected_functions.insert(selected_functions.end(), orbit_functions.begin(),
                             orbit_functions.end());
 
+  UserDefinedCaptureData user_defined_capture_data = data_manager_->user_defined_capture_data();
+
   absl::flat_hash_map<uint64_t, FunctionInfo> selected_functions_map;
+  absl::flat_hash_set<uint64_t> frame_track_function_ids;
+  uint64_t function_id = 1;
```
florian-kuebler, line 871: Maybe add a comment, that you are starting to compute the function_id's here. Also that 0 is a special case.

ronaldfw (approved):
LGTM! 

We should discuss how we expand ids to other timers like GPU debug label timers and manual instrumentation timers, as we also want to be able to create frame tracks, etc. for them.

src/OrbitGl/FunctionsDataView.cpp
```diff
@@ -58,8 +58,21 @@ bool FunctionsDataView::ShouldShowSelectedFunctionIcon(OrbitApp* app,
 }
 
 bool FunctionsDataView::ShouldShowFrameTrackIcon(OrbitApp* app, const FunctionInfo& function) {
-  return app->IsFrameTrackEnabled(function) ||
-         (app->HasCaptureData() && app->HasFrameTrackInCaptureData(function));
+  if (app->IsFrameTrackEnabled(function)) {
```
florian-kuebler, line 61: @ronaldfw, @karupayun Is this still as intended? I remember the very recent changes here.

florian-kuebler (commented):
Thanks for the change! In general LGTM. However, I'd really like to discuss the change along with UserDefinedCaptureData and either @ronaldfw or @karupayun should verify that the code to show the "F" for frame tracks is still correct.

src/OrbitCaptureClient/CaptureClient.cpp
```diff
@@ -53,7 +53,7 @@ ErrorMessageOr<void> CaptureClient::StartCapture(
     ThreadPool* thread_pool, const ProcessData& process,
     const orbit_client_data::ModuleManager& module_manager,
     absl::flat_hash_map<uint64_t, FunctionInfo> selected_functions,
-    TracepointInfoSet selected_tracepoints, UserDefinedCaptureData user_defined_capture_data,
```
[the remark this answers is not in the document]
dimitry-: we want to translate function set to a set of intrumented function ids at the time they become available. Note also that this provides consistency by design since it is clear that frame_track_ids should be in selected function map.

src/OrbitCaptureClient/CaptureEventProcessor.cpp
```diff
@@ -146,7 +146,7 @@ void CaptureEventProcessor::ProcessIntrospectionScope(
   timer_info.set_start(introspection_scope.begin_timestamp_ns());
   timer_info.set_end(introspection_scope.end_timestamp_ns());
   timer_info.set_depth(static_cast<uint8_t>(introspection_scope.depth()));
-  timer_info.set_function_address(0);  // function address n/a, set to invalid value
+  timer_info.set_function_id(0);       // function id n/a, set to 0
```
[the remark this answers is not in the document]
dimitry-: I think I need to find a better place for the constant. It is currently in OrbitGl and this component does not depend on it.

src/OrbitClientGgp/include/OrbitClientGgp/ClientGgp.h
dimitry-: This is actually not used - removed - also fixed the id assignment code to use sequential id instead of address.

src/OrbitClientModel/CaptureData.cpp
dimitry-: At the moment this function is only used when we select a new function. never during capture or when drawing the track. I would rather consider not using it than optimizing this implementation.

src/OrbitClientModel/include/OrbitClientModel/CaptureData.h
dimitry-: I think we want this change. Note that we still use UserDefinedCaptureData and function set in datamanager, but for capturedata this is not needed because here we have ids for these functions.

src/OrbitClientProtos/capture_data.proto
dimitry-: it doesn't break anything and actually it is rather a clarification from the service's point of view - this was always an id but for some reason it was called function_adddress.

src/OrbitGl/App.cpp
dimitry-: Done. and I it is now starting with 7 - so that it is not confused with an index :)

src/OrbitGl/App.cpp
```diff
@@ -1719,69 +1725,97 @@ void OrbitApp::DisableFrameTrack(const FunctionInfo& function) {
...
+void OrbitApp::AddFrameTrack(const FunctionInfo& instrumented_function) {
   if (!HasCaptureData()) {
     return;
   }
-  const uint64_t function_address = GetCaptureData().GetAbsoluteAddress(function);
-  if (GetCaptureData().GetSelectedFunction(function_address) == nullptr) {
+
+  auto instrumented_function_id = GetCaptureData().GetInstrumentedFunctionId(instrumented_function);
+  if (!instrumented_function_id) {
+    return;
```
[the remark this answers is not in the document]
dimitry-: I think it happens when adding frame track on freshly selected function (before starting new capture but having previous capture already available for a different set of instrumented functions).

src/OrbitGl/App.cpp
```diff
@@ -1719,69 +1725,97 @@ void OrbitApp::DisableFrameTrack(const FunctionInfo& function) {
...
   }
 }
 
-void OrbitApp::RemoveFrameTrack(const FunctionInfo& function) {
+void OrbitApp::RemoveFrameTrack(const FunctionInfo& instrumented_function) {
+  CHECK(HasCaptureData());
+  std::optional<uint64_t> instrumented_function_id =
+      GetCaptureData().GetInstrumentedFunctionId(instrumented_function);
+  if (!instrumented_function_id) {
+    return;
```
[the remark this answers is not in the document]
dimitry-: I think the same use-case applies here: removing selected but not (yet) instrumented function.

src/OrbitGl/CallStackDataView.cpp
```diff
@@ -161,7 +161,7 @@ void CallStackDataView::DoFilter() {
     return;
   }
 
-  std::vector<uint32_t> indices;
+  std::vector<uint64_t> indices;
```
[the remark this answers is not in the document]
dimitry-: One view - LiveFunctionView is using ids with map instead of a vector with index (which reminds me that for the vector with an index this should also be uint64_t on lp64). This change is related to change in LiveFunctionsDataView

src/OrbitGl/LiveFunctionsController.cpp
dimitry-: done.

and yes - they are function ids now.

src/OrbitClientModel/CaptureData.cpp
florian-kuebler: Then maybe rename it? e.g. `ComputeInstrumentedFunctionId`/`FindInstrumentedFunctionId` and maybe add a comment?

src/OrbitClientProtos/capture_data.proto
florian-kuebler: Right. Thanks for the clarification.

src/OrbitGl/FrameTrackOnlineProcessor.h
```diff
@@ -22,13 +22,11 @@ class FrameTrackOnlineProcessor {
...
-                     const orbit_client_protos::FunctionInfo& function);
-  void RemoveFrameTrack(const CaptureData& capture_data,
-                        const orbit_client_protos::FunctionInfo& function);
+  void AddFrameTrack(uint64_t function_id);
+  void RemoveFrameTrack(uint64_t function_id);
 
  private:
-  absl::flat_hash_set<uint64_t> current_frame_track_functions_;
+  absl::flat_hash_set<uint64_t> current_frame_track_function_ids_;
   absl::flat_hash_map<uint64_t, uint64_t> previous_timestamp_ns_;
```
dpallotti, line 30: I think it would be beneficial if the name of this field explained what is mapping from and what it's mapping to, it's a bit cryptic like this. E.g., `function_id_to_previous_timestamp_ns_` or `previous_timestamp_ns_of_function_id_`.

src/OrbitGl/LiveFunctionsDataView.cpp
```diff
@@ -128,9 +128,9 @@ void LiveFunctionsDataView::DoSort() {
     return;
   }
   bool ascending = sorting_orders_[sorting_column_] == SortingOrder::kAscending;
-  std::function<bool(int a, int b)> sorter = nullptr;
+  std::function<bool(uint64_t a, uint64_t b)> sorter = nullptr;
 
-  const std::vector<FunctionInfo>& functions = functions_;
+  const auto& functions = functions_;
```
dpallotti, line 133: Can you state the type? Even if this is just a trick for the macros above, it's not immediately clear this is a map.

src/OrbitLinuxTracing/Function.h
```diff
@@ -5,28 +5,35 @@
...
 class Function {
  public:
-  Function(std::string binary_path, uint64_t file_offset, uint64_t virtual_address)
-      : binary_path_{std::move(binary_path)},
-        file_offset_{file_offset},
-        virtual_address_{virtual_address} {}
-
-  const std::string& BinaryPath() const { return binary_path_; }
+  Function(uint64_t id, std::string file_path, uint64_t file_offset)
+      : id_(id), file_path_{std::move(file_path)}, file_offset_{file_offset} {}
```
dpallotti, line 17:
```suggestion
      : id_{id}, file_path_{std::move(file_path)}, file_offset_{file_offset} {}
```
(consistency)

src/OrbitLinuxTracing/Function.h
```diff
@@ -5,28 +5,35 @@
...
+  [[nodiscard]] uint64_t file_offset() const { return file_offset_; }
 
-  uint64_t VirtualAddress() const { return virtual_address_; }
+  bool operator==(const Function& other) {
+    return (this->file_offset_ == other.file_offset_) && (this->file_path_ == other.file_path_);
+  }
 
  private:
-  std::string binary_path_;
+  uint64_t id_;
```
dpallotti, line 28: I think I would prefer `function_id_`. This is not really an id for this specific class, it's an id for the more general "concept" of "function" across the codebase.

src/OrbitLinuxTracing/TracerThread.cpp
```diff
@@ -9,10 +9,11 @@
 #include <absl/meta/type_traits.h>
 #include <absl/strings/str_format.h>
 #include <pthread.h>
-#include <stddef.h>
```
dpallotti, line 12: The reason `<stddef.h>` is used instead of `<cstddef>` is that we use `size_t` and not `std::size_t` and such. I think you are going against IWYU by changing this.

src/OrbitGrpcProtos/capture.proto
```diff
@@ -24,14 +24,16 @@ message CaptureOptions {
   message InstrumentedFunction {
     string file_path = 1;
     uint64 file_offset = 2;
-    uint64 absolute_address = 3;
+    uint64 function_id = 3;
```
dpallotti, line 27: So in the end the absolute address is not sent back?

src/OrbitGrpcProtos/capture.proto
```diff
@@ -24,14 +24,16 @@ message CaptureOptions {
...
 
     enum FunctionType {
       kRegular = 0;
       kTimerStart = 1;
       kTimerStop = 2;
     }
     FunctionType function_type = 4;
+
+    string function_name = 5;
```
dpallotti, line 36: Why is this sent to the service?

src/OrbitGrpcProtos/capture.proto
```diff
@@ -56,7 +58,7 @@ message SchedulingSlice {
 message FunctionCall {
   int32 pid = 1;
   int32 tid = 2;
-  uint64 absolute_address = 3;
+  uint64 function_id = 3;
```
dpallotti, line 61: So in the end the address is not sent back? Are we sure we don't need it?

src/OrbitClientGgp/ClientGgp.cpp
```diff
@@ -92,8 +92,8 @@ bool ClientGgp::RequestStartCapture(ThreadPool* thread_pool) {
   bool collect_thread_state = absl::GetFlag(FLAGS_thread_state);
   bool enable_introspection = false;
   ErrorMessageOr<void> result = capture_client_->StartCapture(
-      thread_pool, target_process_, module_manager_, selected_functions_, selected_tracepoints,
-      UserDefinedCaptureData(), collect_thread_state, enable_introspection);
+      thread_pool, target_process_, module_manager_, selected_functions_, selected_tracepoints, {},
```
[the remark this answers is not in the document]
dimitry-: done.

src/OrbitClientGgp/include/OrbitClientGgp/ClientGgp.h
```diff
@@ -36,7 +36,7 @@
 
 class ClientGgp final : public CaptureListener {
  public:
-  ClientGgp() : main_module_{nullptr} {}
+  ClientGgp() : main_module_{nullptr} {};
```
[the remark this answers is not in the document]
dimitry-: done.

src/OrbitClientModel/CaptureData.cpp
dimitry-: Done.

src/OrbitClientModel/CaptureData.cpp
```diff
@@ -178,14 +191,14 @@ int32_t CaptureData::process_id() const { return process_.pid(); }
 
 std::string CaptureData::process_name() const { return process_.name(); }
 
-void CaptureData::EnableFrameTrack(const FunctionInfo& function) {
-  user_defined_capture_data_.InsertFrameTrack(function);
+void CaptureData::EnableFrameTrack(uint64_t instrumented_function_id) {
+  frame_track_function_ids_.insert(instrumented_function_id);
```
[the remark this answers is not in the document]
dimitry-: user might be able to select the same one twice there is no harm in doing so.

src/OrbitClientProtos/capture_data.proto
```diff
@@ -110,7 +110,8 @@ message CaptureHeader {
 }
 
 message FrameTracksInfo {
-  repeated FunctionInfo frame_track_functions = 1;
+  reserved 1;
```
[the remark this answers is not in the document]
dimitry-: what happens is that when user loads old capture they will not see frame tracks... but I would argue this is minor inconvenience and we shouldn't refuse to load a capture because of that.

src/OrbitGl/App.cpp
```diff
@@ -863,24 +864,30 @@ bool OrbitApp::StartCapture() {
   selected_functions.insert(selected_functions.end(), orbit_functions.begin(),
                             orbit_functions.end());
 
+  UserDefinedCaptureData user_defined_capture_data = data_manager_->user_defined_capture_data();
+
   absl::flat_hash_map<uint64_t, FunctionInfo> selected_functions_map;
+  absl::flat_hash_set<uint64_t> frame_track_function_ids;
+  // We want to keep them small, and non-zero since 0 is reserved for invalid ids.
+  // Not starting with 1 so this is not accidentally used as an index somewhere in the code.
```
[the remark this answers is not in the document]
dimitry-: ok, probably an overkill. changed back to 1.

src/OrbitGl/App.cpp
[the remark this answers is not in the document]
dimitry-: done.

src/OrbitGl/FrameTrackOnlineProcessor.h
dimitry-: renamed

src/OrbitGl/FunctionsDataView.cpp
dimitry-: made it a little bit more readable.

src/OrbitGl/LiveFunctionsController.cpp
dimitry-: ok, now it is invalid (somehow did not change it in previous ps)

src/OrbitGl/LiveFunctionsDataView.cpp
dimitry-: done.

src/OrbitGrpcProtos/capture.proto
dimitry-: we never did send it back, we only sent whatever client sent us in instrumented function list. this did not change - I only renamed it to id.

(and yes - we do not need it in FunctionCall after this change)

src/OrbitLinuxTracing/Function.h
dimitry-: done.

src/OrbitLinuxTracing/Function.h
dimitry-: done.

src/OrbitLinuxTracing/TracerThread.cpp
dimitry-: reverted

src/OrbitGl/App.cpp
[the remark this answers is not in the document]
dpallotti: Maybe also stating when it would happen that we ask for an "instrumented_function" that is not instrumented?

src/OrbitGl/App.cpp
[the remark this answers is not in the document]
dpallotti: As above I guess.

src/OrbitClientModel/CaptureData.cpp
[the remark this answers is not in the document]
dpallotti: But should they be able to? I think it would be a ui error. Or not? At least log an error?

src/OrbitClientProtos/capture_data.proto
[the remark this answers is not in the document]
dpallotti: Makes sense.

src/OrbitGrpcProtos/capture.proto
dpallotti: True.

src/OrbitGrpcProtos/capture.proto
dimitry-: this is to save function_name of the timer in the service side when switching to new capture format.

dpallotti: Hmmm ok, this doesn't really belong to this commit then I guess. Also I didn't really understand: Where does it need to be saved by the service? Why isn't it enough that the service sends back `function_id`s to the client?

dpallotti (approved):
Some minor things but looks good.

src/OrbitGl/FunctionsDataView.cpp
florian-kuebler: Thanks

florian-kuebler (approved):
Let's decide offline on how we proceed with the "UserDefinedCaptureData" or as we said yesterday the ~ "UserCaptureViewSettings".

Otherwise LGTM!

src/OrbitClientModel/CaptureData.cpp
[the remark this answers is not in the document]
dimitry-: ok, added a warning.

src/OrbitGl/App.cpp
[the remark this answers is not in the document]
dimitry-: done.

src/OrbitGl/App.cpp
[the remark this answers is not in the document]
dimitry-: done.

src/OrbitGrpcProtos/capture.proto
dimitry-: agree, this doesn't really belong to this change. removed.

PR #1681, opened 2021-01-15
Copy the Vulkan layer to the build artefacts.

florian-kuebler:
This change prepares the deployment of the Vulkan layer.
When building the Vulkan layer, we now copy the .json file
to the lib directory.
In the Kokoro build script, we now also copy the layer's
shared object file and the json file, such that it will be
part of the resulting artefact .zip file.

Doing so, the layer will be part of the SDK content, and
a new "ggp run --orbit" option can deploy these files
to the gamelet.

Test: Run CI.
Bug: http://b/176814898.


beckerhe (approved):
LGTM! Thanks!

PR #1682, opened 2021-01-15
Fix iterators

dimitry-:
Replacing instrumented function addresses with function_ids
mistakenly changed iterator ids to function ids in
LiveFunctionController, this led to incorrect behaviour
when we add 2 iterators to the same function.

This change fixes it restoring iterator id and changing
iterator_id->function map to iterator_id->function_id map

It also fixes AllNext/AllPrevious behavior which was missed by
previous changed and was using absolute address for lookups.
Now it uses function_ids.

Test: capture add 2 iterators to the same function, click to next/prev
      nextall make sure they work. Delete iterators make sure app does
      not crash.
Bug: http://b/177525225

### DIFF
--- a/src/OrbitGl/LiveFunctionsController.cpp
+++ b/src/OrbitGl/LiveFunctionsController.cpp
@@ -49,7 +49,7 @@ const TextBox* ClosestTo(uint64_t point, const TextBox* box_a, const TextBox* bo
   return box_b;
 }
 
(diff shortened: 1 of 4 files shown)


ronaldfw (approved):
LGTM, thanks for the quick fix.

beckerhe (approved):
LGTM!

PR #1684, opened 2021-01-15
Copy .so from/to "lib" instead of "bin".

florian-kuebler:
As per convention on Linux, shared object files are located under
"lib" rather than "bin". This adjust the src/tgt directory for those
files in the conanfile.py.

This fixes the failing nightly builds due to #1681.

Test: Manually trigger nightly build.


florian-kuebler:
Triggering nightly build before merging...

beckerhe (approved):
If your nightly build is happy, I'm happy as well. :-)

danielfenner (approved):
lgtm

PR #1686, opened 2021-01-18
Fix broken build.

florian-kuebler:
The recent change from "begin timestamp" to "end timestamp" - duration
in the capture events (#1668) was not rebased on #1661, leading
to use begin timestamps that do not exist anymore.

This fixes the use of begin timestamps.

Test: Compile


danielfenner (approved):
lgtm

PR #1687, opened 2021-01-18
Update orbit_tracks E2E test

reichlfl:
Some tracks are not present on the devkit which fails the test. Changed
the MatchTracks test to handle "or" options for tracks and adjusted the
list of tracks.

### DIFF
--- a/contrib/automation_tests/orbit_tracks.py
+++ b/contrib/automation_tests/orbit_tracks.py
@@ -16,8 +16,9 @@ def main(argv):
         ConnectToStadiaInstance(),
         FilterAndSelectFirstProcess(process_filter="hello_ggp"),
         Capture(),
+        # "sdma0" is not present on the DevKits, instead there is "vce0", so this tests for "sdma0 or vce0"
         MatchTracks(expected_names=[
-            "Scheduler", "gfx", "sdma0", "hello_ggp_stand", "hello_ggp_stand", "GgpSwapchain"],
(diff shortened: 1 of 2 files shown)


contrib/automation_tests/orbit_tracks.py
```diff
@@ -16,20 +16,23 @@ def main(argv):
         ConnectToStadiaInstance(),
         FilterAndSelectFirstProcess(process_filter="hello_ggp"),
         Capture(),
+        # "sdma0" is not present on the DevKits, instead there is "vce0", so this tests for "sdma0 or vce0"
         MatchTracks(expected_names=[
-            "Scheduler", "gfx", "sdma0", "hello_ggp_stand", "hello_ggp_stand", "GgpSwapchain"],
+            "Scheduler", ("sdma0", "vce0"), "gfx", "hello_ggp_stand", "hello_ggp_stand"],
```
danielfenner, line 21: Is the change from "GgpSwapchain" to "hello_ggp_stand" intended? There are two "hello_ggp_stand"s now in the list ...

reichlfl: This is intended: There were two "hello_ggp_stand" in the list before (because the main thread is called "hello_ggp_stand", same as the all threads track...). "GgpSwapchain" was removed.

danielfenner: Oh, okay, yes, never mind.

danielfenner (approved):
lgtm

PR #1689, opened 2021-01-18
Fix crash when loading a capture

dimitry-:
Another one missed by absolute_address->function_id change.

Previously capture was only saving function infos for instrumented
functions. After this change it saves functions_ids as well.

Bug: http://b/177800361
Test: capture->save capture->close oribt-> open orbit -> load capture
      make sure it does not crash
Test: run unit-tests

### DIFF
--- a/src/OrbitClientModel/CaptureDeserializer.cpp
+++ b/src/OrbitClientModel/CaptureDeserializer.cpp
@@ -18,12 +18,9 @@
 #include "OrbitBase/MakeUniqueForOverwrite.h"
 #include "OrbitBase/Result.h"
 #include "OrbitClientData/Callstack.h"
-#include "OrbitClientData/FunctionUtils.h"
-#include "OrbitClientData/ModuleData.h"
 #include "OrbitClientData/ModuleManager.h"
 #include "OrbitClientData/ProcessData.h"
 #include "OrbitClientData/TracepointCustom.h"
-#include "OrbitClientData/UserDefinedCaptureData.h"
 #include "absl/strings/str_format.h"
 #include "capture_data.pb.h"
 #include "google/protobuf/io/coded_stream.h"
@@ -159,14 +156,11 @@ void LoadCaptureInfo(const CaptureInfo& capture_info, CaptureListener* capture_l
     return;
   }
 
-  absl::flat_hash_map<uint64_t, orbit_client_protos::FunctionInfo> selected_functions;
-  for (const auto& function : capture_info.selected_functions()) {
-    const auto& module_it = module_map.find(function.loaded_module_path());
-    CHECK(module_it != module_map.end());
-    ModuleData module(module_it->second);
-    uint64_t address = function_utils::GetAbsoluteAddress(function, process, module);
-    selected_functions[address] = function;
+  absl::flat_hash_map<uint64_t, orbit_client_protos::FunctionInfo> instrumented_functions;
+  for (const auto& function : capture_info.instrumented_functions()) {
+    instrumented_functions.insert_or_assign(function.first, function.second);
   }
+
   TracepointInfoSet selected_tracepoints;
   for (const orbit_client_protos::TracepointInfo& tracepoint_info :
        capture_info.tracepoint_infos()) {
@@ -187,7 +181,7 @@ void LoadCaptureInfo(const CaptureInfo& capture_info, CaptureListener* capture_l
     frame_track_function_ids.insert(function_id);
   }
 
-  capture_listener->OnCaptureStarted(std::move(process), std::move(selected_functions),
+  capture_listener->OnCaptureStarted(std::move(process), std::move(instrumented_functions),
                                      std::move(selected_tracepoints),
                                      std::move(frame_track_function_ids));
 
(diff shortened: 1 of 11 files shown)


src/OrbitGl/TimeGraph.cpp
```diff
@@ -74,7 +74,7 @@ void TimeGraph::Clear() {
   std::lock_guard<std::recursive_mutex> lock(mutex_);
 
   batcher_.StartNewFrame();
-  capture_min_timestamp_ = std::numeric_limits<uint64_t>::max();
+  capture_min_timestamp_ = 0;
```
IrinaShkviro, line 77: Does it mean that `capture_min_timestamp_ ` won't be updated after calling this function?

florian-kuebler (approved):
LGTM, thanks!

src/OrbitGl/TimeGraph.cpp
dimitry-: Ok, I see you are referring to UpdateCaptureMinMax. I will revert this part for now, This change was made because by default capture_min_timestamp_ is 0 when class is constructed. I will look into this separately.

IrinaShkviro: ok, thank you!

IrinaShkviro (approved):

src/OrbitClientModel/CaptureDeserializer.cpp
```diff
@@ -159,14 +156,11 @@ void LoadCaptureInfo(const CaptureInfo& capture_info, CaptureListener* capture_l
...
 
-  absl::flat_hash_map<uint64_t, orbit_client_protos::FunctionInfo> selected_functions;
-  for (const auto& function : capture_info.selected_functions()) {
-    const auto& module_it = module_map.find(function.loaded_module_path());
-    CHECK(module_it != module_map.end());
-    ModuleData module(module_it->second);
-    uint64_t address = function_utils::GetAbsoluteAddress(function, process, module);
-    selected_functions[address] = function;
+  absl::flat_hash_map<uint64_t, orbit_client_protos::FunctionInfo> instrumented_functions;
+  for (const auto& function : capture_info.instrumented_functions()) {
```
karupayun, line 160: Given that in the next lines, you are using both of them, why not capture both variables by names using structure bindings?

src/OrbitClientModel/CaptureDeserializerTest.cpp
```diff
@@ -207,24 +209,23 @@ TEST(CaptureDeserializer, LoadCaptureInfoOnCaptureStarted) {
   EXPECT_CALL(listener, OnCaptureStarted(_, _, _, _)).Times(0);
   EXPECT_CALL(listener, OnCaptureStarted(_, _, IsEmpty(), _))
       .Times(1)
-      .WillOnce([selected_function, module_info](
+      .WillOnce([&instrumented_function, kInstrumentedFunctionId](
                     ProcessData&& process,
                     absl::flat_hash_map<uint64_t, orbit_client_protos::FunctionInfo>
-                        actual_selected_functions,
+                        actual_instrumented_functions,
```
karupayun, line 215: actual means current?

src/OrbitGl/TimeGraph.cpp
dimitry-: ok, without this change we have a crash. taking another look.

karupayun (approved):
Thanks Dimitry for the fast solution! :)

BTW, I haven't tested locally.

karupayun (approved):

dimitry-:
Hi, I updated some code in TimeGraph, PTAL.

src/OrbitClientModel/CaptureDeserializerTest.cpp
dimitry-: actual is used to align with GTEST terminology here. EXPECT_* parameters are named acutal and expected.

src/OrbitClientModel/CaptureDeserializer.cpp
dimitry-: done.

florian-kuebler (approved):
Still LGTM, thanks!

src/OrbitGl/TimeGraph.cpp
```diff
@@ -245,6 +245,10 @@ double TimeGraph::GetTime(double ratio) const {
 }
 
 void TimeGraph::ProcessTimer(const TimerInfo& timer_info, const FunctionInfo* function) {
+  if (timer_info.start() < capture_min_timestamp_) {
```
karupayun, line 248: Maybe a stupid question because I'm don't have too much experience with Google C++ Style, but using "min" function here (and "max" after) isn't clearer? (at least less lines of code)

src/OrbitClientModel/CaptureDeserializer.cpp
dimitry-: I reverted this change because it does not sit well with msvc2017 & clang7

karupayun (approved):
LGTM! Also thanks for cleaning TimeGraph a little

IrinaShkviro (approved):

src/OrbitGl/TimeGraph.cpp
dimitry-: done. - this code was modified to use min/max.

PR #1692, opened 2021-01-19
Remove some moves that are not moves

dimitry-:
And some other minor fixes

Test: builds

### DIFF
--- a/src/OrbitQt/orbitmainwindow.cpp
+++ b/src/OrbitQt/orbitmainwindow.cpp
@@ -329,13 +329,13 @@ void OrbitMainWindow::SetupMainWindow(uint32_t font_size) {
   });
 
   app_->SetSamplingReportCallback(
-      [this](DataView* callstack_data_view, std::shared_ptr<SamplingReport> report) {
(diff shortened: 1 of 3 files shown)


beckerhe (approved):
LGTM!

src/OrbitQt/orbitmainwindow.cpp
```diff
@@ -470,7 +470,7 @@ void OrbitMainWindow::SetupCaptureToolbar() {
   // Timer
   toolbar->addWidget(CreateSpacer(toolbar));
   QFontMetrics fm(ui->timerLabel->font());
-  int pixel_width = fm.width("w");
+  int pixel_width = fm.horizontalAdvance("w");
```
florian-kuebler, line 473: This kind of comes out of nothing. Is this part of "minor bugfixes"?

florian-kuebler (approved):
LGTM

src/OrbitQt/orbitmainwindow.cpp
dimitry-: width is deprecated so I replaced it. I do not think this fixes anything. just code cleanup.

PR #1694, opened 2021-01-19
Remove UI flag

antonrohr:
This is the first PR that removes dead code from the old UI. There are also minor refactorings.

Unfortunately here on github it looks like a lot of things changed in main.cpp. There are actually only a few minor changes and I removed one scope ({}) that was not needed anymore. The removal of the scope changes the indentation and that looks like a lot of changes. 

There is still quite some code that should be refactored. I wanted to keep PR smaller, hence you can see this as a first step in an ongoing effort. b/177304549 tracks this.


src/OrbitQt/orbitmainwindow.cpp
```diff
@@ -1302,17 +1227,15 @@ void OrbitMainWindow::OnProcessListUpdated(
...
-  std::optional<orbit_qt::TargetConfiguration> result = std::move(target_configuration_);
-  target_configuration_ = std::nullopt;
-  return result;
+void OrbitMainWindow::ShutdownTarget(const orbit_qt::StadiaTarget& target) {
+  target.GetProcessManager()->SetProcessListUpdateListener(nullptr);
+}
+
+void OrbitMainWindow::ShutdownTarget(const orbit_qt::LocalTarget& target) {
+  target.GetProcessManager()->SetProcessListUpdateListener(nullptr);
+}
```
beckerhe, line 1232: Have you considered adding a member function to the target-classes called `Shutdown` or something similar?

beckerhe (approved):
Minor comments, otherwise LGTM!

src/OrbitQt/orbitmainwindow.cpp
```diff
@@ -765,9 +709,8 @@ void OrbitMainWindow::UpdatePanel(DataViewType a_Type) {
       ui->ModulesList->Refresh();
       break;
     case DataViewType::kProcesses:
-      if (!absl::GetFlag(FLAGS_enable_ui_beta)) {
-        ui->ProcessesList->Refresh();
-      }
+      // TODO (177304549): This is still here from the old UI. Remove ui->HomeTab completely
+      // completely.
```
karupayun, line 709: nit: Remove this completely

src/OrbitQt/orbitmainwindow.h
```diff
@@ -95,7 +90,7 @@ class OrbitMainWindow : public QMainWindow {
 
   void RestoreDefaultTabLayout();
 
-  std::optional<orbit_qt::TargetConfiguration> ClearTargetConfiguration();
+  orbit_qt::TargetConfiguration ClearTargetConfiguration();
```
karupayun, line 93: If we are returning TargetConfiguration, should this function be [[nodiscard]]?

karupayun (approved):
LGTM!! Really happy to see TODO's being cleaned and everything else commented with a link to the bug.

src/OrbitQt/orbitmainwindow.cpp
antonrohr: Yes, but what I just realized, the actual problem is that this is poorly named. 

That the function does is only remove the things from the target that belong to the mainwindow. Not everything of the target will be "shutdown". For example the grpc connection will not be cancelled, since it can be reused again in the startup window.

There are multiple options
1. I could remove the `ShutdownTarget` function and then the `ClearTargetConfiguration` function would look like this:
```
orbit_qt::TargetConfiguration OrbitMainWindow::ClearTargetConfiguration() {
  if (std::holds_alternative<orbit_qt::StadiaTarget>(target_configuration_)) {
    std::get<orbit_qt::StadiaTarget>(target_configuration_)
        .GetProcessManager()
        ->SetProcessListUpdateListener(nullptr);
  } else if (std::holds_alternative<orbit_qt::LocalTarget>(target_configuration_)) {
    std::get<orbit_qt::LocalTarget>(target_configuration_)
        .GetProcessManager()
        ->SetProcessListUpdateListener(nullptr);
  }
  return std::move(target_configuration_);
}
```

2. Rename the `ShutdownTarget` function to one of those (this also needs to be defined for `FileTarget` which is somewhat awkward)
a. `ClearProcessListCallback`
b. `ClearProcessListCallbackOnTarget`
c. `RemoveMainWindowArtifactsFromTarget`
d. `ShutdownProcessListUpdate`

I think I prefer 2 a, but I would like to have your opinions

src/OrbitQt/orbitmainwindow.h
antonrohr: Good catch, thank you!

src/OrbitQt/orbitmainwindow.cpp
antonrohr: Thanks :)

src/OrbitQt/orbitmainwindow.h
antonrohr: Well actually. Forget my last comment. I deliberately chose this not to be to be `[[nodiscard]]`. The result of this function is only used when the target can be reused. In the case that an error with the connection occurs, the target is discarded and a new connection has to be established. 
You can see this in `main.cpp:132`
I hope this explanation makes sense

src/OrbitQt/orbitmainwindow.cpp
beckerhe: To me what is actually happening (clearing the process list callback) is an implementation detail. That's why I don't like 2a, 2b and 2d so much.

Next I would consider which part of the code is actually responsible for this to happen. Is the main window or is it the target class?

Since this code is specific to the main window, I would argue it's the main windows's responsibility. So the code should also live there.

So I would vote for 1.

src/OrbitQt/orbitmainwindow.h
beckerhe: You could consider still making it nodiscard and explicitly discarding it in main.cpp when not needed. By doing this you make clear to others that you thought this code path and that it's not an oversight.

antonrohr: I like that idea, I will do that

src/OrbitQt/orbitmainwindow.cpp
antonrohr: Okay. I change it to option 1. I will rebase at the same time and then merge afterwards

IrinaShkviro:
Creation of MetricsUploader was removed from `main.cpp` with this PR. Was it added somewhere else later?

antonrohr:
> Creation of MetricsUploader was removed from `main.cpp` with this PR. Was it added somewhere else later?

No it wasn't. Sorry for that mishap. Thanks for the other PR, which adds it again

PR #1695, opened 2021-01-19
Fix the partial disappeared capture toolbar on small screens

vickyliu-go4it:
The filter panel in the capture toolbar partially disappears on small screens. The toolbar extension button (i.e., the small grey arrow) on the right side does not do anything.

The reason is that, when a `QToolBar` is not a child of a `QMainWindow`, it loses the ability to populate the extension pop up with widgets added to the toolbar using `addWidget()`. To make the extension button works for the filter panel, we need use the widget action (i.e., `FilterPanelWidgetAction`) created by inheriting QWidgetAction and implementing `QWidgetAction::createWidget()` instead. 

In this PR:
* we move the filter panel from orbitmainwindow.ui to FilterPanelWidget.ui and create the `FilterPanelWidget` class;
* we create the `FilterPanelWidgetAction` class and add the customized widget action to capture toolbar.

Bug: http://b/169397517.

Test: start a new capture / load a capture &#10233; move the splitter / change the size of window &#10233; edit filters in the capture toolbar or in the extension button.

### DIFF
--- a/src/OrbitQt/FilterPanelWidget.cpp
+++ b/src/OrbitQt/FilterPanelWidget.cpp
@@ -1,0 +1,34 @@
+// Copyright (c) 2020 The Orbit Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#include "FilterPanelWidget.h"
+
+#include <QFontMetrics>
+#include <QFrame>
+#include <QWidget>
+
+#include "orbitmainwindow.h"
+#include "ui_FilterPanelWidget.h"
+
+FilterPanelWidget::FilterPanelWidget(QWidget* parent)
+    : QFrame(parent), ui(new Ui::FilterPanelWidget) {
+  ui->setupUi(this);
+
+  connect(ui->filterTracks, &QLineEdit::textChanged, this,
+          &FilterPanelWidget::FilterTracksTextChanged);
+  connect(ui->filterFunctions, &QLineEdit::textChanged, this,
+          &FilterPanelWidget::FilterFunctionsTextChanged);
+
+  QFontMetrics fm(ui->timerLabel->font());
+  int pixel_width = fm.horizontalAdvance("w");
+  ui->timerLabel->setMinimumWidth(5 * pixel_width);
+}
+
+void FilterPanelWidget::SetFilterFunctionsText(const QString& text) {
+  ui->filterFunctions->blockSignals(true);
+  ui->filterFunctions->setText(text);
+  ui->filterFunctions->blockSignals(false);
+}
+
+void FilterPanelWidget::SetTimerLabelText(const QString& text) { ui->timerLabel->setText(text); }
--- a/src/OrbitQt/FilterPanelWidget.h
+++ b/src/OrbitQt/FilterPanelWidget.h
@@ -1,0 +1,37 @@
+// Copyright (c) 2020 The Orbit Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#ifndef ORBIT_QT_FILTER_PANEL_WIDGET_H_
+#define ORBIT_QT_FILTER_PANEL_WIDGET_H_
+
+#include <QFrame>
+#include <QWidget>
+
+#include "ui_FilterPanelWidget.h"
+
+namespace Ui {
+class FilterPanelWidget;
+}
+
+/*
+ * A widget containing a track filter, a function filter, and a timer label. Will be added as a
+ * widget action to the capture toolbar. (see FilterPanelWidgetAction.h for more details)
+ */
+class FilterPanelWidget : public QFrame {
+  Q_OBJECT
+
+ public:
+  explicit FilterPanelWidget(QWidget* parent = nullptr);
+
+  void SetFilterFunctionsText(const QString& text);
+  void SetTimerLabelText(const QString& text);
+
+ signals:
+  void FilterTracksTextChanged(const QString& text);
+  void FilterFunctionsTextChanged(const QString& text);
+
+ private:
+  Ui::FilterPanelWidget* ui;
+};
+#endif  // ORBIT_QT_FILTER_PANEL_WIDGET_H_
--- a/src/OrbitQt/FilterPanelWidgetAction.h
+++ b/src/OrbitQt/FilterPanelWidgetAction.h
@@ -1,0 +1,38 @@
+// Copyright (c) 2020 The Orbit Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#ifndef ORBIT_QT_FILTER_PANEL_WIDGET_ACTION_H_
+#define ORBIT_QT_FILTER_PANEL_WIDGET_ACTION_H_
+
+#include <QFrame>
+#include <QWidget>
+#include <QWidgetAction>
+
+#include "FilterPanelWidget.h"
+
+/*
+ * Widget action used to insert the custom widget, i.e., FilterPanelWidget, into capture toolbar.
+ *
+ * As the capture toolbar is not a child of a QMainWindow, it loses the ability to populate the
+ * extension pop up with widgets added to the toolbar using addWidget(). To insert custom widgets,
+ * we need widget actions created by inheriting QWidgetAction and implementing
+ * QWidgetAction::createWidget() method.
+ */
+class FilterPanelWidgetAction : public QWidgetAction {
+  Q_OBJECT
+
+ public:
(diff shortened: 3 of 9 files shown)


vickyliu-go4it:
Before the change:
The filter panel in capture toolbar disappeared on small screens

After the change:
The filter panel is added to the extension pop up

src/OrbitQt/FilterPanelWidget.cpp
```diff
@@ -0,0 +1,41 @@
...
+
+#include "orbitmainwindow.h"
+#include "ui_FilterPanelWidget.h"
+
+FilterPanelWidget::FilterPanelWidget(QWidget* parent)
+    : QFrame(parent), ui(new Ui::FilterPanelWidget) {
+  ui->setupUi(this);
+
+  QFontMetrics fm(ui->timerLabel->font());
+  int pixel_width = fm.width("w");
```
reichlfl, line 19: Apparently `QFontMetrics::width` is deprecated, check Dimitry's PR here: https://github.com/google/orbit/pull/1692/files#diff-fc6728b247ca2e98948bc9b7413fabad7dd57246578557c0f663571bf133742b

src/OrbitQt/FilterPanelWidgetAction.h
```diff
@@ -0,0 +1,40 @@
...
+#ifndef ORBIT_QT_FILTER_PANEL_WIDGET_ACTION_H_
+#define ORBIT_QT_FILTER_PANEL_WIDGET_ACTION_H_
+
+#include <QFrame>
+#include <QWidget>
+#include <QWidgetAction>
+
+#include "FilterPanelWidget.h"
+
+class FilterPanelWidgetAction : public QWidgetAction {
```
reichlfl, line 14: Add a short comment for this class

src/OrbitQt/FilterPanelWidget.h
```diff
@@ -0,0 +1,45 @@
...
+#include <QFrame>
+#include <QWidget>
+
+#include "ui_FilterPanelWidget.h"
+
+namespace Ui {
+class FilterPanelWidget;
+}
+
+class FilterPanelWidget : public QFrame {
```
reichlfl, line 17: Add a short comment for this class

src/OrbitQt/FilterPanelWidget.h
```diff
@@ -0,0 +1,45 @@
...
+  void SetTimerLabelText(const QString& text);
+
+ private slots:
+  void on_filterFunctions_textChanged(const QString& text);
+  void on_filterTracks_textChanged(const QString& text);
+
+ private:
+  Ui::FilterPanelWidget* ui;
+  std::function<void(const QString& text)> filter_functions_text_changed_callback_;
+  std::function<void(const QString& text)> filter_tracks_text_changed_callback_;
```
reichlfl, line 43: It would probably be easier and more consistent with QT to not handle these with callbacks, but with Signals / Slots as well. You can introduce a public signal `FilterFunctionsTextChanged` in `FilterPanelWidget` and in `FilterPanelWidgetAction`. You should then be able to connect signal to signal and reduce the boilerplate quite a lot, i.e.:

In `FilterPanelWidget`:
```
QObject::connect(ui->filterFunctions, SIGNAL(TextChanged(QString)), this, SIGNAL(FilterFunctionsTextChanged(QString)));
```

And in `FilterPanelWidgetAction`:
```
QObject::connect(filter_panel_, SIGNAL(FilterFunctionsTextChanged(QString)), this, SIGNAL(FilterFunctionsTextChanged(QString)));
```

And then connect `OrbitMainWindow::OnFilterFunctionsTextChanged` to `FilterPanelWidgetAction::FilterFunctionsTextChanged`.

I've written this from the top of my head, so not 100% sure if the syntax is correct, but this should work.

reichlfl (commented):
Thanks a lot for taking care of this! I think this is a nice solution that will allow us to extend the toolbar without thinking too much about space restrictions.

PTAL at my comments regarding signals / slots, I feel this would make the code a lot easier, WDYT?

src/OrbitQt/FilterPanelWidget.cpp
vickyliu-go4it: Changed to use `QFontMetrics::width`. Thanks for that!

src/OrbitQt/FilterPanelWidget.h
vickyliu-go4it: Done. Thanks so much! Using signals & slots simplifies the whole process a lot!

src/OrbitQt/FilterPanelWidgetAction.h
vickyliu-go4it: Done. Thanks for that!

src/OrbitQt/FilterPanelWidget.h
vickyliu-go4it: Done. Thanks for that!

src/OrbitQt/orbitmainwindow.ui
```diff
@@ -58,7 +58,7 @@ QPushButton:disabled {
         <string>MainTabWidget</string>
        </property>
        <property name="currentIndex">
-        <number>0</number>
+        <number>1</number>
```
reichlfl, line 61: Is this intended and does it make a difference after startup?

src/OrbitQt/FilterPanelWidget.cpp
```diff
@@ -0,0 +1,34 @@
...
+#include "ui_FilterPanelWidget.h"
+
+FilterPanelWidget::FilterPanelWidget(QWidget* parent)
+    : QFrame(parent), ui(new Ui::FilterPanelWidget) {
+  ui->setupUi(this);
+
+  connect(ui->filterTracks, &QLineEdit::textChanged, this,
+          &FilterPanelWidget::FilterTracksTextChanged);
+  connect(ui->filterFunctions, &QLineEdit::textChanged, this,
+          &FilterPanelWidget::FilterFunctionsTextChanged);
```
reichlfl, line 21: Nitpick, sorry, didn't spot this before: You could to this in the Designer right away, it was done like this before in the main window. You can define custom slots for your Widget in Designer, and then use the signal / slot editor to connect them.

But we're doing this in large parts in code already, no urgent need to change it, so ignore this if you want. Just wanted to let you know this is possible. In our case, I don't think we have a super strong need to do this because we're building parts of the UI dynamically and need to make the connections in Code. For a window completely defined in Designer, I find this to be easier because everything is done in one place.

reichlfl (approved):
Great change, thanks a lot!

src/OrbitQt/orbitmainwindow.ui
vickyliu-go4it: Thanks for pointing out this!!!! This is not intended. I think when I editing the ui file in qtdesigner, I clicked the capture window but forgot to change back to the home window.

src/OrbitQt/FilterPanelWidget.cpp
vickyliu-go4it: Thanks for the suggestion! I move these connections to the ui file. I totally agree that for a window defined in qtdesigner, it would be better if the connections are added in the same place.

karupayun (commented):
Hi Vicky! Thanks for the change :)

I don't have any comments about the code, LGTM for that! 
The only issue I found is that the button is really difficult to see. It was difficult before, and maybe now it is worse.
You can test in some online tool (like https://juicystudio.com/services/luminositycontrastratio.php) to see how visible is something (there are requirements about contrast ratio for smaller/bigger items). I suggest to aim to 4.5 here, because button is small.

karupayun:
As discussed online, this is an interesting link to Google accessibility training where Color Contrast is discussed:(https://g3doc.corp.google.com/ads/awapps2/shared/a11y/g3doc/codelab/basic/color_contrast.md?cl=head)

And also 2 screenshots where we can see that arrows are not very accessible (difficult to see):
arrows1
arrows2

vickyliu-go4it:
> As discussed online, this is an interesting link to Google accessibility training where Color Contrast is discussed:(https://g3doc.corp.google.com/ads/awapps2/shared/a11y/g3doc/codelab/basic/color_contrast.md?cl=head)
> 
> And also 2 screenshots where we can see that arrows are not very accessible (difficult to see):
> arrows1
> arrows2
> (this last one is incredible difficult to see the button.
> 
> Ps: It isn't a regression.

Cool suggestion! Thanks so much for that. To make the extension button (i.e., the small double arrows) more accessible, the best way would be replace the original icon by `toolBar->findChild<QToolButton*>("qt_toolbar_ext_button")->setIcon(...)`. We can also change the background color, but this might not be a good solution. 

I'm not very familiar about the process of adding new icons to Orbit. I will consider it as a future PR. Thanks again ;-)

PR #1697, opened 2021-01-19
Clean datamanager

antonrohr:
This PR removes all code related to processes from DataManager. These parts are still leftovers from the old ui. 

Since this needed to touch DataManagerUpdateModuleInfosFuzzer and removed DataManager from it, I also renamed it to ModuleListFuzzer. 
@beckerhe I don't know if there are implications with this rename outside of this repository (maybe with oss fuzz configs or similar). If, then we should probably talk about it and I can change those as well.

### DIFF
--- a/src/OrbitGl/ModuleListFuzzer.cpp
+++ b/src/OrbitGl/ModuleListFuzzer.cpp
@@ -9,7 +9,6 @@
 #include <vector>
 
 #include "ModulesDataView.h"
-#include "OrbitBase/Logging.h"
 #include "OrbitClientData/ModuleManager.h"
 #include "OrbitClientData/ProcessData.h"
 #include "absl/flags/flag.h"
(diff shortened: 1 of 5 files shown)


beckerhe (approved):
The name of a fuzzer needs to end with `Fuzzer`. That's the only requirement. Renaming a fuzzer might trigger some emails because existing  reports will be considered new ones. Apart from that it's fine.

LGTM!

src/OrbitGl/ModuleListFuzzer.cpp
```diff
@@ -47,14 +45,10 @@ DEFINE_PROTO_FUZZER(const GetModuleListResponse& module_list) {
...
-  data_manager.UpdateProcessInfos(std::vector{process_info});
-
-  ProcessData* process = data_manager.GetMutableProcessByPid(pid);
-  CHECK(process != nullptr);
-  process->UpdateModuleInfos(modules);
+  process.UpdateModuleInfos(modules);
 
   ModulesDataView modules_data_view{nullptr};
-  modules_data_view.UpdateModules(process);
+  modules_data_view.UpdateModules(&process);
```
dimitry-, line 53: this looks suspicious who owns process after that? is it expected to be a temporary? if so why is it not const ref or passed by value?

dimitry-: oh, this is fuzzer and modules_data_view is also in the same scope, nevermind

PR #1698, opened 2021-01-19
Rename master to main everywhere

antonrohr:
I did this change by searching "master" in the repository. For the kokoro scripts, I don't know how to test if everything works. I also don't know if there are any other locations (outside of this repository) where things have to be changed accordingly. 

This change was prompted because I looked at the iwyu presubmit check in a different PR and saw this:
```
...
The following is a diff that fixes all problems in files this PR touched:
Execute `patch -p1 -i <filename.diff>` in the repo root to apply it to the code base.
fatal: Not a valid object name origin/master
...
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,7 +21,7 @@ Orbit combines sampling and dynamic instrumentation to optimize the profiling wo
 
 **Note**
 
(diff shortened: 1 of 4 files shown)


beckerhe (approved):
Great. Thanks! LGTM!

PR #1701, opened 2021-01-20
Fix Vulkan include

pierricgimmig:


pierricgimmig:
I'm not convinced this is the proper fix, especially after seeing the code block in FindVulkanValidationLayers.cmake regarding vk_layer_dispatch_table.h, but it works on my side on GLinux, let's see if it passes the build checks pass.

florian-kuebler (changes requested):
Yep, that Cmake code was exactly to avoid doing the import like this. I'll have a look on a more proper fix. Anyways, thanks for looking into this.

pierricgimmig:
Deleting the build_* directories (clearing the cmake cache essentially) fixed the issue.

PR #1702, opened 2021-01-20
liborbit.so [WIP]

pierricgimmig:


src/OrbitGl/ScopeTree.cpp
```diff
@@ -0,0 +1,109 @@
...
+
+void ScopeNode::Print(const ScopeNode* node, uint64_t start_time, uint32_t depth) {
+  LOG("%s%lu ScopeNode(%u) %p [%lu, %lu]", std::string(depth, ' '), start_time, node->id_, node,
+      node->timer_info_->start(), node->timer_info_->end());
+  for (auto& pair : node->children_) {
+    Print(pair.second, pair.first, depth + 1);
+  }
+}
+
+void ScopeNode::FindDepth(const ScopeNode* node, size_t* max_depth, size_t current_depth) {
```
karupayun, line 38: O(n), consider saving depth in the nodes if you want this operation to be faster

src/OrbitGl/ScopeTree.cpp
```diff
@@ -0,0 +1,109 @@
...
+
+ScopeNode* ScopeNode::GetFirstChildBeforeOrAtTime(uint64_t time) const {
+  // Get first child before or exactly at "time".
+  if (children_.empty()) return nullptr;
+  auto next_node_it = children_.upper_bound(time);
+  if (next_node_it == children_.begin()) return nullptr;
+  return (--next_node_it)->second;
+}
+
+ScopeNode* ScopeNode::FindDeepestParentForNode(const ScopeNode* node) {
```
karupayun, line 59: DeepestParent and we look for the descendants?

src/OrbitGl/ScopeTree.cpp
```diff
@@ -0,0 +1,109 @@
...
+  return deepest_node;
+}
+
+std::vector<ScopeNode*> ScopeNode::GetChildrenInRange(uint64_t start, uint64_t end) const {
+  // Get children that are enclosed between start and end inclusively.
+  if (children_.empty()) return {};
+  auto node_it = children_.upper_bound(end);
+  if (node_it == children_.begin()) return {};
+  std::vector<ScopeNode*> nodes;
+  while (node_it != children_.begin()) {
```
karupayun, line 77: Going backward is because we want that list in reverse order?

src/OrbitGl/ScopeTree.cpp
```diff
@@ -0,0 +1,109 @@
...
+  }
+}
+
+size_t ScopeNode::Depth() const {
+  size_t max_depth = 0;
+  FindDepth(this, &max_depth);
+  return max_depth;
+}
+
+ScopeNode* ScopeNode::GetFirstChildBeforeOrAtTime(uint64_t time) const {
```
karupayun, line 51: GetLastChildBeforeOrAtTime

src/OrbitGl/ScopeTreeTest.cpp
```diff
@@ -0,0 +1,62 @@
...
+  auto timer_info = std::make_unique<TimerInfo>();
+  timer_info->set_start(start);
+  timer_info->set_end(end);
+  timer_buffer.push_back(std::move(timer_info));
+  return *timer_buffer.back();
+}
+
+TEST(ScopeTree, TreeCreation) {
+  ScopeTree tree;
+  EXPECT_EQ(tree.Size(), 1);
```
karupayun, line 28: Although we are having an initial dummy node, I would expect the size and the Depth to not include it.

src/OrbitGl/ScopeTree.cpp
pierricgimmig: Reimplemented so that the nodes are now in ascending order, thanks for pointing it out. It didn't change the behavior as the nodes were added in a map after, but it's better to have them in ascending order.

src/OrbitGl/ScopeTree.cpp
pierricgimmig: Good point, done.

src/OrbitGl/ScopeTreeTest.cpp
pierricgimmig: Good point, I changed Depth() to Height() which is the number of edges of the longest path, which fits with what you expect.

PR #1709, opened 2021-01-21
Disable qopenglfunctions.h warning to reduce compilation output spam

pierricgimmig:
Disable "qopenglfunctions.h is not compatible with GLEW, GLEW defines will be undefined" warning
to reduce spam in compilation output. This is a known quirk that doesn't cause any ill effect.


reichlfl (approved):
Nice, thanks for taking care of this

beckerhe (approved):
LGTM! Thanks Pierric!

PR #1710, opened 2021-01-21
Improve BatcherId/PickingType encoding logic  

pierricgimmig:
Integrate Henning's suggestion to get rid of annoying warnings related to
BatcherId/PickingType encoding logic. Add additional static asserts to make 
sure we allocate enough bits for the number of entries we need to encode.

### DIFF
--- a/src/OrbitGl/Batcher.cpp
+++ b/src/OrbitGl/Batcher.cpp
@@ -266,6 +266,9 @@ const PickingUserData* Batcher::GetUserData(PickingId id) const {
       return user_data_[id.element_id].get();
     case PickingType::kPickable:
       return nullptr;
+    case PickingType::kCount:
+      UNREACHABLE();
+      return nullptr;
   }
 
   UNREACHABLE();
--- a/src/OrbitGl/Batcher.cpp
+++ b/src/OrbitGl/Batcher.cpp
@@ -268,7 +268,6 @@ const PickingUserData* Batcher::GetUserData(PickingId id) const {
       return nullptr;
     case PickingType::kCount:
       UNREACHABLE();
-      return nullptr;
   }
 
   UNREACHABLE();
--- a/src/OrbitGl/PickingManager.h
+++ b/src/OrbitGl/PickingManager.h
@@ -68,6 +68,8 @@ struct PickingId {
   [[nodiscard]] inline static PickingId Create(PickingType type, uint32_t element_id,
                                                BatcherId batcher_id = BatcherId::kTimeGraph) {
     CHECK(element_id >> kElementIDBitSize == 0);
+    CHECK(type >= PickingType{} && type < PickingType::kCount);
+    CHECK(batcher_id >= BatcherId{} && batcher_id < BatcherId::kCount);
     PickingId result{};
     result.type = type;
     result.element_id = element_id;
@@ -77,10 +79,8 @@ struct PickingId {
 
   [[nodiscard]] inline static PickingId FromPixelValue(uint32_t value) {
     const Layout layout = absl::bit_cast<Layout, uint32_t>(value);
-    CHECK(layout.type >= static_cast<uint32_t>(PickingType::kInvalid) &&
-          layout.type <= static_cast<uint32_t>(PickingType::kPickable));
-    CHECK(layout.batcher_id >= static_cast<uint32_t>(BatcherId::kTimeGraph) &&
-          layout.batcher_id <= static_cast<uint32_t>(BatcherId::kUi));
+    CHECK(layout.type < static_cast<uint32_t>(PickingType::kCount));
+    CHECK(layout.batcher_id < static_cast<uint32_t>(BatcherId::kCount));
 
     PickingId id{};
     id.element_id = layout.element_id;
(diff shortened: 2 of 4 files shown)


dimitry-:
I do not see any warnings when compiling with clang. Is this windows-specific warning? What does it say?

reichlfl:
> I do not see any warnings when compiling with clang. Is this windows-specific warning? What does it say?

Example:
OrbitGl/PickingManager.h:87:15: warning: ‘PickingId::type’ is too small to hold all values of ‘enum class PickingType’

These show up (at least) in the gcc9 compilation done by the CI, see this log for example: https://fusion2.corp.google.com/invocations/1dc4dd17-d5f5-42f2-849a-53207194a988/targets/orbitprofiler%2Fbuilds%2Flinux%2Fgcc9_relwithdebinfo%2Fpresubmit/log

reichlfl:
I'm looping in Henning as well since we briefly talked about this before... this was deliberately changed since `enum class` is preferable, but at that point I was unaware of the log spam. I agree that we should get rid of it since it's cluttering the log, but I'm not sure about reverting to the old unscoped enums TBH.

dimitry-:
Maybe suppress the warning in gcc?

reichlfl:
I've just realized the warning isn't actually gone, at least not on our gcc9 build on the CI. I was a bit suprised that this should get rid of it, because technically, the warning is equally valid as before because the underlying data types of the enums are still unit32_t.

beckerhe:
Thank you a lot Pierric that you're tackling the warnings! Very much appreciated!

May I suggest something like this? https://github.com/beckerhe/orbitprofiler/pull/new/picking_id

This avoids having enums in a bitfields, so it avoids the warnings.

antonrohr:
@beckerhe I think your link was meant to point here:
https://github.com/beckerhe/orbitprofiler/commit/849fbfa5df81c2ea2258e94ac4a809988277b012

pierricgimmig:
@beckerhe , @reichlfl , thanks for the help. I integrated @beckerhe's suggestion and also added static asserts so that we fail at compile time if the number of bits to encode BatcherId/PickingId is not sufficient. PTAL. Thanks!

src/OrbitGl/Batcher.cpp
```diff
@@ -266,6 +266,9 @@ const PickingUserData* Batcher::GetUserData(PickingId id) const {
       return user_data_[id.element_id].get();
     case PickingType::kPickable:
       return nullptr;
+    case PickingType::kCount:
+      UNREACHABLE();
+      return nullptr;
```
dimitry-, line 271: either UNREACHABLE() or return nullptr; I do not think you can do both

src/OrbitGl/PickingManager.h
```diff
@@ -51,41 +45,70 @@ enum class PickingType : uint32_t {
...
+  static_assert(sizeof(Layout) == sizeof(uint32_t),
+                "Layout needs to have the size of a single uint32_t.");
+  static_assert(1 << kPickingTypeBitSize >= static_cast<int>(PickingType::kCount),
+                "PickingType cannot be encoded in the specified number of bits.");
+  static_assert(1 << kBatcherIDBitSize >= static_cast<int>(BatcherId::kCount),
+                "BatcherId cannot be encoded in the specified number of bits.");
 
   [[nodiscard]] inline static PickingId Create(PickingType type, uint32_t element_id,
                                                BatcherId batcher_id = BatcherId::kTimeGraph) {
     CHECK(element_id >> kElementIDBitSize == 0);
```
dimitry-, line 70: add CHECK for batcher_id?

pierricgimmig: Done and also added check for picking type.

src/OrbitGl/Batcher.cpp
pierricgimmig: I thought I'd get an error saying that not all paths return a value, I don't. I removed the "return nullptr".

pierricgimmig:
Thanks for the comments @dimitry- , I modified the code accordingly.

src/OrbitGl/PickingManager.h
```diff
@@ -51,41 +45,70 @@ enum class PickingType : uint32_t {
...
-    PickingId result;
+    CHECK(type >= PickingType{} && type < PickingType::kCount);
+    CHECK(batcher_id >= BatcherId{} && batcher_id < BatcherId::kCount);
```
dimitry-, line 72: these are strongly typed I was thinking about doing what CHECK above does

CHEC(type >> kPickingTypeBitSize == 0);

So that even if somehow there is a value greater than kCount but it does not fit into bits we get an explicit abort here.

pierricgimmig: Isn't what you are suggesting less strict than what we have here? The suggested check would allow for a PickingType greater than PickingType::kCount if it fits in the number of allocated bits, but that should be invalid.

dimitry-: well, if we take static_assert into account - yes. maybe in this case move the static assert here for context?

dimitry- (approved):
otherwise lgtm

beckerhe (approved):
Thanks a lot Pierric for looking into this. LGTM!

PR #1712, opened 2021-01-21
Fix the flickering capture toolbar extension popup

vickyliu-go4it:
When the user moves the splitter, shrinks down the main window, or works on a very low resolution screen, the capture toolbar extension toolbar keeps flickering. 

The reason is that every time when the process list updates (i.e., `OnProcessListUpdated(...)` triggered), we reset the icon of capture toggle button (i.e., in `UpdateCaptureStateDependentWidgets()`) and so cause the toolbar to re-insert every actions.

To fix the above problem, and also consider that these is no need to update every capture state dependent widget when the process connection state changes, we add the `UpdateProcessConnectionStateDependentWidgets()` method which only updates the widget affected by the process connection states when the process list is updated.

Bug: http://b/178063378
Blocking bug: http://b/169397517 (in PR https://github.com/google/orbit/pull/1695)
Test: Move the splitter to the left, click the capture toolbar extension button on the right sider, and then check the extension popup does not flicker any more.

### DIFF
--- a/src/OrbitQt/orbitmainwindow.cpp
+++ b/src/OrbitQt/orbitmainwindow.cpp
@@ -590,6 +590,18 @@ void OrbitMainWindow::UpdateCaptureStateDependentWidgets() {
   }
 }
 
+void OrbitMainWindow::UpdateProcessConnectionStateDependentWidgets() {
+  CaptureClient::State capture_state = app_->GetCaptureState();
+  const bool is_capturing = capture_state != CaptureClient::State::kStopped;
+  const bool is_connected = app_->IsConnectedToInstance();
+  const bool is_target_process_running = target_process_state_ == TargetProcessState::kRunning;
+
+  ui->actionToggle_Capture->setEnabled(
+      capture_state == CaptureClient::State::kStarted ||
+      (capture_state == CaptureClient::State::kStopped && is_target_process_running));
+  ui->actionOpen_Preset->setEnabled(!is_capturing && is_connected);
+}
+
 void OrbitMainWindow::UpdateActiveTabsAfterSelection(bool selection_has_samples) {
   const QTabWidget* capture_parent = FindParentTabWidget(ui->CaptureTab);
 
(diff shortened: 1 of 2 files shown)


vickyliu-go4it:
Example of the capture toolbar extension popup:
Capture toolbar extension popup

src/OrbitQt/orbitmainwindow.cpp
```diff
@@ -1203,7 +1210,7 @@ void OrbitMainWindow::OnProcessListUpdated(
     target_label_->setToolTip({});
     target_process_state_ = TargetProcessState::kRunning;
   }
-  UpdateCaptureStateDependentWidgets();
+  UpdateProcessConnectionStateDependentWidgets();
```
reichlfl, line 1213: If you remove `UpdateCaptureStateDependentWidgets`, you need to duplicate the logic around `assume_target_process_is_running` into `UpdateProcessConnectionStateDependentWidgets` - the Start-Button should be disabled when Orbit is not connected to a process, or when the connection got lost.

reichlfl (changes requested):
Thanks for looking into this. Unfortunately this will break some behavior, see my comment. But I think in general splitting this into the two methods `CaptureStateDependent` and `ProcessConnectionStateDependent` makes a lot of sense, I would suggest to extend the new method to also disable the Start / Stop Capture button accordingly.

src/OrbitQt/orbitmainwindow.cpp
reichlfl: On main, the variable is now called `is_target_process_running`, btw, I didn't have Anton's latest changes.

vickyliu-go4it: Added the logic around `is_target_process_running` into `UpdateProcessConnectionStateDependentWidgets`. Thanks for the comments. Sorry that I didn't realize this. I will be more careful the next time : )

reichlfl (approved):
LGTM!

karupayun (approved):
LGTM!! Thanks for discover and looking into this :)

PR #1714, opened 2021-01-21
Move pid/tid handling to OrbitBase.

danielfenner:
I need these outside of LinuxTracing.
    
Bug: N/A
Test: Unit tests.


src/OrbitBase/include/OrbitBase/ProcessId.h
```diff
@@ -0,0 +1,24 @@
+// Copyright (c) 2020 The Orbit Authors. All rights reserved.
```
dpallotti, line 1: I'm not a fan of the name of this file, it sounds like a type rather than describing what these functions are for.

From the include guards I see you had gone for "LinuxProc", maybe that's also not ideal but I liked it more.

What about "ListProcesses.h"? Or "ListProcessIds.h"?

src/OrbitBase/include/OrbitBase/ProcessId.h
```diff
@@ -0,0 +1,24 @@
...
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#ifndef ORBIT_BASE_LINUX_PROC_H_
+#define ORBIT_BASE_LINUX_PROC_H_
```
dpallotti, line 6: These include guards will have to be fixed (once we settle on the name of the file).

src/OrbitBase/include/OrbitBase/ProcessId.h
danielfenner: Yes, the name is not great. I wanted to change away from "LinuxProc" because I don't want "Linux" in the name (the name should describe what is done not the platform, also this would result in LinuxProcLinux.cpp which is weird).
"ListProcessIds.h" sounds ok-ish to me. 

I looked around in OrbitBase; there is already a file "ThreadUtils.h". I now think moving the stuff there is the best solution. Agree?

src/OrbitBase/ProcessIdLinux.cpp
```diff
@@ -0,0 +1,80 @@
...
+#include "OrbitBase/Logging.h"
+#include "OrbitBase/ProcessId.h"
+#include "OrbitBase/Result.h"
+#include "absl/strings/numbers.h"
+
+namespace orbit_base {
+
+namespace fs = std::filesystem;
+
+namespace {
```
dimitry-, line 18: not: I personally prefer static to anonymous namespaces but this is your call. (or we could have both also)

src/OrbitBase/ProcessIdLinux.cpp
```diff
@@ -0,0 +1,80 @@
+// Copyright (c) 2020 The Orbit Authors. All rights reserved.
```
dimitry-, line 1: nit: 2021 here and below.

src/OrbitBase/ProcessIdLinux.cpp
```diff
@@ -0,0 +1,80 @@
...
+}
+
+}  // namespace
+
+std::vector<pid_t> GetAllPids() {
+  fs::directory_iterator proc{"/proc"};
+  std::vector<pid_t> pids;
+
+  for (const auto& entry : proc) {
+    if (auto pid = ProcEntryToPid(entry); pid.has_value()) {
```
dimitry-, line 44: Move fist part out of if?

src/OrbitBase/ProcessIdLinux.cpp
```diff
@@ -0,0 +1,80 @@
...
+  int potential_pid;
+  if (!absl::SimpleAtoi(entry.path().filename().string(), &potential_pid)) {
+    return std::nullopt;
+  }
+
+  if (potential_pid <= 0) {
+    return std::nullopt;
+  }
+
+  return static_cast<pid_t>(potential_pid);
```
dimitry-, line 34: is this static cast needed here?

src/OrbitBase/include/OrbitBase/ProcessId.h
```diff
@@ -0,0 +1,24 @@
...
+
+#include <sys/types.h>
+
+#include <vector>
+
+namespace orbit_base {
+
+#if defined(__linux)
+std::vector<pid_t> GetAllPids();
```
dimitry-, line 15: Add short comment of what these functions do.

I am also thinking if they should be renamed to Find* of Fetch*, probably fine as it is.

src/OrbitBase/include/OrbitBase/ProcessId.h
dpallotti: I kind of like having them separate, with these free-standing function in `OrbitBase` it's nicer to have small clearly-separate files (matches better the include what you use philosophy).

As for the name, if you choose to rename the functions to something other than "Get" (but for me they're also fine as they are), then the name of the file could match that (e.g., with "Fetch" this could be called "FetchProcessIds.h").

src/OrbitBase/include/OrbitBase/ProcessId.h
dpallotti: "Find" is inaccurate, it's not a search: "Fetch", "Retrieve" or "List" are fine by me (but the current "Get" is also fine by me).

src/OrbitBase/ProcessIdLinux.cpp
dpallotti: This is actually something that in other reviews we agreed to do: when the variable is only used in the if block, it's better to scope it and have it in the condition (imagine there was other code after the if).

src/OrbitBase/ProcessIdLinux.cpp
dpallotti: Not in practice, but `potential_pid` is `int`, not `pid_t`.

dpallotti (commented):
Also replying to some of your comments, Dimitry, as I wrote some of this code.

src/OrbitBase/ProcessIdLinux.cpp
dpallotti: I guess it can be made `pid_t`? I'm not sure why it was like that.

dimitry-: I think it is safer to remove this. we would want to know if pid_t becomes something that needs an explicit cast from int

src/OrbitBase/ProcessIdLinux.cpp
danielfenner: Is this the agreed way of doing this? I often see something like "// Copyright (c) 2020-2021 ... " in other source code.

dimitry-: Just a year when file was created. We do not want constantly updating them, so I would prefer we avoid ranges. (I think the only range we want to update is the one in the LICENSE file)

src/OrbitBase/ProcessIdLinux.cpp
danielfenner: I have a muscle memory to avoid static in these cases (I think the style guide aske for unnanmed namespace in the past). But I really don't care. Changed.

src/OrbitBase/ProcessIdLinux.cpp
danielfenner: Agreed, looks weird.

src/OrbitBase/ProcessIdLinux.cpp
dpallotti: This should be discussed with the team, the examples in http://go/copyright agree with Daniel.

src/OrbitBase/ProcessIdLinux.cpp
danielfenner: done.

src/OrbitBase/ProcessIdLinux.cpp
dpallotti: Well, actually not for creation, never mind.

src/OrbitBase/include/OrbitBase/ProcessId.h
danielfenner: GetProcessIds.{h,cpp} it is.

src/OrbitBase/include/OrbitBase/ProcessId.h
danielfenner: I'm fine with 'Get'. Added comments.

src/OrbitBase/include/OrbitBase/ProcessId.h
danielfenner: done

src/OrbitBase/GetProcessIdsLinux.cpp
```diff
@@ -0,0 +1,77 @@
...
+  return potential_pid;
+}
+
+std::vector<pid_t> GetAllPids() {
+  fs::directory_iterator proc{"/proc"};
+  std::vector<pid_t> pids;
+
+  for (const auto& entry : proc) {
+    auto pid = ProcEntryToPid(entry);
+    if (pid.has_value()) {
```
dpallotti, line 41: Yeah actually here I meant that in other reviews we had gone for how it was before. Something for another religious discussion in a standup.

src/OrbitBase/include/OrbitBase/GetProcessIds.h
```diff
@@ -0,0 +1,27 @@
+// Copyright (c) 2021 The Orbit Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#ifndef ORBIT_BASE_GET_PROCESS_IDS_H
```
dpallotti, line 5: Just a heads-up that the styleguide want the trailing underscore in the include guards. We do it quite consistently.

src/OrbitBase/include/OrbitBase/GetProcessIds.h
```diff
@@ -0,0 +1,27 @@
...
+namespace orbit_base {
+
+#if defined(__linux)
+// Get the process ids of all currently running processes.
+std::vector<pid_t> GetAllPids();
+
+// Get the thread ids of all the threads belonging to process 'pid'.
+std::vector<pid_t> GetTidsOfProcess(pid_t pid);
+
+// Get all thread ids of all currently running processes.
```
dpallotti, line 21:
```suggestion
// Get all thread ids of all currently running threads.
```

dpallotti (approved):

src/OrbitBase/include/OrbitBase/GetProcessIds.h
```diff
@@ -0,0 +1,27 @@
...
+
+#include <sys/types.h>
+
+#include <vector>
+
+namespace orbit_base {
+
+#if defined(__linux)
+// Get the process ids of all currently running processes.
+std::vector<pid_t> GetAllPids();
```
dimitry-, line 16: Get(All)(Running)ProcessIds

I think we follow ThreadId ProcessId naming conventions elsewhere in OrbitBase

src/OrbitBase/include/OrbitBase/GetProcessIds.h
```diff
@@ -0,0 +1,27 @@
...
+
+// Get the thread ids of all the threads belonging to process 'pid'.
+std::vector<pid_t> GetTidsOfProcess(pid_t pid);
```
dimitry-, line 19: GetProcessThreadIds

src/OrbitBase/include/OrbitBase/GetProcessIds.h
```diff
@@ -0,0 +1,27 @@
...
+
+// Get all thread ids of all currently running processes.
+std::vector<pid_t> GetAllTids();
```
dimitry-, line 22: Get(All)(Running)ThreadIds

dimitry- (approved):
lgtm, renaming is up to you, it could also be done separately if there is too many references to these.

src/OrbitBase/GetProcessIdsLinux.cpp
danielfenner: Oh sorry, I skipped over that when addressing comments. 

I think this looks way more readable as it is now. If scope is an issue one can always add additional {}. Yeah, let's (not) discuss at some point :)

src/OrbitBase/include/OrbitBase/GetProcessIds.h
danielfenner: oh, right, done.

src/OrbitBase/include/OrbitBase/GetProcessIds.h
danielfenner: Went with the even more explicit

"Get all thread ids of all the threads in all currently running processes."

PR #1715, opened 2021-01-21
Use std::optional rather than address of const ref or address of map …

florian-kuebler:
…element.

This fixes the flaky CaptureEventProcessor test. `FindMatching*` returned a
pointer to a map element. However, this map is not garantueed to be pointer
stable. Thus, we now return std::optional instead.

Also we used the address of operator on a const-ref., which is also
undefined. Again, we are now using std::optional here.

Test: Run CaptureEventProcessorTest a couple of times.

### DIFF
--- a/src/OrbitCaptureClient/GpuQueueSubmissionProcessor.cpp
+++ b/src/OrbitCaptureClient/GpuQueueSubmissionProcessor.cpp
@@ -12,6 +12,7 @@ using orbit_client_protos::TimerInfo;
 using orbit_grpc_protos::GpuCommandBuffer;
 using orbit_grpc_protos::GpuJob;
 using orbit_grpc_protos::GpuQueueSubmission;
+using orbit_grpc_protos::GpuDebugMarker;
 
 std::vector<TimerInfo> GpuQueueSubmissionProcessor::ProcessGpuQueueSubmission(
     const orbit_grpc_protos::GpuQueueSubmission& gpu_queue_submission,
@@ -23,14 +24,14 @@ std::vector<TimerInfo> GpuQueueSubmissionProcessor::ProcessGpuQueueSubmission(
       gpu_queue_submission.meta_info().pre_submission_cpu_timestamp();
   uint64_t post_submission_cpu_timestamp =
       gpu_queue_submission.meta_info().post_submission_cpu_timestamp();
-  const GpuJob* matching_gpu_job =
+  std::optional<GpuJob> matching_gpu_job_option =
       FindMatchingGpuJob(thread_id, pre_submission_cpu_timestamp, post_submission_cpu_timestamp);
 
   // If we haven't found the matching "GpuJob" or the submission contains "begin" markers (which
   // might have the "end" markers in a later submission), we save the "GpuSubmission" for later.
   // Note that as soon as all "begin" markers have been processed, the "GpuSubmission" will be
   // deleted again.
-  if (matching_gpu_job == nullptr || gpu_queue_submission.num_begin_markers() > 0) {
+  if (!matching_gpu_job_option.has_value() || gpu_queue_submission.num_begin_markers() > 0) {
     tid_to_post_submission_time_to_gpu_submission_[thread_id][post_submission_cpu_timestamp] =
         gpu_queue_submission;
   }
@@ -38,16 +39,16 @@ std::vector<TimerInfo> GpuQueueSubmissionProcessor::ProcessGpuQueueSubmission(
     tid_to_post_submission_time_to_num_begin_markers_[thread_id][post_submission_cpu_timestamp] =
         gpu_queue_submission.num_begin_markers();
   }
-  if (matching_gpu_job == nullptr) {
+  if (!matching_gpu_job_option.has_value()) {
     return {};
   }
 
   std::vector<TimerInfo> result = ProcessGpuQueueSubmissionWithMatchingGpuJob(
-      gpu_queue_submission, *matching_gpu_job, string_intern_pool,
+      gpu_queue_submission, matching_gpu_job_option.value(), string_intern_pool,
       get_string_hash_and_send_to_listener_if_necessary);
 
   if (!HasUnprocessedBeginMarkers(thread_id, post_submission_cpu_timestamp)) {
-    DeleteSavedGpuJob(thread_id, matching_gpu_job->amdgpu_cs_ioctl_time_ns());
+    DeleteSavedGpuJob(thread_id, matching_gpu_job_option->amdgpu_cs_ioctl_time_ns());
   }
   return result;
 }
@@ -58,17 +59,17 @@ std::vector<orbit_client_protos::TimerInfo> GpuQueueSubmissionProcessor::Process
         get_string_hash_and_send_to_listener_if_necessary) {
   int32_t thread_id = gpu_job.tid();
   uint64_t amdgpu_cs_ioctl_time_ns = gpu_job.amdgpu_cs_ioctl_time_ns();
-  const GpuQueueSubmission* matching_gpu_submission =
+  std::optional<GpuQueueSubmission> matching_gpu_submission =
       FindMatchingGpuQueueSubmission(thread_id, amdgpu_cs_ioctl_time_ns);
 
   // If we haven't found the matching "GpuSubmission" or the submission contains "begin" markers
   // (which might have the "end" markers in a later submission), we save the "GpuJob" for later.
   // Note that as soon as all "begin" markers have been processed, the "GpuJob" will be deleted
   // again.
-  if (matching_gpu_submission == nullptr || matching_gpu_submission->num_begin_markers() > 0) {
+  if (!matching_gpu_submission.has_value() || matching_gpu_submission->num_begin_markers() > 0) {
     tid_to_submission_time_to_gpu_job_[thread_id][amdgpu_cs_ioctl_time_ns] = gpu_job;
   }
-  if (matching_gpu_submission == nullptr) {
+  if (!matching_gpu_submission.has_value()) {
     return {};
   }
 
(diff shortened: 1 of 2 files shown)


src/OrbitCaptureClient/GpuQueueSubmissionProcessor.cpp
```diff
@@ -23,31 +24,31 @@ std::vector<TimerInfo> GpuQueueSubmissionProcessor::ProcessGpuQueueSubmission(
       gpu_queue_submission.meta_info().pre_submission_cpu_timestamp();
   uint64_t post_submission_cpu_timestamp =
       gpu_queue_submission.meta_info().post_submission_cpu_timestamp();
-  const GpuJob* matching_gpu_job =
+  std::optional<GpuJob> matching_gpu_job_option =
```
dpallotti, line 27: I find the name "option" a bit weird, shouldn't it be "optional"?

src/OrbitCaptureClient/GpuQueueSubmissionProcessor.cpp
```diff
@@ -103,23 +104,23 @@ const GpuQueueSubmission* GpuQueueSubmissionProcessor::FindMatchingGpuQueueSubmi
...
-    return nullptr;
+  if (matching_gpu_submission.meta_info().pre_submission_cpu_timestamp() > submit_time) {
+    return std::nullopt;
   }
 
   return matching_gpu_submission;
 }
 
-const GpuJob* GpuQueueSubmissionProcessor::FindMatchingGpuJob(
+std::optional<GpuJob> GpuQueueSubmissionProcessor::FindMatchingGpuJob(
```
dpallotti, line 118: In a way I'm still a bit confused. Was the map modified after this function was called? I'm saying that because now a copy is made, which is fine performance-wise, yet feels conceptually wrong.

dimitry- (approved):

dpallotti (approved):

src/OrbitCaptureClient/GpuQueueSubmissionProcessor.cpp
florian-kuebler: To be honest, I don't see where the map may get modified after I retrieved the element. But still this seemed to be the issue. Maybe some weird thing it the MS world...

dimitry-: should probably take another look at this then?

src/OrbitCaptureClient/include/OrbitCaptureClient/GpuQueueSubmissionProcessor.h
```diff
@@ -85,13 +85,15 @@ class GpuQueueSubmissionProcessor {
...
       uint64_t post_submission_cpu_timestamp);
 
   // Finds the GpuQueueSubmission that fully contains the given timestamp and happened on the given
-  // thread id. Returns `nullptr` if there is no such submission.
-  [[nodiscard]] const orbit_grpc_protos::GpuQueueSubmission* FindMatchingGpuQueueSubmission(
+  // thread id. Returns `nullopt` if there is no such submission.
+  [[nodiscard]] std::optional<orbit_grpc_protos::GpuQueueSubmission> FindMatchingGpuQueueSubmission(
       int32_t thread_id, uint64_t submit_time);
   [[nodiscard]] bool HasUnprocessedBeginMarkers(int32_t thread_id,
                                                 uint64_t post_submission_timestamp) const;
```
dimitry-, line 99: out of curiosity does changing flat_hash_map to node_hash_map help?

florian-kuebler: Well, the issue is in the inner std::map (which I use for the ordering (upper_bound/lower_bound).

dimitry-: still if this is an implicit move by outer map it could do something to the outer container. It could be some implementation specific internal data representation of inner map which is tied to an instance and might get cleared if outer container is modified.

src/OrbitCaptureClient/GpuQueueSubmissionProcessor.cpp
florian-kuebler: Okay, I found out what was the original issue and am actually a bit shocked how this never blown up on Linux and only occasionally on Windows.

When looking at this method: We extract a lot of information from the "completed marker", which is a member of the "gpu_queue_submission". Originally that was a pointer to an element inside the map.

Now, if that "completed" maker has information about a begin marker (the true case in line 312), we do some further processing. In particular we set the "matching_begin_submission" to the "gpu_queue_submission", if the meta data is equal. After we are done with the processing of the "matching_begin_submission" we call DecrementUnprocessedBeginMarkers, to eventually clean up state that we might not need anymore, this will (in some cases) remove the "matching_begin_submission" from the map, BUT, in this case it is equal to the "gpu_queue_submission", and this is used right after the method call, so right after it was erased....

I'll move reads of the "gpu_queue_submission" upfront, so that the erase is the last thing that would be done. WDYT?
Especially about std::option vs. pointer inside the std::map.

src/OrbitCaptureClient/GpuQueueSubmissionProcessor.cpp
florian-kuebler: Obsolete now.

src/OrbitCaptureClient/GpuQueueSubmissionProcessor.cpp
florian-kuebler: I changed it, by doing the "Decrement.."  and thus the possible erase operation as last step.

florian-kuebler:
I changed it a bit, such that the call to "Decrement..." will be the last thing done. Please have another look.

src/OrbitCaptureClient/include/OrbitCaptureClient/GpuQueueSubmissionProcessor.h
```diff
@@ -85,13 +85,15 @@ class GpuQueueSubmissionProcessor {
...
+  // Finds the GpuJob that is fully inside the given timestamps and happened on the given thread id.
+  // Returns `nullopt` if there is no such job.
   [[nodiscard]] const orbit_grpc_protos::GpuJob* FindMatchingGpuJob(
       int32_t thread_id, uint64_t pre_submission_cpu_timestamp,
       uint64_t post_submission_cpu_timestamp);
 
   // Finds the GpuQueueSubmission that fully contains the given timestamp and happened on the given
-  // thread id. Returns `nullptr` if there is no such submission.
-  [[nodiscard]] const orbit_grpc_protos::GpuQueueSubmission* FindMatchingGpuQueueSubmission(
+  // thread id. Returns `nullopt` if there is no such submission.
```
dimitry-, line 95: this no longer returns nullopt

src/OrbitCaptureClient/include/OrbitCaptureClient/GpuQueueSubmissionProcessor.h
```diff
@@ -85,13 +85,15 @@ class GpuQueueSubmissionProcessor {
...
+  [[nodiscard]]const orbit_grpc_protos::GpuQueueSubmission* FindMatchingGpuQueueSubmission(
```
dimitry-, line 96: space.

src/OrbitCaptureClient/include/OrbitCaptureClient/GpuQueueSubmissionProcessor.h
dimitry-: I would still use node_map, it is pointer to a map which has pointer stability but I still think moving map does not have to provide same guarantee. It likely does but still.

src/OrbitCaptureClient/GpuQueueSubmissionProcessor.cpp
```diff
@@ -360,9 +371,12 @@ std::vector<TimerInfo> GpuQueueSubmissionProcessor::ProcessGpuDebugMarkers(
         marker_timer.set_thread_id(kUnknownThreadId);
       }
 
-      DecrementUnprocessedBeginMarkers(begin_marker_thread_id,
-                                       matching_begin_job->amdgpu_cs_ioctl_time_ns(),
-                                       begin_marker_post_submission_cpu_timestamp);
+	  // Remember, it would be safe to decrement (and thus possible erease) the "begin marker"
```
dpallotti, line 374: it would *not* be safe?

dpallotti, line 374: Remove one indentation.

dpallotti, line 374: erease -> erase

src/OrbitCaptureClient/include/OrbitCaptureClient/GpuQueueSubmissionProcessor.h
```diff
@@ -85,13 +85,15 @@ class GpuQueueSubmissionProcessor {
   [[nodiscard]] static std::optional<orbit_grpc_protos::GpuCommandBuffer> ExtractFirstCommandBuffer(
       const orbit_grpc_protos::GpuQueueSubmission& gpu_queue_submission);
 
+  // Finds the GpuJob that is fully inside the given timestamps and happened on the given thread id.
+  // Returns `nullopt` if there is no such job.
```
dpallotti, line 89: This also doesn't return `nullopt` now.

dpallotti (approved):

src/OrbitCaptureClient/GpuQueueSubmissionProcessor.cpp
florian-kuebler: Done

florian-kuebler: Done.

florian-kuebler: Done.

src/OrbitCaptureClient/include/OrbitCaptureClient/GpuQueueSubmissionProcessor.h
florian-kuebler: Ah thanks

src/OrbitCaptureClient/include/OrbitCaptureClient/GpuQueueSubmissionProcessor.h
florian-kuebler: Thanks.

src/OrbitCaptureClient/include/OrbitCaptureClient/GpuQueueSubmissionProcessor.h
florian-kuebler: Sure.

florian-kuebler (commented):
Thanks for your reviews. I addressed the comments.

src/OrbitCaptureClient/include/OrbitCaptureClient/GpuQueueSubmissionProcessor.h
florian-kuebler: Done.

PR #1716, opened 2021-01-21
Add new icon for the capture toolbar extension button

vickyliu-go4it:
The original capture toolbar extension button is hard to see. We add a new icon to make it more obvious.

Bug: http://b/178092318
Test: Move the splitter to the left side to see the extension button (i.e., a double-arrow on the right side)


vickyliu-go4it:
Before the change:
original extension button

After the change: 
new extension button

karupayun (approved):
LGTM!

PR #1717, opened 2021-01-21
LinuxTracingIntegrationTest: Lay foundations and add dlopen (mmap) test

dpallotti:
Fork a process and trace it. Send commands to that process via stdin/stdout
redirected to pipe. Inspect the collected `CaptureEvent`s.

Bug: http://b/160756664


src/LinuxTracing/LinuxTracingIntegrationTest.cpp
```diff
@@ -0,0 +1,301 @@
...
+TEST(LinuxTracingIntegrationTest, ModuleUpdateOnDlopen) {
+  if (!CheckIsPerfEventParanoidAtMost(0)) GTEST_SKIP();
+  LinuxTracingIntegrationTestFixture fixture;
+
+  fixture.StartTracing();
+  SleepMs(100);
+  fixture.WriteLineToPuppet(PuppetConstants::kDlopenCommand);
+  while (fixture.ReadLineFromPuppet() != PuppetConstants::kDoneResponse) {
+  }
+  SleepMs(100);
```
dimitry-, line 274: absl::Sleep?

src/LinuxTracing/LinuxTracingIntegrationTest.cpp
```diff
@@ -0,0 +1,301 @@
...
+  CHECK(atoi_succeeded);
+  return perf_event_paranoid;
+}
+
+[[nodiscard]] bool CheckIsPerfEventParanoidAtMost(int max_perf_event_paranoid) {
+  if (IsRunningAsRoot()) {
+    return true;
+  }
+
+  int perf_event_paranoid = ReadPerfEventParanoid();
```
dimitry-, line 51: I can see the value for running locally - but once we have tests for uprobes paranoid will not be enough afaik.

dimitry- (approved):
otherwise lgtm

src/LinuxTracing/LinuxTracingIntegrationTest.cpp
```diff
@@ -0,0 +1,301 @@
...
+  std::optional<Tracer> tracer_ = std::nullopt;
+  std::optional<BufferTracerListener> listener_ = std::nullopt;
+};
+
+void SleepMs(uint64_t ms) { std::this_thread::sleep_for(std::chrono::milliseconds{ms}); }
+
+using PuppetConstants = LinuxTracingIntegrationTestPuppetConstants;
+
+TEST(LinuxTracingIntegrationTest, ModuleUpdateOnDlopen) {
+  if (!CheckIsPerfEventParanoidAtMost(0)) GTEST_SKIP();
```
dimitry-, line 266: << "This test should be run under root or with perf event paranoid value at most 0"

dimitry-: also {}

src/LinuxTracing/LinuxTracingIntegrationTest.cpp
```diff
@@ -0,0 +1,301 @@
...
+  LinuxTracingIntegrationTestFixture fixture;
+
+  fixture.StartTracing();
+  SleepMs(100);
+  fixture.WriteLineToPuppet(PuppetConstants::kDlopenCommand);
+  while (fixture.ReadLineFromPuppet() != PuppetConstants::kDoneResponse) {
```
dimitry-, line 272: continue

src/LinuxTracing/LinuxTracingIntegrationTest.cpp
dpallotti: That's why I added `[[nodiscard]] __attribute__((unused)) bool CheckIsRunningAsRoot()` above, I will use that for uprobes, but I want each test to have the exact condition.

src/LinuxTracing/LinuxTracingIntegrationTest.cpp
dpallotti: That doesn't print the message. `CheckIsPerfEventParanoidAtMost` prints a message, though.

src/LinuxTracing/LinuxTracingIntegrationTest.cpp
dpallotti: Done.

src/LinuxTracing/LinuxTracingIntegrationTest.cpp
dpallotti: Done.

src/LinuxTracing/LinuxTracingIntegrationTest.cpp
```diff
@@ -0,0 +1,300 @@
...
+    return str;
+  }
+
+ private:
+  pid_t child_pid_ = -1;
+  int reading_fd_ = -1;
+  int writing_fd_ = -1;
+};
+
+class BufferTracerListener : public TracerListener {
```
florian-kuebler, line 137: Maybe BufferedTracerListener?

src/LinuxTracing/LinuxTracingIntegrationTest.cpp
```diff
@@ -0,0 +1,300 @@
...
+    events_.emplace_back(std::move(event));
+  }
+
+  void OnModulesUpdate(orbit_grpc_protos::ModuleUpdateEvent module_update_event) override {
+    orbit_grpc_protos::CaptureEvent event;
+    *event.mutable_module_update_event() = std::move(module_update_event);
+    events_.emplace_back(std::move(event));
+  }
+
+  [[nodiscard]] const std::vector<orbit_grpc_protos::CaptureEvent>& GetEvents() const {
```
florian-kuebler, line 199: Again, according to the style guide, simple getters should be snake_case

src/LinuxTracing/LinuxTracingIntegrationTestPuppet.h
```diff
@@ -0,0 +1,25 @@
+// Copyright (c) 2021 The Orbit Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#ifndef LINUX_TRACING_INTEGRATION_TEST_PUPPET_H_
+#define LINUX_TRACING_INTEGRATION_TEST_PUPPET_H_
+
+namespace orbit_linux_tracing {
```
florian-kuebler, line 8: It would probably help to have a high level description here.

florian-kuebler (approved):
LGTM. Thank you for this change!

src/LinuxTracing/LinuxTracingIntegrationTest.cpp
```diff
@@ -0,0 +1,300 @@
...
+  }
+
+  ~ChildProcess() {
+    CHECK(close(reading_fd_) == 0);
+    CHECK(close(writing_fd_) == 0);
+
+    CHECK(waitpid(child_pid_, nullptr, 0) == child_pid_);
+  }
+
+  [[nodiscard]] pid_t GetChildPid() const { return child_pid_; }
```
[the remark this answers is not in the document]
dpallotti: As below.

src/LinuxTracing/LinuxTracingIntegrationTest.cpp
dpallotti: The styleguide says "may", i.e., it's a preference. Personally, I don't mind the snake_case when we have a struct or class that contains only or mostly data and there's almost a one-to-one relation between fields and methods (something like a proto), but I don't like it when that's not the case and the result is just a mix of styles.

src/LinuxTracing/LinuxTracingIntegrationTest.cpp
dpallotti: I think that "buffered" gives the wrong message, I just want to say that this `TracerListener` simply *has* a buffer/vector/list where events are stored. Maybe "buffer" is not ideal? I could also just call this `SimpleTracerListener` if that's less confusing.

src/LinuxTracing/LinuxTracingIntegrationTestPuppet.h
dpallotti: There's a brief high level description in the .cpp file. Should I move it here? The reason why it's not here is that this `LinuxTracingIntegrationTestPuppet` is very similar to an executable (which would only have the .cpp), the header is mostly to share constants.

florian-kuebler: Okay, I missed the one in the cpp file. Is fine for me.

src/LinuxTracing/LinuxTracingIntegrationTest.cpp
florian-kuebler: I am only concerned about the project-wide style. As far as I remember, by now, most simple getters are snake_case. But I am also not strong about this. I'll leave it to you. But we definitely already have classes where it's a big mixture of PascalCase and snake_case.

src/LinuxTracing/LinuxTracingIntegrationTest.cpp
florian-kuebler: Okay, then let's keep BufferTracerListener.

src/LinuxTracing/LinuxTracingIntegrationTest.cpp
dpallotti: > But we definitely already have classes where it's a big mixture of PascalCase and snake_case.

I know, I don't like it... In my opinion that should be reserved for "data" classes.

Take `GetChildPid` above for example. The reason why it would make sense to use `child_pid` is because you know that that's a field. But the field is private, so that's just an implementation detail, the class could have another way of retrieving that pid. While for data classes, it's normal that the accessor is based on a field.

In any case, there classes are private to the test file so it doesn't matter.

PR #1719, opened 2021-01-22
Set -Werror=defaulted-function-deleted only for clang 8 and above

pierricgimmig:
The flag appeared in clang 8. Our CI uses clang 7 which does not
recognize the flag and spammed a lot of warnings.

This removes around 300 occurrences of :
"warning: unknown warning option '-Werror=defaulted-function-deleted'; 
did you mean '-Werror=bad-function-cast'? [-Wunknown-warning-option]"


beckerhe (approved):
Thanks. LGTM!

dimitry- (approved):

PR #1720, opened 2021-01-22
Fix CMakeLists.txt for fuzz testing

antonrohr:
These are fixes for the fuzz tests.


danielfenner (approved):
lgtm

src/OrbitGl/CMakeLists.txt
```diff
@@ -140,6 +140,8 @@ target_link_libraries(
 
 if(TARGET OpenGL::GLX AND TARGET OpenGL::OpenGL)
   target_link_libraries(OrbitGl PUBLIC OpenGL::GLX)
+elseif(TARGET OpenGL::GL)
+  target_link_libraries(OrbitGl PUBLIC OpenGL::GL)
```
beckerhe, line 144: I hope that doesn't break anything. I still remember it was a hassle finding a way that works on platforms. Fingers crossed.

beckerhe: (Just wanted to mention that, to raise awareness that upcoming opengl-problems might be related with that change.)

If it compiles on the CI, I'm fine with it.

beckerhe (approved):
Sorry for the delay. LGTM! Thanks Anton!

src/OrbitGl/CMakeLists.txt
antonrohr: I am somewhat afraid too, but it seems to be working fine so far.  
I think its also somewhat elegant in the way that it only actually links when the target is available. 

I have to admit, I also did not test this on my windows machine yet, but relied on the CI. I will try it on my machine before I merge. (monday morning)

beckerhe: SGTM!

PR #1721, opened 2021-01-22
Add Promise/Future support

beckerhe:
This adds (hopefully) properly tested standard facilities to tell the caller when a background task has been finished.

ThreadPool::Schedule will return a future-type which is connected to the background job. This future can be actively
waited for by calling `MainThreadExecutor::WaitFor`. MainThreadExecutor will take care of scheduling other work on
the main thread while the wait is happening.

The future currently does not have a `Then`-method as it's typical in other programming languages. But this will come
in the future. All facilities are there to chain a main-thread-job to a background-job.

### What is the purpose of all this?

We currently have a lot of code of this structure:
```
thread_pool_->Schedule([&]() {
  // do some background work
  main_thread_executor_->Schedule([&]() {
    // implement the result of the background work in the UI.
  });
});
```

For a lot of use-cases that's fine, but it's not really extensible. In a lot of cases we want to
execute the same background job, wait for it to finish and then react to it on the main-thread.
In the previous design background-task and follow-up-task are tightly coupled. Due to the introduction
of futures, the main-thread-task can be decoupled from the background-task.

### Example

The declaration of `OrbitApp::LoadModules` currently looks like this:
```
  void LoadModules(                                                                                                     
      const std::vector<ModuleData*>& modules,                                                                          
      absl::flat_hash_map<std::string, std::vector<uint64_t>> function_hashes_to_hook_map = {},                         
      absl::flat_hash_map<std::string, std::vector<uint64_t>> frame_track_function_hashes_map = {});
```

The parameters `function_hashes_to_hook_map` and `frame_track_function_hashes_map` are basically
actions that need to happen on the main thread after asynchronously loading the modules in multiple
background threads.

This violates the separation of concern-principle and is very hard to extend. With futures I can declare
something like this:
```
orbit_base::Future<void> LoadModule(ModuleData* module);
```

This interface means the caller can decide if they want to load all modules in parallel and which task to execute
afterwards.

This particular change is not part of the PR, but will follow in another one.


src/OrbitBase/include/OrbitBase/ThreadPool.h
```diff
@@ -38,8 +40,28 @@ class ThreadPool {
...
-    Schedule(CreateAction(std::forward<F>(functor)));
+  auto Schedule(F&& functor) -> orbit_base::Future<std::decay_t<decltype(functor())>> {
+    using ReturnType = std::decay_t<decltype(functor())>;
+
+    orbit_base::Promise<ReturnType> promise;
+    orbit_base::Future<ReturnType> future = promise.GetFuture();
+
+    if constexpr (std::is_same_v<ReturnType, void>) {
+      auto function_wrapper = [functor = std::forward<F>(functor),
+                               promise = std::move(promise)]() mutable {
```
dimitry-, line 51: why is it mutable?

src/OrbitBase/include/OrbitBase/SharedState.h
```diff
@@ -0,0 +1,38 @@
...
+#define ORBIT_BASE_SHARED_STATE_H_
+
+#include <absl/synchronization/mutex.h>
+
+#include <functional>
+#include <optional>
+#include <variant>
+#include <vector>
+
+namespace orbit_base {
```
dimitry-, line 15: maybe use orbit_base_internal here since these are not supposed to be used outside of Promise/Capture

src/OrbitBase/ThreadPoolTest.cpp
```diff
@@ -378,3 +381,120 @@ TEST(ThreadPool, WaitWithoutShutdown) {
...
+
+    EXPECT_TRUE(future.IsValid());
+    EXPECT_FALSE(future.IsFinished());
+
+    EXPECT_TRUE(mutex.AwaitWithTimeout(absl::Condition(
+                                           +[](bool* called) { return *called; }, &called),
+                                       absl::Milliseconds(100)));
+
+    for (int elapsed_ms = 0; !future.IsFinished() && elapsed_ms < 100; ++elapsed_ms) {
+      std::this_thread::sleep_for(std::chrono::milliseconds{1});
```
dimitry-, line 410: nit: absl::Sleep?

src/OrbitBase/FutureTest.cpp
```diff
@@ -0,0 +1,53 @@
...
+// found in the LICENSE file.
+
+#include <gtest/gtest.h>
+
+#include "OrbitBase/Future.h"
+#include "OrbitBase/Promise.h"
+
+namespace orbit_base {
+
+TEST(Future, IsValid) {
```
dimitry-, line 12: In spirit of testing scenarios and not methods: rename this to Create?

src/OrbitBase/FutureTest.cpp
```diff
@@ -0,0 +1,53 @@
...
+  Promise<void> promise{};
+  Future<void> future = promise.GetFuture();
+  EXPECT_TRUE(future.IsValid());
+}
+
+TEST(Future, IsFinished) {
```
dimitry-, line 18: and this one to MarkFinished

src/OrbitBase/FutureTest.cpp
```diff
@@ -0,0 +1,53 @@
...
+  Promise<void> promise{};
+  Future<void> future = promise.GetFuture();
+  EXPECT_FALSE(future.IsFinished());
+
+  promise.MarkFinished();
+  EXPECT_TRUE(future.IsFinished());
+}
+
+TEST(Future, Get) {
```
dimitry-, line 27: FinishedFutureResult?

src/OrbitBase/FutureTest.cpp
```diff
@@ -0,0 +1,53 @@
...
+  ASSERT_TRUE(future.IsFinished());
+
+  Future<int> future2 = std::move(future);
+  EXPECT_FALSE(future.IsValid());
+  EXPECT_FALSE(future.IsFinished());
+
+  ASSERT_TRUE(future2.IsValid());
+  ASSERT_TRUE(future2.IsFinished());
+  EXPECT_EQ(*future2.Get().value(), 42);
+}
```
dimitry-, line 52: when is Future expected to be false? maybe add a test for that too?

src/OrbitBase/PromiseTest.cpp
```diff
@@ -0,0 +1,72 @@
...
+  Promise<int> promise_int{};
+  EXPECT_TRUE(promise_int.IsValid());
+}
+
+TEST(Promise, IsFinished) {
+  Promise<void> promise{};
+  EXPECT_FALSE(promise.IsFinished());
+}
+
+TEST(Promise, HasResult) {
```
dimitry-, line 24: NewPromiseDoesNotHaveResult

dimitry- (commented):
I am going to take another look at the CL later today, these are comments I have so far.

src/OrbitBase/include/OrbitBase/ThreadPool.h
beckerhe: It mutates `promise` by calling `promise.MarkFinished()`.

src/OrbitBase/include/OrbitBase/SharedState.h
beckerhe: Good point. I wasn't sure about our policy here. I like just adding the `_internal` suffix.

src/OrbitBase/include/OrbitBase/ThreadPool.h
beckerhe: Hmm. The more I think about it. Maybe `MarkFinished` should be `const`.

beckerhe: Circling back to this: Making `MarkFinished` const is probably also not a good idea. I guess I prefer to leave the lambda `mutable`.

src/OrbitBase/include/OrbitBase/SharedState.h
beckerhe: Done.

src/OrbitBase/ThreadPoolTest.cpp
beckerhe: Done.

src/OrbitBase/FutureTest.cpp
beckerhe: Sure. I'm very bad at naming tests. So more feedback like that is very much appreciated!

src/OrbitBase/FutureTest.cpp
beckerhe: Yes. Done.

antonrohr:
I will look at this in detail on monday. At first glance it looks pretty awesome

src/OrbitBase/ThreadPoolTest.cpp
```diff
@@ -378,3 +380,120 @@ TEST(ThreadPool, WaitWithoutShutdown) {
...
+      absl::MutexLock lock(&mutex);
+      called = true;
+    });
+
+    EXPECT_TRUE(future.IsValid());
+    EXPECT_FALSE(future.IsFinished());
+
+    EXPECT_TRUE(mutex.AwaitWithTimeout(absl::Condition(
+                                           +[](bool* called) { return *called; }, &called),
+                                       absl::Milliseconds(100)));
```
pierricgimmig, line 406: Isn't it possible for the lambda to start executing before we release the mutex here, in which case we would lock twice? What is preventing the future from being finished by the time "EXPECT_FALSE(future.IsFinished());" is called?

pierricgimmig: Ah nevermind, got it, the mutex inside the lambda will wait until we release it with AwaitWithTimeout

src/OrbitBase/ThreadPoolTest.cpp
```diff
@@ -378,3 +380,120 @@ TEST(ThreadPool, WaitWithoutShutdown) {
...
+
+    for (int elapsed_ms = 0; !future.IsFinished() && elapsed_ms < 100; ++elapsed_ms) {
+      absl::SleepFor(absl::Milliseconds(1));
+    }
```
pierricgimmig, line 410: What is the purpose of this loop? At this point we know the lambda has been executed because it released the mutex. We are waiting for the future to be marked as finished. Why not call future.get()? Isn't there a chance that the marking as finished take more than 100ms?

src/OrbitBase/include/OrbitBase/Future.h
```diff
@@ -0,0 +1,140 @@
...
+
+   A valid Future<T> is created from a Promise<T>. The latter lives in the
+   asynchronous task and its purpose is to notify the Future<T> when the result
+   is available.
+
+   Use Future<T>::IsValid() to check if a future is connected to a promise or holds
+   a result value.
+
+   Call Future<T>::IsFinished() to figure out if the result is already available.
+   If yes, you can call Future<T>::Get() to retrieve the T.
```
pierricgimmig, line 62: I would have expected that calling Get() would block until the result is available like std::future<T>::get. Does this mean that the user always has to implement a loop to check IsFinished() before getting the result?

beckerhe: Good point. I will change that to a blocking behaviour!

src/OrbitBase/ThreadPoolTest.cpp
beckerhe: Yes, this part is not ideal. But this will change with your suggested behaviour change of Get(). Thanks for raising!

src/QtUtils/include/QtUtils/FutureWatcher.h
```diff
@@ -0,0 +1,59 @@
...
+#include <absl/types/span.h>
+
+#include <QObject>
+#include <chrono>
+
+#include "OrbitBase/Future.h"
+
+namespace orbit_qt_utils {
+
+/*
```
dimitry-, line 14: nit: maybe use // comment style everywhere? I think we use it by default

dimitry- (approved):
otherwise lgtm

src/OrbitBase/include/OrbitBase/Future.h
```diff
@@ -0,0 +1,140 @@
...
+ public:
+  [[nodiscard]] bool IsFinished() const {
+    if (shared_state_.use_count() == 0) return false;
+
+    absl::MutexLock lock{&shared_state_->mutex};
+    return shared_state_->finished;
+  }
+
+  // Check Future<T>::RegisterContinuation for a warning about this function.
+  // TLDR: You probably want to use `Future<void>::Then` instead!
```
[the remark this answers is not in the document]
beckerhe: Not yet. `Then` needed a but more thought and will follow in another PR. I hope I have the design sorted out by now. Still have to implement it though.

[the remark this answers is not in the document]
beckerhe: @dimitry- Do you prefer to have it in this PR? I could also still add it.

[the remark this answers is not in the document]
dimitry-: this is fine, next PR sounds good!

src/QtUtils/include/QtUtils/FutureWatcher.h
```diff
@@ -0,0 +1,59 @@
...
+*/
+class FutureWatcher : public QObject {
+  Q_OBJECT
+ public:
+  explicit FutureWatcher(QObject* parent = nullptr);
+
+  enum class Reason { kFutureCompleted, kTimeout, kAbortRequested };
+
+  [[nodiscard]] Reason WaitFor(const orbit_base::Future<void>& future,
+                               std::optional<std::chrono::milliseconds> timeout) const;
```
antonrohr, line 48: Did you consider `absl::Duration`?

src/OrbitBase/ThreadPoolTest.cpp
beckerhe: Done.

src/OrbitBase/include/OrbitBase/Future.h
beckerhe: Done.

src/OrbitGl/MainThreadExecutor.h
```diff
@@ -35,9 +40,42 @@ class MainThreadExecutor {
   virtual void Schedule(std::unique_ptr<Action> action) = 0;
 
   template <typename F>
-  void Schedule(F&& functor) {
-    Schedule(CreateAction(std::forward<F>(functor)));
+  auto Schedule(F&& functor) -> orbit_base::Future<std::decay_t<decltype(functor())>> {
```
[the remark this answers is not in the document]
beckerhe: Done.

src/OrbitQt/MainThreadExecutorImpl.cpp
```diff
@@ -6,29 +6,80 @@
...
+    case Reason::kFutureCompleted:
+      return WaitResult::kCompleted;
+    case Reason::kAbortRequested:
+      return WaitResult::kAborted;
+    case Reason::kTimeout:
+      return WaitResult::kTimedOut;
+  }
+
+  UNREACHABLE();
+  return WaitResult::kTimedOut;
```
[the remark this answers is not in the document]
beckerhe: Fixed. Thanks!

src/QtUtils/FutureWatcher.cpp
```diff
@@ -0,0 +1,136 @@
...
+  // Ensure that no race condition occured. The futured could have completed while we were setting
+  // the continuation.
+  if (future.IsFinished()) return Reason::kFutureCompleted;
+
+  loop.exec();
+
+  if (future.IsFinished()) return Reason::kFutureCompleted;
+
+  constexpr int kInvalidRemainingTime = -1;
+  if (timeout && timer.remainingTime() == kInvalidRemainingTime) return Reason::kTimeout;
```
antonrohr, line 50: Above you used `timeout.has_value()`. Please be consistent

src/QtUtils/FutureWatcherTest.cpp
```diff
@@ -0,0 +1,209 @@
...
+  FutureWatcher watcher{};
+  const auto reason = watcher.WaitFor(std::move(future), std::chrono::milliseconds{5});
+
+  EXPECT_EQ(reason, FutureWatcher::Reason::kTimeout);
+
+  mutex.Unlock();
+  thread_pool->ShutdownAndWait();
+}
+
+TEST(FutureWatcher, WaitForAllWithThreadPool) {
```
antonrohr, line 132: Maybe a test `WaitForAll({}, std::nullopt)` (empty input) is a good idea.

src/QtUtils/include/QtUtils/FutureWatcher.h
beckerhe: Good catch, I wasn't aware of that. Fixed.

antonrohr (approved):
Cannot wait for this to make the symbol loading code simpler! Only a few minor comments

src/OrbitBase/include/OrbitBase/Future.h
```diff
@@ -74,12 +77,11 @@ class Future : public orbit_base_internal::FutureBase<T> {
...
+  const T& Get() const {
+    Wait();
 
     absl::MutexLock lock{&this->shared_state_->mutex};
-    if (!this->shared_state_->result.has_value()) return std::nullopt;
-    return &this->shared_state_->result.value();
+    return this->shared_state_->result.value();
   }
 
   /*
```
dimitry-, line 87: and this.

src/OrbitBase/ThreadPoolTest.cpp
```diff
@@ -60,7 +60,7 @@ TEST(ThreadPool, QueuedActionsExecutedOnShutdown) {
   {
     absl::MutexLock lock(&mutex);
     for (size_t i = 0; i < kNumberOfActions; ++i) {
-      thread_pool->Schedule([&]() {
+      (void)thread_pool->Schedule([&]() {
```
dimitry-, line 63: well, that does not look good :) Maybe have Future<void> without nodiscard? I think this means we want to apply nodiscard to class and omit it for the method.

src/OrbitBase/FutureTest.cpp
```diff
@@ -0,0 +1,54 @@
...
+TEST(Future, Move) {
+  Promise<int> promise{};
+  Future<int> future = promise.GetFuture();
+  EXPECT_FALSE(future.IsFinished());
+
+  promise.SetResult(42);
+  ASSERT_TRUE(future.IsFinished());
+
+  Future<int> future2 = std::move(future);
+  EXPECT_FALSE(future.IsValid());
```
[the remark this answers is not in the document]
beckerhe: Done.

src/QtUtils/FutureWatcher.cpp
beckerhe: Done.

src/QtUtils/FutureWatcherTest.cpp
beckerhe: Good idea. Done.

src/QtUtils/include/QtUtils/FutureWatcher.h
beckerhe: No, I haven't. I was under the impression we prefer std-APIs over third-party-APIs.

src/OrbitBase/ThreadPoolTest.cpp
beckerhe: Yes, I can do that.

beckerhe: Done.

src/OrbitBase/include/OrbitBase/Future.h
beckerhe: Ah yes. Thanks for checking.

beckerhe:
I will wait for Pierric before rebasing and merging.

src/QtUtils/include/QtUtils/FutureWatcher.h
antonrohr: Okay. I don't actually know if there are benefits of absl::Duration over the std:: counterpart. I can only imagine there are some, or otherwise it would not exist. For now leave everything as is. I also dont feel strongly about it.

beckerhe: `std:chrono` has (as of C++17) no support for civil time. Just clocks, time_points and durations. `abseil` offers all of that, including time-zone support. In C++20 `std::chrono` will also get all these features and I expect abseil to slowly moving towards `std::chrono`. Don't know though if this is actually the plan.

src/OrbitBase/include/OrbitBase/Future.h
pierricgimmig: Awesome, thanks!

pierricgimmig (approved):
Hi @beckerhe , thanks for the change, looks really good to me! Cheers

src/QtUtils/include/QtUtils/FutureWatcher.h
antonrohr: Thanks for the explanation, that makes a lot of sense.

PR #1722, opened 2021-01-22
UI fixes after capture toolbar change

reichlfl:
Fixing issues introduced with https://github.com/google/orbit/pull/1695

- Resource file was referenced incorrectly
- Accessible names for "filterFunctions" and "filterTracks" were
changed, causing the track sorting E2E test to fail

This was caught by the E2E tests here: b/178171558


vickyliu-go4it (approved):
LGTM! Thanks for taking care of this. Also sorry that I didn't notice this in my previous changes.

danielfenner (approved):
lgtm!

PR #1723, opened 2021-01-22
Add tests for SchedulingSlices and ThreadStatesSlices to LinuxTracingIntegrationTest

dpallotti:
Bug: http://b/160756664


florian-kuebler (approved):
Well these are two pretty well written and nicely to read tests. LGTM!

Maybe for the future: It seems that there is potential to have a lot of code duplication. If this becomes a problem, we might want to refactor a bit.

dpallotti:
> Maybe for the future: It seems that there is potential to have a lot of code duplication. If this becomes a problem, we might want to refactor a bit.

Yeah, I agree. In particular the part where the capture is taken is basically identical across these tests. But things will get more complicated with sampling and dynamic instrumentation so I want to see how it evolves.

PR #1730, opened 2021-01-25
Add target label to dialog

antonrohr:
This rewords the Ui button to "Start Session" and adds a label to the left of it. 

b/177305224

image

### DIFF
--- a/src/OrbitQt/ConnectToStadiaWidget.h
+++ b/src/OrbitQt/ConnectToStadiaWidget.h
@@ -45,6 +45,9 @@ class ConnectToStadiaWidget : public QWidget {
   [[nodiscard]] const std::shared_ptr<grpc::Channel>& GetGrpcChannel() { return grpc_channel_; }
   void Start(SshConnectionArtifacts* ssh_connection_artifacts,
              std::optional<StadiaConnection> connection = std::nullopt);
+  [[nodiscard]] std::optional<orbit_ggp::Instance> GetSelectedInstance() const {
+    return selected_instance_;
+  }
 
  public slots:
   void SetActive(bool value);
--- a/src/OrbitQt/ProfilingTargetDialog.cpp
+++ b/src/OrbitQt/ProfilingTargetDialog.cpp
@@ -178,9 +178,10 @@ std::optional<TargetConfiguration> ProfilingTargetDialog::Exec() {
 }
 
 void ProfilingTargetDialog::ProcessSelectionChanged(const QModelIndex& current) {
+  emit NoProcessSelected();
+
   if (!current.isValid()) {
     process_ = nullptr;
-    emit NoProcessSelected();
     return;
   }
 
@@ -203,6 +204,7 @@ void ProfilingTargetDialog::SetupStadiaStates() {
   state_stadia_.assignProperty(ui_->stadiaWidget, "active", true);
   state_stadia_.assignProperty(ui_->loadCaptureRadioButton, "checked", false);
   state_stadia_.assignProperty(ui_->localProfilingRadioButton, "checked", false);
+  state_stadia_.assignProperty(ui_->targetLabel, "text", "");
 
   // STATE state_stadia_connecting_
   state_stadia_connecting_.assignProperty(ui_->processesFrame, "enabled", false);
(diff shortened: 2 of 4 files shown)


karupayun (approved):
LGTM! Really like this change. It works perfectly.

src/OrbitQt/ProfilingTargetDialog.cpp
```diff
@@ -254,6 +256,13 @@ void ProfilingTargetDialog::SetupStadiaStates() {
   // STATE s_s_process_selected
   state_stadia_process_selected_.addTransition(this, &ProfilingTargetDialog::NoProcessSelected,
                                                &state_stadia_no_process_selected_);
+  QObject::connect(&state_stadia_process_selected_, &QState::entered, [this] {
```
beckerhe, line 259: This should have `this` as its third argument and the lambda as the fourth. This will tell Qt that `this` needs to be alive, to execute this lambda. Probably that won't be a problem in the current state of the code, but it's a subtle thing. Someone might change the lifetime of `state_stadia_process_selected_` and then this might crash.

src/OrbitQt/ProfilingTargetDialog.cpp
```diff
@@ -320,6 +330,11 @@ void ProfilingTargetDialog::SetupLocalStates() {
   // STATE state_local_process_selected_
   state_local_process_selected_.addTransition(this, &ProfilingTargetDialog::NoProcessSelected,
                                               &state_local_no_process_selected_);
+  QObject::connect(&state_local_process_selected_, &QState::entered, [this] {
```
beckerhe, line 333: Same as my previous comment.

src/OrbitQt/ProfilingTargetDialog.ui
```diff
@@ -182,6 +182,10 @@
               <property name="text">
                <string/>
               </property>
+              <!-- TODO (b/178379643): As part of rethinking the load capture flow, decide if this label is necessary -->
```
beckerhe, line 185: Is this todo preserved by designer changes? If not, is it important enough to impose the manual step of re-adding the comment whenever someones changes this UI file in the designer?

src/OrbitQt/ConnectToStadiaWidget.h
```diff
@@ -45,6 +45,7 @@ class ConnectToStadiaWidget : public QWidget {
   [[nodiscard]] const std::shared_ptr<grpc::Channel>& GetGrpcChannel() { return grpc_channel_; }
   void Start(SshConnectionArtifacts* ssh_connection_artifacts,
              std::optional<StadiaConnection> connection = std::nullopt);
+  std::optional<orbit_ggp::Instance> GetSelectedInstance() { return selected_instance_; }
```
beckerhe, line 48: nodiscard and const?

beckerhe (approved):
Otherwise LGTM!

src/OrbitQt/ProfilingTargetDialog.cpp
antonrohr: I actually thought there wasn't a difference between the two versions. But you are right. 

(https://wiki.qt.io/New_Signal_Slot_Syntax)
> There is no automatic disconnection when the 'receiver' is destroyed because it's a functor with no QObject. However, since 5.2 there is an overload which adds a "context object". When that object is destroyed, the connection is broken (the context is also used for the thread affinity: the lambda will be called in the thread of the event loop of the object used as context).

I might go through other ui code and look for other times `this` was omitted. 

Thank you!

src/OrbitQt/ProfilingTargetDialog.ui
antonrohr: You are right, they are not preserved between saves in designer. I removed the comment and added a note about this in b/178379643 instead. Thanks for catching that!

src/OrbitQt/ProfilingTargetDialog.cpp
antonrohr: Found one more in `ConnectToStadiaWidget.cpp`. done

src/OrbitQt/ConnectToStadiaWidget.h
antonrohr: done.

beckerhe:
Cool. Thanks Anton!

PR #1731, opened 2021-01-25
Re-add metrics uploader

IrinaShkviro:
Creation of MetricsUploader was removed in one of the latest PRs. But I
think that it's too early to deprecate this feature, so I add it again.

Bug: http://b/170389366.

Test: start Orbit, check that the ORBIT_INITIALIZED and
ORBIT_CAPTURE_DURATION log messages appeared on the server.

### DIFF
--- a/src/OrbitQt/main.cpp
+++ b/src/OrbitQt/main.cpp
@@ -39,6 +39,7 @@
 #include "Connections.h"
 #include "DeploymentConfigurations.h"
 #include "ImGuiOrbit.h"
+#include "MetricsUploader/MetricsUploader.h"
 #include "OrbitBase/Logging.h"
 #include "OrbitSsh/Context.h"
 #include "OrbitSshQt/ScopedConnection.h"
@@ -84,6 +85,16 @@ void RunUiInstance(const DeploymentConfiguration& deployment_configuration,
 
   std::optional<orbit_qt::TargetConfiguration> target_config;
 
+  ErrorMessageOr<orbit_metrics_uploader::MetricsUploader> metrics_uploader = ErrorMessage("");
+  if (absl::GetFlag(FLAGS_enable_metrics)) {
+    metrics_uploader = orbit_metrics_uploader::MetricsUploader::CreateMetricsUploader();
+    if (metrics_uploader.has_value()) {
+      LOG("MetricsUploader was initialized successfully");
+    } else {
+      ERROR("%s", metrics_uploader.error().message());
+    }
+  }
+
   while (true) {
     {
       orbit_qt::ProfilingTargetDialog target_dialog{&ssh_connection_artifacts,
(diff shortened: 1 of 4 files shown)


beckerhe:
> Creation of MetricsUploader was removed in one of the latest PRs. But I
> think that it's too early to deprecate this feature, so I add it again.

I can't follow. Why was it removed? By accident?

beckerhe:
> > Creation of MetricsUploader was removed in one of the latest PRs. But I
> > think that it's too early to deprecate this feature, so I add it again.
> 
> I can't follow. Why was it removed? By accident?

I just saw your comment in #1694. Yes, it was removed there. But that's not where the accident happened.
The creation of metrics uploader should have been part of the `RunBetaUiInstance` function. This sounds
a lot like a race-condition to me - I haven't checked it though: Anton may have added `RunBetaUiInstance`
and you had added metrics uploader to `RunUiInstance`.

src/OrbitQt/main.cpp
```diff
@@ -84,6 +85,11 @@ void RunUiInstance(const DeploymentConfiguration& deployment_configuration,
 
   std::optional<orbit_qt::TargetConfiguration> target_config;
 
+  ErrorMessageOr<orbit_metrics_uploader::MetricsUploader> metrics_uploader = ErrorMessage("");
```
beckerhe, line 88: Instead of `ErrorMessageOr`, could you use `std::optional<MetricsUploader>` for holding the instance? When it's `ErrorMessageOr` I ask myself where the error message is used. But it isn't which is a bit suprising.

src/OrbitQt/main.cpp
```diff
@@ -84,6 +85,11 @@ void RunUiInstance(const DeploymentConfiguration& deployment_configuration,
...
+  if (absl::GetFlag(FLAGS_enable_metrics)) {
+    metrics_uploader = orbit_metrics_uploader::MetricsUploader::CreateMetricsUploader();
```
beckerhe, line 90: Could you add some log messages here? Maybe printing the error message, if there is one and if the creation succeeds maybe a message saying that the metrics uploader has been loaded properly or something like that?

beckerhe (approved):
Otherwise LGTM

IrinaShkviro:
> > > Creation of MetricsUploader was removed in one of the latest PRs. But I
> > > think that it's too early to deprecate this feature, so I add it again.
> > 
> > 
> > I can't follow. Why was it removed? By accident?
> 
> I just saw your comment in #1694. Yes, it was removed there. But that's not where the accident happened.
> The creation of metrics uploader should have been part of the `RunBetaUiInstance` function. This sounds
> a lot like a race-condition to me - I haven't checked it though: Anton may have added `RunBetaUiInstance`
> and you had added metrics uploader to `RunUiInstance`.

The PR that adds MetricsUploader was opened on December 2, and I think it happened before new UI was merged. Because of it when I've opened the PR I wasn't able to add creation of the MetricsUploader in the right place (just because it wasn't there).

src/OrbitQt/main.cpp
IrinaShkviro: Discussed offline: I've added logging in the `main.cpp` (and leave the `ErrorMessageOr` type), and removed logging from MetricsUploader::Create().

antonrohr (approved):
LGTM! I dont have anything to add to Hennings comments

PR #1732, opened 2021-01-25
Attach to a process via ptrace.

danielfenner:
Attach to and stop all threads of a process. Also detach again.

Test: Unit Test.


src/UserSpaceInstrumentation/Attach.cpp
```diff
@@ -0,0 +1,105 @@
...
+
+bool ThreadExistsInProcess(pid_t pid, pid_t tid) {
+  const auto process_tids = orbit_base::GetTidsOfProcess(pid);
+  return std::find(process_tids.begin(), process_tids.end(), tid) != process_tids.end();
+}
+
+// Attaches to the thread using ptrace.
+// Returns true if the thread was halted and false if the thread did not exist anymore.
+[[nodiscard]] ErrorMessageOr<bool> AttachAndStopThread(pid_t pid, pid_t tid) {
+  if (ptrace(PTRACE_ATTACH, tid, 0, 0) < 0) {
```
dpallotti, line 36: For the third argument you use 0 here but `nullptr` below? Is there a reason for that?

src/UserSpaceInstrumentation/include/UserSpaceInstrumentation/Attach.h
```diff
@@ -0,0 +1,20 @@
+// Copyright (c) 2021 The Orbit Authors. All rights reserved.
```
dpallotti, line 1: This file is in the "include" directory, i.e., it is a "public" header of the module. Is it meant to be included (possibly transitively) by modules that depend on this one? For how I would expect it to be an implementation detail and hence be with the private sources, but you have the bigger picture.

src/UserSpaceInstrumentation/DummyProcessForTest.cpp
```diff
@@ -0,0 +1,93 @@
...
+#include "DummyProcessForTest.h"
+
+#include <chrono>
+#include <cstdio>
+#include <fstream>
+#include <string>
+#include <vector>
+
+#include "OrbitBase/Logging.h"
+#include "absl/time/clock.h"
```
dpallotti, line 13: I think we decided to use `<>` for `absl` headers (there are still many with `""` but iwyu will use `<>`). Here and in the other files.

src/UserSpaceInstrumentation/DummyProcessForTest.cpp
```diff
@@ -0,0 +1,93 @@
...
+namespace orbit_user_space_instrumentation {
+
+namespace fs = std::filesystem;
+
+namespace {
+
+// Create a file at path.
+void Touch(fs::path p) {
+  std::ofstream ofs(p);
+  ofs << "\n";
```
dpallotti, line 24: Well, technically in Linux terminology "touch" creates an empty file.

src/UserSpaceInstrumentation/DummyProcessForTest.cpp
```diff
@@ -0,0 +1,93 @@
...
+  ofs.close();
```
dpallotti, line 25: The destructor should already call `close` if I'm not mistaken.

src/UserSpaceInstrumentation/DummyProcessForTest.h
```diff
@@ -0,0 +1,42 @@
...
+ private:
+  // Busy wait for 15 ms.
+  void DummyWorker();
+  // Busy loop that spawns and joins threads.
+  // The spawned threads perform a busy wait for 15 ms. Four worker threads are kept active until
+  // flag_file_run_child_ is deleted.
+  void DummyWorkload();
+
+  pid_t pid_;
+  std::mutex joinable_mutex_;
```
dpallotti, line 36: You mean `joinable_threads_mutex_`?

src/UserSpaceInstrumentation/DummyProcessForTest.cpp
```diff
@@ -0,0 +1,93 @@
...
+
+void DummyProcessForTest::DummyWorkload() {
+  size_t kNumThreads = 4;
+  std::vector<std::thread> threads;
+  while (fs::exists(flag_file_run_child_) || !threads.empty()) {
+    // Spawn as many threads as there are missing.
+    while (threads.size() < kNumThreads && fs::exists(flag_file_run_child_)) {
+      threads.emplace_back(std::thread(&DummyProcessForTest::DummyWorker, this));
+    }
+    Touch(flag_file_child_started_);
```
dpallotti, line 73: So this file is "touched" every time a new child is created, yes?

src/UserSpaceInstrumentation/DummyProcessForTest.cpp
```diff
@@ -0,0 +1,93 @@
...
+  if (pid_ == 0) {
+    DummyWorkload();
+    exit(0);
+  } else {
+    while (!fs::exists(flag_file_child_started_)) continue;
+  }
+}
+
+DummyProcessForTest::~DummyProcessForTest() {
+  fs::remove(flag_file_child_started_);
```
dpallotti, line 46: Can it happen that after this is executed, `Touch(flag_file_child_started_);` in `DummyWorkload()` is called again (basically a race condition)? Or does this rely on the fact that the forked process will also run the destructor?

src/UserSpaceInstrumentation/DummyProcessForTest.cpp
```diff
@@ -0,0 +1,93 @@
+// Copyright (c) 2021 The Orbit Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+#include "DummyProcessForTest.h"
```
dpallotti, line 4:
```suggestion

#include "DummyProcessForTest.h"
```

src/UserSpaceInstrumentation/Attach.cpp
```diff
@@ -0,0 +1,105 @@
...
+    if (ThreadExistsInProcess(pid, tid)) {
```
dimitry-, line 37: probably better to check errno here

src/UserSpaceInstrumentation/Attach.cpp
dimitry-, line 36: == -1

src/UserSpaceInstrumentation/Attach.cpp
dimitry-: (likely check for ESRCH) call to ThreadExistsInProcess in this case is flaky and could yield invalid result (false positive).

src/UserSpaceInstrumentation/Attach.cpp
```diff
@@ -0,0 +1,105 @@
...
+    } else {
+      return false;
+    }
+  }
+  // Wait for the traced thread to stop. Timeout after one second.
+  int timeout = 1000;
+  while (true) {
+    const int result_waitpid = waitpid(tid, nullptr, WNOHANG);
+    if (result_waitpid == -1) {
+      if (ThreadExistsInProcess(pid, tid)) {
```
dimitry-, line 49: same probably better to check errno for ECHILD (my best guess)

src/UserSpaceInstrumentation/DummyProcessForTest.cpp
```diff
@@ -0,0 +1,93 @@
...
+
+DummyProcessForTest::~DummyProcessForTest() {
+  fs::remove(flag_file_child_started_);
+  fs::remove(flag_file_run_child_);
+  int status;
+  waitpid(pid_, &status, 0);
+  CHECK(WIFEXITED(status));
+}
+
+void DummyProcessForTest::DummyWorker() {
```
dimitry-, line 53: Please replace Dummy with Fake or Test

src/UserSpaceInstrumentation/DummyProcessForTest.h
dimitry-: also we use absl::Mutex by default. That is unless you need a reentrant one.

src/UserSpaceInstrumentation/Attach.cpp
```diff
@@ -0,0 +1,105 @@
...
+      }
+    }
+    if (result_waitpid > 0) {
+      // waitpid returns greater zero occationally for terminated threads. We catch this here so we
+      // don't count them as succefully halted.
+      // This is not understood. The check here is a workaround. Note that this case is extremely
+      // rare in practice - the thread needs to finish exactly when we attach to the process.
+      return ThreadExistsInProcess(pid, tid);
+    }
+    std::this_thread::sleep_for(std::chrono::milliseconds(1));
```
ronaldfw, line 63: How performance critical is this function? Can we afford to spend full milliseconds waiting?

My initial thought here was that I would probably do this similar to the following:

```
uint64_t start_timestamps_ns = TimestampNowInNanoSeconds();
while(true) {

   uint64_t current_timestamp_ns = TimestampNowInNanoSeconds();
   if(current_timestamp_ns - start_timestamp_ns > 1000 * 1000'000) {
     // Timeout
  }
}

```

I think that would return as quickly as possible (the thread is halted), but of course is busy-waiting, which may cause other problems.

src/UserSpaceInstrumentation/Attach.cpp
```diff
@@ -0,0 +1,105 @@
...
+    if (result_waitpid == -1) {
+      if (ThreadExistsInProcess(pid, tid)) {
+        return ErrorMessage(
+            absl::StrFormat("Can not attach to tid \"%d\": \"%s\"", tid, strerror(errno)));
+      } else {
+        return false;
+      }
+    }
+    if (result_waitpid > 0) {
+      // waitpid returns greater zero occationally for terminated threads. We catch this here so we
```
ronaldfw, line 57: This comment confused me as it seems this whole case is for catching the rare case for > 0 is returned for terminated threads. But if I understand correctly, > 0 is the case we actually want to wait for. So the comment only really refers to the additional check if the thread exists in the process. 

Maybe start the comment with something like:
" This is the success case we wait for and where we want to return true. However, sometimes waitipid returns greater zero also for terminated threads..."

src/UserSpaceInstrumentation/Attach.cpp
```diff
@@ -0,0 +1,105 @@
...
+      // don't count them as succefully halted.
+      // This is not understood. The check here is a workaround. Note that this case is extremely
+      // rare in practice - the thread needs to finish exactly when we attach to the process.
+      return ThreadExistsInProcess(pid, tid);
+    }
```
ronaldfw, line 62: Do we have to handle the case waitpid() return value == 0?

src/UserSpaceInstrumentation/Attach.cpp
```diff
@@ -0,0 +1,105 @@
...
+    for (const auto tid : process_tids) {
+      if (halted_tids.count(tid) == 1) {
+        continue;
+      }
+      OUTCOME_TRY(result, AttachAndStopThread(pid, tid));
+      if (result) {
+        halted_tids.insert(tid);
+      }
+    }
+    // Update the tids; if there are new ones run another round and stop them.
```
ronaldfw, line 89: Can this go on forever?

src/UserSpaceInstrumentation/AttachTest.cpp
```diff
@@ -0,0 +1,35 @@
...
+namespace orbit_user_space_instrumentation {
+
+TEST(AttachTest, AttachAndStopProcess) {
+  DummyProcessForTest dummy_process;
+  const pid_t pid = dummy_process.pid();
+
+  const auto result = AttachAndStopProcess(pid);
+  ASSERT_TRUE(result) << result.error().message();
+
+  // Verify that no new threads get spawned - i.e. the process is not running anymore.
```
ronaldfw, line 25: This check works because the dummy process is spawning threads all the time, right? This comment reads a bit like it's possible to check if a process is running by checking if it keeps spawning threads.

src/UserSpaceInstrumentation/CMakeLists.txt
```diff
@@ -0,0 +1,49 @@
+# Copyright (c) 2020 The Orbit Authors. All rights reserved.
```
ronaldfw, line 1: 2020 --> 2021

src/UserSpaceInstrumentation/Attach.cpp
```diff
@@ -0,0 +1,105 @@
...
+  if (ptrace(PTRACE_ATTACH, tid, 0, 0) < 0) {
+    if (ThreadExistsInProcess(pid, tid)) {
+      return ErrorMessage(
+          absl::StrFormat("Can not attach to tid \"%d\": \"%s\"", tid, strerror(errno)));
+    } else {
+      return false;
+    }
+  }
+  // Wait for the traced thread to stop. Timeout after one second.
+  int timeout = 1000;
```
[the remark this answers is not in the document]
danielfenner: I like being explicit with the bitness. The style guide demands "int" though.  Changed to int32_t.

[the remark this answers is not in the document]
danielfenner: done

src/UserSpaceInstrumentation/Attach.cpp
danielfenner: Both are void* so for clarity they should both be nullptr. Changed accordingly.

src/UserSpaceInstrumentation/include/UserSpaceInstrumentation/Attach.h
danielfenner: I'd like to keep this public for now and potentially restrict visibility later. The reason is that I want to use this from LinuxTracing for trying things out. You might be right about this becoming an implementation detail finally.

src/UserSpaceInstrumentation/DummyProcessForTest.cpp
danielfenner: Why that? That seems wrong. 
<...> is intended for system headers and allows the implemtation to handle them differently from other stuff. The only guarantee you really get is that "..." also uses the same resolving mechanism as <...> does (if it can't find it by its own mechanism).

src/UserSpaceInstrumentation/DummyProcessForTest.cpp
danielfenner: true. removed.

src/UserSpaceInstrumentation/DummyProcessForTest.h
danielfenner: done.

src/UserSpaceInstrumentation/DummyProcessForTest.cpp
danielfenner: Yes, it would be enought to do it in the first round. But why bother?

src/UserSpaceInstrumentation/DummyProcessForTest.cpp
danielfenner: Good catch! From my computer:
```
> ls /tmp/file* -la | wc -l
5270
```
The `fs::remove(flag_file_child_started_);` needs to go after the waitpid.

fwiw: The child is not executing the destructor. It calls `exit(0)` after the workload is done and that only triggers static destruction.

src/UserSpaceInstrumentation/DummyProcessForTest.cpp
danielfenner: done

src/UserSpaceInstrumentation/Attach.cpp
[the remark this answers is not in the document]
dpallotti: To my understanding it asks to use `int16`, `int32` etc over `int16_t`, `int32_t` defined in `base`, "unless your project is guaranteed not to depend on //base", which is our case, so I think we are in line with the styleguide with `int32_t`.

src/UserSpaceInstrumentation/DummyProcessForTest.cpp
dpallotti: Well, I don't have a strong opinion, I don't think it was for technical reasons, but more just to visually separate the includes of our code from includes of library code. I mentioned it because of consistency, especially if IWYU goes the other way.

dpallotti (commented):
You have replied to the review but not updated it, right? I should wait to have another look, right?

src/UserSpaceInstrumentation/Attach.cpp
danielfenner: The ==0 means tid has not yet changed state and we loop and call waitpid again.

src/UserSpaceInstrumentation/Attach.cpp
danielfenner: I understand way better now what happens. The thread exits before it gets descheduled and that is what causes waitpid to return. The cases can be disambiguated by the status value. 
Adjusted comment accordingly.

src/UserSpaceInstrumentation/Attach.cpp
danielfenner: Changed to errno. Sometimes errno is ESRCH. Weirdly errno is EPERM occationally.  Added a comment. This is only understandable after reading the kernel sources :/

src/UserSpaceInstrumentation/Attach.cpp
danielfenner: done

src/UserSpaceInstrumentation/Attach.cpp
danielfenner: Turns out this is not needed any more with the improved handling of errors in PTRACE_ATTACH above. If an error is reported here we can just escalate it.

Also added a better comment for the check below that handels terminated processes which were not in exit state during PTRACE_ATTACH.

src/UserSpaceInstrumentation/DummyProcessForTest.cpp
danielfenner: It's not a fake and I don't like TestProcessForTest. Changed to just TestProcess.

src/UserSpaceInstrumentation/Attach.cpp
```diff
@@ -0,0 +1,105 @@
+// Copyright (c) 2020 The Orbit Authors. All rights reserved.
```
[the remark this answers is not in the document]
danielfenner: done

src/UserSpaceInstrumentation/CMakeLists.txt
danielfenner: done

src/UserSpaceInstrumentation/Attach.cpp
```diff
@@ -0,0 +1,105 @@
...
+#include "absl/strings/str_format.h"
+
+namespace orbit_user_space_instrumentation {
+
+using orbit_base::GetTidsOfProcess;
+
+namespace {
+
+bool ThreadExistsInProcess(pid_t pid, pid_t tid) {
+  const auto process_tids = orbit_base::GetTidsOfProcess(pid);
```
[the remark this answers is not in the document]
danielfenner: function was removed entirely

src/UserSpaceInstrumentation/Attach.cpp
```diff
@@ -0,0 +1,105 @@
...
+  return std::find(process_tids.begin(), process_tids.end(), tid) != process_tids.end();
+}
+
+// Attaches to the thread using ptrace.
+// Returns true if the thread was halted and false if the thread did not exist anymore.
+[[nodiscard]] ErrorMessageOr<bool> AttachAndStopThread(pid_t pid, pid_t tid) {
+  if (ptrace(PTRACE_ATTACH, tid, 0, 0) < 0) {
+    if (ThreadExistsInProcess(pid, tid)) {
+      return ErrorMessage(
+          absl::StrFormat("Can not attach to tid \"%d\": \"%s\"", tid, strerror(errno)));
```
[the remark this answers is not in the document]
danielfenner: comment changed in the meantime

src/UserSpaceInstrumentation/Attach.cpp
```diff
@@ -0,0 +1,105 @@
...
+    }
+  }
+  // Wait for the traced thread to stop. Timeout after one second.
+  int timeout = 1000;
+  while (true) {
+    const int result_waitpid = waitpid(tid, nullptr, WNOHANG);
+    if (result_waitpid == -1) {
+      if (ThreadExistsInProcess(pid, tid)) {
+        return ErrorMessage(
+            absl::StrFormat("Can not attach to tid \"%d\": \"%s\"", tid, strerror(errno)));
```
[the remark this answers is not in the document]
danielfenner: as above

src/UserSpaceInstrumentation/Attach.cpp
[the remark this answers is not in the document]
danielfenner: done

src/UserSpaceInstrumentation/Attach.cpp
```diff
@@ -0,0 +1,105 @@
...
+      } else {
+        return false;
+      }
+    }
+    if (result_waitpid > 0) {
+      // waitpid returns greater zero occationally for terminated threads. We catch this here so we
+      // don't count them as succefully halted.
```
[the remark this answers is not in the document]
danielfenner: done

src/UserSpaceInstrumentation/Attach.cpp
danielfenner: Stopping hundreds of active threads is not a problem with the existing code. So I think it is fine to avoid the busy wait.

src/UserSpaceInstrumentation/Attach.cpp
danielfenner: Theoretically, yes. If the process spawns new threads faster than we can stop them (will never happen). And there are other hard boundaries. E.g.

$ cat /proc/sys/kernel/threads-max
255605

src/UserSpaceInstrumentation/AttachTest.cpp
danielfenner: adjusted comment

src/UserSpaceInstrumentation/Attach.cpp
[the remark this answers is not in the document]
danielfenner: This only refers to the case where you need a variable of a different size. In that case I agree we shpould use *_t.

For general purpose, sane, everyday numbers the style guide suggests:
"We use int very often, for integers we know are not going to be too big, e.g., loop counters. Use plain old int for such things."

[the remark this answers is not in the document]
dpallotti: You are right, your preference then.

PR #1733, opened 2021-01-26
Add JoinFutures and Future::Then helpers

beckerhe:
This PR is foremost adding two additional features to the Promise/Future framework.

1. `orbit_base::JoinFutures`: This free function takes a span of Futures and creates a joined future that completes after all the "child"-futures have been completed. Currently it's a non-templated implementation for `orbit_base::Future<void>`. This can be extended for non-void futures in the future. (no pun intended.)
2. `orbit_base::Future<T>::Then`: This is a member function that allows to chain tasks, even when executed on different executors. The member function takes a pointer to an executor and a function-object as parameters. The function object will be executed on the given executor when the future finishes.

### A typical example for the 1st
(from the upcoming symbol loading refactoring)
```
orbit_base::Future<void> OrbitApp::LoadModules(absl::Span<const ModuleData*> modules) {
  std::vector<orbit_base::Future<void>> futures;
  futures.reserve(modules.size());

  for(const auto& module : modules) {
    futures.emplace_back(LoadModule(module));
  }

  return orbit_base::JoinFutures(futures);  
}
```
All executions of `LoadModule` happen in parallel. `LoadModules` has been simplified by separating concerns.
`LoadModule` can only load one module at a time, while `LoadModules` only contains the logic for handling multiple
`LoadModule` calls.

### A typical example for the 2nd
```
auto future = thread_pool_->Schedule([]() {
  // do something expensive
 return Result{};
});
// or
auto future = MyAsyncOperation(); // Think `LoadModule` as from the previous example

future.Then(main_thread_executor_.get(), [](Result result) {
  // process `result` and update the UI
});
```

Previously the call of `main_thread_executor->Schedule` was part of the background job,
which ties the background task to the processing on the main thread. By using `Then` the
background task gets independent of the continuation. Once more this leads to better separation
of concerns.

### DIFF
--- a/src/OrbitBase/JoinFutures.cpp
+++ b/src/OrbitBase/JoinFutures.cpp
@@ -1,0 +1,52 @@
+// Copyright (c) 2021 The Orbit Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#include "OrbitBase/JoinFutures.h"
+
+#include <memory>
+
+#include "OrbitBase/Future.h"
+#include "OrbitBase/Logging.h"
+#include "OrbitBase/Promise.h"
+#include "absl/synchronization/mutex.h"
+
+namespace orbit_base {
+namespace {
+struct SharedStateJoin {
+  Promise<void> promise;
+  absl::Mutex mutex;
+  int incompleted_futures;
+};
+}  // namespace
+
+Future<void> JoinFutures(absl::Span<const Future<void>> futures) {
+  if (futures.empty()) {
+    Promise<void> promise;
+    promise.MarkFinished();
+    return promise.GetFuture();
+  }
+
+  auto shared_state = std::make_shared<SharedStateJoin>();
+  absl::MutexLock lock{&shared_state->mutex};
+
+  for (const auto& future : futures) {
+    CHECK(future.IsValid());
+    orbit_base::FutureRegisterContinuationResult const result =
+        future.RegisterContinuation([shared_state]() {
+          absl::MutexLock lock{&shared_state->mutex};
+          --shared_state->incompleted_futures;
+
+          if (shared_state->incompleted_futures == 0) {
+            shared_state->promise.MarkFinished();
+          }
+        });
+
+    if (result == orbit_base::FutureRegisterContinuationResult::kSuccessfullyRegistered) {
+      ++shared_state->incompleted_futures;
+    }
+  }
+
+  return shared_state->promise.GetFuture();
+}
+}  // namespace orbit_base
--- a/src/OrbitBase/JoinFuturesTest.cpp
+++ b/src/OrbitBase/JoinFuturesTest.cpp
@@ -1,0 +1,68 @@
+// Copyright (c) 2021 The Orbit Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#include <gtest/gtest.h>
+
+#include "OrbitBase/Future.h"
+#include "OrbitBase/JoinFutures.h"
+#include "OrbitBase/Promise.h"
+#include "absl/types/span.h"
+
+namespace orbit_base {
+
+TEST(JoinFutures, JoinEmptySpan) {
+  Future<void> joined_future = JoinFutures({});
+  EXPECT_TRUE(joined_future.IsValid());
+  EXPECT_TRUE(joined_future.IsFinished());
+}
+
+TEST(JoinFutures, JoinSpanWithOneElement) {
+  Promise<void> promise{};
+  Future<void> future = promise.GetFuture();
+
+  Future<void> joined_future = JoinFutures({future});
+  EXPECT_TRUE(joined_future.IsValid());
+  EXPECT_FALSE(joined_future.IsFinished());
+
+  promise.MarkFinished();
+  EXPECT_TRUE(joined_future.IsFinished());
+}
+
+TEST(JoinFutures, JoinSpanWithManyElements) {
+  Promise<void> promise0{};
+  Future<void> future0 = promise0.GetFuture();
+
+  Promise<void> promise1{};
+  Future<void> future1 = promise1.GetFuture();
+
+  Promise<void> promise2{};
+  Future<void> future2 = promise2.GetFuture();
+
+  Future<void> joined_future = JoinFutures({future0, future1, future2});
+  EXPECT_TRUE(joined_future.IsValid());
+  EXPECT_FALSE(joined_future.IsFinished());
+
+  promise0.MarkFinished();
+  EXPECT_FALSE(joined_future.IsFinished());
+
+  promise2.MarkFinished();
+  EXPECT_FALSE(joined_future.IsFinished());
+
+  promise1.MarkFinished();
+  EXPECT_TRUE(joined_future.IsFinished());
+}
+
+TEST(JoinFutures, JoinSpanWithDuplicateElements) {
+  Promise<void> promise{};
+  Future<void> future = promise.GetFuture();
+
+  Future<void> joined_future = JoinFutures({future, future});
+  EXPECT_TRUE(joined_future.IsValid());
+  EXPECT_FALSE(joined_future.IsFinished());
+
+  promise.MarkFinished();
+  EXPECT_TRUE(joined_future.IsFinished());
+}
+
+}  // namespace orbit_base
--- a/src/OrbitBase/include/OrbitBase/Future.h
+++ b/src/OrbitBase/include/OrbitBase/Future.h
@@ -121,6 +121,16 @@ class [[nodiscard]] Future : public orbit_base_internal::FutureBase<T> {
         &this->shared_state_->result));
   }
 
+  // This is syntactic sugar for MainThreadExecutor (or maybe other executors in the future).
+  // `invocable` will be executed by `executor` after this future has completed.
+  //
+  // Note: Usually `invocable` won't be executed if `executor` gets destroyed before `*this`
+  // completes. Check the docs or implementation of `Executor::ScheduleAfter` to be sure.
+  template <typename Executor, typename Invocable>
+  auto Then(Executor * executor, Invocable && invocable) const {
+    return executor->ScheduleAfter(*this, std::forward<Invocable>(invocable));
+  }
+
  private:
   friend orbit_base_internal::PromiseBase<T>;
 
@@ -167,6 +177,16 @@ class Future<void> : public orbit_base_internal::FutureBase<void> {
     absl::MutexLock lock{&this->shared_state_->mutex};
     shared_state_->mutex.Await(absl::Condition(
         +[](bool* finished) { return *finished; }, &shared_state_->finished));
+  }
+
+  // This is syntactic sugar for MainThreadExecutor (or maybe other executors in the future).
+  // `invocable` will be executed by `executor` after this future has completed.
+  //
+  // Note: Usually `invocable` won't be executed if `executor` gets destroyed before `*this`
+  // completes. Check the docs or implementation of `Executor::ScheduleAfter` to be sure.
+  template <typename Executor, typename Invocable>
+  auto Then(Executor* executor, Invocable&& invocable) const {
+    return executor->ScheduleAfter(*this, std::forward<Invocable>(invocable));
   }
 
  private:
--- a/src/OrbitBase/include/OrbitBase/Future.h
+++ b/src/OrbitBase/include/OrbitBase/Future.h
@@ -47,6 +47,12 @@ class FutureBase {
 
 namespace orbit_base {
 
+enum class FutureRegisterContinuationResult {
+  kSuccessfullyRegistered,
+  kFutureAlreadyCompleted,
+  kFutureNotValid
+};
+
 // A future type, similar to std::future but prepared to integrate with
 // Orbit-specific code like MainThreadExecutor and ThreadPool.
 //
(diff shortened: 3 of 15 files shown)


src/OrbitBase/include/OrbitBase/Future.h
```diff
@@ -90,11 +96,16 @@ class [[nodiscard]] Future : public orbit_base_internal::FutureBase<T> {
   // The continuation is potentially executed on a background thread,
   // which means you have to be aware of race-conditions while registering
   // the continuation and potential mutex deadlocks in the continuation.
-  void RegisterContinuation(std::function<void(const T&)> continuation) const {
-    if (!this->IsValid()) return;
+  [[nodiscard]] FutureRegisterContinuationResult RegisterContinuation(
+      std::function<void(const T&)> continuation) const {
+    if (!this->IsValid()) return FutureRegisterContinuationResult::kFutureNotValid;
```
dimitry-, line 102: add a test for this?

src/OrbitBase/include/OrbitBase/Future.h
```diff
@@ -90,11 +96,16 @@ class [[nodiscard]] Future : public orbit_base_internal::FutureBase<T> {
...
 
     absl::MutexLock lock{&this->shared_state_->mutex};
+    if (this->shared_state_->result.has_value()) {
+      return FutureRegisterContinuationResult::kFutureAlreadyCompleted;
```
dimitry-, line 106: ... and this?

src/OrbitBase/JoinFutures.cpp
```diff
@@ -0,0 +1,50 @@
...
+#include "OrbitBase/Logging.h"
+#include "OrbitBase/Promise.h"
+#include "absl/synchronization/mutex.h"
+
+namespace orbit_base {
+namespace {
+struct SharedStateJoin {
+  Promise<void> promise;
+  absl::Mutex mutex;
+  int uncompleted_futures;
```
dimitry-, line 19: incomplete?

src/OrbitBase/JoinFutures.cpp
```diff
@@ -0,0 +1,50 @@
...
+            shared_state->promise.MarkFinished();
+          }
+        });
+
+    if (result == orbit_base::FutureRegisterContinuationResult::kSuccessfullyRegistered) {
+      ++shared_state->uncompleted_futures;
+    }
+  }
+
+  if (shared_state->uncompleted_futures == 0) {
```
dimitry-, line 44: maybe check features.empty() at line 24 and do early return?

src/OrbitBase/JoinFuturesTest.cpp
```diff
@@ -0,0 +1,56 @@
...
+
+  Future<void> joined_future = JoinFutures({future});
+  EXPECT_TRUE(joined_future.IsValid());
+  EXPECT_FALSE(joined_future.IsFinished());
+
+  promise.MarkFinished();
+  EXPECT_TRUE(joined_future.IsFinished());
+}
+
+TEST(JoinFutures, JoinSpanWithManyElemente) {
```
dimitry-, line 32: typo: s/Elemente/Elements/

src/OrbitQt/MainThreadExecutorImplTest.cpp
```diff
@@ -15,8 +16,8 @@ static int argc = 0;
 TEST(MainThreadExecutorImpl, Schedule) {
   QCoreApplication app{argc, nullptr};
 
-  MainThreadExecutorImpl executor{};
-  (void)executor.Schedule([]() { QCoreApplication::instance()->exit(42); });
+  auto executor = MainThreadExecutorImpl::Create();
+  (void)executor->Schedule([]() { QCoreApplication::exit(42); });
```
dimitry-, line 20: do you need explicit cast to void here? Doesn't it return Future<void>?

src/OrbitGl/MainThreadExecutor.h
```diff
@@ -65,6 +67,134 @@ class MainThreadExecutor {
...
+    orbit_base::Future<ReturnType> resulting_future = promise->GetFuture();
+
+    // Since MSVC2017 crashes with a compiler error when using `if constexpr` inside of lambda, we
+    // have to put the check on the outside. That's unfortunate since it leads to four almost
+    // identical version of the same code.
+    if constexpr (std::is_same_v<ReturnType, void>) {
+      auto continuation = [functor = std::forward<F>(functor), executor_weak_ptr = weak_from_this(),
+                           promise = std::move(promise)]() mutable {
+        auto executor = executor_weak_ptr.lock();
+        if (executor == nullptr) return;
```
dimitry-, line 88: This deserves some comment clarifying if the executer destruction can happen during orderly shutdown. Normally we would want to complete every scheduled action, and it could be argued that ScheduleAfter is also a form of scheduling.

src/OrbitQt/MainThreadExecutorImplTest.cpp
```diff
@@ -22,6 +22,93 @@ TEST(MainThreadExecutorImpl, Schedule) {
...
+TEST(MainThreadExecutorImpl, ScheduleAfterWithExecutorOutOfScope) {
+  QCoreApplication app{argc, nullptr};
+
+  orbit_base::Promise<void> promise;
+  orbit_base::Future<void> future = promise.GetFuture();
+
+  auto executor = MainThreadExecutorImpl::Create();
+
+  bool called = false;
+  auto future2 = executor->ScheduleAfter(future, [&called]() { called = true; });
```
dimitry-, line 101: ok, I see now. this is one case. But I still wonder if this has unintended consequences, if we do not guarantee execution of ScheduledAfter actions.

src/OrbitBase/include/OrbitBase/Future.h
beckerhe: Good point. Done

src/OrbitBase/include/OrbitBase/Future.h
beckerhe: Done as well. Thanks.

src/OrbitBase/JoinFutures.cpp
beckerhe: Yesssss! Fixed.

src/OrbitBase/JoinFutures.cpp
beckerhe: I debated this with myself and decided against it. But now I don't see why anymore. Maybe the code was different at that point.

Fixed. Thanks!

src/OrbitBase/JoinFuturesTest.cpp
beckerhe: Yep. Fixed.

src/OrbitQt/MainThreadExecutorImplTest.cpp
beckerhe: Ahhhh. Noooo. And yes. From the last code review I reverted the nodiscard in ThreadPool, but not in MainThreadExecutor. Which makes no sense, I guess.

I added a commit which reverts the nodiscard and removes all the unnecessary (void) casts.

src/OrbitGl/MainThreadExecutor.h
beckerhe: Good point. I added some comment which explains the function's behaviour. I wanted to do that anyway.

src/OrbitQt/MainThreadExecutorImplTest.cpp
beckerhe: This is very much intended behaviour. At some point I want to introduce the same behaviour for `Schedule`, but it's not that pressing there, since these actions are mostly executed right away. But let me quickly motivate why I think it is a good idea to tie the lifetime of actions to the lifetime of the executor:

We have a lot of Backgroundtasks that work like this:
```
void OrbitApp::Method() {
  thread_pool_->Schedule([this](){
    // do something expensive
    main_thread_executor_->Schedule([result]() {
      // call some member function of OrbitApp and process result
    });
  });
```

The thread pool action usually doesn't check whether `this` is still alive, nor if `main_thread_executor_` is still a valid pointer. 
The same applies to the main_thread_executor_ action. It also usually doesn't check that.

And we had already fixed crashes that occur when the application shuts down. There is also currently an open
bug: Orbit crashes sometimes when you close it while it's loading symbols.

There are two ways to solve that:
1. Make all actions check whether pointers are still valid. That's probably not feasible.
2. Tie the lifetime of actions to the objects they mutate. And that's what I'm trying here. The idea is to give `OrbitApp` (and also other objects if needed) their own instance of MainThreadExecutor. That means all actions running on this instance of MainThreadExecutor can assume that `OrbitApp` is still alive. And when `OrbitApp` gets destructed the local MainThreadExecutor dies first and no action will execute on dangling pointers.

dimitry- (approved):
lgtm, thanks!

src/OrbitQt/MainThreadExecutorImplTest.cpp
dimitry-: makes sense.

src/OrbitGl/MainThreadExecutor.h
```diff
@@ -65,6 +67,142 @@ class MainThreadExecutor {
...
+      -> orbit_base::Future<std::decay_t<decltype(functor())>> {
+    CHECK(future.IsValid());
+
+    using ReturnType = std::decay_t<decltype(functor())>;
+
+    // Since std::function needs to be copyable, promise also needs to be copyable.
+    auto promise = std::make_shared<orbit_base::Promise<ReturnType>>();
+    orbit_base::Future<ReturnType> resulting_future = promise->GetFuture();
+
+    // Since MSVC2017 crashes with a compiler error when using `if constexpr` inside of lambda, we
```
karupayun, line 84: I'm not into the level of details about the crashes in MSVC2017, but what do you think about creating an internal bug to try to merge some of them together in the future (or a TODO)? Maybe you already thought a lot here, in this case please disregard my comment.

src/OrbitBase/include/OrbitBase/Future.h
```diff
@@ -121,6 +121,16 @@ class [[nodiscard]] Future : public orbit_base_internal::FutureBase<T> {
         &this->shared_state_->result));
   }
 
+  // This is syntactic sugar for MainThreadExecutor (or maybe other executors in the future).
+  // `invocable` will be executed by `executor` after this future has completed.
+  //
+  // Note: Usually `invocable` won't be executed if `executor` gets destroyed before `*this`
+  // completes. Check the docs or implementation of `Executor::ScheduleAfter` to be sure.
+  template <typename Executor, typename Invocable>
+  auto Then(Executor * executor, Invocable && invocable) const {
```
karupayun, line 130: Nit: space here. Executor* executor

karupayun (approved):
LGTM Henning!! Only a few small comments :)

I'm not 100% sure why we need to separate between void and other types but I will ping you in private about that.

src/OrbitBase/JoinFuturesTest.cpp
```diff
@@ -0,0 +1,56 @@
...
+
+  Future<void> joined_future = JoinFutures({future});
+  EXPECT_TRUE(joined_future.IsValid());
+  EXPECT_FALSE(joined_future.IsFinished());
+
+  promise.MarkFinished();
+  EXPECT_TRUE(joined_future.IsFinished());
+}
+
+TEST(JoinFutures, JoinSpanWithManyElements) {
```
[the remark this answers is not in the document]
beckerhe: Good idea. I added a test for this.

src/OrbitQt/MainThreadExecutorImplTest.cpp
```diff
@@ -22,6 +22,93 @@ TEST(MainThreadExecutorImpl, Schedule) {
   EXPECT_EQ(app.exec(), 42);
 }
 
+TEST(MainThreadExecutorImpl, ScheduleAfterAllVoid) {
+  QCoreApplication app{argc, nullptr};
+
+  bool called = false;
+  auto executor = MainThreadExecutorImpl::Create();
+  auto future = executor->Schedule([&called]() { called = true; });
```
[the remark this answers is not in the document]
beckerhe: It's not about being obvious. It's rather that I don't want to (and I shouldn't) test the behaviour of `MainThreadExecutor::Schedule`. This test is about `ScheduleAfter` and should ideally depend on
no the behaviour of `ScheduleAfter`. Of course that's hard to do and in most cases impossible.

But the general idea is: Test cases should be disjoint as much as possible. When I make a behavioural change (or introduce a bug), ideally just a single test case fails. Then I immediately know where the bug occurs. If all test cases
fail, it's hard to find the bug.

src/OrbitQt/MainThreadExecutorImplTest.cpp
```diff
@@ -22,6 +22,93 @@ TEST(MainThreadExecutorImpl, Schedule) {
...
+  QCoreApplication app{argc, nullptr};
+
+  auto executor = MainThreadExecutorImpl::Create();
+  auto future = executor->Schedule([]() { return 42; });
+  auto future2 = executor->ScheduleAfter(future, [](int val) { return val + 42; });
+  auto future3 = executor->ScheduleAfter(future2, [](int val) { return val + 42; });
+  auto future4 =
+      executor->ScheduleAfter(future3, [](int val) { QCoreApplication::exit(val + 42); });
+
+  EXPECT_EQ(app.exec(), 4 * 42);
```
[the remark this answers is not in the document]
beckerhe: Good idea. But I opted for a simpler solution by putting checks into the lambdas.

src/OrbitGl/MainThreadExecutor.h
beckerhe: I already have a fix lined up which reduces `ScheduleAfter` to a single function with no `if constexpr`.

PR #1736, opened 2021-01-26
Add test CallstackSamplesAndAddressInfos to LinuxTracingIntegrationTest

dpallotti:
Bug: http://b/160756664

### DIFF
--- a/src/GrpcProtos/capture.proto
+++ b/src/GrpcProtos/capture.proto
@@ -97,7 +97,7 @@ message InternedCallstack {
 }
 
 message CallstackSample {
-  int32 pid = 1;
+  int32 pid = 1;  // Note: pid is currently not set, the client doesn't need it.
   int32 tid = 2;
   oneof callstack_or_key {
     Callstack callstack = 3;
--- a/src/LinuxTracing/LinuxTracingIntegrationTest.cpp
+++ b/src/LinuxTracing/LinuxTracingIntegrationTest.cpp
@@ -2,6 +2,7 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
 
+#include <absl/container/flat_hash_set.h>
 #include <absl/strings/numbers.h>
 #include <absl/time/clock.h>
 #include <gmock/gmock.h>
@@ -14,6 +15,7 @@
 #include <thread>
 
 #include "ElfUtils/ElfFile.h"
+#include "ElfUtils/LinuxMap.h"
 #include "LinuxTracing/Tracer.h"
 #include "LinuxTracing/TracerListener.h"
 #include "LinuxTracingIntegrationTestPuppet.h"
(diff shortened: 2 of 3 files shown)


src/GrpcProtos/capture.proto
```diff
@@ -97,7 +97,7 @@ message InternedCallstack {
 }
 
 message CallstackSample {
-  int32 pid = 1;
+  int32 pid = 1;  // Note that pid is currently not set as the client doesn't need it.
```
dimitry-, line 100: maybe we should fix this...

dimitry-: or if it is really by design, remove this field?

src/LinuxTracing/LinuxTracingIntegrationTest.cpp
```diff
@@ -282,6 +284,44 @@ using PuppetConstants = LinuxTracingIntegrationTestPuppetConstants;
   return fixture->StopTracingAndGetEvents();
 }
 
+std::filesystem::path GetExecutableBinaryPath(pid_t pid) {
```
dimitry-, line 287: GetExecutablePathOrDie?

src/LinuxTracing/LinuxTracingIntegrationTest.cpp
```diff
@@ -420,6 +452,145 @@ TEST(LinuxTracingIntegrationTest, FunctionCalls) {
...
+    if (symbol.name() == PuppetConstants::kInnerFunctionName) {
+      CHECK(inner_function_virtual_address_start == 0 && inner_function_virtual_address_end == 0);
+      inner_function_virtual_address_start = module_info.address_start() + symbol.address();
+      inner_function_virtual_address_end = inner_function_virtual_address_start + symbol.size() - 1;
+    }
+  }
+  CHECK(outer_function_virtual_address_start != 0);
+  CHECK(outer_function_virtual_address_end != 0);
+  LOG("%s virtual addresses: %#lx-%#lx", PuppetConstants::kOuterFunctionName,
+      outer_function_virtual_address_start, outer_function_virtual_address_end);
```
dimitry-, line 487: Is this leftover from debugging or do we want to keep this?

src/GrpcProtos/capture.proto
dimitry-: if it is easy to set we should probably set it and just have it here ready for use if needed.

dpallotti: It then just gets dropped by CaptureEventProcessor as `orbit_client_protos::CallstackEvent` doesn't have a process id. Do we want to do it anyway? Maybe. Do we want to add a process id to `CallstackEvent`? Maybe. In the context of this change I don't know and it's out of scope.

dimitry-: you have discovered this while writing the test, so I would say it is in the scope. I would say we should set it here. If then client ignores it it is fine. This proto will become persistent soon which means that it could become used in the latter version of orbit.

dpallotti: > you have discovered this while writing the test, so I would say it is in the scope.

Why? It was intended behavior.

> This proto will become persistent soon which means that it could become used in the latter version of orbit.

That's why this is out of scope. We should decide what fields we want set and what we don't separately to prepare for a stable version. And that includes the other protos that have the same property (`ThreadName`, `ThreadStateSlice`). I'm not saying we shouldn't set process ids, I'm saying it's not a decision for this change.

florian-kuebler (approved):
LGTM. I guess when we add the frame pointer test here, we need some refactoring, but this is for later.
Thanks!

PR #1737, opened 2021-01-26
Syntax highlighting for C

karupayun:
This PR creates the Syntax highlighter for source code (C++).
For now, the syntax highlighter is only using C-rules. It will be
expanded with C++ rule in a future PR.

It can be tested using CodeViewerDemo. There you will see a few fake unit-tests only to make regex more understandable and easier to modify in the future. 
fake unit tests

Also there is a copy of CaptureServiceImpl.cpp as an
example to test how the Syntax Highlighter works with a real file.
real example

Related to http://b/177651839.

### DIFF
--- a/src/CodeViewerDemo/CodeExamples.h
+++ b/src/CodeViewerDemo/CodeExamples.h
@@ -1,0 +1,69 @@
+// Copyright (c) 2021 The Orbit Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#include <QString>
+
+inline const QString cpp_example = R"(#pragma twice
+// Copyright (c) 2021 The Orbit Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#include <iostream>
+
+#include "TrackManager.h"
+
+class TrackManager {
+ public:
+  explicit TrackManager(TimeGraph* time_graph, OrbitApp* app);
+  void Clear();
+
+  [[nodiscard]] std::vector<Track*> GetAllTracks() const;
+
+ protected:
+  /* This is a protected area. "This is not a string" */
+ private: /* Nobody
+           * can
+           * access
+           * this
+           * part */
+  std::vector<Track*> tracks_;
+}
+
+int main (int argc, char **argv) {
+  char start_capture = 'x';
+  std::string multiline =
+      " This \
+                            is a \
+                       multiline \
+                          string";
+  int* my_ptr = NULL;
+  printf("Hi Orbit's user, if you want to start a capture please press \"%c\"\n", start_capture);
+  return 0;
+
+})";
+
+inline const QString x86Assembly_example = R"(Platform: X86 64 (Intel syntax)
+0x557e1f8091d0:	push         rbp
+0x557e1f8091d1:	mov          rbp, rsp
+0x557e1f8091d4:	sub          rsp, 0x10
+0x557e1f8091d8:	mov          rax, qword ptr fs:[0x28]
+0x557e1f8091e1:	mov          qword ptr [rbp - 8], rax
+0x557e1f8091e5:	xorps        xmm0, xmm0
+0x557e1f8091e8:	movups       xmmword ptr [rdi], xmm0
+0x557e1f8091eb:	mov          qword ptr [rdi + 0x10], 0
+0x557e1f8091f3:	mov          byte ptr [rdi], 0x2a
+0x557e1f8091f6:	movabs       rax, 0x65646f6d20726570
+0x557e1f809200:	mov          qword ptr [rdi + 0xe], rax
+0x557e1f809204:	movups       xmm0, xmmword ptr [rip - 0x36e2d5]
+0x557e1f80920b:	movups       xmmword ptr [rdi + 1], xmm0
+0x557e1f80920f:	mov          byte ptr [rdi + 0x16], 0
+0x557e1f809213:	mov          rax, qword ptr fs:[0x28]
+0x557e1f80921c:	cmp          rax, qword ptr [rbp - 8]
+0x557e1f809220:	jne          0x557e1f80922b
+0x557e1f809222:	mov          rax, rdi
+0x557e1f809225:	add          rsp, 0x10
+0x557e1f809229:	pop          rbp
+0x557e1f80922a:	ret
+0x557e1f80922b:	call         0x557e2000aa70
+0x557e1f809230:)";
--- a/src/CodeViewerDemo/CodeExamples.h
+++ b/src/CodeViewerDemo/CodeExamples.h
@@ -4,6 +4,299 @@
 
 #include <QString>
 
+inline const QString testing_example =
+    R"(
+// Tests to understand how the syntax highlighter works
+  // Number unit tests
+  a = 0x8542;
+  a = 4103;
+  a = 0;
+  a = 01ab;
+  a = a105;
+  a = _103;
+
+  // Capitalized words unit tests
+  OrbitService o;
+  QSyntaxHighlighter q;
+  ORBIT_SCOPE; // no lowercase regex
+  x;
+  kTreeSize == 3;
+  _Tree_;
+
+  // Function unit tests
+  int my_function(void)
+  bool b = a.insert(3);
+  void AddEvent();
+  char _my_underscore_func();
+  int CaptureServiceImpl::Capture() // Function from other class
+  void QSyntaxHighlighter::Play
+  CHECK(false); // no lowercase regex
+
+  // Preprocessor + <> unit tests
+  #include <iostream>
+  std::vector<Typeee>
+  #include<iostream>
+   #pragma<ios>
+  include<iostream>
+
+  // Strings unit tests
+  "Hola mundo\n"
+  "I'm an \"only\" string"
+  "I'm 2" + "strings"
+  "CHECK(false)\\\\"
+  "Multi\
+  Line"
+
+  // Comments unit tests
+  // return
+  // This is a comment
+  return; // This isn't a comment
+  /*This
+  is a
+  long comment*/
+  /* One line comment */ int main;
+
+// CaptureServiceImpl.cpp
+// Copyright (c) 2020 The Orbit Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#include <absl/container/flat_hash_set.h>
+#include <absl/synchronization/mutex.h>
+#include <absl/time/time.h>
+#include <pthread.h>
+#include <stdint.h>
+
+#include <algorithm>
+#include <limits>
+#include <thread>
+#include <utility>
+#include <vector>
+
+#include "CaptureEventBuffer.h"
+#include "CaptureEventSender.h"
+#include "CaptureServiceImpl.h"
+#include "LinuxTracingHandler.h"
+#include "OrbitBase/Logging.h"
+#include "OrbitBase/Tracing.h"
+#include "capture.pb.h"
+
+namespace orbit_service {
+
+using orbit_grpc_protos::CaptureRequest;
+using orbit_grpc_protos::CaptureResponse;
+
+namespace {
+
+using orbit_grpc_protos::CaptureEvent;
+
+class SenderThreadCaptureEventBuffer final : public CaptureEventBuffer {
+ public:
+  explicit SenderThreadCaptureEventBuffer(CaptureEventSender* event_sender)
+      : capture_event_sender_{event_sender} {
+    CHECK(capture_event_sender_ != nullptr);
+    sender_thread_ = std::thread{[this] { SenderThread(); }};
+  }
+
+  void AddEvent(orbit_grpc_protos::CaptureEvent&& event) override {
+    absl::MutexLock lock{&event_buffer_mutex_};
+    if (stop_requested_) {
+      return;
+    }
+    event_buffer_.emplace_back(std::move(event));
+  }
+
+  void StopAndWait() {
+    CHECK(sender_thread_.joinable());
+    {
+      // Protect stop_requested_ with event_buffer_mutex_ so that we can use stop_requested_
+      // in Conditions for Await/LockWhen (specifically, in SenderThread).
+      absl::MutexLock lock{&event_buffer_mutex_};
+      stop_requested_ = true;
+    }
+    sender_thread_.join();
+  }
+
+  ~SenderThreadCaptureEventBuffer() override { CHECK(!sender_thread_.joinable()); }
+
+ private:
+  void SenderThread() {
+    pthread_setname_np(pthread_self(), "SenderThread");
+    constexpr absl::Duration kSendTimeInterval = absl::Milliseconds(20);
+    // This should be lower than kMaxEventsPerResponse in GrpcCaptureEventSender::SendEvents
+    // as a few more events are likely to arrive after the condition becomes true.
+    constexpr uint64_t kSendEventCountInterval = 5000;
+
+    bool stopped = false;
+    while (!stopped) {
+      ORBIT_SCOPE("SenderThread iteration");
+      event_buffer_mutex_.LockWhenWithTimeout(absl::Condition(
+                                                  +[](SenderThreadCaptureEventBuffer* self) {
+                                                    return self->event_buffer_.size() >=
+                                                               kSendEventCountInterval ||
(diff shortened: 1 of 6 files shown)


src/CodeViewerDemo/CodeExamples.h
```diff
@@ -0,0 +1,66 @@
+// Copyright (c) 2021 The Orbit Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#include <QApplication>
```
beckerhe, line 5: QString?

src/CodeViewerDemo/CodeExamples.h
```diff
@@ -0,0 +1,66 @@
...
+
+const QString cpp_example = R"(#pragma twice
```
beckerhe, line 7: If you define these variables in a header you have to declare them `inline`. Otherwise you will get linker error when including them from different cpp-files.

src/SyntaxHighlighter/Cpp.cc
```diff
@@ -0,0 +1,103 @@
+// Copyright (c) 2021 The Orbit Authors. All rights reserved.
```
beckerhe, line 1: We usually use the file extension `.cpp`. Can you rename the file?

src/SyntaxHighlighter/Cpp.cc
```diff
@@ -0,0 +1,103 @@
...
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#include "SyntaxHighlighter/Cpp.h"
+
+#include <qregularexpression.h>
```
beckerhe, line 7: `#include <QRegularExpression>`

The same in the header file.

src/SyntaxHighlighter/include/SyntaxHighlighter/Cpp.h
```diff
@@ -0,0 +1,48 @@
...
+
+#include <absl/container/flat_hash_map.h>
+#include <qregularexpression.h>
+
+#include <QSyntaxHighlighter>
+
+namespace orbit_syntax_highlighter {
+
+/*
+  This a syntax highlighter for cpp.
```
beckerhe, line 16: Very, very nit: cpp -> C++

src/SyntaxHighlighter/include/SyntaxHighlighter/Cpp.h
```diff
@@ -0,0 +1,48 @@
...
+
+  It derives from QSyntaxHighlighter, so check out QSyntaxHighlighter's
+  documentation on how to use it. There are no additional settings or
+  APIs.
+*/
+class Cpp : public QSyntaxHighlighter {
+  Q_OBJECT
+  enum CppHighlighterState { kInitialState = 0, kOpenCommentState = 1, kOpenStringState = 2 };
```
beckerhe, line 24: Do you need the explicit values? If not I would remove the `= 0`, `= 1`, and `= 2`.

beckerhe: And it seems you don't need this enum for the API. So I would move it to the .cpp-file into a anonymous namespace.

src/SyntaxHighlighter/include/SyntaxHighlighter/Cpp.h
```diff
@@ -0,0 +1,48 @@
...
+
+namespace orbit_syntax_highlighter {
+
+/*
+  This a syntax highlighter for cpp.
+
+  It derives from QSyntaxHighlighter, so check out QSyntaxHighlighter's
+  documentation on how to use it. There are no additional settings or
+  APIs.
+*/
```
beckerhe, line 21: One more thing: Dimitry recently pointed out, that we usually use `//` also for block comments. I did this wrong in the assembly syntax highlighter. But we should probably follow that guidance.

src/SyntaxHighlighter/Cpp.cc
```diff
@@ -0,0 +1,103 @@
...
+
+#include <QColor>
+#include <QString>
+#include <QTextCharFormat>
+
+namespace orbit_syntax_highlighter {
+namespace C {
```
beckerhe, line 14: It's fine to put these constants into a namespace with a name. But you still need the anonymous namespace as well.

src/SyntaxHighlighter/CMakeLists.txt
```diff
@@ -5,7 +5,7 @@
 cmake_minimum_required(VERSION 3.15)
 
 project(SyntaxHighlighter)
-add_library(SyntaxHighlighter STATIC)
+add_library(SyntaxHighlighter STATIC include/SyntaxHighlighter/Cpp.h Cpp.cc)
```
beckerhe, line 8: Could you add these files to the `target_sources` directives below instead?

src/CodeViewerDemo/main.cpp
```diff
@@ -4,41 +4,19 @@
...
-0x557e1f809220:	jne          0x557e1f80922b
-0x557e1f809222:	mov          rax, rdi
-0x557e1f809225:	add          rsp, 0x10
-0x557e1f809229:	pop          rbp
-0x557e1f80922a:	ret          
-0x557e1f80922b:	call         0x557e2000aa70
-0x557e1f809230:)";
-
-  dialog.SetSourceCode(content, std::make_unique<orbit_syntax_highlighter::X86Assembly>());
+  const QString content = cpp_example;
```
beckerhe, line 17: Nit: I would remove this variable and directly call SetSourceCode and pass `cpp_example` in.

src/SyntaxHighlighter/include/SyntaxHighlighter/Cpp.h
beckerhe: And could you make it an `enum class`. From the code I've seen, you already scope all usages properly.

src/SyntaxHighlighter/include/SyntaxHighlighter/Cpp.h
```diff
@@ -0,0 +1,48 @@
...
+  QRegularExpression number_regex_;
+  QRegularExpression constant_regex_;
+  QRegularExpression keyword_regex_;
+  QRegularExpression macro_regex_;
+  QRegularExpression include_file_regex_;
+  QRegularExpression string_regex_;
+  QRegularExpression open_string_regex_;
+  QRegularExpression end_string_regex_;
+  QRegularExpression no_end_string_regex_;
+  QRegularExpression uppercase_regex_;
```
beckerhe, line 43: These are public and they shouldn't. (That's also wrong in the Assembly highlighter. Apologies. :see_no_evil: )

src/SyntaxHighlighter/include/SyntaxHighlighter/Cpp.h
```diff
@@ -0,0 +1,48 @@
+// Copyright (c) 2021 The Orbit Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#ifndef SYNTAX_HIGHLIGHTER_CPP_H_
+#define SYNTAX_HIGHLIGHTER_CPP_H_
+
+#include <absl/container/flat_hash_map.h>
```
beckerhe, line 8: This header is not needed here. (Once more the same problem is in the assembly highlighter. I really should review my own code. :see_no_evil:  My bad. )

beckerhe (commented):
Thank you Pablo. Looks already really good. Much more readable than uncolored code.

I added some minor comments. The overall direction looks good. No complaints there. Just a few C++-y things.

src/SyntaxHighlighter/Cpp.cc
```diff
@@ -0,0 +1,103 @@
...
+    format.setForeground(color);
+
+    for (auto it = expression->globalMatch(code); it.hasNext();) {
+      const auto match = it.next();
+      setFormat(match.capturedStart(), match.capturedLength(), format);
+      setCurrentBlockState(new_state);
+    }
+  };
+
+  apply(&number_regex_, C::kBlue);
```
florian-kuebler, line 82: I guess long term, we want to have central place for all colors. For now, I think it is fine as it is.

florian-kuebler (approved):
That's pretty cool. Thanks. I share some of the comments from Henning. So see my review as LGTM as soon as he approved!

src/SyntaxHighlighter/include/SyntaxHighlighter/Cpp.h
karupayun: Do you want me to also change these things in X86Assembly?

beckerhe: No. This was not my intention at all. Of course I won't stop you if you tackle this. But I already made a mental note to address this soonish.

src/SyntaxHighlighter/Cpp.cc
karupayun: Thanks! Do you like it now?

src/SyntaxHighlighter/include/SyntaxHighlighter/Cpp.h
karupayun: Done to the first 2. I didn't make an enum class just because I have to convert to an int for setCurrentBlockState. Is there a nice way to do that?

src/CodeViewerDemo/CodeExamples.h
karupayun: Yes, thanks!

src/SyntaxHighlighter/Cpp.cc
beckerhe: Yes. Looks good!

src/CodeViewerDemo/main.cpp
karupayun: I supposed that was better to have that variable because we are using it twice

src/SyntaxHighlighter/Cpp.cc
karupayun: Yes, completely agree. Is also one of the things in my roadmap that I want to tackle.

src/CodeViewerDemo/main.cpp
beckerhe: Ah sorry. Yes. Sry...

src/SyntaxHighlighter/Cpp.cpp
```diff
@@ -0,0 +1,107 @@
...
+      const auto match = it.next();
+      setFormat(match.capturedStart(), match.capturedLength(), format);
+      setCurrentBlockState(new_state);
+    }
+  };
+
+  apply(&number_regex_, C::kBlue);
+  apply(&uppercase_regex_, C::kViolet);
+  apply(&keyword_regex_, C::kOrange);
+  apply(&include_file_regex_, C::kGreen);
```
ronaldfw, line 89: Ok, I'm really bad with reading regular expressions, so I'm sorry if I'm getting this wrong. 

It looks to me that at least the "include file" and "string" and "macro" regex work together in interesting ways here and that the order matters. 

"Include file regex" will turn green a string like '#include<header.h>'. "String regex" will turn green a string like '"this"' including '"header.h"'. And finally macro regex turns '#pragma', but also '#include', into yellow orange.

So, 

#include <file.h>
#include "header.h"

turn out the same, but due to different rules. 

Doing the "include file regex" after "macro regex" would also lead to a different result (example: #include with <> and "" would look different), right? That may be fine and intended, but I would add a comment to explain that.

src/SyntaxHighlighter/include/SyntaxHighlighter/Cpp.h
beckerhe: No, not really. You have to call `static_cast` all the time.

beckerhe (approved):
LGTM! Thanks Pablo!

src/SyntaxHighlighter/Cpp.cpp
```diff
@@ -0,0 +1,107 @@
...
+    "Static_assert|_Thread_local|asm|typeof|inline|auto|break|case|char|const|continue|default|do|"
+    "double|else|enum|extern|float|for|goto|if|int|long|register|return|short|signed|sizeof|static|"
+    "struct|switch|typedef|union|unsigned|void|volatile|while|)\\b";
+const char* const kCommentRegex =
+    "\\/\\/(?:[^\\r\\n\\\\]|\\\\(?:\\r\\n?|\\n|(?![\\r\\n])))*|\\/\\*[\\s\\S]*?\\*\\/";
+const char* const kOpenCommentRegex = "\\/\\*([^\\*]|[\\*]+[^\\/])*?$";  // Don't work with **/
+const char* const kEndCommentRegex = "[\\s\\S]*\\*\\/";
+const char* const kFunctionRegex = "[a-z_]\\w*(?=\\s*\\()";
+const char* const kNoEndCommentRegex = "([^\\*]|\\*+[^\\/\\*])*$";
+const char* const kMacroRegex = "(^\\s*)#\\s*[a-z_]\\S*";
```
ronaldfw, line 43: Should this better be called kPreprocessorRegex as you are detecting preprocessor directives and not macros (which may be indistinguishable from functions in general)?

src/SyntaxHighlighter/Cpp.cpp
karupayun: There are two different things
1- Yes, I tried to follow the same rules than X86Assembly where we process line by line and apply the rules one on top of another. So, the order matters. If we have **"Hi world, I'm an int variable"**, first **"int"** is painted and after this is overpainted by the comment thing". This leads to one problem, what comes first? Comment or String?
_" /* Hi, I'm an string */. ""_
_/*  "Hi, I'm a comment" */_
Is impossible to solve with the approach, but as they are really rare, we are continuing with our approach. If not, a more complex process is needed. But if you feel that it is important, I can change it.
2- The fact with macro and string was intended. But now, thinking about it better, I'm realizing that I can avoid that without much effort. Will change it.

ronaldfw: As far as I can tell, the approach is fine for our purposes. I would suggest to document it a bit, at least a comment in the code that the order matters and shouldn't be changed unless one knows what they are doing.

src/SyntaxHighlighter/Cpp.cpp
```diff
@@ -0,0 +1,107 @@
...
+#include "SyntaxHighlighter/Cpp.h"
+
+#include <QColor>
+#include <QRegularExpression>
+#include <QString>
+#include <QTextCharFormat>
+
+namespace orbit_syntax_highlighter {
+namespace C {
+namespace {
```
antonrohr, line 14: There might be a reason that I don't see, but why this extra namespace here?

antonrohr (commented):
Only one minor comment in the code and here comes another big-ish one. Feel free to not do it, because it is a bit of work and other people already gave their LGTM, but: How about unit tests? I feel one test for every regex might bring a lot of benefit. It protects us against future accidental changes and it documents what each regex does for the rest of us.

src/SyntaxHighlighter/Cpp.cpp
karupayun: I wanted to have the C-namespace because after I want to include the C++ rules and Henning asked for the anonymous also. @beckerhe ??

src/SyntaxHighlighter/Cpp.cpp
karupayun: Added a few comments in the regex and also when we're applying them. My goal is include at the first the "heuristic ones", where we can modify them following C++/Google conventions and to make the code feels a little bit nicer.

karupayun:
> Only one minor comment in the code and here comes another big-ish one. Feel free to not do it, because it is a bit of work and other people already gave their LGTM, but: How about unit tests? I feel one test for every regex might bring a lot of benefit. It protects us against future accidental changes and it documents what each regex does for the rest of us.

I liked that suggestion, will think about that :)

src/SyntaxHighlighter/Cpp.cpp
beckerhe: `namespace C` for the name. And the unnamed `namespace` to get internal linkage.

beckerhe: There is no way to shorten that. At least none that I know of.

antonrohr: Got it, thank you

karupayun:
So, there are a few more things here:
1- Update PR descriptions with a few screenshots
2- CodeViewerDemo shows "fake unit tests" for interesting regex (and also a real file after that)
3- Comments in every regex

PR #1740, opened 2021-01-27
Set pid for CallstackSample/ThreadName and ThreadStateSlice

dimitry-:
This field was not previously set but since this costs us
nothing lets set them.

Test: run unit and integration tests.

### DIFF
--- a/src/LinuxTracing/ThreadStateManager.h
+++ b/src/LinuxTracing/ThreadStateManager.h
@@ -50,13 +50,14 @@ class ThreadStateManager {
                       orbit_grpc_protos::ThreadStateSlice::ThreadState state);
   void OnNewTask(uint64_t timestamp_ns, pid_t tid);
   [[nodiscard]] std::optional<orbit_grpc_protos::ThreadStateSlice> OnSchedWakeup(
-      uint64_t timestamp_ns, pid_t tid);
+      pid_t pid, uint64_t timestamp_ns, pid_t tid);
   [[nodiscard]] std::optional<orbit_grpc_protos::ThreadStateSlice> OnSchedSwitchIn(
-      uint64_t timestamp_ns, pid_t tid);
+      pid_t pid, uint64_t timestamp_ns, pid_t tid);
   [[nodiscard]] std::optional<orbit_grpc_protos::ThreadStateSlice> OnSchedSwitchOut(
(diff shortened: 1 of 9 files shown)


src/LinuxTracing/ThreadStateManager.h
```diff
@@ -50,13 +50,14 @@ class ThreadStateManager {
...
   [[nodiscard]] std::optional<orbit_grpc_protos::ThreadStateSlice> OnSchedSwitchIn(
-      uint64_t timestamp_ns, pid_t tid);
+      pid_t pid, uint64_t timestamp_ns, pid_t tid);
   [[nodiscard]] std::optional<orbit_grpc_protos::ThreadStateSlice> OnSchedSwitchOut(
-      uint64_t timestamp_ns, pid_t tid, orbit_grpc_protos::ThreadStateSlice::ThreadState new_state);
+      pid_t pid, uint64_t timestamp_ns, pid_t tid,
+      orbit_grpc_protos::ThreadStateSlice::ThreadState new_state);
   [[nodiscard]] std::vector<orbit_grpc_protos::ThreadStateSlice> OnCaptureFinished(
-      uint64_t timestamp_ns);
+      pid_t pid, uint64_t timestamp_ns);
```
dpallotti, line 60: In any case this is a hack. Nothing says that `ThreadStateManager` handles a single process id. For this `ThreadStateManager` should also do bookkeeping of pids internally. It's wrong to give one pid to assign to every slice.

dpallotti (commented):
> this costs us nothing

- I'm fine with this change but it's not for free. While with callstacks and thread names it's a drop in the bucket, thread state causes lot of events, and this can increase the amount of data sent of a non-negligible percentage. Before we archive it as with no costs, we should at least have an idea of whether there is actually an impact by comparing the amount of data sent that gets printed in the logs.

- As stated in the comment, I feel that the way the pid is retrieved for the `task/task_newtask` tracepoint is wrong.

- Especially for the two previous points, this should have been three separate commits. In general the logic around tracepoints and thread states is delicate and changes to it should be well scoped.

- For these changes run a manual end to end test on Trata and verify that everything looks in order: no crashes, samples, scheduling slices, thread names, thread states. Unit tests and a work-in-progress integration test is far from enough.

PR #1741, opened 2021-01-27
Wait for first SchedulingSlice in LinuxTracingIntegrationTest

dpallotti:
Before sending a command to the puppet, instead of sleeping for an arbitrary
amount of time.
This should guarantee that the tracing loop in `TracerThread` has started, hence
that the capture has been fully set up.

### DIFF
--- a/src/LinuxTracing/LinuxTracingIntegrationTest.cpp
+++ b/src/LinuxTracing/LinuxTracingIntegrationTest.cpp
@@ -4,6 +4,7 @@
 
 #include <absl/container/flat_hash_set.h>
 #include <absl/strings/numbers.h>
+#include <absl/synchronization/mutex.h>
 #include <absl/time/clock.h>
 #include <gmock/gmock.h>
 #include <gtest/gtest.h>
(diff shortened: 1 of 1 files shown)


src/LinuxTracing/LinuxTracingIntegrationTest.cpp
```diff
@@ -238,13 +244,22 @@ class LinuxTracingIntegrationTestFixture {
...
     listener_.emplace();
     tracer_->SetListener(&*listener_);
     tracer_->Start();
+
+    // Waiting for the first SchedulingSlice (at least one of which is always expected as long as
+    // capture_options.trace_context_switches() is true) guarantees that the main loop in
+    // TracerThread has started, and hence that the capture has been fully set up.
+    while (!listener_->HasReceivedAtLeastOneSchedulingSlice()) {
+      absl::SleepFor(absl::Milliseconds(10));
+    }
```
dimitry-, line 262: maybe wait on mutex for this condition?

dpallotti: Done.

dimitry- (approved):

PR #1743, opened 2021-01-27
Rename OrbitService -> Service

dimitry-:
Rename folder library and tests. Executable name is
preserved.

Test: builds


src/OrbitProducer/CMakeLists.txt
```diff
@@ -22,9 +22,9 @@ target_include_directories(OrbitProducer PRIVATE ${CMAKE_CURRENT_SOURCE_DIR})
 target_link_libraries(OrbitProducer PUBLIC
         GrpcProtos
         OrbitBase
-        OrbitServiceLib
         concurrentqueue::concurrentqueue
-        CONAN_PKG::abseil)
+        CONAN_PKG::abseil
+        ServiceLib)
```
dpallotti, line 27: What's the order here? I feel like the previous position was fine, in order with our modules and before the external ones.

dpallotti (approved):
Well, as the CI tells you there's one test to fix.

This was one of those components with a short name for which I had got used to the "Orbit" prefix. I hope everyone is fine with this particular renaming (the fact that the executable name is preserved kind of matches this doubt).

src/Service/CMakeLists.txt
```diff
@@ -5,12 +5,12 @@
 # To make OrbitAsioServer and OrbitGrpcServer available for fuzz-testing
 # the target OrbitServiceLib was introduced. OrbitService still exists,
 # but only builds main.cpp and links to OrbitServiceLib.
-project(OrbitServiceLib)
+project(ServiceLib)
```
beckerhe, line 8: At some point I'd like to split the library and the app target into separate directories. But that doesn't need to be in this PR. I just wanted to mention it, because I have struggled so far with finding proper names for these two things. If you happen to have a suggestion I'm all ears.

beckerhe (approved):
Thank you Dimitry. LGTM!

src/OrbitProducer/CMakeLists.txt
dimitry-: trying to keep it alphabetical... but they were not sorted before...

src/Service/CMakeLists.txt
dimitry-: ack.

src/OrbitProducer/CMakeLists.txt
dimitry-: restored.

PR #1744, opened 2021-01-27
Add some helpers for Future-based code

beckerhe:
As far as I can see, this will finally be the last PR for the promise framework.

It adds some tools which should simplify writing code with futures, making it more readable and easier to understand.

### PromiseHelpers.h

These are mainly small trait-like helper classes to deal with semantic difference between `Promise<T>` and `Promise<void>`.
Not so useful for the general use case, but helpful for simplifying `MainThreadExecutor::ScheduleAfter` (2nd commit).

### UnwrapFuture from FutureHelpers.h

This function unwraps (or flattens) nested futures. Imagine you have a type `Future<Future<T>>`. Calling UnwrapFuture on that type will return a future that completes when the inner future completes.

This is often useful when using `Future::Then` with a continuation that returns a Future.

```cpp

Future<ErrorMessageOr<void>> LoadModule(const std::string& module_path) {
  const auto load_module_future = main_thread_executor_->Schedule([this, module_path]() -> Future<ErrorMessageOr<void>> {
    const auto symbols_path = LoadModuleLocally(module_path);

    if (symbols_path.error()) return symbols_path.error();

    return LoadModuleOnRemote(module_path);
  });

  // Since the lambda returns a future, `load_module_future` is of type `Future<Future<ErrorMessageOr<void>>`.
  return UnwrapFuture(load_module_future);
```

### Public constructors for Futures

Sometimes it's useful to construct Futures without its asynchronous nature. Example:

```cpp
Future<ErrorMessageOr<std::string>> process(std::string input) {

  // Return early with error if precondition is not met.
  if (input.empty()) return Future<ErrorMessageOr<std::string>>{ErrorMessage{"input is invalid!"}};

  return thread_pool_->Schedule([input]() {
    // do expensive processing
    return output;
  });
}
```

That's why I added public constructors to the Future type which create valid, and completed futures that are not connected to any promise.

### DIFF
--- a/src/OrbitBase/include/OrbitBase/FutureHelpers.h
+++ b/src/OrbitBase/include/OrbitBase/FutureHelpers.h
@@ -1,0 +1,78 @@
+// Copyright (c) 2021 The Orbit Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#ifndef ORBIT_BASE_FUTURE_HELPERS_H_
+#define ORBIT_BASE_FUTURE_HELPERS_H_
+
+#include <type_traits>
+#include <utility>
+
+#include "OrbitBase/Future.h"
+#include "OrbitBase/Promise.h"
+
+namespace orbit_base {
+
+template <typename T, typename Invocable>
+void RegisterContinuationOrCallDirectly(const Future<T>& future, Invocable continuation) {
+  const auto result = future.RegisterContinuation(continuation);
+
+  if (result == FutureRegisterContinuationResult::kFutureAlreadyCompleted) {
+    continuation(future.Get());
+  }
+}
+
+template <typename Invocable>
+void RegisterContinuationOrCallDirectly(const Future<void>& future, Invocable continuation) {
+  const auto result = future.RegisterContinuation(continuation);
+
+  if (result == FutureRegisterContinuationResult::kFutureAlreadyCompleted) {
+    continuation();
+  }
+}
+
+template <typename T>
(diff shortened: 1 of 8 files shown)


src/OrbitBase/include/OrbitBase/FutureHelpers.h
```diff
@@ -0,0 +1,68 @@
...
+  return future;
+}
+
+template <typename T>
+[[nodiscard]] inline Future<T> UnwrapFuture(const Future<Future<T>>& outer_future) {
+  CHECK(outer_future.IsValid());
+
+  auto promise = std::make_shared<Promise<T>>();
+
+  auto outer_continuation = [promise](const Future<T>& inner_future) {
```
dimitry-, line 45: maybe short explanation of what is happening here. I spent some time unwrapping this code :)

dimitry- (approved):
otherwise lgtm

src/OrbitBase/include/OrbitBase/FutureHelpers.h
beckerhe: Yes, that's a very good idea. I also needed some time to figure out the correct implementation. Unit tests to the rescue. :-)

beckerhe: done.

antonrohr:
I have to admit I only understood ~half of whats going on here. So I am glad Dimitry had a look at this. 
Maybe you can hold a presentation about this for team one day. I think this might be valuable for a lot of us.

PR #1745, opened 2021-01-27
Add test CallstackSamplesWithFramePointersTogetherWithFunctionCalls

dpallotti:
Fix related bug: the `[uprobes]` map entry was not added to the `current_maps_`
in `UprobesUnwindingVisitor`, but we need it.

Bug: http://b/160756664

### DIFF
--- a/src/LinuxTracing/UprobesUnwindingVisitor.cpp
+++ b/src/LinuxTracing/UprobesUnwindingVisitor.cpp
@@ -191,10 +191,25 @@ void UprobesUnwindingVisitor::visit(UretprobesPerfEvent* event) {
 
 void UprobesUnwindingVisitor::visit(MmapPerfEvent* event) {
   CHECK(listener_ != nullptr);
-  const std::string& filename = event->filename();
 
-  ErrorMessageOr<orbit_grpc_protos::ModuleInfo> module_info_result =
-      orbit_elf_utils::CreateModule(filename, event->address(), event->address() + event->length());
+  // Obviously the uprobes map cannot be successfully processed by orbit_elf_utils::CreateModule,
+  // but it's important that current_maps_ contain it.
+  // For example, UprobesReturnAddressManager::PatchCallchain needs it to check whether a program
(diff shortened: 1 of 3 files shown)


src/LinuxTracing/UprobesUnwindingVisitor.cpp
```diff
@@ -191,10 +192,25 @@ void UprobesUnwindingVisitor::visit(UretprobesPerfEvent* event) {
...
+  // Obviously the uprobes map cannot be successfully processed by orbit_elf_utils::CreateModule,
+  // but it's important that current_maps_ contain it.
+  // For example, UprobesReturnAddressManager::PatchCallchain needs it to check whether a program
+  // counter is inside the uprobes map, and UprobesUnwindingVisitor::visit(StackSamplePerfEvent*)
+  // needs it to throw away incorrectly-unwound samples.
+  // As below we are only adding maps successfully parsed with orbit_elf_utils::CreateModule,
+  // we add the uprobes map manually. We are using the same values that that uprobes map would get
+  // if unwindstack::BufferMaps was built by passing the full content of /proc/<pid>/maps to its
+  // constructor.
+  if (event->filename() == "[uprobes]") {
```
dpallotti, line 205: @dimitry- I added this special case but I'm concerned about the fact that we only add an executable entry to the maps if it's actually backed by a file. Is this enough in general?

dimitry-: nice catch.

Yes, the .text section and section used by unwinder .eh_frame are in the first (executable) segment of the elf file.

Other maps reference stuff like data which we do not currently use. We can still add the rest of the segments if needed but at this point there is no need for them.

dimitry- (approved):

florian-kuebler (approved):
LGTM. Also thanks for fixing the maps.

src/LinuxTracing/UprobesUnwindingVisitor.cpp
dpallotti: Sounds good!

PR #1747, opened 2021-01-28
Introspection window is correctly freed on shutdown

reichlfl:
Fixes b/176509021

The introspection window is not parented under the main window, and is therefore not freed automatically when the UI is destroyed. Changing the pointer to unique_ptr fixes the issue.

In addition, closing Orbit while the introspection window is open will now automatically close the introspection.


beckerhe (approved):
Nice catch. LGTM

pierricgimmig:
Thanks @reichlfl !

PR #1749, opened 2021-01-28
Add Session UUID to MetricsUploader

antonrohr:
This adds a uuid for tracking sessions to the orbit_log_event_proto and
generates a uuid based on msdn libraries.

Test: Start orbit with --enable_metrics and then in Orbit.log there should be a line like this: 
[2021-01-28T10:51:39.738459] [MetricsUploader/MetricsUploaderWindows.cpp:77] Session UUID for metrics: b3d65cfa-8438-4c7a-8f59-a2368567b15a

### DIFF
--- a/src/MetricsUploader/MetricsUploaderWindows.cpp
+++ b/src/MetricsUploader/MetricsUploaderWindows.cpp
@@ -2,25 +2,31 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
 
+#include <Rpc.h>
 #include <absl/memory/memory.h>
+
+#include <type_traits>
 
 #include "MetricsUploader/MetricsUploader.h"
 #include "OrbitBase/Logging.h"
+#include "OrbitBase/Result.h"
 
 namespace orbit_metrics_uploader {
 
 constexpr const char* kSendLogEventFunctionName = "SendOrbitLogEvent";
 constexpr const char* kStartUploaderClientFunctionName = "StartUploaderClient";
 
-MetricsUploader::MetricsUploader(std::string client_name,
+MetricsUploader::MetricsUploader(std::string client_name, std::string session_uuid,
                                  Result (*send_log_event_addr)(const uint8_t*, int),
                                  HMODULE metrics_uploader_client_dll)
     : client_name_(std::move(client_name)),
+      session_uuid_(std::move(session_uuid)),
       send_log_event_addr_(send_log_event_addr),
       metrics_uploader_client_dll_(metrics_uploader_client_dll) {}
 
 MetricsUploader::MetricsUploader(MetricsUploader&& other)
     : client_name_(std::move(other.client_name_)),
+      session_uuid_(std::move(other.session_uuid_)),
       metrics_uploader_client_dll_(other.metrics_uploader_client_dll_),
       send_log_event_addr_(other.send_log_event_addr_) {
   other.metrics_uploader_client_dll_ = nullptr;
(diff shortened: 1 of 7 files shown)


src/MetricsUploader/MetricsUploaderWindows.cpp
```diff
@@ -56,34 +60,51 @@ ErrorMessageOr<MetricsUploader> MetricsUploader::CreateMetricsUploader(std::stri
...
+  RPC_CSTR uuid_c_str = nullptr;
+  RPC_STATUS convert_status = UuidToStringA(&uuid, &uuid_c_str);
+  if (convert_status != RPC_S_OK) {
+    return ErrorMessage("Unable to convert UUID to string for metrics uploader");
+  }
+
+  std::string session_uuid{reinterpret_cast<char*>(uuid_c_str)};
+  RpcStringFreeA(&uuid_c_str);
+  LOG("Session UUID for metrics: %s", session_uuid);
+
```
IrinaShkviro, line 78: Maybe we can move it to the separate function?

src/MetricsUploader/include/MetricsUploader/MetricsUploader.h
```diff
@@ -58,7 +58,7 @@ class MetricsUploader {
  private:
 #ifdef _WIN32
   HMODULE metrics_uploader_client_dll_;
-  explicit MetricsUploader(std::string client_name,
+  explicit MetricsUploader(std::string client_name, std::string session_uuid,
```
IrinaShkviro, line 61: I think you should modify the MetricsUploaderLinux as well

IrinaShkviro: Never mind, I'm wrong here

src/MetricsUploader/MetricsUploaderWindows.cpp
```diff
@@ -56,34 +60,51 @@ ErrorMessageOr<MetricsUploader> MetricsUploader::CreateMetricsUploader(std::stri
     return ErrorMessage("MetricsUploader is already created");
   }
 
+  // Generate UUID as a session identifier.
```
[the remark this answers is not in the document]
beckerhe: Ah. Just saw, that Irina made the comment in the meantime. Sorry for bringing it up twice.

src/MetricsUploader/MetricsUploaderWindows.cpp
IrinaShkviro: I'm also not sure that we should fail if the error occurred during session_id generation. I think most of the metrics would still make sense. @antonrohr  @beckerhe WDYT?

beckerhe: In general I agree. But UUID generation doesn't fail nowadays. This API function is very old and was designed when it was still common to deduce the UUID from a bit of randomness and the MAC address.

So I guess we could also just fail when UUID generation fails and simplify our code that way. It will not happen in practice.

src/MetricsUploader/MetricsUploaderWindows.cpp
```diff
@@ -56,34 +60,51 @@ ErrorMessageOr<MetricsUploader> MetricsUploader::CreateMetricsUploader(std::stri
...
+  if (create_status != RPC_S_OK) {
+    return ErrorMessage("Unable to create UUID for metrics uploader");
+  }
+  RPC_CSTR uuid_c_str = nullptr;
+  RPC_STATUS convert_status = UuidToStringA(&uuid, &uuid_c_str);
+  if (convert_status != RPC_S_OK) {
+    return ErrorMessage("Unable to convert UUID to string for metrics uploader");
+  }
+
+  std::string session_uuid{reinterpret_cast<char*>(uuid_c_str)};
```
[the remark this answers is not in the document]
antonrohr: uuid_c_str is a null terminated `unsigned char*`. I tried static cast, but it did not work out of the box. Any other ideas?

src/MetricsUploader/MetricsUploaderWindows.cpp
antonrohr: I moved it into a seperate function and added a quick test case. ptal

src/MetricsUploader/MetricsUploaderWindows.cpp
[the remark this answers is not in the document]
antonrohr: No worries. The unit test was a good idea

src/MetricsUploader/MetricsUploaderWindows.cpp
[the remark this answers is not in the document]
beckerhe: Ah I see. RPC_CSTR is probably `unsigned char*` under the hood. So that's unfortunate. It's probably best to leave the reinterpret_cast, but to protect it against misuse. The problem with reinterpret_cast is that it would also work if the type of RPC_CSTR changed. So I would add a static_assert which ensures that RPC_CSTR is of type `usigned char*`.

```cpp
#include <type_traits>
static_assert(std::is_same_v<RPC_CSTR, unsigned char*>, "The type of RPC_CSTR needs to be castable to const char*");
```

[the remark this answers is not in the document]
beckerhe: Obviously the type of RPC_CSTR will never change. This is more to show others that you thought about potential problems with reinterpret_cast. A comment (or both) could also be enough.

src/MetricsUploader/MetricsUploaderWindowsTest.cpp
```diff
@@ -53,3 +55,26 @@ TEST(MetricsUploader, CreateMetricsUploaderFromNonexistentClient) {
       MetricsUploader::CreateMetricsUploader("NonexistentMetricsUploaderClient");
   EXPECT_EQ(metrics_uploader.has_value(), false);
 }
+
+TEST(MetricsUploader, GenerateUUID) {
```
beckerhe, line 59: I would split this up into 3 tests.
- The first one checks if generation works.
- The second one checks the format.
- The third checks for uniqueness.

For the third I would also check more than 2 UUIDs. Maybe 100 or even 1000. By putting them into a hash set you should be able to easily check the uniqueness.

src/MetricsUploader/MetricsUploaderWindows.cpp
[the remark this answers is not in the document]
antonrohr: This was a great idea! Thank you, I added it

src/MetricsUploader/MetricsUploaderWindowsTest.cpp
antonrohr: done

beckerhe (approved):
Thank you. LGTM!

PR #1750, opened 2021-01-28
Imgui shutdown fix

reichlfl:
Fixes b/176540571 by correcting the lifetimes of ImGui resources and the ImGui context.

The first commit removes the font_size parameters to prepare for this. In hindsight, this might not have been completely required, but IMHO clearer separates the font_size used for our UI from the ImGui font size. 

Detailed comments on the actual change are in the last commit.

### DIFF
--- a/src/OrbitGl/App.cpp
+++ b/src/OrbitGl/App.cpp
@@ -468,7 +468,7 @@ void OrbitApp::RefreshCaptureView() {
   DoZoom = true;  // TODO: remove global, review logic
 }
 
-void OrbitApp::RenderImGui() {
+void OrbitApp::RenderImGuiDebugUI() {
   CHECK(debug_canvas_ != nullptr);
   CHECK(capture_window_ != nullptr);
   ScopeImguiContext context(debug_canvas_->GetImGuiContext());
@@ -485,13 +485,13 @@ void OrbitApp::RenderImGui() {
 
   if (ImGui::BeginTabBar("DebugTabBar", ImGuiTabBarFlags_None)) {
     if (ImGui::BeginTabItem("CaptureWindow")) {
-      capture_window_->RenderImGui();
+      capture_window_->RenderImGuiDebugUI();
       ImGui::EndTabItem();
     }
 
     if (introspection_window_) {
       if (ImGui::BeginTabItem("Introspection")) {
-        introspection_window_->RenderImGui();
+        introspection_window_->RenderImGuiDebugUI();
         ImGui::EndTabItem();
       }
     }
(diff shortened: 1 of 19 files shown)


antonrohr (approved):
Looks like a good change to me. I did not actually check it out and tested if everything imgui related still works. I assume you did?!

karupayun (approved):
LGTM Florian! :) 

Seems that it's also solving my related crash. :)

Small nits in 3rd commit description:
"and ImGui always has a _the_ created context" (erase "the")
"Our implementation **has a strange** mix, **as** it suggests"

src/OrbitGl/App.cpp
```diff
@@ -1167,42 +1168,41 @@ void OrbitApp::LoadSymbols(const std::filesystem::path& symbols_path, ModuleData
...
+                                         std::move(frame_track_function_hashes)] {
+      modules_currently_loading_.erase(module_data->file_path());
+
+      const ProcessData* selected_process = GetTargetProcess();
+      if (selected_process != nullptr &&
+          selected_process->IsModuleLoaded(module_data->file_path())) {
+        functions_data_view_->AddFunctions(module_data->GetFunctions());
+        LOG("Added loaded function symbols for module \"%s\" to the functions tab",
+            module_data->file_path());
+      }
```
beckerhe, line 1181: I think some unrelated clang-format changes have slipped into this commit. Or am I not seeing the change here?

reichlfl: Looks like it, this should not be here. I can revert this and push it in a separate PR since the changes seem to be reasonable, WDYT?

reichlfl:
> LGTM Florian! :)
> 
> Seems that it's also solving my related crash. :)
> 
> Small nits in 3rd commit description:
> "and ImGui always has a _the_ created context" (erase "the")
> "Our implementation **has a strange** mix, **as** it suggests"

Thanks for spotting, I'll fix that with the rebase!

src/OrbitGl/App.cpp
beckerhe: My PR #1752 is going to change that anyway. So for me it's fine if you just remove the change. And I have to admit, my motivation to report this was mainly driven by the fact that this would lead to a merge conflict in my PR. :sweat_smile:

beckerhe: Thank you. :innocent:

reichlfl:
> Looks like a good change to me. I did not actually check it out and tested if everything imgui related still works. I assume you did?!

Yes, works all fine for me. Nevertheless I'll wait with the merge tomorrow to give @pierricgimmig to take a look as he know this part probably best.

pierricgimmig (approved):
Hi @reichlfl , thanks a lot for cleaning this up. The only minor thing is that the ImGui font is now hardcoded whereas before it was tied to the initial font size at Orbit launch. It's not a huge deal, especially since we don't serialize the font size anymore it seems. In any case, it's easy to dynamically change the font size in ImGui, so we could hook it up to our global font size as well if need be. LGTM, thanks!

PR #1751, opened 2021-01-28
Process ThreadNames in order and set their pid

dpallotti:


src/LinuxTracing/PerfEvent.h
```diff
@@ -379,7 +379,10 @@ class TaskNewtaskPerfEvent : public TracepointPerfEvent {
   void Accept(PerfEventVisitor* visitor) override;
 
   // The tracepoint format calls this "pid" but it's effectively the thread id.
-  pid_t GetTid() const { return GetTypedTracepointData<task_newtask_tracepoint>().pid; }
+  // Note that ring_buffer_record.sample_id.cpu.pid and ring_buffer_record.sample_id.cpu.tid
```
dpallotti, line 382: ".cpu" is extra here and below, need to fix that.

src/LinuxTracing/ContextSwitchAndThreadStateVisitor.cpp
```diff
@@ -85,6 +86,16 @@ void ContextSwitchAndThreadStateVisitor::ProcessInitialState(uint64_t timestamp_
 }
 
 void ContextSwitchAndThreadStateVisitor::visit(TaskNewtaskPerfEvent* event) {
+  std::optional<pid_t> new_pid = GetPidOfTid(event->GetNewTid());
+  ThreadName thread_name;
+  if (new_pid.has_value()) {
+    thread_name.set_pid(new_pid.value());
```
dimitry-, line 92: Maybe make it explicit set_pid(new_pid.value_or(-1)) ?

Also when does it happen? (add a comment?)

dimitry- (approved):
otherwise lgtm

src/LinuxTracing/ContextSwitchAndThreadStateVisitor.cpp
dpallotti: Done, more in line with context switches.

This happens if something goes wrong with the bookkeeping of pid-tid associations, so hopefully never. I'd rather not add a comment because this is a common theme of the class, not a special case here.

src/OrbitCaptureClient/CaptureEventProcessor.cpp
```diff
@@ -256,6 +256,7 @@ void CaptureEventProcessor::ProcessGpuQueueSubmission(
 }
 
 void CaptureEventProcessor::ProcessThreadName(const ThreadName& thread_name) {
+  // Note: thread_name.pid() is available, but currently dropped.
```
florian-kuebler, line 259: Will you carry the pid to the timer in an upcoming PR?

florian-kuebler (approved):
LGTM. Thanks!

src/OrbitCaptureClient/CaptureEventProcessor.cpp
dpallotti: It's not in my plans, no, mostly because if I'm not wrong the plan with @dimitry-'s new capture format is that we'll get rid of the "client protos". This change is meant to make sure the pids end up in the protos that have such a field in the saved capture, once we use a single set of protos for both communication and serialization.

PR #1752, opened 2021-01-28
Refactor module loading in OrbitApp

beckerhe:
This PR refactors some function in OrbitApp that are concerned with symbol or module loading.

### The initial problem
The function `OrbitApp::LoadModules` has the following function signature:

```cpp
void LoadModules(
      const std::vector<ModuleData*>& modules,
      absl::flat_hash_map<std::string, std::vector<uint64_t>> function_hashes_to_hook_map = {},
      absl::flat_hash_map<std::string, std::vector<uint64_t>> frame_track_function_hashes_map = {});
```

The whole function works asynchronously, meaning when the function returns the work is not done.
It first loads the modules given by `modules` and then hooks functions given by the second parameter
and enable frame tracks given by the third parameter.

Adding another task that has to happen after loading a module would mean adding another parameter to the function.
I think it's obvious that this does not scale and violates the principle of separation of concerns.

The proper solution is to add a mechanism which can tell a caller of `LoadModules` when module loading is done
and when it's safe to trigger follow-up work. The recently introduced Future-type is made for that purpose.

### The new `LoadModule` method

The function signature of the new `LoadModule` method looks like this:

```cpp
orbit_base::Future<void> LoadModule(const ModuleData* module);
```

It now returns a future which can tell the caller when the asynchronous task is done. It can be either ignored,
or transformed into a synchronous task by using `MainThreadExecutor::WaitFor`:

```cpp

auto future = LoadModule(module);

main_thread_executor->WaitFor(future);
// This call is blocking. MainThreadExecutor will process other UI events in the mean time.

for(const auto& function : functions) SelectFunction(function); // Or other follow-up tasks
```

The third option is chain a continuation using `Future::Then`:

```cpp

LoadModule(module).Then(main_thread_executor_, [functions](){
  // This lambda is executed on the main thread when `LoadModule` is done.

  for(const auto& function : functions) SelectFunction(function);
});
```

Note: Of course there is also a replacement for the previous `LoadModules` method:

```cpp
orbit_base::Future<void> LoadModules(absl::Span<const ModuleData* const> modules);
```

It takes a span (think a vector) of ModuleData pointers and returns a future that completes
when all modules have been loaded (or at least have been tried to load).

### DIFF
--- a/src/OrbitGl/App.cpp
+++ b/src/OrbitGl/App.cpp
@@ -1366,54 +1366,31 @@ ErrorMessageOr<void> OrbitApp::EnableFrameTracksFromHashes(
 }
 
 void OrbitApp::LoadPreset(const std::shared_ptr<PresetFile>& preset_file) {
-  std::vector<ModuleData*> modules_to_load;
-  std::vector<std::string> module_paths_not_found;  // file path of module
-  for (const auto& [module_path, preset_module] : preset_file->preset_info().path_to_module()) {
-    ModuleData* module_data = module_manager_->GetMutableModuleByPath(module_path);
+  std::vector<std::string> module_paths_not_found;
+  std::vector<orbit_base::Future<void>> futures;
 
-    if (module_data == nullptr) {
-      module_paths_not_found.push_back(module_path);
-      continue;
-    }
-    if (module_data->is_loaded()) {
-      std::vector<uint64_t> function_hashes{preset_module.function_hashes().begin(),
-                                            preset_module.function_hashes().end()};
-      const auto selecting_result = SelectFunctionsFromHashes(module_data, function_hashes);
-      if (!selecting_result) {
-        LOG("Warning: %s", selecting_result.error().message());
+  for (const auto& module_pair : preset_file->preset_info().path_to_module()) {
+    auto future = main_thread_executor_->Schedule([this, module_pair, &module_paths_not_found]() {
+      const auto& [module_path, preset_module] = module_pair;
+      const auto result = LoadPresetModule(module_path, preset_module);
+
+      if (result == LoadPresetModuleResult::kModuleNotFound) {
+        module_paths_not_found.emplace_back(module_path);
       }
-      std::vector<uint64_t> frame_track_hashes{preset_module.frame_track_function_hashes().begin(),
-                                               preset_module.frame_track_function_hashes().end()};
-      const auto frame_track_result = EnableFrameTracksFromHashes(module_data, frame_track_hashes);
-      if (!frame_track_result) {
-        LOG("Warning: %s", frame_track_result.error().message());
-      }
-      continue;
-    }
-    modules_to_load.push_back(module_data);
+    });
+    futures.emplace_back(std::move(future));
   }
+
+  const auto result = main_thread_executor_->WaitForAll(absl::MakeSpan(futures));
+  if (result != MainThreadExecutor::WaitResult::kCompleted) return;
+
   if (!module_paths_not_found.empty()) {
     // Note that unloadable presets are disabled.
     SendWarningToUi("Preset only partially loaded",
                     absl::StrFormat("The following modules are not loaded:\n\"%s\"",
                                     absl::StrJoin(module_paths_not_found, "\"\n\"")));
   }
-  if (!modules_to_load.empty()) {
-    absl::flat_hash_map<std::string, std::vector<uint64_t>> function_hashes_to_hook_map;
-    absl::flat_hash_map<std::string, std::vector<uint64_t>> frame_track_function_hashes_map;
-    for (const auto& [module_path, preset_module] : preset_file->preset_info().path_to_module()) {
-      function_hashes_to_hook_map[module_path] = std::vector<uint64_t>{};
-      for (const uint64_t function_hash : preset_module.function_hashes()) {
-        function_hashes_to_hook_map.at(module_path).push_back(function_hash);
-      }
-      frame_track_function_hashes_map[module_path] = std::vector<uint64_t>{};
-      for (const uint64_t function_hash : preset_module.frame_track_function_hashes()) {
-        frame_track_function_hashes_map.at(module_path).push_back(function_hash);
-      }
-    }
-    LoadModules(modules_to_load, std::move(function_hashes_to_hook_map),
-                std::move(frame_track_function_hashes_map));
-  }
+
   FireRefreshCallbacks();
 }
 
--- a/src/OrbitGl/App.cpp
+++ b/src/OrbitGl/App.cpp
@@ -1459,8 +1459,25 @@ void OrbitApp::UpdateProcessAndModuleList(int32_t pid) {
         }
       }
       // (D) Load Modules again (and pass on functions to hook after loading)
-      LoadModules(modules_to_reload, std::move(function_hashes_to_hook_map),
-                  std::move(frame_track_function_hashes_map));
+      const auto future = LoadModules(modules_to_reload);
+
+      if (main_thread_executor_->WaitFor(future) != MainThreadExecutor::WaitResult::kCompleted) {
+        return;
+      }
+
+      for (const auto& [module_path, function_hashes] : function_hashes_to_hook_map) {
+        ModuleData* const module_data = GetMutableModuleByPath(module_path);
+        if (module_data == nullptr) continue;
+        (void)SelectFunctionsFromHashes(module_data, function_hashes);
+        LOG("Auto hooked functions in module \"%s\"", module_data->file_path());
+      }
+
+      for (const auto& [module_path, function_hashes] : frame_track_function_hashes_map) {
+        ModuleData* const module_data = GetMutableModuleByPath(module_path);
+        if (module_data == nullptr) continue;
+        (void)EnableFrameTracksFromHashes(module_data, function_hashes);
+        LOG("Added frame tracks in module \"%s\"", module_data->file_path());
+      }
 
       // Refresh UI
       modules_data_view_->UpdateModules(process);
--- a/src/OrbitGl/App.cpp
+++ b/src/OrbitGl/App.cpp
@@ -1206,56 +1206,34 @@ void OrbitApp::LoadSymbols(const std::filesystem::path& symbols_path, ModuleData
   });
 }
 
-ErrorMessageOr<void> OrbitApp::GetFunctionInfosFromHashes(
-    const ModuleData* module, const std::vector<uint64_t>& function_hashes,
-    std::vector<const FunctionInfo*>* function_infos) {
-  CHECK(module != nullptr);
-  const ProcessData* process = GetTargetProcess();
-  if (process == nullptr) {
-    return ErrorMessage(absl::StrFormat(
-        "Unable to get function infos for module \"%s\", because no process is selected",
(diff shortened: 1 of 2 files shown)


src/OrbitGl/App.cpp
```diff
@@ -1147,6 +1145,13 @@ ErrorMessageOr<std::filesystem::path> OrbitApp::FindSymbolsLocally(
   return ErrorMessage(error_message);
 }
 
+ErrorMessageOr<std::filesystem::path> OrbitApp::FindModuleLocally(
+    const std::filesystem::path& module_path, const std::string& build_id) {
+  const auto scoped_status = CreateScopedStatus(absl::StrFormat(
+      "Searching for symbols on local machine (module: \"%s\")...", module_path.string()));
```
dpallotti, line 1165: Especially since this ends up in the status bar, can we format it a bit better. The parentheses in particular make it a bit unusual. What about something like (but there's probably a better way and it depends on how other status bar message are formatted)
`"Searching for symbols on local machine for module: \"%s\"..."`
?

src/OrbitGl/App.cpp
```diff
@@ -1152,63 +1152,82 @@ ErrorMessageOr<std::filesystem::path> OrbitApp::FindModuleLocally(
...
+                          frame_track_result.error().message());
+                    }
+                    LOG("Added frame tracks in module \"%s\"", module_data->file_path());
+                  }
+
+                  UpdateAfterSymbolLoading();
+                  FireRefreshCallbacks();
+                });
+
+  (void)main_thread_executor_->WaitFor(future);
```
dpallotti: Aaah `[[nodiscard]]`.

src/OrbitGl/App.cpp
```diff
@@ -999,65 +999,73 @@ void OrbitApp::SendErrorToUi(const std::string& title, const std::string& text)
...
-              "Error loading symbols",
-              absl::StrFormat("Did not find symbols locally or on remote for module \"%s\": %s\n%s",
-                              module_data->file_path(), error_message_from_local,
-                              result.error().message()));
-          main_thread_executor_->Schedule([this, module_data]() {
-            modules_currently_loading_.erase(module_data->file_path());
-          });
-          return;
-        }
+      "Searching for symbols on remote instance (module \"%s\")...", module_file_path));
```
dpallotti, line 1008: As written for a previous commit, I think this message should not use the parentheses.

src/OrbitGl/App.cpp
```diff
@@ -999,65 +999,73 @@ void OrbitApp::SendErrorToUi(const std::string& title, const std::string& text)
...
+        const std::filesystem::path local_debug_file_path =
+            symbol_helper_.GenerateCachedFileName(module_file_path);
+
+        scoped_status.UpdateMessage(
+            absl::StrFormat(R"(Copying debug info file for "%s" from remote: "%s"...)",
+                            module_file_path, debug_file_path));
+        SCOPED_TIMED_LOG("Copying \"%s\"", debug_file_path);
+        auto scp_result = secure_copy_callback_(debug_file_path, local_debug_file_path.string());
+        if (!scp_result) {
+          return ErrorMessage{absl::StrFormat("Could not copy debug info file from the remote: %s",
```
dpallotti, line 1033: The message above says "from remote" while here we use the form "from the remote".

src/OrbitGl/App.cpp
```diff
@@ -1206,56 +1206,34 @@ void OrbitApp::LoadSymbols(const std::filesystem::path& symbols_path, ModuleData
...
-    const FunctionInfo* function = module->FindFunctionFromHash(function_hash);
-    if (function == nullptr) {
-      count_missing++;
+ErrorMessageOr<void> OrbitApp::SelectFunctionsFromHashes(
+    const ModuleData* module, const std::vector<uint64_t>& function_hashes) {
+  for (const auto function_hash : function_hashes) {
+    const orbit_client_protos::FunctionInfo* const function_info =
+        module->FindFunctionFromHash(function_hash);
+    if (function_info == nullptr) {
+      LOG("Could not find function hash 0x%x in module %s", function_hash, module->file_path());
```
dpallotti, line 1251: Also put `""` around the module name?

src/OrbitGl/App.cpp
```diff
@@ -1297,6 +1297,44 @@ void OrbitApp::LoadSymbols(const std::filesystem::path& symbols_path, ModuleData
...
+  if (!module_data->is_loaded()) {
+    const auto result = main_thread_executor_->WaitFor(LoadModule(module_data));
+    if (result != MainThreadExecutor::WaitResult::kCompleted) return LoadPresetModuleResult::kAbort;
+  }
+
+  for (const uint64_t function_hash : preset_module.function_hashes()) {
+    const orbit_client_protos::FunctionInfo* const function_info =
+        module_data->FindFunctionFromHash(function_hash);
+    if (function_info == nullptr) {
+      LOG("Could not find function hash 0x%x in module %s", function_hash,
```
dpallotti, line 1223: Well, same as in the first commit.

src/OrbitGl/App.cpp
```diff
@@ -1297,6 +1297,44 @@ void OrbitApp::LoadSymbols(const std::filesystem::path& symbols_path, ModuleData
...
+    }
+    SelectFunction(*function_info);
+  }
+  LOG("Auto hooked functions in module \"%s\"", module_data->file_path());
+
+  for (const uint64_t function_hash : preset_module.frame_track_function_hashes()) {
+    const orbit_client_protos::FunctionInfo* const function_info =
+        module_data->FindFunctionFromHash(function_hash);
+    if (function_info == nullptr) {
+      LOG("Could not find function hash 0x%x in module %s", function_hash,
```
dpallotti, line 1235: And same here.

src/OrbitGl/App.cpp
```diff
@@ -1459,8 +1459,25 @@ void OrbitApp::UpdateProcessAndModuleList(int32_t pid) {
...
+      const auto future = LoadModules(modules_to_reload);
+
+      if (main_thread_executor_->WaitFor(future) != MainThreadExecutor::WaitResult::kCompleted) {
+        return;
+      }
+
+      for (const auto& [module_path, function_hashes] : function_hashes_to_hook_map) {
+        ModuleData* const module_data = GetMutableModuleByPath(module_path);
+        if (module_data == nullptr) continue;
+        (void)SelectFunctionsFromHashes(module_data, function_hashes);
```
dpallotti, line 1378: Shouldn't the returned error used (even if only for a log)?

src/OrbitGl/App.cpp
```diff
@@ -1459,8 +1459,25 @@ void OrbitApp::UpdateProcessAndModuleList(int32_t pid) {
...
+        LOG("Auto hooked functions in module \"%s\"", module_data->file_path());
+      }
+
+      for (const auto& [module_path, function_hashes] : frame_track_function_hashes_map) {
+        ModuleData* const module_data = GetMutableModuleByPath(module_path);
+        if (module_data == nullptr) continue;
+        (void)EnableFrameTracksFromHashes(module_data, function_hashes);
```
dpallotti, line 1385: Same here.

dpallotti (commented):
I'm not sure I fully grasped every detail but it looks good, thanks for these improvements!

The only thing that stands out for me is that we should try to stay consistent with the format of the strings that we show in the status bar.

src/OrbitGl/App.cpp
```diff
@@ -999,121 +1002,130 @@ void OrbitApp::SendErrorToUi(const std::string& title, const std::string& text)
...
-      function_hashes_to_hook = function_hashes_to_hook_map.at(module->file_path());
-    }
+  return orbit_base::JoinFutures(futures);
+}
 
-    std::vector<uint64_t> frame_track_function_hashes;
-    if (frame_track_function_hashes_map.contains(module->file_path())) {
-      frame_track_function_hashes = frame_track_function_hashes_map.at(module->file_path());
-    }
+orbit_base::Future<void> OrbitApp::LoadModule(const ModuleData* module) {
```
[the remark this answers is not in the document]
antonrohr: I actually dont really know how to make this better. Am I assuming correctly, that you did the local variable lambdas to avoid deep nesting with `.then`?

src/OrbitGl/App.cpp
```diff
@@ -1147,166 +1159,145 @@ ErrorMessageOr<std::filesystem::path> OrbitApp::FindSymbolsLocally(
...
+  if (module_data == nullptr) {
+    return LoadPresetModuleResult::kModuleNotFound;
   }
-  if (!process->IsModuleLoaded(module->file_path())) {
-    return ErrorMessage(absl::StrFormat(
-        R"(Unable to get function infos for module "%s", because the module is not loaded by process "%s")",
-        module->file_path(), process->name()));
+
+  if (!module_data->is_loaded()) {
+    const auto result = main_thread_executor_->WaitFor(LoadModule(module_data));
```
antonrohr, line 1215: `WaitFor()` means that this function potentially takes a long time right?. Its not blocking, but "waiting". This is fine, because this function is only used inside a future (in `LoadPreset`), but maybe we should start indicating that in the function name. Maybe `LoadPresetModuleWaiting()`. 
Or maybe this function should have return type `Future<OrbitApp::LoadPresetModuleResult>`? (Did not fully think this through)

src/OrbitGl/App.cpp
```diff
@@ -1147,166 +1159,145 @@ ErrorMessageOr<std::filesystem::path> OrbitApp::FindSymbolsLocally(
...
+      continue;
+    }
+    EnableFrameTrack(*function_info);
   }
-  return outcome::success();
+  LOG("Added frame tracks in module \"%s\"", module_data->file_path());
+  return LoadPresetModuleResult::kSuccess;
 }
 
 ErrorMessageOr<void> OrbitApp::SelectFunctionsFromHashes(
```
antonrohr, line 1245: This function always returns `success` now. Maybe return type should be void. Maybe we should think about whether the information that only some functions from a preset could be hooked is valuable for the user (and display a warning dialog)
Same for EnableFrameTrackFromHashes

src/OrbitGl/App.cpp
```diff
@@ -1147,166 +1159,145 @@ ErrorMessageOr<std::filesystem::path> OrbitApp::FindSymbolsLocally(
...
+        module_paths_not_found.emplace_back(module_path);
       }
-      continue;
-    }
-    modules_to_load.push_back(module_data);
+    });
+    futures.emplace_back(std::move(future));
   }
+
+  const auto result = main_thread_executor_->WaitForAll(absl::MakeSpan(futures));
```
antonrohr, line 1291: Same as my last comment about `WaitFor`

src/OrbitGl/App.cpp
antonrohr: See me other comments. `SelectFunctionsFromHashes()` always returns success

beckerhe:
Hey @antonrohr, hey @dpallotti,
sorry for coming back to this so late. Thanks a lot for your review.

Seeing your remarks I get the feeling that this PR is just too big. Sorry for putting this out in the first place.
I will split this up into smaller chunks, that can be reviewed independently. I will incorporate your feedback.

Sorry again,
Henning

PR #1753, opened 2021-01-28
Rework log files to support multiple instances of Orbit at the same time

vickyliu-go4it:
Currently there is only one log file named as `Orbit.log` in the log directory, and this file is overwritten every time when starting an Orbit instance. Therefore, if multiple Orbit instances are running at the same time, they will write to the same log file.

To address the above issue, we rename the log file in a format of `Orbit-<yyyy_MM_dd_HH_mm_ss>-<pid>.log`, and remove old log files that are older than one week.

Bug: http://b/177978487.
Test: 1) Open several Orbit instances at the same time, check the name of log files in the `%APPDATA%\OrbitProfiler\logs\` folder;
2) Create some log files with names indicate a much earlier time, e.g. `Orbit-2020_01_01_00_00_00-12345.log`.  Open a new Orbit instance and see the old log files are removed.

### DIFF
--- a/src/OrbitBase/Logging.cpp
+++ b/src/OrbitBase/Logging.cpp
@@ -6,14 +6,23 @@
 
 #include <array>
 
+#include "OrbitBase/ThreadUtils.h"
 #include "absl/base/const_init.h"
 #include "absl/debugging/stacktrace.h"
 #include "absl/debugging/symbolize.h"
 #include "absl/strings/str_format.h"
 #include "absl/synchronization/mutex.h"
+#include "absl/time/time.h"
 
 static absl::Mutex log_file_mutex(absl::kConstInit);
 std::ofstream log_file;
+
+std::string GetLogFileName() {
+  std::string timestamp_string =
+      absl::FormatTime(kLogFileNameTimeFormat, absl::Now(), absl::LocalTimeZone());
+  return absl::StrFormat(kLogFileNameDelimiter, timestamp_string,
+                         static_cast<uint32_t>(orbit_base::GetCurrentProcessId()));
+}
 
 void InitLogFile(const std::filesystem::path& path) {
   absl::MutexLock lock(&log_file_mutex);
--- a/src/OrbitBase/Logging.cpp
+++ b/src/OrbitBase/Logging.cpp
@@ -4,15 +4,18 @@
 
 #include "OrbitBase/Logging.h"
 
-#include <array>
+#include <absl/base/const_init.h>
+#include <absl/debugging/stacktrace.h>
+#include <absl/debugging/symbolize.h>
+#include <absl/strings/str_format.h>
+#include <absl/synchronization/mutex.h>
+#include <absl/time/time.h>
 
+#include <array>
+#include <system_error>
+
+#include "LoggingUtils.h"
 #include "OrbitBase/ThreadUtils.h"
-#include "absl/base/const_init.h"
-#include "absl/debugging/stacktrace.h"
-#include "absl/debugging/symbolize.h"
-#include "absl/strings/str_format.h"
-#include "absl/synchronization/mutex.h"
-#include "absl/time/time.h"
 
 static absl::Mutex log_file_mutex(absl::kConstInit);
 std::ofstream log_file;
@@ -22,6 +25,12 @@ std::string GetLogFileName() {
       absl::FormatTime(kLogFileNameTimeFormat, absl::Now(), absl::LocalTimeZone());
   return absl::StrFormat(kLogFileNameDelimiter, timestamp_string,
                          static_cast<uint32_t>(orbit_base::GetCurrentProcessId()));
+}
+
+ErrorMessageOr<void> TryRemoveOldLogFiles(const std::filesystem::path& log_dir) {
+  std::vector<std::filesystem::path> log_file_paths = ListFiles(log_dir);
+  std::vector<std::filesystem::path> old_files = FilterOldLogFiles(log_file_paths);
+  return RemoveFiles(old_files);
 }
 
 void InitLogFile(const std::filesystem::path& path) {
--- a/src/OrbitBase/LoggingUtils.h
+++ b/src/OrbitBase/LoggingUtils.h
@@ -1,0 +1,27 @@
+// Copyright (c) 2021 The Orbit Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#ifndef ORBIT_BASE_LOGGING_UTILS_H_
+#define ORBIT_BASE_LOGGING_UTILS_H_
+
+#include <absl/strings/str_format.h>
+#include <absl/time/time.h>
+
+#include <filesystem>
+
+#include "OrbitBase/Result.h"
+
+constexpr const char* kLogFileNameTimeFormat = "%Y_%m_%d_%H_%M_%S";
+constexpr absl::string_view kLogFileNameDelimiter = "Orbit-%s-%u.log";
+
+[[nodiscard]] std::vector<std::filesystem::path> ListFiles(const std::filesystem::path& log_dir);
+ErrorMessageOr<absl::Duration> GetLogFileLifetime(const std::string& log_file_name);
+[[nodiscard]] std::vector<std::filesystem::path> FilterOldLogFiles(
+    const std::vector<std::filesystem::path>& log_file_paths);
+// This function tries to remove files even when an error is returned. If some files are unable to
+// remove, it returns an error message to record names of those functions and details about the
+// remove failures.
+ErrorMessageOr<void> RemoveFiles(const std::vector<std::filesystem::path>& log_file_paths);
+
+#endif  // ORBIT_BASE_LOGGING_UTILS_H_
--- a/src/OrbitBase/LoggingUtilsTest.cpp
+++ b/src/OrbitBase/LoggingUtilsTest.cpp
@@ -1,0 +1,64 @@
+// Copyright (c) 2021 The Orbit Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#include <absl/strings/str_format.h>
+#include <absl/time/clock.h>
+#include <gmock/gmock.h>
+#include <gtest/gtest.h>
+
+#include "LoggingUtils.h"
+
+namespace {
+std::filesystem::path GenerateTestLogFilePath(const absl::Time& timestamp) {
+  std::filesystem::path test_log_dir = "C:/OrbitAppDataDir/logs";
+  uint32_t test_pid = 12345;
+  std::string timestamp_string =
+      absl::FormatTime(kLogFileNameTimeFormat, timestamp, absl::LocalTimeZone());
+  std::string filename = absl::StrFormat(kLogFileNameDelimiter, timestamp_string, test_pid);
+  return test_log_dir / filename;
+}
+}  // namespace
+
+namespace orbit_base {
+
+TEST(LoggingUtils, GetLogFileLifetime) {
+  const std::string kFilenameInvalidNoTimestamp = "sfsdf-.log ";
+  const std::string kFilenameInvalidValidTimestampWrongFormat =
+      "Orbitfoobar-2021_01_31_00_00_00-.log";
+  const std::string kFilenameValid = "Orbit-2021_01_31_00_00_00-7188.log";
+
+  ErrorMessageOr<absl::Duration> result_extract_failed =
+      GetLogFileLifetime(kFilenameInvalidNoTimestamp);
+  ASSERT_FALSE(result_extract_failed);
+  EXPECT_EQ(result_extract_failed.error().message(),
+            absl::StrFormat("Unable to extract time information from log file: %s",
+                            kFilenameInvalidNoTimestamp));
+
+  ErrorMessageOr<absl::Duration> result_parse_failed =
+      GetLogFileLifetime(kFilenameInvalidValidTimestampWrongFormat);
+  ASSERT_FALSE(result_parse_failed);
+  EXPECT_THAT(
+      result_parse_failed.error().message(),
+      testing::HasSubstr(absl::StrFormat("Error while parsing time information from log file %s",
+                                         kFilenameInvalidValidTimestampWrongFormat)));
+
+  ErrorMessageOr<absl::Duration> result_parse_succeed = GetLogFileLifetime(kFilenameValid);
+  absl::Duration expected_result =
+      absl::Now() - absl::FromCivil(absl::CivilSecond(2021, 1, 31, 0, 0, 0), absl::LocalTimeZone());
+  EXPECT_NEAR(absl::ToDoubleSeconds(result_parse_succeed.value()),
+              absl::ToDoubleSeconds(expected_result), 1);
+}
+
+TEST(LoggingUtils, FilterOldLogFiles) {
+  absl::Time now = absl::Now();
+  std::filesystem::path recent_file = GenerateTestLogFilePath(now - absl::Hours(24));
+  std::filesystem::path old_file = GenerateTestLogFilePath(now - absl::Hours(24 * 14));
+
+  std::vector<std::filesystem::path> test_case({recent_file, old_file});
+  std::vector<std::filesystem::path> result = FilterOldLogFiles(test_case);
+  std::vector<std::filesystem::path> expected_result({old_file});
+  EXPECT_EQ(result, expected_result);
+}
+
+}  // namespace orbit_base
--- a/src/OrbitBase/ThreadUtilsLinux.cpp
+++ b/src/OrbitBase/ThreadUtilsLinux.cpp
@@ -51,4 +51,6 @@ void SetCurrentThreadName(const std::string& thread_name) {
   }
 }
 
+pid_t GetCurrentProcessId() { return getpid(); }
+
 }  // namespace orbit_base
--- a/src/OrbitBase/include/OrbitBase/Logging.h
+++ b/src/OrbitBase/include/OrbitBase/Logging.h
@@ -5,6 +5,9 @@
 #ifndef ORBIT_BASE_LOGGING_H_
 #define ORBIT_BASE_LOGGING_H_
 
+#include <absl/strings/str_format.h>
+#include <absl/synchronization/mutex.h>
+#include <absl/time/time.h>
 #include <stdlib.h>
 
 #include <cstdio>
@@ -16,9 +19,6 @@
 #include <Windows.h>
 #endif
 
-#include "absl/strings/str_format.h"
-#include "absl/synchronization/mutex.h"
-#include "absl/time/time.h"
 
 #ifdef __clang__
 #pragma clang diagnostic push
@@ -113,6 +113,9 @@ constexpr const char* kLogTimeFormat = "%Y-%m-%dT%H:%M:%E6S";
     absl::StrFormat(format, ##__VA_ARGS__)                                                     \
   }
 
+constexpr const char* kLogFileNameTimeFormat = "%Y_%m_%d_%H_%M_%S";
+constexpr absl::string_view kLogFileNameDelimiter = "Orbit-%s-%u.log";
+std::string GetLogFileName();
 extern std::ofstream log_file;
 void InitLogFile(const std::filesystem::path& path);
 void LogToFile(const std::string& message);
--- a/src/OrbitBase/include/OrbitBase/Logging.h
+++ b/src/OrbitBase/include/OrbitBase/Logging.h
@@ -19,6 +19,7 @@
 #include <Windows.h>
 #endif
 
+#include "OrbitBase/Result.h"
 
 #ifdef __clang__
 #pragma clang diagnostic push
@@ -113,9 +114,12 @@ constexpr const char* kLogTimeFormat = "%Y-%m-%dT%H:%M:%E6S";
     absl::StrFormat(format, ##__VA_ARGS__)                                                     \
   }
 
-constexpr const char* kLogFileNameTimeFormat = "%Y_%m_%d_%H_%M_%S";
-constexpr absl::string_view kLogFileNameDelimiter = "Orbit-%s-%u.log";
 std::string GetLogFileName();
+// This function tries to remove log files older than kLogFileLifetime even when an error is
+// returned. An error returns, for instance, if some log files to remove are used by other
+// applications. If so, both the file name and the error message will be recorded in the log file.
+ErrorMessageOr<void> TryRemoveOldLogFiles(const std::filesystem::path& log_dir);
+
 extern std::ofstream log_file;
 void InitLogFile(const std::filesystem::path& path);
 void LogToFile(const std::string& message);
--- a/src/OrbitBase/include/OrbitBase/ThreadUtils.h
+++ b/src/OrbitBase/include/OrbitBase/ThreadUtils.h
@@ -5,18 +5,25 @@
 #ifndef ORBIT_BASE_THREAD_UTILS_H_
 #define ORBIT_BASE_THREAD_UTILS_H_
 
+#ifdef _WIN32
+#include <stdint.h>
+#else
+#include <sys/types.h>
+#include <unistd.h>
+#endif
+
 #include <string>
 
 namespace orbit_base {
 
 #ifdef _WIN32
-#include <stdint.h>
 [[nodiscard]] uint32_t GetCurrentThreadId();
 [[nodiscard]] std::string GetThreadName(uint32_t tid);
+[[nodiscard]] uint32_t GetCurrentProcessId();
 #else
-#include <sys/types.h>
 [[nodiscard]] pid_t GetCurrentThreadId();
 [[nodiscard]] std::string GetThreadName(pid_t tid);
+[[nodiscard]] pid_t GetCurrentProcessId();
 #endif
 
 void SetCurrentThreadName(const std::string& thread_name);
--- a/src/OrbitCore/Path.cpp
+++ b/src/OrbitCore/Path.cpp
@@ -8,6 +8,7 @@
 #include <string>
 
 #include "CoreUtils.h"
+#include "OrbitBase/Logging.h"
 #include "absl/flags/flag.h"
 
 ABSL_FLAG(std::string, log_dir, "", "Set directory for the log.");
@@ -60,7 +61,7 @@ std::filesystem::path Path::CreateOrGetOrbitAppDataDir() {
   return path;
 }
 
-std::filesystem::path Path::GetLogFilePathAndCreateDir() {
+std::filesystem::path Path::CreateOrGetLogDir() {
   std::filesystem::path logs_dir;
   if (!absl::GetFlag(FLAGS_log_dir).empty()) {
     logs_dir = absl::GetFlag(FLAGS_log_dir);
@@ -68,5 +69,9 @@ std::filesystem::path Path::GetLogFilePathAndCreateDir() {
     logs_dir = Path::CreateOrGetOrbitAppDataDir() / "logs";
   }
   std::filesystem::create_directory(logs_dir);
-  return logs_dir / "Orbit.log";
+  return logs_dir;
+}
+
+std::filesystem::path Path::GetLogFilePath() {
+  return Path::CreateOrGetLogDir() / GetLogFileName();
 }
--- a/src/OrbitCore/Path.h
+++ b/src/OrbitCore/Path.h
@@ -14,5 +14,6 @@ namespace Path {
 [[nodiscard]] std::filesystem::path CreateOrGetCaptureDir();
 [[nodiscard]] std::filesystem::path CreateOrGetDumpDir();
 [[nodiscard]] std::filesystem::path CreateOrGetOrbitAppDataDir();
-[[nodiscard]] std::filesystem::path GetLogFilePathAndCreateDir();
+[[nodiscard]] std::filesystem::path CreateOrGetLogDir();
+[[nodiscard]] std::filesystem::path GetLogFilePath();
 };  // namespace Path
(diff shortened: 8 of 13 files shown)


src/OrbitBase/Logging.cpp
```diff
@@ -14,6 +14,24 @@
...
+const int kLogFileLifetimeMinutes = 60 * 24 * 7;  // one week
+
+void RemoveOldLogFiles(const std::filesystem::path& log_dir) {
+  auto now = std::filesystem::file_time_type::clock::now();
+  for (const auto& log_file_path : std::filesystem::recursive_directory_iterator(log_dir)) {
+    std::chrono::minutes file_lifetime_minutes = std::chrono::duration_cast<std::chrono::minutes>(
+        now - std::filesystem::last_write_time(log_file_path));
+    try {
+      if (file_lifetime_minutes.count() > kLogFileLifetimeMinutes) {
+        std::filesystem::remove(log_file_path);
```
dimitry-, line 26: std::filesystem functions always have noexcept couterpart that sets std::error_code. Please use them instead - we try to keep the code exception-free.

src/OrbitBase/Logging.cpp
```diff
@@ -14,6 +14,24 @@
 
 static absl::Mutex log_file_mutex(absl::kConstInit);
 std::ofstream log_file;
+const int kLogFileLifetimeMinutes = 60 * 24 * 7;  // one week
+
+void RemoveOldLogFiles(const std::filesystem::path& log_dir) {
```
dimitry-, line 19: Maybe return ErrorMessageOr<void> here and have caller desided what to do in case removing old files fails.

Or/and(?) rename this TryRemoveOldLogFiles

src/OrbitCore/Path.h
```diff
@@ -14,5 +14,6 @@ namespace Path {
 [[nodiscard]] std::filesystem::path CreateOrGetCaptureDir();
 [[nodiscard]] std::filesystem::path CreateOrGetDumpDir();
 [[nodiscard]] std::filesystem::path CreateOrGetOrbitAppDataDir();
-[[nodiscard]] std::filesystem::path GetLogFilePathAndCreateDir();
+[[nodiscard]] std::filesystem::path CreateOrGetLogDir();
```
dimitry-, line 17: we probably want to test this function as well.

src/OrbitQt/main.cpp
```diff
@@ -266,7 +266,8 @@ int main(int argc, char* argv[]) {
   absl::SetFlagsUsageConfig(absl::FlagsUsageConfig{{}, {}, {}, &orbit_core::GetBuildReport, {}});
   absl::ParseCommandLine(argc, argv);
 
-  InitLogFile(Path::GetLogFilePathAndCreateDir());
+  RemoveOldLogFiles(Path::CreateOrGetLogDir());
```
dimitry-, line 269: Is this the only place where CreateOrGetLogDir is used? if so maybe remove it from Path.h and call in RemoveOldLogFiles implementation?

(if you need this parameter for a test - you can probably make it optional, so that by default the implemenation calls CreateOrGetLogDir())

src/OrbitBase/Logging.cpp
```diff
@@ -14,6 +14,24 @@
...
+  auto now = std::filesystem::file_time_type::clock::now();
+  for (const auto& log_file_path : std::filesystem::recursive_directory_iterator(log_dir)) {
+    std::chrono::minutes file_lifetime_minutes = std::chrono::duration_cast<std::chrono::minutes>(
+        now - std::filesystem::last_write_time(log_file_path));
```
dimitry-, line 23: hmm... another alternative would be to parse filename in order to extract the date, and remove files based on name rather than last_modified time. And this also is easily testable since you can create such a file and then remove it.

src/OrbitBase/Logging.cpp
vickyliu-go4it: Changed accordingly. Thanks for pointing out this!

src/OrbitCore/Path.h
vickyliu-go4it: Updated the test file. Thanks!

src/OrbitBase/Logging.cpp
vickyliu-go4it: Thanks for the suggestion. Changed this method as `ErrorMessageOr<void> TryRemoveOldLogFiles(const std::filesystem::path& log_dir)`.

src/OrbitQt/main.cpp
vickyliu-go4it: Thanks for the suggestion. `CreateOrGetLogDir` is also used in `GetLogFilePath`. I split the original `Path::GetLogFilePathAndCreateDir()` into two parts, i.e., `Path::CreateOrGetLogDir` and `Path::GetLogFilePath`. Therefore, I prefer to keep this method in `Path.h`.

src/OrbitBase/Logging.cpp
vickyliu-go4it: Cool suggestion! It is so easy to test! I changed to extract the time_point from filename. Thanks!

src/OrbitCore/Path.cpp
```diff
@@ -4,11 +4,20 @@
 
 #include "Path.h"
 
+#ifdef _WIN32
+#define getpid GetCurrentProcessId
```
dimitry-, line 8: this is not a good practice, Maybe orbit_base::GetCurrentProcessId() which is cross-platform and calls correct implementation. Like GetCurrentThreadId in OrbitBase ThreadUtils.h

src/OrbitBase/include/OrbitBase/Logging.h
```diff
@@ -114,6 +117,8 @@ constexpr const char* kLogTimeFormat = "%Y-%m-%dT%H:%M:%E6S";
   }
 
 extern std::ofstream log_file;
+std::optional<int> GetLogFileLifetimeMinutes(std::string log_file_name);
```
dimitry-, line 120: It looks like now we have knowledge about logfile format distributed between OrbitCore and OribtBase. I suggest introducing GetLogFileName function to this file and calling it from OrbitCore.

Alternatively you could move both of these functions to OrbitCore.

src/OrbitBase/Logging.cpp
```diff
@@ -5,15 +5,65 @@
...
+  std::string tm_str =
+      log_file_name.substr(first_delimiter_pos + 1, last_delimiter_pos - first_delimiter_pos - 1);
+  if (tm_str.empty()) return std::nullopt;
+
+  std::istringstream istr(tm_str);
+  int year, month, day, hour, minute, second;
+  char delimiter;
+  // The timestamp in the log file name follows a format of "%Y_%m_%d_%H_%M_%S".
+  if (istr >> year >> delimiter >> month >> delimiter >> day >> delimiter >> hour >> delimiter >>
+      minute >> delimiter >> second) {
```
dimitry-, line 33: Use early return

if (! ... ) return std::nullopt;

...
In that case there is one less nested block.

src/OrbitQt/main.cpp
```diff
@@ -303,7 +307,7 @@ int main(int argc, char* argv[]) {
   const std::string handler_path =
       QDir(QCoreApplication::applicationDirPath()).absoluteFilePath(handler_name).toStdString();
   const std::string crash_server_url = CrashServerOptions::GetUrl();
-  const std::vector<std::string> attachments = {Path::GetLogFilePathAndCreateDir().string()};
+  const std::vector<std::string> attachments = {Path::GetLogFilePath().string()};
 
   CrashHandler crash_handler(dump_path, handler_path, crash_server_url, attachments);
 #endif  // ORBIT_CRASH_HANDLING
```
dimitry-, line 313: Add a test for TryRemoveOldLogFiles?

src/OrbitBase/Logging.cpp
```diff
@@ -5,15 +5,65 @@
...
+    return log_file_lifetime_minutes.count();
+  } else {
+    return std::nullopt;
+  }
+}
+
+ErrorMessageOr<void> TryRemoveOldLogFiles(const std::filesystem::path& log_dir) {
+  std::string error_message;
+  for (const auto& log_file_path : std::filesystem::recursive_directory_iterator(log_dir)) {
+    std::string log_file_name = log_file_path.path().filename().u8string();
```
dimitry-, line 50: does u8string yield result different from just string()?

src/OrbitQt/main.cpp
```diff
@@ -266,8 +266,12 @@ int main(int argc, char* argv[]) {
   absl::SetFlagsUsageConfig(absl::FlagsUsageConfig{{}, {}, {}, &orbit_core::GetBuildReport, {}});
   absl::ParseCommandLine(argc, argv);
 
-  InitLogFile(Path::GetLogFilePathAndCreateDir());
+  InitLogFile(Path::GetLogFilePath());
   LOG("You are running Orbit Profiler version %s", orbit_core::GetVersion());
+  ErrorMessageOr<void> remove_old_log_result = TryRemoveOldLogFiles(Path::CreateOrGetLogDir());
+  if (remove_old_log_result.has_error()) {
+    LOG("Fail to remove some old log files:\n%s", remove_old_log_result.error().message());
```
dimitry-, line 273: nit: maybe "Warning: Unable to remove some old log files: "?

src/OrbitCore/Path.cpp
```diff
@@ -60,13 +69,22 @@ std::filesystem::path Path::CreateOrGetOrbitAppDataDir() {
...
   std::filesystem::create_directory(logs_dir);
-  return logs_dir / "Orbit.log";
+  return logs_dir;
+}
+
+std::filesystem::path Path::GetLogFilePath() {
+  std::filesystem::path logs_dir = CreateOrGetLogDir();
+  time_t timestamp = std::chrono::system_clock::to_time_t(std::chrono::system_clock::now());
+  std::string filename =
+      absl::StrCat("Orbit", kLogFileNameDelimiter, orbit_core::FormatTime(timestamp),
```
antonrohr, line 87: Did you consider `absl::FormatTime()` instead of `orbit_core::FormatTime()`? The orbit_core thing looks like some fairly old code, I strongly feel like we should not reuse that if possible. 
If you use absl, there is also the function `absl::ParseTime()` which might simplify the `GetLogFileLifetimeMinutes`. I write a comment there as well

src/OrbitBase/Logging.cpp
```diff
@@ -5,15 +5,65 @@
...
+  std::istringstream istr(tm_str);
+  int year, month, day, hour, minute, second;
+  char delimiter;
+  // The timestamp in the log file name follows a format of "%Y_%m_%d_%H_%M_%S".
+  if (istr >> year >> delimiter >> month >> delimiter >> day >> delimiter >> hour >> delimiter >>
+      minute >> delimiter >> second) {
+    // std::tm will set the year on basis of 1900 and month on basis of January.
+    std::tm log_file_tm = {second, minute, hour, day, month - 1, year - 1900};
+    std::chrono::system_clock::time_point log_file_tp =
+        std::chrono::system_clock::from_time_t(std::mktime(&log_file_tm));
```
antonrohr, line 37: This all looks too complicated and too custom. See my other comment, maybe use `absl::FormatTime()` and `absl::ParseTime()`.

src/OrbitBase/Logging.cpp
```diff
@@ -5,15 +5,65 @@
...
 #include "absl/debugging/symbolize.h"
+#include "absl/strings/str_cat.h"
 #include "absl/strings/str_format.h"
 #include "absl/synchronization/mutex.h"
 
 static absl::Mutex log_file_mutex(absl::kConstInit);
 std::ofstream log_file;
+const int kLogFileLifetimeMinutes = 60 * 24 * 7;  // one week
+
+std::optional<int> GetLogFileLifetimeMinutes(std::string log_file_name) {
```
antonrohr, line 21:
```suggestion
ErrorMessageOr<int> GetLogFileLifetimeMinutes(std::string log_file_name) {
```
I think it would valuable if we see in the logs if this filename parsing fails. Currently if the parsing fails, its just silently omitted.

antonrohr (commented):
Looks good in general, some comments about the time formatting / parsing, that might simplify the code quite a bit. 

As dimitry pointed out, a test for `TryRemoveOldLogFiles` would be nice. Maybe you can split some things in functions that are easier to test. I am thinking something like `std::vector<std::filesystem::path> ListFiles(dir)`, `std::vector<std::filesystem::path> FilterOldFiles(std::vector<std::filesystem::path>)` and `removeFiles(std::vector<std::filesystem::path>)`. Then at least `FilterOldFiles` is easily testable. (This is just an idea, maybe there is a better separation and maybe you want to return `ErrorMessageOr` for these 3, or some of them).

src/OrbitCore/Path.cpp
vickyliu-go4it: Done. Thanks for the suggestion!

src/OrbitBase/include/OrbitBase/Logging.h
vickyliu-go4it: Moved bot `CreateOrGetLogDir` and `GetLogPath` to OrbitBase. Thanks for the suggestion!

src/OrbitBase/Logging.cpp
vickyliu-go4it: Thanks for the suggestion! Changed to use `absl::FormatTime` and `absl::ParseTime` to simpliy the code here.

src/OrbitQt/main.cpp
vickyliu-go4it: I split `TryRemoveOldLogFiles` into three methods, i.e., `ListFiles`, `GetLogFileLifetime`, `FilterOldLogFiles` and `RemoveFiles`;  and added tests for `GetLogFileLifetime` and `FilterOldLogFiles`. Thanks for the suggestion!

src/OrbitBase/Logging.cpp
vickyliu-go4it: Ah, I used `u8string()` when debugging for some issues and forgot to change back. Sorry about that. `u8string()` makes no difference with `string()`. I changed to use `string()`.

src/OrbitQt/main.cpp
vickyliu-go4it: Done! Thanks!

src/OrbitCore/Path.cpp
vickyliu-go4it: Thanks for telling me about this! I changed to use `absl::FormatTime` and `absl::ParseTime` to simplify the code.

src/OrbitBase/Logging.cpp
vickyliu-go4it: Done. Thanks!

src/OrbitBase/Logging.cpp
vickyliu-go4it: Thanks for the suggestion! The parsing failures are also added to the log files now.

src/OrbitBase/Logging.cpp
```diff
@@ -7,56 +7,107 @@
...
+  for (const auto& file_path : std::filesystem::recursive_directory_iterator(log_dir)) {
+    if (file_path.is_regular_file()) {
+      files_in_dir.push_back(file_path.path());
+    }
+  }
+  return files_in_dir;
+}
+
+ErrorMessageOr<absl::Duration> GetLogFileLifetime(const std::string& log_file_name) {
   unsigned first_delimiter_pos = log_file_name.find_first_of(kLogFileNameDelimiter);
```
antonrohr, line 67: From https://www.cplusplus.com/reference/string/string/find_first_of/
> The position of the first character that matches.
If no matches are found, the function returns string::npos.

Have an early return with ErrorMessage here, when `kLogFileNameDelimiter` is not found. Or in other cases where there is something wrong. 

The longer I think about this. Maybe you can make this functions even stronger. The log file name should always have this format: `Orbit-%time%-%pid%.log`. That means the time always starts at character 6. Also the length of the time is fixed.. right? so no need to calculate the length of the time

src/OrbitBase/LoggingTest.cpp
```diff
@@ -0,0 +1,55 @@
...
+  std::filesystem::path path = GetLogFilePath().parent_path();
+  LOG("Testing existence of \"%s\"", path.string());
+  EXPECT_TRUE(std::filesystem::is_directory(path));
+}
+
+TEST(Logging, GetLogFileLifetime) {
+  std::vector<std::string> test_cases({
+      "Orbit.log",
+      "Orbit-2021_01_31T00:00:00-7188.log",
+      "Orbit-2021_01_31_00_00_00-7188.log",
```
antonrohr, line 22: Maybe more weird test examples:
```
Orbitfoobar-2021_01_31_00_00_00-.log
sfsdf-.log
```

src/OrbitBase/include/OrbitBase/Logging.h
```diff
@@ -117,8 +121,20 @@ static const std::string kLogFileNameDelimiter = "-";
...
+[[nodiscard]] std::filesystem::path CreateOrGetLogDir();
+[[nodiscard]] std::filesystem::path GetLogFilePath(const absl::Time& timestamp = absl::Now(),
+                                                   const uint32_t& pid = GetCurrentProcessId());
+}  // namespace orbit_base
+
+[[nodiscard]] std::vector<std::filesystem::path> ListFiles(const std::filesystem::path& log_dir);
+ErrorMessageOr<absl::Duration> GetLogFileLifetime(const std::string& log_file_name);
+[[nodiscard]] std::vector<std::filesystem::path> FilterOldLogFiles(
+    const std::vector<std::filesystem::path>& log_file_paths);
+ErrorMessageOr<void> RemoveFiles(const std::vector<std::filesystem::path>& log_file_paths);
```
antonrohr, line 136: Maybe a comment here, that this function will remove files even when an error is returned. 

As a reader, I might expect this function to not have delete any files when an error is returned

src/OrbitBase/Logging.cpp
```diff
@@ -7,56 +7,107 @@
...
+  std::filesystem::create_directory(logs_dir);
+  return logs_dir;
+}
+
+std::filesystem::path GetLogFilePath(const absl::Time& timestamp, const uint32_t& pid) {
+  std::filesystem::path logs_dir = CreateOrGetLogDir();
+  std::string timestamp_string =
+      absl::FormatTime(kLogFileNameTimeFormat, timestamp, absl::LocalTimeZone());
+  std::string filename = absl::StrCat("Orbit", kLogFileNameDelimiter, timestamp_string,
+                                      kLogFileNameDelimiter, pid, ".log");
```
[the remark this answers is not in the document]
vickyliu-go4it: Changed accordingly. Thanks for the suggestion!

src/OrbitBase/Logging.cpp
vickyliu-go4it: Make sense! I changed to use constants `kTimestampStartPos` and `kTimestampStringLength` to indicate the starting position and the timestamp string length. Thanks!

src/OrbitBase/LoggingTest.cpp
vickyliu-go4it: Updated accordingly. Thanks!

src/OrbitBase/include/OrbitBase/Logging.h
vickyliu-go4it: Added a comments here. Thanks for the suggestion.

src/OrbitBase/include/OrbitBase/Logging.h
```diff
@@ -114,6 +118,15 @@ constexpr const char* kLogTimeFormat = "%Y-%m-%dT%H:%M:%E6S";
...
 
 extern std::ofstream log_file;
+[[nodiscard]] std::vector<std::filesystem::path> ListFiles(const std::filesystem::path& log_dir);
+ErrorMessageOr<absl::Duration> GetLogFileLifetime(const std::string& log_file_name);
+[[nodiscard]] std::vector<std::filesystem::path> FilterOldLogFiles(
+    const std::vector<std::filesystem::path>& log_file_paths);
+ErrorMessageOr<void> RemoveFiles(const std::vector<std::filesystem::path>& log_file_paths);
+// This function tries to remove log files older than kLogFileLifetime even when an error is
+// returned. An error returns, for instance, if some log files to remove are used by other
+// applications. If so, both the file name and the error message will be recored in the Log file.
```
antonrohr, line 128: nit: 
```suggestion
// applications. If so, both the file name and the error message will be recorded in the Log file.
```

src/OrbitBase/include/OrbitBase/Logging.h
antonrohr: nit: The same comment also for `RemoveFiles` (similar issue)

src/OrbitBase/Logging.cpp
antonrohr: Looks great :)

antonrohr (approved):
I have 2 other nitpicks, otherwise LGTM 
Thank you for this Vicky!

src/OrbitCore/Path.cpp
```diff
@@ -60,13 +69,31 @@ std::filesystem::path Path::CreateOrGetOrbitAppDataDir() {
...
+  return logs_dir;
+}
+
+std::filesystem::path Path::GetLogFilePath() {
+  std::filesystem::path logs_dir = CreateOrGetLogDir();
+  std::string timestamp_string =
+      absl::FormatTime(kLogFileNameTimeFormat, absl::Now(), absl::LocalTimeZone());
+  std::string filename =
+      absl::StrFormat(kLogFileNameDelimiter, timestamp_string, Path::GetCurrentProcessId());
+  return logs_dir / filename;
```
dimitry-, line 98: return CreateOrGetLogDir() / GetLogFilename();
GetLogFilename function will be in OrbitBase.

src/OrbitCore/Path.cpp
```diff
@@ -60,13 +69,31 @@ std::filesystem::path Path::CreateOrGetOrbitAppDataDir() {
...
-  return logs_dir / "Orbit.log";
+  return logs_dir;
+}
+
+std::filesystem::path Path::GetLogFilePath() {
+  std::filesystem::path logs_dir = CreateOrGetLogDir();
+  std::string timestamp_string =
+      absl::FormatTime(kLogFileNameTimeFormat, absl::Now(), absl::LocalTimeZone());
+  std::string filename =
+      absl::StrFormat(kLogFileNameDelimiter, timestamp_string, Path::GetCurrentProcessId());
```
dimitry-, line 97: extract this to GetLogFilename function and put it to OrbitBase.

src/OrbitBase/include/OrbitBase/Logging.h
```diff
@@ -114,6 +118,15 @@ constexpr const char* kLogTimeFormat = "%Y-%m-%dT%H:%M:%E6S";
   }
 
 extern std::ofstream log_file;
+[[nodiscard]] std::vector<std::filesystem::path> ListFiles(const std::filesystem::path& log_dir);
+ErrorMessageOr<absl::Duration> GetLogFileLifetime(const std::string& log_file_name);
+[[nodiscard]] std::vector<std::filesystem::path> FilterOldLogFiles(
+    const std::vector<std::filesystem::path>& log_file_paths);
+ErrorMessageOr<void> RemoveFiles(const std::vector<std::filesystem::path>& log_file_paths);
```
dimitry-, line 125: You can move them to src/OrbitBase/ and create separate header file. This will make them private to OrbitBase component but you still can use it from the test.

src/OrbitBase/ThreadUtilsLinux.cpp
```diff
@@ -51,4 +51,10 @@ void SetCurrentThreadName(const std::string& thread_name) {
   }
 }
 
+pid_t GetCurrentProccessId() {
+  pid_t current_pid = getpid();
+  CHECK(current_pid);
```
dimitry-, line 56: no need to check - this method always succeeds according to man page.

src/OrbitBase/ThreadUtilsLinux.cpp
```diff
@@ -51,4 +51,10 @@ void SetCurrentThreadName(const std::string& thread_name) {
   }
 }
 
+pid_t GetCurrentProccessId() {
```
dimitry-, line 54: you have extra 'c' in Process.

src/OrbitCore/Path.cpp
vickyliu-go4it: Done! Thanks so much for the suggestion!

src/OrbitCore/Path.cpp
vickyliu-go4it: Changed accordingly. Thanks!

src/OrbitBase/include/OrbitBase/Logging.h
vickyliu-go4it: Thanks so much for telling me this! I moved these functions to `LoggingUtils.h` and `LoggingUtils.cpp`.

src/OrbitBase/ThreadUtilsLinux.cpp
vickyliu-go4it: Changed accordingly. Thanks!

src/OrbitBase/ThreadUtilsLinux.cpp
vickyliu-go4it: I should be more careful!!! Thanks so much for pointing out this.

src/OrbitBase/LoggingUtilsTest.cpp
```diff
@@ -0,0 +1,60 @@
...
+#include <gmock/gmock.h>
+#include <gtest/gtest.h>
+
+#include "LoggingUtils.h"
+#include "absl/strings/str_format.h"
+#include "absl/time/clock.h"
+
+namespace {
+std::filesystem::path CreateTestLogFilePath(const absl::Time& timestamp) {
+  std::filesystem::path test_log_dir = "C:/OrbitAppDataDir/logs";
```
dimitry-, line 14: does it work on linux? Maybe use https://en.cppreference.com/w/cpp/filesystem/temp_directory_path

vickyliu-go4it: The name of this function might be misleading. It generates a path used for testing and will not create a real directory in the file system. It is a good idea to use `temp_directory_path`, but  `temp_directory_path` returns an empty path on error and we still need to generate a test path in such a case.

Maybe I should rename this function as `GenerateTestLogFilePath` or `GetTestLogFilePath`?

src/OrbitBase/LoggingUtils.h
```diff
@@ -0,0 +1,26 @@
...
+// found in the LICENSE file.
+
+#ifndef ORBIT_BASE_LOGGING_UTILS_H_
+#define ORBIT_BASE_LOGGING_UTILS_H_
+
+#include <filesystem>
+
+#include "OrbitBase/Result.h"
+#include "absl/strings/str_format.h"
+#include "absl/time/time.h"
```
dimitry-, line 12: nit: we use <> style includes for everything external to Orbit

src/OrbitBase/LoggingUtilsTest.cpp
```diff
@@ -0,0 +1,61 @@
...
+}  // namespace
+
+namespace orbit_base {
+
+TEST(LoggingUtils, GetLogFileLifetime) {
+  std::vector<std::string> test_cases({
+      "sfsdf-.log ",
+      "Orbitfoobar-2021_01_31T00_00_00-.log",
+      "Orbit-2021_01_31_00_00_00-7188.log",
+  });
```
dimitry-, line 30: nitpick: Use variables instead of vector

const std::string kFilenameInvalidNoTimestamp = "sfsdf-.log ";
kFilenameInvalidValidTimestampNoPid
kFilenameValid

src/OrbitBase/LoggingUtilsTest.cpp
dimitry-: ok, I see..

dimitry- (approved):
Otherwise LGTM

src/OrbitBase/include/OrbitBase/ThreadUtils.h
```diff
@@ -13,10 +13,13 @@ namespace orbit_base {
 #include <stdint.h>
 [[nodiscard]] uint32_t GetCurrentThreadId();
 [[nodiscard]] std::string GetThreadName(uint32_t tid);
+[[nodiscard]] uint32_t GetCurrentProcessId();
 #else
 #include <sys/types.h>
+#include <unistd.h>
```
dimitry-, line 19: It looks like this include causes build failures... can you move includes outside of namespace declaration, it should fix the build.

src/OrbitBase/include/OrbitBase/ThreadUtils.h
```diff
@@ -6,17 +6,23 @@
...
 #include <string>
+#ifdef _WIN32
+#else
+#include <unistd.h>
+#endif
 
 namespace orbit_base {
 
 #ifdef _WIN32
 #include <stdint.h>
```
dimitry-, line 17: this include

src/OrbitBase/include/OrbitBase/ThreadUtils.h
```diff
@@ -6,17 +6,23 @@
...
 [[nodiscard]] uint32_t GetCurrentThreadId();
 [[nodiscard]] std::string GetThreadName(uint32_t tid);
+[[nodiscard]] uint32_t GetCurrentProcessId();
 #else
 #include <sys/types.h>
```
dimitry-, line 22: and this include as well.

src/OrbitBase/LoggingUtils.h
vickyliu-go4it: Changed for all the files I touched. Thanks!

src/OrbitBase/LoggingUtilsTest.cpp
vickyliu-go4it: Done. Thanks for the suggestion!

src/OrbitBase/include/OrbitBase/ThreadUtils.h
vickyliu-go4it: Done. Thanks so much for pointing out this!

src/OrbitBase/include/OrbitBase/ThreadUtils.h
vickyliu-go4it: Done. Thanks!

PR #1756, opened 2021-01-29
Set pid of orbit_gprc_protos::ThreadStateSlice

dpallotti:
Compared bandwidth usage before and after this change with a simple capture on
Trata: thread state enabled, no dynamic instrumentation.
The average bytes per event goes from 23.1 to 25.7, which is quite an increase.

---

Given what I wrote in the commit message about the increase in data usage,
I'm not sure I want to merge this change, but I'm creating it al least for discussion.

Context: Some protos in `orbit_grpc_protos` had a `pid` field not set. In coordination
with Dimitry, as we are moving towards a stable format, it seems reasonable to have
them set when it's very cheap to do so.
This is the last remaining one but also the one for which such a change is not exactly
free.

### DIFF
--- a/src/LinuxTracing/SwitchesStatesNamesVisitor.cpp
+++ b/src/LinuxTracing/SwitchesStatesNamesVisitor.cpp
@@ -82,7 +82,7 @@ void SwitchesStatesNamesVisitor::ProcessInitialState(uint64_t timestamp_ns, pid_
     ERROR("Parsing thread state char '%c' for tid %d", state_char, tid);
     return;
   }
-  state_manager_.OnInitialState(timestamp_ns, tid, initial_state.value());
+  state_manager_.OnInitialState(timestamp_ns, thread_state_pid_filter_, tid, initial_state.value());
 }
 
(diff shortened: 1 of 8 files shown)


src/LinuxTracing/SwitchesStatesNamesVisitor.cpp
```diff
@@ -168,8 +168,8 @@ void SwitchesStatesNamesVisitor::visit(SchedWakeupPerfEvent* event) {
     return;
   }
 
-  std::optional<ThreadStateSlice> state_slice =
-      state_manager_.OnSchedWakeup(event->GetTimestamp(), event->GetWokenTid());
+  std::optional<ThreadStateSlice> state_slice = state_manager_.OnSchedWakeup(
+      event->GetTimestamp(), thread_state_pid_filter_, event->GetWokenTid());
```
dimitry-, line 172: maybe rename thread_state_pid_filter_ to active_pid_/current_pid_?

dimitry- (approved):

dpallotti:
I'm closing this old change, never merged because of the small increase in bandwidth usage only for the purpose of consistency with the other protos. Refer to http://b/207344825

PR #1757, opened 2021-01-29
Attach to a process via ptrace.

danielfenner:
Attach to and stop all threads of a process. Also detach again.

Test: Unit Test.

### DIFF
--- a/src/UserSpaceInstrumentation/Attach.cpp
+++ b/src/UserSpaceInstrumentation/Attach.cpp
@@ -4,6 +4,7 @@
 
 #include "UserSpaceInstrumentation/Attach.h"
 
+#include <absl/container/flat_hash_set.h>
 #include <absl/strings/str_format.h>
 #include <errno.h>
 #include <stdint.h>
@@ -13,6 +14,7 @@
 #include <chrono>
 #include <memory>
 #include <set>
+
 #include <string>
 #include <thread>
 #include <vector>
@@ -41,8 +43,8 @@ namespace {
         absl::StrFormat("PTRACE_ATTACH failed for %d: \"%s\"", tid, SafeStrerror(errno)));
   }
   // Wait for the traced thread to stop. Timeout after one second.
-  int32_t max_attempts = 1000;
-  while (true) {
+  constexpr int kMaxAttempts = 1000;
+  for (int i = 0; i < kMaxAttempts; i++) {
     int stat_val = 0;
     const int waitpid_result = waitpid(tid, &stat_val, WNOHANG);
     if (waitpid_result == -1) {
@@ -62,25 +64,21 @@ namespace {
           "Wait for thread to get traced yielded unexpected result for tid %d: %d", tid, stat_val));
     }
     std::this_thread::sleep_for(std::chrono::milliseconds(1));
-    if (--max_attempts == 0) {
-      ptrace(PTRACE_DETACH, tid, nullptr, nullptr);
-      return ErrorMessage(
-          absl::StrFormat("Waiting for the traced thread %d to stop timed out.", tid));
-    }
   }
 
-  UNREACHABLE();
+  ptrace(PTRACE_DETACH, tid, nullptr, nullptr);
+  return ErrorMessage(absl::StrFormat("Waiting for the traced thread %d to stop timed out.", tid));
 }
 
 }  // namespace
 
 [[nodiscard]] ErrorMessageOr<void> AttachAndStopProcess(pid_t pid) {
   auto process_tids = GetTidsOfProcess(pid);
-  std::set<pid_t> halted_tids;
+  absl::flat_hash_set<pid_t> halted_tids;
   // Note that the process is still running - it can spawn and end threads at this point.
-  while (true) {
+  while (process_tids.size() != halted_tids.size()) {
     for (const auto tid : process_tids) {
-      if (halted_tids.count(tid) == 1) {
+      if (halted_tids.contains(tid)) {
         continue;
       }
       OUTCOME_TRY(result, AttachAndStopThread(tid));
@@ -88,11 +86,7 @@ namespace {
         halted_tids.insert(tid);
       }
     }
-    // Update the tids; if there are new ones run another round and stop them.
     process_tids = GetTidsOfProcess(pid);
-    if (process_tids.size() == halted_tids.size()) {
-      break;
-    }
   }
   return outcome::success();
 }
--- a/src/UserSpaceInstrumentation/Attach.cpp
+++ b/src/UserSpaceInstrumentation/Attach.cpp
@@ -14,7 +14,6 @@
 #include <chrono>
 #include <memory>
 #include <set>
-
 #include <string>
 #include <thread>
 #include <vector>
@@ -72,7 +71,7 @@ namespace {
 
 }  // namespace
 
-[[nodiscard]] ErrorMessageOr<void> AttachAndStopProcess(pid_t pid) {
+ErrorMessageOr<void> AttachAndStopProcess(pid_t pid) {
   auto process_tids = GetTidsOfProcess(pid);
   absl::flat_hash_set<pid_t> halted_tids;
   // Note that the process is still running - it can spawn and end threads at this point.
@@ -81,8 +80,20 @@ namespace {
       if (halted_tids.contains(tid)) {
         continue;
       }
-      OUTCOME_TRY(result, AttachAndStopThread(tid));
-      if (result) {
+      auto result = AttachAndStopThread(tid);
+      if (!result) {
+        // Try to detach and return an error.
+        for (const auto t : halted_tids) {
+          if (ptrace(PTRACE_DETACH, tid, nullptr, nullptr) == -1) {
+            return ErrorMessage(absl::StrFormat(
+                "Unable to attach to thread: %d. Also unable to clean up. We are still "
+                "attached to thread: %d",
+                tid, t));
+          }
+        }
+        return ErrorMessage(absl::StrFormat("Unable to attach to thread tid: %d", tid));
+      }
+      if (result.value()) {
         halted_tids.insert(tid);
       }
     }
@@ -91,7 +102,7 @@ namespace {
   return outcome::success();
 }
 
-[[nodiscard]] ErrorMessageOr<void> DetachAndContinueProcess(pid_t pid) {
+ErrorMessageOr<void> DetachAndContinueProcess(pid_t pid) {
   auto tids = GetTidsOfProcess(pid);
   for (auto tid : tids) {
     if (ptrace(PTRACE_DETACH, tid, nullptr, nullptr) == -1) {
--- a/src/UserSpaceInstrumentation/Attach.cpp
+++ b/src/UserSpaceInstrumentation/Attach.cpp
@@ -97,11 +97,15 @@ namespace {
   return outcome::success();
 }
 
-void DetachAndContinueProcess(pid_t pid) {
+[[nodiscard]] ErrorMessageOr<void> DetachAndContinueProcess(pid_t pid) {
   auto tids = GetTidsOfProcess(pid);
   for (auto tid : tids) {
-    ptrace(PTRACE_DETACH, tid, nullptr, nullptr);
+    if (ptrace(PTRACE_DETACH, tid, nullptr, nullptr) == -1) {
+      return ErrorMessage(absl::StrFormat("Error while detaching from thread %d: \"%s\"", tid,
+                                          SafeStrerror(errno)));
+    }
   }
+  return outcome::success();
 }
 
 }  // namespace orbit_user_space_instrumentation
--- a/src/UserSpaceInstrumentation/Attach.cpp
+++ b/src/UserSpaceInstrumentation/Attach.cpp
@@ -19,6 +19,7 @@
 
 #include "OrbitBase/GetProcessIds.h"
 #include "OrbitBase/Logging.h"
+#include "OrbitBase/SafeStrerror.h"
 
 namespace orbit_user_space_instrumentation {
 
@@ -35,41 +36,40 @@ namespace {
     // can be looked up in the function 'ptrace_attach' in 'ptrace.c' in the kernel sources.
     if (errno == ESRCH || errno == EPERM) {
       return false;
-    } else {
-      return ErrorMessage(
-          absl::StrFormat("PTRACE_ATTACH failed for \"%d\": \"%s\"", tid, strerror(errno)));
     }
+    return ErrorMessage(
+        absl::StrFormat("PTRACE_ATTACH failed for %d: \"%s\"", tid, SafeStrerror(errno)));
   }
   // Wait for the traced thread to stop. Timeout after one second.
-  int32_t timeout_ms = 1000;
+  int32_t max_attempts = 1000;
   while (true) {
     int stat_val = 0;
-    const int result_waitpid = waitpid(tid, &stat_val, WNOHANG);
-    if (result_waitpid == -1) {
-      return ErrorMessage(absl::StrFormat(
-          "Wait for thread to get traced failed for tid \"%d\": \"%s\"", tid, strerror(errno)));
+    const int waitpid_result = waitpid(tid, &stat_val, WNOHANG);
+    if (waitpid_result == -1) {
+      return ErrorMessage(absl::StrFormat("Wait for thread to get traced failed for tid %d: \"%s\"",
+                                          tid, SafeStrerror(errno)));
     }
-    if (result_waitpid > 0) {
+    if (waitpid_result > 0) {
       // Occasionally the thread is active during PTRACE_ATTACH but terminates before it gets
       // descheduled. So waitpid returns on exit of the thread instead of the expected stop.
       if (WIFEXITED(stat_val)) {
         return false;
-      } else if (WIFSTOPPED(stat_val)) {
-        return true;
-      } else {
-        return ErrorMessage(absl::StrFormat(
-            "Wait for thread to get traced yielded unexpected result for tid \"%d\": \"%d\"", tid,
-            stat_val));
       }
+      if (WIFSTOPPED(stat_val)) {
+        return true;
+      }
+      return ErrorMessage(absl::StrFormat(
+          "Wait for thread to get traced yielded unexpected result for tid %d: %d", tid, stat_val));
     }
     std::this_thread::sleep_for(std::chrono::milliseconds(1));
-    if (--timeout_ms == 0) {
+    if (--max_attempts == 0) {
       ptrace(PTRACE_DETACH, tid, nullptr, nullptr);
       return ErrorMessage(
-          absl::StrFormat("Waiting for the traced thread \"%d\" to stop timed out.", tid));
+          absl::StrFormat("Waiting for the traced thread %d to stop timed out.", tid));
     }
   }
-  return true;
+
+  UNREACHABLE();
 }
 
 }  // namespace
--- a/src/UserSpaceInstrumentation/Attach.cpp
+++ b/src/UserSpaceInstrumentation/Attach.cpp
@@ -1,0 +1,107 @@
+// Copyright (c) 2021 The Orbit Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#include "UserSpaceInstrumentation/Attach.h"
+
+#include <absl/strings/str_format.h>
+#include <errno.h>
+#include <stdint.h>
+#include <sys/ptrace.h>
+#include <sys/wait.h>
+
+#include <chrono>
+#include <memory>
+#include <set>
+#include <string>
+#include <thread>
+#include <vector>
+
+#include "OrbitBase/GetProcessIds.h"
+#include "OrbitBase/Logging.h"
+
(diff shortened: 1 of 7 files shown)


src/UserSpaceInstrumentation/Attach.cpp
```diff
@@ -0,0 +1,107 @@
...
+// Attaches to the thread using ptrace.
+// Returns true if the thread was halted and false if the thread did not exist anymore.
+[[nodiscard]] ErrorMessageOr<bool> AttachAndStopThread(pid_t tid) {
+  if (ptrace(PTRACE_ATTACH, tid, nullptr, nullptr) == -1) {
+    // If tid has ended already we get ESRCH; if the thread was in 'exit state' we get EPERM.
+    // There are a bunch of other (non-relevant) cases. I haven't found documentation on this but it
+    // can be looked up in the function 'ptrace_attach' in 'ptrace.c' in the kernel sources.
+    if (errno == ESRCH || errno == EPERM) {
+      return false;
+    } else {
```
dimitry-, line 38: drop else?

src/UserSpaceInstrumentation/Attach.cpp
```diff
@@ -0,0 +1,107 @@
...
+      return ErrorMessage(
+          absl::StrFormat("PTRACE_ATTACH failed for \"%d\": \"%s\"", tid, strerror(errno)));
```
dimitry-, line 40: Use SafeStrerror from orbit base (the glibc one is not guaranteed to be thread-safe)

src/UserSpaceInstrumentation/Attach.cpp
```diff
@@ -0,0 +1,107 @@
...
+    }
+  }
+  // Wait for the traced thread to stop. Timeout after one second.
+  int32_t timeout_ms = 1000;
```
dimitry-, line 44: nitpick, kind of... these are more like iterations and we we alwasy get more than 1 second. I do not think we need accuarcy here so maybe rename it to max_attemts? In which case the while below can be a for loop.

src/UserSpaceInstrumentation/Attach.cpp
```diff
@@ -0,0 +1,107 @@
...
+    if (result_waitpid == -1) {
+      return ErrorMessage(absl::StrFormat(
+          "Wait for thread to get traced failed for tid \"%d\": \"%s\"", tid, strerror(errno)));
+    }
+    if (result_waitpid > 0) {
+      // Occasionally the thread is active during PTRACE_ATTACH but terminates before it gets
+      // descheduled. So waitpid returns on exit of the thread instead of the expected stop.
+      if (WIFEXITED(stat_val)) {
+        return false;
+      } else if (WIFSTOPPED(stat_val)) {
```
dimitry-, line 57: nit: drop else?

src/UserSpaceInstrumentation/Attach.cpp
```diff
@@ -0,0 +1,107 @@
...
+        return true;
+      } else {
```
dimitry-, line 59: and here.

src/UserSpaceInstrumentation/Attach.cpp
```diff
@@ -0,0 +1,107 @@
...
+      }
+    }
+    std::this_thread::sleep_for(std::chrono::milliseconds(1));
+    if (--timeout_ms == 0) {
+      ptrace(PTRACE_DETACH, tid, nullptr, nullptr);
+      return ErrorMessage(
+          absl::StrFormat("Waiting for the traced thread \"%d\" to stop timed out.", tid));
+    }
+  }
+  return true;
```
dimitry-, line 72: this line seems to be unreachable....

src/UserSpaceInstrumentation/AttachTest.cpp
```diff
@@ -0,0 +1,36 @@
...
+  TestProcess test_process;
+  const pid_t pid = test_process.pid();
+
+  const auto result = AttachAndStopProcess(pid);
+  ASSERT_TRUE(result) << result.error().message();
+
+  // TestProcess is continuously spawning new threads when it is running. Verify that no new threads
+  // get spawned - i.e. the process is not running anymore.
+  const auto tids = orbit_base::GetTidsOfProcess(pid);
+  absl::SleepFor(absl::Milliseconds(50));
```
dimitry-, line 28: this is not related to this PR: I was thinking that maybe we should start moving away from absl time functions. that is if stl already has everything we need.

src/UserSpaceInstrumentation/TestProcess.cpp
```diff
@@ -0,0 +1,94 @@
...
+  while (true) {
+    if (std::chrono::system_clock::now() > deadline) {
+      break;
+    }
+  }
+  absl::MutexLock lock{&joinable_threads_mutex_};
+  joinable_threads_.emplace(std::this_thread::get_id());
+}
+
+void TestProcess::DummyWorkload() {
```
dimitry-, line 66: Fake?

src/UserSpaceInstrumentation/TestProcess.cpp
```diff
@@ -0,0 +1,94 @@
...
+
+TestProcess::~TestProcess() {
+  fs::remove(flag_file_run_child_);
+  int status;
+  waitpid(pid_, &status, 0);
+  CHECK(WIFEXITED(status));
+  fs::remove(flag_file_child_started_);
+}
+
+void TestProcess::DummyWorker() {
```
dimitry-, line 54: Fake?

src/UserSpaceInstrumentation/TestProcess.h
```diff
@@ -0,0 +1,40 @@
...
+
+namespace orbit_user_space_instrumentation {
+
+// TestProcess forks a new process in the constructor and starts a multi threaded dummy load: A busy
+// loop that spawns and joins threads. The spawned threads perform a busy wait for 15 ms. Four
+// worker threads are kept active. When TestProcess goes out of scope the process is ended.
+class TestProcess {
+ public:
+  TestProcess();
+  ~TestProcess();
```
dimitry-, line 21: nit: virtual or make class final

src/UserSpaceInstrumentation/TestProcess.cpp
```diff
@@ -0,0 +1,94 @@
...
+  if (pid_ == 0) {
+    DummyWorkload();
+    exit(0);
+  } else {
+    while (!fs::exists(flag_file_child_started_)) continue;
+  }
+}
+
+TestProcess::~TestProcess() {
+  fs::remove(flag_file_run_child_);
```
dimitry-, line 47: does this variant throw an exception? if so use the one with error_code.

src/UserSpaceInstrumentation/TestProcess.cpp
```diff
@@ -0,0 +1,94 @@
...
+  flag_file_child_started_ = std::tmpnam(nullptr);
+  Touch(flag_file_run_child_);
+  pid_ = fork();
+  CHECK(pid_ != -1);
+  // Start the workload and have the parent wait for the startup to complete.
+  if (pid_ == 0) {
+    DummyWorkload();
+    exit(0);
+  } else {
+    while (!fs::exists(flag_file_child_started_)) continue;
```
dimitry-, line 42: is this noexcept variant?

src/UserSpaceInstrumentation/Attach.cpp
```diff
@@ -0,0 +1,107 @@
...
+    }
+  }
+  // Wait for the traced thread to stop. Timeout after one second.
+  int32_t timeout_ms = 1000;
+  while (true) {
+    int stat_val = 0;
+    const int result_waitpid = waitpid(tid, &stat_val, WNOHANG);
+    if (result_waitpid == -1) {
+      return ErrorMessage(absl::StrFormat(
+          "Wait for thread to get traced failed for tid \"%d\": \"%s\"", tid, strerror(errno)));
```
dpallotti, line 50: Not important, but: do we want pids/tids enclosed in quotes? Unlike with strings, it doesn't seem too useful.

src/UserSpaceInstrumentation/Attach.cpp
```diff
@@ -0,0 +1,107 @@
...
+      break;
+    }
+  }
+  return outcome::success();
+}
+
+void DetachAndContinueProcess(pid_t pid) {
+  auto tids = GetTidsOfProcess(pid);
+  for (auto tid : tids) {
+    ptrace(PTRACE_DETACH, tid, nullptr, nullptr);
```
dpallotti, line 103: Does this always succeed or could there be issues or errors with this?

src/UserSpaceInstrumentation/AttachTest.cpp
dpallotti: Well Dimitry, I'm glad we finally agree on this. But I have to say that the `absl` version is generally easier to use, `chrono` is sometimes unintuitive. So I'm very much in favor of leaving it as preference.

src/UserSpaceInstrumentation/Attach.cpp
dpallotti: We have UNREACHABLE() in `OrbitBase/Logging.h`, it shouldn't cause warnings.

src/UserSpaceInstrumentation/Attach.cpp
danielfenner: done

src/UserSpaceInstrumentation/Attach.cpp
danielfenner: done

src/UserSpaceInstrumentation/Attach.cpp
danielfenner: done

src/UserSpaceInstrumentation/Attach.cpp
danielfenner: done

src/UserSpaceInstrumentation/Attach.cpp
danielfenner: done

src/UserSpaceInstrumentation/TestProcess.cpp
danielfenner: It isn't a fake (in my understanding of fake). Just dropped "dummy".

src/UserSpaceInstrumentation/TestProcess.cpp
danielfenner: as above

src/UserSpaceInstrumentation/TestProcess.h
danielfenner: done

src/UserSpaceInstrumentation/TestProcess.cpp
danielfenner: done

src/UserSpaceInstrumentation/TestProcess.cpp
danielfenner: done

src/UserSpaceInstrumentation/Attach.cpp
```diff
@@ -0,0 +1,107 @@
...
+    } else {
+      return ErrorMessage(
+          absl::StrFormat("PTRACE_ATTACH failed for \"%d\": \"%s\"", tid, strerror(errno)));
+    }
+  }
+  // Wait for the traced thread to stop. Timeout after one second.
+  int32_t timeout_ms = 1000;
+  while (true) {
+    int stat_val = 0;
+    const int result_waitpid = waitpid(tid, &stat_val, WNOHANG);
```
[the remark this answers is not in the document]
danielfenner: done

src/UserSpaceInstrumentation/Attach.cpp
danielfenner: Yeah, this makes no sense.

src/UserSpaceInstrumentation/Attach.cpp
danielfenner: Good question. It is somewhere between can't happen and programming error. The errors that might occure *should* be catched in the part where we stop the threads. To play save I expose the return value to the caller - although there is no mitigation for this. I could probably check fail directly as well.

src/UserSpaceInstrumentation/Attach.cpp
```diff
@@ -0,0 +1,111 @@
...
+    // can be looked up in the function 'ptrace_attach' in 'ptrace.c' in the kernel sources.
+    if (errno == ESRCH || errno == EPERM) {
+      return false;
+    }
+    return ErrorMessage(
+        absl::StrFormat("PTRACE_ATTACH failed for %d: \"%s\"", tid, SafeStrerror(errno)));
+  }
+  // Wait for the traced thread to stop. Timeout after one second.
+  int32_t max_attempts = 1000;
+  while (true) {
```
dimitry-, line 45: If this converted to for (size_t i=0; i < kMaxAttempts; i++) you can drop unreachable below and have lines 65-68 at the end.

src/UserSpaceInstrumentation/Attach.cpp
```diff
@@ -0,0 +1,111 @@
...
+        continue;
+      }
+      OUTCOME_TRY(result, AttachAndStopThread(tid));
+      if (result) {
+        halted_tids.insert(tid);
+      }
+    }
+    // Update the tids; if there are new ones run another round and stop them.
+    process_tids = GetTidsOfProcess(pid);
+    if (process_tids.size() == halted_tids.size()) {
```
dimitry-, line 93: seems like this could be condition of the while loop.

src/UserSpaceInstrumentation/AttachTest.cpp
danielfenner: Hmm, usability in abseil is better. I always need to read up stuff before using std::chrono.

src/UserSpaceInstrumentation/Attach.cpp
```diff
@@ -0,0 +1,111 @@
...
+
+}  // namespace
+
+[[nodiscard]] ErrorMessageOr<void> AttachAndStopProcess(pid_t pid) {
+  auto process_tids = GetTidsOfProcess(pid);
+  std::set<pid_t> halted_tids;
+  // Note that the process is still running - it can spawn and end threads at this point.
+  while (true) {
+    for (const auto tid : process_tids) {
+      if (halted_tids.count(tid) == 1) {
```
dimitry-, line 83: nit: contains?

src/UserSpaceInstrumentation/Attach.cpp
```diff
@@ -0,0 +1,111 @@
...
+  }
+
+  UNREACHABLE();
+}
+
+}  // namespace
+
+[[nodiscard]] ErrorMessageOr<void> AttachAndStopProcess(pid_t pid) {
+  auto process_tids = GetTidsOfProcess(pid);
+  std::set<pid_t> halted_tids;
```
dimitry-, line 79: nit: flat_hash_set?

src/UserSpaceInstrumentation/TestProcess.cpp
```diff
@@ -0,0 +1,100 @@
...
+    }
+  }
+  absl::MutexLock lock{&joinable_threads_mutex_};
+  joinable_threads_.emplace(std::this_thread::get_id());
+}
+
+void TestProcess::Workload() {
+  size_t kNumThreads = 4;
+  std::vector<std::thread> threads;
+  while (fs::exists(flag_file_run_child_) || !threads.empty()) {
```
dimitry-, line 75: this can throw an exception - use noexpect variant?

src/UserSpaceInstrumentation/TestProcess.cpp
```diff
@@ -0,0 +1,100 @@
...
+    // Spawn as many threads as there are missing.
+    while (threads.size() < kNumThreads && fs::exists(flag_file_run_child_)) {
```
dimitry-, line 77: same here.

dimitry- (approved):
otherwise lgtm

src/UserSpaceInstrumentation/Attach.cpp
danielfenner: done

src/UserSpaceInstrumentation/Attach.cpp
danielfenner: done

src/UserSpaceInstrumentation/Attach.cpp
danielfenner: done

src/UserSpaceInstrumentation/Attach.cpp
danielfenner: done

src/UserSpaceInstrumentation/TestProcess.cpp
danielfenner: done

src/UserSpaceInstrumentation/TestProcess.cpp
danielfenner: done

src/UserSpaceInstrumentation/Attach.cpp
```diff
@@ -0,0 +1,105 @@
...
+    std::this_thread::sleep_for(std::chrono::milliseconds(1));
+  }
+
+  ptrace(PTRACE_DETACH, tid, nullptr, nullptr);
+  return ErrorMessage(absl::StrFormat("Waiting for the traced thread %d to stop timed out.", tid));
+}
+
+}  // namespace
+
+[[nodiscard]] ErrorMessageOr<void> AttachAndStopProcess(pid_t pid) {
```
dpallotti, line 75: (Is `[[nodiscard]]` needed in the .cpp?)

src/UserSpaceInstrumentation/include/UserSpaceInstrumentation/Attach.h
```diff
@@ -0,0 +1,20 @@
...
+
+#include <sys/types.h>
+
+#include "OrbitBase/Result.h"
+
+namespace orbit_user_space_instrumentation {
+
+[[nodiscard]] ErrorMessageOr<void> AttachAndStopProcess(pid_t pid);
+
+[[nodiscard]] ErrorMessageOr<void> DetachAndContinueProcess(pid_t pid);
```
dpallotti, line 16: (This looks very much like one of those cases where `[[nodiscard]]` is added because we add it everywhere but where the caller might want to ignore the result. Not that it matters, but I'm still on the fence with out approach to nodiscard.)

src/UserSpaceInstrumentation/Attach.cpp
```diff
@@ -0,0 +1,105 @@
...
+  auto process_tids = GetTidsOfProcess(pid);
+  absl::flat_hash_set<pid_t> halted_tids;
+  // Note that the process is still running - it can spawn and end threads at this point.
+  while (process_tids.size() != halted_tids.size()) {
+    for (const auto tid : process_tids) {
+      if (halted_tids.contains(tid)) {
+        continue;
+      }
+      OUTCOME_TRY(result, AttachAndStopThread(tid));
```
dpallotti, line 84: Suppose this fails for one thread, aren't all the already stopped thread left in a traced state?

src/UserSpaceInstrumentation/TestProcess.cpp
```diff
@@ -0,0 +1,103 @@
...
+  CHECK(!error);
+  int status;
+  waitpid(pid_, &status, 0);
+  CHECK(WIFEXITED(status));
+  fs::remove(flag_file_child_started_, error);
+  CHECK(!error);
+}
+
+void TestProcess::Worker() {
+  constexpr auto kTimeToLiveMs = std::chrono::milliseconds(15);
```
dpallotti, line 61: I would call this only `kTimeToLive`, as it's not a number in milliseconds but an actual `
std::chrono::duration`.

dpallotti (approved):
The only real doubt I have is about the `OUTCOME_TRY` in `AttachAndStopProcess`. Other than that it looks good.

src/UserSpaceInstrumentation/Attach.cpp
danielfenner: It's enought to put it in the declaration (which makes sense - the compiler usually hasn't seen the definition when issueing a warning).
Removed.

src/UserSpaceInstrumentation/include/UserSpaceInstrumentation/Attach.h
danielfenner: Hmm, I think enforcing checking for errors here is valid - the underlying calls to ptrace can fail in unusual cases.

src/UserSpaceInstrumentation/TestProcess.cpp
danielfenner: done

src/UserSpaceInstrumentation/Attach.cpp
danielfenner: Yes, true. There is a similar case in the detach method. Currently I return with an error and leave things in a state that can not be recovered from.

I added code that at least tries to clean up here so that we can recover. However that might go wrong as well. Let's chat briefly offline about how to handle recoverable vs. unrecoverable errors.

PR #1759, opened 2021-01-29
Update abseil to LTS version 20200923.3

beckerhe:
This updates the version of abseil in Orbit, grpc and grpc_codegen.

All packages have been built and uploaded to the internal and external remotes.


pierricgimmig (approved):
Hi Henning, thanks a lot for the quick turnaround, very much appreciated. LGTM! The only question I have is about the setting the compiler to clang for abseil. Does this mean we can't build Orbit strictly with GCC?

beckerhe:
> Hi Henning, thanks a lot for the quick turnaround, very much appreciated. LGTM! The only question I have is about the setting the compiler to clang for abseil. Does this mean we can't build Orbit strictly with GCC?

Hi Pierric,
Orbit will still work fine with GCC. Sadly it's a bit complex: Actually I only want to set the `compiler.cppstd` setting which is needed by the new version of the abseil package. I could either set this globally for all packages or just for abseil. I've chosen to set
it only for abseil for the time being, because setting it globally would trigger a recompilation of all packages.

And when setting the `cppstd`-setting (like so `abseil:compiler.cppstd=17`) only for a single package, conan requires me to also repeat the global `compiler` setting which leads to this:

```
abseil:compiler=clang
abseil:compiler.cppstd=17
```

But if you have a look at the GCC-profiles (or Visual Studio profiles) you won't find clang there:
```
abseil:compiler=gcc
abseil:compiler.cppstd=17
```

pierricgimmig:
Oh yes of course, I missed that the gcc profiles specified "gcc", nothing to see here :-) Thanks for the precision.

beckerhe:
> Oh yes of course, I missed that the gcc profiles specified "gcc", nothing to see here :-) Thanks for the precision.

Great! ;-)

Rebasing now. Will merge afterwards.

PR #1760, opened 2021-01-29
Draw overlaps of timers as triangles.

florian-kuebler:
Especially GPU timers can overlap (due to the pipeline nature).
This change ensures that the text of such timers does not overlap
and that it is clearly visible to the user that these timers do
overlap.

To do so, the overlapping area is drawn as two triangles. Further,
the coloring is slightly adjusted, such that overlapping timers
do not have the same color, and the triangles can be seen.

Test: Need to be done on the feature/vulkan_layer branch.

Screenshot 2021-01-29 at 18 07 14

### DIFF
--- a/src/OrbitGl/Batcher.cpp
+++ b/src/OrbitGl/Batcher.cpp
@@ -14,39 +14,6 @@
 #include "OrbitBase/Logging.h"
 #include "OrbitBase/Tracing.h"
 
-namespace {
-void GetTriangleGradientColors(const Color& color, std::array<Color, 3>* colors,
-                               ShadingDirection shading_direction) {
-  const float kGradientCoeff = 0.94f;
-  Vec3 dark = Vec3(color[0], color[1], color[2]) * kGradientCoeff;
-  Color dark_color = Color(static_cast<uint8_t>(dark[0]), static_cast<uint8_t>(dark[1]),
-                           static_cast<uint8_t>(dark[2]), color[3]);
-
-  switch (shading_direction) {
-    case ShadingDirection::kLeftToRight:
-      (*colors)[0] = dark_color;
-      (*colors)[1] = dark_color;
-      (*colors)[2] = color;
-      break;
-    case ShadingDirection::kRightToLeft:
-      (*colors)[0] = dark_color;
-      (*colors)[1] = color;
-      (*colors)[2] = color;
-      break;
-    case ShadingDirection::kTopToBottom:
-      (*colors)[0] = color;
-      (*colors)[1] = dark_color;
-      (*colors)[2] = color;
-      break;
-    case ShadingDirection::kBottomToTop:
-      (*colors)[0] = dark_color;
-      (*colors)[1] = color;
-      (*colors)[2] = dark_color;
-      break;
-  }
-}
-}  // namespace
-
 void Batcher::AddLine(Vec2 from, Vec2 to, float z, const Color& color,
                       std::unique_ptr<PickingUserData> user_data) {
   Color picking_color = PickingId::ToColor(PickingType::kLine, user_data_.size(), batcher_id_);
@@ -254,19 +221,29 @@ void Batcher::AddTriangle(const Triangle& triangle, const Color& color,
 void Batcher::AddTriangle(const Triangle& triangle, const Color& color, const Color& picking_color,
                           std::unique_ptr<PickingUserData> user_data) {
   std::array<Color, 3> colors;
-  colors[0] = color;
-  colors[1] = color;
-  colors[2] = color;
+  colors.fill(color);
   AddTriangle(triangle, colors, picking_color, std::move(user_data));
 }
 
-void Batcher::AddShadedTriangle(const Triangle& triangle, const Color& color,
-                                std::unique_ptr<PickingUserData> user_data,
-                                ShadingDirection shading_direction) {
-  std::array<Color, 3> colors;
-  GetTriangleGradientColors(color, &colors, shading_direction);
+void Batcher::AddShadedTrapezium(const Vec3& top_left, const Vec3& bottom_left,
+                                 const Vec3& bottom_right, const Vec3& top_right,
+                                 const Color& color, std::unique_ptr<PickingUserData> user_data,
+                                 ShadingDirection shading_direction) {
+  std::array<Color, 4> colors;  // top_left, bottom_left, bottom_right, top_right.
+  GetBoxGradientColors(color, &colors, shading_direction);  // left points will be "darker".
   Color picking_color = PickingId::ToColor(PickingType::kTriangle, user_data_.size(), batcher_id_);
-  AddTriangle(triangle, colors, picking_color, std::move(user_data));
+  Triangle triangle_1{top_left, bottom_left, top_right};
+  std::array<Color, 3> colors_1;
+  colors_1[0] = colors[0];
+  colors_1[1] = colors[1];
+  colors_1[2] = colors[2];
+  AddTriangle(triangle_1, colors_1, picking_color, std::make_unique<PickingUserData>(*user_data));
+  Triangle triangle_2{bottom_left, bottom_right, top_right};
+  std::array<Color, 3> colors_2;
+  colors_2[0] = colors[1];
+  colors_2[1] = colors[2];
+  colors_2[2] = colors[3];
+  AddTriangle(triangle_2, colors_2, picking_color, std::move(user_data));
 }
 
 void Batcher::AddTriangle(const Triangle& triangle, const std::array<Color, 3>& colors,
--- a/src/OrbitGl/Batcher.cpp
+++ b/src/OrbitGl/Batcher.cpp
@@ -225,24 +225,19 @@ void Batcher::AddTriangle(const Triangle& triangle, const Color& color, const Co
   AddTriangle(triangle, colors, picking_color, std::move(user_data));
 }
 
+// Draw a shaded trapezium with two sides parallel to the x-axis or y-axis.
 void Batcher::AddShadedTrapezium(const Vec3& top_left, const Vec3& bottom_left,
                                  const Vec3& bottom_right, const Vec3& top_right,
                                  const Color& color, std::unique_ptr<PickingUserData> user_data,
                                  ShadingDirection shading_direction) {
   std::array<Color, 4> colors;  // top_left, bottom_left, bottom_right, top_right.
-  GetBoxGradientColors(color, &colors, shading_direction);  // left points will be "darker".
+  GetBoxGradientColors(color, &colors, shading_direction);
   Color picking_color = PickingId::ToColor(PickingType::kTriangle, user_data_.size(), batcher_id_);
   Triangle triangle_1{top_left, bottom_left, top_right};
-  std::array<Color, 3> colors_1;
-  colors_1[0] = colors[0];
-  colors_1[1] = colors[1];
-  colors_1[2] = colors[2];
+  std::array<Color, 3> colors_1{colors[0], colors[1], colors[2]};
   AddTriangle(triangle_1, colors_1, picking_color, std::make_unique<PickingUserData>(*user_data));
   Triangle triangle_2{bottom_left, bottom_right, top_right};
-  std::array<Color, 3> colors_2;
-  colors_2[0] = colors[1];
-  colors_2[1] = colors[2];
-  colors_2[2] = colors[3];
+  std::array<Color, 3> colors_2{colors[1], colors[2], colors[3]};
   AddTriangle(triangle_2, colors_2, picking_color, std::move(user_data));
 }
 
--- a/src/OrbitGl/Batcher.cpp
+++ b/src/OrbitGl/Batcher.cpp
@@ -14,6 +14,39 @@
 #include "OrbitBase/Logging.h"
 #include "OrbitBase/Tracing.h"
 
+namespace {
+void GetTriangleGradientColors(const Color& color, std::array<Color, 3>* colors,
+                               ShadingDirection shading_direction) {
+  const float kGradientCoeff = 0.94f;
+  Vec3 dark = Vec3(color[0], color[1], color[2]) * kGradientCoeff;
+  Color dark_color = Color(static_cast<uint8_t>(dark[0]), static_cast<uint8_t>(dark[1]),
+                           static_cast<uint8_t>(dark[2]), color[3]);
+
+  switch (shading_direction) {
+    case ShadingDirection::kLeftToRight:
+      (*colors)[0] = dark_color;
+      (*colors)[1] = dark_color;
+      (*colors)[2] = color;
+      break;
+    case ShadingDirection::kRightToLeft:
+      (*colors)[0] = dark_color;
+      (*colors)[1] = color;
+      (*colors)[2] = color;
+      break;
+    case ShadingDirection::kTopToBottom:
+      (*colors)[0] = color;
+      (*colors)[1] = dark_color;
+      (*colors)[2] = color;
+      break;
+    case ShadingDirection::kBottomToTop:
+      (*colors)[0] = dark_color;
+      (*colors)[1] = color;
+      (*colors)[2] = dark_color;
+      break;
+  }
+}
+}  // namespace
+
 void Batcher::AddLine(Vec2 from, Vec2 to, float z, const Color& color,
                       std::unique_ptr<PickingUserData> user_data) {
   Color picking_color = PickingId::ToColor(PickingType::kLine, user_data_.size(), batcher_id_);
@@ -220,15 +253,33 @@ void Batcher::AddTriangle(const Triangle& triangle, const Color& color,
 
 void Batcher::AddTriangle(const Triangle& triangle, const Color& color, const Color& picking_color,
                           std::unique_ptr<PickingUserData> user_data) {
+  std::array<Color, 3> colors;
+  colors[0] = color;
+  colors[1] = color;
+  colors[2] = color;
+  AddTriangle(triangle, colors, picking_color, std::move(user_data));
+}
+
+void Batcher::AddShadedTriangle(const Triangle& triangle, const Color& color,
+                                std::unique_ptr<PickingUserData> user_data,
+                                ShadingDirection shading_direction) {
+  std::array<Color, 3> colors;
+  GetTriangleGradientColors(color, &colors, shading_direction);
+  Color picking_color = PickingId::ToColor(PickingType::kTriangle, user_data_.size(), batcher_id_);
+  AddTriangle(triangle, colors, picking_color, std::move(user_data));
+}
+
+void Batcher::AddTriangle(const Triangle& triangle, const std::array<Color, 3>& colors,
+                          const Color& picking_color, std::unique_ptr<PickingUserData> user_data) {
   Triangle rounded_tri = triangle;
-  for (size_t v = 0; v < 3; ++v) {
-    rounded_tri.vertices[v][0] = floorf(rounded_tri.vertices[v][0]);
-    rounded_tri.vertices[v][1] = floorf(rounded_tri.vertices[v][1]);
+  for (auto& vertice : rounded_tri.vertices) {
+    vertice[0] = floorf(vertice[0]);
+    vertice[1] = floorf(vertice[1]);
   }
   float layer_z_value = rounded_tri.vertices[0][2];
   auto& buffer = primitive_buffers_by_layer_[layer_z_value];
   buffer.triangle_buffer.triangles_.push_back(rounded_tri);
-  buffer.triangle_buffer.colors_.push_back_n(color, 3);
+  buffer.triangle_buffer.colors_.push_back(colors);
   buffer.triangle_buffer.picking_colors_.push_back_n(picking_color, 3);
   user_data_.push_back(std::move(user_data));
 }
--- a/src/OrbitGl/TimerTrack.cpp
+++ b/src/OrbitGl/TimerTrack.cpp
@@ -75,7 +75,7 @@ WorldXInfo ToWorldX(double start_us, double end_us, double inv_time_window, floa
   double normalized_start = start_us * inv_time_window;
   double normalized_width = width_us * inv_time_window;
 
-  WorldXInfo result;
+  WorldXInfo result{};
   result.world_x_start = static_cast<float>(world_start_x + normalized_start * world_width);
   result.world_x_width = static_cast<float>(normalized_width * world_width);
   result.is_visible_width = normalized_width * canvas_width > 1;
--- a/src/OrbitGl/TimerTrack.cpp
+++ b/src/OrbitGl/TimerTrack.cpp
@@ -111,7 +111,10 @@ void TimerTrack::UpdatePrimitives(uint64_t min_tick, uint64_t max_tick,
 
   for (auto& chain : chains_by_depth) {
     if (!chain) continue;
-    for (auto& block : *chain) {
+    auto chain_iterator = chain->begin();
+    TimerBlock* prev_block = nullptr;
+    while (chain_iterator != chain->end()) {
+      TimerBlock& block = *chain_iterator;
       if (!block.Intersects(min_tick, max_tick)) continue;
 
       // We have to reset this when we go to the next depth, as otherwise we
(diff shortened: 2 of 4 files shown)


florian-kuebler:
@beckerhe I added you basically for the anonymous namespace, so feel free to drop a complete review.

src/OrbitGl/TimerTrack.cpp
```diff
@@ -103,34 +128,100 @@ void TimerTrack::UpdatePrimitives(uint64_t min_tick, uint64_t max_tick,
...
+        UpdateDepth(timer_info.depth() + 1);
+        double start_us = time_graph_->GetUsFromTick(timer_info.start());
+        double end_us = time_graph_->GetUsFromTick(timer_info.end());
+
+        float world_timer_y = GetYFromTimer(timer_info);
+        float box_height = GetTextBoxHeight(timer_info);
+
+        // Check if the previous timer overlaps with the current one, and if so draw the overlap
+        // as triangles rather than as overlapping rectangles.
+        if (k > 0) {
```
ronaldfw, line 148: Have you profiled the performance impact of these changes? If I'm not mistaken, this affects all tracks with timers, and the changes look like they could potentially cause rendering to be slower.

florian-kuebler: Well, I haven't checked it in depth. I've had a trial run on trata though, and didn't noticed anything. But, tbh. a measurable performance impact would highly surprise me. When looking on the change of drawing when the timers do not overlap (which is the common case, and which would be really bad if performance drops) we have the following:
We have 2 additional checks on the number `k` as well as two loads of the timers. Those timers are explicitly stored in a cache friendly datastructure, thus the previous and next timer is very very likely already in the cache. Thus the 2 additional checks on the time prev./next timer should be also pretty fast. And that's it in the case of overlapping timers.
In addition to this, branch prediction, will also give as a couple of cycles buffer to evaluate the checks (as the branch predictor will note that the case of overlaps is very unlikely).

All in all I'd say, if this change effects the performance of our drawing of timers, something is really broken elsewhere. 
And I would really avoid adding custom code only for the GPU timers here.

ronaldfw: It's easy and straightforward to measure UpdatePrimitives time using the introspection window, it's already instrumented. Measured timings are much more convincing than arguments :)

Start Orbit with '-devmode', click the "circled i" button in the capture window and then take a capture in the introspection window. Use a saved capture to get the same benchmark. 

> All in all I'd say, if this change effects the performance of our drawing of timers, something is really broken elsewhere.

Maybe, but it doesn't matter where the problem is, we still need to make sure we don't have performance regressions.

florian-kuebler: Thanks for the suggestion. I completely forgot about this. I verified on a saved capture (~30sec.) that we don't have any notable performance implication in the "non-overlap" case.

florian-kuebler:
Here is a test capture, in case you want to try it on your own:
http://go/orbit-overlapping-timers-example

src/OrbitGl/TimerTrack.cpp
```diff
@@ -103,34 +128,100 @@ void TimerTrack::UpdatePrimitives(uint64_t min_tick, uint64_t max_tick,
...
+        // Check if the next timer overlaps with the current one, and if so draw the overlap
+        // as triangles rather than as overlapping rectangles.
+        if (k + 1 < block.size()) {
+          const TimerInfo& next_timer_info = block[k + 1].GetTimerInfo();
+          // We also compare the type, as for the Gpu timers, timers of different type but same
+          // depth are drawn below each other (and thus do not overlap).
+          if (timer_info.end() > next_timer_info.start() &&
+              next_timer_info.type() == timer_info.type()) {
+            double intersection_end_us = end_us;
+            end_us = time_graph_->GetUsFromTick(next_timer_info.start());
```
karupayun, line 190: I found a little bit confusing the intersection_end_us and the end_us variables.

Could end_us be renamed as intersection_start_us??

src/OrbitGl/TimerTrack.cpp
```diff
@@ -103,34 +128,100 @@ void TimerTrack::UpdatePrimitives(uint64_t min_tick, uint64_t max_tick,
...
+              auto user_data = std::make_unique<PickingUserData>(
+                  &text_box, [&](PickingId id) { return this->GetBoxTooltip(id); });
+              batcher->AddTriangle(triangle, color, std::move(user_data));
+            }
+          }
+        }
+
+        // If the timer is overlapping completely with the previous and next timer, we neither draw
+        // text nor a box.
+        if (start_us > end_us) continue;
```
karupayun, line 210: How many of these are we having? Are them important? Are you planning to draw them in the future?

I would prefer to draw them without text and then having a pop up when you hover the mouse there, because the user could be interested.

karupayun (commented):
Hi Florian, nice PR! I have a few comments.

1- Different Times. The pop up shows a different time than the text in the box (which makes sense because the rectangle itself is smaller)
different times

2- Black lines. When zooming, sometimes a black line appears. Could it be that the rectangle is not being drawn because is small?? Maybe is something unrelated, feel free to disregard and I can create a bug.

black line

3- Box text.
The name is drawn only in the rectangle part and not in the intersection one. I completely understand the reason, but of course it doesn't seem nice. Why the box name is being cut?
name

I don't think we have an easy solution here but I would like to solve it.
3a) One option could be make a bigger text-rectangle including all the part where the text is possible to be drawn. It needs some calculations about the height of the box and the text one to not draw text in the wrong box. 

3b) Another is changing the way (draw smaller sub-rectangles), but I've tried some experiments and sincerely is a little bit ugly, so I'm not going to propose it:

Rectangles Overlap (1)
At first, trying to make some same-height rectangles, at second your way including the parallelograms in the middle (you can realize that I was lazier in the second drawing :P )

src/OrbitGl/TimerTrack.cpp
```diff
@@ -103,34 +128,100 @@ void TimerTrack::UpdatePrimitives(uint64_t min_tick, uint64_t max_tick,
...
+            if (world_x_info.is_visible_width && !is_collapsed) {
+              Vec3 top_left(world_x_info.world_x_start, world_timer_y + box_height, z);
+              float world_intersection_end_x =
+                  world_x_info.world_x_start + world_x_info.world_x_width;
+              Vec3 bottom_right(world_intersection_end_x, world_timer_y, z);
+              Vec3 top_right(world_intersection_end_x, world_timer_y + box_height, z);
+              Triangle triangle(top_left, bottom_right, top_right);
+              auto user_data = std::make_unique<PickingUserData>(
+                  &text_box, [&](PickingId id) { return this->GetBoxTooltip(id); });
+              batcher->AddTriangle(triangle, color, std::move(user_data));
```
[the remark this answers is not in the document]
karupayun: Another possibility is draw the triangles on top of the rectangles, but it would create a new z-layer.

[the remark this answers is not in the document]
reichlfl: I can't see how the difference in color should result from them being in the same z-layer. Usually whatever is drawn last on the same layer should be completely on top of everything else.

By the way, are we still drawing with the z-buffer enabled? Not that it should make a difference in this case.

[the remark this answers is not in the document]
florian-kuebler: We investigated into this, yesterday. It comes from the fact that we add a gradient (with alpha reduction) to the boxes (therefore they are called AddShadedBox :) ). This is originally done, such that you can distinguish two very close timers of the same color.

src/OrbitGl/TimerTrack.cpp
ronaldfw: We should discuss about this. I have a sense there is a way that the overall algorithm could be unified and lead to the result that @karupayun is proposing. 

Every box has a start and end time on the top of its depth (start_top, end_top) and a start and end time on the bottom of its depth (start_bottom, end_bottom). In the non-overlapping case start_top == start_bottom == start and end_top == end_bottom == end. 

In the general case, start_top == start and end_bottom == end always is true. And end_top is always the minimum of end and start(next), and start_bottom is always the maximum of start and end(previous). 

You always need to draw the two triangles (start_top, end_bottom, end_top) and (start_top, start_bottom, end_bottom) if traversing counter-clockwise. 

This would lead to timers that are "squashed" between two other timers to be drawn as a slanted quadrangle that fits in between the others. 

The assumption I'm making is that one command buffer / debug label execution cannot "overtake" another one. That is, if start < start(next), then also end < end(next). 

(The text label is a problem though. I assume that's why you need to draw a quad in the "middle" and the other two triangles?)

florian-kuebler: We discussed this yesterday a bit offline, and this is exactly what I prototype right now. Though, I am not sure, if we want to keep the rectangle as a base case for non-overlaps, as we then need to draw less elements (might be better in perfomance). I can do some experiments.

karupayun: As discussed offline yesterday, I agree with Ronald. I should have written something here, sorry for that.

The 2 triangles proposals works for every case, as Ronald pointed out. But we should avoid it in cases where it's clear that there isn't any intersection for performance.

As Ronald also pointed out, we spoke that we should not print at all timers who are inside of others. (t1_start < t2_start and t1_end > t2_end). With this assumption, we can easily draw the parallelograms.

But there is still a little issue with coloring. We used the shadowBoxes in Timers to differentiate between consecutive timers, and this is what it is causing problems here. So, how to draw the triangles to:
1- Make the feeling that they below to the same parallelogram and
2- Differentiate with the previous/next.?

We have this n%2 issue for that but we are assuming that events come in order. I think that the proposal solution is inserting them with this n%2 algorithm but repainted them after, so when finalizing the capture, we have the invariant: "for every consecutive pair of timers, one is darker and one is lighter". Am I right @florian-kuebler or we still have to discuss that?

src/OrbitGl/TimerTrack.cpp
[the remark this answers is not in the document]
reichlfl: Ah, makes sense, thanks.

florian-kuebler:
Thanks for you reviews so far. Very valuable feedback. I just restructured the change completely to be aligned with the suggestions. Now we are drawing timers always as two rectangles. Also the shading is preserved, for which I added a AddShadedTriangle method. Please have another look.

src/OrbitGl/TimerTrack.cpp
florian-kuebler: Done. We are now drawing always two triangles. I'll do the performance experiments tomorrow.

src/OrbitGl/TimerTrack.cpp
florian-kuebler: Added a new variable for this.

src/OrbitGl/TimerTrack.cpp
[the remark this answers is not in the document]
florian-kuebler: That should be fixed now.

src/OrbitGl/TimerTrack.cpp
```diff
@@ -103,34 +128,100 @@ void TimerTrack::UpdatePrimitives(uint64_t min_tick, uint64_t max_tick,
...
+
+        // Check if the previous timer overlaps with the current one, and if so draw the overlap
+        // as triangles rather than as overlapping rectangles.
+        if (k > 0) {
+          const TimerInfo& prev_timer_info = block[k - 1].GetTimerInfo();
+          // We also compare the type, as for the Gpu timers, timers of different type but same
+          // depth are drawn below each other (and thus do not overlap).
+          if (prev_timer_info.end() > timer_info.start() &&
+              prev_timer_info.type() == timer_info.type()) {
+            // Ensure we can distinguish overlapping timers with the same color:
```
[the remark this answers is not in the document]
florian-kuebler: Obsolete now.

src/OrbitGl/TimerTrack.cpp
```diff
@@ -186,48 +155,82 @@ void TimerTrack::UpdatePrimitives(uint64_t min_tick, uint64_t max_tick,
...
 
-        Vec2 pos(world_x_info.world_x_start, world_timer_y);
-        Vec2 size(world_x_info.world_x_width, GetTextBoxHeight(timer_info));
-        text_box.SetPos(pos);
-        text_box.SetSize(size);
+        // Draw the Timer's text if it is not collapsed.
+        if (!is_collapsed) {
+          // For overlaps, let us make the textbox just a bit wider:
+          double left_overlap_width_us = start_or_prev_end_us - start_us;
+          double text_x_start_us = start_or_prev_end_us - (.25 * left_overlap_width_us);
```
karupayun, line 178: As I told you, I don't like this. There are cases that it won't work and I don't like that. I will think some easier way to do it more correctly.

karupayun: BTW, if you want to make it anyways, you should create a constant.

ronaldfw: I assume this is talking about the .25? Why do we need to make the textbox a bit wider? Is there an alternative?

florian-kuebler: Basically this is the issue:
name

Still thinking this approach works pretty well and does not require "complex" computations, where I would need to calculate the intersection the character height and the triangle.

ronaldfw: Ok, thanks for the screenshot. So, things look a bit funky when you have a long overlap. 

Yes, it seems you would have to know the character height to align it well to the slanted edge as good as possible. 

I don't feel strongly either way. I would say to go with the simple solution and if we see issues and it bothers us a lot, we'll look into it again later.

src/OrbitGl/TimerTrack.cpp
florian-kuebler: Should be better now.

src/OrbitGl/TimerTrack.cpp
```diff
@@ -101,42 +125,112 @@ void TimerTrack::UpdatePrimitives(uint64_t min_tick, uint64_t max_tick,
...
+          double right_overlap_width_us = end_us - end_or_next_start_us;
+          double text_x_end_us = end_or_next_start_us + (.25 * right_overlap_width_us);
+          WorldXInfo world_x_info =
+              ToWorldX(text_x_start_us, text_x_end_us, inv_time_window, world_start_x, world_width);
+
+          if (world_x_info.is_visible_width) {
```
ronaldfw, line 184: Maybe I'm looking at the wrong thing, but it seems `is_visible_width` is never set in `ToWorldX`.

florian-kuebler: Ah thanks, missed that code branch when removing. Should be fixed now.

src/OrbitGl/TimerTrack.cpp
florian-kuebler: Just to clarify, Pablo provided the screenshot. That was from the version prior the 25% change. After that, you basically have the text perfectly extended in the given case. However, it is just an approximation and there might be situations where this is not pixel perfect.

src/OrbitGl/TimerTrack.cpp
ronaldfw: I don't see the change, yet. Can you check again?

src/OrbitGl/TimerTrack.cpp
karupayun: After thinking a little bit more, I regret about my previous comment. I'm totally happy with this solution. Maybe we could modify the left margin to be 10% or 15% for the underscore (you could check the first character of the text, but that could be tooo much). But I agree with FlorianK here.

karupayun (approved):
LGTM Florian!! :) 

I really like the final state of the solution to the intersecting timers.

src/OrbitGl/TimerTrack.cpp
florian-kuebler: Well commit and then push is sometimes hard.... sorry for that. should be there now.

src/OrbitGl/TimerTrack.cpp
```diff
@@ -101,42 +127,122 @@ void TimerTrack::UpdatePrimitives(uint64_t min_tick, uint64_t max_tick,
...
+              &text_box, [&](PickingId id) { return this->GetBoxTooltip(id); });
+          batcher->AddShadedTriangle(left_triangle, color, std::move(user_data_left),
+                                     ShadingDirection::kLeftToRight);
+
+          Triangle triangle(bottom_left, top_right, bottom_right);
+          auto user_data_right = std::make_unique<PickingUserData>(
+              &text_box, [&](PickingId id) { return this->GetBoxTooltip(id); });
+          batcher->AddShadedTriangle(triangle, color, std::move(user_data_right),
+                                     ShadingDirection::kRightToLeft);
+
```
pierricgimmig, line 235: The overlap code is taking a significant portion of the function. Overlaps are only allowed for a very specific type of track, yet its logic is run for all tracks. Could we have an UpdatePrimitives method specifically for tracks that allow overlap and keep the original implementation for all other tracks?

pierricgimmig (commented):
Hi Florian, have you considered something like this:

image

It seems like it would simplify a lot of the logic and leave more room for text.

src/OrbitGl/Batcher.cpp
```diff
@@ -220,15 +253,33 @@ void Batcher::AddTriangle(const Triangle& triangle, const Color& color,
...
+  colors[1] = color;
+  colors[2] = color;
+  AddTriangle(triangle, colors, picking_color, std::move(user_data));
+}
+
+void Batcher::AddShadedTriangle(const Triangle& triangle, const Color& color,
+                                std::unique_ptr<PickingUserData> user_data,
+                                ShadingDirection shading_direction) {
+  std::array<Color, 3> colors;
+  GetTriangleGradientColors(color, &colors, shading_direction);
```
pierricgimmig, line 267: This assumes a certain ordering of the vertices of the triangle, we should probably have a util function that creates a triangles with the proper ordering of vertices (CreateTriangle(Vec3 bottom_left, Vec3 top_left, Vec3  top_right);). Or even something like CreateTriangleFromBox( Box, TriangleId ) where TriangleId would represent which of the 4 possible triangles to pick.

src/OrbitGl/Batcher.cpp
```diff
@@ -220,15 +253,33 @@ void Batcher::AddTriangle(const Triangle& triangle, const Color& color,
 
 void Batcher::AddTriangle(const Triangle& triangle, const Color& color, const Color& picking_color,
                           std::unique_ptr<PickingUserData> user_data) {
+  std::array<Color, 3> colors;
+  colors[0] = color;
+  colors[1] = color;
+  colors[2] = color;
```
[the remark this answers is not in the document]
florian-kuebler: Done. Thanks.

src/OrbitGl/Batcher.cpp
florian-kuebler: I changed the code a bit. We now have AddShadedTrapezium instead of AddShadedTriangle. That gets the for coordinates and simply re-uses the existing GetBoxGradientColors method.

florian-kuebler:
> Hi Florian, have you considered something like this:
> 
> image
> 
> It seems like it would simplify a lot of the logic and leave more room for text.

Interesting idea. However, I don't see a clear benefit of this solution compared to the current one. especially, the code would look pretty much the same. I'd suggest to keep it as it is now, and if it comes up again, rethink.

src/OrbitGl/Batcher.cpp
```diff
@@ -220,15 +220,43 @@ void Batcher::AddTriangle(const Triangle& triangle, const Color& color,
 
 void Batcher::AddTriangle(const Triangle& triangle, const Color& color, const Color& picking_color,
                           std::unique_ptr<PickingUserData> user_data) {
+  std::array<Color, 3> colors;
+  colors.fill(color);
+  AddTriangle(triangle, colors, picking_color, std::move(user_data));
+}
+
```
karupayun, line 227: // Draw a shaded trapezium with two sides parallel to the x-axis or y-axis.

src/OrbitGl/Batcher.cpp
```diff
@@ -220,15 +220,43 @@ void Batcher::AddTriangle(const Triangle& triangle, const Color& color,
...
+void Batcher::AddShadedTrapezium(const Vec3& top_left, const Vec3& bottom_left,
+                                 const Vec3& bottom_right, const Vec3& top_right,
+                                 const Color& color, std::unique_ptr<PickingUserData> user_data,
+                                 ShadingDirection shading_direction) {
+  std::array<Color, 4> colors;  // top_left, bottom_left, bottom_right, top_right.
+  GetBoxGradientColors(color, &colors, shading_direction);  // left points will be "darker".
+  Color picking_color = PickingId::ToColor(PickingType::kTriangle, user_data_.size(), batcher_id_);
+  Triangle triangle_1{top_left, bottom_left, top_right};
+  std::array<Color, 3> colors_1;
```
karupayun, line 236: std::array<Color, 3> colors_1 = {colors[0], colors[1], colors[2]};?

I don't remember the syntax :)

src/OrbitGl/Batcher.cpp
```diff
@@ -220,15 +220,43 @@ void Batcher::AddTriangle(const Triangle& triangle, const Color& color,
...
+  colors_1[0] = colors[0];
+  colors_1[1] = colors[1];
+  colors_1[2] = colors[2];
+  AddTriangle(triangle_1, colors_1, picking_color, std::make_unique<PickingUserData>(*user_data));
+  Triangle triangle_2{bottom_left, bottom_right, top_right};
+  std::array<Color, 3> colors_2;
```
karupayun, line 242: same here

karupayun (approved):

src/OrbitGl/Batcher.cpp
karupayun: > Hi Florian, have you considered something like this:
> 
> image
> 
> It seems like it would simplify a lot of the logic and leave more room for text.

I really like your approach Pierric. In my opinion it seems to be more intuitive and easy to understand. Actually I tried to draw something to recommend it (your image is clearly nicer :) ). But what to do when we have 3 or more events which intersects each other? I wasn't able to draw something nice and simpler to understand, and the trapezium way seems to be clearly straightforward. I propose to have it as a possible future improve, after receiving some feedback from developers. What do you think?

3-intersecting events

pierricgimmig: Actually I feel like I don't have enough information about the data we are trying to draw. It seems to me that it would be possible to have a scope completely embedded in another one, which would require a different approach. I have a meeting planned with Florian tomorrow, hopefully it will clear things up. Thanks for the drawings btw.

src/OrbitGl/Batcher.cpp
```diff
@@ -220,15 +220,43 @@ void Batcher::AddTriangle(const Triangle& triangle, const Color& color,
...
+  colors.fill(color);
+  AddTriangle(triangle, colors, picking_color, std::move(user_data));
+}
+
+void Batcher::AddShadedTrapezium(const Vec3& top_left, const Vec3& bottom_left,
+                                 const Vec3& bottom_right, const Vec3& top_right,
+                                 const Color& color, std::unique_ptr<PickingUserData> user_data,
+                                 ShadingDirection shading_direction) {
+  std::array<Color, 4> colors;  // top_left, bottom_left, bottom_right, top_right.
+  GetBoxGradientColors(color, &colors, shading_direction);  // left points will be "darker".
```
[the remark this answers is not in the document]
florian-kuebler: Right. I just removed the comment. With the one above and the names of the vertexes, it should be clear.

src/OrbitGl/Batcher.cpp
florian-kuebler: Done.

src/OrbitGl/Batcher.cpp
florian-kuebler: Done.

src/OrbitGl/Batcher.cpp
florian-kuebler: Done.

florian-kuebler:
Merging this now. Let me know if we have any follow ups.

PR #1762, opened 2021-02-01
Fix inconsistent locking behaviour in JoinFutures

beckerhe:
JoinFutures could potentially deadlock, at least according to abseil's
deadlock detection.

This commit changes the locking behaviour to smaller locking scopes that
can't deadlock anymore.

### DIFF
--- a/src/OrbitBase/JoinFutures.cpp
+++ b/src/OrbitBase/JoinFutures.cpp
@@ -28,7 +28,7 @@ Future<void> JoinFutures(absl::Span<const Future<void>> futures) {
   }
 
   auto shared_state = std::make_shared<SharedStateJoin>();
-  absl::MutexLock lock{&shared_state->mutex};
+  shared_state->incompleted_futures = static_cast<int>(futures.size());
 
   for (const auto& future : futures) {
     CHECK(future.IsValid());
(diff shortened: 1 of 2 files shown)


src/OrbitBase/JoinFutures.cpp
```diff
@@ -42,11 +42,19 @@ Future<void> JoinFutures(absl::Span<const Future<void>> futures) {
           }
         });
 
-    if (result == orbit_base::FutureRegisterContinuationResult::kSuccessfullyRegistered) {
-      ++shared_state->incompleted_futures;
+    if (result != orbit_base::FutureRegisterContinuationResult::kSuccessfullyRegistered) {
```
dimitry-, line 45: Is this use-case testable? Maybe add a unit-test for this?

dimitry- (approved):
fell free to merge is this cannot be unit-tested. which I suspect is the case.

src/OrbitBase/JoinFutures.cpp
beckerhe: Yes, can be tested. I will add a test!

beckerhe: Done.

beckerhe:
> fell free to merge is this cannot be unit-tested. which I suspect is the case.

It can. Test was added.

PR #1764, opened 2021-02-01
Add environment variable to enable the Vulkan layer.

florian-kuebler:
Prior this, the Vulkan layer was always auto loaded and enabled if
it was present in one of the Vulkan load directories.

With this change, you need to set the ENABLE_ORBIT_VULKAN_LAYER
(which will be done by the "run --orbit" option), to actually enable
the layer.


dpallotti (approved):
Do we need both the `enable_environment` and the `disable_environment`?

florian-kuebler:
From the tutorial:
>It must have an 'off-switch': an environment variable which can overrule the implicit layer and prevent it from being loaded. This is useful if an application finds that there's an incompatibility with an implicit layer, and wants to stop it loading. If both variables are set, the 'off-switch' has priority and the layer will not be loaded.

PR #1766, opened 2021-02-01
Remove unused IsTracing method, make Run non-static

dimitry-:
Test: builds


florian-kuebler (approved):
LGTM, thanks!

dpallotti (approved):
It's a small change that looks good and I appreciate that you write the "test" part of the commit messages, but please run at least a small local test when making these changes, or at least the integration test. "Test: builds" for me is the equivalent of "no testing".

PR #1768, opened 2021-02-02
Callstacks no longer use hashes as ids

dimitry-:
Changed the code to stop using hashes as unique identifiers.
Now callstack lookup is using absl::flat_hash_map<std::vector<uint64_t>,
id>. Ids are assigned sequentially. Changed client code to use ids
provided by the service.

Test: Run integration tests, run capture with instrumentation, save and
     load capture.

### DIFF
--- a/src/ClientProtos/capture_data.proto
+++ b/src/ClientProtos/capture_data.proto
@@ -56,7 +56,7 @@ message FunctionInfo {
 
 message CallstackEvent {
   uint64 time = 1;
-  uint64 callstack_hash = 2;
+  uint64 callstack_id = 2;
   int32 thread_id = 3;
 }
 
@@ -119,12 +119,12 @@ message UserDefinedCaptureInfo {
 }
 
 message CaptureInfo {
-  reserved 1, 2, 3;
+  reserved 1, 2, 3, 6;
   map<uint64, FunctionInfo> instrumented_functions = 16;
   map<int32, string> thread_names = 4;
   repeated ThreadStateSliceInfo thread_state_slices = 12;
   repeated LinuxAddressInfo address_infos = 5;
-  repeated CallstackInfo callstacks = 6;
+  map<uint64, CallstackInfo> callstacks = 17;
   repeated CallstackEvent callstack_events = 7;
   map<uint64, string> key_to_string = 8;
   // TODO(kuebler): consider removing this field completely.
--- a/src/OrbitCaptureClient/CaptureEventProcessor.cpp
+++ b/src/OrbitCaptureClient/CaptureEventProcessor.cpp
@@ -110,17 +110,14 @@ void CaptureEventProcessor::ProcessInternedCallstack(InternedCallstack interned_
 }
 
 void CaptureEventProcessor::ProcessCallstackSample(const CallstackSample& callstack_sample) {
-  Callstack callstack;
-  if (callstack_sample.callstack_or_key_case() == CallstackSample::kCallstackKey) {
-    callstack = callstack_intern_pool[callstack_sample.callstack_key()];
-  } else {
-    callstack = callstack_sample.callstack();
-  }
+  uint64_t callstack_id = callstack_sample.callstack_key();
+  Callstack callstack = callstack_intern_pool[callstack_id];
 
-  uint64_t hash = GetCallstackHashAndSendToListenerIfNecessary(callstack);
+  SendCallstackToListenerIfNecessary(callstack_id, callstack);
+
   CallstackEvent callstack_event;
   callstack_event.set_time(callstack_sample.timestamp_ns());
-  callstack_event.set_callstack_hash(hash);
+  callstack_event.set_callstack_id(callstack_sample.callstack_key());
   // Note: callstack_sample.pid() is available, but currently dropped.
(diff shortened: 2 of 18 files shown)


src/GrpcProtos/capture.proto
```diff
@@ -97,12 +97,10 @@ message InternedCallstack {
...
 
 message CallstackSample {
+  reserved 3, 4;
   int32 pid = 1;
   int32 tid = 2;
-  oneof callstack_or_key {
-    Callstack callstack = 3;
-    uint64 callstack_key = 4;
-  }
+  uint64 callstack_key = 6;
```
dpallotti, line 103: I'm against getting rid of this `oneof` and I don't see a reason for it. The change of signature of `OnCallstackSample` shows how this just causes a hack now: the `CallstackSample` proto is incomplete when sent and the `key` has to be added afterwards. At the same time, the actual callstack has to be sent as a separate vector parameter, which is inconsistent with the rest of the methods in `TracerListener`.

dpallotti (commented):
I haven't look at the full change but I thought we wanted to discuss how to handle interning in general before starting with changing it?

src/GrpcProtos/capture.proto
dimitry-: Ok, here is the thing - because of this one-of one of the client tests was testing non-existing scenario. I think we shouldn't keep fields in proto we do not use for communication purposes. having "simple" interface of TraceListener is not good enough reason. we either move Intern* processing up one level or change the interface.

dpallotti: Ideally, `TracerListener` wouldn't receive protos but normal struct and it would translate to protos. I'm aware of this but it's quite an additional maintenance work. That's why we did it like this. Your proposed solution is a middle ground that ends up being even more confusing: it's not a matter of "simple", it's a matter of "consistent", which is different.
You say that the interface of `TracerListener` is not a good enough reason, but you don't give a reason to remove the field, other than "we shouldn't keep it". It serves our purpose well, it obviously doesn't take actual space and we can test/CHECK appropriately where we don't expect it to be set. And it's conceptually correct: the event can contain the callstack or the key to an interned callstack. If a user of the proto wants to take out the callstack and intern it, it can do it, which is exactly what happens now.

dimitry-: after discussing offline we decided to move this part to follow-up PR

florian-kuebler (approved):
LGTM, but agree with @dpallotti about proto discussion. Sounds good to move this into a different PR.

src/LinuxTracing/LinuxTracingIntegrationTest.cpp
```diff
@@ -155,7 +156,19 @@ class BufferTracerListener : public TracerListener {
   }
 
   void OnCallstackSample(orbit_grpc_protos::CallstackSample callstack_sample) override {
+    static uint64_t next_callstack_id = 1;
+
+    // TODO: Maybe test it on the sender level?
```
dpallotti, line 161: What does this TODO mean? I would make this clear and possibly create a bug, or actually do it, or just remove it altogether.

src/LinuxTracing/LinuxTracingIntegrationTest.cpp
```diff
@@ -155,7 +156,19 @@ class BufferTracerListener : public TracerListener {
...
+    orbit_grpc_protos::CaptureEvent event_interned_callstack;
+    orbit_grpc_protos::InternedCallstack* interned_callstack =
+        event_interned_callstack.mutable_interned_callstack();
+    uint64_t callstack_key = next_callstack_id++;
+    interned_callstack->set_key(callstack_key);
+    *interned_callstack->mutable_intern() = callstack_sample.callstack();
+    events_.emplace_back(std::move(event_interned_callstack));
```
dpallotti, line 168: I'm confused about why callstacks are processed and interned here. This is supposed to test the behavior of `LinuxTracing` without introducing new events.

src/LinuxTracing/LinuxTracingIntegrationTest.cpp
```diff
@@ -565,7 +578,14 @@ void VerifyCallstackSamplesWithOuterAndInnerFunction(
...
   uint64_t first_matching_callstack_timestamp_ns = std::numeric_limits<uint64_t>::max();
   uint64_t last_matching_callstack_timestamp_ns = 0;
+  absl::flat_hash_map<uint64_t, orbit_grpc_protos::Callstack> callstacks;
   for (const auto& event : events) {
+    if (event.event_case() == orbit_grpc_protos::CaptureEvent::kInternedCallstack) {
+      const orbit_grpc_protos::InternedCallstack& interned_callstack = event.interned_callstack();
+      callstacks[interned_callstack.key()] = interned_callstack.intern();
+      continue;
+    }
+
```
dpallotti, line 588: The behavior of `LinuxTracing` hasn't changed and this test shouldn't change. This is relying on new events introduces by `OnCallstackSample` above, which should not happen in the first place.

src/LinuxTracing/LinuxTracingIntegrationTest.cpp
dimitry-: removed

PR #1769, opened 2021-02-02
Stricter check for OpenGL version support

ronaldfw:
We use Qt's OpenGL support and OpenGL via GLEW. We have seen at least
one instance where there was a mismatch between these two and the
version detected by GLEW was < 2.0, which is the minimum version
required by Orbit. We therefore add a stricter check to also directly
check the version obtained by glGetString(GL_VERSION).

Tested: Manually ran Orbit locally.
Bug: http://b/178349019

### DIFF
--- a/src/OrbitQt/opengldetect.cpp
+++ b/src/OrbitQt/opengldetect.cpp
@@ -4,14 +4,64 @@
 
 #include "opengldetect.h"
 
+// This needs to be first because if it is not GL/glew.h
+// complains about being included after gl.h
+// clang-format off
+#include "OpenGl.h" // IWYU pragma: keep
+// clang-format on
+
+#include <absl/strings/numbers.h>
+#include <absl/strings/str_split.h>
+
 #include <QOffscreenSurface>
 #include <QOpenGLContext>
 #include <QSurfaceFormat>
 
 namespace orbit_qt {
+namespace {
 
-// Detects the supported version of Desktop OpenGL by requesting the most
-// recent version of OpenGL and checking what the system could provide.
+// Must have created an OpenGl context before calling this.
+std::optional<OpenGlVersion> GetOpenGlVersionViaDirectCall() {
+  std::string gl_version = reinterpret_cast<const char*>(glGetString(GL_VERSION));
+  std::vector<std::string> version_tokenized = absl::StrSplit(gl_version, '.');
+
+  if (version_tokenized.size() < 2) {
+    return std::nullopt;
+  }
+
+  int major_version;
+  if (!absl::SimpleAtoi(version_tokenized[0], &major_version)) {
+    return std::nullopt;
+  }
+  int minor_version;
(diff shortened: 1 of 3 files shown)


ronaldfw:
@beckerhe Could you take a look what you think about this? I'm not quite happy with the change, as this is somewhat a shot in the dark.

src/OrbitQt/opengldetect.cpp
```diff
@@ -37,7 +67,21 @@ std::optional<OpenGlVersion> DetectOpenGlVersion() {
...
+  std::vector<std::string> version_tokenized = absl::StrSplit(gl_version, '.');
+
+  if (version_tokenized.size() < 2) {
+    return std::nullopt;
+  }
+
+  int major_version;
+  absl::SimpleAtoi(version_tokenized[0], &major_version);
+  int minor_version;
+  absl::SimpleAtoi(version_tokenized[1], &minor_version);
```
beckerhe, line 80: The `SimpleAtoi` calls could fail - theoretically. Probably they won't, but while reading that code the question why comes up. I suggest either an explaining comment or handling the failed-case.

Personally I would put these 10 lines into its own function that returns a `std::optional<OpenGlVersion>`. And I would return `std::nullopt` if either of the `SimpleAtoi` calls fails (returns false).

src/OrbitQt/orbitglwidget.cpp
```diff
@@ -34,6 +34,9 @@
 #define ORBIT_DEBUG_OPEN_GL 0
 
 OrbitGLWidget::OrbitGLWidget(QWidget* parent) : QOpenGLWidget(parent) {
+  QSurfaceFormat requested_format = QSurfaceFormat::defaultFormat();
+  LOG("OpenGL major version requested: %i", requested_format.majorVersion());
+  LOG("OpenGL minor version requested: %i", requested_format.minorVersion());
```
beckerhe, line 39: Nit: I guess this could be a single log line outputting both major and minor version number.

beckerhe (approved):
Very cool. Thanks Ronald. I added two small comments. Otherwise LGTM!

beckerhe:
> @beckerhe Could you take a look what you think about this? I'm not quite happy with the change, as this is somewhat a shot in the dark.

Well yes, it's a shot in the dark. But I also don't see what we can do more at this point - without investing a lot more time. I like the change; all of it. The FATAL will give us better insights in case there is another crash. And the error message should be helpful to the user.

src/OrbitQt/opengldetect.cpp
beckerhe: It seems that clang agrees with me. `SimpleAtoi` is apparently marked as nodiscard.

ronaldfw: Yes, you are right of course! Fixed.

src/OrbitQt/orbitglwidget.cpp
ronaldfw: Agree, changed.

PR #1770, opened 2021-02-02
Remove GCurrentTimeGraph global.

florian-kuebler:
Effectively the global GCurrentTimeGraph is always
app_::capture_window_::time_graph_ (with the fuzzer being the only
exception). This change removes the global and use the value directly
instead.
It also fixes crashes in the fuzzer, which also works with this change
due to the check if capture_window_ exists.

Follow-up: There is an existing bug, that shows the timerange in
the capture window after clearing a capture. We on clearing a capture,
we should not "clear" the time graph, but actually remove it.

Test: Take, clear, load captures.
Bug: http://b/177651082

### DIFF
--- a/src/OrbitGl/App.cpp
+++ b/src/OrbitGl/App.cpp
@@ -176,10 +176,10 @@ void OrbitApp::OnCaptureStarted(ProcessData&& process,
         capture_data_ =
             CaptureData(std::move(process), module_manager_.get(), std::move(selected_functions),
                         std::move(selected_tracepoints), std::move(frame_track_function_ids));
-        capture_window_->GetTimeGraph()->SetCaptureData(&capture_data_.value());
+        GetMutableTimeGraph()->SetCaptureData(&capture_data_.value());
 
         frame_track_online_processor_ =
-            FrameTrackOnlineProcessor(GetCaptureData(), GCurrentTimeGraph);
+            FrameTrackOnlineProcessor(GetCaptureData(), GetMutableTimeGraph());
 
         CHECK(capture_started_callback_);
         capture_started_callback_();
@@ -256,7 +256,7 @@ void OrbitApp::OnCaptureFailed(ErrorMessage error_message) {
 
 void OrbitApp::OnTimer(const TimerInfo& timer_info) {
   if (timer_info.function_id() == 0) {
-    GCurrentTimeGraph->ProcessTimer(timer_info, nullptr);
+    GetMutableTimeGraph()->ProcessTimer(timer_info, nullptr);
     return;
   }
 
@@ -264,7 +264,7 @@ void OrbitApp::OnTimer(const TimerInfo& timer_info) {
   const FunctionInfo& func = capture_data.instrumented_functions().at(timer_info.function_id());
   uint64_t elapsed_nanos = timer_info.end() - timer_info.start();
   capture_data.UpdateFunctionStats(func, elapsed_nanos);
-  GCurrentTimeGraph->ProcessTimer(timer_info, &func);
+  GetMutableTimeGraph()->ProcessTimer(timer_info, &func);
   frame_track_online_processor_.ProcessTimer(timer_info, func);
 }
 
@@ -558,7 +558,7 @@ void OrbitApp::MainTick() {
 
   if (DoZoom && HasCaptureData()) {
     // TODO (b/176077097): TrackManager has to manage sorting by their own.
-    GCurrentTimeGraph->GetTrackManager()->SortTracks();
+    GetMutableTimeGraph()->GetTrackManager()->SortTracks();
     capture_window_->ZoomAll();
     NeedsRedraw();
     DoZoom = false;
@@ -567,7 +567,6 @@ void OrbitApp::MainTick() {
 
 void OrbitApp::SetCaptureWindow(CaptureWindow* capture) {
   CHECK(capture_window_ == nullptr);
-  GCurrentTimeGraph = capture->GetTimeGraph();
   capture_window_ = capture;
   capture_window_->set_draw_help(false);
 }
@@ -691,7 +690,7 @@ void OrbitApp::ClearSelectionBottomUpView() {
 }
 
 std::string OrbitApp::GetCaptureTime() {
-  double time = GCurrentTimeGraph != nullptr ? GCurrentTimeGraph->GetCaptureTimeSpanUs() : 0.0;
+  double time = GetTimeGraph()->GetCaptureTimeSpanUs();
   return GetPrettyTime(absl::Microseconds(time));
 }
 
@@ -786,8 +785,7 @@ PresetLoadState OrbitApp::GetPresetLoadState(
 ErrorMessageOr<void> OrbitApp::OnSaveCapture(const std::string& file_name) {
   const auto& key_to_string_map = string_manager_->GetKeyToStringMap();
 
-  std::vector<std::shared_ptr<TimerChain>> chains =
-      GCurrentTimeGraph->GetAllSerializableTimerChains();
+  std::vector<std::shared_ptr<TimerChain>> chains = GetTimeGraph()->GetAllSerializableTimerChains();
 
   TimerInfosIterator timers_it_begin(chains.begin(), chains.end());
   TimerInfosIterator timers_it_end(chains.end(), chains.end());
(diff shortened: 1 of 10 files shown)


dimitry- (approved):

src/OrbitGl/App.cpp
```diff
@@ -909,17 +906,15 @@ void OrbitApp::AbortCapture() {
 
 void OrbitApp::ClearCapture() {
   ORBIT_SCOPE_FUNCTION;
-  capture_window_->GetTimeGraph()->SetCaptureData(nullptr);
```
florian-kuebler, line 912: Note, that the "set capture data" now happens inside the "clear".

src/OrbitGl/App.cpp
```diff
@@ -1166,42 +1161,41 @@ void OrbitApp::LoadSymbols(const std::filesystem::path& symbols_path, ModuleData
...
+        }
+        LOG("Added frame tracks in module \"%s\"", module_data->file_path());
+      }
 
-          UpdateAfterSymbolLoading();
-          FireRefreshCallbacks();
-        });
+      UpdateAfterSymbolLoading();
+      FireRefreshCallbacks();
+    });
```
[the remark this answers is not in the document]
florian-kuebler: That is feeling strange. I do clang-format all the files that I touch, so it might be an issue here.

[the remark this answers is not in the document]
florian-kuebler: It looks correct and clang-format check accepts it. So I'd just keep it.

src/OrbitGl/App.cpp
```diff
@@ -1166,42 +1161,41 @@ void OrbitApp::LoadSymbols(const std::filesystem::path& symbols_path, ModuleData
...
+                                         std::move(frame_track_function_hashes)] {
+      modules_currently_loading_.erase(module_data->file_path());
+
+      const ProcessData* selected_process = GetTargetProcess();
+      if (selected_process != nullptr &&
+          selected_process->IsModuleLoaded(module_data->file_path())) {
+        functions_data_view_->AddFunctions(module_data->GetFunctions());
+        LOG("Added loaded function symbols for module \"%s\" to the functions tab",
+            module_data->file_path());
+      }
```
beckerhe, line 1176: I can't see what has changed here. Am I blind?

src/OrbitGl/App.cpp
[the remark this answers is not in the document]
beckerhe: Could you put that into a separate commit? It will make rebasing for others a lot easier.

src/OrbitGl/App.h
```diff
@@ -145,6 +145,14 @@ class OrbitApp final : public DataViewFactory, public CaptureListener {
...
 
   void SetCaptureWindow(CaptureWindow* capture);
+  [[nodiscard]] const TimeGraph* GetTimeGraph() const {
+    CHECK(capture_window_ != nullptr);
+    return capture_window_->GetTimeGraph();
+  }
+  [[nodiscard]] TimeGraph* GetMutableTimeGraph() {
+    CHECK(capture_window_ != nullptr);
+    return capture_window_->GetTimeGraph();
+  }
```
beckerhe, line 155: There is an `if (capture_window != nullptr)` somewhere in the code. That tells me that we can't assume that `capture_window_` is always existing. Have you considered returning a `std::optional<TimeGraph*>` and not abort if `capture_window_` does not exist?

src/OrbitGl/LiveFunctionsController.cpp
```diff
@@ -35,9 +35,8 @@ std::pair<uint64_t, uint64_t> ComputeMinMaxTime(
 uint64_t AbsDiff(uint64_t a, uint64_t b) {
   if (a > b) {
     return a - b;
-  } else {
-    return b - a;
   }
+  return b - a;
```
beckerhe, line 39: Nit: This is unrelated and should be a separate commit.

src/OrbitGl/LiveFunctionsController.cpp
```diff
@@ -49,22 +48,22 @@ const TextBox* ClosestTo(uint64_t point, const TextBox* box_a, const TextBox* bo
   return box_b;
 }
 
-const TextBox* SnapToClosestStart(uint64_t function_id) {
-  double min_us = GCurrentTimeGraph->GetMinTimeUs();
-  double max_us = GCurrentTimeGraph->GetMaxTimeUs();
+const TextBox* SnapToClosestStart(const OrbitApp* app, uint64_t function_id) {
```
beckerhe, line 51: When is `OrbitApp` actually needed? As far as I can see a `const TimeGraph*` would be enough.

beckerhe (commented):
Sorry for being late. I added some comments. I hope it's not too late. :see_no_evil:

src/OrbitGl/App.cpp
beckerhe: Nevermind. See other comment.

src/OrbitGl/App.h
florian-kuebler: Well, yeah I though of this. But in the code there is exactly one place where we do the capture_window != nullptr check, on all other places we currently assume that GCurrentTimeGraph is not null (actually without stating it explicitly). I am not sure if we want to add handling for " no capture window present" for all the callers, or if we would end up in adding CHECKs on every caller, in which case we could also just keep it as it is now. It is an improvement to the current state, as we at least do have the checks here.

But I am happy to discuss, and don't have strong feelings about this.

Nit question: Assume we do handle this case here, why not just returning nullptr?

beckerhe: I see your point. Before we ignored it, but it wasn't so obvious. Now, we keep the ignoring part but at least improve in other parts. So I guess it's fine for me to leave it.

Regarding nullptr: I prefer not to return a nullptr, but rather make a syntactical change. We had a lot of crashes over the last few releases which were related to similar getters. `GetProcess`, `GetProcessManager`, `GetModuleByPath`. You never know whether you can assume the returned pointer to be valid or not. And someone might change the behaviour and it will still compile. That's why I advocate for a raw pointer to be never null. If you need an empty value, wrap it into a std::optional.

I want to make a small presentation about that because I feel that is related to a lot of crashes in the past. But of course that shouldn't stop you now. :-)

src/OrbitGl/LiveFunctionsController.cpp
florian-kuebler: Done.

src/OrbitGl/LiveFunctionsController.cpp
florian-kuebler: Done.

src/OrbitGl/App.h
florian-kuebler: Okay, then lets leave it for now.

src/OrbitGl/App.cpp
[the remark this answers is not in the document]
florian-kuebler: Should be obsolete now.

florian-kuebler (commented):
Addressed the comments. Thanks for those.

beckerhe (approved):
Thank you Florian!

src/OrbitGl/LiveFunctionsController.cpp
```diff
@@ -217,12 +217,18 @@ uint64_t LiveFunctionsController::GetStartTime(uint64_t index) {
   return GetCaptureMin();
 }
 
-uint64_t LiveFunctionsController::GetCaptureMin() { return GCurrentTimeGraph->GetCaptureMin(); }
-uint64_t LiveFunctionsController::GetCaptureMax() { return GCurrentTimeGraph->GetCaptureMax(); }
+uint64_t LiveFunctionsController::GetCaptureMin() {
```
karupayun, line 220: const and [[nodiscard]]?

src/OrbitGl/LiveFunctionsController.cpp
```diff
@@ -217,12 +217,18 @@ uint64_t LiveFunctionsController::GetStartTime(uint64_t index) {
...
+  CHECK(app_ != nullptr);
+  return app_->GetTimeGraph()->GetCaptureMin();
+}
+uint64_t LiveFunctionsController::GetCaptureMax() {
```
karupayun, line 224: const and [[nodiscard]]?

src/OrbitGl/LiveFunctionsController.cpp
florian-kuebler: Good catch. I can do a follow-up PR on Monday. There were quite a bit of clang-tidy warnings in this file.

PR #1771, opened 2021-02-02
Replace symbol loading CHECK with Error message

antonrohr:
b/178176772
The solution this change represents is in general not a good idea. But I don't have any clue why the Check fails in the first place and cannot reproduce it. Additionally this code will change with the upcoming #1752. Since #1752 wont make it into 1.59, I made this quick-n-dirty fix. If there are any better ideas, I would like to hear about it.

### DIFF
--- a/src/OrbitGl/App.cpp
+++ b/src/OrbitGl/App.cpp
@@ -1157,7 +1157,15 @@ void OrbitApp::LoadSymbols(const std::filesystem::path& symbols_path, ModuleData
                           frame_track_function_hashes =
                               std::move(frame_track_function_hashes)]() mutable {
     auto symbols_result = SymbolHelper::LoadSymbolsFromFile(symbols_path);
-    CHECK(symbols_result);
+    if (!symbols_result) {
+      std::string error_message{absl::StrFormat("Unable to load symbols for %s from file %s: %s",
+                                                module_data->file_path(), symbols_path.string(),
+                                                symbols_result.error().message())};
+      ERROR("%s", error_message);
+      SendErrorToUi("Unexpected error while loading symbols", error_message);
+      return;
+    }
+
     module_data->AddSymbols(symbols_result.value());
 
     std::string message =


beckerhe:
> b/178176772
> The solution this change represents is in general not a good idea. But I don't have any clue why the Check fails in the first place and cannot reproduce it. Additionally this code will change with the upcoming #1752. Since #1752 wont make it into 1.59, I made this quick-n-dirty fix. If there are any better ideas, I would like to hear about it.

I thought we concluded that this happens in the first place when you try to load symbols from a module without a .symtab section. Is this not the case? (AFAIK stripped executables might not have a symtab section)

dimitry- (approved):
This actually looks fine to me. Especially if this could be triggered by putting stripped file into symbol search path.

antonrohr:
> I thought we concluded that this happens in the first place when you try to load symbols from a module without a .symtab section. Is this not the case? (AFAIK stripped executables might not have a symtab section)

That could be the case, but it actually shouldn't. I just checked all code paths and in theory we never reach this CHECK with a file that does not have a symtab section. 

There are 4 ways to find symbols:
* (local) via SymbolPaths.txt: `SymbolHelper::FindSymbolsWithSymbolsPathFile` calls `SymbolHelper::VerifySymbolsFile`, which checks for the symtab section. 
* (local) in cache: `SymbolHelper::FindSymbolsInCache` also calls `VerifySymbolsFile`
* if we run with --local, the symbol could be included in the module -> also calls `VerifySymbolsFile`
* (remote) : `ServiceUtils::FindSymbolsFilePath()` also checks for symtab

antonrohr:
> Especially if this could be triggered by putting stripped file into symbol search path.

Orbit *should* not consider files that don't have a symtab section, even when they are in the symbol search path (SymbolPaths.txt). See my other comment.
So I really just dont know whats going on.

beckerhe:
> > I thought we concluded that this happens in the first place when you try to load symbols from a module without a .symtab section. Is this not the case? (AFAIK stripped executables might not have a symtab section)
> 
> That could be the case, but it actually shouldn't. I just checked all code paths and in theory we never reach this CHECK with a file that does not have a symtab section.
> 
> There are 4 ways to find symbols:
> 
> * (local) via SymbolPaths.txt: `SymbolHelper::FindSymbolsWithSymbolsPathFile` calls `SymbolHelper::VerifySymbolsFile`, which checks for the symtab section.
> * (local) in cache: `SymbolHelper::FindSymbolsInCache` also calls `VerifySymbolsFile`
> * if we run with --local, the symbol could be included in the module -> also calls `VerifySymbolsFile`
> * (remote) : `ServiceUtils::FindSymbolsFilePath()` also checks for symtab

Ah I see. So this is ensured before even getting to OrbitApp::LoadSymbols. I guess I was fooled by `SymbolHelper::LoadSymbolsFromFile` also checking for symtab.

beckerhe:
> > Especially if this could be triggered by putting stripped file into symbol search path.
> 
> Orbit _should_ not consider files that don't have a symtab section, even when they are in the symbol search path (SymbolPaths.txt). See my other comment.
> So I really just dont know whats going on.

Well we at least assume here that a file on disk (which we might not control) doesn't change in between subsequent accesses. And that's probably not a good idea.

beckerhe (approved):
LGTM

antonrohr:
> Ah I see. So this is ensured before even getting to OrbitApp::LoadSymbols.

I wasn't sure about this when we talked about this offline, but I am sure now.

antonrohr:
Screenshot 2021-02-02 at 17 36 56

The "Close with Comment" and "Comment" button are to close together 😅

src/OrbitGl/App.cpp
```diff
@@ -1157,7 +1157,15 @@ void OrbitApp::LoadSymbols(const std::filesystem::path& symbols_path, ModuleData
                           frame_track_function_hashes =
                               std::move(frame_track_function_hashes)]() mutable {
     auto symbols_result = SymbolHelper::LoadSymbolsFromFile(symbols_path);
-    CHECK(symbols_result);
+    if (!symbols_result) {
+      std::string error_message{absl::StrFormat("Unable to load symbols for %s from file %s: %s",
+                                                module_data->file_path(), symbols_path.string(),
+                                                symbols_result.error().message())};
+      ERROR("%s", error_message);
+      SendErrorToUi("Failure loading symbols", error_message);
```
dimitry-, line 1165: nit: Unexpected error while loading symbols?

antonrohr: Good idea. done

dimitry-:
Ok, after some thinking I think the better version is to replace CHECK with FATAL(result.error_message()); This would provide us with additional context if we get logs alongside the crash. WDYT?

antonrohr:
> Ok, after some thinking I think the better version is to replace CHECK with FATAL(result.error_message()); This would provide us with additional context if we get logs alongside the crash. WDYT?

That is an interesting tradeoff. As you know, this would not be the best user experience, but good for us. I do not want to make this decision. I will open this up to all of us, by posting in the chat about it.

antonrohr:
After the chat discussion was inconclusive, I decided to merge as is

PR #1772, opened 2021-02-03
Replace GetSystemTimeAsFileTime by std::chrono::steady_clock::now()

pierricgimmig:
GetSystemTimeAsFileTime has a resolution of 1ms on our machines. Using
std::chrono's steady_clock gives a resolution of 100ns.

### DIFF
--- a/src/OrbitBase/include/OrbitBase/Profiling.h
+++ b/src/OrbitBase/include/OrbitBase/Profiling.h
@@ -6,7 +6,7 @@
 #define ORBIT_BASE_PROFILING_H_
 
 #ifdef _WIN32
-#include <Windows.h>
+#include <chrono>
 #else
 #include <stdint.h>
 #include <time.h>
@@ -14,9 +14,8 @@
 
 #ifdef _WIN32
 [[nodiscard]] inline uint64_t MonotonicTimestampNs() {
-  __int64 time;
-  GetSystemTimeAsFileTime((FILETIME*)&time);
-  return static_cast<uint64_t>(time) * 100;
+  auto time_since_epoch = std::chrono::high_resolution_clock::now().time_since_epoch();
+  return std::chrono::duration_cast<std::chrono::nanoseconds>(time_since_epoch).count();
 }
 #else
 [[nodiscard]] inline uint64_t MonotonicTimestampNs() {
--- a/src/OrbitBase/include/OrbitBase/Profiling.h
+++ b/src/OrbitBase/include/OrbitBase/Profiling.h
@@ -14,7 +14,7 @@
 
 #ifdef _WIN32
 [[nodiscard]] inline uint64_t MonotonicTimestampNs() {
-  auto time_since_epoch = std::chrono::high_resolution_clock::now().time_since_epoch();
+  auto time_since_epoch = std::chrono::steady_clock::now().time_since_epoch();
   return std::chrono::duration_cast<std::chrono::nanoseconds>(time_since_epoch).count();
 }
 #else


src/OrbitBase/include/OrbitBase/Profiling.h
```diff
@@ -6,17 +6,16 @@
...
 #ifdef _WIN32
 [[nodiscard]] inline uint64_t MonotonicTimestampNs() {
-  __int64 time;
-  GetSystemTimeAsFileTime((FILETIME*)&time);
-  return static_cast<uint64_t>(time) * 100;
+  auto time_since_epoch = std::chrono::high_resolution_clock::now().time_since_epoch();
+  return std::chrono::duration_cast<std::chrono::nanoseconds>(time_since_epoch).count();
 }
 #else
 [[nodiscard]] inline uint64_t MonotonicTimestampNs() {
```
dimitry-, line 21: iirc we want to keep this as clock_gettime(CLOCK_MONOTONIC) for compatibility with perf_event_open timestamps? Maybe add a comment here so that it is not merged with method above? Or even rename this method so that it is clearer why this is not implemented via <chrono> calls.

dimitry- (approved):

src/OrbitBase/include/OrbitBase/Profiling.h
```diff
@@ -6,17 +6,16 @@
...
 #include <stdint.h>
 #include <time.h>
 #endif
 
 #ifdef _WIN32
 [[nodiscard]] inline uint64_t MonotonicTimestampNs() {
-  __int64 time;
-  GetSystemTimeAsFileTime((FILETIME*)&time);
-  return static_cast<uint64_t>(time) * 100;
+  auto time_since_epoch = std::chrono::high_resolution_clock::now().time_since_epoch();
```
dpallotti, line 17: This is the problem with having made this function both Windows and Linux when it was Linux only.

This function is called "monotonic timestamp" because of `CLOCK_MONOTONIC`, which as Dimitry said is the clock used with `perf_event_open` (and in general for profiling events). The Windows implementation here (both the old one and the new one) uses *a* clock, but it doesn't really seem to have an association with `CLOCK_MONOTONIC`. Note in particular that the `chrono` equivalent of it should be `steady_clock`, I don't know if `high_resolution_clock` uses the same clock.

ronaldfw: At least according to https://en.cppreference.com/w/cpp/chrono/high_resolution_clock, high_resolution_clock should be avoided, and it may indeed not be monotonic. From earlier discussions and reading, my take was, as @dpallotti is saying, that one should use steady_clock for interval timing measurements.

ronaldfw: > This is the problem with having made this function both Windows and Linux when it was Linux only.

We need timing measurements on Windows. The name is fine if we use steady_clock, which is guaranteed to be monotonic.

ronaldfw: But I also agree with @dimitry- above that Linux should keep using clock_gettime(CLOCK_MONOTONIC), so maybe two different functions is really the way to go here?

dpallotti: > We need timing measurements on Windows. The name is fine if we use steady_clock, which is guaranteed to be monotonic.

"Monotonic timestamp" didn't mean "a timestamp from *a* clock that is monotonic", rather a timestamp from *the* `CLOCK_MONOTONIC`". That's also the reason why we don't use `steady_clock` on Linux, because we want to make sure that it's the same clock as for perf_event_open. That's a stronger requirement than simply measuring time differences between two points in the code, for which several clocks or way of measuring time can be used.

So I'm not saying there shouldn't be a function returning timestamps for measuring generic time differences, but that it should have been kept separate from the original function. There could be a function that wraps `steady_clock` for convenience as the `chrono` interface is horrible, or maybe `absl` already has something, or one just uses `chrono::steady_clock` directly. But that's a different use case than getting timestamps that fit in the perf_event_open time domain.

dimitry-: maybe we should just use absl::Now() in places where it is not required to be in sync with perf_event_open clock? Move the linux version of the function to LinxuTracing and replace the rest with absl?

pierricgimmig: The introspection code for example is the same on Linux and Windows and uses the current MonotonicTimestampNs() function. I think it does make sense to have a cross platform timestamp provider.  We should consider renaming the function to TimestampNs(). We need all our timestamps to be consistent on a per platform basis, regardless of the clock type. On Linux, it makes sense to use the monotonic clock because that's the same clock we use for perf_event_open, using any other clock would not be advised. On Windows, for now we only have one source, so we don't really care about the clock type, but eventually if we bring back ETW events, we'd have to make sure that TimestampNs() match the ETW clock. The point is that for a given platform, all our timestamps should be on the same time domain, so we should have a single provider to avoid any errors, but the specific clock type is not important.

dpallotti: > On Linux, it makes sense to use the monotonic clock because that's the same clock we use for perf_event_open, using any other clock would not be advised.

> the specific clock type is not important

One can specify what clock `perf_event_open` uses, and if I'm not mistaken it's not even `CLOCK_MONOTONIC` by default, which caused us problems in the past. A more generic `TimestampNs` function dilutes this association even more, especially considering the plethora of available clocks. While it's convenient to have one function to use everywhere, many details are lost in the process, until things stop working.

> The introspection code for example is the same on Linux and Windows and uses the current MonotonicTimestampNs() function.

Can't this just be done differently, e.g., have a wrapper local to introspection that calls the right function depending on Windows or Linux based on the needs of introspection, instead of making this function codebase-global and making the same requirement assumptions for all the other users of such function?

pierricgimmig: > One can specify what clock `perf_event_open` uses, and if I'm not mistaken it's not even `CLOCK_MONOTONIC` by default, which caused us problems in the past. A more generic `TimestampNs` function dilutes this association even more, especially considering the plethora of available clocks. While it's convenient to have one function to use everywhere, many details are lost in the process, until things stop working.

Let's say we decide to change the perf_event_open clock to something else. Should we then go through all the code base and change calls to MonotonicTimestampNs() to whatever new type of clock we use or should we simply change the implementation of a single generic TimestampNs() function?

Similarly, should all users of timestamps define platform specific wrappers or should we use a single generic timestamp provider?

Again, I think that what is important is to have consistency, the type of clock is not important.

dpallotti: > or should we use a single generic timestamp provider?

No.

> should all users of timestamps define platform specific wrappers ...?

There's no need for that unless you need a specific clock, which is a specific use case. In the specific case, yes, it's better to be explicit.

`MonotonicTimestampNs` should only be used where it matters that it's the same clock as perf_event_open, which is how it was originally. It should not be used, say, to measure how long it takes to perform a certain computation in the ui.
At that point, if `MonotonicTimestampNs` changes, you only change it where it really matters, and not in all the other code that has nothing to do with `CLOCK_MONOTONIC`.

What you are proposing is for the entire codebase to use the same clock as perf_event_open needs in the name of "consistency", which makes no sense.

pierricgimmig: > What you are proposing is for the entire codebase to use the same clock as perf_event_open needs in the name of "consistency", which makes no sense.

We are in the end using timestamps to display data from multiple sources on a single timeline, consistency couldn't make more sense, it's required, no way around it. Why should we allow timestamps from different time domains to be used? This is what doesn't make much sense to me.

dpallotti: Consistency here is required for profiling sources, yes, where we need to be in sync with `perf_event_open`, but not for everything else. For everything else (e.g., the entirety of the client), most clocks work (e.g, `steady_clock`), and there shouldn't be a connection with `perf_event_open`. But a generic `TimestampNs` suggests that it should be used everywhere you need a timestamp, which creates a tie with `perf_event_open` that in most cases shouldn't exist.

pierricgimmig: I strongly disagree. We are building this tool to visualize events on a timeline, it's there on a silver platter to also visualize our own internal events, hence the introspection window. I see absolutely no reason to use different clocks, zero added value (do you see one?). Anything that has a timestamp should be allowed on the timeline. The client can have its introspection timers on the same timeline as the local process it's profiling for example, and that is highly desirable.

> But a generic `TimestampNs` suggests that it should be used everywhere you need a timestamp
That's exactly what I suggest, yes.

dpallotti: My point is way more generic. You are talking about introspection on Linux, in which case of course I agree that the clock should be the same as the one used by other sources as introspection on Linux is effectively a profiling source.

But this function is in `OrbitBase` and can be used anywhere, and provides a Windows implementation that has nothing to do with `CLOCK_MONOTONIC`. This is wrong twice: on other Linux code, it ties to perf_event_open even when not necessary (not everything is introspection or a profiling source). On Windows code, `CLOCK_MONOTONIC` just doesn't exist.

So there should be a `MonotonicTimestampNs` for Linux only when the same clock as perf_event_open is needed. Everywhere else, a different function should be used, or `absl`, or `steady_clock` directly, which circles back to the initial point.

ronaldfw: > "Monotonic timestamp" didn't mean "a timestamp from a clock that is monotonic", rather "a timestamp from the CLOCK_MONOTONIC" (which doesn't exist on Windows).

It's ambiguous though. "Monotonic clock" has a clear meaning, namely a clock that is monotonically increasing. If we need a clock that uses CLOCK_MONOTONIC on Linux (which we do) then that shouldn't be in OrbitBase, the name should clearly reflect that and there needs to be clear comments about intended usage. 

> So I'm not saying there shouldn't be a function returning timestamps for measuring generic time differences, but that it should have been kept separate from the original function.

I agree with that. See also my point above that the clock using CLOCK_MONOTONIC should probably not be in OrbitBase.

dpallotti: > "Monotonic timestamp" didn't mean "a timestamp from a clock that is monotonic", rather "a timestamp from the CLOCK_MONOTONIC" (which doesn't exist on Windows).

Well, that was the original name when it was in `LinuxTracing` and it wasn't that ambiguous there. It's certainly ambiguous in `OrbitBase`. We can rename as needed.

But in any case, what makes things particularly bad is that the semantics of the function is returning *a* monotonic timestamp on Windows and *the* `CLOCK_MONOTONIC` on Linux. That's bad and that's why we should have two separate functions, the one returning `CLOCK_MONOTONIC` Linux only (I guess it can stay in `OrbitBase`, but it has to be Linux only).

ronaldfw: Agree on the proposal to have two functions, one Linux-only that uses CLOCK_MONOTONIC, and one that can be used on Windows (and possibly Linux if CLOCK_MONOTONIC is not required). The latter should also return a monotonic timestamp as we need it to compute durations.

I don't feel strongly about it, but IMO moving the CLOCK_MONOTONIC function to LinuxTracing would help reduce misunderstanding about which clock to use in which case.

dpallotti: > IMO moving the CLOCK_MONOTONIC function to LinuxTracing would help reduce misunderstanding about which clock to use in which case

The reason why it was moved out is because it's also needed in other "profiling" modules, e.g., Vulkan layer and Linux UI introspection, we can rename though. But at the same it should be Linux only and clearly for the specific purpose of having a timestamp that ends up in the same timeline as perf_event_open events. In my opinion it shouldn't be for a "generic" cross-platform need for a timestamp difference (completely random example, print how long it took to download a file from an instance), there's `steady_clock` for that, which is my whole point above.

ronaldfw: Ah, I see. That makes sense. Then as you say, let's rename appropriately to make the use case very clear.

pierricgimmig:
@dpallotti , @dimitry- , @ronaldfw , redirecting discussion to the internal doc. In the meantime, since this fixes the issue we have in the introspection window on Windows, I think we should submit it. I changed the clock from high_precision_timer to steady_clock and tested the change. Please take another look.

dimitry- (approved):

ronaldfw:
Just to confirm post-merge: Thanks, agree to make a minimal change here for now to fix the issue on Windows. LGTM!

PR #1773, opened 2021-02-03
Fix always-greyed-out manually instrumented scopes

pierricgimmig:
Filtering for manually instrumented scopes is not yet supported. Make sure all manually 
instrumented scopes are never greyed out. http://b/177968275.

### DIFF
--- a/src/OrbitGl/App.cpp
+++ b/src/OrbitGl/App.cpp
@@ -1445,8 +1445,23 @@ void OrbitApp::SetVisibleFunctionIds(absl::flat_hash_set<uint64_t> visible_funct
   NeedsRedraw();
 }
 
-[[nodiscard]] bool OrbitApp::IsFunctionVisible(uint64_t function_address) {
-  return data_manager_->IsFunctionVisible(function_address);
+bool OrbitApp::IsFunctionVisible(uint64_t function_address) {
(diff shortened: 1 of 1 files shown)


src/OrbitGl/App.cpp
```diff
@@ -1445,8 +1445,23 @@ void OrbitApp::SetVisibleFunctionIds(absl::flat_hash_set<uint64_t> visible_funct
...
 }
 
-[[nodiscard]] bool OrbitApp::IsFunctionVisible(uint64_t function_address) {
-  return data_manager_->IsFunctionVisible(function_address);
+bool OrbitApp::IsFunctionVisible(uint64_t function_address) {
+  if (data_manager_->IsFunctionVisible(function_address)) {
+    return true;
+  }
+
+  // Filtering for manually instrumented scopes is not yet supported. All "Orbit" functions
```
florian-kuebler, line 1453: Nit: Can you add a bug to track this, and convert this into an `TODO (b/...): ....`?

florian-kuebler (approved):
LGTM, thanks!

src/OrbitGl/App.cpp
pierricgimmig: Done.

PR #1775, opened 2021-02-03
Fix crash in ScopedStatus' destructor

beckerhe:
ScopedStatus used to hold a raw pointer to MainThreadExecutor, which
means it's the creators responsibility to ensure that the
MainThreadExecutor instance stays alive as long as the ScopedStatus
instance is alive.

This doesn't work well with the use-case of ScopedStatus. It is meant to
be used in background threads, which means the creator cannot ensure
that the MainThreadExecutor instance still exists, when the ScopedStatus
object is destructed - at least not without bigger changes to the
codebase.

This commit fixes the problem by making ScopedStatus keep a weak_ptr of
MainThreadExecutor. Whenever it is meant to schedule an action it will
first check whether a lock can still be acquired. If not the action will
be ignored.

Note, it's still the users responsiblity to keep the MainThreadExecutor
alive to make ScopeStatus work properly.

### DIFF
--- a/src/OrbitGl/App.cpp
+++ b/src/OrbitGl/App.cpp
@@ -1671,7 +1671,7 @@ bool OrbitApp::IsCapturing() const {
 ScopedStatus OrbitApp::CreateScopedStatus(const std::string& initial_message) {
   CHECK(std::this_thread::get_id() == main_thread_id_);
   CHECK(status_listener_ != nullptr);
-  return ScopedStatus{GetMainThreadExecutor(), status_listener_, initial_message};
+  return ScopedStatus{GetMainThreadExecutor()->weak_from_this(), status_listener_, initial_message};
 }
 
 void OrbitApp::SelectTracepoint(const TracepointInfo& tracepoint) {
--- a/src/OrbitGl/ScopedStatus.h
+++ b/src/OrbitGl/ScopedStatus.h
@@ -36,10 +36,10 @@
 class ScopedStatus final {
  public:
   ScopedStatus() = default;
-  explicit ScopedStatus(MainThreadExecutor* main_thread_executor, StatusListener* status_listener,
-                        const std::string& status_message) {
+  explicit ScopedStatus(std::weak_ptr<MainThreadExecutor> main_thread_executor,
+                        StatusListener* status_listener, const std::string& status_message) {
     data_ = std::make_unique<Data>();
-    data_->main_thread_executor = main_thread_executor;
+    data_->main_thread_executor = std::move(main_thread_executor);
     data_->status_listener = status_listener;
     data_->main_thread_id = std::this_thread::get_id();
     data_->status_id = status_listener->AddStatus(status_message);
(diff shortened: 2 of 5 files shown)


src/OrbitGl/App.cpp
```diff
@@ -1154,8 +1154,9 @@ void OrbitApp::LoadSymbols(const std::filesystem::path& symbols_path, ModuleData
                                          module_data->file_path(), symbols_path.string()));
   thread_pool_->Schedule([this, scoped_status = std::move(scoped_status), symbols_path, module_data,
                           function_hashes_to_hook = std::move(function_hashes_to_hook),
-                          frame_track_function_hashes =
-                              std::move(frame_track_function_hashes)]() mutable {
+                          frame_track_function_hashes = std::move(frame_track_function_hashes),
+                          main_thread_executor =
+                              main_thread_executor_->weak_from_this()]() mutable {
```
dimitry-, line 1159: looks like main_thread_executor is not used in lambda.

beckerhe: Hmmm. You're right. This is from a different PR. Let me quickly fix that!

beckerhe: Done. Thanks for spotting this!

src/OrbitGl/ScopedStatus.h
```diff
@@ -67,10 +67,10 @@ class ScopedStatus final {
...
       data_->status_listener->UpdateStatus(data_->status_id, message);
     } else {
-      data_->main_thread_executor->Schedule(
-          [status_id = data_->status_id, status_listener = data_->status_listener, message] {
-            status_listener->UpdateStatus(status_id, message);
-          });
+      auto executor = data_->main_thread_executor.lock();
+      if (executor == nullptr) return;
+      executor->Schedule([status_id = data_->status_id, status_listener = data_->status_listener,
+                          message] { status_listener->UpdateStatus(status_id, message); });
```
dimitry-, line 73: Looking at this I think it might be a good idea to introduce a function like this:
template <typename F>
bool TrySchedule(std::weak_ptr<MainThreadExecutor> executor, F&& functor);

dimitry- (approved):
otherwise lgtm

src/OrbitGl/ScopedStatus.h
beckerhe: Yes, I like the idea. On it. And I would make it nodiscard, WDYT?

beckerhe:
```
template <typename F>
std::optional<orbit_base::Future<T>> TrySchedule(std::weak_ptr<MainThreadExecutor> executor, F&& functor);
```

The resulting type should probably be an optional future. And I take the nodiscard back.

dimitry-: makes sense.

beckerhe: I added the function, including tests in MainThreadExecutorImplTest.cpp

src/OrbitGl/ScopedStatus.h
```diff
@@ -83,11 +82,10 @@ class ScopedStatus final {
...
     } else {
-      auto executor = data_->main_thread_executor.lock();
-      if (executor == nullptr) return;
-      executor->Schedule([status_listener = data_->status_listener, status_id = data_->status_id] {
-        status_listener->ClearStatus(status_id);
-      });
+      TrySchedule(data_->main_thread_executor,
+                  [status_listener = data_->status_listener, status_id = data_->status_id] {
+                    status_listener->ClearStatus(status_id);
+                  });
```
dimitry-, line 88: this one looks different, you had a return here before. I guess this is correct now because we still want to reset data_

beckerhe: Yes, you're right. My original solution was wrong.

beckerhe: Glad that we talked about it. :sweat_smile:

PR #1776, opened 2021-02-03
Make LoadSymbol use weak_ptr of MainThreadExecutor

beckerhe:
OrbitApp::LoadSymbol (and others) schedule background jobs for
performing expensive tasks.
These background jobs used to capture raw pointers to a
MainThreadExecutor instance, which is accessed at the end of each
background job.

The MainThreadExecutor instance might have been destructed when the
background task reaches its end. That usually happens when the user
closes the main window while the background task is still ongoing.

To avoid the resulting crash, the raw pointer is replaced by a weak_ptr
which allows the background task to check whether the MainThreadExecutor
instance is still alive.

This is a temporary fix, since capturing the main thread executor won't
be necessary anymore with the introduction of Futures.

Bug: http://b/177021435

### DIFF
--- a/src/OrbitGl/App.cpp
+++ b/src/OrbitGl/App.cpp
@@ -1008,7 +1008,8 @@ void OrbitApp::LoadModuleOnRemote(ModuleData* module_data,
       [this, module_data, function_hashes_to_hook = std::move(function_hashes_to_hook),
        frame_track_function_hashes = std::move(frame_track_function_hashes),
        scoped_status = std::move(scoped_status),
-       error_message_from_local = std::move(error_message_from_local)]() mutable {
+       error_message_from_local = std::move(error_message_from_local),
+       main_thread_executor = main_thread_executor_->weak_from_this()]() mutable {
         const auto result = GetProcessManager()->FindDebugInfoFile(module_data->file_path());
 
         if (!result) {
@@ -1017,7 +1018,8 @@ void OrbitApp::LoadModuleOnRemote(ModuleData* module_data,
               absl::StrFormat("Did not find symbols locally or on remote for module \"%s\": %s\n%s",
                               module_data->file_path(), error_message_from_local,
                               result.error().message()));
-          main_thread_executor_->Schedule([this, module_data]() {
+
+          TrySchedule(main_thread_executor, [this, module_data]() {
             modules_currently_loading_.erase(module_data->file_path());
           });
           return;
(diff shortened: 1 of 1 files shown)


src/OrbitGl/App.cpp
```diff
@@ -1017,17 +1018,23 @@ void OrbitApp::LoadModuleOnRemote(ModuleData* module_data,
               absl::StrFormat("Did not find symbols locally or on remote for module \"%s\": %s\n%s",
                               module_data->file_path(), error_message_from_local,
                               result.error().message()));
-          main_thread_executor_->Schedule([this, module_data]() {
-            modules_currently_loading_.erase(module_data->file_path());
-          });
+
+          auto executor = main_thread_executor.lock();
+          if (executor != nullptr) {
```
florian-kuebler, line 1023: Shouldn't you also invert that check and do an early return, as below?

florian-kuebler (approved):
Thanks for fixing this. As said the only nit comment is that I would prefer the early exit in all the cases.

src/OrbitGl/App.cpp
beckerhe: I debated that with myself. Because If I return early I have the `return` twice and only a single statement in between. What do you prefer?

dimitry-: You are probably going to replace this with TrySchedule anyways?

florian-kuebler: > I debated that with myself. Because If I return early I have the `return` twice and only a single statement in between. What do you prefer?

I have a slight preference for the early return. But I guess both have valid pros and cons.

beckerhe: BTW. I agreed with Dimitry on the other branch to introduce a function `TrySchedule` which takes a weak_ptr to an Executor and a function object. With this function, the whole lock-and-check procedure will go away. I want to rebase this PR, when the other one is merged. So this discussion might eventually be obsolete.

beckerhe: I will replace it when the other branch is merged.

src/OrbitGl/App.cpp
```diff
@@ -1174,7 +1178,9 @@ void OrbitApp::LoadSymbols(const std::filesystem::path& symbols_path, ModuleData
     scoped_status.UpdateMessage(message);
     LOG("%s", message);
 
-    main_thread_executor_->Schedule(
+    auto executor = main_thread_executor.lock();
```
florian-kuebler, line 1181: Can't you also use TrySchedule in this case?

florian-kuebler: Ah already fixed. Thanks.

florian-kuebler:
LGTM. Thanks!

PR #1777, opened 2021-02-03
Limit lifetime of MainThreadExecutor actions

beckerhe:
The lifetime of a main thread executor action used to be bound to the lifetime of Qt's QApplication object.
And that's literally the last object that gets destructed in the life of a Orbit process.

But almost all actions require access to resources living in OrbitMainWindow or OrbitApp. These actions
will operate on dangling pointers if executed after OrbitMainWindow has been destructed, leading to a
variety of crashes.

The best solution is to bind the lifetime of an action to the executor itself. Thus it's up to the user of
MainThreadExecutor to maintain its lifetime and the lifetime of its actions.

This fixes a bunch of different crashes which appear when background tasks are still running while
Orbit is shutting down.

### DIFF
--- a/src/OrbitGl/MainThreadExecutor.h
+++ b/src/OrbitGl/MainThreadExecutor.h
@@ -42,6 +42,13 @@ class MainThreadExecutor : public std::enable_shared_from_this<MainThreadExecuto
   // Schedules the action to be performed on the main thread.
   virtual void Schedule(std::unique_ptr<Action> action) = 0;
 
+  // Schedule schedules the function object `functor` to be executed
+  // on `*this`. The execution occures asynchronously, meaning this
+  // function call will only push the function object to a queue. It
+  // will be picked up by an event loop cycle.
+  //
+  // Note: The function object is only executed if `*this` is still alive when the event loop picks
+  // up the scheduled task.
   template <typename F>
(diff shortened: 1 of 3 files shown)


reichlfl (approved):
Nice catch. Code-wise LGTM and makes sense. Did you check if this could cause any actions that were executed e.g. on shutdown before to no longer be executed, and if this could result in any issues? It's hard to judge what consequences this change may have TBH.

beckerhe:
> Nice catch. Code-wise LGTM and makes sense. Did you check if this could cause any actions that were executed e.g. on shutdown before to no longer be executed, and if this could result in any issues? It's hard to judge what consequences this change may have TBH.

That's a good question. I can't say for sure, but I'm pretty confident that it won't be a problem. I have a fairly high visibility into these part of the code due to the work on the Futures. And I don't see any unfinished work that could influence the next session. In all cases I know of we either mutate the state of objects (objects that don't exist anymore when shutting down Orbit), or we show messages to the users (messages the user is not interested anymore when they already have shut down the application).

karupayun (commented):
Nice Henning!!

There a few nits in the comments of Schedule, but as you already merged, I won't comment there :)

beckerhe:
> Nice Henning!!
> 
> There a few nits in the comments of Schedule, but as you already merged, I won't comment there :)

Ahh. I'm sorry, Pablo, for prematurely merging it.

PR #1778, opened 2021-02-03
Fix crash in CopyFileToLocal

beckerhe:
ServiceDeployManager::CopyFileToLocal starts a local QEventLoop to
actively wait for the background task to finish. Qt exits all locally waiting
QEventLoops when the application requests to finish up. This code path has not
been handled properly which leads to a crash (see mentioned bug).

This commit changes `CopyFileToLocal` to expect being quit by the
application without the background task being finished. That means Orbit
won't crash anymore when there is an ongoing file copy while the user
closes the main window.

Bug: http://b/177021435

### DIFF
--- a/src/OrbitQt/servicedeploymanager.cpp
+++ b/src/OrbitQt/servicedeploymanager.cpp
@@ -24,7 +24,9 @@
 
 #include "Error.h"
 #include "EventLoop.h"
+#include "OrbitBase/Future.h"
 #include "OrbitBase/Logging.h"
+#include "OrbitBase/Promise.h"
 #include "OrbitSsh/AddrAndPort.h"
 #include "OrbitSshQt/ScopedConnection.h"
 #include "OrbitSshQt/SftpChannel.h"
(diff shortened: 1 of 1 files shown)


src/OrbitQt/servicedeploymanager.cpp
```diff
@@ -280,12 +283,18 @@ outcome::result<void> ServiceDeployManager::CopyOrbitServicePackage() {
...
   DeferToBackgroundThreadAndWait(
-      this, [&, source = std::move(source), destination = std::move(destination)]() {
-        result = CopyFileToLocalImpl(source, destination);
+      this, [this, source = std::move(source), destination = std::move(destination),
+             promise = std::move(promise)]() mutable {
+        promise.SetResult(CopyFileToLocalImpl(source, destination));
       });
-  return result;
+
+  if (future.IsFinished()) return future.Get();
```
dimitry-, line 295: personal prefference, you can ignore it:  if (!future ... ) return ErrorMessageOr(...);

dimitry- (approved):

src/OrbitQt/servicedeploymanager.cpp
beckerhe: This is actually also my personal preference. Don't know why it wrote it differently.

beckerhe: Fixed. Thanks.

PR #1779, opened 2021-02-03
Update metrics uploader client

IrinaShkviro:
metrics_uploader_client API was updated in the recent cl, so MetricsUploader should be updated as well.


beckerhe (approved):
LGTM!

PR #1780, opened 2021-02-03
Fix parsing of OpenGL version numbers

ronaldfw:
I incorrectly assumed that tokenizing the OpenGL version string can
be done with just splitting on ".", but this does not yield correct
results in cases like "3.1 Mesa 20.1.8". Since all major and minor
version numbers are single digits for all OpenGL versions released so
far, we just check the 0th and 2nd digit of the string and consider
them as our major and minor versions. This is wrong for version numbers
with more than one digit, but in our check will only cause issues for
OpenGL 10.x and higher (if we ever see that).

Tested: Ran Orbit manually.
Bug: http://b/179247771

### DIFF
--- a/src/OrbitQt/opengldetect.cpp
+++ b/src/OrbitQt/opengldetect.cpp
@@ -20,21 +20,19 @@
 namespace orbit_qt {
 namespace {
 
-// Must have created an OpenGl context before calling this.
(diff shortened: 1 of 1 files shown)


beckerhe (approved):
LGTM!

vickyliu-go4it (approved):
LGTM. Thanks for that!

PR #1781, opened 2021-02-03
Connection dialog overlay

reichlfl:
The transparent overlays where not accessible by the E2E tests, so when loading the process list, a "sleep" was added to make sure the overlay disappeared before clicking the table. This PR changes the parenting of the overlays and removes the sleep from the E2E tests.

Bug: b/177037834

### DIFF
--- a/src/OrbitQt/ConnectToStadiaWidget.ui
+++ b/src/OrbitQt/ConnectToStadiaWidget.ui
@@ -41,7 +41,7 @@
       <number>1</number>
      </property>
      <layout class="QGridLayout" name="gridLayout">
-      <item row="0" column="0">
+      <item row="1" column="0">
        <widget class="QFrame" name="contentFrame">
         <property name="lineWidth">
          <number>0</number>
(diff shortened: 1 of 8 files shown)


danielfenner (approved):
lgtm! Thank for getting rid of the sleep!

src/OrbitQt/ConnectToStadiaWidget.ui
```diff
@@ -205,14 +235,17 @@
                 <property name="visible">
                  <bool>false</bool>
                 </property>
-                <property name="accessibleName">
-                 <string>Overlay</string>
-                </property>
+				<property name="accessibleName">
```
antonrohr, line 238: I assume most of these changes are done by Designer automatically?

This specific whitespace thing here seems to be done by hand? Maybe another save in Designer will format this correctly

antonrohr (approved):

src/OrbitQt/ConnectToStadiaWidget.ui
reichlfl: Thanks for catching that, I think I added that by hand indeed. Saving the file fixed it.

PR #1782, opened 2021-02-03
Fix flaky test of FutureWatcher

beckerhe:
The FutureWatcher.WaitForWithThreadPoolAndTimeout test tries to test
whether the WaitFor call properly handles the given timeout.

For this two competing timers have been started and it should be shown
that WaitFor's shorter timer returns first and leads to a timeout
result, since the background job waits for the longer timer.

This fails from time to time. My guess is that the process is not
scheduled while it's waiting for the timers to finish. And it might not
be scheduled for a long time, allowing both timers to time out.

In this case it depends on Qt's implementation of the event loop
which timeout is handled first.

The solution is to remove the second competing timer and never allow the
background job to finish.


antonrohr (approved):
Looks good and the explanation makes sense. Thank you

PR #1783, opened 2021-02-03
Update MetricsUploader/Result

IrinaShkviro:
Add new result types to the MetricsUploader/Result.

Test: check log events on the server.

### DIFF
--- a/src/MetricsUploader/MetricsUploaderWindows.cpp
+++ b/src/MetricsUploader/MetricsUploaderWindows.cpp
@@ -108,9 +108,11 @@ ErrorMessageOr<MetricsUploader> MetricsUploader::CreateMetricsUploader(std::stri
   // set up connection and create a client
   Result result = setup_connection_addr();
   if (result != kNoError) {
-    Result result = shutdown_connection_addr();
-    if (result != kNoError) {
-      ERROR("Error while closing connection: %s", GetErrorMessage(result));
+    if (result != kCannotOpenConnection) {
+      Result shutdown_result = shutdown_connection_addr();
+      if (shutdown_result != kNoError) {
+        ERROR("Error while closing connection: %s", GetErrorMessage(shutdown_result));
+      }
     }
     FreeLibrary(metrics_uploader_client_dll);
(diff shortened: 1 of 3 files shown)


src/MetricsUploader/MetricsUploaderWindows.cpp
```diff
@@ -108,9 +108,11 @@ ErrorMessageOr<MetricsUploader> MetricsUploader::CreateMetricsUploader(std::stri
   // set up connection and create a client
   Result result = setup_connection_addr();
   if (result != kNoError) {
-    Result result = shutdown_connection_addr();
-    if (result != kNoError) {
-      ERROR("Error while closing connection: %s", GetErrorMessage(result));
+    if (result != kCannotOpenConnection) {
```
beckerhe, line 111: I think these two `if`s can be combined.

beckerhe: And one question: Is `shutdown_connection_addr()` the inverse of `setup_connection_addr()`? If yes, then I would say it's untypical to call shutdown when setup didn't succeed. That should not be necessary.

beckerhe (approved):
Otherwise LGTM

src/MetricsUploader/MetricsUploaderWindows.cpp
IrinaShkviro: There also some additional work in SetupConnection that could fail. Because of it I'm checking the error type and closing connection if the error occurred. 
I've renamed the variable for the ShutdownConnection() result as I used the same name for 2 different results.

IrinaShkviro: Regarding the first comment: did you mean something like
```
if (result != kNoError && result != kCannotOpenConnection) {
  shutdown
}
if (result != kNoError) {
  FreeLibrary()
}
```

beckerhe: Ah I'm sorry. I didn't see the `FreeLibrary`. I guess in that case you solution is fine,

IrinaShkviro: No worries, thank you for the review!

PR #1784, opened 2021-02-03
Fix MetricsUploader

IrinaShkviro:
SetupConnection has one argument - string that will be used as suffix in
the metrics_uploader_client log file.

Removed FreeLibrary call that leads to crash.

### DIFF
--- a/src/MetricsUploader/MetricsUploaderWindows.cpp
+++ b/src/MetricsUploader/MetricsUploaderWindows.cpp
@@ -65,9 +65,12 @@ MetricsUploader::~MetricsUploader() {
       ERROR("Error while closing connection: %s", GetErrorMessage(result));
     }
   }
-  if (nullptr != metrics_uploader_client_dll_) {
-    FreeLibrary(metrics_uploader_client_dll_);
-  }
+  // Here FreeLibrary should be called. However, calling it on go-shared library leads to crash.
+  // This should be revisited as soon as the issue in golang is fixed.
+  //
+  // if (nullptr != metrics_uploader_client_dll_) {
+  //   FreeLibrary(metrics_uploader_client_dll_);
+  // }
 }
 
 ErrorMessageOr<MetricsUploader> MetricsUploader::CreateMetricsUploader(std::string client_name) {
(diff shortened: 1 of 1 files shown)


src/MetricsUploader/MetricsUploaderWindows.cpp
```diff
@@ -16,6 +16,7 @@ namespace orbit_metrics_uploader {
 constexpr const char* kSendLogEventFunctionName = "SendOrbitLogEvent";
 constexpr const char* kSetupConnectionFunctionName = "SetupConnection";
 constexpr const char* kShutdownConnectionFunctionName = "ShutdownConnection";
+constexpr const char* kClientLogFileSuffix = "Orbit";
```
beckerhe, line 19: Can you add some more details to the commit message?
Where is that logfile stored? What is the actual filename? And where does the suffix come into play?

src/MetricsUploader/MetricsUploaderWindows.cpp
```diff
@@ -65,9 +58,6 @@ MetricsUploader::~MetricsUploader() {
       ERROR("Error while closing connection: %s", GetErrorMessage(result));
     }
   }
-  if (nullptr != metrics_uploader_client_dll_) {
-    FreeLibrary(metrics_uploader_client_dll_);
-  }
```
beckerhe, line 69: The code should stay here. FreeLibrary should be commented and a comment should be added explaining why we can't call it.

src/MetricsUploader/MetricsUploaderWindows.cpp
```diff
@@ -20,21 +20,17 @@ constexpr const char* kClientLogFileSuffix = "Orbit";
...
                                  Result (*send_log_event_addr)(const uint8_t*, int),
-                                 Result (*shutdown_connection_addr)(),
-                                 HMODULE metrics_uploader_client_dll)
+                                 Result (*shutdown_connection_addr)())
     : client_name_(std::move(client_name)),
       session_uuid_(std::move(session_uuid)),
       send_log_event_addr_(send_log_event_addr),
-      shutdown_connection_addr_(shutdown_connection_addr),
-      metrics_uploader_client_dll_(metrics_uploader_client_dll) {}
+      shutdown_connection_addr_(shutdown_connection_addr) {}
```
beckerhe, line 27: I believe the HMODULE should stay here, even if we currently don't need it.

This class acts as the owner of that handle and we shouldn't throw it away.

beckerhe (commented):
Thank you Irina. I left some comments. :-)

src/MetricsUploader/MetricsUploaderWindows.cpp
IrinaShkviro: ok, thanks. I've left the HMODULE as a class member

src/MetricsUploader/MetricsUploaderWindows.cpp
IrinaShkviro: Updated the code: commented FreeLibrary calls and add explanation in the destructor as well

src/MetricsUploader/MetricsUploaderWindows.cpp
IrinaShkviro: Updated the commit message

IrinaShkviro:
> Thank you Irina. I left some comments. :-)

Thank you for taking a look! I've updated the code

beckerhe (approved):
I had another look. LGTM!

antonrohr (approved):
LGTM!!!

PR #1785, opened 2021-02-03
Open capture in a new window

vickyliu-go4it:
With this change, when the user clicks "Open Capture" in the capture toolbar, the selected capture file will be opened with a new Orbit instance (with the same command line flags as the current Orbit instance).

Bug: http://b/163303287
Test: Open an Orbit instance from the command line with flags, e.g., `--devmode`, and then select a new capture file from the `OrbitMainWindow` . See the selected capture file is opened in a new window with developer mode enabled.

Further discussion:
The `OrbitMainWindow` becomes useless in two cases (i.e., the user cannot reuse the current window): 1) Open Orbit with "Loading a Capture" and then clear the capture; 2) Opening Orbit with "Loading a Capture" and Orbit fails to open the capture file.
The above cases can be avoided by fixing http://b/179470701 (i.e., remove "Clear Capture" button) and http://b/178071836 (i.e., return to the connection window when failing to load the capture file).

### DIFF
--- a/src/OrbitQt/main.cpp
+++ b/src/OrbitQt/main.cpp
@@ -2,6 +2,7 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
 
+#include <absl/container/flat_hash_set.h>
 #include <absl/flags/declare.h>
 #include <absl/flags/flag.h>
 #include <absl/flags/parse.h>
@@ -75,7 +76,8 @@ using Context = orbit_ssh::Context;
 Q_DECLARE_METATYPE(std::error_code);
 
 void RunUiInstance(const DeploymentConfiguration& deployment_configuration,
-                   const Context* ssh_context) {
+                   const Context* ssh_context, const QStringList& command_line_flags,
+                   const std::filesystem::path& capture_file_path = "") {
   qRegisterMetaType<std::error_code>();
 
   const GrpcPort grpc_port{/*.grpc_port =*/absl::GetFlag(FLAGS_grpc_port)};
@@ -95,15 +97,25 @@ void RunUiInstance(const DeploymentConfiguration& deployment_configuration,
     }
   }
 
+  // If Orbit starts with loading a capture file, we skip ProfilingTargetDialog and create a
+  // FileTarget from capture_file_path. After creating the FileTarget, we reset
+  // skip_profiling_target_dialog as false such that if a user ends the previous session, Orbit
+  // will return to a ProfilingTargetDialog.
+  bool skip_profiling_target_dialog = !capture_file_path.empty();
   while (true) {
     {
-      orbit_qt::ProfilingTargetDialog target_dialog{&ssh_connection_artifacts,
-                                                    std::move(target_config)};
-      target_config = target_dialog.Exec();
+      if (skip_profiling_target_dialog) {
+        target_config = orbit_qt::FileTarget(capture_file_path);
+        skip_profiling_target_dialog = false;
+      } else {
+        orbit_qt::ProfilingTargetDialog target_dialog{&ssh_connection_artifacts,
+                                                      std::move(target_config)};
+        target_config = target_dialog.Exec();
 
-      if (!target_config.has_value()) {
-        // User closed dialog
-        break;
+        if (!target_config.has_value()) {
+          // User closed dialog
+          break;
+        }
       }
     }
 
@@ -120,7 +132,8 @@ void RunUiInstance(const DeploymentConfiguration& deployment_configuration,
       }
 
       OrbitMainWindow w(std::move(target_config.value()),
-                        metrics_uploader.has_value() ? &metrics_uploader.value() : nullptr);
+                        metrics_uploader.has_value() ? &metrics_uploader.value() : nullptr,
+                        command_line_flags);
 
       // "resize" is required to make "showMaximized" work properly.
       w.resize(1280, 720);
@@ -159,6 +172,21 @@ void RunUiInstance(const DeploymentConfiguration& deployment_configuration,
       UNREACHABLE();
     }
   }
+}
+
+// Extract command line flags by filtering the positional arguments out from the command line
+// arguments.
+static QStringList ExtractCommandLineFlags(const std::vector<std::string>& command_line_args,
+                                           const std::vector<char*>& positional_args) {
+  QStringList command_line_flags;
+  absl::flat_hash_set<std::string> positional_arg_set(positional_args.begin(),
+                                                      positional_args.end());
+  for (std::string command_line_arg : command_line_args) {
+    if (!positional_arg_set.contains(command_line_arg)) {
+      command_line_flags << QString::fromStdString(command_line_arg);
+    }
+  }
+  return command_line_flags;
 }
 
 static void StyleOrbit(QApplication& app) {
@@ -260,7 +288,16 @@ static bool DevModeEnabledViaEnvironmentVariable() {
 int main(int argc, char* argv[]) {
   absl::SetProgramUsageMessage("CPU Profiler");
   absl::SetFlagsUsageConfig(absl::FlagsUsageConfig{{}, {}, {}, &orbit_core::GetBuildReport, {}});
-  absl::ParseCommandLine(argc, argv);
+  std::vector<char*> positional_args = absl::ParseCommandLine(argc, argv);
+
+  QString orbit_executable = QString(argv[0]);
+  std::vector<std::string> command_line_args;
+  if (argc > 1) {
+    command_line_args.assign(argv + 1, argv + argc);
+  }
+  QStringList command_line_flags = ExtractCommandLineFlags(command_line_args, positional_args);
+  // Skip program name in positional_args[0].
+  std::vector<std::string> capture_file_paths(positional_args.begin() + 1, positional_args.end());
 
   InitLogFile(Path::GetLogFilePath());
   LOG("You are running Orbit Profiler version %s", orbit_core::GetVersion());
@@ -342,6 +379,14 @@ int main(int argc, char* argv[]) {
     return -1;
   }
 
-  RunUiInstance(deployment_configuration, &context.value());
+  // If more than one capture files are provided, start multiple Orbit instances.
+  for (size_t i = 1; i < capture_file_paths.size(); ++i) {
+    QStringList arguments;
+    arguments << QString::fromStdString(capture_file_paths[i]) << command_line_flags;
+    QProcess::startDetached(orbit_executable, arguments);
+  }
+
+  RunUiInstance(deployment_configuration, &context.value(), command_line_flags,
+                capture_file_paths.empty() ? "" : capture_file_paths[0]);
   return 0;
 }
--- a/src/OrbitQt/orbitmainwindow.cpp
+++ b/src/OrbitQt/orbitmainwindow.cpp
@@ -33,6 +33,7 @@
 #include <QMouseEvent>
 #include <QPixmap>
 #include <QPointer>
+#include <QProcess>
 #include <QProgressDialog>
 #include <QPushButton>
 #include <QSettings>
@@ -143,11 +144,13 @@ void OpenDisassembly(const std::string& assembly, const DisassemblyReport& repor
 }  // namespace
 
 OrbitMainWindow::OrbitMainWindow(orbit_qt::TargetConfiguration target_configuration,
-                                 orbit_metrics_uploader::MetricsUploader* metrics_uploader)
+                                 orbit_metrics_uploader::MetricsUploader* metrics_uploader,
+                                 const QStringList& command_line_flags)
     : QMainWindow(nullptr),
       main_thread_executor_{orbit_qt::MainThreadExecutorImpl::Create()},
       app_{OrbitApp::Create(main_thread_executor_.get(), metrics_uploader)},
       ui(new Ui::OrbitMainWindow),
+      command_line_flags_(command_line_flags),
       target_configuration_(std::move(target_configuration)) {
   SetupMainWindow();
 
@@ -1062,7 +1065,10 @@ void OrbitMainWindow::on_actionOpen_Capture_triggered() {
     return;
   }
 
-  OpenCapture(file.toStdString());
+  QString orbit_executable =
+      QString::fromStdString(orbit_base::GetExecutablePath().generic_string());
+  QStringList arguments;
+  QProcess::startDetached(orbit_executable, arguments << file << command_line_flags_);
 }
 
 void OrbitMainWindow::OpenCapture(const std::string& filepath) {
--- a/src/OrbitQt/orbitmainwindow.h
+++ b/src/OrbitQt/orbitmainwindow.h
@@ -56,7 +56,8 @@ class OrbitMainWindow : public QMainWindow {
   static constexpr int kEndSessionReturnCode = 1;
 
   explicit OrbitMainWindow(orbit_qt::TargetConfiguration target_configuration,
-                           orbit_metrics_uploader::MetricsUploader* metrics_uploader = nullptr);
+                           orbit_metrics_uploader::MetricsUploader* metrics_uploader = nullptr,
+                           const QStringList& command_line_flags = QStringList());
   ~OrbitMainWindow() override;
 
   void RegisterGlWidget(OrbitGLWidget* widget) { gl_widgets_.push_back(widget); }
@@ -172,6 +173,7 @@ class OrbitMainWindow : public QMainWindow {
   std::unique_ptr<OrbitGLWidget> introspection_widget_ = nullptr;
   QFrame* hint_frame_ = nullptr;
   QLabel* target_label_ = nullptr;
+  QStringList command_line_flags_;
 
   // Capture toolbar.
   QIcon icon_start_capture_;


src/OrbitQt/orbitmainwindow.cpp
```diff
@@ -1059,14 +1061,24 @@ void OrbitMainWindow::on_actionOpen_Capture_triggered() {
...
 }
 
-void OrbitMainWindow::OpenCapture(const std::string& filepath) {
-  app_->OnLoadCapture(filepath);
-  setWindowTitle(QString::fromStdString(filepath));
-  UpdateCaptureStateDependentWidgets();
-  FindParentTabWidget(ui->CaptureTab)->setCurrentWidget(ui->CaptureTab);
+void OrbitMainWindow::OpenCapture(const std::string& filepath, bool open_in_new_window) {
+  if (open_in_new_window) {
+    QProcess::startDetached(QString::fromStdString(orbit_executable_path_),
```
reichlfl, line 1073: If I'm not mistaken this will drop all other command line flags, is this intended? Not sure if we really need to care since the flags are only meant for development...

src/OrbitQt/orbitmainwindow.h
```diff
@@ -172,6 +172,7 @@ class OrbitMainWindow : public QMainWindow {
   std::unique_ptr<OrbitGLWidget> introspection_widget_ = nullptr;
   QFrame* hint_frame_ = nullptr;
   QLabel* target_label_ = nullptr;
+  std::string orbit_executable_path_;
```
reichlfl, line 175: I think there's no need to have this as a member, is there? It's used only in one place (`OpenCapture`) and can be determined easily there instead of in the constructor, so this feels a bit like unneeded bloat for the class.

src/OrbitQt/orbitmainwindow.cpp
```diff
@@ -1059,14 +1061,24 @@ void OrbitMainWindow::on_actionOpen_Capture_triggered() {
     return;
   }
 
-  OpenCapture(file.toStdString());
+  const bool starts_with_file_target = (target_configuration_.index() == 2);
+  const bool has_data = app_->HasCaptureData();
+  // If the current OrbitMainWindow starts with a FileTarget and the capture file is cleared, open
+  // new capture file in the current OrbitMainWindow. Otherwise, open capture files in new windows.
+  OpenCapture(file.toStdString(), !starts_with_file_target || has_data);
```
reichlfl, line 1068: Do you think the check for an empty capture could be confusing to the user? I get the intention, but as a user, I may wonder why a capture was suddenly opened in the existing window instead of a new one. Did you consider any alternatives to this?

reichlfl (commented):
Thanks a lot for this. Looks good overall, but please take a look at the points I raised.

src/OrbitQt/orbitmainwindow.cpp
vickyliu-go4it: I'm considering the case that the user start an Orbit instance with "Loading capture" and then he "Clear capture". In such a case, the current `OrbitMainWindow` becomes useless (unless he ends the current session): the user cannot take a new capture, and; "Open capture" will open the selected capture file in a new window. 
I'm not sure whether it is a good idea to treat this case differently. So I put this part in a separate commit.

May be still open capture file in new window, and meanwhile provide a hint to tell the user he needs to end this session to reuse this window?

reichlfl: Right, the rationale that the current window basically becomes useless makes sense, I agree. I wonder if, in case there is still a capture loaded, we want to ask the user if they want to open in a new window or re-use the current one.

We could also think about disabling the "Clear Capture" functionality when a capture is loaded from file. Or more generally, do we even need "Clear Capture" at all? With the current UI and this new window functionality, I can't really see a use-case for it. WDYT?

vickyliu-go4it: I agree to disable the "Clear Capture" functionality completely. If a stadia instance is connected, the user can trigger a new capture to clear the current one. And the user can always "End Session" to reload a new capture file. 

Maybe we should also make the "End Session" button more obvious.

src/OrbitQt/orbitmainwindow.cpp
vickyliu-go4it: I add a new commit to get the other command line flags when opening a new window. But I'm not sure whether my implementation is a good choice.

I just checked that we can not use the return value of `absl::ParseCommandLine(argc, argv)` . The return value will ignore all arguments passed after the flag-terminating delimiter (`--`), see [here](https://abseil.io/docs/cpp/guides/flags#parsing-flags-during-startup) for more details. Hence, if we start with command line flags like `--devmode`, we cannot get it from the returned value of `absl::ParseCommandLine(argc, argv)` .

src/OrbitQt/orbitmainwindow.h
vickyliu-go4it: Changed accordingly. Thanks for point out this!

I was thinking this is computed every time so I added it as a member. But I was also considering that it would be a little bit redundant if added as a member. Thanks for helping me to make the decision!

src/OrbitQt/orbitmainwindow.cpp
vickyliu-go4it: Just realize that another case would be the user fails to load a capture due to the wrong format, and the current window also becomes useless. Fixing http://b/178071836 may help to address the problem.

src/OrbitQt/orbitmainwindow.h
```diff
@@ -140,6 +141,7 @@ class OrbitMainWindow : public QMainWindow {
   void SetupMainWindow();
   void SetupHintFrame();
   void SetupTargetLabel();
+  QStringList FilterCommandLineArgs(std::vector<std::string> command_line_args);
```
reichlfl, line 144: `[[nodiscard]] QStringList FilterCommandLineArgs(const std::vector<std::string>& command_line_args) const;`

src/OrbitQt/orbitmainwindow.h
```diff
@@ -56,7 +56,8 @@ class OrbitMainWindow : public QMainWindow {
   static constexpr int kEndSessionReturnCode = 1;
 
   explicit OrbitMainWindow(orbit_qt::TargetConfiguration target_configuration,
-                           orbit_metrics_uploader::MetricsUploader* metrics_uploader = nullptr);
+                           orbit_metrics_uploader::MetricsUploader* metrics_uploader = nullptr,
+                           std::vector<std::string> command_line_args = {});
```
reichlfl, line 60: Does this need to be a copy of the vector, or would `const std::vector<std::string>&` do as well?

src/OrbitQt/main.cpp
```diff
@@ -62,6 +62,9 @@ ABSL_DECLARE_FLAG(std::string, collector);
 ABSL_DECLARE_FLAG(bool, local);
 ABSL_DECLARE_FLAG(bool, devmode);
 ABSL_DECLARE_FLAG(bool, nodeploy);
+ABSL_FLAG(std::string, path_to_capture_file, "",
+          "Path to the capture file. If empty, skip ProfilingTargetDialog and start "
+          "orbitmainwindow with the given capture file.");
```
reichlfl, line 67: Sorry, I just realize that this should probably be done in a different way. Like this, opening a capture from file only works when the filename is explicitly specified as "-path_to_capture_file", right?

If we enable loading a capture with command line args, we should do it the proper way so it also works when one or multiple .orbit files are dropped onto the .exe in windows, or if .orbit is associated with Orbit and a file is opened by double-click. Windows handles this by simply adding all filenames as unnamed parameters, not sure about Linux.

I don't know if absl can handle this, or if we need to parse the arguments ourselves, but I suggest you adjust the code and filtering of parameters to interpret all unnamed parameters as filenames, spawning multiple Orbit instances if more than one filename is given.

src/OrbitQt/orbitmainwindow.cpp
reichlfl: Sounds good. I think you should bring that up in the Standup today so see if anyone objects to remove the "Clear Capture" button. We could maybe still show it in devmode.

reichlfl (commented):
PTAL at my first comment, I think it would be worth doing this. This will change a few things in the code, sorry that I didn't catch that in the first round...

src/OrbitQt/main.cpp
```diff
@@ -74,8 +74,28 @@ using Context = orbit_ssh::Context;
 
 Q_DECLARE_METATYPE(std::error_code);
 
+void FilterCommandLineArgs(const std::vector<std::string>& command_line_args,
```
antonrohr, line 77: I think its a better idea if this is not done by hand, but handled by abseil. 
the call absl::ParseCommandLine returns a vector of arguments that are not flags.

src/OrbitQt/main.cpp
```diff
@@ -342,6 +373,23 @@ int main(int argc, char* argv[]) {
     return -1;
   }
 
-  RunUiInstance(deployment_configuration, &context.value());
+  std::vector<std::string> command_line_args;
+  if (argc > 1) {
+    command_line_args.assign(argv + 1, argv + argc);
+  }
```
antonrohr, line 379: I fear this is too late for making a copy. `argc` and `argv` are passed into the QApplication constructor in line 308 and might be changed then:

From https://doc.qt.io/qt-5/qapplication.html#QApplication
> Note: argc and argv might be changed as Qt removes command line arguments that it recognizes.

Making a copy earlier should solve that

src/OrbitQt/orbitmainwindow.cpp
vickyliu-go4it: Created http://b/179470701 for removing the "Clear Capture" button.

src/OrbitQt/main.cpp
vickyliu-go4it: Thanks so much for the suggestion.

The purpose of this function is to extract commandline flags and capture file paths from the command line arguments.
I didn't find a good way to using abseil to handle this. The return value of `absl::ParseCommandLine(argc, argv)` will ignore all arguments passed after the flag-terminating delimiter (--), see the [document](https://abseil.io/docs/cpp/guides/flags#parsing-flags-during-startup) and the [test cases](https://git.xkool.org/hw/abseil-cpp/-/blob/2eba343b51e0923cd3fb919a6abd6120590fc059/absl/flags/parse_test.cc#L415). 
Hence, if we start with command line flags like `file_name  --flag_name  flag_value` , then the return value includes both `file_name` and `flag_value`.

Therefore, we cannot get the capture file paths from the return value of `absl::ParseCommandLine(argc, argv)`.

src/OrbitQt/orbitmainwindow.h
vickyliu-go4it: Thanks for pointing out this. I modified this method as `void FilterCommandLineArgs(const std::vector<std::string>& command_line_args, QStringList& command_line_flags, std::vector<std::string>& capture_file_paths)`  and put it in main.cpp.

The new method extracts both capture file paths and the command line flags. If multiple capture file paths are provided, we starts multiple Orbit instance to open the provided capture files.

src/OrbitQt/orbitmainwindow.h
vickyliu-go4it: Thanks for pointing out this mistake. I didn't notice here. Changed to use `const std::vector<std::string>&`.

src/OrbitQt/main.cpp
vickyliu-go4it: Thanks for the suggestion. I removed this flag and also added the `void FilterCommandLineArgs(const std::vector<std::string>& command_line_args, QStringList& command_line_flags, std::vector<std::string>& capture_file_paths)` the method in main.cpp to extract capture file paths and the command line flags.

src/OrbitQt/main.cpp
```diff
@@ -95,15 +112,25 @@ void RunUiInstance(const DeploymentConfiguration& deployment_configuration,
     }
   }
 
+  // If Orbit starts with loading a capture file, we skip ProfilingTargetDialog and create a
+  // FileTarget from capture_file_path. After creating the FileTarget, we reset
+  // skip_profiling_target_dialog as false such that if a user ends the previous sesssion, Orbit
```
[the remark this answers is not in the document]
vickyliu-go4it: Great! I didn't know VSCode has this advantage!

Thanks for pointing out this!

src/OrbitQt/main.cpp
vickyliu-go4it: Nice catch! Thanks so much for pointing out this!

src/OrbitQt/main.cpp
antonrohr: I just investigated this and here is my best take (the result is: we should still use the output of `absl::ParseCommandLine()`)

`--flag_name flag_value` actually does not seem to work in general. 
I observed the following:
`./Orbit --devmode true` and `./Orbit --devmode false` both turn on devmode. 
`./Orbit --devmode=true` and `./Orbit --devmode` turns on devmode, `./Orbit --devmode=false` turns it off.

antonrohr: I observed the following:
```
> Executing task: ./Orbit --devmode false --sampling_rate 500 <

[2021-02-05T15:24:10.425171] [                    OrbitQt/main.cpp:265] length 2
[2021-02-05T15:24:10.425212] [                    OrbitQt/main.cpp:267] arg: ./Orbit
[2021-02-05T15:24:10.425219] [                    OrbitQt/main.cpp:267] arg: false
```
this turns ON devmode and sets the sampling_rate to 500
So it seems to work for non boolean values. The 500 is also not part of the std::vector returned by absl::ParseCommandLine

vickyliu-go4it: > I just investigated this and here is my best take (the result is: we should still use the output of `absl::ParseCommandLine()`)
> 
> `--flag_name flag_value` actually does not seem to work in general.
> I observed the following:
> `./Orbit --devmode true` and `./Orbit --devmode false` both turn on devmode.
> `./Orbit --devmode=true` and `./Orbit --devmode` turns on devmode, `./Orbit --devmode=false` turns it off.

Thanks so much for looking into this 👍 

If  we declare devmode as `ABSL_FLAG(bool, devmode, false, "")`, then `--devmode false` turns the devmode off.

If in the future, we add more flags and use `--flag_name flag_value` intentionally or unintentionally, then we will open an Orbit instance with a `flag_value`.
And we still need to get the command line flags by hand. (Not sure about this but I can not find a abseil method to get all the command line flags.)

vickyliu-go4it: May be we use the return value of `absl::ParseCommandLine()` and check whether the elements included in the return value are paths.

antonrohr: > If we declare devmode as ABSL_FLAG(bool, devmode, false, ""), then --devmode false turns the devmode off.

I dont think this is true. I think it turns devmode on and absl::ParseCommandLine returns ["false"]. I tested every way I thought of. Linux & Windows.

antonrohr: Maybe we should talk about this offline

reichlfl (approved):
It took me a bit to understand the idea behind `ExtractCommandLineFlags` and the construct together with `ExtractCaptureFilePaths`. This would be a lot cleaner if absl would provide a way to re-generate a command line from all currently defined flags, but this does not seem to be the case, so I also couldn't come up with a better solution.

Maybe it would help to rename `ExtractCommandLineFlags` to more closely describe what the function is doing, e.g. `RemovePositionalArguments` or something? But feel free to ignore this...

Apart from that, LGTM, thanks for taking care of this!

src/OrbitQt/main.cpp
```diff
@@ -342,6 +387,15 @@ int main(int argc, char* argv[]) {
     return -1;
   }
 
-  RunUiInstance(deployment_configuration, &context.value());
+  // If more than one capture files are provided, start multiple Orbit instances.
+  QString orbit_executable = QString::fromStdString(argv[0]);
```
antonrohr, line 391: `argv[0]` is not an `std::string`, but `char*`, so it would be better to use the constructor `QString(argv[0])`. 
personal preference: I would move this line up. Probably to the very top, but definitely before `QApplication app(argc, argv);`, just in case that QApplication constructor changes `argv`

src/OrbitQt/main.cpp
```diff
@@ -75,7 +76,8 @@ using Context = orbit_ssh::Context;
 Q_DECLARE_METATYPE(std::error_code);
 
 void RunUiInstance(const DeploymentConfiguration& deployment_configuration,
-                   const Context* ssh_context) {
+                   const Context* ssh_context, const QStringList& command_line_flags,
+                   const std::string& capture_file_path = "") {
```
antonrohr, line 80: Since `orbit_qt::FileTarget(std::filesystem::path)` takes a path anyways:
```suggestion
                   const std::filesystem::path& capture_file_path = "") {
```

src/OrbitQt/main.cpp
```diff
@@ -161,6 +174,32 @@ void RunUiInstance(const DeploymentConfiguration& deployment_configuration,
...
+                                                      positional_args.end());
+  for (std::string command_line_arg : command_line_args) {
+    if (!positional_arg_set.contains(command_line_arg)) {
+      command_line_flags << QString::fromStdString(command_line_arg);
+    }
+  }
+  return command_line_flags;
+}
+
+static std::vector<std::string> ExtractCaptureFilePaths(const std::vector<char*>& positional_args) {
```
antonrohr, line 190: I am happy to discuss this and happy to be convinced otherwise: I would not do this filtering. 

I do not see any benefit to restrict the files that can be opened to .orbit files. But I see on downside: If the user tries to open a different file with Orbit (maybe even in windows explorer via "right-click > Open With..") then Orbit will just ignore this file. 

I think we should not filter for ".orbit", but if we do filter, we should at least display a warning to the user.

antonrohr, line 190: Iff this function stays (and is not removed): 
```suggestion
static std::vector<std::filesystem::path> ExtractCaptureFilePaths(const std::vector<char*>& positional_args) {
```

reichlfl: I was thinking the same and then decided it's ok to leave it like this, but I was very much on the fence about this. Now that Anton also points it out, I'd agree :-)

vickyliu-go4it: Thanks for pointing out this. I didn't notice the case of opening other format files with orbit.

Removed the filtering of file extension.

src/OrbitQt/main.cpp
vickyliu-go4it: Changed accordingly. Thanks!

src/OrbitQt/main.cpp
vickyliu-go4it: Changed accordingly. Thanks so much for pointing out this :-) !

src/OrbitQt/main.cpp
vickyliu-go4it: Removed this function! Thanks for the suggestion!

antonrohr (commented):
Sorry for coming up with this so late in the review process, but I just thought about everything again and played around with it. 
I think "Open Capture" should always open a new instance of Orbit. 

Because:
1. When the "Clear Capture" Button is gone, a new Orbit instance will be started every time anyways.
2. There is a bug with the current implementation. If I start Orbit and load "Cap1.orbit" from file. Then I clear the caption and load "Cap2.orbit". Then the top right still displayes "Cap1.orbit"
3. I would be more comfortable if we don't reuse OrbitMainWindow. Aka I think its a good thing if OrbitMainWindow is properly destructed before a new capture is loaded. There might be some data still left from old capture. (Theoretically there shouldn't)

WDYT?

src/OrbitQt/orbitmainwindow.cpp
```diff
@@ -1059,14 +1062,26 @@ void OrbitMainWindow::on_actionOpen_Capture_triggered() {
...
-  UpdateCaptureStateDependentWidgets();
-  FindParentTabWidget(ui->CaptureTab)->setCurrentWidget(ui->CaptureTab);
+void OrbitMainWindow::OpenCapture(const std::string& filepath, bool open_in_new_window) {
+  if (open_in_new_window) {
+    QString orbit_executable =
+        QString::fromStdString(orbit_base::GetExecutablePath().generic_string());
+    QStringList arguments;
+    QProcess::startDetached(orbit_executable,
+                            arguments << QString::fromStdString(filepath) << command_line_flags_);
+  } else {
```
antonrohr, line 1079: personal preference: I would put this into `OrbitMainWindow::on_actionOpen_Capture_triggered` (no need for that extra parameter `open_in_new_window`)

vickyliu-go4it:
> Sorry for coming up with this so late in the review process, but I just thought about everything again and played around with it.
> I think "Open Capture" should always open a new instance of Orbit.
> 
> Because:
> 
> 1. When the "Clear Capture" Button is gone, a new Orbit instance will be started every time anyways.
> 2. There is a bug with the current implementation. If I start Orbit and load "Cap1.orbit" from file. Then I clear the caption and load "Cap2.orbit". Then the top right still displayes "Cap1.orbit"
> 3. I would be more comfortable if we don't reuse OrbitMainWindow. Aka I think its a good thing if OrbitMainWindow is properly destructed before a new capture is loaded. There might be some data still left from old capture. (Theoretically there shouldn't)
> 
> WDYT?

Thanks for pointing out this. This suggestion is also mentioned by Florian. And I will drop the second commit when I reorganize all the commits.

src/OrbitQt/orbitmainwindow.cpp
```diff
@@ -1059,14 +1062,26 @@ void OrbitMainWindow::on_actionOpen_Capture_triggered() {
     return;
   }
 
-  OpenCapture(file.toStdString());
+  const bool starts_with_file_target = (target_configuration_.index() == 2);
```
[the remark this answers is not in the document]
vickyliu-go4it: This part is removed. Also thanks for letting me know this! I didn't find a good way to get the exact type hold by `std::variant`!

src/OrbitQt/orbitmainwindow.cpp
vickyliu-go4it: Nice catch! And by doing so there is no need to change QString (i.e., `file` in `on_actionOpen_Capture_triggered`) to string and then back to QString. 

Changed accordingly. Thanks!

src/OrbitQt/orbitmainwindow.cpp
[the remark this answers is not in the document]
antonrohr: c+= has so many things. I did not know about std::variant until Henning told me about it in december

antonrohr (approved):
Thank you for incorporating ALL the changes I requested and sorry that it took so long. 
The code is really great now!!!

PR #1786, opened 2021-02-04
User ReaderMutexLock when accessing capture_event_buffer_

dimitry-:


src/Service/ProducerSideServiceImpl.cpp
```diff
@@ -348,7 +348,7 @@ void ProducerSideServiceImpl::ReceiveEventsThread(
 
     switch (request.event_case()) {
       case orbit_grpc_protos::ReceiveCommandsAndSendEventsRequest::kBufferedCaptureEvents: {
-        absl::MutexLock lock{&capture_event_buffer_mutex_};
+        absl::ReaderMutexLock lock{&capture_event_buffer_mutex_};
```
dpallotti, line 351: This is correct and in line with the use of `absl::WriterMutexLock` when setting `capture_event_buffer_`, but note that `AddEvent` is called below and it's not `const`. That's fine because of the internal synchronization of `CaptureEventProducer`, but this needs a comment, it took me a while to understand why this change was correct.

Something like
"capture_event_buffer_mutex_ only protects capture_event_buffer_ as a pointer and not the pointee. But the CaptureEventProducer itself is internally synchronized, hence why it's fine to call AddEvent below while only holding a ReaderMutexLock."

src/Service/ProducerSideServiceImpl.cpp
```diff
@@ -348,7 +348,10 @@ void ProducerSideServiceImpl::ReceiveEventsThread(
 
     switch (request.event_case()) {
       case orbit_grpc_protos::ReceiveCommandsAndSendEventsRequest::kBufferedCaptureEvents: {
-        absl::MutexLock lock{&capture_event_buffer_mutex_};
+        // We use ReaderMutexLock because the mutex guards the value of capture_event_buffer_, it does
+        // not guard calls to AddEvent nor the internal state of the object implementing the interface.
+        // The interface implementation is by itself is thread-safe.
```
dpallotti, line 353:
```suggestion
        // The interface implementation is by itself thread-safe.
```

dpallotti (approved):
Thanks!

PR #1787, opened 2021-02-04
Simplify SelectFunctionsFromHashes

beckerhe:
**Note:** This is part of a sequence of refactoring PRs. In this PR some log message will be removed. They will come back in different places at a later stage.

Both `SelectFunctionsFromHashes` and `EnableFrameTracksFromHashes` used to
call a function called `GetFunctionInfosFromHashes`.

`GetFunctionInfosFromHashes` operates on a vector and returns a vector.
The iterations are independent. This violates the separation of concerns
principle.

This commit removes `GetFunctionsInfosFromHashes` and moves the loop
into the callers.

### DIFF
--- a/src/OrbitGl/App.cpp
+++ b/src/OrbitGl/App.cpp
@@ -1194,21 +1194,12 @@ void OrbitApp::LoadSymbols(const std::filesystem::path& symbols_path, ModuleData
           }
 
           if (!function_hashes_to_hook.empty()) {
-            const auto selection_result =
-                SelectFunctionsFromHashes(module_data, function_hashes_to_hook);
-            if (!selection_result) {
-              LOG("Warning, automated hooked incomplete: %s", selection_result.error().message());
-            }
+            SelectFunctionsFromHashes(module_data, function_hashes_to_hook);
             LOG("Auto hooked functions in module \"%s\"", module_data->file_path());
           }
 
           if (!frame_track_function_hashes.empty()) {
-            const auto frame_track_result =
-                EnableFrameTracksFromHashes(module_data, frame_track_function_hashes);
-            if (!frame_track_result) {
-              LOG("Warning, could not insert frame tracks: %s",
-                  frame_track_result.error().message());
-            }
+            EnableFrameTracksFromHashes(module_data, frame_track_function_hashes);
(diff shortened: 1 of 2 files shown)


src/OrbitGl/App.cpp
```diff
@@ -1218,56 +1209,30 @@ void OrbitApp::LoadSymbols(const std::filesystem::path& symbols_path, ModuleData
...
-    const FunctionInfo* function = module->FindFunctionFromHash(function_hash);
-    if (function == nullptr) {
-      count_missing++;
+void OrbitApp::SelectFunctionsFromHashes(const ModuleData* module,
+                                         const std::vector<uint64_t>& function_hashes) {
+  for (const auto function_hash : function_hashes) {
+    const orbit_client_protos::FunctionInfo* const function_info =
+        module->FindFunctionFromHash(function_hash);
+    if (function_info == nullptr) {
+      ERROR("Could not find function hash %#x in module %s", function_hash, module->file_path());
```
dpallotti, line 1218: Does `%#x` also need `ll` here (I never know)?

dpallotti, line 1218: I would use quotes around the `%s` (also the other function).

dpallotti (approved):
Thanks, much cleaner now! Sorry you had to close the other pull request.

src/OrbitGl/App.cpp
beckerhe, line 1218: @dpallotti I changed the error message as suggested in the other PR. You also suggested to use the `ll` length modifier (`%llx`). That's not necessary with abseil.  Quote from the docs:

```
A length modifer to modify the length of the data type. In StrFormat(), these values are ignored (and not needed, since StrFormat() is type-safe) but are allowed for backwards compatibility
```

beckerhe: Ah race condition.

beckerhe: Just added a comment myself. "ll" is not needed in abseil, only in C's printf.

dpallotti: Ah nice, thanks for explaining!

beckerhe: Yes. Good point. Done.

PR #1788, opened 2021-02-04
Add VerifyEventsInOrder to LinuxTracingIntegrationTest

dpallotti:


dimitry- (approved):
I guess if we break it - we can reevaluate the test.

dpallotti:
> I guess if we break it - we can reevaluate the test.

Well, for now the idea would be not to break the ordering, but yes, we can adapt as needed.

PR #1789, opened 2021-02-04
Remove MetricsUploader libraries from installer ZIP

beckerhe:
We generate a bunch of test libraries that mimic the SDK's metrics
uploader client library. They should not be packaged.

image


IrinaShkviro (approved):
Oh, sorry for this. And thank you so much for the fix!

PR #1791, opened 2021-02-04
Don't package shared objects meant for testing

beckerhe:
We have a bunch of test-only libraries (DLLs or SOs). They all have been
packaged and copied to our GCS bucket.

This commit ensures that only explicitly specified libraries are packaged.

This is the proper fix of #1789


florian-kuebler (approved):
LGTM, thanks and good catch.

PR #1792, opened 2021-02-04
Fix frame-track selection in offline capture

dimitry-:
Test: load offline capture enable and disable frame tracks


ronaldfw (approved):
LGTM, thanks for the fix!

src/OrbitGl/LiveFunctionsDataView.cpp
```diff
@@ -223,8 +223,8 @@ std::vector<std::string> LiveFunctionsDataView::GetContextMenu(
       enable_enable_frame_track |= !app_->IsFrameTrackEnabled(selected_function);
       enable_disable_frame_track |= app_->IsFrameTrackEnabled(selected_function);
     } else {
-      enable_enable_frame_track |= !app_->HasFrameTrackInCaptureData(instrumented_function_id);
-      enable_disable_frame_track |= app_->HasFrameTrackInCaptureData(instrumented_function_id);
```
dimitry-, line 227: This is not strictly necessary - should work without it but I think we should in general move away towards getting data from our data structures when possible.

PR #1794, opened 2021-02-04
Use Ninja for CI builds

beckerhe:
I got tired of the slow build times on the CI, so this commit replaces
MSBuild by Ninja when invoked by the CI.

We can't yet change this out in general, because that would break the
Visual Studio development workflow.

Results:
- Usual build time with MSBuild: 22 mins
- With Ninja: 14 mins

I think that's worthwhile even if it is a bit hacky.

### DIFF
--- a/kokoro/builds/build.sh
+++ b/kokoro/builds/build.sh
@@ -78,6 +78,11 @@ if [ -n "$1" ]; then
   }
   trap cleanup EXIT
 
+  # TODO(http://b/179358697): Temporary Ninja hack - will be removed when we move to Ninja in general
+  if [[ $CONAN_PROFILE == "msvc2017_relwithdebinfo" ]]; then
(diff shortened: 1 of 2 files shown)


IrinaShkviro (approved):
That's great!

kokoro/builds/build.sh
```diff
@@ -78,6 +78,11 @@ if [ -n "$1" ]; then
   }
   trap cleanup EXIT
 
+  # Temporary Ninja hack - will be removed when we move to Ninja in general
```
florian-kuebler, line 81: Should you create a bug for that and add it here as a TODO?

florian-kuebler (approved):
Great, that's pretty good!

kokoro/builds/build.sh
beckerhe: done.

PR #1796, opened 2021-02-04
Make OrbitApp::LoadSymbols return Future

beckerhe:
Replaces the old implementation of `LoadSymbols` by one that is based on futures.

Additionally it adds a temporary overload that preserves the old behaviour.
This temporary overload calls the new function to do its thing and will be removed
in a subsequent PR.

### DIFF
--- a/src/OrbitGl/App.cpp
+++ b/src/OrbitGl/App.cpp
@@ -1149,64 +1149,87 @@ ErrorMessageOr<std::filesystem::path> OrbitApp::FindModuleLocally(
   return FindModuleLocallyImpl(symbol_helper_, module_path, build_id);
 }
 
-void OrbitApp::LoadSymbols(const std::filesystem::path& symbols_path, ModuleData* module_data,
-                           std::vector<uint64_t> function_hashes_to_hook,
-                           std::vector<uint64_t> frame_track_function_hashes) {
-  auto scoped_status =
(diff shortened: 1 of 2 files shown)


src/OrbitGl/App.cpp
```diff
@@ -1149,64 +1149,80 @@ ErrorMessageOr<std::filesystem::path> OrbitApp::FindModuleLocally(
...
 
-          if (!frame_track_function_hashes.empty()) {
-            EnableFrameTracksFromHashes(module_data, frame_track_function_hashes);
-            LOG("Added frame tracks in module \"%s\"", module_data->file_path());
-          }
+  const auto handle_hooks_and_frame_tracks =
+      [function_hashes_to_hook = std::move(function_hashes_to_hook),
+       frame_track_function_hashes = std::move(frame_track_function_hashes), module_data,
+       this](const ErrorMessageOr<void>& result) {
+        if (result.has_error()) return;
```
antonrohr, line 1204: Should this error be logged or displayed to the user?

antonrohr (approved):
Looks good. I think the code is very understandable!

src/OrbitGl/App.cpp
beckerhe: Hmmm. Yes, it should I guess. Good catch!

beckerhe: Done.

PR #1798, opened 2021-02-04
Introduce ProducerCaptureEvent/ClientCaptureEvent

dimitry-:
This change achieves 2 objectives

1. Separate Producer<->Service interface from Service<->Client interface on proto level by splitting CaptureEvent into ClientCaptureEvent and ProducerCaptureEvent.
2. Remove dual intent protos like CallstackSample/AddressInfo and provide 2 variants, one with keys another one with full objects

### DIFF
--- a/src/GrpcProtos/capture.proto
+++ b/src/GrpcProtos/capture.proto
@@ -136,20 +136,30 @@ message FullTracepointEvent {
   TracepointInfo tracepoint_info = 5;
 }
 
-message GpuJob {
+message FullGpuJobEvent {
   int32 pid = 1;
   int32 tid = 2;
   uint32 context = 3;
   uint32 seqno = 4;
-  oneof timeline_or_key {
-    string timeline = 5;
-    uint64 timeline_key = 6;
-  }
-  int32 depth = 7;
-  uint64 amdgpu_cs_ioctl_time_ns = 8;
-  uint64 amdgpu_sched_run_job_time_ns = 9;
-  uint64 gpu_hardware_start_time_ns = 10;
-  uint64 dma_fence_signaled_time_ns = 11;
+  int32 depth = 5;
+  uint64 amdgpu_cs_ioctl_time_ns = 6;
+  uint64 amdgpu_sched_run_job_time_ns = 7;
+  uint64 gpu_hardware_start_time_ns = 8;
+  uint64 dma_fence_signaled_time_ns = 9;
+  string timeline = 10;
+}
+
+message InternedGpuJobEvent {
+  int32 pid = 1;
+  int32 tid = 2;
+  uint32 context = 3;
+  uint32 seqno = 4;
+  int32 depth = 5;
+  uint64 amdgpu_cs_ioctl_time_ns = 6;
+  uint64 amdgpu_sched_run_job_time_ns = 7;
+  uint64 gpu_hardware_start_time_ns = 8;
+  uint64 dma_fence_signaled_time_ns = 9;
+  uint64 timeline_key = 10;
 }
 
 message GpuQueueSubmission {
@@ -263,7 +273,7 @@ message ClientCaptureEvent {
     InternedCallstackSample interned_callstack_sample = 16;
     FunctionCall function_call = 4;
     InternedString interned_string = 5;
-    GpuJob gpu_job = 6;
+    InternedGpuJobEvent interned_gpu_job_event = 6;
     GpuQueueSubmission gpu_queue_submission = 13;
     ThreadName thread_name = 7;
     ThreadStateSlice thread_state_slice = 11;
@@ -283,7 +293,7 @@ message ProducerCaptureEvent {
     FullCallstackSample full_callstack_sample = 4;
     FunctionCall function_call = 5;
     InternedString interned_string = 6;
-    GpuJob gpu_job = 7;
+    FullGpuJobEvent full_gpu_job_event = 7;
     GpuQueueSubmission gpu_queue_submission = 8;
     ThreadName thread_name = 9;
     ThreadStateSlice thread_state_slice = 10;
--- a/src/GrpcProtos/capture.proto
+++ b/src/GrpcProtos/capture.proto
@@ -96,14 +96,18 @@ message InternedCallstack {
   Callstack intern = 2;
 }
 
-message CallstackSample {
+message InternedCallstackSample {
   int32 pid = 1;
   int32 tid = 2;
-  oneof callstack_or_key {
-    Callstack callstack = 3;
-    uint64 callstack_key = 4;
-  }
-  uint64 timestamp_ns = 5;
+  uint64 callstack_id = 3;
+  uint64 timestamp_ns = 4;
+}
+
+message FullCallstackSample {
+  int32 pid = 1;
+  int32 tid = 2;
+  Callstack callstack = 3;
+  uint64 timestamp_ns = 4;
 }
 
 message InternedString {
@@ -246,12 +250,12 @@ message ModuleUpdateEvent {
   ModuleInfo module = 3;
 }
 
-message CaptureEvent {
-  reserved 14;
+message ClientCaptureEvent {
+  reserved 3, 14;
   oneof event {
     SchedulingSlice scheduling_slice = 1;
     InternedCallstack interned_callstack = 2;
-    CallstackSample callstack_sample = 3;
+    InternedCallstackSample interned_callstack_sample = 16;
     FunctionCall function_call = 4;
     InternedString interned_string = 5;
     GpuJob gpu_job = 6;
@@ -262,6 +266,26 @@ message CaptureEvent {
     InternedTracepointInfo interned_tracepoint_info = 9;
     TracepointEvent tracepoint_event = 10;
     IntrospectionScope introspection_scope = 12;
+    ModuleUpdateEvent module_update_event = 15;
+  }
+}
+
+message ProducerCaptureEvent {
+  oneof event {
+    SchedulingSlice scheduling_slice = 1;
+    InternedCallstack interned_callstack = 2;
+    InternedCallstackSample interned_callstack_sample = 3;
+    FullCallstackSample full_callstack_sample = 4;
+    FunctionCall function_call = 5;
+    InternedString interned_string = 6;
+    GpuJob gpu_job = 7;
+    GpuQueueSubmission gpu_queue_submission = 8;
+    ThreadName thread_name = 9;
+    ThreadStateSlice thread_state_slice = 10;
+    AddressInfo address_info = 11;
+    InternedTracepointInfo interned_tracepoint_info = 12;
+    TracepointEvent tracepoint_event = 13;
+    IntrospectionScope introspection_scope = 14;
     ModuleUpdateEvent module_update_event = 15;
   }
 }
--- a/src/GrpcProtos/capture.proto
+++ b/src/GrpcProtos/capture.proto
@@ -120,15 +120,20 @@ message InternedTracepointInfo {
   TracepointInfo intern = 2;
 }
 
-message TracepointEvent {
+message InternedTracepointEvent {
   int32 pid = 1;
   int32 tid = 2;
-  uint64 time = 3;
+  uint64 timestamp_ns = 3;
   int32 cpu = 4;
-  oneof tracepoint_info_or_key {
-    TracepointInfo tracepoint_info = 5;
-    uint64 tracepoint_info_key = 6;
-  }
+  uint64 tracepoint_info_key = 5;
+}
+
+message FullTracepointEvent {
+  int32 pid = 1;
+  int32 tid = 2;
+  uint64 timestamp_ns = 3;
+  int32 cpu = 4;
+  TracepointInfo tracepoint_info = 5;
 }
 
 message GpuJob {
@@ -264,7 +269,7 @@ message ClientCaptureEvent {
     ThreadStateSlice thread_state_slice = 11;
     AddressInfo address_info = 8;
     InternedTracepointInfo interned_tracepoint_info = 9;
-    TracepointEvent tracepoint_event = 10;
+    InternedTracepointEvent interned_tracepoint_event = 10;
     IntrospectionScope introspection_scope = 12;
     ModuleUpdateEvent module_update_event = 15;
   }
@@ -284,8 +289,9 @@ message ProducerCaptureEvent {
     ThreadStateSlice thread_state_slice = 10;
     AddressInfo address_info = 11;
     InternedTracepointInfo interned_tracepoint_info = 12;
-    TracepointEvent tracepoint_event = 13;
-    IntrospectionScope introspection_scope = 14;
-    ModuleUpdateEvent module_update_event = 15;
+    InternedTracepointEvent interned_tracepoint_event = 13;
+    FullTracepointEvent full_tracepoint_event = 14;
+    IntrospectionScope introspection_scope = 15;
+    ModuleUpdateEvent module_update_event = 16;
   }
 }
--- a/src/GrpcProtos/capture.proto
+++ b/src/GrpcProtos/capture.proto
@@ -96,7 +96,7 @@ message InternedCallstack {
   Callstack intern = 2;
 }
 
-message InternedCallstackSample {
+message CallstackSample {
   int32 pid = 1;
   int32 tid = 2;
   uint64 callstack_id = 3;
@@ -120,7 +120,7 @@ message InternedTracepointInfo {
   TracepointInfo intern = 2;
 }
 
-message InternedTracepointEvent {
+message TracepointEvent {
   int32 pid = 1;
   int32 tid = 2;
   uint64 timestamp_ns = 3;
@@ -136,7 +136,7 @@ message FullTracepointEvent {
   TracepointInfo tracepoint_info = 5;
 }
 
-message FullGpuJobEvent {
+message FullGpuJob {
   int32 pid = 1;
   int32 tid = 2;
   uint32 context = 3;
@@ -149,7 +149,7 @@ message FullGpuJobEvent {
   string timeline = 10;
 }
 
-message InternedGpuJobEvent {
+message GpuJob {
   int32 pid = 1;
   int32 tid = 2;
   uint32 context = 3;
@@ -246,7 +246,7 @@ message ThreadStateSlice {
   uint64 end_timestamp_ns = 5;
 }
 
-message InternedAddressInfo {
+message AddressInfo {
   uint64 absolute_address = 1;
   uint64 offset_in_function = 2;
   uint64 function_name_key = 3;
@@ -267,40 +267,41 @@ message ModuleUpdateEvent {
 }
 
 message ClientCaptureEvent {
-  reserved 3, 14;
   oneof event {
-    SchedulingSlice scheduling_slice = 1;
-    InternedCallstack interned_callstack = 2;
-    InternedCallstackSample interned_callstack_sample = 16;
-    FunctionCall function_call = 4;
-    InternedString interned_string = 5;
-    InternedGpuJobEvent interned_gpu_job_event = 6;
-    GpuQueueSubmission gpu_queue_submission = 13;
-    ThreadName thread_name = 7;
-    ThreadStateSlice thread_state_slice = 11;
-    InternedAddressInfo interned_address_info = 8;
-    InternedTracepointInfo interned_tracepoint_info = 9;
-    InternedTracepointEvent interned_tracepoint_event = 10;
-    IntrospectionScope introspection_scope = 12;
-    ModuleUpdateEvent module_update_event = 15;
+    // Please keep these alphabetically ordered.
+    AddressInfo address_info = 1;
+    CallstackSample callstack_sample = 2;
+    FunctionCall function_call = 3;
+    GpuJob gpu_job = 4;
+    GpuQueueSubmission gpu_queue_submission = 5;
+    InternedCallstack interned_callstack = 6;
+    InternedString interned_string = 7;
+    InternedTracepointInfo interned_tracepoint_info = 8;
+    IntrospectionScope introspection_scope = 9;
+    ModuleUpdateEvent module_update_event = 10;
+    SchedulingSlice scheduling_slice = 11;
+    ThreadName thread_name = 12;
+    ThreadStateSlice thread_state_slice = 13;
+    TracepointEvent tracepoint_event = 14;
   }
 }
 
 message ProducerCaptureEvent {
   oneof event {
-    SchedulingSlice scheduling_slice = 1;
-    InternedCallstack interned_callstack = 2;
-    InternedCallstackSample interned_callstack_sample = 3;
-    FullCallstackSample full_callstack_sample = 4;
-    FunctionCall function_call = 5;
-    InternedString interned_string = 6;
-    FullGpuJobEvent full_gpu_job_event = 7;
-    GpuQueueSubmission gpu_queue_submission = 8;
-    ThreadName thread_name = 9;
-    ThreadStateSlice thread_state_slice = 10;
-    FullAddressInfo full_address_info = 11;
-    FullTracepointEvent full_tracepoint_event = 12;
-    IntrospectionScope introspection_scope = 13;
-    ModuleUpdateEvent module_update_event = 14;
+    // Please keep these alphabetically ordered.
+    CallstackSample callstack_sample = 1;
+    FullCallstackSample full_callstack_sample = 2;
+    FullAddressInfo full_address_info = 3;
+    FullGpuJob full_gpu_job = 4;
+    FullTracepointEvent full_tracepoint_event = 5;
+    FunctionCall function_call = 6;
+    GpuQueueSubmission gpu_queue_submission = 7;
+    InternedCallstack interned_callstack = 8;
+    InternedString interned_string = 9;
+    IntrospectionScope introspection_scope = 10;
+    ModuleUpdateEvent module_update_event = 11;
+    SchedulingSlice scheduling_slice = 12;
+    ThreadName thread_name = 13;
+    ThreadStateSlice thread_state_slice = 14;
   }
 }
--- a/src/GrpcProtos/capture.proto
+++ b/src/GrpcProtos/capture.proto
@@ -246,17 +246,18 @@ message ThreadStateSlice {
   uint64 end_timestamp_ns = 5;
 }
 
-message AddressInfo {
+message InternedAddressInfo {
   uint64 absolute_address = 1;
-  oneof function_name_or_key {
-    string function_name = 2;
-    uint64 function_name_key = 3;
-  }
-  uint64 offset_in_function = 4;
-  oneof map_name_or_key {
-    string map_name = 5;
-    uint64 map_name_key = 6;
-  }
+  uint64 offset_in_function = 2;
+  uint64 function_name_key = 3;
+  uint64 module_name_key = 4;
+}
+
+message FullAddressInfo {
+  uint64 absolute_address = 1;
+  uint64 offset_in_function = 2;
+  string function_name = 3;
+  string module_name = 4;
 }
 
 message ModuleUpdateEvent {
@@ -277,7 +278,7 @@ message ClientCaptureEvent {
     GpuQueueSubmission gpu_queue_submission = 13;
     ThreadName thread_name = 7;
     ThreadStateSlice thread_state_slice = 11;
-    AddressInfo address_info = 8;
+    InternedAddressInfo interned_address_info = 8;
     InternedTracepointInfo interned_tracepoint_info = 9;
     InternedTracepointEvent interned_tracepoint_event = 10;
     IntrospectionScope introspection_scope = 12;
@@ -297,7 +298,7 @@ message ProducerCaptureEvent {
     GpuQueueSubmission gpu_queue_submission = 8;
     ThreadName thread_name = 9;
     ThreadStateSlice thread_state_slice = 10;
-    AddressInfo address_info = 11;
+    FullAddressInfo full_address_info = 11;
     InternedTracepointInfo interned_tracepoint_info = 12;
     InternedTracepointEvent interned_tracepoint_event = 13;
     FullTracepointEvent full_tracepoint_event = 14;
--- a/src/GrpcProtos/capture.proto
+++ b/src/GrpcProtos/capture.proto
@@ -299,10 +299,8 @@ message ProducerCaptureEvent {
     ThreadName thread_name = 9;
     ThreadStateSlice thread_state_slice = 10;
     FullAddressInfo full_address_info = 11;
-    InternedTracepointInfo interned_tracepoint_info = 12;
-    InternedTracepointEvent interned_tracepoint_event = 13;
-    FullTracepointEvent full_tracepoint_event = 14;
-    IntrospectionScope introspection_scope = 15;
-    ModuleUpdateEvent module_update_event = 16;
+    FullTracepointEvent full_tracepoint_event = 12;
+    IntrospectionScope introspection_scope = 13;
+    ModuleUpdateEvent module_update_event = 14;
   }
 }
--- a/src/GrpcProtos/include/GrpcProtos/Constants.h
+++ b/src/GrpcProtos/include/GrpcProtos/Constants.h
@@ -5,7 +5,10 @@
 #ifndef GRPC_PROTOS_CONSTANTS_H_
 #define GRPC_PROTOS_CONSTANTS_H_
 
+#include <stdint.h>
+
 namespace orbit_grpc_protos {
+constexpr uint64_t kInvalidInternId = 0;
 constexpr uint64_t kInvalidFunctionId = 0;
 }
 
--- a/src/OrbitCaptureClient/CaptureEventProcessorTest.cpp
+++ b/src/OrbitCaptureClient/CaptureEventProcessorTest.cpp
@@ -38,12 +38,12 @@ using orbit_grpc_protos::FunctionCall;
 using orbit_grpc_protos::GpuCommandBuffer;
 using orbit_grpc_protos::GpuDebugMarker;
 using orbit_grpc_protos::GpuDebugMarkerBeginInfo;
-using orbit_grpc_protos::GpuJob;
 using orbit_grpc_protos::GpuQueueSubmission;
 using orbit_grpc_protos::GpuQueueSubmissionMetaInfo;
 using orbit_grpc_protos::GpuSubmitInfo;
 using orbit_grpc_protos::InternedCallstack;
 using orbit_grpc_protos::InternedCallstackSample;
+using orbit_grpc_protos::InternedGpuJobEvent;
 using orbit_grpc_protos::InternedString;
 using orbit_grpc_protos::InternedTracepointEvent;
 using orbit_grpc_protos::InternedTracepointInfo;
@@ -80,6 +80,8 @@ class MockCaptureListener : public CaptureListener {
               (override));
   MOCK_METHOD(void, OnTracepointEvent, (TracepointEventInfo), (override));
 };
+
+}  // namespace
 
 TEST(CaptureEventProcessor, CanHandleSchedulingSlices) {
   MockCaptureListener listener;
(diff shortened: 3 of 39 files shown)


src/GrpcProtos/capture.proto
```diff
@@ -116,15 +120,21 @@ message InternedTracepointInfo {
   TracepointInfo intern = 2;
 }
 
-message TracepointEvent {
+message TracepointEventInfo {
```
dpallotti, line 123: This proto doesn't stand on its own and its name is very generic. 
Could we get rid of this proto and put its field in `InternedTracepointEvent` and `FullTracepointEvent`? 
This would be consistent with what is done with `InternedCallstackSample` and `FullCallstackSample`.

dpallotti:
I only looked and left a comment on the protos. I think I'll wait until the change is complete for everything else.

src/Service/ProducerEventProcessor.h
```diff
@@ -0,0 +1,32 @@
...
+#include "CaptureEventBuffer.h"
+#include "capture.pb.h"
+
+namespace orbit_service {
+//
+// The implementation of this interface is responsible
+// for processing ProducerCaptureEvents from multiple producers
+// It converts them to ClientCaptureEvents and sends to CaptureEventBuffer
+//
+class ProducerEventProcessor {
```
florian-kuebler, line 19: This class should have a test.

src/Service/LinuxTracingHandler.cpp
```diff
@@ -68,156 +71,69 @@ void LinuxTracingHandler::Stop() {
...
 
 void LinuxTracingHandler::OnSchedulingSlice(SchedulingSlice scheduling_slice) {
-  CaptureEvent event;
+  ProducerCaptureEvent event;
   *event.mutable_scheduling_slice() = std::move(scheduling_slice);
-  capture_event_buffer_->AddEvent(std::move(event));
+  producer_event_processor_->ProcessEvent(kLinuxTracingProducerId, std::move(event));
 }
 
-void LinuxTracingHandler::OnCallstackSample(CallstackSample callstack_sample) {
```
florian-kuebler, line 76: The simplifications in this class are pretty neat!

src/Service/ProducerEventProcessor.cpp
```diff
@@ -0,0 +1,405 @@
...
+using orbit_grpc_protos::InternedTracepointInfo;
+using orbit_grpc_protos::IntrospectionScope;
+using orbit_grpc_protos::ModuleUpdateEvent;
+using orbit_grpc_protos::ProducerCaptureEvent;
+using orbit_grpc_protos::SchedulingSlice;
+using orbit_grpc_protos::ThreadName;
+using orbit_grpc_protos::ThreadStateSlice;
+
+template <typename T>
+class InternedCache final {
```
florian-kuebler, line 38: Nit, maybe add a comment about this class here.

src/Service/ProducerEventProcessor.cpp
```diff
@@ -0,0 +1,405 @@
...
+  absl::flat_hash_map<T, uint64_t> entry_to_id_;
+  absl::Mutex entry_to_id_mutex_;
+};
+
+class ProducerEventProcessorImpl : public ProducerEventProcessor {
+ public:
+  ProducerEventProcessorImpl() = delete;
+  explicit ProducerEventProcessorImpl(CaptureEventBuffer* capture_event_buffer)
+      : capture_event_buffer_{capture_event_buffer},
+        callstack_id_counter_{1},
```
florian-kuebler, line 69: Why are the the "1"s here?

src/Service/ProducerEventProcessor.cpp
```diff
@@ -0,0 +1,405 @@
...
+  // <producer_id, producer_string_id> -> client_string_id and
+  // <producer_id, producer_callstack_id> -> client_callstack_id
+  InternedCache<std::pair<uint64_t, uint64_t>> interned_callstack_id_cache_;
+  InternedCache<std::pair<uint64_t, uint64_t>> interned_string_id_cache_;
+  InternedCache<std::pair<uint64_t, uint64_t>> interned_tracepoint_id_cache_;
+};
+
+void ProducerEventProcessorImpl::ProcessAddressInfo(uint64_t producer_id,
+                                                    AddressInfo* address_info) {
+  // TODO: Split address info into 2 separate messages, one with strings another one with keys.
```
florian-kuebler, line 125: Can you create a bug for this?

src/Service/ProducerEventProcessor.cpp
```diff
@@ -0,0 +1,405 @@
...
+}
+
+void ProducerEventProcessorImpl::ProcessFunctionCall(FunctionCall* function_call) {
+  ClientCaptureEvent event;
+  *event.mutable_function_call() = std::move(*function_call);
+  capture_event_buffer_->AddEvent(std::move(event));
+}
+
+void ProducerEventProcessorImpl::ProcessGpuJob(uint64_t producer_id, GpuJob* gpu_job) {
+  // TODO: Split GpuJob into 2 separate messages, one with string and another one with the key
```
florian-kuebler, line 166: Again, either add a bug id or address this now (or also fine for me, to address this right after this PR).

florian-kuebler (commented):
Thanks a lot for this change. This is going into the right direction. I left some comments.

src/GrpcProtos/capture.proto
dimitry-: this makes it convenient when we translate Full->Interned, there is only one field to move. But yes this is not particularly specific to Tracepoints. Also should we have CPU in CallstackStample?

dpallotti: The convenience of writing one line instead of four in a couple of places at the cost of consistency and clarity is not a reason. Let me say it again:
- it's inconsistent wit callstack samples;
- it doesn't stand on its own;
- the name "info" is abused. E.g., `TracepointInfo` means something completely different.

dpallotti: > Also should we have CPU in CallstackStample?

If it wasn't there I guess not for now, I guess we can always add it.

dpallotti: Ah, are you suggesting to make this proto even more generic to reuse it in other protos? As soon as we have one proto that needs one field more or less we have a problem. And it just doesn't fit with the slices. I really don't see the reason for it.

dimitry-: done.. it is going to be more set(get) calls but I agree it looks much better from the interface perspective.

src/GrpcProtos/capture.proto
```diff
@@ -116,15 +120,21 @@ message InternedTracepointInfo {
...
+  uint64 timestamp_ns = 3;
   int32 cpu = 4;
-  oneof tracepoint_info_or_key {
-    TracepointInfo tracepoint_info = 5;
-    uint64 tracepoint_info_key = 6;
-  }
+}
+
+message InternedTracepointEvent {
+  TracepointEventInfo event_info = 1;
```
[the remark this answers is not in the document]
dimitry-: removed did as it is in Callstacks.

src/OrbitCaptureClient/CaptureEventProcessorTest.cpp
```diff
@@ -370,99 +370,24 @@ TEST(CaptureEventProcessor, CanHandleAddressInfosWithInternedStrings) {
   EXPECT_EQ(actual_address_info.module_path(), interned_map_name->intern());
 }
 
-// TODO(b/174145461): Enable the test as soon as the bug is fixed.
-TEST(DISABLED_CaptureEventProcessor, CanHandleOneTracepointEvent) {
```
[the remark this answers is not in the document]
dimitry-: they test scenarios which no longer exist (possible)

src/Service/ProducerSideServiceImpl.h
```diff
@@ -83,8 +84,11 @@ class ProducerSideServiceImpl final : public orbit_grpc_protos::ProducerSideServ
   } service_state_;
   absl::Mutex service_state_mutex_;
 
-  CaptureEventBuffer* capture_event_buffer_ = nullptr;
-  absl::Mutex capture_event_buffer_mutex_;
+  //  CaptureEventBuffer* capture_event_buffer_ = nullptr;
```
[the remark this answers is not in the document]
dimitry-: right. done.

src/Service/ProducerEventProcessor.h
dimitry-: done.

dimitry-:
ptal

src/Service/ProducerEventProcessor.h
```diff
@@ -0,0 +1,32 @@
...
+ public:
+  ProducerEventProcessor() = default;
+  virtual ~ProducerEventProcessor() = default;
+
+  virtual void ProcessEvent(uint64_t producer_id,
+                            orbit_grpc_protos::ProducerCaptureEvent event) = 0;
+};
+
+std::unique_ptr<ProducerEventProcessor> CreateProducerEventProcessor(
+    CaptureEventBuffer* capture_event_buffer);
```
dpallotti, line 29: By the way, this seems like a very Java-like pattern to me, were all object variables are references. Why can't `ProducerEventProcessorImpl` be in its own file and be instantiated directly, instead of having to always be wrapped in a `unique_ptr`?

src/GrpcProtos/capture.proto
```diff
@@ -120,15 +120,20 @@ message InternedTracepointInfo {
   TracepointInfo intern = 2;
 }
 
-message TracepointEvent {
+message InternedTracepointEvent {
```
dpallotti, line 123: Same problem with naming here as with `InternedCallstackSample`. "Interned" here means something different than "Interned" in `InternedTracepointInfo` or `InternedTracepointInfo`. Either drop "Interned" altogether or change prefix.

src/GrpcProtos/capture.proto
```diff
@@ -136,20 +136,30 @@ message FullTracepointEvent {
...
-  uint64 dma_fence_signaled_time_ns = 11;
+  int32 depth = 5;
+  uint64 amdgpu_cs_ioctl_time_ns = 6;
+  uint64 amdgpu_sched_run_job_time_ns = 7;
+  uint64 gpu_hardware_start_time_ns = 8;
+  uint64 dma_fence_signaled_time_ns = 9;
+  string timeline = 10;
+}
+
+message InternedGpuJobEvent {
```
dpallotti, line 152: By now you should have understood that we should change the "Interned" here.

src/GrpcProtos/capture.proto
```diff
@@ -136,20 +136,30 @@ message FullTracepointEvent {
   TracepointInfo tracepoint_info = 5;
 }
 
-message GpuJob {
+message FullGpuJobEvent {
```
dpallotti, line 139: Do we want the "Event" suffix? Isn't it a bit redundant?

src/OrbitCaptureClient/CaptureEventProcessorTest.cpp
```diff
@@ -335,26 +335,31 @@ TEST(CaptureEventProcessor, CanHandleAddressInfos) {
   EXPECT_EQ(actual_address_info.module_path(), address_info->map_name());
 }
 
+static ClientCaptureEvent CreateKeyAndStringEvent(uint64_t key, std::string str) {
```
dpallotti, line 316: Why not `CreateInternedStringEvent`?

src/GrpcProtos/capture.proto
```diff
@@ -245,18 +245,18 @@ message ThreadStateSlice {
   uint64 duration_ns = 6;
   uint64 end_timestamp_ns = 5;
 }
+message InternedAddressInfo {
```
dpallotti, line 248: Aaand we should rename this too.

src/Service/ProducerEventProcessorTest.cpp
```diff
@@ -0,0 +1,1300 @@
+// Copyright (c) 2021 The Orbit Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+#include <gmock/gmock.h>
```
dpallotti, line 4:
```suggestion
// found in the LICENSE file.

#include <gmock/gmock.h>
```

src/Service/ProducerEventProcessorTest.cpp
```diff
@@ -0,0 +1,1300 @@
...
+
+constexpr float kAlpha1 = 0.1f;
+constexpr float kRed1 = 0.2f;
+constexpr float kGreen1 = 0.3f;
+constexpr float kBlue1 = 0.4f;
+
+constexpr float kAlpha2 = 2.1f;
+constexpr float kRed2 = 2.2f;
+constexpr float kGreen2 = 2.3f;
+constexpr float kBlue2 = 2.4f;
```
dpallotti, line 88: If I get it correctly these are constants and hence have internal linkage, but still, I would avoid the confusion and put them in the anonymous namespace (for me the whole file could be in the anonymous namespace).

src/Service/ProducerEventProcessorTest.cpp
```diff
@@ -0,0 +1,1300 @@
...
+  // We do not expect resulting id to be the same, but we also do not
+  // enforce sequential ids in the test. 0 value is reserved.
+  EXPECT_NE(actual_interned_callstack.key(), 0);
+  ASSERT_EQ(actual_interned_callstack.intern().pcs_size(), 3);
+  EXPECT_EQ(actual_interned_callstack.intern().pcs()[0], 1);
+  EXPECT_EQ(actual_interned_callstack.intern().pcs()[1], 2);
+  EXPECT_EQ(actual_interned_callstack.intern().pcs()[2], 3);
+}
+
+TEST(ProducerEventProcessor, TwoInternedCallstackDifferentProducersSameKey) {
```
dpallotti, line 145:
```suggestion
TEST(ProducerEventProcessor, TwoInternedCallstacksDifferentProducersSameKey) {
```

src/Service/ProducerEventProcessorTest.cpp
```diff
@@ -0,0 +1,1300 @@
...
+  EXPECT_EQ(actual_interned_callstack1.intern().pcs()[2], 3);
+
+  EXPECT_NE(actual_interned_callstack2.key(), 0);
+  ASSERT_EQ(actual_interned_callstack2.intern().pcs_size(), 3);
+  EXPECT_EQ(actual_interned_callstack2.intern().pcs()[0], 1);
+  EXPECT_EQ(actual_interned_callstack2.intern().pcs()[1], 2);
+  EXPECT_EQ(actual_interned_callstack2.intern().pcs()[2], 4);
+}
+
+TEST(ProducerEventProcessor, TwoInternedCallstackDifferentProducersSameIntern) {
```
dpallotti, line 194:
```suggestion
TEST(ProducerEventProcessor, TwoInternedCallstacksDifferentProducersSameIntern) {
```

src/Service/ProducerEventProcessorTest.cpp
```diff
@@ -0,0 +1,1300 @@
...
+
+  const InternedCallstackSample& callstack_sample2 =
+      callstack_sameple_event2.interned_callstack_sample();
+  EXPECT_EQ(callstack_sample2.pid(), kPid2);
+  EXPECT_EQ(callstack_sample2.tid(), kTid2);
+  EXPECT_EQ(callstack_sample2.timestamp_ns(), kTimestampNs2);
+  EXPECT_EQ(callstack_sample2.callstack_id(), interned_callstack2.key());
+}
+
+TEST(ProducerEventProcessor, FullCallstackSampleSameCallstack) {
```
dpallotti, line 493: Plural? Or do you want exactly `FullCallstackSample`?

src/Service/ProducerEventProcessorTest.cpp
```diff
@@ -0,0 +1,1300 @@
...
+
+  const InternedCallstackSample& callstack_sample2 =
+      callstack_sameple_event2.interned_callstack_sample();
+  EXPECT_EQ(callstack_sample2.pid(), kPid2);
+  EXPECT_EQ(callstack_sample2.tid(), kTid2);
+  EXPECT_EQ(callstack_sample2.timestamp_ns(), kTimestampNs2);
+  EXPECT_EQ(callstack_sample2.callstack_id(), interned_callstack1.key());
+}
+
+TEST(ProducerEventProcessor, FullTracepointEventDifferentTracepoints) {
```
dpallotti, line 558: Plural? Or do you want exactly `FullTracepointEvent`?

src/Service/ProducerEventProcessorTest.cpp
```diff
@@ -0,0 +1,1300 @@
...
+
+  const InternedTracepointEvent& interned_tracepoint_event2 =
+      tracepoint_event2.interned_tracepoint_event();
+  EXPECT_EQ(interned_tracepoint_event2.pid(), kPid2);
+  EXPECT_EQ(interned_tracepoint_event2.tid(), kTid2);
+  EXPECT_EQ(interned_tracepoint_event2.timestamp_ns(), kTimestampNs2);
+  EXPECT_EQ(interned_tracepoint_event2.tracepoint_info_key(), interned_tracepoint_info2.key());
+}
+
+TEST(ProducerEventProcessor, FullTracepointEventSameTracepoint) {
```
dpallotti, line 626: Same here: plural or is it singular on purpose?

dpallotti (commented):
Very nice!

I haven't looked at `ProducerEventProcessorTest.cpp` too much, I'll do it tomorrow.

The main problem I have is with the naming of the new protos that start with "Interned". I left a few comments, but in general I think it's incorrect and in any case inconsistent with other `Interned...` protos that actually carry an intern.

src/GrpcProtos/capture.proto
```diff
@@ -246,12 +250,12 @@ message ModuleUpdateEvent {
...
 
-message CaptureEvent {
-  reserved 14;
+message ClientCaptureEvent {
+  reserved 3, 14;
   oneof event {
     SchedulingSlice scheduling_slice = 1;
     InternedCallstack interned_callstack = 2;
-    CallstackSample callstack_sample = 3;
+    InternedCallstackSample interned_callstack_sample = 16;
```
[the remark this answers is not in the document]
dpallotti: Yes, we most likely will. I would consider re-resetting the ids the last time we break compatibility to get the most out of it, and then we'll have to just accept higher ids.

src/Service/ProducerEventProcessor.h
dpallotti: Yeah, this is certainly a pro, what I don't like of the pattern is that I'm forcibly given a pointer and for example the class can never be stack allocated. So pros and cons, I'm fine with it.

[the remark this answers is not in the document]
dpallotti: I don't like that too much. For me avoiding a free-standing function in favor of one scoped in the corresponding class would have priority on not having a `static` method in an interface, which doesn't have drawbacks. 

The point is that the function is the only way to get an instance of the interface, so you do want it as close a possible to the interface, and you want to find it just by looking at the interface, not by guessing where the function is.

Also, consider if the interface gets renamed. The IDE can rename all occurrences of the class but will obviously not know that it needs to rename `CreateProducerEventProcessor` to `CreateNewClassName`. Instead if the function was `ProducerEventProcessor::Create` it would naturally be renamed to `NewClassName::Create` (as no renaming actually happens).

src/Service/ProducerSideServiceImpl.cpp
```diff
@@ -115,8 +119,12 @@ grpc::Status ProducerSideServiceImpl::ReceiveCommandsAndSendEvents(
 
   // This thread is responsible for reading from stream, and specifically for
   // receiving CaptureEvents and AllEventsSent messages.
```
dpallotti, line 121:
```suggestion
  // receiving ProducerCaptureEvents and AllEventsSent messages.
```

src/OrbitCaptureClient/CaptureEventProcessorTest.cpp
```diff
@@ -479,21 +393,22 @@ TEST(CaptureEventProcessor, CanHandleInternedTracepointEvents) {
   EXPECT_EQ(actual_tracepoint_info.name(), tracepoint_intern->name());
   EXPECT_EQ(actual_tracepoint_event.pid(), tracepoint->pid());
   EXPECT_EQ(actual_tracepoint_event.tid(), tracepoint->tid());
-  EXPECT_EQ(actual_tracepoint_event.time(), tracepoint->time());
+  EXPECT_EQ(actual_tracepoint_event.time(), tracepoint->timestamp_ns());
   EXPECT_EQ(actual_tracepoint_event.cpu(), tracepoint->cpu());
 }
 
 constexpr int32_t kGpuPid = 1;
 constexpr int32_t kGpuTid = 2;
```
dpallotti, line 401: Now that they're out of the anonymous namespace, I'd prefer these to be explicitly static.

src/OrbitCaptureClient/CaptureEventProcessorTest.cpp
```diff
@@ -502,20 +417,19 @@ GpuJob* CreateGpuJob(CaptureEvent* capture_event, uint64_t sw_queue, uint64_t hw
   return gpu_job;
 }
 
+constexpr uint64_t kTimelineKey = 17;
+constexpr const char* kTimelineString = "timeline";
```
dpallotti, line 421: Same here, the internal linkage is not explicit anymore.

dpallotti (commented):
I'm missing a couple of file but I'll send this for now.

src/Service/ProducerEventProcessor.cpp
```diff
@@ -0,0 +1,371 @@
...
+    }
+
+    uint64_t new_id = id_counter_++;
+    auto [unused_it, inserted] = entry_to_id_.insert_or_assign(entry, new_id);
+    CHECK(inserted);
+    return std::make_pair(new_id, true);
+  }
+
+ private:
+  uint64_t id_counter_ = 1;  // 0 is reserved for invalid_id
```
dpallotti, line 60: Do we need to define a constant for the invalid id?

src/Service/ProducerEventProcessor.cpp
```diff
@@ -0,0 +1,371 @@
...
+  void ProcessEvent(uint64_t producer_id, ProducerCaptureEvent event) override;
+
+ private:
+  void ProcessFullAddressInfo(FullAddressInfo* full_address_info);
+  void ProcessFullCallstackSample(FullCallstackSample* full_callstack_sample);
+  void ProcessFunctionCall(FunctionCall* function_call);
+  void ProcessFullGpuJob(FullGpuJob* full_gpu_job_event);
+  void ProcessGpuQueueSubmission(uint64_t producer_id, GpuQueueSubmission* gpu_queue_submission);
+  void ProcessInternedCallstack(uint64_t producer_id, InternedCallstack* interned_callstack);
+  void ProcessInternedCallstackSample(uint64_t producer_id, CallstackSample* callstack_sample);
```
dpallotti, line 80: Rename to `ProcessCallstackSample`.

src/Service/ProducerEventProcessor.cpp
```diff
@@ -0,0 +1,371 @@
...
+  void SendStringAndKeyEvent(uint64_t key, std::string value);
+
+  CaptureEventBuffer* capture_event_buffer_;
+
+  InternedPool<std::vector<uint64_t>> callstack_cache_;
+  InternedPool<std::string> string_cache_;
+  InternedPool<std::pair<std::string, std::string>> tracepoint_cache_;
+
+  // These are mapping InternStrings and InternedCallstacks from producer ids
+  // to client ids:
```
dpallotti, line 98: Not necessarily here with this comment, but I would add a small explanation somewhere of the fact that we re-intern from different producers to a single "key space" before sending interned strings/callstacks to the client.

src/Service/ProducerEventProcessorTest.cpp
```diff
@@ -0,0 +1,1291 @@
...
+  callstack1->add_pcs(2);
+  callstack1->add_pcs(3);
+
+  ProducerCaptureEvent event2;
+  InternedCallstack* interned_callstack2 = event2.mutable_interned_callstack();
+  interned_callstack2->set_key(kProducer2CallstackKey);
+  interned_callstack2->mutable_intern()->CopyFrom(*callstack1);
+
+  ClientCaptureEvent client_capture_event;
+  // We expect only one call here because we have similar callstacks.
```
dpallotti, line 216: "similar"?

src/Service/ProducerEventProcessorTest.cpp
```diff
@@ -0,0 +1,1291 @@
...
+  EXPECT_CALL(buffer, AddEvent).Times(1).WillOnce(SaveArg<0>(&client_capture_event));
+
+  producer_event_processor->ProcessEvent(kDefaultProducerId, event);
+
+  ASSERT_EQ(client_capture_event.event_case(), ClientCaptureEvent::kInternedCallstack);
+  const InternedCallstack& actual_interned_callstack = client_capture_event.interned_callstack();
+
+  // We do not expect resulting id to be the same, but we also do not
+  // enforce sequential ids in the test. 0 value is reserved.
+  EXPECT_NE(actual_interned_callstack.key(), 0);
```
dpallotti, line 139: I think I mentioned somewhere else about adding a constant for the reserved 0. You could then use it in all these tests.

src/Service/ProducerEventProcessorTest.cpp
```diff
@@ -0,0 +1,1291 @@
...
+    EXPECT_EQ(function_call.depth(), kDepth2);
+    EXPECT_EQ(function_call.return_value(), 42);
+    ASSERT_EQ(function_call.registers_size(), 3);
+    EXPECT_EQ(function_call.registers(0), 42);
+    EXPECT_EQ(function_call.registers(1), 21);
+    EXPECT_EQ(function_call.registers(2), 31);
+  }
+}
+
+TEST(ProducerEventProcessor, FullGpuJobEventDifferentTimelines) {
```
dpallotti, line 753:
```suggestion
TEST(ProducerEventProcessor, FullGpuJobsDifferentTimelines) {
```

src/Service/ProducerEventProcessorTest.cpp
```diff
@@ -0,0 +1,1291 @@
...
+    EXPECT_EQ(gpu_job.depth(), kDepth2);
+    EXPECT_EQ(gpu_job.amdgpu_cs_ioctl_time_ns(), kTimestampNs2);
+    EXPECT_EQ(gpu_job.amdgpu_sched_run_job_time_ns(), kTimestampNs2 + 1);
+    EXPECT_EQ(gpu_job.gpu_hardware_start_time_ns(), kTimestampNs2 + 2);
+    EXPECT_EQ(gpu_job.dma_fence_signaled_time_ns(), kTimestampNs2 + 3);
+    EXPECT_EQ(gpu_job.timeline_key(), interned_string2.interned_string().key());
+  }
+}
+
+TEST(ProducerEventProcessor, FullGpuJobEventSameTimeline) {
```
dpallotti, line 848:
```suggestion
TEST(ProducerEventProcessor, FullGpuJobsSameTimeline) {
```

src/Service/ProducerEventProcessor.cpp
```diff
@@ -0,0 +1,371 @@
...
+
+  ClientCaptureEvent event;
+  *event.mutable_callstack_sample() = std::move(*callstack_sample);
+  capture_event_buffer_->AddEvent(std::move(event));
+}
+
+void ProducerEventProcessorImpl::ProcessInternedString(uint64_t producer_id,
+                                                       InternedString* interned_string) {
+  CHECK(!producer_interned_string_id_to_client_string_id_.contains(
+      {producer_id, interned_string->key()}));
```
dpallotti, line 235: I'm skeptical about this `CHECK`. What if the producer, possibly a different process, makes a mistake and sends an intern twice? Should we really crash? The same for `ProcessInternedCallstack`.

src/Service/ProducerEventProcessor.cpp
```diff
@@ -0,0 +1,371 @@
...
+  *event.mutable_interned_callstack() = std::move(*interned_callstack);
+  capture_event_buffer_->AddEvent(std::move(event));
+}
+
+void ProducerEventProcessorImpl::ProcessInternedCallstackSample(uint64_t producer_id,
+                                                                CallstackSample* callstack_sample) {
+  // translate producer id to client id
+  auto it = producer_interned_callstack_id_to_client_callstack_id_.find(
+      {producer_id, callstack_sample->callstack_id()});
+  CHECK(it != producer_interned_callstack_id_to_client_callstack_id_.end());
```
dpallotti, line 224: Should we crash if something went wrong in the producer (possibly a different process) and didn't send the intern in advance?

dpallotti (commented):
Some naming stuff.

I'm not too convinced about the `CHECK`s in `ProducerEventProcessorImpl` now that we support "external" producers. It means that one producer can crash the service and stop everything.

src/Service/ProducerEventProcessor.cpp
dimitry-: I do not think we can produce valid capture data in this case.

src/Service/ProducerEventProcessor.cpp
dimitry-: I was debating it. we can safely ignore it here, but in this case we won't know if producer starts sending duplicates. Just logging such event won't really have an effect, as a general rule - nobody reads logs. this is why I landed on the side of CHECKing this.

src/Service/ProducerEventProcessor.cpp
dpallotti: Well, we could just drop the callstack and move on. I know a crash is effectively the only way to see this problem, but as we might depend on external processes it feels wrong to crash.

src/GrpcProtos/capture.proto
```diff
@@ -246,22 +266,40 @@ message ModuleUpdateEvent {
...
-    ThreadName thread_name = 7;
-    ThreadStateSlice thread_state_slice = 11;
-    AddressInfo address_info = 8;
-    InternedTracepointInfo interned_tracepoint_info = 9;
-    TracepointEvent tracepoint_event = 10;
-    IntrospectionScope introspection_scope = 12;
-    ModuleUpdateEvent module_update_event = 15;
+    CallstackSample callstack_sample = 1;
+    FullAddressInfo full_address_info = 2;
+    FullTracepointEvent full_tracepoint_event = 3;
```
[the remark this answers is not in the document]
dimitry-: fixed.

src/Service/ProducerEventProcessor.cpp
```diff
@@ -0,0 +1,371 @@
...
+using orbit_grpc_protos::IntrospectionScope;
+using orbit_grpc_protos::ModuleUpdateEvent;
+using orbit_grpc_protos::ProducerCaptureEvent;
+using orbit_grpc_protos::SchedulingSlice;
+using orbit_grpc_protos::ThreadName;
+using orbit_grpc_protos::ThreadStateSlice;
+using orbit_grpc_protos::TracepointEvent;
+
+template <typename T>
+class InternedPool final {
```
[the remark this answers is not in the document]
dimitry-: done.

src/Service/ProducerEventProcessor.cpp
dimitry-: we have one for invalid function id, I guess we could do that for all ids, done.

src/Service/ProducerEventProcessor.cpp
```diff
@@ -0,0 +1,371 @@
...
+  void ProcessThreadStateSlice(ThreadStateSlice* thread_state_slice);
+  void ProcessFullTracepointEvent(FullTracepointEvent* full_tracepoint_event);
+
+  void SendStringAndKeyEvent(uint64_t key, std::string value);
+
+  CaptureEventBuffer* capture_event_buffer_;
+
+  InternedPool<std::vector<uint64_t>> callstack_cache_;
+  InternedPool<std::string> string_cache_;
+  InternedPool<std::pair<std::string, std::string>> tracepoint_cache_;
```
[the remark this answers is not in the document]
dimitry-: done.

src/Service/ProducerEventProcessor.cpp
dimitry-: added above (explains what ProcessIntern* functions do.)

src/Service/ProducerEventProcessor.cpp
dimitry-: tbh I still would prefer a crash here, since this is something that is caused by programming error even if was in a different process, but if you feel strongly about this, we can log, ignore and hope someone reads the service logs.

src/Service/ProducerEventProcessor.h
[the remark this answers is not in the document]
dimitry-: moved to interface.

src/Service/ProducerSideServiceImpl.cpp
dimitry-: done.

src/Service/ProducerEventProcessor.cpp
dimitry-: done

src/Service/ProducerEventProcessorTest.cpp
dimitry-: done.

src/Service/ProducerEventProcessorTest.cpp
dimitry-: done.

src/Service/ProducerEventProcessorTest.cpp
dimitry-: done.

src/Service/ProducerEventProcessorTest.cpp
dimitry-: done.

src/Service/ProducerEventProcessor.cpp
dimitry-: having thought about this a little bit more, I now think the better way here is to add a error. If this happens capture stops and returns a error message to user. This way we do not throw capture data up to this point away and user has an reason of what exactly went wrong. We might need producer name here so that error message has more context. I will create a bug.

dimitry-: added todo:

florian-kuebler (approved):
Done with my review. Probably missed some details, but in general looks good to me. Thanks!

src/Service/ProducerEventProcessor.cpp
dpallotti: Alright, it seems that we disagree a bit on this, for me even stopping the capture is kind of wrong if one process misbehaves. But it's ok, at least we are tracking it.

src/GrpcProtos/capture.proto
```diff
@@ -96,7 +96,7 @@ message InternedCallstack {
   Callstack intern = 2;
 }
 
-message InternedCallstackSample {
```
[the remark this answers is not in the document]
dpallotti: Because the intern is the proto `InternedCallstack` right above, this is just referring to the intern, it's super confusing to name both the intern and the proto referring to the intern with the same "Interned" prefix.

src/OrbitCaptureClient/CaptureEventProcessorTest.cpp
dpallotti: This is very minor but still open (yes, GitHub makes it hard to find).

src/OrbitCaptureClient/CaptureEventProcessorTest.cpp
dpallotti: Same here, very minor but still open.

dpallotti (approved):
I would consider renaming `kInvalidId` to something more reflective of the id it's referring to (unless we want to have a single "invalid id" for everything, which is currently not the case). Other than that it looks good.

src/Service/ProducerEventProcessor.cpp
```diff
@@ -0,0 +1,377 @@
...
+
+ private:
+  void ProcessFullAddressInfo(FullAddressInfo* full_address_info);
+  void ProcessFullCallstackSample(FullCallstackSample* full_callstack_sample);
+  void ProcessFunctionCall(FunctionCall* function_call);
+  void ProcessFullGpuJob(FullGpuJob* full_gpu_job_event);
+  void ProcessGpuQueueSubmission(uint64_t producer_id, GpuQueueSubmission* gpu_queue_submission);
+  // ProcessInterned* functions remap producer intern_ids to the id space used in the client
+  // they keep track of these mappings in producer_interned_callstack_id_to_client_callstack_id_
+  // and producer_interned_string_id_to_client_string_id_
```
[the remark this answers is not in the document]
dimitry-: done.

src/GrpcProtos/include/GrpcProtos/Constants.h
```diff
@@ -6,6 +6,7 @@
 #define GRPC_PROTOS_CONSTANTS_H_
 
 namespace orbit_grpc_protos {
+constexpr uint64_t kInvalidId = 0;
```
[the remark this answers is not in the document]
dimitry-: Renamed to kInvalidInternId (we currently use key and id interchangeably I think we should probably rename all keys to ids at some point)

src/GrpcProtos/capture.proto
```diff
@@ -246,22 +266,40 @@ message ModuleUpdateEvent {
...
+    FunctionCall function_call = 6;
+    GpuQueueSubmission gpu_queue_submission = 7;
+    InternedCallstack interned_callstack = 8;
+    InternedString interned_string = 9;
+    IntrospectionScope introspection_scope = 10;
+    ModuleUpdateEvent module_update_event = 11;
+    SchedulingSlice scheduling_slice = 12;
+    ThreadName thread_name = 13;
+    ThreadStateSlice thread_state_slice = 14;
   }
```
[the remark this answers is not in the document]
dimitry-: done, thanks for reminding, somehow forgot to do that.

src/OrbitCaptureClient/CaptureEventProcessorTest.cpp
dimitry-: done.

src/OrbitCaptureClient/CaptureEventProcessorTest.cpp
dimitry-: done.

PR #1800, opened 2021-02-05
Solve the twice frame-track issue (regression)

karupayun:
We aren't updating function_id in FrameTrackOnlineProcessor. So, while
capturing only one frame track appeared (with function_id = 0). When
finishing the capture you can see duplicated frame_tracks. Here we are
adding the frame_track with updated function_id while capturing.

Screenshots: 
While capturing:
during capture

After finishing:
twice frame track

Related with http://b/179462976

Test: Start/Stop a capture. Save/Load a capture.

### DIFF
--- a/src/OrbitGl/FrameTrackOnlineProcessor.cpp
+++ b/src/OrbitGl/FrameTrackOnlineProcessor.cpp
@@ -44,6 +44,7 @@ void FrameTrackOnlineProcessor::ProcessTimer(const orbit_client_protos::TimerInf
     // TID is meaningless for this timer (start and end can be on two different threads).
     constexpr const int32_t kUnusedThreadId = -1;
     frame_timer.set_thread_id(kUnusedThreadId);
+    frame_timer.set_function_id(function_id);
     frame_timer.set_start(previous_timestamp_ns);
     frame_timer.set_end(timer_info.start());
     // We use user_data_key to keep track of the frame number.


dimitry- (approved):
thanks!

ronaldfw (approved):
LGTM, thanks!

src/OrbitGl/FrameTrackOnlineProcessor.cpp
```diff
@@ -44,6 +44,7 @@ void FrameTrackOnlineProcessor::ProcessTimer(const orbit_client_protos::TimerInf
     // TID is meaningless for this timer (start and end can be on two different threads).
     constexpr const int32_t kUnusedThreadId = -1;
     frame_timer.set_thread_id(kUnusedThreadId);
+    frame_timer.set_function_id(function_id);
```
ronaldfw, line 47: I guess an issue here is (my bad!) that this code is in two places and was only updated in one. Should we change this and avoid the duplication? (Not in this PR, of course.)

florian-kuebler (approved):
LGTM. Thanks!

src/OrbitGl/FrameTrackOnlineProcessor.cpp
karupayun: Agree completely! :)

PR #1801, opened 2021-02-05
ProducerSideService and CaptureEventProducer now support CaptureOptions

dpallotti:
Bug: http://b/179461890

Also performed a simple test using the Vulkan layer.

### DIFF
--- a/src/Service/ProducerSideServiceImpl.cpp
+++ b/src/Service/ProducerSideServiceImpl.cpp
@@ -4,7 +4,6 @@
 
 #include "ProducerSideServiceImpl.h"
 
-#include <absl/container/flat_hash_set.h>
 #include <absl/synchronization/mutex.h>
 #include <absl/time/time.h>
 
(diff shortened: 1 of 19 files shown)


src/Service/ProducerSideServiceImpl.cpp
```diff
@@ -53,6 +54,7 @@ void ProducerSideServiceImpl::OnCaptureStopRequested() {
     }
     LOG("About to send CaptureFinishedCommand to CaptureEventProducers (if any)");
     service_state_.capture_status = CaptureStatus::kCaptureFinished;
+    service_state_.capture_options = std::nullopt;
```
florian-kuebler, line 57: Is there a concrete reason why you clear the options here?

florian-kuebler (approved):
LGTM, thanks!

src/Service/ProducerSideServiceImpl.cpp
dpallotti: Hmm, not really, it just felt like the right thing to do, as they are only kept while they could actually be used (i.e., while `service_state_.capture_status != CaptureStatus::kCaptureFinished`).

dimitry- (approved):

src/Service/ProducerSideServiceImpl.cpp
florian-kuebler: Okay, fine for me. Thanks!

PR #1803, opened 2021-02-05
Make OrbitApp::LoadModuleOnRemote return a Future

beckerhe:
The new version of `LoadModuleOnRemote` returns a future and works
asynchronously. An additional temporary overload was added that
calls the new implemenation but polyfills the old behaviour. This
overload will be removed in a later commit.

### DIFF
--- a/src/OrbitGl/App.cpp
+++ b/src/OrbitGl/App.cpp
@@ -993,66 +993,75 @@ void OrbitApp::SendErrorToUi(const std::string& title, const std::string& text) 
   });
 }
 
+orbit_base::Future<ErrorMessageOr<std::filesystem::path>> OrbitApp::LoadModuleOnRemote(
+    const std::string& module_file_path) {
+  ScopedStatus scoped_status = CreateScopedStatus(absl::StrFormat(
+      "Searching for symbols on remote instance for module \"%s\"...", module_file_path));
+
+  orbit_base::Future<ErrorMessageOr<std::string>> check_file_on_remote =
+      thread_pool_->Schedule([process_manager = GetProcessManager(), module_file_path]() {
+        return process_manager->FindDebugInfoFile(module_file_path);
+      });
+
+  auto download_file =
+      [this, module_file_path, scoped_status = std::move(scoped_status)](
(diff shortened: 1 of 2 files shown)


src/OrbitGl/App.h
```diff
@@ -423,6 +423,9 @@ class OrbitApp final : public DataViewFactory, public CaptureListener {
                                                           const std::string& build_id);
   [[nodiscard]] orbit_base::Future<ErrorMessageOr<void>> LoadSymbols(
       const std::filesystem::path& symbols_path, const std::string& module_file_path);
+
+  [[nodiscard]] orbit_base::Future<ErrorMessageOr<std::filesystem::path>> LoadModuleOnRemote(
```
dpallotti, line 427: The other `LoadModuleOnRemote` will be removed but I still find it weird that the two are not next to each other in this intermediate state.

src/OrbitGl/App.cpp
```diff
@@ -993,66 +993,75 @@ void OrbitApp::SendErrorToUi(const std::string& title, const std::string& text)
   });
 }
 
+orbit_base::Future<ErrorMessageOr<std::filesystem::path>> OrbitApp::LoadModuleOnRemote(
+    const std::string& module_file_path) {
+  ScopedStatus scoped_status = CreateScopedStatus(absl::StrFormat(
+      "Searching for symbols on remote instance (module \"%s\")...", module_file_path));
```
dpallotti, line 999: I see this format is the one that was used in the old implementation, but I suggest moving the module name out of the parenthesis, e.g.:
`"Searching for symbols on remote instance for module \"%s\"..."`
It should also be more consistent with other strings in the status bar related to loading modules.

dpallotti (approved):

src/OrbitGl/App.h
beckerhe: Yes, good point. Fixed. Thanks!

src/OrbitGl/App.cpp
beckerhe: Fixed. Thanks!

PR #1804, opened 2021-02-05
Fix crash on ending session during the capture

IrinaShkviro:
Bug: http://b/179474593
Test: start capture, press "end session", confirm => no crash.


antonrohr (approved):
LGTM! Thank you!

PR #1805, opened 2021-02-05
Fix WillCreateDispatchTableAndVulkanLayerProducerAndAdvanceLinkageOnC…

dimitry-:
…reateInstance test

Previously it was assuming that d-tor is called before it was actually
called, and was checking uninitialized variable. This commit removes bogus
check.

Test: run OrbitVulkanLayerTests


florian-kuebler (approved):
LGTM

dimitry-:
ok, I think I have a better idea...

florian-kuebler (approved):
LGTM! Thanks!

PR #1806, opened 2021-02-06
Replace MonotonicTimestampNs() by orbit_base::CaptureTimestampNs()

pierricgimmig:
This PR is split in two commits.

The first one defines Orbit's main Linux clock as "kOrbitTimelineClock" and uses it
for internal timestamping and for events obtained through the perf_event_open api.
This is done to explicitly link perf_event_open's clock to Orbit's timestamping clock.

The second commit puts the timestamping functions in the orbit_base namespace
and removes the "Monotonic" prefix.

This change stems from a discussion in https://github.com/google/orbit/pull/1772.

### DIFF
--- a/src/OrbitBase/include/OrbitBase/Profiling.h
+++ b/src/OrbitBase/include/OrbitBase/Profiling.h
@@ -13,16 +13,23 @@
 #endif
 
 #ifdef _WIN32
+
 [[nodiscard]] inline uint64_t MonotonicTimestampNs() {
   auto time_since_epoch = std::chrono::steady_clock::now().time_since_epoch();
   return std::chrono::duration_cast<std::chrono::nanoseconds>(time_since_epoch).count();
 }
+
 #else
+
+// Linux clock associated with Orbit's main time domain.
+constexpr clockid_t kOrbitTimelineClock = CLOCK_MONOTONIC;
+
 [[nodiscard]] inline uint64_t MonotonicTimestampNs() {
   timespec ts;
-  clock_gettime(CLOCK_MONOTONIC, &ts);
+  clock_gettime(kOrbitTimelineClock, &ts);
   return 1000000000LL * ts.tv_sec + ts.tv_nsec;
 }
+
 #endif
 
 #endif  // ORBIT_BASE_PROFILING_H_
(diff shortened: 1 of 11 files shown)


dimitry- (approved):

dpallotti:
I don't know, for me what I said in the other pull request still applies with no change. This function has different semantics on Windows and Linux, as on Windows it's a generic monotonic timestamp, and on Linux it's a timestamp from a very specific clock for a specific reason. Also the name now is extremely generic (I thought we wanted to make the name more specific, not the opposite?) and this reason is lost even more.

So in the end I still feel the same: This should be a Linux-only function as it serves a specific Linux purpose&mdash;I really don't get why you want this function to have a Windows implementation. If necessary there could be a more generic timestamp function (or not, just use what you need from `std` or `absl`). The very special case of UI introspection (where we want a specific clock on Linux and a generic monotonic clock on Windows) can easily be handled with an `#ifdef` in the introspection code, not here. 

This said, I don't really want to go through another long discussion for this, so if Ronald and Dimitry are fine I guess I'm fine, worst case we change it again.

src/OrbitBase/include/OrbitBase/Profiling.h
```diff
@@ -12,17 +12,28 @@
...
+
+[[nodiscard]] inline uint64_t TimestampNs() {
   auto time_since_epoch = std::chrono::steady_clock::now().time_since_epoch();
   return std::chrono::duration_cast<std::chrono::nanoseconds>(time_since_epoch).count();
 }
+
 #else
-[[nodiscard]] inline uint64_t MonotonicTimestampNs() {
+
+// Linux clock associated with Orbit's main time domain.
```
ronaldfw, line 26: I would add a comment that this needs to be a monotonic clock.

ronaldfw (approved):
LGTM

dpallotti (approved):
Thanks for bearing with me.

src/OrbitBase/include/OrbitBase/Profiling.h
pierricgimmig: Added in the comment above. Thanks for pointing out.

PR #1807, opened 2021-02-08
Introduce GTest::QtGuiMain and GTest::QtCoreMain

beckerhe:
`GTest::QtCoreMain` and `QTest::QtGuiMain` are alternatives to `GTest::Main`.

They provide a main function that initializes GoogleTest, but that also instantiates
a `QCoreApplication` (or `QApplication`) object.

By doing so, qt-based tests can avoid doing that locally.

### DIFF
--- a/cmake/tests.cmake
+++ b/cmake/tests.cmake
@@ -48,3 +48,51 @@ int main(int argc, char* argv[]) {\n\
   target_link_libraries(GTest_Main PUBLIC GTest::GTest)
   add_library(GTest::Main ALIAS GTest_Main)
 endif()
+
+if(NOT TARGET GTest::QtCoreMain AND TARGET Qt5::Core)
+  set(FILE_PATH ${CMAKE_BINARY_DIR}/test_qtcore_main.cpp)
+  if(NOT EXISTS ${FILE_PATH})
+    set(GTEST_MAIN_CPP "\
+#include <cstdio>\n\
+#include <gtest/gtest.h>\n\
+#include <QCoreApplication>\n\
+\n\
+int main(int argc, char* argv[]) {\n\
+  QCoreApplication app{argc, argv}\;\n\
+  printf(\"Running main() from %s\\n\", __FILE__)\;\n\
+  ::testing::InitGoogleTest(&argc, argv)\;\n\
+  return RUN_ALL_TESTS()\;\n\
+}\n\
+")
+
+    file(WRITE ${FILE_PATH} ${GTEST_MAIN_CPP})
+  endif()
+
+  add_library(GTest_QtCoreMain STATIC EXCLUDE_FROM_ALL ${FILE_PATH})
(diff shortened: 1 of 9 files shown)


cmake/tests.cmake
```diff
@@ -48,3 +48,51 @@ int main(int argc, char* argv[]) {\n\
   target_link_libraries(GTest_Main PUBLIC GTest::GTest)
   add_library(GTest::Main ALIAS GTest_Main)
 endif()
+
+if(NOT TARGET GTest::QtCoreMain AND TARGET Qt5::Core)
```
reichlfl, line 52: Not sure I get why "test_qtcore_main.cpp" and "test_qtgui_main.cpp" aren't part of the repo, but created by the makefile. If we stick with this, shouldn't they at least be part of .gitignore so they are not accidentally added to the repo, in which case later changes to the makefile will not have any effect?

cmake/tests.cmake
```diff
@@ -48,3 +48,51 @@ int main(int argc, char* argv[]) {\n\
...
+  ::testing::InitGoogleTest(&argc, argv)\;\n\
+  return RUN_ALL_TESTS()\;\n\
+}\n\
+")
+
+    file(WRITE ${FILE_PATH} ${GTEST_MAIN_CPP})
+  endif()
+
+  add_library(GTest_QtGuiMain STATIC EXCLUDE_FROM_ALL ${FILE_PATH})
+  target_link_libraries(GTest_QtGuiMain PUBLIC GTest::GTest Qt5::Widgets)
```
reichlfl, line 96: If I see this correctly the only difference between `QtCoreMain` and `QtGuiMain` is that the latter links against `Qt5::Widgets`? Interestingly the Tutorials-Test, which instantiates a UI, only requires `QtCoreMain`, so I wonder in which case we'd need `QtGuiMain`?

reichlfl (commented):
Nice, the less friction we have for new tests the better! Overall this makes sense to me, but I'm not sure about the makefile changes, PTAL.

cmake/tests.cmake
beckerhe: They are created in the build-folder which is part of .gitignore in its entirety.

cmake/tests.cmake
beckerhe: It depends on whether you need a QApplication or QCoreApplication object. The former provides everything the latter does and additionally things like keeping track of windows, handling user input (keyboard and mouse events), clipboard support, handling of screens, etc.

It's not needed to instantiate Widgets. But as soon as you want some of the mentioned features you need a QApplication instance over a QCoreApplication instance.

cmake/tests.cmake
reichlfl: Ah ok, that makes sense, thanks for clearing that up.

reichlfl (approved):
Thanks for addressing my questions, LGTM!

beckerhe:
> Thanks for addressing my questions, LGTM!

Always. Thanks a lot for the very quick review!

PR #1808, opened 2021-02-08
Introduce AnyInvocable

beckerhe:
`AnyInvocable` is a function-wrapper, similar to `std::function`. But it doesn't require its wrapped type to copy-constructible.

This PR introduces the type itself (first commit), then uses it in the Promise/Future framework.

### DIFF
--- a/src/OrbitBase/AnyInvocableTest.cpp
+++ b/src/OrbitBase/AnyInvocableTest.cpp
@@ -1,0 +1,49 @@
+// Copyright (c) 2021 The Orbit Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#include <gtest/gtest.h>
+
+#include "OrbitBase/AnyInvocable.h"
+
+namespace orbit_base {
+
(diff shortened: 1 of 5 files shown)


src/OrbitBase/AnyInvocableTest.cpp
```diff
@@ -0,0 +1,47 @@
...
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#include <gtest/gtest.h>
+
+#include "OrbitBase/AnyInvocable.h"
+
+namespace orbit_base {
+
+static int returns_magic_number() { return 42; }
```
dimitry-, line 11: super nitpick: this is not a magic number either change it to the random number: https://xkcd.com/221, or rename this to get_answer_to_life_universe_and_everything()

dimitry- (approved):
LGTM

src/OrbitBase/AnyInvocableTest.cpp
beckerhe: Oh yes. You're so right about this. :smile: Apologies. Fixed!

PR #1809, opened 2021-02-08
Introduce OrbitApp::LoadModule

beckerhe:
There is a function `OrbitApp::LoadModules` which loads a list of modules.

To simplify the process the first step is to introduce a function `OrbitApp::LoadModule` which
only takes care of loading a single module.

Futures and Promises allow to stich together the asynchronous `LoadModule` implementation to a
simpler implementation of `OrbitApp::LoadModules` that has the same behaviour as the function had before.

### DIFF
--- a/src/OrbitBase/include/OrbitBase/Future.h
+++ b/src/OrbitBase/include/OrbitBase/Future.h
@@ -11,6 +11,7 @@
 #include <memory>
 #include <optional>
 #include <type_traits>
+#include <utility>
 #include <variant>
 
 #include "OrbitBase/Logging.h"
@@ -79,8 +80,22 @@ template <typename T>
 class [[nodiscard]] Future : public orbit_base_internal::FutureBase<T> {
  public:
   // Constructs a completed future
+  /* explicit(false) */ Future(const T& val)
+      : orbit_base_internal::FutureBase<T>{
+            std::make_shared<orbit_base_internal::SharedState<T>>()} {
+    this->shared_state_->result.emplace(val);
+  }
+
+  // Constructs a completed future
+  /* explicit(false) */ Future(T && val)
+      : orbit_base_internal::FutureBase<T>{
+            std::make_shared<orbit_base_internal::SharedState<T>>()} {
+    this->shared_state_->result.emplace(std::move(val));
+  }
+
+  // Constructs a completed future
   template <typename... Args>
-  explicit Future(Args && ... args)
+  explicit Future(std::in_place_t, Args && ... args)
       : orbit_base_internal::FutureBase<T>{
             std::make_shared<orbit_base_internal::SharedState<T>>()} {
     this->shared_state_->result.emplace(std::forward<Args>(args)...);
--- a/src/OrbitGl/App.cpp
+++ b/src/OrbitGl/App.cpp
@@ -45,6 +45,9 @@
 #include "ImGuiOrbit.h"
 #include "MainThreadExecutor.h"
 #include "ModulesDataView.h"
+#include "OrbitBase/Future.h"
+#include "OrbitBase/FutureHelpers.h"
+#include "OrbitBase/JoinFutures.h"
 #include "OrbitBase/Logging.h"
 #include "OrbitBase/Result.h"
 #include "OrbitBase/ThreadConstants.h"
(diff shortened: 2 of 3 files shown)


src/OrbitGl/App.cpp
```diff
@@ -1068,43 +1071,91 @@ void OrbitApp::LoadModules(
...
+  const auto it = modules_currently_loading_.find(module_path);
+  if (it != modules_currently_loading_.end()) {
+    return it->second;
   }
+
+  const auto local_symbols_path = FindModuleLocally(module_path, build_id);
+
+  if (local_symbols_path) {
+    return LoadSymbols(local_symbols_path.value(), module_path);
+    // return orbit_base::Future<ErrorMessageOr<std::filesystem::path>>{local_symbols_path};
```
dpallotti, line 1126: This line was left behind?

src/OrbitGl/App.cpp
```diff
@@ -1068,43 +1071,91 @@ void OrbitApp::LoadModules(
...
+  // connected than process_manager != nullptr
+  if (absl::GetFlag(FLAGS_local) || GetProcessManager() == nullptr) {
+    return orbit_base::Future<ErrorMessageOr<void>>{local_symbols_path.error()};
+  }
+
+  auto load_symbols = [this, module_path, local_error_msg = local_symbols_path.error().message()](
+                          const ErrorMessageOr<std::filesystem::path>& result)
+      -> orbit_base::Future<ErrorMessageOr<void>> {
+    if (result.has_error()) {
+      return {ErrorMessage{
```
dpallotti, line 1139: The outer braces are for the constructor of `Future`?

src/OrbitGl/App.cpp
```diff
@@ -1068,43 +1071,91 @@ void OrbitApp::LoadModules(
...
+    }
+
+    return LoadSymbols(result.value(), module_path);
+  };
+
+  auto load_symbols_result =
+      LoadModuleOnRemote(module_path).Then(main_thread_executor_, std::move(load_symbols));
+
+  auto final_result =
+      UnwrapFuture(load_symbols_result)
```
dpallotti, line 1151: I'm a bit lost here. What's `UnwrapFuture` doing? `load_symbols_result` is a `Future` containing a `Future`?

dpallotti (approved):

src/OrbitBase/include/OrbitBase/Future.h
```diff
@@ -78,9 +79,23 @@ enum class FutureRegisterContinuationResult {
 template <typename T>
 class [[nodiscard]] Future : public orbit_base_internal::FutureBase<T> {
  public:
+  // Constructs a completed future
+  /* explicit(false) */ Future(const T& val)
```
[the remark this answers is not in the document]
beckerhe: No, this has not so much to do with std::optional. But std::optional behaves the same, yes. 

I wrote this to make clear that this constructor is intentionally implicit. `explicit(bool)` is a C++20 feature, so at some point I can remove the comment.

src/OrbitGl/App.cpp
beckerhe: Yes, exactly. It's a Future in a Future. That's not ideal, I know.

I already have some code in flight, that specializes `Future<ErrorMessageOr<T>>` and avoids this situation all the way. 

But for now we are still mixing the happy path with the unhappy path. As soon as that's done I can improve this part again.

src/OrbitBase/include/OrbitBase/Future.h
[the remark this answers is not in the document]
dpallotti: > explicit(bool) is a C++20 feature

Ah nice, this is what I was missing.

src/OrbitGl/App.cpp
beckerhe: Technically no. :slightly_smiling_face:  The outer braces are the constructor of ErrorMessageOr<void>, while the construction of the Future is implicit.

Since the compiler won't chain 2 implicit conversions, we have to make one somehow explicit by the braces.

Here applies the same as I mentioned in your comment from below. This will go away when Future can automatically handle the error case and only call continuations if the previous operation had a positive outcome.

src/OrbitGl/App.cpp
beckerhe: Yes, thank you. I will remove it.

src/OrbitGl/App.cpp
dpallotti: Ah right, I mixed up `ErrorMessage` with `ErrorMessageOr`.

PR #1810, opened 2021-02-08
Introduce ImmediateExecutor

beckerhe:
ImmediateExecutor is a helper class to be used in conjunction with
`Future::Then`. It allows to chain a task to be executed whenever the
future completes. The task will be executed in the execution context
that sets the result of the promise.

This type of executor should only be used for tasks that don't depend on
any outside resources since it can't be guaranteed when and where this
task is executed.

### DIFF
--- a/src/OrbitBase/include/OrbitBase/ImmediateExecutor.h
+++ b/src/OrbitBase/include/OrbitBase/ImmediateExecutor.h
@@ -1,0 +1,64 @@
+// Copyright (c) 2021 The Orbit Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#ifndef ORBIT_BASE_IMMEDIATE_EXECUTOR_H_
+#define ORBIT_BASE_IMMEDIATE_EXECUTOR_H_
+
+#include <type_traits>
+
+#include "OrbitBase/Future.h"
+#include "OrbitBase/FutureHelpers.h"
+#include "OrbitBase/Promise.h"
+#include "OrbitBase/PromiseHelpers.h"
+
+namespace orbit_base {
+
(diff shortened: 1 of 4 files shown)


antonrohr (approved):
LGTM! And I am in dire need of this, so thank you!

I guess dimitry should have a look

beckerhe:
> LGTM! And I am in dire need of this, so thank you!
> 
> I guess dimitry should have a look

Thanks. :laughing: I wanted to give Dimitry a break, so just to point out: Not my fault this time. :see_no_evil:

src/OrbitBase/include/OrbitBase/ImmediateExecutor.h
```diff
@@ -0,0 +1,62 @@
...
+// ScheduleAfter call. When called directly it's just a verbose abstraction around a standard
+// function call.
+//
+// Its usage only makes sense in combination with Future::Then.
+//
+// Example:
+// Future<std::string> result = thread_pool_->Schedule(/* whatever */);
+// ImmediateExecutor immediate_executor{};
+// Future<void> void_future = result.Then(&immediate_executor, [](const std::string& str) { (void)
+// str; }); main_thread_executor_->WaitFor(void_future); // WaitFor only works with Future<void>
```
dimitry-, line 27: nit: formatting is off...

dimitry- (approved):
otherwise lgtm

src/OrbitBase/include/OrbitBase/ImmediateExecutor.h
beckerhe: Ah yes. Of course I blame clang-format. :innocent: 

Thanks for spotting this!

beckerhe:
Thank you both for the reviews!

PR #1813, opened 2021-02-08
Make --local connection faster

antonrohr:
Orbit connects faster to a locally running OrbitService. Orbit does not
wait until the state of the local grpc connection is "ready". Instead
the connection is used immediately after setting it up. The local
connection is determined successful, when Orbit receives the first
response from OrbitService.

Adds a tool tip to the "Connecting..." text with more explanation.

Minor Refactoring in ProfilingTargetDialog.cpp

### DIFF
--- a/src/OrbitQt/ProfilingTargetDialog.cpp
+++ b/src/OrbitQt/ProfilingTargetDialog.cpp
@@ -54,7 +54,6 @@ ABSL_DECLARE_FLAG(bool, local);
 
 namespace {
 constexpr int kProcessesRowHeight = 19;
-constexpr std::chrono::duration kLocalTryConnectTimeout = std::chrono::milliseconds{1000};
 }  // namespace
 
 namespace orbit_qt {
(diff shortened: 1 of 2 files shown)


beckerhe:
This is only affecting the way we connect to a local instance, right? Remote instances are not affected by this change, are they?

src/OrbitQt/ProfilingTargetDialog.cpp
```diff
@@ -282,9 +281,14 @@ void ProfilingTargetDialog::SetupLocalStates() {
   state_local_.assignProperty(ui_->stadiaWidget, "active", false);
   state_local_.assignProperty(ui_->loadCaptureRadioButton, "checked", false);
   state_local_.assignProperty(ui_->targetLabel, "text", "");
+  state_local_connecting_.assignProperty(ui_->localProfilingStatusMessage, "toolTip", "");
```
dpallotti, line 284: Is this supposed to be **`state_local_`**`.assignProperty(ui_->localProfilingStatusMessage, "toolTip", "");` instead of `state_local_connecting_`? Or why is this in the `// STATE state_local_` group?

dpallotti (approved):
Thanks a lot!

src/OrbitQt/ProfilingTargetDialog.cpp
antonrohr: Thank you for catching that! It was a mistake on my side. It was meant to be `state_local_`, but I realized its not necessary at all, since the default of property `toolTip` is empty string. Anyways, I fixed it

beckerhe (approved):
LGTM!

PR #1819, opened 2021-02-08
Also process GPU events in order in the "processing" thread; increase kProcessingDelayMs

dpallotti:
#### Make GpuTracepointEventProcessor a PerfEventVisitor
This prepares the commit that follows, which will move the processing of
gpu tracepoint events from the thread reading from the ring buffers to the
"processing" thread.
#### Process GPU tracepoint events on the ProcessDeferredEvents thread
With this change, all `perf_event_open` events are now processed on a separate
thread than the one that reads from the ring buffers.
This is the last step towards unifying how we process events in `TracerThread`
and towards producing `CaptureEvent`s in order of timestamps.
This also means that individual GPU tracepoint events are now also processed in
order, which solves the potential problem captured by the last test in
`GpuTracepointVisitorTest`: if the tracepoint events are not processed reasonably
in order, some wrong `GpuJob`s could result.
Performance implications of this change are not an issue as these events are
relatively low frequency: small test on this on Trata and Infiltrator with the
next commit.

Note that this currently doesn't work: `dma_fence_signaled` events can come out
of order even in a single ring buffer, which is not supported by
`PerfEventProcessor`/`PerfEventQueue`.
This is going to be solved in the next commit, by adding support for it to
`PerfEventQueue`.
#### PerfEventQueue now also supports events not in order in their fd
`PerfEventQueue` now has two top-level heaps/priority queues to sort events from
the oldest. The old priority queue sorts fifo queues of events ordered in their
ring buffer. The new priority queue sorts a single group of all the other events
that are known to possibly come out of order (these are GPU events).

Tested with both Trata and Infiltrator, I haven't noticed any difference in CPU
utilization.
#### Increase PerfEventProcessor::kProcessingDelayMs
Albeit not a real solution, this is supposed to mitigate http://b/177532389
("discarded as out of order" events) in average captures.

### DIFF
--- a/src/LinuxTracing/GpuTracepointVisitor.cpp
+++ b/src/LinuxTracing/GpuTracepointVisitor.cpp
@@ -2,7 +2,7 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
 
-#include "GpuTracepointEventProcessor.h"
+#include "GpuTracepointVisitor.h"
 
 #include <absl/container/flat_hash_map.h>
 #include <absl/meta/type_traits.h>
@@ -21,9 +21,8 @@ namespace orbit_linux_tracing {
 
 using orbit_grpc_protos::GpuJob;
 
-int GpuTracepointEventProcessor::ComputeDepthForEvent(const std::string& timeline,
-                                                      uint64_t start_timestamp,
-                                                      uint64_t end_timestamp) {
+int GpuTracepointVisitor::ComputeDepthForGpuJob(const std::string& timeline,
+                                                uint64_t start_timestamp, uint64_t end_timestamp) {
   if (!timeline_to_latest_timestamp_per_depth_.contains(timeline)) {
     timeline_to_latest_timestamp_per_depth_.emplace(timeline, std::vector<uint64_t>{});
   }
@@ -47,7 +46,7 @@ int GpuTracepointEventProcessor::ComputeDepthForEvent(const std::string& timelin
   return static_cast<int>(latest_timestamps_per_depth.size() - 1);
 }
 
-void GpuTracepointEventProcessor::CreateGpuExecutionEventIfComplete(const Key& key) {
+void GpuTracepointVisitor::CreateGpuJobAndSendToListenerIfComplete(const Key& key) {
   auto cs_it = amdgpu_cs_ioctl_events_.find(key);
   auto sched_it = amdgpu_sched_run_job_events_.find(key);
   auto dma_it = dma_fence_signaled_events_.find(key);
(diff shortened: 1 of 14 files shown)


src/LinuxTracing/GpuTracepointVisitor.cpp
```diff
@@ -128,33 +127,33 @@ void GpuTracepointEventProcessor::CreateGpuExecutionEventIfComplete(const Key& k
 // map and then try to create a complete GPU execution event. This event is only
 // created when all three types of GPU events have been received.
 
-void GpuTracepointEventProcessor::PushEvent(const AmdgpuCsIoctlPerfEvent& sample) {
```
florian-kuebler, line 131: Can the parameter be const*? Same below.

src/LinuxTracing/PerfEventQueue.cpp
```diff
@@ -25,41 +29,66 @@ void PerfEventQueue::PushEvent(std::unique_ptr<PerfEvent> event) {
...
 
-bool PerfEventQueue::HasEvent() { return !queues_heap_.empty(); }
+bool PerfEventQueue::HasEvent() const {
+  return !ordered_queues_heap_.empty() || !unordered_events_priority_queue_.empty();
+}
 
-PerfEvent* PerfEventQueue::TopEvent() { return queues_heap_.front()->front().get(); }
+PerfEvent* PerfEventQueue::TopEvent() {
+  PerfEvent* top_event = nullptr;
+  if (!unordered_events_priority_queue_.empty()) {
```
florian-kuebler, line 49: I don't quite get what is happening here.
If we have an unordered event at the top that has a timestamp greater then the top of the ordered events, it doesn't mean, that there isn't another unordered event with a smaller timestamp. Is that important?

florian-kuebler (approved):
LGTM, thanks!

dimitry- (approved):

src/LinuxTracing/GpuTracepointVisitor.cpp
dpallotti: This is the signature of the method as defined in `PerfEventVisitor`.

src/LinuxTracing/PerfEventQueue.cpp
dpallotti: Hmm, I don't really understand what you didn't understand. Potentially we have two "top events" (from the two queues) here and we want to get the older of the two. I added a comment with that by the way.

src/LinuxTracing/GpuTracepointVisitor.cpp
florian-kuebler: Right. Well should be fine.

src/LinuxTracing/PerfEventQueue.cpp
florian-kuebler: Thanks for the comment. My point is, that at a first glance, this code really looks like you want to get THE smallest event (by timestamp). But this just not the case, as there might be smaller elements in the unordered_events. However, as this seems to be expected, I am probably fine.

dpallotti: Yes, I want to get the smallest event by timestamp. `unordered_events_priority_queue_` is a priority queue and is sorted by timestamp. Or what am I missing?

dpallotti: Discussed offline, added a commit that makes field names clearer.

PR #1820, opened 2021-02-09
Extend JoinFutures to support arbitrary types

beckerhe:
`JoinFutures` used to be only implemented for `Future<void>`. This new
change implements `JoinFutures` for `Future<T>`. It takes an
`absl::Span<const Future<T>>` as its argument and returns a
`Future<std::vector<T>>`. So far it's not possible to change the
container, that is returned. It's also not allocator-aware.

### DIFF
--- a/src/OrbitBase/include/OrbitBase/JoinFutures.h
+++ b/src/OrbitBase/include/OrbitBase/JoinFutures.h
@@ -6,13 +6,90 @@
 #define ORBIT_BASE_JOIN_FUTURES_H_
 
 #include "OrbitBase/Future.h"
+#include "OrbitBase/FutureHelpers.h"
+#include "OrbitBase/Promise.h"
 #include "absl/types/span.h"
+
+namespace orbit_base_internal {
+
+template <typename T>
+class SharedStateJoin {
+  orbit_base::Promise<std::vector<T>> promise;
+  absl::Mutex mutex;
+  size_t incomplete_futures;
+  std::vector<std::optional<T>> results;
+
+  void SetResultsInPromise() {
+    std::vector<T> output;
+    output.reserve(results.size());
+    std::transform(results.begin(), results.end(), std::back_inserter(output),
+                   [](std::optional<T>& element) {
+                     CHECK(element.has_value());
+                     return *std::move(element);
+                   });
+    results.clear();
+    promise.SetResult(std::move(output));
+  }
+
+ public:
+  void SetNumberOfFutures(size_t size) {
+    absl::MutexLock lock{&mutex};
+    incomplete_futures = size;
+    results.resize(size);
+  }
+
+  void SetResult(size_t index, const T& argument) {
+    absl::MutexLock lock{&mutex};
+    if (!results[index].has_value()) {
+      results[index] = argument;
+      --incomplete_futures;
+    }
+
+    if (incomplete_futures == 0) {
+      SetResultsInPromise();
+    }
+  }
+
+  orbit_base::Future<std::vector<T>> GetFuture() const { return promise.GetFuture(); }
+};
+
+template <>
+struct SharedStateJoin<void> {
+  orbit_base::Promise<void> promise;
+  absl::Mutex mutex;
+  size_t incomplete_futures;
+};
+}  // namespace orbit_base_internal
 
 namespace orbit_base {
 
 // Returns a future which completes when all futures in the argument have completed.
 // Currently only available for Future<void>.
 [[nodiscard]] Future<void> JoinFutures(absl::Span<const Future<void>> futures);
+
+template <typename T>
+Future<std::vector<T>> JoinFutures(absl::Span<const Future<T>> futures) {
+  if (futures.empty()) {
+    Promise<std::vector<T>> promise;
+    promise.SetResult({});
+    return promise.GetFuture();
+  }
+
+  auto shared_state = std::make_shared<orbit_base_internal::SharedStateJoin<T>>();
+  shared_state->SetNumberOfFutures(futures.size());
+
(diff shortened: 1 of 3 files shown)


src/OrbitBase/include/OrbitBase/JoinFutures.h
```diff
@@ -6,13 +6,88 @@
...
+#include "OrbitBase/Promise.h"
 #include "absl/types/span.h"
 
+namespace orbit_base_internal {
+
+template <typename T>
+class SharedStateJoin {
+  orbit_base::Promise<std::vector<T>> promise;
+  absl::Mutex mutex;
+  int incompleted_futures;
```
dimitry-, line 19: incomplete_futures

src/OrbitBase/include/OrbitBase/JoinFutures.h
```diff
@@ -6,13 +6,88 @@
...
+  std::vector<std::optional<T>> results;
+
+  void SetResultsInPromise() {
+    std::vector<T> output;
+    output.reserve(results.size());
+    std::transform(results.begin(), results.end(), std::back_inserter(output),
+                   [](std::optional<T>& element) {
+                     CHECK(element.has_value());
+                     return *std::move(element);
```
dimitry-, line 28: this line made me read documentation on the difference of value() and operator*. I did not know there was a difference. Now I know.

src/OrbitBase/include/OrbitBase/JoinFutures.h
```diff
@@ -6,13 +6,88 @@
...
+  }
+
+  orbit_base::Future<std::vector<T>> GetFuture() const { return promise.GetFuture(); }
+};
+
+template <>
+struct SharedStateJoin<void> {
+  orbit_base::Promise<void> promise;
+  absl::Mutex mutex;
+  int incompleted_futures;
```
dimitry-, line 58: incomplete

src/OrbitBase/include/OrbitBase/JoinFutures.h
dimitry-: Also, it seems like this can be size_t, can it be negative?

antonrohr:
```suggestion
  size_t incompleted_futures;
```

dimitry- (approved):
otherwise lgtm

src/OrbitBase/include/OrbitBase/JoinFutures.h
```diff
@@ -6,13 +6,88 @@
...
+ public:
+  void SetNumberOfFutures(size_t size) {
+    absl::MutexLock lock{&mutex};
+    incompleted_futures = static_cast<int>(size);
+    results.resize(size);
+  }
+
+  void SetResult(int index, const T& argument) {
+    absl::MutexLock lock{&mutex};
+    results[index] = argument;
```
antonrohr, line 43: To make sure the same index cant be set multiple times
```suggestion
    absl::MutexLock lock{&mutex};
    CHECK(!results[index].has_value());
    results[index] = argument;
```

src/OrbitBase/include/OrbitBase/JoinFutures.h
```diff
@@ -6,13 +6,88 @@
...
+  }
+
+ public:
+  void SetNumberOfFutures(size_t size) {
+    absl::MutexLock lock{&mutex};
+    incompleted_futures = static_cast<int>(size);
+    results.resize(size);
+  }
+
+  void SetResult(int index, const T& argument) {
```
antonrohr, line 41:
```suggestion
  void SetResult(size_t index, const T& argument) {
```

src/OrbitBase/include/OrbitBase/JoinFutures.h
beckerhe: Yes and yes, it can be size_t. Fixed. Thanks!

src/OrbitBase/include/OrbitBase/JoinFutures.h
beckerhe: Glad to be at service. :slightly_smiling_face:

beckerhe: And yes. In this case it was intentional to call the "overload" that doesn't check for existence, since there is a CHECK above.

src/OrbitBase/include/OrbitBase/JoinFutures.h
beckerhe: Fixed.

antonrohr (approved):

src/OrbitBase/include/OrbitBase/JoinFutures.h
antonrohr: looks like dimitry was faster

beckerhe: Done!

src/OrbitBase/include/OrbitBase/JoinFutures.h
dimitry-: Also the fact that value() can throw an exception makes me think we should avoid it entirely in our code and always use operator*

src/OrbitBase/include/OrbitBase/JoinFutures.h
beckerhe: Done. But gracefully.

src/OrbitBase/include/OrbitBase/JoinFutures.h
beckerhe: Done.

PR #1822, opened 2021-02-09
Eventtrack refactor

reichlfl:
Refactoring of `EventTrack`, `TracepointTrack` and `ThreadStateTrack` to have them no longer inherit from `Track` and reduce code duplication.

This PR also introduces a new common base class for UI Elements underneath the `TimeGraph` which hopefully will make accessibility and removal of the TimeGraph dependency easier lateron.

### DIFF
--- a/src/OrbitGl/Track.cpp
+++ b/src/OrbitGl/Track.cpp
@@ -18,19 +18,12 @@
 #include "TimeGraphLayout.h"
 
 Track::Track(TimeGraph* time_graph, CaptureData* capture_data)
-    : time_graph_(time_graph),
+    : CaptureViewElement(time_graph),
       collapse_toggle_(std::make_shared<TriangleToggle>(
           TriangleToggle::State::kExpanded,
           [this](TriangleToggle::State state) { OnCollapseToggle(state); }, time_graph)),
       accessibility_(this, &time_graph->GetLayout()),
       capture_data_(capture_data) {
-  mouse_pos_[0] = mouse_pos_[1] = Vec2(0, 0);
-  pos_ = Vec2(0, 0);
-  size_ = Vec2(0, 0);
-  picking_offset_ = Vec2(0, 0);
-  picked_ = false;
-  moving_ = false;
-  canvas_ = nullptr;
   const Color kDarkGrey(50, 50, 50, 255);
   color_ = kDarkGrey;
   num_timers_ = 0;
@@ -90,6 +83,8 @@ void Track::DrawTriangleFan(Batcher* batcher, const std::vector<Vec2>& points, c
 }
 
 void Track::Draw(GlCanvas* canvas, PickingMode picking_mode, float z_offset) {
+  CaptureViewElement::Draw(canvas, picking_mode, z_offset);
+
   Batcher* batcher = canvas->GetBatcher();
 
   const TimeGraphLayout& layout = time_graph_->GetLayout();
@@ -160,7 +155,7 @@ void Track::Draw(GlCanvas* canvas, PickingMode picking_mode, float z_offset) {
   float button_offset = layout.GetCollapseButtonOffset();
   float toggle_y_pos = pos_[1] + half_label_height;
   Vec2 toggle_pos = Vec2(tab_x0 + button_offset, toggle_y_pos);
-  collapse_toggle_->SetPos(toggle_pos);
+  collapse_toggle_->SetPos(toggle_pos[0], toggle_pos[1]);
   collapse_toggle_->Draw(canvas, picking_mode, z_offset);
 
   // Draw label.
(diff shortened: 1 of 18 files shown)


src/OrbitGl/UIElement.h
```diff
@@ -0,0 +1,54 @@
...
+
+#include "PickingManager.h"
+
+class TimeGraph;
+class GlCanvas;
+
+namespace orbit_gl {
+
+/* Base class for UI elements drawn underneath the capture window. */
+class UIElement : public Pickable {
```
dpallotti, line 16: I would really prefer a more specific name that hints to the fact that this is an element of the capture window, as the comment says. For example, `CaptureWindowElement` or `CaptureViewElement` or `TimeGraphElement` something like that (not sure what our favorite way of calling the capture view is).

dpallotti, line 16: By the way, according to the style guide abbreviations should be capitalized as normal words, i.e., `UiElement` (but see the other comment on being more specific with this name).

dpallotti (commented):
Given the "[DRAFT]" prefix, is this ready for review before I go further?

reichlfl:
> Given the "[DRAFT]" prefix, is this ready for review before I go further?

It's ready to review, sorry, I thought I removed the prefix.

dpallotti (commented):
I'll leave it to Pablo and Pierric to verify the behavior, there are a few details I'm not familiar with here. In general the only big comment I have is to give `UiElement` a more specific name.

dpallotti:
And thanks for the change, I'm glad to see this hierarchy being fixed.

src/OrbitGl/Track.h
```diff
@@ -49,26 +46,23 @@ class Track : public Pickable, public std::enable_shared_from_this<Track> {
...
+  void UpdatePrimitives(uint64_t min_tick, uint64_t max_tick, PickingMode picking_mode,
+                        float z_offset = 0) override;
   void OnDrag(int x, int y) override;
-  [[nodiscard]] bool Draggable() override { return true; }
+
+  virtual void SetCaptureData(CaptureData* capture_data) { capture_data_ = capture_data; }
 
   [[nodiscard]] virtual Type GetType() const = 0;
-  [[nodiscard]] virtual bool Movable() { return true; }
+  [[nodiscard]] virtual bool Movable() { return !pinned_; }
```
karupayun, line 58: Only a comment: This might be reversed depending on how we would implement pinning. (for example, divided the TimeGraph in 2 horizontally)

src/OrbitGl/Track.cpp
```diff
@@ -159,7 +161,7 @@ void Track::Draw(GlCanvas* canvas, PickingMode picking_mode, float z_offset) {
   float button_offset = layout.GetCollapseButtonOffset();
   float toggle_y_pos = pos_[1] + half_label_height;
   Vec2 toggle_pos = Vec2(tab_x0 + button_offset, toggle_y_pos);
-  collapse_toggle_->SetPos(toggle_pos);
+  collapse_toggle_->SetPos(toggle_pos[0], toggle_pos[1]);
```
karupayun, line 164: Why this change?

karupayun (approved):
LGTM Florian! Nice change!

BTW, haven't tested anything about tracepoints, but happy that we came back to work there again.

src/OrbitGl/Track.cpp
reichlfl: Before this change, `TriangleToggle` had its own `SetPos` method which received a Vec2, but the new base class only receives x/y. Forgot this in the last commit, but this shouldn't be here, I'll move it to the previous commit. Thanks for finding this!

src/OrbitGl/UIElement.h
reichlfl: Agreed, I changed it to `CaptureViewElement`.

src/OrbitGl/UIElement.cpp
```diff
@@ -0,0 +1,41 @@
...
+namespace orbit_gl {
+
+UIElement::UIElement(TimeGraph* time_graph) : time_graph_(time_graph) {
+  mouse_pos_[0] = mouse_pos_[1] = Vec2(0, 0);
+  pos_ = Vec2(0, 0);
+  size_ = Vec2(0, 0);
+  picking_offset_ = Vec2(0, 0);
+  picked_ = false;
+  canvas_ = nullptr;
+  const Color kDarkGrey(50, 50, 50, 255);
```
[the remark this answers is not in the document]
reichlfl: It shows that I blatantly copied and pasted this initialization from Track.cpp and didn't pay enough attention :-) Thanks for finding this.

src/OrbitGl/UIElement.cpp
```diff
@@ -0,0 +1,41 @@
...
+
+namespace orbit_gl {
+
+UIElement::UIElement(TimeGraph* time_graph) : time_graph_(time_graph) {
+  mouse_pos_[0] = mouse_pos_[1] = Vec2(0, 0);
+  pos_ = Vec2(0, 0);
+  size_ = Vec2(0, 0);
+  picking_offset_ = Vec2(0, 0);
+  picked_ = false;
+  canvas_ = nullptr;
```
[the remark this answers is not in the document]
reichlfl: I completely agree. I moved this part from the original implementation from Track.cpp and didn't change it. Changed it now, thanks.

src/OrbitGl/UIElement.cpp
```diff
@@ -0,0 +1,41 @@
...
+  const Color kDarkGrey(50, 50, 50, 255);
+}
+
+void UIElement::OnPick(int x, int y) {
+  Vec2& mouse_pos = mouse_pos_[0];
+  canvas_->ScreenToWorld(x, y, mouse_pos[0], mouse_pos[1]);
```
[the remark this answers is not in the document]
reichlfl: I've changed this to be clearer (renamed `mouse_pos_[0]` to `mouse_pos_down_` and `mouse_pos_[1]` to `mouse_pos_cur_`), and actually found a bug during this - thanks for pointing this out!

PR #1823, opened 2021-02-09
Reduce flakiness of test GetThreadState.LinuxTracingTestsMainAndAnother

dpallotti:
The sleeps seem to help.

Bug: http://b/179098446

### DIFF
--- a/src/LinuxTracing/LinuxTracingUtilsTest.cpp
+++ b/src/LinuxTracing/LinuxTracingUtilsTest.cpp
@@ -52,6 +52,8 @@ TEST(GetThreadState, LinuxTracingTestsMainAndAnother) {
   pid_t thread_tid = -1;
   std::optional<char> thread_state_holding_mutex;
   std::optional<char> main_state_waiting_mutex;
+
+  mutex.Lock();
   std::thread thread{[&] {
     // Make sure /proc/<pid>/stat is parsed correctly
     // even when the thread name contains spaces and parentheses.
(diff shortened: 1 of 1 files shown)


src/LinuxTracing/LinuxTracingUtilsTest.cpp
```diff
@@ -56,14 +56,16 @@ TEST(GetThreadState, LinuxTracingTestsMainAndAnother) {
     // Make sure /proc/<pid>/stat is parsed correctly
     // even when the thread name contains spaces and parentheses.
     pthread_setname_np(pthread_self(), ") )  )()( )(  )");
+    // Give the main thread the time to get to the mutex.
+    std::this_thread::sleep_for(std::chrono::milliseconds(10));
```
beckerhe, line 60: Have you considered locking the mutex `mutex` before actually creating the thread?

src/LinuxTracing/LinuxTracingUtilsTest.cpp
```diff
@@ -56,14 +56,16 @@ TEST(GetThreadState, LinuxTracingTestsMainAndAnother) {
...
     {
       absl::MutexLock lock{&mutex};
       thread_tid = syscall(SYS_gettid);
       thread_state_holding_mutex = GetThreadState(thread_tid);
       main_state_waiting_mutex = GetThreadState(main_tid);
     }
     // Give the main thread the time to read this thread's state before exiting.
-    std::this_thread::sleep_for(std::chrono::milliseconds(50));
+    std::this_thread::sleep_for(std::chrono::milliseconds(100));
```
beckerhe, line 68: Could the same be achieved by locking a second mutex that was previosuly locked on the main thread? I guess that would also put the thread into a sleep state?

beckerhe (approved):
Hi Davide, thank you for the fix.

I added some comments on how to maybe avoid the sleeps (not sure though). But it's also fine for me to leave it as is.

src/LinuxTracing/LinuxTracingUtilsTest.cpp
dpallotti: Yes, but the idea here was really to get the state while sleeping, the state while holding a mutex is already tested with `main_state_waiting_mutex`.

src/LinuxTracing/LinuxTracingUtilsTest.cpp
dpallotti: Good point, I'll try and get back to you. One problem here is that the case that this one sleep should solve is by far the hardest to reproduce for me.

src/LinuxTracing/LinuxTracingUtilsTest.cpp
beckerhe: Ahh. Sorry. I didn't get that. Please ignore my comment. :see_no_evil:

dpallotti: Because it's not written anywhere. I'll add a comment.

src/LinuxTracing/LinuxTracingUtilsTest.cpp
dpallotti: Done.

src/LinuxTracing/LinuxTracingUtilsTest.cpp
dpallotti: Done.

PR #1824, opened 2021-02-09
Remove the clear capture button

vickyliu-go4it:
As there is no use case of the "Clear Capture" button in the new ui, we remove it completely.

bug: http://b/179470701


karupayun (approved):
LGTM Vicky!

BTW, could you explain why you update the paths to orbiticons.qrc and orbitimages.qrc? Some explanation in the body of the commit.

vickyliu-go4it:
> LGTM Vicky!
> 
> BTW, could you explain why you update the paths to orbiticons.qrc and orbitimages.qrc? Some explanation in the body of the commit.

Thanks for reviewing. I just updated the commit description to mention about this.

PR #1825, opened 2021-02-09
Refactor Preset Loading

beckerhe:
This PR refactors preset loading, especially the `OrbitApp::LoadPreset` function.

It is now split up into all the work that is done per module (`OrbitApp::LoadPresetModule`) and
all the work that is done after all preset modules have been loaded, like updating the UI, etc.

Also note, that `LoadPresetModule` (just like `LoadModule`) does not depend on UI-specific functionality.
That means it can be moved out of OrbitApp more easily in a subsequent PR.

This is necessary for usage in OrbitClientGgp, for example.

I tested the following manually:
- Loading a preset
- Loading a partial preset
- Loading a preset with frametracks
- Loading a partial preset with frametracks

### DIFF
--- a/src/OrbitGl/App.cpp
+++ b/src/OrbitGl/App.cpp
@@ -1300,7 +1300,7 @@ void OrbitApp::LoadSymbols(const std::filesystem::path& symbols_path, ModuleData
 }
 
 void OrbitApp::SelectFunctionsFromHashes(const ModuleData* module,
-                                         const std::vector<uint64_t>& function_hashes) {
+                                         absl::Span<const uint64_t> function_hashes) {
   for (const auto function_hash : function_hashes) {
     const orbit_client_protos::FunctionInfo* const function_info =
         module->FindFunctionFromHash(function_hash);
@@ -1314,7 +1314,7 @@ void OrbitApp::SelectFunctionsFromHashes(const ModuleData* module,
 }
 
 void OrbitApp::EnableFrameTracksFromHashes(const ModuleData* module,
-                                           const std::vector<uint64_t>& function_hashes) {
+                                           absl::Span<const uint64_t> function_hashes) {
   for (const auto function_hash : function_hashes) {
     const orbit_client_protos::FunctionInfo* const function_info =
         module->FindFunctionFromHash(function_hash);
--- a/src/OrbitGl/App.cpp
+++ b/src/OrbitGl/App.cpp
@@ -1114,6 +1114,14 @@ orbit_base::Future<ErrorMessageOr<void>> OrbitApp::LoadModule(const ModuleData* 
 
 orbit_base::Future<ErrorMessageOr<void>> OrbitApp::LoadModule(const std::string& module_path,
                                                               const std::string& build_id) {
+  const ModuleData* module_data = GetModuleByPath(module_path);
+
+  if (module_data == nullptr) {
+    return {ErrorMessage{absl::StrFormat("Module \"%s\" was not found", module_path)}};
+  }
+
+  if (module_data->is_loaded()) return {outcome::success()};
+
   const auto it = modules_currently_loading_.find(module_path);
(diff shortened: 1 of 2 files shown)


src/OrbitGl/App.cpp
```diff
@@ -1347,49 +1348,44 @@ void OrbitApp::EnableFrameTracksFromHashes(const ModuleData* module,
...
+  results.Then(main_thread_executor_, [this](std::vector<std::string> module_paths_not_found) {
+    module_paths_not_found.erase(
+        std::remove_if(module_paths_not_found.begin(), module_paths_not_found.end(),
+                       [](const std::string& path) { return path.empty(); }),
+        module_paths_not_found.end());
+
+    if (!module_paths_not_found.empty()) {
+      // Note that unloadable presets are disabled.
+      SendWarningToUi("Preset only partially loaded",
+                      absl::StrFormat("The following modules are not loaded:\n\"%s\"",
```
dpallotti, line 1383: Maybe "*were* not loaded"?

dpallotti (approved):
Very nice!

ronaldfw (approved):
LGTM!

src/OrbitGl/App.cpp
beckerhe: Done. Thanks for the review!

IrinaShkviro (approved):
Thank you, LGTM!

src/OrbitGl/App.cpp
```diff
@@ -1114,6 +1114,14 @@ orbit_base::Future<ErrorMessageOr<void>> OrbitApp::LoadModule(const ModuleData*
 
 orbit_base::Future<ErrorMessageOr<void>> OrbitApp::LoadModule(const std::string& module_path,
                                                               const std::string& build_id) {
+  const ModuleData* module_data = GetModuleByPath(module_path);
+
+  if (module_data == nullptr) {
+    return {ErrorMessage{absl::StrFormat("Module \"%s\" was not found", module_path)}};
```
antonrohr, line 1121: I wasn't aware that this simplification is possible
`Future<ErrorMessageOr<void>>{ErrorMessage{}}` to `{ErrorMessage{}}`
I like it. Thank you, I learned something!

src/OrbitGl/App.cpp
```diff
@@ -1347,49 +1348,44 @@ void OrbitApp::EnableFrameTracksFromHashes(const ModuleData* module,
...
-        frame_track_function_hashes_map.at(module_path).push_back(function_hash);
-      }
+  // First we try to load all preset modules in parallel
+  for (const auto& [module_path, preset_module] : preset_file->preset_info().path_to_module()) {
+    auto load_preset_result = LoadPresetModule(module_path, preset_module);
+
+    orbit_base::ImmediateExecutor immediate_executor{};
+    auto future = load_preset_result.Then(
+        &immediate_executor,
+        [module_path = module_path](const ErrorMessageOr<void>& result) -> std::string {
```
antonrohr, line 1361: What is this `[module_path = module_path]` and how does it differ to `[module_path]`?

antonrohr (approved):

src/OrbitGl/App.cpp
antonrohr: Is it because `const auto& [module_path, preset_module]` just creates an alias, or something in this direction?

src/OrbitGl/App.cpp
beckerhe: Well it's horrible in some way. The problem is that two implicit conversion need to happen here.

I'm going to introduce a specialization for Future<ErrorMessageOr<T>>. This will also get an additional constructor for ErrorMessage.

src/OrbitGl/App.cpp
beckerhe: Yes, you're on the right track. You can't capture a binding.

PR #1826, opened 2021-02-09
Introduce AnyMovable and use in MainThreadExecutor

beckerhe:
`AnyMovable` is typed-erase Value wrapper for types that fulfill the the `std::movable` concept.
It is similar to `std::any`, but types don't need to be copyable.

In a second commit,  `AnyMovable` is introduced to `MainThreadExecutor`. 
All tasks scheduled with `MainThreadExecutor::ScheduleAfter` are only executed as long as `MainThreadExecutor` exists.
But the function objects itself might live longer than `MainThreadExecutor` because they are owned by the `SharedState`-object
of the future which gets destroyed when the Future completes.

That's a problem when the function object has a non-trivial destructor that accesses resources. These resources might have
already been destroyed when the function object gets destroyed. One example is `ScopedStatus` which calls `StatusListener`
in its destructor. When such a `ScopedStatus` is captured in a lambda, the lambda should not live longer than `MainThreadExecutor` to ensure that the pointer to `StatusListener` is always valid.

The solution: Tasks scheduled by `MainThreadExecutor::ScheduleAfter` live in a `std::list` in `MainThreadExecutor` and the
`SharedState` hold by the `Future` only gets a stable reference to this `std::list`-entry. That ensure that the task is destructed when `MainThreadExecutor` is destructed.

### DIFF
--- a/src/OrbitBase/AnyMovableTest.cpp
+++ b/src/OrbitBase/AnyMovableTest.cpp
@@ -1,0 +1,98 @@
+// Copyright (c) 2021 The Orbit Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#include <gtest/gtest.h>
+
+#include <memory>
+
+#include "OrbitBase/AnyMovable.h"
+
+namespace orbit_base {
+
+TEST(AnyMovable, DefaultConstruction) {
+  AnyMovable any{};
+  EXPECT_FALSE(any.HasValue());
+}
+
+TEST(AnyMovable, CarryInt) {
+  AnyMovable any{42};
+
+  EXPECT_TRUE(any.HasValue());
+  EXPECT_EQ(any.type(), typeid(42));
+}
+
+TEST(AnyMovable, CarryUniquePtr) {
+  AnyMovable any{std::make_unique<int>(42)};
+
+  EXPECT_TRUE(any.HasValue());
+  EXPECT_EQ(any.type(), typeid(std::unique_ptr<int>));
+}
+
+TEST(AnyMovable, InPlaceConstructInt) {
+  AnyMovable any{std::in_place_type<int>, 42};
+
+  EXPECT_TRUE(any.HasValue());
+  EXPECT_EQ(any.type(), typeid(int));
+}
+
+TEST(AnyMovable, InPlaceConstructUniquePtr) {
+  AnyMovable any{std::in_place_type<std::unique_ptr<int>>, new int{42}};
+
+  EXPECT_TRUE(any.HasValue());
+  EXPECT_EQ(any.type(), typeid(std::unique_ptr<int>));
+}
+
+TEST(AnyMovable, EmplaceInt) {
+  AnyMovable any{};
+  any.Emplace<int>(42);
+
+  EXPECT_TRUE(any.HasValue());
+  EXPECT_EQ(any.type(), typeid(int));
+}
+
+TEST(AnyMovable, EmplaceUniquePtr) {
+  AnyMovable any{};
+  any.Emplace<std::unique_ptr<int>>(new int{42});
+
+  EXPECT_TRUE(any.HasValue());
+  EXPECT_EQ(any.type(), typeid(std::unique_ptr<int>));
+}
+
+TEST(any_movable_cast, ExtractInt) {
+  AnyMovable any{42};
+
+  auto* ptr = any_movable_cast<int>(&any);
+  EXPECT_NE(ptr, nullptr);
+  EXPECT_EQ(*ptr, 42);
+}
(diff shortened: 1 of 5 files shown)


src/OrbitBase/include/OrbitBase/AnyMovable.h
```diff
@@ -0,0 +1,121 @@
...
+    virtual void* AccessValue() = 0;
+    virtual ~Base() = default;
+  };
+
+  template <typename T>
+  class Storage final : public Base {
+    T value_;
+
+   public:
+    using Base::Base;
```
dimitry-, line 49: why is this needed?

src/OrbitGl/MainThreadExecutor.h
```diff
@@ -79,16 +81,22 @@ class MainThreadExecutor : public std::enable_shared_from_this<MainThreadExecuto
...
            argument = std::make_tuple(std::forward<decltype(argument)>(argument)...)]() mutable {
             orbit_base::CallTaskAndSetResultInPromise<ReturnType> helper{&promise};
-            std::apply([&](auto... args) { helper.Call(functor, args...); }, std::move(argument));
+            auto functor = orbit_base::any_movable_cast<std::decay_t<F>>(&*function_reference);
+            CHECK(functor != nullptr);
+            std::apply([&](auto... args) { helper.Call(*functor, args...); }, std::move(argument));
+            waiting_continuations_.erase(function_reference);
           };
       executor->Schedule(CreateAction(std::move(function_wrapper)));
     };
```
dimitry-, line 102: what happens if executor is destroyed after lock is released but before function_wrapper started execution?

src/OrbitBase/AnyMovableTest.cpp
```diff
@@ -0,0 +1,98 @@
...
+TEST(AnyMovable, DefaultConstruction) {
+  AnyMovable any{};
+  EXPECT_FALSE(any.HasValue());
+}
+
+TEST(AnyMovable, ShouldCarryInt) {
+  AnyMovable any{42};
+
+  EXPECT_TRUE(any.HasValue());
+  EXPECT_EQ(any.type().hash_code(), typeid(42).hash_code());
```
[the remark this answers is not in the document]
beckerhe: Yes. I was somehow under the impression that it's not a good idea to compare type_info. But yes, it's the right thing here.

src/OrbitBase/include/OrbitBase/AnyMovable.h
beckerhe: I think it isn't. I will remove it.

src/OrbitGl/MainThreadExecutor.h
beckerhe: In that case `function_wrapper` will never be executed. That's guaranteed by `MainThreadExecutor::Schedule` since the lifetime of all actions are tied to the executor.

In our particular implementation we rely on the behaviour of `QMetaObject::invokeMethod`. The line `QMetaObject::invokeMethod(main_thread_executor_, function_wrapper, Qt::QueuedConnection);` will only invoke `function_wrapper` as long as `main_thread_executor_` is still alive.

So we can assume that either function_wrapper is called while MainThreadExecutor is still alive or MainThreadExecutor is destroyed and `function_wrapper` won't be called.

dimitry-: ok, got it.

dimitry- (approved):

src/OrbitGl/MainThreadExecutor.h
beckerhe: I will add a comment explaining this.

src/OrbitBase/AnyMovableTest.cpp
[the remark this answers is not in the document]
beckerhe: Done. Thanks

src/OrbitBase/AnyMovableTest.cpp
```diff
@@ -0,0 +1,98 @@
...
+}
+
+TEST(AnyMovable, ShouldInPlaceConstructInt) {
+  AnyMovable any{std::in_place_type<int>, 42};
+
+  EXPECT_TRUE(any.HasValue());
+  EXPECT_EQ(any.type().hash_code(), typeid(int).hash_code());
+}
+
+TEST(AnyMovable, ShouldInPlaceConstructUniquePtr) {
```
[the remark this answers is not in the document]
beckerhe: Gladly done.

src/OrbitBase/include/OrbitBase/AnyMovable.h
beckerhe: Removed. Works. Thanks.

PR #1829, opened 2021-02-09
Fix crash on unhooking function when capture_data is not available

dimitry-:
Test: Start Orbit, load preset, unhook a function,
      make sure there is no crash.
Bug: http://b/179787649

### DIFF
--- a/src/OrbitGl/App.cpp
+++ b/src/OrbitGl/App.cpp
@@ -1743,7 +1743,11 @@ void OrbitApp::AddFrameTrack(uint64_t instrumented_function_id) {
 }
 
 void OrbitApp::RemoveFrameTrack(const FunctionInfo& function) {
-  CHECK(HasCaptureData());
+  // Ignore this call if there is no capture data
+  if (!HasCaptureData()) {
+    return;
+  }
+
   std::optional<uint64_t> instrumented_function_id =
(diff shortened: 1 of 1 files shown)


ronaldfw (approved):
LGTM, thanks!

karupayun (approved):
LGTM!

src/OrbitGl/App.cpp
```diff
@@ -1743,7 +1743,11 @@ void OrbitApp::AddFrameTrack(uint64_t instrumented_function_id) {
 }
 
 void OrbitApp::RemoveFrameTrack(const FunctionInfo& function) {
-  CHECK(HasCaptureData());
```
florian-kuebler, line 1746: Sorry for being later. I'd argue to keep the CHECK and move the if to the caller. RemoveFrameTrack should remove an existing frametrack from the ui. It does not make sense to call it if there is not CaptureData, thus no existing frametrack.

dimitry-: true - I will do it in main branch... I would keep it as is here

dimitry-: Actually it is used in 5 different places. I think it is better if we just ignore this here. Note that this call is also ignored if the function was not previously instrumented.

PR #1830, opened 2021-02-09
[Rebase to main] Fix crash on unhooking function when capture_data is not available

dimitry-:
Test: Start Orbit, load preset, unhook a function,
      make sure there is no crash.
Bug: http://b/179787649


karupayun (approved):
LGTM!

vickyliu-go4it (approved):
LGTM! Thanks for fixing it so efficiently!

PR #1832, opened 2021-02-09
Increase reliability of listing gamelets.

danielfenner:
I had e2e tests failing due to "List available instances failed." although
the instance looked healthy otherwise.
Let's try to be a bit more persistant (or stubborn).

Bug: http://b/179808977
Test: Compiled, ran, still works fine.


antonrohr (approved):
lgtm

PR #1834, opened 2021-02-10
Combine single selection and multi-selection methods

vickyliu-go4it:
As single selection can be considered as a special cases of the multi-selection, 
some methods and variables of single selection and multi-selection can be combined. 

More specifically, with this change, we 
* Replace `DataView::selected_index_` by using one index in `DataView::selected_indices_`;
* Replace `DataView::GetSelectedIndex` by returning an empty or single-valued vector with `DataView::GetVisibleSelectedIndices`;
* Update the `DataView::OnSelect` method to support both single selection and multi-selection and so remove `DataView::OnMultiSelect` method;
* Replace `OnRowSelected(std::optional<int> row)` and `OnMultiRowsSelected(const std::vector<int>& rows)` by
`OnRowsSelected(const std::vector<int>& rows)` in both `OrbitTreeView` and `OrbitTableModel`.

Bug: http://b/177814945.
Test: Check the selection, sorting and filtering behaviors in Live tab and Sampling tab, it should be same with the previous version.

### DIFF
--- a/src/OrbitGl/DataView.h
+++ b/src/OrbitGl/DataView.h
@@ -69,8 +69,7 @@ class DataView {
   void OnSort(int column, std::optional<SortingOrder> new_order);
   virtual void OnContextMenu(const std::string& action, int menu_index,
                              const std::vector<int>& item_indices);
-  virtual void OnSelect(std::optional<int> /*index*/) {}
-  virtual void OnMultiSelect(const std::vector<int>& /*indices*/) {}
+  virtual void OnSelect(const std::vector<int>& /*indices*/) {}
   // This method returns the intersection of selected indices and visible indices. The returned
   // value contains 0 or 1 index for a DataView with single selection, and contains 0 or
   // multiple indices for a DataView with multi-selection.
--- a/src/OrbitGl/DataView.h
+++ b/src/OrbitGl/DataView.h
@@ -70,8 +70,10 @@ class DataView {
   virtual void OnContextMenu(const std::string& action, int menu_index,
                              const std::vector<int>& item_indices);
   virtual void OnSelect(std::optional<int> /*index*/) {}
-  [[nodiscard]] virtual std::optional<int> GetSelectedIndex();
   virtual void OnMultiSelect(const std::vector<int>& /*indices*/) {}
+  // This method returns the intersection of selected indices and visible indices. The returned
+  // value contains 0 or 1 index for a DataView with single selection, and contains 0 or
+  // multiple indices for a DataView with multi-selection.
   [[nodiscard]] virtual std::vector<int> GetVisibleSelectedIndices();
   virtual void OnDoubleClicked(int /*index*/) {}
   virtual void OnDataChanged();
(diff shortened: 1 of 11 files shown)


vickyliu-go4it:
Hi Florian and Pierric, thanks for reviewing this PR :-) 

As this change is closely related to my previous PR on multi-selection, I think you would be more familiar with this part. I suggest to review it commit by commit. Thanks again!

reichlfl (approved):
LGTM, thanks! Best to wait for Pierric to also take a look to be sure I didn't miss any changes in the logic.

src/OrbitGl/DataView.h
```diff
@@ -70,9 +69,10 @@ class DataView {
   void OnSort(int column, std::optional<SortingOrder> new_order);
   virtual void OnContextMenu(const std::string& action, int menu_index,
                              const std::vector<int>& item_indices);
-  virtual void OnSelect(std::optional<int> /*index*/) {}
-  [[nodiscard]] virtual std::optional<int> GetSelectedIndex() { return selected_index_; }
-  virtual void OnMultiSelect(const std::vector<int>& /*indices*/) {}
+  virtual void OnSelect(const std::vector<int>& /*indices*/) {}
+  // This method returns the intersection of selected indices and visible indices. The returned
+  // value contains 0 or 1 indices for a DataView with single selection, and contains 0 or
```
pierricgimmig, line 74: nit: "0 or 1 index"

pierricgimmig: and later "0 or multiple indices"

src/OrbitGl/LiveFunctionsDataView.cpp
```diff
@@ -98,21 +98,26 @@ std::string LiveFunctionsDataView::GetValue(int row, int column) {
   }
 }
 
-std::optional<int> LiveFunctionsDataView::GetSelectedIndex() {
-  return GetRowFromFunctionId(selected_function_id_);
+std::vector<int> LiveFunctionsDataView::GetVisibleSelectedIndices() {
+  std::optional<int> visible_selected_index = GetRowFromFunctionId(selected_function_id_);
+  std::vector<int> visible_selected_indices;
+  if (visible_selected_index.has_value()) {
+    visible_selected_indices.push_back(visible_selected_index.value());
```
pierricgimmig, line 105: nit. we could use initializer lists here : "return {visible_selected_index.value()};" and "return {}", I find it clearer and takes up fewer lines.

pierricgimmig (approved):
Hi Vicky, thanks for this, it makes the code simpler. LGTM

src/OrbitGl/DataView.h
vickyliu-go4it: Done. Thanks for pointing out this!

src/OrbitGl/LiveFunctionsDataView.cpp
vickyliu-go4it: Make sense! Changed accordingly. Thanks!

PR #1835, opened 2021-02-10
Fix inconsistent min/max capture timestamp.

florian-kuebler:
Whenever we update the capture timestamps in TimeGraph, we should
make sure to update both, such that they are always consistent.

Bug: http://b/179884465
Test: Load the capture linked in the bug.
Note: Loading the capture still fails sometimes, but this is a different bug.


karupayun (approved):
LGTM! :)

PR #1837, opened 2021-02-10
Dump on symbol loading

IrinaShkviro:
Add DumpWithoutCrash for failed symbol loading.


conanfile.py
```diff
@@ -94,7 +94,7 @@ def requirements(self):
                 "libunwindstack/80a734f14@{}#0".format(self._orbit_channel))
         self.requires("zlib/1.2.11#9e0c292b60ce77402bd9be60dd68266f")
 
-        if self.options.with_gui and self.options.with_crash_handling:
+        if self.options.with_orbitgl:
```
beckerhe, line 97: Since crashpad is not optional anymore, can you remove the `with_crash_handling` option?

src/CrashHandler/CMakeLists.txt
```diff
@@ -0,0 +1,32 @@
+# Copyright (c) 2020 The Orbit Authors. All rights reserved.
```
antonrohr, line 1: nit: 2021

src/CrashHandler/include/CrashHandler/CrashHandler.h
```diff
@@ -1,30 +1,33 @@
...
-  std::unique_ptr<crashpad::CrashReportDatabase> crash_report_db_;
-};
-
-#endif  // ORBIT_QT_CRASH_HANDLER_H_
\ No newline at end of file
+// Copyright (c) 2020 The Orbit Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#ifndef ORBIT_CRASH_HANDLER_CRASH_HANDLER_H_
```
antonrohr, line 5: I guess
```suggestion
#ifndef CRASH_HANDLER_CRASH_HANDLER_H_
```

src/OrbitGl/App.cpp
```diff
@@ -1279,6 +1283,9 @@ void OrbitApp::LoadSymbols(const std::filesystem::path& symbols_path, ModuleData
        frame_track_function_hashes = std::move(frame_track_function_hashes), module_data,
        symbols_path, this](const ErrorMessageOr<void>& result) {
         if (result.has_error()) {
+          if (crash_handler_ != nullptr) {
+            crash_handler_->DumpWithoutCrash();
+          }
```
antonrohr, line 1288: personal preference: I would put this between `ERROR` and `SendErrorToUi`.

antonrohr (approved):
LGTM, maybe @beckerhe can take a closer look at the cmake files and the conanfile, I am no expert there

src/CrashHandler/CMakeLists.txt
IrinaShkviro: Updated the copyright

src/CrashHandler/include/CrashHandler/CrashHandler.h
IrinaShkviro: Right, thanks!

src/OrbitGl/App.cpp
IrinaShkviro: Moved DumpWithoutCrash() call there

IrinaShkviro:
I've updated the code to include crashpad package only when it's actually needed. Please take a look again.

src/CrashHandler/CMakeLists.txt
```diff
@@ -0,0 +1,33 @@
...
+target_sources(CrashHandler PUBLIC include/CrashHandler/CrashHandlerBase.h)
+target_link_libraries(CrashHandler PRIVATE OrbitBase)
+
+if(WITH_CRASH_HANDLING)
+target_sources(CrashHandler PUBLIC include/CrashHandler/CrashHandler.h
+                                   include/CrashHandler/CrashOptions.h
+                            PRIVATE CrashHandler.cpp)
+target_link_libraries(CrashHandler PUBLIC CONAN_PKG::crashpad
+                                   PRIVATE OrbitCore
+                                           OrbitVersion)
```
beckerhe, line 27: That means that `OrbitGlTests` and `OrbitQtTests` will link against crashpad (and all other targets that require OrbitQt or OrbitGl). I think that's not a good idea. That's why I suggested to move `CrashpadHandlerBase.h`  into `OrbitBase`. WDYT?

IrinaShkviro: Moved CrashHandlerBase to OrbitBase.

src/OrbitBase/include/OrbitBase/CrashHandlerBase.h
```diff
@@ -0,0 +1,19 @@
+// Copyright (c) 2021 The Orbit Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#ifndef ORBIT_BASE_CRASH_HANDLER_BASE_H_
+#define ORBIT_BASE_CRASH_HANDLER_BASE_H_
+
+namespace orbit_base {
+
+class CrashHandlerBase {
```
beckerhe, line 10: Could you add a small comment explaining that this is just a stub implementation and that the real implementation lives in CrashHandler::?

beckerhe: Personal preference: I would call the class `CrashHandler`. The suffix `Base` sounds a bit like an implementation detail to me.

beckerhe (approved):
Hi Irina, LGTM. Thank you for all your changes. I'm sorry for being so persistent with some of my demands. :see_no_evil: Thank you for that change!

src/OrbitBase/include/OrbitBase/CrashHandlerBase.h
IrinaShkviro: Added a comment and updated the class name

IrinaShkviro:
Thank you so much for the review!

PR #1838, opened 2021-02-10
Fix bug with optimized text rendering

ronaldfw:
There was a small and somewhat subtle issue with the new optimization in
our text renderer: One code path used texture_font_get_glyph without
checking if the glyph was already loaded. This is normally fine as
texture_font_get_glyph internally loads the glyph if it cannot be found.
However, this updates the texture atlas, which the optimization needs to
know about. We fix this by introducing a new helper method that always
checks if a glyph has been loaded already, maybe loads it, and then
gets the glyph and returns it. If the glyph is loaded, the method
indicates that the texture atlas was updated.

Tested: Manually built Orbit and checked if text is rendered correctly.
Bug: http://b/179810089

### DIFF
--- a/src/OrbitGl/TextRenderer.cpp
+++ b/src/OrbitGl/TextRenderer.cpp
@@ -111,6 +111,20 @@ texture_font_t* TextRenderer::GetFont(uint32_t size) {
   return fonts_by_size_[size];
 }
 
+// Always use this method when you need to get a glyph, we need to make sure we know when the
+// texture atlas has been updated. One way to ensure that is to check if we can find the glyph
+// already and, if not, load it explicitly (in which case the texture atlas is updated). Note that
+// texture_font_get_glyph internally may load the glyph if it does not find it. We do not want
+// that as in that case, we do not know that the atlas has actually changed.
+texture_glyph_t* TextRenderer::MaybeLoadAndGetGlyph(texture_font_t* font, const char* character) {
+  if (!texture_font_find_glyph(font, character)) {
+    texture_font_load_glyph(font, character);
+    texture_atlas_changed_ = true;
+  }
+
+  return texture_font_get_glyph(font, character);
+}
+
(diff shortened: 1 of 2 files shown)


src/OrbitGl/TextRenderer.cpp
```diff
@@ -386,7 +390,8 @@ int TextRenderer::GetStringWidthScreenSpace(const char* text, uint32_t font_size
 
   std::size_t len = strlen(text);
   for (std::size_t i = 0; i < len; ++i) {
-    texture_glyph_t* glyph = texture_font_get_glyph(GetFont(font_size), text + i);
+    texture_font_t* font = GetFont(font_size);
+    texture_glyph_t* glyph = MaybeLoadAndGetGlyph(font, text + i);
```
beckerhe, line 394: Am I getting this right? The only behavioural change happens here? We set `texture_atlas_changed_` to true if needed. And that wasn't the case before, right?

ronaldfw: Yes, exactly. If a glyph was loaded here, we would not update the texture atlas, and we would not notice later as we would just find it via texture_font_find_glyph (which doesn't trigger an update). So, unless there would be a different texture atlas update triggered (by lucky coincidence), these glyphs would just be missing from our texture.

beckerhe: Ok, great. Thanks for the explanation!

beckerhe (approved):
LGTM!

pierricgimmig (commented):
LGTM, thanks @ronaldfw! I'm guessing texture_font_find_glyph is fast? This shouldn't have an effect on the big performance gain of the previous change right?

ronaldfw:
> LGTM, thanks @ronaldfw! I'm guessing texture_font_find_glyph is fast? This shouldn't have an effect on the big performance gain of the previous change right?

Yes, this doesn't have a noticeable effect. The gain was from avoiding updating the texture every frame with glTexImage2D.

PR #1840, opened 2021-02-10
Fix load_bias calculation

dimitry-:
Calculate load_bias based on the executable PT_LOAD segment.

Newer linkers create first segment as read only and
put .text section into a separate segment. Which means
load_bias calculation is off if second segment vaddr-offset
difference is different from the first PT_LOAD segment

Test: run unit-tests
Test: run capture on executable with executable PT_LOAD
      segment different from the first PT_LOAD
Bug: http://b/179894572

### DIFF
--- a/src/ElfUtils/ElfFile.cpp
+++ b/src/ElfUtils/ElfFile.cpp
@@ -175,40 +175,34 @@ ErrorMessageOr<ModuleSymbols> ElfFileImpl<ElfT>::LoadSymbols() {
 template <typename ElfT>
 ErrorMessageOr<uint64_t> ElfFileImpl<ElfT>::GetLoadBias() const {
   const llvm::object::ELFFile<ElfT>* elf_file = object_file_->getELFFile();
-
-  uint64_t min_vaddr = UINT64_MAX;
-  bool pt_load_found = false;
   llvm::Expected<typename ElfT::PhdrRange> range = elf_file->program_headers();
 
   if (!range) {
-    std::string error = absl::StrFormat(
-        "Unable to get load bias of ELF file: \"%s\". No program headers "
-        "found.",
-        file_path_.string());
-    ERROR("%s", error.c_str());
+    std::string error =
+        absl::StrFormat("Unable to get load bias of ELF file: \"%s\". No program headers found.",
+                        file_path_.string());
+    ERROR("%s", error);
     return ErrorMessage(std::move(error));
   }
 
+  // Find the executable segment and calculate the load bias based on that segment.
   for (const typename ElfT::Phdr& phdr : range.get()) {
     if (phdr.p_type != llvm::ELF::PT_LOAD) {
       continue;
     }
-    pt_load_found = true;
 
-    if (min_vaddr > phdr.p_vaddr) {
-      min_vaddr = phdr.p_vaddr;
+    if ((phdr.p_flags & llvm::ELF::PF_X) == 0) {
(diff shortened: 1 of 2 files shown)


dimitry-:
need to address the "file without .text segment" use-case

src/ElfUtils/ElfFile.cpp
```diff
@@ -175,40 +180,43 @@ ErrorMessageOr<ModuleSymbols> ElfFileImpl<ElfT>::LoadSymbols() {
...
+
+  if (!text_section_header_) {
+    std::string error =
+        absl::StrFormat("Unable to get load bias of ELF file: \"%s\". .text section was not found.",
+                        file_path_.string());
     ERROR("%s", error.c_str());
     return ErrorMessage(std::move(error));
   }
 
+  // Find the segment that contains .text section and calculate the load bias
```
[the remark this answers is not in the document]
dimitry-: at most one... if you strip debug symbols and store them to a separate file the objcopy generates an empty 0-sized executable pt_load segment. but it does not contain .text section

src/ElfUtils/ElfFile.cpp
```diff
@@ -175,40 +175,34 @@ ErrorMessageOr<ModuleSymbols> ElfFileImpl<ElfT>::LoadSymbols() {
...
-        "found.",
-        file_path_.string());
+    std::string error =
+        absl::StrFormat("Unable to get load bias of ELF file: \"%s\". No program headers found.",
+                        file_path_.string());
     ERROR("%s", error.c_str());
     return ErrorMessage(std::move(error));
   }
 
+  // Find the executable segment and calculate the load bias based on that segment.
```
dpallotti, line 188: I think we briefly mentioned it yesterday: We assume a single executable segment? Could you clarify?

src/ElfUtils/ElfFile.cpp
```diff
@@ -175,40 +175,34 @@ ErrorMessageOr<ModuleSymbols> ElfFileImpl<ElfT>::LoadSymbols() {
...
-    ERROR("%s", error.c_str());
-    return ErrorMessage(std::move(error));
+    return phdr.p_vaddr - phdr.p_offset;
   }
-  return min_vaddr;
+
+  std::string error = absl::StrFormat(
+      "Unable to get load bias of ELF file: \"%s\". No executable PT_LOAD segment found.",
+      file_path_.string());
+  ERROR("%s", error.c_str());
```
dpallotti, line 204: (You don't need `.c_str()` as internally `absl::StrFormat` is used.)

dpallotti (approved):
Thanks a lot!

antonrohr (approved):
I did not test it, but the code lgtm

danielfenner (approved):
lgtm

src/ElfUtils/ElfFile.cpp
dimitry-: In practice we have one, but there is nothing wrong from the elf-file format point of view to have more then one, ideally we want the one with .text segment, but there is a use case where this is not possible, debug symbols file does not have any executable sections. It is however keeps the 0 sized executable segment, I am assuming specifically so that debug tools could calculate load_bias for symbol adresses.

src/ElfUtils/ElfFile.cpp
dimitry-: done.

PR #1841, opened 2021-02-11
Move DataViewType specific refresh operation to DataView::OnRefresh

vickyliu-go4it:
In the current `OrbitTreeView::Refresh` method, we change the behavior of `OrbitTreeView` based on the specific type of `DataView`.
In order to remove the internal information of `DataView` from `OrbitTreeView`, we move the `DataViewType` specific refresh operations to the `DataView::OnRefresh` method. 

bug: http://b/177812629

### DIFF
--- a/src/OrbitGl/DataView.h
+++ b/src/OrbitGl/DataView.h
@@ -40,10 +40,7 @@ class DataView {
   };
 
   explicit DataView(DataViewType type, OrbitApp* app)
-      : update_period_ms_(-1),
-        refresh_mode_(RefreshMode::kOther),
-        type_(type),
-        app_{app} {}
+      : update_period_ms_(-1), type_(type), app_{app} {}
 
   virtual ~DataView() = default;
 
@@ -64,7 +61,8 @@ class DataView {
   // Filter callback set from UI layer.
   using FilterCallback = std::function<void(const std::string&)>;
   void SetUiFilterCallback(FilterCallback callback) { filter_callback_ = std::move(callback); }
-  void SetRefreshMode(RefreshMode mode) { refresh_mode_ = mode; }
+  virtual void OnRefresh(const std::vector<int>& /*visible_selected_indices*/,
+                         const RefreshMode& /*mode*/) {}
 
   void OnSort(int column, std::optional<SortingOrder> new_order);
   virtual void OnContextMenu(const std::string& action, int menu_index,
@@ -95,6 +93,7 @@ class DataView {
 
   int GetUpdatePeriodMs() const { return update_period_ms_; }
   DataViewType GetType() const { return type_; }
+  [[nodiscard]] virtual bool ResetOnRefresh() const { return true; }
 
  protected:
   void InitSortingOrders();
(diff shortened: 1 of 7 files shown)


src/OrbitGl/DataView.h
```diff
@@ -109,6 +108,7 @@ class DataView {
   int update_period_ms_;
   absl::flat_hash_set<int> selected_indices_;
   RefreshMode refresh_mode_;
```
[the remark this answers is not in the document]
reichlfl: As discussed offline: I've realized that it is actually still used, but I think it is worth the effort to remove it by splitting the code in OnSelect() into two methods and calling the part that is required during a refresh explicitly.

src/OrbitGl/DataView.h
```diff
@@ -95,6 +93,7 @@ class DataView {
 
   int GetUpdatePeriodMs() const { return update_period_ms_; }
   DataViewType GetType() const { return type_; }
+  bool GetResetOnRefresh() const { return reset_on_refresh_; }
```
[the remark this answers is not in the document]
vickyliu-go4it: That's really a nice idea! Changed accordingly. Thanks!

src/OrbitGl/DataView.h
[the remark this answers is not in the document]
vickyliu-go4it: Thanks for the suggestion! I split `OnSelect` into two methods for both `LiveFunctionsDataView` and `SamplingReportDataView`. Now the behavior of `OnSelect` won't be affected by the `RefreshMode`.

src/OrbitGl/DataView.h
```diff
@@ -93,7 +93,7 @@ class DataView {
 
   int GetUpdatePeriodMs() const { return update_period_ms_; }
   DataViewType GetType() const { return type_; }
-  bool GetResetOnRefresh() const { return reset_on_refresh_; }
+  virtual bool ResetOnRefresh() const { return true; }
```
reichlfl, line 96: [[nodiscard]]

src/OrbitGl/LiveFunctionsDataView.h
```diff
@@ -42,6 +43,7 @@ class LiveFunctionsDataView : public DataView {
   void OnTimer() override;
   void OnRefresh(const std::vector<int>& visible_selected_indices,
                  const RefreshMode& mode) override;
+  bool ResetOnRefresh() const override { return false; }
```
reichlfl, line 46: [[nodiscard]]

reichlfl (approved):
Thanks for cleaning this up, IMHO that makes it easier to see what's going on. Small nitpicks, otherwise LGTM!

pierricgimmig (approved):
Thanks for this Vicky, the code is cleaner now!

src/OrbitGl/DataView.h
vickyliu-go4it: Done. Thanks!

src/OrbitGl/LiveFunctionsDataView.h
vickyliu-go4it: Done. Thanks for pointing out this.

PR #1842, opened 2021-02-11
Move fuzzers out of OrbitGl

beckerhe:
This is moving fuzzers out of OrbitGl. One fuzzer can be deleted. It just makes no sense. The other one went into OrbitClientModel.

Since these were the last fuzzers in OrbitGl, we can also remove the conan flag `with_orbitgl` which only existed for fuzzing. It allowed to build Orbit without the UI (OrbitQt), but with OrbitGl enabled.

In the last commit I'm adjusting the `libfuzzer_relwithdebinfo` profile which can be used to build and run fuzzers on developer machines. The profile is adjusted to the profile we use in oss-fuzz. The change is not ideal because with that change dependencies won't be built with sanitizers enabled which makes it harder for fuzzers and users to find problems occuring third party dependencies. But the sanitizer build is currently broken anyway. This will come back in a later PR.


IrinaShkviro (approved):
LGTM!

src/ClientProtos/capture_data.proto
```diff
@@ -175,4 +175,5 @@ message Color {
 message CaptureDeserializerFuzzerInfo {
   CaptureInfo capture_info = 1;
   repeated TimerInfo timers = 2;
+  string filename = 3;
```
antonrohr, line 178: This is only used for the fuzz test? There should be a comment about that.

antonrohr (approved):
lgtm. one comment about a comment. comment inception

src/ClientProtos/capture_data.proto
beckerhe: Well the whole proto ist just for the fuzz test. And there is a comment explaining that

beckerhe: I will go ahead and merge that. I hope that's okay. :innocent:

PR #1843, opened 2021-02-11
LinuxTracingIntegrationTest: handle missing second frame with FPs

dpallotti:
This happens when the sample is on `push rbp` or `mov rbp,rsp`.
The goal here is only to handle this gracefully in the tests for now.

Bug: http://b/179376436
Bug: http://b/179976268

### DIFF
--- a/src/LinuxTracing/LinuxTracingIntegrationTest.cpp
+++ b/src/LinuxTracing/LinuxTracingIntegrationTest.cpp
@@ -647,7 +647,7 @@ void VerifyCallstackSamplesWithOuterAndInnerFunction(
     const std::vector<orbit_grpc_protos::CaptureEvent>& events, pid_t pid,
     std::pair<uint64_t, uint64_t> outer_function_virtual_address_range,
     std::pair<uint64_t, uint64_t> inner_function_virtual_address_range, double sampling_rate,
-    const absl::flat_hash_set<uint64_t>* address_infos_received) {
+    const absl::flat_hash_set<uint64_t>* address_infos_received, bool unwound_with_frame_pointers) {
   uint64_t previous_callstack_timestamp_ns = 0;
   size_t matching_callstack_count = 0;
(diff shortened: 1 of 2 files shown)


src/LinuxTracing/LinuxTracingIntegrationTest.cpp
```diff
@@ -676,6 +676,14 @@ void VerifyCallstackSamplesWithOuterAndInnerFunction(
       // address and the caller address should match the "outer" function's address.
       if (callstack.pcs(pc_index) >= inner_function_virtual_address_range.first &&
           callstack.pcs(pc_index) <= inner_function_virtual_address_range.second) {
+        // Frame-pointer unwinding skips the second innermost frame if the sample fell on `push rbp`
+        // (which is one byte) or `mov rbp,rsp`. Disregard such samples. See b/179376436#comment4.
+        if (unwound_with_frame_pointers &&
+            (callstack.pcs(pc_index) == inner_function_virtual_address_range.first ||
+             callstack.pcs(pc_index) == inner_function_virtual_address_range.first + 1)) {
+          continue;
+        }
```
dimitry-, line 685: should we fix unwinding instead?

dimitry-: also it only makes sense for the top pc right?

dpallotti: > should we fix unwinding instead?

The unwinding is done by perf_event_open.

> also it only makes sense for the top pc right?

You mean top of the callstack?

dpallotti: > > also it only makes sense for the top pc right?

> You mean top of the callstack?

Added `pc_index == 0` to the `if`.

dimitry- (approved):
ok, this is not the place to fix-it-up, we can probably do it later on the client-side.

PR #1844, opened 2021-02-11
Fix segfault in TimerTrack::UpdatePrimitives.

florian-kuebler:
PR #1760 introduced handling for overlapping timers, which required
to look forward to the next timer. This code was broken.

This change fixes this and simplifies the logic. The iteration
is now a "sliding window" of previous, current and next pointers
to the TextBoxes (with underlying timers).

Test: Take a capture.

### DIFF
--- a/src/OrbitGl/TimerTrack.cpp
+++ b/src/OrbitGl/TimerTrack.cpp
@@ -251,6 +251,13 @@ void TimerTrack::UpdatePrimitives(uint64_t min_tick, uint64_t max_tick,
 
   for (auto& chain : chains_by_depth) {
     if (!chain) continue;
+    // In order to draw overlaps correctly, we need for every text box to be drawn (current),
+    // its previous and next text box. In order to avoid looking ahead for the next text (which is
+    // error-prone), we are doing just one traversal of the text boxes, while keeping track of the
+    // previous two timers and will draw the previous one, thus the currents iteration value being
+    // the "next" textbox.
+    // Note: This will require us to draw the last timer after the traversal of the text boxes.
+    // Also note: The draw method will take care of nullptr's being passed into.
     TextBox* prev_text_box = nullptr;
     TextBox* current_text_box = nullptr;
     TextBox* next_text_box = nullptr;
@@ -263,6 +270,8 @@ void TimerTrack::UpdatePrimitives(uint64_t min_tick, uint64_t max_tick,
       if (!block.Intersects(min_tick, max_tick)) continue;
 
       for (size_t k = 0; k < block.size(); ++k) {
+        // Remember that the current index (k) points to the "next" text box and we want to draw
+        // the text box from the previous iteration ("current").
         next_text_box = &block[k];
 
         DrawTimer(prev_text_box, current_text_box, next_text_box, min_tick, max_tick, z_offset,
@@ -274,6 +283,7 @@ void TimerTrack::UpdatePrimitives(uint64_t min_tick, uint64_t max_tick,
         current_text_box = next_text_box;
       }
     }
+    // We still need to draw the last timer.
     next_text_box = nullptr;
     DrawTimer(prev_text_box, current_text_box, next_text_box, min_tick, max_tick, z_offset, batcher,
               canvas, world_start_x, world_width, inv_time_window, is_collapsed, z,
--- a/src/OrbitGl/TimerTrack.cpp
+++ b/src/OrbitGl/TimerTrack.cpp
@@ -82,6 +82,147 @@ WorldXInfo ToWorldX(double start_us, double end_us, double inv_time_window, floa
 
 }  // namespace
 
+void TimerTrack::DrawTimer(TextBox* prev_text_box, TextBox* current_text_box,
+                           TextBox* next_text_box, uint64_t min_tick, uint64_t max_tick,
+                           float z_offset, Batcher* batcher, GlCanvas* canvas, float world_start_x,
+                           float world_width, double inv_time_window, bool is_collapsed, float z,
+                           const TextBox* selected_textbox, uint64_t highlighted_function_id,
+                           uint64_t pixel_delta_in_ticks, uint64_t min_timegraph_tick,
+                           uint64_t* min_ignore, uint64_t* max_ignore) {
+  CHECK(min_ignore != nullptr);
+  CHECK(max_ignore != nullptr);
+  if (current_text_box == nullptr) return;
+  const TimerInfo& current_timer_info = current_text_box->GetTimerInfo();
+  if (min_tick > current_timer_info.end() || max_tick < current_timer_info.start()) return;
+  if (current_timer_info.start() >= *min_ignore && current_timer_info.end() <= *max_ignore) return;
+  if (!TimerFilter(current_timer_info)) return;
+
+  UpdateDepth(current_timer_info.depth() + 1);
+  double start_us = time_graph_->GetUsFromTick(current_timer_info.start());
+  double start_or_prev_end_us = start_us;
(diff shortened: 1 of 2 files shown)


src/OrbitGl/TimerTrack.h
```diff
@@ -84,6 +84,13 @@ class TimerTrack : public Track {
     return true;
   }
 
+  void DrawTimer(TextBox* prev_text_box, TextBox* current_text_box, TextBox* next_text_box,
```
karupayun, line 87: I'm not finding any recommendation on limits in the arguments of a function, but seems to be very big. If you want to make a function here (which makes sense), don't you think that we would want to have 2 or 3 different functions here?

src/OrbitGl/TimerTrack.cpp
```diff
@@ -110,151 +251,34 @@ void TimerTrack::UpdatePrimitives(uint64_t min_tick, uint64_t max_tick,
...
-          if (pixel_delta_in_ticks != 0) {
-            min_ignore = min_timegraph_tick +
-                         ((timer_info.start() - min_timegraph_tick) / pixel_delta_in_ticks) *
-                             pixel_delta_in_ticks;
-            max_ignore = min_ignore + pixel_delta_in_ticks;
-          }
-        }
-
-        prev_timer_info = &timer_info;
+        next_text_box = &block[k];
```
karupayun, line 266: Could you explain why in the first iteration we are having:
prev = null
current = null
next = block[0]
?
So, why we need to draw then current is the NULLptr?

More precisely, why not work at first with this border case and making the loop starting where current is the first and next is the second (possibly NULL) timer?

src/OrbitGl/TimerTrack.h
florian-kuebler: Well in the end, there would be one function surrounding which needs all the arguments, so I guess I'd keep it that way.

src/OrbitGl/TimerTrack.cpp
florian-kuebler: So this is intentional:
The issue before was, that we tried when looking at the "current" timer, to look forward to the "next" one and messed up the iterator.

The idea is now, to have the current block index pointing to the "next timer" and remember the previous two timers ("previous and current"). So you need no error prone look-ahead logic.

This still means that we will always require n+1 iterations (which is why I extracted the logic into a method).

Also note, that the DrawTimer method will early return if the "current" is nullptr.
So assume we have a just one timer:
Init:
prev = null
current = null
Enter loop:
next = timer
Call Draw:
current == null -> return
Move pointer:
prev = current == nullptr
current = next == timer
Exit the loop:
next = nullptr
Call Draw:
current != null -> Draw the timer

Now assume we have two timers:
Init:
prev = null
current = null
Enter loop iteration 1:
next = timer1
Call Draw:
current == null -> return
Move pointer:
prev = current == nullptr
current = next == timer1
Enter loop iteration 2:
next = timer2
Call Draw:
current == timer1 ->Draw the timer
Move pointer:
prev = current == timer1
current = next == timer2
Exit the loop:
next = nullptr
Call Draw:
current == timer2 -> Draw the timer

florian-kuebler: Added some comments.

src/OrbitGl/TimerTrack.cpp
```diff
@@ -251,6 +251,13 @@ void TimerTrack::UpdatePrimitives(uint64_t min_tick, uint64_t max_tick,
 
   for (auto& chain : chains_by_depth) {
     if (!chain) continue;
+    // In order to draw overlaps correctly, we need for every text box to be drawn (current),
+    // its previous and next text box. In order to avoid looking ahead for the next text (which is
+    // error-prone), we are doing just one traversal of the text boxes, while keeping track of the
+    // previous two timers and will draw the previous one, thus the currents iteration value being
```
karupayun, line 257: // error-prone), we are doing just one traversal of the text boxes, while keeping track of the
    // previous two timers, thus the currents iteration value being the "next" textbox.

src/OrbitGl/TimerTrack.cpp
```diff
@@ -251,6 +251,13 @@ void TimerTrack::UpdatePrimitives(uint64_t min_tick, uint64_t max_tick,
...
+    // the "next" textbox.
+    // Note: This will require us to draw the last timer after the traversal of the text boxes.
+    // Also note: The draw method will take care of nullptr's being passed into.
```
karupayun, line 260: I would add "first iteration" or erase directly this "also note"
// Also note: The draw method will take care of nullptr's being passed into (first iteration).

src/OrbitGl/TimerTrack.cpp
```diff
@@ -263,6 +270,8 @@ void TimerTrack::UpdatePrimitives(uint64_t min_tick, uint64_t max_tick,
       if (!block.Intersects(min_tick, max_tick)) continue;
 
       for (size_t k = 0; k < block.size(); ++k) {
+        // Remember that the current index (k) points to the "next" text box and we want to draw
```
karupayun, line 273: // The current index (k) points to the "next" text box

karupayun (approved):
LGTM!

src/OrbitGl/TimerTrack.h
pierricgimmig: Since this is probably the code that runs the most in Orbit, there are also performance implications in passing all those arguments. The first six are passed through registers on Linux, but the rest is constantly being pushed and popped from the stack. It could make sense to bundle invariants in single struct that we would pass be reference to the DrawTimer function.

pierricgimmig (approved):
Hi @florian-kuebler , since this fixes a crash, let's get this in ASAP and iterate on the comments if need be. Thanks.

src/OrbitGl/TimerTrack.cpp
```diff
@@ -82,6 +82,147 @@ WorldXInfo ToWorldX(double start_us, double end_us, double inv_time_window, floa
...
+    Vec3 top_right(world_x_info_right_overlap.world_x_start, world_timer_y + box_height, z);
+    Vec3 bottom_right(
+        world_x_info_right_overlap.world_x_start + world_x_info_right_overlap.world_x_width,
+        world_timer_y, z);
+    batcher->AddShadedTrapezium(
+        top_left, bottom_left, bottom_right, top_right, color,
+        std::make_unique<PickingUserData>(current_text_box,
+                                          [&](PickingId id) { return this->GetBoxTooltip(id); }));
+
+  } else {
```
antonrohr, line 202: pers pref:
```suggestion
  return
  } 
```

antonrohr (approved):

src/OrbitGl/TimerTrack.h
florian-kuebler: Maybe adding `inline` annotation would help in this case. I know its not bullet proof, though.

src/OrbitGl/TimerTrack.cpp
florian-kuebler: Done.

src/OrbitGl/TimerTrack.cpp
florian-kuebler: The return is inside the lambda, so I guess I can't do the early return here.

src/OrbitGl/TimerTrack.h
karupayun: I agree with Pierric here, after the PR we should find a better way to organize the parameters.

ronaldfw: For performance: Make sure you measure, it's fairly straightforward with the introspection window.

florian-kuebler:
Merged this for now, but will do some measurements and maybe follow up.

PR #1845, opened 2021-02-11
Fix the selection behavior of frametrack

vickyliu-go4it:
After adding frametrack, when clicking the corresponding function in the live tab or the corresponding function calls in the capture window, the frametrack is also highlighted. This defeats the purpose of the frametrack colors (i.e., indicating long frames) and the selection should not influence the frametrack.

With this change, we disable the highlighting for the frametrack. More specifically:
- When clicking a function in live tab or a function call in the capture window, the color of frametrack will not be affected;
- When clicking the frametrack, the frametrack will not be highlighted, but the other function calls with the same function id in the capture window and the function in the live tab will still be highlighted.

Bug: http://b/179976757
Test: 1) create a frametrack for a function; 2) select the function in the live view, see the corresponding function calls in the thread tracks are highlighted but the frametrack is not affected; 3) select the frametrack, see the corresponding function calls in the thread tracks and the function in the live tab are highlighted, but the frametrack is not affected.

### DIFF
--- a/src/OrbitGl/TimerTrack.cpp
+++ b/src/OrbitGl/TimerTrack.cpp
@@ -27,6 +27,8 @@ using orbit_client_protos::TimerInfo;
 
 ABSL_DECLARE_FLAG(bool, show_return_values);
 
+const Color TimerTrack::kHighlightColor = Color(100, 181, 246, 255);
+
 TimerTrack::TimerTrack(TimeGraph* time_graph, TimeGraphLayout* layout, OrbitApp* app,
                        const CaptureData* capture_data)
     : Track(time_graph, layout, capture_data), app_{app} {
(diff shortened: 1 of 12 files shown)


vickyliu-go4it:
Before the change:
before

After the change:
after

src/OrbitGl/TimerTrack.cpp
```diff
@@ -175,7 +177,7 @@ void TimerTrack::DrawTimer(TextBox* prev_text_box, TextBox* current_text_box,
                         function_id == highlighted_function_id;
 
   static const Color kHighlightColor(100, 181, 246, 255);
```
karupayun, line 179: Please erase this constant if you are not using it :)

karupayun (approved):
LGTM Tingwei! :) 

Thanks for fixing this

src/OrbitGl/TimerTrack.cpp
vickyliu-go4it: Done. Thanks! I didn't notice this when rebasing after PR #1844.

PR #1846, opened 2021-02-11
Erase TimeGraph->GetBatcher() dependence

karupayun:
Part of http://b/176086740.

We are erasing TimeGraph->GetBatcher() dependence by adding the
timegraph-batcher as a parameter of UpdatePrimitives functions.

Also I included a few renamings to differentiate it with the UI-batcher
used in Draw functions. We should use also the batcher from the 
timegraph to draw the tracks themselves, but this will come in another PR.

Note that we are changing the batcher used in the FrameTrack to be consistent.

Test: Load a capture

### DIFF
--- a/src/OrbitGl/CaptureViewElement.h
+++ b/src/OrbitGl/CaptureViewElement.h
@@ -5,6 +5,7 @@
 #ifndef ORBIT_GL_CAPTURE_VIEW_ELEMENT_H_
 #define ORBIT_GL_CAPTURE_VIEW_ELEMENT_H_
 
+#include "Batcher.h"
 #include "PickingManager.h"
 
 class TimeGraph;
@@ -20,7 +21,7 @@ class CaptureViewElement : public Pickable {
     canvas_ = canvas;
   }
 
-  virtual void UpdatePrimitives(uint64_t /*min_tick*/, uint64_t /*max_tick*/,
+  virtual void UpdatePrimitives(Batcher* /*batcher*/, uint64_t /*min_tick*/, uint64_t /*max_tick*/,
                                 PickingMode /*picking_mode*/, float /*z_offset*/ = 0){};
 
   void SetTimeGraph(TimeGraph* timegraph) { time_graph_ = timegraph; }
--- a/src/OrbitGl/EventTrack.cpp
+++ b/src/OrbitGl/EventTrack.cpp
@@ -44,7 +44,7 @@ void EventTrack::Draw(GlCanvas* canvas, PickingMode picking_mode, float z_offset
     return;
   }
 
-  Batcher* batcher = canvas->GetBatcher();
+  Batcher* ui_batcher = canvas->GetBatcher();
 
   // The sample indicators are at z == 0 and do not respond to clicks, but
   // have a tooltip. For picking, we want to draw the event bar over them if
@@ -55,7 +55,7 @@ void EventTrack::Draw(GlCanvas* canvas, PickingMode picking_mode, float z_offset
   event_bar_z += z_offset;
   Color color = color_;
   Box box(pos_, Vec2(size_[0], -size_[1]), event_bar_z);
-  batcher->AddBox(box, color, shared_from_this());
+  ui_batcher->AddBox(box, color, shared_from_this());
 
   if (canvas->GetPickingManager().IsThisElementPicked(this)) {
     color = Color(255, 255, 255, 255);
@@ -66,8 +66,8 @@ void EventTrack::Draw(GlCanvas* canvas, PickingMode picking_mode, float z_offset
   float x1 = x0 + size_[0];
   float y1 = y0 - size_[1];
 
-  batcher->AddLine(pos_, Vec2(x1, y0), event_bar_z, color, shared_from_this());
-  batcher->AddLine(Vec2(x1, y1), Vec2(x0, y1), event_bar_z, color, shared_from_this());
+  ui_batcher->AddLine(pos_, Vec2(x1, y0), event_bar_z, color, shared_from_this());
+  ui_batcher->AddLine(Vec2(x1, y1), Vec2(x0, y1), event_bar_z, color, shared_from_this());
 
   if (picked_) {
     Vec2& from = mouse_pos_last_click_;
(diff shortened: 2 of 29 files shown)


src/OrbitGl/FrameTrack.cpp
```diff
@@ -245,9 +244,10 @@ void FrameTrack::Draw(GlCanvas* canvas, PickingMode picking_mode, float z_offset
   Vec2 white_text_box_position(pos_[0] + layout.GetRightMargin(),
                                y - layout.GetTextBoxHeight() / 2.f);
 
-  batcher->AddLine(from, from + Vec2(layout.GetRightMargin() / 2.f, 0), text_z, kWhiteColor);
-  batcher->AddLine(Vec2(white_text_box_position[0] + white_text_box_size[0], y), to, text_z,
-                   kWhiteColor);
+  Batcher* ui_batcher = canvas->GetBatcher();
```
karupayun, line 247: The batcher of the FrameTrack is changing here

src/OrbitGl/CaptureViewElement.h
```diff
@@ -9,6 +9,7 @@
 
 class TimeGraph;
 class GlCanvas;
+class Batcher;
```
florian-kuebler, line 12: Will we need this in the future?

florian-kuebler (approved):
Thanks! LGTM!

src/OrbitGl/CaptureViewElement.h
reichlfl: It this actually needed right now? Why can't you include Batcher.h here?

src/OrbitGl/EventTrack.cpp
```diff
@@ -44,7 +44,7 @@ void EventTrack::Draw(GlCanvas* canvas, PickingMode picking_mode, float z_offset
     return;
   }
 
-  Batcher* batcher = canvas->GetBatcher();
+  Batcher* ui_batcher = canvas->GetBatcher();
```
reichlfl, line 47: Have you thought about passing the Batcher here as well, so the track does not need to know about where batchers are managed or how many batchers the canvas actually uses?

Do we even need to pass the Canvas here if we do it like this? I would like to get rid of the GlCanvas* parameter here, because it is used very weirdly: We pass the canvas in `Draw`, but then we store it in `canvas_` because some other methods rely on it. These kind of side effects make testing hard.

src/OrbitGl/FrameTrack.cpp
```diff
@@ -199,8 +199,8 @@ std::string FrameTrack::GetTooltip() const {
       GetPrettyTime(absl::Nanoseconds(stats_.average_time_ns())));
 }
 
-std::string FrameTrack::GetBoxTooltip(PickingId id) const {
-  const TextBox* text_box = time_graph_->GetBatcher().GetTextBox(id);
+std::string FrameTrack::GetBoxTooltip(Batcher* batcher, PickingId id) const {
```
reichlfl, line 202: Why does the `GetBoxTooltip` receive a `Batcher*`, `GetSampleTooltip` a `Batcher&`?  If the parameter is intended to be not nullptr, `Batcher&` is preferrable.

src/OrbitGl/FrameTrack.cpp
reichlfl: Was this a bug before?

src/OrbitGl/Track.cpp
```diff
@@ -85,8 +85,7 @@ void Track::DrawTriangleFan(Batcher* batcher, const std::vector<Vec2>& points, c
 void Track::Draw(GlCanvas* canvas, PickingMode picking_mode, float z_offset) {
   CaptureViewElement::Draw(canvas, picking_mode, z_offset);
 
-  Batcher* batcher = canvas->GetBatcher();
-
+  Batcher* ui_batcher = canvas->GetBatcher();
```
reichlfl, line 88: Complete nitpick, feel free to ignore: The naming "ui_batcher" to me infers more knowledge than the track should really need, I thought "batcher" was fine before. I mostly stumbled across this because the `DrawTriangleFan` lines got reformatted :D

reichlfl (approved):
Nice change, thanks a lot! PTAL at the comment regarding `Draw` - not sure if I'm missing something and this can't be changed right now, or if it should just be changed in a subsequent PR.

src/OrbitGl/FrameTrack.cpp
karupayun: I believe that, but I'm not sure and added @ronaldfw because I wanted his opinion. 

Is the only draw function which used the timegraph batcher, and this batcher is only cleaned when we make UpdatePrimitives, so I think we were drawing these lines several times.

But this couldn't make any real performance implication neither causing a crash.

src/OrbitGl/CaptureViewElement.h
karupayun: Not, isn't needed at all! Sorry, I probably saw the other classes and completed automatically without thinking.

src/OrbitGl/EventTrack.cpp
karupayun: Yes, very weirdly. The next step is passing Batcher y TextRenderer here instead of the Canvas, but I'm still thinking about how to solve this "strange storing of canvas".

src/OrbitGl/FrameTrack.cpp
karupayun: Thanks for pointing out. Both should be const &. I tried but used to have a problem with some method that weren't const.

I've modified it now.

src/OrbitGl/Track.cpp
karupayun: mmm, I will disregard this. We used to have some issues with using both batchers without understanding which we had to use. 

I will change this after removing canvas here.

karupayun:
> Nice change, thanks a lot! PTAL at the comment regarding `Draw` - not sure if I'm missing something and this can't be changed right now, or if it should just be changed in a subsequent PR.

Draw methods don't use time_graph->Batcher(), so we don't need to add the batcher there. It's only if we want to erase dependence to GlCanvas (which we want) :)

PR #1847, opened 2021-02-11
Fix when open disassembly while capturing.

florian-kuebler:
App's `sampling_report_` and `capture_data_` are tightly coupled
with both containing the post processed data.
While capturing we end up in having a non-null `sampling_report_`,
that does not yet have any data. In this situation `capture_data_`
does not contain any post processed data at all.

While a larger refactoring is needed, this change ensures for the
current situation two things:
1. Whenever sampling_report_ is not nullptr, capture_data_ contains
post processed sampling data.
2. When accessing post processed sampling data from CaptureData,
we check if it is present directly, rather then checking if there
is a sampling report.

Bug: http://b/179902321
Test: Open disassembly while taking the first capture does not crash.

### DIFF
--- a/src/OrbitClientModel/include/OrbitClientModel/CaptureData.h
+++ b/src/OrbitClientModel/include/OrbitClientModel/CaptureData.h
@@ -198,6 +198,10 @@ class CaptureData {
 
   [[nodiscard]] const ProcessData* process() const { return &process_; }
 
+  [[nodiscard]] bool has_post_processed_sampling_data() const {
+    return post_processed_sampling_data_.has_value();
+  }
+
   [[nodiscard]] const PostProcessedSamplingData& post_processed_sampling_data() const {
     CHECK(post_processed_sampling_data_.has_value());
     return post_processed_sampling_data_.value();
--- a/src/OrbitGl/App.cpp
+++ b/src/OrbitGl/App.cpp
@@ -533,7 +533,7 @@ void OrbitApp::Disassemble(int32_t pid, const FunctionInfo& function) {
     Disassembler disasm;
     disasm.AddLine(absl::StrFormat("asm: /* %s */", function_utils::GetDisplayName(function)));
     disasm.Disassemble(memory.data(), memory.size(), absolute_address, is_64_bit);
-    if (!sampling_report_) {
+    if (!HasCaptureData() || !GetCaptureData().has_post_processed_sampling_data()) {
       DisassemblyReport empty_report(disasm);
       SendDisassemblyToUi(disasm.GetResult(), std::move(empty_report));
       return;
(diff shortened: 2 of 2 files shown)


src/OrbitClientModel/include/OrbitClientModel/CaptureData.h
```diff
@@ -198,6 +198,10 @@ class CaptureData {
 
   [[nodiscard]] const ProcessData* process() const { return &process_; }
 
+  [[nodiscard]] bool has_post_processed_sampling_data() const {
```
dpallotti, line 201: Aaand there you can see why I don't like these snake_case exceptions. This already doesn't fit in the snake_case cases that the style guide allows but at the same time it would be very weird to call it `HasPostProcessedSamplingData`. Just a little rant, ignore it.

src/OrbitGl/App.cpp
```diff
@@ -1658,12 +1658,14 @@ void OrbitApp::UpdateAfterCaptureCleared() {
   PostProcessedSamplingData empty_post_processed_sampling_data;
   absl::flat_hash_map<CallstackID, std::shared_ptr<CallStack>> empty_unique_callstacks;
 
-  SetSamplingReport(empty_post_processed_sampling_data, empty_unique_callstacks);
+  if (sampling_report_ != nullptr) {
+    SetSamplingReport(empty_post_processed_sampling_data, empty_unique_callstacks);
+  }
```
dpallotti, line 1663: Is this part of the fix? I'm a bit confused by it, mostly because I'm not up to date with this code. When is `sampling_report_ != nullptr` if we just cleared the capture? Or when is it `sampling_report_ == nullptr` when we just cleared the capture? How can we end up in two different states after clearing a capture?

dpallotti (approved):
Thanks Florian. 

I'm a bit puzzled by the change in `UpdateAfterCaptureCleared`, mostly because I haven't followed the evolution of the client's model recently (but I'm fine with it).

src/OrbitClientModel/include/OrbitClientModel/CaptureData.h
florian-kuebler: Well thats still the official snake_case google style.

florian-kuebler: I mean it is still a very basic accessor.

src/OrbitGl/App.cpp
florian-kuebler: That is actually enough to do the fix. Before this, we've set sampling report (no matter what it was before) to "empty" sampling report on a start of the capture. In capture_data_ the post processed data is still completely missing, and will only appear after the capture. In the disassemble method we had then a check for sampling_report_ being nullptr and if not access capture data's post processed data. So by this line, sampling_report_ was not null (but empty) and capture data was still not having the data -> crash.

src/OrbitClientModel/include/OrbitClientModel/CaptureData.h
dpallotti: Well, kind of, it explicitly says "get and set functions". It's fine, protos also use `has_`. But when you add the fact that the two styles are mixed and the fact that it's not super clear what functions are included, it would just be cleaner to have a single style.

src/OrbitGl/App.cpp
dpallotti: Alright, thanks for the explanation. Yes, a larger refactor would be needed...

florian-kuebler: Agree, but tbh. I do have a couple of crashes on my stack right now, so lets fix this first and do better later.

src/OrbitClientModel/include/OrbitClientModel/CaptureData.h
florian-kuebler: That's said. In general I agree, and I don't like the style guides "use snake case" suggestion. However, we started with this, so I want to be strict with that, as otherwise, it will be even more mixture.

src/OrbitGl/App.cpp
dpallotti: Of course, of course, I wasn't suggesting we should tackle it right now!

PR #1848, opened 2021-02-11
Fix scheduler track dragging

reichlfl:
Changing the base class of Track resulted in the Scheduler track being draggable, which is not intended.

To reproduce:
Take a capture, drag the scheduler track. This should not be possible.

### DIFF
--- a/contrib/automation_tests/orbit_tracks.py
+++ b/contrib/automation_tests/orbit_tracks.py
@@ -23,6 +23,7 @@ def main(argv):
         SelectTrack(track_index=5),
         DeselectTrack(),
         SelectTrack(track_index=0, expect_failure=True),   # Scheduler track cannot be selected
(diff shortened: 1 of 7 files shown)


karupayun (commented):
I'm not sure if this is consistent. If we make this change, we shouldn't allow other tracks to be moved on top of the Scheduler, what do you think?

pierricgimmig (approved):
LGTM, thanks.

reichlfl:
> I'm not sure if this is consistent. If we make this change, we shouldn't allow other tracks to be moved on top of the Scheduler, what do you think?

True... now that you mention it, I've checked back with a previous release and the scheduler was draggable there as well. I thought this wasn't the case and I broke it with my latest refactor, but the current state is intended behavior. Don't know how I reached that conclusion though :-)

I'll close this, thanks a lot for noticing!

PR #1849, opened 2021-02-11
[Rebase to release/1.59] Fix load_bias calculation

dimitry-:
Calculate load_bias based on the executable PT_LOAD segment.

Newer linkers create first segment as read only and
put .text section into a separate segment. Which means
load_bias calculation is off if second segment vaddr-offset
difference is different from the first PT_LOAD segment

Test: run unit-tests
Test: run capture on executable with executable PT_LOAD
      segment different from the first PT_LOAD
Bug: http://b/179894572


dpallotti (approved):

florian-kuebler:
Thanks for tackling this, but just to be sure:
Do we require this for the 1.59 release? I thought the llvm10 was moved to 1.60?

dpallotti:
> Thanks for tackling this, but just to be sure:
> Do we require this for the 1.59 release? I thought the llvm10 was moved to 1.60?

I think that given b/180003371 the earlier we get it in the better.

florian-kuebler:
Well if llvm10 is not part of 1.59 I would disagree, as it woun't be an issue for Stadia devs, which we are mainly targeting with the releases. But we should discuss offline.

PR #1851, opened 2021-02-11
Remove CHECK that breaks introspection

pierricgimmig:


reichlfl (approved):
LGTM, thanks for fixing

florian-kuebler (approved):
Thanks for that check. LGTM. However, should we try to move towards also having a "capture data" in the introspection? Seeing similar problems raising over again in the future.

pierricgimmig:
> Thanks for that check. LGTM. However, should we try to move towards also having a "capture data" in the introspection? Seeing similar problems raising over again in the future.

Yeah we should probably do that, I'll start thinking about what is involved, thanks.

PR #1852, opened 2021-02-12
Add ScopeTree hierarchical data structure for profiling scopes

pierricgimmig:
ScopeTree is a layer of abstraction above existing scope data. It
provides hierarchical relationship between profiling scopes. The goal
is to be able to generate the scope tree with different streams of
scope data that can arrive out of order. The ScopeTree also maintains
and ordered map of nodes per depth.

As scopes come in, we insert them into the ScopeTree by first finding
the "deepest parent" that completely encompasses the node to add.
We then make the new node child of that parent and migrate the 
original parent's children to the new node. See image below:

image

See usage in ThreadTrack.h/.cpp in the liborbit branch: https://github.com/pierricgimmig/orbit/blob/feature/liborbit/src/OrbitGl/ThreadTrack.cpp#L75
https://github.com/pierricgimmig/orbit/blob/feature/liborbit/src/OrbitGl/ThreadTrack.cpp#L128


src/OrbitGl/ScopeTree.cpp
```diff
@@ -0,0 +1,181 @@
...
+      node->text_box_->GetTimerInfo().end());
+  for (auto& pair : node->children_) {
+    Print(pair.second, pair.first, depth + 1);
+  }
+}
+
+void ScopeNode::FindHeight(const ScopeNode* node, uint32_t* height, uint32_t current_height) {
+  ORBIT_SCOPE_FUNCTION;
+  *height = std::max(*height, current_height);
+  for (auto& pair : node->children_) {
```
karupayun, line 80: same here

src/OrbitGl/ScopeTree.cpp
```diff
@@ -0,0 +1,181 @@
+// Copyright (c) 2021 The Orbit Authors. All rights reserved.
```
florian-kuebler, line 1: General question regarding the implementation:
There are a couple of recursive functions, that write the results into a parameter rather than returning them, is this because auf tail-call optimization? Have you checked if those functions get tail-call optimized? I'd assume that should be critical for performance. 
If so, maybe add a comment.

src/OrbitGl/ScopeTree.cpp
```diff
@@ -0,0 +1,181 @@
...
+
+void ScopeNode::GetAllNodesInSubtree(const ScopeNode* node, std::set<const ScopeNode*>* node_set) {
+  CHECK(node_set != nullptr);
+  node_set->insert(node);
+  for (const auto [unused_time, child] : node->children_) {
+    GetAllNodesInSubtree(child, node_set);
+  }
+}
+
+uint32_t ScopeNode::Height() const {
```
florian-kuebler, line 113: Could you move this method to the other FindHeight method, such that they are close to each other?

src/OrbitGl/ScopeTree.cpp
```diff
@@ -0,0 +1,181 @@
...
+
+ScopeNode* ScopeNode::GetLastChildBeforeOrAtTime(uint64_t time) const {
+  // Get first child before or exactly at "time".
+  if (children_.empty()) return nullptr;
+  auto next_node_it = children_.upper_bound(time);
+  if (next_node_it == children_.begin()) return nullptr;
+  return (--next_node_it)->second;
+}
+
+ScopeNode* ScopeNode::FindDeepestParentForNode(const ScopeNode* node) {
```
florian-kuebler, line 127: Nit question: Is there a reason, why you are not doing this call recursive anymore? Personally, I'd find it probably a bit more readable (I required 3 attempts to fully grasp this method -- but still this method is nice and short, so still totally fine).

src/OrbitGl/ScopeTree.cpp
```diff
@@ -0,0 +1,181 @@
...
+  ORBIT_SCOPE_FUNCTION;
+  *height = std::max(*height, current_height);
+  for (auto& pair : node->children_) {
+    FindHeight(pair.second, height, current_height + 1);
+  }
+}
+
+size_t ScopeNode::CountNodesInSubtree() const {
+  size_t num_nodes = 0;
+  CountNodesInSubtree(this, &num_nodes);
```
karupayun, line 87: Why not making a for loop and sum for each child, child->CountNodesInSubtree?

Then, you don't need the extra function with 2 args.

src/OrbitGl/ScopeTree.cpp
```diff
@@ -0,0 +1,181 @@
...
+      parent_node->GetChildrenInRange(node_to_insert->Start(), node_to_insert->End());
+
+  // Remove all encompassed children from the parent node.
+  for (ScopeNode* encompassed_node : encompassed_nodes) {
+    parent_node->children_.erase(encompassed_node->Start());
+  }
+
+  // Make all encompassed nodes children of the new node.
+  for (ScopeNode* encompassed_node : encompassed_nodes) {
+    node_to_insert->children_.insert({encompassed_node->Start(), encompassed_node});
```
florian-kuebler, line 176: emplace, here and below?

florian-kuebler (commented):
Very nice data structure. Excited to see it in use! Comments are mostly questions/nitpicking nothing show stopping. However, as we are going to use that at least for the timer tracks, a couple of additional tests would be great!

src/OrbitGl/ScopeTree.cpp
```diff
@@ -0,0 +1,181 @@
...
+      deepest_node = current_node;
+    }
+  }
+  return deepest_node;
+}
+
+std::vector<ScopeNode*> ScopeNode::GetChildrenInRange(uint64_t start, uint64_t end) const {
+  // Get children that are enclosed by start and end inclusively.
+  if (children_.empty()) return {};
+  auto node_it = children_.lower_bound(start);
```
karupayun, line 142: Nit: What about a for-loop here?
```
std::vector<ScopeNode*> nodes;
for (auto node_it = children_.lower_bound(start); node_it != children_.end(); node_it++){
    ScopeNode* node = node_it->second;
    if (node->Start() >= start && node->End() <= end) {
      nodes.push_back(node);
    } else {
      break;
    }
}
return nodes;
```

karupayun (approved):
I really like it Pierric! :) :)

I left a few comments, but looking forward to see it working :)

src/OrbitGl/TextBox.h
```diff
@@ -34,6 +34,7 @@ class TextBox {
     timer_info_ = timer_info;
   }
   const orbit_client_protos::TimerInfo& GetTimerInfo() const { return timer_info_; }
+  orbit_client_protos::TimerInfo& GetMutableGetTimerInfo() { return timer_info_; }
```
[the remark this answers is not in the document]
pierricgimmig: Oh yes, thanks.

src/OrbitGl/ScopeTree.h
```diff
@@ -0,0 +1,83 @@
...
+
+// ScopeTree is a layer of abstraction above existing TextBox data. It provides hierachical
+// relationship between profiling scopes. The goal is to be able to generate the scope tree with
+// different streams of scope data that can arrive out of order. The ScopeTree also maintains an
+// ordered map of nodes per depth.
+
+class ScopeNode {
+ public:
+  ScopeNode() = default;
+  ScopeNode(TextBox* text_box, uint64_t id) : text_box_(text_box), id_(id) {}
```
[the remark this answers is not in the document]
pierricgimmig: It was actually to help debug, but it's not needed, I removed it, I'll save memory. Thanks for pointing it out.

src/OrbitGl/ScopeTree.cpp
pierricgimmig: Done, thanks.

src/OrbitGl/ScopeTree.cpp
pierricgimmig: Yes, it's nicer, thanks, changed it.

src/OrbitGl/ScopeTree.cpp
pierricgimmig: Yes! Done, thanks.

src/OrbitGl/ScopeTree.h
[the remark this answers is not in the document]
pierricgimmig: Yeah... I was asking the same question, it's cleaner to make it generic. Done.

src/OrbitGl/ScopeTree.cpp
```diff
@@ -0,0 +1,181 @@
...
+    UpdateDepthInSubtree(child_node, new_depth + 1);
+  }
+
+  // Add node to new depth track.
+  ordered_nodes_by_depth_[new_depth].insert({node_timestamp, node});
+}
+
+size_t ScopeTree::CountOrderedNodes() const {
+  size_t count_from_depth = 0;
+  for (auto& pair : ordered_nodes_by_depth_) {
```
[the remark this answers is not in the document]
pierricgimmig: Done, thanks.

src/OrbitGl/ScopeTree.cpp
```diff
@@ -0,0 +1,181 @@
...
+      LOG("%u: node_depth:%u id:%u", depth, node->Depth(), node->Id());
+    }
+  }
+}
+
+void ScopeNode::Print(const ScopeNode* node, uint64_t start_time, uint32_t depth) {
+  LOG("d%u %s%lu ScopeNode(%u) %p [%lu, %lu]", node->Depth(), std::string(depth, ' '), start_time,
+      node->id_, node, node->text_box_->GetTimerInfo().start(),
+      node->text_box_->GetTimerInfo().end());
+  for (auto& pair : node->children_) {
```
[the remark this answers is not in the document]
pierricgimmig: Done, thanks.

src/OrbitGl/ScopeTree.cpp
pierricgimmig: Done, thanks.

src/OrbitGl/ScopeTree.h
```diff
@@ -0,0 +1,83 @@
...
+  static void Print(const ScopeNode* node, uint64_t start_time = 0, uint32_t depth = 0);
+  static void FindHeight(const ScopeNode* node, uint32_t* height, uint32_t current_height = 0);
+  static void CountNodesInSubtree(const ScopeNode* node, size_t* count);
+  static void GetAllNodesInSubtree(const ScopeNode* node, std::set<const ScopeNode*>* node_set);
+
+ private:
+  TextBox* text_box_ = nullptr;
+  uint64_t id_ = 0;
+  uint32_t depth_ = 0;
+  std::map<uint64_t, ScopeNode*> children_;
```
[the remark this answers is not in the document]
pierricgimmig: Renamed to children_by_start_time_. Concerning absl::btree_map, it's a drop in replacement for std::map, but due to a compiler bug in VS2017, we can't use it yet.

src/OrbitGl/ScopeTree.h
```diff
@@ -0,0 +1,83 @@
...
+ public:
+  ScopeTree();
+  void Insert(TextBox* timer_info);
+  void Print() const;
+
+  [[nodiscard]] const ScopeNode* Root() const { return root_; }
+  [[nodiscard]] size_t Size() const { return size_; }
+  [[nodiscard]] size_t CountOrderedNodes() const;
+  [[nodiscard]] uint32_t Height() const { return root_->Height(); }
+  [[nodiscard]] const std::map<uint32_t, std::map<uint64_t, ScopeNode*>>& GetOrderedNodesByDepth()
```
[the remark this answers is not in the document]
pierricgimmig: done. See previous comment about absl. Thanks.

src/OrbitGl/ScopeTreeTest.cpp
```diff
@@ -0,0 +1,111 @@
...
+  EXPECT_EQ(ordered_nodes_by_depth.at(1).size(), 1);  // node 0
+  EXPECT_EQ(ordered_nodes_by_depth.at(2).size(), 3);  // nodes 1, 2 and 3
+  ValidateTree(tree);
+}
+
+TEST(ScopeTree, EmptyTree) {
+  ScopeTree tree;
+  ValidateTree(tree);
+}
+
```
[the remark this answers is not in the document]
pierricgimmig: I added a test that creates random shuffles of the same scopes and makes sure that the resulting trees are all the same. PTAL.

florian-kuebler (approved):
Had another look, LGTM! Thanks.

src/OrbitGl/ScopeTreeTest.cpp
```diff
@@ -28,6 +29,40 @@ TestScope* CreateScope(uint64_t start, uint64_t end) {
...
+  std::vector<TestScope*>* scope_buffer;
+};
+
+void CreateNonOverlappingTestTree(size_t max_num_nodes, size_t max_depth,
+                                  std::vector<TestScope*>* scope_buffer, size_t depth = 0) {
+  if (depth > max_depth) return;
+  if (scope_buffer->size() >= max_num_nodes) return;
+
+  ScopeTimer timer(scope_buffer, max_num_nodes);
+  constexpr int kNumSiblingNodes = 4;
```
karupayun, line 60: Why not having this as a parameter also?

src/OrbitGl/ScopeTreeTest.cpp
```diff
@@ -28,6 +29,40 @@ TestScope* CreateScope(uint64_t start, uint64_t end) {
...
+  ~ScopeTimer() {
+    if (scope_buffer->size() < max_num_nodes) {
+      scope_buffer->push_back(CreateScope(start, GetFakeTimeStamp()));
+    }
+  }
+  uint64_t start;
+  size_t max_num_nodes;
+  std::vector<TestScope*>* scope_buffer;
+};
+
```
karupayun, line 53: A little comment could help reader to understand, something like:
"Create a test tree using DFS-algorithm to assure that for each pair of intervals, one is completely inside of the other (father-children) or they don't overlap at all"

src/OrbitGl/ScopeTreeTest.cpp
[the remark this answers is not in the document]
karupayun: Perfect Pierric! :)

src/OrbitGl/ScopeTreeTest.cpp
pierricgimmig: Done, thanks.

src/OrbitGl/ScopeTreeTest.cpp
pierricgimmig: Done. Thanks.

PR #1853, opened 2021-02-12
Introducing Future<ErrorMessageOr<T>>::ThenIfSuccess

beckerhe:
`ThenIfSuccess` allows to schedule a continuation that is only executed when the returned `ErrorMessageOr<T>` return type does not contain an error. In case of an error the continuation won't run and the error message will be returned immediately. That means the return type of `ThenIfSuccess` will be a future with `ErrorMessageOr<R>` as its value type, while the `R` is determined by the continuation.

Example:
```cpp
Future<ErrorMessageOr<std::filesystem::path>> file_path = LoadModule("/tmp/OrbitService");

Future<ErrorMessageOr<void>> result = file_path.ThenIfSuccess(main_thread_executor,
                                 [this](const std::filesystem::path& local_file_path) {
  LoadSymbols(local_file_path);
});
```

This allows to chain continuations that only need to take care of the happy path which leads to cleaner code.

### DIFF
--- a/src/OrbitBase/PromiseHelpersTest.cpp
+++ b/src/OrbitBase/PromiseHelpersTest.cpp
@@ -76,4 +76,136 @@ TEST(GetResultFromFutureAndCallContinuation, CallWithResult) {
   EXPECT_TRUE(called);
 }
 
+TEST(HandleErrorAndSetResultInPromise, SuccessVoidInVoidOut) {
+  ErrorMessageOr<void> in{outcome::success()};
+
+  const auto invocable = []() -> void {};
+
+  Promise<ErrorMessageOr<void>> promise;
+  HandleErrorAndSetResultInPromise<ErrorMessageOr<void>> helper{&promise};
+  helper.Call(invocable, in);
+
+  auto future = promise.GetFuture();
+  ASSERT_TRUE(future.IsFinished());
+  EXPECT_TRUE(future.Get().has_value());
+}
+
+TEST(HandleErrorAndSetResultInPromise, SuccessVoidInIntOut) {
+  ErrorMessageOr<void> in{outcome::success()};
+
+  const auto invocable = []() -> int { return 42; };
+
+  Promise<ErrorMessageOr<int>> promise;
+  HandleErrorAndSetResultInPromise<ErrorMessageOr<int>> helper{&promise};
+  helper.Call(invocable, in);
+
+  auto future = promise.GetFuture();
+  ASSERT_TRUE(future.IsFinished());
+  ASSERT_TRUE(future.Get().has_value());
+  EXPECT_EQ(future.Get().value(), 42);
+}
+
+TEST(HandleErrorAndSetResultInPromise, SuccessIntInVoidOut) {
+  ErrorMessageOr<int> in{42};
+
+  const auto invocable = [](int value) { EXPECT_EQ(value, 42); };
+
+  Promise<ErrorMessageOr<void>> promise;
+  HandleErrorAndSetResultInPromise<ErrorMessageOr<void>> helper{&promise};
+  helper.Call(invocable, in);
+
+  auto future = promise.GetFuture();
+  ASSERT_TRUE(future.IsFinished());
+  EXPECT_TRUE(future.Get().has_value());
+}
+
+TEST(HandleErrorAndSetResultInPromise, SuccessIntInIntOut) {
+  ErrorMessageOr<int> in{42};
+
+  const auto invocable = [](int value) -> int {
+    EXPECT_EQ(value, 42);
+    return 42;
+  };
+
+  Promise<ErrorMessageOr<int>> promise;
+  HandleErrorAndSetResultInPromise<ErrorMessageOr<int>> helper{&promise};
+  helper.Call(invocable, in);
+
+  auto future = promise.GetFuture();
+  ASSERT_TRUE(future.IsFinished());
+  ASSERT_TRUE(future.Get().has_value());
+  EXPECT_EQ(future.Get().value(), 42);
+}
+
+TEST(HandleErrorAndSetResultInPromise, FailureVoidInVoidOut) {
+  ErrorMessageOr<void> in{ErrorMessage{"Error"}};
+
+  const auto invocable = []() -> void { FAIL(); };
+
+  Promise<ErrorMessageOr<void>> promise;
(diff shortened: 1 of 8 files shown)


beckerhe:
And once more a bunch of (at least unit tested) new code. Thanks in advance for the review.

src/OrbitBase/include/OrbitBase/PromiseHelpers.h
```diff
@@ -36,6 +40,42 @@ struct CallTaskAndSetResultInPromise<void> {
...
+// arguments.
+//
+// The main benefit of this is the specialization for `T = void`, which requires different syntax.
+// Don't use it if you don't need that particular feature.
+template <typename R>
+struct HandleErrorAndSetResultInPromise {
+  orbit_base::Promise<R>* promise;
+
+  template <typename Invocable, typename T>
+  void Call(Invocable&& invocable, const ErrorMessageOr<T>& result) {
```
dimitry-, line 55: s/result/argument|input/?

src/OrbitBase/PromiseHelpersTest.cpp
```diff
@@ -76,4 +76,128 @@ TEST(GetResultFromFutureAndCallContinuation, CallWithResult) {
...
+  auto future = promise.GetFuture();
+  ASSERT_TRUE(future.IsFinished());
+  EXPECT_TRUE(future.Get().has_value());
+}
+
+TEST(HandleErrorAndSetResultInPromise, SuccessIntInIntOut) {
+  ErrorMessageOr<int> in{42};
+
+  const auto invocable = [](int value) -> int {
+    EXPECT_EQ(value, 42);
```
dimitry-, line 126: interesting, I remember ASSERTS not compiling when called from a function returning non-void. I take it they fixed this.

src/OrbitBase/PromiseHelpersTest.cpp
```diff
@@ -76,4 +76,128 @@ TEST(GetResultFromFutureAndCallContinuation, CallWithResult) {
...
+  auto future = promise.GetFuture();
+  ASSERT_TRUE(future.IsFinished());
+  ASSERT_TRUE(future.Get().has_value());
+  EXPECT_EQ(future.Get().value(), 42);
+}
+
+TEST(HandleErrorAndSetResultInPromise, FailureVoidInVoidOut) {
+  ErrorMessageOr<void> in{ErrorMessage{"Error"}};
+
+  const auto invocable = []() -> void {};
```
dimitry-, line 143: FAIL if this is called?

src/OrbitBase/PromiseHelpersTest.cpp
```diff
@@ -76,4 +76,128 @@ TEST(GetResultFromFutureAndCallContinuation, CallWithResult) {
...
+  auto future = promise.GetFuture();
+  ASSERT_TRUE(future.IsFinished());
+  ASSERT_TRUE(future.Get().has_error());
+  EXPECT_EQ(future.Get().error().message(), "Error");
+}
+
+TEST(HandleErrorAndSetResultInPromise, FailureVoidInIntOut) {
+  ErrorMessageOr<void> in{ErrorMessage{"Error"}};
+
+  const auto invocable = []() -> int { return 42; };
```
dimitry-, line 158: FAIL?

src/OrbitBase/PromiseHelpersTest.cpp
```diff
@@ -76,4 +76,128 @@ TEST(GetResultFromFutureAndCallContinuation, CallWithResult) {
...
+  auto future = promise.GetFuture();
+  ASSERT_TRUE(future.IsFinished());
+  ASSERT_TRUE(future.Get().has_error());
+  EXPECT_EQ(future.Get().error().message(), "Error");
+}
+
+TEST(HandleErrorAndSetResultInPromise, FailureIntInVoidOut) {
+  ErrorMessageOr<int> in{ErrorMessage{"Error"}};
+
+  const auto invocable = [](int value) { EXPECT_EQ(value, 42); };
```
dimitry-, line 173: and probably here

src/QtUtils/MainThreadExecutorImplTest.cpp
```diff
@@ -154,4 +156,81 @@ TEST(MainThreadExecutorImpl, TryScheduleWithInvalidWeakPtr) {
...
+  EXPECT_FALSE(chained_future.IsFinished());
+
+  constexpr const char* const kErrorMessage{"Error"};
+  promise.SetResult(ErrorMessage{kErrorMessage});
+  QCoreApplication::processEvents();
+  EXPECT_FALSE(called);
+  EXPECT_TRUE(chained_future.IsFinished());
+  EXPECT_TRUE(chained_future.Get().has_error());
+  EXPECT_EQ(chained_future.Get().error().message(), kErrorMessage);
+}
```
dimitry-, line 175: Add a test for a chain of 2 futures with ScheduleAfterIfSuccess?

dimitry- (approved):
otherwise LGTM!

src/OrbitBase/include/OrbitBase/PromiseHelpers.h
beckerhe: Oh yes. Sorry for that, Done.

src/OrbitBase/PromiseHelpersTest.cpp
beckerhe: As far as I know `FAIL` and `ASSERT` won't work, but `EXPECT` is fine.

src/OrbitBase/PromiseHelpersTest.cpp
beckerhe: Yes. Done.

src/OrbitBase/PromiseHelpersTest.cpp
beckerhe: Yep. Done.

src/QtUtils/MainThreadExecutorImplTest.cpp
beckerhe: Good idea. Done.

src/OrbitBase/PromiseHelpersTest.cpp
beckerhe: Done.

PR #1854, opened 2021-02-12
Add retry to ssh tunnel connect

antonrohr:
Orbit now retries to establish a ssh tunnel 3 times. There is a waiting
time of 500ms between each retry.

Also restores the OrbitService version logging.

### DIFF
--- a/src/OrbitQt/servicedeploymanager.cpp
+++ b/src/OrbitQt/servicedeploymanager.cpp
@@ -182,7 +182,7 @@ outcome::result<uint16_t> ServiceDeployManager::StartTunnel(
     std::optional<orbit_ssh_qt::Tunnel>* tunnel, uint16_t port) {
   CHECK(QThread::currentThread() == thread());
   emit statusMessage("Setting up port forwarding...");
-  LOG("Setting up tunnel on port %d", grpc_port_.grpc_port);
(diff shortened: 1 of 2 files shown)


danielfenner (approved):
lgtm!

src/OrbitQt/servicedeploymanager.cpp
```diff
@@ -551,7 +551,17 @@ outcome::result<ServiceDeployManager::GrpcPort> ServiceDeployManager::ExecImpl()
         "running...");
   }
 
-  OUTCOME_TRY(local_grpc_port, StartTunnel(&grpc_tunnel_, grpc_port_.grpc_port));
+  outcome::result<uint16_t> local_grpc_port_result =
+      StartTunnel(&grpc_tunnel_, grpc_port_.grpc_port);
+  int retry = 3;
+  while (retry > 0 && local_grpc_port_result.has_error()) {
+    LOG("Failed to establish tunnel. Trying again in 500ms");
```
dpallotti, line 558: I personally see this more as an `ERROR`.

dpallotti (approved):
Thank you!

src/OrbitQt/servicedeploymanager.cpp
antonrohr: Fair point. I force pushed

PR #1856, opened 2021-02-12
Coverage docker & kokoro

antonrohr:
The docker container is already pushed to the cloud project. The successful test is here:

https://fusion.corp.google.com/runanalysis/info/prod:orbitprofiler%2Fbuilds%2Flinux%2Fclang7_relwithdebinfo%2Fcontinuous/prod:orbitprofiler%2Fbuilds%2Flinux%2Fclang7_relwithdebinfo%2Fcontinuous/KOKORO/a888eaee-c02f-44c5-9039-c75b10fa0855/1613132417885/build%20%23709/Targets

Bucket:
https://pantheon.corp.google.com/storage/browser/orbitprofiler-build/gcp_ubuntu_artifacts/prod/orbitprofiler/builds/linux/clang7_relwithdebinfo/continuous/709/20210212-042019;tab=objects?prefix=&forceOnObjectsSortingFiltering=false

(See coverage sub folder)

### DIFF
--- a/kokoro/builds/build.sh
+++ b/kokoro/builds/build.sh
@@ -142,6 +142,13 @@ if [ -n "$1" ]; then
   echo "Starting the build (conan build)."
   conan build -bf "${REPO_ROOT}/build/" "${REPO_ROOT}"
 
+  if [[ $CONAN_PROFILE == "coverage_clang9" ]]; then 
+    echo "Starting to generate unit test coverage report"
+    mkdir -p "${REPO_ROOT}/build/package/coverage"
+    ${REPO_ROOT}/contrib/unit_test_coverage/generate_coverage_report.sh \
+            "${REPO_ROOT}/src" "${REPO_ROOT}/build" "${REPO_ROOT}/build/package/coverage"
+  fi
+
   if [[ $CONAN_PROFILE != "iwyu" ]]; then
     echo "Start the packaging (conan package)."
     conan package -bf "${REPO_ROOT}/build/" "${REPO_ROOT}"
--- a/kokoro/builds/build.sh
+++ b/kokoro/builds/build.sh
@@ -142,14 +142,7 @@ if [ -n "$1" ]; then
   echo "Starting the build (conan build)."
   conan build -bf "${REPO_ROOT}/build/" "${REPO_ROOT}"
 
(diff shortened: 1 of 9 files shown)


kokoro/builds/build.sh
```diff
@@ -142,6 +142,13 @@ if [ -n "$1" ]; then
...
   conan build -bf "${REPO_ROOT}/build/" "${REPO_ROOT}"
 
+  if [[ $CONAN_PROFILE == "coverage_clang9" ]]; then 
+    echo "Starting to generate unit test coverage report"
+    mkdir -p "${REPO_ROOT}/build/package/coverage"
+    ${REPO_ROOT}/contrib/unit_test_coverage/generate_coverage_report.sh \
+            "${REPO_ROOT}/src" "${REPO_ROOT}/build" "${REPO_ROOT}/build/package/coverage"
+  fi
+
   if [[ $CONAN_PROFILE != "iwyu" ]]; then
```
beckerhe, line 152: I think we don't want to call `conan package` when using the `coverage_clang9`, just like include-what-you-use.

If we did, we would store all the binaries in the bucket, again. There is another job doing that already.

beckerhe: The same applies to the ZIP file generation in line 206. This has to be exluded as well.

beckerhe: One more thing. Could you put this part further down in the script? Maybe after the `Analyze include-what-you-use` results?

beckerhe (commented):
Very nice. Thanks a lot, Anton!

I added some minor comments regarding the build.sh script. The rest looks good!

beckerhe:
One more thing: You might wanna add the kokoro build config to this PR. (Could also be a separate PR)

kokoro/builds/build.sh
antonrohr: I fixed this and tested that it works.

antonrohr:
> One more thing: You might wanna add the kokoro build config to this PR. (Could also be a separate PR)

Came to the same conclusion while looking at the kokoro scripts. I added them here. ptal

beckerhe (approved):
Nice. LGTM! Thanks!

PR #1857, opened 2021-02-12
Bring TimerTrack::UpdatePrimitives back to speed again.

florian-kuebler:
The recent fix for the iteration over timers introduced up to 30%
overhead. The time was spent in passing the large number of
arguments.

This change wraps the constant arguments into a struct and then
simply pass this as reference to the Draw function.

Test: Load a capture with timers and introspect.

### DIFF
--- a/src/OrbitGl/TimerTrack.cpp
+++ b/src/OrbitGl/TimerTrack.cpp
@@ -83,17 +83,15 @@ WorldXInfo ToWorldX(double start_us, double end_us, double inv_time_window, floa
 }  // namespace
 
 void TimerTrack::DrawTimer(TextBox* prev_text_box, TextBox* current_text_box,
-                           TextBox* next_text_box, uint64_t min_tick, uint64_t max_tick,
-                           float z_offset, Batcher* batcher, GlCanvas* canvas, float world_start_x,
-                           float world_width, double inv_time_window, bool is_collapsed, float z,
-                           const TextBox* selected_textbox, uint64_t highlighted_function_id,
-                           uint64_t pixel_delta_in_ticks, uint64_t min_timegraph_tick,
+                           TextBox* next_text_box, const internal::DrawData& draw_data,
                            uint64_t* min_ignore, uint64_t* max_ignore) {
   CHECK(min_ignore != nullptr);
   CHECK(max_ignore != nullptr);
   if (current_text_box == nullptr) return;
   const TimerInfo& current_timer_info = current_text_box->GetTimerInfo();
-  if (min_tick > current_timer_info.end() || max_tick < current_timer_info.start()) return;
+  if (draw_data.min_tick > current_timer_info.end() ||
+      draw_data.max_tick < current_timer_info.start())
+    return;
   if (current_timer_info.start() >= *min_ignore && current_timer_info.end() <= *max_ignore) return;
   if (!TimerFilter(current_timer_info)) return;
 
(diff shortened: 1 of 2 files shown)


ronaldfw:
Can you add some timing data and what benchmarks you did? (Size of capture at least.)

src/OrbitGl/TimerTrack.cpp
```diff
@@ -83,17 +83,15 @@ WorldXInfo ToWorldX(double start_us, double end_us, double inv_time_window, floa
 }  // namespace
 
 void TimerTrack::DrawTimer(TextBox* prev_text_box, TextBox* current_text_box,
-                           TextBox* next_text_box, uint64_t min_tick, uint64_t max_tick,
-                           float z_offset, Batcher* batcher, GlCanvas* canvas, float world_start_x,
-                           float world_width, double inv_time_window, bool is_collapsed, float z,
-                           const TextBox* selected_textbox, uint64_t highlighted_function_id,
-                           uint64_t pixel_delta_in_ticks, uint64_t min_timegraph_tick,
+                           TextBox* next_text_box, const internal::DrawData& draw_data,
```
ronaldfw, line 86: Nit: Put function arguments that are input only first.

https://google.github.io/styleguide/cppguide.html#Inputs_and_Outputs

src/OrbitGl/TimerTrack.h
```diff
@@ -29,6 +29,25 @@
 class OrbitApp;
 class TextRenderer;
 
+namespace internal {
+struct DrawData {
```
ronaldfw, line 33: Would it make sense to group data that is a bit related here? The order here seems random. Also, if you order by size, the size of the struct may be smaller, as you can avoid padding (likely with the single bool in the middle).

ronaldfw (approved):
LGTM! Please add some data on benchmarking.

src/OrbitGl/TimerTrack.cpp
```diff
@@ -146,55 +144,60 @@ void TimerTrack::DrawTimer(TextBox* prev_text_box, TextBox* current_text_box,
...
     double right_overlap_width_us = end_us - end_or_next_start_us;
     double text_x_end_us = end_or_next_start_us + (.25 * right_overlap_width_us);
 
-    bool is_visible_width =
-        (text_x_end_us - text_x_start_us) * inv_time_window * canvas->GetWidth() > 1;
-    WorldXInfo world_x_info =
-        ToWorldX(text_x_start_us, text_x_end_us, inv_time_window, world_start_x, world_width);
+    bool is_visible_width = (text_x_end_us - text_x_start_us) * draw_data.inv_time_window *
+                                draw_data.canvas->GetWidth() >
+                            1;
```
karupayun, line 156: Nit: Stupid clang-format thing, could I ask you to put the 1 in the previous line

karupayun (approved):
LGTM Florian! I completely agree with this change but as I told you, I'm still having the sensation that we should have a better way to order this part of the code. I will try to think about. :)

Ps. +1 for the benchmark.

pierricgimmig (approved):
Cool, thanks :-)

src/OrbitGl/TimerTrack.cpp
florian-kuebler: Which arguments do you mean? the text_boxes are input only. The only output arguments are the last two, or am I missing something?

florian-kuebler:
> LGTM! Please add some data on benchmarking.

Sure, I loaded Anton's "long" capture (go/orbit-long-capture), kept it on the default zoom level.
UpdatePrimitive calls on the current main version were around 1.8ms and with this version around 1.4ms.

I will rearrange the struct, but as it will be created only once per call to UpdatePrimitives and passed as ref. only to the Draw method, I don't expect any performance differnce.

src/OrbitGl/TimerTrack.cpp
florian-kuebler: Done. Needed parentheses around the calculation.

src/OrbitGl/TimerTrack.h
florian-kuebler: Done.

src/OrbitGl/TimerTrack.cpp
ronaldfw: If I see it right, there are calls to SetPos(...) and SetSize(...) at least on current_text_box (which I guess makes it an input/output parameter). But it's hard to read what is input / output here due to many pointers being used (the standard Google code style would be: Input type should be const& or by value, input/output or just output would be *).

florian-kuebler: Thanks, missed that part for the "current". I moved it to the other output parameters and made the other text boxes "const".

ronaldfw (approved):
LGTM

PR #1858, opened 2021-02-12
Make ReadFileToString exception-free

dimitry-:
It is also now linux-only since we need it only on linux at
this point.

Test: run OrbitBaseTests (on linux)


beckerhe (approved):
No comments. LGTM. Apparently clang-format prefers some changes though

PR #1859, opened 2021-02-12
Fix module reloading behaviour in OrbitApp

beckerhe:
I accidentally changed the behaviour of module reloading when doing the
preset refactoring, by making `LoadModule` only load a module when it's
not yet loaded.

But the legacy overload of `LoadModules` (which is calling the new
`LoadModule`) expects the module to be reloaded when it is already loaded.

So this PR makes it obvious what's happening by introducing a bunch of
overloads with descriptive names.

**Note:** I recommend to look at the header first, to quickly understand what's going on.

### DIFF
--- a/src/OrbitGl/App.cpp
+++ b/src/OrbitGl/App.cpp
@@ -1099,7 +1099,8 @@ void OrbitApp::LoadModules(
           SelectFunctionsFromHashes(module, select_functions);
           EnableFrameTracksFromHashes(module, frame_tracks);
         };
-    LoadModule(module).Then(main_thread_executor_, std::move(handle_hooks_and_frame_tracks));
+    LoadOrReloadModule(module).Then(main_thread_executor_,
+                                    std::move(handle_hooks_and_frame_tracks));
   }
 }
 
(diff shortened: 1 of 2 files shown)


src/OrbitGl/App.cpp
```diff
@@ -1130,10 +1131,47 @@ orbit_base::Future<ErrorMessageOr<void>> OrbitApp::LoadModule(const std::string&
   const ModuleData* module_data = GetModuleByPath(module_path);
 
   if (module_data == nullptr) {
-    return {ErrorMessage{absl::StrFormat("Module \"%s\" was not found", module_path)}};
+    return ErrorMessage{absl::StrFormat("Module \"%s\" was not found", module_path)};
+  }
+
+  if (module_data->is_loaded()) {
```
beckerhe, line 1137: Is checking `is_loaded` actually the correct thing?
Or do I have to call `GetMutableTargetProcess()->IsModuleLoaded(module_data->file_path)`?

beckerhe: Never mind. `is_loaded` is correct.

antonrohr (approved):
lgtm , thank you!

beckerhe:
I came to the conclusion that this is not the best way forward to fix the problem. Closing this for now. Anyway, Anton, thanks a lot for the review.

PR #1861, opened 2021-02-12
ProcessList::Refresh generates less exceptions

dimitry-:
This change uses noexcept functions for iterating a directory

Test: run ServiceUnitTests

### DIFF
--- a/src/Service/ProcessList.cpp
+++ b/src/Service/ProcessList.cpp
@@ -22,17 +22,35 @@ ErrorMessageOr<void> ProcessList::Refresh() {
 
   // TODO(b/161423785): This for loop should be refactored. For example, when
   //  parts are in a separate function, OUTCOME_TRY could be used to simplify
-  //  error handling. Also use ErrorMessageOr
-  for (const auto& directory_entry : std::filesystem::directory_iterator("/proc")) {
-    if (!directory_entry.is_directory()) continue;
+  //  error handling.
+  std::error_code error;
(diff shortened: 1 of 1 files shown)


src/Service/ProcessList.cpp
```diff
@@ -23,16 +23,34 @@ ErrorMessageOr<void> ProcessList::Refresh() {
   // TODO(b/161423785): This for loop should be refactored. For example, when
   //  parts are in a separate function, OUTCOME_TRY could be used to simplify
   //  error handling. Also use ErrorMessageOr
```
dpallotti, line 25: This was already here, but: Isn't `ErrorMessageOr` already used?

dpallotti (approved):

beckerhe (approved):
LGTM!

src/Service/ProcessList.cpp
```diff
@@ -22,17 +22,35 @@ ErrorMessageOr<void> ProcessList::Refresh() {
...
+    if (error) {
+      return ErrorMessage(
+          absl::StrFormat("Unable to iterate /proc directory: %s", error.message()));
+    }
+
+    bool is_directory = it->is_directory(error);
+    if (error) {
+      ERROR("Unable to stat \"%s\" directory entry: %s", it->path(), error.message());
+      continue;
+    }
```
dimitry-, line 42: actually this whole code makes me think using opendir is better and more readable.

PR #1863, opened 2021-02-15
Erase TimeGraph->GetLayout/FontSize dependences

karupayun:
Part of http://b/176086740.

We are erasing TimeGraph->GetLayout() dependence by adding the layout in
Tracks' constructors. Also TimeGraph->CalculateZoomedFontSize moving that
function into TimeGraphLayout.

Discussed in go/stadia-orbit-technical-debt.

Test: Loaded a capture.

### DIFF
--- a/src/OrbitGl/CaptureViewElement.h
+++ b/src/OrbitGl/CaptureViewElement.h
@@ -7,6 +7,7 @@
 
 #include "Batcher.h"
 #include "PickingManager.h"
+#include "TimeGraphLayout.h"
 
 class TimeGraph;
 class GlCanvas;
@@ -16,7 +17,7 @@ namespace orbit_gl {
 /* Base class for UI elements drawn underneath the capture window. */
 class CaptureViewElement : public Pickable {
  public:
-  explicit CaptureViewElement(TimeGraph* time_graph);
+  explicit CaptureViewElement(TimeGraph* time_graph, TimeGraphLayout* layout);
   virtual void Draw(GlCanvas* canvas, PickingMode /*picking_mode*/, float /*z_offset*/ = 0) {
     canvas_ = canvas;
   }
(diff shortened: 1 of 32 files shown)


src/OrbitGl/CaptureViewElement.h
```diff
@@ -7,6 +7,7 @@
 
 #include "Batcher.h"
 #include "PickingManager.h"
+#include "TimeGraphLayout.h"
```
florian-kuebler, line 10: Was that missing?

florian-kuebler (approved):
LGTM. Thanks!

src/OrbitGl/ThreadBar.h
```diff
@@ -30,6 +32,7 @@ class ThreadBar : public CaptureViewElement, public std::enable_shared_from_this
   void SetCaptureData(CaptureData* capture_data) { capture_data_ = capture_data; }
 
  protected:
+  TimeGraphLayout* layout_;
```
reichlfl, line 35: You could move this to `CaptureViewElement`, I think it's fundamental enough to be in the base class. And think about adding a CHECK for non-null during construction since most Draw / UpdatePrimitives methods rely on this being non-null.

reichlfl (approved):
Nice, thanks a lot. I have a suggestion about where to define the layout, but LGTM also if you choose to keep it like this.

src/OrbitGl/CaptureViewElement.h
karupayun: My bad, forgot to erase it. As first I saved it in CaptureViewElement (as @reichlfl is suggesting in the next comment)

src/OrbitGl/ThreadBar.h
karupayun: I was 50-50, not sure sure how much fundamental it was... Still not sure about that, but will follow your advice and changing it :)

PR #1866, opened 2021-02-15
Make ServiceUtils function noexcept

dimitry-:
Test: run ServiceTests

### DIFF
--- a/src/Service/ServiceUtils.cpp
+++ b/src/Service/ServiceUtils.cpp
@@ -28,6 +28,7 @@
 
 #include "ElfUtils/ElfFile.h"
 #include "OrbitBase/Logging.h"
+#include "OrbitBase/ReadFileToString.h"
 #include "OrbitBase/Result.h"
 #include "absl/strings/str_format.h"
 #include "module.pb.h"
@@ -75,7 +76,7 @@ ErrorMessageOr<std::vector<orbit_grpc_protos::TracepointInfo>> ReadTracepoints()
   return result;
 }
 
-std::optional<Jiffies> GetCumulativeCpuTimeFromProcess(pid_t pid) {
+std::optional<Jiffies> GetCumulativeCpuTimeFromProcess(pid_t pid) noexcept {
   const auto stat = std::filesystem::path{"/proc"} / std::to_string(pid) / "stat";
 
   // /proc/[pid]/stat looks like so (example - all in one line):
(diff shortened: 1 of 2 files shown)


src/Service/ServiceUtils.cpp
```diff
@@ -88,18 +89,26 @@ std::optional<Jiffies> GetCumulativeCpuTimeFromProcess(pid_t pid) {
   // Older kernels might have less fields than in the example. Over time fields had been added to
   // the end, but field indexes stayed stable.
 
-  if (!std::filesystem::exists(stat)) {
-    return {};
+  std::error_code error;
+  bool file_exists = std::filesystem::exists(stat, error);
+  // Even if we couldn't stat we could still be able to read, continue in case of a error.
```
dpallotti, line 94:
```suggestion
  // Even if we couldn't stat we could still be able to read, continue in case of an error.
```

src/Service/ServiceUtils.cpp
```diff
@@ -88,18 +89,26 @@ std::optional<Jiffies> GetCumulativeCpuTimeFromProcess(pid_t pid) {
...
-    return {};
+  ErrorMessageOr<std::string> file_content = orbit_base::ReadFileToString(stat);
+  if (!file_content) {
+    LOG("Could not read \"%s\": %s", stat.string(), file_content.error().message());
+    return std::nullopt;
+  }
+
+  std::vector<std::string> lines = absl::StrSplit(file_content.value(), absl::MaxSplits('\n', 1));
+  if (lines.empty()) {
+    LOG("\"%s\" file is empty", stat.string());
```
dpallotti, line 107: Isn't this more of an `ERROR`?

src/Service/ServiceUtils.cpp
```diff
@@ -88,18 +89,26 @@ std::optional<Jiffies> GetCumulativeCpuTimeFromProcess(pid_t pid) {
...
+    return std::nullopt;
   }
 
-  std::ifstream stream{stat.string()};
-  if (!stream.good()) {
-    LOG("Could not open %s", stat.string());
-    return {};
+  ErrorMessageOr<std::string> file_content = orbit_base::ReadFileToString(stat);
+  if (!file_content) {
+    LOG("Could not read \"%s\": %s", stat.string(), file_content.error().message());
```
dpallotti, line 101: Why not `ERROR`?

src/Service/ServiceUtils.cpp
```diff
@@ -135,8 +144,17 @@ std::optional<Jiffies> GetCumulativeCpuTimeFromProcess(pid_t pid) {
   return Jiffies{utime + stime};
 }
 
-std::optional<TotalCpuTime> GetCumulativeTotalCpuTime() {
-  std::ifstream stat_stream{"/proc/stat"};
+std::optional<TotalCpuTime> GetCumulativeTotalCpuTime() noexcept {
+  ErrorMessageOr<std::string> stat_content = orbit_base::ReadFileToString("/proc/stat");
+  if (!stat_content) {
+    ERROR("%s", stat_content.error().message());
+    return {};
```
dpallotti, line 151:
```suggestion
    return std::nullopt;
```
For consistency with `GetCumulativeCpuTimeFromProcess` and the next if.

src/Service/ServiceUtils.cpp
```diff
@@ -230,7 +247,17 @@ ErrorMessageOr<Path> FindSymbolsFilePath(const Path& module_path,
   std::vector<std::string> error_messages;
 
   for (const auto& symbols_path : search_paths) {
-    if (!std::filesystem::exists(symbols_path)) continue;
+    std::error_code error;
+    bool path_exists = std::filesystem::exists(symbols_path, error);
+    if (error) {
+      std::string error_message =
+          absl::StrFormat("Unable to stat \"%s\": %s", symbols_path, error.message());
+      LOG("%s", error_message);
```
dpallotti, line 255: Should be `ERROR`.

src/Service/ServiceUtils.h
```diff
@@ -23,7 +23,6 @@
 #include "tracepoint.pb.h"
 
 namespace orbit_service::utils {
-using Path = std::filesystem::path;
```
dpallotti, line 26: Thanks!

dpallotti (commented):
So the idea is to stop using `std::istream` and `std::ostream` because they throw exceptions?

beckerhe (approved):
Nothing to add to Davide's points.

src/Service/ServiceUtils.cpp
dimitry-: done.

src/Service/ServiceUtils.cpp
dimitry-: done.

src/Service/ServiceUtils.cpp
dimitry-: done.

src/Service/ServiceUtils.cpp
dimitry-: done. also changed in other places.

src/Service/ServiceUtils.cpp
dimitry-: done.

dimitry- (commented):
Yep, I am still thinking about directory iterator. I think we should probably provide a noexcept list_directory function in OrbitBase and stop using it directly. (but these are for another PRs)

dpallotti (approved):

PR #1867, opened 2021-02-15
Make `capture_data_` const* in all Tracks and remove SetCaptureData.

florian-kuebler:
This makes the `capture_data_` member const* in all Track(like)
classes. Further it removes some unused SetCaptureData methods,
that are not needed anymore, as TimeGraph/TrackManager will be
fresh for every capture, thus for every capture data.

Bug: http://b/180313647
Test: Compile


karupayun (approved):
LGTM! :)

PR #1868, opened 2021-02-15
E2E test assuring that we can instrument libggp.

danielfenner:


contrib/automation_tests/test_cases/live_tab.py
```diff
@@ -87,3 +87,20 @@ def _execute(self, function_name, min_calls=0, max_calls=pow(2, 33)-1):
         self.expect_true(min_calls <= call_count <= max_calls,
                          'Call count for function "%s" expected to be between %s and %s, was %s'
                          % (function_name, min_calls, max_calls, call_count))
+                         
+class VerifyOneFunctionWasCalled(LiveTabTestCase):
+    """
+    Verify that at least one of the functions matching the function name has 
+    received the given number of hits.
+    """
+    def _execute(self, function_name, min_calls=0, max_calls=pow(2, 33)-1):
```
dpallotti, line 96: Isn't `pow(2, 33)` a strange power of two?

contrib/automation_tests/test_cases/live_tab.py
```diff
@@ -87,3 +87,20 @@ def _execute(self, function_name, min_calls=0, max_calls=pow(2, 33)-1):
...
+        children = self.find_control('Tree', parent=self._live_tab).children()
+        for i in range(len(children)):
+            if function_name in children[i].window_text():
+                call_count = int(children[i + 1].window_text())
+                if min_calls <= call_count <= max_calls:
+                    logging.info('Found a call to %s with %s hits',
```
dpallotti, line 102: I'd suggest quotes for the first `%s`.

dpallotti, line 102: Shouldn't the second placeholder be `%d` or something? Python is happy with this?

contrib/automation_tests/test_cases/live_tab.py
```diff
@@ -87,3 +87,20 @@ def _execute(self, function_name, min_calls=0, max_calls=pow(2, 33)-1):
...
+    """
+    Verify that at least one of the functions matching the function name has 
+    received the given number of hits.
+    """
+    def _execute(self, function_name, min_calls=0, max_calls=pow(2, 33)-1):
+        children = self.find_control('Tree', parent=self._live_tab).children()
+        for i in range(len(children)):
+            if function_name in children[i].window_text():
+                call_count = int(children[i + 1].window_text())
+                if min_calls <= call_count <= max_calls:
```
dpallotti, line 101: But isn't `min_calls` 0 here?

dpallotti (commented):
Isn't `min_calls` 0 or what am I missing?

contrib/automation_tests/test_cases/live_tab.py
danielfenner: I just copied it from the code above. The integers look 64 bit wide:
```
>>> sys.maxsize
9223372036854775807
```
So this is just an arbitrary, large number I think.

Florian, can you comment on this?

contrib/automation_tests/test_cases/live_tab.py
danielfenner: Done

danielfenner: Works fine. Probably there is an automatic cast.

contrib/automation_tests/test_cases/live_tab.py
danielfenner: 0 is the default value for the parameter. Can be anything.

dpallotti: Of course, but you `min_calls` is not passed to `VerifyOneFunctionWasCalled` in `orbit_instrument_libs_.py`, so in that test the value of `min_calls` is 0 and you're not actually verifying that the function was called at least once.

contrib/automation_tests/test_cases/live_tab.py
reichlfl: Int in Python 3 is unbounded, so I think the value above is a rather arbitrary maximum default for this... `sys.maxsize` gives you the interpreters word-size and the maximum amount of e.g. elements in a list, but that doesn't affect ints at all.

reichlfl, line 96: I tried to distinguish the parameter names between exact and partial matches in other tests - e.g. `function_name` vs `function_name_contains`. This would make the difference to `VerifyFunctionCallCount` clearer.

contrib/automation_tests/test_cases/live_tab.py
reichlfl: I did the same for `VerifyFunctionCallCount` above, but I'd agree with Davide that the default is a bit strange here. I think it comes down to the name of the test... in `VerifyFunctionCallCount`, verifying a count of 0 sounds like a reasonable usecase. In `VerifyOneFunctionWasCalled`, I'd expect a minumum of 1.

But for consistency, do you maybe want to change the default `min_calls` to 1 in both tests?

reichlfl (approved):
I agree with Davide on the min_count and would suggest to change it for my old test as well. Apart from that LGTM, thanks!

contrib/automation_tests/test_cases/live_tab.py
danielfenner: Ah, okay, yes. That's fine but I can also put the ususal range in there for consistency. done.

danielfenner: Now I finally got it. Yes, zero is not a great value for the check. I'll make it 1.

contrib/automation_tests/test_cases/live_tab.py
danielfenner: Okay, I'll just leave it as is.

danielfenner: Done - renamed parameter to `function_name_contains`.

PR #1871, opened 2021-02-16
Fix dereference of nullptr.

danielfenner:
Bug: http://b/180372294
Test: Ran e2e tests locally. Works.


IrinaShkviro (approved):
From my perspective looks good and crash server is empty, thanks!

reichlfl (approved):
Right, makes sense that this crashes when the TimeGraph is optional... LGTM, thanks!

florian-kuebler (approved):
LGTM, thanks a lot!

PR #1872, opened 2021-02-16
Same warning messages on close

antonrohr:
2 commits. 

1. fixes a bug, where ConnectToStadiaWidget error messages have been displayed, even though the widget was not active.
2. Shows the same warnings when clicking "End Session" and "X"

### DIFF
--- a/src/OrbitQt/orbitmainwindow.cpp
+++ b/src/OrbitQt/orbitmainwindow.cpp
@@ -450,18 +450,8 @@ void OrbitMainWindow::SetupTargetLabel() {
 
   ui->menuBar->setCornerWidget(target_widget, Qt::TopRightCorner);
 
-  QObject::connect(disconnect_target_button, &QPushButton::clicked, this, [this]() {
-    QMessageBox::StandardButton reply = QMessageBox::question(
-        this, QApplication::applicationName(),
-        "This discards any unsaved progress. Are you sure you want to continue?");
-
-    if (reply == QMessageBox::Yes) {
-      if (app_->IsCapturing()) {
-        app_->AbortCapture();
-      }
-      QApplication::exit(kEndSessionReturnCode);
-    }
-  });
+  QObject::connect(disconnect_target_button, &QPushButton::clicked, this,
+                   [this] { RequestExitWithReturnCode(kEndSessionReturnCode); });
 }
 
 void OrbitMainWindow::SetupAccessibleNamesForAutomation() {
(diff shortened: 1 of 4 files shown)


src/OrbitQt/orbitmainwindow.cpp
```diff
@@ -1140,26 +1130,48 @@ bool OrbitMainWindow::eventFilter(QObject* watched, QEvent* event) {
...
+             this, QApplication::applicationName(),
+             "This discards any unsaved progress. Are you sure you want to continue?") ==
+         QMessageBox::Yes;
+}
+
+bool OrbitMainWindow::ShowCaptureAbortWarningAndReturnResult() {
+  return QMessageBox::question(this, "Capture in progress",
+                               "A capture is currently in progress. Do you want to abort the "
+                               "capture and exit Orbit?") == QMessageBox::Yes;
+}
```
IrinaShkviro, line 1144: Optional suggestion: maybe name this functions `ConfirmClose` and `ConfirmAbortCapture` (I think it's a bit more clear which result will be returned and more readable).

src/OrbitQt/orbitmainwindow.h
```diff
@@ -163,6 +164,9 @@ class OrbitMainWindow : public QMainWindow {
 
   static const QString kCollectThreadStatesSettingKey;
   void LoadCaptureOptionsIntoApp();
+  bool ShowCloseWarningAndReturnResult();
+  bool ShowCaptureAbortWarningAndReturnResult();
+  bool ExitRequestWithReturnCode(int return_code);
```
beckerhe, line 169: It took me a while to understand the function name. I was always thinking: "What kind of request is exited here?".

Have you considered calling it `RequestExitWithReturnCode`. I think this is not ambiguous, but I might be wrong.

src/OrbitQt/orbitmainwindow.cpp
```diff
@@ -1140,26 +1130,48 @@ bool OrbitMainWindow::eventFilter(QObject* watched, QEvent* event) {
...
+  }
+
+  if (!ShowCloseWarningAndReturnResult()) return false;
+
+  if (main_thread_executor_) {
+    main_thread_executor_->AbortWaitingJobs();
+  }
+  if (introspection_widget_ != nullptr) {
+    introspection_widget_->close();
+  }
```
IrinaShkviro, line 1163: Both `main_thread_executor_` and `introspection_widget_ ` are smart pointers, but in one case there is implicit conversion, and in other - no. I think we should be consistent.

src/OrbitQt/orbitmainwindow.cpp
beckerhe, line 1144: Nit: Not sure if these two functions make it more readable to me. I tend to say no, since they contain a single statement.

beckerhe (approved):
Otherwise LGTM

IrinaShkviro (approved):
Looks good, couple of small comments

src/OrbitQt/orbitmainwindow.h
antonrohr: Good idea! I renamed it

src/OrbitQt/orbitmainwindow.cpp
antonrohr: I agree and inlined them

antonrohr: I inlined this function (see Hennings comment), hence the name does not exist anymore. Thank you for the suggestion anyways

src/OrbitQt/orbitmainwindow.cpp
antonrohr: Good catch. I compared to nullptr in both ifs.

beckerhe: I guess that's my fault. I introduced the first one. Anton just moved it to a different location.

src/OrbitQt/orbitmainwindow.h
antonrohr: Good catch. I compared to nullptr in both ifs.

vickyliu-go4it (approved):
LGTM! The same as Henning, I would prefer to name `ExitRequestWithReturnCode` as `RequestExitWithReturnCode`. Thanks for that!

PR #1873, opened 2021-02-16
Make ListFilesRecursivelyIgnoreErrors noexcept

dimitry-:
Also changed GetLogFileLifetime->ParseLogFileTimestamp it is easier to
test if it returns timestamp. Switched timestamps to UTC.

Move LoggingUtils methods to orbit_base_internal namespace

Test: run OrbitBaseTests

### DIFF
--- a/src/OrbitBase/Logging.cpp
+++ b/src/OrbitBase/Logging.cpp
@@ -21,16 +21,18 @@ static absl::Mutex log_file_mutex(absl::kConstInit);
 std::ofstream log_file;
 
 std::string GetLogFileName() {
-  std::string timestamp_string =
-      absl::FormatTime(kLogFileNameTimeFormat, absl::Now(), absl::LocalTimeZone());
-  return absl::StrFormat(kLogFileNameDelimiter, timestamp_string,
+  std::string timestamp_string = absl::FormatTime(orbit_base_internal::kLogFileNameTimeFormat,
+                                                  absl::Now(), absl::UTCTimeZone());
+  return absl::StrFormat(orbit_base_internal::kLogFileNameDelimiter, timestamp_string,
                          static_cast<uint32_t>(orbit_base::GetCurrentProcessId()));
 }
 
 ErrorMessageOr<void> TryRemoveOldLogFiles(const std::filesystem::path& log_dir) {
-  std::vector<std::filesystem::path> log_file_paths = ListFiles(log_dir);
-  std::vector<std::filesystem::path> old_files = FilterOldLogFiles(log_file_paths);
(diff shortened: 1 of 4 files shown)


src/OrbitBase/Logging.cpp
```diff
@@ -21,16 +21,18 @@ static absl::Mutex log_file_mutex(absl::kConstInit);
 std::ofstream log_file;
 
 std::string GetLogFileName() {
-  std::string timestamp_string =
-      absl::FormatTime(kLogFileNameTimeFormat, absl::Now(), absl::LocalTimeZone());
-  return absl::StrFormat(kLogFileNameDelimiter, timestamp_string,
+  std::string timestamp_string = absl::FormatTime(orbit_base_internal::kLogFileNameTimeFormat,
+                                                  absl::Now(), absl::LocalTimeZone());
```
dimitry-, line 25: oops. just a sec.

vickyliu-go4it (approved):
LGTM! Thanks for improving this. The code looks much better now : )

src/OrbitBase/LoggingUtils.cpp
```diff
@@ -11,68 +11,96 @@
...
+ErrorMessageOr<absl::Time> ParseLogFileTimestamp(const std::string& log_file_name) noexcept {
   if (log_file_name.size() < kTimestampStartPos + kTimestampStringLength) {
     return ErrorMessage(
         absl::StrFormat("Unable to extract time information from log file: %s", log_file_name));
   }
   std::string timestamp_string = log_file_name.substr(kTimestampStartPos, kTimestampStringLength);
   absl::Time log_file_timestamp;
   std::string parse_time_error;
-  if (!absl::ParseTime(kLogFileNameTimeFormat, timestamp_string, absl::LocalTimeZone(),
+  if (!absl::ParseTime(kLogFileNameTimeFormat, timestamp_string, absl::UTCTimeZone(),
```
antonrohr, line 64: I am unsure about UTCTimeZone. I would agree if we used a iso timestamp like `2021-02-16T16:24:47Z`. The Z indicates about UTC. Maybe we should change the format, or stick to local time.. wdyt?

dimitry-: I went with UTC because it does not require knowledge of local timezones, daylight saving times and it is universally transferable. You can reason about the time the file was created without asking the person who sent it "what is your timezone?"

antonrohr: okay. I guess its better, even without the Z

antonrohr (approved):

PR #1878, opened 2021-02-16
Refactor OrbitMainWindow

antonrohr:
This makes the constructor a little simpler.

Ongoing effort in removing parts of the old UI b/177304549

### DIFF
--- a/src/OrbitGl/App.cpp
+++ b/src/OrbitGl/App.cpp
@@ -270,7 +270,7 @@ void OrbitApp::OnTimer(const TimerInfo& timer_info) {
 }
 
 void OrbitApp::OnKeyAndString(uint64_t key, std::string str) {
-  string_manager_->AddIfNotPresent(key, std::move(str));
+  string_manager_.AddIfNotPresent(key, std::move(str));
 }
 
 void OrbitApp::OnUniqueCallStack(CallStack callstack) {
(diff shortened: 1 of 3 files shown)


vickyliu-go4it (approved):
LGTM. This makes the code much simpler. Thanks, Anton!

src/OrbitGl/App.cpp
```diff
@@ -138,7 +138,8 @@ bool DoZoom = false;
 OrbitApp::OrbitApp(MainThreadExecutor* main_thread_executor,
                    const orbit_base::CrashHandler* crash_handler,
                    orbit_metrics_uploader::MetricsUploader* metrics_uploader)
-    : main_thread_executor_(main_thread_executor),
+    : string_manager_(std::make_shared<StringManager>()),
```
beckerhe, line 141: I see no reason, why StringManager is a shared_ptr in the first place. Could probably be just a value, not even a unique_ptr is needed.

(Sorry, unsolicited comment. Feel free to ignore.)

IrinaShkviro (approved):
LGTM!

src/OrbitGl/App.cpp
antonrohr: No, good that you looked at it. I changed it to simple `StringManager string_manager_`. 
Since this is a fairly small PR, I amended, force pushed and will merge soon

beckerhe: Thanks a lot for including this!

beckerhe (approved):

PR #1879, opened 2021-02-16
Fix warning messages

antonrohr:
My recent change to align the warning messages on quit / end session, did introduce some unwanted behaviour. Before this change, cancelling the loading of a capture would lead to a "this discards any unsaved changes" warning. Same for when the capture loading failed. This PR fixes that.

### DIFF
--- a/src/OrbitQt/orbitmainwindow.cpp
+++ b/src/OrbitQt/orbitmainwindow.cpp
@@ -414,7 +414,7 @@ void OrbitMainWindow::SetupTargetLabel() {
   ui->menuBar->setCornerWidget(target_widget, Qt::TopRightCorner);
 
   QObject::connect(disconnect_target_button, &QPushButton::clicked, this,
-                   [this] { RequestExitWithReturnCode(kEndSessionReturnCode); });
+                   [this] { on_actionEnd_Session_triggered(); });
 }
 
 void OrbitMainWindow::SetupAccessibleNamesForAutomation() {
@@ -814,13 +814,11 @@ void OrbitMainWindow::on_actionOpen_Preset_triggered() {
 }
 
 void OrbitMainWindow::on_actionEnd_Session_triggered() {
-  close();
-  QApplication::exit(kEndSessionReturnCode);
+  if (ConfirmExit()) Exit(kEndSessionReturnCode);
 }
 
 void OrbitMainWindow::on_actionQuit_triggered() {
-  close();
-  QApplication::quit();
+  if (ConfirmExit()) Exit(kQuitOrbitReturnCode);
 }
 
 void OrbitMainWindow::on_actionSave_Preset_As_triggered() {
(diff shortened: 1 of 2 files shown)


src/OrbitQt/orbitmainwindow.cpp
```diff
@@ -1053,13 +1049,13 @@ void OrbitMainWindow::OpenCapture(const std::string& filepath) {
         if (!result.has_value()) {
           QMessageBox::critical(this, "Error while loading capture",
                                 QString::fromStdString(result.error().message()));
-          on_actionEnd_Session_triggered();
+          QApplication::exit(kEndSessionReturnCode);
```
vickyliu-go4it, line 1052: Just a personal preference. Could we add a return code like `kEndSessionWithoutWarningReturnCode` for this case? If so, this exit event could be handled in `RequestExitWithReturnCode` with the new return code. WDYT?

vickyliu-go4it (approved):
Looks good. Thanks for fixing this : )

src/OrbitQt/orbitmainwindow.cpp
beckerhe: Vicky has a good point here. Especially since work like `main_thread_executor->AbortWaitingJobs` should be done. Not strictly necessary, but it's probably good to go through `RequestExitWithReturnCode` in all cases.

But I wouldn't use the return code to encode such specific behaviour, since that's an implementation detail of `OrbitMainWindow` that shouldn't leak into other parts of the code. But a second parameter to `RequestExitWithReturnCode` should do the trick.

beckerhe (approved):
Otherwise LGTM

src/OrbitQt/orbitmainwindow.cpp
antonrohr: Yes, I agree, the important part is doing the actual clean up like `main_thread_executor->AbortWaitingJobs`. I added another commit, it makes 2 functions out of `RequestExitWithReturnCode`. `ExitWithWarning` and `Exit`. The first one calls the second one, when the warnings are dismissed. The second one cleans up. The second one is called when no warning is needed (When clicking "Cancel" on loading)

ptal

src/OrbitQt/orbitmainwindow.cpp
```diff
@@ -1130,25 +1124,32 @@ bool OrbitMainWindow::eventFilter(QObject* watched, QEvent* event) {
   return QObject::eventFilter(watched, event);
 }
 
-bool OrbitMainWindow::RequestExitWithReturnCode(int return_code) {
+bool OrbitMainWindow::ExitWithWarning(int return_code) {
```
beckerhe, line 1127: Nit: From an API design perspective you should consider splitting up getting confirmation and exiting. (separation of concerns)

```cpp
bool OrbitMainWindow::ConfirmExit() const;
void OrbitMainwindow::Exit(int return_code);
```

antonrohr: So `ConfirmExit()` just displays the warning? And then I call it something like this:
```
if (ConfirmExit()) Exit(0);
```
Did I get that right?

beckerhe: Yes, exactly.

antonrohr: I had to use `const_cast` in `ConfirmExit()` since `QMessageBox::warning()` is non const. I think this is a good use of `const_cast`, but I would like to know your opinion. The alternative would be to not make `ConfirmExit()` const.

beckerhe: Ah I'm sorry for suggesting `const`. I would prefer to make it non-const.

You may be right that `const_cast` is okay here, but it sets a bad example, since there is really nothing that blocks us from making the method non-const.

antonrohr: No worries, suggesting const made sense in the first place. And with this little detour I learned about `const_cast` and how to use it :). Removed it, will merge now.

PR #1880, opened 2021-02-16
Fix kokoro cleanup for presubmit coverage

antonrohr:
The presubmit code coverage currently does not produce any results. I think this is because the results are "cleaned up" before kokoro can upload them to the bucket. This commit changes that by having an extra case in the cleanup part, for the `coverage_clang9` profile.


beckerhe (approved):
LGTM

danielfenner (approved):
lgtm

PR #1881, opened 2021-02-16
Introduce absl::Span in ModuleManager and ProcessData

beckerhe:
This simply replaces some `const std::vector<T>&` arguments by
`absl::Span<const T>` arguments. They implementation or caller changes
needed.

I will introduce the first real users of that in a subsequent commit.


antonrohr (approved):
I like these kinda changes!

PR #1882, opened 2021-02-16
Change parameter of SetCurrentThreadName to string_view

dimitry-:
Also slightly optimize Linux implementation

Test: run OrbitBaseTests

### DIFF
--- a/src/OrbitBase/ThreadUtilsLinux.cpp
+++ b/src/OrbitBase/ThreadUtilsLinux.cpp
@@ -40,12 +40,14 @@ std::string GetThreadName(pid_t tid) {
   return comm_content.value();
 }
 
-void SetCurrentThreadName(const std::string& thread_name) {
-  std::string name = thread_name;
-  if (name.length() >= kMaxThreadNameLength) {
-    name[kMaxThreadNameLength - 1] = '\0';
+void SetCurrentThreadName(const char* thread_name) {
+  std::string buf;
+  if (strlen(thread_name) >= kMaxThreadNameLength) {
+    buf.assign(thread_name, kMaxThreadNameLength - 1);
+    thread_name = buf.c_str();
   }
-  int result = pthread_setname_np(pthread_self(), name.data());
+
+  int result = pthread_setname_np(pthread_self(), thread_name);
   if (result != 0) {
     ERROR("Setting thread name for tid %d. Error %d", GetCurrentThreadId(), result);
   }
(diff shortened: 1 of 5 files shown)


src/OrbitBase/ThreadUtilsLinux.cpp
```diff
@@ -40,12 +40,15 @@ std::string GetThreadName(pid_t tid) {
   return comm_content.value();
 }
 
-void SetCurrentThreadName(const std::string& thread_name) {
-  std::string name = thread_name;
-  if (name.length() >= kMaxThreadNameLength) {
-    name[kMaxThreadNameLength - 1] = '\0';
+void SetCurrentThreadName(std::string_view thread_name) {
+  std::string buf;
+  const char* thread_name_cstr = thread_name.data();
```
pierricgimmig, line 45: thread_name_cstr won't be null terminated which is expected by pthread_setname_np expects

src/OrbitBase/ThreadUtilsLinux.cpp
```diff
@@ -40,12 +40,15 @@ std::string GetThreadName(pid_t tid) {
...
+  if (thread_name.length() >= kMaxThreadNameLength) {
+    buf = thread_name.substr(0, kMaxThreadNameLength - 1);
+    thread_name_cstr = buf.data();
```
pierricgimmig, line 48: same thing here.

pierricgimmig (commented):
Hi Dimitry, thanks for this, I think there are issues with the strings not being null terminated however.

src/OrbitBase/ThreadUtilsLinux.cpp
dimitry-: well, this part is fine because std::strings c_str() are null-terminated.

pierricgimmig: well c_str() yes, but not data() I think, no?

src/OrbitBase/ThreadUtilsLinux.cpp
dimitry-: right. fixed.

src/OrbitBase/ThreadUtilsLinux.cpp
pierricgimmig: actually it seems data does return a null terminated string https://en.cppreference.com/w/cpp/string/basic_string/data#:~:text=The%20returned%20array%20is%20null,to%20a%20single%20null%20character.&text=The%20pointer%20obtained%20from%20data,any%20standard%20library%20function%2C%20or

pierricgimmig (approved):
LGTM!

PR #1883, opened 2021-02-17
Add test for Future::Then and Future::ThenIfSuccess

beckerhe:
There was a typo in `Future::Then`'s implementation.

This PR fixes that and adds unit tests for Future::Then and Future::ThenIfSuccess.
These tests were not possible before because they require an executor. But since we
now have the `ImmediateExecutor` adding these missing tests was possible.


dpallotti (approved):
Nice!

dimitry- (approved):

PR #1884, opened 2021-02-17
Add more UnwrapFuture overloads

beckerhe:
This adds some more overloads to UnwrapFuture. These overloads handle
`ErrorMessageOr<Future<T>>` cases.

- `Future<ErrorMessageOr<Future<T>>>` unwraps to
  `Future<ErrorMessageOr<T>>`
- `Future<ErrorMessageOr<Future<ErrorMessageOr<T>>>>` unwraps to
  `Future<ErrorMessageOr<T>>`

Ideally these situations shouldn't occur in a good designed API, but
refactorings might make them necessary temporarily.

### DIFF
--- a/src/OrbitBase/include/OrbitBase/FutureHelpers.h
+++ b/src/OrbitBase/include/OrbitBase/FutureHelpers.h
@@ -73,6 +73,82 @@ template <typename T>
   return promise->GetFuture();
 }
 
+template <typename T>
+[[nodiscard]] Future<ErrorMessageOr<T>> UnwrapFuture(
+    const Future<ErrorMessageOr<Future<T>>>& outer_future) {
+  orbit_base::Promise<ErrorMessageOr<T>> promise{};
+  auto future = promise.GetFuture();
+
+  auto outer_continuation =
+      [promise = std::move(promise)](const ErrorMessageOr<Future<T>>& outer_result) mutable {
+        if (outer_result.has_error()) {
+          promise.SetResult(outcome::failure(outer_result.error()));
+          return;
+        }
+
+        auto inner_continuation = [promise = std::move(promise)](const T& arg) mutable {
+          promise.SetResult(arg);
+        };
+
+        const auto& inner_future = outer_result.value();
+        RegisterContinuationOrCallDirectly(inner_future, std::move(inner_continuation));
+      };
+  RegisterContinuationOrCallDirectly(outer_future, std::move(outer_continuation));
+
+  return future;
+}
+
+template <>
+[[nodiscard]] inline Future<ErrorMessageOr<void>> UnwrapFuture(
+    const Future<ErrorMessageOr<Future<void>>>& outer_future) {
+  orbit_base::Promise<ErrorMessageOr<void>> promise{};
+  auto future = promise.GetFuture();
+
+  auto outer_continuation =
+      [promise = std::move(promise)](const ErrorMessageOr<Future<void>>& outer_result) mutable {
+        if (outer_result.has_error()) {
+          promise.SetResult(outcome::failure(outer_result.error()));
+          return;
+        }
+
+        auto inner_continuation = [promise = std::move(promise)]() mutable {
+          promise.SetResult(outcome::success());
+        };
+
+        const auto& inner_future = outer_result.value();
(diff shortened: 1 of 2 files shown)


src/OrbitBase/include/OrbitBase/FutureHelpers.h
```diff
@@ -73,6 +73,82 @@ template <typename T>
...
+template <typename T>
+[[nodiscard]] inline Future<ErrorMessageOr<T>> UnwrapFuture(
+    const Future<ErrorMessageOr<Future<T>>>& outer_future) {
+  orbit_base::Promise<ErrorMessageOr<T>> promise{};
+  auto future = promise.GetFuture();
+
+  auto outer_continuation =
+      [promise = std::move(promise)](const ErrorMessageOr<Future<T>>& outer_result) mutable {
+        if (outer_result.has_error()) {
+          promise.SetResult(outcome::failure(outer_result.error()));
```
dimitry-, line 85: do you need outcome::failure here? I think you can drop it.

src/OrbitBase/include/OrbitBase/FutureHelpers.h
```diff
@@ -73,6 +73,82 @@ template <typename T>
...
+template <>
+[[nodiscard]] inline Future<ErrorMessageOr<void>> UnwrapFuture(
+    const Future<ErrorMessageOr<Future<void>>>& outer_future) {
+  orbit_base::Promise<ErrorMessageOr<void>> promise{};
+  auto future = promise.GetFuture();
+
+  auto outer_continuation =
+      [promise = std::move(promise)](const ErrorMessageOr<Future<void>>& outer_result) mutable {
+        if (outer_result.has_error()) {
+          promise.SetResult(outcome::failure(outer_result.error()));
```
dimitry-, line 110: same here.

src/OrbitBase/include/OrbitBase/FutureHelpers.h
```diff
@@ -73,6 +73,82 @@ template <typename T>
   return promise->GetFuture();
 }
 
+template <typename T>
+[[nodiscard]] inline Future<ErrorMessageOr<T>> UnwrapFuture(
```
dimitry-, line 77: does inline give us anything here?

src/OrbitBase/include/OrbitBase/FutureHelpers.h
```diff
@@ -73,6 +73,82 @@ template <typename T>
...
+    const Future<ErrorMessageOr<Future<ErrorMessageOr<T>>>>& outer_future) {
+  orbit_base::Promise<ErrorMessageOr<T>> promise{};
+  auto future = promise.GetFuture();
+
+  auto outer_continuation =
+      [promise = std::move(promise)](
+          const ErrorMessageOr<Future<ErrorMessageOr<T>>>& outer_result) mutable {
+        if (outer_result.has_error()) {
+          promise.SetResult(outcome::failure(outer_result.error()));
```
dimitry-, line 136: and here

src/OrbitBase/include/OrbitBase/FutureHelpers.h
beckerhe: Unfortunately not. The implicit constructor of the error case is only available if it's unambiguous.

So for example it won't be available for the type `ErrorMessageOr<ErrorMessage>` or when the `T` in `ErrorMessageOr<T>` has a templated constructor. So to make this universally work, we need outcome::failure.

I have another commit (not yet put out) which will add this to other parts of the code as well.

src/OrbitBase/include/OrbitBase/FutureHelpers.h
beckerhe: No, just a copy-and-paste artefact. I will remove it.

dimitry- (approved):

src/OrbitBase/include/OrbitBase/FutureHelpers.h
beckerhe: Done.

src/OrbitBase/include/OrbitBase/FutureHelpers.h
beckerhe: I'm going to resolve this.

PR #1886, opened 2021-02-17
Remove unused testing code.

danielfenner:


reichlfl (commented):
Was the `orbit_connect_and_capture.py` suite never run? I've extended it in my current branch to check the timers in Scheduler and GFX tracks, but I could as well move that to a different test.

danielfenner:
> Was the `orbit_connect_and_capture.py` suite never run? 
Not on the testing infrastucture. 
The test cases ConnectToStadiaInstance, FilterAndSelectFirstProcess, Capture are used in many other tests so this one really doesn't add anything.

> I've extended it in my current branch to check the timers in Scheduler and GFX tracks, but I could as well move that to a different test.

Then this should be named something with "timers" I think. Does it make anything easier if I keep it here for now?  If so I'll just remove this from the PR.

reichlfl:
> Then this should be named something with "timers" I think. Does it make anything easier if I keep it here for now? If so I'll just remove this from the PR.

Doesn't really matter, then I will create a new test for this.

reichlfl (approved):

PR #1887, opened 2021-02-17
Expose setting to limit the local depth of markers per command buffer.

florian-kuebler:
Adding a lot of Vulkan debug markers can have a significant performance
impact when capturing those.
On the layer, we already have an option to limit the depth of
markers (per command buffer). This change exposes this setting to the
capture settings dialog and transfer the specified value to the layer.

On the UI side, the setting is split into a boolean (checkbox) if
the depth should be limited at all and a text field (only enabled when
the checkbox is clicked and only allowing numeric strings in uint64 range).
Internally, we use uint64::max as to disable limitation of the depth
(as already done in the layer).

Note that the text fields also excepts empty strings (as otherwise
it is inconvenient to change the first digit), but resets those to
0.

Bug: http://b/180388662
Test: Try different settings in the capture options menu and take
 captures with Vulkan layer enabled.

Screenshot 2021-02-17 at 11 21 05
tooltip

### DIFF
--- a/src/OrbitQt/CaptureOptionsDialog.cpp
+++ b/src/OrbitQt/CaptureOptionsDialog.cpp
@@ -4,10 +4,10 @@
 
 #include "CaptureOptionsDialog.h"
 
-#include <QCheckBox>
 #include <QDialog>
 #include <QDialogButtonBox>
 #include <QWidget>
+#include <QtGui/QValidator>
 
 #include "ui_CaptureOptionsDialog.h"
 
@@ -19,6 +19,8 @@ CaptureOptionsDialog::CaptureOptionsDialog(QWidget* parent)
 
   QObject::connect(ui_->buttonBox, &QDialogButtonBox::accepted, this, &QDialog::accept);
   QObject::connect(ui_->buttonBox, &QDialogButtonBox::rejected, this, &QDialog::reject);
+
+  ui_->localMarkerDepthLineEdit->setValidator(&uint64_validator_);
 }
 
 bool CaptureOptionsDialog::GetCollectThreadStates() const {
@@ -28,5 +30,33 @@ bool CaptureOptionsDialog::GetCollectThreadStates() const {
 void CaptureOptionsDialog::SetCollectThreadStates(bool collect_thread_state) {
   ui_->threadStateCheckBox->setChecked(collect_thread_state);
 }
+
+void CaptureOptionsDialog::SetLimitLocalMarkerDepthPerCommandBuffer(
+    bool limit_local_marker_depth_per_command_buffer) {
+  ui_->localMarkerDepthCheckBox->setChecked(limit_local_marker_depth_per_command_buffer);
+}
+
+bool CaptureOptionsDialog::GetLimitLocalMarkerDepthPerCommandBuffer() const {
+  return ui_->localMarkerDepthCheckBox->isChecked();
+}
+
+void CaptureOptionsDialog::SetMaxLocalMarkerDepthPerCommandBuffer(
+    uint64_t local_marker_depth_per_command_buffer) {
+  ui_->localMarkerDepthLineEdit->setText(QString::number(local_marker_depth_per_command_buffer));
+}
+
+uint64_t CaptureOptionsDialog::GetMaxLocalMarkerDepthPerCommandBuffer() const {
+  CHECK(!ui_->localMarkerDepthLineEdit->text().isEmpty());
+  bool valid = false;
+  uint64_t result = ui_->localMarkerDepthLineEdit->text().toULongLong(&valid);
+  CHECK(valid);
+  return result;
+}
+
+void CaptureOptionsDialog::resetLocalMarkerDepthLineEdit() {
+  if (ui_->localMarkerDepthLineEdit->text().isEmpty()) {
+    ui_->localMarkerDepthLineEdit->setText(QString::number(0));
+  }
+};
 
 }  // namespace orbit_qt
--- a/src/OrbitQt/CaptureOptionsDialog.cpp
+++ b/src/OrbitQt/CaptureOptionsDialog.cpp
@@ -53,10 +53,10 @@ uint64_t CaptureOptionsDialog::GetMaxLocalMarkerDepthPerCommandBuffer() const {
   return result;
 }
 
-void CaptureOptionsDialog::resetLocalMarkerDepthLineEdit() {
+void CaptureOptionsDialog::ResetLocalMarkerDepthLineEdit() {
   if (ui_->localMarkerDepthLineEdit->text().isEmpty()) {
     ui_->localMarkerDepthLineEdit->setText(QString::number(0));
   }
-};
+}
 
 }  // namespace orbit_qt
--- a/src/OrbitQt/CaptureOptionsDialog.h
+++ b/src/OrbitQt/CaptureOptionsDialog.h
@@ -8,6 +8,7 @@
 #include <QDialog>
 #include <QObject>
 #include <QString>
+#include <QValidator>
 #include <QWidget>
 #include <memory>
 
@@ -15,6 +16,24 @@
 #include "ui_CaptureOptionsDialog.h"
 
 namespace orbit_qt {
+
+namespace internal {
+class UInt64Validator : public QValidator {
+ public:
+  explicit UInt64Validator(QObject* parent = nullptr) : QValidator(parent) {}
+  QValidator::State validate(QString& input, int& /*pos*/) const override {
+    if (input.isEmpty()) {
+      return QValidator::State::Acceptable;
+    }
+    bool valid = false;
+    input.toULongLong(&valid);
+    if (valid) {
+      return QValidator::State::Acceptable;
+    }
+    return QValidator::State::Invalid;
+  }
+};
+}  // namespace internal
 
 class CaptureOptionsDialog : public QDialog {
   Q_OBJECT
@@ -24,9 +43,17 @@ class CaptureOptionsDialog : public QDialog {
 
   void SetCollectThreadStates(bool collect_thread_state);
   [[nodiscard]] bool GetCollectThreadStates() const;
+  void SetLimitLocalMarkerDepthPerCommandBuffer(bool limit_local_marker_depth_per_command_buffer);
+  [[nodiscard]] bool GetLimitLocalMarkerDepthPerCommandBuffer() const;
+  void SetMaxLocalMarkerDepthPerCommandBuffer(uint64_t local_marker_depth_per_command_buffer);
+  [[nodiscard]] uint64_t GetMaxLocalMarkerDepthPerCommandBuffer() const;
+
+ public slots:
+  void resetLocalMarkerDepthLineEdit();
 
  private:
   std::unique_ptr<Ui::CaptureOptionsDialog> ui_;
+  internal::UInt64Validator uint64_validator_;
 };
 
 }  // namespace orbit_qt
--- a/src/OrbitQt/CaptureOptionsDialog.h
+++ b/src/OrbitQt/CaptureOptionsDialog.h
@@ -17,7 +17,6 @@
 
 namespace orbit_qt {
 
-namespace internal {
 class UInt64Validator : public QValidator {
  public:
   explicit UInt64Validator(QObject* parent = nullptr) : QValidator(parent) {}
@@ -33,7 +32,6 @@ class UInt64Validator : public QValidator {
     return QValidator::State::Invalid;
   }
 };
-}  // namespace internal
 
 class CaptureOptionsDialog : public QDialog {
   Q_OBJECT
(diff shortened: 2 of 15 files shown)


ronaldfw:
Do we know whether the layer is enabled when we show the settings? It would be nice to disable that part when the Vulkan layer is not enabled and have a tooltip that explains the layer needs to be enabled first.

florian-kuebler:
> Do we know whether the layer is enabled when we show the settings? It would be nice to disable that part when the Vulkan layer is not enabled and have a tooltip that explains the layer needs to be enabled first.

Unfortunately, not. Davide and I have discussed this already in a couple of other contexts, but came to the conclusion that we should not try to expose this information by the IPC.

src/OrbitQt/CaptureOptionsDialog.h
```diff
@@ -16,6 +17,24 @@
 
 namespace orbit_qt {
 
+namespace internal {
```
dpallotti, line 20: I thought we didn't want nested namespaces? Why do you need it? Is this header file supposed to be used by a different module?

src/OrbitQt/CaptureOptionsDialog.cpp
```diff
@@ -29,4 +31,32 @@ void CaptureOptionsDialog::SetCollectThreadStates(bool collect_thread_state) {
...
+
+uint64_t CaptureOptionsDialog::GetMaxLocalMarkerDepthPerCommandBuffer() const {
+  CHECK(!ui_->localMarkerDepthLineEdit->text().isEmpty());
+  bool valid = false;
+  uint64_t result = ui_->localMarkerDepthLineEdit->text().toULongLong(&valid);
+  CHECK(valid);
+  return result;
+}
+
+void CaptureOptionsDialog::resetLocalMarkerDepthLineEdit() {
```
dpallotti, line 56: Is "reset" lowercase on purpose?

src/OrbitQt/CaptureOptionsDialog.cpp
```diff
@@ -29,4 +31,32 @@ void CaptureOptionsDialog::SetCollectThreadStates(bool collect_thread_state) {
...
+  if (ui_->localMarkerDepthLineEdit->text().isEmpty()) {
+    ui_->localMarkerDepthLineEdit->setText(QString::number(0));
+  }
+};
```
dpallotti, line 60:
```suggestion
}
```

src/OrbitQt/CaptureOptionsDialog.ui
```diff
@@ -9,36 +9,206 @@
...
-    <height>200</height>
+    <width>440</width>
+    <height>260</height>
    </rect>
   </property>
   <property name="windowTitle">
    <string>Capture Options</string>
   </property>
+  <property name="toolTip">
+   <string>Limits the maximal depth of Vulkan debug markers per command buffer to the given value.
```
dpallotti, line 20: I would use "maximum".

src/OrbitQt/CaptureOptionsDialog.ui
```diff
@@ -9,36 +9,206 @@
...
+     </property>
+     <property name="orientation">
+      <enum>Qt::Horizontal</enum>
+     </property>
+    </widget>
+   </item>
    <item row="0" column="0">
     <layout class="QVBoxLayout" name="verticalLayout">
+     <item>
+      <layout class="QHBoxLayout" name="horizontalLayout_3">
```
dpallotti, line 50: I would suggest using group boxes rather than lines. The way it's done with line-label-line in a horizontal layout is weird to me.

dpallotti (commented):
I'll leave the final word on the look and also on the Qt xml stuff I know little about to Florian, but here's my opinion:

- From the screenshot I find the controls waaay too crowded, they would definitely benefit from some margins and some space to breathe.
- I suggest using group boxes.
- The line above "These setting will only apply..." looks redundant to me. Also it appears to have the wrong spacing to the label compared to the line below the label.

dpallotti:
Another thought: Do we need two labels? "Limit local marker depth per command buffer" and "Max local marker depth per command buffer" basically say the same thing.
We could have the checkbox say "Limit local marker depth per command buffer to:" followed by the textbox?

ronaldfw:
> > Do we know whether the layer is enabled when we show the settings? It would be nice to disable that part when the Vulkan layer is not enabled and have a tooltip that explains the layer needs to be enabled first.
> 
> Unfortunately, not. Davide and I have discussed this already in a couple of other contexts, but came to the conclusion that we should not try to expose this information by the IPC.

Ok, then I would suggest that we at least indicate in the tooltip that these settings only apply when the target process has been started with our Vulkan layer enabled. Right now, as a user seeing these settings, I would assume that the Vulkan layer is auto-enabled, also because the next sentence is "These settings will only apply starting from your next capture". 

We don't have to get into details how to enable it (we have documentation for that), only that the target process has to be started with the layer enabled.

src/OrbitQt/CaptureOptionsDialog.cpp
florian-kuebler: Nope, changed it. Thanks!

src/OrbitQt/CaptureOptionsDialog.cpp
florian-kuebler: Done.

src/OrbitQt/CaptureOptionsDialog.h
florian-kuebler: I was missing the fact that we don't use it. I thought it is internal detail. Removed the namespace, as the header is not exposed.

src/OrbitQt/CaptureOptionsDialog.ui
florian-kuebler: obsolete.

src/OrbitQt/CaptureOptionsDialog.ui
florian-kuebler: Done.

florian-kuebler:
> > > Do we know whether the layer is enabled when we show the settings? It would be nice to disable that part when the Vulkan layer is not enabled and have a tooltip that explains the layer needs to be enabled first.
> > 
> > 
> > Unfortunately, not. Davide and I have discussed this already in a couple of other contexts, but came to the conclusion that we should not try to expose this information by the IPC.
> 
> Ok, then I would suggest that we at least indicate in the tooltip that these settings only apply when the target process has been started with our Vulkan layer enabled. Right now, as a user seeing these settings, I would assume that the Vulkan layer is auto-enabled, also because the next sentence is "These settings will only apply starting from your next capture".
> 
> We don't have to get into details how to enable it (we have documentation for that), only that the target process has to be started with the layer enabled.

I added a tooltip for the whole Vulkan layer group.

Also I addressed the review comments so far. Here it how it looks now (on Linux -- will try Windows tomorrow):
Screenshot 2021-02-17 at 18 11 14

reichlfl:
> > > > Do we know whether the layer is enabled when we show the settings? It would be nice to disable that part when the Vulkan layer is not enabled and have a tooltip that explains the layer needs to be enabled first.
> > > 
> > > 
> > > Unfortunately, not. Davide and I have discussed this already in a couple of other contexts, but came to the conclusion that we should not try to expose this information by the IPC.
> > 
> > 
> > Ok, then I would suggest that we at least indicate in the tooltip that these settings only apply when the target process has been started with our Vulkan layer enabled. Right now, as a user seeing these settings, I would assume that the Vulkan layer is auto-enabled, also because the next sentence is "These settings will only apply starting from your next capture".
> > We don't have to get into details how to enable it (we have documentation for that), only that the target process has to be started with the layer enabled.
> 
> I added a tooltip for the whole Vulkan layer group.
> 
> Also I addressed the review comments so far. Here it how it looks now (on Linux -- will try Windows tomorrow):
> Screenshot 2021-02-17 at 18 11 14

"General" and "Vulkan Layer" are the titles of the group boxes, aren't they? It's weird that they are on top of the boxes, but apparently that's how Qt handles it on Linux, so I wouldn't change the default behavior...

Would it be more clear to explicitly add "(requires game execution with --orbit)" to "Vulkan Layer", or is that too much text?

src/OrbitQt/CaptureOptionsDialog.ui
```diff
@@ -9,34 +9,135 @@
...
+        </item>
+       </layout>
       </widget>
      </item>
      <item>
       <spacer name="verticalSpacer">
        <property name="orientation">
         <enum>Qt::Vertical</enum>
        </property>
+       <property name="sizeHint" stdset="0">
```
reichlfl, line 135: Is this a vspacer of 0 height?

reichlfl (approved):
I have mainly looked at the XML / Look now. Instead of working with spacers in the UI, maybe take a look at the spacing options in the vertical layout that surrounds the group boxes. I think it still looks a bit "condensed", but way better than the initial version. As I wrote above, usually the default values in Qt layouts look good to me, maybe this is also just due to Linux.

src/OrbitQt/CaptureOptionsDialog.ui
```diff
@@ -9,34 +9,135 @@
...
+             <string>0</string>
+            </property>
+           </widget>
+          </item>
+          <item>
+           <spacer name="horizontalSpacer">
+            <property name="orientation">
+             <enum>Qt::Horizontal</enum>
+            </property>
+            <property name="sizeHint" stdset="0">
```
dpallotti, line 117: It this a spacer at the end of the line?

src/OrbitQt/CaptureOptionsDialog.ui
```diff
@@ -9,34 +9,135 @@
...
+          <property name="text">
+           <string>Collect thread states</string>
+          </property>
+         </widget>
+        </item>
+       </layout>
+      </widget>
+     </item>
+     <item>
+      <spacer name="verticalSpacer_2">
```
dpallotti, line 66: Is this spacer between the group boxes? Can you do it with a margin?

src/OrbitQt/CaptureOptionsDialog.ui
```diff
@@ -9,34 +9,135 @@
...
+          <item>
+           <widget class="QLineEdit" name="localMarkerDepthLineEdit">
+            <property name="enabled">
+             <bool>false</bool>
+            </property>
+            <property name="inputMethodHints">
+             <set>Qt::ImhDigitsOnly</set>
+            </property>
+            <property name="text">
+             <string>0</string>
```
dpallotti, line 108: Is 0 the default value for the max depth? When I first check the checkbox, is 0 the first value that is automatically inserted? I would think it makes more sense to have a positive value that we think is sensible.

dpallotti (approved):
I agree with @reichlfl and strongly suggest to use margins instead of spacers for fixed-sized spaces between components.

I would also change the default limit from 0, I find it a bit confusing (as it means that checking the checkbox without changing the value is equivalent to disabling markers altogether).

Other than that it looks good.

src/OrbitQt/CaptureOptionsDialog.ui
florian-kuebler: Done. Used the layout options of the verticalLayout.

src/OrbitQt/CaptureOptionsDialog.ui
florian-kuebler: Changed it to 3 and re-added the tooltip.

src/OrbitQt/CaptureOptionsDialog.ui
florian-kuebler: Oh, yeah I missed to remove this one. Done.

src/OrbitQt/CaptureOptionsDialog.ui
florian-kuebler: That was also a leftover (with zero effect). I've set this to default again.

florian-kuebler:
> > > > > Do we know whether the layer is enabled when we show the settings? It would be nice to disable that part when the Vulkan layer is not enabled and have a tooltip that explains the layer needs to be enabled first.
> > > > 
> > > > 
> > > > Unfortunately, not. Davide and I have discussed this already in a couple of other contexts, but came to the conclusion that we should not try to expose this information by the IPC.
> > > 
> > > 
> > > Ok, then I would suggest that we at least indicate in the tooltip that these settings only apply when the target process has been started with our Vulkan layer enabled. Right now, as a user seeing these settings, I would assume that the Vulkan layer is auto-enabled, also because the next sentence is "These settings will only apply starting from your next capture".
> > > We don't have to get into details how to enable it (we have documentation for that), only that the target process has to be started with the layer enabled.
> > 
> > 
> > I added a tooltip for the whole Vulkan layer group.
> > Also I addressed the review comments so far. Here it how it looks now (on Linux -- will try Windows tomorrow):
> > Screenshot 2021-02-17 at 18 11 14
> 
> "General" and "Vulkan Layer" are the titles of the group boxes, aren't they? It's weird that they are on top of the boxes, but apparently that's how Qt handles it on Linux, so I wouldn't change the default behavior...
> 
> I think the space between the two group boxes is still a bit small, is that a vertical layout with default settings? Usually Qt does a decent job with its default spacings and alignments on Windows.
> 
> Would it be more clear to explicitly add "(requires game execution with --orbit)" to "Vulkan Layer", or is that too much text?

I removed the vertical spacer and adjusted the layout options a bit. Also added a note to the Vulkan layer title, and set the default value to 3.

Screenshot 2021-02-18 at 10 11 26

florian-kuebler:
Merged this now. Feel free to reach out for follow up improvements.

PR #1888, opened 2021-02-17
Demangle function names when processing FullAddressInfo

dimitry-:
Test: run ServiceTests
Bug: http://b/180397173

### DIFF
--- a/src/LinuxTracing/UprobesUnwindingVisitor.cpp
+++ b/src/LinuxTracing/UprobesUnwindingVisitor.cpp
@@ -6,6 +6,7 @@
 
 #include <absl/container/flat_hash_map.h>
 #include <asm/perf_regs.h>
+#include <llvm/Demangle/Demangle.h>
 #include <sys/mman.h>
 #include <unwindstack/MapInfo.h>
 #include <unwindstack/Unwinder.h>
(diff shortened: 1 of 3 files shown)


florian-kuebler (approved):
LGTM, thanks!

dpallotti (approved):
Where were they demangled before your big refactoring of protos?

dimitry-:
> Where were they demangled before your big refactoring of protos?

yes - it LinuxTracingHandler (previous implementation)

I wonder if I should change the function_name in FullAddressInfo to mangled_function_name for clarity... this is kind of implicit now. We have mangled string in FullAdressInfo and demangled in AddressInfo

dpallotti:
> We have mangled string in FullAdressInfo and demangled in AddressInfo

Ok, this is a bit weird honestly. What about demangling somewhere else then, like right after unwinding where `FullAddressInfo` is produced or even on the client side? Where are names from the symbols demangled?

dimitry-:
> > We have mangled string in FullAdressInfo and demangled in AddressInfo
> 
> Ok, this is a bit weird honestly. What about demangling somewhere else then, like right after unwinding where `FullAddressInfo` is produced or even on the client side? Where are names from the symbols demangled?

makes sense. At least we are consistent in this case.

dpallotti:
> > > We have mangled string in FullAdressInfo and demangled in AddressInfo
> > 
> > 
> > Ok, this is a bit weird honestly. What about demangling somewhere else then, like right after unwinding where `FullAddressInfo` is produced or even on the client side? Where are names from the symbols demangled?
> 
> makes sense. At least we are consistent in this case.

Not sure which one of the two you are referring to, but if I get it correctly names from loaded modules are unmangled as soon as possible in `ElfFile`, so I would just do something similar and unwind before producing `FullAddressInfo`. Or otherwise we could unmangle everything on the client side, but then it should be everything.

dpallotti (approved):

florian-kuebler (approved):
Still LGTM. Thanks!

PR #1889, opened 2021-02-17
Ensure a stable ordering of the GpuTracks -- align marker tracks.

florian-kuebler:
This stablelizes the ordering of GpuTracks by also using an ordered
std::map<string,track> (as already done for other tracks).
Main benefit is, that the marker tracks are always below and alligned
to their corresponding queue track.

Test: Take multiple captures with Vulkan layer and sort tracks.
Bug: http://b/179888337

### DIFF
--- a/src/OrbitGl/TrackManager.cpp
+++ b/src/OrbitGl/TrackManager.cpp
@@ -338,11 +338,13 @@ ThreadTrack* TrackManager::GetOrCreateThreadTrack(int32_t tid) {
 
 GpuTrack* TrackManager::GetOrCreateGpuTrack(uint64_t timeline_hash) {
   std::lock_guard<std::recursive_mutex> lock(mutex_);
-  std::shared_ptr<GpuTrack> track = gpu_tracks_[timeline_hash];
+  std::string timeline =
+      app_->GetStringManager()->Get(timeline_hash).value_or(std::to_string(timeline_hash));
+  std::shared_ptr<GpuTrack> track = gpu_tracks_[timeline];
   if (track == nullptr) {
     track = std::make_shared<GpuTrack>(time_graph_, layout_, timeline_hash, app_, capture_data_);
     AddTrack(track);
-    gpu_tracks_[timeline_hash] = track;
+    gpu_tracks_[timeline] = track;
   }
   return track.get();
 }
--- a/src/OrbitGl/TrackManager.h
+++ b/src/OrbitGl/TrackManager.h
@@ -77,7 +77,7 @@ class TrackManager {
   std::unordered_map<int32_t, std::shared_ptr<ThreadTrack>> thread_tracks_;
   std::map<std::string, std::shared_ptr<AsyncTrack>> async_tracks_;
   std::map<std::string, std::shared_ptr<GraphTrack>> graph_tracks_;
-  // Mapping from timeline to GPU tracks.
+  // Mapping from timeline to GPU tracks. Timeline name is used for stable ordering.
   std::map<std::string, std::shared_ptr<GpuTrack>> gpu_tracks_;
   // Mapping from function address to frame tracks.
   // TODO (b/175865913): Use Function info instead of their address as key to FrameTracks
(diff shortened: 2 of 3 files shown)


src/OrbitGl/TrackManager.cpp
```diff
@@ -338,11 +338,13 @@ ThreadTrack* TrackManager::GetOrCreateThreadTrack(int32_t tid) {
 
 GpuTrack* TrackManager::GetOrCreateGpuTrack(uint64_t timeline_hash) {
   std::lock_guard<std::recursive_mutex> lock(mutex_);
-  std::shared_ptr<GpuTrack> track = gpu_tracks_[timeline_hash];
+  std::string timeline =
+      app_->GetStringManager()->Get(timeline_hash).value_or(std::to_string(timeline_hash));
```
karupayun, line 342: Why not use Track's name? If you don't like that there could be several with the same name (""), you could add the default value "timeline_hash" to Track's name in the constructor.

karupayun (approved):
LGTM! Nice and logical change, but please consider my comment :)

ronaldfw:
Inverting this change (using timeline_hash instead of the timeline string) is a "nice" optimization that breaks intended behavior and there's not really anything preventing someone from doing that (not likely to happen now, but in 6 months?). 

I would suggest: Add a comment why you are using timeline string and not timeline hash as the key and that std::map should be used (correct sorting behavior, ordering in std::map, etc.). A test would be better but that's probably a lot of work as this class doesn't have a test yet.

src/OrbitGl/TrackManager.cpp
florian-kuebler: I can't, as the track might have not yet been created. However, I also adjusted the Track's name.

florian-kuebler:
> Inverting this change (using timeline_hash instead of the timeline string) is a "nice" optimization that breaks intended behavior and there's not really anything preventing someone from doing that (not likely to happen now, but in 6 months?).
> 
> I would suggest: Add a comment why you are using timeline string and not timeline hash as the key and that std::map should be used (correct sorting behavior, ordering in std::map, etc.). A test would be better but that's probably a lot of work as this class doesn't have a test yet.

I added a comment. Was also looking into how much work it would be to add a test. I guess the main burden is the call to "app_->GetStringManager", which would require a bit of refactoring to be able to mock it.
AFAIK, making TrackManager testable is on our roadmap. I will have an eye on that and add a test as soon is this effort starts.

src/OrbitGl/TrackManager.h
```diff
@@ -77,8 +77,8 @@ class TrackManager {
   std::unordered_map<int32_t, std::shared_ptr<ThreadTrack>> thread_tracks_;
   std::map<std::string, std::shared_ptr<AsyncTrack>> async_tracks_;
   std::map<std::string, std::shared_ptr<GraphTrack>> graph_tracks_;
-  // Mapping from timeline hash to GPU tracks.
-  std::unordered_map<uint64_t, std::shared_ptr<GpuTrack>> gpu_tracks_;
+  // Mapping from timeline to GPU tracks. Timeline name is used for stable ordering.
```
ronaldfw, line 80: I would add a bit more information here. Why do we want stable ordering for example? Just from the current comment alone it's still hard to understand why I can't change this from timeline name to timeline hash. The ordering is just as stable in that case (if I use std::map), the exact order of tracks is just not what we want then. We explicitly want the ordering on timeline strings (so that tracks like "gfx" and "gfx_markers" appear next to each other in the UI).

src/OrbitGl/TrackManager.cpp
karupayun: Yes, you are right. I only was worried about using twice this "GetTimeline". I've just checked if we can only give timeline instead of the hash to the GpuTrack constructor, but seems to be that we also need the timeline_hash for other functions (GetLeft and GetRight).

karupayun (approved):
LGTM by my side!!

src/OrbitGl/TrackManager.h
florian-kuebler: I extended the comment.

ronaldfw (approved):
LGTM, thanks

PR #1890, opened 2021-02-17
Replace fstream usage with ReadFileToString in ElfFileTests

dimitry-:


beckerhe (approved):
The change looks good to me. But I guess we need a windows version of ReadFileToString, don't we?

dimitry-:
> The change looks good to me. But I guess we need a windows version of ReadFileToString, don't we?

indeed - working on it.

beckerhe (approved):
Nice! LGTM!

PR #1892, opened 2021-02-17
Add "Go To Source" action to context menus

beckerhe:
This adds a feature flag and the source code view action
to the context menus in all data views.

This context menu entry is only shown when the feature flags is set to
true (provided on the command line)

### DIFF
--- a/src/OrbitGl/App.cpp
+++ b/src/OrbitGl/App.cpp
@@ -90,6 +90,7 @@
 ABSL_DECLARE_FLAG(bool, devmode);
 ABSL_DECLARE_FLAG(bool, local);
 ABSL_DECLARE_FLAG(bool, enable_tracepoint_feature);
+ABSL_DECLARE_FLAG(bool, enable_source_code_view);
 
 using orbit_client_protos::CallstackEvent;
 using orbit_client_protos::FunctionInfo;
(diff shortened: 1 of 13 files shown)


src/OrbitGl/ClientFlags.cpp
```diff
@@ -35,3 +35,5 @@ ABSL_FLAG(bool, show_return_values, false, "Show return values on time slices");
 
 ABSL_FLAG(bool, enable_tracepoint_feature, false,
           "Enable the setting of the panel of kernel tracepoints");
+
+ABSL_FLAG(bool, enable_sourcecode_view, false, "Enable the experimental source code view");
```
antonrohr, line 39: nit: Since everywhere else its `SourceCode`
```suggestion
ABSL_FLAG(bool, enable_source_code_view, false, "Enable the experimental source code view");
```

src/OrbitGl/App.cpp
```diff
@@ -550,6 +551,11 @@ void OrbitApp::Disassemble(int32_t pid, const FunctionInfo& function) {
   });
 }
 
+void OrbitApp::ShowSourceCode(const orbit_client_protos::FunctionInfo& function) {
+  // For now this is just a stub. This will change in a subsequent commit
```
antonrohr, line 555: pers pref: //TODO(bug/feature)

antonrohr (approved):
just two nits

src/OrbitGl/ClientFlags.cpp
beckerhe: Done.

src/OrbitGl/App.cpp
beckerhe: I'd agree if there were a bug which had further information about the todo. And creating a bug just for the sole purpose of having something to reference defeats the purpose.

PR #1897, opened 2021-02-18
Implement WriteStringToFile function

dimitry-:
For the cases when we want to write a string to file.

Test: run OrbitBaseTests

### DIFF
--- a/src/OrbitBase/WriteStringToFile.cpp
+++ b/src/OrbitBase/WriteStringToFile.cpp
@@ -1,0 +1,72 @@
+// Copyright (c) 2021 The Orbit Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#include "OrbitBase/WriteStringToFile.h"
+
+#include <absl/strings/str_format.h>
+#include <errno.h>
+#include <fcntl.h>
+#include <stdint.h>
+#include <sys/stat.h>
+#include <sys/types.h>
+
+#if defined(__linux)
+#include <unistd.h>
+#elif defined(_WIN32)
+#include <io.h>
+// Windows does not have TEMP_FAILURE_RETRY - define a shortcut
+#define TEMP_FAILURE_RETRY(expression) (expression)
+#endif
+
+#include "OrbitBase/Logging.h"
+#include "OrbitBase/SafeStrerror.h"
+#include "OrbitBase/UniqueResource.h"
+
+namespace orbit_base {
+
+ErrorMessageOr<void> WriteStringToFileImpl(int fd, std::string_view content) {
+  int64_t bytes_left = content.size();
+  const char* current_position = content.data();
+  while (bytes_left > 0) {
+    int64_t bytes_written = TEMP_FAILURE_RETRY(write(fd, current_position, bytes_left));
+    if (bytes_written == -1) {
+      return ErrorMessage{SafeStrerror(errno)};
+    }
+    current_position += bytes_written;
+    bytes_left -= bytes_written;
+  }
+
+  CHECK(bytes_left == 0);
+  return outcome::success();
+}
+
+ErrorMessageOr<void> WriteStringToFile(const std::filesystem::path& file_name,
+                                       std::string_view content) {
+#if defined(__linux)
+  constexpr int open_flags = O_WRONLY | O_CREAT | O_TRUNC | O_CLOEXEC;
+  constexpr int open_mode = 0600;
+#elif defined(_WIN32)
+  constexpr int open_flags = O_WRONLY | O_CREAT | O_TRUNC | O_BINARY;
+  constexpr int open_mode = _S_IREAD | _S_IWRITE;
+#endif  // defined(__linux)
+  auto fd = unique_resource(
+      TEMP_FAILURE_RETRY(open(file_name.string().c_str(), open_flags, open_mode)), [](int fd) {
+        if (fd != -1) close(fd);
+      });
+  if (fd == -1) {
+    return ErrorMessage(
+        absl::StrFormat("Unable to open file \"%s\": %s", file_name.string(), SafeStrerror(errno)));
+  }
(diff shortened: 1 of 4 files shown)


src/OrbitBase/include/OrbitBase/WriteStringToFile.h
```diff
@@ -0,0 +1,18 @@
+// Copyright (c) 2021 The Orbit Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#ifndef ORBIT_BASE_WRITE_STRING_TO_FILE_H_
+#define ORBIT_BASE_WRITE_STRING_TO_FILE_H_
+
+#include <filesystem>
+#include <string>
```
beckerhe, line 9: <string_view>

beckerhe (approved):
Otherwise LGTM

src/OrbitBase/include/OrbitBase/WriteStringToFile.h
dimitry-: done.

src/OrbitBase/WriteStringToFile.cpp
```diff
@@ -0,0 +1,72 @@
...
+
+#if defined(__linux)
+#include <unistd.h>
+#elif defined(_WIN32)
+#include <io.h>
+// Windows does not have TEMP_FAILURE_RETRY - define a shortcut
+#define TEMP_FAILURE_RETRY(expression) (expression)
+#endif
+
+#include "OrbitBase//Logging.h"
```
dpallotti, line 22:
```suggestion
#include "OrbitBase/Logging.h"
```

src/OrbitBase/WriteStringToFile.cpp
```diff
@@ -0,0 +1,72 @@
...
+        if (fd != -1) close(fd);
+      });
+  if (fd == -1) {
+    return ErrorMessage(
+        absl::StrFormat("Unable to open file \"%s\": %s", file_name.string(), SafeStrerror(errno)));
+  }
+
+  ErrorMessageOr<void> result = WriteStringToFileImpl(fd, content);
+  if (!result) {
+    unlink(file_name.string().c_str());
```
dpallotti, line 64: I think we should use `remove`, `unlink` is not portable.

src/OrbitBase/WriteStringToFileTest.cpp
```diff
@@ -0,0 +1,31 @@
...
+  size_t content_size = 16;
+  const char* expected_content = "content\nnew line";
+  std::string_view content(full_content, content_size);
+
+  ErrorMessageOr<void> result = WriteStringToFile(temp_file_name, content);
+  ASSERT_TRUE(result) << result.error().message();
+  ErrorMessageOr<std::string> actual_content = ReadFileToString(temp_file_name);
+  ASSERT_TRUE(actual_content) << actual_content.error().message();
+  EXPECT_EQ(actual_content.value(), expected_content);
+  unlink(temp_file_name);  // TODO(http://b/180574275): see above
```
dpallotti, line 29: Same here, why not `remove`?

src/OrbitBase/WriteStringToFileTest.cpp
```diff
@@ -0,0 +1,31 @@
...
+#include "OrbitBase/ReadFileToString.h"
+#include "OrbitBase/Result.h"
+#include "OrbitBase/WriteStringToFile.h"
+
+namespace orbit_base {
+TEST(WriteStringToFile, Smoke) {
+  // TODO(http://b/180574275): Replace this with temporary_file once it becomes available
+  const char* temp_file_name = std::tmpnam(nullptr);
+  const char* full_content = "content\nnew line(this text is not written)";
+  size_t content_size = 16;
```
dpallotti, line 20: Make it a constant?

dpallotti (approved):

src/OrbitBase/WriteStringToFile.cpp
dimitry-: done.

src/OrbitBase/WriteStringToFile.cpp
dimitry-: so the difference between remove and unlink is that remove can remove directories where unlink is specifically for files. It seems to be portable between linux and windows. what am I missing here?

src/OrbitBase/WriteStringToFileTest.cpp
dimitry-: done.

src/OrbitBase/WriteStringToFile.cpp
dpallotti: `unlink` is defined in `<unistd.h>`, that's enough of a sign for me. Yes, the same is provided on Windows in this case, but it's a deprecated alias: https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/unlink?view=msvc-160. 
So both work but `unlink` is the weird choice here.

dimitry-: same goes for open/read/write and remove for that matter. I think at this point we accept the fact that even thought microsoft is trying to deprecate posix there is enough users for this to be just a wish.

dpallotti: `remove` is defined in `<cstdio>`. For me the fact that we have to use posix functions for some things doesn't necessarily mean that we always have to do that even when there's an alternative.

dpallotti: This said you clearly seem to prefer `unlink` and this is really not important so in the end I'm fine with both.

dimitry-: tbh I do not see the point - they are both supported on windows and linux, on linux there is a clear difference between unlink and remove. but well, I do not have strong feeling here - will change to remove.

PR #1898, opened 2021-02-18
Fix connection loss

antonrohr:
This PR mainly adds support for ssh connection loses. Also some improved error messages, appropriate disabling of actions,  refactoring and general moving around of things. I tried to split it up in multiple commits to make the review easier.

This is tracked by http://b/177305866
This also properly fixes http://b/180327548

### DIFF
--- a/src/OrbitQt/ConnectToStadiaWidget.cpp
+++ b/src/OrbitQt/ConnectToStadiaWidget.cpp
@@ -88,10 +88,23 @@ void ConnectToStadiaWidget::SetActive(bool value) {
   ui_->radioButton->setChecked(value);
 }
 
-void ConnectToStadiaWidget::Start(SshConnectionArtifacts* ssh_connection_artifacts,
-                                  std::optional<StadiaConnection> connection) {
+void ConnectToStadiaWidget::SetSshConnectionArtifacts(
+    SshConnectionArtifacts* ssh_connection_artifacts) {
   CHECK(ssh_connection_artifacts != nullptr);
   ssh_connection_artifacts_ = ssh_connection_artifacts;
+}
+
+void ConnectToStadiaWidget::SetConnection(StadiaConnection connection) {
+  selected_instance_ = std::move(connection.instance_);
+  service_deploy_manager_ = std::move(connection.service_deploy_manager_);
+  grpc_channel_ = std::move(connection.grpc_channel_);
+}
+
+void ConnectToStadiaWidget::Start() {
+  if (ssh_connection_artifacts_ == nullptr) {
+    ERROR("Unable to start ConnectToStadiaWidget: ssh_connection_artifacts_ is nullptr");
+    return;
+  }
 
   auto client_result = orbit_ggp::Client::Create(this);
   if (!client_result) {
@@ -101,10 +114,7 @@ void ConnectToStadiaWidget::Start(SshConnectionArtifacts* ssh_connection_artifac
   }
   ggp_client_ = client_result.value();
 
-  if (connection.has_value()) {
-    selected_instance_ = std::move(connection->instance_);
-    service_deploy_manager_ = std::move(connection->service_deploy_manager_);
-    grpc_channel_ = std::move(connection->grpc_channel_);
+  if (grpc_channel_ != nullptr && grpc_channel_->GetState(false) == GRPC_CHANNEL_READY) {
     state_machine_.setInitialState(&s_connected_);
   } else {
     state_machine_.setInitialState(&s_instances_loading_);
--- a/src/OrbitQt/ProfilingTargetDialog.cpp
+++ b/src/OrbitQt/ProfilingTargetDialog.cpp
@@ -87,6 +87,7 @@ ProfilingTargetDialog::ProfilingTargetDialog(
   CHECK(ssh_connection_artifacts != nullptr);
 
   ui_->setupUi(this);
+  ui_->stadiaWidget->SetSshConnectionArtifacts(ssh_connection_artifacts);
   ui_->processesTableOverlay->raise();
 
   state_machine_.setGlobalRestorePolicy(QStateMachine::RestoreProperties);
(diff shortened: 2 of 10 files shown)


src/OrbitQt/orbitmainwindow.cpp
```diff
@@ -1131,12 +1133,25 @@ void OrbitMainWindow::closeEvent(QCloseEvent* event) {
...
+                   [this, instance_name = target.GetConnection()->GetInstance().display_name](
+                       std::error_code error) {
+                     QMessageBox::critical(
+                         this, "Connection error",
+                         QString("The connection to instance \"%1\" failed with error message: %2")
+                             .arg(instance_name)
+                             .arg(QString::fromStdString(error.message())),
+                         QMessageBox::Ok);
+                     QApplication::exit(kEndSessionReturnCode);
+                   });
```
beckerhe: As far as I can see this connection could technically be made twice between the same objects, when SetTarget is called twice. So I would add the Qt::UniqueConnection flag. https://doc.qt.io/qt-5/qt.html#ConnectionType-enum

src/OrbitQt/ConnectToStadiaWidget.cpp
```diff
@@ -5,6 +5,7 @@
 #include "ConnectToStadiaWidget.h"
 
 #include <absl/strings/str_format.h>
+#include <grpc/impl/codegen/connectivity_state.h>
```
beckerhe, line 8: Is clangd suggesting that? Looks like an internal header to me. I assume iwyu suggests something else instead.

src/OrbitQt/ConnectToStadiaWidget.cpp
```diff
@@ -88,10 +89,23 @@ void ConnectToStadiaWidget::SetActive(bool value) {
   ui_->radioButton->setChecked(value);
 }
 
-void ConnectToStadiaWidget::Start(SshConnectionArtifacts* ssh_connection_artifacts,
-                                  std::optional<StadiaConnection> connection) {
+void ConnectToStadiaWidget::SetSshConnectionArtifacts(
+    SshConnectionArtifacts* ssh_connection_artifacts) {
```
beckerhe, line 93: From an API design perspective this operation should be reversible, even if the class is currently used only in one situation in the code. The user needs some way to clear the pointer again, even though the one existing user doesn't care. So maybe add a `ClearSshConnectionArtifacts` method which sets the internal pointer to `nullptr`?

src/OrbitQt/orbitmainwindow.h
```diff
@@ -168,6 +169,8 @@ class OrbitMainWindow : public QMainWindow {
   [[nodiscard]] bool ConfirmExit();
   void Exit(int return_code);
 
+  void StadiaConnectionError(std::error_code error);
```
beckerhe, line 172: It's a bit unclear if that function is an event handler which is reacting to a connection error or if this is actually an event source which is called from something internal. From the implementation I know it's the former. So maybe something like `OnStadiaConnectionError`?

beckerhe (approved):
Otherwise LGTM.

But it's a large PR. Someone else should approve as well.

src/OrbitQt/ProfilingTargetDialog.cpp
```diff
@@ -495,11 +492,16 @@ void ProfilingTargetDialog::ConnectToLocal() {
...
+    process_manager_ = std::move(target.process_manager_);
+    process_ = std::move(target.process_);
+
+    state_stadia_.setInitialState(&state_stadia_connected_);
+    state_stadia_history_.setDefaultState(&state_stadia_connected_);
+  }
 
-  state_stadia_.setInitialState(&state_stadia_connected_);
-  state_stadia_history_.setDefaultState(&state_stadia_connected_);
   state_machine_.setInitialState(&state_stadia_);
```
vickyliu-go4it, line 505: May be we can also rename the function `SetInitalStateFromTarget` as `SetStateMachineInitalStateFromTarget` and move this line for setting initial state for `state_machine_` there. If so, we can also have an early return in this method.

vickyliu-go4it (approved):
Overall look good! I have some minor suggestions, but they are mostly some personal preferences. Please feel free to ignore them. Thanks for making this change! LGTM.

src/OrbitQt/orbitmainwindow.cpp
antonrohr: I did not know about those. Good idea, I added it.

src/OrbitQt/ConnectToStadiaWidget.cpp
antonrohr: done

src/OrbitQt/orbitmainwindow.h
antonrohr: done

src/OrbitQt/orbitmainwindow.h
```diff
@@ -201,7 +206,9 @@ class OrbitMainWindow : public QMainWindow {
 
   enum class TargetProcessState { kRunning, kEnded };
 
-  TargetProcessState target_process_state_ = TargetProcessState::kRunning;
+  TargetProcessState target_process_state_ = TargetProcessState::kEnded;
+
+  bool is_connected_ = false;
```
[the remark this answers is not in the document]
antonrohr: good idea! I added a comment

src/OrbitQt/ProfilingTargetDialog.cpp
```diff
@@ -139,20 +140,24 @@ ProfilingTargetDialog::ProfilingTargetDialog(
...
-      state_machine_.setInitialState(&state_stadia_);
-    } else {
-      state_machine_.setInitialState(&state_file_);
-    }
-    ui_->stadiaWidget->Start(ssh_connection_artifacts);
+    SetInitialState();
+  }
+}
+
+void ProfilingTargetDialog::SetInitialState() {
```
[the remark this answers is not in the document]
antonrohr: I agree it might be confusing. I added the `StateMachine` part of the name

src/OrbitQt/ProfilingTargetDialog.cpp
antonrohr: I changed it. ptal

vickyliu-go4it: Looks good. Thanks for making the change : )

src/OrbitQt/ConnectToStadiaWidget.cpp
antonrohr: Yes it was clangd. iwyu did not complain when it was in there. I removed it and everything still works, also iwyu is not complaining as far as I can tell

antonrohr:
@beckerhe I would rebase and merge this soon, if there are no objections.

beckerhe:
> @beckerhe I would rebase and merge this soon, if there are no objections.

Sry, yes. Sure.

PR #1899, opened 2021-02-18
Add code_viewer::Dialog::GoToLineNumber

beckerhe:
This allows programmatically to jump to a specific line in the code
viewer.

### DIFF
--- a/src/CodeViewer/Dialog.cpp
+++ b/src/CodeViewer/Dialog.cpp
@@ -48,4 +48,12 @@ void Dialog::SetLineNumberMargins(FontSizeInEm left, FontSizeInEm right) {
 
 void Dialog::SetEnableLineNumbers(bool enabled) { ui_->viewer->SetEnableLineNumbers(enabled); }
 
+void Dialog::GoToLineNumber(size_t line_number) {
+  const QTextBlock block =
+      ui_->viewer->document()->findBlockByLineNumber(static_cast<int>(line_number));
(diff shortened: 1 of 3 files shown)


src/CodeViewer/Dialog.cpp
```diff
@@ -48,4 +48,12 @@ void Dialog::SetLineNumberMargins(FontSizeInEm left, FontSizeInEm right) {
 
 void Dialog::SetEnableLineNumbers(bool enabled) { ui_->viewer->SetEnableLineNumbers(enabled); }
 
+void Dialog::GoToLineNumber(size_t line_number) {
```
karupayun, line 51: Stupid question: You don't like int here and avoid the cast because line_number should be an unsigned and you prefer to make that clear in the interface?

karupayun (approved):
LGTM! :)

src/CodeViewer/Dialog.cpp
beckerhe: That's not a stupid question at all!

In this specific case: I first designed the interface (using size_t) because of what you said (to make it clear it's a quantity).
And then later when writing the implementation I realized that Qt uses `int`. So I had to add the cast.

But I don't really feel strongly about either way. There are good reasons for always using signed numbers and avoiding unsigned types completely. I think we tend to use unsigned types in the code base, that's why I left it. In other cases I might have changed the code to avoid the cast.

PR #1902, opened 2021-02-18
Make SymbolHelper functions exception-free

dimitry-:
Test: run OrbitCoreTests

### DIFF
--- a/src/OrbitCore/SymbolHelper.cpp
+++ b/src/OrbitCore/SymbolHelper.cpp
@@ -7,18 +7,20 @@
 #include <absl/strings/match.h>
 #include <absl/strings/str_format.h>
 #include <absl/strings/str_replace.h>
+#include <absl/strings/str_split.h>
 
 #include <algorithm>
 #include <filesystem>
-#include <fstream>
 #include <memory>
 #include <outcome.hpp>
 #include <set>
 
 #include "ElfUtils/ElfFile.h"
 #include "OrbitBase/Logging.h"
+#include "OrbitBase/ReadFileToString.h"
 #include "OrbitBase/Result.h"
 #include "OrbitBase/Tracing.h"
+#include "OrbitBase/WriteStringToFile.h"
 #include "Path.h"
 
 using orbit_grpc_protos::ModuleSymbols;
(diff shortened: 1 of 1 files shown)


src/OrbitCore/SymbolHelper.cpp
```diff
@@ -30,39 +33,61 @@ namespace {
...
   fs::path file_name = Path::GetSymbolsFileName();
-  if (!fs::exists(file_name)) {
-    std::ofstream outfile(file_name);
-    outfile << "//-------------------" << std::endl
-            << "// Orbit Symbol Locations" << std::endl
-            << "//-------------------" << std::endl
-            << "// Orbit will scan the specified directories for symbol files." << std::endl
-            << "// Enter one directory per line, like so:" << std::endl
+  std::error_code error;
+  ErrorMessageOr<bool> file_exists = fs::exists(file_name, error);
```
dimitry-, line 37: well.. something went wrong here...

dimitry-: fixed.

src/OrbitCore/SymbolHelper.cpp
```diff
@@ -116,7 +141,14 @@ ErrorMessageOr<fs::path> SymbolHelper::FindSymbolsWithSymbolsPathFile(
...
   for (const auto& symbols_path : search_paths) {
-    if (!fs::exists(symbols_path)) continue;
+    std::error_code error;
+    bool exists = fs::exists(symbols_path, error);
+    if (error) {
+      ERROR("Unable to stat \"%s\": %s", symbols_path.string(), error.message());
+      continue;
+    }
+
+    if (!exists) continue;
```
pierricgimmig, line 151: Should we warn here?

pierricgimmig (approved):
Hi Dimitry, LGTM

pierricgimmig:
Also, nit. in description: "excpetion-free" -> "exception-free"

src/OrbitCore/SymbolHelper.cpp
dimitry-: I believe this is perfectly fine since this method basically check the list which is constructed without knowledge of which files exists and which do not. it may contain for example "/path1/liba.so.debug", "/path2/lib.so.debug" if one of these exists symbol will be found. it is ok for both of them to not exists because in this case the caller will try to download them from the service or do something else.

PR #1903, opened 2021-02-19
Source Code Syntax Highlighter 1.0

karupayun:
This PR upgrades the C Syntax Highlighter adding C++ rules. More
specific:

 - Upgrades number and keyword regex.

 - Modifies function regex to only include function definitions
 (as discussed)

 - Introduces class members and add examples (track_, m_track, ..)

 - Introduces comma regex (to paint ',' and ';'), and also modifies
 capitalized words (to not paint '*' and '&' in variables).

 - Update few colors following CLion color conventions and uses a new
 one to paint class' names and types like std::vector.

Screenshots:
file
tests

To test it:
 - Run CodeViewerDemo. You will see there some examples and a real file.

Last step in http://b/177651839.

### DIFF
--- a/src/CodeViewerDemo/CodeExamples.h
+++ b/src/CodeViewerDemo/CodeExamples.h
@@ -19,17 +19,25 @@ inline const QString testing_example =
   OrbitService o;
   QSyntaxHighlighter q;
   ORBIT_SCOPE; // no lowercase regex
-  x;
-  kTreeSize == 3;
-  _Tree_;
+  x = _Tree_;
+  bool P = (kTreeSize == 3);
+
+  // Class members
+  Track* track_;
+  int __int__ = 3;
+  m_PascalCase = 5;
+  Track::SortTracks();
+  class TrackManager {};
 
   // Function unit tests
-  int my_function(void)
+  int my_function(void);
+  int MyFunction(void);
   bool b = a.insert(3);
   void AddEvent();
   char _my_underscore_func();
-  int CaptureServiceImpl::Capture() // Function from other class
-  void QSyntaxHighlighter::Play
+  int CaptureServiceImpl::Capture(); // Function from other class
+  a = absl::Milliseconds(20); // method that we don't have to highlight
+  int QSyntaxHighlighter::PlayInteger;
   CHECK(false); // no lowercase regex
 
   // Preprocessor + <> unit tests
--- a/src/SyntaxHighlighter/Cpp.cpp
+++ b/src/SyntaxHighlighter/Cpp.cpp
@@ -10,10 +10,8 @@
 #include <QTextCharFormat>
 
 namespace orbit_syntax_highlighter {
-namespace C {
 namespace {
-enum HighlighterState { kInitialState, kOpenCommentState, kOpenStringState };
-
+// Colors are taken from Clion Darcula theme
 const QColor kGrey{0x80, 0x80, 0x80};
 const QColor kBlue{0x61, 0x96, 0xcc};
 const QColor kYellow{0xa0, 0xa0, 0x33};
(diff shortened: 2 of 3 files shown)


src/CodeViewerDemo/CodeExamples.h
```diff
@@ -23,12 +23,21 @@ inline const QString testing_example =
   kTreeSize == 3;
   _Tree_;
 
+  // Class members
+  Track* track_;
+  __int__;
+  m_Text_y;
```
florian-kuebler, line 29: If I am not wrong, the code style here would be `m_PascalCase`. Not sure if that matters.

src/CodeViewerDemo/CodeExamples.h
```diff
@@ -23,12 +23,21 @@ inline const QString testing_example =
...
+  Track::SortTracks();
+  class TrackManager {
+
   // Function unit tests
   int my_function(void)
+  int MyFunction(void)
```
florian-kuebler, line 35: should this end with a `;`? Similar above.

src/CodeViewerDemo/CodeExamples.h
```diff
@@ -23,12 +23,21 @@ inline const QString testing_example =
   kTreeSize == 3;
   _Tree_;
 
+  // Class members
+  Track* track_;
+  __int__;
+  m_Text_y;
+  Track::SortTracks();
+  class TrackManager {
+
```
florian-kuebler, line 32: Should you add a closing `}`?

src/SyntaxHighlighter/Cpp.cpp
```diff
@@ -43,8 +46,8 @@ const char* const kOpenCommentRegex = "\\/\\*([^\\*]|[\\*]+[^\\/])*?$";
 const char* const kEndCommentRegex = "[\\s\\S]*\\*\\/";
 // Match with a line without a closing comment
 const char* const kNoEndCommentRegex = "([^\\*]|\\*+[^\\/\\*])*$";
-// A word who is followed with an open-parenthesis (
-const char* const kFunctionRegex = "(?<=[^\\(\\:\\.\\w\\>])[a-z_]\\w*(?=\\s*\\()";
+// int Function( or Namespace::FunctionName patterns
+const char* const kFunctionDefinitionRegex = "(?<=\\w)\\s+(\\w*::)*\\w*(?=\\()";
```
florian-kuebler, line 50: so you don't allow `_` in function names? Remember Google Style Guide as `snake_case()`/`set_snake_case(T t)` for simple accessor methods. If that is not easily doable, also feel free to ignore.

florian-kuebler (approved):
That's pretty nice! LGTM!

src/CodeViewerDemo/CodeExamples.h
karupayun: Doesn't matter, but I will change it.

src/CodeViewerDemo/CodeExamples.h
karupayun: I will polish a little bit CodeExamples to make code which have some sense.

src/SyntaxHighlighter/Cpp.cpp
beckerhe: `\w` includes a-z, A-Z, 0-9 ,and "_".

beckerhe: The last `\w*` should probably be `\w+`, no? And the one before `::` as well, isn't it?

src/SyntaxHighlighter/Cpp.cpp
```diff
@@ -53,35 +56,47 @@ const char* const kStringRegex = "\"([^\\\\\"]|\\\\.)*\"|\'[^\']*\'";
 const char* const kOpenStringRegex = "\"([^\\\\\"]|\\\\.)*\\\\$";
 const char* const kEndStringRegex = "([^\\\\\"]|\\\\.)*\"";
 const char* const kNoEndStringRegex = "([^\\\\\"]|\\\\.)*\\\\$";
-const char* const kCapitalizedRegex = "(?<=[\\s\\(<])[A-Z][\\w\\*\\&]*";
+const char* const kCapitalizedRegex = "(?<=[\\s\\(<])[A-Z][\\w]*";
 const char* const kNoLowercaseRegex = "(?<=[\\s\\(])[A-Z][0-9A-Z\\_\\*\\&]*(?=[\\;\\,\\(\\s\\)])";
+const char* const kCommaRegex = "[\\;\\,]";
+// All related with namespaces and classes (e.g Namespace:: , std::cout, absl::Milliseconds) except
+// function definitions
+const char* const kNamespaceRegex = "[A-Z]\\w*::|[a-z]\\w*::\\w*(?=\\()";
```
beckerhe, line 64: Why is there a differentiation between upper and lower case?

src/SyntaxHighlighter/Cpp.cpp
```diff
@@ -10,31 +10,34 @@
...
-    "struct|switch|typedef|union|unsigned|void|volatile|while|)\\b";
-// Match with comments starting and ending in the same line
+    "\\b(?:alignas|alignof|asm|auto|bool|break|case|catch|char|char8_t|char16_t|char32_t|class|"
+    "compl|concept|const|consteval|constexpr|constinit|const_cast|continue|co_await|co_return|co_"
+    "yield|decltype|default|delete|do|double|dynamic_cast|else|enum|explicit|export|extern|false|"
+    "final|float|for|friend|goto|if|inline|int|int8_t|int16_t|int32_t|int64_t|uint8_t|uint16_t|"
+    "uint32_t|uint64_t|long|mutable|namespace|new|noexcept|nullptr|operator|override|private|"
+    "protected|public|register|reinterpret_cast|requires|return|short|signed|sizeof|static|static_"
+    "assert|static_cast|struct|switch|template|this|thread_local|throw|true|try|typedef|typeid|"
+    "typename|union|unsigned|using|virtual|void|volatile|wchar_t|while)\\b";
```
beckerhe, line 40: Nice, we already support `concept` and `requires`. You could also add `import` and `export`.

src/SyntaxHighlighter/Cpp.cpp
```diff
@@ -95,33 +110,38 @@ void Cpp::highlightBlock(const QString& code) {
...
+  apply(&number_regex_, kBlue);
+  apply(&class_member_regex_, kViolet);
+  apply(&comma_regex_, kOrange);
 
-  // Special C/C++ words: The order here doesn't matter.
-  apply(&keyword_regex_, C::kOrange);
-  apply(&constant_regex_, C::kOlive);
-  apply(&include_file_regex_, C::kGreen);
-  apply(&preprocessor_regex_, C::kYellow);
+  // Special C/C++ words: The order here doesn't matter. Order doesn't matter.
```
beckerhe, line 125: Nit: comment. comment.

beckerhe (approved):
Very nice Pablo. Thank you a lot for that!

src/SyntaxHighlighter/Cpp.cpp
karupayun: \w includes '_' :)
You can see 2 examples there in the screenshots with functions having '_'.

I'm differentiating them using a horrible heuristic here. I'm asking for a character before the function(

In that way:
"int function();'' has the 't' of int
and "lets_execute_function()" not.

"(?<=\\w)\\s+" is doing that.
?<= means looking behind without painting it
and \\s+ is the spaces between that character and the function itself.

BTW, thanks for your comment, I've just realized that I have to polish this a little bit to avoid having number as a first character here. :)

src/SyntaxHighlighter/Cpp.cpp
karupayun: Ohh, sorry! :(

beckerhe: No worries. :slightly_smiling_face:

src/SyntaxHighlighter/Cpp.cpp
karupayun: Nice catch! I started making different things and after changed it and never review after that.

src/SyntaxHighlighter/Cpp.cpp
karupayun: Yes, thanks Henning

src/SyntaxHighlighter/Cpp.cpp
karupayun: I should rewrite this to include lower and uppercase together. Thanks for pointing out. It's something that I used to need, but not anymore, I think.

It isn't a difference between lower and upper, it's about the left and right. The first part is catching the left part of the " : : " and the second the right part which shouldn't be a function (because we are painting them yellow).

src/SyntaxHighlighter/Cpp.cpp
```diff
@@ -10,31 +10,34 @@
 #include <QTextCharFormat>
 
 namespace orbit_syntax_highlighter {
-namespace C {
 namespace {
-enum HighlighterState { kInitialState, kOpenCommentState, kOpenStringState };
-
 const QColor kGrey{0x80, 0x80, 0x80};
```
dpallotti, line 14: I would suggest a quick comment on where these colors are taken from (CLion Darcula?), for documentation purposes and in case one wants to add more.

karupayun: better?

src/SyntaxHighlighter/Cpp.cpp
karupayun: Changed: "(?<=\\w)\\s+([A-Za-z_]\\w*::)*[A-Za-z_]\\w*(?=\\()"

Instead of **_\w+_**, I used **_[A-Za-z_]\w*_**, because it shouldn't start with a number

src/SyntaxHighlighter/Cpp.cpp
karupayun: Divided into two:
kNamespaceRegex match the first part before "::" and kNamespaceVariableRegex match the right part. Turn out to be simpler after that division and cleaner.

src/SyntaxHighlighter/Cpp.cpp
dpallotti: Thanks (don't wait for my approval)

florian-kuebler (commented):
Thanks for the update. LGTM!

PR #1908, opened 2021-02-19
Extract useful functions to File.h

dimitry-:
Extract OpenForWriting/Reading and ReadFully from
ReadFileFromString and WriteFileToString.
Explicitly introduce unique_fd (unique_resource specialization).

Test: run OrbitBaseTests

### DIFF
--- a/src/OrbitBase/File.cpp
+++ b/src/OrbitBase/File.cpp
@@ -1,0 +1,64 @@
+// Copyright (c) 2021 The Orbit Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#include "OrbitBase/File.h"
+
+#include "OrbitBase/Logging.h"
+
+namespace orbit_base {
+
+static void FdDeleter(int fd) {
+  if (fd != -1) close(fd);
+};
+
+ErrorMessageOr<unique_fd> OpenFileForReading(const std::filesystem::path& path) {
+#if defined(__linux)
+  constexpr int open_flags = O_RDONLY | O_CLOEXEC;
+#elif defined(_WIN32)
+  constexpr int open_flags = O_RDONLY | O_BINARY;
+#endif  // defined(__linux)
+  int fd = TEMP_FAILURE_RETRY(open(path.string().c_str(), open_flags));
+  if (fd == -1) {
+    return ErrorMessage(
+        absl::StrFormat("Unable to open file \"%s\": %s", path.string(), SafeStrerror(errno)));
+  }
+
+  return unique_fd{fd, FdDeleter};
+}
+
+ErrorMessageOr<unique_fd> OpenFileForWriting(const std::filesystem::path& path) {
+#if defined(__linux)
+  constexpr int open_flags = O_WRONLY | O_CREAT | O_TRUNC | O_CLOEXEC;
+  constexpr int open_mode = 0600;
+#elif defined(_WIN32)
+  constexpr int open_flags = O_WRONLY | O_CREAT | O_TRUNC | O_BINARY;
+  constexpr int open_mode = _S_IREAD | _S_IWRITE;
+#endif  // defined(__linux)
+  int fd = TEMP_FAILURE_RETRY(open(path.string().c_str(), open_flags, open_mode));
+
+  if (fd == -1) {
+    return ErrorMessage(
+        absl::StrFormat("Unable to open file \"%s\": %s", path.string(), SafeStrerror(errno)));
+  }
+
+  return unique_fd{fd, FdDeleter};
+}
+
+ErrorMessageOr<void> WriteFully(const unique_fd& fd, std::string_view content) {
+  int64_t bytes_left = content.size();
+  const char* current_position = content.data();
+  while (bytes_left > 0) {
+    int64_t bytes_written = TEMP_FAILURE_RETRY(write(fd, current_position, bytes_left));
+    if (bytes_written == -1) {
+      return ErrorMessage{SafeStrerror(errno)};
+    }
+    current_position += bytes_written;
+    bytes_left -= bytes_written;
+  }
+
+  CHECK(bytes_left == 0);
+  return outcome::success();
+}
+
+}  // namespace orbit_base
--- a/src/OrbitBase/include/OrbitBase/File.h
+++ b/src/OrbitBase/include/OrbitBase/File.h
@@ -1,0 +1,43 @@
+// Copyright (c) 2021 The Orbit Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#ifndef ORBIT_BASE_FILE_H_
+#define ORBIT_BASE_FILE_H_
+
+#include <absl/strings/str_format.h>
+#include <fcntl.h>
+
+#include <filesystem>
+#include <string_view>
+
+#include "OrbitBase/Result.h"
+#include "OrbitBase/SafeStrerror.h"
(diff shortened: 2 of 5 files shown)


src/OrbitBase/include/OrbitBase/File.h
```diff
@@ -0,0 +1,43 @@
...
+#include <unistd.h>
+#elif defined(_WIN32)
+#include <io.h>
+// Windows does not have TEMP_FAILURE_RETRY - define a shortcut
+#define TEMP_FAILURE_RETRY(expression) (expression)
+#endif
+
+namespace orbit_base {
+
+using unique_fd = unique_resource<int, std::function<void(int)>>;
```
beckerhe, line 28: Since `FdDeleteter` is captureless, have you considered using a function pointer as the deleter?
That would be one indirection less and also takes less space.

`using unique_fd = unique_resource<int, void(*)(int)>;`

src/OrbitBase/include/OrbitBase/File.h
```diff
@@ -0,0 +1,43 @@
...
+#include "OrbitBase/SafeStrerror.h"
+#include "OrbitBase/UniqueResource.h"
+
+#if defined(__linux)
+#include <unistd.h>
+#elif defined(_WIN32)
+#include <io.h>
+// Windows does not have TEMP_FAILURE_RETRY - define a shortcut
+#define TEMP_FAILURE_RETRY(expression) (expression)
+#endif
```
beckerhe, line 24: Could this be moved to the CPP file? Or is there still a reason for the user to use TEMP_FAILURE_RETRY?

beckerhe (approved):
Otherwise LGTM! Thanks!

src/OrbitBase/File.cpp
```diff
@@ -0,0 +1,64 @@
...
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#include "OrbitBase/File.h"
+
+#include "OrbitBase/Logging.h"
+
+namespace orbit_base {
+
+static constexpr auto FdDeleter = [](int fd) {
```
dpallotti, line 11: Why not a function?

src/OrbitBase/include/OrbitBase/File.h
dpallotti, line 28: Why is snake_case being used for these types instead of PascalCase? It seems to be in clear constrast with the style guide.

src/OrbitBase/File.cpp
```diff
@@ -0,0 +1,64 @@
...
+
+  if (fd == -1) {
+    return ErrorMessage(
+        absl::StrFormat("Unable to open file \"%s\": %s", path.string(), SafeStrerror(errno)));
+  }
+
+  return unique_fd{fd, FdDeleter};
+}
+
+ErrorMessageOr<void> WriteFully(const unique_fd& fd, std::string_view content) {
```
dpallotti, line 48: Are you planning on adding a `ReadFully`?

src/OrbitBase/include/OrbitBase/File.h
```diff
@@ -0,0 +1,43 @@
...
+
+namespace orbit_base {
+
+using unique_fd = unique_resource<int, std::function<void(int)>>;
+
+// Following functions provides convenience wrappers for working with file
```
dpallotti, line 30:
```suggestion
// The following functions provide convenience wrappers for working with file
```

src/OrbitBase/include/OrbitBase/File.h
```diff
@@ -0,0 +1,43 @@
...
+// descriptors without having having to do TEMP_FAILURE_RETRY and
+// different subroutines for Windows/Linux. They also traslate errors to
```
dpallotti, line 32:
```suggestion
// different subroutines for Windows/Linux. They also translate errors to
```

src/OrbitBase/File.cpp
dpallotti, line 48: Don't you want to also support passing a bare file descriptor (`int`) instead of exclusively a `unique_fd`?

dpallotti (approved):

pierricgimmig (approved):

src/OrbitBase/File.cpp
dimitry-: yes - this one is on purpose, we should avoid using bare fds in our code.

src/OrbitBase/include/OrbitBase/File.h
dimitry-: It is currently used in ReadFileToString.cpp

src/OrbitBase/include/OrbitBase/File.h
dimitry-: done.

src/OrbitBase/File.cpp
dimitry-: done.

src/OrbitBase/include/OrbitBase/File.h
dimitry-: same reason why unique_resource is snake-case (as well as absl::bit_cast<>)

src/OrbitBase/File.cpp
dimitry-: Probably, I was thinking about ReadLine because this is the case we use in most places. But ReadFully would also make sense.

src/OrbitBase/include/OrbitBase/File.h
dimitry-: done.

src/OrbitBase/include/OrbitBase/File.h
dimitry-: done.

src/OrbitBase/File.cpp
dpallotti: Hhm, I don't have a strong opinion nor I see it being a problem, but conceptually I think this is the wrong place to enforce it.
At least from the name, `unique_fd` seems to serve the purpose of resource management of file descriptors, not the purpose of completely hiding the presence of file descriptors (which is something that `FILE` would do). 
Compare with `unique_ptr`: you want to use them, but it makes no sense for a function to accept `const unique_ptr&`, you could just pass the pointer. This is the same: you want to *create* `unique_fd`s, but it still makes sense to pass the bare file descriptor around.

src/OrbitBase/include/OrbitBase/File.h
dpallotti: And what's the reason why `unique_resource` is snake_case?

dimitry-: https://google.github.io/styleguide/cppguide.html#Exceptions_to_Naming_Rules (stl-like exception)

dpallotti: Alright.

src/OrbitBase/File.cpp
dimitry-: well, unique_ptr is not something we want to enforce because we do not want to decide for the user how they do memory management, is it unique_ptr, shared_ptr, weak_ptr or something else. For this we kind of want to not have bare fds in the code for files. I would default on unique_fd and if an reasonable use case requiring bare fd comes up we can change it.

dpallotti: Yes, we can always change it. One thing I was thinking is that in theory we could just make the fact that `unique_fd` is managing a file descriptor completely opaque. For example, it could just be `File` or `FileHandle`. It would fit better if this is meant to become a completely independent library to read and write files. I don't know if it's necessarily better though.

PR #1909, opened 2021-02-19
Clang-format

pierricgimmig:
@beckerhe , @florian-kuebler , if I understand correctly, our CI runs clang-format only on the lines that changed. Would it be possible to run it on the entire file instead so that we don't risk submitting a change that is not properly formatted in the context of the entire file?

### DIFF
--- a/src/CrashHandler/CrashHandler.cpp
+++ b/src/CrashHandler/CrashHandler.cpp
@@ -4,13 +4,13 @@
 
 #include "CrashHandler/CrashHandler.h"
 
-#include <algorithm>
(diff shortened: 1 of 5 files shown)


beckerhe (approved):
LGTM!

And regarding your question: Yes, it's possible to run clang-format for the entire file. But I don't think it will prevent these kind of errors.

I don't see right away why these slipped through. Could be caused by race-conditions (two PRs change the same files). I have to investigate more.

florian-kuebler (approved):
LGTM. Thanks!

PR #1910, opened 2021-02-20
Add "API" module which generates liborbit.so

pierricgimmig:
Add module that defines the Orbit programming interface and generates
the shared library that is dynamically loaded by user code in order
to interact programmatically with Orbit.

This the first part of a bigger change in the manual instrumentation API.

This PR copies src/Orbit.h into src/Api/include/Api/Orbit.h and reimplements the 
API as a shared library, liborbit.so, which is dynamically loaded on the first api 
function call. The new version will coexist with the old one until it is fully in place, 
at which time the old Orbit.h will disappear. For the review, I'd suggest comparing
both Orbit.h versions.

The main things to review are:
- "Api" project to generate liborbit.so
- Dynamic loading logic to src/Api/include/Api/Orbit.h (lines 305-440)

Code that was moved around:
- Existing EncodedEvent class from Orbit.h to EncodedEvent.h
- All of the Orbit.h api section (macros + function prototypes)

Note that the new Orbit.h API is now in "C". The plan is to be able to use it from both
C++ and C. In C++, the logic to dynamically load liborbit.so is directly in the header file
for ease of integration (a single .h file is all the user code needs). For a "C" project
however, the dynamic loading logic can't live in the header file, it will live in Orbit.c but this
is beyond the scope of this PR. See b/180976430.


src/Api/include/Api/Orbit.h
```diff
@@ -0,0 +1,440 @@
...
+// To disable manual instrumentation macros, define ORBIT_API_ENABLED as 0.
+#define ORBIT_API_ENABLED 1
+
+#if ORBIT_API_ENABLED
+
+// Call once at application start, this is only needed when using the Api from pure C code.
+#define ORBIT_API_INIT() orbit_api_init()
+
+// Call once at application exit.
+#define ORBIT_API_DEINIT() orbit_api_deinit()
```
dpallotti, line 60: What happens if this is not called? Isn't cleanup done anyway?

src/Api/include/Api/Orbit.h
```diff
@@ -0,0 +1,440 @@
...
+#else  // ORBIT_API_INTERNAL_IMPL
+
+#if __linux__
+
+#include <dlfcn.h>
+#include <stdio.h>
+
+inline void* orbit_api_get_lib_orbit() {
+  static void* liborbit = dlopen("./liborbit.so", RTLD_LAZY);
+  if (liborbit == nullptr) printf("%s", "ERROR. liborbit.so not found, Orbit API is disabled.\n");
```
dpallotti, line 314: I'm personally used to `ERROR:` instead of `ERROR.` from our logging (but feel free to keep as is).

dpallotti, line 314: Why the `"%s"` and then the string instead of just the string?

src/Api/include/Api/Orbit.h
```diff
@@ -0,0 +1,440 @@
...
+  return liborbit;
+}
+
+inline void* orbit_api_get_proc_address(const char* name) {
+  static void* liborbit = orbit_api_get_lib_orbit();
+  void* address = liborbit != nullptr ? dlsym(liborbit, name) : nullptr;
+  printf("orbit_api_get_proc_address for %s : %p\n", name, address);
```
dpallotti, line 321: I'd add quotes around `%s` and remove the space that follows it.

dpallotti, line 321: If I had to choose I would use `stderr` for this kind of logging, I generally feel that we shouldn't interfere with the `stdout` of a program, it could be used for some specific expected output (although this is likely irrelevant for games).

dpallotti (approved):
Looks good overall. Various minor comments and a couple of questions.

dpallotti:
>  For a "C" project
however, the dynamic loading logic can't live in the header file, it will live in Orbit.c but this
is beyond the scope of this PR. See b/180976430.

I remember this from back in the days of experimenting with Perfetto. Very annoying.

src/Api/Orbit.cpp
```diff
@@ -0,0 +1,78 @@
...
+
+// Temporary dummy function, proper implementation is in next PR.
+// See https://github.com/google/orbit/pull/1702 for fully working prototype.
+static void EnqueueIntermediateEvent(const orbit_api::ApiEvent& event) { (void)event; }
+
+static void EnqueueApiEvent(orbit_api::EventType type, const char* name = nullptr,
+                            uint64_t data = 0, orbit_api_color color = kOrbitColorAuto) {
+  uint64_t timestamp_ns = orbit_base::CaptureTimestampNs();
+  static pid_t pid = orbit_base::GetCurrentProcessId();
+  thread_local uint32_t tid = orbit_base::GetCurrentThreadId();
```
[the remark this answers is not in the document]
pierricgimmig: Done, thanks.

src/Api/Orbit.cpp
```diff
@@ -0,0 +1,78 @@
...
+
+  orbit_api::ApiEvent api_event(pid, tid, timestamp_ns, type, name, data, color);
+  EnqueueIntermediateEvent(api_event);
+}
+
+extern "C" {
+
+void orbit_api_start(const char* name, orbit_api_color color) {
+  EnqueueApiEvent(orbit_api::EventType::kScopeStart, name, /*data*/ 0, color);
```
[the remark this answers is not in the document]
pierricgimmig: Yes that's it, thanks, done.

src/Api/include/Api/EncodedEvent.h
```diff
@@ -0,0 +1,110 @@
...
+constexpr size_t kMaxEventStringSize = 34;
+struct Event {
+  uint8_t version;                 // 1
+  uint8_t type;                    // 1
+  char name[kMaxEventStringSize];  // 34
+  orbit_api_color color;           // 4
+  uint64_t data;                   // 8
+};
+
+// EncodedEvent is used for encoding an Orbit API event into the 6 integer register of the LInux
```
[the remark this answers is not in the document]
pierricgimmig: Done, thanks.

src/Api/include/Api/Orbit.h
```diff
@@ -0,0 +1,440 @@
...
+  return address;
+}
+
+#else
+
+inline void* orbit_api_get_proc_address(const char*) {
```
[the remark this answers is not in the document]
pierricgimmig: Done, thanks.

src/Api/include/Api/Orbit.h
```diff
@@ -0,0 +1,440 @@
...
+template <typename OrbitFunctionType>
+class OrbitFunctor {
+ public:
+  OrbitFunctor() = delete;
+  explicit OrbitFunctor(const char* proc_name)
+      : func_(reinterpret_cast<OrbitFunctionType>(orbit_api_get_proc_address(proc_name))) {}
+
+  template <typename... Args>
+  inline void operator()(const Args&... args) {
+    if (func_ != nullptr) func_(args...);
```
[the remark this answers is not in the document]
pierricgimmig: Yeah that's interesting. Nothing is moved so I didn't use std::forward, but I see your point. However, it seems forwarding all arguments at once doesn't work because of "const char*" in some of the functions that would have the "const" casted away. Not sure how to make this work actually, I'll leave it as is for now.

src/Api/include/Api/Orbit.h
pierricgimmig: Yeah looking at this again, it is incomplete. I actually wanted a way to let the user unload the shared object, but then it would break the program as the cached function addresses would point to freed memory. I opted to remove this altogether as it's not really needed to explicitly unload the library at the end of the program.

src/Api/include/Api/Orbit.h
pierricgimmig: Yep, changed it, thanks.

src/Api/include/Api/Orbit.h
pierricgimmig: Done, thanks.

src/Api/include/Api/Orbit.h
pierricgimmig: Good question, changed it :-)

src/Api/include/Api/Orbit.h
pierricgimmig: Changed it to stderr, it's also more consistent with our logging in general.

src/Api/include/Api/Orbit.h
```diff
@@ -0,0 +1,440 @@
...
+  }
+
+ private:
+  OrbitFunctionType func_;
+};
+
+extern "C" {
+
+// in C++, orbit_api_init() does nothing.
```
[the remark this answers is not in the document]
pierricgimmig: Done, thanks.

pierricgimmig:
Thanks for the review @dpallotti !

PR #1911, opened 2021-02-22
Fix incorrect notebook import for Python code

ronaldfw:
I renamed the Python file orbit_capture.py and forgot to update the
import in the Jupyter notebook (which is not unit tested). This change
fixes the import.


beckerhe (approved):
LGTM

ronaldfw:
Thanks for the quick review, @beckerhe

PR #1912, opened 2021-02-22
Measure and LOG clock resolution on start of OrbitService

ronaldfw:
We measure and LOG clock resolution on start of OrbitService to be able
to potentially debug issues encountered by users in the wild.

Bug: http://b/180720516

### DIFF
--- a/src/OrbitBase/Profiling.cpp
+++ b/src/OrbitBase/Profiling.cpp
@@ -1,0 +1,39 @@
+// Copyright (c) 2021 The Orbit Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#include "OrbitBase/Profiling.h"
+
+#include <algorithm>
+#include <limits>
+
+namespace orbit_base {
+uint64_t EstimateClockResolution() {
+  uint64_t minimum_resolution_found = std::numeric_limits<uint64_t>::max();
+
+  // We limit the amount of time spent to measure timer resolution.
+  uint64_t iteration_start_time_ns = orbit_base::CaptureTimestampNs();
+  uint64_t duration_ns = 0;
+  constexpr uint64_t kMaximumDurationNs = 1000 * 1000;  // 1 ms
+
+  while (duration_ns < kMaximumDurationNs) {
+    // Count to limit the number of iterations in the while loop below. Note that the loop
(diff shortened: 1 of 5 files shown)


src/Service/OrbitService.cpp
```diff
@@ -149,6 +159,8 @@ void OrbitService::Run(std::atomic<bool>* exit_requested) {
   LOG("**********************************");
 #endif
 
+  PrintClockResolution();
```
alenl, line 162: I would recommend checking that the resolution is some minimum sane resolution, and if it is not, at least warn the user. For example, if the resolution is as low as 1ms then you maybe don't want to even attempt profiling. 
(As it seems to has happened with introspection on Windows, which was the motivation for this change, if I understood correctly?)

src/OrbitBase/Profiling.cpp
```diff
@@ -0,0 +1,37 @@
...
+namespace orbit_base {
+int64_t EstimateClockResolution() {
+  int64_t minimum_resolution_found = std::numeric_limits<int64_t>::max();
+
+  // We take the minimum resolution found over 1000 tries. This should already yield stable
+  // results for far fewer iterations (less than 10), but we multiply by a factor of
+  // 100x for robustness. This is harmless in terms of runtime and may cover unforeseen
+  // cases that we can't test easily.
+  constexpr int kIterationsForMinimum = 1000;
+  for (int i = 0; i < kIterationsForMinimum; ++i) {
```
alenl, line 19: Perhaps consider using time, as an additional limit for this? Because e.g, if the resolution is 1ms, then this loop will add 1s to launch time of the service.

alenl (approved):
I have added comments about two suggestions - if you'd like, but even without it, it LGTM.

src/Service/OrbitService.cpp
ronaldfw: Good suggestion, and I think we should do that. We currently don't have a direct way of sending warnings/errors from the service to the UI to be displayed, so this needs to be discussed first how we  want to do that. As for now, I'll not address this in the current PR.

src/OrbitBase/Profiling.cpp
ronaldfw: Done.

ronaldfw:
Thanks for the review, @alenl

PR #1913, opened 2021-02-22
Add line highlighting support in CodeViewer

beckerhe:
This adds the possbility to highlight the current line. The current line
is the one with the cursor.

It's highlighted by making the background of that particular line lighter.

It works in conjunction with GoToLine.

There is also another commit adjusting the handling of line numbers. It seems
line numbers are always one-indexed, never zero-indexed. So let's change the API
to expect a one-indexed line number.

Source code view with line number 51 highlighted
Example: Line number 51 is highlighted here!

### DIFF
--- a/src/CodeViewer/Viewer.cpp
+++ b/src/CodeViewer/Viewer.cpp
@@ -137,10 +137,10 @@ void Viewer::UpdateLeftSidebarWidth() {
   left_sidebar_widget_.SetSizeHint({overall_width_px, 0});
 }
 
-void Viewer::SetEnableLineNumbers(bool enabled) {
-  if (line_numbers_enabled_ == enabled) return;
+void Viewer::SetEnableLineNumbers(bool is_enabled) {
+  if (line_numbers_enabled_ == is_enabled) return;
 
-  line_numbers_enabled_ = enabled;
+  line_numbers_enabled_ = is_enabled;
   UpdateLeftSidebarWidth();
 }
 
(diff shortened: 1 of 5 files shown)


src/CodeViewer/include/CodeViewer/Viewer.h
```diff
@@ -71,6 +74,9 @@ class Viewer : public QPlainTextEdit {
 
   FontSizeInEm heatmap_bar_width_ = FontSizeInEm{0.0f};
   HeatmapSource heatmap_source_;
+
+  bool highlight_current_line_ = false;
```
karupayun, line 78: not sure here, but is_current_line_highlighted_??

src/CodeViewer/include/CodeViewer/Viewer.h
```diff
@@ -71,6 +74,9 @@ class Viewer : public QPlainTextEdit {
...
+  void HighlightCurrentLine();
```
karupayun, line 79: Please move this private method near the others

src/CodeViewer/Viewer.cpp
```diff
@@ -165,4 +165,33 @@ void Viewer::ClearHeatmapSource() {
   UpdateLeftSidebarWidth();
 }
 
+void Viewer::SetHighlightCurrentLine(bool enabled) {
```
karupayun, line 168: bool is_enabled?

src/CodeViewer/Viewer.cpp
```diff
@@ -165,4 +165,33 @@ void Viewer::ClearHeatmapSource() {
...
+  highlight_current_line_ = enabled;
+
+  if (!highlight_current_line_) {
+    QObject::disconnect(this, &Viewer::cursorPositionChanged, this, &Viewer::HighlightCurrentLine);
+    return;
+  }
+
+  QObject::connect(this, &Viewer::cursorPositionChanged, this, &Viewer::HighlightCurrentLine,
+                   Qt::UniqueConnection);
+  HighlightCurrentLine();
```
karupayun, line 180: I don't understand why this function is only called once and not every time that we select a different line, but as it is working, I don't have anything to complain.

karupayun (approved):
LGTM Henning! Seems nice. I've left a few comments, feel free to disregard most of them :)

src/CodeViewer/include/CodeViewer/Viewer.h
beckerhe: Yes, good point. Will change it!

src/CodeViewer/include/CodeViewer/Viewer.h
beckerhe: Done.

src/CodeViewer/include/CodeViewer/Viewer.h
beckerhe: Done.

src/CodeViewer/Viewer.cpp
beckerhe: Well the `QObject::connect` ensure that it is called whenever the cursor position changes.

And then we just call it once to highlight the current line. Otherwise it would a cursor movement to start the highlighting.

src/CodeViewer/Viewer.cpp
beckerhe: Done.

src/CodeViewer/Viewer.cpp
karupayun: Nice!

PR #1914, opened 2021-02-22
Split module and symbole loading

beckerhe:
The function `OrbitApp::LoadModule` used to retrieve a module file and
then load its symbols automatically. This PR changes that by renaming
`LoadModule` to `LoadModuleAndSymbols`.

Furthermore it introduces a new function called `OrbitApp::LoadModule`
which only retrieves the module and ensure that it is available in a
local file path. Maybe nothing needs to be done, because it's in a local
search path, or it needs to be copied from the remote instance and put
in the local cache directory. Either way, `LoadModule` will take care of
it and return the local path, if successful.

Test: Manual test
Bug: http://b/179241832

### DIFF
--- a/src/OrbitGl/App.cpp
+++ b/src/OrbitGl/App.cpp
@@ -968,7 +968,7 @@ void OrbitApp::SendErrorToUi(const std::string& title, const std::string& text) 
   });
 }
 
-orbit_base::Future<ErrorMessageOr<std::filesystem::path>> OrbitApp::LoadModuleOnRemote(
+orbit_base::Future<ErrorMessageOr<std::filesystem::path>> OrbitApp::RetrieveModuleFromRemote(
     const std::string& module_file_path) {
   ScopedStatus scoped_status = CreateScopedStatus(absl::StrFormat(
       "Searching for symbols on remote instance for module \"%s\"...", module_file_path));
@@ -1006,12 +1006,12 @@ orbit_base::Future<ErrorMessageOr<std::filesystem::path>> OrbitApp::LoadModuleOn
   return check_file_on_remote.Then(main_thread_executor_, std::move(download_file));
 }
 
-void OrbitApp::LoadModuleOnRemote(ModuleData* module_data,
-                                  std::vector<uint64_t> function_hashes_to_hook,
-                                  std::vector<uint64_t> frame_track_function_hashes,
-                                  std::string error_message_from_local) {
+void OrbitApp::RetrieveModuleFromRemote(ModuleData* module_data,
+                                        std::vector<uint64_t> function_hashes_to_hook,
+                                        std::vector<uint64_t> frame_track_function_hashes,
+                                        std::string error_message_from_local) {
   // This overload will be removed in a subsequent commit.
-  auto local_file_path = LoadModuleOnRemote(module_data->file_path());
+  auto local_file_path = RetrieveModuleFromRemote(module_data->file_path());
 
   auto load_symbols = [this, module_data,
                        function_hashes_to_hook = std::move(function_hashes_to_hook),
@@ -1058,12 +1058,13 @@ void OrbitApp::LoadModules(
           SelectFunctionsFromHashes(module, select_functions);
           EnableFrameTracksFromHashes(module, frame_tracks);
         };
-    LoadModuleAndSymbols(module).Then(main_thread_executor_,
-                                      std::move(handle_hooks_and_frame_tracks));
+    RetrieveModuleAndLoadSymbols(module).Then(main_thread_executor_,
+                                              std::move(handle_hooks_and_frame_tracks));
   }
 }
 
-orbit_base::Future<void> OrbitApp::LoadModules(absl::Span<const ModuleData* const> modules) {
+orbit_base::Future<void> OrbitApp::RetrieveModulesAndLoadSymbols(
+    absl::Span<const ModuleData* const> modules) {
   std::vector<orbit_base::Future<void>> futures;
   futures.reserve(modules.size());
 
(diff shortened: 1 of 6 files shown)


src/OrbitGl/App.cpp
```diff
@@ -1074,64 +1075,81 @@ orbit_base::Future<void> OrbitApp::LoadModules(absl::Span<const ModuleData* cons
...
-                          const ErrorMessageOr<std::filesystem::path>& result)
-      -> orbit_base::Future<ErrorMessageOr<void>> {
-    if (result.has_error()) {
-      return {ErrorMessage{
-          absl::StrFormat("Did not find symbols locally or on remote for module \"%s\": %s\n%s",
-                          module_path, local_error_msg, result.error().message())}};
-    }
-
-    return LoadSymbols(result.value(), module_path);
-  };
```
beckerhe, line 1124: This move upwards into `LoadModuleAndSymbols`. And it also gets simpler due to `ThenIfSuccess`

src/OrbitGl/App.cpp
```diff
@@ -1074,64 +1075,81 @@ orbit_base::Future<void> OrbitApp::LoadModules(absl::Span<const ModuleData* cons
...
 
   return orbit_base::JoinFutures(futures);
 }
 
-orbit_base::Future<ErrorMessageOr<void>> OrbitApp::LoadModule(const ModuleData* module) {
+orbit_base::Future<ErrorMessageOr<void>> OrbitApp::LoadModuleAndSymbols(const ModuleData* module) {
+  return LoadModuleAndSymbols(module->file_path(), module->build_id());
+}
+
+orbit_base::Future<ErrorMessageOr<std::filesystem::path>> OrbitApp::LoadModule(
```
dpallotti, line 1089: Shouldn't this be moved next to the other overload, that this method calls?

src/OrbitGl/App.h
```diff
@@ -289,11 +289,20 @@ class OrbitApp final : public DataViewFactory, public CaptureListener {
...
       absl::flat_hash_map<std::string, std::vector<uint64_t>> frame_track_function_hashes_map);
-  orbit_base::Future<ErrorMessageOr<void>> LoadModule(const ModuleData* module);
-  orbit_base::Future<ErrorMessageOr<void>> LoadModule(const std::string& module_path,
-                                                      const std::string& build_id);
+
+  // LoadModule retrieves a module file and returns the local file path (potentially from the local
+  // cache). Symbols won't be loaded.
+  orbit_base::Future<ErrorMessageOr<std::filesystem::path>> LoadModule(const ModuleData* module);
+  orbit_base::Future<ErrorMessageOr<std::filesystem::path>> LoadModule(
+      const std::string& module_path, const std::string& build_id);
```
dpallotti, line 297: The comments certainly clarify it, but from the name of the methods the word "load" might be bit overloaded. What do you think of `RetrieveModule` here and `RetrieveModuleAndLoadSymbols` below?

src/OrbitGl/App.h
```diff
@@ -289,11 +289,20 @@ class OrbitApp final : public DataViewFactory, public CaptureListener {
...
   orbit_base::Future<void> LoadModules(absl::Span<const ModuleData* const> modules);
 
+  // LoadModuleAndSymbols is a helper function which first retrieves the module by calling
+  // `LoadModule` and afterwards load the symbols by calling `LoadSymbols`.
+  orbit_base::Future<ErrorMessageOr<void>> LoadModuleAndSymbols(const ModuleData* module);
+  orbit_base::Future<ErrorMessageOr<void>> LoadModuleAndSymbols(const std::string& module_path,
+                                                                const std::string& build_id);
```
antonrohr, line 304: Is this callback making sense? It is only called from `LoadModuleAndSymbols(const ModuleData* module)`. I am mainly curious whether you designed it like this on purpose

antonrohr (commented):
I agree with @dpallotti that the naming could be made better. I also think this PR is a good place for that. 
I like the `Retrieve` idea. Or maybe something with `Find`, since some kind of searching/finding is involved. 
Also, since its technically not the "Module" that is retrieved/loaded/found/whatever, but the file that contains the debug information / symbols: maybe `RetrieveSymbolsFile(ModuleData* module)`. 

And then maybe instead of `LoadModuleAndSymbol`: `LoadSymbolsIntoModule()`. 

I dont have strong preferences, these are just ideas

src/OrbitGl/App.cpp
```diff
@@ -1074,64 +1075,81 @@ orbit_base::Future<void> OrbitApp::LoadModules(absl::Span<const ModuleData* cons
...
+  if (module_data->is_loaded()) {
+    return {symbol_helper_.GenerateCachedFileName(std::filesystem::path{module_path})};
+  }
 
   const auto it = modules_currently_loading_.find(module_path);
   if (it != modules_currently_loading_.end()) {
     return it->second;
   }
 
-  const auto local_symbols_path = FindModuleLocally(module_path, build_id);
```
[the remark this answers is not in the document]
beckerhe: cause of RVO

src/OrbitGl/App.h
beckerhe: I guess you mean overload, do you? Why do you think it doesn't make sense? You can call it whenever you don't have a ModuleData object. You actually don't need a full ModuleData object, a path and a build_id is fine.

src/OrbitGl/App.cpp
```diff
@@ -1074,64 +1075,81 @@ orbit_base::Future<void> OrbitApp::LoadModules(absl::Span<const ModuleData* cons
...
   const ModuleData* module_data = GetModuleByPath(module_path);
 
   if (module_data == nullptr) {
     return {ErrorMessage{absl::StrFormat("Module \"%s\" was not found", module_path)}};
   }
 
-  if (module_data->is_loaded()) return {outcome::success()};
+  if (module_data->is_loaded()) {
+    return {symbol_helper_.GenerateCachedFileName(std::filesystem::path{module_path})};
+  }
```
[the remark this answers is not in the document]
beckerhe: Ah yes. I missed to remove that check entirely. This actually moved into `LoadModuleAndSymbols`

src/OrbitGl/App.h
antonrohr: Yes I meant overload instead of callback. 
> You actually don't need a full ModuleData object, a path and a build_id is fine.

But there currently does not seem to be a scenario where only path and build_id is available, but not the full ModuleData object. Are you anticipating there will be a scenario in the future, or is there any other reason? 

I could imagine that this is good design in the sense of that `LoadModuleAndSymbols(path, build_id)` is the minimal function that is maybe more easily testable or other better for other reasons? And `LoadModuleAndSymbols(module_data)` is the convenience overload. Is this your thinking behind it, or something else?

src/OrbitGl/App.cpp

src/OrbitGl/App.h
beckerhe: Yes, that's a very good point. I also thought about this already and I like "Retrieve". Changed.

src/OrbitGl/App.cpp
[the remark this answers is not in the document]
beckerhe: Done. Thanks.

src/OrbitGl/App.h
beckerhe: > Yes I meant overload instead of callback.
> 
> > You actually don't need a full ModuleData object, a path and a build_id is fine.
> 
> But there currently does not seem to be a scenario where only path and build_id is available, but not the full ModuleData object. Are you anticipating there will be a scenario in the future, or is there any other reason?
> 
> I could imagine that this is good design in the sense of that `LoadModuleAndSymbols(path, build_id)` is the minimal function that is maybe more easily testable or other better for other reasons? And `LoadModuleAndSymbols(module_data)` is the convenience overload. Is this your thinking behind it, or something else?

Yes, that's exactly my reasoning here. I try to explicitly state the dependencies and keep them minimal. That helps with decoupling in the long run as well. Do you disagree?

beckerhe:
@antonrohr Are you going to have another look?

beckerhe:
> @antonrohr Are you going to have another look?

I'm going to merge this in half an hour.

PR #1915, opened 2021-02-22
Fix crash on debug markers across multiple submissions.

florian-kuebler:
If debug markers are across different submissions, and we captured
the "begin" marker in the Vulkan layer, but missed a tracepoint
for the "GpuJobs, we end up in a situation where we do have
the "begin" submission, but not the "begin" job.

We now handle this gracefully (similar to the case where we do not
have the "begin" submission), but using the begin capture time
as approximation.

Test: Use Vulkan layer on Infiltrator.

### DIFF
--- a/src/OrbitCaptureClient/CaptureEventProcessorTest.cpp
+++ b/src/OrbitCaptureClient/CaptureEventProcessorTest.cpp
@@ -892,8 +892,6 @@ TEST(CaptureEventProcessor, CanHandleGpuDebugMarkersWithNoBeginJobRecorded) {
 
   event_processor.ProcessEvent(queue_submission_event_1);
 
-  testing::Mock::VerifyAndClearExpectations(&listener);
-
   uint64_t actual_command_buffer_key;
   EXPECT_CALL(listener, OnKeyAndString(_, "command buffer"))
       .Times(1)
--- a/src/OrbitCaptureClient/CaptureEventProcessorTest.cpp
+++ b/src/OrbitCaptureClient/CaptureEventProcessorTest.cpp
@@ -852,6 +852,81 @@ TEST(CaptureEventProcessor, CanHandleGpuDebugMarkersWithNoBeginRecorded) {
                            actual_marker_key);
 }
 
+TEST(CaptureEventProcessor, CanHandleGpuDebugMarkersWithNoBeginJobRecorded) {
+  MockCaptureListener listener;
+  CaptureEventProcessor event_processor(&listener);
+
+  ClientCaptureEvent timeline_string = CreateInternedStringEvent(kTimelineKey, kTimelineString);
+
+  ClientCaptureEvent gpu_job_event_2;
+  GpuJob* gpu_job_2 = CreateGpuJob(&gpu_job_event_2, kTimelineKey, 50, 60, 70, 80);
+
+  ClientCaptureEvent marker_string_event = CreateInternedStringEvent(42, "marker");
+
+  ClientCaptureEvent queue_submission_event_1;
+  GpuQueueSubmission* submission_1 = queue_submission_event_1.mutable_gpu_queue_submission();
+  GpuQueueSubmissionMetaInfo* meta_info_1 = CreateGpuQueueSubmissionMetaInfo(submission_1, 9, 11);
+  GpuSubmitInfo* submit_info_1 = submission_1->add_submit_infos();
+  AddGpuCommandBufferToGpuSubmitInfo(submit_info_1, 115, 119);
+  AddGpuCommandBufferToGpuSubmitInfo(submit_info_1, 120, 124);
+  submission_1->set_num_begin_markers(1);
+
+  ClientCaptureEvent queue_submission_event_2;
+  GpuQueueSubmission* submission_2 = queue_submission_event_2.mutable_gpu_queue_submission();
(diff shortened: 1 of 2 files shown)


src/OrbitCaptureClient/GpuQueueSubmissionProcessor.cpp
```diff
@@ -348,25 +350,31 @@ std::vector<TimerInfo> GpuQueueSubmissionProcessor::ProcessGpuDebugMarkers(
...
-
-      // Convert the GPU time to CPU time, based on the CPU time of the HW execution begin and the
-      // GPU timestamp of the begin of the first command buffer. Note that we will assume that the
-      // first command buffer starts execution right away as an approximation.
-      marker_timer.set_start(completed_marker.begin_marker().gpu_timestamp_ns() +
-                             matching_begin_job->gpu_hardware_start_time_ns() -
-                             begin_submission_first_command_buffer->begin_gpu_timestamp_ns());
+
+      uint64_t begin_submission_time_ns = 0;
+      // We might have bad luck and captured the "begin" submission, but not the matching job.
```
ronaldfw, line 355: This reads as if this comment applies to the "if" case, but the "bad luck" part is actually handled in the "else" clause. Move this comment into the "else" block?

src/OrbitCaptureClient/GpuQueueSubmissionProcessor.cpp
```diff
@@ -212,7 +211,10 @@ void GpuQueueSubmissionProcessor::DeleteSavedGpuJob(int32_t thread_id,
   if (!tid_to_submission_time_to_gpu_job_.contains(thread_id)) {
     return;
   }
-  auto& submission_time_to_gpu_job = tid_to_submission_time_to_gpu_job_.at(thread_id);
+  // This method might be also called if the the last Gpu Job that was was not part of the capture,
```
ronaldfw, line 214:
```suggestion
  // This method might be also called if the last Gpu Job that was not part of the capture,
```

ronaldfw: This sentence also seems incomplete? Maybe the "that" shouldn't be there?

src/OrbitCaptureClient/GpuQueueSubmissionProcessor.cpp
dpallotti, line 355: I'd say "We might have had bad luck and have captured"

src/OrbitCaptureClient/GpuQueueSubmissionProcessor.cpp
dpallotti: Yep, I'm a bit confused by this sentence.

src/OrbitCaptureClient/CaptureEventProcessorTest.cpp
```diff
@@ -852,6 +852,81 @@ TEST(CaptureEventProcessor, CanHandleGpuDebugMarkersWithNoBeginRecorded) {
...
+  EXPECT_CALL(listener, OnKeyAndString(_, "hw execution")).Times(1);
+
+  EXPECT_CALL(listener, OnTimer).Times(3);
+
+  event_processor.ProcessEvent(timeline_string);
+  event_processor.ProcessEvent(gpu_job_event_2);
+
+  testing::Mock::VerifyAndClearExpectations(&listener);
+
+  event_processor.ProcessEvent(queue_submission_event_1);
```
dpallotti, line 893: Is it on purpose that no expectations are set for this but `VerifyAndClearExpectations` is called right after? You just want to ignore the calls to the mock that this causes?

src/OrbitCaptureClient/GpuQueueSubmissionProcessor.cpp
florian-kuebler: You are right. That sentence is completely broken. I've re-written it.

src/OrbitCaptureClient/GpuQueueSubmissionProcessor.cpp
florian-kuebler: Done.

florian-kuebler: Done.

florian-kuebler (commented):
Thanks for the review comments. I addressed the ones so far.

src/OrbitCaptureClient/GpuQueueSubmissionProcessor.cpp
```diff
@@ -212,7 +211,10 @@ void GpuQueueSubmissionProcessor::DeleteSavedGpuJob(int32_t thread_id,
   if (!tid_to_submission_time_to_gpu_job_.contains(thread_id)) {
     return;
   }
-  auto& submission_time_to_gpu_job = tid_to_submission_time_to_gpu_job_.at(thread_id);
+  // This method might be called if the "capture start" falls directly inside a GpuJob, and we thus
```
dpallotti, line 214: I think "if" is slightly inaccurate (that's not the only case when this method is called). Maybe "even when" would be more precise?

dpallotti (approved):

src/OrbitCaptureClient/GpuQueueSubmissionProcessor.cpp
florian-kuebler: Done.

PR #1916, opened 2021-02-22
Add ElfFile::CalculateDebuglinkChecksum

beckerhe:
Adds a function which allows to verify the CRC32 checksum as given by a
`.gnu_debuglink` section.

### DIFF
--- a/src/ElfUtils/ElfFile.cpp
+++ b/src/ElfUtils/ElfFile.cpp
@@ -13,6 +13,7 @@
 #include <llvm/Object/ELF.h>
 #include <llvm/Object/ELFTypes.h>
 #include <llvm/Object/SymbolicFile.h>
+#include <llvm/Support/CRC.h>
 #include <llvm/Support/Casting.h>
 #include <llvm/Support/Error.h>
 #include <llvm/Support/MathExtras.h>
@@ -22,7 +23,10 @@
 #include <utility>
 #include <vector>
 
+#include "OrbitBase/File.h"
 #include "OrbitBase/Logging.h"
+#include "OrbitBase/ReadFileToString.h"
+#include "OrbitBase/Result.h"
 #include "absl/strings/str_cat.h"
 #include "absl/strings/str_format.h"
 #include "llvm/DebugInfo/Symbolize/Symbolize.h"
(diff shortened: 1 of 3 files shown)


florian-kuebler (approved):
LGTM!

src/ElfUtils/ElfFile.cpp
```diff
@@ -385,4 +388,11 @@ ErrorMessageOr<std::unique_ptr<ElfFile>> ElfFile::Create(
   return result;
 }
 
+ErrorMessageOr<uint32_t> ElfFile::CalculateDebuglinkChecksum(
+    const std::filesystem::path& file_path) {
+  ErrorMessageOr<std::string> buffer_or_error = orbit_base::ReadFileToString(file_path);
```
dimitry-, line 393: as far as I understand this file could be quite big (it most likely has .debug_info section)... maybe we will need to change to use iterative crc32 calculation instead?

dimitry-: (add a todo?)

beckerhe: Yes, very good point. As discussed OOB, I will add a TODO and change that to an interative approach when `orbit_base::ReadFully` is implemented in a couple of days.

beckerhe: Added the TODO

beckerhe: Since `ReadFully` is already implemented, I will wait until it's merged.

beckerhe: @dimitry- I changed the implementation. PTAL!

src/ElfUtils/ElfFile.cpp
```diff
@@ -385,4 +389,33 @@ ErrorMessageOr<std::unique_ptr<ElfFile>> ElfFile::Create(
...
+  std::vector<char> buffer(kBufferSize);
+
+  uint32_t rolling_checksum = 0;
+
+  while (true) {
+    ErrorMessageOr<size_t> chunksize_or_error =
+        orbit_base::ReadFully(fd_or_error.value(), buffer.data(), buffer.size());
+    if (chunksize_or_error.has_error()) return chunksize_or_error.error();
+
+    const llvm::StringRef str{buffer.data(), chunksize_or_error.value()};
```
dimitry-, line 412: chunksize_or_error could be 0 in case file size is divisible by 4MiB (not sure if StringRef handles this correctly), maybe just check for 0 and break here instead of doing the check on L415. We can afford extra read once we hit the end of file.

dimitry- (approved):
otherwise LGTM

src/ElfUtils/ElfFile.cpp
beckerhe: done.

PR #1918, opened 2021-02-23
Collect total memory usage information

vickyliu-go4it:
With this change, Orbit is able to periodically collect the total memory usage information, which includes the total memory, free memory, available memory, memory of buffers and cached memory retrieved from the `/proc/meminfo` file.

Bugs: http://b/179000525, http://b/179000844.
Test: Tested by logging out the received memory events in `CaptureEventProcesser::ProcessEvent`.


src/GrpcProtos/capture.proto
```diff
@@ -271,6 +271,16 @@ message ModuleUpdateEvent {
   ModuleInfo module = 3;
 }
 
+message TotalMemoryUsageInfo {
```
dimitry-, line 277: I would also add a pid here, and populate it.

src/GrpcProtos/capture.proto
```diff
@@ -48,6 +48,9 @@ message CaptureOptions {
   bool enable_introspection = 9;
 
   uint64 max_local_marker_depth_per_command_buffer = 10;
+
+  bool collect_memory_info = 11;
+  int64 memory_sampling_period_ns = 12;
```
dimitry-, line 53: unit64?

src/Memory/CMakeLists.txt
```diff
@@ -0,0 +1,43 @@
...
+target_link_libraries(Memory PUBLIC
+        GrpcProtos
+        OrbitBase
+        CONAN_PKG::abseil)
+
+add_executable(MemoryTests)
+
+target_compile_options(MemoryTests PRIVATE ${STRICT_COMPILE_FLAGS})
+
+target_sources(MemoryTests PRIVATE MemoryUtilsTest.cpp)
```
dimitry-, line 36: The naming conventions is to get a header file and have a test named after it:
MemoryInfoListener.h -> MemoryInfoListenerTests.cpp

src/Memory/MemoryUtils.cpp
```diff
@@ -0,0 +1,67 @@
...
+  constexpr int kBase = 10;
+  constexpr int kMissingInfo = -1;
+  int mem_total = kMissingInfo, mem_free = kMissingInfo, mem_available = kMissingInfo,
+      mem_buffers = kMissingInfo, mem_cached = kMissingInfo;
+  for (int i = 0, end = lines.size() > kNumLines ? kNumLines : lines.size(); i < end; ++i) {
+    const std::string& current_line = lines[i];
+    std::vector<std::string> splits = absl::StrSplit(current_line, ' ', absl::SkipWhitespace{});
+    if (splits.size() < 2) continue;
+    if (splits[0] == "MemTotal:") {
+      mem_total = std::strtoll(splits[1].c_str(), nullptr, kBase);
```
dimitry-, line 34: absl::SimpleAtoi?

src/Memory/include/Memory/MemoryInfoProducer.h
```diff
@@ -0,0 +1,45 @@
...
+  void SetListener(MemoryInfoListener* listener) { listener_ = listener; }
+
+  void Start();
+  void Stop();
+
+ private:
+  void Run();
+  void ProduceTotalMemoryUsageInfoAndSendToListener();
+
+  bool collect_memory_info_;
```
dimitry-, line 37: maybe avoid starting this producer if this option is set to false instead of checking it when we start the thread?

src/Memory/include/Memory/MemoryInfoProducer.h
```diff
@@ -0,0 +1,45 @@
...
+
+#include "Memory/MemoryInfoListener.h"
+#include "capture.pb.h"
+
+namespace orbit_memory {
+class MemoryInfoProducer {
+ public:
+  explicit MemoryInfoProducer(const orbit_grpc_protos::CaptureOptions& capture_options)
+      : collect_memory_info_(capture_options.collect_memory_info()),
+        sampling_period_ns_(capture_options.memory_sampling_period_ns()) {}
```
dimitry-, line 19: also probably just pass memory_sampling_period_ns here, it does not seem that MemoryInfoProcuder needs anything else from that message..

src/GrpcProtos/capture.proto
```diff
@@ -300,7 +300,8 @@ message ClientCaptureEvent {
     SchedulingSlice scheduling_slice = 11;
     ThreadName thread_name = 12;
     ThreadStateSlice thread_state_slice = 13;
-    TracepointEvent tracepoint_event = 14;
+    TotalMemoryUsageInfo total_memory_usage_info = 14;
+    TracepointEvent tracepoint_event = 15;
```
dimitry-, line 304: when adding new field to proto message do not change field number for existing ones. They are needed for backwards compatibility.

src/Service/MemoryInfoHandler.cpp
```diff
@@ -0,0 +1,34 @@
...
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#include "MemoryInfoHandler.h"
+
+#include "OrbitBase/Logging.h"
+
+namespace orbit_service {
+
+constexpr uint64_t kMemoryInfoProducerId = 1;
```
dimitry-, line 11: we probably want one file with all reserved ids moved to somewhere more visible. Maybe to GrpcProtos/include/...

also move the starting id in ProducerSideServiceImpl.h to kExtranalProducerStartingId and set it to 1024 (or some other number >> 1)

dimitry- (commented):
Hi, I left some preliminary comments, I will be doing another pass on this change tomorrow.

src/GrpcProtos/capture.proto
vickyliu-go4it: Added a `pid` to `TotalMemoryUsageInfo` and also updated `MemoryUtils::GetTotalMemoryUsageInfo` to set up the pid.
Thanks for the suggestion.

src/GrpcProtos/capture.proto
vickyliu-go4it: Change accordingly. Thanks so much for finding this!

src/Memory/MemoryUtils.cpp
vickyliu-go4it: Changed to use `absl::SimpleAtoi`. Thanks for telling me about this method.

src/Memory/include/Memory/MemoryInfoProducer.h
vickyliu-go4it: Done. Now I let the `MemoryInfoHandler` to check `CaptureOptions::collect_memory_info_` and decide whether to start the producer. Thanks!

src/Memory/include/Memory/MemoryInfoProducer.h
vickyliu-go4it: Changed accordingly. Thanks for the suggestion : )

src/GrpcProtos/capture.proto
vickyliu-go4it: Thanks for telling me about this! Changed accordingly.

src/Service/MemoryInfoHandler.cpp
vickyliu-go4it: Make sense. 
I moved all reserved producer IDs to GrpcProtos/constants.h, including:
* `kLinuxTracingProducerId` used in LinuxTracingHandler.cpp
* `kMemoryInfoProducerId` used in MemoryInfoHandler.cpp
* `kExternalProducerStartingId` used as the starting id in ProducerSideServiceImpl.h

Thanks for the suggestion.

src/Memory/MemoryUtils.cpp
```diff
@@ -0,0 +1,69 @@
...
+    } else if (splits[0] == "Cached:" && absl::SimpleAtoi(splits[1], &mem_cached)) {
+      memory_info.set_cached_kb(mem_cached);
+    }
+  }
+
+  return memory_info;
+}
+
+std::optional<TotalMemoryUsageInfo> GetTotalMemoryUsageInfo(uint64_t timestamp_ns) noexcept {
+  ErrorMessageOr<std::string> reading_result = orbit_base::ReadFileToString("/proc/meminfo");
```
dimitry-, line 57: I just realized this information is not process specific, it is system-wide. is this intentional?

dimitry-: (in which case pid does not make sense in proto, maybe rename proto to SystemMemoryUsage?)

vickyliu-go4it: > I just realized this information is not process specific, it is system-wide. is this intentional?

Yes. We consider providing the system wide memory usage information at the first stage. 

Sorry about my misleading naming. I'm really not good at it :-)

vickyliu-go4it: > (in which case pid does not make sense in proto, maybe rename proto to SystemMemoryUsage?)

Thanks for suggesting the name ` SystemMemoryUsage`!

src/Memory/MemoryUtilsTest.cpp
```diff
@@ -0,0 +1,110 @@
...
+MemFree:        $1 kB
+SwapCached:      0 kB)",
+                                                       kMemTotal, kMemFree);
+
+  const std::string kEmptyMeminfo = "";
+
+  std::optional<TotalMemoryUsageInfo> parsing_result;
+
+  // TODO(vickyliu): Check whether we are able to user testing::EqualsProto from the #include
+  // "testing/base/public/gmock.h" library
```
dpallotti, line 89: I would remove this TODO, this is a problem that comes up over and over and would apply to many tests. Maybe create a bug instead.

src/Memory/MemoryInfoProducer.cpp
```diff
@@ -0,0 +1,52 @@
...
+void MemoryInfoProducer::Stop() {
+  *exit_requested_ = true;
+  if (thread_ != nullptr && thread_->joinable()) {
+    thread_->join();
+  }
+  thread_.reset();
+}
+
+void MemoryInfoProducer::Run() {
+  pthread_setname_np(pthread_self(), "MemoryInfoProducer::Run");
```
dpallotti, line 33: The thread name will not be set as it exceeds 15 characters.

src/Memory/include/Memory/MemoryInfoProducer.h
```diff
@@ -0,0 +1,43 @@
...
+  void Start();
+  void Stop();
+
+ private:
+  void Run();
+  void ProduceTotalMemoryUsageInfoAndSendToListener();
+
+  uint64_t sampling_period_ns_;
+  MemoryInfoListener* listener_ = nullptr;
+  std::shared_ptr<std::atomic<bool>> exit_requested_ = std::make_unique<std::atomic<bool>>(true);
```
dpallotti, line 38: Do you need this to be wrapped inside a smart pointer?

src/Memory/include/Memory/MemoryInfoProducer.h
```diff
@@ -0,0 +1,43 @@
...
+  std::shared_ptr<std::thread> thread_;
```
dpallotti, line 39: Do you need this to be wrapped inside a smart pointer?

src/Memory/include/Memory/MemoryInfoProducer.h
```diff
@@ -0,0 +1,43 @@
+// Copyright (c) 2021 The Orbit Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#ifndef ORBIT_MEMORY_INFO_PRODUCER_H_
```
dpallotti, line 5: Include guard should be the name of the module followed by the name of the file.

src/Memory/MemoryInfoProducer.cpp
```diff
@@ -0,0 +1,52 @@
...
+void MemoryInfoProducer::ProduceTotalMemoryUsageInfoAndSendToListener() {
+  std::optional<TotalMemoryUsageInfo> total_memory_usage_info;
+  uint64_t current_timestamp_ns = orbit_base::CaptureTimestampNs();
+  while (!(*exit_requested_)) {
+    total_memory_usage_info = GetTotalMemoryUsageInfo(current_timestamp_ns);
+    if (total_memory_usage_info.has_value()) {
+      listener_->OnTotalMemoryUsageInfo(total_memory_usage_info.value());
+    }
+
+    std::this_thread::sleep_for(std::chrono::nanoseconds(sampling_period_ns_));
```
dpallotti, line 48: I think that the sampling period should be handled differently. Only sleeping for `sampling_period_ns_` doesn't account for how long the sleep will actually be (this value passed is just a lower bound) and for how long the rest of the code in the `while` will take. This means that over time the sampling period will be higher than the one passed. An approach similar to what games do to compute how much time is left before the next frame should probably be used.

src/Service/MemoryInfoHandler.h
```diff
@@ -0,0 +1,37 @@
...
+#include "Memory/MemoryInfoProducer.h"
+#include "ProducerEventProcessor.h"
+#include "capture.pb.h"
+
+namespace orbit_service {
+
+class MemoryInfoHandler : public orbit_memory::MemoryInfoListener {
+ public:
+  explicit MemoryInfoHandler(ProducerEventProcessor* producer_event_processor)
+      : producer_event_processor_{producer_event_processor} {}
```
dpallotti, line 17: Maybe check that `producer_event_processor` is not `nullptr`?

src/GrpcProtos/capture.proto
```diff
@@ -271,6 +274,17 @@ message ModuleUpdateEvent {
...
 
+message TotalMemoryUsageInfo {
+  int32 pid = 1;
+  uint64 timestamp_ns = 2;
+  // These fields are retrieved from /proc/meminfo.
+  int64 total_kb = 3;
+  int64 free_kb = 4;
+  int64 available_kb = 5;
+  int64 buffers_kb = 6;
+  int64 cached_kb = 7;
```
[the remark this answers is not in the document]
vickyliu-go4it: I'm considering the case that we fail to extract the information from `/proc/meminfo`. I use the constant -1 to indicate missing informations.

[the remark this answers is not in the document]
dpallotti: Ah right. I would consider adding that constant to `GrpcProtos/Constants.h` so that it can also be used all the way to the client.

src/Memory/MemoryInfoProducer.cpp
dimitry-: also - it is better to use the method from OrbitBase/ThreadUtils.h

src/Memory/MemoryUtils.cpp
```diff
@@ -0,0 +1,69 @@
...
+#include <string>
+
+#include "OrbitBase/Logging.h"
+#include "OrbitBase/ReadFileToString.h"
+#include "OrbitBase/Result.h"
+#include "OrbitBase/ThreadUtils.h"
+
+using orbit_grpc_protos::TotalMemoryUsageInfo;
+
+std::optional<TotalMemoryUsageInfo> ParseMemInfo(const std::string& meminfo_content) {
```
florian-kuebler, line 20: Nit: `ParseMemoryInfo` and `memory_info_content`.  Per style guide, the use of abbreviations should be minimal.
https://google.github.io/styleguide/cppguide.html#General_Naming_Rules

src/Memory/MemoryUtils.h
```diff
@@ -0,0 +1,15 @@
+// Copyright (c) 2021 The Orbit Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#ifndef ORBIT_MEMORY_UTILS_H_
+#define ORBIT_MEMORY_UTILS_H_
+
+#include "capture.pb.h"
+
+std::optional<orbit_grpc_protos::TotalMemoryUsageInfo> ParseMemInfo(
```
florian-kuebler, line 10: I guess we want to use a namespace around those.

src/Memory/MemoryInfoProducer.cpp
```diff
@@ -0,0 +1,52 @@
...
+}
+
+void MemoryInfoProducer::ProduceTotalMemoryUsageInfoAndSendToListener() {
+  std::optional<TotalMemoryUsageInfo> total_memory_usage_info;
+  uint64_t current_timestamp_ns = orbit_base::CaptureTimestampNs();
+  while (!(*exit_requested_)) {
+    total_memory_usage_info = GetTotalMemoryUsageInfo(current_timestamp_ns);
+    if (total_memory_usage_info.has_value()) {
+      listener_->OnTotalMemoryUsageInfo(total_memory_usage_info.value());
+    }
```
florian-kuebler, line 46: I am not sure if this is already possible (there was a similar discussion with @ronaldfw). It would be nice, to notify the UI in case of an error. If that is not yet possible (which I assume is the case), would be nice if you could create a bug and add a TODO here.

src/Memory/MemoryUtilsTest.cpp
```diff
@@ -0,0 +1,110 @@
...
+  const std::string kPartialMeminfo = absl::Substitute(R"(MemTotal:       $0 kB
+MemFree:        $1 kB
+SwapCached:      0 kB)",
+                                                       kMemTotal, kMemFree);
+
+  const std::string kEmptyMeminfo = "";
+
+  std::optional<TotalMemoryUsageInfo> parsing_result;
+
+  // TODO(vickyliu): Check whether we are able to user testing::EqualsProto from the #include
```
florian-kuebler, line 88: "to use"

Also you can have a Look at the "VulkanLayerControllerTest" line 132 to see gmock matchers in action.

src/OrbitCaptureClient/CaptureEventProcessor.cpp
```diff
@@ -79,6 +79,9 @@ void CaptureEventProcessor::ProcessEvent(const ClientCaptureEvent& event) {
     case ClientCaptureEvent::kModuleUpdateEvent:
       // TODO (http://b/168797897): Process module update events
       break;
+    case ClientCaptureEvent::kTotalMemoryUsageInfo:
+      // TODO: Process the total memory usage information.
```
florian-kuebler, line 83: Please add a bug id (or at least your LDAP).

src/Memory/include/Memory/MemoryInfoListener.h
```diff
@@ -0,0 +1,20 @@
...
+// found in the LICENSE file.
+
+#ifndef MEMORY_INFO_LISTENER_H_
+#define MEMORY_INFO_LISTENER_H_
+
+#include "capture.pb.h"
+
+namespace orbit_memory {
+
+class MemoryInfoListener {
```
florian-kuebler, line 12: The following should be done for all new classes:
Every class should have a brief comment what it is good for/how to use it:
https://google.github.io/styleguide/cppguide.html#Class_Comments

florian-kuebler (commented):
Very clean and nice PR. Just some minor comments. In general looks good.

src/Memory/MemoryUtils.cpp
dpallotti: The reason for the name of the function is that the file is called `/proc/meminfo`.

florian-kuebler: Makes sense.

src/GrpcProtos/capture.proto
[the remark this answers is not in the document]
vickyliu-go4it: I have moved the constant kMissingInfo to `GrpcProtos/Constants.h`. Thanks for the suggestion.

Also, I need to apologize that I rebased the fixup commits (before receiving the new comments). So sorry for the inconvenience. Changes corresponding to your suggestions are included in fixup commits starting from "[fixup] Rename the Memory module as MemoryTracing"

src/LinuxTracing/LinuxTracingIntegrationTest.cpp
```diff
@@ -420,6 +420,8 @@ void VerifyOrderOfAllEvents(const std::vector<orbit_grpc_protos::ProducerCapture
         EXPECT_GE(event.module_update_event().timestamp_ns(), previous_event_timestamp_ns);
         previous_event_timestamp_ns = event.module_update_event().timestamp_ns();
         break;
+      case orbit_grpc_protos::ProducerCaptureEvent::kTotalMemoryUsageInfo:
+        break;
```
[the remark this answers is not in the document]
vickyliu-go4it: Changed accordingly. Thanks for the suggestion.

src/Memory/CMakeLists.txt
```diff
@@ -0,0 +1,43 @@
+# Copyright (c) 2021 The Orbit Authors. All rights reserved.
+# Use of this source code is governed by a BSD-style license that can be
+# found in the LICENSE file.
+
+cmake_minimum_required(VERSION 3.15)
+
+project(Memory)
```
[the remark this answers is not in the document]
vickyliu-go4it: Thanks for pointing out this, I have renamed this module as `MemoryTracing` as it sounds similar to `LinuxTracing`.
(Relevant changes are included in the commit *[fixup] Rename the Memory module as MemoryTracing*).

src/Service/MemoryInfoHandler.h
```diff
@@ -0,0 +1,37 @@
+// Copyright (c) 2021 The Orbit Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#ifndef MEMORY_INFO_HANDLER_H_
```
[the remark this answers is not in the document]
vickyliu-go4it: Thanks for pointing out this mistake. I have added the module names to the include guards. 
(Relevant changes are included in the commit *[fixup] Add module name to the include guards*)

src/Service/CaptureServiceImpl.cpp
```diff
@@ -166,7 +167,7 @@ class GrpcCaptureEventSender final : public CaptureEventSender {
 // for example until all CaptureEvents from external producers have been received.
 // Hence why these methods need to be called in parallel on different threads.
 static void StopTracingHandlerAndCaptureStartStopListenersInParallel(
```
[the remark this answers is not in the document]
vickyliu-go4it: Make sense. I changed the method name accordingly. Please see the commit *[fixup] Address other review comments*.

src/OrbitCaptureClient/CaptureEventProcessor.cpp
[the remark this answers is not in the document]
vickyliu-go4it: Thanks for the suggestion. I added the corresponding bug here. Please refer to the commit *[fixup] Adjust the field numbers in ClientCaptureEvent*.

src/Memory/include/Memory/MemoryInfoListener.h
```diff
@@ -0,0 +1,20 @@
+// Copyright (c) 2021 The Orbit Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#ifndef MEMORY_INFO_LISTENER_H_
```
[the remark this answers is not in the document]
vickyliu-go4it: Thanks for pointing out this. I have updated the include guards.

src/Memory/MemoryUtils.cpp
```diff
@@ -0,0 +1,69 @@
...
+  TotalMemoryUsageInfo memory_info;
+  memory_info.set_total_kb(kMissingInfo);
+  memory_info.set_free_kb(kMissingInfo);
+  memory_info.set_available_kb(kMissingInfo);
+  memory_info.set_buffers_kb(kMissingInfo);
+  memory_info.set_cached_kb(kMissingInfo);
+
+  constexpr int kNumLines = 5;
+  int mem_total, mem_free, mem_available, mem_buffers, mem_cached;
+  for (int i = 0, end = lines.size() > kNumLines ? kNumLines : lines.size(); i < end; ++i) {
```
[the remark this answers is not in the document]
vickyliu-go4it: Changed accordingly. Thanks! (Changes are in commit *[fixup] Address other review comments*).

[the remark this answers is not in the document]
vickyliu-go4it: Done. Thanks for pointing out this! (Changes are in commit [fixup] Address other review comments)

src/Memory/MemoryUtils.cpp
```diff
@@ -0,0 +1,69 @@
...
+std::optional<TotalMemoryUsageInfo> GetTotalMemoryUsageInfo(uint64_t timestamp_ns) noexcept {
+  ErrorMessageOr<std::string> reading_result = orbit_base::ReadFileToString("/proc/meminfo");
+  if (!reading_result) {
+    ERROR("%s", reading_result.error().message());
+    return std::nullopt;
+  }
+
+  std::optional<TotalMemoryUsageInfo> parsing_result = ParseMemInfo(reading_result.value());
+  if (parsing_result.has_value()) {
+    parsing_result.value().set_pid(orbit_base::GetCurrentProcessId());
```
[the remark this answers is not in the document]
vickyliu-go4it: Just renamed the `TotalMemoryUsageInfo` as `SystemMemoryUsage`, and removed `pid` from that.
Thanks pointing out this.

src/Memory/MemoryUtils.cpp
```diff
@@ -0,0 +1,69 @@
...
+      memory_info.set_buffers_kb(mem_buffers);
+    } else if (splits[0] == "Cached:" && absl::SimpleAtoi(splits[1], &mem_cached)) {
+      memory_info.set_cached_kb(mem_cached);
+    }
+  }
+
+  return memory_info;
+}
+
+std::optional<TotalMemoryUsageInfo> GetTotalMemoryUsageInfo(uint64_t timestamp_ns) noexcept {
```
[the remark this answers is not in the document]
vickyliu-go4it: Thanks for the suggestion. I changed to let this method (now renamed as `GetSystemMemoryUsage`) itself to get and set the timestamp. Changes are included in the commit *[fixup] Handel the sampling period with sleep_until*.

src/Memory/MemoryUtils.cpp
```diff
@@ -0,0 +1,69 @@
+// Copyright (c) 2021 The Orbit Authors. All rights reserved.
```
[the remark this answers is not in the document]
vickyliu-go4it: Thanks for pointing out this mistake. I added the namespace in the commit *[fixup] Rename MemoryUtils as MemoryTracingUtils and add namespace*.

src/Memory/MemoryUtils.h
```diff
@@ -0,0 +1,15 @@
+// Copyright (c) 2021 The Orbit Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#ifndef ORBIT_MEMORY_UTILS_H_
+#define ORBIT_MEMORY_UTILS_H_
+
+#include "capture.pb.h"
+
```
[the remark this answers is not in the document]
vickyliu-go4it: Done. Thanks!

src/Memory/MemoryUtilsTest.cpp
vickyliu-go4it: Just removed this TODO. And I will spend sometime later to figure out how this can be resolved. Thanks for the suggestion.

src/Memory/MemoryInfoProducer.cpp
vickyliu-go4it: I changed to use methods from the ThreadUtils.h and set the thread name as `orbit_base::SetCurrentThreadName("MIP::Run")`.

This change is included in the commit *[fixup] Set thread name with the ThreadUtils method*.

Sorry again for the inconvenience brought by my rebasing :-(

src/Memory/include/Memory/MemoryInfoProducer.h
vickyliu-go4it: Make sense. I just removed the smart pointer. Thanks for pointing out this. Please see the commit *[fixup] Not use smart pointer in MemoryInfoProducer*.

src/Memory/include/Memory/MemoryInfoProducer.h
vickyliu-go4it: Changed to use `std::thread* thread_;` Thanks for pointing out this.

src/Memory/include/Memory/MemoryInfoProducer.h
vickyliu-go4it: Changed accordingly. Thanks for pointing out this mistake. Please see the commit *[fixup] Add module name to the include guards*)

src/Memory/MemoryInfoProducer.cpp
vickyliu-go4it: Thanks for pointing out this and providing the helpful suggestions!

I changed to compute the scheduled running time for the next memory sampling (i.e., current scheduled time + sampling period), and use the `sleep_until` method to block the thread until the scheduled running time.

I'm not sure about how this has been done in games, but I take a look at the code in Perfetto. 
After finishing current round of memory sampling, they create a delayed task for the next round of memory sampling.
The delayed task includes the memory sampling function and the scheduled time for running.  They use a `UnixTaskRunner` to manage all delayed tasks in a `std::multimap<TimeMillis, std::function<void()>> delayed_tasks_`. This `UnixTaskRunner` will keep comparing the scheduled time of the delayed tasks with the current time to determine whether this task should be run and erased.

src/Service/MemoryInfoHandler.h
vickyliu-go4it: Changed accordingly. Thanks! Please refer to the commit *[fixup] Address other review comments*.

src/Memory/CMakeLists.txt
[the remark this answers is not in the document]
vickyliu-go4it: Thanks for the suggestion! I like the name `MemoryTracing`!

Also I need to apologize that I rebased the fixup commits. I'm not good at handling a PR with many commits :-( Sorry for the inconvenience. Changes corresponding to your suggestions are included in fixup commits starting from "[fixup] Rename the Memory module as MemoryTracing".

src/Memory/MemoryUtils.h
vickyliu-go4it: Thanks for pointing out this mistake. I added the namespace in the commit *[[fixup] Rename MemoryUtils as MemoryTracingUtils and add namespace](https://github.com/google/orbit/pull/1918/commits/8fac00527b70ed498b52a9c6d24757db62b6a7b8)*.

src/Memory/MemoryInfoProducer.cpp
vickyliu-go4it: Thanks for the suggestion. Should I add it here or in the `GetSystemMemoryUsage()` method (the `GetTotalMemoryUsageInfo()` method is now renamed as `GetSystemMemoryUsage()`) where we generate an error if fails to read the `/proc/meminfo` file? I feel like it is a little bit strange as there is no error mentioned in this file.

src/Memory/MemoryUtilsTest.cpp
vickyliu-go4it: Thanks so much for providing the information about `Matcher`.

But I had some trouble when applying the `Field` method to the message. `VkExtensionProperties` is a struct so we can handle the filed in the way of `Field("specVersion", &VkExtensionProperties::specVersion, expected.specVersion)`. But I cannot process the filed of `SystemMemoryUsage` in a similar way, e.g.,  `Field("total_kb", &SystemMemoryUsage::total_kb(), expected.total_kb())`. There is an error about the second param, which says that I cannot access a non-static member without an object.

I removed the comments and keep the rest part as it is for now. I will try to address this problem later.

src/OrbitCaptureClient/CaptureEventProcessor.cpp
vickyliu-go4it: Thanks for the suggestion. I added the corresponding bug here. Please refer to the commit *[[fixup] Adjust the field numbers in ClientCaptureEvent](https://github.com/google/orbit/pull/1918/commits/9375c9730a4c3547ce0778f932d5df17178ee99e)*.

src/Memory/include/Memory/MemoryInfoListener.h
vickyliu-go4it: Thanks so much for pointing out this. I have added descriptions for all newly added classes in the commit 
*[[fixup] Add class descriptions](https://github.com/google/orbit/pull/1918/commits/5f6e2176ad33d13ac242a42f8e026ce3a8462af3)*.

vickyliu-go4it:
Hi Dimitry, Davide and Florian! Thanks so much for reviewing this PR. I feel really sorry for rebasing the fixup commits. I'm not good at handling large PRs (maybe I should split them into smaller ones). And when I realized that rebasing is a really bad idea to organize the fixup commits, I already did it :-( ...

Sorry for the inconvenience in the re-reviewing! I added the relevant commit name and link in the responses to your comments. Hope this can help a little bit.

src/Memory/MemoryInfoProducer.cpp
florian-kuebler: If we would have that communication to the UI, it would be added here (something like `listener_->OnError`, or `listener_->OnMetaInfo`). So I would add the TODO here.
Also note that this is totally optional. I was just wondering, as I remembered the discussion on other places.

src/Memory/MemoryUtilsTest.cpp
florian-kuebler: Have you checked the gMock cookbook? https://github.com/google/googletest/blob/master/docs/gmock_cook_book.md#validating-a-member-of-an-object

But in general it is totally fine to leave it as it is. I'd don't think its worth to invest a lot of time here, other then learning the gMock API. Also, you could simply move the `EXPECT`s into a helper method.

src/Memory/include/Memory/MemoryInfoProducer.h
florian-kuebler: But now we are creating the thread with `new`, which seems very suspicious to me. Couldn't it be just a member of the producer? E.g. as `std::optional`, if you need to "reset" it on capture stop?
If not it should be probably still a `unique_ptr`.

But passing `std::thread*` is better than the shared_ptr, I agree.

src/MemoryTracing/MemoryInfoProducer.cpp
```diff
@@ -0,0 +1,53 @@
...
+  exit_requested_ = false;
+  thread_ = new std::thread(&MemoryInfoProducer::Run, this);
+}
+
+void MemoryInfoProducer::Stop() {
+  exit_requested_ = true;
+  if (thread_ != nullptr && thread_->joinable()) {
+    thread_->join();
+  }
+  delete thread_;
```
dimitry-, line 29: use unique_ptr<> for thread, and call reset() here.

src/MemoryTracing/include/MemoryTracing/MemoryInfoProducer.h
```diff
@@ -0,0 +1,49 @@
...
+  void Stop();
+
+ private:
+  void Run();
+  void ProduceSystemMemoryUsageAndSendToListener();
+
+  uint64_t sampling_period_ns_;
+  MemoryInfoListener* listener_ = nullptr;
+  std::atomic<bool> exit_requested_ = true;
+  std::thread* thread_;
```
dimitry-, line 44: unique_ptr<>

src/MemoryTracing/MemoryTracingUtils.cpp
```diff
@@ -0,0 +1,76 @@
...
+  memory_info.set_available_kb(kMissingInfo);
+  memory_info.set_buffers_kb(kMissingInfo);
+  memory_info.set_cached_kb(kMissingInfo);
+
+  constexpr size_t kNumLines = 5;
+  size_t end = lines.size() > kNumLines ? kNumLines : lines.size();
+  int mem_total, mem_free, mem_available, mem_buffers, mem_cached;
+  for (size_t i = 0; i < end; ++i) {
+    const std::string& current_line = lines[i];
+    std::vector<std::string> splits = absl::StrSplit(current_line, ' ', absl::SkipWhitespace{});
```
dimitry-, line 42: maybe use absl::Span<> here and split in 2 by using absl::MaxSplits(' ', 2)?

src/MemoryTracing/MemoryInfoProducer.cpp
```diff
@@ -0,0 +1,53 @@
...
+void MemoryInfoProducer::ProduceSystemMemoryUsageAndSendToListener() {
+  std::optional<SystemMemoryUsage> system_memory_usage;
+  std::chrono::time_point scheduled_time = std::chrono::steady_clock::now();
+  while (!exit_requested_) {
+    system_memory_usage = GetSystemMemoryUsage();
+    if (system_memory_usage.has_value()) {
+      listener_->OnSystemMemoryUsage(system_memory_usage.value());
+    }
+    scheduled_time += std::chrono::nanoseconds(sampling_period_ns_);
+    std::this_thread::sleep_until(scheduled_time);
```
dimitry-, line 49: if sampling_period_ns_ is too big, stop would block for quite considerable time. You can avoid it by protecting stop_requested by a absl::Mutex, and wait on it with timeout with early interruption if flag was set.

Example: https://github.com/google/orbit/blob/main/src/OrbitBase/ThreadPool.cpp#L142

In that case if flag is set - the thread will wake - check the flag and exit.

src/MemoryTracing/MemoryTracingUtils.cpp
```diff
@@ -0,0 +1,76 @@
...
+  memory_info.set_total_kb(kMissingInfo);
+  memory_info.set_free_kb(kMissingInfo);
+  memory_info.set_available_kb(kMissingInfo);
+  memory_info.set_buffers_kb(kMissingInfo);
+  memory_info.set_cached_kb(kMissingInfo);
+
+  constexpr size_t kNumLines = 5;
+  size_t end = lines.size() > kNumLines ? kNumLines : lines.size();
+  int mem_total, mem_free, mem_available, mem_buffers, mem_cached;
+  for (size_t i = 0; i < end; ++i) {
```
dimitry-, line 40: id seems you don't use i below, you can probably replace this with `for (const std::string& line : lines) { ...`

src/MemoryTracing/MemoryInfoProducer.cpp
florian-kuebler: Would you prefer unique_ptr over std::optional?

dimitry-: the class owns the thread here - std::optional won't delete it so I would say this should be std::unique_ptr<>
Rule of thumb: use unique_ptr to establish ownership of the object.

dpallotti: Why won't `std::optional` delete it?

dimitry-: Well, because I was thinking about std::optional<std::thread*>.
so, yes, you can use optional<thread> here.

src/GrpcProtos/capture.proto
```diff
@@ -308,5 +322,6 @@ message ProducerCaptureEvent {
     SchedulingSlice scheduling_slice = 12;
     ThreadName thread_name = 13;
     ThreadStateSlice thread_state_slice = 14;
+    SystemMemoryUsage system_memory_usage = 15;
```
dpallotti, line 325: Move up to keep the alphabetical order.

src/MemoryTracing/MemoryInfoProducer.cpp
```diff
@@ -0,0 +1,53 @@
...
+  exit_requested_ = true;
+  if (thread_ != nullptr && thread_->joinable()) {
+    thread_->join();
+  }
+  delete thread_;
+  thread_ = nullptr;
+}
+
+void MemoryInfoProducer::Run() {
+  orbit_base::SetCurrentThreadName("MIP::Run");
```
dpallotti, line 34: Tiny suggestion: "MemInfoPr::Run" so that you fit in 15 characters but still with more info.

src/MemoryTracing/MemoryInfoProducer.cpp
```diff
@@ -0,0 +1,53 @@
...
+#include "OrbitBase/ThreadUtils.h"
+#include "capture.pb.h"
+
+namespace orbit_memory_tracing {
+
+using orbit_grpc_protos::SystemMemoryUsage;
+
+void MemoryInfoProducer::Start() {
+  exit_requested_ = false;
+  thread_ = new std::thread(&MemoryInfoProducer::Run, this);
```
dpallotti, line 21: As mentioned below, never use `new` and `delete` (unless there's a very specific reason like Qt ownership and such).

src/Memory/MemoryInfoProducer.cpp
dpallotti: This makes sense to me, thanks.

dpallotti:
> I feel really sorry for rebasing the fixup commits.

Don't worry too much! Either way GitHub is a bit of a mess with or without rebasing.

src/Memory/MemoryUtilsTest.cpp
vickyliu-go4it: Thanks for providing this helpful resource for gMock. Also I moved the `EXPECT`s into a helper method called `CheckMemInfoParsingResult(...)`.

src/MemoryTracing/MemoryInfoProducer.cpp
vickyliu-go4it: Done. Thanks for the suggestion! Changed to use `unique_ptr` for both `exit_requested_` and `thread_`.

src/MemoryTracing/include/MemoryTracing/MemoryInfoProducer.h
vickyliu-go4it: Done. Thanks for the suggestion!

src/MemoryTracing/MemoryTracingUtils.cpp
vickyliu-go4it: Thanks for pointing out this and providing helpful suggestions! 

I'm considering that lines in `/proc/meminfo` end with "KB", for example "MemTotal:       1234567 kB". If we use `absl::MaxSplits(' ', 2)`, then the "KB" will be included in the second element of the splitting result.

src/MemoryTracing/MemoryInfoProducer.cpp
vickyliu-go4it: Thanks so much for raising this issue! I changed to use `absl::Mutex` and `AwaitWithDeadline` in the `MemoryInfoProducer`.

src/MemoryTracing/MemoryTracingUtils.cpp
vickyliu-go4it: Done. Thanks for the suggestion!

src/GrpcProtos/capture.proto
vickyliu-go4it: I forgot to change the order after the renaming! Thanks so much for pointing out this!

src/MemoryTracing/MemoryInfoProducer.cpp
vickyliu-go4it: Thanks for suggesting the thread name! Changed to use "MemInfoPr::Run".

src/MemoryTracing/MemoryInfoProducer.cpp
vickyliu-go4it: Changed to use `unique_ptr` and `reset`. Also thanks for telling me about this! I will pay attention in the future!

src/MemoryTracing/MemoryTracingUtils.cpp
dimitry-: hmm, does it mean we need to parse KB MB GB as well? they seem to be ignored now.

src/MemoryTracing/MemoryInfoProducer.cpp
```diff
@@ -0,0 +1,61 @@
...
+  orbit_base::SetCurrentThreadName("MemInfoPr::Run");
+
+  CHECK(listener_ != nullptr);
+  ProduceSystemMemoryUsageAndSendToListener();
+}
+
+void MemoryInfoProducer::ProduceSystemMemoryUsageAndSendToListener() {
+  std::optional<SystemMemoryUsage> system_memory_usage;
+  absl::Time scheduled_time = absl::Now();
+  exit_requested_mutex_.ReaderLock();
```
vickyliu-go4it, line 49: I feel like that with this read lock, we will not able to acquire a write lock and so a large `sampling_period_ns_`.

But when I try to remove these reader lock and unlock in line 49 and 58, and add the following method for reading `exit_requested_` in the current line 50 and line 56:
```cpp
bool MemoryInfoProducer::GetExitRequested() {
  exit_requested_mutex_.ReaderLock();
  bool exit_requested = *exit_requested_;
  exit_requested_mutex_.ReaderUnlock();
  return exit_requested;
}
```
I found the following error in the log file
```shell
[                OrbitService] [mutex.cc : 2410] RAW: thread should hold at least a read lock on Mutex
```

dimitry-: yes - you probably want a regular lock here.. note that it will unlock on line 56, so the setter will be able to write at this point.

src/MemoryTracing/MemoryTracingUtils.cpp
vickyliu-go4it: Yes. In the current version I split  "MemTotal: 1234567 kB" into three pieces, i.e., "MemTotal:", "1234567" and "kB", and only extract information from the second splitting piece. 
With `absl::MaxSplits(' ', 2)` we will get two pieces, i.e.,  "MemTotal:" and "1234567 kB". Do we need also parse KB, MB, GB? I'm assuming that the content of `/proc/meminfo` has a fixed format.

src/MemoryTracing/MemoryInfoProducer.cpp
vickyliu-go4it: Thanks so much! I got stuck on this for a long time!

src/MemoryTracing/MemoryInfoProducer.cpp
```diff
@@ -0,0 +1,61 @@
...
+
+void MemoryInfoProducer::Stop() {
+  SetExitRequested(true);
+  if (thread_ != nullptr && thread_->joinable()) {
+    thread_->join();
+  }
+  thread_.reset();
+}
+
+void MemoryInfoProducer::SetExitRequested(bool status) {
```
dpallotti, line 34: I would rename the parameter to `exit_requested`.

src/MemoryTracing/include/MemoryTracing/MemoryInfoProducer.h
```diff
@@ -0,0 +1,52 @@
...
+  void Stop();
+
+ private:
+  void SetExitRequested(bool status);
+  void Run();
+  void ProduceSystemMemoryUsageAndSendToListener();
+
+  uint64_t sampling_period_ns_;
+  MemoryInfoListener* listener_ = nullptr;
+  std::unique_ptr<bool> exit_requested_ = std::make_unique<bool>(true);
```
dpallotti, line 45: Do you need to wrap the `bool` in a `unique_ptr`? Can't you have a simple `bool`, and simply take its address when using it with `absl::Condition`?

src/MemoryTracing/MemoryTracingUtilsTest.cpp
```diff
@@ -0,0 +1,123 @@
...
+
+#include "GrpcProtos/Constants.h"
+#include "MemoryTracingUtils.h"
+
+using orbit_grpc_protos::kMissingInfo;
+using orbit_grpc_protos::SystemMemoryUsage;
+
+namespace {
+
+void CheckMemInfoParsingResult(std::optional<SystemMemoryUsage> parsing_result,
```
dpallotti, line 19: I would suggest `ExpectMemInfoParsingResult` instead of `CheckMemInfoParsingResult` as with "check" we mean something specific in our codebase, while with "expect" you hint at the fact that you use `EXPECT_...` internally.

dpallotti (commented):
In general it looks good, I only left a couple of minor comments.

But I think we should double check whether we need to parse the unit in `/proc/meminfo` as Dimitry pointed out. It does seem like `kB` is always used, but the manpage of `procfs` simply says 
`Each line of the file consists of a parameter
              name, followed by a colon, the value of the parameter, and
              an option unit of measurement (e.g., "kB").`. So I'm not sure we should make the assumption that it's always kilobytes.

src/MemoryTracing/include/MemoryTracing/MemoryInfoProducer.h
vickyliu-go4it: Ah, I mixed up the suggestion on `thread_` and `exit_requested_`! Will change to use simple boolean for this.

src/MemoryTracing/MemoryInfoProducer.cpp
```diff
@@ -0,0 +1,61 @@
...
+  orbit_base::SetCurrentThreadName("MemInfoPr::Run");
+
+  CHECK(listener_ != nullptr);
+  ProduceSystemMemoryUsageAndSendToListener();
+}
+
+void MemoryInfoProducer::ProduceSystemMemoryUsageAndSendToListener() {
+  std::optional<SystemMemoryUsage> system_memory_usage;
+  absl::Time scheduled_time = absl::Now();
+  exit_requested_mutex_.Lock();
```
dimitry-, line 49: you can now use absl::MutexLock here.

src/MemoryTracing/MemoryInfoProducer.cpp
```diff
@@ -0,0 +1,61 @@
...
+  while (!(*exit_requested_)) {
```
dimitry-: as Davide mentioned this should probably be plain bool now

dimitry- (approved):
otherwise LGTM

src/MemoryTracing/MemoryInfoProducer.cpp
vickyliu-go4it: Make sense. Changed. Thanks!

src/MemoryTracing/MemoryInfoProducer.cpp
vickyliu-go4it: Changed! Thanks for raising this issue. I mixed up the suggestion on `thread_` and `exit_requested_`.

src/MemoryTracing/MemoryInfoProducer.cpp
vickyliu-go4it: Changed accordingly. Thanks!

src/MemoryTracing/MemoryTracingUtilsTest.cpp
vickyliu-go4it: Changed accordingly. Thanks for suggesting the name.

vickyliu-go4it:
> In general it looks good, I only left a couple of minor comments.
> 
> But I think we should double check whether we need to parse the unit in `/proc/meminfo` as Dimitry pointed out. It does seem like `kB` is always used, but the manpage of `procfs` simply says
> `Each line of the file consists of a parameter name, followed by a colon, the value of the parameter, and an option unit of measurement (e.g., "kB").`. So I'm not sure we should make the assumption that it's always kilobytes.

Thanks for the detailed information on the units! I will create a small follow up PR to also parse the unit and update the unit tests.

dpallotti (approved):
> I will create a small follow up PR to also parse the unit and update the unit tests.

Sounds good. Maybe consider adding a TODO(+bug).

florian-kuebler (approved):
LGTM, thanks!

PR #1919, opened 2021-02-23
File.h: Implement ReadFully

dimitry-:
Test: run unit-tests

### DIFF
--- a/src/OrbitBase/File.cpp
+++ b/src/OrbitBase/File.cpp
@@ -6,6 +6,17 @@
 
 #include "OrbitBase/Logging.h"
 
+#if defined(__linux)
+#include <unistd.h>
+#elif defined(_WIN32)
+#include <io.h>
+#endif
+
+#if defined(_WIN32)
+// Windows never returns EINTR - so there is no need for TEMP_FAILURE_RETRY implementation
+#define TEMP_FAILURE_RETRY(expression) (expression)
+#endif
+
 namespace orbit_base {
 
 static void FdDeleter(int fd) {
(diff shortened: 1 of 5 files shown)


src/OrbitBase/File.cpp
```diff
@@ -6,6 +6,17 @@
...
 #include "OrbitBase/Logging.h"
 
+#if defined(__linux)
+#include <unistd.h>
+#elif defined(_WIN32)
+#include <io.h>
+#endif
+
+#if defined(_WIN32)
+// Windows does not have TEMP_FAILURE_RETRY - define a shortcut
```
dpallotti, line 16: It's not really a shortcut, more like a fallback.

src/OrbitBase/File.cpp
```diff
@@ -61,4 +72,22 @@ ErrorMessageOr<void> WriteFully(const unique_fd& fd, std::string_view content) {
   return outcome::success();
 }
 
+ErrorMessageOr<size_t> ReadFully(const unique_fd& fd, void* buffer, size_t size) {
```
dpallotti, line 75: I guess that this depends on what use you have in mind for this function, but it's annoying to have to provide a preallocated buffer and to have to guess the size. It makes for a more complicated replacement for `read` while basically providing the same interface. In particular, in doesn't provide the convenience that one would expect from `Fully`.
Can't a vector of bytes be returned, maybe passing a *maximum size* instead of the size of the buffer?

dimitry-: There are couple of reason for doing it this way:
1. Usually the pattern in which this is used is when we repeatedly read chunks from file and do some computations.
2. The only(!) case when actual is less then requested size is end of file.
3. There are 2 other functions that let you read whole file and (coming soon) the one that reads one line, they return std::string

The convenience is still there since you do not need to handle interruptions by signals (see 2).

src/OrbitBase/File.cpp
dimitry-: I still have in my todo to look if EINTR is generated by windows, I suspect this error code is not at all used. Does anyone happen to know btw?

beckerhe: I think you can ignore EINTR. It can't occur on Windows (Ref: https://docs.microsoft.com/en-us/cpp/c-runtime-library/errno-constants?view=msvc-160)

beckerhe (approved):
Thank you for the quick implementation! LGTM!

src/OrbitBase/File.cpp
dpallotti: Sure, but then it shouldn't be called `ReadFully`, all your three points are basically stating that "fully" is the opposite of what this function is doing.

PR #1920, opened 2021-02-23
Handle errors when exporting to CSV

dimitry-:
This change also gets rid of ofstream in favor of OrbitBase/File.h
which does not throw exceptions.

Also according to RCF CSV lines should end with CRLF.

Test: export a table to CSV - open in google drive

### DIFF
--- a/src/OrbitGl/DataView.h
+++ b/src/OrbitGl/DataView.h
@@ -17,6 +17,7 @@
 
 #include "DataViewTypes.h"
 #include "OrbitBase/Logging.h"
+#include "OrbitBase/Result.h"
 #include "absl/container/flat_hash_set.h"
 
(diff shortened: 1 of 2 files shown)


src/OrbitGl/DataView.h
```diff
@@ -17,6 +17,7 @@
 
 #include "DataViewTypes.h"
 #include "OrbitBase/Logging.h"
+#include "OrbitBase/Result.h"
```
beckerhe, line 20: `<filesystem>` should be included as well.

beckerhe (approved):
LGTM!

src/OrbitGl/DataView.h
dimitry-: done.

PR #1921, opened 2021-02-23
Use File Descriptor for loading presets

dimitry-:
Test: Start Orbit -> Load presets

### DIFF
--- a/src/OrbitGl/App.cpp
+++ b/src/OrbitGl/App.cpp
@@ -24,7 +24,6 @@
 #include <cstddef>
 #include <cstdint>
 #include <exception>
-#include <fstream>
 #include <memory>
 #include <outcome.hpp>
 #include <ratio>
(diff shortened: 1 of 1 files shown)


dimitry-:
(forgot to fix the saving presets part). working on it.

src/OrbitGl/App.cpp
```diff
@@ -691,36 +691,39 @@ ErrorMessageOr<void> OrbitApp::SavePreset(const std::string& filename) {
...
-  std::ofstream file(filename_with_ext, std::ios::binary);
-  if (file.fail()) {
-    ERROR("Saving preset in \"%s\": %s", filename_with_ext, "file.fail()");
-    return ErrorMessage(
-        absl::StrFormat("Error opening the file \"%s\" for writing", filename_with_ext));
+  auto open_result = orbit_base::OpenFileForWriting(filename_with_ext);
+  if (open_result.has_error()) {
+    ERROR("Failed to open \"%s\": %s", filename_with_ext, open_result.error().message());
+    return ErrorMessage(absl::StrFormat("Failed to open \"%s\": %s", filename_with_ext,
+                                        open_result.error().message()));
```
IrinaShkviro, line 698: Maybe define a message and use it in ERROR and ErrorMessage (as it's done in line 716)? There also similar cases below

IrinaShkviro (approved):
Thank Dimitry! I've left one comment below

src/OrbitGl/App.cpp
dimitry-: done.

PR #1922, opened 2021-02-23
Make source code view work for modules with integrated debuginfo

beckerhe:
This enabled the source code view for the very simple use case, when code, symbols, and debuginfo are in the same file.

### DIFF
--- a/src/OrbitGl/App.cpp
+++ b/src/OrbitGl/App.cpp
@@ -136,12 +136,16 @@ PresetLoadState GetPresetLoadStateForProcess(
 
 bool DoZoom = false;
 
-OrbitApp::OrbitApp(MainThreadExecutor* main_thread_executor,
+OrbitApp::OrbitApp(orbit_gl::MainWindowInterface* main_window,
+                   MainThreadExecutor* main_thread_executor,
                    const orbit_base::CrashHandler* crash_handler,
                    orbit_metrics_uploader::MetricsUploader* metrics_uploader)
-    : main_thread_executor_(main_thread_executor),
+    : main_window_{main_window},
+      main_thread_executor_(main_thread_executor),
       crash_handler_(crash_handler),
       metrics_uploader_(metrics_uploader) {
+  CHECK(main_window_ != nullptr);
+
   thread_pool_ = ThreadPool::Create(4 /*min_size*/, 256 /*max_size*/, absl::Seconds(1));
   main_thread_id_ = std::this_thread::get_id();
   data_manager_ = std::make_unique<DataManager>(main_thread_id_);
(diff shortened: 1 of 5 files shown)


src/OrbitQt/orbitmainwindow.cpp
```diff
@@ -1270,4 +1268,29 @@ orbit_qt::TargetConfiguration OrbitMainWindow::ClearTargetConfiguration() {
...
+  code_viewer_dialog.SetHighlightCurrentLine(true);
+
+  QFile source_code_file{QString::fromStdString(file_path.string())};
+
+  if (!source_code_file.open(QIODevice::ReadOnly)) {
+    ERROR("Could not open file %s", file_path.string());
+    return;
+  }
+
+  auto syntax_highlighter = std::make_unique<orbit_syntax_highlighter::Cpp>();
```
IrinaShkviro, line 1291: Why it's unique_ptr and not just value?

IrinaShkviro (approved):
For me it looks good, but I think you should wait for other reviewers who worked more closely with modules loading.

src/OrbitGl/App.cpp
```diff
@@ -136,10 +136,12 @@ PresetLoadState GetPresetLoadStateForProcess(
 
 bool DoZoom = false;
 
-OrbitApp::OrbitApp(MainThreadExecutor* main_thread_executor,
+OrbitApp::OrbitApp(orbit_gl::MainWindowInterface* main_window,
+                   MainThreadExecutor* main_thread_executor,
                    const orbit_base::CrashHandler* crash_handler,
                    orbit_metrics_uploader::MetricsUploader* metrics_uploader)
-    : main_thread_executor_(main_thread_executor),
+    : main_window_{main_window},
```
dpallotti, line 144: Maybe add a check that `main_window` is not `nullptr`?

src/OrbitQt/orbitmainwindow.cpp
```diff
@@ -1272,4 +1273,24 @@ orbit_qt::TargetConfiguration OrbitMainWindow::ClearTargetConfiguration() {
...
+void OrbitMainWindow::ShowSourceCode(const std::filesystem::path& file_path, size_t line_number) {
+  orbit_code_viewer::Dialog code_viewer_dialog{this};
+
+  code_viewer_dialog.SetEnableLineNumbers(true);
+  code_viewer_dialog.SetHighlightCurrentLine(true);
+
+  QFile source_code_file{QString::fromStdString(file_path.string())};
+
+  if (!source_code_file.open(QIODevice::ReadOnly)) {
+    ERROR("Could not open file %s", file_path.string());
```
dpallotti, line 1287: I'd add quotes around `%s` or a colon before it.

src/OrbitGl/MainWindowInterface.h
```diff
@@ -0,0 +1,26 @@
...
+#include <string_view>
+
+namespace orbit_gl {
+
+// This abstract base class is an attempt to simplify callbacks
+// in OrbitApp and make it easier to refactor things in the future.
+//
+// OrbitMainWindow and Mocks can derive from this and offer a fixed interface
+// to OrbitApp.
+class MainWindowInterface {
```
dpallotti, line 20: Nice!

dpallotti (approved):

src/OrbitQt/orbitmainwindow.cpp
beckerhe: That's a requirement coming from Qt. This class inherits from QObject. And QObject is neither movable nor copyable. So I have to heap-allocate it. And then I can move it into the dialog object in the next line.

But in general you're right: A value would be much more desirable.

IrinaShkviro: Thanks for explanation!

src/OrbitGl/App.cpp
beckerhe: Done.

src/OrbitQt/orbitmainwindow.cpp
beckerhe: Done.

src/OrbitGl/MainWindowInterface.h
beckerhe: Glad you like it as well!

PR #1923, opened 2021-02-23
Capture deserialization: Replace ifstream with file descriptor

dimitry-:
Test: run unit-tests
Test: save capture/load capture

### DIFF
--- a/src/OrbitClientModel/CaptureDeserializer.cpp
+++ b/src/OrbitClientModel/CaptureDeserializer.cpp
@@ -10,11 +10,11 @@
 
 #include <algorithm>
 #include <cstdint>
-#include <fstream>
 #include <memory>
 #include <utility>
 #include <vector>
 
+#include "OrbitBase/File.h"
 #include "OrbitBase/Logging.h"
 #include "OrbitBase/MakeUniqueForOverwrite.h"
 #include "OrbitBase/Result.h"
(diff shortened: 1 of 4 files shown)


src/OrbitClientModel/CaptureDeserializerTest.cpp
```diff
@@ -157,9 +169,15 @@ TEST(CaptureDeserializer, LoadNoCaptureInfo) {
   stream << std::string(absl::bit_cast<char*>(&size_of_header), sizeof(size_of_header))
          << serialized_header;
```
beckerhe, line 170: what about replacing this as well?

beckerhe: Should work in all three tests.

src/OrbitClientModel/CaptureDeserializer.cpp
```diff
@@ -43,39 +43,38 @@ using orbit_grpc_protos::ProcessInfo;
...
   SCOPED_TIMED_LOG("Loading capture from \"%s\"", file_name);
 
-  // Binary
-  std::ifstream file(file_name, std::ios::binary);
-  if (file.fail()) {
-    std::string error = absl::StrFormat("Error opening file \"%s\" for reading", file_name);
+  auto open_result = orbit_base::OpenFileForReading(file_name);
+  if (open_result.has_error()) {
+    std::string error = absl::StrFormat("Error opening file \"%s\" for reading: %s",
+                                        file_name.string(), open_result.error().message());
```
beckerhe, line 55: If I remember correctly `OpenFileForReading` already wraps a generic message around SafeStrerror.

beckerhe: Just checked. The message would look like this: 
```
Error opening file "filename.xyz" for reading: Unable to open file "filename.xyz": <the actual error from strerror>
```

I think it's fine to just return `open_result.error()`

beckerhe (approved):
Otherwise LGTM

src/OrbitClientModel/CaptureDeserializerTest.cpp
beckerhe: That's what I meant. Thank you. Looks good!

PR #1925, opened 2021-02-23
Add SymbolHelper::FindDebugInfoFileLocally

beckerhe:
This method goes through the search paths and checks for debuginfo file matching a certain filename and a checksum.

That's needed to find a debuginfo file which was referred to by a .gnu_debuglink section.

Test: Ran unit tests

### DIFF
--- a/src/OrbitCore/SymbolHelper.cpp
+++ b/src/OrbitCore/SymbolHelper.cpp
@@ -199,3 +199,49 @@ fs::path SymbolHelper::GenerateCachedFileName(const fs::path& file_path) const {
   auto file_name = absl::StrReplaceAll(file_path.string(), {{"/", "_"}});
   return cache_directory_ / file_name;
 }
+
+[[nodiscard]] static bool IsMatchingDebugInfoFile(const std::filesystem::path& debuginfo_file_path,
+                                                  uint32_t checksum) {
+  std::error_code error;
+  bool exists = fs::exists(debuginfo_file_path, error);
+  if (error) {
+    ERROR("Unable to stat \"%s\": %s", debuginfo_file_path.string(), error.message());
+    return false;
+  }
+
+  if (!exists) return false;
+
+  const auto checksum_or_error = ElfFile::CalculateDebuglinkChecksum(debuginfo_file_path);
+  if (checksum_or_error.has_error()) {
+    LOG("Unable to calculate checksum of \"%s\": \"%s\"", debuginfo_file_path.filename().string(),
+        checksum_or_error.error().message());
+    return false;
+  }
+
+  if (checksum_or_error.value() != checksum) {
+    LOG("Found file with matching name \"%s\", but the checksums do not match. Expected: %#x. "
+        "Actual: %#x",
+        debuginfo_file_path.string(), checksum, checksum_or_error.value());
+    return false;
+  }
+
+  LOG("Found debug info in file \"%s\"", debuginfo_file_path.string());
+  return true;
+}
+
+ErrorMessageOr<fs::path> SymbolHelper::FindDebugInfoFileLocally(std::string_view filename,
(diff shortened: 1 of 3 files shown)


src/OrbitCore/SymbolHelper.cpp
```diff
@@ -199,3 +199,44 @@ fs::path SymbolHelper::GenerateCachedFileName(const fs::path& file_path) const {
...
+
+    const auto checksum_or_error = ElfFile::CalculateDebuglinkChecksum(debuginfo_file_path);
+    if (checksum_or_error.has_error()) {
+      LOG("Unable to calculate checksum of \"%s\": \"%s\"", filename,
+          checksum_or_error.error().message());
+      continue;
+    }
+
+    if (checksum_or_error.value() != checksum) {
+      LOG("Found file with matching name \"%s\", but the checksum do not match. Expected: %#x. "
```
ronaldfw, line 229:
```suggestion
      LOG("Found file with matching name \"%s\", but the checksums do not match. Expected: %#x. "
```

src/OrbitCore/SymbolHelper.cpp
```diff
@@ -199,3 +199,44 @@ fs::path SymbolHelper::GenerateCachedFileName(const fs::path& file_path) const {
...
+    search_paths.insert(directory / filename);
+  }
+
+  LOG("Trying to find debuginfo file with filename \"%s\"", filename);
+  for (const auto& debuginfo_file_path : search_paths) {
+    std::error_code error;
+    bool exists = fs::exists(debuginfo_file_path, error);
+    if (error) {
+      ERROR("Unable to stat \"%s\": %s", debuginfo_file_path.string(), error.message());
+      continue;
```
ronaldfw, line 216: To consider, but totally fine as-is: I always find "continue" a bit more problematic than a return in a function. Maybe move the part in the loop into a function that returns true/false if the file was successfully found? Then all these continues become 'return false'.

ronaldfw (approved):
LGTM, though I'm missing some context, so might be better to wait for Anton and/or Dimitry! One small typo, and one nit that is just for consideration.

dimitry- (approved):
LGTM

src/OrbitCore/SymbolHelper.cpp
beckerhe: clang-tidy agrees with you. Too complex. Too many nestings. I will change it. Thanks!

src/OrbitCore/SymbolHelper.cpp
beckerhe: done.

src/OrbitCore/SymbolHelper.cpp
beckerhe: done.

ronaldfw (approved):
Perfect, thanks!

src/OrbitCore/SymbolHelper.cpp
```diff
@@ -199,3 +199,49 @@ fs::path SymbolHelper::GenerateCachedFileName(const fs::path& file_path) const {
...
+  if (checksum_or_error.has_error()) {
+    LOG("Unable to calculate checksum of \"%s\": \"%s\"", debuginfo_file_path.filename().string(),
+        checksum_or_error.error().message());
+    return false;
+  }
+
+  if (checksum_or_error.value() != checksum) {
+    LOG("Found file with matching name \"%s\", but the checksums do not match. Expected: %#x. "
+        "Actual: %#x",
+        debuginfo_file_path.string(), checksum, checksum_or_error.value());
```
dimitry-, line 224: also maybe add a test for this code path?

beckerhe: Yes, good point. https://github.com/google/orbit/pull/1929

PR #1926, opened 2021-02-23
Retry finalizing submits when querying for some timestamps fails

ronaldfw:
When finalizing submits it can happen that queries for timestamps can
fail when the timestamp is not yet ready to be queried. This aborted
with a CHECK for now, but needs to be worked around by re-enqueing the
submits where timestamp queries failed for later processing. This is
now fixed with this change.

Tested: Manually ran example application and capture multiple times.
Bug: http://179649352

### DIFF
--- a/src/OrbitVulkanLayer/SubmissionTracker.h
+++ b/src/OrbitVulkanLayer/SubmissionTracker.h
@@ -66,6 +66,12 @@ class SubmissionTracker : public VulkanLayerProducer::CaptureStatusListener {
   struct SubmittedCommandBuffer {
     std::optional<uint32_t> command_buffer_begin_slot_index;
     uint32_t command_buffer_end_slot_index;
+
+    // These are set when the timestamps were queried successfully. When these have values, the
+    // corresponding slot indices for timestamp queries have been reset. Do not query again in
+    // that case and only get the timestamps through these members.
+    std::optional<uint64_t> begin_timestamp;
+    std::optional<uint64_t> end_timestamp;
   };
 
   // A persistent version of a debug marker (either begin or end) that was submitted and its slot
@@ -78,6 +84,11 @@ class SubmissionTracker : public VulkanLayerProducer::CaptureStatusListener {
   struct SubmittedMarker {
     SubmissionMetaInformation meta_information;
     uint32_t slot_index;
+
+    // This is set when the timestamp was queried successfully. When this has a value, the
+    // corresponding slot index for the timestamp query has been reset. Do not query again in
+    // that case and only get the timestamp through this member.
+    std::optional<uint64_t> timestamp;
   };
 
   // Represents a color to be used to in debug markers. The values are all in range [0.f, 1.f].
@@ -109,6 +120,7 @@ class SubmissionTracker : public VulkanLayerProducer::CaptureStatusListener {
   // `completed_markers` are all the debug markers that got completed (via "End") within this
   // submission. Their "Begin" might still be in a different submission.
   struct QueueSubmission {
+    VkQueue queue;
     SubmissionMetaInformation meta_information;
     std::vector<SubmitInfo> submit_infos;
     std::vector<SubmittedMarkerSlice> completed_markers;
@@ -296,7 +308,7 @@ class SubmissionTracker : public VulkanLayerProducer::CaptureStatusListener {
   // It also takes a timestamp before the execution of the driver code for the submission.
   // This allows us to map submissions from the Vulkan layer to the driver submissions.
   [[nodiscard]] std::optional<QueueSubmission> PersistCommandBuffersOnSubmit(
-      uint32_t submit_count, const VkSubmitInfo* submits) {
+      VkQueue queue, uint32_t submit_count, const VkSubmitInfo* submits) {
     if (!IsCapturing()) {
       // `PersistDebugMarkersOnSubmit` and `OnCaptureFinished` will take care of clean up and
       // slot resetting.
@@ -304,6 +316,7 @@ class SubmissionTracker : public VulkanLayerProducer::CaptureStatusListener {
     }
 
     QueueSubmission queue_submission;
+    queue_submission.queue = queue;
     queue_submission.meta_information.pre_submission_cpu_timestamp =
         orbit_base::CaptureTimestampNs();
     queue_submission.meta_information.thread_id = orbit_base::GetCurrentThreadId();
@@ -413,7 +426,7 @@ class SubmissionTracker : public VulkanLayerProducer::CaptureStatusListener {
               // If there is a begin marker slot from a previous submission, this is our chance to
               // reset the slot.
               if (marker_state.begin_info.has_value() && !queue_submission_optional.has_value()) {
-                marker_slots_to_reset.push_back(marker_state.begin_info.value().slot_index);
+                marker_slots_to_reset.push_back(marker_state.begin_info->slot_index);
               }
 
               // If the begin marker was discarded for exceeding the maximum depth, but the end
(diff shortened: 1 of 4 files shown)


ronaldfw:
The approach here taken can be summarized as follows:

When completing current submits:
1) Query all timestamps and write those into the QueueSubmission data structure if successful. 
2) If any query fails, re-enqueu QueueSubmission to try again later. 
3) For all QueueSubmission's where all timestamps were queried successfully, send the data out of the layer.

ronaldfw:
@florian-kuebler Can you advise on where and if I need to adjust the tests?

src/OrbitVulkanLayer/SubmissionTracker.h
```diff
@@ -481,17 +500,29 @@ class SubmissionTracker : public VulkanLayerProducer::CaptureStatusListener {
...
-    for (const auto& completed_submission : completed_submissions) {
+    std::vector<QueueSubmission> submissions_to_send = {};
+
+    for (auto& completed_submission : completed_submissions) {
+      bool command_buffer_queries_succeeded = QueryCommandBufferTimestamps(
+          &completed_submission, &query_slots_to_reset, device, query_pool, timestamp_period);
+      bool marker_queries_succeeded = QueryDebugMarkerTimestamps(
+          &completed_submission, &query_slots_to_reset, device, query_pool, timestamp_period);
+
+      if (command_buffer_queries_succeeded && marker_queries_succeeded) {
```
florian-kuebler, line 511: As written offline, I am a bit concerned that this change breaks the "ordering" property of the event's we sent to the UI, i.e. vulkan events of a specific queue are ordered per submission cpu timestamp. 
I gave this another thought (what it would mean to the client and if we can work around), but I guess the best would be to keep this property on the producer/layer side. And it is actually quite easy to adjust here.
Currently, PullCompletedSubmissions returns a sorted list. So we are iterating here in order. We could simply give up as soon as we find a submission that is incomplete (i.e. add another bool here, set this in the else case, and then in the following iterations always push the submissions to the retry block).

src/OrbitVulkanLayer/SubmissionTracker.h
```diff
@@ -98,6 +104,12 @@ class SubmissionTracker : public VulkanLayerProducer::CaptureStatusListener {
     std::string label_name;
     Color color;
     size_t depth;
+
+    // These are set when the timestamps were queried successfully. When these have values, the
+    // corresponding slot indices for timestamp queries have been reset. Do not query again in
+    // that case and only get the timestamps through these members.
+    std::optional<uint64_t> begin_timestamp;
```
florian-kuebler, line 111: Nit: I am wondering if we should put the optional timestamp into the `SubmittedMarker`. Note that the state begin_info == nullopt, begin_timestamp == value would be invalid.

src/OrbitVulkanLayer/SubmissionTracker.h
```diff
@@ -472,6 +484,13 @@ class SubmissionTracker : public VulkanLayerProducer::CaptureStatusListener {
     std::vector<QueueSubmission> completed_submissions =
         PullCompletedSubmissions(device, query_pool);
 
+    // Completing a submission is not possible when a timestamp is not yet ready to be queried (and
+    // this is in the hands of Vulkan). We therefore maintain a list of submissions to be retried
+    // later and append this list here for the current round.
+    completed_submissions.insert(completed_submissions.end(), retry_to_complete_.begin(),
```
florian-kuebler, line 490: Is there are reason to have `retry_to_complete_` as a member and add them to the completed submissions at the beginning of the method here?
I'd guess it would be a bit more straightforward to have a local vector here, and add it to the `queue_to_submissions` at the end of the method?

florian-kuebler (commented):
Thank you so much for jumping in here. In general looks good, but it would be great to consider keeping the ordering property. Also I don't feel we should have the submissions to retry as a member. But I hope both should be easy to address.

Regarding the tests:
If I am not wrong, this change should not break any existing test, which is great!
In order to write a test for the breaking scenario you could probably take a look e.g. at `CanRetrieveDebugMarkerAcrossTwoSubmissions` as well as the `CanRetrieveCommandBufferTimestampsForACompleteSubmissionAtALaterTime` in SubmissionTrackerTest.
Probably a simple case with two submissions, each containing a single command buffer should work. Ideally we would have a second test where the debug markers are not ready, but not sure if we really need this.

src/OrbitVulkanLayer/SubmissionTracker.h
ronaldfw: Yes. The issue is that queue_to_submissions_ is a map from queue to submissions and at this point I don't know the queue anymore. We could plumb this through in PullCompletedSubmissions, etc., but using this additional vector seemed to be the easier solution.

src/OrbitVulkanLayer/SubmissionTracker.h
ronaldfw: Yes, I think that would make sense.

src/OrbitVulkanLayer/SubmissionTracker.h
ronaldfw: Yes, that seems like a simple fix. I'll change that.

src/OrbitVulkanLayer/SubmissionTracker.h
florian-kuebler: Could we not just add the queue to the `QueueSubmission` struct? I guess that would be cleaner.

src/OrbitVulkanLayer/SubmissionTracker.h
florian-kuebler: Thanks!

src/OrbitVulkanLayer/SubmissionTracker.h
```diff
@@ -84,6 +84,12 @@ class SubmissionTracker : public VulkanLayerProducer::CaptureStatusListener {
   struct SubmittedMarker {
     SubmissionMetaInformation meta_information;
     uint32_t slot_index;
+
+    // These are set when the timestamps were queried successfully. When these have values, the
+    // corresponding slot indices for timestamp queries have been reset. Do not query again in
+    // that case and only get the timestamps through these members.
+    std::optional<uint64_t> begin_timestamp;
```
florian-kuebler, line 91: I guess with the new change, you only need one timestamp here.
A slice `SubmittedMarker`s for both start and end (well start is optional...). So in the code below you are always accessing either end_marker.end_timestamp or begin_marker->begin_timestamp.

florian-kuebler (approved):
Thanks for addressing the comments. I feel it is a bit cleaner now. LGTM. Feel free to omit the tests for this PR and hand them over to me. However, if you are going to write tests, just re-request a review.

src/OrbitVulkanLayer/SubmissionTracker.h
ronaldfw: Ah, you are right, will fix right away.

src/OrbitVulkanLayer/SubmissionTracker.h
ronaldfw: Should be fixed now.

src/OrbitVulkanLayer/SubmissionTracker.h
ronaldfw: Done.

src/OrbitVulkanLayer/SubmissionTracker.h
ronaldfw: Done.

src/OrbitVulkanLayer/SubmissionTracker.h
ronaldfw: Done.

ronaldfw:
I verified that the current unit tests still pass and will merge this once the CI is done.

I will follow-up with additional unit tests that isolate the issue fixed here.

PR #1927, opened 2021-02-24
Add LockFreeApiEventProducer

pierricgimmig:
This is another piece of the new manual instrumentation change that uses the side-channel
to send event from the target process to OrbitService. In this PR, we define a simple ApiEvent 
capture event that wraps raw orbit_api::ApiEvent's. 

The LockFreeApiEventProducer and LockFreeApiEventBulkProducer are two variants of event
producers that will be used to compare the performance of different approaches for sending
events. They both push raw api events to a lockless queue from multiple threads and have a single
consumer thread that translates the raw events into capture events (protobufs). The difference is that
the "Bulk" version wraps multiple raw events into a single CaptureEvent to amortize the cost of 
creating protobufs. The goal will be to iterate on the non-bulk implementation to see how close we
can get in terms of performance to the bulk approach.

This PR focuses on the production and sending of events from the target process to OrbitService. 
The next PR will actually process those events.

Note that LockFreeBufferBulkCaptureEventProducer.h is basically a copy of the existing LockFreeBufferCaptureEventProducer.h with the minor modification of allowing to translate multiple
intermediate events at once. Maybe both classes could be merged as this could be an optimization 
even for the normal use case (instead of calling translate function once per event, we could call it
once for all the dequeued events).

### DIFF
--- a/src/Api/CMakeLists.txt
+++ b/src/Api/CMakeLists.txt
@@ -19,6 +19,7 @@ target_include_directories(Api PUBLIC
 
 target_sources(Api PUBLIC
         include/Api/EncodedEvent.h
+        include/Api/LockFreeApiEventProducer.h
         include/Api/Orbit.h)
 
 target_sources(Api PRIVATE
--- a/src/Api/Orbit.cpp
+++ b/src/Api/Orbit.cpp
@@ -6,21 +6,20 @@
 #include "Api/Orbit.h"
 
 #include "Api/EncodedEvent.h"
+#include "Api/LockFreeApiEventProducer.h"
 #include "OrbitBase/Profiling.h"
 #include "OrbitBase/ThreadUtils.h"
 
-// Temporary dummy function, proper implementation is in next PR.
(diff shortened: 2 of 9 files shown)


dpallotti (commented):
Long story short: I'm strongly against this change.
It adds two implementations of the same class in `LockFreeApiEventProducer.h` and a second almost identical (and not tested) implementation of `LockFreeBufferCaptureEventProducer`. 
Performance investigation should be done first, and then we should decide how to proceed. Plus bulking multiple events in a single `ApiEvent` is something I've already expressed I'm against as it basically bypasses the whole idea of protobufs. Have you tried getting rid of `bytes` in `ApiEvent` and comparing the performance achieved with the new `Arena` allocation to your solution? Let's try that first, let's benchmark carefully, and then let's see which way to go.
We should follow up offline.

pierricgimmig:
@dpallotti , thanks for the input, let's take it offline as you suggested!

PR #1929, opened 2021-02-24
Add test for SymbolHelper::IsMatchingDebugInfoFile

beckerhe:


src/OrbitCore/SymbolHelper.h
```diff
@@ -33,6 +33,7 @@ class SymbolHelper {
 
   [[nodiscard]] fs::path GenerateCachedFileName(const fs::path& file_path) const;
 
+  [[nodiscard]] static bool IsMatchingDebugInfoFile(const fs::path& file_path, uint32_t checksum);
```
dimitry-, line 36: probably did not mean to make this static

dimitry- (approved):
otherwise lgtm

src/OrbitCore/SymbolHelper.h
beckerhe: Hmm why not? It does not need access to any member variables. I could also make it a free function if you prefer that?

dimitry-: nevermind, missed the part where you moved it to SymbolHelper class.

PR #1930, opened 2021-02-24
Follow .gnu_debuglink when looking for DWARF information

beckerhe:
This ensures that we follow the .gnu_debuglink section when the original module contains symbols, but no debug (DWARF) information.

### DIFF
--- a/src/OrbitGl/App.cpp
+++ b/src/OrbitGl/App.cpp
@@ -1223,6 +1223,52 @@ orbit_base::Future<ErrorMessageOr<std::filesystem::path>> OrbitApp::RetrieveModu
   return final_result;
 }
 
+orbit_base::Future<ErrorMessageOr<std::filesystem::path>> OrbitApp::RetrieveModuleWithDebugInfo(
+    const ModuleData* module_data) {
+  return RetrieveModuleWithDebugInfo(module_data->file_path(), module_data->build_id());
+}
+
+orbit_base::Future<ErrorMessageOr<std::filesystem::path>> OrbitApp::RetrieveModuleWithDebugInfo(
+    const std::string& module_path, const std::string& build_id) {
+  auto loaded_module = RetrieveModule(module_path, build_id);
+  return loaded_module.ThenIfSuccess(
+      main_thread_executor_,
+      [this, module_path](
+          const std::filesystem::path& local_file_path) -> ErrorMessageOr<std::filesystem::path> {
+        auto elf_file = orbit_elf_utils::ElfFile::Create(local_file_path);
+
+        if (elf_file.has_error()) return elf_file.error();
+
+        if (elf_file.value()->HasDebugInfo()) return local_file_path;
+
+        if (!elf_file.value()->HasGnuDebuglink()) {
+          return ErrorMessage{
+              absl::StrFormat("Module \"%s\" neither includes debug info, nor does it contain "
+                              "a .gnu_debuglink section which could refer to a separate debug "
+                              "info file.",
+                              module_path)};
+        }
+
+        const auto debuglink = elf_file.value()->GetGnuDebugLinkInfo().value();
+        ErrorMessageOr<std::filesystem::path> local_debuginfo_path =
+            symbol_helper_.FindDebugInfoFileLocally(debuglink.path.filename().string(),
+                                                    debuglink.crc32_checksum);
+        if (local_debuginfo_path.has_error()) {
+          return ErrorMessage{absl::StrFormat(
+              "Module \"%s\" doesn't include debug info, and a separate "
+              "debuginfo file wasn't found on this machine, when searching "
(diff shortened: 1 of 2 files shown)


src/OrbitGl/App.cpp
```diff
@@ -1198,6 +1189,53 @@ orbit_base::Future<ErrorMessageOr<std::filesystem::path>> OrbitApp::RetrieveModu
...
+          ErrorMessageOr<std::filesystem::path> local_debuginfo_path =
+              symbol_helper_.FindDebugInfoFileLocally(debuglink.path.filename().string(),
+                                                      debuglink.crc32_checksum);
+          if (local_debuginfo_path.has_error()) {
+            return ErrorMessage{absl::StrFormat(
+                "Module \"%s\" doesn't include debug info, and a separate "
+                "debuginfo file wasn't found on this machine, when searching "
+                "the paths from your SymbolsPath.txt. Please make sure the "
+                "debuginfo file can be found in one of the listed directories. According to "
+                "the .gnu_debuglink section, the debuginfo file must be called \"%s\"",
```
dpallotti, line 1227:
```suggestion
                "the .gnu_debuglink section, the debuginfo file must be called \"%s\".",
```
As this probably ends up in a message dialog.

src/OrbitGl/App.h
```diff
@@ -306,6 +306,13 @@ class OrbitApp final : public DataViewFactory, public CaptureListener {
   orbit_base::Future<ErrorMessageOr<void>> RetrieveModuleAndLoadSymbols(
       const std::string& module_path, const std::string& build_id);
 
+  // This method is pretty similar to `RetrieveModule`, but it also requires debug information to be
+  // present. `RetrieveModule` will be satisfied with a symbol section.
```
dpallotti, line 310: I don't understand the sentence "`RetrieveModule` will be satisfied with a symbol section."

dpallotti (approved):

src/OrbitGl/App.cpp
```diff
@@ -1198,6 +1189,53 @@ orbit_base::Future<ErrorMessageOr<std::filesystem::path>> OrbitApp::RetrieveModu
...
+orbit_base::Future<ErrorMessageOr<std::filesystem::path>> OrbitApp::RetrieveModuleWithDebugInfo(
+    const std::string& module_path, const std::string& build_id) {
+  auto loaded_module = RetrieveModule(module_path, build_id);
+  return loaded_module.ThenIfSuccess(
+      main_thread_executor_,
+      [this, module_path](
+          const std::filesystem::path& local_file_path) -> ErrorMessageOr<std::filesystem::path> {
+        auto elf_file = orbit_elf_utils::ElfFile::Create(local_file_path);
+
+        if (elf_file.has_error()) return elf_file.error();
```
antonrohr, line 1206: OUTCOME_TRY(elf_file, orbit_elf_utils::ElfFile::Create(local_file_path));

src/OrbitGl/App.cpp
```diff
@@ -1198,6 +1189,53 @@ orbit_base::Future<ErrorMessageOr<std::filesystem::path>> OrbitApp::RetrieveModu
...
+                "debuginfo file can be found in one of the listed directories. According to "
+                "the .gnu_debuglink section, the debuginfo file must be called \"%s\"",
+                module_path, debuglink.path.string())};
+          }
+
+          elf_file = orbit_elf_utils::ElfFile::Create(local_debuginfo_path.value());
+          if (elf_file.has_error()) return elf_file.error();
+        }
+
+        return local_file_path;
```
antonrohr, line 1235: I think here is something wrong
Either at the end you need to `return elf_file.value().file_path();` or 
```
if (elf_file.has_error()) {
  return elf_file.error();
} else {
  return local_debuginfo_path.value();
}
```

beckerhe: Yes, good catch

src/OrbitGl/App.cpp
beckerhe: I'd rather not use OUTCOME_TRY before we haven't figured out if we move to absl::StatusOr

src/OrbitGl/App.cpp
beckerhe: Done.

src/OrbitGl/App.h
beckerhe: Yes, this was weird. I removed it and adjusted the comment for `RetrieveModule` accordingly. It should be clearer now.

src/OrbitGl/App.cpp
beckerhe: Hopefully fixed. PTAL!

antonrohr: LGreatTM. I like the change a lot (all the other early returns)

antonrohr (approved):
LGTM. Thank you for putting in the changes!

PR #1931, opened 2021-02-24
Make usage of directory_iterators exception-free

dimitry-:
This change also handles it.increment failure by stopping
iterations instead of retrying them.

Test: Start Orbit -> connect to instance, check that presets are populated

### DIFF
--- a/src/OrbitGl/App.cpp
+++ b/src/OrbitGl/App.cpp
@@ -387,13 +387,38 @@ static std::vector<std::filesystem::path> ListRegularFilesWithExtension(
     const std::filesystem::path& directory, std::string_view extension) {
   std::vector<std::filesystem::path> files;
 
-  for (const auto& file : std::filesystem::directory_iterator(directory)) {
-    if (std::filesystem::is_regular_file(file)) {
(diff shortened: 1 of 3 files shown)


src/OrbitGl/App.cpp
```diff
@@ -387,13 +387,38 @@ static std::vector<std::filesystem::path> ListRegularFilesWithExtension(
...
+    if (error) {
+      ERROR("Iterating directory \"%s\": %s (increment failed, stopping)", directory.string(),
+            error.message());
+      break;
+    }
+
+    const auto& path = it->path();
+    bool is_regular_file = std::filesystem::is_regular_file(path, error);
+    if (error) {
+      ERROR("Unable to stat \"%s\": %s (ingnoring)", path.string(), error.message());
```
dpallotti, line 409: ingnoring -> ignoring

dpallotti (approved):

src/OrbitGl/App.cpp
dimitry-: done.

PR #1932, opened 2021-02-24
Refactor ReadTracepoints - make it exception-free

dimitry-:
Test: run ServiceTests under root

### DIFF
--- a/src/Service/ServiceUtils.cpp
+++ b/src/Service/ServiceUtils.cpp
@@ -41,36 +41,63 @@ using ::orbit_elf_utils::ElfFile;
 using orbit_grpc_protos::ModuleInfo;
 using orbit_grpc_protos::TracepointInfo;
 
-static const char* kLinuxTracingEvents = "/sys/kernel/debug/tracing/events/";
+static const char* kLinuxTracingEventsDirectory = "/sys/kernel/debug/tracing/events/";
 
 ErrorMessageOr<std::vector<orbit_grpc_protos::TracepointInfo>> ReadTracepoints() {
   std::vector<TracepointInfo> result;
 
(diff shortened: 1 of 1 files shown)


src/Service/ServiceUtils.cpp
```diff
@@ -41,36 +41,57 @@ using ::orbit_elf_utils::ElfFile;
...
+    }
+
+    for (auto it = fs::begin(name_directory_iterator), end = fs::end(name_directory_iterator);
+         it != end; it.increment(error)) {
+      if (error) {
+        return ErrorMessage{absl::StrFormat("Unable to scan \"%s\" directory: %s",
+                                            category_path.string(), error.message())};
+      }
+
+      if (it->path().filename() == "enable" || it->path().filename() == "filter") {
```
dimitry-, line 86: looks like this should be check for is_directory instead.

dimitry-: done.

danielfenner (approved):
lgtm!

src/Service/ServiceUtils.cpp
danielfenner: I think this is wrong - we need to filter out the "enable" and "filter" files.

dimitry-: but they are files... we are not counting files, only directories. So if Linux decides to add a file in the category directory - we are still fine.

danielfenner: Oh, I see. You are right: "enable" and "filter" are currently the only files here - so ingoring files is equivalent to the current behaviour and potentially more stable.

danielfenner (approved):
lgtm!

PR #1934, opened 2021-02-24
Properly set max local debug marker depth in tests.

florian-kuebler:
A recent change introduced setting the maximal local marker depth
per command buffer via the capture options.
This change was not reflected in the tests.

Test: Run tests.

### DIFF
--- a/src/OrbitVulkanLayer/SubmissionTrackerTest.cpp
+++ b/src/OrbitVulkanLayer/SubmissionTrackerTest.cpp
@@ -53,10 +53,14 @@ class MockVulkanLayerProducer : public VulkanLayerProducer {
 
   MOCK_METHOD(void, SetCaptureStatusListener, (CaptureStatusListener*), (override));
 
-  void StartCapture() {
+  void StartCapture(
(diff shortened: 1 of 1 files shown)


src/OrbitVulkanLayer/SubmissionTrackerTest.cpp
```diff
@@ -1388,7 +1390,7 @@ TEST_F(SubmissionTrackerTest, CanLimitNestedDebugMarkerDepthPerCommandBuffer) {
 
   Color expected_color{1.f, 0.8f, 0.6f, 0.4f};
 
-  producer_->StartCapture();
+  producer_->StartCapture(1);
```
ronaldfw, line 1393: Nit: Maybe define a constant kMaxDepth = 1 to indicate the meaning of "1" here.

ronaldfw (approved):
LGTM

src/OrbitVulkanLayer/SubmissionTrackerTest.cpp
florian-kuebler: Done.

PR #1935, opened 2021-02-24
Delete operator bool for ErrorMessageOr<bool>

dimitry-:
Test: builds

### DIFF
--- a/src/OrbitBase/include/OrbitBase/Result.h
+++ b/src/OrbitBase/include/OrbitBase/Result.h
@@ -25,7 +25,18 @@ template <typename T, typename E>
 using Result = outcome::result<T, E, outcome::policy::terminate>;
 
 template <typename T>
-using ErrorMessageOr = Result<T, ErrorMessage>;
+class ErrorMessageOr : public Result<T, ErrorMessage> {
+ public:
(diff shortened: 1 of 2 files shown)


src/OrbitBase/include/OrbitBase/Result.h
```diff
@@ -25,7 +25,17 @@ template <typename T, typename E>
...
+class ErrorMessageOr : public Result<T, ErrorMessage> {
+ public:
+  using Result<T, ErrorMessage>::Result;
+};
+
+template <>
+class ErrorMessageOr<bool> : public Result<bool, ErrorMessage> {
+ public:
+  using Result<bool, ErrorMessage>::Result;
+  operator bool() = delete;
```
beckerhe, line 37: Is that also deleting the const overload? Probably. I honestly don't know.

beckerhe (approved):
That was fast. Thanks a lot! LGTM!

danielfenner (approved):
lgtm

src/OrbitBase/include/OrbitBase/Result.h
dimitry-: It is according to errors I am getting but I explicitly deleted const variant anyways

PR #1936, opened 2021-02-24
Backup, modify and restore register state.

danielfenner:
Register state of a halted thread can be read, modified, and written
back to the halted thread.

bug: http://b/181117089
test: unit test.

### DIFF
--- a/src/UserSpaceInstrumentation/RegisterState.cpp
+++ b/src/UserSpaceInstrumentation/RegisterState.cpp
@@ -1,0 +1,140 @@
+// Copyright (c) 2021 The Orbit Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#include "UserSpaceInstrumentation/RegisterState.h"
+
+#include <cpuid.h>
+#include <elf.h>
+#include <errno.h>
+#include <sys/ptrace.h>
+#include <sys/uio.h>
+
+#include <string>
+
+#include "OrbitBase/Logging.h"
+#include "OrbitBase/SafeStrerror.h"
+#include "absl/strings/str_format.h"
+
+namespace orbit_user_space_instrumentation {
+
+namespace {
+
+// Return the size of the XSave area on this cpu.
+[[nodiscard]] ErrorMessageOr<size_t> GetXSaveAreaSize() {
+  uint32_t eax = 0;
+  uint32_t ebx = 0;
+  uint32_t ecx = 0;
+  uint32_t edx = 0;
+  if (!__get_cpuid(1, &eax, &ebx, &ecx, &edx) || !(ecx & bit_XSAVE)) {
+    return ErrorMessage("XSAVE is not supported by Cpu.");
+  }
+  if (!__get_cpuid_count(0x0d, 0x00, &eax, &ebx, &ecx, &edx)) {
+    return ErrorMessage("XSAVE is not supported by Cpu.");
+  }
+  return static_cast<size_t>(ecx);
+}
+
+// Return offset of the YMMx registers inside the extended section of the XSave area.
+[[nodiscard]] ErrorMessageOr<size_t> GetAvxOffset() {
+  uint32_t eax = 0;
+  uint32_t ebx = 0;
+  uint32_t ecx = 0;
+  uint32_t edx = 0;
+  if (!__get_cpuid(0x01, &eax, &ebx, &ecx, &edx) || !(ecx & bit_AVX)) {
+    return ErrorMessage("AVX is not supported by Cpu.");
+  }
+  if (!__get_cpuid_count(0x0d, 0x02, &eax, &ebx, &ecx, &edx)) {
+    return ErrorMessage("AVX offset query failed.");
+  }
+  return static_cast<size_t>(ebx);
+}
+
+}  // namespace
+
+bool RegisterState::Hasx87DataStored() {
+  return (static_cast<uint64_t>(GetXSaveHeader()->xstate_bv) &
+          static_cast<uint64_t>(XSaveHeader::StateComponents::kX87)) != 0;
+}
+
+bool RegisterState::HasSseDataStored() {
+  return (static_cast<uint64_t>(GetXSaveHeader()->xstate_bv) &
+          static_cast<uint64_t>(XSaveHeader::StateComponents::kSse)) != 0;
+}
+
+bool RegisterState::HasAvxDataStored() {
+  return (static_cast<uint64_t>(GetXSaveHeader()->xstate_bv) &
+          static_cast<uint64_t>(XSaveHeader::StateComponents::kAvx)) != 0;
+}
+
+[[nodiscard]] ErrorMessageOr<void> RegisterState::BackupRegisters(pid_t tid) {
+  tid_ = tid;
+
+  iovec iov;
+  iov.iov_base = &general_purpose_registers_;
+  iov.iov_len = sizeof(GeneralPurposeRegisters);
+  auto result = ptrace(PTRACE_GETREGSET, tid, NT_PRSTATUS, &iov);
+  if (result == -1) {
+    return ErrorMessage(absl::StrFormat("PTRACE_GETREGS, NT_PRSTATUS failed with errno: %d: \"%s\"",
+                                        errno, SafeStrerror(errno)));
+  }
+
+  if (iov.iov_len == sizeof(GeneralPurposeRegisters32)) {
+    bitness_ = Bitness::k32Bit;
+  } else if (iov.iov_len == sizeof(GeneralPurposeRegisters64)) {
+    bitness_ = Bitness::k64Bit;
+  } else {
+    CHECK(false);
+  }
+
+  auto xsave_area_size = GetXSaveAreaSize();
+  if (xsave_area_size.has_error()) {
+    return xsave_area_size.error();
+  }
+  xsave_area_.resize(xsave_area_size.value());
+
+  iov.iov_len = xsave_area_.size();
+  iov.iov_base = xsave_area_.data();
+  result = ptrace(PTRACE_GETREGSET, tid, NT_X86_XSTATE, &iov);
+  if (result == -1) {
+    return ErrorMessage(absl::StrFormat(
+        "PTRACE_GETREGS, NT_X86_XSTATE failed with errno: %d: \"%s\"", errno, SafeStrerror(errno)));
+  }
+
+  auto avx_offset = GetAvxOffset();
+  if (!avx_offset.has_error()) {
+    avx_offset_ = avx_offset.value();
+  } else {
+    avx_offset_ = 0;
+  }
+
+  return outcome::success();
+}
+
+[[nodiscard]] ErrorMessageOr<void> RegisterState::RestoreRegisters() {
+  // BackupRegisters needs to be called before RestoreRegisters.
+  CHECK(tid_ != -1);
+
+  iovec iov;
+  iov.iov_base = &general_purpose_registers_;
+  iov.iov_len = bitness_ == Bitness::k32Bit ? sizeof(GeneralPurposeRegisters32)
+                                            : sizeof(GeneralPurposeRegisters64);
+  auto result = ptrace(PTRACE_SETREGSET, tid_, NT_PRSTATUS, &iov);
+  if (result == -1) {
+    return ErrorMessage(
+        absl::StrFormat("PTRACE_SETREGSET failed to write NT_PRSTATUS with errno: %d: \"%s\"",
+                        errno, SafeStrerror(errno)));
+  }
+
+  iov.iov_len = xsave_area_.size();
+  iov.iov_base = xsave_area_.data();
+  result = ptrace(PTRACE_SETREGSET, tid_, NT_X86_XSTATE, &iov);
+  if (result == -1) {
+    ErrorMessage(
+        absl::StrFormat("PTRACE_SETREGSET failed to write NT_X86_XSTATE with errno: %d: \"%s\"",
+                        errno, SafeStrerror(errno)));
+  }
+  return outcome::success();
+}
+
+}  // namespace orbit_user_space_instrumentation
--- a/src/UserSpaceInstrumentation/RegisterState.cpp
+++ b/src/UserSpaceInstrumentation/RegisterState.cpp
@@ -21,11 +21,12 @@ namespace orbit_user_space_instrumentation {
 namespace {
 
 // Some notes reguarding the calls to cpuid below:
-// Cpuid takes one parameter in eax. In Intel's terminology, this is called the Cpuid "leaf". Some
-// leaves have "sub-leafs" i.e. they take a second paramter in ecx. The sub-leaf is sometimes called
-// "count". The return values end up in registers eax, ... , edx.
-// The wrappers from cpuid.h simplify error and parameter handling. cpuid.h also has some defines
-// and useful comments to figure out what can be queried. More comprehensive info:
+// Cpuid can be used to query all sorts of information about the cpu (presence of features,
+// specifications, ...). Cpuid takes one parameter in eax. In Intel's terminology, this is called
+// the Cpuid "leaf". Some leaves have "sub-leafs" i.e. they take a second paramter in ecx. The
+// sub-leaf is sometimes called "count". The return values end up in registers eax, ... , edx. The
+// wrappers from cpuid.h simplify error and parameter handling. cpuid.h also has some defines and
+// useful comments to figure out what can be queried. More comprehensive info:
 // https://www.sandpile.org/x86/cpuid.htm
 
 // Return the size of the XSave area on this cpu.
--- a/src/UserSpaceInstrumentation/RegisterState.cpp
+++ b/src/UserSpaceInstrumentation/RegisterState.cpp
@@ -20,17 +20,25 @@ namespace orbit_user_space_instrumentation {
 
 namespace {
 
+// Some notes reguarding the calls to cpuid below:
+// Cpuid takes one parameter in eax. In Intel's terminology, this is called the Cpuid "leaf". Some
+// leaves have "sub-leafs" i.e. they take a second paramter in ecx. The sub-leaf is sometimes called
+// "count". The return values end up in registers eax, ... , edx.
+// The wrappers from cpuid.h simplify error and parameter handling. cpuid.h also has some defines
+// and useful comments to figure out what can be queried. More comprehensive info:
+// https://www.sandpile.org/x86/cpuid.htm
+
 // Return the size of the XSave area on this cpu.
 [[nodiscard]] ErrorMessageOr<size_t> GetXSaveAreaSize() {
   uint32_t eax = 0;
   uint32_t ebx = 0;
   uint32_t ecx = 0;
   uint32_t edx = 0;
-  if (!__get_cpuid(1, &eax, &ebx, &ecx, &edx) || !(ecx & bit_XSAVE)) {
-    return ErrorMessage("XSAVE is not supported by Cpu.");
+  if (!__get_cpuid(0x01, &eax, &ebx, &ecx, &edx) || !(ecx & bit_XSAVE)) {
+    return ErrorMessage("XSAVE is not supported by the Cpu.");
   }
   if (!__get_cpuid_count(0x0d, 0x00, &eax, &ebx, &ecx, &edx)) {
-    return ErrorMessage("XSAVE is not supported by Cpu.");
+    return ErrorMessage("XSAVE is not supported by the Cpu.");
   }
   return static_cast<size_t>(ecx);
 }
@@ -42,7 +50,7 @@ namespace {
   uint32_t ecx = 0;
   uint32_t edx = 0;
   if (!__get_cpuid(0x01, &eax, &ebx, &ecx, &edx) || !(ecx & bit_AVX)) {
-    return ErrorMessage("AVX is not supported by Cpu.");
+    return ErrorMessage("AVX is not supported by the Cpu.");
   }
   if (!__get_cpuid_count(0x0d, 0x02, &eax, &ebx, &ecx, &edx)) {
     return ErrorMessage("AVX offset query failed.");
@@ -67,7 +75,7 @@ bool RegisterState::HasAvxDataStored() {
           static_cast<uint64_t>(XSaveHeader::StateComponents::kAvx)) != 0;
 }
 
-[[nodiscard]] ErrorMessageOr<void> RegisterState::BackupRegisters(pid_t tid) {
+ErrorMessageOr<void> RegisterState::BackupRegisters(pid_t tid) {
   tid_ = tid;
 
   iovec iov;
@@ -84,7 +92,7 @@ bool RegisterState::HasAvxDataStored() {
   } else if (iov.iov_len == sizeof(GeneralPurposeRegisters64)) {
     bitness_ = Bitness::k64Bit;
   } else {
-    CHECK(false);
+    FATAL("Bitness is neither 32 or 64 bit.");
   }
 
   auto xsave_area_size = GetXSaveAreaSize();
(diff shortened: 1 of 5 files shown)


src/UserSpaceInstrumentation/include/UserSpaceInstrumentation/RegisterState.h
```diff
@@ -0,0 +1,240 @@
...
+  union {
+    std::array<uint8_t, 10> bytes;
+    MmsAs80BitFloat as_float;
+  };
+  std::array<uint8_t, 6> reserved;
+};
+static_assert(sizeof(MmsRegister) == 16, "MmsRegister is not 16 bytes of size");
+
+struct XmmRegister {
+  uint8_t bytes[16];
```
dpallotti, line 95: Is there a reason why you use `std::array` above but a "normal" array here?

src/UserSpaceInstrumentation/include/UserSpaceInstrumentation/RegisterState.h
```diff
@@ -0,0 +1,240 @@
...
+
+// The upper 128 bit of the YMM0..15 registers. The lower bits are shared with XMM registers above.
+struct YmmHi {
+  std::array<YmmHiRegister, 16> ymm;
+};
+PACKED_END
+
+// Backup, modify and restore register state of a halted thread.
+//
+// Requires the XSave extension (added around 2008) to be supported by the Cpu otherwise
```
dpallotti, line 152: I think that the piece missing in all the comments here is a simple definition of the "XSave area". One gets the idea but it would be clearer to have an explanation right away.

src/UserSpaceInstrumentation/RegisterState.cpp
```diff
@@ -0,0 +1,140 @@
...
+
+namespace {
+
+// Return the size of the XSave area on this cpu.
+[[nodiscard]] ErrorMessageOr<size_t> GetXSaveAreaSize() {
+  uint32_t eax = 0;
+  uint32_t ebx = 0;
+  uint32_t ecx = 0;
+  uint32_t edx = 0;
+  if (!__get_cpuid(1, &eax, &ebx, &ecx, &edx) || !(ecx & bit_XSAVE)) {
```
dpallotti, line 29: Could you add a comment somewhere on what `__get_cpuid` and `__get_cpuid_count` do and how they are used?

src/UserSpaceInstrumentation/RegisterState.cpp
```diff
@@ -0,0 +1,140 @@
...
+    return ErrorMessage("XSAVE is not supported by Cpu.");
```
dpallotti, line 30: by Cpu -> by the Cpu (here and below)?

src/UserSpaceInstrumentation/RegisterState.cpp
```diff
@@ -0,0 +1,140 @@
...
+  }
+  if (!__get_cpuid_count(0x0d, 0x00, &eax, &ebx, &ecx, &edx)) {
+    return ErrorMessage("XSAVE is not supported by Cpu.");
```
dpallotti, line 33: This error message is the same as the one above, but in `GetAvxOffset` the error message for `__get_cpuid_count` is different. Is this on purpose?

src/UserSpaceInstrumentation/RegisterState.cpp
```diff
@@ -0,0 +1,140 @@
...
+  return (static_cast<uint64_t>(GetXSaveHeader()->xstate_bv) &
+          static_cast<uint64_t>(XSaveHeader::StateComponents::kSse)) != 0;
+}
+
+bool RegisterState::HasAvxDataStored() {
+  return (static_cast<uint64_t>(GetXSaveHeader()->xstate_bv) &
+          static_cast<uint64_t>(XSaveHeader::StateComponents::kAvx)) != 0;
+}
+
+[[nodiscard]] ErrorMessageOr<void> RegisterState::BackupRegisters(pid_t tid) {
```
dpallotti, line 70: One thing I find slightly weird of the interface of this class is that the `tid` is passed to `BackupResisters` but not to `RestoreRegisters`, instead the `tid_` stored by `BackupRegisters` is the used in `RestoreRegisters`. I can see the point but I would find it more consistent if both methods accepted the `tid` instead of storing it in the class. After all, `tid_` is not conceptually part of the "register state".

src/UserSpaceInstrumentation/RegisterState.cpp
```diff
@@ -0,0 +1,140 @@
...
+  tid_ = tid;
+
+  iovec iov;
+  iov.iov_base = &general_purpose_registers_;
+  iov.iov_len = sizeof(GeneralPurposeRegisters);
+  auto result = ptrace(PTRACE_GETREGSET, tid, NT_PRSTATUS, &iov);
+  if (result == -1) {
+    return ErrorMessage(absl::StrFormat("PTRACE_GETREGS, NT_PRSTATUS failed with errno: %d: \"%s\"",
+                                        errno, SafeStrerror(errno)));
```
dpallotti, line 79: Isn't `SafeStrerror(errno)` enough in this message (and below)? Do you really want the exact number?

src/UserSpaceInstrumentation/RegisterState.cpp
```diff
@@ -0,0 +1,140 @@
...
+  }
+
+  if (iov.iov_len == sizeof(GeneralPurposeRegisters32)) {
+    bitness_ = Bitness::k32Bit;
+  } else if (iov.iov_len == sizeof(GeneralPurposeRegisters64)) {
+    bitness_ = Bitness::k64Bit;
+  } else {
+    CHECK(false);
```
dpallotti, line 87: Use `UNREACHABLE()`, or `FATAL` with a message.

src/UserSpaceInstrumentation/RegisterState.cpp
```diff
@@ -0,0 +1,140 @@
...
+  if (result == -1) {
+    return ErrorMessage(absl::StrFormat(
+        "PTRACE_GETREGS, NT_X86_XSTATE failed with errno: %d: \"%s\"", errno, SafeStrerror(errno)));
+  }
+
+  auto avx_offset = GetAvxOffset();
+  if (!avx_offset.has_error()) {
+    avx_offset_ = avx_offset.value();
+  } else {
+    avx_offset_ = 0;
```
dpallotti, line 108: Is `0` the real value in case of error or is this an assumption? In other words, why isn't an `ErrorMessage` returned here instead?

src/UserSpaceInstrumentation/RegisterState.cpp
```diff
@@ -0,0 +1,140 @@
...
+  return outcome::success();
+}
+
+[[nodiscard]] ErrorMessageOr<void> RegisterState::RestoreRegisters() {
+  // BackupRegisters needs to be called before RestoreRegisters.
+  CHECK(tid_ != -1);
+
+  iovec iov;
+  iov.iov_base = &general_purpose_registers_;
+  iov.iov_len = bitness_ == Bitness::k32Bit ? sizeof(GeneralPurposeRegisters32)
```
dpallotti, line 120: Add `()` around `bitness_ == Bitness::k32Bit`, this is not super readable.

src/UserSpaceInstrumentation/RegisterStateTest.cpp
```diff
@@ -0,0 +1,96 @@
...
+// verification.
+void Child() {
+  CHECK(ptrace(PTRACE_TRACEME, 0, NULL, 0) != -1);
+
+  uint64_t rax = 0xaabbccdd;
+  uint8_t avx_bytes[32];
+  for (int i = 0; i < 32; ++i) {
+    avx_bytes[i] = i;
+  }
+  __asm__ __volatile__(
```
dpallotti, line 30: Oh wow, please add a comment on what's happening here.

src/UserSpaceInstrumentation/RegisterStateTest.cpp
```diff
@@ -0,0 +1,96 @@
...
+      "mov (%0), %%rax\n\t"
```
dpallotti, line 31: What's the trailing tab for by the way (I have no clue how `__asm__` works)?

src/UserSpaceInstrumentation/RegisterStateTest.cpp
```diff
@@ -0,0 +1,96 @@
...
+#include <sys/wait.h>
+
+#include "OrbitBase/Logging.h"
+#include "UserSpaceInstrumentation/RegisterState.h"
+
+namespace orbit_user_space_instrumentation {
+
+namespace {
+
+// Let the parent trace us, write into rax and ymm0, then break. While the child is breaked the
```
dpallotti, line 18: What do you mean with "break" in this context?

src/UserSpaceInstrumentation/RegisterStateTest.cpp
```diff
@@ -0,0 +1,96 @@
...
+  for (int i = 0; i < 16; ++i) {
+    s.GetFxSave()->xmm[0].bytes[i] += 0x10;
+    s.GetAvxHiRegisters()->ymm[0].bytes[i] += 0x10;
+  }
+  EXPECT_TRUE(s.RestoreRegisters());
+
+  // Continue child.
+  CHECK(ptrace(PT_CONTINUE, pid, 1, 0) == 0);
+
+  // Wait for the child to exit. Exit status is zero if the modified register could be verified.
```
dpallotti, line 89:
```suggestion
  // Wait for the child to exit. Exit status is zero if the modified registers could be verified.
```

src/UserSpaceInstrumentation/RegisterState.cpp
```diff
@@ -0,0 +1,140 @@
...
+  if (!avx_offset.has_error()) {
+    avx_offset_ = avx_offset.value();
+  } else {
+    avx_offset_ = 0;
+  }
+
+  return outcome::success();
+}
+
+[[nodiscard]] ErrorMessageOr<void> RegisterState::RestoreRegisters() {
```
dpallotti, line 114: I don't think you need the `[[nodiscard]]` in the .cpp as well. Same above for `BackupRegisters`.

src/UserSpaceInstrumentation/include/UserSpaceInstrumentation/RegisterState.h
```diff
@@ -0,0 +1,240 @@
...
+  bool HasSseDataStored();
+  bool HasAvxDataStored();
+
+  // Structure access to the different parts of the XSave area. Detail, again, can be found at
+  // "Intel 64 and IA-32 Architectures Software Developer’s Manual, Volume 1", section 13.
+  // "GetFxSave" can be used to access fpu, mmx, sse registers. "GetAvxHiRegisters" give access to
+  // the upper half of the avx registers (lower half in store in the sse registers).
+  FxSave* GetFxSave() { return reinterpret_cast<FxSave*>(xsave_area_.data()); }
+  XSaveHeader* GetXSaveHeader() { return reinterpret_cast<XSaveHeader*>(xsave_area_.data() + 512); }
+  YmmHi* GetAvxHiRegisters() { return reinterpret_cast<YmmHi*>(xsave_area_.data() + avx_offset_); }
```
dpallotti, line 228: `[[nodiscard]]` here too, since you use it above and these are accessors?

dpallotti (commented):
This is all pretty cool. I added some comments about where some additional explanation would I think be beneficial, plus some minor comments.

src/UserSpaceInstrumentation/include/UserSpaceInstrumentation/RegisterState.h
```diff
@@ -0,0 +1,240 @@
...
+  // be in their initial state. So before accessing this data one needs to call the "Has*DataStored"
+  // methods first.
+  bool Hasx87DataStored();
+  bool HasSseDataStored();
+  bool HasAvxDataStored();
+
+  // Structure access to the different parts of the XSave area. Detail, again, can be found at
+  // "Intel 64 and IA-32 Architectures Software Developer’s Manual, Volume 1", section 13.
+  // "GetFxSave" can be used to access fpu, mmx, sse registers. "GetAvxHiRegisters" give access to
+  // the upper half of the avx registers (lower half in store in the sse registers).
```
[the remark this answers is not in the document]
danielfenner: done

src/UserSpaceInstrumentation/include/UserSpaceInstrumentation/RegisterState.h
```diff
@@ -0,0 +1,240 @@
...
+// For the general purpose registers this is somewhat straightforward with the only complication
+// of bitness of the process. Users of this class should query the bitness of the process using
+// "GetBitness" and only access the respective member of the union returned by
+// "GetGeneralPurposeRegisters".
+//
+// For fpu / vector data the situation is slightly more complicated. A comprehensive write-up can be
+// found in [1]. A good, readable introduction can be found here: [2].
+// Different registers may or may not be present on a specific Cpu (e.g. the Cpu may or may not
+// support Avx). Even if the Cpu has the respective registers they might not have been saved by
+// "BackupRegisters" since they are in their initial state. Hence every meaningful access to this
```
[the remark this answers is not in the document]
danielfenner: I added a sentence there: Registers are initialized at process startup. If a program never touches these registers the data is not stored. 
This was a bit surprising for me - I tried it out with a process not using avx and indeed the resulting xsave header had the flag for avx set to zero. It makes sense though that this is optimized a lot: The instructions doing the register backup used for task switching.

src/UserSpaceInstrumentation/include/UserSpaceInstrumentation/RegisterState.h
```diff
@@ -0,0 +1,240 @@
...
+
+#include "OrbitBase/Result.h"
+
+namespace orbit_user_space_instrumentation {
+
+// The structs below are defined to match the memory layout of the binary blobs containing the
+// register state of the cpu (as returned by the kernel). They should not used directly - see
+// comment on class "RegisterState" below.
+#define PACKED_START _Pragma("pack(push, 1)")
+#define PACKED_END _Pragma("pack(pop)")
```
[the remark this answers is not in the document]
danielfenner: Yep, thanks, this looks mildly dangerous. Done.

src/UserSpaceInstrumentation/include/UserSpaceInstrumentation/RegisterState.h
danielfenner: no, just forget it. done.

src/UserSpaceInstrumentation/include/UserSpaceInstrumentation/RegisterState.h
danielfenner: Added some line here. done.

src/UserSpaceInstrumentation/RegisterState.cpp
danielfenner: I added something above these two functions. Not sure if this is a good place. Let me know if you have a better idea...

src/UserSpaceInstrumentation/RegisterState.cpp
danielfenner: done

src/UserSpaceInstrumentation/RegisterState.cpp
danielfenner: Yes, the cpuid leaf 0x01 returns in ecx & bit_XSAVE wheather the processor supports XSAVE. The subleave 0x0d, 0x00 return the maxímum size of the xsave area in ecx. 
Both can only fail when the processor is acient and does not support xsave. So the messages are fine I think.

I'm unsure if the second one can happen; probably the first will already catch all cases where the second call fails.

For the Avx case there might be different reasons (well, theoretically, I'm not sure if the second one can be hit).

src/UserSpaceInstrumentation/RegisterState.cpp
danielfenner: Yes, I thought about that. I could not come up with any case in which injecting the registers of one thread into another leads to anything but desaster. So I chose to forbid it.

src/UserSpaceInstrumentation/RegisterState.cpp
danielfenner: errno might be helpful when looking into the sources where this is emitted.

src/UserSpaceInstrumentation/RegisterState.cpp
danielfenner: done

src/UserSpaceInstrumentation/RegisterState.cpp
danielfenner: done

src/UserSpaceInstrumentation/RegisterStateTest.cpp
danielfenner: There is a comment above the method. 
Let's take this offline and see what makes to add here.

src/UserSpaceInstrumentation/RegisterStateTest.cpp
danielfenner: Just the way clang format likes inline assembly.

src/UserSpaceInstrumentation/RegisterStateTest.cpp
danielfenner: Entering a breakpoint. The int3 in the asm.  Let's chat about that offline as well ...

src/UserSpaceInstrumentation/RegisterStateTest.cpp
danielfenner: done

src/UserSpaceInstrumentation/RegisterState.cpp
danielfenner: done

src/UserSpaceInstrumentation/include/UserSpaceInstrumentation/RegisterState.h
danielfenner: done

src/UserSpaceInstrumentation/include/UserSpaceInstrumentation/RegisterState.h
```diff
@@ -0,0 +1,254 @@
...
+  GeneralPurposeRegisters* GetGeneralPurposeRegisters() { return &general_purpose_registers_; }
+
+  // Some registers do not get stored in RegisterState. The Cpu might not support them or they might
+  // be in their initial state. So before accessing this data one needs to call the "Has*DataStored"
+  // methods first.
+  bool Hasx87DataStored();
+  bool HasSseDataStored();
+  bool HasAvxDataStored();
+
+  // Structure access to the different parts of the XSave area. Detail, again, can be found at
```
dpallotti, line 236:
```suggestion
  // Structure access to the different parts of the XSave area. Details, again, can be found at
```

src/UserSpaceInstrumentation/RegisterStateTest.cpp
danielfenner: done

src/UserSpaceInstrumentation/include/UserSpaceInstrumentation/RegisterState.h
```diff
@@ -147,17 +147,29 @@ struct YmmHi {
...
 // Requires the XSave extension (added around 2008) to be supported by the Cpu otherwise
-// "BackupRegisters" will return an error. RegisterState requires to be initialized by a call to
-// "BackupRegisters" before anything else can be called on the object. RegisterState stores the
-// general purpose registers as well as all floating point and vector registers.
+// "BackupRegisters" will return an error. The XSave feature set includes instructions that save and
+// restore the XSave managed "state components" to and from memory. These state components
+// correspond to sets of registers introduced for different processor features (e.g. Avx, Avx-512,
+// ...). The processor organizes the state components in a region of memory called an XSave area.
+// The XSave area comprises the legacy region (roughly: storing everything up to SSE2), the XSave
+// header (roughly: flags indicating what is in the xtended region) and the extended region itself
```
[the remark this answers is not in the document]
danielfenner: done

src/UserSpaceInstrumentation/include/UserSpaceInstrumentation/RegisterState.h
```diff
@@ -166,8 +178,10 @@ PACKED_END
 // found in [1]. A good, readable introduction can be found here: [2].
 // Different registers may or may not be present on a specific Cpu (e.g. the Cpu may or may not
 // support Avx). Even if the Cpu has the respective registers they might not have been saved by
-// "BackupRegisters" since they are in their initial state. Hence every meaningful access to this
-// data needs to call the "Has*DataStored" functions first.
+// "BackupRegisters" since they are in their initial state (e.g. Avx registers are initialized with
+// zeros on process startup; if the process makes no use of Avx storing these registers is skipped
```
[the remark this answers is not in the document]
danielfenner: done

src/UserSpaceInstrumentation/include/UserSpaceInstrumentation/RegisterState.h
danielfenner: done

src/UserSpaceInstrumentation/RegisterStateTest.cpp
danielfenner: Added something. ptal.

src/UserSpaceInstrumentation/RegisterState.cpp
```diff
@@ -0,0 +1,146 @@
...
+
+#include "OrbitBase/Logging.h"
+#include "OrbitBase/SafeStrerror.h"
+#include "absl/strings/str_format.h"
+
+namespace orbit_user_space_instrumentation {
+
+namespace {
+
+// Some notes reguarding the calls to cpuid below:
```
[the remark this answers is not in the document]
danielfenner: done. ptal ...

src/UserSpaceInstrumentation/include/UserSpaceInstrumentation/RegisterState.h
```diff
@@ -0,0 +1,258 @@
...
+  // Returns a pointer to a union of 32 and 64 bit general purpose registers. Call "GetBitness" to
+  // determine which member of the union is valid.
+  GeneralPurposeRegisters* GetGeneralPurposeRegisters() { return &general_purpose_registers_; }
+
+  // Some registers do not get stored in RegisterState. The Cpu might not support them or they might
+  // be in their initial state. So before accessing this data one needs to call the "Has*DataStored"
+  // methods first.
+  bool Hasx87DataStored();
+  bool HasSseDataStored();
+  bool HasAvxDataStored();
```
dpallotti, line 234: `[[nodiscard]]`

dpallotti (approved):

src/UserSpaceInstrumentation/RegisterStateTest.cpp
```diff
@@ -0,0 +1,100 @@
...
+
+#include "OrbitBase/Logging.h"
+#include "UserSpaceInstrumentation/RegisterState.h"
+
+namespace orbit_user_space_instrumentation {
+
+namespace {
+
+// Let the parent trace us, write into rax and ymm0, then enter a breakpoint. While the child is
+// stopped the parent modifies the register and continues the child. The child then reads back the
```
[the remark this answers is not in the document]
danielfenner: done

src/UserSpaceInstrumentation/RegisterStateTest.cpp
```diff
@@ -0,0 +1,100 @@
...
+// of that verification.
+void Child() {
+  CHECK(ptrace(PTRACE_TRACEME, 0, NULL, 0) != -1);
+
+  uint64_t rax = 0xaabbccdd;
+  uint8_t avx_bytes[32];
+  for (int i = 0; i < 32; ++i) {
+    avx_bytes[i] = i;
+  }
+  // The first two lines move the memory to the registers. The "%0" and "%1" refer to the adresses
```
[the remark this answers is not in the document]
danielfenner: done

src/UserSpaceInstrumentation/RegisterStateTest.cpp
```diff
@@ -0,0 +1,100 @@
...
+  // of "rax" and "avx_bytes" given in the second line from the bottom. "int3" just is the
+  // breakpoint. The parent does "waitpid" for that. line four and five move the registers back into
```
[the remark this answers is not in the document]
danielfenner: done

src/UserSpaceInstrumentation/include/UserSpaceInstrumentation/RegisterState.h
danielfenner: done

PR #1938, opened 2021-02-24
Preparation steps for integrating Python util code in our CI

ronaldfw:


conanfile.py
```diff
@@ -147,6 +150,7 @@ def configure(self):
 
 
     def build(self):
+        print("build")
```
ronaldfw, line 153: Remove.

conanfile.py
```diff
@@ -161,6 +165,9 @@ def build(self):
         cmake.build(target=str(self.options.build_target) if self.options.build_target else None)
         if self.options.run_tests and not tools.cross_building(self.settings, skip_x64_x86=True) and self.settings.get_safe("os.platform") != "GGP":
             cmake.test(output_on_failure=True)
+        print(self.options.run_python_tests)
```
ronaldfw, line 168: Remove.

contrib/jupyter/run_tests.py
```diff
@@ -0,0 +1,8 @@
+import orbitutils.run_all_unittests as unit_tests
```
ronaldfw, line 1: Copyright header missing.

contrib/jupyter/build.py
```diff
@@ -5,24 +5,39 @@
...
+
+    command = "protoc --python_out=" + orbitutils_path + " --proto_path=" + proto_path + " " + capture_data_proto
+    ret = os.system(command)
+    if ret:
+        print("Command failed: ", command)
+        sys.exit(ret)
+
+    install_orbitutils(current_directory)
+
+    subprocess.check_call(
```
ronaldfw, line 37: Calling this as a subprocess ensures that orbitutils is installed when executing the imports in run_tests.py. There are instructions how to do an "try import; install if failed; retry import" on https://stackoverflow.com/questions/12332975/installing-python-module-within-code, but this did not work for me. So, right now I don't know if there is a more elegant solution. I might just be doing something wrong.

beckerhe: I think the subprocess approach is viable. I consider it coincidence that the language we use for writing this script is the same language as the code under test. If this was a bash script this would be totally fine.

FWIW, I think you might need to reload your import path:
```
import site
from importlib import reload
reload(site)
```
Ref: https://stackoverflow.com/questions/25384922/how-to-refresh-sys-path

contrib/jupyter/build.py
```diff
@@ -5,24 +5,39 @@
...
 
-file_directory = os.path.dirname(os.path.abspath(__file__))
-orbitutils_path = os.path.join(file_directory, "orbitutils")
 
-# We need to specify the "root" of the src tree. With this choice, we can then
-# import the proto as "import ClientProtos.capture_data_pb2.py"
-proto_path = os.path.join(file_directory, "../../src/")
+def install_orbitutils(current_directory):
+    subprocess.check_call(
+        [sys.executable, "-m", "pip", "install", "-e", current_directory])
```
beckerhe, line 14: Nit: current_directory -> installation_directory or target_directory?

contrib/jupyter/build.py
ronaldfw: Hmm, I couldn't figure out how to do the reload in a few minutes (it also seemed to have changed with Python 3), so I think I'll just go with the subprocess approach.

contrib/jupyter/build.py
ronaldfw: The directory passed actually needs to be the directory that contains the setup.py file and also contains the "orbitutils" folder. I renamed to "orbitutils_setup_path" and added a comment.

ronaldfw:
@beckerhe Please have another look, thank you!

beckerhe:
Ok, cool. Looks good. I think the next step is to actually enable it for the CI. For this you have to set the run_python_test option in one of the profiles that run on the CI. I suggest we use clang7_relwithdebinfo (`third_party/conan/configs/linux/profiles/clang7_relwithdebinfo`):

```
[options]
OrbitProfiler:run_python_tests=True
```

The options-section should aready be there but empty. So you just have to add the line below.

ronaldfw:
> Ok, cool. Looks good. I think the next step is to actually enable it for the CI. For this you have to set the run_python_test option in one of the profiles that run on the CI. I suggest we use clang7_relwithdebinfo (`third_party/conan/configs/linux/profiles/clang7_relwithdebinfo`):
> 
> ```
> [options]
> OrbitProfiler:run_python_tests=True
> ```
> 
> The options-section should aready be there but empty. So you just have to add the line below.

My intention was to add this in a follow-up PR. Then it's easier to revert enabling the CI without losing the changes in the current PR. WDYT?

beckerhe:
> > Ok, cool. Looks good. I think the next step is to actually enable it for the CI. For this you have to set the run_python_test option in one of the profiles that run on the CI. I suggest we use clang7_relwithdebinfo (`third_party/conan/configs/linux/profiles/clang7_relwithdebinfo`):
> > ```
> > [options]
> > OrbitProfiler:run_python_tests=True
> > ```
> > 
> > 
> > The options-section should aready be there but empty. So you just have to add the line below.
> 
> My intention was to add this in a follow-up PR. Then it's easier to revert enabling the CI without losing the changes in the current PR. WDYT?

Sure. That's also fine with me.

beckerhe (approved):

PR #1941, opened 2021-02-25
Add test for retrying of submission completion when queries fail

ronaldfw:
We recently added retrying completing a submission upon failure of
timestamp queries in the Vulkan layer. Here we add a unit test that checks for the retry
on query fail.

Tested: Ran unit test (only unit test code was changed).
Bug: http://b/179649352

### DIFF
--- a/src/OrbitVulkanLayer/SubmissionTrackerTest.cpp
+++ b/src/OrbitVulkanLayer/SubmissionTrackerTest.cpp
@@ -2,6 +2,8 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
 
+#include <array>
+
 #include "OrbitBase/ThreadUtils.h"
 #include "SubmissionTracker.h"
 #include "VulkanLayerProducer.h"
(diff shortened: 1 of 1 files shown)


src/OrbitVulkanLayer/SubmissionTrackerTest.cpp
```diff
@@ -536,6 +538,99 @@ TEST_F(SubmissionTrackerTest,
...
+  post_submit_times[1] = orbit_base::CaptureTimestampNs();
+
+  tracker_.CompleteSubmits(device_);
+  tracker_.CompleteSubmits(device_);
+
+  EXPECT_THAT(actual_reset_slots1, UnorderedElementsAre(kSlotIndex1, kSlotIndex2, kSlotIndex4));
+  // We failed on the 4th GetQuerySlots call above, which belongs to the "begin" timestamp of the
+  // second command buffer (note that the order of calls when querying for timestamps is to query
+  // for the end timestamp first). Therefore, this must be kSlotIndex3.
+  // TODO: Consider if we should just check the size of the reset slots? Depending on code order
```
ronaldfw, line 624: Thoughts?

florian-kuebler: I was also thinking about this. I guess it is fine for the tests. We had the tests from quite early on, and changed the code and the tests quite a bit. AFAIR we did't had issues with slots indices. So I think this is fine for the tests. Alternative would be to collect the slots into the same vector, and just check that all of them are reset exactly once.

florian-kuebler (approved):
LGTM. Thanks!

src/OrbitVulkanLayer/SubmissionTrackerTest.cpp
ronaldfw: Ok, if we find it bothers us, we can change it later. I'll merge as-is. Thanks for the input!

PR #1944, opened 2021-02-25
VkFreeCommandBuffers also rollback the timer slots and remove the state.

florian-kuebler:
A command buffer can actually also be "free"ed while it is in
recording or executable state. On free, we remove the command
buffer from our internal maps, but not from the state map.

This leads to a crash in OnCaptureStopped, as the device mapping
is gone, but the state remains.

This change prevents this, by also resetting the slots on Free
and erasing the state.

Bug: http://b/181217248
Test: Run tests or stop a capture at the end of infiltrator (when
the gpu is done)

### DIFF
--- a/src/OrbitVulkanLayer/SubmissionTracker.h
+++ b/src/OrbitVulkanLayer/SubmissionTracker.h
@@ -181,6 +181,13 @@ class SubmissionTracker : public VulkanLayerProducer::CaptureStatusListener {
     for (uint32_t i = 0; i < count; ++i) {
       VkCommandBuffer command_buffer = command_buffers[i];
       associated_command_buffers.erase(command_buffer);
+
+      if (command_buffer_to_state_.contains(command_buffer)) {
+        ResetCommandBufferUnsafe(command_buffer);
+
+        command_buffer_to_state_.erase(command_buffer);
+      }
+
       CHECK(command_buffer_to_device_.contains(command_buffer));
       CHECK(command_buffer_to_device_.at(command_buffer) == device);
       command_buffer_to_device_.erase(command_buffer);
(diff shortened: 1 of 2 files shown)


src/OrbitVulkanLayer/SubmissionTracker.h
```diff
@@ -934,6 +920,32 @@ class SubmissionTracker : public VulkanLayerProducer::CaptureStatusListener {
     }
   }
 
+  // This method does not acquire a lock and MUST NOT be called without holding the `mutex_`.
+  void ResetCommandBufferUnsafe(VkCommandBuffer command_buffer) {
```
dpallotti, line 924: "Unsafe" sounds a bit worse that it actually is. But I don't really know what a better word would be. "Unsynchronized"/"Unsynch" maybe? Or feel free to leave it as is if you don't like it.

dpallotti (approved):
Like the other change I think it would make sense to move some details from the commit message into one (or multiple if necessary) comments, so that the explanation doesn't get lost. Looks good otherwise.

src/OrbitVulkanLayer/SubmissionTracker.h
florian-kuebler: I've seen that in a couple of other projects, also used for missing synchronisation. The nice thing is that it actually sounds bad. So when you use it, you will look at the method, to check "why is it unsafe".

However, I don't feel strongly about this and could also use "Unsynchronized".

dpallotti: Hehe that's an interesting point of view. I don't have a strong opinion either, I just wanted to mention it.

florian-kuebler: I guess I'll leave it like this.

florian-kuebler: > Like the other change I think it would make sense to move some details from the commit message into one (or multiple if necessary) comments, so that the explanation doesn't get lost. Looks good otherwise.

Thanks. Added two comments.

PR #1907, opened 2021-02-19
Add Capture and preset load and save metrics

antonrohr:
This belongs together with http://cl/358374603

PR #1945, opened 2021-02-25
Refactor metrics uploader

antonrohr:
The first commit just does some simple refactoring of MetricsUploaderWindowsTest.cpp (was part of #1907 <- will close soon).

The second commit makes MetricsUploader an abstract class, so it can be more easily mocked in the future. (Pre work for upcoming PR).


src/MetricsUploader/include/MetricsUploader/MetricsUploader.h
```diff
@@ -36,39 +36,20 @@ const std::string kMetricsUploaderClientDllName = "metrics_uploader_client";
...
-  MetricsUploader();
-#endif
-
-  Result (*send_log_event_addr_)(const uint8_t*, int) = nullptr;
-  Result (*shutdown_connection_addr_)() = nullptr;
-  std::string client_name_;
-  std::string session_uuid_;
+  virtual bool SendLogEvent(
+      OrbitLogEvent_LogEventType log_event_type,
+      std::chrono::milliseconds event_duration = std::chrono::milliseconds::zero()) = 0;
```
IrinaShkviro, line 52: Maybe we can add base implementation for this function instead of making it pure virtual? In this case we can return from CreateMetricsUploader the base class instance if we don't have `.dll` available (or in case other errors), and avoid checking `MetricsUploader*` for nullptr when we want to send metrics.

src/MetricsUploader/include/MetricsUploader/MetricsUploader.h
```diff
@@ -36,39 +36,20 @@ const std::string kMetricsUploaderClientDllName = "metrics_uploader_client";
...
 class MetricsUploader {
  public:
+  MetricsUploader() = default;
+  virtual ~MetricsUploader() = default;
+
   // Create a MetricsUploader instance, load metrics uploader client library if available and starts
   // metrics uploader client when called first time. Return the instance if there are no errors and
   // ErrorMessage otherwise.
-  [[nodiscard]] static ErrorMessageOr<MetricsUploader> CreateMetricsUploader(
+  [[nodiscard]] static ErrorMessageOr<std::unique_ptr<MetricsUploader>> CreateMetricsUploader(
```
IrinaShkviro, line 45: Why it's better to have a `unique_ptr` here instead of value?

src/MetricsUploader/MetricsUploaderWindows.cpp
```diff
@@ -18,17 +19,43 @@ constexpr const char* kSetupConnectionFunctionName = "SetupConnection";
...
-                                 Result (*shutdown_connection_addr)(),
-                                 HMODULE metrics_uploader_client_dll)
+class MetricsUploaderImpl : public MetricsUploader {
+ public:
+  explicit MetricsUploaderImpl(std::string client_name, std::string session_uuid,
+                               Result (*send_log_event_addr)(const uint8_t*, int),
+                               Result (*shutdown_connection)(),
+                               HMODULE metrics_uploader_client_dll);
+  MetricsUploaderImpl(const MetricsUploaderImpl& other) = delete;
+  MetricsUploaderImpl(MetricsUploaderImpl&& other);
```
beckerhe, line 29: noexcept

src/MetricsUploader/MetricsUploaderWindows.cpp
```diff
@@ -18,17 +19,43 @@ constexpr const char* kSetupConnectionFunctionName = "SetupConnection";
...
+  MetricsUploaderImpl& operator=(const MetricsUploaderImpl& other) = delete;
+  MetricsUploaderImpl& operator=(MetricsUploaderImpl&& other);
```
beckerhe, line 31: noexcept

beckerhe (approved):
Otherwise LGTM

src/MetricsUploader/include/MetricsUploader/MetricsUploader.h
antonrohr: I am going to be quite honest, I wasn't sure if this is a good idea or not, so I talked to Henning about it. He confirmed to me that the current implementation with `MetricsUploader` only being an interface is actually pretty nice. For the reason that if someone else wants to write another child class to `MetricsUploader`, there will be compiler errors if it was forgotten to implement a method (in this case only `SendLogEvent`). If it's not a pure virtual class, it could happen that the base implementation is accidentally used, because there are no errors.  

I hope this explanation makes sense.

Conclusion: I would leave everything as is, wdyt?

src/MetricsUploader/include/MetricsUploader/MetricsUploader.h
antonrohr: Its not better, its necessary because of the pure virtual situation (see other comment). Indeed it would be nicer to just have the value and not a pointer.

IrinaShkviro (approved):

PR #1946, opened 2021-02-26
Fix an error in Highlighting m_variables

karupayun:
m_VariableName is widely used to refer to class members. Actually, we
used to have them in Orbit. There is an error in that regex, we should have a separator before that variable, because currently is matching for example with "pam_variable". This PR is fixing this.

Test: Run CodeViewerDemo.

### DIFF
--- a/src/CodeViewerDemo/CodeExamples.h
+++ b/src/CodeViewerDemo/CodeExamples.h
@@ -28,6 +28,7 @@ inline const QString testing_example =
   m_PascalCase = 5;
   Track::SortTracks();
   class TrackManager {};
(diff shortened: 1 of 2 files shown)


beckerhe (approved):

florian-kuebler:
What happens to `foo.m_Member` or `foo->m_Member`? Is that also covered?

karupayun:
> What happens to `foo.m_Member` or `foo->m_Member`? Is that also covered?

In both cases I'm painting "m_Member", shouldn't I?

florian-kuebler:
Perfect.

PR #1947, opened 2021-02-26
Reduce connection backoff time of ProducerSideChannel

dpallotti:
One alternative would be to set "grpc.testing.fixed_reconnect_backoff_ms"
instead of min and max. This seems to be somewhat common, but I decided against
it because of "testing" and the fact that it doesn't have a constant (rather,
macro) associated.

Bug: http://b/181121581 (Vulkan layer producer not always connected)

Test: run Infiltrator with the layer a couple of times and verify it connects
quickly.

### DIFF
--- a/src/ProducerSideChannel/include/ProducerSideChannel/ProducerSideChannel.h
+++ b/src/ProducerSideChannel/include/ProducerSideChannel/ProducerSideChannel.h
@@ -22,7 +22,19 @@ constexpr std::string_view kProducerSideUnixDomainSocketPath = "/tmp/orbit-produ
 inline std::shared_ptr<grpc::Channel> CreateProducerSideChannel(
     std::string_view unix_domain_socket_path = kProducerSideUnixDomainSocketPath) {
   std::string server_address = absl::StrFormat("unix:%s", unix_domain_socket_path);
-  return grpc::CreateChannel(server_address, grpc::InsecureChannelCredentials());
+
+  grpc::ChannelArguments channel_arguments;
+  // Significantly reduce the gRPC channel's reconnection backoff time. Defaults for min and max
+  // would be 20 seconds and 2 minutes. That's too much for us, as we want a producer to quickly
+  // connect to OrbitService after Orbit is started, so that when starting a capture the producer
+  // can already send data.
+  constexpr int kMinReconnectBackoffMs = 1000;
+  constexpr int kMaxReconnectBackoffMs = 1000;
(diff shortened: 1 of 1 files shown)


src/ProducerSideChannel/include/ProducerSideChannel/ProducerSideChannel.h
```diff
@@ -22,7 +22,19 @@ constexpr std::string_view kProducerSideUnixDomainSocketPath = "/tmp/orbit-produ
...
   std::string server_address = absl::StrFormat("unix:%s", unix_domain_socket_path);
-  return grpc::CreateChannel(server_address, grpc::InsecureChannelCredentials());
+
+  grpc::ChannelArguments channel_arguments;
+  // Reduce the gRPC channel's reconnection backoff time. Defaults for min and max would be 20
+  // seconds and 2 minutes. That's too much for us, as we want a producer to quickly connect to
+  // OrbitService after Orbit is started, so that when starting a capture the producer can already
+  // send data.
+  constexpr int kMinReconnectBackoffMs = 1000;
+  constexpr int kMaxReconnectBackoffMs = 5000;
```
florian-kuebler, line 32: Would it make sense to go even lower?

dpallotti: Possibly, yes. I don't know if there's a downside but I would guess none measurable. Keep in mind that we still only retry to connect every five seconds in `OrbitProducer`. So with these values it can take up to 10 seconds to actually connect. Bringing `kMaxReconnectBackoffMs` down to, say, 1 second should bring this max down to 6 seconds, so it should be worth it. Should I do that?

florian-kuebler: Yeah, I would try to get overall connection time down to lets say one (or at most two seconds).
A power-user of Orbit might actually quite fast in connecting, selecting the game and starting a capture.

florian-kuebler: Otherwise, LGTM, thanks!

florian-kuebler (approved):

src/ProducerSideChannel/include/ProducerSideChannel/ProducerSideChannel.h
dpallotti: I changed both min and max to one second. Let's follow up on the other retry timeout separately.

florian-kuebler: Thanks. Sounds good!

PR #1948, opened 2021-02-26
Fix segfault in SubmissionTracker::SetVulkanLayerProducer

dpallotti:
Bug: http://b/181234480 (Infiltrator reports segfault on VkDestroyInstance)

Test: Update unit tests. Run Infiltrator with the layer and verify that no crash
is reported.

### DIFF
--- a/src/OrbitVulkanLayer/SubmissionTracker.h
+++ b/src/OrbitVulkanLayer/SubmissionTracker.h
@@ -143,11 +143,14 @@ class SubmissionTracker : public VulkanLayerProducer::CaptureStatusListener {
   // We will also register ourselves as CaptureStatusListener, in order to get notified on
   // capture finish (OnCaptureFinished). That's when we will reset the open query slots.
   void SetVulkanLayerProducer(VulkanLayerProducer* vulkan_layer_producer) {
+    // De-register from the previous VulkanLayerProducer.
+    if (vulkan_layer_producer_ != nullptr) {
+      vulkan_layer_producer_->SetCaptureStatusListener(nullptr);
+    }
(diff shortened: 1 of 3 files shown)


src/OrbitVulkanLayer/SubmissionTracker.h
```diff
@@ -143,11 +143,14 @@ class SubmissionTracker : public VulkanLayerProducer::CaptureStatusListener {
   // We will also register ourselves as CaptureStatusListener, in order to get notified on
   // capture finish (OnCaptureFinished). That's when we will reset the open query slots.
   void SetVulkanLayerProducer(VulkanLayerProducer* vulkan_layer_producer) {
+    // De-register from the previous VulkanLayerProducer.
```
dpallotti, line 146: Do you know if de-registering was also the intent of the previous code? It seems like it but I'm not 100% sure if that was the reason for `vulkan_layer_producer->SetCaptureStatusListener(nullptr)`.

florian-kuebler: I belief it was intended, but I am totally unsure, why? Is it because we can handle multiple devices/instances (in theory)?

florian-kuebler (approved):

src/OrbitVulkanLayer/SubmissionTracker.h
dpallotti: I would guess it was mainly for completeness.

florian-kuebler: Probably.

florian-kuebler (approved):
LGTM, thanks!

PR #1949, opened 2021-02-26
Fix crash in vkCreateDevice.

florian-kuebler:
In order to check if our required extensions are available,
we use our next layer's "vkGetInstanceProcAddr" to query for
the "vkEnumerateDeviceExtensionProperties". Unlike the "Instance"
version of the enumeration function, this can not be called with
nullptr as instance (see: https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/vkGetInstanceProcAddr.html).

This establishes a mapping from dispatchable object to instance
in the dispatch table (created on "vkCreateInstance").

Test: Run Infiltrator and Trata with the layer.


ronaldfw (approved):
LGTM

PR #1950, opened 2021-02-26
Improve preset loading error message

antonrohr:
Beforehand it was assumed that a preset cannot be loaded when non of the modules are loaded by the process. This is actually wrong. When a preset is loaded via File > Open Preset, it can happen that a preset is not loaded at all. This PR fixes that. The case that a preset is not loaded at all will be covered by a metric

### DIFF
--- a/src/OrbitGl/App.cpp
+++ b/src/OrbitGl/App.cpp
@@ -1399,10 +1399,17 @@ void OrbitApp::LoadSymbols(const std::filesystem::path& symbols_path, ModuleData
 
 orbit_base::Future<ErrorMessageOr<void>> OrbitApp::LoadPresetModule(
     const std::string& module_path, const orbit_client_protos::PresetModule& preset_module) {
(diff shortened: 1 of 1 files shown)


src/OrbitGl/App.cpp
```diff
@@ -1399,10 +1399,17 @@ void OrbitApp::LoadSymbols(const std::filesystem::path& symbols_path, ModuleData
 
 orbit_base::Future<ErrorMessageOr<void>> OrbitApp::LoadPresetModule(
     const std::string& module_path, const orbit_client_protos::PresetModule& preset_module) {
+  if (!GetTargetProcess()->IsModuleLoaded(module_path)) {
+    return {ErrorMessage{"Module not loaded by process."}};
```
beckerhe, line 1403: BTW, The additional curly braces are not needed anymore. Future now has a special constructor which takes an error message.

beckerhe (approved):
LGTM

PR #1952, opened 2021-02-26
Metrics Capture & Preset, load & save

antonrohr:
This is redoing PR #1907.

It now introduces ScopedMetric and uses it for capture and preset loading and saving metrics. 
The comments from the last PR are also addressed.

### DIFF
--- a/src/MetricsUploader/MetricsUploaderWindows.cpp
+++ b/src/MetricsUploader/MetricsUploaderWindows.cpp
@@ -182,24 +182,28 @@ ErrorMessageOr<std::string> GenerateUUID() {
 
 bool MetricsUploaderImpl::SendLogEvent(OrbitLogEvent_LogEventType log_event_type,
                                        std::chrono::milliseconds event_duration) {
-  if (send_log_event_addr_ != nullptr) {
-    OrbitLogEvent log_event;
-    log_event.set_log_event_type(log_event_type);
-    log_event.set_orbit_version(orbit_core::GetVersion());
-    log_event.set_event_duration_milliseconds(event_duration.count());
-    log_event.set_session_uuid(session_uuid_);
-
-    int message_size = log_event.ByteSize();
-    std::vector<uint8_t> buffer(message_size);
-    log_event.SerializeToArray(buffer.data(), message_size);
-
-    Result result = send_log_event_addr_(buffer.data(), message_size);
-    if (result == kNoError) {
-      return true;
-    }
-    ERROR("Can't start the metrics uploader client: %s", GetErrorMessage(result));
+  if (send_log_event_addr_ == nullptr) {
+    ERROR("Unable to send metric, send_log_event_addr_ is nullptr");
+    return false;
   }
-  return false;
+
+  OrbitLogEvent log_event;
+  log_event.set_log_event_type(log_event_type);
+  log_event.set_orbit_version(orbit_core::GetVersion());
+  log_event.set_event_duration_milliseconds(event_duration.count());
+  log_event.set_session_uuid(session_uuid_);
+
+  int message_size = log_event.ByteSize();
+  std::vector<uint8_t> buffer(message_size);
+  log_event.SerializeToArray(buffer.data(), message_size);
+
+  Result result = send_log_event_addr_(buffer.data(), message_size);
+  if (result != kNoError) {
+    ERROR("Unable to send metrics event: %s", GetErrorMessage(result));
+    return false;
+  }
+
+  return true;
 }
 
 }  // namespace orbit_metrics_uploader
--- a/src/MetricsUploader/MetricsUploaderWindows.cpp
+++ b/src/MetricsUploader/MetricsUploaderWindows.cpp
@@ -11,6 +11,7 @@
 #include "OrbitBase/Logging.h"
 #include "OrbitBase/Result.h"
 #include "OrbitVersion/OrbitVersion.h"
+#include "orbit_log_event.pb.h"
 
 namespace orbit_metrics_uploader {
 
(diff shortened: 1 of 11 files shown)


src/MetricsUploader/MetricsUploaderWindows.cpp
```diff
@@ -181,7 +181,8 @@ ErrorMessageOr<std::string> GenerateUUID() {
 }
 
 bool MetricsUploaderImpl::SendLogEvent(OrbitLogEvent_LogEventType log_event_type,
-                                       std::chrono::milliseconds event_duration) {
+                                       std::chrono::milliseconds event_duration,
+                                       OrbitLogEvent_StatusCode status_code) {
```
IrinaShkviro: Maybe we can use partially filled `OrbitLogEvent` as an argument instead of log_event_type, event_duration and status_code (I guess list will grow...)?

beckerhe:
Sorry, I completely missed this. I will have a look in a couple of minutes.

src/MetricsUploader/ScopedMetric.cpp
```diff
@@ -0,0 +1,49 @@
...
+    : uploader_(uploader),
+      log_event_type_(log_event_type),
+      start_(std::chrono::steady_clock::now()) {}
+
+ScopedMetric::~ScopedMetric() {
+  auto duration = std::chrono::duration_cast<std::chrono::milliseconds>(
+      std::chrono::steady_clock::now() - start_);
+  LOG("called dtor, duration would be %d", duration.count());
+  if (uploader_ == nullptr) return;
+  LOG("dtor called for real, duration %d", duration.count());
```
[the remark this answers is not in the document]
antonrohr: Sorry these logs are actually from my testing and "debugging" I remove them completely

src/MetricsUploader/include/MetricsUploader/MetricsUploader.h
```diff
@@ -49,7 +49,8 @@ class MetricsUploader {
   // Returns true on success and false otherwise.
   virtual bool SendLogEvent(
       OrbitLogEvent_LogEventType log_event_type,
-      std::chrono::milliseconds event_duration = std::chrono::milliseconds::zero()) = 0;
+      std::chrono::milliseconds event_duration = std::chrono::milliseconds::zero(),
+      OrbitLogEvent_StatusCode status_code = OrbitLogEvent_StatusCode_SUCCESS) = 0;
```
beckerhe, line 53: Default arguments for virtual function are not ideal, and shouldn't be used according to the style guide: https://google.github.io/styleguide/cppguide.html#Default_Arguments

beckerhe: Sorry, for raising this now. This probably slipped through one of the previous reviews.

beckerhe (approved):
Very nice work. I'm glad that we talked about it. The API is really good now.

LGTM, apart from some minor comments. And what Irina said also makes sense.

antonrohr:
I did 2 things, ptal. 

1. I changed the enum COMMAND_FAILURE to INTERNAL_ERROR. This is because of http://cl/360192320

2. I redid the API of `MetricsUploader.h` uploader a little bit. There are now these 3 overloads:
 ```
 bool SendLogEvent(OrbitLogEvent_LogEventType log_event_type)
bool SendLogEvent(OrbitLogEvent_LogEventType log_event_type, std::chrono::milliseconds event_duration)
bool SendLogEvent(OrbitLogEvent_LogEventType log_event_type, std::chrono::milliseconds event_duration, OrbitLogEvent_StatusCode status_code)
```
Which internally all use `bool MetricsUploaderImpl::FillAndSendLogEvent(OrbitLogEvent partial_filled_event)`. I am unsure about this design choice, maybe there is a better way of solving that. Making the method `FillAndSendLogEvent(OrbitLogEvent)` public is certainly an option, but I feel a weird about exposing this "the user has to create a half proto, and MetricsUploader does the rest" workflow. 

Please let me know what you think

src/MetricsUploader/include/MetricsUploader/MetricsUploader.h
antonrohr: No problem at all. Better late than never. I wrote a comment in the main thread and changed some things. ptal

src/MetricsUploader/MetricsUploaderWindows.cpp
antonrohr: I am unsure how I feel about having an API for MetricsUploader, where the user has to provide "half" a proto. I rather have some overloads, like I did now. I don't think this will grow much much further. ptal at my comment in the "main thread"

beckerhe:
> I did 2 things, ptal.
> 
> 1. I changed the enum COMMAND_FAILURE to INTERNAL_ERROR. This is because of http://cl/360192320
> 2. I redid the API of `MetricsUploader.h` uploader a little bit. There are now these 3 overloads:
> 
> ```
> bool SendLogEvent(OrbitLogEvent_LogEventType log_event_type)
> bool SendLogEvent(OrbitLogEvent_LogEventType log_event_type, std::chrono::milliseconds event_duration)
> bool SendLogEvent(OrbitLogEvent_LogEventType log_event_type, std::chrono::milliseconds event_duration, OrbitLogEvent_StatusCode status_code)
> ```
> 
> Which internally all use `bool MetricsUploaderImpl::FillAndSendLogEvent(OrbitLogEvent partial_filled_event)`. I am unsure about this design choice, maybe there is a better way of solving that. Making the method `FillAndSendLogEvent(OrbitLogEvent)` public is certainly an option, but I feel a weird about exposing this "the user has to create a half proto, and MetricsUploader does the rest" workflow.
> 
> Please let me know what you think

I think it's fine. But I also see Irina's point regarding the proto. In general I would argue that decoupling the protos from the API is good thing, but here we are using the enum declarations from the proto, so this doesn't really apply here. It's already coupled.

Furthermore you have to consider what "half-a-proto" means. Do all possible "halves" work? To me it looks like we need at least an event type, right? So I would prefer an API which clearly states what are the mandatory fields. And plain simple function arguments do that.

IrinaShkviro:
> I did 2 things, ptal.
> 
> 1. I changed the enum COMMAND_FAILURE to INTERNAL_ERROR. This is because of http://cl/360192320
> 2. I redid the API of `MetricsUploader.h` uploader a little bit. There are now these 3 overloads:
> 
> ```
> bool SendLogEvent(OrbitLogEvent_LogEventType log_event_type)
> bool SendLogEvent(OrbitLogEvent_LogEventType log_event_type, std::chrono::milliseconds event_duration)
> bool SendLogEvent(OrbitLogEvent_LogEventType log_event_type, std::chrono::milliseconds event_duration, OrbitLogEvent_StatusCode status_code)
> ```
> 
> Which internally all use `bool MetricsUploaderImpl::FillAndSendLogEvent(OrbitLogEvent partial_filled_event)`. I am unsure about this design choice, maybe there is a better way of solving that. Making the method `FillAndSendLogEvent(OrbitLogEvent)` public is certainly an option, but I feel a weird about exposing this "the user has to create a half proto, and MetricsUploader does the rest" workflow.
> 
> Please let me know what you think

Are there situations where we want to set status code, but not duration? I mean that if we have `n` "optional" fields in OrbitLogEvent then amount of functions will be 2^n. I'm actually don't insist on half-proto solution, what I actually mean is that
1. we probably don't want to have too much functions
2. it should be easy to add a new field in OrbitLogEvent proto. It may be not applicable if we know for sure that we have only 2 "optional" fields (duration and status code for now), I just don't want that this fact stops us from adding "optional" fields if needed.

antonrohr:
> Are there situations when we want to set status code, but not duration? I mean that if we have n "optional" fields in OrbitLogEvent then amount of functions will be 2^n. I'm actually don't insist on half-proto solution, what I actually mean is that
> 1. we probably don't want to have too much functions
> 2. it should be easy to add a new field in OrbitLogEvent proto. It may be not applicable if we know for sure that we have only 2 "optional" fields (duration and status code for now), I just don't want that this fact stops us from adding "optional" fields if needed.

Since I dont see a solution without tradeoffs, I decided to leave everything as is for now. Currently its only 3 functions and I am unsure how many more will be added in the future. It might be changed rather sooner than later, but I am okay with that.

PR #1953, opened 2021-02-27
Add ErrorMessage constructor that calls StrFormat()

dimitry-:
Test: builds

### DIFF
--- a/src/OrbitBase/include/OrbitBase/Result.h
+++ b/src/OrbitBase/include/OrbitBase/Result.h
@@ -5,6 +5,8 @@
 #ifndef ORBIT_BASE_RESULT_H_
 #define ORBIT_BASE_RESULT_H_
 
+#include <absl/strings/str_format.h>
+
 #include <string>
 #include <type_traits>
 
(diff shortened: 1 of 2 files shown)


src/OrbitBase/include/OrbitBase/Result.h
```diff
@@ -13,7 +15,10 @@
 class ErrorMessage final {
  public:
   ErrorMessage() = default;
-  explicit ErrorMessage(std::string message) : message_(std::move(message)) {}
+  explicit ErrorMessage(std::string message) : message_{std::move(message)} {}
+  template <typename... Args>
+  explicit constexpr ErrorMessage(const absl::FormatSpec<Args...>& format, const Args&... args)
+      : message_{absl::StrFormat(format, args...)} {}
```
beckerhe, line 21: Is there a chance that this new constructor is considered over the `std::string`-constructor when converting from something else? Maybe we should require at least 2 arguments for this constructor (format string and one arg) to be on the safe side, WDYT?

beckerhe (approved):
Very nice. Thanks. Otherwise LGTM.

src/OrbitBase/include/OrbitBase/Result.h
dimitry-: it does not seem to help, gcc is still unhappy.

beckerhe: I don't see the problem directly either. I will have a look.

dimitry-:
well, that apparently works only for clang.. I am going to close this for now. Not sure if there is a solution that would satisfy gcc and msvc

beckerhe:
I'm sorry I haven't had time to really look into this. I believe there might be some SFINAE missing. But I need to understand how abseil handles FormatSpec.

PR #1955, opened 2021-03-01
Fix broken timers when command buffer begin timestamp is missing.

florian-kuebler:
If we miss a command buffer's begin -- as we were not already
capturing -- we send "0" as special timestamp. However,
we use the first command buffer of a submission as a reference
to align the Gpu timestamps to our Cpu timeline. If this timestamp
was missing, we failed and resulted in timers with length of days.

This fixes this (for now) on the client site by discarding the
complete submission if the first command buffer has no begin
marker affected. Note that the "align" debug markers to the
begin of capture if their "begin" info is missing, remains working.

Follow up: We should re-consider sending incomplete submissions.

Test: Take a capture in the first actual frames of infiltrator.
Bug: http://b/181195585

### DIFF
--- a/src/OrbitCaptureClient/GpuQueueSubmissionProcessor.cpp
+++ b/src/OrbitCaptureClient/GpuQueueSubmissionProcessor.cpp
@@ -159,6 +159,13 @@ std::vector<TimerInfo> GpuQueueSubmissionProcessor::ProcessGpuQueueSubmissionWit
   std::optional<GpuCommandBuffer> first_command_buffer =
       ExtractFirstCommandBuffer(gpu_queue_submission);
 
+  // The first command buffer acts as our reference needed to align GPU time based events in the
+  // CPU timeline. If we have no result for it -- which is the case if we started capturing within
+  // its execution -- we need to give up for this.
+  if (first_command_buffer.has_value() && first_command_buffer->begin_gpu_timestamp_ns() == 0) {
+    return result;
+  }
+
   std::vector<TimerInfo> command_buffer_timers =
       ProcessGpuCommandBuffers(gpu_queue_submission, matching_gpu_job, first_command_buffer,
                                timeline_key, get_string_hash_and_send_to_listener_if_necessary);
--- a/src/OrbitCaptureClient/GpuQueueSubmissionProcessor.cpp
+++ b/src/OrbitCaptureClient/GpuQueueSubmissionProcessor.cpp
@@ -161,7 +161,7 @@ std::vector<TimerInfo> GpuQueueSubmissionProcessor::ProcessGpuQueueSubmissionWit
 
   // The first command buffer acts as our reference needed to align GPU time based events in the
   // CPU timeline. If we have no result for it -- which is the case if we started capturing within
-  // its execution -- we need to give up for this.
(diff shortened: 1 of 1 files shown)


src/OrbitCaptureClient/GpuQueueSubmissionProcessor.cpp
```diff
@@ -159,6 +159,13 @@ std::vector<TimerInfo> GpuQueueSubmissionProcessor::ProcessGpuQueueSubmissionWit
   std::optional<GpuCommandBuffer> first_command_buffer =
       ExtractFirstCommandBuffer(gpu_queue_submission);
 
+  // The first command buffer acts as our reference needed to align GPU time based events in the
+  // CPU timeline. If we have no result for it -- which is the case if we started capturing within
+  // its execution -- we need to give up for this.
```
dpallotti, line 164: Could you rephrase "give up for this"? What does "this" refer to? The command buffer (i.e., the queue submission)?

src/OrbitCaptureClient/GpuQueueSubmissionProcessor.cpp
```diff
@@ -159,6 +159,13 @@ std::vector<TimerInfo> GpuQueueSubmissionProcessor::ProcessGpuQueueSubmissionWit
...
+  if (first_command_buffer.has_value() && first_command_buffer->begin_gpu_timestamp_ns() == 0) {
```
dpallotti, line 165: What happens if `!first_command_buffer.has_value()`? Is that handled correctly in the calls that follow?

florian-kuebler: Yes. The "first command buffer" is simply the first command buffer in this submission, if there are any. The code that follows always iterates of the command buffers, and does stuff for a concrete command buffer. Thus, in those cases the "first command buffer" is always set --  otherwise, we would just skip the loops.

src/OrbitCaptureClient/GpuQueueSubmissionProcessor.cpp
florian-kuebler: Done.

src/OrbitCaptureClient/GpuQueueSubmissionProcessor.cpp
```diff
@@ -159,6 +159,13 @@ std::vector<TimerInfo> GpuQueueSubmissionProcessor::ProcessGpuQueueSubmissionWit
   std::optional<GpuCommandBuffer> first_command_buffer =
       ExtractFirstCommandBuffer(gpu_queue_submission);
 
+  // The first command buffer acts as our reference needed to align GPU time based events in the
+  // CPU timeline. If we have no result for it -- which is the case if we started capturing within
```
dpallotti, line 163: Maybe change "result" to "initial GPU timestamp" (if I got it correctly)?

dpallotti (approved):

src/OrbitCaptureClient/GpuQueueSubmissionProcessor.cpp
florian-kuebler: Somewhat. I changed it to "first timestamp of the submission"

PR #1956, opened 2021-03-01
Handle inline code in ElfFile::GetLineInfo

beckerhe:
GetLineInfo used not to take inlining into account, meaning the
addresses from a particular function might point to source files all
over the place because parts have been inlined.

For the source code view we want samples to count for the code line in
the original function, not the inlined function.

So this change makes GetLineInfo take inlining into account and always
return the location of outmost inlining frame which is the function we
are interested in.


src/ElfUtils/ElfFile.cpp
```diff
@@ -294,25 +294,32 @@ std::filesystem::path ElfFileImpl<ElfT>::GetFilePath() const {
...
   if (!line_info_or_error) {
     return ErrorMessage(absl::StrFormat(
         "Unable to get line number info for \"%s\", address=0x%x: %s", object_file_->getFileName(),
         address, llvm::toString(line_info_or_error.takeError())));
   }
 
   auto& symbolizer_line_info = line_info_or_error.get();
+  const uint32_t number_of_frames = symbolizer_line_info.getNumberOfFrames();
+
+  // Getting back zero frames means there was some kind of problem. We will return a error.
```
dimitry-, line 308: interesting, if it is easy to reproduce maybe add a test for this use-case?

src/ElfUtils/ElfFile.cpp
```diff
@@ -294,25 +294,32 @@ std::filesystem::path ElfFileImpl<ElfT>::GetFilePath() const {
 template <typename ElfT>
 ErrorMessageOr<LineInfo> orbit_elf_utils::ElfFileImpl<ElfT>::GetLineInfo(uint64_t address) {
   CHECK(has_debug_info_section_);
-  auto line_info_or_error = symbolizer_.symbolizeCode(
-      *object_file_, {address, llvm::object::SectionedAddress::UndefSection});
+  auto line_info_or_error = symbolizer_.symbolizeInlinedCode(
+      object_file_->getFileName(), {address, llvm::object::SectionedAddress::UndefSection});
```
dimitry-, line 298: this probably deserves a test..

dimitry- (approved):
otherwise lgtm

src/ElfUtils/ElfFile.cpp
beckerhe: I probably have to recreate the testdata for that. Do we have a script for that?

dimitry-: I honestly do not remember, I need to find it and write a readme so that I do not forget again.

beckerhe: Don't worry. I will try to write a Makefile.

beckerhe: So I looked into this. And it's non-trivial to recreate the hello_world_elf as there are a lot of tests which check hard-coded addresses, especially in LinuxMapTest.cpp.

So I will create an additional elf file to test the inlining stuff.

src/ElfUtils/ElfFile.cpp
beckerhe: Ah, this is a tricky one. I haven't yet figured out when this conditon occurs and when the other one below occurs. So, unfortunately it's not so easy to reproduce.

src/ElfUtils/ElfFile.cpp
beckerhe: I added a test for this. @dimitry-  PTAL

dimitry-: > So I will create an additional elf file to test the inlining stuff.

Yeah, I remember now. we cannot compile because it won't produce stable result and without stable result we do not have expected values to test against.

src/ElfUtils/ElfFileTest.cpp
```diff
@@ -272,4 +272,81 @@ TEST(ElfFile, CalculateDebuglinkChecksumNotFound) {
...
+      optimized_symbols_result.value().symbol_infos().begin(),
+      optimized_symbols_result.value().symbol_infos().end());
+
+  const auto optimized_main_symbol =
+      std::find_if(optimized_symbols.begin(), optimized_symbols.end(),
+                   [](const SymbolInfo& symbol) { return symbol.name() == "main"; });
+  ASSERT_NE(optimized_main_symbol, optimized_symbols.end());
+
+  for (uint64_t offset = 0; offset < optimized_main_symbol->size(); ++offset) {
+    const auto& address = optimized_main_symbol->address() + offset;
```
dimitry-, line 340: why is this a reference?

src/ElfUtils/ElfFileTest.cpp
```diff
@@ -272,4 +272,81 @@ TEST(ElfFile, CalculateDebuglinkChecksumNotFound) {
...
+
+    auto line_info = optimized_program.value()->GetLineInfo(address);
+    ASSERT_TRUE(line_info) << line_info.error().message();
```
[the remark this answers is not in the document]
beckerhe: I rebased.

src/ElfUtils/ElfFileTest.cpp
```diff
@@ -272,4 +272,81 @@ TEST(ElfFile, CalculateDebuglinkChecksumNotFound) {
...
+
+    // Check whether the filepath matches
+    EXPECT_EQ(line_info.value().source_file(), reference_file_path);
+
+    // Check whether the reported line number falls into the reference program's range.
+    EXPECT_LE(min_line, line_info.value().source_line());
+    EXPECT_LE(line_info.value().source_line(), max_line);
+  }
```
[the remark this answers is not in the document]
beckerhe: I changed the test to do that. PTAL!

src/ElfUtils/ElfFileTest.cpp
```diff
@@ -272,4 +272,81 @@ TEST(ElfFile, CalculateDebuglinkChecksumNotFound) {
...
+  for (uint64_t offset = 0; offset < reference_main_symbol->size(); ++offset) {
+    const auto& address = reference_main_symbol->address() + offset;
+
+    auto line_info = reference_program.value()->GetLineInfo(address);
+    ASSERT_TRUE(line_info) << line_info.error().message();
+
+    reference_file_path = line_info.value().source_file();
+
+    min_line = std::min(min_line, line_info.value().source_line());
+    max_line = std::max(min_line, line_info.value().source_line());
```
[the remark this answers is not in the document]
beckerhe: done.

dimitry- (approved):
thanks!

PR #1958, opened 2021-03-01
Add source code report

beckerhe:
Just like DisassemblyReport this adds a class which implements the `CodeReport` interface for source code files.

### DIFF
--- a/src/OrbitGl/SourceCodeReport.cpp
+++ b/src/OrbitGl/SourceCodeReport.cpp
@@ -1,0 +1,50 @@
+// Copyright (c) 2021 The Orbit Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#include "SourceCodeReport.h"
+
+#include "OrbitBase/Logging.h"
+#include "OrbitClientData/PostProcessedSamplingData.h"
+
+namespace orbit_gl {
+
+SourceCodeReport::SourceCodeReport(std::string_view source_file,
+                                   const orbit_client_protos::FunctionInfo& function,
+                                   uint64_t absolute_address, orbit_elf_utils::ElfFile* elf_file,
+                                   const PostProcessedSamplingData& sampling_data,
+                                   uint32_t total_samples_in_capture)
+    : total_samples_in_capture_(total_samples_in_capture) {
+  for (size_t offset = 0; offset < function.size(); ++offset) {
+    const ThreadSampleData* summary = sampling_data.GetSummary();
+    const auto it = summary->raw_address_count.find(absolute_address + offset);
+    if (it == summary->raw_address_count.end()) continue;
+
+    const uint32_t current_samples = it->second;
+    if (current_samples == 0) continue;
+
+    const auto maybe_current_line_info = elf_file->GetLineInfo(function.address() + offset);
+    if (!maybe_current_line_info.has_value()) continue;
+
+    const auto& current_line_info = maybe_current_line_info.value();
+    if (source_file != current_line_info.source_file()) {
+      ERROR(
+          "Was trying to gather sampling data for function \"%s\" but the debug information "
+          "tells me the function address %#x is defined in a different source file.",
+          function.pretty_name(), function.address() + offset);
+      ERROR("Expected: %s", source_file);
+      ERROR("Actual: %s", current_line_info.source_file());
+      continue;
+    }
+
+    number_of_samples_per_line_[current_line_info.source_line()] += current_samples;
+    total_samples_in_function_ += current_samples;
+  }
+}
+
+uint32_t SourceCodeReport::GetNumSamplesAtLine(size_t line) const {
+  if (!number_of_samples_per_line_.contains(line)) return 0;
+  return number_of_samples_per_line_.at(line);
+}
+
+}  // namespace orbit_gl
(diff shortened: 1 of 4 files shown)


src/OrbitGl/SourceCodeReport.cpp
```diff
@@ -0,0 +1,51 @@
...
+
+    const uint32_t current_samples = it->second;
+    if (current_samples == 0) continue;
+
+    const auto maybe_current_line_info = elf_file->GetLineInfo(function.address() + offset);
+    if (!maybe_current_line_info.has_value()) continue;
+
+    const auto& current_line_info = maybe_current_line_info.value();
+    if (source_file != current_line_info.source_file()) {
+      LOG("Was trying to gather sampling data for function \"%s\" but the debug information "
```
florian-kuebler, line 31: Isn't that actually an error? When does this happen?

src/OrbitGl/SourceCodeReport.cpp
```diff
@@ -0,0 +1,51 @@
...
+          "tells me the function address %#x is defined in a different source file.",
+          function.pretty_name(), function.address() + offset);
+      LOG("Expected: %s", source_file);
+      LOG("Actual: %s", current_line_info.source_file());
+      continue;
+    }
+
+    number_of_samples_per_line[current_line_info.source_line()] += current_samples;
+    LOG("Attributed %d samples to line %d", current_samples, current_line_info.source_line());
```
florian-kuebler, line 40: Not sure if we want to keep that LOG. Looks like this may spam the output for long captures.

src/OrbitGl/SourceCodeReport.h
```diff
@@ -0,0 +1,46 @@
...
+                            uint32_t total_samples_in_capture);
+
+  [[nodiscard]] uint32_t GetNumSamplesInFunction() const override {
+    return total_samples_in_function_;
+  }
+  [[nodiscard]] uint32_t GetNumSamples() const override { return total_samples_in_capture_; }
+  [[nodiscard]] uint32_t GetNumSamplesAtLine(size_t line) const override;
+
+ private:
+  absl::flat_hash_map<size_t, size_t> number_of_samples_per_line;
```
florian-kuebler, line 39: "_" missing.

src/OrbitGl/SourceCodeReport.cpp
```diff
@@ -0,0 +1,51 @@
...
+    total_samples_in_function_ += current_samples;
+  }
+}
+
+uint32_t SourceCodeReport::GetNumSamplesAtLine(size_t line) const {
+  const auto it = number_of_samples_per_line.find(line);
+  if (it == number_of_samples_per_line.end()) return 0;
```
florian-kuebler, line 47: Personal preference: As this is a `absl` map, I'd go for `contains` over `find`. As this is a bit more readable (but of course a bit less efficient).

florian-kuebler:
In general looks good. Thanks!

src/OrbitGl/SourceCodeReport.h
antonrohr, line 39: I think:
```suggestion
  absl::flat_hash_map<uint32_t, uint32_t> number_of_samples_per_line;
```

antonrohr (approved):
besides the point @florian-kuebler raised, lgtm

src/OrbitGl/SourceCodeReport.cpp
beckerhe: This can happen with inlining, but should be fixed by #1956. It'd still prefer to keep this check, because LLVM's documentation is sparse and I am not sure if there might be any other condition leading to this. If it was possible it'd put a metric on this event, to see what's happening in the wild. But that's not yet possible.

src/OrbitGl/SourceCodeReport.cpp
beckerhe: Yes, good point. I forgot to remove that.

src/OrbitGl/SourceCodeReport.cpp
florian-kuebler: Then, I'd still argue to use ERROR over LOG.

beckerhe: Sure. No problem.

beckerhe: Done.

src/OrbitGl/SourceCodeReport.cpp
beckerhe: Done.

src/OrbitGl/SourceCodeReport.h
beckerhe: done.

src/OrbitGl/SourceCodeReport.cpp
beckerhe: Changed,

src/OrbitGl/SourceCodeReport.h
beckerhe: Done.

beckerhe:
@florian-kuebler PTAL

florian-kuebler (approved):
LGTM, thanks!

beckerhe:
> LGTM, thanks!

Thank you both for the reviews.

PR #1962, opened 2021-03-01
Simplify use of locks in our Vulkan layer's SubmissionTracker class

ronaldfw:
We simplify the use of locks in the SubmissionTracker class to make it
easier to reason about correctness of the code and avoid sometimes
leaking query slots.

For example, in methods like MarkCommandBufferBegin, the mutex was
locked, then freed again, then it was checked if we were capturing,
then the mutex was locked again to record the timestamp. It was possible
that capturing was stopped in the middle and OnCaptureFinished called.
This lead to query slots that should have been cleaned up in
OnCaptureFinished to not get cleaned up at all.

We also tighten the check on whether we are capturing and directly tie
this check to the calls to OnCaptureFinished. The previous use of the
IsCapturing() method sometimes led to imcomplete command buffer states
that were neither cleaned up in OnCaptureFinished nor properly
completed.

With this change, all query slots are now properly cleaned up after the 
call to OnCaptureFinished and after all remaining submits have been 
completed.

Tested: Manually ran Vulkan layer with two sample applications.
Bug: http://b/181097789

### DIFF
--- a/src/OrbitVulkanLayer/SubmissionTracker.h
+++ b/src/OrbitVulkanLayer/SubmissionTracker.h
@@ -209,6 +209,7 @@ class SubmissionTracker : public VulkanLayerProducer::CaptureStatusListener {
   }
 
   void MarkCommandBufferBegin(VkCommandBuffer command_buffer) {
+    absl::WriterMutexLock lock(&mutex_);
     // Even when we are not capturing we create state for this command buffer to allow the
     // debug marker tracking. In order to compute the correct depth of a debug marker and being able
     // to match an "end" marker with the corresponding "begin" marker, we maintain a stack of all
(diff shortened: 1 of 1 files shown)


src/OrbitVulkanLayer/SubmissionTracker.h
```diff
@@ -218,48 +219,44 @@ class SubmissionTracker : public VulkanLayerProducer::CaptureStatusListener {
...
-      absl::WriterMutexLock lock(&mutex_);
       CHECK(command_buffer_to_state_.contains(command_buffer));
       command_buffer_to_state_.at(command_buffer).command_buffer_begin_slot_index =
           std::make_optional(slot_index);
     }
   }
 
   void MarkCommandBufferEnd(VkCommandBuffer command_buffer) {
-    if (!IsCapturing()) {
+    absl::WriterMutexLock lock(&mutex_);
```
florian-kuebler, line 238: Just to mention that with the argument below, a read lock would be fine, but agree to make it simpler to reason about.

src/OrbitVulkanLayer/SubmissionTracker.h
```diff
@@ -499,6 +494,7 @@ class SubmissionTracker : public VulkanLayerProducer::CaptureStatusListener {
   // This method also resets all the timer slots that have been read.
   // It is assumed to be called periodically, e.g. on `vkQueuePresentKHR`.
   void CompleteSubmits(VkDevice device) {
+    absl::WriterMutexLock lock(&mutex_);
```
florian-kuebler, line 497: Should we really lock the complete present? I guess it is fine for now, however.

florian-kuebler (approved):
LGTM.

src/OrbitVulkanLayer/SubmissionTracker.h
ronaldfw: This is writing command_buffer_end_slot_index, which may also be written / reset by OnCaptureFinished at the same time. Or am I missing something?

florian-kuebler: Thanks. I was missing this.

src/OrbitVulkanLayer/SubmissionTracker.h
ronaldfw: Yes, I think the scope here can be a bit tighter, but the write lock is at least needed for the PullCompletedSubmissions call (it was there before) and the first loop here (which is potentially adding submissions back into the queue). I think I'll keep it as-is for simplicity but we can definitely keep this in mind for an improvement.

florian-kuebler: Sounds good.

PR #1964, opened 2021-03-01
Use effective_capture_start_timestamp_ns_ for the initial ThreadNames

dpallotti:
This makes sure that no event that follows the initial `ThreadName`s has a lower
timestamp. This was caught by `LinuxTracingIntegrationTest` failing, albeit very
rarely. Note that this is basically the same as what was already done, but
previously a new timestamp was taken at the beginning of
`RetrieveThreadNamesSystemWide`.


florian-kuebler (approved):
LGTM

PR #1967, opened 2021-03-01
Draw sample counts per line in source code

karupayun:
We want to include sample counts and percentages when showing source code
and disassembly view. This PR aims that. For that, we are creating a widget to 
be in the right, where we show that information. Also, we introduced a widget 
in the top with column titles.

This PR includes:

 - A right widget with information about sample counts 

 - A top-widget where we draw titles for every column ("Samples",
 "Function" & "Total").

 - Making the pop-up window wider

Related to http://b/159809330 from Source Code feature.

To test: Simply run CodeViewerDemo.

Screenshots with sample info, only to show how it looks:
Before:
dialog before

After:
source code

### DIFF
--- a/src/CodeViewer/Viewer.cpp
+++ b/src/CodeViewer/Viewer.cpp
@@ -14,6 +14,7 @@
 #include <QPlainTextEdit>
 #include <QRect>
 #include <QResizeEvent>
+#include <QScrollBar>
 #include <QString>
 #include <QStringList>
 #include <QTextBlock>
@@ -28,30 +29,45 @@ static const QColor kTextEditBackgroundColor{30, 30, 30};
 static const QColor kTextEditForegroundColor{189, 189, 189};
 static const QColor kHeatmapColor{Qt::red};
 
-int DetermineLineNumberWidthInPixels(const QFontMetrics& font_metrics, int max_line_number) {
-  return font_metrics.horizontalAdvance(QString::number(max_line_number));
+static int StringWidthInPixels(const QFontMetrics& font_metrics, QString string) {
+  return font_metrics.horizontalAdvance(string);
 }
 
-Viewer::Viewer(QWidget* parent) : QPlainTextEdit(parent), left_sidebar_widget_{this} {
-  UpdateLeftSidebarWidth();
-  QObject::connect(this, &QPlainTextEdit::blockCountChanged, this, &Viewer::UpdateLeftSidebarWidth);
+int DetermineLineNumberWidthInPixels(const QFontMetrics& font_metrics, int max_line_number) {
+  return StringWidthInPixels(font_metrics, QString::number(max_line_number));
+}
+
+Viewer::Viewer(QWidget* parent)
+    : QPlainTextEdit(parent), left_sidebar_widget_{this}, right_sidebar_widget_{this} {
+  UpdateSidebarsWidth();
+  QObject::connect(this, &QPlainTextEdit::blockCountChanged, this, &Viewer::UpdateSidebarsWidth);
 
   QObject::connect(&left_sidebar_widget_, &PlaceHolderWidget::PaintEventTriggered, this,
                    &Viewer::DrawLineNumbers);
 
-  const auto update_viewport_area = [&](const QRect& rect, int dy) {
-    const auto update_caused_by_scroll = [](int dy) { return dy != 0; };
+  QObject::connect(&right_sidebar_widget_, &PlaceHolderWidget::PaintEventTriggered, this,
+                   &Viewer::DrawSampleCounters);
 
-    if (update_caused_by_scroll(dy)) {
+  const auto update_viewport_area = [&](const QRect& rect, int dy) {
+    bool update_caused_by_scroll = (dy != 0);
+    if (update_caused_by_scroll) {
       left_sidebar_widget_.scroll(0, dy);
+      right_sidebar_widget_.scroll(0, dy);
     } else {
       QRect line_number_rect = rect;
+      QRect samples_info_rect = rect;
 
-      // We keep the (vertical) y-coordinates and adjust the horizontal x-coordinates for the line
-      // number area.
+      // We keep the (vertical) y-coordinates and adjust the horizontal x-coordinates for the each
+      // bar's area
       line_number_rect.setLeft(0);
       line_number_rect.setWidth(left_sidebar_widget_.sizeHint().width());
       left_sidebar_widget_.update(line_number_rect);
+
+      int total_width_without_scroll_bar = width() - verticalScrollBar()->width();
+      int right_sidebar_width = right_sidebar_widget_.sizeHint().width();
+      samples_info_rect.setLeft(total_width_without_scroll_bar - right_sidebar_width);
+      samples_info_rect.setWidth(right_sidebar_width);
+      right_sidebar_widget_.update(line_number_rect);
     }
   };
   QObject::connect(this, &QPlainTextEdit::updateRequest, this, update_viewport_area);
@@ -72,15 +88,14 @@ Viewer::Viewer(QWidget* parent) : QPlainTextEdit(parent), left_sidebar_widget_{t
 void Viewer::resizeEvent(QResizeEvent* ev) {
   QPlainTextEdit::resizeEvent(ev);
 
-  auto left_sidebar = contentsRect();
-  left_sidebar.setWidth(left_sidebar_widget_.sizeHint().width());
-  left_sidebar_widget_.setGeometry(left_sidebar);
+  UpdateSidebarPositions();
 }
 
 void Viewer::wheelEvent(QWheelEvent* ev) {
   QPlainTextEdit::wheelEvent(ev);
 
-  UpdateLeftSidebarWidth();
+  UpdateSidebarsWidth();
+  UpdateSidebarPositions();
 }
 
 void Viewer::DrawLineNumbers(QPaintEvent* event) {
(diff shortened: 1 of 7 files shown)


beckerhe (commented):
Very nice. Thanks Pablo. Looks good in general. I added some comments, mostly minor.

src/CodeViewer/Viewer.cpp
```diff
@@ -28,27 +29,37 @@ static const QColor kTextEditBackgroundColor{30, 30, 30};
...
 
     if (update_caused_by_scroll(dy)) {
       left_sidebar_widget_.scroll(0, dy);
+      right_sidebar_widget_.scroll(0, dy);
     } else {
       QRect line_number_rect = rect;
 
       // We keep the (vertical) y-coordinates and adjust the horizontal x-coordinates for the line
       // number area.
+      // TODO: Is this really needed? Which update can break it?
```
florian-kuebler, line 62: You are going to address this TODO within this PR, right?

src/CodeViewer/Viewer.cpp
```diff
@@ -72,15 +83,14 @@ Viewer::Viewer(QWidget* parent) : QPlainTextEdit(parent), left_sidebar_widget_{t
 void Viewer::resizeEvent(QResizeEvent* ev) {
   QPlainTextEdit::resizeEvent(ev);
 
-  auto left_sidebar = contentsRect();
-  left_sidebar.setWidth(left_sidebar_widget_.sizeHint().width());
-  left_sidebar_widget_.setGeometry(left_sidebar);
+  UpdateSidebarPositions();
```
florian-kuebler, line 86: Why is it not also UpdateSidebarWidth?

src/CodeViewer/Viewer.cpp
```diff
@@ -122,47 +132,143 @@ void Viewer::DrawLineNumbers(QPaintEvent* event) {
   }
 }
 
-void Viewer::UpdateLeftSidebarWidth() {
+// For example, from a value = 0.5, we want to print "50.00 %"
+std::string PercentageFromDouble(double v) {
```
florian-kuebler, line 136: Nit: `PercentageStringFromDouble` or even `DoubleToPercentageString`?

src/CodeViewer/Viewer.cpp
```diff
@@ -122,47 +132,143 @@ void Viewer::DrawLineNumbers(QPaintEvent* event) {
...
+  std::clamp(v, 0., 1.);
```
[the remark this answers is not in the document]
florian-kuebler: Ideally we would have tests at least for the free functions. However, this can be done post-release, I guess.

src/CodeViewer/Viewer.cpp
```diff
@@ -194,4 +300,20 @@ void Viewer::HighlightCurrentLine() {
...
 
+int Viewer::WidthPercentageColumn() const {
+  const std::string kWidestPercentage = "100.00 %";
+  return StringWidthInPixels(fontMetrics(), kWidestPercentage);
+}
+
+int Viewer::WidthSampleCounterColumn() const {
+  const std::string kSampleColumnTitle = "Samples";
+  return StringWidthInPixels(fontMetrics(),
+                             kSampleColumnTitle);  // TODO: titles might have smaller font
```
florian-kuebler, line 311: Again, are you addressing this in this PR?

florian-kuebler (approved):
LGTM, some minor comments. Thanks!

src/CodeViewer/Viewer.cpp
karupayun: Yes

src/CodeViewer/Viewer.cpp
karupayun: Yes

src/CodeViewer/Viewer.cpp
karupayun: Maybe I am wrong, but because they only change when the font-size change (so, after zooming in-out and not a resize event)?

src/CodeViewer/Viewer.cpp
```diff
@@ -28,27 +29,37 @@ static const QColor kTextEditBackgroundColor{30, 30, 30};
...
+  QObject::connect(this, &QPlainTextEdit::blockCountChanged, this, &Viewer::UpdateSidebarsWidth);
 
   QObject::connect(&left_sidebar_widget_, &PlaceHolderWidget::PaintEventTriggered, this,
                    &Viewer::DrawLineNumbers);
 
+  QObject::connect(&right_sidebar_widget_, &PlaceHolderWidget::PaintEventTriggered, this,
+                   &Viewer::DrawSampleCounters);
+
   const auto update_viewport_area = [&](const QRect& rect, int dy) {
     const auto update_caused_by_scroll = [](int dy) { return dy != 0; };
```
vickyliu-go4it, line 52: I'm not very clear about the purpose for using a lambda expression here. `update_cuased_by_scroll` seems only used in line 54. Why we are not using a boolean `update_cuased_by_scroll` here?

src/CodeViewer/Viewer.cpp
florian-kuebler: That might be the case. Basically wanted to point you to this, as before it has the "left_sidebar.SetWidth".

vickyliu-go4it (approved):
Nice PR! Thanks for that. I agree Florian's comments on the TODOs. If they will not be addressed in this PR, may be you can put the relevant bug in the TODOs.

LGTM!

src/CodeViewer/Viewer.cpp
```diff
@@ -122,47 +132,143 @@ void Viewer::DrawLineNumbers(QPaintEvent* event) {
   }
 }
 
-void Viewer::UpdateLeftSidebarWidth() {
+// For example, from a value = 0.5, we want to print "50.00 %"
+std::string PercentageFromDouble(double v) {
+  std::clamp(v, 0., 1.);
+  int printed_digits = static_cast<int>(round(v * 10000.));
+  return (std::to_string(printed_digits / 100) + "." + std::to_string(printed_digits % 100 / 10) +
+          std::to_string(printed_digits % 10) + " %");
```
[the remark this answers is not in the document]
karupayun: Nice link. Will switch to QString!

src/CodeViewer/Viewer.cpp
```diff
@@ -122,47 +132,143 @@ void Viewer::DrawLineNumbers(QPaintEvent* event) {
...
 
-  setViewportMargins(QMargins{overall_width_px, 0, 0, 0});
-  left_sidebar_widget_.SetSizeHint({overall_width_px, 0});
+void Viewer::UpdateSidebarPositions() {
+  auto left_sidebar = contentsRect();
+  int total_width = left_sidebar.width();
+  left_sidebar.setWidth(left_sidebar_widget_.sizeHint().width());
+  left_sidebar_widget_.setGeometry(left_sidebar);
+
+  auto slider_width = QAbstractScrollArea().verticalScrollBar()->sizeHint().width();
```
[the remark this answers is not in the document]
karupayun: Thanks! BTW, it is not solving the issue :(

src/CodeViewer/Viewer.cpp
karupayun: Yes, but we aren't changing the width there. If you read UpdateSidebarPositions we are also making that SetWidth.

This is code that was there before, but basically:
1 - Create a new left_sidebar with the entire windows size (  auto left_sidebar = contentsRect(); )
2- We set the width to the previous width (  left_sidebar.setWidth(left_sidebar_widget_.sizeHint().width()); )
3- We change our left_sidebar for the new one.

All these steps is to simply move the left_sidebar to the new position of the window without changing their size.

src/CodeViewer/Viewer.cpp
karupayun: This was here before, and I also found it strange. I think that it was used to give the user a better explanation than a simply if(dy != 0). I changed it, What do you think?

src/CodeViewer/Viewer.cpp
karupayun: Actually, I want @beckerhe opinion here.

src/CodeViewer/Viewer.cpp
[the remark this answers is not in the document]
beckerhe: Too bad.

src/CodeViewer/Viewer.cpp
beckerhe: I think this is needed. At least it doesn't hurt.

beckerhe (approved):
Thank you for addressing my remarks.

src/CodeViewer/Viewer.cpp
beckerhe: `UpdateSidebarWidth` is only needed when something occured which leads to a different width of the sidebar. Since line numbers are shown there, one event that could cause a width-change is changing the shown code. A different number of lines might need more or less sidebar width due to more or fewer digits.

The same applies for changing the settings (enabling/disabling linenumbers/heatmap).

But in this event handler the only thing that actually changes is the size of the viewport, which means we might need to position the sidebar somewhere else, but we don't need to update its width.

Does that make sense?

src/CodeViewer/Viewer.cpp
beckerhe: I had a closer look into the qt docs. There are two situations here:

1. We need to handle scrolling - which is clear I guess
2. We need to handle content updates (changes of the source code)

The second doesn't really appear in our use-case because it's read-only. But It's still good style to handle it here. Just imagine we implement two different styles of line-break in the future. most editors allow to visually break a line when it is too long. We could have the same. And if the user toggles that every line with such a visual line break will have twice the height from before. And that's what's updated here.

beckerhe: TLDR: Yes, we need that! It's our duty to handle that case, even if it is unlikely right now, that this code path is taken.

src/CodeViewer/Viewer.cpp
florian-kuebler: thanks

src/CodeViewer/Viewer.cpp
vickyliu-go4it: Thanks for the explanation and making the change :-) I understand the purpose for the readability. But just curious why not a simple `bool update_cuased_by_scroll = (dy != 0)` like we are using in other cases.

src/CodeViewer/Viewer.cpp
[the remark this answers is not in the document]
karupayun: Solved by using MoveLeft instead of MoveRight! :)

src/CodeViewer/Viewer.cpp
karupayun: I changed. WDYT?

src/CodeViewer/Viewer.cpp
karupayun: Nice suggestion!! I don't know why, but we are using it now :)

karupayun:
I know that everyone did LGTM, but I'm asking for re-reviews. Sorry for that.

New changes: 
 - Addressed comments and TODO/bugs
 - We are using real data
 - We have a top_bar_widget now
 - The pop-up windows is bigger
 
 Change I need to do tomorrow:
  - Rebase on Henning's PR to include optional data (and not draw any number)
  
 Change I won't do:
  - Problem with zoom-in a lot (will set a bug tomorrow)

vickyliu-go4it (approved):
Very nice and clear PR! And I really like the way that you organize tasks and TODOs in the PR! Thanks for that!

src/CodeViewer/Viewer.cpp
```diff
@@ -198,4 +355,21 @@ void Viewer::HighlightCurrentLine() {
   setExtraSelections({selection});
 }
 
+int Viewer::WidthPercentageColumn() const {
```
florian-kuebler, line 358: This (and the next two functions) can be also static, right?

florian-kuebler (approved):
LGTM. Consider making the three functions static.

src/CodeViewer/Viewer.cpp
karupayun: Yes, I need to add "static" to my mental checking list

beckerhe: AFAIK, they need to be member function because the access `fontMetrics()`.

florian-kuebler: Oh, sorry. Haven't noticed this one. Thanks!

karupayun: Yes, Henning is right. Sorry for my comment and thanks Henning!

beckerhe (approved):

src/CodeViewer/Viewer.cpp
karupayun: ``  QRect titles_bar = rect;
      QRect line_number_rect = rect;
      QRect samples_info_rect = rect;

      // We keep the (vertical) y-coordinates and adjust the horizontal x-coordinates for the each
      // bar's area
      titles_bar.setLeft(0);
      titles_bar.setWidth(top_bar_widget_.sizeHint().width());
      top_bar_widget_.update(line_number_rect);

      line_number_rect.setLeft(0);
      line_number_rect.setWidth(left_sidebar_widget_.sizeHint().width());
      left_sidebar_widget_.update(line_number_rect);

      samples_info_rect.setLeft(0);
      samples_info_rect.setWidth(right_sidebar_widget_.sizeHint().width());
      right_sidebar_widget_.update(line_number_rect); ``

karupayun: Reverted the change:
 
     ```
 QRect titles_bar = rect;
      QRect line_number_rect = rect;
      QRect samples_info_rect = rect;

      // We keep the (vertical) y-coordinates and adjust the horizontal x-coordinates for the each
      // bar's area
      titles_bar.setLeft(0);
      titles_bar.setWidth(top_bar_widget_.sizeHint().width());
      top_bar_widget_.update(line_number_rect);

      line_number_rect.setLeft(0);
      line_number_rect.setWidth(left_sidebar_widget_.sizeHint().width());
      left_sidebar_widget_.update(line_number_rect);

      samples_info_rect.setLeft(0);
      samples_info_rect.setWidth(right_sidebar_widget_.sizeHint().width());
      right_sidebar_widget_.update(line_number_rect);
```

File: src/CodeViewer/Viewer.cpp [hunk 10]
```diff
@@ -26,32 +27,57 @@ static const QColor kLineNumberBackgroundColor{50, 50, 50};
...
+      titles_bar.setWidth(top_bar_widget_.sizeHint().width());
+      top_bar_widget_.update(line_number_rect);
 
-      // We keep the (vertical) y-coordinates and adjust the horizontal x-coordinates for the line
-      // number area.
       line_number_rect.setLeft(0);
       line_number_rect.setWidth(left_sidebar_widget_.sizeHint().width());
       left_sidebar_widget_.update(line_number_rect);
+
+      samples_info_rect.setLeft(0);
```
beckerhe, line 78: I think the left coordinate needs to be the same as in `UpdateBarPositions`, so `total_width_without_scroll_bar - right_sidebar_width`

src/CodeViewer/Viewer.cpp
```diff
@@ -26,32 +27,57 @@ static const QColor kLineNumberBackgroundColor{50, 50, 50};
...
     } else {
+      QRect titles_bar = rect;
       QRect line_number_rect = rect;
+      QRect samples_info_rect = rect;
+
+      // We keep the (vertical) y-coordinates and adjust the horizontal x-coordinates for the each
+      // bar's area
+      titles_bar.setLeft(0);
+      titles_bar.setWidth(top_bar_widget_.sizeHint().width());
+      top_bar_widget_.update(line_number_rect);
```
beckerhe, line 72: The title bar doesn't need to be updated at all, because a text change can't change anything here. I think you can remove it entirely.

src/CodeViewer/Viewer.cpp
```diff
@@ -128,47 +192,156 @@ void Viewer::DrawLineNumbers(QPaintEvent* event) {
...
+  };
+
+  for (auto block = firstVisibleBlock(); block.isValid() && top_of(block) <= event->rect().bottom();
+       block = block.next()) {
+    if (!block.isVisible() || bottom_of(block) < event->rect().top()) continue;
+
+    const int left = left_margin_.ToPixels(fontMetrics());
+
+    auto opt_num_samples_in_lines = code_report_->GetNumSamplesAtLine(block.blockNumber() + 1);
+    if (opt_num_samples_in_lines.has_value()) {
```
beckerhe, line 222: Nit: `if (!opt_num_samples_in_lines.has_value()) return;`

src/OrbitQt/orbitmainwindow.cpp
```diff
@@ -1375,6 +1376,7 @@ void OrbitMainWindow::ShowSourceCode(const std::filesystem::path& file_path, siz
   orbit_code_viewer::Dialog code_viewer_dialog{this};
 
   code_viewer_dialog.SetEnableLineNumbers(true);
+  code_viewer_dialog.SetEnableSampleCounters(true);
```
beckerhe, line 1379: I think this should only be enabled when `maybe_code_report.has_value()` is true. Can you move it down to line 1390?

beckerhe (approved):
Otherwise LGTM

src/CodeViewer/Viewer.cpp
karupayun: I wasn't sure there, but I will change it!

src/OrbitQt/orbitmainwindow.cpp
karupayun: nice catch!

src/CodeViewer/Viewer.cpp
karupayun: Of course, nice catch again!

src/CodeViewer/Viewer.cpp
karupayun: BTW, I should write _continue_ instead, not return

PR #1968, opened 2021-03-01
Provide CodeReport to code viewer dialog

beckerhe:
This is needed to get samples counts in the code viewer.

### DIFF
--- a/src/CodeViewer/Viewer.cpp
+++ b/src/CodeViewer/Viewer.cpp
@@ -100,10 +100,14 @@ void Viewer::DrawLineNumbers(QPaintEvent* event) {
        block = block.next()) {
     if (!block.isVisible() || bottom_of(block) < event->rect().top()) continue;
 
-    if (heatmap_bar_width_.Value() > 0.0 && heatmap_source_) {
+    if (heatmap_bar_width_.Value() > 0.0 && code_report_ != nullptr) {
       const QRect heatmap_rect{0, top_of(block), heatmap_bar_width_.ToPixels(fontMetrics()),
                                fontMetrics().height()};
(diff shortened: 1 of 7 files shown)


src/CodeViewer/Viewer.cpp
```diff
@@ -100,10 +100,14 @@ void Viewer::DrawLineNumbers(QPaintEvent* event) {
        block = block.next()) {
     if (!block.isVisible() || bottom_of(block) < event->rect().top()) continue;
 
-    if (heatmap_bar_width_.Value() > 0.0 && heatmap_source_) {
+    if (heatmap_bar_width_.Value() > 0.0 && code_report_ != nullptr) {
```
karupayun, line 103: I would split these 2 conditions.

Here only if (code_report_ != nullptr) and the width of the heatmap before calculating intensity, what do you think? 

BTW, as we can skip conflicts between our PRs, I can change it in my PR. I'm currently thinking how to mix it when drawing the samples column.

karupayun (approved):
LGTM Henning! :)

src/CodeViewer/Viewer.cpp
beckerhe: I don't understand. There is no work that I can do with either of these conditions not true.

beckerhe: What code should be executed when code_report != nullptr, but heatmap_bar_width.Value() == 0?

beckerhe: As discussed OOB, we will follow up tomorrow and I'm going to merge this as-is.

karupayun: Maybe we could want to hide the heatmap when hiding line numbers but letting the samples in?

PR #1969, opened 2021-03-02
Integrate SourcePathsMapping into source code loading

beckerhe:
This adds the source paths mapping step to the show source code process.

When the path found in debuginfo is not found locally, the source paths
mappings will be loaded and tried to apply. If no mapping succeeds the
user will be prompted to choose the file via the file open dialog.

### DIFF
--- a/src/OrbitQt/orbitmainwindow.cpp
+++ b/src/OrbitQt/orbitmainwindow.cpp
@@ -6,6 +6,7 @@
 
 #include <absl/flags/declare.h>
 
+#include <QAbstractButton>
 #include <QAction>
 #include <QApplication>
 #include <QCheckBox>
@@ -84,6 +85,8 @@
 #include "Path.h"
 #include "QtUtils/MainThreadExecutorImpl.h"
 #include "SamplingReport.h"
+#include "SourcePathsMapping/Mapping.h"
+#include "SourcePathsMapping/MappingManager.h"
 #include "StatusListenerImpl.h"
 #include "SyntaxHighlighter/Cpp.h"
 #include "SyntaxHighlighter/X86Assembly.h"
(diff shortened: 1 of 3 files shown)


src/OrbitQt/orbitmainwindow.cpp
```diff
@@ -1269,21 +1272,88 @@ void OrbitMainWindow::ShowTooltip(std::string_view message) {
...
+                          QString("Could not find the source code file \"%1\" on this machine.")
+                              .arg(QString::fromStdString(file_path.string())),
+                          QMessageBox::Cancel, this};
+  auto* pick_file_button = message_box.addButton("Choose file...", QMessageBox::ActionRole);
+
+  constexpr const char* kAutocreateMappingKey = "auto_create_mapping";
+  constexpr const char* kPreviousSourcePathsMappingDirectoryKey =
+      "previous_source_paths_mapping_directory";
+  auto checkbox = std::make_unique<QCheckBox>(
+      "Automatically create a source paths mapping from my chosen file.");
```
antonrohr, line 1297: Pers pref: Ever since the whole Orbit doc got reviewed, I like `selected` more than `chosen`.

src/OrbitQt/orbitmainwindow.cpp
```diff
@@ -1269,21 +1272,88 @@ void OrbitMainWindow::ShowTooltip(std::string_view message) {
...
+  if (user_chosen_file.isEmpty()) return std::nullopt;
+
+  settings.setValue(kAutocreateMappingKey, message_box.checkBox()->isChecked());
+  settings.setValue(kPreviousSourcePathsMappingDirectoryKey, QFileInfo{user_chosen_file}.path());
+  if (message_box.checkBox()->isChecked()) {
+    auto maybe_mapping = orbit_source_paths_mapping::InferMappingFromExample(
+        file_path, std::filesystem::path{user_chosen_file.toStdString()});
+    if (maybe_mapping.has_value()) {
+      orbit_source_paths_mapping::MappingManager mapping_manager{};
+      mapping_manager.AppendMapping(maybe_mapping.value());
```
IrinaShkviro, line 1330: Should we append mapping only if we can open it successfully?

src/OrbitQt/orbitmainwindow.cpp
```diff
@@ -1269,21 +1272,88 @@ void OrbitMainWindow::ShowTooltip(std::string_view message) {
...
+  auto checkbox = std::make_unique<QCheckBox>(
+      "Automatically create a source paths mapping from my chosen file.");
+  checkbox->setToolTip(
+      "If enabled Orbit will automatically try to create a source paths mapping from it. The "
+      "common suffix between the path given in the debug information and the local file path will "
+      "be stripped. From the rest a mapping will be created.");
+  QSettings settings{};
+  checkbox->setCheckState(settings.value(kAutocreateMappingKey, true).toBool() ? Qt::Checked
+                                                                               : Qt::Unchecked);
+  message_box.setCheckBox(checkbox.release());
```
antonrohr, line 1305: This line took me a minute to understand. A comment about transferring of ownership would be helpful

src/OrbitQt/orbitmainwindow.cpp
```diff
@@ -1269,21 +1272,88 @@ void OrbitMainWindow::ShowTooltip(std::string_view message) {
                      QString::fromUtf8(message.data(), static_cast<int>(message.size())), this);
 }
 
+std::optional<QString> OrbitMainWindow::LoadSourceCode(const std::filesystem::path& file_path) {
```
IrinaShkviro, line 1275: There are several sequential steps in this function: load the file itself, then if it's not successful try to use existing mappings, then if not successful ask user to provide a mapping... I think with comments about these steps it would be easier to read this function

IrinaShkviro (approved):
LGTM, I've left couple of small comments

src/OrbitQt/orbitmainwindow.cpp
beckerhe: I don't like comments too much, but I can split this up into several functions which is actually a good idea. Thanks!

src/OrbitQt/orbitmainwindow.cpp
beckerhe: I don't have a strong opinion for either ways. So yes, I can do that.

src/OrbitQt/orbitmainwindow.cpp
```diff
@@ -1269,21 +1272,88 @@ void OrbitMainWindow::ShowTooltip(std::string_view message) {
...
+
+    user_chosen_file = QFileDialog::getOpenFileName(
+        this, QString{"Choose %1"}.arg(QString::fromStdString(file_path.string())),
+        previous_directory.filePath(file_name), file_name);
+    message_box.accept();
+  });
+
+  message_box.exec();
+
+  if (user_chosen_file.isEmpty()) return std::nullopt;
```
antonrohr, line 1321: pers pref:
```suggestion
    if (user_chosen_file.isEmpty()) {
      message_box.cancel();
      return
    } 
    message_box.accept();
  });

  QDialogCode rc = message_box.exec();

  if (rc == QDialog::Rejected) return std::nullopt;
```

antonrohr (approved):
just nits. thank you!

src/OrbitQt/orbitmainwindow.cpp
beckerhe: Done.

src/OrbitQt/orbitmainwindow.cpp
beckerhe: Great. Thanks. I'm always unsure about wording.

src/OrbitQt/orbitmainwindow.cpp
beckerhe: Done.

src/OrbitQt/orbitmainwindow.cpp
beckerhe: Done.

src/OrbitQt/orbitmainwindow.cpp
beckerhe: Done.

PR #1970, opened 2021-03-02
Reassing field numbers in *CaptureEvent protos

dimitry-:
Use numbers 1-16 for high frequency events
and number starting with 16 for events with lower frequency

Bug: http://b/181014656

### DIFF
--- a/src/GrpcProtos/capture.proto
+++ b/src/GrpcProtos/capture.proto
@@ -286,42 +286,60 @@ message SystemMemoryUsage {
 
 message ClientCaptureEvent {
   oneof event {
+    // Note that field numbers from 1-15 take only 1 byte to encode
+    // https://developers.google.com/protocol-buffers/docs/proto#assigning_field_numbers
+    // use them for high frequency events. For the rest please assign
+    // numbers starting with 16.
+    //
     // Please keep these alphabetically ordered.
-    AddressInfo address_info = 1;
-    CallstackSample callstack_sample = 2;
-    FunctionCall function_call = 3;
-    GpuJob gpu_job = 4;
-    GpuQueueSubmission gpu_queue_submission = 5;
-    InternedCallstack interned_callstack = 6;
-    InternedString interned_string = 7;
-    InternedTracepointInfo interned_tracepoint_info = 8;
-    IntrospectionScope introspection_scope = 9;
-    ModuleUpdateEvent module_update_event = 10;
-    SchedulingSlice scheduling_slice = 11;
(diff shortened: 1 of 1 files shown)


src/GrpcProtos/capture.proto
```diff
@@ -273,40 +273,50 @@ message ModuleUpdateEvent {
...
-    InternedCallstack interned_callstack = 6;
-    InternedString interned_string = 7;
-    InternedTracepointInfo interned_tracepoint_info = 8;
-    IntrospectionScope introspection_scope = 9;
-    ModuleUpdateEvent module_update_event = 10;
-    SchedulingSlice scheduling_slice = 11;
-    ThreadName thread_name = 12;
-    ThreadStateSlice thread_state_slice = 13;
-    TracepointEvent tracepoint_event = 14;
+    AddressInfo address_info = 16;
```
dpallotti, line 282: I think that this can be somewhat high frequency (one for each sampled address), although it's hard to give numbers. For comparison, I think it can be higher frequency than `gpu_job`. Same for `full_address_info` in `ProcuderCaptureEvent`.

dimitry-: yeah, in this case I was kind of looking into the future where we do not have address_info for frame-pointer based unwinding.

dpallotti: Ah alright, then I'll leave the decision to you. Maybe add a comment about it.

dpallotti (approved):

src/GrpcProtos/capture.proto
```diff
@@ -273,40 +273,50 @@ message ModuleUpdateEvent {
...
+    CallstackSample callstack_sample = 1;
+    FunctionCall function_call = 2;
+    GpuJob gpu_job = 3;
+    GpuQueueSubmission gpu_queue_submission = 4;
+    InternedCallstack interned_callstack = 17;
```
dimitry-, line 287: I am not quite sure about this event.

dpallotti: Me neither... One would expect low frequency because it's interned, but of course it's easy to have many different callstacks. Does this become more or less critical once we increase the sampling frequency with frame pointers?

src/GrpcProtos/capture.proto
dimitry-: added a comment.

src/GrpcProtos/capture.proto
dimitry-: Still unsure but moved to 1-15 range.

PR #1971, opened 2021-03-02
Integrate Sample statistics in SourceCodeView

beckerhe:
The integrates sample statistics for the source code view. This has already been there for the disassembly view for a very long time. Now it's also coming for the source code view.

For now the sample data will only fill the heatmap on the left. Pablo is working on bringing actual numbers to a right side bar.

### DIFF
--- a/src/OrbitGl/App.cpp
+++ b/src/OrbitGl/App.cpp
@@ -69,6 +69,7 @@
 #include "Path.h"
 #include "PresetsDataView.h"
 #include "SamplingReport.h"
+#include "SourceCodeReport.h"
 #include "StringManager.h"
 #include "SymbolHelper.h"
 #include "TimeGraph.h"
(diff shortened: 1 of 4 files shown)


src/OrbitGl/App.cpp
```diff
@@ -545,16 +546,32 @@ void OrbitApp::ShowSourceCode(const orbit_client_protos::FunctionInfo& function)
           [this, module,
            function](const std::filesystem::path& local_file_path) -> ErrorMessageOr<void> {
             const auto elf_file = orbit_elf_utils::ElfFile::Create(local_file_path);
-            const auto line_info = elf_file.value()->GetLineInfo(function.address());
+            const auto first_line_or_error = elf_file.value()->GetLineInfo(function.address());
```
dpallotti, line 549: I would suggest `first_line_info_or_error`. It matches `line_info` below better. Or just `line_info_or_error`.

src/OrbitQt/orbitmainwindow.cpp
```diff
@@ -1370,7 +1370,8 @@ std::optional<QString> OrbitMainWindow::LoadSourceCode(const std::filesystem::pa
   return std::nullopt;
 }
 
-void OrbitMainWindow::ShowSourceCode(const std::filesystem::path& file_path, size_t line_number) {
+void OrbitMainWindow::ShowSourceCode(const std::filesystem::path& file_path, size_t line_number,
+                                     std::optional<std::unique_ptr<CodeReport>> maybe_code_report) {
```
dpallotti, line 1374: The name of the parameter should match the name in the interface and in the declaration.

src/OrbitGl/MainWindowInterface.h
```diff
@@ -20,7 +23,8 @@ namespace orbit_gl {
 class MainWindowInterface {
  public:
   virtual void ShowTooltip(std::string_view message) = 0;
-  virtual void ShowSourceCode(const std::filesystem::path& file_path, size_t line_number) = 0;
+  virtual void ShowSourceCode(const std::filesystem::path& file_path, size_t line_number,
+                              std::optional<std::unique_ptr<CodeReport>> code_report) = 0;
```
dpallotti, line 27: `optional`+`unique_ptr` looks particularly ugly to me. Is it a problem to only use `unique_ptr` and just pass a `unique_ptr` holding `nullptr` when the `CodeReport` is not available? If you rename to `maybe_code_report` like in the definition of `OrbitMainWindow::ShowSourceCode` then it should be clear enough.

dpallotti (approved):
I would really move away from `std::optional<std::unique_ptr<CodeReport>>` but I don't know if you feel the same. Looks good otherwise.

src/OrbitGl/MainWindowInterface.h
beckerhe: I know, it's ugly. But it's so much better than having bugs because of unexpected nullpointers. And we had a lot of them.

I will change the name!

beckerhe:
> I would really move away from `std::optional<std::unique_ptr<CodeReport>>` but I don't know if you feel the same. Looks good otherwise.

I rather feel strong about this. I could have two overloads, one without the pointer and one with. Would that be better?

dpallotti:
> I could have two overloads, one without the pointer and one with. Would that be better?

I'm not sure, maybe, but then again that's then two methods to override and at that point `optional` works just fine. I'll leave it to you.

src/OrbitGl/MainWindowInterface.h
antonrohr: I wanted to suggest `std::optional<CodeReport>`, but since this is not possible, I like `std::optional<std::unique_ptr<CodeReport>>` more than just `unique_ptr`

antonrohr (approved):

beckerhe:
> > I could have two overloads, one without the pointer and one with. Would that be better?
> 
> I'm not sure, maybe, but then again that's then two methods to override and at that point `optional` works just fine. I'll leave it to you.

Thanks. In that case I tend to leave it as-is.

src/OrbitGl/App.cpp
beckerhe: Done. Thanks.

src/OrbitQt/orbitmainwindow.cpp
beckerhe: Done. Thanks.

IrinaShkviro (approved):

PR #1972, opened 2021-03-02
Gracefully handle running out of query slots in our Vulkan layer

ronaldfw:
We simply CHECKed that we found a query slot so far, which sometimes
failed on some of our test applications. We now handle this case
gracefully by not aborting and continuing execution, however without
recording timestamps for command buffers and markers where we don't get
one of the slots (either begin or end).

Due to the logic in how we distribute slots, once we run out of slots
it usually happens that no further command buffer and debug marker
timeslices are created, as all slots are given to begin events. We free
them but freed ones are immediately given to begin slots again.

There is also a performance issue in NextReadyQuerySlot when running out
of slots as this function runs over all slots (currently 65536) every
time a slot is requested. This will be fixed in a follow-up change.

Tested: Manually ran layer on two sample applications with both small
and large numbers of total slots (100, 1000, 20000, 65535).
Bug: http://b/181097789

### DIFF
--- a/src/OrbitVulkanLayer/SubmissionTracker.h
+++ b/src/OrbitVulkanLayer/SubmissionTracker.h
@@ -65,7 +65,7 @@ class SubmissionTracker : public VulkanLayerProducer::CaptureStatusListener {
   // associated to the `SubmissionMetaInformation`.
   struct SubmittedCommandBuffer {
     std::optional<uint32_t> command_buffer_begin_slot_index;
-    uint32_t command_buffer_end_slot_index;
+    std::optional<uint32_t> command_buffer_end_slot_index;
 
     // These are set when the timestamps were queried successfully. When these have values, the
     // corresponding slot indices for timestamp queries have been reset. Do not query again in
@@ -226,8 +226,8 @@ class SubmissionTracker : public VulkanLayerProducer::CaptureStatusListener {
       return;
     }
 
-    uint32_t slot_index = RecordTimestamp(command_buffer, VK_PIPELINE_STAGE_TOP_OF_PIPE_BIT);
-    {
+    uint32_t slot_index;
+    if (RecordTimestamp(command_buffer, VK_PIPELINE_STAGE_TOP_OF_PIPE_BIT, &slot_index)) {
       CHECK(command_buffer_to_state_.contains(command_buffer));
       command_buffer_to_state_.at(command_buffer).command_buffer_begin_slot_index =
           std::make_optional(slot_index);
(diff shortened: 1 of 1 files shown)


src/OrbitVulkanLayer/SubmissionTracker.h
```diff
@@ -362,6 +373,10 @@ class SubmissionTracker : public VulkanLayerProducer::CaptureStatusListener {
       }
     }
 
+    if (!query_slots_to_reset.empty()) {
+      timer_query_pool_->ResetQuerySlots(device, query_slots_to_reset);
```
florian-kuebler, line 377: This is a bit problematic. We would set the slot directly as available again. However, the slot is wired into the command buffer, being submitted. Thus, it will write into this slot index and our results might be wrong, if the slot is being picked up again.
However, this will be addressed by my change with the re-use of command buffers. Could you add a TODO for now, you can reference the b/181098020.

florian-kuebler (approved):
LGTM!

src/OrbitVulkanLayer/SubmissionTracker.h
ronaldfw: > However, the slot is wired into the command buffer, being submitted.

It's possible I misunderstand something, but we reset the slot in the command buffer above (the new "if" above). Doesn't that mean that we don't get the timestamp there?

ronaldfw: Ah, I get it, it's in the command buffer we send to the GPU. Yes, that's a problem.

florian-kuebler: So in the if above:
end is nullopt, begin has lets say slot i. So in the command buffer, there is a "CmdWriteTimestamp(i)".
This command buffer is submitted exactly now, so it will be executed.
What we are doing here, is reset the slot on the hardware (done immediately, so before the write timestamp command was executed) and mark the slot as free again. Further, above, we ensure, that we will not read the slot in the present.

Now if right after this code, another command buffer (or marker or so) begins, and thus requests a slot, the pool may reserve "i" for it (as it was resetted).
Now we have two command buffers writing into the same slot. Thus undefined what the result will be. Still that case is very very unlikely. Especially, as the new command buffer must be created, submitted and finished with the execution, BEFORE, the one from the current submission executes, but still, it is possible.

ronaldfw: Added a TODO and comment as you suggested. I think it's not that much of a problem right now as this only happens in the case when we run out of slots, in which case we typically don't get any timeslices anyway. But should be addressed at some point.

florian-kuebler: But as I said, this is solvable with the exact same logic, I am currently implementing. We would mark those slots as "done reading", and just wait until their command buffers get "reset" to free the slot again.

PR #1974, opened 2021-03-02
Make CodeReport::NumSamplesAtLine return an optional

beckerhe:
This way we can differentiate between lines without any samples and
lines without any data.

The latter occurs in source code files. We always show the whole file
but there is only data available for a single function (might change in
the future). Hence we need a way to indicate to the user whether data is
available or not.

### DIFF
--- a/src/CodeViewer/Viewer.cpp
+++ b/src/CodeViewer/Viewer.cpp
@@ -104,10 +104,12 @@ void Viewer::DrawLineNumbers(QPaintEvent* event) {
       const QRect heatmap_rect{0, top_of(block), heatmap_bar_width_.ToPixels(fontMetrics()),
                                fontMetrics().height()};
 
-      const size_t num_samples_in_line = code_report_->GetNumSamplesAtLine(block.blockNumber() + 1);
-      const size_t num_samples_in_function = code_report_->GetNumSamplesInFunction();
-      const float intensity =
-          std::clamp(static_cast<float>(num_samples_in_line) / num_samples_in_function, 0.0f, 1.0f);
+      const uint32_t num_samples_in_line =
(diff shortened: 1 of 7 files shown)


src/CodeViewer/Viewer.cpp
```diff
@@ -104,10 +104,12 @@ void Viewer::DrawLineNumbers(QPaintEvent* event) {
       const QRect heatmap_rect{0, top_of(block), heatmap_bar_width_.ToPixels(fontMetrics()),
                                fontMetrics().height()};
 
-      const size_t num_samples_in_line = code_report_->GetNumSamplesAtLine(block.blockNumber() + 1);
-      const size_t num_samples_in_function = code_report_->GetNumSamplesInFunction();
-      const float intensity =
-          std::clamp(static_cast<float>(num_samples_in_line) / num_samples_in_function, 0.0f, 1.0f);
+      const uint32_t num_samples_in_line =
+          code_report_->GetNumSamplesAtLine(block.blockNumber() + 1).value_or(0);
```
beckerhe, line 108: The important change is here `.value_or(0)`. The rest is some minor cleanup.

karupayun (approved):
Nice Henning! Thanks for this commit! :)

What do you prefer to show in CodeSamples when there is no data?
 - Nothing at all
 - "--" string

beckerhe:
> Nice Henning! Thanks for this commit! :)
> 
> What do you prefer to show in CodeSamples when there is no data?
> 
> * Nothing at all
> * "--" string

I think I have to see it to know for sure what I like best.

I would try showing nothing at all first. And if that worked well, I would just stick to it.

antonrohr (approved):

PR #1975, opened 2021-03-02
Address flakiness of LinuxTracingIntegrationTest

dpallotti:
Bug: http://b/181634108

### DIFF
--- a/src/LinuxTracing/UprobesUnwindingVisitor.cpp
+++ b/src/LinuxTracing/UprobesUnwindingVisitor.cpp
@@ -44,7 +44,21 @@ void UprobesUnwindingVisitor::visit(StackSamplePerfEvent* event) {
   const std::vector<unwindstack::FrameData>& libunwindstack_callstack = unwinder_.Unwind(
       current_maps_.get(), event->GetRegisters(), event->GetStackData(), event->GetStackSize());
 
+  // LibunwindstackUnwinder::Unwind signals an unwinding error with an empty callstack.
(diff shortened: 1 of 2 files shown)


src/LinuxTracing/UprobesUnwindingVisitor.cpp
```diff
@@ -44,13 +44,24 @@ void UprobesUnwindingVisitor::visit(StackSamplePerfEvent* event) {
...
 
+  // LibunwindstackUnwinder::Unwind signals an unwinding error with an empty callstack.
   if (libunwindstack_callstack.empty()) {
     if (unwind_error_counter_ != nullptr) {
       ++(*unwind_error_counter_);
     }
     return;
   }
 
+  // Callstacks with only one frame (the sampled address) are also unwinding errors, that were not
```
florian-kuebler, line 55: What is if we have a sample inside main?

dpallotti: This is not the case of main because there are other frames outside of main.

florian-kuebler: I see. Could you add that concrete statement to the comment? Otherwise, LGTM.

florian-kuebler (approved):

src/LinuxTracing/UprobesUnwindingVisitor.cpp
dpallotti: Good point, done.

PR #1976, opened 2021-03-02
Optimize finding free timestamp query slots in our Vulkan layer

ronaldfw:
We optimize finding a free timestamp query slot. The previous code ran
over a vector of all possible slots until a free one (based on the
vector entry) was found. This was problematic when many, and in
particular all, slots were occupied already, leading to massive game
slowdown and a lot of CPU time spent in finding the next free slot.

This is mostly noticeable when running out of query slots. With a
maximum of 10'000 slots, the previous code had ~84% of exclusive
callstack samples in TimerQueryPool::NextReadyQuerySlot. For the new
code, this number went down to ~0.14%.

For validation purposes we keep the vector of slot states and check
allocated slots and reset slots against the expected state.

Tested: Manually ran Vulkan layer on two test applications. Profiled
Orbit layer / game performance using Orbit.
Bug: http://b/181639638

### DIFF
--- a/src/OrbitVulkanLayer/TimerQueryPool.h
+++ b/src/OrbitVulkanLayer/TimerQueryPool.h
@@ -5,6 +5,9 @@
 #ifndef ORBIT_VULKAN_LAYER_TIMER_QUERY_POOL_H_
 #define ORBIT_VULKAN_LAYER_TIMER_QUERY_POOL_H_
 
+#include <numeric>
+#include <vector>
+
 #include "OrbitBase/Logging.h"
 #include "absl/container/flat_hash_map.h"
 #include "absl/synchronization/mutex.h"
@@ -51,7 +54,10 @@ class TimerQueryPool {
       std::vector<SlotState> slots{num_timer_query_slots_};
       std::fill(slots.begin(), slots.end(), SlotState::kReadyForQueryIssue);
       device_to_query_slots_[device] = slots;
-      device_to_potential_next_free_index_[device] = 0;
+      std::vector<uint32_t> free_slots(num_timer_query_slots_);
+      // At the beginning all slot indices in [0, num_timer_query_slots) are free.
+      std::iota(free_slots.begin(), free_slots.end(), 0);
+      device_to_free_slots_[device] = free_slots;
     }
   }
 
@@ -61,7 +67,7 @@ class TimerQueryPool {
     CHECK(device_to_query_pool_.contains(device));
     VkQueryPool query_pool = device_to_query_pool_.at(device);
     device_to_query_slots_.erase(device);
-    device_to_potential_next_free_index_.erase(device);
+    device_to_free_slots_.erase(device);
 
     dispatch_table_->DestroyQueryPool(device)(device, query_pool, nullptr);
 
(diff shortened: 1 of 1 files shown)


src/OrbitVulkanLayer/TimerQueryPool.h
```diff
@@ -51,7 +54,9 @@ class TimerQueryPool {
       std::vector<SlotState> slots{num_timer_query_slots_};
       std::fill(slots.begin(), slots.end(), SlotState::kReadyForQueryIssue);
       device_to_query_slots_[device] = slots;
-      device_to_potential_next_free_index_[device] = 0;
+      std::vector<uint32_t> free_slots(num_timer_query_slots_);
+      std::iota(free_slots.begin(), free_slots.end(), 0);
```
florian-kuebler, line 58: Maybe add a comment, that at the beginning, every slot is considered empty (I needed to look up the iota method).

src/OrbitVulkanLayer/TimerQueryPool.h
```diff
@@ -76,30 +81,24 @@ class TimerQueryPool {
...
-  // In case it successfully found a slot, the index will be written to the given `allocated_index`.
+  // Returns a free query slot from the device's pool if one still exists. It returns `false` if all
+  // slots are occupied and true otherwise. If successful, the index will be written to the given
+  // `allocated_index`.
   //
   // Note that the pool must be initialized using `InitializeTimerQueryPool` before.
   // See also `ResetQuerySlots` to make occupied slots available again.
   [[nodiscard]] bool NextReadyQuerySlot(VkDevice device, uint32_t* allocated_index) {
     absl::WriterMutexLock lock(&mutex_);
-    CHECK(device_to_potential_next_free_index_.contains(device));
```
florian-kuebler, line 87: That should probably be a CHECK(device_to_free_slots_.contains(device));

src/OrbitVulkanLayer/TimerQueryPool.h
```diff
@@ -76,30 +81,24 @@ class TimerQueryPool {
...
-
-    return false;
+    std::vector<uint32_t>& free_slots = device_to_free_slots_.at(device);
+    if (free_slots.empty()) {
+      return false;
+    }
+    *allocated_index = free_slots.back();
+    CHECK(device_to_query_slots_.at(device)[*allocated_index] == SlotState::kReadyForQueryIssue);
+    device_to_query_slots_.at(device)[*allocated_index] = SlotState::kQueryPendingOnGpu;
+    free_slots.pop_back();
```
florian-kuebler, line 100: Nit/Personal preference: I would move the free_slots.pop_back up to the free_slots.back();

florian-kuebler (approved):
Thanks. LGTM!

src/OrbitVulkanLayer/TimerQueryPool.h
ronaldfw: Done.

src/OrbitVulkanLayer/TimerQueryPool.h
ronaldfw: Yes, good idea. Done.

src/OrbitVulkanLayer/TimerQueryPool.h
ronaldfw: Yes, agree. Done.

src/OrbitVulkanLayer/TimerQueryPool.h
florian-kuebler: Thanks!

src/OrbitVulkanLayer/TimerQueryPool.h
florian-kuebler: Thanks!

src/OrbitVulkanLayer/TimerQueryPool.h
florian-kuebler: Thanks!

ronaldfw:
Thanks for the review, @florian-kuebler

PR #1977, opened 2021-03-02
Fix SourcePathsMapping test

beckerhe:
It seems a race condition occured and the file testfile.txt was moved to
a different location, which means the tests can't find it anymore.

This commit fixes the test by using a different file.


IrinaShkviro (approved):
lgtm

beckerhe:
> lgtm

Thanks Irina!

PR #1978, opened 2021-03-02
Integrate Source Paths Mapping Editor

beckerhe:
The PRs adds the source paths mappings editor to Orbit. It allows users to change source paths mappings which have been
automatically infered and it also allows them to add new ones, remove old ones, etc.

The editor can be launched from the capture options dialog:
CaptureOptions

The editor looks like this:
SourcePathsMappings


vickyliu-go4it (approved):
Nice PR! Thanks for that and LGTM.

PR #1979, opened 2021-03-02
Factor out methods to access memory in tracee.

danielfenner:
Previously these lived in the unnamed namespace of the AllocateInTracee
implementation. Turns out I'll need these again for injecting a dynamic
lib. So they are moved into a separate file and written in a slightly
more general way.

bug: http://b/181304023
test: Unit test in PR.

### DIFF
--- a/src/UserSpaceInstrumentation/AccessTraceesMemory.cpp
+++ b/src/UserSpaceInstrumentation/AccessTraceesMemory.cpp
@@ -32,7 +32,7 @@ using orbit_base::ReadFileToString;
     const uint64_t data = ptrace(PTRACE_PEEKDATA, pid, address_start + i * 8, NULL);
     if (errno) {
       return ErrorMessage(
-          absl::StrFormat("Failed to PTRACE_PEEKDATA for pid: %d with errno %d: \"%s\"", pid, errno,
+          absl::StrFormat("Failed to PTRACE_PEEKDATA for pid %d with errno %d: \"%s\"", pid, errno,
                           SafeStrerror(errno)));
     }
     std::memcpy(bytes.data() + (i * sizeof(uint64_t)), &data, 8);
@@ -55,8 +55,7 @@ using orbit_base::ReadFileToString;
   return outcome::success();
 }
 
-[[nodiscard]] ErrorMessageOr<void> GetFirstExecutableMemoryRegion(pid_t pid, uint64_t* addr_start,
-                                                                  uint64_t* addr_end) {
+[[nodiscard]] ErrorMessageOr<AddressRange> GetFirstExecutableMemoryRegion(pid_t pid) {
   auto result_read_maps = ReadFileToString(absl::StrFormat("/proc/%d/maps", pid));
   if (result_read_maps.has_error()) {
     return result_read_maps.error();
@@ -68,9 +67,10 @@ using orbit_base::ReadFileToString;
     if (tokens.size() < 2 || tokens[1].size() != 4 || tokens[1][2] != 'x') continue;
     const std::vector<std::string> addresses = absl::StrSplit(tokens[0], '-');
     if (addresses.size() != 2) continue;
-    if (!absl::numbers_internal::safe_strtou64_base(addresses[0], addr_start, 16)) continue;
-    if (!absl::numbers_internal::safe_strtou64_base(addresses[1], addr_end, 16)) continue;
-    return outcome::success();
+    AddressRange result;
+    if (!absl::numbers_internal::safe_strtou64_base(addresses[0], &result.first, 16)) continue;
+    if (!absl::numbers_internal::safe_strtou64_base(addresses[1], &result.second, 16)) continue;
+    return result;
   }
   return ErrorMessage(absl::StrFormat("Unable to locate executable memory area in pid: %d", pid));
 }
--- a/src/UserSpaceInstrumentation/AccessTraceesMemory.cpp
+++ b/src/UserSpaceInstrumentation/AccessTraceesMemory.cpp
@@ -14,8 +14,8 @@
 #include <string>
 
 #include "OrbitBase/Logging.h"
-#include "OrbitBase/SafeStrerror.h"
 #include "OrbitBase/ReadFileToString.h"
+#include "OrbitBase/SafeStrerror.h"
 
 namespace orbit_user_space_instrumentation {
 
@@ -43,7 +43,7 @@ using orbit_base::ReadFileToString;
   do {
     // Pack 8 byte for writing into `data`.
     uint64_t data = 0;
-    std::memcpy(&data, bytes.data() + pos , sizeof(uint64_t));
+    std::memcpy(&data, bytes.data() + pos, sizeof(uint64_t));
     if (ptrace(PTRACE_POKEDATA, pid, address_start + pos, data) == -1) {
       ErrorMessage(absl::StrFormat("Unable to write data into tracees (pid: %d) memory.", pid));
     }
--- a/src/UserSpaceInstrumentation/AccessTraceesMemory.cpp
+++ b/src/UserSpaceInstrumentation/AccessTraceesMemory.cpp
@@ -1,0 +1,78 @@
+// Copyright (c) 2021 The Orbit Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#include "AccessTraceesMemory.h"
+
+#include <absl/strings/numbers.h>
+#include <absl/strings/str_format.h>
+#include <absl/strings/str_split.h>
+#include <sys/ptrace.h>
+
+#include <cerrno>
+#include <cstring>
+#include <string>
+
+#include "OrbitBase/Logging.h"
+#include "OrbitBase/SafeStrerror.h"
+#include "OrbitBase/ReadFileToString.h"
+
+namespace orbit_user_space_instrumentation {
+
+using orbit_base::ReadFileToString;
+
+[[nodiscard]] ErrorMessageOr<void> ReadTraceesMemory(pid_t pid, uint64_t address_start,
+                                                     uint64_t length, std::vector<uint8_t>* bytes) {
+  // Round up length to next multiple of eight.
+  length = ((length + 7) / 8) * 8;
+  bytes->resize(length);
+  for (size_t i = 0; i < length / 8; i++) {
+    const uint64_t data = ptrace(PTRACE_PEEKDATA, pid, address_start + i * 8, NULL);
+    if (errno) {
+      return ErrorMessage(absl::StrFormat("Failed to PTRACE_PEEKDATA with errno %d: \"%s\"", errno,
+                                          SafeStrerror(errno)));
+    }
+    std::memcpy(bytes->data() + (i * sizeof(uint64_t)), &data, 8);
+  }
+  return outcome::success();
+}
+
+[[nodiscard]] ErrorMessageOr<void> WriteTraceesMemory(pid_t pid, uint64_t address_start,
+                                                      const std::vector<uint8_t>& bytes) {
+  size_t pos = 0;
+  do {
+    // Pack 8 byte for writing into `data`.
+    uint64_t data = 0;
+    for (size_t i = 0; i < sizeof(data) && pos + i < bytes.size(); i++) {
+      uint64_t t = bytes[pos + i];
+      data |= t << (8 * i);
+    }
+    if (ptrace(PTRACE_POKEDATA, pid, address_start + pos, data) == -1) {
+      ErrorMessage(absl::StrFormat("Unable to write data into tracees (pid: %d) memory.", pid));
+    }
+    pos += 8;
+  } while (pos < bytes.size());
+  return outcome::success();
+}
+
+[[nodiscard]] ErrorMessageOr<void> GetFirstExecutableMemoryRegion(pid_t pid, uint64_t* addr_start,
+                                                                  uint64_t* addr_end) {
+  auto result_read_maps = ReadFileToString(absl::StrFormat("/proc/%d/maps", pid));
+  if (result_read_maps.has_error()) {
+    return result_read_maps.error();
+  }
+  const std::vector<std::string> lines =
+      absl::StrSplit(result_read_maps.value(), '\n', absl::SkipEmpty());
+  for (const auto& line : lines) {
+    const std::vector<std::string> tokens = absl::StrSplit(line, ' ', absl::SkipEmpty());
+    if (tokens.size() < 2 || tokens[1].size() != 4 || tokens[1][2] != 'x') continue;
+    const std::vector<std::string> addresses = absl::StrSplit(tokens[0], '-');
+    if (addresses.size() != 2) continue;
+    if (!absl::numbers_internal::safe_strtou64_base(addresses[0], addr_start, 16)) continue;
+    if (!absl::numbers_internal::safe_strtou64_base(addresses[1], addr_end, 16)) continue;
+    return outcome::success();
+  }
+  return ErrorMessage(absl::StrFormat("Unable to locate executable memory area in pid: %d", pid));
+}
+
+}  // namespace orbit_user_space_instrumentation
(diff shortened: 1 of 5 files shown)


src/UserSpaceInstrumentation/AllocateInTracee.cpp
```diff
@@ -99,14 +55,19 @@ void WriteBytesIntoTraceesMemory(pid_t pid, uint64_t address_start,
...
 
   // Backup first 8 bytes.
-  const uint64_t backup_module_code = ptrace(PTRACE_PEEKDATA, pid, address_start, NULL);
-  if (errno) {
-    return ErrorMessage(absl::StrFormat("Failed to PTRACE_PEEKDATA with errno %d: \"%s\"", errno,
-                                        SafeStrerror(errno)));
+  std::vector<uint8_t> original_data;
+  auto result_backup_code = ReadTraceesMemory(pid, address_start, 8, &original_data);
+  if (result_backup_code.has_error()) {
+    return ErrorMessage(absl::StrFormat("Failed to read from tracees: \"%s\"",
```
dpallotti, line 61:
```suggestion
    return ErrorMessage(absl::StrFormat("Failed to read from tracee: \"%s\"",
```

src/UserSpaceInstrumentation/AllocateInTracee.cpp
```diff
@@ -99,14 +55,19 @@ void WriteBytesIntoTraceesMemory(pid_t pid, uint64_t address_start,
...
+                                        result_backup_code.error().message()));
   }
 
   // Write `syscall` into memory. Machine code is `0x0f05`.
-  WriteBytesIntoTraceesMemory(pid, address_start, std::vector<uint8_t>{0x0f, 0x05});
+  auto result_write_code = WriteTraceesMemory(pid, address_start, std::vector<uint8_t>{0x0f, 0x05});
+  if (result_write_code.has_error()) {
+    return ErrorMessage(
+        absl::StrFormat("Failed to write to tracees: \"%s\"", result_write_code.error().message()));
```
dpallotti, line 69:
```suggestion
        absl::StrFormat("Failed to write to tracee: \"%s\"", result_write_code.error().message()));
```

src/UserSpaceInstrumentation/AccessTraceesMemoryTest.cpp
```diff
@@ -0,0 +1,67 @@
...
+#include <vector>
+
+#include "AccessTraceesMemory.h"
+#include "OrbitBase/Logging.h"
+#include "UserSpaceInstrumentation/Attach.h"
+#include "UserSpaceInstrumentation/RegisterState.h"
+
+namespace orbit_user_space_instrumentation {
+
+namespace {}  // namespace
```
dpallotti, line 24: Remove.

src/UserSpaceInstrumentation/AccessTraceesMemory.cpp
```diff
@@ -43,7 +43,7 @@ using orbit_base::ReadFileToString;
   do {
     // Pack 8 byte for writing into `data`.
     uint64_t data = 0;
-    std::memcpy(&data, bytes.data() + pos , sizeof(uint64_t));
+    std::memcpy(&data, bytes.data() + pos, sizeof(uint64_t));
```
dimitry-, line 46: absl::bit_cast<> does the same if I am not mistaken. It does require the data to be aligned though, I guess it should be in this case?

src/UserSpaceInstrumentation/AccessTraceesMemory.h
```diff
@@ -0,0 +1,39 @@
...
+
+#include "OrbitBase/Result.h"
+
+namespace orbit_user_space_instrumentation {
+
+// Read `length` bytes from process `pid` starting at `address_start`. `length` will be rounded up
+// to a multiple of eight. The data is returned in `bytes`.
+// Assumes we are already attached to the tracee `pid` e.g. using `AttachAndStopProcess`.
+[[nodiscard]] ErrorMessageOr<void> ReadTraceesMemory(pid_t pid, uint64_t address_start,
+                                                     uint64_t length, std::vector<uint8_t>* bytes);
```
dpallotti, line 21: I think that this function should return `ErrorMessageOr<std::vector<uint8_t>>` instead of using an output parameter.

src/UserSpaceInstrumentation/AccessTraceesMemory.cpp
```diff
@@ -0,0 +1,75 @@
...
+
+[[nodiscard]] ErrorMessageOr<void> WriteTraceesMemory(pid_t pid, uint64_t address_start,
+                                                      const std::vector<uint8_t>& bytes) {
+  size_t pos = 0;
+  do {
+    // Pack 8 byte for writing into `data`.
+    uint64_t data = 0;
+    std::memcpy(&data, bytes.data() + pos, sizeof(uint64_t));
+    if (ptrace(PTRACE_POKEDATA, pid, address_start + pos, data) == -1) {
+      ErrorMessage(absl::StrFormat("Unable to write data into tracees (pid: %d) memory.", pid));
```
dpallotti, line 48: OCD thought: This message is very different from the one in `ReadTraceesMemory`, but the two functions have a clear parallel. I would make them similar if it's possible.

src/UserSpaceInstrumentation/AccessTraceesMemory.cpp
```diff
@@ -0,0 +1,75 @@
...
+#include "OrbitBase/SafeStrerror.h"
+
+namespace orbit_user_space_instrumentation {
+
+using orbit_base::ReadFileToString;
+
+[[nodiscard]] ErrorMessageOr<void> ReadTraceesMemory(pid_t pid, uint64_t address_start,
+                                                     uint64_t length, std::vector<uint8_t>* bytes) {
+  // Round up length to next multiple of eight.
+  length = ((length + 7) / 8) * 8;
```
dpallotti, line 27: I would define a constant for the 8, something like `kPeekdataReadSize` or similar.

src/UserSpaceInstrumentation/AccessTraceesMemory.cpp
```diff
@@ -0,0 +1,75 @@
...
+    std::memcpy(bytes->data() + (i * sizeof(uint64_t)), &data, 8);
+  }
+  return outcome::success();
+}
+
+[[nodiscard]] ErrorMessageOr<void> WriteTraceesMemory(pid_t pid, uint64_t address_start,
+                                                      const std::vector<uint8_t>& bytes) {
+  size_t pos = 0;
+  do {
+    // Pack 8 byte for writing into `data`.
```
dpallotti, line 44:
```suggestion
    // Pack 8 bytes for writing into `data`.
```

src/UserSpaceInstrumentation/AccessTraceesMemory.cpp
danielfenner: absl::bit_cast only works with types of identical size:

https://github.com/abseil/abseil-cpp/blob/a76698790753d2ec71f655cdc84d61bcb27780d4/absl/base/casts.h#L42

So I'll stick with the memcpy.

src/UserSpaceInstrumentation/AllocateInTracee.cpp
danielfenner: done

src/UserSpaceInstrumentation/AllocateInTracee.cpp
danielfenner: done

src/UserSpaceInstrumentation/AccessTraceesMemoryTest.cpp
danielfenner: done

src/UserSpaceInstrumentation/AccessTraceesMemory.h
danielfenner: done

src/UserSpaceInstrumentation/AccessTraceesMemory.cpp
danielfenner: Changed it a bit. The functions are more different than it might appear though: PTRACE_POKEDATA returns -1 on error. PTRACE_PEEKDATA sets errno.

src/UserSpaceInstrumentation/AccessTraceesMemory.cpp
danielfenner: done

src/UserSpaceInstrumentation/AccessTraceesMemory.cpp
danielfenner: `kEight` maybe? :) I don't think that helps.

dpallotti: Well no, but that's because it repeats the content. 
But it helps if it reflects the reason for the content, for example why you round to the next multiple of 8. If I get it correctly this is because `PTRACE_PEEKDATA` reads "a **word** at the address addr in the tracee's memory". That's why I suggested `kPeekdataReadSize`.

src/UserSpaceInstrumentation/AccessTraceesMemory.cpp
```diff
@@ -0,0 +1,78 @@
...
+                                                                     uint64_t address_start,
+                                                                     uint64_t length) {
+  // Round up length to next multiple of eight.
+  length = ((length + 7) / 8) * 8;
+  std::vector<uint8_t> bytes(length);
+  for (size_t i = 0; i < length / 8; i++) {
+    const uint64_t data = ptrace(PTRACE_PEEKDATA, pid, address_start + i * 8, NULL);
+    if (errno) {
+      return ErrorMessage(
+          absl::StrFormat("Failed to PTRACE_PEEKDATA for pid: %d with errno %d: \"%s\"", pid, errno,
```
dpallotti, line 35:
```suggestion
          absl::StrFormat("Failed to PTRACE_PEEKDATA for pid %d with errno %d: \"%s\"", pid, errno,
```
It flows better.

src/UserSpaceInstrumentation/AccessTraceesMemory.cpp
```diff
@@ -0,0 +1,78 @@
...
+
+[[nodiscard]] ErrorMessageOr<void> WriteTraceesMemory(pid_t pid, uint64_t address_start,
+                                                      const std::vector<uint8_t>& bytes) {
+  size_t pos = 0;
+  do {
+    // Pack 8 bytes for writing into `data`.
+    uint64_t data = 0;
+    std::memcpy(&data, bytes.data() + pos, sizeof(uint64_t));
+    if (ptrace(PTRACE_POKEDATA, pid, address_start + pos, data) == -1) {
+      ErrorMessage(absl::StrFormat("Failed to PTRACE_POKEDATA for pid: %d.", pid));
```
dpallotti, line 51: Isn't `errno` set even if `ptrace` itself also returns a status code with `POKEDATA`?

src/UserSpaceInstrumentation/AccessTraceesMemory.cpp
```diff
@@ -0,0 +1,78 @@
...
+    }
+    pos += sizeof(uint64_t);
+  } while (pos < bytes.size());
+  return outcome::success();
+}
+
+[[nodiscard]] ErrorMessageOr<void> GetFirstExecutableMemoryRegion(pid_t pid, uint64_t* addr_start,
```
dpallotti, line 58: I think Dimitry mentioned in the previous change, but have you considered returning a `pair` or a dedicated `struct` instead of using two out parameters?

src/UserSpaceInstrumentation/AccessTraceesMemoryTest.cpp
```diff
@@ -0,0 +1,65 @@
...
+
+  auto result_read_back = ReadTraceesMemory(pid, address_start, address_end - address_start);
+  CHECK(result_read_back.has_value());
+
+  EXPECT_EQ(new_data, result_read_back.value());
+
+  CHECK(WriteTraceesMemory(pid, address_start, result_backup.value()).has_value());
+
+  // Detach and end child.
+  CHECK(DetachAndContinueProcess(pid).has_value());
```
dpallotti, line 60: Maybe `!has_error()` makes a bit more sense than `has_value()` with these functions that return `ErrorMessageOr<void>`.

src/UserSpaceInstrumentation/AllocateInTracee.cpp
```diff
@@ -99,14 +55,18 @@ void WriteBytesIntoTraceesMemory(pid_t pid, uint64_t address_start,
   }
 
   // Backup first 8 bytes.
-  const uint64_t backup_module_code = ptrace(PTRACE_PEEKDATA, pid, address_start, NULL);
-  if (errno) {
-    return ErrorMessage(absl::StrFormat("Failed to PTRACE_PEEKDATA with errno %d: \"%s\"", errno,
-                                        SafeStrerror(errno)));
+  auto result_backup_code = ReadTraceesMemory(pid, address_start, 8);
+  if (result_backup_code.has_error()) {
+    return ErrorMessage(absl::StrFormat("Failed to read from tracee: \"%s\"",
```
dpallotti, line 60: Maybe "from memory of tracee"?

src/UserSpaceInstrumentation/AllocateInTracee.cpp
```diff
@@ -99,14 +55,18 @@ void WriteBytesIntoTraceesMemory(pid_t pid, uint64_t address_start,
...
+                                        result_backup_code.error().message()));
   }
 
   // Write `syscall` into memory. Machine code is `0x0f05`.
-  WriteBytesIntoTraceesMemory(pid, address_start, std::vector<uint8_t>{0x0f, 0x05});
+  auto result_write_code = WriteTraceesMemory(pid, address_start, std::vector<uint8_t>{0x0f, 0x05});
+  if (result_write_code.has_error()) {
+    return ErrorMessage(
+        absl::StrFormat("Failed to write to tracee: \"%s\"", result_write_code.error().message()));
```
dpallotti, line 68: Maybe "to memory of tracee"?

dpallotti (approved):
Looks good otherwise.

src/UserSpaceInstrumentation/AccessTraceesMemory.cpp
danielfenner: done

src/UserSpaceInstrumentation/AccessTraceesMemory.cpp
danielfenner: To my understanding errno is only set by the  PTRACE_PEEK* stuff since the retunr value is user for the actual data. At least this is how I understand this:

  https://man7.org/linux/man-pages/man2/ptrace.2.html

   _C library/kernel differences
       At the system call level, the PTRACE_PEEKTEXT, PTRACE_PEEKDATA,
       and PTRACE_PEEKUSER requests have a different API: they store the
       result at the address specified by the data parameter, and the
       return value is the error flag.  The glibc wrapper function
       provides the API given in DESCRIPTION above, with the result
       being returned via the function return value._

dpallotti: Yep, sounds like it.

src/UserSpaceInstrumentation/AccessTraceesMemory.cpp
danielfenner: Funny! The style guide evolved a lot from what I remembered. This was the recommended way of doing things up until 2017. Let's go with a type alias for a pair.

src/UserSpaceInstrumentation/AccessTraceesMemoryTest.cpp
danielfenner: Yeah, this is a bit funny ("has value of type void"). Probably just a mildly poor wording in the result class. Changed for WriteTraceesMemory and  DetachAndContinueProcess.

src/UserSpaceInstrumentation/AllocateInTracee.cpp
danielfenner: done

src/UserSpaceInstrumentation/AllocateInTracee.cpp
danielfenner: done

PR #1980, opened 2021-03-02
Fix the selection behavior after loading symbols

vickyliu-go4it:
The current sampling tab does not correctly restore the selection after right-clicking loading symbols.

This is because both `SamplingReportDataview::functions_` and `SamplingReportDataview::indices_` are changed after loading symbols. Therefore, the mapping relationship between functions in `SamplingReportDataview::functions_`  and rows in the sampling tab is also changed. If still using `selected_indices_` to keep user's previous selections in the sampling tab, the re-selections will be made based on previous mapping relationship between functions and rows.

We fixed this problem by replacing `selected_indices_` with `selected_function_ids_` in the sampling tab. Also, as the
`selected_indices_` is not used anywhere else, we removed it from the DataView.

Bug: http://b/181549382
Test: In the sampling tab, select one or more functions and right click "Load Symbol". Check whether the selections in the sampling tab are correctly restored after loading symbols.

### DIFF
--- a/src/OrbitGl/SamplingReportDataView.cpp
+++ b/src/OrbitGl/SamplingReportDataView.cpp
@@ -278,27 +278,43 @@ void SamplingReportDataView::OnContextMenu(const std::string& action, int menu_i
   }
 }
 
-void SamplingReportDataView::UpdateSelectedAddressesAndTid(const std::vector<int>& indices) {
+void SamplingReportDataView::UpdateSelectedIndicesAndFunctionIds(
+    const std::vector<int>& selected_indices) {
+  selected_indices_.clear();
+  selected_function_ids_.clear();
+  for (int row : selected_indices) {
+    selected_indices_.insert(indices_[row]);
+    selected_function_ids_.insert(GetSampledFunction(row).absolute_address);
+  }
+}
+
+void SamplingReportDataView::RestoreSelectedIndicesAfterFunctionsChanged() {
+  selected_indices_.clear();
+  for (size_t row = 0; row < functions_.size(); ++row) {
+    if (selected_function_ids_.contains(functions_[row].absolute_address)) {
+      selected_indices_.insert(static_cast<int>(row));
+    }
+  }
+}
+
+void SamplingReportDataView::UpdateVisibleSelectedAddressesAndTid(
+    const std::vector<int>& visible_selected_indices) {
   absl::flat_hash_set<uint64_t> addresses;
-  for (int index : indices) {
+  for (int index : visible_selected_indices) {
     addresses.insert(GetSampledFunction(index).absolute_address);
   }
   sampling_report_->OnSelectAddresses(addresses, tid_);
 }
 
 void SamplingReportDataView::OnSelect(const std::vector<int>& indices) {
-  UpdateSelectedAddressesAndTid(indices);
-
-  selected_indices_.clear();
-  for (int row : indices) {
-    selected_indices_.insert(indices_[row]);
-  }
+  UpdateSelectedIndicesAndFunctionIds(indices);
+  UpdateVisibleSelectedAddressesAndTid(indices);
 }
 
 void SamplingReportDataView::OnRefresh(const std::vector<int>& visible_selected_indices,
                                        const RefreshMode& mode) {
   if (mode != RefreshMode::kOnFilter && mode != RefreshMode::kOnSort) return;
-  UpdateSelectedAddressesAndTid(visible_selected_indices);
+  UpdateVisibleSelectedAddressesAndTid(visible_selected_indices);
 }
 
 void SamplingReportDataView::LinkDataView(DataView* data_view) {
@@ -310,6 +326,7 @@ void SamplingReportDataView::LinkDataView(DataView* data_view) {
 
 void SamplingReportDataView::SetSampledFunctions(const std::vector<SampledFunction>& functions) {
   functions_ = functions;
+  RestoreSelectedIndicesAfterFunctionsChanged();
 
   size_t num_functions = functions_.size();
   indices_.resize(num_functions);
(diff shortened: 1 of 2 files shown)


reichlfl (commented):
I'm not sure about this... it looks to me like this is removing functionality from the base class (DataView) that made sense there, and I have a hard time following the implementation in the SamplingReportDataView.

Would it help to allow classes inheriting from DataView to customize the concept of "index" for the selection? Afaik this is quite similar to how Qt does it (and we're back at implementing functionality we should have out of the box...) - something like `MapRowToIndex`, and the default implementation just maps this to the row index, while the SamplingDataView could map it to the index of the function in the function list.

Not sure if this is helpful, I'll need to take a longer look at this.

src/OrbitGl/DataView.h
```diff
@@ -73,7 +73,7 @@ class DataView {
   // This method returns the intersection of selected indices and visible indices. The returned
   // value contains 0 or 1 index for a DataView with single selection, and contains 0 or
   // multiple indices for a DataView with multi-selection.
-  [[nodiscard]] virtual std::vector<int> GetVisibleSelectedIndices();
+  [[nodiscard]] virtual std::vector<int> GetVisibleSelectedIndices() { return std::vector<int>{}; }
```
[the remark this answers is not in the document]
vickyliu-go4it: We only have `OnSelect` implemented for the `LiveFunctionsDataView` and `SamplingReportDataView`. And  `OnSelect` is the place where we used to update user's selection (i.e., `selected_indices_`, `selected_function_id_` or `selected_function_ids_`). For other DataViews, the `selected_indices_` is never updated and the `GetVisibleSelectedIndices()` method needs the user's selection information in `selected_indices_` (or `selected_function_id_`, `selected_function_ids_`).

vickyliu-go4it:
> I'm not sure about this... it looks to me like this is removing functionality from the base class (DataView) that made sense there, and I have a hard time following the implementation in the SamplingReportDataView.
> 
> Would it help to allow classes inheriting from DataView to customize the concept of "index" for the selection? Afaik this is quite similar to how Qt does it (and we're back at implementing functionality we should have out of the box...) - something like `MapRowToIndex`, and the default implementation just maps this to the row index, while the SamplingDataView could map it to the index of the function in the function list.
> 
> Not sure if this is helpful, I'll need to take a longer look at this.

I'm also thinking about what is a right way to store user's selection information in DataViews.

In the past, when we have `ProcessDataView`, we actually use the `selected_process_id_` to update the `selected_index_`.

I'm considering that, whether we should keep the previous `selected_indices_` and `GetVisibleSelectedIndices` for the base class DataView and meanwhile allow the child classes (e.g., `LiveFunctionsDataView` and `SamplingReportDataView`) to use additional information (e.g., function id or ids) to update `selected_indices_`. WDYT?

src/OrbitGl/DataView.h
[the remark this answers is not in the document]
reichlfl: I see. Ok, this makes more sense then. But the more we mess with the DataViews, the more I think we should really try to move them to QItemModel tbh.

vickyliu-go4it:
As discussed offline: I kept the `DataView::selected_indices_` and `DataView::GetVisibleSelectedIndices` and then added `SamplingReportDataView::selected_function_ids_` to record the user's selected functions in the sampling tab. If the `SamplingReportDataView::functions_` changes, the `selected_indices_` can still be updated according to the selected function ids.

This is a quick bug ugly fix for the bug. We still need to consider how to replace `DataView` with a good `QItemModel` design.

src/OrbitGl/SamplingReportDataView.h
```diff
@@ -36,7 +36,10 @@ class SamplingReportDataView : public DataView {
   void OnRefresh(const std::vector<int>& visible_selected_indices,
                  const RefreshMode& mode) override;
 
-  void UpdateSelectedAddressesAndTid(const std::vector<int>& indices);
+  void UpdateSelectedIndicesAndFunctionIds(const std::vector<int>& selected_indices);
+  void UpdateSelectedIndicesOnFunctionsChanged();
+  // The callstack view will be updated according to the visible selected addresses and thread id.
+  void UpdateVisibleSelectedAddressesAndTid(const std::vector<int>& visible_selected_indices);
```
antonrohr, line 42: I think these can be private

src/OrbitGl/SamplingReportDataView.cpp
```diff
@@ -278,27 +278,43 @@ void SamplingReportDataView::OnContextMenu(const std::string& action, int menu_i
...
+    const std::vector<int>& selected_indices) {
+  selected_indices_.clear();
+  selected_function_ids_.clear();
+  for (int row : selected_indices) {
+    selected_indices_.insert(indices_[row]);
+    selected_function_ids_.insert(GetSampledFunction(row).absolute_address);
+  }
+}
+
+void SamplingReportDataView::UpdateSelectedIndicesOnFunctionsChanged() {
```
beckerhe, line 291: Nit: All the new method start with "Update" which is not wrong. But I would prefer for this function a name that makes it clear that this function is different. Maybe something like: `RestoreSelectedIndicesAfterFunctionsChanged`

antonrohr (approved):
> This is a quick bug ugly fix for the bug

I am okay with the quick fix as a tradeoff.

beckerhe (approved):
Yes, it's kind of ugly, but that's not your fault. I think it's the best we can do for now. I agree with Florian; having better abstractions for Indexes, Ids, etc. will make the code a lot easier, but that's out of scope for a quick fix like that.

LGTM! I added one minor comment: Feel free to ignore.

src/OrbitGl/SamplingReportDataView.h
vickyliu-go4it: Ah, right! Thanks for pointing out this. Changed accordingly.

src/OrbitGl/SamplingReportDataView.cpp
vickyliu-go4it: Thanks for suggesting the name! Changed to use `RestoreSelectedIndicesAfterFunctionsChanged`.

PR #1981, opened 2021-03-02
Test lib for the injection mechanism.

danielfenner:
In order to load shared object into attached process we need a simple
.so first to test / develop that. Thats why the produced binary is
checked in to the testdata folder.

bug: http://b/181304023
test: NA


src/UserSpaceInstrumentation/CMakeLists.txt
```diff
@@ -33,6 +33,22 @@ target_link_libraries(UserSpaceInstrumentation PUBLIC
         OrbitBase
         CONAN_PKG::abseil)
 
+# Test Lib is merely used in UserSpaceInstrumentationTests below. A binary
+# libTestLib.so created from this target is checked in to testdata folder and
+# that binary is used to test the injection mechanism.
+add_library(TestLib SHARED)
```
dpallotti, line 39: The name of the library should be expanded, for example to `UserSpaceInstrumentationTestLib`. If I'm not mistaken this still ends up in the `lib` folder and `libTestLib.so` is too generic.

src/UserSpaceInstrumentation/TestLib.cpp
```diff
@@ -0,0 +1,52 @@
...
+}  // namespace
+
+void InitTestLib() {
+  auto p = GetTmpFilePath();
+  std::cout << "Init Lib. Tmp file is: " << p << std::endl;
+}
+
+void UseTestLib(const std::string& s) {
+  auto p = GetTmpFilePath();
+  std::ofstream ofs;
```
dpallotti, line 30: I get that this is just for testing, but still I thought we wanted to get rid of `std` streams because of exceptions? I guess we should be consistent. There are utilities in `OrbitBase` to read and write an entire text file (after all they are nicer to use than stream directly).

src/UserSpaceInstrumentation/TestLib.h
```diff
@@ -0,0 +1,28 @@
...
+
+#ifndef USER_SPACE_INSTRUMENTATION_TEST_LIB_H_
+#define USER_SPACE_INSTRUMENTATION_TEST_LIB_H_
+
+#include <string>
+
+namespace orbit_user_space_instrumentation {
+
+// Library functions for some trivial logging. This library is merely used in tests: A binary
+// produced by this code is checked in into the testdata folder. The test injects this binary into
```
dpallotti, line 13: Do you actually need to put it into the testdata folder? The library ends up in the `lib` directory anyway and you can reference it. This is what we did with `LinuxTracingIntegrationTestPuppetSharedObject` and it works fine. It also makes sure that the .so and the code are actually always in sync.

src/UserSpaceInstrumentation/TestLib.cpp
```diff
@@ -0,0 +1,52 @@
...
+}
+
+}  // namespace
+
+void InitTestLib() {
+  auto p = GetTmpFilePath();
+  std::cout << "Init Lib. Tmp file is: " << p << std::endl;
+}
+
+void UseTestLib(const std::string& s) {
```
dpallotti, line 28: Not that it matters, but I guess you can use `string_view` in this case.

src/UserSpaceInstrumentation/CMakeLists.txt
danielfenner: Yep, that's true. UserSpaceInstrumentationTestLib it is.

src/UserSpaceInstrumentation/TestLib.cpp
danielfenner: done

src/UserSpaceInstrumentation/TestLib.cpp
danielfenner: I'd like to keep this as "vanilla" as possible for so I decided against using OrbitBase. 

This will not be linked to Orbit or even any test so I think exceptions don't matter in this case.

src/UserSpaceInstrumentation/TestLib.h
danielfenner: Thanks! Indeed this looks way better.

Adjusted comments and removed the binary in testdata .

danielfenner:
Besides addressing the comments I also removed the namespace in the lib; I'd like to have this as easy to work with as possible (nice short symbol names).

src/UserSpaceInstrumentation/CMakeLists.txt
```diff
@@ -33,6 +33,22 @@ target_link_libraries(UserSpaceInstrumentation PUBLIC
         OrbitBase
         CONAN_PKG::abseil)
 
+# Test Lib is merely used in UserSpaceInstrumentationTests below. The binary
```
dpallotti, line 36: Change "Test Lib" to the name of the CMake library or to "This test lib".

src/UserSpaceInstrumentation/UserSpaceInstrumentationTestLib.h
```diff
@@ -0,0 +1,23 @@
+// Copyright (c) 2021 The Orbit Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#ifndef USER_SPACE_INSTRUMENTATION_TEST_LIB_H_
+#define USER_SPACE_INSTRUMENTATION_TEST_LIB_H_
+
+#include <string_view>
+
```
dpallotti, line 9: > I also removed the namespace in the lib; I'd like to have this as easy to work with as possible (nice short symbol names).

Maybe add a comment around here with this explanation so that the reason why the namespace is missing doesn't get lost.

src/UserSpaceInstrumentation/UserSpaceInstrumentationTestLib.h
```diff
@@ -0,0 +1,23 @@
...
+// Library functions for some trivial logging. This library is merely used in tests: The test
+// injects a binary produced by this code into its child.
+
+// Call first to initialize the library.
+void InitTestLib();
+
+// Log a string into a temporary file.
+void UseTestLib(std::string_view s);
+
+// Call to end using the library. Prints the entire log to standard out and removes the temporary
```
dpallotti, line 19: out -> output

dpallotti (approved):
> I'd like to keep this as "vanilla" as possible for so I decided against using OrbitBase.
>
> This will not be linked to Orbit or even any test so I think exceptions don't matter in this case.

Is there a specific gain you have in mind with keeping it vanilla? My personal feeling is that if we are afraid of linking OrbitBase we might be doing something wrong in OrbitBase, and that consistency (in this case, not using streams) is helpful in general, even if this is just a test. (But in any case I don't feel strongly about it.)

src/UserSpaceInstrumentation/CMakeLists.txt
danielfenner: done

src/UserSpaceInstrumentation/UserSpaceInstrumentationTestLib.h
danielfenner: done

danielfenner:
> Is there a specific gain you have in mind with keeping it vanilla? My personal feeling is that if we are afraid of linking OrbitBase we might be doing something wrong in OrbitBase, and that consistency (in this case, not using streams) is helpful in general, even if this is just a test. (But in any case I don't feel strongly about it.)

I really don't know yet. I just want to keep it as simple as possible for experimenting. Good chance I'll make it more similar to a lib we want really want to inject in the next PR.

src/UserSpaceInstrumentation/UserSpaceInstrumentationTestLib.h
danielfenner: As mentioned below: This will change quite a bit no matter what and it really doesn't matter what namespace this is in since it will not link to anything.

PR #1982, opened 2021-03-02
Metrics next

antonrohr:
This PR does multiple things, they are properly separated into commits. 

1. Make `MetricsUploader.SendLogEvent` return `void` instead of `bool`. The return type was not used at all, only in tests. And there is also no need to return anything when an event cannot be sent. An error message should be printed, but the user should not be bothered. That is what happens and the test checks if that error is printed.
2. Refactor MetricsUploader to use more const. Also touches OrbitApp and OrbitMainWindow. 
3. Presets delete metric (http://b/181648088)
4. Iterator Add & Remove metric (http://b/181649379)

### DIFF
--- a/src/MetricsUploader/proto/orbit_log_event.proto
+++ b/src/MetricsUploader/proto/orbit_log_event.proto
@@ -20,6 +20,7 @@ message OrbitLogEvent {
     ORBIT_PRESET_LOAD = 6;
     ORBIT_PRESET_DELETE = 7;
     ORBIT_ITERATOR_ADD = 8;
+    ORBIT_ITERATOR_REMOVE = 9;
   }
   enum StatusCode {
     UNKNOWN_STATUS = 0;
--- a/src/MetricsUploader/proto/orbit_log_event.proto
+++ b/src/MetricsUploader/proto/orbit_log_event.proto
@@ -19,6 +19,7 @@ message OrbitLogEvent {
     ORBIT_PRESET_SAVE = 5;
     ORBIT_PRESET_LOAD = 6;
     ORBIT_PRESET_DELETE = 7;
+    ORBIT_ITERATOR_ADD = 8;
   }
   enum StatusCode {
     UNKNOWN_STATUS = 0;
--- a/src/MetricsUploader/proto/orbit_log_event.proto
+++ b/src/MetricsUploader/proto/orbit_log_event.proto
@@ -18,6 +18,7 @@ message OrbitLogEvent {
     ORBIT_CAPTURE_LOAD = 4;
     ORBIT_PRESET_SAVE = 5;
     ORBIT_PRESET_LOAD = 6;
+    ORBIT_PRESET_DELETE = 7;
   }
   enum StatusCode {
     UNKNOWN_STATUS = 0;
--- a/src/OrbitGl/PresetsDataView.h
+++ b/src/OrbitGl/PresetsDataView.h
@@ -13,13 +13,15 @@
 #include <vector>
 
 #include "DataView.h"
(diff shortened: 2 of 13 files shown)


src/MetricsUploader/MetricsUploaderWindowsTest.cpp
```diff
@@ -10,6 +10,12 @@
 #include "OrbitBase/Result.h"
 #include "absl/container/flat_hash_set.h"
 #include "absl/strings/ascii.h"
+#include "gmock/gmock-matchers.h"
+#include "gtest/internal/gtest-port.h"
```
vickyliu-go4it, line 14: Nit: May be use <> style includes for everything external to Orbit.

src/MetricsUploader/proto/orbit_log_event.proto
```diff
@@ -20,6 +20,7 @@ message OrbitLogEvent {
     ORBIT_PRESET_LOAD = 6;
     ORBIT_PRESET_DELETE = 7;
     ORBIT_ITERATOR_ADD = 8;
+    ORBIT_ITERATOR_REMOVE = 9;
```
vickyliu-go4it, line 23: Nit: As more metrics will be added in the future, may be we can put them in the alphabetically ordered without changing the numbering.

vickyliu-go4it (approved):
Very nice and well organized PR. Thanks for that! I only have some nit comments. LGTM!

src/OrbitGl/PresetsDataView.h
```diff
@@ -68,6 +69,9 @@ class PresetsDataView : public DataView {
 
   static const std::string kMenuActionLoad;
   static const std::string kMenuActionDelete;
+
+ private:
+  const orbit_metrics_uploader::MetricsUploader* metrics_uploader_;
```
IrinaShkviro, line 74: Looks like all DataViews need metrics_uploader_ anyway. Should we add it to the base class then?

IrinaShkviro (approved):
LGTM, but I personally prefer the separation that you've made in the PR description than in commits.

src/MetricsUploader/MetricsUploaderWindowsTest.cpp
antonrohr: obsolete

src/MetricsUploader/proto/orbit_log_event.proto
antonrohr: I generally like the idea, I am going to do it the next time I add something and not in this PR. This is because there is a proto in g3, that is basically a copy of the one here. I will then change both protos, so they are in sync.

antonrohr:
Hey y'all, 

After talking offline with beckerhe, I decided to actually discard some of the changes I made here. If you are interested I am happy to talk about the reasons. Since it threw stuff away, I decided to rebase and force push.

Please take a look again.

beckerhe (approved):
This is awesome. Thanks a lot for the changes. LGTM!

src/OrbitGl/PresetsDataView.h
antonrohr: For now not all DataViews need it, so I would rather not add it to the base class. But I also don't feel strongly about it and can add it, no problem. wdyt?

vickyliu-go4it (approved):
LGTM!  Thanks!

src/OrbitGl/PresetsDataView.h
IrinaShkviro: Could be done when all the DataViews have it, so not in this PR

IrinaShkviro (approved):
Thanks Anton!

PR #1983, opened 2021-03-02
Don't crash on reused command buffers.

florian-kuebler:
Command buffers can be reused without calling vkResetCommandBuffer
(followed by vkBegin/EndCommandBuffer). In this case, the timer
slots that are baked into the command buffer, need to be invalidated,
as we could have "Submit"->"Wait"->"Submit"->"Present" (note that
we are only querying for timeslots on present, and thus can not
uniquely identify the slots anymore).

This fixes this by introducing a two staged reset of slots.
We mark slots that we will not read anymore. Further we mark
slots that were part of a vkResetCommandBuffer.
The actual slot is only reset once marked with both.

Thus for the case of a re-use command buffer, we will never reset
the slot completely (at least unless vkResetCommandBuffer is called).

To not query resubmitted slots again, we keep the command buffer state
until reset/new begin, and added an optional timestamp for the submission.
This allows us to only persist (and therefore read) a slot if
it originates from this submission.

Test: Run Tests, Capture on Trata, Infiltrator and hello_ggp
Bug: http://b/181098020.

### DIFF
--- a/src/OrbitVulkanLayer/SubmissionTracker.h
+++ b/src/OrbitVulkanLayer/SubmissionTracker.h
@@ -223,8 +223,9 @@ class SubmissionTracker : public VulkanLayerProducer::CaptureStatusListener {
         ERROR("Calling vkBeginCommandBuffer on a command buffer that is not in the initial state.");
         // We end up in this case, if we have used the command buffer before and want to write new
         // commands to it without resetting the command buffer. This is prohibited by the
-        // specification. However, we've seen this in the wild. Let's assume there was a reset call,
-        // which for use mean, we need to reset the slots associated with this command buffer.
+        // specification. However, we've seen this in the wild. The "vkBeginCommandBuffer" will,
+        // however, invalidate the previous commands here. So we can reset potentially blocked slots
+        // associated with this command buffer.
         ResetCommandBufferUnsafe(command_buffer);
       }
       command_buffer_to_state_[command_buffer] = {};
@@ -334,6 +335,9 @@ class SubmissionTracker : public VulkanLayerProducer::CaptureStatusListener {
     }
 
     VkDevice device = VK_NULL_HANDLE;
+    // Collect slots of command buffer "begins" that are baked into the command buffer but for which
+    // the "end" is missing. We will never read those, but need to wait until the command buffer
+    // gets reset.
     std::vector<uint32_t> query_slots_not_needed_to_read;
 
     QueueSubmission queue_submission;
@@ -351,7 +355,7 @@ class SubmissionTracker : public VulkanLayerProducer::CaptureStatusListener {
         VkCommandBuffer command_buffer = submit_info.pCommandBuffers[command_buffer_index];
         CHECK(command_buffer_to_state_.contains(command_buffer));
         CommandBufferState& state = command_buffer_to_state_.at(command_buffer);
-        bool already_submitted = state.pre_submission_timestamp.has_value();
+        bool has_been_submitted_before = state.pre_submission_cpu_timestamp.has_value();
 
         // Mark that this command buffer in the current state was already submitted. If the command
         // buffer is being submitted again, without a "vkResetCommandBuffer" call, the same slot
@@ -361,7 +365,7 @@ class SubmissionTracker : public VulkanLayerProducer::CaptureStatusListener {
         // Note that we are using an std::optional with the submission time to protect us against an
         // "OnCaptureFinished" call, right between pre and post submission, which would try to reset
         // the slots.
-        state.pre_submission_timestamp =
+        state.pre_submission_cpu_timestamp =
             queue_submission.meta_information.pre_submission_cpu_timestamp;
 
         if (device == VK_NULL_HANDLE) {
@@ -384,7 +388,7 @@ class SubmissionTracker : public VulkanLayerProducer::CaptureStatusListener {
         // indices are not unique anymore and can't be used by us. Note, that the slots will be
         // marked as "done with reading" when the first submission of this command buffer is done
         // and will eventually be reset on a "vkResetCommandBuffer" call.
-        if (!already_submitted) {
+        if (!has_been_submitted_before) {
           SubmittedCommandBuffer submitted_command_buffer{
               .command_buffer_begin_slot_index = state.command_buffer_begin_slot_index,
               .command_buffer_end_slot_index = state.command_buffer_end_slot_index.value()};
@@ -451,9 +455,9 @@ class SubmissionTracker : public VulkanLayerProducer::CaptureStatusListener {
         for (const Marker& marker : state.markers) {
           std::optional<SubmittedMarker> submitted_marker = std::nullopt;
           CHECK(!queue_submission_optional.has_value() ||
-                state.pre_submission_timestamp.has_value());
+                state.pre_submission_cpu_timestamp.has_value());
           if (marker.slot_index.has_value() && queue_submission_optional.has_value() &&
-              (state.pre_submission_timestamp.value() ==
+              (state.pre_submission_cpu_timestamp.value() ==
                queue_submission_optional->meta_information.pre_submission_cpu_timestamp)) {
             submitted_marker = {.meta_information = queue_submission_optional->meta_information,
                                 .slot_index = marker.slot_index.value()};
@@ -633,7 +637,7 @@ class SubmissionTracker : public VulkanLayerProducer::CaptureStatusListener {
     VkDevice device = VK_NULL_HANDLE;
 
     for (auto& [command_buffer, command_buffer_state] : command_buffer_to_state_) {
-      if (command_buffer_state.pre_submission_timestamp.has_value()) continue;
+      if (command_buffer_state.pre_submission_cpu_timestamp.has_value()) continue;
       if (device == VK_NULL_HANDLE) {
         CHECK(command_buffer_to_device_.contains(command_buffer));
         device = command_buffer_to_device_.at(command_buffer);
@@ -700,7 +704,7 @@ class SubmissionTracker : public VulkanLayerProducer::CaptureStatusListener {
   struct CommandBufferState {
     std::optional<uint32_t> command_buffer_begin_slot_index;
     std::optional<uint32_t> command_buffer_end_slot_index;
-    std::optional<uint64_t> pre_submission_timestamp;
+    std::optional<uint64_t> pre_submission_cpu_timestamp;
     std::vector<Marker> markers;
     uint32_t local_marker_stack_size;
   };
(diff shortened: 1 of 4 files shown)


src/OrbitVulkanLayer/TimerQueryPool.h
```diff
@@ -104,62 +111,119 @@ class TimerQueryPool {
...
-      if (rollback_only) {
-        continue;
-      }
-      VkQueryPool query_pool = device_to_query_pool_.at(device);
-      dispatch_table_->ResetQueryPoolEXT(device)(device, query_pool, slot_index, 1);
     }
   }
 
+ private:
+  enum class SlotState { kReadyForQueryIssue = 0, kQueryPendingOnGpu };
```
ronaldfw, line 225: Couldn't we expand the states here into something like 

{ kReadyForQueryIssue = 0, kQueryPendingOnGpu = 1, kDoneReading = 2 }

and then maintain the current state in the (already existing) device_to_query_slots_ vectors? 

You always have the indices of the slots, so looking up the current state and modifying it is easy and fast (definitely a lot faster than looking up indices in a set).

src/OrbitVulkanLayer/SubmissionTracker.h
```diff
@@ -327,7 +334,7 @@ class SubmissionTracker : public VulkanLayerProducer::CaptureStatusListener {
     }
 
     VkDevice device = VK_NULL_HANDLE;
-    std::vector<uint32_t> query_slots_to_reset;
+    std::vector<uint32_t> query_slots_not_needed_to_read;
```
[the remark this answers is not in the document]
florian-kuebler: Those are the slots, that are part baked into the command buffer for the "begin" but we ran out of slots on the "end". So we will never read the result.
I added a comment.

src/OrbitVulkanLayer/SubmissionTracker.h
```diff
@@ -219,7 +219,14 @@ class SubmissionTracker : public VulkanLayerProducer::CaptureStatusListener {
     // submission. We will not write timestamps in this case and thus don't store any information
     // other than the debug markers then.
     {
-      CHECK(!command_buffer_to_state_.contains(command_buffer));
+      if (command_buffer_to_state_.contains(command_buffer)) {
+        ERROR("Calling vkBeginCommandBuffer on a command buffer that is not in the initial state.");
+        // We end up in this case, if we have used the command buffer before and want to write new
+        // commands to it without resetting the command buffer. This is prohibited by the
+        // specification. However, we've seen this in the wild. Let's assume there was a reset call,
```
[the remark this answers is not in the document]
florian-kuebler: It is formulated a bit badly from myside. But, in general yes this is safe. ResetCommandBuffer, will free potentially blocked slots, and afterwards we are clearing the state here. 
This method is called on vkBeginCommandBuffer, after that the command buffer will definitely forget the old commands and record new ones.

src/OrbitVulkanLayer/SubmissionTracker.h
```diff
@@ -219,7 +219,14 @@ class SubmissionTracker : public VulkanLayerProducer::CaptureStatusListener {
...
+        // which for use mean, we need to reset the slots associated with this command buffer.
```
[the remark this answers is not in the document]
florian-kuebler: Obsolete

src/OrbitVulkanLayer/SubmissionTracker.h
```diff
@@ -344,6 +351,18 @@ class SubmissionTracker : public VulkanLayerProducer::CaptureStatusListener {
         VkCommandBuffer command_buffer = submit_info.pCommandBuffers[command_buffer_index];
         CHECK(command_buffer_to_state_.contains(command_buffer));
         CommandBufferState& state = command_buffer_to_state_.at(command_buffer);
+        bool already_submitted = state.pre_submission_timestamp.has_value();
```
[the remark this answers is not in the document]
florian-kuebler: Sounds good.

src/OrbitVulkanLayer/SubmissionTracker.h
```diff
@@ -671,6 +700,7 @@ class SubmissionTracker : public VulkanLayerProducer::CaptureStatusListener {
   struct CommandBufferState {
     std::optional<uint32_t> command_buffer_begin_slot_index;
     std::optional<uint32_t> command_buffer_end_slot_index;
+    std::optional<uint64_t> pre_submission_timestamp;
```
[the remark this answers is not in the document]
florian-kuebler: Had no specific reason. Renamed it.

src/OrbitVulkanLayer/TimerQueryPool.h
florian-kuebler: We would need 4 states then, right?
{ kReadyForQueryIssue = 0, kQueryPendingOnGpu = 1, kDoneReading = 2, kResetRequested = 3 }

In general sounds not too bad. I will address this tomorrow.

ronaldfw: Yes, might be that we need four states. The key idea here would be to "flatten" the state space into one dimension. Right now it's spread out over this enum and two sets, with multiple possible states that are illegal (like having state kReadyForQueryIssue and being in either of the sets, slots being in both sets, etc.). With just the enum the code should be a lot easier to understand and it will be easier to assert correctness.

The one redundancy in terms of state is the "free slots" list, but that exists for performance, and I don't see how we can get rid of that while also being fast.

florian-kuebler: Makes sense!

florian-kuebler: Done.

florian-kuebler:
Thanks for the review. I changed the logic, to only make use of the state and added a state machine diagram + description to the class description.

src/OrbitVulkanLayer/TimerQueryPool.h
```diff
@@ -5,25 +5,54 @@
...
- */
+// This class wraps Vulkan's VkQueryPool explicitly for timestamp queries, and provides utility
+// methods to (1) initialize a pool, (2) retrieve an available slot index and (3) reset slot
+// indices.
+// In order to do so, it stores the internal `SlotState` for each index.
+//
+// Slots can be requested using `NextReadyQuerySlot, which will block them until being reset again.
+// If the command buffer storing the slot index gets reset before it was even submitted,
+// `RollbackPendingQuerySlots` can be called to mark the slot as being free (ready) again without
+// the need of actually resetting the timer value underlying to the slot.
```
ronaldfw, line 29:
```suggestion
// the need of actually resetting the timer value underlying the slot.
```

src/OrbitVulkanLayer/TimerQueryPool.h
```diff
@@ -5,25 +5,54 @@
...
+// Once submitted, the slot is baked into the command buffer (until the command buffer is being
+// reset again). In order to free that slot, two things need to happen:
+// 1. The client (this Vulkan layer) needs to communicate that it will not do any attempts to read
+//    that slot anymore (done by calling `MarkSlotDoneReading`).
+// 2. The command buffer needs to reset, to ensure that the slot index is not baked into it anymore
+//    (done by calling `MarkSlotForReset`).
+//
+// The following state machine represents the allowed calls.
```
ronaldfw, line 37: Nice diagram!

src/OrbitVulkanLayer/TimerQueryPool.h
```diff
@@ -104,62 +133,112 @@ class TimerQueryPool {
...
+    std::vector<uint32_t>& free_slots = device_to_free_slots_.at(device);
+    for (uint32_t slot_index : slot_indices) {
+      CHECK(slot_index < num_timer_query_slots_);
+      const SlotState& current_state = slot_states.at(slot_index);
+      CHECK(current_state == SlotState::kQueryPendingOnGpu ||
+            current_state == SlotState::kResetRequested);
+      if (current_state == SlotState::kQueryPendingOnGpu) {
+        slot_states.at(slot_index) = SlotState::kDoneReading;
+        continue;
+      }
```
ronaldfw, line 164: Nit: You could just have: 

CHECK(current_state == SlotState::kResetRequested); 

after the "if" clause. I find that a bit easier to read.

src/OrbitVulkanLayer/TimerQueryPool.h
```diff
@@ -104,62 +133,112 @@ class TimerQueryPool {
...
+      return;
+    }
+    absl::WriterMutexLock lock(&mutex_);
+    CHECK(device_to_query_slots_.contains(device));
+    std::vector<SlotState>& slot_states = device_to_query_slots_.at(device);
+    CHECK(device_to_free_slots_.contains(device));
+    std::vector<uint32_t>& free_slots = device_to_free_slots_.at(device);
+    for (uint32_t slot_index : slot_indices) {
+      CHECK(slot_index < num_timer_query_slots_);
+      const SlotState& current_state = slot_states.at(slot_index);
```
ronaldfw, line 158: Use []?

src/OrbitVulkanLayer/TimerQueryPool.h
```diff
@@ -104,62 +133,112 @@ class TimerQueryPool {
...
```
ronaldfw, line 194: Use []?

src/OrbitVulkanLayer/TimerQueryPool.h
```diff
@@ -104,62 +133,112 @@ class TimerQueryPool {
...
+      CHECK(current_state == SlotState::kQueryPendingOnGpu ||
+            current_state == SlotState::kDoneReading);
+      if (current_state == SlotState::kQueryPendingOnGpu) {
+        slot_states.at(slot_index) = SlotState::kResetRequested;
+        continue;
+      }
```
ronaldfw, line 200: Same as above, just a single CHECK(current_state == SlotState::kDoneReading) here?

ronaldfw (approved):
LGTM, just some minor comments left

src/OrbitVulkanLayer/TimerQueryPool.h
florian-kuebler: Done.

src/OrbitVulkanLayer/TimerQueryPool.h
florian-kuebler: Agree. Thanks

src/OrbitVulkanLayer/TimerQueryPool.h
florian-kuebler: Done.

src/OrbitVulkanLayer/TimerQueryPool.h
florian-kuebler: Done.

florian-kuebler (commented):
Thanks for the review.

src/OrbitVulkanLayer/SubmissionTracker.h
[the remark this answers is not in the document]
ronaldfw: I thought about it but didn't like any of the ideas I came up with. And technically, the naming is still true: If you don't need to read them and even if you never read them, you are done reading. I'd leave it as-is.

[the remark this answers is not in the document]
florian-kuebler: Sounds good.

PR #1984, opened 2021-03-02
Fix crash when running out of slots and discarding the submission.

florian-kuebler:
If we run out of timer slots, it may happen that we have not single
command buffer timer slot, and thus discard the submission (even
if that contained a valid "begin marker").

If the submission containing the "end marker" is not discarded,
we ran into a CHECK on the UI. This will discard the debug marker
on the UI.

Test: Capture Infiltrator for a long time since beginning.
Bug: http://177652631

### DIFF
--- a/src/OrbitCaptureClient/GpuQueueSubmissionProcessor.cpp
+++ b/src/OrbitCaptureClient/GpuQueueSubmissionProcessor.cpp
@@ -347,7 +347,8 @@ std::vector<TimerInfo> GpuQueueSubmissionProcessor::ProcessGpuDebugMarkers(
             begin_marker_thread_id, begin_marker_post_submission_cpu_timestamp);
         // Note that we receive submissions of a single queue in order (by CPU submission time).
         // However, if we are out of timer slot indices, we might discard submissions (if it
-        // contains no command buffer timers).
+        // contains no command buffer timers). If we don't have a matching submission for the
+        // "begin" marker, we have to discard the entire marker.
         if (matching_begin_submission == nullptr) {
           ERROR("Discarding debug marker timer.");
           continue;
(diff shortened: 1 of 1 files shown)


src/OrbitCaptureClient/GpuQueueSubmissionProcessor.cpp
```diff
@@ -345,10 +345,13 @@ std::vector<TimerInfo> GpuQueueSubmissionProcessor::ProcessGpuDebugMarkers(
       } else {
         const GpuQueueSubmission* matching_begin_submission = FindMatchingGpuQueueSubmission(
             begin_marker_thread_id, begin_marker_post_submission_cpu_timestamp);
-        // Note that we receive submissions of a single queue in order (by CPU submission time). So
-        // if there is no matching "begin submission", the "begin" was submitted before the "end"
-        // and we lost the record of the "begin submission" (which should not happen).
-        CHECK(matching_begin_submission != nullptr);
+        // Note that we receive submissions of a single queue in order (by CPU submission time).
+        // However, if we are out of timer slot indices, we might discard submissions (if it
```
ronaldfw, line 349: Can we / should we discard these earlier?

ronaldfw (approved):
LGTM, modulo one question

src/OrbitCaptureClient/GpuQueueSubmissionProcessor.cpp
florian-kuebler: I don't know how. So I guess it is fine.

ronaldfw (approved):
LGTM, maybe add a brief additional comment (see suggestion)

src/OrbitCaptureClient/GpuQueueSubmissionProcessor.cpp
florian-kuebler: Exactly, this is the case.

src/OrbitCaptureClient/GpuQueueSubmissionProcessor.cpp
```diff
@@ -345,10 +345,13 @@ std::vector<TimerInfo> GpuQueueSubmissionProcessor::ProcessGpuDebugMarkers(
...
+        // contains no command buffer timers).
```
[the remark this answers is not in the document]
florian-kuebler: Slightly rephrased it, but in general makes sense.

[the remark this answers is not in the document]
ronaldfw: Looks great, thanks!

PR #1987, opened 2021-03-03
Increase the TimerQueryPool size.

florian-kuebler:
Based on experiments with Infiltrator, the previous size of
the timer query pool was a bit to small, as we were loosing
events at the beginning. This doubles to size.

Test: Run on Infiltrator from the beginning.


ronaldfw (approved):
LGTM

PR #1992, opened 2021-03-03
Avoid passing through 0 color values from the Vulkan layer

ronaldfw:
We don't want to render 0 color values in our GPU tracks, as they are
invisible (simply because alpha is 0). However, we have only checked
for colors being not 0.f in all components. This made it possible that
really small values, close to 0.f, were still passed through by the
layer, and rendered as quanitize (0, 0, 0, 0) in 8 bit precision.

We now check the color values against 0 after quantizing to 8 bits. This
fixes the issue of "invisible" debug markers.

The fix is debatable, because it might be user intention to actually
use those values, but the same is true for exactly (0.f, 0.f, 0.f 0.f).

Tested: Manually ran layer with example application.
Bug: http://b/181747686


florian-kuebler (approved):
Thanks for addressing this. In general i guess it is a bit problematic (as this is still an explicitly defined color). However, as this ends up being basically unusable in Orbit, I'd say lets go with this.

PR #1993, opened 2021-03-03
Fix issue in source code view when there is no samples

karupayun:
Show 0.00 % if there isn't any sample.

### DIFF
--- a/src/CodeViewer/Viewer.cpp
+++ b/src/CodeViewer/Viewer.cpp
@@ -195,9 +195,10 @@ void Viewer::DrawLineNumbers(QPaintEvent* event) {
 }
 
 // For example, from a value = 0.5, we want to print "50.00 %"
-static QString DoubleToPercentageString(double v) {
-  v = std::clamp(v, 0., 1.);
-  return QString{"%1 %"}.arg(v * 100., 0, 'f', 2);
+static QString FractionToPercentageString(int a, int b) {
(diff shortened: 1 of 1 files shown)


beckerhe:
I think we misunderstood each other. If there is no data available (the returned optional  has no value, then we want to show nothing.) we only want to show a number if we have data available.

src/CodeViewer/Viewer.cpp
```diff
@@ -239,9 +239,13 @@ void Viewer::DrawSampleCounters(QPaintEvent* event) {
                                fontMetrics().height()};
       painter.setPen(kLineNumberForegroundColor);
 
+      double function_percentage =
+          (code_report_->GetNumSamplesInFunction()
+               ? static_cast<double>(num_samples_in_line) / code_report_->GetNumSamplesInFunction()
+               : 0.);
```
beckerhe, line 245: Could you make a small static function out of this?

beckerhe (approved):
Otherwise LGTM

src/CodeViewer/Viewer.cpp
karupayun: I modified DoubleToPercentageString for FractionToPercentageString. Seems to be easier to read.

beckerhe: Even better.

PR #1994, opened 2021-03-03
Return 0 samples inside function

beckerhe:
`SourceCodeReport` used to return an empty optional for every line without a sample.

This commit changes the following: For every line inside of the function without samples 0 is returned. For every line outside of the function an empty optional is returned.

### DIFF
--- a/src/OrbitGl/SourceCodeReport.cpp
+++ b/src/OrbitGl/SourceCodeReport.cpp
@@ -4,6 +4,8 @@
 
 #include "SourceCodeReport.h"
 
+#include <algorithm>
+#include <limits>
 #include <optional>
 
 #include "OrbitBase/Logging.h"
@@ -39,14 +41,23 @@ SourceCodeReport::SourceCodeReport(std::string_view source_file,
       continue;
     }
 
+    min_line_number_ = std::min(min_line_number_, current_line_info.source_line());
+    max_line_number_ = std::max(max_line_number_, current_line_info.source_line());
+
     number_of_samples_per_line_[current_line_info.source_line()] += current_samples;
     total_samples_in_function_ += current_samples;
   }
 }
 
 std::optional<uint32_t> SourceCodeReport::GetNumSamplesAtLine(size_t line) const {
-  if (!number_of_samples_per_line_.contains(line)) return std::nullopt;
-  return number_of_samples_per_line_.at(line);
+  // Outside the function, no data. So we return an empty optional.
+  if (line < min_line_number_ || line > max_line_number_) return std::nullopt;
+
+  // We are inside the function and we have samples
+  if (number_of_samples_per_line_.contains(line)) return number_of_samples_per_line_.at(line);
+
+  // We are inside the function, but have no samples. So we return 0
+  return 0;
 }
 
 }  // namespace orbit_gl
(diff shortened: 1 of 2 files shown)


src/OrbitGl/SourceCodeReport.h
```diff
@@ -39,6 +39,8 @@ class SourceCodeReport : public CodeReport {
   absl::flat_hash_map<uint32_t, uint32_t> number_of_samples_per_line_;
   uint32_t total_samples_in_function_ = 0;
   uint32_t total_samples_in_capture_ = 0;
+  uint32_t min_line_number_ = 0;
```
karupayun, line 42: Please initialize this variables here instead of inside the constructor

src/OrbitGl/SourceCodeReport.cpp
```diff
@@ -29,6 +33,10 @@ SourceCodeReport::SourceCodeReport(std::string_view source_file,
     if (!maybe_current_line_info.has_value()) continue;
 
     const auto& current_line_info = maybe_current_line_info.value();
+
+    min_line_number_ = std::min(min_line_number_, current_line_info.source_line());
```
karupayun: Shouldn't these updates be before the continue??

src/OrbitGl/SourceCodeReport.cpp
```diff
@@ -39,14 +44,23 @@ SourceCodeReport::SourceCodeReport(std::string_view source_file,
...
     number_of_samples_per_line_[current_line_info.source_line()] += current_samples;
     total_samples_in_function_ += current_samples;
   }
 }
 
 std::optional<uint32_t> SourceCodeReport::GetNumSamplesAtLine(size_t line) const {
-  if (!number_of_samples_per_line_.contains(line)) return std::nullopt;
-  return number_of_samples_per_line_.at(line);
+  // We are inside the function and we have samples
+  if (number_of_samples_per_line_.contains(line)) return number_of_samples_per_line_.at(line);
```
karupayun, line 57: Nit: I don't like the order here.

I would prefer:
- get nullopt checking lines
- this line
- return 0

karupayun (approved):
LGTM Henning!

BTW, I'm also not sure if blank lines are working properly because Im not able to test it. Let me know :)

src/OrbitGl/SourceCodeReport.cpp
```diff
@@ -29,6 +33,7 @@ SourceCodeReport::SourceCodeReport(std::string_view source_file,
     if (!maybe_current_line_info.has_value()) continue;
 
     const auto& current_line_info = maybe_current_line_info.value();
+
```
karupayun, line 36: Nit: erase this line.

beckerhe: Done.

src/OrbitGl/SourceCodeReport.cpp
beckerhe: Done.

src/OrbitGl/SourceCodeReport.cpp
beckerhe: No the loop is not over line numbers! The loop is over instruction addresses. So we have to obtain the line number. I actually moved this even later.

src/OrbitGl/SourceCodeReport.h
beckerhe: changed.

PR #1997, opened 2021-03-03
Eventtrack accessibility update

reichlfl:
This PR enables basic accessibility information for the thread bars (EventTrack, TracepointTrack, ThreadStateTrack), and updates E2E tests to verify the newly available information:

- Tracks now report the existence and locations of thread bars underneath, but not individual samples
- Tracks expose the existence of timers: an additional TimerPane is reported if at lease one timer is visible in the current viewport, but no individual timers are reported
- The E2E tests for thread states, instrument function and frame track check for the respective content in the capture window

Suggest to go through it commit by commit as the first few commits lay some ground work.

### DIFF
--- a/contrib/automation_tests/test_cases/capture_window.py
+++ b/contrib/automation_tests/test_cases/capture_window.py
@@ -28,8 +28,11 @@ class CaptureWindowE2ETestCaseBase(E2ETestCase):
         self._time_graph = self.find_control('Image', name='TimeGraph', parent=suite.top_window())
         super().execute(suite=suite)
 
-    def _find_tracks(self):
-        return self._time_graph.children()
+    def _find_tracks(self, name_contains: str = None):
+        tracks = self._time_graph.children()
+        if name_contains is not None:
+            tracks = filter(lambda track: name_contains in track.texts()[0], tracks)
+        return list(tracks)
 
 
 class SelectTrack(CaptureWindowE2ETestCaseBase):
@@ -239,3 +242,42 @@ class Capture(E2ETestCase):
         self._set_capture_options(collect_thread_states)
         self._take_capture(length_in_seconds)
         self._verify_existence_of_tracks()
+
+
+class CheckThreadStates(CaptureWindowE2ETestCaseBase):
+    def _execute(self, track_name_contains: str, expect_exists: bool = True):
+        tracks = self._find_tracks(track_name_contains)
+        self.expect_true(len(tracks) > 0, 'Found tracks matching %s' % track_name_contains)
+        logging.info('Checking for thread states in %s tracks', len(tracks))
+        for track in tracks:
+            track = Track(track)
+            if expect_exists:
+                self.expect_true(track.thread_states is not None, 'Track %s has a thread state pane' % track.name)
+            else:
+                self.expect_true(track.thread_states is None, 'Track %s has a no thread state pane' % track.name)
+
+
+class CheckTimers(CaptureWindowE2ETestCaseBase):
+    def _execute(self, track_name_contains: str, expect_exists: bool = True):
+        tracks = self._find_tracks(track_name_contains)
+        self.expect_true(len(tracks) > 0, 'Found tracks matching "%s"' % track_name_contains)
+        logging.info('Checking for timers in %s tracks', len(tracks))
+        for track in tracks:
+            track = Track(track)
+            if expect_exists:
+                self.expect_true(track.timers is not None, 'Track "%s" has timers pane' % track.name)
+            else:
+                self.expect_true(track.timers is None, 'Track "%s" has no timers pane' % track.name)
+
+
+class CheckEvents(CaptureWindowE2ETestCaseBase):
+    def _execute(self, track_name_contains: str, expect_exists: bool = True):
+        tracks = self._find_tracks(track_name_contains)
+        self.expect_true(len(tracks) > 0, 'Found tracks matching "%s"' % track_name_contains)
+        logging.info('Checking for events in %s tracks', len(tracks))
+        for track in tracks:
+            track = Track(track)
+            if expect_exists:
+                self.expect_true(track.events is not None, 'Track "%s" has events pane' % track.name)
+            else:
+                self.expect_true(track.events is None, 'Track "%s" has no events pane' % track.name)
--- a/src/OrbitGl/AccessibleThreadBar.cpp
+++ b/src/OrbitGl/AccessibleThreadBar.cpp
@@ -1,0 +1,58 @@
+// Copyright (c) 2020 The Orbit Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#include "AccessibleThreadBar.h"
+
+#include "GlCanvas.h"
+#include "OrbitBase/Logging.h"
+#include "ThreadBar.h"
+#include "Track.h"
+
+namespace orbit_gl {
+
+int orbit_gl::AccessibleThreadBar::AccessibleChildCount() const { return 0; }
+
+const orbit_accessibility::AccessibleInterface* AccessibleThreadBar::AccessibleParent() const {
+  CHECK(thread_bar_ != nullptr);
+
+  if (thread_bar_->GetParent() == nullptr) return nullptr;
+  if (thread_bar_->GetParent()->GetAccessibleInterface() == nullptr) return nullptr;
+
+  // This relies on knowledge about AccessibleTrack: Accessibility is split into "virtual" controls
+  // that only exist as children in the accessibility tree. In particular, the track is split into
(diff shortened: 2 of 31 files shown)


contrib/automation_tests/test_cases/capture_window.py
```diff
@@ -239,3 +242,42 @@ def _execute(self, length_in_seconds: int = 5, collect_thread_states: bool = Fal
         self._set_capture_options(collect_thread_states)
         self._take_capture(length_in_seconds)
         self._verify_existence_of_tracks()
+
+
+class CheckThreadStates(CaptureWindowE2ETestCaseBase):
+    def _execute(self, track_name_contains: str, expect_exists: bool = True):
+        tracks = self._find_tracks(track_name_contains)
+        self.expect_true(len(tracks) > 0, 'Found tracks matching %s' % track_name_contains)
```
dpallotti, line 250: I'd wrap `%s` in quotes. Here and in the classes below.

contrib/automation_tests/test_cases/capture_window.py
```diff
@@ -239,3 +242,42 @@ def _execute(self, length_in_seconds: int = 5, collect_thread_states: bool = Fal
...
+        logging.info('Checking for threadstates in %s tracks', len(tracks))
```
dpallotti, line 251:
```suggestion
        logging.info('Checking for thread states in %s tracks', len(tracks))
```

contrib/automation_tests/test_cases/capture_window.py
```diff
@@ -239,3 +242,42 @@ def _execute(self, length_in_seconds: int = 5, collect_thread_states: bool = Fal
...
+        for track in tracks:
+            track = Track(track)
+            if expect_exists:
+                self.expect_true(track.thread_states is not None, 'Track %s has a threadstate pane' % track.name)
```
dpallotti, line 255:
```suggestion
                self.expect_true(track.thread_states is not None, 'Track %s has a thread state pane' % track.name)
```

contrib/automation_tests/test_cases/capture_window.py
```diff
@@ -239,3 +242,42 @@ def _execute(self, length_in_seconds: int = 5, collect_thread_states: bool = Fal
...
+            else:
+                self.expect_true(track.thread_states is None, 'Track %s has a no threadstate pane' % track.name)
```
dpallotti, line 257:
```suggestion
                self.expect_true(track.thread_states is None, 'Track %s has a no thread state pane' % track.name)
```

src/OrbitGl/EventTrack.cpp
```diff
@@ -32,8 +32,10 @@ using orbit_client_protos::CallstackEvent;
 namespace orbit_gl {
 
 EventTrack::EventTrack(OrbitApp* app, TimeGraph* time_graph, TimeGraphLayout* layout,
-                       const CaptureData* capture_data, ThreadID thread_id)
-    : ThreadBar(app, time_graph, layout, capture_data, thread_id), color_{0, 255, 0, 255} {}
+                       const CaptureData* capture_data, ThreadID thread_id,
+                       CaptureViewElement* parent)
+    : ThreadBar(app, time_graph, layout, capture_data, thread_id, parent, "Events"),
```
dpallotti, line 37: Seeing `"Events"` passed as `name` and used in the E2E tests reminded me that `CallstackTrack` and `"Callstacks"` would be more appropriate. Just throwing it there...

dpallotti (approved):
Minor comments. Thanks for the change!

src/OrbitGl/CaptureViewElement.h
```diff
@@ -35,6 +35,11 @@ class CaptureViewElement : public Pickable {
   void SetSize(float width, float height) { size_ = Vec2(width, height); }
   [[nodiscard]] Vec2 GetSize() const { return size_; }
 
+  [[nodiscard]] orbit_accessibility::AccessibleInterface* GetOrCreateAccessibleInterface();
+  [[nodiscard]] const orbit_accessibility::AccessibleInterface* GetAccessibleInterface() const {
+    return accessible_interface_.get();
+  }
```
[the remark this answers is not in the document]
reichlfl: Exactly, it's only for the `const` case. Accessible interfaces may need this to retrieve the parent-interface in `AccessibleParent`, which is const and can safely assume that the parent has already created the interface since the tree is always traversed from the root. And in this case, it also needs to be public.

src/OrbitGl/AccessibleThreadBar.cpp
```diff
@@ -0,0 +1,56 @@
...
+
+std::string AccessibleThreadBar::AccessibleName() const { return thread_bar_->GetName(); }
+
+orbit_accessibility::AccessibilityRect orbit_gl::AccessibleThreadBar::AccessibleLocalRect() const {
+  CHECK(thread_bar_ != nullptr);
+
+  Vec2 pos = thread_bar_->GetPos();
+
+  // TODO(b/177350599): This could be cleaned up with clearer coordinate systems
+  if (thread_bar_->GetParent() != nullptr) {
```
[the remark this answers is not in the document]
reichlfl: Yes, kind of... I think the core issue is that we're specifying absolute positions when it would make more sense to specify positions relative to the parent, then `pos` would be correct in both cases.

This can't actually happen since we're parenting all elements under `TimeGraph` currently, I just wanted to make sure this case works and is well-defined to have it unit-testable (when tracks are finally testable).

src/OrbitGl/AccessibleTimeGraph.cpp
```diff
@@ -30,7 +30,9 @@ int TimeGraphAccessibility::AccessibleChildCount() const {
 }
 
 const AccessibleInterface* TimeGraphAccessibility::AccessibleChild(int index) const {
-  return time_graph_->GetTrackManager()->GetVisibleTracks()[index]->AccessibilityInterface();
+  return time_graph_->GetTrackManager()
```
florian-kuebler, line 33: Why is this reformatted? Was that intended?

florian-kuebler (approved):
Thanks for that change, in general looks pretty good. I guess the most relevant points are pointed out by @dpallotti.

src/OrbitGl/EventTrack.cpp
reichlfl: I was going to rename those classes to not be called "Track" anymore, I think `CallstackThreadBar` would be more fitting given the base class is called `ThreadBar`. But I wanted to do this in a follow-up PR to make the review easier.

src/OrbitGl/AccessibleTimeGraph.cpp
reichlfl: Yes, the function name changed from `AccessibilityInterface` to `GetOrCreateAccessibilityInterface`

src/OrbitGl/AccessibleThreadBar.cpp
```diff
@@ -0,0 +1,56 @@
...
+    Vec2 parent_pos = thread_bar_->GetParent()->GetPos();
+    pos[0] -= parent_pos[0];
+    pos[1] = parent_pos[1] - pos[1];
```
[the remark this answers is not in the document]
reichlfl: Right, that's better in terms of readability - changed accordingly, thanks!

src/OrbitGl/AccessibleTrack.cpp
```diff
@@ -17,25 +17,35 @@
...
 using orbit_accessibility::AccessibilityState;
 using orbit_accessibility::AccessibleInterface;
 
 namespace orbit_gl {
 
-int AccessibleTrackContent::AccessibleChildCount() const { return 0; }
+int AccessibleTrackContent::AccessibleChildCount() const {
+  std::vector<CaptureViewElement*> children = track_->GetVisibleChildren();
+  int result = static_cast<int>(children.size());
+  return result;
```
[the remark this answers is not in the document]
reichlfl: This was changed in a later commit anyways... I cleaned it up nevertheless.

src/OrbitGl/TimerTrack.cpp
```diff
@@ -98,6 +98,8 @@ void TimerTrack::DrawTimer(const TextBox* prev_text_box, const TextBox* next_tex
   if (current_timer_info.start() >= *min_ignore && current_timer_info.end() <= *max_ignore) return;
   if (!TimerFilter(current_timer_info)) return;
 
+  ++visible_timer_count_;
```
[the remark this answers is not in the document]
reichlfl: True, that's definitely better. Completely agree, changed.

reichlfl:
Thanks for the comments, @dpallotti and @florian-kuebler. I addressed all suggested fixes except the renaming, will do in another PR. I rebased and reordered commits, if you want to have another look it should be sufficient to take a look at the fixup! commits. In any case, I'd merge later today since you already approved and the fixes were straightforward, unless there's any complaints.

PR #1999, opened 2021-03-04
[CHERRY-PICK] Fix off-by-one-line samples in Disassembly View

beckerhe:
Test: Manual test with Infiltrator
Bug: http://b/181819973

### DIFF
--- a/src/OrbitGl/DisassemblyReport.cpp
+++ b/src/OrbitGl/DisassemblyReport.cpp
@@ -5,6 +5,9 @@
 #include "DisassemblyReport.h"
 
 std::optional<uint32_t> DisassemblyReport::GetNumSamplesAtLine(size_t line) const {
+  // The given line number will be 1-indexed, but `Disassembler` works with 0-indexed line numbers.
+  line -= 1;
+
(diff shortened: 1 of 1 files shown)


src/OrbitGl/DisassemblyReport.cpp
```diff
@@ -5,6 +5,10 @@
 #include "DisassemblyReport.h"
 
 std::optional<uint32_t> DisassemblyReport::GetNumSamplesAtLine(size_t line) const {
+  // The given line number will be 1-indexed, but `Disassembler` works with 0-indexed line numbers.
+  // So we subtract 1 here:
```
ronaldfw, line 9: Nit: This line in the comment is pretty much exactly saying what the next code line does, so I'd remove it (the above line is great and exactly covers the "why").

ronaldfw (approved):
LGTM

src/OrbitGl/DisassemblyReport.cpp
beckerhe: Removed. Thanks!

PR #2002, opened 2021-03-04
Remove DialogCode from QMessageBox usage

beckerhe:
QMessageBox's exec function does not return a QDialogCode, which means
we can't use `QDialog::accept` and `QDialog::reject` to set the dialog's
status.


IrinaShkviro (approved):
Looks good!

PR #2004, opened 2021-03-04
Improve error message from "ggp version"

dimitry-:
Also log stderr and stdout from ggp

Test: Start orbit see that error with more context is displayed as a
      tooltip, and that stderr content is in the logs.


antonrohr (approved):
Thank you!

PR #2005, opened 2021-03-04
[CHERRY-PICK] Avoid LoadSymbols in parallel

beckerhe:
We used to check whether symbols are already loaded by checking
`module->is_loaded`, but that flag doesn't tell whether symbols are
currently being loaded. And this might lead to a race condition where
two competing `LoadSymbols` calls have been scheduled at the same time.

We avoid this by doing the same that we also do for module. We do
bookkeeping in a hash map about which module are currently being symbols
loaded for.

### DIFF
--- a/src/OrbitGl/App.cpp
+++ b/src/OrbitGl/App.cpp
@@ -1378,6 +1378,11 @@ void OrbitApp::AddSymbols(const std::filesystem::path& module_file_path,
 
 orbit_base::Future<ErrorMessageOr<void>> OrbitApp::LoadSymbols(
     const std::filesystem::path& symbols_path, const std::string& module_file_path) {
+  const auto it = symbols_currently_loading_.find(module_file_path);
+  if (it != symbols_currently_loading_.end()) {
+    return it->second;
+  }
+
   auto scoped_status = CreateScopedStatus(absl::StrFormat(
       R"(Loading symbols for "%s" from file "%s"...)", module_file_path, symbols_path.string()));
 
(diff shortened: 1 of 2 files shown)


antonrohr (approved):
I looked through quite some code paths to figure out if this PR is actually enough. It seems to me that it is enough, but we should still think about this in the future and make a more resilient design. I am still afraid that a module can somehow be loaded again.

vickyliu-go4it (approved):
Tested on my windows workstation. It works perfectly! Thanks for fixing it!

beckerhe:
> I looked through quite some code paths to figure out if this PR is actually enough. It seems to me that it is enough, but we should still think about this in the future and make a more resilient design. I am still afraid that a module can somehow be loaded again.

What do you mean be loaded again? Retrieved again?

beckerhe:
So, I assume, we decide to merge that?

antonrohr:
Yes, merge

antonrohr:
> What do you mean be loaded again? Retrieved again?

Sorry, I meant adding symbols to a module again. Aka have this `LoadSymbols()` method called again, after the future is already removed from `symbols_currently_loading_`

beckerhe:
> > What do you mean be loaded again? Retrieved again?
> 
> Sorry, I meant adding symbols to a module again. Aka have this `LoadSymbols()` method called again, after the future is already removed from `symbols_currently_loading_`

Yes, I tried to make this bullet proof. But the old still has different assumption than the new code. This will all get better when we move all these functions into a separate class.

PR #2006, opened 2021-03-04
Fix PerfEventQueue::TopEvent,PopEvent returning different events

dpallotti:
When the `PerfEvent` at the top of `priority_queue_of_events_not_ordered_by_fd_`
and the `PerfEvent` at the top of `heap_of_queues_of_events_ordered_by_fd_` have
the exact same timestamp, `TopEvent` was returning the former, while `PopEvent`
was popping and returning the latter. This was particularly problematic when
`PopEvent` is called based on the result of `TopEvent`.

Bug: http://b/181840879

Test: Follow instructions in the bug and verify it's fixed. Add unit test that
was previously failing and now succeeds.

### DIFF
--- a/src/LinuxTracing/PerfEventQueue.cpp
+++ b/src/LinuxTracing/PerfEventQueue.cpp
@@ -47,7 +47,8 @@ bool PerfEventQueue::HasEvent() const {
 
 PerfEvent* PerfEventQueue::TopEvent() {
   // As we effectively have two priority queues, get the older event between the two events at the
-  // top of the two queues.
+  // top of the two queues. In case those two events have the exact same timestamp, return the one
+  // at the top of priority_queue_of_events_not_ordered_by_fd_ (and do the same in PopEvent).
   PerfEvent* top_event = nullptr;
   if (!priority_queue_of_events_not_ordered_by_fd_.empty()) {
     top_event = priority_queue_of_events_not_ordered_by_fd_.top().get();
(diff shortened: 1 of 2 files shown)


src/LinuxTracing/PerfEventQueue.cpp
```diff
@@ -65,10 +66,12 @@ PerfEvent* PerfEventQueue::TopEvent() {
 std::unique_ptr<PerfEvent> PerfEventQueue::PopEvent() {
   if (!priority_queue_of_events_not_ordered_by_fd_.empty() &&
       (heap_of_queues_of_events_ordered_by_fd_.empty() ||
-       priority_queue_of_events_not_ordered_by_fd_.top()->GetTimestamp() <
+       priority_queue_of_events_not_ordered_by_fd_.top()->GetTimestamp() <=
```
dpallotti, line 69: This is the actual change.

src/LinuxTracing/PerfEventQueueTest.cpp
```diff
@@ -251,4 +251,18 @@ TEST(PerfEventQueue, OrderedFdsAndNotOrderedInAnyFileDescriptor) {
...
+    TopEventAndPopEventReturnTheSameWhenAnEventOrderedByFdAndAnEventNotOrderedInAnyFdHaveTheSameTimestamp) {
+  PerfEventQueue event_queue;
+  constexpr uint64_t kCommonTimestamp = 100;
+
+  event_queue.PushEvent(MakeTestEvent(11, kCommonTimestamp));
+  event_queue.PushEvent(MakeTestEvent(PerfEvent::kNotOrderedInAnyFileDescriptor, kCommonTimestamp));
+
+  PerfEvent* top_event = event_queue.TopEvent();
+  std::unique_ptr<PerfEvent> popped_event = event_queue.PopEvent();
+  EXPECT_EQ(top_event, popped_event.get());
```
dpallotti, line 265: I'm about to expand this slightly.

ronaldfw (approved):
LGTM

src/LinuxTracing/PerfEventQueueTest.cpp
dpallotti: Done.

dimitry- (approved):

PR #2007, opened 2021-03-04
[REBASE] Avoid LoadSymbols in parallel

beckerhe:
We used to check whether symbols are already loaded by checking
`module->is_loaded`, but that flag doesn't tell whether symbols are
currently being loaded. And this might lead to a race condition where
two competing `LoadSymbols` calls have been scheduled at the same time.

We avoid this by doing the same that we also do for module. We do
bookkeeping in a hash map about which module are currently being symbols
loaded for.


beckerhe:
So has already been reviewed in #2005 . So when the CI is done I will go ahead and merge it.

PR #2010, opened 2021-03-04
Fix Source Paths Mapping problem

beckerhe:
This fixes the problem that we had seen in the demo. With this change, source paths mappings in Unreal work well.

### DIFF
--- a/src/SourcePathsMapping/Mapping.cpp
+++ b/src/SourcePathsMapping/Mapping.cpp
@@ -16,38 +16,18 @@ namespace fs = std::filesystem;
 
 namespace orbit_source_paths_mapping {
 
-std::optional<fs::path> MapToFirstMatchingTarget(absl::Span<const Mapping> mappings,
-                                                 const fs::path& source_path) {
-  for (const auto& mapping : mappings) {
-    if (absl::StartsWith(source_path.string(), mapping.source_path.string())) {
-      std::string target = mapping.target_path.string();
-      target.append(
-          std::string_view{source_path.string()}.substr(mapping.source_path.string().size()));
-      return fs::path{target};
-    }
+static bool IsRegularFile(const std::filesystem::path& target_path) {
+  std::error_code error{};
+  if (fs::is_regular_file(target_path, error)) return true;
+  if (error.value() != 0) {
+    ERROR("Failed to 'stat' the file \"%s\": %s", target_path.string(), error.message());
   }
-
-  return std::nullopt;
+  return false;
 }
 
 std::optional<fs::path> MapToFirstExistingTarget(absl::Span<const Mapping> mappings,
                                                  const fs::path& source_path) {
-  for (const auto& mapping : mappings) {
-    if (absl::StartsWith(source_path.string(), mapping.source_path.string())) {
-      std::string target = mapping.target_path.string();
-      target.append(
-          std::string_view{source_path.string()}.substr(mapping.source_path.string().size()));
-      auto target_path = fs::path{target};
-
-      std::error_code error{};
-      if (fs::is_regular_file(target_path, error)) return target_path;
-      if (error.value() != 0) {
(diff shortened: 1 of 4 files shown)


beckerhe:
@antonrohr No hurries. This has a low priority regarding the release going on in parallel.

antonrohr:
I havent gotten to this yet. I did not forget this. I will do it on Monday

antonrohr (approved):
In general looks good to me!

1. Can you add a test case that tests case insensitivity. Its probably not strictly necessary, but it would give me some piece of mind knowing that this will still work even with `c:\` vs `C:\`
2. May I ask what the problem was before and why it works now?

beckerhe:
> In general looks good to me!
> 
> 1. Can you add a test case that tests case insensitivity. Its probably not strictly necessary, but it would give me some piece of mind knowing that this will still work even with `c:\` vs `C:\`
> 2. May I ask what the problem was before and why it works now?

I can add a test for case-sensitivity. Regarding the problem: It is described in the commit message, but here in short:
Previously the matching was made based on string prefix matching. That was fine as long as the path from debuginfo was just taken as-is and matched. But paths in debuginfo sometimes contain "/../" segments which should be stripped before doing the matching. This operation is called normalization and it also changes the path segment separator from `/` to `\` or vice-versa. So the mapping might not match anymore.

The solution: No string-base comparison, but segment-wise comparison.

beckerhe:
> > In general looks good to me!
> > 
> > 1. Can you add a test case that tests case insensitivity. Its probably not strictly necessary, but it would give me some piece of mind knowing that this will still work even with `c:\` vs `C:\`
> > 2. May I ask what the problem was before and why it works now?
> 
> I can add a test for case-sensitivity. Regarding the problem: It is described in the commit message, but here in short:
> Previously the matching was made based on string prefix matching. That was fine as long as the path from debuginfo was just taken as-is and matched. But paths in debuginfo sometimes contain "/../" segments which should be stripped before doing the matching. This operation is called normalization and it also changes the path segment separator from `/` to `\` or vice-versa. So the mapping might not match anymore.
> 
> The solution: No string-base comparison, but segment-wise comparison.

Okay, on a second thought regarding case-sensitivity: I don't think that's a good idea. Doing string comparisons in the first approach was done on purpose: Because if the paths don't match byte-wise, it's clear that the mapping shouldn't match. Consider the following examples:

Two devs compile unreal, the first one uses C:\ue4 as its root directory, the second C:\UE4. These paths also end up in the debug information. Seeing a path C:\ue4\Engine\Source\whatever.cpp shouldn't match a mapping for C:\UE4 because this mapping was clearly created from different debug information. WDYT?

antonrohr:
> The solution: No string-base comparison, but segment-wise comparison. 

Thank you for the explanation! I indeed did not look at the commit messages. It makes a lot of sense now. 

Regarding the case sensitivity. I am still a little torn. I think your explanation to not match with different capitalization makes sense and I lean towards you should keep it case sensitive. But also, I think Windows filesystem is on default case insensitive, so `C:\UE4` actually is the same location as `c:\ue4`. 

What I care about is, that it is clearly defined and ideally I can look at the tests to see how the matching behaves. So maybe one test for case sensitivity with 1 good (can be matched because same capitalization) and 1 bad example (cant be matched because different capitalization) is a good idea.

beckerhe:
> > The solution: No string-base comparison, but segment-wise comparison.
> 
> Thank you for the explanation! I indeed did not look at the commit messages. It makes a lot of sense now.
> 
> Regarding the case sensitivity. I am still a little torn. I think your explanation to not match with different capitalization makes sense and I lean towards you should keep it case sensitive. But also, I think Windows filesystem is on default case insensitive, so `C:\UE4` actually is the same location as `c:\ue4`.
> 
> What I care about is, that it is clearly defined and ideally I can look at the tests to see how the matching behaves. So maybe one test for case sensitivity with 1 good (can be matched because same capitalization) and 1 bad example (cant be matched because different capitalization) is a good idea.

Thanks. Done!

PR #2015, opened 2021-03-05
Adjust frame track colors to be similar when durations are similar

ronaldfw:
When all frame times are fairly similar, we want all colors to be fairly
similar, as there is no need to point out small differences by using
a different color. We solve this by adjusting the maximum upwards a bit
when computing colors (and only then!) and when the minimum and
maximum are close to each other. Colors are still interpolated between
the real minimum and the adjusted maximum, so there will be still some
variation in color.

When the maximum is "large enough" compared with the minimum, 
the behavior is not changed.

Tested: Manually tested frame track.
Bug: http://b/174224286

### DIFF
--- a/src/OrbitGl/FrameTrack.cpp
+++ b/src/OrbitGl/FrameTrack.cpp
@@ -110,21 +110,28 @@ Color FrameTrack::GetTimerColor(const orbit_client_protos::TimerInfo& timer_info
   Vec4 color;
   uint64_t timer_duration_ns = timer_info.end() - timer_info.start();
 
+  // A note on overflows here and below: The times in uint64_t represent durations of events
+  // in nanoseconds. This means the maximum duration is ~600 years. That is, multiplying by values
+  // in the single digits (and even much higher) as done here does not cause any issues.
   if (timer_duration_ns >= kHeightCapAverageMultipleUint64 * stats_.average_time_ns()) {
     color = warn_color;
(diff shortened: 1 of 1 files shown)


ronaldfw:
Screenshot before the change:
frame_track_original

Screenshot after the change:
frame_track_new

karupayun (approved):
LGTM Ronald!!! Clearly an improvement!

I posted 2 images here.

The first one is VkQueuePresent not being that constantly as your screenshot. You could seem there some blue columns, which I think that they are ok.
first VkQueuePresent

But this second one has one VkQueue frame timer with duration 12.29s which makes all the rest being too blueish. 
second example

This makes me think that we should still use the average instead of the minumum. And instead of making a difference with the bigger/smaller than the average, choose average + 10% instead. So, the difference started to be really noticed when you are 10% bigger than the average. 

What do you think?

ronaldfw:
> What do you think?

You are right, as long as we are using minimum and only the minimum, single outliers can completely mess up this scheme. I like your suggestion to base this on the average.

ronaldfw:
For your capture:

Old code (before this PR):

example_karupayun_before

New code (current version, based on your suggestion):

example_karupayun_new

karupayun:
I like it! :)

LGTM again!

ronaldfw:
Thanks for the review @karupayun

PR #2019, opened 2021-03-06
unique_fd is no longer unique_resource

dimitry-:
unique_fd can no longer be reset to another (int) fd. In can only be
constructed or moved from anther unique_fd.

Test: run OrbitBase unit-tests

### DIFF
--- a/src/OrbitBase/include/OrbitBase/File.h
+++ b/src/OrbitBase/include/OrbitBase/File.h
@@ -13,11 +13,54 @@
 
 #include "OrbitBase/Result.h"
 #include "OrbitBase/SafeStrerror.h"
-#include "OrbitBase/UniqueResource.h"
+
+#if defined(__linux)
+#include <unistd.h>
+#elif defined(_WIN32)
+#include <io.h>
+#endif
 
 namespace orbit_base {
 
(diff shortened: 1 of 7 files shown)


src/OrbitBase/include/OrbitBase/File.h
```diff
@@ -13,11 +13,52 @@
...
+  ~unique_fd() noexcept { release(); }
+
+  unique_fd(const unique_fd&) = delete;
+  unique_fd& operator=(const unique_fd&) = delete;
+
+  constexpr unique_fd(unique_fd&& other) noexcept : fd_{other.fd_} { other.fd_ = kInvalidFd; }
+
+  unique_fd& operator=(unique_fd&& other) noexcept {
+    reset(other.fd_);
+    other.fd_ = kInvalidFd;
```
beckerhe, line 40: I think you need to guard againat self-move!

beckerhe (approved):
Otherwise LGTM

src/OrbitBase/include/OrbitBase/File.h
dimitry-: right, thanks!

dimitry-: done, and added a test.

PR #2023, opened 2021-03-07
Add method to load dynsym symbols from elf file.

danielfenner:
Previously only symtab was considered. Now there are two
methods - on for symtab, one for dynsym.

The call sites of `LoadSymbols()` have been updated to use the
new name `LoadSymbolsFromSymtab()`

Added unit test.

Bug: http://b/181304023
Test: Unit tests.


src/ElfUtils/ElfFile.cpp
```diff
@@ -197,34 +230,14 @@ ErrorMessageOr<ModuleSymbols> ElfFileImpl<ElfT>::LoadSymbols() {
...
-      continue;
-    }
-
-    uint64_t symbol_address = symbol_ref.getValue();
-    SymbolInfo* symbol_info = module_symbols.add_symbol_infos();
-    symbol_info->set_name(name);
-    symbol_info->set_demangled_name(demangled_name);
-    symbol_info->set_address(symbol_address);
-    symbol_info->set_size(symbol_ref.getSize());
+  for (const llvm::object::ELFSymbolRef& symbol_ref : object_file_->getDynamicSymbolIterators()) {
```
dimitry-, line 233: this only makes sense if .symtab is not present. The dynamic symbol set is always a subset of .symtab

src/ElfUtils/ElfFile.cpp
```diff
@@ -182,12 +189,38 @@ void ElfFileImpl<ElfT>::InitSections() {
   }
 }
 
+bool AddSymbol(const llvm::object::ELFSymbolRef& symbol_ref, ModuleSymbols& module_symbols,
+               std::string_view file_path) {
```
dimitry-, line 193: maybe return std::optional here or ErrorMessageOr<SymbolInfo>. The caller can then move it by doing
*module_symbols.add_symbol_infos() = std::move(symbol_or_error.value());

danielfenner:
I'm not sure if this is done well: Previously the elf file has been used to provide the full debug info. I now added parsing the dynsym's also. Adding the check in SymbolHelper restores the old behaviour and this is probably what we want in this case: If the full information is missing we want the SymbolHelper to fail. 

Alternatively we could replace the `LoadSymbols` method by `LoadSymtabSymbols` and `LoadDynsymSymbols`. Having looked at the change after I pushed it this seems now the right thing to do.

Any thoughts?

Also currently the symbols that occur in dynsym and also symtab gets added twice - that would be easy to fix if we agree the PR is good in principle. And there needs to be some test.

src/ElfUtils/ElfFile.cpp
dimitry-: (It might make sense to split it into LoadSymbols, LoadSymbolsFromSymtab, LoadSymbolsFromDynsym depending on your use-case, with LoadSymbols trying to load fromSymtab and if it is not available falling back to dynsym)

src/OrbitCore/SymbolHelper.cpp
```diff
@@ -232,6 +232,10 @@ ErrorMessageOr<ModuleSymbols> SymbolHelper::LoadSymbolsFromFile(const fs::path&
                                         file_path.string(), elf_file_or_error.error().message()));
   }
 
+  if (!elf_file_or_error.value()->HasSymtab()) {
+    return ErrorMessage("ELF file does not have a .symtab section.");
+  }
```
dimitry-, line 237: in which case this could just call LoadSymbolsFromSymtab if we don't want to fall-back to .dynsym

dimitry-:
(and also please add a test/tests for this)

src/ElfUtils/ElfFile.cpp
danielfenner: I cleaned up the method a bit and it and kept the adding of the symbol inside this method. Turns out the returned bool is really not needed.
Please take a look, if you still think the add_symbol_infos should go into the caller please let me know.

src/OrbitCore/SymbolHelper.cpp
danielfenner: done

src/ElfUtils/ElfFile.cpp
danielfenner: splitted into two methods as suggested

src/ElfUtils/ElfFile.cpp
dimitry-: personal preference: I would still rather have it return SymbolInfo, and do the add to container part on the caller side. And maybe rename it to CreateSymbolInfo(const llvm::object::ELFSymbolRef& symbol_ref). file_path seems to only be used for logging, if this function is moved to ElfFileImpl<ElfT> it can just access it from a field.

dimitry- (approved):
otherwise lgtm

src/ElfUtils/ElfFile.cpp
danielfenner: added CreateSymbolInfo to ElfFileImpl

PR #2025, opened 2021-03-08
Fix margin's size after zoom in CodeViewer

karupayun:
We were using fontsize from TopWidget but we weren't updated it after
every zoom. So, as we are using the same font size in the entire
CodeViewer, we took it from there.

[b/181837657](http://b/181837657)

### DIFF
--- a/src/CodeViewer/Viewer.cpp
+++ b/src/CodeViewer/Viewer.cpp
@@ -390,6 +390,6 @@ int Viewer::WidthMarginBetweenColumns() const {
   return StringWidthInPixels(fontMetrics(), kTwoSpaces);
 }
 
-int Viewer::TopWidgetHeight() const { return top_bar_widget_.fontMetrics().height(); }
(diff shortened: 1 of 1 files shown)


beckerhe (approved):

beckerhe:
> We were using fontsize from TopWidget but we weren't updated it after
> every zoom. So, as we are using the same font size in the entire
> CodeViewer, we took it from there.
> 
> [b/181837657](http://b/181837657)

Just for your interest: The original code should actually work, because the top_bar_widget should inherit the fontMetrics from the parent widget. But it does not work, when the Orbit-Qt-style is enabled which is definitely a bug.

karupayun:
> Just for your interest: The original code should actually work, because the top_bar_widget should inherit the fontMetrics from the parent widget. But it does not work, when the Orbit-Qt-style is enabled which is definitely a bug.

Ahh, makes sense. I didn't understand why it worked for CodeViewerDemo's dialog. Shall we create a bug for that?

PR #2027, opened 2021-03-08
Fix expected counts of tracks after filtering

reichlfl:
This fixes newly introduced issues in the E2E test "orbit_tracks":

1) The expected count after filtering was wrong - the "all_threads" thread was renamed to report its name as "All Threads" instead of the process name, but this does not affect filtering, so the expected # of tracks matching "hello_ggp_" in the E2E test is still 2...
2) The E2E test expects a minimum of 5 tracks, but proceeds to test movement for track with zero-based index 5. This could potentially fail.

### DIFF
--- a/contrib/automation_tests/orbit_tracks.py
+++ b/contrib/automation_tests/orbit_tracks.py
@@ -20,18 +20,18 @@ def main(argv):
         MatchTracks(expected_names=[
             "Scheduler", ("sdma0", "vce0"), "gfx", "All Threads", "hello_ggp_stand"],
             allow_additional_tracks=True),
-        SelectTrack(track_index=5),
+        SelectTrack(track_index=4),
         DeselectTrack(),
         SelectTrack(track_index=0, expect_failure=True),   # Scheduler track cannot be selected
-        MoveTrack(track_index=5, new_index=0),
+        MoveTrack(track_index=4, new_index=0),
         MoveTrack(track_index=0, new_index=3),
-        MoveTrack(track_index=3, new_index=5),
+        MoveTrack(track_index=3, new_index=4),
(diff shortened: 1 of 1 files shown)


antonrohr (approved):
lgtm. Thank you for fixing that!!

contrib/automation_tests/orbit_tracks.py
```diff
@@ -20,18 +20,18 @@ def main(argv):
...
         MoveTrack(track_index=0, new_index=3),
-        MoveTrack(track_index=3, new_index=5),
+        MoveTrack(track_index=3, new_index=4),
         # TODO: The numbers below are very pessimistic, but it's not assured additional tracks like
         # GgpSwapChain, GgpVideoIpcRead etc are present - GgpSwapChain is missing on the DevKit, others
         # depend on the samples that have been taken
-        FilterTracks(filter_string="hello", expected_count=1),
-        FilterTracks(filter_string="Hello", expected_count=1),
-        FilterTracks(filter_string="ggp", expected_count=1, allow_additional_tracks=True),
+        FilterTracks(filter_string="hello", expected_count=2),
```
karupayun, line 32: Why 2? Which extra track we are counting? Scheduler one?

reichlfl: The main thread and the all_threads track.

karupayun (approved):
LGTM! BTW, after solving http://b/181671054, we should also change this expected behavior. Am I wrong?

reichlfl:
> LGTM! BTW, after solving http://b/181671054, we should also change this expected behavior. Am I wrong?

Yes, good point. Fixing this should change the expected numbers of tracks for all filtering tests.

PR #2030, opened 2021-03-08
Add integration test for memory tracing

vickyliu-go4it:
With this change, we add the integration test for the memory tracing.

Bug: http://b/182119930

### DIFF
--- a/src/MemoryTracing/MemoryTracingIntegrationTest.cpp
+++ b/src/MemoryTracing/MemoryTracingIntegrationTest.cpp
@@ -1,0 +1,135 @@
+// Copyright (c) 2021 The Orbit Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#include <absl/strings/substitute.h>
+#include <absl/synchronization/mutex.h>
+#include <absl/time/clock.h>
+#include <gmock/gmock.h>
+#include <gtest/gtest.h>
+
+#include "GrpcProtos/Constants.h"
+#include "MemoryTracing/MemoryInfoListener.h"
+#include "MemoryTracing/MemoryInfoProducer.h"
+#include "OrbitBase/Logging.h"
+#include "capture.pb.h"
+
+namespace orbit_memory_tracing {
+namespace {
+
+using orbit_grpc_protos::kMissingInfo;
+using orbit_grpc_protos::ProducerCaptureEvent;
+using orbit_grpc_protos::SystemMemoryUsage;
+using orbit_memory_tracing::MemoryInfoListener;
+using orbit_memory_tracing::MemoryInfoProducer;
+
+class BufferMemoryInfoListener : public MemoryInfoListener {
+ public:
+  void OnSystemMemoryUsage(SystemMemoryUsage system_memory_usage) override {
+    ProducerCaptureEvent event;
+    *event.mutable_system_memory_usage() = std::move(system_memory_usage);
+    {
+      absl::MutexLock lock{&events_mutex_};
+      events_.emplace_back(std::move(event));
+    }
+  }
+
+  [[nodiscard]] std::vector<ProducerCaptureEvent> GetAndClearEvents() {
+    absl::MutexLock lock{&events_mutex_};
+    std::vector<ProducerCaptureEvent> events = std::move(events_);
+    events_.clear();
+    return events;
+  }
+
+ private:
+  std::vector<ProducerCaptureEvent> events_;
+  absl::Mutex events_mutex_;
+};
+
+class MemoryTracingIntegrationTestFixture {
+ public:
+  explicit MemoryTracingIntegrationTestFixture(uint64_t memory_sampling_period_ns)
+      : memory_sampling_period_ns_(memory_sampling_period_ns) {}
+
+  void StartTracing() {
+    CHECK(!producer_.has_value());
+    CHECK(!listener_.has_value());
+    producer_.emplace(memory_sampling_period_ns_);
+    listener_.emplace();
+    producer_->SetListener(&*listener_);
+    producer_->Start();
+  }
+
+  [[nodiscard]] std::vector<ProducerCaptureEvent> StopTracingAndGetEvents() {
+    CHECK(producer_.has_value());
+    CHECK(listener_.has_value());
+    producer_->Stop();
+    producer_.reset();
+    std::vector<ProducerCaptureEvent> events = listener_->GetAndClearEvents();
+    listener_.reset();
+    return events;
+  }
+
+ private:
+  uint64_t memory_sampling_period_ns_;
+  std::optional<MemoryInfoProducer> producer_ = std::nullopt;
+  std::optional<BufferMemoryInfoListener> listener_ = std::nullopt;
+};
+
+[[nodiscard]] std::vector<ProducerCaptureEvent> TraceAndGetEvents(
+    MemoryTracingIntegrationTestFixture* fixture, absl::Duration tracing_period) {
+  CHECK(fixture != nullptr);
+
+  fixture->StartTracing();
+  absl::SleepFor(tracing_period);
+  return fixture->StopTracingAndGetEvents();
+}
+
+void VerifyOrderAndContentOfEvents(const std::vector<ProducerCaptureEvent>& events) {
+  uint64_t previous_event_timestamp_ns = 0;
+  for (const auto& event : events) {
+    CHECK(event.event_case() == ProducerCaptureEvent::kSystemMemoryUsage);
+
+    // Verify whether events are in order of their timestamps.
+    uint64_t current_event_timestamp_ns = event.system_memory_usage().timestamp_ns();
+    EXPECT_GE(current_event_timestamp_ns, previous_event_timestamp_ns);
+
+    // Verify whether the contents of events are valid.
+    EXPECT_TRUE(event.system_memory_usage().total_kb() > 0);
+    EXPECT_TRUE(event.system_memory_usage().free_kb() > 0);
+    EXPECT_TRUE(event.system_memory_usage().available_kb() > 0);
+    EXPECT_TRUE(event.system_memory_usage().buffers_kb() > 0);
+    EXPECT_TRUE(event.system_memory_usage().cached_kb() > 0);
+
+    previous_event_timestamp_ns = current_event_timestamp_ns;
+  }
+}
+
+// Verify whether the memory_sampling_period_ns_ works as expected by checking the number of
+// received events.
+void VerifyEventCounts(const std::vector<ProducerCaptureEvent>& events,
+                       size_t expected_event_counts) {
+  constexpr size_t kEventCountsErrorTolerance = 2;
+
+  size_t received_event_counts = events.size();
+  EXPECT_GE(received_event_counts, expected_event_counts - kEventCountsErrorTolerance);
+  EXPECT_LE(received_event_counts, expected_event_counts + kEventCountsErrorTolerance);
+}
+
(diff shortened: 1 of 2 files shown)


vickyliu-go4it:
As I'm not sure whether it is a good idea to verify whether the `memory_sampling_period_ns_` works as we expected, I put the relevant changes in a separated commit (i.e., the second one). I feel like it is hard to determine the tolerance of sampling time error, especially when the  `memory_sampling_period_ns_`  is small. 

I would like to hear more suggestions about the verification of `memory_sampling_period_ns_` . If this verification is unnecessary, I will drop the second commit.

src/MemoryTracing/MemoryTracingIntegrationTest.cpp
```diff
@@ -0,0 +1,129 @@
...
+    }
+
+    previous_event_timestamp_ns = current_event_timestamp_ns;
+  }
+}
+
+}  // namespace
+
+TEST(MemoryTracingIntegrationTest, SystemMemoryUsageTracing) {
+  constexpr uint64_t kMemorySamplingPeriodNs = 100000000;
```
dimitry-, line 121: nitpick: you can use "'" to separate groups of 000 so that it is easier to read: as in `100'000'000`. Or better yet use `absl::Milliseconds(100) / absl::Nanoseconds(1)`

src/MemoryTracing/MemoryTracingIntegrationTest.cpp
```diff
@@ -0,0 +1,129 @@
...
+  fixture->StartTracingAndWaitForTracingLoopStarted();
+
+  constexpr absl::Duration kSleepBeforeStopTracing = absl::Nanoseconds(1000000000);
+  absl::SleepFor(kSleepBeforeStopTracing);
+  return fixture->StopTracingAndGetEvents();
+}
+
+void VerifyOrderAndSamplingPeriodOfEvents(const std::vector<ProducerCaptureEvent>& events,
+                                          uint64_t sampling_period_ns) {
+  constexpr uint64_t kSamplingTimeErrorToleranceNs = 1000000;
```
dimitry-, line 90: same here use `absl::Milliseconds(10?)/absl::Nanoseconds(1)`

src/MemoryTracing/MemoryTracingIntegrationTest.cpp
```diff
@@ -0,0 +1,129 @@
...
+      uint64_t time_since_last_successful_sampling =
+          current_event_timestamp_ns - previous_event_timestamp_ns;
+      uint64_t periods_since_last_successful_sampling =
+          time_since_last_successful_sampling / sampling_period_ns;
+      // If there are failed attempts of memory sampling, the sampling time error would be
+      // difficult to decide. Hence, we only check the case of successive successful attempts of
+      // sampling.
+      if (periods_since_last_successful_sampling <= 1) {
+        EXPECT_GE(time_since_last_successful_sampling,
+                  sampling_period_ns - kSamplingTimeErrorToleranceNs);
```
dimitry-, line 108: this kind of seems flaky - I would say check number of events received against expected with +-1/2. It should be much less flaky since the main loop in the sampling thread is timed by deadlines.

src/MemoryTracing/MemoryTracingIntegrationTest.cpp
```diff
@@ -0,0 +1,129 @@
...
+  std::optional<BufferMemoryInfoListener> listener_ = std::nullopt;
+};
+
+[[nodiscard]] std::vector<ProducerCaptureEvent> TraceAndGetEvents(
+    MemoryTracingIntegrationTestFixture* fixture) {
+  CHECK(fixture != nullptr);
+
+  fixture->StartTracingAndWaitForTracingLoopStarted();
+
+  constexpr absl::Duration kSleepBeforeStopTracing = absl::Nanoseconds(1000000000);
```
dimitry-, line 83: use abls::Milliseconds() or Seconds()

src/MemoryTracing/MemoryTracingIntegrationTest.cpp
```diff
@@ -0,0 +1,129 @@
...
+  }
+}
+
+}  // namespace
+
+TEST(MemoryTracingIntegrationTest, SystemMemoryUsageTracing) {
+  constexpr uint64_t kMemorySamplingPeriodNs = 100000000;
+  MemoryTracingIntegrationTestFixture fixture(kMemorySamplingPeriodNs);
+
+  std::vector<ProducerCaptureEvent> events = TraceAndGetEvents(&fixture);
```
dimitry-, line 124: do you want to add some sanity checks for data? Check that memory consumption is non-zero for example.

dimitry- (approved):

src/MemoryTracing/MemoryTracingIntegrationTest.cpp
vickyliu-go4it: Changed to use `absl::Milliseconds(100) / absl::Nanoseconds(1)`. Thanks for the suggestion!

src/MemoryTracing/MemoryTracingIntegrationTest.cpp
vickyliu-go4it: Changed to use `absl::Milliseconds(1) / absl::Nanoseconds(1)`. Thanks!

src/MemoryTracing/MemoryTracingIntegrationTest.cpp
vickyliu-go4it: Changed to use `absl::Miliseconds(1000)`. Thanks for the suggestion!

src/MemoryTracing/MemoryTracingIntegrationTest.cpp
vickyliu-go4it: Updated this method to also checking the content of `SystemMemoryUsage`. Thanks for the suggestion!

src/MemoryTracing/MemoryTracingIntegrationTest.cpp
vickyliu-go4it: It is also difficult to decide the error tolerance of number of events, i.e., `+- n`, since we allow the failure of extracting `SystemMemoryUsage` from `/proc/meminfo`. When the `memory_sampling_period_ns_` becomes small, `# of received events - # of expected received events` would increase.

src/MemoryTracing/MemoryTracingIntegrationTest.cpp
```diff
@@ -0,0 +1,145 @@
...
+  }
+
+ private:
+  std::vector<ProducerCaptureEvent> events_;
+  absl::Mutex events_mutex_;
+};
+
+class MemoryTracingIntegrationTestFixture {
+ public:
+  MemoryTracingIntegrationTestFixture(uint64_t memory_sampling_period_ns)
```
dpallotti, line 51: `explicit`

src/MemoryTracing/MemoryTracingIntegrationTest.cpp
```diff
@@ -0,0 +1,145 @@
...
+      : memory_sampling_period_ns_(memory_sampling_period_ns) {}
+
+  void StartTracingAndWaitForTracingLoopStarted() {
```
dpallotti, line 54: From the name, what tracing loop is this waiting for?

src/MemoryTracing/MemoryTracingIntegrationTest.cpp
```diff
@@ -0,0 +1,145 @@
...
+  absl::SleepFor(kSleepBeforeStopTracing);
+  return fixture->StopTracingAndGetEvents();
+}
+
+void VerifyOrderAndContentOfEvents(const std::vector<ProducerCaptureEvent>& events,
+                                   uint64_t sampling_period_ns) {
+  const uint64_t kSamplingTimeErrorToleranceNs = absl::Milliseconds(1) / absl::Nanoseconds(1);
+  uint64_t previous_event_timestamp_ns = 0;
+  for (const auto& event : events) {
+    if (event.event_case() != ProducerCaptureEvent::kSystemMemoryUsage) UNREACHABLE();
```
dpallotti, line 95: It's strange to use `UNREACHABLE` like this, that's more for branches of `switch`es and `if`s that cannot be avoided.
In this case I would just write
`CHECK(event.event_case() == ProducerCaptureEvent::kSystemMemoryUsage)`

src/MemoryTracing/MemoryTracingIntegrationTest.cpp
dpallotti: > since we allow the failure of extracting SystemMemoryUsage from /proc/meminfo

Is this something we normally expect? I would imagine not, unless something is super strange about the machine, I guess `/proc/meminfo` should always be well formed.

src/MemoryTracing/MemoryTracingIntegrationTest.cpp
```diff
@@ -0,0 +1,145 @@
...
+    EXPECT_TRUE(event.system_memory_usage().total_kb() == kMissingInfo ||
+                event.system_memory_usage().total_kb() > 0);
+    EXPECT_TRUE(event.system_memory_usage().free_kb() == kMissingInfo ||
+                event.system_memory_usage().free_kb() > 0);
+    EXPECT_TRUE(event.system_memory_usage().available_kb() == kMissingInfo ||
+                event.system_memory_usage().available_kb() > 0);
+    EXPECT_TRUE(event.system_memory_usage().buffers_kb() == kMissingInfo ||
+                event.system_memory_usage().buffers_kb() > 0);
+    EXPECT_TRUE(event.system_memory_usage().cached_kb() == kMissingInfo ||
+                event.system_memory_usage().cached_kb() > 0);
```
dpallotti, line 128: Do we actually normally expect information to be missing? I would expect them to always be there and hence the part after the `||` to be enough.

src/MemoryTracing/MemoryTracingIntegrationTest.cpp
vickyliu-go4it: Make sense! Changed accordingly. Thanks!

src/MemoryTracing/MemoryTracingIntegrationTest.cpp
vickyliu-go4it: Thanks for raising this issue.

In the `ParseMemInfo` method, I initialize the fields in the `SystemMemoryUsage` event with `kMissingInfo`. If we assume the `proc/meminfo` file is always well-formatted and so we can successfully extract the information we need, should I change the logic in the `ParseMemInfo` method? I'm thinking that the expected information in the test and in the method where we produce the event should be the same.

src/MemoryTracing/MemoryTracingIntegrationTest.cpp
dimitry-: I think checking for count is somewhat less flaky because the implementation compensates for wait time flactuations. And we expect it to.

dpallotti (approved):
Looks good otherwise.

dimitry- (approved):
LGTM, agree with @dpallotti comment about kernel expected behavior,

src/MemoryTracing/MemoryTracingIntegrationTest.cpp
vickyliu-go4it: Thanks for the explanation! I removed the comparison with `kMissingInfo`.

src/MemoryTracing/MemoryTracingIntegrationTest.cpp
vickyliu-go4it: Thanks for the explanation! Changed to check the received event counts.

PR #2031, opened 2021-03-08
Main timer checks the dirty flag before updates

reichlfl:
Reduces the GPU resources required in idle mode from 3% to 0.1% - see b/179978894

To reproduce: Start Orbit without taking a capture, open the task manager. There's a few percent of GPU utilization before this fix which should disappear afterwards.

Tests: Redraws and interaction with the capture window, debug window and introspection window work fine afterwards.

### DIFF
--- a/src/OrbitQt/orbitmainwindow.cpp
+++ b/src/OrbitQt/orbitmainwindow.cpp
@@ -782,10 +782,14 @@ void OrbitMainWindow::OnTimer() {
   app_->MainTick();
 
   for (OrbitGLWidget* gl_widget : gl_widgets_) {
-    gl_widget->update();
+    if (gl_widget->GetCanvas() != nullptr && gl_widget->GetCanvas()->GetNeedsRedraw()) {
+      gl_widget->update();
+    }
   }
 
-  filter_panel_action_->SetTimerLabelText(QString::fromStdString(app_->GetCaptureTime()));
+  if (app_->IsCapturing()) {
+    filter_panel_action_->SetTimerLabelText(QString::fromStdString(app_->GetCaptureTime()));
+  }
 }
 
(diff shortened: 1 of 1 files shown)


src/OrbitQt/orbitmainwindow.cpp
```diff
@@ -782,10 +782,14 @@ void OrbitMainWindow::OnTimer() {
   app_->MainTick();
 
   for (OrbitGLWidget* gl_widget : gl_widgets_) {
-    gl_widget->update();
+    if (gl_widget->GetCanvas() != nullptr && gl_widget->GetCanvas()->GetNeedsRedraw()) {
```
karupayun, line 785: Probably a stupid question, but are you sure about this? What happens if GetNeedsRedraw was set into false after a redrawing the CaptureWindow?

reichlfl: I'm not sure I understand the issue you're seeing. "Needs redraw" is set whenever anything happened that requires redraw, and is only cleared when that redraw happened. If it returns `false` at this point, it should be save to assume all contents of the capture window are drawn and up-to-date, or is there some case I'm missing?

karupayun: Yes, of course. Was a stupid question :P

karupayun (approved):
LGTM! :) Nice improvement!

pierricgimmig:
Hi @reichlfl , thanks for the change! I tested it and it seems to work perfectly. I added logs to see how often the event filtering is done in "OrbitGLWidget::eventFilter", it's way less than I anticipated (I am seeing < 1 event/second). I wonder if something changed in Qt or maybe it's a side effect of something on our side. Anyhow, thanks for the fix.

pierricgimmig (approved):

PR #2032, opened 2021-03-08
Make ShowSourceCode work in LiveFunctionsDataView::OnContextMenu

beckerhe:
`LiveFunctionsDataView::OnContextMenu` was never calling
`ShowSourceCode` since the action is checked twice in two nested if's.

Only the second if was taking the source code view into account. No need
to say that this needs to be refactored.


florian-kuebler (approved):
LGTM, thanks!

PR #2033, opened 2021-03-08
Add "process_name" parameter

reichlfl:
When specified, Orbit will automatically connect to this process and
open the main window after instance connection is established.

When combined with automatic instance connection, this lets Orbit go
through the complete connection process automatically.

I've mainly added this to speed up my own iterations - which in hindsight
wasn't needed had I known about local ssh tunneling, but IMO this is a neat
addition and a first step towards automatically connecting to the last
process, WDYT?

### DIFF
--- a/src/OrbitQt/ProfilingTargetDialog.cpp
+++ b/src/OrbitQt/ProfilingTargetDialog.cpp
@@ -51,6 +51,7 @@
 #include "servicedeploymanager.h"
 
 ABSL_DECLARE_FLAG(bool, local);
+ABSL_DECLARE_FLAG(std::string, process_name);
 
 namespace {
 constexpr int kProcessesRowHeight = 19;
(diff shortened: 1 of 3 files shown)


src/OrbitQt/ProfilingTargetDialog.cpp
```diff
@@ -462,10 +463,15 @@ void ProfilingTargetDialog::OnProcessListUpdate(
     // TrySelectProcess attempts to select the process again if it exists
     // in process_model_.
     if (process_ != nullptr) {
-      bool success = TrySelectProcess(*process_);
+      bool success = TrySelectProcess(process_->name());
```
beckerhe, line 466: What happens when two processes have the same name? (Maybe a game forks or whatever). Is there a chance, that the selection jumps around between these two (or more) processes with the same name?

beckerhe (approved):
Otherwise LGTM!

src/OrbitQt/ProfilingTargetDialog.cpp
reichlfl: Good point. Yes, that could happen, if the processes switch positions due to changing CPU load. If you specify a `process_name` parameter, the Dialog will automatically close after finding a matching process, so you won't see any jumping of selections, but will end up with the first of multiple matches selected.

The jumping will occur when you click "End Session" and go back to the startup window. But this problem isn't newly introduced, the implementation before was only checking the process name in `TrySelectProcess` as well...

I'm not sure if this is something we should take care of. We could pass an optional PID to `TrySelectProcess` and pick that if there's multiple options, WDYT? But I'd probably create another PR for this.

beckerhe: Ahh, I'm sorry. I didn't realize that the previous implementation had the same problem. So I'm fine with shipping it like this.

Of course, matching a PID would be better, but I don't want to put the burden of fixing random discovered bugs on you.

PR #2034, opened 2021-03-09
IntrospectionWindow scopes its own instrumented methods

reichlfl:
I'm not sure if this was clear to everyone, but the introspection window reports
its own `Draw` and `UpdatePrimitives` methods in the same timeline, which makes 
it hard to figure out which calls you're actually interested in.

A quick fix: I've overloaded all methods that are instrumented in `CaptureWindow`
and added a scope on top of that in `IntrospectionWindow`.
This makes it obvious where the introspection window is reporting
performance about itself rather than the main window if you take a look at the hierarchy
of instrumented functions.

Example image: *Left* is a stack from the main capture window, *Right* is from the 
introspection.

image

### DIFF
--- a/src/OrbitGl/IntrospectionWindow.cpp
+++ b/src/OrbitGl/IntrospectionWindow.cpp
@@ -50,6 +50,21 @@ void IntrospectionWindow::StartIntrospection() {
 }
 void IntrospectionWindow::StopIntrospection() { introspection_listener_ = nullptr; }
 
+void IntrospectionWindow::Draw() {
+  ORBIT_SCOPE_FUNCTION;
+  CaptureWindow::Draw();
+}
+
+void IntrospectionWindow::DrawScreenSpace() {
+  ORBIT_SCOPE_FUNCTION;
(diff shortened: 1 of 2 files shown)


src/OrbitGl/IntrospectionWindow.cpp
```diff
@@ -50,6 +50,21 @@ void IntrospectionWindow::StartIntrospection() {
 }
 void IntrospectionWindow::StopIntrospection() { introspection_listener_ = nullptr; }
 
+void IntrospectionWindow::Draw() {
+  ORBIT_SCOPE("IntrospectionWindow::Draw");
+  CaptureWindow::Draw();
```
pierricgimmig, line 55: ORBIT_SCOPE_FUNCTION could be used also here

pierricgimmig (approved):
Cool, thanks for addressing the issue!

src/OrbitGl/IntrospectionWindow.cpp
reichlfl: Thanks, I missed that... changed.

PR #2035, opened 2021-03-09
Add nodiscard to ErrorMessage and ErrorMessageOr

dimitry-:
Test: builds


beckerhe (approved):
Ahhh: :sweat_smile:  It seems we missed that last time. Thanks a lot!

PR #2036, opened 2021-03-09
Automatically upload Vulkan layer symbols.

florian-kuebler:


kokoro/builds/upload_symbols.sh
```diff
@@ -72,6 +72,7 @@ function upload_symbol_file {
 function upload_debug_symbols {
   local api_key=$1
   local bin_folder=$2
```
IrinaShkviro, line 74: We also can change the second argument to the `build_folder`, and add '/bin' or '/lib' suffix below (when call upload_symbol_file function). But I don't have a strong opinion on this, so whatever you prefer.

IrinaShkviro (approved):
LGTM!

beckerhe (approved):
LGTM!

But I suggest we test this before merging. We could start a nightly build from this branch. The logfile should tell us whether it worked or not.

florian-kuebler:
> LGTM!
> 
> But I suggest we test this before merging. We could start a nightly build from this branch. The logfile should tell us whether it worked or not.

Sounds good. Thanks!

kokoro/builds/upload_symbols.sh
florian-kuebler: I thought about this. The main reason why I did it that way, was to break if I missed a usage of that function.

florian-kuebler:
> > LGTM!
> > But I suggest we test this before merging. We could start a nightly build from this branch. The logfile should tell us whether it worked or not.
> 
> Sounds good. Thanks!

I've run the nightly collect build. It is green, but from the log file, I don't see any output related to symbol uploading.

PR #2037, opened 2021-03-09
Log and potentially warn about low clock resolution on the client

ronaldfw:
We had seen an issue where low clock resolution on our Windows client
led to highly inaccurate introspection timings (resolution was 1ms). We
now log and (in devmode) warn users when the estimated clock resolution
is low and hence timings may be inaccurate.

I tested this with the Windows clock we used previously and indeed
had to adjust some constants in the clock resolution estimation to even
get the right resolution.

Tested: Manually ran Orbit with -devmode to check if the warning
appears with the clock we used previously on Windows.
Bug: http://b/180995185

### DIFF
--- a/src/OrbitQt/main.cpp
+++ b/src/OrbitQt/main.cpp
@@ -43,6 +43,7 @@
 #include "MetricsUploader/MetricsUploader.h"
 #include "OrbitBase/CrashHandler.h"
 #include "OrbitBase/Logging.h"
+#include "OrbitBase/Profiling.h"
 #include "OrbitSsh/Context.h"
 #include "OrbitSshQt/ScopedConnection.h"
 #include "OrbitVersion/OrbitVersion.h"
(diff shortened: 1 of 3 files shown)


src/OrbitQt/main.cpp
```diff
@@ -257,6 +258,30 @@ static bool DevModeEnabledViaEnvironmentVariable() {
   return env.contains("ORBIT_DEV_MODE") || env.contains("ORBIT_DEVELOPER_MODE");
 }
 
+static void LogAndMaybeWarnAboutClockResolution() {
+  uint64_t estimated_clock_resolution = orbit_base::EstimateClockResolution();
+  LOG("%s",
+      absl::StrFormat("Clock resolution on client: %" PRIu64 " (ns)", estimated_clock_resolution));
```
beckerhe, line 264: Nit: PRIu64 is not needed for absl::StrFormat. `%d` is totally fine here.

beckerhe: Ref: https://abseil.io/docs/cpp/guides/format#conversion-specifiers

beckerhe (approved):
LGTM

src/OrbitQt/main.cpp
```diff
@@ -257,6 +258,30 @@ static bool DevModeEnabledViaEnvironmentVariable() {
...
+
+  // Since a low clock resolution on the client only affects our own introspection and logging
+  // timings, we only show a warning dialogue when running in devmode.
+  constexpr uint64_t kWarnThresholdClockResolutionNs = 10 * 1000;  // 10 us
+  if (absl::GetFlag(FLAGS_devmode) &&
+      estimated_clock_resolution > kWarnThresholdClockResolutionNs) {
+    DisplayErrorToUser(
+        QString("Warning, clock resolution is low (estimated as %1 ns)! Introspection "
```
beckerhe, line 272: After seeing the screenshot: You could use `(estimated as %L1 ns)` to get localized number formatting. This should give you the appropriate thousands-separator (if appropriate at all)

ronaldfw:
Thanks for the review @beckerhe and the good suggestions!

PR #2039, opened 2021-03-09
Simplify `CompleteSubmits` by using a priority queue.

florian-kuebler:
Prior to this change, `CompleteSubmits` (which gets called
on `vkQueuePresetKHR`) started by calling `PullCompletedSubmissions`,
which estimated which submissions were finished (by inspecting the
last command buffer in there). However, it could still be
that those submissions were not actually finished, so
we implemented a push back mechanism. Also, we needed to sort
the result every time in `PullCompletedSubmissions`.

This change simplifies this logic, by maintaining a priority
queue, sorted (inverse) by the cpu timestamps.
In `CompleteSubmits`, we now iterate over that queue (starting
by the oldest submission), and immediately stop as soon as
the first timestamp is not ready. This way the code is less
complex and also we don't query results twice.

Test: Run tests & take captures on Infiltrator/Trata

### DIFF
--- a/src/OrbitVulkanLayer/SubmissionTracker.h
+++ b/src/OrbitVulkanLayer/SubmissionTracker.h
@@ -5,6 +5,8 @@
 #ifndef ORBIT_VULKAN_LAYER_SUBMISSION_TRACKER_H_
 #define ORBIT_VULKAN_LAYER_SUBMISSION_TRACKER_H_
 
+#include <functional>
+#include <queue>
 #include <stack>
 
 #include "OrbitBase/Logging.h"
@@ -525,18 +527,28 @@ class SubmissionTracker : public VulkanLayerProducer::CaptureStatusListener {
       return;
     }
 
-    if (!queue_to_submissions_.contains(queue)) {
-      queue_to_submissions_[queue] = {};
+    if (!submissions_queue_.contains(queue)) {
+      std::priority_queue<QueueSubmission, std::vector<QueueSubmission>,
+                          std::function<bool(QueueSubmission, QueueSubmission)>>
+          priority_queue{[](const QueueSubmission& lhs, const QueueSubmission& rhs) -> bool {
+            return lhs.meta_information.pre_submission_cpu_timestamp >
+                   rhs.meta_information.pre_submission_cpu_timestamp;
+          }};
+      submissions_queue_[queue] = priority_queue;
     }
-    CHECK(queue_to_submissions_.contains(queue));
-    queue_to_submissions_.at(queue).emplace_back(std::move(queue_submission_optional.value()));
+
+    submissions_queue_.at(queue).emplace(std::move(queue_submission_optional.value()));
   }
 
   // This method is responsible for retrieving all the timestamps for the "completed" submissions,
-  // for transforming the information of those submissions (in particlular about the command buffers
+  // for transforming the information of those submissions (in particular about the command buffers
   // and debug markers) into the `GpuQueueSubmission` proto, and for sending it to the
-  // `VulkanLayerProducer`. We consider a submission to be "completed" if its last command buffer
-  // timestamp is ready (see `PullCompletedSubmissions`).
+  // `VulkanLayerProducer`. We consider a submission to be "completed" all timestamps that are
+  // associated with this submission are ready.
+  // We maintain a priority queue (for every `VkQueue`) `submissions_queue_` and process submissions
+  // with the oldest CPU timestamp, until we encounter the first "incomplete" submission.
+  // This way, we ensure, that we will send the submission information per queue in ordered by the
+  // CPU timestamp.
   // Beside the timestamps of command buffers and the meta information of the submission, the proto
   // also contains the debug markers, "begin" (even if submitted in a different submission) and
   // "end", that got completed in this submission.
@@ -546,10 +558,8 @@ class SubmissionTracker : public VulkanLayerProducer::CaptureStatusListener {
   void CompleteSubmits(VkDevice device) {
     absl::WriterMutexLock lock(&mutex_);
     VkQueryPool query_pool = timer_query_pool_->GetQueryPool(device);
-    std::vector<QueueSubmission> completed_submissions =
-        PullCompletedSubmissions(device, query_pool);
 
-    if (completed_submissions.empty()) {
+    if (submissions_queue_.empty()) {
       return;
     }
 
@@ -560,29 +570,31 @@ class SubmissionTracker : public VulkanLayerProducer::CaptureStatusListener {
     std::vector<uint32_t> query_slots_done_reading = {};
     std::vector<QueueSubmission> submissions_to_send = {};
 
-    // The submits in completed_submissions are sorted by "pre submission CPU" timestamp and we want
-    // to make sure we send events to the client in that order. Therefore, we must retry all
-    // submissions after we encounter the first submission that we can't complete yet and push them
-    // back into the queue_to_submissions_ queue.
-    bool all_succeeded_so_far = true;
-    for (auto& completed_submission : completed_submissions) {
-      bool command_buffer_queries_succeeded = QueryCommandBufferTimestamps(
-          &completed_submission, &query_slots_done_reading, device, query_pool, timestamp_period);
-      bool marker_queries_succeeded = QueryDebugMarkerTimestamps(
-          &completed_submission, &query_slots_done_reading, device, query_pool, timestamp_period);
+    // The submits of a specific queue in `submissions_queue_` are sorted by "pre submission CPU"
+    // timestamp and we want to make sure we send events to the client in that order. Therefore, we
+    // stop as soon as a query failed.
+    for (auto& [unused_queue, submissions] : submissions_queue_) {
+      while (!submissions.empty()) {
+        QueueSubmission completed_submission = submissions.top();
+        submissions.pop();
+        bool command_buffer_queries_succeeded = QueryCommandBufferTimestamps(
+            &completed_submission, &query_slots_done_reading, device, query_pool, timestamp_period);
 
-      all_succeeded_so_far &= command_buffer_queries_succeeded && marker_queries_succeeded;
-
-      if (all_succeeded_so_far) {
-        submissions_to_send.push_back(completed_submission);
-      } else {
-        VkQueue queue = completed_submission.queue;
-
-        if (!queue_to_submissions_.contains(queue)) {
-          queue_to_submissions_[queue] = {};
+        // We only need to read the debug marker timestamps, if querying the command buffers
+        // succeeded.
+        bool marker_queries_succeeded = false;
+        if (command_buffer_queries_succeeded) {
+          marker_queries_succeeded =
+              QueryDebugMarkerTimestamps(&completed_submission, &query_slots_done_reading, device,
+                                         query_pool, timestamp_period);
         }
-        CHECK(queue_to_submissions_.contains(queue));
-        queue_to_submissions_.at(queue).emplace_back(std::move(completed_submission));
+
+        if (command_buffer_queries_succeeded && marker_queries_succeeded) {
+          submissions_to_send.emplace_back(std::move(completed_submission));
+        } else {
+          submissions.emplace(std::move(completed_submission));
+          break;
+        }
       }
     }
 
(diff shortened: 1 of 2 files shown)


src/OrbitVulkanLayer/SubmissionTracker.h
```diff
@@ -1029,7 +985,10 @@ class SubmissionTracker : public VulkanLayerProducer::CaptureStatusListener {
   absl::flat_hash_map<VkCommandBuffer, VkDevice> command_buffer_to_device_;
 
   absl::flat_hash_map<VkCommandBuffer, CommandBufferState> command_buffer_to_state_;
-  absl::flat_hash_map<VkQueue, std::vector<QueueSubmission>> queue_to_submissions_;
+  absl::flat_hash_map<VkQueue,
```
ronaldfw, line 988: Have you tried just using a single priority queue? That is, one for all queues. That would simplify the code a bit more, but I'm a bit worried that a single queue might block the others in that case (e.g. long compute jobs in the compute queue).

src/OrbitVulkanLayer/SubmissionTracker.h
```diff
@@ -592,15 +604,19 @@ class SubmissionTracker : public VulkanLayerProducer::CaptureStatusListener {
...
 
       WriteMetaInfo(completed_submission.meta_information, submission_proto->mutable_meta_info());
-      WriteCommandBufferTimings(completed_submission, submission_proto);
-      WriteDebugMarkers(completed_submission, submission_proto);
+      bool has_command_buffer_timestamps =
+          WriteCommandBufferTimings(completed_submission, submission_proto);
+      bool has_debug_marker_timestamps = WriteDebugMarkers(completed_submission, submission_proto);
 
-      if (vulkan_layer_producer_ != nullptr) {
+      if (vulkan_layer_producer_ != nullptr &&
```
ronaldfw, line 611: Not related to this change directly, but what is actually the case when vulkan_layer_producer_ is nullptr? In that case we never send any events, so all work done in the layer doesn't actually have any outcome. Can we early out in this function (or even earlier?) when vulkan_layer_producer_ is nullptr or do we still need to do the work to maintain the state? In other words, can this be a CHECK here with graceful error handling done earlier?

src/OrbitVulkanLayer/SubmissionTracker.h
florian-kuebler: I haven't tried this. Unless there are compute queue jobs that takes several seconds, I would not worry too much about this. So it might be worth trying. I'll do some tests tomorrow. 

In general, I like the split here, as I actually find it more understandable. Also it may allow us to do more advanced locking in the vkQueueSubmit part.

src/OrbitVulkanLayer/SubmissionTracker.h
```diff
@@ -856,33 +804,32 @@ class SubmissionTracker : public VulkanLayerProducer::CaptureStatusListener {
...
                                                   std::vector<uint32_t>* query_slots_to_reset,
                                                   VkDevice device, VkQueryPool query_pool,
                                                   float timestamp_period) {
-    bool all_queries_succeeded = true;
     for (auto& completed_submit : completed_submission->submit_infos) {
       for (auto& completed_command_buffer : completed_submit.command_buffers) {
-        all_queries_succeeded &= QuerySingleCommandBufferTimestamps(
+        bool queries_succeeded = QuerySingleCommandBufferTimestamps(
             &completed_command_buffer, query_slots_to_reset, device, query_pool, timestamp_period);
+        if (!queries_succeeded) return false;
```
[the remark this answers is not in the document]
florian-kuebler: I found it more readable, which might be an indication that the function name should be renamed... Will think about this.

src/OrbitVulkanLayer/SubmissionTracker.h
florian-kuebler: @dpallotti correct me if I am wrong, but I think this may only happen if the producer actually dies. Otherwise, we should not end up here. Might be, that we can even turn this into a CHECK.

In that case, I guess we need to do the work, especially to reset the timer slots.

src/OrbitVulkanLayer/SubmissionTracker.h
```diff
@@ -525,18 +527,28 @@ class SubmissionTracker : public VulkanLayerProducer::CaptureStatusListener {
...
 
-    if (!queue_to_submissions_.contains(queue)) {
-      queue_to_submissions_[queue] = {};
+    if (!submissions_queue_.contains(queue)) {
+      std::priority_queue<QueueSubmission, std::vector<QueueSubmission>,
+                          std::function<bool(QueueSubmission, QueueSubmission)>>
+          priority_queue{[](const QueueSubmission& lhs, const QueueSubmission& rhs) -> bool {
+            return lhs.meta_information.pre_submission_cpu_timestamp >
+                   rhs.meta_information.pre_submission_cpu_timestamp;
+          }};
```
dpallotti, line 536: I suggest defining the comparator once as `static constexpr` right before the definition of `submissions_queue_`.

src/OrbitVulkanLayer/SubmissionTracker.h
```diff
@@ -525,18 +527,28 @@ class SubmissionTracker : public VulkanLayerProducer::CaptureStatusListener {
...
   }
 
   // This method is responsible for retrieving all the timestamps for the "completed" submissions,
-  // for transforming the information of those submissions (in particlular about the command buffers
+  // for transforming the information of those submissions (in particular about the command buffers
   // and debug markers) into the `GpuQueueSubmission` proto, and for sending it to the
-  // `VulkanLayerProducer`. We consider a submission to be "completed" if its last command buffer
-  // timestamp is ready (see `PullCompletedSubmissions`).
+  // `VulkanLayerProducer`. We consider a submission to be "completed" all timestamps that are
+  // associated with this submission are ready.
```
dpallotti, line 547: There something missing in this sentence, probably "when".

src/OrbitVulkanLayer/SubmissionTracker.h
```diff
@@ -525,18 +527,28 @@ class SubmissionTracker : public VulkanLayerProducer::CaptureStatusListener {
...
+  // We maintain a priority queue (for every `VkQueue`) `submissions_queue_` and process submissions
+  // with the oldest CPU timestamp, until we encounter the first "incomplete" submission.
+  // This way, we ensure, that we will send the submission information per queue in ordered by the
```
dpallotti, line 550:
```suggestion
  // This way, we ensure that we will send the submission information per queue in ordered by the
```

dpallotti, line 550: "in order" or "ordered"

src/OrbitVulkanLayer/SubmissionTracker.h
```diff
@@ -1029,7 +985,10 @@ class SubmissionTracker : public VulkanLayerProducer::CaptureStatusListener {
   absl::flat_hash_map<VkCommandBuffer, VkDevice> command_buffer_to_device_;
 
   absl::flat_hash_map<VkCommandBuffer, CommandBufferState> command_buffer_to_state_;
-  absl::flat_hash_map<VkQueue, std::vector<QueueSubmission>> queue_to_submissions_;
+  absl::flat_hash_map<VkQueue,
+                      std::priority_queue<QueueSubmission, std::vector<QueueSubmission>,
+                                          std::function<bool(QueueSubmission, QueueSubmission)>>>
+      submissions_queue_;
```
dpallotti, line 991: Do you want the singular here? These are multiple priority queues. Or consider expanding the name to something like `queue_to_submission_priority_queue_`.

src/OrbitVulkanLayer/SubmissionTracker.h
dpallotti: > I think this may only happen if the producer actually dies

What do you mean with the producer dying? I guess that if the producer crashes the whole game crashes, as it's the same process.

Either way, I'm not really sure about the conditions on `vulkan_layer_producer_` in `SubmissionTracker`. I guess we kept it generic and support both cases of set and not set. In particular because `vulkan_layer_producer_` is set on `OnCreateInstance` and removed on `OnDestroyInstance`.

src/OrbitVulkanLayer/SubmissionTracker.h
dpallotti: Then you might (maybe) even be able to just write `submissions_queue_.emplace(queue, {kTheNameYouGaveTheComparator}`.

dpallotti (approved):

src/OrbitVulkanLayer/SubmissionTracker.h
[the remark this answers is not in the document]
ronaldfw: The call with all the arguments is also very long (in number of characters). So maybe just leave as-is.

src/OrbitVulkanLayer/SubmissionTracker.h
ronaldfw: If my understanding is correct, then even jobs that extend over a small number of frames might cause issues, but I don't know how common this is. 

For instance, you mentioned you have seen single calls to vkQueuePresent (because of what it has to do in CompleteSubmits) of around ~33ms for all the command buffers / debug markers in a single frame. If these pile up even over only a small number of frames (>= 2) because a single job is blocking progress on the queue, this will introduce quite some hiccups. 

The ~33ms time indicates that we would probably want to distribute the work done in CompleteSubmits more rather than less. This is probably not practical and doing it in vkQueuePresent is probably what we'll keep doing. But we shouldn't make the situation worse. 

So, unless any of what I'm saying here is wrong, I'd be in favor of keeping the priority queue per Vulkan queue.

florian-kuebler: First, I remeasured without instrumenting the "GetTimestampFunctions", in which case it was ~8ms (still pretty bad).

In general, my understanding is, that it would be still fine. We would end up in sending those events the amount of frames later, but we would also exit pretty early. There might be a problem, as we temporary might need more timer indices as before.

I would suggest to keep it as it is.

src/OrbitVulkanLayer/SubmissionTracker.h
florian-kuebler: Nice, done!

src/OrbitVulkanLayer/SubmissionTracker.h
florian-kuebler: Yes, thanks!

src/OrbitVulkanLayer/SubmissionTracker.h
florian-kuebler: Done.

florian-kuebler: Done.

src/OrbitVulkanLayer/SubmissionTracker.h
florian-kuebler: Renamed it as you suggested.

src/OrbitVulkanLayer/SubmissionTracker.h
florian-kuebler: This should be however, in a different PR.

florian-kuebler (commented):
Thanks for the comments.

src/OrbitVulkanLayer/SubmissionTracker.h
ronaldfw: > In general, my understanding is, that it would be still fine. We would end up in sending those events the amount of frames later, but we would also exit pretty early.

I'm worried about the case where a lot of submissions (that all finish quickly) would pile up while waiting for the topmost job to finish (which causes an early exit in CompleteSubmits). These all cause a large spike once that one job clears.

florian-kuebler: Yeah that might happen.

src/OrbitVulkanLayer/SubmissionTracker.h
```diff
@@ -1029,7 +980,16 @@ class SubmissionTracker : public VulkanLayerProducer::CaptureStatusListener {
   absl::flat_hash_map<VkCommandBuffer, VkDevice> command_buffer_to_device_;
 
   absl::flat_hash_map<VkCommandBuffer, CommandBufferState> command_buffer_to_state_;
-  absl::flat_hash_map<VkQueue, std::vector<QueueSubmission>> queue_to_submissions_;
+
+  static constexpr auto kPreSubmissionCPUTimestampComparator =
```
dpallotti, line 984: Prefer "Cpu" to "CPU" as per styleguide's "prefer to capitalize abbreviations as single words, e.g., StartRpc() rather than StartRPC()".

florian-kuebler: Done

PR #2041, opened 2021-03-09
Add E2E test for the source code view

beckerhe:
It covers the basic workflow:
- Load Symbols
- Click Go To Source an a function
- Tell Orbit where the sources can be found
- Show source code


contrib/automation_tests/orbit_source_code_view.py
```diff
@@ -0,0 +1,49 @@
+"""
+Copyright (c) 2020 The Orbit Authors. All rights reserved.
```
danielfenner, line 2: 2021

beckerhe: Done. Thanks!

contrib/automation_tests/test_cases/symbols_tab.py
```diff
@@ -153,3 +154,100 @@ def _try_verify_functions_are_hooked(self):
...
+
+        logging.info("File Edit was found. Entering file path...")
+        file_edit = self.find_control(
+            control_type="Edit", name="File name:", parent=file_open_dialog, recurse=True)
+        file_edit.set_focus()
+        file_edit.set_edit_text('')
+
+        # The environment variable allows to overwrite the path easily for local testing
+        file_path = os.getenv(
+            "ORBIT_SOURCE_CODE_VIEW_E2E_TEST_FILE_PATH", "C:\\build\\scratch\\test\\hello_ggp\\src\\main.c")
```
reichlfl, line 222: Could you use a relative path instead and keep the file next to the test itself so you wouldn't have to configure the test, or is that problematic due to open source? This is quite opaque to devs who want to run the test locally - wouldn't a command line parameter (using absl flags) a better option if a relative path can't be used?

How does the file get to this location on the testing machines? It's not mentioned in the comments for the test.

reichlfl (approved):
Thanks for splitting the method! I'd prefer to get rid of the absolute path, but if this is not possible, LGTM as is.

contrib/automation_tests/test_cases/symbols_tab.py
beckerhe: Yes, several points here:

1. Yes, we can't check-in this file into our repo since it's not open source, which means we need to download it from an internal bucket to the E2E test machine. And even if we could, this would mean we need to package this random C-file with the official distribution of Orbit which sounds not ideal to me.
2. Yes a command line parameter does work as well. I will add it.
3. I will improve the documentation.

reichlfl: I'd argue that the random c-file is test data and should, as such, be distributed in a data-directory with the tests. But anyways, doesn't matter since we can't do that. In that case the configurable path solution sounds good.

ronaldfw: We could go through the process of open sourcing hello_ggp, but it might not be worth the effort. But wanted to mention that the option is there.

beckerhe: Yes, good point. I also thought about this and concluded that it's not worth it. I assume the whole process will take several weeks with at least a dev-week of work.

reichlfl: Agree - we may want to keep this in mind if we end up using more of the source code across tests, but I can't see that happening, I don't think it's worth it for a single test.

beckerhe:
Status: I'm currently running another round of tests on the E2E test infrastructure with the latest version of the scripts. If that goes well, I will rebase and merge!

PR #2042, opened 2021-03-09
Fix ConnectToStadiaWidget scaling

antonrohr:
The label of the radiobutton of ConnectToStadiaWidget used to be cut off
in high dpi scaling modes. This change fixes that by detaching the
radiobutton right after the first draw instead of during construction.

To fix http://b/177500652

Before at 175% scaling:
175% before

After at 175% scaling:
175% after

Proof that no scaling still works:

Linux still looks normal
linux

### DIFF
--- a/src/OrbitQt/ConnectToStadiaWidget.cpp
+++ b/src/OrbitQt/ConnectToStadiaWidget.cpp
@@ -59,8 +59,6 @@ ConnectToStadiaWidget::ConnectToStadiaWidget(QWidget* parent)
   ui_->setupUi(this);
   ui_->instancesTableOverlay->raise();
 
-  DetachRadioButton();
-
   QSettings settings;
   if (settings.contains(kRememberChosenInstance)) {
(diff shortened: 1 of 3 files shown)


src/OrbitQt/ConnectToStadiaWidget.cpp
```diff
@@ -435,4 +434,9 @@ void ConnectToStadiaWidget::TrySelectRememberedInstance() {
   }
 }
 
+void ConnectToStadiaWidget::showEvent(QShowEvent* event) {
+  QWidget::showEvent(event);
+  DetachRadioButton();
```
beckerhe, line 439: This line would benefit from a short explanatory comment. Otherwise this might get removed in a year or so, because nobody remembers.

beckerhe (approved):
LGTM! Thanks for fixing this!

src/OrbitQt/ConnectToStadiaWidget.cpp
antonrohr: done

PR #2043, opened 2021-03-10
New Manual Instrumentation Implementation

pierricgimmig:
Reimplement manual instrumentation using a side channel to send data 
directly from the target process to OrbitService. The old uprobe-based
implementation is still supported but will be deprecated. The new approach
yields massive performance gains (tested with >300K events/second).

Here is an overview of the new architecture:
image

This is a big PR, but a lot of code is boiler plate. Here are the components in order of importance:

// Add LockFreeApiEventProducer
src/Api/Orbit.cpp
src/Api/include/Api/LockFreeApiEventProducer.h

// Add ApiEvent proto
src/ClientProtos/capture_data.proto
src/GrpcProtos/capture.proto
src/LinuxTracing/LinuxTracingIntegrationTest.cpp
src/Service/ProducerEventProcessor.cpp

// Add ApiEventProcessor
src/OrbitCaptureClient/ApiEventProcessor.cpp
src/OrbitCaptureClient/ApiEventProcessorTest.cpp
src/OrbitCaptureClient/CMakeLists.txt
src/OrbitCaptureClient/CaptureEventProcessor.cpp
src/OrbitCaptureClient/include/OrbitCaptureClient/ApiEventProcessor.h
src/OrbitCaptureClient/include/OrbitCaptureClient/CaptureEventProcessor.h

// UI
src/OrbitGl/ThreadTrack.cpp
src/OrbitGl/TimeGraph.cpp
src/OrbitGl/TimeGraph.h

// Add ApiInterface target for introspection
src/Api/CMakeLists.txt
src/OrbitBase/CMakeLists.txt

// Build
CMakeLists.txt
kokoro/builds/build.sh

// Adapt OrbitTest to use the new manual instrumentation API
src/OrbitTest/CMakeLists.txt
src/OrbitTest/OrbitTest.cpp

// Use new C API
src/Api/include/Api/Orbit.h
src/LinuxTracing/TracerThread.cpp
src/OrbitBase/OrbitApiTest.cpp
src/OrbitBase/Tracing.cpp
src/OrbitBase/include/OrbitBase/Tracing.h

// Maintain support for previous manual instrumentation (uprobe based)
src/OrbitGl/ManualInstrumentationManager.cpp
src/OrbitGl/ManualInstrumentationManager.h

// Remove previous version of Orbit.h
src/Orbit.h 

Note: The use of the new ScopeTree data structure to merge manual and 
dynamic instrumentation data will come in a subsequent PR.


src/Api/CMakeLists.txt
```diff
@@ -29,3 +31,14 @@ target_link_libraries(Api PUBLIC
         OrbitBase
         OrbitProducer
         ProducerSideChannel)
+endif()
+
+add_library(ApiInterface INTERFACE)
```
dpallotti, line 36: What's the difference between this and `Api` above and what's the reason for having both? It's strange that they share files. It's possible that a common target can be extracted (for example, `Api` above can depend on this one).

src/Api/include/Api/LockFreeApiEventProducer.h
```diff
@@ -0,0 +1,46 @@
+// Copyright (c) 2021 The Orbit Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#ifndef ORBIT_API_LOCK_FREE_API_EVENT_PRODUCER_H_
```
dpallotti, line 5: Our include guards reflect the module/directory name, not the namespace, so drop the `ORBIT_` prefix.

src/Api/Orbit.cpp
```diff
@@ -6,21 +6,21 @@
...
 #include "OrbitBase/ThreadUtils.h"
 
-// Temporary dummy function, proper implementation is in next PR.
-// See https://github.com/google/orbit/pull/1702 for fully working prototype.
-static void EnqueueIntermediateEvent(const orbit_api::ApiEvent& event) { (void)event; }
-
 static void EnqueueApiEvent(orbit_api::EventType type, const char* name = nullptr,
                             uint64_t data = 0, orbit_api_color color = kOrbitColorAuto) {
-  uint64_t timestamp_ns = orbit_base::CaptureTimestampNs();
+  static orbit_api::LockFreeApiEventProducer producer;
```
dpallotti, line 15: This statement is associated with quite some initialization work. If I get it correctly this happens right after the .so gets loaded, so it's already the case that the first call to `EnqueueApiEvent` happens together with initialization work, but I want to confirm that this is the case and it's intended.

src/Api/include/Api/LockFreeApiEventProducer.h
```diff
@@ -0,0 +1,46 @@
...
+#define ORBIT_API_LOCK_FREE_API_EVENT_PRODUCER_H_
+
+#include "Api/EncodedEvent.h"
+#include "OrbitProducer/LockFreeBufferCaptureEventProducer.h"
+#include "ProducerSideChannel/ProducerSideChannel.h"
+
+namespace orbit_api {
+
+// This class is used to enqueue orbit_api::ApiEvent events from multiple threads and relay them to
+// OrbitService in the form of orbit_grpc_protos::ProducerCaptureEventCaptureEvent events.
```
dpallotti, line 15: Aren't they called `ApiEvent`?

src/Api/include/Api/LockFreeApiEventProducer.h
```diff
@@ -0,0 +1,46 @@
...
+ public:
+  LockFreeApiEventProducer() {
+    BuildAndStart(orbit_producer_side_channel::CreateProducerSideChannel());
+  }
+
+  ~LockFreeApiEventProducer() { ShutdownAndWait(); }
+
+ protected:
+  [[nodiscard]] virtual orbit_grpc_protos::ProducerCaptureEvent* TranslateIntermediateEvent(
+      orbit_api::ApiEvent&& raw_api_event, google::protobuf::Arena* arena) override {
```
dpallotti, line 27: Maybe rename `raw_api_event` to just `api_event`.

src/GrpcProtos/capture.proto
```diff
@@ -298,6 +314,7 @@ message ClientCaptureEvent {
     // it is going to go away in the future when we switch to
     // frame-pointer based unwinding.
     AddressInfo address_info = 16;
+    ApiEvent api_event = 24;
```
dpallotti, line 317: As per the comment at the beginning of the `oneof`, this should probably be in the 1-15 range as it's high frequency. Unless we already know that the field number will likely change, in which case we don't want to take away a field number in the 1-15 range.

src/GrpcProtos/capture.proto
```diff
@@ -323,6 +340,7 @@ message ProducerCaptureEvent {
     // numbers starting with 16.
     //
     // Please keep these alphabetically ordered.
+    ApiEvent api_event = 23;
```
dpallotti, line 343: Same as above.

src/GrpcProtos/capture.proto
```diff
@@ -95,6 +95,22 @@ message IntrospectionScope {
...
 
+message ApiEvent {
+  int32 pid = 1;
+  int32 tid = 2;
+  uint64 timestamp_ns = 3;
+
+  // The fields below are used to encode orbit_api::EncodedEvent events. This is
+  // done mainly for the performance gain of not having to allocate memory out
+  // of the Arena for strings.
+  fixed64 r0 = 4;
```
dpallotti, line 106: What does "r" stand for? Raw? I would consider expanding these names even if we know they just stand for row data.

src/OrbitBase/CMakeLists.txt
```diff
@@ -71,6 +71,7 @@ target_sources(OrbitBase PRIVATE
 endif()
 
 target_link_libraries(OrbitBase PUBLIC
+        ApiInterface
```
dpallotti, line 74: Such a dependency for `OrbitBase` is a bit worrisome. What's the reason for it? I assume because of `Tracing.cpp` (i.e., introspection). But in that case I think that `Tracing.cpp` should be moved to its own module, that modules that use introspection can explicitly depend on. Or introspection and manual instrumentation should just be independent.

src/OrbitTest/OrbitTest.cpp
```diff
@@ -15,16 +15,26 @@
...
-#include "../Orbit.h"
 #include "OrbitBase/ThreadUtils.h"
 #include "absl/strings/str_format.h"
 
+// OrbitBase defines an implementation of the Orbit API for introspection. To prevent symbol clash
+// with the header-only use of the API, we wrap the include of Orbit.h in a namespace. This is only
+// needed for internal projects that link with OrbitBase, basically only OrbitTest.
+namespace orbit_api_wrapper {
+#include "Api/Orbit.h"
+}
```
dpallotti, line 26: This is quite weird. Is there no alternative? Do we even need the same interface for manual instrumentation and for introspection?

src/OrbitGl/TimeGraph.cpp
```diff
@@ -299,13 +305,41 @@ void TimeGraph::ProcessOrbitFunctionTimer(FunctionInfo::OrbitType type,
...
+    case orbit_api::kScopeStopAsync:
+      manual_instrumentation_manager_->ProcessAsyncTimer(timer_info);
+      break;
+
+    case orbit_api::kTrackInt:
+    case orbit_api::kTrackInt64:
+    case orbit_api::kTrackUint:
+    case orbit_api::kTrackUint64:
+    case orbit_api::kTrackFloat:
+    case orbit_api::kTrackDouble:
```
dpallotti, line 334: Use `[[fallthrough]]` where applicable.

src/OrbitGl/TimeGraph.cpp
```diff
@@ -284,6 +284,12 @@ void TimeGraph::ProcessTimer(const TimerInfo& timer_info, const FunctionInfo* fu
       track->OnTimer(timer_info);
       break;
     }
+    case TimerInfo::kApiEvent: {
+      if (timer_info.type() == TimerInfo::kApiEvent) {
```
dpallotti, line 288: I assume this `if` is redundant.

src/OrbitCaptureClient/ApiEventProcessor.cpp
```diff
@@ -0,0 +1,128 @@
...
+#include "OrbitCaptureClient/ApiEventProcessor.h"
+
+#include "OrbitBase/Logging.h"
+
+using orbit_client_protos::TimerInfo;
+using orbit_grpc_protos::ApiEvent;
+
+ApiEventProcessor::ApiEventProcessor(CaptureListener* listener) : capture_listener_(listener) {}
+
+static inline TimerInfo TimerInfoFromEncodedEvent(const orbit_api::EncodedEvent& encoded_event,
```
dpallotti, line 14: What is `inline` for here?

src/OrbitCaptureClient/include/OrbitCaptureClient/ApiEventProcessor.h
```diff
@@ -0,0 +1,33 @@
...
+#ifndef ORBIT_CAPTURE_CLIENT_API_EVENT_PROCESSOR_H_
+#define ORBIT_CAPTURE_CLIENT_API_EVENT_PROCESSOR_H_
+
+#include <absl/container/flat_hash_map.h>
+
+#include "Api/EncodedEvent.h"
+#include "OrbitCaptureClient/CaptureListener.h"
+#include "capture.pb.h"
+
+class ApiEventProcessor {
```
dpallotti, line 14: This file needs the namespace.

src/OrbitCaptureClient/include/OrbitCaptureClient/ApiEventProcessor.h
```diff
@@ -0,0 +1,33 @@
...
+ public:
+  ApiEventProcessor(CaptureListener* listener);
```
dpallotti, line 16: `explicit`.

src/Api/Orbit.cpp
florian-kuebler, line 15: Just to clarify that I got it right here:
This will compiled into orbit.so.
The game will load orbit.so as soon as it is available.
In this case we are creating a static producer, which thus will be destructed on unload of orbit.so (or game exit).

Is there any issue with this?

src/OrbitCaptureClient/ApiEventProcessor.cpp
```diff
@@ -0,0 +1,128 @@
...
+// found in the LICENSE file.
+
+#include "OrbitCaptureClient/ApiEventProcessor.h"
+
+#include "OrbitBase/Logging.h"
+
+using orbit_client_protos::TimerInfo;
+using orbit_grpc_protos::ApiEvent;
+
+ApiEventProcessor::ApiEventProcessor(CaptureListener* listener) : capture_listener_(listener) {}
```
dpallotti, line 12: `CHECK` that `listener` is not `nullptr`.

src/Api/include/Api/LockFreeApiEventProducer.h
```diff
@@ -0,0 +1,46 @@
...
+    auto* capture_event =
+        google::protobuf::Arena::CreateMessage<orbit_grpc_protos::ProducerCaptureEvent>(arena);
+    auto* api_event = capture_event->mutable_api_event();
```
florian-kuebler, line 30: Not sure if this is already fix, as it is basically the first time I see it. But I find "ApiEvent" a bit generic. Isn't it in the end a "ManualInstrumentationEvent"?

src/OrbitCaptureClient/include/OrbitCaptureClient/ApiEventProcessor.h
```diff
@@ -0,0 +1,33 @@
...
+ private:
+  void ProcessApiEvent(const orbit_api::ApiEvent& api_event);
+  void ProcessStartEvent(const orbit_api::ApiEvent& api_event);
+  void ProcessStopEvent(const orbit_api::ApiEvent& api_event);
+  void ProcessAsyncStartEvent(const orbit_api::ApiEvent& api_event);
+  void ProcessAsyncStopEvent(const orbit_api::ApiEvent& api_event);
+  void ProcessTrackingEvent(const orbit_api::ApiEvent& api_event);
+
+ private:
+  CaptureListener* capture_listener_ = nullptr;
```
dpallotti, line 28: Have you considered aligning this class to the model of `GpuQueueSubmissionProcessor` and returning `TimerInfo`s instead of passing the `CaptureListener` to this class? It's weird to me that `GpuQueueSubmissionProcessor` and this class follow opposite models when there's a clear parallel between the two.

src/Api/include/Api/LockFreeApiEventProducer.h
```diff
@@ -0,0 +1,46 @@
...
+    api_event->set_timestamp_ns(raw_api_event.timestamp_ns);
+    api_event->set_pid(raw_api_event.pid);
+    api_event->set_tid(raw_api_event.tid);
+    api_event->set_r0(raw_api_event.encoded_event.args[0]);
```
florian-kuebler, line 34: Are there any plans for the (near future) to move from raw data to some more structured data?

src/OrbitCaptureClient/CaptureEventProcessor.cpp
```diff
@@ -82,6 +82,10 @@ void CaptureEventProcessor::ProcessEvent(const ClientCaptureEvent& event) {
     case ClientCaptureEvent::kSystemMemoryUsage:
       // TODO (http://b/179000848): Process the system memory usage information.
       break;
+    case ClientCaptureEvent::kApiEvent: {
```
dpallotti, line 85: `CaptureEventProcessorTest.cpp` should probably be expanded accordingly. Refer to Florian and Dimitry, they worked on it the most.

src/OrbitCaptureClient/ApiEventProcessor.cpp
```diff
@@ -0,0 +1,128 @@
...
+      break;
+    case orbit_api::kScopeStopAsync:
+      ProcessAsyncStopEvent(api_event);
+      break;
+    case orbit_api::kTrackInt:
+    case orbit_api::kTrackInt64:
+    case orbit_api::kTrackUint:
+    case orbit_api::kTrackUint64:
+    case orbit_api::kTrackFloat:
+    case orbit_api::kTrackDouble:
```
dpallotti, line 69: Use `[[fallthrough]]`.

src/OrbitCaptureClient/ApiEventProcessor.cpp
```diff
@@ -0,0 +1,128 @@
...
+  if (event_stack.empty()) {
+    // We received a stop event with no matching start event, which is possible if the capture was
+    // started between the event's start and stop times.
+    return;
+  }
+
+  const orbit_api::ApiEvent& start_event = event_stack.back();
+  TimerInfo timer_info = TimerInfoFromEncodedEvent(
+      start_event.encoded_event, start_event.timestamp_ns, stop_event.timestamp_ns, stop_event.pid,
+      stop_event.tid, /*depth*/ event_stack.size() - 1);
```
dpallotti, line 96:
```suggestion
      stop_event.tid, /*depth=*/ event_stack.size() - 1);
```

src/OrbitCaptureClient/ApiEventProcessor.cpp
```diff
@@ -0,0 +1,128 @@
...
+  if (asynchronous_events_by_id_.count(event_id) == 0) {
+    // We received a stop event with no matching start event, which is possible if the capture was
+    // started between the event's start and stop times.
+    return;
+  }
+
+  orbit_api::ApiEvent& start_event = asynchronous_events_by_id_[event_id];
+  TimerInfo timer_info = TimerInfoFromEncodedEvent(
+      start_event.encoded_event, start_event.timestamp_ns, stop_event.timestamp_ns, stop_event.pid,
+      stop_event.tid, /*depth*/ 0);
```
dpallotti, line 117:
```suggestion
      stop_event.tid, /*depth=*/ 0);
```

src/OrbitBase/Tracing.cpp
```diff
@@ -79,16 +79,14 @@ static std::vector<TracingScope>& GetThreadLocalScopes() {
   return thread_local_scopes;
 }
 
-namespace orbit_api {
-
-void Start(const char* name, orbit::Color color) {
+void orbit_api_start(const char* name, orbit_api_color color) {
```
florian-kuebler, line 82: Is there a special reason for the snake_case syntax? Feels a bit inconsistent. At least I would expect a comment.

src/OrbitCaptureClient/ApiEventProcessor.cpp
dpallotti, line 117: Is the depth really meant to be always zero? Can't the same id be used to nest async scopes?

src/OrbitCaptureClient/include/OrbitCaptureClient/ApiEventProcessor.h
florian-kuebler, line 14: Could you add a brief description of the class:
https://google.github.io/styleguide/cppguide.html#Class_Comments

src/OrbitGl/ManualInstrumentationManager.h
```diff
@@ -28,6 +28,7 @@ class ManualInstrumentationManager {
   void AddAsyncTimerListener(AsyncTimerInfoListener* listener);
   void RemoveAsyncTimerListener(AsyncTimerInfoListener* listener);
   void ProcessAsyncTimer(const orbit_client_protos::TimerInfo& timer_info);
+  void ProcessAsyncTimerDeprecated(const orbit_client_protos::TimerInfo& timer_info);
```
dpallotti, line 31: Move this up one line or move down the definition in the .cpp to have the same order.

src/OrbitCaptureClient/CaptureEventProcessor.cpp
```diff
@@ -82,6 +82,10 @@ void CaptureEventProcessor::ProcessEvent(const ClientCaptureEvent& event) {
...
+      api_event_processor_.ProcessApiEvent(event.api_event());
```
florian-kuebler, line 86: Could you add one (or more) tests in CaptureEventProcessorTest?

dpallotti (commented):
I added a couple of reviewers as it's a big change and it touches many parts.

I've mentioned tests for the additions in `CaptureEventProcessor` but have you thought about unit tests for other parts?

src/OrbitCaptureClient/ApiEventProcessor.cpp
```diff
@@ -0,0 +1,128 @@
...
+    case orbit_api::kTrackInt64:
+    case orbit_api::kTrackUint:
+    case orbit_api::kTrackUint64:
+    case orbit_api::kTrackFloat:
+    case orbit_api::kTrackDouble:
+    case orbit_api::kString:
+      ProcessTrackingEvent(api_event);
+      break;
+    case orbit_api::kNone:
+      ERROR("Received event of type orbit_api::kNone");
```
florian-kuebler, line 74: Can this actually happen? Otherwise, consider using a CHECK.

src/OrbitGl/TimeGraph.cpp
```diff
@@ -299,13 +305,41 @@ void TimeGraph::ProcessOrbitFunctionTimer(FunctionInfo::OrbitType type,
...
+    case orbit_api::kTrackInt64:
+    case orbit_api::kTrackUint:
+    case orbit_api::kTrackUint64:
+    case orbit_api::kTrackFloat:
+    case orbit_api::kTrackDouble:
+    case orbit_api::kString:
+      ProcessValueTrackingTimer(timer_info);
+      break;
+    default:
+      ERROR("Unhandled API event");
```
florian-kuebler, line 339: UNREACHABLE?

florian-kuebler (commented):
Thanks for that change. In general it looks good. However, I'd love to see tests for this. Not sure if it needs to be in this PR, but should be added soon.

src/Api/Orbit.cpp
dimitry-: since this is inside the function it is going to be initialized on the first call.

src/OrbitBase/CMakeLists.txt
pierricgimmig: This is an improvement over what we had before (including the header user "../../src/Orbit.h" but I agree that introspection should probably be in its own module. Added b/182473973

src/Api/include/Api/LockFreeApiEventProducer.h
pierricgimmig: There is significant performance overhead associated with "string" in protobufs as it doesn't use the Arena to allocate memory. That was verified through profiling. There is a bigger plan to maybe move away from protobufs altogether in favor of (maybe) flatbuffers. In that case, we would have proper structured data.

src/GrpcProtos/capture.proto
dimitry-: have you tried using bytes in please of string? does it improve the performance?

src/Service/ProducerEventProcessor.cpp
```diff
@@ -318,6 +320,12 @@ void ProducerEventProcessorImpl::ProcessSystemMemoryUsage(SystemMemoryUsage* sys
   capture_event_buffer_->AddEvent(std::move(event));
 }
 
+void ProducerEventProcessorImpl::ProcessApiEvent(ApiEvent* api_event) {
+  ClientCaptureEvent event;
+  *event.mutable_api_event() = std::move(*api_event);
```
dimitry-, line 325: we probably want to convert it to variable-encoding friendly proto for the service<->client communication

src/Api/CMakeLists.txt
pierricgimmig: Fixed, Api depends on ApiInterface now. The "ApiInterface" is used if a project wants to define its own implementation, this is done so that introspection shares the same interface as the manual instrumentation API we ship to users.

src/Api/include/Api/LockFreeApiEventProducer.h
pierricgimmig: Done, thanks.

src/Api/include/Api/LockFreeApiEventProducer.h
pierricgimmig: Done, thanks.

src/Api/include/Api/LockFreeApiEventProducer.h
pierricgimmig: I want to make it clear that the intermediate events are "raw" events and not protos. Also api_event is used for another variable below.

src/OrbitCaptureClient/ApiEventProcessor.cpp
pierricgimmig: It's not required by the style guide. "A common exception is consecutive case labels without intervening code, in which case no annotation is needed."

src/OrbitCaptureClient/ApiEventProcessor.cpp
pierricgimmig: thanks

src/OrbitGl/ManualInstrumentationManager.h
pierricgimmig: done, thanks.

src/OrbitBase/Tracing.cpp
pierricgimmig: The reason is to signal that those are "C" functions, not "C++" anymore.

src/OrbitCaptureClient/ApiEventProcessor.cpp
pierricgimmig: Thanks for catching this, I had made a mental note to review this code, it now aborts in the default case (I removed the specific kNone case).

src/OrbitGl/TimeGraph.cpp
pierricgimmig: done, thanks

src/OrbitCaptureClient/ApiEventProcessor.cpp
```diff
@@ -0,0 +1,128 @@
...
+      break;
+    default:
```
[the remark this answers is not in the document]
pierricgimmig: Yep, I had made a mental note to fix this, but didn't, thank you and Florian for pointing this out, fixed now with UNREACHABLE.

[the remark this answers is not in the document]
dimitry-: UNREACHABLE is generating runtime abort if someone adds a new value, whereas omitting default results in compile-time error. The difference is that former could be missed because it needs for new event happen in a test. Latter is much harder to miss.

src/OrbitCaptureClient/ApiEventProcessor.cpp
pierricgimmig: Yes this is meant to be zero. Async scopes are not nested, they end up being stacked by the UI but there is no hierarchical relationships between async events.

src/OrbitCaptureClient/ApiEventProcessor.cpp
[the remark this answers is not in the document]
pierricgimmig: yes, good point, changed it to omit default, thanks for the precision.

src/GrpcProtos/capture.proto
pierricgimmig: Yeah it stands for registers, might not be the best name, but it maps directly to the previous uprobe-based implementation of manual instrumentation. I modified the comment. Concerning the use of bytes, it would improve performance over strings, yes, but not over fixed64. I'd keep the fixed64 fields until we deprecate the old implementation. For now it's quite convenient to have 6 separate 64 bit values to emulate registers. Concerning performance, as you know, there is a bigger investigation going on as to how we can reduce OrbitService's CPU usage. It seems like the constant decoding/encoding of protobufs is taking the majority of the time on the service side, this will probably be my main focus after the manual instrumentation change.

src/Api/Orbit.cpp
pierricgimmig: That's right. The first api call also triggers the loading of liborbit.so. We could possibly add an initialization method to explicitly trigger the loading of the .so and the initialization of the side channel if need be. Currently we have no way to unload liborbit.so manually, so having an "init" without a "deinit" is a bit weird, but it's something I'd like to add.

pierricgimmig: Yeah that's right, it should be very lightweight when we are not capturing, I think it's fine. But as mentioned above, one thing I'd like to add an api method to unload liborbit.so manually, although it might be tricky to implement.

src/GrpcProtos/capture.proto
pierricgimmig: cool, I made the change, thanks.

src/GrpcProtos/capture.proto
pierricgimmig: done, thanks.

src/OrbitCaptureClient/ApiEventProcessor.cpp
pierricgimmig: Done.

src/OrbitCaptureClient/ApiEventProcessor.cpp
pierricgimmig: As a hint to the compiler to inline the function. Is static and inline mutually exclusive?

src/OrbitCaptureClient/include/OrbitCaptureClient/ApiEventProcessor.h
pierricgimmig: Not all calls to ProcessApiEvent could return a TimerInfo, "start" events are cached until a corresponding "stop" event arrives, only then can a TimerInfo be created and passed to a listener.

src/OrbitGl/TimeGraph.cpp
pierricgimmig: it sure is :-)

src/OrbitGl/TimeGraph.cpp
pierricgimmig: I think it's not needed here. "A common exception is consecutive case labels without intervening code, in which case no annotation is needed" https://google.github.io/styleguide/cppguide.html#Loops_and_Switch_Statements

src/OrbitTest/OrbitTest.cpp
pierricgimmig: Yeah it is definitely weird, but it's only for this one internal test project. Another option would be to not link with OrbitBase, but that's annoying because we're using ThreadPools and Logging. We could duplicate the interface, but then we'd maintain essentially 2 copies of the same file. I also find it cool that we're dogfooding the API.

src/OrbitCaptureClient/include/OrbitCaptureClient/ApiEventProcessor.h
pierricgimmig: yep, done.

src/OrbitCaptureClient/include/OrbitCaptureClient/ApiEventProcessor.h
pierricgimmig: which namespace?

pierricgimmig:
@dpallotti , @florian-kuebler , @dimitry- , thanks a lot for your comments, and thanks for bearing with the big PR. I addressed most of the issues, but I'm still writing the tests, will have those in tomorrow.

src/GrpcProtos/capture.proto
```diff
@@ -323,6 +341,7 @@ message ProducerCaptureEvent {
     // numbers starting with 16.
     //
     // Please keep these alphabetically ordered.
+    ApiEvent api_event = 10;
```
dimitry-, line 344: I do not think it matters, we are saving 1 byte by using low number but on the other hand the structure of the proto wasting a lot of them, not just absence of variant encoding for most fields we also pass unnecessary padding over the wire. I think this is one case I would argue this does not make sense. We better off saving low number for compact protos.

src/Api/Orbit.cpp
florian-kuebler: Okay, thanks for clarification

src/Api/include/Api/LockFreeApiEventProducer.h
dpallotti: You're right, I had gotten this wrong.

src/GrpcProtos/capture.proto
dpallotti: @dimitry- `bytes` is internally a heap-allocated `std::string` just like `string`, the only difference is that `string` requires `UTF-8` encoding.

src/OrbitCaptureClient/ApiEventProcessor.cpp
dpallotti: No no, but we basically never use it just to hint the compiler to try to inline (I guess that for a `static` function the compiler will probably inline anyway?). But I'm totally fine with having it if that's the intent.

src/OrbitCaptureClient/ApiEventProcessor.cpp
dpallotti: Ok ok, makes sense, I had previous reviews asking for `[[fallthrough]]` in such a case so it seems we are not fully aligned. It's preference I guess.

src/OrbitCaptureClient/include/OrbitCaptureClient/ApiEventProcessor.h
dpallotti: I would have said `orbit_capture_client`, but it seems that the whole module lacks the namespace so I guess it's fine as it is.

src/OrbitCaptureClient/include/OrbitCaptureClient/ApiEventProcessor.h
```diff
@@ -0,0 +1,38 @@
...
+#ifndef ORBIT_CAPTURE_CLIENT_API_EVENT_PROCESSOR_H_
+#define ORBIT_CAPTURE_CLIENT_API_EVENT_PROCESSOR_H_
+
+#include <absl/container/flat_hash_map.h>
+
+#include "Api/EncodedEvent.h"
+#include "OrbitCaptureClient/CaptureListener.h"
+#include "capture.pb.h"
+
+// The ApiEventProcessor is responsible processing orbit_grpc_protos::ApiEvent events and
```
dpallotti, line 14:
```suggestion
// The ApiEventProcessor is responsible for processing orbit_grpc_protos::ApiEvent events and
```

src/OrbitTest/OrbitTest.cpp
dpallotti: > We could duplicate the interface, but then we'd maintain essentially 2 copies of the same file. I also find it cool that we're dogfooding the API.

I see your points. On the other hand it seems that we have to go through various hacks in order to achieve this (such as this one or `#ifdef`s) to have only one file. And introspection and manual instrumentation work quite differently under the hood. So I'm not really sold on the fact that it's better to have a single "copy".

src/OrbitCaptureClient/include/OrbitCaptureClient/ApiEventProcessor.h
dpallotti: You could just return `std::optional<TimerInfo>`.

src/OrbitBase/CMakeLists.txt
dpallotti: It's an improvement in the sense that we don't just include a header bypassing CMake, but it's also quite worse because now we explicitly have `OrbitBase` depending on one of our modules. We should not introduce exceptions to the "rule" that `OrbitBase` shouldn't depend on other modules. In my opinion this should be addressed, creating a bug doesn't really do anything.

pierricgimmig: The dependency was already there, this is making it explicit. OrbitBase itself uses introspection btw, which make that "rule" difficult to enforce. Note that introspection is not only using Orbit.h, but also EncodedEvent.h. I made a bug because I think it's a significant change that goes out of the scope of this PR. It also ties in with the deprecation of the old API that will allow us to remove EncodedEvent.h altogether. I think it will be easier to tackle this then.

src/OrbitCaptureClient/CaptureEventProcessor.cpp
pierricgimmig: Please see new ApiEventProcessorTest.cpp

src/OrbitCaptureClient/include/OrbitCaptureClient/ApiEventProcessor.h
pierricgimmig: Done, thanks.

src/Api/include/Api/LockFreeApiEventProducer.h
pierricgimmig: I really prefer ApiEvent for its conciseness, I think it's specific enough to understand exactly what they are. The existing "IntrospectionScope" could also be considered a "ManualInstrumentationEvent" which would be confusing I think.

src/GrpcProtos/capture.proto
pierricgimmig: Ok, let's change it back then, thanks.

src/OrbitCaptureClient/include/OrbitCaptureClient/ApiEventProcessor.h
pierricgimmig: I like that ApiEventProccessor can work both in isolation and as part of a CaptureEventProcessor (this is used in ApiEventProcessorTest.cpp). I prefer keeping it as is instead of introducing std::optional which I think could be confusing. I would find it easier as a reader to look at code sites where the listener is used rather than trying to figure out why the ApiEventProcessor sometimes returns a TimerInfo and sometimes doesn't.

src/OrbitCaptureClient/CaptureEventProcessor.cpp
pierricgimmig: Added ApiEventProcessorTest.cpp which tests the processing of api events for both CaptureEventProcessor and ApiEventProcessor.

src/OrbitTest/OrbitTest.cpp
pierricgimmig: Yeah, I'm not particularly happy with this namespace trick, I'll think about an alternative. Maybe having separate interfaces for introspection and manual instrumentation, we'll see, but as mentioned I think it'll be easier to tackle this once the original manual instrumentation is deprecated.

pierricgimmig:
@dpallotti , @florian-kuebler , @dimitry- , again thanks for your help. I think I have addressed all the comments, please let me know if I missed anything. I added ApiEventProcessorTest.cpp that tests the ApiEventProcessor both in isolation and as part of a CaptureEventProcessor. Please take another look, thanks!

src/Api/include/Api/LockFreeApiEventProducer.h
florian-kuebler: Gave this another thought. I am fine with ApiEvent.

florian-kuebler (approved):
Thanks for adding the tests. In general they look nice. You could consider doing them with gmock, but this would mainly be for consistency with the other tests, you achieved all that was needed and remain readable. So I'll leave this to you.

src/OrbitBase/CMakeLists.txt
dpallotti: > OrbitBase itself uses introspection btw, which make that "rule" difficult to enforce.

If something is already using introspection in OrbitBase then ok, this goes beyond this change. I would at least add a todo referring to the bug though. But in my opinion OrbitBase should not use introspection. That goes completely against OrbitBase being a "base" component, i.e., independent of Orbit's technicalities, and it introduces unwanted dependencies.

src/OrbitCaptureClient/include/OrbitCaptureClient/ApiEventProcessor.h
dpallotti: > I like that ApiEventProccessor can work both in isolation and as part of a CaptureEventProcessor

Returning `std::optional<TimerInfo>` increases isolation because it removes the dependency on the listener, so I don't understand this point. 

> introducing std::optional which I think could be confusing. I would find it easier as a reader to look at code sites where the listener is used rather than trying to figure out why the ApiEventProcessor sometimes returns a TimerInfo and sometimes doesn't.

I'm not sure about this. Either way, the inconsistency with `GpuQueueSubmissionProcessor` is more confusing. For sure as a reader I'm wondering why one class uses one approach and the other a different one.

src/OrbitCaptureClient/CaptureEventProcessor.cpp
dpallotti: I think it would still be beneficial to have a small test in `CaptureEventProcessorTest.cpp` that covers the fact that `kApiEvent`s are handled properly. Then of course all the details of `ApiEventProcessor` are tested in isolation in `ApiEventProcessorTest.cpp`.

src/OrbitCaptureClient/CaptureEventProcessor.cpp
dpallotti: See the other comment: there should be a small test regarding `CaptureEventProcessor` with `kApiEvent` in `CaptureEventProcessorTest.cpp`.

src/OrbitTest/OrbitTest.cpp
dpallotti: Sounds good. Maybe add a todo (possibly with bug) so that we don't lose track.

src/OrbitCaptureClient/ApiEventProcessorTest.cpp
```diff
@@ -0,0 +1,238 @@
...
+    EXPECT_EQ(CountTimersOfType(orbit_api::kScopeStart), num_timers);
+    return *this;
+  }
+
+  ApiTester& CheckNumAsyncScopeTimers(size_t num_timers) {
+    EXPECT_EQ(CountTimersOfType(orbit_api::kScopeStartAsync), num_timers);
+    return *this;
+  }
+
+  ApiTester& CheckNumTrackingTimers(size_t num_timers) {
```
dpallotti, line 106: I would suggest renaming these methods from `Check...` to `Expect...`. They better hint at the fact that `EXPECT_...` is used and *not* `CHECK`, which has a different meaning.

src/OrbitCaptureClient/ApiEventProcessorTest.cpp
```diff
@@ -0,0 +1,238 @@
...
+    timers_.emplace_back(timer_info);
+  }
+  std::vector<TimerInfo> timers_;
+};
+
+// ApiTester exposes methods that mocks the Orbit API and internally creates grpc events that it
+// forwards to both a CaptureEventProcessor and a ApiEventProcessor. The test makes sure that both
+// processors forward the same information to their listener. Note that a CaptureEventProcessor owns
+// a ApiEventProcessor to which it forwards ApiEvent events. To help readability of test code, the
+// api methods return a ApiTester reference so that we can chain function calls.
```
dpallotti, line 54:
```suggestion
// ApiTester exposes methods that mocks the Orbit API and internally creates grpc events that it
// forwards to both a CaptureEventProcessor and an ApiEventProcessor. The test makes sure that both
// processors forward the same information to their listener. Note that a CaptureEventProcessor owns
// an ApiEventProcessor to which it forwards ApiEvent events. To help readability of test code, the
// api methods return an ApiTester reference so that we can chain function calls.
```

Or maybe this was on purpose. Just pointing it out.

src/OrbitCaptureClient/ApiEventProcessorTest.cpp
```diff
@@ -0,0 +1,238 @@
+// Copyright (c) 2020 The Orbit Authors. All rights reserved.
```
dpallotti, line 1: As partly mentioned by Florian, I would at least consider using gTest functionalities a bit more.
For example, `ApiEventCaptureListener` seems a good candidate for being a gMock, as you are basically counting how many times `onTimer` has been called with which arguments.
Or `ApiTester` seems to be a good candidate for using a test fixture.
I don't know if it's necessarily better (gMock has a non-negligible learning curve, although it's useful to learn it), but it would at least a more idiomatic use of gTest.

src/OrbitCaptureClient/ApiEventProcessorTest.cpp
pierricgimmig: yep thanks

src/OrbitCaptureClient/ApiEventProcessorTest.cpp
pierricgimmig: Done, thanks.

src/OrbitBase/CMakeLists.txt
dpallotti, line 74: Please add a TODO linking http://b/182473973 around here.

dpallotti (approved):

PR #2044, opened 2021-03-10
Unify behaviour of Disassembly and source code view

beckerhe:
- Current lines are now highlighted in both views
- The right sidebar is hidden when no capture was taken in both views

Furthermore there is a fix for a potential crash.


karupayun (approved):
LGTM! :)

PR #2047, opened 2021-03-10
Rename redraw / update methods, fix tooltips

reichlfl:
This cleans up the naming around NeedsUpdate / NeedsRedraw, and changes
the logic of "GetNeedsUpdate()": This now not only returns True if
explicitely requested, but also if required implicitely (in the current
case: if mouse hover needs to be evaluated).

The update behavior did not work for tooltips before: The need to update
on mouse hover was determined in `PreRender`, but this was never called
due to the early out on `GetNeedsUpdate` in the main timer.

The main change lies in the the logic behind the new `IsRedrawNeeded`,
sorry for mixing the renaming in there.

### DIFF
--- a/src/OrbitGl/TimeGraph.cpp
+++ b/src/OrbitGl/TimeGraph.cpp
@@ -83,7 +83,7 @@ void TimeGraph::ZoomAll() {
   min_time_us_ = max_time_us_ - (GNumHistorySeconds * 1000 * 1000);
   if (min_time_us_ < 0) min_time_us_ = 0;
 
-  NeedsUpdate();
+  RequestUpdatePrimitives();
 }
 
 void TimeGraph::Zoom(uint64_t min, uint64_t max) {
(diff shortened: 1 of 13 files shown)


antonrohr (approved):

reichlfl:
I've updated the PR to address b/182349451 - @antonrohr If you want to take another look, I've added the changes in subsequent commits and did not modify the original one.

pierricgimmig (approved):
Hi @reichlfl , thanks for fixing this, the code is also clearer now. LGTM

src/OrbitGl/TimeGraph.cpp
```diff
@@ -1,9 +1,8 @@
+#include "TimeGraph.h"
```
antonrohr, line 1: I assume clang-format or ide messed up somehow

antonrohr (approved):
one include mess up, everything else still LGTM!

src/OrbitGl/TimeGraph.cpp
reichlfl: Weird, probably that was due to Renaming in VS. Thanks for spotting, fixed it.

PR #2048, opened 2021-03-10
Increase timeout for symbol loading in E2E tests.

danielfenner:
Symbols can be huge so the 5 seconds default is not always enought.

bug: http://b/182354655
Test: Ran the test locally.


reichlfl (approved):
LGTM

PR #2049, opened 2021-03-10
Make zooming in Dialog's sidebars works as expected

karupayun:
Currently, we can't zoom-in/out in Code Source Dialog when your mouse
is on the sidebars. This is an issue if you zoom-in a lot because user
is not able to come back to the original zooming-level. Also it isn't
intuitive at all because sidebars seem to be part of the dialog, instead
of a margin.

This PR makes zooming in the sidebar works as expected.

[b/181838868](http://b/181838868)

Tested using CodeViewerDialog.


beckerhe (approved):
LGTM! Thanks!

PR #2050, opened 2021-03-10
Make e2e test more reliable.

danielfenner:
This is a workaround for a bug that is not understood. It is unclear why the
method getting the ProcessList occationally throws.

Bug: http://b/181769235

### DIFF
--- a/contrib/automation_tests/test_cases/connection_window.py
+++ b/contrib/automation_tests/test_cases/connection_window.py
@@ -61,7 +61,7 @@ class FilterAndSelectFirstProcess(E2ETestCase):
                 filter_edit = self.find_control('Edit', 'FilterProcesses')
                 break
             except KeyError:
-                logging.info('Find ProcessList failed. Try again.')        
+                logging.info('Find FilterProcesses failed. Try again.')
 
         while (True):
             try:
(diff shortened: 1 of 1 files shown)


contrib/automation_tests/test_cases/connection_window.py
```diff
@@ -54,9 +54,15 @@ class FilterAndSelectFirstProcess(E2ETestCase):
     Select the first process in the process list and verify there is at least one entry in the list
     """
     def _execute(self, process_filter):
-        filter_edit = self.find_control('Edit', 'FilterProcesses')
```
reichlfl, line 57: You could do the same using `wait_for_condition`, which also re-tries after an exception.

reichlfl (commented):
The solution LGTM if we expect the filter_edit to be accessible eventually, but this could also cause an infinite wait if e.g. Orbit crashed before searching for the filter edit... `wait_for_condition` would achieve the same and time-out after 30 seconds (configurable).

contrib/automation_tests/test_cases/connection_window.py
danielfenner: Sorry, I have no idea how to do this. `wait_for_condition` takes a lambda, right? How would it be possible to extract the return value?

reichlfl: Good point, I just realized it's not that straightforward in Python and not even possible before 3.8.
It's doable like this: https://stackoverflow.com/questions/6282042/assignment-inside-lambda-expression-in-python

But this might make it less readable and enforces Python 3.8, so I'm open to leaving it as is.

reichlfl (approved):

contrib/automation_tests/test_cases/connection_window.py
danielfenner: Okay, I'll leave it as this then.

PR #2051, opened 2021-03-10
Fix Crashes on Misuse of Command Buffer Lifecycle.

florian-kuebler:
We've seen some misuses of the command buffer lifecycle, such as
submit a command buffer right after reset.
This handles this gracefully, to not crash, but also not record
timestamps.

Further, this change will log the error message on misuses only once, to not spam the output.

Test: Run tests and Run on affected samples.
Bug: http://b/182282087.

### DIFF
--- a/src/OrbitVulkanLayer/SubmissionTracker.h
+++ b/src/OrbitVulkanLayer/SubmissionTracker.h
@@ -966,7 +966,7 @@ class SubmissionTracker : public VulkanLayerProducer::CaptureStatusListener {
       return;
     }
     CHECK(command_buffer_to_state_.contains(command_buffer));
-    CommandBufferState& state = command_buffer_to_state_.at(command_buffer);
+    const CommandBufferState& state = command_buffer_to_state_.at(command_buffer);
 
     for (const Marker& marker : state.markers) {
       std::optional<SubmittedMarker> submitted_marker = std::nullopt;
--- a/src/OrbitVulkanLayer/SubmissionTracker.h
+++ b/src/OrbitVulkanLayer/SubmissionTracker.h
@@ -986,9 +986,9 @@ class SubmissionTracker : public VulkanLayerProducer::CaptureStatusListener {
           }
           CHECK(marker.label_name.has_value());
           CHECK(marker.color.has_value());
-          MarkerState marker_state{.label_name = marker.label_name.value(),
+          MarkerState marker_state{.begin_info = submitted_marker,
+                                   .label_name = marker.label_name.value(),
                                    .color = marker.color.value(),
-                                   .begin_info = submitted_marker,
                                    .depth = markers->marker_stack.size(),
                                    .depth_exceeds_maximum = marker.cut_off};
           markers->marker_stack.push(std::move(marker_state));
@@ -1019,9 +1019,9 @@ class SubmissionTracker : public VulkanLayerProducer::CaptureStatusListener {
             queue_submission_optional->value().completed_markers.emplace_back(
                 SubmittedMarkerSlice{.begin_info = marker_state.begin_info,
                                      .end_info = submitted_marker.value(),
-                                     .depth = marker_state.depth,
+                                     .label_name = std::move(marker_state.label_name),
                                      .color = marker_state.color,
-                                     .label_name = std::move(marker_state.label_name)});
+                                     .depth = marker_state.depth});
           }
 
           break;
--- a/src/OrbitVulkanLayer/SubmissionTracker.h
+++ b/src/OrbitVulkanLayer/SubmissionTracker.h
@@ -224,17 +224,10 @@ class SubmissionTracker : public VulkanLayerProducer::CaptureStatusListener {
     // other than the debug markers then.
     {
       if (command_buffer_to_state_.contains(command_buffer)) {
-        if (!logged_begin_when_not_initial_) {
-          ERROR(
-              "Calling vkBeginCommandBuffer on a command buffer that is not in the initial state "
-              "(i.e. neither freshly allocated nor reset with vkResetCommandBuffer).");
-          logged_begin_when_not_initial_ = true;
-        }
         // We end up in this case, if we have used the command buffer before and want to write new
-        // commands to it without resetting the command buffer. This is prohibited by the
-        // specification. However, we've seen this in the wild. The "vkBeginCommandBuffer" will,
-        // however, invalidate the previous commands here. So we can reset potentially blocked slots
-        // associated with this command buffer.
+        // commands to it without resetting the command buffer. Per specification,
+        // "vkBeginCommandBuffer" does also reset the command buffer, in addition to putting it
+        // into the executable state.
         ResetCommandBufferUnsafe(command_buffer);
       }
       command_buffer_to_state_[command_buffer] = {};
@@ -257,12 +250,10 @@ class SubmissionTracker : public VulkanLayerProducer::CaptureStatusListener {
       return;
     }
     if (!command_buffer_to_state_.contains(command_buffer)) {
-      if (!logged_end_when_initial_) {
-        ERROR(
-            "Calling vkEndCommandBuffer on a command buffer that is in the initial state "
-            "(i.e. either freshly allocated or reset with vkResetCommandBuffer).");
-        logged_end_when_initial_ = true;
-      }
+      LogErrorOnce(
+          "Calling vkEndCommandBuffer on a command buffer that is in the initial state "
+          "(i.e. either freshly allocated or reset with vkResetCommandBuffer).",
+          &logged_end_when_initial_);
       return;
     }
 
@@ -283,13 +274,11 @@ class SubmissionTracker : public VulkanLayerProducer::CaptureStatusListener {
     bool marker_depth_exceeds_maximum;
     {
       if (!command_buffer_to_state_.contains(command_buffer)) {
-        if (!logged_marker_begin_when_initial_) {
-          ERROR(
-              "Calling vkCmdDebugMarkerBeginEXT/vkCmdBeginDebugUtilsLabelEXT on a command buffer "
-              "that is in the initial state (i.e. either freshly allocated or reset with "
-              "vkResetCommandBuffer).");
-          logged_marker_begin_when_initial_ = true;
-        }
+        LogErrorOnce(
+            "Calling vkCmdDebugMarkerBeginEXT/vkCmdBeginDebugUtilsLabelEXT on a command buffer "
+            "that is in the initial state (i.e. either freshly allocated or reset with "
+            "vkResetCommandBuffer).",
+            &logged_marker_begin_when_initial_);
         return;
       }
       CHECK(command_buffer_to_state_.contains(command_buffer));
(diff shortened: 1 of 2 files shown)


src/OrbitVulkanLayer/SubmissionTracker.h
```diff
@@ -221,7 +221,12 @@ class SubmissionTracker : public VulkanLayerProducer::CaptureStatusListener {
     // other than the debug markers then.
     {
       if (command_buffer_to_state_.contains(command_buffer)) {
-        ERROR("Calling vkBeginCommandBuffer on a command buffer that is not in the initial state.");
+        if (!logged_begin_when_not_initial_) {
```
ronaldfw, line 224: This (and below) seems to be out-of-scope for the change? At least according to the pull request description.

src/OrbitVulkanLayer/SubmissionTracker.h
```diff
@@ -265,6 +279,16 @@ class SubmissionTracker : public VulkanLayerProducer::CaptureStatusListener {
...
     {
+      if (!command_buffer_to_state_.contains(command_buffer)) {
+        if (!logged_marker_begin_when_initial_) {
+          ERROR(
+              "Calling vkCmdDebugMarkerBeginEXT/vkCmdBeginDebugUtilsLabelEXT on a command buffer "
+              "that is in the initial state (i.e. either freshly allocated or reset with "
+              "vkResetCommandBuffer).");
+          logged_marker_begin_when_initial_ = true;
+        }
+        return;
```
ronaldfw, line 290: I guess this return here (and below, and the "continue") is the actual fix for the crash?

src/OrbitVulkanLayer/SubmissionTracker.h
ronaldfw, line 224: Can you move this out into a function? I think it becomes easier to read if the call here (and elsewhere) is:

LogErrorOnce("error string", &logged_being_when_not_initial);

The code is already quite nested as-is.

Of course, an actual ERROR_ONCE macro would be preferable, but probably out-of-scope for this change, and I'm not sure this can be done with performance that is appropriate here (as you need to keep track of messages that have been logged and check if something has been logged before).

src/OrbitVulkanLayer/SubmissionTracker.h
```diff
@@ -293,6 +317,16 @@ class SubmissionTracker : public VulkanLayerProducer::CaptureStatusListener {
     absl::WriterMutexLock lock(&mutex_);
     bool marker_depth_exceeds_maximum;
     {
```
ronaldfw, line 319: We can actually remove this scope, I think, as we moved the lock to the full scope anyway. Would help with the nestedness.

src/OrbitVulkanLayer/SubmissionTracker.h
```diff
@@ -355,6 +389,15 @@ class SubmissionTracker : public VulkanLayerProducer::CaptureStatusListener {
       for (uint32_t command_buffer_index = 0; command_buffer_index < submit_info.commandBufferCount;
            ++command_buffer_index) {
         VkCommandBuffer command_buffer = submit_info.pCommandBuffers[command_buffer_index];
```
ronaldfw, line 391: Nit: Not sure this is practical here, but when you start having multiple continues in a loop body, I generally find it easier to read when you move the loop body into a function and replace the continues with returns.

src/OrbitVulkanLayer/SubmissionTracker.h
```diff
@@ -1049,6 +1101,14 @@ class SubmissionTracker : public VulkanLayerProducer::CaptureStatusListener {
   // either in OnCaptureFinished or when completing submits. Note that calling
   // vulkan_layer_producer_->IsCapturing() is not a correct replacement for checking this boolean.
   bool is_capturing_ = false;
+
+  // The layer can handle some misuse of the Vulkan command buffer lifetime. The following bools
+  // are used to log errors on missuses only once and not spam the log.
```
ronaldfw, line 1106:
```suggestion
  // are used to log errors on misuses only once and not spam the log.
```

ronaldfw (commented):
Is the intention to also do the "log once" change in this PR?

florian-kuebler:
> Is the intention to also do the "log once" change in this PR?

Yes, I adjusted the PR description (which I will use as commit message when squashing).

src/OrbitVulkanLayer/SubmissionTracker.h
ronaldfw: As @beckerhe pointed out this can be done with a static bool per macro invocation. So I think for the use case here, we should do that.

florian-kuebler (commented):
Thanks for the review.

src/OrbitVulkanLayer/SubmissionTracker.h
florian-kuebler: So I should do that right away in this PR? Can do that tomorrow.

florian-kuebler: Done.

src/OrbitVulkanLayer/SubmissionTracker.h
```diff
@@ -295,18 +305,24 @@ class SubmissionTracker : public VulkanLayerProducer::CaptureStatusListener {
...
+    }
+    CHECK(command_buffer_to_state_.contains(command_buffer));
+    CommandBufferState& state = command_buffer_to_state_.at(command_buffer);
+    marker_depth_exceeds_maximum =
+        state.local_marker_stack_size > max_local_marker_depth_per_command_buffer_;
+    Marker marker{.type = MarkerType::kDebugMarkerEnd, .cut_off = marker_depth_exceeds_maximum};
+    state.markers.emplace_back(std::move(marker));
+    // We might see more "ends" than "begins", as the "begins" can be on a different command
+    // buffer.
+    if (state.local_marker_stack_size != 0) {
```
ronaldfw, line 324: Nit: I find it slightly more readable if you say > 0 here. (But since local_marker_stack_size is unsigned, this doesn't actually make a different.)

src/OrbitVulkanLayer/SubmissionTracker.h
```diff
@@ -976,6 +893,140 @@ class SubmissionTracker : public VulkanLayerProducer::CaptureStatusListener {
     command_buffer_to_state_.erase(command_buffer);
   }
 
+  void PersistSingleCommandBufferOnSubmitUnsafe(
```
ronaldfw, line 896: Why "unsafe"? Because this doesn't lock the mutex? Unless I'm missing something, this should never be called without actually holding the mutex. I would remove the "unsafe" in the name and add `mutex_.AssertHeld();` at the beginning.

src/OrbitVulkanLayer/SubmissionTracker.h
```diff
@@ -976,6 +893,140 @@ class SubmissionTracker : public VulkanLayerProducer::CaptureStatusListener {
...
+          .command_buffer_begin_slot_index = state.command_buffer_begin_slot_index,
+          .command_buffer_end_slot_index = state.command_buffer_end_slot_index.value()};
+      submitted_submit_info->command_buffers.emplace_back(submitted_command_buffer);
+    }
+  }
+
+  void PersistDebugMarkersOfASingleCommandBufferOnSubmitUnsafe(
+      VkCommandBuffer command_buffer, std::optional<QueueSubmission>* queue_submission_optional,
+      QueueMarkerState* markers, std::vector<uint32_t>* marker_slots_not_needed_to_read) {
+    CHECK(queue_submission_optional != nullptr);
```
ronaldfw, line 955: Same as above, add `mutex_.AssertHeld();` and consider removing the "unsafe" in the name.

src/OrbitVulkanLayer/SubmissionTracker.h
florian-kuebler: Agree. Done.

src/OrbitVulkanLayer/SubmissionTracker.h
florian-kuebler: Yes that was regarding the Mutex. I used this pattern kind of across the whole layer. We used this in my previous project a lot to make it explicit to the caller of a method that is has some very specific assumptions, which I kind of liked. 
Anyways, the `AssertHeld` makes this assumption even more explicit, so I added it. For now, I removed the "unsafe" part of the name. But we are already in a mixture of this in the layer code base. I will make my mind about this, and address this in a separate PR (if needed).

src/OrbitVulkanLayer/SubmissionTracker.h
florian-kuebler: Done.

florian-kuebler (commented):
Thanks for the review. I should have addressed all your comments.

src/OrbitVulkanLayer/SubmissionTracker.h
ronaldfw: IMO, with `AssertHeld`, the "unsafe" is redundant or even wrong. You can't call it in an unsafe way anyway, so there's no need to know about it in the name. If you use it wrong, you will learn about it. 

An "unsafe" version of a method without `AssertHeld` suggests that there are use cases where I can call the function without holding the mutex. So unless these use cases exist, I would always use the assert and remove the "unsafe".

florian-kuebler: Yep, with the concrete examples and the asserts, I agree.

src/OrbitVulkanLayer/SubmissionTracker.h
```diff
@@ -977,6 +893,143 @@ class SubmissionTracker : public VulkanLayerProducer::CaptureStatusListener {
...
+    CHECK(marker_slots_not_needed_to_read != nullptr);
+
+    if (!command_buffer_to_state_.contains(command_buffer)) {
+      ERROR_ONCE(
+          "Calling vkQueueSubmit on a command buffer that is in the initial state (i.e. "
+          "either freshly allocated or reset with vkResetCommandBuffer).");
+      return;
+    }
+    CHECK(command_buffer_to_state_.contains(command_buffer));
+    CommandBufferState& state = command_buffer_to_state_.at(command_buffer);
```
ronaldfw, line 969: Can this be a const ref? That would make it more obvious that AsserReaderHeld() is enough above.

src/OrbitVulkanLayer/SubmissionTracker.h
```diff
@@ -977,6 +893,143 @@ class SubmissionTracker : public VulkanLayerProducer::CaptureStatusListener {
...
+          .command_buffer_begin_slot_index = state.command_buffer_begin_slot_index,
+          .command_buffer_end_slot_index = state.command_buffer_end_slot_index.value()};
+      submitted_submit_info->command_buffers.emplace_back(submitted_command_buffer);
+    }
+  }
+
+  void PersistDebugMarkersOfASingleCommandBufferOnSubmit(
+      VkCommandBuffer command_buffer, std::optional<QueueSubmission>* queue_submission_optional,
+      QueueMarkerState* markers, std::vector<uint32_t>* marker_slots_not_needed_to_read) {
+    mutex_.AssertReaderHeld();
```
ronaldfw, line 957: Why is read enough here? (See also question below.)

florian-kuebler: It is not. Thanks. That was a simple copy&paste error.

src/OrbitVulkanLayer/SubmissionTracker.h
florian-kuebler: No it can't. I changed the assertion.

florian-kuebler (commented):
Thanks. I changed the assertions to `AssertHeld`.

src/OrbitVulkanLayer/SubmissionTracker.h
ronaldfw: I'm probably missing something, but what is the reason, this can't be a const&? I see a loop with const& over state.markers, and no other access to `state`.

florian-kuebler: Sorry for the confusion (it seems to be one of those Mondays). I looked at the `PersistSingleCommandBufferOnSubmit` method, where the `state` can not be const. Here, you are totally right. I made it const ref.

src/OrbitVulkanLayer/SubmissionTracker.h
```diff
@@ -977,6 +893,143 @@ class SubmissionTracker : public VulkanLayerProducer::CaptureStatusListener {
...
+          .command_buffer_begin_slot_index = state.command_buffer_begin_slot_index,
+          .command_buffer_end_slot_index = state.command_buffer_end_slot_index.value()};
+      submitted_submit_info->command_buffers.emplace_back(submitted_command_buffer);
+    }
+  }
+
+  void PersistDebugMarkersOfASingleCommandBufferOnSubmit(
+      VkCommandBuffer command_buffer, std::optional<QueueSubmission>* queue_submission_optional,
+      QueueMarkerState* markers, std::vector<uint32_t>* marker_slots_not_needed_to_read) {
+    mutex_.AssertHeld();
```
ronaldfw, line 957: I think now this can be an AsserReaderHeld().

florian-kuebler: My understanding is that we are writing to `queue_submission_optional` (line 1019) as well as to the `markers->markers_stack`.

Even those are passed in as pointer arguments, and are not read directly for the members holding them, they should not be written simultaneously.  So I would keep the lock as it is.

ronaldfw: That makes sense, thanks for clarifying!

ronaldfw (approved):
LGTM

PR #2052, opened 2021-03-10
Target label

antonrohr:
This makes the target label the same thing in the ProfilingTargetDialog and the Orbitmainwindow (and adds tests)


src/OrbitQt/CMakeLists.txt
```diff
@@ -112,9 +114,9 @@ target_link_libraries(
           SyntaxHighlighter
           qtpropertybrowser::qtpropertybrowser)
 
-if(WITH_CRASH_HANDLING)
-  target_link_libraries(OrbitQt PRIVATE CrashHandler)
-  target_compile_definitions(OrbitQt PRIVATE ORBIT_CRASH_HANDLING)
+if(WITH_CRASH_HANDLING)
+  target_link_libraries(OrbitQt PRIVATE CrashHandler)
+  target_compile_definitions(OrbitQt PRIVATE ORBIT_CRASH_HANDLING)
```
beckerhe, line 119: What has changed here?

src/OrbitQt/TargetLabel.h
```diff
@@ -0,0 +1,35 @@
...
+#define ORBIT_QT_TARGET_LABEL_H_
+
+#include <QLabel>
+
+#include "OrbitClientData/ProcessData.h"
+#include "OrbitGgp/Instance.h"
+
+namespace orbit_qt {
+
+class TargetLabel : public QLabel {
```
beckerhe, line 15: I think you shouldn't inherit from QLabel here. To me it looks like the label should be a private member instead.

You're trying to enforce invariants by setting the label's properties in the setters of `TargetLabel`. But if you inherit from QLabel, the user of this class will be able to mess with this.

I see two solutions:
- Either make it a private member, so you have full control over the invariance and the user can rely on that.
- Or make it obvious that it's the user's responsibility, by making the setters free functions which take a label as an argument. `SetFileTarget(QLabel* target_label, const std::filesystem::path& path);`

Obviously the first is closer to the current solution.

src/OrbitQt/TargetLabelTest.cpp
```diff
@@ -0,0 +1,115 @@
...
+#include "process.pb.h"
+
+namespace orbit_qt {
+
+namespace fs = std::filesystem;
+
+TEST(TargetLabel, Constructor) {
+  TargetLabel label{};
+  EXPECT_TRUE(label.text().isEmpty());
+  EXPECT_TRUE(label.styleSheet().isEmpty());
```
beckerhe, line 20: I'm not saying you should remove it, only to think about it if this couldn't be checked differently.

It looks a lot like an implementation detail to me. In the end for the user of the class it should be of no concern
how the font coloring is done. You could also use QPalette, or overwrite the paintEvent method, or even role your own label implementation. The only important thing in the end is the color of the text, not the stylesheet.

And by putting that check into the tests you even guarantee to the user that this is officially observable behaviour. That means someone could start checking the stylesheet property to figure out in which mode the target label currently is. Still unlikely for this use-case, but I hope you get the idea.

src/OrbitQt/TargetLabelTest.cpp
```diff
@@ -0,0 +1,115 @@
...
+  EXPECT_TRUE(label.styleSheet().contains(TargetLabel::kRedColor));
+  EXPECT_EQ(label.toolTip(), error_message);
+}
+
+TEST(TargetLabel, StateChanging) {
+  TargetLabel label;
+
+  label.SetProcessAlive();
+  EXPECT_TRUE(label.styleSheet().contains(TargetLabel::kGreenColor));
+  EXPECT_TRUE(label.toolTip().isEmpty());
```
reichlfl, line 103: Nit: Since you test this twice (here and in `SetProcessAlive`), it would be easier to maintain and update if you'd move it to a function. Same for the three tests below.

src/OrbitQt/TargetLabel.cpp
```diff
@@ -0,0 +1,50 @@
...
+}
+
+void TargetLabel::SetLocalTarget(const ProcessData& process) {
+  setText(QString{"%1 @ localhost"}.arg(QString::fromStdString(process.name())));
+  SetProcessAlive();
+}
+
+void TargetLabel::SetProcessDead() {
+  setStyleSheet(kStyleSheetFormat.arg(kOrangeColor));
+  setToolTip("The process ended on the instance");
```
beckerhe, line 37: Nit: It's a sentence, I would make a fullstop at the end.

src/OrbitQt/TargetLabel.h
```diff
@@ -0,0 +1,35 @@
...
+ public:
+  explicit TargetLabel(QWidget* parent = nullptr);
+  void SetFileTarget(const std::filesystem::path& path);
+  void SetStadiaTarget(const ProcessData& process, const orbit_ggp::Instance& instance);
+  void SetLocalTarget(const ProcessData& process);
```
reichlfl, line 20: I'm not super happy with the API here - there's 3 methods that look like regular setters, but their usage is mutually exclusive.
Maybe just rename them to make this clear, e.g. `ChangeToXXXTarget`.

src/OrbitQt/TargetLabel.h
```diff
@@ -0,0 +1,35 @@
...
+  void SetProcessDead();
+  void SetProcessAlive();
+  void SetConnectionDead(const QString& error_message);
```
reichlfl, line 23: Same as above. Maybe `SetStatus` with an enum parameter, and an optional tooltip parameter?

src/OrbitQt/TargetLabel.h
```diff
@@ -0,0 +1,35 @@
...
+
+  static const QString kDefaultColor;
+  static const QString kGreenColor;
+  static const QString kOrangeColor;
+  static const QString kRedColor;
```
reichlfl, line 28: Nit: These should be private, and I assume they are public because of the unit test? I'd avoid making something public just to be testable - it's good practice to test through the public API and ignore internals, so this indicates you're testing internals.

However - since this whole widget has purely visual purposes, this level of detail may be fine here... I just wanted to point this out, but I'm undecided.

reichlfl (commented):
This is nice, the encapsulation makes it a lot cleaner. PTAL at my comments regarding the API, I'd suggest to at least rename the methods to sound less like setters.

beckerhe (approved):
Otherwise LGTM

src/OrbitQt/TargetLabel.h
reichlfl: I second this, because it'll also make it easier to extend it by e.g. adding a "Connected" icon as we discussed.

src/OrbitQt/CMakeLists.txt
antonrohr: Took me a while to figure it out. I think vscode or clang format fixed line endings here (CRLF -> LF). 
Screenshot 2021-03-11 at 10 49 37

beckerhe: So nothing changed and it can be removed from this commit?

antonrohr: Well. The file contained mixed line endings before this commit. Which is a bad thing in the first place. Now its just LF line endings and thats a good thing. I could add a dedicated "Fix line endings" commit, but for the sake of simplicity, I would leave it as is. wdyt?

beckerhe: I personally value clean commits a lot higher than consistent line endings, but I also don't feel strongly at all about this. So if you prefer to leave it, that's fine with me!

src/OrbitQt/TargetLabelTest.cpp
antonrohr: outdated

src/OrbitQt/TargetLabel.h
antonrohr: I changed it, ptal

src/OrbitQt/TargetLabel.h
antonrohr: I changed it, ptal

src/OrbitQt/TargetLabel.h
antonrohr: I made them an implementation detail and dont test for correct color anymore, only for change in color. ptal

src/OrbitQt/TargetLabel.h
antonrohr: I implemented the first solution (in a way). I mark this as resolved

src/OrbitQt/TargetLabelTest.cpp
antonrohr: As we talked offline, I changed the testing of this. It now uses QPalette instead of spreadsheet and I only test for change in color.

src/OrbitQt/TargetLabel.cpp
antonrohr: done

antonrohr:
I changed quite a bit in this PR and I needed the cmake change that was submitted yesterday, hence I rebased and force pushed. @reichlfl @beckerhe please take a look again. 

This now also adds cpu utilization and icons and therefore does everything tracked in http://b/177305224

src/OrbitQt/ProfilingTargetDialog.cpp
```diff
@@ -265,13 +264,14 @@ void ProfilingTargetDialog::SetupStadiaStates() {
...
     CHECK(process_ != nullptr);
-    QString process_name = QString::fromStdString(process_->name());
     CHECK(ui_->stadiaWidget->GetSelectedInstance().has_value());
-    QString instance_name = ui_->stadiaWidget->GetSelectedInstance().value().display_name;
-    ui_->targetLabel->setText(QString("%1 @ %2").arg(process_name).arg(instance_name));
+    ui_->targetLabel->ChangeToStadiaTarget(*process_,
+                                           ui_->stadiaWidget->GetSelectedInstance().value());
   });
+  QObject::connect(&state_stadia_process_selected_, &QState::exited, this,
+                   [this] { ui_->targetLabel->Clear(); });
```
beckerhe, line 274: why not
QObject::connect(&state_stadia_process_selected_, &QState::exited, ui_->targetLabel, &TargetLabel::Clear);?

src/OrbitQt/ProfilingTargetDialog.cpp
```diff
@@ -458,32 +459,45 @@ void ProfilingTargetDialog::OnProcessListUpdate(
...
-    if (ui_->processesTableView->selectionModel()->hasSelection()) return;
+    // In case there is a selection already, do not change anything, only update the cpu usage
+    if (ui_->processesTableView->selectionModel()->hasSelection()) {
+      const auto it =
+          std::find_if(process_list.begin(), process_list.end(), [&](const auto& process) {
+            return process.pid() == ui_->processesTableView->selectionModel()
+                                        ->currentIndex()
+                                        .data(Qt::UserRole)
+                                        .value<const ProcessInfo*>()
+                                        ->pid();
```
beckerhe, line 470: The right hand side of this comparison is constant, isn't it? It's hard to understand at first that this doesn't change. I'd suggest to put this pid into a variable first, and then search for the correct process by comparing to this variable.

beckerhe: One more thing: The "current index" is somehow independent from a selection, and mainly meant for keyboard navigation.

You can have a selection and a different current index, or maybe no current index at all. I think you wanna ask the selection model for the selectedRows() and then take the first one.

src/OrbitQt/orbitmainwindow.cpp
```diff
@@ -1179,9 +1173,8 @@ void OrbitMainWindow::OnStadiaConnectionError(std::error_code error) {
                               .arg(target.GetConnection()->GetInstance().display_name)
                               .arg(QString::fromStdString(error.message()));
 
-  target_label_->setStyleSheet(
-      kTargetLabelDefaultStyleSheet.arg(kTargetLabelColorTargetDisconnected));
-  target_label_->setToolTip(error_message);
+  target_label_->SetConnectionDead(error_message);
+  ui->menuBar->adjustSize();
```
beckerhe, line 1177: Nit: This feels a bit like a dependency on the implemenation of TargetLabel, because you need to know which method of TargetLabel changes the widget's size. Ideally this is decoupled and TargetLabel tells you when a call of adjustSize is needed.

This could be done by a signal in TargetLabel. Maybe there is already one in QWidget which could be used. Haven't checked though.

beckerhe (commented):
Very Nice, Anton. I added some comments

src/OrbitQt/TargetLabel.h
```diff
@@ -0,0 +1,68 @@
...
+
+ private:
+  std::unique_ptr<Ui::TargetLabel> ui_;
+  QString process_ = "";
+  QString machine_ = "";
+  std::optional<Icon> icon_;
+
+  const QPixmap process_alive_icon_;
+  const QPixmap process_ended_icon_;
+  const QPixmap connection_ended_icon_;
```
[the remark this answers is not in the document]
antonrohr: They are actually not there because of the tests, but because I "compute" them. Aka take the white icon and color it pixel by pixel (see TargetDialog.cpp anon namespace `ColorizeIcon()`). I did not want to make this computation everytime I use them, but once. I test via the enum. 

The static is a good idea. I will change it to static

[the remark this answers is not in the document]
beckerhe: But that still doesn't explain why they need to be in the class. Why not static variables in the cpp file? You could still compute them.

src/OrbitQt/orbitmainwindow.cpp
antonrohr: I like this idea a lot. QWidget did not have a signal that was useful, so I added a `SizeChanged` signal. done

src/OrbitQt/ProfilingTargetDialog.cpp
antonrohr: done

src/OrbitQt/ProfilingTargetDialog.cpp
antonrohr: done

src/OrbitQt/ProfilingTargetDialog.cpp
```diff
@@ -397,8 +396,10 @@ void ProfilingTargetDialog::SetupFileStates() {
   QObject::connect(&state_file_selected_, &QState::entered, this, [this] {
     QString filename = QString::fromStdString(selected_file_path_.filename().string());
     ui_->selectedFileLabel->setText(filename);
-    ui_->targetLabel->setText(filename);
+    ui_->targetLabel->ChangeToFileTarget(selected_file_path_);
   });
+  QObject::connect(&state_file_selected_, &QState::exited, this,
+                   [this] { ui_->targetLabel->Clear(); });
```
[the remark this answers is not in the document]
antonrohr: done

src/OrbitQt/ProfilingTargetDialog.cpp
```diff
@@ -341,11 +340,12 @@ void ProfilingTargetDialog::SetupLocalStates() {
...
                                               &state_local_no_process_selected_);
-  QObject::connect(&state_local_process_selected_, &QState::entered, [this] {
+  QObject::connect(&state_local_process_selected_, &QState::entered, this, [this] {
     CHECK(process_ != nullptr);
-    QString process_name = QString::fromStdString(process_->name());
-    ui_->targetLabel->setText(QString("%1 @ localhost").arg(process_name));
+    ui_->targetLabel->ChangeToLocalTarget(*process_);
   });
+  QObject::connect(&state_local_process_selected_, &QState::exited, this,
+                   [this] { ui_->targetLabel->Clear(); });
```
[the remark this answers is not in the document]
antonrohr: Thank you! Missed that. done

src/OrbitQt/TargetLabel.h
```diff
@@ -0,0 +1,68 @@
...
+  bool SetProcessCpuUsage(double cpu_usage);
+  bool SetProcessEnded();
+  bool SetConnectionDead(const QString& error_message);
+
+  void Clear();
+
+  [[nodiscard]] QColor GetColor() const;
+  [[nodiscard]] QString GetText() const;
+  [[nodiscard]] QString GetToolTip() const { return toolTip(); }
+  [[nodiscard]] std::optional<Icon> GetIcon() const { return icon_; }
```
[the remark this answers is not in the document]
antonrohr: done

src/OrbitQt/TargetLabel.cpp
```diff
@@ -0,0 +1,171 @@
...
+  process_ = process_name;
+  machine_ = kLocalhostName;
+  SetProcessCpuUsage(cpu_usage);
+}
+
+bool TargetLabel::SetProcessCpuUsage(double cpu_usage) {
+  if (process_.isEmpty() || machine_.isEmpty()) return false;
+
+  ui_->textLabel->setText(
+      QString{"%1 (%2%) @ %3"}.arg(process_, QString::number(cpu_usage, 'f', 0), machine_));
```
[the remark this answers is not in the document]
antonrohr: Did not know about that arg overload. neat! done

src/OrbitQt/TargetLabel.cpp
```diff
@@ -0,0 +1,171 @@
...
+}
+
+void TargetLabel::ChangeToLocalTarget(const QString& process_name, double cpu_usage) {
+  Clear();
+  process_ = process_name;
+  machine_ = kLocalhostName;
+  SetProcessCpuUsage(cpu_usage);
+}
+
+bool TargetLabel::SetProcessCpuUsage(double cpu_usage) {
```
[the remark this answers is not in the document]
antonrohr: done

src/OrbitQt/TargetLabel.h
```diff
@@ -0,0 +1,68 @@
...
+namespace Ui {
+class TargetLabel;  // IWYU pragma: keep
+}
+
+namespace orbit_qt {
+
+class TargetLabel : public QWidget {
+  Q_OBJECT
+ public:
+  enum class Icon { kProcessAlive, kProcessEnded, kConnectionDead };
```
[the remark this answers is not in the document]
antonrohr: I renamed it to IconType. For easy reviewing its a separate fixup rename commit. ptal

src/OrbitQt/TargetLabel.h
```diff
@@ -0,0 +1,68 @@
...
+
+  [[nodiscard]] QColor GetColor() const;
+  [[nodiscard]] QString GetText() const;
+  [[nodiscard]] QString GetToolTip() const { return toolTip(); }
+  [[nodiscard]] std::optional<Icon> GetIcon() const { return icon_; }
+
+ private:
+  std::unique_ptr<Ui::TargetLabel> ui_;
+  QString process_ = "";
+  QString machine_ = "";
```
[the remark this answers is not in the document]
antonrohr: done

src/OrbitQt/TargetLabel.h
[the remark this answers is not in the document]
antonrohr: done

src/OrbitQt/orbitmainwindow.cpp
```diff
@@ -5,6 +5,8 @@
 #include "orbitmainwindow.h"
 
 #include <absl/flags/declare.h>
+#include <qmenubar.h>
+#include <qobject.h>
```
beckerhe, line 9: Nit

src/OrbitQt/TargetLabel.cpp
```diff
@@ -4,6 +4,8 @@
 
 #include "TargetLabel.h"
 
+#include <qpixmap.h>
+
```
beckerhe, line 8: Nit

beckerhe (approved):
Thanks a lot for the fixes. LGTM

antonrohr:
@reichlfl I think we are good here code wise with Hennings review, maybe you could just look at these screenshots to make sure we are on the same page.

The blurriness in the red color is an encoding artifact of chrome remote desktop. I dont know what the encoders problem with red color is, but its definitely sharp without CRD.

Screenshot 2021-03-18 at 10 56 00
Screenshot 2021-03-18 at 10 55 00
Screenshot 2021-03-18 at 10 55 20
Screenshot 2021-03-18 at 10 54 54
Screenshot 2021-03-18 at 10 54 42

src/OrbitQt/orbitmainwindow.cpp
antonrohr: One day I will find time to setup auto fixing iwyu on every save. This gets really annoying. Thank you for pointing it out. done (and rebased)

src/OrbitQt/TargetLabel.h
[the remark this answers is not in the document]
beckerhe: Sounds good.

reichlfl:
> @reichlfl I think we are good here code wise with Hennings review, maybe you could just look at these screenshots to make sure we are on the same page.

Sorry, I didn't realize I had open discussion items here. I'll trust in your and Hennings judgment on this, no need to wait for me to do a second review.

Regarding the screenshots: Really nice, I like the inclusion of the icon. A nitpick, the red looks really aggressive to a point where it's a bit hard to read on the grey background, maybe it should be a bit of a lighter shade. But that might also be due to the compression.

Orange indicates the process connection was lost, but the instance is still alive? A different icon for this could be nice, but not urgently needed. Could also be changed at a later point. Do we have tooltips that indicate the difference between the red and orange state? It's not clear from just looking at the label / icon.

Maybe we just use a "Warning" icon when the process connection is lost? https://material.io/resources/icons/?search=problem&icon=report_problem&style=outline
In the future, I can imagine you can still take a capture as we more and more become a system profiler (e.g. memory, scheduling and GPU information is still valid), so "Warning" may be appropriate.

reichlfl (approved):

antonrohr:
Thanks for the suggestions Florian. I changed the icon and made the red a bit brighter. There are also tooltips. The process has ended one could be a bit more verbose, in case you have better idea. 

Screenshot 2021-03-18 at 13 31 58
Screenshot 2021-03-18 at 13 41 29
Capture

reichlfl:
> Thanks for the suggestions Florian. I changed the icon and made the red a bit brighter. There are also tooltips. The process has ended one could be a bit more verbose, in case you have better idea.
> 

More nitpicking, I'd probably make the icons a bit smaller - but this is really a minor detail at this point.
Can we already automatically reconnect to the process? Then I'd state that in the tooltip (e.g. "Restart a process with a matching name to continue profiling").

antonrohr:
I made the icons smaller. 

The reconnection to a restarted process is not yet implemented. The actual change is just a couple of lines, but I am somewhat afraid to break something at a different.  This will be a different PR. (I will remember to adjust the tooltip)

Screenshot 2021-03-18 at 16 33 25
Screenshot 2021-03-18 at 16 32 57

reichlfl:
> I made the icons smaller.
> 
> The reconnection to a restarted process is not yet implemented. The actual change is just a couple of lines, but I am somewhat afraid to break something at a different. This will be a different PR. (I will remember to adjust the tooltip)

Makes sense, thanks!

PR #2054, opened 2021-03-11
Add LOG_ONCE and ERROR_ONCE macros.

florian-kuebler:
This only logs once on each Macro definition.

```
LOG_ONCE("FOO");
LOG_ONCE("FOO");
```
Will print "FOO" twice.

Test: Manually insert the Macro and check the output.

### DIFF
--- a/src/OrbitBase/include/OrbitBase/Logging.h
+++ b/src/OrbitBase/include/OrbitBase/Logging.h
@@ -48,6 +48,17 @@ constexpr const char* kLogTimeFormat = "%Y-%m-%dT%H:%M:%E6S";
 
 #define ERROR(format, ...) LOG("Error: " format, ##__VA_ARGS__)
 
+#define LOG_ONCE(format, ...)           \
+  do {                                  \
+    static bool already_logged = false; \
+    if (!already_logged) {              \
+      already_logged = true;            \
+      LOG(format, ##__VA_ARGS__);       \
+    }                                   \
+  } while (0)
+
+#define ERROR_ONCE(format, ...) LOG_ONCE("Error: " format, ##__VA_ARGS__)
+
 #define FATAL(format, ...)                \
   do {                                    \
     LOG("Fatal: " format, ##__VA_ARGS__); \
--- a/src/OrbitBase/include/OrbitBase/Logging.h
+++ b/src/OrbitBase/include/OrbitBase/Logging.h
@@ -12,6 +12,7 @@
 
 #include <cstdio>
 #include <filesystem>
+#include <mutex>
 #include <string>
 
 #ifdef _WIN32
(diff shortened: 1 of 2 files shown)


beckerhe (approved):
Is thread-safety a concern? If yes we could think about using a std::atomic<bool> instead. Otherwise LGTM!

florian-kuebler:
> Is thread-safety a concern? If yes we could think about using a std::atomic instead. Otherwise LGTM!

For the concrete example not. But better be safe!

florian-kuebler:
Made it an atomic<bool>

beckerhe:
> Made it an atomic

That's not enough. You also need to make the if-condition and the assignment an atomic operation, using `std::atomic<T>::compare_exchange_weak`.

I think
```
bool already_logged_false = false;
if (already_logged.compare_exchange_weak(alread_logged_false, true)) {
```

should work, but I haven't tested this.

dpallotti:
> You also need to make the if-condition and the assignment an atomic operation, using std::atomic<T>::compare_exchange_weak.

Absolutely.

But IIRC `compare_exchange_weak` is usually meant to be used in a loop, that's why it's "weak". Not super sure if that applies here though.

What about using `std::once_flag` and `std::call_once` instead, so that we don't have to deal with the technicalities of atomics?

beckerhe:
> > You also need to make the if-condition and the assignment an atomic operation, using std::atomic::compare_exchange_weak.
> 
> Absolutely.
> 
> But IIRC `compare_exchange_weak` is usually meant to be used in a loop, that's why it's "weak". Not super sure if that applies here though.
> 
> What about using `std::once_flag` and `std::call_once` instead, so that we don't have to deal with the technicalities of atomics?

Ah right. I always forget about those. Yes, I think that's the way to go here!

florian-kuebler:
> > You also need to make the if-condition and the assignment an atomic operation, using std::atomic::compare_exchange_weak.
> 
> Absolutely.
> 
> But IIRC `compare_exchange_weak` is usually meant to be used in a loop, that's why it's "weak". Not super sure if that applies here though.
> 
> What about using `std::once_flag` and `std::call_once` instead, so that we don't have to deal with the technicalities of atomics?

Done.

src/OrbitBase/include/OrbitBase/Logging.h
```diff
@@ -48,6 +49,14 @@ constexpr const char* kLogTimeFormat = "%Y-%m-%dT%H:%M:%E6S";
 
 #define ERROR(format, ...) LOG("Error: " format, ##__VA_ARGS__)
 
+#define LOG_ONCE(format, ...)                                                  \
+  do {                                                                         \
+    static std::once_flag already_logged_flag;                                 \
+    std::call_once(already_logged_flag, []() { LOG(format, ##__VA_ARGS__); }); \
```
beckerhe, line 55: The lambda might need `&` as a capture argument

beckerhe: And could you add some usage of these macros somewhere? Otherwise they won't be tested.

florian-kuebler: I tried it manually (as I was intending to use it in a different PR #2051). It worked without the `&`, but I can safely add it. I also added a usage.

beckerhe: Yes, it worked because `format` is a constexpr expression, and __VA_ARGS__ is empty in your example. Have you tried it with an argument and it still worked? I think it would break when the ampersand is not there and the macro gets an argument.

Anyways, thanks a lot for addressing the comment. LGTM!

florian-kuebler: You are totally right, tested it locally, and it did not compile.

florian-kuebler:
Note, that the gcc build error is unrelated, so I am going to submit this change now.

PR #2056, opened 2021-03-11
fix gcc9 compilation error

dimitry-:


src/OrbitCaptureClient/CaptureClient.cpp
```diff
@@ -58,6 +58,7 @@ static CaptureOptions::InstrumentedFunction::FunctionType InstrumentedFunctionTy
         FunctionInfo_OrbitType_FunctionInfo_OrbitType_INT_MAX_SENTINEL_DO_NOT_USE_:
       UNREACHABLE();
   }
+  UNREACHABLE();
```
dpallotti, line 61: Why not just restore the `default` in the switch? I don't think we should use those values directly.

dimitry-: because we want a compile time error when enum class changes.

dpallotti: But with the addition of this `UNREACHABLE()` would you still get that compile time error now? If I get this correctly this new `UNREACHABLE()` acts like `default` now.

dimitry-: yes, if there is no default in the switch - and you add another value to enum you get compile time error. (it is a warning promoted to a error)

dpallotti: Makes sense.

dpallotti (approved):

florian-kuebler (approved):
LGTM, thanks

PR #2059, opened 2021-03-11
Add Vulkan headers to conan dependencies.

florian-kuebler:
This adds the "vulkan-headers/1.1.77.0" requirement to the Linux
build.
Afterwards, gen_lockfile.sh was run and and the changed hashes
of existing entries were reset.

In order to run the gen_lockfile.sh, conan needed to be upgraded.

### DIFF
--- a/bootstrap-orbit.sh
+++ b/bootstrap-orbit.sh
@@ -26,8 +26,7 @@ done
 readonly REQUIRED_PACKAGES=( build-essential libglu1-mesa-dev mesa-common-dev \
                              libxmu-dev libxi-dev libopengl0 qtbase5-dev \
                              qtwebengine5-dev libqt5webchannel5-dev \
-                             libqt5websockets5-dev libxxf86vm-dev python3-pip \
-                             libvulkan-dev vulkan-validationlayers-dev )
+                             libqt5websockets5-dev libxxf86vm-dev python3-pip )
 
 function add_ubuntu_universe_repo {
   sudo add-apt-repository universe
@@ -80,7 +79,7 @@ echo "Checking if conan is available..."
 which conan >/dev/null
 if [ $? -ne 0 ]; then
   echo "Couldn't find conan. Trying to install via pip..."
-  pip3 install --user conan==1.29.2 || exit $?
+  pip3 install --user conan==1.32.0 || exit $?
 
   which conan >/dev/null
   if [ $? -ne 0 ]; then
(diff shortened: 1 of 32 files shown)


bootstrap-orbit.sh
```diff
@@ -80,7 +80,7 @@ echo "Checking if conan is available..."
 which conan >/dev/null
 if [ $? -ne 0 ]; then
   echo "Couldn't find conan. Trying to install via pip..."
-  pip3 install --user conan==1.29.2 || exit $?
+  pip3 install --user conan==1.32.0 || exit $?
```
beckerhe, line 83: If we change that version, we should do it everywhere. Also in the bootstrap-orbit.ps1 for Windows, and also in the Kokoro scripts for the CI.

florian-kuebler: Good catch. Hope I have all the files now.

florian-kuebler:
> On top of that: I think we should also add the cmake integration here, since this will be useless if we figure out later that it doesn't work as expected. Have you tried that already?

Trying to this right now. But keep fighting with the CI.

florian-kuebler:
I also needed to add the CONAN_PKG::vulkan-headers in the FindVulkanValidationLayers.cmake. However, the vk_layer_dispatch_table.h header there does not seem the be compatible with the Vulkan 1.1.77 and does not compile. Not unexpected, but still requires to also do the vk_layer_dispatch_table.h change along with this.

florian-kuebler:
Okay, now everything seems working:

I needed to bump to 1.1.114, as the vk_layer_dispatch_table.h is not there in 1.1.77: https://github.com/KhronosGroup/Vulkan-Loader/blob/sdk-1.1.114/loader/generated/vk_layer_dispatch_table.h

I removed the WITH_VULKAN_LAYER flag and always compile the layers in Linux (hope that also ensures the tests get executed).

Adjusted the VulkanValidattionLayers package accordingly.

beckerhe (approved):
Thanks Florian. This looks good to me!

Regarding the coverage build failure. Davide had the same problem and has fixed that on main. So you just need to rebase and the error should go away.

Regarding the GCC9 build error: I assume nobody tried to build the layer code with GCC9 before. But I guess should be a simple enough fix.

florian-kuebler:
> Regarding the GCC9 build error: I assume nobody tried to build the layer code with GCC9 before. But I guess should be a simple enough fix.
 
Yep, we never did this. There is a bit of a hickhack with named initialisation of structs. Working on it, would be nice if you could have another look as soon as this is done.

Will also rebase to main.

beckerhe:
> > Regarding the GCC9 build error: I assume nobody tried to build the layer code with GCC9 before. But I guess should be a simple enough fix.
> 
> Yep, we never did this. There is a bit of a hickhack with named initialisation of structs. Working on it, would be nice if you could have another look as soon as this is done.
> 
> Will also rebase to main.

Sure. Just ping me when you're done.

PR #2061, opened 2021-03-11
Inject a shared library into a tracee.

danielfenner:
The code of the tracee can then be overridden to invoke functions in
the injected lib.

Bug: http://b/181304023
Test: Unit tests in PR.

### DIFF
--- a/src/UserSpaceInstrumentation/InjectLibraryInTracee.cpp
+++ b/src/UserSpaceInstrumentation/InjectLibraryInTracee.cpp
@@ -7,7 +7,6 @@
 #include <absl/base/casts.h>
 #include <absl/strings/match.h>
 #include <absl/strings/str_format.h>
-#include <dlfcn.h>
 #include <sys/ptrace.h>
 #include <sys/wait.h>
 
@@ -31,7 +30,7 @@ using orbit_elf_utils::ElfFile;
 using orbit_elf_utils::ReadModules;
 
 // Size of the small amount of memory we need in the tracee to write machine code into.
-constexpr uint64_t kScratchPadSize = 1024;
+constexpr uint64_t kCodeScratchPadSize = 1024;
 
 // In certain error conditions the tracee is damaged and we don't try to recover from that. We just
 // abort with a fatal log message. None of these errors are expected to occur in operation
@@ -62,7 +61,7 @@ void ExecuteOrDie(pid_t pid, RegisterState& original_registers, uint64_t address
   registers_set_rip.GetGeneralPurposeRegisters()->x86_64.rip = address_code;
   auto result_restore_registers = registers_set_rip.RestoreRegisters();
   if (result_restore_registers.has_error()) {
-    FATAL("Unable to continue tracee with PTRACE_CONT.");
+    FATAL("Unable to set registers in tracee: \"%s\"", result_restore_registers.error().message());
   }
   if (ptrace(PTRACE_CONT, pid, 0, 0) != 0) {
     FATAL("Unable to continue tracee with PTRACE_CONT.");
@@ -74,12 +73,38 @@ void ExecuteOrDie(pid_t pid, RegisterState& original_registers, uint64_t address
   }
 }
 
+// Returns the absolute virtual address of a function in a module of a process as
+// FindFunctionAddress does but accepts a fallback symbol if the primary one cannot be resolved.
+ErrorMessageOr<uint64_t> FindFunctionAddressWithFallback(pid_t pid, std::string_view function,
+                                                         std::string_view module,
+                                                         std::string_view fallback_function,
+                                                         std::string_view fallback_module) {
+  ErrorMessageOr<uint64_t> primary_address_or_error = FindFunctionAddress(pid, function, module);
+  if (primary_address_or_error.has_value()) {
+    return primary_address_or_error.value();
+  }
+  ErrorMessageOr<uint64_t> fallback_address_or_error =
+      FindFunctionAddress(pid, fallback_function, fallback_module);
+  if (fallback_address_or_error.has_value()) {
+    return fallback_address_or_error.value();
+  }
+
+  return ErrorMessage(absl::StrFormat(
+      "Failed to load symbol \"%s\" from module \"%s\" with error: \"%s\"\nAnd also "
+      "failed to load fallback symbol \"%s\" from module \"%s\" with error: \"%s\"",
+      function, module, primary_address_or_error.error().message(), fallback_function,
+      fallback_module, fallback_address_or_error.error().message()));
+}
+
 }  // namespace
+
+
 
 [[nodiscard]] ErrorMessageOr<void*> DlopenInTracee(pid_t pid, std::filesystem::path path,
                                                    uint32_t flag) {
   // Figure out address of dlopen in libc.
-  OUTCOME_TRY(address_dlopen, FindFunctionAddress(pid, "dlopen", "libc-"));
+  OUTCOME_TRY(address_dlopen, FindFunctionAddressWithFallback(pid, "dlopen", "libdl-",
+                                                              "__libc_dlopen_mode", "libc-"));
 
   // Backup registers.
   RegisterState original_registers;
@@ -87,27 +112,28 @@ void ExecuteOrDie(pid_t pid, RegisterState& original_registers, uint64_t address
 
   // Allocate small memory area in the tracee. This is used for the code and the path name.
   const uint64_t path_length = path.string().length() + 1;  // Include terminating zero.
-  const uint64_t memory_size = kScratchPadSize + path_length;
+  const uint64_t memory_size = kCodeScratchPadSize + path_length;
   OUTCOME_TRY(address_code, AllocateInTracee(pid, memory_size));
 
-  // Write the name of the .so into memory at address_code with offset of kScratchPadSize.
+  // Write the name of the .so into memory at address_code with offset of kCodeScratchPadSize.
   std::vector<uint8_t> path_as_vector(path_length);
   memcpy(path_as_vector.data(), path.c_str(), path_length);
-  const uint64_t address_so_path = address_code + kScratchPadSize;
+  const uint64_t address_so_path = address_code + kCodeScratchPadSize;
   auto result_write_path = WriteTraceesMemory(pid, address_so_path, path_as_vector);
   if (result_write_path.has_error()) {
     FreeMemoryOrDie(pid, address_code, memory_size);
     return result_write_path.error();
   }
+
   // We want to do the following in the tracee:
-  // return_value = dlopen(path, RTLD_LAZY);
+  // return_value = dlopen(path, flag);
   // The calling convention is to put the parameters in registers rdi and rsi.
-  // So the address of the file path goes to rdi. The flag argument goes into rsi (RTLD_LAZY is
-  // equal to 1). Then we load the address of dlopen into rax and do the call. Assembly in Intel
-  // syntax (destination first), machine code on the right:
+  // So the address of the file path goes to rdi. The flag argument goes into rsi. Then we load the
+  // address of dlopen into rax and do the call. Assembly in Intel syntax (destination first),
+  // machine code on the right:
 
   // movabs rax, address_so_path      48 b8 address_so_path
-  // mov rbx, 1                       48 c7 c3 01 00 00 00
+  // mov rbx, 2                       48 c7 c3 flag
   // mov rdi, rax                     48 8b f8
   // mov rsi, rbx                     48 8b f3
   // movabs rax, address_dlopen       48 b8 address_dlopen
@@ -142,7 +168,8 @@ void ExecuteOrDie(pid_t pid, RegisterState& original_registers, uint64_t address
 
 [[nodiscard]] ErrorMessageOr<void*> DlsymInTracee(pid_t pid, void* handle, std::string symbol) {
   // Figure out address of dlsym in libc.
-  OUTCOME_TRY(address_dlsym, FindFunctionAddress(pid, "dlsym", "libc-"));
+  OUTCOME_TRY(address_dlsym,
+              FindFunctionAddressWithFallback(pid, "dlsym", "libdl-", "__libc_dlsym", "libc-"));
 
   // Backup registers.
   RegisterState original_registers;
@@ -150,13 +177,13 @@ void ExecuteOrDie(pid_t pid, RegisterState& original_registers, uint64_t address
 
   // Allocate small memory area in the tracee. This is used for the code and the symbol name.
   const size_t symbol_name_length = symbol.length() + 1;  // include terminating zero
-  const uint64_t memory_size = kScratchPadSize + symbol_name_length;
+  const uint64_t memory_size = kCodeScratchPadSize + symbol_name_length;
   OUTCOME_TRY(address_code, AllocateInTracee(pid, memory_size));
 
-  // Write the name of symbol into memory at address_code with offset of kScratchPadSize.
+  // Write the name of symbol into memory at address_code with offset of kCodeScratchPadSize.
   std::vector<uint8_t> symbol_name_as_vector(symbol_name_length);
   memcpy(symbol_name_as_vector.data(), symbol.c_str(), symbol_name_length);
-  const uint64_t address_symbol_name = address_code + kScratchPadSize;
+  const uint64_t address_symbol_name = address_code + kCodeScratchPadSize;
   auto result_write_symbol_name =
       WriteTraceesMemory(pid, address_symbol_name, symbol_name_as_vector);
   if (result_write_symbol_name.has_error()) {
@@ -207,14 +234,15 @@ void ExecuteOrDie(pid_t pid, RegisterState& original_registers, uint64_t address
 
 [[nodiscard]] ErrorMessageOr<void> DlcloseInTracee(pid_t pid, void* handle) {
   // Figure out address of dlclose.
-  OUTCOME_TRY(address_dlclose, FindFunctionAddress(pid, "dlclose", "libc-"));
+  OUTCOME_TRY(address_dlclose,
+              FindFunctionAddressWithFallback(pid, "dlclose", "libdl-", "__libc_dlclose", "libc-"));
 
   // Backup registers.
   RegisterState original_registers;
   OUTCOME_TRY(original_registers.BackupRegisters(pid));
 
   // Allocate small memory area in the tracee.
-  OUTCOME_TRY(address_code, AllocateInTracee(pid, kScratchPadSize));
+  OUTCOME_TRY(address_code, AllocateInTracee(pid, kCodeScratchPadSize));
 
   // We want to do the following in the tracee:
   // dlclose(handle);
@@ -237,7 +265,7 @@ void ExecuteOrDie(pid_t pid, RegisterState& original_registers, uint64_t address
       .AppendBytes({0xcc});
   auto result_write_code = WriteTraceesMemory(pid, address_code, code.GetResultAsVector());
   if (result_write_code.has_error()) {
-    FreeMemoryOrDie(pid, address_code, kScratchPadSize);
+    FreeMemoryOrDie(pid, address_code, kCodeScratchPadSize);
     return result_write_code.error();
   }
 
@@ -249,9 +277,11 @@ void ExecuteOrDie(pid_t pid, RegisterState& original_registers, uint64_t address
 
   // Cleanup memory and registers.
   RestoreRegistersOrDie(original_registers);
-  FreeMemoryOrDie(pid, address_code, kScratchPadSize);
+  FreeMemoryOrDie(pid, address_code, kCodeScratchPadSize);
   return outcome::success();
 }
+
+
 
 ErrorMessageOr<uint64_t> FindFunctionAddress(pid_t pid, std::string_view function,
                                              std::string_view module) {
@@ -263,13 +293,13 @@ ErrorMessageOr<uint64_t> FindFunctionAddress(pid_t pid, std::string_view functio
   std::string module_file_path;
   uint64_t module_base_address = 0;
   for (const auto& m : modules.value()) {
-    if (absl::StrContains(m.name(), module)) {
+    if (absl::StartsWith(m.name(), module)) {
       module_file_path = m.file_path();
       module_base_address = m.address_start();
     }
   }
   if (module_file_path.empty()) {
-    return ErrorMessage(absl::StrFormat("There is no module %s in process %u.", module, pid));
+    return ErrorMessage(absl::StrFormat("There is no module %s in process %d.", module, pid));
   }
 
   auto elf_file = ElfFile::Create(module_file_path);
@@ -279,20 +309,18 @@ ErrorMessageOr<uint64_t> FindFunctionAddress(pid_t pid, std::string_view functio
 
   auto syms = elf_file.value()->LoadSymbolsFromDynsym();
   if (syms.has_error()) {
-    return ErrorMessage(absl::StrFormat("Failed to load symbols for module %s: %s", module,
+    return ErrorMessage(absl::StrFormat("Failed to load symbols for module \"%s\": %s", module,
                                         syms.error().message()));
   }
 
-  auto it = syms.value().symbol_infos().begin();
-  while (it != syms.value().symbol_infos().end()) {
-    if (absl::StrContains(it->name(), function)) {
-      return it->address() + module_base_address - syms.value().load_bias();
+  for (const auto& sym : syms.value().symbol_infos()) {
+    if (sym.name() == function) {
+      return sym.address() + module_base_address - syms.value().load_bias();
     }
-    it++;
   }
 
-  return ErrorMessage(
-      absl::StrFormat("Unable to locate function symbol %s in module %s.", function, module));
+  return ErrorMessage(absl::StrFormat("Unable to locate function symbol \"%s\" in module \"%s\".",
+                                      function, module));
 }
 
 }  // namespace orbit_user_space_instrumentation
--- a/src/UserSpaceInstrumentation/InjectLibraryInTracee.cpp
+++ b/src/UserSpaceInstrumentation/InjectLibraryInTracee.cpp
@@ -1,0 +1,301 @@
+// Copyright (c) 2021 The Orbit Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#include "InjectLibraryInTracee.h"
+
+#include <absl/base/casts.h>
+#include <absl/strings/match.h>
+#include <absl/strings/str_format.h>
+#include <dlfcn.h>
+#include <sys/ptrace.h>
+#include <sys/wait.h>
+
+#include <csignal>
+#include <string>
+#include <vector>
+
+#include "AccessTraceesMemory.h"
+#include "AllocateInTracee.h"
+#include "ElfUtils/ElfFile.h"
+#include "ElfUtils/LinuxMap.h"
+#include "MachineCode.h"
+#include "OrbitBase/Logging.h"
+#include "UserSpaceInstrumentation/RegisterState.h"
+
+namespace orbit_user_space_instrumentation {
+
+namespace {
+
+using orbit_elf_utils::ElfFile;
+using orbit_elf_utils::ReadModules;
+
+// Size of the small amount of memory we need in the tracee to write machine code into.
+constexpr uint64_t kScratchPadSize = 1024;
+
+// In certain error conditions the tracee is damaged and we don't try to recover from that. We just
+// abort with a fatal log message. None of these errors are expected to occur in operation
+// obvioulsy. That's what the *OrDie methods below are for.
+void FreeMemoryOrDie(pid_t pid, uint64_t address_code, uint64_t size) {
+  auto result = FreeInTracee(pid, address_code, size);
+  FAIL_IF(result.has_error(), "Unable to free previously allocated memory in tracee: \"%s\"",
+          result.error().message());
+}
+
+void RestoreRegistersOrDie(RegisterState& register_state) {
+  auto result = register_state.RestoreRegisters();
+  FAIL_IF(result.has_error(), "Unable to restore register state in tracee: \"%s\"",
+          result.error().message());
+}
+
+uint64_t GetReturnValueOrDie(pid_t pid) {
+  RegisterState return_value_registers;
+  auto result_return_value = return_value_registers.BackupRegisters(pid);
+  FAIL_IF(result_return_value.has_error(), "Unable to read registers after function called :\"%s\"",
+          result_return_value.error().message());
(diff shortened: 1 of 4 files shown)


src/UserSpaceInstrumentation/InjectLibraryInTracee.cpp
```diff
@@ -0,0 +1,298 @@
...
+uint64_t GetReturnValueOrDie(pid_t pid) {
+  RegisterState return_value_registers;
+  auto result_return_value = return_value_registers.BackupRegisters(pid);
+  FAIL_IF(result_return_value.has_error(), "Unable to read registers after function called :\"%s\"",
+          result_return_value.error().message());
+  return return_value_registers.GetGeneralPurposeRegisters()->x86_64.rax;
+}
+
+// Execute the code at `address_code`. The code segment has to end with an `int3`.
+void ExecuteOrDie(pid_t pid, RegisterState& original_registers, uint64_t address_code) {
```
dpallotti, line 60: Shouldn't it be `code_address`?

src/UserSpaceInstrumentation/InjectLibraryInTracee.cpp
```diff
@@ -0,0 +1,298 @@
...
+  RegisterState registers_set_rip = original_registers;
+  registers_set_rip.GetGeneralPurposeRegisters()->x86_64.rip = address_code;
+  auto result_restore_registers = registers_set_rip.RestoreRegisters();
+  if (result_restore_registers.has_error()) {
+    FATAL("Unable to continue tracee with PTRACE_CONT.");
```
dpallotti, line 65: Do you want to also print `result_restore_registers.error()`?

src/UserSpaceInstrumentation/InjectLibraryInTracee.cpp
```diff
@@ -0,0 +1,298 @@
...
+  // Figure out address of dlopen in libc.
+  OUTCOME_TRY(address_dlopen, FindFunctionAddress(pid, "dlopen", "libc-"));
+
+  // Backup registers.
+  RegisterState original_registers;
+  OUTCOME_TRY(original_registers.BackupRegisters(pid));
+
+  // Allocate small memory area in the tracee. This is used for the code and the path name.
+  const uint64_t path_length = path.string().length() + 1;  // Include terminating zero.
+  const uint64_t memory_size = kScratchPadSize + path_length;
```
dpallotti, line 90: As I was a bit surprised that the actual allocated size was not `kScratchPadSize` but `kScratchPadSize + something`, I suggest renaming `kScratchPadSize` to `kCodeScratchPadSize` or similar.

src/UserSpaceInstrumentation/InjectLibraryInTracee.cpp
```diff
@@ -0,0 +1,298 @@
...
+  OUTCOME_TRY(address_code, AllocateInTracee(pid, memory_size));
+
+  // Write the name of the .so into memory at address_code with offset of kScratchPadSize.
+  std::vector<uint8_t> path_as_vector(path_length);
+  memcpy(path_as_vector.data(), path.c_str(), path_length);
+  const uint64_t address_so_path = address_code + kScratchPadSize;
```
dpallotti, line 96: `so_path_address`?

src/UserSpaceInstrumentation/InjectLibraryInTracee.cpp
```diff
@@ -0,0 +1,298 @@
...
+  auto result_write_path = WriteTraceesMemory(pid, address_so_path, path_as_vector);
+  if (result_write_path.has_error()) {
+    FreeMemoryOrDie(pid, address_code, memory_size);
+    return result_write_path.error();
+  }
+  // We want to do the following in the tracee:
```
dpallotti, line 102:
```suggestion
  }

  // We want to do the following in the tracee:
```

src/UserSpaceInstrumentation/InjectLibraryInTracee.cpp
```diff
@@ -0,0 +1,298 @@
...
+  // return_value = dlopen(path, RTLD_LAZY);
+  // The calling convention is to put the parameters in registers rdi and rsi.
+  // So the address of the file path goes to rdi. The flag argument goes into rsi (RTLD_LAZY is
+  // equal to 1). Then we load the address of dlopen into rax and do the call. Assembly in Intel
+  // syntax (destination first), machine code on the right:
+
+  // movabs rax, address_so_path      48 b8 address_so_path
+  // mov rbx, 1                       48 c7 c3 01 00 00 00
+  // mov rdi, rax                     48 8b f8
+  // mov rsi, rbx                     48 8b f3
```
dpallotti, line 112: What's the reason for loading into `rax` and `rbx` just to immediately copy their content into `rdi` and `rsi` (and also immediately overwriting `rax`)? Can't you load directly into `rdi` and `rsi`?

src/UserSpaceInstrumentation/InjectLibraryInTracee.cpp
```diff
@@ -0,0 +1,298 @@
...
+  if (result_write_code.has_error()) {
+    FreeMemoryOrDie(pid, address_code, memory_size);
+    return result_write_code.error();
+  }
+
+  ExecuteOrDie(pid, original_registers, address_code);
+
+  void* return_value = absl::bit_cast<void*>(GetReturnValueOrDie(pid));
+
+  // Cleanup memory and registers.
```
dpallotti, line 137:
```suggestion
  // Clean up memory and registers.
```

src/UserSpaceInstrumentation/InjectLibraryInTracee.cpp
```diff
@@ -0,0 +1,298 @@
...
+  RestoreRegistersOrDie(original_registers);
+  FreeMemoryOrDie(pid, address_code, memory_size);
+  return return_value;
+}
+
+[[nodiscard]] ErrorMessageOr<void*> DlsymInTracee(pid_t pid, void* handle, std::string symbol) {
```
dpallotti, line 143: I'm wondering if it would make sense and if it would be possible to extract some code from `DlopenInTracee`, `DlsymInTracee` and `DlcloseInTracee`, as there is a certain degree of code duplication.

src/UserSpaceInstrumentation/InjectLibraryInTracee.cpp
```diff
@@ -0,0 +1,298 @@
...
+  // return_value = dlsym(handle, symbol);
+  // The calling convention is to put the parameters in registers rdi and rsi.
+  // So the handle goes to rdi and the address of the symbol name goes to rsi. Then we load the
+  // address of dlsym into rax and do the call. Assembly in Intel syntax (destination first),
+  // machine code on the right:
+
+  // movabs rax, handle               48 b8 handle
+  // mov rdi, rax                     48 89 c7
+  // movabs rax, address_symbol_name  48 b8 address_symbol_name
+  // mov rsi, rax                     48 89 c6
```
dpallotti, line 177: Same question as above: can't you write `rdi` and `rsi` directly instead of going through `rax`?
Also, the pattern here is different than `DlopenInTracee`, as you reuse `rax` here, while above you use `rax` and `rbx`. I would try to be consistent.

src/UserSpaceInstrumentation/InjectLibraryInTracee.cpp
```diff
@@ -0,0 +1,298 @@
...
+  OUTCOME_TRY(address_code, AllocateInTracee(pid, kScratchPadSize));
+
+  // We want to do the following in the tracee:
+  // dlclose(handle);
+  // The calling convention is to put the parameter in registers rdi. Then we load the address of
+  // address_dlclose into rax and do the call. Assembly in Intel syntax (destination first), machine
+  // code on the right:
+
+  // movabs rax, handle               48 b8 handle
+  // mov rdi, rax                     48 8b f8
```
dpallotti, line 226: Well, again, can you not do `movabs rdi handle`?

src/UserSpaceInstrumentation/InjectLibraryInTracee.cpp
```diff
@@ -0,0 +1,298 @@
...
+  std::string module_file_path;
+  uint64_t module_base_address = 0;
+  for (const auto& m : modules.value()) {
+    if (absl::StrContains(m.name(), module)) {
+      module_file_path = m.file_path();
+      module_base_address = m.address_start();
+    }
+  }
+  if (module_file_path.empty()) {
+    return ErrorMessage(absl::StrFormat("There is no module %s in process %u.", module, pid));
```
dpallotti, line 272:
```suggestion
    return ErrorMessage(absl::StrFormat("There is no module \"%s\" in process %d.", module, pid));
```

`pid` is signed.

src/UserSpaceInstrumentation/InjectLibraryInTracee.cpp
```diff
@@ -0,0 +1,298 @@
...
+  }
+
+  auto elf_file = ElfFile::Create(module_file_path);
+  if (elf_file.has_error()) {
+    return elf_file.error();
+  }
+
+  auto syms = elf_file.value()->LoadSymbolsFromDynsym();
+  if (syms.has_error()) {
+    return ErrorMessage(absl::StrFormat("Failed to load symbols for module %s: %s", module,
```
dpallotti, line 282:
```suggestion
    return ErrorMessage(absl::StrFormat("Failed to load symbols for module \"%s\": %s", module,
```

src/UserSpaceInstrumentation/InjectLibraryInTracee.cpp
```diff
@@ -0,0 +1,298 @@
...
+  auto it = syms.value().symbol_infos().begin();
+  while (it != syms.value().symbol_infos().end()) {
+    if (absl::StrContains(it->name(), function)) {
+      return it->address() + module_base_address - syms.value().load_bias();
+    }
+    it++;
+  }
+
+  return ErrorMessage(
+      absl::StrFormat("Unable to locate function symbol %s in module %s.", function, module));
```
dpallotti, line 295:
```suggestion
      absl::StrFormat("Unable to locate function symbol \"%s\" in module \"%s\".", function, module));
```

src/UserSpaceInstrumentation/InjectLibraryInTracee.cpp
```diff
@@ -0,0 +1,298 @@
...
+                                        syms.error().message()));
+  }
+
+  auto it = syms.value().symbol_infos().begin();
+  while (it != syms.value().symbol_infos().end()) {
+    if (absl::StrContains(it->name(), function)) {
+      return it->address() + module_base_address - syms.value().load_bias();
+    }
+    it++;
+  }
```
dpallotti, line 292: Can't this be a foreach loop?

dpallotti (commented):
Looks very good overall. Some minor comments.

src/UserSpaceInstrumentation/InjectLibraryInTracee.cpp
```diff
@@ -0,0 +1,298 @@
...
+  std::vector<uint8_t> path_as_vector(path_length);
+  memcpy(path_as_vector.data(), path.c_str(), path_length);
+  const uint64_t address_so_path = address_code + kScratchPadSize;
+  auto result_write_path = WriteTraceesMemory(pid, address_so_path, path_as_vector);
+  if (result_write_path.has_error()) {
+    FreeMemoryOrDie(pid, address_code, memory_size);
+    return result_write_path.error();
+  }
+  // We want to do the following in the tracee:
+  // return_value = dlopen(path, RTLD_LAZY);
```
dimitry-, line 103: nit: use RTLD_NOW

src/UserSpaceInstrumentation/InjectLibraryInTracee.cpp
```diff
@@ -0,0 +1,298 @@
+// Copyright (c) 2021 The Orbit Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#include "InjectLibraryInTracee.h"
+
+#include <absl/base/casts.h>
+#include <absl/strings/match.h>
+#include <absl/strings/str_format.h>
+#include <dlfcn.h>
```
dimitry-, line 10: is this needed?

src/UserSpaceInstrumentation/InjectLibraryInTracee.cpp
```diff
@@ -0,0 +1,298 @@
...
+    FATAL("Failed to wait for sigtrap after PTRACE_CONT.");
+  }
+}
+
+}  // namespace
+
+[[nodiscard]] ErrorMessageOr<void*> DlopenInTracee(pid_t pid, std::filesystem::path path,
+                                                   uint32_t flag) {
+  // Figure out address of dlopen in libc.
+  OUTCOME_TRY(address_dlopen, FindFunctionAddress(pid, "dlopen", "libc-"));
```
dimitry-, line 82: interesting, not libdl?

src/UserSpaceInstrumentation/InjectLibraryInTracee.cpp
```diff
@@ -0,0 +1,298 @@
...
+
+  auto syms = elf_file.value()->LoadSymbolsFromDynsym();
+  if (syms.has_error()) {
+    return ErrorMessage(absl::StrFormat("Failed to load symbols for module %s: %s", module,
+                                        syms.error().message()));
+  }
+
+  auto it = syms.value().symbol_infos().begin();
+  while (it != syms.value().symbol_infos().end()) {
+    if (absl::StrContains(it->name(), function)) {
```
dimitry-, line 288: I think in this case we should lookup by exact name. I see now why it works. This is because libc has a private symbol which contains dlopen but is not dlopen.

src/UserSpaceInstrumentation/InjectLibraryInTracee.cpp
```diff
@@ -0,0 +1,298 @@
...
+                                             std::string_view module) {
+  auto modules = ReadModules(pid);
+  if (modules.has_error()) {
+    return modules.error();
+  }
+
+  std::string module_file_path;
+  uint64_t module_base_address = 0;
+  for (const auto& m : modules.value()) {
+    if (absl::StrContains(m.name(), module)) {
```
dimitry-, line 266: Maybe go with starts_with here? if we search for libdl.so it should find any version of it.

src/UserSpaceInstrumentation/InjectLibraryInTracee.h
```diff
@@ -0,0 +1,36 @@
...
+#include <filesystem>
+#include <string_view>
+
+#include "OrbitBase/Result.h"
+
+namespace orbit_user_space_instrumentation {
+
+// Open, use, and close dynamic library in the tracee. The functions here resemble the respective
+// functions offered by libdl as documented e.g. here: https://linux.die.net/man/3/dlopen. We rely
+// on libc being loaded into the tracee.
```
dimitry-, line 20: libdl? Also since we currntly are not relying on libdl's dlopen but rather on libc's private '__libc_dlopen_mode@@GLIBC_PRIVATE' symbol, is it guaranteed to work if libdl is not loaded?

dimitry- (commented):
some preliminary questions.

src/UserSpaceInstrumentation/InjectLibraryInTracee.cpp
danielfenner: It's pretty much the same I think. The code has an address which is the address of the code.

src/UserSpaceInstrumentation/InjectLibraryInTracee.cpp
danielfenner: Jep, also the string is just copied from the error below. It should mention setting the registers not PTRACE_CONT. Done.

src/UserSpaceInstrumentation/InjectLibraryInTracee.cpp
danielfenner: done

src/UserSpaceInstrumentation/InjectLibraryInTracee.cpp
danielfenner: Same as above. "Address of .so Path"

src/UserSpaceInstrumentation/InjectLibraryInTracee.cpp
danielfenner: done

src/UserSpaceInstrumentation/InjectLibraryInTracee.cpp
danielfenner: My understanding is that there is simply no machine code for this. One only can move 64 bit immediates to eax.

src/UserSpaceInstrumentation/InjectLibraryInTracee.cpp
danielfenner: done

src/UserSpaceInstrumentation/InjectLibraryInTracee.cpp
danielfenner: done

src/UserSpaceInstrumentation/InjectLibraryInTracee.cpp
danielfenner: done

src/UserSpaceInstrumentation/InjectLibraryInTracee.h
dimitry-: (answering my own question: __libc_dlopen_mode calls into ld-linux.so and does not need libdl.so)

src/UserSpaceInstrumentation/InjectLibraryInTracee.cpp
dpallotti: That's the point, in English it's either "address of the code" or "code address", "address code" means something else and it's weird to read.

src/UserSpaceInstrumentation/InjectLibraryInTracee.cpp
dpallotti: Same as above, either `address_of_so_path` or `so_path_address`, in English "address so path" reads wrong.

src/UserSpaceInstrumentation/InjectLibraryInTracee.cpp
danielfenner: Yes, I spend more time on this then I should have. 
tldr: The repeated field is stored in a class RepeatedField and range based loops work fine for classes that implement begin and end. ("6.5.4 The range-based for statement")

src/UserSpaceInstrumentation/InjectLibraryInTracee.cpp
danielfenner: Changed the comment to what actually happens in the code: we forward the flag parameter here. 

Changed to using RTLD_NOW in the test as discussed offline.

src/UserSpaceInstrumentation/InjectLibraryInTracee.cpp
danielfenner: No, left over from previous version ... removed.

src/UserSpaceInstrumentation/InjectLibraryInTracee.cpp
danielfenner: As discussed offline: We now try libdl first and use libc as a fallback. I neither is there we return an error.

src/UserSpaceInstrumentation/InjectLibraryInTracee.cpp
danielfenner: As discussed: I changed this to match the symbol exactly and the module needs to start with the given string to allow for different versions to be matched.

src/UserSpaceInstrumentation/InjectLibraryInTracee.h
danielfenner: As dicuseed offline ...

src/UserSpaceInstrumentation/InjectLibraryInTracee.cpp
danielfenner: see above.

src/UserSpaceInstrumentation/InjectLibraryInTracee.cpp
danielfenner: Sorry, I really don't get it. "address of the code" and "address code" are identical to me. 

What happens in my brain when I read "address_code": It's an adress! What might it be the address of? Ah, the code!
A bit like a subscript in a math symbol: F_gravity.

dpallotti: Ah, I get it now. For me this is weird and something that I haven't really seen done purposefully, especially not in Orbit.
The name of the variable (excluding special names like `i`) should read like English, where the attributive noun (the noun that acts like an adjective) precedes the noun, hence "code address" = "address of (the) code". "address code" does have a meaning and it's the opposite, that is, "code of the address". The underscore here is not indicating a subscript, but a space.
The same applies to all the `result_` variables in the end, but with "result" it's naturally clearer (and also they're "just" local variables). But yes, personally I would also write those differently.
I can see that I'm not changing your mind but again, I think it's not good practice to sometimes not follow English rules in parameter names (and I don't really get the criterion).

dpallotti: To clarify further, here's a question: Why is the function below called `FindFunctionAddress`? Based on your reasoning it should be called `FindAddressFunction`, but of course that would make no sense. So I don't understand why there you almost always apply English rules but then here you don't.

src/UserSpaceInstrumentation/InjectLibraryInTracee.h
```diff
@@ -0,0 +1,36 @@
...
+[[nodiscard]] ErrorMessageOr<void*> DlsymInTracee(pid_t pid, void* handle, std::string symbol);
+[[nodiscard]] ErrorMessageOr<void> DlcloseInTracee(pid_t pid, void* handle);
+
+// Returns the absolute virtual address of a function in a module of a process as resolved by the
+// dynsym section of the file that module is associated with.
+// The function name has to match the symbol name exactly. The module name needs start with the
+// string given to the function; this is done to allow for different versions of a library to be
+// matched.
+[[nodiscard]] ErrorMessageOr<uint64_t> FindFunctionAddress(pid_t pid, std::string_view function,
+                                                           std::string_view module);
```
dpallotti, line 32: I suggest renaming `module` to `module_prefix`. This makes it clearer that the function name is matched exactly but the library names is matched by prefix.

src/UserSpaceInstrumentation/InjectLibraryInTracee.cpp
```diff
@@ -0,0 +1,322 @@
...
+
+  // We want to do the following in the tracee:
+  // return_value = dlopen(path, flag);
+  // The calling convention is to put the parameters in registers rdi and rsi.
+  // So the address of the file path goes to rdi. The flag argument goes into rsi. Then we load the
+  // address of dlopen into rax and do the call. Assembly in Intel syntax (destination first),
+  // machine code on the right:
+
+  // movabs rax, address_so_path      48 b8 address_so_path
+  // mov rbx, 2                       48 c7 c3 flag
```
dpallotti, line 134: So by now I get that you cannot do `movabs rdi address_so_path`, but can't you do `mov rsi 2` instead of going through `rbx`?

src/UserSpaceInstrumentation/InjectLibraryInTracee.cpp
dpallotti: I think you've answered this.

src/UserSpaceInstrumentation/InjectLibraryInTracee.cpp
dpallotti: Are you going to address this?

src/UserSpaceInstrumentation/InjectLibraryInTracee.cpp
dpallotti: Have you seen this comment?

src/UserSpaceInstrumentation/InjectLibraryInTracee.cpp
```diff
@@ -0,0 +1,322 @@
...
+  std::string module_file_path;
+  uint64_t module_base_address = 0;
+  for (const auto& m : modules.value()) {
+    if (absl::StartsWith(m.name(), module)) {
+      module_file_path = m.file_path();
+      module_base_address = m.address_start();
+    }
+  }
+  if (module_file_path.empty()) {
+    return ErrorMessage(absl::StrFormat("There is no module %s in process %d.", module, pid));
```
dpallotti, line 298:
```suggestion
    return ErrorMessage(absl::StrFormat("There is no module \"%s\" in process %d.", module, pid));
```

dpallotti (approved):
There were still a couple of minor comments from the initial review, maybe you missed those. I added a couple of other minor comments. It seems that we don't agree on variable names and I'm still a bit puzzled but I guess I'm not changing your mind. Anyway, looks good to me!

src/UserSpaceInstrumentation/InjectLibraryInTracee.cpp
```diff
@@ -0,0 +1,322 @@
...
+      fallback_module, fallback_address_or_error.error().message()));
+}
+
+}  // namespace
+
+[[nodiscard]] ErrorMessageOr<void*> DlopenInTracee(pid_t pid, std::filesystem::path path,
+                                                   uint32_t flag) {
+  // Figure out address of dlopen in libc.
+  OUTCOME_TRY(address_dlopen, FindFunctionAddressWithFallback(pid, "dlopen", "libdl-",
+                                                              "__libc_dlopen_mode", "libc-"));
```
dimitry-, line 105: why is there a dash in the library name?

src/UserSpaceInstrumentation/InjectLibraryInTracee.h
dimitry-: and maybe rename function to function_name

dimitry- (approved):
otherwise lgtm (I am really curious about dash, I expected the name to be something like libdl.so.6)

src/UserSpaceInstrumentation/InjectLibraryInTracee.cpp
dimitry-: ok, looks like I have an answer:
```
$ ls -la /usr/lib/x86_64-linux-gnu/libc.so.6   
lrwxrwxrwx 1 root root 12 Jan 14  2018 /usr/lib/x86_64-linux-gnu/libc.so.6 -> libc-2.24.so
```

src/UserSpaceInstrumentation/InjectLibraryInTracee.cpp
```diff
@@ -0,0 +1,322 @@
...
+                                             std::string_view module) {
+  auto modules = ReadModules(pid);
+  if (modules.has_error()) {
+    return modules.error();
+  }
+
+  std::string module_file_path;
+  uint64_t module_base_address = 0;
+  for (const auto& m : modules.value()) {
+    if (absl::StartsWith(m.name(), module)) {
```
dimitry-, line 292: I still a bit worry about false positives here. "libc-core.so" for example. Maybe add a todo to replace it with regexp or keep iterating libraries if we have not found a function in this one?

dimitry-: I have a fix for this - but it needs to go to ElfFile.cpp, you can merge it as is I will do a followup PR

dimitry-: with #2070 module name will be set to soname which will let you to do exact comparison for module name with libdl.so and libc.so.

src/UserSpaceInstrumentation/InjectLibraryInTracee.cpp
dimitry-: ok, I merged the change to module name now we can use "libdl.so"/"libc.so" here (it should now work for all versions).

src/UserSpaceInstrumentation/InjectLibraryInTracee.cpp
danielfenner: Hmm, it really appears totally naturally to me to use it this way around. I would probably not use FindAddressFunction because there is a verb and I would not omit the "Of" then. My code is probably full of these (as you mentioned: result_*). I would honestly never have come up with the idea that anyone would do it differently then adding suffixes to disambiguate different entities. No clue what to make out of this.

src/UserSpaceInstrumentation/InjectLibraryInTracee.h
danielfenner: done

src/UserSpaceInstrumentation/InjectLibraryInTracee.cpp
danielfenner: Sorry, missed this and some others. It seems not all comments show up under the "Conversation" tab. I only see the second one there. That's a bit of a mess.

danielfenner: done

src/UserSpaceInstrumentation/InjectLibraryInTracee.cpp
danielfenner: Sorry, again, this is not showing up in the "conversation" tab....

Yes, I thought about this but have not found a good way yet. There is a similar block of code in the test when calling into `TrivialFunction` and I'll need that whenever I call something in the injected lib.

I'll revisit this in a follow up PR if there is a sensible portion that is generic.

src/UserSpaceInstrumentation/InjectLibraryInTracee.cpp
danielfenner: Okay, I added a solution with a regex here already. I'll submit that and we can chat offline about  how to use module name.

src/UserSpaceInstrumentation/InjectLibraryInTracee.cpp
danielfenner: done

src/UserSpaceInstrumentation/InjectLibraryInTracee.cpp
danielfenner: I'll prepare a follow up PR for this.

src/UserSpaceInstrumentation/InjectLibraryInTracee.cpp
danielfenner: Okay, I digged deeper into this and it turns out the stuff I claimed above was rubbish: One can move immediate 64 bit values into all registers. The same is true for 32 bit values. The "2" in the comment needs to be changed to "flag".
I changed that here and also in the other places where I write machine code directly.

dpallotti: Nice, much cleaner.

PR #2062, opened 2021-03-12
grpc_protos: Replace string with bytes

dimitry-:
Passing strings in proto triggers UTF-8 validation which
we do not really need. This change replaces strings with bytes
to avoid triggering validation step.

Test: builds

### DIFF
--- a/src/GrpcProtos/capture.proto
+++ b/src/GrpcProtos/capture.proto
@@ -121,7 +121,8 @@ message FullCallstackSample {
 
 message InternedString {
   uint64 key = 1;
-  string intern = 2;
+  // This is a string, we use bytes to avoid UTF-8 validation.
+  bytes intern = 2;
 }
 
 message InternedTracepointInfo {
@@ -155,7 +156,8 @@ message FullGpuJob {
   uint64 amdgpu_sched_run_job_time_ns = 7;
   uint64 gpu_hardware_start_time_ns = 8;
   uint64 dma_fence_signaled_time_ns = 9;
-  string timeline = 10;
+  // This is a string, we use bytes to avoid UTF-8 validation.
+  bytes timeline = 10;
 }
 
 message GpuJob {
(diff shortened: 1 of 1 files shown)


ronaldfw:
It sounds like this is a performance improvement. Do you have a benchmark that confirms this and gives us an idea of how much we win?

pierricgimmig:
Hi @dimitry- , I would also like to see some benchmarks. Also, do you know if this will complicate python parsing in the Jupyter notebooks? In C++ the both byte and string are represented as std::string, but it seems it's different in Python: https://groups.google.com/g/protobuf/c/au6eQBkRT5s

dimitry-:
Sorry for delay, I do not have any benchmarks on this. Regarding type mapping between languages this is a table:
https://developers.google.com/protocol-buffers/docs/proto#scalar

I do not think it is going to be a problem to handle bytes as string in python or Java

ronaldfw:
> Sorry for delay, I do not have any benchmarks on this. Regarding type mapping between languages this is a table:
> https://developers.google.com/protocol-buffers/docs/proto#scalar
> 
> I do not think it is going to be a problem to handle bytes as string in python or Java

Yeah, I don't think this will be a problem in Python, but I can also try this out if needed. 

While it would be nice to get a benchmark, unless anyone has concerns that this change makes something more inconvenient somewhere, I'm also fine with skipping this. Maybe it doesn't help much, but it's highly unlikely this makes anything worse in terms of performance.

ronaldfw:
Maybe add comments that the `bytes` fields represent strings?

florian-kuebler (approved):

dimitry-:
> Maybe add comments that the `bytes` fields represent strings?

done

dpallotti (approved):
I'm also fine with this, but what do you think about using a single comment at the top of the file instead of a comment for each string? It's a bit repetitive at the moment. Unless we think that there could be `bytes` fields that do *not* represent strings (but it's not the case now, or is it?) and then it could get confusing.

dimitry-:
> I'm also fine with this, but what do you think about using a single comment at the top of the file instead of a comment for each string? It's a bit repetitive at the moment. Unless we think that there could be `bytes` fields that do _not_ represent strings (but it's not the case now, or is it?) and then it could get confusing.

I think having comment next to field is better for readability.

PR #2063, opened 2021-03-12
Return errors from Allocate/FreeInTracee

dimitry-:
Test: builds


danielfenner (approved):
Wow, thanks, that's a weird bug.

lgtm

PR #2065, opened 2021-03-12
Offer command line option to clear source paths mappings

beckerhe:
This is a prerequisite to fix the failing E2E test. It fails because a source paths mapping already exists, while test expects an error message to pop up.

The solution: We clear all existing source paths mappings prior to starting Orbit. This could be done by directly deleting keys from the registry, but the nicer solution is in this PR: Let Orbit offer a command line option to clear this setting which means we don't have to worry in the E2E tests how and where the mappings are persisted.

### DIFF
--- a/src/OrbitQt/main.cpp
+++ b/src/OrbitQt/main.cpp
@@ -49,6 +49,7 @@
 #include "OrbitVersion/OrbitVersion.h"
 #include "Path.h"
 #include "ProfilingTargetDialog.h"
+#include "SourcePathsMapping/MappingManager.h"
 #include "TargetConfiguration.h"
 #include "opengldetect.h"
 #include "orbitmainwindow.h"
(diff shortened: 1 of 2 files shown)


danielfenner (approved):
lgtm!

src/OrbitQt/main.cpp
```diff
@@ -66,6 +67,8 @@ ABSL_DECLARE_FLAG(bool, local);
 ABSL_DECLARE_FLAG(bool, devmode);
 ABSL_DECLARE_FLAG(bool, nodeploy);
 
+ABSL_FLAG(bool, clear_source_paths_mappings, false, "Clear all the stored sources paths mappings");
```
ronaldfw, line 70:
```suggestion
ABSL_FLAG(bool, clear_source_paths_mappings, false, "Clear all the stored source paths mappings");
```

(Which is also what you use in other places.)

beckerhe: Right. Thanks.

ronaldfw, line 70: I wonder if we should have a comment that states that the main (only?) intention for usage of this flag is as part of e2e tests?

beckerhe: A user-facing comment, or for us in the code?

ronaldfw: For us in the code.

beckerhe: I added a comment.

beckerhe: I will go ahead and merge this.

PR #2066, opened 2021-03-12
Make OrbitQt a library target

beckerhe:
Previously OrbitQt was an executable target (Orbit.exe) which makes
problems for tests as these tests can't just link against OrbitQt.

So this commit converts OrbitQt into a library target and moves the
main.cpp into a separate executable target called "Orbit".

Hence, the target OrbitQtTests can just link against OrbitQt.

Test: Ran unit tests.


vickyliu-go4it (approved):
Thanks for resolving this. LGTM!

src/OrbitQt/FilterPanelWidget.h
```diff
@@ -8,8 +8,6 @@
 #include <QFrame>
 #include <QWidget>
 
-#include "ui_FilterPanelWidget.h"
-
 namespace Ui {
 class FilterPanelWidget;
 }
```
antonrohr, line 13: Just to clarify. The best practice here is to include `ui_FilterPanelWidget.h` in FilterPanelWidget.cpp and in FilterPanelWidget.h this forward declaration (i hope this is the right term) aka
```
namespace Ui {
class FilterPanelWidget;
}
```
?

antonrohr (approved):
Thank you very very much! Especially since this blocked me on the target label stuff

src/OrbitQt/FilterPanelWidget.h
beckerhe: Yes, exactly. (Forward declaration is the right term)

beckerhe:
If the CI is happy, I will merge! Thank you both for the review!

PR #2067, opened 2021-03-12
Remove LoadTextureFromFile and stb

beckerhe:
Since the toolbar moved to Qt, "LoadTextureFromFile" has not been in use
anymore. Since this is the only reason why we require "stb", let's just
remove it.

Test: compiles


beckerhe:
@pierricgimmig Do you still need that functionality? If not, let's remove it.

pierricgimmig (approved):
Hi Henning, no it's not required anymore, if we ever need it again it will be easy to revive. LGTM

beckerhe:
> Hi Henning, no it's not required anymore, if we ever need it again it will be easy to revive. LGTM

Yep. That's what I thought as well! Thanks!

PR #2070, opened 2021-03-15
ElfUtils: Use soname (DT_SONAME) as module name if present

dimitry-:
The filename is often not the name of the library, we should use soname for it if present. 
This change add function to get a soname for DSO.

Sets module name to soname if it is preset and to file name if the soname is not present.

Test: run unit tests


src/ElfUtils/testdata/Makefile
```diff
@@ -8,8 +8,6 @@
 CC=clang
 PHONY=all clean
 
-all: line_info_test_binary line_info_test_binary_compressed test_library
-
```
beckerhe: This was probably not intended, was it?

src/ElfUtils/ElfFile.cpp
```diff
@@ -134,6 +138,42 @@ ElfFileImpl<ElfT>::ElfFileImpl(std::filesystem::path file_path,
...
+        }
+        dynamic_string_table.emplace(absl::bit_cast<const char*>(strtab_addr_or_error.get()));
+        break;
+      }
+      default:
+        break;
+    }
+  }
+  if (soname_offset.has_value() && dynamic_string_table.has_value()) {
+    soname_ = dynamic_string_table.value().data() + soname_offset.value();
```
beckerhe, line 173: I assume this works because `dynamic_string_table` is a null-separated list of strings?
What happens if `SONAME` is the laste entry in the list? Does LLVM ensure that `dynamic_string_table` ends with a null-character or should we also take the length into account?

src/ElfUtils/ElfFile.cpp
```diff
@@ -134,6 +138,42 @@ ElfFileImpl<ElfT>::ElfFileImpl(std::filesystem::path file_path,
   InitSections();
 }
 
+template <typename ElfT>
+void ElfFileImpl<ElfT>::InitDynamicEntries(const llvm::object::ELFFile<ElfT>* elf_file) {
+  std::optional<uint64_t> soname_offset;
+  std::optional<llvm::StringRef> dynamic_string_table;
```
beckerhe, line 144: Nit: These could be moved further down.

beckerhe (approved):
Otherwise LGTM

danielfenner (approved):
lgtm

src/ElfUtils/testdata/Makefile
dimitry-: it was, I figured in most cases we do not want to compile every single test library, and muscle memory is not really helping here.

src/ElfUtils/ElfFile.cpp
dimitry-: come to think - we should probably never trust that format is correct and consistent, I am adding boundary checks and null-termination check.

src/ElfUtils/testdata/Makefile
beckerhe: Okay. Makes sense. But in that case you should also remove `all` from the list of phony targets.

src/ElfUtils/ElfFile.cpp
dimitry-: done.

src/ElfUtils/ElfFile.cpp
dimitry-: done.

src/ElfUtils/testdata/Makefile
dimitry-: done.

src/ElfUtils/ElfFile.cpp
```diff
@@ -158,19 +159,38 @@ void ElfFileImpl<ElfT>::InitDynamicEntries(const llvm::object::ELFFile<ElfT>* el
...
+    // Check boundaries and that dynamic string table is null-terminated
+    if (soname_offset.value() > dynamic_string_table_size.value()) {
+      ERROR(
+          "Soname offset is out of bounds of the string table (file=\"%s\", offset=%u strtab "
+          "size=%u)",
+          file_path_.string(), soname_offset.value(), dynamic_string_table_size.value());
+      return;
+    }
+
+    if (dynamic_string_table_ptr.value()[dynamic_string_table_size.value() - 1] != '\0') {
```
dimitry-, line 187: actually this is also not safe... let me fix this as well.

dimitry-: done.

beckerhe (approved):
LGTM! Thanks!

PR #2071, opened 2021-03-15
Make Disassembly and SourceCode view non-modal

beckerhe:
As requested by users, the code views should be non-modal such that the user can interact with Orbit while one or multiple code views are open in the background or on a different screen.

To make this happen, the code viewer dialog needs to own all its required resources. And it needs to delete itself when the dialog is closed.

The former is solved by introducing `orbit_code_viewer::OwningDialog`, a variant of `orbit_code_viewer::Dialog` that owns all its resources.

The latter can solved by Qt which offers the attribute `Qt::WA_DeleteOnClose` for a classes inheriting from `QWidget`. To make this somehow safe to handle, I introduced a free function `OpenAndDeleteOnClose`. It takes a unique pointer to an OwningDialog by value. It will open the dialog, ensure it will be deleted on close, and will release the unique pointer.


florian-kuebler (approved):
LGTM. Thanks!

beckerhe:
Thanks Florian! I just rebased. If the CI agrees I'm going to merge soon.

PR #2073, opened 2021-03-15
Fix unit test InjectLibraryInTraceeTest.

danielfenner:


src/UserSpaceInstrumentation/InjectLibraryInTraceeTest.cpp
```diff
@@ -38,10 +38,10 @@ TEST(InjectLibraryInTraceeTest, FindFunctionAddress) {
   // Stop the child process using our tooling.
   ASSERT_TRUE(AttachAndStopProcess(pid).has_value());
 
-  auto function_address_or_error = FindFunctionAddress(pid, "printf", "libc-");
+  auto function_address_or_error = FindFunctionAddress(pid, "printf", "libc");
   EXPECT_TRUE(function_address_or_error.has_value());
 
-  function_address_or_error = FindFunctionAddress(pid, "NOT_A_SYMBOL", "libc-");
```
dimitry-, line 44: well, as I mentioned in original PR, you can now safely use libc.so here no need to do contains/startswith anymore.

danielfenner: If I'm not totally mistaken the sonames are "libc.so.6" and "libdl.so.2". Let's have a look tomorrow.

PR #2074, opened 2021-03-15
Compare FunctionInfo by address and module_path

dimitry-:
We are going to end up in the situation where function_info is not fully
populated for instrumented function when loading capture from current file version.

This is done to preserve compatibility until flip a switch to new
capture format.

Test: run OrbitClientDataTests


florian-kuebler (approved):
A bit unfortunate that we need to go this path, but thanks for fixing it. LGTM.

PR #2076, opened 2021-03-16
Add timeline visualization for the memory information

vickyliu-go4it:
We would like to add the timeline visualization for the system memory usage information and make it obvious when the used memory exceeds the production limit threshold.

This PR includes:
* Processing the `SystemMemoryUsage` events received on the Orbit Client side and encoding it into a `TimerInfo` with the type of `kSystemUsage`;
* Adding `warning_threshold_`, `value_upper_bound_` and some relevant methods to the `GraphTrack` so we can visualize those values when they are provided;
* Create three memory graph tracks (i.e., "Memory used (kb)", "Memory buffers or cached (kb)" and "Memory unused") with total memory size as `value_upper_bound_` and the production limit as `warning_threshold_`;
* Draw text box and line to visualize the production limit and draw text box to visualize the total memory size;
* Update the CaptureOptionsDialog for supporting input the memory tracing parameters.

Other things might be done in this PR:
* Removing the "Memory Unused" track given the fact that Memory Total = Used + Unused + Buffers or cached;
* Enable hiding & showing memory tracks;
* (If not ship) Hide the memory tracing behind the flag.

Bugs: http://b/178999638, http://b/179000848.

After the change:
Screenshot 2021-03-16 115941
Screenshot 2021-03-16 120005

### DIFF
--- a/src/OrbitCaptureClient/CaptureClient.cpp
+++ b/src/OrbitCaptureClient/CaptureClient.cpp
@@ -60,13 +60,15 @@ static InstrumentedFunction::FunctionType InstrumentedFunctionTypeFromOrbitType(
   UNREACHABLE();
 }
 
+// TODO(vickyliu): This method contains a lot of arguments. Consider making it more structured.
 Future<ErrorMessageOr<CaptureListener::CaptureOutcome>> CaptureClient::Capture(
     ThreadPool* thread_pool, const ProcessData& process,
     const orbit_client_data::ModuleManager& module_manager,
     absl::flat_hash_map<uint64_t, FunctionInfo> selected_functions,
     TracepointInfoSet selected_tracepoints, absl::flat_hash_set<uint64_t> frame_track_function_ids,
     double samples_per_second, UnwindingMethod unwinding_method, bool collect_thread_state,
-    bool enable_introspection, uint64_t max_local_marker_depth_per_command_buffer) {
+    bool enable_introspection, uint64_t max_local_marker_depth_per_command_buffer,
+    bool collect_memory_info, uint64_t memory_sampling_period_ns) {
   absl::MutexLock lock(&state_mutex_);
   if (state_ != State::kStopped) {
     return {
@@ -87,11 +89,13 @@ Future<ErrorMessageOr<CaptureListener::CaptureOutcome>> CaptureClient::Capture(
        selected_functions = std::move(selected_functions), selected_tracepoints,
        frame_track_function_ids = std::move(frame_track_function_ids), collect_thread_state,
        samples_per_second, unwinding_method, enable_introspection,
-       max_local_marker_depth_per_command_buffer]() mutable {
+       max_local_marker_depth_per_command_buffer, collect_memory_info,
+       memory_sampling_period_ns]() mutable {
         return CaptureSync(std::move(process), module_manager, std::move(selected_functions),
                            std::move(selected_tracepoints), std::move(frame_track_function_ids),
                            samples_per_second, unwinding_method, collect_thread_state,
-                           enable_introspection, max_local_marker_depth_per_command_buffer);
+                           enable_introspection, max_local_marker_depth_per_command_buffer,
+                           collect_memory_info, memory_sampling_period_ns);
       });
 
   return capture_result;
@@ -102,7 +106,8 @@ ErrorMessageOr<CaptureListener::CaptureOutcome> CaptureClient::CaptureSync(
     const absl::flat_hash_map<uint64_t, FunctionInfo>& selected_functions,
     TracepointInfoSet selected_tracepoints, absl::flat_hash_set<uint64_t> frame_track_function_ids,
     double samples_per_second, UnwindingMethod unwinding_method, bool collect_thread_state,
-    bool enable_introspection, uint64_t max_local_marker_depth_per_command_buffer) {
+    bool enable_introspection, uint64_t max_local_marker_depth_per_command_buffer,
+    bool collect_memory_info, uint64_t memory_sampling_period_ns) {
   ORBIT_SCOPE_FUNCTION;
   writes_done_failed_ = false;
   try_abort_ = false;
@@ -128,6 +133,9 @@ ErrorMessageOr<CaptureListener::CaptureOutcome> CaptureClient::CaptureSync(
       capture_options->set_unwinding_method(CaptureOptions::kDwarf);
     }
   }
+
+  capture_options->set_collect_memory_info(collect_memory_info);
+  capture_options->set_memory_sampling_period_ns(memory_sampling_period_ns);
 
   capture_options->set_trace_thread_state(collect_thread_state);
   capture_options->set_trace_gpu_driver(true);
--- a/src/OrbitGl/App.cpp
+++ b/src/OrbitGl/App.cpp
@@ -967,12 +967,15 @@ void OrbitApp::StartCapture() {
   uint64_t max_local_marker_depth_per_command_buffer =
       data_manager_->max_local_marker_depth_per_command_buffer();
 
+  bool collect_memory_info = data_manager_->collect_memory_info();
+  uint64_t memory_sampling_period_ns = data_manager_->memory_sampling_period_ns();
+
   CHECK(capture_client_ != nullptr);
   Future<ErrorMessageOr<CaptureOutcome>> capture_result = capture_client_->Capture(
       thread_pool_.get(), *process, *module_manager_, std::move(selected_functions_map),
       std::move(selected_tracepoints), std::move(frame_track_function_ids), samples_per_second,
       unwinding_method, collect_thread_states, enable_introspection,
-      max_local_marker_depth_per_command_buffer);
+      max_local_marker_depth_per_command_buffer, collect_memory_info, memory_sampling_period_ns);
 
   capture_result.Then(main_thread_executor_, [this](ErrorMessageOr<CaptureOutcome> capture_result) {
     if (capture_result.has_error()) {
--- a/src/OrbitGl/App.cpp
+++ b/src/OrbitGl/App.cpp
@@ -284,6 +284,32 @@ void OrbitApp::OnTimer(const TimerInfo& timer_info) {
   frame_track_online_processor_.ProcessTimer(timer_info, func);
 }
 
+void OrbitApp::OnSystemMemoryUsage(
+    const orbit_grpc_protos::SystemMemoryUsage& system_memory_usage) {
+  TimerInfo timer_info;
+  timer_info.set_type(TimerInfo::kSystemMemoryUsage);
+  timer_info.set_start(system_memory_usage.timestamp_ns());
+  timer_info.set_end(system_memory_usage.timestamp_ns());
+
+  constexpr size_t kNumOfEncodedValues = 5;
+  timer_info.mutable_registers()->Reserve(kNumOfEncodedValues);
+  timer_info.mutable_registers()->Set(static_cast<size_t>(SystemMemoryUsageEncodingIndex::kTotalKb),
+                                      orbit_api::Encode<uint64_t>(system_memory_usage.total_kb()));
+  timer_info.mutable_registers()->Set(static_cast<size_t>(SystemMemoryUsageEncodingIndex::kFreeKb),
+                                      orbit_api::Encode<uint64_t>(system_memory_usage.free_kb()));
+  timer_info.mutable_registers()->Set(
+      static_cast<size_t>(SystemMemoryUsageEncodingIndex::kAvailableKb),
+      orbit_api::Encode<uint64_t>(system_memory_usage.available_kb()));
+  timer_info.mutable_registers()->Set(
+      static_cast<size_t>(SystemMemoryUsageEncodingIndex::kBuffersKb),
+      orbit_api::Encode<uint64_t>(system_memory_usage.buffers_kb()));
+  timer_info.mutable_registers()->Set(
+      static_cast<size_t>(SystemMemoryUsageEncodingIndex::kCachedKb),
+      orbit_api::Encode<uint64_t>(system_memory_usage.cached_kb()));
+
+  GetMutableTimeGraph()->ProcessTimer(timer_info, nullptr);
+}
+
 void OrbitApp::OnKeyAndString(uint64_t key, std::string str) {
   string_manager_.AddIfNotPresent(key, std::move(str));
 }
--- a/src/OrbitGl/App.h
+++ b/src/OrbitGl/App.h
@@ -367,6 +367,20 @@ class OrbitApp final : public DataViewFactory, public CaptureListener {
   void SetUnwindingMethod(orbit_grpc_protos::UnwindingMethod unwinding_method);
   void SetMaxLocalMarkerDepthPerCommandBuffer(uint64_t max_local_marker_depth_per_command_buffer);
 
+  void SetCollectMemoryInfo(bool collect_memory_info) {
+    data_manager_->set_collect_memory_info(collect_memory_info);
+  }
+  [[nodiscard]] bool GetCollectMemoryInfo() const { return data_manager_->collect_memory_info(); }
+  void SetMemorySamplingPeriodNs(uint64_t memory_sampling_period_ns) {
+    data_manager_->set_memory_sampling_period_ns(memory_sampling_period_ns);
+  }
+  void SetMemoryWarningThresholdKb(uint64_t memory_warning_threshold_kb) {
+    data_manager_->set_memory_warning_threshold_kb(memory_warning_threshold_kb);
+  }
+  [[nodiscard]] uint64_t GetMemoryWarningThresholdKb() const {
+    return data_manager_->memory_warning_threshold_kb();
+  }
+
   // TODO(kuebler): Move them to a separate controler at some point
   void SelectFunction(const orbit_client_protos::FunctionInfo& func);
   void DeselectFunction(const orbit_client_protos::FunctionInfo& func);
--- a/src/OrbitGl/App.h
+++ b/src/OrbitGl/App.h
@@ -71,6 +71,7 @@
 #include "SymbolHelper.h"
 #include "TextBox.h"
 #include "TracepointsDataView.h"
+#include "capture.pb.h"
 #include "capture_data.pb.h"
 #include "preset.pb.h"
 #include "services.pb.h"
@@ -146,6 +147,16 @@ class OrbitApp final : public DataViewFactory, public CaptureListener {
   void OnUniqueTracepointInfo(uint64_t key,
                               orbit_grpc_protos::TracepointInfo tracepoint_info) override;
   void OnTracepointEvent(orbit_client_protos::TracepointEventInfo tracepoint_event_info) override;
+
+  enum class SystemMemoryUsageEncodingIndex {
+    kTotalKb,
+    kFreeKb,
+    kAvailableKb,
+    kBuffersKb,
+    kCachedKb
+  };
+  void OnSystemMemoryUsage(
+      const orbit_grpc_protos::SystemMemoryUsage& system_memory_usage) override;
 
   void OnValidateFramePointers(std::vector<const ModuleData*> modules_to_validate);
 
--- a/src/OrbitGl/GraphTrack.cpp
+++ b/src/OrbitGl/GraphTrack.cpp
@@ -94,7 +94,8 @@ void GraphTrack::Draw(GlCanvas* canvas, PickingMode picking_mode, float z_offset
   float text_z = GlCanvas::kZValueEvent + z_offset;
 
   // Add warning threshold text box and line.
-  Batcher* batcher = canvas->GetBatcher();
+  Batcher* ui_batcher = canvas->GetBatcher();
+  uint32_t font_size = layout_->CalculateZoomedFontSize();
   if (warning_threshold_.has_value()) {
     const Color kThresholdColor(244, 67, 54, 255);
 
@@ -105,7 +106,6 @@ void GraphTrack::Draw(GlCanvas* canvas, PickingMode picking_mode, float z_offset
     Vec2 to(x + size_[0], y);
 
     std::string text = warning_threshold_.value().first;
-    uint32_t font_size = layout_->CalculateZoomedFontSize();
     float string_width = canvas->GetTextRenderer().GetStringWidth(text.c_str(), font_size);
     Vec2 text_box_size(string_width, layout_->GetTextBoxHeight());
     Vec2 text_box_position(pos_[0] + layout_->GetRightMargin(),
@@ -114,15 +114,15 @@ void GraphTrack::Draw(GlCanvas* canvas, PickingMode picking_mode, float z_offset
                                       text_box_position[1] + layout_->GetTextOffset(), text_z,
                                       kThresholdColor, font_size, text_box_size[0]);
 
-    batcher->AddLine(from, from + Vec2(layout_->GetRightMargin() / 2.f, 0), text_z,
-                     kThresholdColor);
-    batcher->AddLine(Vec2(text_box_position[0] + text_box_size[0], y), to, text_z, kThresholdColor);
+    ui_batcher->AddLine(from, from + Vec2(layout_->GetRightMargin() / 2.f, 0), text_z,
+                        kThresholdColor);
+    ui_batcher->AddLine(Vec2(text_box_position[0] + text_box_size[0], y), to, text_z,
+                        kThresholdColor);
   }
 
   // Add value upper bound text box (e.g., the "Memory Total" text box for the memory tracks).
   if (value_upper_bound_.has_value()) {
     std::string text = value_upper_bound_.value().first;
-    uint32_t font_size = layout_->CalculateZoomedFontSize();
     float string_width = canvas->GetTextRenderer().GetStringWidth(text.c_str(), font_size);
     Vec2 text_box_size(string_width, layout_->GetTextBoxHeight());
     Vec2 text_box_position(pos_[0] + size_[0] - text_box_size[0] - layout_->GetRightMargin() -
@@ -133,11 +133,25 @@ void GraphTrack::Draw(GlCanvas* canvas, PickingMode picking_mode, float z_offset
                                       kWhite, font_size, text_box_size[0]);
   }
 
+  // Add value lower bound text box.
+  if (value_lower_bound_.has_value()) {
+    std::string text = value_lower_bound_.value().first;
+    float string_width = canvas->GetTextRenderer().GetStringWidth(text.c_str(), font_size);
+    Vec2 text_box_size(string_width, layout_->GetTextBoxHeight());
+    Vec2 text_box_position(pos_[0] + size_[0] - text_box_size[0] - layout_->GetRightMargin() -
+                               layout_->GetSliderWidth(),
+                           pos_[1] - size_[1]);
+    canvas->GetTextRenderer().AddText(text.c_str(), text_box_position[0],
+                                      text_box_position[1] + layout_->GetTextOffset(), text_z,
+                                      kWhite, font_size, text_box_size[0]);
+  }
+
   // Draw label
   uint64_t current_mouse_time_ns = time_graph_->GetCurrentMouseTimeNs();
   auto previous_point = GetPreviousValueAndTime(current_mouse_time_ns);
   double value =
       previous_point.has_value() ? previous_point.value().second : values_.begin()->second;
+
   uint64_t first_time = values_.begin()->first;
   uint64_t label_time = std::max(current_mouse_time_ns, first_time);
   float point_x = time_graph_->GetWorldFromTick(label_time);
@@ -210,4 +224,28 @@ float GraphTrack::GetHeight() const {
   float height = layout_->GetTextBoxHeight() + layout_->GetSpaceBetweenTracksAndThread() +
                  layout_->GetEventTrackHeight() + layout_->GetTrackBottomMargin();
   return height;
+}
+
+void GraphTrack::SetWarningThresholdWhenEmpty(
+    const std::optional<std::pair<std::string, double>>& warning_threshold) {
+  if (warning_threshold_.has_value() || !warning_threshold.has_value()) return;
+  warning_threshold_ = warning_threshold;
+  UpdateMinAndMax(warning_threshold_.value().second);
+}
+
+void GraphTrack::SetValueUpperBoundWhenEmpty(
+    const std::optional<std::pair<std::string, double>>& value_upper_bound) {
+  if (value_upper_bound_.has_value() || !value_upper_bound.has_value()) return;
+  value_upper_bound_ = value_upper_bound;
+  UpdateMinAndMax(value_upper_bound_.value().second);
+}
+void GraphTrack::SetValueLowerBoundWhenEmpty(
+    const std::optional<std::pair<std::string, double>>& value_lower_bound) {
+  value_lower_bound_ = value_lower_bound;
+  UpdateMinAndMax(value_lower_bound_.value().second);
+}
+
+void GraphTrack::UpdateMinAndMax(double value) {
+  max_ = std::max(max_, value);
+  min_ = std::min(min_, value);
 }
--- a/src/OrbitGl/GraphTrack.cpp
+++ b/src/OrbitGl/GraphTrack.cpp
@@ -5,6 +5,7 @@
 #include "GraphTrack.h"
 
 #include <GteVector.h>
+#include <absl/strings/str_format.h>
 
 #include <algorithm>
 #include <cstdint>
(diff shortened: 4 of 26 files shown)


vickyliu-go4it:
After adjusting the unit:
Screenshot 2021-03-16 193440

ronaldfw:
This looks really great! A small nitpick I have is that reading kB values when the ranges are in the GB is really hard. I would consider doing all this in MB, which seems to be a good trade-off between readability and precision.

vickyliu-go4it:
> This looks really great!
> 
> A small nitpick I have is that reading kB values when the ranges are in the GB is really hard. I would consider doing all this in MB, which seems to be a good trade-off between readability and precision. Similarly for the settings and putting the sampling period in ns there, which is too precise anyway (users can't put 1 ns there, for example, as it's impossible to sample at that frequency). Have you tested what minimum sampling period is possible? We should tell users what is a feasible range there.

Thanks for the suggestions! 

Totally agree that showing memory tracks in "MB" provides a good trade-off between readability and precision. Will change accordingly. Also I have changed to set sampling period in milliseconds in the CaptureOptionsDialog. I have checked whether the performance would be affected by high-frequency memory sampling with introspection. It seems that >= 1ms would be precise enough without affecting the performance.

src/OrbitQt/CaptureOptionsDialog.cpp
```diff
@@ -77,4 +79,36 @@ void CaptureOptionsDialog::ShowSourcePathsMappingEditor() {
...
+bool CaptureOptionsDialog::GetCollectMemoryInfo() const {
+  return ui_->collectMemoryInfoCheckBox->isChecked();
+}
+
+void CaptureOptionsDialog::SetMemorySamplingPeriodNs(uint64_t memory_sampling_period_ns) {
+  ui_->memorySamplingPeriodNsLineEdit->setText(QString::number(memory_sampling_period_ns));
+}
+
+uint64_t CaptureOptionsDialog::GetMemorySamplingPeriodNs() const {
+  CHECK(!ui_->memorySamplingPeriodNsLineEdit->text().isEmpty());
```
antonrohr, line 98: This does not work, it crashes when the user leaves the field empty

src/OrbitCaptureClient/CaptureClient.cpp
```diff
@@ -58,6 +58,7 @@ static CaptureOptions::InstrumentedFunction::FunctionType InstrumentedFunctionTy
         FunctionInfo_OrbitType_FunctionInfo_OrbitType_INT_MAX_SENTINEL_DO_NOT_USE_:
       UNREACHABLE();
   }
+  UNREACHABLE();
```
antonrohr, line 61: Just out of curiosity, why is this necessary? Did a build fail?

src/OrbitGl/TimeGraph.cpp
```diff
@@ -380,40 +380,50 @@ void TimeGraph::ProcessValueTrackingTimer(const TimerInfo& timer_info) {
...
+  const std::string kTotalLabel = "Memory Total";
   const std::string kUnusedLabel = "Memory Unused (kB)";
   const std::string kBuffersOrCachedLabel = "Memory Buffers / Cached (kB)";
   const std::string kUsedLabel = "Memory Used (kB)";
+  const std::string kWarningThresholdLabel = "Production Limit";
 
-  std::optional<std::pair<std::string, uint64_t>> warning_threshold_label_and_kb = std::nullopt;
+  std::optional<std::pair<std::string, uint64_t>> warning_threshold_pretty_name_and_raw_value =
+      std::nullopt;
+  const uint64_t kKilobytesToBytes = 1024;
```
antonrohr, line 391: nit: constexpr

src/OrbitQt/CaptureOptionsDialog.h
```diff
@@ -48,7 +48,7 @@ class CaptureOptionsDialog : public QDialog {
 
   void SetCollectMemoryInfo(bool collect_memory_info);
   [[nodiscard]] bool GetCollectMemoryInfo() const;
-  void SetMemorySamplingPeriodNs(uint64_t memory_sampling_period_ns);
+  void SetMemorySamplingPeriodMs(uint64_t memory_sampling_period_ns);
```
antonrohr, line 51: This does not make much sense. The method has the suffix `Ms` and the parameter is still `ns`. The implementation also still wants a nanosecond value.

antonrohr (commented):
I tested it and it works quite nicely, but I have some remarks in the code

ronaldfw:
> > This looks really great!
> > A small nitpick I have is that reading kB values when the ranges are in the GB is really hard. I would consider doing all this in MB, which seems to be a good trade-off between readability and precision. Similarly for the settings and putting the sampling period in ns there, which is too precise anyway (users can't put 1 ns there, for example, as it's impossible to sample at that frequency). Have you tested what minimum sampling period is possible? We should tell users what is a feasible range there.
> 
> Thanks for the suggestions!
> 
> Totally agree that showing memory tracks in "MB" provides a good trade-off between readability and precision. Will change accordingly. Also I have changed to set sampling period in milliseconds in the CaptureOptionsDialog. I have checked whether the performance would be affected by high-frequency memory sampling with introspection. It seems that >= 1ms would be precise enough without affecting the performance.

Sounds perfect, thanks for testing this!

src/OrbitQt/CaptureOptionsDialog.cpp
```diff
@@ -77,4 +79,36 @@ void CaptureOptionsDialog::ShowSourcePathsMappingEditor() {
...
+  bool valid = false;
+  uint64_t result = ui_->memorySamplingPeriodNsLineEdit->text().toULongLong(&valid);
+  CHECK(valid);
```
karupayun, line 102: Are we going to crash if the user complete this with a non-numeric value?

src/OrbitQt/CaptureOptionsDialog.cpp
```diff
@@ -77,4 +79,36 @@ void CaptureOptionsDialog::ShowSourcePathsMappingEditor() {
...
+
+void CaptureOptionsDialog::SetMemoryWarningThresholdKb(uint64_t memory_warning_threshold_kb) {
+  ui_->memoryWarningThresholdKbLineEdit->setText(QString::number(memory_warning_threshold_kb));
+}
+
+uint64_t CaptureOptionsDialog::GetMemoryWarningThresholdKb() const {
+  CHECK(!ui_->memoryWarningThresholdKbLineEdit->text().isEmpty());
+  bool valid = false;
+  uint64_t result = ui_->memoryWarningThresholdKbLineEdit->text().toULongLong(&valid);
+  CHECK(valid);
```
karupayun, line 114: same here

src/OrbitQt/CaptureOptionsDialog.ui
```diff
@@ -61,6 +61,79 @@
...
+          <item>
+           <widget class="QLineEdit" name="memoryWarningThresholdKbLineEdit">
+            <property name="enabled">
+             <bool>false</bool>
+            </property>
+            <property name="inputMethodHints">
+             <set>Qt::ImhDigitsOnly</set>
+            </property>
+            <property name="text">
+             <string>8388608</string>
```
karupayun, line 128: Can we write a comment there? "#8GB" or "8 x 1024 x 1024"

src/OrbitCaptureClient/CaptureClient.cpp
```diff
@@ -66,7 +66,8 @@ Future<ErrorMessageOr<CaptureListener::CaptureOutcome>> CaptureClient::Capture(
     absl::flat_hash_map<uint64_t, FunctionInfo> selected_functions,
     TracepointInfoSet selected_tracepoints, absl::flat_hash_set<uint64_t> frame_track_function_ids,
     bool collect_thread_state, bool enable_introspection,
-    uint64_t max_local_marker_depth_per_command_buffer) {
+    uint64_t max_local_marker_depth_per_command_buffer, bool collect_memory_info,
```
karupayun, line 70: This function has a lot of arguments, it should be some struct, could you add a TODO about that?

src/OrbitGl/App.h
```diff
@@ -361,6 +361,20 @@ class OrbitApp final : public DataViewFactory, public CaptureListener {
   void SetCollectThreadStates(bool collect_thread_states);
   void SetMaxLocalMarkerDepthPerCommandBuffer(uint64_t max_local_marker_depth_per_command_buffer);
 
+  void SetCollectMemoryInfo(bool collect_memory_info) {
+    data_manager_->set_collect_memory_info(collect_memory_info);
+  }
+  [[nodiscard]] bool GetCollectMemoryInfo() const { return data_manager_->collect_memory_info(); }
+  void SetMemorySamplingPeriodNs(uint64_t memory_sampling_period_ns) {
```
karupayun, line 368: Maybe stupid question: We don't have GetMemorySamplingPeriod because we are not using it?

src/OrbitQt/orbitmainwindow.cpp
```diff
@@ -854,16 +854,38 @@ void OrbitMainWindow::on_actionSave_Preset_As_triggered() {
...
   app_->SetCollectThreadStates(settings.value(kCollectThreadStatesSettingKey, false).toBool());
 
+  app_->SetCollectMemoryInfo(settings.value(kCollectMemoryInfoSettingKey, false).toBool());
+  uint64_t memory_sampling_period_ns = kMemorySamplingPeriodNsDefaultValue;
+  uint64_t memory_warning_threshold_kb = kMemoryWarningThresholdKbDefaultValue;
+  if (app_->GetCollectMemoryInfo()) {
+    memory_sampling_period_ns = settings
+                                    .value(kMemorySamplingPeriodNsSettingKey,
+                                           QVariant::fromValue(kMemorySamplingPeriodNsDefaultValue))
+                                    .toULongLong();
```
karupayun, line 879: Are we checking that nothing could crash if the user insert an invalid number here?

src/OrbitGl/GraphTrack.cpp
```diff
@@ -89,23 +89,12 @@ void GraphTrack::Draw(GlCanvas* canvas, PickingMode picking_mode, float z_offset
...
-  float point_x = time_graph_->GetWorldFromTick(label_time);
-  double normalized_value = (value - min_) * inv_value_range_;
-  float point_y = pos_[1] - size_[1] * (1.f - static_cast<float>(normalized_value));
   const Color kBlack(0, 0, 0, 255);
   const Color kWhite(255, 255, 255, 255);
   float text_z = GlCanvas::kZValueEvent + z_offset;
-  DrawLabel(canvas, Vec2(point_x, point_y), std::to_string(value), kBlack, kWhite, text_z);
 
   // Add warning threshold text box and line.
+  Batcher* batcher = canvas->GetBatcher();
```
karupayun, line 97: This was here before, but please rename to ui_batcher to be consistent with other tracks.

src/OrbitGl/TimeGraph.cpp
```diff
@@ -380,40 +380,50 @@ void TimeGraph::ProcessValueTrackingTimer(const TimerInfo& timer_info) {
...
   }
 
   int64_t total_kb = orbit_api::Decode<int64_t>(timer_info.registers(0));
-  std::optional<std::pair<std::string, int64_t>> total_label_and_kb = std::nullopt;
-  if (total_kb != kMissingInfo) total_label_and_kb = std::make_pair(kTotalLabel, total_kb);
+  std::optional<std::pair<std::string, int64_t>> total_pretty_name_and_raw_value = std::nullopt;
+  if (total_kb != kMissingInfo) {
+    std::string pretty_size = GetPrettySize(total_kb * kKilobytesToBytes);
+    std::string pretty_name = absl::StrFormat("%s: %s", kTotalLabel, pretty_size);
+    total_pretty_name_and_raw_value = std::make_pair(pretty_name, total_kb);
```
karupayun, line 404: This variable name doesn't say me anything. I won't comment a suggestion because I know that you are modifying code here.

src/OrbitGl/GraphTrack.cpp
```diff
@@ -88,20 +89,60 @@ void GraphTrack::Draw(GlCanvas* canvas, PickingMode picking_mode, float z_offset
...
+                           pos_[1] - layout_->GetTextBoxHeight() / 2.f);
+    canvas->GetTextRenderer().AddText(text.c_str(), text_box_position[0],
+                                      text_box_position[1] + layout_->GetTextOffset(), text_z,
+                                      kWhite, font_size, text_box_size[0]);
+  }
+
   // Draw label
   uint64_t current_mouse_time_ns = time_graph_->GetCurrentMouseTimeNs();
   auto previous_point = GetPreviousValueAndTime(current_mouse_time_ns);
-
```
karupayun, line 94: This line was erased on purpose?

src/OrbitGl/GraphTrack.h
```diff
@@ -31,10 +31,21 @@ class GraphTrack : public Track {
                         PickingMode picking_mode, float z_offset = 0) override;
   [[nodiscard]] float GetHeight() const override;
   void AddValue(double value, uint64_t time);
-  [[nodiscard]] std::optional<std::pair<uint64_t, double> > GetPreviousValueAndTime(
+  [[nodiscard]] std::optional<std::pair<uint64_t, double>> GetPreviousValueAndTime(
       uint64_t time) const;
   [[nodiscard]] bool IsEmpty() const override { return values_.empty(); }
 
+  void SetWarningThreshold(const std::optional<std::pair<std::string, double>>& warning_threshold) {
```
karupayun, line 38: I don't like to have this type in the set function.

What do you think about:
- Have a ClearWarningThreshold function to set this to false.
- Have also a SetWarningThreshold which takes an string and also a double and assumes that it is true?

And also don't ask for them in the constructors. 

Also maybe for a future PR, I would like to have it only in a MemoryTrack, and not in all the GraphTracks. So, MemoryTrack inherit from GraphTrack and also having this SetWarningThreshold and ClearWarningThreshold methods.

karupayun (commented):
Nice PR Vicky! I'm really happy about having memory information displayed in Orbit!! :) :)

I made a few comments, most of them are nitpicks which you can disregard.

And as I told you in the last meeting, I'm not sure of having this memory message in the top right. IMHO it doesn't seem very nice, but maybe I'm losing something. Why we are having both the Production Limit threshold and the information about the total memory??

And lastly, are we sure that we need the three graphs? Isn't enough with having two of them (for example Used + Cached + .... and Unused)?

src/OrbitGl/App.cpp
```diff
@@ -279,6 +279,23 @@ void OrbitApp::OnTimer(const TimerInfo& timer_info) {
...
 
+void OrbitApp::OnSystemMemoryUsage(
+    const orbit_grpc_protos::SystemMemoryUsage& system_memory_usage) {
+  TimerInfo timer_info;
+  timer_info.set_type(TimerInfo::kSystemMemoryUsage);
+  timer_info.set_start(system_memory_usage.timestamp_ns());
+  timer_info.set_end(system_memory_usage.timestamp_ns());
+
+  timer_info.mutable_registers()->Reserve(5);
+  timer_info.add_registers(static_cast<uint64_t>(system_memory_usage.total_kb()));
```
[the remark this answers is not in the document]
vickyliu-go4it: Thanks for pointing out this!

We use `kMissingInfo = -1` in `SystemMemoryUsage` to indicate that case that the require information is missing. This is the reason why we have `int64_t` rather than `uint64_t` in the memory size raw values.

Now I changed to use `orbit_api::Encode` to encode the value into `unit_64`. This is the same way we use in value tracking: we use `orbit_api::Encode` to encode `int64_t`, `float` and other types into `uint64_t.`

src/OrbitGl/TimeGraph.cpp
```diff
@@ -375,6 +379,34 @@ void TimeGraph::ProcessValueTrackingTimer(const TimerInfo& timer_info) {
...
 }
 
+void TimeGraph::ProcessMemoryTrackingTimer(const TimerInfo& timer_info) {
+  const std::string kTotalLabel = "Memory Total (kB)";
+  const std::string kUnusedLabel = "Memory Unused (kB)";
+  const std::string kBuffersOrCachedLabel = "Memory Buffers / Cached (kB)";
+  const std::string kUsedLabel = "Memory Used (kB)";
+
+  GraphTrack* track;
+  int64_t total_kb = orbit_api::Decode<int64_t>(timer_info.registers(0));
```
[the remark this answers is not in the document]
vickyliu-go4it: Really good suggestion! I have added the `enum SystemMemoryUsageEncodingIndex` in `OrbitApp` so they can be used when decoding the memory information.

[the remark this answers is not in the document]
vickyliu-go4it: Thanks for pointing out this issue!

I have changed to use `orbit_api::Encode` and `orbit_api::Decode` such that we can recover the value successfully.

Using `static_cast` seems can not preserve the value. For example, the following test case:
```
    int64_t raw = -1;
    uint64_t encoded;
    int64_t decoded;
    
    encoded = static_cast<uint64_t>(raw);
    decoded = static_cast<int64_t>(decoded);
    
    cont << raw << "==>" << encoded << "==>" << decoded;
```

will output `-1==>18446744073709551615==>0`.

src/OrbitGl/TimeGraph.cpp
```diff
@@ -375,6 +379,34 @@ void TimeGraph::ProcessValueTrackingTimer(const TimerInfo& timer_info) {
...
+  int64_t unused_kb = orbit_api::Decode<int64_t>(timer_info.registers(1));
+  if (unused_kb != kMissingInfo) {
```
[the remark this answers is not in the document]
vickyliu-go4it: Changed to use `orbit_api::Encode` and `orbit_api::Decode` method which can help recover the negative value succesfully. Thanks!

src/OrbitGl/TimeGraph.cpp
```diff
@@ -375,6 +379,34 @@ void TimeGraph::ProcessValueTrackingTimer(const TimerInfo& timer_info) {
...
+    track = track_manager_->GetOrCreateGraphTrack(kUnusedLabel);
+    track->AddValue(unused_kb, timer_info.start());
```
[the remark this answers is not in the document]
vickyliu-go4it: Thanks! I have added `static_cast<double>` accordingly!

src/OrbitGl/TrackManager.h
```diff
@@ -56,7 +56,10 @@ class TrackManager {
   SchedulerTrack* GetOrCreateSchedulerTrack();
   ThreadTrack* GetOrCreateThreadTrack(int32_t tid);
   GpuTrack* GetOrCreateGpuTrack(uint64_t timeline_hash);
-  GraphTrack* GetOrCreateGraphTrack(const std::string& name);
+  GraphTrack* GetOrCreateGraphTrack(
+      const std::string& name,
+      const std::optional<std::pair<std::string, double>>& warning_threshold = std::nullopt,
+      const std::optional<std::pair<std::string, double>>& value_upper_bound = std::nullopt);
```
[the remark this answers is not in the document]
vickyliu-go4it: That's really nice! I changed to call `track->SetWarningThresholdWhenEmpty` and `track->SetValueUpperBoundWhenEmpty`.

src/OrbitQt/CaptureOptionsDialog.cpp
vickyliu-go4it: Thanks for pointing out this issue.
I have added the `ResetMemorySamplingPeriodMsLineEditWhenEmpty` method such that the value will be set to the default value (i.e., 100ms) if empty.

src/OrbitCaptureClient/CaptureClient.cpp
vickyliu-go4it: Some checks failed with the error of "control reaches the end of a non-void function".

Screenshot 2021-03-18 132908

src/OrbitGl/TimeGraph.cpp
vickyliu-go4it: Thanks! Changed accordingly!

src/OrbitQt/CaptureOptionsDialog.h
vickyliu-go4it: Thanks for pointing out this issue! 

We are using nanoseconds in the proto but we would like to only allow the user to set the sampling period in milliseconds in the ui. So I change to rename some components and methods by replacing "Ns" with "Ms".

I change the component in the ui to be `memorySamplingPeriodMsLineEdit`. And here when setting the sampling period programmatically, we pass the value in nanoseconds and this `SetMemorySamplingPeriodMs` method will convert it to milliseconds and then set up the value in the ui.

When getting the sampling period value in the ui with `GetMemorySamplingPeriodNs`, we get the raw value in milliseconds and then convert it to nanoseconds.

src/OrbitQt/CaptureOptionsDialog.cpp
vickyliu-go4it: The user is not able to input a non-numeric value here. We set `<set>Qt::ImhDigitsOnly</set>` in the ui, such that only digits are allowed.

src/OrbitQt/CaptureOptionsDialog.cpp
vickyliu-go4it: Thanks for pointing out this. For those line edit components, we only allow the digit inputs by adding `<set>Qt::ImhDigitsOnly</set>`.

src/OrbitQt/orbitmainwindow.cpp
vickyliu-go4it: Yep! We have the `UInt64Validator` to check the input number is valid or not.

src/OrbitQt/CaptureOptionsDialog.ui
vickyliu-go4it: Good idea. I added a tooltip "The default value is 8388608 KB, i.e., 8GB." for this.

src/OrbitGl/GraphTrack.cpp
vickyliu-go4it: Changed! Thanks!

src/OrbitGl/GraphTrack.cpp
vickyliu-go4it: Ah. It is removed by mistake. I added it back for the readability.

src/OrbitCaptureClient/CaptureClient.cpp
vickyliu-go4it: Done. Thanks for the suggestion!

src/OrbitGl/App.h
vickyliu-go4it: Yep. This is not stupid at or :-)! For `memory_sampling_period_ns`, we will set it to `grpc_protos::CaptureOptions`. And when we need the value in the Service side, we can use `capture_option.memory_sampling_period_ns()`. 
`warning_threshold` is not a part of `grpc_protos::CaptureOptions`, so we have a Get method for it.

src/OrbitGl/GraphTrack.h
vickyliu-go4it: Make sense. As we discussed offline, we may have a specific implementation for the memory tracks. This would be helpful especially when we do a type based filtering to hide and show tracks. Because memory timers, value tracking timers and introspection timers can all be GraphTracks, so with `Track::Type` we cannot differentiate the purpose of different tracks. So I agree that we can let the MemoryTrack to be inherit from the GraphTrack.

vickyliu-go4it:
> Nice PR Vicky! I'm really happy about having memory information displayed in Orbit!! :) :)
> 
> I made a few comments, most of them are nitpicks which you can disregard.
> 
> And as I told you in the last meeting, I'm not sure of having this memory message in the top right. IMHO it doesn't seem very nice, but maybe I'm losing something. Why we are having both the Production Limit threshold and the information about the total memory??
Thanks for the suggestions.

The memory size limit is 16GB for developer and 8GB for production. So for the developers, it is helpful to make it more obvious when the memory usage exceeds the production limits.

> 
> And lastly, are we sure that we need the three graphs? Isn't enough with having two of them (for example Used + Cached + .... and Unused)?

Just removed the unused memory track and only kept two of them.

vickyliu-go4it:
Thanks for all the valuable review comments. I have made changes according to the review comments and the suggestions received in today's discussion.

This is how the memory tracks look like currently 
Screenshot 2021-03-18 152926

Screenshot 2021-03-18 152912

I will keep working on fix the check failure caused by the commit 
[Add label unit and format value for the memory track … 946e228](https://github.com/google/orbit/pull/2076/commits/946e228863ffcf1e2d1ee5754fe59f07f62ef3fb).

src/OrbitGl/GraphTrack.h
karupayun: Sure! And what do you think about having this SetWarningThreshold (string , double) instead of the optional <pair> (what I wrote in the top part)?

src/OrbitQt/CaptureOptionsDialog.ui
karupayun: I was thinking more in adding a comment to have in the .ui file, but your change probably is better :)

src/OrbitGl/GraphTrack.h
vickyliu-go4it: Sounds good! So we can keep the type of `warning_threshold_` but set it with string and double instead of optional. I will change accordingly : )

src/OrbitGl/TimeGraph.cpp
karupayun: mmm, if you still want to use the optional in the Set method, I would rename the variables:
pretty_total_size
pretty_total_label
pretty_total_label_and_value

src/OrbitGl/GraphTrack.cpp
```diff
@@ -170,3 +227,57 @@ float GraphTrack::GetHeight() const {
...
+}
+
+// `absl::StrFormat` only allows constexpr as the first parameter. Hence, we are not able to
+// configure the value decimal digits by adding a variable format string as the first parameter in
+// `absl::StrFormat`. As the default format has 6 decimal digits, we only enumerate the cases where
+// 0 <= value_decimal_digits_ <=5.
+std::string GraphTrack::FormatValue(double value) {
+  if (!value_decimal_digits_.has_value()) return std::to_string(value);
+
+  switch (value_decimal_digits_.value()) {
```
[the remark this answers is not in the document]
vickyliu-go4it: That really helps! Changed accordingly! Thanks so much for the suggestions!

src/OrbitGl/TimeGraph.cpp
vickyliu-go4it: Removed optional arguments in the Set methods and also renamed the variables. Thanks!

karupayun (approved):
By my side, LGTM!!! 

Again, I'm really happy to have memory information now!! :)

src/OrbitGl/GraphTrack.cpp
```diff
@@ -170,3 +227,43 @@ float GraphTrack::GetHeight() const {
...
+  max_ = std::max(max_, value);
+  min_ = std::min(min_, value);
+}
+
+std::string GraphTrack::FormatValue(double value) {
+  if (!value_decimal_digits_.has_value()) return std::to_string(value);
+
+  std::stringstream stream;
+  stream << std::fixed << std::setprecision(value_decimal_digits_.value()) << value;
+  return stream.str();
```
antonrohr, line 268: This can be done with: (Sorry for not realizing this yesterday. I needed to google this myself)
```
absl::StrFormat("%.*f", value_decimal_digits_.value(), value)
```

Which means you can remove the method `FormatValue`. Which is a good thing, because the naming `FormatValue` is a not a good name. It does not tell me anything about what the method does, only some vary vague formatting.

src/OrbitGl/TimeGraph.cpp
```diff
@@ -375,6 +379,70 @@ void TimeGraph::ProcessValueTrackingTimer(const TimerInfo& timer_info) {
...
+  uint64_t warning_threshold_kb = app_->GetMemoryWarningThresholdKb();
+  std::string warning_threshold_pretty_size =
+      GetPrettySize(warning_threshold_kb * kKilobytesToBytes);
+  std::string warning_threshold_pretty_label =
+      absl::StrFormat("%s: %s", kWarningThresholdLabel, warning_threshold_pretty_size);
+  double warning_threshold_raw_value =
+      static_cast<double>(warning_threshold_kb) / kMegabytesToKilobytes;
+
+  int64_t total_kb = orbit_api::Decode<int64_t>(timer_info.registers(OrbitApp::kTotalKb));
+  std::string total_pretty_label = "";
```
antonrohr, line 406: nit
```suggestion
  std::string total_pretty_label;
```
You can rely on the default constructor here

src/OrbitQt/CaptureOptionsDialog.h
antonrohr: Sorry, I dont think I was clear enough with what I meant. From an api design standpoint you want to design and name your methods in a way that it is clear how to use them. 

If I as a user read the name `SetMemorySamplingPeriodMs`, I assume I can give this a value in milliseconds and it sets the memory sampling period. But this is not the case. I actually have to give a value in nanoseconds. This is confusing for someone who wants to use that method.

src/OrbitGl/GraphTrack.cpp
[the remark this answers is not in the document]
antonrohr: see my comment, this can be done with `absl::StrFormat("%.*f", value_decimal_digits_.value(), value)`.

antonrohr:
I realized one more thing. The sampling period can be set to 0 ms. I think we should not allow that. 

Screenshot 2021-03-19 at 12 13 28

src/OrbitGl/App.h
```diff
@@ -151,6 +149,10 @@ class OrbitApp final : public DataViewFactory, public CaptureListener {
                               orbit_grpc_protos::TracepointInfo tracepoint_info) override;
   void OnTracepointEvent(orbit_client_protos::TracepointEventInfo tracepoint_event_info) override;
 
+  enum SystemMemoryUsageEncodingIndex { kTotalKb, kFreeKb, kAvailableKb, kBuffersKb, KCachedKb };
```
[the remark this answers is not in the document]
vickyliu-go4it: Thanks for pointing out this! Changed accordingly.

src/OrbitCaptureClient/CaptureClient.cpp
```diff
@@ -58,15 +58,18 @@ static CaptureOptions::InstrumentedFunction::FunctionType InstrumentedFunctionTy
         FunctionInfo_OrbitType_FunctionInfo_OrbitType_INT_MAX_SENTINEL_DO_NOT_USE_:
       UNREACHABLE();
   }
+  UNREACHABLE();
 }
 
+// TODO: This method contains a lot of arguments. Consider making it more structured.
```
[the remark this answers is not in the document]
vickyliu-go4it: Done. Thanks!

src/OrbitGl/App.cpp
```diff
@@ -287,11 +287,11 @@ void OrbitApp::OnSystemMemoryUsage(
...
-  timer_info.add_registers(static_cast<uint64_t>(system_memory_usage.total_kb()));
-  timer_info.add_registers(static_cast<uint64_t>(system_memory_usage.free_kb()));
-  timer_info.add_registers(static_cast<uint64_t>(system_memory_usage.available_kb()));
-  timer_info.add_registers(static_cast<uint64_t>(system_memory_usage.buffers_kb()));
-  timer_info.add_registers(static_cast<uint64_t>(system_memory_usage.cached_kb()));
+  timer_info.add_registers(orbit_api::Encode<uint64_t>(system_memory_usage.total_kb()));
+  timer_info.add_registers(orbit_api::Encode<uint64_t>(system_memory_usage.free_kb()));
+  timer_info.add_registers(orbit_api::Encode<uint64_t>(system_memory_usage.available_kb()));
+  timer_info.add_registers(orbit_api::Encode<uint64_t>(system_memory_usage.buffers_kb()));
+  timer_info.add_registers(orbit_api::Encode<uint64_t>(system_memory_usage.cached_kb()));
```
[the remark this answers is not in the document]
vickyliu-go4it: Thanks for the suggestion! changed accordingly.

src/OrbitGl/GraphTrack.cpp
vickyliu-go4it: Thanks for suggesting a better solution! Changed accordingly!

src/OrbitGl/TimeGraph.cpp
vickyliu-go4it: Done!

src/OrbitQt/CaptureOptionsDialog.h
vickyliu-go4it: Thanks for the explanation! I changed to use only milliseconds in the CaptureOptionsDialog .h/.cpp/.ui files. 

The conversions to nanoseconds are moved to the OrbitMainWindow.

vickyliu-go4it:
> I realized one more thing. The sampling period can be set to 0 ms. I think we should not allow that.
> 
> Screenshot 2021-03-19 at 12 13 28

Thanks for finding this! I have updated the validator to support checking the value lower bound (if it is provided).

src/OrbitQt/CaptureOptionsDialog.h
antonrohr: Thanks!!

src/OrbitQt/CaptureOptionsDialog.h
```diff
@@ -20,17 +20,22 @@ namespace orbit_qt {
...
-    if (valid) {
+    uint64_t input_value = input.toULongLong(&valid);
+    if (valid && input_value >= bottom_) {
       return QValidator::State::Acceptable;
     }
     return QValidator::State::Invalid;
   }
+
+ private:
+  uint64_t bottom_ = 0;
```
antonrohr, line 38: nit:
```suggestion
  uint64_t minimum_ = 0;
```

antonrohr (approved):
LGTM! Thank you for incorporating all the changes I requested!

PR #2077, opened 2021-03-16
Pass function name for instrumetned functions

dimitry-:
The name is used in new capture file format when saving on the service
side. It is not directly used in capture.

Test: run integration test/unit tests

### DIFF
--- a/src/GrpcProtos/capture.proto
+++ b/src/GrpcProtos/capture.proto
@@ -12,6 +12,20 @@ import "tracepoint.proto";
 // This is needed by LockFreeBufferCaptureEventProducer.
 option cc_enable_arenas = true;
 
+message InstrumentedFunction {
+  string file_path = 1;
+  uint64 file_offset = 2;
+  uint64 function_id = 3;
+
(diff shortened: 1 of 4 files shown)


florian-kuebler (approved):
LGTM

src/GrpcProtos/capture.proto
```diff
@@ -12,6 +12,20 @@ import "tracepoint.proto";
...
+  uint64 file_offset = 2;
+  uint64 function_id = 3;
+
+  enum FunctionType {
+    kRegular = 0;
+    kTimerStart = 1;
+    kTimerStop = 2;
+  }
+  FunctionType function_type = 4;
+  bytes function_name = 5;
```
dpallotti, line 26: I think we should keep using `string` until we are sure that this is not a problem. You are not gaining a lot here anyway.

dpallotti (approved):

src/GrpcProtos/capture.proto
dimitry-: done.

PR #2078, opened 2021-03-16
Fix SourceCodeView E2E test after non-modal change

beckerhe:
The code viewer dialog is not a modal dialog anymore. That means it will
appear as a top level window in the accessibility tree, instead of as a
child element of the main window.

This commit adjusts the E2E test accordingly.


karupayun (approved):
LGTM!

PR #2080, opened 2021-03-16
Introduce CaptureStarted message

dimitry-:
The message passed all necessary information about initial state of process,
and (in case the capture is consumed by process different from the one that initiated it)
CaptureOptions used to start this capture.

Test: run all unit-tests
Test: run Orbit instrument functions, take capture.
Test: run LinuxTracingIntegrationTest
Test: save capture and load saved capture.


dimitry-:
(actually LinuxTracingIntegrationTest.ThreadNames integration test if failing, working on it)

src/Service/LinuxTracingHandler.cpp
```diff
@@ -68,6 +66,20 @@ void LinuxTracingHandler::Stop() {
...
+  absl::MutexLock lock(&capture_started_mutex_);
+  capture_started_mutex_.Await(absl::Condition(&capture_started_event_sent_));
+}
+
+void LinuxTracingHandler::OnCaptureStarted(orbit_grpc_protos::CaptureStarted capture_started) {
+  ProducerCaptureEvent event;
+  *event.mutable_capture_started() = std::move(capture_started);
+  producer_event_processor_->ProcessEvent(kLinuxTracingProducerId, std::move(event));
+
+  absl::MutexLock lock(&capture_started_mutex_);
```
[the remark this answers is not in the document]
dimitry-: Await function unlocks the mutex.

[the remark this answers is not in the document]
florian-kuebler: That makes sense. Thanks!

src/GrpcProtos/capture.proto
```diff
@@ -286,6 +286,16 @@ message SystemMemoryUsage {
   int64 cached_kb = 6;
 }
 
+message CaptureStarted {
+  int32 process_id = 1;
+  string executable_path = 2;
+  string executable_build_id = 3;
+  uint64 capture_start_timestamp_ns = 4;
+  repeated ModuleInfo modules = 5;
+  repeated ThreadName thread_names = 6;
```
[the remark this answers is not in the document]
dimitry-: part of initial state of the system in this case - these are list of thread at the moment when we start capturing. I will add a comment to this and modules.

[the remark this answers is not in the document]
florian-kuebler: Cool, thanks!

dimitry-:
fixed the test. This is ready for review, ptal.

florian-kuebler (approved):
LGTM

src/Service/CaptureServiceImpl.cpp
```diff
@@ -216,6 +216,11 @@ grpc::Status CaptureServiceImpl::Capture(
   LOG("Read CaptureRequest from Capture's gRPC stream: starting capture");
 
   tracing_handler.Start(request.capture_options());
+
+  // Tracing handler is responsible for generating CaptureStarted event which should always
+  // go first, we wait until it is done before continuing.
+  tracing_handler.WaitForCaptureStartedEvent();
```
dpallotti, line 222: I think there's a discussion to be had here about `CaptureStarted` being generated by `LinuxTracing`. This effectively makes `LinuxTracing` the "master" producer. Secondarily, it also adds some code to `TracerThread::Startup` that doesn't necessarily belong there (although there's already so much stuff in `LinuxTracing` anyway).
This doesn't seem very clean. What's the reason for this? Can't `CaptureStarted` be generate in a dedicated place? The only thing that really seems to belong in `LinuxTracing` is the initial thread names, but that can be done differently, e.g., the same way it was done before or with a dedicated `InitialThreadNames` message.
Have you considered alternatives?

dpallotti (commented):
I'm not done but I'm sending it now for the big conceptual question about `LinuxTracing` generating `CaptureStarted`.

src/OrbitClientModel/CaptureSerializerTest.cpp
```diff
@@ -38,30 +35,30 @@ using orbit_client_protos::FunctionInfo;
 using orbit_client_protos::FunctionStats;
 using orbit_client_protos::LinuxAddressInfo;
 using orbit_client_protos::TracepointEventInfo;
+using orbit_grpc_protos::CaptureStarted;
 using orbit_grpc_protos::TracepointInfo;
 using ::testing::ElementsAreArray;
 
+ABSL_FLAG(uint16_t, sampling_rate, 1000, "Frequency of callstack sampling in samples per second");
+ABSL_FLAG(bool, frame_pointer_unwinding, false, "Use frame pointers for unwinding");
```
dpallotti, line 43: What changed here such that these flags are needed, but only these?

src/OrbitClientModel/CaptureDeserializerLoadFuzzer.cpp
```diff
@@ -18,13 +19,14 @@
 #include "OrbitClientModel/CaptureDeserializer.h"
 #include "OrbitClientModel/CaptureSerializer.h"
 
+ABSL_FLAG(uint16_t, sampling_rate, 1000, "Frequency of callstack sampling in samples per second");
+ABSL_FLAG(bool, frame_pointer_unwinding, false, "Use frame pointers for unwinding");
```
dpallotti, line 23: Same here: what are these flags for here and why only these?

src/OrbitClientModel/CaptureDeserializer.cpp
```diff
@@ -178,9 +182,36 @@ ErrorMessageOr<CaptureListener::CaptureOutcome> LoadCaptureInfo(
...
   }
 
-  capture_listener->OnCaptureStarted(std::move(process), std::move(instrumented_functions),
-                                     std::move(selected_tracepoints),
-                                     std::move(frame_track_function_ids));
+  CaptureStarted capture_started;
+  capture_started.set_process_id(capture_info.process().pid());
+  capture_started.set_executable_path(process.full_path());
+  capture_started.set_process_id(process.pid());
+  capture_started.set_capture_start_timestamp_ns(0);  // Do we have this information anywhere?
```
dpallotti, line 189: > Do we have this information anywhere?

I don't remember that we do. I believe just the timestamp of the first event, or the min timestamp.

But I think that such comment shouldn't be left around, rather it should be stated why 0 is used.

src/OrbitCaptureClient/include/OrbitCaptureClient/CaptureListener.h
```diff
@@ -20,11 +21,8 @@ class CaptureListener {
   virtual ~CaptureListener() = default;
 
   // Called after capture started but before the first event arrived.
```
dpallotti, line 23: This comment is now no longer correct, is it?

src/LinuxTracing/TracerThread.cpp
```diff
@@ -590,12 +617,47 @@ void TracerThread::Startup() {
...
-  RetrieveInitialThreadNamesSystemWideAndNotifyListener(effective_capture_start_timestamp_ns_);
+  std::vector<ThreadName> thread_names =
+      RetrieveInitialThreadNamesSystemWide(effective_capture_start_timestamp_ns_);
+
+  *capture_started.mutable_thread_names() = {thread_names.begin(), thread_names.end()};
+
+  capture_started.set_capture_start_timestamp_ns(effective_capture_start_timestamp_ns_);
+  capture_started.mutable_capture_options()->CopyFrom(capture_options_);
+
+  listener_->OnCaptureStarted(std::move(capture_started));
```
dpallotti, line 660: If this code stays in `TracerThread`, I think that extracting it to `CreateCaptureStartedAndSendToListener` would be useful.

src/LinuxTracing/TracerThread.cpp
```diff
@@ -590,12 +617,47 @@ void TracerThread::Startup() {
...
+  auto modules_or_error = orbit_elf_utils::ReadModules(target_pid_);
+  if (modules_or_error.has_value()) {
+    const std::vector<ModuleInfo>& modules = modules_or_error.value();
+    *capture_started.mutable_modules() = {modules.begin(), modules.end()};
+  } else {
+    ERROR("Unable to load modules for %d: %s", target_pid_, modules_or_error.error().message());
+  }
+
+  // Get the initial thread names to construct CaptureStarted event and notify the listener_.
   // All ThreadName events generate by this call will have effective_capture_start_timestamp_ns_ as
```
dpallotti, line 648:
```suggestion
  // All ThreadName events generated by this call will have effective_capture_start_timestamp_ns_ as
```
My bad I believe.

src/LinuxTracing/LinuxTracingIntegrationTest.cpp
```diff
@@ -377,6 +386,11 @@ void VerifyOrderOfAllEvents(const std::vector<orbit_grpc_protos::ProducerCapture
   uint64_t previous_event_timestamp_ns = 0;
   for (const auto& event : events) {
     switch (event.event_case()) {
+      case orbit_grpc_protos::ProducerCaptureEvent::kCaptureStarted:
```
dpallotti, line 389: We could probably also test that it's sent at all.

src/LinuxTracing/LinuxTracingIntegrationTest.cpp
```diff
@@ -946,24 +960,34 @@ TEST(LinuxTracingIntegrationTest, ThreadNames) {
...
       continue;
     }
 
     const orbit_grpc_protos::ThreadName& thread_name = event.thread_name();
-    if (thread_name.tid() != fixture.GetPuppetPid()) {
+    if (thread_name.pid() != fixture.GetPuppetPid()) {
       continue;
     }
 
-    EXPECT_EQ(thread_name.pid(), fixture.GetPuppetPid());
```
dpallotti, line 960: Why is this being changed? The point here was to filter based on the `tid` and then verify that the `pid` is equal to the `tid`. This is because the puppet is single threaded. You are basically removing the test on `tid == pid` like this.

src/LinuxTracing/LinuxTracingIntegrationTest.cpp
```diff
@@ -946,24 +960,34 @@ TEST(LinuxTracingIntegrationTest, ThreadNames) {
...
-  std::vector<std::string> collected_event_names;
+  std::vector<std::string> changed_thread_names;
+  std::vector<std::string> initial_thread_names;
   for (const auto& event : events) {
+    if (event.event_case() == orbit_grpc_protos::ProducerCaptureEvent::kCaptureStarted) {
+      const orbit_grpc_protos::CaptureStarted& capture_started = event.capture_started();
+      for (const auto& thread_name : capture_started.thread_names()) {
+        if (thread_name.pid() != fixture.GetPuppetPid()) {
+          continue;
+        }
```
dpallotti, line 971: As for the comment below, this doesn't test the `tid`, which we can test.

src/Service/CaptureServiceImpl.cpp
dimitry-: I do not think initial thread names should be sent separately, we should have a snapshot of a process state at the start of the capture. It makes sense to include this to CaptureStartedEvent.

I was considering moving compilation of CaptureSartedEvent to ProducerEventProcessor but this is much bigger change. I was thinking that the initial memory snapshot should also be included to CaptureStarted event. And decided to postpone this to a separate change as well, this is because we might want to consider having initial messages for all other producers, this in will have affect on delaying all other messages sent to the client until all Start messages gathered and sent. And I am not sure we should do that since this could have a negative impact on memory consumption in the service and might create latency problems.

Do you have anything in particular you want to be addressed?

src/OrbitCaptureClient/CaptureEventProcessor.cpp
```diff
@@ -88,6 +91,11 @@ void CaptureEventProcessor::ProcessEvent(const ClientCaptureEvent& event) {
   }
 }
 
+void CaptureEventProcessor::ProcessCaptureStarted(
+    const orbit_grpc_protos::CaptureStarted& capture_started) {
+  capture_listener_->OnCaptureStarted(capture_started, frame_track_function_ids_);
```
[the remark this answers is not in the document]
dpallotti: > Same thing that happens if any first message is delayed for a time...

In my understanding, previously, `OnCaptureStarted` was called right after sending `CaptureRequest`, not after receiving the first event. So this is quite different. `OnCaptureStarted` is actually `OrbitApp::OnCaptureStarted`, which prepares the UI for the new capture, so I think that this is quite a change.

src/Service/CaptureServiceImpl.cpp
dpallotti: > I do not think initial thread names should be sent separately, we should have a snapshot of a process state at the start of the capture.

This was a minor detail. 

But I didn't really get an opinion on the main point, which is architectural and was the following: "This effectively makes LinuxTracing the "master" producer." In other words, this makes `LinuxTracing` indispensable and it makes all other producers wait for part of its initialization (dropping all their previous events).

This is a very strong requirement with potential future implications (e.g., a capture that collects data only from the Vulkan layer is not a valid one). So I want to understand if and why it's not possible to do it differently (for example, dedicated code separated from `LinuxTracing`, dedicated producer, etc).

src/OrbitClientModel/CaptureDeserializerLoadFuzzer.cpp
dpallotti: Do we need a TODO, or are you already working on that?

src/OrbitCaptureClient/CaptureEventProcessor.cpp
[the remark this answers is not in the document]
dpallotti: Discussed offline, my bad: we've already had a "starting" (but not started) capture state for a long time now.

src/GrpcProtos/capture.proto
```diff
@@ -286,6 +286,18 @@ message SystemMemoryUsage {
...
 }
 
+message CaptureStarted {
+  int32 process_id = 1;
+  string executable_path = 2;
+  string executable_build_id = 3;
+  uint64 capture_start_timestamp_ns = 4;
+  // The list of loaded modules of the process.
+  repeated ModuleInfo modules = 5;
+  // System-wide list of thread names.
```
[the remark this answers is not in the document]
dimitry-: these are moved to snapshot events

src/LinuxTracing/TracerThread.h
```diff
@@ -133,6 +132,7 @@ class TracerThread {
   bool trace_thread_state_;
   bool trace_gpu_driver_;
   std::vector<orbit_grpc_protos::TracepointInfo> instrumented_tracepoints_;
+  orbit_grpc_protos::CaptureOptions capture_options_;
```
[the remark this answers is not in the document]
dimitry-: this is reverted now.

src/LinuxTracing/LinuxTracingIntegrationTest.cpp
dimitry-: no longer part of LinuxTracingHandler

src/LinuxTracing/LinuxTracingIntegrationTest.cpp
dimitry-: if tid != puppet_pid, this one is always true, I restore the check in but now it is checking for tid.

src/LinuxTracing/LinuxTracingIntegrationTest.cpp
dimitry-: added a check

src/LinuxTracing/TracerThread.cpp
dimitry-: moved out of TracerThread

src/Service/CaptureServiceImpl.cpp
dimitry-: we now have snapshot events.

src/OrbitClientModel/CaptureDeserializer.cpp
```diff
@@ -178,9 +182,36 @@ ErrorMessageOr<CaptureListener::CaptureOutcome> LoadCaptureInfo(
...
+  capture_options->set_pid(capture_info.process().pid());
+  for (const auto& function : capture_info.instrumented_functions()) {
+    InstrumentedFunction instrumented_function;
+    instrumented_function.set_function_id(function.first);
+    instrumented_function.set_function_type(
+        CaptureClient::InstrumentedFunctionTypeFromOrbitType(function.second.orbit_type()));
+
+    instrumented_function.set_file_path(function.second.loaded_module_path());
+    const ModuleData* module_data =
+        module_manager->GetModuleByPath(function.second.loaded_module_path());
```
dimitry-, line 202: ok, this is the part where we need modules to be available right away. Without it we cannot construct functions. Sending snapshot after capture started event breaks this part.

I think I am going to revert everything now, sorry it did not work out.

src/Service/CaptureServiceImpl.cpp
dimitry-: and now we dont.

src/OrbitClientModel/CaptureDeserializer.cpp
dpallotti: I don't fully understand so let's discuss offline. Just a couple of notes:
- This feels like it should be solved differently and that we should be able to handle snapshots. One reason is that if we need all modules at the very beginning, this also doesn't work with module updates.
- Maybe this is just a problem of the current capture format that we are moving away from anyway and we can make it work in the new format?

dimitry-:
will return to this after fixing instrumented functions in capturedata.

src/Service/CaptureServiceImpl.cpp
dimitry-: and now we do again.

src/OrbitClientModel/CaptureDeserializerLoadFuzzer.cpp
dimitry-: #2103

src/OrbitClientModel/CaptureDeserializer.cpp
dimitry-: removed.

src/OrbitCaptureClient/include/OrbitCaptureClient/CaptureListener.h
dimitry-: removed.

dimitry-:
this is ready for review, ptal

src/LinuxTracing/TracerThread.cpp
```diff
@@ -590,12 +617,31 @@ void TracerThread::Startup() {
...
+  auto modules_or_error = orbit_elf_utils::ReadModules(target_pid_);
+  if (modules_or_error.has_value()) {
+    const std::vector<ModuleInfo>& modules = modules_or_error.value();
+    *modules_snapshot.mutable_modules() = {modules.begin(), modules.end()};
+    listener_->OnModulesSnapshot(std::move(modules_snapshot));
+  } else {
+    ERROR("Unable to load modules for %d: %s", target_pid_, modules_or_error.error().message());
+  }
+
+  // Get the initial thread names to construct CaptureStarted event and notify the listener_.
```
dpallotti, line 632: "to construct CaptureStarted event" is no longer correct.

src/Service/ProducerEventProcessor.cpp
```diff
@@ -328,6 +353,9 @@ void ProducerEventProcessorImpl::ProcessApiEvent(ApiEvent* api_event) {
 
 void ProducerEventProcessorImpl::ProcessEvent(uint64_t producer_id, ProducerCaptureEvent event) {
```
dpallotti, line 354: Do these additions need new tests in `ProducerEventProcessorTest` (I'm not sure if those tests cover everything or only the delicate parts)?

dpallotti (commented):
It looks good, I like it much better than before, thanks for making the changes.

The one thing I don't understand is why `capture_started_timestamp_ns` is passed to `LinuxTracingHandler` and `TracerThread` when it doesn't seem to be used (which is good, I think it shouldn't be used).

src/LinuxTracing/TracerThread.cpp
```diff
@@ -685,10 +731,11 @@ void TracerThread::ProcessOneRecord(PerfEventRingBuffer* ring_buffer) {
   }
 }
 
-void TracerThread::Run(const std::shared_ptr<std::atomic<bool>>& exit_requested) {
+void TracerThread::Run(const std::shared_ptr<std::atomic<bool>>& exit_requested,
+                       uint64_t capture_started_timestamp_ns) {
```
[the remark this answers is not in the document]
dimitry-: right... I think I wanted to implement delta encoding here and then decided to do it in separate PR. I can remove this for now.

dimitry-:
after offline discussion with @antonrohr - since module reloading as it is currently done could lead to unwanted crashes this needs to be addressed before the change goes it. I am converting it back to draft for now.

src/GrpcProtos/capture.proto
```diff
@@ -308,6 +308,27 @@ message SystemMemoryUsage {
...
 
+message ModulesSnapshot {
+  int32 pid = 1;
+  uint64 timestamp_ns = 2;
+  repeated ModuleInfo modules = 3;
+}
+
+message ThreadNamesSnapshot {
+  // This is system-wide snapshot
+  // this is why it does not have pid
```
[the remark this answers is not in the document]
dimitry-: done.

src/GrpcProtos/capture.proto
```diff
@@ -308,6 +308,27 @@ message SystemMemoryUsage {
...
+  uint64 timestamp_ns = 1;
+  repeated ThreadName thread_names = 2;
+}
+
+message CaptureStarted {
+  int32 process_id = 1;
+  string executable_path = 2;
+  string executable_build_id = 3;
+  uint64 capture_start_timestamp_ns = 4;
+  CaptureOptions capture_options = 7;
```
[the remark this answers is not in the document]
dimitry-: done.

src/LinuxTracing/TracerThread.cpp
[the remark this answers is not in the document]
dimitry-: removed.

src/LinuxTracing/TracerThread.cpp
dimitry-: removed.

src/Service/ProducerEventProcessor.cpp
dimitry-: added the tests.

antonrohr:
I will not get to reviewing this today. Its on my todo list for tomorrow morning

src/LinuxTracing/TracerThread.cpp
```diff
@@ -506,6 +512,27 @@ bool TracerThread::OpenInstrumentedTracepoints(const std::vector<int32_t>& cpus)
   return !tracepoint_event_open_errors;
 }
 
+static std::vector<ThreadName> RetrieveInitialThreadNamesSystemWide(uint64_t initial_timestamp_ns) {
```
dpallotti, line 515: Why was `RetrieveInitialThreadNamesSystemWideAndNotifyListener` changed to this static function? Couldn't just the logic be changed? This is now a bit inconsistent with `RetrieveInitialTidToPidAssociationSystemWide` and `RetrieveInitialThreadStatesOfTarget`.

src/OrbitCaptureClient/ApiEventProcessorTest.cpp
```diff
@@ -24,15 +24,18 @@ namespace {
 // Empty implementation of CaptureListener used as helper to reduce the complexity in subclasses.
 class EmptyCaptureListener : public CaptureListener {
  public:
-  void OnCaptureStarted(ProcessData&&,
-                        absl::flat_hash_map<uint64_t, orbit_grpc_protos::InstrumentedFunction>,
-                        TracepointInfoSet, absl::flat_hash_set<uint64_t>) override {}
+  void OnCaptureStarted(const orbit_grpc_protos::CaptureStarted& /*capture_started*/,
```
dpallotti, line 27: I like the parameter names here but this is now inconsistent with the other methods. I guess the parameter names should be added to the other methods, too.

src/Service/CaptureServiceImpl.cpp
```diff
@@ -193,6 +200,37 @@ static void StopInternalProducersAndCaptureStartStopListenersInParallel(
...
+  if (executable_path_or_error.has_value()) {
+    const std::string& executable_path = executable_path_or_error.value();
+    capture_started->set_executable_path(executable_path);
+
+    ErrorMessageOr<std::unique_ptr<orbit_elf_utils::ElfFile>> elf_file_or_error =
+        orbit_elf_utils::ElfFile::Create(executable_path);
+    if (elf_file_or_error.has_value()) {
+      capture_started->set_executable_build_id(elf_file_or_error.value()->GetBuildId());
+    } else {
+      ERROR("Unable to load module: %s", elf_file_or_error.error().message());
```
dpallotti, line 222: The module for the error is already in `elf_file_or_error.error().message()`, yes?

src/GrpcProtos/capture.proto
```diff
@@ -324,17 +344,20 @@ message ClientCaptureEvent {
...
     CallstackSample callstack_sample = 1;
+    CaptureStarted capture_started = 24;
     FunctionCall function_call = 2;
     GpuJob gpu_job = 3;
     GpuQueueSubmission gpu_queue_submission = 4;
     InternedCallstack interned_callstack = 5;
     InternedString interned_string = 18;
     InternedTracepointInfo interned_tracepoint_info = 19;
     IntrospectionScope introspection_scope = 20;
+    ModulesSnapshot modules_snapshot = 25;
```
dpallotti, line 355: Isn't `module_update_event` before `modules_snapshot` because of the space? Same below.

dpallotti (approved):
I guess it's useful if someone else also takes a look but for me it looks good. Thanks!

src/LinuxTracing/TracerThread.cpp
dimitry-: This was clang-tidy warning I believe telling that this function can be made status, since it does not use anything from the class. I can change others as well. I guess I did not notice because wasn't modifying them.

src/OrbitCaptureClient/ApiEventProcessorTest.cpp
dimitry-: this is for another change. I do not want to add too many "drive by" changes to this one.

src/GrpcProtos/capture.proto
dimitry-: sort disagrees
```
$ echo "modules_snapshot
> module_update" | sort
modules_snapshot
module_update

```

florian-kuebler (approved):
LGTM. Thanks!

src/LinuxTracing/TracerThread.cpp
dpallotti: What I meant is that before the function was also notifying the listener, hence using a member of the class. I guess you like it better now, but what I'm saying is that it wasn't really necessary as having a method with side effects instead was consistent with the other two methods I mentioned.

src/GrpcProtos/capture.proto
dpallotti: Which means that I disagree with sort for this...

src/OrbitGl/App.cpp
```diff
@@ -349,6 +342,16 @@ void OrbitApp::OnTracepointEvent(orbit_client_protos::TracepointEventInfo tracep
       is_same_pid_as_target);
 }
 
+void OrbitApp::OnModuleUpdate(uint64_t /*timestamp_ns*/, ModuleInfo module_info) {
+  module_manager_->AddOrUpdateModules({module_info});
+  GetMutableCaptureData().mutable_process()->AddOrUpdateModuleInfo(module_info);
+}
+
+void OrbitApp::OnModulesSnapshot(uint64_t /*timestamp_ns*/, std::vector<ModuleInfo> module_infos) {
+  module_manager_->AddOrUpdateModules(module_infos);
```
antonrohr, line 351: This call returns a `std::vector<ModuleData*>` of modules that are not loaded (do not contain symbols) anymore after this call. I think the `AddOrUpdateModules` method should be `[[nodiscard]]`, because this rare case needs to be taken care of. After this call, the functions data view might contain pointers to FunctionInfos that do not exist anymore. 

There is the method `OrbitApp::ReloadModules` which is the only place `AddOrUpdateModules` was called before. `ReloadModules` attempts to load the symbols again for modules unloaded, which should not be done here. So probably the best way forward here is to split `ReloadModules` into 2 parts, the first that does the proper unloading (including refreshing the dataviews) and a second part that loads symbols again. 

@beckerhe touched this `ReloadModules` most recently, and is planning to move symbol loading related code into a separate module.

src/OrbitGl/App.cpp
```diff
@@ -349,6 +342,16 @@ void OrbitApp::OnTracepointEvent(orbit_client_protos::TracepointEventInfo tracep
...
+  GetMutableCaptureData().mutable_process()->UpdateModuleInfos(module_infos);
```
antonrohr, line 352: Besides the process in capture data, there is also `ProcessData* process_` (`GetMutableTargetProcess()`) in OrbitApp (its a pointer to a process living in a `TargetConfiguration` which is owned by orbitmainwindow). 

I think on capture start the target process is copied into capture data. 

I guess the modules of this target process should also be updated (and refresh module list dataview). Maybe on capture end the target process should be overridden with the version from capture data.

Having these 2 processes is not great in the first place. I guess we could also do something about that, but that might entail a lot.  
Let me know what you think, I am also happy to chat about it offline

antonrohr (commented):
I am not fully done with the review, but I found a potential problem

src/OrbitGl/App.cpp
dimitry-: I think the process managed by the app should be updated once capture is done (or pereodically but outside of capture-process thread). It does not have any effect on what we are doing here anyways. I believe we cannot hook/unhook functions during capture right?

src/OrbitGl/App.cpp
dimitry-: Looking at the LiveFunctions it seems like this should result in one of following for hooked functions:
1. Service will (probably) stop sending events if module has been changed and was reloaded by the process. In which case we it does not matter whether or not we refresh the view.
2. Service will keep sending events because even though module was updated and loaded the old image is still there and the executable might still be calling it. In which case we want live view to have old functions in the list. Note that Live Function view keeps copies of FunctionInfos.

I will check again there could be problems with that but from the first look it seems like this is exactly what we want. we do not want to rehook the functions, because this will result in undefined behavior for the second case.

src/LinuxTracing/TracerThread.cpp
dimitry-: yes my personal preference is to do it this way, I think object-file static is more private than the private in the class (because it does not have to be in the header).  And looking at the other functions it seems they can be modified in the same way, they do not have to call the listener - it could be done on the caller side.

src/OrbitGl/App.cpp
dimitry-: after offline discussion - we still have a problem here with FunctionDataView because it uses pointers to FuctionInfos. And since in the case we have too many things that could go wrong or result in incorrect data in callstacks/sampling reports when a module without buildid gets reloaded - the decision here is to generate a error and stop/abort.

src/OrbitGl/App.cpp
```diff
@@ -352,6 +345,35 @@ void OrbitApp::OnTracepointEvent(orbit_client_protos::TracepointEventInfo tracep
...
+void OrbitApp::UpdateModulesAbortIfModuleWithoutBuildIdReloader(
+    absl::Span<const ModuleInfo> module_infos) {
+  auto updated_modules = module_manager_->AddOrUpdateModules(module_infos);
+
+  if (!updated_modules.empty()) {
+    std::string module_paths;
+    for (const auto* updated_module : updated_modules) {
+      if (!module_paths.empty()) module_paths += ", ";
+      module_paths += updated_module->file_path();
+      FATAL(
```
antonrohr, line 357: Display Error to user instead?
And then abort capture? There should be methods for that.

dimitry-: I will add a todo here.

src/OrbitGl/App.cpp
dimitry-: Since in this case we want to not unload the modules so that after aborting the capture the client is in the consistent state, I will do it in the follow-up change. It ends up being bigger then I expected.

src/OrbitGl/App.cpp
dimitry-: Since in this case we want to not unload the modules so that after aborting the capture the client is in the consistent state, I will do it in the follow-up change. It ends up being bigger then I expected.

PR #2081, opened 2021-03-16
Set status_code to SUCCESS by deault

IrinaShkviro:
Previously sent logs had UNKNOWN_STATUS by default, but it's quite hard
to analyze if we want to understand amount of "failed" and "not failed"
events. In most cases we the actual status code is SUCCESS, so this
commit updates the default value.

Test: start Orbit, check that status_code in ORBIT_INITIALIZED event is
"SUCCESS".


antonrohr (approved):
Thanks

PR #2082, opened 2021-03-16
Add missing return-keywords

beckerhe:
The oss-fuzz build complained about unused values and I assumed they were meant to be returned.


danielfenner (approved):
Thanks for catching! I had an identical issue in some other file and Dimitry found that one; I'll grep for "ErrorMessage" in my folder ...

beckerhe:
> Thanks for catching! I had an identical issue in some other file and Dimitry found that one; I'll grep for "ErrorMessage" in my folder ...

Sure thing. Thanks for the review!

PR #2087, opened 2021-03-17
Support for loading symbols from virtual modules

beckerhe:
Adds support for loading symbols from virtual modules in Orbit by doing the following steps:

1. Add a function that can create `ModuleInfo` by reading the ELF file from a buffer, instead of a file
2. Changing the `/proc/<pid>/maps` parsing code to consider all executable maps at first. When creating `ModuleInfo` objects, we will filter out all maps that aren't mapped ELF files in memory. Handling of file-backed maps stay the same. They are still read from filesystem
3. Extend `SymbolHelper` to allow considering the `.dynsym` section as a fallback symbol source.
4. Add function `OrbitApp::RetrieveVirtualModuleFromRemote` which can load virtual modules by using the `ReadProcessMemory` gRPC call which is currently used for the disassembly view as well. The resulting ELF will be written into the symbol cache and can be considered a regular ELF file from that point on.
5. Wire all those extra functions into the normal `LoadSymbols`-code path.

### DIFF
--- a/src/ElfUtils/LinuxMap.cpp
+++ b/src/ElfUtils/LinuxMap.cpp
@@ -27,8 +27,8 @@ namespace orbit_elf_utils {
 using orbit_elf_utils::ElfFile;
 using orbit_grpc_protos::ModuleInfo;
 
-ErrorMessageOr<ModuleInfo> CreateModule(const std::filesystem::path& module_path,
-                                        uint64_t start_address, uint64_t end_address) {
+ErrorMessageOr<ModuleInfo> CreateModuleFromFile(const std::filesystem::path& module_path,
+                                                uint64_t start_address, uint64_t end_address) {
   // This excludes mapped character or block devices.
   if (absl::StartsWith(module_path.string(), "/dev/")) {
     return ErrorMessage(absl::StrFormat(
@@ -67,6 +67,35 @@ ErrorMessageOr<ModuleInfo> CreateModule(const std::filesystem::path& module_path
   module_info.set_address_end(end_address);
   module_info.set_build_id(elf_file_or_error.value()->GetBuildId());
   module_info.set_load_bias(load_bias_or_error.value());
+  module_info.set_is_virtual_module(false);
+
+  return module_info;
+}
+
+ErrorMessageOr<ModuleInfo> CreateModuleFromBuffer(std::string module_name, std::string_view buffer,
+                                                  uint64_t start_address, uint64_t end_address) {
+  ErrorMessageOr<std::unique_ptr<ElfFile>> elf_file_or_error =
+      ElfFile::CreateFromBuffer({}, buffer.data(), buffer.size());
+  if (elf_file_or_error.has_error()) {
+    return ErrorMessage(
+        absl::StrFormat("Unable to load module: %s", elf_file_or_error.error().message()));
+  }
+
+  ErrorMessageOr<uint64_t> load_bias_or_error = elf_file_or_error.value()->GetLoadBias();
+  // Every loadable module contains a load bias.
+  if (load_bias_or_error.has_error()) {
+    return load_bias_or_error.error();
+  }
+
+  ModuleInfo module_info;
+  module_info.set_name(module_name);
+  module_info.set_file_path(std::move(module_name));
+  module_info.set_file_size(buffer.size());
+  module_info.set_address_start(start_address);
+  module_info.set_address_end(end_address);
+  module_info.set_build_id(elf_file_or_error.value()->GetBuildId());
+  module_info.set_load_bias(load_bias_or_error.value());
+  module_info.set_is_virtual_module(true);
 
   return module_info;
 }
@@ -119,7 +148,7 @@ ErrorMessageOr<std::vector<ModuleInfo>> ParseMaps(std::string_view proc_maps_dat
     if (!address_range.is_executable) continue;
 
     ErrorMessageOr<ModuleInfo> module_info_or_error =
-        CreateModule(module_path, address_range.start_address, address_range.end_address);
+        CreateModuleFromFile(module_path, address_range.start_address, address_range.end_address);
     if (module_info_or_error.has_error()) {
       ERROR("Unable to create module: %s", module_info_or_error.error().message());
       continue;
--- a/src/ElfUtils/LinuxMap.cpp
+++ b/src/ElfUtils/LinuxMap.cpp
@@ -83,8 +83,8 @@ ErrorMessageOr<ModuleInfo> CreateModuleFromBuffer(std::string module_name, std::
   ErrorMessageOr<std::unique_ptr<ElfFile>> elf_file_or_error =
       ElfFile::CreateFromBuffer({}, buffer.data(), buffer.size());
   if (elf_file_or_error.has_error()) {
-    return ErrorMessage(
-        absl::StrFormat("Unable to load module: %s", elf_file_or_error.error().message()));
+    return ErrorMessage(absl::StrFormat("Unable to load module \"%s\": %s", module_name,
+                                        elf_file_or_error.error().message()));
   }
 
   ErrorMessageOr<uint64_t> load_bias_or_error = elf_file_or_error.value()->GetLoadBias();
--- a/src/ElfUtils/LinuxMap.cpp
+++ b/src/ElfUtils/LinuxMap.cpp
@@ -21,6 +21,7 @@
 #include "ElfUtils/ElfFile.h"
 #include "OrbitBase/Logging.h"
 #include "OrbitBase/ReadFileToString.h"
+#include "OrbitBase/Result.h"
 
 namespace orbit_elf_utils {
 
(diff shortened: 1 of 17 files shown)


dpallotti (approved):

src/ElfUtils/LinuxMapTest.cpp
```diff
@@ -184,3 +185,30 @@ TEST(LinuxMap, ParseMaps) {
...
+  using orbit_grpc_protos::ModuleInfo;
+
+  const std::filesystem::path hello_world_path =
+      orbit_base::GetExecutableDir() / "testdata" / "hello_world_elf";
+  ASSERT_TRUE(std::filesystem::exists(hello_world_path));
+
+  constexpr uint64_t kStartAddress = 23;
+  constexpr uint64_t kEndAddress = 8004;
+  auto buffer_or_error = orbit_base::ReadFileToString(hello_world_path);
+  ASSERT_TRUE(buffer_or_error.has_value());
```
dimitry-, line 235: << buffer_or_error.error().message()

src/Service/ProcessUtils.cpp
```diff
@@ -0,0 +1,71 @@
...
+  if (map_entry.end_address <= map_entry.start_address) {
+    return ErrorMessage{
+        absl::StrFormat("Invalid address range for module \"%s\".", map_entry.module_path)};
+  }
+
+  const auto size = map_entry.end_address - map_entry.start_address;
+
+  // 8 MiB - That's also the limit that we support in the `ReadProcessMemory` gRPC endpoint.
+  // It's more than enough for reading [vdso] which is typically around 8 KiB.
+  if (size > 8 * 1024 * 1024) {
```
dimitry-, line 54: Passing large files over grpc channel could run into DEADLINE_EXCEEDED errors. I wonder if we still want to do that via dump to file and then scp like we do for other modules. scp does not have a deadline and we already have some progress indicators ready to use there.

src/OrbitClientData/include/OrbitClientData/ModuleData.h
```diff
@@ -32,6 +32,10 @@ class ModuleData final {
   [[nodiscard]] const std::string& build_id() const { return module_info_.build_id(); }
   [[nodiscard]] uint64_t load_bias() const { return module_info_.load_bias(); }
   [[nodiscard]] bool is_loaded() const;
+  [[nodiscard]] bool is_virtual_module() const { return module_info_.is_virtual_module(); }
+  [[nodiscard]] uint64_t address_start() const { return module_info_.address_start(); }
+  [[nodiscard]] uint64_t address_end() const { return module_info_.address_end(); }
```
dimitry-, line 37: why do we need address_start/address_end in module_info?

dimitry- (commented):
Maybe add another grpc call which will dump file to disk and return the path which could be used to scp?

src/Service/ProcessUtils.cpp
beckerhe: So you're saying the current limit for the  `GetProcessMemory` rpc call is already too high?

For my use-case I only need 8 KiB. Is that also going to be a problem? Supporting other potentially bigger virtual modules is something that we basically got for free, that's why I added that as well. But I could also remove that.

I feel that writing that to a file while the only use-case we know of won't rellay benefit from that, is a bit overkill.

src/OrbitClientData/include/OrbitClientData/ModuleData.h
beckerhe: That's needed to make the GetProcessMemory RPC call.

src/Service/ProcessUtilsTest.cpp
```diff
@@ -0,0 +1,67 @@
...
+  const auto maps = orbit_elf_utils::ParseMaps(maps_data.value());
+
+  const auto is_vdso = [](const orbit_elf_utils::MapEntry& entry) {
+    return entry.module_path == "[vdso]";
+  };
+
+  const auto it = std::find_if(maps.begin(), maps.end(), is_vdso);
+
+  if (it == maps.end()) {
+    GTEST_SKIP() << "The test process has no [vdso] module, so we can't test loading it.";
```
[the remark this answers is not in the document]
beckerhe: Good question. I saw that here: https://github.com/google/googletest/blob/master/googletest/test/gtest_skip_test.cc

src/ElfUtils/LinuxMapTest.cpp
beckerhe: done.

src/Service/ProcessUtilsTest.cpp
```diff
@@ -0,0 +1,67 @@
...
+#include <gtest/gtest.h>
+
+#include <algorithm>
+
+#include "ElfUtils/LinuxMap.h"
+#include "OrbitBase/ExecutablePath.h"
+#include "ProcessUtils.h"
+#include "module.pb.h"
+
+TEST(ProcessUtils, CreateProcessMemoryBackedModuleFromMapEntry) {
```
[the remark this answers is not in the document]
beckerhe: done.

src/Service/ProcessUtils.h
```diff
@@ -0,0 +1,28 @@
...
+#include "OrbitBase/Result.h"
+#include "module.pb.h"
+
+namespace orbit_service {
+
+// This function is similar to orbit_elf_utils::ReadModules, but it also takes virtual modules
+// into account.
+ErrorMessageOr<std::vector<orbit_grpc_protos::ModuleInfo>> ReadModulesFromProcMaps(int32_t pid);
+
+// CreateProcessMemoryBackedModuleFromMapEntry can create a ModuleInfo object by reading the ELF
```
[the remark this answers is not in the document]
beckerhe: done.

src/ElfUtils/LinuxMap.cpp
```diff
@@ -67,6 +67,35 @@ ErrorMessageOr<ModuleInfo> CreateModule(const std::filesystem::path& module_path
...
+  return module_info;
+}
+
+ErrorMessageOr<ModuleInfo> CreateModuleFromBuffer(std::string module_name, std::string_view buffer,
+                                                  uint64_t start_address, uint64_t end_address) {
+  ErrorMessageOr<std::unique_ptr<ElfFile>> elf_file_or_error =
+      ElfFile::CreateFromBuffer({}, buffer.data(), buffer.size());
+  if (elf_file_or_error.has_error()) {
+    return ErrorMessage(
+        absl::StrFormat("Unable to load module: %s", elf_file_or_error.error().message()));
```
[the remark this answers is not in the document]
beckerhe: done.

src/Service/ProcessUtilsTest.cpp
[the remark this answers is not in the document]
beckerhe: This was fixed here: https://github.com/google/googletest/pull/2517

I guess we need to update gtest at some point.

src/Service/ProcessUtils.cpp
dimitry-: I remember we were having deadline problems when reading symbols and sending them over the network, I think it could repeat itself, not for vdso but if user desided to load something bigger (and dynamically allocated).

beckerhe: I made some tests. And indeed 8 MiB are too large. There seems to be a limit in gRPC around 4 MiB.

So new I changed the limit to 3 MiB and made some tests by artificially hacking in a bigger size for vdso. Works fine on both of my machines.

So now the user will not see any modules larger than 3 MiB. I could also change the behaviour that modules are shown no matter which size, but we show an error message when the user tries to load symbols for them.

WDYT?

dpallotti: Isn't 4 MB the default limit for gRPC messages?

dpallotti: I remember that when we were sending symbols over gRPC we had to increase that limit somewhere.

beckerhe: > Isn't 4 MB the default limit for gRPC messages?

Yes, it seems so. I took the 8 MiB because the `GetProcessMemory` callback limits its result to 8 MiB, which would obviously not work and should be fixed as well.

dimitry- (approved):
the only reservation I have is timeouts, otherwise lgtm.

beckerhe:
> the only reservation I have is timeouts, otherwise lgtm.

So what do we consider a gRPC message size that does not produce timeouts? What is our limit?

dimitry-:
> > the only reservation I have is timeouts, otherwise lgtm.
> 
> So what do we consider a gRPC message size that does not produce timeouts? What is our limit?

If I had to guess I would say Mbytes are very risky, hundreds of kilobytes is something to be avoided. single digit kilobytes are fine.

Another thing here. If we go with file transfer we do not have to impose size restrictions on the user. Files can be large.

I suggest we merge it as is now but revisit soon (create a follow-up bug?)

dpallotti:
> > > the only reservation I have is timeouts, otherwise lgtm.
> > 
> > 
> > So what do we consider a gRPC message size that does not produce timeouts? What is our limit?
> 
> If I had to guess I would say Mbytes are very risky, hundreds of kilobytes is something to be avoided. single digit kilobytes are fine.
> 
> Another thing here. If we go with file transfer we do not have to impose size restrictions on the user. Files can be large.
> 
> I suggest we merge it as is now but revisit soon (create a follow-up bug?)

We have file transfer already in place so I don't know if this makes sense, but the gRPC-only way to do it is a streaming RPC.

beckerhe:
Okay, thanks. I'm out right now, but I can take of this PR in an hour or so.

beckerhe:
Closing this for now, since the same is achieved by Davide's change as well. I don't see a lot of additional benefit in this solution, at least when the virtual ELF file has no debug information or whatsoever.

PR #2089, opened 2021-03-18
Fix scroll position not being updated after filtering

reichlfl:
The height and position of the tracks was only updated on rendering, but the overall height of the TimeGraph is needed to determine the extent of the scrollbar.

Changed the behavior to have the height / pos calculation independent of the rendering.

Bug: b/182978978
To verify:
1) Take a capture
2) Scroll down in the capture window
3) Filter tracks

Before, the vertical scroll position was not updated. Now, it's updated.

### DIFF
--- a/src/OrbitGl/TimeGraph.cpp
+++ b/src/OrbitGl/TimeGraph.cpp
@@ -579,7 +579,7 @@ void TimeGraph::UpdatePrimitives(PickingMode picking_mode) {
 
   track_manager_->SortTracks();
   track_manager_->UpdateMovingTrackSorting();
-  track_manager_->UpdateTracks(&batcher_, min_tick, max_tick, picking_mode);
+  track_manager_->UpdateTrackPrimitives(&batcher_, min_tick, max_tick, picking_mode);
 
   update_primitives_requested_ = false;
 }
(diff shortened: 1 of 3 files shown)


src/OrbitGl/TimeGraph.cpp
```diff
@@ -579,7 +579,7 @@ void TimeGraph::UpdatePrimitives(PickingMode picking_mode) {
 
   track_manager_->SortTracks();
   track_manager_->UpdateMovingTrackSorting();
-  track_manager_->UpdateTracks(&batcher_, min_tick, max_tick, picking_mode);
+  track_manager_->UpdateTrackPrimitives(&batcher_, min_tick, max_tick, picking_mode);
```
karupayun, line 582: nice renaming!! :)

karupayun (approved):
LGTM! The division there in UpdateTrackPosition and UpdateTrackPrimitives is really nice!! :) :)


BTW, the bug (b/182978978) is still there. Maybe you solved another one, but if you move the screen too far below, there will be no information after writing in the filter tracks tab until you move the sliders.

reichlfl:
> LGTM! The division there in UpdateTrackPosition and UpdateTrackPrimitives is really nice!! :) :)
> 
> BTW, the bug (b/182978978) is still there. Maybe you solved another one, but if you move the screen too far below, there will be no information after writing in the filter tracks tab until you move the sliders.

As discussed offline, it's not as easy to reproduce as initially thought, and there may be some other side effects at play here. This PR fixes one clearly reproducible issue, but we'll leave the bug open for further investigation.

karupayun:
Seems to be fixed now.

PR #2090, opened 2021-03-18
Use DWARF's subroutine information to determine the beginning of a function in source code

beckerhe:
We used to use the line number of the first instruction to determine the beginning of a function. This can obviously go wrong and might not even be close to the beginning of a function in source code.

So this PR changes that and reads the subroutine attributions DW_AT_DECL_LINE and DW_AT_DECL_FILE to determine the actual beginning of a function in source.

This information is used to jump to the correct line in the source code view.

### DIFF
--- a/src/ElfUtils/ElfFile.cpp
+++ b/src/ElfUtils/ElfFile.cpp
@@ -12,6 +12,9 @@
 #include <llvm/ADT/Twine.h>
 #include <llvm/ADT/iterator_range.h>
 #include <llvm/BinaryFormat/ELF.h>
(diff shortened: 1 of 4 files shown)


karupayun (approved):
LGTM! But as I didn't read too much code in these files, I would like to also have Anton's approval.

antonrohr (approved):

antonrohr:
Looks good to me!. 
I merge this because @beckerhe is OOO

beckerhe:
Thanks Anton!

PR #2092, opened 2021-03-18
Less restrictive remember checkbox

antonrohr:
This enables the "Automatically connect to this instance next time" immediately after a selection was made. Before hand it was only enabled after the connection was established.


beckerhe (approved):
LGTM

PR #2093, opened 2021-03-18
Add custom unwindstack::Memory to support unwinding vDSO samples

dpallotti:
`StackAndProcessMemory` still carries the stack sample collected with
`perf_event_open` like done before with `CreateOfflineMemory`, but it also
allows to read online from the memory of the process.
This allows to unwind callstacks that involve virtual modules.
It really brings down the number of unwinding errors, on Trata from ~0.8 % to
almost zero, on Infiltrator from ~2.5% to ~0.3%.

Bug: http://b/182895433

Test: Tested on Trata and Infiltrator to see that callstacks still look good,
and that we get a reasonable number of `[vdso]` callstacks now (relatively a lot
of them on Infiltrator, actually).

### DIFF
--- a/src/LinuxTracing/LibunwindstackUnwinder.cpp
+++ b/src/LinuxTracing/LibunwindstackUnwinder.cpp
@@ -15,6 +15,56 @@
 
 namespace orbit_linux_tracing {
 
+namespace {
+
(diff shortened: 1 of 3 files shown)


dimitry- (approved):
Nice!

beckerhe (approved):
Yes, very nice! LGTM!

PR #2094, opened 2021-03-18
Reconnect to ended process

antonrohr:
This "reconnects" to a process that ended and was restarted. Whether its the same process is based on its path

I think everything else should still work after this change, but I have not extensively tested. If this makes it into the release, I will make sure to test it well. http://b/180580029


antonrohr:
Will reopen with green builds and better tested

PR #2095, opened 2021-03-18
Disable filter panels when no capture is present

reichlfl:
Bug: b/183111923

Filtering in the capture window causes a crash when no capture data is present. Solved by disabling the filtering panel before taking a capture as there is no use for it, and this avoids getting us into an inconsistent state.

In addition, this clears the filters when restarting a capture. This was a bug before:
1) Start a capture
2) Filter tracks
3) Restart capture

This leaves the filter in place, but it does not affect the displayed tracks.


karupayun (commented):
I'm still seeing the filter when starting orbit and if we write something there, we still have a crash in the same line.

track-filter


BTW, the second commit seems to solve another issue after restarting a capture that we will want to solve.

ronaldfw:
Merged as we need for this for the release we are branching today.

PR #2097, opened 2021-03-19
Enable ScopeTree data structure in ThreadTrack

pierricgimmig:
The new manual instrumentation requires the use of the ScopeTree data
structure to merge scope data coming from different sources. Add a
ThreadTrack specific "UpdatePrimitives" method that takes advantage of
the ScopeTree's sorted scopes for rendering timeslices.

Note: The ScopeTree was added at the ThreadTrack level because not all 
TimerTracks support it (the AsyncTrack for example does not have 
hierarchical relationship between scopes).

The UpdatePrimitives in ThreadTrack is similar to the one we had before 
the GPU overlap change. I tried to clean it up as much as possible for 
readability. The min_ignore/max_ignore were merged into a single 
next_pixel_start_time_ns value. @ronaldfw , please let me know if it respects
the spirit of the initial optimization.

Tested with a mix of manual and dynamic instrumentation (OrbitTest, Unreal)

I'd like to do a bit more profiling before this goes into a release.

### DIFF
--- a/src/OrbitGl/TextBox.h
+++ b/src/OrbitGl/TextBox.h
@@ -20,6 +20,10 @@ class TextBox {
 
   void SetSize(const Vec2& size) { size_ = size; }
   void SetPos(const Vec2& pos) { pos_ = pos; }
+  void SetPosAndSize(const Vec2& pos, const Vec2& size) {
+    pos_ = pos;
+    size_ = size;
+  }
 
   const Vec2& GetSize() const { return size_; }
   const Vec2& GetPos() const { return pos_; }
--- a/src/OrbitGl/TextBox.h
+++ b/src/OrbitGl/TextBox.h
@@ -45,6 +45,7 @@ class TextBox {
   // Start() and End() are required in order to be used as node in a ScopeTree.
   uint64_t Start() const { return timer_info_.start(); }
   uint64_t End() const { return timer_info_.end(); }
+  uint64_t Duration() const { return End() - Start(); }
 
  protected:
   Vec2 pos_;
--- a/src/OrbitGl/TextBox.h
+++ b/src/OrbitGl/TextBox.h
@@ -20,10 +20,6 @@ class TextBox {
 
   void SetSize(const Vec2& size) { size_ = size; }
   void SetPos(const Vec2& pos) { pos_ = pos; }
-  void SetPosAndSize(const Vec2& pos, const Vec2& size) {
-    pos_ = pos;
-    size_ = size;
-  }
 
   const Vec2& GetSize() const { return size_; }
   const Vec2& GetPos() const { return pos_; }
--- a/src/OrbitGl/ThreadTrack.cpp
+++ b/src/OrbitGl/ThreadTrack.cpp
@@ -417,7 +417,7 @@ void ThreadTrack::OnTimer(const TimerInfo& timer_info) {
   CHECK(timer_chain != nullptr);
   {
     absl::MutexLock lock(&scope_tree_mutex_);
-    scope_tree_.Insert(timer_chain->Last());
+    scope_tree_.Insert(timer_chain->GetLast());
   }
 }
 
@@ -453,13 +453,13 @@ void ThreadTrack::UpdatePrimitives(Batcher* batcher, uint64_t min_tick, uint64_t
   UpdatePrimitivesOfSubtracks(batcher, min_tick, max_tick, picking_mode, z_offset);
   UpdateBoxHeight();
 
-  internal::DrawData draw_data = GetDrawData(
+  const internal::DrawData draw_data = GetDrawData(
       min_tick, max_tick, z_offset, batcher, time_graph_, collapse_toggle_->IsCollapsed(),
       app_->selected_text_box(), app_->GetFunctionIdToHighlight());
 
   absl::MutexLock lock(&scope_tree_mutex_);
 
-  for (auto& [depth, ordered_nodes] : scope_tree_.GetOrderedNodesByDepth()) {
+  for (const auto& [depth, ordered_nodes] : scope_tree_.GetOrderedNodesByDepth()) {
     auto first_node_to_draw = ordered_nodes.lower_bound(min_tick);
     if (first_node_to_draw != ordered_nodes.begin()) --first_node_to_draw;
 
--- a/src/OrbitGl/ThreadTrack.cpp
+++ b/src/OrbitGl/ThreadTrack.cpp
@@ -102,12 +102,16 @@ std::string ThreadTrack::GetBoxTooltip(const Batcher& batcher, PickingId id) con
   const TimerInfo& timer_info = text_box->GetTimerInfo();
 
   const InstrumentedFunction* func =
-      capture_data_
-          ? capture_data_->GetInstrumentedFunctionById(timer_info.function_id())
-          : nullptr;
+      capture_data_ ? capture_data_->GetInstrumentedFunctionById(timer_info.function_id())
+                    : nullptr;
+
+  FunctionInfo::OrbitType type{FunctionInfo::kNone};
+  if (func != nullptr) {
+    type = function_utils::GetOrbitTypeByName(func->function_name());
+  }
 
   std::string function_name;
-  bool is_manual = (func != nullptr && func->orbit_type() == FunctionInfo::kOrbitTimerStart) ||
+  bool is_manual = (func != nullptr && type == FunctionInfo::kOrbitTimerStart) ||
                    timer_info.type() == TimerInfo::kApiEvent;
 
   if (!func && !is_manual) {
@@ -122,7 +126,7 @@ std::string ThreadTrack::GetBoxTooltip(const Batcher& batcher, PickingId id) con
   }
 
   std::string module_name =
-      func != nullptr ? function_utils::GetLoadedModuleName(*func->file_path()) : "unknown";
+      func != nullptr ? function_utils::GetLoadedModuleNameByPath(func->file_path()) : "unknown";
 
   return absl::StrFormat(
       "<b>%s</b><br/>"
--- a/src/OrbitGl/ThreadTrack.cpp
+++ b/src/OrbitGl/ThreadTrack.cpp
@@ -99,24 +99,30 @@ std::string ThreadTrack::GetBoxTooltip(const Batcher& batcher, PickingId id) con
     return "";
   }
 
+  const TimerInfo& timer_info = text_box->GetTimerInfo();
+
   const InstrumentedFunction* func =
       capture_data_
-          ? capture_data_->GetInstrumentedFunctionById(text_box->GetTimerInfo().function_id())
+          ? capture_data_->GetInstrumentedFunctionById(timer_info.function_id())
           : nullptr;
 
-  if (!func) {
+  std::string function_name;
+  bool is_manual = (func != nullptr && func->orbit_type() == FunctionInfo::kOrbitTimerStart) ||
+                   timer_info.type() == TimerInfo::kApiEvent;
+
+  if (!func && !is_manual) {
     return text_box->GetText();
   }
 
-  std::string function_name;
-  bool is_manual = func->function_type() == InstrumentedFunction::kTimerStart;
   if (is_manual) {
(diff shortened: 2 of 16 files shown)


src/OrbitGl/TimerTrack.h
```diff
@@ -46,6 +46,13 @@ struct DrawData {
   float z;
   bool is_collapsed;
 };
+
+struct Rect {
```
florian-kuebler, line 50: Nit: `Rectangle`, as per [style guide?](https://google.github.io/styleguide/cppguide.html#General_Naming_Rules)?

src/OrbitGl/ThreadTrack.h
```diff
@@ -56,19 +57,25 @@ class ThreadTrack final : public TimerTrack {
...
+  void UpdatePrimitivesOfSubtracks(Batcher* batcher, uint64_t min_tick, uint64_t max_tick,
+                                   PickingMode picking_mode, float z_offset);
   void UpdateMinMaxTimestamps();
 
   std::shared_ptr<orbit_gl::ThreadStateBar> thread_state_bar_;
   std::shared_ptr<orbit_gl::CallstackThreadBar> event_bar_;
   std::shared_ptr<orbit_gl::TracepointThreadBar> tracepoint_bar_;
+
+  absl::Mutex scope_tree_mutex_;
+  ScopeTree<TextBox> scope_tree_;
```
florian-kuebler, line 78: We should start using `GUARDED_BY` annotations for new mutex guarded elements: https://abseil.io/docs/cpp/guides/synchronization#annotation-guidelines

src/OrbitGl/ThreadTrack.cpp
```diff
@@ -405,3 +397,100 @@ float ThreadTrack::GetHeaderHeight() const {
...
+  UpdatePrimitivesOfSubtracks(batcher, min_tick, max_tick, picking_mode, z_offset);
+  UpdateBoxHeight();
+  visible_timer_count_ = 0;
+  internal::DrawData draw_data = GetDrawData(
+      min_tick, max_tick, z_offset, batcher, time_graph_, collapse_toggle_->IsCollapsed(),
+      app_->selected_text_box(), app_->GetFunctionIdToHighlight());
+
+  absl::MutexLock lock(&scope_tree_mutex_);
+
+  for (auto& [depth, ordered_nodes] : scope_tree_.GetOrderedNodesByDepth()) {
```
florian-kuebler, line 449: `const auto &`?

src/OrbitGl/ThreadTrack.cpp
```diff
@@ -162,6 +162,15 @@ bool ThreadTrack::IsTrackSelected() const {
   return ToColor(static_cast<uint64_t>(event.color));
 }
 
+Color ThreadTrack::GetTimerColor(const TextBox& text_box, const internal::DrawData& draw_data) {
+  const TimerInfo& timer_info = text_box.GetTimerInfo();
+  uint64_t function_id = timer_info.function_id();
```
florian-kuebler, line 167: Nit/Personal Preference: `const` for the variables?

src/OrbitGl/ThreadTrack.cpp
```diff
@@ -405,3 +397,100 @@ float ThreadTrack::GetHeaderHeight() const {
...
+    absl::MutexLock lock(&scope_tree_mutex_);
+    scope_tree_.Insert(timer_chain->Last());
+  }
+}
+
+[[nodiscard]] static inline internal::Rect GetTimerInfoRect(const TimerInfo& timer_info,
+                                                            const internal::DrawData& draw_data,
+                                                            const TimeGraph* time_graph,
+                                                            float world_pos_y, float world_size_y) {
+  double start_us = time_graph->GetUsFromTick(timer_info.start());
```
florian-kuebler, line 417: Nit/Personal Preference: `const` for the variables?

src/OrbitGl/ThreadTrack.cpp
```diff
@@ -405,3 +397,100 @@ float ThreadTrack::GetHeaderHeight() const {
...
+
+// We minimize overdraw when drawing lines for small events by discarding events that would just
+// draw over an already drawn pixel line. When zoomed in enough that all events are drawn as boxes,
+// this has no effect. When zoomed  out, many events will be discarded quickly.
+void ThreadTrack::UpdatePrimitives(Batcher* batcher, uint64_t min_tick, uint64_t max_tick,
+                                   PickingMode picking_mode, float z_offset) {
+  UpdatePrimitivesOfSubtracks(batcher, min_tick, max_tick, picking_mode, z_offset);
+  UpdateBoxHeight();
+  visible_timer_count_ = 0;
+  internal::DrawData draw_data = GetDrawData(
```
florian-kuebler, line 443: Again, const?

florian-kuebler (approved):
Thanks for integrating this. Code-wise, LGTM.

In general, I would highly prefer to have this change in a super-class (like TimerTrack). Maybe, the TimerTrack class should be split-up into several classes. As an example, this may be also useful for the Gpu timers (needs to be discussed). Also it feels strange to reimplement logic, that is essentially already there in the super class.

I would also appreciate a more general split between the Track properties, like drawing the name, the rounded background, etc., and the specific content. Currently I am working on a composed Track, and it feels kind of (As I need to call the subtrack's UpdratePrimitives, but not the subtrack's Draw method). Of course this is something that we should address in a different PR.

If we would have a more consistent split here, we could also move the ScopeTree logic to a super class of AsyncTrack, or GpuTrack. Those tracks would then be essentially composed tracks, with a variable number of subtracks.

src/OrbitGl/TextBox.h
```diff
@@ -20,6 +20,10 @@ class TextBox {
 
   void SetSize(const Vec2& size) { size_ = size; }
   void SetPos(const Vec2& pos) { pos_ = pos; }
+  void SetPosAndSize(const Vec2& pos, const Vec2& size) {
```
ronaldfw, line 23: Is this necessary for performance reasons? Could this be implemented using `SetSize` and `SetPos`?

src/OrbitGl/ThreadTrack.cpp
```diff
@@ -405,3 +397,100 @@ float ThreadTrack::GetHeaderHeight() const {
...
+          &text_box, [&, batcher](PickingId id) { return this->GetBoxTooltip(*batcher, id); });
+
+      if ((timer_info.end() - timer_info.start()) > draw_data.ns_per_pixel) {
+        SetTimesliceText(timer_info, draw_data.world_start_x, z_offset, &text_box);
+        batcher->AddShadedBox(rect.pos, rect.size, draw_data.z, color, std::move(user_data));
+      } else {
+        batcher->AddVerticalLine(rect.pos, rect.size[1], draw_data.z, color, std::move(user_data));
+      }
+
+      // Use the time at boundary of the next pixel as a threshold to avoid overdraw.
```
ronaldfw, line 476: Using only the next pixel at the boundary and discarding everything earlier looks easier and more elegant. The only issue I can see is that this requires that timers are sorted. 

Otherwise with something like the following draw order (assume that "1" is drawn first), "3" and "2" will not get drawn at all.
3  2  1
 |   |   | 

If we assume that timers are sorted by end time and don't overlap, then they are also, at the same depth, sorted by start time, which will avoid this problem. For GPU debug label timers, there's potentially a problem here, as we know they can overlap, but if this is only in `ThreadTrack`, it should work fine. In either case, it will likely be unnoticeable in practice but that argument is hand-wavy at best.

Of course, it would be good to verify there's no performance regression, but sounds like you are planning to do that anyway.

src/OrbitGl/TimerChain.h
```diff
@@ -108,6 +114,7 @@ class TimerChain {
   ~TimerChain();
 
   void push_back(const TextBox& item) { current_->Add(item); }
+  [[nodiscard]] TextBox* Last() { return current_->Last(); }
```
ronaldfw, line 117: Nit: Other methods that return a TextBox* are called `GetSomething`. Call this `GetLast`?

src/OrbitGl/TimerTrack.h
ronaldfw: `Rect` is widely accepted in graphics / UI frameworks as a shortening of Rectangle and we also use this elsewhere. For consistency and brevity I'd keep `Rect`.

ronaldfw (commented):
> If we would have a more consistent split here, we could also move the ScopeTree logic to a super class of AsyncTrack, or GpuTrack. Those tracks would then be essentially composed tracks, with a variable number of subtracks.

I agree the architecture needs some thoughts here. I think the consensus was that we want to avoid overriding functions like `UpdatePrimitives` with completely different behavior and rather have parallel implementations like `SimpleTimerTrack`, `ScopeTimerTrack`, `OverlapTimerTrack` that implement specifically what those tracks need. 


>  ScopeTree logic to a super class of AsyncTrack,

Specifically this we shouldn't do as AsyncTrack does not have a scoped hierarchy.

ronaldfw:
On profiling this change: Can there be pathological cases where ScopeTree becomes very slow compared with not using this data structure? Something with a very deep hierarchy at medium/low frequency so that the overall number of timers is still reasonably small, but so that inserting into the map becomes comparatively expensive. 

That said, I'm not sure we care unless this already happens at relatively shallow hierarchies. But it might be good to understand if something really bad can happen, as it means there's some tipping point beyond which performance degrades.

src/OrbitGl/TimerTrack.h
pierricgimmig: I actually removed the struct altogether when further cleaning up ThreadTrack::UpdatePrimitives.

src/OrbitGl/ThreadTrack.h
pierricgimmig: Interesting, yeah I think we should look into it. I feel I need to understand it a bit deeper before starting applying it however.

src/OrbitGl/ThreadTrack.cpp
pierricgimmig: yes! thanks

src/OrbitGl/ThreadTrack.cpp
pierricgimmig: I'm a bit on the fence about it, but why not! thanks

src/OrbitGl/ThreadTrack.cpp
pierricgimmig: Here I find it helps understand that the draw_data is read-only in a relatively big function where draw_data is used at multiple spots. I added const, thanks.

src/OrbitGl/ThreadTrack.cpp
pierricgimmig: same as above

src/OrbitGl/TextBox.h
pierricgimmig: It was to make the code shorter, but I don't need it anymore actually so I removed it in favor of using the individual SetPos and SetSize.

src/OrbitGl/TimerChain.h
pierricgimmig: Yes, thanks. I also moved the function alongside the other getters.

src/OrbitGl/ThreadTrack.cpp
pierricgimmig: Since we are using ScopeTree here, the timers are guaranteed to be sorted by start time. In the case of ThreadTrack, there is no overlap. But I agree that we need to keep that in mind for other tracks allowing overlap or not using the ScopeTree at all.

ronaldfw:
Let me know when/if I should take another look at this PR!

pierricgimmig:
> On profiling this change: Can there be pathological cases where ScopeTree becomes very slow compared with not using this data structure? Something with a very deep hierarchy at medium/low frequency so that the overall number of timers is still reasonably small, but so that inserting into the map becomes comparatively expensive.
> 
> That said, I'm not sure we care unless this already happens at relatively shallow hierarchies. But it might be good to understand if something really bad can happen, as it means there's some tipping point beyond which performance degrades.

I've been profiling the change and so far I see a slight advantage for the new implementation actually, which is a bit surprising. I will submit this PR and open a new one that allows to easily toggle between the two implementations for further investigation.

PR #2099, opened 2021-03-19
Combine Gpu debug marker and submission tracks into one.

florian-kuebler:
Combine Gpu debug marker and submission tracks into one.
    
    Prior this, we had two Gpu tracks per queue (when the Vulkan layer
    was enabled). However, those belong tightly together.
    
    This change makes `GpuTrack` being a composed track of `GpuDebugMarkerTrack`
    and `GpuSubmissionTrack`. It refactors out the logic of each kind of
    prior GpuTrack.
    
    In order to do so, tracks can be now transparent and may have an intent.
    
    Test: Take a capture with the Vulkan layer.
    Bug: http://b/182751932.

Screenshot 2021-03-19 at 14 55 30
Screenshot 2021-03-19 at 14 56 43

### DIFF
--- a/contrib/automation_tests/test_cases/capture_window.py
+++ b/contrib/automation_tests/test_cases/capture_window.py
@@ -29,10 +29,25 @@ class CaptureWindowE2ETestCaseBase(E2ETestCase):
         self._time_graph = self.find_control('Image', name='TimeGraph', parent=suite.top_window())
         super().execute(suite=suite)
 
-    def _find_tracks(self, name_filter: str = None):
-        tracks = self._time_graph.children()
-        if name_filter is not None:
-            tracks = filter(lambda track: fnmatch(track.texts()[0], name_filter), tracks)
+    def _find_tracks(self, name_filter: str = None, recursive: bool = False):
+        if not recursive:
+            tracks = self._time_graph.children()
+            if name_filter is not None:
+                tracks = filter(lambda track: fnmatch(track.texts()[0], name_filter), tracks)
+            return list(tracks)
+
+        work_list = self._time_graph.children()
+        tracks = list()
+
+        while work_list:
+            track = work_list.pop()
+            if name_filter is None or fnmatch(track.texts()[0], name_filter):
+                tracks.append(track)
+            # Find all sub tracks of this track. We identify a track by its first child being a tab.
+            for sub_track in track.children():
+                if self.find_control("TabItem", parent=sub_track, raise_on_failure=False):
+                    work_list.append(sub_track)
+
         return list(tracks)
 
 
@@ -125,6 +140,7 @@ class MatchTracks(CaptureWindowE2ETestCaseBase):
     """
     Verify that the existing visible tracks match the expected tracks
     """
+
     def _execute(self, expected_count: int = None, expected_names: List[str or Iterable[str]] = None,
                  allow_additional_tracks=False):
         """
@@ -212,6 +228,7 @@ class ExpandTrack(CollapsingTrackBase):
     """
     Click on a collapsed track's triangle toggle, verify that the height increased.
     """
+
     def _verify_height(self, prev_click_height: int, post_click_height: int, track_name: str):
         self.expect_true(prev_click_height < post_click_height, "Expanding track '{}' changed its height from {:d} to {:d}".format(
             track_name, prev_click_height, post_click_height))
@@ -221,6 +238,7 @@ class CollapseTrack(CollapsingTrackBase):
     """
     Click on an expanded track's triangle toggle, verify that the height decreased.
     """
+
     def _verify_height(self, prev_click_height: int, post_click_height: int, track_name: str):
         self.expect_true(prev_click_height > post_click_height, "Collapsing track '{}' changed its height from {:d} to {:d}".format(
             track_name, prev_click_height, post_click_height))
@@ -262,10 +280,11 @@ class Capture(E2ETestCase):
         logging.info('Opening "Capture Options" dialog')
         capture_options_button = self.find_control('Button', 'Capture Options', parent=capture_tab)
         capture_options_button.click_input()
-        
+
         capture_options_dialog = self.find_control('Window', 'Capture Options')
 
-        collect_thread_states_checkbox = self.find_control('CheckBox', 'Collect thread states', parent=capture_options_dialog)
+        collect_thread_states_checkbox = self.find_control('CheckBox', 'Collect thread states',
+                                                           parent=capture_options_dialog)
         if collect_thread_states_checkbox.get_toggle_state() != collect_thread_states:
             logging.info('Toggling "Collect thread states" checkbox')
             collect_thread_states_checkbox.click_input()
@@ -322,8 +341,8 @@ class CheckThreadStates(CaptureWindowE2ETestCaseBase):
 
 
 class CheckTimers(CaptureWindowE2ETestCaseBase):
-    def _execute(self, track_name_filter: str, expect_exists: bool = True):
-        tracks = self._find_tracks(track_name_filter)
+    def _execute(self, track_name_filter: str, expect_exists: bool = True, recursive: bool = False):
+        tracks = self._find_tracks(track_name_filter, recursive)
         self.expect_true(len(tracks) > 0, 'Found tracks matching "{}"'.format(track_name_filter))
         logging.info('Checking for timers in {} tracks'.format(len(tracks)))
         for track in tracks:
--- a/src/OrbitGl/AsyncTrack.cpp
+++ b/src/OrbitGl/AsyncTrack.cpp
@@ -34,8 +34,9 @@ using orbit_grpc_protos::InstrumentedFunction;
 AsyncTrack::AsyncTrack(CaptureViewElement* parent, TimeGraph* time_graph,
                        orbit_gl::Viewport* viewport, TimeGraphLayout* layout,
                        const std::string& name, OrbitApp* app,
-                       const orbit_client_model::CaptureData* capture_data)
-    : TimerTrack(parent, time_graph, viewport, layout, app, capture_data) {
+                       const orbit_client_model::CaptureData* capture_data,
+                       uint32_t indentation_level)
+    : TimerTrack(parent, time_graph, viewport, layout, app, capture_data, indentation_level) {
   SetName(name);
   SetLabel(name);
 }
--- a/src/OrbitGl/GpuDebugMarkerTrack.cpp
+++ b/src/OrbitGl/GpuDebugMarkerTrack.cpp
@@ -1,0 +1,141 @@
+// Copyright (c) 2021 The Orbit Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#include "GpuDebugMarkerTrack.h"
+
+#include <absl/time/time.h>
+
+#include <memory>
+
+#include "App.h"
+#include "Batcher.h"
+#include "CoreUtils.h"
+#include "GlCanvas.h"
+#include "GlUtils.h"
+#include "OrbitBase/Logging.h"
+#include "OrbitBase/ThreadConstants.h"
+#include "TimeGraph.h"
+#include "TimeGraphLayout.h"
+#include "TimerChain.h"
+#include "TriangleToggle.h"
+#include "absl/strings/str_format.h"
+
+using orbit_client_protos::TimerInfo;
+
+GpuDebugMarkerTrack::GpuDebugMarkerTrack(CaptureViewElement* parent, TimeGraph* time_graph,
+                                         orbit_gl::Viewport* viewport, TimeGraphLayout* layout,
+                                         OrbitApp* app,
+                                         const orbit_client_model::CaptureData* capture_data,
+                                         uint32_t indentation_level)
+    : TimerTrack(parent, time_graph, viewport, layout, app, capture_data, indentation_level) {
+  SetLabel("Debug Markers");
+  draw_background_ = false;
+  string_manager_ = app->GetStringManager();
+
+  // Gpu tracks are collapsed by default.
+  collapse_toggle_->SetState(TriangleToggle::State::kCollapsed,
+                             TriangleToggle::InitialStateUpdate::kReplaceInitialState);
+}
+
+std::string GpuDebugMarkerTrack::GetTooltip() const {
+  return "Shows execution times for Vulkan debug markers";
+}
+
+Color GpuDebugMarkerTrack::GetTimerColor(const TimerInfo& timer_info, bool is_selected,
+                                         bool is_highlighted) const {
+  CHECK(timer_info.type() == TimerInfo::kGpuDebugMarker);
+  const Color kInactiveColor(100, 100, 100, 255);
+  const Color kSelectionColor(0, 128, 255, 255);
+  if (is_highlighted) {
+    return TimerTrack::kHighlightColor;
+  }
+  if (is_selected) {
+    return kSelectionColor;
+  }
+  if (!IsTimerActive(timer_info)) {
+    return kInactiveColor;
+  }
+  if (timer_info.has_color()) {
+    CHECK(timer_info.color().red() < 256);
+    CHECK(timer_info.color().green() < 256);
+    CHECK(timer_info.color().blue() < 256);
+    CHECK(timer_info.color().alpha() < 256);
+    return Color(static_cast<uint8_t>(timer_info.color().red()),
+                 static_cast<uint8_t>(timer_info.color().green()),
+                 static_cast<uint8_t>(timer_info.color().blue()),
+                 static_cast<uint8_t>(timer_info.color().alpha()));
+  }
+  std::string marker_text = string_manager_->Get(timer_info.user_data_key()).value_or("");
+  return TimeGraph::GetColor(marker_text);
+}
+
+void GpuDebugMarkerTrack::SetTimesliceText(const TimerInfo& timer_info, float min_x, float z_offset,
+                                           TextBox* text_box) {
+  CHECK(timer_info.type() == TimerInfo::kGpuDebugMarker);
+
+  if (text_box->GetText().empty()) {
+    std::string time = GetPrettyTime(absl::Nanoseconds(timer_info.end() - timer_info.start()));
+    text_box->SetElapsedTimeTextLength(time.length());
+    std::string text = absl::StrFormat(
+        "%s  %s", string_manager_->Get(timer_info.user_data_key()).value_or(""), time.c_str());
+    text_box->SetText(text);
+  }
+
+  const Color kTextWhite(255, 255, 255, 255);
+  const Vec2& box_pos = text_box->GetPos();
+  const Vec2& box_size = text_box->GetSize();
+  float pos_x = std::max(box_pos[0], min_x);
+  float max_size = box_pos[0] + box_size[0] - pos_x;
+  text_renderer_->AddTextTrailingCharsPrioritized(
+      text_box->GetText().c_str(), pos_x, text_box->GetPos()[1] + layout_->GetTextOffset(),
+      GlCanvas::kZValueBox + z_offset, kTextWhite, text_box->GetElapsedTimeTextLength(),
+      layout_->CalculateZoomedFontSize(), max_size);
+}
+
+std::string GpuDebugMarkerTrack::GetBoxTooltip(const Batcher& batcher, PickingId id) const {
+  const TextBox* text_box = batcher.GetTextBox(id);
+  if (text_box != nullptr) {
+    return "";
+  }
+
+  const TimerInfo& timer_info = text_box->GetTimerInfo();
+  CHECK(timer_info.type() == TimerInfo::kGpuDebugMarker);
+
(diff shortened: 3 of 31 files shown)


florian-kuebler:
It collapses the complete subtracks to just displaying the "hw_execution" timers. Which is actually default. See in the first screenshot the compute queue. And in contrast to this the graphics queue and the DMA queue on the same screenshot.

I think it makes sense to have both, as the default, which is not showing a lot is much more clean.

ronaldfw:
> It collapses the complete subtracks to just displaying the "hw_execution" timers. Which is actually default. See in the first screenshot the compute queue. And in contrast to this the graphics queue and the DMA queue on the same screenshot.
> 
> I think it makes sense to have both, as the default, which is not showing a lot is much more clean.

Ok,makes sense!

florian-kuebler:
I've set the font size of the subtrack labels to 90% of the parent's size:

Screenshot 2021-03-22 at 15 32 08

florian-kuebler:
Made this a draft request, as the accessibility interface is not implemented correctly.

src/OrbitGl/GpuTrack.h
```diff
@@ -31,50 +33,48 @@ std::string MapGpuTimelineToTrackLabel(std::string_view timeline);
 
 }  // namespace orbit_gl
 
-class GpuTrack : public TimerTrack {
+// A track to display Gpu related information (of a certain command queue). It contains two,
```
reichlfl, line 36: Nit: Remove the comma at the end of the line

src/OrbitGl/GpuTrack.cpp
```diff
@@ -54,312 +46,234 @@ std::string MapGpuTimelineToTrackLabel(std::string_view timeline) {
...
-    CHECK(timer_info.color().alpha() < 256);
-    return Color(static_cast<uint8_t>(timer_info.color().red()),
-                 static_cast<uint8_t>(timer_info.color().green()),
-                 static_cast<uint8_t>(timer_info.color().blue()),
-                 static_cast<uint8_t>(timer_info.color().alpha()));
-  }
-  if (timer_info.type() == TimerInfo::kGpuDebugMarker) {
-    std::string marker_text = string_manager_->Get(timer_info.user_data_key()).value_or("");
-    return TimeGraph::GetColor(marker_text);
+void GpuTrack::UpdatePositionOnfSubtracks() {
```
reichlfl, line 110: Typo in method name (O**n**fSubtracks)

src/OrbitGl/GpuTrack.cpp
```diff
@@ -54,312 +46,234 @@ std::string MapGpuTimelineToTrackLabel(std::string_view timeline) {
...
-                                float z_offset, TextBox* text_box) {
-  if (text_box->GetText().empty()) {
-    std::string time = GetPrettyTime(absl::Microseconds(elapsed_us));
+void GpuTrack::Draw(GlCanvas* canvas, PickingMode picking_mode, float z_offset) {
+  float track_height = GetHeight();
+  float track_width = canvas->GetWorldWidth();
 
-    text_box->SetElapsedTimeTextLength(time.length());
+  SetPos(canvas->GetWorldTopLeftX(), pos_[1]);
+  SetSize(track_width, track_height);
```
reichlfl, line 152: I wonder if all of this is really needed... You probably just copied this from the other Tracks, but this seems somewhere between unneeded and broken:
* `GetHeight()` is only correct after drawing, so this is incorrect for at least one frame
* `SetPos` doesn't change anything since `canvas->GetWorldTopLeftX()` always returns 0

But if I'm not mistaken all Tracks suffer from this, so let's keep it consistent and clean it up at some point.

src/OrbitGl/GpuTrack.cpp
```diff
@@ -54,312 +46,234 @@ std::string MapGpuTimelineToTrackLabel(std::string_view timeline) {
...
 
-    CHECK(timer_info.type() == TimerInfo::kGpuActivity ||
-          timer_info.type() == TimerInfo::kGpuCommandBuffer ||
-          timer_info.type() == TimerInfo::kGpuDebugMarker);
+  UpdatePositionOnfSubtracks();
```
reichlfl, line 154: This looks dangerous: When the subtrack positions change, `UpdatePrimitives` is needed. So either this is not required here, or you could be missing calls to `UpdatePrimitives`.

src/OrbitGl/GpuTrack.h
```diff
@@ -31,50 +33,48 @@ std::string MapGpuTimelineToTrackLabel(std::string_view timeline);
...
-                        float z_offset, TextBox* text_box) override;
-  [[nodiscard]] std::string GetBoxTooltip(const Batcher& batcher, PickingId id) const override;
+  [[nodiscard]] int GetVisiblePrimitiveCount() const override { return visible_timer_count_; }
 
  private:
+  void UpdatePositionOnfSubtracks();
+  const std::shared_ptr<GpuSubmissionTrack> submission_track_;
+  const std::shared_ptr<GpuDebugMarkerTrack> marker_track_;
+
+  int visible_timer_count_ = 0;
```
reichlfl, line 75: Is this needed? I guess it's for the Accessibility, but it would be less error-prone to not cache this here but simply query directly from the children in `GetVisiblePrimitiveCount()`.

reichlfl (commented):
I did a first review mostly of the combined track and left some comments. I did not look too much into the GpuDebugMarkerTrack, GpuSubmissionTrack, and tests yet.

florian-kuebler:
Added intent-based behaviour:
Screenshot 2021-04-07 at 11 41 54

florian-kuebler (commented):
I reworked the PR and adjusted the E2E test.

contrib/automation_tests/test_cases/capture_window.py
```diff
@@ -125,6 +140,7 @@ class MatchTracks(CaptureWindowE2ETestCaseBase):
     """
     Verify that the existing visible tracks match the expected tracks
     """
+
```
ronaldfw, line 143: Nit: Was this auto-formatted? I would normally minimize whitespace changes.

src/OrbitGl/AsyncTrack.cpp
```diff
@@ -31,8 +31,9 @@ using orbit_client_protos::TimerInfo;
 using orbit_grpc_protos::InstrumentedFunction;
 
 AsyncTrack::AsyncTrack(CaptureViewElement* parent, TimeGraph* time_graph, TimeGraphLayout* layout,
-                       const std::string& name, OrbitApp* app, const CaptureData* capture_data)
-    : TimerTrack(parent, time_graph, layout, app, capture_data) {
+                       const std::string& name, OrbitApp* app, const CaptureData* capture_data,
+                       uint32_t intent)
```
ronaldfw, line 35: I don't understand what intent here means. What's the purpose?

src/OrbitGl/TriangleToggle.cpp
```diff
@@ -17,13 +17,15 @@
...
-                               TimeGraph* time_graph, TimeGraphLayout* layout, Track* track)
+                               TimeGraph* time_graph, TimeGraphLayout* layout, Track* track,
+                               uint32_t intent)
     : CaptureViewElement(track, time_graph, layout),
       track_(track),
       state_(initial_state),
       initial_state_(initial_state),
       handler_(std::move(handler)) {
-  SetSize(10.f, 10.f);
+  uint32_t capped_intent = (intent > 5) ? 5 : intent;
```
ronaldfw, line 27: Ah, is this "indentation"?

src/OrbitGl/Track.cpp
```diff
@@ -110,11 +111,14 @@ void Track::Draw(GlCanvas* canvas, PickingMode picking_mode, float z_offset) {
   float half_label_width = 0.5f * label_width;
   float tab_x0 = x0 + layout_->GetTrackTabOffset();
 
-  Box box(Vec2(tab_x0, y0), Vec2(label_width, -label_height), track_z);
-  ui_batcher->AddBox(box, kBackgroundColor, shared_from_this());
+  const float intent_x0 = tab_x0 + (intent_ * layout_->GetTrackIntentOffset());
+  if (!transparent_) {
```
ronaldfw, line 115: Why is the transparency needed and in particular the case distinction? Or, ask differently: Why is transparency not simply an alpha value in the color?

src/OrbitGl/GpuTrack.cpp
```diff
@@ -54,325 +46,232 @@ std::string MapGpuTimelineToTrackLabel(std::string_view timeline) {
...
+    case TimerInfo::kGpuDebugMarker:
+      return marker_track_->GetDown(textbox);
+    default:
+      UNREACHABLE();
+  }
+}
+void GpuTrack::OnCollapseToggle(TriangleToggle::State state) {
+  Track::OnCollapseToggle(state);
+  if (state == TriangleToggle::State::kCollapsed) {
+    submission_track_->Collapse();
```
ronaldfw, line 275: Only the submission track or also the marker track?

src/OrbitGl/GpuSubmissionTrack.h
```diff
@@ -0,0 +1,76 @@
...
+// different submissions that would overlap, they are displayed on a different depth with a thin gap
+// between them.
+// This track is meant to be used as a subtrack of `GpuTrack`.
+class GpuSubmissionTrack : public TimerTrack {
+ public:
+  explicit GpuSubmissionTrack(CaptureViewElement* parent, TimeGraph* time_graph,
+                              TimeGraphLayout* layout, uint64_t timeline_hash, OrbitApp* app,
+                              const CaptureData* capture_data, uint32_t intent);
+  ~GpuSubmissionTrack() override = default;
+  [[nodiscard]] Type GetType() const override { return Type::kGpuTrack; }
```
ronaldfw, line 38: Should this have its own type (kGpuSubTrack)?

src/OrbitGl/GpuSubmissionTrack.cpp
```diff
@@ -0,0 +1,294 @@
...
+
+float GpuSubmissionTrack::GetYFromTimer(const TimerInfo& timer_info) const {
+  auto adjusted_depth = static_cast<float>(timer_info.depth());
+  if (collapse_toggle_->IsCollapsed()) {
+    adjusted_depth = 0.f;
+  }
+  CHECK(timer_info.type() == TimerInfo::kGpuActivity ||
+        timer_info.type() == TimerInfo::kGpuCommandBuffer);
+
+  // We are drawing a small gap between each depths, as the timers of different depths do not
```
ronaldfw, line 128:
```suggestion
  // We are drawing a small gap between each depth, as the timers of different depths do not
```

src/OrbitGl/GpuSubmissionTrack.cpp
```diff
@@ -0,0 +1,294 @@
...
+  // correlate. There won't be a gap between "hw execution"timers and command buffer timers, which
```
ronaldfw, line 129: What does "do not correlate" mean here?

ronaldfw (commented):
Added a bunch of comments, but didn't look at all code in detail. Overall agree with this change, the new tracks look really nice.

src/OrbitGl/AsyncTrack.cpp
ronaldfw: Ok, I think this is a typo and should really be "indentation"?

src/OrbitGl/GpuSubmissionTrack.h
florian-kuebler: Honestly, there simply shouldn't be any value called type. The type was used by the track manager/timegraph to separate tracks and to pass the "correct" timers to them.  This has however already changed, and should be done by the actual runtime type of the track. It is currently only used by the "AddFrameTrack" functions to insert the track at the desired position.

Anyways, sorry for the rant. There is no need for subtracks to have a type, I am going to use the already existing `kUnknown`, in both subtracks.

src/OrbitGl/GpuTrack.cpp
florian-kuebler: Exactly. This is somewhat of a hack. When the complete GpuTrack is collapsed, it looks exactly the same way as it would currently look if you collapse a submission track. In order to do so, we collapse the subtrack, such that the timers get collapsed.

src/OrbitGl/Track.cpp
florian-kuebler: The meaning of is `transparent` is: "Should the track's background (+tab background and rounded corners) be drawn". I could rename it to `draw_background`? Not sure if this is better.
Of course, we could simply adjust the kBackgroundColor in this case, to use an alpha of 0, but I don't feel that this makes things better. WDYT? I am happy to adjust this here.

src/OrbitGl/TriangleToggle.cpp
florian-kuebler: yap. Sorry for the confusion.

florian-kuebler (commented):
Thanks for the review, I addressed the comments.

src/OrbitGl/GpuTrack.cpp
ronaldfw: If this is somewhat of a hack, you should add a comment explaining why we are doing what we are doing here.

src/OrbitGl/GpuSubmissionTrack.h
ronaldfw: Add a comment explaining the choice of kUnknown.

contrib/automation_tests/test_cases/capture_window.py
ronaldfw: Ok, then just keep it.

reichlfl (commented):
I've added some comments - I didn't yet fully go through everything, sorry, just wanted to make sure the feedback isn't lost.

src/OrbitGl/Track.cpp
ronaldfw: I think `draw_background` is better as it makes the intent clearer. Transparency is closely related to alpha value and is easy to misunderstand. 

Actually, I do like using alpha of 0 even more, as it simplifies the code structure. You should need fewer case distinctions which is usually a win.

reichlfl: It should not make a difference here I assume, but I want to point out that alpha blending is disabled when rendering in picking mode.

src/OrbitGl/GpuDebugMarkerTrack.cpp
```diff
@@ -0,0 +1,136 @@
...
+    return kSelectionColor;
+  }
+  if (!IsTimerActive(timer_info)) {
+    return kInactiveColor;
+  }
+  if (timer_info.has_color()) {
+    CHECK(timer_info.color().red() < 256);
+    CHECK(timer_info.color().green() < 256);
+    CHECK(timer_info.color().blue() < 256);
+    CHECK(timer_info.color().alpha() < 256);
```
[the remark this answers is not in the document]
florian-kuebler: Since TimerInfo is simply a proto, we can only check this on the usage sides (unless we create a wrapper).

Re: ColorFromProtoColor, I kind of agree. However, my question would be where to put it. Currently we do have a couple of static GetColor functions in TimeGraph. But I would be afraid of adding more noise to the timegraph. Refactoring those out would be out of scope (as the scope of this PR is already a quite large).

src/OrbitGl/GpuTrack.cpp
florian-kuebler: Done.

src/OrbitGl/Track.cpp
florian-kuebler: I renamed the member to `draw_background_` and let it simply adjust the the background color's alpha value.

src/OrbitGl/GpuSubmissionTrack.h
florian-kuebler: Done.

src/OrbitGl/AsyncTrack.cpp
```diff
@@ -31,8 +31,9 @@ using orbit_client_protos::TimerInfo;
 using orbit_grpc_protos::InstrumentedFunction;
 
 AsyncTrack::AsyncTrack(CaptureViewElement* parent, TimeGraph* time_graph, TimeGraphLayout* layout,
-                       const std::string& name, OrbitApp* app, const CaptureData* capture_data)
-    : TimerTrack(parent, time_graph, layout, app, capture_data) {
+                       const std::string& name, OrbitApp* app, const CaptureData* capture_data,
+                       uint32_t indentation_level)
+    : TimerTrack(parent, time_graph, layout, app, capture_data, indentation_level) {
```
ronaldfw, line 36: Is indentation level the same as the level in the hierarchy of tracks? That is, in our case the main track has indentation_level 0, and the two subtracks each have indentation level 1 (and if they had subtracks they would have indentation level 2)?

src/OrbitGl/TriangleToggle.cpp
```diff
@@ -17,13 +17,15 @@
...
-                               TimeGraph* time_graph, TimeGraphLayout* layout, Track* track)
+                               TimeGraph* time_graph, TimeGraphLayout* layout, Track* track,
+                               uint32_t indentation_level)
     : CaptureViewElement(track, time_graph, layout),
       track_(track),
       state_(initial_state),
       initial_state_(initial_state),
       handler_(std::move(handler)) {
-  SetSize(10.f, 10.f);
+  uint32_t capped_indentation_level = (indentation_level > 5) ? 5 : indentation_level;
```
ronaldfw, line 27: Why is there a cap at 5?

And I would write this as
```
constexpr uint32_t kMaxIndentationLevel = 5; 
uint32_t capped_indentation_level = std::min(indentation_level, kMaxIndentationLevel);
```

src/OrbitGl/TriangleToggle.cpp
```diff
@@ -17,13 +17,15 @@
 #include "Track.h"
 
 TriangleToggle::TriangleToggle(State initial_state, StateChangeHandler handler,
-                               TimeGraph* time_graph, TimeGraphLayout* layout, Track* track)
+                               TimeGraph* time_graph, TimeGraphLayout* layout, Track* track,
+                               uint32_t indentation_level)
```
ronaldfw, line 21: Should the `TriangleToggle` class "know" about indentation level? This seems like an unnecessary coupling of meaning. Just pass the size here?

src/OrbitGl/AsyncTrack.cpp
florian-kuebler: Exactly.

src/OrbitGl/TriangleToggle.cpp
florian-kuebler: Otherwise, the shape is simply to small. I added a comment.

src/OrbitGl/TriangleToggle.cpp
florian-kuebler: Good idea. Done.

src/OrbitGl/Track.cpp
```diff
@@ -157,15 +167,15 @@ void Track::Draw(GlCanvas* canvas, PickingMode picking_mode, float z_offset) {
...
-  collapse_toggle_->Draw(canvas, picking_mode, z_offset);
+  const float toggle_y_pos = DrawCollapsingTriangle(canvas, picking_mode, z_offset);
 
   // Draw label.
   if (!picking) {
     uint32_t font_size = layout_->CalculateZoomedFontSize();
+    // For the first 10 indentations, we decrease the font_size by 10 percent points (per
+    // indentation).
+    uint32_t capped_intent = (indentation_level_ > 9) ? 10 : indentation_level_;
+    font_size = (font_size * (10 - capped_intent)) / 10;
```
ronaldfw, line 178: So, starting at indentation level 10, the font size will be 0. Is this intended? And even at earlier indentation levels, the font size will be really small. 

I'm also wondering: We are only using indentation level 0 and 1 right now (correct?). Should we just remove any logic that deals with higher indentation levels?

src/OrbitGl/Track.cpp
```diff
@@ -174,7 +184,7 @@ void Track::Draw(GlCanvas* canvas, PickingMode picking_mode, float z_offset) {
         IsTrackSelected() ? GlCanvas::kTabTextColorSelected : Color(255, 255, 255, 255);
 
     text_renderer.AddTextTrailingCharsPrioritized(
-        label_.c_str(), tab_x0 + label_offset_x, toggle_y_pos - label_offset_y, text_z, kColor,
+        label_.c_str(), intent_x0 + label_offset_x, toggle_y_pos - label_offset_y, text_z, kColor,
```
ronaldfw, line 187: Typo?

ronaldfw (approved):
LGTM, with two smaller comments left. Note that I have not tested this change.

src/OrbitGl/Track.cpp
florian-kuebler: Agree. I will adjust this to be capped at 5 indentations (so the same as for the triangle toggle). I would still argue that it makes sense to keep the code being flexible. It does not make the code more complicated and in the future we might want to have tracks with more indentation (e.g. for the pipeline stats).

src/OrbitGl/Track.cpp
florian-kuebler: Ah, thanks.

PR #2101, opened 2021-03-19
[Release (maybe)] Quick Fix for hot reloading

antonrohr:
This fixes an issue when reloading a module that changed.
It is not trivial why this works and its not a good fix. 

@beckerhe We should talk about this offline on Monday morning


beckerhe (approved):
LGTM

antonrohr:
I just checked, and this is a regression. It does not crash in 1.60

PR #2102, opened 2021-03-19
Fix a crash caused by wrongly used mutex

karupayun:
We used to have an only mutex in TimeGraph and we splitted into two when
we create TrackManager. In this PR, we moved the problematic functions
(GetNumTimers and GetTracksMinMaxTime) to TrackManager to mutex them properly.

Solve http://b/179467994.

### DIFF
--- a/src/OrbitGl/TrackManager.cpp
+++ b/src/OrbitGl/TrackManager.cpp
@@ -403,3 +403,31 @@ FrameTrack* TrackManager::GetOrCreateFrameTrack(uint64_t function_id,
 
   return track.get();
 }
+
+uint32_t TrackManager::GetNumTimers() const {
+  uint32_t num_timers = 0;
+  std::lock_guard<std::recursive_mutex> lock(mutex_);
+  for (const auto& track : GetAllTracks()) {
+    num_timers += track->GetNumTimers();
+  }
+  // Frame tracks are removable by users and cannot simply be thrown into the
+  // tracks_ vector.
+  for (const auto& track : GetFrameTracks()) {
+    num_timers += track->GetNumTimers();
+  }
+  return num_timers;
+}
+
+std::pair<uint64_t, uint64_t> TrackManager::GetTracksMinMaxTimestamps() const {
+  uint64_t min_time = std::numeric_limits<uint64_t>::max();
+  uint64_t max_time = std::numeric_limits<uint64_t>::min();
+
+  std::lock_guard<std::recursive_mutex> lock(mutex_);
+  for (auto& track : GetAllTracks()) {
+    if (!track->IsEmpty()) {
+      min_time = std::min(min_time, track->GetMinTime());
+      max_time = std::max(max_time, track->GetMaxTime());
+    }
+  }
+  return std::make_pair(min_time, max_time);
+}
--- a/src/OrbitGl/TrackManager.h
+++ b/src/OrbitGl/TrackManager.h
@@ -54,6 +54,9 @@ class TrackManager {
 
   [[nodiscard]] float GetTracksTotalHeight() const { return tracks_total_height_; }
 
+  [[nodiscard]] uint32_t GetNumTimers() const;
+  [[nodiscard]] std::pair<uint64_t, uint64_t> GetTracksMinMaxTimestamps() const;
+
   SchedulerTrack* GetOrCreateSchedulerTrack();
   ThreadTrack* GetOrCreateThreadTrack(int32_t tid);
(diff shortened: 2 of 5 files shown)


src/OrbitGl/TrackManager.cpp
```diff
@@ -403,3 +403,31 @@ FrameTrack* TrackManager::GetOrCreateFrameTrack(uint64_t function_id,
...
+  // tracks_ vector.
+  for (const auto& track : GetFrameTracks()) {
+    num_timers += track->GetNumTimers();
+  }
+  return num_timers;
+}
+
+std::pair<uint64_t, uint64_t> TrackManager::GetTracksMinMaxTime() const {
+  uint64_t min_time = std::numeric_limits<uint64_t>::max();
+  uint64_t max_time = std::numeric_limits<uint64_t>::lowest();
```
florian-kuebler, line 423: I'd rather use `min` for consistency, even though for `uint64_t` it does not make a difference.

florian-kuebler (approved):
LGTM. Thanks!

src/OrbitGl/TrackManager.cpp
```diff
@@ -403,3 +403,31 @@ FrameTrack* TrackManager::GetOrCreateFrameTrack(uint64_t function_id,
...
+  }
+  // Frame tracks are removable by users and cannot simply be thrown into the
+  // tracks_ vector.
+  for (const auto& track : GetFrameTracks()) {
+    num_timers += track->GetNumTimers();
+  }
+  return num_timers;
+}
+
+std::pair<uint64_t, uint64_t> TrackManager::GetTracksMinMaxTime() const {
```
dpallotti, line 421: What about calling this `GetTracksMinMaxTimestamps` for consistency with `UpdateCaptureMinMaxTimestamps` and in general because "timestamp" is more specific than "time"?

src/OrbitGl/TrackManager.h
```diff
@@ -53,6 +53,8 @@ class TrackManager {
                              PickingMode picking_mode);
 
   [[nodiscard]] float GetTracksTotalHeight() const { return tracks_total_height_; }
+  [[nodiscard]] uint32_t GetNumTimers() const;
```
dpallotti, line 56:
```suggestion
  [[nodiscard]] float GetTracksTotalHeight() const { return tracks_total_height_; }

  [[nodiscard]] uint32_t GetNumTimers() const;
```

dpallotti (approved):

src/OrbitGl/TrackManager.cpp
karupayun: Of course! Changed.

src/OrbitGl/TrackManager.cpp
ronaldfw: Consistency with what though? Use of `std::min` below? Or because we cause `std::numeric_limits<unit64_t>::min()` elsewhere? 

I learned the hard way that `std::numeric_limits<T>::min()` is mathematically not necessarily the minimum of the range of T, and one should always use lowest when one wants the mathematically smallest possible value.

Thus, `lowest` is actually more consistent with `std::min` and less likely to lead to mistakes.

florian-kuebler: We are using `min` in most other paces with similar semantics, but especially here for the member variables.

I am not strongly against using lowest here. But we should then do it everywhere else (and probably in a different PR).

dpallotti: > Use of std::min below? Or because we cause std::numeric_limits<unit64_t>::min() elsewhere?

I would say both and more, as I would argue that `min` is way more common than `lowest` in general. Using `lowest` here is making me ask what the reason for it is. But it is true that one needs to be aware of the catch with floating-point types. If we want to always prefer `lowest` we should do it consistently.

karupayun: Don't have an strong opinion here, I like lowest for consistency everywhere, but as we are not using float pointers and they pointed that, shouldn't be an issue. BTW, we can change it in a future PR if we agree on that.

ronaldfw: > We are using min in most other paces with similar semantics, but especially here for the member variables.

Semantically, I would say we are using `min` to mean mathematically smallest value of the type, which in C++ `numeric_limits` is called `lowest`.

I argue for using `lowest` as it's almost always the correct choice, unless there is a special reason where one needs to smallest positive values for floating points (which should be rare).

PR #2105, opened 2021-03-20
Add cross-platform GetFilePathFromFd

dimitry-:
To be used in TemporaryFile

Bug: http://b/180574275
Test: build, run OrbitBaseTests


dimitry-:
actually this is not needed.

PR #2106, opened 2021-03-21
Implement TemporaryFile and use it in place of std::tmpnam

dimitry-:
Bug: http://b/180574275

### DIFF
--- a/src/OrbitBase/File.cpp
+++ b/src/OrbitBase/File.cpp
@@ -19,19 +19,23 @@
 
 namespace orbit_base {
 
-ErrorMessageOr<unique_fd> OpenFileForReading(const std::filesystem::path& path) {
-#if defined(__linux)
-  constexpr int kOpenFlags = O_RDONLY | O_CLOEXEC;
-#elif defined(_WIN32)
-  constexpr int kOpenFlags = O_RDONLY | O_BINARY;
-#endif  // defined(__linux)
-  int fd = TEMP_FAILURE_RETRY(open(path.string().c_str(), kOpenFlags));
+static ErrorMessageOr<unique_fd> OpenFile(const std::filesystem::path& path, int flags, int mode) {
+  int fd = TEMP_FAILURE_RETRY(open(path.string().c_str(), flags, mode));
   if (fd == kInvalidFd) {
     return ErrorMessage(
         absl::StrFormat("Unable to open file \"%s\": %s", path.string(), SafeStrerror(errno)));
   }
 
   return unique_fd{fd};
+}
+
+ErrorMessageOr<unique_fd> OpenFileForReading(const std::filesystem::path& path) {
+#if defined(__linux)
+  constexpr int kOpenFlags = O_RDONLY | O_CLOEXEC;
+#elif defined(_WIN32)
+  constexpr int kOpenFlags = O_RDONLY | O_BINARY;
+#endif  // defined(__linux)
+  return OpenFile(path, kOpenFlags, 0);
 }
 
 ErrorMessageOr<unique_fd> OpenFileForWriting(const std::filesystem::path& path) {
@@ -42,14 +46,18 @@ ErrorMessageOr<unique_fd> OpenFileForWriting(const std::filesystem::path& path) 
   constexpr int kOpenFlags = O_WRONLY | O_CREAT | O_TRUNC | O_BINARY;
   constexpr int kOpenMode = _S_IREAD | _S_IWRITE;
 #endif  // defined(__linux)
-  int fd = TEMP_FAILURE_RETRY(open(path.string().c_str(), kOpenFlags, kOpenMode));
+  return OpenFile(path, kOpenFlags, kOpenMode);
+}
 
-  if (fd == kInvalidFd) {
-    return ErrorMessage(
-        absl::StrFormat("Unable to open file \"%s\": %s", path.string(), SafeStrerror(errno)));
-  }
-
-  return unique_fd{fd};
+ErrorMessageOr<unique_fd> OpenNewFileForReadWrite(const std::filesystem::path& path) {
+#if defined(__linux)
+  constexpr int kOpenFlags = O_RDWR | O_CREAT | O_EXCL | O_CLOEXEC;
+  constexpr int kOpenMode = 0600;
+#elif defined(_WIN32)
+  constexpr int kOpenFlags = O_RDWR | O_CREAT | O_EXCL | O_BINARY;
+  constexpr int kOpenMode = _S_IREAD | _S_IWRITE;
+#endif  // defined(__linux)
+  return OpenFile(path, kOpenFlags, kOpenMode);
 }
 
 ErrorMessageOr<void> WriteFully(const unique_fd& fd, std::string_view content) {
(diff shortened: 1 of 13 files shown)


src/OrbitBase/TemporaryFile.cpp
```diff
@@ -0,0 +1,58 @@
...
+#include "OrbitBase/TemporaryFile.h"
+
+#include <stdlib.h>
+
+namespace orbit_base {
+
+void TemporaryFile::Create() noexcept {
+  std::error_code error;
+  std::filesystem::path temporary_dir = std::filesystem::temp_directory_path(error);
+  std::string file_path = (temporary_dir / "orbit_XXXXXX").string();
```
danielfenner, line 14: The man page and the examples I saw all use capital X's. Since the tests work this doesn't seem to matter though.

src/OrbitBase/include/OrbitBase/TemporaryFile.h
```diff
@@ -0,0 +1,35 @@
...
+
+#ifndef ORBIT_BASE_TEMPORARY_FILE_H_
+#define ORBIT_BASE_TEMPORARY_FILE_H_
+
+#include "OrbitBase/File.h"
+
+namespace orbit_base {
+class TemporaryFile final {
+ public:
+  TemporaryFile() { Create(); }
```
danielfenner, line 13: Why not simply implement the ctor?

Also: The Create() method has a bunch of error conditions. It might make sense to add a method to query whether a TemporaryFile was successfully created. A static factory might be an alternative:

`static std::unique_ptr<TemporaryFile> Create() {...}`

src/OrbitBase/File.cpp
```diff
@@ -42,14 +46,18 @@ ErrorMessageOr<unique_fd> OpenFileForWriting(const std::filesystem::path& path)
...
-
-  if (fd == kInvalidFd) {
-    return ErrorMessage(
-        absl::StrFormat("Unable to open file \"%s\": %s", path.string(), SafeStrerror(errno)));
-  }
+  return OpenFile(path, kOpenFlags, kOpenMode);
+}
 
-  return unique_fd{fd};
+ErrorMessageOr<unique_fd> OpenNewFileForReadWrite(const std::filesystem::path& path) {
```
dpallotti, line 52: I feel like `OpenFileForWriting` should be renamed to `OpenNewFileForWriting` or this should just be `OpenFileForReadWrite`. The naming is a bit confusing at the moment.

src/OrbitBase/TemporaryFile.cpp
dpallotti, line 14: Move this after `if (error)`, it's not used before that.

dpallotti, line 14: So this is because we want to it start with "orbit_"? Is it necessary? Why not just use `std::tmpnam`?

src/OrbitBase/include/OrbitBase/TemporaryFile.h
```diff
@@ -0,0 +1,35 @@
...
+  ~TemporaryFile() { CloseAndRemove(); }
+
+  TemporaryFile(const TemporaryFile&) = delete;
+  TemporaryFile& operator=(const TemporaryFile&) = delete;
+  TemporaryFile(TemporaryFile&&) = delete;
+  TemporaryFile& operator=(TemporaryFile&&) = delete;
```
dpallotti, line 19: Why shouldn't this class be moveable?

src/OrbitBase/include/OrbitBase/TemporaryFile.h
```diff
@@ -0,0 +1,35 @@
...
+#ifndef ORBIT_BASE_TEMPORARY_FILE_H_
+#define ORBIT_BASE_TEMPORARY_FILE_H_
+
+#include "OrbitBase/File.h"
+
+namespace orbit_base {
+class TemporaryFile final {
+ public:
+  TemporaryFile() { Create(); }
+  ~TemporaryFile() { CloseAndRemove(); }
```
dpallotti, line 14: I feel like having both the destructor and a public `CloseAndRemove` is a bit confusing. Can't the class just be purely RAII? This makes for a mix of styles and a not-so-clean interface.

src/OrbitBase/TemporaryFileTest.cpp
```diff
@@ -46,4 +46,24 @@ TEST(TemporaryFile, Destruction) {
...
+TEST(TemporaryFile, CleanupAfterReopen) {
+  std::filesystem::path file_path_copy;
+  {
+    TemporaryFile tmp_file;
+    EXPECT_TRUE(tmp_file.fd().valid());
+    EXPECT_THAT(tmp_file.file_path().string(), HasSubstr("orbit_"));
+    EXPECT_THAT(tmp_file.file_path().string(), Not(HasSubstr("XXXXXX")));
+    file_path_copy = tmp_file.file_path();
+
+    // Check that the file removed even if we call RemoveAndClose and then recreate it.
```
dpallotti, line 58: "that the file is removed"

src/OrbitBase/TemporaryFileTest.cpp
```diff
@@ -46,4 +46,24 @@ TEST(TemporaryFile, Destruction) {
...
+    tmp_file.CloseAndRemove();
+    auto fd_or_error = OpenFileForWriting(file_path_copy);
+    ASSERT_TRUE(fd_or_error.has_value()) << fd_or_error.error().message();
```
dpallotti, line 61: This kind of supports my comment that having both `CloseAndRemove` and the destructor do the same thing in a RAII-like class is confusing. This is a bit weird.

src/OrbitBase/File.cpp
dimitry-: well, it does have O_EXCL therefore this is not necesseraly new file

src/OrbitBase/TemporaryFile.cpp
dimitry-: aren't we also using capital X's? Not sure I understand what you mean here.

dimitry-: tmpnam is not safe it generates a lot of warnings. The reason is that it does not actually create a file.

src/OrbitBase/TemporaryFileTest.cpp
dimitry-: The primary goal for this class is to be used in tests. The scope here is very much intentional.

src/OrbitBase/TemporaryFileTest.cpp
dimitry-: done.

src/OrbitBase/TemporaryFile.cpp
danielfenner: Wow, these look like small x's - I did not even bother to copy them over.  You are right, these are capital letters.

danielfenner (approved):
lgtm!

src/OrbitBase/TemporaryFile.cpp
dimitry-: done.

src/OrbitBase/include/OrbitBase/TemporaryFile.h
dpallotti: Defining the move constructor but deleting the move assignment is super weird, they should both be implemented.

PR #2107, opened 2021-03-22
[FOR RELEASE] Fix timer label always showing 1s for loaded captures and empty captures.

florian-kuebler:
Within #2031, we changed the main tick to only update the timer label
when capturing. For loaded captures, this update was never done.

Furthermore, the default value is 1s. For empty captures (not started
yet), this is actually wrong. So this changes the default also to
0s.

Test: Load a capture. Connect to a process. Check capture time label.
Bug: http://b/183203015

### DIFF
--- a/src/OrbitQt/FilterPanelWidget.ui
+++ b/src/OrbitQt/FilterPanelWidget.ui
@@ -79,7 +79,7 @@
    <item>
     <widget class="QLabel" name="timerLabel">
      <property name="text">
-      <string>1s</string>
+      <string>0s</string>
      </property>
     </widget>
    </item>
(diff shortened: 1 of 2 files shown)


florian-kuebler:
I am actually wondering if this is enough. [This change](https://github.com/google/orbit/pull/2031/files#diff-fc6728b247ca2e98948bc9b7413fabad7dd57246578557c0f663571bf133742bR790), could lead to the capture time being up to 16ms off. Imagine the following:
We are capturing and drawing the timer label.
~16ms later we are stopping the capture.
In the next main tick, the capture is already stopped, so we are not updating the timer label.
---> We are displaying a wrong result.

A proper fix could be to remember the last capture duration, and only update if we have a new value. Not sure, if computation-wise this makes any difference to always drawing the timer label. Maybe we should just revert this particular line?

karupayun (approved):
LGTM! Thanks for solving the issue :)

reichlfl (approved):

PR #2108, opened 2021-03-22
[release/1.61] Fix crash when calling ClearEdits() on invalid widget pointer

ronaldfw:
As Qt is handling the lifecycle of widgets such as the track filter
panel, we cannot directly call methods on widgets such as
FilterPanelWidget. We use Qt's signal/slot mechanism to avoid a direct
call for ClearEdits(), which fixes a crash.

Tested: Manually ran Orbit multiple times and took captures. Also tested
if the filter text panels are actually cleared on starting a capture.
Bug: http://b/183199646


vickyliu-go4it (approved):
LGTM! Thanks for that!

karupayun (approved):
LGTM! Is this the issue we were having when starting a capture? Thanks for solving this :)

ronaldfw:
> LGTM! Is this the issue we were having when starting a capture? Thanks for solving this :)

Yes, exactly!

ronaldfw:
Thanks for the reviews!

PR #2109, opened 2021-03-22
Allocate memory suitable for trampolines.

danielfenner:
Previously the allocation in the tracee happened at arbitrary memory
positions. For the trampolines we need chunks of memory in a 32 bit
offset range of the executable code segments of the trace.

This change contains:

a) Changes to facilitate allocation of memory at a specified position
(compare AllocateInTracee.h) and follow up changes to this.

b) The logic to determine the position of the memory chunk for the
trampolines (compare Trampoline.h) and the code doing the actual
allocation.

c) Fix for a glitch in FreeInTracee: Previously it was possible that
FreeInTracee used the memory segment it was meant to unmap as an
intermediate working area (which causes a crash).

### DIFF
--- a/src/UserSpaceInstrumentation/AllocateInTracee.cpp
+++ b/src/UserSpaceInstrumentation/AllocateInTracee.cpp
@@ -133,15 +133,16 @@ namespace {
 [[nodiscard]] ErrorMessageOr<uint64_t> AllocateInTracee(pid_t pid, uint64_t address,
                                                         uint64_t size) {
   // Syscall will be equivalent to:
-  // `mmap(address, size, PROT_READ | PROT_WRITE | PROT_EXEC, MAP_PRIVATE | MAP_ANONYMOUS | MAP_FIXED, -1, 0)`
-  // unless address it 0 in which case it will be (without the `MAP_FIXED`): 
+  // `mmap(address, size, PROT_READ | PROT_WRITE | PROT_EXEC, MAP_PRIVATE | MAP_ANONYMOUS |
+  // MAP_FIXED, -1, 0)`
+  // Unless address it 0 in which case it will be (without the `MAP_FIXED`):
   // `mmap(0, size, PROT_READ | PROT_WRITE | PROT_EXEC, MAP_PRIVATE | MAP_ANONYMOUS, -1, 0)`
   const int flags =
       (address == 0) ? (MAP_PRIVATE | MAP_ANONYMOUS) : (MAP_PRIVATE | MAP_ANONYMOUS | MAP_FIXED);
   constexpr uint64_t kSyscallNumberMmap = 9;
   auto result_or_error =
       SyscallInTracee(pid, kSyscallNumberMmap, address, size, PROT_READ | PROT_WRITE | PROT_EXEC,
-                      flags, static_cast<uint64_t>(-1), 0, /*exclude_address=*/ 0);
+                      flags, static_cast<uint64_t>(-1), 0, /*exclude_address=*/0);
   if (result_or_error.has_error()) {
     return ErrorMessage(absl::StrFormat("Failed to execute mmap syscall: \"%s\"",
                                         result_or_error.error().message()));
--- a/src/UserSpaceInstrumentation/AllocateInTracee.cpp
+++ b/src/UserSpaceInstrumentation/AllocateInTracee.cpp
@@ -144,8 +144,9 @@ namespace {
   }
   const uint64_t result = result_or_error.value();
   if (address != 0 && result != address) {
-    FAIL_IF(FreeInTracee(pid, result, size).has_error(),
-            "Unable to free proviously allocated memory.");
+    auto error_or_void = FreeInTracee(pid, result, size);
+    FAIL_IF(error_or_void.has_error(), "Unable to free proviously allocated memory: \"%s\"",
+            error_or_void.error().message());
     return ErrorMessage(
         absl::StrFormat("AllocateInTracee wanted to allocate memory at %#x but got memory at a "
                         "different adress: %#x. The memory has been freed again.",
--- a/src/UserSpaceInstrumentation/AllocateInTracee.cpp
+++ b/src/UserSpaceInstrumentation/AllocateInTracee.cpp
@@ -135,9 +135,9 @@ namespace {
   // Syscall will be equivalent to:
   // `mmap(address, size, PROT_READ | PROT_WRITE | PROT_EXEC, MAP_PRIVATE | MAP_ANONYMOUS, -1, 0)`
   constexpr uint64_t kSyscallNumberMmap = 9;
-  auto result_or_error =
-      SyscallInTracee(pid, kSyscallNumberMmap, address, size, PROT_READ | PROT_WRITE | PROT_EXEC,
-                      MAP_PRIVATE | MAP_ANONYMOUS, static_cast<uint64_t>(-1), 0, /*exclude_address=*/0);
+  auto result_or_error = SyscallInTracee(
+      pid, kSyscallNumberMmap, address, size, PROT_READ | PROT_WRITE | PROT_EXEC,
+      MAP_PRIVATE | MAP_ANONYMOUS, static_cast<uint64_t>(-1), 0, /*exclude_address=*/0);
   if (result_or_error.has_error()) {
     return ErrorMessage(absl::StrFormat("Failed to execute mmap syscall: \"%s\"",
                                         result_or_error.error().message()));
--- a/src/UserSpaceInstrumentation/AllocateInTracee.cpp
+++ b/src/UserSpaceInstrumentation/AllocateInTracee.cpp
@@ -30,10 +30,13 @@ namespace {
 // list: https://github.com/torvalds/linux/blob/master/arch/x86/entry/syscalls/syscall_64.tbl
 // The parameters and the ordering are the same as in the C wrapper:
 // https://man7.org/linux/man-pages/dir_section_2.html
+// Optionally one can specify `exclude_address`. This prevents the method from using an address
+// range containing `exclude_address` as a working area. This is required for the munmap syscall
+// which might otherwise choose the mapping it is removing as a working area.
 [[nodiscard]] ErrorMessageOr<uint64_t> SyscallInTracee(pid_t pid, uint64_t syscall, uint64_t arg_0,
-                                                       uint64_t arg_1 = 0, uint64_t arg_2 = 0,
-                                                       uint64_t arg_3 = 0, uint64_t arg_4 = 0,
-                                                       uint64_t arg_5 = 0) {
+                                                       uint64_t arg_1, uint64_t arg_2,
+                                                       uint64_t arg_3, uint64_t arg_4,
+                                                       uint64_t arg_5, uint64_t exclude_address) {
   RegisterState original_registers;
   auto result_backup = original_registers.BackupRegisters(pid);
   if (result_backup.has_error()) {
@@ -46,7 +49,7 @@ namespace {
   }
 
   // Get an executable memory region.
-  auto result_memory_region = GetFirstExecutableMemoryRegion(pid);
+  auto result_memory_region = GetFirstExecutableMemoryRegion(pid, exclude_address);
   if (result_memory_region.has_error()) {
     return ErrorMessage(absl::StrFormat("Failed to find executable memory region: \"%s\"",
                                         result_memory_region.error().message()));
@@ -127,13 +130,18 @@ namespace {
 
 }  // namespace
 
-[[nodiscard]] ErrorMessageOr<uint64_t> AllocateInTracee(pid_t pid, uint64_t size) {
+[[nodiscard]] ErrorMessageOr<uint64_t> AllocateInTracee(pid_t pid, uint64_t address,
+                                                        uint64_t size) {
   // Syscall will be equivalent to:
-  // `mmap(NULL, size, PROT_READ | PROT_WRITE | PROT_EXEC, MAP_PRIVATE | MAP_ANONYMOUS, -1, 0)`
+  // `mmap(address, size, PROT_READ | PROT_WRITE | PROT_EXEC, MAP_PRIVATE | MAP_ANONYMOUS | MAP_FIXED, -1, 0)`
+  // unless address it 0 in which case it will be (without the `MAP_FIXED`): 
+  // `mmap(0, size, PROT_READ | PROT_WRITE | PROT_EXEC, MAP_PRIVATE | MAP_ANONYMOUS, -1, 0)`
+  const int flags =
+      (address == 0) ? (MAP_PRIVATE | MAP_ANONYMOUS) : (MAP_PRIVATE | MAP_ANONYMOUS | MAP_FIXED);
   constexpr uint64_t kSyscallNumberMmap = 9;
-  auto result_or_error = SyscallInTracee(pid, kSyscallNumberMmap, static_cast<uint64_t>(NULL), size,
-                                         PROT_READ | PROT_WRITE | PROT_EXEC,
-                                         MAP_PRIVATE | MAP_ANONYMOUS, static_cast<uint64_t>(-1), 0);
+  auto result_or_error =
+      SyscallInTracee(pid, kSyscallNumberMmap, address, size, PROT_READ | PROT_WRITE | PROT_EXEC,
+                      flags, static_cast<uint64_t>(-1), 0, /*exclude_address=*/ 0);
   if (result_or_error.has_error()) {
     return ErrorMessage(absl::StrFormat("Failed to execute mmap syscall: \"%s\"",
                                         result_or_error.error().message()));
@@ -145,7 +153,8 @@ namespace {
   // Syscall will be equivalent to:
   // `munmap(address, size)`
   constexpr uint64_t kSyscallNumberMunmap = 11;
-  auto result_or_error = SyscallInTracee(pid, kSyscallNumberMunmap, address, size);
+  auto result_or_error =
+      SyscallInTracee(pid, kSyscallNumberMunmap, address, size, 0, 0, 0, 0, address);
   if (result_or_error.has_error()) {
     return ErrorMessage(absl::StrFormat("Failed to execute munmap syscall: \"%s\"",
                                         result_or_error.error().message()));
--- a/src/UserSpaceInstrumentation/AllocateInTracee.cpp
+++ b/src/UserSpaceInstrumentation/AllocateInTracee.cpp
@@ -133,21 +133,25 @@ namespace {
 [[nodiscard]] ErrorMessageOr<uint64_t> AllocateInTracee(pid_t pid, uint64_t address,
                                                         uint64_t size) {
   // Syscall will be equivalent to:
-  // `mmap(address, size, PROT_READ | PROT_WRITE | PROT_EXEC, MAP_PRIVATE | MAP_ANONYMOUS |
-  // MAP_FIXED, -1, 0)`
-  // Unless address it 0 in which case it will be (without the `MAP_FIXED`):
-  // `mmap(0, size, PROT_READ | PROT_WRITE | PROT_EXEC, MAP_PRIVATE | MAP_ANONYMOUS, -1, 0)`
-  const int flags =
-      (address == 0) ? (MAP_PRIVATE | MAP_ANONYMOUS) : (MAP_PRIVATE | MAP_ANONYMOUS | MAP_FIXED);
+  // `mmap(address, size, PROT_READ | PROT_WRITE | PROT_EXEC, MAP_PRIVATE | MAP_ANONYMOUS, -1, 0)`
   constexpr uint64_t kSyscallNumberMmap = 9;
   auto result_or_error =
       SyscallInTracee(pid, kSyscallNumberMmap, address, size, PROT_READ | PROT_WRITE | PROT_EXEC,
-                      flags, static_cast<uint64_t>(-1), 0, /*exclude_address=*/0);
+                      MAP_PRIVATE | MAP_ANONYMOUS, static_cast<uint64_t>(-1), 0, /*exclude_address=*/0);
   if (result_or_error.has_error()) {
     return ErrorMessage(absl::StrFormat("Failed to execute mmap syscall: \"%s\"",
                                         result_or_error.error().message()));
   }
-  return result_or_error.value();
+  const uint64_t result = result_or_error.value();
+  if (address != 0 && result != address) {
+    FAIL_IF(FreeInTracee(pid, result, size).has_error(),
+            "Unable to free proviously allocated memory.");
+    return ErrorMessage(
+        absl::StrFormat("AllocateInTracee wanted to allocate memory at %#x but got memory at a "
+                        "different adress: %#x. The memory has been freed again.",
+                        address, result));
+  }
+  return result;
 }
 
 [[nodiscard]] ErrorMessageOr<void> FreeInTracee(pid_t pid, uint64_t address, uint64_t size) {
--- a/src/UserSpaceInstrumentation/AllocateInTracee.h
+++ b/src/UserSpaceInstrumentation/AllocateInTracee.h
@@ -14,9 +14,13 @@
 namespace orbit_user_space_instrumentation {
 
 // Allocate `size` bytes of memory in the tracee's address space using mmap. The memory will have
-// read, write, and execute permissions.
-// Assumes we are already attached to the tracee `pid` e.g. using `AttachAndStopProcess`.
-[[nodiscard]] ErrorMessageOr<uint64_t> AllocateInTracee(pid_t pid, uint64_t size);
+// read, write, and execute permissions. The memory allocated will start at `address`. `address`
+// needs to be aligned to page boundaries. Make sure the requested memory range is available;
+// otherwise the overlapped part of existing mappings will be discarded.
+// If `address` is zero the placement of memory will be arbitray (compare the documenation of mmap:
+// https://man7.org/linux/man-pages/man2/mmap.2.html). Assumes we are already attached to the tracee
+// `pid` e.g. using `AttachAndStopProcess`.
+[[nodiscard]] ErrorMessageOr<uint64_t> AllocateInTracee(pid_t pid, uint64_t address, uint64_t size);
 
 // Free address range previously allocated with AllocateInTracee using munmap.
 // Assumes we are already attached to the tracee `pid` e.g. using `AttachAndStopProcess`.
--- a/src/UserSpaceInstrumentation/AllocateInTracee.h
+++ b/src/UserSpaceInstrumentation/AllocateInTracee.h
@@ -15,8 +15,8 @@ namespace orbit_user_space_instrumentation {
 
 // Allocate `size` bytes of memory in the tracee's address space using mmap. The memory will have
 // read, write, and execute permissions. The memory allocated will start at `address`. `address`
-// needs to be aligned to page boundaries. Make sure the requested memory range is available;
-// otherwise the overlapped part of existing mappings will be discarded.
+// needs to be aligned to page boundaries. If the memory mapping can not be placed at `address` an
+// error is returned.
 // If `address` is zero the placement of memory will be arbitray (compare the documenation of mmap:
 // https://man7.org/linux/man-pages/man2/mmap.2.html). Assumes we are already attached to the tracee
 // `pid` e.g. using `AttachAndStopProcess`.
--- a/src/UserSpaceInstrumentation/AllocateInTraceeTest.cpp
+++ b/src/UserSpaceInstrumentation/AllocateInTraceeTest.cpp
@@ -2,6 +2,7 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
 
+#include <absl/strings/numbers.h>
 #include <absl/strings/str_format.h>
 #include <absl/strings/str_split.h>
 #include <gtest/gtest.h>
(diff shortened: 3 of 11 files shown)


src/UserSpaceInstrumentation/AllocateInTracee.h
```diff
@@ -14,9 +14,11 @@
 namespace orbit_user_space_instrumentation {
 
 // Allocate `size` bytes of memory in the tracee's address space using mmap. The memory will have
-// read, write, and execute permissions.
+// read, write, and execute permissions. The memory allocated will start at `address`. If `address`
```
dimitry-, line 17: not exactly. since you are not using MAP_FIXED the address is a hint, kernel will try to allocate it at the specified address if possible and if not it will allocate nearby that address for some value of nearby. If the expected behaviour to allocate at exact address there should be additional check of return value of mmap, and some handling if this did not happen.

src/UserSpaceInstrumentation/AllocateInTraceeTest.cpp
```diff
@@ -65,14 +66,25 @@ TEST(AllocateInTraceeTest, AllocateAndFree) {
 
   // Allocate a megabyte in the tracee.
   constexpr uint64_t kMemorySize = 1024 * 1024;
-  auto result_allocate = AllocateInTracee(pid, kMemorySize);
-  CHECK(result_allocate.has_value());
-
+  auto result_allocate = AllocateInTracee(pid, 0, kMemorySize);
+  ASSERT_TRUE(result_allocate.has_value());
```
dimitry-, line 70: add a test verifying that AllocateInTracee works as expected when allocated twice at the same address?

src/UserSpaceInstrumentation/Trampoline.h
```diff
@@ -0,0 +1,55 @@
...
+#include <cstdint>
+#include <optional>
+#include <utility>
+#include <vector>
+
+#include "OrbitBase/Result.h"
+
+namespace orbit_user_space_instrumentation {
+
+using AddressRange = std::pair<uint64_t, uint64_t>;
```
dimitry-, line 19: nit: maybe introduce a trivial class to have start/end instead of first/second, makes code more readable

src/UserSpaceInstrumentation/Trampoline.cpp
```diff
@@ -0,0 +1,167 @@
...
+    if (DoAddressRangesOverlap(all_ranges[i], range)) {
+      return i;
+    }
+  }
+  return std::nullopt;
+}
+
+std::optional<size_t> HighestIntersectingAddressRange(const std::vector<AddressRange>& all_ranges,
+                                                      const AddressRange& range) {
+  size_t i = all_ranges.size();
```
dimitry-, line 86: what happens if size() == 0? add a test?

src/UserSpaceInstrumentation/Trampoline.cpp
```diff
@@ -0,0 +1,167 @@
...
+      return i;
+    }
+  } while (i > 0);
+  return std::nullopt;
+}
+
+ErrorMessageOr<AddressRange> FindAddressRangeForTrampoline(
+    const std::vector<AddressRange>& taken_ranges, const AddressRange& code_range, uint64_t size) {
+  uint64_t page_size = 0;
+  auto optional_pagesize = orbit_base::ExecuteCommand("getconf PAGESIZE");
```
dimitry-, line 99: sysconf(_SC_PAGE_SIZE)

src/UserSpaceInstrumentation/AllocateInTracee.h
```diff
@@ -14,9 +14,11 @@
...
+// is null the placement of memory will be arbitray (compare the documenation of mmap:
```
dpallotti, line 18: As `address` is a `uint64_t`, technically "if address is zero".

src/UserSpaceInstrumentation/TrampolineTest.cpp
```diff
@@ -0,0 +1,244 @@
...
+using orbit_base::ReadFileToString;
+
+extern "C" __attribute__((noinline)) int DoubleAndIncrease(int i) {
+  i = 2 * i;
+  return i + 1;
+}
+
+}  // namespace
+
+TEST(TrampolineTest, DoAddressRangesOverlap) {
```
dpallotti, line 37: Add tests with extremes touching, e.g., {1,3},{3,7} and {3,7},{7,9}

src/UserSpaceInstrumentation/TrampolineTest.cpp
```diff
@@ -0,0 +1,244 @@
...
+                                             {kOneGb + 0x01 * kOneMb - 1, kOneGb + 0x10 * kOneMb},
+                                             {kOneGb + 0x11 * kOneMb - 1, kOneGb + 0x20 * kOneMb},
+                                             {kOneGb + 0x21 * kOneMb - 1, kOneGb + 0x30 * kOneMb},
+                                             {kOneGb + 0x31 * kOneMb - 1, kOneGb + 0x40 * kOneMb}};
+  address_range_or_error = FindAddressRangeForTrampoline(kTakenRanges6, {kOneMb, kOneGb}, kOneMb);
+  ASSERT_FALSE(address_range_or_error.has_error()) << address_range_or_error.error().message();
+  EXPECT_EQ(kOneGb + 0x40 * kOneMb, address_range_or_error.value().first);
+
+  // No space to the left and the last segment nearly fills up the 64 bit address space. So no
+  // placement is possibel.
```
dpallotti, line 154:
```suggestion
  // placement is possible.
```

src/UserSpaceInstrumentation/TrampolineTest.cpp
```diff
@@ -0,0 +1,244 @@
...
+}
+
+TEST(TrampolineTest, FindAddressRangeForTrampoline) {
+  constexpr uint64_t k64Kb = 0x10000;
+  constexpr uint64_t kOneMb = 0x100000;
+  constexpr uint64_t k256Mb = 0x10000000;
+  constexpr uint64_t kOneGb = 0x40000000;
+
+  // Trivial placement to the left.
+  std::vector<AddressRange> kTakenRanges1 = {
```
dpallotti, line 100: Maybe make these `kTakenRanges#` `const` as they start with `k`.

src/UserSpaceInstrumentation/TrampolineTest.cpp
```diff
@@ -0,0 +1,244 @@
...
+  address_range_or_error = FindAddressRangeForTrampoline(
+      kTakenRanges8, {0x12 * kOneGb + kOneMb - 1, 0x12 * kOneGb + 2 * kOneMb}, kOneMb);
+  ASSERT_FALSE(address_range_or_error.has_error()) << address_range_or_error.error().message();
+  constexpr uint64_t kPageSize = 4096;
+  constexpr uint64_t kNextPage =
+      (((0x12 * kOneGb + 2 * kOneMb + 42) + (kPageSize - 1)) / kPageSize) * kPageSize;
+  EXPECT_EQ(kNextPage, address_range_or_error.value().first);
+
+  // There is no sufficiently large gap in the mappings in the 2GB below the code segment. And there
+  // also is no gap large enough in the 2GB above the code segment. So no placement is possibel.
```
dpallotti, line 183:
```suggestion
  // also is no gap large enough in the 2GB above the code segment. So no placement is possible.
```

src/UserSpaceInstrumentation/AllocateInTracee.h
danielfenner: Great, thanks, I did not know about MAP_FIXED which does exactly what we need. I planned to handle possible misplacement "above" this code (in which case the comment would have been misleading). 
With MAP_FIXED I can adjust the code to the comment.

src/UserSpaceInstrumentation/Trampoline.h
danielfenner: done

src/UserSpaceInstrumentation/Trampoline.cpp
danielfenner: done

src/UserSpaceInstrumentation/Trampoline.cpp
danielfenner: done

src/UserSpaceInstrumentation/AllocateInTracee.cpp
```diff
@@ -127,13 +131,14 @@ namespace {
...
   // Syscall will be equivalent to:
-  // `mmap(NULL, size, PROT_READ | PROT_WRITE | PROT_EXEC, MAP_PRIVATE | MAP_ANONYMOUS, -1, 0)`
+  // `mmap(address, size, PROT_READ | PROT_WRITE | PROT_EXEC, MAP_PRIVATE | MAP_ANONYMOUS, -1, 0)`
   constexpr uint64_t kSyscallNumberMmap = 9;
-  auto result_or_error = SyscallInTracee(pid, kSyscallNumberMmap, static_cast<uint64_t>(NULL), size,
-                                         PROT_READ | PROT_WRITE | PROT_EXEC,
-                                         MAP_PRIVATE | MAP_ANONYMOUS, static_cast<uint64_t>(-1), 0);
+  auto result_or_error =
+      SyscallInTracee(pid, kSyscallNumberMmap, address, size, PROT_READ | PROT_WRITE | PROT_EXEC,
+                      MAP_PRIVATE | MAP_ANONYMOUS, static_cast<uint64_t>(-1), 0, 0);
```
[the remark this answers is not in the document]
danielfenner: Now that you mention that: The `exclude_address` was an optional parameter. I could have omitted it. But with the already length signature I prefer to make it non optional. Added the comment as well.

src/UserSpaceInstrumentation/Trampoline.cpp
```diff
@@ -0,0 +1,167 @@
...
+namespace internal {
+
+namespace {
+
+using absl::numbers_internal::safe_strtou64_base;
+using orbit_base::ReadFileToString;
+
+// Parses the /proc/pid/maps file of a process and returns all the taken address ranges (joining
+// directly neighboring ones). We also add a range [0, /proc/sys/vm/mmap_min_addr] to block the
+// lowest addresses in the process space which mmap can not use.
```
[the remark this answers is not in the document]
danielfenner: done

src/UserSpaceInstrumentation/Trampoline.cpp
```diff
@@ -0,0 +1,167 @@
...
+    if (addresses.size() != 2) {
+      continue;
+    }
+    uint64_t address_begin = 0;
+    uint64_t address_end = 0;
+    if (!safe_strtou64_base(addresses[0], &address_begin, 16) ||
+        !safe_strtou64_base(addresses[1], &address_end, 16)) {
+      continue;
+    }
+    // Join with previous segment ...
```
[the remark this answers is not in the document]
danielfenner: done

src/UserSpaceInstrumentation/Trampoline.cpp
```diff
@@ -0,0 +1,167 @@
...
+      return i;
+    }
+  }
+  return std::nullopt;
+}
+
+std::optional<size_t> HighestIntersectingAddressRange(const std::vector<AddressRange>& all_ranges,
+                                                      const AddressRange& range) {
+  size_t i = all_ranges.size();
+  do {
```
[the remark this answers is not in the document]
danielfenner: That is the endless loop I had here before.

src/UserSpaceInstrumentation/AllocateInTracee.h
danielfenner: done

src/UserSpaceInstrumentation/TrampolineTest.cpp
danielfenner: done

src/UserSpaceInstrumentation/TrampolineTest.cpp
danielfenner: done

src/UserSpaceInstrumentation/TrampolineTest.cpp
danielfenner: done

src/UserSpaceInstrumentation/TrampolineTest.cpp
danielfenner: done

src/UserSpaceInstrumentation/AllocateInTracee.h
dimitry-: well, there is another thing you should probably know before doing that - MAP_FIXED will create the map even if there was already another map at this address. It will "override" it, and I am not sure this is safe.

src/UserSpaceInstrumentation/Trampoline.cpp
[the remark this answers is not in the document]
dimitry-: ssize_t

src/UserSpaceInstrumentation/Trampoline.cpp
```diff
@@ -0,0 +1,167 @@
...
+  uint64_t page_size = 0;
+  auto optional_pagesize = orbit_base::ExecuteCommand("getconf PAGESIZE");
+  if (!optional_pagesize.has_value() || !absl::SimpleAtoi(optional_pagesize.value(), &page_size)) {
+    return ErrorMessage("Unable to determine system's page size.");
+  }
+
+  // Try to fit an interval of length `size` below `code_range`.
+  auto optional_range_index = LowestIntersectingAddressRange(taken_ranges, code_range);
+  if (!optional_range_index) {
+    return ErrorMessage("code_range is not in taken_ranges.");
```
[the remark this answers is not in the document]
danielfenner: done

src/UserSpaceInstrumentation/Trampoline.cpp
```diff
@@ -0,0 +1,167 @@
...
+  }
+  while (optional_range_index.value() > 0) {
+    // Place directly to the left of the take interval we are in...
+    uint64_t address_trampoline = taken_ranges[optional_range_index.value()].first - size;
+    // ... but round down to the next page boundary.
```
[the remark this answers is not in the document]
danielfenner: done

src/UserSpaceInstrumentation/Trampoline.cpp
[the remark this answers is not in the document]
danielfenner: ..., and this is effectively an int. The style guide advises to "Use plain old int for such things." so I'll go with that.
Still looks better than the awkward "do ... while" thing.

src/UserSpaceInstrumentation/Trampoline.cpp
```diff
@@ -0,0 +1,167 @@
...
+    address_trampoline = (address_trampoline / page_size) * page_size;
+    AddressRange range_trampoline = {address_trampoline, address_trampoline + size};
+    optional_range_index = LowestIntersectingAddressRange(taken_ranges, range_trampoline);
+    if (!optional_range_index) {
+      // We do not intersect any taken interval. Check if we are close enough to code_range:
+      // code_range is above range_trampoline we will need to jmp back and forth in these ranges
+      // with 32 bit offsets. If no distance is greater than 0x7fffffff this is safe.
+      if (code_range.second - range_trampoline.first <= 0x7fffffff) {
```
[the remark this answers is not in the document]
danielfenner: done

src/UserSpaceInstrumentation/Trampoline.cpp
```diff
@@ -0,0 +1,167 @@
...
+        return range_trampoline;
+      }
+      // If we are already beyond the close range there is no need to go any further.
+      break;
+    }
+  }
+  // Try to fit an interval of length `size` above `code_range`.
```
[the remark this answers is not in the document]
danielfenner: done

src/UserSpaceInstrumentation/Trampoline.cpp
```diff
@@ -0,0 +1,167 @@
...
+  optional_range_index = HighestIntersectingAddressRange(taken_ranges, code_range);
+  if (!optional_range_index) {
+    return ErrorMessage("code_range is not in taken_ranges.");
+  }
+  do {
+    const uint64_t address_trampoline =
+        ((taken_ranges[optional_range_index.value()].second + (page_size - 1)) / page_size) *
+        page_size;
+    // Check if we run out of address space.
```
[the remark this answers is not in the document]
danielfenner: done

src/UserSpaceInstrumentation/Trampoline.cpp
```diff
@@ -0,0 +1,167 @@
...
+      // with 32 bit offsets. If no distance is greater than 0x7fffffff this is safe.
+      if (range_trampoline.second - code_range.first <= 0x7fffffff) {
+        return range_trampoline;
+      }
+      // If we are already beyond the close range there is no need to go any further.
+      break;
+    }
+  } while (optional_range_index.value() < taken_ranges.size());
+
+  return ErrorMessage("No place to fit size bytes close to code range.");
```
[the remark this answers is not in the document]
danielfenner: done

src/UserSpaceInstrumentation/Trampoline.h
```diff
@@ -0,0 +1,55 @@
...
+
+// Functions in the internal namespace are merely exposed in the header to facilitate testing. They
+// should not be used directly.
+namespace internal {
```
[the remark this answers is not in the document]
danielfenner: Yep, good point.

src/UserSpaceInstrumentation/AllocateInTraceeTest.cpp
danielfenner: While looking into this I recognized that it makes sense to move all the stuff that handeles AddressRange's into a separate header. There is already another definition of `AddressRange` in AccessTraceesMemory.h. I need to get rid of that. This will also make testing this easier.
I don't want to squeeze these changes into this PR. So let me do this in a followup.

danielfenner:
+pierric 
Mainly FYI since we talked about this already - this facilitates allocating trampolines close to code segments.

pierricgimmig (commented):
Hi Daniel, looks good to me overall, but please see minor comments. Thanks!

src/UserSpaceInstrumentation/TrampolineTest.cpp
```diff
@@ -0,0 +1,249 @@
...
+#include "Trampoline.h"
+
+namespace orbit_user_space_instrumentation {
+
+namespace {
+
+using absl::numbers_internal::safe_strtou64_base;
+using orbit_base::ReadFileToString;
+
+extern "C" __attribute__((noinline)) int DoubleAndIncrease(int i) {
```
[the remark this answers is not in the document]
danielfenner: done

src/UserSpaceInstrumentation/Trampoline.cpp
```diff
@@ -0,0 +1,155 @@
...
+
+std::optional<size_t> HighestIntersectingAddressRange(const std::vector<AddressRange>& all_ranges,
+                                                      const AddressRange& range) {
+  for (int i = all_ranges.size() - 1; i >= 0; i--) {
+    if (DoAddressRangesOverlap(all_ranges[i], range)) {
+      return i;
+    }
+  }
+  return std::nullopt;
+}
```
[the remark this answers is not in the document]
danielfenner: Yes, these ranges need to be sorted. And we only ever hand over sorted ranges since this is the way get them from GetTakenAddressRanges. 
So I'd rather make this explicit in the naming and documentation than adding a sort that would be an no-op.

src/UserSpaceInstrumentation/Trampoline.cpp
```diff
@@ -0,0 +1,155 @@
...
+      if (code_range.end - range_trampoline.start <= kMax32BitOffset) {
+        return range_trampoline;
+      }
+      // If we are already beyond the close range there is no need to go any further.
+      break;
+    }
+  }
+
+  // Try to fit an interval of length `size` above `code_range`.
+  optional_range_index = HighestIntersectingAddressRange(taken_ranges, code_range);
```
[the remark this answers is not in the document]
danielfenner: If `code_range` is intersecting `taken_ranges` we can only fit `code_range` above the highest of these `taken_ranges`. So we are looking for the lowest position about the highest one we are intersecting.

An easier way of looking at it: We are sliding `code_range` to the upwards until we don't intersect anything that is taken.

src/UserSpaceInstrumentation/Trampoline.cpp
```diff
@@ -0,0 +1,155 @@
...
+  }
+  return std::nullopt;
+}
+
+ErrorMessageOr<std::vector<AddressRange>> GetTakenAddressRanges(pid_t pid) {
+  std::vector<AddressRange> result;
+  OUTCOME_TRY(mmap_min_addr, ReadFileToString("/proc/sys/vm/mmap_min_addr"));
+  uint64_t mmap_min_addr_as_uint64 = 0;
+  if (!absl::SimpleAtoi(mmap_min_addr, &mmap_min_addr_as_uint64)) {
+    return ErrorMessage("Failed to read/proc/sys/vm/mmap_min_addr");
```
dimitry-, line 54: failed to parse probably.

src/UserSpaceInstrumentation/Trampoline.cpp
```diff
@@ -0,0 +1,155 @@
...
+    const std::vector<AddressRange>& ranges_sorted, const AddressRange& range) {
+  for (int i = ranges_sorted.size() - 1; i >= 0; i--) {
+    if (DoAddressRangesOverlap(ranges_sorted[i], range)) {
+      return i;
+    }
+  }
+  return std::nullopt;
+}
+
+ErrorMessageOr<std::vector<AddressRange>> GetTakenAddressRanges(pid_t pid) {
```
dimitry-, line 49: GetAllocated maybe?

src/UserSpaceInstrumentation/Trampoline.cpp
```diff
@@ -0,0 +1,155 @@
...
+  // Try to fit an interval of length `size` below `code_range`.
+  auto optional_range_index = LowestIntersectingAddressRange(taken_ranges, code_range);
+  if (!optional_range_index) {
+    return ErrorMessage(absl::StrFormat("code_range %#x-%#x is not in taken_ranges.",
+                                        code_range.start, code_range.end));
+  }
+  while (optional_range_index.value() > 0) {
+    // Place directly to the left of the take interval we are in...
+    uint64_t address_trampoline = taken_ranges[optional_range_index.value()].start - size;
+    // ... but round down to the previous page boundary.
```
dimitry-, line 101: nit: just "round down to the page boundary" "previous" is a bit confusing there.

src/UserSpaceInstrumentation/Trampoline.cpp
```diff
@@ -0,0 +1,155 @@
...
+    address_trampoline = (address_trampoline / page_size) * page_size;
```
dimitry-, line 102: completely optional: page_size is a power of 2 so you can replace it with something like address_trampoline & ~(page_size - 1)

src/UserSpaceInstrumentation/Trampoline.cpp
```diff
@@ -0,0 +1,155 @@
...
+  if (!optional_range_index) {
+    return ErrorMessage(absl::StrFormat("code_range %#x-%#x is not in taken_ranges.",
+                                        code_range.start, code_range.end));
+  }
+  do {
+    const uint64_t address_trampoline =
+        ((taken_ranges[optional_range_index.value()].end + (page_size - 1)) / page_size) *
+        page_size;
+    // Check if we ran out of address space.
+    if (address_trampoline >= 0xffffffffffffffff - size) {
```
dimitry-, line 128: is this UINTPTR_MAX?

src/UserSpaceInstrumentation/Trampoline.cpp
```diff
@@ -0,0 +1,155 @@
...
+      // We do not intersect any taken interval. Check if we are close enough to code_range:
+      // code_range is below range_trampoline we will need to jump back and forth in these ranges
+      // with 32 bit offsets. If no distance is greater than 0x7fffffff this is safe.
+      if (range_trampoline.end - code_range.start <= kMax32BitOffset) {
+        return range_trampoline;
+      }
+      // If we are already beyond the close range there is no need to go any further.
+      break;
+    }
+  } while (optional_range_index.value() < taken_ranges.size());
```
dimitry-, line 143: isn't this always true?

src/UserSpaceInstrumentation/Trampoline.cpp
```diff
@@ -0,0 +1,155 @@
...
+  constexpr uint64_t kMax32BitOffset = 0x7fffffff;
+  const uint64_t page_size = sysconf(_SC_PAGE_SIZE);
+
+  // Try to fit an interval of length `size` below `code_range`.
+  auto optional_range_index = LowestIntersectingAddressRange(taken_ranges, code_range);
+  if (!optional_range_index) {
+    return ErrorMessage(absl::StrFormat("code_range %#x-%#x is not in taken_ranges.",
+                                        code_range.start, code_range.end));
+  }
+  while (optional_range_index.value() > 0) {
```
dimitry-, line 98: this implies we know here that the very first segment is always starts with 0 address. Maybe add a comment about that?

src/UserSpaceInstrumentation/TrampolineTest.cpp
```diff
@@ -0,0 +1,249 @@
...
+
+TEST(TrampolineTest, FindAddressRangeForTrampoline) {
+  constexpr uint64_t k64Kb = 0x10000;
+  constexpr uint64_t kOneMb = 0x100000;
+  constexpr uint64_t k256Mb = 0x10000000;
+  constexpr uint64_t kOneGb = 0x40000000;
+
+  // Trivial placement to the left.
+  const std::vector<AddressRange> kTakenRanges1 = {
+      {0, k64Kb}, {kOneGb, 2 * kOneGb}, {3 * kOneGb, 4 * kOneGb}};
```
dimitry-, line 107: Maybe add a test for ranges which do not have the very first one start with 0 address?

src/UserSpaceInstrumentation/Trampoline.cpp
dimitry-: and probably a check since things could go wrong if someone passes list of addresses without 0-min_vaddr

src/UserSpaceInstrumentation/AllocateInTracee.cpp
```diff
@@ -133,21 +133,25 @@ namespace {
...
+                      MAP_PRIVATE | MAP_ANONYMOUS, static_cast<uint64_t>(-1), 0, /*exclude_address=*/0);
   if (result_or_error.has_error()) {
     return ErrorMessage(absl::StrFormat("Failed to execute mmap syscall: \"%s\"",
                                         result_or_error.error().message()));
   }
-  return result_or_error.value();
+  const uint64_t result = result_or_error.value();
+  if (address != 0 && result != address) {
+    FAIL_IF(FreeInTracee(pid, result, size).has_error(),
+            "Unable to free proviously allocated memory.");
```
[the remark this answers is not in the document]
danielfenner: done.

src/UserSpaceInstrumentation/Trampoline.cpp
danielfenner: done

src/UserSpaceInstrumentation/Trampoline.cpp
danielfenner: The range from zero to mmap_min_addr is not really allocated but just made unavailable via mmap for security reasons. The other ones are allocated (many of them but the loader I suppose). I agree that "Taken" is a bit hard to get. 
Changed to "GetUnavailableAddressRanges", agree?

src/UserSpaceInstrumentation/Trampoline.cpp
danielfenner: done

src/UserSpaceInstrumentation/Trampoline.cpp
danielfenner: Hmm, true, but I'll keep as is since it is mildly more readable imo.

src/UserSpaceInstrumentation/Trampoline.cpp
danielfenner: Indeed. This was left over from a prvious version. Changed to "true".

src/UserSpaceInstrumentation/Trampoline.cpp
danielfenner: Probably yes, but that is a coincidence. It is the UINTPTR_MAX of the other process. Currently both are always 64 bit. I hesitate a bit to put UINTPTR_MAX  here since it is misleading in this way. 

I dreamed up my own name "kMax64BitAdress". Not sure if this is better.

src/UserSpaceInstrumentation/Trampoline.cpp
danielfenner: Hmm, the method is *really* specific to the output of GetUnavailableAddressRanges. Strictly speaking I could also add a check for non overlapping regions. The only reason they are not tied together (either calling GetUnavailableAddressRanges from within FindAddressRangeForTrampoline or at least putting GetUnavailableAddressRanges  into an unnnamed namespace) is that I'd like to have them available for testing.

Added a check and comment. done.

src/UserSpaceInstrumentation/TrampolineTest.cpp
danielfenner: Added EXPECT_DEATH test.

src/UserSpaceInstrumentation/Trampoline.cpp
```diff
@@ -85,23 +85,29 @@ ErrorMessageOr<std::vector<AddressRange>> GetTakenAddressRanges(pid_t pid) {
 }
 
 ErrorMessageOr<AddressRange> FindAddressRangeForTrampoline(
-    const std::vector<AddressRange>& taken_ranges, const AddressRange& code_range, uint64_t size) {
+    const std::vector<AddressRange>& unavailable_ranges, const AddressRange& code_range,
+    uint64_t size) {
   constexpr uint64_t kMax32BitOffset = 0x7fffffff;
+  constexpr uint64_t kMax64BitAdress = 0xffffffffffffffff;
```
dimitry-, line 91: is it UINT64_MAX? :) It is very hard to count number of bytes this way, maybe separate every 4 bytes/digits with `'`?

src/UserSpaceInstrumentation/Trampoline.cpp
```diff
@@ -85,23 +85,29 @@ ErrorMessageOr<std::vector<AddressRange>> GetTakenAddressRanges(pid_t pid) {
 }
 
 ErrorMessageOr<AddressRange> FindAddressRangeForTrampoline(
-    const std::vector<AddressRange>& taken_ranges, const AddressRange& code_range, uint64_t size) {
+    const std::vector<AddressRange>& unavailable_ranges, const AddressRange& code_range,
+    uint64_t size) {
   constexpr uint64_t kMax32BitOffset = 0x7fffffff;
```
dimitry-, line 90: I am guessing this is INT32_MAX?

dimitry- (approved):
otherwise lgtm

src/UserSpaceInstrumentation/Trampoline.cpp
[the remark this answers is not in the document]
pierricgimmig: Ah I see, thanks for the precision.

pierricgimmig (approved):

src/UserSpaceInstrumentation/Trampoline.cpp

src/UserSpaceInstrumentation/Trampoline.cpp

PR #2110, opened 2021-03-22
[rebase on main] Fix crash when calling ClearEdits() on invalid widget pointer

ronaldfw:
As Qt is handling the lifecycle of widgets such as the track filter
panel, we cannot directly call methods on widgets such as
FilterPanelWidget. We use Qt's signal/slot mechanism to avoid a direct
call for ClearEdits(), which fixes a crash.

Tested: Manually ran Orbit multiple times and took captures. Also tested
if the filter text panels are actually cleared on starting a capture.
Bug: http://b/183199646


karupayun (approved):
LGTM!

PR #2111, opened 2021-03-22
[rebase on main] Fix a crash caused by wrongly used mutex

karupayun:
We used to have an only mutex in TimeGraph and we splitted into two when
we create TrackManager. In this PR, we moved the problematic functions
(GetNumTimers and GetTracksMinMaxTimestamps) to TrackManager to mutex
them properly.

Solve http://b/179467994.

Rebasing https://github.com/google/orbit/pull/2102 from the release branch.


florian-kuebler:
Could you link the original PR in the description?

florian-kuebler (approved):
LGTM, thanks!

karupayun:
> Could you link the original PR in the description?

Done!

PR #2112, opened 2021-03-22
Remove unused code from test library.

danielfenner:
This caused headaches when removing std::tmpnam from the code. Since the
functionality is not used at all let's get rid of it and solve this in
case it is really needed.


dimitry- (approved):
thanks!

PR #2115, opened 2021-03-22
[REBASE] Fix timer label always showing 1s for loaded captures and empty captures.

florian-kuebler:
Within #2031, we changed the main tick to only update the timer label
when capturing. For loaded captures, this update was never done.

Furthermore, the default value is 1s. For empty captures (not started
yet), this is actually wrong. So this changes the default also to
0s.

Original PR: #2107

Test: Load a capture. Connect to a process. Check capture time label.
Bug: http://b/183203015


karupayun (approved):
LGTM!

PR #2118, opened 2021-03-22
Move ORBIT_CRASH_HANDLING macro definition

beckerhe:
When moving `OrbitQt/main.cpp` to `Orbit/main.cpp` I missed also moving
the macro definition of `ORBIT_CRASH_HANDLING` which lead to crash
handling never be enabled.


karupayun (approved):
LGTM! Thanks for solving this so quickly!

antonrohr (approved):

PR #2120, opened 2021-03-22
Fix process loading stuck

antonrohr:
This fixes http://b/182996343 by starting `state_machine_` before registering the `OnProcessListUpdate` callback with `process_manager_` (Only the "Fix ProfilingTargetDialog being stuck" is relevant. The other commits are just refactoring/renaming). 

I am pretty sure this solved the problem, but it would be appreciated if someone could test this PR (@karupayun, the problem seems to occur often on your machine, maybe you can test it?)

This is a PR for main and I dont think it needs to be on release, since I dont think this is a regression, but I am happy to be convinced otherwise

### DIFF
--- a/src/OrbitQt/ProfilingTargetDialog.cpp
+++ b/src/OrbitQt/ProfilingTargetDialog.cpp
@@ -450,7 +450,7 @@ void ProfilingTargetDialog::SelectFile() {
   }
 }
 
-bool ProfilingTargetDialog::TrySelectProcess(const std::string& process_name) {
+bool ProfilingTargetDialog::TrySelectProcessByName(const std::string& process_name) {
   QModelIndexList matches = process_proxy_model_.match(
       process_proxy_model_.index(0, static_cast<int>(ProcessItemModel::Column::kName)),
       Qt::DisplayRole, QVariant::fromValue(QString::fromStdString(process_name)));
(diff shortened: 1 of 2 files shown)


beckerhe:
Haven't looked at it yet. But I wanted to say I agree. This doesn't need to be in the release.

src/OrbitQt/ProfilingTargetDialog.cpp
```diff
@@ -144,7 +144,8 @@ ProfilingTargetDialog::ProfilingTargetDialog(
   if (target_configuration_opt.has_value()) {
     TargetConfiguration config = std::move(target_configuration_opt.value());
     target_configuration_opt = std::nullopt;
-    SetStateMachineInitialStateFromTarget(std::move(config));
+    std::visit([this](auto target) { SetTargetAndStateMachineInitialState(std::move(target)); },
```
beckerhe, line 147: Nit: `auto&& target` - Will save a move. But is also super-nit!

beckerhe (approved):
LGTM

Thanks a lot for splitting this up so fine-grained. Made the review a breeze!

I can also test this later or tomorrow morning.

beckerhe:
Trying this now.

src/OrbitQt/ProfilingTargetDialog.cpp
antonrohr: I added `&&` and removed one `std::move`.

beckerhe:
> Trying this now.

Seems to work. I can't reproduce the problem anymore! Thanks!

karupayun (approved):
LGTM! Thanks for solving this!

Before these 4 commits, I made 5 tries before reproduced it (20%), after I tried 12 times and couldn't reproduce it.

No comments in the code.

antonrohr:
Okay, this sounds good!
There is another issue with switching between local and stadia that @reichlfl found (see http://b/182996343). I will address this in different PR.

PR #2121, opened 2021-03-23
Add OrbitTestLegacyApi to test original Orbit.h support

pierricgimmig:
Since OrbitTest now uses the new manual instrumentation Api, we need a new
test executable to test the old version of Orbit.h. This is basically the same code
as OrbitTest.cpp.

### DIFF
--- a/src/OrbitTest/main_legacy_api.cpp
+++ b/src/OrbitTest/main_legacy_api.cpp
@@ -1,0 +1,132 @@
+// Copyright (c) 2020 The Orbit Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#include <stdint.h>
+
+#include <algorithm>
+#include <chrono>
+#include <cmath>
+#include <cstdint>
+#include <iostream>
+#include <mutex>
+#include <string>
+#include <thread>
+#include <vector>
+
+#include "../Orbit.h"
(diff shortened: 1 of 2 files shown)


src/OrbitTest/main_legacy_api.cpp
```diff
@@ -0,0 +1,132 @@
...
+  void Stop() {
+    std::lock_guard lock(mutex_);
+    for (uint64_t id : async_scope_ids_to_stop_) {
+      std::string str = absl::StrFormat(
+          "This is a very long dynamic string: The quick brown fox jumps over the lazy dog. This "
+          "string is associated with task id %u.",
+          id);
+      ORBIT_ASYNC_STRING(str.c_str(), id);
+      ORBIT_STOP_ASYNC(id);
+      std::this_thread::sleep_for(std::chrono::milliseconds(2));
```
ronaldfw, line 39: Nit: I would add a comment to indicate what the purpose of the sleep here is. Is this for throttling the amount of events caused by this code here?

ronaldfw (approved):
LGTM, thanks

karupayun (approved):
LGTM!!

src/OrbitTest/main_legacy_api.cpp
pierricgimmig: Yep, good, point, added a comment, thanks.

PR #2122, opened 2021-03-23
Enable running Python unit tests for Jupyter notebooks in our CI

ronaldfw:


ronaldfw:
Based on suggestion from beckerhe@ here:
https://github.com/google/orbit/pull/1938#issuecomment-788698605

beckerhe:
Nice. But something seems to be not right yet. The python tests are executed, and they fail, but the error seems not to be propagated.

ronaldfw:
> Nice. But something seems to be not right yet. The python tests are executed, and they fail, but the error seems not to be propagated.

There was a `return` and `sys.exit(main())` missing in one of the files. I fixed this, let's see if this causes the CI to fail now. Then I'll take a look how to fix the actual test error (missing package).

ronaldfw:
Ok, looks like now the failed test propagates its result properly.

contrib/jupyter/build.py
```diff
@@ -7,6 +7,9 @@
 import os, sys
 import subprocess
 
+def install_external_packages():
+    subprocess.check_call([sys.executable, "-m", "pip", "install", "protobuf"])
```
beckerhe, line 11: I'm not sure if that's the best way to do it. `protobuf` is a dependency of the `orbitutils` package, right?

If yes, I think it should go into the `install_requires` property of `setup.py`.

ronaldfw: Ah, yes, that sounds better!

contrib/jupyter/setup.py
```diff
@@ -14,5 +14,5 @@
     url='http://github.com/google/orbit',
     license='BSD 2-Clause License',
     description='Orbit capture utilities',
-    install_requires=[],
+    install_requires=["protobuf"],
```
beckerhe, line 17: Sorry for nitpicking: It might make sense to stick to a fixed, maybe the one we use on the C++ side.

This would be `install_requires=["protobuf==3.9.1"],`. The latest version would be  `install_requires=["protobuf==3.15.6"],`

ronaldfw: No worries, I'm glad you are pointing this out. Will fix!

beckerhe: Thanks. If the CI agrees I think this will be good to go!

beckerhe (approved):

ronaldfw:
Thanks for the review. If the CI succeeds, I'll submit the change.

PR #2125, opened 2021-03-23
Update CaptureWindow position after filtering

karupayun:
We had the issue that after filtering sometimes the capture windows
doesn't show anything because as we are not showing some tracks anymore,
the old position doesn't have any track. In this PR we ask for an update
after updating tracks position, so the view y-position is clamped for the
new top-y and bottom-y coordinates.

Something to consider:
 - I don't like TrackManager asking CaptureWindows to update position.
 We should have a better way to do this (CaptureWindows updating its
 position after drawing, but as it wasn't that simple) I left as it is
 now.

 - I overloaded the method UpdateWorldTopLeftY because I feel that it is
 nicer (can be convinced about the opposite)

 - This solution came from the idea in
 https://github.com/google/orbit/pull/2089 (almost copy-paste) but without splitting
 UpdateTrackPosition and UpdateTrackPrimitives

Solve http://b/182978978.
Test: Load a capture, move the screen, filter.

### DIFF
--- a/src/OrbitGl/TimeGraph.cpp
+++ b/src/OrbitGl/TimeGraph.cpp
@@ -643,6 +643,9 @@ void TimeGraph::UpdatePrimitives(PickingMode picking_mode) {
   track_manager_->SortTracks();
   track_manager_->UpdateMovingTrackSorting();
   track_manager_->UpdateTracks(&batcher_, min_tick, max_tick, picking_mode);
+  // Coordinates from CaptureWindows could need an update if we modified the vertical size of some
+  // track.
+  GetCanvas()->UpdateWorldTopLeftY();
 
   update_primitives_requested_ = false;
 }
--- a/src/OrbitGl/GlCanvas.h
+++ b/src/OrbitGl/GlCanvas.h
@@ -95,6 +95,7 @@ class GlCanvas {
   [[nodiscard]] float GetWorldMaxY() const { return world_max_y_; }
   [[nodiscard]] float GetWorldTopLeftX() const { return world_top_left_x_; }
   [[nodiscard]] float GetWorldTopLeftY() const { return world_top_left_y_; }
+  void UpdateWorldTopLeftY() { UpdateWorldTopLeftY(GetWorldTopLeftY()); }
   virtual void UpdateWorldTopLeftY(float val) { world_top_left_y_ = val; }
 
   [[nodiscard]] TextRenderer& GetTextRenderer() { return text_renderer_; }


florian-kuebler (commented):
Do you have an idea how to address this outside of trackmanager?

florian-kuebler:
> Do you have an idea how to address this outside of trackmanager?

Quickly looked into the code. `TimeGraph` is the only caller of `TackManager::UpdateTracks`. Shouldn't we just move the call to `GetCanvas()->UpdateWorldTopLeftY()` right after this call in `TimeGraph`?

florian-kuebler (approved):
LGTM. Thanks.

karupayun:
> > Do you have an idea how to address this outside of trackmanager?
> 
> Quickly looked into the code. `TimeGraph` is the only caller of `TackManager::UpdateTracks`. Shouldn't we just move the call to `GetCanvas()->UpdateWorldTopLeftY()` right after this call in `TimeGraph`?

Yes, thanks for catching this. I only was thinking on solving this inside CaptureWindows and I forget this quick-win. I still think this should to be manage with CaptureWindows itself, but we are one step ahead xD

src/OrbitGl/CaptureWindow.cpp
```diff
@@ -686,6 +686,8 @@ void CaptureWindow::UpdateVerticalSliderFromWorld() {
   vertical_slider_->SetOrthogonalSliderPixelHeight(slider_->GetPixelHeight());
 }
 
+void CaptureWindow::UpdateWorldTopLeftY() { UpdateWorldTopLeftY(GetWorldTopLeftY()); }
```
reichlfl, line 689: Nit: This implementation should already be in GlCanvas, I'd argue the method should not even be virtual.
But feel free to ignore it for this PR, the inheritance of GlCanvas and CaptureWindow is a bit of a mess anyways and I'm working on cleaning that up.

src/OrbitGl/TimeGraph.cpp
```diff
@@ -643,6 +643,9 @@ void TimeGraph::UpdatePrimitives(PickingMode picking_mode) {
   track_manager_->SortTracks();
   track_manager_->UpdateMovingTrackSorting();
   track_manager_->UpdateTracks(&batcher_, min_tick, max_tick, picking_mode);
+  // Coordinates from CaptureWindows could need an update if we modified the vertical size of some
+  // track.
+  GetCanvas()->UpdateWorldTopLeftY();
```
reichlfl, line 648: I'm a bit suprised this fixes the issue, and I'm not sure it will work in all cases. Problem is, this is happening after redraw, so it forces another redraw / update primitives if the height of any track changed. This happens in `UpdateWorldTopLeftY()`, but the line right below this resets `update_primitives_requested_` to false... this should mean that the next frame will call `Draw`, but not `UpdatePrimitives` on all tracks, shouldn't this lead to issues?

Second issue: Re-requesting `Draw` / `UpdatePrimitives` could lead to issues with picking. If this happens when picking is active, AFAIK there will be no picking-redraw...

I think the only clean solution would be completely decoupling position / size calculation from drawing.

reichlfl (approved):
Thanks for this, the separate overload of `UpdateWorldTopLeftY()` is nice IMO. See my comment in TimeGraph.cpp - this is definitely better than before, so I'm fine with merging this, but I don't think it fixes the issue in all cases.

src/OrbitGl/CaptureWindow.cpp
karupayun: I was a bit unsure here, because it is an issue with CaptureWindow. Having this method in GlCanvas doesn't have sense, because it's not doing anything.

But I will move it because I also thought about that.

src/OrbitGl/TimeGraph.cpp
karupayun: I'm merging without addressing this issue. Feel free to open a bug for checking what it is actually happening in the drawing process of CaptureWindows.

PR #2127, opened 2021-03-23
Fix "stop capture" in E2E tests

reichlfl:
The Capture() test case in the E2E tests did not account for the "Finalizing Capture" dialog that may be shown if a capture takes time to finalize.

With this fix, all E2E tests wait for the "Finalizing Capture" dialog to disappear after the capture has been stopped.

### DIFF
--- a/contrib/automation_tests/test_cases/capture_window.py
+++ b/contrib/automation_tests/test_cases/capture_window.py
@@ -10,7 +10,7 @@ import time
 from typing import Tuple, List, Iterable
 
 from core.common_controls import Track
-from core.orbit_e2e import E2ETestCase, E2ETestSuite
+from core.orbit_e2e import E2ETestCase, E2ETestSuite, wait_for_condition
 from pywinauto.base_wrapper import BaseWrapper
 from pywinauto import mouse, keyboard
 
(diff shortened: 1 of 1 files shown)


contrib/automation_tests/test_cases/capture_window.py
```diff
@@ -233,11 +233,18 @@ def _take_capture(self, length_in_seconds: int):
...
+        self._wait_for_capture_completion()
 
     def _verify_existence_of_tracks(self):
         logging.info("Verifying existence of at least one track...")
         MatchTracks(expected_count=1, allow_additional_tracks=True).execute(self.suite)
 
+    def _wait_for_capture_completion(self):
+        logging.info("Waiting for capture to finalize...")
+        wait_for_condition(lambda: self.find_control('Window', 'Finalizing capture',
+                                                     recurse=False, raise_on_failure=False) is None, max_seconds=30)
```
florian-kuebler, line 245: Maybe we should increase the max seconds? On my machine it actually took a reasonable amount of time already (didn't measured though).

florian-kuebler (approved):
Thanks, I would increase the timer, but besides that, LGTM!

danielfenner (approved):
lgtm!

contrib/automation_tests/test_cases/capture_window.py
danielfenner: Good point, the video of the failed test shows the dialog for ~16 seconds.

reichlfl: You're right. I've changed it to 120 seconds now - better to be safe here, as the tests may run for multiple minutes anyways.

PR #2130, opened 2021-03-24
Fix orders of production limit line & white label

vickyliu-go4it:
The previous code puts the production limit line and the white label in
the same layer. As we draw primitives in the same layer in the order of
"Box --> Line --> Triangle", the red line of production limit appears
over the white label.

With this change, we fix the wrong ordering by putting the while label
in a different layer with the production limit.

Bug: http://b/183492649

Before the change:
red line

After:
Screenshot 2021-03-24 085503

### DIFF
--- a/src/OrbitGl/GlCanvas.cpp
+++ b/src/OrbitGl/GlCanvas.cpp
@@ -27,7 +27,8 @@ float GlCanvas::kZValueTrack = 0.01f;
 float GlCanvas::kZValueEventBar = 0.03f;
 float GlCanvas::kZValueBox = 0.05f;
 float GlCanvas::kZValueEvent = 0.07f;
-float GlCanvas::kZValueTrackText = 0.09f;
+float GlCanvas::kZValueTrackText = 0.08f;
+float GlCanvas::kZValueTrackLabel = 0.09f;
 float GlCanvas::kZValueOverlay = 0.43f;
 float GlCanvas::kZValueOverlayTextBackground = 0.45f;
 float GlCanvas::kZValueEventBarPicking = 0.49f;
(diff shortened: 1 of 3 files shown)


src/OrbitGl/GraphTrack.cpp
```diff
@@ -92,6 +92,7 @@ void GraphTrack::Draw(GlCanvas* canvas, PickingMode picking_mode, float z_offset
   const Color kBlack(0, 0, 0, 255);
   const Color kWhite(255, 255, 255, 255);
   float text_z = GlCanvas::kZValueEvent + z_offset;
+  float label_z = GlCanvas::kZValueTrackText + z_offset;
```
karupayun, line 95: I'm not sure here. We are using the Event layer to draw text and the text layer to draw the label. I feel we need a new label for this elements which has to be on top of the text. I'm not sure about the name, could be simply kZValueTrackLabel and I will propose (in GlCanvas):
Event -> 0.07
TrackText -> 0.08
TrackLabel ->0.09

After that I would write here:
```
  float text_z = GlCanvas::kZValueTrackText + z_offset;
  float label_z = GlCanvas::kZValueTrackLabel + z_offset;
```

karupayun (commented):
Thanks for solving this so fast Vicky! I made a suggestion in order to solved this as proper as possible and not using a patch. BTW, the code wasn't the perfect one (the text was drawn on the event layer), so you will also solve a possible future issue there :)

src/OrbitGl/GraphTrack.cpp
vickyliu-go4it: Thanks for the suggestion and also explaining the purpose of different layers to me! I have changed the code accordingly.

src/OrbitGl/GlCanvas.cpp
```diff
@@ -27,7 +27,8 @@ float GlCanvas::kZValueTrack = 0.01f;
 float GlCanvas::kZValueEventBar = 0.03f;
 float GlCanvas::kZValueBox = 0.05f;
 float GlCanvas::kZValueEvent = 0.07f;
-float GlCanvas::kZValueTrackText = 0.09f;
+float GlCanvas::kZValueTrackText = 0.08f;
+float GlCanvas::kZValueTrackLabel = 0.09f;
```
karupayun, line 31: Can I also ask you to modify line 46-48 with...
```
constexpr unsigned kNumberOriginalLayers = 17;
constexpr unsigned kExtraLayersForMovingTracks = 6;
constexpr unsigned kExtraLayersForPinnedTracks = 6;
```
?

karupayun (approved):
LGTM!!

src/OrbitGl/GlCanvas.cpp
vickyliu-go4it: Done! :-)

karupayun: Not your problem, we should change this manually updated numbers :-)

karupayun:
Thanks for addressing my comments! :)

PR #2131, opened 2021-03-24
Change track name matching to perfect match

reichlfl:
All functions that match track names now require a perfect match with
the expected search string. Wildcards (*) are allowed to achieve the
previous functionality.

Affected tests have been adjusted accordingly.

### DIFF
--- a/contrib/automation_tests/orbit_tracks.py
+++ b/contrib/automation_tests/orbit_tracks.py
@@ -18,7 +18,7 @@ def main(argv):
         Capture(),
         # "sdma0" is not present on the DevKits, instead there is "vce0", so this tests for "sdma0 or vce0"
         MatchTracks(expected_names=[
-            "Scheduler", ("sdma0", "vce0"), "gfx", "All Threads", "hello_ggp_stand"],
+            "Scheduler", ("*sdma0*", "*vce0*"), "gfx", "All Threads", "hello_ggp_stand"],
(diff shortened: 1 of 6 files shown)


florian-kuebler (approved):
LGTM. Thanks!

contrib/automation_tests/orbit_tracks.py
```diff
@@ -18,7 +18,7 @@ def main(argv):
         Capture(),
         # "sdma0" is not present on the DevKits, instead there is "vce0", so this tests for "sdma0 or vce0"
         MatchTracks(expected_names=[
-            "Scheduler", ("sdma0", "vce0"), "gfx", "All Threads", "hello_ggp_stand"],
+            "Scheduler", ("*sdma0*", "*vce0*"), "gfx", "All Threads", "hello_ggp_stand"],
```
danielfenner, line 21: Is this matching done on purpose? Iirc the track names are "sdma0" or "vce0" exactly. Doesn't really matter though I think.

danielfenner (approved):
lgtm!

contrib/automation_tests/orbit_tracks.py
reichlfl: I wasn't 100% sure about the DevKits, so I thought it's better to be safe here. This is basically equivalent to the old behavior.

PR #2142, opened 2021-03-31
Fix Crashs in E2E Tests.

florian-kuebler:
With PR #2138, a couple of CHECKs for the child index were introduced.
However, it seems that on a local setup, invalid indexes gets querried.

Test: Run Vulkan layer E2E Test locally.


reichlfl (approved):
LGTM!

PR #2144, opened 2021-04-01
GlCanvas cleanup

reichlfl:
This is a first refactor of GlCanvas / CaptureWindow. Sorry for the large PR, I kept running into small inconsistencies during the refactor that made it hard to find a reasonable cutoff point, and I wanted to avoid submitting it half-done because I had to revisit and change previous commits quite a bit as my assumptions continued to be incorrect...

### Goals of this PR

**Remove the code duplication and inconsistent behavior of `GlCanvas` and `CaptureWindow`** 

A lot of functionality (e.g. picking) was partially implemented in `GlCanvas`, but only worked for `CaptureWindow`. Other functionality was implemented in `GlCanvas` and then reimplemented (sometimes with minor changes) in `CaptureWindow`. Now, `GlCanvas` fully takes care of picking and tracking mouse interactions, whereas `CaptureWindow` add all functionality related to `TimeGraph` and Scrollbars.

**Break down GlCanvas to improve Unit Test coverage**

As a first step, all functionality related to viewport sizes, coordinate conversion, and virtual "camera placement" (e.g. position of the screen within the 2D world) moved to a new class `Viewport`. This new class is properly unit tested.

### Next steps
* Pass `Batcher*` and `Viewport*` instead of `GlCanvas*` where possible. This will decouple most components from `GlCanvas`, and will decouple rendering from coordinate translation.
* Change horizontal scrolling and scaling to use the `Viewport` functionality. Currently, this is handled by `TimeGraph`, which is quite inconsistent (`Viewport::GetWorldTopLeft()[0]` is always 0!)
* Move the layering definitions into a separate class
* Move the actual "Renderpass" functionality into a new class that is linked to a viewport - so we get rid of the implicit assumptions that "some elements are drawn in screen space, others in world space", and remove the implicit 2nd renderpass that happens on picking, as this causes issues with state changes in between those two renderpasses.

### DIFF
--- a/src/OrbitGl/AccessibleThreadBar.cpp
+++ b/src/OrbitGl/AccessibleThreadBar.cpp
@@ -8,6 +8,7 @@
 #include "OrbitBase/Logging.h"
 #include "ThreadBar.h"
 #include "Track.h"
+#include "Viewport.h"
 
 namespace orbit_gl {
 
@@ -45,12 +46,12 @@ orbit_accessibility::AccessibilityRect orbit_gl::AccessibleThreadBar::Accessible
 
   GlCanvas* canvas = thread_bar_->GetCanvas();
   CHECK(canvas != nullptr);
+  const Viewport& viewport = canvas->GetViewport();
 
   return orbit_accessibility::AccessibilityRect(
-      canvas->GetViewport().WorldToScreenWidth(local_pos[0]),
-      canvas->GetViewport().WorldToScreenHeight(local_pos[1]),
-      canvas->GetViewport().WorldToScreenWidth(thread_bar_->GetSize()[0]),
-      canvas->GetViewport().WorldToScreenHeight(thread_bar_->GetSize()[1]));
+      viewport.WorldToScreenWidth(local_pos[0]), viewport.WorldToScreenHeight(local_pos[1]),
+      viewport.WorldToScreenWidth(thread_bar_->GetSize()[0]),
+      viewport.WorldToScreenHeight(thread_bar_->GetSize()[1]));
 }
 
 orbit_accessibility::AccessibilityState AccessibleThreadBar::AccessibleState() const {
--- a/src/OrbitGl/AccessibleThreadBar.cpp
+++ b/src/OrbitGl/AccessibleThreadBar.cpp
@@ -47,9 +47,10 @@ orbit_accessibility::AccessibilityRect orbit_gl::AccessibleThreadBar::Accessible
   CHECK(canvas != nullptr);
 
   return orbit_accessibility::AccessibilityRect(
-      canvas->WorldToScreenWidth(local_pos[0]), canvas->WorldToScreenHeight(local_pos[1]),
-      canvas->WorldToScreenWidth(thread_bar_->GetSize()[0]),
-      canvas->WorldToScreenHeight(thread_bar_->GetSize()[1]));
+      canvas->GetViewport().WorldToScreenWidth(local_pos[0]),
+      canvas->GetViewport().WorldToScreenHeight(local_pos[1]),
+      canvas->GetViewport().WorldToScreenWidth(thread_bar_->GetSize()[0]),
+      canvas->GetViewport().WorldToScreenHeight(thread_bar_->GetSize()[1]));
 }
 
 orbit_accessibility::AccessibilityState AccessibleThreadBar::AccessibleState() const {
--- a/src/OrbitGl/AccessibleTimeGraph.cpp
+++ b/src/OrbitGl/AccessibleTimeGraph.cpp
@@ -11,15 +11,19 @@
 #include "TimeGraph.h"
 #include "Track.h"
 #include "TrackManager.h"
+#include "Viewport.h"
 
 using orbit_accessibility::AccessibilityRect;
 using orbit_accessibility::AccessibilityState;
 using orbit_accessibility::AccessibleInterface;
 
+namespace orbit_gl {
+
 AccessibilityRect TimeGraphAccessibility::AccessibleLocalRect() const {
   GlCanvas* canvas = time_graph_->GetCanvas();
-  return AccessibilityRect(0, 0, canvas->GetViewport().GetWidth(),
-                           canvas->GetViewport().GetHeight());
+  const Viewport& viewport = canvas->GetViewport();
+
+  return AccessibilityRect(0, 0, viewport.GetScreenWidth(), viewport.GetScreenHeight());
 }
 
 AccessibilityState TimeGraphAccessibility::AccessibleState() const {
@@ -39,3 +43,5 @@ const AccessibleInterface* TimeGraphAccessibility::AccessibleChild(int index) co
 const AccessibleInterface* TimeGraphAccessibility::AccessibleParent() const {
   return time_graph_->GetCanvas()->GetOrCreateAccessibleInterface();
 }
+
+}  // namespace orbit_gl
--- a/src/OrbitGl/AccessibleTimeGraph.cpp
+++ b/src/OrbitGl/AccessibleTimeGraph.cpp
@@ -18,7 +18,8 @@ using orbit_accessibility::AccessibleInterface;
 
 AccessibilityRect TimeGraphAccessibility::AccessibleLocalRect() const {
   GlCanvas* canvas = time_graph_->GetCanvas();
-  return AccessibilityRect(0, 0, canvas->GetWidth(), canvas->GetHeight());
+  return AccessibilityRect(0, 0, canvas->GetViewport().GetWidth(),
+                           canvas->GetViewport().GetHeight());
 }
 
 AccessibilityState TimeGraphAccessibility::AccessibleState() const {
--- a/src/OrbitGl/AccessibleTrack.cpp
+++ b/src/OrbitGl/AccessibleTrack.cpp
@@ -184,8 +184,9 @@ AccessibilityRect AccessibleTrack::AccessibleLocalRect() const {
   // in a final result with width / height of 0.
 
   // First: Clamp bottom
-  if (screen_pos[1] + screen_height > canvas->GetHeight()) {
-    screen_height = std::max(0, canvas->GetHeight() - static_cast<int>(screen_pos[1]));
+  if (screen_pos[1] + screen_height > canvas->GetViewport().GetHeight()) {
+    screen_height =
+        std::max(0, canvas->GetViewport().GetHeight() - static_cast<int>(screen_pos[1]));
   }
   // Second: Clamp top
   if (screen_pos[1] < 0) {
--- a/src/OrbitGl/AccessibleTrack.cpp
+++ b/src/OrbitGl/AccessibleTrack.cpp
@@ -105,11 +105,12 @@ AccessibilityRect AccessibleTrackContent::AccessibleLocalRect() const {
   CHECK(track_ != nullptr);
 
   GlCanvas* canvas = track_->GetCanvas();
+  const Viewport& viewport = canvas->GetViewport();
 
-  return AccessibilityRect(canvas->GetViewport().WorldToScreenWidth(0),
-                           canvas->GetViewport().WorldToScreenHeight(layout_->GetTrackTabHeight()),
-                           canvas->GetViewport().WorldToScreenWidth(track_->GetSize()[0]),
-                           canvas->GetViewport().WorldToScreenHeight(track_->GetSize()[1]));
+  return AccessibilityRect(viewport.WorldToScreenWidth(0),
+                           viewport.WorldToScreenHeight(layout_->GetTrackTabHeight()),
+                           viewport.WorldToScreenWidth(track_->GetSize()[0]),
+                           viewport.WorldToScreenHeight(track_->GetSize()[1]));
 }
 
 AccessibilityState AccessibleTrackContent::AccessibleState() const {
@@ -135,11 +136,12 @@ AccessibilityRect AccessibleTrackTab::AccessibleLocalRect() const {
   CHECK(track_ != nullptr);
 
   GlCanvas* canvas = track_->GetCanvas();
+  const Viewport& viewport = canvas->GetViewport();
 
-  return AccessibilityRect(canvas->GetViewport().WorldToScreenWidth(layout_->GetTrackTabOffset()),
-                           canvas->GetViewport().WorldToScreenHeight(0),
-                           canvas->GetViewport().WorldToScreenWidth(layout_->GetTrackTabWidth()),
-                           canvas->GetViewport().WorldToScreenHeight(layout_->GetTrackTabHeight()));
+  return AccessibilityRect(viewport.WorldToScreenWidth(layout_->GetTrackTabOffset()),
+                           viewport.WorldToScreenHeight(0),
+                           viewport.WorldToScreenWidth(layout_->GetTrackTabWidth()),
+                           viewport.WorldToScreenHeight(layout_->GetTrackTabHeight()));
 }
 
 AccessibilityState AccessibleTrackTab::AccessibleState() const {
@@ -175,18 +177,19 @@ AccessibilityRect AccessibleTrack::AccessibleLocalRect() const {
   Vec2 size = track_->GetSize();
   size[1] += layout_->GetTrackTabHeight();
 
-  Vec2i screen_pos = canvas->GetViewport().WorldToScreenPos(pos);
-  int screen_width = canvas->GetViewport().WorldToScreenWidth(size[0]);
-  int screen_height = canvas->GetViewport().WorldToScreenHeight(size[1]);
+  const Viewport& viewport = canvas->GetViewport();
+
+  Vec2i screen_pos = viewport.WorldToScreenPos(pos);
+  int screen_width = viewport.WorldToScreenWidth(size[0]);
+  int screen_height = viewport.WorldToScreenHeight(size[1]);
 
   // Adjust the coordinates to clamp the result to an on-screen rect
   // This will "cut" any part that is offscreen due to scrolling, and may result
   // in a final result with width / height of 0.
 
   // First: Clamp bottom
-  if (screen_pos[1] + screen_height > canvas->GetViewport().GetHeight()) {
-    screen_height =
-        std::max(0, canvas->GetViewport().GetHeight() - static_cast<int>(screen_pos[1]));
+  if (screen_pos[1] + screen_height > viewport.GetScreenHeight()) {
+    screen_height = std::max(0, viewport.GetScreenHeight() - static_cast<int>(screen_pos[1]));
   }
   // Second: Clamp top
   if (screen_pos[1] < 0) {
--- a/src/OrbitGl/AccessibleTrack.cpp
+++ b/src/OrbitGl/AccessibleTrack.cpp
@@ -106,10 +106,10 @@ AccessibilityRect AccessibleTrackContent::AccessibleLocalRect() const {
 
   GlCanvas* canvas = track_->GetCanvas();
 
-  return AccessibilityRect(canvas->WorldToScreenWidth(0),
-                           canvas->WorldToScreenHeight(layout_->GetTrackTabHeight()),
-                           canvas->WorldToScreenWidth(track_->GetSize()[0]),
-                           canvas->WorldToScreenHeight(track_->GetSize()[1]));
+  return AccessibilityRect(canvas->GetViewport().WorldToScreenWidth(0),
+                           canvas->GetViewport().WorldToScreenHeight(layout_->GetTrackTabHeight()),
+                           canvas->GetViewport().WorldToScreenWidth(track_->GetSize()[0]),
+                           canvas->GetViewport().WorldToScreenHeight(track_->GetSize()[1]));
 }
 
 AccessibilityState AccessibleTrackContent::AccessibleState() const {
@@ -136,10 +136,10 @@ AccessibilityRect AccessibleTrackTab::AccessibleLocalRect() const {
 
   GlCanvas* canvas = track_->GetCanvas();
 
-  return AccessibilityRect(canvas->WorldToScreenWidth(layout_->GetTrackTabOffset()),
-                           canvas->WorldToScreenHeight(0),
-                           canvas->WorldToScreenWidth(layout_->GetTrackTabWidth()),
-                           canvas->WorldToScreenHeight(layout_->GetTrackTabHeight()));
+  return AccessibilityRect(canvas->GetViewport().WorldToScreenWidth(layout_->GetTrackTabOffset()),
+                           canvas->GetViewport().WorldToScreenHeight(0),
+                           canvas->GetViewport().WorldToScreenWidth(layout_->GetTrackTabWidth()),
+                           canvas->GetViewport().WorldToScreenHeight(layout_->GetTrackTabHeight()));
 }
 
 AccessibilityState AccessibleTrackTab::AccessibleState() const {
@@ -175,9 +175,9 @@ AccessibilityRect AccessibleTrack::AccessibleLocalRect() const {
   Vec2 size = track_->GetSize();
   size[1] += layout_->GetTrackTabHeight();
 
-  Vec2 screen_pos = canvas->WorldToScreen(pos);
-  int screen_width = canvas->WorldToScreenWidth(size[0]);
-  int screen_height = canvas->WorldToScreenHeight(size[1]);
+  Vec2i screen_pos = canvas->GetViewport().WorldToScreenPos(pos);
+  int screen_width = canvas->GetViewport().WorldToScreenWidth(size[0]);
+  int screen_height = canvas->GetViewport().WorldToScreenHeight(size[1]);
 
   // Adjust the coordinates to clamp the result to an on-screen rect
   // This will "cut" any part that is offscreen due to scrolling, and may result
--- a/src/OrbitGl/AccessibleTriangleToggle.cpp
+++ b/src/OrbitGl/AccessibleTriangleToggle.cpp
@@ -17,9 +17,11 @@ orbit_accessibility::AccessibilityRect AccessibleTriangleToggle::AccessibleLocal
   Vec2 pos = triangle_toggle_->GetPos();
   Vec2 size = triangle_toggle_->GetSize();
 
-  Vec2i screen_pos = canvas->GetViewport().WorldToScreenPos(pos);
-  int screen_width = canvas->GetViewport().WorldToScreenWidth(size[0]);
-  int screen_height = canvas->GetViewport().WorldToScreenHeight(size[1]);
+  const Viewport& viewport = canvas->GetViewport();
+
+  Vec2i screen_pos = viewport.WorldToScreenPos(pos);
+  int screen_width = viewport.WorldToScreenWidth(size[0]);
+  int screen_height = viewport.WorldToScreenHeight(size[1]);
 
   orbit_accessibility::AccessibilityRect parent_rect =
       parent_->GetOrCreateAccessibleInterface()->AccessibleLocalRect();
--- a/src/OrbitGl/AccessibleTriangleToggle.cpp
+++ b/src/OrbitGl/AccessibleTriangleToggle.cpp
@@ -17,16 +17,16 @@ orbit_accessibility::AccessibilityRect AccessibleTriangleToggle::AccessibleLocal
   Vec2 pos = triangle_toggle_->GetPos();
   Vec2 size = triangle_toggle_->GetSize();
 
-  Vec2 screen_pos = canvas->WorldToScreen(pos);
-  int screen_width = canvas->WorldToScreenWidth(size[0]);
-  int screen_height = canvas->WorldToScreenHeight(size[1]);
+  Vec2i screen_pos = canvas->GetViewport().WorldToScreenPos(pos);
+  int screen_width = canvas->GetViewport().WorldToScreenWidth(size[0]);
+  int screen_height = canvas->GetViewport().WorldToScreenHeight(size[1]);
 
   orbit_accessibility::AccessibilityRect parent_rect =
       parent_->GetOrCreateAccessibleInterface()->AccessibleLocalRect();
 
-  return orbit_accessibility::AccessibilityRect(static_cast<int>(screen_pos[0]) - parent_rect.left,
-                                                static_cast<int>(screen_pos[1]) - parent_rect.top,
-                                                screen_width, screen_height);
+  return orbit_accessibility::AccessibilityRect(screen_pos[0] - parent_rect.left,
+                                                screen_pos[1] - parent_rect.top, screen_width,
+                                                screen_height);
 }
 
 orbit_accessibility::AccessibilityState AccessibleTriangleToggle::AccessibleState() const {
--- a/src/OrbitGl/CaptureWindow.cpp
+++ b/src/OrbitGl/CaptureWindow.cpp
@@ -69,7 +69,6 @@ using orbit_client_protos::TimerInfo;
 
 CaptureWindow::CaptureWindow(OrbitApp* app) : GlCanvas(), app_{app} {
   draw_help_ = true;
-  draw_filter_ = false;
   world_top_left_x_ = 0;
   world_top_left_y_ = 0;
   world_max_y_ = 0;
@@ -97,8 +96,6 @@ CaptureWindow::CaptureWindow(OrbitApp* app) : GlCanvas(), app_{app} {
   slider_->SetOrthogonalSliderPixelHeight(vertical_slider_->GetPixelHeight());
 }
 
-void CaptureWindow::OnTimer() { GlCanvas::OnTimer(); }
-
 void CaptureWindow::ZoomAll() {
   ResetHoverTimer();
   RequestUpdatePrimitives();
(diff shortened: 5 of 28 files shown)


src/OrbitGl/ViewportTest.cpp
```diff
@@ -0,0 +1,176 @@
...
+  viewport.SetWorldTopLeftY(-200.f);
+  EXPECT_EQ(viewport.GetWorldTopLeft(), Vec2(100.f, -100.f));
+  EXPECT_EQ(viewport.IsDirty(), false);
+
+  // Resizing should clamp
+  viewport.SetWorldExtents(150, 250);
+  EXPECT_EQ(viewport.GetWorldTopLeft(), Vec2(50.f, -50.f));
+}
+
+TEST(Viewport, ScrollingWithZoom) {
```
karupayun, line 114: Could we also add a test where we zoom in the middle of the test and check if the point_new_position is where it should be?

src/OrbitGl/ViewportTest.cpp
```diff
@@ -0,0 +1,176 @@
...
+
+  EXPECT_EQ(viewport.QtToGlScreenPos(screen_pos),
+            Vec2i(screen_pos[0], viewport.GetHeight() - screen_pos[1]));
+}
+
+TEST(Viewport, CoordinateConversion) {
+  Viewport viewport(10, 100);
+  // Extents are large enough to scroll
+  viewport.SetWorldExtents(500, 500);
+  viewport.SetWorldWidth(20.f);
```
karupayun, line 161: A small comment here to explain the zoom-level?
Something like this:
// horizontal zoom_level = 2, vertical zoom-level = 0.5?

src/OrbitGl/GlCanvas.cpp
```diff
@@ -176,11 +167,16 @@ void GlCanvas::MouseMoved(int x, int y, bool left, bool /*right*/, bool /*middle
   mouse_screen_y_ = mouse_y;
 
   // Pan
-  if (left && !im_gui_active_) {
-    world_top_left_x_ = world_click_x_ -
-                        static_cast<float>(mouse_x) / static_cast<float>(GetWidth()) * world_width_;
-    world_top_left_y_ = world_click_y_ + static_cast<float>(mouse_y) /
-                                             static_cast<float>(GetHeight()) * world_height_;
+  if (left && !picking_manager_.IsDragging()) {
```
karupayun, line 146: Why the dragging here? Were there a bug?

karupayun (commented):
I'm only commenting the first commit here. In the second commit, everything seems to makes sense, but I would like to review the GlCanvas part.

pierricgimmig (commented):
Hi @reichlfl , thanks a lot for this, looks really good. I love to see all this red in the diff view :-). Please see comments. Also, I tested the branch and I have found a few bugs:

- Green selection rectangle is not sticky anymore (is this intended?). It is an important feature for the upcoming selection stats.
- Right click used to give Orbit focus, now if Orbit is not in focus and we try to right-click and drag, the green rectangle does not appear
- Vertical scrolling in a single direction makes the vertical slider jump up and down (see video)


https://user-images.githubusercontent.com/3687222/113441777-baed5380-93a3-11eb-972d-ecd50b5aa217.mov

src/OrbitGl/GlCanvas.cpp
```diff
@@ -241,29 +195,35 @@ void GlCanvas::LeftUp() {
...
 
   Orbit_ImGui_MouseButtonCallback(imgui_context_, 1, true);
   RequestRedraw();
 }
 
 bool GlCanvas::RightUp() {
   Orbit_ImGui_MouseButtonCallback(imgui_context_, 1, false);
-  is_selecting_ = true;
+  is_selecting_ = false;
+  select_start_ = select_stop_;
```
pierricgimmig, line 217: This line makes the green selection rectangle disappear on "right-up" which is a change in behavior. I think we should keep the previous behavior.

src/OrbitGl/Viewport.h
```diff
@@ -0,0 +1,85 @@
...
+// found in the LICENSE file.
+
+#ifndef ORBIT_GL_VIEWPORT_H_
+#define ORBIT_GL_VIEWPORT_H_
+
+#include "CoreMath.h"
+
+namespace orbit_gl {
+
+/**
```
florian-kuebler, line 12: Nit: If I am not wrong we avoid block comments and use `//` at all places. However, even if not it should be
```
/*
 *
 */
```

src/OrbitGl/Viewport.h
```diff
@@ -0,0 +1,85 @@
...
+y coordinates are negative). Viewport clamps all X values to be >= 0, and all Y values to be <= 0.
+
+See TODO(b/177350599): Unify QtScreen and GlScreen
+**/
+class Viewport {
+ public:
+  explicit Viewport(int width, int height);
+
+  void Resize(int width, int height);
+  [[nodiscard]] int GetWidth() const;
```
florian-kuebler, line 40: When reading this and the class level comment, I am wondering if we should make it more clear, that those values refer to the screen, (e.g. `GetScreenWidth`, as for the world below).

src/OrbitGl/Viewport.h
```diff
@@ -0,0 +1,85 @@
...
+  void SetWorldWidth(float width);
+  void SetWorldHeight(float height);
+  [[nodiscard]] float GetWorldWidth() const;
+  [[nodiscard]] float GetWorldHeight() const;
+
+  // Absolute size of the world
+  void SetWorldExtents(float width, float height);
+  [[nodiscard]] Vec2 GetWorldExtents();
+
+  void SetWorldTopLeftY(float y);
```
florian-kuebler, line 53: Will they be called individually? Otherwise, consider passing a vec. to it.

Also I think there should be a comment.

src/OrbitGl/Viewport.h
```diff
@@ -0,0 +1,85 @@
...
+**/
+class Viewport {
+ public:
+  explicit Viewport(int width, int height);
+
+  void Resize(int width, int height);
+  [[nodiscard]] int GetWidth() const;
+  [[nodiscard]] int GetHeight() const;
+
+  // Visible size of the world
```
florian-kuebler, line 43: Nit: Add `.` and below.
https://google.github.io/styleguide/cppguide.html#Punctuation,_Spelling_and_Grammar

src/OrbitGl/Viewport.h
```diff
@@ -0,0 +1,85 @@
...
+  // Absolute size of the world
+  void SetWorldExtents(float width, float height);
+  [[nodiscard]] Vec2 GetWorldExtents();
+
+  void SetWorldTopLeftY(float y);
+  void SetWorldTopLeftX(float x);
+  [[nodiscard]] const Vec2& GetWorldTopLeft() const;
+
+  [[nodiscard]] Vec2 ScreenToWorldPos(const Vec2i& screen_pos) const;
+  [[nodiscard]] float ScreenToWorldHeight(int height) const;
```
florian-kuebler, line 58: `Height` or `Y`? Same for the width.

src/OrbitGl/Viewport.h
```diff
@@ -0,0 +1,85 @@
...
+  [[nodiscard]] float ScreenToWorldWidth(int width) const;
+
+  [[nodiscard]] Vec2i WorldToScreenPos(const Vec2& world_pos) const;
+  [[nodiscard]] int WorldToScreenHeight(float height) const;
+  [[nodiscard]] int WorldToScreenWidth(float width) const;
+
+  [[nodiscard]] Vec2i QtToGlScreenPos(const Vec2i& qt_pos) const;
+
+  [[nodiscard]] bool IsDirty() const { return is_dirty_; }
```
florian-kuebler, line 67: Could you add a comment here? To me it is not clear what `Dirty` means in this context.

src/OrbitGl/ViewportTest.cpp
```diff
@@ -0,0 +1,176 @@
...
+#include "Viewport.h"
+
+namespace orbit_gl {
+
+TEST(Viewport, ResizingAndDirty) {
+  Viewport viewport(100, 200);
+  EXPECT_TRUE(viewport.IsDirty());
+  viewport.ClearDirtyFlag();
+
+  // Test initial values
```
florian-kuebler, line 16: Nit: Consider adding punctation to the comments here.

src/OrbitGl/ViewportTest.cpp
```diff
@@ -0,0 +1,176 @@
...
+#include <gtest/gtest.h>
+
+#include "Viewport.h"
+
+namespace orbit_gl {
+
+TEST(Viewport, ResizingAndDirty) {
+  Viewport viewport(100, 200);
+  EXPECT_TRUE(viewport.IsDirty());
+  viewport.ClearDirtyFlag();
```
florian-kuebler, line 14: Also add an `EXPECT_FALSE(viewport.IsDirty()` afterwards?

src/OrbitGl/Viewport.h
```diff
@@ -0,0 +1,85 @@
...
+
+Screen:
+  (0, 0) ---->
+    |
+    |
+    v
+
+where world(0, 0) is initially anchored at screen(0, 0), i.e. top left (meaning basically all world
+y coordinates are negative). Viewport clamps all X values to be >= 0, and all Y values to be <= 0.
+
```
florian-kuebler, line 32: I would also expect at least some very brief comments about the usage (e.g. that the ClearDirtyFlag needs to be called).

src/OrbitGl/ViewportTest.cpp
```diff
@@ -0,0 +1,176 @@
...
+  viewport.SetWorldHeight(600.f);
+  EXPECT_TRUE(viewport.IsDirty());
+  viewport.ClearDirtyFlag();
+
+  EXPECT_EQ(viewport.GetWidth(), 1000);
+  EXPECT_EQ(viewport.GetHeight(), 2000);
+  EXPECT_EQ(viewport.GetWorldWidth(), 500.f);
+  EXPECT_EQ(viewport.GetWorldHeight(), 600.f);
+  EXPECT_EQ(viewport.GetWorldTopLeft(), Vec2(0.f, 0.f));
+  EXPECT_EQ(viewport.GetWorldExtents(), Vec2(100, 200));
```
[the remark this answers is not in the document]
florian-kuebler: I agree with this being not super easy to grasp at the first glance. Maybe consider adding a more detailed comment on the method (IDE's should show the comments also on the caller).

src/OrbitGl/Viewport.cpp
```diff
@@ -0,0 +1,130 @@
...
+
+#include "Viewport.h"
+
+#include <algorithm>
+
+#include "OrbitBase/Logging.h"
+
+namespace orbit_gl {
+
+Viewport::Viewport(int width, int height) : width_(width), height_(height) {
```
florian-kuebler, line 13: Nit: also `world_width_(width), world_height_(height)`?

src/OrbitGl/Viewport.cpp
```diff
@@ -0,0 +1,130 @@
...
+  CHECK(height > 0);
+
+  if (width == width_ && height == height_) return;
+
+  width_ = width;
+  height_ = height;
+  FlagAsDirty();
+}
+
+int Viewport::GetWidth() const { return width_; }
```
florian-kuebler, line 30: Nit: Those one-liners can be moved to the header.

src/OrbitGl/Viewport.h
```diff
@@ -0,0 +1,85 @@
...
+  (0, 0) ---->
+
+Screen:
+  (0, 0) ---->
+    |
+    |
+    v
+
+where world(0, 0) is initially anchored at screen(0, 0), i.e. top left (meaning basically all world
+y coordinates are negative). Viewport clamps all X values to be >= 0, and all Y values to be <= 0.
```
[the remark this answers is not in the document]
reichlfl: IMO we shouldn't, because it's part of the Viewport's functionality to do this, and happened in other places in GlCanvas / CaptureWindow before. The unit tests explicitly check for this as intended behavior.
I could see value in making the min-value (currently always 0) configurable, then this would be the equivalent to worldExtents, WDYT?

src/OrbitGl/Viewport.cpp
```diff
@@ -0,0 +1,130 @@
...
+
+int Viewport::GetHeight() const { return height_; }
+
+void Viewport::SetWorldWidth(float width) {
+  if (width == world_width_) return;
+
+  world_width_ = width;
+  // Recalculate required scrolling
```
florian-kuebler, line 38: Nit: Add `.` and below.

src/OrbitGl/ViewportTest.cpp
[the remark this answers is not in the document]
reichlfl: Good point. I was thinking the same and was considering to rename `WorldWidth` to `WorldVisibleWidth` or something, and decided against it to make it easier to review (as the naming doesn't change so much). But I think I'll actually add the renaming to this PR.

src/OrbitGl/ThreadTrack.cpp
```diff
@@ -276,17 +276,17 @@ void ThreadTrack::Draw(GlCanvas* canvas, PickingMode picking_mode, float z_offse
   const float tracepoint_track_height = layout_->GetEventTrackHeight();
 
   if (!thread_state_bar_->IsEmpty()) {
-    thread_state_bar_->SetSize(canvas->GetWorldWidth(), thread_state_track_height);
+    thread_state_bar_->SetSize(canvas->GetViewport().GetWorldWidth(), thread_state_track_height);
```
florian-kuebler, line 279: Nit: Extract the world_width into a local var.?

src/OrbitGl/GlCanvas.cpp
reichlfl: `picking_manager_.IsDragging() == true` means that the user clicked on an element in the capture window and is dragging across it. In this case, there should be no panning happening in GlCanvas, the current element will handle any drag-related actions. This is the same logic as before, it was just moved from `CaptureWindow` to `GlCanvas`.

src/OrbitGl/CaptureWindow.cpp
```diff
@@ -642,13 +470,12 @@ void CaptureWindow::UpdateHorizontalScroll(float ratio) {
 
 void CaptureWindow::UpdateVerticalScroll(float ratio) {
   if (time_graph_ == nullptr) return;
-  float min = world_max_y_;
-  float max = world_height_ - time_graph_->GetTrackManager()->GetTracksTotalHeight();
+  float min = 0.0f;
```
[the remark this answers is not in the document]
reichlfl: `world_max_y_` was 0 before since all coordinates in the world have negative y. This is now managed by the `Viewport`, but maybe it makes sense to retrieve the max-value from there instead of hardcoding it.

reichlfl:
> * Green selection rectangle is not sticky anymore (is this intended?). It is an important feature for the upcoming selection stats.
> * Right click used to give Orbit focus, now if Orbit is not in focus and we try to right-click and drag, the green rectangle does not appear
> * Vertical scrolling in a single direction makes the vertical slider jump up and down (see video)

The intent was to keep the existing behavior, so all changes mentioned above can be considered bugs. I'll take a look and fix them.

src/OrbitGl/Viewport.cpp
```diff
@@ -0,0 +1,130 @@
...
+
+void Viewport::Resize(int width, int height) {
+  CHECK(width > 0);
+  CHECK(height > 0);
+
+  if (width == width_ && height == height_) return;
+
+  width_ = width;
+  height_ = height;
+  FlagAsDirty();
```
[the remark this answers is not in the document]
reichlfl: Yes, any action that changes the visible or total area of the world. Basically anything that would require `UpdatePrimitives` on the time graph.

src/OrbitGl/CaptureWindow.cpp
```diff
@@ -247,276 +188,161 @@ void CaptureWindow::SelectTextBox(const TextBox* text_box) {
...
-  screen_click_y_ = y;
-  Pick();
+  GlCanvas::PostRender();
 
-  is_selecting_ = true;
-  select_start_ = Vec2(world_click_x_, world_click_y_);
-  select_stop_ = select_start_;
-  time_start_ = time_graph_->GetTickFromWorld(world_click_x_);
-  time_stop_ = time_start_;
+  // TODO: This should not be needed, but the extents of track heights may change after drawing
```
florian-kuebler, line 198: Consider creating a bug for this.

src/OrbitGl/CaptureWindow.cpp
```diff
@@ -675,25 +502,17 @@ void CaptureWindow::UpdateHorizontalSliderFromWorld() {
 
 void CaptureWindow::UpdateVerticalSliderFromWorld() {
   if (time_graph_ == nullptr) return;
-  float min = world_max_y_;
-  float max = world_height_ - time_graph_->GetTrackManager()->GetTracksTotalHeight();
-  float ratio = (world_top_left_y_ - min) / (max - min);
-  float vertical_ratio = world_height_ / time_graph_->GetTrackManager()->GetTracksTotalHeight();
+  float min = 0.0f;
```
florian-kuebler, line 505: Same as Pierric mentioned above.

src/OrbitGl/GlCanvas.cpp
```diff
@@ -1,16 +1,16 @@
+#include "GlCanvas.h"
```
florian-kuebler, line 1: The licence declaration should be the first thing in a file:
https://google.github.io/styleguide/cppguide.html#File_Comments

florian-kuebler: Also, it seems that you are including `GlCanvas` twice.

src/OrbitGl/CaptureWindow.cpp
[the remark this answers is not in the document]
karupayun: Yes, or at least a comment to explain why it is 0

src/OrbitGl/Viewport.cpp
```diff
@@ -0,0 +1,130 @@
...
+  SetWorldTopLeftX(world_top_left_[0]);
+  SetWorldTopLeftY(world_top_left_[1]);
+
+  FlagAsDirty();
+}
+
+Vec2 Viewport::GetWorldExtents() { return world_extents_; }
+
+void Viewport::SetWorldTopLeftY(float y) {
+  float clamped = std::min(std::max(y, world_height_ - world_extents_[1]), 0.f);
```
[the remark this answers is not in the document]
reichlfl: I've decided against this for 2 reasons:
1) I want the user of the class to be able to zoom out to more than 100% - for which you need to set world height > world extents
2) I don't want changes in the visible height to affect the extents because this could lead to unwanted side-effects.

I couldn't use `std::clamp` unfortunately since it doesn't deal with the case `world_extents_[1] < world_height_` correctly.

src/OrbitGl/GlCanvas.h
```diff
@@ -73,18 +72,14 @@ class GlCanvas {
 
   [[nodiscard]] TextRenderer& GetTextRenderer() { return text_renderer_; }
 
-  // TODO: Only needed for Graph labels
+  // Only needed for Graph labels - should we find a better way?
```
karupayun, line 75: We should create a bug for that

src/OrbitGl/Viewport.h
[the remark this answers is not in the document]
karupayun: I like it! But if it isn't trivial, don't rush in making it for this PR.

src/OrbitGl/Viewport.cpp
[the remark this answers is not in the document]
karupayun: I see your point now (and actually I think I modified it in the past when I made the vertical zoom and thought the same). My plan seems to be clearly more dirty and bug-prone. Thanks for the explanation :)

src/OrbitGl/ViewportTest.cpp
```diff
@@ -0,0 +1,176 @@
...
+
+  EXPECT_EQ(viewport.GetWorldTopLeft(), Vec2(200.f, -400.f));
+  EXPECT_EQ(viewport.IsDirty(), true);
+  viewport.ClearDirtyFlag();
+
+  // Scrolling further after clamping should not result in a dirty flag
+  viewport.SetWorldTopLeftX(350.f);
+  viewport.SetWorldTopLeftY(-600.f);
+  EXPECT_EQ(viewport.GetWorldTopLeft(), Vec2(200.f, -400.f));
+  EXPECT_EQ(viewport.IsDirty(), false);
```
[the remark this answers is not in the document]
reichlfl: I would leave it in - it tests the same behavior, but the preconditions are different, and this test makes sure that, when zooming is present, the clamping still works as expected.

src/OrbitGl/ViewportTest.cpp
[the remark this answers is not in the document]
reichlfl: I renamed all methods:

VisibleWorld[Width/Height]: Size of the world visible on the screen
WorldExtents: Max values of the current world
WorldMin: Min values of the current world (*technically*, min and max are swapped for y because all values we use are negative, but conceptually, I think I'll stick with this notion of min / max)
ScreenSize: Size of the screen in pixels

I added some comments above each method block, hope that makes the usage clearer.

src/OrbitGl/Viewport.h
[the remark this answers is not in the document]
reichlfl: Done, see comment below.

src/OrbitGl/Viewport.cpp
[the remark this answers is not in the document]
reichlfl: I've added a comment and a brief explanation in the file header.

src/OrbitGl/Viewport.h
```diff
@@ -0,0 +1,85 @@
...
+  [[nodiscard]] int WorldToScreenWidth(float width) const;
+
+  [[nodiscard]] Vec2i QtToGlScreenPos(const Vec2i& qt_pos) const;
+
+  [[nodiscard]] bool IsDirty() const { return is_dirty_; }
+  void FlagAsDirty() { is_dirty_ = true; }
+  void ClearDirtyFlag() { is_dirty_ = false; }
+
+ private:
+  int width_;
```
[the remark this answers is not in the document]
reichlfl: Good idea, done, thanks!

src/OrbitGl/ViewportTest.cpp
reichlfl: Done

src/OrbitGl/ViewportTest.cpp
reichlfl: It should be covered by the coordinate conversion test. I've extended that one to include a zoomed and non-zoomed version of all tests.

src/OrbitGl/CaptureWindow.cpp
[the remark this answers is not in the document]
reichlfl: Done

src/OrbitGl/AccessibleThreadBar.cpp
```diff
@@ -46,9 +46,10 @@ orbit_accessibility::AccessibilityRect orbit_gl::AccessibleThreadBar::Accessible
   CHECK(canvas != nullptr);
 
   return orbit_accessibility::AccessibilityRect(
-      canvas->WorldToScreenWidth(local_pos[0]), canvas->WorldToScreenHeight(local_pos[1]),
-      canvas->WorldToScreenWidth(thread_bar_->GetSize()[0]),
-      canvas->WorldToScreenHeight(thread_bar_->GetSize()[1]));
+      canvas->GetViewport().WorldToScreenWidth(local_pos[0]),
+      canvas->GetViewport().WorldToScreenHeight(local_pos[1]),
+      canvas->GetViewport().WorldToScreenWidth(thread_bar_->GetSize()[0]),
+      canvas->GetViewport().WorldToScreenHeight(thread_bar_->GetSize()[1]));
```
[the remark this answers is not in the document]
reichlfl: I've done that where appropriate now. At some point, this can probably disappear anyways because we should try to pass the Viewport instead of GlCanvas where possible.

src/OrbitGl/AccessibleTimeGraph.cpp
```diff
@@ -18,7 +18,8 @@ using orbit_accessibility::AccessibleInterface;
 
 AccessibilityRect TimeGraphAccessibility::AccessibleLocalRect() const {
   GlCanvas* canvas = time_graph_->GetCanvas();
-  return AccessibilityRect(0, 0, canvas->GetWidth(), canvas->GetHeight());
+  return AccessibilityRect(0, 0, canvas->GetViewport().GetWidth(),
+                           canvas->GetViewport().GetHeight());
```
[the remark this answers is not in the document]
reichlfl: Done

src/OrbitGl/AccessibleTrack.cpp
```diff
@@ -106,10 +106,10 @@ AccessibilityRect AccessibleTrackContent::AccessibleLocalRect() const {
...
   GlCanvas* canvas = track_->GetCanvas();
 
-  return AccessibilityRect(canvas->WorldToScreenWidth(0),
-                           canvas->WorldToScreenHeight(layout_->GetTrackTabHeight()),
-                           canvas->WorldToScreenWidth(track_->GetSize()[0]),
-                           canvas->WorldToScreenHeight(track_->GetSize()[1]));
+  return AccessibilityRect(canvas->GetViewport().WorldToScreenWidth(0),
+                           canvas->GetViewport().WorldToScreenHeight(layout_->GetTrackTabHeight()),
+                           canvas->GetViewport().WorldToScreenWidth(track_->GetSize()[0]),
+                           canvas->GetViewport().WorldToScreenHeight(track_->GetSize()[1]));
```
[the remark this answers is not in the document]
reichlfl: Done

src/OrbitGl/AccessibleTrack.cpp
```diff
@@ -136,10 +136,10 @@ AccessibilityRect AccessibleTrackTab::AccessibleLocalRect() const {
...
   GlCanvas* canvas = track_->GetCanvas();
 
-  return AccessibilityRect(canvas->WorldToScreenWidth(layout_->GetTrackTabOffset()),
-                           canvas->WorldToScreenHeight(0),
-                           canvas->WorldToScreenWidth(layout_->GetTrackTabWidth()),
-                           canvas->WorldToScreenHeight(layout_->GetTrackTabHeight()));
+  return AccessibilityRect(canvas->GetViewport().WorldToScreenWidth(layout_->GetTrackTabOffset()),
+                           canvas->GetViewport().WorldToScreenHeight(0),
+                           canvas->GetViewport().WorldToScreenWidth(layout_->GetTrackTabWidth()),
+                           canvas->GetViewport().WorldToScreenHeight(layout_->GetTrackTabHeight()));
```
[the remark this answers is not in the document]
reichlfl: Done

src/OrbitGl/AccessibleTrack.cpp
```diff
@@ -175,17 +175,18 @@ AccessibilityRect AccessibleTrack::AccessibleLocalRect() const {
   Vec2 size = track_->GetSize();
   size[1] += layout_->GetTrackTabHeight();
 
-  Vec2 screen_pos = canvas->WorldToScreen(pos);
-  int screen_width = canvas->WorldToScreenWidth(size[0]);
-  int screen_height = canvas->WorldToScreenHeight(size[1]);
+  Vec2i screen_pos = canvas->GetViewport().WorldToScreenPos(pos);
+  int screen_width = canvas->GetViewport().WorldToScreenWidth(size[0]);
+  int screen_height = canvas->GetViewport().WorldToScreenHeight(size[1]);
```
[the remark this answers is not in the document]
reichlfl: Done

src/OrbitGl/AccessibleTriangleToggle.cpp
```diff
@@ -17,16 +17,16 @@ orbit_accessibility::AccessibilityRect AccessibleTriangleToggle::AccessibleLocal
   Vec2 pos = triangle_toggle_->GetPos();
   Vec2 size = triangle_toggle_->GetSize();
 
-  Vec2 screen_pos = canvas->WorldToScreen(pos);
-  int screen_width = canvas->WorldToScreenWidth(size[0]);
-  int screen_height = canvas->WorldToScreenHeight(size[1]);
+  Vec2i screen_pos = canvas->GetViewport().WorldToScreenPos(pos);
+  int screen_width = canvas->GetViewport().WorldToScreenWidth(size[0]);
+  int screen_height = canvas->GetViewport().WorldToScreenHeight(size[1]);
```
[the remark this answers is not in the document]
reichlfl: Done

src/OrbitGl/TextRenderer.cpp
```diff
@@ -297,13 +297,16 @@ void TextRenderer::AddText(const char* text, float x, float y, float z, const Co
...
+        Vec2i(static_cast<int>(out_screen_pos.x), static_cast<int>(inv_y)));
   }
 
   if (out_text_size) {
-    (*out_text_size)[0] = canvas_->ScreenToWorldWidth(static_cast<int>(out_screen_size.x));
-    (*out_text_size)[1] = canvas_->ScreenToWorldHeight(static_cast<int>(out_screen_size.y));
+    (*out_text_size)[0] =
+        canvas_->GetViewport().ScreenToWorldWidth(static_cast<int>(out_screen_size.x));
+    (*out_text_size)[1] =
+        canvas_->GetViewport().ScreenToWorldHeight(static_cast<int>(out_screen_size.y));
```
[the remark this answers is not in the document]
reichlfl: Done

src/OrbitGl/TextRenderer.cpp
```diff
@@ -433,17 +436,17 @@ std::vector<float> TextRenderer::GetLayers() const {
...
-  float world_min_left_y = canvas_->GetWorldTopLeftY() - world_height;
+  float world_width = canvas_->GetViewport().GetWorldWidth();
+  float world_height = canvas_->GetViewport().GetWorldHeight();
+  float world_top_left_x = canvas_->GetViewport().GetWorldTopLeft()[0];
+  float world_min_left_y = canvas_->GetViewport().GetWorldTopLeft()[1] - world_height;
 
-  o_x = ((x - world_top_left_x) / world_width) * canvas_->GetWidth();
-  o_y = ((y - world_min_left_y) / world_height) * canvas_->GetHeight();
+  o_x = ((x - world_top_left_x) / world_width) * canvas_->GetViewport().GetWidth();
+  o_y = ((y - world_min_left_y) / world_height) * canvas_->GetViewport().GetHeight();
```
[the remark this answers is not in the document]
reichlfl: Done

src/OrbitGl/Viewport.h
```diff
@@ -0,0 +1,85 @@
...
+    ^
+    |
+    |
+  (0, 0) ---->
+
+Screen:
+  (0, 0) ---->
+    |
+    |
+    v
```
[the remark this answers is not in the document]
reichlfl: Done

src/OrbitGl/Viewport.h
```diff
@@ -0,0 +1,85 @@
...
+
+where world(0, 0) is initially anchored at screen(0, 0), i.e. top left (meaning basically all world
+y coordinates are negative). Viewport clamps all X values to be >= 0, and all Y values to be <= 0.
+
+See TODO(b/177350599): Unify QtScreen and GlScreen
+**/
+class Viewport {
+ public:
+  explicit Viewport(int width, int height);
```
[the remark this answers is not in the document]
reichlfl: Yes, good point, done!

src/OrbitGl/GlCanvas.cpp
reichlfl: Thanks for catching this, fixed now!

src/OrbitQt/orbitglwidget.cpp
```diff
@@ -190,7 +190,10 @@ void OrbitGLWidget::messageLogged(const QOpenGLDebugMessage& msg) {
 void OrbitGLWidget::resizeGL(int w, int h) {
   if (gl_canvas_) {
     gl_canvas_->Resize(w, h);
-    gl_canvas_->SetMainWindowSize(this->geometry().width(), this->geometry().height());
+    // TODO: I've added this to make sure removing the "MainWindowSize" won't break anything
```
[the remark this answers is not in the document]
reichlfl: Removed

src/OrbitGl/Viewport.h
reichlfl: I've changed it to //

src/OrbitGl/Viewport.h
reichlfl: Yeah, true, I hope it's now clearer from the comments and the renaming of the methods.

src/OrbitGl/Viewport.h
reichlfl: I thought about that too for the sake of API consistency, but these are called individually quite often (e.g. when manipulating through the scrollbars).

src/OrbitGl/Viewport.h
reichlfl: Done.

src/OrbitGl/Viewport.h
reichlfl: Added a comment here. There's a difference between Pos and Width / Height, that's why we need both methods.

src/OrbitGl/Viewport.h
reichlfl: Done

src/OrbitGl/ViewportTest.cpp
reichlfl: Done

src/OrbitGl/ViewportTest.cpp
reichlfl: Makes sense, thanks. Added.

src/OrbitGl/Viewport.h
reichlfl: Done

src/OrbitGl/Viewport.cpp
reichlfl: And extents_, actually :-) Changed.

src/OrbitGl/Viewport.cpp
reichlfl: I prefer to have them in the cpp - I assume this class will be included quite a lot at some point, and we already suffer from compilation times whenever we change anything in GlCanvas.h, so I'd like to keep the potential for header changes minimal here.

src/OrbitGl/Viewport.cpp
reichlfl: Done

src/OrbitGl/CaptureWindow.cpp
reichlfl: Done

src/OrbitGl/GlCanvas.cpp
reichlfl: Hm, seems that clang-format messed that up, thanks for catching.

src/OrbitGl/GlCanvas.cpp
```diff
@@ -275,26 +262,14 @@ void GlCanvas::KeyPressed(unsigned int key_code, bool ctrl, bool shift, bool alt
 void GlCanvas::KeyReleased(unsigned int key_code, bool ctrl, bool shift, bool alt) {
   UpdateSpecialKeys(ctrl, shift, alt);
   Orbit_ImGui_KeyCallback(imgui_context_, static_cast<int>(key_code), false, ctrl, shift, alt);
+
```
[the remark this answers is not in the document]
reichlfl: Done

src/OrbitGl/GlCanvas.h
```diff
@@ -95,14 +88,13 @@ class GlCanvas {
   [[nodiscard]] float GetWorldMaxY() const { return world_max_y_; }
   [[nodiscard]] float GetWorldTopLeftX() const { return world_top_left_x_; }
   [[nodiscard]] float GetWorldTopLeftY() const { return world_top_left_y_; }
+
   void UpdateWorldTopLeftY() { UpdateWorldTopLeftY(GetWorldTopLeftY()); }
   virtual void UpdateWorldTopLeftY(float val) { world_top_left_y_ = val; }
+  virtual void UpdateWorldTopLeftX(float val) { world_top_left_x_ = val; }
```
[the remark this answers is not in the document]
reichlfl: I think this is a perfectly valid complaint :D This commit is outdated, but I adjusted it accordingly in Viewport.h.

src/OrbitGl/GlCanvas.h
reichlfl: I've actually removed this, the method now exposes the last mouse pos (x and y), that makes it a bit more generically useful.

src/OrbitGl/ThreadTrack.cpp
reichlfl: Done

reichlfl:
Thanks for the thorough reviews, @pierricgimmig @karupayun @florian-kuebler. I've addressed the comments in follow-up commits, PTAL.

Pierric: I did not yet find the reason for the weird scrollbar jumping, but I'll investigate further before merging this. I'll update as soon as I find something.

pierricgimmig (approved):
Cool, thanks for the changes @reichlfl , LGTM, cheers!

src/OrbitGl/CaptureWindow.cpp
reichlfl: This was moved into `CaptureWindow::Draw` now and the comment was removed.

reichlfl:
> * Right click used to give Orbit focus, now if Orbit is not in focus and we try to right-click and drag, the green rectangle does not appear

@pierricgimmig On windows, this works as expected... sure this is related to the changes here? I did not touch anything above OrbitGl, and can't see how those chances should influence window focus as all of this handling is done in OrbitQt.

> * Vertical scrolling in a single direction makes the vertical slider jump up and down (see video)

@pierricgimmig I've found the reason and fixed it. This was actually a one-line change, I moved the update of Viewport Extents to be done right after track rendering.

src/OrbitGl/ViewportTest.cpp
```diff
@@ -151,20 +164,26 @@ void VerifyConversion(Viewport& viewport, const Vec2i& screen_pos, const Vec2& w
...
 
-  // NOTE: World space is y-down, but (0,0) is top left
   Vec2i screen_pos = Vec2i(8, 20);
-  Vec2 world_pos = Vec2(16.f, -10.f);
-  Vec2 world_size = Vec2(16.f, 10.f);
+  Vec2 world_pos = Vec2(8.f, -20.f);
+  Vec2 world_size = Vec2(8.f, 20.f);
+  VerifyConversion(viewport, screen_pos, world_pos, world_size);
+
+  // Change zoom: Zoom out 100% horizontally, zoom in 50% vertically.
```
karupayun, line 180: 200% horizontally

karupayun (approved):
Only a small nit, otherwise LGTM!!! :)

Thanks for this

src/OrbitGl/ViewportTest.cpp
reichlfl: *By* 100% was what I wanted to say here. But this is not clear, I'll change the comment, thanks!

karupayun: I've just seen "zoom-out". Still prefers 200% and 50%.

reichlfl: Changed

PR #2145, opened 2021-04-01
Add ExpandTrack/CollapseTrack E2E-Tests

florian-kuebler:
Test: Run E2E Test locally.

### DIFF
--- a/contrib/automation_tests/test_cases/capture_window.py
+++ b/contrib/automation_tests/test_cases/capture_window.py
@@ -123,7 +123,7 @@ class MoveTrack(CaptureWindowE2ETestCaseBase):
 
 class MatchTracksBase(CaptureWindowE2ETestCaseBase):
     """
-    Provides a static method track names.
+    Provides a static method to match track names.
     """
     @staticmethod
     def _match(expected_name: str or Iterable[str], found_name: str) -> bool:
--- a/contrib/automation_tests/test_cases/capture_window.py
+++ b/contrib/automation_tests/test_cases/capture_window.py
@@ -121,7 +121,22 @@ class MoveTrack(CaptureWindowE2ETestCaseBase):
                        (expected_new_index, index))
 
 
-class MatchTracks(CaptureWindowE2ETestCaseBase):
+class MatchTracksBase(CaptureWindowE2ETestCaseBase):
+    """
+    Provides a static method track names.
+    """
+    @staticmethod
+    def _match(expected_name: str or Iterable[str], found_name: str) -> bool:
+        if isinstance(expected_name, str):
+            return fnmatch(found_name, expected_name)
+        else:
+            for option in expected_name:
+                if fnmatch(found_name, option):
+                    return True
+        return False
+
+
+class MatchTracks(MatchTracksBase):
     """
     Verify that the existing visible tracks match the expected tracks
     """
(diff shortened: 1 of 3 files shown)


contrib/automation_tests/test_cases/capture_window.py
```diff
@@ -167,15 +182,58 @@ def _execute(self, expected_count: int = None, expected_names: List[str or Itera
...
-            return fnmatch(found_name, expected_name)
-        else:
-            for option in expected_name:
-                if fnmatch(found_name, option):
-                    return True
-        return False
+
+class CollapsingTrackBase(MatchTracksBase):
+    """
+    Clicks on a track's triangle toggle and calls compares the track's height by calling `_verify_height`.
```
danielfenner, line 188: ~~calls~~

danielfenner (approved):
lgtm

contrib/automation_tests/test_cases/capture_window.py
```diff
@@ -167,15 +182,58 @@ def _execute(self, expected_count: int = None, expected_names: List[str or Itera
...
+        :param expected_name: The exact name of the track to be collapsed. "*" is allowed as placeholder.
+        """
+        assert (expected_name is not "")
+
+        tracks = self._find_tracks()
+
+        found_track = False
+        for track in [Track(t) for t in tracks]:
+            track_name = track.name
+            if self._match(expected_name, track_name):
```
reichlfl, line 202: `self._find_tracks()` has a filter parameter that you can use to only return tracks matching a certain pattern, equal to what's checked in `self._match`. I'd recomment to use this and remove the inheritance from `MatchTracksBase`.

contrib/automation_tests/test_cases/capture_window.py
```diff
@@ -167,15 +182,58 @@ def _execute(self, expected_count: int = None, expected_names: List[str or Itera
...
-    def _match(expected_name: str or Iterable[str], found_name: str) -> bool:
-        if isinstance(expected_name, str):
-            return fnmatch(found_name, expected_name)
-        else:
-            for option in expected_name:
-                if fnmatch(found_name, option):
-                    return True
-        return False
+
+class CollapsingTrackBase(MatchTracksBase):
```
reichlfl, line 186: I don't think this is a good use of inheritance. As far as I can see you only want the `_match` method?  Probably you don't need that, see my comment below, but even if you do, I'd suggest to either make this a free function or introduce it in the `CaptureWindowE2ETestCaseBase`.

contrib/automation_tests/test_cases/capture_window.py
```diff
@@ -167,15 +182,58 @@ def _execute(self, expected_count: int = None, expected_names: List[str or Itera
...
+                triangle_toggle.click_input(button_up=False)
+                time.sleep(0.1)
+                triangle_toggle.click_input(button_down=False)
+                post_height = track.container.rectangle().height()
+                self._verify_height(prev_height, post_height, expected_name)
+                break
+
+        self.expect_true(found_track, "Found a match for track name '%s'" % expected_name)
+
+    def _verify_height(self, prev_click_height, post_click_height, track_name):
```
reichlfl, line 217: I'd advocate for type hints everywhere: `def _verify_height(self, prev_click_height: int, post_click_height: int, track_name: str)`

contrib/automation_tests/test_cases/capture_window.py
florian-kuebler: Ah thanks. Done. I also deleted the MatchTracksBase class.

contrib/automation_tests/test_cases/capture_window.py
florian-kuebler: Done.

contrib/automation_tests/test_cases/capture_window.py
florian-kuebler: Agree. I wasn't sure about the best way to do so in python. However, this is obsolete with using the find_tracks method with a filter.

contrib/automation_tests/test_cases/capture_window.py
florian-kuebler: Done. Thanks.

reichlfl (commented):
Thanks for the changes, LGTM!

reichlfl (approved):

PR #2147, opened 2021-04-01
Enable saving memory tracks in capture files

vickyliu-go4it:
With this change, we enable saving the memory tracks in the capture file.

Bug: http://b/183172079
Test: Take a capture with enabling memory tracing and save the capture file, then open the saved capture file.
> - Taking a capture with enabling memory tracing
> Taking a capture with enabling memory tracing
> - Open the capture file with memory tracks
> Loading the capture file with memory tracks

### DIFF
--- a/src/OrbitGl/App.cpp
+++ b/src/OrbitGl/App.cpp
@@ -291,21 +291,24 @@ void OrbitApp::OnSystemMemoryUsage(
   timer_info.set_start(system_memory_usage.timestamp_ns());
   timer_info.set_end(system_memory_usage.timestamp_ns());
 
-  constexpr size_t kNumOfEncodedValues = 5;
-  timer_info.mutable_registers()->Reserve(kNumOfEncodedValues);
-  timer_info.mutable_registers()->Set(static_cast<size_t>(SystemMemoryUsageEncodingIndex::kTotalKb),
-                                      orbit_api::Encode<uint64_t>(system_memory_usage.total_kb()));
-  timer_info.mutable_registers()->Set(static_cast<size_t>(SystemMemoryUsageEncodingIndex::kFreeKb),
-                                      orbit_api::Encode<uint64_t>(system_memory_usage.free_kb()));
-  timer_info.mutable_registers()->Set(
-      static_cast<size_t>(SystemMemoryUsageEncodingIndex::kAvailableKb),
-      orbit_api::Encode<uint64_t>(system_memory_usage.available_kb()));
-  timer_info.mutable_registers()->Set(
-      static_cast<size_t>(SystemMemoryUsageEncodingIndex::kBuffersKb),
-      orbit_api::Encode<uint64_t>(system_memory_usage.buffers_kb()));
-  timer_info.mutable_registers()->Set(
-      static_cast<size_t>(SystemMemoryUsageEncodingIndex::kCachedKb),
-      orbit_api::Encode<uint64_t>(system_memory_usage.cached_kb()));
+  // The RepeatedField<Element>::Set(int index, const Element & value) method won't update the
+  // current_size_ of the repeated field. If the current_size_ == 0, the repeated field will be
+  // ignored when copying the protobuf. Therefore, if encoding the following memory information
+  // into TimerInfo registers with the Set method, the registers field of the copied TimerInfo
+  // would be empty.
+  std::vector<uint64_t> encoded_values(static_cast<size_t>(SystemMemoryUsageEncodingIndex::kEnd));
+  encoded_values[static_cast<size_t>(SystemMemoryUsageEncodingIndex::kTotalKb)] =
+      orbit_api::Encode<uint64_t>(system_memory_usage.total_kb());
+  encoded_values[static_cast<size_t>(SystemMemoryUsageEncodingIndex::kFreeKb)] =
+      orbit_api::Encode<uint64_t>(system_memory_usage.free_kb());
+  encoded_values[static_cast<size_t>(SystemMemoryUsageEncodingIndex::kAvailableKb)] =
+      orbit_api::Encode<uint64_t>(system_memory_usage.available_kb());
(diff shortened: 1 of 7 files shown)


src/OrbitGl/GraphTrack.h
```diff
@@ -41,6 +41,12 @@ class GraphTrack : public Track {
   void SetLabelUnitWhenEmpty(const std::string& label_unit);
   void SetValueDecimalDigitsWhenEmpty(uint8_t value_decimal_digits);
 
+  void OnTimer(const orbit_client_protos::TimerInfo& timer_info) override;
+  std::vector<std::shared_ptr<TimerChain>> GetAllChains() const override;
+  std::vector<std::shared_ptr<TimerChain>> GetAllSerializableChains() const override {
```
antonrohr, line 46: Since now all `Track` subclasses implement these 2 methods, you could think about making them pure virtual in `Track.h`

antonrohr (commented):
This is a great follow up for the memory usage feature. I tested it and it works :)

Did you look at unit tests for this? I know we do have CaptureSerializerTest.cpp and CaptureDeserializerTest.cpp, but I am not too familiar on what they test. Maybe adding something there for memory usage is a good idea.

src/OrbitGl/App.cpp
```diff
@@ -291,21 +291,28 @@ void OrbitApp::OnSystemMemoryUsage(
...
-      orbit_api::Encode<uint64_t>(system_memory_usage.available_kb()));
-  timer_info.mutable_registers()->Set(
-      static_cast<size_t>(SystemMemoryUsageEncodingIndex::kBuffersKb),
-      orbit_api::Encode<uint64_t>(system_memory_usage.buffers_kb()));
-  timer_info.mutable_registers()->Set(
-      static_cast<size_t>(SystemMemoryUsageEncodingIndex::kCachedKb),
-      orbit_api::Encode<uint64_t>(system_memory_usage.cached_kb()));
+  for (uint64_t value : encoded_values) {
+    timer_info.mutable_registers()->Add(value);
+  }
```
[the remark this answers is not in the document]
karupayun: or also using the insert function

src/OrbitGl/App.cpp
```diff
@@ -291,21 +291,28 @@ void OrbitApp::OnSystemMemoryUsage(
   timer_info.set_start(system_memory_usage.timestamp_ns());
   timer_info.set_end(system_memory_usage.timestamp_ns());
 
+  // The RepeatedField<Element>::Set(int index, const Element & value) method won't update the
+  // current_size_ of the repeated field. If the current_size_ == 0, the repeated field will be
+  // ignored when copying the protobuf. Therefore, if encoding the following memory information
+  // into TimerInfo registers with the Set method, the registers field of the copied TimerInfo
+  // would be empty.
   constexpr size_t kNumOfEncodedValues = 5;
```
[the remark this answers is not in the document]
karupayun: I like the map here, is it hard to change it?

Also, we can skip the issue of using static_cast<> 5 times.

src/OrbitGl/GraphTrack.h
karupayun: Don't forget the lovely FrameTrack!

src/OrbitGl/GraphTrack.h
```diff
@@ -41,6 +41,12 @@ class GraphTrack : public Track {
...
+    return GetAllChains();
```
karupayun, line 47: I'm not sure why we have both GetAllSerializableChains and GetAllChains

karupayun (approved):
LGTM Tingwei! :)

src/OrbitGl/App.cpp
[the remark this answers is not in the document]
vickyliu-go4it: Thanks for the suggestions! I put the `SystemMemoryUsageEncodingIndex::kEnd` based solution in the first fixup and the `std::map<SystemMemoryUsageEncodingIndex, unit64_t> encoded_values` based solution in the second fixup. If we all vote for the map solution, I will take that one!

src/OrbitGl/App.cpp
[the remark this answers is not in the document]
vickyliu-go4it: Changed accordingly. Thanks!

src/OrbitGl/GraphTrack.h
vickyliu-go4it: Thanks for the suggestions! Changed accordingly.

src/OrbitGl/GraphTrack.h
vickyliu-go4it: To summarize our offline discussion:
- `GetAllChains` returns all the timer chains in the track while `GetAllSerializableChains` returns the timer chains we would like to save in the capture file;
- For most types of tracks, `GetAllSerializableChains` just return the result of `GetAllChains`, but in the `AsyncTrack`, `GetAllSerializableChains` returns an empty vector to avoid serializing the async timer twice (the timers in the track already serialized on their initial thread tracks).

vickyliu-go4it:
> This is a great follow up for the memory usage feature. I tested it and it works :)
> 
> Did you look at unit tests for this? I know we do have CaptureSerializerTest.cpp and CaptureDeserializerTest.cpp, but I am not too familiar on what they test. Maybe adding something there for memory usage is a good idea.

Thanks for the suggestion! Totally agree that it's a good idea to test the saving and loading of memory tracks.

I take a look at the unit tests in CaptureSerializerTest.cpp and CaptureDeserializerTest.cpp. There are well designed tests for methods in CaptureSerializer.h and CaptureDeserializer.h, e.g., methods that generate and load the `CaptureInfo`. But there is no test for saving and loading tracks. I have created a bug http://b/184641371 for this.

antonrohr (approved):
lgtm, I would only change `encoded_values` back to vector. Like discussed offline, the change to map was not meant for `encoded_values`, but instead of using `TimerInfo.registers()`.

PR #2151, opened 2021-04-06
Lengthen sleep in LinuxTracingIntegrationTestPuppet's SleepRepeatedly

dpallotti:
The previous value of 10 microseconds seems to not be enough to cause a
consistent number of thread state slices. See also http://b/183229462#comment4.

Bug: http://b/183229462

Test: Run LinuxTracingIntegrationTest.ThreadStateSlices a few thousand times
to see that the flakiness is gone...

### DIFF
--- a/src/LinuxTracing/LinuxTracingIntegrationTestPuppet.cpp
+++ b/src/LinuxTracing/LinuxTracingIntegrationTestPuppet.cpp
@@ -27,7 +27,7 @@ using PuppetConstants = LinuxTracingIntegrationTestPuppetConstants;
 
 static void SleepRepeatedly() {
   for (uint64_t i = 0; i < PuppetConstants::kSleepCount; ++i) {
-    absl::SleepFor(absl::Microseconds(10));
+    absl::SleepFor(absl::Microseconds(100));
   }
 }
 


florian-kuebler (approved):
The change itself LGTM. However, I'd be curious if we could actually address the root problem. Do you have an idea, why we are having issues with 10micros?

dpallotti:
> However, I'd be curious if we could actually address the root problem. Do you have an idea, why we are having issues with 10micros?

I've spent most of the day trying various things (I briefly mentioned them in the bug) and my current conclusion is that this doesn't depend on our tracing code.
My current only guess is that the closer we are to the order of magnitude of the time that a context switch takes, the more likely it is that the kernel will do something different than switching the thread to interruptible sleep. After all, I don't think it's documented anywhere that a sleep will actually make the kernel switch the thread to interruptible sleep.
Of course I might be completely wrong and I might be missing something obvious. But I also don't really think that for the time being it's worth investing more time in investigating this.

src/LinuxTracing/LinuxTracingIntegrationTestPuppet.cpp
```diff
@@ -27,7 +27,7 @@ using PuppetConstants = LinuxTracingIntegrationTestPuppetConstants;
 
 static void SleepRepeatedly() {
   for (uint64_t i = 0; i < PuppetConstants::kSleepCount; ++i) {
-    absl::SleepFor(absl::Microseconds(10));
+    absl::SleepFor(absl::Microseconds(100));
```
dimitry-, line 30: There are multiple layers here:
1. absl
2. glibc
3. kernel

I think we might want to call nanosleep here for the starters and see if it returns EINTR, and it case it does repeat. It could be that abseil here is doing smart think by checking if time is already passed when EINTR is returned and return with noop. and 10 microseconds could be just at the limit here.

worth a try?

dimitry-: (and probably add a comment about why the test uses nanosleep instead of absl)

dpallotti: > I think we might want to call nanosleep here for the starters and see if it returns EINTR, and it case it does repeat.

This made sense to me but doesn't work. Even when the test fails I don't see `EINTR` ever being returned. After all, why would we expect the target to receive signals?

By the way, what Abseil does is here: https://chromium.googlesource.com/external/github.com/abseil/abseil-cpp/+/HEAD/absl/time/clock.cc#564

dimitry-: ok, nevermind. I was under mistaken impression that nanosleep syscall operates in duration - but it actually uses timespec. which means that increasing timeout seems to be the only way here.

dimitry- (approved):

florian-kuebler:
> > However, I'd be curious if we could actually address the root problem. Do you have an idea, why we are having issues with 10micros?
> 
> I've spent most of the day trying various things (I briefly mentioned them in the bug) and my current conclusion is that this doesn't depend on our tracing code.
> My current only guess is that the closer we are to the order of magnitude of the time that a context switch takes, the more likely it is that the kernel will do something different than switching the thread to interruptible sleep. After all, I don't think it's documented anywhere that a sleep will actually make the kernel switch the thread to interruptible sleep.
> Of course I might be completely wrong and I might be missing something obvious. But I also don't really think that for the time being it's worth investing more time in investigating this.

Okay, thanks for the explanation. Let's hope that it does not bubble up again. And if so, investigate once more.

PR #2152, opened 2021-04-06
Stable track order when adding/removing FrameTrack

karupayun:
This PR make the track order stable after both adding and removing a
FrameTrack. As discussed in a daily meeting, the Frame Track is added
after the Scheduler one.

Also, we rename a method (UpdateVisibleTrackList).

Solved the quick-win http://b/175534459.

Test: Load a capture, add/remove a FrameTrack. Also start/stop a
capture.

### DIFF
--- a/src/OrbitGl/TrackManager.cpp
+++ b/src/OrbitGl/TrackManager.cpp
@@ -142,17 +142,17 @@ void TrackManager::SortTracks() {
 
     last_thread_reorder_.Restart();
 
-    UpdateFilteredTrackList();
+    UpdateVisibleTrackList();
   }
   sorting_invalidated_ = false;
 }
 
 void TrackManager::SetFilter(const std::string& filter) {
   filter_ = absl::AsciiStrToLower(filter);
-  UpdateFilteredTrackList();
+  UpdateVisibleTrackList();
 }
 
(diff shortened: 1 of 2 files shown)


src/OrbitGl/TrackManager.cpp
```diff
@@ -306,12 +306,23 @@ void TrackManager::AddTrack(const std::shared_ptr<Track>& track) {
   sorting_invalidated_ = true;
 }
 
+void TrackManager::AddFrameTrack(const std::shared_ptr<FrameTrack>& frame_track) {
+  frame_tracks_[frame_track->GetFunctionId()] = frame_track;
+  auto scheduler_track_pos =
+      find(sorted_tracks_.begin(), sorted_tracks_.end(), scheduler_track_.get());
+  if (scheduler_track_pos != sorted_tracks_.end()) {
+    sorted_tracks_.insert(++scheduler_track_pos, frame_track.get());
```
dimitry-, line 314: maybe rename this to ordered_tracks_, I was initially under impression this is std::sorted (in which case we would have sorted this by pointers) but it is not. (not for this cl)

dimitry- (approved):
LGTM thanks!

ronaldfw (approved):
LGTM, thanks!

PR #2153, opened 2021-04-06
Add Track::IsFilterable() method to allow non-filterable tracks

pierricgimmig:
Scheduler track is now always visible (b/181671054). 
"Pinned" tracks are also not be filterable.

### DIFF
--- a/src/OrbitGl/TrackManager.cpp
+++ b/src/OrbitGl/TrackManager.cpp
@@ -161,9 +161,13 @@ void TrackManager::UpdateFilteredTrackList() {
   visible_tracks_.clear();
   std::vector<std::string> filters = absl::StrSplit(filter_, ' ', absl::SkipWhitespace());
   for (const auto& track : sorted_tracks_) {
+    if (!track->IsFilterable()) {
+      visible_tracks_.push_back(track);
+      continue;
+    }
     std::string lower_case_label = absl::AsciiStrToLower(track->GetLabel());
     for (auto& filter : filters) {
-      if (!track->IsFilterable() || absl::StrContains(lower_case_label, filter)) {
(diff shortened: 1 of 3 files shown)


reichlfl (commented):
C++ part LGTM, but this will break some E2E tests that use the track filtering - at least orbit_create_frame_track, and it will make the results from orbit_tracks less meaningful.

I'd suggest to change the create_frame_track test to expect the scheduler to be visible - filter_string should already work with a list of names, so you should be able to include the scheduler there. If not, you can split the FilterTracks test to not verify the # of tracks, and use MatchTracks yourself.

For orbit_tracks, at least increase the expected # of tracks where needed.

src/OrbitGl/TrackManager.cpp
```diff
@@ -163,7 +163,7 @@ void TrackManager::UpdateFilteredTrackList() {
   for (const auto& track : sorted_tracks_) {
     std::string lower_case_label = absl::AsciiStrToLower(track->GetLabel());
     for (auto& filter : filters) {
-      if (absl::StrContains(lower_case_label, filter)) {
+      if (!track->IsFilterable() || absl::StrContains(lower_case_label, filter)) {
```
karupayun, line 166: Nit: consider moving this if to the track-loop.

karupayun (approved):
LGTM Pierric! Thanks for solving it!

src/OrbitGl/TrackManager.cpp
pierricgimmig: Oh yes, of course, thanks!

pierricgimmig:
> but this will break some E2E tests

Arrgh, I will implement your suggestions, thanks @reichlfl

PR #2155, opened 2021-04-07
Fix y-coordinate of GPU timers with command buffers.

florian-kuebler:
On a GPU track, we have a small gap (2px) between the different depths.
When we have command buffer timers, the depth is multiplied by 2.
However, as there is no gap between the hw execution timer and the
command buffer timers, the number of gaps must remain the same in both cases.

Test: Take a capture with the Vulkan layer and expand the track.
Bug: http://b/184720175

### DIFF
--- a/src/OrbitGl/GpuTrack.cpp
+++ b/src/OrbitGl/GpuTrack.cpp
@@ -172,14 +172,14 @@ float GpuTrack::GetYFromTimer(const TimerInfo& timer_info) const {
   CHECK(timer_info.type() == TimerInfo::kGpuActivity ||
         timer_info.type() == TimerInfo::kGpuCommandBuffer);
 
+  float gap_space = adjusted_depth * layout_->GetSpaceBetweenGpuDepths();
+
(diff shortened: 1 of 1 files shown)


src/OrbitGl/GpuTrack.cpp
```diff
@@ -172,15 +172,15 @@ float GpuTrack::GetYFromTimer(const TimerInfo& timer_info) const {
   CHECK(timer_info.type() == TimerInfo::kGpuActivity ||
         timer_info.type() == TimerInfo::kGpuCommandBuffer);
 
+  float gap_space = adjusted_depth * layout_->GetSpaceBetweenGpuDepths();
```
ronaldfw, line 175: Add a comment why this is done here and not later (as before).

ronaldfw (approved):
LGTM, with just one small nitpick to add a comment

src/OrbitGl/GpuTrack.cpp
florian-kuebler: Done. Thanks for the review.

PR #2156, opened 2021-04-07
Move LinuxTracingIntegrationTest to a separate module

dpallotti:
For better modularization, as we expect `LinuxTracingIntegrationTest` to still
grow and as it should only use the public interface of `LinuxTracing`.


beckerhe (approved):
Nice. LGTM!

dimitry- (approved):

PR #2157, opened 2021-04-07
Fix executing function in tracee.

danielfenner:
Previously we called the function without adjusting the stack pointer.
There are two problems with this:
a) The 128 bytes below the stack pointer can be used for temporary
variables (without moving rsp). The function we interrupt and the
function we execute might interfer here.
b) rsp needs to be a multiple of 16 when calling a function (but not
during the execution of the function we interrupt).

Now we adjust rsp accordingly before executing the function in the
tracee. Note that all registers get restored before the regular
execution of the tracee continues - so rsp is restored as well.

Bug: http://b/182908805
Test: The previously flaky unit test works fine now.

### DIFF
--- a/src/UserSpaceInstrumentation/InjectLibraryInTracee.cpp
+++ b/src/UserSpaceInstrumentation/InjectLibraryInTracee.cpp
@@ -81,9 +81,18 @@ ErrorMessageOr<uint64_t> ExecuteOrDie(pid_t pid, uint64_t address_code, uint64_t
   RegisterState original_registers;
   OUTCOME_TRY(original_registers.BackupRegisters(pid));
 
-  RegisterState registers_set_rip = original_registers;
-  registers_set_rip.GetGeneralPurposeRegisters()->x86_64.rip = address_code;
-  OUTCOME_TRY(registers_set_rip.RestoreRegisters());
+  RegisterState registers_for_execution = original_registers;
+  registers_for_execution.GetGeneralPurposeRegisters()->x86_64.rip = address_code;
+  // The calling convention for x64 assumes the 128 bytes below rsp to be usable as a scratch pad
+  // for the current function. This area is called the 'red zone'. The function we interrupted might
+  // have stored temporary data in the red zone and the function we are about to execute might do
+  // the same. To keep them separated we decrement rsp by 128 bytes.
+  // The calling convention also asks for rsp being a multiple of 16 so we additionally round down
+  // to the next multiople of 16.
+  const uint64_t old_rsp = original_registers.GetGeneralPurposeRegisters()->x86_64.rsp;
+  const uint64_t aligned_rsp_below_red_zone = (old_rsp - 128) / 16 * 16;
+  registers_for_execution.GetGeneralPurposeRegisters()->x86_64.rsp = aligned_rsp_below_red_zone;
+  OUTCOME_TRY(registers_for_execution.RestoreRegisters());
   if (ptrace(PTRACE_CONT, pid, 0, 0) != 0) {
     FATAL("Unable to continue tracee with PTRACE_CONT.");
   }


src/UserSpaceInstrumentation/InjectLibraryInTracee.cpp
```diff
@@ -81,9 +81,18 @@ ErrorMessageOr<uint64_t> ExecuteOrDie(pid_t pid, uint64_t address_code, uint64_t
   RegisterState original_registers;
   OUTCOME_TRY(original_registers.BackupRegisters(pid));
 
-  RegisterState registers_set_rip = original_registers;
-  registers_set_rip.GetGeneralPurposeRegisters()->x86_64.rip = address_code;
-  OUTCOME_TRY(registers_set_rip.RestoreRegisters());
+  RegisterState registers_for_execution = original_registers;
+  registers_for_execution.GetGeneralPurposeRegisters()->x86_64.rip = address_code;
+  // The calling convention for x64 assumes the 128 bytes below rsp to be usable as a scratch pad
```
pierricgimmig, line 86: nit. "128 bytes below rsp" might be confusing given that stack addresses decrease as we push. Consider changing to "128 bytes below rsp" or maybe use something like "The calling convention for x64 assumes a scratch pad of 128 bytes for the current function.

pierricgimmig (approved):
Hi @danielfenner, thanks for looking into this. I'm not sure the issue is completely resolved unfortunately. It doesn't seem to fix the segmentation fault when injecting code into OrbitTest.

src/UserSpaceInstrumentation/InjectLibraryInTracee.cpp
pierricgimmig: I guess it depends on when you look at rsp, I'd say that if it's from the perspective of the called function, then "above" should be used. From the caller, "below" is fine. Feel free to disregard.

src/UserSpaceInstrumentation/InjectLibraryInTracee.cpp
```diff
@@ -81,9 +81,18 @@ ErrorMessageOr<uint64_t> ExecuteOrDie(pid_t pid, uint64_t address_code, uint64_t
...
+  // for the current function. This area is called the 'red zone'. The function we interrupted might
+  // have stored temporary data in the red zone and the function we are about to execute might do
+  // the same. To keep them separated we decrement rsp by 128 bytes.
+  // The calling convention also asks for rsp being a multiple of 16 so we additionally round down
+  // to the next multiople of 16.
+  const uint64_t old_rsp = original_registers.GetGeneralPurposeRegisters()->x86_64.rsp;
+  const uint64_t aligned_rsp_below_red_zone = (old_rsp - 128) / 16 * 16;
+  registers_for_execution.GetGeneralPurposeRegisters()->x86_64.rsp = aligned_rsp_below_red_zone;
```
pierricgimmig, line 94: I think doing the offset here is dangerous as any injected code that pushes to the stack will mess with the alignment of rsp and with the red zone. Should we instead do this right before the call instruction in the injected code?

src/UserSpaceInstrumentation/InjectLibraryInTracee.cpp
```diff
@@ -81,9 +81,18 @@ ErrorMessageOr<uint64_t> ExecuteOrDie(pid_t pid, uint64_t address_code, uint64_t
...
+  RegisterState registers_for_execution = original_registers;
+  registers_for_execution.GetGeneralPurposeRegisters()->x86_64.rip = address_code;
+  // The calling convention for x64 assumes the 128 bytes below rsp to be usable as a scratch pad
+  // for the current function. This area is called the 'red zone'. The function we interrupted might
+  // have stored temporary data in the red zone and the function we are about to execute might do
+  // the same. To keep them separated we decrement rsp by 128 bytes.
+  // The calling convention also asks for rsp being a multiple of 16 so we additionally round down
+  // to the next multiople of 16.
+  const uint64_t old_rsp = original_registers.GetGeneralPurposeRegisters()->x86_64.rsp;
+  const uint64_t aligned_rsp_below_red_zone = (old_rsp - 128) / 16 * 16;
```
dimitry-, line 93: this number may be less than 128 `(old_rsp - 128 - 15) & ~15`?

dimitry-: oh wait it is rounding *down*. nevermind.

dimitry- (approved):

pierricgimmig (commented):
I approved too fast, please see comments.

src/UserSpaceInstrumentation/InjectLibraryInTracee.cpp
pierricgimmig: Actually, I read a bit more about the red zone, it's not at all similar to the shadow space on Windows. If I understand correctly, we don't need to change RSP at all, the red zone is simply a guarantee that the 128 bytes below RSP "will not be asynchronously clobbered by signals or interrupt handlers, and thus can use it for scratch data, without explicitly moving the stack pointer." (https://eli.thegreenplace.net/2011/09/06/stack-frame-layout-on-x86-64#the-red-zone). 

We shouldn't modify RSP other than by making sure it is 16 bytes aligned.

src/UserSpaceInstrumentation/InjectLibraryInTracee.cpp
danielfenner: It is "below" from the perspective of the called function. Unlike what is done on Windows the function might access rsp-8, rsp-16, ... without changing rsp. Typically only leaf functions will do that.

src/UserSpaceInstrumentation/InjectLibraryInTracee.cpp
danielfenner: I could do `(old_rsp - 128) & ~15` but I think it is slightly more readable this way.

src/UserSpaceInstrumentation/InjectLibraryInTracee.cpp
danielfenner: I found that same document. It is linked in the bug, maybe I should add the link here.

>  it's not at all similar to the shadow space on Windows

Well, it is space on the stack usable for temporary variables without changing rsp. On windows it is above rsp and one can rely on the caller to allocate it. Here the area below rsp is implicitly allocated. 

>will not be asynchronously clobbered ...

But this is what we might do: Assume that the function we interrupt writes to the red zone (it might be a leaf function). The function we call might do the same. When we continue execution of the interrupted function the red zone will contain rubbish.
To avoid that I move rsp below the current red zone to a untouched part of the stack where the calling convention does not make any guarantees. When the regular execution continues rsp is restored and the interrupted function gets back its unmodified red zone.

dpallotti: > will not be asynchronously clobbered by signals or interrupt handlers

If I understand correctly the point here is exactly that we are "asynchronously clobbering" it, hence breaking the calling convention if we don't change rsp.

src/UserSpaceInstrumentation/InjectLibraryInTracee.cpp
dpallotti: What about using "after" instead? It be slightly clearer that it means in the direction in which the stack grows.

dpallotti (approved):

danielfenner:
> Hi @danielfenner, thanks for looking into this. I'm not sure the issue is completely resolved unfortunately. It doesn't seem to fix the segmentation fault when injecting code into OrbitTest.

I'll look into this today. It might very well be that there is an additional, unrelated problem when attaching to more complex code.

danielfenner:
> I'll look into this today. It might very well be that there is an additional, unrelated problem when attaching to more complex code.

I can reproduce the crash with OrbitTest.

The problem seems to be that we stop OrbitTest in the middle of a syscall every time we attach to it. This will need special handling. But it is an unrelated issue - I suggest doing this in a follow up PR.

src/UserSpaceInstrumentation/InjectLibraryInTracee.cpp
pierricgimmig: Yes, never mind, it makes sense to modify RSP before executing the injected code precisely because we don't want to clobber the existing red zone. I wonder if we are moving it enough then. The red zone is 128 bytes after the local variables. Could it be that a function has used the red zone but not yet incremented RSP for its local variables (which we could have more than 128 bytes of? In that case calling our injected code could clobber the red zone.

pierricgimmig:
> The problem seems to be that we stop OrbitTest in the middle of a syscall every time we attach to it. This will need special handling. But it is an unrelated issue - I suggest doing this in a follow up PR.

Ok sounds good, I still wonder if it's possible that a function uses the red zone before it decrements rsp for its local variables, (maybe to backup registers without decrementing RSP?). But since this seems to fix the tests, I'm ok with submitting it and iterating if need be. Thanks!

pierricgimmig (approved):
LGTM

src/UserSpaceInstrumentation/InjectLibraryInTracee.cpp
dpallotti: > Could it be that a function has used the red zone but not yet decremented RSP for its local variables (which we could have more than 128 bytes of)?

Is that a problem? The way I understand it the red zone is defined based on the current value of RSP in every moment, so if a function uses more than 128 bytes past RSP before initially decrementing RSP then the function is already breaking the calling convention.

pierricgimmig: Ah yes, this makes more sense now, thanks for the input.

danielfenner:
> I still wonder if it's possible that a function uses the red zone before it decrements rsp for its local variables, (maybe to backup registers without decrementing RSP?). 

That would be bad. But the way I read section 3.2..2 here: https://refspecs.linuxfoundation.org/elf/x86_64-abi-0.99.pdf  is that stack frame should be set up (this includes setting rsp) before anything in the function happens.

src/UserSpaceInstrumentation/InjectLibraryInTracee.cpp
danielfenner: I'll keep the "below". "after" sounds a bit like "toward larger numbers" to me.

src/UserSpaceInstrumentation/InjectLibraryInTracee.cpp
danielfenner: Okay, yes that also makes sense. 

As commented under "conversations": Section 3.2.2 here: https://refspecs.linuxfoundation.org/elf/x86_64-abi-0.99.pdf reads as if the stack frame is set up before anything else happens. But it could be made clearer there.

danielfenner:
Thanks for having a look everyone!

PR #2159, opened 2021-04-08
Add MetrcicsUploaderStub

IrinaShkviro:
With MetricsUploaderStub we will always return a valid result from
CreateMetricsUploader(): MetricsUploaderImpl or MetricsUploaderStub. It
allows us to avoid nullptr checks every time when we want to call
MetricsUploader::SendLogEvent().
The stub is returned on Linux, or in case of errors on Windows.

Bug: http://b/183177789
Test: start Orbit with and without dll.

### DIFF
--- a/src/MetricsUploader/MetricsUploaderWindowsTest.cpp
+++ b/src/MetricsUploader/MetricsUploaderWindowsTest.cpp
@@ -7,65 +7,71 @@
 #include <chrono>
 
 #include "MetricsUploader/MetricsUploader.h"
+#include "MetricsUploader/MetricsUploaderStub.h"
 #include "OrbitBase/Result.h"
 #include "absl/container/flat_hash_set.h"
 #include "absl/strings/ascii.h"
 
 namespace orbit_metrics_uploader {
 
+using orbit_metrics_uploader::MetricsUploaderStub;
+
+[[nodiscard]] static bool IsMetricsUploaderStub(
+    const std::unique_ptr<MetricsUploader>& metrics_uploader) {
+  return dynamic_cast<MetricsUploaderStub*>(metrics_uploader.get()) != nullptr;
+}
+
 TEST(MetricsUploader, CreateMetricsUploaderFromClientWithoutSendEvent) {
   auto metrics_uploader =
       MetricsUploader::CreateMetricsUploader("MetricsUploaderClientWithoutSendEvent");
-  EXPECT_FALSE(metrics_uploader.has_value());
+  EXPECT_TRUE(IsMetricsUploaderStub(metrics_uploader));
 }
 
 TEST(MetricsUploader, CreateMetricsUploaderFromClientWithoutSetup) {
   auto metrics_uploader =
       MetricsUploader::CreateMetricsUploader("MetricsUploaderClientWithoutSetup");
-  EXPECT_FALSE(metrics_uploader.has_value());
+  EXPECT_TRUE(IsMetricsUploaderStub(metrics_uploader));
 }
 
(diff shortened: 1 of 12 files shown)


src/Orbit/main.cpp
```diff
@@ -93,16 +93,9 @@ void RunUiInstance(const DeploymentConfiguration& deployment_configuration,
...
-  auto metrics_uploader_result = orbit_metrics_uploader::MetricsUploader::CreateMetricsUploader();
-  if (metrics_uploader_result.has_value()) {
-    metrics_uploader = std::move(metrics_uploader_result.value());
-    LOG("MetricsUploader was initialized successfully");
-    metrics_uploader->SendLogEvent(
-        orbit_metrics_uploader::OrbitLogEvent_LogEventType_ORBIT_METRICS_UPLOADER_START);
-  } else {
-    ERROR("%s", metrics_uploader_result.error().message());
-  }
+  auto metrics_uploader = orbit_metrics_uploader::MetricsUploader::CreateMetricsUploader();
```
antonrohr, line 96: nit: I prefer the full type here instead of `auto`, since its not clear that its going to be a unique_ptr, based on `orbit_metrics_uploader::MetricsUploader::CreateMetricsUploader()`

antonrohr (approved):
lgtm! Thank you

src/MetricsUploader/MetricsUploaderWindowsTest.cpp
```diff
@@ -7,65 +7,66 @@
...
 
 namespace orbit_metrics_uploader {
 
+using orbit_metrics_uploader::MetricsUploaderStub;
+
 TEST(MetricsUploader, CreateMetricsUploaderFromClientWithoutSendEvent) {
   auto metrics_uploader =
       MetricsUploader::CreateMetricsUploader("MetricsUploaderClientWithoutSendEvent");
-  EXPECT_FALSE(metrics_uploader.has_value());
+  EXPECT_NE(dynamic_cast<MetricsUploaderStub*>(metrics_uploader.get()), nullptr);
```
beckerhe, line 22: Nit: On first sight it's a bit hard to understand what's happening here. This would benefit from separating the check into its own function `[[nodiscard]] static bool IsMetricsUploaderStub(const std::unique_ptr<MetricsUploader>&)`

src/MetricsUploader/include/MetricsUploader/MetricsUploader.h
```diff
@@ -42,7 +42,7 @@ class MetricsUploader {
   // Create a MetricsUploader instance, load metrics uploader client library if available and starts
   // metrics uploader client when called first time. Return the instance if there are no errors and
   // ErrorMessage otherwise.
```
beckerhe, line 44: This comment needs an update

beckerhe (approved):
Thanks Irina!

src/Orbit/main.cpp
IrinaShkviro: Done, thanks!

src/MetricsUploader/MetricsUploaderWindowsTest.cpp
IrinaShkviro: Agree, I've moved this logic to the separate function in this test file, thank you!.

src/MetricsUploader/include/MetricsUploader/MetricsUploader.h
IrinaShkviro: Ah, right, fixed now

beckerhe:
FYI: The msvc2019 build failure will go away after rebasing.

PR #2160, opened 2021-04-08
Package Visual C++ runtime on Windows

beckerhe:
This will allow us to update the compiler without making changes to the
installer.

Tried on a fresh Windows VM without the Visual C++ redistributable installed. Works as expected.


danielfenner (approved):
Thanks! lgtm!

PR #2161, opened 2021-04-08
Handling of tracees interrupted in a syscall.

danielfenner:
Previously we ignored the special handling of the kernel when a
process gets interrupted during a syscall. We now inhibit that
special handling for the intermediate restarts.

Bug: http://b/184159315
Test: New unit test with tracee stuck in a syscall.

### DIFF
--- a/src/UserSpaceInstrumentation/InjectLibraryInTracee.cpp
+++ b/src/UserSpaceInstrumentation/InjectLibraryInTracee.cpp
@@ -90,8 +90,17 @@ ErrorMessageOr<uint64_t> ExecuteOrDie(pid_t pid, uint64_t address_code, uint64_t
   // The calling convention also asks for rsp being a multiple of 16 so we additionally round down
   // to the next multiople of 16.
   const uint64_t old_rsp = original_registers.GetGeneralPurposeRegisters()->x86_64.rsp;
-  const uint64_t aligned_rsp_below_red_zone = (old_rsp - 128) / 16 * 16;
+  constexpr int kSizeRedZone = 128;
+  constexpr int kStackAlignment = 16;
+  const uint64_t aligned_rsp_below_red_zone =
+      ((old_rsp - kSizeRedZone) / kStackAlignment) * kStackAlignment;
   registers_for_execution.GetGeneralPurposeRegisters()->x86_64.rsp = aligned_rsp_below_red_zone;
+  // In case we stopped the process in the middle of a syscall orig_rax holds the number of that
+  // syscall. The kernel uses that to trigger the restart of the interrupted syscall. By setting
+  // orig_rax to -1 we bypass this logic for the PTRACE_CONT below.
+  // The syscall will be restarted when we restore the original registers and detach to continue the
+  // normal operation.
+  registers_for_execution.GetGeneralPurposeRegisters()->x86_64.orig_rax = -1;
   OUTCOME_TRY(registers_for_execution.RestoreRegisters());
   if (ptrace(PTRACE_CONT, pid, 0, 0) != 0) {
     FATAL("Unable to continue tracee with PTRACE_CONT.");
--- a/src/UserSpaceInstrumentation/InjectLibraryInTraceeTest.cpp
+++ b/src/UserSpaceInstrumentation/InjectLibraryInTraceeTest.cpp
@@ -82,7 +82,10 @@ void OpenUseAndCloseLibrary(pid_t pid) {
   RegisterState registers_for_execution = original_registers;
   registers_for_execution.GetGeneralPurposeRegisters()->x86_64.rip = address_code;
   const uint64_t old_rsp = original_registers.GetGeneralPurposeRegisters()->x86_64.rsp;
-  const uint64_t aligned_rsp_below_red_zone = (old_rsp - 128) / 16 * 16;
+  constexpr int kSizeRedZone = 128;
+  constexpr int kStackAlignment = 16;
+  const uint64_t aligned_rsp_below_red_zone =
+      ((old_rsp - kSizeRedZone) / kStackAlignment) * kStackAlignment;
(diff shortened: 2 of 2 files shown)


pierricgimmig (approved):
Very cool, can't wait to try it out, LGTM. Thanks!

src/UserSpaceInstrumentation/InjectLibraryInTracee.cpp
```diff
@@ -90,8 +90,17 @@ ErrorMessageOr<uint64_t> ExecuteOrDie(pid_t pid, uint64_t address_code, uint64_t
   // The calling convention also asks for rsp being a multiple of 16 so we additionally round down
   // to the next multiople of 16.
   const uint64_t old_rsp = original_registers.GetGeneralPurposeRegisters()->x86_64.rsp;
-  const uint64_t aligned_rsp_below_red_zone = (old_rsp - 128) / 16 * 16;
+  constexpr int kSizeRedZone = 128;
+  constexpr int kStackAlignment = 16;
+  const uint64_t aligned_rsp_below_red_zone =
+      ((old_rsp - kSizeRedZone) / kStackAlignment) * kStackAlignment;
```
dimitry-, line 96: I know this actually belongs to the previous CL, but this can be tested right? Have puppet allocate something prime-number aligned on the stack and check that it doesn't crash?

dimitry-: and I still think & ~(alignment - 1) looks better.

dimitry-: maybe even introduce constexpr helper function that aligns arbitrary value to a power of 2. (and static_assert for power_of_two)

dimitry- (approved):
otherwise lgtm

src/UserSpaceInstrumentation/InjectLibraryInTracee.cpp
dpallotti: > and I still think & ~(alignment - 1) looks better

I actually find dividing and multiplying easier to read (it's encountered much more often as it works for any alignment).

dpallotti (approved):

src/UserSpaceInstrumentation/InjectLibraryInTracee.cpp
danielfenner: > I actually find dividing and multiplying easier to read

Same here but I have to admit that I saw 
`and $0xfffffff0,%esp`
in handwritten asssembler code and also emitted from the compiler in the last few days.

dimitry-: note that this is highly unusual to align something to not a power of two value. So it would be nice to have a check in align() function.

danielfenner: > , but this can be tested right?

For a meaningful test I would need something that crashes (without the enforced alignment) first.
I spend some time to produces a crash due to this but - surprisingly - didn't succeed. Even in the trivial example the 16 byte alignment is violated very often; I see rsp ending in "8" before the call and the function is happily executed (with the aligning commented out).
My understanding is that there are two reasons to enforce the alignment: performance penilties for unaligned loads and SIMD instructions that crash on unaligned access. It is surprisingly hard to get the later from within C++ - even with intrinsics the compiler tends to shuffle things around such that nothing bad happens. 
I then wrote inline assembly to cause a misaligned access on the stack:

  ` __asm__ __volatile__("movaps -7(%%rsp),%%xmm1\n\t" : : : "%rsp");`

This this does not crash either although the intel manual clearly states that "a general-protection exception (#GP) will be generated". 

I'll spend a bit more time today looking into this but currently I don't have a test for this.

src/UserSpaceInstrumentation/InjectLibraryInTraceeTest.cpp
```diff
@@ -117,9 +79,13 @@ TEST(InjectLibraryInTraceeTest, OpenUseCloseLibrary) {
   ASSERT_FALSE(result_write_code.has_error());
 
   // Set rip to the code address and execute.
```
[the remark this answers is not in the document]
danielfenner: All of this is a code duplication from ExecuteOrDie in the implementation.  I'll prepare a follow up CL to clean this up.

src/UserSpaceInstrumentation/InjectLibraryInTraceeTest.cpp
```diff
@@ -140,10 +106,76 @@ TEST(InjectLibraryInTraceeTest, OpenUseCloseLibrary) {
...
+  }
+
+  // Stop the child process using our tooling.
+  ASSERT_TRUE(AttachAndStopProcess(pid).has_value());
+
+  auto function_address_or_error = FindFunctionAddress(pid, "printf", "libc.so.6");
+  EXPECT_TRUE(function_address_or_error.has_value());
+
+  function_address_or_error = FindFunctionAddress(pid, "NOT_A_SYMBOL", "libc.so.6");
+  EXPECT_TRUE(function_address_or_error.has_error());
```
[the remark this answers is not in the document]
danielfenner: done

src/UserSpaceInstrumentation/InjectLibraryInTraceeTest.cpp
```diff
@@ -140,10 +106,76 @@ TEST(InjectLibraryInTraceeTest, OpenUseCloseLibrary) {
...
+  kill(pid, SIGKILL);
+  waitpid(pid, NULL, 0);
+}
+
+TEST(InjectLibraryInTraceeTest, OpenUseAndCloseLibraryInSyscall) {
+  pid_t pid = fork();
+  ASSERT_TRUE(pid != -1);
+  if (pid == 0) {
+    while (true) {
+      // Child wil be stuck in syscall sys_clock_nanosleep.
```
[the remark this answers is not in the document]
danielfenner: done

PR #2163, opened 2021-04-09
Remove code duplication.

danielfenner:
Cleanup from previous PR. Expose ExecuteMachineCode in header so we
don't need to duplicate it in the test.

Bug: N/A
Test: Existing unit tests.

### DIFF
--- a/src/UserSpaceInstrumentation/InjectLibraryInTracee.cpp
+++ b/src/UserSpaceInstrumentation/InjectLibraryInTracee.cpp
@@ -287,8 +287,8 @@ ErrorMessageOr<uint64_t> ExecuteMachineCode(pid_t pid, uint64_t address_code, ui
   // for the current function. This area is called the 'red zone'. The function we interrupted might
   // have stored temporary data in the red zone and the function we are about to execute might do
   // the same. To keep them separated we decrement rsp by 128 bytes.
-  // The calling convention also asks for rsp being a multiple of 16 so we additionally round down
-  // to the next multiople of 16.
+  // The calling convention also asks for rsp being to be aligned to 16 bytes so we additionally
+  // round down to the next multiple of 16.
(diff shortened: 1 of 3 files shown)


src/UserSpaceInstrumentation/InjectLibraryInTracee.cpp
```diff
@@ -326,4 +269,57 @@ ErrorMessageOr<uint64_t> FindFunctionAddress(pid_t pid, std::string_view functio
...
+  OUTCOME_TRY(original_registers.BackupRegisters(pid));
+
+  RegisterState registers_for_execution = original_registers;
+  registers_for_execution.GetGeneralPurposeRegisters()->x86_64.rip = address_code;
+  // The calling convention for x64 assumes the 128 bytes below rsp to be usable as a scratch pad
+  // for the current function. This area is called the 'red zone'. The function we interrupted might
+  // have stored temporary data in the red zone and the function we are about to execute might do
+  // the same. To keep them separated we decrement rsp by 128 bytes.
+  // The calling convention also asks for rsp being a multiple of 16 so we additionally round down
+  // to the next multiople of 16.
```
dimitry-, line 291: nit: to be aligned to 16 bytes

dimitry- (approved):

src/UserSpaceInstrumentation/InjectLibraryInTracee.cpp
pierricgimmig: or "to the previous multiple"?

pierricgimmig (approved):
Cool, thanks!

src/UserSpaceInstrumentation/InjectLibraryInTracee.cpp
danielfenner: both done.

PR #2164, opened 2021-04-09
Revert "Stable track order when adding/removing FrameTrack"

karupayun:
This reverts commit 7854e6de4de0cf9a769d52d3e6aa564ba619b877.

The commit produced regressions (http://b/184939724) and a one-time reproducible crash. Will prepare a new PR.

### DIFF
--- a/src/OrbitGl/TrackManager.cpp
+++ b/src/OrbitGl/TrackManager.cpp
@@ -142,17 +142,17 @@ void TrackManager::SortTracks() {
 
     last_thread_reorder_.Restart();
 
-    UpdateVisibleTrackList();
+    UpdateFilteredTrackList();
   }
   sorting_invalidated_ = false;
 }
 
 void TrackManager::SetFilter(const std::string& filter) {
   filter_ = absl::AsciiStrToLower(filter);
-  UpdateVisibleTrackList();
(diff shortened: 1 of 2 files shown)


src/OrbitGl/TrackManager.cpp
```diff
@@ -142,17 +142,17 @@ void TrackManager::SortTracks() {
...
 }
 
 void TrackManager::SetFilter(const std::string& filter) {
   filter_ = absl::AsciiStrToLower(filter);
-  UpdateVisibleTrackList();
+  UpdateFilteredTrackList();
 }
 
-void TrackManager::UpdateVisibleTrackList() {
+void TrackManager::UpdateFilteredTrackList() {
```
pierricgimmig, line 155: Why the name change if the function still updates the "visible_tracks_" field? It seems "UpdateVisibleTrackList" better reflects what the function is doing no?

pierricgimmig (approved):
Hi Pablo, LGTM, but please see minor comment, thanks!

src/OrbitGl/TrackManager.cpp
karupayun: It's a plane revert. I didn't make anything else. I will make this PR again taking care of the regressions.

PR #2166, opened 2021-04-12
Track refactoring

florian-kuebler:
### Adjust Positioning System of Tracks - Make it Self-contained.

CaptureViewElements have position and size information for their
appearance in the world coordinate system. This was only partially
used for tracks. The track's header/tab was not included by the
position/size of the track itself. Instead, the layout's gap size
between tracks contained the size of the header. The track relied
on this implicit space and drawn the header out of its scope.

This change adjusted the track's size. It is now the complete size
of a track.

Further, -- since there were bugs and it was unused -- this change
removes the handling for pinned tracks. The were drawn differently,
and remain untested.

Follow-up:
1. Simplify the Accessibility Interface (for tracks).
2. The nice thing about the state before this change, was that
   drawing the track's content did not required any knowledge about
   the header (and other margins). However, the clean solution would
   be to split the Track into a TrackHeader/Tab, (possibly multiple)
   TrackContentPane and TrackFooter.

Test: Take a capture including all types of tracks.

-----

### Introduce AccessibleCaptureViewElement as base class for accessibility.

With CaptureViewElement's following the same positioning (with
their pos/size set correctly) as well as adding knowledge about
their parent, this change extracts some computations for accessibility
into a base class.

Follow-ups:
1. Remove the need for virtual controls, by splitting Track into
   multiple CaptureViewElements.
2. Let the AccessibleCaptureViewElement take more parameters,
   instead of overriding for specialisations.

Test: Use AccessibilityInsights.

_____
### AccessibilityInterface now uses absolute positions for the rect.

Postions of CaptureViewElements as well as in AccessibilityAdapter::rect()
are absolute. Prior this change, we converted CaptureViewElement's position
to a relative one in order to convert it back to an absolute one.
This makes the AccessibilityInterfaces unnecessarily complicated.

This change switch also to absolute positions in the accessibility.
Further, it introduces FakeTimerPane and FakeTrackTab as capture view
elements. Now all position calculations for accessibility are done
by AccessibleCaptureViewElement.

Follow up: The fake classes should be actual children of a Track.
  This requires the actual drawing to be also moved to those classes.

Test: Take a capture and inspect with the AccessibilityInsights.

### DIFF
--- a/src/OrbitGl/AccessibleThreadBar.cpp
+++ b/src/OrbitGl/AccessibleThreadBar.cpp
@@ -12,47 +12,12 @@
 
 namespace orbit_gl {
 
+AccessibleThreadBar::AccessibleThreadBar(const ThreadBar* thread_bar)
+    : AccessibleCaptureViewElement(thread_bar), thread_bar_(thread_bar) {}
+
 int orbit_gl::AccessibleThreadBar::AccessibleChildCount() const { return 0; }
 
-const orbit_accessibility::AccessibleInterface* AccessibleThreadBar::AccessibleParent() const {
-  CHECK(thread_bar_ != nullptr);
-
-  if (thread_bar_->GetParent() == nullptr) return nullptr;
-  if (thread_bar_->GetParent()->GetAccessibleInterface() == nullptr) return nullptr;
-
-  // This relies on knowledge about AccessibleTrack: Accessibility is split into "virtual" controls
-  // that only exist as children in the accessibility tree. In particular, the track is split into
-  // a tab and a content area, and the thread bar is parented underneath the content area (child 1).
-  // This could be cleaned up by introducing "real" control elements for tab and content area.
-  return thread_bar_->GetParent()->GetAccessibleInterface()->AccessibleChild(1);
-}
-
 std::string AccessibleThreadBar::AccessibleName() const { return thread_bar_->GetName(); }
-
-orbit_accessibility::AccessibilityRect orbit_gl::AccessibleThreadBar::AccessibleLocalRect() const {
-  CHECK(thread_bar_ != nullptr);
-
-  Vec2 pos = thread_bar_->GetPos();
-  Vec2 local_pos;
-
-  const CaptureViewElement* parent = thread_bar_->GetParent();
-  // TODO(b/177350599): This could be cleaned up with clearer coordinate systems
-  if (parent != nullptr) {
-    Vec2 parent_pos = parent->GetPos();
-    local_pos = Vec2(pos[0] - parent_pos[0], parent_pos[1] - pos[1]);
-  } else {
-    local_pos = pos;
-  }
-
-  GlCanvas* canvas = thread_bar_->GetCanvas();
-  CHECK(canvas != nullptr);
-  const Viewport& viewport = canvas->GetViewport();
-
-  return orbit_accessibility::AccessibilityRect(
-      viewport.WorldToScreenWidth(local_pos[0]), viewport.WorldToScreenHeight(local_pos[1]),
-      viewport.WorldToScreenWidth(thread_bar_->GetSize()[0]),
-      viewport.WorldToScreenHeight(thread_bar_->GetSize()[1]));
-}
 
 orbit_accessibility::AccessibilityState AccessibleThreadBar::AccessibleState() const {
   return orbit_accessibility::AccessibilityState::Focusable;
--- a/src/OrbitGl/AccessibleTrack.cpp
+++ b/src/OrbitGl/AccessibleTrack.cpp
@@ -22,6 +22,72 @@ using orbit_accessibility::AccessibleInterface;
 
 namespace orbit_gl {
 
+namespace {
+
+class FakeTrackTab : public CaptureViewElement {
+ public:
+  explicit FakeTrackTab(Track* track, TimeGraphLayout* layout)
+      : CaptureViewElement(track, track->GetTimeGraph(), layout), track_(track) {
+    SetCanvas(track->GetCanvas());
+
+    // Compute and set the position and size (which would usually be done by the parent):
+    Vec2 track_pos = track_->GetPos();
+    SetPos(track_pos[0], track_pos[1]);
+
+    SetSize(track->GetSize()[0], layout_->GetTrackTabHeight());
+  }
+
+  std::unique_ptr<AccessibleInterface> CreateAccessibleInterface() override {
+    return std::make_unique<AccessibleTrackTab>(this, track_);
+  }
+
+ private:
+  Track* track_;
+};
+
+class FakeTimerPane : public CaptureViewElement {
+ public:
+  explicit FakeTimerPane(Track* track, TimeGraphLayout* layout, CaptureViewElement* track_tab)
+      : CaptureViewElement(track, track->GetTimeGraph(), layout),
+        track_(track),
+        track_tab_(track_tab) {
+    SetCanvas(track->GetCanvas());
+
+    // Compute and set the position and size (which would usually be done by the parent):
+    std::vector<CaptureViewElement*> track_children = track_->GetVisibleChildren();
+
+    // We are looking for the track's child that is right above the timer pane.
+    CaptureViewElement* predecessor = track_tab_;
+
+    if (!track_children.empty()) {
+      predecessor = track_children[track_children.size() - 1];
+    }
+
+    Vec2 pos = predecessor->GetPos();
+    Vec2 tab_size = predecessor->GetSize();
+    pos[1] -= tab_size[1];
+    SetPos(pos[0], pos[1]);
+
+    Vec2 track_size = track_->GetSize();
+    float track_header_height = track_->GetPos()[1] - pos[1];
+
+    SetSize(track_size[0], track_size[1] - track_header_height);
+  }
+
+  std::unique_ptr<AccessibleInterface> CreateAccessibleInterface() override {
+    return std::make_unique<AccessibleTimerPane>(this);
+  }
+
+ private:
+  Track* track_;
+  CaptureViewElement* track_tab_;
+};
+
+}  //  namespace
+
+AccessibleTrackTab::AccessibleTrackTab(CaptureViewElement* fake_track_tab, Track* track)
+    : AccessibleCaptureViewElement(fake_track_tab), track_(track) {}
+
 const AccessibleInterface* AccessibleTrackTab::AccessibleChild(int index) const {
   if (index == 0) {
     return track_->GetTriangleToggle()->GetOrCreateAccessibleInterface();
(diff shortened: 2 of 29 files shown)


florian-kuebler:
Best to review each commit separately.

florian-kuebler:
Added @vickyliu-go4it FYI

ronaldfw:
Does this have to be a single PR? I haven't looked at detail in the code, maybe there is a dependency I'm missing that requires that these are done as one PR. But it would make reviewing easier if this were split into two PRs.

florian-kuebler:
> Does this have to be a single PR? I haven't looked at detail in the code, maybe there is a dependency I'm missing that requires that these are done as one PR. But it would make reviewing easier if this were split into two PRs.

I thought about this. The issue is, that each commit would break the E2E tests if taking alone. This could be fixed, by adjusting the Accessebility in the first commit. But we would then end up in a very temporal change that would be nearly as large as the second commit.

ronaldfw:
> I thought about this. The issue is, that each commit would break the E2E tests if taking alone. This could be fixed, by adjusting the Accessebility in the first commit. But we would then end up in a very temporal change that would be nearly as large as the second commit.

Ok,  got it. Then it's probably better to do this as a single PR.

karupayun (approved):
LGTM! Thanks for addressing this issue :)

src/OrbitGl/Track.cpp
```diff
@@ -135,11 +123,11 @@ void Track::Draw(GlCanvas* canvas, PickingMode picking_mode, float z_offset) {
     uint32_t sides = static_cast<uint32_t>(layout_->GetRoundingNumSides() + 0.5f);
     auto rounded_corner = GetRoundedCornerMask(radius, sides);
     Vec2 bottom_left(x0, y1);
-    Vec2 bottom_right(tab_x0 + label_width, y0 + top_margin);
-    Vec2 top_right(tab_x0 + label_width, y0 + label_height);
-    Vec2 top_left(tab_x0, y0 + label_height);
+    Vec2 bottom_right(tab_x0 + label_width, y0 - label_height + top_margin);
```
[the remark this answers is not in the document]
florian-kuebler: Done, PTAL.

src/OrbitGl/FrameTrack.cpp
```diff
@@ -79,14 +79,14 @@ FrameTrack::FrameTrack(CaptureViewElement* parent, TimeGraph* time_graph, TimeGr
...
+float FrameTrack::GetHeaderHeight() const { return layout_->GetTrackTabHeight(); }
 
 float FrameTrack::GetHeight() const {
-  return GetMaximumBoxHeight() + layout_->GetTrackBottomMargin();
+  return GetHeaderHeight() + GetMaximumBoxHeight() + layout_->GetTrackBottomMargin();
 }
 
 float FrameTrack::GetYFromTimer(const TimerInfo& /*timer_info*/) const {
-  return pos_[1] - GetMaximumBoxHeight();
+  return pos_[1] - GetHeaderHeight() - GetMaximumBoxHeight();
```
reichlfl, line 89: This (and the overriding `GetYFromTimer`) looks like it would be more straightforward if we renamed it to something like `GetYOffsetFromTimer`, have it return a position relative to the start of the "Timer Area", and have the caller do the required math to convert this into an absolute position, right?

It's probably not something to change in this PR though, but we should keep this in mind when moving to relative positioning.

src/OrbitGl/ThreadTrack.cpp
```diff
@@ -413,6 +413,14 @@ float ThreadTrack::GetHeaderHeight() const {
   return header_height;
 }
 
+float ThreadTrack::GetYFromDepth(uint32_t depth) const {
```
reichlfl, line 416: I'm not sure I understand why this override is needed now. Isn't one of the reasons `GetHeaderHeight` was introduced to avoid overriding `GetYFromDepth` for every class deriving from `TimerTrack` and provide a single template method to customize the timer positioning?

This looks like `gap_between_tracks_and_timers` should be part of the header height calculation, so the logic to check for thread bars is only required in `GetHeaderHeight`.

src/OrbitGl/TimerTrack.cpp
```diff
@@ -59,8 +59,7 @@ float TimerTrack::GetYFromTimer(const TimerInfo& timer_info) const {
 }
 
 float TimerTrack::GetYFromDepth(uint32_t depth) const {
-  return pos_[1] - GetHeaderHeight() - layout_->GetSpaceBetweenTracksAndThread() -
-         box_height_ * static_cast<float>(depth + 1);
+  return pos_[1] - GetHeaderHeight() - box_height_ * static_cast<float>(depth + 1);
```
reichlfl, line 62: See my related comment above, why the change in the usage of `layout_->GetSpaceBetweenTracksAndThread()`?

src/OrbitGl/ThreadTrack.h
```diff
@@ -42,6 +42,7 @@ class ThreadTrack final : public TimerTrack {
   void UpdatePrimitives(Batcher* batcher, uint64_t min_tick, uint64_t max_tick,
                         PickingMode picking_mode, float z_offset = 0) override;
   void OnTimer(const orbit_client_protos::TimerInfo& timer_info) override;
+  [[nodiscard]] virtual float GetYFromDepth(uint32_t depth) const override;
```
reichlfl, line 45: ... and then this would not need to be virtual, because its behavior is customized through `GetHeaderHeight`.

reichlfl (commented):
Thanks a lot, I think this is moving into a much clearer direction. I have a few comments - I think the only one that should be addressed is the one in `AccessibleCaptureViewElement`. The change in `GetYFromDepth` isn't ideal IMHO, but as we discussed yesterday we should handle timer positioning in another way anyways, so I'm ok with leaving it for now given that it will be temporary.

src/OrbitGl/FrameTrack.cpp
florian-kuebler: That would change completely as soon as we have the ContentPane/TimerPane's. In this case, the code would look like before this change, as the pane is already positioned and all computations only need to be relative to its own position.I would keep it as it for now, as this will change anyways soon.

src/OrbitGl/ThreadTrack.cpp
florian-kuebler: Yeah that thing is a bit funcy, but for now, we need to keep it as it is. It is all basically because of the last space between the event bar and the timers pane.
We only want to draw it if there are timers (otherwise, "empty" tracks would have some strange space). However, in the GetHeaderHeight, we don't know about whether we have a timers pane or not.

src/OrbitGl/ThreadTrack.h
florian-kuebler: See above

src/OrbitGl/TimerTrack.cpp
florian-kuebler: Here it is a bit simpler to explain: TimerTrack should not know at all about the possibility of event/thread bars above the timer pane. Only if the track itself actually has an event bar, it should handle this.

As above, it will become obsolete by the refactorings anyways.

src/OrbitGl/ThreadTrack.cpp
reichlfl: Ok, SGTM, let's fix it the proper way in a later PR.

florian-kuebler:
I refactored the GetLocalRect to be GetRect and using absolute positions and added (partial) capture view element implementations for the TrackTab and TimerPane. Accessibility now always uses the AccessibleCaptureViewElement's positioning method.

src/OrbitQt/AccessibilityAdapter.cpp
```diff
@@ -205,14 +205,25 @@ QAccessibleInterface* AccessibilityAdapter::childAt(int x, int y) const {
...
+
+  // Find the outer most AccessibleInterface to convert the coordinates from
+  // pixel relative to the CaptureWindow to screen positions.
+
+  const orbit_accessibility::AccessibleInterface* bridge_info = info_->AccessibleParent();
+  const QAccessibleInterface* bridge = parent();
+  while (bridge_info != nullptr && bridge_info->AccessibleParent() != nullptr) {
+    bridge_info = bridge_info->AccessibleParent();
+    bridge = bridge->parent();
+  }
```
reichlfl, line 218: This could be easier to read if you only trace `AccessibleParent()` upwards, and rename `bridge_info` to `root` or similar.

Once `root->AccessibleParent() == nullptr`, you should be able to retrieve `QAccessibleInterface` from `AdapterRegistry::GetOrCreateAdapter`.

At least consider naming the variables differently so it's clear you're searching for the root of our GL accessibility tree here, I think "bridge" doesn't really help to understand this.

reichlfl (approved):
LGTM, thanks a lot for cleaning this up. I have some nitpicks, but I'm also fine with merging it in the current state since it's not going to be the final iteration.

src/OrbitGl/AccessibleTrack.cpp
```diff
@@ -17,188 +17,137 @@
...
-class AccessibleTimerPane : public AccessibleInterface {
- public:
-  AccessibleTimerPane(orbit_gl::AccessibleTrackContent* parent) : parent_(parent) {}
 
-  [[nodiscard]] int AccessibleChildCount() const override { return 0; }
-  [[nodiscard]] const AccessibleInterface* AccessibleChild(int /*index*/) const override {
-    return nullptr;
-  }
-  [[nodiscard]] const AccessibleInterface* AccessibleParent() const override { return parent_; }
+class FakeTrackTab : public CaptureViewElement {
```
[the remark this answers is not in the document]
florian-kuebler: Yes it should. See the track refactoring doc.

src/OrbitGl/AccessibleThreadBar.cpp
```diff
@@ -12,48 +12,13 @@
...
-  if (thread_bar_->GetParent() == nullptr) return nullptr;
-  if (thread_bar_->GetParent()->GetAccessibleInterface() == nullptr) return nullptr;
-
-  // This relies on knowledge about AccessibleTrack: Accessibility is split into "virtual" controls
-  // that only exist as children in the accessibility tree. In particular, the track is split into
-  // a tab and a content area, and the thread bar is parented underneath the content area (child 1).
-  // This could be cleaned up by introducing "real" control elements for tab and content area.
-  return thread_bar_->GetParent()->GetAccessibleInterface()->AccessibleChild(1);
-}
+int orbit_gl::AccessibleThreadBar::AccessibleChildCount() const { return 0; }
```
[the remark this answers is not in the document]
florian-kuebler: Done. In the future, the capture view element should actually just know the childs.

src/OrbitGl/AccessibleTrack.cpp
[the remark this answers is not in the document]
florian-kuebler: I added a Bug for this.

src/OrbitGl/AccessibleTrack.cpp
```diff
@@ -17,188 +17,137 @@
...
-}
+    // Compute and set the position and size (which would usually be done by the parent):
+    std::vector<CaptureViewElement*> track_children = track_->GetVisibleChildren();
 
-const AccessibleInterface* AccessibleTrackContent::AccessibleChild(int index) const {
-  if (index < 0) {
-    return nullptr;
-  }
+    // We are looking for the track's child that is right above the timer pane.
+    CaptureViewElement* predecessor = track_tab_;
```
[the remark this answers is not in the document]
florian-kuebler: Done

src/OrbitGl/AccessibleTrack.cpp
```diff
@@ -219,7 +168,7 @@ AccessibilityState AccessibleTrack::AccessibleState() const {
     }
   }
 
-  if (AccessibleLocalRect().height == 0) {
+  if (AccessibleRect().height == 0) {
```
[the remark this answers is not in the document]
florian-kuebler: I have the feeling the complete state handling could be simplified in the AccessibleCaptureView. I am going to omit this for now, as I don't want to get pulled into another rabbit hole.

src/OrbitQt/AccessibilityAdapter.cpp
florian-kuebler: Done.

PR #2167, opened 2021-04-12
Quote all fields in exported CSV files

ronaldfw:
Fields in our exported CSV files may contain commas, for example,
strings that represent functions with arguments. To ensure that these
CSV files can be correctly imported in spreadsheet applications, values
for these fields need to be quoted. We also escape any existing quotes
in values. Added a unit test to ensure correctness of the formatting 
code. 

Tested: Exported CSV file from sampled callstack, imported in Google
Sheets. Unit test.
Bug: http://b/181847321

### DIFF
--- a/src/OrbitGl/DataView.cpp
+++ b/src/OrbitGl/DataView.cpp
@@ -4,11 +4,23 @@
 
 #include "DataView.h"
 
+#include <absl/strings/str_replace.h>
+
 #include <memory>
 
 #include "App.h"
 #include "OrbitBase/File.h"
 #include "OrbitBase/Logging.h"
+
+namespace orbit_gl {
+std::string FormatValueForCsv(std::string_view value) {
+  std::string result;
+  result.append("\"");
+  result.append(absl::StrReplaceAll(value, {{"\"", "\"\""}}));
+  result.append("\"");
+  return result;
+}
+}  // namespace orbit_gl
 
 void DataView::InitSortingOrders() {
   sorting_orders_.clear();
@@ -105,7 +117,7 @@ ErrorMessageOr<void> DataView::ExportCSV(const std::filesystem::path& file_path)
   {
     std::string header_line;
     for (size_t i = 0; i < num_columns; ++i) {
-      header_line.append(GetColumns()[i].header);
+      header_line.append(orbit_gl::FormatValueForCsv(GetColumns()[i].header));
       if (i < num_columns - 1) header_line.append(", ");
     }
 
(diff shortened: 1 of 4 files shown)


src/OrbitGl/DataView.cpp
```diff
@@ -102,10 +102,19 @@ ErrorMessageOr<void> DataView::ExportCSV(const std::filesystem::path& file_path)
 
   size_t num_columns = GetColumns().size();
 
+  // Values in the dataview may contain commas, for example, functions with arguments. We quote all
+  // values to ensure the CSV file can be imported correctly, following
+  // https://tools.ietf.org/html/rfc4180.
+  auto append_with_quotes = [](std::string* s, const std::string_view value) {
+    s->append("\"");
+    s->append(value);
```
beckerhe, line 110: AFAIK, quotes in `value` need to be escaped for this to work. Or are we sure that quotes cannot appear in data view fields? If so, this might warrant a comment explaining that.

beckerhe (approved):
Thanks. Otherwise LGTM!

src/OrbitGl/DataView.cpp
ronaldfw: Yes, you are right. I was assuming this doesn't occur, but I'll just fix this right now.

ronaldfw: Ok, should be fixed now.

ronaldfw:
I've addressed your comment that quotes need to be escaped (with another quote) when quoting the fields. I added a unit test for the CSV formatting function. The CSV formatting code and test could be moved to OrbitBase, but since they are not used anywhere else I figured it would be better to keep it local in DataView. But happy to address any additional comments if you don't agree with the choices here.

beckerhe (approved):
Very cool. Thanks. LGTM!

src/OrbitGl/DataView.cpp
```diff
@@ -9,6 +9,20 @@
 #include "App.h"
 #include "OrbitBase/File.h"
 #include "OrbitBase/Logging.h"
+#include "absl/strings/str_join.h"
+#include "absl/strings/str_split.h"
```
dimitry-, line 13: nit: we use `<>` include style for external to Orbit headers

ronaldfw: Ah, thanks, will fix right away!

ronaldfw: Should be fixed now.

src/OrbitGl/DataView.cpp
```diff
@@ -4,12 +4,27 @@
...
 #include "App.h"
 #include "OrbitBase/File.h"
 #include "OrbitBase/Logging.h"
 
+namespace orbit_gl {
+std::string FormatValueForCsv(std::string_view value) {
+  std::string result;
+  result.append("\"");
+  std::vector<std::string> split_value = absl::StrSplit(value, "\"");
+  std::string value_with_escaped_quotes = absl::StrJoin(split_value, "\"\"");
```
dimitry-, line 21: Maybe `absl::StrReplaceAll({"\"", "\"\""})`?

ronaldfw: Great, that's better, will change it!

dimitry- (approved):
thanks!

PR #2168, opened 2021-04-12
Add flag for the memory warning threshold

vickyliu-go4it:
With this change, we add the feature flag for the memory warning
threshold (i.e., the production limit) and set the default value as
false. This will be remove once we have a clear explanation of the
production limit.

### DIFF
--- a/src/OrbitGl/ClientFlags.cpp
+++ b/src/OrbitGl/ClientFlags.cpp
@@ -41,3 +41,8 @@ ABSL_FLAG(bool, enable_tracepoint_feature, false,
 
 // TODO(b/181736566): Remove this flag entirely
 ABSL_FLAG(bool, enable_source_code_view, true, "Enable the experimental source code view");
+
+// TODO(b/185099421): Remove this flag once we have a clear explanation of the memory warning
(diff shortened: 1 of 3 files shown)


antonrohr (approved):
lgtm, thanks for doing this with a flag :)

src/OrbitGl/ClientFlags.cpp
```diff
@@ -41,3 +41,8 @@ ABSL_FLAG(bool, enable_tracepoint_feature, false,
 
 // TODO(b/181736566): Remove this flag entirely
 ABSL_FLAG(bool, enable_source_code_view, true, "Enable the experimental source code view");
+
+// TODO(vickyliu): Remove this flag once we have a clear explanation of the memory warning threshold
+// (i.e., production limit).
```
karupayun, line 46: Please link a bug for this (we need it anyway :) )

karupayun (approved):
LGTM!

src/OrbitGl/ClientFlags.cpp
vickyliu-go4it: Done! Thanks for the suggestion!

PR #2170, opened 2021-04-12
Do not draw timer's text when collapsed.

florian-kuebler:
With the introduction of the scope tree datastructure,
ThreadTrack::UpdatePrimitives was reimplemented, overriding
the one in TimerTrack. However, before that change, the text
was not renedered on callapsed timers.

Test: Run with --devmode, collapse some tracks.
Bug: http://b/184938662


reichlfl (approved):
Thanks for fixing, LGTM

PR #2172, opened 2021-04-12
Set pid of FullGpuJob, show it in GpuTrack's tooltips

dpallotti:
#### Set pid of FullGpuJob
`FullGpuJob` has a pid that `LinuxTracing` didn't set. But
`ProducerEventProcessor` copies that pid into `GpuJob` and
`CaptureEventProcessor` copies that pid into `TimerInfo`.

This was not a problem for the client as the pid is currently not shown.

As the frequency of these events is low and the impact on bandwidth is
negligible (the average number of bytes per event on a 10-second capture stays
around 25.0 in the range of measurement error), just set the pid.

I noticed this while adding tests for `GpuJob`s to
`LinuxTracingIntegrationTests`.

Bug: http://b/184942487

Test: Update unit tests. Capture Trata and see that gpu jobs in the timeline are
as usual.

#### Add "Submitted from process" to GpuTrack's tooltips
Since we now have the information from `GpuJob`s.
This can be relevant when GPU activity is caused by a process different than the
target.

Bug: http://b/185202201

Test: Capture Trata and see that the tooltips show the new information.
Also do the same with the Vulkan layer and verify the tooltips of command
buffers and debug markers.

### DIFF
--- a/src/LinuxTracing/GpuTracepointVisitor.cpp
+++ b/src/LinuxTracing/GpuTracepointVisitor.cpp
@@ -60,6 +60,7 @@ void GpuTracepointVisitor::CreateGpuJobAndSendToListenerIfComplete(const Key& ke
   }
 
   std::string timeline = cs_it->second.timeline;
+  pid_t pid = cs_it->second.pid;
   pid_t tid = cs_it->second.tid;
 
   // We assume that GPU jobs (command buffer submissions) immediately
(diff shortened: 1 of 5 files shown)


dimitry- (approved):

ronaldfw (approved):
Nice, thank you! LGTM

florian-kuebler (approved):
LGTM, thanks!

PR #2173, opened 2021-04-12
Faster access to memory of tracee.

danielfenner:
Previously we used the PTRACE_PEEKDATA / PTRACE_POKEDATA interface to
access the tracees's memory. This is slow since we can only access eight
byte at a time and an access requires two context switches. Also the
access can only happen in multiples of eight bytes which complicates
things.

The filesystem based interface in /proc/pid/mem is both faster and more
convenient.

Bug: http://b/183115358
Test: Existing unit tests.

### DIFF
--- a/src/OrbitBase/File.cpp
+++ b/src/OrbitBase/File.cpp
@@ -15,8 +15,11 @@
 #endif
 
 #if defined(_WIN32)
+
 // Windows never returns EINTR - so there is no need for TEMP_FAILURE_RETRY implementation
 #define TEMP_FAILURE_RETRY(expression) (expression)
+
+namespace {
 
 using ssize_t = int64_t;
 
@@ -52,6 +55,9 @@ ssize_t pwrite(int fd, const void* buffer, size_t size, off_t offset) {
   }
   return bytes_written;
 }
+
+}  // namespace
+
 #endif
 
 namespace orbit_base {
--- a/src/OrbitBase/File.cpp
+++ b/src/OrbitBase/File.cpp
@@ -85,6 +85,7 @@ ErrorMessageOr<void> WriteFullyAtOffset(const unique_fd& fd, const void* buffer,
 
   while (size > 0) {
     int64_t bytes_written = TEMP_FAILURE_RETRY(pwrite(fd.get(), current_position, size, offset));
+    CHECK(bytes_written != 0);
     if (bytes_written == -1) {
       return ErrorMessage{SafeStrerror(errno)};
     }
@@ -122,22 +123,22 @@ ErrorMessageOr<size_t> ReadFullyAtOffset(const unique_fd& fd, void* buffer, size
   FATAL("Not implemented on Windows.")
 #elif defined(__linux)
   uint8_t* current_position = static_cast<uint8_t*>(buffer);
-  size_t bytes_received = 0;
-  while (bytes_received < size) {
+  size_t bytes_read = 0;
+  while (bytes_read < size) {
     int64_t pread_result = TEMP_FAILURE_RETRY(pread(fd.get(), current_position, size, offset));
     if (pread_result == -1) {
       return ErrorMessage{SafeStrerror(errno)};
     }
     if (pread_result == 0) {
-      return bytes_received;
+      break;
     }
-    bytes_received += pread_result;
+    bytes_read += pread_result;
     current_position += pread_result;
     size -= pread_result;
     offset += pread_result;
   }
 
-  return bytes_received;
+  return bytes_read;
 #endif
 }
 
--- a/src/OrbitBase/File.cpp
+++ b/src/OrbitBase/File.cpp
@@ -76,6 +76,24 @@ ErrorMessageOr<void> WriteFully(const unique_fd& fd, std::string_view content) {
   return outcome::success();
 }
 
+ErrorMessageOr<void> WriteFullyAtOffset(const unique_fd& fd, const void* buffer, size_t size,
+                                        off_t offset) {
+  const char* current_position = static_cast<const char*>(buffer);
+
+  while (size > 0) {
+    int64_t bytes_written = TEMP_FAILURE_RETRY(pwrite(fd.get(), current_position, size, offset));
+    if (bytes_written == -1) {
+      return ErrorMessage{SafeStrerror(errno)};
+    }
+    current_position += bytes_written;
+    offset += bytes_written;
+    size -= bytes_written;
+  }
+
+  CHECK(size == 0);
+  return outcome::success();
+}
+
 ErrorMessageOr<size_t> ReadFully(const unique_fd& fd, void* buffer, size_t size) {
   size_t bytes_left = size;
   auto current_position = static_cast<uint8_t*>(buffer);
@@ -92,6 +110,27 @@ ErrorMessageOr<size_t> ReadFully(const unique_fd& fd, void* buffer, size_t size)
   }
 
   return size - bytes_left;
+}
+
+ErrorMessageOr<size_t> ReadFullyAtOffset(const unique_fd& fd, void* buffer, size_t size,
+                                         off_t offset) {
+  uint8_t* current_position = static_cast<uint8_t*>(buffer);
+  size_t bytes_received = 0;
+  while (bytes_received < size) {
+    int64_t pread_result = TEMP_FAILURE_RETRY(pread(fd.get(), current_position, size, offset));
+    if (pread_result == -1) {
+      return ErrorMessage{SafeStrerror(errno)};
+    }
+    if (pread_result == 0) {
+      return bytes_received;
+    }
+    bytes_received += pread_result;
+    current_position += pread_result;
+    size -= pread_result;
+    offset += pread_result;
+  }
+
+  return bytes_received;
 }
 
 }  // namespace orbit_base
--- a/src/OrbitBase/File.cpp
+++ b/src/OrbitBase/File.cpp
@@ -4,6 +4,8 @@
 
 #include "OrbitBase/File.h"
 
+#include <cstdint>
+
 #include "OrbitBase/Logging.h"
 
 #if defined(__linux)
@@ -15,6 +17,41 @@
 #if defined(_WIN32)
 // Windows never returns EINTR - so there is no need for TEMP_FAILURE_RETRY implementation
 #define TEMP_FAILURE_RETRY(expression) (expression)
+
+using ssize_t = int64_t;
+
+// There is no implementation for pread and pwrite in 'io.h' so we implement them on top of lseek,
+// read and write here.
+ssize_t pread(int fd, void* buffer, size_t size, off_t offset) {
+  off_t old_position = lseek(fd, 0, SEEK_CUR);
+  if (old_position == -1) {
+    return -1;
+  }
+  if (lseek(fd, offset, SEEK_SET) == -1) {
+    return -1;
+  }
+  ssize_t bytes_read = read(fd, buffer, size);
+  if ((lseek(fd, old_position, SEEK_SET)) == -1) {
+    return -1;
+  }
+  return bytes_read;
+}
+
+ssize_t pwrite(int fd, const void* buffer, size_t size, off_t offset) {
+  off_t cpos, opos;
+  off_t old_position = lseek(fd, 0, SEEK_CUR);
+  if (old_position == -1) {
+    return -1;
+  }
+  if (lseek(fd, offset, SEEK_SET) == -1) {
+    return -1;
+  }
+  ssize_t bytes_written = write(fd, buffer, size);
+  if (lseek(fd, old_position, SEEK_SET) == -1) {
+    return -1;
+  }
+  return bytes_written;
+}
 #endif
 
 namespace orbit_base {
@@ -78,9 +115,6 @@ ErrorMessageOr<void> WriteFully(const unique_fd& fd, std::string_view content) {
 
 ErrorMessageOr<void> WriteFullyAtOffset(const unique_fd& fd, const void* buffer, size_t size,
                                         off_t offset) {
-#if defined(_WIN32)
-  FATAL("Not implemented on Windows.");
-#elif defined(__linux)
   const char* current_position = static_cast<const char*>(buffer);
 
   while (size > 0) {
@@ -96,7 +130,6 @@ ErrorMessageOr<void> WriteFullyAtOffset(const unique_fd& fd, const void* buffer,
 
   CHECK(size == 0);
   return outcome::success();
-#endif
 }
 
 ErrorMessageOr<size_t> ReadFully(const unique_fd& fd, void* buffer, size_t size) {
@@ -119,9 +152,6 @@ ErrorMessageOr<size_t> ReadFully(const unique_fd& fd, void* buffer, size_t size)
 
 ErrorMessageOr<size_t> ReadFullyAtOffset(const unique_fd& fd, void* buffer, size_t size,
                                          off_t offset) {
-#if defined(_WIN32)
-  FATAL("Not implemented on Windows.");
-#elif defined(__linux)
   uint8_t* current_position = static_cast<uint8_t*>(buffer);
   size_t bytes_read = 0;
   while (bytes_read < size) {
@@ -139,7 +169,6 @@ ErrorMessageOr<size_t> ReadFullyAtOffset(const unique_fd& fd, void* buffer, size
   }
 
   return bytes_read;
-#endif
 }
 
 }  // namespace orbit_base
--- a/src/OrbitBase/File.cpp
+++ b/src/OrbitBase/File.cpp
@@ -79,7 +79,7 @@ ErrorMessageOr<void> WriteFully(const unique_fd& fd, std::string_view content) {
 ErrorMessageOr<void> WriteFullyAtOffset(const unique_fd& fd, const void* buffer, size_t size,
                                         off_t offset) {
 #if defined(_WIN32)
-  FATAL("Not implemented on Windows.")
+  FATAL("Not implemented on Windows.");
 #elif defined(__linux)
   const char* current_position = static_cast<const char*>(buffer);
 
@@ -120,7 +120,7 @@ ErrorMessageOr<size_t> ReadFully(const unique_fd& fd, void* buffer, size_t size)
 ErrorMessageOr<size_t> ReadFullyAtOffset(const unique_fd& fd, void* buffer, size_t size,
                                          off_t offset) {
 #if defined(_WIN32)
-  FATAL("Not implemented on Windows.")
+  FATAL("Not implemented on Windows.");
 #elif defined(__linux)
   uint8_t* current_position = static_cast<uint8_t*>(buffer);
   size_t bytes_read = 0;
--- a/src/OrbitBase/File.cpp
+++ b/src/OrbitBase/File.cpp
@@ -78,6 +78,9 @@ ErrorMessageOr<void> WriteFully(const unique_fd& fd, std::string_view content) {
 
 ErrorMessageOr<void> WriteFullyAtOffset(const unique_fd& fd, const void* buffer, size_t size,
                                         off_t offset) {
+#if defined(_WIN32)
+  FATAL("Not implemented on Windows.")
+#elif defined(__linux)
   const char* current_position = static_cast<const char*>(buffer);
 
   while (size > 0) {
@@ -92,6 +95,7 @@ ErrorMessageOr<void> WriteFullyAtOffset(const unique_fd& fd, const void* buffer,
 
   CHECK(size == 0);
   return outcome::success();
+#endif
 }
 
 ErrorMessageOr<size_t> ReadFully(const unique_fd& fd, void* buffer, size_t size) {
@@ -114,6 +118,9 @@ ErrorMessageOr<size_t> ReadFully(const unique_fd& fd, void* buffer, size_t size)
 
 ErrorMessageOr<size_t> ReadFullyAtOffset(const unique_fd& fd, void* buffer, size_t size,
                                          off_t offset) {
+#if defined(_WIN32)
+  FATAL("Not implemented on Windows.")
+#elif defined(__linux)
   uint8_t* current_position = static_cast<uint8_t*>(buffer);
   size_t bytes_received = 0;
   while (bytes_received < size) {
@@ -131,6 +138,7 @@ ErrorMessageOr<size_t> ReadFullyAtOffset(const unique_fd& fd, void* buffer, size
   }
 
   return bytes_received;
+#endif
 }
 
 }  // namespace orbit_base
--- a/src/OrbitBase/FileTest.cpp
+++ b/src/OrbitBase/FileTest.cpp
@@ -83,6 +83,65 @@ TEST(File, OpenNewFileForReadWrite) {
   ASSERT_TRUE(fd_or_error.has_value()) << fd_or_error.error().message();
 }
 
+TEST(File, WriteFullySmoke) {
+  auto temporary_file_or_error = TemporaryFile::Create();
+  ASSERT_TRUE(temporary_file_or_error.has_value()) << temporary_file_or_error.error().message();
+  TemporaryFile temporary_file = std::move(temporary_file_or_error.value());
+
+  // Write buffer into file.
+  const std::string buffer = "blub\nbla\n";
+  ErrorMessageOr<void> write_result_or_error = WriteFully(temporary_file.fd(), buffer);
+  ASSERT_FALSE(write_result_or_error.has_error()) << write_result_or_error.error().message();
+
+  // Read back and compare content.
+  ErrorMessageOr<unique_fd> fd_or_error = OpenFileForReading(temporary_file.file_path().string());
+  ASSERT_FALSE(fd_or_error.has_error()) << fd_or_error.error().message();
+  std::array<char, 64> read_back = {};
+  ErrorMessageOr<size_t> result_or_error = ReadFully(fd_or_error.value(), read_back.data(), 9);
+  ASSERT_FALSE(result_or_error.has_error()) << result_or_error.error().message();
+  EXPECT_EQ(result_or_error.value(), 9);
+  EXPECT_STREQ(read_back.data(), buffer.c_str());
+
+  temporary_file.CloseAndRemove();
+}
+
+TEST(File, WriteFullyAtOffsetSmoke) {
+  auto temporary_file_or_error = TemporaryFile::Create();
+  ASSERT_TRUE(temporary_file_or_error.has_value()) << temporary_file_or_error.error().message();
+  TemporaryFile temporary_file = std::move(temporary_file_or_error.value());
+
+  // Write at the beginning of the previously empty file.
+  std::array<char, 64> buffer{'a', 'b', '\n', 'c', 'd', '\n'};
+  ErrorMessageOr<void> write_result_or_error =
+      WriteFullyAtOffset(temporary_file.fd(), buffer.data(), 6, 0);
+  ASSERT_FALSE(write_result_or_error.has_error()) << write_result_or_error.error().message();
+
+  // Read back and compare content.
+  std::array<char, 64> read_back = {};
+  ErrorMessageOr<unique_fd> fd_or_error = OpenFileForReading(temporary_file.file_path().string());
+  ASSERT_FALSE(fd_or_error.has_error()) << fd_or_error.error().message();
+  ErrorMessageOr<size_t> result_or_error = ReadFully(fd_or_error.value(), read_back.data(), 64);
+  ASSERT_FALSE(result_or_error.has_error()) << result_or_error.error().message();
+  EXPECT_EQ(result_or_error.value(), 6);
+  EXPECT_STREQ(read_back.data(), buffer.data());
+
+  // Overrride at a given offset.
+  buffer = {'e', 'f'};
+  write_result_or_error = WriteFullyAtOffset(temporary_file.fd(), buffer.data(), 2, 3);
+  ASSERT_FALSE(write_result_or_error.has_error()) << write_result_or_error.error().message();
+
+  // Read back and compare content.
+  read_back = {};
+  fd_or_error = OpenFileForReading(temporary_file.file_path().string());
+  ASSERT_FALSE(fd_or_error.has_error()) << fd_or_error.error().message();
+  result_or_error = ReadFully(fd_or_error.value(), read_back.data(), 64);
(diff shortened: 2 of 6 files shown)


src/UserSpaceInstrumentation/AccessTraceesMemory.cpp
```diff
@@ -25,44 +26,61 @@ using orbit_base::ReadFileToString;
...
-  for (size_t i = 0; i < length / 8; i++) {
-    const uint64_t data = ptrace(PTRACE_PEEKDATA, pid, address_start + i * 8, NULL);
-    if (errno) {
-      return ErrorMessage(
-          absl::StrFormat("Failed to PTRACE_PEEKDATA for pid %d with errno %d: \"%s\"", pid, errno,
-                          SafeStrerror(errno)));
-    }
-    std::memcpy(bytes.data() + (i * sizeof(uint64_t)), &data, 8);
+  if (length == 0) {
+    return ErrorMessage("Tried to read 0 bytes.");
```
dimitry-, line 30: shouldn't this be a CHECK?

src/UserSpaceInstrumentation/AccessTraceesMemory.cpp
```diff
@@ -25,44 +26,61 @@ using orbit_base::ReadFileToString;
...
   }
-  return bytes;
+
+  OUTCOME_TRY(fd, orbit_base::OpenFileForReading(absl::StrFormat("/proc/%d/mem", pid)));
+
+  std::vector<uint8_t> data(length);
+  ssize_t result = pread(fd.get(), data.data(), length, address_start);
```
dimitry-, line 36: maybe implement orbit_base::ReadFullyAtOffset() note that pread can return value less then length in case it was interrupted by signal, so we want to just continue reading in this case.

src/UserSpaceInstrumentation/AccessTraceesMemory.cpp
```diff
@@ -25,44 +26,61 @@ using orbit_base::ReadFileToString;
...
+
+  if (result == -1) {
+    return ErrorMessage(
+        absl::StrFormat("Failed to read from memory file of process %d with errno %d: \"%s\"", pid,
+                        errno, SafeStrerror(errno)));
+  }
+
+  if (result < static_cast<ssize_t>(length)) {
```
dimitry-, line 44: "Note that it is not an error for a successful call to transfer fewer bytes than requested (see read(2) and write(2))."

src/UserSpaceInstrumentation/AccessTraceesMemory.cpp
```diff
@@ -25,44 +26,61 @@ using orbit_base::ReadFileToString;
...
-    if (ptrace(PTRACE_POKEDATA, pid, address_start + pos, data) == -1) {
-      return ErrorMessage(absl::StrFormat("Failed to PTRACE_POKEDATA for pid: %d.", pid));
-    }
-    pos += sizeof(uint64_t);
-  } while (pos < bytes.size());
+  if (bytes.empty()) {
+    return ErrorMessage("Tried to write 0 bytes.");
+  }
+  OUTCOME_TRY(fd, orbit_base::OpenFileForWriting(absl::StrFormat("/proc/%d/mem", pid)));
+  ssize_t result = pwrite(fd.get(), bytes.data(), bytes.size(), address_start);
```
dimitry-, line 59: same here - WriteFullyAtOffset, same thing about having written less than length is not a error.

src/UserSpaceInstrumentation/AccessTraceesMemoryTest.cpp
```diff
@@ -18,47 +21,191 @@
...
+    FATAL("Not a number.");
+  }
+  if (!absl::numbers_internal::safe_strtou64_base(addresses[1], &result.second, 16)) {
+    FATAL("Not a number.");
+  }
+  return result;
+}
+
+// Get the adress range of the first consecutive mappings. We can read this range but not more.
+[[nodiscard]] ErrorMessageOr<AddressRange> GetFirstContinuesAdressRange(pid_t pid) {
```
dpallotti, line 47: Continues -> Continuous
Adress -> Address

src/UserSpaceInstrumentation/AccessTraceesMemoryTest.cpp
```diff
@@ -18,47 +21,191 @@
...
+  if (pid == 0) {
+    // Child just runs an endless loop.
+    while (true) {
+    }
+  }
+
+  // Stop the child process using our tooling.
+  ASSERT_TRUE(AttachAndStopProcess(pid).has_value());
+
+  auto continues_range_or_error = GetFirstContinuesAdressRange(pid);
```
dpallotti, line 82: continuous

src/UserSpaceInstrumentation/AccessTraceesMemoryTest.cpp
```diff
@@ -18,47 +21,191 @@
...
```
dpallotti, line 131: continuous

src/UserSpaceInstrumentation/AccessTraceesMemoryTest.cpp
```diff
@@ -18,47 +21,191 @@
...
+  result = WriteTraceesMemory(-1, address, bytes);
+  ASSERT_TRUE(result.has_error());
+  EXPECT_TRUE(absl::StrContains(result.error().message(), "Unable to open file"));
+
+  // Write 0 bytes.
+  result = WriteTraceesMemory(pid, address, std::vector<uint8_t>());
+  ASSERT_TRUE(result.has_error());
+  EXPECT_TRUE(absl::StrContains(result.error().message(), "Tried to write 0 bytes."));
+
+  // Read past the end of the mappings.
```
dpallotti, line 156:
```suggestion
  // Write past the end of the mappings.
```

src/UserSpaceInstrumentation/AccessTraceesMemoryTest.cpp
```diff
@@ -18,47 +21,191 @@
...
+  bytes.push_back(0);
+  result = WriteTraceesMemory(pid, address, bytes);
+  ASSERT_TRUE(result.has_error());
+  EXPECT_TRUE(absl::StrContains(result.error().message(), "Only wrote"));
+
+  // Read from bad address.
```
dpallotti, line 162:
```suggestion
  // Write to bad address.
```

src/UserSpaceInstrumentation/AccessTraceesMemoryTest.cpp
```diff
@@ -18,47 +21,191 @@
...
+  ASSERT_FALSE(WriteTraceesMemory(pid, address, backup.value()).has_error());
+  ASSERT_FALSE(DetachAndContinueProcess(pid).has_error());
+  kill(pid, SIGKILL);
+  waitpid(pid, NULL, 0);
+}
+
 TEST(AccessTraceesMemoryTest, ReadWriteRestore) {
   pid_t pid = fork();
-  CHECK(pid != -1);
+  ASSERT_TRUE(pid != -1);
```
dpallotti, line 176: I feel like this and a few below were better as `CHECK` as they are not really part assessing what is being tested, rather than the setup for the test is working as expected.

src/UserSpaceInstrumentation/AccessTraceesMemory.cpp
danielfenner: done

src/UserSpaceInstrumentation/AccessTraceesMemoryTest.cpp
```diff
@@ -18,47 +21,191 @@
...
+  if (!absl::numbers_internal::safe_strtou64_base(addresses[0], &result.first, 16)) {
+    FATAL("Not a number.");
+  }
+  if (!absl::numbers_internal::safe_strtou64_base(addresses[1], &result.second, 16)) {
+    FATAL("Not a number.");
+  }
+  return result;
+}
+
+// Get the adress range of the first consecutive mappings. We can read this range but not more.
```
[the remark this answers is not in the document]
danielfenner: done

src/UserSpaceInstrumentation/AccessTraceesMemoryTest.cpp
danielfenner: done

src/UserSpaceInstrumentation/AccessTraceesMemoryTest.cpp
danielfenner: done

src/UserSpaceInstrumentation/AccessTraceesMemoryTest.cpp
danielfenner: done

src/UserSpaceInstrumentation/AccessTraceesMemoryTest.cpp
danielfenner: done

src/UserSpaceInstrumentation/AccessTraceesMemoryTest.cpp
danielfenner: done

src/UserSpaceInstrumentation/AccessTraceesMemoryTest.cpp
danielfenner: Yep, this came up before and is tracked here: http://b/185312970 
I changed it in the places I touched here; the other stuff will be handled in a followup PR.

src/OrbitBase/File.cpp
```diff
@@ -76,6 +76,24 @@ ErrorMessageOr<void> WriteFully(const unique_fd& fd, std::string_view content) {
   return outcome::success();
 }
 
+ErrorMessageOr<void> WriteFullyAtOffset(const unique_fd& fd, const void* buffer, size_t size,
+                                        off_t offset) {
+  const char* current_position = static_cast<const char*>(buffer);
+
+  while (size > 0) {
+    int64_t bytes_written = TEMP_FAILURE_RETRY(pwrite(fd.get(), current_position, size, offset));
```
dimitry-, line 84: CHECK(bytes_written != 0);

src/OrbitBase/File.cpp
```diff
@@ -94,4 +112,25 @@ ErrorMessageOr<size_t> ReadFully(const unique_fd& fd, void* buffer, size_t size)
...
+                                         off_t offset) {
+  uint8_t* current_position = static_cast<uint8_t*>(buffer);
+  size_t bytes_received = 0;
+  while (bytes_received < size) {
+    int64_t pread_result = TEMP_FAILURE_RETRY(pread(fd.get(), current_position, size, offset));
+    if (pread_result == -1) {
+      return ErrorMessage{SafeStrerror(errno)};
+    }
+    if (pread_result == 0) {
+      return bytes_received;
```
dimitry-, line 125: break;

src/OrbitBase/FileTest.cpp
```diff
@@ -83,6 +83,65 @@ TEST(File, OpenNewFileForReadWrite) {
...
+  // Write buffer into file.
+  const std::string buffer = "blub\nbla\n";
+  ErrorMessageOr<void> write_result_or_error = WriteFully(temporary_file.fd(), buffer);
+  ASSERT_FALSE(write_result_or_error.has_error()) << write_result_or_error.error().message();
+
+  // Read back and compare content.
+  ErrorMessageOr<unique_fd> fd_or_error = OpenFileForReading(temporary_file.file_path().string());
+  ASSERT_FALSE(fd_or_error.has_error()) << fd_or_error.error().message();
+  std::array<char, 64> read_back = {};
+  ErrorMessageOr<size_t> result_or_error = ReadFully(fd_or_error.value(), read_back.data(), 9);
```
dimitry-, line 100: maybe read 11 to check that it doesn't read past of eof. also use buffer.size()+3 instead of hardcoding the size of the buffer.

src/OrbitBase/include/OrbitBase/File.h
```diff
@@ -87,6 +90,9 @@ ErrorMessageOr<void> WriteFully(const unique_fd& fd, std::string_view content);
 // used for non-blocking reads from sockets/pipes - it does not handle EAGAIN.
 ErrorMessageOr<size_t> ReadFully(const unique_fd& fd, void* buffer, size_t size);
 
+ErrorMessageOr<size_t> ReadFullyAtOffset(const unique_fd& fd, void* buffer, size_t size,
```
dimitry-, line 93: same disclaimer about `don't use for non-blocking reads` as for ReadFully applies here (maybe add a comment about that)

src/UserSpaceInstrumentation/AccessTraceesMemory.cpp
```diff
@@ -25,44 +26,37 @@ using orbit_base::ReadFileToString;
...
   std::vector<uint8_t> bytes(length);
-  for (size_t i = 0; i < length / 8; i++) {
-    const uint64_t data = ptrace(PTRACE_PEEKDATA, pid, address_start + i * 8, NULL);
-    if (errno) {
-      return ErrorMessage(
-          absl::StrFormat("Failed to PTRACE_PEEKDATA for pid %d with errno %d: \"%s\"", pid, errno,
-                          SafeStrerror(errno)));
-    }
-    std::memcpy(bytes.data() + (i * sizeof(uint64_t)), &data, 8);
+  OUTCOME_TRY(result, ReadFullyAtOffset(fd, bytes.data(), length, address_start));
```
dimitry-, line 34: this error does not provide context. It will return something like "Permission denied." maybe add an if (result.has_error) { provide some context about what it was trying to do (read memory of a process with pid=%d) }

dimitry-: My personal preference is not to use OUTCOME_TRY.

src/UserSpaceInstrumentation/AccessTraceesMemory.cpp
```diff
@@ -25,44 +26,37 @@ using orbit_base::ReadFileToString;
...
-    if (ptrace(PTRACE_POKEDATA, pid, address_start + pos, data) == -1) {
-      return ErrorMessage(absl::StrFormat("Failed to PTRACE_POKEDATA for pid: %d.", pid));
-    }
-    pos += sizeof(uint64_t);
-  } while (pos < bytes.size());
+  CHECK(!bytes.empty());
+
+  OUTCOME_TRY(fd, orbit_base::OpenFileForWriting(absl::StrFormat("/proc/%d/mem", pid)));
+
+  OUTCOME_TRY(WriteFullyAtOffset(fd, bytes.data(), bytes.size(), address_start));
```
dimitry-, line 51: same here about the error message.

src/UserSpaceInstrumentation/AccessTraceesMemoryTest.cpp
```diff
@@ -18,11 +21,101 @@
...
 namespace orbit_user_space_instrumentation {
 
-TEST(AccessTraceesMemoryTest, ReadWriteRestore) {
+namespace {
+
+AddressRange AddressRangeFromString(const std::string& s) {
+  AddressRange result;
+  const std::vector<std::string> addresses = absl::StrSplit(s, '-');
+  if (addresses.size() != 2) {
+    FATAL("Not an address range.");
```
dimitry-, line 35: what is not an adress range? add %s, s

src/UserSpaceInstrumentation/AccessTraceesMemoryTest.cpp
```diff
@@ -18,11 +21,101 @@
...
 #include "OrbitBase/Logging.h"
+#include "OrbitBase/ReadFileToString.h"
 #include "RegisterState.h"
 
 namespace orbit_user_space_instrumentation {
 
-TEST(AccessTraceesMemoryTest, ReadWriteRestore) {
+namespace {
+
+AddressRange AddressRangeFromString(const std::string& s) {
```
dimitry-, line 31: s/s/address_string/

src/UserSpaceInstrumentation/AccessTraceesMemoryTest.cpp
```diff
@@ -18,11 +21,101 @@
...
+  AddressRange result;
+  const std::vector<std::string> addresses = absl::StrSplit(s, '-');
+  if (addresses.size() != 2) {
+    FATAL("Not an address range.");
+  }
+  if (!absl::numbers_internal::safe_strtou64_base(addresses[0], &result.first, 16)) {
+    FATAL("Not a number.");
```
dimitry-, line 38: %s, addresses[0]

src/UserSpaceInstrumentation/AccessTraceesMemoryTest.cpp
```diff
@@ -18,11 +21,101 @@
...
+  }
+  if (!absl::numbers_internal::safe_strtou64_base(addresses[1], &result.second, 16)) {
+    FATAL("Not a number.");
```
dimitry-, line 41: same here.

src/UserSpaceInstrumentation/AccessTraceesMemoryTest.cpp
```diff
@@ -18,11 +21,101 @@
...
+  const uint64_t length = continuous_range.second - continuous_range.first;
+
+  // Good read.
+  ErrorMessageOr<std::vector<uint8_t>> result = ReadTraceesMemory(pid, address, length);
+  EXPECT_TRUE(result.has_value());
+
+  // Process does not exist.
+  result = ReadTraceesMemory(-1, address, length);
+  ASSERT_TRUE(result.has_error());
+  EXPECT_TRUE(absl::StrContains(result.error().message(), "Unable to open file"))
```
dimitry-, line 95: nitpick: EXPECT_THAT(result.error().message()), testing::HasSubstr("..."));

src/UserSpaceInstrumentation/AccessTraceesMemoryTest.cpp
```diff
@@ -18,11 +21,101 @@
...
+      << result.error().message();
+
+  // Read 0 bytes.
+  EXPECT_DEATH(auto unused_result = ReadTraceesMemory(pid, address, 0), "Check failed");
+
+  // Read past the end of the mappings.
+  result = ReadTraceesMemory(pid, address, length + 1);
+  ASSERT_TRUE(result.has_error());
+  EXPECT_TRUE(absl::StrContains(result.error().message(), "Only got")) << result.error().message();
```
dimitry-, line 104: same here and below.

src/UserSpaceInstrumentation/AccessTraceesMemoryTest.cpp
```diff
@@ -32,33 +125,88 @@ TEST(AccessTraceesMemoryTest, ReadWriteRestore) {
...
+  }
+
+  // Stop the child process using our tooling.
+  ASSERT_TRUE(AttachAndStopProcess(pid).has_value());
 
-  auto result_memory_region = GetFirstExecutableMemoryRegion(pid);
-  CHECK(result_memory_region.has_value());
-  const uint64_t address_start = result_memory_region.value().first;
+  auto memory_region_or_error = GetFirstExecutableMemoryRegion(pid);
+  ASSERT_TRUE(memory_region_or_error.has_value());
```
dimitry-, line 188: if it fails here - we loose the error message. Maybe we want to introduce a macro that checks `ErrorMessageOr<>`  and always adds the actual error message.

dimitry- (approved):
otherwise lgtm

src/OrbitBase/File.cpp
```diff
@@ -94,4 +112,25 @@ ErrorMessageOr<size_t> ReadFully(const unique_fd& fd, void* buffer, size_t size)
   return size - bytes_left;
 }
 
+ErrorMessageOr<size_t> ReadFullyAtOffset(const unique_fd& fd, void* buffer, size_t size,
+                                         off_t offset) {
+  uint8_t* current_position = static_cast<uint8_t*>(buffer);
+  size_t bytes_received = 0;
```
dpallotti, line 118: `bytes_read`?

src/OrbitBase/File.cpp
```diff
@@ -94,4 +112,25 @@ ErrorMessageOr<size_t> ReadFully(const unique_fd& fd, void* buffer, size_t size)
   return size - bytes_left;
 }
 
+ErrorMessageOr<size_t> ReadFullyAtOffset(const unique_fd& fd, void* buffer, size_t size,
+                                         off_t offset) {
+  uint8_t* current_position = static_cast<uint8_t*>(buffer);
```
dpallotti, line 117: I find it inconsistent that `uint8_t` is used here but `char` is used in `WriteFullyAtOffset`.

src/OrbitBase/FileTest.cpp
```diff
@@ -83,6 +83,65 @@ TEST(File, OpenNewFileForReadWrite) {
   ASSERT_TRUE(fd_or_error.has_value()) << fd_or_error.error().message();
 }
 
+TEST(File, WriteFullySmoke) {
+  auto temporary_file_or_error = TemporaryFile::Create();
+  ASSERT_TRUE(temporary_file_or_error.has_value()) << temporary_file_or_error.error().message();
```
dpallotti, line 88: Shouldn't this be `CHECK` as it's setup (same in `WriteFullyAtOffsetSmoke`)?

src/OrbitBase/FileTest.cpp
```diff
@@ -83,6 +83,65 @@ TEST(File, OpenNewFileForReadWrite) {
...
+  // Read back and compare content.
+  ErrorMessageOr<unique_fd> fd_or_error = OpenFileForReading(temporary_file.file_path().string());
+  ASSERT_FALSE(fd_or_error.has_error()) << fd_or_error.error().message();
+  std::array<char, 64> read_back = {};
+  ErrorMessageOr<size_t> result_or_error = ReadFully(fd_or_error.value(), read_back.data(), 9);
+  ASSERT_FALSE(result_or_error.has_error()) << result_or_error.error().message();
+  EXPECT_EQ(result_or_error.value(), 9);
+  EXPECT_STREQ(read_back.data(), buffer.c_str());
+
+  temporary_file.CloseAndRemove();
```
dpallotti, line 105: The destructor already does this, right (same in `WriteFullyAtOffsetSmoke`)?

src/OrbitBase/FileTest.cpp
```diff
@@ -111,4 +170,51 @@ TEST(File, ReadFullySmoke) {
   EXPECT_STREQ(buf.data(), "");
 }
 
+TEST(File, ReadFullyAtOffsetSmoke) {
+  const auto fd_or_error =
+      OpenFileForReading(GetExecutableDir() / "testdata" / "OrbitBase" / "textfile.bin");
+  ASSERT_FALSE(fd_or_error.has_error()) << fd_or_error.error().message();
```
dpallotti, line 176: `CHECK`?

src/UserSpaceInstrumentation/AccessTraceesMemory.cpp
```diff
@@ -25,44 +26,37 @@ using orbit_base::ReadFileToString;
...
-      return ErrorMessage(
-          absl::StrFormat("Failed to PTRACE_PEEKDATA for pid %d with errno %d: \"%s\"", pid, errno,
-                          SafeStrerror(errno)));
-    }
-    std::memcpy(bytes.data() + (i * sizeof(uint64_t)), &data, 8);
+  OUTCOME_TRY(result, ReadFullyAtOffset(fd, bytes.data(), length, address_start));
+
+  if (result < length) {
+    return ErrorMessage(absl::StrFormat(
+        "Failed to read %u bytes from memory file of process pid %d. Only got %d bytes.", length,
```
dpallotti, line 38: `process %d` or `process with pid %d`

dpallotti (approved):

src/OrbitBase/File.cpp
danielfenner: Hmm, is this really an error?

_The number of bytes written may be less than count if, for example, there is insufficient space on the underlying physical medium, or the RLIMIT_FSIZE resource limit is encountered (see setrlimit(2)), or the call was interrupted by a signal handler after having written less than count bytes. (See also pipe(7).)_

If pwrite gets interrupted with zero bytes read we should retry, right?

src/OrbitBase/File.cpp
danielfenner: Well, it doesn't matter, does it? Changed anyways.

src/OrbitBase/FileTest.cpp
danielfenner: Reading back buffer.size()+3 and check the result for size buffer.size() to check that nothing got written beyond what was intended.

src/OrbitBase/include/OrbitBase/File.h
danielfenner: done

src/OrbitBase/File.cpp
dimitry-: hmm, I assume that for write if 0 bytes are written it returns -1 with EITNR, at least this is what write does

_EINTR  The call was interrupted by a signal before any data was written; see signal(7)._

so it can never be 0.

dimitry-: and EINTR is handled by TEMP_FAILURE_RETRY macro.

src/UserSpaceInstrumentation/AccessTraceesMemory.cpp
danielfenner: I thought the same but started liking OUTCOME_TRY for keeping the code size under control. In many cases the error messages are good enough and it is reasonably clear what is going on. The caller still has the information that they called a "ReadTraceesMemory" and "Permission denied." is not the worst response there.

src/UserSpaceInstrumentation/AccessTraceesMemoryTest.cpp
danielfenner: done

src/UserSpaceInstrumentation/AccessTraceesMemoryTest.cpp
danielfenner: done

src/UserSpaceInstrumentation/AccessTraceesMemoryTest.cpp
danielfenner: done

src/UserSpaceInstrumentation/AccessTraceesMemoryTest.cpp
danielfenner: I think the right way to get this into gtest would be to write a matchers to enable something like this:
```
EXPECT_THAT(result, ::testing::HasValue())
EXPECT_THAT(result, ::testing::HasError())
ASSERT_THAT(result, ::testing::HasErrorContaining("permission"))
```
I dislike the `<< result.error().message()` I added for its unneccessary verbosity.

Not in this PR thought: http://b/185464948

src/OrbitBase/File.cpp
danielfenner: done

src/OrbitBase/File.cpp
danielfenner: Yes, I just copied that from the ReadFully / WriteFully methods. 

A much worse inconsistency is that WriteFully takes a std::string_view as an input and the other methods taks `void*`. I guess a absl::Span would be the right solution but I haven't put any thought into this yet.

src/OrbitBase/FileTest.cpp
danielfenner: done

src/OrbitBase/FileTest.cpp
danielfenner: done

src/OrbitBase/FileTest.cpp
danielfenner: done

src/UserSpaceInstrumentation/AccessTraceesMemory.cpp
danielfenner: done

src/UserSpaceInstrumentation/AccessTraceesMemory.cpp
danielfenner: see above

src/UserSpaceInstrumentation/AccessTraceesMemoryTest.cpp
danielfenner: done

src/UserSpaceInstrumentation/AccessTraceesMemoryTest.cpp
danielfenner: done

src/UserSpaceInstrumentation/AccessTraceesMemoryTest.cpp
danielfenner: done

src/OrbitBase/File.cpp
danielfenner: > EINTR The call was interrupted by a signal before any data was written; see signal(7).

Okay, then the Check should be save. Added. Done.

src/UserSpaceInstrumentation/AccessTraceesMemory.cpp
dimitry-: the missing part here of course if that it was trying to read from /proc/pid/mem.

src/OrbitBase/File.cpp
```diff
@@ -78,6 +78,9 @@ ErrorMessageOr<void> WriteFully(const unique_fd& fd, std::string_view content) {
 
 ErrorMessageOr<void> WriteFullyAtOffset(const unique_fd& fd, const void* buffer, size_t size,
                                         off_t offset) {
+#if defined(_WIN32)
+  FATAL("Not implemented on Windows.")
```
dimitry-, line 82: maybe use lseek/read? should be compatible with windows in this case.

danielfenner: Oh, right. The unit test for the file stuff is also running on windows. I'll see if there is a simple solution.

pierricgimmig (approved):

src/OrbitBase/File.cpp
danielfenner: will leave it as is for now

src/OrbitBase/File.cpp
danielfenner: see pread/pwrite implementation for WIN32

PR #2174, opened 2021-04-13
Add the MemoryTrack class for memory tracks

vickyliu-go4it:
With this change, we add the `MemoryTrack` class such that the memory tracks can be treated separately from the general `GraphTrack`.

More specifically, the setting and drawing of `warning_threshold_`, `value_upper_bound_` and `value_lower_bound_` are moved to the `MemoryTrack`, while the setting of `label_unit_` and `value_decimal_digits_` (i.e., label value digits) still remain in the `GraphTrack`.

Bug: http://b/185204557
Test: take a capture with memory tracks

Follow up tasks can be found in [Memory feature task list](https://docs.google.com/spreadsheets/d/1PdYLwV4tvEDghWQP1jJDdOIUXaDAbczbh6p-VVebINw/edit#gid=894146394).

### DIFF
--- a/src/OrbitGl/MemoryTrack.cpp
+++ b/src/OrbitGl/MemoryTrack.cpp
@@ -1,0 +1,96 @@
+// Copyright (c) 2021 The Orbit Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#include "MemoryTrack.h"
+
+#include <algorithm>
+
+#include "GlCanvas.h"
+
+namespace orbit_gl {
+
+void MemoryTrack::Draw(GlCanvas* canvas, PickingMode picking_mode, float z_offset) {
+  GraphTrack::Draw(canvas, picking_mode, z_offset);
+
+  if (values_.empty() || picking_mode != PickingMode::kNone) return;
+
+  float text_z = GlCanvas::kZValueTrackText + z_offset;
+  Batcher* ui_batcher = canvas->GetBatcher();
+  uint32_t font_size = layout_->CalculateZoomedFontSize();
+
+  // Add warning threshold text box and line.
+  if (warning_threshold_.has_value()) {
+    const Color kThresholdColor(244, 67, 54, 255);
+    double normalized_value = (warning_threshold_.value().second - min_) * inv_value_range_;
+    float x = pos_[0];
+    float y = pos_[1] - size_[1] + static_cast<float>(normalized_value) * size_[1];
+    Vec2 from(x, y);
+    Vec2 to(x + size_[0], y);
+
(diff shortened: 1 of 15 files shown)


karupayun (approved):
LGTM Tingwei!! Thanks for this refactor :)

src/OrbitGl/MemoryTrack.cpp
```diff
@@ -0,0 +1,92 @@
...
+#include <algorithm>
+
+#include "GlCanvas.h"
+
+void MemoryTrack::Draw(GlCanvas* canvas, PickingMode picking_mode, float z_offset) {
+  GraphTrack::Draw(canvas, picking_mode, z_offset);
+
+  if (values_.empty() || picking_mode != PickingMode::kNone) return;
+
+  const Color kWhite(255, 255, 255, 255);
```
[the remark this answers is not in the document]
vickyliu-go4it: SGTM! Thanks for the suggestion!

src/OrbitGl/MemoryTrack.h
```diff
@@ -0,0 +1,35 @@
...
+
+#ifndef ORBIT_GL_MEMORY_TRACK_H_
+#define ORBIT_GL_MEMORY_TRACK_H_
+
+#include <string>
+
+#include "GraphTrack.h"
+#include "Track.h"
+
+class MemoryTrack final : public GraphTrack {
```
antonrohr, line 13: Since your are doing this new class (which is awesome). Can you put this in namespace `orbit_gl`? It is currently somewhat annoying, because a few files in OrbitGl have the namespace and lots of them don't, but I think its the right direction for the future.

src/OrbitGl/Track.h
```diff
@@ -37,6 +37,7 @@ class Track : public orbit_gl::CaptureViewElement, public std::enable_shared_fro
     kGpuTrack,
     kSchedulerTrack,
     kAsyncTrack,
+    kMemoryTrack,
```
antonrohr, line 40: Would you do me a favor and change this from `enum Type` to `enum class Type`. Since you are not technically touching this, feel free to dismiss. But I think its a fairly small change, that could be its own commit and would help a lot.

antonrohr (approved):
lgtm, thanks for improving this code

src/OrbitGl/MemoryTrack.h
vickyliu-go4it: Make sense! Changed accordingly. Thanks!

src/OrbitGl/Track.h
vickyliu-go4it: I'm glad to do this. Thanks for the suggestion!

PR #2175, opened 2021-04-13
Metric symbol loading

antonrohr:


src/OrbitGl/App.cpp
```diff
@@ -1222,18 +1222,33 @@ orbit_base::Future<ErrorMessageOr<void>> OrbitApp::RetrieveModuleAndLoadSymbols(
...
 
-  return orbit_base::UnwrapFuture(
+  orbit_base::Future<ErrorMessageOr<void>> load_result = orbit_base::UnwrapFuture(
       RetrieveModule(module_path, build_id)
           .ThenIfSuccess(main_thread_executor_, [this, module_path, build_id](
                                                     const std::filesystem::path& local_file_path) {
             return LoadSymbols(local_file_path, module_path, build_id);
           }));
+
+  return load_result.Then(main_thread_executor_,
```
beckerhe, line 1243: Nit: To me it would be more natural to split the data flow, instead of chaining it. So you could call `Future::Then` on `load_result` and still return `load_result`.

beckerhe: But of course, the current solution is also fine!

karupayun (approved):
LGTM!!

beckerhe (approved):
LGTM!

src/OrbitGl/App.cpp
antonrohr: This is a good idea! I will change it

PR #2176, opened 2021-04-13
Stable track order for Frame Tracks

karupayun:
This PR makes the track order stable after adding and removing FrameTracks.
They are always sorted now by function_id (the user is able to modify it
later). This implied a change from hash map to a regular map.

Frame Track now are inserted just after the last track which should be
before them (the scheduler itself or a FrameTrack with lower function_id).

Also we rename a method (UpdateVisibleTrackList) and we replace some
std-containers for absl's ones.

Solved http://b/175534459 and http://b/185224990.

Test: Load a capture, add/remove a Frame Track. Also tested with running
captures.


dimitry- (approved):

ronaldfw (approved):
LGTM, thanks

PR #2177, opened 2021-04-13
Add LinuxTracingIntegrationTest.GpuJobs, relying on a simplified version of the Vulkan tutorial

dpallotti:
#### Add VulkanTutorial module
This contains `OffscreenRenderingVulkanTutorialLib`, a simplified
implementation of the Vulkan Tutorial triangle that renders to memory instead of
to a surface. This will be used, possibly expanded, for GPU-related tests,
starting with `LinuxTracingIntegrationTests`.
It also contains `OffscreenRenderingVulkanTutorial` with a `main` that allows to
run it.

Bug: http://b/182793342

Test: Run `OffscreenRenderingVulkanTutorial`, take a capture, see the expected
GPU jobs in the capture.

#### Add LinuxTracingIntegrationTest.GpuJobs
Bug: http://b/182793342


dpallotti:
Some files don't have our usual license header. I need some guidance about that. For example, for the two .h files without it, should I just add it?

third_party/conan/lockfiles/base.lock
```diff
@@ -320,364 +321,370 @@
     "context": "host"
    },
    "31": {
-    "ref": "crashpad/20200624@orbitdeps/stable#8c19cb575eb819de0b050cf7d1f317b6",
+    "ref": "volk/1.2.170#faf80e25f935df05ffb094e31b6cedf7",
```
florian-kuebler, line 324: Haven't looked at the actual PR, but this is Vulkan, right?
Shouldn't it be also 1.1.114? Or should we move the headers to 1.2.17?

dpallotti: Yes and no, see https://conan.io/center/volk and https://github.com/zeux/volk.
I guess that the version of volk is supposed to match the version of Vulkan, although I don't know if it's a problem.
But there's no volk 1.1.114. Maybe we should update the headers then?

beckerhe:
> Some files don't have our usual license header. I need some guidance about that. For example, for the two .h files without it, should I just add it?

Do we have the copyright for all the vulkan tutorial files? Or are they from some public source?

third_party/conan/lockfiles/base.lock
beckerhe: volk is backwards compatible, meaning volk/1.2.170 works with vulkan-headers up to version 1.2.170.0, but not the other way around.

dpallotti:
> > Some files don't have our usual license header. I need some guidance about that. For example, for the two .h files without it, should I just add it?
> 
> Do we have the copyright for all the vulkan tutorial files? Or are they from some public source?

I'm not too sure how to answer so I'll just give you some misc information. The Vulkan tutorial is here: https://vulkan-tutorial.com/. The code of the tutorial is here https://vulkan-tutorial.com/code/15_hello_triangle.cpp although the vast majority of the code in this change comes from following the tutorial itself (I assumed this is not a problem as hello_ggp is also based on this tutorial).

third_party/conan/lockfiles/base.lock
dpallotti: So we're fine as it is?

beckerhe: I think we are, yes!

beckerhe:
> > > Some files don't have our usual license header. I need some guidance about that. For example, for the two .h files without it, should I just add it?
> > 
> > 
> > Do we have the copyright for all the vulkan tutorial files? Or are they from some public source?
> 
> I'm not too sure how to answer so I'll just give you some misc information. The Vulkan tutorial is here: https://vulkan-tutorial.com/. The code of the tutorial is here https://vulkan-tutorial.com/code/15_hello_triangle.cpp although the vast majority of the code in this change comes from following the tutorial itself (I assumed this is not a problem as hello_ggp is also based on this tutorial).

Ah ok. So we can't put our copyright header on files that aren't completely written by us. So if some of your files contain code form the vulkan tutorial, they need to live under `third_party/` and may not have our copyright header. So that's not really a big deal, as you can just move the files there and reference in your CMakeLists.txt. But you also need a license file for these files.

By the way, what is the license of this code? We also need to figure out if it compatible to Google policies.

dpallotti:
> > > > Some files don't have our usual license header. I need some guidance about that. For example, for the two .h files without it, should I just add it?
> > > 
> > > 
> > > Do we have the copyright for all the vulkan tutorial files? Or are they from some public source?
> > 
> > 
> > I'm not too sure how to answer so I'll just give you some misc information. The Vulkan tutorial is here: https://vulkan-tutorial.com/. The code of the tutorial is here https://vulkan-tutorial.com/code/15_hello_triangle.cpp although the vast majority of the code in this change comes from following the tutorial itself (I assumed this is not a problem as hello_ggp is also based on this tutorial).
> 
> Ah ok. So we can't put our copyright header on files that aren't completely written by us. So if some of your files contain code form the vulkan tutorial, they need to live under `third_party/` and may not have our copyright header. So that's not really a big deal, as you can just move the files there and reference in your CMakeLists.txt. But you also need a license file for these files.
> 
> By the way, what is the license of this code? We also need to figure out if it compatible to Google policies.

It seems that I underestimated this quite a bit.

https://github.com/Overv/VulkanTutorial#license says 
"The code listings in the code directory are licensed as [CC0 1.0 Universal](https://creativecommons.org/publicdomain/zero/1.0/). By contributing to that directory, you agree to license your contributions to the public under that same public domain-like license."

beckerhe:
> > > > > Some files don't have our usual license header. I need some guidance about that. For example, for the two .h files without it, should I just add it?
> > > > 
> > > > 
> > > > Do we have the copyright for all the vulkan tutorial files? Or are they from some public source?
> > > 
> > > 
> > > I'm not too sure how to answer so I'll just give you some misc information. The Vulkan tutorial is here: https://vulkan-tutorial.com/. The code of the tutorial is here https://vulkan-tutorial.com/code/15_hello_triangle.cpp although the vast majority of the code in this change comes from following the tutorial itself (I assumed this is not a problem as hello_ggp is also based on this tutorial).
> > 
> > 
> > Ah ok. So we can't put our copyright header on files that aren't completely written by us. So if some of your files contain code form the vulkan tutorial, they need to live under `third_party/` and may not have our copyright header. So that's not really a big deal, as you can just move the files there and reference in your CMakeLists.txt. But you also need a license file for these files.
> > By the way, what is the license of this code? We also need to figure out if it compatible to Google policies.
> 
> It seems that I underestimated this quite a bit.
> 
> https://github.com/Overv/VulkanTutorial#license says
> "The code listings in the code directory are licensed as [CC0 1.0 Universal](https://creativecommons.org/publicdomain/zero/1.0/). By contributing to that directory, you agree to license your contributions to the public under that same public domain-like license."

CC0 is not ideal - to say the least. As far as I know it is not allowed for Google-internal stuff or might at least require a review. Not sure if the same policy apply for an open source project like Orbit, especially since we only need it for testing.

dpallotti:
> > > > > > Some files don't have our usual license header. I need some guidance about that. For example, for the two .h files without it, should I just add it?
> > > > > 
> > > > > 
> > > > > Do we have the copyright for all the vulkan tutorial files? Or are they from some public source?
> > > > 
> > > > 
> > > > I'm not too sure how to answer so I'll just give you some misc information. The Vulkan tutorial is here: https://vulkan-tutorial.com/. The code of the tutorial is here https://vulkan-tutorial.com/code/15_hello_triangle.cpp although the vast majority of the code in this change comes from following the tutorial itself (I assumed this is not a problem as hello_ggp is also based on this tutorial).
> > > 
> > > 
> > > Ah ok. So we can't put our copyright header on files that aren't completely written by us. So if some of your files contain code form the vulkan tutorial, they need to live under `third_party/` and may not have our copyright header. So that's not really a big deal, as you can just move the files there and reference in your CMakeLists.txt. But you also need a license file for these files.
> > > By the way, what is the license of this code? We also need to figure out if it compatible to Google policies.
> > 
> > 
> > It seems that I underestimated this quite a bit.
> > https://github.com/Overv/VulkanTutorial#license says
> > "The code listings in the code directory are licensed as [CC0 1.0 Universal](https://creativecommons.org/publicdomain/zero/1.0/). By contributing to that directory, you agree to license your contributions to the public under that same public domain-like license."
> 
> CC0 is not ideal - to say the least. As far as I know it is not allowed for Google-internal stuff or might at least require a review. Not sure if the same policy apply for an open source project like Orbit, especially since we only need it for testing.
> 
> EDIT: I checked again and found also claims that CC0 is okay.

As discussed offline, I moved most of the code from the `VulkanTutorial` module to `third_party`, so that we stay on the safe side.

beckerhe (approved):
LGTM

I just started building packages for volk. Will still take a couple of minutes. After that it's good to go!

third_party/VulkanTutorial/include/VulkanTutorial/OffscreenRenderingVulkanTutorial.h
```diff
@@ -0,0 +1,83 @@
...
+  VkRenderPass render_pass_ = VK_NULL_HANDLE;
+  VkPipelineLayout pipeline_layout_ = VK_NULL_HANDLE;
+  VkPipeline graphics_pipeline_ = VK_NULL_HANDLE;
+  VkFramebuffer framebuffer_ = VK_NULL_HANDLE;
+  VkCommandPool command_pool_ = VK_NULL_HANDLE;
+  VkCommandBuffer command_buffer_ = VK_NULL_HANDLE;
+  VkFence fence_ = VK_NULL_HANDLE;
+
+  absl::Mutex stop_requested_mutex_;
+  bool stop_requested_ = false;
```
florian-kuebler, line 78: Could you add the GUARDED_BY annotation?
https://abseil.io/docs/cpp/guides/synchronization#thread-annotations

src/VulkanTutorial/OffscreenRenderingVulkanTutorialMain.cpp
```diff
@@ -0,0 +1,46 @@
...
+namespace {
+std::atomic<bool> exit_requested = false;
+
+void SigintHandler(int signum) {
+  if (signum == SIGINT) {
+    exit_requested = true;
+  }
+}
+
+void InstallSigintHandler() {
```
florian-kuebler, line 20: Nit: It would be nice to have a brief comment here, what this does/why this is needed.

src/LinuxTracingIntegrationTests/LinuxTracingIntegrationTest.cpp
```diff
@@ -1049,5 +1068,68 @@ TEST(LinuxTracingIntegrationTest, ModuleUpdateOnDlopen) {
...
+    }
+
+    // The puppet is single-threaded.
+    EXPECT_EQ(gpu_job.tid(), fixture.GetPuppetPid());
+
+    EXPECT_LT(gpu_job.amdgpu_cs_ioctl_time_ns(), gpu_job.amdgpu_sched_run_job_time_ns());
+    // If no other job is running on the GPU (which is the case if the puppet is the only process
+    // using the GPU), then we assume (it's the best we can do) that the job starts running on the
+    // hardware at the same time as it is scheduled by the driver, hence the EXPECT_EQ.
+    // Otherwise, use EXPECT_LE.
```
florian-kuebler, line 1117: Are we sure about this?

third_party/VulkanTutorial/OffscreenRenderingVulkanTutorial.cpp
```diff
@@ -0,0 +1,575 @@
...
+}
+
+void OffscreenRenderingVulkanTutorial::CreateInstance() {
+  VkApplicationInfo app_info{
+      .sType = VK_STRUCTURE_TYPE_APPLICATION_INFO,
+      .pApplicationName = "VulkanTutorial",
+      .applicationVersion = VK_MAKE_VERSION(1, 0, 0),
+      .pEngineName = "",
+      .engineVersion = VK_MAKE_VERSION(1, 0, 0),
+      .apiVersion = VK_API_VERSION_1_0,
```
florian-kuebler, line 99: Should this be 1.1, as our headers are that version?

third_party/VulkanTutorial/OffscreenRenderingVulkanTutorial.cpp
```diff
@@ -0,0 +1,575 @@
...
+}
+
+bool OffscreenRenderingVulkanTutorial::AreValidationLayersSupported() {
+  uint32_t layer_count = 0;
+  CHECK_VK_SUCCESS(vkEnumerateInstanceLayerProperties(&layer_count, nullptr));
+
+  std::vector<VkLayerProperties> available_layers(layer_count);
+  CHECK_VK_SUCCESS(vkEnumerateInstanceLayerProperties(&layer_count, available_layers.data()));
+
+  for (const char* layer_name : kValidationLayerNames) {
```
florian-kuebler, line 125: Theoretically, you might also want to check for the version.

florian-kuebler (approved):
LGTM, thanks!

src/VulkanTutorial/OffscreenRenderingVulkanTutorialMain.cpp
dpallotti: Done.

third_party/VulkanTutorial/include/VulkanTutorial/OffscreenRenderingVulkanTutorial.h
dpallotti: Done.

third_party/VulkanTutorial/OffscreenRenderingVulkanTutorial.cpp
dpallotti: I guess it makes sense, done.

src/LinuxTracingIntegrationTests/LinuxTracingIntegrationTest.cpp
dpallotti: Are you referring to a specific part?
The idea here is that the tutorial is always only running one job at a time (offscreen rendering might be playing a role in this, I'm not sure how the full tutorial would behave). 
When no other job is running on the gpu, we assume it runs immediately after being scheduled and the middle time interval has 0 duration.
We check if we can make this assumption for every single job by seeing if there were no jobs from other processes while we were tracing.
Do you see some part that could be problematic in this?

third_party/VulkanTutorial/OffscreenRenderingVulkanTutorial.cpp
dpallotti: Do you know what I should check exactly? That `layer_properties.specVersion >= VK_API_VERSION_1_1`?

src/LinuxTracingIntegrationTests/LinuxTracingIntegrationTest.cpp
florian-kuebler: >we assume it runs immediately after being scheduled and the middle time interval has 0 duration.

I am not sure. I was basically wondering about the confidence that this is true. Couldn't there be a super tiny amount of time until the second trace point gets hidden, even if there is nothing else on the gpu?

third_party/VulkanTutorial/OffscreenRenderingVulkanTutorial.cpp
florian-kuebler: For your concrete layer there is the `VK_EXT_VALIDATION_FLAGS_SPEC_VERSION`. But I am not 100% sure if this is really suggested to check for layers. 

For extensions there is something similar, and there it is suggested.

src/LinuxTracingIntegrationTests/LinuxTracingIntegrationTest.cpp
dpallotti: No no, this is our own assumption, that we make in `GpuTracepointVisitor.cpp`. The problem here is that we have three tracepoints for four time points, so this is the best we can do.

From the code:
```
  // We assume that GPU jobs (command buffer submissions) immediately
  // start running on the hardware when they are scheduled by the
  // driver (this is the best we can do), *unless* there is already a
  // job running. We keep track of when jobs finish in
  // timeline_to_latest_dma_signal_. If a previous job is still running
  // at the timestamp of scheduling the current job, we push the start
  // time for starting on the hardware back.
```
and
```
  // We do not have an explicit event for the following timestamp. We
  // assume that, when the GPU queue corresponding to timeline is
  // not executing a job, that this job starts exactly when it is
  // scheduled by the driver. Otherwise, we assume it starts exactly
  // when the previous job has signaled that it is done. Since we do
  // not have an explicit signal here, this is the best we can do.
```

florian-kuebler: Ahhh. Thanks for the clarification.

dpallotti: In other words, the first, second, and fourth time point come from a tracepoint, while the third we have to assume based on what else is happening on the gpu. When nothing else is running on the gpu, we assume that the second and third time points coincide.

third_party/VulkanTutorial/OffscreenRenderingVulkanTutorial.cpp
dpallotti: As discussed offline, let's leave as it is.

PR #2178, opened 2021-04-14
Ground work for injecting liborbit.so

pierricgimmig:
Instead of dynamically loading liborbit.so on the first api call,
Orbit will automatically inject the library when needed and patch
a global function table which will enable the API in the tracee.

This PR focuses on Api, OrbitBase and OrbitTest. Note that in its
current state, it breaks manual instrumentation in OrbitTest. This
is temporary as other PRs are in the pipeline. See PR #2140 for
end-to-end liborbit.so injection.

### DIFF
--- a/src/Api/CMakeLists.txt
+++ b/src/Api/CMakeLists.txt
@@ -34,6 +34,24 @@ target_link_libraries(Api PUBLIC
         OrbitBase
         OrbitProducer
         ProducerSideChannel)
+
+
+add_library(ApiLib STATIC)
+
+target_compile_options(ApiLib PRIVATE ${STRICT_COMPILE_FLAGS})
+
+target_compile_features(ApiLib PUBLIC cxx_std_17)
+
+target_sources(ApiLib PUBLIC
+        include/Api/InitializeInTracee.h)
+
+target_sources(ApiLib PRIVATE
+        InitializeInTracee.cpp)
+
+target_link_libraries(ApiLib PUBLIC
+        ApiInterface
+        OrbitBase
+        UserSpaceInstrumentation)
 endif()
 
 
--- a/src/Api/InitializeInTracee.cpp
+++ b/src/Api/InitializeInTracee.cpp
@@ -11,9 +11,8 @@ using orbit_grpc_protos::CaptureOptions;
 
 namespace orbit_api {
 
-ErrorMessageOr<void> InitializeInTracee(const CaptureOptions& capture_options) {
+ErrorMessageOr<void> InitializeInTracee(const CaptureOptions& /*capture_options*/) {
   // Dummy implementation until "ExecuteInProcess" functionality makes it into main.
-  (void)capture_options;
   return outcome::success();
 }
 
--- a/src/Api/InitializeInTracee.cpp
+++ b/src/Api/InitializeInTracee.cpp
@@ -1,0 +1,20 @@
+// Copyright (c) 2021 The Orbit Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#include "Api/InitializeInTracee.h"
+
+#include "Api/Orbit.h"
+#include "OrbitBase/Result.h"
+
+using orbit_grpc_protos::CaptureOptions;
+
+namespace orbit_api {
+
+ErrorMessageOr<void> InitializeInTracee(const CaptureOptions& capture_options) {
+  // Dummy implementation until "ExecuteInProcess" functionality makes it into main.
+  (void)capture_options;
+  return outcome::success();
+}
+
+}  // namespace orbit_api
--- a/src/Api/Orbit.cpp
+++ b/src/Api/Orbit.cpp
@@ -5,11 +5,12 @@
 #define ORBIT_API_INTERNAL_IMPL
 #include "Api/Orbit.h"
 
+#include <absl/base/casts.h>
+
 #include "Api/EncodedEvent.h"
 #include "Api/LockFreeApiEventProducer.h"
 #include "OrbitBase/Profiling.h"
 #include "OrbitBase/ThreadUtils.h"
-#include "absl/base/casts.h"
 
 static void EnqueueApiEvent(orbit_api::EventType type, const char* name = nullptr,
                             uint64_t data = 0, orbit_api_color color = kOrbitColorAuto) {
@@ -80,16 +81,11 @@ void orbit_api_async_string(const char* str, uint64_t id, orbit_api_color color)
 static void orbit_api_initialize_v0(uint64_t address) {
   // The api function table is accessed by user code using this pattern:
   //
+  // std::atomic_thread_fence(std::memory_order_acquire)
   // if(api->active && api->orbit_api_function_name) api->orbit_api_function_name(...);
   //
-  // We don't use locking nor memory barriers here for these reasons:
-  // 1. We rely on the fact that 64 bit read/writes on aligned uint64_t are atomic on x64 so we
-  //    assume there cannot be torn reads/writes on a function address in the table.
-  // 2. Nothing bad happens if "active" is 0 but other threads see non-null functions pointers.
-  // 3. Nothing bad happens if "active" is 1 but other threads see a null function pointer.
-  //
-  // If these assumptions became false, then we should revisit this mechanism both in this function
-  // and in the user code.
+  // We use acquire and release semantics to make sure that when api->active is set, all the
+  // function pointers have been assigned and are visible to other cores.
 
   orbit_api_v0* api = absl::bit_cast<orbit_api_v0*>(address);
   if (api->initialized != 0) return;
@@ -104,6 +100,7 @@ static void orbit_api_initialize_v0(uint64_t address) {
   api->track_uint64 = &orbit_api_track_uint64;
   api->track_float = &orbit_api_track_float;
   api->track_double = &orbit_api_track_double;
+  std::atomic_thread_fence(std::memory_order_release);
   api->initialized = 1;
   api->active = 1;
 }
--- a/src/Api/Orbit.cpp
+++ b/src/Api/Orbit.cpp
@@ -9,6 +9,7 @@
 #include "Api/LockFreeApiEventProducer.h"
 #include "OrbitBase/Profiling.h"
 #include "OrbitBase/ThreadUtils.h"
+#include "absl/base/casts.h"
 
 static void EnqueueApiEvent(orbit_api::EventType type, const char* name = nullptr,
                             uint64_t data = 0, orbit_api_color color = kOrbitColorAuto) {
@@ -74,5 +75,48 @@ void orbit_api_async_string(const char* str, uint64_t id, orbit_api_color color)
     EnqueueApiEvent(orbit_api::EventType::kString, buffer, id, color);
     str += chunk_size;
   }
+}
+
+static void orbit_api_initialize_v0(uint64_t address) {
+  // The api function table is accessed by user code using this pattern:
+  //
+  // if(api->active && api->orbit_api_function_name) api->orbit_api_function_name(...);
+  //
+  // We don't use locking nor memory barriers here for these reasons:
+  // 1. We rely on the fact that 64 bit read/writes on aligned uint64_t are atomic on x64 so we
+  //    assume there cannot be torn reads/writes on a function address in the table.
+  // 2. Nothing bad happens if "active" is 0 but other threads see non-null functions pointers.
+  // 3. Nothing bad happens if "active" is 1 but other threads see a null function pointer.
+  //
+  // If these assumptions became false, then we should revisit this mechanism both in this function
+  // and in the user code.
+
+  orbit_api_v0* api = absl::bit_cast<orbit_api_v0*>(address);
+  if (api->initialized != 0) return;
+  api->start = &orbit_api_start;
+  api->stop = &orbit_api_stop;
+  api->start_async = &orbit_api_start_async;
+  api->stop_async = &orbit_api_stop_async;
+  api->async_string = &orbit_api_async_string;
+  api->track_int = &orbit_api_track_int;
+  api->track_int64 = &orbit_api_track_int64;
+  api->track_uint = &orbit_api_track_uint;
+  api->track_uint64 = &orbit_api_track_uint64;
+  api->track_float = &orbit_api_track_float;
+  api->track_double = &orbit_api_track_double;
+  api->initialized = 1;
+  api->active = 1;
+}
+
+void orbit_api_initialize(uint64_t address, uint64_t api_version) {
+  LOG("Initializing Orbit API at address 0x%x, version %u", address, api_version);
+  if (api_version > kOrbitApiVersion) {
+    ERROR(
+        "Orbit API version in tracee (%u) is newer than the max supported version (%u). "
+        "Some features will be unavailable.",
+        api_version, kOrbitApiVersion);
+  }
+
+  if (api_version >= 0) orbit_api_initialize_v0(address);
 }
 }
--- a/src/Api/Orbit.cpp
+++ b/src/Api/Orbit.cpp
@@ -107,7 +107,7 @@ static void orbit_api_initialize_v0(uint64_t address) {
 }
 
 void orbit_api_initialize(uint64_t address, uint64_t api_version) {
-  LOG("Initializing Orbit API at address 0x%x, version %u", address, api_version);
+  LOG("Initializing Orbit API at address %#x, version %u", address, api_version);
   if (api_version > kOrbitApiVersion) {
     ERROR(
         "Orbit API version in tracee (%u) is newer than the max supported version (%u). "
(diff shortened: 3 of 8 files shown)


src/Api/InitializeInTracee.cpp
```diff
@@ -0,0 +1,20 @@
...
+#include "Api/InitializeInTracee.h"
+
+#include "Api/Orbit.h"
+#include "OrbitBase/Result.h"
+
+using orbit_grpc_protos::CaptureOptions;
+
+namespace orbit_api {
+
+ErrorMessageOr<void> InitializeInTracee(const CaptureOptions& capture_options) {
```
dimitry-, line 14: const CaptureOptions& /*capture_options*/

src/Api/InitializeInTracee.cpp
```diff
@@ -0,0 +1,20 @@
...
+  // Dummy implementation until "ExecuteInProcess" functionality makes it into main.
+  (void)capture_options;
```
dimitry-, line 16: In which case you can drop this line

src/Api/Orbit.cpp
```diff
@@ -9,6 +9,7 @@
 #include "Api/LockFreeApiEventProducer.h"
 #include "OrbitBase/Profiling.h"
 #include "OrbitBase/ThreadUtils.h"
+#include "absl/base/casts.h"
```
dimitry-, line 12: we use `<>` includes for all headers external to Orbit

src/OrbitTest/OrbitTest.cpp
```diff
@@ -24,7 +24,8 @@
 // removed once we split the introspection and manual instrumentation apis TODO(b/182848566).
 namespace orbit_api_wrapper {
 #include "Api/Orbit.h"
-}
+orbit_api_v0 g_orbit_api_v0;
```
dimitry-, line 27: does it work as expected if this is inside namespace? I guess compiler ignores namespace in this case because of extern "C" in the header. what does `readelf -d` say?

dimitry- (approved):

src/Api/CMakeLists.txt
```diff
@@ -34,6 +34,24 @@ target_link_libraries(Api PUBLIC
         OrbitBase
         OrbitProducer
         ProducerSideChannel)
+
+
+add_library(ApiLib STATIC)
```
dpallotti, line 39: I would try to find a more specific name for this target. The relationship with the `Api` is not clear from the name. Is it the target responsible for loading `Api` when the game is running? Then it could probably be called `ApiLoader`. In that case it should also be in a separate directory.

src/Api/include/Api/Orbit.h
```diff
@@ -31,32 +33,18 @@
...
-// On Linux/Stadia, our current dynamic instrumentation implementation, which relies on uprobes and
-// uretprobes, incur some non-negligible overhead (>5us per instrumented function call). Please
-// note that instrumenting too many functions will possibly cause some noticeable performance
-// degradation. Reducing overhead is our highest priority and we are actively working on a new
-// implementation that should be at least one order of magnitude faster.
-//
 // Integration:
-// To integrate the manual instrumentation API in your code base, simply include this header file.
+// To integrate the manual instrumentation API in your code base, simply include this header file
+// and instanciate the "g_orbit_api_v0" global variable in your code. Orbit will automatically
```
dpallotti, line 38:
```suggestion
// and instantiate the "g_orbit_api_v0" global variable in your code. Orbit will automatically
```
Much more common.

src/Api/include/Api/Orbit.h
```diff
@@ -269,144 +276,32 @@ typedef enum {
...
-inline void orbit_api_start_async(const char* name, uint64_t id, orbit_api_color color) {
-  static OrbitFunctor<void (*)(const char*, uint64_t, orbit_api_color)> f("orbit_api_start_async");
-  f(name, id, color);
-}
-
-inline void orbit_api_stop_async(uint64_t id) {
-  static OrbitFunctor<void (*)(uint64_t)> f("orbit_api_stop_async");
-  f(id);
-}
+// User-instanciated global variable.
```
dpallotti, line 297:
```suggestion
// User-instantiated global variable.
```
As above.

src/Api/Orbit.cpp
```diff
@@ -75,4 +76,47 @@ void orbit_api_async_string(const char* str, uint64_t id, orbit_api_color color)
...
+  api->track_uint = &orbit_api_track_uint;
+  api->track_uint64 = &orbit_api_track_uint64;
+  api->track_float = &orbit_api_track_float;
+  api->track_double = &orbit_api_track_double;
+  api->initialized = 1;
+  api->active = 1;
+}
+
+void orbit_api_initialize(uint64_t address, uint64_t api_version) {
+  LOG("Initializing Orbit API at address 0x%x, version %u", address, api_version);
```
dpallotti, line 112:
```suggestion
  LOG("Initializing Orbit API at address %#x, version %u", address, api_version);
```

src/Api/Orbit.cpp
```diff
@@ -75,4 +76,47 @@ void orbit_api_async_string(const char* str, uint64_t id, orbit_api_color color)
     str += chunk_size;
   }
 }
+
+static void orbit_api_initialize_v0(uint64_t address) {
+  // The api function table is accessed by user code using this pattern:
+  //
+  // if(api->active && api->orbit_api_function_name) api->orbit_api_function_name(...);
+  //
+  // We don't use locking nor memory barriers here for these reasons:
```
dpallotti, line 85: I have to say, I'm not very convinced, it feels like there should at least be atomics involved. But I've forgotten the little I had learned from this 3-hour talk https://www.youtube.com/watch?v=A8eCGOqgvH4 about the subtleties of this so I don't know.
But in any case: Isn't this code executed while the target is stopped? Or when only one thread is stopped? And does that make a difference?

dpallotti (approved):
If I get it correctly, the `InitializeInTracee` should be a different module.

I'm not too convinced about the assumptions on not needing any synchronization at all but I don't know enough.

src/Api/Orbit.cpp
ronaldfw: If I'm not mistaken, this is at least undefined behavior and should therefore be avoided. Is there a benefit to avoiding memory barriers here?

src/Api/Orbit.cpp
```diff
@@ -75,4 +76,47 @@ void orbit_api_async_string(const char* str, uint64_t id, orbit_api_color color)
...
+  api->track_int64 = &orbit_api_track_int64;
+  api->track_uint = &orbit_api_track_uint;
+  api->track_uint64 = &orbit_api_track_uint64;
+  api->track_float = &orbit_api_track_float;
+  api->track_double = &orbit_api_track_double;
+  api->initialized = 1;
+  api->active = 1;
+}
+
+void orbit_api_initialize(uint64_t address, uint64_t api_version) {
```
danielfenner, line 111: Not sure if I understand this correctly: 
`address` is a `orbit_api_v0*`, right? Why pass it around as a uint64_t?

src/Api/Orbit.cpp
danielfenner: > But in any case: Isn't this code executed while the target is stopped? Or when only one thread is stopped? 

All threads are stopped. Then one thread gets hijacked for doing the write. Then that thread gets restored. Then all threads are restarted.

>And does that make a difference?

I don't know.

I think the only thing we need to avoid is tearing and I think this does not occur when the data we write is eight byte aligned (no, I don't have a link to the place in the intel manual but I'd love to see one). I'm not so sure if stopping the threads helps with that. I think an unaligned access gets split into two accesses (internal to the processor, not by the compiler, the machine code is still only one instruction) and if the stop happens inbetween (is this possible?) we get tearing. All of this might be wrong!

danielfenner: [This looks related](https://docs.microsoft.com/en-us/windows/win32/dxtecharts/lockless-programming?redirectedfrom=MSDN#non-atomic-operations):

_On all modern processors, you can assume that reads and writes of naturally aligned native types are atomic. As long as the memory bus is at least as wide as the type being read or written, the CPU reads and writes these types in a single bus transaction, making it impossible for other threads to see them in a half-completed state. On x86 and x64 there, is no guarantee that reads and writes larger than eight bytes are atomic. This means that 16-byte reads and writes of streaming SIMD extension (SSE) registers, and string operations, might not be atomic._

_Reads and writes of types that are not naturally aligned—for instance, writing DWORDs that cross four-byte boundaries—are not guaranteed to be atomic. The CPU may have to do these reads and writes as multiple bus transactions, which could allow another thread to modify or see the data in the middle of the read or write._

So making sure the pointers are 8-byte aligned in the target should do the trick.

src/Api/InitializeInTracee.cpp
pierricgimmig: Done, thanks.

src/Api/InitializeInTracee.cpp
pierricgimmig: yep, thanks.

src/Api/Orbit.cpp
pierricgimmig: Done, thanks.

src/Api/Orbit.cpp
pierricgimmig: Ok so I revisited this to make sure that no one can access the function pointers until the "active" flag has been set.  To achieve this, we use acquire and release semantics in the form of memory fences. It's essentially what is described here (https://preshing.com/20130823/the-synchronizes-with-relation/). Note that since access to naturally aligned uint32_t is atomic, we don't need to make the guard variable a std::atomic<uint32_t> (this was verified with the author of the post).

See also https://preshing.com/20120913/acquire-and-release-semantics/

ronaldfw: I don't know what the code is as it's not yet updated with fences, but I'm still not sure about not using atomic on the boolean. In C++, a data race, e.g. when one thread writes to the bool and another reads at the same time, is undefined behavior (4.7.1 in the C++17 standard), unless I'm reading something wrong. That is, whether you get correct behavior here is not just a property of the underlying hardware, but also of the compiler (which in the case of undefined behavior may do what it wants). 

What's the benefit of not using an atomic? Performance?

dimitry-: std::atomic is equivalent to using load/store with memory_order_seq_cst, which is not necessary in this case, since we only want to make sure that initalized filag is read before function pointer. Note also that atomic_load/store will fail if the value passed is not properly aligned.

ronaldfw: But isn't there a potential concurrent write and read on the "initialized" or "active" boolean? (Not sure which is the one that is checked by the target process threads.)

dimitry-: I am sure they are - but atomic_load/store with acquire/release semantics is not much different from memory_order_seq_cst in that respect, or is it?

ronaldfw: I'm still not sure what the proposed code will look like. I think to be on the safe side, the code should be something like:

std::atomic<bool> active;

// On thread initializing the function pointers
api->start = &orbit_api_start;
[...]
active.store(true, std::memory_order_release);

// On threads reading the function pointers
if (active.load(std::memory_order_acquire)) {
  // use function pointer
}

pierricgimmig: > What's the benefit of not using an atomic? Performance?
Orbit.h is pure "C" so we don't get the std::atomic, but maybe there is the equivalent in C11? I'll look into it.

> std::atomic is equivalent to using load/store with memory_order_seq_cst
Well in this case we'd use memory_order_relaxed for the guard variable (active)

src/Api/CMakeLists.txt
pierricgimmig: Yeah good point, I like ApiLoader, I'll implement your suggestion.

src/Api/Orbit.cpp
ronaldfw: Ah, sorry, I didn't realize that Orbit.h was C only. I'll wait for the code update.

src/Api/Orbit.cpp
pierricgimmig: I should add more comments. "orbit_api_initialized" is (will be) called from OrbitService with the address of the function tables to patch. For a given process, there could be different tables with different api versions if different shared libraries use the api.

src/Api/Orbit.cpp
pierricgimmig: I pushed the change. Concerning the fact that Orbit.h is C, there is an additional annoyance that we can't include <stdatomic.h> from C++ (we get compilation errors: https://godbolt.org/z/3nqzM6he7, it seems to be a known issue). 

Since we use this guard variable as a flag, it can be either 0 or 1, even if the write was not atomic, I have a hard time imagining a scenario where that could be an issue. In our case, what is important is that by the time the flag is set to 1, all the function pointers are visible.

pierricgimmig: A non-atomic write would only delay the api enabling. Also, it's worth noting that those function pointers are written to once and never changed again.

ronaldfw: My (only) point here is that not using atomic on the guard flag is undefined behavior in C++ (and presumably in C11 as well?). But if we can't use https://en.cppreference.com/w/c/language/atomic due to the compilation error, I'm not sure what we can do.

dpallotti: Would it be possible and how complicated would it be to switch between the C++ atomics and the C atomics depending on whether it's C++ or C here?
With all the subtleties at compiler and CPU level (more about memory order than tearing) and the undefined behavior I think it would be cleaner and clearer to use atomics.

pierricgimmig: > Would it be possible and how complicated would it be to switch between the C++ atomics and the C atomics depending on whether it's C++ or C here?

Because we access this atomic from C++ in liborbit.so, we'd have to reinterpret the C atomic of Orbit.h as a C++ atomic, which seems risky. The fact that we can't include <stdatomic.h> in C++ is really annoying. I agree that it would be cleaner to simply have an atomic here, but we are dealing with an operation we know to be atomic on our system and pretty much every system:

"For example, it’s common knowledge that on all modern x86, x64, Itanium, SPARC, ARM and PowerPC processors, plain 32-bit integer assignment is atomic as long as the target variable is naturally aligned. You can verify it by consulting your processor manual and/or compiler documentation. In the games industry, I can tell you that a lot of 32-bit integer assignments rely on this particular guarantee." (https://preshing.com/20130618/atomic-vs-non-atomic-operations/#:~:text=For%20example%2C%20it's%20common%20knowledge,target%20variable%20is%20naturally%20aligned.)

Moreover, it seems that in our case, even if the write wasn't atomic, it wouldn't change the fact that whenever the guard variable's value became a potentially torn non-zero value, we'd be sure that the function table was fully visible (this is achieved with the memory barriers).

I'll keep thinking about how we could possibly share an atomic between C and C++, but so far I don't see how to do it cleanly. I'm definitely open to suggestions.

pierricgimmig: https://www.youtube.com/watch?v=A8eCGOqgvH4&t=4779s gives some hint about the fact that the guard variable can be an aligned int. Also, in https://preshing.com/20130922/acquire-and-release-fences/, we see `g_guard.store(1, std::memory_order_relaxed);` . The use of std::memory_order_relaxed here means that we only care about atomicity of the guard variable and not about memory ordering. Memory ordering is taken care of by the acquire and release fences.

I will submit the current implementation.

src/OrbitTest/OrbitTest.cpp
pierricgimmig: `readelf -d` doesn't return anything, but 'readelf -s' gives me:

`736: 0000000000068110    96 OBJECT  GLOBAL DEFAULT   29 g_orbit_api_v0
`

src/Api/CMakeLists.txt
pierricgimmig: Will do so in another PR shortly actually, b/185530194

src/Api/InitializeInTracee.cpp
```diff
@@ -0,0 +1,20 @@
...
+
+#include "Api/Orbit.h"
+#include "OrbitBase/Result.h"
+
+using orbit_grpc_protos::CaptureOptions;
+
+namespace orbit_api {
+
+ErrorMessageOr<void> InitializeInTracee(const CaptureOptions& capture_options) {
+  // Dummy implementation until "ExecuteInProcess" functionality makes it into main.
```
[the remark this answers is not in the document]
pierricgimmig: The PR is ready actually, I'll submit it shortly.

pierricgimmig:
Thanks to all for the comments. I will submit as so that I can submit subsequent PRs that rely on this ground work. This is by no means set in stone and I'll be happy to modify further if need be.

src/Api/Orbit.cpp
dpallotti: > The use of std::memory_order_relaxed here means that we only care about atomicity of the guard variable and not about memory ordering.

This is fair. As I wrote in my comment, my issue is generally not with possible tearing (thanks to the atomicity that comes from aligned writes), but about memory order, as this can cause incredibly unexpected issues. If we already take care of memory ordering, then I'm happy as long as it's properly documented.

dpallotti: > we'd have to reinterpret the C atomic of Orbit.h as a C++ atomic, which seems risky

Yeah yeah, this seems super risky, we should not do that.

PR #2179, opened 2021-04-14
Don't discard data symbols when parsing elf file

pierricgimmig:
In order to have access to global variables, don't discard data
symbols when parsing elf file.

For a bit of context:
This is a somewhat naive implementation. What I'd like to have
is the ability to query all modules for the presence of a particular
global variable, without populating all the debug information. This
will be used to discover the Orbit api function table in different 
modules so we can punch in the proper function pointers from 
Orbit.

See https://github.com/google/orbit/pull/2140 for full usage.

### DIFF
--- a/src/ElfUtils/ElfFile.cpp
+++ b/src/ElfUtils/ElfFile.cpp
@@ -5,6 +5,7 @@
 #include "ElfUtils/ElfFile.h"
 
 #include <absl/base/casts.h>
+#include <absl/strings/match.h>
 #include <absl/strings/str_cat.h>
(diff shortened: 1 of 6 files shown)


antonrohr (changes requested):
Looks generally good, but I would like some unit tests. The classes you touch here are some of the well unit tested ones, lets keep it this way

pierricgimmig:
@antonrohr , as mentioned, this was the wrong approach for the problem, sorry for not realizing that sooner. This change could still potentially be useful, but I'll close the PR for now since there is no immediate need for it. Thanks for the review.

PR #2181, opened 2021-04-15
Fix not showing timer's text on inactive tracks.

florian-kuebler:
A timer's text slice is currently only drawn on expanded tracks
(for ThreadTracks). However, tracks having just one depth of
timers are not "expanded", but "inactive". Thus, the check needs
to be "not collapsed".

Test: Take a capture with one function hooked.
Bug: http://b/185362423

### DIFF
--- a/src/OrbitGl/ThreadTrack.cpp
+++ b/src/OrbitGl/ThreadTrack.cpp
@@ -486,7 +486,7 @@ void ThreadTrack::UpdatePrimitives(Batcher* batcher, uint64_t min_tick, uint64_t
       const Vec2& size = text_box.GetSize();
 
       if (text_box.Duration() > draw_data.ns_per_pixel) {
-        if (collapse_toggle_->IsExpanded()) {
+        if (!collapse_toggle_->IsCollapsed()) {
           SetTimesliceText(text_box.GetTimerInfo(), draw_data.world_start_x, z_offset, &text_box);
         }
         batcher->AddShadedBox(pos, size, draw_data.z, color, std::move(user_data));


src/OrbitGl/ThreadTrack.cpp
```diff
@@ -486,7 +486,7 @@ void ThreadTrack::UpdatePrimitives(Batcher* batcher, uint64_t min_tick, uint64_t
       const Vec2& size = text_box.GetSize();
 
       if (text_box.Duration() > draw_data.ns_per_pixel) {
-        if (collapse_toggle_->IsExpanded()) {
+        if (!collapse_toggle_->IsCollapsed()) {
```
reichlfl, line 489: Should we (in a future PR) maybe change this behavior? I think it's reasonable to expect that `IsCollapsed() != IsExpanded` holds... we could only provide one of both methods, and `IsCollapsible()` in addition to deal with single-level tracks, WDYT? Could you create a bug for this?

reichlfl (approved):
LGTM, thanks!

src/OrbitGl/ThreadTrack.cpp
reichlfl: Reading your initial commit, maybe the solution should rather be to have this a tri-state Collapsed / Expanded / Inactive.

florian-kuebler: >Reading your initial commit, maybe the solution should rather be to have this a tri-state Collapsed / Expanded / Inactive.

This is how it is currently implemented. There is this tri-state and there are methods for IsCollapsed, IsExpanded, IsInactive. I simply was not aware of the third state. Not sure about a nice solution here. It is probably not too bad as it is now.

PR #2182, opened 2021-04-15
Fix crash when scheduling an event while shutting down the tracing li…

florian-kuebler:
…stener.

The global tracing listener has a field "active_" that is used
as a guard for scheduling events to its "thread_pool_". However,
in the tracing listener's destructor, the thread pool is first
shutdown, before the "active_" guard is set to false.
In this case, it could happen that while destructing the listener,
a new event would be scheduled to the pool, leading to a crash.

Test: Take a capture and while capturing start&stop introsepction.
Bug: http://b/185443567


beckerhe (approved):
As discussed offline, LGTM!

PR #2184, opened 2021-04-15
[Release] Remove the selection overlay when zooming in

reichlfl:
This was introduced with my Viewport changes: When using right-click / drag to zoom (hold CTRL), the green overlay should not be rendered after zooming is done.

Test: Hold Ctrl + Right-click drag, the green overlay disappears.
Bug: b/185440963


karupayun (approved):
LGTM!!

PR #2186, opened 2021-04-15
[RELEASE] Fix bug regarding ORBIT_EXIT scoped metric

antonrohr:
Before the ScopedMetric was not properly done, because of a
implementation mistake.


karupayun (approved):
LGTM Anton!!! Thanks for fixing this!

PR #2187, opened 2021-04-15
[REBASE TO MAIN] Fix crash related with wrongly used erase

karupayun:
We were always erasing one Frame Track, and this crashed when the Frame
Track function has no samples, because there is no FrameTrack. Now it is
fixed.

Fixed: http://b/185448604.


ronaldfw (approved):
LGTM, thanks!

PR #2194, opened 2021-04-15
Add command line prameters to log.

danielfenner:
This came up in the debug session.

Bug: n/a
Test: Started Orbit, works.

### DIFF
--- a/src/Orbit/main.cpp
+++ b/src/Orbit/main.cpp
@@ -293,6 +293,16 @@ static void ClearSourcePathsMappings() {
   LOG("Cleared the saved source paths mappings.");
 }
 
+// Put the command line into the log.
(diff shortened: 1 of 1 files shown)


src/Orbit/main.cpp
```diff
@@ -293,6 +293,16 @@ static void ClearSourcePathsMappings() {
   LOG("Cleared the saved source paths mappings.");
 }
 
+// Put the command line into the log.
+void LogCommandLine(int argc, char* argv[]) {
```
beckerhe, line 297: `static`

beckerhe (approved):
Otherwise LGTM. Thanks!

src/Orbit/main.cpp
danielfenner: done

PR #2195, opened 2021-04-15
Make UpdateVisibleTrack run only in the main thread

karupayun:
We had sometimes a crash when starting a capture because
multi-threading (http://b/185459021). This only happened when
we had at least one FrameTrack enabled.

In this PR, instead of calling UpdateVisibleTrack when adding/removing a
FrameTrack we set a flag, so the main thread could update them.

For that, we also force SortTracks when we finish to process the
capture. This simplifies a little the logic of SortTracks, and we can
only actually sorting when the sorting is invalidated or when the
capture is running and we didn't make it in the last second.


dpallotti:
I guess that as you are discussing it in person with Henning and Florian then I'll leave it to them.

beckerhe (approved):
LGTM. Thanks Pablo!

PR #2196, opened 2021-04-15
Reset the default value of memory sampling period

vickyliu-go4it:
When the memory sampling period is set to a high number, e.g., 1000 ms,
the update in the capture window looks to be "stuck". This is because
update of events in the other tracks might be lagging far behind the
update of events in memory tracks.

When the memory sampling period is set to 10 ms, the capture window
updates smoothly and the performance on the service side is affected
very slightly. Hence, in this change, we change the default value of the
sampling period to be 10ms.

Bug: http://b/185473068


karupayun (approved):
LGTM Vicky!! Thanks for this hotfix

reichlfl (approved):
LGTM, thanks

PR #2199, opened 2021-04-15
Syncronize ProcessData

dimitry-:
Since memory_map in process data can be accessed from
capture thread we need to syncronize this structure.

Note that absl::Mutex renders ProcessData uncopiable and
unmovable so all the refernces changed from std::optional<>
to std::unique_ptr<>

Test: start oribit, start and stop capture 4-5 times.


src/OrbitClientData/ProcessData.cpp
```diff
@@ -17,7 +17,43 @@ using orbit_grpc_protos::ModuleInfo;
...
 ProcessData::ProcessData() { process_info_.set_pid(-1); }
 
+void ProcessData::SetProcessInfo(const orbit_grpc_protos::ProcessInfo& process_info) {
+  absl::MutexLock lock(&mutex_);
+  process_info_ = process_info;
+}
+
+int32_t ProcessData::pid() const {
+  absl::MutexLock lock(&mutex_);
+  return process_info_.pid();
```
dpallotti, line 27: Would it make sense to use `ReaderMutexLock`s and `WriterMutexLock`s as most accesses are reads?

dpallotti (approved):
Thanks a lot!

dpallotti:
It seems that there's a failing test.

src/OrbitClientData/ProcessDataTest.cpp
```diff
@@ -180,9 +180,9 @@ TEST(ProcessData, IsModuleLoaded) {
   ProcessData process(ProcessInfo{});
   process.UpdateModuleInfos(module_infos);
 
-  EXPECT_NE(process.FindModuleByPath(file_path_1), nullptr);
-  EXPECT_NE(process.FindModuleByPath(file_path_2), nullptr);
-  EXPECT_EQ(process.FindModuleByPath("not/loaded/module"), nullptr);
+  EXPECT_TRUE(process.FindModuleByPath(file_path_1).has_value());
+  EXPECT_TRUE(process.FindModuleByPath(file_path_2).has_value());
+  EXPECT_TRUE(process.FindModuleByPath("not/loaded/module").has_value());
```
dpallotti, line 185: This should be `EXPECT_FALSE`?

src/OrbitClientData/ProcessData.cpp
dimitry-: access to this class is not contentious enough to warrant suing reader/writer locks. Let's keep it as it is.

src/OrbitClientData/ProcessDataTest.cpp
dimitry-: done.

PR #2202, opened 2021-04-16
Inject liborbit.so using UserSpaceInstrumentation lib

pierricgimmig:


src/Api/InitializeInTracee.cpp
```diff
@@ -4,16 +4,47 @@
...
+ErrorMessageOr<void> InitializeInTracee(const CaptureOptions& capture_options) {
+  if (capture_options.api_table_infos().size() == 0) {
+    return ErrorMessage("No api table to initilize.");
+  }
+
+  ErrorMessageOr<void> result = outcome::success();
+  int32_t pid = capture_options.pid();
+
+  OUTCOME_TRY(AttachAndStopProcess(pid));
+  {
```
danielfenner, line 30: The additional scope is not needed here. Restarting on return is just fine.

src/Api/InitializeInTracee.cpp
```diff
@@ -4,16 +4,47 @@
...
+        pid, [&result](int32_t pid) { result = DetachAndContinueProcess(pid); }};
+
+    // Load liborbit.so and find api table initialization function.
+    const std::string kLibPath = orbit_base::GetExecutableDir() / "liborbit.so";
+    const char* kInitFunction = "orbit_api_initialize";
+    OUTCOME_TRY(handle, DlopenInTracee(pid, kLibPath, RTLD_NOW));
+    OUTCOME_TRY(address, DlsymInTracee(pid, handle, kInitFunction));
+
+    // Initialize all api function tables.
+    for (const ApiTableInfo& info : capture_options.api_table_infos()) {
```
danielfenner, line 42: This loop could be moved into the init function, right? Not sure if it is worth it.

src/OrbitCaptureClient/CaptureClient.cpp
```diff
@@ -95,8 +95,38 @@ Future<ErrorMessageOr<CaptureListener::CaptureOutcome>> CaptureClient::Capture(
   return capture_result;
 }
 
+static void SetupApiTableInfos(const ProcessData& process,
+                               const orbit_client_data::ModuleManager& module_manager,
+                               CaptureOptions* capture_options) {
+  size_t kMaxApiVersion = 0;
```
danielfenner, line 101: `constexpr size_t kMaxApiVersion = 0;`

src/OrbitCaptureClient/CaptureClient.cpp
```diff
@@ -95,8 +95,38 @@ Future<ErrorMessageOr<CaptureListener::CaptureOutcome>> CaptureClient::Capture(
...
+  const char* kOrbitApiPrefix = "g_orbit_api_v";
```
danielfenner, line 102: const constexpr char* kOrbitApiPrefix = "g_orbit_api_v";

src/OrbitCaptureClient/CaptureClient.cpp
```diff
@@ -95,8 +95,38 @@ Future<ErrorMessageOr<CaptureListener::CaptureOutcome>> CaptureClient::Capture(
...
+    // module_data->GetDataSymbolsFromName(kOrbitApiPrefix)
+    std::vector<SymbolInfo*> dummy_global_variables_symbol_infos;
+
+    for (const SymbolInfo* symbol_info : dummy_global_variables_symbol_infos) {
+      for (size_t i = 0; i <= kMaxApiVersion; ++i) {
+        std::string orbit_api_table_name = absl::StrFormat("%s%u", kOrbitApiPrefix, i);
+        if (symbol_info->name() != orbit_api_table_name) continue;
+        ApiTableInfo* api_table_info = capture_options->add_api_table_infos();
+        api_table_info->set_module_path(module_data->file_path());
+        api_table_info->set_api_table_address(symbol_info->address() + base_address.value());
```
danielfenner, line 120: Is the load bias missing here?

src/UserSpaceInstrumentation/include/UserSpaceInstrumentation/InjectLibraryInTracee.h
```diff
@@ -5,6 +5,7 @@
 #ifndef USER_SPACE_INSTRUMENTATION_INJECT_LIBRARY_IN_TRACEE_H_
 #define USER_SPACE_INSTRUMENTATION_INJECT_LIBRARY_IN_TRACEE_H_
 
+#include <dlfcn.h>
```
danielfenner, line 8: I get it, but a automatic check would complain since this violates iwyu. Suggestion: make `RTLD_NOW` the default for `flag` parameter below.

src/Api/InitializeInTracee.cpp
```diff
@@ -4,16 +4,47 @@
...
+  }
+
+  ErrorMessageOr<void> result = outcome::success();
+  int32_t pid = capture_options.pid();
+
+  OUTCOME_TRY(AttachAndStopProcess(pid));
+  {
+    // Make sure we resume the target process, even on early-outs.
+    orbit_base::unique_resource scope_exit{
+        pid, [&result](int32_t pid) { result = DetachAndContinueProcess(pid); }};
```
dimitry-, line 33: I feel like we want a proper scope_guard here (one that does not need a resource)

src/UserSpaceInstrumentation/include/UserSpaceInstrumentation/InjectLibraryInTracee.h
pierricgimmig: Good idea, thanks, I did just that.

pierricgimmig: Actually I'll submit the change in a separate PR and remove RTLD_NOW from other call sites.

src/Api/InitializeInTracee.cpp
```diff
@@ -4,16 +4,47 @@
...
+
 #include "Api/Orbit.h"
-#include "OrbitBase/Result.h"
+#include "OrbitBase/ExecutablePath.h"
+#include "OrbitBase/UniqueResource.h"
+#include "UserSpaceInstrumentation/Attach.h"
+#include "UserSpaceInstrumentation/ExecuteInProcess.h"
+#include "UserSpaceInstrumentation/InjectLibraryInTracee.h"
 
+using namespace orbit_user_space_instrumentation;
```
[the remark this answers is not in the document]
pierricgimmig: oh, right, thanks.

src/OrbitCaptureClient/CaptureClient.cpp
pierricgimmig: Done.

src/Api/InitializeInTracee.cpp
pierricgimmig: Well that's interesting, if we get rid of the scope, then "result" will always be outcome::success because the scope exit will be executed after the return value has been assigned.

pierricgimmig: Done.

src/Api/InitializeInTracee.cpp
```diff
@@ -4,16 +4,47 @@
...
+
+  OUTCOME_TRY(AttachAndStopProcess(pid));
+  {
+    // Make sure we resume the target process, even on early-outs.
+    orbit_base::unique_resource scope_exit{
+        pid, [&result](int32_t pid) { result = DetachAndContinueProcess(pid); }};
+
+    // Load liborbit.so and find api table initialization function.
+    const std::string kLibPath = orbit_base::GetExecutableDir() / "liborbit.so";
+    const char* kInitFunction = "orbit_api_initialize";
```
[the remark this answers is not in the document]
pierricgimmig: Done.

src/Api/InitializeInTracee.cpp
pierricgimmig: Do we have one?

src/OrbitCaptureClient/CaptureClient.cpp
pierricgimmig: Actually replaced by kOrbitApiVersion from Orbit.h.

src/Api/InitializeInTracee.cpp
pierricgimmig: Oh but we'd have to pass in arrays to the init function, seems complicated, or am I missing something?

src/OrbitCaptureClient/CaptureClient.cpp
pierricgimmig: yikes, good point, thanks. I think this is the proper way to use the load bias here?

```
api_table_info->set_api_table_address(symbol_info->address() + base_address.value() -
                                              module_data->load_bias());
```

src/Api/InitializeInTracee.cpp
danielfenner: Hmm, yes, right. You would need to allocate memory in the tracee for that and pass the pointer. Let's keep it as is. If we really need this at some point we'll come up with some tooling around it.

danielfenner (approved):
lgtm!

src/OrbitCaptureClient/CaptureClient.cpp
```diff
@@ -79,13 +80,12 @@ Future<ErrorMessageOr<CaptureListener::CaptureOutcome>> CaptureClient::Capture(
...
-       selected_functions = std::move(selected_functions), selected_tracepoints,
-       frame_track_function_ids = std::move(frame_track_function_ids), collect_thread_state,
-       samples_per_second, unwinding_method, enable_introspection,
+      [this, &process, &module_manager, selected_functions = std::move(selected_functions),
+       selected_tracepoints, frame_track_function_ids = std::move(frame_track_function_ids),
+       collect_thread_state, samples_per_second, unwinding_method, enable_introspection,
        max_local_marker_depth_per_command_buffer, collect_memory_info,
        memory_sampling_period_ns]() mutable {
-        return CaptureSync(process_id, module_manager, selected_functions, selected_tracepoints,
+        return CaptureSync(process, module_manager, selected_functions, selected_tracepoints,
```
dimitry-, line 88: I would say pass api_table_infos here instead of process.. and generate it before this call, the way we do it for everything else here.

src/GrpcProtos/capture.proto
```diff
@@ -27,6 +27,13 @@ message InstrumentedFunction {
   string function_name = 5;
 }
 
+message ApiTableInfo {
+  string module_path = 1;
+  uint64 api_table_address = 2;
```
dimitry-, line 32: is this absolute address? maybe we need to pass it in form of file_path and offset, the way it is done for instrumented functions and have the service side to figure out where the module is loaded?

Also this will let us do it on the fly without involvement of the client if a module is not yet loaded.

src/Api/InitializeInTracee.cpp
dimitry-: not yet :)

src/Api/InitializeInTracee.cpp
```diff
@@ -4,15 +4,51 @@
...
+using orbit_user_space_instrumentation::AttachAndStopProcess;
+using orbit_user_space_instrumentation::DetachAndContinueProcess;
+using orbit_user_space_instrumentation::DlopenInTracee;
+using orbit_user_space_instrumentation::DlsymInTracee;
+using orbit_user_space_instrumentation::ExecuteInProcess;
 
 namespace orbit_api {
+ErrorMessageOr<void> InitializeInTracee(const CaptureOptions& capture_options) {
+  if (capture_options.api_table_infos().size() == 0) {
+    return ErrorMessage("No api table to initilize.");
```
dpallotti, line 28:
```suggestion
    return ErrorMessage("No api table to initialize.");
```

src/Api/InitializeInTracee.cpp
```diff
@@ -4,15 +4,51 @@
...
+  }
+
+  int32_t pid = capture_options.pid();
+
+  OUTCOME_TRY(AttachAndStopProcess(pid));
+
+  // Make sure we resume the target process, even on early-outs.
+  orbit_base::unique_resource scope_exit{pid, [](int32_t pid) {
+                                           if (DetachAndContinueProcess(pid).has_error())
+                                             ERROR("Detaching from %i", pid);
```
dpallotti, line 38: Use braces.

src/UserSpaceInstrumentation/include/UserSpaceInstrumentation/InjectLibraryInTracee.h
dpallotti: I don't get it, what's the reason for adding the default `RTLD_NOW` value? This to seems to be a case where using a default argument reduces readability at the call sites.

src/Api/InitializeInTracee.cpp
pierricgimmig: Once we have it I'll be happy to use it :-)

src/Api/InitializeInTracee.cpp
pierricgimmig: Done.

src/OrbitCaptureClient/CaptureClient.cpp
```diff
@@ -95,8 +95,38 @@ Future<ErrorMessageOr<CaptureListener::CaptureOutcome>> CaptureClient::Capture(
...
 }
 
+static void SetupApiTableInfos(const ProcessData& process,
+                               const orbit_client_data::ModuleManager& module_manager,
+                               CaptureOptions* capture_options) {
+  constexpr const char* kOrbitApiPrefix = "g_orbit_api_v";
+  for (const ModuleData* module_data : module_manager.GetAllModuleData()) {
+    std::optional<uint64_t> base_address = process.GetModuleBaseAddress(module_data->file_path());
+    if (!base_address.has_value()) {
+      ERROR("No base address found for module %s", module_data->file_path());
```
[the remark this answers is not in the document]
pierricgimmig: done.

src/UserSpaceInstrumentation/include/UserSpaceInstrumentation/InjectLibraryInTracee.h
pierricgimmig: The main thing I want to do is include <dlfcn.h> here. Maybe "// IWYU pragma: keep" is what we need?

pierricgimmig: I did just that, although I didn't mind the default parameter.

src/GrpcProtos/capture.proto
pierricgimmig: Changed so that we only send the relative address, module_path and build_id. The service verifies the build-id's match and computes the absolute address.

src/OrbitCaptureClient/CaptureClient.cpp
```diff
@@ -95,8 +95,38 @@ Future<ErrorMessageOr<CaptureListener::CaptureOutcome>> CaptureClient::Capture(
...
+      continue;
+    }
+
+    // TODO: The global variable symbols will be obtained in the next PR.
```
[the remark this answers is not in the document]
pierricgimmig: Outdated, added the implementation.

src/GrpcProtos/capture.proto
```diff
@@ -27,6 +27,13 @@ message InstrumentedFunction {
   string function_name = 5;
 }
 
+message ApiTableInfo {
```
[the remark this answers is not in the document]
pierricgimmig: I changed the name to ApiFunctionInfo which is more self describing. The logic changed as well, we now call a function in the tracee to retrieve the address of the function table.

src/OrbitCaptureClient/CaptureClient.cpp
pierricgimmig: Reverted to original signature.

src/OrbitCaptureClient/CaptureClient.cpp
```diff
@@ -95,8 +95,38 @@ Future<ErrorMessageOr<CaptureListener::CaptureOutcome>> CaptureClient::Capture(
   return capture_result;
 }
 
+static void SetupApiTableInfos(const ProcessData& process,
```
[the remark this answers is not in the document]
pierricgimmig: Good point, I added a paragraph.

src/OrbitCaptureClient/CaptureClient.cpp
```diff
@@ -95,8 +95,38 @@ Future<ErrorMessageOr<CaptureListener::CaptureOutcome>> CaptureClient::Capture(
...
+                               const orbit_client_data::ModuleManager& module_manager,
+                               CaptureOptions* capture_options) {
```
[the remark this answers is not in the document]
pierricgimmig: Done.

pierricgimmig:
Hi @dpallotti , @dimitry- , @danielfenner . I added a few commits to this PR so that the change works end-to-end with no TODOs. In essence, the changes are:
- addressed comments
- don't rely on global variable symbol, but only on function symbols to find the address where to punch in functions (see FindApiFunctions in CaptureClient.cpp)
- Don't use absolute address when sending ApiFunctionInfos
- Call a function in tracee to get api function table address

Not done: moving api loading logic to ApiLoader project. Will do in separate PR, b/185862867.

I know the PR has gotten a little bigger, but hopefully it is now clearer. Thanks again for your help.

src/Api/InitializeInTracee.cpp
```diff
@@ -4,15 +4,101 @@
...
+    // Filter api functions.
+    constexpr const char* kOrbitApiGetAddressPrefix = "orbit_api_get_function_table_address_v";
+    if (absl::StrContains(api_function.name(), kOrbitApiGetAddressPrefix) == false) continue;
+
+    // Get ModuleInfo associated with function.
+    const ModuleInfo* module_info = FindModuleInfoForApiFunction(api_function, modules_by_path);
+    if (module_info == nullptr) continue;
+
+    // Get address of function table by calling "orbit_api_get_function_table_address_vN" in tracee.
+    void* api_function_address = absl::bit_cast<void*>(
```
danielfenner, line 93: Nothing for this PR but I wonder if `void*` is the right choice to identify addresses in the trace. Both processes are 64 bit so it doesn't matter. I might consider to change the tracee side pointers to uint64_t in InjectLibraryInTracee.h and ExecuteInProcess.h.

src/OrbitClientData/ModuleManager.cpp
```diff
@@ -108,4 +108,12 @@ std::vector<FunctionInfo> ModuleManager::GetOrbitFunctionsOfProcess(
   return result;
 }
 
-}  // namespace orbit_client_data
\ No newline at end of file
+std::vector<const ModuleData*> ModuleManager::GetAllModuleData() const {
+  std::vector<const ModuleData*> result;
```
dimitry-, line 112: MutexLock

src/OrbitCaptureClient/CaptureClient.cpp
```diff
@@ -95,6 +96,34 @@ Future<ErrorMessageOr<CaptureListener::CaptureOutcome>> CaptureClient::Capture(
...
+  // the Api after having injected liborbit.so.
+
+  std::vector<ApiFunction> api_functions;
+  for (const ModuleData* module_data : module_manager.GetAllModuleData()) {
+    constexpr const char* kOrbitApiGetAddressPrefix = "orbit_api_get_function_table_address_v";
+    for (size_t i = 0; i <= kOrbitApiVersion; ++i) {
+      std::string orbit_api_function_name = absl::StrFormat("%s%u", kOrbitApiGetAddressPrefix, i);
+      uint64_t function_hash = function_utils::GetHash(orbit_api_function_name);
+
+      const FunctionInfo* function_info = module_data->FindFunctionFromHash(function_hash);
```
dimitry-, line 113: maybe add FindFunctionByName (you can use hash internally for now but at least do the verification for cases of hash collision)

dimitry- (approved):
otherwise lgtm

src/Api/InitializeInTracee.cpp
```diff
@@ -4,15 +4,101 @@
...
 using orbit_grpc_protos::CaptureOptions;
+using orbit_grpc_protos::ModuleInfo;
+using orbit_user_space_instrumentation::AttachAndStopProcess;
+using orbit_user_space_instrumentation::DetachAndContinueProcess;
+using orbit_user_space_instrumentation::DlopenInTracee;
+using orbit_user_space_instrumentation::DlsymInTracee;
+using orbit_user_space_instrumentation::ExecuteInProcess;
+
+static ErrorMessageOr<absl::flat_hash_map<std::string, ModuleInfo>> GetModulesByPathForPid(
+    int32_t pid) {
```
danielfenner, line 32: I use `pid_t` from `#include <sys/types.h>` as a type for `pid`. Probably doesn't matter since it is an int everywhere.

danielfenner (approved):
lgtm!

src/Service/LinuxTracingHandler.cpp
```diff
@@ -60,6 +62,13 @@ void LinuxTracingHandler::SetupIntrospection() {
       });
 }
 
+void LinuxTracingHandler::SetupOrbitApi(const CaptureOptions& capture_options) {
```
dpallotti, line 65: This should not be in `LinuxTracingHandler` but further up in `OrbitService`, closer to `CaptureServiceImpl::Capture` (probably just in that file, this is really not a lot of code after all).
Basically the idea is that now `LinuxTracingHandler` is not the "master" producer anymore, but is just one of the producers, together with the memory producer and the receivers from the Unix-socket channel (which manual instrumentation also uses).

src/Service/LinuxTracingHandler.cpp
```diff
@@ -60,6 +62,13 @@ void LinuxTracingHandler::SetupIntrospection() {
...
+  auto result = orbit_api::InitializeInTracee(capture_options);
+  if (result.has_error()) {
+    LOG("Warning: Could not initialize Orbit Api: %s", result.error().message());
```
dpallotti, line 68: I think this could just be an `ERROR`, we don't have a distinction between `ERROR` and less severe warnings (and this is actually quite severe).

src/OrbitCaptureClient/CaptureClient.cpp
```diff
@@ -95,6 +96,34 @@ Future<ErrorMessageOr<CaptureListener::CaptureOutcome>> CaptureClient::Capture(
   return capture_result;
 }
 
+[[nodiscard]] static std::vector<ApiFunction> FindApiFunctions(
+    const orbit_client_data::ModuleManager& module_manager) {
+  // Api functions are declared in Orbit.h. They are implemented in user code through the
```
dpallotti, line 101: I would move this comment outside of the block (above the function) for better visibility.

src/OrbitCaptureClient/CaptureClient.cpp
```diff
@@ -95,6 +96,34 @@ Future<ErrorMessageOr<CaptureListener::CaptureOutcome>> CaptureClient::Capture(
...
+  // ORBIT_API_INSTANTIATE macro. Those functions are used to query the tracee for Orbit specific
+  // information such as the memory location where Orbit should write function pointers to enable
```
dpallotti, line 103:
```suggestion
  // information, such as the memory location where Orbit should write function pointers to enable
```
This is a long sentence.

src/GrpcProtos/capture.proto
```diff
@@ -27,6 +27,14 @@ message InstrumentedFunction {
   string function_name = 5;
 }
 
+message ApiFunction {
```
dpallotti, line 30: I still feel like a small documentation comment here would help, even if the name changed, because the purpose is quite obscure unless one has reviewed this change. It can also point you to `FindApiFunctions`.

src/Api/include/Api/Orbit.h
```diff
@@ -35,9 +35,9 @@
 //
 // Integration:
 // To integrate the manual instrumentation API in your code base, simply include this header file
-// and instantiate the "g_orbit_api_v0" global variable in your code. Orbit will automatically
+// and place the ORBIT_API_INSTANTIATE macro in an implementation file. Orbit will automatically
 // deploy and dynamically load liborbit.so into the target process. Orbit will then write the proper
-// function addresses into the "g_orbit_api_v0" table.
+// function addresses into the "g_orbit_api_v[N]" table.
```
dpallotti, line 40:
```suggestion
// function addresses into the "g_orbit_api_v<N>" table.
```
Maybe a weird suggestion, but it looked like an array otherwise.

src/OrbitClientData/ModuleManager.cpp
pierricgimmig: Oh yes, thanks! Done.

src/Api/InitializeInTracee.cpp
```diff
@@ -4,15 +4,101 @@
...
+
+  // Make sure we resume the target process, even on early-outs.
+  orbit_base::unique_resource scope_exit{pid, [](int32_t pid) {
+                                           if (DetachAndContinueProcess(pid).has_error()) {
+                                             ERROR("Detaching from %i", pid);
+                                           }
+                                         }};
+
+  // Load liborbit.so and find api table initialization function.
+  const std::string kLibPath = orbit_base::GetExecutableDir() / "liborbit.so";
```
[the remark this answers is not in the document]
pierricgimmig: Done, thanks.

src/Api/InitializeInTracee.cpp
pierricgimmig: Yeah it's a good point, we should decide if we want to support 32 bit tracees or not, ideally we would, but then we need to add 32 bit versions for functions in InjectLibraryInTracee.h and ExecuteInProcess.h

src/Api/InitializeInTracee.cpp
pierricgimmig: yeah, I think it's fine

src/Service/LinuxTracingHandler.cpp
pierricgimmig: Yeah, switched to ERROR, thanks

src/OrbitCaptureClient/CaptureClient.cpp
pierricgimmig: Done.

src/OrbitCaptureClient/CaptureClient.cpp
pierricgimmig: Very good, thanks. Done.

src/OrbitCaptureClient/CaptureClient.cpp
pierricgimmig: Yeah good idea, thanks, I did just that.

src/GrpcProtos/capture.proto
pierricgimmig: Added some doc.

src/Service/LinuxTracingHandler.cpp
pierricgimmig: Sounds good, thanks for the thorough review @dpallotti . I moved that logic directly in CaptureServiceImpl::Capture, basically calling 'orbit_api::InitializeInTracee' directly, without the 'SetupOrbitApi' wrapper.

pierricgimmig:
Thanks @danielfenner , @dimitry- and @dpallotti  for the thorough reviews! I addressed all the comments.

src/Api/InitializeInTracee.cpp
```diff
@@ -4,15 +4,101 @@
...
+  constexpr const char* kInitFunction = "orbit_api_initialize";
+  OUTCOME_TRY(handle, DlopenInTracee(pid, liborbit_path, RTLD_NOW));
+  OUTCOME_TRY(orbit_init_function, DlsymInTracee(pid, handle, kInitFunction));
+
+  // Initialize all api function tables.
+  OUTCOME_TRY(modules_by_path, GetModulesByPathForPid(pid));
+  for (const ApiFunction& api_function : capture_options.api_functions()) {
+    // Filter api functions.
+    constexpr const char* kOrbitApiGetAddressPrefix = "orbit_api_get_function_table_address_v";
+    if (absl::StrContains(api_function.name(), kOrbitApiGetAddressPrefix) == false) continue;
```
dpallotti, line 86:
```suggestion
    if (!absl::StrContains(api_function.name(), kOrbitApiGetAddressPrefix)) continue;
```

src/GrpcProtos/capture.proto
```diff
@@ -27,6 +27,18 @@ message InstrumentedFunction {
   string function_name = 5;
 }
 
+// Api functions are declared in Orbit.h. They are implemented in user code
+// through the ORBIT_API_INSTANTIATE macro. Those functions are used to query
+// the tracee for Orbit specific information. See "FindApiFunction" form more
```
dpallotti, line 32:
```suggestion
// the tracee for Orbit specific information. See "FindApiFunction" for more
```

dpallotti (approved):

src/GrpcProtos/capture.proto
pierricgimmig: Arrgh, thanks for that.

PR #2203, opened 2021-04-16
Package liborbit.so

pierricgimmig:
Remove unused command that would try package .so files from /bin and 
replace by new one that explicitly packages liborbit.so.

### DIFF
--- a/conanfile.py
+++ b/conanfile.py
@@ -229,9 +229,9 @@ class OrbitConan(ConanFile):
     def package(self):
         if self.options.debian_packaging:
             shutil.rmtree(self.package_folder)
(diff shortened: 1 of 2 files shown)


beckerhe (approved):
LGTM.

Just one more thing: The log file says that liborbit.so is rather large. I guess
symbols are not stripped. That can be easily done with our `strip_symbols`
CMake command and might be worthwhile. An example can be found in `src/LinuxTracingIntegrationTests/CMakeLists.txt`.

pierricgimmig:
> I guess symbols are not stripped. That can be easily done with our `strip_symbols`

Ah great, thanks for the info, will look into it.

PR #2209, opened 2021-04-16
[RELEASE] Fix text position update

reichlfl:
This allows the GlCanvas::Draw() method to update the viewport, which
will correctly mark it as dirty again and trigger a new rendering next
frame.

This is required because we change the height of tracks during TimeGraph::UpdatePrimitives(),
but in the same method, text is rendered that relies on the extents of the viewport
that have been calculated before.

To reproduce, take a look at the description in this bug: b/185363608

### DIFF
--- a/src/OrbitGl/CaptureWindow.cpp
+++ b/src/OrbitGl/CaptureWindow.cpp
@@ -359,14 +359,14 @@ std::unique_ptr<AccessibleInterface> CaptureWindow::CreateAccessibleInterface() 
   return std::make_unique<AccessibleCaptureWindow>(this);
 }
 
-void CaptureWindow::Draw() {
+void CaptureWindow::Draw(bool viewport_was_dirty) {
   ORBIT_SCOPE("CaptureWindow::Draw");
   if (ShouldAutoZoom()) {
     ZoomAll();
   }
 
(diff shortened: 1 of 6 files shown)


florian-kuebler (approved):
LGTM

src/OrbitGl/CaptureWindow.cpp
```diff
@@ -359,14 +359,14 @@ std::unique_ptr<AccessibleInterface> CaptureWindow::CreateAccessibleInterface()
   return std::make_unique<AccessibleCaptureWindow>(this);
 }
 
-void CaptureWindow::Draw() {
+void CaptureWindow::Draw(bool viewport_was_updated) {
```
karupayun, line 362: I prefer "viewport_was_dirty"

karupayun (approved):
LGTM! But not 100% sure if we should rush this for the release. But I'm OK of doing it anyway.

src/OrbitGl/CaptureWindow.cpp
reichlfl: I've changed the name and fixed the gcc warning, PTAL.

florian-kuebler: Looks good.

PR #2210, opened 2021-04-16
Cleanup CHECK vs ASSERT.

danielfenner:
A bunch of ASSERTs are not really testing something related to the test.
Made these CHECKs.

Bug: http://b/185312970
Test: Unit tests.

### DIFF
--- a/src/UserSpaceInstrumentation/AccessTraceesMemoryTest.cpp
+++ b/src/UserSpaceInstrumentation/AccessTraceesMemoryTest.cpp
@@ -172,7 +172,7 @@ TEST(AccessTraceesMemoryTest, WriteFailures) {
 
 TEST(AccessTraceesMemoryTest, ReadWriteRestore) {
   pid_t pid = fork();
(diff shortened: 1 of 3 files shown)


dpallotti (approved):

src/UserSpaceInstrumentation/AccessTraceesMemoryTest.cpp
```diff
@@ -172,18 +172,18 @@ TEST(AccessTraceesMemoryTest, WriteFailures) {
...
+  CHECK(pid != -1);
   if (pid == 0) {
     // Child just runs an endless loop.
     while (true) {
     }
   }
 
   // Stop the child process using our tooling.
-  ASSERT_TRUE(AttachAndStopProcess(pid).has_value());
+  CHECK(!AttachAndStopProcess(pid).has_error());
```
dimitry-, line 183: Are you sure we want to loose error message here? I feel like this making it harder to diagnose broken tests.

PR #2212, opened 2021-04-16
Cleanup: Rename variable.

danielfenner:
Restricted this PR to more or less trivial, mechanical changes.
e.g. `address_code` -> `code_address`. There might be more occurrences
but this should cover most of the issue.

Bug: http://b/185254470
Test: unit test.


pierricgimmig (approved):
Great! thanks for this @danielfenner

PR #2213, opened 2021-04-19
Introduce CaptureFinished message

dimitry-:
This is last event sent for the capture. This
commit establishes the message and sends it at the
end of the capture. For now it is always a success.

Bug: http://b/185754690
Test: run capture in the client.

### DIFF
--- a/src/Service/CaptureServiceImpl.cpp
+++ b/src/Service/CaptureServiceImpl.cpp
@@ -31,6 +31,7 @@
 
 namespace orbit_service {
 
+using orbit_grpc_protos::CaptureFinished;
 using orbit_grpc_protos::CaptureOptions;
 using orbit_grpc_protos::CaptureRequest;
 using orbit_grpc_protos::CaptureResponse;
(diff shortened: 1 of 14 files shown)


florian-kuebler (approved):
LGTM. Thanks.

Just one question: Deserialization will not be affected by those new messages, right? Do we make sure that this remains without problems?

dimitry-:
> Just one question: Deserialization will not be affected by those new messages, right? Do we make sure that this remains without problems?

Actually, maybe it should. Looking.

dimitry-:
> > Just one question: Deserialization will not be affected by those new messages, right? Do we make sure that this remains without problems?
> 
> Actually, maybe it should. Looking.

Ok, it doesn't really have any effect on capture deserialization for current format. Going to leave it as it is.

florian-kuebler:
> > > Just one question: Deserialization will not be affected by those new messages, right? Do we make sure that this remains without problems?
> > 
> > 
> > Actually, maybe it should. Looking.
> 
> Ok, it doesn't really have any effect on capture deserialization for current format. Going to leave it as it is.

Agree. Just wanted to have this checked.

src/Service/CaptureServiceImpl.cpp
```diff
@@ -200,6 +201,13 @@ static void StopInternalProducersAndCaptureStartStopListenersInParallel(
   }
 }
 
+static ClientCaptureEvent CreateCaptureFinishedEvent() {
```
dpallotti, line 204: Maybe move after `CreateCaptureStartedEvent` as it seems more logical.

dpallotti (approved):

src/Service/CaptureServiceImpl.cpp
dimitry-: done.

PR #2214, opened 2021-04-19
Set Visual Studio startup project to Orbit

beckerhe:
This was forgotten when we changed the OrbitQt into a library target and
added a new executable target "Orbit".

So without that changes Visual Studio tries to executed the library
target which obviously fails.

This commit changes the default startup project to "Orbit" which is the
new name of the executable target.


florian-kuebler (approved):
Ah, thanks a lot!

karupayun (approved):
Thanks for being so fast! :)

PR #2215, opened 2021-04-19
Fix missing comment in header.

danielfenner:


src/UserSpaceInstrumentation/include/UserSpaceInstrumentation/Attach.h
```diff
@@ -11,8 +11,13 @@
 
 namespace orbit_user_space_instrumentation {
 
+// Attaches to process `pid` using ptrace.
+// Attaches to and stopps all threads of the process using `PTRACE_ATTACH`. Being attached to a
```
dpallotti, line 15:
```suggestion
// Attaches to and stops all threads of the process using `PTRACE_ATTACH`. Being attached to a
```

dpallotti (approved):
Considering that these comments are short and could easily be merged I'm not sure about the need to have a short summary and a longer version (consider `DetachAndContinueProcess`, they are both one line). I don't think this affects IDE behavior in showing documentation either. 

For example you could just have:
```
Attaches to and stops all threads of the process `pid` using `ptrace` with `PTRACE_ATTACH`. Being attached to a process using this function is a precondition for using any of the tooling provided here.
```
and
```
Detaches from all threads of the process `pid` and continues the execution.
```

But I also have no problem with the current version, just pointing this out.

src/UserSpaceInstrumentation/include/UserSpaceInstrumentation/Attach.h
danielfenner: done

PR #2216, opened 2021-04-19
Add a clang-format script to be invoked as pre-commit hook.

florian-kuebler:
This is a first step into having an automated formatting upon commits.
It adds a script that can be executed from the pre-commit hook directly.

All the files touched by the commit will be formatted completely and
the complete file will be added to the commit. Thus it does not work
with hunks.

Test: Add "source path/to/file/" to the pre-commit hook script.

### DIFF
--- a/contrib/hooks/clang-format-pre-commit.sh
+++ b/contrib/hooks/clang-format-pre-commit.sh
@@ -1,0 +1,21 @@
+#!/bin/bash
+
+# Copyright (c) 2021 The Orbit Authors. All rights reserved.
+# Use of this source code is governed by a BSD-style license that can be
+# found in the LICENSE file.
+
+# A pre-commit hook that applies clang-format-9 to all the changes
+# (staged and unstaged).
+# After the formatting has been applied, it will ask you which changes should
+
+echo "Formatting all changed files now..."
+git clang-format-9 -f
+# Only show this to the user of there are unstaged changes.
+if ! (git diff -- --quiet)
+then
+  exec < /dev/tty
+  echo "\n\n"
+  echo "Please select the changes to be included in this commit ('a' in include all):"
+  git add -p .
+fi
+


contrib/hooks/clang-format-pre-commit.sh
```diff
@@ -0,0 +1,17 @@
...
+# commit.
+# Note, it will format and add the complete file, and thus does not work with
+# hunks.
+
+echo "Formatting changes now..."
+while read line; do
+  echo "Formatting ${line}"
+  clang-format-9 -i ${line}
+  git add ${line}
+done <<< $(git diff-index --cached --name-only HEAD | grep -iE '\.(h|cpp|cc|proto)')
```
dimitry-, line 17: I am not sure if this one has the same behavior, but if I amend commit to amend only a commit message (so there are no changes) it displays a error. Maybe add a check that the list of modified files is not empty?

beckerhe:
Hi Florian,
have you also considered using `clang-format-diff` instead and not format the whole file? I fear that by formatting the whole file we will increase the number of unsolicited code changes which is a burden for reviews.

contrib/hooks/clang-format-pre-commit.sh
florian-kuebler: Should be better now.

florian-kuebler:
I rewrote the complete script. It now formats only the changes, but in all touched files (staged and unstaged). In order to prevent adding unstaged (additional) changes, it will ask you for each change, if it should be added (`git add -p .`). If you want to commit all the changes, you can do that by typing `a`.

beckerhe:
> I rewrote the complete script. It now formats only the changes, but in all touched files (staged and unstaged). In order to prevent adding unstaged (additional) changes, it will ask you for each change, if it should be added (`git add -p .`). If you want to commit all the changes, you can do that by typing `a`.

I see the CI has been triggered, but I don't see the changes. Is it GitHub's fault or was there something wrong with the update?

florian-kuebler:
> > I rewrote the complete script. It now formats only the changes, but in all touched files (staged and unstaged). In order to prevent adding unstaged (additional) changes, it will ask you for each change, if it should be added (`git add -p .`). If you want to commit all the changes, you can do that by typing `a`.
> 
> I see the CI has been triggered, but I don't see the changes. Is it GitHub's fault or was there something wrong with the update?

Sorry for the confusion, my usual `git add src/` didn't helped for obvious reasons. Now it should be there.

beckerhe (approved):
Thanks a lot!

contrib/hooks/clang-format-pre-commit.sh
```diff
@@ -0,0 +1,21 @@
...
+# Copyright (c) 2021 The Orbit Authors. All rights reserved.
+# Use of this source code is governed by a BSD-style license that can be
+# found in the LICENSE file.
+
+# A pre-commit hook that applies clang-format-9 to all the changes
+# (staged and unstaged).
+# After the formatting has been applied, it will ask you which changes should
+
+echo "Formatting all changed files now..."
+git clang-format-9 -f
```
dimitry-, line 12: I have a feeling this does not work in windows git shell

florian-kuebler: Yes, this would be a separate PR. However, we want to fix the clang-format version. As we decided once, to not fix the clang-format rules for version specific stuff, we would end up in a difference between the CI and the local formatting.

dimitry- (approved):

PR #2217, opened 2021-04-19
Remove duplicate definition of AddressRange.

danielfenner:
Bug: http://b/185254608
Test: Existing unit test.

### DIFF
--- a/src/UserSpaceInstrumentation/AddressRange.h
+++ b/src/UserSpaceInstrumentation/AddressRange.h
@@ -1,0 +1,24 @@
+// Copyright (c) 2021 The Orbit Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#ifndef USER_SPACE_INSTRUMENTATION_ADDRESS_RANGE_H_
+#define USER_SPACE_INSTRUMENTATION_ADDRESS_RANGE_H_
+
+#include <cstdint>
+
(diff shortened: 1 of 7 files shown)


src/UserSpaceInstrumentation/AddressRange.h
```diff
@@ -0,0 +1,24 @@
...
+
+#ifndef USER_SPACE_INSTRUMENTATION_ADDRESS_RANGE_H_
+#define USER_SPACE_INSTRUMENTATION_ADDRESS_RANGE_H_
+
+#include <cstdint>
+
+namespace orbit_user_space_instrumentation {
+
+// Represents an address range in the tracee. Mostly useful in the context of parsing the content of
+// "/proc/pid/maps". As in the maps file the address range is stored using the first and "one past
```
dpallotti, line 13:
```suggestion
// "/proc/pid/maps". As in the maps file, the address range is stored using the first and "one past
```
Without it, the sentence sounds like "Because in the maps file..." and it took me a bit to get it.
I'm actually not sure if this is proper English or if it should be "Similarly to the maps file" (but whatever).

dpallotti (approved):
Better, thanks.

dimitry- (approved):

src/UserSpaceInstrumentation/AddressRange.h
danielfenner: done

PR #2218, opened 2021-04-19
Manage memory in tracee via unique_resource.

danielfenner:
Added method AllocateInTraceeAsUniqueResource returning the memory
address wrapped in a unique_resource. Replaced almost all occurrences of
AllocateInTracee by AllocateInTraceeAsUniqueResource.
The only exception is AllocateMemoryForTrampolines: It is not obvious
yet if it is preferable to automatically free the memory or have this
done manually. Note that the memory for the trampolines stays allocated
after DetachAndContinueProcess.

Bug: http://b/185315205
Test: Unit tests.

### DIFF
--- a/src/UserSpaceInstrumentation/AllocateInTracee.cpp
+++ b/src/UserSpaceInstrumentation/AllocateInTracee.cpp
@@ -173,8 +173,10 @@ AllocateInTraceeAsUniqueResource(pid_t pid, uint64_t address, uint64_t size) {
   OUTCOME_TRY(allocated_address, AllocateInTracee(pid, address, size));
   std::function<void(uint64_t)> deleter = [pid, size](uint64_t address) {
     auto result = FreeInTracee(pid, address, size);
-    FAIL_IF(result.has_error(), "Unable to free previously allocated memory in tracee: \"%s\"",
+    if (result.has_error()) {
+      ERROR("Unable to free previously allocated memory in tracee: \"%s\"",
             result.error().message());
+    }
   };
   return orbit_base::unique_resource(allocated_address, deleter);
 }
--- a/src/UserSpaceInstrumentation/AllocateInTracee.cpp
+++ b/src/UserSpaceInstrumentation/AllocateInTracee.cpp
@@ -168,4 +168,15 @@ namespace {
   return outcome::success();
 }
 
+ErrorMessageOr<orbit_base::unique_resource<uint64_t, std::function<void(uint64_t)>>>
+AllocateInTraceeAsUniqueResource(pid_t pid, uint64_t address, uint64_t size) {
+  OUTCOME_TRY(allocated_address, AllocateInTracee(pid, address, size));
+  std::function<void(uint64_t)> deleter = [pid, size](uint64_t address) {
+    auto result = FreeInTracee(pid, address, size);
+    FAIL_IF(result.has_error(), "Unable to free previously allocated memory in tracee: \"%s\"",
+            result.error().message());
+  };
+  return orbit_base::unique_resource(allocated_address, deleter);
+}
+
 }  // namespace orbit_user_space_instrumentation
--- a/src/UserSpaceInstrumentation/AllocateInTracee.h
+++ b/src/UserSpaceInstrumentation/AllocateInTracee.h
@@ -28,7 +28,7 @@ namespace orbit_user_space_instrumentation {
 // Assumes we are already attached to the tracee `pid` e.g. using `AttachAndStopProcess`.
 [[nodiscard]] ErrorMessageOr<void> FreeInTracee(pid_t pid, uint64_t address, uint64_t size);
 
-// Same as `AllocateInTracee` above but wraps the memory in a unique_resource that handle
+// Same as `AllocateInTracee` above but wraps the memory in a unique_resource that handles
 // deallocating the memory. Note that we still need to be attached (or attached again) to the tracee
 // when the unique_resource goes out of scope.
 [[nodiscard]] ErrorMessageOr<orbit_base::unique_resource<uint64_t, std::function<void(uint64_t)>>>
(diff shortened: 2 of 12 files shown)


src/UserSpaceInstrumentation/InjectLibraryInTracee.cpp
```diff
@@ -173,9 +158,7 @@ ErrorMessageOr<uint64_t> FindFunctionAddressWithFallback(pid_t pid, std::string_
                                                                kLibcSoname, kDlcloseInLibc));
 
   // Allocate small memory area in the tracee.
-  OUTCOME_TRY(address, AllocateInTracee(pid, 0, kCodeScratchPadSize));
-  orbit_base::unique_resource code_address{
-      address, [&pid](uint64_t address) { FreeMemoryOrDie(pid, address, kCodeScratchPadSize); }};
+  OUTCOME_TRY(code_address, AllocateInTraceeAsUniqueResource(pid, 0, kCodeScratchPadSize));
 
   // We want to do the following in the tracee:
   // dlclose(handle);
```
dimitry-, line 164: it is probably because we have implicit cast in `unique_resource<>`, but I think we should remove it and change line 181 to explicitly call `code_address.get()`. Otherwise it is unclear while reading the line 181 where does ownership belong, it is not transferred right?

src/UserSpaceInstrumentation/InjectLibraryInTracee.cpp
```diff
@@ -126,10 +114,7 @@ ErrorMessageOr<uint64_t> FindFunctionAddressWithFallback(pid_t pid, std::string_
...
   const size_t symbol_name_length = symbol.length() + 1;  // include terminating zero
   const uint64_t memory_size = kCodeScratchPadSize + symbol_name_length;
-  OUTCOME_TRY(address, AllocateInTracee(pid, 0, memory_size));
-  orbit_base::unique_resource code_address{address, [&pid, &memory_size](uint64_t address) {
-                                             FreeMemoryOrDie(pid, address, memory_size);
-                                           }};
+  OUTCOME_TRY(code_address, AllocateInTraceeAsUniqueResource(pid, 0, memory_size));
 
   // Write the name of symbol into memory at code_address with offset of kCodeScratchPadSize.
   std::vector<uint8_t> symbol_name_as_vector(symbol_name_length, 0);
```
dimitry-, line 120: same goes for line 122

src/UserSpaceInstrumentation/InjectLibraryInTracee.cpp
```diff
@@ -76,10 +67,7 @@ ErrorMessageOr<uint64_t> FindFunctionAddressWithFallback(pid_t pid, std::string_
...
   const uint64_t path_length = path.string().length() + 1;  // Include terminating zero.
   const uint64_t memory_size = kCodeScratchPadSize + path_length;
-  OUTCOME_TRY(address, AllocateInTracee(pid, 0, memory_size));
-  orbit_base::unique_resource code_address{address, [&pid, &memory_size](uint64_t address) {
-                                             FreeMemoryOrDie(pid, address, memory_size);
-                                           }};
+  OUTCOME_TRY(code_address, AllocateInTraceeAsUniqueResource(pid, 0, memory_size));
 
   // Write the name of the .so into memory at code_address with offset of kCodeScratchPadSize.
   std::vector<uint8_t> path_as_vector(path_length);
```
dimitry-, line 73: and line 75

dimitry- (commented):
looping in @beckerhe for removing implicit cast from `unique_resource<>`, wdyt?

src/UserSpaceInstrumentation/InjectLibraryInTracee.cpp
danielfenner: Right, ownership stays with this instance of the unique_resource.

I removed all the `.get()`'s when I found out about the implicit cast. I'm also fine with the more explicit / verbose version. No real preference here.

src/UserSpaceInstrumentation/AllocateInTracee.cpp
```diff
@@ -168,4 +168,15 @@ namespace {
   return outcome::success();
 }
 
+ErrorMessageOr<orbit_base::unique_resource<uint64_t, std::function<void(uint64_t)>>>
+AllocateInTraceeAsUniqueResource(pid_t pid, uint64_t address, uint64_t size) {
+  OUTCOME_TRY(allocated_address, AllocateInTracee(pid, address, size));
+  std::function<void(uint64_t)> deleter = [pid, size](uint64_t address) {
+    auto result = FreeInTracee(pid, address, size);
+    FAIL_IF(result.has_error(), "Unable to free previously allocated memory in tracee: \"%s\"",
```
pierricgimmig, line 176: I wonder if FAIL_IF is too strong here. Is it possible that we FreeInTracee fail if the tracee is no longer alive? In that case I think we should warn rather than kill Orbit, no?

pierricgimmig (approved):
Cool! LGTM, apart from the FAIL_IF comment.

beckerhe:
> looping in @beckerhe for removing implicit cast from `unique_resource<>`, wdyt?

Fine with me! :+1:

dimitry- (approved):
It can be done in the followup change - or this one, your preference.

src/UserSpaceInstrumentation/AllocateInTracee.h
```diff
@@ -26,6 +28,12 @@ namespace orbit_user_space_instrumentation {
 // Assumes we are already attached to the tracee `pid` e.g. using `AttachAndStopProcess`.
 [[nodiscard]] ErrorMessageOr<void> FreeInTracee(pid_t pid, uint64_t address, uint64_t size);
 
+// Same as `AllocateInTracee` above but wraps the memory in a unique_resource that handle
```
dpallotti, line 31:
```suggestion
// Same as `AllocateInTracee` above but wraps the memory in a unique_resource that handles
```

dpallotti (approved):

PR #2219, opened 2021-04-20
Fix flicker of target_label_

antonrohr:
http://b/184644006

This seems to have solved the flicker on my system, please check it out and test.


karupayun (commented):
I still see the flicker

karupayun (approved):
Sorry, I was testing in a wrong PR. Seems to be solved :)

PR #2221, opened 2021-04-20
Make Track Manager manage their own sorting

karupayun:
In this refactoring, we move the logic related to sorting the tracks to inside TrackManager.

We created a method (UpdateTracksOrder) to manage all the logic related to the order of the tracks.

http://b/176077097.

### DIFF
--- a/src/OrbitGl/TrackManager.cpp
+++ b/src/OrbitGl/TrackManager.cpp
@@ -66,97 +66,86 @@ std::vector<FrameTrack*> TrackManager::GetFrameTracks() const {
 }
 
 void TrackManager::SortTracks() {
-  if (!app_->IsCapturing() && !sorted_tracks_.empty() && !sorting_invalidated_ &&
-      !visible_track_list_needs_update_)
-    return;
+  std::lock_guard<std::recursive_mutex> lock(mutex_);
+  // Gather all tracks regardless of the process in sorted order
+  std::vector<Track*> all_processes_sorted_tracks;
 
-  // Reorder threads if sorting isn't valid or once per second when capturing
-  if (sorting_invalidated_ ||
-      (app_->IsCapturing() && last_thread_reorder_.ElapsedMillis() > 1000.0)) {
-    std::lock_guard<std::recursive_mutex> lock(mutex_);
-    // Gather all tracks regardless of the process in sorted order
-    std::vector<Track*> all_processes_sorted_tracks;
-
-    // Frame tracks.
-    for (const auto& name_and_track : frame_tracks_) {
-      all_processes_sorted_tracks.push_back(name_and_track.second.get());
-    }
-
-    // Gpu tracks.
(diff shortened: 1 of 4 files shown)


src/OrbitGl/TrackManager.cpp
```diff
@@ -66,97 +66,86 @@ std::vector<FrameTrack*> TrackManager::GetFrameTracks() const {
...
-      !visible_track_list_needs_update_)
-    return;
-
-  // Reorder threads if sorting isn't valid or once per second when capturing
-  if (sorting_invalidated_ ||
-      (app_->IsCapturing() && last_thread_reorder_.ElapsedMillis() > 1000.0)) {
-    std::lock_guard<std::recursive_mutex> lock(mutex_);
-    // Gather all tracks regardless of the process in sorted order
-    std::vector<Track*> all_processes_sorted_tracks;
+  std::lock_guard<std::recursive_mutex> lock(mutex_);
```
florian-kuebler, line 69: Can you use this opportunity to add the [GUARDED_BY](https://abseil.io/docs/cpp/guides/synchronization#annotation-guidelines) annotation to the members that you know are protected by the mutex?

reichlfl (approved):
Nice change, I like this much better than before. One suggestion: It could be useful for the future to have a separate `Update()` and `Draw()` method (which, in our current weird naming scheme, would be `UpdateTracks()` and `UpdateTracksOrder()`. This way, we can lateron split Update and Rendering (to calculate height before rendering), and `UpdateTracksOrder()` can more easily be unit-tested because it doesn't need any parameters and no Batcher.

Of course this would leave us with 2 calls to `TrackManager` in `TimeGraph::UpdatePrimitives()`, which was one reason for this change... so I'm fine with adding this split later. But if you want to move towards unit-testing the sorting, this should probably happen anyways.

florian-kuebler (approved):
LGTM. Would be nice if you can add the annotation to the mutex, but could be also done in a different PR.

src/OrbitGl/TrackManager.cpp
karupayun: Added to the correspondent bug: http://b/177200020. I feel we should revise it properly, and not only add what I think that are being protected.

karupayun:
> Nice change, I like this much better than before. One suggestion: It could be useful for the future to have a separate `Update()` and `Draw()` method (which, in our current weird naming scheme, would be `UpdateTracks()` and `UpdateTracksOrder()`. This way, we can lateron split Update and Rendering (to calculate height before rendering), and `UpdateTracksOrder()` can more easily be unit-tested because it doesn't need any parameters and no Batcher.
> 
> Of course this would leave us with 2 calls to `TrackManager` in `TimeGraph::UpdatePrimitives()`, which was one reason for this change... so I'm fine with adding this split later. But if you want to move towards unit-testing the sorting, this should probably happen anyways.

I will do it as part of a different PR. I only wanted here to simplify the logic related with sorting.

PR #2222, opened 2021-04-20
Deploy liborbit.so in development mode

pierricgimmig:


src/OrbitQt/servicedeploymanager.cpp
```diff
@@ -536,6 +551,7 @@ outcome::result<ServiceDeployManager::GrpcPort> ServiceDeployManager::ExecImpl()
   } else if (std::holds_alternative<BareExecutableAndRootPasswordDeployment>(
                  *deployment_configuration_)) {
     OUTCOME_TRY(CopyOrbitServiceExecutable());
+    OUTCOME_TRY(CopyOrbitApiLibrary());
     OUTCOME_TRY(StartOrbitServicePrivileged());
```
antonrohr, line 555: personal preference and it is somewhat unrelated to your commit and all the functions are `private`, so feel free to dismiss. 

This would bring the `std::holds_alternative` close to `std::get` call. And also remove the hidden prerequisite of `CopyOrbitApiLibrary`, which is that `deployment_configuration_` holds a `BareExecutableAndRootPasswordDeployment`.
 
```suggestion
    const auto& config = std::get<orbit_qt::BareExecutableAndRootPasswordDeployment>(*deployment_configuration_);
    OUTCOME_TRY(CopyOrbitServiceExecutable(config));
    OUTCOME_TRY(CopyOrbitApiLibrary(config));
    OUTCOME_TRY(StartOrbitServicePrivileged(config));
```

antonrohr (approved):

beckerhe (approved):
Thanks Pierric. LGTM

src/OrbitQt/servicedeploymanager.cpp
pierricgimmig: Ah yes, I agree, I'll do that in other PR in case a new conversation spawns.

PR #2223, opened 2021-04-20
Tooling for trampolines.

danielfenner:
AllInstructionPointersFromProcess returns the set of the instruction
pointers from all the threads of a halted tracee.

LengthOfOverriddenInstructions returns the number of bytes occupied by
the instructions that got hit when overriding a given number of bytes in
the beginning of a piece of code. (Proper documentation in the header)

Bug: http://b/185933738, http://b/185935100
Test: Unit tests in PR.

### DIFF
--- a/src/UserSpaceInstrumentation/Trampoline.cpp
+++ b/src/UserSpaceInstrumentation/Trampoline.cpp
@@ -17,6 +17,7 @@
 #include "OrbitBase/GetProcessIds.h"
 #include "OrbitBase/Logging.h"
 #include "OrbitBase/ReadFileToString.h"
+#include "OrbitBase/UniqueResource.h"
 #include "RegisterState.h"
 
 namespace orbit_user_space_instrumentation {
@@ -160,7 +161,7 @@ ErrorMessageOr<uint64_t> AllocateMemoryForTrampolines(pid_t pid, const AddressRa
   return AllocateInTracee(pid, address_range.start, size);
 }
 
-ErrorMessageOr<absl::flat_hash_set<uint64_t>> AllInstructionPointersFromProcess(pid_t pid) {
+ErrorMessageOr<absl::flat_hash_set<uint64_t>> GetInstructionPointersFromProcess(pid_t pid) {
   absl::flat_hash_set<uint64_t> result;
   std::vector<pid_t> tids = orbit_base::GetTidsOfProcess(pid);
   for (pid_t tid : tids) {
@@ -175,21 +176,21 @@ std::optional<int> LengthOfOverriddenInstructions(csh handle, const std::vector<
                                                   int bytes_to_override) {
   cs_insn* instruction = cs_malloc(handle);
   CHECK(instruction != nullptr);
+  orbit_base::unique_resource scope_exit{instruction,
+                                         [](cs_insn* instruction) { cs_free(instruction, 1); }};
 
   const uint8_t* code_pointer = code.data();
   size_t code_size = code.size();
   uint64_t address = 0;
   int length = 0;
-
   while (cs_disasm_iter(handle, &code_pointer, &code_size, &address, instruction)) {
     length += instruction->size;
     if (length >= bytes_to_override) break;
   }
-  cs_free(instruction, 1);
 
   // Function too short?
   if (length < bytes_to_override) {
-    return {};
+    return std::nullopt;
   }
   return length;
 }
--- a/src/UserSpaceInstrumentation/Trampoline.cpp
+++ b/src/UserSpaceInstrumentation/Trampoline.cpp
@@ -172,8 +172,8 @@ ErrorMessageOr<absl::flat_hash_set<uint64_t>> GetInstructionPointersFromProcess(
   return result;
 }
 
-std::optional<int> LengthOfOverriddenInstructions(csh handle, const std::vector<uint8_t>& code,
-                                                  int bytes_to_override) {
+std::optional<int> LengthOfOverwrittenInstructions(csh handle, const std::vector<uint8_t>& code,
+                                                   int bytes_to_overwrite) {
   cs_insn* instruction = cs_malloc(handle);
   CHECK(instruction != nullptr);
   orbit_base::unique_resource scope_exit{instruction,
@@ -185,11 +185,11 @@ std::optional<int> LengthOfOverriddenInstructions(csh handle, const std::vector<
   int length = 0;
   while (cs_disasm_iter(handle, &code_pointer, &code_size, &address, instruction)) {
     length += instruction->size;
-    if (length >= bytes_to_override) break;
+    if (length >= bytes_to_overwrite) break;
   }
 
   // Function too short?
-  if (length < bytes_to_override) {
+  if (length < bytes_to_overwrite) {
     return std::nullopt;
   }
   return length;
--- a/src/UserSpaceInstrumentation/Trampoline.cpp
+++ b/src/UserSpaceInstrumentation/Trampoline.cpp
@@ -14,8 +14,10 @@
 
 #include "AllocateInTracee.h"
 #include "OrbitBase/ExecuteCommand.h"
+#include "OrbitBase/GetProcessIds.h"
 #include "OrbitBase/Logging.h"
 #include "OrbitBase/ReadFileToString.h"
+#include "RegisterState.h"
 
 namespace orbit_user_space_instrumentation {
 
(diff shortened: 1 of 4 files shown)


src/UserSpaceInstrumentation/Trampoline.cpp
```diff
@@ -158,4 +160,38 @@ ErrorMessageOr<uint64_t> AllocateMemoryForTrampolines(pid_t pid, const AddressRa
   return AllocateInTracee(pid, address_range.start, size);
 }
 
+ErrorMessageOr<absl::flat_hash_set<uint64_t>> AllInstructionPointersFromProcess(pid_t pid) {
```
pierricgimmig, line 163: nit. I'd prefer something like "GetInstructionPointersFromProcess", "l+l+I" can be hard to read.

src/UserSpaceInstrumentation/Trampoline.cpp
```diff
@@ -158,4 +160,38 @@ ErrorMessageOr<uint64_t> AllocateMemoryForTrampolines(pid_t pid, const AddressRa
...
+  absl::flat_hash_set<uint64_t> result;
```
pierricgimmig, line 164: All threads could be at exactly the same instruction, in which case we'd return a single IP, is that what we want?

src/UserSpaceInstrumentation/Trampoline.cpp
```diff
@@ -158,4 +160,38 @@ ErrorMessageOr<uint64_t> AllocateMemoryForTrampolines(pid_t pid, const AddressRa
...
+    RegisterState registers;
+    OUTCOME_TRY(registers.BackupRegisters(tid));
+    result.insert(registers.GetGeneralPurposeRegisters()->x86_64.rip);
+  }
+  return result;
+}
+
+std::optional<int> LengthOfOverriddenInstructions(csh handle, const std::vector<uint8_t>& code,
+                                                  int bytes_to_override) {
+  cs_insn* instruction = cs_malloc(handle);
```
pierricgimmig, line 176: nit. seems like this would be a good candidate for unique_resource

src/UserSpaceInstrumentation/Trampoline.cpp
pierricgimmig: Or maybe GetUniqueInstructionPointersFromProcess? Is this function for test purposes only? I don't see why we would need a set, it seems to me that a hash_map would with tid as key would be more useful, no?

pierricgimmig (commented):
Hi Daniel, this looks good in general, but please see comments.

src/UserSpaceInstrumentation/Trampoline.h
```diff
@@ -56,6 +58,35 @@ namespace orbit_user_space_instrumentation {
...
+//
+// Beginning of the code we want to override:
+// 0x5583ba2837a0: push         rbp
+// 0x5583ba2837a1: mov          rbp, rsp
+// 0x5583ba2837a4: sub          rsp, 0x10
+// 0x5583ba2837a8: mov          rax, qword ptr fs:[0x28]
+//
+// We'd like to override from 0x5583ba2837a0 with a jmp which is five bytes long:
+// 0x5583ba2837a0: jmp          0x5583ba0bf000
+// 0x5583ba2837a5: XYZ
```
dimitry-, line 76: nop?

src/UserSpaceInstrumentation/Trampoline.cpp
```diff
@@ -158,4 +160,38 @@ ErrorMessageOr<uint64_t> AllocateMemoryForTrampolines(pid_t pid, const AddressRa
...
+
+  while (cs_disasm_iter(handle, &code_pointer, &code_size, &address, instruction)) {
+    length += instruction->size;
+    if (length >= bytes_to_override) break;
+  }
+  cs_free(instruction, 1);
+
+  // Function too short?
+  if (length < bytes_to_override) {
+    return {};
```
dimitry-, line 192: nit: std::nullopt

src/UserSpaceInstrumentation/Trampoline.cpp
```diff
@@ -158,4 +160,38 @@ ErrorMessageOr<uint64_t> AllocateMemoryForTrampolines(pid_t pid, const AddressRa
...
+                                                  int bytes_to_override) {
+  cs_insn* instruction = cs_malloc(handle);
+  CHECK(instruction != nullptr);
+
+  const uint8_t* code_pointer = code.data();
+  size_t code_size = code.size();
+  uint64_t address = 0;
+  int length = 0;
+
+  while (cs_disasm_iter(handle, &code_pointer, &code_size, &address, instruction)) {
```
dimitry-, line 184: what happens if this is invalid instruction?

danielfenner: Luckily it just returns false. Added test case, adjusted function documentation.

src/UserSpaceInstrumentation/Trampoline.cpp

src/UserSpaceInstrumentation/Trampoline.h
danielfenner: I don't get this. Are you saying I should override this memory with nops in the procduction code? Maybe, but not really needed. 

Or are you asking if these are nops? Not at all, it's just garbage as described below. These are the last three bytes of the encoding of `sub rsp, 0x10`.

danielfenner: Tried to clarify this in the example.

src/UserSpaceInstrumentation/Trampoline.h
```diff
@@ -56,6 +58,36 @@ namespace orbit_user_space_instrumentation {
...
+// 0x5583ba2837a8: mov          rax, qword ptr fs:[0x28]
+//
+// We'd like to override from 0x5583ba2837a0 with a jmp which is five bytes long:
+// 0x5583ba2837a0: jmp          0x5583ba0bf000
+// 0x5583ba2837a5: (last three bytes of the encoding of `sub rsp, 0x10`)
+//
+// Since the instruction lengths do not align we end up with three bytes of garbage, the next intact
+// instruction starts at 0x5583ba2837a8. So LengthOfOverriddenInstructions would return eight.
+//
+// `handle` is a handle to the capstone library returned by cs_open. `code` contains  the machine
```
dpallotti, line 81:
```suggestion
// `handle` is a handle to the capstone library returned by cs_open. `code` contains the machine
```

src/UserSpaceInstrumentation/Trampoline.h
```diff
@@ -56,6 +58,36 @@ namespace orbit_user_space_instrumentation {
...
+// code of the function that should be overridden. `bytes_to_override` is the number of bytes to
+// override at the beginning of `code`.
+// The return value is either the number of bytes as described above or unset if there was a problem
```
dpallotti, line 84: Maybe "unset" -> "nullopt".

src/UserSpaceInstrumentation/Trampoline.h
```diff
@@ -56,6 +58,36 @@ namespace orbit_user_space_instrumentation {
...
+// with the function. That might either be that the function is too short or that we were unable to
+// disassemble enough code to cover `bytes_to_override` bytes.
+[[nodiscard]] std::optional<int> LengthOfOverriddenInstructions(csh handle,
```
dpallotti, line 87: You use the verb "override" here and in the comment. Did you mean "overwrite" or is "override" intended here?

src/UserSpaceInstrumentation/TrampolineTest.cpp
```diff
@@ -256,4 +257,53 @@ TEST(TrampolineTest, AllocateMemoryForTrampolines) {
...
+  CHECK(error_code == CS_ERR_OK);
+
+  constexpr int kBytesToOverride = 5;
+  const std::vector<uint8_t> kFiveNops{0x90, 0x90, 0x90, 0x90, 0x90};
+  std::optional<int> result = LengthOfOverriddenInstructions(handle, kFiveNops, kBytesToOverride);
+  ASSERT_TRUE(result.has_value());
+  EXPECT_EQ(5, result.value());
+
+  const std::vector<uint8_t> kProlog{0x55, 0x48, 0x89, 0xe5, 0x48, 0x83, 0xec, 0x10, 0x64,
+                                     0x48, 0x8b, 0x04, 0x25, 0x28, 0x00, 0x00, 0x00};
```
dpallotti, line 293: I assume these are real instructions? If yes, it would be better to comment which ones so that it's clear why the result would be eight. Are those the ones in the comment for `LengthOfOverriddenInstructions`?

dpallotti (approved):
Looks good, just some minor comments.
I would personally consider the use of "overwrite" over "override".

src/UserSpaceInstrumentation/Trampoline.h
danielfenner: done

src/UserSpaceInstrumentation/Trampoline.h
danielfenner: done

src/UserSpaceInstrumentation/Trampoline.h
danielfenner: Jep, thanks, agreed, should be "overwrite".

src/UserSpaceInstrumentation/TrampolineTest.cpp
danielfenner: > in the comment for LengthOfOverriddenInstructions

Yep, exactly, added comment.

PR #2224, opened 2021-04-21
Add matcher for ErrorMessageOr.

danielfenner:
Bug: http://b/185464948
Test: Used in a bunch of unit tests.

### DIFF
--- a/src/OrbitBase/include/OrbitBase/TestUtils.h
+++ b/src/OrbitBase/include/OrbitBase/TestUtils.h
@@ -26,13 +26,6 @@ MATCHER(HasNoValue, absl::StrCat(negation ? "Has a" : "Has no", " value.")) {
   return !arg.has_value();
 }
 
-MATCHER(HasError, absl::StrCat(negation ? "Has no" : "Has an", " error.")) {
-  if (arg.has_error()) {
-    *result_listener << "Error: " << arg.error().message();
-  }
-  return arg.has_error();
-}
-
 MATCHER(HasNoError, absl::StrCat(negation ? "Has an" : "Has no", " error.")) {
   if (arg.has_error()) {
     *result_listener << "Error: " << arg.error().message();
--- a/src/OrbitBase/include/OrbitBase/TestUtils.h
+++ b/src/OrbitBase/include/OrbitBase/TestUtils.h
@@ -12,8 +12,6 @@
 
 namespace orbit_base {
 
-namespace testing {
-
 MATCHER(HasValue, absl::StrCat(negation ? "Has no" : "Has a", " value.")) {
   if (arg.has_error()) {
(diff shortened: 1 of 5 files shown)


src/OrbitBase/include/OrbitBase/TestUtils.h
```diff
@@ -0,0 +1,58 @@
...
+#define ORBIT_BASE_TEST_UTILS_H_
+
+#include <absl/strings/match.h>
+#include <absl/strings/str_cat.h>
+#include <absl/strings/str_format.h>
+#include <gmock/gmock.h>
+
+namespace orbit_base {
+
+namespace testing {
```
dimitry-, line 15: we try to avoid nested namespaces. maybe ommit this, or rename orbit_base to orbit_base_testing?

src/OrbitBase/include/OrbitBase/TestUtils.h
```diff
@@ -0,0 +1,58 @@
...
+  }
+  return arg.has_value();
+}
+
+MATCHER(HasNoValue, absl::StrCat(negation ? "Has a" : "Has no", " value.")) {
+  if (arg.has_error()) {
+    *result_listener << "Error: " << arg.error().message();
+  }
+  return !arg.has_value();
+}
```
dimitry-, line 29: I feel like we can go with HasError/HasNoError instead of these. Also maybe HasError should be HasErrorContaining, if you check for a error it will encourage to check the message too.

danielfenner: Since outcome::result offers the somewhat redundant `has_error`, `has_value` functions I'd like to mirror this here.

Sometimes one of the variants is more readable than the other. Eg I'd prefer 
```
ASSERT_THAT(result, HasValue());
EXPECT_EQ(42, result.value());
```
over
```
ASSERT_THAT(result, HasNoError());
EXPECT_EQ(42, result.value());
```
The other way around, if `result` is a `ErrorMessageOr<void>`, I prefer the `Has*Error` methods since there is no value.

> maybe HasError should be HasErrorContaining

I don't think it hurts to keep both. e.g. one could

```
std::vector<ErrorMessageOr<int>> results;
...
EXPECT_THAT(results, Each(HasError()));
```
even whne the messages differ.

dimitry-: do we have examples of checking vector for each having a error? seems like a very unusual check to have.

danielfenner: No, this was just example. go/matchers has lots of ways of composing matchers and I think it would be weird to leave an obvious one.

dimitry- (approved):
Thanks!

PR #2225, opened 2021-04-21
Add e2e test for the memory feature

vickyliu-go4it:
With this change, we add automation test which covers the following basic workflow for the memory feature:
 - start Orbit, connect to a gamelet, and then select a process
 - enable memory tracing, set and validate the memory sampling period
 - take a capture with collecting the system memory usage information and verify the presence of memory tracks

Bug: http://b/183201381


contrib/automation_tests/test_cases/capture_window.py
```diff
@@ -338,3 +345,55 @@ def _execute(self, track_name_filter: str, expect_exists: bool = True):
                 self.expect_true(track.callstacks is not None, 'Track "%s" has callstacks pane' % track.name)
             else:
                 self.expect_true(track.callstacks is None, 'Track "%s" has no callstacks pane' % track.name)
+
+
+class SetAndCheckMemorySamplingPeriod(E2ETestCase):
```
reichlfl, line 350: As we discussed offline, I am not sure about this part of the test. The part of adding it to the Capture test and verifying the existence of the track is great, but in this test, we're basically testing the behavior and correct configuration of the QLineEdit. This should IMO not happen at this level, but rather in a lower level test. It also adds a bunch of duplicated code (to show / hide the options window, and to enable the memory collection), and now that I look at it, I think it should be removed and replaced by another test if possible. @danielfenner WDYT?

At the least, it should be added in a separate commit or even PR.

contrib/automation_tests/orbit_memory_tracing.py
```diff
@@ -0,0 +1,39 @@
...
+    test_cases = [
+        ConnectToStadiaInstance(),
+        FilterAndSelectFirstProcess(process_filter='hello_ggp'),
+        SetAndCheckMemorySamplingPeriod(memory_sampling_period='030'),
+        SetAndCheckMemorySamplingPeriod(memory_sampling_period=''),
+        SetAndCheckMemorySamplingPeriod(memory_sampling_period='ab'),
+        SetAndCheckMemorySamplingPeriod(memory_sampling_period='0'),
+        Capture(collect_system_memory_usage=True),
+        FilterTracks(filter_string="Memory", expected_count=2),
+        Capture(collect_thread_states=False),
```
beckerhe, line 31: Nit: That's a bit confusing to me. `collect_thread_states` is by default False, isn't it? And so is `collect_system_memory_usage`. But just one is mentioned here.

src/OrbitQt/CaptureOptionsDialog.ui
```diff
@@ -85,6 +85,9 @@
           </item>
           <item>
            <widget class="QLineEdit" name="memorySamplingPeriodMsLineEdit">
+            <property name="accessibleName">
+             <string>MemorySamplingPeriodEdit</string>
```
beckerhe, line 89: This "name" looks more like an identifier. If a user was using a screen reader this would be read to them to identify the field. I think we should try to make the name pronouncable for a screen reader and ideally clear to the user. What about just "Memory Sampling Period in Milliseconds" or something similar?

contrib/automation_tests/test_cases/capture_window.py
```diff
@@ -338,3 +345,55 @@ def _execute(self, track_name_filter: str, expect_exists: bool = True):
...
+          # If the user clean the input, sampling period should be reset to the default value 
+          expected = DEFAULT_SAMPLING_PERIOD
+      elif (not memory_sampling_period.isdecimal()) or (not memory_sampling_period.lstrip('0')):
+          # If the input contains invalid characters (e.g., letters) or its value equals to 0, 
+          # sampling period should keep the same as the previous value.
+          expected = memory_sampling_period_edit.texts()[0]
+      else:
+          expected = memory_sampling_period.lstrip('0')
+
+      logging.info('Entering "%s" in "Sampling period (ms)" edit' % memory_sampling_period)
```
beckerhe, line 388: Do we have a policy for python2-style format strings? I personally prefer the python3 ones (`"".format(...)`), but I see the old ones are in use also in other parts of the code. Looping in @reichlfl for the discussion.

beckerhe (approved):
Thank you Vicky. And sorry for being late. I added some comments. Otherwise LGTM.

src/OrbitQt/CaptureOptionsDialog.ui
reichlfl: To be fair, this is only following what we are doing in other places as well, because our accessible names are mainly used for E2E testing - it's unfortunately not realistic (especially for the capture window) that Orbit can be used reasonably with a screen reader in the near future.

Also, there is a difference between accessible name and accessible description (which we are leaving blank...), though I don't know how they are commonly used by screen readers.

I do get the point, so if we change this to be more in line with the intended use, we should try to consistently do this for the other elements as well in the future.

contrib/automation_tests/test_cases/capture_window.py
reichlfl: Not that I know of. The %s formatting is probably my fault to some extent, because I've never gotten used to `.format()` after switching from Python 2 to 3. I don't have a preference.

contrib/automation_tests/test_cases/capture_window.py
reichlfl: Ok, thanks for the input, then let's leave this part of the test in. I'd argue that if this was C++ code, we would probably advocate to factor this out and remove the duplication, and we should not treat our Python Code as less important when it comes to proper architecture.

I don't feel super strong about it in this case, I'm just worried that this will then "spread" as soon as we add another test that wants to also verify something in the options dialog. But I'd leave it up to @vickyliu-go4it, I'm also fine with leaving it as it is.

reichlfl (approved):

contrib/automation_tests/orbit_memory_tracing.py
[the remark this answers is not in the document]
reichlfl: +1, looks like a mistake.

contrib/automation_tests/test_cases/capture_window.py
beckerhe: Okay good. Let's just leave it as it is.

src/OrbitQt/CaptureOptionsDialog.ui
beckerhe: I wasn't aware that we have similar identifiers already in other places. Under these circumstances it's probably best to leave it as-is.

contrib/automation_tests/orbit_memory_tracing.py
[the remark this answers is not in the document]
vickyliu-go4it: Sorry this is a mistake :-( I took the suggested input by the VSCode and didn't notice this! Sorry for my careless 😂

vickyliu-go4it: Sorry for my careless 😂  this is a typo!

danielfenner (approved):
lgtm

contrib/automation_tests/test_cases/capture_window.py
vickyliu-go4it: Thanks for the suggestion😄!  I put the setting and validating of memory sampling period part to a separate commit, and also add a TODO to mention that we will move the capture options dialog utilities to a common base class.

PR #2228, opened 2021-04-21
Remove some implicit dependencies in ServiceDeployManager

pierricgimmig:
As suggested by @antonrohr in #2222


antonrohr (approved):
Thank you!

PR #2229, opened 2021-04-22
Two tweaks to CaptureClient::Capture

dpallotti:
#### Pass process_id instead of process to CaptureClient::Capture
For simplicity reasons, as the methods only needs the pid.

Test: Take a simple capture with Trata and verify that the capture is as
expected.
#### Add collect_scheduling_info parameter to CaptureClient::Capture
So that `CaptureClient` can be used with scheduling information collection off.
I will need it for performance testing of OrbitService in different
configurations.

Test: Capture Trata and verify that the capture still looks in order.


dimitry- (approved):
lgtm

PR #2230, opened 2021-04-22
Replace GLEW by GLAD

beckerhe:
GLAD is another popular loader for OpenGL. It has the advantage that it can be chained to another OpenGL loader.
In that case GLAD will only provide the OpenGL API in terms of a header file (glad/glad.h), but all actual function pointer
load requests (GetProcAddress) will be forwarded to another loader through a callback.

That means for us we can have Qt load OpenGL and have GLAD call into Qt. GLAD will provide the API headers needed in
OrbitGL and forward any function pointer requests to Qt. So we can be sure that both parts of Orbit always use the same
OpenGL implementation and we can rely on Qt's built-in fallback mechanism in case OpenGL is not available.

The latter still requires Orbit to ship with the Mesa software rasterizer (opengl32sw.dll) which will come in a follow-up step.


src/OrbitQt/orbitglwidget.cpp
```diff
@@ -88,9 +88,10 @@ void OrbitGLWidget::initializeGL() {
 
   initializeOpenGLFunctions();
 
-  if (gl_canvas_) {
-    gl_canvas_->Initialize();
-  }
+  gladLoadGLLoader([](const char* name) {
+    // NOLINTNEXTLINE
+    return reinterpret_cast<void*>(QOpenGLContext::currentContext()->getProcAddress(name));
+  });
```
beckerhe, line 94: That's where the magic happens.

conanfile.py
```diff
@@ -99,9 +99,7 @@ def requirements(self):
 
         if self.options.with_gui:
             self.requires("freetype/2.10.0@bincrafters/stable#0")
-            self.requires(
-                "freetype-gl/8d9a97a@{}#2836d28f3d91c308ec9652c2054015db".format(self._orbit_channel))
-            self.requires("glew/2.1.0@{}#0".format(self._orbit_channel))
+            self.requires("freetype-gl/79b03d9@{}".format(self._orbit_channel))
```
danielfenner, line 102: I'd have expected glad to be added here.

danielfenner: Hmm, it is added in the freetyp conanfile.py below. I guess this makes glad available in OrbitGL as well, right?

danielfenner (approved):
lgtm, thanks a bunch for digging into this!

ronaldfw (approved):
This is awesome, thank you so much for this wonderful solution! LGTM!

conanfile.py
beckerhe: Yes, you're right. it works by accident because freetype-gl requires GLAD. But we really should add it here in the include-what-you-use manner. Will fix...

beckerhe: Done.

src/OrbitGl/ImGuiOrbit.cpp
```diff
@@ -416,10 +416,11 @@ void Orbit_ImGui_RenderDrawLists(ImDrawData* draw_data) {
   glUseProgram(g_ShaderHandle);
   glUniform1i(g_AttribLocationTex, 0);
   glUniformMatrix4fv(g_AttribLocationProjMtx, 1, GL_FALSE, &ortho_projection[0][0]);
-#ifdef GL_SAMPLER_BINDING
-  glBindSampler(0, 0);  // We use combined texture/sampler state. Applications
-                        // using GL 3.3 may set that otherwise.
-#endif
+
+  if (GLAD_GL_VERSION_3_3) {
+    glBindSampler(0, 0);  // We use combined texture/sampler state. Applications
```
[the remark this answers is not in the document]
beckerhe: Yes. Done.

PR #2234, opened 2021-04-22
Move UserDefinedCaptureInfo message to a separate file

dimitry-:
We are going to reuse it in new capture format.

Test: load previously saved capture.

### DIFF
--- a/src/ClientProtos/user_defined_capture_info.proto
+++ b/src/ClientProtos/user_defined_capture_info.proto
@@ -1,0 +1,16 @@
+// Copyright (c) 2021 The Orbit Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+syntax = "proto3";
+
+package orbit_client_protos;
+
+message FrameTracksInfo {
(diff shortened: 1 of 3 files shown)


src/ClientProtos/user_defined_capture_info.proto
```diff
@@ -0,0 +1,16 @@
+// Copyright (c) 2020 The Orbit Authors. All rights reserved.
```
dpallotti, line 1:
```suggestion
// Copyright (c) 2021 The Orbit Authors. All rights reserved.
```
Since you changed the other one.

dpallotti (approved):
Is it going to stay in `ClientProtos` even if it is for the new capture format?

src/ClientProtos/user_defined_capture_info.proto
dimitry-: > Is it going to stay in `ClientProtos` even if it is for the new capture format?

Yes, the additional sections are going to use protos from this module.

florian-kuebler (approved):
LGTM.

PR #2236, opened 2021-04-24
Remove line and file columns from functions

dimitry-:
Bug: http://b/154580143
Test: run capture check that views do not have file and line fields


dpallotti (approved):
Thanks!

PR #2238, opened 2021-04-26
Remove unused OrbitUtils/utils.cmake

dpallotti:
As far as I can tell, this is very old and completely unused.


beckerhe (approved):
Nice. Thanks!

dpallotti:
Ah nice, it seems that I already have a review, so I'll just merge. Thanks Henning.

PR #2240, opened 2021-04-26
Rename OrbitProducer to CaptureEventProducer and adjust namespace

dpallotti:
- Remove `Orbit` prefix.
- Make the name more specific (match the name of the "main" class).

Test: build, take a capture with the Vulkan layer.


beckerhe (approved):
LGTM! Thanks!

PR #2242, opened 2021-04-26
Fix CMake dev warning in iwyu.cmake

dpallotti:
This gets rid of the following CMake warning that I was always getting in CLion:

```
CMake Warning (dev) at cmake/iwyu.cmake:20:
  Syntax Warning in cmake code at column 85

  Argument not separated from preceding token by whitespace.
Call Stack (most recent call first):
  CMakeLists.txt:119 (include)
This warning is for project developers.  Use -Wno-dev to suppress it.
```

Test: Run IWYU before and after and see that the resulting `iwyu.diff`s are
identical.


beckerhe (approved):
LGTM

PR #2243, opened 2021-04-26
Add and expose enable_api and enable_introspection capture options

pierricgimmig:
Add option to toggle introspection and manual instrumentation from the capture settings UI.

Note that the api toggle works only from disabled to enabled, we can't go from enabled to 
disabled yet. This will come in a subsequent PR.

Test: toggle introspection, toggle api from disabled to enabled.

### DIFF
--- a/src/CaptureClient/include/CaptureClient/CaptureClient.h
+++ b/src/CaptureClient/include/CaptureClient/CaptureClient.h
@@ -78,7 +78,7 @@ class CaptureClient {
       const absl::flat_hash_map<uint64_t, orbit_client_protos::FunctionInfo>& selected_functions,
       const TracepointInfoSet& selected_tracepoints, double samples_per_second,
       orbit_grpc_protos::UnwindingMethod unwinding_method, bool collect_scheduling_info,
-      bool collect_thread_state, bool enable_introspection, bool enable_api,
+      bool collect_thread_state, bool enable_api, bool enable_introspection,
       uint64_t max_local_marker_depth_per_command_buffer, bool collect_memory_info,
       uint64_t memory_sampling_period_ns, CaptureEventProcessor* capture_event_processor);
 
--- a/src/CaptureClient/include/CaptureClient/CaptureClient.h
+++ b/src/CaptureClient/include/CaptureClient/CaptureClient.h
@@ -44,7 +44,7 @@ class CaptureClient {
       absl::flat_hash_map<uint64_t, orbit_client_protos::FunctionInfo> selected_functions,
       TracepointInfoSet selected_tracepoints, double samples_per_second,
       orbit_grpc_protos::UnwindingMethod unwinding_method, bool collect_scheduling_info,
-      bool collect_thread_state, bool enable_introspection, bool enable_api,
+      bool collect_thread_state, bool enable_api, bool enable_introspection,
       uint64_t max_local_marker_depth_per_command_buffer, bool collect_memory_info,
       uint64_t memory_sampling_period_ns,
       std::unique_ptr<CaptureEventProcessor> capture_event_processor);
--- a/src/CaptureClient/include/CaptureClient/CaptureClient.h
+++ b/src/CaptureClient/include/CaptureClient/CaptureClient.h
@@ -44,7 +44,7 @@ class CaptureClient {
       absl::flat_hash_map<uint64_t, orbit_client_protos::FunctionInfo> selected_functions,
       TracepointInfoSet selected_tracepoints, double samples_per_second,
       orbit_grpc_protos::UnwindingMethod unwinding_method, bool collect_scheduling_info,
-      bool collect_thread_state, bool enable_introspection,
+      bool collect_thread_state, bool enable_introspection, bool enable_api,
       uint64_t max_local_marker_depth_per_command_buffer, bool collect_memory_info,
       uint64_t memory_sampling_period_ns,
       std::unique_ptr<CaptureEventProcessor> capture_event_processor);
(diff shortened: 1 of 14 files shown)


ronaldfw:
It looks like the enable introspection option will also be there in non-dev mode? If yes, do we want to expose this setting to users?

src/CaptureClient/include/CaptureClient/CaptureClient.h
```diff
@@ -78,7 +78,7 @@ class CaptureClient {
       const absl::flat_hash_map<uint64_t, orbit_client_protos::FunctionInfo>& selected_functions,
       const TracepointInfoSet& selected_tracepoints, double samples_per_second,
       orbit_grpc_protos::UnwindingMethod unwinding_method, bool collect_scheduling_info,
-      bool collect_thread_state, bool enable_introspection,
+      bool collect_thread_state, bool enable_introspection, bool enable_api,
```
dpallotti, line 81: Careful that the order of `enable_api` and `enable_introspection` is different in the .cpp.

src/OrbitQt/CaptureOptionsDialog.ui
```diff
@@ -58,6 +58,20 @@
...
+          </property>
+         </widget>
+        </item>
+        <item>
+         <widget class="QCheckBox" name="introspectionCheckBox">
+          <property name="text">
+           <string>Enable introspection</string>
+          </property>
+         </widget>
+        </item>
```
dpallotti, line 74: Basically what Ronald said. I think that we should hide this when not in `--devmode`, and also make sure that the when not in `--devmode` introspection is disabled no matter what the last setting for it was.

antonrohr (commented):
I agree with Ronald that we should think about whether we want to expose this to users. 

I have one nit, basically about clean commit history.

Maybe there is a more expressive naming than just `enable_api`, it is very generic. I dont know if `enable_orbit_api` is necessary better, just something to think about. 

Besides these points, everything looks good

src/OrbitQt/orbitmainwindow.cpp
```diff
@@ -924,6 +928,8 @@ void OrbitMainWindow::on_actionCaptureOptions_triggered() {
 
   orbit_qt::CaptureOptionsDialog dialog{this};
   dialog.SetCollectThreadStates(settings.value(kCollectThreadStatesSettingKey, false).toBool());
+  dialog.SetEnableApi(settings.value(kEnableApiSettingKey, false).toBool());
```
dpallotti, line 931: Do we want the default to be false? Wouldn't the user expect to get manual instrumentation by default if they introduced it in their code? Or is there a reason why it should be false by default?

pierricgimmig: Yes you're right, I changed it to default to true, and same for introspection to maintain the same behavior as before by default. Thanks.

src/OrbitQt/CaptureOptionsDialog.ui
pierricgimmig: Yep, done and done, thanks.

src/CaptureClient/include/CaptureClient/CaptureClient.h
pierricgimmig: yikes, thanks.

pierricgimmig:
> It looks like the enable introspection option will also be there in non-dev mode? If yes, do we want to expose this setting to users?

Yeah was on the fence about this, but I think we should maintain the previous behavior which is to keep it as a devmode-only option. I made the change, thanks for bringing it up.

pierricgimmig:
> I have one nit, basically about clean commit history.

The commits will be squashed in a single commit.

pierricgimmig:
> Maybe there is a more expressive naming than just `enable_api`

Yeah, I agree that it is pretty generic, but we have an "Api" project, and "ApiFunctions", etc. so the naming is consistent. I think that if the api evolves into separate functionalities, then we might want to consider exposing multiple toggles more appropriately named, for now I'll leave it as is.

antonrohr (approved):
LGTM, thank you for addressing the comments!

src/OrbitClientGgp/ClientGgp.cpp
```diff
@@ -105,15 +105,17 @@ bool ClientGgp::RequestStartCapture(ThreadPool* thread_pool) {
...
   UnwindingMethod unwinding_method = options_.use_framepointer_unwinding
                                          ? UnwindingMethod::kFramePointerUnwinding
                                          : UnwindingMethod::kDwarfUnwinding;
 
   Future<ErrorMessageOr<CaptureOutcome>> result = capture_client_->Capture(
       thread_pool, target_process_->pid(), module_manager_, selected_functions_,
       selected_tracepoints, options_.samples_per_second, unwinding_method, collect_scheduling_info,
-      collect_thread_state, enable_introspection, max_local_marker_depth_per_command_buffer, false,
-      0, CaptureEventProcessor::CreateForCaptureListener(this, absl::flat_hash_set<uint64_t>{}));
+      collect_thread_state, enable_introspection, enable_api,
```
dpallotti, line 116: `enable_api` should precede `enable_introspection` here.

dpallotti (approved):
Please fix the wrong order in `ClientGgp.cpp`. Looks good otherwise.

src/OrbitClientGgp/ClientGgp.cpp
pierricgimmig: Arrgh, sorry, properly fixed now, thanks.

PR #2246, opened 2021-04-27
Fixes to CaptureFileOutputStream

dimitry-:
Add IsOpen method, improve error handling on Close. Fix logic for writing multiple messages, we need a size there in order to correctly read them. Fix the test accordingly.

### DIFF
--- a/src/CaptureFile/CaptureFileOutputStream.cpp
+++ b/src/CaptureFile/CaptureFileOutputStream.cpp
@@ -29,6 +29,7 @@ class CaptureFileOutputStreamImpl final : public CaptureFileOutputStream {
   [[nodiscard]] ErrorMessageOr<void> WriteCaptureEvent(
       const orbit_grpc_protos::ClientCaptureEvent& event) override;
   void Close() noexcept override;
+  [[nodiscard]] bool IsOpen() noexcept override;
 
  private:
   [[nodiscard]] ErrorMessageOr<void> WriteHeader();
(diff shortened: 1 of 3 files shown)


src/CaptureFile/CaptureFileOutputStream.cpp
```diff
@@ -46,9 +48,9 @@ class CaptureFileOutputStreamImpl final : public CaptureFileOutputStream {
 };
 
 CaptureFileOutputStreamImpl::~CaptureFileOutputStreamImpl() noexcept {
-  // This is not default to make sure close for streams and the file are called in the correct
-  // order.
-  CaptureFileOutputStreamImpl::Close();
+  // The destructor is not default to make sure close for streams and the file are called in
+  // the correct order.
+  (void)CaptureFileOutputStreamImpl::Close();
```
dimitry-, line 53: this should be reset()

src/CaptureFile/CaptureFileOutputStream.cpp
```diff
@@ -66,14 +68,26 @@ ErrorMessageOr<void> CaptureFileOutputStreamImpl::Initialize() {
   return outcome::success();
 }
 
-void CaptureFileOutputStreamImpl::Close() noexcept {
+ErrorMessageOr<void> CaptureFileOutputStreamImpl::Close() noexcept {
+  coded_output_->Trim();
+  if (coded_output_->HadError()) {
+    return HandleWriteError("Unknown", SafeStrerror(file_output_stream_->GetErrno()));
```
florian-kuebler, line 74: Don't we need also need to close the streams in case of an error?

dimitry-: HandleWriteError resets the stream and file and also removes it.

florian-kuebler: Okay, thanks.

florian-kuebler (approved):
LGTM

PR #2249, opened 2021-04-28
Fix requirements for tests in LinuxTracingIntegrationTests

dpallotti:
Events generated from tracepoints don't need root for `perf_event_open`, but
we still need root to read the tracepoint id.
Among those events are `SchedulingSlices`, but they are used to make sure that
the main tracing loop has started. So now, when not running as root (e.g., for
debugging from CLion), sleep for a long time instead.
The sleep should be long enough also when running the test from a debug build
on one of the machines with plenty of cores.
More details in the bug.

Bug: http://b/186609456

Test: Run `LinuxTracingIntegrationTests` on gLinux and on gamelet not as root
with various values of `perf_event_paranoid`.


dimitry- (approved):

florian-kuebler (approved):
LGTM, thanks!

PR #2250, opened 2021-04-28
Implement OpenExitingFileForReadWrite

dimitry-:
Test: run OrbitBaseTests

### DIFF
--- a/src/OrbitBase/File.cpp
+++ b/src/OrbitBase/File.cpp
@@ -103,6 +103,15 @@ ErrorMessageOr<unique_fd> OpenNewFileForReadWrite(const std::filesystem::path& p
   return OpenFile(path, kOpenFlags, kOpenMode);
 }
 
+ErrorMessageOr<unique_fd> OpenExistingFileForReadWrite(const std::filesystem::path& path) {
+#if defined(__linux)
+  constexpr int kOpenFlags = O_RDWR | O_CLOEXEC;
+#elif defined(_WIN32)
+  constexpr int kOpenFlags = O_RDWR | O_BINARY;
+#endif  // defined(__linux)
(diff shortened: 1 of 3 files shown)


src/OrbitBase/File.cpp
```diff
@@ -92,6 +92,15 @@ ErrorMessageOr<unique_fd> OpenFileForWriting(const std::filesystem::path& path)
   return OpenFile(path, kOpenFlags, kOpenMode);
 }
 
+ErrorMessageOr<unique_fd> OpenExistingFileForReadWrite(const std::filesystem::path& path) {
```
dpallotti, line 95: I would say to move it after `OpenNewFileForReadWrite` to keep the same order as in the `.h`.

dpallotti (approved):

src/OrbitBase/FileTest.cpp
```diff
@@ -83,6 +83,19 @@ TEST(File, OpenNewFileForReadWrite) {
   ASSERT_TRUE(fd_or_error.has_value()) << fd_or_error.error().message();
 }
 
+TEST(File, OpenEsitingFileForReadWrite) {
```
florian-kuebler, line 86:
```suggestion
TEST(File, OpenExistingFileForReadWrite) {
```

florian-kuebler (approved):
Besides the typo, lgtm! thanks!

src/OrbitBase/FileTest.cpp
dimitry-: done.

src/OrbitBase/File.cpp
dimitry-: done.

PR #2251, opened 2021-04-28
Capture 128 Bytes of the Stack and Registers on FP-based unwinding.

florian-kuebler:
In order to patch callchains of leaf functions resulting from
frame-pointer-based unwinding, we will need the top of the stack
to unwind the second-last frame. Further, the register content
will be required.

This prepares this change, by adding 128 bytes of the stack as well
as the register content.

Test: LinuxTracingIntegrationTest does not get broken.

### DIFF
--- a/src/LinuxTracing/PerfEvent.h
+++ b/src/LinuxTracing/PerfEvent.h
@@ -46,38 +46,8 @@ class PerfEvent {
   int ordered_in_file_descriptor_ = kNotOrderedInAnyFileDescriptor;
 };
 
-namespace {
 std::array<uint64_t, PERF_REG_X86_64_MAX> perf_event_sample_regs_user_all_to_register_array(
-    const perf_event_sample_regs_user_all& regs) {
-  std::array<uint64_t, PERF_REG_X86_64_MAX> registers{};
-  registers[PERF_REG_X86_AX] = regs.ax;
-  registers[PERF_REG_X86_BX] = regs.bx;
-  registers[PERF_REG_X86_CX] = regs.cx;
-  registers[PERF_REG_X86_DX] = regs.dx;
-  registers[PERF_REG_X86_SI] = regs.si;
-  registers[PERF_REG_X86_DI] = regs.di;
-  registers[PERF_REG_X86_BP] = regs.bp;
-  registers[PERF_REG_X86_SP] = regs.sp;
-  registers[PERF_REG_X86_IP] = regs.ip;
-  registers[PERF_REG_X86_FLAGS] = regs.flags;
-  registers[PERF_REG_X86_CS] = regs.cs;
-  registers[PERF_REG_X86_SS] = regs.ss;
-  // Registers ds, es, fs, gs do not actually exist.
-  registers[PERF_REG_X86_DS] = 0ul;
-  registers[PERF_REG_X86_ES] = 0ul;
-  registers[PERF_REG_X86_FS] = 0ul;
-  registers[PERF_REG_X86_GS] = 0ul;
-  registers[PERF_REG_X86_R8] = regs.r8;
-  registers[PERF_REG_X86_R9] = regs.r9;
-  registers[PERF_REG_X86_R10] = regs.r10;
-  registers[PERF_REG_X86_R11] = regs.r11;
-  registers[PERF_REG_X86_R12] = regs.r12;
-  registers[PERF_REG_X86_R13] = regs.r13;
-  registers[PERF_REG_X86_R14] = regs.r14;
-  registers[PERF_REG_X86_R15] = regs.r15;
-  return registers;
-}
-}  // namespace
+    const perf_event_sample_regs_user_all& regs);
 
 class ContextSwitchPerfEvent : public PerfEvent {
  public:
@@ -240,7 +210,7 @@ class CallchainSamplePerfEvent : public PerfEvent {
   dynamically_sized_perf_event_sample_stack_user stack;
 
   explicit CallchainSamplePerfEvent(uint64_t callchain_size, uint64_t dyn_stack_size)
-      : ips(callchain_size), stack{dyn_stack_size} {
+      : ips(callchain_size), stack(dyn_stack_size) {
     ring_buffer_record.nr = callchain_size;
   }
 
--- a/src/LinuxTracing/PerfEvent.h
+++ b/src/LinuxTracing/PerfEvent.h
@@ -46,6 +46,39 @@ class PerfEvent {
   int ordered_in_file_descriptor_ = kNotOrderedInAnyFileDescriptor;
 };
 
+namespace {
+std::array<uint64_t, PERF_REG_X86_64_MAX> perf_event_sample_regs_user_all_to_register_array(
+    const perf_event_sample_regs_user_all& regs) {
+  std::array<uint64_t, PERF_REG_X86_64_MAX> registers{};
+  registers[PERF_REG_X86_AX] = regs.ax;
+  registers[PERF_REG_X86_BX] = regs.bx;
+  registers[PERF_REG_X86_CX] = regs.cx;
+  registers[PERF_REG_X86_DX] = regs.dx;
+  registers[PERF_REG_X86_SI] = regs.si;
+  registers[PERF_REG_X86_DI] = regs.di;
+  registers[PERF_REG_X86_BP] = regs.bp;
+  registers[PERF_REG_X86_SP] = regs.sp;
+  registers[PERF_REG_X86_IP] = regs.ip;
+  registers[PERF_REG_X86_FLAGS] = regs.flags;
+  registers[PERF_REG_X86_CS] = regs.cs;
+  registers[PERF_REG_X86_SS] = regs.ss;
+  // Registers ds, es, fs, gs do not actually exist.
+  registers[PERF_REG_X86_DS] = 0ul;
+  registers[PERF_REG_X86_ES] = 0ul;
+  registers[PERF_REG_X86_FS] = 0ul;
+  registers[PERF_REG_X86_GS] = 0ul;
+  registers[PERF_REG_X86_R8] = regs.r8;
+  registers[PERF_REG_X86_R9] = regs.r9;
+  registers[PERF_REG_X86_R10] = regs.r10;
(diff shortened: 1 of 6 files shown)


src/LinuxTracing/PerfEvent.h
```diff
@@ -46,6 +46,39 @@ class PerfEvent {
   int ordered_in_file_descriptor_ = kNotOrderedInAnyFileDescriptor;
 };
 
+namespace {
+std::array<uint64_t, PERF_REG_X86_64_MAX> perf_event_sample_regs_user_all_to_register_array(
```
dpallotti, line 50: Maybe move the definition to the .cpp.

src/LinuxTracing/PerfEventOpen.h
```diff
@@ -102,6 +102,7 @@ static constexpr uint64_t SAMPLE_REGS_USER_SP_IP_ARGUMENTS =
 //  sample, this constant should be a parameters and should be made available in
 //  some setting.
 static constexpr uint16_t SAMPLE_STACK_USER_SIZE = 65000;
+static constexpr uint16_t SAMPLE_STACK_USER_SIZE_ON_CALLCHAINS = 128;
```
dpallotti, line 105: Why exactly 128? Is this arbitrary? Either way, also add a comment about the reasoning.

src/LinuxTracing/PerfEventRecords.h
```diff
@@ -110,6 +116,9 @@ struct __attribute__((__packed__)) perf_event_callchain_sample_fixed {
   perf_event_sample_id_tid_time_streamid_cpu sample_id;
   uint64_t nr;
   // The rest of the sample is a uint64_t[nr] that we read dynamically.
+  // Following this there is:
+  // perf_event_sample_regs_user_sp_ip_arguments regs;
+  // perf_event_sample_stack_user_8bytes stack;
```
dpallotti, line 121: Instead of just adding to the previous comment, rewrite it in a single style. "The rest of the sample is a uint64_t[nr]" still sounds like the `uint64_t[nr]` fills the rest of the sample entirely.

dpallotti (commented):
I have a general consideration: By making this change we basically give up the possibility of comparing the overhead on the target of simple callchain unwinding vs callchain+DWARF (which is something we haven't done yet and we probably should do, at least for us to know). And we get rid of callchain-only unwinding completely.

One possibility in this transition phase is to add the new unwinding method (callchain+DWARF) instead of replacing the old one. This could make our life easier when trying to compare the two approaches.

What do you think?

---

> Test: LinuxTracingIntegrationTest does not get broken.

For this kind of changes, at least also take a real capture and check that callstacks are in order.

src/LinuxTracing/PerfEventReaders.cpp
```diff
@@ -120,17 +120,47 @@ std::unique_ptr<CallchainSamplePerfEvent> ConsumeCallchainSamplePerfEvent(
...
   uint64_t nr = 0;
   ring_buffer->ReadValueAtOffset(&nr, offsetof(perf_event_callchain_sample_fixed, nr));
-  auto event = std::make_unique<CallchainSamplePerfEvent>(nr);
+
+  uint64_t size_in_bytes = nr * sizeof(uint64_t) / sizeof(char);
+
+  // We expect the following layout of the perf event:
+  //  struct {
+  //    struct perf_event_header header;
+  //    u64 sample_id; /* if PERF_SAMPLE_IDENTIFIER */
```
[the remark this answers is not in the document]
florian-kuebler: Done.

src/LinuxTracing/PerfEventReaders.cpp
```diff
@@ -120,17 +120,47 @@ std::unique_ptr<CallchainSamplePerfEvent> ConsumeCallchainSamplePerfEvent(
...
+  //    char data[size]; /* if PERF_SAMPLE_STACK_USER */
+  //    u64 dyn_size;    /* if PERF_SAMPLE_STACK_USER &&
+  //                        size != 0 */
+  //  };
+  // Unfortunately, the number of `ips` is dynamic, so we need to compute the offsets by hand,
+  // rather than relying on a struct.
+
+  auto offset_of_ips = offsetof(perf_event_callchain_sample_fixed, nr) +
+                       sizeof(perf_event_callchain_sample_fixed::nr);
+  auto offset_of_regs_user = offset_of_ips + size_in_bytes;
```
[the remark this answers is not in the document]
florian-kuebler: this is just bad naming. regs_user is referring to the struct containing `abi` + all `regs`. I renamed it a bit.

src/LinuxTracing/PerfEventReaders.cpp
```diff
@@ -120,17 +120,47 @@ std::unique_ptr<CallchainSamplePerfEvent> ConsumeCallchainSamplePerfEvent(
...
+  auto offset_of_size = offset_of_regs_user + sizeof(perf_event_sample_regs_user_all);
+  auto offset_of_data = offset_of_regs_user + sizeof(perf_event_sample_regs_user_all);
```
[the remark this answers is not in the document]
florian-kuebler: Thanks a lot. Fixed that.

src/LinuxTracing/PerfEventReaders.cpp
```diff
@@ -120,17 +120,47 @@ std::unique_ptr<CallchainSamplePerfEvent> ConsumeCallchainSamplePerfEvent(
...
+
+  uint64_t dyn_stack_size = 0;
```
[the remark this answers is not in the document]
florian-kuebler: Done.

src/LinuxTracing/PerfEventReaders.cpp
```diff
@@ -120,17 +120,47 @@ std::unique_ptr<CallchainSamplePerfEvent> ConsumeCallchainSamplePerfEvent(
...
+  ring_buffer->ReadRawAtOffset(&dyn_stack_size, offset_of_size, sizeof(uint64_t));
```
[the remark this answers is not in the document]
florian-kuebler: You are completely right. Thanks for that.

src/LinuxTracing/PerfEventRecords.h
```diff
@@ -98,6 +98,12 @@ struct __attribute__((__packed__)) perf_event_sample_stack_user_8bytes {
   uint64_t dyn_size;
 };
 
+struct __attribute__((__packed__)) perf_event_sample_stack_user_128bytes {
+  uint64_t size;
+  char data[SAMPLE_STACK_USER_SIZE];
```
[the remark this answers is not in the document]
florian-kuebler: Good catch, thanks!

src/LinuxTracing/PerfEventRecords.h
florian-kuebler: Thanks. Done.

florian-kuebler (commented):
Thanks for the good review and spotting those issues!

Regarding the higher-level question: As discussed offline, I'll extend my performance overhead experiment today, before merging this change.

src/LinuxTracing/PerfEventReaders.cpp
```diff
@@ -120,17 +120,51 @@ std::unique_ptr<CallchainSamplePerfEvent> ConsumeCallchainSamplePerfEvent(
...
+  auto offset_of_ips = offsetof(perf_event_callchain_sample_fixed, nr) +
+                       sizeof(perf_event_callchain_sample_fixed::nr);
+  auto offset_of_regs_user_struct = offset_of_ips + size_of_ips_in_bytes;
+  auto offset_of_size = offset_of_regs_user_struct + sizeof(perf_event_sample_regs_user_all);
+  auto offset_of_data = offset_of_size + sizeof(uint64_t);
+
+  uint64_t size = 0;
+  ring_buffer->ReadRawAtOffset(&size, offset_of_size, sizeof(uint64_t));
+
+  auto offset_of_dyn_size = offset_of_data + (size * sizeof(char));
```
dpallotti, line 154: I honestly prefer `size_t` like in `ConsumeMmapPerfEvent`. I prefer being explicit in this low level code.

src/LinuxTracing/PerfEventOpen.h
```diff
@@ -103,6 +103,9 @@ static constexpr uint64_t SAMPLE_REGS_USER_SP_IP_ARGUMENTS =
 //  some setting.
 static constexpr uint16_t SAMPLE_STACK_USER_SIZE = 65000;
 
+// Arbitrary small value, that is still large enough to contain the complete last frame.
+static constexpr uint16_t SAMPLE_STACK_USER_SIZE_ON_CALLCHAINS = 128;
```
dpallotti, line 107: As this is used in a `struct` called `perf_event_sample_stack_user_128bytes`, and as the constant below is called `SAMPLE_STACK_USER_SIZE_8BYTES`, I would suggest `SAMPLE_STACK_USER_SIZE_128BYTES`. I know it's uglier but I think it's more consistent with the rest of the code.

dpallotti (commented):
Some comments on naming and comments. Looks good in general.

src/LinuxTracing/PerfEventOpen.cpp
```diff
@@ -92,6 +92,12 @@ int callchain_sample_event_open(uint64_t period_ns, pid_t pid, int32_t cpu) {
   pe.sample_max_stack = 127;
   pe.exclude_callchain_kernel = true;
 
+  // Also capture a small part of the stack and the registers to allow patching the callers of
+  // leaf functions. This is done by unwinding the first frame using DWARF.
```
dpallotti, line 96: The first two? Or is the first one enough and then we already have all the information to determine the caller?

src/LinuxTracing/PerfEventOpen.h
```diff
@@ -103,6 +103,10 @@ static constexpr uint64_t SAMPLE_REGS_USER_SP_IP_ARGUMENTS =
 //  some setting.
 static constexpr uint16_t SAMPLE_STACK_USER_SIZE = 65000;
 
+// Arbitrary small value, that is should be large enough to contain the complete last frame.
```
dpallotti, line 106:
```suggestion
// Arbitrary small value, that should be large enough to contain the complete last frame.
```

dpallotti (approved):

PR #2252, opened 2021-04-28
Import opengl32sw.dll with Orbit

beckerhe:
This is adding support for the LLVMPIPE OpenGL software renderer.

Our conan recipe has now an options which enables importing that
software renderer library such that it will be packaged with Orbit.

This also enables this option for the `msvc2019_relwithdebinfo_ninja`
which we use on the CI.

### DIFF
--- a/conanfile.py
+++ b/conanfile.py
@@ -31,7 +31,8 @@ class OrbitConan(ConanFile):
                "with_crash_handling": [True, False],
                "run_tests": [True, False],
                "run_python_tests": [True, False],
-               "build_target": "ANY"}
+               "build_target": "ANY",
+               "deploy_opengl_software_renderer": [True, False]}
     default_options = {"system_mesa": True,
                        "system_qt": True, "with_gui": True,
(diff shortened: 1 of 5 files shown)


beckerhe:
@danielfenner I think we should talk this through. Can you ping me when you look at this?

ronaldfw:
Do we need to add a LICENSE file or does that come with the conan package (or some other way)?

beckerhe:
> Do we need to add a LICENSE file or does that come with the conan package (or some other way)?

This comes automatically with the conan package and will automatically be added to our NOTICE file.

danielfenner (approved):
As discussed offline - lgtm

ronaldfw:
> This comes automatically with the conan package and will automatically be added to our NOTICE file.

Perfect, thanks!

ronaldfw (approved):
LGTM

beckerhe:
Thank you both for the review. I just rebased and also added the llvmpipe recipe as discussed with Daniel oob.

PR #2254, opened 2021-04-28
Package new version of Orbit.h

pierricgimmig:
The old version of Orbit.h has been renamed to LegacyOrbit.h and
will be kept until full deprecation of the old api implementation.


ronaldfw:
Just to make sure I fully understand this. Is the following correct:

- The new version is the one in Api/Orbit.h and that will now be packaged with the release for users
- The old version is only kept around for testing. If yes, is this testing needed for future releases?

Do users have to change anything when they upgrade to the new version or will things just work for them?

beckerhe (approved):
LGTM from the technical / deployment side of things. Thanks!

pierricgimmig:
> Just to make sure I fully understand this. Is the following correct:
> 
> * The new version is the one in Api/Orbit.h and that will now be packaged with the release for users
That's right, yes.

> * The old version is only kept around for testing. If yes, is this testing needed for future releases?
We support both versions of the Api for now, until we decide to fully deprecate the old one.  We have OrbitTestLegacyApi
executable which uses the old implementation and which we should use as part of our release testing.
 
> Do users have to change anything when they upgrade to the new version or will things just work for them?
The only thing a user will have to do is place the new "ORBIT_API_INSTANTIATE" macro in an implementation file. This will have to be communicated.

PR #2255, opened 2021-04-29
Disable Orbit Api on capture stop

pierricgimmig:
Make sure Orbit Api is fully disabled on capture stop by setting
the "enabled" flag to false which makes api functions early out
at call site. This change also makes it possible to disable the api
from the UI (it was only possible to turn it on, not off previously).

Test: Ran OrbitTest and verified that the "enabled" flag is indeed
set to false on capture stop and that we can toggle back and forth
with expected results.

### DIFF
--- a/src/Api/EnableInTracee.cpp
+++ b/src/Api/EnableInTracee.cpp
@@ -2,7 +2,7 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
 
-#include "Api/InitializeInTracee.h"
+#include "Api/EnableInTracee.h"
 
 #include <absl/base/casts.h>
 #include <absl/container/flat_hash_map.h>
@@ -56,7 +56,8 @@ static ErrorMessageOr<absl::flat_hash_map<std::string, ModuleInfo>> GetModulesBy
 }
 
 namespace orbit_api {
-ErrorMessageOr<void> InitializeInTracee(const CaptureOptions& capture_options) {
+ErrorMessageOr<void> EnableApiInTracee(const CaptureOptions& capture_options, bool enabled) {
+  SCOPED_TIMED_LOG("%s Api in tracee", enabled ? "Enabling" : "Disabling");
   if (capture_options.api_functions().size() == 0) {
     return ErrorMessage("No api table to initialize.");
   }
@@ -74,9 +75,9 @@ ErrorMessageOr<void> InitializeInTracee(const CaptureOptions& capture_options) {
 
   // Load liborbit.so and find api table initialization function.
   const std::string liborbit_path = orbit_base::GetExecutableDir() / "liborbit.so";
-  constexpr const char* kInitFunction = "orbit_api_initialize";
+  constexpr const char* kEnableFunction = "orbit_api_enable";
   OUTCOME_TRY(handle, DlopenInTracee(pid, liborbit_path, RTLD_NOW));
-  OUTCOME_TRY(orbit_init_function, DlsymInTracee(pid, handle, kInitFunction));
+  OUTCOME_TRY(orbit_api_enable_function, DlsymInTracee(pid, handle, kEnableFunction));
 
   // Initialize all api function tables.
   OUTCOME_TRY(modules_by_path, GetModulesByPathForPid(pid));
@@ -94,9 +95,9 @@ ErrorMessageOr<void> InitializeInTracee(const CaptureOptions& capture_options) {
         module_info->address_start() + api_function.address() - module_info->load_bias());
     OUTCOME_TRY(function_table_address, ExecuteInProcess(pid, api_function_address));
 
-    // Call "orbit_api_initialize" in tracee with table address and api version as parameters.
-    OUTCOME_TRY(ExecuteInProcess(pid, orbit_init_function, function_table_address,
-                                 api_function.api_version()));
+    // Call "orbit_api_enable" in tracee.
+    OUTCOME_TRY(ExecuteInProcess(pid, orbit_api_enable_function, function_table_address,
+                                 api_function.api_version(), enabled ? 1 : 0));
   }
 
   return outcome::success();
--- a/src/Api/EnableInTracee.cpp
+++ b/src/Api/EnableInTracee.cpp
@@ -56,7 +56,7 @@ ErrorMessageOr<absl::flat_hash_map<std::string, ModuleInfo>> GetModulesByPathFor
   return &module_info;
 }
 
-ErrorMessageOr<void> EnableApiInTracee(const CaptureOptions& capture_options, bool enabled) {
+ErrorMessageOr<void> SetApiEnabledInTracee(const CaptureOptions& capture_options, bool enabled) {
   SCOPED_TIMED_LOG("%s Api in tracee", enabled ? "Enabling" : "Disabling");
   if (capture_options.api_functions().size() == 0) {
     return ErrorMessage("No api table to initialize.");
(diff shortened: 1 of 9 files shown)


src/Api/Orbit.cpp
```diff
@@ -101,20 +100,25 @@ static void orbit_api_initialize_v0(uint64_t address) {
...
   if (api_version > kOrbitApiVersion) {
     ERROR(
         "Orbit API version in tracee (%u) is newer than the max supported version (%u). "
         "Some features will be unavailable.",
         api_version, kOrbitApiVersion);
   }
 
-  if (api_version >= 0) orbit_api_initialize_v0(address);
+  orbit_api_v0* api_v0 = absl::bit_cast<orbit_api_v0*>(address);
+  orbit_api_initialize_v0(api_v0);
```
danielfenner, line 121: I'd expected something like 

`if (!api_v0.initialized) orbit_api_initialize_v0(api_v0);`

We only need this on the first call, right? Probably doesn't really hurt to do it again though.

danielfenner (approved):
lgtm

src/Api/include/Api/Orbit.h
```diff
@@ -316,14 +316,14 @@ extern ORBIT_EXPORT void* orbit_api_get_function_table_address_v0();
   orbit_api_v0 g_orbit_api_v0; \
   void* orbit_api_get_function_table_address_v0() { return &g_orbit_api_v0; }
 
-inline bool orbit_api_active() {
-  bool active = g_orbit_api_v0.active;
+inline bool orbit_api_enabled() {
+  bool enabled = g_orbit_api_v0.initialized && g_orbit_api_v0.enabled;
```
dpallotti, line 320: Personal preference: In a way I preferred "active" as the name of the local variable to distinguish between the result of the `&&` and `g_orbit_api_v0.enabled` itself.

src/Api/Orbit.cpp
```diff
@@ -78,17 +78,16 @@ void orbit_api_async_string(const char* str, uint64_t id, orbit_api_color color)
   }
 }
 
-static void orbit_api_initialize_v0(uint64_t address) {
+static void orbit_api_initialize_v0(orbit_api_v0* api) {
   // The api function table is accessed by user code using this pattern:
   //
-  // bool active = api->active;
+  // bool enabled = api.initialized && api.enabled;
```
dpallotti, line 84: See below on "enabled" vs "active".

src/Api/include/Api/Orbit.h
dpallotti, line 320: Is it intended that you use `bool` here but `uint32_t` in `orbit_api_v0`? Are we assuming that C supports `bool` or not?

src/OrbitBase/Tracing.cpp
```diff
@@ -200,9 +200,9 @@ void InitializeTracing() {
   g_orbit_api_v0.track_uint64 = &orbit_api_track_uint64;
   g_orbit_api_v0.track_float = &orbit_api_track_float;
   g_orbit_api_v0.track_double = &orbit_api_track_double;
-  g_orbit_api_v0.initialized = 1;
   std::atomic_thread_fence(std::memory_order_release);
-  g_orbit_api_v0.active = 1;
+  g_orbit_api_v0.initialized = 1;
+  g_orbit_api_v0.enabled = 1;
```
dpallotti, line 205: Now that you have two assignments I'm not sure that all the reasoning around that fence still applies. Does it?

src/Api/Orbit.cpp
```diff
@@ -101,20 +100,25 @@ static void orbit_api_initialize_v0(uint64_t address) {
...
     ERROR(
         "Orbit API version in tracee (%u) is newer than the max supported version (%u). "
         "Some features will be unavailable.",
         api_version, kOrbitApiVersion);
   }
 
-  if (api_version >= 0) orbit_api_initialize_v0(address);
+  orbit_api_v0* api_v0 = absl::bit_cast<orbit_api_v0*>(address);
+  orbit_api_initialize_v0(api_v0);
+  api_v0->enabled = enabled;
```
dpallotti, line 122: As in the other comment, now you have this assignment following `api->initialized = 1` in `orbit_api_initialize_v0`. Are there new implications regarding atomicity here?

src/Api/include/Api/EnableInTracee.h
```diff
@@ -2,15 +2,16 @@
...
+#define ORBIT_API_ENABLE_IN_TRACEE_H_
 
 #include "OrbitBase/Result.h"
 #include "capture.pb.h"
 
 namespace orbit_api {
 
-ErrorMessageOr<void> InitializeInTracee(const orbit_grpc_protos::CaptureOptions& capture_options);
+ErrorMessageOr<void> EnableApiInTracee(const orbit_grpc_protos::CaptureOptions& capture_options,
+                                       bool enabled);
```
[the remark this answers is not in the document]
ronaldfw: Typo: decreases readability? 

And agree on the suggestion here.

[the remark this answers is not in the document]
dpallotti: > Typo: decreases readability?

Whoops, yes.

src/Api/Orbit.cpp
pierricgimmig: Yeah I agree, I moved the condition out of the function so that when we have more versions to initialize, we only check the `initialized` flag once.

src/Api/include/Api/EnableInTracee.h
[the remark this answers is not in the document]
pierricgimmig: Totally agree, done.

src/Api/include/Api/Orbit.h
pierricgimmig: Good, point, done.

src/OrbitBase/Tracing.cpp
pierricgimmig: I think so, yes. I added a paragraph in Orbit.cpp:

  // By the time we reach this, the "initialized" guard variable has been set to 1 and we know that
  // all function pointers have been written to and published to other cores by the use of
  // acquire/release fences. The "enabled" flag serves as a global toggle which is always used in
  // conjunction with the "initialized" flag to determine if the Api is active. See
  // orbit_api_active() in Orbit.h.

src/Api/Orbit.cpp
pierricgimmig: We do have atomicity but there could have been an ordering issue, which I don't think we have. I see "enabled" as a global toggle whose value is only valid if "initialized' is 1, in which case we know that all function pointers have been written.

src/Api/include/Api/Orbit.h
pierricgimmig: Oh good observation. I'll fix this in another PR since it's not related to the change. Also, I have a C only version of OrbitTest locally that I need to submit. Thanks.

src/Api/Orbit.cpp
pierricgimmig: Yep, done.

src/Api/EnableInTracee.cpp
```diff
@@ -55,8 +56,8 @@ static ErrorMessageOr<absl::flat_hash_map<std::string, ModuleInfo>> GetModulesBy
   return &module_info;
 }
 
-namespace orbit_api {
-ErrorMessageOr<void> InitializeInTracee(const CaptureOptions& capture_options) {
+ErrorMessageOr<void> EnableApiInTracee(const CaptureOptions& capture_options, bool enabled) {
```
dpallotti, line 59: I suggest calling this differently than the "public" `EnableApiInTracee` below. Maybe `EnableOrDisableApiInTracee`, or `EnableApiInTraceeInternal`.

src/OrbitBase/Tracing.cpp
dpallotti: Makes sense.

src/Api/Orbit.cpp
dpallotti: Yeah, with "atomicity" I meant the whole package, including ordering.

dpallotti (approved):

src/Api/include/Api/Orbit.h
dpallotti: Maybe create a bug (in case you haven't already).

PR #2256, opened 2021-04-29
Add utility methods to collect cgroup & process memory usage information

vickyliu-go4it:
With this change, we add utility methods to collect the following cgroup & process memory usage information:
* Basic memory usage information (collected for the memory track): 
  - process memory usage retrieved from the `rss` field in the `/proc/<pid>/stat` file;
  - cgroup memory limit retrieved from the `/sys/fs/cgroup/memory/<cgroup_name>/memory.limit_in_bytes` file;
  - cgroup memory usage retrieved from the `rss`  and `mapped_file` fields in the `/sys/fs/cgroup/memory/<cgroup_name>/memory.stat` file.
* Page fault statistics (collected for the page fault track):
  - system page fault statistics retrieved from the `minflt` and `majflt` fields in the `/proc/vmstat` file;
  - process page fault statistics retrieved from the `pgfault` and `pgmajfault` fields in the `/proc/<pid>stat` file;
  - cgroup page fault statistics retrieved from the  `pgfault` and `pgmajfault` fields in the `/sys/fs/cgroup/memory/<cgroup_name>/memory.stat` file.
* Detailed cgroup memory usage (collected for the memory track tooltips):
  - values retrieved from the `unevictable`, `inactive_anon`, `active_anon`, `inactive_file`, `active_file` fields.

Relevant bugs: http://b/185107554, http://b/185107588, http://b/187918073.

### DIFF
--- a/src/GrpcProtos/capture.proto
+++ b/src/GrpcProtos/capture.proto
@@ -325,6 +325,24 @@ message SystemMemoryUsage {
   int64 cached_kb = 6;
 }
 
+message ProcessMemoryUsage {
+  uint32 pid = 1;
+  uint64 timestamp_ns = 2;
+  // These fields are retrieved from /proc/<pid>/status.
+  int64 vm_rss_kb = 3;
+}
+
+message CGroupMemoryUsage {
+  string cgroup_name = 1;
+  uint64 timestamp_ns = 2;
+  // This field is retrieved from
+  // /sys/fs/cgroup/memory/<cgroup_name>/memory.stat.
+  int64 rss_bytes = 3;
+  // This field is retrieved from
+  // /sys/fs/cgroup/memory/<cgroup_name>/memory.limit_in_bytes.
+  int64 limit_bytes = 4;
+}
+
 message ModulesSnapshot {
   int32 pid = 1;
   uint64 timestamp_ns = 2;
--- a/src/GrpcProtos/capture.proto
+++ b/src/GrpcProtos/capture.proto
@@ -323,13 +323,18 @@ message SystemMemoryUsage {
   int64 available_kb = 4;
   int64 buffers_kb = 5;
   int64 cached_kb = 6;
+  // These fields are retrieved from /proc/vmstat.
+  int64 pgfault = 7;
+  int64 pgmajfault = 8;
 }
 
 message ProcessMemoryUsage {
   uint32 pid = 1;
   uint64 timestamp_ns = 2;
-  // These fields are retrieved from /proc/<pid>/status.
-  int64 vm_rss_kb = 3;
+  // These fields are retrieved from /proc/<pid>/stat.
+  int64 rss_pages = 3;
+  int64 minflt = 4;
+  int64 majflt = 5;
 }
 
 message CGroupMemoryUsage {
@@ -342,6 +347,8 @@ message CGroupMemoryUsage {
   // /sys/fs/cgroup/memory/<cgroup_name>/memory.stat.
   int64 rss_bytes = 4;
   int64 mapped_file_bytes = 5;
+  int64 pgfault = 6;
+  int64 pgmajfault = 7;
 }
 
 message ModulesSnapshot {
--- a/src/GrpcProtos/capture.proto
+++ b/src/GrpcProtos/capture.proto
@@ -336,11 +336,12 @@ message CGroupMemoryUsage {
   string cgroup_name = 1;
   uint64 timestamp_ns = 2;
   // This field is retrieved from
-  // /sys/fs/cgroup/memory/<cgroup_name>/memory.stat.
-  int64 rss_bytes = 3;
-  // This field is retrieved from
   // /sys/fs/cgroup/memory/<cgroup_name>/memory.limit_in_bytes.
-  int64 limit_bytes = 4;
+  int64 limit_bytes = 3;
+  // These fields are retrieved from
+  // /sys/fs/cgroup/memory/<cgroup_name>/memory.stat.
+  int64 rss_bytes = 4;
+  int64 mapped_file_bytes = 5;
 }
 
 message ModulesSnapshot {
--- a/src/GrpcProtos/capture.proto
+++ b/src/GrpcProtos/capture.proto
@@ -317,13 +317,15 @@ message ModuleUpdateEvent {
 
 message SystemMemoryUsage {
   uint64 timestamp_ns = 1;
-  // These fields are retrieved from /proc/meminfo.
+  // These fields are retrieved from /proc/meminfo. See
+  // https://man7.org/linux/man-pages/man5/proc.5.html for the documentation.
   int64 total_kb = 2;
   int64 free_kb = 3;
   int64 available_kb = 4;
   int64 buffers_kb = 5;
   int64 cached_kb = 6;
-  // These fields are retrieved from /proc/vmstat.
+  // These fields are retrieved from /proc/vmstat. See
+  // https://man7.org/linux/man-pages/man5/proc.5.html for the documentation.
   int64 pgfault = 7;
   int64 pgmajfault = 8;
 }
@@ -331,7 +333,8 @@ message SystemMemoryUsage {
 message ProcessMemoryUsage {
   uint32 pid = 1;
   uint64 timestamp_ns = 2;
-  // These fields are retrieved from /proc/<pid>/stat.
+  // These fields are retrieved from /proc/<pid>/stat. See
+  // https://man7.org/linux/man-pages/man5/proc.5.html for the documentation.
   int64 rss_pages = 3;
   int64 minflt = 4;
   int64 majflt = 5;
@@ -341,14 +344,23 @@ message CGroupMemoryUsage {
   string cgroup_name = 1;
   uint64 timestamp_ns = 2;
   // This field is retrieved from
-  // /sys/fs/cgroup/memory/<cgroup_name>/memory.limit_in_bytes.
+  // /sys/fs/cgroup/memory/<cgroup_name>/memory.limit_in_bytes. See
+  // https://www.kernel.org/doc/Documentation/cgroup-v1/memory.txt for the
+  // documentation.
   int64 limit_bytes = 3;
   // These fields are retrieved from
-  // /sys/fs/cgroup/memory/<cgroup_name>/memory.stat.
+  // /sys/fs/cgroup/memory/<cgroup_name>/memory.stat. See
+  // https://www.kernel.org/doc/Documentation/cgroup-v1/memory.txt section 5.2
+  // for the documentation.
   int64 rss_bytes = 4;
   int64 mapped_file_bytes = 5;
   int64 pgfault = 6;
   int64 pgmajfault = 7;
+  int64 unevictable_bytes = 8;
+  int64 inactive_anon_bytes = 9;
+  int64 active_anon_bytes = 10;
+  int64 inactive_file_bytes = 11;
+  int64 active_file_bytes = 12;
 }
 
 message ModulesSnapshot {
--- a/src/MemoryTracing/MemoryTracingUtils.cpp
+++ b/src/MemoryTracing/MemoryTracingUtils.cpp
@@ -24,13 +24,23 @@ using orbit_grpc_protos::kMissingInfo;
 using orbit_grpc_protos::ProcessMemoryUsage;
 using orbit_grpc_protos::SystemMemoryUsage;
 
+SystemMemoryUsage CreateAndInitializeSystemMemoryUsage() {
+  SystemMemoryUsage system_memory_usage;
+  system_memory_usage.set_total_kb(kMissingInfo);
+  system_memory_usage.set_free_kb(kMissingInfo);
+  system_memory_usage.set_available_kb(kMissingInfo);
+  system_memory_usage.set_buffers_kb(kMissingInfo);
+  system_memory_usage.set_cached_kb(kMissingInfo);
+  system_memory_usage.set_pgfault(kMissingInfo);
+  system_memory_usage.set_pgmajfault(kMissingInfo);
+  return system_memory_usage;
+}
+
 void GetValuesFromMemInfo(std::string_view meminfo_content,
                           SystemMemoryUsage* system_memory_usage) {
   if (meminfo_content.empty()) return;
 
   std::vector<std::string> lines = absl::StrSplit(meminfo_content, '\n');
-  if (lines.empty()) return;
-
   constexpr size_t kNumLines = 5;
   std::vector<std::string> top_lines(
       lines.begin(), lines.begin() + (lines.size() > kNumLines ? kNumLines : lines.size()));
@@ -66,8 +76,6 @@ void GetValuesFromVmStat(std::string_view vmstat_content, SystemMemoryUsage* sys
   if (vmstat_content.empty()) return;
 
   std::vector<std::string> lines = absl::StrSplit(vmstat_content, '\n');
-  if (lines.empty()) return;
-
   for (std::string_view line : lines) {
     // Each line of the /proc/vmstat file consists a single name-value pair, delimited by white
     // space. In /proc/vmstat, the pgfault and pgmajfault fields report cumulative values.
@@ -86,15 +94,8 @@ void GetValuesFromVmStat(std::string_view vmstat_content, SystemMemoryUsage* sys
 }
 
 ErrorMessageOr<SystemMemoryUsage> GetSystemMemoryUsage() noexcept {
-  SystemMemoryUsage system_memory_usage;
+  SystemMemoryUsage system_memory_usage = CreateAndInitializeSystemMemoryUsage();
   system_memory_usage.set_timestamp_ns(orbit_base::CaptureTimestampNs());
-  system_memory_usage.set_total_kb(kMissingInfo);
-  system_memory_usage.set_free_kb(kMissingInfo);
-  system_memory_usage.set_available_kb(kMissingInfo);
-  system_memory_usage.set_buffers_kb(kMissingInfo);
-  system_memory_usage.set_cached_kb(kMissingInfo);
-  system_memory_usage.set_pgfault(kMissingInfo);
-  system_memory_usage.set_pgmajfault(kMissingInfo);
 
   // Extract system-wide memory usage from the /proc/meminfo file.
   ErrorMessageOr<std::string> reading_result = orbit_base::ReadFileToString("/proc/meminfo");
@@ -115,6 +116,14 @@ ErrorMessageOr<SystemMemoryUsage> GetSystemMemoryUsage() noexcept {
   return system_memory_usage;
 }
 
+ProcessMemoryUsage CreateAndInitializeProcessMemoryUsage() {
+  ProcessMemoryUsage process_memory_usage;
+  process_memory_usage.set_rss_pages(kMissingInfo);
+  process_memory_usage.set_minflt(kMissingInfo);
+  process_memory_usage.set_majflt(kMissingInfo);
+  return process_memory_usage;
+}
+
 void GetValuesFromProcessStat(std::string_view stat_content,
                               ProcessMemoryUsage* process_memory_usage) {
   if (stat_content.empty()) return;
@@ -122,7 +131,7 @@ void GetValuesFromProcessStat(std::string_view stat_content,
   // According to the kernel code
   // https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/fs/proc/array.c,
   // the /proc/<PID>/stat file records 52 process status information on a single line, in a fixed
-  // order. We are intersted in the following fields:
+  // order. We are interested in the following fields:
   //   Field index | Name   | Format | Meaning
   //    24         | rss    | %ld    | # of pages the process has in real memory
   //    10         | minflt | %lu    | # of minor faults the process has made
@@ -137,7 +146,7 @@ void GetValuesFromProcessStat(std::string_view stat_content,
 }
 
 ErrorMessageOr<ProcessMemoryUsage> GetProcessMemoryUsage(uint32_t pid) noexcept {
-  ProcessMemoryUsage process_memory_usage;
+  ProcessMemoryUsage process_memory_usage = CreateAndInitializeProcessMemoryUsage();
   process_memory_usage.set_pid(pid);
   process_memory_usage.set_timestamp_ns(orbit_base::CaptureTimestampNs());
 
@@ -153,12 +162,25 @@ ErrorMessageOr<ProcessMemoryUsage> GetProcessMemoryUsage(uint32_t pid) noexcept 
   return process_memory_usage;
 }
 
+CGroupMemoryUsage CreateAndInitializeCGroupMemoryUsage() {
+  CGroupMemoryUsage cgroup_memory_usage;
+  cgroup_memory_usage.set_limit_bytes(kMissingInfo);
+  cgroup_memory_usage.set_rss_bytes(kMissingInfo);
+  cgroup_memory_usage.set_mapped_file_bytes(kMissingInfo);
+  cgroup_memory_usage.set_pgfault(kMissingInfo);
+  cgroup_memory_usage.set_pgmajfault(kMissingInfo);
+  cgroup_memory_usage.set_unevictable_bytes(kMissingInfo);
+  cgroup_memory_usage.set_inactive_anon_bytes(kMissingInfo);
+  cgroup_memory_usage.set_active_anon_bytes(kMissingInfo);
+  cgroup_memory_usage.set_inactive_file_bytes(kMissingInfo);
+  cgroup_memory_usage.set_active_file_bytes(kMissingInfo);
+  return cgroup_memory_usage;
+}
+
 std::string GetProcessMemoryCGroupName(std::string_view cgroup_content) {
   if (cgroup_content.empty()) return "";
 
   std::vector<std::string> lines = absl::StrSplit(cgroup_content, '\n');
-  if (lines.empty()) return "";
-
   for (std::string_view line : lines) {
     std::vector<std::string> splits = absl::StrSplit(line, absl::MaxSplits(':', 2));
     // If find the memory cgroup, return the cgroup name without the leading "/".
@@ -184,8 +206,6 @@ void GetValuesFromCGroupMemoryStat(std::string_view memory_stat_content,
   if (memory_stat_content.empty()) return;
 
   std::vector<std::string> lines = absl::StrSplit(memory_stat_content, '\n');
-  if (lines.empty()) return;
-
   for (std::string_view line : lines) {
     std::vector<std::string> splits = absl::StrSplit(line, ' ', absl::SkipWhitespace{});
     // According to the document https://www.kernel.org/doc/Documentation/cgroup-v1/memory.txt:
@@ -204,6 +224,16 @@ void GetValuesFromCGroupMemoryStat(std::string_view memory_stat_content,
       cgroup_memory_usage->set_pgfault(value);
     } else if (splits[0] == "pgmajfault") {
       cgroup_memory_usage->set_pgmajfault(value);
+    } else if (splits[0] == "unevictable") {
+      cgroup_memory_usage->set_unevictable_bytes(value);
+    } else if (splits[0] == "inactive_anon") {
+      cgroup_memory_usage->set_inactive_anon_bytes(value);
+    } else if (splits[0] == "active_anon") {
+      cgroup_memory_usage->set_active_anon_bytes(value);
+    } else if (splits[0] == "inactive_file") {
+      cgroup_memory_usage->set_inactive_file_bytes(value);
+    } else if (splits[0] == "active_file") {
+      cgroup_memory_usage->set_active_file_bytes(value);
     }
   }
 }
@@ -226,12 +256,9 @@ ErrorMessageOr<CGroupMemoryUsage> GetCGroupMemoryUsage(uint32_t pid) noexcept {
     return ErrorMessage{std::move(error_message)};
   }
 
-  CGroupMemoryUsage cgroup_memory_usage;
+  CGroupMemoryUsage cgroup_memory_usage = CreateAndInitializeCGroupMemoryUsage();
   cgroup_memory_usage.set_cgroup_name(cgroup_name);
   cgroup_memory_usage.set_timestamp_ns(current_timestamp_ns);
-  cgroup_memory_usage.set_limit_bytes(kMissingInfo);
-  cgroup_memory_usage.set_rss_bytes(kMissingInfo);
-  cgroup_memory_usage.set_mapped_file_bytes(kMissingInfo);
 
   // Extract cgroup memory limit from /sys/fs/cgroup/memory/<cgroup_name>/memory.limit_in_bytes.
   reading_result = orbit_base::ReadFileToString(
--- a/src/MemoryTracing/MemoryTracingUtils.cpp
+++ b/src/MemoryTracing/MemoryTracingUtils.cpp
@@ -143,36 +143,40 @@ std::string GetProcessMemoryCGroupName(std::string_view cgroup_content) {
   return "";
 }
 
-int64_t GetCGroupMemoryLimitInBytes(std::string_view memory_limit_in_bytes_content) {
-  if (memory_limit_in_bytes_content.empty()) return kMissingInfo;
+void GetCGroupMemoryLimitInBytes(std::string_view memory_limit_in_bytes_content,
+                                 CGroupMemoryUsage* cgroup_memory_usage) {
+  if (memory_limit_in_bytes_content.empty()) return;
 
   // The memory.limit_in_bytes file use "bytes" as the size unit.
   int64_t memory_limit_in_bytes;
   if (absl::SimpleAtoi(memory_limit_in_bytes_content, &memory_limit_in_bytes)) {
-    return memory_limit_in_bytes;
+    cgroup_memory_usage->set_limit_bytes(memory_limit_in_bytes);
   }
-
-  return kMissingInfo;
 }
 
-int64_t GetRssFromCGroupMemoryStat(std::string_view memory_stat_content) {
-  if (memory_stat_content.empty()) return kMissingInfo;
+void GetValuesFromCGroupMemoryStat(std::string_view memory_stat_content,
+                                   CGroupMemoryUsage* cgroup_memory_usage) {
+  if (memory_stat_content.empty()) return;
 
   std::vector<std::string> lines = absl::StrSplit(memory_stat_content, '\n');
-  if (lines.empty()) return kMissingInfo;
+  if (lines.empty()) return;
 
   for (std::string_view line : lines) {
     std::vector<std::string> splits = absl::StrSplit(line, ' ', absl::SkipWhitespace{});
-    if (splits.size() == 2 && splits[0] == "rss") {
-      // According to the document https://www.kernel.org/doc/Documentation/cgroup-v1/memory.txt,
-      // the memory.stat file use "bytes" as the size unit.
-      int64_t rss_in_bytes;
-      if (absl::SimpleAtoi(splits[1], &rss_in_bytes)) return rss_in_bytes;
-      return kMissingInfo;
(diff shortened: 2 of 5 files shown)


src/GrpcProtos/capture.proto
```diff
@@ -325,6 +325,24 @@ message SystemMemoryUsage {
   int64 cached_kb = 6;
 }
 
+message ProcessMemoryUsage {
+  uint32 pid = 1;
+  uint64 timestamp_ns = 2;
+  // These fields are retrieved from /proc/<pid>/status.
+  int64 rss_kb = 3;
```
ronaldfw, line 332: I would use the exact same names as in /proc/pid/status, adjusted to our code style. Adding the unit at the end is a good idea, though. That is, this should be `vm_rss_kb`.

src/GrpcProtos/capture.proto
```diff
@@ -325,6 +325,24 @@ message SystemMemoryUsage {
...
+}
+
+message CGroupMemoryUsage {
+  string cgroup_name = 1;
+  uint64 timestamp_ns = 2;
+  // This field is retrieved from
+  // /sys/fs/cgroup/memory/<cgroup_name>/memory.stat.
+  int64 rss_kb = 3;
```
ronaldfw, line 340: Since the original value in the memory.stat file is in bytes, I would also keep the value in bytes. The rounding practically will not make a difference for users, but it's easier to avoid mistakes when parsing the data and when someone else needs to touch the code in the future. Reading this here, it's easy to assume that the value in the memory.stat file is in kb.

src/GrpcProtos/capture.proto
```diff
@@ -325,6 +325,24 @@ message SystemMemoryUsage {
...
+  // This field is retrieved from
+  // /sys/fs/cgroup/memory/<cgroup_name>/memory.limit_in_bytes.
+  int64 limit_kb = 4;
```
ronaldfw, line 343: Same here, store as bytes.

src/MemoryTracing/MemoryTracingUtils.cpp
```diff
@@ -85,4 +90,139 @@ std::optional<SystemMemoryUsage> GetSystemMemoryUsage() noexcept {
   return parsing_result;
 }
 
+int64_t ParseProcessStatus(const std::string& status_content) {
```
ronaldfw, line 93: I would consider a different name for this to make it clearer what this function is doing. For example, GetVmRssFromProcessStatus. You can also make it more generic and call it GetValueFromProcessStatus and pass both the status file content and desired field (VmRSS in this case) into the function.

ronaldfw, line 93: Nit: const std::string& should be string_view.

src/MemoryTracing/MemoryTracingUtils.cpp
```diff
@@ -85,4 +90,139 @@ std::optional<SystemMemoryUsage> GetSystemMemoryUsage() noexcept {
...
+
+  ProcessMemoryUsage process_memory_usage;
+  process_memory_usage.set_pid(pid);
+  process_memory_usage.set_timestamp_ns(current_timestamp_ns);
+  process_memory_usage.set_rss_kb(process_rss_kb);
+
+  return process_memory_usage;
+}
+
+std::string ParseProcessCGroup(const std::string& cgroup_content) {
```
ronaldfw, line 136: Similarly here, make the name more specific to what the function is actually doing.

ronaldfw (commented):
Not sure if this was ready for review, yet, but I'm interested in this topic and had a quick look. Mostly LGTM modulo some comments. I think throughout you should change const string& to string_view.

src/GrpcProtos/capture.proto
vickyliu-go4it: Make sense! Changed to use `vm_rss_kb`. Thanks for the suggestion!

src/GrpcProtos/capture.proto
vickyliu-go4it: Totally agree that use the same unit size as the original value would be less misleading! Changed to use `rss_bytes` here. Thanks!

src/GrpcProtos/capture.proto
vickyliu-go4it: Rename as `limit_bytes` and store the value in bytes now. Thanks!

src/MemoryTracing/MemoryTracingUtils.cpp
vickyliu-go4it: Thanks for suggesting the name! Changed to `GetVmRssFromProcessStatus` now. Thanks!

vickyliu-go4it: Thanks for bringing this to my attention! Changed accordingly.

src/MemoryTracing/MemoryTracingUtils.cpp
vickyliu-go4it: Thanks for the suggestion! Rename it as `GetProcessMemoryCGroupName` now.

src/MemoryTracing/MemoryTracingUtils.cpp
```diff
@@ -85,4 +88,137 @@ std::optional<SystemMemoryUsage> GetSystemMemoryUsage() noexcept {
...
+}
+
+std::optional<ProcessMemoryUsage> GetProcessMemoryUsage(uint32_t pid) noexcept {
+  uint64_t current_timestamp_ns = orbit_base::CaptureTimestampNs();
+
+  // Extract the VmRSS value in kilobytes from the /proc/<pid>/status file.
+  ErrorMessageOr<std::string> reading_result =
+      orbit_base::ReadFileToString(absl::StrFormat("/proc/%d/status", pid));
+  if (!reading_result.has_value()) {
+    ERROR("%s", reading_result.error().message());
```
ronaldfw, line 121: Since you can get an error here, return ErrorMessageOr instead of optional?

src/MemoryTracing/MemoryTracingUtils.cpp
```diff
@@ -85,4 +88,137 @@ std::optional<SystemMemoryUsage> GetSystemMemoryUsage() noexcept {
...
+
+std::string GetProcessMemoryCGroupName(std::string_view cgroup_content) {
+  if (cgroup_content.empty()) return "";
+
+  std::vector<std::string> lines = absl::StrSplit(cgroup_content, '\n');
+  if (lines.empty()) return "";
+
+  for (std::string_view line : lines) {
+    std::vector<std::string> splits = absl::StrSplit(line, absl::MaxSplits(':', 2));
+    CHECK(splits.size() == 3);
```
ronaldfw, line 142: Similarly here, consider returning an empty string when the format is not correct.

src/MemoryTracing/MemoryTracingUtils.cpp
```diff
@@ -85,4 +88,137 @@ std::optional<SystemMemoryUsage> GetSystemMemoryUsage() noexcept {
...
+
+  return "";
+}
+
+int64_t GetCGroupMemoryLimitInBytes(std::string_view memory_limit_in_bytes_content) {
+  if (memory_limit_in_bytes_content.empty()) return kMissingInfo;
+
+  // The memory.limit_in_bytes file use "bytes" as the size unit.
+  int64_t memory_limit_in_bytes;
+  if (!absl::SimpleAtoi(memory_limit_in_bytes_content, &memory_limit_in_bytes)) return kMissingInfo;
```
ronaldfw, line 155: Maybe I'm missing something why this is necessary, but I noticed you use different styles how to handle the if here and above. 

Above it's

```
      int64_t rss_kb;
      if (absl::SimpleAtoi(splits[1], &rss_kb)) return rss_kb;
      return kMissingInfo;
```
so the success case is handled in the `if` block. 

Here it's
```
int64_t memory_limit_in_bytes;
if (!absl::SimpleAtoi(memory_limit_in_bytes_content, &memory_limit_in_bytes)) return kMissingInfo;
return memory_limit_in_bytes;
```
so the failure case is handled in the `if` block.

This should be consistent.

src/MemoryTracing/MemoryTracingUtils.cpp
```diff
@@ -85,4 +88,137 @@ std::optional<SystemMemoryUsage> GetSystemMemoryUsage() noexcept {
...
+  if (lines.empty()) return kMissingInfo;
+
+  for (std::string_view line : lines) {
+    std::vector<std::string> splits = absl::StrSplit(line, ' ', absl::SkipWhitespace{});
+    CHECK(splits.size() == 2);
+    if (splits[0] == "rss") {
+      // According to the document https://www.kernel.org/doc/Documentation/cgroup-v1/memory.txt,
+      // the memory.stat file use "bytes" as the size unit.
+      int64_t rss_in_bytes;
+      if (!absl::SimpleAtoi(splits[1], &rss_in_bytes)) return kMissingInfo;
```
ronaldfw, line 173: Similarly here, this should be consistent with other instances of this pattern.

src/MemoryTracing/MemoryTracingUtils.cpp
```diff
@@ -85,4 +88,137 @@ std::optional<SystemMemoryUsage> GetSystemMemoryUsage() noexcept {
...
+}
+
+std::optional<CGroupMemoryUsage> GetCGroupMemoryUsage(uint32_t pid) noexcept {
+  uint64_t current_timestamp_ns = orbit_base::CaptureTimestampNs();
+
+  // Extract cgroup name from the /proc/<pid>/cgroup file.
+  ErrorMessageOr<std::string> reading_result =
+      orbit_base::ReadFileToString(absl::StrFormat("/proc/%d/cgroup", pid));
+  if (!reading_result.has_value()) {
+    ERROR("%s", reading_result.error().message());
```
ronaldfw, line 188: Consider returning ErrorMessageOr.

src/MemoryTracing/MemoryTracingUtils.h
```diff
@@ -5,14 +5,23 @@
...
 
+#include <string_view>
+
 #include "capture.pb.h"
 
 namespace orbit_memory_tracing {
 
-std::optional<orbit_grpc_protos::SystemMemoryUsage> ParseMemInfo(
-    const std::string& meminfo_content);
+std::optional<orbit_grpc_protos::SystemMemoryUsage> ParseMemInfo(std::string_view meminfo_content);
```
ronaldfw, line 14: All functions that return something should be marked with `[[nodiscard]]`.

src/MemoryTracing/MemoryTracingUtils.cpp
```diff
@@ -85,4 +88,137 @@ std::optional<SystemMemoryUsage> GetSystemMemoryUsage() noexcept {
...
+  if (lines.empty()) return kMissingInfo;
+
+  for (std::string_view line : lines) {
+    if (absl::StartsWith(line, "VmRSS")) {
+      std::vector<std::string> splits = absl::StrSplit(line, ' ', absl::SkipWhitespace{});
+      // According to the kernel code
+      // https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/fs/proc/task_mmu.c,
+      // the size unit of the VmRSS field in the /proc/<PID>/status file is fixed to "kB", which
+      // implies 1024 Bytes. We report in "KB" in ProcessSystemUsage.
+      CHECK(splits.size() == 3 && splits[2] == "kB");
```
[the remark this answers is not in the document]
vickyliu-go4it: Thanks for the suggestion! I changed to return `kMissingInfo` or empty string when the file content format or the size unit is incorrect.

src/MemoryTracing/MemoryTracingUtils.cpp
vickyliu-go4it: Good idea! Changed to return `ErrorMessageOr` in `GetCGroupMemoryUsage` and `GetProcessMemoryUsage`.
Also add one more commit to change the `GetSystemMemoryUsage` to return `ErrorMessageOr`.

Thanks!

src/MemoryTracing/MemoryTracingUtils.cpp
vickyliu-go4it: Changed accordingly. Thanks!

src/MemoryTracing/MemoryTracingUtils.cpp
vickyliu-go4it: Changed accordingly. Thanks!

src/MemoryTracing/MemoryTracingUtils.cpp
vickyliu-go4it: Thanks for the suggestion! Changed accordingly.

src/MemoryTracing/MemoryTracingUtils.h
vickyliu-go4it: Thanks for pointing out this! Changed accordingly.

src/MemoryTracing/MemoryTracingUtils.h
```diff
@@ -22,8 +22,10 @@ namespace orbit_memory_tracing {
     uint32_t pid) noexcept;
 
 [[nodiscard]] std::string GetProcessMemoryCGroupName(std::string_view cgroup_content);
-[[nodiscard]] int64_t GetCGroupMemoryLimitInBytes(std::string_view memory_limit_in_bytes_content);
-[[nodiscard]] int64_t GetRssFromCGroupMemoryStat(std::string_view memory_stat_content);
+void GetCGroupMemoryLimitInBytes(std::string_view memory_limit_in_bytes_content,
```
karupayun, line 27: I'm not against this change (actually it makes sense), but I would like to have @ronaldfw opinion here (about having a void Get method with the only returning value in the parameters).

src/GrpcProtos/capture.proto
```diff
@@ -323,13 +323,18 @@ message SystemMemoryUsage {
   int64 available_kb = 4;
   int64 buffers_kb = 5;
   int64 cached_kb = 6;
+  // These fields are retrieved from /proc/vmstat.
+  int64 pgfault = 7;
```
karupayun, line 327: shouldn't we include minor in the name?

minorpgfault and majpgfault?

karupayun (commented):
Hi Vicky! I left comments for the first 5 commits.

src/MemoryTracing/MemoryTracingUtils.h
vickyliu-go4it: Thanks for the suggestion!

Fields in `CGroupMemoryUsage` are collected from different files. I changed this as a void Get method in order to keep consistent with the `GetValuesFromCGroupMemoryStat` method. But I'm also fine with returning the value.

vickyliu-go4it: Thanks for the suggestion!

Fields in `CGroupMemoryUsage` are collected from different files. I changed this as a void Get method in order to keep consistent with the `GetValuesFromCGroupMemoryStat` method. But I'm also fine with returning the value.

src/GrpcProtos/capture.proto
vickyliu-go4it: Thanks for the suggestion!

For these fields, we are using the exact same names as in the file `memory.stat`. The `memory.stat.pgmajfault` reports the number of major page faults but the `memory.stat.pgfault` reports the total number of page faults (rather than the number of minor page faults).

I would like to extract the raw value in the file and do the conversion to minor page faults when we need to use these values. What do you think?

src/MemoryTracing/MemoryTracingUtilsTest.cpp
```diff
@@ -120,4 +120,175 @@ SwapCached:      0 kB)",
...
+  EXPECT_EQ(parsing_result, kCGroupName);
+
+  parsing_result = GetProcessMemoryCGroupName(kPartialProcessCGroup);
+  EXPECT_TRUE(parsing_result.empty());
+
+  parsing_result = GetProcessMemoryCGroupName(kEmptyProcessCGroup);
+  EXPECT_TRUE(parsing_result.empty());
+}
+
+TEST(MemoryUtils, GetCGroupMemoryLimitInBytes) {
```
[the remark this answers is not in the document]
vickyliu-go4it: Make sense. I will erase this test!

src/MemoryTracing/MemoryTracingUtils.cpp
```diff
@@ -24,18 +24,12 @@ using orbit_grpc_protos::kMissingInfo;
...
-  memory_info.set_total_kb(kMissingInfo);
-  memory_info.set_free_kb(kMissingInfo);
-  memory_info.set_available_kb(kMissingInfo);
-  memory_info.set_buffers_kb(kMissingInfo);
-  memory_info.set_cached_kb(kMissingInfo);
-
-  if (meminfo_content.empty()) return memory_info;
+void GetValuesFromMemInfo(std::string_view meminfo_content,
+                          SystemMemoryUsage* system_memory_usage) {
+  if (meminfo_content.empty()) return;
```
karupayun, line 29: Shouldn't in this case reset the values from system_memory_usage to "kMissingInfo"?

src/MemoryTracing/MemoryTracingUtils.cpp
```diff
@@ -24,18 +24,12 @@ using orbit_grpc_protos::kMissingInfo;
...
 
   std::vector<std::string> lines = absl::StrSplit(meminfo_content, '\n');
-  if (lines.empty()) return memory_info;
+  if (lines.empty()) return;
```
karupayun, line 32: this lines.empty() is not the same case than meminfo_content.empty()?

src/MemoryTracing/MemoryTracingUtils.cpp
```diff
@@ -55,75 +49,106 @@ SystemMemoryUsage ParseMemInfo(std::string_view meminfo_content) {
...
+    }
+  }
 }
 
 ErrorMessageOr<SystemMemoryUsage> GetSystemMemoryUsage() noexcept {
-  uint64_t current_timestamp_ns = orbit_base::CaptureTimestampNs();
-
+  SystemMemoryUsage system_memory_usage;
+  system_memory_usage.set_timestamp_ns(orbit_base::CaptureTimestampNs());
+  system_memory_usage.set_total_kb(kMissingInfo);
```
karupayun, line 91: Again, I would prefer to have all of them initialized by default to kMissingInfo

src/MemoryTracing/MemoryTracingUtils.cpp
```diff
@@ -55,75 +49,106 @@ SystemMemoryUsage ParseMemInfo(std::string_view meminfo_content) {
...
+}
+
+void GetValuesFromProcessStat(std::string_view stat_content,
+                              ProcessMemoryUsage* process_memory_usage) {
+  if (stat_content.empty()) return;
+
+  // According to the kernel code
+  // https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/fs/proc/array.c,
+  // the /proc/<PID>/stat file records 52 process status information on a single line, in a fixed
+  // order. We are intersted in the following fields:
```
karupayun, line 125: nit: "interested"

src/MemoryTracing/MemoryTracingUtils.cpp
```diff
@@ -55,75 +49,106 @@ SystemMemoryUsage ParseMemInfo(std::string_view meminfo_content) {
...
+  //   Field index | Name   | Format | Meaning
+  //    24         | rss    | %ld    | # of pages the process has in real memory
+  //    10         | minflt | %lu    | # of minor faults the process has made
+  //    12         | majflt | %lu    | # of major faults the process has made
+  std::vector<std::string> splits = absl::StrSplit(stat_content, ' ', absl::SkipWhitespace{});
+  if (splits.size() != 52) return;
```
karupayun, line 131: Why returning here? If we are sure we always have 52 fields, I would print an error msg and if not I would like to know to think in the fields that matter for us are still there in the same order.

Lastly, again, I would reset the process_memory_usage.

src/GrpcProtos/capture.proto
```diff
@@ -349,6 +349,11 @@ message CGroupMemoryUsage {
   int64 mapped_file_bytes = 5;
   int64 pgfault = 6;
   int64 pgmajfault = 7;
+  int64 unevictable_bytes = 8;
```
karupayun, line 352: Maybe a link to some documentation is enough here for every field, or a short comment in each one.

karupayun (commented):
Hi Vicky! I've left some comments. 

I'm glad to see the memory feature growing and becoming nicer in each PR. :)

src/MemoryTracing/MemoryTracingUtilsTest.cpp
```diff
@@ -115,89 +147,224 @@ SwapCached:      0 kB)",
...
-
-  parsing_result = GetVmRssFromProcessStatus(kEmptyProcessStatus);
-  EXPECT_EQ(parsing_result, kMissingInfo);
+TEST(MemoryUtils, GetValuesFromProcessStat) {
+  const int kRssPages = 2793;
+  const int kMinorPageFaults = 20;
+  const int kMajorPageFaults = 1;
+
+  const std::string kValidProcessStat = absl::Substitute(
+      R"(9562 (TargetProcess) S 9561 9561 9561 0 -1 123456789 $1 3173 $2 0 7 18 1 7 20 0 10 0 123456789 123456789 $0 123456789 1 1 0 0 0 0 0 0 2 0 0 0 17 6 0 0 0 0 0 0 0 0 0 0 0 0 0)",
```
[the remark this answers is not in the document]
vickyliu-go4it: Maybe because this is a raw string? I found more details about how raw strings are formatted in go/clang-format-raw-string-literal.

src/MemoryTracing/MemoryTracingUtils.cpp
vickyliu-go4it: As we discussed offline: I will move these value setting lines to the corresponding Get method. If so, we can remove those Reset methods in the unit test.

src/GrpcProtos/capture.proto
vickyliu-go4it: Thanks for the suggestion. I added the link to the documentation to the memory protos.

src/MemoryTracing/MemoryTracingUtils.cpp
vickyliu-go4it: Done! Thanks for pointing out this typo.

src/MemoryTracing/MemoryTracingUtils.cpp
vickyliu-go4it: Ah right, I didn't notice this. Thanks for pointing out it! 

I removed some redundant `lines.empty()`.

src/GrpcProtos/capture.proto
```diff
@@ -323,13 +323,18 @@ message SystemMemoryUsage {
...
 }
 
 message ProcessMemoryUsage {
   uint32 pid = 1;
   uint64 timestamp_ns = 2;
-  // These fields are retrieved from /proc/<pid>/status.
-  int64 vm_rss_kb = 3;
+  // These fields are retrieved from /proc/<pid>/stat.
+  int64 rss_pages = 3;
+  int64 minflt = 4;
```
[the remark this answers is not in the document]
vickyliu-go4it: The name `minflt` is provided in the `/proc/<pid>/stat` file and it reports the number of minor page faults. For the system & cgroup ones, we collect the `pgfault` which reports the number of total page faults. 

Hence I prefer to keep the same name as the file.

src/GrpcProtos/capture.proto
```diff
@@ -325,6 +325,24 @@ message SystemMemoryUsage {
   int64 cached_kb = 6;
 }
 
+message ProcessMemoryUsage {
+  uint32 pid = 1;
+  uint64 timestamp_ns = 2;
+  // These fields are retrieved from /proc/<pid>/status.
+  int64 vm_rss_kb = 3;
```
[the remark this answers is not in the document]
vickyliu-go4it: As we discussed offline, added the link to the official documentation.

src/GrpcProtos/capture.proto
```diff
@@ -325,6 +325,24 @@ message SystemMemoryUsage {
...
+}
+
+message CGroupMemoryUsage {
+  string cgroup_name = 1;
+  uint64 timestamp_ns = 2;
+  // This field is retrieved from
+  // /sys/fs/cgroup/memory/<cgroup_name>/memory.stat.
+  int64 rss_bytes = 3;
```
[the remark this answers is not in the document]
vickyliu-go4it: As we discussed offline, added the link to the official documentation.

src/MemoryTracing/MemoryTracingUtils.h
```diff
@@ -5,14 +5,23 @@
...
-std::optional<orbit_grpc_protos::SystemMemoryUsage> ParseMemInfo(
-    const std::string& meminfo_content);
+std::optional<orbit_grpc_protos::SystemMemoryUsage> ParseMemInfo(std::string_view meminfo_content);
 std::optional<orbit_grpc_protos::SystemMemoryUsage> GetSystemMemoryUsage() noexcept;
 
+int64_t GetVmRssFromProcessStatus(std::string_view status_content);
+std::optional<orbit_grpc_protos::ProcessMemoryUsage> GetProcessMemoryUsage(uint32_t pid) noexcept;
+
+std::string GetProcessMemoryCGroupName(std::string_view cgroup_content);
+int64_t GetCGroupMemoryLimitInBytes(std::string_view memory_limit_in_bytes_content);
```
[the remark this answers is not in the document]
vickyliu-go4it: Thanks for the suggestion! As in the latest updates, we also pass the `CGroupMemoryUsage` as the parameter and update the `CGroupMemoryUsage::limit_bytes` field accordingly, I prefer to keep the current name.

src/GrpcProtos/capture.proto
```diff
@@ -336,11 +336,12 @@ message CGroupMemoryUsage {
...
-  // /sys/fs/cgroup/memory/<cgroup_name>/memory.stat.
-  int64 rss_bytes = 3;
-  // This field is retrieved from
   // /sys/fs/cgroup/memory/<cgroup_name>/memory.limit_in_bytes.
-  int64 limit_bytes = 4;
+  int64 limit_bytes = 3;
+  // These fields are retrieved from
+  // /sys/fs/cgroup/memory/<cgroup_name>/memory.stat.
+  int64 rss_bytes = 4;
+  int64 mapped_file_bytes = 5;
```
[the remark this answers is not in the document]
vickyliu-go4it: Thanks for the suggestion! As we discussed offline, added the link to the official documentation.

src/MemoryTracing/MemoryTracingUtils.cpp
vickyliu-go4it: Thanks for the suggestion. I moved the resetting of system_memory_usage fields to the corresponding Get method.

src/GrpcProtos/capture.proto
karupayun: Yes, of course. Agree 100%

src/GrpcProtos/capture.proto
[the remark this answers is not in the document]
karupayun: mmm, ok. So strange this name. Maybe adding a comment before: `// minor page faults` ?

src/MemoryTracing/MemoryTracingUtilsTest.cpp
```diff
@@ -115,89 +147,224 @@ SwapCached:      0 kB)",
 
   const std::string kEmptyMeminfo = "";
 
-  SystemMemoryUsage parsing_result = ParseMemInfo(kValidMeminfo);
-  ExpectSystemMemoryUsageEq(parsing_result, kMemTotal, kMemFree, kMemAvailable, kBuffers, kCached);
+  SystemMemoryUsage system_memory_usage;
+  ResetSystemMemoryUsage(&system_memory_usage);
```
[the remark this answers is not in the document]
vickyliu-go4it: Thanks for the suggestion. I changed the `ResetSystemMemoryUsage` method to the `CreateAndInitializeSystemMemoryUsage` method, and moved it to the header file `MemoryTracingUtilis.h`.

src/MemoryTracing/MemoryTracingUtilsTest.cpp
```diff
@@ -115,89 +147,224 @@ SwapCached:      0 kB)",
...
+  GetValuesFromMemInfo(kValidMeminfo, &system_memory_usage);
+  ExpectSystemMemoryUsageEq(system_memory_usage, kMemTotal, kMemFree, kMemAvailable, kBuffers,
+                            kCached);
 
-  parsing_result = ParseMemInfo(kPartialMeminfo);
-  ExpectSystemMemoryUsageEq(parsing_result, kMemTotal, kMemFree);
+  ResetSystemMemoryUsage(&system_memory_usage);
```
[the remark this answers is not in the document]
vickyliu-go4it: Thanks for the suggestion! I added the `CreateAndInitializeSystemMemoryUsage` method tp create the memory proto with initializing value of fields to `kMissingInfo`.

src/MemoryTracing/MemoryTracingUtilsTest.cpp
```diff
@@ -115,89 +147,224 @@ SwapCached:      0 kB)",
...
+swap_ra_hit 207052
+nr_unstable 0)",
+                                                        kPageFaults, kMajorPageFaults);
+
+  const std::string kPartialProcVmStat = absl::Substitute(R"(pgfault $0)", kPageFaults);
+
+  const std::string kEmptyProcVmStat = "";
+
+  SystemMemoryUsage system_memory_usage;
+  ResetSystemMemoryUsage(&system_memory_usage);
```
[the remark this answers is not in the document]
vickyliu-go4it: Thanks! Changed this as the `CreateAndInitializeSystemMemoryUsage` method.

src/MemoryTracing/MemoryTracingUtils.cpp
vickyliu-go4it: We initialize the values in the memory proto to  `kMissingInfo` when we create it. When calling the Get method to update corresponding fields in the memory proto, we only update the proto when the format of the file content is correct.

If the format of the file content is incorrect, we return without updating the corresponding fields in the memory proto. According to the source code, we can make sure that this file content contains 52 fields. 

Only in very wired case the format of this linux virtual file could be wrong. The '/proc/<pid>/stat/` file provides only values and there is no parameter name. If some fields are missing in the file, even though the order might be correct but we still cannot find the value we would like to extract.

src/MemoryTracing/MemoryTracingUtils.h
```diff
@@ -12,17 +12,24 @@
 
 namespace orbit_memory_tracing {
 
+[[nodiscard]] orbit_grpc_protos::SystemMemoryUsage CreateAndInitializeSystemMemoryUsage();
+// All the following GetValues methods only update the input memory proto when the input file
```
karupayun, line 16: I'm ok with this (as discussed online) but I would like to have an additional opinion from @ronaldfw

src/MemoryTracing/MemoryTracingUtilsTest.cpp
```diff
@@ -161,17 +132,16 @@ SwapCached:      0 kB)",
 
   const std::string kEmptyMeminfo = "";
 
-  SystemMemoryUsage system_memory_usage;
-  ResetSystemMemoryUsage(&system_memory_usage);
+  SystemMemoryUsage system_memory_usage = CreateAndInitializeSystemMemoryUsage();
```
karupayun, line 135: as I said before, I'd rather prefer to have these sub-tests divided by "{", "}"

```
{
  SystemMemoryUsage system_memory_usage = CreateAndInitializeSystemMemoryUsage();
  GetValuesFromMemInfo(kValidMeminfo, &amp;system_memory_usage);  
 ExpectSystemMemoryUsageEq(system_memory_usage, kMemTotal, kMemFree, kMemAvailable, kBuffers,                            kCached);
}


{
  SystemMemoryUsage system_memory_usage = CreateAndInitializeSystemMemoryUsage();  
  GetValuesFromMemInfo(kPartialMeminfo, &amp;system_memory_usage);  
  ExpectSystemMemoryUsageEq(system_memory_usage, kMemTotal, kMemFree);
} 

```
.. etc

src/MemoryTracing/MemoryTracingUtils.cpp
ronaldfw: I agree we should error out if the format is incorrect. Should we return an error from this function, though? Or at leaset log an error here so it's easier to see why this failed should it fail?

src/MemoryTracing/MemoryTracingUtils.cpp
```diff
@@ -85,4 +88,137 @@ std::optional<SystemMemoryUsage> GetSystemMemoryUsage() noexcept {
...
+  // Extract cgroup memory limit from /sys/fs/cgroup/memory/<cgroup_name>/memory.limit_in_bytes.
+  reading_result = orbit_base::ReadFileToString(
+      absl::StrFormat("/sys/fs/cgroup/memory/%s/memory.limit_in_bytes", cgroup_name));
+  if (!reading_result.has_value()) {
+    ERROR("%s", reading_result.error().message());
+    return std::nullopt;
+  }
+  int64_t cgroup_memory_limit_in_bytes = GetCGroupMemoryLimitInBytes(reading_result.value());
+
+  // Extract cgroup memory usage from the /sys/fs/cgroup/memory/<cgroup_name>/memory.stat file.
```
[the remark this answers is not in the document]
ronaldfw: I agree it's a nice idea to use constants to clearly indicate what string literals represent.

src/MemoryTracing/MemoryTracingUtils.h
ronaldfw: I think it's fine to not update the proto behind the pointer in case there is a failure in the method. I would consider returning an error message in case there is a problem. Or is there a reason you are not returning an error?

src/MemoryTracing/MemoryTracingUtils.h
ronaldfw: I think it's fine for consistency with the rest. In Google code style, parameters passed as * are in/out parameters, so it's pretty obvious that the value is returned as part of the proto that is passed in via pointer. 

You _could_ consider a different name (not Get), maybe something like `InitializeCGroupMemoryLimitInBytes`, but I don't feel too strongly about this (and I'm not sure what the best name here is).

src/MemoryTracing/MemoryTracingUtils.h
vickyliu-go4it: Thanks for the suggestions. I changed to return an `ErrorMessage` in case there is a problem in extracting information from the file content.

src/MemoryTracing/MemoryTracingUtilsTest.cpp
vickyliu-go4it: Thanks for the suggestion! Changed accordingly.

src/MemoryTracing/MemoryTracingUtils.cpp
vickyliu-go4it: I changed to return an error message which explains the reason of failing to extract information from the file content, for instance, empty file content, wrong format in which line, or fail to extract the value. Thanks for the suggestion.

src/MemoryTracing/MemoryTracingUtils.h
vickyliu-go4it: Thanks! I renamed these GetValues methods with the format Update<proto name>From<file name>, for instance, `UpdateSystemMemoryUsageFromMemInfo`, `UpdateCgroupMemoryUsageFromMemoryStat`.

src/MemoryTracing/MemoryTracingUtils.cpp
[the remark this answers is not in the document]
vickyliu-go4it: Thanks for the suggestion! Changed to use constants and removed some redundant comments.

src/MemoryTracing/MemoryTracingUtils.cpp
ronaldfw: Sounds great, thanks!

src/MemoryTracing/MemoryTracingUtils.h
ronaldfw: I really like the new names, looks great!

ronaldfw (approved):
LGTM, thanks!

PR #2259, opened 2021-04-29
Fix the crash when loading capture files with memory tracks

vickyliu-go4it:
With this change, we fix the crash when loading capture files with memory tracks.

If the user deselect "Collect system-wide memory usage information" in the capture option dialog and then load a capture file with memory tracks, orbit crashes because of a redundant check in `TimeGraph::ProcessMemoryTrackingTimer`.

Bug: http://b/186738639
Test: 1) deselect "Collect system-wide memory usage information" in the capture option dialog, and then 2) load a capture file with memory tracks.


antonrohr (approved):
LGTM! The explanation sounds reasonable to me. I did not check the code out, I assume you tested this. Let me know if you want me to test this.

reichlfl (approved):

PR #2260, opened 2021-04-29
Refactor OrbitApp, add Future as some return type

antonrohr:
This is needed for the capture metrics that is coming soon


beckerhe (approved):
LGTM

PR #2261, opened 2021-04-29
Add capture metric

antonrohr:
Introduces CaptureMetric, which is sent after a capture ends. 
orbit_metrics_uploader::CaptureStartData and orbit_metrics_uploader::CaptureCompleteData will be extended to hold more values in the future. 

This can be reviewed now, since orbit_log_event.proto is highly unlikely to still change. Anyways, I will only merge this once  http://cl/371061415 is submitted

### DIFF
--- a/src/MetricsUploader/CaptureMetricTest.cpp
+++ b/src/MetricsUploader/CaptureMetricTest.cpp
@@ -1,0 +1,163 @@
+// Copyright (c) 2021 The Orbit Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#include <gmock/gmock.h>
+#include <gtest/gtest.h>
+
+#include <chrono>
+#include <thread>
+
+#include "MetricsUploader/CaptureMetric.h"
+#include "MetricsUploader/MetricsUploader.h"
+#include "orbit_log_event.pb.h"
+
+namespace orbit_metrics_uploader {
+
+namespace {
+
+constexpr CaptureStartData kTestStartData{
+    1 /*number_of_instrumented_functions*/,
+    2 /*number_of_frame_tracks*/,
+    3 /*number_of_manual_start_timers*/,
+    4 /*number_of_manual_stop_timers*/,
+    5 /*number_of_manual_start_async_timers*/,
+    6 /*number_of_manual_stop_async_timers*/,
+    7 /*number_of_manual_tracked_values*/
+};
+
+bool HasSameCaptureStartData(const OrbitCaptureData& capture_data,
+                             const CaptureStartData& start_data) {
+  return capture_data.number_of_instrumented_functions() ==
+             start_data.number_of_instrumented_functions &&
+         capture_data.number_of_frame_tracks() == start_data.number_of_frame_tracks &&
+         capture_data.number_of_manual_start_timers() == start_data.number_of_manual_start_timers &&
+         capture_data.number_of_manual_stop_timers() == start_data.number_of_manual_stop_timers &&
+         capture_data.number_of_manual_start_async_timers() ==
+             start_data.number_of_manual_start_async_timers &&
+         capture_data.number_of_manual_stop_async_timers() ==
+             start_data.number_of_manual_stop_async_timers &&
+         capture_data.number_of_manual_tracked_values() ==
+             start_data.number_of_manual_tracked_values;
+}
+
+}  // namespace
+
+using ::testing::_;
+using ::testing::Ge;
+
+class MockUploader : public MetricsUploader {
+ public:
+  MOCK_METHOD(bool, SendLogEvent, (OrbitLogEvent_LogEventType /*log_event_type*/), (override));
+  MOCK_METHOD(bool, SendLogEvent,
+              (OrbitLogEvent_LogEventType /*log_event_type*/,
+               std::chrono::milliseconds /*event_duration*/),
+              (override));
+  MOCK_METHOD(bool, SendLogEvent,
+              (OrbitLogEvent_LogEventType /*log_event_type*/,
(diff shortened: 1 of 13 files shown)


src/MetricsUploader/CaptureMetricTest.cpp
```diff
@@ -0,0 +1,159 @@
...
+    3 /*number_of_manual_start_timers*/,
+    4 /*number_of_manual_stop_timers*/,
+    5 /*number_of_manual_start_async_timers*/,
+    6 /*number_of_manual_stop_async_timers*/,
+    7 /*number_of_manual_tracked_values*/
+};
+
+bool CaptureDataProtoContainsCaptureStartData(const OrbitCaptureData& capture_data,
+                                              const CaptureStartData& start_data) {
+  return capture_data.number_of_instrumented_functions() ==
```
karupayun, line 31: I'm not sure if I understand. Why we are making this comparisons here?

src/MetricsUploader/CaptureMetricTest.cpp
```diff
@@ -0,0 +1,159 @@
...
+        EXPECT_TRUE(CaptureDataProtoContainsCaptureStartData(capture_data, kTestStartData));
+        return true;
+      });
+
+  CaptureMetric metric{&uploader, kTestStartData};
+  metric.SetCaptureComplete(complete_data);
+  EXPECT_TRUE(metric.Send());
+}
+
+TEST(CaptureMetric, MultipleSetAndSend) {
```
karupayun, line 133: This could be a real situation? Why we are testing this?

src/OrbitGl/App.cpp
```diff
@@ -151,6 +154,41 @@ PresetLoadState GetPresetLoadStateForProcess(
...
   return PresetLoadState::kPartiallyLoadable;
 }
+
+orbit_metrics_uploader::CaptureStartData CreateCaptureStartData(
+    const std::vector<FunctionInfo>& all_instrumented_functions, int64_t number_of_frame_tracks) {
+  orbit_metrics_uploader::CaptureStartData capture_start_data{};
+
+  for (const auto& function : all_instrumented_functions) {
+    switch (function.orbit_type()) {
+      case orbit_client_protos::FunctionInfo_OrbitType_kNone:
```
karupayun, line 164: All of them are instrumented functions?

karupayun (approved):
LGTM! I made a few questions about parts I didn't understand. Will wait Henning approval as well.

src/OrbitGl/App.cpp
antonrohr: The proto field `number_of_instrumented_functions` has the following comment:
```
  // Number of functions the user dynamically instrumented (hooked) before the
  // start of this capture. This does not include functions that are
  // instrumented via the manual instrumentation api (Orbit.h).
```

But I see how this still might be confusing. Do you think maybe the name `number_of_dynamically_instrumented_functions` is better? Or any other suggestions. 

There is also going to be a different field soon:
```
  // Total number of timers of all dynamically instrumented (hooked) functions
  // that were recorded in this capture. This does not include anything from
  // the manual instrumentation api (Orbit.h).
  int64 number_of_instrumented_function_timers = 10;
```
I could also call this `number_of_dynamically_instrumented_function_timers`

src/MetricsUploader/CaptureMetricTest.cpp
antonrohr: No, currently this is not used. However, the class `CaptureMetric` might be used by someone else in the future (maybe ClientGgp). So I designed the class in a way that the outcome of these setters is always the same. The setters basically set the state of the class. Is is confusing for users of the class, if a call to a setter has a different outcome depending on the state of the class before the call. So I wrote this test to make sure the class adheres to this. I hope this explanation makes a little bit of sense. 

The idea actually came from @beckerhe (feel free to chime in).

karupayun: mmm, in this case I will rename the test to CheckingSameOutcomeSetAndSend and maybe 3 similar CaptureMetrics metric1, metric2 and metric3?

src/OrbitGl/App.cpp
karupayun: Yes, I will rename it in every place you like.

> I could also call this number_of_dynamically_instrumented_function_timers

And yes, to not confuse with the manual ones, thanks :)

src/MetricsUploader/CaptureMetricTest.cpp
antonrohr: The function `CaptureDataProtoContainsCaptureStartData` is just a helper function to compare the input of a test to the output of that test. This is to have less duplicate code. The alternative is to do an EXPECT_EQ() for each of these.

karupayun: Ah, ok. It makes sense now. I would like to rename it, for me "Contains" it's about if we have this data more than that the data has the same value.

I would rename it to something like "HasSameCaptureStartData"

beckerhe (approved):
LGTM. Thank you Anton. The changes with the Setters and the separate Send function look good!

src/OrbitGl/App.cpp
antonrohr: As discussed offline, sadly not possible anymore, since this is only a copy from the source of truth

src/MetricsUploader/CaptureMetricTest.cpp
antonrohr: I improved the test

src/MetricsUploader/CaptureMetricTest.cpp
antonrohr: done :)

PR #2263, opened 2021-04-30
Create AccessibleInterfaceProvider as a base class for Accessibility

karupayun:
In the process of erasing GlCanvas dependence, we are creating a common interface, so TimeGraph's doesn't have to know which type is its parent.

As a result GetCanvas() from TimeGraph is not used anymore.

https://b/185214650.

### DIFF
--- a/src/OrbitGl/AccessibleInterfaceProvider.cpp
+++ b/src/OrbitGl/AccessibleInterfaceProvider.cpp
@@ -1,0 +1,17 @@
+// Copyright (c) 2021 The Orbit Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#include "AccessibleInterfaceProvider.h"
+
+namespace orbit_gl {
+
+orbit_accessibility::AccessibleInterface*
+AccessibleInterfaceProvider::GetOrCreateAccessibleInterface() {
+  if (accessibility_ == nullptr) {
+    accessibility_ = CreateAccessibleInterface();
+  }
+  return accessibility_.get();
+}
+
+}  // namespace orbit_gl
--- a/src/OrbitGl/AccessibleInterfaceProvider.h
+++ b/src/OrbitGl/AccessibleInterfaceProvider.h
@@ -1,0 +1,30 @@
+// Copyright (c) 2021 The Orbit Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#ifndef ORBIT_GL_ACCESSIBLE_INTERFACE_PROVIDER_H_
+#define ORBIT_GL_ACCESSIBLE_INTERFACE_PROVIDER_H_
+
+#include "OrbitAccessibility/AccessibleInterface.h"
+
+namespace orbit_gl {
+
+/* Base class to provide accessibility in the capture window. */
+class AccessibleInterfaceProvider {
+ public:
+  explicit AccessibleInterfaceProvider() {}
+
+  [[nodiscard]] orbit_accessibility::AccessibleInterface* GetOrCreateAccessibleInterface();
+  [[nodiscard]] const orbit_accessibility::AccessibleInterface* GetAccessibleInterface() const {
+    return accessibility_.get();
+  }
+
+ private:
+  [[nodiscard]] virtual std::unique_ptr<orbit_accessibility::AccessibleInterface>
+  CreateAccessibleInterface() = 0;
+  std::unique_ptr<orbit_accessibility::AccessibleInterface> accessibility_;
+};
+
+}  // namespace orbit_gl
+
+#endif
--- a/src/OrbitGl/AccessibleTimeGraph.cpp
+++ b/src/OrbitGl/AccessibleTimeGraph.cpp
@@ -6,8 +6,8 @@
 
 #include <vector>
 
+#include "AccessibleInterfaceProvider.h"
 #include "AccessibleTrack.h"
-#include "GlCanvas.h"
 #include "TimeGraph.h"
 #include "Track.h"
 #include "TrackManager.h"
@@ -40,7 +40,8 @@ const AccessibleInterface* TimeGraphAccessibility::AccessibleChild(int index) co
 }
 
 const AccessibleInterface* TimeGraphAccessibility::AccessibleParent() const {
-  return time_graph_->GetCanvas()->GetOrCreateAccessibleInterface();
+  // Special handling given than GlCanvas is not a CaptureViewElement.
+  return time_graph_->GetAccessibleParent()->GetOrCreateAccessibleInterface();
 }
 
 }  // namespace orbit_gl
--- a/src/OrbitGl/CaptureViewElement.h
+++ b/src/OrbitGl/CaptureViewElement.h
@@ -5,6 +5,7 @@
 #ifndef ORBIT_GL_CAPTURE_VIEW_ELEMENT_H_
 #define ORBIT_GL_CAPTURE_VIEW_ELEMENT_H_
 
+#include "AccessibleInterfaceProvider.h"
 #include "Batcher.h"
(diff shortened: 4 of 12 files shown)


src/OrbitGl/AccessibleInterfaceProvider.h
```diff
@@ -0,0 +1,33 @@
...
+class AccessibleInterfaceProvider {
+ public:
+  explicit AccessibleInterfaceProvider(AccessibleInterfaceProvider* parent) : parent_(parent) {}
+
+  [[nodiscard]] orbit_accessibility::AccessibleInterface* GetOrCreateAccessibleInterface();
+  [[nodiscard]] const orbit_accessibility::AccessibleInterface* GetAccessibleInterface() const {
+    return accessibility_.get();
+  }
+
+  [[nodiscard]] virtual AccessibleInterfaceProvider* GetParent() const { return parent_; }
```
florian-kuebler, line 22: This doesn't need to be virtual anymore.

src/OrbitGl/AccessibleTimeGraph.cpp
```diff
@@ -7,7 +7,6 @@
 #include <vector>
 
 #include "AccessibleTrack.h"
-#include "GlCanvas.h"
```
florian-kuebler, line 10: Nit: Include the AccessibleInterfaceProvider.h.

florian-kuebler (approved):
Besides the two nit comments, that LGTM. Thanks!

src/OrbitGl/CaptureViewElement.h
```diff
@@ -17,9 +18,9 @@ class GlCanvas;
 namespace orbit_gl {
 
 /* Base class for UI elements drawn underneath the capture window. */
-class CaptureViewElement : public Pickable {
+class CaptureViewElement : public Pickable, public AccessibleInterfaceProvider {
  public:
-  explicit CaptureViewElement(CaptureViewElement* parent, TimeGraph* time_graph,
+  explicit CaptureViewElement(AccessibleInterfaceProvider* parent, TimeGraph* time_graph,
```
reichlfl, line 23: As discussed offline: I would prefer to have the parent / children of our Composite pattern to be of the same type, so children can have access to the parent's position, size etc.

So I would leave this as `CaptureViewElement*`, and keep the special case in `TimeGraph` - but replace `GetCanvas()` with a new method that returns an `AccessibleInterfaceProvider`.

src/OrbitGl/AccessibleInterfaceProvider.cpp
```diff
@@ -0,0 +1,17 @@
...
+#include "AccessibleInterfaceProvider.h"
+
+namespace orbit_gl {
+
+orbit_accessibility::AccessibleInterface*
+AccessibleInterfaceProvider::GetOrCreateAccessibleInterface() {
+  if (accessibility_ == nullptr) {
+    accessibility_ = CreateAccessibleInterface();
+  }
+  return accessibility_.get();
```
vickyliu-go4it, line 14: (Personal preference 😃 ) I prefer to use `return GetAccessibleInterface();` here for the readability.

vickyliu-go4it (approved):
LGTM. Nice PR! Thanks Pablo!

src/OrbitGl/CaptureViewElement.h
florian-kuebler: Hmm, but maybe then we should simply rename `AccessibleInterfaceProvider` to something more general that has a position and size. GlCanvas being a sort of `CaptureViewElement` would be the cleanest solution, but `CaptureViewElement` already contains more than just pos, size, parent and accessibility. Essentially, GlCanvas is the parent of TimeGraph, so GetParent should be used.

reichlfl: IMO we shouldn't have this driven too much by the accessibility. The way it's build right now, `TimeGraph` for me is the clear root of the hierarchy of rendered elements, not `GlCanvas`. Having TimeGraph return `nullptr` as parent is logical in that sense.

I've always thought of it rather this way: How would we do it if accessibility wasn't involved? That should be the clean way. The accessibility needs to bridge that gap from GlCanvas somehow, and it's already a bit of a complex construct because we jump from the Qt world to Gl, and I would like to have this special cases hidden in the accessibility classes instead of the control hierarchy.

I already said this offline: I do not have a strong veto against this solution, and everything except the change in type for the parent makes sense to me, but I feel that the type change will cause issues in the future and does not reflect the meaning of "parent" because it is too prominently driven by accessibility.

florian-kuebler: To keep it short: I also don't have a strong opinion in this, so I'll let Pablo decide.

I think you could argue that GlCanvas could be considered being a parent of TimeGraph even completely without accessibility (where I totally agree with you that accessibility impl. details should not drive our design), esp. with respect to pos/size.

karupayun: I feel that ideally CaptureWindow might be the root of the OrbitGl tree. It depends on how we want to divide different TimeGraphs. But for sure, the type of all the elements in this case (CaptureWindows included) should be CaptureViewElement. So, for now I will keep this (CaptureViewElement as type for parents, nullptr as the father of TimeGraph) by having a different type for TimeGraph's parent, until we reach an state of several TimeGraph, when we should take the decision about which should be the root.

src/OrbitGl/AccessibleInterfaceProvider.cpp
karupayun: Sure!

src/OrbitGl/AccessibleInterfaceProvider.h
reichlfl, line 22: You don't really need this here anymore, do you? It's overridden in `CaptureViewElement` with a different return type, but since `TimeGraph` has its own accessibility class and special handling of this case, there's no need to declare it here.

src/OrbitGl/AccessibleInterfaceProvider.h
```diff
@@ -0,0 +1,33 @@
...
+
+ private:
+  AccessibleInterfaceProvider* parent_;
```
reichlfl, line 25: Same for this - it's not used, you even declare a variable of the same name in `CaptureViewElement` as type `CaptureViewElement*`, and an additional `AccessibleInterfaceProvider*` in `TimeGraph`. I'd just remove the declaration here, it shouldn't be needed. This will create chaos from multiple inheritance otherwise.

src/OrbitGl/AccessibleInterfaceProvider.h
karupayun: GlCanvas is using it. I prefer to have it here than in GlCanvas.

karupayun: You are right, thanks. I was sure that GlCanvas needed the GetParent() method. My bad.

src/OrbitGl/AccessibleInterfaceProvider.h
karupayun: Again, you are right, thanks.

src/OrbitGl/CaptureViewElement.h
```diff
@@ -47,12 +48,6 @@ class CaptureViewElement : public Pickable {
   void OnDrag(int x, int y) override;
   [[nodiscard]] bool Draggable() override { return true; }
 
-  // Accessibility
-  [[nodiscard]] orbit_accessibility::AccessibleInterface* GetOrCreateAccessibleInterface();
-  [[nodiscard]] const orbit_accessibility::AccessibleInterface* GetAccessibleInterface() const {
-    return accessibility_.get();
-  }
-
   [[nodiscard]] virtual CaptureViewElement* GetParent() const { return parent_; }
```
reichlfl, line 51: Nit: Not sure if this still needs to be virtual?

reichlfl (approved):
LGTM, thanks for the changes!

PR #2267, opened 2021-05-03
Remove bincrafters repo and update bintray repo url

beckerhe:
That's only the first step moving from the old bintray repo to JFrogs new repo offering for open source projects.

Due to missing bincrafters packages this won't work out of the box yet, but we're getting there.


pierricgimmig (approved):
Thanks for jumping on this quickly, @beckerhe! Lgtm

PR #2269, opened 2021-05-03
Implement reading from Main Section of Capture File

dimitry-:
Implement CaptureSectionInputStream for reading from the
main section of the capture file.

Test: run CaptureFileTests

### DIFF
--- a/src/CaptureFile/CaptureFile.cpp
+++ b/src/CaptureFile/CaptureFile.cpp
@@ -5,11 +5,14 @@
 #include "CaptureFile/CaptureFile.h"
 
 #include "CaptureFileConstants.h"
+#include "CaptureSectionInputStreamImpl.h"
 #include "OrbitBase/File.h"
 
 namespace orbit_capture_file {
 
 namespace {
+
+using orbit_base::unique_fd;
 
 struct CaptureFileHeader {
   std::array<char, kFileSignatureSize> signature;
@@ -37,15 +40,43 @@ class CaptureFileImpl : public CaptureFile {
   ErrorMessageOr<void> ReadFromSection(uint64_t section_number, uint64_t offset_in_section,
                                        void* data, size_t size) override;
 
+  std::unique_ptr<CaptureSectionInputStream> CreateCaptureSectionInputStream() override;
+
  private:
   // Parse header and validate version/file-format
-  ErrorMessageOr<void> ReadHeaderAndSectionList();
+  ErrorMessageOr<void> ReadHeader();
+  ErrorMessageOr<void> ReadSectionList();
+  ErrorMessageOr<void> CalculateCaptureSectionSize();
 
   std::filesystem::path file_path_;
-  orbit_base::unique_fd fd_;
+  unique_fd fd_;
   CaptureFileHeader header_{};
+
+  // This is used for boundary checks so that we do not end up
+  // reading from sections following capture section, this is not
+  // the exact size of the section but it is always >= the actual
+  // size of the section. The user must rely on CaptureFinished
+  // message to detect the last message for the capture section.
+  uint64_t capture_section_size_ = 0;
+
   std::vector<CaptureFileSection> section_list_;
 };
+
+template <uint64_t alignment>
+constexpr uint64_t AlignUp(uint64_t value) {
+  // alignment must be a power of 2
+  static_assert((alignment & (alignment - 1)) == 0);
+  return (value + (alignment - 1)) & ~(alignment - 1);
+}
+
+ErrorMessageOr<uint64_t> GetEndOfFileOffset(const unique_fd& fd) {
+  off_t end_of_file = lseek(fd.get(), 0, SEEK_END);
+  if (end_of_file == -1) {
+    return ErrorMessage{SafeStrerror(errno)};
+  }
+
+  return end_of_file;
+}
 
 ErrorMessageOr<void> CaptureFileImpl::Initialize() {
   auto fd_or_error = orbit_base::OpenExistingFileForReadWrite(file_path_);
@@ -55,23 +86,40 @@ ErrorMessageOr<void> CaptureFileImpl::Initialize() {
 
   fd_ = std::move(fd_or_error.value());
 
-  OUTCOME_TRY(ReadHeaderAndSectionList());
+  OUTCOME_TRY(ReadHeader());
+  OUTCOME_TRY(ReadSectionList());
+  OUTCOME_TRY(CalculateCaptureSectionSize());
+
   return outcome::success();
 }
 
-ErrorMessageOr<void> CaptureFileImpl::ReadHeaderAndSectionList() {
-  OUTCOME_TRY(header, orbit_base::ReadFullyAtOffset<CaptureFileHeader>(fd_, 0));
-  header_ = header;
-
-  if (std::memcmp(header_.signature.data(), kFileSignature, header_.signature.size()) != 0) {
-    return ErrorMessage{"Invalid file signature"};
+ErrorMessageOr<void> CaptureFileImpl::CalculateCaptureSectionSize() {
+  // If there are no additional sections the capture section ends at the EOF
+  if (header_.section_list_offset == 0) {
+    OUTCOME_TRY(end_of_file_offset, GetEndOfFileOffset(fd_));
+    capture_section_size_ = end_of_file_offset - header_.capture_section_offset;
+    return outcome::success();
   }
 
-  if (header_.version != kFileVersion) {
-    return ErrorMessage{
-        absl::StrFormat("Incompatible version %d, expected %d", header_.version, kFileVersion)};
+  // Otherwise it ends at the start of the next section
+  uint64_t min_section_offset = std::numeric_limits<uint64_t>::max();
+
+  CHECK(!section_list_.empty());
+
+  for (const auto& section : section_list_) {
+    if (section.offset < min_section_offset) {
+      min_section_offset = section.offset;
+    }
   }
 
+  CHECK(min_section_offset < std::numeric_limits<uint64_t>::max());
+
+  capture_section_size_ = min_section_offset - header_.capture_section_offset;
+
+  return outcome::success();
+}
+
+ErrorMessageOr<void> CaptureFileImpl::ReadSectionList() {
   if (header_.section_list_offset == 0) {
     return outcome::success();
   }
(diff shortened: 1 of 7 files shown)


src/CaptureFile/include/CaptureFile/CaptrureSectionInputStream.h
```diff
@@ -0,0 +1,23 @@
...
+
+#ifndef CAPTURE_FILE_MAIN_SECTION_INPUT_STREAM_H_
+#define CAPTURE_FILE_MAIN_SECTION_INPUT_STREAM_H_
+
+#include "OrbitBase/Result.h"
+#include "capture.pb.h"
+
+namespace orbit_capture_file {
+
+class CaptureSectionInputStream {
```
florian-kuebler, line 13: Could you add a super brief comment for this class?

src/CaptureFile/CaptureSectionInputStreamImpl.h
```diff
@@ -0,0 +1,31 @@
...
+#ifndef CAPTURE_SECTION_INPUT_STREAM_IMPL_H_
+#define CAPTURE_SECTION_INPUT_STREAM_IMPL_H_
+
+#include "CaptureFile/CaptrureSectionInputStream.h"
+#include "FileFragmentInputStream.h"
+#include "OrbitBase/File.h"
+
+namespace orbit_capture_file_internal {
+
+class CaptureSectionInputStreamImpl : public orbit_capture_file::CaptureSectionInputStream {
```
florian-kuebler, line 14: Could you add a brief comment to the class?

src/CaptureFile/CaptureSectionInputStreamImpl.cpp
```diff
@@ -0,0 +1,31 @@
...
+#include "OrbitBase/MakeUniqueForOverwrite.h"
+
+namespace orbit_capture_file_internal {
+using orbit_grpc_protos::ClientCaptureEvent;
+
+ErrorMessageOr<ClientCaptureEvent> CaptureSectionInputStreamImpl::ReadEvent() {
+  uint64_t message_size = 0;
+
+  if (!coded_input_stream_.ReadVarint64(&message_size)) {
+    return file_fragment_input_stream_.GetLastError().value_or(
```
florian-kuebler, line 16: It feels to me that I a missing a small bit. Why is `file_fragment_input_stream_` only used for reading the last error?
How does it correlate with the `coded_input_stream_`?
Could you add a comment somewhere to describe the relationship (could be in the class level comment)?

florian-kuebler: Edit: Okay after reading the header again, I see that coded_input_stream_ is backed by the file_fragment_input_stream_. Still think that a brief comment could help.

src/CaptureFile/CaptureFileTest.cpp
```diff
@@ -20,15 +20,127 @@ static constexpr const char* kNotAnAnswerString = "Some odd number, not the answ
...
+
+  auto output_stream_or_error = CaptureFileOutputStream::Create(temp_file_name);
+  ASSERT_TRUE(output_stream_or_error.has_value()) << output_stream_or_error.error().message();
+
+  std::unique_ptr<CaptureFileOutputStream> output_stream =
+      std::move(output_stream_or_error.value());
+
+  EXPECT_TRUE(output_stream->IsOpen());
+
+  orbit_grpc_protos::ClientCaptureEvent event1 =
```
florian-kuebler, line 99: I guess you could extract the creation of the capture file into a helper method, as you already have the same code above.

src/CaptureFile/CaptureSectionInputStreamImpl.cpp
dimitry-: done.

src/CaptureFile/CaptureSectionInputStreamImpl.h
dimitry-: done

src/CaptureFile/include/CaptureFile/CaptrureSectionInputStream.h
dimitry-: done.

src/CaptureFile/CaptureSectionInputStreamImpl.h
```diff
@@ -0,0 +1,32 @@
+// Copyright (c) 2021 The Orbit Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#ifndef CAPTURE_SECTION_INPUT_STREAM_IMPL_H_
+#define CAPTURE_SECTION_INPUT_STREAM_IMPL_H_
+
+#include "CaptureFile/CaptrureSectionInputStream.h"
```
antonrohr, line 8:
```suggestion
#include "CaptureFile/CaptureSectionInputStream.h"
```

src/CaptureFile/CaptureFileTest.cpp
florian-kuebler: As discussed offline, lets keep it as is.

florian-kuebler (approved):
LGTM, thanks!

src/CaptureFile/include/CaptureFile/CaptrureSectionInputStream.h
```diff
@@ -0,0 +1,25 @@
+// Copyright (c) 2021 The Orbit Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#ifndef CAPTURE_FILE_MAIN_SECTION_INPUT_STREAM_H_
```
antonrohr, line 5:
```suggestion
#ifndef CAPTURE_SECTION_INPUT_STREAM_H_
```

dpallotti, line 5: Adjust to `CAPTURE_FILE_CAPTURE_SECTION_INPUT_STREAM_H_`.

src/CaptureFile/include/CaptureFile/CaptrureSectionInputStream.h
```diff
@@ -0,0 +1,25 @@
+// Copyright (c) 2021 The Orbit Authors. All rights reserved.
```
dpallotti, line 1: The filename has a typo (captrure -> capture).

src/CaptureFile/CaptureFileTest.cpp
dpallotti: Why does using `CHECK` hurt readability of failures? Because you can't use `<<`? Why not use `FAIL_IF`? We can add `FAIL_UNLESS` if you don't want to invert the condition. There's a bit of duplicated code here.

src/CaptureFile/CaptureFile.cpp
```diff
@@ -55,23 +86,39 @@ ErrorMessageOr<void> CaptureFileImpl::Initialize() {
...
 
-ErrorMessageOr<void> CaptureFileImpl::ReadHeaderAndSectionList() {
-  OUTCOME_TRY(header, orbit_base::ReadFullyAtOffset<CaptureFileHeader>(fd_, 0));
-  header_ = header;
-
-  if (std::memcmp(header_.signature.data(), kFileSignature, header_.signature.size()) != 0) {
-    return ErrorMessage{"Invalid file signature"};
+ErrorMessageOr<void> CaptureFileImpl::CalculateCaptureSectionSize() {
+  // Calculate capture section size
+  if (header_.section_list_offset == 0) {
```
dpallotti, line 98: This means that there are no additional sections, right?

src/CaptureFile/CaptureFile.cpp
```diff
@@ -188,6 +240,10 @@ ErrorMessageOr<void> orbit_capture_file::CaptureFileImpl::ReadFromSection(
 
   return outcome::success();
 }
+std::unique_ptr<CaptureSectionInputStream> CaptureFileImpl::CreateCaptureSectionInputStream() {
```
dpallotti, line 243:
```suggestion
}

std::unique_ptr<CaptureSectionInputStream> CaptureFileImpl::CreateCaptureSectionInputStream() {
```

src/CaptureFile/CaptureFileTest.cpp
```diff
@@ -20,15 +20,127 @@ static constexpr const char* kNotAnAnswerString = "Some odd number, not the answ
...
+    EXPECT_EQ(event_or_error.value().interned_string().intern(), kAnswerString);
+  }
+
+  {
+    auto event_or_error = capture_section->ReadEvent();
+    ASSERT_TRUE(event_or_error.has_value()) << event_or_error.error().message();
+    ASSERT_EQ(event_or_error.value().event_case(), ClientCaptureEvent::kInternedString);
+    EXPECT_EQ(event_or_error.value().interned_string().key(), kNotAnAnswerKey);
+    EXPECT_EQ(event_or_error.value().interned_string().intern(), kNotAnAnswerString);
+  }
```
antonrohr, line 81: Maybe call `ReadEvent()` a third time. I would expect an error then

src/CaptureFile/CaptureSectionInputStreamImpl.cpp
```diff
@@ -0,0 +1,34 @@
...
+
+#include "OrbitBase/MakeUniqueForOverwrite.h"
+
+namespace orbit_capture_file_internal {
+using orbit_grpc_protos::ClientCaptureEvent;
+
+ErrorMessageOr<ClientCaptureEvent> CaptureSectionInputStreamImpl::ReadEvent() {
+  uint64_t message_size = 0;
+
+  // Note that in case there was an error CodedInputStream does not provide error messages/coded.
```
dpallotti, line 15:
```suggestion
  // Note that in case there was an error CodedInputStream does not provide error messages/codes.
```

dpallotti (approved):

antonrohr (approved):
lgtm!

src/CaptureFile/CaptureFileTest.cpp
dimitry-: I initially had this code but then I realized that:
1. File section can have a padding (filled with 0's) it used to align sections to 8bytes boundary.
2. Reading from stream filled with 0's yields empty (default) message - so this actually succeeds in the case where we have an additional section.
3. We now have CaptureFinished message to serve as terminator of a stream. This message is expected to be used in real world scenarios.

In conclusion - reading past CaptureFinished message is actually undefined even though the stream is bound to the capture section (sometimes it ends in reading default message other times it results in error, it depends if there is a padding in the main section).

And we should probably not test undefined behavior.

antonrohr: I agree. Thank you for the explanation! I would like this to be documented somewhere. A comment at `ReadEvent()` in the Impl.h file seems a good idea

src/CaptureFile/CaptureFile.cpp
dimitry-: yes. added a comment.

src/CaptureFile/CaptureFile.cpp
dimitry-: done

src/CaptureFile/CaptureFileTest.cpp
dimitry-: added a comment to ReadEvent method

src/CaptureFile/CaptureFileTest.cpp
dimitry-: yep, using FAIL_IF kind of solves it but it relies on the fact that our tests are isolated and if we use another environment which does not use isolation FAIL_IF aborts the executable, not just one test and that means other tests are not executed. It also feels awkward using FAIL_IF in place of ASSERT_*.

I think having a bit of duplication in tests is fine if doing otherwise results in worse readability of failure results.

src/CaptureFile/CaptureSectionInputStreamImpl.cpp
dimitry-: done.

src/CaptureFile/CaptureSectionInputStreamImpl.h
dimitry-: done.

src/CaptureFile/include/CaptureFile/CaptrureSectionInputStream.h
dimitry-: done.

src/CaptureFile/include/CaptureFile/CaptrureSectionInputStream.h
dimitry-: done - kept CAPTURE_FILE_ prefix.

dimitry-: done.

PR #2271, opened 2021-05-03
Fix TestInstrumentFunction Test

florian-kuebler:
With #2099 we combined the different Gpu tracks into one. Thus,
the "gfx" track itself does not have timers anymore, but instead
there is a subtrack which does.

This adjusts the test accordingly.

Test: Run affected test locally.
Bug: http://b/187013635


danielfenner (approved):
lgtm! Thanks for fixing!

florian-kuebler:
Merging this now to fix the e2e tests. @reichlfl please let me know if there is anything regarding this.

PR #2272, opened 2021-05-03
Implement CaptureSaveToFileProcessor

dimitry-:
This implementation of CaptureEventProcessor saves
events to main section of the capture file. It also
saves frame_tracks to USER_DATA section of the file.

Test: run CaptureClientTests

### DIFF
--- a/src/CaptureClient/CMakeLists.txt
+++ b/src/CaptureClient/CMakeLists.txt
@@ -24,9 +24,11 @@ target_sources(CaptureClient PRIVATE
         ApiEventProcessor.cpp
         CaptureClient.cpp
         CaptureEventProcessor.cpp
-        GpuQueueSubmissionProcessor.cpp)
+        GpuQueueSubmissionProcessor.cpp
+        SaveToFileCaptureEventProcessor.cpp)
 
 target_link_libraries(CaptureClient PUBLIC
+        CaptureFile
         GrpcProtos
         OrbitCore)
 
@@ -41,7 +43,8 @@ target_compile_options(CaptureClientTests PRIVATE ${STRICT_COMPILE_FLAGS})
 
 target_sources(CaptureClientTests PRIVATE
         ApiEventProcessorTest.cpp
-        CaptureEventProcessorTest.cpp)
+        CaptureEventProcessorTest.cpp
+        SaveToFileCaptureEventProcessorTest.cpp)
 
 target_link_libraries(
         CaptureClientTests PRIVATE
--- a/src/CaptureClient/SaveToFileCaptureEventProcessor.cpp
+++ b/src/CaptureClient/SaveToFileCaptureEventProcessor.cpp
@@ -1,0 +1,141 @@
+// Copyright (c) 2021 The Orbit Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#include "CaptureClient/CaptureEventProcessor.h"
+#include "CaptureFile/CaptureFile.h"
+#include "CaptureFile/CaptureFileOutputStream.h"
+#include "OrbitBase/MakeUniqueForOverwrite.h"
+#include "capture_data.pb.h"
+
+using orbit_capture_file::CaptureFile;
+using orbit_capture_file::CaptureFileOutputStream;
+using orbit_client_protos::UserDefinedCaptureInfo;
+using orbit_grpc_protos::ClientCaptureEvent;
+
+namespace orbit_capture_client {
+
+namespace {
+
+class SaveToFileCaptureEventProcessor : public CaptureEventProcessor {
+ public:
+  explicit SaveToFileCaptureEventProcessor(std::filesystem::path file_path,
+                                           absl::flat_hash_set<uint64_t> frame_track_function_ids,
+                                           std::function<void(const ErrorMessage&)> error_handler)
+      : frame_track_function_ids_(std::move(frame_track_function_ids)),
+        file_path_{std::move(file_path)},
+        error_handler_{std::move(error_handler)},
+        state_{State::kProcessing} {}
+  ~SaveToFileCaptureEventProcessor() override = default;
+
+  ErrorMessageOr<void> Initialize();
+  void ProcessEvent(const ClientCaptureEvent& event) override;
+
+ private:
+  enum class State {
+    kProcessing,
+    kCaptureFinished,
+    kErrorReported,
+  };
+
+  ErrorMessageOr<void> WriteUserData();
+  void ReportError(const ErrorMessage& error);
+
+  absl::flat_hash_set<uint64_t> frame_track_function_ids_;
+  std::filesystem::path file_path_;
+  std::function<void(const ErrorMessage&)> error_handler_;
+  std::unique_ptr<CaptureFileOutputStream> output_stream_;
+  State state_;
+};
+
+ErrorMessageOr<void> SaveToFileCaptureEventProcessor::Initialize() {
+  auto stream_or_error = CaptureFileOutputStream::Create(file_path_);
+  if (stream_or_error.has_error()) {
+    return ErrorMessage{absl::StrFormat("Failed to initialize CaptureSaveToFileProcessor: %s",
+                                        stream_or_error.error().message())};
+  }
+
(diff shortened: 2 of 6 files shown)


src/CaptureClient/CaptureSaveToFileProcessor.cpp
```diff
@@ -0,0 +1,117 @@
...
+using orbit_capture_file::CaptureFile;
+using orbit_capture_file::CaptureFileOutputStream;
+using orbit_client_protos::UserDefinedCaptureInfo;
+using orbit_grpc_protos::ClientCaptureEvent;
+
+namespace orbit_capture_client {
+
+namespace {
+
+class CaptureSaveToFileProcessor : public CaptureEventProcessor {
```
dpallotti, line 20: I don't really like this name, I think it's weird to have "SaveToFile" in the middle. I would go for something like `SaveToFileCaptureEventProcessor`.

src/CaptureClient/CaptureSaveToFileProcessor.cpp
```diff
@@ -0,0 +1,117 @@
...
+  for (uint64_t function_id : frame_track_function_ids_) {
+    user_defined_capture_data.mutable_frame_tracks_info()->add_frame_track_function_ids(
+        function_id);
+  }
+
+  size_t message_size = user_defined_capture_data.ByteSizeLong();
+  auto buf = make_unique_for_overwrite<uint8_t[]>(message_size);
+  // SerializeToArray returns false if the size of the buffer is too small or too big
+  // https://github.com/protocolbuffers/protobuf/blob/5e84a6169cf0f9716c9285c95c860bcb355dbdc1/src/google/protobuf/message_lite.cc#L484
+  CHECK(user_defined_capture_data.SerializeToArray(buf.get(), message_size));
```
dpallotti, line 92: Are we fine with temporarily serializing the full message in memory (I guess it's very small, isn't it?) and then writing to file?

src/CaptureClient/CaptureSaveToFileProcessor.cpp
```diff
@@ -0,0 +1,117 @@
...
+  }
+
+  output_stream_ = std::move(stream_or_error.value());
+
+  return outcome::success();
+}
+
+void CaptureSaveToFileProcessor::ProcessEvent(const ClientCaptureEvent& event) {
+  CHECK(output_stream_ != nullptr);
+  if (!output_stream_->IsOpen()) return;
```
dpallotti, line 56: Could this be reported to the `error_handler_`? For example in case an event is received after the `CaptureFinished` event?

dimitry-: The another case here is after I/O error happened. In this case reporting this to error_handler_ will result in spam. Note that error_handler_ is not resposible for cleaning up the file in case I/O happens it is done by the CaptureFileOutputStream. After first error ProcessEvent should ignore all other errors.

src/CaptureClient/CaptureSaveToFileProcessor.cpp
dimitry-: Yes, they are very small - we shouldn't have anything big here.

src/CaptureClient/CMakeLists.txt
```diff
@@ -41,6 +43,7 @@ target_compile_options(CaptureClientTests PRIVATE ${STRICT_COMPILE_FLAGS})
 
 target_sources(CaptureClientTests PRIVATE
         ApiEventProcessorTest.cpp
+        CaptureSaveToFileProcessorTest.cpp
```
dpallotti, line 46: Rename to `SaveToFileCaptureEventProcessorTest.cpp`.

src/CaptureClient/CaptureSaveToFileProcessorTest.cpp
```diff
@@ -0,0 +1,129 @@
...
+  return event;
+}
+
+static ClientCaptureEvent CreateCaptureFinishedEvent() {
+  ClientCaptureEvent event;
+  event.mutable_capture_finished()->set_status(CaptureFinished::kSuccessful);
+  return event;
+}
+
+TEST(CaptureSaveToFileProcessor, SaveAndLoadSimpleCaptureWithFrameTracks) {
```
dpallotti, line 40: Rename the tests from `CaptureSaveToFileProcessor` to the new class name.

dpallotti (approved):
Some renaming left to do, looks good otherwise.

src/CaptureClient/CaptureSaveToFileProcessor.cpp
dpallotti: Yeah, I guess also having multiple events after the `CaptureFinished` event would result in spam. But in both cases you could report the error only if it's the first time.

src/CaptureClient/include/CaptureClient/CaptureEventProcessor.h
```diff
@@ -32,6 +32,10 @@ class CaptureEventProcessor {
 
   static std::unique_ptr<CaptureEventProcessor> CreateForCaptureListener(
       CaptureListener* capture_listener, absl::flat_hash_set<uint64_t> frame_track_function_ids);
+  static ErrorMessageOr<std::unique_ptr<CaptureEventProcessor>> CreateSaveToFileProcessor(
```
florian-kuebler, line 35: Just to clarify, long term this one will replace the other one, right?

src/CaptureClient/SaveToFileCaptureEventProcessor.cpp
```diff
@@ -0,0 +1,117 @@
...
+namespace {
+
+class SaveToFileCaptureEventProcessor : public CaptureEventProcessor {
+ public:
+  explicit SaveToFileCaptureEventProcessor(std::filesystem::path file_path,
+                                      absl::flat_hash_set<uint64_t> frame_track_function_ids,
+                                      std::function<void(const ErrorMessage&)> error_handler)
+      : frame_track_function_ids_(std::move(frame_track_function_ids)),
+        file_path_{std::move(file_path)},
+        error_handler_{std::move(error_handler)} {}
```
florian-kuebler, line 27: Should Initialize be called in the constructor?

src/CaptureClient/SaveToFileCaptureEventProcessor.cpp
```diff
@@ -0,0 +1,117 @@
...
+  if (!output_stream_->IsOpen()) return;
+
+  auto write_result = output_stream_->WriteCaptureEvent(event);
+  if (write_result.has_error()) {
+    error_handler_(write_result.error());
+    return;
+  }
+
+  if (event.event_case() == ClientCaptureEvent::kCaptureFinished) {
+    // we are done - close the stream
```
florian-kuebler, line 65:
```suggestion
    // We are done - close the stream.
```

src/CaptureClient/CMakeLists.txt
dimitry-: done.

src/CaptureClient/CaptureSaveToFileProcessorTest.cpp
dimitry-: done.

src/CaptureClient/SaveToFileCaptureEventProcessor.cpp
dimitry-: Constructors can't return ErrorMessageOr<>, this is one of the reasons we have a factory method here.

src/CaptureClient/SaveToFileCaptureEventProcessor.cpp
dimitry-: done.

src/CaptureClient/include/CaptureClient/CaptureEventProcessor.h
dimitry-: Not in the foreseeable future, for now we want the UI client to do both save to file and process. There is going to be a combination/chain of CaptureEventProcessors

src/CaptureClient/CaptureSaveToFileProcessor.cpp
dimitry-: ok.. I introduced additional state to catch "an event after CaptureFinished" error.

PR #2273, opened 2021-05-03
Fix public build

beckerhe:
This fixes the public build which cannot rely on prebuilt packages. Tested on my Linux machine.

### DIFF
--- a/conanfile.py
+++ b/conanfile.py
@@ -103,8 +103,7 @@ class OrbitConan(ConanFile):
         self.requires("zlib/1.2.11#9e0c292b60ce77402bd9be60dd68266f")
 
         if self.options.with_gui and self.options.with_crash_handling:
-            self.requires(
-                "crashpad/20200624@{}#8c19cb575eb819de0b050cf7d1f317b6".format(self._orbit_channel))
+            self.requires("crashpad/20200624@{}".format(self._orbit_channel))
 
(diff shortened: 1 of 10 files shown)


conanfile.py
```diff
@@ -103,8 +103,7 @@ def requirements(self):
         self.requires("zlib/1.2.11#9e0c292b60ce77402bd9be60dd68266f")
 
         if self.options.with_gui and self.options.with_crash_handling:
-            self.requires(
-                "crashpad/20200624@{}#8c19cb575eb819de0b050cf7d1f317b6".format(self._orbit_channel))
+            self.requires("crashpad/20200624@{}".format(self._orbit_channel))
```
pierricgimmig, line 106: in the lock file we have ""ref": "crashpad/20200624@orbitdeps/stable#5a03308d7b9a649eced1ee4d63c3d6b4", is there a potential mismatch?

pierricgimmig (approved):
Hi Henning, again, thanks for the quick turnover. LGTM, but please see small comment.

conanfile.py
beckerhe: No, this is intended. The hash changed due to the package update.

And removing it from the conanfile was just a clean-up. The recipe hashes are not needed in the conanfile anymore since we have the lockfiles.

PR #2274, opened 2021-05-04
Add GetTracerPidOfProcess to detect existing "Tracer" process

pierricgimmig:
GetTracerPidOfProcess returns the pid of the attached tracer process
for the given pid. If no process is attached, the function returns 0.
Internally, the function reads the content of "/proc/<pid>/status" to
find the "TracerPid" entry and returns its value.

This functionality is used in the second commit, which makes sure that no
tracer process is present before attempting to attach to the target process.
    
Tested on instance with and without debugger attached with expected results.

### DIFF
--- a/src/OrbitBase/GetProcessIdsLinux.cpp
+++ b/src/OrbitBase/GetProcessIdsLinux.cpp
@@ -4,10 +4,12 @@
 
 #include <filesystem>
 #include <optional>
+#include <string>
 #include <system_error>
 
 #include "OrbitBase/GetProcessIds.h"
 #include "OrbitBase/Logging.h"
+#include "OrbitBase/ReadFileToString.h"
 #include "OrbitBase/Result.h"
 #include "absl/strings/numbers.h"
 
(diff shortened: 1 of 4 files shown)


src/OrbitBase/GetProcessIdsLinux.cpp
```diff
@@ -86,4 +88,30 @@ std::vector<pid_t> GetTidsOfProcess(pid_t pid) {
...
+  std::optional<pid_t> tracer_pid;
+  constexpr const char* kTracerPidStr = "TracerPid:";
+
+  while (std::getline(status_file_content_ss, line)) {
+    if (line.find(kTracerPidStr) == std::string::npos) continue;
+    int potential_pid;
+    if (!absl::SimpleAtoi(line.substr(line.find_last_of(kTracerPidStr) + 1), &potential_pid)) {
+      return ErrorMessage(absl::StrFormat("Could not extract pid from line %s", line));
+    }
+    tracer_pid = potential_pid;
```
danielfenner, line 105: You could just `return potential_pid;` here. That would get rid of `tracer_pid` entirely and after the loop there would just be an `return ErrorMessage(...`.

It's a bit shorter but equivalent to what is already there - so feel free to ignore.

danielfenner (approved):
Thanks! lgtm!

PR #2275, opened 2021-05-04
Connect error handler when reusing a connection

antonrohr:
http://b/186741866

Reproduce before:
1. Connect to instance
2. Select whatever process
3. Start Session
4. End Session
5. ssh into instance, kill OrbitService
6. No Error message appears, the list of processes just stops updating

After this change: A error message is displayed and the instances are loaded to start another connection


beckerhe (approved):
Nice. Thanks!

PR #2276, opened 2021-05-04
Erase Canvas in the Track::Draw methods

karupayun:
In this PR, we are erasing canvas dependence from the Tracks.

For that, we are mainly replacing it in the Draw methods for a Batcher and a TextRenderer (both needed to draw) and also we are including the mouse_time (needed in the GraphTrack). There are also a few more modifications that I had to include in the same commit. Sorry for that.

With this we achieved 3 goals:
 - The dependences in Draw are easier to understand and to refactor in the future
 - canvas_ is not needed as a member in CaptureViewElements
 - Only 2 canvas references left in TimeGraph (one about the PickingManager which will be passed as a parameter in a future PR) and one to force a Redraw in GlCanvas (which should be managed in the opposite direction).

Focusing:
https://b/185214650.

Small TODO: Make the same in UpdatePrimitives and give a reference of the batcher instead of a pointer.

Test:
Run a capture and load a capture.

### DIFF
--- a/src/OrbitGl/AccessibleTrack.cpp
+++ b/src/OrbitGl/AccessibleTrack.cpp
@@ -11,7 +11,6 @@
 #include "AccessibleThreadBar.h"
 #include "AccessibleTimeGraph.h"
 #include "CoreMath.h"
-#include "GlCanvas.h"
 #include "OrbitBase/Logging.h"
 #include "TimeGraph.h"
 #include "Track.h"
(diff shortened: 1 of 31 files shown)


florian-kuebler (approved):
LGTM

beckerhe (approved):
LGTM. Thanks!

reichlfl (approved):

PR #2277, opened 2021-05-04
Add collect_gpu_jobs to CaptureClient and flag to OrbitFakeClient

dpallotti:
I want `OrbitFakeClient` to be able to disable collection of (almost) every
event.

Bug: http://b/187168022

Test:
- Take a capture -> GpuJobs still in order.
- Capture with `OrbitFakeClient` with all options off, and a capture with all
  options off except `-gpu_jobs`, notice the difference in number of events.

### DIFF
--- a/src/CaptureClient/CaptureClient.cpp
+++ b/src/CaptureClient/CaptureClient.cpp
@@ -63,15 +63,16 @@ InstrumentedFunction::FunctionType CaptureClient::InstrumentedFunctionTypeFromOr
   UNREACHABLE();
 }
 
-// TODO(vickyliu): This method contains a lot of arguments. Consider making it more structured.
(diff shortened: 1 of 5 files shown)


florian-kuebler (approved):
LGTM, but we need to make sure that we are not allowing captures with the Vulkan layer and without the collect_gpu_jobs (this is for a different change though).

dpallotti:
> LGTM, but we need to make sure that we are not allowing captures with the Vulkan layer and without the collect_gpu_jobs (this is for a different change though).

Yep, I thought about it. In the end this change is at a lower level, as the problem is matching the events from the two sources in the client, not generating or receiving the events themselves. Either way, until/unless we allow disabling `collect_gpu_jobs` in the UI (which doesn't seem necessary), this won't be a problem.

PR #2278, opened 2021-05-04
Re-add header of "Address" column in "Live" tab

dpallotti:
This was removed by 9f69e6783056895249194dbbdafdb7e470142347 but I don't know
if there's a reason (as the column is still there).

Bug: http://b/187178875

Test: Capture Trata with a function instrumented.


dpallotti:
@dimitry- Was there a reason or was it just an accident?

dpallotti:
Also should this go in the release?

dimitry-:
> @dimitry- Was there a reason or was it just an accident?

That was by accident... thanks for catching!

dimitry- (approved):

dimitry-:
we should probably cherry-pick it since this was a regression.

PR #2282, opened 2021-05-05
Fix deadlock on Capture finalizing

antonrohr:
http://b/187172673

This is a rather quick fix to not have a deadlock anymore. I filed b/187250643 to clean this up.

### DIFF
--- a/src/OrbitGl/App.cpp
+++ b/src/OrbitGl/App.cpp
@@ -1093,6 +1093,8 @@ void OrbitApp::StartCapture() {
       enable_introspection, max_local_marker_depth_per_command_buffer, collect_memory_info,
       memory_sampling_period_ns, std::move(capture_event_processor));
 
+  // TODO (187250643): Refactor this to be more readable and maybe remove parts that are not needed
+  // here (capture cancelled)
   capture_result.Then(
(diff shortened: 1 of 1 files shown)


src/OrbitGl/App.cpp
```diff
@@ -1093,6 +1093,8 @@ void OrbitApp::StartCapture() {
       enable_introspection, max_local_marker_depth_per_command_buffer, collect_memory_info,
       memory_sampling_period_ns, std::move(capture_event_processor));
 
+  // TODO (187250643): Refactor this to be more readable and maybe remove parts that are not needed
```
dpallotti, line 1096:
```suggestion
  // TODO (b/187250643): Refactor this to be more readable and maybe remove parts that are not needed
```

dpallotti (approved):
Thanks!

src/OrbitGl/App.cpp
antonrohr: done

dimitry-: Interestingly enough - if you do http://b/<number> here CLion will show you the status of the bug in the editor.

dpallotti: Yep, `b/` is enough to see the status, hence the suggestion.

PR #2283, opened 2021-05-05
[REBASE] Fix crash when hovering on manually instrumented async scope

karupayun:
TimerInfo::kApiEvent don't have an "InstrumentedFunction" associated
with them so don't expect one.

Test: profile OrbitTest with manual instrumentation enabled, hover on
a profiling scope in one of the "async" tracks.


pierricgimmig (approved):
Ah thanks a lot @karupayun ! Will do that straight away next time :-)

PR #2287, opened 2021-05-05
Allow instrumenting a function with OrbitFakeClient

dpallotti:
Bug: http://b/187177718

Test: Use `OrbitFakeClient` and Orbit UI to instrument
`eva::graphics::rendering::spatial_geometry::world_sphere() const`
(path: `/srv/game/assets/benchmark`, load bias: `0x200000`, address: `0x6ed730`,
offset: `0x4ed730`). Notice how OrbitService reports a similar number of
instrumentation events.

### DIFF
--- a/src/FakeClient/FakeClientMain.cpp
+++ b/src/FakeClient/FakeClientMain.cpp
@@ -12,12 +12,14 @@
 #include <atomic>
 #include <csignal>
 #include <cstdlib>
+#include <filesystem>
 #include <limits>
 #include <memory>
 #include <thread>
 
 #include "CaptureClient/CaptureClient.h"
 #include "CaptureClient/CaptureListener.h"
+#include "ElfUtils/ElfFile.h"
 #include "FakeCaptureEventProcessor.h"
 #include "GrpcProtos/Constants.h"
 #include "OrbitBase/Logging.h"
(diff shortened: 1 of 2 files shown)


src/FakeClient/FakeClientMain.cpp
```diff
@@ -57,6 +59,33 @@ void InstallSigintHandler() {
...
+  module_info.set_file_path(file_path);
+  constexpr const char* kFakeBuildId = "id";
+  module_info.set_build_id(kFakeBuildId);
+  constexpr uint64_t kFakeLoadBias = 0;
+  module_info.set_load_bias(kFakeLoadBias);
+  CHECK(module_manager->AddOrUpdateModules({module_info}).empty());
+
+  orbit_client_protos::FunctionInfo function_info;
+  function_info.set_module_path(file_path);
+  function_info.set_module_build_id(kFakeBuildId);
```
dimitry-, line 81: will this break once we have build_id validation on the service side?

florian-kuebler (approved):

src/FakeClient/FakeClientMain.cpp
dpallotti: Are we planning on adding validation of `InstrumentedFunction.file_build_id` to the service? If yes, then yes. At that point we can add another command line flag for it, or read it from the binary. I'm not particularly worried, but I also don't know what the specific plan is.

dimitry-: yes - there is bug for it: http://b/183380723

dpallotti: Fine, let's make this cleaner then. It should be easy to get the build id with `ElfFile`, right?

dimitry-: yes - it is pretty straightforward, you can also do load_bias correctly here using that.

dpallotti: Done.

src/FakeClient/FakeClientMain.cpp
```diff
@@ -57,6 +61,40 @@ void InstallSigintHandler() {
...
+  orbit_grpc_protos::ModuleInfo module_info;
+  module_info.set_file_path(file_path);
+  module_info.set_build_id(build_id);
+  module_info.set_load_bias(load_bias);
+  CHECK(module_manager->AddOrUpdateModules({module_info}).empty());
+
+  orbit_client_protos::FunctionInfo function_info;
+  function_info.set_module_path(file_path);
+  function_info.set_module_build_id(build_id);
+  function_info.set_address(load_bias + file_offset);
```
dpallotti, line 91: Should we pass the address instead of the offset as command line argument? What do you think?

dimitry- (approved):
thanks!

PR #2288, opened 2021-05-05
Save capture in new format on the client side.

dimitry-:
At the moment it is feature behind the enable_capture_autosave flag.

Also CompositeEventProcessor now assumes ownership over
CaptureEventProcessors.

Test: Manual - start Orbit with --enable_capture_autosave check the
      created file using hexdump -C (mostly header and some strings)

### DIFF
--- a/src/CaptureClient/CompositeEventProcessorTest.cpp
+++ b/src/CaptureClient/CompositeEventProcessorTest.cpp
@@ -15,16 +15,28 @@ class MockEventProcessor : public CaptureEventProcessor {
 };
 
 TEST(CompositeEventProcessor, Smoke) {
-  MockEventProcessor processor1;
-  MockEventProcessor processor2;
-  MockEventProcessor processor3;
+  // CompositeEventProcessor takes ownership of event processors used to construct it.
+  // But in the test we want to keep pointers to them to check that they are called.
+  std::vector<std::unique_ptr<CaptureEventProcessor>> event_processors;
+
+  auto processor = std::make_unique<MockEventProcessor>();
+  MockEventProcessor* processor1 = processor.get();
+  event_processors.push_back(std::move(processor));
+
+  processor = std::make_unique<MockEventProcessor>();
+  MockEventProcessor* processor2 = processor.get();
+  event_processors.push_back(std::move(processor));
+
+  processor = std::make_unique<MockEventProcessor>();
+  MockEventProcessor* processor3 = processor.get();
+  event_processors.push_back(std::move(processor));
 
   auto composite_processor =
-      CaptureEventProcessor::CreateCompositeProcessor({&processor1, &processor2, &processor3});
+      CaptureEventProcessor::CreateCompositeProcessor(std::move(event_processors));
 
-  EXPECT_CALL(processor1, ProcessEvent).Times(1);
-  EXPECT_CALL(processor2, ProcessEvent).Times(1);
-  EXPECT_CALL(processor3, ProcessEvent).Times(1);
+  EXPECT_CALL(*processor1, ProcessEvent).Times(1);
+  EXPECT_CALL(*processor2, ProcessEvent).Times(1);
+  EXPECT_CALL(*processor3, ProcessEvent).Times(1);
 
   orbit_grpc_protos::ClientCaptureEvent event;
   composite_processor->ProcessEvent(event);
--- a/src/OrbitGl/App.cpp
+++ b/src/OrbitGl/App.cpp
@@ -97,6 +97,7 @@ ABSL_DECLARE_FLAG(bool, devmode);
 ABSL_DECLARE_FLAG(bool, local);
 ABSL_DECLARE_FLAG(bool, enable_tracepoint_feature);
 ABSL_DECLARE_FLAG(bool, enable_source_code_view);
+ABSL_DECLARE_FLAG(bool, enable_capture_autosave);
 
 using orbit_capture_client::CaptureClient;
 using orbit_capture_client::CaptureEventProcessor;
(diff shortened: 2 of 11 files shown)


src/CaptureClient/CompositeEventProcessorTest.cpp
```diff
@@ -15,16 +15,28 @@ class MockEventProcessor : public CaptureEventProcessor {
 };
 
 TEST(CompositeEventProcessor, Smoke) {
-  MockEventProcessor processor1;
-  MockEventProcessor processor2;
-  MockEventProcessor processor3;
+  // CompositeEventProcessor assumes ownership of event processors used to construct it.
```
dpallotti, line 18:
```suggestion
  // CompositeEventProcessor takes ownership of event processors used to construct it.
```
"assume" is kind of ambiguous.

src/OrbitGl/ClientFlags.cpp
```diff
@@ -46,3 +46,7 @@ ABSL_FLAG(bool, enable_source_code_view, true, "Enable the experimental source c
 // threshold (i.e., production limit).
 ABSL_FLAG(bool, enable_warning_threshold, false,
           "Enable setting and showing the memory warning threshold");
+
+// TODO(dimitry): Set default to true in 1.65
+// TODO(dimitry): Remove this flag in the next release 1.66
```
dpallotti, line 51: You should probably create task bugs blocking the feature.

src/OrbitGl/App.cpp
```diff
@@ -1083,8 +1084,34 @@ void OrbitApp::StartCapture() {
 
   CHECK(capture_client_ != nullptr);
 
-  auto capture_event_processor =
-      CaptureEventProcessor::CreateForCaptureListener(this, std::move(frame_track_function_ids));
+  std::unique_ptr<CaptureEventProcessor> capture_event_processor;
+  auto capture_for_listener_processor =
```
dpallotti, line 1088: The order of the words is off. I suggest `processor_for_capture_listener` or something like that.

src/OrbitGl/App.cpp
```diff
@@ -1083,8 +1084,34 @@ void OrbitApp::StartCapture() {
...
+          SendErrorToUi("Error saving capture", error.message());
+        });
+    if (save_to_file_processor_or_error.has_value()) {
+      std::vector<std::unique_ptr<CaptureEventProcessor>> event_processors;
+      event_processors.push_back(std::move(capture_for_listener_processor));
+      event_processors.push_back(std::move(save_to_file_processor_or_error.value()));
+      capture_event_processor =
+          CaptureEventProcessor::CreateCompositeProcessor(std::move(event_processors));
+    } else {
+      SendErrorToUi("Error saving capture",
```
dpallotti, line 1107: Does this happen for example if the file already exists?

dpallotti (approved):
Looks good otherwise.

src/OrbitGl/App.cpp
```diff
@@ -1093,8 +1120,8 @@ void OrbitApp::StartCapture() {
       enable_introspection, max_local_marker_depth_per_command_buffer, collect_memory_info,
       memory_sampling_period_ns, std::move(capture_event_processor));
 
-  // TODO (187250643): Refactor this to be more readable and maybe remove parts that are not needed
-  // here (capture cancelled)
+  // TODO (http://b/187250643): Refactor this to be more readable and maybe remove parts that are
```
dpallotti, line 1123: I don't think you need the `http://`.

dimitry-: My personal preference to have http:// everywhere it is implied. Easier for tools to recognize that this is a link.

src/OrbitGl/App.cpp
```diff
@@ -1083,8 +1084,34 @@ void OrbitApp::StartCapture() {
...
+        });
+    if (save_to_file_processor_or_error.has_value()) {
+      std::vector<std::unique_ptr<CaptureEventProcessor>> event_processors;
+      event_processors.push_back(std::move(capture_for_listener_processor));
+      event_processors.push_back(std::move(save_to_file_processor_or_error.value()));
+      capture_event_processor =
+          CaptureEventProcessor::CreateCompositeProcessor(std::move(event_processors));
+    } else {
+      SendErrorToUi("Error saving capture",
+                    absl::StrFormat("Unable to save capture to \"%s\": %s", file_path.string(),
```
antonrohr, line 1108: This error might be a confusing for the user. This error potentially shows up on capture start, correct? Then maybe add some more context like: "Unable to set up automatic capture saving"

src/OrbitGl/App.cpp
```diff
@@ -1083,8 +1084,34 @@ void OrbitApp::StartCapture() {
 
   CHECK(capture_client_ != nullptr);
 
-  auto capture_event_processor =
-      CaptureEventProcessor::CreateForCaptureListener(this, std::move(frame_track_function_ids));
+  std::unique_ptr<CaptureEventProcessor> capture_event_processor;
+  auto capture_for_listener_processor =
+      CaptureEventProcessor::CreateForCaptureListener(this, frame_track_function_ids);
+
+  if (absl::GetFlag(FLAGS_enable_capture_autosave)) {
```
antonrohr, line 1091: This whole if else tree would be more readable if there are early returns. Please consider extracting it to an extra method so you can use return

src/OrbitGl/App.cpp
antonrohr: Sorry for that. Forgot to git add before amending yesterday. Also fixed it in https://github.com/google/orbit/pull/2291

antonrohr (approved):
I agree with @dpallotti's comments + 2 more nits

src/OrbitGl/App.cpp
dpallotti: Do you have specific tools that recognize it with `http://` and not without? The styleguide says `b/buganizer_id` and being consistent across our TODOs doesn't hurt.

antonrohr: vscode does not recognize it without the http:// 

I agree consistency would be nice. I will take this to the chat room and can then take care of the refactor for the whole codebase.

PR #2289, opened 2021-05-05
Also Expand Gpu-Subtracks

florian-kuebler:
When expanding the GpuTrack, the user wants to look into the Gpu related information.
They will probably also be interessted in submissions and debug markers. So save them
a click and also expand those tracks.

Bug: http://b/187269015
Test: Take a capture with the Vulkan layer, and collapse/expand the Gpu track.

### DIFF
--- a/src/OrbitGl/GpuSubmissionTrack.cpp
+++ b/src/OrbitGl/GpuSubmissionTrack.cpp
@@ -28,7 +28,7 @@ constexpr const char* kHwQueueString = "hw queue";
 constexpr const char* kHwExecutionString = "hw execution";
 constexpr const char* kCmdBufferString = "command buffer";
 
-GpuSubmissionTrack::GpuSubmissionTrack(CaptureViewElement* parent, TimeGraph* time_graph,
+GpuSubmissionTrack::GpuSubmissionTrack(Track* parent, TimeGraph* time_graph,
                                        orbit_gl::Viewport* viewport, TimeGraphLayout* layout,
                                        uint64_t timeline_hash, OrbitApp* app,
                                        const orbit_client_model::CaptureData* capture_data,
@@ -39,9 +39,10 @@ GpuSubmissionTrack::GpuSubmissionTrack(CaptureViewElement* parent, TimeGraph* ti
   text_renderer_ = time_graph->GetTextRenderer();
   timeline_hash_ = timeline_hash;
   string_manager_ = app->GetStringManager();
+  parent_ = parent;
 
-  // Gpu tracks are collapsed by default.
-  collapse_toggle_->SetState(TriangleToggle::State::kCollapsed,
+  // Gpu subtracks are expanded by default, but are however not shown until the parent is collapsed.
+  collapse_toggle_->SetState(TriangleToggle::State::kExpanded,
                              TriangleToggle::InitialStateUpdate::kReplaceInitialState);
 }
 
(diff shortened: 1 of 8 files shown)


src/OrbitGl/GpuTrack.cpp
```diff
@@ -270,10 +270,24 @@ const TextBox* GpuTrack::GetDown(const TextBox* textbox) const {
...
   // When being collapsed, we only show the submission track, and we want those timers also being
-  // collapsed. The marker track does not need to be touched, as it will not be drawn.
+  // collapsed. Further, when expanding the track the user likely wants to get more insights, thus
+  // we expand the subtracks.
   if (state == TriangleToggle::State::kCollapsed) {
-    submission_track_->Collapse();
+    if (submission_track_->IsCollapsible()) {
+      submission_track_->Collapse();
+    }
+    if (marker_track_->IsCollapsed()) {
```
dpallotti, line 280: Shouldn't this be `IsCollapsible`?

src/OrbitGl/GpuTrack.cpp
```diff
@@ -270,10 +270,24 @@ const TextBox* GpuTrack::GetDown(const TextBox* textbox) const {
...
+      marker_track_->Collapse();
+    }
+  } else if (state == TriangleToggle::State::kExpanded) {
+    if (submission_track_->IsCollapsible()) {
+      submission_track_->Expand();
+    }
+    if (marker_track_->IsCollapsed()) {
```
dpallotti, line 287: Shouldn't this be `IsCollapsible`?

src/OrbitGl/GpuTrack.cpp
```diff
@@ -270,10 +270,24 @@ const TextBox* GpuTrack::GetDown(const TextBox* textbox) const {
   }
 }
 void GpuTrack::OnCollapseToggle(TriangleToggle::State state) {
-  Track::OnCollapseToggle(state);
   // When being collapsed, we only show the submission track, and we want those timers also being
-  // collapsed. The marker track does not need to be touched, as it will not be drawn.
+  // collapsed. Further, when expanding the track the user likely wants to get more insights, thus
```
dpallotti, line 274: Doesn't this reset the collapsed state of the subtracks every time you expand or collapse the parent track? That could be annoying.

Could we do it that the subtracks start expanded even if the parent track starts collapsed, so that when you expand the parent track for the first time the subtracks are expanded, but then their collapsed state is independent of the parent track? (I hope what I'm saying makes sense.)

dpallotti:
By the way, thanks for looking into this right away.

src/OrbitGl/GpuTrack.cpp
```diff
@@ -270,10 +270,24 @@ const TextBox* GpuTrack::GetDown(const TextBox* textbox) const {
...
+  } else if (state == TriangleToggle::State::kExpanded) {
+    if (submission_track_->IsCollapsible()) {
+      submission_track_->Expand();
+    }
+    if (marker_track_->IsCollapsed()) {
+      marker_track_->Expand();
+    }
   }
+
+  Track::OnCollapseToggle(state);
```
karupayun, line 292: Nit: I will do this at first of the method and after make the real function in the subtracks but I don't have a "why-explanation"

karupayun (approved):
Nice PR. I agree with this solution and also with Davide's comments.

florian-kuebler:
As discussed offline, I changed it in the way that the subtracks are expanded by default and that they remember their state. The parent track being collapsed is realised by the the submission track knowing the parent.

src/OrbitGl/GpuSubmissionTrack.cpp
```diff
@@ -124,7 +125,7 @@ Color GpuSubmissionTrack::GetTimerColor(const TimerInfo& timer_info, bool is_sel
 
 float GpuSubmissionTrack::GetYFromTimer(const TimerInfo& timer_info) const {
   auto adjusted_depth = static_cast<float>(timer_info.depth());
-  if (collapse_toggle_->IsCollapsed()) {
+  if (IsCollapsed() || GetParent()->IsCollapsed()) {
```
dpallotti, line 128: I suggest adding an helper method, e.g., `ShouldShowCollapsed`, and using it in these three occurrences.

PR #2292, opened 2021-05-06
Refactor TODO comments

antonrohr:


dpallotti:
Can we avoid the spaces between `TODO` and `(` like in the style guide's examples?

dimitry- (approved):

src/OrbitTest/OrbitTest.cpp
```diff
@@ -21,7 +21,8 @@
 // OrbitBase defines an implementation of the Orbit API for introspection. To prevent symbol clash
 // with the header-only use of the API, we wrap the include of Orbit.h in a namespace. This is only
 // needed for internal projects that link with OrbitBase, basically only OrbitTest. This should be
-// removed once we split the introspection and manual instrumentation apis TODO(b/182848566).
+// removed once we split the introspection and manual instrumentation apis
+// TODO(b/182848566).
```
dpallotti, line 25: Could you move the TODO at the beginning of the paragraph?

dpallotti (approved):
Thanks!

src/OrbitGl/App.h
```diff
@@ -391,7 +391,7 @@ class OrbitApp final : public DataViewFactory, public orbit_capture_client::Capt
     return data_manager_->memory_warning_threshold_kb();
   }
 
-  // TODO(kuebler): Move them to a separate controler at some point
+  // TODO(kuebler): Move them to a separate controller at some point
```
ronaldfw, line 394: Should be a bug link here?

antonrohr: A bug id is preferred over ldap, but there are ~20 other locations where an ldap is used. 

I only accidentally saw this typo and decided to fix it. @florian-kuebler if you want to file a bug feel free.

src/OrbitTest/OrbitTest.cpp
antonrohr: done

src/OrbitGl/App.h
ronaldfw: Ah, it looked like this is the only place, but I guess there are TODOs that are already formatted correctly :) In that case, keep it as-is, and let's fix usernames later.

PR #2293, opened 2021-05-06
Load capture in new file format

dimitry-:
With this change loading capture first tries to load from the new
capture file format and then falls back to the old capture file format.

Test: Open automatically saved capture - visually compare it to original
      capture.


dpallotti (approved):

dpallotti:
Nice to see all the pieces coming together!

PR #2295, opened 2021-05-06
Make rounding corners work again

karupayun:
Regression from yesterday's refactor. It was a confusion between capture window background color and the track background color.

In addition, we rename track background color method and variables to avoid this problem in the future.

Test: Start a capture

### DIFF
--- a/src/OrbitGl/Track.cpp
+++ b/src/OrbitGl/Track.cpp
@@ -109,9 +109,9 @@ void Track::Draw(Batcher& batcher, TextRenderer& text_renderer, uint64_t current
   float text_z = GlCanvas::kZValueTrackText + z_offset;
   float top_margin = layout_->GetTrackTopMargin();
 
-  Color background_color = GetBackgroundColor();
+  Color track_background_color = GetTrackBackgroundColor();
   if (!draw_background_) {
-    background_color[3] = 0;
+    track_background_color[3] = 0;
   }
 
   // Draw tab.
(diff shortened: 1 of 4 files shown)


src/OrbitGl/Track.cpp
```diff
@@ -148,12 +148,16 @@ void Track::Draw(Batcher& batcher, TextRenderer& text_renderer, uint64_t current
     Vec2 content_top_right(top_left[0] + size_[0] - right_margin,
                            top_left[1] - label_height + top_margin);
 
-    DrawTriangleFan(batcher, rounded_corner, top_left, background_color, -90.f, track_z);
-    DrawTriangleFan(batcher, rounded_corner, tab_top_right, background_color, 180.f, track_z);
-    DrawTriangleFan(batcher, rounded_corner, tab_bottom_right, background_color, 0, track_z);
-    DrawTriangleFan(batcher, rounded_corner, content_bottom_left, background_color, 0, track_z);
-    DrawTriangleFan(batcher, rounded_corner, tab_bottom_right, background_color, 0, track_z);
-    DrawTriangleFan(batcher, rounded_corner, content_bottom_left, background_color, 0, track_z);
+    DrawTriangleFan(batcher, rounded_corner, top_left, GlCanvas::kBackgroundColor, -90.f, track_z);
```
florian-kuebler, line 151: Why is this GlCanvas::BackgroundColor and not `track_background_color`?

karupayun: Because we have to paint the top left corner with the background color from canvas on top of the track color to make the grey rounding corner

florian-kuebler: I am still a bit confused why this is changed now. But I trust that you checked the outcome.

florian-kuebler (approved):

src/OrbitGl/Track.cpp
karupayun: I've messed up. I wrote track background color thinking that was the same than GlCanvas::kBackgroundColor, it's a revert + a renaming.

PR #2296, opened 2021-05-06
Rename project OrbitClientData to ClientData, adjust include headers, and add namespace

dpallotti:


dimitry- (approved):

antonrohr (approved):

dimitry-:
I have just notices that you have introduced a nested namespace for function_utils. We want to avoid nested namespaces, maybe rename it to oribit_function_utils? or keep orbit_client_data and drop function_utils all together.

dpallotti:
> > I have just notices that you have introduced a nested namespace for function_utils. We want to avoid nested namespaces, maybe rename it to oribit_function_utils? or keep orbit_client_data and drop function_utils all together.
> 
> @dimitry-
> That was basically already a nested namespace, I just added the missing outer one.
> We have discussed this in the past in the same context. If we want some some way of grouping certain free-standing functions together we either use a nested namespace or we use a containing `struct`. For me we can also just drop `function_utils`, I have nothing against it, but then those functions are no longer grouped together in the module, so we need to be fine with it.
> But I disagree with renaming to `orbit_function_utils` as it breaks the rule of the namespace matching the module.

By the way, we also have `orbit_service::utils`, so I'm not even sure that "we want to avoid nested namespaces" actually applies.

dimitry-:
> By the way, we also have `orbit_service::utils`, so I'm not even sure that "we want to avoid nested namespaces" actually applies.

Yep we want would want remove it too.

PR #2298, opened 2021-05-06
Make capture loading look interactive

dimitry-:
While we do not have additional section we may as well enjoy some
interactivity. It now looks like capturing while loading from file.

Test: Open capture file in new format, check that it loads interactively
Bug: http://b/166094073


florian-kuebler (approved):

antonrohr:
LGTM

PR #2299, opened 2021-05-07
Update Live functions while loading capture from disk

dimitry-:
This also fixes a case with updating live function list if
a module gets loaded after the live capture is started.

Bug: http://b/187463865
Test: load capture from disk see that live function tab is populated and
      updated while the capture is loading.


florian-kuebler (approved):
LGTM

dpallotti (approved):
It looks good, but I'm slightly confused about when this happens. Why does this happen only when loading a capture (except for the cases of a module being loaded late by the target)?

dimitry-:
> Why does this happen only when loading a capture (except for the cases of a module being loaded late by the target)?

Since you cannot select a function without loading a module the function info for instrumented function is always available in the case of live capture. In the case of loading from disk we do have information about the module but we do not have functions because the module is not loaded (it actually is never loaded in this case).

antonrohr (approved):

PR #2303, opened 2021-05-10
Fix possible compilation errors in Orbit.h

pierricgimmig:
Wrap ORBIT_CALL macro in "do{...}while(0)" to avoid "dangling else" issue

The fact that there was an if statement in the macro made this code fail:
    
  if(cond)
    ORBIT_START(...);
  else
    ORBIT_STOP();

Also, fix C compilation of Orbit.h.

Tested in Unity and in C test application.

### DIFF
--- a/src/Api/include/Api/Orbit.h
+++ b/src/Api/include/Api/Orbit.h
@@ -5,6 +5,7 @@
 #ifndef ORBIT_API_ORBIT_H_
 #define ORBIT_API_ORBIT_H_
 
+#include <stdbool.h>
 #include <stdint.h>
 
 // =================================================================================================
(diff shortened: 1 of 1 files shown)


src/Api/include/Api/Orbit.h
```diff
@@ -323,7 +324,7 @@ inline bool orbit_api_active() {
 }
 
 #define ORBIT_CALL(function_name) \
-  if (orbit_api_active() && g_orbit_api_v0.function_name) g_orbit_api_v0.function_name
+  !(orbit_api_active() && g_orbit_api_v0.function_name) ? (void)0 : g_orbit_api_v0.function_name
```
dimitry-, line 327: question, can you incorporate function call inside the macro?

something like 
```
ORBIT_CALL(function_name, ...)
do { 
  if (...) function_name(__VA_ARGS__);
} while ...
```

dpallotti: That would be my thought, too.

dimitry- (approved):

PR #2306, opened 2021-05-10
Log time of loading from new capture file format

dimitry-:
Test: Load capture from old format and from new one, check logs


src/OrbitGl/App.cpp
```diff
@@ -993,8 +993,9 @@ ErrorMessageOr<void> OrbitApp::OnSaveCapture(const std::filesystem::path& file_n
 }
 
 static ErrorMessageOr<CaptureListener::CaptureOutcome> LoadCaptureFromNewFormat(
-    CaptureListener* listener, CaptureFile* capture_file,
+    const std::filesystem::path& file_path, CaptureListener* listener, CaptureFile* capture_file,
```
dpallotti, line 996: Can `CaptureFile` have a method that exposes `CaptureFileImpl::file_path_` or is there a reason against it?

It seems redundant to pass both `file_path` (only for logging reasons) and the respective `CaptureFile`.

src/CaptureFile/CaptureFile.cpp
```diff
@@ -259,6 +261,9 @@ std::optional<uint64_t> orbit_capture_file::CaptureFileImpl::FindSectionByType(
 
   return std::nullopt;
 }
+const std::filesystem::path& orbit_capture_file::CaptureFileImpl::GetFilePath() const {
```
dpallotti, line 264:
```suggestion
}

const std::filesystem::path& CaptureFileImpl::GetFilePath() const {
```

dpallotti: You don't need to specify `orbit_capture_file::`, right? (Also `FindSectionByType` right above)

dpallotti (approved):

src/CaptureFile/CaptureFile.cpp
dimitry-: yep - clion always does it to me for some reason.

dpallotti: I feel you.

dimitry-: done.

PR #2307, opened 2021-05-10
Patch Callstacks of Leaf Function Samples.

florian-kuebler:
In leaf functions (or samples that hit the prologue/epilogue),
`$rbp` might not point to the beginning of the frame. In those cases
perf_event_open will miss the sample's direct caller in the resulting
callstack (on frame-pointer-based unwinding).

This change will in addition to perf_event_open, unwind the first
two frames with libunwindstack, and if the sample's caller is
different, we know the missing pc. We will patch the sample in this
case.

Now that we handle the case that our sample falls inside the prologue,
the test does not need to discard those samples anymore.

Test: Compile a target with -momit-leaf-frame-pointer, inspect the
      callstack. Further, run the integration and unit tests.

Bug: http://b/160399871

### DIFF
--- a/src/ClientProtos/capture_data.proto
+++ b/src/ClientProtos/capture_data.proto
@@ -71,6 +71,7 @@ message CallstackInfo {
     kFramePointerUnwindingError = 2;
     kInUprobes = 3;
     kUprobesPatchingFailed = 4;
+    kFramePointerDwarfStackTooSmallError = 5;
     // These are set by the client and are in addition to the ones in
     // Callstack::CallstackType.
     kFilteredByMajorityOutermostFrame = 100;
--- a/src/ClientProtos/capture_data.proto
+++ b/src/ClientProtos/capture_data.proto
@@ -71,7 +71,8 @@ message CallstackInfo {
     kFramePointerUnwindingError = 2;
     kInUprobes = 3;
     kUprobesPatchingFailed = 4;
-    kFramePointerDwarfStackTooSmallError = 5;
+    kStackTopForDwarfUnwindingTooSmall = 5;
+    kStackTopDwarfUnwindingError = 6;
     // These are set by the client and are in addition to the ones in
     // Callstack::CallstackType.
     kFilteredByMajorityOutermostFrame = 100;
--- a/src/GrpcProtos/capture.proto
+++ b/src/GrpcProtos/capture.proto
@@ -142,7 +142,8 @@ message Callstack {
     kFramePointerUnwindingError = 2;
     kInUprobes = 3;
     kUprobesPatchingFailed = 4;
-    kFramePointerDwarfStackTooSmallError = 5;
+    kStackTopForDwarfUnwindingTooSmall = 5;
+    kStackTopDwarfUnwindingError = 6;
   }
   CallstackType type = 2;
 }
--- a/src/GrpcProtos/capture.proto
+++ b/src/GrpcProtos/capture.proto
@@ -142,6 +142,7 @@ message Callstack {
     kFramePointerUnwindingError = 2;
     kInUprobes = 3;
     kUprobesPatchingFailed = 4;
+    kFramePointerDwarfStackTooSmallError = 5;
   }
   CallstackType type = 2;
 }
--- a/src/LinuxTracing/LeafFunctionCallManager.cpp
+++ b/src/LinuxTracing/LeafFunctionCallManager.cpp
@@ -39,7 +39,7 @@ Callstack::CallstackType LeafFunctionCallManager::PatchLeafFunctionCaller(
 
   uint64_t stack_size =
       event->GetRegisters()[PERF_REG_X86_BP] - event->GetRegisters()[PERF_REG_X86_SP];
-  if (stack_size > SAMPLE_STACK_USER_SIZE_128BYTES) {
+  if (stack_size > SAMPLE_STACK_USER_SIZE_512BYTES) {
     return Callstack::kFramePointerDwarfStackTooSmallError;
   }
 
--- a/src/LinuxTracing/LeafFunctionCallManager.cpp
+++ b/src/LinuxTracing/LeafFunctionCallManager.cpp
@@ -22,14 +22,14 @@ using orbit_grpc_protos::Callstack;
 // Therefore, libunwindstack will be able to compute two frames, where the outer one is the missing
 // caller and needs to be patched in.
 // As libunwindstack will try to unwind even further, unwinding errors will always be reported.
-// We need to fallback to plausibility checks to detect actual unwinding errors (such as whether the
-// frames are executable).
+// We need to fall back to plausibility checks to detect actual unwinding errors (such as whether
+// the frames are executable).
 // Note that we cannot simply set libunwindstack to unwind always two frames and compare the outer
 // frame with the respective one in the callchain carried by the perf_event_open event, as in case
 // of uprobes overriding the return addresses, both addresses would be identical even if the actual
 // addresses (after uprobe patching) are not.
 // More (internal) documentation on this in: go/stadia-orbit-leaf-frame-pointer
-Callstack::CallstackType LeafFunctionCallManager::PatchLeafFunctionCaller(
+Callstack::CallstackType LeafFunctionCallManager::PatchCallerOfLeafFunction(
     CallchainSamplePerfEvent* event, LibunwindstackMaps* current_maps,
     LibunwindstackUnwinder* unwinder) {
   CHECK(event != nullptr);
@@ -37,10 +37,16 @@ Callstack::CallstackType LeafFunctionCallManager::PatchLeafFunctionCaller(
   CHECK(unwinder != nullptr);
   CHECK(event->GetCallchainSize() > 2);
 
-  uint64_t stack_size =
-      event->GetRegisters()[PERF_REG_X86_BP] - event->GetRegisters()[PERF_REG_X86_SP];
+  uint64_t rbp = event->GetRegisters()[PERF_REG_X86_BP];
+  uint64_t rsp = event->GetRegisters()[PERF_REG_X86_SP];
+
+  if (rbp < rsp) {
+    return Callstack::kFramePointerUnwindingError;
+  }
+
+  uint64_t stack_size = rbp - rsp;
   if (stack_size > SAMPLE_STACK_USER_SIZE_512BYTES) {
-    return Callstack::kFramePointerDwarfStackTooSmallError;
+    return Callstack::kStackTopForDwarfUnwindingTooSmall;
   }
 
   const LibunwindstackResult& libunwindstack_result =
@@ -52,7 +58,7 @@ Callstack::CallstackType LeafFunctionCallManager::PatchLeafFunctionCaller(
   if (libunwindstack_callstack.empty()) {
     ERROR("Discarding sample as DWARF-based unwinding resulted in empty callstack (error: %s)",
           LibunwindstackUnwinder::LibunwindstackErrorString(libunwindstack_result.error_code()));
-    return Callstack::kDwarfUnwindingError;
+    return Callstack::kStackTopDwarfUnwindingError;
   }
 
   // In case of an intact frame pointer, the region from $rsp to $rbp will only include the locals
@@ -83,7 +89,7 @@ Callstack::CallstackType LeafFunctionCallManager::PatchLeafFunctionCaller(
   // If the caller is not executable, we have an unwinding error.
   unwindstack::MapInfo* map_info = current_maps->Find(libunwindstack_leaf_caller_pc);
   if (map_info == nullptr || (map_info->flags & PROT_EXEC) == 0) {
-    return Callstack::kDwarfUnwindingError;
+    return Callstack::kStackTopDwarfUnwindingError;
   }
 
   // perf_event_open's callstack always contains the return address. Libunwindstack has already
--- a/src/LinuxTracing/LeafFunctionCallManager.cpp
+++ b/src/LinuxTracing/LeafFunctionCallManager.cpp
@@ -12,6 +12,8 @@
 
 namespace orbit_linux_tracing {
 
+using orbit_grpc_protos::Callstack;
+
 // Let's unwind the stack from $rsp to $rbp using libunwindstack. If $rbp points to the current
 // frame, this will only include the locals and not the return address, so libunwindstack will
 // only be able to unwind one frame (the instruction pointer). If $rpb points to the previous frame,
@@ -20,16 +22,16 @@ namespace orbit_linux_tracing {
 // Therefore, libunwindstack will be able to compute two frames, where the outer one is the missing
 // caller and needs to be patched in.
 // As libunwindstack will try to unwind even further, unwinding errors will always be reported.
-// We need to fallback to plausibility checks to detect actual unwinding errors (such as are the
-// frames executable).
-// Note, that we can not simply set libunwindstack to unwind always two frames and compare the outer
-// frame with the respective one in the perf_event_open event, as in case of uprobes overriding the
-// return addresses, both addresses would be identical even if the actual addresses (after uprobe
-// patching) are not.
+// We need to fallback to plausibility checks to detect actual unwinding errors (such as whether the
+// frames are executable).
+// Note that we cannot simply set libunwindstack to unwind always two frames and compare the outer
+// frame with the respective one in the callchain carried by the perf_event_open event, as in case
+// of uprobes overriding the return addresses, both addresses would be identical even if the actual
+// addresses (after uprobe patching) are not.
 // More (internal) documentation on this in: go/stadia-orbit-leaf-frame-pointer
-bool LeafFunctionCallManager::PatchLeafFunctionCaller(CallchainSamplePerfEvent* event,
-                                                      LibunwindstackMaps* current_maps,
-                                                      LibunwindstackUnwinder* unwinder) {
+Callstack::CallstackType LeafFunctionCallManager::PatchLeafFunctionCaller(
+    CallchainSamplePerfEvent* event, LibunwindstackMaps* current_maps,
+    LibunwindstackUnwinder* unwinder) {
   CHECK(event != nullptr);
   CHECK(current_maps != nullptr);
   CHECK(unwinder != nullptr);
@@ -38,9 +40,7 @@ bool LeafFunctionCallManager::PatchLeafFunctionCaller(CallchainSamplePerfEvent* 
   uint64_t stack_size =
       event->GetRegisters()[PERF_REG_X86_BP] - event->GetRegisters()[PERF_REG_X86_SP];
   if (stack_size > SAMPLE_STACK_USER_SIZE_128BYTES) {
-    ERROR("Discarding sample where the stack dump is to small (required %lu bytes, got %lu bytes)",
-          stack_size, SAMPLE_STACK_USER_SIZE_128BYTES);
-    return false;
+    return Callstack::kFramePointerDwarfStackTooSmallError;
   }
 
   const LibunwindstackResult& libunwindstack_result =
@@ -50,26 +50,22 @@ bool LeafFunctionCallManager::PatchLeafFunctionCaller(CallchainSamplePerfEvent* 
       libunwindstack_result.frames();
 
   if (libunwindstack_callstack.empty()) {
-    ERROR(
-        "Discarding sample as DWARF-based unwinding resulted in empty callstack (error code: %s).",
-        LibunwindstackUnwinder::LibunwindstackErrorString(libunwindstack_result.error_code()));
-    return false;
+    ERROR("Discarding sample as DWARF-based unwinding resulted in empty callstack (error: %s)",
+          LibunwindstackUnwinder::LibunwindstackErrorString(libunwindstack_result.error_code()));
+    return Callstack::kDwarfUnwindingError;
   }
 
   // In case of an intact frame pointer, the region from $rsp to $rbp will only include the locals
   // of the current frame, and NOT the return address. Thus, unwinding will only report one frame
   // (the instruction pointer) and we know that the original callchain is already correct.
   if (libunwindstack_callstack.size() == 1) {
-    return true;
+    return Callstack::kComplete;
   }
 
   // If unwinding results in more than two frames, $rbp was also not set correctly by the caller,
-  // thus frame pointers are not in all non-leaf functions, and our assumptions do not hold.
+  // thus frame pointers are not all in non-leaf functions, and our assumptions do not hold.
   if (libunwindstack_callstack.size() > 2) {
-    ERROR(
-        "Discarding sample as DWARF-based unwinding resulted in more than two frames, indicating "
-        "the absence of frame pointers.");
-    return false;
+    return Callstack::kFramePointerUnwindingError;
   }
 
   const std::vector<uint64_t> original_callchain = event->ips;
@@ -87,12 +83,7 @@ bool LeafFunctionCallManager::PatchLeafFunctionCaller(CallchainSamplePerfEvent* 
   // If the caller is not executable, we have an unwinding error.
   unwindstack::MapInfo* map_info = current_maps->Find(libunwindstack_leaf_caller_pc);
   if (map_info == nullptr || (map_info->flags & PROT_EXEC) == 0) {
-    ERROR(
-        "Discarding sample as DWARF-based unwounded caller with address 0x%x does not point to "
-        "executable code (error code: %s).",
-        libunwindstack_leaf_caller_pc,
-        LibunwindstackUnwinder::LibunwindstackErrorString(libunwindstack_result.error_code()));
-    return false;
+    return Callstack::kDwarfUnwindingError;
   }
 
   // perf_event_open's callstack always contains the return address. Libunwindstack has already
@@ -106,7 +97,7 @@ bool LeafFunctionCallManager::PatchLeafFunctionCaller(CallchainSamplePerfEvent* 
   event->ring_buffer_record.nr = result.size();
   event->ips = std::move(result);
 
-  return true;
+  return Callstack::kComplete;
 }
 
-}  //  namespace orbit_linux_tracing
+}  // namespace orbit_linux_tracing
--- a/src/LinuxTracing/LeafFunctionCallManager.cpp
+++ b/src/LinuxTracing/LeafFunctionCallManager.cpp
@@ -1,0 +1,112 @@
+// Copyright (c) 2021 The Orbit Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#include "LeafFunctionCallManager.h"
+
+#include <sys/mman.h>
+
+#include <vector>
+
+#include "OrbitBase/Logging.h"
+
+namespace orbit_linux_tracing {
+
+// Let's unwind the stack from $rsp to $rbp using libunwindstack. If $rbp points to the current
+// frame, this will only include the locals and not the return address, so libunwindstack will
+// only be able to unwind one frame (the instruction pointer). If $rpb points to the previous frame,
+// as the leaf function does not contain frame pointers, the region from $rsp to $rbp will contain
+// the complete innermost frame, plus the locals of the caller's frame (but not the return address).
+// Therefore, libunwindstack will be able to compute two frames, where the outer one is the missing
+// caller and needs to be patched in.
+// As libunwindstack will try to unwind even further, unwinding errors will always be reported.
+// We need to fallback to plausibility checks to detect actual unwinding errors (such as are the
+// frames executable).
+// Note, that we can not simply set libunwindstack to unwind always two frames and compare the outer
+// frame with the respective one in the perf_event_open event, as in case of uprobes overriding the
+// return addresses, both addresses would be identical even if the actual addresses (after uprobe
+// patching) are not.
+// More (internal) documentation on this in: go/stadia-orbit-leaf-frame-pointer
+bool LeafFunctionCallManager::PatchLeafFunctionCaller(CallchainSamplePerfEvent* event,
+                                                      LibunwindstackMaps* current_maps,
+                                                      LibunwindstackUnwinder* unwinder) {
+  CHECK(event != nullptr);
+  CHECK(current_maps != nullptr);
+  CHECK(unwinder != nullptr);
+  CHECK(event->GetCallchainSize() > 2);
+
+  uint64_t stack_size =
+      event->GetRegisters()[PERF_REG_X86_BP] - event->GetRegisters()[PERF_REG_X86_SP];
+  if (stack_size > SAMPLE_STACK_USER_SIZE_128BYTES) {
+    ERROR("Discarding sample where the stack dump is to small (required %lu bytes, got %lu bytes)",
+          stack_size, SAMPLE_STACK_USER_SIZE_128BYTES);
+    return false;
+  }
+
+  const LibunwindstackResult& libunwindstack_result =
+      unwinder->Unwind(event->GetPid(), current_maps->Get(), event->GetRegisters(),
+                       event->GetStackData(), stack_size, true);
+  const std::vector<unwindstack::FrameData>& libunwindstack_callstack =
+      libunwindstack_result.frames();
+
+  if (libunwindstack_callstack.empty()) {
+    ERROR(
+        "Discarding sample as DWARF-based unwinding resulted in empty callstack (error code: %s).",
+        LibunwindstackUnwinder::LibunwindstackErrorString(libunwindstack_result.error_code()));
+    return false;
+  }
+
+  // In case of an intact frame pointer, the region from $rsp to $rbp will only include the locals
+  // of the current frame, and NOT the return address. Thus, unwinding will only report one frame
+  // (the instruction pointer) and we know that the original callchain is already correct.
+  if (libunwindstack_callstack.size() == 1) {
+    return true;
+  }
+
+  // If unwinding results in more than two frames, $rbp was also not set correctly by the caller,
+  // thus frame pointers are not in all non-leaf functions, and our assumptions do not hold.
+  if (libunwindstack_callstack.size() > 2) {
+    ERROR(
+        "Discarding sample as DWARF-based unwinding resulted in more than two frames, indicating "
+        "the absence of frame pointers.");
+    return false;
+  }
+
+  const std::vector<uint64_t> original_callchain = event->ips;
+  CHECK(original_callchain.size() > 2);
+
+  std::vector<uint64_t> result;
+  result.reserve(original_callchain.size() + 1);
+  for (size_t i = 0; i < 2; ++i) {
+    result.push_back(original_callchain[i]);
+  }
+
+  CHECK(libunwindstack_callstack.size() == 2);
+  uint64_t libunwindstack_leaf_caller_pc = libunwindstack_callstack[1].pc;
+
+  // If the caller is not executable, we have an unwinding error.
+  unwindstack::MapInfo* map_info = current_maps->Find(libunwindstack_leaf_caller_pc);
+  if (map_info == nullptr || (map_info->flags & PROT_EXEC) == 0) {
+    ERROR(
+        "Discarding sample as DWARF-based unwounded caller with address 0x%x does not point to "
+        "executable code (error code: %s).",
+        libunwindstack_leaf_caller_pc,
+        LibunwindstackUnwinder::LibunwindstackErrorString(libunwindstack_result.error_code()));
+    return false;
+  }
+
+  // perf_event_open's callstack always contains the return address. Libunwindstack has already
+  // decreased the address by one. To not mix them, increase the address again.
+  result.push_back(libunwindstack_leaf_caller_pc + 1);
+
+  for (size_t i = 2; i < original_callchain.size(); ++i) {
+    result.push_back(original_callchain[i]);
+  }
+
+  event->ring_buffer_record.nr = result.size();
+  event->ips = std::move(result);
+
+  return true;
+}
+
+}  //  namespace orbit_linux_tracing
--- a/src/LinuxTracing/LeafFunctionCallManager.h
+++ b/src/LinuxTracing/LeafFunctionCallManager.h
@@ -2,8 +2,8 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
 
-#ifndef LINUX_TRACING_LEAF_FUNCTION_UTILS_H_
-#define LINUX_TRACING_LEAF_FUNCTION_UTILS_H_
+#ifndef LINUX_TRACING_LEAF_FUNCTION_CALL_MANAGER_H_
+#define LINUX_TRACING_LEAF_FUNCTION_CALL_MANAGER_H_
 
 #include <capture.pb.h>
 
@@ -13,9 +13,9 @@
 
 namespace orbit_linux_tracing {
 
-// This class provides the `PatchLeafFunctionCaller` method to fix a frame-pointer based callchain,
-// where the leaf function does not have frame-pointers.
-// Note that, this is wrapped in a class to allow tests to mock this implementation.
+// This class provides the `PatchCallerOfLeafFunction` method to fix a frame-pointer based
+// callchain, where the leaf function does not have frame-pointers. Note that this is wrapped in a
+// class to allow tests to mock this implementation.
 class LeafFunctionCallManager {
  public:
   virtual ~LeafFunctionCallManager() = default;
@@ -26,16 +26,16 @@ class LeafFunctionCallManager {
   // callchain), the respective `CallstackType` will be returned and the event remains untouched.
   // If the innermost frame has frame-pointers, this function will return `kComplete` and keeps the
   // callchain event untouched.
-  // Otherwise, that is the caller of the leaf function is missing and there are no unwinding
+  // Otherwise, that is if the caller of the leaf function is missing and there are no unwinding
   // errors, the callchain event gets updated, such that it contains the missing caller, and
   // `kComplete` will be returned.
   // Note that the address of the caller address is computed by decreasing the return address by
   // one in libunwindstack, to match the format of perf_event_open.
-  virtual orbit_grpc_protos::Callstack::CallstackType PatchLeafFunctionCaller(
+  virtual orbit_grpc_protos::Callstack::CallstackType PatchCallerOfLeafFunction(
       CallchainSamplePerfEvent* event, LibunwindstackMaps* current_maps,
       LibunwindstackUnwinder* unwinder);
 };
 
 }  //  namespace orbit_linux_tracing
 
-#endif  // LINUX_TRACING_LEAF_FUNCTION_UTILS_H_
+#endif  // LINUX_TRACING_LEAF_FUNCTION_CALL_MANAGER_H_
--- a/src/LinuxTracing/LeafFunctionCallManager.h
+++ b/src/LinuxTracing/LeafFunctionCallManager.h
@@ -1,0 +1,39 @@
+// Copyright (c) 2021 The Orbit Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#ifndef LINUX_TRACING_LEAF_FUNCTION_UTILS_H_
+#define LINUX_TRACING_LEAF_FUNCTION_UTILS_H_
+
+#include "LibunwindstackMaps.h"
+#include "LibunwindstackUnwinder.h"
+#include "PerfEvent.h"
+
+namespace orbit_linux_tracing {
+
+// This class provides the `PatchLeafFunctionCaller` method to fix a frame-pointer based callchain,
+// where the leaf function does not have frame-pointers.
+// Note that, this is wrapped in a class to allow tests to mock this implementation.
+class LeafFunctionCallManager {
+ public:
+  virtual ~LeafFunctionCallManager() = default;
+
+  // Computes the actual caller of a leaf function (that may not have frame-pointers) based on
+  // libunwindstack and modifies the given callchain event, if needed.
+  // In case of any unwinding error (either from libunwindstack or in the frame-pointer based
+  // callchain), `false` will be returned and the event remains untouched.
+  // If the innermost frame has frame-pointers, this function will return `true` and keeps the
+  // callchain event untouched.
+  // Otherwise, that is the caller of the leaf function is missing and there are no unwinding
+  // errors, the callchain event gets updated, such that it contains the missing caller, and `true`
+  // will be returned.
+  // Note that, the address of the caller address is computed by decreasing the return address by
+  // one in libunwindstack, to match the format of perf_event_open.
+  virtual bool PatchLeafFunctionCaller(CallchainSamplePerfEvent* event,
+                                       LibunwindstackMaps* current_maps,
+                                       LibunwindstackUnwinder* unwinder);
+};
+
+}  //  namespace orbit_linux_tracing
+
+#endif  // LINUX_TRACING_LEAF_FUNCTION_UTILS_H_
--- a/src/LinuxTracing/LeafFunctionCallManager.h
+++ b/src/LinuxTracing/LeafFunctionCallManager.h
@@ -5,6 +5,8 @@
 #ifndef LINUX_TRACING_LEAF_FUNCTION_UTILS_H_
 #define LINUX_TRACING_LEAF_FUNCTION_UTILS_H_
 
+#include <capture.pb.h>
+
 #include "LibunwindstackMaps.h"
 #include "LibunwindstackUnwinder.h"
 #include "PerfEvent.h"
@@ -21,17 +23,17 @@ class LeafFunctionCallManager {
   // Computes the actual caller of a leaf function (that may not have frame-pointers) based on
   // libunwindstack and modifies the given callchain event, if needed.
   // In case of any unwinding error (either from libunwindstack or in the frame-pointer based
-  // callchain), `false` will be returned and the event remains untouched.
-  // If the innermost frame has frame-pointers, this function will return `true` and keeps the
+  // callchain), the respective `CallstackType` will be returned and the event remains untouched.
+  // If the innermost frame has frame-pointers, this function will return `kComplete` and keeps the
   // callchain event untouched.
   // Otherwise, that is the caller of the leaf function is missing and there are no unwinding
-  // errors, the callchain event gets updated, such that it contains the missing caller, and `true`
-  // will be returned.
-  // Note that, the address of the caller address is computed by decreasing the return address by
+  // errors, the callchain event gets updated, such that it contains the missing caller, and
+  // `kComplete` will be returned.
+  // Note that the address of the caller address is computed by decreasing the return address by
   // one in libunwindstack, to match the format of perf_event_open.
-  virtual bool PatchLeafFunctionCaller(CallchainSamplePerfEvent* event,
-                                       LibunwindstackMaps* current_maps,
-                                       LibunwindstackUnwinder* unwinder);
+  virtual orbit_grpc_protos::Callstack::CallstackType PatchLeafFunctionCaller(
+      CallchainSamplePerfEvent* event, LibunwindstackMaps* current_maps,
+      LibunwindstackUnwinder* unwinder);
 };
 
 }  //  namespace orbit_linux_tracing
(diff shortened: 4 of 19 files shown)


src/LinuxTracing/UprobesUnwindingVisitor.cpp
```diff
@@ -98,12 +99,80 @@ void UprobesUnwindingVisitor::visit(CallchainSamplePerfEvent* event) {
...
+    return;
+  }
+
+  // The leaf function is not guaranteed to have the frame pointer for all our targets. Though, we
+  // assume that $rbp remains untouched by the leaf functions, such that we can rely on
+  // perf_event_open to give us "almost" correct callstacks (the caller of the leaf function will be
+  // missing). We can sanity check this assumption by checking if the callstack only contains
+  // executable code.
+  // TODO(b/187690455): As soon as we actually have frame pointers in all non-leaf functions, we can
+  //  "discard" the sample. Till that point, this check will always fail, as libc does not have
```
dpallotti, line 126: "as" would actually have to be two "as"s, or the sentence is broken.
I suggest, "Till that point this check will always fail, because the caller of __libc_start_main will be invalid since libc doesn't have frame-pointers".

src/LinuxTracing/UprobesUnwindingVisitor.cpp
```diff
@@ -98,12 +99,80 @@ void UprobesUnwindingVisitor::visit(CallchainSamplePerfEvent* event) {
...
+  // function is never the outermost frame. For example, for the main thread the outermost function
+  // is _start, followed by __libc_start_main. For other threads, the outermost function is clone.
+  if (libunwindstack_callstack.size() != 2) {
+    if (unwind_error_counter_ != nullptr) {
+      ++(*unwind_error_counter_);
+    }
+    return;
+  }
+
+  // Note, both, frame-pointer-based unwinding and libunwindstack use the return address of a
```
dpallotti, line 161: "Note:" or "Note that".

dpallotti, line 161:
```suggestion
  // Note, both frame-pointer-based unwinding and libunwindstack use the return address of a
```

src/LinuxTracing/UprobesUnwindingVisitor.cpp
```diff
@@ -98,12 +99,80 @@ void UprobesUnwindingVisitor::visit(CallchainSamplePerfEvent* event) {
...
+  // function call as the caller's address (which is the instruction after the call instruction), as
+  // they don't know the size of the call instruction. To fix this, libunwindstack decreases the
```
dpallotti, line 163: "this" here refers to the fact that the caller used is the instruction after the call instruction, but it's not really clear, as that's not the core of the previous sentence, rather it's in parenthesis. I think this should be slightly reformulated.

src/LinuxTracing/UprobesUnwindingVisitor.cpp
```diff
@@ -98,12 +99,80 @@ void UprobesUnwindingVisitor::visit(CallchainSamplePerfEvent* event) {
...
+  }
+
+  std::optional<uint64_t> leaf_function_caller_address;
+
+  // Unwind the first two frames with libunwindstack. If the outer frame (i.e. the caller of the
+  // function at $pc) is different to the one reported by perf_event_open, it was a leaf function,
+  // and we need to add this frame to perf_event_open's callstack.
+  const LibunwindstackCallstack& libunwindstack_result =
+      unwinder_.Unwind(event->GetPid(), current_maps_.get(), event->GetRegisters(),
+                       event->GetStackData(), event->GetStackSize(), 2);
```
dpallotti, line 145: Define a constant for 2.

src/LinuxTracing/UprobesUnwindingVisitor.cpp
```diff
@@ -98,12 +99,80 @@ void UprobesUnwindingVisitor::visit(CallchainSamplePerfEvent* event) {
...
+  const std::vector<unwindstack::FrameData>& libunwindstack_callstack =
+      libunwindstack_result.frames;
+
+  // We expect exactly two frames (as we've set `max_frames` to 2), otherwise it is an unwinding
+  // error.
+  // Note that this doesn't exclude samples inside the main function of any thread as the main
+  // function is never the outermost frame. For example, for the main thread the outermost function
+  // is _start, followed by __libc_start_main. For other threads, the outermost function is clone.
+  if (libunwindstack_callstack.size() != 2) {
```
dpallotti, line 154: Use the constant for 2 that you should define above.

dpallotti (commented):
I want to take another look. I'm leaving some comments in the meantime, most are minor.

I'd like `LinuxTracingIntegrationTests` to be run many many times to check everything is still in order, it was really taking a lot of runs to get all corner cases.

src/LinuxTracing/UprobesUnwindingVisitor.cpp
```diff
@@ -98,12 +99,80 @@ void UprobesUnwindingVisitor::visit(CallchainSamplePerfEvent* event) {
   CHECK(listener_ != nullptr);
```
dimitry-, line 99: maybe extract this of most of this function to something that is unit-testable?

dimitry-: I am kind of interested how this function react to a tail call landing to a leaf function.

florian-kuebler: Was also thinking of this, so I agree. I'll look into this. Also might look once more into tail-calls.

src/LinuxTracing/UprobesUnwindingVisitor.cpp
```diff
@@ -98,12 +99,80 @@ void UprobesUnwindingVisitor::visit(CallchainSamplePerfEvent* event) {
...
+  // executable code.
+  // TODO(b/187690455): As soon as we actually have frame pointers in all non-leaf functions, we can
+  //  "discard" the sample. Till that point, this check will always fail, as libc does not have
+  //  frame-pointers, the caller of __libc_start_main will be invalid.
+  //  Note, that at this point in time, we don't want to actually throw the sample away, but rather
+  //  report it as "broken".
+  for (uint64_t frame_index = 1; frame_index < event->GetCallchainSize(); ++frame_index) {
+    unwindstack::MapInfo* map_info = current_maps_->Find(event->GetCallchain()[frame_index]);
+    if (map_info == nullptr || (map_info->flags & PROT_EXEC) == 0) {
+      ERROR("$rbp was modified.");
```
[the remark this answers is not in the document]
florian-kuebler: Thanks., I've taken your suggestion.

src/LinuxTracingIntegrationTests/LinuxTracingIntegrationTest.cpp
```diff
@@ -894,12 +866,12 @@ TEST(LinuxTracingIntegrationTest, CallstackSamplesWithFramePointers) {
...
 
   // Note that this test requires that the "inner" function of the puppet use frame pointers.
-  VerifyCallstackSamplesWithOuterAndInnerFunctionForFramePointerUnwinding(
+  VerifyCallstackSamplesWithOuterAndInnerFunction(
       events, fixture.GetPuppetPid(), outer_function_virtual_address_range,
       inner_function_virtual_address_range, sampling_rate, /*address_infos_received=*/nullptr);
 }
 
-TEST(LinuxTracingIntegrationTest, CallstackSamplesWithFramePointersTogetherWithFunctionCalls) {
+TEST(LinuxTracingIntegrationTest, mmi) {
```
dimitry-, line 874: I do not think this was intentional right?

src/LinuxTracing/UprobesUnwindingVisitor.cpp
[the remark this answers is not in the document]
florian-kuebler: Yeah I was thinking the same. But when adding `ERROR_ONCE` to the code base, we decided not to go for an implementation that takes the content of the string. Sure we can change this, but as this example here is very temporary, I would not go this route.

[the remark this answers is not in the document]
dpallotti: I didn't mean using `ERROR_ONCE`, I meant having custom logic here to only print the error if the address hasn't already been seen.

src/LinuxTracing/LibunwindstackUnwinder.h
```diff
@@ -20,16 +20,26 @@
 
 namespace orbit_linux_tracing {
 
+struct LibunwindstackResult {
+  std::vector<unwindstack::FrameData> frames;
+  // If "true", Libunwindstack successfully unwound the callstack, but the actual stack has more
+  // frames than what was given as "max_frames", so this callstack is incomplete.
```
dpallotti, line 26: Came here to copy this `struct`. Just a heads up that this comment is outdated.

florian-kuebler: Thanks

florian-kuebler:
Updated the implementation with the findings from go/stadia-orbit-leaf-frame-pointer. It should now work also together with uprobes. Please take a new look.

src/LinuxTracing/LeafFunctionCallManager.cpp
```diff
@@ -0,0 +1,112 @@
...
+// only be able to unwind one frame (the instruction pointer). If $rpb points to the previous frame,
+// as the leaf function does not contain frame pointers, the region from $rsp to $rbp will contain
+// the complete innermost frame, plus the locals of the caller's frame (but not the return address).
+// Therefore, libunwindstack will be able to compute two frames, where the outer one is the missing
+// caller and needs to be patched in.
+// As libunwindstack will try to unwind even further, unwinding errors will always be reported.
+// We need to fallback to plausibility checks to detect actual unwinding errors (such as are the
+// frames executable).
+// Note, that we can not simply set libunwindstack to unwind always two frames and compare the outer
+// frame with the respective one in the perf_event_open event, as in case of uprobes overriding the
```
dpallotti, line 26: "in the perf_event_open event": you should mention the callchain (e.g., "in the callchain carried by the perf_event_open event").

src/LinuxTracing/LeafFunctionCallManager.cpp
```diff
@@ -0,0 +1,112 @@
...
+                                                      LibunwindstackUnwinder* unwinder) {
+  CHECK(event != nullptr);
+  CHECK(current_maps != nullptr);
+  CHECK(unwinder != nullptr);
+  CHECK(event->GetCallchainSize() > 2);
+
+  uint64_t stack_size =
+      event->GetRegisters()[PERF_REG_X86_BP] - event->GetRegisters()[PERF_REG_X86_SP];
+  if (stack_size > SAMPLE_STACK_USER_SIZE_128BYTES) {
+    ERROR("Discarding sample where the stack dump is to small (required %lu bytes, got %lu bytes)",
```
dpallotti, line 41:
```suggestion
    ERROR("Discarding sample where the stack dump is too small (required %lu bytes, got %lu bytes)",
```

dpallotti, line 41: This should basically never happen, right? Or is there a risk of log spamming?

src/LinuxTracing/LeafFunctionCallManager.cpp
```diff
@@ -0,0 +1,112 @@
...
+
+  const LibunwindstackResult& libunwindstack_result =
+      unwinder->Unwind(event->GetPid(), current_maps->Get(), event->GetRegisters(),
+                       event->GetStackData(), stack_size, true);
+  const std::vector<unwindstack::FrameData>& libunwindstack_callstack =
+      libunwindstack_result.frames();
+
+  if (libunwindstack_callstack.empty()) {
+    ERROR(
+        "Discarding sample as DWARF-based unwinding resulted in empty callstack (error code: %s).",
```
dpallotti, line 54: Maybe just "error" instead of "error code" as you're translating to the string.

src/LinuxTracing/LeafFunctionCallManager.cpp
```diff
@@ -0,0 +1,112 @@
...
+
+  // In case of an intact frame pointer, the region from $rsp to $rbp will only include the locals
+  // of the current frame, and NOT the return address. Thus, unwinding will only report one frame
+  // (the instruction pointer) and we know that the original callchain is already correct.
+  if (libunwindstack_callstack.size() == 1) {
+    return true;
+  }
+
+  // If unwinding results in more than two frames, $rbp was also not set correctly by the caller,
+  // thus frame pointers are not in all non-leaf functions, and our assumptions do not hold.
```
dpallotti, line 67:
```suggestion
  // thus frame pointers are not all in non-leaf functions, and our assumptions do not hold.
```

src/LinuxTracing/LeafFunctionCallManager.cpp
dpallotti, line 54:
```suggestion
        "Discarding sample as DWARF-based unwinding resulted in empty callstack (error code: %s)",
```

src/LinuxTracing/LeafFunctionCallManager.cpp
```diff
@@ -0,0 +1,112 @@
...
+  if (libunwindstack_callstack.size() > 2) {
+    ERROR(
+        "Discarding sample as DWARF-based unwinding resulted in more than two frames, indicating "
+        "the absence of frame pointers.");
```
dpallotti, line 71:
```suggestion
        "the absence of frame pointers");
```

src/LinuxTracing/LeafFunctionCallManager.cpp
```diff
@@ -0,0 +1,112 @@
...
+  // of the current frame, and NOT the return address. Thus, unwinding will only report one frame
+  // (the instruction pointer) and we know that the original callchain is already correct.
+  if (libunwindstack_callstack.size() == 1) {
+    return true;
+  }
+
+  // If unwinding results in more than two frames, $rbp was also not set correctly by the caller,
+  // thus frame pointers are not in all non-leaf functions, and our assumptions do not hold.
+  if (libunwindstack_callstack.size() > 2) {
+    ERROR(
```
dpallotti, line 69: Couldn't this be quite frequent if it just indicated the absence of frame pointers?

src/LinuxTracing/LeafFunctionCallManager.cpp
```diff
@@ -0,0 +1,112 @@
...
+  }
+
+  CHECK(libunwindstack_callstack.size() == 2);
+  uint64_t libunwindstack_leaf_caller_pc = libunwindstack_callstack[1].pc;
+
+  // If the caller is not executable, we have an unwinding error.
+  unwindstack::MapInfo* map_info = current_maps->Find(libunwindstack_leaf_caller_pc);
+  if (map_info == nullptr || (map_info->flags & PROT_EXEC) == 0) {
+    ERROR(
+        "Discarding sample as DWARF-based unwounded caller with address 0x%x does not point to "
```
dpallotti, line 91:
```suggestion
        "Discarding sample as DWARF-based unwounded caller with address %#x does not point to "
```

src/LinuxTracing/LeafFunctionCallManager.cpp
```diff
@@ -0,0 +1,112 @@
...
+        "executable code (error code: %s).",
```
dpallotti, line 92: "error code" -> "error"?

src/LinuxTracing/LeafFunctionCallManager.cpp
```diff
@@ -0,0 +1,112 @@
...
+    result.push_back(original_callchain[i]);
+  }
+
+  CHECK(libunwindstack_callstack.size() == 2);
+  uint64_t libunwindstack_leaf_caller_pc = libunwindstack_callstack[1].pc;
+
+  // If the caller is not executable, we have an unwinding error.
+  unwindstack::MapInfo* map_info = current_maps->Find(libunwindstack_leaf_caller_pc);
+  if (map_info == nullptr || (map_info->flags & PROT_EXEC) == 0) {
+    ERROR(
```
dpallotti, line 90: Couldn't this be very frequent and spam the logs if this an unwinding error?

src/LinuxTracing/UprobesUnwindingVisitorTest.cpp
```diff
@@ -89,6 +98,7 @@ class UprobesUnwindingVisitorTest : public ::testing::Test {
   MockUprobesReturnAddressManager return_address_manager_;
   MockLibunwindstackMaps maps_;
   MockLibunwindstackUnwinder unwinder_;
+  MockLeafFunctionCallerManager leaf_function_call_manager_;
```
dpallotti, line 101: MockLeafFunctionCallerManager -> MockLeafFunctionCallManager

src/LinuxTracing/LeafFunctionCallManager.h
```diff
@@ -0,0 +1,39 @@
...
+  // Computes the actual caller of a leaf function (that may not have frame-pointers) based on
+  // libunwindstack and modifies the given callchain event, if needed.
+  // In case of any unwinding error (either from libunwindstack or in the frame-pointer based
+  // callchain), `false` will be returned and the event remains untouched.
+  // If the innermost frame has frame-pointers, this function will return `true` and keeps the
+  // callchain event untouched.
+  // Otherwise, that is the caller of the leaf function is missing and there are no unwinding
+  // errors, the callchain event gets updated, such that it contains the missing caller, and `true`
+  // will be returned.
+  // Note that, the address of the caller address is computed by decreasing the return address by
```
dpallotti, line 30:
```suggestion
  // Note that the address of the caller address is computed by decreasing the return address by
```

src/LinuxTracing/LibunwindstackUnwinder.h
```diff
@@ -45,9 +45,15 @@ class LibunwindstackUnwinder {
 
   virtual LibunwindstackResult Unwind(pid_t pid, unwindstack::Maps* maps,
                                       const std::array<uint64_t, PERF_REG_X86_64_MAX>& perf_regs,
-                                      const void* stack_dump, uint64_t stack_dump_size) = 0;
+                                      const void* stack_dump, uint64_t stack_dump_size,
+                                      bool offline_memory = false,
```
dpallotti, line 49: I would call this "offline_memory_only" to be clearer, when this is false *some* offline memory is still used.

src/LinuxTracing/LibunwindstackUnwinder.cpp
```diff
@@ -67,20 +67,13 @@ class LibunwindstackUnwinderImpl : public LibunwindstackUnwinder {
  public:
   LibunwindstackResult Unwind(pid_t pid, unwindstack::Maps* maps,
                               const std::array<uint64_t, PERF_REG_X86_64_MAX>& perf_regs,
-                              const void* stack_dump, uint64_t stack_dump_size) override;
+                              const void* stack_dump, uint64_t stack_dump_size, bool offline_memory,
```
dpallotti, line 70: The default value of `offline_memory` is not specified here, while the default value of `max_frames` is. I guess they should both be specified (or none?)?

src/LinuxTracing/LibunwindstackUnwinder.cpp
```diff
@@ -93,17 +86,22 @@ const std::array<size_t, unwindstack::X86_64_REG_LAST>
...
   unwindstack::RegsX86_64 regs{};
   for (size_t perf_reg = 0; perf_reg < unwindstack::X86_64_REG_LAST; ++perf_reg) {
     regs[perf_reg] = perf_regs.at(kUnwindstackRegsToPerfRegs[perf_reg]);
   }
 
-  std::shared_ptr<unwindstack::Memory> memory = StackAndProcessMemory::Create(
-      pid, static_cast<const uint8_t*>(stack_dump), regs[unwindstack::X86_64_REG_RSP],
-      regs[unwindstack::X86_64_REG_RSP] + stack_dump_size);
+  std::shared_ptr<unwindstack::Memory> memory =
+      (offline_memory)
```
dpallotti, line 96: Consider an `if`-`else`, I think it's more readable.

src/LinuxTracing/UprobesUnwindingVisitor.cpp
```diff
@@ -129,9 +130,12 @@ void UprobesUnwindingVisitor::visit(CallchainSamplePerfEvent* event) {
...
 
-  // The top of a callchain is always inside the kernel code and we don't expect samples to be only
-  // inside the kernel.
-  if (event->GetCallchainSize() <= 1) {
+  // Callstacks with only two frames (the first is in the kernel, the second is the sampled address)
+  // are unwinding errors.
+  // Note that this doesn't exclude samples inside the main function of any thread as the main
+  // function is never the outermost frame. For example, for the main thread the outermost function
+  // is _start, followed by __libc_start_main. For other threads, the outermost function is clone.
+  if (event->GetCallchainSize() <= 2) {
```
dpallotti, line 138: Just a note that I'm planning to split this in two, `== 2` as unwinding error and `== 1` as unexpected (there should always be at least one (real) frame).

src/LinuxTracing/UprobesUnwindingVisitor.cpp
```diff
@@ -172,6 +161,47 @@ void UprobesUnwindingVisitor::visit(CallchainSamplePerfEvent* event) {
...
+  // TODO(b/187690455): As soon as we actually have frame pointers in all non-leaf functions, we can
+  //  "discard" the sample. Till that point this check will always fail, because the caller of
+  //  __libc_start_main will be invalid since libc doesn't have frame-pointers.
+  //  Note that, that at this point in time, we don't want to actually throw the sample away, but
+  //  rather report it as "broken".
+  for (uint64_t frame_index = 1; frame_index < event->GetCallchainSize(); ++frame_index) {
+    unwindstack::MapInfo* map_info = current_maps_->Find(event->GetCallchain()[frame_index]);
+    if (map_info == nullptr || (map_info->flags & PROT_EXEC) == 0) {
+      ERROR(
+          "Unwinding failed, $rbp was likely modified in the leaf-function and no longer pointer "
```
dpallotti, line 178:
```suggestion
          "Unwinding failed, $rbp was likely modified in the leaf-function and no longer points "
```

src/LinuxTracing/UprobesUnwindingVisitor.cpp
```diff
@@ -172,6 +161,47 @@ void UprobesUnwindingVisitor::visit(CallchainSamplePerfEvent* event) {
...
+  // contains executable code.
+  // TODO(b/187690455): As soon as we actually have frame pointers in all non-leaf functions, we can
+  //  "discard" the sample. Till that point this check will always fail, because the caller of
+  //  __libc_start_main will be invalid since libc doesn't have frame-pointers.
+  //  Note that, that at this point in time, we don't want to actually throw the sample away, but
+  //  rather report it as "broken".
+  for (uint64_t frame_index = 1; frame_index < event->GetCallchainSize(); ++frame_index) {
+    unwindstack::MapInfo* map_info = current_maps_->Find(event->GetCallchain()[frame_index]);
+    if (map_info == nullptr || (map_info->flags & PROT_EXEC) == 0) {
+      ERROR(
```
dpallotti, line 177: This could spam the logs I believe.

src/LinuxTracing/UprobesUnwindingVisitor.cpp
```diff
@@ -172,6 +161,47 @@ void UprobesUnwindingVisitor::visit(CallchainSamplePerfEvent* event) {
...
 
+  // The leaf function is not guaranteed to have the frame pointer for all our targets. Though, we
+  // assume that $rbp remains untouched by the leaf functions, such that we can rely on
+  // perf_event_open to give us "almost" correct callstacks (the caller of the leaf function will be
+  // missing). We do a plausibility check for this assumption by checking if the callstack only
+  // contains executable code.
+  // TODO(b/187690455): As soon as we actually have frame pointers in all non-leaf functions, we can
+  //  "discard" the sample. Till that point this check will always fail, because the caller of
+  //  __libc_start_main will be invalid since libc doesn't have frame-pointers.
+  //  Note that, that at this point in time, we don't want to actually throw the sample away, but
```
dpallotti, line 172:
```suggestion
  //  Note that, at this point in time, we don't want to actually throw the sample away, but
```

src/LinuxTracing/UprobesUnwindingVisitor.cpp
```diff
@@ -172,6 +161,47 @@ void UprobesUnwindingVisitor::visit(CallchainSamplePerfEvent* event) {
...
+  //  rather report it as "broken".
```
dpallotti, line 173: I'm confused by this TODO. Don't we want to do it right away?

src/LinuxTracing/LeafFunctionCallManagerTest.cpp
```diff
@@ -0,0 +1,333 @@
...
+
+    EXPECT_CALL(maps_, Find(AllOf(Ge(kTargetMapsStart), Lt(kTargetMapsEnd))))
+        .WillRepeatedly(Return(&kTargetMapInfo));
+
+    EXPECT_CALL(maps_, Find(AllOf(Ge(kNonExecutableMapsStart), Lt(kNonExecutableMapsEnd))))
+        .WillRepeatedly(Return(&kNonExecutableMapInfo));
+  }
+
+  void TearDown() override {}
+  MockLibunwindstackMaps maps_;
```
dpallotti, line 53:
```suggestion
  void TearDown() override {}

  MockLibunwindstackMaps maps_;
```

src/LinuxTracing/LeafFunctionCallManagerTest.cpp
```diff
@@ -0,0 +1,333 @@
...
+  static constexpr uint64_t kNonExecutableMapsStart = 500;
+  static constexpr uint64_t kNonExecutableMapsEnd = 600;
+
+  static constexpr uint64_t kKernelAddress = 11;
+
+  static constexpr uint64_t kTargetAddress1 = 100;
+  //  static constexpr uint64_t kUprobesAddress = 42;
+  static constexpr uint64_t kTargetAddress2 = 200;
+  static constexpr uint64_t kTargetAddress3 = 300;
+  //  static constexpr uint64_t kNonExecutableAddress3 = 500;
```
dpallotti, line 73: These comments should be removed I guess.

src/LinuxTracing/LeafFunctionCallManagerTest.cpp
```diff
@@ -0,0 +1,333 @@
...
+  };
+  event.ring_buffer_record.sample_id = sample_id;
+  event.ips = callchain;
+  event.regs.bp = SAMPLE_STACK_USER_SIZE_128BYTES;
+  event.regs.sp = 10;
+
+  EXPECT_CALL(maps_, Get).WillRepeatedly(Return(nullptr));
+
+  // When libunwindstack reports exactly one frame (the ip), the innermost function has frame
+  // pointers.
```
dpallotti, line 316: But we have two frames here. This refers to the previous test, right?

src/LinuxTracingIntegrationTests/LinuxTracingIntegrationTest.cpp
```diff
@@ -758,14 +758,6 @@ void VerifyCallstackSamplesWithOuterAndInnerFunction(
       // address and the caller address should match the "outer" function's address.
       if (callstack.pcs(pc_index) >= inner_function_virtual_address_range.first &&
           callstack.pcs(pc_index) <= inner_function_virtual_address_range.second) {
-        // Frame-pointer unwinding skips the second innermost frame if the sample fell on `push rbp`
```
dpallotti, line 761: Have you run the affected tests at least, say, 1000 times? Just to be sure.

src/LinuxTracing/UprobesUnwindingVisitorTest.cpp
```diff
@@ -615,15 +662,8 @@ TEST_F(UprobesUnwindingVisitorTest,
...
-
-  std::vector<unwindstack::FrameData> libunwindstack_callstack;
-  libunwindstack_callstack.push_back(kFrame1);
-  libunwindstack_callstack.push_back(kFrame2);
-
-  EXPECT_CALL(unwinder_, Unwind(kPid, nullptr, _, _, kStackSize))
-      .WillRepeatedly(Return(
-          LibunwindstackResult{libunwindstack_callstack, unwindstack::ErrorCode::ERROR_NONE}));
+  EXPECT_CALL(leaf_function_call_manager_, PatchLeafFunctionCaller).Times(1).WillOnce(Return(true));
+  EXPECT_CALL(return_address_manager_, PatchCallchain).Times(1).WillRepeatedly(Return(false));
```
dpallotti, line 666: `WillOnce`?

src/LinuxTracing/UprobesUnwindingVisitorTest.cpp
```diff
@@ -668,40 +708,47 @@ TEST_F(UprobesUnwindingVisitorTest, VisitCallchainSampleInsideUprobeCodeSendsInU
...
 
   EXPECT_EQ(unwinding_errors, 0);
-  EXPECT_EQ(discarded_samples_in_uretprobes_counter, 1);
+  EXPECT_EQ(discarded_samples_in_uretprobes_counter, 0);
 }
 
-// TODO(b/160399871): Disabled until we have support for handling leaf calls without frame pointers.
 TEST_F(UprobesUnwindingVisitorTest,
-       DISABLED_VisitLeafCallOptimizedCallchainSampleWithoutUprobesSendsCompleteCallstack) {
+       VisitLeafCallOptimizedCallchainSampleWherePatchingLeafFunctionCallerFails) {
```
dpallotti, line 751: Fails => SendsFramePointerUnwindingErrorCallstack (yes, a bit long, but more consistent)

dpallotti (commented):
Looks good overall. One common theme in my comments are all the `ERROR`s, it seems to me that some of them would be frequent and hence spam the logs. Plus some questions here and there.

src/LinuxTracing/LeafFunctionCallManager.cpp
florian-kuebler: Right, that SHOULD never happen.

src/LinuxTracingIntegrationTests/LinuxTracingIntegrationTest.cpp
florian-kuebler: There still seem to be an issue. I look into this right now.

florian-kuebler: there are actually also quite large frames. I changed the size to 512 bytes. Still, there are some samples that gets discarded then, so I adjusted the test to also except this error kind.

src/LinuxTracing/LeafFunctionCallManager.h
```diff
@@ -0,0 +1,41 @@
...
+  // In case of any unwinding error (either from libunwindstack or in the frame-pointer based
+  // callchain), the respective `CallstackType` will be returned and the event remains untouched.
+  // If the innermost frame has frame-pointers, this function will return `kComplete` and keeps the
+  // callchain event untouched.
+  // Otherwise, that is the caller of the leaf function is missing and there are no unwinding
+  // errors, the callchain event gets updated, such that it contains the missing caller, and
+  // `kComplete` will be returned.
+  // Note that the address of the caller address is computed by decreasing the return address by
+  // one in libunwindstack, to match the format of perf_event_open.
+  virtual orbit_grpc_protos::Callstack::CallstackType PatchLeafFunctionCaller(
```
dpallotti, line 34: I was thinking, for clarity, of renaming this to `PatchCallerOfLeafFunction`. I think it's clearer. Otherwise every time I read it I expected "leaf function call" and instead there's "caller". I had the same problem when reviewing this last time so I think it makes sense. This would require some renaming in the tests (test names, mocks, ...) and comments as well.

src/LinuxTracing/LeafFunctionCallManager.h
```diff
@@ -0,0 +1,41 @@
...
+
+#include "LibunwindstackMaps.h"
+#include "LibunwindstackUnwinder.h"
+#include "PerfEvent.h"
+
+namespace orbit_linux_tracing {
+
+// This class provides the `PatchLeafFunctionCaller` method to fix a frame-pointer based callchain,
+// where the leaf function does not have frame-pointers.
+// Note that, this is wrapped in a class to allow tests to mock this implementation.
```
dpallotti, line 18:
```suggestion
// Note that this is wrapped in a class to allow tests to mock this implementation.
```

src/LinuxTracing/LeafFunctionCallManager.h
```diff
@@ -0,0 +1,41 @@
...
+ public:
+  virtual ~LeafFunctionCallManager() = default;
+
+  // Computes the actual caller of a leaf function (that may not have frame-pointers) based on
+  // libunwindstack and modifies the given callchain event, if needed.
+  // In case of any unwinding error (either from libunwindstack or in the frame-pointer based
+  // callchain), the respective `CallstackType` will be returned and the event remains untouched.
+  // If the innermost frame has frame-pointers, this function will return `kComplete` and keeps the
+  // callchain event untouched.
+  // Otherwise, that is the caller of the leaf function is missing and there are no unwinding
```
dpallotti, line 29:
```suggestion
  // Otherwise, that is if the caller of the leaf function is missing and there are no unwinding
```

src/LinuxTracing/LeafFunctionCallManager.h
```diff
@@ -0,0 +1,41 @@
+// Copyright (c) 2021 The Orbit Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#ifndef LINUX_TRACING_LEAF_FUNCTION_UTILS_H_
```
dpallotti, line 5: Update the include guard.

src/LinuxTracing/LeafFunctionCallManager.cpp
```diff
@@ -0,0 +1,103 @@
...
+
+// Let's unwind the stack from $rsp to $rbp using libunwindstack. If $rbp points to the current
+// frame, this will only include the locals and not the return address, so libunwindstack will
+// only be able to unwind one frame (the instruction pointer). If $rpb points to the previous frame,
+// as the leaf function does not contain frame pointers, the region from $rsp to $rbp will contain
+// the complete innermost frame, plus the locals of the caller's frame (but not the return address).
+// Therefore, libunwindstack will be able to compute two frames, where the outer one is the missing
+// caller and needs to be patched in.
+// As libunwindstack will try to unwind even further, unwinding errors will always be reported.
+// We need to fallback to plausibility checks to detect actual unwinding errors (such as whether the
```
dpallotti, line 25:
```suggestion
// We need to fall back to plausibility checks to detect actual unwinding errors (such as whether the
```

src/LinuxTracing/LeafFunctionCallManager.cpp
```diff
@@ -0,0 +1,103 @@
...
+
+#include <vector>
+
+#include "OrbitBase/Logging.h"
+
+namespace orbit_linux_tracing {
+
+using orbit_grpc_protos::Callstack;
+
+// Let's unwind the stack from $rsp to $rbp using libunwindstack. If $rbp points to the current
```
dpallotti, line 17: Non-actionable comment to say that I really like this comment.

src/LinuxTracing/LeafFunctionCallManager.cpp
```diff
@@ -0,0 +1,103 @@
...
+  const LibunwindstackResult& libunwindstack_result =
+      unwinder->Unwind(event->GetPid(), current_maps->Get(), event->GetRegisters(),
+                       event->GetStackData(), stack_size, true);
+  const std::vector<unwindstack::FrameData>& libunwindstack_callstack =
+      libunwindstack_result.frames();
+
+  if (libunwindstack_callstack.empty()) {
+    ERROR("Discarding sample as DWARF-based unwinding resulted in empty callstack (error: %s)",
+          LibunwindstackUnwinder::LibunwindstackErrorString(libunwindstack_result.error_code()));
+    return Callstack::kDwarfUnwindingError;
```
dpallotti, line 55: This is the type that is then sent with the incomplete callstack (correct me if I'm wrong). I was wondering whether it would be better the have a separate type for dwarf unwinding errors that happen here (e.g., `kStackTopDwarfUnwindingError`), to distinguish between these and the ones that happen when actually using full DWARF-unwinding? Did you think about this or what do you think?

src/LinuxTracing/LeafFunctionCallManager.cpp
```diff
@@ -0,0 +1,103 @@
...
+    result.push_back(original_callchain[i]);
+  }
+
+  CHECK(libunwindstack_callstack.size() == 2);
+  uint64_t libunwindstack_leaf_caller_pc = libunwindstack_callstack[1].pc;
+
+  // If the caller is not executable, we have an unwinding error.
+  unwindstack::MapInfo* map_info = current_maps->Find(libunwindstack_leaf_caller_pc);
+  if (map_info == nullptr || (map_info->flags & PROT_EXEC) == 0) {
+    return Callstack::kDwarfUnwindingError;
```
dpallotti, line 86: Same thought here about possibly having a separate type for DWARF errors happening in this function.

src/LinuxTracing/LeafFunctionCallManagerTest.cpp
```diff
@@ -0,0 +1,338 @@
...
+      .res = 0,
+  };
+  event.ring_buffer_record.sample_id = sample_id;
+  event.ips = callchain;
+  event.regs.bp = SAMPLE_STACK_USER_SIZE_512BYTES;
+  event.regs.sp = 10;
+
+  EXPECT_CALL(maps_, Get).WillRepeatedly(Return(nullptr));
+
+  // Unwinding errors could result in empty callstacks:
```
dpallotti, line 178: I agree with supporting (and testing) this but this comment confuses me a bit as I think that in general we don't expect empty callstacks from libunwindstack (we should always get at least the IP).

src/LinuxTracing/UprobesUnwindingVisitor.cpp
dpallotti: Sorry 😅

src/LinuxTracingIntegrationTests/LinuxTracingIntegrationTest.cpp
florian-kuebler: Exactly. I've seen a sample where rbp and rsp was more than 1000 bytes apart. Also I've seen a sample, where the difference was negative (probably by rbp not being set for whatever reason).

src/LinuxTracing/LeafFunctionCallManager.cpp
florian-kuebler: Would be fine for me. My intend was that it is actually the same kind of error: We do Libunwindstack and that fails (for whatever internal reason).

src/LinuxTracing/LeafFunctionCallManager.cpp
```diff
@@ -0,0 +1,103 @@
...
+// More (internal) documentation on this in: go/stadia-orbit-leaf-frame-pointer
+Callstack::CallstackType LeafFunctionCallManager::PatchLeafFunctionCaller(
+    CallchainSamplePerfEvent* event, LibunwindstackMaps* current_maps,
+    LibunwindstackUnwinder* unwinder) {
+  CHECK(event != nullptr);
+  CHECK(current_maps != nullptr);
+  CHECK(unwinder != nullptr);
+  CHECK(event->GetCallchainSize() > 2);
+
+  uint64_t stack_size =
```
dpallotti, line 40: So related to your answer to the other comment, could this be negative (in theory, it's a `uint64_t`)? Or did I misunderstand? How does `Unwind` react to a then very big number?

src/LinuxTracing/UprobesUnwindingVisitor.cpp
florian-kuebler: Yes, at least until we have frame pointers in glibc. Otherwise, we will end for all frame-pointer based samples in being unwinding errors. That would be a large burden for us, as we could not test the frame-pointer code anymore. Same will be probably true for our integration tests, though.

dpallotti: Ah, so basically this is the main place where we check if frame pointer unwinding was successful, and we want to get wrong but complete callstacks (instead of the unwind error) for now for experimental purposes. Is that correct? In that case that's fine but maybe that should be clarified in the comment.
I'm still wondering whether the `ERROR` is spammy when experimenting but then that's for you to say.

src/GrpcProtos/capture.proto
```diff
@@ -142,6 +142,7 @@ message Callstack {
     kFramePointerUnwindingError = 2;
     kInUprobes = 3;
     kUprobesPatchingFailed = 4;
+    kFramePointerDwarfStackTooSmallError = 5;
```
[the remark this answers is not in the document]
florian-kuebler: I am going for `kStackTopForDwarfUnwindingTooSmall`

src/ClientProtos/capture_data.proto
```diff
@@ -71,6 +71,7 @@ message CallstackInfo {
     kFramePointerUnwindingError = 2;
     kInUprobes = 3;
     kUprobesPatchingFailed = 4;
+    kFramePointerDwarfStackTooSmallError = 5;
```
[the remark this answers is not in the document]
florian-kuebler: Renamed it to `kStackTopForDwarfUnwindingTooSmall`

src/LinuxTracing/LeafFunctionCallManager.cpp
florian-kuebler: Done.

src/LinuxTracing/LeafFunctionCallManager.cpp
florian-kuebler: That's a good point, I explicitly handled the case that rbp is smaller than rsp now and abort early.

src/LinuxTracing/LeafFunctionCallManager.cpp
florian-kuebler: Added a `kStackTopDwarfUnwindingError`

src/LinuxTracing/LeafFunctionCallManager.cpp
florian-kuebler: Added a `kStackTopDwarfUnwindingError`

src/LinuxTracing/LeafFunctionCallManager.h
florian-kuebler: Done.

src/LinuxTracing/LeafFunctionCallManager.h
florian-kuebler: Done.

src/LinuxTracing/LeafFunctionCallManager.h
florian-kuebler: Done.

src/LinuxTracing/LeafFunctionCallManager.h
florian-kuebler: Sounds good, done!

src/LinuxTracing/LeafFunctionCallManagerTest.cpp
florian-kuebler: Rephrased the comment.

src/LinuxTracing/LibunwindstackUnwinder.cpp
```diff
@@ -67,20 +67,14 @@ class LibunwindstackUnwinderImpl : public LibunwindstackUnwinder {
  public:
   LibunwindstackResult Unwind(pid_t pid, unwindstack::Maps* maps,
                               const std::array<uint64_t, PERF_REG_X86_64_MAX>& perf_regs,
-                              const void* stack_dump, uint64_t stack_dump_size) override;
+                              const void* stack_dump, uint64_t stack_dump_size,
+                              bool offline_memory_only = false,
+                              size_t max_frames = kMaxFrames) override;
 
  private:
   static constexpr size_t kMaxFrames = 1024;  // This is arbitrary.
```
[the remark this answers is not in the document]
florian-kuebler: Ah thanks. Removed it.

src/LinuxTracing/LibunwindstackUnwinder.h
```diff
@@ -45,9 +45,15 @@ class LibunwindstackUnwinder {
...
-                                      const void* stack_dump, uint64_t stack_dump_size) = 0;
+                                      const void* stack_dump, uint64_t stack_dump_size,
+                                      bool offline_memory_only = false,
+                                      size_t max_frames = kMaxFrames) = 0;
 
   static std::unique_ptr<LibunwindstackUnwinder> Create();
+  static std::string LibunwindstackErrorString(unwindstack::ErrorCode error_code);
+
+ protected:
+  static constexpr size_t kMaxFrames = 1024;  // This is arbitrary.
```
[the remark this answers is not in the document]
florian-kuebler: Done.

src/LinuxTracing/UprobesUnwindingVisitor.cpp
```diff
@@ -154,16 +155,17 @@ void UprobesUnwindingVisitor::visit(CallchainSamplePerfEvent* event) {
...
-
-  if (!return_address_manager_->PatchCallchain(event->GetTid(), event->GetCallchain(),
-                                               event->GetCallchainSize(), current_maps_)) {
+  // Callstacks with only two frames (the first is in the kernel, the second is the sampled address)
+  // are unwinding errors.
+  // Note that this doesn't exclude samples inside the main function of any thread as the main
+  // function is never the outermost frame. For example, for the main thread the outermost function
+  // is _start, followed by __libc_start_main. For other threads, the outermost function is clone.
+  if (event->GetCallchainSize() == 2) {
+    ERROR("Callchain has only %lu frames", event->GetCallchainSize());
```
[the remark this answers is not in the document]
florian-kuebler: Removed the `ERROR`.

src/LinuxTracing/UprobesUnwindingVisitor.cpp
```diff
@@ -184,6 +186,49 @@ void UprobesUnwindingVisitor::visit(CallchainSamplePerfEvent* event) {
     return;
   }
 
+  // The leaf function is not guaranteed to have the frame pointer for all our targets. Though, we
+  // assume that $rbp remains untouched by the leaf functions, such that we can rely on
+  // perf_event_open to give us "almost" correct callstacks (the caller of the leaf function will be
+  // missing). We do a plausibility check for this assumption by checking if the callstack only
+  // contains executable code.
+  // TODO(b/187690455): As soon as we actually have frame pointers in all non-leaf functions, we can
+  //  "discard" the sample. Till that point this check will always fail, because the caller of
```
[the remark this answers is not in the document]
florian-kuebler: Rephrased it.

src/LinuxTracing/UprobesUnwindingVisitor.cpp
florian-kuebler: I removed the error and changed the comment. Left the rest of the code as is for now.

src/LinuxTracing/UprobesUnwindingVisitorTest.cpp
```diff
@@ -671,40 +719,49 @@ TEST_F(UprobesUnwindingVisitorTest, VisitCallchainSampleInsideUprobeCodeSendsInU
...
-  EXPECT_EQ(discarded_samples_in_uretprobes_counter, 1);
+  EXPECT_EQ(discarded_samples_in_uretprobes_counter, 0);
 }
 
-// TODO(b/160399871): Disabled until we have support for handling leaf calls without frame pointers.
-TEST_F(UprobesUnwindingVisitorTest,
-       DISABLED_VisitLeafCallOptimizedCallchainSampleWithoutUprobesSendsCompleteCallstack) {
+TEST_F(
+    UprobesUnwindingVisitorTest,
+    VisitLeafCallOptimizedCallchainSampleWherePatchingLeafFunctionCallerSendsFramePointerUnwindingErrorCallstack) {
```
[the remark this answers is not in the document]
florian-kuebler: Oh no, thanks. Added the `Fails` to the name.

florian-kuebler (commented):
Thanks for the review! I addressed the comments

src/LinuxTracing/LeafFunctionCallManagerTest.cpp
```diff
@@ -0,0 +1,339 @@
...
+      .pc = kNonExecutableMapsStart,
+      .function_name = "???",
+      .function_offset = 0,
+      .map_name = kNonExecutableName,
+  };
+};
+
+}  // namespace
+
+TEST_F(LeafFunctionCallManagerTest, PatchLeafFunctionCallerReturnsErrorOnSmallStackSamples) {
```
dpallotti, line 120: Yeah, now all these tests should be renamed changing "PatchLeafFunctionCaller" to "PatchCallerOfLeafFunction".

src/LinuxTracing/UprobesUnwindingVisitor.cpp
```diff
@@ -184,6 +185,48 @@ void UprobesUnwindingVisitor::visit(CallchainSamplePerfEvent* event) {
...
+  // perf_event_open to give us "almost" correct callstacks (the caller of the leaf function will be
+  // missing). We do a plausibility check for this assumption by checking if the callstack only
+  // contains executable code.
+  // TODO(b/187690455): As soon as we actually have frame pointers in all non-leaf functions, we can
+  //  report an unwinding error here. Till that point this check will always fail, because the
+  //  caller of __libc_start_main will be invalid since libc doesn't have frame-pointers. This
+  //  prevents us from testing the current implementation, which will have "almost" correct
+  //  callstack.
+  //  Note that, at this point in time, we don't want to actually throw the sample away, but rather
+  //  report it as "broken".
```
dpallotti, line 199: Is this last part still relevant, now that in general we are no longer throwing samples away?

src/LinuxTracingIntegrationTests/LinuxTracingIntegrationTest.cpp
dpallotti: Do you think that the cases with such big frames (for example 336) were actually caused by a big frame, or could it be that rbp was just completely wrong? In other words, is it actually worth it to increase from 128 to 512? I'm wondering because increasing to 512 B has a performance impact (for example with regard to the performance impact investigation you had done on this).

src/LinuxTracing/UprobesUnwindingVisitorTest.cpp
```diff
@@ -671,40 +719,49 @@ TEST_F(UprobesUnwindingVisitorTest, VisitCallchainSampleInsideUprobeCodeSendsInU
...
-  libunwindstack_callstack.push_back(kFrame1);
-  libunwindstack_callstack.push_back(kFrame2);
-
-  EXPECT_CALL(unwinder_, Unwind(kPid, nullptr, _, _, kStackSize))
-      .WillRepeatedly(Return(
-          LibunwindstackResult{libunwindstack_callstack, unwindstack::ErrorCode::ERROR_NONE}));
+  EXPECT_CALL(maps_, Find).WillRepeatedly(Return(&kTargetMapInfo));
+  EXPECT_CALL(return_address_manager_, PatchCallchain).Times(1).WillRepeatedly(Return(true));
+
+  auto fake_patch_leaf_function_caller = [](CallchainSamplePerfEvent* event,
```
dpallotti, line 725: I guess `fake_patch_caller_of_leaf_function`.

dpallotti (approved):
Looks good!

I have a last minute doubt regarding increasing 128 B to 512 B for the top of the stack.

src/LinuxTracing/UprobesUnwindingVisitorTest.cpp
florian-kuebler: Done

src/LinuxTracing/UprobesUnwindingVisitor.cpp
florian-kuebler: Good point, removed it.

src/LinuxTracing/LeafFunctionCallManagerTest.cpp
florian-kuebler: Done.

src/LinuxTracingIntegrationTests/LinuxTracingIntegrationTest.cpp
florian-kuebler: As discussed offline. Keep it as it for now. We need to do some further experiments anyways.

PR #2308, opened 2021-05-10
Erase RequestUpdatePrimitives in Tracks

karupayun:
We solve it by making every CaptureViewElement tell this parent that it need to be updated, so we don't need TimeGraph for that anymore.

Also moved GetThreadName to ThreadTrack.

http://b/176086740.

Test: Start/stop + Load a capture

### DIFF
--- a/src/OrbitGl/CaptureViewElement.cpp
+++ b/src/OrbitGl/CaptureViewElement.cpp
@@ -24,12 +24,18 @@ void CaptureViewElement::OnPick(int x, int y) {
 
 void CaptureViewElement::OnRelease() {
   picked_ = false;
-  time_graph_->RequestUpdatePrimitives();
+  RequestUpdate();
 }
 
 void CaptureViewElement::OnDrag(int x, int y) {
   mouse_pos_cur_ = viewport_->ScreenToWorldPos(Vec2i(x, y));
-  time_graph_->RequestUpdatePrimitives();
+  RequestUpdate();
+}
+
+void CaptureViewElement::RequestUpdate() {
+  if (parent_ != nullptr) {
+    parent_->RequestUpdate();
+  }
 }
 
 }  // namespace orbit_gl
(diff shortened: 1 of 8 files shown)


src/OrbitGl/CaptureViewElement.h
```diff
@@ -46,6 +46,7 @@ class CaptureViewElement : public Pickable, public AccessibleInterfaceProvider {
   [[nodiscard]] bool Draggable() override { return true; }
 
   [[nodiscard]] virtual CaptureViewElement* GetParent() const { return parent_; }
+  virtual void RequestUpdate() { return parent_->RequestUpdate(); }
```
reichlfl, line 49: We should allow `parent_ == nullptr` IMO.

src/OrbitGl/TimeGraph.h
```diff
@@ -56,6 +56,7 @@ class TimeGraph : public orbit_gl::CaptureViewElement {
   void DrawText(float layer);
 
   void RequestUpdatePrimitives();
+  void RequestUpdate() override { RequestUpdatePrimitives(); }
```
reichlfl, line 59: Should we completely remove `RequestUpdatePrimitives` and replace it with `RequestUpdate` for consistency?

reichlfl (approved):
I'd add handling of `parent_ == nullptr`, otherwise LGTM, thanks!

src/OrbitGl/TimeGraph.h
karupayun: I will replace it.

src/OrbitGl/CaptureViewElement.cpp
```diff
@@ -32,4 +32,10 @@ void CaptureViewElement::OnDrag(int x, int y) {
   RequestUpdate();
 }
 
+void CaptureViewElement::RequestUpdate() {
+  if (parent_ != nullptr) {
+    return parent_->RequestUpdate();
```
reichlfl, line 37: You're returning `void`, I assume this is not intended?

reichlfl (approved):
I'd remove the `return` I pointed out - otherwise LGTM, thanks!

src/OrbitGl/CaptureViewElement.cpp
karupayun: oh, thanks! Yeah, of course

karupayun: Actually it was before, I didn't see it.

PR #2311, opened 2021-05-11
Save function names in presets

dimitry-:
Get rid of hashes in preset files.

### DIFF
--- a/src/ClientData/include/ClientData/ModuleData.h
+++ b/src/ClientData/include/ClientData/ModuleData.h
@@ -62,6 +62,9 @@ class ModuleData final {
   orbit_grpc_protos::ModuleInfo module_info_;
   bool is_loaded_;
   std::map<uint64_t, std::unique_ptr<orbit_client_protos::FunctionInfo>> functions_;
+  absl::flat_hash_map<std::string_view, orbit_client_protos::FunctionInfo*>
+      name_to_function_info_map_;
+
   // TODO(b/168799822) This is a map of hash to function used for preset loading. Currently presets
   // are based on a hash of the functions pretty name. This should be changed to not use hashes
   // anymore.
--- a/src/ClientProtos/preset.proto
+++ b/src/ClientProtos/preset.proto
@@ -6,12 +6,21 @@ syntax = "proto3";
 
 package orbit_client_protos;
 
+// This message is used for loading presets from
+// old format, do not use it to save presets.
 message PresetModule {
   repeated uint64 function_hashes = 1;
   repeated uint64 frame_track_function_hashes = 2;
 }
 
+message PresetModuleInfo {
+  repeated string function_names = 1;
+  repeated string frame_track_function_names = 2;
+}
+
 message PresetInfo {
   reserved 1;
+  // This field is deprecated, use modules instead.
   map<string, PresetModule> path_to_module = 2;
+  map<string, PresetModuleInfo> modules = 3;
 }
--- a/src/ClientProtos/preset.proto
+++ b/src/ClientProtos/preset.proto
@@ -15,9 +15,3 @@ message PresetInfo {
   reserved 1;
   map<string, PresetModule> path_to_module = 2;
 }
-
-// this message isn't used for serialization
-message PresetFile {
-  string file_name = 1;
-  PresetInfo preset_info = 2;
-}
--- a/src/OrbitGl/App.cpp
+++ b/src/OrbitGl/App.cpp
@@ -118,7 +118,6 @@ using orbit_client_protos::CallstackEvent;
 using orbit_client_protos::FunctionInfo;
 using orbit_client_protos::FunctionStats;
 using orbit_client_protos::LinuxAddressInfo;
-using orbit_client_protos::PresetFile;
 using orbit_client_protos::PresetInfo;
 using orbit_client_protos::TimerInfo;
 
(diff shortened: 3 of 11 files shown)


src/OrbitGl/App.cpp
```diff
@@ -1673,6 +1674,19 @@ void OrbitApp::SelectFunctionsFromHashes(const ModuleData* module,
   }
 }
 
+void OrbitApp::SelectFunctionsByName(const ModuleData* module,
+                                     absl::Span<const std::string> function_names) {
+  for (const auto& function_name : function_names) {
+    const orbit_client_protos::FunctionInfo* const function_info =
+        module->FindFunctionFromPrettyName(function_name);
```
pierricgimmig, line 1681: Follow up from previous comment: we could use the hash "function_name" to retrieve the FunctionInfo

src/OrbitGl/PresetFile.cpp
```diff
@@ -0,0 +1,72 @@
...
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#include "PresetFile.h"
+
+#include "OrbitBase/Logging.h"
+
+namespace orbit_gl {
+
+static std::vector<std::filesystem::path> GetPresetModulePathsLegacy(
```
pierricgimmig, line 11: nodiscard?

src/ClientProtos/preset.proto
```diff
@@ -6,18 +6,21 @@ syntax = "proto3";
...
 message PresetModule {
   repeated uint64 function_hashes = 1;
   repeated uint64 frame_track_function_hashes = 2;
 }
 
+message PresetModuleInfo {
+  repeated string function_names = 1;
+  repeated string frame_track_function_names = 2;
+}
+
```
pierricgimmig, line 20: Could we have added the new fields in "PresetModule" so that we'd have had the same code path for both the old and new format? 

Something like: 

message PresetModule {
  repeated uint64 function_hashes = 1;
  repeated uint64 frame_track_function_hashes = 2;
  repeated string function_names = 3;
  repeated string frame_track_function_names = 4;
}

And then we eventually deprecate 1 and 2?

pierricgimmig (commented):
Hi Dimitry, thanks a lot for tackling this! Please see comments, my main concern is with string duplication.

src/ClientProtos/preset.proto
dimitry-: In which case it could be that one module has combination of hashes and function names next one only function names and the last one only hashes. I think it is a bit clearer when we separate it on the top level. We do not want to accidently end up with mixed file format.

src/ClientData/include/ClientData/ModuleData.h
```diff
@@ -62,6 +62,8 @@ class ModuleData final {
   orbit_grpc_protos::ModuleInfo module_info_;
   bool is_loaded_;
   std::map<uint64_t, std::unique_ptr<orbit_client_protos::FunctionInfo>> functions_;
+  absl::flat_hash_map<std::string, orbit_client_protos::FunctionInfo*> name_to_function_info_map_;
```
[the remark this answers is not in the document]
dimitry-: because hashes collide. and with comparing number of possible values of hashes and strings they mathematically have to collide. I do not think there is a correct way of translating a string to uint64_t without a collision.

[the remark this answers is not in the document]
beckerhe: `std::string_view` might work here as well, right?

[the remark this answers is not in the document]
dimitry-: Interesting. I haven't thought about this. It will solve the problem. I will try that.

src/OrbitGl/PresetFile.cpp
```diff
@@ -0,0 +1,72 @@
...
+bool PresetFile::IsLegacyFileFormat() const { return preset_info_.modules().empty(); }
+
+std::vector<uint64_t> PresetFile::GetSelectedFunctionHashesForModuleLegacy(
+    const std::filesystem::path& module_path) const {
+  CHECK(IsLegacyFileFormat());
+  const auto& function_hashes =
+      preset_info_.path_to_module().at(module_path.string()).function_hashes();
+  return {function_hashes.begin(), function_hashes.end()};
+}
+std::vector<uint64_t> PresetFile::GetFrameTrackFunctionHashesForModuleLegacy(
```
beckerhe, line 51: Super nit: An empty line would be nice.

beckerhe (approved):
Thanks a lot Dimitry. LGTM!

src/OrbitGl/PresetFile.cpp
dimitry-: done.

src/OrbitGl/PresetFile.cpp
dimitry-: done.

antonrohr:
I did not look at the code, but I assume this will break compatibility with old preset files? If that is the case, please make sure the corresponding bug is included in the sprint hotlist and has a release note

dimitry-:
> I did not look at the code, but I assume this will break compatibility with old preset files? If that is the case, please make sure the corresponding bug is included in the sprint hotlist and has a release note

It won't.

src/ClientProtos/preset.proto
pierricgimmig: ok sounds good to me

PR #2316, opened 2021-05-12
Automatically convert old presets to new format (but do not save it yet)

dimitry-:
Automatic conversion will be enabled after new format becomes human readable.

Bug: http://b/168799822
Test: Open old preset -> check preset folder, it now has .backup file
      and new preset file.

PR #2317, opened 2021-05-13
Make new preset format human-readable

dimitry-:
And enable auto-conversion from old to new format on successful preset load.

Bug: http://b/168799822
Test: Load preset in old format, check that it is in new format
      and human readable.

### DIFF
--- a/src/OrbitGl/App.cpp
+++ b/src/OrbitGl/App.cpp
@@ -119,7 +119,7 @@ using orbit_client_protos::FunctionInfo;
 using orbit_client_protos::FunctionStats;
 using orbit_client_protos::LinuxAddressInfo;
 using orbit_client_protos::PresetInfo;
-using orbit_client_protos::PresetModuleInfo;
+using orbit_client_protos::PresetModule;
 using orbit_client_protos::TimerInfo;
 
 using orbit_client_services::CrashManager;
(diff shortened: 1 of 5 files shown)


dimitry-:
first commit is duplicate of #2316 you can look only at the second commit.

src/OrbitGl/PresetFile.cpp
```diff
@@ -4,12 +4,46 @@
...
 
+#include <absl/strings/match.h>
+#include <google/protobuf/io/zero_copy_stream_impl_lite.h>
+#include <google/protobuf/text_format.h>
+
+#include "OrbitBase/File.h"
 #include "OrbitBase/Logging.h"
+#include "OrbitBase/ReadFileToString.h"
 
 namespace orbit_gl {
```
dpallotti, line 15: Not related to this change, but what's the reason for this to be in `OrbitGl`?

src/OrbitGl/App.cpp
```diff
@@ -1773,16 +1745,15 @@ void OrbitApp::LoadPreset(const PresetFile& preset_file) {
...
-      //            absl::StrFormat("Unable to convert preset file \"%s\" to new file format: %s",
-      //                            preset_file.file_path().string(),
-      //                            convertion_result.error().message()));
-      //      }
+    } else {
+      // Then if load was successful and the preset is in old format - convert it to new one.
+      auto convertion_result = ConvertPresetToNewFormatIfNecessary(preset_file);
+      if (convertion_result.has_error()) {
+        SendWarningToUi(
+            "Conversion to new format failed",
```
dpallotti, line 1753: Should we really show this to the user?

dpallotti (approved):

src/OrbitGl/PresetFile.cpp
dimitry-: I guess it wants to be its own component... I will create a followup CL for this

src/OrbitGl/App.cpp
dimitry-: Converted to ERROR, also added the code to restore backed up file in case SaveToFile fails.

dimitry-: not sure where my comment went. I converted it to ERROR and added "restore from .backup" functionality to ConvertPresetToNewFormatIfNeccessary().

PR #2318, opened 2021-05-14
Replace std::map by absl::btree_map for better performance in ScopeTree

pierricgimmig:


src/OrbitGl/ScopeTree.h
```diff
@@ -36,8 +38,8 @@ class ScopeNode {
 
   [[nodiscard]] ScopeNode* GetLastChildBeforeOrAtTime(uint64_t time) const;
   [[nodiscard]] std::vector<ScopeNode*> GetChildrenInRange(uint64_t start, uint64_t end) const;
-  [[nodiscard]] std::map<uint64_t, ScopeNode*>& GetChildrenByStartTime() {
-    return children_by_start_time_;
+  [[nodiscard]] absl::btree_map<uint64_t, ScopeNode*>& GetChildrenByStartTime() const {
```
dimitry-, line 41: const method should return const reference, otherwise it is misleading, if caller modifies the map than method is not const.

src/OrbitGl/ScopeTree.h
```diff
@@ -59,7 +61,7 @@ class ScopeNode {
  private:
   ScopeT* scope_ = nullptr;
   uint32_t depth_ = 0;
-  std::map<uint64_t, ScopeNode<ScopeT>*> children_by_start_time_;
+  std::unique_ptr<absl::btree_map<uint64_t, ScopeNode<ScopeT>*>> children_by_start_time_;
```
dimitry-, line 64: why is it unique_ptr?

src/OrbitGl/ScopeTree.h
```diff
@@ -25,7 +25,9 @@ template <typename ScopeT>
 class ScopeNode {
  public:
   ScopeNode() = default;
```
dimitry-, line 27: children_by_start_time_ is going to be nullptr after default c-tor, is this intended?

dimitry- (approved):

src/OrbitGl/ScopeTree.h
pierricgimmig: Added a comment.

src/OrbitGl/ScopeTree.h
pierricgimmig: Fixed by using non-static data member initialization.

src/OrbitGl/ScopeTree.h
pierricgimmig: Added const.

PR #2319, opened 2021-05-14
Add "emplace_back" method to BlockChain and TimerChain

pierricgimmig:
Remove "Add", "push_back" and "Last" in favor or "emplace_back" which
uses placement-new to avoid unnacessary copies when appending to
chains. Note that "emplace_back" returns a reference to the newly
appended element, which makes "Last" obsolete.

### DIFF
--- a/src/OrbitGl/ThreadTrack.cpp
+++ b/src/OrbitGl/ThreadTrack.cpp
@@ -429,14 +429,29 @@ float ThreadTrack::GetYFromDepth(uint32_t depth) const {
 }
 
 void ThreadTrack::OnTimer(const TimerInfo& timer_info) {
-  TimerTrack::OnTimer(timer_info);
+  UpdateDepth(timer_info.depth() + 1);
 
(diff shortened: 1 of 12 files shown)


src/OrbitGl/ThreadTrack.cpp
```diff
@@ -429,14 +429,29 @@ float ThreadTrack::GetYFromDepth(uint32_t depth) const {
...
 void ThreadTrack::OnTimer(const TimerInfo& timer_info) {
-  TimerTrack::OnTimer(timer_info);
+  UpdateDepth(timer_info.depth() + 1);
+
+  if (process_id_ == -1) {
+    process_id_ = timer_info.process_id();
+  }
+
+  // Thread tracks use a ScopeTree so we don't need to create one TimerChain per depth.
+  // Allocate a single TimerChain into which all timers will.
```
dimitry-, line 439: will go?

dimitry- (approved):
LGTM

PR #2320, opened 2021-05-14
Capture loading optimizations

pierricgimmig:
Don't render time-graph when loading a capture to avoid contention
between the main thread and the loading thread. Also, move the ScopeTree
creation at the end of capture loading in prevision of parallelization.

### DIFF
--- a/src/OrbitGl/App.cpp
+++ b/src/OrbitGl/App.cpp
@@ -285,7 +285,7 @@ void OrbitApp::OnCaptureStarted(const CaptureStarted& capture_started,
 }
 
 Future<void> OrbitApp::OnCaptureComplete() {
-  for (auto thread_track : GetTimeGraph()->GetTrackManager()->GetThreadTracks()) {
+  for (auto thread_track : GetMutableTimeGraph()->GetTrackManager()->GetThreadTracks()) {
     thread_track->OnCaptureComplete();
   }
 
--- a/src/OrbitGl/App.cpp
+++ b/src/OrbitGl/App.cpp
@@ -65,6 +65,7 @@
 #include "OrbitBase/Result.h"
 #include "OrbitBase/ThreadConstants.h"
 #include "OrbitBase/Tracing.h"
+#include "OrbitBase/UniqueResource.h"
 #include "Path.h"
 #include "PresetsDataView.h"
 #include "SamplingReport.h"
@@ -260,6 +261,8 @@ void OrbitApp::OnCaptureStarted(const CaptureStarted& capture_started,
         capture_data_ = std::make_unique<CaptureData>(module_manager_.get(), capture_started,
                                                       std::move(frame_track_function_ids));
         capture_window_->CreateTimeGraph(capture_data_.get());
+        TrackManager* track_manager = GetMutableTimeGraph()->GetTrackManager();
+        track_manager->SetIsDataFromSavedCapture(is_loading_capture_);
 
         frame_track_online_processor_ =
             orbit_gl::FrameTrackOnlineProcessor(GetCaptureData(), GetMutableTimeGraph());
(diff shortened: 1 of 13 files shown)


src/OrbitGl/App.cpp
```diff
@@ -2191,8 +2202,10 @@ bool OrbitApp::IsCapturing() const {
   return capture_client_ != nullptr && capture_client_->IsCapturing();
 }
 
+bool OrbitApp::IsLoadingCapture() const { return is_loading_capture_; }
+
 bool OrbitApp::IsCapturingOrLoading() const {
-  return is_capture_loading_ || (capture_client_ != nullptr && capture_client_->IsCapturing());
+  return is_loading_capture_ || (capture_client_ != nullptr && capture_client_->IsCapturing());
```
dimitry-, line 2208: do we need this here? I added it for interactivity which is now disabled. I think we can drop IsCapturingOrLoading and make IsCapturing public

src/OrbitGl/TimeGraph.h
```diff
@@ -68,7 +68,7 @@ class TimeGraph : public orbit_gl::CaptureViewElement {
   [[nodiscard]] const orbit_client_model::CaptureData* GetCaptureData() const {
     return capture_data_;
   }
-  [[nodiscard]] TrackManager* GetTrackManager() { return track_manager_.get(); }
+  [[nodiscard]] TrackManager* GetTrackManager() const { return track_manager_.get(); }
```
dimitry-, line 71: it should probably return const TrackManager*

src/OrbitGl/TimeGraph.cpp
```diff
@@ -682,6 +682,9 @@ void TimeGraph::Draw(Batcher& batcher, TextRenderer& text_renderer, uint64_t cur
                      PickingMode picking_mode, float z_offset) {
   ORBIT_SCOPE("TimeGraph::Draw");
 
+  // Don't draw while loading to avoid contention with the loading thread.
+  if (app_->IsLoadingCapture()) return;
```
karupayun, line 686: I rather would prefer app_ not calling draw from TimeGraph to avoid this dependence. TimeGraph shouldn't know about if a capture is being loaded.

src/OrbitGl/ThreadTrack.h
```diff
@@ -57,6 +58,8 @@ class ThreadTrack final : public TimerTrack {
 
   [[nodiscard]] std::vector<CaptureViewElement*> GetVisibleChildren() override;
 
+  void OnCaptureComplete();
```
karupayun, line 61: I would receive the flag scope_tree_update_type_  here instead of receiving it in the constructor.

karupayun (commented):
Nice optimizations Pierric! I've really enjoyed reading your doc. I wrote a few comments because I don't want to generate another dependence from TimeGraph to App.

src/OrbitGl/ThreadTrack.h
pierricgimmig: The flag is needed earlier so that we can avoid constructing the ScopeTree as timers arrive in the case of capture loading.

pierricgimmig: See ThreadTrack::OnTimer

pierricgimmig: Hmm, I might have missed a file when splitting up my PR... :-) Pushing now...

pierricgimmig: It's in now, PTAL, that was an important piece...

dimitry- (approved):

src/OrbitGl/ThreadTrack.h
karupayun: ahhh, nice to see. BTW, I'm still not sure. I will re-read again on Monday if you don't merge it today.

src/OrbitGl/App.cpp
pierricgimmig: Done.

src/OrbitGl/TimeGraph.h
pierricgimmig: I reverted the change actually.

src/OrbitGl/TimeGraph.cpp
pierricgimmig: I totally agree, I moved it up to CaptureWindow.

PR #2325, opened 2021-05-17
Make OpenDisassembly a main window member function

beckerhe:
This is mainly moving code around in preparation of the combined source code view.

No functional changes in here.

### DIFF
--- a/src/OrbitGl/App.cpp
+++ b/src/OrbitGl/App.cpp
@@ -1305,8 +1305,7 @@ bool OrbitApp::IsDevMode() const { return absl::GetFlag(FLAGS_devmode); }
 void OrbitApp::SendDisassemblyToUi(std::string disassembly, DisassemblyReport report) {
   main_thread_executor_->Schedule(
       [this, disassembly = std::move(disassembly), report = std::move(report)]() mutable {
-        CHECK(disassembly_callback_);
-        disassembly_callback_(std::move(disassembly), std::move(report));
+        main_window_->ShowDisassembly(disassembly, std::move(report));
       });
 }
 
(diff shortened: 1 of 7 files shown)


src/OrbitGl/App.cpp
```diff
@@ -1305,8 +1305,7 @@ bool OrbitApp::IsDevMode() const { return absl::GetFlag(FLAGS_devmode); }
 void OrbitApp::SendDisassemblyToUi(std::string disassembly, DisassemblyReport report) {
```
karupayun, line 1305: Should we also rename this method?

karupayun (approved):
LGTM!

src/OrbitGl/App.cpp
beckerhe: Can you explain why? How would you name that method?

karupayun: Because only seeing the naming, before we were moving the disassembly and now we are showing it? I don't understand the relationship between SendDisassemblyToUi and ShowDisassembly

beckerhe: There are multiple functions called `Send*ToUi`, `SendTooltipToUi` for examples. They all have in common that you can call them from any thread - they are thread safe.

Changing the name of `SendDisassemblyToUi` to something else would mean it's not clear anymore that this function belongs to this group of functions which are thread-safe.

I renamed `Disassembly` to `ShowDisassembly` to make the name similar to `ShowSourceCode`.

karupayun: Thanks for explaining it! LGTM

PR #2326, opened 2021-05-17
Add missing std::clamp

beckerhe:
Qt throws some error messages if alpha is not in between 0 and 256.

### DIFF
--- a/src/CodeViewer/Viewer.cpp
+++ b/src/CodeViewer/Viewer.cpp
@@ -208,10 +208,16 @@ void Viewer::DrawLineNumbers(QPaintEvent* event) {
       const uint32_t num_samples_in_line =
           code_report_->GetNumSamplesAtLine(line_number).value_or(0);
       const uint32_t num_samples_in_function = code_report_->GetNumSamplesInFunction();
-      const float intensity = std::clamp(
-          static_cast<float>(num_samples_in_line) / static_cast<float>(num_samples_in_function),
-          0.0f, 1.0f);
-      const auto scaled_intensity = static_cast<int>(std::sqrt(intensity) * 255);
+
+      const int scaled_intensity = [&] {
+        if (num_samples_in_function == 0) return 0;
+
+        const float intensity = std::clamp(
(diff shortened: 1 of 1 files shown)


src/CodeViewer/Viewer.cpp
```diff
@@ -211,7 +211,8 @@ void Viewer::DrawLineNumbers(QPaintEvent* event) {
       const float intensity = std::clamp(
           static_cast<float>(num_samples_in_line) / static_cast<float>(num_samples_in_function),
           0.0f, 1.0f);
-      const auto scaled_intensity = static_cast<int>(std::sqrt(intensity) * 255);
+      const auto scaled_intensity =
+          std::clamp(static_cast<int>(std::sqrt(intensity) * 255), 0, 255);
```
karupayun, line 215: I don't understand the cause of this.

Is because static_cast <int> (0.0) is answering -1? Should we use round instead of static_cast?

beckerhe: `static_cast<float>(num_samples_in_line) / static_cast<float>(num_samples_in_function)` is `NaN` when `num_samples_in_function` is 0. Then `intensity` gets `NaN` and `std::sqrt(intensity)` also returns `NaN`. The `static_cast<int>` invokes UB.

While thinking about it, maybe we should fix it differently.

karupayun: isn't the intensity clamped from 0.0 to 1.0?

beckerhe: Doesn't matter. `clamp(Nan)` is still NaN.

beckerhe: I pushed a revised version. PTAL!

karupayun (approved):
LGTM!

src/CodeViewer/Viewer.cpp
karupayun: Yes, of course. thanks for explaining it.

beckerhe: Sure. Thanks for asking!

PR #2327, opened 2021-05-17
Rename ElfUtils to ObjectUtils for more general object file support

ronaldfw:
This is a first step in adding support for different types of object
files such as Common Object File Format (COFF) in the future. All ELF
specific naming is generalized (ObjectUtils, ObjectFile, LoadSymbols)
or kept specifically in ElfFile (derived from ObjectFile). There are
still some methods in ElfFile that we probably want at the general
ObjectFile level (for example, to get line information), but for now
I've kept the changes to what is needed to load symbols only.

Key changes:
- Add ObjectFile base class with common methods.
- Renaming.

Bug: http://b/187665628
Tested: Unit tests; manually tested loading symbols with various .so
files.


src/ObjectUtils/ElfFile.cpp
```diff
@@ -54,16 +54,16 @@ class ElfFileImpl : public ElfFile {
   ElfFileImpl(std::filesystem::path file_path,
               llvm::object::OwningBinary<llvm::object::ObjectFile>&& owning_binary);
 
-  [[nodiscard]] ErrorMessageOr<ModuleSymbols> LoadSymbolsFromSymtab() override;
+  [[nodiscard]] ErrorMessageOr<ModuleSymbols> LoadSymbols() override;
```
dimitry-, line 57: I think this becomes too vague.. maybe LoadDebugSymbols, and add a comment that for the elf-file it is .symtab section.

src/ObjectUtils/ElfFile.cpp
```diff
@@ -54,16 +54,16 @@ class ElfFileImpl : public ElfFile {
...
   [[nodiscard]] ErrorMessageOr<ModuleSymbols> LoadSymbolsFromDynsym() override;
   [[nodiscard]] ErrorMessageOr<uint64_t> GetLoadBias() const override;
-  [[nodiscard]] bool HasSymtab() const override;
+  [[nodiscard]] bool HasSymbols() const override;
```
dimitry-, line 60: same here.

src/ObjectUtils/ElfFile.cpp
```diff
@@ -54,16 +54,16 @@ class ElfFileImpl : public ElfFile {
...
   [[nodiscard]] bool HasDynsym() const override;
   [[nodiscard]] bool HasDebugInfo() const override;
   [[nodiscard]] bool HasGnuDebuglink() const override;
   [[nodiscard]] bool Is64Bit() const override;
   [[nodiscard]] std::string GetBuildId() const override;
-  [[nodiscard]] std::string GetSoname() const override;
+  [[nodiscard]] std::string GetPrettyName() const override;
```
dimitry-, line 66: soname is more like ID. I do not think pretty_name fits here. Also is there windows counterpart for this?

src/ObjectUtils/ElfFile.cpp
ronaldfw: I had this first as LoadSymbolsFromSymbolTable(), which also exists for COFF files. WDYT?

src/ObjectUtils/ElfFile.cpp
ronaldfw: I was going by how this is used. We set the "name" field in ModuleInfo to either what we get as soname, or as the filename. This is what is displayed in the UI. (Not sure if it's used elsewhere, but it seems to me that once that name is in ModuleInfo, all special meaning of it has been lost anyway.)

My suggestion would be to keep GetPrettyName(), as it's intended for UI usage (unless I'm missing something). A basic implementation would be just using the filename. 

If we need soname for ELF files with the special identifier meaning, I would suggest to add that as an extra method in ElfFile.

src/ObjectUtils/ElfFile.cpp
dimitry-: seems like it has the same semantics - it is a debug symbols table.

Which is deprecated if I read this correctly:
https://docs.microsoft.com/en-us/windows/win32/debug/pe-format#coff-file-header-object-and-image
"The file offset of the COFF symbol table, or zero if no COFF symbol table is present. This value should be zero for an image because COFF debugging information is deprecated."

dimitry-: and dynamic symbols (or exported symbols) are coming from edata section.

src/ObjectUtils/ElfFile.cpp
dimitry-: It is possible soname should be a property of an elf-file.

dimitry-: it is also used by dynamic instrumentation to correctly find libc/libdl

src/ObjectUtils/ElfFile.cpp
ronaldfw: Ok, I'll go with LoadDebugSymbols to keep other options open (such as reading them from DWARF).

src/ObjectUtils/ElfFile.cpp
ronaldfw: I see, it seems we don't distinguish between soname and name of the module, at least in the user space instrumentation code. The code seems to expect that ModuleInfo::name is the soname of the module.

So there is some ambiguity in ModuleInfo, as it's not clear that "name" should be soname when a soname is present. 

Is the code in the user space instrumentation correct if the name field was filled with the filename and not the soname? Or asked differently, if an ELF file doesn't have an soname, is the filename the correct choice then?

Not sure what the best option here is: Should we just call the method GetName() (which is then using the same name as ModuleInfo)? Should ModuleInfo get a soname field? 

My suggestion:
- GetName() in base class (filename, if no better choice).
- ModuleInfo gets added soname field (it already has other ELF specific fields)
- GetSoname() in ElfFile
- ElfFile implements GetName() as GetSoname() if present, o/w as filename.
- user space instrumentation code is changed to read soname(), not name(). 

WDYT?

dimitry-: sgtm!

src/ObjectUtils/include/ObjectUtils/ObjectFile.h
```diff
@@ -0,0 +1,43 @@
...
+  [[nodiscard]] virtual std::string GetPrettyName() const = 0;
+  [[nodiscard]] virtual std::filesystem::path GetFilePath() const = 0;
+
+  [[nodiscard]] static ErrorMessageOr<std::unique_ptr<ElfFile>> CreateElfFile(
+      const std::filesystem::path& file_path);
+  [[nodiscard]] static ErrorMessageOr<std::unique_ptr<ElfFile>> CreateElfFile(
+      const std::filesystem::path& file_path,
+      llvm::object::OwningBinary<llvm::object::ObjectFile>&& file);
+  [[nodiscard]] static ErrorMessageOr<std::unique_ptr<ElfFile>> CreateElfFileFromBuffer(
+      const std::filesystem::path& file_path, const void* buf, size_t len);
```
beckerhe, line 38: I think these should be free functions and declared in `ElfFile.h`. This would also avoid the forward declaration of `ElfFile`.

beckerhe (approved):
Otherwise LGTM. Thanks!

src/ObjectUtils/ElfFile.cpp
danielfenner: I don't get why `LoadSymbolsFromSymtab`, `HasSymtab` are renamed in the first place. The names are elf specific, but this is ElfFile.cpp so that is fine, right? Or do you intend to move them later into ObjectFile?

ronaldfw: Both are already in ObjectFile and are needed for core functionality that we need across both ELF and COFF. 

The idea is that LoadDebugSymbols is the method used by the UI to get symbols for a module if the symbols are embedded in the module (either as a symbol table or as DWARF). We currently use symtab for ELF files, but IMO this is already an implementation detail and we could just as well use DWARF if a symbol table is not present for whatever reason. 

My understanding is that LoadSymbolsFromDynsym is only used to inject code into the target process by finding the right functions in ld and/or libc (so we don't need to generalize this, yet). Correct?

src/ObjectUtils/ElfFile.cpp
ronaldfw: Done.

src/ObjectUtils/ElfFile.cpp
dimitry-: it is also used to normalize callstacks in the sampling report.

dimitry-: We were also considering using Dynsym as a fallback when Symtab is not available.

ronaldfw: Yes, that would make sense to me. 

Side-note: I think the same can be done for PE/COFF (by looking at the export directory), but some care must be taken about forwarded symbols: A dll can forward a symbol to a different dll and not actually contain a definition for the symbol. It's not quite clear how we want to treat this in the UI, but a simple initial solution would be to just exclude forwarded symbols.

I'll take a note to look at adding this as a fallback, but will not make any additional changes in the current PR.

dimitry- (approved):

src/ObjectUtils/ElfFile.cpp
ronaldfw: @danielfenner I'll resolve this, but we can discuss if there are further concerns or ideas on how we should structure the code. I'll keep working on this part of the code.

PR #2328, opened 2021-05-17
Introduce module CodeReport

beckerhe:
This module contains the abstract base class CodeReport and its implementations DisassemblyReport and SourceCodeReport, as well as the class Disassembler which is a dependency of DisassemblyReport.

All these classes used to live in OrbitGl. The PR also adds a first simple test for Disassembler.


florian-kuebler (commented):
LGTM, thanks!

florian-kuebler (approved):

src/CodeReport/CMakeLists.txt
```diff
@@ -5,9 +5,13 @@
 cmake_minimum_required(VERSION 3.15)
 
 project(CodeReport)
-add_library(CodeReport INTERFACE)
+add_library(CodeReport OBJECT)
```
dpallotti, line 8: It's the first time I see `OBJECT` used in Orbit. Do we prefer this to `STATIC` or what's the reason?

src/CodeReport/include/CodeReport/DisassemblyReport.h
```diff
@@ -39,4 +40,5 @@ class DisassemblyReport : public CodeReport {
   uint32_t samples_count_;
 };
 
-#endif  // ORBIT_GL_DISASSEMBLY_REPORT_H_
+}  // namespace orbit_code_report
+#endif  // CODE_REPORT_DISASSEMBLY_REPORT_H_
```
dpallotti, line 44:
```suggestion
}  // namespace orbit_code_report

#endif  // CODE_REPORT_DISASSEMBLY_REPORT_H_
```

dpallotti (approved):

src/CodeReport/CMakeLists.txt
beckerhe: Yes, we prefer OBJECT :slightly_smiling_face:  An OBJECT library is just a bunch of .o-files, while a static library is always linked into a .a/.lib-file. Since we don't ship the the static libraries, it's just a waste of time and disk space to generate them.

At some point I will move everything to OBJECT targets.

src/CodeReport/include/CodeReport/DisassemblyReport.h
beckerhe: Done. Thx.

src/CodeReport/CMakeLists.txt
dpallotti: Hehe I was always wondering whether all those generated static libraries were really necessary. Thanks!

beckerhe: Feel free to try out OBJECT next time. :smile:

PR #2330, opened 2021-05-17
[DRAFT] Highlight Block Cpp

karupayun:
A PR made to receive comments from Henning.

First step of http://b/187388546.

Basically:
 1- All QRegularExpressions became public and static.
 2- HighlightBlock now returns the new state (because we aren't able to use that from QSyntaxHighlighter class).

The function below is how it was before.

### DIFF
--- a/src/SyntaxHighlighter/Cpp.cpp
+++ b/src/SyntaxHighlighter/Cpp.cpp
@@ -97,6 +97,62 @@ Cpp::Cpp() : QSyntaxHighlighter{static_cast<QObject*>(nullptr)} {
   class_member_regex_ = QRegularExpression{CppRegex::kClassMemberRegex};
 }
 
+CppHighlighterState HighlightBlockCpp(
+    const QString& code, CppHighlighterState previous_block_state,
+    std::function<void(int, int, const QTextCharFormat&)> set_format) {
+  CppHighlighterState next_block_state = CppHighlighterState::kInitialState;
+  const auto apply = [&code, &set_format, &next_block_state](
+                         QRegularExpression* expression, const QColor& color,
+                         CppHighlighterState new_state = CppHighlighterState::kInitialState) {
+    QTextCharFormat format{};
+    format.setForeground(color);
+
+    for (auto it = expression->globalMatch(code); it.hasNext();) {
+      const auto match = it.next();
+      set_format(match.capturedStart(), match.capturedLength(), format);
+      next_block_state = new_state;
+    }
+  };
+
+  // We are processing line by line and trying to find all substrings that match with these
+  // patterns. As each one paints over the others, order matters.
+
+  // Ordered heuristics for painting certain word patterns. Should be at the first.
+  apply(&Cpp::capitalized_regex_, kGreyViolet);
+  apply(&Cpp::namespace_variables_regex_, kGreyViolet);
+  apply(&Cpp::function_definition_regex_, kYellowOrange);
+  apply(&Cpp::namespace_regex_, kGreyViolet);
+  apply(&Cpp::only_uppercase_regex_, kOlive);
+
+  // Extra patterns which make the syntax highlighter nicer. Order doesn't matter.
+  apply(&Cpp::class_name_regex_, kGreyViolet);
+  apply(&Cpp::number_regex_, kBlue);
+  apply(&Cpp::class_member_regex_, kViolet);
+  apply(&Cpp::comma_regex_, kOrange);
+
+  // Special C/C++ patterns. Order doesn't matter.
+  apply(&Cpp::keyword_regex_, kOrange);
+  apply(&Cpp::constant_regex_, kOlive);
+  apply(&Cpp::include_file_regex_, kGreen);
+  apply(&Cpp::preprocessor_regex_, kYellow);
+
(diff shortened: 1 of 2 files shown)


src/SyntaxHighlighter/Cpp.cpp
```diff
@@ -97,6 +97,62 @@ Cpp::Cpp() : QSyntaxHighlighter{static_cast<QObject*>(nullptr)} {
...
+    apply(&Cpp::no_end_string_regex_, kGreen, CppHighlighterState::kOpenStringState);
+    apply(&Cpp::end_string_regex_, kGreen);
+  }
+  if (previous_block_state == CppHighlighterState::kOpenCommentState) {
+    apply(&Cpp::no_end_comment_regex_, kGrey, CppHighlighterState::kOpenCommentState);
+    apply(&Cpp::end_comment_regex_, kGrey);
+  }
+  apply(&Cpp::open_string_regex_, kGreen, CppHighlighterState::kOpenStringState);
+  apply(&Cpp::open_comment_regex_, kGrey, CppHighlighterState::kOpenCommentState);
+}
```
beckerhe, line 154: a `return next_block_state` is missing here, right?

karupayun: Yep!

src/SyntaxHighlighter/include/SyntaxHighlighter/Cpp.h
```diff
@@ -22,28 +22,27 @@ class Cpp : public QSyntaxHighlighter {
...
+  static QRegularExpression open_string_regex_;
+  static QRegularExpression end_string_regex_;
+  static QRegularExpression no_end_string_regex_;
+  static QRegularExpression comma_regex_;
+  static QRegularExpression only_uppercase_regex_;
+  static QRegularExpression capitalized_regex_;
+  static QRegularExpression namespace_regex_;
+  static QRegularExpression namespace_variables_regex_;
+  static QRegularExpression class_name_regex_;
+  static QRegularExpression class_member_regex_;
```
beckerhe, line 45: Do they need to be declared in a header file at all? Can't they just be constants with internal linkage in the cpp file?

src/SyntaxHighlighter/Cpp.cpp
```diff
@@ -97,6 +97,62 @@ Cpp::Cpp() : QSyntaxHighlighter{static_cast<QObject*>(nullptr)} {
   class_member_regex_ = QRegularExpression{CppRegex::kClassMemberRegex};
 }
 
+CppHighlighterState HighlightBlockCpp(
```
beckerhe, line 100: This function needs also be declared in the header file. Otherwise you can't call it from somewhere else.

src/SyntaxHighlighter/Cpp.cpp
```diff
@@ -97,6 +97,62 @@ Cpp::Cpp() : QSyntaxHighlighter{static_cast<QObject*>(nullptr)} {
...
+  }
+  if (previous_block_state == CppHighlighterState::kOpenCommentState) {
+    apply(&Cpp::no_end_comment_regex_, kGrey, CppHighlighterState::kOpenCommentState);
+    apply(&Cpp::end_comment_regex_, kGrey);
+  }
+  apply(&Cpp::open_string_regex_, kGreen, CppHighlighterState::kOpenStringState);
+  apply(&Cpp::open_comment_regex_, kGrey, CppHighlighterState::kOpenCommentState);
+}
+
 void Cpp::highlightBlock(const QString& code) {
```
beckerhe, line 156: Cpp::highlightBlock should used `CppHighlightBlock` to do its formatting. But I guess you already knew that.

src/SyntaxHighlighter/include/SyntaxHighlighter/Cpp.h
karupayun: Similar to the regex in the .cpp file?

src/SyntaxHighlighter/Cpp.cpp
karupayun: Yes, of course. I only wanted to receive comments about how to implement it, sorry for that.

src/SyntaxHighlighter/Cpp.cpp
karupayun: Actually I was going to erase this function and always use the free functions instead. WDYT?

beckerhe: You still need to have this function because it overrides a virtual function from the base class and that's how you customize the behaviour. But all the formatting should go into the free function, yes. This function should really only do whatever is necessary to call the free function.

src/SyntaxHighlighter/include/SyntaxHighlighter/Cpp.h
beckerhe: Yes, just like the regex string literals.

src/SyntaxHighlighter/Cpp.cpp
karupayun: I hadn't had realized that, thanks for saving a future half-an-hour :)

PR #2331, opened 2021-05-17
Extend DisassemblyReport

beckerhe:
This is doing two things which are needed for the combined source code view:

1. Extend `Disassembler` and `DisassemblyReport` to allow address to line number lookups. A hash map is used under the hood. I tried to implement binary search on top of the existing line number to address mapping but that doesn't work well since lines might not map to any address.
2. Add `FunctionInfo` as a parameter to `ShowDisassembly`. This will also be needed for the combined source code view in a later step.


antonrohr (approved):
Thank you for separating into small commits, review was a breeze. LGTM!

PR #2335, opened 2021-05-18
Add thread states to CapturreMetric

antonrohr:
This updates orbit_log_event.proto from the source of truth and adds information about thread states to CaptureMetrics. 
Whether thread states are enabled or not is known when the user starts a capture

http://b/187914133

Simple Test: run unit tests. 
Extensive Test: download newest metrics_uploader_client.dll, start capture, check sherlog

### DIFF
--- a/src/MetricsUploader/CaptureMetric.cpp
+++ b/src/MetricsUploader/CaptureMetric.cpp
@@ -23,6 +23,7 @@ CaptureMetric::CaptureMetric(MetricsUploader* uploader, const CaptureStartData& 
   capture_data_.set_number_of_manual_stop_async_timers(
       start_data.number_of_manual_stop_async_timers);
   capture_data_.set_number_of_manual_tracked_values(start_data.number_of_manual_tracked_values);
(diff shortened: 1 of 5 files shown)


florian-kuebler (approved):
LGTM, thanks!

ronaldfw:
Could you add some more info to the PR title and PR description, including a bug link? (Unless you have good reasons for not including details.) Ideally, the commit messages should also reflect the same details (not just "what" you are doing, but also "why"; the "why" can often be covered with a bug link). Thanks!

antonrohr:
> Could you add some more info to the PR title and PR description, including a bug link? (Unless you have good reasons for not including details.) Ideally, the commit messages should also reflect the same details (not just "what" you are doing, but also "why"; the "why" can often be covered with a bug link). Thanks!

You are right, this was sloppy on my side. I added the bug links and updated title and description.

ronaldfw:
> You are right, this was sloppy on my side. I added the bug links and updated title and description.

Looks great now, thank you!

PR #2336, opened 2021-05-18
Make Scheduler and Gpu Timers appear again

karupayun:
In https://github.com/google/orbit/pull/2319 we switched to emplace_back but in the process we forgot to add the text_box to the TimerChain.

Solved https://b/188375720.

Test: Start/stop a capture. Load an old one.

### DIFF
--- a/src/OrbitGl/TimerTrack.cpp
+++ b/src/OrbitGl/TimerTrack.cpp
@@ -331,6 +331,7 @@ void TimerTrack::OnTimer(const TimerInfo& timer_info) {
 
   TextBox& text_box = timer_chain->emplace_back();
   text_box.SetTimerInfo(timer_info);
+  timer_chain->emplace_back(text_box);
 
   ++num_timers_;
   if (timer_info.start() < min_time_) min_time_ = timer_info.start();


ronaldfw:
Thanks for the fix!

Does this change anything about the performance conclusions from the original change?

src/OrbitGl/TimerTrack.cpp
```diff
@@ -331,6 +331,7 @@ void TimerTrack::OnTimer(const TimerInfo& timer_info) {
 
   TextBox& text_box = timer_chain->emplace_back();
   text_box.SetTimerInfo(timer_info);
+  timer_chain->emplace_back(text_box);
```
beckerhe, line 334: That doesn't make sense to me. The text_box is added two lines above: `TextBox& text_box = timer_chain->emplace_back();`

Now we are adding it twice?

beckerhe: Maybe emplace_back is not working as it should?

dpallotti: Also there is code similar to this in `GraphTrack` and `ThreadTrack`. Does this also fix that? I would assume not.

beckerhe: My guess is that is has something to do with `min_timestamp_` and `max_timestamp_` in `TimerBlock`. They are only updated when `emplace_back` is called, but not when `text_box.SetTimerInfo` is called.

karupayun: Yes, you are right. I only realized that was working and didn't analize the code.

dpallotti: > My guess is that is has something to do with min_timestamp_ and max_timestamp_

Yes.

karupayun: Nono, the fix is working, the question is why.

karupayun: emplace_back is not returning void?

text_box = timer_chain-&gt;emplace_back();

karupayun: Ignore my previous comment. I don't understand why the timers are appearing.

beckerhe: It behaves like emplace_black in the standard containers, so yes, it makes sense.

But this doesn't work with preserving invariance in TimerChain

beckerhe: I'm not sure how to fix this. Of course a quick fix would/should work:

```
TextBox text_box{};
text_box.SetTimerInfo(timer_info);
timer_chain->emplace_back(std::move(text_box));
```

But the underlying problem persists.

dpallotti: Doesn't that kind of defeat the performance optimization that the `emplace_back` change was supposed to introduce?

A short term fix is to remove `SetTimerInfo` and use a constructor parameter.

karupayun:
Closing this PR. @dpallotti  is working on a better solution.

PR #2338, opened 2021-05-18
Add basic implementation of COFF object files

ronaldfw:
This adds a basic implementation of COFF object file support for
loading symbols from COFF files using the symbol table. A short unit
test is added.

Bug: http://b/187665628
Tested: Unit test (code is not yet integrated).


src/ObjectUtils/include/ObjectUtils/CoffFile.h
```diff
@@ -0,0 +1,33 @@
...
+
+#include "ObjectUtils/ObjectFile.h"
+#include "OrbitBase/Result.h"
+#include "llvm/Object/Binary.h"
+#include "llvm/Object/ObjectFile.h"
+#include "symbol.pb.h"
+
+namespace orbit_object_utils {
+
+class CoffFile : public ObjectFile {
```
beckerhe, line 19: I guess you opted for having `CoffFile` and `CoffFileImpl` for symmetry reasons with `ElfFile`?

I personally find it simpler to put the implementation in `CoffFile` directly, but I also don't feel strongly about that.

src/ObjectUtils/CoffFile.cpp
```diff
@@ -0,0 +1,155 @@
...
+CoffFileImpl::CoffFileImpl(std::filesystem::path file_path,
+                           llvm::object::OwningBinary<llvm::object::ObjectFile>&& owning_binary)
+    : file_path_(std::move(file_path)),
+      owning_binary_(std::move(owning_binary)),
+      has_symbol_table_(false) {
+  object_file_ = llvm::dyn_cast<llvm::object::COFFObjectFile>(owning_binary_.getBinary());
+  Initialize();
+}
+
+void CoffFileImpl::Initialize() {
```
beckerhe, line 51: Are you planning on putting more stuff into this function? If not, I'd opt for putting this one line into the constructor.

beckerhe (approved):
LGTM

src/ObjectUtils/CoffFile.cpp
ronaldfw: Yeah, this kind of grew out of what we have in ElfFile, but the separate method there actually has a clear meaning and it makes sense to separate it out. Here it doesn't really make sense for now. I moved the line into the constructor as you suggested.

src/ObjectUtils/include/ObjectUtils/CoffFile.h
ronaldfw: Yes, it's purely for symmetry. I don't feel strongly either way, so I'll keep it as-is for now.

src/ObjectUtils/CoffFile.cpp
```diff
@@ -0,0 +1,150 @@
...
+using orbit_grpc_protos::SymbolInfo;
+
+class CoffFileImpl : public CoffFile {
+ public:
+  CoffFileImpl(std::filesystem::path file_path,
+               llvm::object::OwningBinary<llvm::object::ObjectFile>&& owning_binary);
+  [[nodiscard]] ErrorMessageOr<orbit_grpc_protos::ModuleSymbols> LoadDebugSymbols() override;
+  [[nodiscard]] bool HasDebugSymbols() const override;
+  [[nodiscard]] std::string GetName() const override;
+  [[nodiscard]] std::filesystem::path GetFilePath() const override;
```
dimitry-, line 31: should this be a const-ref?

src/ObjectUtils/CoffFile.cpp
```diff
@@ -0,0 +1,150 @@
...
+                                        name, file_path_.string()));
+  }
+  // Limit list of symbols to functions. Ignore sections and variables.
+  if (symbol_ref.getType().get() != llvm::object::SymbolRef::ST_Function) {
+    return ErrorMessage("Symbol is not a function.");
+  }
+
+  // Symbol addresses are relative virtual addresses (RVAs) that are relative to the
+  // start of the section. To get the address relative to the start of the object
+  // file, we need to find the section and add the base address of the section.
```
dimitry-, line 71: extract to a separate function? Also it is kind of counterintuitive: are there multiple sections with symbols? I was under impression there is one.

ronaldfw: I don't know if it happens in practice, but the specifications mentions multiple text sections. 

For example, this text part mentions multiple sections: 
"SizeOfCode Size of the code (text) section, or the sum of all code sections if there are multiple sections." (**https://courses.cs.washington.edu/courses/cse378/03wi/lectures/LinkerFiles/coff.pdf**)

Looking up the section is also not expensive, it's all just indices returned and some index arithmetic (it's not iterating over the sections to find the section, for example).

I will pull this out into a separate function, though.

src/ObjectUtils/CoffFile.cpp
ronaldfw: I think it could be. That would mean this has to be purely a getter of the member, right? This is one of the methods that was already there and I took the existing signature, so not sure if there was a deeper reason why this is not const&. Also, our codebase seems to not do this consistently throughout.

src/ObjectUtils/CoffFile.cpp
```diff
@@ -48,6 +49,21 @@ CoffFileImpl::CoffFileImpl(std::filesystem::path file_path,
...
+    const llvm::object::SymbolRef& symbol_ref) {
+  // Symbol addresses are relative virtual addresses (RVAs) that are relative to the
+  // start of the section. To get the address relative to the start of the object
+  // file, we need to find the section and add the base address of the section.
+  llvm::object::COFFSymbolRef coff_symbol_ref = object_file_->getCOFFSymbol(symbol_ref);
+  int32_t section_id = coff_symbol_ref.getSectionNumber();
+  const llvm::object::coff_section* section = nullptr;
+  std::error_code err = object_file_->getSection(section_id, section);
+  if (err) {
+    return ErrorMessage("Section of symbol not found.");
```
dimitry-, line 62: add/return err.message()?

src/ObjectUtils/CoffFile.cpp
ronaldfw: Done.

src/ObjectUtils/CoffFile.cpp
dimitry-: I am not sure why it is returned by value. Probably because in this case this is an interface and we cannot speculate about implementation. If it stores the value or generates it based on something else. My personal preference here would be to use const-ref until proven otherwise.

src/ObjectUtils/CoffFile.cpp
ronaldfw: Done.

src/ObjectUtils/CoffFile.cpp
ronaldfw: Done.

PR #2339, opened 2021-05-18
Guard all TrackManager methods which access tracks

karupayun:
After a new crash https://b/188521605, we are locking all methods which access all_tracks and visible_tracks to avoid having race conditions.

This should be a temporal solution until we move ProcessTimer + OnTimer functions outside OrbitGl and TimeGraph and therefore we assure the UI is only accessed by the main thread.

We are solving http://b/177200020, but I'm not sure about the performance regression that we can have from this change. I will make some experiments.


dpallotti:
As discussed, as this doesn't solve the issue I would say either lock every method, or investigate which specific methods should be locked and which shouldn't.

PR #2340, opened 2021-05-18
Remove `TextBox::SetTimerInfo`, pass to the constructor instead

dpallotti:
#### Remove `TextBox::SetTimerInfo`, pass to the constructor instead
`TimerChain` needs the added `TextBox` to already have the `TimerInfo` set, to
correctly update the `min_timestamp_` and `max_timestamp` of a `TimerBlock`.
Also addressing some lint warnings as there's too much yellow in my IDE.

Bug: http://b/188375720

Test: Capture Trata with Vulkan layer and instrumented functions: notice that
scheduler and GPU tracks are now filled. Also enable a frame track and notice
that it also now works, although sometimes affected by the unrelated crash
http://b/188521605.
#### Delete copy- and move- assignment of `TextBox`
As discussed, so that elements of `TimerChain` cannot be reassigned.

Bug: http://b/188375720

### DIFF
--- a/src/OrbitGl/CaptureStatsTest.cpp
+++ b/src/OrbitGl/CaptureStatsTest.cpp
@@ -4,25 +4,12 @@
 
 #include <gtest/gtest.h>
 
+#include <list>
+
 #include "CaptureStats.h"
 #include "SchedulerTrack.h"
 #include "SchedulingStats.h"
 #include "TextBox.h"
-
-TextBox* CreateScope(int32_t pid, int32_t tid, int32_t cpu, uint64_t start, uint64_t end) {
-  const size_t kNumScopes = 1024;
-  static TextBox scope_buffer[kNumScopes];
-  static size_t index = 0;
-  EXPECT_TRUE(index < kNumScopes - 2);
-  TextBox& scope = scope_buffer[index++];
-  orbit_client_protos::TimerInfo& timer_info = scope.GetMutableTimerInfo();
-  timer_info.set_start(start);
-  timer_info.set_end(end);
-  timer_info.set_thread_id(tid);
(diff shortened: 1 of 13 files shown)


beckerhe (approved):
LGTM. Thanks a lot, Davide!

florian-kuebler (approved):
LGTM, Thanks!

src/OrbitGl/CaptureStatsTest.cpp
```diff
@@ -10,18 +10,14 @@
 #include "TextBox.h"
 
 TextBox* CreateScope(int32_t pid, int32_t tid, int32_t cpu, uint64_t start, uint64_t end) {
-  const size_t kNumScopes = 1024;
-  static TextBox scope_buffer[kNumScopes];
-  static size_t index = 0;
-  EXPECT_TRUE(index < kNumScopes - 2);
```
karupayun, line 16: I know that we are inside a test, but we are not having this invariant anymore or simply we are not checking it?

karupayun (approved):
LGTM! Thanks Davide!

src/OrbitGl/CaptureStatsTest.cpp
dpallotti: As far as I understand, this was just a limit posed by using an array, but a vector can just be used instead.

dpallotti: At the same time now a test fails on Windows in this file, so let's see.

dpallotti: Rearranged this code a bit and switched to a list.

pierricgimmig:
Thanks @dpallotti !

PR #2341, opened 2021-05-19
Highlight independent functions

karupayun:
We are creating free functions to Syntax Highlight a block, so that they can be used in any other syntax highlighter class.

Also we are refactoring a little some files constructing directly the QRegularExpressions from strings.

First step of http://b/187388546.


beckerhe (approved):
LGTM

PR #2342, opened 2021-05-19
Use fixed font only for the source code contents

beckerhe:
orbit_code_viewer::Viewer used to show line numbers and sample numbers
in fixed font. With this change, only the actual source code is shown
in fixed font.

This is how it looks like now (zoomed in to make the different fonts more obvious):
image

EDIT: Yes, I know, this is not fixing the out-of-line bug. Line numbers are still a little bit too low.

### DIFF
--- a/src/CodeViewer/Viewer.cpp
+++ b/src/CodeViewer/Viewer.cpp
@@ -106,8 +106,6 @@ Viewer::Viewer(QWidget* parent)
   };
   QObject::connect(this, &QPlainTextEdit::updateRequest, this, update_viewport_area);
 
-  setFont(QFontDatabase::systemFont(QFontDatabase::SystemFont::FixedFont));
-
   constexpr int kTabStopInWhitespaces = 4;
   setTabStopDistance(fontMetrics().horizontalAdvance(' ') * kTabStopInWhitespaces);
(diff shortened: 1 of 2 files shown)


karupayun (approved):
LGTM!

src/CodeViewer/Viewer.cpp
```diff
@@ -126,8 +124,16 @@ void Viewer::resizeEvent(QResizeEvent* ev) {
 }
 
 void Viewer::wheelEvent(QWheelEvent* ev) {
+  auto document_default_font = document()->defaultFont();
```
reichlfl, line 127: Should this (and below) be an explicit type? https://google.github.io/styleguide/cppguide.html#Type_deduction
(Disclaimer: I personally like auto for cases like this, but I was corrected about this in multiple PRs now :D)

reichlfl (approved):

src/CodeViewer/Viewer.cpp
beckerhe: You're probably right. I will fix it. :-)

beckerhe: Done.

PR #2346, opened 2021-05-19
Correct function order in UpdateTrackOrder

karupayun:
Regression from 1.63 (more precisely from https://github.com/google/orbit/pull/2221).

In that PR we alter the order by mistake of the functions "UpdateMovingTrackSorting" and "UpdateVisibleTrackList".

This PR we are reverting that change and renaming UpdateMovingTrackSorting to a name which explicit the fact that we are updating the position in the visible tracks.

It solved the crash http://b/188521605.

TODO: I'm updating my linux machine to test the crash there.

### DIFF
--- a/src/OrbitGl/TrackManager.cpp
+++ b/src/OrbitGl/TrackManager.cpp
@@ -204,7 +204,7 @@ std::vector<ThreadTrack*> TrackManager::GetSortedThreadTracks() {
   return sorted_tracks;
 }
 
-void TrackManager::UpdateMovingTrackSorting() {
+void TrackManager::UpdateMovingTrackPositionInVisibleTracks() {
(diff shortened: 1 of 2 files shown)


src/OrbitGl/TrackManager.cpp
```diff
@@ -204,7 +204,7 @@ std::vector<ThreadTrack*> TrackManager::GetSortedThreadTracks() {
   return sorted_tracks;
 }
 
-void TrackManager::UpdateMovingTrackSorting() {
+void TrackManager::UpdateMovingTrackPosInVisibleTracks() {
```
dpallotti, line 207: I think you shouldn't abbreviate and just use "Position".

dpallotti (approved):
Tried it, works for me.

PR #2348, opened 2021-05-19
Improvements in target selection window

reichlfl:
Two small improvements to the target selection:

1) Pressing "enter" in the process filter will start the session (b/183942482)
2) Instance list can now be sorted (b/187412728)

Tests:
1) Type a filter string in the process filter, press Enter
2) With 2 reserved instances, tested if sorting in all columns works, if the correct instance is connected afterwards, and if "Remember this instance" works correctly


antonrohr (approved):
I tested it and the code looks great! LGTM

PR #2350, opened 2021-05-19
Metrics: number of instrumented function timers + number of vulkan timers

antonrohr:
There is some explanation in the commit messages, together with bug links.


florian-kuebler (approved):
Thanks!

PR #2357, opened 2021-05-20
 Show unwinding errors in the capture timeline

dpallotti:
#### Add CallstackType type to orbit_grpc_protos::Callstack
And add support for it on the service.

Bug: http://b/188178748
Bug: http://b/188498204

Test: Update affected tests. Capture Trata and verify samples. Run
`LinuxTracingIntegrationTests` 1000 times.
#### Add CallstackType type to orbit_client_protos::CallstackInfo
Note that the logic of
`CallstackData::FilterCallstackEventsBasedOnMajorityStart` had to be slightly
changed (the method was renamed accordingly): instead of throwing away
`CallstackEvent`s, we now change the type of the referenced `CallstackInfo`.

Bug: http://b/188178748
Bug: http://b/188498206

Test: Update affected unit tests. Capture Trata and verify callstacks (the ones
that are not unwinding errors) are in order in the timeline and in the
aggregated views. Play around with selections to verify that non-`kComplete`
callstacks don't end up in them.
#### Show unwinding errors in the timeline 
Bug: http://b/188496243

Test: Capture Trata, look for unwinding errors in the timeline, check their
tooltips. One way to get a lot of them is to instrument a function called
frequently, as all the samples falling directly inside uprobes will be
errors.

Screenshot from Linux:
Screenshot from 2021-05-20 13-04-11_mod

### DIFF
--- a/src/ClientData/CallstackData.cpp
+++ b/src/ClientData/CallstackData.cpp
@@ -5,6 +5,7 @@
 #include "ClientData/CallstackData.h"
 
 #include <absl/container/flat_hash_map.h>
+#include <absl/container/flat_hash_set.h>
 
 #include <cstdint>
 #include <utility>
@@ -14,6 +15,7 @@
 #include "capture_data.pb.h"
 
 using orbit_client_protos::CallstackEvent;
+using orbit_client_protos::CallstackInfo;
 
 namespace orbit_client_data {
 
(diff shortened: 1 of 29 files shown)


dimitry- (approved):

src/LinuxTracing/LibunwindstackUnwinder.h
```diff
@@ -16,18 +16,37 @@
...
 
+class LibunwindstackResult {
+ public:
+  explicit LibunwindstackResult(
+      std::vector<unwindstack::FrameData> frames,
+      unwindstack::ErrorCode error_code = unwindstack::ErrorCode::ERROR_NONE)
+      : frames_{std::move(frames)}, error_code_{error_code} {}
+
+  [[nodiscard]] const std::vector<unwindstack::FrameData>& frames() const { return frames_; }
+  [[nodiscard]] std::vector<unwindstack::FrameData>& frames() { return frames_; }
```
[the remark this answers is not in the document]
dpallotti: Nope, I probably just added it mechanically. Removed.

src/OrbitGl/CallstackThreadBar.cpp
```diff
@@ -195,8 +189,14 @@ bool CallstackThreadBar::IsEmpty() const {
 }
 
 [[nodiscard]] std::string CallstackThreadBar::SafeGetFormattedFunctionName(
-    uint64_t addr, int max_line_length) const {
+    const orbit_client_protos::CallstackInfo& callstack, int frame_index,
+    int max_line_length) const {
   CHECK(capture_data_ != nullptr);
+  if (frame_index >= callstack.frames_size()) {
```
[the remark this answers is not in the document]
dpallotti: Yes, that's why I added it. The condition seemed to be fitting with the name of the function, "Safe...". But I'm open to suggestions.

dpallotti:
It appears that there's a corner case where the clang_format in the check behaves differently than mine.

src/LinuxTracing/UprobesUnwindingVisitorTest.cpp
```diff
@@ -35,7 +35,7 @@ class MockLibunwindstackMaps : public LibunwindstackMaps {
 
 class MockLibunwindstackUnwinder : public LibunwindstackUnwinder {
  public:
-  MOCK_METHOD(std::vector<unwindstack::FrameData>, Unwind,
+  MOCK_METHOD(LibunwindstackResult, Unwind,
```
florian-kuebler, line 38: Hehe, I see, I will have much fun merging my change and this one :)

dpallotti: I already had a lot of fun rebasing this change on your previous one 😅

florian-kuebler: Oh yeah, I can imagine this 💯 .

src/ClientData/CallstackData.cpp
```diff
@@ -238,34 +244,56 @@ void CallstackData::FilterCallstackEventsBasedOnMajorityStart() {
       }
     }
 
-    static constexpr double kFilterSupermajorityThreshold = 0.6;
+    static constexpr double kFilterSupermajorityThreshold = 0.75;
```
florian-kuebler, line 247: Is that modification intended? There should probably be then a comment. Is this value arbitrary?

florian-kuebler (approved):
Thanks!

src/ClientData/CallstackData.cpp
dpallotti: It is arbitrary, and it is intended, I wanted to make it stricter. Added a comment.

PR #2359, opened 2021-05-20
Avoid reporting wrong format error on empty line

vickyliu-go4it:
While splitting the linux file contents into lines, the previous code
will generate an empty line in the end and report wrong format on the
empty line. With this change, we avoid producing empty lines when
splitting the file content.

Bug: http://b/188758824


antonrohr (approved):
Thanks for fixing it that quickly. I tested it, LGTM!

PR #2361, opened 2021-05-20
Remove dead code from CoreUtils.h

dimitry-:
Also replace usages of Fill with array.fill() and
ToLower with absl::AsciiStrToLower.

Test: builds


beckerhe (approved):
Nice. Thanks!

src/OrbitCore/Path.cpp
```diff
@@ -51,13 +51,13 @@ std::filesystem::path Path::CreateOrGetCacheDir() {
...
 }
 
-std::filesystem::path Path::CreateOrGetPresetDir() {
+std::filesystem::path Path::CreateOrGetPresetDirPriorTo1_65() {
   std::filesystem::path preset_dir = Path::CreateOrGetOrbitAppDataDir() / "presets";
   std::filesystem::create_directory(preset_dir);
   return preset_dir;
 }
 
-std::filesystem::path Path::CreateOrGetCaptureDir() {
```
dimitry-, line 60: ok, this is accidental. reverting.

PR #2362, opened 2021-05-20
Differentiate main content with annotating lines

karupayun:
In this PR we are setting annotating lines to have lighter text color and lighter background. Now we can differentiate them from the main_content.

Second step in http://b/187388546.

Tested by CodeViewerDemo.
Syntax Highlighter combined

### DIFF
--- a/src/CodeViewer/Viewer.cpp
+++ b/src/CodeViewer/Viewer.cpp
@@ -24,6 +24,8 @@
 #include <QWheelEvent>
 #include <cmath>
 
+#include "SyntaxHighlighter/HighlightingMetadata.h"
+
 namespace orbit_code_viewer {
 
 static const QColor kLineNumberBackgroundColor{50, 50, 50};
@@ -43,7 +45,7 @@ int DetermineLineNumberWidthInPixels(const QFontMetrics& font_metrics, int max_l
 }
 
 namespace {
-struct Metadata : QTextBlockUserData {
+struct Metadata : public orbit_syntax_highlighter::HighlightingMetadata {
   enum LineType { kMainContent, kAnnotatingLine };
 
   LineType line_type = kMainContent;
@@ -52,6 +54,8 @@ struct Metadata : QTextBlockUserData {
   explicit Metadata(LineType line_type, uint64_t line_number)
       : line_type(line_type), line_number(line_number) {}
   explicit Metadata() = default;
+
+  [[nodiscard]] bool IsMainContentLine() const override { return line_type == kMainContent; }
 };
 }  // namespace
 
--- a/src/CodeViewer/Viewer.cpp
+++ b/src/CodeViewer/Viewer.cpp
@@ -29,6 +29,7 @@ namespace orbit_code_viewer {
 static const QColor kLineNumberBackgroundColor{50, 50, 50};
 static const QColor kLineNumberForegroundColor{189, 189, 189};
 static const QColor kTextEditBackgroundColor{30, 30, 30};
+static const QColor kAnnotatingLinesBackgroundColor{50, 50, 50};
 static const QColor kTextEditForegroundColor{189, 189, 189};
 static const QColor kTitleBackgroundColor{30, 65, 89};
 static const QColor kHeatmapColor{Qt::red};
@@ -474,10 +475,13 @@ void Viewer::SetHighlightCurrentLine(bool enabled) {
 bool Viewer::IsCurrentLineHighlighted() const { return is_current_line_highlighted_; }
 
 void Viewer::HighlightCurrentLine() {
-  const QColor highlight_color = palette().base().color().lighter();
+  QColor current_background_color = textCursor().block().blockFormat().background().color();
+  if (current_background_color == Qt::black) {  // color hasn't been set in the block
+    current_background_color = palette().base().color();
+  }
 
   QTextEdit::ExtraSelection selection{};
-  selection.format.setBackground(highlight_color);
+  selection.format.setBackground(current_background_color.lighter());
   selection.format.setProperty(QTextFormat::FullWidthSelection, true);
   selection.cursor = textCursor();
   selection.cursor.clearSelection();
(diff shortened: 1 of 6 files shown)


src/SyntaxHighlighter/X86Assembly.cpp
```diff
@@ -11,6 +11,18 @@
 #include <QString>
 #include <QTextCharFormat>
 
+// TODO: What is the best way to do this? Move the Metadata to another file? Declare it as extern?
```
beckerhe, line 14: Yes, a separate file is probably best. But also a lot of work, because it needs to go into its own cmake module.

I'm thinking it might be better to detect an assemble line by looking at the code. All assembly lines start with the address `0x...`.

src/SyntaxHighlighter/X86Assembly.cpp
```diff
@@ -216,9 +229,34 @@ const QRegularExpression kPlatformRegex{"^Platform:.*$"};
 X86Assembly::X86Assembly() : QSyntaxHighlighter{static_cast<QObject*>(nullptr)} {}
 
 void X86Assembly::highlightBlock(const QString& code) {
-  HighlightBlockAssembly(code, [this](int start, int count, const QTextCharFormat& format) {
-    setFormat(start, count, format);
-  });
+  const auto* const metadata = static_cast<const Metadata*>(currentBlock().userData());
```
beckerhe, line 232: If we go with currentBlock().userData() you also have to check whether it returns a null pointer.

If only main content is set, `userData` is not used. It will only be populated when there are also annotating lines.

src/SyntaxHighlighter/include/SyntaxHighlighter/HighlightingMetadata.h
```diff
@@ -0,0 +1,15 @@
+// Copyright (c) 2021 The Orbit Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#ifndef HIGHLIGHTING_METADATA_H_
+#define HIGHLIGHTING_METADATA_H_
+
+#include <QTextBlockUserData>
+
+class HighlightingMetadata : public QTextBlockUserData {
```
beckerhe, line 10: This needs to go into the `orbit_syntax_highlighter` namespace.

src/SyntaxHighlighter/include/SyntaxHighlighter/HighlightingMetadata.h
```diff
@@ -0,0 +1,15 @@
...
+ public:
+  virtual bool IsMainContentLine() const = 0;
```
beckerhe, line 12: nodiscard

src/SyntaxHighlighter/X86Assembly.cpp
```diff
@@ -216,9 +219,35 @@ const QRegularExpression kPlatformRegex{"^Platform:.*$"};
 X86Assembly::X86Assembly() : QSyntaxHighlighter{static_cast<QObject*>(nullptr)} {}
 
 void X86Assembly::highlightBlock(const QString& code) {
-  HighlightBlockAssembly(code, [this](int start, int count, const QTextCharFormat& format) {
-    setFormat(start, count, format);
-  });
+  const auto* const highlighting_metadata =
+      static_cast<const HighlightingMetadata*>(currentBlock().userData());
```
beckerhe, line 223: Since we can't ensure that userdata is always of type `const HighlightingMetadata*` we should use `dynamic_cast` here.

src/SyntaxHighlighter/X86Assembly.cpp
```diff
@@ -216,9 +219,35 @@ const QRegularExpression kPlatformRegex{"^Platform:.*$"};
...
+  if (highlighting_metadata->IsMainContentLine()) {
```
beckerhe, line 224: `highlighting_metadata` might be `nullptr`. We have to check that. If it's `nullptr` we can assume that it's main content.

src/CodeViewer/Viewer.cpp
```diff
@@ -474,10 +479,13 @@ void Viewer::SetHighlightCurrentLine(bool enabled) {
 bool Viewer::IsCurrentLineHighlighted() const { return is_current_line_highlighted_; }
 
 void Viewer::HighlightCurrentLine() {
-  const QColor highlight_color = palette().base().color().lighter();
+  QColor current_color = textCursor().block().blockFormat().background().color();
```
beckerhe, line 482: Nit: I would name this color `current_background_color`. WDYT?

src/SyntaxHighlighter/include/SyntaxHighlighter/X86Assembly.h
```diff
@@ -30,6 +30,9 @@ class X86Assembly : public QSyntaxHighlighter {
 
 void HighlightBlockAssembly(const QString& code,
                             std::function<void(int, int, const QTextCharFormat&)> set_format);
+void HighlightAnnotatingBlock(const QString& code,
+                              std::function<void(int, int, const QTextCharFormat&)> set_format,
+                              const QColor& default_color = Qt::white);
```
beckerhe, line 35: Sorry I missed this in the last commit already: We shouldn't copy the `std::function` object. A const-ref should be more performant.

src/SyntaxHighlighter/X86Assembly.cpp
karupayun: yes, yes. I was making that fixup, you were faster than me :P

karupayun: BTW, it's there

src/SyntaxHighlighter/X86Assembly.cpp
karupayun: As discussed offline, this won't work properly because the first lines in the assembly file will be highlight as C++ lines. We decided to not go in this direction.

src/SyntaxHighlighter/X86Assembly.cpp
karupayun: Yes, you are right!

beckerhe (approved):
I'm approving now since the important comments have been adressed. Thanks Pablo!

src/SyntaxHighlighter/include/SyntaxHighlighter/X86Assembly.h
beckerhe: ty

PR #2363, opened 2021-05-20
Integrate Combined Source Code View

beckerhe:
The first couple of commits are various small refactorings. The actual integration work happens in the last two commits (integration and test).

Test: There is actually a quite comprehensive unit-like test. On top I did a lot of manual testing.


antonrohr:
First nit that I am doing here as a separate comment: I would squash the commits "Move TryAskingTheUserAndReadSourceFile to separate module" and "Use SourcePathsMappingUI module in OrbitQt". The first commit actually adds code and the second one removes code. Having both in one commit makes it more clear that the code was actually moved.

src/SourcePathsMappingUI/AskUserForFile.cpp
```diff
@@ -0,0 +1,85 @@
+// Copyright (c) 2021 The Orbit Authors. All rights reserved.
```
reichlfl, line 1: This and the next commit should probably be merged or it may cause linker errors due to the redefinition of TryAskingTheUserAndReadSourceFile, won't it?

src/SourcePathsMappingUI/AskUserForFile.cpp
```diff
@@ -43,39 +46,61 @@ std::optional<QString> TryAskingTheUserAndReadSourceFile(QWidget* parent,
...
 
-  if (message_box.checkBox()->isChecked()) {
+  if (infer_source_paths_mapping) {
     auto maybe_mapping = orbit_source_paths_mapping::InferMappingFromExample(
         file_path, std::filesystem::path{user_chosen_file.toStdString()});
     if (maybe_mapping.has_value()) {
       orbit_source_paths_mapping::MappingManager mapping_manager{};
       mapping_manager.AppendMapping(maybe_mapping.value());
+    } else {
+      LOG("Unable to infer a mapping from \"%s\" to \"%s\"", file_path.string(),
```
antonrohr, line 102: Nit: I think this should be an ERROR

src/SourcePathsMappingUI/AskUserForFile.cpp
```diff
@@ -0,0 +1,85 @@
...
+
+namespace orbit_source_paths_mapping_ui {
+
+std::optional<QString> TryAskingTheUserAndReadSourceFile(QWidget* parent,
+                                                         const std::filesystem::path& file_path) {
+  QMessageBox message_box{QMessageBox::Warning, "Source code file not found",
+                          QString("Could not find the source code file \"%1\" on this machine.")
+                              .arg(QString::fromStdString(file_path.string())),
+                          QMessageBox::Cancel, parent};
+  auto* pick_file_button = message_box.addButton("Choose file...", QMessageBox::ActionRole);
```
reichlfl, line 34: I know that this was probably there before and should have been spotted in another PR, but most (all?) of the `auto` declarations should be replaced with types.

src/SourcePathsMappingUI/include/SourcePathsMappingUI/AskUserForFile.h
```diff
@@ -10,6 +10,8 @@
 #include <filesystem>
 #include <optional>
 
+#include "OrbitBase/Logging.h"
+
```
antonrohr, line 14: Include in the .cpp instead .h ?!

src/SourcePathsMappingUI/AskUserForFile.cpp
```diff
@@ -0,0 +1,85 @@
...
+
+  constexpr const char* kAutocreateMappingKey = "auto_create_mapping";
+  constexpr const char* kPreviousSourcePathsMappingDirectoryKey =
+      "previous_source_paths_mapping_directory";
+  auto checkbox = std::make_unique<QCheckBox>(
+      "Automatically create a source paths mapping from my selected file.");
+  checkbox->setToolTip(
+      "If enabled Orbit will automatically try to create a source paths mapping from it. The "
```
reichlfl, line 39: "If enabled, Orbit will..."

src/OrbitQt/orbitmainwindow.cpp
```diff
@@ -1352,29 +1353,24 @@ void OrbitMainWindow::ShowTooltip(std::string_view message) {
...
 }
 
 std::optional<QString> OrbitMainWindow::LoadSourceCode(const std::filesystem::path& file_path) {
-  auto maybe_source_code = TryReadSourceFile(QString::fromStdString(file_path.string()));
-  if (maybe_source_code.has_value()) return maybe_source_code.value();
+  auto source_code_or_error = orbit_base::ReadFileToString(file_path);
+  if (source_code_or_error.has_value()) return QString::fromStdString(source_code_or_error.value());
 
-  maybe_source_code = TryApplyMappingAndReadSourceFile(file_path);
+  auto maybe_source_code = TryApplyMappingAndReadSourceFile(file_path);
```
reichlfl, line 1374: More auto nitpicking.

src/OrbitQt/AnnotatingSourceCodeHandler.cpp
```diff
@@ -0,0 +1,178 @@
...
+    LOG("Error while loading debug information for the disassembly view: %s",
+        local_file_path_or_error.error().message());
+
+    dialog_->SetStatusMessage(QString::fromStdString(local_file_path_or_error.error().message()),
+                              std::nullopt);
+
+    return;
+  }
+
+  auto& local_file_path = local_file_path_or_error.value();
```
reichlfl, line 70: Also here, do we want this `auto`?

src/OrbitQt/AnnotatingSourceCodeHandler.cpp
```diff
@@ -0,0 +1,178 @@
...
+
+  auto elf_or_error = orbit_object_utils::CreateElfFile(local_file_path);
```
reichlfl, line 72: `auto`

src/OrbitQt/AnnotatingSourceCodeHandler.cpp
```diff
@@ -0,0 +1,178 @@
...
+  if (elf_or_error.has_error()) {
+    dialog_->SetStatusMessage(QString::fromStdString(local_file_path_or_error.error().message()),
+                              std::nullopt);
+    awaited_button_action_ = ButtonAction::kNone;
+    return;
+  }
+  elf_file_ = std::move(elf_or_error.value());
+
+  auto location_or_error = elf_file_->GetDeclarationLocationOfFunction(function_info_.address());
```
reichlfl, line 81: `auto`

src/OrbitQt/AnnotatingSourceCodeHandlerTest.cpp
```diff
@@ -0,0 +1,116 @@
...
+      orbit_base::GetExecutableDir() / "testdata" / "line_info_test_binary";
+
+  {
+    orbit_source_paths_mapping::MappingManager manager{};
+    manager.SetMappings({});
+    manager.AppendMapping(orbit_source_paths_mapping::Mapping{
+        std::filesystem::path{"."} / "..", orbit_base::GetExecutableDir() / "testdata"});
+  }
+
+  auto program = orbit_object_utils::CreateElfFile(file_path);
```
reichlfl, line 46: `auto`

src/OrbitQt/AnnotatingSourceCodeHandler.cpp
```diff
@@ -0,0 +1,178 @@
...
+#include "OrbitBase/ImmediateExecutor.h"
+#include "OrbitBase/ReadFileToString.h"
+#include "OrbitBase/Result.h"
+#include "QtUtils/MainThreadExecutorImpl.h"
+#include "SourcePathsMapping/MappingManager.h"
+#include "SourcePathsMappingUI/AskUserForFile.h"
+
+namespace orbit_qt {
+AnnotatingSourceCodeHandler::AnnotatingSourceCodeHandler(
+    const QPointer<orbit_code_viewer::OwningDialog>& dialog,
```
antonrohr, line 22: maybe CHECK(dialog != nullptr) ? Or would this class work with a nullptr?

reichlfl (commented):
I only did a quick sweep so far, I'll take another look after my meetings. I'm not sure if we have a common guideline on when we're using auto - as pointed out in another PR, I am also a fan of this, but got pushback from others, so now I'm passing that on :D

src/OrbitQt/AnnotatingSourceCodeHandler.h
```diff
@@ -0,0 +1,84 @@
...
+ public:
+  // Same interface as OrbitApp::RetrieveModuleWithDebugInfo;
+  using RetrieveModuleWithDebugInfoCallback =
+      std::function<orbit_base::Future<ErrorMessageOr<std::filesystem::path>>(const std::string&,
+                                                                              const std::string&)>;
+  explicit AnnotatingSourceCodeHandler(const QPointer<orbit_code_viewer::OwningDialog>& dialog,
+                                       orbit_client_protos::FunctionInfo function_info,
+                                       orbit_code_report::DisassemblyReport report,
+                                       RetrieveModuleWithDebugInfoCallback callback,
+                                       QObject* parent = nullptr);
```
antonrohr, line 51: `QObject* parent = nullptr` I assume you want to adhere to qt style, but iff a `orbit_code_viewer::OwningDialog` is required anyways, maybe skip that additional parent parameter and always use `dialog` as parent?

I bet you thought about it and have a reason why the current way is good. I'd be interested in that thought

antonrohr (commented):
I am still reviewing the last 2 commits, here are the first few nits.

src/SourcePathsMappingUI/AskUserForFile.cpp
```diff
@@ -43,39 +46,61 @@ std::optional<QString> TryAskingTheUserAndReadSourceFile(QWidget* parent,
...
   message_box.exec();
+
+  if (maybe_file_contents.has_value()) {
+    settings.setValue(kAutocreateMappingKey, message_box.checkBox()->isChecked());
+  }
+
+  return maybe_file_contents;
+}
+
+ErrorMessageOr<std::optional<QString>> ShowFileOpenDialogAndReadSourceFile(
```
reichlfl, line 72: I find these "DoXAndY" methods a bit confusing and long. Wouldn't it be easier for the code flow to split them into two methods that each do one thing? Same for `TryAskingTheUserAndReadSourceFile`.

src/OrbitQt/AnnotatingSourceCodeHandler.h
reichlfl: I was thinking exactly the same when reading this. What is the logical difference between owning dialog and parent, especially if the owning dialog is supposed to manage the lifetime?

Also about argument order, should (if it really is required separately) `parent` be the first or second argument, following how QT is usually defining the constructors?

reichlfl: Also, can you elaborate why `QPointer` instead of `OwningDialog*`? AFAIK we're not using this anywhere else?

src/OrbitQt/AnnotatingSourceCodeHandler.cpp
```diff
@@ -0,0 +1,178 @@
...
+      ChooseFile();
+      break;
+    case ButtonAction::kAddAnnotations:
+      HandleAnnotations();
+      break;
+  }
+}
+
+void AnnotatingSourceCodeHandler::HandleDebugInfo(
+    const ErrorMessageOr<std::filesystem::path>& local_file_path_or_error) {
```
reichlfl, line 56: Personal preference, I'd split this method into a few smaller methods.

reichlfl (commented):
Left some more comments - I have not yet looked at the tests in detail.

src/SourcePathsMappingUI/include/SourcePathsMappingUI/AskUserForFile.h
```diff
@@ -0,0 +1,37 @@
...
+// machine. The user now has to choice to abort the operation or choose a file manually with a
+// file-open-dialog.
+// If the users chooses the latter the file will be read and the contents returned as a QString.
+//
+// The user also has the choice to infer a source paths mapping from their chosen file path.
+std::optional<QString> TryAskingTheUserAndReadSourceFile(QWidget* parent,
+                                                         const std::filesystem::path& file_path);
+
+// Shows a file open dialog and returns the contents of the chosen file.
+// It will also try to create a source paths mapping when `infer_source_paths_mapping` is set to
```
karupayun, line 27: Nit: I would continue the sentence in the same line. "... chosen file. It will ... "

src/SourcePathsMappingUI/AskUserForFile.cpp
karupayun: I was thinking exactly the same. By the way, it's not always easy to do. I would prefer to have a method with all the logic together which call a bunch of small methods where each one only does one thing.

src/OrbitQt/AnnotatingSourceCodeHandler.h
beckerhe: This is implemented like this for a couple of reasons:

1. `OwningDialog` doesn't need to manage the lifetime. For example the test creates the object on the stack.
2. I want lifetime management to be explicit here. The user of this class needs to know what they are doing. If I remove the `parent` parameter it gets implicit and it gets unclear to users what is happening when they just read the line that is instantiating this object. By sticking to the Qt-standard-pattern. This is much more clear.
3. We need a QPointer to know when `OwningDialog` gets out of scope. A raw pointer might get dangling. That's why you see a couple of `if (dialog_.IsNull()) return;` statements in the code. Note, that this requirement won't go away even if we made the `parent` implicit and always set it to `dialog_` because all children are get destroyed after `OwningDialog` gets destroyed.

We use QPointer at a couple of locations in `OrbitGgp` and `OrbitSshQt` and `QtUtils`.

beckerhe: I also don't believe that "not used anywhere else" is a strong argument for not using it.

src/SourcePathsMappingUI/AskUserForFile.cpp
beckerhe: I will try to split it up.

src/SourcePathsMappingUI/AskUserForFile.cpp
beckerhe: It won't cause a linker error since they don't have the same symbol name. But I will merge it as suggested at the end of the review. :+1:

src/OrbitQt/AnnotatingSourceCodeHandlerTest.cpp
```diff
@@ -0,0 +1,116 @@
...
+  ASSERT_TRUE(program.has_value()) << program.error().message();
```
karupayun, line 47: I don't understand this. You are forwarding that error_message if the program doesn't have any value?

beckerhe: gtest will show the given error message if the assertion fails. That's a pattern we have in almost every test. Here is an example in the official documentation: https://google.github.io/googletest/primer.html#assertions

src/SourcePathsMappingUI/AskUserForFile.cpp
beckerhe: done.

src/OrbitQt/orbitmainwindow.cpp
beckerhe: done.

src/OrbitQt/AnnotatingSourceCodeHandler.cpp
beckerhe: done.

src/OrbitQt/AnnotatingSourceCodeHandler.cpp
beckerhe: done

src/OrbitQt/AnnotatingSourceCodeHandler.cpp
beckerhe: done

src/OrbitQt/AnnotatingSourceCodeHandlerTest.cpp
beckerhe: done.

src/OrbitQt/orbitmainwindow.cpp
```diff
@@ -1352,29 +1353,24 @@ void OrbitMainWindow::ShowTooltip(std::string_view message) {
...
-  return source_code_file.readAll();
-}
-
 static std::optional<QString> TryApplyMappingAndReadSourceFile(
     const std::filesystem::path& file_path) {
   orbit_source_paths_mapping::MappingManager mapping_manager{};
   const auto maybe_mapping_file_path = mapping_manager.MapToFirstExistingTarget(file_path);
   if (maybe_mapping_file_path.has_value()) {
-    return TryReadSourceFile(QString::fromStdString(maybe_mapping_file_path->string()));
+    auto result = orbit_base::ReadFileToString(*maybe_mapping_file_path);
```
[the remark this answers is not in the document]
beckerhe: done.

src/SourcePathsMappingUI/AskUserForFile.cpp
beckerhe: done.

src/SourcePathsMappingUI/include/SourcePathsMappingUI/AskUserForFile.h
beckerhe: yes. thanks. done.

src/OrbitQt/AnnotatingSourceCodeHandler.cpp
beckerhe: done.

src/SourcePathsMappingUI/include/SourcePathsMappingUI/AskUserForFile.h
beckerhe: done

src/SourcePathsMappingUI/AskUserForFile.cpp
beckerhe: done.

src/OrbitQt/AnnotatingSourceCodeHandlerTest.cpp
karupayun: Yes, I saw them a lot but I wasn't 100% sure. Thanks for that!

src/OrbitQt/AnnotatingSourceCodeHandler.cpp
```diff
@@ -0,0 +1,178 @@
...
+    return;
+  }
+
+  auto& local_file_path = local_file_path_or_error.value();
+
+  auto elf_or_error = orbit_object_utils::CreateElfFile(local_file_path);
+  if (elf_or_error.has_error()) {
+    dialog_->SetStatusMessage(QString::fromStdString(local_file_path_or_error.error().message()),
+                              std::nullopt);
+    awaited_button_action_ = ButtonAction::kNone;
```
antonrohr, line 62: why this line here? 
And iff its necessary, then maybe also add it in the last if

src/OrbitQt/AnnotatingSourceCodeHandler.cpp
```diff
@@ -0,0 +1,178 @@
...
+  HandleSourceCode(result.value().value());
+}
+
+void AnnotatingSourceCodeHandler::HandleSourceCode(const QString& source_file_contents) {
+  annotations_ = orbit_code_report::AnnotateDisassemblyWithSourceCode(
+      function_info_, location_info_, source_file_contents.toStdString(), elf_file_.get(), report_);
+
+  // When loading the source code takes less than the given time we won't ask the user if they want
+  // to have the source code added for context. We will add it right away - as it's done in Visual
+  // Studio.
```
antonrohr, line 169: I like this!

src/OrbitQt/AnnotatingSourceCodeHandler.cpp
```diff
@@ -0,0 +1,178 @@
...
+  constexpr std::chrono::steady_clock::duration kMaxWaitingTime = std::chrono::milliseconds{250};
+
+  // ButtonAction::kChooseFile means the user had to select the source file path manually. In this
+  // case we don't keep them waiting even if loading took longer than the limit.
+  if (std::chrono::steady_clock::now() - starting_time_ < kMaxWaitingTime ||
+      awaited_button_action_ == ButtonAction::kChooseFile) {
```
antonrohr, line 175: pers pref: in combination with how the comment is worded
`awaited_button_action_ == ButtonAction::kChooseFile || std::chrono::steady_clock::now() - starting_time_ < kMaxWaitingTime`

src/OrbitQt/AnnotatingSourceCodeHandler.cpp
beckerhe: Done. PTAL!

src/SourcePathsMappingUI/AskUserForFile.cpp
```diff
@@ -0,0 +1,85 @@
...
+  checkbox->setToolTip(
+      "If enabled Orbit will automatically try to create a source paths mapping from it. The "
+      "common suffix between the path given in the debug information and the local file path will "
+      "be stripped. From the rest a mapping will be created.");
+  QSettings settings{};
+  checkbox->setCheckState(settings.value(kAutocreateMappingKey, true).toBool() ? Qt::Checked
+                                                                               : Qt::Unchecked);
+
+  // Ownership will be transferred to message_box
+  message_box.setCheckBox(checkbox.release());
```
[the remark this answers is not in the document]
beckerhe: done.

antonrohr (approved):
just nits. LGTM

src/OrbitQt/AnnotatingSourceCodeHandler.h
reichlfl: Not a strong argument, but I do think "consistency" is an argument. But sure, if it makes sense then this should not be a reason against it.

Regarding lifetime, isn't the Qt standard pattern that parent manages the lifetime, or am I mixing this up now?

src/OrbitQt/AnnotatingSourceCodeHandler.cpp
reichlfl: Thanks, I like the update a lot! This provides a super clear picture of the flow without actually reading the details.

One nitput, and this may be controversial: I would pass the input (e.g. `elf_file_`) required in each of the steps as an argument to make it explicit, even if it it stored in a member variable. Right now, you have implicit input and output locations, which makes it less readable.

But feel free to ignore.

beckerhe: I completely agree. This is why I actually prefer the previous solution. :sweat_smile: 

But I could make these functions free functions and pass in all in- and outputs. WDYT?

reichlfl: Would be an option, that would also minimize the header file. I don't think that having the output implicit is a big issue though - as for the input, this prevents others that may extend the class to just call the function with wrong preconditions and then try to track down the bug.

Whatever you choose :-)

src/OrbitQt/AnnotatingSourceCodeHandler.h
beckerhe:
```
Not a strong argument, but I do think "consistency" is an argument. But sure, if it makes sense then this should not be a reason against it.
```

It's not about consistency. QPointer solves a problem which can't be solved by any other mechanism.

```
Regarding lifetime, isn't the Qt standard pattern that parent manages the lifetime, or am I mixing this up now?
```
No you are correct. Parents delete children when they get deleted theirselves. But this happens in the destructor of QObject, so after `OwningDialog`'s destructor has been executed. That means there is a small period of time when `AnnotatingSourceCodeHandler` is alive and `OwningDialog` is not.

This is not a problem as long as `AnnotatingSourceCodeHandler` does not access `OwningDialog` during that time. And we ensure that by checking if the QPointer is null. Qt's data views and data models do similar things to ensure that destruction oder doesn't matter.

If we want to avoid the whole QPointer-thing we can make `OwningDialog` a member of `AnnotatingSourceCodeHandler`. That would ensure a strict lifetime hierarchy and the problem can't occur. Maybe that's not the worst idea. WDYT? Of course we would rename it to `AnnotatingSourceCodeDialog` or something similar.

reichlfl: It's not so much about `QPointer`, I've just not seen us use that before, that's why it stood out to me, and really the thing that struck me most odd was that we pass `dialog` and `parent` separately. Getting rid of this would feel clearer and more "familiar" to me.

But I don't have a super strong opinion now that I understand the reasoning behind it, I'm also fine with leaving it as is, maybe then this reasoning should be pointed out in a comment.

beckerhe: Okay, I will have a quick look if changing this makes it simpler.

beckerhe: A refactored the class and there is now an `AnnotatingSourceCodeDialog` which inherits from `orbit_code_viewer::Dialog`. PTAL

src/SourcePathsMappingUI/AskUserForFile.cpp
beckerhe: Done. PTAL.

src/OrbitQt/AnnotatingSourceCodeHandler.cpp
beckerhe: So I tried this. But it doesn't realy work. It makes a whole mess out of everything. So I tend to refrain from making them free functions and leaving them as non-const member functions with implicit state changes.

src/OrbitQt/AnnotatingSourceCodeHandler.cpp
```diff
@@ -0,0 +1,178 @@
...
+    const ErrorMessageOr<std::filesystem::path>& local_file_path_or_error) {
+  // No dialog means the user aborted the operation
+  if (dialog_.isNull()) return;
+
+  if (local_file_path_or_error.has_error()) {
+    LOG("Error while loading debug information for the disassembly view: %s",
+        local_file_path_or_error.error().message());
+
+    dialog_->SetStatusMessage(QString::fromStdString(local_file_path_or_error.error().message()),
+                              std::nullopt);
```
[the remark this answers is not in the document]
beckerhe: Done.

src/OrbitQt/AnnotatingSourceCodeHandler.cpp
beckerhe: I think this is obsolete now due to the dismiss-button.

src/OrbitQt/AnnotatingSourceCodeHandler.cpp
beckerhe: I agree. Done.

src/OrbitQt/AnnotatingSourceCodeDialog.h
```diff
@@ -0,0 +1,106 @@
...
+  std::filesystem::path local_source_file_path_;
+  std::unique_ptr<orbit_object_utils::ElfFile> elf_file_;
+  orbit_grpc_protos::LineInfo location_info_;
+  std::vector<orbit_code_report::AnnotatingLine> annotations_;
+};
+
+// This function opens the given dialog and ensures it is deleted when closed.
+// Note, this function returns immediately after opening the dialog, NOT when it is closed. Use
+// `QDialog::exec` to wait for the dialog.
+QPointer<AnnotatingSourceCodeDialog> OpenAndDeleteOnClose(
```
reichlfl, line 102: Nit and not a strong opinion: This could be a static member of `AnnotatingSourceCodeDialog` since it's tightly coupled to that scope.

reichlfl (approved):
IMO the new `AnnotatingSourceCodeDialog` is a good improvement, thanks! LGTM.

src/OrbitQt/AnnotatingSourceCodeHandler.cpp
reichlfl: SGTM.

src/OrbitQt/AnnotatingSourceCodeHandler.h
reichlfl: I like this, thanks!

src/SourcePathsMappingUI/AskUserForFile.cpp
reichlfl: LGTM!

beckerhe:
Thank you all for the reviews. I will rebase this, clean up all the fixup commits, conduct another round of testing, and merge it when done.

karupayun (approved):

PR #2365, opened 2021-05-21
Always hook vkQueuePresentKHR in FakeClient to output avg frame time

dpallotti:
But only if the target has loaded `libvulkan.so.1`.

This might slightly affect the existing metrics in `OrbitService` performance
testing, but it seems only within measurement error. See below.

Bug: http://b/187827094

Test:

Tried on Trata, Infiltrator, and a binary that doesn't to graphics.
With Trata and Infiltrator, compare results with Orbit UI's frame tracks and
with the "Frame rate rendered" of the Stadia extension.
Note that Infiltrator doesn't hit 60 fps.

Compared the results of the OrbitService performance testing on YHITI with the
previous and with this version of `OrbitFakeClient`.
Previous: http://sponge2/6cd37b5c-792f-47b1-84db-6c4240377e77
This change: http://sponge2/8114a173-2bc6-4e7f-a165-1f6658eb3150
There doesn't seem to be any relevant difference.

### DIFF
--- a/src/FakeClient/FakeCaptureEventProcessor.h
+++ b/src/FakeClient/FakeCaptureEventProcessor.h
@@ -10,43 +10,83 @@
 
 namespace orbit_fake_client {
 
-// This implementation of CaptureEventProcessor simply discards all the events it receives, but it
-// keeps track of their number and total size and then outputs these statistics to file.
+// This implementation of CaptureEventProcessor mostly discard all events it receives, except for:
+// - keeping track of their number and total size, and writing these statistics to file;
+// - keeping track of the calls to the frame boundary function, and possibly writing the average
+//   frame time to file;
 class FakeCaptureEventProcessor : public orbit_capture_client::CaptureEventProcessor {
  public:
   void ProcessEvent(const orbit_grpc_protos::ClientCaptureEvent& event) override {
     ++event_count_;
     byte_count_ += event.ByteSizeLong();
+
+    // Keep track of the number of calls to the frame boundary function, of the timestamp of the
+    // first call, and of the timestamp of the last call. Below, the average frame time is then
+    // naively computed as (max_timestamp - min_timestamp) / (call_count - 1).
+    if (!event.has_function_call()) {
+      return;
+    }
+    const orbit_grpc_protos::FunctionCall& function_call = event.function_call();
+    if (function_call.function_id() != kFrameBoundaryFunctionId) {
+      return;
+    }
+    ++frame_boundary_count_;
+    uint64_t start_timestamp_ns = function_call.end_timestamp_ns() - function_call.duration_ns();
+    frame_boundary_min_timestamp_ns_ =
+        std::min(frame_boundary_min_timestamp_ns_, start_timestamp_ns);
+    frame_boundary_max_timestamp_ns_ =
+        std::max(frame_boundary_max_timestamp_ns_, start_timestamp_ns);
   }
 
   ~FakeCaptureEventProcessor() override {
     {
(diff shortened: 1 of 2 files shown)


src/FakeClient/FakeCaptureEventProcessor.h
```diff
@@ -10,43 +10,79 @@
...
+// - keeping track of their number and total size, and writing these statistics to file;
+// - keeping track of the calls to the frame boundary function, and possibly writing the average
+//   frame time to file;
 class FakeCaptureEventProcessor : public orbit_capture_client::CaptureEventProcessor {
  public:
   void ProcessEvent(const orbit_grpc_protos::ClientCaptureEvent& event) override {
     ++event_count_;
     byte_count_ += event.ByteSizeLong();
+
+    if (event.has_function_call()) {
```
dimitry-, line 23: early return?

src/FakeClient/FakeCaptureEventProcessor.h
```diff
@@ -10,43 +10,79 @@
...
+      const orbit_grpc_protos::FunctionCall& function_call = event.function_call();
+      if (function_call.function_id() == kFrameBoundaryFunctionId) {
```
dimitry-, line 25: same here

src/FakeClient/FakeCaptureEventProcessor.h
```diff
@@ -10,43 +10,79 @@
...
+        uint64_t start_timestamp_ns =
```
danielfenner, line 26: I was a bit confused here. I expected just summing up all the `function_call.duration_ns()`, count the frames and do a division below. But this should work as well.

danielfenner (approved):
lgtm!

dimitry- (approved):
otherwise lgtm

src/FakeClient/FakeCaptureEventProcessor.h
```diff
@@ -10,43 +10,79 @@
...
+            function_call.end_timestamp_ns() - function_call.duration_ns();
+        ++frame_boundary_count_;
```
florian-kuebler, line 28: So you are basically counting the frames and then compute (max_timestamp - min_timestamp)/#frames? I have the feeling that you should add a brief comment here.

src/FakeClient/FakeCaptureEventProcessor.h
dpallotti: If I sum the durations of the calls I get the total time spent in `vkQueuePresentKHR`, while I want the sum of the times between the calls. Or did I understand incorrectly?

florian-kuebler (approved):
LGTM

src/FakeClient/FakeCaptureEventProcessor.h
dpallotti: Exactly. Added comment.

src/FakeClient/FakeCaptureEventProcessor.h
dpallotti: Oh yes, done.

src/FakeClient/FakeCaptureEventProcessor.h
dpallotti: Done.

src/FakeClient/FakeCaptureEventProcessor.h
danielfenner: Oh, right, I was fooled by the word "duration". As you said that is the duration of the call `vkQueuePresentKHR` not the frame.

PR #2366, opened 2021-05-25
Add some helper functions to File.h

dimitry-:
They work with std::filesystem::* counterparts in exception free way.

Test: run OrbitBaseTests

### DIFF
--- a/src/OrbitBase/FileTest.cpp
+++ b/src/OrbitBase/FileTest.cpp
@@ -8,6 +8,7 @@
 #include "OrbitBase/ExecutablePath.h"
 #include "OrbitBase/File.h"
 #include "OrbitBase/TemporaryFile.h"
+#include "OrbitBase/TestUtils.h"
 
 namespace orbit_base {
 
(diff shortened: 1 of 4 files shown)


beckerhe (approved):
It looks good to me, but it seems Windows doesn't agree.

src/OrbitBase/FileTest.cpp
```diff
@@ -267,4 +269,52 @@ TEST(File, ReadStructureAtOffset) {
...
+TEST(File, ListFilesInDirectory) {
+  auto tmp_file_or_error = TemporaryFile::Create();
+  ASSERT_THAT(tmp_file_or_error, HasNoError());
+  auto tmp_file = std::move(tmp_file_or_error.value());
+  auto file_list_or_error = ListFilesInDirectory(tmp_file.file_path().parent_path());
+  ASSERT_THAT(file_list_or_error, HasNoError());
+  const auto& file_list = file_list_or_error.value();
+  auto it = find(file_list.begin(), file_list.end(), tmp_file.file_path());
+  EXPECT_TRUE(it != file_list.end())
+      << "File " << tmp_file.file_path() << "was not found in the list: "
```
dpallotti, line 314:
```suggestion
      << "File " << tmp_file.file_path() << " was not found in the list: "
```

src/OrbitBase/FileTest.cpp
```diff
@@ -267,4 +269,52 @@ TEST(File, ReadStructureAtOffset) {
...
+  EXPECT_TRUE(exists_or_error.value());
+
+  exists_or_error = FileExists(tmp_file.file_path());
+  ASSERT_THAT(exists_or_error, HasNoError());
+  EXPECT_FALSE(exists_or_error.value());
+
+  remove(new_path.string().c_str());
+}
+
+TEST(File, ListFilesInDirectory) {
```
dpallotti, line 305: Some vertical space in this test wouldn't hurt.

dpallotti (approved):

src/OrbitBase/FileTest.cpp
dimitry-: done.

src/OrbitBase/FileTest.cpp
dimitry-: done.

src/OrbitBase/TemporaryFile.cpp
```diff
@@ -44,6 +44,9 @@ ErrorMessageOr<void> TemporaryFile::Init() {
         absl::StrFormat("Unable to create a temporary file: %s", SafeStrerror(errnum))};
   }
 
+  // get rid of '\0' at the end
+  file_path = std::filesystem::path(file_path.string().c_str());
```
danielfenner, line 48: `file_path `is a string already. Perhaps make a copy of the `file_path `above and append the` \0` to that copy.

danielfenner (approved):
lgtm

src/OrbitBase/TemporaryFile.cpp
dimitry-: but then we will need to copy it over.. _mktemp_s modifies the string. Fixed by replacing fs::path with std::string() here.

PR #2368, opened 2021-05-25
[ON RELEASE] Set pretty(demangled)_name when constructing FunctionInfo from Instru…

dimitry-:
…mentedFunction

Bug: http://b/189222632
Test: run capture on OrbitTestLegacyApi

### DIFF
--- a/src/OrbitGl/LiveFunctionsDataView.cpp
+++ b/src/OrbitGl/LiveFunctionsDataView.cpp
@@ -10,10 +10,9 @@
 #include <absl/strings/str_format.h>
 #include <absl/strings/str_split.h>
 #include <absl/time/time.h>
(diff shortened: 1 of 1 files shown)


beckerhe (approved):
I'm not on top of what you discussed, but code-wise it looks good to me and it makes sense.

antonrohr:
By now it tested this PR like this:

Checkout the PR (lets call it 1.64),
take capture with 1.64 and OrbitTestLegacyApi 
save that capture
load it with 1.64 
load it with 1.63 
take a capture with 1.63 and OrbitTestLegacyApi
save that capture
load it with 1.63 
load it with 1.64 

It all lgtm to me

antonrohr:
Merging this now

PR #2371, opened 2021-05-26
Mention minimum Ubuntu release in DEVELOPMENT.md

beckerhe:
A common problem when compiling Orbit is a too old version of Ubuntu LTS. Let's mention that at least in the DEVELOPMENT.md file.


karupayun (approved):
LGTM!

PR #2373, opened 2021-05-27
Remove TimeGraph dependency from LiveFunctionsDataView

beckerhe:
Eventually LiveFunctionsDataView should be moved into a separate DataView module to make testing easier.
For this to happen all the DataViews may not depend on TimeGraph anymore.

So this PR removes TimeGraph dependency from LiveFunctionsDataView.

### DIFF
--- a/src/OrbitGl/App.cpp
+++ b/src/OrbitGl/App.cpp
@@ -2355,6 +2355,33 @@ bool OrbitApp::HasFrameTrackInCaptureData(uint64_t instrumented_function_id) con
   return GetTimeGraph()->HasFrameTrack(instrumented_function_id);
 }
 
+void OrbitApp::JumpToTextBoxAndZoom(uint64_t function_id, JumpToTextBoxMode selection_mode) {
+  switch (selection_mode) {
+    case JumpToTextBoxMode::kFirst: {
+      const auto* first_box = GetTimeGraph()->FindNextFunctionCall(
+          function_id, std::numeric_limits<uint64_t>::lowest());
+      if (first_box != nullptr) GetMutableTimeGraph()->SelectAndZoom(first_box);
+      break;
+    }
+    case JumpToTextBoxMode::kLast: {
+      const auto* last_box = GetTimeGraph()->FindPreviousFunctionCall(
+          function_id, std::numeric_limits<uint64_t>::max());
+      if (last_box != nullptr) GetMutableTimeGraph()->SelectAndZoom(last_box);
+      break;
+    }
+    case JumpToTextBoxMode::kMin: {
+      auto [min_box, unused_max_box] = GetTimeGraph()->GetMinMaxTextBoxForFunction(function_id);
+      if (min_box != nullptr) GetMutableTimeGraph()->SelectAndZoom(min_box);
+      break;
+    }
+    case JumpToTextBoxMode::kMax: {
+      auto [unused_min_box, max_box] = GetTimeGraph()->GetMinMaxTextBoxForFunction(function_id);
+      if (max_box != nullptr) GetMutableTimeGraph()->SelectAndZoom(max_box);
+      break;
+    }
+  }
+}
+
 void OrbitApp::RefreshFrameTracks() {
   CHECK(HasCaptureData());
   CHECK(std::this_thread::get_id() == main_thread_id_);
--- a/src/OrbitGl/App.h
+++ b/src/OrbitGl/App.h
@@ -452,6 +452,9 @@ class OrbitApp final : public DataViewFactory, public orbit_capture_client::Capt
 
   [[nodiscard]] bool HasFrameTrackInCaptureData(uint64_t instrumented_function_id) const;
 
+  enum class JumpToTextBoxMode { kFirst, kLast, kMin, kMax };
+  void JumpToTextBoxAndZoom(uint64_t function_id, JumpToTextBoxMode selection_mode);
+
  private:
   void UpdateModulesAbortCaptureIfModuleWithoutBuildIdNeedsReload(
       absl::Span<const orbit_grpc_protos::ModuleInfo> module_infos);
--- a/src/OrbitGl/LiveFunctionsDataView.cpp
+++ b/src/OrbitGl/LiveFunctionsDataView.cpp
@@ -293,33 +293,19 @@ void LiveFunctionsDataView::OnContextMenu(const std::string& action, int menu_in
   } else if (action == kMenuActionJumpToFirst) {
     CHECK(item_indices.size() == 1);
     auto function_id = GetInstrumentedFunctionId(item_indices[0]);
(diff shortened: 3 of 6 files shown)


src/OrbitGl/LiveFunctionsDataView.cpp
```diff
@@ -312,14 +310,14 @@ void LiveFunctionsDataView::OnContextMenu(const std::string& action, int menu_in
   } else if (action == kMenuActionJumpToMin) {
     CHECK(item_indices.size() == 1);
     uint64_t function_id = GetInstrumentedFunctionId(item_indices[0]);
-    auto [min_box, _] = GetMinMax(function_id);
+    auto [min_box, _] = app_->GetTimeGraph()->GetMinMaxTextBoxForFunction(function_id);
```
dpallotti: `_` -> `unused_max_box`

src/OrbitGl/LiveFunctionsDataView.cpp
```diff
@@ -312,14 +310,14 @@ void LiveFunctionsDataView::OnContextMenu(const std::string& action, int menu_in
...
     if (min_box != nullptr) {
       app_->GetMutableTimeGraph()->SelectAndZoom(min_box);
     }
   } else if (action == kMenuActionJumpToMax) {
     CHECK(item_indices.size() == 1);
     uint64_t function_id = GetInstrumentedFunctionId(item_indices[0]);
-    auto [_, max_box] = GetMinMax(function_id);
+    auto [_, max_box] = app_->GetTimeGraph()->GetMinMaxTextBoxForFunction(function_id);
```
dpallotti: `_` -> `unused_min_box`

src/OrbitGl/App.cpp
```diff
@@ -2355,6 +2355,33 @@ bool OrbitApp::HasFrameTrackInCaptureData(uint64_t instrumented_function_id) con
...
+      break;
+    }
+    case SelectionMode::kLast: {
+      const auto* last_box = GetTimeGraph()->FindPreviousFunctionCall(
+          function_id, std::numeric_limits<uint64_t>::max());
+      if (last_box != nullptr) GetMutableTimeGraph()->SelectAndZoom(last_box);
+      break;
+    }
+    case SelectionMode::kMin: {
+      auto [min_box, _] = GetTimeGraph()->GetMinMaxTextBoxForFunction(function_id);
```
dpallotti, line 2373: `_` -> `unused_max_box`

src/OrbitGl/App.cpp
```diff
@@ -2355,6 +2355,33 @@ bool OrbitApp::HasFrameTrackInCaptureData(uint64_t instrumented_function_id) con
...
+      if (min_box != nullptr) GetMutableTimeGraph()->SelectAndZoom(min_box);
+      break;
+    }
+    case SelectionMode::kMax: {
+      auto [_, max_box] = GetTimeGraph()->GetMinMaxTextBoxForFunction(function_id);
```
dpallotti, line 2378: `_` -> `unused_min_box`

src/OrbitGl/App.h
```diff
@@ -452,6 +452,9 @@ class OrbitApp final : public DataViewFactory, public orbit_capture_client::Capt
 
   [[nodiscard]] bool HasFrameTrackInCaptureData(uint64_t instrumented_function_id) const;
 
+  enum class SelectionMode { kFirst, kLast, kMin, kMax };
```
dpallotti, line 455: I don't like the name of this `enum` and of the method too much, I think "selection" is too generic. If I think of selection I think more of a user clicking or dragging on a visual element, while here what really happens is a search, that results in moving to a different position in the capture, and finally a selection. I propose two alternatives:
- Rename this `enum` to something like `JumpToTextBoxMode` or `JumpToAndSelectTextBoxMode` or `FindAndSelectTextBoxMode`, and the method to something like `JumpToTextBoxAndZoom` or `JumpToAndSelectTextBox` or `FindAndSelectTextBox` ("find" is probably my favorite);
- Have four different methods instead of the switch, and call them something like `JumpToFirstTextBoxAndZoom`, etc., or `JumpToAndSelectFirstTextBoxAndZoom`, etc., or `FindAndSelectFirstTextBoxAndZoom`, etc.
What do you think?

src/OrbitGl/App.h
```diff
@@ -452,6 +452,9 @@ class OrbitApp final : public DataViewFactory, public orbit_capture_client::Capt
...
+  void SelectTextBoxAndZoom(uint64_t function_id, SelectionMode selection_mode);
```
karupayun, line 456: I'm undecided if this function should be inside TimeGraph or CaptureWindow, but I think that it shouldn't be inside App.

beckerhe: All the other functions of this type are currently in App (See SelectTextBox for example) and it needs to be in App if we want to introduce `AppInterface` as discussed in the doc.

It might move somewhere else in the long run but for now there is no way around it.

src/OrbitGl/App.h
beckerhe: I agree. I will go for `JumpToTextBoxMode` and `JumpToTextBoxAndZoom`.

beckerhe: Done.

src/OrbitGl/App.cpp

src/OrbitGl/App.cpp

src/OrbitGl/App.h
karupayun: Ok. I will include some wording in the last commit about why we are moving this method to App, and our next steps.

dpallotti (approved):

src/OrbitGl/App.h
dpallotti: Great, thanks!

src/OrbitGl/TimeGraph.cpp
```diff
@@ -972,6 +972,33 @@ const TextBox* TimeGraph::FindDown(const TextBox* from) {
...
+    uint64_t function_id) const {
+  const TextBox* min_box = nullptr;
+  const TextBox* max_box = nullptr;
+  std::vector<std::shared_ptr<TimerChain>> chains = GetAllThreadTrackTimerChains();
+  for (auto& chain : chains) {
+    if (!chain) continue;
+    for (auto& block : *chain) {
+      for (size_t i = 0; i < block.size(); i++) {
+        const TextBox& box = block[i];
+        if (box.GetTimerInfo().function_id() == function_id) {
```
dimitry-, line 985: nit: != and continue?

beckerhe: done.

src/OrbitGl/App.h
beckerhe: Done. PTAL

karupayun (approved):
LGTM!

antonrohr (approved):

PR #2376, opened 2021-05-27
Add more exception free wrappers to OrbitBase and some cleanups

dimitry-:


src/OrbitBase/FileTest.cpp
```diff
@@ -305,6 +305,22 @@ TEST(File, MoveFile) {
   EXPECT_FALSE(exists_or_error.value());
 }
 
+TEST(File, RemoveFile) {
+  auto tmp_file_or_error = TemporaryFile::Create();
+  ASSERT_THAT(tmp_file_or_error, HasNoError());
+  auto tmp_file = std::move(tmp_file_or_error.value());
+
+  // Windows is weird, for some reason it does not allow to move opened files.
```
beckerhe, line 313: This comment is from the `MoveFile` test, isn't it? It makes no sense here

src/OrbitBase/File.cpp
```diff
@@ -231,6 +231,16 @@ ErrorMessageOr<void> RemoveFile(const std::filesystem::path& file_path) {
   return outcome::success();
 }
 
+ErrorMessageOr<void> CreateDirectory(const std::filesystem::path& file_path) {
+  std::error_code error;
+  std::filesystem::create_directories(file_path, error);
```
beckerhe, line 236: Have you considered `ErrorMessageOr<bool>` as the return type? `create_directories` will return false if the directory already exists. This might be useful in some situations.

beckerhe (approved):
Otherwise LGTM

src/Orbit/main.cpp
```diff
@@ -157,7 +159,7 @@ static QStringList ExtractCommandLineFlags(const std::vector<std::string>& comma
   QStringList command_line_flags;
   absl::flat_hash_set<std::string> positional_arg_set(positional_args.begin(),
                                                       positional_args.end());
-  for (std::string command_line_arg : command_line_args) {
+  for (const std::string& command_line_arg : command_line_args) {
```
vickyliu-go4it, line 162: Nit: Maybe use `string_view` instread of `const std::string&`.

vickyliu-go4it (approved):
LGTM! Only some minor suggestions : )

src/Orbit/main.cpp
dimitry-: command_line_args is a vector of strings.. what are advantages of using string_view in this case?

src/OrbitBase/File.cpp
dimitry-: done.

src/OrbitBase/FileTest.cpp
dimitry-: fixed comment.

src/Orbit/main.cpp
vickyliu-go4it: I was thinking that `const std::string&` can be replace by `string_view` as the latter one would be faster in general cases (go/totw/179). 

But I agree that there is no advantage in this case.

src/OrbitBase/File.cpp
```diff
@@ -221,6 +221,26 @@ ErrorMessageOr<void> MoveFile(const std::filesystem::path& from, const std::file
   return outcome::success();
 }
 
+ErrorMessageOr<void> RemoveFile(const std::filesystem::path& file_path) {
```
dpallotti, line 224: Building on Henning's other comment, if I understand correctly `remove` also returns a `bool`. If I understand correctly, when the file didn't exists then it returns false without setting the error.

src/OrbitCore/Path.cpp
```diff
@@ -33,9 +34,17 @@ static std::string GetEnvVar(const char* variable_name) {
   return var;
 }
 
+static void CreateDirectoryOrDie(const std::filesystem::path& directory) {
+  auto create_directory_result = orbit_base::CreateDirectory(directory);
+  if (create_directory_result.has_error()) {
+    FATAL("Unable to create directory \"%s\": %s", directory.string(),
```
dpallotti, line 40: Are we fine with just crashing in all the functions below?

src/OrbitBase/FileTest.cpp
```diff
@@ -305,6 +305,40 @@ TEST(File, MoveFile) {
   EXPECT_FALSE(exists_or_error.value());
 }
 
+TEST(File, RemoveFile) {
+  auto tmp_file_or_error = TemporaryFile::Create();
+  ASSERT_THAT(tmp_file_or_error, HasNoError());
+  auto tmp_file = std::move(tmp_file_or_error.value());
+
+  // Windows is weird, for some reason it does not allow to remove opened files.
```
dpallotti, line 313: That's weird from a Linux perspective, but I've seen the "file is open in another program" error often enough to not find this weird at all. 😀 (Nothing to do here anyway.)

dpallotti (approved):

src/OrbitCore/Path.cpp
dimitry-: we have done it before just implicitly, but I agree there is a better way to do it, we will need to separate create and get have create return ErrorMessageOr and call initialization from somewhere we can display the error. Orbit should still probably not start but at least user will know why and it won't be a crash.

But this is for another PR, I did not want to do it now.

src/OrbitBase/File.cpp
dimitry-: right.

dimitry-: done.

src/OrbitQt/MoveFilesDialog.ui
```diff
@@ -0,0 +1,84 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
dpallotti, line 1: This file doesn't belong in this change, does it?

src/OrbitBase/FileTest.cpp
dpallotti: In a way this comment could have stayed, it was still helpful. Maybe just "Windows doesn't allow to remove opened files."

src/OrbitQt/MoveFilesDialog.ui
dimitry-: ouch... right sorry. removing.

src/OrbitBase/FileTest.cpp
dimitry-: removing the file it is now part of the test-case we want to make sure RemoveFile on non-existing file returns false.

dpallotti: Ah yeah, sorry.

PR #2378, opened 2021-05-27
Relocate instructions.

danielfenner:
Tooling to move instructions in memory. Interesting instructions are
e.g. relative jumps or instructions using ip relative addressing.

In order to keep the size of the PR limited this contains only the
code to move one individual instruction so there are no tests yet
that actually run the relocated instruction. This will follow in the
next PR.

Bug: http://b/183113989
Test: Unit tests in PR

### DIFF
--- a/src/UserSpaceInstrumentation/Trampoline.cpp
+++ b/src/UserSpaceInstrumentation/Trampoline.cpp
@@ -199,11 +199,16 @@ ErrorMessageOr<RelocatedInstruction> RelocateInstruction(cs_insn* instruction, u
     // Most of these are optional - at least one byte of opcode needs to be present.
     // Many instructions that refer to an operand in memory have an addressing-form specifier byte
     // (called the modrm byte) following the primary opcode.
-    // In case (modrm & 0xC7 == 0x05) this modrm bytes encodes a memory operand that is computed as
+    // In case (modrm & 0xC7 == 0x05) this modrm byte encodes a memory operand that is computed as
     // the rip of the next instruction plus the 32 bit offset encoded in the four address
     // displacement bytes of the instruction.
     // See "Intel 64 and IA-32 Architectures Software Developer’s Manual Vol. 2A" Chapter 2.1.
     // Specifically table 2-2.
+    //
+    // Example of original code (add one to memory location at offset 0x123456 from rip):
+    // add [rip + 0x123456], 1       48 83 05 56 34 12 00 01
+    // The relocated instruction looks the same - we merely adjust the 0x123456 such that we address
+    // the same memory location but relative to the new code location.
     const int32_t old_displacement = *absl::bit_cast<int32_t*>(
         instruction->bytes + instruction->detail->x86.encoding.disp_offset);
     const uint64_t old_absolute_address = old_address + instruction->size + old_displacement;
@@ -222,15 +227,10 @@ ErrorMessageOr<RelocatedInstruction> RelocateInstruction(cs_insn* instruction, u
   } else if (instruction->detail->x86.opcode[0] == 0xeb ||
              instruction->detail->x86.opcode[0] == 0xe9) {
     // This handles unconditional jump to relative immediate parameter (32 bit or 8 bit).
-    // Examples of original code
-    // (endless loop doing nothing)
-    // nop                          90
-    // jmp 0xfc                     eb fc
-    // (jump to a 32 bit offset)
+    // Example of original code (jump to a 32 bit offset):
     // jmp 0x01020304               e9 01 20 03 04
-    //
-    // In both cases we compute the absolute address of the jump target, store it in memory in the
-    // trampoline and jump there:
+    // In both cases (8 and 32 bit offsets) we compute the absolute address of the jump target,
+    // store it in memory in the trampoline and jump there:
     // jmp [rip + 0]                ff 25 00 00 00 00
     // .byte absolute_address       01 02 03 04 05 06 07 08
     const int32_t immediate =
@@ -241,13 +241,13 @@ ErrorMessageOr<RelocatedInstruction> RelocateInstruction(cs_insn* instruction, u
                                        instruction->detail->x86.encoding.imm_offset);
     const uint64_t absolute_address = old_address + instruction->size + immediate;
     MachineCode code;
-    code.AppendBytes({0xff, 0x25})
-        .AppendImmediate32(0x00000000)
-        .AppendImmediate64(absolute_address);
+    code.AppendBytes({0xff, 0x25}).AppendImmediate32(0).AppendImmediate64(absolute_address);
     result.code = code.GetResultAsVector();
     result.position_of_absolute_address = 6;
   } else if (instruction->detail->x86.opcode[0] == 0xe8) {
     // Call function at relative immediate parameter.
+    // Example of original code (call function at offset 0x01020304):
+    // call 0x01020304              e8 04 03 02 01
     // We compute the absolute address of the called function and call it like this:
     // call [rip+2]                 ff 15 02 00 00 00
     // jmp label;                   eb 08
@@ -265,8 +265,11 @@ ErrorMessageOr<RelocatedInstruction> RelocateInstruction(cs_insn* instruction, u
     result.position_of_absolute_address = 8;
   } else if ((instruction->detail->x86.opcode[0] & 0xf0) == 0x70) {
     // 0x7? are conditional jumps to an 8 bit immediate.
-    // We invert the condition of the jump and construct the following code sequence.
-    // j(!cc) 0x0e                  7? 0e  // 0x0e == 14 = 6 bytes jmp + 8 bytes address
+    // Example of original code (jump backwards 10 bytes if last result was not zero):
+    // jne 0xf6                     75 f6
+    // We invert the condition of the jump, compute the absolute address of the jump target and
+    // construct the following code sequence.
+    // je 0x0e                      74 0e  // 0x0e == 14 = 6 bytes jmp + 8 bytes address
     // jmp [rip + 0]                ff 25 00 00 00 00
     // .byte absolute_address       01 02 03 04 05 06 07 08
     const int8_t immediate =
@@ -283,9 +286,11 @@ ErrorMessageOr<RelocatedInstruction> RelocateInstruction(cs_insn* instruction, u
     result.position_of_absolute_address = 8;
   } else if (instruction->detail->x86.opcode[0] == 0x0f &&
              (instruction->detail->x86.opcode[1] & 0xf0) == 0x80) {
-    // 0x0f 0x8? are conditional jumps to a 32 bit immediate
+    // 0x0f 0x8? are conditional jumps to a 32 bit immediate.
+    // Example of original code (jump backwards 10 bytes if last result was not zero):
+    // jne                          0f 85 f6 ff ff ff
     // We invert the condition of the jump and construct the following code sequence.
-    // j(!cc) 0x0e                  7? 0e
+    // je 0x0e                      74 0e  // 0x0e == 14 = 6 bytes jmp + 8 bytes address
     // jmp [rip + 0]                ff 25 00 00 00 00
     // .byte absolute_address       01 02 03 04 05 06 07 08
     const int32_t immediate = *absl::bit_cast<int32_t*>(
--- a/src/UserSpaceInstrumentation/Trampoline.cpp
+++ b/src/UserSpaceInstrumentation/Trampoline.cpp
@@ -29,7 +29,7 @@ namespace {
 using absl::numbers_internal::safe_strtou64_base;
 using orbit_base::ReadFileToString;
 
-std::string InstructionBytesAsString(cs_insn* instruction) {
+[[nodiscard]] std::string InstructionBytesAsString(cs_insn* instruction) {
   std::string result;
   for (int i = 0; i < instruction->size; i++) {
     result = absl::StrCat(result, i == 0 ? absl::StrFormat("%#0.2x", instruction->bytes[i])
@@ -177,13 +177,15 @@ ErrorMessageOr<uint64_t> AllocateMemoryForTrampolines(pid_t pid, const AddressRa
 }
 
 ErrorMessageOr<int32_t> AddressDifferenceAsInt32(uint64_t a, uint64_t b) {
-  const uint64_t abs_diff = (a > b) ? (a - b) : (b - a);
   constexpr uint64_t kAbsMaxInt32AsUint64 =
       static_cast<uint64_t>(std::numeric_limits<int32_t>::max());
   constexpr uint64_t kAbsMinInt32AsUint64 =
       static_cast<uint64_t>(-static_cast<int64_t>(std::numeric_limits<int32_t>::min()));
-  if ((a > b && abs_diff > kAbsMaxInt32AsUint64) || (b > a && abs_diff > kAbsMinInt32AsUint64)) {
-    return ErrorMessage("Difference is larger than +-2GB.");
+  if ((a > b) && (a - b > kAbsMaxInt32AsUint64)) {
+    return ErrorMessage("Difference is larger than +2GB.");
+  }
+  if ((b > a) && (b - a > kAbsMinInt32AsUint64)) {
+    return ErrorMessage("Difference is larger than -2GB.");
   }
   return a - b;
 }
--- a/src/UserSpaceInstrumentation/Trampoline.cpp
+++ b/src/UserSpaceInstrumentation/Trampoline.cpp
@@ -4,6 +4,7 @@
 
 #include "Trampoline.h"
 
+#include <absl/base/casts.h>
 #include <absl/strings/numbers.h>
 #include <absl/strings/str_format.h>
 #include <absl/strings/str_split.h>
@@ -13,6 +14,7 @@
 #include <string>
 
 #include "AllocateInTracee.h"
+#include "MachineCode.h"
 #include "OrbitBase/ExecuteCommand.h"
 #include "OrbitBase/GetProcessIds.h"
 #include "OrbitBase/Logging.h"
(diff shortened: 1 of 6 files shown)


src/UserSpaceInstrumentation/Trampoline.cpp
```diff
@@ -161,4 +176,124 @@ ErrorMessageOr<uint64_t> AllocateMemoryForTrampolines(pid_t pid, const AddressRa
...
+          "While trying to relocate an instruction with rip relative addressing the target was out "
+          "of range from the trampoline. old address: %#x, new address :%#x instruction: %s",
+          old_address, new_address, InstructionBytesAsString(instruction)));
+    }
+    result.code.resize(instruction->size);
+    memcpy(result.code.data(), instruction->bytes, instruction->size);
+    *absl::bit_cast<int32_t*>(result.code.data() + instruction->detail->x86.encoding.disp_offset) =
+        new_displacement_or_error.value();
+  } else if (instruction->detail->x86.opcode[0] == 0xeb ||
+             instruction->detail->x86.opcode[0] == 0xe9) {
```
dpallotti, line 214: For all the branches, I would suggest also showing the assembly of the instruction (or one example if the branch covers multiple opcodes) before the relocation, in addition to the description. I think it helps understanding the different cases.

src/UserSpaceInstrumentation/Trampoline.cpp
```diff
@@ -161,4 +176,124 @@ ErrorMessageOr<uint64_t> AllocateMemoryForTrampolines(pid_t pid, const AddressRa
...
+    code.AppendBytes({0xff, 0x15})
+        .AppendImmediate32(0x00000002)
+        .AppendBytes({0xeb, 0x08})
+        .AppendImmediate64(absolute_address);
+    result.code = code.GetResultAsVector();
+    result.position_of_absolute_address = 8;
+  } else if ((instruction->detail->x86.opcode[0] & 0xf0) == 0x70) {
+    // 0x7? are conditional jumps to an 8 bit immediate.
+    // We invert the condition of the jump and construct the following code sequence.
+    // j(!cc) 0x0e                  7? 0e
```
dpallotti, line 250: So 0x0e makes you jump after the absolute_address, yes?

src/UserSpaceInstrumentation/TrampolineTest.cpp
```diff
@@ -257,4 +264,213 @@ TEST(TrampolineTest, AllocateMemoryForTrampolines) {
   waitpid(pid, nullptr, 0);
 }
 
+TEST(TrampolineTest, AddressDifferenceAsInt32) {
+  // Result of the difference is negative; in the first case it just fits the second case overflows.
```
dpallotti, line 268:
```suggestion
  // Result of the difference is negative; in the first case it just fits, the second case overflows.
```

src/UserSpaceInstrumentation/TrampolineTest.cpp
```diff
@@ -257,4 +264,213 @@ TEST(TrampolineTest, AllocateMemoryForTrampolines) {
...
+  const uint64_t kAddr1 = 0x6012345612345678;
+  const uint64_t kAddr2Larger = kAddr1 - std::numeric_limits<int32_t>::min();
+  auto result = AddressDifferenceAsInt32(kAddr1, kAddr2Larger);
+  ASSERT_THAT(result, HasNoError());
+  EXPECT_EQ(std::numeric_limits<int32_t>::min(), result.value());
+  result = AddressDifferenceAsInt32(kAddr1, kAddr2Larger + 1);
+  EXPECT_THAT(result, HasError("Difference is larger than +-2GB"));
+
+  // Result of the difference is positive; in the first case it just fits the second case overflows.
```
dpallotti, line 277:
```suggestion
  // Result of the difference is positive; in the first case it just fits, the second case overflows.
```

src/UserSpaceInstrumentation/TrampolineTest.cpp
```diff
@@ -257,4 +264,213 @@ TEST(TrampolineTest, AllocateMemoryForTrampolines) {
...
+  Disassemble(code);
+
+  ErrorMessageOr<RelocatedInstruction> result =
+      RelocateInstruction(instruction_, 0x0100000000, 0x0200000000);
+  ASSERT_THAT(result, HasValue());
+  EXPECT_THAT(result.value().code, ElementsAreArray({0x90}));
+  EXPECT_FALSE(result.value().position_of_absolute_address.has_value());
+}
+
+TEST_F(RelocateInstructionTest, RipRelativeAddressing) {
```
dpallotti, line 335: For all these tests, could you make it (even) more immediate to match what branch they refer to. I would suggest just a comment with the content of the "if".

src/UserSpaceInstrumentation/TrampolineTest.cpp
```diff
@@ -257,4 +264,213 @@ TEST(TrampolineTest, AllocateMemoryForTrampolines) {
...
+  constexpr int32_t kOffset = 0x969433;
+  // add qword ptr [rip + kOffset], 1
+  code.AppendBytes({0x48, 0x83, 0x05}).AppendImmediate32(kOffset).AppendBytes({0x01});
+  Disassemble(code);
+
+  uint64_t kOriginalAddress = 0x0100000000;
+  ErrorMessageOr<RelocatedInstruction> result =
+      RelocateInstruction(instruction_, kOriginalAddress, kOriginalAddress + kOffset - 0x123456);
+  ASSERT_THAT(result, HasValue());
+  EXPECT_THAT(result.value().code,
```
dpallotti, line 346: Could you write the disassembly of this and the next `EXPECT_THAT` (like in the other cases)?

src/UserSpaceInstrumentation/Trampoline.cpp
```diff
@@ -161,4 +176,124 @@ ErrorMessageOr<uint64_t> AllocateMemoryForTrampolines(pid_t pid, const AddressRa
...
+    // .byte absolute_address       01 02 03 04 05 06 07 08
+    const int32_t immediate =
+        (instruction->detail->x86.opcode[0] == 0xe9)
+            ? *absl::bit_cast<int32_t*>(instruction->bytes +
+                                        instruction->detail->x86.encoding.imm_offset)
+            : *absl::bit_cast<int8_t*>(instruction->bytes +
+                                       instruction->detail->x86.encoding.imm_offset);
+    const uint64_t absolute_address = old_address + instruction->size + immediate;
+    MachineCode code;
+    code.AppendBytes({0xff, 0x25}).AppendImmediate32(0x0000000).AppendImmediate64(absolute_address);
```
dpallotti, line 227:
```suggestion
    code.AppendBytes({0xff, 0x25}).AppendImmediate32(0x00000000).AppendImmediate64(absolute_address);
```

src/UserSpaceInstrumentation/Trampoline.cpp
```diff
@@ -161,4 +176,124 @@ ErrorMessageOr<uint64_t> AllocateMemoryForTrampolines(pid_t pid, const AddressRa
...
+    result.code = code.GetResultAsVector();
+    result.position_of_absolute_address = 6;
+  } else if (instruction->detail->x86.opcode[0] == 0xe8) {
+    // Call function at relative immediate parameter.
+    // We compute the absolute address of the called function and call it like this:
+    // Call [rip+2]                 ff 15 02 00 00 00
```
dpallotti, line 233:
```suggestion
    // call [rip+2]                 ff 15 02 00 00 00
```

src/UserSpaceInstrumentation/TrampolineTest.cpp
```diff
@@ -257,4 +264,213 @@ TEST(TrampolineTest, AllocateMemoryForTrampolines) {
...
+  MachineCode code;
+  constexpr int32_t kOffset = 0x01020304;
+  // call [rip + kOffset]
+  code.AppendBytes({0xe8}).AppendImmediate32(kOffset);
+  Disassemble(code);
+
+  ErrorMessageOr<RelocatedInstruction> result =
+      RelocateInstruction(instruction_, 0x0100000000, 0x0200000000);
+  ASSERT_THAT(result, HasValue());
+  // Call [rip + 2]               ff 15 02 00 00 00
```
dpallotti, line 412:
```suggestion
  // call [rip + 2]               ff 15 02 00 00 00
```

src/UserSpaceInstrumentation/TrampolineTest.cpp
```diff
@@ -257,4 +264,213 @@ TEST(TrampolineTest, AllocateMemoryForTrampolines) {
...
+  // jmp [rip + kOffset]
+  code.AppendBytes({0xe9}).AppendImmediate32(kOffset);
+  Disassemble(code);
+
+  ErrorMessageOr<RelocatedInstruction> result =
+      RelocateInstruction(instruction_, 0x0100000000, 0x0200000000);
+  ASSERT_THAT(result, HasValue());
+  // jmp  [rip + 0]               ff 25 00 00 00 00
+  // absolute_address             09 03 02 01 01 00 00 00
+  // original jump instruction end on 0x0100000000 + 0x05. Adding kOffset yields 0x0101020309.
```
dpallotti, line 395:
```suggestion
  // original jump instruction ends on 0x0100000000 + 0x05. Adding kOffset yields 0x0101020309.
```

dpallotti (commented):
Just some comments on (possibly adding some) comments.

src/UserSpaceInstrumentation/Trampoline.cpp
danielfenner: I added examples of original code .... please take a look if this solves the problem.

src/UserSpaceInstrumentation/Trampoline.cpp

src/UserSpaceInstrumentation/Trampoline.cpp

src/UserSpaceInstrumentation/Trampoline.cpp
danielfenner: exactly, added a comment.

src/UserSpaceInstrumentation/Trampoline.h
```diff
@@ -58,6 +58,57 @@ namespace orbit_user_space_instrumentation {
...
                                                                     uint64_t size);
 
+// Returns the signed 32 bit difference (a-b) between two absolute virtual 64 bit addresses or an
+// error if the difference is too large.
+[[nodiscard]] ErrorMessageOr<int32_t> AddressDifferenceAsInt32(uint64_t a, uint64_t b);
+
+// Merely serves as a return value for the function below.
+struct RelocatedInstruction {
+  // Machine code of the relocated instruction. Might contain multiple instructions to emulate what
+  // the original instruction acchieved.
```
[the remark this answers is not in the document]
danielfenner: done

src/UserSpaceInstrumentation/Trampoline.h
```diff
@@ -58,6 +58,57 @@ namespace orbit_user_space_instrumentation {
...
+  std::vector<uint8_t> code;
+
+  // Some relocated instructions contain an absolute address that needs to be adjusted once all the
+  // relocations are done. The position of this absolute address is stored here.
+  // Example: A conditional jump to a forward position needs to know the position of a instruction
```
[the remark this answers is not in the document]
danielfenner: done

src/UserSpaceInstrumentation/TrampolineTest.cpp
danielfenner: done

src/UserSpaceInstrumentation/TrampolineTest.cpp
danielfenner: done

src/UserSpaceInstrumentation/TrampolineTest.cpp
danielfenner: done

src/UserSpaceInstrumentation/TrampolineTest.cpp
danielfenner: done

src/UserSpaceInstrumentation/Trampoline.h
```diff
@@ -58,6 +58,57 @@ namespace orbit_user_space_instrumentation {
...
+  // 0x0200: j(!cc) rip+08 (== 0x0210)
+  // 0x0202: jmp [rip+0] (== [0x0208])
+  // 0x0208: 8 byte destination address == address of relocated InstructionB == 0x0217
+  // 0x0210: InstructionA'
+  // 0x0217: InstructionB'
+  //
+  // The conditional jump at 0x0100 is translated into the first three lines of the result. The
+  // address (at 0x0208) of InstructionB' is not yet known at the point of the translation. So it
+  // needs to be recorded and handled later. In this case the `position_of_absolute_address` below
+  // would be 8.
```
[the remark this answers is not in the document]
danielfenner: Yes, or even more straightforward: It is the position in the 'code' stored above. 

Added comment - please have a look if this makes sense.

src/UserSpaceInstrumentation/TrampolineTest.cpp
danielfenner: Added comments and moved trivial case to the bottom so they occur in the same order as in 'RelocateInstruction'

src/UserSpaceInstrumentation/TrampolineTest.cpp
danielfenner: done

danielfenner (commented):
@dpallotti: Thanks for the review! I left a few comments where I added a bunch of explanations unresolved - can you have a look if this makes more sense now?

src/UserSpaceInstrumentation/Trampoline.h
```diff
@@ -58,6 +58,58 @@ namespace orbit_user_space_instrumentation {
...
+  std::optional<size_t> position_of_absolute_address = std::nullopt;
```
pierricgimmig, line 96: In the case of a conditional jump, there is always only one absolute address required right? That's because if the conditional jump finishes at the boundary of the overwritten 5 bytes, the path where the jump is not taken will automatically continue to the original code. In the case where the relative jump finishes before the 5 bytes boundary, then the following instruction is also in the trampoline, right? Maybe add a comment?

pierricgimmig (approved):
Can't wait to try it out! Thanks for this Daniel, LGTM, please see minor comments.

src/UserSpaceInstrumentation/Trampoline.h
danielfenner: I think I don't get the point. 

If you have a conditional jump you get an 8 bit or 32 bit offset. From that you compute the absolute address of the jump target and that is what is written into `code`. And the index of this is stored in `position_of_absolute_address`; it needs to be updated later in case the jump target is also relocated into the trampoline.

Independently from that: If the jump is not taken we just execute the next instruction (as the original conditional jump would do). This is either another relocated instruction or the jump at the end of the trampoline.

src/UserSpaceInstrumentation/Trampoline.cpp
```diff
@@ -161,4 +176,143 @@ ErrorMessageOr<uint64_t> AllocateMemoryForTrampolines(pid_t pid, const AddressRa
...
+ErrorMessageOr<RelocatedInstruction> RelocateInstruction(cs_insn* instruction, uint64_t old_address,
+                                                         uint64_t new_address) {
+  RelocatedInstruction result;
+  if ((instruction->detail->x86.modrm & 0xC7) == 0x05) {
+    // The encoding of an x86 instruction contains instruction prefixes, an opcode, the modrm and
+    // sib bytes, 1, 2 or 4 bytes address displacement and 1, 2 or 4 bytes of immediate data.
+    // Most of these are optional - at least one byte of opcode needs to be present.
+    // Many instructions that refer to an operand in memory have an addressing-form specifier byte
+    // (called the modrm byte) following the primary opcode.
+    // In case (modrm & 0xC7 == 0x05) this modrm bytes encodes a memory operand that is computed as
```
dpallotti, line 202:
```suggestion
    // In case (modrm & 0xC7 == 0x05) this modrm byte encodes a memory operand that is computed as
```

src/UserSpaceInstrumentation/Trampoline.cpp
```diff
@@ -161,4 +176,143 @@ ErrorMessageOr<uint64_t> AllocateMemoryForTrampolines(pid_t pid, const AddressRa
...
+        new_displacement_or_error.value();
+  } else if (instruction->detail->x86.opcode[0] == 0xeb ||
+             instruction->detail->x86.opcode[0] == 0xe9) {
+    // This handles unconditional jump to relative immediate parameter (32 bit or 8 bit).
+    // Examples of original code
+    // (endless loop doing nothing)
+    // nop                          90
+    // jmp 0xfc                     eb fc
+    // (jump to a 32 bit offset)
+    // jmp 0x01020304               e9 01 20 03 04
```
dpallotti, line 230: I would make it clearer these are two separate examples, I was somehow confused at the beginning.

src/UserSpaceInstrumentation/Trampoline.cpp
dpallotti: I was thinking of the same also for the branches below, or do you think it doesn't really make sense/it's already enough in the tests?

src/UserSpaceInstrumentation/Trampoline.cpp
```diff
@@ -161,4 +176,143 @@ ErrorMessageOr<uint64_t> AllocateMemoryForTrampolines(pid_t pid, const AddressRa
...
+    // is "jump if not equal").
+    const uint8_t opcode = 0x01 ^ instruction->detail->x86.opcode[0];
+    code.AppendBytes({opcode, 0x0e})
+        .AppendBytes({0xff, 0x25, 0x00, 0x00, 0x00, 0x00})
+        .AppendImmediate64(absolute_address);
+    result.code = code.GetResultAsVector();
+    result.position_of_absolute_address = 8;
+  } else if (instruction->detail->x86.opcode[0] == 0x0f &&
+             (instruction->detail->x86.opcode[1] & 0xf0) == 0x80) {
+    // 0x0f 0x8? are conditional jumps to a 32 bit immediate
```
dpallotti, line 286:
```suggestion
    // 0x0f 0x8? are conditional jumps to a 32 bit immediate.
```

dpallotti (approved):
Thanks!

dimitry- (approved):
nit: I still think 0 is more readable than 0x0000000000

src/UserSpaceInstrumentation/Trampoline.cpp
danielfenner: I just removed one of the examples. The difference between 8 and 32 bit offsets shpuld be small enough.

PR #2379, opened 2021-05-31
Move files to new more visible location

dimitry-:
On Orbit start move captures and presets from the old
    location to the new one. From now on we save presets
    and capture files in %HOME%/Documents/Google/Orbit
    
    Test: Start orbit - see that it moved files to new location
    Bug: http://b/159605581

### DIFF
--- a/src/Orbit/main.cpp
+++ b/src/Orbit/main.cpp
@@ -21,14 +21,15 @@
 #include <cstdint>
 #include <filesystem>
 #include <optional>
-#include <outcome.hpp>
 #include <string>
 #include <system_error>
 #include <utility>
-#include <variant>
 #include <vector>
 
 #include "MetricsUploader/ScopedMetric.h"
+#include "OrbitBase/File.h"
+#include "OrbitQt/MoveFilesDialog.h"
+#include "OrbitQt/MoveFilesProcess.h"
 
 #ifdef _WIN32
 #include <process.h>
@@ -255,6 +256,59 @@ static void LogAndMaybeWarnAboutClockResolution() {
   }
 }
 
+static bool IsDirectoryEmpty(const std::filesystem::path& directory) {
+  auto exists_or_error = orbit_base::FileExists(directory);
+  if (exists_or_error.has_error()) {
+    ERROR("Unable to stat \"%s\": %s", directory.string(), exists_or_error.error().message());
+    return false;
+  }
+
+  if (!exists_or_error.value()) {
+    return true;
+  }
+
+  auto file_list_or_error = orbit_base::ListFilesInDirectory(directory);
+  if (file_list_or_error.has_error()) {
+    ERROR("Unable to list directory \"%s\": %s", directory.string(),
+          file_list_or_error.error().message());
+    return false;
+  }
+
+  return file_list_or_error.value().empty();
+}
+
+static void TryMoveSavedDataLocationIfNeeded() {
+  if (IsDirectoryEmpty(orbit_core::GetPresetDirPriorTo1_65()) &&
+      IsDirectoryEmpty(orbit_core::GetCaptureDirPriorTo1_65())) {
+    return;
+  }
+
+  orbit_qt::MoveFilesDialog dialog;
+  orbit_qt::MoveFilesProcess process;
+
+  QObject::connect(&process, &orbit_qt::MoveFilesProcess::generalError,
+                   [&dialog](const std::string& error_message) {
+                     dialog.AddText(absl::StrFormat("Error: %s", error_message));
+                   });
+
+  QObject::connect(
+      &process, &orbit_qt::MoveFilesProcess::moveStarted,
+      [&dialog](const std::filesystem::path& from_dir, const std::filesystem::path& to_dir,
+                size_t number_of_files) {
+        dialog.AddText(absl::StrFormat(R"(Moving %d files from "%s" to "%s" ...)", number_of_files,
+                                       from_dir.string(), to_dir.string()));
+      });
+
+  QObject::connect(&process, &orbit_qt::MoveFilesProcess::moveDone,
+                   [&dialog]() { dialog.AddText("... done"); });
+
+  QObject::connect(&process, &orbit_qt::MoveFilesProcess::processFinished,
+                   [&dialog]() { dialog.EnableCloseButton(); });
+
+  process.Start();
+  dialog.exec();
+}
+
 // Removes all source paths mappings from the persistent settings storage.
 static void ClearSourcePathsMappings() {
   orbit_source_paths_mapping::MappingManager mapping_manager{};
@@ -375,6 +429,8 @@ int main(int argc, char* argv[]) {
   }
 
   LogAndMaybeWarnAboutClockResolution();
+
+  TryMoveSavedDataLocationIfNeeded();
 
   const DeploymentConfiguration deployment_configuration = FigureOutDeploymentConfiguration();
 
--- a/src/OrbitCore/Path.cpp
+++ b/src/OrbitCore/Path.cpp
@@ -60,14 +60,18 @@ std::filesystem::path CreateOrGetCacheDir() {
   return cache_dir;
 }
 
+std::filesystem::path GetPresetDirPriorTo1_65() { return CreateOrGetOrbitAppDataDir() / "presets"; }
+
+std::filesystem::path GetCaptureDirPriorTo1_65() { return CreateOrGetOrbitAppDataDir() / "output"; }
+
 std::filesystem::path CreateOrGetPresetDir() {
-  std::filesystem::path preset_dir = CreateOrGetOrbitAppDataDir() / "presets";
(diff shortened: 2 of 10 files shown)


dimitry-:
first commit is not what it should look like - working on it.

dimitry-:
this is ready for review - ptal.

ronaldfw:
Our AppData folder is called "OrbitProfiler". While this is not really visible to users, I feel that the naming should be consistent. I also think the extra "Google" folder is just extra complication for users. So, I'd suggest that we save files in %HOME%/Documents/OrbitProfiler

dimitry-:
> Our AppData folder is called "OrbitProfiler". While this is not really visible to users, I feel that the naming should be consistent. I also think the extra "Google" folder is just extra complication for users. So, I'd suggest that we save files in %HOME%/Documents/OrbitProfiler

I think orbitprofiler is there from before we renamed the project from "ORBIT Profiler" to "Orbit".

ronaldfw:
> I think orbitprofiler is there from before we renamed the project from "ORBIT Profiler" to "Orbit".

Then maybe just "Orbit" directly under Documents?

dimitry-:
> > I think orbitprofiler is there from before we renamed the project from "ORBIT Profiler" to "Orbit".
> 
> Then maybe just "Orbit" directly under Documents?

done.

src/OrbitQt/include/OrbitQt/MoveFilesProcess.h
```diff
@@ -0,0 +1,42 @@
...
+namespace orbit_qt {
+
+class MoveFilesProcess : public QObject {
+  Q_OBJECT
+
+ public:
+  explicit MoveFilesProcess();
+  ~MoveFilesProcess() noexcept override;
+
+  void Run();
```
beckerhe, line 21: Nit: To me `Run` sounds like a synchronous function. I would prefer either `Start` or something like `RunAsync`.

src/OrbitQt/MoveFilesProcess.cpp
```diff
@@ -0,0 +1,83 @@
...
+  if (file_exists_or_error.has_error()) {
+    ReportError(absl::StrFormat("Unable to stat \"%s\": %s", src_dir.string(),
+                                file_exists_or_error.error().message()));
+  } else {
+    file_exists = file_exists_or_error.value();
+  }
+
+  if (!file_exists) {
+    return;
+  }
```
beckerhe, line 45: what about `else if (!file_exists_or_error.value()) { return; }`

This would avoid the extra `bool file_exists`

beckerhe: Or is there a return missing in the `ReportError` case above?

src/OrbitQt/MoveFilesDialog.cpp
```diff
@@ -0,0 +1,42 @@
...
+#include <QPlainTextEdit>
+#include <QScrollBar>
+
+#include "ui_MoveFilesDialog.h"
+
+namespace orbit_qt {
+
+namespace {
+
+class MoveFilesDialogImpl : public MoveFilesDialog {
```
beckerhe, line 17: Why an Impl class?

src/OrbitQt/MoveFilesDialog.cpp
```diff
@@ -0,0 +1,42 @@
+// Copyright (c) 2021 The Orbit Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#include "OrbitQt/MoveFilesDialog.h"
+
+#include <QLabel>
+#include <QPlainTextEdit>
+#include <QScrollBar>
```
beckerhe, line 9: Is QLabel and QScrollbar used in the code below? I can't spot them. But QString should probably be here.

src/Orbit/main.cpp
```diff
@@ -255,6 +256,59 @@ static void LogAndMaybeWarnAboutClockResolution() {
...
+      });
+
+  QObject::connect(&process, &orbit_qt::MoveFilesProcess::moveDone,
+                   [&dialog]() { dialog->AddText("... done"); });
+
+  QObject::connect(&process, &orbit_qt::MoveFilesProcess::processFinished,
+                   [&dialog]() { dialog->EnableCloseButton(); });
+
+  process.Run();
+  dialog->exec();
```
beckerhe, line 309: What happens when the user presses Escape? Will this close the dialog while moving files is still in progress? I would assume that the dialog closes, `process`'s destructor will block in that case and Orbit's UI will be frozen until moving files is done.

dimitry-: this is a bit hard to test - the process is very fast since when we move on the same device it effectively renames. I will try to insert some sleeps and see what happens. I expect the same. Btw UI is not yet up at this point so the user will see nothing until the process is done.

src/OrbitQt/MoveFilesDialog.cpp
dimitry-: well.. It started when I decided to introduce include folder to OrbitQt, once I did that CMake operating under assumption that .ui file must me in the same folder as file that includes it, refused to compile the header. I have decided to move include to .cpp file which led to this (by Impl we avoid mentioning Ui::MoveFilesDialog in the header file).

src/OrbitCore/Path.cpp
```diff
@@ -88,6 +92,18 @@ std::filesystem::path CreateOrGetOrbitAppDataDir() {
   return path;
 }
 
+std::filesystem::path CreateOrGetOrbitUserDataDir() {
+#ifdef WIN32
+  std::filesystem::path path = std::filesystem::path(GetEnvVar("USERPROFILE"));
+#else
+  std::filesystem::path path = std::filesystem::path(GetEnvVar("HOME"));
+#endif
+  path = path / "Documents" / "Orbit";
```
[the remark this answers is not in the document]
dimitry-: what would be correct way to detect Documents folder? is there another environment variable for it?
(we may want to make it configurable, but this is probably for separate PR)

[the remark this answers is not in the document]
beckerhe: AFAIK https://docs.microsoft.com/en-us/windows/win32/api/shlobj_core/nf-shlobj_core-shgetfolderpatha should be the correct Win32 API function. Alternatively you could also read it from the registry or use Qt's https://doc.qt.io/qt-5/qstandardpaths.html. The latter is probably easiest if you can inject the path and not use a free function on OrbitCore to determine it.

src/OrbitQt/MoveFilesDialog.cpp
beckerhe: Ah, I see. In Qt you usually forward declare `Ui::MoveFilesDialog` in the header and only include `ui_MoveFilesDialog.cpp` in the cpp file. For that to work you can't have a compiler-generated destructor. You have to define it explicitly in the cpp file, for example by defaulting it: `MoveFilesDialog::~MoveFilesDialog() = default;`.

But I guess it doesn't make a huge difference. I personally would prefer the solution without Impl, but feel free to ignore.

src/OrbitCore/Path.cpp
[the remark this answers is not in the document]
dimitry-: I think I am going to leave it for latter PRs. It looks like this can be done easier if we moved from global functions to a Locations object that initialized and passed around and this requires some refactoring.

vickyliu-go4it (approved):
LGTM now! Thanks for that.

src/Orbit/main.cpp
beckerhe: Should we do something about it? We could override the `closeEvent` method (https://doc.qt.io/qt-5/qwidget.html#closeEvent) and ignore the event if we aren't done yet.

beckerhe (approved):
Approving now since only nits are left.

src/Orbit/main.cpp
dimitry-: I must be doing something wrong. closeEvent is not getting triggered on escape.

src/OrbitQt/MoveFilesDialog.cpp
```diff
@@ -0,0 +1,34 @@
...
+
+MoveFilesDialog::~MoveFilesDialog() noexcept = default;
+
+void MoveFilesDialog::AddText(std::string_view text) {
+  ui_->log->append(QString::fromUtf8(text.data(), text.size()));
+}
+
+void MoveFilesDialog::EnableCloseButton() { ui_->closeButton->setEnabled(true); }
+
+void MoveFilesDialog::closeEvent(QCloseEvent* event) {
```
dimitry-, line 26: When I debug it - this function in never called.

beckerhe: Hmmm. Weird. I was expecting this should trigger on all kinds closes. I don't have a better suggestion though. So feel free to ignore me here. :sweat_smile:

dimitry-: I removed it for now.

PR #2380, opened 2021-05-31
Show existing files in CSV export

reichlfl:
This will correctly configure csv saving functionality:

* Existing files are shown
* When overwriting an existing file, a confirmation dialog is displayed

b/167546024
To test:
* Export any DataView to csv
* In the "Save File" dialog, observe that existing csv files in the same folder are displayed
* When overwriting an existing file, Orbit asks for confirmation

### DIFF
--- a/src/OrbitQt/orbitmainwindow.cpp
+++ b/src/OrbitQt/orbitmainwindow.cpp
@@ -729,9 +729,17 @@ void OrbitMainWindow::OnNewSelectionBottomUpView(
 }
 
 std::string OrbitMainWindow::OnGetSaveFileName(const std::string& extension) {
-  std::string filename =
-      QFileDialog::getSaveFileName(this, "Specify a file to save...", nullptr, extension.c_str())
-          .toStdString();
+  QFileDialog dialog(this);
+  dialog.setFileMode(QFileDialog::FileMode::AnyFile);
+  dialog.setAcceptMode(QFileDialog::AcceptMode::AcceptSave);
+  dialog.setNameFilter(QString::fromStdString(extension + " (*" + extension + ")"));
+  dialog.setWindowTitle("Specify a file to save...");
+  dialog.setDirectory(nullptr);
+  std::string filename;
+
+  if (dialog.exec() == QDialog::Accepted && dialog.selectedFiles().size() > 0) {
+    filename = dialog.selectedFiles()[0].toStdString();
+  }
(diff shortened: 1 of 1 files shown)


src/OrbitQt/orbitmainwindow.cpp
```diff
@@ -729,9 +729,17 @@ void OrbitMainWindow::OnNewSelectionBottomUpView(
 }
 
 std::string OrbitMainWindow::OnGetSaveFileName(const std::string& extension) {
-  std::string filename =
-      QFileDialog::getSaveFileName(this, "Specify a file to save...", nullptr, extension.c_str())
-          .toStdString();
+  QFileDialog dialog(this);
+  dialog.setFileMode(QFileDialog::FileMode::AnyFile);
+  dialog.setAcceptMode(QFileDialog::AcceptMode::AcceptSave);
+  dialog.setNameFilter((extension + " (*" + extension + ")").c_str());
```
beckerhe, line 735: We should use QString::fromStdString instead of std::string::c_str since the former is more efficient and safer.

src/OrbitQt/orbitmainwindow.cpp
```diff
@@ -729,9 +729,17 @@ void OrbitMainWindow::OnNewSelectionBottomUpView(
...
+  dialog.setWindowTitle("Specify a file to save...");
+  dialog.setDirectory(nullptr);
+  std::string filename;
+
+  if (dialog.exec() != 0 && dialog.selectedFiles().size() > 0) {
```
beckerhe, line 740: Nit: `QDialog::Rejected` instead of `0` or `dialog.exec() == QDialog::Accepted` (as you prefer).

src/OrbitQt/orbitmainwindow.cpp
beckerhe: And you can also consider to use format strings like `QString{"%1 (*%1)"}.arg(QString::fromStdString(extension))`. Looks slightly more readable to me, but I don't feel strongly about it.

beckerhe (approved):
Just minor things. Feel free to ignore. LGTM.

src/OrbitQt/orbitmainwindow.cpp
reichlfl: Changed, thanks. I've not updated this to use the format string though, I actually prefer the other one readability-wise.

PR #2381, opened 2021-06-01
Add CaptureFileInfo class and test

antonrohr:
This is the start of a series of PRs for the new "recent capture files" features. 

In case you want to see where this is going, I am actively working in [this](https://github.com/antonrohr/orbit/tree/recent_capture_files) branch in my fork.

### DIFF
--- a/src/CaptureFileInfo/include/CaptureFileInfo/CaptureFileInfo.h
+++ b/src/CaptureFileInfo/include/CaptureFileInfo/CaptureFileInfo.h
@@ -1,0 +1,37 @@
+// Copyright (c) 2021 The Orbit Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#ifndef CAPTURE_FILE_INFO_CAPTURE_FILE_INFO_H_
+#define CAPTURE_FILE_INFO_CAPTURE_FILE_INFO_H_
+
+#include <QDateTime>
+#include <QFileInfo>
+#include <filesystem>
+#include <utility>
+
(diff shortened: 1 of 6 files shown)


src/CaptureFileInfo/include/CaptureFileInfo/CaptureFileInfo.h
```diff
@@ -0,0 +1,39 @@
...
+
+  [[nodiscard]] QDateTime LastUsed() const { return last_used_; }
+  [[nodiscard]] QDateTime Created() const { return file_info_.birthTime(); }
+
+  [[nodiscard]] bool FileExists() const { return file_info_.exists() && file_info_.isFile(); }
+
+  void Touch() { last_used_ = QDateTime::currentDateTime(); }
+
+ private:
+  QFileInfo file_info_;
```
dimitry-, line 33: out of curiosity - how does it handle io errors?

dimitry- (approved):

src/CaptureFileInfo/include/CaptureFileInfo/CaptureFileInfo.h
antonrohr: I dont know. I googled and I did not find a lot of information. I assume when a call to (for example) `.exists()` fails it just returns false. In this case maybe `QFileDevice::error()` [link](https://doc.qt.io/qt-5/qfiledevice.html#error) returns the error code why it failed. This would be corresponding to the `std::filesystem::exists(path, error_code)` call. But this is only my best guess, I dont know for sure.

PR #2382, opened 2021-06-01
Switch capture state to kStarted after recieving CaptureStarted event

dimitry-:
This fixes the problem with inconsistent CaptureClient state introduced
in 0f389fe00bac44a36287e31ef8c5f517fd09f398.

With this change Capture States switches from kStarting to kStarted
after CaptureEvent event is processed.

Bug: http://b/189803997
Test: introduce a delay on the service side, click start and stop
      capture very fast. Check that it does not crash.

### DIFF
--- a/src/CaptureClient/CaptureClient.cpp
+++ b/src/CaptureClient/CaptureClient.cpp
@@ -38,6 +38,7 @@ using orbit_grpc_protos::ApiFunction;
 using orbit_grpc_protos::CaptureOptions;
 using orbit_grpc_protos::CaptureRequest;
 using orbit_grpc_protos::CaptureResponse;
+using orbit_grpc_protos::ClientCaptureEvent;
 using orbit_grpc_protos::InstrumentedFunction;
 using orbit_grpc_protos::TracepointInfo;
 using orbit_grpc_protos::UnwindingMethod;
@@ -84,6 +85,7 @@ Future<ErrorMessageOr<CaptureListener::CaptureOutcome>> CaptureClient::Capture(
   }
 
   state_ = State::kStarting;
+  LOG("State is now kStarting");
 
   auto capture_result = thread_pool->Schedule(
       [this, process_id, &module_manager, selected_functions = std::move(selected_functions),
@@ -219,11 +221,6 @@ ErrorMessageOr<CaptureListener::CaptureOutcome> CaptureClient::CaptureSync(
   }
   LOG("Sent CaptureRequest on Capture's gRPC stream: asking to start capturing");
 
-  {
-    absl::MutexLock lock{&state_mutex_};
-    state_ = State::kStarted;
-  }
-
   while (!writes_done_failed_ && !try_abort_) {
     CaptureResponse response;
     bool read_succeeded;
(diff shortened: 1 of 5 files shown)


src/CaptureClient/CaptureClient.cpp
```diff
@@ -84,6 +85,7 @@ Future<ErrorMessageOr<CaptureListener::CaptureOutcome>> CaptureClient::Capture(
   }
 
   state_ = State::kStarting;
+  LOG("State is now kStarting");
```
ronaldfw, line 88: Was this added as debug output or do you want to keep it? (Seems like it could be useful to have in the log.)

src/CaptureClient/CaptureClient.cpp
```diff
@@ -265,16 +262,16 @@ bool CaptureClient::StopCapture() {
   {
     absl::MutexLock lock(&state_mutex_);
     if (state_ == State::kStarting) {
-      // Block and wait until the state is not kStarting
-      bool (*is_not_starting)(State*) = [](State* state) { return *state != State::kStarting; };
-      state_mutex_.Await(absl::Condition(is_not_starting, &state_));
+      LOG("StopCapture ignored, because it is starting and cannot be stopped at this stage.");
```
ronaldfw, line 265: From a user perspective, this means, they have to press the button again to stop the capture, right? 

What's the reason for not keeping the wait here?

dimitry-: it leads to a deadlock. OnCaptureStarted waits on something to be executed on the main thread while main thread waits for OnCaptureStarted to complete.

src/CaptureClient/CaptureClient.cpp
dimitry-: Yep, it could be useful.

src/CaptureClient/CaptureClient.cpp
dpallotti, line 265:
```suggestion
      LOG("StopCapture ignored, because it is starting and cannot be stopped at this stage");
```

src/CaptureClient/CaptureClient.cpp
```diff
@@ -265,16 +262,16 @@ bool CaptureClient::StopCapture() {
...
+      return false;
     }
 
     if (state_ != State::kStarted) {
```
dpallotti, line 269: This condition now includes the one above and does basically the same, just with a different string. I would consider listing each state explicitly (`kStarting`, `kStopping`, `kStopped`) and printing a string dedicated to each ("it is starting and cannot be stopped", "it is already stopping", "it is already stopped").

dpallotti (approved):

src/CaptureClient/CaptureClient.cpp
ronaldfw: Got it, thanks for clarifying!

dimitry-: done.

src/CaptureClient/CaptureClient.cpp
dimitry-: done.

ronaldfw (approved):
LGTM

PR #2383, opened 2021-06-01
Add CaptureFileInfo Manager class and tests

antonrohr:
http://b/178379643

I naively named the class `Manager`, because it is in the namespace `orbit_capture_file_info` and the naming `orbit_capture_file_info::CaptureFileInfoManager` seemed a bit verbose. But the name `Manager` is very generic, so maybe its not a good idea. Please let me know what you think.

### DIFF
--- a/src/CaptureFileInfo/ManagerTest.cpp
+++ b/src/CaptureFileInfo/ManagerTest.cpp
@@ -1,0 +1,147 @@
+// Copyright (c) 2021 The Orbit Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#include <gtest/gtest.h>
+
+#include <QCoreApplication>
+#include <QDateTime>
+#include <chrono>
+#include <filesystem>
+#include <thread>
+
+#include "CaptureFileInfo/Manager.h"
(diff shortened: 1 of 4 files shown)


src/CaptureFileInfo/ManagerTest.cpp
```diff
@@ -0,0 +1,144 @@
...
+#include <thread>
+
+#include "CaptureFileInfo/Manager.h"
+#include "OrbitBase/ExecutablePath.h"
+
+namespace orbit_capture_file_info {
+
+constexpr const char* kOrgName = "The Orbit Authors";
+
+TEST(Manager, Clear) {
```
beckerhe, line 19: I think we should call the test suite `CaptureFileInfoManager` instead of `Manager` because if it fails on the CI it's hard to figure out from the log that it refers to a failing test in `CaptureFileInfo` if it just says the test `Manager.Clear` failed.

beckerhe (approved):
Sorry for being late. LGTM.

src/CaptureFileInfo/ManagerTest.cpp
antonrohr: done

PR #2384, opened 2021-06-01
Show exclusive/inclusive count in addition to % in "Sampling" tab

dpallotti:
Similarly to the top-down and bottom-up view.

Bug: http://b/189808486

Test: Update `SamplingDataPostProcessorTest`.
Capture Trata and verify the numbers make sense.

---

Plus "Renamings in OrbitQt/orbitsamplingreport".


beckerhe (approved):
Sorry for being late. LGTM

PR #2385, opened 2021-06-01
Create CaptureData object in IntrospectionWindow

pierricgimmig:
This avoids null-checking "capture_data_" in introspection code
as the variable can't be "nullptr" anymore. It also makes sense for
introspection to have it's own capture data, especially if we add
sampling and scheduling information eventually.

This also fixes a crash when using the introspection window.

### DIFF
--- a/src/OrbitGl/ThreadTrack.cpp
+++ b/src/OrbitGl/ThreadTrack.cpp
@@ -60,8 +60,7 @@ ThreadTrack::ThreadTrack(CaptureViewElement* parent, TimeGraph* time_graph,
 }
 
 std::string ThreadTrack::GetThreadNameFromTid(uint32_t thread_id) {
-  const std::string kEmptyString;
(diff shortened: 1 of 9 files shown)


src/OrbitGl/ThreadTrack.cpp
```diff
@@ -61,7 +61,7 @@ ThreadTrack::ThreadTrack(CaptureViewElement* parent, TimeGraph* time_graph,
 
 std::string ThreadTrack::GetThreadNameFromTid(uint32_t thread_id) {
   const std::string kEmptyString;
```
karupayun, line 63: We can also get rid of this variable now.

karupayun (approved):
LGTM!

src/OrbitGl/ThreadTrack.cpp
pierricgimmig: Yes, thanks.

pierricgimmig:
> LGTM! Thanks for this Pierric

Thanks @karupayun !

PR #2386, opened 2021-06-02
Block/Timer-Chain optimization

pierricgimmig:
Use std::vector instead of array so that we can allocate a block's
memory without calling the default constructor on all elements through
the use of the "reserve" method. Elements are created in place using the
emplace_back method.

Also, increased the size of the TimerChain's blocks to 64K to decrease
the number of allocations. This is the same block size we use in the 
batcher for example.

Overall, this gives a 5% improvement on capture load time but benefits any 
BlockChain/TimerChain usage.

Test: Load Dimitry's test capture data. Profile an application.

### DIFF
--- a/src/OrbitGl/TimerChain.h
+++ b/src/OrbitGl/TimerChain.h
@@ -30,19 +30,19 @@ class TimerBlock {
   explicit TimerBlock(TimerBlock* prev)
       : prev_(prev),
         next_(nullptr),
-        size_(0),
         min_timestamp_(std::numeric_limits<uint64_t>::max()),
-        max_timestamp_(std::numeric_limits<uint64_t>::min()) {}
+        max_timestamp_(std::numeric_limits<uint64_t>::min()) {
+    data_.reserve(kBlockSize);
+  }
 
   // Append a new element to the end of the block using placement-new.
   template <class... Args>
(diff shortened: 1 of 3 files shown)


dimitry- (approved):
thanks!

src/OrbitGl/TimerChain.h
```diff
@@ -14,7 +14,7 @@
 #include "OrbitBase/Logging.h"
 #include "TextBox.h"
 
-static constexpr int kBlockSize = 1024;
+static constexpr int kBlockSize = 64 * 1024;
```
dpallotti, line 17: Just one note that, if I remember correctly, in some places (some `UpdatePrimitives` I guess) we go through the blocks and, based on `min_timestamp_` and `max_timestamp_`, we either skip the content of the block completely or go through each the `TextBox` in the block to see whether they are in the time range we are interested in. This might have no actual effect on that or even be an optimization, but it affects more than just memory allocation.

dpallotti (approved):

src/OrbitGl/TimerChain.h
pierricgimmig: Good point, thanks, I'm reverting the change in block size for now to be sure.

PR #2387, opened 2021-06-02
Move orbit_qt::EventLoop into QtUtils

beckerhe:
EventLoop does not need to link against QtGUI, QtCore is enough. So by
moving this test into QtUtils the test will run on headless Windows
machines as well (like our CI machines).


antonrohr (approved):
LGTM. Thanks!

PR #2388, opened 2021-06-02
Remove clamp from CoreMath.h

beckerhe:
This function can be replaced by std::clamp.

### DIFF
--- a/src/OrbitGl/CoreMath.h
+++ b/src/OrbitGl/CoreMath.h
@@ -22,8 +22,3 @@ using Vec3i = gte::Vector3<int>;
 using Vec4i = gte::Vector4<int>;
 
 using Color = gte::Vector4<unsigned char>;
-
-template <class T>
-constexpr const T& clamp(const T& v, const T& lo, const T& hi) {
-  return std::min(std::max(v, lo), hi);
-}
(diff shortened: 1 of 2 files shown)


dimitry- (approved):

src/OrbitGl/CoreMath.h
```diff
@@ -21,9 +21,4 @@ using Vec2i = gte::Vector2<int>;
 using Vec3i = gte::Vector3<int>;
 using Vec4i = gte::Vector4<int>;
 
-using Color = gte::Vector4<unsigned char>;
-
-template <class T>
-constexpr const T& clamp(const T& v, const T& lo, const T& hi) {
-  return std::min(std::max(v, lo), hi);
-}
+using Color = gte::Vector4<unsigned char>;
```
dimitry-, line 24: seems like this file is missing include guard?

beckerhe: Hmm yes. You're right. I didn't even notice that. I will add them.

beckerhe: Ah. It has `pragma once`. I will change it to include guards

beckerhe: done.

PR #2390, opened 2021-06-02
Introduce generic Executor base class and add tests

beckerhe:
This is moving around a lot of stuff regarding executors.

1. There is now a new base class `orbit_base::Executor` which implements `Schedule`, `ScheduleAfter` and `ScheduleAfterIfSuccess`.
2. `MainThreadExecutor` and `ThreadPool` inherit from `Executor`.
3. There are now tests for Executor
4. There is now a `SimpleExecutor` which also implements `Executor` and can be used in tests. It's different to `ImmediateExecutor` in the sense that scheduled actions are stored until the user executes them manually by calling `ExecuteScheduledTasks`.

### DIFF
--- a/src/OrbitBase/ExecutorTest.cpp
+++ b/src/OrbitBase/ExecutorTest.cpp
@@ -1,0 +1,199 @@
+// Copyright (c) 2021 The Orbit Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#include <gmock/gmock.h>
+#include <gtest/gtest.h>
+
+#include "OrbitBase/Action.h"
+#include "OrbitBase/Executor.h"
+
+namespace {
+class MockExecutor : public orbit_base::Executor {
+ public:
+  MOCK_METHOD(void, ScheduleImpl, (std::unique_ptr<Action> action), (override));
+
+  [[nodiscard]] static std::shared_ptr<MockExecutor> Create() {
+    return std::make_shared<MockExecutor>();
+  }
+};
+
+// Since testing::SaveArg can't deal with move-only arguments we define our own helper here.
+template <typename T>
+auto SaveUniquePtr(std::unique_ptr<T>* destination) {
+  return [destination](std::unique_ptr<T> arg) { *destination = std::move(arg); };
+}
+}  // namespace
+
+namespace orbit_base {
+
+TEST(Executor, ScheduledTaskShouldBeCalledSimpleWithVoid) {
+  auto executor = MockExecutor::Create();
+  std::unique_ptr<Action> action;
+  EXPECT_CALL(*executor, ScheduleImpl(testing::_)).Times(1).WillOnce(SaveUniquePtr(&action));
+
+  bool called = false;
+  auto future = executor->Schedule([&called]() { called = true; });
+  ASSERT_NE(action, nullptr);
+  EXPECT_FALSE(called);
+  EXPECT_FALSE(future.IsFinished());
+  action->Execute();
+  EXPECT_TRUE(called);
+  EXPECT_TRUE(future.IsFinished());
+}
+
+TEST(Executor, ScheduledTaskShouldBeCalledSimpleWithInt) {
+  auto executor = MockExecutor::Create();
+  std::unique_ptr<Action> action;
+  EXPECT_CALL(*executor, ScheduleImpl(testing::_)).Times(1).WillOnce(SaveUniquePtr(&action));
+
+  bool called = false;
+  auto future = executor->Schedule([&called]() {
+    called = true;
+    return 42;
+  });
+
+  ASSERT_NE(action, nullptr);
+  EXPECT_FALSE(called);
+  action->Execute();
+  EXPECT_TRUE(called);
+  ASSERT_TRUE(future.IsFinished());
+  EXPECT_EQ(future.Get(), 42);
+}
+
+TEST(Executor, ChainedTaskedShouldBeCalledSimple) {
+  auto executor = MockExecutor::Create();
+  std::unique_ptr<Action> action;
+  EXPECT_CALL(*executor, ScheduleImpl(testing::_)).Times(1).WillOnce(SaveUniquePtr(&action));
+
+  Promise<void> promise{};
+  auto future = promise.GetFuture();
+  auto chained_future = executor->ScheduleAfter(future, []() {});
+  EXPECT_EQ(action, nullptr);
+  EXPECT_FALSE(chained_future.IsFinished());
+
+  promise.MarkFinished();
+  ASSERT_NE(action, nullptr);
+  action->Execute();
+  EXPECT_TRUE(chained_future.IsFinished());
+}
+
+TEST(Executor, ScheduleAfterIfSuccessShortCircuitOnErrorVoid) {
+  auto executor = MockExecutor::Create();
+  std::unique_ptr<Action> action;
+  EXPECT_CALL(*executor, ScheduleImpl(testing::_)).Times(1).WillOnce(SaveUniquePtr(&action));
+
+  Promise<ErrorMessageOr<void>> promise{};
+  auto future = promise.GetFuture();
+  auto chained_future = executor->ScheduleAfterIfSuccess(future, []() {});
+  EXPECT_FALSE(chained_future.IsFinished());
+
+  constexpr const char* const kErrorMessage{"Error"};
+  promise.SetResult(ErrorMessage{kErrorMessage});
+  ASSERT_TRUE(chained_future.IsFinished());
+  EXPECT_TRUE(chained_future.Get().has_error());
+  EXPECT_EQ(chained_future.Get().error().message(), kErrorMessage);
+
+  EXPECT_EQ(executor->GetNumberOfWaitingContinuations(), 1);
+  ASSERT_NE(action, nullptr);
+  action->Execute();
+  EXPECT_EQ(executor->GetNumberOfWaitingContinuations(), 0);
+}
+
+TEST(Executor, ScheduleAfterIfSuccessShortCircuitOnErrorInt) {
+  auto executor = MockExecutor::Create();
+  std::unique_ptr<Action> action;
+  EXPECT_CALL(*executor, ScheduleImpl(testing::_)).Times(1).WillOnce(SaveUniquePtr(&action));
+
+  Promise<ErrorMessageOr<int>> promise{};
+  auto future = promise.GetFuture();
+  auto chained_future = executor->ScheduleAfterIfSuccess(future, [](int value) {
+    EXPECT_EQ(value, 42);
+    return 1 + value;
+  });
+  EXPECT_FALSE(chained_future.IsFinished());
+
+  constexpr const char* const kErrorMessage{"Error"};
+  promise.SetResult(ErrorMessage{kErrorMessage});
+  ASSERT_TRUE(chained_future.IsFinished());
+  EXPECT_TRUE(chained_future.Get().has_error());
+  EXPECT_EQ(chained_future.Get().error().message(), kErrorMessage);
+
+  EXPECT_EQ(executor->GetNumberOfWaitingContinuations(), 1);
+  ASSERT_NE(action, nullptr);
+  action->Execute();
+  EXPECT_EQ(executor->GetNumberOfWaitingContinuations(), 0);
+}
+
+TEST(Executor, ScheduleAfterIfSuccessCallOnSuccessVoid) {
+  auto executor = MockExecutor::Create();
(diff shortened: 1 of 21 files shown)


src/OrbitBase/SimpleExecutor.cpp
```diff
@@ -0,0 +1,33 @@
...
+      action->Execute();
+      mutex_.Lock();
+    }
+    scheduled_tasks_.pop_front();
+  }
+}
+
+std::shared_ptr<SimpleExecutor> SimpleExecutor::Create() {
+  // NOLINTNEXTLINE
+  return std::shared_ptr<SimpleExecutor>{new SimpleExecutor{}};
```
antonrohr, line 31: Why not `std::make_shared`?

src/OrbitBase/SimpleExecutorTest.cpp
```diff
@@ -0,0 +1,127 @@
...
+  bool called = false;
+  Promise<ErrorMessageOr<void>> promise{};
+  auto future = promise.GetFuture();
+  auto chained_future = executor->ScheduleAfterIfSuccess(future, [&called]() { called = true; });
+  EXPECT_FALSE(called);
+  EXPECT_FALSE(chained_future.IsFinished());
+
+  constexpr const char* const kErrorMessage{"Error"};
+  promise.SetResult(ErrorMessage{kErrorMessage});
+  executor->ExecuteScheduledTasks();
```
antonrohr, line 58: Is `chained_future` already finished between these 2 lines?

src/OrbitBase/SimpleExecutorTest.cpp
```diff
@@ -0,0 +1,127 @@
...
+  auto executor = SimpleExecutor::Create();
+  bool called = false;
+  Promise<ErrorMessageOr<int>> promise{};
+  auto future = promise.GetFuture();
+  auto chained_future = executor->ScheduleAfterIfSuccess(future, [&called](int value) {
+    EXPECT_EQ(value, 42);
+    called = true;
+    return 1 + value;
+  });
+  executor->ExecuteScheduledTasks();
```
antonrohr, line 75: At this point the lambda is still in `Executor::waiting_continuations`, correct? I wonder if checking the value of `GetNumberOfWaitingContinuations` here, is a good or a bad idea. I think it might help explaining whats going on to the reader of this test, but it also unnecessarily ties this test suite to the implementation of the `Executor` base class.

antonrohr (approved):
1. 
 > It might also be interesting for folks who want to use promises and
futures in parts of Orbit which can't depend on Qt (like the service).

Is that actually the case, or is `Immediate Executor` the better choice? And while we are at it, is it planned that `ImmediateExecutor` also inherits from `Executor`? 

2. 
Thank you for all the tests (it took me a minute to wrap my head around `SaveUniquePtr`). I generally like them a lot and how they only EXPECT things in the realm of the test. Opposed to checking every pre and post condition. I made some remarks to the tests that might very well be bad ideas, so please take them with a grain of salt. 

3. 
The code looks good to me, but I am not comfortable judging on a conceptual level, whether things could be done differently / better. Maybe @dimitry- could take a look

src/OrbitBase/SimpleExecutor.cpp
beckerhe: SimpleExecutor has a private constructor and make_shared/unique can't be used.

src/OrbitBase/SimpleExecutorTest.cpp
beckerhe: No. `promise.SetResult` schedules the continuation, and `ExecuteScheduledTasks` executes the continuation and completes the future. Does that make sense?

src/OrbitBase/SimpleExecutorTest.cpp
beckerhe: There is also another point to consider: Ideally we wanna have orthogonal tests, meaning whenever I introduce a single bug, ideally I want to have a single test fail. This will allow me to pinpoint the problem very quickly. Of course that's not always possible, or let's say most of the time that's very hard because of dependencies.

But whenever possible I prefer not to test behaviour that is already covered by another test. In this case it's covered by the `Executor.*` tests.

Needless to say that these tests are far away from consistently following this guideline because most of them are just copied from what has already been there. But I tried at least not to introduce more double-coverage.

Nevertheless, if you believe this test (these tests) requires some explanation for the reader I can totally improve on that. I guess in that case comments would be the way to go. WDYT?

src/OrbitBase/include/OrbitBase/Executor.h
```diff
@@ -0,0 +1,175 @@
...
+#include "OrbitBase/FutureHelpers.h"
+#include "OrbitBase/Promise.h"
+#include "OrbitBase/PromiseHelpers.h"
+
+namespace orbit_base {
+
+// Executor is a common base class for ThreadPool and MainThreadExecutor.
+// Check out these two for details.
+class Executor : public std::enable_shared_from_this<Executor> {
+  // Schedules the action to be performed on the main thread.
```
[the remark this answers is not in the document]
beckerhe: good catch. done. thanks

src/OrbitBase/ExecutorTest.cpp
```diff
@@ -0,0 +1,195 @@
...
+  auto future = promise.GetFuture();
+  auto chained_future = executor->ScheduleAfterIfSuccess(future, [](int value) {
+    EXPECT_EQ(value, 42);
+    return 1 + value;
+  });
+
+  EXPECT_EQ(action, nullptr);
+  EXPECT_FALSE(chained_future.IsFinished());
+
+  promise.SetResult(42);
```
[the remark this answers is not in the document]
beckerhe: good point. done.

src/OrbitBase/ExecutorTest.cpp
```diff
@@ -0,0 +1,195 @@
...
+  auto executor = MockExecutor::Create();
+  std::unique_ptr<Action> action;
+  EXPECT_CALL(*executor, ScheduleImpl(testing::_)).Times(1).WillOnce(SaveUniquePtr(&action));
+
+  bool called = false;
+  auto future = executor->Schedule([&called]() { called = true; });
+  ASSERT_NE(action, nullptr);
+  EXPECT_FALSE(called);
+  action->Execute();
+  EXPECT_TRUE(called);
```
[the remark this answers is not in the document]
beckerhe: yes. done.

src/OrbitBase/include/OrbitBase/Executor.h
```diff
@@ -0,0 +1,175 @@
...
+  virtual void ScheduleImpl(std::unique_ptr<Action> action) = 0;
+
+ public:
+  virtual ~Executor() = default;
+
+  // Schedule schedules the function object `functor` to be executed
+  // on `*this`. The execution occures asynchronously, meaning this
+  // function call will only push the function object to a queue. It
+  // will be picked up by an event loop cycle.
```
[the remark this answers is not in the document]
beckerhe: done. I hope I caught all occurrences.

src/OrbitBase/SimpleExecutorTest.cpp
antonrohr: I looked at the code for a while and then decided to just test it. 
The following "is true" (which I tried to ask earlier)
```
promise.SetResult(ErrorMessage{kErrorMessage});
EXPECT_TRUE(chained_future.IsFinished());
executor->ExecuteScheduledTasks();
```

I looked more at the code and it makes sense. This is the "Short Circuit Pathway" in `Executor::ScheduleAfterIfSuccess`. On `promise.SetResult` the continuations of the corresponding future are called and with `ScheduleAfterIfSuccess` this continuation "short circuits" and effectively also calls "`chained_promise.SetResult`" (promise of `chained_future`). In this case another action is scheduled to erase the old functor (`[&called]() { called = true; }`), which was never executed. 

This is all complicated stuff. But I think it all checks out in the end. Also maybe you misunderstood my initial question. 

Just for completion, this is how I augmented the test to understand it:
```
...
  auto future = promise.GetFuture();
  EXPECT_EQ(executor->GetNumberOfWaitingContinuations(), 0);
  auto chained_future = executor->ScheduleAfterIfSuccess(future, [&called]() { called = true; });
  EXPECT_FALSE(called);
  EXPECT_FALSE(chained_future.IsFinished());

  EXPECT_EQ(executor->GetNumberOfWaitingContinuations(), 1);
  constexpr const char* const kErrorMessage{"Error"};
  promise.SetResult(ErrorMessage{kErrorMessage});
  EXPECT_TRUE(chained_future.IsFinished());
  EXPECT_EQ(executor->GetNumberOfWaitingContinuations(), 1);
  executor->ExecuteScheduledTasks();
  EXPECT_EQ(executor->GetNumberOfWaitingContinuations(), 0);
  EXPECT_FALSE(called);
...
```

src/OrbitBase/SimpleExecutorTest.cpp
antonrohr: Using the term orthogonal, plus the explanation is a great way to put in words what was a somewhat fuzzy feeling in my head. I have to remember that. 

So yes, I agree, lets leave the tests as is. I just tried to come up with explaining comments, but at this point I think I fully understood all the code and then the tests also are clear the way they currently are. So im going to resolve the conversation

src/OrbitBase/SimpleExecutorTest.cpp
beckerhe: Ahh, yes. You're right. Sorry. I didn't get that this was about the error case.

antonrohr (approved):
I know I requested Dimitry's review earlier, but after letting it sit for a while and looking at it again, I now feel more comfortable that I fully understood it. Also its mainly moving things around. So feel free to merge even without another review

src/OrbitBase/SimpleExecutorTest.cpp
beckerhe: I quickly thought about it and I'm not entirely sure if that behaviour is an implementation detail or not. Nevertheless I will add the check! On first thought it makes sense to me, we can still remove it in the future!

beckerhe: Done.

beckerhe:
> I know I requested Dimitry's review earlier, but after letting it sit for a while and looking at it again, I now feel more comfortable that I fully understood it. Also its mainly moving things around. So feel free to merge even without another review

Thanks. I just rebased and I will merge when the CI is done.

src/OrbitBase/ThreadPool.cpp
```diff
@@ -191,8 +191,8 @@ void ThreadPoolImpl::WorkerFunction() {
 
 };  // namespace
 
-std::unique_ptr<ThreadPool> ThreadPool::Create(size_t thread_pool_min_size,
+std::shared_ptr<ThreadPool> ThreadPool::Create(size_t thread_pool_min_size,
                                                size_t thread_pool_max_size,
                                                absl::Duration thread_ttl) {
-  return std::make_unique<ThreadPoolImpl>(thread_pool_min_size, thread_pool_max_size, thread_ttl);
+  return std::make_shared<ThreadPoolImpl>(thread_pool_min_size, thread_pool_max_size, thread_ttl);
```
dimitry-, line 197: why did it become shared?

src/OrbitBase/SimpleExecutor.cpp
```diff
@@ -0,0 +1,33 @@
...
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#include "OrbitBase/SimpleExecutor.h"
+
+#include <memory>
+
+namespace orbit_base {
+void SimpleExecutor::ScheduleImpl(std::unique_ptr<Action> action) {
+  scheduled_tasks_.emplace_back(std::move(action));
```
dimitry-, line 11: MutexLock

src/OrbitBase/ThreadPool.cpp
beckerhe: This is explained in the commit message. `Executor` - the new base class - uses `std::enable_shared_from_this`.

src/OrbitBase/SimpleExecutor.cpp
beckerhe: Ahhh yes. Thanks for spotting this. Fix: #2411

src/OrbitBase/ThreadPool.cpp
dimitry-: Maybe add a comment here as well, so that reader doesn't hunt for commit message. Is is far enough away from ScheduleAfter...

beckerhe: Done. See #2411

PR #2391, opened 2021-06-02
Show file location in TargetLabel

dimitry-:
It is set on start capture and displayed next to connection status and the target

Bug: http://b/189153398

### DIFF
--- a/src/ClientModel/CaptureDeserializer.cpp
+++ b/src/ClientModel/CaptureDeserializer.cpp
@@ -219,7 +219,8 @@ ErrorMessageOr<CaptureListener::CaptureOutcome> LoadCaptureInfo(
     frame_track_function_ids.insert(function_id);
   }
 
-  capture_listener->OnCaptureStarted(capture_started, frame_track_function_ids);
+  capture_listener->OnCaptureStarted(capture_started, std::filesystem::path(),
+                                     frame_track_function_ids);
 
   for (const auto& address_info : capture_info.address_infos()) {
     if (*cancellation_requested) {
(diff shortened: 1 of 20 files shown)


beckerhe (approved):
LGTM, but Anton should have a look as well!

florian-kuebler (approved):
LGTM

src/ClientModel/CaptureDeserializer.cpp
```diff
@@ -219,7 +219,8 @@ ErrorMessageOr<CaptureListener::CaptureOutcome> LoadCaptureInfo(
     frame_track_function_ids.insert(function_id);
   }
 
-  capture_listener->OnCaptureStarted(capture_started, frame_track_function_ids);
+  capture_listener->OnCaptureStarted(capture_started, std::filesystem::path(),
```
antonrohr, line 222: This `std::filesystem::path()` unfortunately bites you later on. `OrbitApp::OnCaptureStarted` calls `capture_started_callback_(file_path)` which is defined in `orbitmainwindow::SetCaptureStartedCallback` and this then calls `target_label_->SetFile(path)`. This means the file target will effectively be cleared (when loading an old capture from file) because the path is here set to `{}`. 

There are multiple ways of fixing that. You could do some more `std::optional` things, but I would recommend to just add a `const std::filesystem::path& file_path` parameter to `LoadCaptureInfo(...)`. `LoadCaptureInfo` is only called once (besides tests) and the file path is available at the call site.
Since you then need to touch `CaptureDeserializerTest` as well, maybe you could also quickly test that, by adding an `EXPECT` to the test `LoadCaptureInfoOnCaptureStarted`

antonrohr (changes requested):
I only have one thing, but its important

src/ClientModel/CaptureDeserializer.cpp
dimitry-: nice catch! I think I fixed that (by not reseting the label if file_path is not specified), we do not want to save the file name since it is going to be used to save on modification, and we do not want to save on modification in legacy file format.

antonrohr (approved):
thanks for addressing that

PR #2392, opened 2021-06-04
Show [unknown@address] instead of "???" in CallstackThreadBar

dpallotti:
"???" provides no information, but when the function name is not known, we can
at least show the virtual address.
This is the same format as used in the top-down and bottom-up view.

Test: Cause several unwinding errors by instrumenting a function called very
frequently. Hover over one of those in the timeline.


karupayun (approved):
LGTM!

florian-kuebler (approved):
LGTM!

PR #2395, opened 2021-06-06
Automatically save user data on modification

dimitry-:
Bug: http://b/189145024
Test: Run Capture, modify frame-tracks, restart Orbit, load file and
      check that it is correct.

### DIFF
--- a/src/OrbitGl/App.cpp
+++ b/src/OrbitGl/App.cpp
@@ -34,6 +34,7 @@
 #include "CallstackDataView.h"
 #include "CaptureClient/CaptureListener.h"
 #include "CaptureFile/CaptureFile.h"
+#include "CaptureFile/CaptureFileHelpers.h"
 #include "CaptureWindow.h"
 #include "ClientData/CallstackData.h"
 #include "ClientData/FunctionUtils.h"
@@ -323,6 +324,7 @@ Future<void> OrbitApp::OnCaptureComplete() {
   return main_thread_executor_->Schedule(
       [this, sampling_profiler = std::move(post_processed_sampling_data)]() mutable {
         ORBIT_SCOPE("OnCaptureComplete");
+        TrySaveUserDefinedCaptureInfo();
         RefreshFrameTracks();
         GetMutableCaptureData().set_post_processed_sampling_data(sampling_profiler);
         RefreshCaptureView();
@@ -1107,8 +1109,8 @@ static std::unique_ptr<CaptureEventProcessor> CreateCaptureEventProcessor(
                                     orbit_client_model::capture_serializer::GenerateCaptureFileName(
                                         process_name, absl::Now(), "_autosave");
 
-  auto save_to_file_processor_or_error = CaptureEventProcessor::CreateSaveToFileProcessor(
-      file_path, frame_track_function_ids, error_handler);
+  auto save_to_file_processor_or_error =
+      CaptureEventProcessor::CreateSaveToFileProcessor(file_path, error_handler);
 
   if (save_to_file_processor_or_error.has_error()) {
     error_handler(ErrorMessage{
@@ -2301,6 +2303,7 @@ void OrbitApp::AddFrameTrack(const FunctionInfo& function) {
 
 void OrbitApp::AddFrameTrack(uint64_t instrumented_function_id) {
   CHECK(instrumented_function_id != 0);
+  CHECK(std::this_thread::get_id() == main_thread_id_);
   if (!HasCaptureData()) {
     return;
   }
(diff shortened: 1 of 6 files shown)


pierricgimmig (approved):
Hi Dimitry, thanks for this, the code is also cleaner now. LGTM but please see minor comments.

src/OrbitGl/App.cpp
```diff
@@ -2564,3 +2569,32 @@ void OrbitApp::CaptureMetricProcessApiTimer(const orbit_client_protos::TimerInfo
...
   }
 }
+
+void OrbitApp::SaveCaptureUserInfo() {
+  CHECK(std::this_thread::get_id() == main_thread_id_);
+  if (!HasCaptureData()) return;
+  if (IsCapturing()) {
+    // We are going to save it at the end of capture anyways.
+    return;
+  }
```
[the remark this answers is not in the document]
dimitry-: yes, that is right - we cannot serialize user_data section while capturing.

src/OrbitGl/App.cpp
```diff
@@ -2564,3 +2569,32 @@ void OrbitApp::CaptureMetricProcessApiTimer(const orbit_client_protos::TimerInfo
...
+
+  const auto& file_path = GetCaptureData().file_path();
+  if (!file_path.has_value()) {
+    LOG("Warning: capture is not backed by a file, skipping the save.");
```
dpallotti, line 2583: Is this expected or not? If not I would just use `ERROR`. If yes, I would just drop the "Warning".

dpallotti, line 2583:
```suggestion
    LOG("Warning: capture is not backed by a file, skipping the save");
```

dpallotti (approved):

src/OrbitGl/App.cpp
```diff
@@ -2564,3 +2569,32 @@ void OrbitApp::CaptureMetricProcessApiTimer(const orbit_client_protos::TimerInfo
...
+  orbit_client_protos::UserDefinedCaptureInfo capture_info;
+  const auto& frame_track_function_ids = GetCaptureData().frame_track_function_ids();
+  *capture_info.mutable_frame_tracks_info()->mutable_frame_track_function_ids() = {
+      frame_track_function_ids.begin(), frame_track_function_ids.end()};
+  thread_pool_->Schedule([this, capture_info = std::move(capture_info),
+                          file_path = file_path.value()] {
+    LOG("Saving user defined capture info to \"%s\"", file_path.string());
+    auto write_result = orbit_capture_file::WriteUserData(file_path, capture_info);
+    if (write_result.has_error()) {
+      SendErrorToUi("Save failed", absl::StrFormat("Save to \"%s\" failed: %s", file_path.string(),
```
antonrohr, line 2596: The following is basically a different idea how to structure this whole function. It requires significantly more code at call site and maybe thats a bad tradeoff. So feel free to not do this

This error is displayed to the user directly and in my opinion might be confusing to the user. So maybe `SaveCaptureUserInfo` can return `ErrorMessageOr<void>` and return this string here as an error message. Then the callsite could augment this error with more context (e.g. Unable to add frame track to the capture file on disk, error: Save to ... failed: ...)
But this also means the call site needs to decide whether to call `SaveCaptureUserInfo` and thats maybe uncomfortable. 

In case you dont do this, maybe also consider naming this function `TrySaveCaptureUserInfo`.

src/OrbitGl/App.cpp
```diff
@@ -2301,6 +2303,7 @@ void OrbitApp::AddFrameTrack(const FunctionInfo& function) {
 
 void OrbitApp::AddFrameTrack(uint64_t instrumented_function_id) {
   CHECK(instrumented_function_id != 0);
+  CHECK(std::this_thread::get_id() == main_thread_id_);
   if (!HasCaptureData()) {
```
antonrohr, line 2307: Maybe put a comment here. // This is only the case when an old capture file was loaded

antonrohr (approved):

src/OrbitGl/App.cpp
dimitry-: > But this also means the call site needs to decide whether to call SaveCaptureUserInfo and thats maybe uncomfortable.

I do not think I got this part. Why does caller side need to decide?

src/OrbitGl/App.cpp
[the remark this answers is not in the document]
dimitry-: done.

dimitry-: done.

src/OrbitGl/App.cpp
dimitry-: Also you probably meant `Future<ErrorMessageOr<void>>`

dimitry-: and I will need to implement ScheduleAfterIfError()... I am going to leave this as it is for now.

antonrohr: > I do not think I got this part. Why does caller side need to decide?

I just meant that most of the checks that you are doing in this function would need to be done at call site (for example `if (!file_path.has_value())`. But now it doesnt matter anyways ;)

PR #2396, opened 2021-06-07
Remove save button/menu item and warning dialog about unsaved progress

dimitry-:
Bug: http://b/189159722


beckerhe (approved):
Nice. Thanks. LGTM

src/OrbitQt/orbitmainwindow.cpp
```diff
@@ -1065,35 +1064,6 @@ void OrbitMainWindow::OnTimerSelectionChanged(const orbit_client_protos::TimerIn
   ui->liveFunctions->OnRowSelected(selected_row);
 }
 
-void OrbitMainWindow::on_actionSave_Capture_triggered() {
-  ShowCaptureOnSaveWarningIfNeeded();
```
dpallotti, line 1069: I think this is the only use of this method and as such it can be removed too. I might be wrong.

dpallotti (approved):

src/OrbitQt/orbitmainwindow.cpp
dimitry-: right - removed.

antonrohr (approved):
lgtm

PR #2399, opened 2021-06-08
Move screen vertically after double clicking in an event 

karupayun:
Move the screen accordingly to assure that the event and its track are fully visible. Also rename Select method to make it more clear and fix an issue with the vertical alignment (http://b/190463784).

Solve http://b/174304405.

Test: Start a capture, stop, make some jumps, double click in some timers.


vickyliu-go4it (approved):
LGTM! Thanks for fixing this and I like the new name `SelectAndMakeVisible`.

PR #2400, opened 2021-06-08
Send FullAddressInfo for addresses falling directly inside uprobes

dpallotti:
We can send a meaningful `FullAddressInfo` even for addresses falling inside the
`[uprobes]` map, by treating it as a single function (which is the case, only
two possible addresses seem to be involved).

Test: Instrument Trata's very frequently called
`eva::concurrent::thread_aware_task_sink::execute_task(bool)`. Verify the
expected result in "Sampling", "Top-down", "Bottom-up" (instead of the previous
"???" or "unknown@0x7fffffffe001").

### DIFF
--- a/src/LinuxTracing/UprobesUnwindingVisitor.cpp
+++ b/src/LinuxTracing/UprobesUnwindingVisitor.cpp
@@ -41,6 +41,25 @@ static void SendFullAddressInfoToListener(TracerListener* listener,
   listener->OnAddressInfo(std::move(address_info));
 }
 
+// For addresses falling directly inside u(ret)probes code, unwindstack::FrameData has limited
+// information. Nonetheless, we can send a perfectly meaningful FullAddressInfo, treating
+// u(ret)probes code as a single function. This makes sense as the only affected virtual addresses I
+// observed are 0x7fffffffe000 (~1% of uprobes addresses) and 0x7fffffffe001 (~99%). This way the
+// client can show more information for such a frame, in particular when associated with the
+// corresponding unwinding error.
+static void SendUprobesFullAddressInfoToListener(
+    TracerListener* listener, const unwindstack::FrameData& libunwindstack_frame) {
+  CHECK(listener != nullptr);
+
+  FullAddressInfo address_info;
+  address_info.set_absolute_address(libunwindstack_frame.pc);
+  address_info.set_function_name("[uprobes]");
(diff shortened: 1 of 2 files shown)


src/LinuxTracing/UprobesUnwindingVisitor.cpp
```diff
@@ -41,6 +41,25 @@ static void SendFullAddressInfoToListener(TracerListener* listener,
...
+// corresponding unwinding error.
+static void SendUprobesFullAddressInfoToListener(
+    TracerListener* listener, const unwindstack::FrameData& libunwindstack_frame) {
+  CHECK(listener != nullptr);
+
+  FullAddressInfo address_info;
+  address_info.set_absolute_address(libunwindstack_frame.pc);
+  address_info.set_function_name("[uprobes]");
+  address_info.set_offset_in_function(libunwindstack_frame.pc - libunwindstack_frame.map_start);
+  address_info.set_module_name("[uprobes]");
```
florian-kuebler, line 58: I guess one issue is, that we are not showing the [uprobes] module in the modules list. So that might be suprising.

dpallotti: Is that an issue? We can also just leave this empty and keep having "???", but I don't see how that's better than giving some additional information that we have.

florian-kuebler: I agree with giving additional information, but we should be careful to not confuse the user. They probably don't know about uprobes.

dpallotti: That's the point, if they don't know about it they can find out more, while if they do that makes immediately sense to them. Otherwise they have some 0x7fffffffe001 address with "???" they can't explain and that's way more confusing.

florian-kuebler: Yeah, agree with that. We should probably at some later pointer also point them to the right resources (e.g. in the tooltip).

dpallotti: Yep, agree, but I really wanted to keep this as a quick win.

dpallotti:
I forgot that I need to update a test, working on that.

florian-kuebler (approved):
LGTM, thanks!

dpallotti:
The test should be updated properly now.

PR #2401, opened 2021-06-08
Recent capture files pr

antonrohr:
I tried to structure the commits nicely, so not a lot of explanation is necessary. 

This is mainly adding the ItemModel for CaptureFileInfo and adds FillFromDirectory to the Manager. The later will be used to only once to fill the recent capture files from the output directory

### DIFF
--- a/src/CaptureFileInfo/CMakeLists.txt
+++ b/src/CaptureFileInfo/CMakeLists.txt
@@ -48,8 +48,11 @@ target_link_libraries(
           QtUtils
           Qt5::Test)
 
-add_custom_command(TARGET CaptureFileInfoTests POST_BUILD
-  COMMAND ${CMAKE_COMMAND} -E copy_directory ${CMAKE_CURRENT_LIST_DIR}/testdata
-  $<TARGET_FILE_DIR:CaptureFileInfoTests>/testdata)
+add_custom_command(TARGET CaptureFileInfoTests POST_BUILD 
+          COMMAND ${CMAKE_COMMAND} -E remove_directory 
+            $<TARGET_FILE_DIR:CaptureFileInfoTests>/testdata/CaptureFileInfo
+          COMMAND ${CMAKE_COMMAND} -E copy_directory 
+            ${CMAKE_CURRENT_SOURCE_DIR}/testdata/
+            $<TARGET_FILE_DIR:CaptureFileInfoTests>/testdata/CaptureFileInfo)
 
 register_test(CaptureFileInfoTests)
--- a/src/CaptureFileInfo/CMakeLists.txt
+++ b/src/CaptureFileInfo/CMakeLists.txt
@@ -12,18 +12,24 @@ target_compile_options(CaptureFileInfo PRIVATE ${STRICT_COMPILE_FLAGS})
 target_sources(
   CaptureFileInfo
   PUBLIC  include/CaptureFileInfo/CaptureFileInfo.h
+          include/CaptureFileInfo/ItemModel.h
           include/CaptureFileInfo/Manager.h)
 
 target_sources(
   CaptureFileInfo
   PRIVATE CaptureFileInfo.cpp
+          ItemModel.cpp
           Manager.cpp)
 
 target_include_directories(CaptureFileInfo PUBLIC include/)
 
 target_link_libraries(
   CaptureFileInfo
-  PUBLIC  Qt5::Core)
+  PUBLIC  OrbitBase
+          OrbitCore
+          Qt5::Core)
+
+set_target_properties(CaptureFileInfo PROPERTIES AUTOMOC ON)
 
 add_executable(CaptureFileInfoTests)
 
@@ -32,13 +38,15 @@ target_compile_options(CaptureFileInfoTests PRIVATE ${STRICT_COMPILE_FLAGS})
 target_sources(
   CaptureFileInfoTests 
   PRIVATE CaptureFileInfoTest.cpp
+          ItemModelTest.cpp
           ManagerTest.cpp)
 
 target_link_libraries(
   CaptureFileInfoTests
-  PRIVATE CaptureFileInfo 
-          OrbitBase
-          GTest::Main)
+  PRIVATE CaptureFileInfo
+          GTest::Main
+          QtUtils
+          Qt5::Test)
 
 add_custom_command(TARGET CaptureFileInfoTests POST_BUILD
   COMMAND ${CMAKE_COMMAND} -E copy_directory ${CMAKE_CURRENT_LIST_DIR}/testdata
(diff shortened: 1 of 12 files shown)


src/CaptureFileInfo/CaptureFileInfoTest.cpp
```diff
@@ -85,4 +85,23 @@ TEST(CaptureFileInfo, Touch) {
...
+  {
+    const QString non_existing_path{"test/path/file.ext"};
+
+    CaptureFileInfo capture_file_info{non_existing_path};
+
+    EXPECT_EQ(capture_file_info.FileSize(), 0);
+  }
+  {
+    const std::filesystem::path path =
+        orbit_base::GetExecutableDir() / "testdata" / "not_existing_test_file.txt";
```
beckerhe: `not_existing_test_file`? Does that file actually exist?

src/CaptureFileInfo/ItemModel.cpp
```diff
@@ -0,0 +1,78 @@
...
+#include "CaptureFileInfo/CaptureFileInfo.h"
+#include "CoreUtils.h"
+#include "OrbitBase/Logging.h"
+
+namespace orbit_capture_file_info {
+
+void ItemModel::SetCaptureFileInfos(std::vector<CaptureFileInfo> capture_file_infos) {
+  beginInsertRows(QModelIndex{}, 0, static_cast<int>(capture_file_infos.size()) - 1);
+  capture_files_ = std::move(capture_file_infos);
+  endInsertRows();
```
beckerhe, line 16: You're replacing all entries in capture_files_ which means we have to express that as deleting all rows first and then adding new ones. So this needs some `beginRemoveRows` as well.

beckerhe: This is also not taking into account that `capture_file_infos` could be empty.

src/CaptureFileInfo/Manager.cpp
```diff
@@ -48,14 +48,14 @@ void Manager::SaveCaptureFileInfos() {
   settings.endArray();
 }
 
-void Manager::AddOrTouchCaptureFile(const QString& path) {
+void Manager::AddOrTouchCaptureFile(const std::filesystem::path& path) {
```
beckerhe, line 53: I don't see why you're doing this. Maybe you could add an explanation to the commit message?

src/CaptureFileInfo/Manager.cpp
```diff
@@ -77,4 +79,19 @@ void Manager::PurgeNonExistingFiles() {
...
 }
 
+ErrorMessageOr<void> Manager::FillFromDirectory(const std::filesystem::path& directory) {
+  OUTCOME_TRY(files, orbit_base::ListFilesInDirectory(directory));
+
+  for (const auto& file : files) {
+    if (file.extension() != ".orbit") continue;
+
+    QFileInfo tmp_file_info{QString::fromStdString(file.string())};
+    capture_file_infos_.emplace_back(tmp_file_info.filePath(), tmp_file_info.birthTime());
```
beckerhe, line 89: What happens when FillFromDirectory has already been called, or a particular file is already present in `capture_file_infos_`. There is no deduplication, right? Maybe this method should clear `capture_file_infos_` before filling it.

src/CaptureFileInfo/ManagerTest.cpp
```diff
@@ -144,4 +147,33 @@ TEST(CaptureFileInfoManager, Persistency) {
...
+TEST(CaptureFileInfoManager, FillFromDirectory) {
+  QCoreApplication::setOrganizationName(kOrgName);
+  QCoreApplication::setApplicationName("CaptureFileInfo.Manager.FillFromDirectory");
+
+  Manager manager;
+  manager.Clear();
+
+  {  // Fail
+    ErrorMessageOr<void> result = manager.FillFromDirectory("/non/existent/path/to/dir");
+    ASSERT_TRUE(result.has_error());
```
beckerhe, line 159: I think this is superfluous when you use the HasError matcher

src/CaptureFileInfo/CMakeLists.txt
```diff
@@ -48,8 +48,11 @@ target_link_libraries(
...
 
-add_custom_command(TARGET CaptureFileInfoTests POST_BUILD
-  COMMAND ${CMAKE_COMMAND} -E copy_directory ${CMAKE_CURRENT_LIST_DIR}/testdata
-  $<TARGET_FILE_DIR:CaptureFileInfoTests>/testdata)
+add_custom_command(TARGET CaptureFileInfoTests POST_BUILD COMMAND ${CMAKE_COMMAND}
+          -E remove_directory $<TARGET_FILE_DIR:CaptureFileInfoTests>/testdata/CaptureFileInfo)
+
+add_custom_command(TARGET CaptureFileInfoTests POST_BUILD COMMAND ${CMAKE_COMMAND}
+          -E copy_directory ${CMAKE_CURRENT_SOURCE_DIR}/testdata/
+          $<TARGET_FILE_DIR:CaptureFileInfoTests>/testdata/CaptureFileInfo)
```
beckerhe: When using two rules you don't have control over the execution order. They could be executed in parallel.

But you can use multiple COMMAND directives in the same `add_custom_command` call. They will be executed in order.

src/CaptureFileInfo/CaptureFileInfoTest.cpp
antonrohr: I rebased wrong. This was a mistake that I fixed, but the fix ended up one commit later. This means also the test will fail at this commit. I will fix that

beckerhe: Feel free to force push.

src/CaptureFileInfo/CMakeLists.txt
antonrohr: I copied this from the CMakeLists.txt of OrbitBase. I fixed it here and added a separate commit for fixing it in OrbitBase

src/CaptureFileInfo/Manager.cpp
antonrohr: done

src/CaptureFileInfo/Manager.cpp
antonrohr: I will add an explanation to the commit message when doing the rebase.

src/CaptureFileInfo/ItemModel.cpp
antonrohr: done

src/CaptureFileInfo/CMakeLists.txt
beckerhe: Even better. Thanks!

beckerhe (approved):
Thank you for all the fixes. LGTM!

dimitry- (approved):

PR #2402, opened 2021-06-08
Fix Heap-use-after-free in GpuQueueSubmissionProcessor

florian-kuebler:
Esentially, the `matching_gpu_job` can already get deleted by
the `ProcessGpuQueueSubmissionWithMatchingGpuJob`.

Bug: https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=34934
Test: Run the Fuzzer with the given input


beckerhe (approved):
Great. Thanks a lot for taking care of this!

PR #2403, opened 2021-06-09
Add missing tests to EventLoop

beckerhe:
coverage, coverage, coverage! and some useful tests.


antonrohr (approved):
34.29% coverage ??? Gotta pump those numbers up, those are rookie numbers

PR #2405, opened 2021-06-09
Introduce module Symbols

beckerhe:
This is a first step to move a symbol loading related code into a single module which can be tested properly.

It moves SymbolHelper into a module "Symbols" and thereby cleans up the OrbitCore cmake config.

A second commit increases test coverage for SymbolHelper and fixes some issues around the SymbolPaths.txt which have been recently raised by an internal user.


dpallotti (approved):
Thanks!

PR #2407, opened 2021-06-09
Add -readability-function-cognitive-complexity to .clang-tidy

dpallotti:
After updating CLion, this caused way too many warnings.

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -26,6 +26,7 @@ Checks: >
   performance-*,
   -performance-noexcept-move-constructor,
   readability-*,
+  -readability-function-cognitive-complexity,
   -readability-magic-numbers,
   -readability-non-const-parameter,
   -readability-uppercase-literal-suffix,


.clang-tidy
```diff
@@ -26,6 +26,7 @@ Checks: >
   performance-*,
   -performance-noexcept-move-constructor,
   readability-*,
+  -readability-function-cognitive-complexity,
```
dimitry-, line 29: we use some number of macros in our code (OUTCOME_TRY, ERROR, etc). I have a feeling that this check works after all macros are pre-processed which makes it not very useful. Have you looked into this?

My data-point is CLion complaining about cognitive complexity of a small method which uses a lot of OUTCOME_TRYs.

dimitry-: nevermind - you meant to disable it not enable.

dimitry- (approved):

.clang-tidy
dpallotti: Yeah, sorry, "Add" followed by "-" was not my brightest commit message.

PR #2410, opened 2021-06-09
Add additional MainThreadExecutor tests

beckerhe:
This is adding tests for the different use cases and outcomes of
`MainThreadExecutor::WaitFor` and `MainThreadExecutor::WaitForAll`.

### DIFF
--- a/src/QtUtils/MainThreadExecutorImplTest.cpp
+++ b/src/QtUtils/MainThreadExecutorImplTest.cpp
@@ -5,6 +5,7 @@
 #include <gtest/gtest.h>
 
 #include <QCoreApplication>
+#include <chrono>
 #include <memory>
 
 #include "OrbitBase/Future.h"
@@ -121,13 +122,6 @@ TEST(MainThreadExecutorImpl, ScheduleAfterWithExecutorOutOfScope) {
   promise.MarkFinished();
   QCoreApplication::processEvents();
   EXPECT_FALSE(called);
-}
-
-TEST(MainThreadExecutorImpl, Wait) {
-  bool called = false;
-  auto executor = MainThreadExecutorImpl::Create();
-  orbit_base::Future<void> future = executor->Schedule([&called]() { called = true; });
-  EXPECT_EQ(executor->WaitFor(future), MainThreadExecutor::WaitResult::kCompleted);
 }
 
 TEST(MainThreadExecutorImpl, ChainFuturesWithThen) {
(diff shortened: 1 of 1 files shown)


src/QtUtils/MainThreadExecutorImplTest.cpp
```diff
@@ -266,4 +260,99 @@ TEST(MainThreadExecutorImpl, ScheduleAfterIfSuccessTwice) {
...
+TEST(MainThreadExecutorImpl, WaitForNoTimeoutCompletedFutures) {
+  auto executor = MainThreadExecutorImpl::Create();
+  orbit_base::Future<void> future{};  // Constructs completed future
+  EXPECT_EQ(executor->WaitFor(future), MainThreadExecutor::WaitResult::kCompleted);
+}
+
+TEST(MainThreadExecutorImpl, WaitForWithTimeoutCompletedFutures) {
+  auto executor = MainThreadExecutorImpl::Create();
+  orbit_base::Future<void> future{};  // Constructs completed future
+  EXPECT_EQ(executor->WaitFor(future, std::chrono::milliseconds{1000}),
```
dimitry-, line 285: does it work with 10? if not maybe change timeout in the next test to 1 second as well.

src/QtUtils/MainThreadExecutorImplTest.cpp
```diff
@@ -266,4 +260,99 @@ TEST(MainThreadExecutorImpl, ScheduleAfterIfSuccessTwice) {
...
+            MainThreadExecutor::WaitResult::kCompleted);
+}
+
+TEST(MainThreadExecutorImpl, WaitForAllWithExceedingTimeout) {
+  auto executor = MainThreadExecutorImpl::Create();
+
+  // We won't mark these promises as finished, so we would wait forever if there wasn't a time out.
+  std::array promises{orbit_base::Promise<void>{}, orbit_base::Promise<void>{}};
+  std::array futures = {promises[0].GetFuture(), promises[1].GetFuture()};
+  EXPECT_EQ(executor->WaitForAll(absl::MakeSpan(futures), std::chrono::milliseconds{10}),
```
dimitry-, line 343: and here.

src/QtUtils/MainThreadExecutorImplTest.cpp
beckerhe: I just tried that. Yes it works. (Ran the tests 10,000 times).

But I'm not sure what to gain from this change. Shouldn't make a difference functionality-wise and might potentially increase likelyhood of flaky test. What am I missing?

dimitry-: the point here if this one does not work for some timeout the next test might yield false positive result. If a test times out for 10 on completed future we shouldn't assume it will not timeout on incorrectly working incomplete future.

dimitry- (approved):

src/QtUtils/MainThreadExecutorImplTest.cpp
beckerhe: Got it. Okay. Makes sense. I will leave it at 10ms.

PR #2413, opened 2021-06-10
Fix crash after a module was remapped

dimitry-:
This fixed problem of detecting module by absolute address
when module was reloaded or a second copy of the module was loaded
(which is unlikely but could happen).

Now we store full module_in_memory info for both locations
and can return correct result.

This commit aslo marks some problematic functions which may return
incomplete results in this case as [[depracated]]. To be addressed in
the followup PRs

Test: run OrbitClientDataTests
Bug: http://b/190357159


src/ClientData/ProcessData.cpp
```diff
@@ -92,11 +92,10 @@ std::optional<ModuleInMemory> ProcessData::FindModuleByPath(const std::string& m
...
 void ProcessData::AddOrUpdateModuleInfo(const ModuleInfo& module_info) {
   absl::MutexLock lock(&mutex_);
-  module_memory_map_.insert_or_assign(
-      module_info.file_path(),
-      ModuleInMemory{module_info.address_start(), module_info.address_end(),
-                     module_info.file_path(), module_info.build_id()});
-  start_addresses_.insert_or_assign(module_info.address_start(), module_info.file_path());
+  ModuleInMemory module_in_memory{module_info.address_start(), module_info.address_end(),
+                                  module_info.file_path(), module_info.build_id()};
+  module_memory_map_.insert_or_assign(module_info.file_path(), module_in_memory);
```
ronaldfw, line 97: Just to clarify: This line is the reason why some functions are now marked as deprecated as they may return only the most recently inserted mapped module, correct?

Nit: If the above is correct, maybe add a brief comment. But if you are addressing this anyway in follow-up PRs, can also be omitted.

ronaldfw (commented):
General question: Since we now keep all entries of a module (indexed by start address), can it now happen that we have modules in start_addresses_ that overlap with their address ranges? As we don't handle unmap events. Previously we implicitly did.

O/w LGTM

src/ClientData/ProcessData.cpp
dimitry-: yes - it is correct.

dimitry-:
> Also I have absolutely no idea why the test fails on CI and does not fail locally.

 was running the test under the old name...

ronaldfw:
> Actually we didn't since start_address remained untouched, it was just pointing to invalid module_in_memory. But it is a good point - we should probably check for collisions and remove modules that intersect with freshly mapped addresses.

Ah, you are right. Thanks for clarifying.

src/ClientData/ProcessData.cpp
```diff
@@ -66,15 +66,15 @@ void ProcessData::UpdateModuleInfos(absl::Span<const ModuleInfo> module_infos) {
 
   for (const auto& module_info : module_infos) {
     {
-      const auto [it, success] = module_memory_map_.try_emplace(
-          module_info.file_path(),
-          ModuleInMemory{module_info.address_start(), module_info.address_end(),
-                         module_info.file_path(), module_info.build_id()});
+      const auto [unused, success] = module_memory_map_.try_emplace(
```
dpallotti, line 69:
```suggestion
      const auto [unused_it, success] = module_memory_map_.try_emplace(
```
This is the format we had decided on.

src/ClientData/ProcessData.cpp
```diff
@@ -66,15 +66,15 @@ void ProcessData::UpdateModuleInfos(absl::Span<const ModuleInfo> module_infos) {
...
+          module_info.file_path(), module_info.address_start(), module_info.address_end(),
+          module_info.file_path(), module_info.build_id());
       CHECK(success);
     }
     {
-      const auto [it, success] =
-          start_addresses_.try_emplace(module_info.address_start(), module_info.file_path());
+      const auto [unused, success] = start_addresses_.try_emplace(
```
dpallotti, line 75:
```suggestion
      const auto [unused_it, success] = start_addresses_.try_emplace(
```

src/ClientData/include/ClientData/ProcessData.h
```diff
@@ -73,10 +73,17 @@ class ProcessData final {
...
+  [[deprecated]] std::optional<uint64_t> GetModuleBaseAddress(const std::string& module_path) const;
 
-  [[nodiscard]] std::optional<ModuleInMemory> FindModuleByPath(
+  // This function is deprecated since it relies on the fact that only one instance
+  // of module is loaded in the process which is not always true.
+  [[deprecated]] absl::node_hash_map<std::string, ModuleInMemory> GetMemoryMapCopy() const;
+
+  // This function is deprecated since it relies on the fact that only one instance
+  // of module is loaded in the process which is not always true.
+  [[nodiscard, deprecated]] std::optional<ModuleInMemory> FindModuleByPath(
```
dpallotti, line 86: I'm a bit skeptical about adding `[[deprecated]]` in this case, as it's not like there's an alternative to it at the time of this commit. It's not a problem if these are just going to be removed soon.

src/ClientData/ProcessDataTest.cpp
```diff
@@ -341,4 +343,63 @@ TEST(ProcessData, FindModuleByAddress) {
...
+  ModuleInfo module_info;
+  module_info.set_file_path(kModulePath);
+  module_info.set_build_id(kBuildId);
+  module_info.set_address_start(kStartAddress);
+  module_info.set_address_end(kEndAddress);
+
+  process.UpdateModuleInfos({module_info});
+
+  {
+    // start address
```
dpallotti, line 371: What is this comment really saying?

src/ClientData/include/ClientData/ProcessData.h
```diff
@@ -91,7 +98,7 @@ class ProcessData final {
 
   // This is a map from module_path to the space in memory where that module is loaded
   absl::node_hash_map<std::string, ModuleInMemory> module_memory_map_;
-  std::map<uint64_t, std::string> start_addresses_;
+  std::map<uint64_t, ModuleInMemory> start_addresses_;
```
dpallotti, line 101: Can we take rename this to `start_address_to_module_in_memory_` or something like that? The name is currently hinting at this being a set.

dpallotti (approved):
Looks good otherwise.

antonrohr (approved):
Im okay with this PR, because it doesn't make anything worse, but also only some things better. 

We still have the problem that there could be overlapping address spaces when a module was unloaded by the process which Orbit currently does not notice (And another module is afterwards loaded in an overlapping space). This can effectively lead to a wrong resolution from absolute address to module

src/ClientData/include/ClientData/ProcessData.h
dimitry-: we will have one peace of code that will be using it after fixing up other references and that is Deseraalization from the old file format. I think we should be able to move disassebly away from this (but it could also end up using it for some time).

I would like to avoid to have new code using it - hence [[depracated]]

dpallotti: > that will be using it after fixing up other references

> I would like to avoid to have new code using it

Yes, I agree, this is what it's for. What I meant is that since you said that that's addressed in an upcoming change, I would have expected the `[[deprecated]]` to also be added in that change, instead of in advance (before the alternative is added).

src/ClientData/ProcessData.cpp
dimitry-: done.

src/ClientData/ProcessData.cpp
dimitry-: done.

src/ClientData/ProcessDataTest.cpp
dimitry-: removed.

src/ClientData/include/ClientData/ProcessData.h
dimitry-: makes sense - removed deprecated attribute from this commit - will introduce in the follow-up.

src/ClientData/include/ClientData/ProcessData.h
dimitry-: done.

PR #2423, opened 2021-06-11
Fix log file name issue in main.cpp

antonrohr:
Beforehand there were 2 calls to orbit_core::GetLogFilePath(), which
both generated a filename. This was never a problem, because they were
usually called within one second and then the generated file name does
not change.
This commit changes this so the log file name is only generated once.


vickyliu-go4it (approved):
LGTM. Thanks for fixing it!

PR #2426, opened 2021-06-11
Refactor ModuleDataView to use origianl Memory Map

dimitry-:
This avoids a problem with not displaying a module if it was mapped
twice.

This change also introduces separate method to retrieve module
paths and build_ids from the process memory map.

Test: run unit-tests, Start Orbit and check that ModuleView works.
Bug: http://b/190805289

### DIFF
--- a/src/ClientData/ProcessData.cpp
+++ b/src/ClientData/ProcessData.cpp
@@ -131,15 +131,26 @@ std::optional<uint64_t> ProcessData::GetModuleBaseAddress(const std::string& mod
   return module_memory_map_.at(module_path).start();
 }
 
-absl::node_hash_map<std::string, ModuleInMemory> ProcessData::GetMemoryMapCopy() const {
+std::map<uint64_t, ModuleInMemory> ProcessData::GetMemoryMapCopy() const {
   absl::MutexLock lock(&mutex_);
-  return module_memory_map_;
+  return start_address_to_module_in_memory_;
 }
 
 bool ProcessData::IsModuleLoadedByProcess(const ModuleData* module) const {
   const std::optional<ModuleInMemory> loaded_module = FindModuleByPath(module->file_path());
   if (!loaded_module.has_value()) return false;
   return loaded_module->build_id() == module->build_id();
+}
+
+std::vector<std::pair<std::string, std::string>> ProcessData::GetUniqueModulesPathAndBuildId()
+    const {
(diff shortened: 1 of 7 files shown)


src/ClientData/include/ClientData/ProcessData.h
```diff
@@ -77,9 +77,8 @@ class ProcessData final {
   // of module is loaded in the process which is not always true.
   std::optional<uint64_t> GetModuleBaseAddress(const std::string& module_path) const;
 
-  // This function is deprecated since it relies on the fact that only one instance
-  // of module is loaded in the process which is not always true.
-  absl::node_hash_map<std::string, ModuleInMemory> GetMemoryMapCopy() const;
+  std::map<uint64_t, ModuleInMemory> GetMemoryMapCopy() const;
+  std::vector<std::pair<std::string, std::string>> GetUniqueModulesPathAndBuildId() const;
```
beckerhe, line 81: nodiscard

src/ClientData/ProcessData.cpp
```diff
@@ -142,4 +142,15 @@ bool ProcessData::IsModuleLoadedByProcess(const ModuleData* module) const {
   return loaded_module->build_id() == module->build_id();
 }
 
+std::vector<std::pair<std::string, std::string>> ProcessData::GetUniqueModulesPathAndBuildId()
+    const {
+  absl::MutexLock lock(&mutex_);
+  std::set<std::pair<std::string, std::string>> module_keys;
```
beckerhe, line 148: What about making `ModuleKey` a struct instead of using a `std::pair`? It's used in quite some places and it's hard to remember whether the first element is the module path or the second. It might also be error-prone since both are strings.

beckerhe (approved):
Otherwise LGTM

src/ClientData/ProcessData.cpp
dimitry-: The reason I was hesitant of doing that is this is something used in sets which means we will need std::hash, absl::Hash, less, equals and then test for all of them. With pair you just get it for free. I am not sure if it is worth it.

beckerhe: Hmm, yes. True. Feel free to choose what makes more sense to you.

src/ClientData/include/ClientData/ProcessData.h
dimitry-: done.

PR #2427, opened 2021-06-14
Add tests for PresetsDataView

beckerhe:
This is adding tests for PresetsDataView and is preparing to add tests for all the other DataViews.

The plan is the following:
- We introduce an abstract base class of OrbitApp called `AppInterface`. This base class declares all member functions of `OrbitApp` that are needed in the data views as pure virtual and they will be marked as overriden in `OrbitApp`.
- This allows us to replace the `OrbitApp*` pointers in the data views by `AppInterface*` pointers. 
- `AppInterface` can be mocked by gmock and so we can test the data views without OrbitApp.
- On top of that this is moving all the data view related stuff into a new module `DataViews` which does not depend on `OrbitGl` anymore.

Don't be too alarmed by the number of commits. Most of them are pretty small and are only doing a single thing
and are mainly moving code around. The actual work is happening in the last commits which is adding tests for `PresetsDataView`.


src/DataViews/include/DataViews/DataViewTypes.h
```diff
@@ -1,9 +1,11 @@
-// Copyright (c) 2020 The Orbit Authors. All rights reserved.
+// Copyright (c) 2021 The Orbit Authors. All rights reserved.
```
dpallotti, line 1: Consider renaming the file to `DataViewType.h` to match the name of the `enum`.

src/DataViews/PresetsDataViewTest.cpp
```diff
@@ -0,0 +1,441 @@
...
+#include "DataViews/PresetsDataView.h"
+#include "MetricsUploader/MetricsUploaderStub.h"
+#include "MockAppInterface.h"
+#include "OrbitBase/File.h"
+#include "OrbitBase/ReadFileToString.h"
+#include "OrbitBase/TemporaryFile.h"
+#include "OrbitBase/TestUtils.h"
+#include "PresetFile/PresetFile.h"
+#include "preset.pb.h"
+namespace {
```
dpallotti, line 27:
```suggestion
#include "preset.pb.h"

namespace {
```

src/DataViews/PresetsDataViewTest.cpp
```diff
@@ -0,0 +1,441 @@
...
+
+  // We don't enforce an order here. That might change in the future though.
+  EXPECT_THAT(absl::StrSplit(view.GetValue(0, 2), '\n'),
+              testing::UnorderedElementsAre("main_module", "other_module"));
+  // Column 3 lists the number of functions in each module.
+  EXPECT_THAT(absl::StrSplit(view.GetValue(0, 3), '\n'),
+              testing::UnorderedElementsAre(std::to_string(module0.function_names_size()),
+                                            std::to_string(module1.function_names_size())));
+}
+TEST(PresetsDataView, Sorting) {
```
dpallotti, line 186:
```suggestion
}

TEST(PresetsDataView, Sorting) {
```

src/DataViews/PresetsDataViewTest.cpp
```diff
@@ -0,0 +1,441 @@
...
+    return std::tie(lhs.r, lhs.g, lhs.b) == std::tie(rhs.r, rhs.g, rhs.b);
+  }
+  friend bool operator!=(const Color& lhs, const Color& rhs) { return !(lhs == rhs); }
+};
+}  // namespace
+
+TEST(PresetsDataView, Columns) {
+  orbit_metrics_uploader::MetricsUploaderStub metrics_uploader{};
+  orbit_data_views::MockAppInterface app{};
+  orbit_data_views::PresetsDataView view{&app, &metrics_uploader};
```
dpallotti, line 42: You could use a test fixture and avoid redefining this for every test.

dpallotti (approved):
Thanks!

src/DataViews/include/DataViews/AppInterface.h
```diff
@@ -5,11 +5,27 @@
...
+#include "capture_data.pb.h"
+
 namespace orbit_data_views {
 
 class AppInterface {
  public:
   virtual ~AppInterface() noexcept = default;
+
+  virtual void SetClipboard(const std::string& contents) = 0;
+  [[nodiscard]] virtual std::string GetSaveFile(const std::string& extension) const = 0;
```
dimitry-, line 28: what is savefile? also should it return std::filesystem::path?

src/DataViews/PresetsDataViewTest.cpp
```diff
@@ -0,0 +1,441 @@
...
+#include "OrbitBase/ReadFileToString.h"
+#include "OrbitBase/TemporaryFile.h"
+#include "OrbitBase/TestUtils.h"
+#include "PresetFile/PresetFile.h"
+#include "preset.pb.h"
+namespace {
+// This is just a helper type to handle colors. Note that Color from `OrbitGl/CoreMath.h` is not
+// available in this module.
+struct Color {
+  unsigned char r, g, b;
```
dimitry-, line 31: nit: uint8_t?

src/DataViews/PresetsDataViewTest.cpp
```diff
@@ -0,0 +1,441 @@
...
+}
+
+TEST(PresetsDataView, LoadState) {
+  orbit_metrics_uploader::MetricsUploaderStub metrics_uploader{};
+  orbit_data_views::MockAppInterface app{};
+  orbit_data_views::PresetsDataView view{&app, &metrics_uploader};
+
+  // GetPresetLoadState is called once per `GetValue` and `GetToolTip` call.
+  auto load_state = orbit_data_views::PresetLoadState::kLoadable;
+  EXPECT_CALL(app, GetPresetLoadState).Times(9).WillRepeatedly(testing::ReturnPointee(&load_state));
```
dimitry-, line 66: Do we want to fix the number of times this method called in the test? You are using AnyNumber in other tests...

dimitry-:
I am going to look more closely into tests on my second pass later today.

src/DataViews/PresetsDataViewTest.cpp
```diff
@@ -0,0 +1,441 @@
...
+struct Color {
+  unsigned char r, g, b;
+  friend bool operator==(const Color& lhs, const Color& rhs) {
+    return std::tie(lhs.r, lhs.g, lhs.b) == std::tie(rhs.r, rhs.g, rhs.b);
+  }
+  friend bool operator!=(const Color& lhs, const Color& rhs) { return !(lhs == rhs); }
+};
+}  // namespace
+
+TEST(PresetsDataView, Columns) {
```
dimitry-, line 39: ColumnHeadersNotEmpty

src/DataViews/PresetsDataViewTest.cpp
```diff
@@ -0,0 +1,441 @@
...
+TEST(PresetsDataView, Empty) {
+  orbit_metrics_uploader::MetricsUploaderStub metrics_uploader{};
+  orbit_data_views::MockAppInterface app{};
+  orbit_data_views::PresetsDataView view{&app, &metrics_uploader};
+
+  EXPECT_EQ(view.GetNumElements(), 0);
+  EXPECT_EQ(view.GetLabel(), "Presets");
+}
+
+TEST(PresetsDataView, LoadState) {
```
dimitry-, line 59: CheckLabelAndColorForLoadStates

In general it is better to name test after scenarios rather then members, so that it is clearer what is being tested.

src/DataViews/PresetsDataViewTest.cpp
```diff
@@ -0,0 +1,441 @@
...
+
+  // We don't test for specific color values here, but we will ensure that the colors are different,
+  // hence that the load state is indicated by color.
+  EXPECT_TRUE(view.WantsDisplayColor());
+  EXPECT_NE(color_loadable_state, color_partially_loadable_state);
+  EXPECT_NE(color_loadable_state, color_not_loadable_state);
+  EXPECT_NE(color_partially_loadable_state, color_not_loadable_state);
+}
+
+TEST(PresetsDataView, PresetName) {
```
dimitry-, line 103: PresetNameIsFileName?

src/DataViews/PresetsDataViewTest.cpp
```diff
@@ -0,0 +1,441 @@
...
+  orbit_client_protos::PresetInfo preset_info0{};
+  const std::filesystem::path preset_filename0{"/path/filename.xyz"};
+  orbit_preset_file::PresetFile preset_file0{preset_filename0, preset_info0};
+  view.SetPresets({preset_file0});
+
+  EXPECT_EQ(view.GetNumElements(), 1);
+  EXPECT_EQ(view.GetValue(0, 1), preset_filename0.filename().string());
+}
+
+TEST(PresetsDataView, OnDataChanged) {
```
dimitry-, line 117: ViewIsUpdatedAfterSetPresets?

dimitry- (approved):
otherwise LGTM

src/DataViews/include/DataViews/DataViewTypes.h
beckerhe: Yes, good point. Done.

src/DataViews/PresetsDataViewTest.cpp
beckerhe: Done.

src/DataViews/PresetsDataViewTest.cpp
beckerhe: Done.

src/DataViews/include/DataViews/AppInterface.h
beckerhe: `GetSaveFile` opens a Qt file picker dialog and returns a file path. Yes it could probably return a std::filesystem::path, but this commit is merely defining an abstract interface for an already existing function. Improving on that is out of scope or do you see a need this to change right now that I missed?

src/DataViews/PresetsDataViewTest.cpp
beckerhe: Done.

src/DataViews/PresetsDataViewTest.cpp
beckerhe: This was very much intentional. Since this test is about the preset load state I check the number of calls here. But I won't check it in the other tests since there it's not under test. If somethings breaks around `GetPresetLoadState` ideally only this test fails and it will be easy to pinpoint the problem. Does that make sense?

src/DataViews/PresetsDataViewTest.cpp
beckerhe: Yes. Done.

src/DataViews/PresetsDataViewTest.cpp
beckerhe: Yes. Thanks for suggestion some better names. Is also changed some more.

src/DataViews/PresetsDataViewTest.cpp
beckerhe: Done.

src/DataViews/PresetsDataViewTest.cpp
beckerhe: Done.

src/DataViews/PresetsDataViewTest.cpp
beckerhe: Done.

src/DataViews/include/DataViews/AppInterface.h
dimitry-: it is out of scope yes.

antonrohr (approved):
Very nice! Thank you for splitting it up in so many commits! 
Only nits regarding namespace orbit_data_views

src/DataViews/PresetsDataViewTest.cpp
```diff
@@ -0,0 +1,402 @@
...
+}
+
+TEST_F(PresetsDataViewTest, Empty) {
+  EXPECT_EQ(view_.GetNumElements(), 0);
+  EXPECT_EQ(view_.GetLabel(), "Presets");
+}
+
+TEST_F(PresetsDataViewTest, CheckLabelAndColorForLoadStates) {
+  // GetPresetLoadState is called once per `GetValue` and `GetToolTip` call.
+  auto load_state = orbit_data_views::PresetLoadState::kLoadable;
```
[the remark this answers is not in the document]
beckerhe: What would be the advantage?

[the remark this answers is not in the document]
antonrohr: A little more readability, just by having less text. What would be the disadvantage?

src/OrbitGl/App.cpp
```diff
@@ -482,13 +482,15 @@ void OrbitApp::UpdateModulesAbortCaptureIfModuleWithoutBuildIdNeedsReload(
 void OrbitApp::OnModuleUpdate(uint64_t /*timestamp_ns*/, ModuleInfo module_info) {
   UpdateModulesAbortCaptureIfModuleWithoutBuildIdNeedsReload({module_info});
   GetMutableCaptureData().mutable_process()->AddOrUpdateModuleInfo(module_info);
-  main_thread_executor_->Schedule([this]() { FireRefreshCallbacks(DataViewType::kLiveFunctions); });
+  main_thread_executor_->Schedule(
+      [this]() { FireRefreshCallbacks(orbit_data_views::DataViewType::kLiveFunctions); });
```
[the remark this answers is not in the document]

src/OrbitQt/orbitmainwindow.cpp
```diff
@@ -158,14 +158,17 @@ OrbitMainWindow::OrbitMainWindow(orbit_qt::TargetConfiguration target_configurat
...
-  ui->ModulesList->Initialize(data_view_factory->GetOrCreateDataView(DataViewType::kModules),
-                              SelectionType::kExtended, FontType::kDefault);
-  ui->FunctionsList->Initialize(data_view_factory->GetOrCreateDataView(DataViewType::kFunctions),
-                                SelectionType::kExtended, FontType::kDefault);
-  ui->PresetsList->Initialize(data_view_factory->GetOrCreateDataView(DataViewType::kPresets),
-                              SelectionType::kDefault, FontType::kDefault,
-                              /*is_main_instance=*/true, /*uniform_row_height=*/false,
-                              /*text_alignment=*/Qt::AlignTop | Qt::AlignLeft);
+  ui->ModulesList->Initialize(
+      data_view_factory->GetOrCreateDataView(orbit_data_views::DataViewType::kModules),
```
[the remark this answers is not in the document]
beckerhe: done.

src/DataViews/PresetsDataViewTest.cpp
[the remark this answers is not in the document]
beckerhe: done.

src/DataViews/PresetsDataView.cpp
```diff
@@ -39,15 +39,18 @@ constexpr const float kModulesColumnWidth = 0.34f;
...
 }
 }  // namespace
 
-PresetsDataView::PresetsDataView(OrbitApp* app,
+namespace orbit_data_views {
+
+PresetsDataView::PresetsDataView(orbit_data_views::AppInterface* app,
                                  orbit_metrics_uploader::MetricsUploader* metrics_uploader)
-    : DataView(DataViewType::kPresets, app), metrics_uploader_(metrics_uploader) {}
+    : orbit_data_views::DataView(orbit_data_views::DataViewType::kPresets, app),
```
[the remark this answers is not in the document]
beckerhe: done.

PR #2430, opened 2021-06-15
Remove most of usages of GetAbsoluteAddress

dimitry-:
This change removes almost all usages of GetAbsoluteAddress and
deprecates the function. We have left with disassebly use-case still
relying on absolute addresses. Since it requires service-side support
this part is left for the follow-up PRs.

Bug: http://b/190805289
Test: Start Orbit, instrument a function and take a capture.

### DIFF
--- a/src/ClientData/FunctionUtils.cpp
+++ b/src/ClientData/FunctionUtils.cpp
@@ -5,6 +5,7 @@
 #include "ClientData/FunctionUtils.h"
 
 #include <absl/strings/str_cat.h>
+#include <absl/strings/str_join.h>
 
 #include <filesystem>
 #include <utility>
@@ -44,12 +45,25 @@ uint64_t Offset(const FunctionInfo& func, const ModuleData& module) {
 
 std::optional<uint64_t> GetAbsoluteAddress(const orbit_client_protos::FunctionInfo& func,
                                            const ProcessData& process, const ModuleData& module) {
-  std::optional<uint64_t> base_address = process.GetModuleBaseAddress(func.module_path());
-  if (!base_address.has_value()) {
+  std::vector<uint64_t> base_addresses =
+      process.GetModuleBaseAddresses(module.file_path(), module.build_id());
+  if (base_addresses.empty()) {
     return std::nullopt;
   }
 
-  return func.address() + base_address.value() - module.load_bias();
+  if (base_addresses.size() > 1) {
+    ERROR(
+        "Found multiple mappings for \"%s\" with build_id=%s [%s]: "
+        "will use the first one as a base address",
+        module.file_path(), module.build_id(),
+        absl::StrJoin(base_addresses, ",", [](std::string* out, uint64_t address) {
+          return out->append(absl::StrFormat("%#x", address));
+        }));
+  }
+
+  CHECK(!base_addresses.empty());
+
+  return func.address() + base_addresses.at(0) - module.load_bias();
 }
 
 bool IsOrbitFunctionFromType(const FunctionInfo::OrbitType& type) {
--- a/src/ClientData/ProcessDataTest.cpp
+++ b/src/ClientData/ProcessDataTest.cpp
@@ -24,6 +24,8 @@ using orbit_base::HasNoError;
 using orbit_grpc_protos::ModuleInfo;
 using orbit_grpc_protos::ProcessInfo;
 
+using testing::ElementsAre;
+
 namespace orbit_client_data {
 
 TEST(ProcessData, Constructor) {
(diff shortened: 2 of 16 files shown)


src/ClientData/FunctionUtils.cpp
```diff
@@ -44,12 +45,23 @@ uint64_t Offset(const FunctionInfo& func, const ModuleData& module) {
...
+  if (base_address.size() > 1) {
+    ERROR(
+        "Found multiple mappings for \"%s\" with build_id=%s [%s]: "
+        "will use the first one as a base address",
+        module.file_path(), module.build_id(),
+        absl::StrJoin(base_address, ",", [](std::string* out, uint64_t address) {
+          return out->append(absl::StrFormat("%#x", address));
+        }));
+  }
+
```
florian-kuebler, line 63: Nit: Maybe add a CHECK(base_address.size() == 0) here.

src/ClientModel/CaptureData.cpp
```diff
@@ -175,15 +175,21 @@ const std::string& CaptureData::GetFunctionNameByAddress(uint64_t absolute_addre
 // callstack.
 std::optional<uint64_t> CaptureData::FindFunctionAbsoluteAddressByAddress(
```
florian-kuebler, line 176: Kind of unrelated to this change, but I find the name (including the parameter name) pretty confusing. Maybe we can use the opportunity to improve here.

src/OrbitGl/FunctionsDataView.cpp
```diff
@@ -108,23 +108,8 @@ std::string FunctionsDataView::GetValue(int row, int column) {
...
-      CHECK(process != nullptr);
-      const ModuleData* module =
-          app_->GetModuleByPathAndBuildId(function.module_path(), function.module_build_id());
-      CHECK(module != nullptr);
-      return absl::StrFormat("0x%llx", orbit_client_data::function_utils::GetAbsoluteAddress(
-                                           function, *process, *module)
-                                           .value_or(0));
-    }
+    case kColumnAddress:
+      return absl::StrFormat("%#" PRIx64, function.address());
```
florian-kuebler, line 112: So we are only changing the address in those data view to be the relative one and not in the disassembly? Which addresses are displayed in the sampling view? We should take care that we remain somehow consistent. E.g. if I have a callstack sample and go to disassembly, I should be able to immediately correlate that sample with the instruction. I think this is still the case, but just wanted to check back.

dimitry-: Sampling view uses absolute addresses. But here we do not have them because function list in this view is coming from local modules not server-side mapping so it is fine to have elf-addresses here. Note also that sort currently uses relative addresses which was not consistent with getvalue.

florian-kuebler: Makes sense. Thanks!

florian-kuebler (approved):

florian-kuebler:
LGTM!

src/ClientModel/CaptureData.cpp
dimitry-: Do you have a name in mind?

src/ClientData/FunctionUtils.cpp
dimitry-: we have an early return in line 50

dimitry-: I will add it anyways.

src/ClientData/include/ClientData/FunctionUtils.h
```diff
@@ -34,7 +34,11 @@ namespace function_utils {
 [[nodiscard]] uint64_t Offset(const orbit_client_protos::FunctionInfo& func,
                               const ModuleData& module);
 
-[[nodiscard]] std::optional<uint64_t> GetAbsoluteAddress(
+// This function should not be used, since it could return incomplete or invalid
+// result in the case when one module is mapped 2 or more times. Try to find another way
+// of solving the problem, for example by converting an absolute address to a module offset and
+// then operating on that.
```
antonrohr, line 40: I dont know if we have a bug for deprecating the "old" capture format. Iff we do, please add a TODO

antonrohr (approved):
lgtm. One nit

src/ClientData/FunctionUtils.cpp
dpallotti: Florian, you mean `== 1`?

src/ClientData/include/ClientData/FunctionUtils.h
```diff
@@ -34,7 +34,11 @@ namespace function_utils {
 [[nodiscard]] uint64_t Offset(const orbit_client_protos::FunctionInfo& func,
                               const ModuleData& module);
 
-[[nodiscard]] std::optional<uint64_t> GetAbsoluteAddress(
+// This function should not be used, since it could return incomplete or invalid
+// result in the case when one module is mapped 2 or more times. Try to find another way
```
dpallotti, line 38:
```suggestion
// result in the case when one module is mapped two or more times. Try to find another way
```
Wrote this on one of Anton's recent changes, I feel like I should do the same here :)

src/ClientModel/CaptureData.cpp
dpallotti: Yeah, it's confusing, looking at the name it's almost like it just returns the argument. Maybe `FindFunctionAbsoluteAddressByInstructionAbsoluteAddress` (if I even got it correctly)?

src/OrbitGl/LiveFunctionsDataView.cpp
```diff
@@ -93,10 +93,8 @@ std::string LiveFunctionsDataView::GetValue(int row, int column) {
       return GetPrettyTime(absl::Nanoseconds(stats.max_ns()));
     case kColumnModule:
       return function.module_path();
-    case kColumnAddress: {
-      const CaptureData& capture_data = app_->GetCaptureData();
-      return absl::StrFormat("0x%" PRIx64, capture_data.GetAbsoluteAddress(function).value_or(0));
-    }
+    case kColumnAddress:
+      return absl::StrFormat("%#" PRIx64, function.address());
```
dpallotti, line 97: I think that we don't need `PRIx64` with `absl::StrFormat`, we can just use `"%#x"`.

src/OrbitGl/FunctionsDataView.cpp
dpallotti, line 112: I agree we need to be careful. We use absolute addresses everywhere else, so this is confusing. I would suggest to at least rename the column, e.g., to "Address in module".

src/ClientModel/CaptureData.cpp
```diff
@@ -175,15 +175,21 @@ const std::string& CaptureData::GetFunctionNameByAddress(uint64_t absolute_addre
...
+
+  const ModuleData* module =
+      module_manager_->GetModuleByPathAndBuildId(module_path, module_build_id);
+  if (module == nullptr) return std::nullopt;
+
+  const uint64_t offset = absolute_address - module_base_address;
+  const auto* function_info = module->FindFunctionByOffset(offset, false);
+  if (function_info == nullptr) return std::nullopt;
+
+  return module_base_address - module->load_bias() + function_info->address();
```
dimitry-, line 192: I kind of accidentally dropped address-info related logic from here. Will need to revisit.

src/ClientData/FunctionUtils.cpp
florian-kuebler: For sure, I meant `== 1`, sorry. Also, yes that is not needed, as we do have the early two early returns, and probably the code will not evolve too much, but I feel a bit saver with the check (e.g. if we add some code after the early return).

src/ClientData/include/ClientData/FunctionUtils.h
dimitry-: done.

src/ClientModel/CaptureData.cpp
dimitry-: done.

src/OrbitGl/FunctionsDataView.cpp
dimitry-: done.

src/OrbitGl/LiveFunctionsDataView.cpp
dimitry-: done.

src/ClientModel/CaptureData.cpp
dimitry-: done.

src/ClientData/FunctionUtils.cpp
dimitry-: done.

src/ClientData/FunctionUtils.cpp
```diff
@@ -44,12 +45,25 @@ uint64_t Offset(const FunctionInfo& func, const ModuleData& module) {
 
 std::optional<uint64_t> GetAbsoluteAddress(const orbit_client_protos::FunctionInfo& func,
                                            const ProcessData& process, const ModuleData& module) {
-  std::optional<uint64_t> base_address = process.GetModuleBaseAddress(func.module_path());
-  if (!base_address.has_value()) {
+  std::vector<uint64_t> base_address =
```
dpallotti, line 48: `base_addresses`

src/ClientData/ProcessDataTest.cpp
```diff
@@ -240,33 +242,71 @@ TEST(ProcessData, IsModuleLoadedByProcess) {
 
 TEST(ProcessData, GetModuleBaseAddress) {
```
dpallotti, line 243:
```suggestion
TEST(ProcessData, GetModuleBaseAddresses) {
```

src/OrbitGl/FunctionsDataView.cpp
```diff
@@ -108,23 +108,8 @@ std::string FunctionsDataView::GetValue(int row, int column) {
...
-      }
-      CHECK(process != nullptr);
-      const ModuleData* module =
-          app_->GetModuleByPathAndBuildId(function.module_path(), function.module_build_id());
-      CHECK(module != nullptr);
-      return absl::StrFormat("0x%llx", orbit_client_data::function_utils::GetAbsoluteAddress(
-                                           function, *process, *module)
-                                           .value_or(0));
-    }
+    case kColumnAddress:
```
dpallotti, line 111: Maybe rename the constant as well to `kColumnAddressInModule` (although these names are not very precise).

dpallotti (approved):

src/ClientData/FunctionUtils.cpp
dimitry-: done.

src/ClientData/ProcessDataTest.cpp
dimitry-: done.

src/ClientData/include/ClientData/FunctionUtils.h
dimitry-: added todo for disassembly bug - once this is fixed this function can be removed.

PR #2431, opened 2021-06-15
Refactor client for unit tests

antonrohr:
I suppressed my urge to further refactor this (aka completely rewrite the class), its basically the bare minimum to make it testable. The plan is to plug in a mock ggp client. The timeout is also a parameter, to be able to test a timeout without waiting 10seconds. 

http://b/185089025

### DIFF
--- a/src/OrbitGgp/include/OrbitGgp/Client.h
+++ b/src/OrbitGgp/include/OrbitGgp/Client.h
@@ -9,6 +9,7 @@
 #include <QPointer>
 #include <QString>
 #include <QVector>
+#include <chrono>
 #include <functional>
 #include <outcome.hpp>
 #include <string>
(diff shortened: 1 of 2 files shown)


src/OrbitGgp/include/OrbitGgp/Client.h
```diff
@@ -17,21 +18,34 @@
 #include "OrbitBase/Result.h"
 #include "SshInfo.h"
 
+namespace {
+
+constexpr const char* kDefaultGgpProgram{"ggp"};
+constexpr std::chrono::milliseconds kDefaultTimeout{10'000};
+
+}  // namespace
```
antonrohr, line 26: Clang-tidy is upset about anonymous namespaces in header files and that variable definitions can lead to ODR violations. I would appreciate a suggestion on how to make this better.

beckerhe: If you want to declare and define variables in a header file you want to mark them `inline` which avoids the ODR issue.
Since `constexpr` implies `inline` you can just leave out the anonymous namespace.

beckerhe (approved):
Otherwise

src/OrbitGgp/include/OrbitGgp/Client.h
antonrohr: We talked about the constexpr and namespace situation before. I did forget about it :/ Thank you for the explanation!

PR #2432, opened 2021-06-15
Disable exceptions in our code

dimitry-:
Compile everything with -fno-exceptions

Test: builds


beckerhe (approved):
LGTM. Let's see if that makes problems. :-)

PR #2433, opened 2021-06-16
Fix Windows build problems

beckerhe:
This is hopefully fixing the Windows build problems.

My working theory is that the build whenever windeployqt is called twice simultaneously, so basically a race condition that only happens sometimes.


danielfenner (approved):
Thanks a lot for looking into this!  lgtm!

antonrohr (approved):

PR #2434, opened 2021-06-16
Add tests for OrbitGgp::Client

antonrohr:
http://b/185089025

### DIFF
--- a/src/OrbitGgp/CMakeLists.txt
+++ b/src/OrbitGgp/CMakeLists.txt
@@ -36,12 +36,22 @@ target_link_libraries(
 set_target_properties(OrbitGgp PROPERTIES AUTOMOC ON)
 set_target_properties(OrbitGgp PROPERTIES AUTOUIC ON)
 
+# Mock ggp clients
+add_executable(OrbitMockGgpFailing)
+target_compile_options(OrbitMockGgpFailing PRIVATE ${STRICT_COMPILE_FLAGS})
+target_sources(OrbitMockGgpFailing PRIVATE MockGgp/Failing.cpp)
+
+add_executable(OrbitMockGgpWorking)
+target_compile_options(OrbitMockGgpWorking PRIVATE ${STRICT_COMPILE_FLAGS})
+target_sources(OrbitMockGgpWorking PRIVATE MockGgp/Working.cpp)
+
 # Tests
 add_executable(OrbitGgpTests)
 
 target_compile_options(OrbitGgpTests PRIVATE ${STRICT_COMPILE_FLAGS})
 
 target_sources(OrbitGgpTests PRIVATE 
+          ClientTest.cpp
           InstanceTests.cpp
           InstanceItemModelTests.cpp
           SshInfoTests.cpp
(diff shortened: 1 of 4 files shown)


src/OrbitGgp/CMakeLists.txt
```diff
@@ -36,12 +36,22 @@ target_link_libraries(
 set_target_properties(OrbitGgp PROPERTIES AUTOMOC ON)
 set_target_properties(OrbitGgp PROPERTIES AUTOUIC ON)
 
+# Mock ggp cli
```
vickyliu-go4it, line 39: Nit: ```# Mock ggp clients```

src/OrbitGgp/MockGgp/Working.cpp
```diff
@@ -0,0 +1,81 @@
...
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#include <chrono>
+#include <cstring>
+#include <iostream>
+#include <thread>
+
+int GgpVersion(char* argv[]) {
+  if (strcmp(argv[1], "version") != 0) {
```
beckerhe, line 11: I think we should refrain from using C-functions and stick to C++. In this case that would mean casting to `std::string_view` and use its internal comparison functionality. Something like: `std::string_view{argv[1]} == "version"`.

src/OrbitGgp/MockGgp/Working.cpp
```diff
@@ -0,0 +1,81 @@
...
+ {\n\
+  \"displayName\": \"displayName-2\",\n\
+  \"id\": \"id/of/instance2\",\n\
+  \"ipAddress\": \"123.456.789.012\",\n\
+  \"lastUpdated\": \"2012-12-12T12:12:12Z\",\n\
+  \"owner\": \"owner@\",\n\
+  \"pool\": \"pool-of-test_instance_2\"\n\
+ }\n\
+]\n\
+";
```
beckerhe, line 45: The JSON strings would get more readable with raw string literals. This would eliminate the backslash at the end of each line, the newline characters and the escaping for quotes. WDYT?

src/OrbitGgp/MockGgp/Working.cpp
```diff
@@ -0,0 +1,81 @@
...
+ \"knownHostsPath\": \"example/path/to/known_hosts\",\n\
+ \"port\": \"12345\",\n\
+ \"user\": \"example_user\"\n\
+}\n\
+";
+  return 0;
+}
+
+int main(int argc, char* argv[]) {
+  std::this_thread::sleep_for(std::chrono::milliseconds{30});
```
beckerhe, line 68: Could you add a comment explaining this waiting? I assume the idea is to test the timeout feature?

beckerhe (approved):
Otherwise LGTM

src/OrbitGgp/MockGgp/Working.cpp
antonrohr: This is a great idea. I did not think of raw string literals while I wrote this. Fixed!

src/OrbitGgp/MockGgp/Working.cpp
antonrohr: done

src/OrbitGgp/MockGgp/Working.cpp
antonrohr: done

src/OrbitGgp/CMakeLists.txt
antonrohr: done

PR #2435, opened 2021-06-16
Add and send some `MetadataEvent`s, show in new capture log, show some as message boxes

dpallotti:
#### Add capture log QTextEdit and corresponding QPushButton
The QTextEdit is shown when capturing in the bottom right and hidden when the
capture is finished. The checkable button in the status bar allows to hide it
while capturing or to re-show it when the capture is done.
Also added the first log lines: capture started and finished.

Bug: http://b/188182133

Test: Take a capture, verify expected behavior.
#### Add MainWindowInterface::ShowWarningWithDontShowAgainCheckboxIfNeeded
For now, use it to replace
`OrbitMainWindow::ShowEmptyFrameTrackWarningIfNeeded`.

Test: Hook a function that is not called, capture, then enable the frame track.
#### Fix initial default value of "Enable Orbit Api in target"
Bug: http://b/191236151

Test: Delete all settings, capture a binary with manual instrumentation.
#### Add MetadataEvent with some initial sub-events and handle it/them
Bug: http://b/188182133
Bug: http://b/180235290
Bug: http://b/187176721
#### Don't report "No api table to initialize" as an error
This is not really an error, as it's the equivalent of having enabled all the
known api tables, like in the loop below.

An alternative would be to also have
`if (capture_options.api_functions().empty())` in the callers and keep the error
here.

Test: Capture a binary with no manual instrumentation, and a binary with manual
instrumentation but symbols not loaded.
#### Don't distinguish between uprobes and other perf_event_open for errors
While a long time ago a low `perf_event_paranoid` was enough without root for
everything except uprobes, this no longer holds, as reading tracepoint ids
also requires root.
#### Add, send, process ErrorsWithPerfEventOpenEvent
Bug: http://b/188182133
Bug: http://b/180235290

Test: Take a capture with `OrbitService` not running as root. Update
`LinuxTracingIntegrationTests`.
#### Don't show capture warning message boxes when loading a capture


src/OrbitGl/App.cpp
```diff
@@ -266,10 +269,28 @@ OrbitApp::~OrbitApp() {
 }
 
 void OrbitApp::OnCaptureFinished(const CaptureFinished& capture_finished) {
-  if (capture_finished.status() == CaptureFinished::kFailed) {
-    SendErrorToUi("Capture Failed", capture_finished.error_message());
-    ERROR("Capture Finished with error: %s", capture_finished.error_message());
-  }
+  main_thread_executor_->Schedule([this, capture_finished]() {
+    switch (capture_finished.status()) {
+      case orbit_grpc_protos::CaptureFinished_Status_kSuccessful: {
```
beckerhe, line 274: Nit: Using the abbreviated constants like `CaptureFinished::kFailed` etc. would make it more readable. At least to me.

src/OrbitGl/MainWindowInterface.h
```diff
@@ -32,6 +32,10 @@ class MainWindowInterface {
                                const std::string& assembly,
                                orbit_code_report::DisassemblyReport report) = 0;
 
+  enum class CaptureLogSeverity { kInfo, kWarning, kSevereWarning, kError };
+  virtual void AppendToCaptureLog(CaptureLogSeverity severity, std::string_view capture_time,
```
beckerhe, line 40: From an interface perspect I would expect the capture_time to be a chrono or abseil timestamp or duration, not a string, such that formatting happens in a single place. Not sure though if that change is worth it. I assume you built this around existing functions which return strings?

src/OrbitQt/orbitmainwindow.cpp
```diff
@@ -436,6 +437,32 @@ void OrbitMainWindow::SetupTargetLabel() {
                    });
 }
 
+void OrbitMainWindow::SetupStatusBarLogButton() {
+  // The Qt Designer doesn't seem to support adding children to a StatusBar.
+  auto* capture_log_widget = new QWidget(statusBar());
```
beckerhe, line 439: Nit: You could add a `// NOLINT` or `// NOLINTNEXTLINE` comment to silence clang-tidy. I assume it will complain about the raw `new`.

beckerhe (approved):
Nice. Otherwise LGTM

src/OrbitGl/App.cpp
dpallotti: Oh yeah, the cases were autogenerated by the IDE and I forgot to change them. Done.

src/OrbitQt/orbitmainwindow.cpp
dpallotti: Oh thanks, I like that, added a couple in this method.

src/OrbitGl/MainWindowInterface.h
dpallotti: Yep, basically you said it all, I agree the interface is weird, but at the same time this uses `OrbitApp`'s `GetCaptureTime` and the new `GetCaptureTimeAt`, which already return formatted strings. 

Internally these methods call `GetPrettyTime` so I could just call that directly in `OrbitMainWindow` (although `GetPrettyTime` is currently only called in `OrbitGl`). That would also allow the capture log to choose to change the way it formats timestamps. In that case I think that `GetCaptureTime` should just be changed to return an `absl::Duration` and all the places it's called from should use `GetPrettyTime` (there are not that many).

What do you think?

beckerhe:
> Nice. Otherwise LGTM

I just realized I haven't reviewed all the commits.

src/OrbitGl/MainWindowInterface.h
beckerhe: Yes, that sounds perfect to me!

beckerhe (approved):
So, this time I went through all the commits. This looks all good to me, but there are quite some changes in areas I'm not super familiar with. So I suggest at least one other dev also has a look.

src/OrbitGl/App.cpp
```diff
@@ -926,6 +1036,18 @@ std::string OrbitApp::GetCaptureTime() const {
...
 }
 
+std::string OrbitApp::GetCaptureTimeAt(uint64_t timestamp_ns) const {
+  const TimeGraph* time_graph = GetTimeGraph();
+  if (time_graph == nullptr) {
+    return GetPrettyTime(absl::ZeroDuration());
+  }
+  const uint64_t capture_min_timestamp_ns = time_graph->GetCaptureMin();
+  if (timestamp_ns < capture_min_timestamp_ns) {
+    return GetPrettyTime(absl::ZeroDuration());
```
pierricgimmig, line 1046: A negative time could be useful to differentiate from the case where time_graph is null and to give and idea about how far from the capture min we are.

src/OrbitQt/orbitmainwindow.cpp
```diff
@@ -436,6 +434,52 @@ void OrbitMainWindow::SetupTargetLabel() {
                    });
 }
 
+void OrbitMainWindow::SetupStatusBarLogButton() {
+  // The Qt Designer doesn't seem to support adding children to a StatusBar.
+  auto* capture_log_widget = new QWidget(statusBar());  // NOLINT
```
pierricgimmig, line 439: Out of curiosity, what is the linter complaining about?

src/OrbitQt/orbitmainwindow.cpp
```diff
@@ -919,7 +970,7 @@ constexpr uint64_t kMemoryWarningThresholdKbDefaultValue = 1024 * 1024 * 8;  //
 void OrbitMainWindow::LoadCaptureOptionsIntoApp() {
   QSettings settings;
   app_->SetCollectThreadStates(settings.value(kCollectThreadStatesSettingKey, false).toBool());
-  app_->SetEnableApi(settings.value(kEnableApiSettingKey, false).toBool());
+  app_->SetEnableApi(settings.value(kEnableApiSettingKey, true).toBool());
```
pierricgimmig, line 973: thanks!

pierricgimmig (approved):
This is cool! Thanks. LGTM, please see minor comments. Also, I tried it out and the capture log automatically disappears on capture end when we automatically switch to the sampling tab. It would have expected it to stick.

src/OrbitQt/orbitmainwindow.cpp
dpallotti: clang-tidy complains (rightly so) that the result of `new` is not passed to an owner.

src/GrpcProtos/capture.proto
```diff
@@ -409,6 +409,43 @@ message CaptureStarted {
...
+}
+
+message ClockResolutionEvent {
+  uint64 timestamp_ns = 1;
+  uint64 clock_resolution_ns = 2;
+}
+
+message ErrorsWithPerfEventOpenEvent {
+  uint64 timestamp_ns = 1;
+}
```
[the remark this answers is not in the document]
dpallotti: We can pass an additional message or `enum`, but what level of detail do you have in mind? What class of events failed to open (e.g., sampling, dynamic instrumentation), or more specific? One problem with being specific is that basically every times there's one failure, there are many failure, so I feel like we would return either only the first one, or a ton of them.

dpallotti:
> the capture log automatically disappears on capture end when we automatically switch to the sampling tab

@pierricgimmig This is intended, the code for this is at the end of `OrbitMainWindow::UpdateCaptureStateDependentWidgets`. The reasoning is that the user has already seen the messages during the capture, when the log takes vertical space from the right but most of the data is on the left (unless you have instrumented many functions and you want to see all of them in the "Live" tab), while after the capture the messages are in most cases no longer interesting, and I would prefer to make space for all the data that is now on the right, without the user having the close the log every single time, which I believe would be very annoying.

src/Api/EnableInTracee.cpp
```diff
@@ -33,7 +33,7 @@ namespace {
 ErrorMessageOr<absl::flat_hash_map<std::string, ModuleInfo>> GetModulesByPathForPid(int32_t pid) {
   OUTCOME_TRY(module_infos, orbit_object_utils::ReadModules(pid));
   absl::flat_hash_map<std::string, ModuleInfo> result;
-  for (const ModuleInfo& module_info : module_infos) {
+  for (ModuleInfo& module_info : module_infos) {
```
dimitry-, line 36: interesting. I would have expected compiler to complain on move here.

src/OrbitGl/App.cpp
dpallotti: Hmm, I don't think we should have show negative times. It can happen here if `GetCaptureMin()` is still returning `std::numeric_limits<uint64_t>::max()`, but ideally we should always have non-negative times here (it otherwise indicates an error on our side).

src/OrbitGl/App.cpp
```diff
@@ -266,10 +269,28 @@ OrbitApp::~OrbitApp() {
...
+      case orbit_grpc_protos::CaptureFinished::kFailed: {
+        SendErrorToUi("Capture Failed", capture_finished.error_message());
+        ERROR("Capture Finished with error: %s", capture_finished.error_message());
+        main_window_->AppendToCaptureLog(
+            MainWindowInterface::CaptureLogSeverity::kError, GetCaptureTime(),
+            absl::StrFormat("Capture finished with error: %s.", capture_finished.error_message()));
+      } break;
+      case orbit_grpc_protos::
+          CaptureFinished_Status_CaptureFinished_Status_INT_MIN_SENTINEL_DO_NOT_USE_:
+        [[fallthrough]];
```
dimitry-, line 286: I think in case of completely empty case we do not want to use [[fallthrough]], it actually makes it harder to read.

I would reserve it only for cases where there is some work to be done and after that fallthrought is intended.

src/OrbitGl/MainWindowInterface.h
dpallotti: Done.

src/OrbitGl/App.cpp
```diff
@@ -490,6 +514,92 @@ void OrbitApp::OnModulesSnapshot(uint64_t /*timestamp_ns*/, std::vector<ModuleIn
...
+                                           GetCaptureTimeAt(timestamp_ns), message);
+
+          if (!IsLoadingCapture()) {
+            constexpr const char* kDontShowAgainHighClockResolutionWarningKey =
+                "DontShowAgainHighClockResolutionWarning";
+            main_window_->ShowWarningWithDontShowAgainCheckboxIfNeeded(
+                "High clock resolution", message, kDontShowAgainHighClockResolutionWarningKey);
+          }
+        }
+      } break;
```
dimitry-, line 577: does this want to be a function?

dimitry- (approved):
couple of nits. otherwise lgtm

src/GrpcProtos/capture.proto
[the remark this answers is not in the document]
pierricgimmig: I was thinking of specifying in what context perf_event_open failed, i.e. while enabling sampling, scheduling, uprobes, etc. Maybe adding a string to the event would be enough, this way we could know exactly which events we would not receive in the capture.

src/OrbitQt/orbitmainwindow.cpp
pierricgimmig: Ah ok, thanks.

src/OrbitGl/App.cpp
pierricgimmig: The suggestion was to get more info on when the event occurred for debugging, but maybe we should be stricter here and actually CHECK since receiving an event from before the capture is an error on our side. Perhaps we could display the raw timestamp when we receive such events?

pierricgimmig:
> @pierricgimmig This is intended, the code for this is at the end of `OrbitMainWindow::UpdateCaptureStateDependentWidgets`. The reasoning is that the user has already seen the messages during the capture, when the log takes vertical space from the right but most of the data is on the left (unless you have instrumented many functions and you really really want to see all of them in the "Live" tab), while after the capture the messages are in most cases no longer interesting, and I would prefer to make space for all the data that is now on the right, without the user having the close the log every single time, which I believe would be very annoying.

I see your point. The closer the messages are to the end of the capture, the less likely the user will have time to see them however. Really not a huge deal. Eventually we could add an option to "pin" the log section.

dpallotti (commented):
It looks like I forgot to send this reply on Friday.

dpallotti:
> The closer the messages are to the end of the capture, the less likely the user will have time to see them however.

@pierricgimmig That is true. I'm relying on the fact that at the moment most of the messages come at the very beginning. If/once we add more events (the new dynamic instrumentation should be a new source of messages) we can adjust.

src/OrbitGl/App.cpp
dpallotti: Regarding debugging and raw timestamps, this is not really for our debugging, but to actually display information to the user, so we should keep it clean. For example, raw timestamps are timestamps since boot on the instance and they don't say a lot if we just print them. We should use `LOG` and `ERROR` for these things.

There is a more general problem with how we track capture min (and max) timestamps. The problem here is that `TimeGraph` is doing that and it shouldn't. That's why we can't really `CHECK` here. That is a bit bigger and out of scope from this change. I've created http://b/191631291 and I'll set this aside for now, as it doesn't really affect the capture log here.

src/GrpcProtos/capture.proto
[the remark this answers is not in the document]
dpallotti: Makes sense, looking into this.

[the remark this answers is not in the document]
dpallotti: Done (added a commit).

PR #2436, opened 2021-06-16
Remove manual crashes via nullptr dereferencing.

danielfenner:
Depending on the compiler this is translated into an illegal opcode. It doesn't
add much to the functionality so we just get rid of it.

### DIFF
--- a/src/GrpcProtos/services.proto
+++ b/src/GrpcProtos/services.proto
@@ -98,7 +98,6 @@ service FramePointerValidatorService {
 message CrashOrbitServiceRequest {
   enum CrashType {
     CHECK_FALSE = 0;
-    NULL_POINTER_DEREFERENCE = 1;
     STACK_OVERFLOW = 2;
   }
   CrashType crash_type = 1;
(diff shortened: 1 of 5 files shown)


src/GrpcProtos/services.proto
```diff
@@ -98,7 +98,6 @@ service FramePointerValidatorService {
 message CrashOrbitServiceRequest {
   enum CrashType {
     CHECK_FALSE = 0;
-    NULL_POINTER_DEREFERENCE = 1;
```
beckerhe, line 101: Is there some best-practice with removing enum values here? Should we mark the id as reserved or something?

beckerhe (approved):
LGTM

src/GrpcProtos/services.proto
danielfenner: Yes, deleting from a proto hurt a bit when typing it. But since we do not care about staying compatible with different versions of service <-> client we don't care. And this thing is never serialized.

However we can keep the field and deprecate it which would be the more robust version of handling this. I'll bring this up in a standup.

danielfenner: "deprecated = true" did not work for unclear reason, prefixing the member with OBSOLETE_ is another way to mark deprecated members.

beckerhe: Thanks. SGTM!

PR #2438, opened 2021-06-17
[On release] Fix file name conflict issue

dimitry-:
And sequence number to file name if file already
exists.

Test: Manual - press x many times see that files with suffixes are
      created
Bug: http://b/191247811

### DIFF
--- a/src/OrbitGl/App.cpp
+++ b/src/OrbitGl/App.cpp
@@ -1113,6 +1113,25 @@ static std::unique_ptr<CaptureEventProcessor> CreateCaptureEventProcessor(
                                     orbit_client_model::capture_serializer::GenerateCaptureFileName(
                                         process_name, absl::Now(), "_autosave");
 
+  uint64_t suffix_number = 0;
+  while (true) {
+    auto file_exists_or_error = orbit_base::FileExists(file_path);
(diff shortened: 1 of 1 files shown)


src/OrbitGl/App.cpp
```diff
@@ -1113,6 +1113,25 @@ static std::unique_ptr<CaptureEventProcessor> CreateCaptureEventProcessor(
                                     orbit_client_model::capture_serializer::GenerateCaptureFileName(
                                         process_name, absl::Now(), "_autosave");
 
+  uint64_t suffix_number = 0;
+  while (true) {
+    auto file_exists_or_error = orbit_base::FileExists(file_path);
+    if (file_exists_or_error.has_error()) {
+      ERROR("Unable to stat \"%s\": %s", file_path.string(),
```
dpallotti, line 1120: I think this should be "Unable to check for existence of". (`stat` would return an error if the file doesn't exist, but this is about having an error on the check for existence itself)

dpallotti (approved):
Looks good otherwise.

src/OrbitGl/App.cpp
dimitry-: done.

PR #2440, opened 2021-06-17
Get rid of ProcessData::module_memory_map_

dimitry-:
Also modify FindModuleByPath to return vector of build_ids
instead since this is what the callers need.

Bug: http://b/190805289
Test: Start Orbit - load a preset.

### DIFF
--- a/src/ClientData/ProcessData.cpp
+++ b/src/ClientData/ProcessData.cpp
@@ -61,40 +61,34 @@ const std::string& ProcessData::build_id() const {
 
 void ProcessData::UpdateModuleInfos(absl::Span<const ModuleInfo> module_infos) {
   absl::MutexLock lock(&mutex_);
-  module_memory_map_.clear();
   start_address_to_module_in_memory_.clear();
 
   for (const auto& module_info : module_infos) {
-    {
-      const auto [unused_it, success] = module_memory_map_.try_emplace(
-          module_info.file_path(), module_info.address_start(), module_info.address_end(),
-          module_info.file_path(), module_info.build_id());
-      CHECK(success);
-    }
-    {
-      const auto [unused_it, success] = start_address_to_module_in_memory_.try_emplace(
-          module_info.address_start(), module_info.address_start(), module_info.address_end(),
-          module_info.file_path(), module_info.build_id());
-      CHECK(success);
-    }
+    const auto [unused_it, success] = start_address_to_module_in_memory_.try_emplace(
+        module_info.address_start(), module_info.address_start(), module_info.address_end(),
+        module_info.file_path(), module_info.build_id());
+    CHECK(success);
   }
 }
 
-std::optional<ModuleInMemory> ProcessData::FindModuleByPath(const std::string& module_path) const {
+std::vector<std::string> ProcessData::FindModuleBuildIdsByPath(
+    const std::string& module_path) const {
   absl::MutexLock lock(&mutex_);
-  auto it = module_memory_map_.find(module_path);
-  if (it == module_memory_map_.end()) {
-    return std::nullopt;
+  std::set<std::string> build_ids;
+
+  for (const auto& [unused_address, module_in_memory] : start_address_to_module_in_memory_) {
+    if (module_in_memory.file_path() == module_path) {
+      build_ids.insert(module_in_memory.build_id());
+    }
   }
 
-  return it->second;
+  return {build_ids.begin(), build_ids.end()};
 }
 
 void ProcessData::AddOrUpdateModuleInfo(const ModuleInfo& module_info) {
   absl::MutexLock lock(&mutex_);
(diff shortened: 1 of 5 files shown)


src/ClientData/include/ClientData/ProcessData.h
```diff
@@ -97,8 +95,6 @@ class ProcessData final {
   mutable absl::Mutex mutex_;
   orbit_grpc_protos::ProcessInfo process_info_;
 
-  // This is a map from module_path to the space in memory where that module is loaded
-  absl::node_hash_map<std::string, ModuleInMemory> module_memory_map_;
   std::map<uint64_t, ModuleInMemory> start_address_to_module_in_memory_;
```
beckerhe, line 98: Since we are now iterating quite a lot over this map we would probably benefitting from making it a flat_hash_map. As far as I can see we don't need this to be an ordered container, do we?

src/OrbitGl/App.cpp
```diff
@@ -1694,34 +1694,54 @@ orbit_base::Future<ErrorMessageOr<void>> OrbitApp::LoadSymbols(
...
-  const ModuleData* module_data = module_data_or_error.value();
+  std::vector<const ModuleData*>& modules_data = modules_data_or_error.value();
+
+  if (modules_data.empty()) {
+    return ErrorMessage{
+        absl::StrFormat("Module \"%s\" is not loaded by process.", module_path.string())};
+  }
+
+  // TODO(b/191240539): Ask the user which build_id they prefer.
+  if (modules_data.size() > 1) {
```
beckerhe, line 1735: Just for my understanding: What would be the use-case for this? I assume we expect not to see this situation very often, only with esoteric module setups, right?

beckerhe (approved):
Otherwise LGTM

src/OrbitGl/App.cpp
dimitry-: if a module was reloaded on the fly but mapping for old version of the module is still there (and it might be used for some time). This is not normal use-case as far as I can speculate.

beckerhe: Thanks. Makes sense.

src/ClientData/include/ClientData/ProcessData.h
dimitry-: FindModuleByAddress assumes this is sorted by start_address because it uses upper_bound.

beckerhe: Ahh, missed that. Thanks.

src/OrbitGl/App.cpp
```diff
@@ -1694,34 +1694,54 @@ orbit_base::Future<ErrorMessageOr<void>> OrbitApp::LoadSymbols(
...
+    ERROR("Found multiple build_ids for module \"%s\" build_ids=[%s], will choose the first one",
```
dpallotti, line 1736: I would probably format this better, for example

`"Found multiple build_ids (%s) for module \"%s\", will choose the first one"`

src/OrbitGl/App.cpp
```diff
@@ -1694,34 +1694,54 @@ orbit_base::Future<ErrorMessageOr<void>> OrbitApp::LoadSymbols(
...
+    const std::filesystem::path& module_path) {
+  std::vector<std::string> build_ids =
+      GetTargetProcess()->FindModuleBuildIdsByPath(module_path.string());
+
+  std::vector<const ModuleData*> result;
+  for (const auto& build_id : build_ids) {
+    const ModuleData* module_data =
+        module_manager_->GetModuleByPathAndBuildId(module_path.string(), build_id);
+    if (module_data == nullptr) {
+      ERROR("module \"%s\" was loaded by the process, but is not part of module manager",
```
dpallotti, line 1707:
```suggestion
      ERROR("Module \"%s\" was loaded by the process, but is not part of module manager",
```

dpallotti (approved):

src/ClientData/ProcessData.cpp
```diff
@@ -61,40 +61,34 @@ const std::string& ProcessData::build_id() const {
...
-    }
+    const auto [unused_it, success] = start_address_to_module_in_memory_.try_emplace(
+        module_info.address_start(), module_info.address_start(), module_info.address_end(),
+        module_info.file_path(), module_info.build_id());
+    CHECK(success);
   }
 }
 
-std::optional<ModuleInMemory> ProcessData::FindModuleByPath(const std::string& module_path) const {
+std::vector<std::string> ProcessData::FindModuleBuildIdsByPath(
```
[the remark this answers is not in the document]
dimitry-: done.

src/OrbitGl/App.cpp
dimitry-: done.

src/OrbitGl/App.cpp
dimitry-: done.

PR #2441, opened 2021-06-17
Fix Qt multithreading issue

antonrohr:
Beforehand an QObject::connect overload was used which did not execute
the lambda on the correct (main) thread. This commit fixes that and adds
a CHECK to be sure.

### DIFF
--- a/src/Orbit/main.cpp
+++ b/src/Orbit/main.cpp
@@ -283,27 +283,37 @@ static void TryMoveSavedDataLocationIfNeeded() {
     return;
   }
 
+  std::thread::id main_thread_id = std::this_thread::get_id();
+
   orbit_qt::MoveFilesDialog dialog;
   orbit_qt::MoveFilesProcess process;
 
-  QObject::connect(&process, &orbit_qt::MoveFilesProcess::generalError,
-                   [&dialog](const std::string& error_message) {
+  QObject::connect(&process, &orbit_qt::MoveFilesProcess::generalError, &dialog,
+                   [&dialog, main_thread_id](const std::string& error_message) {
+                     CHECK(main_thread_id == std::this_thread::get_id());
                      dialog.AddText(absl::StrFormat("Error: %s", error_message));
                    });
 
   QObject::connect(
-      &process, &orbit_qt::MoveFilesProcess::moveStarted,
-      [&dialog](const std::filesystem::path& from_dir, const std::filesystem::path& to_dir,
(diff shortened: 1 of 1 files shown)


src/Orbit/main.cpp
```diff
@@ -283,27 +283,37 @@ static void TryMoveSavedDataLocationIfNeeded() {
...
+  std::thread::id main_thread_id = std::this_thread::get_id();
+
   orbit_qt::MoveFilesDialog dialog;
   orbit_qt::MoveFilesProcess process;
 
-  QObject::connect(&process, &orbit_qt::MoveFilesProcess::generalError,
-                   [&dialog](const std::string& error_message) {
+  QObject::connect(&process, &orbit_qt::MoveFilesProcess::generalError, &dialog,
+                   [&dialog, main_thread_id](const std::string& error_message) {
+                     CHECK(main_thread_id == std::this_thread::get_id());
```
beckerhe, line 293: Why this check? The behaviour of `QObject::connect` is guaranteed by Qt. When could this fail?

beckerhe (approved):

ronaldfw:
Is this a fix for b/191258690?

src/Orbit/main.cpp
beckerhe: Ah, I just read the description which mentions the CHECK. I personally find it confusing because I immediately think about when this could fail and I think it can't. But maybe that's just me. Feel free to ignore.

antonrohr: I know about the guarantee. This cannot fail currently.@dimitry- suggested to leave the check in. I guess for clarity while reading the code. I do not feel strongly about this. @dimitry- maybe you have a opinion on this.

antonrohr:
> Is this a fix for b/191258690?

Yes, it is a fix, but we decided to rather revert instead of fix.

src/Orbit/main.cpp
dimitry-: without the pointer to object (dialog in this case) there is no guarantee, which was not obvious to me, I assumed it will take the caller thread, hence I find the check useful but again this is my lack of knowledge of qt things.

beckerhe: I see. That makes sense to me. Thanks for the explanation!

dpallotti:
This was incorporated in https://github.com/google/orbit/pull/2451, so I'm closing this. Thanks Anton for the fix!

PR #2443, opened 2021-06-17
[Rebase on main] Fix file name conflict issue

dimitry-:
And sequence number to file name if file already
exists.

Test: Manual - press x many times see that files with suffixes are
      created
Bug: http://b/191247811


danielfenner (approved):
lgtm

PR #2444, opened 2021-06-17
[Rebase on main] Correctly handle errors when unable to open the file.

dimitry-:
Making file_path_ std::optional lets us to correctly set
file_path in capture data when Orbit was unable to open it.

Bug: http://b/191247811
Test: Trigger the error by clicking stop-start very fast for some
      time. Finish capture after the error and see that there are
      no errors.


danielfenner (approved):
lgtm

PR #2445, opened 2021-06-17
Move gtest's main functions into separate module

beckerhe:
gtest's main functions used to be created by CMake on the fly. This
commit is moving them into its own cmake module and subdirectory and
treats their targets as if they were standard library targets.

It is also changing when QApplication and QCoreApplication instances are
created. By using a test listener we postpone the creation to right
before the test start.

That's important because this way gtest command line options like
`--help` or `--gtest_list_tests` won't need a QApplication instance and
hance they work in headless setups.

I discovered this problem while investigating the flaky Windows build, but
it seems unrelated to that.


src/Test/test_qtcore_main.cpp
```diff
@@ -0,0 +1,43 @@
...
+#include <cstdio>
+#include <optional>
+
+namespace {
+// QCoreApplication objects needs to be on the stack or in static storage. Putting this variable
+// into the QCoreApplicationStarter class leads to a crash during destruction. I'm not sure what is
+// going on.
+std::optional<QCoreApplication> app;
+
+// QCoreApplicationStarter is a gtest event listener which takes are of delayed construction of the
```
danielfenner, line 17: Sentence derailed.

danielfenner (approved):
lgtm

src/Test/test_qtcore_main.cpp
beckerhe: Oh yes. There is a `c` missing which makes it completely not understandable. Thanks. :sweat_smile:

beckerhe: Fixed.

PR #2446, opened 2021-06-17
Remove enable_source_code_view flag

beckerhe:
The flag was defaulted to true more than two releases ago, so let's
finally remove the flag.


karupayun (approved):
LGTM!

PR #2448, opened 2021-06-18
Fix OrbitGgp Client and ClientTest

antonrohr:
Remove an additional second waiting time from RunProcessWithTimeout.
Increase the delay of MockGgpWorking by 20ms. Fix a test string in
ClientTest. (Re)Enable OrbitGgpTests

I let the test in question run extensively before I made this PR to make sure it is not flaky.

### DIFF
--- a/src/OrbitGgp/Client.cpp
+++ b/src/OrbitGgp/Client.cpp
@@ -22,8 +22,6 @@ namespace orbit_ggp {
 
 namespace {
 
-constexpr int kAdditionalSynchronousTimeoutInMs = 20;
-
 void RunProcessWithTimeout(const QString& program, const QStringList& arguments,
                            std::chrono::milliseconds timeout, QObject* parent,
                            const std::function<void(outcome::result<QByteArray>)>& callback) {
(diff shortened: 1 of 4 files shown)


src/OrbitGgp/Client.cpp
```diff
@@ -35,12 +33,11 @@ void RunProcessWithTimeout(const QString& program, const QStringList& arguments,
 
   QObject::connect(timeout_timer, &QTimer::timeout, parent,
                    [process, timeout_timer, timeout, callback]() {
-                     if (process && !process->waitForFinished(kAdditionalSynchronousTimeoutInMs)) {
+                     if (process && !process->waitForFinished(0)) {
```
beckerhe, line 36: What does `0` as the argument mean in that regard? I guess it will basically tell you whether the process is still running or not. In that case I would prefer to use `QProcess::state` instead.

If you meant to wait forever you have to pass in `-1` according to the docs.

beckerhe (approved):
Otherwise LGTM

src/OrbitGgp/Client.cpp
antonrohr: I used QProcessState now

src/OrbitGgp/Client.cpp
```diff
@@ -17,13 +17,12 @@
 #include "OrbitGgp/Error.h"
 #include "OrbitGgp/Instance.h"
 #include "OrbitGgp/SshInfo.h"
+#include "qprocess.h"
```
beckerhe, line 20: Nit

antonrohr: oh thanks!

PR #2449, opened 2021-06-18
Move FunctionsDataView and add tests

beckerhe:
This is adding tests for FunctionsDataView in a similar manner as it has already been done for PresetsDataView.

### DIFF
--- a/src/DataViews/FunctionsDataViewTest.cpp
+++ b/src/DataViews/FunctionsDataViewTest.cpp
@@ -1,0 +1,673 @@
+// Copyright (c) 2021 The Orbit Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#include <absl/flags/flag.h>
+#include <absl/strings/str_split.h>
+#include <absl/types/span.h>
+#include <gmock/gmock-actions.h>
+#include <gmock/gmock-cardinalities.h>
+#include <gmock/gmock-matchers.h>
+#include <gmock/gmock-more-actions.h>
+#include <gtest/gtest.h>
+
+#include "process.pb.h"
+
+#ifdef _WIN32
+#include "oqpi.hpp"
+
+#define OQPI_USE_DEFAULT
+#endif
+
+#include "ClientData/FunctionUtils.h"
+#include "ClientData/ModuleData.h"
+#include "ClientData/ModuleManager.h"
+#include "ClientData/ProcessData.h"
+#include "ClientModel/CaptureData.h"
+#include "DataViews/DataView.h"
+#include "DataViews/FunctionsDataView.h"
+#include "MockAppInterface.h"
+#include "OrbitBase/ReadFileToString.h"
+#include "OrbitBase/TemporaryFile.h"
+#include "OrbitBase/TestUtils.h"
+#include "capture.pb.h"
+#include "capture_data.pb.h"
+
+namespace {
+struct FunctionsDataViewTest : public testing::Test {
+ public:
+  explicit FunctionsDataViewTest() : view_{&app_} {
+#ifdef _WIN32
+    oqpi::default_helpers::start_default_scheduler();
+#endif
+
+    orbit_client_protos::FunctionInfo function0;
+    function0.set_name("foo");
+    function0.set_pretty_name("void foo()");
+    function0.set_module_path("/path/to/module");
+    function0.set_module_build_id("buildid");
+    function0.set_address(12);
+    function0.set_size(16);
+    functions_.emplace_back(std::move(function0));
+
+    orbit_client_protos::FunctionInfo function1;
+    function1.set_name("main");
+    function1.set_pretty_name("main(int, char**)");
+    function1.set_module_path("/path/to/other");
+    function1.set_module_build_id("buildid2");
+    function1.set_address(0x100);
+    function1.set_size(42);
+    functions_.emplace_back(std::move(function1));
+
+    orbit_client_protos::FunctionInfo function2;
+    function2.set_name("_ZeqRK1AS1_");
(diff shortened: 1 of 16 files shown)


src/DataViews/FunctionsDataViewTest.cpp
```diff
@@ -0,0 +1,655 @@
...
+
+  is_function_selected = {false, true, false};
+  is_frame_track_enabled = {false, true, false};
+  EXPECT_THAT(view_.GetContextMenu(0, {0, 1, 2}), can_hook_and_unhook());
+  EXPECT_THAT(view_.GetContextMenu(0, {0, 1, 2}), can_enable_and_disable_frame_tracks());
+
+  is_function_selected = {false, true, true};
+  is_frame_track_enabled = {false, true, false};
+  EXPECT_THAT(view_.GetContextMenu(0, {0, 1, 2}), can_hook_and_unhook());
+  EXPECT_THAT(view_.GetContextMenu(0, {0, 1, 2}), can_enable_and_disable_frame_tracks());
```
antonrohr, line 441: pers pref: Add 2 more blocks
1. 
```
is_function_selected = {true, true, true};  
is_frame_track_enabled = {true, true, false};
```
2. 
```
is_function_selected = {true, true, true};  
is_frame_track_enabled = {true, true, true};
```

src/DataViews/FunctionsDataViewTest.cpp
```diff
@@ -0,0 +1,655 @@
...
+                             ", %s, %d, %s, %#x\n",
+                             functions_[0].pretty_name(), functions_[0].size(),
+                             std::filesystem::path{functions_[0].module_path()}.filename().string(),
+                             functions_[0].address()));
+  }
+
+  // Export to CSV
+  {
+    const auto export_to_csv_idx =
+        std::find(context_menu.begin(), context_menu.end(), "Copy Selection") -
```
antonrohr, line 485: I think this was wrongly copied from the test before. The string should be "Export to CSV". 

Researching why this test succeeds, even with this wrong string, led me to the following: The parameter `menu_index` in `DataView::OnContextMenu()` seems to be not used at all in the code base. Maybe consider removing it.

src/DataViews/FunctionsDataViewTest.cpp
```diff
@@ -0,0 +1,655 @@
...
+        context_menu.begin();
+    ASSERT_LT(export_to_csv_idx, context_menu.size());
+
+    ErrorMessageOr<orbit_base::TemporaryFile> temporary_file_or_error =
+        orbit_base::TemporaryFile::Create();
+    ASSERT_THAT(temporary_file_or_error, orbit_base::HasNoError());
+    const std::filesystem::path temporary_file_path = temporary_file_or_error.value().file_path();
+    temporary_file_or_error.value().CloseAndRemove();
```
antonrohr, line 493: I was confused, why this file is closed and removed before the context menu action can write to it. I think I understood now, that you actually only need a tmp file path and the detour via creating a temporary file gives you a file path in "/tmp" (or windows equivalent). 

I think as the scope of this PR, there should be a comment explaining that. (The even better solution would be having a tested function in OrbitBase that creates a tmp file path. And this function is then used by TemporaryFile)

Also. Consider deleting the file after this test finishes.

antonrohr (approved):
Nice work! Only nits

src/DataViews/FunctionsDataViewTest.cpp
```diff
@@ -0,0 +1,655 @@
...
+#include <absl/flags/flag.h>
+#include <absl/strings/str_split.h>
+#include <absl/types/span.h>
+#include <gmock/gmock-actions.h>
+#include <gmock/gmock-cardinalities.h>
+#include <gmock/gmock-matchers.h>
+#include <gmock/gmock-more-actions.h>
+#include <gtest/gtest.h>
+
+#include "process.pb.h"
```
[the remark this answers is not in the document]
beckerhe: I don't understand. In what way is it tripping clang-format - something on Windows? Works fine for me. The CI seems also to be fine. I also added the include at the exact location where clang-format would place it if it wasn't in an #ifdef.

src/DataViews/FunctionsDataViewTest.cpp
beckerhe: Yeah, I won't change any code from the data views here. Otherwise I won't be finished for Christmas. :innocent: But I will fix the literal. Thanks for spotting this.

src/DataViews/FunctionsDataViewTest.cpp
beckerhe: `TemporaryFile::CloseAndRemove` exactly does what you suggest. It makes `TemporaryFile` only create a temporary file path and it will still delete the file at the end of the test. We also have multiple similar usages in tests in the codebase. This was surprising to me as well, maybe the API can be improved here but that feels out of scope for this PR.

I will add a comment though. That's a good suggestion.

src/DataViews/FunctionsDataViewTest.cpp
beckerhe: Done.

src/DataViews/FunctionsDataViewTest.cpp
beckerhe: Done.

src/DataViews/FunctionsDataViewTest.cpp
beckerhe: Done.

src/DataViews/FunctionsDataViewTest.cpp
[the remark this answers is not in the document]
beckerhe: I will go ahead and merge this. Let's discuss this in chat.

PR #2450, opened 2021-06-18
Remove intersecting modules on AddOrUpdateModuleInfo

dimitry-:
Also fix a bug where module was incorrectly found when looking
for it by the end address. Note that end address could be
another module start address.

Test: run ClientDataTests
Bug: http://b/190805289

### DIFF
--- a/src/ClientData/ProcessDataTest.cpp
+++ b/src/ClientData/ProcessDataTest.cpp
@@ -365,11 +365,8 @@ TEST(ProcessData, FindModuleByAddress) {
   {
     // exactly end address
     const auto result = process.FindModuleByAddress(end_address);
-    ASSERT_FALSE(result.has_error());
-    EXPECT_EQ(result.value().file_path(), module_path);
-    EXPECT_EQ(result.value().start(), start_address);
-    EXPECT_EQ(result.value().end(), end_address);
-    EXPECT_EQ(result.value().build_id(), kBuildId);
+    EXPECT_THAT(result, HasError("Unable to find module for address"));
+    EXPECT_THAT(result, HasError("No module loaded at this address"));
   }
   {
     // after end address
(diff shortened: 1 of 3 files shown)


src/ClientData/include/ClientData/ProcessData.h
```diff
@@ -69,7 +69,13 @@ class ProcessData final {
   [[nodiscard]] const std::string& build_id() const;
 
   void UpdateModuleInfos(absl::Span<const orbit_grpc_protos::ModuleInfo> module_infos);
-  void AddOrUpdateModuleInfo(const orbit_grpc_protos::ModuleInfo& module_infos);
+
+  // This method removes all modules with addresses intersecting with module_info.
+  // Some background on this. The service currently does not send any unmap event
+  // to the client, which means that if the client receives a module with mapping
+  // intersecting with exiting mapping the old module was likely unloaded.
```
dpallotti, line 76: But could it happen that we have absolute addresses in the capture that refer to the unloaded module? Wouldn't it be a problem if we have removed the module from the mapping?

What would happen if we kept both? That then we wouldn't know how to treat absolute addresses that fall in the intersection hence making things even worse?

src/ClientData/ProcessDataTest.cpp
```diff
@@ -483,4 +480,298 @@ TEST(ProcessData, RemapModule) {
...
+    ASSERT_THAT(result, HasNoError());
+    EXPECT_EQ(result.value().file_path(), kNewModulePath);
+    EXPECT_EQ(result.value().build_id(), kNewBuildId);
+    EXPECT_EQ(result.value().start(), 100);
+    EXPECT_EQ(result.value().end(), 300);
+  }
+}
+
+TEST_F(ProcessDataModuleIntersectionTest, FullyInsideAnotherModuleAddressRange) {
+  // address range in fully inside another module address range
```
dpallotti, line 616:
```suggestion
  // Address range is fully inside another module's address range
```

src/ClientData/ProcessDataTest.cpp
```diff
@@ -483,4 +480,298 @@ TEST(ProcessData, RemapModule) {
...
+    EXPECT_EQ(result.value().end(), 190);
+
+    // Original module is gone
+    EXPECT_THAT(process_.FindModuleByAddress(108), HasError("Unable to find module for address"));
+    EXPECT_THAT(process_.FindModuleByAddress(190), HasError("Unable to find module for address"));
+  }
+}
+
+TEST_F(ProcessDataModuleIntersectionTest, OverlapsWithEverything) {
+  // address range in fully inside another module address range
```
dpallotti, line 666: Likely a copy-paste leftover.

src/ClientData/ProcessDataTest.cpp
```diff
@@ -483,4 +480,298 @@ TEST(ProcessData, RemapModule) {
...
+    ASSERT_THAT(result, HasNoError());
+    EXPECT_EQ(result.value().file_path(), kNewModulePath);
+    EXPECT_EQ(result.value().build_id(), kNewBuildId);
+    EXPECT_EQ(result.value().start(), 10);
+    EXPECT_EQ(result.value().end(), 450);
+  }
+}
+
+TEST_F(ProcessDataModuleIntersectionTest, ReplaceFirstModule) {
+  // address range in fully inside another module address range
```
dpallotti, line 683:
```suggestion
  // Address range is fully inside the first module's address range
```

src/ClientData/ProcessDataTest.cpp
```diff
@@ -483,4 +480,298 @@ TEST(ProcessData, RemapModule) {
...
+    ASSERT_THAT(result, HasNoError());
+    EXPECT_EQ(result.value().file_path(), kModulePath3);
+    EXPECT_EQ(result.value().build_id(), kBuildId3);
+    EXPECT_EQ(result.value().start(), kStartAddress3);
+    EXPECT_EQ(result.value().end(), kEndAddress3);
+  }
+}
+
+TEST_F(ProcessDataModuleIntersectionTest, ReplaceLastModule) {
+  // address range in fully inside another module address range
```
dpallotti, line 731: This doesn't apply to this test.

dpallotti (commented):
The code looks good, but I'm wondering whether removing a module in this very rare case would end up being a problem (see the comment in `ProcessData.h`).

src/ClientData/include/ClientData/ProcessData.h
dimitry-: Let me start by saying the proper fix is to have a timeline for module load events but this is out of the scope of this PR.

The most likely use-case is that the  same address is remapped in which case this does not make it worth. 
The second use case is having 2 consecutive modules unmapped and 1 (bigger) module remapped in their place, in which case this PR makes things better since it it removes the second module before mapping the bigger one. Without it looking up addresses after former start of the second module will result in incorrect module/function addresses being used.

The third use case if somehow the mapping ended up intersecting 1 module. This could happen for example if someone unloads a module then calls mmap and then loads different module. In which case with old logic we will end up with the checking the second module in most cases since anything bigger new module address will look only at new module start address. even if it end up not being in the requested address range.

The only use case which makes things a bit worse is when we happen to keep old module around and address lands intersection between start of old module and start of new one. In which case old logic accidentally gets it right.

dpallotti: Thanks. I think it would be useful to have a summary of this explanation in a comment, or at the very least in the commit message.

dpallotti (approved):
Consider moving some of your explanation to where it's easily accessible, e.g., a comment. Looks good otherwise.

src/ClientData/include/ClientData/ProcessData.h
dimitry-: done (moved parts of this into commit message)

PR #2451, opened 2021-06-18
Move captures and presets from `%appdata%\OrbitProfiler` to `Documents\Orbit`

dpallotti:
Reverts the revert of https://github.com/google/orbit/pull/2379, fixes a bug with using https://github.com/google/orbit/pull/2441, uses `SHGetKnownFolderPath` to get the path to Documents, makes some visual improvements and some other minor improvements.


src/OrbitCore/Path.cpp
```diff
@@ -88,6 +97,43 @@ std::filesystem::path CreateOrGetOrbitAppDataDir() {
   return path;
 }
 
+static std::filesystem::path GetDocumentsPath() {
+#ifdef WIN32
+  PWSTR ppszPath;
+  HRESULT result = SHGetKnownFolderPath(FOLDERID_Documents, 0, nullptr, &ppszPath);
```
dimitry-, line 103: SHGetKnownFolderPathA?

src/OrbitCore/Path.cpp
```diff
@@ -88,6 +97,43 @@ std::filesystem::path CreateOrGetOrbitAppDataDir() {
...
+
+  if (result != S_OK) {
+    CoTaskMemFree(ppszPath);
+    LPTSTR lpBuffer = nullptr;
+    FormatMessage(
```
dimitry-, line 108: FormatMessageA?

src/OrbitCore/Path.cpp
dpallotti: I don't think that exists.

src/OrbitCore/Path.cpp
dpallotti: Done.

src/OrbitQt/MoveFilesDialog.ui
```diff
@@ -24,76 +21,59 @@
...
-    <sizepolicy hsizetype="Expanding" vsizetype="Expanding">
-     <horstretch>0</horstretch>
-     <verstretch>0</verstretch>
-    </sizepolicy>
-   </property>
-   <property name="readOnly">
-    <bool>true</bool>
-   </property>
-  </widget>
+  <layout class="QVBoxLayout">
```
dimitry-, line 40: Thanks! I spent non-trivial amount of time trying to apply a layout :)

src/OrbitCore/Path.cpp
```diff
@@ -92,14 +97,36 @@ std::filesystem::path CreateOrGetOrbitAppDataDir() {
...
+        reinterpret_cast<LPSTR>(&error_string), 0, nullptr);
+    std::filesystem::path path = std::filesystem::path(GetEnvVar("USERPROFILE")) / "Documents";
+    ERROR("Retrieving path to Documents (defaulting to \"%s\"): %s", path.string(), error_string);
+    LocalFree(error_string);
+    return path;
+  }
+
+  std::wstring wpath = ppszPath;
+  CoTaskMemFree(ppszPath);
+  std::string path{wpath.begin(), wpath.end()};
```
beckerhe, line 120: This conversion could be problematic if the path contains non-ascii-characters. What about creating a `std::filesystem::path` directory from `wpath`. There is actually a constructor for that purpose.

beckerhe (approved):
Awesome. Thanks a lot. Looks all fine to me.

src/Orbit/main.cpp
```diff
@@ -296,23 +296,37 @@ static void TryMoveSavedDataLocationIfNeeded() {
...
-  QObject::connect(&process, &orbit_qt::MoveFilesProcess::moveDone, &dialog,
+  QObject::connect(&process, &orbit_qt::MoveFilesProcess::moveDirectoryDone, &dialog,
                    [&dialog, main_thread_id]() {
                      CHECK(main_thread_id == std::this_thread::get_id());
-                     dialog.AddText("... done");
+                     dialog.AddText("Done.\n");
+                   });
+
+  QObject::connect(
+      &process, &orbit_qt::MoveFilesProcess::moveFileStarted, &dialog,
```
dimitry-, line 314: I was actually thinking about this a decided against it because this ended up generating too many messages.

src/OrbitQt/MoveFilesDialog.cpp
```diff
@@ -33,4 +33,20 @@ void MoveFilesDialog::EnableCloseButton() {
   ui_->closeButton->setEnabled(true);
 }
 
+void MoveFilesDialog::closeEvent(QCloseEvent* event) {
+  if (ui_->closeButton->isEnabled()) {
+    event->accept();
+  } else {
+    event->ignore();
+  }
+}
```
dimitry-, line 42: does it prevent close on escape?

src/OrbitQt/MoveFilesDialog.cpp
```diff
@@ -33,4 +33,20 @@ void MoveFilesDialog::EnableCloseButton() {
...
+
+void MoveFilesDialog::keyPressEvent(QKeyEvent* event) {
+  if (event->key() != Qt::Key_Escape || ui_->closeButton->isEnabled()) {
```
dimitry-, line 45: ok, I see... but in this case does closeEvent do anything?

dimitry- (approved):
otherwise lgtm

dpallotti:
@beckerhe I will also need you to take a look at http://go/stadia-orbit-migration-to-documents first.

src/Orbit/main.cpp
dimitry-: but this is fine with me too.

src/OrbitQt/MoveFilesDialog.cpp
dpallotti: `closeEvent` is called when the window is closed with the X or the equivalent in some menu, that's what it prevents. But not on ESC.

src/OrbitQt/MoveFilesDialog.cpp
dpallotti: Nope, that's the reason for `keyPressEvent` below (as you have seen), but we still need this.

src/Orbit/main.cpp
dpallotti: When the process takes long (e.g., when migrating across partitions) we need to give the user more information on what's happening and a sense of how long this is taking. I think that giving information file by file goes a long way. Yes, it's many messages, but oh well.

src/OrbitQt/MoveFilesDialog.ui
dpallotti: It's indeed not obvious how to do it in the designer (right click the the container > "Lay out" > "Lay Out Vertically", and NOT dragging the desired layout from the widgets on the left).

dpallotti:
As discussed offline, I'll add the possibility of interrupting the migration process. You will still need to wait for the current file to be moved and the migration will continue next time you open Orbit.

src/OrbitCore/Path.cpp
dpallotti: Oh good point, done. (`LOG` still needs `string` it seems)

dpallotti:
I added two commits:
- The second-last one allows suspending the move for now until Orbit is started the next time, as discussed.
- The last one just move this code out of OrbitQt into a separate module.
So I'll need you to have a look.

src/MoveFilesToDocuments/CMakeLists.txt
```diff
@@ -0,0 +1,38 @@
+# Copyright (c) 2021 The Orbit Authors. All rights reserved.
+# Use of this source code is governed by a BSD-style license that can be
+# found in the LICENSE file.
+
+project(MoveFilesToDocuments CXX)
```
dpallotti, line 5: Let me know if you think I should use a different name.

beckerhe: SGTM.

src/Orbit/main.cpp
```diff
@@ -327,9 +327,20 @@ static void TryMoveSavedDataLocationIfNeeded() {
...
+  QObject::connect(&process, &orbit_qt::MoveFilesProcess::processInterrupted, &dialog,
+                   [&dialog, main_thread_id]() {
+                     CHECK(main_thread_id == std::this_thread::get_id());
+                     dialog.AddText("Interrupted.");
+                     dialog.OnMoveInterrupted();
+                   });
+
+  // Intentionally use the version of `QObject::connect` with three arguments.
+  QObject::connect(&dialog, &orbit_qt::MoveFilesDialog::interruptionRequested,
+                   [&process]() { process.RequestInterruption(); });
```
beckerhe: I don't fully understand. Why not `QObject::connect(&dialog, &orbit_qt::MoveFilesDialog::interruptionRequested, &process, &orbit_qt::MoveFilesProcess::RequestInterruption);`?

src/OrbitQt/MoveFilesProcess.cpp
```diff
@@ -24,13 +24,16 @@ void MoveFilesProcess::Start() {
   QMetaObject::invokeMethod(this, [this]() { Run(); });
```
beckerhe: Nit: For the current use-case this is not needed, but from a design perspective the user of this class could call `Start` multiple times with the same object. To make this work `interruption_requested_` needs to be reset to false here.

src/MoveFilesToDocuments/CMakeLists.txt
```diff
@@ -0,0 +1,38 @@
...
+
+add_library(MoveFilesToDocuments OBJECT)
+
+target_compile_options(MoveFilesToDocuments PRIVATE ${STRICT_COMPILE_FLAGS})
+
+target_include_directories(MoveFilesToDocuments PUBLIC
+        ${CMAKE_CURRENT_SOURCE_DIR}/include)
+
+target_include_directories(MoveFilesToDocuments PRIVATE
+        ${CMAKE_CURRENT_SOURCE_DIR})
```
beckerhe, line 15: Nit: This is not needed.

src/Orbit/main.cpp
```diff
@@ -256,95 +255,6 @@ static void LogAndMaybeWarnAboutClockResolution() {
   }
 }
 
-static bool IsDirectoryEmpty(const std::filesystem::path& directory) {
```
beckerhe: Thanks a lot for moving this!

beckerhe (approved):
Otherwise LGTM

beckerhe:
@dpallotti  BTW, when doing fixups for this PR with a lot of commits you might appreciate https://symflower.com/en/company/blog/2021/git-autofixup/

src/OrbitQt/MoveFilesDialog.ui
```diff
@@ -78,7 +78,7 @@
  <resources/>
  <connections>
   <connection>
-   <sender>closeButton</sender>
+   <sender>button</sender>
    <signal>clicked()</signal>
    <receiver>MoveFilesDialog</receiver>
    <slot>accept()</slot>
```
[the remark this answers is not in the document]
dpallotti: Aaaaaah this is where that connection happens, I guess this was originally introduced by Dimitry and I missed it. I thought it was done automatically by the dialog and that's why I was overriding `accept` to undo it. Thanks, I'll definitely remove it!

src/Orbit/main.cpp
dpallotti: Because then it happens on the event loop (or whatever) of `MoveFilesProcess`, which is already executing the move, while I need to set `interruption_requested_` from another thread. Is there a better way?

dpallotti: It might have something to do with `moveToThread(&background_thread_);`. That was from Dimitry's initial change so I'm not sure if that's needed.

src/OrbitQt/MoveFilesDialog.ui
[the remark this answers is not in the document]
dpallotti: Done.

src/MoveFilesToDocuments/CMakeLists.txt
dpallotti: Done.

src/OrbitQt/MoveFilesProcess.cpp
dpallotti: Oh good point, done.

src/Orbit/main.cpp
beckerhe: Ahh, now I see what is happening here. Thanks for the explanation. I think the code is fine.

The following would be more Qt-ish but also more work than just a quick fixup. I'm not even sure if it improves the situation:

1. Copying all the files happens in a single job on the event loop inside of the QThread in `MoveFilesProcess`.
2. QThread already has a mechanism to stop what it's currently doing. `QThread::quit`. It will quit the current event loop but the event loop will still finish executing the currently running task.
3. To make use of that you have to split `MoveFilesProcess::Run` into smaller tasks which all need to be queued on the event loop (with `QMetaObject::invokeMethod(..., Qt::QueuedConnection)` for example).
4. Then you could call `QThread::quit` from wherever you want since it's a thread-safe function. No atomic required etc.

dpallotti: Thanks for the detailed explanation. I'm certainly leaning towards keeping it as it is as it seems to work and the new approach would need quite some more testing. But you understand what's going on here much better than me so you tell me whether you think it's worth investing the time.

beckerhe: I think you should leave it as-is. It's not worth it, at least not at the moment.

dpallotti: In the meantime I clarified the comment.

dpallotti: Alright, thanks!

dpallotti:
> BTW, when doing fixups for this PR with a lot of commits you might appreciate https://symflower.com/en/company/blog/2021/git-autofixup/

Oh thanks, that indeed looks like something I should use!

PR #2452, opened 2021-06-21
Remove autosave flag

dimitry-:
Bug: http://b/187388305
Test: builds

### DIFF
--- a/src/OrbitGl/App.cpp
+++ b/src/OrbitGl/App.cpp
@@ -98,7 +98,6 @@
 ABSL_DECLARE_FLAG(bool, devmode);
 ABSL_DECLARE_FLAG(bool, local);
 ABSL_DECLARE_FLAG(bool, enable_tracepoint_feature);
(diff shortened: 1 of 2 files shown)


florian-kuebler (approved):

florian-kuebler:
LGTM

dpallotti (approved):
I'm fine with removing it but have you consider keeping it for internal use? Would we have downsides?

dimitry-:
> I'm fine with removing it but have you consider keeping it for internal use? Would we have downsides?

I do not really see any use-case for disabling this flag... do you have something specific in mind?

dpallotti:
> > I'm fine with removing it but have you consider keeping it for internal use? Would we have downsides?
> 
> I do not really see any use-case for disabling this flag... do you have something specific in mind?

I was just thinking that when developing and taking many captures it could be useful to just turn autosave off to prevent accumulating captures. But on the other hand maybe we shouldn't get used to a workflow that we don't support outside of development.

PR #2453, opened 2021-06-21
Cleanup conan profiles

beckerhe:
This PR is deduplicating all the conan profiles for both platforms - Windows and Linux.

Conan profiles support including other conan profiles and this way inherit all the settings from the included profile.

We use this feature to establish a hierarchy, for example: `linux_common` -> `clang_common` -> `clang9_common` -> `clang9_relwithdebinfo`.

@danielfenner @dimitry- I'm adding you both as reviewers since you both recently touched the CXXFLAGS and CFLAGS. I would be great if you could focus in the review especially on the flags. I hopefully didn't miss any and hopefully didn't make any copy-and-paste error. This could go rather undetected because it often still compiles even if flags we want are missing.


dimitry- (approved):

danielfenner (approved):
lgtm! thanks!

PR #2454, opened 2021-06-21
Add CI specific conan profiles

beckerhe:
This is adding a generic mechanism to have CI-specific conan configurations which differ from the settings we use for developer builds.

Of course in general we want to keep CI-builds as close to dev-builds as possible. But there are two use-cases which make this viable:

- We use Ninja for Windows builds on the CI because it's faster than MSBuild. We can't move (yet) to Ninja for dev-builds because that breaks the Visual Studio development workflow.
- We need to deploy the MESA llvmpipe OpenGL software rendering library with Orbit releases. This is done by conan - but only in CI builds.

Test: Checked CI logfile. The new config is found and installed and the build is done with Ninja, etc.

### DIFF
--- a/third_party/conan/configs/install.ps1
+++ b/third_party/conan/configs/install.ps1
@@ -40,6 +40,12 @@ if (Test-Path Env:ORBIT_OVERRIDE_ARTIFACTORY_URL) {
     if ($process.ExitCode -ne 0) { Throw "Error while adding conan remote." }
 
     conan_disable_public_remotes
+
+    if (Test-Path -Path "$PSScriptRoot\windows_ci") {
+      Write-Host "Found CI-specific Conan config. Installing that as well..."
+      $process = Start-Process $conan.Path -Wait -NoNewWindow -ErrorAction Stop -PassThru -ArgumentList "config", "install", "$PSScriptRoot\windows_ci"
+      if ($process.ExitCode -ne 0) { Throw "Error while installing conan config." }
+    }
   } catch {
     try {
       $response = Invoke-WebRequest -URI http://orbit-artifactory/ -ErrorAction Ignore -MaximumRedirection 0 -UseBasicParsing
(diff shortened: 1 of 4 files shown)


third_party/conan/configs/install.ps1
```diff
@@ -40,6 +40,12 @@ if (Test-Path Env:ORBIT_OVERRIDE_ARTIFACTORY_URL) {
     if ($process.ExitCode -ne 0) { Throw "Error while adding conan remote." }
 
     conan_disable_public_remotes
+
+    if (Test-Path -Path "$PSScriptRoot\windows_ci") {
+      Write-Host "Found CI-specific Conan config. Installing that as well..."
+      $process = Start-Process $conan.Path -Wait -NoNewWindow -ErrorAction Stop -PassThru -ArgumentList "config","install","$PSScriptRoot\windows_ci"
```
florian-kuebler, line 46:
```suggestion
      $process = Start-Process $conan.Path -Wait -NoNewWindow -ErrorAction Stop -PassThru -ArgumentList "config", "install", "$PSScriptRoot\windows_ci"
```

If that is possible (not sure if the syntax prevents you from adding the space).

florian-kuebler (approved):
Thanks! I assume that the ci related profiles are already available on the ci, so that the `if` will be taken?

If this is the case, LGTM!

beckerhe:
> Thanks! I assume that the ci related profiles are already available on the ci, so that the `if` will be taken?
> 
> If this is the case, LGTM!

Yes, this actually happens as part of this PR. I move one profile over into `windows_ci`.

third_party/conan/configs/install.ps1
beckerhe: I think it's possible. Will do that.

beckerhe: Done.

florian-kuebler:
> > Thanks! I assume that the ci related profiles are already available on the ci, so that the `if` will be taken?
> > If this is the case, LGTM!
> 
> Yes, this actually happens as part of this PR. I move one profile over into `windows_ci`.

Ah, I see in the path of the last file. Missed that. Thanks! LGTM!

PR #2455, opened 2021-06-21
Simplify TriangleToggle State and Usage.

florian-kuebler:
Before this change, the track's triangle toggle had a tri-state,
expanded, collapsed, inactive (meanning not-collapsible). This
required the maintainance of the "initial state", but most
importantly, IsExpanded() was not !IsCollapsed(), which is
error-prone.

Now, similar to the track itself, the toggle can be collapsible
(or not) and independent of this, the toggle has is collapsed
(or not). The default is not collapsed, and (but) collapsible.

Test: Take a capture with all kinds of tracks. Collapse/uncollapse.
Bug: http://b/185465747

### DIFF
--- a/src/OrbitGl/AccessibleTriangleToggle.cpp
+++ b/src/OrbitGl/AccessibleTriangleToggle.cpp
@@ -15,7 +15,7 @@ AccessibleTriangleToggle::AccessibleTriangleToggle(TriangleToggle* triangle_togg
       parent_{parent} {}
 
 orbit_accessibility::AccessibilityState AccessibleTriangleToggle::AccessibleState() const {
-  if (triangle_toggle_->IsInactive()) {
+  if (triangle_toggle_->IsCollapsible()) {
     return orbit_accessibility::AccessibilityState::Disabled;
   }
   return orbit_accessibility::AccessibilityState::Normal;
--- a/src/OrbitGl/AccessibleTriangleToggle.cpp
+++ b/src/OrbitGl/AccessibleTriangleToggle.cpp
@@ -16,9 +16,9 @@ AccessibleTriangleToggle::AccessibleTriangleToggle(TriangleToggle* triangle_togg
 
 orbit_accessibility::AccessibilityState AccessibleTriangleToggle::AccessibleState() const {
   if (triangle_toggle_->IsCollapsible()) {
-    return orbit_accessibility::AccessibilityState::Disabled;
+    return orbit_accessibility::AccessibilityState::Normal;
   }
-  return orbit_accessibility::AccessibilityState::Normal;
+  return orbit_accessibility::AccessibilityState::Disabled;
 }
 
 const orbit_accessibility::AccessibleInterface* AccessibleTriangleToggle::AccessibleParent() const {
--- a/src/OrbitGl/GpuTrack.cpp
+++ b/src/OrbitGl/GpuTrack.cpp
@@ -70,8 +70,9 @@ GpuTrack::GpuTrack(CaptureViewElement* parent, TimeGraph* time_graph, orbit_gl::
   submission_track_->SetName(absl::StrFormat("%s_submissions", timeline));
   marker_track_->SetName(absl::StrFormat("%s_marker", timeline));
 
-  collapse_toggle_->SetState(TriangleToggle::State::kCollapsed,
-                             TriangleToggle::InitialStateUpdate::kReplaceInitialState);
+  // Gpu are expanded by default. Their subtracks are expanded by default, but are however not shown
+  // while the Gpu track is collapsed.
+  collapse_toggle_->SetCollapsed(true);
 }
 
 void GpuTrack::OnTimer(const TimerInfo& timer_info) {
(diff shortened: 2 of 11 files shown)


src/OrbitGl/GpuTrack.cpp
```diff
@@ -70,8 +70,9 @@ GpuTrack::GpuTrack(CaptureViewElement* parent, TimeGraph* time_graph, orbit_gl::
   submission_track_->SetName(absl::StrFormat("%s_submissions", timeline));
   marker_track_->SetName(absl::StrFormat("%s_marker", timeline));
 
-  collapse_toggle_->SetState(TriangleToggle::State::kCollapsed,
-                             TriangleToggle::InitialStateUpdate::kReplaceInitialState);
+  // Gpu are expanded by default. Their subtracks are expanded by default, but are however not shown
```
karupayun, line 73: // Gpu Tracks are collapsed by default

src/OrbitGl/TimerTrack.cpp
```diff
@@ -337,9 +337,7 @@ void TimerTrack::OnTimer(const TimerInfo& timer_info) {
 }
 
 float TimerTrack::GetHeight() const {
-  bool is_collapsed = collapse_toggle_->IsCollapsed();
-  uint32_t collapsed_depth = (GetNumTimers() == 0) ? 0 : 1;
```
karupayun, line 341: I will still have this variable ("collapsed_depth" with this value: std::min<uint32_t>(1, GetDepth())), for legibility

src/OrbitGl/Track.cpp
```diff
@@ -161,11 +161,7 @@ void Track::Draw(Batcher& batcher, TextRenderer& text_renderer, uint64_t current
   }
 
   // Collapse toggle state management.
-  if (!this->IsCollapsible()) {
-    collapse_toggle_->SetState(TriangleToggle::State::kInactive);
-  } else if (collapse_toggle_->IsInactive()) {
-    collapse_toggle_->ResetToInitialState();
-  }
+  collapse_toggle_->SetIsCollapsible(this->IsCollapsible());
```
karupayun, line 164: Why we need this line?

src/OrbitGl/TriangleToggle.h
```diff
@@ -17,13 +17,10 @@ class Track;
...
  public:
-  enum class State { kInactive, kExpanded, kCollapsed };
-  enum class InitialStateUpdate { kKeepInitialState, kReplaceInitialState };
-
-  using StateChangeHandler = std::function<void(TriangleToggle::State)>;
-  explicit TriangleToggle(State initial_state, StateChangeHandler handler, TimeGraph* time_graph,
-                          orbit_gl::Viewport* viewport, TimeGraphLayout* layout, Track* track,
-                          float size);
+  using StateChangeHandler = std::function<void(bool)>;
+  explicit TriangleToggle(bool is_collapsed, bool is_collapsible, StateChangeHandler handler,
```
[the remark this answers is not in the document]
karupayun: Other possibility could be have both at the end and is_collapsed set to false by default. I would like to not have this value for every track, since that almost every track are expanded by default.

src/OrbitGl/Track.cpp
florian-kuebler: That is exactly the point when we tell the triangle to be collapsible or not. Unfortunately, we do not have an event system for tracks yet to get notified when the track itself becomes collapsible or not (which is determined differently for the subtracks, e.g. the timer track checks if there are timers). Currently, we basically pull that state every draw call. In the future we should probably use some event system (i.e. at the point when the state changes, notify the toggle).

But that is out of scope for this PR.

src/OrbitGl/AccessibleTriangleToggle.cpp
```diff
@@ -15,7 +15,7 @@ AccessibleTriangleToggle::AccessibleTriangleToggle(TriangleToggle* triangle_togg
       parent_{parent} {}
 
 orbit_accessibility::AccessibilityState AccessibleTriangleToggle::AccessibleState() const {
-  if (triangle_toggle_->IsInactive()) {
+  if (triangle_toggle_->IsCollapsible()) {
```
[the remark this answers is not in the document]
florian-kuebler: Thanks. I switch the Normal/Disabled below, to make it more readable.

src/OrbitGl/GpuTrack.cpp
florian-kuebler: Thanks. Done.

src/OrbitGl/TimerTrack.cpp
florian-kuebler: Done.

src/OrbitGl/TriangleToggle.h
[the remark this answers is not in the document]
florian-kuebler: I removed it from the constructor and defaulted it with collapsible and not collapsed.

karupayun (approved):
LGTM! Thanks for addressing all my comments :)

PR #2456, opened 2021-06-22
Add separate CMake conan profile

beckerhe:
Whenever we want to change our CMake version we can do that now in one
place (each for Windows and Linux).


danielfenner (approved):
lgtm!

PR #2457, opened 2021-06-22
Unify our Dockerfiles

beckerhe:
This PR cleans up our Docker images.

- All the Linux images will be based on Ubuntu 20.04 LTS.
- It also adds hash based versioning for all Docker images which allows updating and testing containers in a PR without breaking the CI. It also makes updating Docker containers much less error-prone.
- The Windows image is not updated but it is also locked by hash
- Furthermore this PR removes a lot of old crap - like old Kokoro build configs and old Docker files.

Anton, Florian, I will ask you for review since you both made changes to the Docker images before.


third_party/conan/docker/Dockerfile.clang7_opengl_qt
```diff
@@ -1,102 +1,33 @@
...
-    # remove all __pycache__ directories created by pyenv
-    && find /opt/pyenv -iname __pycache__ -print0 | xargs -0 rm -rf \
-    && update-alternatives --install /usr/bin/python python /opt/pyenv/shims/python 100 \
-    && update-alternatives --install /usr/bin/python3 python3 /opt/pyenv/shims/python3 100 \
-    && update-alternatives --install /usr/bin/pip pip /opt/pyenv/shims/pip 100 \
-    && update-alternatives --install /usr/bin/pip3 pip3 /opt/pyenv/shims/pip3 100
+RUN useradd -Ums /bin/bash conan && \
+    apt-get -qq update && \
+    DEBIAN_FRONTEND=noninteractive \
+        apt-get install -y --no-install-recommends \
```
florian-kuebler, line 10: Super Nit: "--yes" is more readable for readers not too familiar.

If you decide to do that change, also consider the other docker files.

florian-kuebler (approved):
That is great. Thanks for that!

antonrohr (approved):
lgtm

third_party/conan/docker/Dockerfile.clang7_opengl_qt
beckerhe: Yes, I agree. But this time I will refrain from doing it because it's a lot of work since I have to rebuild all docker images afterwards and I feel this is not justified.

Thanks anyway, appreciated!

PR #2458, opened 2021-06-22
2 Ui tests

antonrohr:
This adds a pretty exhaustive test for LoadCaptureWidget and a smoke test for ProfilingTargetDialog. 

I am planning on adding more tests for ProfilingTargetDialog, but the class needs to be slightly refactored before that is possible. This will be an upcoming PR

### DIFF
--- a/src/CaptureFileInfo/LoadCaptureWidgetTest.cpp
+++ b/src/CaptureFileInfo/LoadCaptureWidgetTest.cpp
@@ -1,0 +1,114 @@
+// Copyright (c) 2021 The Orbit Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#include <gtest/gtest.h>
+
+#include <QApplication>
+#include <QPushButton>
+#include <QRadioButton>
+#include <QTableView>
+#include <QTest>
+#include <filesystem>
+
+#include "CaptureFileInfo/LoadCaptureWidget.h"
+#include "CaptureFileInfo/Manager.h"
+#include "OrbitBase/ExecutablePath.h"
+
+namespace orbit_capture_file_info {
+
+constexpr const char* kOrgName = "The Orbit Authors";
+
+TEST(LoadCaptureWidget, IsActiveSetActive) {
+  QCoreApplication::setOrganizationName(kOrgName);
+  QCoreApplication::setApplicationName("LoadCaptureWidget.IsActiveSetActive");
+
+  LoadCaptureWidget widget{};
+
+  auto* radio_button = widget.findChild<QRadioButton*>("radioButton");
+  ASSERT_TRUE(radio_button != nullptr);
+  auto* select_file_button = widget.findChild<QPushButton*>("selectFileButton");
+  ASSERT_TRUE(select_file_button != nullptr);
+  auto* table_view = widget.findChild<QTableView*>("tableView");
+  ASSERT_TRUE(table_view != nullptr);
+
+  // The widget needs to be shown so that the detaching of the radio button takes place. Otherwise
+  // the radioButton is not detached from the contentFrame and will be disabled when the widget is
+  // set inactive.
+  widget.show();
+  QApplication::processEvents();
+
+  // default is active
+  EXPECT_TRUE(widget.IsActive());
+  EXPECT_TRUE(select_file_button->isEnabled());
+  EXPECT_TRUE(table_view->isEnabled());
+  // radio button is always enabled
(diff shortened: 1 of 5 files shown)


src/CaptureFileInfo/LoadCaptureWidgetTest.cpp
```diff
@@ -0,0 +1,115 @@
+// Copyright (c) 2021 The Orbit Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#include <gtest/gtest.h>
+#include <qapplication.h>
```
beckerhe, line 6: Nit

src/CaptureFileInfo/LoadCaptureWidgetTest.cpp
```diff
@@ -0,0 +1,115 @@
...
+  // radio button is always enabled
+  EXPECT_TRUE(radio_button->isEnabled());
+
+  // set active to false and then activate by clicking on the radiobutton
+  widget.SetActive(false);
+  bool signal_received = false;
+  QObject::connect(&widget, &LoadCaptureWidget::Activated,
+                   [&signal_received]() { signal_received = true; });
+  QTest::mouseClick(radio_button, Qt::MouseButton::LeftButton, Qt::KeyboardModifier::NoModifier);
+  EXPECT_TRUE(signal_received);
```
beckerhe, line 73: No need to change this, but I wanted to mention that there is QSignalSpy in QTest (https://doc.qt.io/qt-5/qsignalspy.html) to check whether a signal was emitted or not. Maybe this comes handy in the future.

src/OrbitQt/ProfilingTargetDialogTest.cpp
```diff
@@ -0,0 +1,39 @@
...
+
+  SshConnectionArtifacts ssh_artifacts{&ssh_context.value(), ServiceDeployManager::GrpcPort{0},
+                                       &config};
+
+  orbit_metrics_uploader::MetricsUploaderStub uploader{};
+
+  ProfilingTargetDialog dialog{&ssh_artifacts, std::nullopt, &uploader};
+
+  QApplication::processEvents();
+  EXPECT_TRUE(dialog.isEnabled());
```
beckerhe, line 36: What about at least checking the current state of the state machine? This probably requires setting object names for the QStates though.

beckerhe (approved):

src/CaptureFileInfo/LoadCaptureWidgetTest.cpp
```diff
@@ -0,0 +1,115 @@
...
+  LoadCaptureWidget widget{};
+
+  auto* radio_button = widget.findChild<QRadioButton*>("radioButton");
+  ASSERT_TRUE(radio_button != nullptr);
+  auto* select_file_button = widget.findChild<QPushButton*>("selectFileButton");
+  ASSERT_TRUE(select_file_button != nullptr);
+  auto* table_view = widget.findChild<QTableView*>("tableView");
+  ASSERT_TRUE(table_view != nullptr);
+
+  // The widget needs to be show so the detaching of the radio button takes place. Otherwise the
```
dpallotti, line 36:
```suggestion
  // The widget needs to be shown so that the detaching of the radio button takes place. Otherwise the
```

src/OrbitQt/ProfilingTargetDialogTest.cpp
```diff
@@ -0,0 +1,39 @@
...
+#include "MetricsUploader/MetricsUploaderStub.h"
+#include "OrbitSsh/Context.h"
+#include "ProfilingTargetDialog.h"
+
+namespace orbit_qt {
+
+// TODO(b/189838600) Add tests that check correctness.
+
+TEST(ProfilingTargetDialog, SmokeTest) {
+  // This test does nothing else that creating a ProfilingTargetDialog.
```
dpallotti, line 21:
```suggestion
  // This test does nothing else other than creating a ProfilingTargetDialog.
```

dpallotti (approved):

src/CaptureFileInfo/LoadCaptureWidgetTest.cpp
antonrohr: Thanks!

src/OrbitQt/ProfilingTargetDialogTest.cpp
antonrohr: I did a call to Exec and reject() via invoke method

PR #2460, opened 2021-06-22
 TargetLabel's fileLabel allow opening file's containing path

dpallotti:
With double click and context menu action.

Bug: http://b/191121155#comment2

### DIFF
--- a/src/OrbitQt/DoubleClickableLabel.h
+++ b/src/OrbitQt/DoubleClickableLabel.h
@@ -1,0 +1,32 @@
+// Copyright (c) 2021 The Orbit Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#ifndef ORBIT_QT_DOUBLE_CLICKABLE_LABEL_H_
+#define ORBIT_QT_DOUBLE_CLICKABLE_LABEL_H_
+
+#include <QLabel>
+#include <QMouseEvent>
+#include <QWidget>
+
+namespace orbit_qt {
+
+class DoubleClickableLabel : public QLabel {
+  Q_OBJECT
+
+ public:
+  using QLabel::QLabel;
+
+ signals:
+  void DoubleClicked();
+
+ protected:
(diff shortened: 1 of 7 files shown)


src/OrbitQt/DoubleClickableLabel.h
```diff
@@ -0,0 +1,32 @@
...
+#include <QMouseEvent>
+#include <QWidget>
+
+namespace orbit_qt {
+
+class DoubleClickableLabel : public QLabel {
+  Q_OBJECT
+
+ public:
+  explicit DoubleClickableLabel(QWidget* parent = nullptr) : QLabel{parent} {}
```
beckerhe, line 18: Nit: `using QLabel::QLabel;`

src/OrbitQt/DoubleClickableLabelTest.cpp
```diff
@@ -0,0 +1,31 @@
...
+
+#include "DoubleClickableLabel.h"
+
+namespace orbit_qt {
+
+TEST(DoubleClickableLabel, DoubleClickedSignalledOnDoubleClick) {
+  DoubleClickableLabel label;
+  bool double_clicked_signalled = false;
+  QObject::connect(&label, &DoubleClickableLabel::DoubleClicked,
+                   [&double_clicked_signalled] { double_clicked_signalled = true; });
```
beckerhe, line 17: Nit: For testing signals there is `QSignalSpy` (https://doc.qt.io/qt-5/qsignalspy.html). But the current solution is also fine with me.

beckerhe (approved):
Very nice. Only super minor things. Feel free to ignore all of them.

antonrohr (approved):
Very nice, thank you!

src/OrbitQt/DoubleClickableLabel.h
dpallotti: It even crossed my mind. Done.

src/OrbitQt/DoubleClickableLabelTest.cpp
dpallotti: It seems to work well for this case, done.

PR #2461, opened 2021-06-22
Collect RssAnon instead of Rss for the process

vickyliu-go4it:
With this change, we collect the size of resident anonymous memory `RssAnon`
rather than the resident set size `Rss` for the process memory usage.

`Rss` = `RssAnon` + `RssFile` + `RssShmem` and `RssFile` counts the shared files.
However the cgroup memory usage counts a mapped file only if the memory 
cgroup is the owner. Using `Rss` as the process memory usage will result in the case
that process memory usage is larger than the cgroup memory usage.

More detailed discussion can be found in http://b/191758791.

Bug: http://b/191758791
Test: Covered by unit test.

### DIFF
--- a/src/GrpcProtos/capture.proto
+++ b/src/GrpcProtos/capture.proto
@@ -347,13 +347,16 @@ message SystemMemoryUsage {
 }
 
 message ProcessMemoryUsage {
+  reserved 3;
   int32 pid = 1;
   uint64 timestamp_ns = 2;
   // These fields are retrieved from /proc/<pid>/stat. See
   // https://man7.org/linux/man-pages/man5/proc.5.html for the documentation.
-  int64 rss_pages = 3;
   int64 minflt = 4;
   int64 majflt = 5;
+  // This field is retrieved from /proc/<pid>/status. See
+  // https://man7.org/linux/man-pages/man5/proc.5.html for the documentation.
+  int64 rss_anon_kb = 6;
 }
 
 message CGroupMemoryUsage {
--- a/src/MemoryTracing/MemoryTracingUtils.cpp
+++ b/src/MemoryTracing/MemoryTracingUtils.cpp
@@ -150,7 +150,7 @@ ErrorMessageOr<SystemMemoryUsage> GetSystemMemoryUsage() noexcept {
 
 ProcessMemoryUsage CreateAndInitializeProcessMemoryUsage() {
   ProcessMemoryUsage process_memory_usage;
-  process_memory_usage.set_rss_pages(kMissingInfo);
+  process_memory_usage.set_rss_anon_kb(kMissingInfo);
   process_memory_usage.set_minflt(kMissingInfo);
   process_memory_usage.set_majflt(kMissingInfo);
   return process_memory_usage;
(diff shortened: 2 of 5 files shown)


src/MemoryTracing/MemoryTracingUtils.cpp
```diff
@@ -197,24 +190,62 @@ ErrorMessageOr<void> UpdateProcessMemoryUsageFromProcessStat(
   return outcome::success();
 }
 
+ErrorMessageOr<void> UpdateProcessMemoryUsageFromProcessStatus(
+    std::string_view status_content, ProcessMemoryUsage* process_memory_usage) {
```
beckerhe, line 194: As far as I can see this functions only calls `set_rss_anon_kb` from `ProcessMemoryUsage`. Have you considered returning thos one value instead of using an out-parameter? I mean like that: `ErrorMessageOr<int64_t> ExtractRssAnonFromProcessStatus(std::string_view status_content);`?

beckerhe (approved):
Very nice. Otherwise LGTM

src/GrpcProtos/capture.proto
```diff
@@ -349,9 +349,11 @@ message SystemMemoryUsage {
 message ProcessMemoryUsage {
   int32 pid = 1;
   uint64 timestamp_ns = 2;
+  // This field is retrieved from /proc/<pid>/status. See
+  // https://man7.org/linux/man-pages/man5/proc.5.html for the documentation.
+  int64 rss_anon_kb = 3;
   // These fields are retrieved from /proc/<pid>/stat. See
   // https://man7.org/linux/man-pages/man5/proc.5.html for the documentation.
-  int64 rss_pages = 3;
```
florian-kuebler, line 354: That field was already used (in a released version), right? 
For backwards-compatibility, we should mark the field number 3 as `reserved` ([see here](https://developers.google.com/protocol-buffers/docs/proto#reserved)).

For the `rss_pages` you should then use a new number.

src/MemoryTracing/MemoryTracingUtilsTest.cpp
```diff
@@ -133,28 +95,43 @@ SwapCached:      0 kB)",
   const std::string kEmptyMeminfo = "";
 
   {
+    const SystemMemoryUsage kExpectedSystemMemoryUsage = [&] {
```
florian-kuebler, line 98: I think for the test here, that is fine, but in general I would try to avoid capturing all locals (neither per reference nor per value) and explicitly list the used ones.
But do you need any capture here at all, that static values should be available anyways?

florian-kuebler (approved):
Otherwise, LGTM!

src/GrpcProtos/capture.proto
beckerhe: The new field also has the id. So it's a rename only. No breaking change, right?

florian-kuebler: Good point. I am not 100% if it is fine in that case, but I guess you are right.

beckerhe: I guess the more important question is: What happens when a user opens a 1.65 capture with 1.66. We will read the capture and interpet the RSS value as AnonRSS. Is that a problem? Is it better to make a breaking change? Is it worth to invest more and make it backwards compatible while not showing wrong things?

src/MemoryTracing/MemoryTracingUtils.cpp
vickyliu-go4it: Changed accordingly. Thanks for the suggestion!

src/GrpcProtos/capture.proto
vickyliu-go4it: Thanks for pointing it out! I didn't think about this 🙈. As we discussed offline, I changed to make the field number 3 reserved.

src/MemoryTracing/MemoryTracingUtilsTest.cpp
vickyliu-go4it: Ah right! Thanks for pointing it out!

PR #2462, opened 2021-06-23
Add clang9 debug job

beckerhe:
This is fixing a bug in our ObjectUtils code which is only triggered in a debug build. Furthermore it adds the job config for a debug-build presubmit job.

### DIFF
--- a/src/ObjectUtils/CoffFile.cpp
+++ b/src/ObjectUtils/CoffFile.cpp
@@ -93,7 +93,10 @@ ErrorMessageOr<SymbolInfo> CoffFileImpl::CreateSymbolInfo(
   symbol_info.set_name(name);
   symbol_info.set_demangled_name(llvm::demangle(name));
   symbol_info.set_address(symbol_ref.getValue() + section_offset.value());
-  symbol_info.set_size(symbol_ref.getCommonSize());
+
+  llvm::object::COFFSymbolRef coff_symbol_ref = object_file_->getCOFFSymbol(symbol_ref);
+  symbol_info.set_size(coff_symbol_ref.getValue());
+
   return symbol_info;
 }
 
--- a/kokoro/builds/linux/clang9_debug/common.cfg
+++ b/kokoro/builds/linux/clang9_debug/common.cfg
@@ -1,0 +1,12 @@
+# Format: //devtools/kokoro/config/proto/build.proto
+
+# Location of the bash script. Should have value <github_scm.name>/<path_from_repository_root>.
+# github_scm.name is specified in the job configuration (next section).
+build_file: "orbitprofiler/kokoro/builds/build.sh"
+
+action {
+  define_artifacts {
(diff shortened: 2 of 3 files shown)


src/ObjectUtils/CoffFile.cpp
```diff
@@ -93,7 +93,10 @@ ErrorMessageOr<SymbolInfo> CoffFileImpl::CreateSymbolInfo(
   symbol_info.set_name(name);
   symbol_info.set_demangled_name(llvm::demangle(name));
   symbol_info.set_address(symbol_ref.getValue() + section_offset.value());
-  symbol_info.set_size(symbol_ref.getCommonSize());
```
dimitry-, line 96: link a bug in the commit message?

dimitry- (approved):

src/ObjectUtils/CoffFile.cpp
beckerhe: Oh yes, good point. Of course there is already a bug - I completely missed that. Done.

src/ObjectUtils/CoffFile.cpp
```diff
@@ -93,7 +93,10 @@ ErrorMessageOr<SymbolInfo> CoffFileImpl::CreateSymbolInfo(
...
+
+  llvm::object::COFFSymbolRef coff_symbol_ref = object_file_->getCOFFSymbol(symbol_ref);
+  symbol_info.set_size(coff_symbol_ref.getValue());
```
ronaldfw, line 98: This code doesn't look right, or I'm completely missing something. 

First, you do not need to get the COFFSymbolRef. You can get the value from `symbol_ref` as a few lines above, the getValue() method will do basically what you are doing here.

Second, `value` is the address offset, not the size. I wanted to learn more about this, but it seems that COFF symbols do not have a size; see e.g. symbol table description in https://docs.microsoft.com/en-us/windows/win32/debug/pe-format#coff-symbol-table. Also, compare ELFSymbolRef, which has a method getSize(), with COFFSymbolRef. I earlier thought common size was the right alternative, but that is not the case. 

I suspect we do not need the size at all, things worked fine when it was just set to zero. Therefore my suggestion is to remove these two lines completely.

dimitry-: interesting, don't we have a unit-test for it?

ronaldfw: Yeah, indeed interesting, there is a unit test and it has a non-zero size as well. I need to look into it.

ronaldfw: If what I'm saying above is correct, I'd expect the test should have failed.

ronaldfw: As discussed, the code actually seems to be alright for now. I'll take a closer look as a follow-up.

Thanks for the (offline) clarification @beckerhe

PR #2463, opened 2021-06-23
Fix the memory tracing e2e test

vickyliu-go4it:
In our latest change of memory track (https://github.com/google/orbit/pull/2422), we changed to visualize all the system memory usage related information in a single `MemoryTrack` with multivariate stacked graph. This change will update the e2e test accordingly.

Bug: http://b/191840564


danielfenner (approved):
lgtm! Thanks!

PR #2464, opened 2021-06-23
Fix actionOpenUserDataDirectory, add actionOpenAppDataDirectory

dpallotti:
"Open User Data Directory..." still pointed to `%APPDATA%\OrbitProfiler`, fix it
to send the user to `Documents\Orbit`.
Add "Open App Data Directory" to still be able to easily reach
`%APPDATA%\OrbitProfiler`.

Bug: http://b/191121155

### DIFF
--- a/src/OrbitQt/orbitmainwindow.ui
+++ b/src/OrbitQt/orbitmainwindow.ui
@@ -306,6 +306,9 @@ QPushButton:disabled {
     <property name="title">
      <string>Help</string>
     </property>
+    <property name="toolTipsVisible">
+     <bool>true</bool>
+    </property>
     <addaction name="actionOpenUserDataDirectory"/>
     <addaction name="actionOpenAppDataDirectory"/>
     <addaction name="separator"/>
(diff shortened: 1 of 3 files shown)


dimitry- (approved):

src/OrbitQt/orbitmainwindow.ui
```diff
@@ -522,6 +524,15 @@ QPushButton:disabled {
...
   </action>
+  <action name="actionOpenAppDataDirectory">
+   <property name="icon">
+    <iconset resource="../../icons/orbiticons.qrc">
+     <normaloff>:/actions/folder</normaloff>:/actions/folder</iconset>
+   </property>
+   <property name="text">
+    <string>Open App Data Directory...</string>
+   </property>
+  </action>
```
beckerhe, line 535: What about adding a tooltip to this action telling the user more details like this used to be the old location for captures and presets?

beckerhe (approved):
LGTM

src/OrbitQt/orbitmainwindow.ui
dpallotti: I think that they are quite hidden, but they also can't hurt, so I added tooltips for both actions. Let me know what you think of them.

beckerhe: Perfect. Thanks a lot!

PR #2465, opened 2021-06-23
Add ORBIT_DONT_MOVE_FROM_APPDATA env var to avoid moving from %APPDATA%

dpallotti:
Bug: http://b/189887834

Test: Move presets and captures back to %APPDATA%, set
`ORBIT_DONT_MOVE_FROM_APPDATA`, start Orbit, verify that no move happened.


beckerhe (approved):
LGTM, but Pierric should have a look as well

dpallotti:
> Pierric should have a look as well

Definitely, I'm waiting for his opinion, but I already had this ready so might as well open it.

pierricgimmig (approved):
Very cool, thanks, this will allow using both the old and new versions of Orbit together. Thanks Davide.

PR #2467, opened 2021-06-23
Move session setup

antonrohr:
1st commit: Move session setup related classes into component SessionSetup.

My idea behind this was to clean up the somewhat crowded component OrbitQt by extracting files into a new SessionSetup component. Now that I did this change, I am questioning if this is a good idea. Please let me know what you think.

2nd commit: Rename ProfilingTargetDialog to SessionSetupDialog.
3rd commit: Rename files `servicedeploymanager.*` to `ServiceDeployManager.*`


src/Orbit/main.cpp
```diff
@@ -193,9 +193,9 @@ static std::optional<std::string> GetCollectorPath(const QProcessEnvironment& pr
   return std::nullopt;
 }
 
-static orbit_qt::DeploymentConfiguration FigureOutDeploymentConfiguration() {
+static DeploymentConfiguration FigureOutDeploymentConfiguration() {
```
beckerhe, line 196: Would it make sense to move this function into the new session setup module?

src/SessionSetup/CMakeLists.txt
```diff
@@ -0,0 +1,87 @@
...
+          ClientServices
+          CONAN_PKG::grpc
+          CONAN_PKG::abseil
+          MetricsUploader
+          OrbitBase
+          OrbitGgp
+          OrbitSsh
+          OrbitSshQt
+          Qt5::Widgets
+          QtUtils)
```
beckerhe, line 57: Could you put our internal targets at the top and sort them alphabetically?

beckerhe (approved):
I think it makes a lot of sense. LGTM!

src/Orbit/main.cpp
antonrohr: yes, done!

src/SessionSetup/CMakeLists.txt
antonrohr: done

beckerhe: thx

PR #2468, opened 2021-06-23
Fix tests in debug build.

danielfenner:
Deactivated InstrumentFunctionTest.TooShort for debug builds. The
test depends on the length of the code emitted by the compiler (this
isn't great but it is exactly what I want to test here). In debug
mode an additional ud2 instruction is added at the end of the
function (to guard against unwanted continuation of execution I
guess).

CheckFloatParameters and CheckM256iParameters tests where using inline
assembly that implicitly relied on the specific variant of the jmp
instruction being picked by the compiler. The debug configuration
picked the unexpected variant. The code is altered such that it works
fine reguardless of what the compiler chooses to emit.


beckerhe (approved):
Wow, nice. Thank you. Looks definitely like a non-trivial fix that took some time!

PR #2469, opened 2021-06-23
Have build_and_upload_dependencies.sh use the new docker image hashes

beckerhe:
Docker images are now referenced by digests - basically upload hashes. This allow us to pin a specific version of a Docker image to each commit in the repo.

This change broke the `build_and_upload_dependencies.sh` script, so this PR cleans up this script, improves it, adds support for Docker digests and adds support for a potential public build of Orbit dependencies.


third_party/conan/docker/utils.sh
```diff
@@ -0,0 +1,24 @@
...
+
+function find_container_for_conan_profile {
+    readonly CONAN_PROFILE="$1"
+
+    # Use a specific tag from the mapping. If none is specified, fall back to `latest`.
+    readonly DOCKER_IMAGE_TAG="${docker_image_tag_mapping[${CONAN_PROFILE}]-latest}"
+
+    CONTAINER="gcr.io/orbitprofiler/${CONAN_PROFILE}:${DOCKER_IMAGE_TAG}"
+
+    if [ ${docker_image_digest_mapping[${CONAN_PROFILE}]+abc} ]; then
```
danielfenner, line 18: Looks like there is a digest for every profile. Why does this not completely replace the docker_image_tag_mapping and the code above?

beckerhe: Can probably be removed now. This mechanism was designed as having fallbacks. But yes, I can probably remove that now. I just thinking whether this has any disadvantages.

beckerhe: Fixed!

beckerhe:
@danielfenner I'm finally coming back to this. Could you have another look?

danielfenner (approved):
lgtm!

PR #2470, opened 2021-06-23
Move or remove most of the content of CoreUtils

dpallotti:


src/OrbitBase/include/OrbitBase/Append.h
```diff
@@ -0,0 +1,19 @@
...
+
+#include <vector>
+
+#ifndef ORBIT_BASE_APPEND_H_
+#define ORBIT_BASE_APPEND_H_
+
+namespace orbit_base {
+
+template <class T>
+inline void Append(std::vector<T>& dest, const std::vector<T>& source) {
```
dimitry-, line 13: add unit-tests? also probably drop inline - let the compiler figure this out.

src/OrbitCore/CoreUtils.h
```diff
@@ -33,10 +32,6 @@
 #include <utility>
 #include <vector>
 
-inline uint64_t StringHash(std::string_view str) {
-  return XXH64(str.data(), str.size(), 0xBADDCAFEDEAD10CC);
```
beckerhe, line 44: Was this also the last usage of xxHash64 in OrbitCore? If yes, could you also remove it from the dependencies in `CMakeLists.txt`?

beckerhe (approved):
Thank you soo much for this!

src/OrbitCore/CoreUtils.h
dimitry-: I believe presets still use it - old format and conversion function.

src/OrbitBase/include/OrbitBase/Append.h
dpallotti: Isn't inline needed because the definition is in the header, or does this not apply to templates?

src/OrbitCore/CoreUtils.h
dpallotti: It's used in `ClientData` but I can drop it from `OrbitCore?CMakeLists.txt`.

src/OrbitBase/include/OrbitBase/Append.h
dimitry-: I do not think it does.

dpallotti: Which part of my sentence are you referring to?

dimitry-: ... apply to templates

src/OrbitCore/CoreUtils.h
dpallotti: This is done.

src/OrbitBase/include/OrbitBase/Append.h
dpallotti: Right. Both done.

dimitry- (approved):

PR #2471, opened 2021-06-24
Make Python2 the default in Docker images

beckerhe:
Build tools like GN require that Python2 can be called via `python` on
the command line. This is fixing that.

GN is needed to compile crashpad and other dependencies.

@florian-kuebler I still have your suggested `-y`/`--yes` change in mind, but unfortunately it's not in this commit since this was already done when you added the comment. I will add this in a later change. I assume there are more changes coming now since it got so easy to update the containers.


florian-kuebler (approved):
LGTM, thanks!

PR #2472, opened 2021-06-24
Clean up StringManager and move out of OrbitCore

dpallotti:


src/OrbitGl/StringManager.cpp
```diff
@@ -10,31 +10,31 @@
...
 
+namespace orbit_gl {
+
 // TODO(b/181207737): Make this assert that it is not present and rename to "Add".
 bool StringManager::AddIfNotPresent(uint64_t key, std::string_view str) {
   absl::MutexLock lock{&mutex_};
-  if (key_to_string_.contains(key)) {
-    ERROR("String collision for key: %ul and string: %s.", key, str);
-    return false;
+  bool inserted = key_to_string_.emplace(key, str).second;
```
dimitry-, line 18: this changes the logic - we now replace the string in the case of collision. try_emplace maybe?

src/OrbitGl/StringManagerTest.cpp
```diff
@@ -0,0 +1,59 @@
...
+
+TEST(StringManager, AddOrReplace) {
+  StringManager string_manager;
+  EXPECT_TRUE(string_manager.AddIfNotPresent(0, "test1"));
+  EXPECT_FALSE(string_manager.AddOrReplace(0, "test2"));
+  EXPECT_TRUE(string_manager.AddOrReplace(1, "test3"));
+  EXPECT_FALSE(string_manager.AddOrReplace(1, "test4"));
+
+  EXPECT_TRUE(string_manager.Contains(0));
+  EXPECT_EQ("test2", string_manager.Get(0).value_or(""));
```
dimitry-, line 28: value_or("no value")?

src/OrbitGl/StringManagerTest.cpp
```diff
@@ -0,0 +1,59 @@
...
+#include "StringManager.h"
+
+namespace orbit_gl {
+
+TEST(StringManager, AddIfNotPresent) {
+  StringManager string_manager;
+  EXPECT_TRUE(string_manager.AddIfNotPresent(0, "test1"));
+  EXPECT_FALSE(string_manager.AddIfNotPresent(0, "test2"));
+
+  EXPECT_EQ("test1", string_manager.Get(0).value_or(""));
```
dimitry-, line 16: this is interesting, I was expecting emplace to replace the value.

src/OrbitGl/StringManager.cpp
dimitry-: ok, apparently emplace does not replace the element if it exists, I assumed it does.

dimitry- (approved):

src/OrbitGl/StringManager.cpp
dpallotti: This doesn't change the logic.

`emplace` and `try_emplace` do almost the same thing. The difference among those is something else and is more subtle, the high level semantics is the same.

src/OrbitGl/StringManagerTest.cpp
dpallotti: Sure, why not, done.

PR #2474, opened 2021-06-24
Add the cgroup memory track

vickyliu-go4it:
We continue the plan to refactor graph related tracks and add / update memory tacks proposed in https://github.com/google/orbit/pull/2474.  

And this PR is for the STEP 5 -- Add the `MemoryTrack` for process & cgroup memory usage.

In order to add the `MemoryTrack` for process & cgroup memory usage which takes the information provided by both `ProcessMemoryInfoProvider` and `CGroupMemoryInfoProvider`, we need to synchronize the service-side memory events.
The related discussion can be found in a doc mentioned in http://b/185107593.

Bugs: http://b/185107371, http://b/185107593
Test: Integration tests are updated. And you can run orbit with the `--enable_cgroup_memory` flag to try this feature.

Screenshots for the collapsed and uncollapsed versions.
* Trata benchmark, collapsed:
Screenshot 2021-06-24 123803

* Trata benchmark, uncollapsed:
Screenshot 2021-06-24 123810

Another example which show all the colors:
* sysmon collapsed
Screen Shot 2021-06-24 at 12 14 28

* sysmon uncollapsed
Screen Shot 2021-06-24 at 12 14 13


src/GrpcProtos/capture.proto
```diff
@@ -521,6 +521,7 @@ message ProducerCaptureEvent {
     InternedCallstack interned_callstack = 7;
     InternedString interned_string = 18;
     IntrospectionScope introspection_scope = 19;
+    MemoryEventWrapper memory_event_wrapper = 29;
```
beckerhe, line 524: This is just a question: Why are the "old" memory events removed from `ClientCaptureEvent`, but not from ProducerCaptureEvent`?

beckerhe: This makes me realize that I'm not very familiar with this code and probably can't review it properly. I'm not sure how familier @karupayun is with this. So maybe we should add another reviewer? Davide for example? Pablo, WDYT?

src/Service/MemoryInfoHandler.cpp
```diff
@@ -49,26 +54,64 @@ void MemoryInfoHandler::Stop() {
...
+        wrapper_is_ready && wrapper.has_process_memory_usage() && wrapper.has_cgroup_memory_usage();
+  }
+  if (!wrapper_is_ready) return;
+
+  uint64_t synchronized_timestamp_ns = wrapper.system_memory_usage().timestamp_ns();
+  if (enable_cgroup_memory_) {
+    synchronized_timestamp_ns = static_cast<uint64_t>(
+        (synchronized_timestamp_ns + wrapper.cgroup_memory_usage().timestamp_ns() +
+         wrapper.process_memory_usage().timestamp_ns()) /
+        3.0);
```
beckerhe, line 107: This deserves a comment explaining why you went with the artihmetic mean and not with any other averaging function.

beckerhe: Also, why `3.0`? This means it will be a floating point division and then immediately casted back to uint64_t. Can you explain the advantage of making a float point division here?

beckerhe: Also we're getting pretty close to the maximum value of `uint64_t` when adding up 3 nanosecond timestamps.

This is fine for now, but if someone wants to add more timestamps to the averaging in the future this might overflow. I suggest to add a comment and to warn future editors of this code.

src/Service/MemoryInfoHandler.cpp
```diff
@@ -49,26 +54,64 @@ void MemoryInfoHandler::Stop() {
...
+  *in_progress_wrappers_[sampling_window_id].mutable_system_memory_usage() =
+      std::move(system_memory_usage);
+  ProcessMemoryEventWrapperIfReady(sampling_window_id);
+}
+
+uint64_t MemoryInfoHandler::GetSamplingWindowId(uint64_t sample_timestamp_ns) const {
+  return static_cast<uint64_t>(
+      std::round(static_cast<double>(sample_timestamp_ns - sampling_start_timestamp_ns_) /
+                 sampling_period_ns_));
+}
```
beckerhe, line 89: This can be a static free function - doesn't need to be a member function.

beckerhe: I'm also not sure about the sampling window calculation. To me it would be more logical to put the decision boundary in the middle between to sampling periods. But here it seems to be right between the sampling periods.

This might be fine if samples can only be late, but never early. And if we cannot get out-of-sync and can have no error in the start timestamp. In this case this deserves a comment.

src/Service/MemoryInfoHandler.h
```diff
@@ -36,10 +38,18 @@ class MemoryInfoHandler : public orbit_memory_tracing::MemoryInfoListener {
...
+  void ProcessMemoryEventWrapperIfReady(uint64_t sampling_window_id);
+
   ProducerEventProcessor* producer_event_processor_;
   std::unique_ptr<orbit_memory_tracing::MemoryInfoProducer> cgroup_memory_info_producer_;
   std::unique_ptr<orbit_memory_tracing::MemoryInfoProducer> process_memory_info_producer_;
   std::unique_ptr<orbit_memory_tracing::MemoryInfoProducer> system_memory_info_producer_;
+  uint64_t sampling_start_timestamp_ns_;
+  uint64_t sampling_period_ns_;
+  bool enable_cgroup_memory_ = false;
+  absl::flat_hash_map<uint64_t, orbit_grpc_protos::MemoryEventWrapper> in_progress_wrappers_;
```
beckerhe, line 51: Nit: You could use abseil's `GUARDED_BY` to make clear that this map is synchronized by the mutex below. See https://abseil.io/docs/cpp/guides/synchronization#annotation-guidelines for details.

src/ClientProtos/capture_data.proto
```diff
@@ -170,6 +171,7 @@ message TimerInfo {
   uint64 timeline_hash = 11;
   repeated uint64 registers = 12;
   Color color = 13;
+  string cgroup_name = 14;
```
beckerhe, line 174: I can't really judge if this could have any performance implications. I guess not, but someone else should have a look - in particular at this change to the TimeInfo proto.

src/OrbitGl/App.cpp
```diff
@@ -411,35 +411,67 @@ void OrbitApp::OnTimer(const TimerInfo& timer_info) {
...
-      orbit_api::Encode<uint64_t>(system_memory_usage.available_kb());
-  encoded_values[static_cast<size_t>(SystemMemoryUsageEncodingIndex::kBuffersKb)] =
-      orbit_api::Encode<uint64_t>(system_memory_usage.buffers_kb());
-  encoded_values[static_cast<size_t>(SystemMemoryUsageEncodingIndex::kCachedKb)] =
-      orbit_api::Encode<uint64_t>(system_memory_usage.cached_kb());
-
-  *timer_info.mutable_registers() = {encoded_values.begin(), encoded_values.end()};
-
-  GetMutableTimeGraph()->ProcessTimer(timer_info, nullptr);
+  if (memory_event_wrapper.has_system_memory_usage()) {
```
beckerhe, line 414: Nit: Have you considered making separate functions for processing system and cgroup memory usage? This way we would avoid these large if-branches.

src/OrbitGl/TrackManager.h
```diff
@@ -69,6 +69,11 @@ class TrackManager {
   orbit_gl::SystemMemoryTrack* GetSystemMemoryTrack() const { return system_memory_track_.get(); }
   orbit_gl::SystemMemoryTrack* CreateAndGetSystemMemoryTrack(
       const std::array<std::string, orbit_gl::kSystemMemoryTrackDimension>& series_names);
+  orbit_gl::CGroupAndProcessMemoryTrack* GetCGroupAndProcessMemoryTrack() const {
+    return cgroup_and_process_memory_track_.get();
+  }
+  orbit_gl::CGroupAndProcessMemoryTrack* CreateAndGetCGroupAndProcessMemoryTrack(
```
beckerhe, line 75: Nit: nodiscard for both functions?

src/OrbitGl/TrackManager.cpp
```diff
@@ -441,6 +445,24 @@ SystemMemoryTrack* TrackManager::CreateAndGetSystemMemoryTrack(
   return system_memory_track_.get();
 }
 
+CGroupAndProcessMemoryTrack* TrackManager::CreateAndGetCGroupAndProcessMemoryTrack(
+    const std::array<std::string, orbit_gl::kCGroupAndProcessMemoryTrackDimension>& series_names) {
+  if (cgroup_and_process_memory_track_ == nullptr) {
+    constexpr uint8_t kTrackValueDecimalDigits = 2;
+    const std::string kTrackValueLabelUnit = "MB";
```
beckerhe, line 453: We should discuss whether we wanna write `MB` or `MiB`. I'm strongly in favor of the latter since we divide by 1024. :slightly_smiling_face:

src/MemoryTracing/MemoryTracingIntegrationTest.cpp
```diff
@@ -28,31 +34,37 @@ using orbit_memory_tracing::MemoryInfoProducer;
...
-    {
-      absl::MutexLock lock{&events_mutex_};
-      events_.emplace_back(std::move(event));
-    }
+    uint64_t sampling_window_id = GetSamplingWindowId(cgroup_memory_usage.timestamp_ns());
+
+    absl::MutexLock lock(&in_progress_wrappers_mutex_);
+    *in_progress_wrappers_[sampling_window_id].mutable_cgroup_memory_usage() =
+        std::move(cgroup_memory_usage);
+    ProcessMemoryEventWrapperIfReady(sampling_window_id);
```
beckerhe, line 49: This seems to be the same code as in `MemoryInfoHandler`. Can't we inherit from `MemoryInfoHandler` and re-use a bunch of this code - or move the shared code into `MemoryInfoListener`? To me it looks like the only difference is what we do with events once they are created.

beckerhe (commented):
Very nice Vicky. I added some comments. :sweat_smile:

src/ClientProtos/capture_data.proto
ronaldfw: @dimitry- @dpallotti Can you take a look if this is alright? 

Other strings, like e.g. the timeline string, are stored elsewhere and the TimerInfo just has a string hash. I don't have the full context if we should do the same here. I'm also not sure if we have changed our string handling in any way.

src/GrpcProtos/capture.proto
vickyliu-go4it: Removing old memory events from `ProducerCpatureEvent` and updating the integration test need to be done at the same commit. So I put this part in a later commit which updates the integration test.

src/Service/MemoryInfoHandler.cpp
vickyliu-go4it: `std::round` will find the nearest integer, which means we are putting *the decision boundary* in the middle between two sampling periods. For example, if the sampling period is 1 second, then:
* timestamp 0.99 ==> `sampling_window_id` is 1;
* timestamp 1.1 ==> `sampling_window_id` is 1;
* timestamp 1.6 ==> `sampling_window_id` is 2.

beckerhe: Ah yes. Of course. Thanks for the explanation!

src/GrpcProtos/capture.proto
```diff
@@ -452,6 +459,8 @@ message MetadataEvent {
 }
 
 message ClientCaptureEvent {
+  reserved 23, 28, 29;
```
dpallotti, line 462: @dimitry- What happens with backwards compatibility here?

src/GrpcProtos/capture.proto
karupayun: I'm seeing that Davide is also a reviewer now.

src/ClientProtos/capture_data.proto
dpallotti: Yeah, this probably duplicates the string for every "timer" and hence has memory implications. I guess a key should be used like some fields to above. 

The good thing is that we don't use `TimerInfo` for serialization anymore (only for deserialization from the old format).

dpallotti: You can have a look at `GpuDebugMarkerTrack` and `GpuSubmissionTrack` and what they do with `App::string_manager_`.

src/GrpcProtos/capture.proto
dimitry-: these events are going to be ignored when loading from old files.

src/Service/MemoryInfoHandler.cpp
karupayun: Does someone know how timestamp_ns is measuring? I've just tested 2 different captures with 2 days and an hour of difference and the timestamps seems to be 6.29 days and 8.33 days. Unless we arrive to 500 years, we shouldn't have any issue.

But we can add a comment to assure future editors we won't have a problem there.

beckerhe: So it's not a unix timestamp?

dpallotti: It's a monotonic timestamp, nanoseconds since boot.

karupayun: BTW, I still think than using average could confuse users. Shouldn't we use instead min or max timestamp to combine them?

beckerhe: > It's a monotonic timestamp, nanoseconds since boot.
Then it shouldn't be a problem.

> BTW, I still think than using average could confuse users. Shouldn't we use instead min or max timestamp to combine them?
That sounds even more confusing to me.

karupayun: ok

src/GrpcProtos/capture.proto
```diff
@@ -382,6 +382,13 @@ message CGroupMemoryUsage {
   int64 active_file_bytes = 12;
 }
 
+message MemoryEventWrapper {
```
dimitry-, line 385: nit: maybe drop Wrapper, just MemoryEvent?

src/Service/MemoryInfoHandler.cpp
dimitry-:
```
return (sample_timestamp_ns - sampling_start_timestamp_ns_ + (sampling_period_ns_/2))/sampling_period_ns_;
```

src/CaptureClient/CaptureEventProcessor.cpp
```diff
@@ -147,14 +147,8 @@ void CaptureEventProcessorForListener::ProcessEvent(const ClientCaptureEvent& ev
...
       ProcessModuleUpdate(event.module_update_event());
       break;
-    case ClientCaptureEvent::kCgroupMemoryUsage:
-      // TODO (http://b/185107371): Process the cgroup memory usage information.
-      break;
-    case ClientCaptureEvent::kProcessMemoryUsage:
-      // TODO (http://b/185107593): Process the process memory usage information.
-      break;
-    case ClientCaptureEvent::kSystemMemoryUsage:
-      ProcessSystemMemoryUsage(event.system_memory_usage());
```
dimitry-, line 157: I kind of think we should probably add an explicit `default: break;` for this break, so that we explicitly do not process anything that was removed or added in capture file.

src/GrpcProtos/capture.proto
dimitry-: also see my commend for the case in the client

src/OrbitGl/TimeGraph.cpp
```diff
@@ -481,6 +485,59 @@ void TimeGraph::ProcessMemoryTrackingTimer(const TimerInfo& timer_info) {
...
+  double cgroup_rss_anon_mb = static_cast<double>(cgroup_rss_bytes) / kMegabytesToBytes;
+  double cgroup_mapped_file_mb = static_cast<double>(cgroup_mapped_file_bytes) / kMegabytesToBytes;
+  double process_rss_anon_mb = static_cast<double>(process_rss_anon_kb) / kMegabytesToKilobytes;
+  double other_rss_anon_mb = cgroup_rss_anon_mb - process_rss_anon_mb;
+  double unused_mb = cgroup_limit_mb - cgroup_rss_anon_mb - cgroup_mapped_file_mb;
+  track->AddValues(timer_info.start(),
+                   {process_rss_anon_mb, other_rss_anon_mb, cgroup_mapped_file_mb, unused_mb});
+  track->OnTimer(timer_info);
+
+  if (!track->GetValueUpperBound().has_value()) {
```
karupayun, line 525: I know this was before in the other track but if it has a value we are not updating it? So, can I assume than this cgroup_limit_bytes is a value who doesn't change?

src/MemoryTracing/MemoryTracingIntegrationTest.cpp
```diff
@@ -131,120 +186,77 @@ class MemoryTracingIntegrationTestFixture {
...
-
-  // Verify the number of cgroup memory events only if the process's memory cgroup and the cgroup
-  // memory.stat file can be found successfully
-  if (num_received_cgroup_memory_events == 0) return;
-  EXPECT_GE(num_received_cgroup_memory_events,
-            num_expected_events_per_type - kEventCountsErrorTolerance);
-  EXPECT_LE(num_received_cgroup_memory_events,
-            num_expected_events_per_type + kEventCountsErrorTolerance);
+  EXPECT_GE(num_received_wrappers, expected_counts - kEventCountsErrorTolerance);
+  EXPECT_LE(num_received_wrappers, expected_counts + kEventCountsErrorTolerance);
```
karupayun, line 259: I could expect to receive less but how we could receive more events?

dpallotti:
@vickyliu-go4it Would you also like me to take a look at this change? There are already many reviewers here so it's probably not necessary. But I can also totally have a look.

vickyliu-go4it:
> @vickyliu-go4it Would you also like me to take a look at this change? There are already many reviewers here so it's probably not necessary. But I can also totally have a look.

If it is convenient to you, I would be happy to have your reviews and suggestions.

src/OrbitGl/App.cpp
```diff
@@ -411,35 +411,67 @@ void OrbitApp::OnTimer(const TimerInfo& timer_info) {
...
+    encoded_values[static_cast<size_t>(SystemMemoryUsageEncodingIndex::kFreeKb)] =
+        orbit_api::Encode<uint64_t>(system_memory_usage.free_kb());
+    encoded_values[static_cast<size_t>(SystemMemoryUsageEncodingIndex::kAvailableKb)] =
+        orbit_api::Encode<uint64_t>(system_memory_usage.available_kb());
+    encoded_values[static_cast<size_t>(SystemMemoryUsageEncodingIndex::kBuffersKb)] =
+        orbit_api::Encode<uint64_t>(system_memory_usage.buffers_kb());
+    encoded_values[static_cast<size_t>(SystemMemoryUsageEncodingIndex::kCachedKb)] =
+        orbit_api::Encode<uint64_t>(system_memory_usage.cached_kb());
+
+    *timer_info.mutable_registers() = {encoded_values.begin(), encoded_values.end()};
```
dpallotti, line 439: I'm confused here. `TimerInfo::registers`, which are specific to manual and dynamic instrumentation, are used to store memory data. And `orbit_api::Encode`, which is specific to manual instrumentation, is used as a tool for this. This is quite a hack. I think it should be implemented differently with dedicated fields.

src/OrbitGl/App.cpp
```diff
@@ -411,35 +411,67 @@ void OrbitApp::OnTimer(const TimerInfo& timer_info) {
...
+    encoded_values[static_cast<size_t>(CGroupAndProcessMemoryUsageEncodingIndex::kCGroupRssBytes)] =
+        orbit_api::Encode<uint64_t>(cgroup_memory_usage.rss_bytes());
+    encoded_values[static_cast<size_t>(
+        CGroupAndProcessMemoryUsageEncodingIndex::kCGroupMappedFileBytes)] =
+        orbit_api::Encode<uint64_t>(cgroup_memory_usage.mapped_file_bytes());
+    encoded_values[static_cast<size_t>(
+        CGroupAndProcessMemoryUsageEncodingIndex::kProcessRssAnonKb)] =
+        orbit_api::Encode<uint64_t>(process_memory_usage.rss_anon_kb());
+
+    *timer_info.mutable_registers() = {encoded_values.begin(), encoded_values.end()};
```
dpallotti, line 471: Same as above.

src/OrbitGl/TimeGraph.cpp
```diff
@@ -481,6 +485,59 @@ void TimeGraph::ProcessMemoryTrackingTimer(const TimerInfo& timer_info) {
...
+
+  constexpr double kMegabytesToBytes = 1024.0 * 1024.0;
+  constexpr double kMegabytesToKilobytes = 1024.0;
+  orbit_gl::CGroupAndProcessMemoryTrack* track = track_manager_->GetCGroupAndProcessMemoryTrack();
+  if (track == nullptr) {
+    const std::array<std::string, orbit_gl::kCGroupAndProcessMemoryTrackDimension> kSeriesNames = {
+        absl::StrFormat("Process [%s] Resident Anonymous Memory", capture_data_->process_name()),
+        "Other Processes Resident Anonymous Memory",
+        absl::StrFormat("CGroup [%s] Mapped File", timer_info.cgroup_name()),
+        absl::StrFormat("CGroup [%s] Unused", timer_info.cgroup_name())};
```
dpallotti, line 512: From the screenshot, these seem to be very wide, potentially going past the width of the track even when the window is quite large.

src/OrbitGl/TrackManager.cpp
dpallotti: I'm rather neutral on the topic, but note that until now in all other places where memory units are involved (not many though) we use the `MB` version (see `GetDisplaySize`), so we also need to be consistent (one way or the other).

src/OrbitGl/MemoryTrack.h
```diff
@@ -52,6 +52,19 @@ using SystemMemoryTrack = MemoryTrack<kSystemMemoryTrackDimension>;
...
+    Color(231, 68, 53, 255),  // red
+    Color(246, 196, 0, 255),  // orange
+    Color(87, 166, 74, 255)   // green
+};
+
+const std::array<Color, kCGroupAndProcessMemoryTrackDimension> kCGroupAndProcessMemoryTrackColors{
+    Color(231, 68, 53, 255),   // red
+    Color(249, 96, 111, 255),  // warm red
+    Color(246, 196, 0, 255),   // orange
+    Color(87, 166, 74, 255)    // green
```
dpallotti, line 65: Where were these specific color values taken from? Or were they hand-picked? I'm asking because if there is a special "source" it would be useful to have a comment in case one then needs to add more or change them.

karupayun:
> * Trata benchmark, collapsed:
>   Screenshot 2021-06-24 123803
> * Trata benchmark, uncollapsed:
>   Screenshot 2021-06-24 123810
> 

It doesn't seem to be a straight line in top of the graphics, but I don't understand why. Could you explain me?

src/GrpcProtos/capture.proto
```diff
@@ -382,6 +382,13 @@ message CGroupMemoryUsage {
...
+  uint64 timestamp_ns = 1;
```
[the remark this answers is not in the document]
vickyliu-go4it: This is intended. The timestamp in a specific memory event (e.g., `SystemMemoryUsage`, `CGroupMemoryUsage`, or `ProcessMemoryUsage`) records the time when this memory event is actually collected. 

On one hand, for events collected in the same sampling window, we can check whether the difference between those timestamps is small enough in MemoryTracingIntegrationTest; on the other hand, they are recorded in case we need the more precise timestamp information.

vickyliu-go4it:
> > * Trata benchmark, collapsed:
> >   Screenshot 2021-06-24 123803
> > * Trata benchmark, uncollapsed:
> >   Screenshot 2021-06-24 123810
> 
> It doesn't seem to be a straight line in top of the graphics (all the stacked graphs added together), but I don't understand why. Could you explain me?

The `MultvariateTimeSeries` stores time series value as `double` type, however the y position (computed from the time series value) is `float` type. Due to the loss of precision, the top of stacked graphs are lot flat.

karupayun:
> > > * Trata benchmark, collapsed:
> > >   Screenshot 2021-06-24 123803
> > > * Trata benchmark, uncollapsed:
> > >   Screenshot 2021-06-24 123810
> > 
> > 
> > It doesn't seem to be a straight line in top of the graphics (all the stacked graphs added together), but I don't understand why. Could you explain me?
> 
> The `MultvariateTimeSeries` stores time series value as `double` type, however the y position (computed from the time series value) is `float` type. Due to the loss of precision, the top of stacked graphs are lot flat.

Are you sure that it is related with precision? If this is the case, I would like to round them. As an user I will expect the top part of the graphics to be flat.

vickyliu-go4it:
> > > > * Trata benchmark, collapsed:
> > > >   Screenshot 2021-06-24 123803
> > > > * Trata benchmark, uncollapsed:
> > > >   Screenshot 2021-06-24 123810
> > > 
> > > 
> > > It doesn't seem to be a straight line in top of the graphics (all the stacked graphs added together), but I don't understand why. Could you explain me?
> > 
> > 
> > The `MultvariateTimeSeries` stores time series value as `double` type, however the y position (computed from the time series value) is `float` type. Due to the loss of precision, the top of stacked graphs are lot flat.
> 
> Are you sure that it is related with precision? If this is the case, I would like to round them. As an user I will expect the top part of the graphics to be flat.

Thanks for the suggestion. I just added one commit to round the precision of memory usage data. Now when you check the sum of numbers listed in the white label, the sum is fixed. And the  top of system track becomes flat now. 

But the top of cgroup tracks still looks not very flat. I checked that the sum of numbers listed in the white label is fixed now. So I guess the problem is caused by the loss of precision when converting the time series value to the graph y position.

src/Service/MemoryInfoHandler.cpp
vickyliu-go4it: Thanks for the suggestion. This part has been changed accordingly and moved to the `MemoryInfoListener::GetSynchronizedSamplingTimestamp` method which returns the arithmetic mean.

src/Service/MemoryInfoHandler.cpp
vickyliu-go4it: Changed to return ` (sample_timestamp_ns - sampling_start_timestamp_ns_ + (sampling_period_ns_/2))/sampling_period_ns_;
` and moved to `MemoryInfoListener.cpp` as a static non-member fucntion. Thanks for the suggestion!

src/Service/MemoryInfoHandler.h
vickyliu-go4it: Changed accordingly. Now this part is moved in to `MemoryInfoListener.h`. Thanks for the suggestion!

src/ClientProtos/capture_data.proto
vickyliu-go4it: Thanks for the suggestion. Now I change to store the string hash in the `TimerInfo` and recover the string value with `App::string_manager_`.

Also the `OnMemoryUsageEvents` method, which converts a  `MemoryUsageEvent` into a `TimerInfo`, has been moved from `App.cpp` to `CaptureEventProcessor.cpp`. And the corresponding test has been added to `CaptureEventProcessorTest.cpp`.

src/OrbitGl/App.cpp
vickyliu-go4it: Thanks for the suggestion. Now they are put into separate functions, i.e., `CaptureEventProcessorForListener::ExtractAndProcessSystemMemoryTrackingTimer` and `CaptureEventProcessorForListener::ExtractAndProcessCGroupAndProcessMemoryTrackingTimer` in `CaptureEventProcessor.cpp`.

src/OrbitGl/TrackManager.h
vickyliu-go4it: Done! Thanks!

src/OrbitGl/TrackManager.cpp
vickyliu-go4it: Thanks for the suggestion. I will prepare a follow up PR to change to consistently use `KiB` and `MiB` as the size units.

src/MemoryTracing/MemoryTracingIntegrationTest.cpp
vickyliu-go4it: Thanks for the suggestion. As we discussed offline, this part has been moved to `MemoryInfoListener`.

src/GrpcProtos/capture.proto
vickyliu-go4it: Thanks for the suggestion but I cannot find the comments for the case in the client 😂

src/GrpcProtos/capture.proto
vickyliu-go4it: Thanks for all the suggestions! Changed to use `MemoryUsageEvent`.

src/CaptureClient/CaptureEventProcessor.cpp
vickyliu-go4it: Good point! Added accordingly. Thanks!

src/OrbitGl/TimeGraph.cpp
vickyliu-go4it: We can update the cgroup limit but we usually don't do this. We can assume that `cgroup_limit_bytes` is a fixed value at least won't change while taking the capture.

src/MemoryTracing/MemoryTracingIntegrationTest.cpp
vickyliu-go4it: Ah, right, the `EXPECT_LE` case is redundant. Removed now. Thanks!

src/Service/ProducerEventProcessor.cpp
```diff
@@ -347,22 +346,10 @@ void ProducerEventProcessorImpl::ProcessFullTracepointEvent(
   capture_event_buffer_->AddEvent(std::move(event));
 }
 
-void ProducerEventProcessorImpl::ProcessCGroupMemoryUsage(CGroupMemoryUsage* cgroup_memory_usage) {
+void ProducerEventProcessorImpl::ProcessMemoryEventWrapper(
```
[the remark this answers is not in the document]
vickyliu-go4it: Thanks for the suggestion! Test has been added now.

src/OrbitGl/MemoryTrack.h
vickyliu-go4it: Thanks for the suggestioin! I have added a comment to mention the place to get the colors https://convertingcolors.com/.

Now the basic idea is to guarantee the used memories (e.g., `Used` in system memory track, `Process RssAnon` in the cgroup memory track) are visualized in different red color; the unused memories (e.g., `Uused` in system memory track and the cgroup memory track) are visualized in different green colors; and the cached memories (e.g., `Buffers / Cached` in system memory track and `CGroup Mapped Files`) are visualized in different yellow colors.

src/OrbitGl/TimeGraph.cpp
karupayun: Ok, so I would like to not have this every time we process a new timer. Instead moving this logic to the constructor of this track, and having a method to set it again. Not to do in this PR, but should be part of the refactor planned from the tracks and the timers. Could you add a TODO?

src/CaptureClient/CaptureEventProcessorTest.cpp
```diff
@@ -549,6 +552,89 @@ TEST(CaptureEventProcessor, CanHandleGpuJobs) {
...
 }
 
+TEST(CaptureEventProcessor, CanHandleMemoryEventWrapper) {
+  MockCaptureListener listener;
+  auto event_processor =
+      CaptureEventProcessor::CreateForCaptureListener(&listener, std::filesystem::path{}, {});
+
+  ClientCaptureEvent event;
+  MemoryEventWrapper* memory_event_wrapper = event.mutable_memory_event_wrapper();
+  memory_event_wrapper->set_timestamp_ns(110);
```
karupayun: Could I ask for a comment explaining that we are using an average here and moving this line after setting the timestamp from all the MemoryUsage or where we are checking that?

src/OrbitGl/TimeGraph.cpp
vickyliu-go4it: Thanks! I have added a todo and created a bug for this!

karupayun (commented):
Thanks for addressing my comments. I've made a question and also I would be interested in solving the issue with the top part of the stacked graph not being flat. Could be a round solve the issue there also if it is related with precision?

src/CaptureClient/CaptureEventProcessorTest.cpp
vickyliu-go4it: Done! Thanks for the suggestion!

vickyliu-go4it:
> Thanks for addressing my comments. I've made a question and also I would be interested in solving the issue with the top part of the stacked graph not being flat. Could be a round solve the issue there also if it is related with precision?

It becomes flat now https://screenshot.googleplex.com/4254FNZVMUpr4ZW! Thanks for the suggestion!

src/OrbitGl/App.cpp
dpallotti: > I'm not sure whether it is a good idea to add all the fields into TimerInfo as they are used only for the memory tracks.

Your doubt comes from the fact that `TimerInfo` is a big mess, unfortunately. There are various fields, but most only makes sense in some cases, based on the `type` of `TimerInfo`. Considering the current state, I don't see a problem adding fields only used for memory tracks. I think that reusing fields that were introduced for other purposes (like `registers`) is worse.

You could wrap memory information into a sub-message `MemoryTimerInfo` (the name is just an example) of `TimerInfo`. Or you could do something similar to adding a repeated field of `VariableInfo` to `TimerInfo` as you suggested, but I would change `VariableInfo` to something that makes it clearer it's for memory, e.g., `MemoryValue` or something like that.

Now, all this said, I just noticed that `MemoryTrack` is not a `TimerTrack` (it inherits from `GraphTrack` which is just a `Track`). In that case, I think that you should just not use `TimerInfo`s at all, nor the `OnTimer` methods, it just makes things more complicated. I see that `GraphTrack` already uses `TimerInfo`s but I'm not sure of the point. In fact, you don't need to use protobufs at all (`TimerInfo` is a proto to load from the old capture format). You should probably just route instances of dedicated classes or structs to the `MemoryTrack`s.

src/CaptureClient/CaptureEventProcessor.cpp
```diff
@@ -363,9 +367,78 @@ void CaptureEventProcessorForListener::ProcessGpuQueueSubmission(
...
+}
+
+void CaptureEventProcessorForListener::ExtractAndProcessSystemMemoryTrackingTimer(
+    uint64_t synchronized_timestamp_ns,
     const orbit_grpc_protos::SystemMemoryUsage& system_memory_usage) {
-  capture_listener_->OnSystemMemoryUsage(system_memory_usage);
+  TimerInfo timer;
+  timer.set_type(TimerInfo::kSystemMemoryUsage);
+  timer.set_start(synchronized_timestamp_ns);
+  timer.set_end(synchronized_timestamp_ns);
```
dpallotti, line 391: This supports even more that `TimerInfo` should just not be used here, this is a single point in time, not a time slice, which is what `TimerInfo` is supposed to represent.

src/MemoryTracing/include/MemoryTracing/MemoryInfoListener.h
```diff
@@ -13,9 +17,24 @@ namespace orbit_memory_tracing {
...
+      const std::vector<uint64_t>& sampling_timestamps);
+
+  absl::flat_hash_map<uint64_t, orbit_grpc_protos::MemoryUsageEvent>
+      in_progress_memory_usage_events_ GUARDED_BY(in_progress_memory_usage_events_mutex_);
+  absl::Mutex in_progress_memory_usage_events_mutex_;
+  uint64_t sampling_start_timestamp_ns_;
+  uint64_t sampling_period_ns_;
+
+ private:
+  virtual void ProcessMemoryUsageEventIfReady(uint64_t sampling_window_id) = 0;
```
dpallotti, line 37: I am quite skeptical about the design of this class. The subclasses should just have to implement a method to handle the complete `ProducerCaptureEvent` (e.g., `OnMemoryUsageEvent`) containing the `MemoryUsageEvent`. `ProcessMemoryUsageEventIfReady` should be private here and implemented, and call `OnMemoryUsageEvent` when the event is ready. 

Instead, right now the subclasses now have to implement a lot of very specific logic based on `protected` fields and methods, which should be `private` instead. Consider how much logic is in `BufferMemoryInfoListener`, which should instead just be a very simple class, as it's just a fake for tests.

src/MemoryTracing/MemoryTracingIntegrationTest.cpp
```diff
@@ -131,120 +152,76 @@ class MemoryTracingIntegrationTestFixture {
...
+      EXPECT_TRUE(cgroup_memory_usage.limit_bytes() >= 0);
+      EXPECT_TRUE(cgroup_memory_usage.rss_bytes() >= 0);
+      EXPECT_TRUE(cgroup_memory_usage.mapped_file_bytes() >= 0);
+      EXPECT_TRUE(cgroup_memory_usage.pgfault() >= 0);
+      EXPECT_TRUE(cgroup_memory_usage.pgmajfault() >= 0);
+      EXPECT_TRUE(cgroup_memory_usage.unevictable_bytes() >= 0);
+      EXPECT_TRUE(cgroup_memory_usage.inactive_anon_bytes() >= 0);
+      EXPECT_TRUE(cgroup_memory_usage.active_anon_bytes() >= 0);
+      EXPECT_TRUE(cgroup_memory_usage.inactive_file_bytes() >= 0);
+      EXPECT_TRUE(cgroup_memory_usage.active_file_bytes() >= 0);
```
dpallotti, line 204: Prefer `EXPECT_GE` to `EXPECT_TRUE(... > ...)`

src/OrbitGl/MemoryTrack.h
dpallotti: I'm not sure I understand. That just looks like a website to convert between color spaces. How were the specific shades of these colors chosen?

src/CaptureClient/CaptureEventProcessor.cpp
vickyliu-go4it: Make sense. I have added a bug for using dedicated structs for the memory tracks and also add a TODO here. Thanks for pointing out this!

src/MemoryTracing/include/MemoryTracing/MemoryInfoListener.h
vickyliu-go4it: Thanks for the suggestion! Changed accordingly.

src/MemoryTracing/MemoryTracingIntegrationTest.cpp
vickyliu-go4it: Done. Thanks!

src/OrbitGl/MemoryTrack.h
vickyliu-go4it: I changed the comments to explain how the colors are chosen. Sorry for the previous unclear comments🙈

dpallotti (approved):
I'm still not a big fan of how `TimerInfo`'s `registers` are abused, but apart from that it looks good. I'm not too familiar with `Track`s and `TrackManager`, maybe you want to wait for someone else's review.

karupayun:
> I'm still not a big fan of how `TimerInfo`'s `registers` are abused, but apart from that it looks good. I'm not too familiar with `Track`s and `TrackManager`, maybe you want to wait for someone else's review.

We will need to do something for the next steps, as we won't use timer_info anymore. I won't think this should be part of this PR, but this bug should be addressed shortly in the next months.

karupayun (approved):
Thanks for addressing my comments. By my side, LGTM!

src/Service/ProducerEventProcessorTest.cpp
```diff
@@ -1610,4 +1615,54 @@ TEST(ProducerEventProcessor, MetadataEvent) {
   EXPECT_EQ(actual_metadata_event.info_event().message(), kMessage);
 }
 
+TEST(ProducerEventProcessor, MemoryEventWrapper) {
+  MockCaptureEventBuffer buffer;
+  auto producer_event_processor = ProducerEventProcessor::Create(&buffer);
```
beckerhe, line 1620: Nit: Could you move these two lines down to where `buffer` is used first time?

src/MemoryTracing/include/MemoryTracing/MemoryInfoListener.h
```diff
@@ -5,6 +5,10 @@
 #ifndef MEMORY_TRACING_MEMORY_INFO_LISTENER_H_
 #define MEMORY_TRACING_MEMORY_INFO_LISTENER_H_
 
+#include <absl/base/internal/thread_annotations.h>
```
beckerhe, line 8: I think this should be `#include <absl/base/thread_annotations.h>` (without the `internal`).

src/MemoryTracing/include/MemoryTracing/MemoryInfoListener.h
```diff
@@ -13,9 +17,38 @@ namespace orbit_memory_tracing {
...
+  }
+
+  void OnSystemMemoryUsage(orbit_grpc_protos::SystemMemoryUsage system_memory_usage);
+  void OnCGroupMemoryUsage(orbit_grpc_protos::CGroupMemoryUsage cgroup_memory_usage);
+  void OnProcessMemoryUsage(orbit_grpc_protos::ProcessMemoryUsage process_memory_usage);
+
+ private:
+  // This method compute the arithmetic mean of input timestamps.
+  [[nodiscard]] static uint64_t GetSynchronizedSamplingTimestamp(
+      const std::vector<uint64_t>& sampling_timestamps);
```
beckerhe, line 41: Nit: This does not need to be declared in the class, does it? It could be a simple free static function in `MemoryInfoListener.cpp`, couldn't it?

beckerhe (approved):
Thank you Vicky. Sorry for being so late. LGTM!

src/Service/ProducerEventProcessorTest.cpp
vickyliu-go4it: Done! Thanks!

src/MemoryTracing/include/MemoryTracing/MemoryInfoListener.h
vickyliu-go4it: Fixed! Thanks !

src/MemoryTracing/include/MemoryTracing/MemoryInfoListener.h
vickyliu-go4it: Changed accordingly! Thanks!

PR #2475, opened 2021-06-24
Improve error handling for ElfFile

dimitry-:
Split construction and initialization of ElfFile,
report and propogate errors that happen on initialization.

Test: run ObjectFileTests

### DIFF
--- a/src/ObjectUtils/ElfFile.cpp
+++ b/src/ObjectUtils/ElfFile.cpp
@@ -54,10 +54,13 @@ class ElfFileImpl : public ElfFile {
   ElfFileImpl(std::filesystem::path file_path,
               llvm::object::OwningBinary<llvm::object::ObjectFile>&& owning_binary);
 
+  ErrorMessageOr<void> Initialize();
+
   // Loads symbols from the .symtab section.
   [[nodiscard]] ErrorMessageOr<ModuleSymbols> LoadDebugSymbols() override;
   [[nodiscard]] ErrorMessageOr<ModuleSymbols> LoadSymbolsFromDynsym() override;
-  [[nodiscard]] ErrorMessageOr<uint64_t> GetLoadBias() const override;
+  [[nodiscard]] uint64_t GetLoadBias() const override;
+  [[nodiscard]] uint64_t GetExecutableSegmentOffset() const override;
   [[nodiscard]] bool HasDebugSymbols() const override;
   [[nodiscard]] bool HasDynsym() const override;
   [[nodiscard]] bool HasDebugInfo() const override;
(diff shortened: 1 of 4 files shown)


src/ObjectUtils/ElfFile.cpp
```diff
@@ -563,11 +602,15 @@ ErrorMessageOr<std::unique_ptr<ElfFile>> CreateElfFile(
...
+        std::make_unique<ElfFileImpl<llvm::object::ELF32LE>>(file_path, std::move(file));
+    OUTCOME_TRY(instance->Initialize());
+    result = std::move(instance);
   } else if (llvm::dyn_cast<llvm::object::ELF64LEObjectFile>(object_file) != nullptr) {
-    result = std::unique_ptr<ElfFile>(
-        new ElfFileImpl<llvm::object::ELF64LE>(file_path, std::move(file)));
+    auto instance =
+        std::make_unique<ElfFileImpl<llvm::object::ELF64LE>>(file_path, std::move(file));
+    OUTCOME_TRY(instance->Initialize());
+    result = std::move(instance);
```
beckerhe, line 613: I assume they code was different at some point, but now I don't see a reason for having `result`. Why not return early in each of these cases?

beckerhe (approved):
Nice. Otherwise LGTM

src/ObjectUtils/ElfFile.cpp
dimitry-: makes sense. done.

src/ObjectUtils/ElfFile.cpp
```diff
@@ -178,57 +197,66 @@ void ElfFileImpl<ElfT>::InitDynamicEntries(const llvm::object::ELFFile<ElfT>* el
...
   }
 
   auto strtab_last_byte_or_error = elf_file->toMappedAddr(dynamic_string_table_addr.value() +
                                                           dynamic_string_table_size.value() - 1);
   if (!strtab_last_byte_or_error) {
-    LOG("Unable get last byte address of dynamic string table \"%s\": %s", file_path_.string(),
-        llvm::toString(strtab_last_byte_or_error.takeError()));
-    return;
+    auto error_message =
+        absl::StrFormat("Unable get last byte address of dynamic string table \"%s\": %s",
```
ronaldfw, line 207:
```suggestion
        absl::StrFormat("Unable to get last byte address of dynamic string table \"%s\": %s",
```

src/ObjectUtils/include/ObjectUtils/ElfFile.h
```diff
@@ -53,7 +53,8 @@ class ElfFile : public ObjectFile {
   //
   // This method returns load bias for the elf-file if program headers are
   // available. This should be the case for all loadable elf-files.
-  [[nodiscard]] virtual ErrorMessageOr<uint64_t> GetLoadBias() const = 0;
+  [[nodiscard]] virtual uint64_t GetLoadBias() const = 0;
+  [[nodiscard]] virtual uint64_t GetExecutableSegmentOffset() const = 0;
```
ronaldfw, line 57: Is this a function you need for the absolute address changes you are working on? (Just wondering, as this is not yet used anywhere other than tests it seems.)

ronaldfw (approved):
LGTM

I suppose similar changes can be / need to be done for the COFFFile class eventually? (I can take care of these later today, if we need them. I need to look at the code there anyway.)

src/ObjectUtils/include/ObjectUtils/ElfFile.h
dimitry-: Yes, it is going to be used in the next PR; the initial change was too big - so I split it to 2.

src/ObjectUtils/ElfFile.cpp
dimitry-: done.

dimitry-: done.

PR #2478, opened 2021-06-25
Move introspection from OrbitBase to new Introspection module

dpallotti:
#### Don't introspect ThreadPool
This prepares removing the dependency of `OrbitBase` on introspection.
Either way, `ThreadPool` is quite a low level component to be introspected.
If we want to keep it nonetheless, the alternative is to just move `ThreadPool`
to its own module.

Bug: http://b/182473973
#### Move introspection from OrbitBase to new Introspection module
Removes the dependency of `OrbitBase` on `ApiInterface`.

Bug: http://b/182473973

Test:
- Capture `OrbitTest` to see that manual instrumentation is still working.
- Run in `--devmode` to see that introspection (both service and client side) is
  still working.


src/OrbitBase/ThreadPool.cpp
```diff
@@ -179,17 +168,14 @@ void ThreadPoolImpl::WorkerFunction() {
       break;
     }
 
-    bool auto_profile = auto_profile_;
     mutex_.Unlock();
-    if (auto_profile) ORBIT_START("Execute Action");
     action->Execute();
-    if (auto_profile) ORBIT_STOP();
```
dpallotti, line 186: @pierricgimmig I'm for just removing this, but let me know if you need introspection for `ThreadPool` to be kept. In that case I could move `ThreadPool` out of `OrbitBase` to its own module.

pierricgimmig: I'd prefer keeping it, yes, as it catches a lot of interesting scopes automatically. Could we set some callbacks on the ThreadPool so that we don't have to move it out of OrbitBase? Something like OnActionStarted/OnActionCompleted maybe? I was thinking also that at some point we could provide a name for actions so that they could show up in the introspection scope, that would be a different change.

dpallotti: Where or by whom would those callbacks be set?

pierricgimmig (approved):
Thanks for this Davide. I'm totally for putting introspection into its own module. However, we're removing useful functionality by not instrumenting thread pools anymore. If we can't easily find an alternative implementation, let's open a tracking bug. Otherwise, lgtm.

dpallotti:
> If we can't easily find an alternative implementation, let's open a tracking bug.

If you feel that we should keep the introspection in the `ThreadPool`, as I said above the easy solution is to move the `ThreadPool` to its own module (really not a big deal). Let me know if you prefer that.

src/OrbitBase/ThreadPool.cpp
pierricgimmig: I was thinking wherever ThreadPool::Create is called, we could specify callbacks that would in turn call ORBIT_START/STOP.

pierricgimmig:
> If you feel that we should keep the introspection in the `ThreadPool`, as I said above the easy solution is to move the `ThreadPool` to its own module (really not a big deal). Let me know if you prefer that.

If that's the easy solution, let's do that then, thanks!

src/OrbitBase/ThreadPool.cpp
dpallotti: @dimitry- What do you think? I thought this would complicate the call sites, but it seems that only the `ThreadPool` created in `App.cpp` would be affected.

dpallotti: @dimitry- As discussed, I added the possibility to wrap the `Action` in `ScheduleImpl`. Please have a look.

src/Introspection/IntrospectionTest.cpp
```diff
@@ -1,7 +1,8 @@
-// Copyright (c) 2020 The Orbit Authors. All rights reserved.
+// Copyright (c) 2021 The Orbit Authors. All rights reserved.
```
dimitry-, line 1: mm? I do not think we want to do it on move...

dimitry- (approved):
otherwise lgtm

src/Introspection/IntrospectionTest.cpp
dpallotti: The internal guidelines say that changing this on modification is optional (because it would be annoying). But I don't know why I did it so let me just change it back.

dpallotti:
> > If you feel that we should keep the introspection in the `ThreadPool`, as I said above the easy solution is to move the `ThreadPool` to its own module (really not a big deal). Let me know if you prefer that.
> 
> If that's the easy solution, let's do that then, thanks!

So we went with the other way. The `ThreadPool` in `OrbitApp` is instrumented.

PR #2480, opened 2021-06-25
Make timebar lighter and opaque

karupayun:
Now tracks are not going to be drawn behind it. Also we fixed the fact
that the last track didn't fit in view (http://b/191335338).

I wanted to make a refactor but after speaking with @reichlfl, I decided to create a bug there.

timebar lighter and opaque


florian-kuebler (approved):
Thanks

reichlfl (approved):

PR #2481, opened 2021-06-26
Report only executable segmetents in Module events

dimitry-:
Since modules may end up being mapped between other modules,
and sometimes do (see http://b/191732187), we only report
now executable segment mappings for modules. This also brings
snapshot event data in line with module update data.

Since exectuable segment is located at the offset different
from 0 we need to take it into account while translating
absolute addresses to function address (and vice versa).
For this reason we introduce exectuable_segment_offset
to ModuleInfo

Bug: http://b/191732187
Test: Run orbit, intstrument a function, take a capture, go to
      Disassebly.

### DIFF
--- a/src/ClientData/FunctionUtils.cpp
+++ b/src/ClientData/FunctionUtils.cpp
@@ -63,7 +63,8 @@ std::optional<uint64_t> GetAbsoluteAddress(const orbit_client_protos::FunctionIn
 
   CHECK(!base_addresses.empty());
 
-  return func.address() + base_addresses.at(0) - module.load_bias();
+  return func.address() + base_addresses.at(0) - module.load_bias() -
+         module.executable_segment_offset();
 }
 
 bool IsOrbitFunctionFromType(const FunctionInfo::OrbitType& type) {
--- a/src/ObjectUtils/LinuxMap.cpp
+++ b/src/ObjectUtils/LinuxMap.cpp
@@ -67,18 +67,12 @@ ErrorMessageOr<ModuleInfo> CreateModule(const std::filesystem::path& module_path
     CHECK(elf_file != nullptr);
     module_info.set_build_id(elf_file->GetBuildId());
     module_info.set_soname(elf_file->GetSoname());
-
-    ErrorMessageOr<uint64_t> load_bias_or_error = elf_file->GetLoadBias();
-    // Every loadable ELF module contains a load bias.
-    if (load_bias_or_error.has_error()) {
-      return load_bias_or_error.error();
-    }
-    module_info.set_load_bias(load_bias_or_error.value());
+    module_info.set_load_bias(elf_file->GetLoadBias());
+    module_info.set_executable_segment_offset(elf_file->GetExecutableSegmentOffset());
   }
 
   // All fields we need to set for COFF files are already set, no need to handle COFF
   // specifically here.
-
   return module_info;
 }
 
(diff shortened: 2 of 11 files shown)


src/ObjectUtils/LinuxMap.cpp
```diff
@@ -113,30 +102,16 @@ ErrorMessageOr<std::vector<ModuleInfo>> ParseMaps(std::string_view proc_maps_dat
...
-      address_range.is_executable |= is_executable;
-    }
-  }
-
-  std::vector<ModuleInfo> result;
-  for (const auto& [module_path, address_range] : address_map) {
-    // Filter out entries which are not executable
-    if (!address_range.is_executable) continue;
+    // Skip non-executable mappings
+    if (!is_executable) continue;
```
danielfenner, line 106: This should do the fix for the case documented in b/191732187. I wonder if there are real world cases with multiple executable mappings per module. Probably there is little use in asking that - we'll find out ...

danielfenner (approved):
lgtm

src/ClientData/FunctionUtils.cpp
```diff
@@ -63,7 +63,13 @@ std::optional<uint64_t> GetAbsoluteAddress(const orbit_client_protos::FunctionIn
 
   CHECK(!base_addresses.empty());
 
-  return func.address() + base_addresses.at(0) - module.load_bias();
+  auto result = func.address() + base_addresses.at(0) - module.load_bias() -
+                module.executable_segment_offset();
+  LOG("Calculating absolute address for %s: %#x(function_address) + %#x(base_address) - "
+      "%#x(load_bias) - %#x(segment_offset) = %#x",
+      func.pretty_name(), func.address(), base_addresses.at(0), module.load_bias(),
+      module.executable_segment_offset(), result);
```
dpallotti, line 71: How often is this function called? Doesn't this spam the logs? This looks more intended for development.

src/ClientData/FunctionUtils.cpp
```diff
@@ -63,7 +63,13 @@ std::optional<uint64_t> GetAbsoluteAddress(const orbit_client_protos::FunctionIn
 
   CHECK(!base_addresses.empty());
 
-  return func.address() + base_addresses.at(0) - module.load_bias();
+  auto result = func.address() + base_addresses.at(0) - module.load_bias() -
+                module.executable_segment_offset();
+  LOG("Calculating absolute address for %s: %#x(function_address) + %#x(base_address) - "
```
dpallotti, line 68:
```suggestion
  LOG("Calculating absolute address for \"%s\": %#x(function_address) + %#x(base_address) - "
```

src/ObjectUtils/LinuxMap.cpp
dimitry-: yep - this is currently assumes that there is only one executable segment. I think for now it is good enough.

dpallotti (approved):
I'm just wondering about the `LOG`, looks good otherwise.

ronaldfw (approved):
LGTM

COFF file support should still work as nothing changes if executable segment offset is 0 (if I'm not mistaken), o/w I'll fix it

src/ClientData/FunctionUtils.cpp
dimitry-: ops - this was debugging leftover log - removed.

src/ClientData/FunctionUtils.cpp
dimitry-: removed.

PR #2483, opened 2021-06-28
Fix build

florian-kuebler:
I missed to rebase #2455 before merging. This should fix the broken
build

Test: Compile

### DIFF
--- a/src/OrbitGl/GraphTrack.cpp
+++ b/src/OrbitGl/GraphTrack.cpp
@@ -29,9 +29,6 @@ GraphTrack<Dimension>::GraphTrack(CaptureViewElement* parent, TimeGraph* time_gr
       series_(MultivariateTimeSeries<Dimension>(series_names)) {
   SetName(name);
   SetLabel(name);
-
-  collapse_toggle_->SetState(TriangleToggle::State::kExpanded,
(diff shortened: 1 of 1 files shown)


src/OrbitGl/GraphTrack.cpp
```diff
@@ -29,9 +29,6 @@ GraphTrack<Dimension>::GraphTrack(CaptureViewElement* parent, TimeGraph* time_gr
       series_(MultivariateTimeSeries<Dimension>(series_names)) {
   SetName(name);
   SetLabel(name);
-
-  collapse_toggle_->SetState(TriangleToggle::State::kExpanded,
-                             TriangleToggle::InitialStateUpdate::kReplaceInitialState);
```
vickyliu-go4it, line 34: Thanks for making this change. May I know what is the default state of `GraphTrack` now? Is it expanded by default?

vickyliu-go4it (approved):
LGTM. Thanks!

src/OrbitGl/GraphTrack.cpp
florian-kuebler: I've split the state into collapsed/expanded and collapsable/not-collapsable. The default is expanded and collapsable.

